PR #3629, opened 2020-12-08
t: Refactor and mark multiple test modules for 100% statement coverage

okurz:
* t: Refactor and mark multiple test modules for 100% statement coverage
* t: Use more concise one-line Selenium WDKeys init as well
* t: Use more concise one-line Selenium driver initialization (2nd)

### DIFF
--- a/t/24-worker-jobs.t
+++ b/t/24-worker-jobs.t
@@ -47,8 +47,8 @@ sub wait_for_job {
     my $error;
     my $timer = Mojo::IOLoop->timer(
         $timeout => sub {
-            $error = "'$check_message' not happened after $timeout seconds";
-            Mojo::IOLoop->stop;
+            $error = "'$check_message' not happened after $timeout seconds";    # uncoverable statement
+            Mojo::IOLoop->stop;                                                 # uncoverable statement
         });
 
     # Watch the status event for changes
@@ -95,7 +95,7 @@ sub wait_until_uploading_logs_and_assets_concluded {
     has pool_directory  => undef;
 }
 {
-    package Test::FakeClient;
+    package Test::FakeClient;    # uncoverable statement count:2
     use Mojo::Base -base;
     has worker_id            => 1;
     has webui_host           => 'not relevant here';
@@ -104,7 +104,6 @@ sub wait_until_uploading_logs_and_assets_concluded {
     has sent_messages        => sub { [] };
     has websocket_connection => sub { OpenQA::Test::FakeWebSocketTransaction->new };
     has ua                   => sub { Mojo::UserAgent->new };
-    has url                  => sub { Mojo::URL->new };
     has register_called      => 0;
     has last_error           => undef;
     has fail_job_duplication => 0;
@@ -132,7 +131,7 @@ sub wait_until_uploading_logs_and_assets_concluded {
     }
 }
 {
-    package Test::FakeEngine;
+    package Test::FakeEngine;    # uncoverable statement count:2
     use Mojo::Base -base;
     has pid        => 1;
     has errored    => 0;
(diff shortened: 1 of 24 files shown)


okurz:
The overall coverage increases. The patch is not fully covered likely due to https://codecov.io/gh/os-autoinst/openQA/pull/3629/diff?src=pr&el=tree-more which is exactly what I like to cover with your help if we understand what this code section should be needed for if at all.

t/24-worker-jobs.t
```diff
@@ -95,7 +95,7 @@ sub wait_until_uploading_logs_and_assets_concluded {
     has pool_directory  => undef;
 }
 {
-    package Test::FakeClient;
+    package Test::FakeClient;                                                          # uncoverable statement count:2
```
Martchus, line 98: Just a note: That this and similar lines aren't accounted it likely a limitation of the coverage tracking.

t/24-worker-jobs.t
```diff
@@ -104,7 +104,7 @@ sub wait_until_uploading_logs_and_assets_concluded {
     has sent_messages        => sub { [] };
     has websocket_connection => sub { OpenQA::Test::FakeWebSocketTransaction->new };
     has ua                   => sub { Mojo::UserAgent->new };
-    has url                  => sub { Mojo::URL->new };
+    has url                  => sub { Mojo::URL->new };                                # uncoverable statement count:2
```
Martchus, line 107: I've just tried removing this line completely and the test still passes. So I suggest to remove the line completely.

t/24-worker-jobs.t
```diff
@@ -204,22 +204,23 @@ $engine_mock->redefine(
...
+sub upload_mock {
+    # uncoverable subroutine
+    my ($key, $self, @args) = @_;     # uncoverable statement
+    my $shared_hash = shared_hash;    # uncoverable statement
+
+    # uncoverable statement count:1
+    # uncoverable statement count:2
+    push @{$shared_hash->{$key}}, \@args;
+    shared_hash $shared_hash;                # uncoverable statement
+    return $shared_hash->{upload_result};    # uncoverable statement
```
Martchus, line 217: I suppose these statements are only not tracked because this function runs within a sub process. (Otherwise the use of shared has wouldn't make any sense.) So maybe we handle this in the same way as other sub processes?

t/24-worker-webui-connection.t
```diff
@@ -119,10 +119,6 @@ $client->on(
         }
         return undef;
     }
-    sub skip_job {
-        my ($self, $job_id, $reason) = @_;
-        $self->skipped_jobs->{$job_id} = $reason;
-    }
```
Martchus, line 125: You can also remove `has skipped_jobs            => sub { {}; };`. (I've just tested it.)

t/43-scheduling-and-worker-scalability.t
```diff
@@ -154,9 +154,15 @@ subtest 'assign and run jobs' => sub {
...
+        # this branch is uncovered in
+        # https://codecov.io/gh/os-autoinst/openQA/src/master/t/43-scheduling-and-worker-scalability.t
+        # , why should we keep it?
         is(scalar @$allocated, $job_count, 'each job has a worker assigned');
     }
     else {
+        # this branch is uncovered in
+        # https://codecov.io/gh/os-autoinst/openQA/src/master/t/43-scheduling-and-worker-scalability.t
+        # , why should we keep it?
         is(scalar @$allocated, $job_count, 'all jobs assigned and all workers busy');
```
Martchus, line 166: This test is configurable. One might configure an equal number of workers and jobs. In this case we expect that the scheduler assigns all jobs and all workers have a job assigned. One might also configure more jobs than workers or more workers than jobs. In these cases the checks are obviously more limited and if there are more jobs than workers we obviously need to schedule multiple times until all jobs are processes.

I really prefer to keep the test configurable because it was helpful in the past to reproduce issues with job assignment locally which we otherwise only saw in production. For that playing around with the parameters was very useful.

You may mark the blocks not executed within the CI as unrecoverable with the notes "only executed when there are more jobs than workers" and "only executed when the number of workers is jobs are equal".

Martchus (changes requested):

t/24-worker-jobs.t
okurz: even better :)

t/24-worker-jobs.t
okurz: I thought the problem is rather the mocking. But other mocked functions are still properly covered. I don't know what's wrong here

t/24-worker-jobs.t
okurz: Yes, I agree :)

t/24-worker-jobs.t
okurz: done

t/24-worker-webui-connection.t
okurz: thx, done

t/43-scheduling-and-worker-scalability.t
okurz: ok, done that. It's not pretty but I think it's ok to keep it with my "uncoverable" comments as it's really an exception which we don't have anywhere else in the same way

t/24-worker-jobs.t
Martchus: > I don't know what's wrong here

So may explanation that this runs in a sub process which prevents the coverage tracking it wrong? Note that with "handle this in the same way as other sub processes" I meant specifically to use `Devel::Cover::report() if Devel::Cover->can('report');` here.

But of course it is questionable whether we actually want this because this is just about test code anyways.

okurz: Well, I don't know where we actually start such subprocess. However, I tried with a `Devel::Cover::report() if Devel::Cover->can('report');` and that failed the tests reproducibly with

```
    not ok 14 - expected REST-API calls happened
    #   Failed test 'expected REST-API calls happened'
    #   at t/24-worker-jobs.t line 762.
    #          $got->[4]{json}{status}{test_order} = Does not exist
    #     $expected->[4]{json}{status}{test_order} = ARRAY(0x560f2f589050)
```

Maybe the additional line is making some internal REST-API call timeout. I would leave it as is if you are ok with that.

Martchus: I'm ok with leaving it as it is.

Out of my head I suppose the forking happens in the upload code of the worker (which is under test here).

okurz: ok, thx. Maybe you wanna give it a try or we can hack on this together another day. I would like that :) Wanna approve the PR?

Martchus (approved):

PR #3632, opened 2020-12-09
docker: Include the current web UI build alternatives

ilausuch:
Currently the web UI can be compiled in three different ways,
to be used with all its services in the same container using
apache as a reverse proxy with or without SSL certificate,
or with its services separately to be used with nginx as a
reverse proxy for example in the docker-compose case.

This PR provides a docker file for each of the different models.

https://progress.opensuse.org/issues/80520

### DIFF
--- a/docker/README.md
+++ b/docker/README.md
@@ -32,11 +32,18 @@ The Dockerfiles included in this project are for openSUSE:
 
     # Fedora and docker-compose
     docker build -t openqa_data ./openqa_data
-    docker build -t openqa_webui ./webui
     docker build -t openqa_worker ./worker
+
+    # Fedora
+    docker build -t openqa_webui ./webui
+
+    # Fedora providing SSL certificates
+    docker build -f webui/Dockerfile-ssl -t openqa_webui ./webui
+    # NOTE: you must provide in the same directory the servcer.crt, server.key and ca.crt
 
     # docker-compose only
     docker build -f webui/Dockerfile-lb -t openqa_webui_lb ./webui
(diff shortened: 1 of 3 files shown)


okurz (changes requested):
Two things:

* Before adding more and more Dockerfiles I suggest to get tests in place for all Dockerfiles, at least some static checks
* Instead of duplicating the same calls I suggest to 
[DRY](https://en.wikipedia.org/wiki/Don't_repeat_yourself). You are calling the same code multiple times, please [extract method](http://refactoring.com/catalog/extractMethod.html).

ilausuch:
I started to do a test over the current dockerfile
https://github.com/os-autoinst/openQA/pull/3633

Martchus (commented):
Because you've requested a review: I could only state what @okurz mentioned before.

ilausuch:
This PR is not longer necessary as we have a initial script for the webUI that covers all cases
https://github.com/os-autoinst/openQA/blob/master/container/webui/run_openqa.sh

okurz:
@ilausuch instead of closing PRs that fast I suggest to please stick to the suggestions we provide because also your new solution proposals lack the same points. And at the very least please amake reviewing a bit easier for us by linking the according PRs. It is as simple as mentioning the URL like this:
https://github.com/os-autoinst/openQA/pull/3737

I think it would be beneficial if you could put yourself into the shoes of reviewers and think about how others can read, understand and review your PRs same as you are very welcome to review all other's PRs.

PR #3633, opened 2020-12-10
docker: add web UI container test

ilausuch:
This test tests the webui container as separate services as they would
be used in an environment with an external load balancer. As provided by
docker-compose or would be done with kubernetes.

This test checks:
- The build of the images
- The execution of the different containers
- And that the services are correctly configured and listening on their
respective ports.

https://progress.opensuse.org/issues/80520

### DIFF
--- a/t/46-webui-containers.t
+++ b/t/46-webui-containers.t
@@ -1,0 +1,146 @@
+#!/usr/bin/env perl
+# Copyright (C) 2020 SUSE LLC
+#
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation; either version 2 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License along
+# with this program; if not, see <http://www.gnu.org/licenses/>.
+
+use Test::Most;
+use Test::Warnings ':report_warnings';
+use FindBin;
+use lib "$FindBin::Bin/../lib", "$FindBin::Bin/../../external/os-autoinst-common/lib";
+use Mojo::File qw(tempdir);
+use Cwd qw(cwd);
+END { clean(); }
+
+#plan skip_all => "set SCHEDULER_FULLSTACK=1 (be careful)" unless $ENV{SCHEDULER_FULLSTACK};
+
+my $workdir = tempdir();
+my $PWD     = cwd;
+my $prefix  = 'webui_' . int(rand(10000)) . '_';
+
+sub clean {
+    system("docker rm -f ${prefix}gru 2>/dev/null || echo");
+    system("docker rm -f ${prefix}livehandler 2>/dev/null || echo");
+    system("docker rm -f ${prefix}websockets 2>/dev/null || echo");
+    system("docker rm -f ${prefix}scheduler 2>/dev/null || echo");
+    system("docker rm -f ${prefix}webui 2>/dev/null || echo");
+    system("docker rm -f ${prefix}data 2>/dev/null || echo");
+    system("docker rm -f ${prefix}db 2>/dev/null || echo");
+    system("docker network rm ${prefix}testing 2>/dev/null || echo");
+}
+
+sub wait_for {
+    my ($container) = @_;
+    my $count = 30;
+    while (system("docker logs $container 2>&1 | grep Listening >/dev/null") != 0) {
+        sleep 1;
+        $count = $count - 1;
+        last if ($count == 0);
+    }
+
+    is(system("docker logs $container 2>&1 | grep Listening >/dev/null"), 0, "container $container is listening");
+}
+
+is(system("docker network create ${prefix}testing"), 0, 'docker cannot create the network');
+
+my $volumes
+  = "-v \"$workdir/data/factory:/data/factory\" -v \"$workdir/data/tests:/data/tests\" -v \"$PWD/docker/webui/conf:/data/conf:ro\"";
+
+is(system('docker', 'build', '-t', 'openqa_data',  'docker/openqa_data'), 0, 'data container image can be built');
+is(system('docker', 'build', '-t', 'openqa_webui', 'docker/webui'),       0, 'webui container image can be built');
+is(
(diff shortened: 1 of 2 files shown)


ilausuch:
Why is that if I add some line to .circleci/ci-packages.txt the ci/circleci: t step doesn't appear in the checks?
And also there is a problem with ci/circleci: cache with
Package 'perl-Mojolicious-8.65' not found.
Package 'perl-Perl-Tidy-20201001' not found.
?

okurz:
> Package 'perl-Mojolicious-8.65' not found.
> Package 'perl-Perl-Tidy-20201001' not found.

As I don't see this problem reproduced in my PRs I assume you would be able to solve that with rebasing on a more recent version of openQA master. I already suggested that many times in other of your PRs. In general it's a good practice to rebase on the main branch before you push your changes or create new pull requests. The main file in the main fork in https://github.com/os-autoinst/openQA/blob/master/.circleci/ci-packages.txt#L141 has perl-Mojolicious-8.66 which was brought in 8 days ago in https://github.com/os-autoinst/openQA/commit/aac49b57105dea4608d54967c03df58de3358e4a#diff-3370d5adc0943f8551d72bbdf89f62427b77ded3d8751d9e9bc476a9b32ef65f and https://github.com/os-autoinst/openQA/blob/master/.circleci/ci-packages.txt#L177 has perl-Perl-Tidy-20201202 from 6 days ago.

okurz (changes requested):
https://github.com/os-autoinst/openQA/pull/3633#issuecomment-742511120

ilausuch:
Perfect! you we're right, thanks

t/46-webui-containers.t
```diff
@@ -0,0 +1,54 @@
...
+    system('docker rm -f webui 2>/dev/null');
+    system('docker rm -f db 2>/dev/null');
+    system('docker network rm testing 2>/dev/null');
+    system('rm -rf workdir 2>/dev/null');
+}
+
+clean();
+
+is(system('docker network create testing'), 0, 'docker cannot create the network');
+is(system('mkdir workdir'), 0, 'cannot create the working dir');
```
kalikiana, line 34: Minor point, but you could use `mkdir` here w/o invoking a shell.

t/46-webui-containers.t
```diff
@@ -0,0 +1,54 @@
...
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License along
+# with this program; if not, see <http://www.gnu.org/licenses/>.
+
+use Test::Most;
+use Test::Warnings ':report_warnings';
+
+sub clean(){
+    system('docker rm -f gru 2>/dev/null');
```
kalikiana, line 21: These names are very generic and likely to delete existing containers. I think there should be a prefix and/ or a check like this at the top:

    plan skip_all => "set SCHEDULER_FULLSTACK=1 (be careful)" unless $ENV{SCHEDULER_FULLSTACK};

t/46-webui-containers.t
```diff
@@ -0,0 +1,54 @@
...
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License along
+# with this program; if not, see <http://www.gnu.org/licenses/>.
+
+use Test::Most;
+use Test::Warnings ':report_warnings';
+
+sub clean(){
```
Martchus, line 20: Maybe this function should be invoked within the end block to be sure it is also invoked when an exception is thrown. Additionally, a loop would help to reduce the repetition here.

t/46-webui-containers.t
```diff
@@ -0,0 +1,54 @@
...
+    system('docker rm -f gru 2>/dev/null');
+    system('docker rm -f livehandler 2>/dev/null');
+    system('docker rm -f websockets 2>/dev/null');
+    system('docker rm -f scheduler 2>/dev/null');
+    system('docker rm -f webui 2>/dev/null');
+    system('docker rm -f db 2>/dev/null');
+    system('docker network rm testing 2>/dev/null');
+    system('rm -rf workdir 2>/dev/null');
```
Martchus, line 28: We could do that in Perl without forking.

t/46-webui-containers.t
```diff
@@ -0,0 +1,54 @@
...
+}
+
+clean();
+
+is(system('docker network create testing'), 0, 'docker cannot create the network');
+is(system('mkdir workdir'), 0, 'cannot create the working dir');
+
+my volumes = "-v $(pwd)/workdir/data/factory:/data/factory -v $(pwd)/workdir/data/tests:/data/tests -v $(pwd)/docker/webui/conf:/data/conf";
```
Martchus, line 36: Hopefully `$(pwd)` doesn't return a path with spaces here. Besides, you could also simply use `$PWD` instead of invoking `pwd` three times.

Martchus (changes requested):
+ everything the others have already mentioned

ilausuch:
I pushed a commit without taking in consideration your comments, sorry I did before to read all of them, I had to be sure of fix some problems I found. However all your comments are still valid, and I will incorporate them in the next comment. Thanks

ilausuch:
Again I pushed a commit without your comments because I was changing things deeply, I'll include them in my next one, sorry

t/46-webui-containers.t
ilausuch: Thanks, but in the new version it won't be necessary. I consider more interesting using temporary directories 
my $workdir = tempdir( );

t/46-webui-containers.t
ilausuch: Won't be necessary because the reason I mentioned with the temp directories Thanks

t/46-webui-containers.t
okurz:
```perl

okurz (commented):
@ilausuch I strongly suggest you look into the conceptual design problems before trying to beautify the details. I honestly don't know what the right approach for "test containers" is but if I would invest more time in the story this is the first thing I would research about, same as we discussed in chat, in review and also in video-chat yesterday. Do you have any news regarding the general approach or state-of-art in the industry?

t/46-webui-containers.t
ilausuch: Applied

t/46-webui-containers.t
ilausuch: Done

ilausuch:
I think this PR could be closed because the problem to test the openQA containers is covered by https://github.com/os-autoinst/os-autoinst-distri-openQA/pull/60

PR #3634, opened 2020-12-10
docker: Add a test for openQA workers containers

ilausuch:
This PR add a test for the openQA workers containers

The current execution with an appropriate environment and configuration
is

```
[info] [pid:21] worker 1:
 - config file:           /etc/openqa/workers.ini
 - worker hostname:       f2fa029ca2ae
 - isotovideo version:    20
 - websocket API version: 1
 - web UI hosts:          http://localhost:9526
 - class:                 qemu_i386,qemu_x86_64
 - no cleanup:            no
 - pool directory:        /var/lib/openqa/pool/1
API key and secret are needed for the worker connecting
http://localhost:9526
```

The test probes that the container image can be build and can be used
deploying a worker. But because the is not a web UI is going to fail
in the execution. The test checks that show some part of this information

https://progress.opensuse.org/issues/80682

### DIFF
--- a/t/46-worker-container.t
+++ b/t/46-worker-container.t
@@ -1,0 +1,28 @@
+#!/usr/bin/env perl
+# Copyright (C) 2020 SUSE LLC
+#
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation; either version 2 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License along
+# with this program; if not, see <http://www.gnu.org/licenses/>.
+
+use Test::Most;
+use Test::Warnings ':report_warnings';
+use File::Copy;
+use File::Temp qw/ tempdir /;
+
+my $dir = tempdir( CLEANUP => 1 );
+copy("./t/data/workers.ini", "$dir/workers.ini") or die "Copy failed: $!";
+
+is(system('docker', 'build', '-t', 'openqa_worker', 'docker/worker'), 0, 'worker container image can be built');
+is(system("docker run -t -v $dir/workers.ini:/data/conf/workers.ini openqa_worker | grep 'worker connecting'"), 0, 'worker container starts');
+
+done_testing();
--- a/.circleci/ci-packages.txt
+++ b/.circleci/ci-packages.txt
@@ -256,3 +256,4 @@ python3-setuptools-40.5.0
 python3-yamllint-1.22.1
 ShellCheck-0.7.1
 sqlite3-3.28.0
+docker


t/46-worker-container.t
```diff
@@ -0,0 +1,23 @@
...
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License along
+# with this program; if not, see <http://www.gnu.org/licenses/>.
+
+use Test::Most;
+use Test::Warnings ':report_warnings';
+
+is(system('docker', 'build', '-t', 'openqa_worker', 'docker/worker'), 0, 'worker container image can be built');
+is(system('docker run -t -v $(pwd)/t/data/workers.ini:/data/conf/workers.ini openqa_worker | grep "worker connecting"'), 0, 'worker container starts');
```
okurz, line 21: Better not rely on the current directory being the root of the working copy. Better use FindBin as in other tests.

okurz (changes requested):
As can be seen in CI output this does not work because there is no docker aemon running:

```
[16:35:42] t/46-worker-container.t ................................... 1/? 
#   Failed test 'worker container image can be built'
#   at t/46-worker-container.t line 20.
#          got: '256'
#     expected: '0'
docker: Cannot connect to the Docker daemon at unix:///var/run/docker.sock. Is the docker daemon running?.
See 'docker run --help'.

#   Failed test 'worker container starts'
#   at t/46-worker-container.t line 21.
#          got: '256'
#     expected: '0'
# Looks like you failed 2 tests of 3.
                                                                            [16:35:42] t/46-worker-container.t ................................... Dubious, test returned 2 (wstat 512, 0x200)
Failed 2/3 subtests
```

There were already many other related comments. Please take them into consideration. It seems like you created another PR without addressing all the comments we already provided.

t/46-worker-container.t
kalikiana, line 21: These names are very generic and likely to delete existing containers. I think there should be a prefix and/ or a check like this at the top:

    plan skip_all => "set SCHEDULER_FULLSTACK=1 (be careful)" unless $ENV{SCHEDULER_FULLSTACK};

t/46-worker-container.t
```diff
@@ -0,0 +1,28 @@
...
+use Test::Most;
+use Test::Warnings ':report_warnings';
+use File::Copy;
+use File::Temp qw/ tempdir /;
+
+my $dir = tempdir( CLEANUP => 1 );
+copy("./t/data/workers.ini", "$dir/workers.ini") or die "Copy failed: $!";
+
+is(system('docker', 'build', '-t', 'openqa_worker', 'docker/worker'), 0, 'worker container image can be built');
+is(system("docker run -t -v $dir/workers.ini:/data/conf/workers.ini openqa_worker | grep 'worker connecting'"), 0, 'worker container starts');
```
okurz, line 26: As discussed you don't need the tempdir at all. I would not care if etc/workers.ini is changed by the worker as it should not happen anyway.

As you asked in video chat regarding waiting for something in logs you could do more or less the following pseudo-code:

```perl
for (1 .. timeout) {
  last if grep 'my expected string' $logfile;
  sleep .1;
}
```

okurz (changes requested):

ilausuch:
This problem is covered by https://github.com/os-autoinst/os-autoinst-distri-openQA/pull/60

PR #3635, opened 2020-12-10
Ensure usage of results disk is below configurable threshold

Martchus:
Just sketching the basic idea here, not really tested/implemented yet.

See [poo#76984](https://progress.opensuse.org/issues/76984)


okurz (commented):
looks nice so far. If I understand this right then you plan to implement additional cleaning next to the already, independant result cleanup. This would ensure to trigger additional result cleanup if there is not enough free space, right? So this is not – what we can consider independantly – to *not* cleanup results when there already is still enough space available, right?

Martchus:
> So this is not – what we can consider independantly – to not cleanup results when there already is still enough space available, right?

Yes, it only deletes *more* but never *less*. And yes, I would consider that independently - especially because the related ticket https://progress.opensuse.org/issues/76984 does not really cover this and is already big enough. I'd also like to note that especially for assets cleaning up less is likely not useful for us because in production we always over-allocate.

okurz:
I agree. With this you certainly address the more important issue.

lib/OpenQA/Task/Job/Limit.pm
```diff
@@ -146,4 +151,106 @@ sub _limit_screenshots {
...
+    my $free_percentage   = $available_bytes / $total_bytes * 100;
+    my $margin_percentage = $free_percentage - $min_free_percentage;
+    my $margin_bytes      = $margin_percentage / 100 * $total_bytes;
+    $job->note(df                => $df);
+    $job->note(margin_percentage => $margin_percentage);
+    $job->note(margin_bytes      => $margin_bytes);
+    return $margin_bytes;
+}
+
+sub _ensure_results_below_thresold {
```
okurz, line 171: s/thresold/threshold/

lib/OpenQA/Task/Job/Limit.pm
```diff
@@ -146,4 +151,106 @@ sub _limit_screenshots {
...
+    while (my $job = $relevant_jobs->next) {
+        log_debug 'Deleting video of job ' . $job->id;
+        return $job->finish('Done after deleting videos from non-important jobs') if $margin_bytes += $job->delete_video;
+    }
+
+    log_debug "Deleting results from non-important jobs startinng from oldest job (balance is $margin_bytes)";
+    # TODO: Do we want to delete the whole job or only logs or all results but keep the DB entry? The phrasing
+    #       in the progress ticket sounds like the last option. However, we currently don't have this "grade of
+    #       cleanup" implemented so that would be something new.
+    # TODO: Implement delete_results.
```
okurz, line 231: would you recommend to implement the "delete only results but keep job in db" first?

Martchus: I plan to make it an extra commit similar to `Add function to determine screenshots exclusively used by a job`. Same counts for the video deletion. (I split things up as soon as I have them tested.)

okurz: Alright, sounds good. I suggest then to merge these features step by step if possible as likely something can break badly and we will have a hard time realizing … and then there is christmas vacation … ;)

Martchus:
The unit tests should now cover all the branches. It still lacks testing with real data and documentation and the customization point for providing an own df script. The code has some caveats/simplifications (see comments within the code) but at least it looks clean and shouldn't be hard to understand.

etc/openqa/openqa.ini
```diff
@@ -144,6 +144,9 @@ blocklist = job_grab job_done
 # Specify the number of screenshot ID ranges (with a size as configured by screenshot_cleanup_batch_size) to process in a single Minion
 # job (reduce to avoid Minion jobs from running very long and possibly being interrupted, increase to reduce the number of Minion jobs)
 #screenshot_cleanup_batches_per_minion_job = 450
+# Extends the job result cleanup to ensure the partition results are stored on does not become too full
+# (still experimental, relies on df)
+#results_min_free_disk_space_percentage = 0
```
kalikiana, line 149: I wonder if it wouldn't be more consistent to go with gigabytes like we do for asset sizes. `_percentage` is clear enough in that sense. Just wanted to mention that it seems like a deviation.

lib/OpenQA/Schema/Result/Jobs.pm
```diff
@@ -1174,9 +1181,69 @@ sub delete_logs {
...
+    my $job_id                          = $self->id;
+    my $exclusively_used_screenshot_ids = $self->exclusively_used_screenshot_ids;
+    my $schema                          = $self->result_source->schema;
+    my $screenshots                     = $schema->resultset('Screenshots');
+    my $dbh                             = $schema->storage->dbh;
+    my $delete_screenshot_query         = $dbh->prepare('DELETE FROM screenshots WHERE id = ?');
+    my $imagesdir                       = imagesdir();
+    $self->screenshot_links->delete;
+    for my $screenshot_id (@$exclusively_used_screenshot_ids) {
+        # get rid of screenshot, similar to OpenQA::Task::Job::Limit
```
kalikiana, line 1219: Why do we need to do this here? Existing code seems to assume that `screenshots left unused are deleted later in the job limit task`.

etc/openqa/openqa.ini
Martchus: When it comes to preserving a minimum amount of free disk space tools usually let you specify a percentage so I preferred this.

lib/OpenQA/Schema/Result/Jobs.pm
Martchus: We need to delete screenshots here immediately because otherwise we would not be able to see a difference immediately when re-checking df again.

kalikiana: Right, that makes sense in hindsight. Thanks for clarifying!

lib/OpenQA/Schema/Result/Jobs.pm
```diff
@@ -1174,9 +1181,69 @@ sub delete_logs {
...
+        my $screenshot_filename = $screenshots->find($screenshot_id)->filename;
+        my $screenshot_path     = catfile($imagesdir, $screenshot_filename);
+        my $thumb_path = catfile($imagesdir, dirname($screenshot_filename), '.thumbs', basename($screenshot_filename));
+        next unless eval { $delete_screenshot_query->execute($screenshot_id); 1 };
+        unless (unlink($screenshot_path, $thumb_path) == 2) {
+            log_debug qq{Can't remove screenshot "$screenshot_filename"} if -f $screenshot_filename;
+            log_debug qq{Can't remove thumbnail "$thumb_path"}           if -f $thumb_path;
```
kalikiana, line 1226: I guess these two cases aren't covered by the tests yet

okurz:
could you please take a look into the coverage drop for lib/OpenQA/Schema/Result/Jobs.pm and lib/OpenQA/Worker/Engines/isotovideo.pm

Martchus:
The coverage drop in `lib/OpenQA/Worker/Engines/isotovideo.pm` (a file I didn't touch) is likely due to rebasing multiple times because the coverage diff corresponds to changes on master.

In `lib/OpenQA/Schema/Result/Jobs.pm` it is really just the 2 error cases already mentioned by @kalikiana.

lib/OpenQA/Schema/Result/Jobs.pm
```diff
@@ -1174,9 +1181,69 @@ sub delete_logs {
...
+    my $screenshots                     = $schema->resultset('Screenshots');
+    my $dbh                             = $schema->storage->dbh;
+    my $delete_screenshot_query         = $dbh->prepare('DELETE FROM screenshots WHERE id = ?');
+    my $imagesdir                       = imagesdir();
+    $self->screenshot_links->delete;
+    for my $screenshot_id (@$exclusively_used_screenshot_ids) {
+        # get rid of screenshot, similar to OpenQA::Task::Job::Limit
+        my $screenshot_filename = $screenshots->find($screenshot_id)->filename;
+        my $screenshot_path     = catfile($imagesdir, $screenshot_filename);
+        my $thumb_path = catfile($imagesdir, dirname($screenshot_filename), '.thumbs', basename($screenshot_filename));
```
Amrysliu, line 1222: how about  `my $thumb_path = catfile($dirname($screenshot_path), '.thumbs', basename($screenshot_filename));`

Martchus:
I'm removing the WIP here because the unit tests should be complete now and merging it shouldn't disturb anyone. A test with real data, further documentation and allowing to run a custom script instead of df are still on the TODO list. However, I'd like to merge this at some point before it grows too big.

By the way, this approach could be extended in the future to copy old results to a storage server instead and only delete anything if the storage server is full as well.

Amrysliu (approved):
LGTM

lib/OpenQA/ScreenshotDeletion.pm
```diff
@@ -0,0 +1,51 @@
...
+    my ($self, $screenshot_id, $screenshot_filename) = @_;
+
+    my $screenshot_path = catfile($self->{_imagesdir}, $screenshot_filename);
+    my $thumb_path      = catfile(dirname($screenshot_path), '.thumbs', basename($screenshot_filename));
+
+    # delete screenshot in database first
+    # note: This might fail due to foreign key violation because a new screenshot link might
+    #       have just been created. In this case the screenshot should not be deleted in the
+    #       database or the file system.
+    return undef unless eval { $self->{_deletion_query}->execute($screenshot_id); 1 };
```
kalikiana, line 43: Can/ should we keep in mind other types of error? Although after typing this question, I guess the statement is prepared before so other errors should occur in the new function anyway.

kalikiana (approved):

kalikiana:
Apparently #3694 and this PR are hitting [poo#66718](https://progress.opensuse.org/issues/66718)

okurz:
> By the way, this approach could be extended in the future to copy old results to a storage server instead and only delete anything if the storage server is full as well.

yes, that was the idea :)

okurz (approved):
It's already way too big to properly review in detail so  I can only go with approval :D

lib/OpenQA/ScreenshotDeletion.pm
Martchus: I know; unexpected errors should be logged. However, I'm just moving this code around.

kalikiana:
> It's already way too big to properly review in detail so I can only go with approval :D

A big-boned PR 😎

Seems like this and other PRs are just stuck on OBS now...

Martchus:
Yes. At least it builds for TW again.

okurz:
Likely https://github.com/os-autoinst/openQA/pull/3690 caused the conflict but that should be easy to fix.

Martchus:
Now it fails to build the cache for the CI. Maybe I'll merge it anyways as the package build succeeds. (The rebasing-diff should be entirely tested by building the packages.)

okurz:
Just be patient and let it run :)

Martchus:
But it is stuck within that state because the old Mojolicious version is not present anymore to make the cache and only merging the dependency PR would fix that. However, it is not 100 % clear yet whether we actually like to merge it as-is. So I'm still merging this PR because as mentioned in my previous comment it should be fine.

PR #3636, opened 2020-12-11
Allow configuring the worker to terminate after all jobs have been processed

Martchus:
See https://progress.opensuse.org/issues/80986

Besides the unit tests I've also conducted 2 full test runs (one with and one without the configuration) and both worked as expected.

### DIFF
--- a/lib/OpenQA/Worker/Settings.pm
+++ b/lib/OpenQA/Worker/Settings.pm
@@ -52,6 +52,7 @@ sub new {
 
     # read global settings from environment variables
     $global_settings{LOG_DIR} = $ENV{OPENQA_WORKER_LOGDIR} if $ENV{OPENQA_WORKER_LOGDIR};
+    $global_settings{TERMINATE_AFTER_JOBS_DONE} //= $ENV{OPENQA_WORKER_TERMINATE_AFTER_JOBS_DONE};
 
     # read global settings specified via CLI arguments
     $global_settings{LOG_LEVEL} = 'debug' if $cli_options->{verbose};
--- a/Makefile
+++ b/Makefile
@@ -81,11 +81,10 @@ install-generic:
 	for i in systemd/*.{service,target,timer}; do \
 		install -m 644 $$i "$(DESTDIR)"/usr/lib/systemd/system ;\
 	done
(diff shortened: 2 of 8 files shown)


okurz (changes requested):
the code you changed looks fine, I would appreciate the following:

* Extend etc/openqa/workers.ini
* Optionally extend documentation. Not sure if we have a section for possible worker config, feel free to add one
* Put more details in the git commit message to explain the use case

Martchus:
Ok, and see https://github.com/os-autoinst/openQA/pull/3639#issuecomment-744259023 regarding the warning. It is just a deprecation.

okurz:
take a look into https://github.com/os-autoinst/openQA/blob/master/Makefile#L84 for how we create a "no-cleanup" convenience systemd service definition.

lib/OpenQA/Worker/Settings.pm
```diff
@@ -52,6 +52,7 @@ sub new {
 
     # read global settings from environment variables
     $global_settings{LOG_DIR} = $ENV{OPENQA_WORKER_LOGDIR} if $ENV{OPENQA_WORKER_LOGDIR};
+    $global_settings{TERMINATE_AFTER_JOBS_DONE} //= $ENV{OPENQA_WORKER_TERMINATE_AFTER_JOBS_DONE};
```
kalikiana, line 55: We should probably have an uncommented example in the default `workers.ini`?

okurz: has been done

okurz (approved):

okurz:
I have approved but added a label so that you can show "acceptance tests", e.g. from your local env or a staging env.

lib/OpenQA/Worker/Settings.pm
Martchus: I have added a comment to `workers.ini`.

Martchus:
As mentioned, I've done 2 tests for both configurations locally. There's nothing special to see in the test results which I could share. What I still want to test is the systemd service file.

Martchus:
I've tested the systemd settings locally and it works: https://w3.suse.de/~mkittler/openQA/worker-log-with-restart.log

kalikiana (approved):

Martchus:
The package build seems to fail because of the deprecation warning.

PR #3639, opened 2020-12-14
Dependency cron 2020-12-14

openqabot:


okurz:
@kraih can you help here?

okurz:
Same problem as in https://github.com/os-autoinst/openQA/pull/3637

Martchus:
Seems like they deprecated something and now the "turn warnings into errors" approach takes effect. Unfortunately the problematic usage is in `AssetPack`.

andrii-suse:
It seems that logs are confusing (at least for me), but just for reference - the actual error message is probably : `Stopping openqa-webapi process because a Perl warning occurred: Mojolicious::Routes::Route::route is DEPRECATED in favor of Mojolicious::Routes::Route::any at /usr/lib/perl5/vendor_perl/5.26.1/Mojolicious/Plugin/AssetPack.pm line 14`

Martchus:
I've been adding a commit to avoid making use of deprecated functions a test error. We should still try to improve it of course but the deprecated function is used within the `AssetPack` module and not within our code so it will likely take some time to get the fix in.

t/lib/OpenQA/Test/Utils.pm
```diff
@@ -215,7 +215,9 @@ sub _setup_sub_process {
     $0 = $process_name;
     note "PID of $process_name: $$";
     $SIG{__WARN__} = sub {
-        log_error "Stopping $process_name process because a Perl warning occurred: @_";
+        my $warning = "@_";
+        log_error "Stopping $process_name process because a Perl warning occurred: $warning"
+          unless $warning =~ qr/DEPRECATED/i;
```
okurz, line 220: as discussed please create a ticket for that. Then I am completely fine to skip that. Good.

okurz (approved):

t/lib/OpenQA/Test/Utils.pm
kalikiana: This doesn't seem to work:

```perl
#   Failed test 'no (unexpected) warnings (via done_testing)'
#   at t/39-scheduled_products-table.t line 88.
# Got the following unexpected warnings:
#   1: Mojolicious::Routes::Route::route is DEPRECATED in favor of Mojolicious::Routes::Route::any at /usr/lib/perl5/vendor_perl/5.26.1/Mojolicious/Plugin/AssetPack.pm line 14.
#   2: Mojolicious::Routes::Route::via is DEPRECATED in favor of Mojolicious::Routes::Route::methods at /usr/lib/perl5/vendor_perl/5.26.1/Mojolicious/Plugin/AssetPack.pm line 14.
```

okurz:
still failing on warnings

okurz:
still fails in https://app.circleci.com/pipelines/github/os-autoinst/openQA/5262/workflows/72308481-e2b6-48ea-b23c-a3f5fd65d0a2/jobs/49914/steps

Martchus:
Yes, so far I only added a workaround to my hack for Selenium tests. However, the usual `use Test::Warnings ':report_warnings';` is still blocking. Not sure how to add an exception there. I also want to avoid touching all test files for that.

kalikiana:
> Yes, so far I only added a workaround to my hack for Selenium tests. However, the usual `use Test::Warnings ':report_warnings';` is still blocking. Not sure how to add an exception there. I also want to avoid touching all test files for that.

The [docs on Test::Warnings](https://metacpan.org/pod/Test::Warnings#CAVEATS) seem to confirm there's no support for that( yet). And we'd need to add an `if $ENV{}` to the import in every file to disable it. 🤔

Martchus:
Ok, so I'm not blind. (I've read the docs before as well.)

okurz:
So let's fix AssetPack?

okurz:
https://build.opensuse.org/request/show/855720 is already in staging. I think we can wait for that if we don't have a good workaround

Martchus:
By the way, that's the ticket I've been creating for this: https://progress.opensuse.org/issues/81038

And yes, without a workaround we would naturally wait until it is accepted.

PR #3640, opened 2020-12-14
Avoid worker from appearing stuck (e.g. not reacting to SIGTERM) on cache check

Martchus:
When the worker checks for the availability of the cache service the normal
behavior of the cache service client is applied (huge number of attempts,
high inactivity timeout). This is not nice for the availability check
because it leads to the worker getting stuck and not even reacting to
signals.

---

By the way, to reproduce this, just configure the worker to use the cache
service but don't start it. Then you can not stop the worker via `Ctrl+C`.

### DIFF
--- a/lib/OpenQA/Worker.pm
+++ b/lib/OpenQA/Worker.pm
@@ -260,7 +260,8 @@ sub init {
     # check the setup (pool directory, worker cache, ...)
     # note: This assigns $self->current_error if there's an error and therefore prevents us from grabbing
     #       a job while broken. The error is propagated to the web UIs.
-    $self->check_availability();
+    $self->configure_cache_client;
+    $self->check_availability;
 
     # register error handler to stop the current job when a critical/unhandled error occurs
     Mojo::IOLoop->singleton->reactor->on(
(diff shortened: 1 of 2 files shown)


lib/OpenQA/Worker.pm
```diff
@@ -601,7 +601,11 @@ sub check_availability {
 
     # check whether the cache service is available if caching enabled
     if ($self->settings->global_settings->{CACHEDIRECTORY}) {
-        my $error = OpenQA::CacheService::Client->new->info->availability_error;
+        # reduce default timeout and attempts availability check
```
okurz, line 604: Could you also explain in the comment why we need the code?

lib/OpenQA/Worker.pm
```diff
@@ -601,7 +601,11 @@ sub check_availability {
...
+        my $client = OpenQA::CacheService::Client->new;
+        $client->attempts(1);
+        $client->ua->inactivity_timeout(10);
```
okurz, line 607: Hm, a magic number. Why 10? Make it configurable?

okurz (commented):
suggestions resolved, awaiting coverage results …

PR #3641, opened 2020-12-14
Allow stopping worker so it still finishes assigned jobs

Martchus:
~~Based #3636 and #3640. Just check the last commit for the actual change.
5c3df15~~

---

Allow stopping worker so it still finishes assigned jobs

* This allows deploying new worker versions or applying configuration
  changes without interrupting currently running job
* See https://progress.opensuse.org/issues/80908

See [below](#issuecomment-748147996) for further details.

---

When using systemd, one could simply use the `openqa-worker@`
service and send SIGHUP via systemctl, e.g.
`systemctl kill --signal SIGHUP openqa-worker@2.service`.

WIP because unit tests are missing or need adaption. However, I've already
tested the mentioned `systemctl` command.


Martchus:
It would likely be possible to make systemd send the `SIGHUB` signal on a regular service restart. However, then we'd likely also increase the timeout to avoid the worker from getting killed and we likely don't generally want to.

By the way, I simply chose `SIGHUB` because NGINX is also using it for reloading the config and restarting processes: http://nginx.org/en/docs/control.html  
(Of course NGINX doesn't rely on systemd restarting the service because they only restart worker services.)

script/worker
```diff
@@ -133,8 +132,8 @@ $worker->log_setup_info();
 sub catch_exit {
     my ($sig) = @_;
 
-    log_info("Quit due to signal $sig");
-    $worker->stop(WORKER_COMMAND_QUIT);
+    log_info("Received signal $sig");
+    $worker->handle_signal($sig);
 }
```
okurz, line 137: As we do not know anymore what the signal will result in I suggest to delete the function completely and instead use the new method `$worker->handle_signal` as signal handler directly … or the other way around as `handle_signal` would only be called from here anyway. You might want to keep it in the worker though to simplify reaching it during test runs for higher coverage

okurz (changes requested):

okurz:
> service and send SIGHUB via systemctl, e.g.
> `systemctl kill --signal SIGHUP openqa-worker@2.service`.

mind the spelling, "HUP" not "HUB"

lib/OpenQA/Constants.pm
```diff
@@ -66,10 +66,11 @@ use constant {
     WORKER_SR_BROKEN        => 'worker broken',    # worker setup is generally broken, e.g. cache service not started
     WORKER_SR_DONE          => 'done',             # backend exited normally
     WORKER_SR_DIED          => 'died',             # backend died
+    WORKER_SR_FINISH_OFF    => 'finish-off',       # the worker is supposed to terminate after finishing assigned jobs
```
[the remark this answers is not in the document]
Martchus: I've chosen it because the worker will also turn itself off after finishing the current jobs. So I was going for the 2nd definition "to do the last part of something; to make something¹ end by doing one last thing²" from https://www.oxfordlearnersdictionaries.com/definition/english/finish-off?q=finish+off.

---

¹ execution of the worker
² processing assigned jobs

script/worker
Martchus: I want to keep it in the worker because then I can simply call `handle_signal` from within the tests. And no, `handle_signal` can not be assigned directly to the signal handler because it needs the worker as first argument.

lib/OpenQA/Worker.pm
```diff
@@ -711,7 +715,7 @@ sub _handle_job_status_changed {
 
             # ensure we actually skip the next jobs in the queue if user stops the worker with Ctrl+C right
             # after the last job has concluded
-            $reason = 'worker terminates' if $reason eq WORKER_SR_DONE;
+            $reason = 'worker terminates' if $reason eq WORKER_SR_DONE && !$self->{_finishing_off};
```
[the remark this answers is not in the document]
Martchus: I'm going to cover everything. Since it seems nobody raised any objections about the approach in general I'll continue with that now.

script/worker
okurz: ok, then maybe `sub catch_exit { $worker->handle_signal }`? (or is it `sub catch_exit { $worker->handle_signal(@_) }`?)

Martchus: Ok, I make it an one-liner. (I actually already have it but I'm waiting to push because I want to check the new coverage report first.)

lib/OpenQA/Worker.pm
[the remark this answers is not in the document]
Martchus: This is now definitely tested. (I put a `die` here and the tests fails as expected.)

okurz:
sorry, I am not sure I understand what this is offering on top of https://github.com/os-autoinst/openQA/pull/3636 . Is this like an alternative to the openqa-worker-auto-restart service?

Martchus:
I extended the commit message of `Allow stopping worker so it still finishes assigned jobs` to explain the benefit compared to `TERMINATE_AFTER_JOBS_DONE`:
* The TERMINATE_AFTER_JOBS_DONE setting only stops the service *after*
  at least one job has been processed. So it has no affect on idling
  workers which would therefore still process at least one job with
  the old code/configuration and only terminate after that.
* This change allows stopping the worker via SIGHUP which not only
  terminates the worker after processing its current jobs but also
  immediately if it is currently idling.

And yes, this now turned into an alternative to the `TERMINATE_AFTER_JOBS_DONE` setting. In the first place I wanted to make the worker to react to `SIGHUP` only when idling so `TERMINATE_AFTER_JOBS_DONE` would still be required to deal with busy workers. However, it seems simpler to implement and also simpler to use and test if `SIGHUP` has an effect regardless of the worker state.

Note that in production we don't want the service to just terminate but also restart. So we would still need to use the new `…-auto-restart` service introduced in https://github.com/os-autoinst/openQA/pull/3636/commits/2f2eecaef78869318f51941cd0a0e8beb24a39d9 for the `s/Restart=on-failure/Restart=always/` part.

okurz (approved):

Martchus:
By the way, that's how the salt diff to make use of this could look like (just a draft): https://gitlab.suse.de/openqa/salt-states-openqa/-/merge_requests/423/diffs

script/worker
```diff
@@ -130,12 +129,7 @@ $options{instance} ||= 1;
 my $worker = OpenQA::Worker->new(\%options);
 $worker->log_setup_info();
 
-sub catch_exit {
-    my ($sig) = @_;
-
-    log_info("Quit due to signal $sig");
-    $worker->stop(WORKER_COMMAND_QUIT);
-}
+sub catch_exit { $worker->handle_signal(@_) }
```
okurz, line 132: this seems to be the only line that is not covered by tests which is what we expected, right? Can you add an `# uncoverable statement` here?

kalikiana (approved):

PR #3643, opened 2020-12-15
Install qemu-hw-display-virtio-gpu-pci in openqa-bootstrap

asdil12:
Ticket: https://progress.opensuse.org/issues/81064

qemu-hw-display-virtio-gpu-pci is not required but only recommended.
So it needs to be explicitly installed.

### DIFF
--- a/script/openqa-bootstrap
+++ b/script/openqa-bootstrap
@@ -23,6 +23,11 @@ if [ "$(uname -m)" = "aarch64" ]; then
     zypper -n install --no-recommends qemu-uefi-aarch64
 fi
 
+# this was split into a separate package on newer dist versions - so install it if available
+if zypper -n search -x qemu-hw-display-virtio-gpu-pci ; then
+    zypper -n install --no-recommends qemu-hw-display-virtio-gpu-pci
+fi
+
 
 # setup database
 systemctl enable --now postgresql


Martchus (approved):
It is just 14 KiB so I guess it can not hurt to install it.

ggardet:
This is fine for Tumbleweed, but I think it would break on older distro, such as Leap 15.2.

asdil12:
The new code should only install the pkg if it is available.

ggardet (approved):
LGTM, thanks!

okurz (changes requested):
A shell script that is meant for generic openQA bootstrapping should not do package dependency management. Also the package is not useful nor necessary for many use cases. Maybe it's better solved with convenience sub-packages within the spec file.

Also this looks like it's really depending on what we do for os-autoinst by default so probably better solved in os-autoinst

asdil12:
@okurz I have no idea what you mean. How should adding more subpackages (to whatever spec) resole an issue that was created by someone moving a qemu functionality into a subpackage?
And how do you think could this be resolved in os-autoinst? By calling qemu in a way that doesn't require that dependency?

I'm not sure if we are talking about the same topic. Are you maybe talking about https://bugzilla.opensuse.org/show_bug.cgi?id=1180048 ? I thing this is a separate issue that has nothing to do with that odd openQA dependency problem.

okurz (approved):
No, this has nothing to do with https://bugzilla.opensuse.org/show_bug.cgi?id=1180048 which is solved within https://progress.opensuse.org/issues/81038 and https://github.com/os-autoinst/openQA/pull/3644

> And how do you think could this be resolved in os-autoinst? By calling qemu in a way that doesn't require that dependency?

No, not call qemu differently. I thought of specifying the dependencies in https://github.com/os-autoinst/os-autoinst/blob/master/dist/rpm/os-autoinst.spec in a new sub-package similar to https://github.com/os-autoinst/os-autoinst/blob/master/dist/rpm/os-autoinst.spec#L120 for example.

As an alternative consider to change it in the tests https://github.com/os-autoinst/os-autoinst-distri-opensuse/tree/master/tests/openqa

By the way, os-autoinst does not *require* qemu because there are many backends which do not have a relation to or dependency on qemu and it is for the user to define in what environment and with what backend os-autoinst should be run. For example for a s390x z/VM openQA worker no qemu is required, neither for baremetal support.

On the other hand of course something like "openqa-bootstrap" should provide something like a good convenient default support which can include qemu so that the user can "right away" run usual tests. 
I see now that we already install multiple other packages for the same reason. Hence I can approve because you are not making it much worse :) But I wonder if we would not be better off with installing with recommendations included.

kalikiana:
@asdil12 I think this needs a rebase

script/openqa-bootstrap
```diff
@@ -23,6 +23,11 @@ if [ "$(uname -m)" = "aarch64" ]; then
     zypper -n install --no-recommends qemu-uefi-aarch64
 fi
 
+# this was split into a separate package on newer dist versions - so install it if available
+if zypper -n search -x qemu-hw-display-virtio-gpu-pci ; then
```
ggardet, line 27: I Think this should be negated with a `!`:
`if ! zypper -n search -x qemu-hw-display-virtio-gpu-pci ; then`

PR #3644, opened 2020-12-15
Avoid using deprecated Mojolicious functions

Martchus:
* See https://progress.opensuse.org/issues/81038
* See https://progress.opensuse.org/issues/81068
* See https://bugzilla.opensuse.org/show_bug.cgi?id=1180048

### DIFF
--- a/script/openqa-cli
+++ b/script/openqa-cli
@@ -22,5 +22,7 @@ use lib "$FindBin::RealBin/../lib";
 
 use OpenQA::CLI;
 
+@ARGV = () if ($ARGV[0] // '') =~ qr/^(-h|(--)?help)$/;
+
 my $ret = OpenQA::CLI->new->run(@ARGV);
 exit $ret unless $ENV{MOJO_HELP};
--- a/templates/webapi/not_found.html.ep
+++ b/templates/webapi/not_found.html.ep
@@ -23,7 +23,7 @@ table pre {
             % $pattern = "+$pattern" if $depth;
             <pre><%= '  ' x $depth %><%= $pattern %></pre>
         </td>
-        <td><%= uc(join ',', @{$route->via || []}) || '*' %></td>
+        <td><%= uc(join ',', @{($route->can('methods') ? $route->methods : $route->via) || []}) || '*' %></td>
         <td>
             % my $api_description = stash('api_description') // {};
             % if (my $description = $api_description->{$route->name}) {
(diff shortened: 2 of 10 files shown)


okurz (approved):

templates/webapi/not_found.html.ep
```diff
@@ -23,7 +23,7 @@ table pre {
             % $pattern = "+$pattern" if $depth;
             <pre><%= '  ' x $depth %><%= $pattern %></pre>
         </td>
-        <td><%= uc(join ',', @{$route->via || []}) || '*' %></td>
+        <td><%= uc(join ',', @{$route->methods || []}) || '*' %></td>
```
okurz, line 26: this fails tests because the old Mojo version does not have this method yet. Instead of breaking support for older Mojo versions immediately I suggest to fallback to `via` if `methods` is not defined and only bump the version in `.circleci/ci-packages.txt`, not introduce a version requirement.

Martchus:
The CI passes so I was able to retain compatibility with the previous Mojolicious version. Let's see whether the packages build (currently still scheduled) to verify whether it also works with the current Mojolicious version.

Martchus:
The scripts test also fails locally. Looks like the new behavior of the command system doesn't work as intended, e.g.:

```
[martchus@linux-9lzf openQA]$ [deprecated] script/openqa --help
[2020-12-15 17:35:31.73324] [3027] [warn] Deprecated use of config key '[audit]: blacklist'. Use '[audit]: blocklist' instead
[2020-12-15 17:35:31.84132] [3027] [info] Loading external plugin ObsRsync
[2020-12-15 17:35:31.84443] [3027] [error] Routes not configured, plugin ObsRsync will be disabled
[2020-12-15 17:35:31.84448] [3027] [error] API routes not configured, plugin ObsRsync will not have API configured
Invalid command "--help".
[martchus@linux-9lzf openQA]$ [deprecated] echo $?
255
```

Martchus:
Finally, all tests pass in the CI (old Mojolicious version) and the Leap 15.2 package build works as well (using new Mojolicious version through our devel repo). The TW package build is currently blocked.

script/openqa-cli
```diff
@@ -22,5 +22,7 @@ use lib "$FindBin::RealBin/../lib";
 
 use OpenQA::CLI;
 
+@ARGV = () if ($ARGV[0] // '') =~ qr/^(-h|(--)?help)$/;
```
perlpunk, line 25: is this a leftover and you meant to replace it with `fix_top_level_help`?

Martchus: No. I wanted to avoid a dependency to `OpenQA::Utils` here (and all other dependencies it pulls).

perlpunk (approved):

Issue #3645, opened 2020-12-15
worker: Make log messages about "cache service unavailable" and "webui not available" consistent


okurz:
## Observation

If I start a worker with CACHEDIRECTORY configured but no cache service started (or when it crashed or failed) *and* multiple web UIs configured I see recurring info/warn messages about the retrying but no information that the cache service connection is retried, e.g.:

```
[error] Worker cache not available: Cache service info error: Connection refused
[info] CACHE: caching is enabled, setting up /var/lib/openqa/cache/e13.suse.de
[info] Project dir for host e13.suse.de is /var/lib/openqa/share
[info] CACHE: caching is enabled, setting up /var/lib/openqa/cache/yast-openqa.suse.cz
[info] Project dir for host yast-openqa.suse.cz is /var/lib/openqa/share
[info] Registering with openQA e13.suse.de
[warn] Failed to register at e13.suse.de - connection error: No route to host - trying again in 10 seconds
[info] Registering with openQA yast-openqa.suse.cz
[warn] Failed to register at yast-openqa.suse.cz - connection error: Connection refused - trying again in 10 seconds
[info] Registering with openQA e13.suse.de
[warn] Failed to register at e13.suse.de - connection error: No route to host - trying again in 10 seconds
[info] Registering with openQA yast-openqa.suse.cz
[warn] Failed to register at yast-openqa.suse.cz - connection error: Connection refused - trying again in 10 seconds
[info] Registering with openQA e13.suse.de
[warn] Failed to register at e13.suse.de - connection error: No route to host - trying again in 10 seconds
[info] Registering with openQA yast-openqa.suse.cz
[warn] Failed to register at yast-openqa.suse.cz - connection error: Connection refused - trying again in 10 seconds
```

## Suggestion

Likely in
https://github.com/os-autoinst/openQA/blob/3a581bf953ce34dd4082e740011cfdabb6a2bd4a/lib/OpenQA/Worker.pm#L215
we can handle a false return value of `check_availability` to log something like `log_info "cache service unavailable, retrying in $timeout"`

okurz:
Fixed by https://github.com/os-autoinst/openQA/pull/3646

PR #3646, opened 2020-12-15
Improve logging of availability error to show retry interval

Martchus:
Based on #3640

---

It will look like this:
```
[warn] Worker cache not available: Cache service info error: Connection refused - checking again for web UI 'http://localhost:9526' in 100.00 s
```

See #3645

### DIFF
--- a/lib/OpenQA/Worker/WebUIConnection.pm
+++ b/lib/OpenQA/Worker/WebUIConnection.pm
@@ -420,19 +420,20 @@ sub send_status {
     # send the worker status (unless the websocket connection has been lost)
     my $websocket_connection = $self->websocket_connection;
     return undef unless $websocket_connection;
+    my $status = $self->worker->status;
+
     $websocket_connection->send(
-        {json => $self->worker->status},
+        {json => $status},
         sub {
 
             # continue sending status updates (unless the websocket connection has been lost)
             return undef unless $self->websocket_connection;
 
             my $status_update_interval = $self->_calculate_status_update_interval;
-            $self->{_send_status_timer} = Mojo::IOLoop->timer(
-                $status_update_interval,
-                sub {
-                    $self->send_status;
-                });
+            my $webui_host             = $self->webui_host;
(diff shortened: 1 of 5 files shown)


lib/OpenQA/Worker/WebUIConnection.pm
```diff
@@ -420,19 +420,20 @@ sub send_status {
...
             my $status_update_interval = $self->_calculate_status_update_interval;
-            $self->{_send_status_timer} = Mojo::IOLoop->timer(
-                $status_update_interval,
-                sub {
-                    $self->send_status;
-                });
+            my $webui_host             = $self->webui_host;
+            log_warning "$status->{reason} - checking again for web UI '$webui_host' in $status_update_interval s"
+              if $status->{reason};
+            $self->{_send_status_timer} = Mojo::IOLoop->timer($status_update_interval, sub { $self->send_status });
```
kalikiana, line 436: Neat approach. And still simple!

How often does this re-try, though? Should it have a limit?

okurz: The actual retry behaviour is not impacted by this change. The retry happens on different intervals depending on the actual worker state and AFAIK it's without limit and so far is ok to not have a limit :)

okurz (approved):

kalikiana (approved):

okurz:
hm, this now implicitly merged https://github.com/os-autoinst/openQA/pull/3640 now as well. We don't have full patch coverage.

@Martchus as long as your memory about your changes is still fresh can you please try to come up with 100% coverage for your recently modified code lines?

Martchus:
Yes, I've seen it as well. Note that the code in `t/lib/OpenQA/Test/Utils.pm` is definitely run but can not be collected because it is a sub process. Not sure how gracefully we stop the sub process. Since it is just test code I tend to mark the block as uncoverable. Apart from that it is only the `$self->_setup_pool_directory` case in `check_availability`.

PR #3649, opened 2020-12-17
Improve worker tests

Martchus:
See particular commit messages and https://github.com/os-autoinst/openQA/pull/3646#issuecomment-746967848

I wasn't able to fix the coverage for the `c_worker` function. Somehow the coverage collection via `Devel::Cover::report() if Devel::Cover->can('report');` doesn't work. (Increasing the time before we send `SIGKILL` didn't work.)


Martchus:
> I wasn't able to fix the coverage for…

At least the included fixes seem to have worked (considering the codecov report).

PR #3652, opened 2020-12-21
worker: Fix clearing current error on status updates when idling

Martchus:
* Fix regression of d3d33b54affa617d9c54a2eac40804d9d1da2f51
* See https://progress.opensuse.org/issues/81228

### DIFF
--- a/lib/OpenQA/Worker.pm
+++ b/lib/OpenQA/Worker.pm
@@ -223,6 +223,7 @@ sub status {
     }
     else {
         $status{status} = 'free';
+        $self->current_error(undef);
     }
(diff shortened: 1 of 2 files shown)


Martchus:
By the way, I've also tested this locally by starting the worker while the cache service is offline, starting the cache service and waiting until the cache service is checked again (takes 100 seconds in my case). Then the worker does not only appear as online but also takes the next job.

Amrysliu (approved):

okurz (approved):

okurz:
OBS Factory build seems to be blocked for longer

PR #3655, opened 2020-12-31
Improve t/lib/OpenQA/Test/Utils.pm

kalikiana:
TBD (I don't know yet what my fixes will amount to if anything so please don't bother asking for a better title or description)

### DIFF
--- a/t/lib/OpenQA/Test/Utils.pm
+++ b/t/lib/OpenQA/Test/Utils.pm
@@ -449,7 +449,9 @@ sub create_worker {
 sub unstable_worker {
     my ($apikey, $apisecret, $host, $instance, $ticks, $sleep) = @_;
     note("Starting unstable worker. Instance: $instance for host $host");
-    $ticks = 1 unless defined $ticks;
+    # Block the event loop for the specified number of events or indefinitely if -1 is specified
+    $ticks = $ticks == -1 ? ~0 >> 1 : ($ticks // 1);
+    $sleep //= 0;
 
     my $h = _setup_sigchld_handler 'openqa-worker-unstable', start sub {
         _setup_sub_process 'openqa-worker-unstable';
@@ -462,18 +464,11 @@ sub unstable_worker {
             });
         setup_worker($worker, $host);
         $worker->init();
-        if ($ticks < 0) {
-            Mojo::IOLoop->singleton->start;
-        }
-        else {
-            Mojo::IOLoop->singleton->one_tick for (0 .. $ticks);
-        }
+        Mojo::IOLoop->singleton->one_tick for (0 .. $ticks);
+        sleep $sleep;
         Devel::Cover::report() if Devel::Cover->can('report');
-        if ($sleep) {    # uncoverable statement
-            1 while sleep $sleep;    # uncoverable statement
-        }    # uncoverable statement
     };
-    sleep $sleep if $sleep;
+    sleep $sleep;
     return $h;
 }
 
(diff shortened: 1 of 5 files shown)


t/lib/OpenQA/Test/Utils.pm
```diff
@@ -467,13 +465,14 @@ sub unstable_worker {
             });
         setup_worker($worker, $host);
         $worker->init();
+        Devel::Cover::report() if Devel::Cover->can('report');
         if ($ticks < 0) {
+            # This call will effectively never return
```
okurz, line 470:
```suggestion
            # uncoverable statement This call will effectively never return
```

t/lib/OpenQA/Test/Utils.pm
```diff
@@ -467,13 +465,14 @@ sub unstable_worker {
...
             Mojo::IOLoop->singleton->start;
         }
         else {
             Mojo::IOLoop->singleton->one_tick for (0 .. $ticks);
```
okurz, line 474: as you moved the report line above this one would also never be collected in coverage

t/lib/OpenQA/Test/Utils.pm
```diff
@@ -467,13 +465,14 @@ sub unstable_worker {
             });
         setup_worker($worker, $host);
         $worker->init();
```
okurz, line 467: maybe we can also add a signal handler here and call `Mojo::IOLoop->singleton->stop` on SIGTERM:

```perl

$SIG{TERM} = sub { Mojo::IOLoop->singleton->stop };
```

Then we should reach the `cover()` call reliably later on

t/lib/OpenQA/Test/Utils.pm
```diff
@@ -467,13 +465,14 @@ sub unstable_worker {
...
+        Devel::Cover::report() if Devel::Cover->can('report');
         if ($ticks < 0) {
+            # This call will effectively never return
             Mojo::IOLoop->singleton->start;
```
okurz, line 471: maybe instead of creating a subprocess and then calling `Mojo::IOLoop::start` we could use the functionality from `Mojo::IOLoop` directly and do something like `Mojo::IOLoop::subprocess`

t/lib/OpenQA/Test/Utils.pm
```diff
@@ -467,13 +465,14 @@ sub unstable_worker {
...
         }
         else {
             Mojo::IOLoop->singleton->one_tick for (0 .. $ticks);
         }
-        Devel::Cover::report() if Devel::Cover->can('report');
         if ($sleep) {    # uncoverable statement
             1 while sleep $sleep;    # uncoverable statement
         }    # uncoverable statement
```
okurz, line 478: likely this can be even simplified to

```suggestion
        1 while sleep $sleep;    # uncoverable statement
```

or

```suggestion
        sleep $sleep if $sleep;    # uncoverable statement
```

it seems one can even just do

```suggestion
        sleep $sleep;    # uncoverable statement
```

okurz (changes requested):

t/lib/OpenQA/Test/Utils.pm
kalikiana: This seems to break the tests.

t/lib/OpenQA/Test/Utils.pm
kalikiana: I want for a ridiculous amount of ticks, which means we can use the same code path

PR #3658, opened 2021-01-05
Clear any uncommitted changes that would prevent a rebase

kalikiana:
See: https://progress.opensuse.org/issues/81180

### DIFF
--- a/script/fetchneedles
+++ b/script/fetchneedles
@@ -15,6 +15,7 @@
 : "${needles_branch:="master"}"
 
 : "${updateall:="0"}"
+: "${force:="0"}"
 
 dir="/var/lib/openqa/share/tests"
 if [ -w / ]; then
@@ -37,6 +38,11 @@ if [ -w / ]; then
     exit 1
 fi
 
+fail() {
+    echo "$*" >&2
+    exit 1
+}
+
 needlesdir() {
     # new layout since
     # https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/920
@@ -51,7 +57,9 @@ git_update() {
     branch="${1:-"master"}"
     git gc --auto --quiet
     git fetch -q origin
-    git rebase -q origin/"$branch"
+    # Clear any uncommitted changes that would prevent a rebase
+    [ "$force" = 1 ] && git reset --hard HEAD
+    git rebase -q origin/"$branch" || fail 'Use force=1 to discard uncommitted changes before rebasing'
 }
 
 # For needles repos because of needle saving we might end up in conflict, i.e.


okurz (approved):

Martchus (commented):
Why not just using `get reset --hard HEAD` unconditionally? If there's no diff it should be a no-op and likely not more expensive than `git diff-index`.

script/fetchneedles
```diff
@@ -51,6 +51,8 @@ git_update() {
     branch="${1:-"master"}"
     git gc --auto --quiet
     git fetch -q origin
+    # Clear any uncommitted changes that would prevent a rebase
+    git reset --hard HEAD
```
okurz, line 55: I would feel safer if we introduce this together with a parameter `--force` or environment variables "force". Can you do that?

kalikiana: Sounds good. I added a check for `force=1` which defaults to `0` and if the rebase fails the option is suggested as a way to handle it automatically.

script/fetchneedles
```diff
@@ -51,7 +52,9 @@ git_update() {
     branch="${1:-"master"}"
     git gc --auto --quiet
     git fetch -q origin
-    git rebase -q origin/"$branch"
+    # Clear any uncommitted changes that would prevent a rebase
+    [ "$force" = 1 ] && git reset --hard HEAD
+    git rebase -q origin/"$branch" || echo "Use force=1 to discard uncommitted changes before rebasing"
```
okurz, line 57: Because of your echo call the script would not fail anymore in case of error but continue and likely fail later. Better also fail the script

okurz (changes requested):

script/fetchneedles
kalikiana: Silly me, forgot to add an `&& exit 1` there. Thank you for catching it!

I tested this again, with a dirty staging area and also wrong branch names for good measure (not sure if that's used in the wild, but the script supports it and should fail in that case).

script/fetchneedles
```diff
@@ -51,7 +52,9 @@ git_update() {
     branch="${1:-"master"}"
     git gc --auto --quiet
     git fetch -q origin
-    git rebase -q origin/"$branch"
+    # Clear any uncommitted changes that would prevent a rebase
+    [ "$force" = 1 ] && git reset --hard HEAD
+    git rebase -q origin/"$branch" || echo "Use force=1 to discard uncommitted changes before rebasing" && exit 1
```
okurz, line 57: this won't work either, try

```bash
echo "pass" || echo "fail" && echo "exit 1"
```

locally to see what I mean. You will see

```
pass
exit 1
```

See https://github.com/os-autoinst/openQA/blob/master/script/openqa-clone-custom-git-refspec#L36 for a solution.

Define

```bash
fail() {
    echo "$*" >&2
    exit 1
}
```

then call

```suggestion
    git rebase -q origin/"$branch" || fail "Use force=1 to discard uncommitted changes before rebasing"
```

okurz (changes requested):

script/fetchneedles
kalikiana: I'm confused. I tested it and it worked perfectly. I guess it might depend on the shell or whatever 😒

Will use the *fail* function. Thank you for the idea.

Martchus (approved):

okurz (approved):

PR #3665, opened 2021-01-11
RFC: never query isotovideo interface version

kalikiana:
- Querying the version has the same overhead as executing a job
- We don't use the version except to display it
- isotovideo already logs the version

I just prepared this PR since I had tested out briefly to see what the difference was. It's possible that I missed something. But I figure I might as well share it. I suspect I'll learn something from it either way.

### DIFF
--- a/t/24-worker-engine.t
+++ b/t/24-worker-engine.t
@@ -18,7 +18,7 @@ use Test::Most;
 use FindBin;
 use lib "$FindBin::Bin/lib", "$FindBin::Bin/../external/os-autoinst-common/lib";
 use OpenQA::Test::TimeLimit '10';
-use OpenQA::Constants 'WORKER_EC_ASSET_FAILURE';
+use OpenQA::Constants qw(WEBSOCKET_API_VERSION WORKER_EC_ASSET_FAILURE);
 use Test::Fatal;
 use Test::Warnings ':report_warnings';
 use OpenQA::Worker;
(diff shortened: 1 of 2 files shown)


t/24-worker-engine.t
```diff
@@ -58,7 +58,7 @@ subtest 'isotovideo version' => sub {
 
     my $isotovideo = path($FindBin::Bin)->child('fake-isotovideo.pl');
     my $worker2 = OpenQA::Worker->new({apikey => 'foo', apisecret => 'bar', instance => 1, isotovideo => $isotovideo});
-    is($worker2->isotovideo_interface_version, 15, 'isotovideo version set from os-autoinst');
+    is($worker2->isotovideo_interface_version, WEBSOCKET_API_VERSION, 'isotovideo version set from os-autoinst');
```
Martchus, line 61: We should consequently remove `isotovideo_interface_version` and the entire test as well.

coolo:
What's wrong with displaying the proper version of os-autoinst installed? That's quite useful for the admin to know before merging new tests.

Martchus (commented):
At this point the worker doesn't talk to isotovideo (or its command server) anymore via web sockets. So from the worker's point of view this versioning is indeed useless.

However, it passes the version to the web UI which then makes it available to the user. So one can easily see within the web UI which interface version is deployed on which worker. That's maybe something we don't want to remove.

---

> Querying the version has the same overhead as executing a job

You mean as a no-op job. And yes, it takes indeed a short moment (0.349 s here).

Likely we should also remove all places where we print that version in the worker log and show it within the web UI because otherwise it is confusing to just see the expected version where usually the actual version was displayed. But as I mentioned, I don't know whether we really want to remove that.

t/24-worker-engine.t
Amrysliu: Agree. If we don't need os-autoinst version, why we return `WEBSOCKET_API_VERSION`? is it used somewhere?

Martchus:
I would take the comments as a "no".

kalikiana:
> I would take the comments as a "no".

Well, this was obviously a proof of concept 😁 I could've filed a ticket but w/o a benchmark it'll probably end up as future anyway. So I simpliy don't know if it's worth it.

PR #3666, opened 2021-01-11
Add service files to restart worker after config changes without interrupting running jobs

Martchus:
* As long as e.g. `openqa-soft-restart-worker@1.path` is running
  the worker started via the corresponding service unit
  `openqa-worker-auto-restart@1.service` is restarted as soon as it
  becomes idle (and immediately if it is already idling) when
  `/etc/openqa/workers.ini` is edited.
* See https://progress.opensuse.org/issues/80910
  and https://progress.opensuse.org/issues/80908#note-8

---

* ~~We could actually also watch for a packaged file so it would also apply
  to package updates.~~ Maybe that's better done on RPM level as suggested in https://progress.opensuse.org/issues/80908#note-8 (point 2.)?
* Then we could also get rid of `OPENQA_WORKER_TERMINATE_AFTER_JOBS_DONE=1`
  in the auto-restart service file.

### DIFF
--- a/dist/rpm/openQA.spec
+++ b/dist/rpm/openQA.spec
@@ -566,6 +566,8 @@ fi
 %{_unitdir}/openqa-worker-cacheservice.service
 %{_unitdir}/openqa-worker-no-cleanup@.service
 %{_unitdir}/openqa-worker-auto-restart@.service
+%{_unitdir}/openqa-reload-worker-auto-restart@.service
+%{_unitdir}/openqa-reload-worker-auto-restart@.path
 %{_unitdir}/openqa-slirpvde.service
 %{_unitdir}/openqa-vde_switch.service
 %{_datadir}/openqa/script/openqa-slirpvde
--- a/docs/Installing.asciidoc
+++ b/docs/Installing.asciidoc
@@ -588,6 +588,9 @@ package provides further systemd units:
 ** Stops/restarts *all* worker units when stopped/restarted.
 ** Is restarted automatically when the `openQA-worker` package is updated
    (unless `DISABLE_RESTART_ON_UPDATE="yes"` is set in `/etc/sysconfig/services`).
+* `openqa-reload-worker-auto-restart@.path`: allows to restart the worker service
+  automatically on configuration changes without interrupting jobs (see next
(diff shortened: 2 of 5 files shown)


Martchus:
This seems to work as expected.

docs/Installing.asciidoc
```diff
@@ -604,6 +604,10 @@ ongoing testing. Example:
 systemctl kill --signal SIGHUP 'openqa-worker-auto-restart@*.service'
 --------------------------------------------------------------------------------
 
+There is also the systemd unit `openqa-soft-restart-worker@.path` which invokes
+the command above whenever the worker configuration under `/etc/openqa/workers.ini`
+changes.
+
```
okurz, line 610: We still have that list of worker service files above this section, i.e. above line 591. Should both openqa-soft-restart-worker@.{path,service} also be mentioned there? Also I think it would be good if you could make it clear to the person setting up an installation if that needs any action or is done automatically.

dist/rpm/openQA.spec
```diff
@@ -566,6 +566,8 @@ fi
 %{_unitdir}/openqa-worker-cacheservice.service
 %{_unitdir}/openqa-worker-no-cleanup@.service
 %{_unitdir}/openqa-worker-auto-restart@.service
+%{_unitdir}/openqa-soft-restart-worker@.service
```
okurz, line 569: not sure how well defined a "soft-restart" is. Maybe just "reload" to explain that the complete worker service is sustained but with new config?

okurz (changes requested):

dist/rpm/openQA.spec
Martchus: Makes sense. I renamed it to `openqa-reload-worker-auto-restart@.path`. I added the `auto-restart` part to make it clear that this unit only affects the `…-worker-auto-restart@.service` but also added a note within the documentation.

docs/Installing.asciidoc
Martchus: I've mentioned it in the list above and that the unit is not enabled by default.

okurz (approved):

PR #3667, opened 2021-01-12
Document further systemd units for the worker

Martchus:


kalikiana (approved):

docs/Installing.asciidoc
```diff
@@ -568,6 +568,24 @@ Once you got workers running they should show up in the admin section of openQA
...
+package provides further systemd units:
+
+* `openqa-worker-no-cleanup@.service`: see
+  <<WritingTests.asciidoc#snapshots-for-each-module,enabling snapshots>>
+* `openqa-worker-auto-restart@.service`: worker that restarts automatically
+  after processing assigned jobs
+* `openqa-worker-cacheservice`/`openqa-worker-cacheservice-minion`: services
+  for <<Installing.asciidoc#asset-caching,the asset cache>>
+* `openqa-worker.target`
+** starts `openqa-worker@1.service` (but no other worker units) when started
```
okurz, line 582: why is that? https://github.com/os-autoinst/openQA/blob/8784296ab3689882f3e2d0f4b70b7e3875a61bac/systemd/openqa-worker%40.service#L9 says just `PartOf=openqa-worker.target` and I have not found any other place where the first instance would be handled in a special way.

Martchus: When testing it yesterday, it *always* started `openqa-worker@1.service` and never e.g. `openqa-worker@2.service`. It did not matter what instances were already running or had been enabled or were running when stopping the target the last time. However, when testing it again today, it always starts `openqa-worker@1.service` and `openqa-worker@2.service`. Before I was wondering as well where the "1" comes from. Now I'm wondering where this whole list of services comes from and how it can be modified.

Note that `PartOf=openqa-worker.target` has apparently no effect when it comes to deciding what services are started when starting the target. For instance, `openqa-worker-cacheservice.service` has it as well but it is not started when the target is started. (But it is stopped when the target is stopped as well as *all* instances.)

Martchus: I've asked on systemd's IRC channel and someone helped me to figure it out.

It depends on the symlinks within the `openqa-worker.target.wants` directory which is searched for under various locations (see `man systemd.unit`). In this case it is `/run/systemd/generator/openqa-worker.target.wants`. This directory is generated by https://github.com/os-autoinst/openQA/blob/master/systemd/systemd-openqa-generator.

In conclusion the "@"-parameters services are started for correspond to the available pool directories at the last time the generators were run and one can always use `systemctl daemon-reload` to refresh it.

okurz:
please make sure the "unreliable coverage report" is handled in tickets so that we won't be annoyed by that

Martchus:
You had already a ticket for that but I guess it is already closed at this point so I'll create a new one.

okurz:
Martchus created https://github.com/os-autoinst/openQA/pull/3669

okurz (approved):

PR #3624, opened 2020-12-07
Clone incompletes caused by cache service errors automatically

Martchus:
* It is possible to configure the regex being used
* It is possible to disable this kind of automatic cloning
* See https://progress.opensuse.org/issues/80202#note-15

---

* ~~Based on https://github.com/os-autoinst/openQA/pull/3618 which could be reviewed individually first~~ rebased on master after #3618 was merged
* See last commit for only the additional change

PR #3672, opened 2021-01-14
Automatically rerun incompleted jobs because of no space left

Amrysliu:
Relate: https://progress.opensuse.org/issues/80334

### DIFF
--- a/lib/OpenQA/Worker/Job.pm
+++ b/lib/OpenQA/Worker/Job.pm
@@ -35,6 +35,7 @@ use Mojo::JSON 'decode_json';
 use Mojo::File 'path';
 use Try::Tiny;
 use Scalar::Util 'looks_like_number';
+use File::Map 'map_file';
 
 # define attributes for public properties
 has 'worker';
@@ -568,11 +569,19 @@ sub _format_reason {
         }
     }
     catch {
+        # read autoinst-log.txt to check the reason, see poo#80334
+        my $msg = '';
+        eval {
+            map_file my $log_content, path($self->worker->pool_directory, 'autoinst-log.txt'), '<';
+            $msg = ': No space left on device' if ($log_content =~ /No space left on device/);
+        };
+        log_warning($@) if $@;
+
         if ($reason eq WORKER_SR_DONE) {
-            $reason = "$reason: terminated with corrupted state file";
+            $reason = "$reason: terminated with corrupted state file$msg";
         }
         else {
-            $reason = "$reason: terminated prematurely with corrupted state file, see log output for details";
+            $reason = "terminated prematurely: Encountered corrupted state file$msg, see log output for details";
         }
         log_warning("Found $state_file but failed to parse the JSON: $_");
     };
--- a/t/24-worker-jobs.t
+++ b/t/24-worker-jobs.t
@@ -443,7 +443,7 @@ subtest 'Job aborted, broken state file' => sub {
     combined_like { wait_until_job_status_ok($job, 'stopped') } qr/failed to parse.*JSON/, 'warning about corrupt JSON';
     is(
         @{$client->sent_messages}[-1]->{reason},
-        'died: terminated prematurely with corrupted state file, see log output for details',
+        'terminated prematurely: Encountered corrupted state file, see log output for details',
         'reason propagated'
     ) or diag explain $client->sent_messages;
     combined_like {
@@ -1236,6 +1236,29 @@ subtest 'known images and files populated from status update' => sub {
     $job->_upload_results_step_0_prepare();
     is_deeply($job->known_images, \@fake_known_images, 'known images populated from status update');
     is_deeply($job->known_files,  \@fake_known_files,  'known files populated from status update');
+};
+
+subtest 'write no space left to job reason by parsing autoinst-log' => sub {
+    my $job = OpenQA::Worker::Job->new($worker, $client, {id => 12, URL => $engine_url});
+    $engine_mock->redefine(
+        engine_workit => sub {
+            $pool_directory->child('base_state.json')->spurt(qq(foo boo));
+            $pool_directory->child('autoinst-log.txt')
+              ->spurt(
+'[debug] Unable to serialize fatal error: Can\'t write to file "base_state.json": No space left on device at /usr/lib/os-autoinst/bmwqemu.pm line 86.'
+              );
+            $job->stop(WORKER_SR_DIED);
+            return {error => 'worker interrupted'};
+        });
+    $job->accept;
+    wait_until_job_status_ok($job, 'accepted');
+    $job->start;
+    combined_like { wait_until_job_status_ok($job, 'stopped') } qr/failed to parse.*JSON/, 'warning about corrupt JSON';
+    is(
+        @{$client->sent_messages}[-1]->{reason},
+        'terminated prematurely: Encountered corrupted state file: No space left on device, see log output for details',
+        'The job incomplete reason includes "No space left on device"'
+    ) or diag explain $client->sent_messages;
 };
 
 subtest 'Cache setup error handling' => sub {
--- a/.circleci/ci-packages.txt
+++ b/.circleci/ci-packages.txt
@@ -244,6 +244,7 @@ perl-XML-Simple-2.24
 perl-YAML-1.24
 perl-YAML-LibYAML-0.82
 perl-YAML-PP-0.026
+perl-File-Map
 psmisc-23.0
 python3-appdirs-1.4.3
 python3-EditorConfig-0.12.2
--- a/container/travis_test/Dockerfile
+++ b/container/travis_test/Dockerfile
@@ -129,6 +129,7 @@ RUN zypper in -y -C \
        'perl(Time::ParseDate)' \
        'perl(XSLoader) >= 0.24' \
        'perl(XML::SemanticDiff)' \
+       'perl(File::Map)' \
        perl-Archive-Extract \
        perl-Test-Simple \
        'perl(aliased)' \
--- a/cpanfile
+++ b/cpanfile
@@ -35,6 +35,7 @@ requires 'Fcntl';
 requires 'File::Basename';
 requires 'File::Copy';
 requires 'File::Copy::Recursive';
+requires 'File::Map';
 requires 'File::Path';
 requires 'File::Spec';
 requires 'FindBin';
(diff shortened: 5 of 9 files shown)


lib/OpenQA/Schema/Result/Jobs.pm
```diff
@@ -1994,6 +1994,12 @@ sub done {
         my $auto_clone_regex = OpenQA::App->singleton->config->{global}->{auto_clone_regex};
         $restart = 1 if $result eq INCOMPLETE && $auto_clone_regex && $reason =~ $auto_clone_regex;
 
+        # for poo#80334
+        my $autoinst_log_file = path($self->result_dir, "autoinst-log.txt");
+        my $command           = "grep \"No space left on device\" $autoinst_log_file > /dev/null";
+        $restart = 1
+          if ($restart == 0 && $reason =~ /terminated prematurely with corrupted state file/ && system($command) == 0);
```
Martchus, line 2001: It would be better to grep within the log file directly from the Perl script instead of starting a shell and grep. I would also avoid any extra checks unless the result is incomplete.

Martchus (changes requested):

t/api/04-jobs.t
```diff
@@ -1307,6 +1307,27 @@ subtest 'show parent group name and id when getting job details' => sub {
     is $job->{parent_group_id}, $parent_group_id, 'parent group id was shown correctly';
 };
 
+subtest 're-trigger jobs when it incomplete becase no space left' => sub {
```
okurz, line 1310: s/becase/because/

okurz (changes requested):
In general I would favor a more general solution. Grepping the log files for certain terms is something that we already support and use with "job done" hooks. If we want to include that into openQA code directly then we should do it in a more generic style that works with multiple, different "known issues" and would need to configure the search terms at best outside the actual source code.

@Martchus you recently introduce the possibility to better automatically retrigger jobs for known causes. Can we find some middle ground of errors that we detect as "nothing that openQA can solve" but worth to retrigger without needing to parse the logfiles?

I hope you understand what I mean. If I can not make myself clear I am happy to discuss the topic in a video call the next time we come together.

Martchus:
> If the reason "No space" is not visible in the reason field then this simply means that we need to ensure this is included in the reason. This work is done by the worker so scales always better than anything the webUI needs to do.

This is a very good point so I'd also say you should implement this on the worker-side.

lib/OpenQA/Schema/Result/Jobs.pm
kalikiana:
```perl
open my $fh, "<", $autoinst_log_file or die "Failed to open $autoinst_log_file";
my @lines = sort grep /No space left on device/, <$fh>
```

Something like this 🤔

Amrysliu: > It would be better to grep within the log file directly from the Perl script instead of starting a shell and grep. 

Could you give me more details?  read content from the file and do grep? just like @kalikiana wrote. Thanks.

Amrysliu:
> In general I would favor a more general solution. Grepping the log files for certain terms is something that we already support and use with "job done" hooks. If we want to include that into openQA code directly then we should do it in a more generic style that works with multiple, different "known issues" and would need to configure the search terms at best outside the actual source code.
> 
> @Martchus introduced the possibility to better automatically retrigger jobs for known causes based on the reason in #3624 . Can we find some middle ground of errors that we detect as "nothing that openQA can solve" but worth to retrigger without needing to parse the logfiles but just relying on the logfile parsing? If the reason "No space" is not visible in the reason field then this simply means that we need to ensure this is included in the reason. This work is done by the worker so scales always better than anything the webUI needs to do.
> 
> I hope you understand what I mean. If I can not make myself clear I am happy to discuss the topic in a video call the next time we come together.

Thanks for your comments. do you mean that: we should put the `no space left` into the reason, and configure `auto_clone_regex` in openqa.ini to re-trigger the job?  

>Can we find some middle ground of errors that we detect as "nothing that openQA can solve" but worth to retrigger without needing to parse the logfiles but just relying on the logfile parsing?

Could we parse the logfiles on worker side?

Martchus:
> Thanks for your comments. do you mean that: we should put the no space left into the reason, and configure auto_clone_regex in openqa.ini to re-trigger the job?

yes

> Could we parse the logfiles on worker side?

I think that's the plan. Of course the sentence "without needing to parse the logfiles but just relying on the logfile parsing" is self-contradicting and makes no sense (so just ignore it).

lib/OpenQA/Worker/Job.pm
```diff
@@ -568,11 +569,22 @@ sub _format_reason {
...
+        # read autoinst-log.txt to check the reason, see poo#80334
+        my $msg               = '';
+        my $autoinst_log_file = path($self->worker->pool_directory, 'autoinst-log.txt');
+        if (!-e $autoinst_log_file) {
+            log_warning("File $autoinst_log_file does not exist");
+        }
+        else {
+            map_file my $log_content, $autoinst_log_file, '<';
+            $msg = ': No space left on device' if ($log_content =~ /No space left on device/);
+        }
```
Martchus, line 581: `map_file` will throw an exception if the file can not be opened. I would rather catch this exception instead of testing this condition before. (There might be other exceptions as well and the file might just vanish between `-e …` and `map_file …`.

```suggestion
        eval {
            map_file my $log_content, path($self->worker->pool_directory, 'autoinst-log.txt'), '<';
            $msg = ': No space left on device' if ($log_content =~ /No space left on device/);
        }
        log_warning $@ if $@;
```

t/24-worker-jobs.t
```diff
@@ -1238,6 +1238,29 @@ subtest 'known images and files populated from status update' => sub {
     is_deeply($job->known_files,  \@fake_known_files,  'known files populated from status update');
 };
 
+subtest 'write no space left to job failed reason by parsing autoinst-log' => sub {
```
Martchus, line 1241:
```suggestion
subtest 'write no space left to job reason by parsing autoinst-log' => sub {
```

(Just because this is not really tied to the job result "failed".)

Martchus (changes requested):

okurz:
> Of course the sentence "without needing to parse the logfiles but just relying on the logfile parsing" is self-contradicting and makes no sense (so just ignore it).

Yes, sorry about that. What I mean is: I think we should not need to parse the logfile. We should have everything in the reason field. If the reason field was not updated we should still have a generic "Unexpected premature termination encountered, no reason could be found, retriggering automatically"-reason and automatic retrigger, no log parsing.

Amrysliu:
> > Of course the sentence "without needing to parse the logfiles but just relying on the logfile parsing" is self-contradicting and makes no sense (so just ignore it).
> 
> Yes, sorry about that. What I mean is: I think we should not need to parse the logfile. We should have everything in the reason field. If the reason field was not updated we should still have a generic "Unexpected premature termination encountered, no reason could be found, retriggering automatically"-reason and automatic retrigger, no log parsing.
> 
> Again, if a host runs out of memory or storage that is not for openQA to handle, the admin+monitoring need to take care about that.

So do you mean we only need to add a prefix for the reason, such as `Unexpected died: terminated prematurely with corrupted state file`, without parsing log file?
@Martchus WDYT?

Amrysliu:
@Martchus Could you give me some details, what I need to do to let the CI check pass? I add the `File::Map` in dependencies.yaml, is there somewhere I need to modify it? Thanks

Martchus:
I still don't understand @okurz comment.

> Yes, sorry about that. What I mean is: I think we should not need to parse the logfile.

Where is the information to come from then?

> …, retriggering automatically

I would not put "retriggering automatically" into the reason (at least not on the worker-side when the retriggering is actually done by the web UI).

okurz:
> I still don't understand @okurz comment.
> 
> > Yes, sorry about that. What I mean is: I think we should not need to parse the logfile.
> 
> Where is the information to come from then?

Again, if a host runs out of memory or storage that is not for openQA to handle, the admin+monitoring need to take care about that. So if we fail to write a json file which includes a useful reason then the worker should just describe that clearly in the reason field and abort. IMHO the worker should not try to be too intelligent, parsing the logfile, which might fail for the same reason "out of space" because maybe some temporary file or something could not be created.

> > …, retriggering automatically
> 
> I would not put "retriggering automatically" into the reason (at least not on the worker-side when the retriggering is actually done by the web UI).

you can still come up with some text. I just wanted to provide an example.

okurz:
You know what: By now I think for the time being we should rely on auto-review and not do anything more from the side of the worker

Martchus:
Of course parsing the log file is hacky. I personally also don't think it is worth it but somehow seemed very wanted in previous discussions to restart in this particular case. Of course we can alternatively just say that we want to restart all incompletes which can not write the state file. However, that's a different set of incompletes.

Amrysliu:
> You know what: By now I think for the time being we should rely on auto-review and not do anything more from the side of the worker

so, as a result, I could close this pr? we do not need to change any code because auto-review already does well.

lib/OpenQA/Worker/Job.pm
```diff
@@ -568,11 +569,19 @@ sub _format_reason {
...
+        };
+        log_warning($@) if $@;
+
         if ($reason eq WORKER_SR_DONE) {
-            $reason = "$reason: terminated with corrupted state file";
+            $reason = "$reason: terminated with corrupted state file$msg";
         }
         else {
-            $reason = "$reason: terminated prematurely with corrupted state file, see log output for details";
+            $reason = "$reason: terminated prematurely with corrupted state file$msg, see log output for details";
```
okurz, line 584: As discussed we can change the reason to be more specific. I suggest:

```suggestion
            $reason = "terminated prematurely: Encountered corrupted state file$msg, see log output for details";
```

okurz (changes requested):
with the new prefix suggestion "terminated prematurely" please add that in etc/openqa/openqa.ini in "auto_clone_regex". Please keep the log parsing. It's ok. Sorry about my previous confusing comment stating otherwise.

okurz:
please see http://open.qa/docs/#_perl_and_other_packages about adding dependencies. And please let us know or correct if it's wrong.

okurz (changes requested):
please also extend `auto_clone_regex            => '^cache failure: '` in lib/OpenQA/Setup.pm

Amrysliu:
> please see http://open.qa/docs/#_perl_and_other_packages about adding dependencies. And please let us know or correct if it's wrong.

The document doesn't say how to let the ci check pass, expect that, the rest are fine. I modified the ci-packages.txt, not sure if it's the right way. Seems this file is written by other processes, not handly, right?

Martchus:
> Seems this file is written by other processes, not handly, right?

Yes. But it is nevertheless correct to simply add the dependency to the file in the way you did to make a new dependency immediately available within the CI.

Martchus (approved):

okurz (approved):

PR #3676, opened 2021-01-15
Workaround getaddrinfo() being stuck in error state for 127.0.0.1

Martchus:
* See https://progress.opensuse.org/issues/78390#note-38
* If this actually helps we could try the adjustments to underlying
  Perl libraries mentioned in https://progress.opensuse.org/issues/78390#note-41

### DIFF
--- a/lib/OpenQA/Worker.pm
+++ b/lib/OpenQA/Worker.pm
@@ -16,6 +16,33 @@
 package OpenQA::Worker;
 use Mojo::Base -base;
 
+BEGIN {
+    use Socket;
+    use Scalar::Util;
+    use OpenQA::Log qw(log_debug);
+    use Mojo::Util qw(monkey_patch);
+
+    # workaround getaddrinfo() being stuck in error state "Address family for hostname not supported"
+    # for local connections (see https://progress.opensuse.org/issues/78390#note-38)
+    my $original = \&Socket::getaddrinfo;
+    monkey_patch 'Socket', getaddrinfo => sub {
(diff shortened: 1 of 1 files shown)


kraih:
It's pretty ugly, but if this works out in production i'll make an upstream change in Mojolicious for a clean solution (new `Mojo::UserAgent` API to pass additional socket parameters to `IO::Socket::IP`).

okurz (approved):
I am happy to see that you came up with what looks like a workaround. Of course I also can agree with @kraih that it's not that pretty but I guess that's only natural and expected as you work around issues from upstream. So if we can really see this helping in production I would also be very happy to see that being transformed into a proper upstream fix.

kalikiana:
@Martchus I appreciate your effort in finding the correct solution... but if this is an alternative to #3673 is it worth the hassle? Or perhaps it makes sense to go the systemd route until we have an upstream fix - but maybe there's a drawback that I'm not seeing. To me this just seems potentially very hard to maintain code

okurz:
I mentioned in #3673 and in our voice chat last weekly meekly meeting that the systemd dependency is likely introducing other problems and not a good solution

kraih (approved):

Martchus:
> Or perhaps it makes sense to go the systemd route until we have an upstream fix

Definitely not. The upstream change might be some effort. Only if this change actually does the trick we know that it is actually worth it. We also don't know whether the systemd route will actually work as expected. And besides: https://github.com/os-autoinst/openQA/pull/3673#issuecomment-762299562

PR #3677, opened 2021-01-18
Adjust Bugzilla links for SLE 15 SP3 products to add new "PUBLIC" prefix

Martchus:
According to the mail "Opening our Bugzilla - SLE 15 SP3 QA bugs" sent to
the "qa-team" mailing list the Bugzilla product name for certain for SLE 15
SP3 product changes. This commit adjusts our error reporting template
accordingly. It takes only the products mentioned within that mail into
account but the approach allows to add more public products easily later.

### DIFF
--- a/templates/webapi/branding/openSUSE/external_reporting.html.ep
+++ b/templates/webapi/branding/openSUSE/external_reporting.html.ep
@@ -36,6 +36,11 @@
     Leanos => 'Server',
     Installer => 'Server',
 ); %>
+<% my %public_sle_products = (
+    Server => 'Server',
+    Desktop => 'Desktop',
+    'High Availability Extension' => 'High Availability', # the public version leaves out the "Extension" suffix
+); %>
 % my $distri = $distri_to_prod{$job->DISTRI} // 'DISTRI NOT FOUND: Adjust templates/openSUSE/external_reporting.html.ep';
 % my $product = '';
 % if ($job->DISTRI eq 'sle') {
@@ -49,7 +54,15 @@
 %             $version = "12 SP1\x{00A0}(SLED 12 SP1)";
 %         }
 %         # fall back to 'Server' as the most common flavor
-%         $product = join(' ', $flavor_to_prod_sle{$subproduct} // 'Server', $version);
+%         my $sle_product        = $flavor_to_prod_sle{$subproduct} // 'Server';
+%         my $public_sle_product = $public_sle_products{$sle_product};
+%         if ($public_sle_product && ($version =~ qr/(\d+) SP(\d+)/)) {
+%             if ($1 == 15 && $2 >= 3) {
+%                 $distri      = "PUBLIC $distri" ;
+%                 $sle_product = $public_sle_product;
+%             }
+%         }
+%         $product = "$sle_product $version";
 %     }
 % }
 % elsif ($job->DISTRI eq 'opensuse' || $job->DISTRI eq 'microos') {


Martchus:
By the way, I've tested this manually using an arbitrary job on my local instance and adjusting the relevant job settings manually via SQL. It didn't look like we have any automatic tests for this logic and maybe it is also not worth adding one.

templates/webapi/branding/openSUSE/external_reporting.html.ep
```diff
@@ -49,7 +54,15 @@
...
 %         # fall back to 'Server' as the most common flavor
-%         $product = join(' ', $flavor_to_prod_sle{$subproduct} // 'Server', $version);
+%         my $sle_product        = $flavor_to_prod_sle{$subproduct} // 'Server';
+%         my $public_sle_product = $public_sle_products{$sle_product};
+%         if ($public_sle_product && ($version =~ qr/(\d+) SP(\d+)/)) {
+%             if ($1 == 15 && $2 >= 3) {
+%                 $distri      = "PUBLIC $distri" ;
+%                 $sle_product = $public_sle_product;
+%             }
+%         }
```
okurz, line 64: I see you have foreseen all possible problems already :) I hope the maintainers of products on bugzilla realize how much mess they cause with their inconsistent naming and changes :(

kalikiana:
> By the way, I've tested this manually using an arbitrary job on my local instance and adjusting the relevant job settings manually via SQL. It didn't look like we have any automatic tests for this logic and maybe it is also not worth adding one.

Yeah. I'd still _like_ to see coverage for it eventually but there's no significant breakage if you get it wrong.

okurz (approved):

templates/webapi/branding/openSUSE/external_reporting.html.ep
Martchus: Likely they'll change patterns again for SP4 but then we can still adjust the `>=`.

templates/webapi/branding/openSUSE/external_reporting.html.ep
```diff
@@ -36,6 +36,11 @@
     Leanos => 'Server',
     Installer => 'Server',
 ); %>
+<% my %public_sle_products = (
+    Server => 'Server',
+    Desktop => 'Desktop',
+    'High Availability Extension' => 'High Availability', # the public version leaves out the "Extension" suffix
+); %>
```
kraih, line 43: You could also just prefix each line with a `% ` instead of wrapping them in `<% ... %>`. Works the same.

templates/webapi/branding/openSUSE/external_reporting.html.ep
kraih: Maybe worth turning into a helper? Logic definitely seems complex enough.

Martchus: Maybe later (one can open a ticket for that). Since this is not a usual template but only part of an optional and distro-specific branding we might not want to mix it into our usual helpers.

Martchus:
The TW build is blocked but it shouldn't be relevant anyways. I'm merging it now so we don't forget it and it will be deployed on Wednesday. (The mail mentions it should be done until Friday 22th Jan.)

templates/webapi/branding/openSUSE/external_reporting.html.ep
okurz: I agree, this is very site-specific hence in branding. Would be interesting if there could be "site-specific plugin code" elsewhere and then integrated

ldevulder:
According to https://confluence.suse.com/pages/viewpage.action?spaceKey=SUSELinuxEnterpriseServer15SP3&title=Opening+our+Bugzilla+-+FAQ only SLES, SLED and SLEHA should have public bugs opened by default, but since this PR SLES4SAP or SLE-HPC bugs are also opened as public by default... which is not expected AFAIK.

Martchus:
Ok, but I'm not sure whether I understand this right. It would be helpful if you could talk a bit more in "openQA job settings terms". That's because I don't know to which job settings e.g. "SLES4SAP" corresponds to. Of course you can also come up with a PR on your own.

ldevulder:
> Ok, but I'm not sure whether I understand this right. It would be helpful if you could talk a bit more in "openQA job settings terms". That's because I don't know to which job settings e.g. "SLES4SAP" corresponds to. Of course you can also come up with a PR on your own.

I don't know the "openQA job settings terms" otherwise I would have use it :). The issue is that if we try to open a bug for tests in for example these Job Groups: https://openqa.suse.de/group_overview/146 and https://openqa.suse.de/group_overview/130 the bugs are opened in "PUBLIC" but it should not be the case.

@schlad FYI, I think that we can be concerned with HPC ;-)

Martchus:
I've just picked a random job from each of the mentioned groups (https://openqa.suse.de/tests/5405017, https://openqa.suse.de/tests/5400992). In these jobs DISTRI, VERSION and FLAVOR are sle, 15-SP3 and Online/Full. These job settings don't look very special. Maybe a special FLAVOR could be used to distinguish these jobs? We could also opt-out of the "PUBLIC " prefix if the job settings SLES4SAP_MODE and SCC_REGCODE_HPC are present or simply if the job is part of the mentioned groups.

PR #3679, opened 2021-01-19
client: Log what config file is being read

kalikiana:
When trying to debug why e.g. `/etc/openqa/client.conf` is not working it might be because `~/.config/openqa/client.conf` takes precedence. That could be seen more easily if we print the filename verbosely.

See: [poo#820670](https://progress.opensuse.org/issues/820670)

### DIFF
--- a/lib/OpenQA/UserAgent.pm
+++ b/lib/OpenQA/UserAgent.pm
@@ -16,6 +16,8 @@
 package OpenQA::UserAgent;
 use Mojo::Base 'Mojo::UserAgent';
 
+use OpenQA::Log 'log_info';
+
 use Mojo::Util 'hmac_sha1_sum';
 use Config::IniFiles;
 use Scalar::Util ();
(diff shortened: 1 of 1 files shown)


perlpunk (approved):

Martchus (approved):

lib/OpenQA/UserAgent.pm
```diff
@@ -38,6 +40,7 @@ sub new {
         for my $path (@cfgpaths) {
             my $file = $path . '/client.conf';
             next unless $file && -r $file;
+            log_info("Reading config from $file");
```
okurz, line 43: not sure if this is a good idea to print output by default when AFAIK the client does not offer an interface to increase/decrease the log level. I think "log_debug" + command line parameters to enable debug messages would be a better approach

perlpunk: Yeah, this actually resulted in https://progress.opensuse.org/issues/88053
So we have to find an alternative

Martchus: What about just writing it to stderr? Users would usually only pass stdout to their JSON parsers.

PR #3680, opened 2021-01-19
clone-custom-git-refspec: Call clone-job from same install

kalikiana:
I spent a little while being confused why a change wasn't working like I expected. Turns out clone-custom-refspec relies on the system path.

I would expect all commands calling into each other to be from the same install. Which is predictable. And avoids potential incompatibilities.

### DIFF
--- a/script/openqa-clone-custom-git-refspec
+++ b/script/openqa-clone-custom-git-refspec
@@ -135,7 +135,9 @@ Please try 'curl $json_url' or select another job, e.g. in the same scenario: $h
     local casedir="${casedir:-"$repo#$branch"}"
     local GROUP="${GROUP:-0}"
     local dry_run="${dry_run:-""}"
-    local cmd="$dry_run openqa-clone-job $clone_args \"$host\" \"$job\" _GROUP=\"$GROUP\" TEST=\"$test\" BUILD=\"$build\" CASEDIR=\"$casedir\" PRODUCTDIR=\"$productdir\" NEEDLES_DIR=\"$needles_dir\""
+    local scriptdir
+    scriptdir=$(dirname "${BASH_SOURCE[0]}")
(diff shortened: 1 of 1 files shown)


Martchus (approved):

Martchus (changes requested):
I suppose shellcheck is at least right with the point that you need to use double quotes. (Yes, you need to use double quotes within double quotes to avoid expansion of `BASH_SOURCE`.)

Martchus (approved):

script/openqa-clone-custom-git-refspec
```diff
@@ -135,7 +135,9 @@ Please try 'curl $json_url' or select another job, e.g. in the same scenario: $h
     local casedir="${casedir:-"$repo#$branch"}"
     local GROUP="${GROUP:-0}"
     local dry_run="${dry_run:-""}"
-    local cmd="$dry_run openqa-clone-job $clone_args \"$host\" \"$job\" _GROUP=\"$GROUP\" TEST=\"$test\" BUILD=\"$build\" CASEDIR=\"$casedir\" PRODUCTDIR=\"$productdir\" NEEDLES_DIR=\"$needles_dir\""
+    local scriptdir
+    scriptdir=$(dirname "${BASH_SOURCE[0]}")
```
Amrysliu, line 139: I am not familiar with shell script, just a  small question: why not write `local script=$(dirname "${BASH_SOURCE[0]}")` directly?

Martchus: Likely shellcheck didn't like it.

kalikiana: Yes. Supposedly it's better to do it separately because it could hide problems otherwise.

Amrysliu (approved):

okurz:
In before one could overwrite which command is called by specifying PATH. That's not possible anymore. Have you thought of that caveat? :)

PR #3678, opened 2021-01-19
Add SUSE Linux Enterprise Micro in the external reporting

jlausuch:
We have SLE Micro product which is different than OpenSUSE MicroOS

PR #3683, opened 2021-01-20
Fix error reporting link for SLE Micro

Martchus:
@jlausuch 
> Looks like the change `Add SUSE Linux Enterprise Micro in the external reporting`
> in this PR https://github.com/os-autoinst/openQA/pull/3678
> didn’t work..
> I don’t see the “bug” icon when clicking on some frame…
> https://openqa.suse.de/tests/5318460#step/disk_boot/7

> Any ideas why?

It needs it own branch and apparently the product name does not contain the `VERSION` here.

### DIFF
--- a/templates/webapi/branding/openSUSE/external_reporting.html.ep
+++ b/templates/webapi/branding/openSUSE/external_reporting.html.ep
@@ -10,20 +10,20 @@
 
 <% my %distri_to_product_url_new = (
     sle => 'https://bugzilla.suse.com/enter_bug.cgi',
+    'sle-micro' => 'https://bugzilla.suse.com/enter_bug.cgi',
     opensuse => 'https://bugzilla.opensuse.org/enter_bug.cgi',
     caasp => 'https://bugzilla.suse.com/enter_bug.cgi',
     openqa => 'https://progress.opensuse.org/projects/openqav3/issues/new',
     kubic => 'https://bugzilla.opensuse.org/enter_bug.cgi',
     microos => 'https://bugzilla.opensuse.org/enter_bug.cgi',
-    'sle-micro' => 'https://bugzilla.suse.com/enter_bug.cgi',
 );%>
 <% my %distri_to_prod = (
     sle => 'SUSE Linux Enterprise',
+    'sle-micro' => 'SUSE Linux Enterprise',
     opensuse => 'openSUSE',
     caasp => 'SUSE CaaS Platform',
     kubic => 'openSUSE',
     microos => 'openSUSE',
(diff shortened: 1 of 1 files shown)


templates/webapi/branding/openSUSE/external_reporting.html.ep
```diff
@@ -10,20 +10,20 @@
...
     opensuse => 'https://bugzilla.opensuse.org/enter_bug.cgi',
     caasp => 'https://bugzilla.suse.com/enter_bug.cgi',
     openqa => 'https://progress.opensuse.org/projects/openqav3/issues/new',
     kubic => 'https://bugzilla.opensuse.org/enter_bug.cgi',
     microos => 'https://bugzilla.opensuse.org/enter_bug.cgi',
-    'sle-micro' => 'https://bugzilla.suse.com/enter_bug.cgi',
 );%>
 <% my %distri_to_prod = (
     sle => 'SUSE Linux Enterprise',
+    'sle-micro' => 'SUSE Linux Enterprise',
```
jlausuch, line 22:
```suggestion
    'sle-micro' => 'SUSE Linux Enterprise Micro',
```

jlausuch:
Thanks for the fix. I guess at some point we can remove caasp entries here.

Martchus:
I wanted to reply on your comment but it didn't work. Maybe you've removed it. Nevertheless, here the answer:

I'm mapping this just to "SUSE Linux Enterprise" here on purpose to stay consistent with the code which is already there, e.g. "opensuse" and "microos" are both mapped to "openSUSE". For the " Micro" appendix the `$product` variable is set while `%distri_to_prod` remains defining the mapping from `DISTRI` to "generic products". By the way, I've tested this locally with the job settings you've mentioned in your mail and it produces a link like the one from your initial PR.

jlausuch:
> I wanted to reply on your comment but it didn't work. Maybe you've removed it. Nevertheless, here the answer:
> 
> I'm mapping this just to "SUSE Linux Enterprise" here on purpose to stay consistent with the code which is already there, e.g. "opensuse" and "microos" are both mapped to "openSUSE". For the " Micro" appendix the `$product` variable is set while `%distri_to_prod` remains defining the mapping from `DISTRI` to "generic products". By the way, I've tested this locally with the job settings you've mentioned in your mail and it produces a link like the one from your initial PR.

yeah, I removed cause I found the reason, anyway, thanks for the clarification. Makes sense.

okurz (approved):

Amrysliu (approved):

okurz:
I now deployed this version on osd

jlausuch:
Thanks a lot for your support!

PR #3684, opened 2021-01-20
CloneJob: Set log level depending on verbose switch

kalikiana:
Also, add tests with and without verbose, to confirm the script is silent or reports settings and config filename.

See: https://progress.opensuse.org/issues/88053

### DIFF
--- a/lib/OpenQA/Script/CloneJob.pm
+++ b/lib/OpenQA/Script/CloneJob.pm
@@ -49,6 +49,7 @@ sub is_global_setting {
 sub clone_job_apply_settings {
     my ($argv, $depth, $settings, $options) = @_;
 
+    $ENV{MOJO_LOG_LEVEL} = 'debug' if $options->{verbose};
     delete $settings->{NAME};         # usually autocreated
     $settings->{is_clone_job} = 1;    # used to figure out if this is a clone operation
 
--- a/lib/OpenQA/UserAgent.pm
+++ b/lib/OpenQA/UserAgent.pm
@@ -16,6 +16,8 @@
 package OpenQA::UserAgent;
 use Mojo::Base 'Mojo::UserAgent';
 
+use OpenQA::Log 'log_info';
+
 use Mojo::Util 'hmac_sha1_sum';
 use Config::IniFiles;
 use Scalar::Util ();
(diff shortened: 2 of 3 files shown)


t/09-job_clone.t
```diff
@@ -93,7 +94,19 @@ subtest 'get job' => sub {
     qr/failed to get job '$job_id'/, 'invalid job id results in error';
 
     $job_id = 99937;
-    lives_ok { clone_job_get_job($job_id, $remote, $remote_url, \%options) } 'got job';
+    combined_like { clone_job_get_job($job_id, $remote, $remote_url, \%options) } qr/^$/ , 'got job';
```
okurz, line 97: does this really work? Would this fail if there *is* output? I would assume that /^$/ maybe also matches a newline after actual output

kalikiana: It fails if `verbose=>1` is added to the test, so yes, if we added anything else in the future producing output, like other logs in libs, it would fail.

lib/OpenQA/Script/CloneJob.pm
```diff
@@ -49,6 +49,7 @@ sub is_global_setting {
 sub clone_job_apply_settings {
     my ($argv, $depth, $settings, $options) = @_;
 
+    $ENV{MOJO_LOG_LEVEL} = 'info' if $options->{verbose};
```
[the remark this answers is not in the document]
kalikiana: I'm not sure the `-vv` bit is trivial to add, since it would touch both script and lib code, if we want to use it elsewhere. I'll just make it *debug* for now.

lib/OpenQA/UserAgent.pm
```diff
@@ -38,6 +40,7 @@ sub new {
         for my $path (@cfgpaths) {
             my $file = $path . '/client.conf';
             next unless $file && -r $file;
+            log_info("Reading config from $file");
```
Martchus, line 43: I guess we still shouldn't log messages unconditionally in library code. And yes, it depends on the global log level but I'd rather tend to have a switch within the library. Or the library is simply able to return the path it is using to the tool which can then decide whether to log it or not.

kalikiana: Maybe we can discuss it in the weekly later. From my perspective we do this all over the place in openQA library code. It's not "new". Except we have more than one config file handler so unconditional log calls do different things depending on the frontend.

PR #3685, opened 2021-01-20
fullstack: Improve error/subtest handling

kalikiana:
- The subtest fails cleaner if it has a check before the wait
- We can have a more useful message in the wait for the dev console

See: https://progress.opensuse.org/issues/80800

Split off from #3655

### DIFF
--- a/t/lib/OpenQA/Test/FullstackUtils.pm
+++ b/t/lib/OpenQA/Test/FullstackUtils.pm
@@ -186,11 +186,11 @@ sub wait_for_developer_console_available {
     my ($driver) = @_;
 
     wait_for_or_bail_out {
-        note('waiting for worker to propagate URL for os-autoinst cmd srv');
         $driver->refresh;
         wait_for_ajax(msg => 'developer console available');
         my $console_form = $driver->find_element('#ws_console_form');
         my $text         = $console_form->get_text() // '';
+        note("worker response: $text");
         return $text =~ qr/The command server is not available./ ? 0 : 1;
     }
     'URL for os-autoinst cmd srv', {timeout => 120, interval => 2};
--- a/t/full-stack.t
+++ b/t/full-stack.t
@@ -121,9 +121,9 @@ sub start_worker_and_schedule {
 sub autoinst_log { path($resultdir, '00000', sprintf("%08d", shift) . "-$job_name")->child('autoinst-log.txt') }
 
 start_worker_and_schedule;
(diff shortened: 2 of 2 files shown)


t/lib/OpenQA/Test/FullstackUtils.pm
```diff
@@ -186,11 +186,11 @@ sub wait_for_developer_console_available {
     my ($driver) = @_;
 
     wait_for_or_bail_out {
-        note('waiting for worker to propagate URL for os-autoinst cmd srv');
         $driver->refresh;
         wait_for_ajax(msg => 'developer console available');
         my $console_form = $driver->find_element('#ws_console_form');
         my $text         = $console_form->get_text() // '';
+        note("worker response: $text");
```
okurz, line 193: could you please provide an example output to explain where the new note structure would be beneficial?

kalikiana: Well, the code checks for it in the next line

Martchus: We don't get a response from the worker here so the note makes no sense. We're just waiting until the developer console becomes available and the note you've been removing refers to what happens in the background.

For debugging it might indeed be useful to see what the developer console contained, though.

Martchus (changes requested):

t/lib/OpenQA/Test/FullstackUtils.pm
kalikiana: `waiting for worker to propagate URL for os-autoinst cmd srv` and `URL for os-autoinst cmd srv` say the same thing, no?

kalikiana: Maybe you have ideas on what else to add? I added the `$text` so in case of an error it'd be more obvious from logs that it contained something odd.

kalikiana: Of course if it's not running at all, that's not much help. Maybe we should log the page itself if it's not rendered correctly or not responding 🤔

Martchus: No, they don't say the same thing. They just state we're waiting for the worker to do something. They don't state that the output within the developer console is a direct response from the worker.

kalikiana: Perhaps I should've said, to me, they say the same thing. In that spirit I'll close the proposed changes if they don't make sense to others.

PR #3686, opened 2021-01-20
fullstack: Unify unstable_worker loop and sleep

kalikiana:
- Always use IOLOOP, also for indefinite wait
- Clarify the intention of the code
- Sleep unconditionally
- Report coverage at the end

Split off from #3655

### DIFF
--- a/t/lib/OpenQA/Test/Utils.pm
+++ b/t/lib/OpenQA/Test/Utils.pm
@@ -454,7 +454,9 @@ sub unstable_worker {
     # the help of the Doctor would be really appreciated here.
     my ($apikey, $apisecret, $host, $instance, $ticks, $sleep) = @_;
     note("Starting unstable worker. Instance: $instance for host $host");
-    $ticks = 1 unless defined $ticks;
+    # Block the event loop for the specified number of events or 'inf' for indefinitely
+    $ticks = $ticks // 1;
+    $sleep //= 0;
 
     my $h = _setup_sigchld_handler 'openqa-worker-unstable', start sub {
         _setup_sub_process 'openqa-worker-unstable';
@@ -467,18 +469,16 @@ sub unstable_worker {
             });
         setup_worker($worker, $host);
         $worker->init();
-        if ($ticks < 0) {
-            Mojo::IOLoop->singleton->start;
-        }
-        else {
-            Mojo::IOLoop->singleton->one_tick for (0 .. $ticks);
-        }
+        sleep $sleep;
+        my $timer = Mojo::IOLoop->timer(
+            $ticks => sub {
+                Mojo::IOLoop->stop;    # uncoverable statement
+            });
+        Mojo::IOLoop->start;
+        Mojo::IOLoop->remove($timer);
         Devel::Cover::report() if Devel::Cover->can('report');
-        if ($sleep) {    # uncoverable statement
-            1 while sleep $sleep;    # uncoverable statement
-        }    # uncoverable statement
     };
-    sleep $sleep if $sleep;
+    sleep $sleep;
     return $h;
 }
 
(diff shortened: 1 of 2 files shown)


t/lib/OpenQA/Test/Utils.pm
```diff
@@ -454,7 +454,9 @@ sub unstable_worker {
     # the help of the Doctor would be really appreciated here.
     my ($apikey, $apisecret, $host, $instance, $ticks, $sleep) = @_;
     note("Starting unstable worker. Instance: $instance for host $host");
-    $ticks = 1 unless defined $ticks;
+    # Block the event loop for the specified number of events or indefinitely if -1 is specified
+    $ticks = $ticks == -1 ? ~0 >> 1 : ($ticks // 1);
```
okurz, line 458: I don't understand this. Isn't `-1` always true?

kalikiana: this should be equivalent to:

```perl
if ($ticks == -1) $ticks = ~0 >> 1;
else { $ticks = $ticks // 1 };
```

Martchus:
> Always use ticks, also for indefinite wait

Not sure whether that will cause any disturbances. Maybe @kraih knows.

t/lib/OpenQA/Test/Utils.pm
okurz: oh right. Stupid me missed the `==`.

okurz:
have you seen the CI failure?:

```
[19:15:51] t/full-stack.t .. 374/? # full result panel contents:
# State: scheduled
# Scheduled product: job has not been created by posting an ISO
# 50
[19:15:51] t/full-stack.t .. 376/? 
#   Failed test 'test 1 is running'
#   at t/full-stack.t line 124.
Bailout called.  Further testing stopped:  URL for os-autoinst cmd srv not available
FAILED--Further testing stopped: URL for os-autoinst cmd srv not available
```

t/lib/OpenQA/Test/Utils.pm
kraih: I don't find `~0 >> 1` particularly readable. How about just using a boring old constant?

kraih:
> > Always use ticks, also for indefinite wait
> 
> Not sure whether that will cause any disturbances. Maybe @kraih knows.

Possible. If we were to rely on worker code calling `Mojo::IOLoop->stop` or similar that wouldn't work anymore. I guess the patch makes it a little more consistent, but i'm having a hard time seeing real benefits.

t/lib/OpenQA/Test/Utils.pm
perlpunk: Does the method accept `'inf'` as a value for the ticks? :)

kraih: I'm assuming the worker here is unstable because it only runs a few ticks and will therefore just vanish at some point while processing a job. It's a bit sketchy, but if the number is low enough should be reliable enough because I/O events will be spaced out enough. If it were relying more on timers this would be super inconsistent and unstable.

A much better strategy would have been to always use `Mojo::IOLoop->start` and then hook into the worker instance to call `Mojo::IOLoop->stop` at a very specific point to simulate an unstable worker more consistently.

kraih: Really not a fan of relying on randomness for tests.

kraih: External factors like the size of the kernel send buffer will influence what this test does, because ticks are mostly one I/O event per socket. So the more packets there are to read the more ticks you waste with it... and so on. Not a great strategy. I'd rather monkey patch a `Mojo::IOLoop->stop` into the worker.

Martchus:
I'm also having a hard time to see the benefits. The ideas from @kraih in the other comment thread would be a better alternative (if we want to improve here).

kalikiana:
> I'm also having a hard time to see the benefits. The ideas from @kraih in the other comment thread would be a better alternative (if we want to improve here).

The benefit is to unify the code paths and make coverage consistent. Because the infinite case is the same as the  one-off case.

PR #3688, opened 2021-01-21
Revert "client: Log what config file is being read"

perlpunk:
This reverts commit b3ee9d2139dd77d6ad8ddebfb597fd81af7ad9f8. (https://github.com/os-autoinst/openQA/pull/3679)

Issue: https://progress.opensuse.org/issues/88053

### DIFF
--- a/lib/OpenQA/UserAgent.pm
+++ b/lib/OpenQA/UserAgent.pm
@@ -16,8 +16,6 @@
 package OpenQA::UserAgent;
 use Mojo::Base 'Mojo::UserAgent';
 
-use OpenQA::Log 'log_info';
-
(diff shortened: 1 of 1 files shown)


Martchus (approved):
Maybe we can write to stderr instead. However, it is generally questionable to do logging unconditionally in library code. It might even be used by a graphical fronted. For now, let's just get rid of it again.

kalikiana:
> Maybe we can write to stderr instead. However, it is generally questionable to do logging unconditionally in library code. It might even be used by a graphical fronted. For now, let's just get rid of it again.

That's what #3684 addresses. But I'm okay with the revert, too.

kalikiana (approved):

kraih (approved):

PR #3690, opened 2021-01-22
Convert OpenQA::WebAPI to a more modern Perl dialect with signatures

kraih:
This is a little experiment to get a conversation started about the topic. [Signatures](https://perldoc.perl.org/perlsub#Signatures) have been available as an experimental feature since Perl 5.20 (released 2014) and have since been very stable. In fact the upstream [Mojolicious documentation](https://docs.mojolicious.org/) only uses this style in all examples since last year, without encountering any problems at all.

Aside from resulting in shorter and cleaner code, signatures also enforce an arity check, that means if the wrong number of arguments is passed to a subroutine it will result in an exception. Which is very nice for uncovering many bugs early on.

```
$ perl -Mojo -E 'sub foo ($bar) { } foo()'
Too few arguments for subroutine 'main::foo' at -e line 1.
$ perl -Mojo -E 'sub foo ($bar) { } foo(1, 2)'
Too many arguments for subroutine 'main::foo' at -e line 1.
```

As this PR shows, we do not have to covert all code at once. Modules and scripts can be converted slowly over time (even partially, since `@_` still works), and we could even decide to leave certain parts like the worker related modules with the old style so the Perl version requirement is lower for workers (i would not recommend it, but it's an option we have). Almost forgot to mention, since Perl 5.28 or so signatures are also significantly faster than using `@_`.

Of course there are risks too. The feature is technically still experimental, and that means it could break with the next Perl major release (5.34 in summer this year). But the risks are mitigated by the fact that Perl releases happen on a schedule, and we would be aware of any significant breakage months in advance. Considering that signatures are already widely used across the Perl community, breaking changes are highly unlikely to happen at all though.

This PR is intentionally very limited in scope so we can safely test the waters. And if it does end up causing too many problems for our users we can quickly revert it again.

Looking forward to your thoughts on the matter. 😃


kraih:
While making this patch i noticed that we do not technically have a Perl version requirement for openQA. I'm guessing that the implicit Perl 5.16 one from Mojolicious might be the highest at the moment. So this would be a bump up to Perl 5.20.

perlpunk (approved):
If it's ok to increase the perl version, I'm all for it.
I also use signatures in some of my projects already.

kalikiana:
> available as an experimental feature since Perl 5.20

So, as far as I understand there's still a chance of bug-fixes in the ABI in very specific cases, but we don't have to expect radical changes or this going away altogether.

I like the gradual approach. We can see if, for some reason even the minimum requirement could be a problem, and these small changes shouldn't be a big hassle to undo if needed. We have a lot more style fixes/ tidy churn than these anyway.

kraih:
Fwiw. i use signatures in all my new projects and have recently started [converting Cavil](https://github.com/openSUSE/cavil/commit/67b4dabd61bd62a06c96c15b0ae4f2c60cae35cc) too.

Martchus (commented):
This means it can not run under Leap 42.x anymore. I suppose that's ok at this point. I've only added the "not-ready" flag to prevent mergify from merging this while we likely want to give people some time to intervene.

okurz (approved):
I like it, you remove more lines than adding :)

Martchus:
@AdamWill Would requiring at least Perl 5.20 ok?

AdamWill:
No problem, we're at 5.32. Thanks for checking.

kraih:
> I like it, you remove more lines than adding :)

If we do end up converting everything that's going to be thousands of lines removed. 😁

Martchus (approved):

kraih:
Since everybody approved i'm going to remove the label. 😉

PR #3604, opened 2020-11-30
Symlink CASEDIR and set values to relative path

Amrysliu:
1. symlink test repository from CASEDIR to POOLDIR (using CASEDIR's 
basename as the destdir name).
2. set the CASEDIR and PRODUCTDIR as relative values.
3. when uing `openqa-clone-custom-git-refspec`, set the NEEDLES_DIR to
`needles`, because code in openQA worker will symlink the needles
directory to it.

Verify run:
The source job run on a remote worker: http://10.67.19.103/tests/413
The clone job run on a local worker: http://10.67.19.103/tests/414#settings
A simple case use relative path as CASEDIR and PRODUCTDIR: http://10.67.19.103/tests/408/file/vars.json

Relate: https://progress.opensuse.org/issues/80372

PR #3693, opened 2021-01-23
Revert "Symlink CASEDIR and set values to relative path"

okurz:
Reverts os-autoinst/openQA#3604 due to the report in
https://github.com/os-autoinst/openQA/pull/3604#issuecomment-765888161


okurz:
@coolo @Amrysliu

PR #3694, opened 2021-01-25
tests: Test expected output from job cloning

kalikiana:
This adds test coverage for the regression addressed by #3688.

- In non-verbose mode there should be no output in normal operation.
- In verbose mode a dump of the job settings is expected.

### DIFF
--- a/t/09-job_clone.t
+++ b/t/09-job_clone.t
@@ -28,6 +28,7 @@ use OpenQA::Test::Utils qw(create_webapi stop_service);
 use OpenQA::Test::TimeLimit '20';
 use Test::Mojo;
 use Test::Warnings ':report_warnings';
+use Test::Output 'combined_like';
 
 OpenQA::Test::Database->new->create(fixtures_glob => '01-jobs.pl');
 my $t        = client(Test::Mojo->new('OpenQA::WebAPI'));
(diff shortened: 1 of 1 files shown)


Martchus (approved):

t/09-job_clone.t
```diff
@@ -93,7 +94,21 @@ subtest 'get job' => sub {
...
-    lives_ok { clone_job_get_job($job_id, $remote, $remote_url, \%options) } 'got job';
+    combined_like { clone_job_get_job($job_id, $remote, $remote_url, \%options) } qr/^$/, 'got job';
+};
+
+subtest 'get job with verbose output' => sub {
+    my $config = tempdir;
+    $config->child("client.conf")->touch;
+    $ENV{OPENQA_CONFIG} = $config;
+    my $temp_assetdir = tempdir;
+    my %options       = (dir => $temp_assetdir, host => $host, from => $host, verbose => 1);
```
Amrysliu, line 105: why the `$ENV{OPENQA_CONFIG}` and `$temp_assetdir` are set here?  If I understand correctly, the functions `create_url_handler` and `clone_job_get_job` don't need them.

okurz (commented):
+1

t/09-job_clone.t
kalikiana: I added those because otherwise these code paths are skipped. The regression with logging in `create_url_handler` could only reproduced with a `.conf` file in place. So this is not needed to pass the test, but it's necessary to have better coverage. Otherwise a regression would not be caught by this test.

kalikiana: I suppose I'll add a comment to point this out. Thank you for asking.

okurz:
failed in https://circleci.com/gh/os-autoinst/openQA/51805?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link . Please make sure this is covered in tests, please don't just retrigger or something :)

kalikiana:
> failed in https://circleci.com/gh/os-autoinst/openQA/51805?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link . Please make sure this is covered in tests, please don't just retrigger or something :)

Not introduced here. Also seen in #3635. This is [poo#66718](https://progress.opensuse.org/issues/66718)

okurz (approved):
thanks for crosschecking

PR #3695, opened 2021-01-25
tests: Drop unstable_worker and tests based on it

kalikiana:
Going by the observation that none of us is really happy with how these tests work we might be better off with other existing tests covering the same thing, or even new tests.

### DIFF
--- a/t/05-scheduler-full.t
+++ b/t/05-scheduler-full.t
@@ -44,7 +44,7 @@ use OpenQA::Test::Utils qw(
   mock_service_ports setup_mojo_app_with_default_worker_timeout
   setup_fullstack_temp_dir create_user_for_workers
   create_webapi setup_share_dir create_websocket_server
-  stop_service unstable_worker
+  stop_service
   unresponsive_worker broken_worker rejective_worker
 );
 use OpenQA::Test::TimeLimit '150';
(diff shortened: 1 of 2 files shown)


t/05-scheduler-full.t
```diff
@@ -136,146 +136,6 @@ subtest 'Scheduler worker job allocation' => sub {
...
-    @workers = broken_worker(@$worker_settings, 3, 'out of order');
-    wait_for_worker($schema, 5);
-    $allocated = scheduler_step();
-    is @$allocated, 0, 'scheduler does not consider broken worker for allocating job';
-    stop_workers;
-    dead_workers($schema);
-
-    # simulate a worker in idle state that rejects all jobs assigned to it
-    @workers = rejective_worker(@$worker_settings, 3, 'rejection reason');
-    wait_for_worker($schema, 5);
```
Martchus, line 162: The tests with the broken and rejective worker should stay. These workers should behave consistently.

t/05-scheduler-full.t
```diff
@@ -136,146 +136,6 @@ subtest 'Scheduler worker job allocation' => sub {
...
-        last if $jobs->find(99982)->state eq OpenQA::Jobs::Constants::DONE;
-        sleep .2;
-    }
-
-    my $job = $jobs->find(99982);
-    is $job->state, OpenQA::Jobs::Constants::DONE,
-      'running job set to done if its worker re-connects claiming not to work on it anymore';
-    is $job->result, OpenQA::Jobs::Constants::INCOMPLETE,
-      'running job incompleted if its worker re-connects claiming not to work on it anymore';
-    like $job->reason, qr/abandoned: associated worker .+:\d+ re-connected but abandoned the job/, 'reason is set';
```
Martchus, line 224: Not sure whether this is already covered (and also actually checked!) somewhere else.

Martchus (changes requested):

t/05-scheduler-full.t
kalikiana: I was hoping to see coverage results telling me that.

okurz:
Please also crosscheck to keep code coverage up. The idea was that we delete test code assuming there is coverage provided by other tests. Better start small, so as @Martchus suggested, keep the "broken and rejective workers"

okurz:
@kalikiana maybe it helps readability of coverage reports if we merge https://github.com/os-autoinst/openQA/pull/3728 first, wait for tests in master to produce proper coverage base reports and then you simply try again and we see what parts are not covered after your code removal. The idea was really to not need to add sophisticated tests but rather remove unreliabe existing stuff, right?

okurz:
please rebase now so that we can see how the current coverage report looks like

PR #3696, opened 2021-01-26
Rework symlink CASEDIR and set values to relative path

Amrysliu:
One commit is the same as #3604.
Another commit is used to fix the regression issue: https://github.com/os-autoinst/openQA/pull/3604#issuecomment-765888161

Verify run: http://10.67.19.103/tests/432

### DIFF
--- a/lib/OpenQA/Utils.pm
+++ b/lib/OpenQA/Utils.pm
@@ -85,6 +85,7 @@ our @EXPORT  = qw(
   change_sec_to_word
   find_video_files
   fix_top_level_help
+  looks_like_url_with_scheme
 );
 
 our @EXPORT_OK = qw(
@@ -898,5 +899,7 @@ sub find_video_files { path(shift)->list_tree->grep(VIDEO_FILE_NAME_REGEX) }
 
 # workaround https://github.com/mojolicious/mojo/issues/1629
 sub fix_top_level_help { @ARGV = () if ($ARGV[0] // '') =~ qr/^(-h|(--)?help)$/ }
+
+sub looks_like_url_with_scheme { return !!Mojo::URL->new(shift)->scheme }
 
 1;
--- a/lib/OpenQA/Utils.pm
+++ b/lib/OpenQA/Utils.pm
@@ -139,7 +139,7 @@ sub productdir {
     my ($distri, $version, $rootfortests) = @_;
 
     my $dir = testcasedir($distri, $version, $rootfortests);
-    return $dir . "/products/$distri" if -e "$dir/products/$distri";
+    return "$dir/products/$distri" if $distri && -e "$dir/products/$distri";
     return $dir;
 }
 
--- a/lib/OpenQA/Worker/Engines/isotovideo.pm
+++ b/lib/OpenQA/Worker/Engines/isotovideo.pm
@@ -20,7 +20,7 @@ use warnings;
 
 use OpenQA::Constants qw(WORKER_SR_DONE WORKER_EC_CACHE_FAILURE WORKER_EC_ASSET_FAILURE WORKER_SR_DIED);
 use OpenQA::Log qw(log_error log_info log_debug log_warning get_channel_handle);
-use OpenQA::Utils qw(asset_type_from_setting base_host locate_asset);
+use OpenQA::Utils qw(asset_type_from_setting base_host locate_asset looks_like_url_with_scheme);
 use POSIX qw(:sys_wait_h strftime uname _exit);
 use Mojo::JSON 'encode_json';    # booleans
 use Cpanel::JSON::XS ();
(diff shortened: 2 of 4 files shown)


lib/OpenQA/Utils.pm
```diff
@@ -899,4 +900,6 @@ sub find_video_files { path(shift)->list_tree->grep(VIDEO_FILE_NAME_REGEX) }
 # workaround https://github.com/mojolicious/mojo/issues/1629
 sub fix_top_level_help { @ARGV = () if ($ARGV[0] // '') =~ qr/^(-h|(--)?help)$/ }
 
+sub is_url { return shift =~ /http(s)?\:\/\//; }
```
Martchus, line 903: This checks for an URL only with the schemes http/https and an authority (even if empty). This makes sense here but maybe `is_http_url` would be a better name.

Martchus (commented):
I hope that was the last special case.

lib/OpenQA/Utils.pm
```diff
@@ -139,7 +140,7 @@ sub productdir {
     my ($distri, $version, $rootfortests) = @_;
 
     my $dir = testcasedir($distri, $version, $rootfortests);
-    return $dir . "/products/$distri" if -e "$dir/products/$distri";
+    return $dir . "/products/$distri" if $distri && -e "$dir/products/$distri";
```
Martchus, line 143:
```suggestion
    return "$dir/products/$distri" if $distri && -e "$dir/products/$distri";
```

lib/OpenQA/Utils.pm
Martchus, line 903:
```suggestion
sub is_http_url { shift =~ /http(s)?\:\/\// }
```

Martchus (commented):
Some style suggestions since you're changing these places anyways.

Amrysliu:
> I hope that was the last special case.

I have tested the cases I could think out. we could test it on o3?

Martchus (approved):

Martchus:
> we could test it on o3?

If you like you can hot-patch the change on o3 (or relevant OSD workers) before merging this PR. (If not, remove the "acceptance-tests-needed" label again.)

kalikiana (approved):

lib/OpenQA/Utils.pm
```diff
@@ -899,4 +900,6 @@ sub find_video_files { path(shift)->list_tree->grep(VIDEO_FILE_NAME_REGEX) }
 # workaround https://github.com/mojolicious/mojo/issues/1629
 sub fix_top_level_help { @ARGV = () if ($ARGV[0] // '') =~ qr/^(-h|(--)?help)$/ }
 
+sub is_http_url { shift =~ /http(s)?\:\/\// }
```
okurz, line 903: As we just forward the URL to os-autoinst which calls `git clone` we should not only accept http(s) but anything that git takes. In the commit message you correctly stated just "URL", not that it has to be HTTP(S). Better don't try to roll your own but use something from a common perl library to detect if we have a URL. E.g. use https://docs.mojolicious.org/Mojo/URL

okurz (changes requested):

lib/OpenQA/Utils.pm
okurz: maybe you got it the wrong way around, see https://github.com/os-autoinst/openQA/pull/3696#discussion_r564587717 . I think the method name was correct, the implementation was wrong and should cover more than just http(s)

lib/OpenQA/Utils.pm
okurz: I guess we can live with the limitation about the colon

lib/OpenQA/Worker/Engines/isotovideo.pm
```diff
@@ -326,11 +338,27 @@ sub engine_workit {
...
+    $vars{CASEDIR}    //= $target_name;
+    $vars{PRODUCTDIR} //= substr($productdir, rindex($casedir, $target_name));
+
+    if (!is_http_url($vars{CASEDIR})) {
+        if (my $error = _link_repo($casedir, $pooldir, $target_name)) { return $error }
+    }
 
+    # if NEEDLES_DIR is an absolute path, it means that users or openqa-clone-custom-git-refspec specify it,
+    # if NEEDLES_DIR is an URL address, it means that users specify it and want to get the needles from URL.
+    # These two scenarios, we do not need to do the symlink.
```
okurz, line 355: Can you phrase the comment in a way that it describes why we can not create links rather than "not need it"?

Amrysliu:
> > we could test it on o3?
> 
> If you like you can hot-patch the change on o3 (or relevant OSD workers) before merging this PR. (If not, remove the "acceptance-tests-needed" label again.)

how to hot-patch the change on o3 (or relevant OSD workers)?  I thought we could deploy it on o3 first, and see the results, if everything is ok, then deploy it on OSD.

lib/OpenQA/Utils.pm
Amrysliu: Not sure if I understand correctly, so I should change the function like this:
```
my $url = Mojo::URL->new(shift);
return 1 if ($url->scheme);
return 0;
```

okurz:
```suggestion
sub is_url { return !!Mojo::URL->new(shift)->scheme }
```

Martchus:
I was just mentioning possibilities. On o3 workers it is likely not possible to easily hot patch because the workers are transactional servers so using selected OSD workers would be an alternative. But we can also just merge it and let it go though the usual deployment.

lib/OpenQA/Utils.pm
```diff
@@ -899,4 +900,6 @@ sub find_video_files { path(shift)->list_tree->grep(VIDEO_FILE_NAME_REGEX) }
 # workaround https://github.com/mojolicious/mojo/issues/1629
 sub fix_top_level_help { @ARGV = () if ($ARGV[0] // '') =~ qr/^(-h|(--)?help)$/ }
 
+sub is_url { return !!Mojo::URL->new(shift)->scheme }
```
Martchus, line 903: In consistency with `looks_like_number` from `Scalar::Utils`:

```suggestion
sub looks_like_url_with_scheme { Mojo::URL->new(shift)->scheme ne '' }
```

I guess `!!` vs. `ne ''` makes no real difference in practice but I wouldn't exclude other falsy values like `0` here. (Scheme is not supposed to return `undef`.)

Amrysliu: The function name is better.

Amrysliu: Scheme returns `undef`, see the log
```
Use of uninitialized value in string ne at /home/liuxiaojing/code/repos/openQA/lib/OpenQA/Utils.pm line 903.
```

Martchus: Ah, then you can keep your `!!` version.

lib/OpenQA/Worker/Engines/isotovideo.pm
Amrysliu: modified. Please help me confirm that if the new words make sense.

okurz (approved):

PR #3701, opened 2021-01-29
Update patch for B::Deparse

perlpunk:
Issue: https://progress.opensuse.org/issues/40895

Now also $expr2 seems to be uninitialized.

    Use of uninitialized value $expr2 in concatenation (.) or string at /home/squamata/project/t/lib/OpenQA/Test/PatchDeparse.pm line 44.
    Use of uninitialized value $expr2 in substitution (s///) at /home/squamata/project/t/lib/OpenQA/Test/PatchDeparse.pm line 43.

Also update the version to 1.54 (perl 5.32.0) since the sub hasn't changed.

### DIFF
--- a/.circleci/ci-packages.txt
+++ b/.circleci/ci-packages.txt
@@ -139,7 +139,7 @@ perl-Module-Pluggable-5.2
 perl-Module-Runtime-0.016
 perl-Module-Runtime-Conflicts-0.003
 perl-Mojo-IOLoop-ReadWriteProcess-0.28
-perl-Mojolicious-8.71
+perl-Mojolicious-8.72
 perl-Mojolicious-Plugin-AssetPack-2.10
 perl-Mojolicious-Plugin-OAuth2-1.58
 perl-Mojo-Pg-4.23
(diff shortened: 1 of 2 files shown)


okurz (approved):

Martchus (approved):

.circleci/ci-packages.txt
```diff
@@ -139,7 +139,7 @@ perl-Module-Pluggable-5.2
 perl-Module-Runtime-0.016
 perl-Module-Runtime-Conflicts-0.003
 perl-Mojo-IOLoop-ReadWriteProcess-0.28
-perl-Mojolicious-8.71
+perl-Mojolicious-8.72
```
kraih, line 142: Why was the Mojolicious version update needed?

kraih: Just out of curiosity... not that i would ever complain about a Mojolicious version update. 😁

perlpunk: Because the previous version was already unavailable :)

PR #3702, opened 2021-01-30
Dependency cron 2021-01-30

openqabot:


perlpunk (approved):

okurz:
@Martchus this trivial PR is one reason why I think it's better to split PRs that introduce new dependencies because now your original PR can not be cleanly reverted anymore

PR #3703, opened 2021-01-30
Revert "Rework symlink CASEDIR and set values to relative path"

Amrysliu:
Reverts os-autoinst/openQA#3696

Here is a case that I did not consider: https://openqa.opensuse.org/tests/1609706#


okurz:
ok. I can merge that. On https://openqa.opensuse.org/tests?match=last_good_test I also saw the suspicion confirmed that any "last_good_tests" investigation job since this day (since the new version is deployed) is incomplete except of the ones on the old aarch64 workers which still have the old version

PR #3704, opened 2021-02-01
Use Mojo::UserAgent::socket_options if available to avoid relying on monkey patch

Martchus:
* Mojo::UserAgent::socket_options is available as of Mojolicious 8.72 so
  when using at least this version "Running patched getaddrinfo() …" should
  no longer occur within the debug log
* See https://progress.opensuse.org/issues/78390

### DIFF
--- a/lib/OpenQA/CacheService/Client.pm
+++ b/lib/OpenQA/CacheService/Client.pm
@@ -1,4 +1,4 @@
-# Copyright (C) 2018-2020 SUSE LLC
+# Copyright (C) 2018-2021 SUSE LLC
 #
 # This program is free software; you can redistribute it and/or modify
 # it under the terms of the GNU General Public License as published by
@@ -14,7 +14,7 @@
 # with this program; if not, see <http://www.gnu.org/licenses/>.
 
 package OpenQA::CacheService::Client;
-use Mojo::Base -base;
+use Mojo::Base -base, -signatures;
 
 use OpenQA::Worker::Settings;
 use OpenQA::CacheService::Request::Asset;
@@ -22,6 +22,7 @@ use OpenQA::CacheService::Request::Sync;
 use OpenQA::CacheService::Response::Info;
 use OpenQA::CacheService::Response::Status;
 use OpenQA::Utils qw(base_host service_port);
+use Socket qw(AF_INET IPPROTO_TCP SOCK_STREAM pack_sockaddr_in inet_aton);
 use Mojo::URL;
 use Mojo::File 'path';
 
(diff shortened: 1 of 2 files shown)


lib/OpenQA/Worker.pm
```diff
@@ -377,8 +377,21 @@ sub configure_cache_client {
...
+    # not supported" for local connections (see https://progress.opensuse.org/issues/78390#note-38)
+    # note: The socket_options function has only been added in Mojolicious 8.72. For older versions we still rely
+    #       on the monkey_patch within the BEGIN block. It could be removed when we stop supporting older Mojolicious
+    #       versions.
+    my %cache_service_address = (
+        family   => AF_INET,
+        protocol => IPPROTO_TCP,
+        socktype => SOCK_STREAM,
+        addr     => pack_sockaddr_in('9530', inet_aton('127.0.0.1')));
+    $ua->socket_options({PeerAddrInfo => [\%cache_service_address]}) if $ua->can('socket_options');
```
kraih, line 394: Better use `$ua->socket_options->{PeerAddrInfo} = [\%cache_service_address]`. Then you don't end up accidentally overriding other options that might have been set somewhere else (perhaps even inside `Mojo::UserAgent`).

kraih:
It would probably be better to just depend on Mojolicious 8.72.

lib/OpenQA/Worker.pm
Martchus: Right. I also moved it to the client module itself to catch all occurrences.

kraih (approved):

okurz (approved):

lib/OpenQA/CacheService/Client.pm
```diff
@@ -31,7 +32,27 @@ has attempts   => $ENV{OPENQA_CACHE_ATTEMPTS}           // 60;
...
+    my %cache_service_address = (
+        family   => AF_INET,
+        protocol => IPPROTO_TCP,
+        socktype => SOCK_STREAM,
+        addr     => pack_sockaddr_in(service_port('cache_service'), inet_aton('127.0.0.1')));
+    $ua->socket_options->{PeerAddrInfo} = [\%cache_service_address] if $ua->can('socket_options');
+    return $ua;
+};
+
+sub set_port ($self, $port) {
```
kraih, line 52: Oh, you added signatures after i did my review. This might require a spec file update for the worker to also depend on `perl >= 5.20.0`.

Martchus: #3707

PR #3709, opened 2021-02-05
script: Fallback to running openqa-bootstrap-container without pipe

okurz:
Older versions of systemd-run do not have the `-P` parameter. We
originally envisioned that newer versions of Leap would get support but
as the main version of systemd stays the same as in the SLE15 codestream
understandably also the parameters stay the same.

This commit actively checks if the parameter is available. If it is
available it is used, otherwise it will output a message to the user
accordingly.

This is based on a user support question brought up in text chat.

### DIFF
--- a/script/openqa-bootstrap-container
+++ b/script/openqa-bootstrap-container
@@ -1,9 +1,6 @@
 #!/bin/bash -e
 [ "$1" = "-h" ] || [ "$1" = "--help" ] && echo "Setup a working openQA installation in a systemd-nspawn container" && exit
 
-# This script doesn't work on Leap 15.0 as it makes use of the -P option of systemd-run
-# which is not available there
-
 if [ "$(id -ru)" != "0" ]; then
     echo "$0 must be run as root"
     exit 1
@@ -14,6 +11,14 @@ set -x
 
 CONTAINER_NAME="openqa1"
 CONTAINER_PATH="/var/lib/machines/${CONTAINER_NAME}"
+
+systemd_run_params=(-q -M "$CONTAINER_NAME")
+if systemd-run --help | grep '\-P'; then
+    systemd_run_params+=(-P)
+else
+    echo "Your version of systemd-run does not support the '-P' parameter,
+piped output from the container will not be available here"
+fi
 
 ARCH="${ARCH:=$(arch)}"
 if [ "$ARCH" = "x86_64" ]; then
@@ -79,6 +84,6 @@ systemctl start systemd-nspawn-openqa@$CONTAINER_NAME
 # ensure that the container is really running
 # ignore expected errors about 'Failed to create bus connection: Protocol error' and restarting error
 while ! timeout -s9 2 systemd-run -qPM $CONTAINER_NAME /bin/bash -c whoami /dev/null 2>&1 ; do systemctl restart systemd-nspawn-openqa@$CONTAINER_NAME.service || true ; sleep 3 ; done
-systemd-run -qPM $CONTAINER_NAME /bin/bash -c '/usr/share/openqa/script/openqa-bootstrap'
+systemd-run "${systemd_run_params[@]}" /bin/bash -c '/usr/share/openqa/script/openqa-bootstrap'
 
 echo -e "$(tput setaf 2;tput bold)Your openQA container has been created. Run 'systemd-run -tM $CONTAINER_NAME /bin/bash' to get a shell in the container$(tput sgr0)"


script/openqa-bootstrap-container
```diff
@@ -15,6 +12,14 @@ set -x
 CONTAINER_NAME="openqa1"
 CONTAINER_PATH="/var/lib/machines/${CONTAINER_NAME}"
 
+systemd_run_params="${systemd_run_params:-"-qM $CONTAINER_NAME"}"
+if systemd-run --help | grep '\-P'; then
+    systemd_run_params="$systemd_run_params -P"
+else
+    echo "Your version of systemd-run does not support the '-P' parameter,
+    piped output from the container will not be available here"
```
Martchus, line 20: This way of wrapping the string will lead to surplus white-spaces being printed.

script/openqa-bootstrap-container
```diff
@@ -79,6 +84,6 @@ systemctl start systemd-nspawn-openqa@$CONTAINER_NAME
 # ensure that the container is really running
 # ignore expected errors about 'Failed to create bus connection: Protocol error' and restarting error
 while ! timeout -s9 2 systemd-run -qPM $CONTAINER_NAME /bin/bash -c whoami /dev/null 2>&1 ; do systemctl restart systemd-nspawn-openqa@$CONTAINER_NAME.service || true ; sleep 3 ; done
-systemd-run -qPM $CONTAINER_NAME /bin/bash -c '/usr/share/openqa/script/openqa-bootstrap'
+systemd-run "$systemd_run_params" /bin/bash -c '/usr/share/openqa/script/openqa-bootstrap'
```
Martchus, line 87: Does it really work to pass these parameters as one argument? Likely you'll have to get rid of the quoting. Note that having white-spaces within the `CONTAINER_NAME` didn't work before as well. To support white-spaces you could of course just use an array (and then keep the quotes).

Martchus (changes requested):

script/openqa-bootstrap-container
okurz: Well, I wanted the second line to be indented in the output same as is seen in the source, i.e. indented by 4 spaces. Do you have other preferences?

script/openqa-bootstrap-container
okurz: it does work like this as far as I can observe. Trying just `systemd-run "-qh"` seems to do the same as `systemd-run -qh`.

script/openqa-bootstrap-container
Martchus: Yes, it looks weird when multiple lines of one sentence have different indentation levels.

script/openqa-bootstrap-container
Martchus: Of course `systemd-run "-qh"` is equivalent to the unquoted version. My point was whether `… "-qM some-container-name"` is equivalent to `… -qM some-container-name"`. In the first case everything is passed as one argument (including the container name) and in the second case the container name is passed as separate argument. I suspect that only the last invocation will work, especially if the first character of the container name is also a valid short option.

script/openqa-bootstrap-container
okurz: ok, fine. I have deleted the indention on the second line

script/openqa-bootstrap-container
okurz: right. I changed to a bash array now.

Martchus (approved):

kalikiana (approved):

okurz:
Skipping tests as codecov is currently (likely still) broken and the changed code is not covered by any tests triggered during PR checks

script/openqa-bootstrap-container
```diff
@@ -79,6 +84,6 @@ systemctl start systemd-nspawn-openqa@$CONTAINER_NAME
 # ensure that the container is really running
 # ignore expected errors about 'Failed to create bus connection: Protocol error' and restarting error
 while ! timeout -s9 2 systemd-run -qPM $CONTAINER_NAME /bin/bash -c whoami /dev/null 2>&1 ; do systemctl restart systemd-nspawn-openqa@$CONTAINER_NAME.service || true ; sleep 3 ; done
-systemd-run -qPM $CONTAINER_NAME /bin/bash -c '/usr/share/openqa/script/openqa-bootstrap'
+systemd-run "${systemd_run_params[@]}" /bin/bash -c '/usr/share/openqa/script/openqa-bootstrap'
```
asdil12, line 87: Was this actually tested? Because this seems to drop the `-q` and the `-M $CONTAINER_NAME` parameters and I don't see how this could work without them.
Also in the line above `while ! timeout -s9 2 systemd-run -qPM $CONTAINER_NAME …` the `-P` option is used as well.

asdil12: sorry - somehow overlooked the line that set that params.

PR #3710, opened 2021-02-05
Use full path in openqa-reload-….service to support

Martchus:
to avoid
```
openqaworker-arm-1 systemd[1]: /usr/lib/systemd/system/openqa-reload-worker-auto-restart@.service:6: Executable path is not absolute: systemctl kill --signal SIGHUP openqa-worker-auto-restart@%i.service
```
when using it on production workers with older systemd version.

Related ticket: https://progress.opensuse.org/issues/80908#note-8


Martchus:
I've tested it on openqaworker-arm-1.

PR #3711, opened 2021-02-08
Fix problem in fullstack test leading to occasional failures

Martchus:
Sometimes the actual error message for the failed download is available and
therefore that message is logged (instead of the more generic one). The
reason should always contain the "Failed to download…" message so I'm
adding a check for that.

So far I've noticed that problem only when testing locally (to reproduce
https://progress.opensuse.org/issues/80800 which is **not** fixed by this
change).

### DIFF
--- a/t/full-stack.t
+++ b/t/full-stack.t
@@ -1,6 +1,6 @@
 #!/usr/bin/env perl
 
-# Copyright (C) 2016-2020 SUSE LLC
+# Copyright (C) 2016-2021 SUSE LLC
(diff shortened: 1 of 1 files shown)


okurz:
All failed tests timed out. Seems like everything slowed down a lot?!?

Martchus:
Not the first time I've seen this. Recently the tests also failed often due to the overall timeout.

kalikiana (approved):
Nice catch!

PR #3713, opened 2021-02-09
Prevent test timeouts with the introduction of "heavy" tests

okurz:
Combine "scheduler"+"developer" test runs into one to save time
    
Both the scheduler-tests as well as developer-tests in relation to other
test runs do not take that long in comparison. Hence we can save time by
combining both test runs, save one circleCI job while saving the time
for duplicate test run setup.

Mark known slow tests as "heavy" so that it is easier to separate good
and fast tests from slow ones and to prevent premature Makefile-level
timeouts due to recent slowdown observed in CI environments.

Related progress issue: https://progress.opensuse.org/issues/88496

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -251,10 +251,7 @@ jobs:
   unstable:
     <<: *test-template
 
-  developer:
-    <<: *test-fullstack-template
-
-  scheduler:
+  fullstack-unstable:
     <<: *test-fullstack-template
 
   codecov:
@@ -327,9 +324,7 @@ workflows:
       - fullstack: &requires_fullstack
           requires:
             - cache.fullstack
-      - developer:
-          <<: *requires_fullstack
-      - scheduler:
+      - fullstack-unstable:
           <<: *requires_fullstack
       - codecov:
           requires:
@@ -338,8 +333,7 @@ workflows:
             - ui
             - unstable
             - fullstack
-            - developer
-            - scheduler
+            - fullstack-unstable
       - build-docs:
           <<: *requires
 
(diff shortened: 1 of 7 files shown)


Martchus:
Makes sense. Let's see how long the "heavy" tests take now.

okurz:
hm, it's problematic that the former scheduler tests are "unstable" hence also the additional heavy tests are rerun multiple times. Maybe we create a new section "unstable-heavy"?

okurz:
Renamed the combined "scheduler+developer" into "fullstack-unstable" as they have a full stack pulled in as requirement and in comparison to t/full-stack.t are unstable. Then split out an explicit "heavy" set for all the heavy tests that are just slow but not unstable to prevent them to rerun multiple times, especially because they take so long. So in total the same number of circleCI jobs as currently in master but reshuffled.

okurz:
As can be seen in https://app.circleci.com/pipelines/github/os-autoinst/openQA/5554/workflows/c0765fa0-a6a9-4a5b-8a72-de89ccba6694 the runtime of "t" reduced to 15m now and in parallel "heavy" running for "14m". I think that's good parallelization :)

okurz:
for the coverage drop in OpenQA::Worker::WebUIConnection see https://github.com/os-autoinst/openQA/pull/3718 as a minor improvement

Martchus (approved):
The new heavy tests now take roughly the same time than all the regular tests together so I guess that's good.

kalikiana (approved):
I guess the selection of tests is not semantic anymore. But I like idea. And I like the renaming, that makes more sense.

PR #3714, opened 2021-02-09
Stabilize scalability and fullstack test

Martchus:
by ensuring the worker property WEBSOCKET_API_VERSION (which is checked by
the scheduler) is passed before considering worker idle.

---

I thought I might be able to reproduce poo#80800 via the scalability test, see https://progress.opensuse.org/issues/80800#note-12. However, I didn't actually manage to do this. This PR fixes *another* issue I found along the way.

### DIFF
--- a/t/43-scheduling-and-worker-scalability.t
+++ b/t/43-scheduling-and-worker-scalability.t
@@ -1,5 +1,5 @@
 #!/usr/bin/env perl
-# Copyright (C) 2020 SUSE LLC
+# Copyright (C) 2020-2021 SUSE LLC
 #
 # This program is free software; you can redistribute it and/or modify
 # it under the terms of the GNU General Public License as published by
@@ -30,6 +30,7 @@ use OpenQA::Scheduler::Model::Jobs;
 use OpenQA::Utils qw(service_port);
 use OpenQA::Test::Database;
 use OpenQA::Jobs::Constants;
+use OpenQA::Log qw(setup_log);
 use OpenQA::Test::Utils
   qw(mock_service_ports setup_mojo_app_with_default_worker_timeout),
   qw(create_user_for_workers create_webapi setup_share_dir create_websocket_server),
@@ -142,14 +143,19 @@ subtest 'wait for workers to be idle' => sub {
     my @non_idle_workers;
     for my $worker ($workers->all) {
         $worker_ids{$worker->id} = 1;
-        push(@non_idle_workers, $worker->id) if $worker->status ne 'idle';
+        push(@non_idle_workers, $worker->info) if $worker->status ne 'idle';
     }
     ok(!@non_idle_workers, 'all workers idling') or diag explain \@non_idle_workers;
 };
 
 subtest 'assign and run jobs' => sub {
-    my $allocated = OpenQA::Scheduler::Model::Jobs->singleton->schedule;
-    BAIL_OUT('Unable to assign jobs to (idling) workers') unless ref($allocated) eq 'ARRAY' && @$allocated > 0;
+    my $scheduler = OpenQA::Scheduler::Model::Jobs->singleton;
+    my $allocated = $scheduler->schedule;
+    unless (ref($allocated) eq 'ARRAY' && @$allocated > 0) {
+        diag explain 'Allocated: ', $allocated;
+        diag explain 'Scheduled: ', $scheduler->scheduled_jobs;
+        BAIL_OUT('Unable to assign jobs to (idling) workers');
+    }
 
     my $remaining_jobs = $job_count - $worker_count;
     note("Assigned jobs: " . dumper($allocated));
--- a/t/43-scheduling-and-worker-scalability.t
+++ b/t/43-scheduling-and-worker-scalability.t
@@ -26,6 +26,7 @@ use Mojo::Util 'dumper';
 use IPC::Run qw(start);
 use FindBin;
 use lib "$FindBin::Bin/lib", "$FindBin::Bin/../external/os-autoinst-common/lib";
+use OpenQA::Constants qw(WEBSOCKET_API_VERSION);
 use OpenQA::Scheduler::Model::Jobs;
 use OpenQA::Utils qw(service_port);
 use OpenQA::Test::Database;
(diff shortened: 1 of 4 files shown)


t/43-scheduling-and-worker-scalability.t
```diff
@@ -133,23 +135,30 @@ my $polling_tries_workers      = $seconds_to_wait_per_worker / $polling_interval
...
     }
     ok(!@non_idle_workers, 'all workers idling') or diag explain \@non_idle_workers;
 };
 
 subtest 'assign and run jobs' => sub {
-    my $allocated = OpenQA::Scheduler::Model::Jobs->singleton->schedule;
-    BAIL_OUT('Unable to assign jobs to (idling) workers') unless ref($allocated) eq 'ARRAY' && @$allocated > 0;
+    my $scheduler = OpenQA::Scheduler::Model::Jobs->singleton;
+    my $allocated = $scheduler->schedule;
+    unless (ref($allocated) eq 'ARRAY' && @$allocated > 0) {
```
kalikiana, line 157: I guess these lines in the block need `# uncoverable statement`

kalikiana (approved):

okurz (changes requested):
Please check the patch coverage

okurz:
Hm, you added debug logs into production code but just for testing? Temporarily this is fine but I would prefer to not merge changes that introduce testing changes into production code. Maybe it helps to split the methods and have tests based in results returned by these methods?

Martchus:
> Hm, you added debug logs into production code but just for testing?

No, the additional debug log is not enabled in production. It is only logged when the code runs a part of a fullstack test.

> Maybe it helps to split the methods and have tests based in results returned by these methods?

Not sure whether I understand this correctly. Should remove the fullstack test then in its current form and only resort to testing on lower levels? Note that I'm trying to debug the test setup of the fullstack test here, not the code which is being tested (which seems fine, also judging by other tests we already have which are more low-level).

Martchus:
By the way, with the debugging code in-place I could already find some further clues locally. However, I thought that it would be helpful if one could get the clues right away from the CI log instead of having to run the (slow) test locally lots of times.

okurz:
> > Hm, you added debug logs into production code but just for testing?
> 
> No, the additional debug log is not enabled in production. It is only logged when the code runs a part of a fullstack test.

No, I mean I do not think it is good style to have code that would only be executed in tests.

Martchus:
I don't want to enable this logging in production because there we have a significantly higher amount of workers and jobs and therefore the code would log a lot. It is also not very important to have this information in production because one can simply use the web UI and/or database to access it. During the test runs it is quite different: The additional logging is not much (and only goes into the file downloadable under artifacts) and can be very helpful to understand the test setup at a certain point of the test execution which is otherwise not easily possible.

> No, I mean I do not think it is good style to have code that would only be executed in tests.

Tests and test helpers are code which is only executed in tests. So is writing tests a bad style? I suppose you mean it is bad that this code is mixed into the production code. I personally also don't like that. However, it seems the simplest and most pragmatic solution.

okurz:
But again, what if you split methods accordingly so that you can test for the expected output instead of checking log message content? Maybe you want to chat about that later this evening or tomorrow morning?

Martchus:
You're really trying to lure me into refactoring the scheduler code. Ok, let's open this can of worms. The additional checks within the fullstack test would then be using that refactored code but not to test the code itself but to verify that the test setup is as expected. This would definitely be more code than the logging solution but I can do it of course but maybe I'll need this anyways in order to actually fix the problem.

kraih:
Have to agree. Also not a fan of code that is only executed for tests.

t/full-stack.t
```diff
@@ -33,6 +33,7 @@ BEGIN {
 
 use FindBin;
 use lib "$FindBin::Bin/lib", "$FindBin::Bin/../external/os-autoinst-common/lib";
+use Mojo::Base -signatures;
```
kraih, line 36: Since this activates a lexical feature for the whole namespace (similar to `strict`/`warnings`), should this go to the top, near `use Test::Most`?

kraih: Oh wait, there's also `autodie` already below, that's pretty much the same case.

Martchus: The pattern is/was to use `Test::Most` first, then comes `FindBind` and only then other modules (because they might depend on the setup done via `FindBind`). So considering these rules I've included it as soon as possible.

t/full-stack.t
```diff
@@ -113,15 +145,19 @@ my $job_page_url = $driver->get_current_url();
 like(status_text, qr/State: scheduled/, 'test 1 is scheduled');
 ok javascript_console_has_no_warnings_or_errors, 'no javascript warnings or errors after test 1 was scheduled';
 
-sub start_worker_and_schedule {
-    $worker = start_worker(get_connect_args());
-    ok($worker, "Worker started as $worker");
-    schedule_one_job;
+sub assign_jobs ($worker_class = undef) {
+    check_scheduled_job_and_wait_for_free_worker $worker_class // 'qemu_i386';
```
kalikiana, line 149: Since you're using the signature here... would `$worker_class = 'qemu_i386'` work here?

Martchus: No, this way I'd needed to hard-code the string `'qemu_i386'` in two places (because both functions should have that default value).

Martchus:
It looks still good - no failures after 35 local runs. This means the test (and the code under test) is generally quite stable and I couldn't find any other issues so far.

t/full-stack.t
kalikiana: To be clear, what I meant is to use the signature so the `//` in the next line isn't needed:

```suggestion
sub assign_jobs ($worker_class = 'qemu_i386') {
    check_scheduled_job_and_wait_for_free_worker $worker_class;
```

t/full-stack.t
okurz: Not really: https://github.com/os-autoinst/openQA/blob/master/docs/Contributing.asciidoc#code-style-suggestions says `"strict", "warnings" or other modules that provide static checks` and also put Mojo::Base *before* `use lib`

t/full-stack.t
Martchus: Yes, but then I need to do this in both of the functions and would duplicate the string.

t/full-stack.t
Martchus: `-signatures` is neither "strict", "warnings" and also doesn't provide any static checks. (It enables a lexer feature which most code does not need.)

t/full-stack.t
kalikiana: That isn't what I suggested. But feel free to leave it as-is. It's not a big difference.

t/full-stack.t
kraih: Actually it does provide static checks, because signatures require the code to be parsed differently. Different things become syntax errors.

kraih: We should probably just consider `use feature` (and all derivatives) one level below `use warnings`.

okurz (approved):

okurz:
@Martchus do you think this PR might help with https://progress.opensuse.org/issues/65082 so that we can mark the test as stable again without retries?

Martchus:
Seems likely, I'll have a look.

PR #3715, opened 2021-02-10
Fix worker container to allow job executions relying on KVM

ilausuch:
Problem: Container workers are unable to execute jobs because they
do not have permissions on the kvm device
Solution: Include the _openqa-worker user in the same group than /dev/kvm

Problem: qemu-tools is necessary when running a job. Otherwise it shows
the error: no kvm-img/qemu-img found

Notes: Using docker the container has to be runned using --device=/dev/kvm

https://progress.opensuse.org/issues/76978

### DIFF
--- a/container/worker/run_openqa_worker.sh
+++ b/container/worker/run_openqa_worker.sh
@@ -1,4 +1,5 @@
 #!/bin/bash
+# shellcheck disable=SC2012
 set -e
 
 if [[ -z $OPENQA_WORKER_INSTANCE ]]; then
@@ -6,4 +7,12 @@ if [[ -z $OPENQA_WORKER_INSTANCE ]]; then
 fi
 
 /root/qemu/kvm-mknod.sh
+
+if [ -e "/dev/kvm" ] && getent group kvm > /dev/null; then
+  groupmod -g "$(ls -lhn /dev/kvm | cut -d ' ' -f 4)" kvm
+  usermod -a -G kvm _openqa-worker
+else
+  echo "Warning: /dev/kvm doesn't exist. If you want to use KVM, run the container with --device=/dev/kvm"
+fi
+
 su _openqa-worker -c "/usr/share/openqa/script/worker --verbose --instance \"$OPENQA_WORKER_INSTANCE\""
--- a/container/worker/Dockerfile
+++ b/container/worker/Dockerfile
@@ -7,7 +7,7 @@ RUN zypper ar -p 95 -f http://download.opensuse.org/repositories/devel:openQA/op
     zypper ar -p 90 -f http://download.opensuse.org/repositories/devel:openQA:Leap:15.2/openSUSE_Leap_15.2 devel_openQA_Leap && \
     zypper --gpg-auto-import-keys ref && \
     zypper --non-interactive in ca-certificates-mozilla curl gzip && \
-    zypper --non-interactive in openQA-worker qemu-arm qemu-ppc qemu-x86 && \
+    zypper --non-interactive in openQA-worker qemu-arm qemu-ppc qemu-x86 qemu-tools && \
     zypper --non-interactive in kmod && \
     (zypper --non-interactive in qemu-ovmf-x86_64 || true) && \
     (zypper --non-interactive in qemu-uefi-aarch64 || true)


Martchus (commented):
Did you try to just make `/dev/kvm` accessible to the `_openqa-worker` user, e.g. by adding the user to the group `kvm`? That should do the trick in a non containerized environment.

ilausuch:
> Did you try to just make `/dev/kvm` accessible to the `_openqa-worker` user, e.g. by adding the user to the group `kvm`? That should do the trick in a non containerized environment.

Yes, I tried that, but doesn't work
$ groups _openqa-worker
_openqa-worker : nogroup kvm

Martchus:
What are the permissions and ownerships for `/dev/kvm` within and outside of the container? Maybe the group ID inside the container doesn't match the group ID used by the actual system? Not sure what the usual approach for tackling such problems is (apart from the "sudo" hammer).

ilausuch:
> What are the permissions and ownerships for `/dev/kvm` within and outside of the container? Maybe the group ID inside the container doesn't match the group ID used by the actual system? Not sure what the usual approach for tackling such problems is (apart from the "sudo" hammer).

And why don't execute the worker as sudo?

Martchus:
I simply would not recommend to run the worker as root for security reasons. At least that applies to normal setups. I don't know whether it applies to the containerized setup as well. In any case it seems needless to grant a process permission to do everything if just the permission to a specific device would be sufficient.

okurz (changes requested):
I think we already discussed this. The openQA worker is *not* just for qemu and it is *not* just for os-autoinst-distri-opensuse-deps. Why would a Debian developer need to install os-autoinst-distri-opensuse-deps to develop Debian tests?

Martchus:
Of course that raises the question: How is one supposed to make additional dependencies available within that container? I suppose we need to provide a way to install additional packages within the container before starting the tests.

okurz:
Yes, how about an optional ENV variable which takes a list of packages to install if defined

ilausuch:
> I think we already discussed this. The openQA worker is _not_ just for qemu and it is _not_ just for os-autoinst-distri-opensuse-deps. Why would a Debian developer need to install os-autoinst-distri-opensuse-deps to develop Debian tests?


> Yes, how about an optional ENV variable which takes a list of packages to install if defined

Instead of in building time, it can be moved to deploy time.

ilausuch:
Trying to figure out how to solve the kvm problem without sudo mode.
I will remove the installation of deps, this could be installed after deploy the worker

container/worker/run_openqa_worker.sh
```diff
@@ -6,4 +6,9 @@ if [[ -z $OPENQA_WORKER_INSTANCE ]]; then
 fi
 
 /root/qemu/kvm-mknod.sh
+
+kvm_original_group=$(ls -lh /dev/kvm | cut -d " " -f 4)
+groupmod -g $kvm_original_group kvm
```
Martchus, line 11: So my suspicion that the group IDs within and outside of the container don't match was correct. Note that `groupmod -g` expects a group ID but the ls command might return a group name in case there's actually a valid mapping. Hence you need the `-n` flag. I'd also change the quoting and get rid of the extra line:

```suggestion
groupmod -g "$(ls -lhn /dev/kvm | cut -d ' ' -f 4)" kvm
```

Martchus (changes requested):

container/worker/run_openqa_worker.sh
ilausuch: Interesting.. because I tested this code and works. However I am going to include your suggestion to see if works too

Martchus: Of course "it works" … until it breaks when there's coincidentally a mapping for that group ID :-)

ilausuch: Thanks, done and tested locally

Martchus (approved):
I'd called the PR "Fix worker container to allow job executions relying on KVM" but I guess it is ok.

container/worker/run_openqa_worker.sh
```diff
@@ -6,4 +6,8 @@ if [[ -z $OPENQA_WORKER_INSTANCE ]]; then
 fi
 
 /root/qemu/kvm-mknod.sh
+
+groupmod -g "$(ls -lhn /dev/kvm | cut -d ' ' -f 4)" kvm
+usermod -a -G kvm _openqa-worker
```
okurz, line 11: I assume this will fail unless the group "kvm" exists which I would not rely on as already stated multiple times: qemu is the most common but just *one* backend, not the only one that can be used

Martchus: It would make sense to check whether the group `kvm` and `/dev/kvm` both exist before running this code.

ilausuch: Ok thanks

ilausuch: done

container/worker/run_openqa_worker.sh
```diff
@@ -6,4 +6,16 @@ if [[ -z $OPENQA_WORKER_INSTANCE ]]; then
...
 /root/qemu/kvm-mknod.sh
+
+if [ -e "/dev/kvm" ]; then
+  if [ ! $(getent group kvm) ]; then
+    groupmod -g "$(ls -lhn /dev/kvm | cut -d ' ' -f 4)" kvm
+  fi
+
+  usermod -a -G kvm _openqa-worker
+else
+  echo "Warning: /dev/kvm doesn't exist. If you whant to use that, run the container with --device=/dev/kvm"
```
Martchus, line 17:
```suggestion
  echo "Warning: /dev/kvm doesn't exist. If you want to use KVM, run the container with --device=/dev/kvm"
```

container/worker/run_openqa_worker.sh
```diff
@@ -6,4 +6,16 @@ if [[ -z $OPENQA_WORKER_INSTANCE ]]; then
 fi
 
 /root/qemu/kvm-mknod.sh
+
+if [ -e "/dev/kvm" ]; then
+  if [ ! $(getent group kvm) ]; then
```
Martchus, line 11: This should be much simpler since `getent` seems to return `0` only if the group exists:
```suggestion
  if getent group kvm > /dev/null; then
```

My version also negates the condition so `groupmod` is only executed when the group exists. I suppose that's the intention here (because the `groupmod` command would simply fail otherwise).

container/worker/run_openqa_worker.sh
```diff
@@ -6,4 +6,16 @@ if [[ -z $OPENQA_WORKER_INSTANCE ]]; then
...
+    groupmod -g "$(ls -lhn /dev/kvm | cut -d ' ' -f 4)" kvm
+  fi
+
+  usermod -a -G kvm _openqa-worker
```
Martchus, line 15: This line should also only be executed if the group `kvm` exists (see https://github.com/os-autoinst/openQA/pull/3715#discussion_r575214849).

Martchus (changes requested):
See my inline-comments. Note that `groupmod -g …` alters the ID of the existing group but does *not* ensure a new group is created. Alternatively to moving the `usermod` command into the `if` you could also attempt to actually create the group if it doesn't exist.

container/worker/run_openqa_worker.sh
ilausuch: Ups sorry, thanks

container/worker/run_openqa_worker.sh
ilausuch: The problem here is that /dev/kvm could has groups that are not in the system (and with an ls you can se a number) So the objective with groupmod is to create the kvm group using the current group. So only if the group kvm doesn't exist should be executed groupmod

Martchus: But the `groupmod` command does *not* create a new group, e.g.:

```
LANG=C sudo /usr/sbin/groupmod -g 123 foo
groupmod: group 'foo' does not exist
```

As I said, it only alters the group ID of the existing group (in this case to match the group ID of `/dev/kvm`).

(The command to add a group is `groupadd` btw.)

Martchus: Just to be on the same page: Your `groupmod` command fixes a mismatch between the underlying group ID of the group `kvm` in and outside the container. It adjusts the kvm group within the container to have the same group ID as the kvm group outside the container which fixes the permission errors because the device mapped into the container still uses the group ID from outside the container. And again, the command does *not* create a group which is missing within the container (although this might be nice to have as well as mentioned in the other comments).

container/worker/run_openqa_worker.sh
ilausuch: done

container/worker/run_openqa_worker.sh
ilausuch: Thanks for the detailed explanation

ilausuch:
About the checkstyle problem, do you think could be ignored?

kalikiana:
> About the checkstyle problem, do you think could be ignored?

I'm not sure it makes sense to use `find` here... you can use a special comment like the following to ignore this suggestion.

> In container/worker/run_openqa_worker.sh line 12:
>   groupmod -g "$(ls -lhn /dev/kvm | cut -d ' ' -f 4)" kvm
>                  ^--------------^ SC2012: Use find instead of ls to better handle non-alphanumeric filenames.

    # shellcheck disable=SC2012

ilausuch:
> > About the checkstyle problem, do you think could be ignored?

Thanks chris

container/worker/run_openqa_worker.sh
```diff
@@ -1,9 +1,19 @@
 #!/bin/bash
+# shellcheck disable=SC2012
 set -e
+set -x
```
Martchus, line 4: Leftover from debugging?

ilausuch: I thought was interesting to provide more logs. But removed :)

Martchus (approved):

container/worker/run_openqa_worker.sh
```diff
@@ -1,9 +1,18 @@
...
   OPENQA_WORKER_INSTANCE=1
 fi
 
 /root/qemu/kvm-mknod.sh
+
+if [ -e "/dev/kvm" ] && getent group kvm > /dev/null; then
+  groupmod -g "$(ls -lhn /dev/kvm | cut -d ' ' -f 4)" kvm
+  usermod -a -G kvm _openqa-worker
+else
+  echo "Warning: /dev/kvm doesn't exist. If you want to use KVM, run the container with --device=/dev/kvm"
```
okurz, line 15: I don't think this is the right approach as, as stated multiple times, kvm is not a requirement to have a working openQA worker. However, to move forward and as this has no functional impact I will accept it.

okurz (approved):

PR #3716, opened 2021-02-10
Fix/improve reloading worker

Martchus:
* See particular commit messages for details
* See https://progress.opensuse.org/issues/80908#note-8 and https://progress.opensuse.org/issues/80910#note-25

---

I've tested this on openqaworker11 by hot patching its service files and `Worker.pm`. I've also tested the globbing with `systemctl reload` used within the spec file. I've also tested the changes to the Makefile. Everything worked as expected.


okurz (approved):
Very nice :)

PR #3717, opened 2021-02-10
Use signatures for all OpenQA::CacheService::Client methods

kraih:
This conversion was started in #3704 with one method. Converting the rest now too should be very harmless.

### DIFF
--- a/lib/OpenQA/CacheService/Client.pm
+++ b/lib/OpenQA/CacheService/Client.pm
@@ -54,30 +54,22 @@ sub set_port ($self, $port) {
       if $self->ua->can('socket_options');
 }
 
-sub info {
-    my $self = shift;
-
+sub info ($self) {
     my $tx   = $self->_request('get', $self->url('info'));
     my $err  = $self->_error('info', $tx);
     my $data = $tx->res->json // {};
-
     return OpenQA::CacheService::Response::Info->new(data => $data, error => $err);
 }
 
-sub status {
-    my ($self, $request) = @_;
-
+sub status ($self, $request) {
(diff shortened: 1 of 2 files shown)


Martchus (approved):

okurz (approved):

perlpunk (approved):

kraih:
The failed ui test is unrelated, other PRs hang there too it seems.

okurz:
Hm, the tests failed in

```
[11:26:20] t/ui/12-needle-edit.t .................. 17/? Bailout called.  Further testing stopped:  test exceeds runtime limit of '120' seconds
FAILED--Further testing stopped: test exceeds runtime limit of '120' seconds
```

haven't seen that in before. Maybe related to other recent dependency updates. However we should not simply accept unstable tests. I suggest you either look into that as part of the PR even though for sure it's very unlikely to cause these failures or you create a ticket for that.

kraih:
Running `t/ui/12-needle-edit.t` locally the test output looks clean, and it takes about 34 seconds. Also checked with current master, and it's exactly the same. So this PR has nothing to do with it. But i changed the timeout to 40 seconds to fix the problem.

ilausuch (approved):

okurz:
and now

```
[14:07:17] t/api/14-plugin_obs_rsync.t ........ 11/? Bailout called.  Further testing stopped:  test exceeds runtime limit of '180' seconds
FAILED--Further testing stopped: test exceeds runtime limit of '180' seconds
```

😕

kraih:
Tests passed. 😁

PR #3720, opened 2021-02-12
Dependency cron 2021-02-12

openqabot:


okurz:
So it seems the codecov problems are here to stay. I suggest see try to look for updates to the code of client

perlpunk:
> So it seems the codecov problems are here to stay. I suggest see try to look for updates to the code of client

https://github.com/codecov/codecov-perl/pull/42

okurz:
soon we have :1: on the PR by the whole team :)

PR #3722, opened 2021-02-12
Fix codecov with a temporary hack

kraih:
Work in progress.

Upstream fix is in the works too. https://github.com/codecov/codecov-perl/pull/42


kraih:
This workaround is no longer necessary. Codecov has fixed the API endpoint on their end.

PR #3726, opened 2021-02-15
Fix Mojolicious 9.0 compatibility

kraih:
1. Reserved stash values.
```
Route pattern "/children/:status" contains a reserved stash value at lib/OpenQA/WebAPI.pm line 361.
```
It is forbidden to use reserved stash values as placeholders in routes.

2. New log formatting

Individual log messages are now joined with a whitespace character instead of a newline. Additionally `Mojo::Log::Colored` is incompatible with Mojolicious 9.0 (reported upstream).

3. Missing controllers are now an exception
```
[error] [1Whlu6Jt] Controller "OpenQA::WebAPI::Search" does not exist at extlib/lib/perl5/Mojolicious.pm line 124.
```
One of the routes used `->to('search#search')` as a hack to render the template `search/search.html.ep`.

4. `Mojo::IOLoop::Delay` no longer exists
```
Can't locate object method "delay" via package "Mojo::IOLoop" at t/26-controllerrunning.t line 45.
```
The module is no longer part of Mojolicious, since `Mojo::Promise` exists as a much better alternative.

Progress: https://progress.opensuse.org/issues/88609

### DIFF
--- a/lib/OpenQA/WebAPI/Controller/API/V1/Mm.pm
+++ b/lib/OpenQA/WebAPI/Controller/API/V1/Mm.pm
@@ -52,20 +52,20 @@ JSON block with the list.
 # IMHO it should be replaced with get_children and removed
 sub get_children_status {
     my ($self) = @_;
-    my $status = $self->stash('status');
-    if ($status eq 'running') {
-        $status = OpenQA::Jobs::Constants::RUNNING;
+    my $state = $self->stash('state');
+    if ($state eq 'running') {
+        $state = OpenQA::Jobs::Constants::RUNNING;
     }
-    elsif ($status eq 'scheduled') {
-        $status = OpenQA::Jobs::Constants::SCHEDULED;
+    elsif ($state eq 'scheduled') {
+        $state = OpenQA::Jobs::Constants::SCHEDULED;
     }
     else {
-        $status = OpenQA::Jobs::Constants::DONE;
+        $state = OpenQA::Jobs::Constants::DONE;
     }
     my $jobid = $self->stash('job_id');
 
     my @res = $self->schema->resultset('Jobs')
-      ->search({'parents.parent_job_id' => $jobid, state => $status}, {columns => ['id'], join => 'parents'});
+      ->search({'parents.parent_job_id' => $jobid, state => $state}, {columns => ['id'], join => 'parents'});
     my @res_ids = map { $_->id } @res;
     return $self->render(json => {jobs => \@res_ids}, status => 200);
 }
--- a/.circleci/ci-packages.txt
+++ b/.circleci/ci-packages.txt
@@ -140,7 +140,7 @@ perl-Module-Pluggable-5.2
 perl-Module-Runtime-0.016
 perl-Module-Runtime-Conflicts-0.003
 perl-Mojo-IOLoop-ReadWriteProcess-0.28
-perl-Mojolicious-8.73
+perl-Mojolicious-9.01
 perl-Mojolicious-Plugin-AssetPack-2.10
 perl-Mojolicious-Plugin-OAuth2-1.58
 perl-Mojo-Pg-4.24
--- a/lib/OpenQA/Log.pm
+++ b/lib/OpenQA/Log.pm
@@ -44,9 +44,6 @@ our @EXPORT_OK = qw(
 my %CHANNELS;
 my %LOG_DEFAULTS = (LOG_TO_STANDARD_CHANNEL => 1, CHANNELS => []);
 
-my $log_module = "Mojo::Log";
-eval 'use Mojo::Log::Colored; $log_module = "Mojo::Log::Colored"';
-
 # logging helpers - _log_msg wrappers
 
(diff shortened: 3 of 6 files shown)


Martchus (dismissed):

lib/OpenQA/WebAPI/Controller/API/V1/Mm.pm
```diff
@@ -52,7 +52,7 @@ JSON block with the list.
 # IMHO it should be replaced with get_children and removed
 sub get_children_status {
     my ($self) = @_;
-    my $status = $self->stash('status');
+    my $status = $self->stash('state');
```
okurz, line 55: could you rename the variable name as well to be consistent? And how about the method name and hence the routing target?

kraih: Once i have all the other incompatibilities fixed.

kraih:
Side note, i'm a little concerned that `Mojo::Log::Colored` was used automatically when it was installed, but had no tests whatsoever.

kraih:
I believe all issues are resolved now and the code is compatible with 8.73 and 9.0.

Martchus (approved):

okurz (changes requested):

I can recommend the tool [hub](https://github.com/github/hub) (zypper in rubygem-hub) for easier PR creation. Also, I myself use a script [pr-last](http://github.com/okurz/scripts/tree/master/pr-last) to create a PR with proper description for these simple one-commit PRs:
```bash
$ cat $(which git-pr-last )
#!/bin/sh -e
target="${target:-"$USER"}"
```

And the "unstable" test t/ui/27-plugin_obs_rsync_status_details.t is becoming really annoying. I have the suspicion that another recent dependency change made it even more unreliable. In https://github.com/os-autoinst/openQA/pull/3728 I applied further workarounds with longer timeout and more retries. After you fixed the git commit message please rebase on top.

okurz (changes requested):
I think I have a better idea instead of simply removing color support: Please take a look into https://github.com/os-autoinst/os-autoinst/pull/1403/files for an approach which I think you can apply here as well unless you would prefer to fix Mojo::Log::Colored and wait for that. I would appreciate if you would withhold any breaking submission of Mojolicious 9.0 until we can resolve this.

@Martchus I see that you have approved the change without further comments. Maybe you have overlooked that kraih removed the support for colors. I am not a friend of removing features without further consideration.

kraih:
That really wasn't a feature. It was barely a hack. Since there's no tests we should really first discuss the actual requirements for the feature. That seems rather out of scope for this bug fix. I don't mind doing it, but it should really be done as a separate PR.

kraih:
Probably worth mentioning that `Mojo::Log` already supports journald highlighting natively.

okurz:
> Probably worth mentioning that `Mojo::Log` already supports journald highlighting natively.

sounds good. How would we make use of that?

kraih:
> > Probably worth mentioning that `Mojo::Log` already supports journald highlighting natively.
> 
> sounds good. How would we make use of that?

It's the "short" logging option (since journald has its own timestamps etc.). https://docs.mojolicious.org/Mojo/Log#short

kraih:
Actually, we might just solve the `Mojo::Log` issue upstream and make colour highlighting a core mojo feature too. In addition to the journald format. Then for openQA we could just switch around with environment variables depending on use case (`MOJO_LOG_SHORT=1` or `MOJO_LOG_COLOUR=1`).

Martchus (approved):
I approved it because I don't care about the colors. If the dependency breaks with the latest Mojolicious version I wouldn't try to support it any longer.

kraih:
Yes, i consider this PR finished and will just add some more details to the commit message.

kraih:
@okurz Commit message has been updated, and log coloring will be handled upstream. I'm still interested in the specific use cases we had for that though, so the upstream changes don't end up not addressing them.

kraih:
Colored log messages have been added upstream and should be in Mojolicious 9.01. https://github.com/mojolicious/mojo/compare/363ea5575334...0cf164f56956

okurz:
I appreciate the details in the git commit message but I would prefer if we can make "drop of a feature" mentioned prominently in a git commit message subject line so that we can see it explicitly in the package changelogs as well. So it would be great if you can just move that specific removal part into its own git commit, then we can accept it

Martchus:
I'd like to merge this PR rather soon. Otherwise all other PRs are blocked because the CI is not working.

okurz:
sure, as soon as we do not hide removal of a feature behind "fix ..." :)

kraih:
I'm afraid it will take some time for me to figure out how to cleanly separate the commits.

Martchus:
The commit for 9.01 should work. I was going to push it as well but took the opportunity to test the command from another PR (and it apparently takes forever with all the downloading and installation involved): https://github.com/os-autoinst/openQA/pull/3733/files#r576846271

okurz (approved):

okurz:
package tests are fine, we should not block other problems due to incomplete patch coverage

PR #3727, opened 2021-02-15
Remove invisible WIP logos from logo.svg

dcermak:
The svg of the logo still contained a bunch of WIP logos that were left in there
but were invisible because they were outside of the shown image. This commit
removes them, because they just bloat the file, albeit they are quite
interesting to look at ;-)


Martchus (approved):
Tested locally, still looks good.

okurz:
> By the way, mergify might not merge this automatically because the commit message and PR title contain "WIP".

Yeah I double checked as well that the "WIP" is not the prefix :D

PR #3728, opened 2021-02-15
Fix broken lookup of paths in coverage reports

okurz:
Related progress issue: https://progress.opensuse.org/issues/88594

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -153,7 +153,7 @@ test-api:
 # put unstable tests in unstable_tests.txt and uncomment in circle CI to handle unstables with retries
 .PHONY: test-unstable
 test-unstable:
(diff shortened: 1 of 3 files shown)


okurz:
path lookup seems to be ok again, e.g. following https://codecov.io/gh/os-autoinst/openQA/pull/3728/diff?src=pr&el=tree#diff-dC8xMC1qb2JzLnQ= . But paths like https://codecov.io/gh/os-autoinst/openQA/pull/3728/src/lib/OpenQA.pm complain about "File/folder was not found on base commit. No compare to display.". I guess we need this PR merged first.

kalikiana (approved):
I don't see any errors, although I need to click through the UI to see results rather than (re)loading a specific URL which doesn't seem to reveal the same result. So I assume this should work.

Martchus (approved):

PR #3730, opened 2021-02-15
t: Mark uncoverable statements in 01-test-utilities.t+25-downloader.t

okurz:


t/01-test-utilities.t
```diff
@@ -52,11 +52,12 @@ subtest 'warnings in sub processes are fatal test failures' => sub {
     $test_would_have_failed = 0;
     my $ipc_run_harness = OpenQA::Test::Utils::_setup_sigchld_handler 'test-process-2', start sub {
         OpenQA::Test::Utils::_setup_sub_process 'test-process-2';
+        # uncoverable statement
         note "waiting at most $signal_timeout seconds for SIGTERM (sleep is supposed to be interrupted by SIGTERM)";
-        Devel::Cover::report() if Devel::Cover->can('report');
-        sleep $signal_timeout;                            # uncoverable statement
-        note 'timeout for receiving SIGTERM exceeded';    # uncoverable statement
-        exit -1;                                          # uncoverable statement
+        Devel::Cover::report() if Devel::Cover->can('report');    # uncoverable statement
```
kalikiana, line 57: This seems counter-intuitive. This statement is what gives us coverage up to and including this line, no? We don't do this anywhere else either.

Martchus: But we might not even get there because this subprocess is terminated via SIGTERM as part of the test. Hence note comment about the sleep but of course also `Devel::Cover::report()` might already be interrupted.

okurz: Exactly. It's hard to find the right line where to put a comment to explain that. Would have been nice if in perl we could e.g. create a block, add a comment there to explain and then mark the whole block as uncoverable

kalikiana (approved):

Martchus (approved):

PR #3732, opened 2021-02-16
Stabilize t/ui/27-plugin_obs_rsync_status_details

andrii-suse:
The problem happens when gru is slow performing background task. It is reproducible by injecting sleep(2) into update_obs_builds_text().
Previously the test tried to refresh page in hope to get correct info. But that doesn't work for "obs_builds" column.
The fix performs correct UI action instead.

### DIFF
--- a/t/ui/27-plugin_obs_rsync_status_details.t
+++ b/t/ui/27-plugin_obs_rsync_status_details.t
@@ -125,22 +125,22 @@ my %params = (
 );
 
 sub _wait_helper {
-    my ($element, $test_break) = @_;
+    my ($element, $test_break, $refresh) = @_;
     my $ret;
-    for (0 .. 50) {
-        $ret = $driver->find_element($element)->get_text();
-        return $ret if $test_break->($ret);    # uncoverable statement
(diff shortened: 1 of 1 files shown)


okurz:
Great! Let's see how this works then.

kraih:
This PR should not have been merged. It did not pass the perltidy check, which now delays more important fixes.

okurz:
That is correct. However yesterday, before perl-Mojolicious-9.0 was accepted into openSUSE:Factory multiple PRs - I think including the dependency upgrade one - was failing due to recent more sporadic failures in t/ui/27-plugin_obs_rsync_status_details.t

Martchus:
I guess it doesn't matter whether we merge the chicken or the egg first.

okurz:
"merge a chicken"? Funny picture :D

kraih:
> I guess it doesn't matter whether we merge the chicken or the egg first.

Either way, now we've got [waffles](https://github.com/mojolicious/mojo/blob/v9.0/Changes#L3). 😆

PR #3734, opened 2021-02-16
Improve help of openqa-clone-job script

Martchus:
* Unify style
* Add comment explaining what examples from SYNOPSIS actually do
* Add example for cloning a parallel cluster
* Don't have two SYNOPSIS sections and instead add a DESCRIPTION section
* Show lengthy list of options at the end
* Clarify what's meant with "parent" and "child" because for especially for
  parallel jobs that might not be obvious
* Add note that cloning directly chained dependencies does not work
* Tweak certain details


script/openqa-clone-job
```diff
@@ -21,111 +21,114 @@
 
 =head1 openqa-clone-job
 
-openqa-clone-job - clone job from local or remote openQA instances
+Clones a job from the local or another openQA instance. Downloads all assets associated
```
okurz, line 24: what was wrong with "remote"?

script/openqa-clone-job
```diff
@@ -21,111 +21,114 @@
...
+with the job (unless --skip-download is specified). Optionally settings can be modified.
 
 =head1 SYNOPSIS
 
   openqa-clone-job [OPTIONS] JOBREF [KEY=[VALUE] ...]
 
+  # clones job 42 (and parents) from "openqa.opensuse.org" to the local openQA instance
```
okurz, line 31: I wonder if this is a common format for SYNOPSIS in perl applications. Have you seen this elsewhere?

Also I would write "(and any existing parents)"

script/openqa-clone-job
```diff
@@ -21,111 +21,114 @@
...
+=head1 DESCRIPTION
+
+Call with either a full URL pointing to a test job to clone from or one of
+both parameters C<--from> or C<--within-instance>. The job ID can be specified
+as part of the URL or as its own parameter.
+
+Any parent jobs (chained or parallel) are also cloned unless C<--skip-deps> or
+C<--skip-chained-deps> is specified. If C<--skip-chained-deps> is specified
+published assets generated by parent jobs are downloaded to be directly used
+instead of generated. Keep in mind that by default any additionally specified job settinngns
```
okurz, line 55: s/settinngns/settings/

and it looks like you haven't consistently wrapped at 80 characters. I suggest you find a way how to auto-wrap in your editor of choice as we do not have automatic checks for that. As this text is shown to the user we should be consistent there

script/openqa-clone-job
```diff
@@ -21,111 +21,114 @@
...
+Increases verbosity.
 
 =item B<--help, -h>
 
-print help
+Prints help.
 
 =back
 
-=head1 SYNOPSIS
```
okurz, line 103: Do you know why there had been two SYNOPSIS sections in before? Maybe that actually makes sense in Perl POD?

script/openqa-clone-job
```diff
@@ -141,7 +144,7 @@ my $jobid;
 
 sub usage($) {
     my $r = shift;
-    eval { require Pod::Usage; Pod::Usage::pod2usage($r); };
+    eval { require Pod::Usage; Pod::Usage::pod2usage(-exitval => $r, -verbose => 99) };
```
okurz, line 147: interesting. What is the meaning of this "verbose => 99"?

okurz (changes requested):

script/openqa-clone-job
perlpunk: No, two sections with the same name don't really make sense

script/openqa-clone-job
Martchus: This was actually "Clone job from another instance." from below and when merging redundant sentences it got mixed up. I can keep using "remote" here.

script/openqa-clone-job
Martchus: > Have you seen this elsewhere?

I don't know. I just find it annoying to have "nice" examples but one still has to figure out what these examples actually do.

script/openqa-clone-job
Martchus: I can make the wrapping consistent within the code but I'd like to note that the Perl module we use to display this as `--help`-text applies its own wrapping so this is not actually shown to the user.

script/openqa-clone-job
Martchus: It causes `Pod::Usage::pod2usage` to display all the sections. Maybe the duplication of SYNOPSIS was previously done to workaround the default behavior (to answer the question from your other comment).

script/openqa-clone-job
Martchus: Even if Perl people would disagree: I don't think having the same heading twice at the same level is generally considered a good way to structure a document.

script/openqa-clone-job
perlpunk: It looks fine to me, I don't see a problem

script/openqa-clone-job
okurz: ok, I would prefer that. Because "local or remote" reads better than "local or another".

script/openqa-clone-job
okurz: ok, thanks for the confirmation. Keep it then.

script/openqa-clone-job
okurz: I know but there can be weird practices for good reasons :) Ok, keep it then

script/openqa-clone-job
okurz: I see. Could be. all good then.

okurz (approved):

ilausuch (approved):
This is great!

PR #3737, opened 2021-02-17
[WIP] Add a script to deploy webui and worker container locally easily

ilausuch:
Problem: To create a webui and worker deploy some steps have to be
followed.
Solution: This script executes these steps

Additionally:
- It checks that all is correct in each step
- A previous step could be redeployed
- It guides to the user with the manual steps

https://progress.opensuse.org/issues/76978

### DIFF
--- a/container/deploy_openqa_containers.sh
+++ b/container/deploy_openqa_containers.sh
@@ -1,5 +1,5 @@
 #!/bin/bash
-#set -x
+set -e
 
 FORCE=0
 PREFIX="openqa"
@@ -12,14 +12,12 @@ WORKER_IMAGE="registry.opensuse.org/devel/openqa/containers15.2/openqa_worker:la
 function wait_for_container_ready {
   container=$1
 
-  count=30
-  while [ $count -gt 0 ]; do
+  for ((count=0; count<30; count++))
+  do
     sleep 1
     if docker inspect openqa2_db | grep '"Status": "running"' >/dev/null; then
       echo "OK - Container $container started correcty"
       break
-    else
-      count=$((count-1))
     fi
   done
 }
@@ -29,22 +27,20 @@ function wait_for_container_appropiate_logs {
   message=$2
 
   if [ -n "$3" ]; then
-    count=$(($3))
+    countMax=$(($3))
   else  
-    count=30
+    countMax=30
   fi
 
-  while [ $count -gt 0 ]; do
+  for ((count=0; count<countMax; count++))
+  do
     if docker logs "$container" 2>&1 | grep "$message" >/dev/null; then
       echo "OK - Container $container stand-up correctly - found in logs $message"
       break
-    else
-      count=$((count-1))
-      sleep 1
     fi    
   done
 
-  if [ $count -eq 0 ]; then
+  if [[ $count -eq countMax ]]; then
     echo "ERROR - Sorry the container $container didn't stand-up correctly"
     exit 2
   fi
@@ -59,6 +55,8 @@ function show_usage {
   echo " - worker, create the worker container If uses --force the current configuration and worker container will be removed and ecreated"
   echo "Options:"
   echo "-f|--force, destroys and recreate the previous data or container associated to the action"
+  echo "--key, api key. Required deploying the worker"
+  echo "--secret, api secret. Required deploying the worker"
   echo "--prefix [network_name], set a prefix network and container creation (the default is openqa)"
   echo "--port, changes the port for the UI (default 80)"
   echo "--webui-image, to select what image will be used for the webui container (the default  is registry.opensuse.org/devel/openqa/containers15.2/openqa_webui:latest)"
@@ -68,12 +66,12 @@ function show_usage {
 }
 
 function prepare {
-  [ $FORCE ] && rm -rf data 2>/dev/null
+  [[ $FORCE -eq 1 ]] && rm -rf data 2>/dev/null
   mkdir -p data/factory/{iso,hdd,other,tmp} data/{testresults,tests,conf} data/certs/{ssl.crt,ssl.key}
   chmod a+w data/testresults
-  
-  [ $FORCE ] && docker network rm "$NETWORK" 2>/dev/null
-  docker network create "$NETWORK" 2>/dev/null
+
+  [[ $FORCE -eq 1 ]] && docker network inspect "$NETWORK" >/dev/null 2>&1 && docker network rm "$NETWORK" >/dev/null
+  docker network create "$NETWORK" >/dev/null || exit 1
   
   echo "Next step (db), creation of the data base"
 }
@@ -81,8 +79,8 @@ function prepare {
 function db {
   CONT_NAME="${PREFIX}_db"
 
-  [ $FORCE ] && docker rm -f "$CONT_NAME" 2>/dev/null
-  docker run -d --network "$NETWORK" -e POSTGRES_PASSWORD=openqa -e POSTGRES_USER=openqa -e POSTGRES_DB=openqa --net-alias=db --name "$CONT_NAME" postgres
+  [[ $FORCE -eq 1 ]] && set +e && docker rm -f "$CONT_NAME" >/dev/null 2>&1; set -e
+  docker run -d --network "$NETWORK" -e POSTGRES_PASSWORD=openqa -e POSTGRES_USER=openqa -e POSTGRES_DB=openqa --net-alias=db --name "$CONT_NAME" postgres >/dev/null
 
   wait_for_container_ready "$CONT_NAME"
   wait_for_container_appropiate_logs "$CONT_NAME" "database system is ready to accept connections"
@@ -91,26 +89,28 @@ function db {
 }
 
 function webui {
-  [ ! -d data ] && echo "Please execute first prepare command" && exit 1
+  [[ ! -d data ]] && echo "Please execute first prepare command" && exit 1
 
   CONT_NAME="${PREFIX}_webui"
   
   pushd .
   cd data/conf || exit 1
-  wget https://raw.githubusercontent.com/os-autoinst/openQA/master/container/webui/conf/openqa.ini || exit 1
+  wget https://raw.githubusercontent.com/os-autoinst/openQA/master/container/webui/conf/openqa.ini >/dev/null
   cp openqa.ini /tmp/
   sed 's/method = OpenID/method = Fake/' </tmp/openqa.ini >openqa.ini
-  wget https://raw.githubusercontent.com/os-autoinst/openQA/master/container/webui/conf/database.ini || exit 1
-  wget https://raw.githubusercontent.com/os-autoinst/openQA/master/container/openqa_data/data.template/conf/client.conf || exit 1
+  wget https://raw.githubusercontent.com/os-autoinst/openQA/master/container/webui/conf/database.ini  >/dev/null
+  wget https://raw.githubusercontent.com/os-autoinst/openQA/master/container/openqa_data/data.template/conf/client.conf >/dev/null
   popd || exit 1
 
-  openssl req -newkey rsa:4096 -x509 -sha256 -days 365 -nodes -subj '/CN=www.mydom.com/O=My Company Name LTD./C=DE' -out data/certs/ssl.crt/server.crt -keyout data/certs/ssl.key/server.key
+  openssl req -newkey rsa:4096 -x509 -sha256 -days 365 -nodes -subj '/CN=www.mydom.com/O=My Company Name LTD./C=DE' \
+    -out data/certs/ssl.crt/server.crt -keyout data/certs/ssl.key/server.key >/dev/null
+
   cp data/certs/ssl.crt/server.crt data/certs/ssl.crt/ca.crt
 
   volumes="-v $(pwd)/data:/data"
   certificates="-v $(pwd)/data/certs/ssl.crt:/etc/apache2/ssl.crt -v $(pwd)/data/certs/ssl.key:/etc/apache2/ssl.key"
 
-  [ $FORCE ] && docker rm -f "$CONT_NAME" 2>/dev/null
+  [[ $FORCE -eq 1 ]] && set +e && docker rm -f "$CONT_NAME" 2>/dev/null 2>&1; set -e
   docker run -d --network "$NETWORK" $volumes $certificates -p $PORT:80 --net-alias=openqa_webui --name "$CONT_NAME" "$WEBUI_IMAGE" || exit 1
   
   wait_for_container_ready "$CONT_NAME"
@@ -124,20 +124,20 @@ function webui {
 }
 
 function worker {
-  [ -z "$KEY" ] || [ -z "$SECRET" ] && echo "--key and --secret are required" && exit 1
+  [[ -z $KEY ]] || [[ -z $SECRET ]] && echo "--key and --secret are required" && exit 1
 
-  [ ! -d data ] && echo "Please execute first prepare command" && exit 1
+  [[ ! -d data ]] && [[ ! -d data/conf ]] && echo "Please execute first prepare command" && exit 1
 
   CONT_NAME="${PREFIX}_worker"
 
   pushd .
-  cd data/conf || exit 1
-  wget https://raw.githubusercontent.com/os-autoinst/openQA/master/container/openqa_data/data.template/conf/workers.ini >/dev/null 2>/dev/null || exit 1
+  cd data/conf
+  wget https://raw.githubusercontent.com/os-autoinst/openQA/master/container/openqa_data/data.template/conf/workers.ini >/dev/null 2>/dev/null
   popd || exit 1
 
   echo -e "[openqa_webui]\nkey = $KEY\nsecret = $SECRET" > data/conf/client.conf
 
-  [ $FORCE ] && docker rm -f "$CONT_NAME" >/dev/null 2>/dev/null
+  [[ $FORCE -eq 1 ]] && set +e && docker rm -f "$CONT_NAME" >/dev/null 2>&1 && set -e
   docker run -d --network "$NETWORK" -v "$(pwd)/data:/data" --device=/dev/kvm --privileged --name "$CONT_NAME" "$WORKER_IMAGE" >/dev/null
 
   wait_for_container_ready "$CONT_NAME"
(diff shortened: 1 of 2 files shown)


container/deployer.sh
```diff
@@ -0,0 +1,217 @@
+#!/bin/bash
+#set -x
```
Martchus, line 2: Using `set -e` would actually make sense.

container/deployer.sh
```diff
@@ -0,0 +1,217 @@
...
+PREFIX="openqa"
+NETWORK="${PREFIX}_net"
+PORT="80"
+ACTION=""
+
+function wait_for_container_ready {
+  container=$1
+
+  count=30
+  while [ $count -gt 0 ]; do
```
Martchus, line 16: There are for loops and local variables in bash.

Martchus (commented):
I haven't looked very closely what this script does but found some general coding style issues. When using Bash, try to use `[[` if possible (instead of `[`) to avoid forking `test`.

It would be good if the help of the script would contain an overall description what is achieved by executing it. (Avoid generic words like "deploy".)

container/deployer.sh
ilausuch: done

container/deployer.sh
ilausuch: done

container/deployer.sh
```diff
@@ -0,0 +1,217 @@
...
+  CONT_NAME="${PREFIX}_webui"
+  IMAGE="registry.opensuse.org/devel/openqa/containers15.2/openqa_webui:latest"
+  
+  pushd .
+  cd data/conf || exit 1
+  wget https://raw.githubusercontent.com/os-autoinst/openQA/master/container/webui/conf/openqa.ini || exit 1
+  cp openqa.ini /tmp/
+  sed 's/method = OpenID/method = Fake/' </tmp/openqa.ini >openqa.ini
+  wget https://raw.githubusercontent.com/os-autoinst/openQA/master/container/webui/conf/database.ini || exit 1
+  wget https://raw.githubusercontent.com/os-autoinst/openQA/master/container/openqa_data/data.template/conf/client.conf || exit 1
```
okurz: Now this gets tricky. So you want to mix the upstream container images with the current config files in git which might not be compatible.

okurz (changes requested):
Overall this seems like you are duplicating the effort from other bootstrap scripts that we already have or even what container management solutions provide including the already present container/webui/docker-compose.yaml which is still untested. I suggest to look into these existing solutions rather than coming up with yet another wrapper. One more problem I have in reviewing is that you just provide a script out of context which is not called anywhere so I would not know how this is integrated. Please think about https://en.wikipedia.org/wiki/Test-driven_development or at least https://tom.preston-werner.com/2010/08/23/readme-driven-development.html

container/deployer.sh
```diff
@@ -0,0 +1,217 @@
...
+  echo "-f|--force, destroys and recreate the previous data or container associated to the action"
+  echo "--prefix [network_name], set a prefix network and container creation (the default is openqa)"
+  echo "Notes:"
+  echo "  For initial deployments execute in order the actions: prepare, db, webui, worker"
+}
+
+function prepare {
+  [ $FORCE ] && rm -rf data 2>/dev/null
+  mkdir -p data/factory/{iso,hdd,other,tmp} data/{testresults,tests,conf} data/certs/{ssl.crt,ssl.key}
+  chmod a+w data/testresults
```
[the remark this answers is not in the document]
ilausuch: Yes, this is estrange, but no, these are directories and the certificates has to be inside

container/deployer.sh
```diff
@@ -0,0 +1,217 @@
...
+  wait_for_container_appropiate_logs "$CONT_NAME" "database system is ready to accept connections"
+
+  echo "Next step (webui), stand-up the webui"
+}
+
+function webui {
+  [ ! -d data ] && echo "Please execute first prepare command" && exit 1
+
+  CONT_NAME="${PREFIX}_webui"
+  IMAGE="registry.opensuse.org/devel/openqa/containers15.2/openqa_webui:latest"
```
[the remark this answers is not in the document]
ilausuch: Good point

container/deploy_openqa_containers.sh
```diff
@@ -0,0 +1,228 @@
...
+FORCE=0
+PREFIX="openqa"
+NETWORK="${PREFIX}_net"
+PORT="80"
+ACTION=""
+WEBUI_IMAGE="registry.opensuse.org/devel/openqa/containers15.2/openqa_webui:latest"
+WORKER_IMAGE="registry.opensuse.org/devel/openqa/containers15.2/openqa_worker:latest"
+
+function wait_for_container_ready {
+  container=$1
```
Martchus, line 13: Bash still supports local variables.

container/deploy_openqa_containers.sh
```diff
@@ -0,0 +1,228 @@
...
+
+  for ((count=0; count<30; count++))
+  do
```
Martchus, line 16: I'd prefer this style:
```suggestion
  for ((count=0; count<30; count++)); do
```

container/deploy_openqa_containers.sh
```diff
@@ -0,0 +1,228 @@
...
+    sleep 1
+    if docker inspect openqa2_db | grep '"Status": "running"' >/dev/null; then
+      echo "OK - Container $container started correcty"
```
Martchus, line 19:
```suggestion
      echo "OK - Container $container started correctly"
```

container/deploy_openqa_containers.sh
```diff
@@ -0,0 +1,228 @@
...
+
+function prepare {
+  [[ $FORCE -eq 1 ]] && rm -rf data 2>/dev/null
+  mkdir -p data/factory/{iso,hdd,other,tmp} data/{testresults,tests,conf} data/certs/{ssl.crt,ssl.key}
+  chmod a+w data/testresults
+
+  [[ $FORCE -eq 1 ]] && docker network inspect "$NETWORK" >/dev/null 2>&1 && docker network rm "$NETWORK" >/dev/null
+  docker network create "$NETWORK" >/dev/null || exit 1
+  
+  echo "Next step (db), creation of the data base"
```
Martchus, line 76:
```suggestion
  echo "Next step (db), creation of the database"
```

container/deploy_openqa_containers.sh
```diff
@@ -0,0 +1,228 @@
...
+}
+
+function webui {
+  [[ ! -d data ]] && echo "Please execute first prepare command" && exit 1
+
+  CONT_NAME="${PREFIX}_webui"
+  
+  pushd .
+  cd data/conf || exit 1
+  wget https://raw.githubusercontent.com/os-autoinst/openQA/master/container/webui/conf/openqa.ini >/dev/null
```
Martchus, line 98: I'd avoid repeating the same path again and again:
```suggestion
  local openqa_url_prefix=https://raw.githubusercontent.com/os-autoinst/openQA/master
  wget "$openqa_url_prefix/container/webui/conf/openqa.ini" >/dev/null
```

container/deploy_openqa_containers.sh
```diff
@@ -0,0 +1,228 @@
...
+  echo "All ready. Enjoy!"
+  echo "You have to provide your own tests or the upstream tests and neeles and store them in the directory data/tests"
+  echo "e.g. of usage: docker exec -ti ${PREFIX}_webui /usr/share/openqa/script/clone_job.pl --apikey $KEY --apisecret $SECRET <upstream_job>"
+}
+
+function clean {
+  set +e
+  docker rm -f "${PREFIX}_worker"
+  docker rm -f "${PREFIX}_webui"
+  docker rm -f "${PREFIX}_db"
```
Martchus, line 157: I suppose this should be possible:
```suggestion
  docker rm -f "${PREFIX}_worker" "${PREFIX}_webui" "${PREFIX}_db"
```
Otherwise you could use a loop.

container/deploy_openqa_containers.sh
```diff
@@ -0,0 +1,228 @@
...
+    exit 2
+  fi
+}
+
+function show_usage {
+  echo "Usage: $0 [options] <action>"
+  echo "Actions:"
+  echo " - prepare, prepare the environment. If uses --force the current directories will be removed and recreated"
+  echo " - db, create the database container. If uses --force the current db container will be removed and recreated"
+  echo " - webui, create the webui container. If uses --force the current configuration and webui container will be removed and ecreated"
```
Martchus, line 54:
```suggestion
  echo " - webui, create the webui container. If uses --force the current configuration and webui container will be removed and recreated"
```

container/deploy_openqa_containers.sh
```diff
@@ -0,0 +1,228 @@
...
+  echo " - worker, create the worker container If uses --force the current configuration and worker container will be removed and ecreated"
```
Martchus, line 55:
```suggestion
  echo " - worker, create the worker container If uses --force the current configuration and worker container will be removed and recreated"
```

container/deploy_openqa_containers.sh
```diff
@@ -0,0 +1,228 @@
...
+  echo "Options:"
+  echo "-f|--force, destroys and recreate the previous data or container associated to the action"
```
Martchus, line 57:
```suggestion
  echo "-f|--force, destroys and recreates the previous data or container associated to the action"
```

container/deploy_openqa_containers.sh
```diff
@@ -0,0 +1,228 @@
...
+  echo "--key, api key. Required deploying the worker"
```
Martchus, line 58:
```suggestion
  echo "--key, api key. Required to deploy the worker"
```

(same counts for the next line)

container/deploy_openqa_containers.sh
```diff
@@ -0,0 +1,228 @@
...
+  echo "--secret, api secret. Required deploying the worker"
+  echo "--prefix [network_name], set a prefix network and container creation (the default is openqa)"
+  echo "--port, changes the port for the UI (default 80)"
+  echo "--webui-image, to select what image will be used for the webui container (the default  is registry.opensuse.org/devel/openqa/containers15.2/openqa_webui:latest)"
```
Martchus, line 62:
```suggestion
  echo "--webui-image, to select what image will be used for the web UI container (the default is registry.opensuse.org/devel/openqa/containers15.2/openqa_webui:latest)"
```

container/deploy_openqa_containers.sh
```diff
@@ -0,0 +1,228 @@
...
+  echo "--worker-image, to select what image will be used for the worker container (the default is registry.opensuse.org/devel/openqa/containers15.2/openqa_worker:latest)"
+  echo "Notes:"
+  echo "  For initial deployments execute in order the actions: prepare, db, webui, worker"
```
Martchus, line 65:
```suggestion
  echo "  For initial deployments execute the actions in the following order: prepare, db, webui, worker"
```

container/deploy_openqa_containers.sh
```diff
@@ -0,0 +1,228 @@
...
+  done
+
+  if [[ $count -eq countMax ]]; then
+    echo "ERROR - Sorry the container $container didn't stand-up correctly"
+    exit 2
+  fi
+}
+
+function show_usage {
+  echo "Usage: $0 [options] <action>"
```
Martchus, line 50: It likely makes sense to use just one `echo` and a multi-line `""`.

container/deploy_openqa_containers.sh
```diff
@@ -0,0 +1,228 @@
...
+  echo "Actions:"
+  echo " - prepare, prepare the environment. If uses --force the current directories will be removed and recreated"
```
Martchus, line 52:
```suggestion
  echo " - prepare, prepare the environment. If using --force the current directories will be removed and recreated"
```

(same counts for the next line)

container/deploy_openqa_containers.sh
```diff
@@ -0,0 +1,228 @@
...
+    fi    
+  done
+
+  if [[ $count -eq countMax ]]; then
+    echo "ERROR - Sorry the container $container didn't stand-up correctly"
+    exit 2
+  fi
+}
+
+function show_usage {
```
Martchus, line 49: The usage is detailed but leaves some important questions: What *does* this script and what are the pre-conditions to run it? Do I need to run it as root? Is it supposed to run within a container? What is meant with "deployment" (which is a rather generic term)? Where is it deployed to? Maybe the script's name could also be changed to be more specific.

Martchus (changes requested):
Please read this time all the inline-comments. Note that the inline-comments count for every occurrence of the same problem.

kalikiana:
There is three user-supplied aspects here that I see
- certificates
- custom authentication
- creating folders in the test results directory

I don't see why we want these bits
- There's a lot of waiting/checking/logging 🤔
- There's some CLI with `--key`, `--port` and such, imho that can be configured with docker/podman commands
- There's the *clean* function with things like `docker rm -f "${PREFIX}_worker"` which arguably is the same as `docker compose down -v`

Shouldn't we rely on `docker-compose up/down` for these things rather than creating a custom CLI?

container/deployer.sh
ilausuch: Maybe a solution could be to deliver this script with the configuration files. Therefore we ensure that won't be a discrepancy between the script and the config

container/deployer.sh
```diff
@@ -0,0 +1,217 @@
...
+  pushd .
+  cd data/conf || exit 1
+  wget https://raw.githubusercontent.com/os-autoinst/openQA/master/container/webui/conf/openqa.ini || exit 1
+  cp openqa.ini /tmp/
+  sed 's/method = OpenID/method = Fake/' </tmp/openqa.ini >openqa.ini
+  wget https://raw.githubusercontent.com/os-autoinst/openQA/master/container/webui/conf/database.ini || exit 1
+  wget https://raw.githubusercontent.com/os-autoinst/openQA/master/container/openqa_data/data.template/conf/client.conf || exit 1
+  popd || exit 1
+
+  openssl req -newkey rsa:4096 -x509 -sha256 -days 365 -nodes -subj '/CN=www.mydom.com/O=My Company Name LTD./C=DE' -out data/certs/ssl.crt/server.crt -keyout data/certs/ssl.key/server.key
```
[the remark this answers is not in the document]
ilausuch: I have a question about that. RPM is not linux distro interdependent. Do you think is a good idea to put this on the script? Other option is to include the certificates on the container image (webUI) but I am checking that by default this package perl-Mojolicous is not listed (I checked this entering to a running container and trying to install with zypper)

container/deployer.sh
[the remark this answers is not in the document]
ilausuch: See that https://github.com/os-autoinst/openQA/blob/feca60d86048c2c80a8595695cc57584356cb083/container/webui/openqa-ssl.conf#L18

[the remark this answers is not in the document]
okurz: ok, I see. Thanks.

container/deployer.sh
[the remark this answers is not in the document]
okurz: > I have a question about that. RPM is not linux distro interdependent.

I assume you mean "independent" and that non-rpm Linux distributions would not have this RPM package? Yes, that is true but I mentioned the above not that you need to play with rpm within the script but only for you to execute in your development environment to see which files are available.

> Do you think is a good idea to put this on the script? Other option is to include the certificates on the container image (webUI) but I am checking that by default this package perl-Mojolicous is not listed (I checked this entering to a running container and trying to install with zypper)

I don't know what you mean with "not listed". But the perl package https://metacpan.org/pod/Mojolicious is a hard dependency of openQA so this is why we can rely upon it in all environments where we want to run openQA. For containers of course would mean **within** the container.

Again as in https://github.com/os-autoinst/openQA/pull/3737#pullrequestreview-596157558 I don't know where and how you intend to run this script so *please* address this first before you try to fix details in a shell script.

@Martchus and @kalikiana just addressed the very same issue: We do not yet understand the basic concept that you have in mind so **please do not** try to beautify your shell script before we could even get that sorted.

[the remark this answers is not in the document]
Martchus: The last paragraph is important. Maybe we should sort this out in a meeting.

ilausuch:
> There is three user-supplied aspects here that I see
> 
>     * certificates
In the scripts, the certificates are generated when are needed. But if you think it's better to give the opportunity to the user to provide them, it could be done. e.g. with the flag --cert-key, --cert-crt, --cert-ca
> 
>     * custom authentication
By default it is used the Fake authentification. If you run this a local (localhost) the openID doesn't work. The same than the previous task, it could be configured. e.g. --auth
> 
>     * creating folders in the test results directory
I think the content in testresults directory is created by the worker. If you mean in the tests then should be independent to this script to maintain the independence to SUSE. Is the same that installing  
os-autoinst-distri-opensuse-deps. 

> 
> I don't see why we want these bits
> 
>     * There's a lot of waiting/checking/logging thinking
About waiting and checking, why not? It is better if we can provide a check about the action. About login, yes, could be hidden.

>     * There's some CLI with `--key`, `--port` and such, imho that can be configured with docker/podman commands
Sorry, I don't see your point

> 
>     * There's the _clean_ function with things like `docker rm -f "${PREFIX}_worker"` which arguably is the same as `docker compose down -v`
> 
> 
> Shouldn't we rely on `docker-compose up/down` for these things rather than creating a custom CLI?
> To illustrate what I _thought_ we wanted you could look at this [https://github.com/docker/awesome-compose/tree/master/gitea-postgres](https://github.com/docker/awesome-compose/tree/master/gitea-postgres?rgh-link-date=2021-02-23T11%3A11%3A50Z)

I agree with you that will be better, but the problem with docker-compose up/down is that there are some steps in the middle between the webUI and the worker that are manual. Obtain the key/secret. In this case we should have two docker-compose one for db and web UI and other for the worker. I am not see if this is your idea. But yes, let's try this to see how far I can get.

ilausuch:
following your comments...
We could take this https://github.com/os-autoinst/openQA/blob/master/container/webui/docker-compose.yaml and there are different options
- Create an other docker-compose to deploy the worker
- Create a unified docker-compose (if is possible because the key/secret required in the worker initialization)

In both cases, openqa_data should be fixed with some missing directories https://github.com/os-autoinst/openQA/tree/master/container/openqa_data

kalikiana:
> We could take this https://github.com/os-autoinst/openQA/blob/master/container/webui/docker-compose.yaml and there are different options
> 
> * Create an other docker-compose to deploy the worker
> * Create a unified docker-compose (if is possible because the key/secret required in the worker initialization)

I would guess **as a user** that I could pull/run the worker image. And maybe I have to 1. set environment variables to connect with the host/credentials 2. network connect.

Or maybe **as a user** I have to modify the compose yaml and add the workers I want.

> We don't know if all parts are running correctly without checking the logs (in kubernetes is possible using Liveness tests)

Do we need to check that here? I assume that's done in CI in static checks, openQA tests, etc and we extend those as needed. The examples in *awesome-compose* don't have such checks either. Of course if you wanted to write this in the tutorial you can say, "check this to verify..." and explain what services are running.

Note that I picked the awesome examples as a baseline. Since those are probably considered nice examples in general. I couldn't find one with a variable amount of servers, though.

PR #3740, opened 2021-02-18
Fix remove link configuration for "Untracked" groups in /admin/assets

ilausuch:
Problem: Clicking on "Untracked" on /admin/assets yields "Not Found" page
Solution: Remove the clickable link at all

https://progress.opensuse.org/issues/88745


ilausuch:
I based that over the name of the group, but groupId=0 could be used too

Martchus (approved):

okurz (approved):

PR #3741, opened 2021-02-18
Add tests for the scheduler's starvation protection

Martchus:
… and a few other tests for certain error cases which were not covered so far.

The motivation is https://progress.opensuse.org/issues/88459 although these tests actually show that starvation should be handled correctly.


okurz (approved):
I see the coverage increased so this seems to be a useful addition also in this regard

Amrysliu (approved):

Martchus:
Codecov rejects the upload after re-triggering it due to another networking issue:

```
404 Not Found
{"detail":"Build has already finished, uploads rejected."}
```

When trying to access the previous report I've ran into `upstream request timeout` but it should be good.

PR #3742, opened 2021-02-18
Make offset for scheduler's starvation protection configurable

Martchus:
See: https://progress.opensuse.org/issues/88459#change-383696

### DIFF
--- a/lib/OpenQA/Scheduler.pm
+++ b/lib/OpenQA/Scheduler.pm
@@ -1,4 +1,4 @@
-# Copyright (C) 2015-2020 SUSE LLC
+# Copyright (C) 2015-2021 SUSE LLC
 #
 # This program is free software; you can redistribute it and/or modify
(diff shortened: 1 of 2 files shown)


okurz (approved):
How about patching o3 with a number higher than 1 and see if that helps the problems we observed there?

kalikiana (commented):
Can we check the env variable in a unit test? Although I guess the implementation is basically the same.

Might also be good to document, though maybe like other scheduler variables we consider it semi-private (i.e. OPENQA_SCHEDULER_MAX_JOB_ALLOCATION or OPENQA_SCHEDULER_SCHEDULE_TICK_MS are not documented either)

Martchus:
When https://github.com/os-autoinst/openQA/pull/3741 is merged I can extend the tests to cover this feature as well.

Amrysliu (approved):

PR #3743, opened 2021-02-19
codecov.yml: Introduce specific coverage thresholds

okurz:


kalikiana:
Would you mind adding a description? What is this addressing? Is this related to a ticket or PR? Maybe we discussed it before, but reading the YAML I don't know what to look for

okurz:
let me make that a draft. I was experimenting what is actually possible before deciding which portions to handle in what specific way. I don't see a codecov report so that's the problem for now

okurz:
Still the problem that codecov does not provide a report. Anyone has an idea how we can ensure that we have a complete report like before or also separate ones per path?

Martchus:
I assume the idea here is to have two distinct coverage statistics for tests code and the actual code.

I can read the codecov documentation myself next week. Maybe one needs to keep the default?

okurz:
Trying in https://github.com/os-autoinst/openQA/pull/3938 from my personal fork instead of a branch from main repo

PR #3744, opened 2021-02-19
Remove custom AssetPack pipe

kraih:
If i'm not mistaken, `OpenQA::WebAPI::AssetPipe` does nothing since the original issue was resolved [upstream back in 2016](https://github.com/jhthorsen/mojolicious-plugin-assetpack/commit/0a5fd6deb03d4334c62bb841bb57fd15ccf3ff48).

Progress: https://progress.opensuse.org/issues/71476


kraih:
The whole thing seems very odd, since the fix was committed just two weeks after `OpenQA::WebAPI::AssetPack` was added to openQA. Has this really been missed for 4 years, or is there perhaps a new use for it not covered by tests yet?

Martchus:
When trying to migrate to WebPack I've also noticed that and maybe it makes sense to get rid of it first.

> Has this really been missed for 4 years, or is there perhaps a new use for it not covered by tests yet?

I don't know that much of the history.

---

Our RPM build checks should cover this, right? So let's see whether they fail now. By the way, that's how one might test it manually: http://open.qa/docs/#update-asset-cache#update-asset-cache

kraih:
I did run `tools/generate-packed-assets` to regenerate the asset cache, which seemed to work just fine too.

Martchus (approved):

kalikiana (approved):

ilausuch (approved):

PR #3745, opened 2021-02-19
Allow specifying a custom commit message in needle editor

Martchus:
See https://progress.opensuse.org/issues/88564

### DIFF
--- a/lib/OpenQA/Task/Needle/Save.pm
+++ b/lib/OpenQA/Task/Needle/Save.pm
@@ -1,4 +1,4 @@
-# Copyright (C) 2019 SUSE LLC
+# Copyright (C) 2019-2021 SUSE LLC
 #
 # This program is free software; you can redistribute it and/or modify
 # it under the terms of the GNU General Public License as published by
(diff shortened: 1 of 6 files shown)


okurz:
Could you please crosscheck the runtime of t/ui/12-needle-edit.t? in your test runs it took 154s(!): 

```
t/ui/12-needle-edit.t .................. ok   153403 ms
```

in master it took also very long but still less https://app.circleci.com/pipelines/github/os-autoinst/openQA/5684/workflows/7278b681-3255-4fde-93ca-a3d60c9405a6/jobs/53642:

```
t/ui/12-needle-edit.t .................. ok   138640 ms
```

Amrysliu (approved):
LGTM

kalikiana (approved):

lib/OpenQA/Task/Needle/Save.pm
```diff
@@ -151,8 +152,8 @@ sub _save_needle {
     if ($git->enabled) {
         my $error = $git->commit(
             {
-                add     => ["$needledir/$needlename.json", "$needledir/$needlename.png"],
-                message => sprintf("%s for %s", $needlename, $openqa_job->name),
+                add     => ["$needlename.json", "$needlename.png"],
```
ilausuch, line 155: One question, why needledir is not longer necessary?

Martchus: We're basically using `git -C "$needledir" …` here anyways so it was never necessary. For the tests it was easier to get rid of the full path.

Martchus:
The test takes indeed longer, likely because I've enabled using Git. Locally it isn't that notable because the test is generally faster but according to @okurz's figures it is 14 seconds in production. I'm not sure yet how to improve that because the test is using forked processes so I can not easily change the configuration in the middle (to test just in one case with Git enabled).

okurz:
Hm, maybe try that coverage analysis is actually not running within any forked processes?

Martchus:
Looks like the last commit helped:

```
[13:03:05] t/ui/12-needle-edit.t .................. ok   130460 ms ( 0.19 usr  0.02 sys + 213.68 cusr  5.93 csys = 219.82 CPU)
```

okurz:
Great that you found a way to speedup the test. As this concerns graphical features, could you please provide a screenshot in the PR? This can also help for users finding this later althought I don't think we need to extend the documentation as the UI should be self-explanatory.

Martchus:
screenshot_20210223_110231

It shows the default commit message as placeholder text (when nothing is entered).

templates/webapi/step/edit.html.ep
```diff
@@ -152,6 +152,14 @@
                                     <input tupe="input" name="workaround_desc" id="input_workaround_desc" class="form-control" value="">
                                 </span>
                             </p>
+                            % if ($git_enabled) {
+                            <p>
+                                <label for="needleeditor_commit_message">Commit message:</label>
```
okurz, line 157: We should make it clear to the user that the git commit message details are optional. How about

s/Commit message:/Optional commit message details (leave empty for default text):/

okurz (changes requested):

templates/webapi/step/edit.html.ep
Martchus: If it is left empty it looks like this:

screenshot_20210223_115520

I've mentioned the use of the placeholder text before. I think with this we can omit the verbose explanation (which wouldn't even tell the user what the default text is).

okurz: Sure, I like the placeholder text but for me as a user it's not obvious if the grey text will be used by default or if it's just an example shown to the user but never ending in the git commit message. Just compare that to the text field in the lower right with the text "Add new tags here". This text is not automatically added as a tag if the user does not overwrite it, it's an explanation that is just shown to the UI user, not ending in the needle JSON.

okurz (approved):

PR #3746, opened 2021-02-19
Fix local upload for larger assets

psladek:
Asset file is now readable from server side
https://progress.opensuse.org/issues/88347


okurz (approved):
Thanks for supplying that solution directly yourself :)

Martchus (approved):

ilausuch (approved):

PR #3749, opened 2021-02-22
Attempt to fix path for os-autoinst

kalikiana:
See: https://progress.opensuse.org/issues/88915

### DIFF
--- a/codecov.yml
+++ b/codecov.yml
@@ -6,6 +6,7 @@ comment:
 # https://progress.opensuse.org/issues/88594
 fixes:
   - "/home/circleci/openQA::"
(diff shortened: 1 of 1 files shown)


Martchus (approved):
Basically the same as https://github.com/os-autoinst/openQA/pull/3728 but for os-autoinst's files.

okurz:
Actually I think we should not even collect coverage for any other source except openQA itself. Can you instead exclude the os-autoinst path from collecting?

kalikiana:
I'll leave this at that since the issue was solved and had nothing to do with the path. And I'm not sure what/how these paths are relevant to real issues.

PR #3750, opened 2021-02-22
Add "early-abort" in asset/results cleanup jobs based on df-output

Martchus:
* Make the minimum used disk space configurable having the new behavior not
  enabled by default
* See https://progress.opensuse.org/issues/88121#note-10

### DIFF
--- a/etc/openqa/openqa.ini
+++ b/etc/openqa/openqa.ini
@@ -139,6 +139,9 @@ blocklist = job_grab job_done
 
 [misc_limits]
 #untracked_assets_storage_duration = 14
+# Skips the cleanup of results/assets if the used disk space on the given partition is below the specified percentage
+#result_cleanup_min_percentage = 0
+#asset_cleanup_min_percentage = 0
 # Specify the screenshot ID range to query at once from the database (reduce to avoid big queries, increase to lower query overhead)
 #screenshot_cleanup_batch_size = 200000
 # Specify the number of screenshot ID ranges (with a size as configured by screenshot_cleanup_batch_size) to process in a single Minion
--- a/etc/openqa/openqa.ini
+++ b/etc/openqa/openqa.ini
@@ -139,9 +139,9 @@ blocklist = job_grab job_done
 
 [misc_limits]
 #untracked_assets_storage_duration = 14
-# Skips the cleanup of results/assets if the used disk space on the given partition is below the specified percentage
-#result_cleanup_min_percentage = 0
-#asset_cleanup_min_percentage = 0
+# Performs the cleanup of results/assets only if the free disk space on the relevant partition is below the specified percentage (and aborts early otherwise)
+#result_cleanup_max_free_percentage = 100
+#asset_cleanup_max_free_percentage = 100
 # Specify the screenshot ID range to query at once from the database (reduce to avoid big queries, increase to lower query overhead)
 #screenshot_cleanup_batch_size = 200000
 # Specify the number of screenshot ID ranges (with a size as configured by screenshot_cleanup_batch_size) to process in a single Minion
--- a/lib/OpenQA/Setup.pm
+++ b/lib/OpenQA/Setup.pm
@@ -1,4 +1,4 @@
-# Copyright (C) 2017-2020 SUSE LLC
+# Copyright (C) 2017-2021 SUSE LLC
 #
 # This program is free software; you can redistribute it and/or modify
 # it under the terms of the GNU General Public License as published by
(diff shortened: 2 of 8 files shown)


etc/openqa/openqa.ini
```diff
@@ -139,6 +139,9 @@ blocklist = job_grab job_done
 
 [misc_limits]
 #untracked_assets_storage_duration = 14
+# Skips the cleanup of results/assets if the used disk space on the given partition is below the specified percentage
+#result_cleanup_min_percentage = 0
+#asset_cleanup_min_percentage = 0
```
okurz, line 144: As we commonly use the plural form I suggest for consistency to use that here as well

```suggestion
#results_cleanup_min_percentage = 0
#assets_cleanup_min_percentage = 0
```

e.g. also see the lines above and below

Also I suggested to use the "free" capacity, not "used" capacity. Any specific reason you chose "used"?

Reasons for using "free" is that commonly one wants to keep a bit of free space regardless of how much is actually used. And when one increases the available storage space the number for "used" might need adjustment but not necessarily for "free". Commonly it's a good practice to keep something like 10% of a filesystem free to provide enough buffer as well as keep a filesystem performant. This number would always stay the same, regardless of how much more one adds as storage. Of course everyone could calculate that again to be "90% used" but why ask the user to calculate that and you then calculate "100 - …" in code as well.

okurz (changes requested):
I see that you add asset/results cleanup in the same step. Can we try to focus on one at a time, e.g. results only first?

Martchus:
> I see that you add asset/results cleanup in the same step. Can we try to focus on one at a time, e.g. results only first?

If it really has to be I can split the commit retrospectively (which likely means rebasing and solving conflicts of the 2nd commit later). However, I don't think it is necessary because the behavior can still be configured individually.

etc/openqa/openqa.ini
Martchus: I don't care about this. I guess I've chosen it because `0` then naturally becomes the "disabled" value which is quite common. With your suggestion I'd change the default to `100` which will then be the "disabled" value.

How would you write the explaining comment then?

I suppose

> Skips the cleanup of results/assets if the used disk space on the given partition is below the specified percentage

would turn into

> Performs the cleanup of results/assets only if the free disk space on the relevant partition is below the specified percentage (and aborts early otherwise)

The keys would be e.g. `results_cleanup_max_free_percentage` then (unless you have a better idea).

okurz: > With your suggestion I'd change the default to 100 which will then be the "disabled" value.

Do you mean that "disabled" would "disable the feature to skip cleanup"?

> The keys would be e.g. "results_cleanup_max_free_percentage"

I thought of "results_min_free_percentage" as in "trying to keep at least the specified percentage free". And you don't need the word "cleanup" because "min_free" already implies that space is "kept free", of course by removing stuff.

okurz:
> > I see that you add asset/results cleanup in the same step. Can we try to focus on one at a time, e.g. results only first?
> 
> If it really has to be I can split the commit retrospectively (which likely means rebasing and solving conflicts of the 2nd commit later). However, I don't think it is necessary because the behavior can still be configured individually.

Sure. Whatever works best for you. Just wanted to suggest a way to get independent parts completed quicker, including deployment, risk in production, potential reverts, etc.

etc/openqa/openqa.ini
Martchus: > Do you mean that "disabled" would "disable the feature to skip cleanup"?

Yes. With my first approach `0` would retain the current default behavior and with using the free percentage `100` would retain the current default behavior.

"results_min_free_percentage" seems misleading because this doesn't configure something to ensure a minimum free percentage. It configures not to take care if there's more free disk space than the specified percentage. I don't think "cleanup" is redundant. "results_min_free_percentage" sounds like openQA might cease to take more jobs to ensure the min free disk space and it is not necessarily clear that the thresholds is cleanup related.

okurz: hm. Looking forward for the plans we had I still think "results_min_free_percentage" might be better but in the current implementation of course your suggestion is better fitting. Ok, we can keep it as you proposed and see if later we want to keep it, introduce another setting or add additional ones

okurz: I see that you changed it now and have used the default 100. As I wrote above I would have accepted your original proposal. And I find the default value of 100 as tricky. If you want to keep the new names how about treating 0% as a special disabled value?

Martchus: 100 as default is not really tricky. I've just figured that this was the reason I intuitively make it work the other way around. We can still treat 0 as special value but I'm not sure whether it makes things better.

okurz: fine. Anyway, looks good now.

Amrysliu (approved):
LGTM

okurz (approved):
Awesome test, easy to read :)

PR #3751, opened 2021-02-22
Use codecov-bash to upload coverage reports

kalikiana:
As per suggestion in [poo#88915](https://progress.opensuse.org/issues/88915) since it was't much work and I was still investigating the upload errors. And despite this being reproduced in other projects, *codecov-bash* seems to work fine.

- Add JSON::MaybeXS and use JSON for codecov db in CI
- Update and use makefile targets in CI
- Replace codecov (perl) with codecov-bash

**Note**: [Devel::Cover::Report::Codecovbash](https://metacpan.org/pod/Devel::Cover::Report::Codecovbash) is used in CI via CPAN

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -111,6 +111,7 @@ images:
       # that context in the prove calling context.
       PERL_TEST_HARNESS_DUMP_TAP: test-results
       HARNESS: --harness TAP::Harness::JUnit --timer
+      DEVEL_COVER_DB_FORMAT: JSON
       COVEROPT: -MDevel::Cover=-select_re,'^/lib',+ignore_re,lib/perlcritic/Perl/Critic/Policy,-coverage,statement
       COMMIT_AUTHOR_EMAIL: skynet@open.qa
 
@@ -269,10 +270,16 @@ jobs:
       - run: *install_cached_packages
       - attach_workspace:
           at: .
-      - run: cover -write cover_db cover_db*
-      - run: cover -select_re '^(lib|script|t)/' -report html_minimal
+      - run:
+          name: Merge coverage databases
+          command: |
+            cover -write cover_db cover_db*
+            ls cover_db/*cov*
+      - run: make coverage-html
       - store_artifacts: *store_cover_db
-      - run: cover -select_re '^(lib|script|t)/' -report codecov
+      - run: sudo cpanm -nq 'Devel::Cover::Report::Codecovbash'
+      - run: make coverage-codecov
+      - run: bash <(curl -s https://codecov.io/bash)
 
   build-docs: &docs-template
     name: "Generating docs"
--- a/Makefile
+++ b/Makefile
@@ -204,12 +204,12 @@ coverage:
 COVER_REPORT_OPTS ?= -select_re '^(lib|script|t)/'
 
 .PHONY: coverage-codecov
-coverage-codecov: coverage
-	cover $(COVER_REPORT_OPTS) -report codecov
+coverage-codecov:
+	cover $(COVER_REPORT_OPTS) -report codecovbash
 
 .PHONY: coverage-html
-coverage-html: coverage
-	cover $(COVER_REPORT_OPTS) -report html_basic
+coverage-html:
+	cover $(COVER_REPORT_OPTS) -report html_minimal
 
 public/favicon.ico: assets/images/logo.svg
 	for w in 16 32 64 128; do \
--- a/dependencies.yaml
+++ b/dependencies.yaml
@@ -72,6 +72,7 @@ devel_no_selenium_requires:
   xorg-x11-fonts:
   perl(Devel::Cover):
   perl(Devel::Cover::Report::Codecov):
+  perl(JSON::MaybeXS):
   perl(Perl::Tidy): '== 20210111'
 
 devel_requires:
--- a/.circleci/ci-packages.txt
+++ b/.circleci/ci-packages.txt
@@ -122,6 +122,7 @@ perl-JavaScript-Minifier-XS-0.11
 perl-JSON-2.97001
 perl-JSON-Validator-3.08
 perl-JSON-XS-3.04
+perl-JSON-MaybeXS
 perl-libwww-perl-6.31
 perl-Lingua-EN-Inflect-1.903
 perl-List-MoreUtils-0.428
--- a/cpanfile
+++ b/cpanfile
@@ -116,6 +116,7 @@ on 'test' => sub {
 on 'devel' => sub {
     requires 'Devel::Cover';
     requires 'Devel::Cover::Report::Codecov';
+    requires 'JSON::MaybeXS';
     requires 'Perl::Tidy', '== 20210111';
 
 };
--- a/dist/rpm/openQA.spec
+++ b/dist/rpm/openQA.spec
@@ -71,7 +71,7 @@
 %define qemu qemu
 %endif
 # The following line is generated from dependencies.yaml
-%define devel_no_selenium_requires %build_requires %qemu %test_requires curl perl(Devel::Cover) perl(Devel::Cover::Report::Codecov) perl(Perl::Tidy) postgresql-devel rsync sudo tar xorg-x11-fonts
+%define devel_no_selenium_requires %build_requires %qemu %test_requires curl perl(Devel::Cover) perl(Devel::Cover::Report::Codecov) perl(JSON::MaybeXS) perl(Perl::Tidy) postgresql-devel rsync sudo tar xorg-x11-fonts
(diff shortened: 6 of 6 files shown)


okurz:
well, https://app.circleci.com/pipelines/github/os-autoinst/openQA/5709/workflows/c0706f2f-c472-43e0-ac38-205f1bc2fddf/jobs/53855 looks good but codecov did not provide a report comment here. Also I think you need to adjust the dependencies. See https://github.com/os-autoinst/os-autoinst/pull/1576/files for the approach I took for os-autoinst

kalikiana:
> well, https://app.circleci.com/pipelines/github/os-autoinst/openQA/5709/workflows/c0706f2f-c472-43e0-ac38-205f1bc2fddf/jobs/53855 looks good but codecov did not provide a report comment here. Also I think you need to adjust the dependencies. See https://github.com/os-autoinst/os-autoinst/pull/1576/files for the approach I took for os-autoinst

Ack. I'm doubtful now that this won't hit the bug due to codecov/codecov-bash#411 but will try anyway

okurz:
https://app.circleci.com/pipelines/github/os-autoinst/openQA/5712/workflows/6f8cc47a-099f-4fe0-ad8e-7082677adf86/jobs/53894

```
Reading database from /home/squamata/project/cover_db
Can't read /home/squamata/project/cover_db/cover.14 with Cpanel::JSON::XS: malformed JSON string, neither tag, array, object, number, string or atom, at character offset 0 (before "pst0\x{5}\n\x{3}\x{0}...") at /usr/lib/perl5/vendor_perl/5.26.1/x86_64-linux-thread-multi/Devel/Cover/DB/IO/JSON.pm line 35, <$fh> chunk 1.
Can't read /home/squamata/project/cover_db/structure/01432dcb40271f5f1f39aebec8308817 with Cpanel::JSON::XS: malformed JSON string, neither tag, array, object, number, string or atom, at character offset 0 (before "pst0\x{5}\n\x{3}\x{0}...") at /usr/lib/perl5/vendor_perl/5.26.1/x86_64-linux-thread-multi/Devel/Cover/DB/IO/JSON.pm line 35, <$fh> chunk 1.
```

Have you tried that locally first?

kalikiana:
> https://app.circleci.com/pipelines/github/os-autoinst/openQA/5712/workflows/6f8cc47a-099f-4fe0-ad8e-7082677adf86/jobs/53894
> 
> ```
> Reading database from /home/squamata/project/cover_db
> Can't read /home/squamata/project/cover_db/cover.14 with Cpanel::JSON::XS: malformed JSON string, neither tag, array, object, number, string or atom, at character offset 0 (before "pst0\x{5}\n\x{3}\x{0}...") at /usr/lib/perl5/vendor_perl/5.26.1/x86_64-linux-thread-multi/Devel/Cover/DB/IO/JSON.pm line 35, <$fh> chunk 1.
> Can't read /home/squamata/project/cover_db/structure/01432dcb40271f5f1f39aebec8308817 with Cpanel::JSON::XS: malformed JSON string, neither tag, array, object, number, string or atom, at character offset 0 (before "pst0\x{5}\n\x{3}\x{0}...") at /usr/lib/perl5/vendor_perl/5.26.1/x86_64-linux-thread-multi/Devel/Cover/DB/IO/JSON.pm line 35, <$fh> chunk 1.
> ```
> 
> Have you tried that locally first?

Yes and I was hoping it was something I broke because I was running too many commands. But it seems like the same happens here. Not sure yet what's happening here 🤔

okurz:
> To expand on that, locally I had seen this, deleted the db and after running the commands again it was fine... like this works fine here:
> 
> ```
> rm -Rf cover_db
> make coverage TESTS=t/03-auth.t
> cover -write cover_db cover_db*
> cover -select_re '^(lib|script|t)/' -report html_minimal
> cover -select_re '^(lib|script|t)/' -report codecovbash
> bash <(curl -s https://codecov.io/bash)
> ```

It's a good idea to limit the coverage to a single module so that it's way faster and also rules out that some components included in the source we want to gather coverage on cause the problem. Maybe you can just do the equivalent within circleCI and see if you get a coverage report from codecov for just `t/03-auth.t`?

Makefile
```diff
@@ -204,12 +204,12 @@ coverage:
...
 .PHONY: coverage-codecov
-coverage-codecov: coverage
-	cover $(COVER_REPORT_OPTS) -report codecov
+coverage-codecov:
+	cover $(COVER_REPORT_OPTS) -report codecovbash
 
 .PHONY: coverage-html
-coverage-html: coverage
-	cover $(COVER_REPORT_OPTS) -report html_basic
+coverage-html:
```
okurz, line 211: This is changing the previous notion of `coverage-html` implying that the coverage is being collected. What is the specific reason you needed to change this?

dependencies.yaml
```diff
@@ -72,6 +72,7 @@ devel_no_selenium_requires:
   xorg-x11-fonts:
   perl(Devel::Cover):
   perl(Devel::Cover::Report::Codecov):
+  perl(JSON::MaybeXS):
```
okurz, line 75: what is this specifically needed for? Shouldn't it be an implicit dependency of some other package?

.circleci/config.yml
```diff
@@ -269,10 +270,16 @@ jobs:
       - run: *install_cached_packages
       - attach_workspace:
           at: .
-      - run: cover -write cover_db cover_db*
-      - run: cover -select_re '^(lib|script|t)/' -report html_minimal
+      - run:
+          name: Merge coverage databases
+          command: |
+            cover -write cover_db cover_db*
+            ls cover_db/*cov*
```
okurz, line 277: Is this still necessary or debugging code that should be removed?

.circleci/config.yml
```diff
@@ -269,10 +270,16 @@ jobs:
...
+      - run: make coverage-html
       - store_artifacts: *store_cover_db
-      - run: cover -select_re '^(lib|script|t)/' -report codecov
+      - run: sudo cpanm -nq 'Devel::Cover::Report::Codecovbash'
```
okurz, line 280: Please also ensure that the package is submitted to openSUSE:Factory. This can be done as part of the ticket, pending acceptance not blocking the PR.

okurz (changes requested):
You made coverage work again and the report does not even reveal weird coverage drops \o/

dependencies.yaml
kraih: I believe it should.

Makefile
kalikiana: Because I wasted some of my time investigating with incorrectly replicating the steps done in CI. I'd like to merge the steps. At least this is a step in that direction.

okurz: Sorry, I don't understand. The line `cover $(COVER_REPORT_OPTS) -report html_minimal` only creates the final report but does not conduct any tests to initially collect the coverage data. Where do you "merge steps"?

.circleci/config.yml
kalikiana: I found it useful to find out if things work correctly.

dependencies.yaml
okurz: Unless it's the only missing dependency for the module that cdywan downloads from CPAN?

kalikiana: Yes, except *Devel::Cover::Report::Codecovbash* is not a package yet.

.circleci/config.yml
okurz: Of course. Question is if this should be kept in production code or remove it again as it was only useful for you at the time.

dependencies.yaml
perlpunk: But since you are installing it via cpanm, the dependency should be installed automatically

okurz: ok, shouldn't matter. We can keep it as a task for later in the related ticket or a followup to use properly packaged dependencies.

okurz (approved):
Ok, let's try to expedite this fix. I can approve and we can handle open points in the ticket which I noted there.

dependencies.yaml
Martchus: For now it makes sense to have the dependency so we need to install as less as possible via cpanm.

Martchus (approved):

Martchus (approved):

dependencies.yaml
kraih: `JSON::MaybeXS` is a super light depedency. Installing it via `cpanm` should be no problem at all.

Makefile
kalikiana: I meant making sure CI uses the same makefile targets humans use. The HTML didn't even work for me when I tried it locally, so I don't think it was used at all.

PR #3752, opened 2021-02-24
Mark t/ui/27-plugin_obs_rsync_status_details.t as stable again

okurz:


Martchus (approved):
Since it previously happened 1/2 of the time I suppose your testing is sufficient.

okurz:
well, I still assume that circleCI has different environments whenever new jobs are spawned. What I have not done is restart complete circleCI job itself multiple times or on multiple days but this is something we will surely learn after merge :)

andrii-suse (approved):
Thx for covering this!

PR #3712, opened 2021-02-09
Rework symlink CASEDIR and set values to relative path

Amrysliu:
The third commit is used to fix the issues mentioned in:
https://progress.opensuse.org/issues/67723#note-44

Here is also another about needles_dir. If users use `openqa-clone-custom-git-refspec` to trigger a new job whose CASEDIR is a basename, the needles_dir will be set as a relative path, the job will incomplete because fail to do NEEDLES_DIR symlink.

Verify run:
For PRODUCTDIR: http://10.67.19.103/tests/448#settings
For NEEDLES_DIR: http://10.67.19.103/tests/449#settings

PR #3753, opened 2021-02-24
Restore cover and run makefile targets

kalikiana:
- CI can use the new -report variants of the targets
- Cover format and options are enabled via COVERAGE=1

```
COVERAGE=1 make test-with-database
make coverage-report-html
```

and

    make coverage-html

are analoguous, the former way is used in CI.

See [poo#88915](https://progress.opensuse.org/issues/88915)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -171,12 +171,14 @@ test-fullstack-unstable:
 .PHONY: test-with-database
 test-with-database:
 	test -d $(TEST_PG_PATH) && (pg_ctl -D $(TEST_PG_PATH) -s status >&/dev/null || pg_ctl -D $(TEST_PG_PATH) -s start) || ./t/test_postgresql $(TEST_PG_PATH)
-	PERL5OPT="$(PERL5OPT) -It/lib -I$(PWD)/t/lib -MOpenQA::Test::PatchDeparse" $(MAKE) test-unit-and-integration TEST_PG="DBI:Pg:dbname=openqa_test;host=$(TEST_PG_PATH)"
+	$(MAKE) test-unit-and-integration TEST_PG="DBI:Pg:dbname=openqa_test;host=$(TEST_PG_PATH)"
 	-[ $(KEEP_DB) = 1 ] || pg_ctl -D $(TEST_PG_PATH) stop
 
 .PHONY: test-unit-and-integration
 test-unit-and-integration:
 	export GLOBIGNORE="$(GLOBIGNORE)";\
+	export DEVEL_COVER_DB_FORMAT=JSON;\
+	export PERL5OPT="$(COVEROPT)$(PERL5OPT) -It/lib -I$(PWD)/t/lib -MOpenQA::Test::PatchDeparse";\
 	RETRY=${RETRY} timeout -s SIGINT -k 5 -v ${TIMEOUT_RETRIES} tools/retry prove ${PROVE_LIB_ARGS} ${PROVE_ARGS}
 
 # prepares running the tests within a container (eg. pulls os-autoinst) and then runs the tests considering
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -171,12 +171,13 @@ test-fullstack-unstable:
...
 test-with-database:
 	test -d $(TEST_PG_PATH) && (pg_ctl -D $(TEST_PG_PATH) -s status >&/dev/null || pg_ctl -D $(TEST_PG_PATH) -s start) || ./t/test_postgresql $(TEST_PG_PATH)
-	PERL5OPT="$(PERL5OPT) -It/lib -I$(PWD)/t/lib -MOpenQA::Test::PatchDeparse" $(MAKE) test-unit-and-integration TEST_PG="DBI:Pg:dbname=openqa_test;host=$(TEST_PG_PATH)"
+	PERL5OPT="$(COVEROPT) $(PERL5OPT) -It/lib -I$(PWD)/t/lib -MOpenQA::Test::PatchDeparse" $(MAKE) test-unit-and-integration TEST_PG="DBI:Pg:dbname=openqa_test;host=$(TEST_PG_PATH)"
 	-[ $(KEEP_DB) = 1 ] || pg_ctl -D $(TEST_PG_PATH) stop
 
 .PHONY: test-unit-and-integration
 test-unit-and-integration:
 	export GLOBIGNORE="$(GLOBIGNORE)";\
+	export DEVEL_COVER_DB_FORMAT=JSON;\
```
okurz, line 180: Doesn't that change the format for manually generated reports or does it only influence the intermediate format used by Devel::Cover itself?

okurz (approved):
I am happy that you managed to get rid of the near-duplicate definition of the coverage options \o/

Makefile
kalikiana: This is what's referred to [in Devel::Cover docs](https://metacpan.org/pod/Devel::Cover#Developer-variables) as `DB format`. Otherwise the format depends on the installed packages and is the reason #3751 was failing with weird JSON errors at one point.

Using it here also means you always get the same as what we use in CI, and you get an error if you don't have the dependencies needed to use that format.

kalikiana:
Small update here:
- I made a tiny fix to add the `,` so the arguments are always valid
- test-stylecheck got legitimately confused because PERL5OPT was set to partial arguments, so I decided to change this to a COVERDB_SUFFIX variable and define all args in the makefile

Martchus (approved):
Looks like there's some setup error in the style check but it might be unrelated.

kalikiana:
> Looks like there's some setup error in the style check but it might be unrelated.

Hrmmmm `The specified paths did not match any files in /home/squamata/project` no idea what this means, for now I'll re-run. I'm not aware that I changed anything to the paths used or such, that's all in `codecov.yml` 🤔

okurz:
I am pretty sure the problem is related to your changes but can't put my finger onto it

kalikiana:
> I am pretty sure the problem is related to your changes but can't put my finger onto it

Turns out `PERL5OPT` must be set in *test-unit-and-integration* because that's what *checkstyle* is based on. Otherwise there's no coverage. Previously CI would have overriden this and locally it just didn't work.

okurz (changes requested):
now checks in OBS fail like in https://build.opensuse.org/package/live_build_log/devel:openQA:TestGithub:PR-3753/openQA/openSUSE_Factory/x86_64 with

```
[   64s] export PERL5OPT=", -It/lib -I/home/abuild/rpmbuild/BUILD/openQA-4.6.1614787879.d3309e4d1/t/lib -MOpenQA::Test::PatchDeparse";\
[   64s] RETRY=0 timeout -s SIGINT -k 5 -v $((60 * (0 + 1) ))m tools/retry prove -l -r -v
[   64s] Illegal switch in PERL5OPT: -,.
```

okurz (approved):
Assuming all checks pass

kalikiana:
Seems like this is stuck at 5/10 similar to #3712 but I don't see any pending builds

okurz:
Right now it does not seem stuck but "blocked" by a Factory rebuild. IMHO ok to wait here as well.

PR #3756, opened 2021-03-02
Limit execution time of hook scripts run within Minion

Martchus:
This should prevent blocking Minion workers for a long time and other Minion jobs piling up for too long.

See https://progress.opensuse.org/issues/89224

### DIFF
--- a/lib/OpenQA/Task/Job/FinalizeResults.pm
+++ b/lib/OpenQA/Task/Job/FinalizeResults.pm
@@ -1,4 +1,4 @@
-# Copyright (C) 2020 SUSE LLC
+# Copyright (C) 2020-2021 SUSE LLC
 #
 # This program is free software; you can redistribute it and/or modify
 # it under the terms of the GNU General Public License as published by
@@ -14,7 +14,7 @@
 # with this program; if not, see <http://www.gnu.org/licenses/>.
 
 package OpenQA::Task::Job::FinalizeResults;
-use Mojo::Base 'Mojolicious::Plugin';
+use Mojo::Base 'Mojolicious::Plugin', -signatures;
 use OpenQA::Jobs::Constants 'CANCELLED';
 
 sub register {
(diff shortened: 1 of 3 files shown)


lib/OpenQA/Task/Job/FinalizeResults.pm
```diff
@@ -14,7 +14,7 @@
 # with this program; if not, see <http://www.gnu.org/licenses/>.
 
 package OpenQA::Task::Job::FinalizeResults;
-use Mojo::Base 'Mojolicious::Plugin';
+use Mojo::Base 'Mojolicious::Plugin', -signatures;
```
kraih, line 17: Don't you want to convert the whole module if you're the one introducing `-signatures` into it?

Martchus: Definitely not as part of the same commit and not necessarily as part of this PR. I personally would simply convert every function at the time its parameters are changed anyways.

okurz (approved):

lib/OpenQA/Task/Job/FinalizeResults.pm
```diff
@@ -49,15 +49,16 @@ sub _finalize_results {
     return if $carried_over;
     my $key = 'job_done_hook_' . $openqa_job->result;
     if (my $hook = $ENV{'OPENQA_' . uc $key} // $app->config->{hooks}->{lc $key}) {
-        my $ret = _done_hook_new_issue($openqa_job, $hook);
+        my $timeout      = $ENV{'OPENQA_JOB_DONE_HOOK_TIMEOUT'}      // '5m';
+        my $kill_timeout = $ENV{'OPENQA_JOB_DONE_HOOK_KILL_TIMEOUT'} // '30s';
```
kraih, line 53: There is no need to quote the hash keys (`$ENV{OPENQA_JOB_DONE_HOOK_KILL_TIMEOUT}`).

kraih: It's actually inconsistent with the other uses of the same env var in the test.

kraih (approved):

PR #3757, opened 2021-03-02
Add personal access tokens as an alternative to X-API-* headers

kraih:
A personal access token can be constructed from existing API keys and secrets. The format is `USER:KEY:SECRET`.

This allows for other clients than OpenQA::Client to access the openQA API with authentication for operator and admin privileges. Since the token is submitted via Basic authentication, which pretty much all HTTP clients support, it is as easy as `curl -u kraih:KEY:SECRET ...` to use.

As a security precaution, personal access tokens are currently limited to HTTPS requests originating from any host, and HTTP requests originating from localhost. This will prevent API secrets from being sent in cleartext over the network.

Reverse proxy setups might need to add an `X-Forwarded-Proto: https` header to properly signal to the application server that the connection to the client is actually secure.

Progress: https://progress.opensuse.org/issues/87698

### DIFF
--- a/lib/OpenQA/Shared/Controller/Auth.pm
+++ b/lib/OpenQA/Shared/Controller/Auth.pm
@@ -14,26 +14,24 @@
 # with this program; if not, see <http://www.gnu.org/licenses/>.
 
 package OpenQA::Shared::Controller::Auth;
-use Mojo::Base 'Mojolicious::Controller';
+use Mojo::Base 'Mojolicious::Controller', -signatures;
 
 use OpenQA::Schema;
-use OpenQA::Log 'log_debug';
-use Mojo::Util 'hmac_sha1_sum';
+use OpenQA::Log qw(log_debug);
+use Mojo::Util qw(hmac_sha1_sum secure_compare);
 
-sub check {
-    my $self = shift;
-
-    my $req = $self->req;
+sub check ($self) {
     if ($self->app->config->{no_localhost_auth}) {
         return 1 if $self->is_local_request;
     }
 
+    my $req       = $self->req;
     my $headers   = $req->headers;
     my $key       = $headers->header('X-API-Key');
     my $hash      = $headers->header('X-API-Hash');
     my $timestamp = $headers->header('X-API-Microtime');
     my $user;
-    log_debug($key ? "API key from client: *$key*" : "No API key from client.");
+    log_debug($key ? "API key from client: *$key*" : 'No API key from client');
 
     my $schema  = OpenQA::Schema->singleton;
     my $api_key = $schema->resultset('ApiKeys')->find({key => $key});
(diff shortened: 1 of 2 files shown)


kraih:
We are reusing the same workflow as the existing API key/secret authentication uses. So only the actual authentication mechanism is exchanged, and code changes are very limited in scope.

I'd really like to add some documentation too, but am not sure where it would best fit in. If nobody has a better suggestion i'll add something to the [Client/Authentication](http://open.qa/docs/#client) section.

lib/OpenQA/Shared/Controller/Auth.pm
```diff
@@ -56,62 +54,82 @@ sub check {
...
+                my ($username, $key, $secret) = ($1, $2, $3);
+                $log->debug(qq{Personal access token for user "$username"});
+                if ($self->is_local_request || $self->req->is_secure) {
+                    my $ip         = $self->tx->remote_address;
+                    my $reject_msg = qq{Rejecting personal access token for user "$username" with ip "$ip"};
+                    if (my $api_key = $self->schema->resultset('ApiKeys')->find({key => $key})) {
+                        my $real_user = $api_key->user;
+                        if ($real_user && secure_compare($real_user->username, $username)) {
+                            if (secure_compare($api_key->secret, $secret)) { ($user, $reason) = ($real_user, undef) }
+                            else                                           { $log->error("$reject_msg, wrong secret") }
```
Martchus, line 83: If a user mis-types its secret, should this really be logged as an error? This is not a problem an admin should be concerned about. The same counts for the other errors.

lib/OpenQA/Shared/Controller/Auth.pm
```diff
@@ -56,62 +54,82 @@ sub check {
     return undef;
 }
 
-sub auth {
-    my $self = shift;
-    my $log  = $self->app->log;
-    my $user;
-
-    my $reason = "Not authorized";
+sub auth ($self) {
```
Martchus, line 57: Maybe it is time to split the code into smaller functions.

Martchus:
About the documentation: I think the section you've found is already a good fit. At least the sentence "And that is usually where you reach the limits of curl" needs to be adjusted anyways.

lib/OpenQA/Shared/Controller/Auth.pm
kraih: I've based this on the API key auth logging. There it's also an error message.

kraih: Personally, i don't think it's all that useful and would be happy to change it to whatever we can agree on.

lib/OpenQA/Shared/Controller/Auth.pm
kraih: Fair point, i was a bit focused on keeping all the old logic that i ignored opportunities for making the code cleaner.

lib/OpenQA/Shared/Controller/Auth.pm
Martchus: I guess it makes sense to adapt the existing API key auth logging then.

lib/OpenQA/Shared/Controller/Auth.pm
Martchus: It also makes sense to do this kind of refactoring as a 2nd step (e.g. 2nd commit or further PR).

okurz (approved):
This looks great already! I checked the test and there is no unexpected output and runtime is also fine. Of you like, as suggested by Martchus you can show up with some style changes, multiple commits would be appreciated. Also I suggest to reference the ticket in the git commit message

kalikiana (approved):

kraih:
Ok, i'll make separate PRs for documentation and code cleanup. 😀

PR #3758, opened 2021-03-02
CI: Add compose tests via GitHub workflow

kalikiana:
Testing out a simple GHA to test the `docker-compose.yml` for the *webui* using [isbang/compose-action](https://github.com/marketplace/actions/docker-compose-action).

Related to #3755

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -135,7 +135,7 @@ endif
 endif
 
 .PHONY: test-checkstyle
-test-checkstyle: test-checkstyle-standalone test-tidy-compile
+test-checkstyle: test-checkstyle-standalone test-tidy-compile test-containers-compose
 
 .PHONY: test-t
 test-t:
@@ -279,6 +279,10 @@ tidy-js: check-js-beautify
 .PHONY: tidy
 tidy: tidy-js
 	tools/tidy
+
+.PHONY: test-containers-compose
+test-containers-compose:
+	for i in $(wildcard container/**/docker-compose.yaml); do sudo docker-compose up -d -f $$i; done
 
 .PHONY: update-deps
 update-deps:
--- a/.github/workflows/compose.yml
+++ b/.github/workflows/compose.yml
@@ -1,0 +1,11 @@
+---
+name: compose
+# yamllint disable-line rule:truthy
+on: [push, pull_request]
+jobs:
+  webui-docker-compose:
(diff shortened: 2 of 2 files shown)


okurz:
Good idea, I don't see it executed though. Do we still need to enable something?

kalikiana:
> Good idea, I don't see it executed though. Do we still need to enable something?

Unsure. I'm getting notifications with build results so it is being executed... somewhere

And I'm getting [this error](https://github.com/kalikiana/openQA/runs/2014727215?check_suite_focus=true):

```
Run isbang/compose-action@v0.1
Building with native build. Learn about native build in Compose here: https://docs.docker.com/go/compose-native-build/
Creating network "webui_default" with the default driver
Pulling data (openqa_data:)...
The image for the service you're trying to recreate has been removed. If you continue, volume data could be lost. Consider backing up your data before continuing.
Continue with the new image? [yN]
Error: The operation was canceled.
```

And I get the same running `sudo docker-compose up` locally. @ilausuch can you confirm that the yml in master is not working?

Makefile
```diff
@@ -280,6 +280,10 @@ tidy-js: check-js-beautify
 tidy: tidy-js
 	tools/tidy
 
+.PHONY: test-containers-compose
+test-containers-compose:
```
okurz, line 284: Hm, that would be the first target that is not included when calling "make test" without any other arguments. I suggest to include it there as well. See how we currently split targets explicitly for different circleCI jobs but still have it included when one calls "make test" locally

kalikiana: I was a bit worried to get pushback if I add more to `test` :-D I'll add it then

kalikiana:
I'm not sure if requesting review here makes sense - since we were also discussing with @ilausuch and @okurz to propose this together with fixes needed to make compose work. But I'll make it "ready" now, and we see from there

Makefile
okurz: Which would mean you would need my changes first

ilausuch (approved):
As far as is commented about the test-checkstyle I approve that

kalikiana:
Note that the "errors" in checkstyle are because now CircleCI runs the test, rather than just GHA

kalikiana:
Closing in favor of pulling the test into #3773

PR #3761, opened 2021-03-03
CI: Add hadolint validation via GitHub workflow

kalikiana:
Using [hadolint via GHA](https://github.com/marketplace/actions/hadolint-action), locally tested like so:

    podman run --rm -i hadolint/hadolint <container/worker/Dockerfile
    podman run --rm -i hadolint/hadolint <container/webui/Dockerfile

See: [poo#88451](https://progress.opensuse.org/issues/88451)

### DIFF
--- a/.github/workflows/hadolint.yml
+++ b/.github/workflows/hadolint.yml
@@ -1,0 +1,11 @@
+---
+name: hadolint
+# yamllint disable-line rule:truthy
+on: [push, pull_request]
+jobs:
+  hadolint:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - name: Validate container images
+        run: make test-containers-hadolint
--- a/Makefile
+++ b/Makefile
@@ -280,6 +280,11 @@ tidy-js: check-js-beautify
 tidy: tidy-js
 	tools/tidy
 
+.PHONY: test-containers-hadolint
+test-containers-hadolint:
+	docker run --rm -i hadolint/hadolint <container/webui/Dockerfile
+	docker run --rm -i hadolint/hadolint <container/worker/Dockerfile
+
 .PHONY: update-deps
 update-deps:
 	tools/update-deps --specfile dist/rpm/openQA.spec
--- a/container/webui/Dockerfile
+++ b/container/webui/Dockerfile
@@ -3,11 +3,13 @@ FROM opensuse/leap:15.2
 LABEL maintainer Jan Sedlak <jsedlak@redhat.com>, Josef Skladanka <jskladan@redhat.com>, wnereiz <wnereiz@eienteiland.org>, Sergio Lindo Mansilla <slindomansilla@suse.com>
 LABEL version="0.3"
 
+# hadolint ignore=DL3037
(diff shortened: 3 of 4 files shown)


.github/workflows/hadolint.yml
```diff
@@ -0,0 +1,14 @@
...
+  hadolint:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - uses: brpaz/hadolint-action@v1.2.1
+        with:
+          dockerfile: './container/webui/Dockerfile'
+      - uses: brpaz/hadolint-action@v1.2.1
+        with:
+          dockerfile: './container/worker/Dockerfile'
```
okurz, line 14: I guess using our own shell style call for this could help to catch all Dockerfiles, not just the ones that are explicitly specified

kalikiana: There is [#3](https://github.com/brpaz/hadolint-action/issues/3#issuecomment-646366920) about this, so basically we could run `hadolint ./container/**/Dockerfile` with `container: hadolint/hadolint`. Upstream seems happy to take fixes, not sure how easy that is.

okurz: Yes, but what I am suggesting is to not rely on github action specific stuff that people can not reproduce locally and also is not complete but rather use simple shell scripts that solve both points, same as in https://github.com/os-autoinst/os-autoinst/pull/1604/files#diff-1af57dc1e7a5a58067566f2787f0030953af77313ae8d5488b06f71a851640b7R11

kalikiana: That's a totally different point. But yes, I also brought it up in a daily last week, where these points came up:

- You can see the reproducer in the PR description but by default GHAs don't advertise how to reproduce them locally.
- GHAs are nice and concise snippets which we don't have to maintain.
- Usually makefile targets are the most convenient to reproduce something locally.

So maybe what we want here is a `test-containers-hadolint` target, and `test-containers-compose` for #3758 respectively.

kalikiana: Well, not as straightforward as I thought, I'm getting all kinds of funny errors trying to simply translate the glob to something running in the container... but added the new target with two individual lines anyway

okurz: Have seen and understood the line
https://github.com/os-autoinst/os-autoinst/pull/1604/files#diff-1af57dc1e7a5a58067566f2787f0030953af77313ae8d5488b06f71a851640b7R11
?

Maybe you forgot to pass the current working directory into the container because the globbing works on passing paths to hadolint within the container so hadolint needs to be able to find the Dockerfile's *within* the container environment

kalikiana: If this is more obvious to you, I'd welcome a one-line suggestion to make it work in the `Makefile`.

kalikiana: Variants including `-v $(pwd):/root --workdir=/root` and `$(readlink -f "$(dirname "$0")"/..):/root --workdir=/root`

okurz: Well, I consider the best approach doing the same as in os-autoinst, so trying with #3772

PR #3763, opened 2021-03-03
Split up large auth method into multiple methods

kraih:
Small followup to #3757. Just splitting up the large auth method and changing the log level consistently to debug.


Martchus (approved):

okurz (approved):
codecov already explained what is missing :)

PR #3770, opened 2021-03-08
Add auth provider column to users table to support multiple providers later

Martchus:
This helps with https://progress.opensuse.org/issues/89020 and the existing
PR https://github.com/os-autoinst/openQA/pull/3769.

Note that this change will not improve support for multiple auth providers.
It merely contains the database migration.


okurz (approved):

kraih (approved):

Martchus:
The OBS build has actually succeeded now.

PR #3771, opened 2021-03-08
Document how to take out worker slots without interrupting running jobs

Martchus:


okurz (approved):

docs/Installing.asciidoc
```diff
@@ -612,6 +612,18 @@ invokes the command above (for the specified slot) whenever the worker configura
 under `/etc/openqa/workers.ini` changes. This unit is not enabled by default and
 only affects `openqa-worker-auto-restart@.service` but not other worker services.
 
+This kind of setup makes it easy to take out worker slots temporarily without
```
kalikiana, line 615:
```suggestion
This kind of setup makes it easy to disable worker slots temporarily without
```

Same as a service you enable/disable. At least to my mind that's a more natural way to phrase it.

Martchus: I would actually avoid this phrasing because it could easily be confused with "disabling" the service in the systemd-sense. However, the service will actually not be "disabled" but "masked" and also be stopped (gracefully).

Martchus:
The OBS build has actually succeeded now.

docs/Installing.asciidoc
kalikiana: Maybe call it mask then? I honestly don't know what I would expect "take out" to mean here on its own

Martchus: It means stopping them without interrupting currently running jobs. Likely you've just read the diff and not the sections heading. I suppose if you consider the section heading (and previous paragraphs) it wouldn't be hard to make sense of it.

kalikiana: I'm not saying I can't make sense of it, I'm saying the phrasing makes no sense without investigating what it could mean or asking here

kalikiana: You could refute my concerns by quoting other docs we have using similar terms if there are any, or systemd or whatever else. But since this is documentation I simply wonder if other readers will also be wondering what it means

kalikiana: You can also ignore my concerns if you disagree. But I hope it's clear that I'm not asking for an out of band explanation here

PR #3772, opened 2021-03-09
Add static Dockerfile checks as in os-autoinst

okurz:
Related progress issue: https://progress.opensuse.org/issues/88451

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -8,6 +8,9 @@ KEEP_DB ?= 0
 # TESTS: Specify individual test files in a space separated lists. As the user
 # most likely wants only the mentioned tests to be executed and no other
 # checks this implicitly disables CHECKSTYLE
+# CONTAINER_TEST: Set to 0 to exclude container tests needing a container
+# runtime environment
+CONTAINER_TEST ?= 1
 TESTS ?=
 ifeq ($(TESTS),)
 PROVE_ARGS ?= --trap -r -v
@@ -242,6 +245,9 @@ prepare-and-launch-docker-to-run-tests-within: docker-test-build launch-docker-t
 # all additional checks not called by prove
 .PHONY: test-checkstyle-standalone
 test-checkstyle-standalone: test-shellcheck test-yaml test-critic test-js-style
+ifeq ($(CONTAINER_TEST),1)
+test-checkstyle-standalone: test-check-containers
+endif
 
 .PHONY: test-critic
 test-critic:
@@ -275,6 +281,10 @@ test-js-style: check-js-beautify
 tidy-js: check-js-beautify
 	@# Fall back to find if there is no git, e.g. in package builds
 	for i in $$(git ls-files "*.js" 2>/dev/null || find assets/javascripts/ -name '*.js'); do js-beautify ${JSBEAUTIFIER_OPTS} $$i > $$i.new; mv $$i.new $$i; done
+
+.PHONY: test-check-containers
+test-check-containers:
+	tools/static_check_containers
 
 .PHONY: tidy
 tidy: tidy-js
--- a/tools/static_check_containers
+++ b/tools/static_check_containers
@@ -1,0 +1,15 @@
+#!/bin/bash
+
+set -euo pipefail
+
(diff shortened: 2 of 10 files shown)


.github/workflows/ci.yml
```diff
@@ -0,0 +1,12 @@
+---
```
kalikiana, line 1: I suggest to use a proper name for the workflow i.e. `hadolint` and a filename like `hadolint.yml`

.github/workflows/ci.yml
```diff
@@ -0,0 +1,12 @@
...
+name: ci
+# yamllint disable-line rule:truthy
+on: [push, pull_request]
+jobs:
+  static-check-containers:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - name: Test our container definitions
+        run: tools/static_check_containers
```
kalikiana, line 11: Please make this a makefile target e.g. `test-containers-hadolint` as discussed on #3761 to make it trivial to reproduce locally

tools/static_check_containers
```diff
@@ -0,0 +1,14 @@
+#!/bin/bash
+
+set -euo pipefail
+
+# use github internal registry if available, e.g. within github actions
+if [ "$(curl -s -w "%{http_code}" https://ghcr.io/v2/ -o /dev/null)" = "401" ]; then
+    img=hadolint/hadolint
+else
+    img=ghcr.io/hadolint/hadolint
```
kalikiana, line 9: I don't know if/why the image thing is useful (it's not used by the GHA). But this script should probably be moved to the "common" repo then

kalikiana (changes requested):

tools/static_check_containers
okurz: On GHA ghcr.io/hadolint/hadolint should be used to save bandwidth on GHA side and hence runtime for us

okurz:
https://github.com/okurz/openQA/runs/2066090651?check_suite_focus=true shows the successful run.

tools/static_check_containers
kalikiana: Maybe a comment could be added to mention that. I was just surprised since "the" GHA for hadolint doesn't seem to be using it and other uses of it I came across didn't. If this makes things faster, awesome 👍🏾

Makefile
```diff
@@ -8,6 +8,9 @@ KEEP_DB ?= 0
 # TESTS: Specify individual test files in a space separated lists. As the user
 # most likely wants only the mentioned tests to be executed and no other
 # checks this implicitly disables CHECKSTYLE
+# CONTAINER_TEST: Set to 0 to exclude container tests needing a container
+# runtime environment
+CONTAINER_TEST ?= 1
```
kalikiana, line 13: Since you're adding this flag I'm thinking a message like "Set CONTAINER_TEST=0 to skip container checks" would be good to have. Otherwise it won't be easy to find. And this would be similar to other flags we have for heavy tests etc.

tools/static_check_containers
okurz: Well, the run in GHA would automatically fallback to "hadolint/hadolint" so we don't ensure the one on ghcr.io is actually used. Better to double-check

Makefile
kalikiana: CHECKSTYLE is different since you don't need to set it. You can use the targets without knowing about it

kalikiana (approved):

Makefile
okurz: Yes, sure. I selected this approach for CONTAINER_TEST with the default to be enabled because `make test` would still try to run all tests, including the container checks. We only exclude that from circleCI because we explicitly wanted to run more in GHA and not all in circleCI.

ilausuch (approved):

Martchus:
Shouldn't a GitHub action now show up in the list of CI checks?

perlpunk:
> Shouldn't a GitHub action now show up in the list of CI checks?

The first PR in a repo without an existing github workflow will only run the test if the branch is in the original repository, not in a fork.

PR #3773, opened 2021-03-09
docker: Allow docker-compose build images

ilausuch:
Now docker-compose is able to build the images before try to start
the containers

https://progress.opensuse.org/issues/89719

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -135,6 +135,9 @@ test: test-with-database
 else
 test: test-checkstyle-standalone test-with-database
 endif
+ifeq ($(CONTAINER_TEST),1)
+test: test-containers-compose
+endif
 endif
 
 .PHONY: test-checkstyle
@@ -294,6 +297,10 @@ test-check-containers:
 .PHONY: tidy
 tidy: tidy-js
 	tools/tidy
+
+.PHONY: test-containers-compose
+test-containers-compose:
+	tools/test_containers_compose
 
 .PHONY: update-deps
 update-deps:
--- a/dist/rpm/openQA.spec
+++ b/dist/rpm/openQA.spec
@@ -290,6 +290,10 @@ rm \
 # within CI systems, e.g. OBS. See t/lib/OpenQA/Test/TimeLimit.pm
 export CI=1
 export OPENQA_TEST_TIMEOUT_SCALE_CI=10
+# Skip container tests that would need additional requirements, e.g.
+# docker-compose. Also, these tests are less relevant (or not relevant) for
+# packaging
+export CONTAINER_TEST=0
 make test PROVE_ARGS='-r -v' CHECKSTYLE=0 TEST_PG_PATH=%{buildroot}/DB
 rm -rf %{buildroot}/DB
 %endif
--- a/tools/test_containers_compose
+++ b/tools/test_containers_compose
@@ -1,0 +1,12 @@
+#!/bin/bash
+
+set -euo pipefail
+
+for i in webui; do
+  (
+    cd container/$i/ &&
+    sudo docker-compose build -q --parallel &&
+    sudo docker-compose up -d &&
+   (docker-compose ps --services --filter status=stopped | grep "^[[:space:]]*$") || (docker-compose logs; sudo docker-compose ps; exit 1)
+  ) || exit 1;
+done
--- a/.github/workflows/compose.yml
+++ b/.github/workflows/compose.yml
@@ -1,0 +1,12 @@
+---
+name: compose
+# yamllint disable-line rule:truthy
+on: [push, pull_request]
+jobs:
+  webui-docker-compose:
+    runs-on: ubuntu-latest
+    timeout-minutes: 30
+    steps:
+      - uses: actions/checkout@v2
+      - name: Verify that containers can be composed
+        run: make test-containers-compose
--- a/container/webui/docker-compose.yaml
+++ b/container/webui/docker-compose.yaml
@@ -2,6 +2,7 @@ version: '3.8'
 services:
   data:
     image: openqa_data
+    build: ../openqa_data/
     restart: always
     volumes:
       - ./workdir/data/factory:/data/factory
@@ -22,6 +23,7 @@ services:
 
   websockets:
     image: openqa_webui
+    build: .
     ports:
       - "9527:9527"  # worker connection
     volumes:
@@ -37,6 +39,7 @@ services:
 
   gru:
     image: openqa_webui
+    build: .
     volumes:
       - ./workdir/data/factory:/data/factory
       - ./workdir/data/tests:/data/tests
@@ -49,6 +52,7 @@ services:
 
   livehandler:
     image: openqa_webui
+    build: .
     ports:
       - "9528:9528"  # handle live view
     volumes:
@@ -64,6 +68,7 @@ services:
 
   webui:
     image: openqa_webui
+    build: .
     ports:
       - "9526"
     volumes:
@@ -91,6 +96,9 @@ services:
 
   nginx:
     image: openqa_webui_lb
+    build:
+      context: .
+      dockerfile: Dockerfile-lb
     restart: always
     ports:
       - "9526:9526"


kalikiana:
> > Still wondering: will this PR have CI as discussed before?
> 
> The CI test in #3758 checks this problem too.

So how would that get merged, though?

ilausuch:
> > > Still wondering: will this PR have CI as discussed before?
> > 
> > 
> > The CI test in #3758 checks this problem too.
> 
> So how would that get merged, though?

But this Test is not going to work without this, and this cannot be checked without this test... then, what is the best solution?

kalikiana:
> > > > Still wondering: will this PR have CI as discussed before?
> > > The CI test in #3758 checks this problem too.
> > So how would that get merged, though?
> But this Test is not going to work without this, and this cannot be checked without the test... then, what is the best solution?

I'll make it simple. I'm closing my PR. You can grab the commit from there.

Makefile
```diff
@@ -135,7 +135,7 @@ endif
 endif
 
 .PHONY: test-checkstyle
-test-checkstyle: test-checkstyle-standalone test-tidy-compile
+test-checkstyle: test-checkstyle-standalone test-tidy-compile test-containers-compose
```
okurz, line 138: As I learned from cdywan the compose test is a dynamic, functional test, not a static style check so please include that not as part of "checkstyle" but as part of the dynamic tests, e.g. above in line 131+133

okurz (changes requested):

kalikiana (approved):
I assume this should be good, provided CI passes. Although I fear only @ilausuch will be able to confirm GHA runs until this hits master and I've not found a way to fix that.

ilausuch:
> I assume this should be good, provided CI passes. Although I fear only @ilausuch will be able to confirm GHA runs until this hits master and I've not found a way to fix that.

Does it can be tested previously to be merged? I mean, the code of docker-compose up -d seems good but not sure about the CI related code

kalikiana:
> > I assume this should be good, provided CI passes. Although I fear only @ilausuch will be able to confirm GHA runs until this hits master and I've not found a way to fix that.
> 
> Does it can be tested previously to be merged? I mean, the code of docker-compose up -d seems good but not sure about the CI related code

You should get a notification if it fails. Conversely if you didn't it *should* be passing afaik.

Makefile
```diff
@@ -128,7 +128,7 @@ ifeq ($(TRAVIS),true)
 test: run-tests-within-container
 else
 ifeq ($(CHECKSTYLE),0)
-test: test-with-database
+test: test-with-database test-containers-compose
```
okurz, line 131: It does not make sense to only call this when skipping checkstyle

okurz (changes requested):

Makefile
ilausuch: Modified, I think now is the correct place

okurz: No, sorry. It also does not make sense to only call this when checkstyle is included. have you seen my comment in https://github.com/os-autoinst/openQA/pull/3773#pullrequestreview-607477161 ? Please read carefully that I mentioned there *two* lines. However this will still not be clean. This is why there is also https://github.com/os-autoinst/openQA/pull/3772/files which cdywan has invited you to review two days ago. I suggest we finish https://github.com/os-autoinst/openQA/pull/3772 first and then extend here.

okurz:
please rebase after https://github.com/os-autoinst/openQA/pull/3772 merged

Makefile
```diff
@@ -251,7 +251,7 @@ prepare-and-launch-docker-to-run-tests-within: docker-test-build launch-docker-t
 .PHONY: test-checkstyle-standalone
 test-checkstyle-standalone: test-shellcheck test-yaml test-critic test-js-style
 ifeq ($(CONTAINER_TEST),1)
-test-checkstyle-standalone: test-check-containers
+test-checkstyle-standalone: test-check-containers test-containers-compose
```
okurz, line 254: still wrong. Again, `test-containers-compose` is not a "style check"

okurz (changes requested):

Makefile
kalikiana: May I recommend, on a general note, that you try and suggest what the correct place is or how to identify it? Instead of saying "wrong" repeatedly.

okurz: https://github.com/os-autoinst/openQA/pull/3773#discussion_r592213208

kalikiana: > not as part of "checkstyle" but as part of the dynamic tests, e.g. above in line 131+133

I guess that refers to the first two variants of the `test:` target 🤔 Should both be guarded by `ifeq ($(CONTAINER_TEST),1)`?

okurz: I don't know exactly how it should look like but it should be clear that we want to run the functional, dynamic container tests as part of complete test runs as well as in exactly one circleCI test job while not calling it a "style check". If you want a complete description in code how it would look like I would be better off doing the change myself which I don't want to do unless you ask me because previously there were also complains when I did something like this.

kalikiana:
```make
.PHONY: test        
ifeq ($(TRAVIS),true)
test: run-tests-within-container
else
ifeq ($(CHECKSTYLE),0)
test: test-with-database
else                               
test: test-checkstyle-standalone test-with-database
endif
ifeq ($(CONTAINER_TEST),1)
test: test-containers-compose
endif
endif
```

I **think** this is what we want, i.e. add the new target to `test` if `CONTAINER_TEST` is set, but outside the `TRAVIS` guard - not tested, though

ilausuch: Thanks, changed

okurz: Not resolved though. As long as test-containers-compose is called as part of style checks I will still consider this wrong

Makefile
```diff
@@ -135,6 +135,9 @@ test: test-with-database
 else
 test: test-checkstyle-standalone test-with-database
 endif
+ifeq ($(CONTAINER_TEST),1)
+test: test-containers-compose
+endif
```
okurz, line 140: Why do you want to exclude that from the "TRAVIS==true" branch?

Makefile
ilausuch: > I don't know exactly how it should look like but it should be clear that we want to run the functional, dynamic container tests as part of complete test runs as well as in exactly one circleCI test job while not calling it a "style check". If you want a complete description in code how it would look like I would be better off doing the change myself which I don't want to do unless you ask me because previously there were also complains when I did something like this.

Please, go ahead

Makefile
kalikiana: That was my suggestion above. As far as I can see TRAVIS is supposed to run `run-tests-within-container` which is a script that spawns a container.

okurz: Yeah, we don't really know if this would work at all. I guess it does not matter either way.

okurz: The above looks ok but you should also set the default of CONTAINER_TEST=1

Makefile
okurz: Just remove this one-line change here and it's ok

okurz (changes requested):

ilausuch:
About the CONTAINER_TEST default to 1, is already in the code https://github.com/os-autoinst/openQA/blob/c6677f1db58846327ba64f0d5ee841f3691bc589/Makefile#L13 I am not wrong

okurz (approved):
Yes, sorry about your "default" comment. I thought I had removed that before posting. Of course it's there already

okurz:
have you seen the error on OBS?:

```
[ 1378s] for i in container/webui/docker-compose.yaml; do sudo docker-compose up -d -f $i;  done
[ 1378s] /bin/sh: line 1: sudo: command not found
```

ilausuch:
The new commit only is to test the OBS problem removing the sudo command

kalikiana:
> The purpose of the new commit is only to test the OBS problem removing the sudo command. It works will be stashed

The dependencies aren't actually installed, so it fails. Maybe we should just skip container tests on OBS in general?

```
[ 1812s] for i in container/webui/docker-compose.yaml; do docker-compose up -d -f $i;  done
[ 1812s] /bin/sh: docker-compose: command not found
```

kalikiana:
Also this, but I assume it's unrelated, just mentining it for completeness and in case it pops up somewhere else.

```
t/ui/25-developer_mode.t ................... 4/? 
    #   Failed test '#developer-panel visible'
```

okurz:
I also consider the tests on OBS less relevant for OBS so I suggest to skip them there. This is why we have the possibility to control the tests with build flags, e.g. just set CONTAINER_TEST=0 in the spec file explaining why they are not enabled

ilausuch:
Before disabling, we could test install the dependencies, couldn't we?

kalikiana:
> Before disabling, we could test install the dependencies, couldn't we?

What would the test do on OBS that it doesn't do in the GHA? It seems redundant to me.

kalikiana:
> I also consider the tests on OBS less relevant for OBS so I suggest to skip them there. This is why we have the possibility to control the tests with build flags, e.g. just set CONTAINER_TEST=0 in the spec file explaining why they are not enabled

Discussed this a bit with @ilausuch to clarify GHA vs OBS vs other stuff.

"OBS" in this context means `openQA.spec` which uses rpm macros to define the test setup. Something like `ag CI=1` would show you where this is done. And what Oli suggested is to add `export CONTAINER_TEST=0` there to avoid executing these tests. At the same time, we want the default `make test` to run these.

okurz:
I pushed https://github.com/os-autoinst/openQA/pull/3773/commits/65150683aaf74dd07ae1fa4eb17c2cc3d6dd1716 to exclude the container tests from packaging tests

okurz:
Can you help me find the latest github action run including the new rules?

okurz (changes requested):
please rebase to get rid of the "merge" commit. Maybe https://medium.com/singlestone/a-git-workflow-using-rebase-1b1210de83e5 helps

Martchus (changes requested):
Don't merge things locally. Please use `git rebase -i origin/master` and delete all unrelated commits in the editor prompt.

okurz (commented):
the title still says "WIP", is it ready? Then please update the title

okurz (changes requested):
https://github.com/os-autoinst/openQA/pull/3773/checks?check_run_id=2129340678#step:3:4 shows just the help output and unfortunately seems to still return an exit code of success. I suggest you first ensure that a problem in the docker-compose test makes the github actions step actually report that as failure

dist/rpm/openQA.spec
```diff
@@ -290,6 +290,10 @@ rm \
 # within CI systems, e.g. OBS. See t/lib/OpenQA/Test/TimeLimit.pm
 export CI=1
 export OPENQA_TEST_TIMEOUT_SCALE_CI=10
+# Skip container tests that would need additional requirements, e.g.
+# docker-compose. Also, these tests are less relevant (or not relevant) for
+# packaging
+export CONTAINER_TEST=0
```
kalikiana, line 296: I would've said "tests we already run via GHA" 🤔 what is less relevant?

okurz: Well, my point was that building a container is actually a different use case than installing packages. Of course we use RPM packages within the containers but that's already tested itself as well.

okurz:
https://github.com/os-autoinst/openQA/pull/3773/checks?check_run_id=2129505864 is passed now. Looks acceptable. Can you show https://github.com/os-autoinst/openQA/pull/3773#pullrequestreview-614093753 . Also the "WIP" in the title is left

Makefile
```diff
@@ -295,6 +298,10 @@ test-check-containers:
 tidy: tidy-js
 	tools/tidy
 
+.PHONY: test-containers-compose
+test-containers-compose:
+	for i in $(wildcard container/**/docker-compose.yaml); do cd "$$(echo $$i | cut -d "/" -f-2)"; sudo docker-compose up -d || exit 1; cd ../../; done
```
perlpunk, line 303: I would avoid the `cd ../../` by doing the `cd` in a subprocess:

`(cd ...; sudo ...)`

ilausuch:
> [#3773 (checks)](https://github.com/os-autoinst/openQA/pull/3773/checks?check_run_id=2129505864) is passed now. Looks acceptable. Can you show [#3773 (review)](https://github.com/os-autoinst/openQA/pull/3773#pullrequestreview-614093753) . Also the "WIP" in the title is left

I am not sure, because the logs stop of the job "compose" stops in a point that I didn't expected
In my environment I see that after "Verify: matching certificate signature" there are more lines
```
Verify: matching certificate signature
/etc/apache2/ssl.crt/df739aa4f868-server.crt: OK
....
```
This is part of the build of the image, but I missing the standup of the openQA in these logs

Makefile
perlpunk, line 303:
```suggestion
	for i in $(dir $(wildcard container/**/docker-compose.yaml)); do (cd $$i; sudo docker-compose up -d || exit 1); done
```

okurz: Not sure this works as expected though. I think this will only exit the subshell but not stop the execution of the loop.

`(cd $$i; sudo docker-compose up -d || exit 1) || exit 1` might work though

perlpunk: ah right. this should do it:
```
do (cd $$i; sudo docker-compose up -d) || exit 1
```

ilausuch: Thanks, seems work perfectly

Makefile
```diff
@@ -295,6 +298,10 @@ test-check-containers:
 tidy: tidy-js
 	tools/tidy
 
+.PHONY: test-containers-compose
+test-containers-compose:
+	for i in $(dir $(wildcard container/**/docker-compose.yaml)); do (cd $$i; sudo docker-compose up -d) || exit 1; done
```
Martchus, line 303: ~~I'm wondering about the `|| exit 1`. It seems useless unless you really wanted to enforce the exit code `1` in favor of any other non-zero exit code. Or did you wanted to break the loop here? That's not possible this way because `(…)` forks and the `exit` only affects the fork.~~ Ok, I see that you've just updated the commit so it should be fine now.

---

By the way, **if the `sudo` version in this env is new enough**:
```suggestion
	for i in $(dir $(wildcard container/**/docker-compose.yaml)); do sudo --chdir="$$i" docker-compose up -d; done
```

Martchus (approved):

Makefile
okurz: You could check it out if it works by using a container environment using the same image. I suggest to reserve this change for later though. Let's focus on getting the PR in quickly

ilausuch:
I think now we have a good green check on compose job

okurz (approved):
Ok, looks good. I am only still missing a verification that the github action check would actually fail in case of a failed test. So please still provide that regardless if before or after merge

kalikiana:
> Ok, looks good. I am only still missing a verification that the github action check would actually fail in case of a failed test. So please still provide that regardless if before or after merge

Screenshot from 2021-03-17 15-42-06

Seems to [look good](https://github.com/os-autoinst/openQA/pull/3773/checks?check_run_id=2131044953)!

okurz:
@kalikiana You seem to misunderstand. I can see the "passed" check for a test run that looks successful. But I wonder if there would be an actual "failed" check for a test run that has a problem.

ilausuch:
> Ok, looks good. I am only still missing a verification that the github action check would actually fail in case of a failed test. So please still provide that regardless if before or after merge

Is not trivial

There is an option to start the docker-compose an forces an exit and error code using --abort-on-container-exit but if all works, the docker-compose never finish. So we have to continue using -d. I am testing that here:

https://github.com/os-autoinst/openQA/pull/3796

ilausuch:
I applied the same modifications here to check when all is right

okurz:
https://github.com/os-autoinst/openQA/pull/3796/checks?check_run_id=2132109256 shows a failure correctly detected but now https://github.com/os-autoinst/openQA/pull/3773/checks?check_run_id=2132326640#step:3:41 fails as well. Is this what you expected?

ilausuch:
> [#3796 (checks)](https://github.com/os-autoinst/openQA/pull/3796/checks?check_run_id=2132109256) shows a failure correctly detected but now [#3773 (checks)](https://github.com/os-autoinst/openQA/pull/3773/checks?check_run_id=2132326640#step:3:41) fails as well. Is this what you expected?

I didn't expected a failure, in my dev env works, I am adding more logs to see what is the problem

tools/test_containers_compose
```diff
@@ -0,0 +1,14 @@
+#!/bin/bash
+
+set -euo pipefail
+
+for i in webui; do
+  (
+    cd container/$i/ && \
+    sudo docker-compose build -q --parallel && \
+    sudo docker-compose up -d && \
+    sleep 30 && \
```
okurz, line 10: As usual I have my problems with approving something that introduces sleep with more than 1 second :)

okurz (changes requested):

tools/test_containers_compose
ilausuch: Ok, I check if is possible to remove that

ilausuch: But the reason for that is because maybe we have an Up in our containers in the the second 1 but not 10 seconds later. And this is because they do some work when are Up, for instance standing up the DB

ilausuch: As talked in a conversation we'll consider these case in an other ticket

ilausuch:
The test is good now, this is ready for review. thanks

ilausuch:
No, there is a problem confirmed by https://github.com/os-autoinst/openQA/pull/3796

ilausuch:
Ok, solved, please review

okurz:
In detail, https://github.com/os-autoinst/openQA/pull/3796/checks?check_run_id=2139502104 shows how it's correctly handled in case of test failures. Looks good!

okurz (approved):

Martchus (approved):

PR #3774, opened 2021-03-09
Personal access tokens need to compare the nickname

kraih:
Depending on how the user was created the username column can be very different from the nickname returned by the `->name` method. For example `https://www.opensuse.org/openid/user/kraih` instead of the expected `kraih`.

Progress: https://progress.opensuse.org/issues/87698

### DIFF
--- a/t/api/03-auth.t
+++ b/t/api/03-auth.t
@@ -180,7 +180,7 @@ subtest 'personal access token' => sub {
     $t->delete_ok('/api/v1/assets/1')->status_is(403)->json_is({error => 'no api key'});
 
     # Valid access token
-    $t->$userinfo('arthur:ARTHURKEY01:EXCALIBUR')->delete_ok('/api/v1/assets/1')->status_is(404);
+    $t->$userinfo('artie:ARTHURKEY01:EXCALIBUR')->delete_ok('/api/v1/assets/1')->status_is(404);
 
     # Invalid access token
     $t->$userinfo('invalid:invalid')->delete_ok('/api/v1/assets/1')->status_is(403)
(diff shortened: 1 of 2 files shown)


kraih:
In #3757 i made the mistake to copy the `$user->username` calls from the API key auth debug code, instead of using the `$user->name` call from the template that shows the logged in user in the webui.

t/api/03-auth.t
```diff
@@ -180,7 +180,7 @@ subtest 'personal access token' => sub {
     $t->delete_ok('/api/v1/assets/1')->status_is(403)->json_is({error => 'no api key'});
 
     # Valid access token
-    $t->$userinfo('arthur:ARTHURKEY01:EXCALIBUR')->delete_ok('/api/v1/assets/1')->status_is(404);
+    $t->$userinfo('artie:ARTHURKEY01:EXCALIBUR')->delete_ok('/api/v1/assets/1')->status_is(404);
```
kalikiana, line 183: So... where does `artie` come from I wonder.. and it seems like in `t/fixtures/03-users.pl` this is the `nickname` of the same user. Maybe the test should use `lancelot` as well, since that has `https://openid.camelot.uk/lancelot` for the `username`?
In `Auth.pm` you're checking `->name` now so that means `username unless nickname` I guess.

Martchus (approved):

t/api/03-auth.t
kraih: It's actually more complicated than that, `->name` is not just a field but [a method with logic](https://github.com/os-autoinst/openQA/blob/0b1531c352c415d0d8fb3d20cf22284ed193fd92/lib/OpenQA/Schema/Result/Users.pm#L73). The main point here is that i believe the personal access token user should be identical to the one we show in the navbar after login, and that is whatever `->name` returns.

kalikiana (approved):

PR #3775, opened 2021-03-09
Adjust priorities of Minion tasks and enable use of "fast lane"

Martchus:
* See particular commit message for details (it is maybe also easier to review the commits individually)
* Related ticket: https://progress.opensuse.org/issues/89224#note-7

### DIFF
--- a/lib/OpenQA/WebAPI/Command/gru/run.pm
+++ b/lib/OpenQA/WebAPI/Command/gru/run.pm
@@ -1,4 +1,4 @@
-# Copyright (C) 2018-2020 SUSE LLC
+# Copyright (C) 2018-2021 SUSE LLC
 #
 # This program is free software; you can redistribute it and/or modify
 # it under the terms of the GNU General Public License as published by
@@ -24,7 +24,7 @@ has usage       => sub { shift->extract_usage };
 sub run {
     my ($self, @args) = @_;
 
-    getopt \@args, 'o|oneshot' => \my $oneshot, 'reset-locks' => \my $reset_locks;
+    getopt \@args, ['pass_through'], 'o|oneshot' => \my $oneshot, 'reset-locks' => \my $reset_locks;
 
     my $minion = $self->app->minion;
     return $minion->perform_jobs if $oneshot;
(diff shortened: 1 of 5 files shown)


kraih:
The feature is on the way to factory now. https://build.opensuse.org/request/show/878032

kraih (approved):
The proposed changes make sense to me.

lib/OpenQA/WebAPI/Command/gru/run.pm
```diff
@@ -24,7 +24,7 @@ has usage       => sub { shift->extract_usage };
 sub run {
     my ($self, @args) = @_;
 
-    getopt \@args, 'o|oneshot' => \my $oneshot, 'reset-locks' => \my $reset_locks;
+    getopt \@args, ['pass_through'], 'o|oneshot' => \my $oneshot, 'reset-locks' => \my $reset_locks;
```
kalikiana, line 27: I don't understand what this does. The commit says "Allow passing CLI options to minion" but nothing else is being added there. That's the only change in the commit besides the copyright 🤔

kraih: Looks a bit weird, but is very sensible here. It's an option of the underlying param parsing module that leaves unknown flags in `@args` to be processed somewhere else. https://docs.mojolicious.org/Mojo/Util#getopt

kraih: I'm actually not sure why it's missing, should have been there already.

Martchus: The commit message means that the change allows to *pass* options through "this layer of abstraction". So additional `@args` which go in are passed to the run function of the base class which then handles them (whatever they are). So you can e.g. specify `--jobs 10` and these args are passed to the Minion worker with no further handling from our side.

kalikiana: Thank you for clarifying! Confusion resolved. 👍🏾

kalikiana (approved):

PR #3776, opened 2021-03-09
Test personal access tokens with an OpenID user

kraih:
Followup to #3774 with a few extra tests.


kalikiana (approved):
Nice. Thanks for extending the test further!

PR #3777, opened 2021-03-10
BuildRequires openSUSE-release only on openSUSE

ggardet:


okurz (approved):

okurz:
Makes absolute sense. Not sure if this needs to be manually maintained in the spec file or if dependencies.yaml needs to reflect this. We will see. Maybe that can be solved even better by replacing "openSUSE" with a variable as I assume there is also a "SLE-release" package

ggardet:
> Makes absolute sense. Not sure if this needs to be manually maintained in the spec file or if dependencies.yaml needs to reflect this. We will see. Maybe that can be solved even better by replacing "openSUSE" with a variable as I assume there is also a "SLE-release" package

There is `sles-release` package, but the question is more: do we want to package the systemd services for SLE as well?

Martchus (approved):
> Not sure if this needs to be manually maintained in the spec file or if dependencies.yaml needs to reflect this.

There's no `# The following line is generated from dependencies.yaml` so I wouldn't expect that. However, I've double checked and it is not necassary to change anything else (`git checkout ggardet/fix_build_on_sle`; `make update-deps` still generates no diff).

PR #3780, opened 2021-03-11
Update apparmor profile for git for openSUSE Tumbleweed

ggardet:
since git moved from /usr/lib/git/ to /usr/libexec/git/

More information at https://en.opensuse.org/openSUSE:Packaging_Conventions_RPM_Macros#.25_libexecdir

This is related to https://progress.opensuse.org/issues/89620 and tested locally.


okurz:
Thanks a lot. Haven't thought about that change in Tumbleweed

PR #3781, opened 2021-03-11
[WIP] containers: Add worker container to the docker-compose

ilausuch:
- Worker service is part of the compose workflow
- The number of workers is configurable.

https://progress.opensuse.org/issues/89752


okurz (changes requested):
Let's get tests done first before looking at this

container/webui/docker-compose.yaml
```diff
@@ -79,6 +79,25 @@ services:
...
+      - ./workdir/data/factory/hdd:/data/factory/hdd:rw
+      - ./workdir/data/factory/other:/data/factory/other:rw
+      - ./workdir/data/factory/tmp:/data/factory/tmp:rw
+      - ./workdir/data/testresults:/data/testresults:rw
+      - ./workdir/data/tests:/data/tests:ro
+      - ./conf:/data/conf:ro
+    depends_on:
+      - nginx
+    deploy:
+      replicas: ${OPENQA_WORKER_REPLICAS}
```
kalikiana, line 99: Would it make sense to have a `healthcheck` here? Something to see if the container came up 🤔

ilausuch: Done

kalikiana: I'll wait for #3773 since it's needed for the *healthcheck* to be run in CI. Please remind me to re-review once this PR has the check running

ilausuch:
I only did a rebase from master as a experiment to see how this works with the new CI test for containers

container/webui/docker-compose.yaml
```diff
@@ -120,6 +120,32 @@ services:
...
+    devices:
+      - /dev/kvm
+    volumes:
+      - ./workdir/data/factory/iso:/data/factory/iso:rw
+      - ./workdir/data/factory/hdd:/data/factory/hdd:rw
+      - ./workdir/data/factory/other:/data/factory/other:rw
+      - ./workdir/data/factory/tmp:/data/factory/tmp:rw
+      - ./workdir/data/testresults:/data/testresults:rw
+      - ./workdir/data/tests:/data/tests:ro
+      - ./conf:/data/conf:ro
```
dcermak, line 136:
```suggestion
    volumes: *volumes
```

Martchus:
With https://github.com/os-autoinst/openQA/pull/4160 I improved the web UI's and the worker's docker compose files so one can use them together. However, I refrained from adding the worker to the web UI docker-compose file itself because I don't think that's something we generally like to have. It should be sufficient if both docker-compose setups can work together and see no need to tie them together that closely. So I'm inclined to close this PR which is now in conflict with other changes anyways.

PR #3782, opened 2021-03-11
Package auto-update part on openSUSE only

ggardet:
This is a follow-up of https://github.com/os-autoinst/openQA/pull/3777

Build tested for SLE15-SP2/ https://build.opensuse.org/package/show/home:Guillaume_G:branches:devel:openQA/openQA

### DIFF
--- a/dist/rpm/openQA.spec
+++ b/dist/rpm/openQA.spec
@@ -242,6 +242,7 @@ Group:          Development/Tools/Other
 Documentation material covering installation, configuration, basic test writing, etc.
 Covering both openQA and also os-autoinst test engine.
 
+%if 0%{?is_opensuse}
 %package auto-update
 Summary:        Automatically upgrade and reboot the system when required
 Group:          Development/Tools/Other
@@ -251,6 +252,7 @@ Requires:       rebootmgr
 %description auto-update
 Use this package to install and enable a systemd service for nightly upgrading
 and rebooting the system if devel:openQA packages are stable.
+%endif
 
 %prep
 %setup -q -a1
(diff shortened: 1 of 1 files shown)


Martchus (commented):
I'm wondering how installing this package is causing the additional files to be installed. We have

```
os := $(shell grep opensuse /etc/os-release)
.PHONY: install
ifeq ($(os),)
install: install-generic
else
install: install-opensuse
endif
```
within the `Makefile` but it only checks for `opensuse`.

ggardet:
> I'm wondering how installing this package is causing the additional files to be installed. We have
> 
> ```
> os := $(shell grep opensuse /etc/os-release)
> .PHONY: install
> ifeq ($(os),)
> install: install-generic
> else
> install: install-opensuse
> endif
> ```
> 
> within the `Makefile` but it only checks for `opensuse`.

Indeed, my local test pointed this out as well. I will then skip the `auto-update` part on SLE.

Martchus (approved):
I'm fine with skipping these systemd files. Alternatively one could also invoke the `install-opensuse` target explicitly or extend the `grep` command to cover SLE as well.

okurz:
Do you have a link on OBS for verification?

ggardet:
> Do you have a link on OBS for verification?

It was a local build. Just did it on OBS: https://build.opensuse.org/package/show/home:Guillaume_G:branches:devel:openQA/openQA

ggardet:
> I'm fine with skipping these systemd files. Alternatively one could also invoke the `install-opensuse` target explicitly or extend the `grep` command to cover SLE as well.

I guess this was on purpose to differentiate openSUSE from other targets.

PR #3783, opened 2021-03-11
container: Fix missing shared directories and its permissions

ilausuch:
- Check and complete with the missing directories
The directories 'tmp' and 'others' were missing.
Added all the required directories as volumes.

- Check and fix the permissions over these directories for the gekotest user
Because this directories are shared between all the containers and also
with the host. This PR assign the write permissions for the directories that
the containers write.

https://progress.opensuse.org/issues/89755

### DIFF
--- a/container/webui/docker-compose.yaml
+++ b/container/webui/docker-compose.yaml
@@ -1,24 +1,20 @@
 version: '3.8'
-services:
-  data:
-    image: openqa_data
-    build: ../openqa_data/
-    restart: always
-    volumes:
-      - ./workdir/data/factory:/data/factory
-      - ./workdir/data/tests:/data/tests
 
+services:
   scheduler:
     image: openqa_webui
-    volumes:
-      - ./workdir/data/factory:/data/factory
-      - ./workdir/data/tests:/data/tests
-      - ./conf:/data/conf
+    volumes: &volumes
+      - ./workdir/data/factory/iso:/data/factory/iso:rw
+      - ./workdir/data/factory/hdd:/data/factory/hdd:rw
+      - ./workdir/data/factory/other:/data/factory/other:rw
+      - ./workdir/data/factory/tmp:/data/factory/tmp:rw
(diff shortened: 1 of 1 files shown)


okurz:
please rebase after https://github.com/os-autoinst/openQA/pull/3772 merged

ilausuch:
rebased

okurz (changes requested):
Quite a lot of directories to maintain manually. What would be needed to reduce the duplication?

perlpunk:
> Quite a lot of directories to maintain manually. What would be needed to reduce the duplication?

The directory list seems to be the same in every case, so it can probably be replaced with an `&anchor` and aliases.

ilausuch:
@perlpunk Good point. Done

okurz (approved):
Looks good so far. Please keep this open until we have CI tests for docker-compose merged.

ilausuch:
Since https://github.com/os-autoinst/openQA/pull/3773 is merged, I rebased to ensure that this is code is effectively checked on the CI

container/webui/docker-compose.yaml
```diff
@@ -1,88 +1,70 @@
...
+    volumes: *volumes
     environment:
       MODE: "webui"
       MOJO_LISTEN: "http://0.0.0.0:9526"
     depends_on:
-      - data
       - db
     deploy:
       replicas: ${OPENQA_WEBUI_REPLICAS}
+    entrypoint: "sh -c 'chmod -R a+rw /data/{factory,testresults}; /root/run_openqa.sh'"
```
Martchus, line 66: Those are directories so it would make sense to add `rwx` permissions instead of just `rw`.

okurz: Better use `a+rwX`

ilausuch: done

ilausuch:
This is ready for your review. Thanks

Martchus (approved):

PR #3784, opened 2021-03-11
Wait for the element instead of constantly refreshing the page

kraih:
This is a followup to #3732 and should further stabilize the test results. Instead of just doing `find_element`/`sleep` we now use `wait_for_element`, like most other ui tests.

Progress: https://progress.opensuse.org/issues/89935

### DIFF
--- a/t/ui/27-plugin_obs_rsync_status_details.t
+++ b/t/ui/27-plugin_obs_rsync_status_details.t
@@ -17,6 +17,7 @@ use Test::Most;
 
 use FindBin;
 use lib "$FindBin::Bin/../lib", "$FindBin::Bin/../../external/os-autoinst-common/lib";
+use Mojo::Base -signatures;
 use Test::Warnings;
 use Test::Mojo;
 use OpenQA::Test::TimeLimit '60';
@@ -52,8 +53,7 @@ sub fake_api_server {
     $mock->mode('test');
 
     $mock->routes->get(
-        '/public/build/:proj/_result' => sub {
-            my $c       = shift;
+        '/public/build/:proj/_result' => sub ($c) {
             my $proj    = $c->stash('proj');
             my $package = $c->param('package') // '';
             return $c->render(
@@ -124,24 +124,24 @@ my %params = (
     'Batch1'            => ['191216_150610', 'containers', 'BatchedProj', '470.2, 469.1'],
 );
 
-sub _wait_helper {
-    my ($element, $test_break, $refresh) = @_;
-    my $ret;
-    for (0 .. 3) {
-        for (0 .. 30) {
-            $ret = $driver->find_element($element)->get_text();
-            return $ret if $test_break->($ret);    # uncoverable statement
-            sleep .1;                              # uncoverable statement
-        }
+sub _wait_for_change ($selector, $break_cb, $refresh_cb = undef) {
+    my $text;
+    my $limit = int OpenQA::Test::TimeLimit::scale_timeout(10);
+    for my $i (0 .. $limit) {
+
         # sometimes gru is not fast enough, so let's refresh the page and see if that helped
-        if ($refresh) {                            # uncoverable statement
-            $refresh->();                          # uncoverable statement
+        if ($i > 0) {
+            sleep 1;
+            note qq{Refreshing page, waiting for "$selector" to change};
+            $refresh_cb ? $refresh_cb->() : $driver->refresh;
         }
-        else {
-            $driver->refresh();                    # uncoverable statement
-        }
+
+        wait_for_element(selector => $selector);
+        $text = $driver->find_element($selector)->get_text();
+        return $text if $break_cb->(local $_ = $text);
     }
-    return $driver->find_element($element)->get_text();    # uncoverable statement
+
+    BAIL_OUT qq{Wait limit of $limit seconds exceeded for "$selector", no change: $text};
 }
 
 foreach my $proj (sort keys %params) {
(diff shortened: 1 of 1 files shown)


t/ui/27-plugin_obs_rsync_status_details.t
```diff
@@ -124,24 +124,23 @@ my %params = (
     'Batch1'            => ['191216_150610', 'containers', 'BatchedProj', '470.2, 469.1'],
 );
 
-sub _wait_helper {
-    my ($element, $test_break, $refresh) = @_;
+sub _wait_for_change ($selector, $break_cb, $refresh_cb = undef) {
```
Martchus, line 127: Thanks for improving the variable names as well.

Martchus (commented):
The code looks good but the test fails now:

```
[17:09:42] t/ui/27-plugin_obs_rsync_status_details.t .. 101/? 
#   Failed test 'Timestamp on dirty status is updated'
#   at t/ui/27-plugin_obs_rsync_status_details.t line 211.
#          got: 'published on 2021-03-11 17:10:39 +0000 (standard)'
#     expected: anything else
[17:09:42] t/ui/27-plugin_obs_rsync_status_details.t .. 113/? [17:09:42] t/ui/27-plugin_obs_rsync_status_details.t .. 116/? [17:09:42] t/ui/27-plugin_obs_rsync_status_details.t .. 120/? [17:09:42] t/ui/27-plugin_obs_rsync_status_details.t .. 121/? [17:09:42] t/ui/27-plugin_obs_rsync_status_details.t .. 122/? [17:09:42] t/ui/27-plugin_obs_rsync_status_details.t .. 123/? # Looks like you failed 1 test of 127.
                                                               [17:09:42] t/ui/27-plugin_obs_rsync_status_details.t .. Dubious, test returned 1 (wstat 256, 0x100)
Failed 1/127 subtests 
```

kraih:
Still unstable i guess. Tried it a bunch of times and results were consistent... 😞

kraih:
Maybe the 15 second timeout i used originally was still too small for Circle CI, i've doubled it now to 30. But will re-rerun the test more tomorrow to make sure it's stable.

kraih:
I understand now why the coverage is so flaky though The frequent refreshing of the page hits different parts of the plugin depending on the state of the Minion jobs. So everything changes depending on the timing of the refreshes. Crazy.

t/ui/27-plugin_obs_rsync_status_details.t
```diff
@@ -124,24 +124,23 @@ my %params = (
...
+    for my $i (0 .. 30) {
+
         # sometimes gru is not fast enough, so let's refresh the page and see if that helped
-        if ($refresh) {                            # uncoverable statement
-            $refresh->();                          # uncoverable statement
-        }
-        else {
-            $driver->refresh();                    # uncoverable statement
+        if ($i > 0) {
+            sleep 1;
```
okurz, line 134: previously there was `sleep .1`, now it's `sleep 1`. Can this not unnecessarily increase the wait time?

okurz (changes requested):

Martchus:
> Maybe the 15 second timeout i used originally was still too small for Circle CI, i've doubled it now to 30.

Note that there's also `OpenQA::Test::TimeLimit::scale_timeout` to cope with that problem.

> The frequent refreshing of the page hits different parts of the plugin depending on the state of the Minion jobs. So everything changes depending on the timing of the refreshes.

Maybe it is best to mark lines which are not executed all the time as uncoverable.

t/ui/27-plugin_obs_rsync_status_details.t
kraih: That's actually the point. The background jobs are very slow. For Circle CI i actually had to increase the overall timeout from 15 to 30 seconds max.

kraih:
To be clear, i',m not trying to solve the flaky test coverage in this PR. This PR is only about making the test result stable. Coverage will still be flaky afterwards, i just coincidentally stumbled over the reason why coverage is flaky.

t/ui/27-plugin_obs_rsync_status_details.t
kraih: Since this now uses a `wait_for_element` too, i don't think the 1 second increase has a negative impact. The previous more frequent polling is just not needed anymore because of that. Now the `sleep` is only for waiting for background jobs, not the page to be rendered.

okurz: True, still for an individual test waiting 1 second when it's not needed is wasteful. Is it ok to keep `sleep .1` here even if it's not the same thing we wait for as in before?

kraih: I just tried it (reducing `sleep` to `.1`) and the result is 100% CPU usage in my VM and the test timing out almost all the time.

Martchus: If the scale we're dealing here with is seconds then `sleep 1` makes sense.

Martchus (approved):

t/ui/27-plugin_obs_rsync_status_details.t
kraih: Pretty sure the test is faster with `sleep 1`, because with Selenium refreshing the page less often the background jobs get more CPU cycles.

kraih: Actually, before this patch:
```
ok
All tests successful.
Files=1, Tests=61, 39 wallclock secs ( 0.07 usr  0.01 sys +  3.36 cusr  0.90 csys =  4.34 CPU)
Result: PASS
```
And after:
```
ok
All tests successful.
Files=1, Tests=91, 34 wallclock secs ( 0.09 usr  0.03 sys +  3.17 cusr  0.88 csys =  4.17 CPU)
Result: PASS
```

kraih: A little counter intuitive, but the longer `sleep` is better for test runtime. 😀

okurz: Still, a test runtime of more than 30s for just one tiny feature of openQA is rather horrible so this is why I am looking for ways to improve. Granted, that one little statement obviously isn't the problem :D

kraih:
I've re-run the tests in Circle CI a bunch of times now and it seems stable. Only one failure happened, but that was in another unrelated test `t/ui/13-admin.t`, something about YAML. https://app.circleci.com/pipelines/github/os-autoinst/openQA/5873/workflows/bedb3cb7-fa93-4bb5-a219-cdc62938341c/jobs/55361

t/ui/27-plugin_obs_rsync_status_details.t
kraih: Sorry, i really don't want to work on that plugin to improve performance (and that will be necessary). 😬 More stability will have to be enough for today.

kraih: Maybe worth mentioning, the whole thing is still just a plugin. So nothing prevents us from moving it into a separate GitHub repo (and RPM package).

okurz (approved):

PR #3786, opened 2021-03-12
Revert "Rework symlink CASEDIR and set values to relative path"

okurz:
Reverts os-autoinst/openQA#3712


czerw:
@okurz  Thx a lot for quick action.

PR #3787, opened 2021-03-12
containers: Fix worker container entrypoint script

ilausuch:
Problem:
The current entrypoint.sh is generating this error
groupmod GID '0' already exists
    
Solution: Remove the group number in groupmod
    
Additionally
- Add qemu_no_kvm flag
- Provided a test for qemu
    
https://progress.opensuse.org/issues/88754

### DIFF
--- a/container/worker/run_openqa_worker.sh
+++ b/container/worker/run_openqa_worker.sh
@@ -1,18 +1,24 @@
 #!/bin/bash
-# shellcheck disable=SC2012
+# shellcheck disable=SC2012,SC2154
 set -e
 
 if [[ -z $OPENQA_WORKER_INSTANCE ]]; then
   OPENQA_WORKER_INSTANCE=1
 fi
 
-/root/qemu/kvm-mknod.sh
+if [[ -z $qemu_no_kvm ]] || [[ $qemu_no_kvm -eq 0 ]]; then
+  if [ -e "/dev/kvm" ] && getent group kvm > /dev/null; then
+    /root/qemu/kvm-mknod.sh
 
-if [ -e "/dev/kvm" ] && getent group kvm > /dev/null; then
-  groupmod -g "$(ls -lhn /dev/kvm | cut -d ' ' -f 4)" kvm
-  usermod -a -G kvm _openqa-worker
-else
-  echo "Warning: /dev/kvm doesn't exist. If you want to use KVM, run the container with --device=/dev/kvm"
+    group=$(ls -lhn /dev/kvm | cut -d ' ' -f 4)
+    groupmod -g "$group" --non-unique kvm
+    usermod -a -G kvm _openqa-worker
+  else
+    echo "Warning: /dev/kvm doesn't exist. If you want to use KVM, run the container with --device=/dev/kvm"
+  fi
 fi
+
+qemu-system-x86_64 -S &
+kill $!
 
 su _openqa-worker -c "/usr/share/openqa/script/worker --verbose --instance \"$OPENQA_WORKER_INSTANCE\""
--- a/container/worker/conf/client.conf
+++ b/container/worker/conf/client.conf
@@ -1,0 +1,3 @@
+[openqa_webui]
+key = 1234567890ABCDEF
+secret = 1234567890ABCDEF
--- a/container/worker/conf/workers.ini
+++ b/container/worker/conf/workers.ini
@@ -1,0 +1,3 @@
+[global]
+BACKEND = qemu
+HOST = http://openqa_webui
--- a/container/worker/docker-compose.yaml
+++ b/container/worker/docker-compose.yaml
@@ -1,0 +1,14 @@
+version: '3.8'
+services:
+  worker:
+    image: openqa_worker
+    build: .
+    privileged: true
+    devices:
+      - /dev/kvm
+    volumes:
+      - ./workdir/data/factory:/data/factory:rw
+      - ./workdir/data/tests:/data/tests:ro
+      - ./conf:/data/conf:ro
+    environment:
+      qemu_no_kvm: 1
--- a/tools/test_containers_compose
+++ b/tools/test_containers_compose
@@ -2,7 +2,7 @@
 
 set -euo pipefail
 
-for i in webui; do
+for i in webui worker; do
   (
     cd container/$i/ &&
     sudo docker-compose build -q --parallel &&


Martchus (changes requested):
The idea of the previous code was to ensure the group ID of `/dev/kvm` and the `kvm` group within the container match because the might be different. This change doesn't cover this anymore so I'd assume using KVM breaks by this unless the group IDs match anyways.

Of course I get what you're trying to fix: If the group ID is already used within the container by another group the current code breaks.

Maybe you can just add the `--non-unique` flag to the `groupmod` command. Otherwise you may be able to change the group ID of the ownership of `/dev/kvm`. (Not sure whether that's possible for devices mapped into a container.) Otherwise you could simply add the user to that other group which already uses the matching ID.

ilausuch:
> --non-unique

Thanks @Martchus I tested locally with that option and seems that works and the execution continues

Martchus (approved):
Since we're only messing around within the container I suppose using `--non-unique` is ok.

container/worker/run_openqa_worker.sh
```diff
@@ -9,7 +9,7 @@ fi
 /root/qemu/kvm-mknod.sh
 
 if [ -e "/dev/kvm" ] && getent group kvm > /dev/null; then
-  groupmod -g "$(ls -lhn /dev/kvm | cut -d ' ' -f 4)" kvm
+  groupmod -g "$(ls -lhn /dev/kvm | cut -d ' ' -f 4)" --non-unique kvm
```
kalikiana, line 12: Any chance we can test this? With e.g. a `healthcheck` or something else?

okurz (commented):
I would also appreciate a proper test

container/worker/run_openqa_worker.sh
ilausuch: I added two checks. First to checks that the kvm group don't have already the id group 0. And an other after the groupmod execution to check hat the id is changed.

However, I think to test this should be done on the openQA tests

container/worker/run_openqa_worker.sh
```diff
@@ -9,7 +9,9 @@ fi
 /root/qemu/kvm-mknod.sh
 
 if [ -e "/dev/kvm" ] && getent group kvm > /dev/null; then
-  groupmod -g "$(ls -lhn /dev/kvm | cut -d ' ' -f 4)" kvm
+  group=$(ls -lhn /dev/kvm | cut -d ' ' -f 4)
+  [[ $(getent group kvm | cut -d ":" -f3) -ne "$group" ]] && groupmod -g "$group" --non-unique kvm
+  [[ $(getent group kvm | cut -d ":" -f3) -ne "$group" ]] && { echo "Error: The kvm group id must be 0"; exit 1; }
```
Martchus, line 14: The error message makes so sense. How many times have I already written that this code sets the ID of the `kvm` group within the container to the group ID of the mapped `/dev/kvm` so both match? Hence the group id is *not* necessarily supposed to be 0. A better error message would be "Unable to ensure the ID of the kvm matches the group ID of /dev/kvm. KVM will not work.".

container/worker/run_openqa_worker.sh
```diff
@@ -9,7 +9,9 @@ fi
 /root/qemu/kvm-mknod.sh
 
 if [ -e "/dev/kvm" ] && getent group kvm > /dev/null; then
-  groupmod -g "$(ls -lhn /dev/kvm | cut -d ' ' -f 4)" kvm
+  group=$(ls -lhn /dev/kvm | cut -d ' ' -f 4)
+  [[ $(getent group kvm | cut -d ":" -f3) -ne "$group" ]] && groupmod -g "$group" --non-unique kvm
```
Martchus, line 13: The groupmod command will not do any harm if the group ID won't change so I'd get rid of the check:
```suggestion
groupmod -g "$group" --non-unique kvm
```

Martchus (changes requested):

container/worker/run_openqa_worker.sh
ilausuch: A possible option is to run `kvm -S daemonize` that creates a vm without execution. If the user doesn't have permissions this command exits with code 1, other ways creates the VM in background. However that needs the installation of some additional packages.

Amrysliu:
I don't know much about `groupmod kvm`.  I am working on the openQA container test case, so I added `assert_script_run('sed -i \'s/\" kvm/\" --non-unique kvm/\' openQA/container/worker/run_openqa_worker.sh', timeout => 300);`  in the test code to confirm that. The test result shows: https://openqa.opensuse.org/tests/1668416#step/worker/3. we could see there is no `groupmod GID '0' already exists` in it. I guess only adding `--non-unique` should be ok.

Martchus (changes requested):
The solution stated in the first commit message is now longer matching the actual contents of the commit. I'm also wondering what is gained by the 2nd commit. Why did you add it? I'd just add the `--non-unique`. If `groupmod` fails the script will abort at this point anyways. By the way, when others requested a test I don't think they had further checks within `run_openqa_worker.sh` itself in mind.

ilausuch:
> The solution stated in the first commit message is now longer matching the actual contents of the commit. I'm also wondering what is gained by the 2nd commit. Why did you add it? I'd just add the `--non-unique`. If `groupmod` fails the script will abort at this point anyways. By the way, when others requested a test I don't think they had further checks within `run_openqa_worker.sh` itself in mind.

@kalikiana and @okurz asked for a test. I tried to create a sort of, but maybe this repo is not the correct place and should be included in https://github.com/os-autoinst/os-autoinst-distri-openQA as complement of the running the container tests

Martchus:
> @kalikiana and @okurz asked for a test.

Yes, but I don't think they wanted to you add a check to the code which is supposed to be tested itself. Likely the error you're introducing here would never be raised because the script would stop before anyways due to `set -e`. So I would just drop the 2nd commit. Not sure how easy it'll be to add a real test because it depends on the group ID of `/dev/kdm` and the default group ID of the `kvm` group within the container.

ilausuch:
> > @kalikiana and @okurz asked for a test.
> 
> Yes, but I don't think they wanted to you add a check to the code which is supposed to be tested itself. Likely the error you're introducing here would never be raised because the script would stop before anyways due to `set -e`. So I would just drop the 2nd commit. Not sure how easy it'll be to add a real test because it depends on the group ID of `/dev/kdm` and the default group ID of the `kvm` group within the container.

In these terms, I agree with you. Let's see what the others reviewers say about this before removing this commit

okurz:
I agree with what Martchus has written. I don't think the script would even continue if the first call was not effective. So now rather I would wonder what a "test of the check" would look like :D When I said "test" of course I meant something that would be actually executed not error handling that has no chance to be called. https://github.com/os-autoinst/openQA/pull/3787#discussion_r594489053 sounds promising. Having tests within the repo which are executed for every PR is always better than tests that are just called after merge, e.g. os-autoinst-distri-openQA but that again is still better than no test :)

ilausuch:
> I agree with what Martchus has written. I don't think the script would even continue if the first call was not effective. So now rather I would wonder what a "test of the check" would look like :D When I said "test" of course I meant something that would be actually executed not error handling that has no chance to be called. [#3787 (comment)](https://github.com/os-autoinst/openQA/pull/3787#discussion_r594489053) sounds promising. Having tests within the repo which are executed for every PR is always better than tests that are just called after merge, e.g. os-autoinst-distri-openQA but that again is still better than no test :)

Ok trying to implement [#3787 (comment)](https://github.com/os-autoinst/openQA/pull/3787#discussion_r594489053)

container/worker/run_openqa_worker.sh
kalikiana: If this is going to be an openQA test, I suggest we get this PR merged now. My question was assuming a trivial `healthcheck` or one-line check within `run_openqa_worker.sh`. And I take your answer to mean that's not feasible.

kalikiana: So I guess this is what you were adding to *run_openqa_worker.sh*:

    qemu-kvm -S &
    kill $!

Mentioning it here to keep the conversation context. I'm not fond of this approach personally as it requires additional deps only used by this check.

okurz:
Could you please clarify where the corresponding test results for your changes are? Now that we have a docker compose test this should be easier to do

ilausuch:
> Could you please clarify where the corresponding test results for your changes are? Now that we have a docker compose test this should be easier to do

Now there is a test for this piece of code. I created a docker-compose for testing the worker and add this to the test_containers_compose test.

As in the case of the webui test this checks that the worker container does not end unexpectedly because of the creation of a qemu-kvm server. Otherways the worker enters in a loop to try to connect to the webui and the container will remain working

container/worker/run_openqa_worker.sh
```diff
@@ -9,8 +9,15 @@ fi
 /root/qemu/kvm-mknod.sh
 
 if [ -e "/dev/kvm" ] && getent group kvm > /dev/null; then
-  groupmod -g "$(ls -lhn /dev/kvm | cut -d ' ' -f 4)" kvm
+  group=$(ls -lhn /dev/kvm | cut -d ' ' -f 4)
+  groupmod -g "$group" --non-unique kvm
+  [[ $(getent group kvm | cut -d ":" -f3) -ne "$group" ]] && { echo "Unable to ensure the ID of the kvm matches the group ID of /dev/kvm. KVM will not work."; exit 1; }
```
Martchus, line 14: Didn't we come to the conclusion that this line is useless because it basically checks whether the previous command succeeds but with `set -e` it won't be executed anyways?

container/worker/run_openqa_worker.sh
```diff
@@ -9,8 +9,15 @@ fi
 /root/qemu/kvm-mknod.sh
 
 if [ -e "/dev/kvm" ] && getent group kvm > /dev/null; then
-  groupmod -g "$(ls -lhn /dev/kvm | cut -d ' ' -f 4)" kvm
+  group=$(ls -lhn /dev/kvm | cut -d ' ' -f 4)
+  groupmod -g "$group" --non-unique kvm
+  [[ $(getent group kvm | cut -d ":" -f3) -ne "$group" ]] && { echo "Unable to ensure the ID of the kvm matches the group ID of /dev/kvm. KVM will not work."; exit 1; } 
   usermod -a -G kvm _openqa-worker
+  qemu-kvm -S &
```
Martchus, line 16: This is just a symlink to e.g. `qemu-system-x86_64` which you're hard-coding below anyways. So I'd just use `qemu-system-x86_64 -enable-kvm`. You maybe need `-nographic` as well to prevent the GTK3 error in a headless environment.

container/worker/run_openqa_worker.sh
```diff
@@ -9,8 +9,15 @@ fi
...
+  qemu_pid=$!
+  sleep 1
+  (pgrep qemu-system-x86 >/dev/null) || (echo "Unnable to start qemu-kvm"; exit 1)
```
Martchus, line 19: I'd just get completely rid of this. If the line above fails these lines won't be executed anyways due to `set -e`. So just `kill $!` should be sufficient.

Martchus (changes requested):

ilausuch:
The test failed because the container is not running in privileged mode although is specified in the docker compose. The reason could be that circleci doesn't allow privileged mode runs.

```
worker_1  | mknod: invalid minor device number ''
worker_1  | Unable to make /dev/kvm node; software emulation will be used
worker_1  | (This can happen if the container is run without -privileged)
```

container/worker/run_openqa_worker.sh
```diff
@@ -9,8 +9,11 @@ fi
 /root/qemu/kvm-mknod.sh
 
 if [ -e "/dev/kvm" ] && getent group kvm > /dev/null; then
-  groupmod -g "$(ls -lhn /dev/kvm | cut -d ' ' -f 4)" kvm
+  group=$(ls -lhn /dev/kvm | cut -d ' ' -f 4)
+  groupmod -g "$group" --non-unique kvm
   usermod -a -G kvm _openqa-worker
+  qemu-system-x86_64 -enable-kvm -S &
```
okurz, line 15: this is arch-specific and should be avoided

Martchus (commented):
The change looks ok but it would make sense to squash the commits as they just introduce and then remove certain changes. (We can do this when merging via GitHub although Mergify doesn't help with it.)

ilausuch:
Ok in this push the worker entrypoint checks if is defined and enabled the variable qemu_no_kvm. We have now the test pass because the docker-conpose file includes the qemu_no_kvm set to 1 and the test is not running this part of the code in the entrypoint. What do you think? Should we find an other way to test the kvm part?

ilausuch:
Now this test runs a qemu test. If it's ok to you then I can squash the commits

container/worker/run_openqa_worker.sh
ilausuch: I provided a new way to check the qemu without using kvm

container/worker/Dockerfile
```diff
@@ -8,7 +8,7 @@ RUN zypper ar -p 95 -f http://download.opensuse.org/repositories/devel:openQA/op
     zypper ar -p 90 -f http://download.opensuse.org/repositories/devel:openQA:Leap:15.2/openSUSE_Leap_15.2 devel_openQA_Leap && \
     zypper --gpg-auto-import-keys ref && \
     zypper in -y ca-certificates-mozilla curl gzip && \
-    zypper in -y openQA-worker qemu-arm qemu-ppc qemu-x86 qemu-tools && \
+    zypper in -y openQA-worker qemu-arm qemu-ppc qemu-x86 qemu-tools qemu-kvm && \
```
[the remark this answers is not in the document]
ilausuch: This is removed now

container/worker/run_openqa_worker.sh
```diff
@@ -1,18 +1,25 @@
...
-  echo "Warning: /dev/kvm doesn't exist. If you want to use KVM, run the container with --device=/dev/kvm"
+    group=$(ls -lhn /dev/kvm | cut -d ' ' -f 4)
+    groupmod -g "$group" --non-unique kvm
+    usermod -a -G kvm _openqa-worker
+  else
+    echo "Warning: /dev/kvm doesn't exist. If you want to use KVM, run the container with --device=/dev/kvm"
+  fi
 fi
 
+qemu-system-x86_64 -S &
```
Martchus, line 22: You could try to run it with `-enable-kvm` parameter unless `$qemu_no_kvm` is set.

Martchus (approved):
I guess this is good enough for now.

container/worker/run_openqa_worker.sh
ilausuch: I did in a previous commit, but I understood that was not a good idea (https://github.com/os-autoinst/openQA/pull/3787#discussion_r603274782)

container/worker/run_openqa_worker.sh
```diff
@@ -1,18 +1,25 @@
 #!/bin/bash
-# shellcheck disable=SC2012
+# shellcheck disable=SC2012,SC2154
 set -e
+set -x
```
okurz, line 4: this looks like temporary debugging code

container/worker/run_openqa_worker.sh
```diff
@@ -1,18 +1,25 @@
...
+    group=$(ls -lhn /dev/kvm | cut -d ' ' -f 4)
+    groupmod -g "$group" --non-unique kvm
+    usermod -a -G kvm _openqa-worker
+  else
+    echo "Warning: /dev/kvm doesn't exist. If you want to use KVM, run the container with --device=/dev/kvm"
+  fi
 fi
 
+qemu-system-x86_64 -S &
+kill $!
```
okurz, line 23: can you explain this line please?

Martchus: It tests whether qemu starts and terminates with zero return code.

okurz: yeah ok, that makes sense for testing but it is helpful for the user as well?

ilausuch: I understand, if I am not wrong that with that, we are checking that this worker can execute jobs

container/worker/run_openqa_worker.sh
ilausuch: Removed

ilausuch:
I re-based from mater because of a checkstyle problem not related

kalikiana (approved):

Amrysliu (approved):

Amrysliu (approved):

PR #3798, opened 2021-03-18
Dependency cron 2021-03-18

openqabot:


Martchus (commented):
A downgrade? That looks wrong. It likely happens because https://build.opensuse.org/package/show/devel:openQA:Leap:15.2/perl-Mojo-Pg is unresolvable. I'll try linking `perl-SQL-Abstract-Pg`.

Martchus:
I linked the package and the build succeeds again. I also covered the other devel repos (besides Leap 15.2). The next dependency cron PR should contain the right version.

okurz:
I triggered a rerun in https://app.circleci.com/pipelines/github/os-autoinst/openQA/5944/workflows/40a09bdf-f648-4325-9eef-eb17aee90ed7 to have updated dependencies to ensure new pull requests can be properly tested

PR #3804, opened 2021-03-18
Allow to override isotovideo executable with command line

okurz:
os-autoinst includes multiple backends which can require changes to 
support different technologies or specific testing needs. Also, the
worker environment needs to provide requirements to run tests which
include more package than originally required for os-autoinst or newer
versions.

This commit introduces the possibility to customize the actual worker
engine that is started with keeping the default of starting "perl
$isotovideo -d". For example a different location for isotovideo can be 
specified or a container providing different versions of packages or 
even a completely different worker engine that is interface-compatible.

Working on https://github.com/os-autoinst/os-autoinst/pull/1558 this
idea came back to mind. I had already thought about doing something
similar couple of times, also based on requests by other users to be
able to test os-autoinst changes more easily. With this change it
should be possible to execute an arbitrary command within the worker
based on a job setting. For example the use case I envision is to
build a container image from a still open pull request of os-autoinst
and execute tests on a production instance without harmful impact of
the general environment, e.g.
`podman run --rm -it registry.opensuse.org/devel/openqa/...`.

This could also be helpful for customized backend environments needed
to access other, more exotic hardware and such.

https://progress.opensuse.org/issues/90362

### DIFF
--- a/lib/OpenQA/Worker/Engines/isotovideo.pm
+++ b/lib/OpenQA/Worker/Engines/isotovideo.pm
@@ -371,7 +371,7 @@ sub engine_workit {
 
     # os-autoinst's commands server
     $job_info->{URL}
-      = "http://localhost:" . ($job_info->{settings}->{QEMUPORT} + 1) . "/" . $job_info->{settings}->{JOBTOKEN};
+      = "http://localhost:" . ($job_settings->{QEMUPORT} + 1) . "/" . $job_settings->{JOBTOKEN};
 
     # create cgroup within /sys/fs/cgroup/systemd
     log_info('Preparing cgroup to start isotovideo');
--- a/lib/OpenQA/Worker/Engines/isotovideo.pm
+++ b/lib/OpenQA/Worker/Engines/isotovideo.pm
@@ -426,7 +426,9 @@ sub engine_workit {
             # PERL5OPT may have Devel::Cover options, we don't need and want
             # them in the spawned process as it does not belong to openQA code
             local $ENV{PERL5OPT} = "";
-            exec "perl", "$isotovideo", '-d';
+            # Allow to override isotovideo executable with an arbitrary
+            # command line based on a config option
+            exec $job_settings->{ISOTOVIDEO} ? $job_settings->{ISOTOVIDEO} : ('perl', $isotovideo, '-d');
             die "exec failed: $!\n";
         });
     $child->on(
--- a/.circleci/unstable_tests.txt
+++ b/.circleci/unstable_tests.txt
@@ -1,3 +1,4 @@
 t/25-cache-service.t
 t/ui/01-list.t
 t/ui/26-jobs_restart.t
+t/ui/13-admin.t


okurz:
https://github.com/os-autoinst/openQA/pull/3804/files#diff-53098d895ef28f5c7db596e2dc505d7c8ee24b6a9a4cca23865fb225e10cd9fcR380 is the most important line, the rest is refactoring

Martchus (commented):
I'm not sure whether that's the best idea from a security perspective. I haven't though it through but that's something we should keep in mind.

okurz:
Sure, I thought about that as well but we don't have any different layers of security enforcement between os-autoinst and test code which can be arbitrary perl code as well. And do override the settings one needs operator permissions, same as when scheduling arbitrary code in tests. However we have cgroups on the next outer layer as well apparmor

lib/OpenQA/Worker/Engines/isotovideo.pm
```diff
@@ -390,7 +390,7 @@ sub engine_workit {
             # PERL5OPT may have Devel::Cover options, we don't need and want
             # them in the spawned process as it does not belong to openQA code
             local $ENV{PERL5OPT} = "";
-            exec "perl", "$isotovideo", '-d';
+            exec $job_settings->{ISOTOVIDEO} ? split(' ', $job_settings->{ISOTOVIDEO}) : "perl", "$isotovideo", '-d';
```
Martchus, line 393: This simple splitting will not handle quoting/escaping correctly. Maybe it makes more sense to just pass the path as-is? And for readability wouldn't it make sense to swap out the `, "$isotovideo" ` part as well? Or is the code already doing that (braces would help)?

Martchus (commented):
I suppose this really doesn't open any doors which aren't already open.

lib/OpenQA/Worker/Engines/isotovideo.pm
okurz: yes, you are right, passing as-is seems to be a better choice. https://perldoc.perl.org/functions/exec is much smarter than my split.

I guess braces help. By the way, to test locally:

```bash
env MY_COMMAND="echo -e 'my custom\tcommand'" perl -e 'exec $ENV{MY_COMMAND} ? $ENV{MY_COMMAND} : ("echo", "-e", "foo", "\n")';
```

lib/OpenQA/Worker/Engines/isotovideo.pm
```diff
@@ -426,7 +426,7 @@ sub engine_workit {
             # PERL5OPT may have Devel::Cover options, we don't need and want
             # them in the spawned process as it does not belong to openQA code
             local $ENV{PERL5OPT} = "";
-            exec "perl", "$isotovideo", '-d';
+            exec $job_settings->{ISOTOVIDEO} ? $job_settings->{ISOTOVIDEO} : ("perl", "$isotovideo", '-d');
```
Martchus, line 429: Looks good, I would only avoid unnecessary (double) quoting.
```suggestion
            exec $job_settings->{ISOTOVIDEO} ? $job_settings->{ISOTOVIDEO} : ('perl', $isotovideo, '-d');
```

okurz: done

lib/OpenQA/Worker/Engines/isotovideo.pm
```diff
@@ -371,7 +371,7 @@ sub engine_workit {
 
     # os-autoinst's commands server
     $job_info->{URL}
-      = "http://localhost:" . ($job_info->{settings}->{QEMUPORT} + 1) . "/" . $job_info->{settings}->{JOBTOKEN};
+      = "http://localhost:" . ($job_settings->{QEMUPORT} + 1) . "/" . $job_settings->{JOBTOKEN};
```
kraih, line 374: Technically there's unnecessary double quoting here too.

Martchus (approved):

lib/OpenQA/Worker/Engines/isotovideo.pm
okurz: ok, I have a draft PR with further refactoring in https://github.com/os-autoinst/openQA/pull/3807 where I will include your proposal

okurz: done

kraih (approved):

PR #3807, opened 2021-03-19
worker: Simplify use of job_settings hash lookup

okurz:


lib/OpenQA/Worker/Engines/isotovideo.pm
```diff
@@ -273,52 +262,16 @@ sub do_asset_caching {
...
-        OPENQA_URL      => $openqa_url,
+    my %vars = (
+        OPENQA_URL      => $webui_host,
         WORKER_INSTANCE => $instance,
-        WORKER_ID       => $workerid,
+        WORKER_ID       => $worker_id,
         PRJDIR          => OpenQA::Utils::sharedir(),
-        %$job_settings
-    );
+        %$job->info->{settings});
```
okurz, line 274: this triggers `Can't locate object method "info" via package "21"`. Anyone has an idea how to solve that better?

perlpunk:
```suggestion
        %{ $job->info->{settings} })
```

okurz: Right, that gives me `Use of uninitialized value in concatenation (.) or string`

perlpunk: there is no string concatenation happening in this part of the code?

Martchus: I was wondering about the same. Generally the suggested code change is supposed to work.

okurz: ah ok, I see that it's now about a different line. Well, that helps, I can go on :)

okurz: Well, couldn't understand that part while being in another casual video meeting so for now I just left out that commit :)

Martchus:
Looks generally good (besides the missing patch coverage).

Martchus (approved):

perlpunk:
checkstyle failed

perlpunk:
https://github.com/os-autoinst/openQA/pull/4078 for a fix

okurz:
thanks. I wonder if maybe auto-merge skips some checks and maybe that was my problem?

PR #3810, opened 2021-03-22
Allow client to get YAML job templates from a file

alvarocarvajald:
Currently, openQA client's help recommends the following to configure a job group from a YAML template:
```
  # Post job template
  openqa-cli api -X POST job_templates_scheduling/1 \
    schema=JobTemplates-01.yaml preview=0 template="$(cat foo.yaml)"
```
A potential issue with this, is that when the YAML file reaches a certain size, such a command can hit the `Argument list too long` error on a shell, for example:
```
openqa-cli api --host http://mango.qa.suse.de --apikey **** --apisecret **** -X POST \
    job_templates_scheduling/2 schema=JobTemplates-01.yaml preview=1 \
    template="$(cat sle15_sp3_ha.yml)"
-bash: /usr/bin/openqa-cli: Argument list too long
```
Or:
```
cat sle15_sp3_ha.yml | openqa-cli api --host http://mango.qa.suse.de --apikey **** --apisecret **** \
    -X POST job_templates_scheduling/2 schema=JobTemplates-01.yaml preview=1 \
    template="$(cat -)"
-bash: /usr/bin/openqa-cli: Argument list too long
```
With this PR, I'm suggesting adding a new option to `OpenQA::CLI::api` titled `--param-file`, which can be used to pass parameters whose value is read from files by the client before connection to the server, instead of given on the command line. An example of the previous calls would be:
```
openqa-cli api --host http://mango.qa.suse.de --apikey **** --apisecret *** -X POST \
    job_templates_scheduling/2 schema=JobTemplates-01.yaml preview=1 \
    --param-file template=sle15_sp3_ha.yml
```
This way we avoid the `Argument list too long` error from the shell, and move the task of reading the file to the client.

--

Tests performed before and after applying changes to my local instance:

1. `Argument list too long` error with current version of `openqa-cli`, YAML file contents in command line:
```
# openqa-cli api --host http://mango.qa.suse.de --apikey *** --apisecret *** -X POST \
    job_templates_scheduling/2 schema=JobTemplates-01.yaml preview=1 \
    template="$(cat sle15_sp3_ha.yml)"
-bash: /usr/bin/openqa-cli: Argument list too long
```
2. `Expected object - got string` error with current version of `openqa-cli`, YAML file path in command line:
```
# openqa-cli api --host http://mango.qa.suse.de --apikey **** --apisecret **** -X POST \
    job_templates_scheduling/2 schema=JobTemplates-01.yaml preview=1 \
    template=file:./sle15_sp3_ha.yml
400 Bad Request
{"error":[{"message":"Expected object - got string.","path":"\/"}],"error_status":400}
```
3. `Expected object - got string` error with `openqa-cli` **after changes were applied**, YAML file path in command line without new option:
```
# openqa-cli api --host http://mango.qa.suse.de --apikey **** --apisecret **** -X POST \
    job_templates_scheduling/2 schema=JobTemplates-01.yaml preview=1 \
    template=file:./sle15_sp3_ha.yml 
400 Bad Request
{"error":[{"message":"Expected object - got string.","path":"\/"}],"error_status":400}
```
4. File not found error with `openqa-cli` **after changes were applied**, YAML file path in command line with new option but wrongly formatted:
```
# openqa-cli api --host http://mango.qa.suse.de --apikey **** --apisecret **** -X POST \
    job_templates_scheduling/2 schema=JobTemplates-01.yaml preview=1 \
   --param-file template=file:./sle15_sp3_ha.yml 
Can't open file "file:./sle15_sp3_ha.yml": No such file or directory at /usr/share/openqa/script/../lib/OpenQA/Command.pm line 95.
```
5. Template processing error (job already defined in another group) with `openqa-cli` **after changes were applied**, YAML file path in command line:
```
# openqa-cli api --host http://mango.qa.suse.de --apikey **** --apisecret **** -X POST \
    job_templates_scheduling/2 schema=JobTemplates-01.yaml preview=1 \
    --param-file template=./sle15_sp3_ha.yml
400 Bad Request
{"error":["Job template name 'create_hdd_ha_textmode' with sle-15-SP3-Online-aarch64 and aarch64 is already used in job group 'HA'"],"error_status":400,"id":2,"job_group_id":2}
```
6. Success with `openqa-cli` **after changes were applied**, YAML file path in command line:
```
# openqa-cli api --host http://mango.qa.suse.de --apikey **** --apisecret **** -X POST \
    job_templates_scheduling/2 schema=JobTemplates-01.yaml preview=1 \
    --param-file template=./development.yml 
{"id":2,"ids":[495,496,497],"job_group_id":2,"preview":1}
```
--

As suggested in the comments, changes are only applied to `openqa-cli` and not to `openqa-client`.

### DIFF
--- a/lib/OpenQA/CLI/api.pm
+++ b/lib/OpenQA/CLI/api.pm
@@ -1,4 +1,4 @@
-# Copyright (C) 2020 SUSE LLC
+# Copyright (C) 2020-2021 SUSE LLC
 #
 # This program is free software; you can redistribute it and/or modify
 # it under the terms of the GNU General Public License as published by
@@ -35,6 +35,7 @@ sub command {
       'd|data=s'      => \$data,
       'f|form'        => \my $form,
       'j|json'        => \my $json,
+      'param-file=s'  => \my @param_file,
       'p|pretty'      => \my $pretty,
       'q|quiet'       => \my $quiet,
       'X|method=s'    => \(my $method = 'GET'),
@@ -45,7 +46,7 @@ sub command {
 
     $data = path($data_file)->slurp if $data_file;
     my @data   = ($data);
-    my $params = $form ? decode_json($data) : $self->parse_params(@args);
+    my $params = $form ? decode_json($data) : $self->parse_params(\@args, \@param_file);
     @data = (form => $params) if keys %$params;
 
     my $headers = $self->parse_headers(@headers);
@@ -105,26 +106,34 @@ sub command {
     openqa-cli api -X POST job_templates_scheduling/1 \
       schema=JobTemplates-01.yaml preview=0 template="$(cat foo.yaml)"
 
+    # Post job template (from file)
+    openqa-cli api -X POST job_templates_scheduling/1 \
+      schema=JobTemplates-01.yaml preview=0 --param-file template=foo.yaml
+
     # Post job template (from JSON file)
     openqa-cli api --data-file form.json -X POST job_templates_scheduling/1
 
   Options:
-        --apibase <path>        API base, defaults to /api/v1
-        --apikey <key>          API key
-        --apisecret <secret>    API secret
-    -a, --header <name:value>   One or more additional HTTP headers
-    -D, --data-file <path>      Load content to send with request from file
-    -d, --data <string>         Content to send with request, alternatively you
-                                can also pipe data to openqa-cli
-    -f, --form                  Turn JSON object into form parameters
-        --host <host>           Target host, defaults to http://localhost
-    -h, --help                  Show this summary of available options
-    -j, --json                  Request content is JSON
-        --osd                   Set target host to http://openqa.suse.de
-        --o3                    Set target host to https://openqa.opensuse.org
-    -p, --pretty                Pretty print JSON content
-    -q, --quiet                 Do not print error messages to STDERR
-    -X, --method <method>       HTTP method to use, defaults to GET
-    -v, --verbose               Print HTTP response headers
+        --apibase <path>          API base, defaults to /api/v1
+        --apikey <key>            API key
+        --apisecret <secret>      API secret
+    -a, --header <name:value>     One or more additional HTTP headers
+    -D, --data-file <path>        Load content to send with request from file
+    -d, --data <string>           Content to send with request, alternatively
+                                  you can also pipe data to openqa-cli
+    -f, --form                    Turn JSON object into form parameters
+        --host <host>             Target host, defaults to http://localhost
+    -h, --help                    Show this summary of available options
+    -j, --json                    Request content is JSON
+        --osd                     Set target host to http://openqa.suse.de
+        --o3                      Set target host to https://openqa.opensuse.org
+        --param-file <param=file> Load content of params from files instead of
+                                  from command line arguments. Multiple params
+                                  may be specified by adding the option
+                                  multiple times
+    -p, --pretty                  Pretty print JSON content
+    -q, --quiet                   Do not print error messages to STDERR
+    -X, --method <method>         HTTP method to use, defaults to GET
+    -v, --verbose                 Print HTTP response headers
 
 =cut
--- a/lib/OpenQA/Command.pm
+++ b/lib/OpenQA/Command.pm
@@ -1,4 +1,4 @@
-# Copyright (C) 2020 SUSE LLC
+# Copyright (C) 2020-2021 SUSE LLC
 #
 # This program is free software; you can redistribute it and/or modify
 # it under the terms of the GNU General Public License as published by
@@ -21,10 +21,12 @@ use OpenQA::Client;
 use Mojo::IOLoop;
 use Mojo::Util qw(decode getopt);
 use Mojo::URL;
+use Mojo::File qw(path);
 use Term::ANSIColor qw(colored);
 
 my $JSON = Cpanel::JSON::XS->new->utf8->canonical->allow_nonref->allow_unknown->allow_blessed->convert_blessed
   ->stringify_infnan->escape_slash->allow_dupkeys->pretty;
+my $PARAM_RE = qr/^([[:alnum:]_\[\]\.]+)=(.*)$/s;
 
 has apibase => '/api/v1';
 has [qw(apikey apisecret host)];
@@ -81,12 +83,17 @@ sub parse_headers {
 }
 
 sub parse_params {
-    my ($self, @args) = @_;
+    my ($self, $args, $param_file) = @_;
 
     my %params;
-    for my $arg (@args) {
-        next unless $arg =~ /^([[:alnum:]_\[\]\.]+)=(.*)$/s;
+    for my $arg (@{$args}) {
+        next unless $arg =~ $PARAM_RE;
         push @{$params{$1}}, $2;
+    }
+
+    for my $arg (@{$param_file}) {
+        next unless $arg =~ $PARAM_RE;
+        push @{$params{$1}}, path($2)->slurp;
     }
 
     return \%params;
--- a/t/43-cli-api.t
+++ b/t/43-cli-api.t
@@ -1,4 +1,4 @@
-# Copyright (C) 2020 SUSE LLC
+# Copyright (C) 2020-2021 SUSE LLC
 #
 # This program is free software; you can redistribute it and/or modify
 # it under the terms of the GNU General Public License as published by
@@ -493,6 +493,38 @@ subtest 'PIPE input' => sub {
     my ($stdout, @result) = capture_stdout sub { $api->run(@host, 'test/pub/http') };
     my $data = decode_json $stdout;
     is $data->{body}, 'Hello openQA!', 'request body';
+};
+
+subtest 'YAML Templates' => sub {
+    my ($stdout, $stderr, @result) = capture sub { $api->run(@host, '-X', 'POST', 'job_groups', 'name=Test'); };
+    my $data = decode_json $stdout;
+    is $data->{id}, '1', 'create job group';
+    is_deeply \@result, [0], 'create job group - exit code';
+    is $stderr, '', 'create job group - stderr quiet';
+    my $yaml_text = "products: {}\nscenarios: {}";
+    ($stdout, $stderr, @result) = capture sub {
+        $api->run(@host, '-X', 'POST', 'job_templates_scheduling/1', 'schema=JobTemplates-01.yaml', 'preview=1',
+            "template=$yaml_text");
+    };
+    $data = decode_json $stdout;
+    is $data->{job_group_id}, '1', 'YAML template as param';
+    is_deeply \@result, [0], 'YAML template as param - exit code';
+    is $stderr, '', 'YAML template as param - stderr quiet';
+    my $file = tempfile->spurt($yaml_text);
+    ($stdout, $stderr, @result) = capture sub {
+        $api->run(@host, '-X', 'POST', 'job_templates_scheduling/1', 'schema=JobTemplates-01.yaml', 'preview=1',
+            "template=$file");
+    };
+    is_deeply \@result, [1], 'File name as template - right error code';
+    like $stderr, qr/400 Bad Request/, 'File name as template - right error msg';
+    ($stdout, $stderr, @result) = capture sub {
+        $api->run(@host, '-X', 'POST', 'job_templates_scheduling/1', 'schema=JobTemplates-01.yaml', 'preview=1',
+            '--param-file', "template=$file");
+    };
+    $data = decode_json $stdout;
+    is $data->{job_group_id}, '1', 'YAML template by file';
+    is_deeply \@result, [0], 'YAML template by file - exit code';
+    is $stderr, '', 'YAML template by file - stderr quiet';
 };
 
 done_testing();


lib/OpenQA/Command.pm
```diff
@@ -89,6 +89,16 @@ sub parse_params {
         push @{$params{$1}}, $2;
     }
 
+    return \%params unless (defined $params{template});
+
+    foreach my $t (@{$params{template}}) {
+        if ($t =~ /^file:(.+)$/) {
+            open(my $fh, '<', $1) || die "ERROR: cannot open file [$1]: $!\n";
+            $t = do { local $/; <$fh> };
+            close $fh;
```
Martchus, line 98: It would be nicer to use `Mojo::File::slurp` instead. You could also use `Mojo::URL` for checking whether the schema is `file`.

lib/OpenQA/Script/Client.pm
```diff
@@ -150,6 +150,12 @@ sub run {
         $url->query([%params]) if %params;
     }
 
+    if (defined $params{template} && $params{template} =~ /^file:(.+)$/) {
```
Martchus, line 153: The same counts as for the occurrence in `Command.pm`. I'm also wondering the the duplication is necessary here.

t/31-client.t
```diff
@@ -105,4 +105,19 @@ $options{json} = 1;
 stderr_like \&wrong_call, qr/$code.*Not Found/, 'Error reported for undocumented "json" parameter';
 is $ret, undef, 'undef shows error for undocumented parameter';
 
+subtest 'YAML Templates' => sub {
+    my $template_openqa = "$FindBin::Bin/data/job-templates/openqa.yaml";
+    my @common_args     = qw(job_templates_scheduling/1 post schema=JobTemplates-01.yaml preview=1);
+    my $ret;
+    stderr_is { $ret = run(\%options, qw(job_groups post name=Test)) } '', 'create job group';
+    warn "RET is [$ret]";
+    is $ret, { id => 1 }, 'create job group';
```
Martchus: I'd remove that unconditional warning. If `$res` is not the expected value the actual value should be logged by the subsequent line if you use `is_deeply`.

t/43-cli-api.t
```diff
@@ -495,4 +498,31 @@ subtest 'PIPE input' => sub {
     is $data->{body}, 'Hello openQA!', 'request body';
 };
 
+subtest 'YAML Templates' => sub {
+    my ($stdout, $stderr, @result) = capture sub { $api->run(@host, '-X', 'POST', 'job_groups', 'name=Test'); };
+    warn "stdout=[$stdout], stderr[$stderr], results[@result]";
```
Martchus: Please remove the unconditional warnings in this subtest as well.

Martchus (changes requested):
Looks generally good (besides the issues mentioned in the inline comments). I appreciate that you're extending the tests.

lib/OpenQA/Script/Client.pm
alvarocarvajald: Figured I needed to extend both `openqa-client` and `openqa-cli` so the option is available in both, and couldn't think of anything better than to duplicate the code. In a sense, `OpenQA::Command::parse_params()` is already a duplicate of code from `OpenQA::Script::Client`. I'll see how it looks after I apply changes with `Mojo::File::slurp` and `Mojo::URL` and make a decision regarding this.

t/31-client.t
alvarocarvajald: Oh, I already removed it. This was added just to debug the tests, as they're working in my local system.

t/43-cli-api.t
alvarocarvajald: Same reason as above. Sorry for the noise.

lib/OpenQA/Script/Client.pm
Martchus: Ah, ok. If it is just for the both clients it makes sense. Since it'll be just one line using slurp I suppose it wouldn't make sense to create a separate function.

lib/OpenQA/Command.pm
alvarocarvajald: Done.

lib/OpenQA/Script/Client.pm
alvarocarvajald: > Ah, ok. If it is just for the both clients it makes sense. Since it'll be just one line using slurp I suppose it wouldn't make sense to create a separate function.

So it's 2 lines inside an if-block in both modules. Should I attempt to factorize this so it's in a single location? I think it'd add more complexity to be honest. Let me know what you think.

lib/OpenQA/Command.pm
```diff
@@ -89,6 +90,13 @@ sub parse_params {
         push @{$params{$1}}, $2;
     }
 
+    return \%params unless (defined $params{template});
+
+    foreach my $tval (@{$params{template}}) {
+        my $t = Mojo::URL->new($tval);
+        $tval = Mojo::File->new($t->path)->slurp if ($t->scheme eq 'file');
```
Martchus, line 97: `scheme` might not be defined so better use `// ''` to avoid a warning:
```suggestion
        $tval = Mojo::File->new($t->path)->slurp if ($t->scheme // '') eq 'file';
```
The same counts for the other occurrence.

Martchus (changes requested):

lib/OpenQA/Command.pm
alvarocarvajald: Done.

Martchus (approved):

t/43-cli-api.t
```diff
@@ -69,6 +69,9 @@ my @host = ('--host', $host);
 # Default options for authentication tests
 my @auth = ('--apikey', 'ARTHURKEY01', '--apisecret', 'EXCALIBUR', @host);
 
+# Job group template for template tests
+my $template_openqa = "$FindBin::Bin/data/job-templates/openqa.yaml";
```
kraih, line 73: No reason for this to be outside the subtest.

okurz:
What you put into the description of the pull request could also be in the commit message details. I don't know how you created your pull request but your git commit messages has *only* the subject line while the pull request description has more details. Please keep in mind that the github pull request description is only visible on github, the commit can be considered permanent information storage. I suggest to use the last commit which has the same title as the PR and squash some other debugging commits.

```bash
$ cat $(which git-pr-last )
#!/bin/sh -e
target="${target:-"$USER"}"
```

Also see https://github.com/os-autoinst/os-autoinst-distri-opensuse/blob/master/CONTRIBUTING.md#coding-style for more details

lib/OpenQA/Script/Client.pm
```diff
@@ -150,6 +151,11 @@ sub run {
         $url->query([%params]) if %params;
     }
 
+    if (defined $params{template}) {
+        my $t = Mojo::URL->new($params{template});
+        $params{template} = Mojo::File->new($t->path)->slurp if ($t->scheme // '') eq 'file';
+    }
+
```
okurz, line 158: Why does this need to be duplicated? How about moving the code into OpenQA::Client into a function and calling that function in both places where needed?

t/31-client_job_groups.t
```diff
@@ -0,0 +1,53 @@
+# Copyright (C) 2020 SUSE LLC
```
okurz, line 1: s/2020/2021/

t/31-client_job_groups.t
```diff
@@ -0,0 +1,53 @@
...
+
+use Mojo::Base -strict;
+
+use FindBin;
+use lib "$FindBin::Bin/lib", "$FindBin::Bin/../external/os-autoinst-common/lib";
+
+use Test::More;
+use Mojo::Server::Daemon;
+use Mojo::JSON qw(decode_json);
+use OpenQA::Script::Client;
```
okurz, line 24: please add `use Test::Warnings ':report_warnings';`

t/31-client_job_groups.t
```diff
@@ -0,0 +1,53 @@
...
+my $host = "http://127.0.0.1:$port";
+
+my %options = (host => $host);
+
+my $template_openqa = "$FindBin::Bin/data/job-templates/openqa.yaml";
+my @common_args     = qw(job_templates_scheduling/1 post schema=JobTemplates-01.yaml preview=1);
+my $ret             = '';
+is $ret, '', 'Initial test';
+$ret = run(\%options, qw(job_groups post name=Test));
+warn "RET is [$ret]";
```
okurz, line 44: this looks like temporary debugging code?

okurz (changes requested):
Please see inline comments and the failed tests in https://app.circleci.com/pipelines/github/os-autoinst/openQA/6043/workflows/9adbecb7-d2cd-4384-b655-4a083ac49c3f/jobs/57018 related to your changes. As I could not yet see the tests pass also make sure that when the tests pass there is no unhandled output, same as for the other test modules. There should only be a single line from prove like `t/31-client_job_groups.t .................................. ok`

t/31-client_job_groups.t
alvarocarvajald: Yes. And it's still not working. I'm dropping that file and look for a different solution to test this.

t/31-client_job_groups.t
alvarocarvajald: File dropped.

t/31-client_job_groups.t
alvarocarvajald: File dropped.

lib/OpenQA/Script/Client.pm
alvarocarvajald: See my reply to @Martchus in https://github.com/os-autoinst/openQA/pull/3810#discussion_r599802300

Basically I figured that this needed to be extended in both clients, hence the code duplication as already there is some duplicated code among both clients. I will move it to a function in `OpenQA::Client` as you suggest.

alvarocarvajald:
> What you put into the description of the pull request could also be in the commit message details.

OK. I'll add it. Should I include the examples as well?

> I don't know how you created your pull request but your git commit messages has _only_ the subject line while the pull request description has more details.

Yes, I figured it explained enough about the committed changes, just as https://github.com/os-autoinst/openQA/commit/f2665c99f90fcbfae47c62b3660d8e1c51e3fff5 has only a subject line.

I'll add more information on the commit itself as in https://github.com/os-autoinst/openQA/commit/0323c017ec5973124e4c0d2e2aba44419eebf96b

> Please keep in mind that the github pull request description is only visible on github, the commit can be considered permanent information storage. I suggest to use the last commit which has the same title as the PR and squash some other debugging commits.

Will do.

> I can recommend the tool [hub](https://github.com/github/hub) (zypper in rubygem-hub) for easier PR creation. Also, I myself use a script [git-pr-last](https://github.com/okurz/scripts/tree/master/git-pr-last) to create a PR with proper description for these simple one-commit PRs:

Thanks. Will look into [hub](https://github.com/github/hub).

> Also see https://github.com/os-autoinst/os-autoinst-distri-opensuse/blob/master/CONTRIBUTING.md#coding-style for more details

So, besides updating copyright years and more information on the commit message, did you have anything else in mind?

t/43-cli-api.t
alvarocarvajald: What do you mean? To test the changes to the `template=` parameter I need to pass a YaML configuration file. See L516.

Should I generate a file on the go from the subtest itself?

lib/OpenQA/Script/Client.pm
alvarocarvajald: Taking a look at `OpenQA::Client`, I don't feel this would be a good place for such a function. WDYT about placing it in `OpenQA::YAML` instead?

t/43-cli-api.t
alvarocarvajald: > No reason for this to be outside the subtest.

Understood. Removed that, and I'm now using a tempfile created on the go.

lib/OpenQA/Script/Client.pm
kraih: Maybe just don't expand the old client with the functionality at all.

perlpunk:
I haven't looked too closely at the PR so far, sorry, but what came to my mind at first was if there maybe should be general support for reading files in the client, not only for one specific parameter? Of course we would need a different syntax, then.

alvarocarvajald:
> I haven't looked too closely at the PR so far, sorry, but what came to my mind at first was if there maybe should be general support for reading files in the client, not only for one specific parameter? Of course we would need a different syntax, then.

It would be more or less trivial in `openqa-cli` to extend what I'm doing for `template=` to all parameters, but what worries me is this breaking other use cases. For example `SOME_TEST_SETTING=file:///path/to/file/in/SUT` in the command line could produce a file not found error or send to the test a wrong setting, so - as you say - a different syntax may be needed.

What I thought about initially was a list of parameters that can be read from command line or from files, and wrap those in a loop as:
```
my @file_supported_arguments = qw(template other_argument_that_can_read_from_file etc);
...
foreach my $argument (@file_supported_arguments) {
    next unless defined $params{$argument};
    foreach my $tval (@{$params{$argument}) {
        ...
    }
}
```
That was the reason for my question in the very last line of the description.

perlpunk:
Maybe there could be a new cli option. I can't think of a good name, but lets say `--param-file`:
```
job_templates_scheduling/1 post --form schema=JobTemplates-01.yaml
        preview=0 --param-file template=foo.yaml
```
(If there is more than one file parameter, the option can be repeated.)

lib/OpenQA/Script/Client.pm
okurz: that would be ok as well

okurz:
> > What you put into the description of the pull request could also be in the commit message details.
> 
> OK. I'll add it. Should I include the examples as well?

Sure

> > Also see [https://github.com/os-autoinst/os-autoinst-distri-opensuse/blob/master/CONTRIBUTING.md#coding-style](https://github.com/os-autoinst/os-autoinst-distri-opensuse/blob/master/CONTRIBUTING.md?rgh-link-date=2021-03-24T11%3A37%3A14Z#coding-style) for more details
> 
> So, besides updating copyright years and more information on the commit message, did you have anything else in mind?

well, you already saw the inline comments so that's fine

perlpunk:
btw, since you repeatedly use `YaML` with a lowercase "a" - it should be all uppercase `YAML` - "YAML Ain't Markup Language"

alvarocarvajald:
> btw, since you repeatedly use `YaML` with a lowercase "a" - it should be all uppercase `YAML` - "YAML Ain't Markup Language"

Fixing it ... but "ain't" is such an ugly contraction .... it doesn't deserve capitalization ;)

lib/OpenQA/Script/Client.pm
alvarocarvajald: Understood. I will drop the changes from `openqa-client`, and then explore the `--param-file` suggestion by @perlpunk

lib/OpenQA/Command.pm
```diff
@@ -89,6 +90,13 @@ sub parse_params {
         push @{$params{$1}}, $2;
     }
 
+    return \%params unless (defined $params{template});
+
+    foreach my $tval (@{$params{template}}) {
```
kalikiana, line 95: I feel like `reference` should get the same treatment since you will run into the same problems.

kalikiana: If you haven't used it: `reference` is the expected current YAML document that the openQA instance knows about to avoid race conditions.

alvarocarvajald: But isn't this `reference` in the openQA instance itself? I mean, on the server side? Not sure how this is impacted by command line arguments processing on the client side.

alvarocarvajald:
> Maybe there could be a new cli option. I can't think of a good name, but lets say `--param-file`:
> 
> ```
> job_templates_scheduling/1 post --form schema=JobTemplates-01.yaml
>         preview=0 --param-file template=foo.yaml
> ```
> 
> (If there is more than one file parameter, the option can be repeated.)

Added a `--param-file` option as suggested.

lib/OpenQA/Command.pm
alvarocarvajald: I changed that part of the code. Not sure if your comment still applies.

lib/OpenQA/CLI/api.pm
```diff
@@ -105,26 +106,34 @@ sub command {
...
+        --apisecret <secret>      API secret
+    -a, --header <name:value>     One or more additional HTTP headers
+    -D, --data-file <path>        Load content to send with request from file
+    -d, --data <string>           Content to send with request, alternatively
+                                  you can also pipe data to openqa-cli
+    -f, --form                    Turn JSON object into form parameters
+        --host <host>             Target host, defaults to http://localhost
+    -h, --help                    Show this summary of available options
+    -j, --json                    Request content is JSON
+        --param-file <param=file> Load content of params from files instead of
```
kraih, line 128: Options are ordered alphabetically.

lib/OpenQA/Command.pm
```diff
@@ -81,14 +82,19 @@ sub parse_headers {
...
 
     my %params;
-    for my $arg (@args) {
+    for my $arg (@{$args}) {
         next unless $arg =~ /^([[:alnum:]_\[\]\.]+)=(.*)$/s;
         push @{$params{$1}}, $2;
     }
 
+    for my $arg (@{$param_file}) {
+        next unless $arg =~ /^([[:alnum:]_\[\]\.]+)=(.*)$/s;
```
kraih, line 94: The regex is long enough that it should be factored out into a global for reuse.
```perl
$PARAM_RE = qr/^([[:alnum:]_\[\]\.]+)=(.*)$/s;
```
and then here
```perl
next unless $arg =~ $PARAM_RE;
```

lib/OpenQA/CLI/api.pm
alvarocarvajald: Fixed.

lib/OpenQA/Command.pm
alvarocarvajald: Done.

lib/OpenQA/Command.pm
```diff
@@ -81,14 +83,19 @@ sub parse_headers {
...
-    for my $arg (@args) {
-        next unless $arg =~ /^([[:alnum:]_\[\]\.]+)=(.*)$/s;
+    for my $arg (@{$args}) {
+        next unless $arg =~ $PARAM_RE;
         push @{$params{$1}}, $2;
     }
 
+    for my $arg (@{$param_file}) {
+        next unless $arg =~ $PARAM_RE;
+        push @{$params{$1}}, path($2)->slurp;
```
kraih, line 96: Actually, this would probably more efficient as `push @{$params{$1}}, {file => $2};`, since it allows `Mojo::UserAgent` to stream the file in chunks, instead of loading everything into memory.

alvarocarvajald: That doesn't seem to work as expected:
```
# openqa-cli api --host http://mango.qa.suse.de --apikey **** --apisecret **** -X POST \
    job_templates_scheduling/2 schema=JobTemplates-01.yaml preview=1 \
    --param-file template=sle15_sp3_ha.yml
400 Bad Request
{"error":["Job template name 'create_hdd_ha_textmode' with sle-15-SP3-Online-aarch64 and aarch64 is already used in job group 'HA'"],"error_status":400,"id":2,"job_group_id":2}
# vi ./lib/OpenQA/Command.pm
# openqa-cli api --host http://mango.qa.suse.de --apikey **** --apisecret **** -X POST \
    job_templates_scheduling/2 schema=JobTemplates-01.yaml preview=1 --param-file template=sle15_sp3_ha.yml
400 Bad Request
{"error":[{"message":"Expected object - got string.","path":"\/"}],"error_status":400}
```
Perhaps I am mistaken, but don't we require to load the whole file into memory before submitting the request to the server?

lib/OpenQA/CLI/api.pm
```diff
@@ -105,26 +106,34 @@ sub command {
...
+    -D, --data-file <path>        Load content to send with request from file
+    -d, --data <string>           Content to send with request, alternatively
+                                  you can also pipe data to openqa-cli
+    -f, --form                    Turn JSON object into form parameters
+        --host <host>             Target host, defaults to http://localhost
+    -h, --help                    Show this summary of available options
+    -j, --json                    Request content is JSON
+        --osd                     Set target host to http://openqa.suse.de
+        --o3                      Set target host to https://openqa.opensuse.org
+        --param-file <param=file> Load content of params from files instead of
```
Martchus, line 130: In consistency with `--data-file <path>`:
```suggestion
        --param-file <path> Load content of params from files instead of
```

perlpunk: But we need the parameter name before the path

alvarocarvajald: Use cases are different for both. `--data-file` takes one file and only one file, typically in JSON format, with a list of parameters and its values; in conjunction with `--form` the contents of this file take precedence over parameters passed on the command line itself.

For the examples I've used above, an example of a data file would be:
```
{
    "schema" : "JobTemplates-01.yaml",
    "preview" : "1"
}
```
And not sure if we could add a `template` value there .... in my tests JSON and YAML together in a file did not mix too well (or at all), so job template probably needs to be translated from YAML to JSON in order to include it in the data file, which defeats the purpose of having the template in YAML in the first place.

`--param-file` OTOH can be used several times to read several parameters from several files. On the above examples, think of: `--param-file schema=myschema.txt --param-file preview=mypreview.txt --param-file template=foo.yml` with `myschema.txt` and `mypreview.txt` containing `JobTemplates-01.yaml` and `1` respectively. Of course, it would not make sense to have files for parameters with short values, but it does for something such as the YAML template.

Hope this clears it up.

lib/OpenQA/Command.pm
perlpunk: Yeah, now it is not needed anymore. But with the previous approach it would have been, since the `reference` parameter can be given in the same way as the `template` parameter to avoid overwriting intermediate changes.

lib/OpenQA/CLI/api.pm
Martchus: Sorry, I've missed that.

okurz (approved):

t/43-cli-api.t
```diff
@@ -495,4 +495,36 @@ subtest 'PIPE input' => sub {
...
+    is_deeply \@result, [0], 'zero exit code';
+    is $stderr, '', 'quiet';
+    my $yaml_text = "products: {}\nscenarios: {}";
+    ($stdout, $stderr, @result) = capture sub {
+        $api->run(@host, '-X', 'POST', 'job_templates_scheduling/1', 'schema=JobTemplates-01.yaml', 'preview=1',
+            "template=$yaml_text");
+    };
+    $data = decode_json $stdout;
+    is $data->{job_group_id}, '1', 'YAML Text';
+    is_deeply \@result, [0], 'zero exit code';
```
okurz, line 511: I already approved the PR but in case you still have a bit of time to improve or revisit this anyway then I suggest to improve the test strings with a bit more detail and also make them unique, e.g. 'zero exit code' is used multiple times. It's easier to pinpoint the corresponding source code from failing tests if the test messages are unambiguous

alvarocarvajald: Sure. Will do that. Used mostly the same names as it seemed to be some sort of standard, but I'm happy to change them.

lib/OpenQA/Command.pm
alvarocarvajald: Ah, now I get it. `reference` is also a parameter.

perlpunk (approved):

kalikiana (approved):

lib/OpenQA/Command.pm
Martchus: By the way, this is actually still a valid point for improvement. Not sure how to make it work exactly within Mojolicious but generally it would be preferable to avoid reading the whole file into memory.

kraih: I think i see now why it didn't work. The current implementation of `{file => '...'}` forces it to be a file upload with filename. Which doesn't work with the server side validation we have that expects a simple form value.

Martchus: Although the memory usage of the YAML parser on the server-side might just be the next thing to explode then :-)

kraih: I've made an upstream change in Mojolicious to allow this in the future. https://github.com/mojolicious/mojo/compare/529af01cdb4d...884d8f1ee196

PR #3811, opened 2021-03-24
Replace js-beautifier with standard

kalikiana:
Replace js-beautifier with standard
    
- Makefile targets are the same as before.
- standard needs to be installed globally via npm
- npm/ standard is installed in CircleCI
    
See https://github.com/standard/standard


Martchus (commented):
I must say that I'd prefer to keep the semicolons and the 4-spaces indentation so the JavaScript code looks more in-line with the Perl code. It is also not nice to add a dependency which is only available via NPM just for the sake of styling.

ilausuch (approved):

Martchus (changes requested):
See my previous comment

okurz:
I suggest we don't introduce a second zypper call but the other changes look unavoidable.

@Martchus see https://github.com/standard/standard#i-disagree-with-rule-x-can-you-change-it :)

Martchus:
I don't like one-size-fits-all-solutions. In our case it still makes sense to use a style which is more similar to Perl. Note that the changes are completely avoidable. It is as easy as closing the PR.

okurz:
It's funny to see that semistandard exists but I don't buy into the idea of "make javascript look similar to perl". Anyway, neither +1 nor -1 from me yet, let others have their opinion as well

okurz:
as discussed a suggestion was to use https://github.com/mojolicious/server-starter/blob/master/.eslintrc.json and use eslint

Martchus:
I guess it is now obsoleted by https://github.com/os-autoinst/openQA/pull/3900.

PR #3815, opened 2021-03-30
Improve error handling in cli/clone-job

Martchus:
See particular commit messages and https://progress.opensuse.org/issues/90038


Martchus:
By the way, when testing this via `script/openqa-clone-job … --skip-download --clone-children https://openqa.opensuse.org/tests/1683985` where `1683985` is the chained parent `create_hdd_gnome@64bit` I've got the following:

screenshot_20210330_174653

This is the original cluster:

screenshot_20210330_174916

So it did not clone any grand children. That might simply be a known limitation but it is weird that the parent appears as grand child instead. Did someone go on a time-traveling bug with openQA jobs?

Amrysliu:
> So it did not clone any grand children. That might simply be a known limitation but it is weird that the parent appears as grand child instead. Did someone go on a time-traveling trip with openQA jobs?
> 

It did not clone any grand children? From the picture, the new job has many children and their status is `user canceled`. maybe the children should not be triggered in this scenario. Is `--clone-children` only used for `parallel` dependent jobs?

okurz:
Looks good but as I know you are up for the challenge: Can you bump to 100% patch statement coverage? :)

Martchus:
I'm currently working on that and I thought it would make sense to cover code paths for dependencies as well. The test looks ok but I can actually reproduce the issue from the previous screenshot with it so it is failing. I'll see whether I can fix it.

Martchus:
Since it is not that trivial to fix I made the new test tolerate the buggy behavior and added an explaining comment. If we'd wanted to fix the problem than there's now at least already a test which one could start with.

PR #3816, opened 2021-03-31
Fix request routing for Mojolicious 9.11

AdamWill:
Mojolicious 9.11 disables request "format detection" - where a route declared for `/foo` will be used for `/foo.iso` by default, with the 'format' included in the request stash - to improve security. There are several cases where we relied on this; these commits change those routes to explicitly declare the formats they accept. I hope this covers everything, it at least makes all the tests we run on Fedora package build work with Mojolicious 9.15.

### DIFF
--- a/lib/OpenQA/WebAPI.pm
+++ b/lib/OpenQA/WebAPI.pm
@@ -131,7 +131,8 @@ sub startup ($self) {
     $r->get('/search')->name('search')->to(template => 'search/search');
 
     $r->get('/tests')->name('tests')->to('test#list');
-    $r->get('/tests/overview')->name('tests_overview')->to('test#overview');
+    $r->get('/tests/overview' => [format => ['json', 'html']])->name('tests_overview')
+      ->to('test#overview', format => 'html');
     $r->get('/tests/latest')->name('latest')->to('test#latest');
 
     $r->get('/tests/export')->name('tests_export')->to('test#export');
@@ -195,7 +196,8 @@ sub startup ($self) {
 
     $r->get('/group_overview/<groupid:num>' => [format => ['json', 'html']])->name('group_overview')
       ->to('main#job_group_overview', format => 'html');
-    $r->get('/parent_group_overview/<groupid:num>')->name('parent_group_overview')->to('main#parent_group_overview');
+    $r->get('/parent_group_overview/<groupid:num>' => [format => ['json', 'html']])->name('parent_group_overview')
+      ->to('main#parent_group_overview', format => 'html');
 
     # Favicon
     $r->get('/favicon.ico'             => sub ($c) { $c->render_static('favicon.ico') });
--- a/t/22-dashboard.t
+++ b/t/22-dashboard.t
@@ -139,8 +139,8 @@ sub check_test_parent {
     is_deeply(
         \@urls,
         [
-            '/tests/overview?distri=opensuse&version=13.1&build=0091&groupid=1001',
-            '/tests/overview?distri=opensuse&version=13.1&build=0091&groupid=1002'
+            '/tests/overview.html?distri=opensuse&version=13.1&build=0091&groupid=1001',
+            '/tests/overview.html?distri=opensuse&version=13.1&build=0091&groupid=1002'
         ],
         'link URLs'
     );
@@ -339,12 +339,12 @@ $t->get_ok('/dashboard_build_results?limit_builds=20&show_tags=0')->status_is(20
 is(scalar @urls, 12, 'now builds belong to different versions and are split');
 is(
     $urls[1]->attr('href'),
-    '/tests/overview?distri=suse&version=14.2&build=87.5011&groupid=1001',
+    '/tests/overview.html?distri=suse&version=14.2&build=87.5011&groupid=1001',
     'most recent version/build'
 );
 is(
     $urls[-1]->attr('href'),
-    '/tests/overview?distri=opensuse&version=13.1&build=0091&groupid=1002',
+    '/tests/overview.html?distri=opensuse&version=13.1&build=0091&groupid=1002',
     'oldest version/build still shown'
 );
 
@@ -365,7 +365,7 @@ subtest 'build which has jobs with different DISTRIs links to overview with all 
     my $first_url = $urls[1]->attr('href');
     is(
         $first_url,
-        '/tests/overview?distri=opensuse&distri=suse&version=14.2&build=87.5011&groupid=1001',
+        '/tests/overview.html?distri=opensuse&distri=suse&version=14.2&build=87.5011&groupid=1001',
         'both distris present in overview link'
     );
     $job_with_different_distri->delete;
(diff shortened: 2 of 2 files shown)


AdamWill:
@kraih for review.

AdamWill:
I'm not sure if we need to handle this for asset downloads too, that might be tricky if so. At least no tests fail on it. I'll test it manually in a bit too.

AdamWill:
Ah, so one problem appears: this syntax apparently doesn't work on older Mojolicious, so everything this *fixes* on 9.11+, it *breaks* on older versions (not sure when this syntax got added, but whenever it was, I guess anything older than that is broken...)

okurz (changes requested):
Well, unfortunately quite some tests fail. And even if this works it is far from obvious why the changed routes would need an explicit statement about the format. It would be ok for me in case you can add a test that ensures everything that would be necessary for the python client. We would likely have reacted already on any breakage of tests within https://build.opensuse.org/package/show/devel:openQA/python-openqa_client but so far that package shows no problems as well.

kraih:
The syntax works just fine on older versions, the patch probably doesn't cover all the cases. I just did the same change for [Cavil](https://github.com/openSUSE/cavil/commit/9baf2b2c831ee14d69bdb4153d9994d9fb05d63d) and tests pass fine on 9.01 and 9.13.

kraih:
The problem itself that forced us to disable format detection was a security issue in versions prior to 9.11. You can test for it by requesting [https://openqa.opensuse.org/.json](https://openqa.opensuse.org/.json) and triggering a 500 error.

t/lib/OpenQA/Test/Utils.pm
```diff
@@ -131,7 +131,7 @@ sub fake_asset_server {
             $c->render(text => 'This file was not compressed!', format => 'txt');
         });
     $r->get(
-        '/tests/:job/asset/:type/:filename' => sub {
+        '/tests/:job/asset/:type/:filename' => [format => ['qcow2']] => sub {
```
kraih, line 134: This looks wrong to me. There's no content negotiation here that would need a format. It was probably kinda broken before already, and the `:filename` should have been a `*filename`, to capture the whole filename.

kraih:
The right approach to fixing this would be to go through all cases where we use content negotiation (like `->respond_to()`) in controllers, and then add the formats from there to the route.

t/lib/OpenQA/Test/Utils.pm
AdamWill: Well, I was just following the guide in the changelog, which didn't mention that being a possibility. It clearly broke because of the change, and this was what the changelog suggested doing to fix it, and it fixed it, so...

AdamWill:
@kraih that is literally what I did. The tests all pass when built against Mojo 9.15. They only fail when building against older Mojo. I don't know why.

kraih:
@AdamWill I'll take a look and run some tests.

t/lib/OpenQA/Test/Utils.pm
AdamWill: changed, though.

AdamWill:
OK, so this latest version doesn't change the route for `tests/overview` or `parent_group_overview`, because nothing in the tests accesses those with an extension.

The basic trade-off for any affected route here, AFAICT, is: if we don't change it, accessing anything via that route with an extension (`.json` or `.html`) will stop working with Mojo 9.11+, but if we change it, generated URLs using that route will have an extension where they did not before. For each affected route it's a question of which impact we want. So far as the *tests* go, if a test uses a URL via an affected route with a file extension, we can *either* change the route *or* change the test not to do that. So to an extent the question is, "why does the test do that?"

As it stands, this PR changes the route for each case where the tests use a URL with an extension:

* `/dashboard_build_results.json` is used by `t/api/04-jobs.t`, `t/api/10-jobgroups.t`, and `t/ui/14-dashboard.t`
* `/admin/workers.json` is used by `t/api/12-admin-workers.t`
* `/group_overview/(someid).json` is used by `t/api/09-comments.t` and `t/api/04-jobs.t`

I'm not 100% clear *why* each of those tests uses a URL with an extension, though. I can't spot a case in which we would generate a link that would have an extension. I think they may date back to very early implementation of JSON representations, and didn't actually match how those routes would 'normally' be accessed. For instance, the use of `/group_overview/1001.json` in `t/api/04-jobs.t` ultimately traces back to 70adf2bab9feac765969cdf708e99291a440ec44 , but even at that time, the route was implemented the same as it is now, and the actual representation was added in 37426074459a9b07d02bd4e2a116cddbfe68febc , which says "JSON response can be triggered by adding HTTP header 'Accept: application/json'".

I'll look into the other cases, but if we've never actually documented or templated access to a route with an extension, arguably we should just fix the tests not to use one and leave the route alone, rather than changing it. I might amend this PR further after doing some more research.

kraih:
All those URLs are generated with `url_for()` referencing the route names. The mechanism exists specifically so changes in the router do not require template changes and propagate automatically. So i'd argue that this is specifically the reason why we use `url_for` there in the first place. 😉

AdamWill:
OK, so having thought about it a bit, I think I'd vote for choice 1. So I've changed the PR (back) to implementing that; it now updates *all* routes that could previously work this way so they still do (and adjusts the tests accordingly). @kraih 's #3820 (as I write this) implements choice 2. We don't have a PR for choice 3, but it would be easy to do if someone wants it.

AdamWill:
oh, there are a couple of POST / DELETE routes that I didn't update. I don't know if it's worth doing those too.

Martchus:
I guess for POST/DELETE we're using the API anyways and the API should still be tied to JSON, right? Not sure whether the failing fullstack test is related, though.

AdamWill:
It *shouldn't* be. I find it hard to interpret fullstack test failures, though. Can we just kick it and make it try again? I can't offhand think of a reason why it would fail on this version of the PR but pass on a previous version.

okurz:
well, because I like you I will do so and retrigger the full-stack test ;) But please keep in mind that we invested quite some effort to make the full-stack test rather fast and very stable. We should not accept flakyness.

For further reference the failed full stack run is https://app.circleci.com/pipelines/github/os-autoinst/openQA/6118/workflows/b9f9ca33-472f-4acf-a67a-4486e9e5a62c/jobs/57756/tests

now restarted as https://app.circleci.com/pipelines/github/os-autoinst/openQA/6118/workflows/08b44232-4ba1-4a00-837a-bba62568cf45/jobs/57848

AdamWill:
I saw it, but I can't really make head or tail of the logs. It's not even clear to me what's going wrong.

okurz (approved):

AdamWill:
so since #1820 got merged I just rebased this to be, hopefully (if nothing got mangled) all the additional routes that #1820 *didn't* do. so with #1820 merged we have changed only routes that affected tests, this changes other routes that act the same but where no test happened to be using filename extension format detection.

okurz (approved):

Martchus (approved):

PR #3817, opened 2021-03-31
Set using relative path as the default operation

Amrysliu:
Users can disable this feature by setting `ABSOLUTE_TEST_CONFIG_PATHS=1`.

See: https://progress.opensuse.org/issues/90293#note-11

### DIFF
--- a/lib/OpenQA/Worker/Engines/isotovideo.pm
+++ b/lib/OpenQA/Worker/Engines/isotovideo.pm
@@ -344,7 +344,7 @@ sub engine_workit {
     my $target_name = path($casedir)->basename;
 
     $vars{ASSETDIR} //= OpenQA::Utils::assetdir();
-    $vars{CASEDIR}  //= $vars{ENABLE_RELATIVE_PATH} ? $target_name : $casedir;
+    $vars{CASEDIR}  //= $vars{ABSOLUTE_TEST_CONFIG_PATHS} ? $casedir : $target_name;
 
     if ($vars{CASEDIR} eq $target_name) {
         $vars{PRODUCTDIR} //= substr($productdir, rindex($casedir, $target_name));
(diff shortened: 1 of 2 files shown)


lib/OpenQA/Worker/Engines/isotovideo.pm
```diff
@@ -356,7 +356,7 @@ sub engine_workit {
     # if NEEDLES_DIR is an URL address, it means that users specify it and want to get the needles from URL.
     # In these two scenarios, doing symlink is useless and the job may incomplete because fail to do symlink.
     if (   $vars{NEEDLES_DIR}
-        && $vars{ENABLE_RELATIVE_PATH}
+        && !$vars{DISABLE_RELATIVE_PATH}
```
Martchus, line 359: I suppose the alternative would be:

```suggestion
        && ($vars{ENABLE_RELATIVE_PATH} // 1)
```

It would have the advantage of not changing the name of the setting afterwards.

Martchus: This was more an RFC but I like that you've turned it around.

Martchus:
Have we actually fixed the last problem we've encountered with the relative paths? I believe some LTP tests were affected. If not, maybe we should take care to add `ENABLE_RELATIVE_PATH=0` in the relevant tests.

Amrysliu:
> Have we actually fixed the last problem we've encountered with the relative paths? I believe some LTP tests were affected. If not, maybe we should take care to add `ENABLE_RELATIVE_PATH=0` in the relevant tests before merging this.

we fixed that issue. the related pr is https://github.com/os-autoinst/os-autoinst/pull/1634

Martchus (approved):

okurz (changes requested):
I would prefer an explicit variable to *disable* the feature instead of having a switch that sounds like one can enable but it's already enabled by default. I do not see a problem with renaming this feature switch. How about "ABSOLUTE_TEST_CONFIG_PATHS" (by default not enabled)?

Amrysliu:
> I would prefer an explicit variable to _disable_ the feature instead of having a switch that sounds like one can enable but it's already enabled by default. I do not see a problem with renaming this feature switch. How about "ABSOLUTE_TEST_CONFIG_PATHS" (by default not enabled)?

Modified. @Martchus are you ok with that?

okurz (approved):

Martchus (approved):

PR #3819, opened 2021-03-31
Move containers documentation to regular documentation

Martchus:
* Change format from Markdown to AsciiDoc
* Use proper cross-references
* Fix minor details/typos
* Mention that the `docker-compose` setup is still experimental
* See https://progress.opensuse.org/issues/76990


okurz (approved):

okurz:
@ilausuch

docs/ContainerizedSetup.asciidoc
```diff
@@ -131,39 +137,34 @@ ports:
...
+   - "443:443"
+   ```
+   You can change the exported port if 443 is already used in your computer, for
+   instance:
+   ```
+   - "10443:443"
+   ```
+
+2. Provide an SSL certificate:
+   ```
```
ilausuch, line 156: This volumes has to be in the section volumes of the docker-compose service but this section is currently missing on the yaml file and cannot be included if there are not volumes. So please include "volumes:" before these two lines.

ilausuch (changes requested):

docs/ContainerizedSetup.asciidoc
Martchus: The YAML looks like this:

```
    ports:
      - "9526:9526"
      # Uncomment for SSL access
      # - "443:443"
    environment:
      replicas: ${OPENQA_WEBUI_REPLICAS}
      # If is needed to set a certificate for SSL
      # - cert.crt:/etc/ssl/certs/opeqa.crt
      # - cert.key:/etc/ssl/certs/opeqa.key
```

So I would assume the documentation's first point asks me to uncomment the first comment line and the second point asks me to uncomment the two last comments. Not sure how volumes would come into play here. How would the YAML look like after following step 2? Maybe we should change the YAML so one has to uncomment lines if that's not already possible.

okurz:
https://github.com/os-autoinst/openQA/pull/3822 for the UI test instability. You can retrigger for this PR or rebase on my changes.

docs/ContainerizedSetup.asciidoc
ilausuch: I add this change here https://github.com/os-autoinst/openQA/pull/3821#issuecomment-811774254 but I am not sure is the place, feel free to add this here

Martchus: I would keep my PR as-is and when you do changes which require documentation updates you should to these documentation updates as part of these changes.

ilausuch: ok

ilausuch (approved):

PR #3820, opened 2021-03-31
Fix Mojolicious 9.11 compatibility

kraih:
Alternative to #3816. This should be the minimal fix to make openQA work with Mojolicious 9.01 and 9.11. [Here's](https://github.com/os-autoinst/openQA/compare/k/mojo_911) the same patch in a different branch running CI tests against 9.14.
```
 sudo perl - -M https://cpan.metacpan.org -n Mojolicious@9.14
 curl -L https://cpanmin.us
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100  295k  100  295k    0     0  5097k      0 --:--:-- --:--:-- --:--:-- 5097k
--> Working on Mojolicious
Fetching https://cpan.metacpan.org/authors/id/S/SR/SRI/Mojolicious-9.14.tar.gz ... OK
Configuring Mojolicious-9.14 ... OK
Building Mojolicious-9.14 ... OK
Successfully installed Mojolicious-9.14 (upgraded from 9.01)
1 distribution installed
```

### DIFF
--- a/lib/OpenQA/WebAPI.pm
+++ b/lib/OpenQA/WebAPI.pm
@@ -193,13 +193,15 @@ sub startup ($self) {
     # due to the split md5_dirname having a /
     $r->get('/image/:md5_1/:md5_2/.thumbs/#md5_basename')->to('file#thumb_image');
 
-    $r->get('/group_overview/<groupid:num>')->name('group_overview')->to('main#job_group_overview');
+    $r->get('/group_overview/<groupid:num>' => [format => ['json', 'html']])->name('group_overview')
+      ->to('main#job_group_overview', format => 'html');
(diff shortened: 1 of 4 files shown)


Martchus (approved):

AdamWill:
My current version of my PR is now similar to this, but also see my notes [here](https://github.com/os-autoinst/openQA/pull/3816#issuecomment-811371688). I think it's possible we don't need to change any routes, just change tests not to use file extensions, but I'm going to do a bit more research into whether there's any legitimate reason for the tests to be doing that first.

okurz (approved):

PR #3821, opened 2021-04-01
containers: Add healthchecks to organize the docker-compose startup

ilausuch:
Eventually the startup using docker-compose fails several seconds after all is up. The reason is because the webui do operations over the DB and because we have concurrent webui containers can cause race conditions

For instance:

```
Webui_1 failed to run SQL in
/usr/share/openqa/script/../dbicdh/PostgreSQL/deploy/92/001-auto-__VERSION.sql:
DBIx::Class::DeploymentHandler::DeployMethod::SQL::Translator::try {...} ():
 DBI Exception: DBD::Pg::db do failed:
ERROR:  duplicate key value violates unique constraint "pg_type_typname_nsp_index"
```

An example of this failure could be found at: #3818

The reason for the problem is that all webui containers try to create and initialize the database at the same time, and since all webui replicas are launched at the same time by the docker-compose and we have more than one, they overlap and generate the issue

The solution approach has been to create an initial webui container with a health check. This is the first container in the entire web UI system. So from it, using a cascade of appropriate dependencies so that the whole system starts up in an orderly way. And the health checks allow that the next step is not going to start until the previous one is not complete.

See [poo#89731](https://progress.opensuse.org/issues/89731)

### DIFF
--- a/container/webui/docker-compose.yaml
+++ b/container/webui/docker-compose.yaml
@@ -15,7 +15,14 @@ services:
       MODE: "scheduler"
       MOJO_LISTEN: "http://0.0.0.0:9529"
     depends_on:
-      - db
+      webui:
+        condition: service_healthy
+    healthcheck:
+      test: ["CMD", "curl", "-f", "http://localhost:9529"]
+      interval: 5s
+      timeout: 2s
+      retries: 3
+      start_period: 10s
 
   websockets:
     image: openqa_webui
@@ -27,7 +34,14 @@ services:
       MODE: "websockets"
       MOJO_LISTEN: "http://0.0.0.0:9527"
     depends_on:
-      - db
+      webui:
+        condition: service_healthy
+    healthcheck:
+      test: ["CMD", "curl", "-f", "http://localhost:9527"]
+      interval: 5s
+      timeout: 2s
+      retries: 3
+      start_period: 10s
 
   gru:
     image: openqa_webui
@@ -36,7 +50,15 @@ services:
     environment:
       MODE: "gru"
     depends_on:
-      - db
+      webui:
+        condition: service_healthy
+    entrypoint: "sh -c '/root/run_openqa.sh|tee /var/log/gru.log'"
+    healthcheck:
+      test: ["CMD", "grep", "started", "/var/log/gru.log"]
+      interval: 5s
+      timeout: 2s
+      retries: 3
+      start_period: 10s
 
   livehandler:
     image: openqa_webui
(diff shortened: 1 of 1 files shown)


Martchus (commented):
I'm not sure how these healthchecks work in detail but the commands you've been adding look reasonable. I've just re-triggered the successful CI check whether it is actually stable now. (We might want to re-trigger it a few more times to be sure.)

I assume that the error prevented by this change is:
```
 webui_1        | Waiting for DB creation
webui_1        | failed to run SQL in /usr/share/openqa/script/../dbicdh/PostgreSQL/deploy/92/001-auto-__VERSION.sql: DBIx::Class::DeploymentHandler::DeployMethod::SQL::Translator::try {...} (): DBI Exception: DBD::Pg::db do failed: ERROR:  duplicate key value violates unique constraint "pg_type_typname_nsp_index"
webui_1        | DETAIL:  Key (typname, typnamespace)=(dbix_class_deploymenthandler_versions_id_seq, 2200) already exists. at inline delegation in DBIx::Class::DeploymentHandler for deploy_method->deploy (attribute declared in /usr/lib/perl5/vendor_perl/5.26.1/DBIx/Class/DeploymentHandler/WithApplicatorDumple.pm at line 51) line 18
webui_1        |  (running line 'CREATE TABLE dbix_class_deploymenthandler_versions ( id serial NOT NULL, version character varying(50) NOT NULL, ddl text, upgrade_sql text, PRIMARY KEY (id), CONSTRAINT dbix_class_deploymenthandler_versions_version UNIQUE (version) )') at /usr/lib/perl5/vendor_perl/5.26.1/DBIx/Class/DeploymentHandler/DeployMethod/SQL/Translator.pm line 263.
webui_1        | DBIx::Class::Storage::TxnScopeGuard::DESTROY(): A DBIx::Class::Storage::TxnScopeGuard went out of scope without explicit commit or error. Rolling back. at /usr/share/openqa/script/openqa line 0
webui_1        | DBIx::Class::Storage::TxnScopeGuard::DESTROY(): A DBIx::Class::Storage::TxnScopeGuard went out of scope without explicit commit or error. Rolling back. at /usr/share/openqa/script/openqa line 0
```

It would make sense to mention the concrete error (maybe a little bit elided) in the commit message. You might also fix the grammar mistakes in the commit message btw.

ilausuch:
> I'm not sure how these healthchecks work in detail but the commands you've been adding look reasonable. I've just re-triggered the successful CI check whether it is actually stable now. (The 2nd run succeeded as well. We might want to re-trigger it a few more times to be sure.)

I rewrite the commit message, hope now is more clear :)

ilausuch:
Waiting to https://github.com/os-autoinst/openQA/pull/3827

kalikiana (approved):
I love how you use the dependencies here, and the health checks for specific services! Looks good to me!

Martchus (approved):

okurz (approved):

Amrysliu (approved):

PR #3827, opened 2021-04-07
containers: fix github test webui-docker-compose timeout

ilausuch:
The test webui-docker-compose exceeds 30 minutes and is cancelled with the message
"The job running on runner Hosted Agent has exceeded the maximum execution time of 30 minutes."

Increased to 60 minutes

https://progress.opensuse.org/issues/90767

### DIFF
--- a/.github/workflows/compose.yml
+++ b/.github/workflows/compose.yml
@@ -5,7 +5,7 @@ on: [push, pull_request]
 jobs:
   webui-docker-compose:
     runs-on: ubuntu-latest
-    timeout-minutes: 30
+    timeout-minutes: 60
     steps:
       - uses: actions/checkout@v2
       - name: Verify that containers can be composed


Martchus:
Wait, this runs 30 minutes and now even longer? I doubt the docker-compose tests should be that time consuming. In other places we try to make the tests more efficient, please try to make these tests more efficient as well.

Martchus:
Now it says "compose / webui-docker-compose (pull_request) Successful in 29m" so the previous limit was really close. I think this is already too long. Unfortunately the log is not verbose and without timestamps so it isn't clear where the time is spent. Why does it take so long?

okurz (approved):
I agree that 30m is way too long for a good test however in most cases we would not bother as reviewers see the PR after 30m. And OBS tests can sometimes take longer as well as circleCI when we have many tests queued. So this PR would help in these cases.

okurz:
I noted down the problem of "test takes too long" in https://progress.opensuse.org/issues/90767#note-3 and I will merge this PR as quick-fix

ilausuch:
> Wait, this runs 30 minutes and now even longer? I doubt the docker-compose tests should be that time consuming. In other places we try to make the tests more efficient, please try to make these tests more efficient as well.

Please refer to my comment about this at https://progress.opensuse.org/issues/89731#note-19

PR #3830, opened 2021-04-08
Fix content negotiation to use Accept headers again

kraih:
Currently only extensions like `.json` work with content negotiation, since we've made `.html` the default. That means content negotiation headers like `Accept: application/json` do nothing. We'll have to wait for `perl-Mojolicious` 9.11+ to actually [make it to factory](https://build.opensuse.org/request/show/880885) though before this PR can be applied.

Progress: https://progress.opensuse.org/issues/90818


.circleci/config.yml
```diff
@@ -53,6 +53,7 @@ aliases:
         else
           sudo rpm -i -f $(find /var/cache/zypp/packages/ | grep '.rpm$')
         fi
+        curl -L https://cpanmin.us | sudo perl - -M https://cpan.metacpan.org -n Mojolicious
```
kraih, line 56: This needs to be removed before we can apply the PR! ⚠️

kraih:
Ok, Mojolicious 9.14 is now in factory, so i've removed the Circle CI hack and the PR should be ready for merge.

perlpunk (approved):

Martchus (approved):
Added "not-ready" because we should still wait until it is in TW.

okurz (approved):

Martchus:
It is now in TW.

okurz:
> It is now in TW. Can you check whether the failing fullstack test is related?

what happened?

kraih:
What failing fullstack test? 🤷

AdamWill:
Ah, well, this leaves me in a bit of an awkward spot by dropping support for Mojo 8. Fedora 34 released with Mojo 8. So now I either have to do a major, incompatible version bump of Mojo in Fedora 34, or introduce some kind of new "mojolicious9" package, or run our openQA instance on Rawhide, or leave it outdated (or revert this patch) for the next six months. None of those feel like fantastic options...:(

PR #3833, opened 2021-04-08
Improve container documentation

Martchus:
* Separate instructions for "Fedora" and "docker-compose" setups more from
  each other because the interleaving made it quite hard to follow
* Improve wording
* Fix references to "Keeping all data …" sections
* Merge section explaining the data storage with the directory structure
  setup because these seem to be very related
* See https://progress.opensuse.org/issues/76990

### DIFF
--- a/docs/ContainerizedSetup.asciidoc
+++ b/docs/ContainerizedSetup.asciidoc
@@ -56,6 +56,8 @@ setup SELinux properly. If you are having problems with it, run this command:
 You can either build the images locally or use Fedora images from Docker Hub.
 
 For the `docker-compose` setup it is required to build the images locally.
+However, it is done automatically so this section can be skipped unless one
+wants to build the containers manually.
 
 === Download Fedora-based images from the Docker Hub
 
@@ -63,7 +65,7 @@ For the `docker-compose` setup it is required to build the images locally.
     docker pull fedoraqa/openqa_webui
     docker pull fedoraqa/openqa_worker
 
-=== Build openSUSE-based images locally (mandatory when using docker-compose)
+=== Build openSUSE-based images locally
 
     docker build -t openqa_data ./openqa_data
     docker build -t openqa_webui ./webui
--- a/docs/ContainerizedSetup.asciidoc
+++ b/docs/ContainerizedSetup.asciidoc
@@ -232,7 +232,8 @@ instance:
    - cert.key:/etc/ssl/certs/openqa.key
 ----
 
-3. Modify `nginx.cfg` to use this certificate. Uncomment the lines
+3. Modify `nginx.cfg` to use this certificate. Append ` ssl` to the listen command
+for port 443 and uncomment the lines:
 +
 [source]
 ----
--- a/docs/ContainerizedSetup.asciidoc
+++ b/docs/ContainerizedSetup.asciidoc
@@ -15,6 +15,31 @@ The second one uses `docker-compose` to deploy a complete web UI with HA for the
 API and the dashboard. This setup is under development and currently considered
 proof-of-concept. Do *not* expect that everything works yet.
 
+== Get container images
+
+You can either build the images locally or use Fedora images from Docker Hub.
+
+For the `docker-compose` setup it is required to build the images locally.
+However, it is done automatically so this section can be skipped unless one
+wants to build the containers manually.
+
+=== Download Fedora-based images from the Docker Hub
+
+    docker pull fedoraqa/openqa_data
+    docker pull fedoraqa/openqa_webui
+    docker pull fedoraqa/openqa_worker
+
+=== Build openSUSE-based images locally
+
+    docker build -t openqa_data ./openqa_data
+    docker build -t openqa_webui ./webui
+    docker build -t openqa_worker ./worker
+
+    # for docker-compose setup
+    docker build -f webui/Dockerfile-lb -t openqa_webui_lb ./webui
+
+== Setup with Fedora-based images
+
 == Data storage and directory structure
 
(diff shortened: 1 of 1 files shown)


docs/ContainerizedSetup.asciidoc
```diff
@@ -9,65 +9,72 @@ This section describes two ways to deploy the containers for the openQA web UI
...
+=== Build openSUSE-based images locally (mandatory when using docker-compose)
 
-    # Fedora and docker-compose
     docker build -t openqa_data ./openqa_data
     docker build -t openqa_webui ./webui
     docker build -t openqa_worker ./worker
 
-    # docker-compose only
+    # for docker-compose setup
     docker build -f webui/Dockerfile-lb -t openqa_webui_lb ./webui
```
ilausuch, line 73: I think we should mention that the images can be built using docker-compose build

docs/ContainerizedSetup.asciidoc
```diff
@@ -109,13 +102,82 @@ next two steps, you will set an OpenID provider (if necessary), create the API
...
+repository in `/tools` directory. Note that you have to have
+`libguestfs-tools` and `libguestfs-xfs` installed.
+
+== Setup with openSUSE-based images and docker-compose
+
+=== Run the data and web UI containers in HA mode
 
     # To create the containers
     # in the directory openQA/docker/webui execute:
     docker-compose up -d
```
ilausuch, line 178: Here and related with I mentioned previously, docker-compose up is going to build the images if don't exist. Maybe is interesting to comment that too because simplifies all the setup

docs/ContainerizedSetup.asciidoc
```diff
@@ -9,65 +9,72 @@ This section describes two ways to deploy the containers for the openQA web UI
...
 proof-of-concept. Do *not* expect that everything works yet.
 
-If a section or action is tagged as "Fedora", it should be applied only for a
-Fedora deployment. If the tag "docker-compose" is used it is supposed to be used
-for the `docker-compose` deployment.
+== Data storage and directory structure
+
+Our intent was to create universal `webui` and `worker` containers and move
+all data storage and configurations to a third container, called `openqa_data`.
+`openqa_data` is a so called
```
ilausuch, line 22: We have to consider that openqa_data is not longer necessary in docker-compose setup. See https://github.com/os-autoinst/openQA/blob/master/container/webui/docker-compose.yaml. The volumes are covered here https://github.com/os-autoinst/openQA/blob/22b668bd77a95f557ab79d8cff89d5e6eaf006a3/container/webui/docker-compose.yaml#L6 and the web UI configuration files in the webUI/config directory https://github.com/os-autoinst/openQA/blob/22b668bd77a95f557ab79d8cff89d5e6eaf006a3/container/webui/docker-compose.yaml#L13

docs/ContainerizedSetup.asciidoc
```diff
@@ -9,65 +9,72 @@ This section describes two ways to deploy the containers for the openQA web UI
...
+<<ContainerizedSetup.asciidoc#_keeping_all_data_in_the_data_volume_container,Keeping all data in the Data Volume Container>>
+and
+<<ContainerizedSetup.asciidoc#_keeping_all_data_on_the_host_system,Keeping all data on the host system>>
+respectively.
+
+Otherwise, when you want to have the big files (isos and disk images, tests and
+needles) outside of the Volume container, you should create this file
+structure from within the directory you are going to execute the container:
+
+    mkdir -p data/factory/{iso,hdd} data/tests
```
ilausuch, line 47: Empirically I checked that there are missing directories
mkdir -p data/factory/{iso,hdd,other,tmp} data/tests data/testresults
See https://github.com/os-autoinst/openQA/blob/22b668bd77a95f557ab79d8cff89d5e6eaf006a3/container/webui/docker-compose.yaml#L6
Maybe with the "chmod -R 777 data" documentation line is covered, but not sure, this would be checked

docs/ContainerizedSetup.asciidoc
ilausuch: Or see my next comment, docker-compose is going to build the images if don't exist when call docker-compose up

ilausuch (changes requested):

docs/ContainerizedSetup.asciidoc
Martchus: That would be good to know. I suppose I'll have to try it out myself to see how it is actually supposed to be used *without* building any images before manually.

docs/ContainerizedSetup.asciidoc
Martchus: Ok, so that section is completely irrelevant when using docker-compose. I'll move it into the Fedora-specific part then.

docs/ContainerizedSetup.asciidoc
Martchus: Ok, I'll add this as well.

Martchus:
@ilausuch I'll do the changes in a separate commit to separate the restructuring from the actual content updating which I haven't done so far.

docs/ContainerizedSetup.asciidoc
Martchus: I moved the section completely to the Fedora-specific setup. That's better anyways because it also references other sections which seem to be Fedora-specific.

okurz (approved):

ilausuch (approved):

docs/ContainerizedSetup.asciidoc
Martchus: I moved it under the Fedora section after all.

PR #3840, opened 2021-04-13
CI: Fix db & webui_init startup

ilausuch:
In #89731 we introduced a initial webui container in charge of initializing
the database. We have a test where the health check failed
https://github.com/os-autoinst/openQA/pull/3838/checks?check_run_id=2329551052

There are several detected problems:
- The db healthcheck command doesn't generate an errorcode when fails
echo 'select * from api_keys;' | psql -U openqa openqa
- There is a deathlock between db and webui: db healthcheck checks that
the table api_keys exists, but this table is created by webui_init, and
webui_init depends on the healthy start of the db to start

Solutions:

- Fix the DB healthcheck
- DB and webui_init are started at the same time
- webui pool starts when DB is healthy

https://progress.opensuse.org/issues/91046

### DIFF
--- a/container/webui/docker-compose.yaml
+++ b/container/webui/docker-compose.yaml
@@ -89,7 +89,7 @@ services:
       MODE: "webui"
       MOJO_LISTEN: "http://0.0.0.0:9526"
     depends_on:
-      webui_db_init:
+      db:
         condition: service_healthy
     deploy:
       replicas: ${OPENQA_WEBUI_REPLICAS}
@@ -109,9 +109,6 @@ services:
     environment:
       MODE: "webui"
       MOJO_LISTEN: "http://0.0.0.0:9526"
-    depends_on:
-      db:
-        condition: service_healthy
     entrypoint: "sh -c 'chmod -R a+rwX /data/{factory,testresults}; /root/run_openqa.sh'"
     healthcheck:
       test: ["CMD", "curl", "-f", "http://localhost:9526/login"]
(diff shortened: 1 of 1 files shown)


okurz (commented):
there are some typos in the git commit message, e.g. "heatlhcheck", "deahlock", "creeated", "webui_nit"

container/webui/docker-compose.yaml
```diff
@@ -109,9 +109,6 @@ services:
     environment:
       MODE: "webui"
       MOJO_LISTEN: "http://0.0.0.0:9526"
-    depends_on:
-      db:
-        condition: service_healthy
```
Martchus, line 114: So we're relying on `wait_for_db_creation()` in `run_openqa.sh` here? I suppose this is actually good enough to avoid the deadlock.

Martchus (changes requested):
The commit message has just too many typos (see comment of @okurz) and grammar mistakes (e.g. "doesn't generates").

container/webui/docker-compose.yaml
ilausuch: Yes, for this container we have the webui mode, so this line will be executed https://github.com/os-autoinst/openQA/blob/5c4cb4a68d872f81f8111fc297d3b37a214b6eed/container/webui/run_openqa.sh#L27 therefore will wait for db creation

ilausuch:
> there are some typos in the git commit message, e.g. "heatlhcheck", "deahlock", "creeated", "webui_nit"

Yes, sorry about that

okurz (approved):
actually it's "deadlock" not "deathlock" :D

Martchus (approved):

kalikiana (approved):

PR #3842, opened 2021-04-14
CI: Remove -q flag in docker-compose build in test_containers_compose 

ilausuch:
The flag -q is silencing all the logs when the container images
are building. So when it fails we don't have enough information about
the failure.

This commit removes this flag.

https://progress.opensuse.org/issues/91097

### DIFF
--- a/tools/test_containers_compose
+++ b/tools/test_containers_compose
@@ -5,7 +5,7 @@ set -euo pipefail
 for i in webui worker; do
   (
     cd container/$i/ &&
-    sudo docker-compose build -q --parallel &&
+    sudo docker-compose build --parallel &&
     sudo docker-compose up -d &&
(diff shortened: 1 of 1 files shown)


tools/test_containers_compose
```diff
@@ -5,7 +5,7 @@ set -euo pipefail
 for i in webui worker; do
   (
     cd container/$i/ &&
-    sudo docker-compose build -q --parallel &&
+    sudo docker-compose build --parallel &&
```
kalikiana, line 8: Where do I see that this worked? The tests pass and I'm not sure where to find the change

Martchus (approved):

tools/test_containers_compose
Martchus: Just look at the very verbose output within the CI pipeline compose / webui-docker-compose. It now shows all the zypper output which is likely what we're interested in.

okurz (approved):

PR #3844, opened 2021-04-15
CI: Retry if docker-compose build fails in webui-docker-compose test

ilausuch:
Eventually docker-compose build fails because of the zypper
respositories. This commit sets 3 retries before consider the test
failed

https://progress.opensuse.org/issues/90614

### DIFF
--- a/tools/test_containers_compose
+++ b/tools/test_containers_compose
@@ -5,8 +5,8 @@ set -euo pipefail
 for i in webui worker; do
   (
     cd container/$i/ &&
-    sudo docker-compose build --parallel &&
+     for retry in {2..0}; do sudo docker-compose build && break; echo "Remaining retries $retry"; done &&
     sudo docker-compose up -d &&
-   (docker-compose ps --services --filter status=stopped | grep "^[[:space:]]*$") || (docker-compose logs; sudo docker-compose ps; exit 1)
+    (docker-compose ps --services --filter status=stopped | grep "^[[:space:]]*$") || (docker-compose logs; sudo docker-compose ps; exit 1)
   ) || exit 1;
 done


ilausuch:
I am re-running the job to try to catch a failure that and prove how this works. I also could force a failure if it is necessary

ilausuch:
I forced a test that fails because the url of a repo is invalid
`[devel_openQA|http://download.opensuse.org/repositories/devel:openQA/openSUSE_Leap_15.22] Valid metadata not found at specified URL`

You can check that retries 3 times the build
https://github.com/os-autoinst/openQA/pull/3844/checks?check_run_id=2351562046

tools/test_containers_compose
```diff
@@ -5,8 +5,8 @@ set -euo pipefail
 for i in webui worker; do
   (
     cd container/$i/ &&
-    sudo docker-compose build --parallel &&
+    (for retry in 2 1 0; do sudo docker-compose build && exit 0; echo "Remaining reties $retry"; done; exit 1) &&
```
okurz, line 8: s/reties/retries/

Also how about:

```suggestion
    for retry in {2..0}; do sudo docker-compose build && break; echo "Remaining retries $retry"; done &&
```

You should not need the subshell with just "break"

okurz (changes requested):

tools/test_containers_compose
ilausuch: Done

tools/test_containers_compose
```diff
@@ -5,8 +5,8 @@ set -euo pipefail
 for i in webui worker; do
   (
     cd container/$i/ &&
-    sudo docker-compose build --parallel &&
```
okurz, line 8: with the removal of `--parallel', what is the impact?

ilausuch: This flag is being ignored
`Flag '--parallel' is ignored when building with COMPOSE_DOCKER_CLI_BUILD=1`

For instance, see this run
https://github.com/os-autoinst/openQA/runs/2342931733

kalikiana (approved):
Looks good to me

okurz (approved):

Martchus (approved):

PR #3847, opened 2021-04-16
Prevent recursive apparmor profile inclusion

Martchus:
* Remove `#include <local/usr.share.openqa.script.openqa>
  from that file itself so that it no longer recursively includes itsself
* Move the content of that local/ file to the main profile
* See https://progress.opensuse.org/issues/91284
* See https://bugzilla.suse.com/show_bug.cgi?id=1184838#c0


okurz (approved):
hm, well, the "local" file was done on purpose because the use case of the extended allowed rules might not be to the liking of all users. But as https://bugzilla.suse.com/show_bug.cgi?id=1184838#c0 mentioned that was not the idea of "local". Ok, good enough :)

PR #3854, opened 2021-04-21
Revert "worker: Make uploading test order on final upload (when stopping job) work"

okurz:
Reverts os-autoinst/openQA#3849 due to
https://progress.opensuse.org/issues/91461


Martchus:
We rolled back all workers, right? So let's see whether I can fix it later. If not we can merge this.

okurz:
> We rolled back all workers, right?

yes, o3+osd

> So let's see whether I can fix it later. If not we can merge this.

Sure. For that you give me an estimate how long you need and I give back a ridiculously lower time estimate and then again you finish it way faster :D

Martchus:
Shouldn't be required if we merge https://github.com/os-autoinst/openQA/pull/3855 instead.

PR #3856, opened 2021-04-21
worker: Avoid running optipng/upload twice

Martchus:
Making the worker go though every test module again during the final
upload (https://github.com/os-autoinst/openQA/pull/3849) introduced the
problem of uploading images again during the final upload. This should have
been prevented by the usual skipping of known images. However, since we now
re-read the test module results the checksums were re-computed during the
final upload which breaks the matching. That's because the checksums
actually differ due to running optipng before the image upload. This change
restores the matching by simply caching and reusing the md5sum from before
the optimization.

Related issue: https://progress.opensuse.org/issues/90152

### DIFF
--- a/lib/OpenQA/Worker/Job.pm
+++ b/lib/OpenQA/Worker/Job.pm
@@ -90,6 +90,7 @@ sub new {
     $self->{_files_to_send}                = {};
     $self->{_known_images}                 = [];
     $self->{_known_files}                  = [];
+    $self->{_md5sums}                      = {};
     $self->{_last_screenshot}              = '';
     $self->{_current_test_module}          = undef;
     $self->{_progress_info}                = {};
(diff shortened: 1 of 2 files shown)


kalikiana (approved):
That looks pretty straightforward.

I'm wondering if it might be simpler to do the optimization earlier. But you added a unit test for it so I'd say that's fine anyway and addresses the issue at hand.

Martchus:
I also considered doing the optimization earlier. However, it would have been a more involved code change and I want to avoid causing any further breakage. Besides, changing the way the md5sums of screenshots are computed would mean new screenshots would never match existing ones and hence would break de-duplication with existing screenshots.

coolo:
The point is that we only optimize images we're going to upload - i.e. those that aren't known already.

PR #3858, opened 2021-04-22
Add support for archived jobs

Martchus:
* See https://progress.opensuse.org/issues/91347 and particular commit messages
* Still WIP and related ticket is only "spike" ticket

### DIFF
--- a/lib/OpenQA/Schema/Result/Jobs.pm
+++ b/lib/OpenQA/Schema/Result/Jobs.pm
@@ -1243,7 +1243,7 @@ END_SQL
 sub num_prefix_dir {
     my ($self)    = @_;
     my $numprefix = sprintf "%05d", $self->id / 1000;
-    return catfile(resultdir(), $numprefix);
+    return catfile(resultdir($self->archived), $numprefix);
 }
 
 sub create_result_dir {
--- a/lib/OpenQA/Schema/Result/Jobs.pm
+++ b/lib/OpenQA/Schema/Result/Jobs.pm
@@ -37,7 +37,7 @@ use OpenQA::Jobs::Constants;
 use OpenQA::JobDependencies::Constants;
 use OpenQA::ScreenshotDeletion;
 use File::Basename qw(basename dirname);
-use File::Copy::Recursive 'dircopy';
+use File::Copy::Recursive qw();
 use File::Spec::Functions 'catfile';
 use File::Path ();
 use DBIx::Class::Timestamps 'now';
@@ -276,7 +276,7 @@ sub archive ($self) {
     return undef unless my $normal_result_dir = $self->archivable_result_dir;
 
     my $archived_result_dir = $self->add_result_dir_prefix($self->remove_result_dir_prefix($normal_result_dir), 1);
-    if (!dircopy($normal_result_dir, $archived_result_dir)) {
+    if (!File::Copy::Recursive::dircopy($normal_result_dir, $archived_result_dir)) {
         my $error = $!;
         File::Path::rmtree($archived_result_dir);    # avoid leftovers
         die "Unable to copy '$normal_result_dir' to '$archived_result_dir': $error";
--- a/lib/OpenQA/Schema/Result/Jobs.pm
+++ b/lib/OpenQA/Schema/Result/Jobs.pm
@@ -37,6 +37,7 @@ use OpenQA::Jobs::Constants;
 use OpenQA::JobDependencies::Constants;
 use OpenQA::ScreenshotDeletion;
 use File::Basename qw(basename dirname);
+use File::Copy::Recursive 'dircopy';
 use File::Spec::Functions 'catfile';
 use File::Path ();
 use DBIx::Class::Timestamps 'now';
@@ -264,6 +265,26 @@ sub delete {
     return $ret;
 }
 
+sub archive ($self) {
+    my $normal_result_dir = $self->result_dir;
+    return undef
+      if (!$normal_result_dir || !-d $normal_result_dir)
+      || $self->archived
+      || OpenQA::Jobs::Constants::meta_state($self->state) ne OpenQA::Jobs::Constants::FINAL;
+
+    my $archived_result_dir = $self->add_result_dir_prefix($self->remove_result_dir_prefix($normal_result_dir), 1);
+    if (!dircopy($normal_result_dir, $archived_result_dir)) {
+        my $error = $!;
+        File::Path::rmtree($archived_result_dir);    # avoid leftovers
+        die "Unable to copy '$normal_result_dir' to '$archived_result_dir': $error";
+    }
+
+    $self->update({archived => 1});
+    $self->discard_changes;
+    File::Path::remove_tree($normal_result_dir);
+    return $archived_result_dir;
+}
+
 sub name {
     my ($self) = @_;
     return $self->{_name} if $self->{_name};
@@ -434,14 +455,12 @@ sub settings_hash {
     return $settings;
 }
 
-sub add_result_dir_prefix {
-    my ($self, $rd) = @_;
-    return $rd ? catfile($self->num_prefix_dir, $rd) : undef;
+sub add_result_dir_prefix ($self, $result_dir, $archived = undef) {
+    return $result_dir ? catfile($self->num_prefix_dir($archived), $result_dir) : undef;
 }
 
-sub remove_result_dir_prefix {
-    my ($self, $rd) = @_;
-    return $rd ? basename($rd) : undef;
+sub remove_result_dir_prefix ($self, $result_dir) {
+    return $result_dir ? basename($result_dir) : undef;
 }
 
 sub set_prio {
@@ -1240,10 +1259,9 @@ END_SQL
     return [map { $_->[0] } @{$sth->fetchall_arrayref // []}];
 }
 
-sub num_prefix_dir {
-    my ($self)    = @_;
+sub num_prefix_dir ($self, $archived = undef) {
     my $numprefix = sprintf "%05d", $self->id / 1000;
-    return catfile(resultdir($self->archived), $numprefix);
+    return catfile(resultdir($archived // $self->archived), $numprefix);
 }
 
 sub create_result_dir {
--- a/lib/OpenQA/Schema/Result/Jobs.pm
+++ b/lib/OpenQA/Schema/Result/Jobs.pm
@@ -265,12 +265,15 @@ sub delete {
     return $ret;
 }
 
-sub archive ($self) {
-    my $normal_result_dir = $self->result_dir;
+sub archivable_result_dir ($self) {
     return undef
-      if (!$normal_result_dir || !-d $normal_result_dir)
-      || $self->archived
-      || OpenQA::Jobs::Constants::meta_state($self->state) ne OpenQA::Jobs::Constants::FINAL;
+      if $self->archived || OpenQA::Jobs::Constants::meta_state($self->state) ne OpenQA::Jobs::Constants::FINAL;
+    my $result_dir = $self->result_dir;
+    return $result_dir && -d $result_dir ? $result_dir : undef;
+}
+
+sub archive ($self) {
+    return undef unless my $normal_result_dir = $self->archivable_result_dir;
 
     my $archived_result_dir = $self->add_result_dir_prefix($self->remove_result_dir_prefix($normal_result_dir), 1);
     if (!dircopy($normal_result_dir, $archived_result_dir)) {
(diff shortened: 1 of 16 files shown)


dbicdh/PostgreSQL/upgrade/92-93/001-auto.sql
```diff
@@ -0,0 +1,12 @@
+-- Convert schema '/hdd/openqa-devel/repos/openQA/script/../dbicdh/_source/deploy/92/001-auto.yml' to '/hdd/openqa-devel/repos/openQA/script/../dbicdh/_source/deploy/93/001-auto.yml':;
+
+;
+BEGIN;
+
+;
+ALTER TABLE jobs ADD COLUMN is_archived boolean DEFAULT '0' NOT NULL;
```
okurz, line 7: "is_archived" sounds more like a method. I think in other cases of boolean columns we did not add the "is_" prefix.

lib/OpenQA/Schema/Result/Jobs.pm
```diff
@@ -2061,7 +2079,7 @@ sub done {
         # limit length of the reason
         # note: The reason can be anything the worker picked up as useful information so better cut it at a
         #       reasonable, human-readable length. This also avoids growing the database too big.
-        $reason = substr($reason, 0, 300) . decode('UTF-8', '…') if defined $reason && length $reason > 300;
+        $reason = substr($reason, 0, 300) . '…' if defined $reason && length $reason > 300;
```
okurz, line 2082: interesting. Why this change?

okurz (commented):
looks nice. Have you found a good place where to trigger the archiving? E.g. following our idea in the ticket: Should we consider "archived" == "important" after some grace time when an important job is moved to a potentially slower archive? So at the time when we trigger the cleanup for non-important jobs we also look at the important jobs and make sure that they are archived or queued for archiving (if that's a more costly process)

dbicdh/PostgreSQL/upgrade/92-93/001-auto.sql
Martchus: DBIx maps it to a method so I thought it would be a good fit. We already do the same for the boolean columns within the users table (`is_operator`, `is_admin`).

lib/OpenQA/Schema/Result/Jobs.pm
Martchus: see https://github.com/os-autoinst/openQA/pull/3859

dbicdh/PostgreSQL/upgrade/92-93/001-auto.sql
Martchus: To be fair, there are also other boolean columns *not* using that prefix. I don't really care but would postpone polishing like this until we decide that we actually want to merge the spike solution.

Martchus:
No, triggering the archiving is not included yet. However, the idea from the ticket / our previous discussion would make sense. Of course I'd make it a configuration option (and before enabling it one needed to ensure that the archive directory/mountpoint actually exist). And I would likely make it an extra Minion job (triggered by the the cleanup Minion job) because the cleanup job is already long-running so it would make sense to split it up.

There's also some polishing missing but considering it is a spike solution I'd like to do this only at the very end (when we decide that we actually like to merge it):

* Rename the table columns to remove `is_` prefix if that's generally preferred (amending the commits I've created so far)
* Indication within the UI
* Documentation
* Test coverage of error case

Martchus:
Adding a Minion task for this actually raised some questions, see `lib/OpenQA/Task/Job/ArchiveResults.pm`. The last commit is incomplete (only covers the config option so far).

dbicdh/PostgreSQL/upgrade/92-93/001-auto.sql
okurz: sure, not important right now. Note for later :)

lib/OpenQA/Task/Job/ArchiveResults.pm
```diff
@@ -0,0 +1,43 @@
...
+sub register ($self, $app, @args) {
+    $app->minion->add_task(archive_job_results => \&_archive_results);
+}
+
+sub _archive_results ($minion_job, @args) {
+    my ($openqa_job_id) = @args;
+    my $app = $minion_job->app;
+    return $minion_job->fail('No job ID specified.') unless defined $openqa_job_id;
+
+    # FIXME: Should we ensure that the job results are finalized?
```
okurz, line 28: how about we simply ignore non-finalized job results (can we detect that?) as they should be found as archiving candidates the next time

lib/OpenQA/Task/Job/ArchiveResults.pm
```diff
@@ -0,0 +1,43 @@
...
+    # FIXME: Should we take care of possible interference with cleanup task?
```
okurz, line 29: do you mean results or screenshots cleanup? Because when triggering synchronously with results cleanup then that at least should not interfer. And screenshots are symlinks which should still be valid after archiving a job unless any screenshot cleanup happens in parallel to the directory copy, right? In that case, how about a minion lock on each corresponding job?

lib/OpenQA/Task/Job/ArchiveResults.pm
```diff
@@ -0,0 +1,43 @@
...
+    # FIXME: This breaks jobs currently loading the results tab. I suppose we have to live
+    #        with it.
```
okurz, line 31: good point. I wonder how that would break and if we can intervene within the results tab loading to force-refresh loading that page?

lib/OpenQA/Task/Job/ArchiveResults.pm
Martchus: > can we detect that?

That's a good question. I think we'd have to check whether there are any `.txt` files within the results dir.

> as they should be found as archiving candidates the next time

If finalizing a job fails for some reason it is not automatically re-attempted. So we should at least try to re-attempt the finalizing.

lib/OpenQA/Task/Job/ArchiveResults.pm
Martchus: > do you mean results or screenshots cleanup? Because when triggering synchronously with results cleanup then that at least should not interfer.

I mean the result cleanup. I suppose it wouldn't interfere in practice but I can think of some contrived examples, e.g. a job is no longer considered important during the cleanup and then would be deleted after all (while the archiving is ongoing).

> And screenshots are symlinks which should still be valid after archiving a job unless any screenshot cleanup happens in parallel to the directory copy, right?

Yes, screenshots are not touched by this. And the symlinks remain valid as long as they use absolute paths which they luckily do. If screenshot cleanup happens in the background that shouldn't be a problem because it wouldn't delete screenshots used by existing jobs (regardless of their archiving state).

lib/OpenQA/Task/Job/ArchiveResults.pm
Martchus: I also wonder how that would break. Only HTTP requests which are pending when deleting the old results directory and had already been started before the job was marked as archived would cause problems.

We have controller code reading result files and some files are served directly by Apache. Likely the deletion of the original files during the archiving would fail if the result file if there are still file handles open for reading at this point. So we would likely see failed Minion jobs and leftover results. We could make it more resilient against that problem by trying to delete the results again after a short delay.

If it was actually possible to delete the files while there are still file handles open for reading I suppose the readers would run into I/O errors and at least some parts of the details tab wouldn't be shown correctly.

Martchus:
There's an even bigger problem. So far I've only tested this using Mojolicious HTTP server directly. However, when using the Apache reverse proxy setup loading images fails because Apache would still expect them within the usual results directory.

lib/OpenQA/Task/Job/ArchiveResults.pm
```diff
@@ -0,0 +1,43 @@
...
+
+    return $minion_job->retry({delay => 60})
+      unless my $guard1 = $app->minion->guard('archiving_results', 86400, {limit => 5});
+    return $minion_job->retry({delay => 30})
+      unless my $guard2 = $app->minion->guard("finalize_job_results_for_$openqa_job_id", 86400);
```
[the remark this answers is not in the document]
okurz: also see https://github.com/os-autoinst/openQA/pull/3861

okurz (changes requested):

lib/OpenQA/Task/Job/ArchiveResults.pm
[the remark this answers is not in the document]
Martchus: I'd prefer to do this kind of refactoring separately.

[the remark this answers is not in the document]
okurz: yes, this is why I created my PR. You can ignore the suggestions for here now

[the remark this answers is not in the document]
Martchus: If the refactoring PR gets merged after this i can still apply it here as a separate commit.

okurz:
by the way, https://github.com/os-autoinst/openQA/pull/3862 is ready to reviewed

Martchus:
Ticket for finalizing the PR: https://progress.opensuse.org/issues/91782

dbicdh/PostgreSQL/upgrade/92-93/001-auto.sql
kraih: I'd also just call this column `archived`.

okurz:
Looks great. Can you split out the database migration into a separate PR to merge that earlier?

Martchus:
I've been creating https://github.com/os-autoinst/openQA/pull/3897 but I still need to check the FIXMEs.

lib/OpenQA/Schema/Result/Jobs.pm
```diff
@@ -260,6 +265,30 @@ sub delete {
...
+sub archivable_result_dir ($self) {
+    return undef
+      if $self->archived || OpenQA::Jobs::Constants::meta_state($self->state) ne OpenQA::Jobs::Constants::FINAL;
+    my $result_dir = $self->result_dir;
+    return $result_dir && -d $result_dir ? $result_dir : undef;
+}
+
+sub archive ($self) {
+    my $normal_result_dir = $self->archivable_result_dir;
+    return undef unless $normal_result_dir;
```
okurz, line 277:
```suggestion
    return undef unless my $normal_result_dir = $self->archivable_result_dir;
```

if you like :) That's what we use already in other places

lib/OpenQA/Task/Job/ArchiveResults.pm
```diff
@@ -0,0 +1,43 @@
...
+    my $app = $minion_job->app;
+    return $minion_job->fail('No job ID specified.') unless defined $openqa_job_id;
+
+    # FIXME: Should we ensure that the job results are finalized?
+    # FIXME: Should we take care of possible interference with cleanup task?
+    # FIXME: This breaks jobs currently loading the results tab. I suppose we have to live
+    #        with it.
+
+    return $minion_job->retry({delay => 60})
+      unless my $guard1 = $app->minion->guard('archiving_results', 86400, {limit => 5});
```
okurz, line 34:
```suggestion
      unless my $guard1 = $app->minion->guard('archiving_results', ONE_DAY, {limit => 5});
```

lib/OpenQA/Task/Job/ArchiveResults.pm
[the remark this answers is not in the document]
okurz: you can use `ONE_DAY` here now as well

lib/OpenQA/Task/Job/Limit.pm
```diff
@@ -62,9 +62,19 @@ sub _limit {
     my $schema = $app->schema;
     $schema->resultset('JobGroups')->new({})->limit_results_and_logs;
 
-    my $groups = $schema->resultset('JobGroups');
+    my $groups  = $schema->resultset('JobGroups');
+    my $gru     = $app->gru;
+    my %options = (priority => 0, ttl => 172800);
```
okurz, line 67:
```suggestion
    my %options = (priority => 0, ttl => 2 * ONE_DAY);
```

okurz (changes requested):

Martchus:
See the commit message of the last commit for some caveats.

okurz (approved):
Approved. Please let mergify do some work, i.e. don't merge manually to give a bit more time to reviewers as well as to ensure all required checks are actually passed :)

Amrysliu (approved):
LGTM

AdamWill:
should we add the archive directory to the package spec?

Martchus:
Likely one would create a bind mount there or configure a different path because the idea is to make use of a different storage. So this would likely be not that useful considering one needs to do some manual setup anyways.

AdamWill:
hmm, yeah, that makes sense.

PR #3859, opened 2021-04-22
Enable signatures in …::Result::Jobs and fix issues

Martchus:
* Fix declaration of needle_dir
* Fix string literal '…' as this apparently also pulls in changed UTF-8 handling

---

When enabling signatures in `…::Result::Jobs` in https://github.com/os-autoinst/openQA/pull/3858 I've noticed that this is not completely straight forward as these two issues need to be fixed. So I've created a separate PR to have these fixes independently of https://github.com/os-autoinst/openQA/pull/3858.

### DIFF
--- a/lib/OpenQA/Schema/Result/Jobs.pm
+++ b/lib/OpenQA/Schema/Result/Jobs.pm
@@ -18,9 +18,9 @@ package OpenQA::Schema::Result::Jobs;
 use strict;
 use warnings;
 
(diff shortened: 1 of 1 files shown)


perlpunk (approved):

kalikiana (approved):

Martchus:
This is only waiting for the WIP check to report back. It is maybe stuck so I'm force-merging.

PR #3860, opened 2021-04-23
Add back to top button to layout

kraih:
test

The button looks a little different than on the QAM Dashboard because openQA doesn't use Bootstrap defaults. But it's consistent with the look of other buttons.

Progress: https://progress.opensuse.org/issues/91601

### DIFF
--- a/assets/javascripts/back_to_top.js
+++ b/assets/javascripts/back_to_top.js
@@ -1,0 +1,18 @@
+function backToTop() {
+    $(document).ready(function() {
+        $(window).scroll(function() {
+            // Increase the value to not show the button on shorter pages
+            if ($(this).scrollTop() > 50) {
+                $('#back-to-top').fadeIn();
+            } else {
+                $('#back-to-top').fadeOut();
+            }
+        });
+        $('#back-to-top').click(function() {
+            $('#back-to-top').tooltip('hide');
+            $('body, html').animate({ scrollTop: 0 }, 800);
+            return false;
+        });
+        $('#back-to-top').tooltip('show');
+    });
+}
--- a/assets/javascripts/back_to_top.js
+++ b/assets/javascripts/back_to_top.js
@@ -13,6 +13,5 @@ function backToTop() {
             $('body, html').animate({ scrollTop: 0 }, 800);
             return false;
         });
-        $('#back-to-top').tooltip('show');
     });
 }
--- a/assets/stylesheets/back_to_top.css
+++ b/assets/stylesheets/back_to_top.css
@@ -1,10 +1,13 @@
 .back-to-top {
     bottom: 20px;
+    border-radius: 50rem;
     cursor: pointer;
     display: none;
     position: fixed;
     right: 20px;
     z-index: 1000;
+    width: 2.2rem;
+    height: 2.2rem;
 }
 .back-to-top i {
     padding-left: 3pt !important;
--- a/assets/stylesheets/back_to_top.css
+++ b/assets/stylesheets/back_to_top.css
@@ -1,0 +1,11 @@
+.back-to-top {
+    bottom: 20px;
+    cursor: pointer;
+    display: none;
+    position: fixed;
+    right: 20px;
+    z-index: 1000;
+}
+.back-to-top i {
+    padding-left: 3pt !important;
+}
--- a/templates/webapi/layouts/bootstrap.html.ep
+++ b/templates/webapi/layouts/bootstrap.html.ep
@@ -29,6 +29,7 @@
           %= content_for 'head_javascript'
           $(function() {
             setupForAll();
+            backToTop();
             %= content_for 'ready_function'
             % if (current_user) {
               newFeature(<%= current_user->feature_version %>);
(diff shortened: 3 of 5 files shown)


templates/webapi/layouts/bootstrap.html.ep
```diff
@@ -29,6 +29,7 @@
           %= content_for 'head_javascript'
           $(function() {
             setupForAll();
+            backToTop();
```
Martchus, line 32: Where does this function come from? I've tested it locally and got:

```
jquery-3.3.1.js:3827 Uncaught ReferenceError: backToTop is not defined
    at HTMLDocument.<anonymous> ((index):64)
    at mightThrow (jquery-3.3.1.js:3534)
    at process (jquery-3.3.1.js:3602)
```

Martchus (changes requested):

templates/webapi/layouts/bootstrap.html.ep
kraih: Ah, yea, my latest force push was completely busted. No idea what happened.

assets/javascripts/back_to_top.js
```diff
@@ -0,0 +1,17 @@
...
+        $(window).scroll(function() {
+            if ($(this).scrollTop() > 50) {
+                $('#back-to-top').fadeIn();
+            } else {
+                $('#back-to-top').fadeOut();
+            }
+        });
+        $('#back-to-top').click(function() {
+            $('#back-to-top').tooltip('hide');
+            $('body, html').animate({ scrollTop: 0 }, 800);
```
Martchus, line 12: 800 feels quite slow (but that's of course a matter of preference).

assets/stylesheets/back_to_top.css
```diff
@@ -0,0 +1,8 @@
+.back-to-top {
+    bottom: 20px;
+    cursor: pointer;
+    display: none;
+    position: fixed;
+    right: 20px;
+    z-index: 1000;
+}
```
Martchus, line 8: I'd also add `.back-to-top i { padding-right: initial; }`. Otherwise the icon isn't centered correctly within the button.

assets/stylesheets/back_to_top.css
```diff
@@ -0,0 +1,8 @@
+.back-to-top {
+    bottom: 20px;
+    cursor: pointer;
+    display: none;
+    position: fixed;
+    right: 20px;
+    z-index: 1000;
```
Martchus, line 7: Maybe add

```
    border-radius: 50rem;
    width: 2.2rem;
    height: 2.2rem;
```

for consistency with the icon-only buttons we're using in the comment form.

assets/javascripts/back_to_top.js
okurz: if that is the same value than on cavil then it's good :)

kraih: Yes, i did not change any of the values from Cavil/QAMDashboard.

assets/stylesheets/back_to_top.css
kraih: Doesn't look like it changes anything.

assets/javascripts/back_to_top.js
Martchus: @okurz "cavil" means making an unnecessary objection so for the sake of irony I say your statement doesn't count.

assets/stylesheets/back_to_top.css
kraih: For anyone wondering, that makes it a circle. I'll wait for more comments/votes on that.

assets/stylesheets/back_to_top.css
kraih: With Bootstrap defaults the icon is perfectly centered, must be one of the custom settings openQA uses.

assets/javascripts/back_to_top.js
```diff
@@ -0,0 +1,17 @@
...
+            return false;
+        });
+        $('#back-to-top').tooltip('show');
+    });
+}
```
kraih, line 17: It bugs me that we forbid newlines at the end of JS files. 😬

kraih:
test-1

Ok, the button icon is now properly in the center and looks good.

assets/stylesheets/back_to_top.css
kraih: It's not properly in the center.

kraih:
test-2
And this is what it would look like with a circular button.

Martchus (approved):
Looks good. I'd vote for the circle would also approve without :-)

kraih:
I kinda like the circle too, so i've added a commit for it.

Martchus:
The fullstack test is failing with `Uncaught Error: Please use show on visible elements`. This could actually be a real error because the new button might indeed hide another element.

kraih:
That would be unfortunate. And i have no clue how to solve that i'm afraid.

kraih:
That would be unfortunate. And i have no clue how to solve that i'm afraid.

kraih:
But that's not actually what "visible" means in the DOM. It's just a property of the element and not affected by overlapping elements blocking the view for the user.

Martchus:
As far as I know Selenium actually also cares about the element being *actually* visible, e.g. it wouldn't allow to click on an element while another hides it (and would therefore in reality receive the click).

kraih:
Turns out the problem was not selenium at all, the error was just in the console, and came from the Bootstrap tooltip code. So i just disabled the tooltip. 😁

PR #3861, opened 2021-04-24
Use Time::Seconds instead of magic numbers

okurz:


Martchus (approved):

Martchus (changes requested):
Considering the test failures there must be something wrong. Maybe it would have helped to split the introduction of signatures and use of `Time::Seconds` into separate commits. Note that I'd only convert to signatures "along the way" when touching the argument list anyways. Otherwise it would be much cleaner to create a dedicated commit for the conversion. By the way, enabling signatures might *not* just work, see https://github.com/os-autoinst/openQA/pull/3859. Maybe you're hitting a similar problem.

okurz:
yes, you are right and I was always open for that. I first opted for not doing separate commits because within a "gitpod" session `git add -p` did not allow to split the patches on single lines. I will amend the commits by using `git gui` to the rescue and also split to introduce Time::Seconds independantly of signatures

okurz:
Let me switch to a proper branch from my fork instead: https://github.com/os-autoinst/openQA/pull/3862

PR #3863, opened 2021-04-26
Do symlink assets as default

Amrysliu:
In `vars.json`, our assets are recorded as absolute pathes. When users want
to run a case by using `vars.json`, they need to prepare the directory
for assets.

In this pull request, we do symlink assets to working directory as default,
and record the assets' name in `vars.json`.

Users can disable this feature by setting `ABSOLUTE_TEST_CONFIG_PATHS=1`

Related: https://progress.opensuse.org/issues/81703

Blocked by: https://github.com/os-autoinst/os-autoinst/pull/1658


okurz (approved):
applying "not-ready" to wait for the os-autoinst change first.

lib/OpenQA/Worker/Engines/isotovideo.pm
```diff
@@ -14,6 +14,7 @@
 # with this program; if not, see <http://www.gnu.org/licenses/>.
 
 package OpenQA::Worker::Engines::isotovideo;
+use Mojo::Base 'Mojo::EventEmitter', -signatures;
 
 use strict;
 use warnings;
```
okurz, line 20: try

```suggestion
use Mojo::Base 'Mojo::EventEmitter', -base, -signatures;
```

see https://docs.mojolicious.org/Mojo/Base#DESCRIPTION

Amrysliu: Thanks. I removed this line. I agree with your words `not mix features, bugfixes and signature updates`

okurz:
os-autoinst change merged. I suggest you update the os-autoinst hash to use in the full stack tests to use that btw version or wait until the automatic update does it and then rebase this PR on top

Martchus (approved):

PR #3865, opened 2021-04-27
Only consider bugrefs/labels for black certificate icon

Martchus:
* Do *not* count any kind of comment anymore when computing black
  certificate icon; only count bugrefs/labels in consistency with TODO
  parameter of test results overview page
* Use the same code for checking bugrefs/labels for certificate, test
  results overview page and 'Next & previous results' tab
* See https://progress.opensuse.org/issues/91658

### DIFF
--- a/lib/OpenQA/BuildResults.pm
+++ b/lib/OpenQA/BuildResults.pm
@@ -59,8 +59,9 @@ sub count_job {
             return;
         }
         if (grep { $job->result eq $_ } OpenQA::Jobs::Constants::NOT_OK_RESULTS) {
+            my $comment_data = $labels->{$job->id};
             $jr->{failed}++;
-            $jr->{labeled}++ if $labels->{$job->id};
+            $jr->{labeled}++ if $comment_data && ($comment_data->{bugs} || $comment_data->{label});
             return;
         }
         # note: Incompletes and timeouts are accounted to both categories - failed and skipped.
@@ -214,31 +215,20 @@ sub compute_build_results {
         }
 
         my %seen;
-        my @ids = map { $_->id } $jobs->all;
-        # prefetch comments to count. Any comment is considered a label here
-        # so a build is considered as 'reviewed' if all failures have at least
-        # a comment. This could be improved to distinguish between
-        # "only-labels", "mixed" and such
-        my $c = $group->result_source->schema->resultset('Comments')->search({job_id => {in => \@ids}});
-        my %labels;
-        while (my $comment = $c->next) {
-            $labels{$comment->job_id}++;
-        }
-        $jobs->reset;
-
-        while (my $job = $jobs->next) {
+        my $comment_data = $group->result_source->schema->resultset('Comments')->comment_data_for_jobs($jobs);
+        for my $job ($jobs->all) {
             $jr{distris}->{$job->DISTRI} = 1;
             my $key = $job->TEST . "-" . $job->ARCH . "-" . $job->FLAVOR . "-" . $job->MACHINE;
             next if $seen{$key}++;
 
             $jr{oldest} = $job->t_created if $job->t_created < $jr{oldest};
-            count_job($job, \%jr, \%labels);
+            count_job($job, \%jr, $comment_data);
             if ($jr{children}) {
                 my $child = $jr{children}->{$job->group_id};
                 $child->{distris}->{$job->DISTRI} = 1;
                 $child->{version} //= $job->VERSION;
                 $child->{build}   //= $job->BUILD;
-                count_job($job, $child, \%labels);
+                count_job($job, $child, $comment_data);
                 add_review_badge($child);
             }
         }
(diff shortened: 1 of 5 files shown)


lib/OpenQA/BuildResults.pm
```diff
@@ -59,8 +59,9 @@ sub count_job {
             return;
         }
         if (grep { $job->result eq $_ } OpenQA::Jobs::Constants::NOT_OK_RESULTS) {
+            my $comment_data = $labels->{$job->id};
             $jr->{failed}++;
-            $jr->{labeled}++ if $labels->{$job->id};
+            $jr->{labeled}++ if $comment_data && ($comment_data->{bugs} || $comment_data->{label});
```
okurz, line 64: even if it may look simple I think we should not have a logic like "bugs || label" in here. Maybe we need to define how we call "bugs || label" but anyway please extract in a function and use it here ...

lib/OpenQA/BuildResults.pm
```diff
@@ -214,31 +215,20 @@ sub compute_build_results {
...
-        # "only-labels", "mixed" and such
-        my $c = $group->result_source->schema->resultset('Comments')->search({job_id => {in => \@ids}});
-        my %labels;
-        while (my $comment = $c->next) {
-            $labels{$comment->job_id}++;
-        }
-        $jobs->reset;
-
-        while (my $job = $jobs->next) {
+        my $commant_data = $group->result_source->schema->resultset('Comments')->comment_data_for_jobs($jobs);
```
okurz, line 218: s/commant_data/comment_data/ ?

lib/OpenQA/Schema/Result/Jobs.pm
```diff
@@ -2236,15 +2236,13 @@ sub _overview_result_done {
...
-          || ($self->result eq OpenQA::Jobs::Constants::SOFTFAILED
-            && ($job_labels->{$jobid}{label} || !$self->has_failed_modules));
-    }
+    # skip all jobs NOT needed to be labeled for the black certificate icon to show up
+    my $comment_data = $job_labels->{$jobid};
+    return undef
+      if $todo
+      && ( ($overall eq PASSED)
+        || ($overall eq SOFTFAILED && !$self->has_failed_modules)
+        || ($comment_data->{bugs} || $comment_data->{label}));
```
okurz, line 2245: ... and here

Martchus:
It is slightly slower as my local tests show. With OSD data from yesterday the dashboard takes most of the time at least ~ 100 ms longer to load. I guess it is ok. At some point it might make sense to store this data directly in the database instead of parsing the comments over and over again.

okurz:
Could you please look into the code coverage? Not sure I read the codecov parts right but maybe this is a good opportunity to try to reach for 100% code coverage on all changed files so that our future review gets easier.

okurz:
@DimStar77 @lkocman @ggardet @michelmno do you see problems with the change? It should only impact the black certificate icon on the group overview page, not anything else. So I assume TTM is also not behaving any different

lib/OpenQA/Schema/ResultSet/Comments.pm
```diff
@@ -41,4 +41,43 @@ sub referenced_bugs {
...
+        elsif (my $label = $comment->label) {
+            $res->{label}    = $label;
+            $res->{reviewed} = 1;
+        }
+        else {
+            $res->{comments}++;
+        }
+        # note: Previous labels are overwritten here so only the most recent label is returned.
+    }
+    return \%res;
```
Amrysliu, line 80: maybe a stupid question,  I don't find the place which sets  values for `%res` in this function.

okurz: For example line 76.

$foo->{bar} sets the value for the key "bar" in a hash %foo

Amrysliu: I thoguht `$res->{comments}++`  was used to set the value for the `$res` in line 61 `my ($bugrefs, $res) = ($comment->bugrefs, $res{$comment->job_id} //= {});`

okurz: Ok, sorry, I think I did not look or explain well enough You took a much better look than I did :)

So

```perl
my ($bugrefs, $res) = ($comment->bugrefs, $res{$comment->job_id} //= {});
```

adds a reference `$res` to `$res{$comment->job_id}` and AFAICS at the same time it initializes `$res{$comment->job_id}` to an empty hash if it has no value yet.

Over the reference `$res` pointing to individual values, the complete hash `%res` is updated in lines like `$res->{comments}++`.

Good enough? :)

Amrysliu: Thanks for your explanation. It's really good.

Martchus:
> Could you please look into the code coverage?

The entire diff is actually covered.

okurz (approved):

DimStar77:
> @DimStar77 @lkocman @ggardet @michelmno do you see problems with the change? It should only impact the black certificate icon on the group overview page, not anything else. So I assume TTM is also not behaving any different

No problem; to me this sounds about like: black label is only added, if there is nothing left on 'todo', which sounds anyway better than the current state (used to be like this until the investigation jobs/comments appeared; before the cog only landed after somebody labeled or at least added a comment to a test)

PR #3868, opened 2021-04-28
Let the worker retry uploading when updating module results failed

Martchus:
* Avoid incompletes with the reason "Failed modules: …"
    * Weaken error handling introduced in
      a34d89c4a2a78d874c84596257ac28f27f20d892 so the worker would try the
      upload again
    * Affected tests actually have the failing test modules so it is likely
      that a retry would help
* See https://progress.opensuse.org/issues/91902

### DIFF
--- a/lib/OpenQA/Schema/Result/Jobs.pm
+++ b/lib/OpenQA/Schema/Result/Jobs.pm
@@ -1547,7 +1547,10 @@ sub update_status {
     }
     $ret->{known_images} = [sort keys %known_image];
     $ret->{known_files}  = [sort keys %known_files];
-    $ret->{error}        = 'Failed modules: ' . join ', ', @failed_modules if @failed_modules;
+    if (@failed_modules) {
+        $ret->{error}        = 'Failed modules: ' . join ', ', @failed_modules;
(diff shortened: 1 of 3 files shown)


Martchus:
I've been testing this to check the behavior with https://github.com/os-autoinst/openQA/pull/3869 and noticed that I can not use the return code 425 because this also makes the worker consider the web UI being busy and introduced a retry delay. So I'm now using the custom return code 499.

okurz:
do you have an idea what caused the updating errors in the first place even though module results do exist?

Martchus:
I haven't looked that much into it as the ticket isn't actually in ready. I mainly created this PR to demonstrate the suggestion I made in the ticket and as a quick workaround (although it might be the best/simplest solution anyways).

okurz:
@kraih WDYT about the return code?

kraih:
I think 499 already has a special meaning, and you can get it returned from some web servers like nginx.

Martchus:
I've read about it. Maybe I use just something else, e.g. 490 (or the teapot).

okurz (approved):

PR #3869, opened 2021-04-28
Consider jobs with no modules incomplete

Martchus:
* Consider jobs with no modules incomplete when the worker does not
  explicitly pass a result
* Use `no test modules scheduled/uploaded` as reason if there's no other
  reason
    * Compilation errors, needle initialization errors and other early
      errors (before the test module creation) still have a specific reason
* Tested manually by breaking controller code to simulate
  https://progress.opensuse.org/issues/90152

### DIFF
--- a/t/api/04-jobs.t
+++ b/t/api/04-jobs.t
@@ -1216,15 +1216,26 @@ subtest 'marking job as done' => sub {
         $t->post_ok('/api/v1/jobs/99961/set_done?result=incomplete&reason=test&worker_id=1');
         $t->status_is(400, 'set_done with worker_id rejected if job no longer assigned');
         $t->json_like('/error', qr/Refusing.*because.*re-scheduled/, 'error message returned');
+
         $schema->txn_begin;
         $jobs->find(99961)->update({assigned_worker_id => 1});
         $t->post_ok('/api/v1/jobs/99961/set_done?result=incomplete&reason=test&worker_id=42');
         $t->status_is(400, 'set_done with worker_id rejected if job assigned to different worker');
         $t->json_like('/error', qr/Refusing.*because.*assigned to worker 1/, 'error message returned');
-        $t->post_ok('/api/v1/jobs/99961/set_done?result=incomplete&reason=test&worker_id=1');
+        $t->post_ok('/api/v1/jobs/99961/set_done?result=failed&reason=test&worker_id=1');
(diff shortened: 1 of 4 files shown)


t/api/04-jobs.t
```diff
@@ -1216,15 +1216,26 @@ subtest 'marking job as done' => sub {
...
+        my $job = $jobs->find(99961);
+        is $job->clone_id, undef, 'job not cloned when reason does not match configured regex';
+        is $job->result, FAILED, 'result is failure (as passed via param)';
+        $schema->txn_rollback;
+
+        $schema->txn_begin;
+        $t->post_ok('/api/v1/jobs/99961/set_done');
+        $job = $jobs->find(99961);
+        is $job->result, INCOMPLETE, 'result is incomplete (no modules and no reason explicitely specified)';
+        is $job->reason, 'no test modules scheduled/uploaded', 'reason for incomplete set';
```
okurz, line 1236: considering that t/api/04-jobs.t takes rather long already and is one of the longer running tests I would appreciate if you can consider splitting it (or optimising in other ways):

```
[14:51:33] t/api/04-jobs.t .................... ok   142093 ms ( 0.37 usr  0.04 sys + 127.56 cusr  4.57 csys = 132.54 CPU)
```

Martchus: This is almost nothing. I don't think I can optimize the new code. Maybe there's something else within `t/api/04-jobs.t` I can optimize. Splitting it will increase the overall test execution time.

okurz: yes, of course, the new code is straightforward and not introducing any significant slowdown. Would be awesome if you can find something else to optimize in htere

Martchus: Looks like this is only slow when collecting coverage because collection of coverage data for forked processes is very inefficient. Unfortunately the code under test is forking during the screenshot upload.

okurz: hm, yeah, coverage collection in these subprocesses is costly. Would be better if we could disable coverage analysis in subprocesses unless for the test code that intends to cover the relevant processes. Well, I guess nevermind then?

okurz (approved):

t/api/04-jobs.t
Martchus: I've been creating https://github.com/os-autoinst/openQA/pull/3874 to speed up the test.

PR #3870, opened 2021-04-29
[WIP] API: Add job list by issue_id

ilausuch:
This new feature allows to the user via API get the jobs
that are related with an issue_id

https://progress.opensuse.org/issues/91542

### DIFF
--- a/lib/OpenQA/WebAPI/Controller/API/V1/Issue.pm
+++ b/lib/OpenQA/WebAPI/Controller/API/V1/Issue.pm
@@ -1,0 +1,67 @@
+# Copyright (C) 2015-2016 SUSE LLC
+#
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation; either version 2 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License along
+# with this program; if not, see <http://www.gnu.org/licenses/>.
+
+package OpenQA::WebAPI::Controller::API::V1::Issue;
+use Mojo::Base 'Mojolicious::Controller';
+
+# use DBIx::Class::ResultClass::HashRefInflator;
+use OpenQA::Utils;
+
+use constant JOB_QUERY_LIMIT => 10000;
+
+=pod
+
+=head1 NAME
+
+OpenQA::WebAPI::Controller::API::V1::Issue
+
+=head1 SYNOPSIS
+
+  use OpenQA::WebAPI::Controller::API::V1::Issue;
+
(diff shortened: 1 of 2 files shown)


lib/OpenQA/WebAPI/Controller/API/V1/Issue.pm
```diff
@@ -0,0 +1,67 @@
...
+
+OpenQA::WebAPI::Controller::API::V1::Issue
+
+=head1 SYNOPSIS
+
+  use OpenQA::WebAPI::Controller::API::V1::Issue;
+
+=head1 DESCRIPTION
+
+OpenQA API implementation for assets handling methods.
```
okurz, line 36: seems like a copy-paste error?

lib/OpenQA/WebAPI/Controller/API/V1/Issue.pm
```diff
@@ -0,0 +1,67 @@
...
+
+=head1 METHODS
+
+=back
+
+=cut
+
+sub jobs {
+    my $self   = shift;
```
okurz, line 45: for new classes I suggest to start with signatures directly:

```suggestion
use Mojo::Base 'Mojolicious::Controller' -base, -signatures;
...
sub jobs($self) {
```

lib/OpenQA/WebAPI/Controller/API/V1/Issue.pm
```diff
@@ -0,0 +1,67 @@
...
+    my $schema = $self->schema;
+
+    my $validation = $self->validation;
+    $validation->optional('limit')->num(0);
+
+    my $limit = $validation->param('limit') // JOB_QUERY_LIMIT;
+    return $self->render(json => {error => 'Limit exceeds maximum'}, status => 400) unless $limit <= JOB_QUERY_LIMIT;
+
+    my $rs = $schema->resultset("JobSettings")->search({
+        key => {like => '%ISSUE%'},
```
okurz, line 55: As this definition of "ISSUE" is something that SUSE QAM invented for their own internal use cases but there are other objects called "issue" within openQA itself this would be a bad idea.

okurz (changes requested):

lib/OpenQA/WebAPI/Controller/API/V1/Issue.pm
ilausuch: maybe could be a parametrized option, or search in all the settings

okurz: yes, that sounds good. Like "list jobs by setting" :)

ilausuch: In this case, I think the /v1/issues API entry is not the best option. 
I am thinking that could be even better to do a search by settings in jobs

lib/OpenQA/WebAPI/Controller/API/V1/Issue.pm
perlpunk: @okurz  missing a comma:
```
use Mojo::Base 'Mojolicious::Controller', -base, -signatures;
```

ilausuch:
I close this PR because is not longer necessary

PR #3874, opened 2021-04-29
Speed up execution of t/api/04-jobs.t with coverage analysis enabled

Martchus:
* Avoid forking which slows down the test execution speed massively
* It now runs within 27 seconds locally; previously the limit of 120
  seconds (includes scaling for run with coverage analysis) was exceeded


t/api/04-jobs.t
```diff
@@ -55,6 +57,13 @@ my $chunk_size = 10000000;
 
 OpenQA::Events->singleton->on('chunk_upload.end' => sub { Devel::Cover::report() if Devel::Cover->can('report'); });
 
+my $io_loop_mock = Test::MockModule->new('Mojo::IOLoop');
```
okurz, line 60: If you could add a comment what this is about I would be ok with that :)

okurz (approved):

okurz:
the coverage report tells me that we have no other test code that ensures coverage of the missed parts than the parts that you excluded from coverage analysis. So either we have specific tests covering these parts or we ditch the approach. WDYT?

Martchus:
Strange - this shouldn't have been the case. The code I'm mocking is not part of the code we consider for coverage analysis. The decrease of -69.92% within `t/api/04-jobs.t` seems also not very plausible. Unfortunately I can not access the coverage report on codecov see any details.

Martchus:
Removing `OpenQA::Events->singleton->on('chunk_upload.end' => sub { Devel::Cover::report() if Devel::Cover->can('report'); });` did the trick.

The fullstack test failed but it was unrelated:

```
RETRY=0 timeout -s SIGINT -k 5 -v $((9 * (0 + 1) ))m tools/retry prove -l --harness TAP::Harness::JUnit --timer t/full-stack.t
[12:18:35] t/full-stack.t .. 23/? timeout: sending signal INT to command ‘tools/retry’
```

PR #3876, opened 2021-04-30
openqa_bootstrap: Fix indentation

asdil12:


okurz:
Is there a static check we can have to ensure such style for the future? I don't know if maybe an option to shellcheck would help

kalikiana:
> Is there a static check we can have to ensure such style for the future? I don't know if maybe an option to shellcheck would help

I like the idea and filed [poo#92146](https://progress.opensuse.org/issues/92146).

okurz (approved):

PR #3877, opened 2021-04-30
Simplify handling of potential not-implemented states of jobs

okurz:
Helps with statement coverage in tests as well :) 

Encountered while looking at code statement coverage reports on codecov,
in particular
https://codecov.io/gh/os-autoinst/openQA/src/master/lib/OpenQA/BuildResults.pm

Related progress issue: https://progress.opensuse.org/issues/55364

### DIFF
--- a/lib/OpenQA/BuildResults.pm
+++ b/lib/OpenQA/BuildResults.pm
@@ -75,11 +75,9 @@ sub count_job {
         return;
     }
     my $state = $job->state;
-    if (grep { /$state/ } (OpenQA::Jobs::Constants::PENDING_STATES)) {
-        $jr->{unfinished}++;
-        return;
-    }
-    log_error("MISSING S:" . $job->state . " R:" . $job->result);
+    log_error("Encountered not-implemented state:" . $job->state . " result:" . $job->result)
+      unless grep { /$state/ } (OpenQA::Jobs::Constants::PENDING_STATES);
+    $jr->{unfinished}++;
     return;
 }
 


Martchus (changes requested):
In production this will result in a 500 error page for the user and we'd have to dig into the logs to find the exact error message. That's also not very nice. I'd rather put a warning somewhere on the web page if we consider this important enough or completely remove the code path. Or we'd simply add a category "other" where we would add the states/results to.

okurz:
Right, so if the die is handled that way by presenting 500 to the user then surely this PR in the current form is no improvement. I have a different preference though. An error in the logs is still better than in the webUI. The webUI is meant for users, the log for admins. That we don't look into logs (anymore) is our problem as admins. For that we also have https://github.com/os-autoinst/openqa-logwarn . But maybe an adjustment of your "other" proposal we can just count the unidentified as "unfinished" together. See my updated proposal.

Martchus (approved):
I guess counting these into unfinished makes sense. We handle already cancelled and done and it is unlikely that we introduce more "finished" states in the future.

kalikiana:
Can you please add a description with e.g. a ticket or discription of what you're trying to achieve?

okurz:
Is the second line "Helps with statement coverage in tests as well :)" in the commit message enough?

kalikiana:
> Is the second line "Helps with statement coverage in tests as well :)" in the commit message enough?

Not really. I'm still wondering what conversation or investigation this helps with "as well".

okurz:
@kalikiana updated the git commit message with more details and also a ticket reference and copied that into the PR description.

okurz:
created https://progress.opensuse.org/issues/92164 for the test failure which I consider unrelated. Will retrigger.

kalikiana:
11:17:17] t/01-test-utilities.t ..................................... 
    #   Failed test 'test would have failed'
    #   at t/01-test-utilities.t line 49.

    #   Failed test 'manual termination via stop_service does not trigger _fail_and_exit'
    #   at t/01-test-utilities.t line 63.
    #          got: '1'
    #     expected: '0'
    # Looks like you failed 2 tests of 5.

PR #3880, opened 2021-05-04
Add favicon for open.qa page

okurz:


Martchus (approved):

Martchus:
It has a white background (instead a transparent one like the favicon of openQA itself). I suppose a transparent background would look better in most cases.

okurz:
Really? I just copied over the one from master. And it looked correct to me on open.qa in chrome

Martchus:
It doesn't look correct on Firefox and Chromium. Of course you'll have to switch to another tab in Chromium because the default background color of an active tab button is white.

okurz:
Right. I could see that now. Well, I would not mind if anyone changes that but I don't plan to do that right now.

PR #3881, opened 2021-05-05
[WIP] Add 2 spaces to indent the logs of the download assets

ilausuch:
This commit helps users distinguish between log lines and the output
of asset download operations. This adds two spaces before every output
line

https://progress.opensuse.org/issues/91527

### DIFF
--- a/lib/OpenQA/Utils.pm
+++ b/lib/OpenQA/Utils.pm
@@ -903,4 +903,6 @@ sub fix_top_level_help { @ARGV = () if ($ARGV[0] // '') =~ qr/^(-h|(--)?help)$/ 
 
 sub looks_like_url_with_scheme { return !!Mojo::URL->new(shift)->scheme }
 
+
+sub indent_log_raw_output { @ARGV = () if ($ARGV[0] // '') =~ s/\r\n/,\s\s/gm;}
 1;
--- a/lib/OpenQA/Worker/Engines/isotovideo.pm
+++ b/lib/OpenQA/Worker/Engines/isotovideo.pm
@@ -20,7 +20,7 @@ use warnings;
 
 use OpenQA::Constants qw(WORKER_SR_DONE WORKER_EC_CACHE_FAILURE WORKER_EC_ASSET_FAILURE WORKER_SR_DIED);
 use OpenQA::Log qw(log_error log_info log_debug log_warning get_channel_handle);
-use OpenQA::Utils qw(asset_type_from_setting base_host locate_asset looks_like_url_with_scheme testcasedir productdir);
+use OpenQA::Utils qw(asset_type_from_setting base_host locate_asset looks_like_url_with_scheme testcasedir productdir indent_log_raw_output);
 use POSIX qw(:sys_wait_h strftime uname _exit);
 use Mojo::JSON 'encode_json';    # booleans
 use Cpanel::JSON::XS ();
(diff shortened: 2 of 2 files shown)


lib/OpenQA/Worker/Engines/isotovideo.pm
```diff
@@ -140,7 +140,10 @@ sub cache_assets {
             $status = $cache_client->status($asset_request);
         }
         my $msg = "Download of $asset_uri processed";
-        if (my $output = $status->output) { $msg .= ":\n$output" }
+        if (my $output = $status->output) {
```
okurz, line 143: I suggest to move your git commit explanation into the source code as it's not clear to source code users what this is for. An alternative can always be a function that by its name already tells what it tries to achieve

okurz (changes requested):
s/ident/indent/

Martchus:
@okurz was faster. I now changed the PR title but you still need to change the commit message.

lib/OpenQA/Worker/Engines/isotovideo.pm
ilausuch: Yes, I thought about the lib function with a good name to no need comments. Let's do that

ilausuch: How about now?

lib/OpenQA/Utils.pm
```diff
@@ -903,4 +903,6 @@ sub fix_top_level_help { @ARGV = () if ($ARGV[0] // '') =~ qr/^(-h|(--)?help)$/
 
 sub looks_like_url_with_scheme { return !!Mojo::URL->new(shift)->scheme }
 
+
+sub indent_log_raw_output { @ARGV = () if ($ARGV[0] // '') =~ s/\r\n/,\s\s/gm;}
```
perlpunk, line 907: `@ARGV`? that's the array with command line arguments to the script

okurz, line 907: I suggest to not add a function here when it is only used in a single location. OpenQA::Utils is not a dumping ground. Just move the function back into the class where used.

The name is fine, the argument is wrong :)

okurz (changes requested):

okurz:
oh, and as we discussed about "Extreme Programming" today morning I would like to reference http://www.extremeprogramming.org/rules/testfirst.html so obviously you did not test your changes because @ARGV is not what you need so I suggest to first try to come up with a test, only then add the implementation :)

lib/OpenQA/Utils.pm
perlpunk: Even with `@_`, this doesn't work at all.
i wonder if it even compiles for you.
`\s` on the right hand side of a substitution should generate a warning:
```
Unrecognized escape \s passed through
```
Trying to modify a constant string `... // ''` results in an error here:
```
Can't modify defined or (//) in substitution (s///)
```

ilausuch:
Sure

ilausuch:
@perlpunk I think the best option in this case is to include the indentation `=~ s/^/ /gm;` in https://github.com/os-autoinst/openQA/blob/c034e36b0b1e168e9c8b120728c102c7a1bfe717/lib/OpenQA/Log.pm#L86. In this way we'll solve this and others

What do you think?

ilausuch:
Continues at #3908

PR #3882, opened 2021-05-05
 Do not enqueue more Minion jobs while waiting in 27-plugin_obs_rsync_status_details.t

Martchus:
* Enqueuing even more `obs_rsync_update_builds_text` jobs only decreases
  the likeliness of processing all Minion jobs which need to run within the
  test in time
* See https://progress.opensuse.org/issues/89935

### DIFF
--- a/t/ui/27-plugin_obs_rsync_status_details.t
+++ b/t/ui/27-plugin_obs_rsync_status_details.t
@@ -122,13 +122,14 @@ sub _wait_for_change ($selector, $break_cb, $refresh_cb = undef) {
     my $text;
     my $limit = int OpenQA::Test::TimeLimit::scale_timeout(10);
     for my $i (0 .. $limit) {
-        note 'Pending Minion jobs: ' . $minion->jobs({states => [qw(inactive active)]})->total;
 
         # sometimes gru is not fast enough, so let's refresh the page and see if that helped
         if ($i > 0) {
             sleep 1;
+            my $pending_minion_jobs = $minion->jobs({states => [qw(inactive active)]});
+            note 'Pending Minion jobs: ' . $pending_minion_jobs->total;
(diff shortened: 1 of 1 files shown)


okurz:
https://github.com/os-autoinst/openQA/pull/3884 for t/ui/25-developer_mode.t (see https://progress.opensuse.org/issues/80374), retriggering here

t/ui/27-plugin_obs_rsync_status_details.t
```diff
@@ -36,13 +36,27 @@ use Mojo::IOLoop::ReadWriteProcess 'process';
 use Mojo::IOLoop::ReadWriteProcess::Session 'session';
 use Time::HiRes 'sleep';
 
-my $fixtures = '01-jobs.pl 03-users.pl';
-OpenQA::Test::Case->new->init_data(fixtures_glob => $fixtures);
+my $port          = Mojo::IOLoop::Server->generate_port;
+my $host          = "http://127.0.0.1:$port";
+my $url           = "http://127.0.0.1:$port/public/build/%%PROJECT/_result";
```
okurz, line 41:
```suggestion
my $url           = "$host/public/build/%%PROJECT/_result";
```

okurz (changes requested):
From the commit messages it's not entirely clear if your changes should increase stability. If that is the case then I would appreciate some statistics in a commit message, e.g. "before" and "after" comparison

Martchus:
That's because I'm not sure whether it will increase stability. However, it should speed up the test in case Minion jobs are executed slowly. Not sure how to include figures without having them. When running the test locally it always passes and within the CI it also passes most of the time.

okurz:
"most of the time" isn't easily quantifiable ;)

PR #3887, opened 2021-05-05
Makefile: set DEVEL_COVER_DB_FORMAT

perlpunk:
To reproduce:

    make coverage  TESTS=t/01-test-utilities.t

This didn't work for Oli and me on Leap 15.2

cover was failing with:

    Can't read /home/okurz/local/os-autoinst/openQA/cover_db/runs/1620219503.14849.03766/cover.14 with Sereal: Sereal: Error: Bad Sereal header: Not a valid Sereal document. at offset 1 of input at srl_decoder.c line 600 at /usr/lib/perl5/vendor_perl/5.26.1/x86_64-linux-thread-multi/Devel/Cover/DB/IO/Sereal.pm line 34, <$fh> chunk 1.

So apparently DEVEL_COVER_DB_FORMAT got lost somewhere.

Issue: https://progress.opensuse.org/issues/92179

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -207,6 +207,7 @@ endif
 
 .PHONY: coverage
 coverage:
+	export DEVEL_COVER_DB_FORMAT=JSON;\
 	COVERAGE=1 cover ${COVER_OPTS} -test
 
 COVER_REPORT_OPTS ?= -select_re '^(lib|script|t)/'


Martchus (approved):

Martchus:
Thanks to CMake I'm not very familiar with Makefiles but I assume that the modifications to the environment are not preserved when `make` leaves a directory (the `make[1]: Leaving directory '/home/okurz/local/os-autoinst/openQA'` line). If that's true it wouldn't be a surprise that the environment variable is lost.

okurz (approved):
I could verify that this works with `make coverage KEEP_DB=1 TESTS=t/01-test-utilities.t` that previously failed.

PR #3889, opened 2021-05-06
Do not allow to deprioritize/obsolete build when specifying TEST

Martchus:
When only scheduling a single scenario it is likely not intended to
deprioritize/obsolete which can be rather confusing. We've seen this
problem in production where a product has been scheduled with
`TEST=containers_sle_image_on_sle_host` and `_DEPRIORITIZEBUILD=1` which
deprioritized all previous jobs with the same DISTRI/VERSION/FLAVOR/ARCH.

### DIFF
--- a/lib/OpenQA/Schema/Result/ScheduledProducts.pm
+++ b/lib/OpenQA/Schema/Result/ScheduledProducts.pm
@@ -1,4 +1,4 @@
-# Copyright (C) 2019-2020 SUSE LLC
+# Copyright (C) 2019-2021 SUSE LLC
 #
 # This program is free software; you can redistribute it and/or modify
 # it under the terms of the GNU General Public License as published by
@@ -226,6 +226,13 @@ sub _schedule_iso {
     my $obsolete           = delete $args->{_OBSOLETE}                 // 0;
     my $onlysame           = delete $args->{_ONLY_OBSOLETE_SAME_BUILD} // 0;
     my $skip_chained_deps  = delete $args->{_SKIP_CHAINED_DEPS}        // 0;
+    my $force              = delete $args->{_FORCE_DEPRIORITIZEBUILD};
+    $force = delete $args->{_FORCE_OBSOLETE} || $force;
+    if (($deprioritize || $obsolete) && $args->{TEST} && !$force) {
+        return {error => 'One must not specify TEST and _DEPRIORITIZEBUILD=1/_OBSOLETE=1 at the same time as it is'
+              . 'likely not intended to deprioritize the whole build when scheduling a single scenario.'
+        };
+    }
 
     my $result = $self->_generate_jobs($args, \@notes, $skip_chained_deps);
     return {error => $result->{error_message}, error_code => $result->{error_code} // 400}
--- a/t/39-scheduled_products-table.t
+++ b/t/39-scheduled_products-table.t
@@ -1,5 +1,5 @@
 #!/usr/bin/env perl
-# Copyright (C) 2019-2020 SUSE LLC
+# Copyright (C) 2019-2021 SUSE LLC
 #
 # This program is free software; you can redistribute it and/or modify
 # it under the terms of the GNU General Public License as published by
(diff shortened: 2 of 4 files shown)


t/39-scheduled_products-table.t
```diff
@@ -50,11 +50,17 @@ my $scheduled_product;
...
     is_deeply(
         $scheduled_product->schedule_iso({REPO_0 => ''}),
         {error => 'Asset type and name must not be empty.'},
         'schedule_iso prevents adding assets with empty name',
     );
+    $schema->txn_rollback;
+
+    like $scheduled_product->schedule_iso({_DEPRIORITIZEBUILD => 1, TEST => 'foo'})->{error},
+      qr/One must not specify TEST and _DEPRIORITIZEBUILD.*/,
+      'schedule_iso prevents deprioritization/obsoleten when scheduling single scenario';
```
okurz, line 63: s/obsoleten/obsoletion/

lib/OpenQA/Schema/Result/ScheduledProducts.pm
```diff
@@ -226,6 +226,11 @@ sub _schedule_iso {
     my $obsolete           = delete $args->{_OBSOLETE}                 // 0;
     my $onlysame           = delete $args->{_ONLY_OBSOLETE_SAME_BUILD} // 0;
     my $skip_chained_deps  = delete $args->{_SKIP_CHAINED_DEPS}        // 0;
+    if (($deprioritize || $obsolete) && $args->{TEST}) {
+        return {error => 'One must not specify TEST and _DEPRIORITIZEBUILD=1/_OBSOLETE=1 at the same time as it is'
+              . 'likely not intended to deprioritize the whole build when scheduling a single scenario.'
```
okurz, line 231: If a complete build is scheduled and then one wants to schedule individual tests but prioritize them over the build I think test+deprio still makes sense, don't you think?

okurz (changes requested):

lib/OpenQA/Schema/Result/ScheduledProducts.pm
Martchus: I would leave that to the users to decide, see discussion in Rocket Chat.

okurz: sorry, I don't understand. How can users decide if you prevent that case with an error?

Martchus:
Just for the record: With a little bit more effort I could also add `TEST` to the query params for jobs to deprioritize so it would actually work as expected and only deprioritize relevant jobs. However, it would be equally intrusive in the sense that we change existing behavior (point raised in https://github.com/os-autoinst/openQA/pull/3889#discussion_r627256460).

coolo:
But I'd claim that this would be the expected behaviour - the current one can surely be considered a bug.

coolo:
Well, if someone wanted a single TEST to be scheduled and the rest of the build deprioritzed, I would say: get your own openqa installation.

okurz:
there is a big difference between "single job in one build of one product" vs. "the complete instance is affected".

Martchus:
@okurz I can follow @coolo and I don't know what you're getting at with your last statement.

> Well, if someone wanted a single TEST to be scheduled and the rest of the build deprioritzed, I would say: get your own openqa installation.

And if that's too much to ask for we could of course still add a force parameter. That's likely better than keeping a dangerous default-behavior.

lib/OpenQA/Schema/Result/ScheduledProducts.pm
```diff
@@ -226,6 +226,11 @@ sub _schedule_iso {
     my $obsolete           = delete $args->{_OBSOLETE}                 // 0;
     my $onlysame           = delete $args->{_ONLY_OBSOLETE_SAME_BUILD} // 0;
     my $skip_chained_deps  = delete $args->{_SKIP_CHAINED_DEPS}        // 0;
+    if (($deprioritize || $obsolete) && $args->{TEST}) {
```
okurz, line 229: how about

```suggestion
    if (($deprioritize || $obsolete) && $args->{TEST} && !$args->{_FORCE_DEPRIORITIZEBUILD} && !$args->{_FORCE_OBSOLETE}) {
```

okurz (approved):

okurz (approved):

lib/OpenQA/Schema/Result/ScheduledProducts.pm
Amrysliu: just one small question: users could use `_FORCE_OBSOLETE` or `_FORCE_DEPRIORITIZEBUILD` to obsolete or deprioritize when scheduling jobs, why we provide two different parameters? one should work, right?

Martchus: Yes, one would work. I was just following @okurz 's idea because I also thought it makes sense to have one force option matching each setting.

Amrysliu (approved):

PR #3890, opened 2021-05-06
CI: Fix 01-test-utilities.t

perlpunk:
It can happen that the warning in the subprocess happens before the
SIGCHLD handler was installed.

The sleep will fix this in most cases.

(To be really sure one would have to add some method in Utils.pm to check if
the handler has been installed.)

Also I changed combined_like to combined_from, which made it easier to debug
the actual output.

Issue: https://progress.opensuse.org/issues/92164

I ran this in a loop in CircleCI [50 times](https://app.circleci.com/pipelines/github/perlpunk/openQA/171/workflows/a48cf629-ecf2-4ec4-9a1c-f238f0b26968/jobs/751), it always passed.

### DIFF
--- a/t/01-test-utilities.t
+++ b/t/01-test-utilities.t
@@ -21,7 +21,7 @@ use lib "$FindBin::Bin/lib", "$FindBin::Bin/../external/os-autoinst-common/lib";
 use OpenQA::Test::TimeLimit '15';
 use OpenQA::Test::Utils 'stop_service';
 use IPC::Run 'start';
-use Test::Output 'combined_like';
+use Test::Output qw(combined_from);
 use Test::MockModule;
 
 my $signal_timeout = OpenQA::Test::TimeLimit::scale_timeout 7;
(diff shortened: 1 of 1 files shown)


Martchus (approved):

Martchus:
> (To be really sure one would have to add some method in Utils.pm to check if
the handler has been installed.)

And likely that wouldn't be that simple. I mean, what do we do if the check determines that the handler hasn't been installed but the child process has already terminated? This is still a situation we can't really handle keeping a positive test result.

We could register the handler before starting the child process (without knowing the PID and recording unknown PIDs) and when the child PID is known we make the handler aware of it but also handle the case when the handler was already called. That's of course rather complicated and maybe not worth it.

t/01-test-utilities.t
```diff
@@ -35,16 +35,18 @@ subtest 'warnings in sub processes are fatal test failures' => sub {
             isnt(shift, 0, 'exit code of test failure is non-zero');
             $test_would_have_failed = 1;
         });
-    combined_like {
+    my $out = combined_from {
         # start a sub process like the test helper do and simulate a Perl warning
         OpenQA::Test::Utils::_setup_sigchld_handler 'test-process-1', start sub {
+            sleep 1;                                                     # Give Utils.pm a chance to install $SIG{CHLD}
```
okurz, line 41: We now need an "uncoverable statement" in the comment as well

okurz:
Meh, those god-forsaken sleep calls. Does this slow down tests by second?

t/01-test-utilities.t
perlpunk: ah right. done

okurz (approved):

PR #3894, opened 2021-05-10
worker: Improve responsiveness while waiting for worker cache

Martchus:
See particular commit messages and https://progress.opensuse.org/issues/91764

---

It might make sense to checkout the commits individually.

### DIFF
--- a/docs/GettingStarted.asciidoc
+++ b/docs/GettingStarted.asciidoc
@@ -168,8 +168,8 @@ overall result from the following list.
   that workaround needles are in place, a softfailure has been recorded explicitly
   via `record_soft_failure` (from os-autoinst) or a job failure has been ignored
   explicitly via a <<UsersGuide.asciidoc#_show_bug_or_label_icon_on_overview_if_labeled_gh550,job label>>.
-* *timeout_exceeded* The job was aborted because `MAX_JOB_TIME` has been exceeded, see
-  <<WritingTests.asciidoc#_changing_timeout,Changing timeout>> for details.
+* *timeout_exceeded* The job was aborted because `MAX_JOB_TIME` or `MAX_SETUP_TIME`
+  has been exceeded, see <<WritingTests.asciidoc#_changing_timeout,Changing timeout>> for details.
 * *skipped* Dependencies failed so the job was not started.
 * *obsoleted* The job was superseded by scheduling a new product.
 * *parallel_failed*/*parallel_restarted* The job could not continue because a job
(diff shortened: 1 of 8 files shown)


Martchus:
This PR now addresses the two problems from https://progress.opensuse.org/issues/91764.

* Cancellation of a job is now also possible while waiting for asset downloads and syncing tests.
* The time it takes to download assets and sync tests is now counted into `MAX_JOB_TIME` and the job will be stopped if `MAX_JOB_TIME` is exceeded during that time.

The last point might be problematic because it means we would give jobs less time than before. If that sounds too dangerous I could increase the default `MAX_JOB_TIME` or add an additional variable.

okurz:
>     * The time it takes to download assets and sync tests is now counted into `MAX_JOB_TIME` and the job will be stopped if `MAX_JOB_TIME` is exceeded during that time.
> 
> The last point might be problematic because it means we would give jobs less time than before. If that sounds too dangerous I could increase the default `MAX_JOB_TIME` or add an additional variable.

I agree that this is problematic and a change to the previous behaviour. Especially when considering a slow infrastructure with longer times needed to sync assets between a central webUI instance and the workers. IMHO we can either just see MAX_JOB_TIME as "last resort catch all" timeout as you did now or we use specific timeouts for the individual states. How does this compare to the times we have for "broken/stale worker detection"? Should we at least try to define such timeouts at the same place and over the same way to prevent silly misconfigurations which don't make sense?

Martchus:
> How does this compare to the times we have for "broken/stale worker detection"?

This is done by the web UI (and the timeout is globally configured for the instance) but this change is on the worker side and it would make more sense to configure it per job (like `MAX_JOB_TIME`). I could add `MAX_DOWNLOAD_TIME` and set it to one hour by default.

okurz:
> I could add `MAX_DOWNLOAD_TIME` and set it to one hour by default.

that's what I mean with "specific timeouts for the individual states" but I would use the same terminology as we already have for the different states, i.e. from the constants.

For now I would go with your original approach to include the downloading within MAX_JOB_TIME. After all, if jobs spend the majority of their time downloading and not running that's not a healthy ratio.

Martchus:
I now already implemented it (as `MAX_SETUP_TIME`). So this is actually following the terminology of job states. This approach also differentiates the cases as the reason is set accordingly. This could be useful because job stuck waiting for the cache service like on openqaworker13 would be good candidates for automatic restarting.

okurz (approved):
yes, MAX_SETUP_TIME sounds better :)

PR #3897, opened 2021-05-11
Add new column to jobs table to flag jobs as archived

Martchus:
* Part of https://github.com/os-autoinst/openQA/pull/3858
* See https://progress.opensuse.org/issues/91347


okurz:
The PR itself looks fine. I suggest to undraft it and we can accept it even some days before you are ready with https://github.com/os-autoinst/openQA/pull/3858 which I am sure we will be able to complete eventually :)

Issue #3892, opened 2021-05-08
support mojo8 for a bit longer?


okurz:
Ah, well, this leaves me in a bit of an awkward spot by dropping support for Mojo 8. Fedora 34 released with Mojo 8. So now I either have to do a major, incompatible version bump of Mojo in Fedora 34, or introduce some kind of new "mojolicious9" package, or run our openQA instance on Rawhide, or leave it outdated (or revert this patch) for the next six months. None of those feel like fantastic options...:(

_Originally posted by @AdamWill in https://github.com/os-autoinst/openQA/issues/3830#issuecomment-834803586_

okurz:
can we do something about this?

AdamWill:
I have to go back and look at it again, but I think when I was first looking at this, it should at least be possible to conditionalize the routes on Mojo version. A bit ugly, but plausible.

I am currently looking through things that depend on Mojo in Fedora and seeing how things look for Mojo 9 compatibility, but it would be rather against Fedora policies to do a major release bump in a stable Fedora even if all the deps were portable. In theory we care about third parties, who might not appreciate whole chunks of Mojo's API disappearing in a stable update.

Then of course, we'd have to not introduce any other Mojo 9-isms for the next ~6 months until I could bump our servers to F35.

If I'd known this was happening before F34 Final freeze I could've stuffed Mojo 9 into it, but I didn't :(

okurz:
fixed by https://github.com/os-autoinst/openQA/pull/3898

PR #3898, opened 2021-05-12
Tweak routing again to work for Mojo < 9 or >=9.11

AdamWill:
We cut off support for Mojo before 9 pretty suddenly with #3830,
and it leaves Fedora in an awkward position as F34 shipped with
Mojo 8 and updating to 9 within a stable release would be against
Fedora's update policies. This adjusts the routing to work the
old way (before #3816, #3820 or #3830) for Mojolicious versions
before 9, and the new way for versions newer than 9.10; for
anything in the middle, we die with an error.


AdamWill:
@kraih for thoughts. this is a way we could achieve #3892 if we want to.

okurz (approved):

Martchus (commented):
It would be good if the `elsif`/`else` branches would be marked as uncoverable. I'm also wondering why you don't just provide a newer version of Mojolicious and openQA via an extra repository and use that for your deployments.

okurz:
> It would be good if the `elsif`/`else` branches would be marked as uncoverable. I'm also wondering why you don't just provide a newer version of Mojolicious and openQA via an extra repository and use that for your deployments.

are you questioning the development model of all non-rolling-release operating systems? :)

kraih:
Mojolicious 8.55 is almost  a year old (34 releases) and we are not testing new openQA releases with it. So it will definitely break for you every now and then in unexpected ways. Even if there are two separate code paths for the most obvious cases.

AdamWill:
It's just an option. Updating Mojo is another option. I don't particularly like side repositories, but that's also an option. Not merging this but carrying it downstream till F35 is another option.

AdamWill:
oh, btw, we're not using 8.55. I just put the number back to what it was before the change to 9.11. We are on 8.73.
People are pretty heavily against updating F34 to Mojo 9, so I think I'm going to use this one way or another; either merged upstream or carried downstream.

okurz:
I pushed a patch to fix the dependency check

AdamWill:
oh, thanks, sorry, I forgot that needed doing! thanks for the merge. As long as I'm updating the package for F34 I should notice if anything breaks too terribly; once I stop updating to new git bumps I'll try and remember to let you know so we can just drop this and go back to 9 only.

PR #3899, opened 2021-05-12
Revert "Do symlink assets as default"

Amrysliu:
Reverts os-autoinst/openQA#3863


Amrysliu:
Fix the regression issue on https://github.com/os-autoinst/os-autoinst/pull/1667. So we can wait for more time to see the results on o3.

okurz:
So far we have not seen cases where the original change caused problems and we would have a conflict here anyway. Closing. If you need to reopen please rebase.

PR #3900, opened 2021-05-12
ESLint (take a look, leave feedback)

kraih:
This patch replaces jsbeautifier with eslint and a [semistandard](https://github.com/standard/semistandard) config. Both are not packaged, and probably will never be packaged, since they are npm modules. Huge advantage is that eslint works with common IDEs like VSCode and that we can have a GitHub Action to validate the JavaScript. The GitHub Action automatically adds annotations to the diff for ever rule violation.

Locally you need to have `node` installed, and then you can install eslint with `npm`:
```
$ npm install
...
```
Many rule violations can be automatically fixed (the `fix` script is part of the `package.json` file):
```
npm run fix
...
```
Similarly you can also just do a plain lint run:
```
npm run lint
...
```
Even after fixing about 20k rule violations automatically, our JavaScript code still has about 500 violations left that would need manual intervention. For starters i've made all of them warnings. We can decide one by one if we want to fix them, or just keep them as warnings. Although for a lot of them i would strongly recommend fixing, since there's good reasons for why they are considered bad style.

Don't consider this a complete proposal, it's more of a discussion starter about what JavaScript style we want to choose going forward.


kraih:
You can see the remaining rule violation warnings in the GitHub Action [here](https://github.com/os-autoinst/openQA/pull/3900/checks?check_run_id=2567625740#step:5:10). Each one can be turned into an error with a change in `.eslintrc.json`.

.eslintrc.json
```diff
@@ -0,0 +1,69 @@
...
+  "parserOptions": {
+    "ecmaVersion": 12
+  },
+  "rules": {
+    "max-len": [
+      "warn",
+      {
+        "code": 120
+      }
+    ],
```
kraih, line 19: The extra rules below this line are the ones our code still violates, and which would need to be manually fixed.

kraih:
And i just learned that warnings also get annotations, they just don't make the GitHub Action fail.
warning

Martchus:
Looks good and tests still seem to pass. It would make sense to update the section `Checking for JavaScript problems` within the contributor documentation.

Martchus:
Of course the list of warnings shows that the tool is looking at files individually and hence is unaware that functions defined in another file exist. It is also unaware of JavaScript code within `on…=""` attributes. Maybe we should simply disable related warnings.

It also warns about things which I so far suppressed in `jshint` via
```
// jshint multistr: true
// jshint esversion: 6
```
in some files. These comments should likely be changed to work with `eslint`.

kraih:
Obsoleted by #4042.

PR #3902, opened 2021-05-13
Indent new lines for Logs

ilausuch:
Previosly the new lines appear without indentation. But for a
better reading and parsing the new lines that belongs to a
log entry are indented

(Pendent example)

progress.opensuse.org/issues/91527

### DIFF
--- a/lib/OpenQA/Log.pm
+++ b/lib/OpenQA/Log.pm
@@ -173,12 +173,18 @@ sub add_log_channel {
 
 # The default format for logging
 sub log_format_callback {
-    my ($time, $level, @lines) = @_;
+    my ($time, $level, @items) = @_;
+
+    my $lines = join("\n", @items, '');
+
+    # ensure indentation for multi-line output
+    $lines =~ s/(?<!\A)^/  /gm;
+
     # Unfortunately $time doesn't have the precision we want. So we need to use Time::HiRes
     $time = gettimeofday;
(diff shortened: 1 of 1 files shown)


lib/OpenQA/Log.pm
```diff
@@ -174,6 +174,11 @@ sub add_log_channel {
 # The default format for logging
 sub log_format_callback {
     my ($time, $level, @lines) = @_;
+    
+    foreach ( @lines ) {
+        $_ =~ s/(?<!\A)^ */  /gm;
+    }
```
okurz, line 180: please apply the same as in os-autoinst for the code style

okurz (changes requested):

lib/OpenQA/Log.pm
ilausuch: Done

Martchus:
This is now about the web UI and worker log, right? So the example from the commit message doesn't really apply. I'm also wondering whether we need this here at all. These logs are usually not that cluttered with unintended multi-line log messages.

ilausuch:
Yes, you are correct, these logs will be written on worker.log Not sure if this is necessary, but will maybe will help to better read of this log.

okurz (approved):

ilausuch:
Ok, let's close this to prevent confusion

PR #3903, opened 2021-05-14
Give permission of dmesg to worker

Amrysliu:
We would like to read the dmesg and find out if the qemu is killed
because of out of memory. So we give the permisson to worker.

Relate: https://progress.opensuse.org/issues/90974

### DIFF
--- a/profiles/apparmor.d/usr.share.openqa.script.worker
+++ b/profiles/apparmor.d/usr.share.openqa.script.worker
@@ -163,6 +163,11 @@
 
   }
 
+  profile /usr/lib/os-autoinst/check_qemu_oom {
+    #include <abstractions/base>
+    /usr/bin/dmesg rix,
+  }
+
   profile /usr/bin/lscpu {
     #include <abstractions/base>
 


Martchus (commented):
Wouldn't it make more sense to do the grepping in Perl instead of using an external application?

okurz:
I suggest the same. We should mention technical suggestions in https://github.com/os-autoinst/os-autoinst/pull/1672

Amrysliu:
> Wouldn't it make more sense to do the grepping in Perl instead of using an external application?

Thanks, fixed

Martchus:
It would have made sense to adjust the commit message as well.

coolo:
How can you ensure it's only os-autoinst code that runs this command and not the test code branched? dmesg reveals quite some delicate information about the system. Not the system under the test but about the infrastructure.

coolo:
... If you had an external application you could make sure in apparmor profile that only that application can run dmesg and test code can only find out if qemu was killed. Just saying

okurz:
hm, ok. good point.

@Amrysliu could you please try that? E.g. create a new profile section in https://github.com/os-autoinst/openQA/pull/3903/files#diff-d11aa1b125bd1f3f70fdc39be1066c1b64c870bc8849d568b347661e2c750e9aR201 for a new script that you put into the os-autoinst repo, e.g. check_qemu_oom and allow dmesg just for that?

coolo:
And yes, security kills all the good ideas :)

Amrysliu:
I update the os-autoinst pull request, and did tests in my local. The script `check_qemu_oom` doesn't have the permission issue. Do I need to modify the apparmor file?
Here is an example: http://10.67.19.157/tests/853

kalikiana:
> I update the os-autoinst pull request, and did tests in my local. The script `check_qemu_oom` doesn't have the permission issue. Do I need to modify the apparmor file?
> Here is an example: http://10.67.19.157/tests/853

The way I read it the suggestion is to expose a specific command, and not just dmesg, to avoid leaking info. As-is dmesg would be exposed completely.

okurz:
> EDIT: Is this redundant because of [os-autoinst/os-autoinst#1672](https://github.com/os-autoinst/os-autoinst/pull/1672) ?

It's not redundant but relying on what https://github.com/os-autoinst/os-autoinst/pull/1672 does

Amrysliu:
> 
> I assume you started a custom worker with your custom os-autoinst version. As soon as `check_qemu_oom` is called from a package installed os-autoinst and openQA then apparmor needs to cover it.

I see. Thanks. Modified the apparmor file.

Martchus (commented):
Did you test this? I'm just wondering whether we need permission for `check_qemu_oom` itself (as we have also rules for isotovideo and the videoencoder). Although I'm wondering whether this breaks the additional security you want to gain by making it a custom script again.

Amrysliu:
> Did you test this? I'm just wondering whether we need permission for `check_qemu_oom` itself (as we have also rules for isotovideo and the videoencoder). Although I'm wondering whether this breaks the additional security you want to gain by making it a custom script again.

I didn't test this. Because my tests are under my custom os-autoinst, it doesn't need any permission.  I agree with your point `maybe we need permission for check_qemu_oom`. 
@okurz WDYT?

Amrysliu:
> Did you test this? I'm just wondering whether we need permission for `check_qemu_oom` itself (as we have also rules for isotovideo and the videoencoder). Although I'm wondering whether this breaks the additional security you want to gain by making it a custom script again.

I did a test by those steps:
 1. Download a os-autoinst package from boo (os-autoinst-4.6.1622014432.10433fa2-lp152.8.1.x86_64.rpm). I got the link from the CI check `OBS Package Build` details.
 2. run `rpm -U xxx`  on my local environment.
 3. zypper up openQA 
 4. Run tests.
 It didn't report the issue that lack of permission of `check_qemu_oom` or `dmesg`.
Is my workflow correct or I miss something?

okurz (approved):
Your test workflow sounds elaborate and complete :)

Martchus:
> Is my workflow correct or I miss something?

Unless you haven't started apparmor yet, you need to start it first (`systemctl start apparmor.service`).

Amrysliu:
> > Is my workflow correct or I miss something?
> 
> Unless you haven't started apparmor yet, you need to start it first (`systemctl start apparmor.service`).

Sorry, I missed this question.

I remembered the apparmor was active at that time. 
But in order to confirm that I run the tests again,  here are the steps:
```
linux-jths:~ # systemctl status apparmor.service 
● apparmor.service - Load AppArmor profiles
   Loaded: loaded (/usr/lib/systemd/system/apparmor.service; enabled; vendor preset: enabled)
   Active: active (exited) since Thu 2021-05-27 10:05:04 CST; 5 days ago
 Main PID: 514 (code=exited, status=0/SUCCESS)
    Tasks: 0
   CGroup: /system.slice/apparmor.service

May 27 10:05:03 linux-jths apparmor.systemd[514]: Restarting AppArmor
May 27 10:05:03 linux-jths apparmor.systemd[514]: Reloading AppArmor profiles
May 27 10:05:04 linux-jths systemd[1]: Started Load AppArmor profiles.
linux-jths:~ # systemctl restart apparmor.service       

linux-jths:~ # systemctl reload apparmor.service

linux-jths:~ # rpm -qa|grep os-autoinst 
os-autoinst-4.6.1622014432.10433fa2-lp152.8.1.x86_64
os-autoinst-openvswitch-4.6.1615799823.a1b2e4ba-lp152.690.2.x86_64

linux-jths:~ # rpm -qf /usr/lib/os-autoinst/check_qemu_oom 
os-autoinst-4.6.1622014432.10433fa2-lp152.8.1.x86_64
linux-jths:~ # for i in {1..25}; do openqa-clone-job http://10.67.19.157/t1002 QEMURAM=4096; done
```
Here is an example: http://10.67.19.157/tests/1036  (The incomplete reason is: backend died: QEMU was killed due to the system being out of memory)

PR #3906, opened 2021-05-18
Install auto-update systemd files for SUSE as well and not only openSUSE

ggardet:
This will reduce the gap between Leap and SLE and it will also fix the build when targeting Backports repos in OBS.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -110,7 +110,7 @@ install-generic:
 
 	cp -Ra dbicdh "$(DESTDIR)"/usr/share/openqa/dbicdh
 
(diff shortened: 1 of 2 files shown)


okurz:
If it works this will be good. Could you try to prepare a temporary OBS project that shows that this works? E.g. branch devel:openQA or the generated OBS project from this PR checks and extend the project config to not only test openSUSE but also the various Leap+SLE ones?

okurz (approved):

Martchus (approved):

PR #3908, opened 2021-05-18
Add 2 spaces to indent the logs of the download assets

ilausuch:
This commit helps users distinguish between log lines and the output
of asset download operations. This adds two spaces before every output
line

https://progress.opensuse.org/issues/91527

### DIFF
--- a/lib/OpenQA/Worker/Engines/isotovideo.pm
+++ b/lib/OpenQA/Worker/Engines/isotovideo.pm
@@ -80,6 +80,8 @@ sub _save_vars {
     close($fd);
 }
 
+sub indent_log_raw_output ($text) { $text =~ s/(?<!\A)^/ /gmr }
+
 sub detect_asset_keys {
     my ($vars) = @_;
 
@@ -140,7 +142,7 @@ sub cache_assets {
             $status = $cache_client->status($asset_request);
         }
         my $msg = "Download of $asset_uri processed";
-        if (my $output = $status->output) { $msg .= ":\n$output" }
+        if (my $output = $status->output) { $msg .= ":\n" . indent_log_raw_output($output) }
         log_info($msg, channels => 'autoinst');
 
         $asset = $cache_client->asset_path($webui_host, $asset_uri)


ilausuch:
Replaces #3881

lib/OpenQA/Worker/Engines/isotovideo.pm
```diff
@@ -80,6 +80,11 @@ sub _save_vars {
     close($fd);
 }
 
+sub indent_log_raw_output {
```
Martchus, line 83: I would prefer to use signatures in new code (which could make this a one-liner function).

Martchus (commented):
It would be good to have tests. We've started to add tests for these functions in `t/24-worker-engine.t` at some point so you could extend these existing tests.

lib/OpenQA/Worker/Engines/isotovideo.pm
```diff
@@ -80,6 +80,11 @@ sub _save_vars {
...
+    my $text = @_;
```
perlpunk, line 84: This will assign the number of items in the `@_` array to `$text`.

lib/OpenQA/Worker/Engines/isotovideo.pm
```diff
@@ -80,6 +80,11 @@ sub _save_vars {
...
+    if ($text[0] // '') =~ s/(?<!\A)^/ /gm;
```
perlpunk, line 85: This doesn't even compile for me

lib/OpenQA/Worker/Engines/isotovideo.pm
```diff
@@ -80,6 +80,12 @@ sub _save_vars {
     close($fd);
 }
 
+sub indent_log_raw_output {
+    my ($text) = @_;
+    $text =~ s/(?<!\A)^/ /gm;
```
perlpunk, line 85: It would be possible to use the `/r` flag on the substitution, then the return value will be the substituted text. Combined with a signature like @Martchus suggested, it will be a very short one liner:
```
sub indent_log_raw_output($text) { $text =~ s/(?<!\A)^/ /gmr }
```

okurz (changes requested):
https://github.com/os-autoinst/openQA/pull/3908/files#r634455689

lib/OpenQA/Worker/Engines/isotovideo.pm
ilausuch: Thanks, done

lib/OpenQA/Worker/Engines/isotovideo.pm
ilausuch: Done

perlpunk (commented):
Finally a test would be good :)

lib/OpenQA/Worker/Engines/isotovideo.pm
```diff
@@ -140,7 +142,7 @@ sub cache_assets {
             $status = $cache_client->status($asset_request);
         }
         my $msg = "Download of $asset_uri processed";
-        if (my $output = $status->output) { $msg .= ":\n$output" }
+        if (my $output = $status->output) { $msg .= ":\n" . indent_log_raw_output($output) }
```
perlpunk, line 145: I'm wondering why the first line shouldn't be indented here.
After all the return value of that function is already going to be a continuation line of the log message, so *all* lines should be indented by that function.

The description of the PR should include a before and after example.

PR #3910, opened 2021-05-19
Update docs about the special case of  BACKEND variable precedence.

radolin:
The docs state that variable values set as part of the API request that triggers the jobs will
'win' over values set at any of the other locations, but this is not true for the special case of BACKEND variable.


Martchus (approved):

okurz:
Please delete the dot at the end of line of the git commit message

radolin:
> Please delete the dot at the end of line of the git commit message

Done, sorry, I hope to get used to not putting a dot or write a git pre-commit check.

okurz (approved):

PR #3912, opened 2021-05-19
Do not build openQA-devel for archs where it would be unresolvable

Martchus:
chromedriver does not exist under ppc ppc64 ppc64le s390x so only the
package opennQA-no-selenium-devel can be used there.


okurz (approved):

Martchus:
I've been adding ppc64le to the project of the OBS package check and `openQA-devel-test` shows up as excluded for that architecture as expected.

okurz:
yes, but now other failures show up as you added ppc64le but for a repository that does not provide the according packages. You can force-merge though

Martchus:
Yes, I suppose it was wrong to enable it for TW as well.

PR #3915, opened 2021-05-20
containers: Fix scheduler and websockets URLs in webui container

ilausuch:
The webui container tries to access to the scheduler
but we obtain this error

[warn] Unable to wakeup scheduler: Connection refused

The reason is that is trying to access to the default url: localhost

https://progress.opensuse.org/issues/92833


Martchus (approved):

okurz (changes requested):
s/Te reason is that is trying/The reason is that the services are trying/

Also please follow http://www.extremeprogramming.org/rules/bugs.html

You found a bug so please have a test *first* that shows the problem, then apply a fix which makes the test pass

tools/test_containers_compose
```diff
@@ -9,4 +9,13 @@ for i in webui worker; do
     sudo docker-compose up -d &&
     (docker-compose ps --services --filter status=stopped | grep "^[[:space:]]*$") || (docker-compose logs; sudo docker-compose ps; exit 1)
   ) || exit 1;
+
+  if [[ $i == "webui" ]]; then
+    (
+      docker-compose exec webui /usr/share/openqa/script/clone_job.pl --from https://openqa.opensuse.org --host http://nginx:9526 1738950 || echo "Error executing a job";
+      sleep 1;
+      docker-compose logs webui;
+      docker-compose logs webui | grep "Unable to wakeup scheduler: Connection refused" && exit 1
```
okurz, line 18:
```suggestion
      docker-compose exec webui openqa-clone-job --from https://openqa.opensuse.org --host http://nginx:9526 1738950 || echo "Error executing a job"
      sleep 1
      docker-compose logs webui
      docker-compose logs webui | grep "Unable to wakeup scheduler: Connection refused" && exit 1
```

okurz (changes requested):

tools/test_containers_compose
```diff
@@ -9,4 +9,13 @@ for i in webui worker; do
     sudo docker-compose up -d &&
     (docker-compose ps --services --filter status=stopped | grep "^[[:space:]]*$") || (docker-compose logs; sudo docker-compose ps; exit 1)
   ) || exit 1;
+
+  if [[ $i == "webui" ]]; then
+    (
+      docker-compose exec webui /usr/share/openqa/script/clone_job.pl --from https://openqa.opensuse.org --host http://nginx:9526 1738950 || echo "Error executing a job";
```
Martchus, line 15: This job might be cleaned up on o3 at some point so don't rely on that.

tools/test_containers_compose
```diff
@@ -9,4 +9,13 @@ for i in webui worker; do
...
+      sleep 1;
```
Martchus, line 16: Shouldn't there be a better way?

Martchus (changes requested):

tools/test_containers_compose
ilausuch: I put that in not ready because I am still working on that test. I'll move to WIP or draft

tools/test_containers_compose
ilausuch: yes, this is a bad point of doing this in this way... should be a better way to solve this

ilausuch: However we are testing that API can connect with the scheduler to launch tests...

kalikiana: I might be missing context but relying on any data on o3 seems is a bad idea unless you're validating the deployment of this instance. This should use a fixture or create a job on the fly.

ilausuch: Yes I agree

okurz: It feels like you try to achieve too much here. I think having just a manual test properly defined in the commit message is enough. Having an automatic test that relies on o3 is too much

tools/test_containers_compose
ilausuch: I agree

ilausuch:
Ok, I created 3 commits
1. Solves the problem
2. Test the problem and its solution
3. Refactor the test

tools/test_containers_compose
ilausuch: I did a new approach, please check this new commit

tools/test_containers_compose
ilausuch: Now is solved with an easy way

container/webui/conf/openqa.ini
```diff
@@ -1,5 +1,5 @@
 [auth]
-method = OpenID
+method = Fake
```
okurz, line 2: This is not something we should introduce into our production templates of the container setup. Otherwise people might risk spawning an openQA instance with no proper authentication

tools/test_containers_compose
```diff
@@ -2,11 +2,44 @@
...
   ) || exit 1;
-done
+}
+
+#test webui
+(
+  cd container/webui/ &&
+  setup_containers &&
+  docker-compose exec -T webui openqa-cli api -X POST jobs ISO=foo.iso DISTRI=my-distri FLAVOR=my-flavor VERSION=42 BUILD=42 TEST=my-test --host http://nginx:9526 || (echo "Error executing a job"; exit 1) &&
+  sleep 1 &&
```
okurz, line 18: You are using `sleep()` or are simulating a sleep-like functionality. This is introducing the risk of either unnecessarily wasting time or just shifting race conditions and making them even harder to investigate. You *can* use sleep but there are only very limited cases where is the best solution (e.g. yield threads in low-level system code). Here you have the classical problem in software design of [Synchronization](https://en.wikipedia.org/wiki/Synchronization_(computer_science)) . sleep should not be used in place of proper synchronization methods. https://blogs.msmvps.com/peterritchie/2007/04/26/thread-sleep-is-a-sign-of-a-poorly-designed-program/ is one of multiple references that explains this: "The original problem is likely a timing/synchronization issue, ignoring it by hiding it with Thread.Sleep is only going to delay the problem and make it occur in random, hard to reproduce ways." And if the problem does not occur at least time is wasted. https://testing.googleblog.com/2021/03/test-flakiness-one-of-main-challenges.html states that as well: "Do NOT add arbitrary delays as these can become flaky again over time and slow down the test unnecessarily."

We are waiting because some condition changes some time ... keyword(s) is/are *some time*. It is important to know the exact condition that the code should wait for and check for that instead of delaying program execution for an arbitrary amount of time. For openQA tests in general just try to do what you as a human user would also do. You would not look on your watch waiting for exactly 100s before you look at the screen again, right? :)

Likely you want a loop call to `docker-compose logs` with a timeout, e.g. `timeout 60 until ...`

tools/test_containers_compose
```diff
@@ -2,11 +2,44 @@
...
+  (! docker-compose logs webui | grep "Unable to wakeup scheduler") || (docker-compose logs webui; exit 1)
+) || exit 1;
+
+#test worker
+(
+  cd container/worker/ &&
+  setup_containers
+) || exit 1;
+
+# for i in webui worker; do
```
okurz, line 28: Don't add disabled code as nobody but you will understand why this is not active. Better leave it out and keep in your local git repository, either in `git stash` or a temporary "WIP"-commit

okurz (changes requested):
s/Added/Add/ in second commit

container/webui/docker-compose.yaml
```diff
@@ -90,6 +90,7 @@ services:
       MOJO_LISTEN: "http://0.0.0.0:9526"
       OPENQA_SCHEDULER_HOST: "scheduler"
       OPENQA_WEB_SOCKETS_HOST: "websockets"
+      MOJO_CLIENT_DEBUG: 1
```
Martchus, line 93: This isn't supposed to stay when we merge the PR, right?

ilausuch: No, you have to consider that this is still not-ready

tools/test_containers_compose
ilausuch: Done

container/webui/conf/openqa.ini
ilausuch: Done

tools/test_containers_compose
```diff
@@ -2,11 +2,44 @@
...
+    sudo MOJO_CLIENT_DEBUG=1 docker-compose up -d &&
     (docker-compose ps --services --filter status=stopped | grep "^[[:space:]]*$") || (docker-compose logs; sudo docker-compose ps; exit 1)
   ) || exit 1;
-done
+}
+
+#test webui
+(
+  cd container/webui/ &&
+  sed -e "s/method = OpenID/method = Fake/" < conf/openqa.ini > /tmp/openqa.ini &&
```
okurz, line 32: this means that executing the test would change the files in the git checkout, right? I consider this bad style and suggest instead what we follow in other tests: Create a temporary directory and generate files with the correct setup in there, e.g.

```suggestion
conf=$(mktemp -d)
trap 'rm -rf $conf' EXIT
…
  sed -e "s/method = OpenID/method = Fake/" < conf/openqa.ini > $conf/openqa.ini
  …
  export OPENQA_CONFIG=$conf
```

See for example https://github.com/os-autoinst/openQA/blob/master/t/03-auth.t#L32

okurz (changes requested):

tools/test_containers_compose
```diff
@@ -2,11 +2,44 @@
...
+  mv /tmp/openqa.ini conf/openqa.ini &&
+  printf "[nginx]\nkey = 1234567890ABCDEF\nsecret = 1234567890ABCDE\n" > conf/client.conf &&
+  setup_containers &&
+  docker-compose exec -T webui openqa-cli api -X POST jobs ISO=foo.iso DISTRI=my-distri FLAVOR=my-flavor VERSION=42 BUILD=42 TEST=my-test \
+   --host http://nginx:9526 --apikey 1234567890ABCDEF --apisecret 1234567890ABCDEF || (echo "Error executing a job"; exit 1) &&
```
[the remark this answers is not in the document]
ilausuch: Good catch!

tools/test_containers_compose
ilausuch: I will copy all files from container/webui to the temp dir, because, also, the docker-compose creates a new directory "workdir"

tools/test_containers_compose
```diff
@@ -2,11 +2,53 @@
...
-done
+}
+
+test_webui(){
+  (
+    workspace=$(mktemp -d) &&
+    cp -r container/webui "$workspace" &&
+    cd "$workspace/webui" &&
+    sed -e "s/method = OpenID/method = Fake/" < conf/openqa.ini > /tmp/openqa.ini &&
+    mv /tmp/openqa.ini conf/openqa.ini &&
```
okurz, line 35:
```suggestion
    sed -i -e "s/method = OpenID/method = Fake/" conf/openqa.ini &&
```

tools/test_containers_compose
```diff
@@ -2,11 +2,53 @@
...
+    for retry in {2..0}; do sudo docker-compose build && break; echo "Remaining retries $retry"; done &&
+    sudo MOJO_CLIENT_DEBUG=1 docker-compose up -d &&
     (docker-compose ps --services --filter status=stopped | grep "^[[:space:]]*$") || (docker-compose logs; sudo docker-compose ps; exit 1)
   ) || exit 1;
-done
+}
+
+test_webui(){
+  (
+    workspace=$(mktemp -d) &&
```
okurz, line 31: please check that the temporary directory will be removed again on both passed as well as failed test runs. This is why I suggested the "trap".

okurz (changes requested):

tools/test_containers_compose
ilausuch: done

tools/test_containers_compose
ilausuch: done

ilausuch:
With this last commit, a new command is added to the trap to remove all the container and prevent undesired containers running

tools/test_containers_compose
```diff
@@ -2,11 +2,54 @@
...
+    trap 'docker-compose down; sudo rm -rf "$workspace"' EXIT &&
```
Martchus, line 32: Somehow I'm not comfortable running `sudo rm -rf` on an path which is the unvalidated output of another command. Are the `sudo` and the `-f` flag really necessary?

tools/test_containers_compose
```diff
@@ -2,11 +2,54 @@
 
 set -euo pipefail
 
-for i in webui worker; do
+wait_until(){
```
Martchus, line 5:
```suggestion
wait_until() {
```

The same counts for the other instances.

tools/test_containers_compose
```diff
@@ -2,11 +2,54 @@
...
-    sudo docker-compose up -d &&
+    for retry in {2..0}; do sudo docker-compose build && break; echo "Remaining retries $retry"; done &&
+    sudo MOJO_CLIENT_DEBUG=1 docker-compose up -d &&
     (docker-compose ps --services --filter status=stopped | grep "^[[:space:]]*$") || (docker-compose logs; sudo docker-compose ps; exit 1)
   ) || exit 1;
-done
+}
+
+test_webui(){
+  (
```
Martchus, line 30: I'm wondering why one would spawn a sub-shell all the time. This doesn't make the code more readable.

tools/test_containers_compose
ilausuch: Sudo yes because the workdir is created by docker-compose that is executed by sudo.

tools/test_containers_compose
ilausuch: I have to use this to prevent to contextualize the workspace for each test and preventing the use of pushd and popd t (or using cd..)
I remember en in previous review that we prefer sub-shells

tools/test_containers_compose
ilausuch: works without force (-f)

okurz (approved):

Martchus (approved):

PR #3918, opened 2021-05-21
Simplify OpenQA::Schema::ResultSet::Jobs

okurz:


Amrysliu (commented):
LGTM

lib/OpenQA/Schema/ResultSet/Jobs.pm
```diff
@@ -202,9 +200,7 @@ sub create_from_settings {
     }
 
     # assign default for WORKER_CLASS
-    if (!$settings{WORKER_CLASS}) {
-        $settings{WORKER_CLASS} = 'qemu_' . ($new_job_args{ARCH} // 'x86_64');
-    }
+    $settings{WORKER_CLASS} //= 'qemu_' . ($new_job_args{ARCH} // 'x86_64');
```
Martchus, line 203: I suppose

```suggestion
    $settings{WORKER_CLASS} ||= 'qemu_' . ($new_job_args{ARCH} // 'x86_64');
```

would be the syntax for *not* changing the behavior.

lib/OpenQA/Schema/ResultSet/Jobs.pm
```diff
@@ -394,23 +359,13 @@ sub complex_query {
...
-        if (my $order_by = $args{order_by}) {
-            $attrs{order_by} = $order_by;
-        }
-    }
-    else {
-        $attrs{order_by} = ['me.id DESC'];
-    }
-
-    $attrs{join} = \@joins if @joins;
+    $attrs{order_by} = $args{order_by} ? $args{order_by} : ['me.id DESC'];
```
Martchus, line 367:
```suggestion
    $attrs{order_by} = $args{order_by} || ['me.id DESC'];
```

Martchus (approved):

Martchus (approved):

Martchus (dismissed):
The test failure might actually be related.

lib/OpenQA/Schema/ResultSet/Jobs.pm
```diff
@@ -215,203 +210,135 @@ sub create_from_settings {
...
-        push(
-            @conds,
-            {
-                -or => [
-                    'me.t_created'  => $agecond,
-                    'me.t_started'  => $agecond,
-                    'me.t_finished' => $agecond
-                ]});
+    if ($args->{failed_modules}) {
+        push @joins, 'modules';
```
okurz: this is interesting to see that this code section is not covered. Means we can completely delete it and no tests will fail. Maybe this is actually obsoleted and replaced by the query for "modules" together with "modules_result". Should we just delete this? Interesting though is that "failed_modules" as a word does appear in tests.

EDIT: I have now added three more commits deleting dead code or at least uncovered code

perlpunk: Seems it was removed from the HTML form in 4865792863e7f298a23b3640052c748b05d265e0 but forgotten in the backend

okurz: ah, perfect! I will include that reference in the git commit message. Thank you.

okurz:
Added one more commit https://github.com/os-autoinst/openQA/pull/3918/commits/8686c474fcea4420bdc9be09b052aecd68bf5fdf which removes the uncovered code about "assetid". With this I expect that the codecov patch check will be happy and we will be ready to merge.

Martchus (commented):
I hope `failed_modules` is not exposed (or at least used) via the API.

okurz:
> I hope `failed_modules` is not exposed (or at least used) via the API.

Not sure if it's exposed but it's not documented, not tested, we did not have any reports from users about it and very likely not used. Can you approve the PR or do you have other open points that I overlooked?

kalikiana (approved):
Just for reference, my thoughts:
- Love the simplifications
- It seems the test is actually faster now since we can drop HEAVY, which is awesome.
- We seem to be dropping `assetid` and `failed_modules`. Probably fine for undocumented/unsupported API. On the off chance that it is used somewhere, this comment may help a person investigating what happened.

Martchus (changes requested):
> Not sure if it's exposed

I've just checked and it is explicitly exposed by the code in `OpenQA::WebAPI::Controller::API::V1::Job::list` so at least the parameter handling should be removed there as well for consistency and to make it clear that we're doing an API breaking change here.

> but it's not documented

If we'd go by that we could delete a lot of code I suppose.

> not tested

Not sure whether this makes the code itself worse, though. Maybe somebody tested it manually at some point and it still works.

---

I don't know whether we should keep it but as mentioned in the first paragraph, at least remove the parameter handling in the API controller as well.

lib/OpenQA/Schema/ResultSet/Jobs.pm
```diff
@@ -541,29 +444,15 @@ sub mark_job_linked {
...
-    elsif ($referer) {
-        log_debug("Unrecognized referer '$referer'");
-    }
+    return undef unless $referer;
+    return log_debug("Unrecognized referer '$referer'")
+      unless grep { $referer eq $_ } @{$app->config->{global}->{recognized_referers}};
+    my $job = $self->find({id => $jobid});
+    return unless $job;
+    my $comments = $job->comments;
+    return undef if grep { $_->label // '' eq 'linked' } $comments->all;
```
perlpunk, line 453: A `grep` might be slower than a loop with `last` because grep iterates over all items. `List::Util::first()` might be better here.

Martchus: The `label` function is a bit costly so I guess using `List::Util::first` is a good idea.

perlpunk: oh, and `->all()` will instantiate all comments, even when using `List::Util::first(), so maybe going back to iterating via `next` is even better.

lib/OpenQA/WebAPI/Controller/API/V1/Job.pm
```diff
@@ -102,9 +102,9 @@ sub list {
 
     my %args;
     $args{limit} = $limit;
-    my @args = qw(build iso distri version flavor maxage scope group
-      groupid page before after arch hdd_1 test machine worker_class
-      failed_modules modules modules_result);
+    my @args = qw(build iso distri version flavor scope group groupid page
```
Martchus, line 105: Looks like you now also got rid of `maxage`. Is it really useless to be able to filter out old jobs?

Martchus (dismissed):

lib/OpenQA/WebAPI/Controller/API/V1/Job.pm
okurz: Yes. Is it used somewhere?

okurz:
> > Not sure if it's exposed
> 
> I've just checked and it is explicitly exposed by the code in `OpenQA::WebAPI::Controller::API::V1::Job::list` so at least the parameter handling should be removed there as well for consistency and to make it clear that we're doing an API breaking change here.

I think I did that already in https://github.com/os-autoinst/openQA/pull/3918/commits/755f541aa24df3723b771022279b8d1aab10e2df , didn't I?

Martchus:
Oh, sorry if the commit was already present before I wrote that. "Not sure if it's exposed" sounded like you didn't take any effort into looking how the API route uses the schema.

lib/OpenQA/WebAPI/Controller/API/V1/Job.pm
Martchus: You'll have to ask our users. (This is not an internal API.)

lib/OpenQA/Schema/ResultSet/Jobs.pm
okurz: ok. changed to https://metacpan.org/pod/List::Util#first now. Unless you show me a nicer one-line way or some performance figures I would prefer to keep the array of comment references. In python it's even expected that list operations are faster than iterating over individual members.

okurz:
> Oh, sorry if the commit was already present before I wrote that. "Not sure if it's exposed" sounded like you didn't take any effort into looking how the API route uses the schema.

Well, I wasn't sure if that line I changed is actually the only "exposing" place.

Martchus:
It is. The API for the overview seems to handle the parameter differently (not relying on `complex_query`).

lib/OpenQA/WebAPI/Controller/API/V1/Job.pm
okurz: I already asked and nobody responded. Can we please expect the missing removal to be an oversight of https://github.com/os-autoinst/openQA/commit/4865792863e7f298a23b3640052c748b05d265e0 and if someone complains we will add it back?

lib/OpenQA/Schema/ResultSet/Jobs.pm
perlpunk: Python is not perl, and even in python it would be slower to instantiate all objects and iterate over that list instead of using an iterator, in case it's expected that the loop will be exited before the end.

If the comment objects include the DateTime fields here and there are many comments, that would be unfortunate.

okurz: But it's only an array of *references* at this point, isn't it?

perlpunk: Yes, but the array of references is an array of fully instantiated objects, which include *all* columns of the comments table, including two extra costly DateTime columns.
When iterating via `next`, each object will be instantiated in the loop.
I'm not sure yet when this code is executed, so it might not make much difference, and probably there are usually not that many comments per job, but I would use an iterator.
I don't insist, I give up :)

lib/OpenQA/WebAPI/Controller/API/V1/Job.pm
okurz: Same reasoning as for `failed_modules` applies for `maxage`, see https://github.com/os-autoinst/openQA/pull/3918/commits/f055c31b6cc8744bfe066474a70d3877dc08bfb1 . It's undocumented, untested, not used
anywhere within openQA and very likely not used anywhere outside the
project scope as well as no one ever again reported an issue about that, asked a question or brought it up in any context.

lib/OpenQA/Schema/ResultSet/Jobs.pm
Martchus: I'd opt for going back to the good old loop. It is as easy to understand as the new code, just a little bit longer. Saving a few lines here is likely not worth anything (even if the performance impact is very slight it is more than anything).

okurz: If Perl is that disappointing and saving full objects in the array then, yes, it could be less efficient. We might need to change that. But reg. "not worth anything" I did it also to fix the patch code coverage check :)

perlpunk: perl is not disappointing here.
`next()` will retrieve row by row via DB cursor, and then in each iteration another object is instantiated.
That is what `next()` is for - to avoid fetching data from DB (to avoid having all data in an array, or to exit the loop before the cursor is finished).
If you call `all()` then perl/DBIC does what it is **asked** and fetches **all** items from the database. And of course it has to instantiate the objects then. When should it instantiate the objects otherwise? The data must be saved somewhere.
If you want to iterate with a `for` loop or a `map` over data, then all the data must be there before the first loop iteration.

okurz: Right, now I get it. Thank you

okurz:
Updated:
* Extracted https://github.com/os-autoinst/openQA/pull/3963/commits/91e52d7fe731bba1558f4e1bd0295ef1e28c17a7 into a separate https://github.com/os-autoinst/openQA/pull/3963 as it's actually independant (the other commits could not easily be split or turned around in order)
* Using `return undef unless $comments->find({text => {like => 'label:linked%'}})` for comment label lookup

lib/OpenQA/Schema/ResultSet/Jobs.pm
okurz: Using `return undef unless $comments->find({text => {like => 'label:linked%'}})` for comment label lookup

Martchus:
Btw, Have you checked with production data whether using `return undef unless $comments->find({text => {like => 'label:linked%'}})` is notably slower (as it might be an additional SQL query)?

okurz:
> Btw, have you checked with production data whether using `return undef unless $comments->find({text => {like => 'label:linked%'}})` is notably slower (as it might be an additional SQL query)?

No, I haven't. As the change is only in the referral handler for openQA links coming from elsewhere so not that common I opted to do that change without

lib/OpenQA/WebAPI/Controller/API/V1/Job.pm
ilausuch: Should be groupids (plural) in this list too?

ilausuch:
Is groupids in the test?

ilausuch (approved):
To me this is ok, but It would be nice to have a test for multiple groupid case. For instance, in t/10-tests-overview.t

```
$t->get_ok('/tests/overview?groupid=1001&groupid=1002')->status_is(200);
$t->element_exists('#res_DVD_x86_64_kde');        # Included groupid 1001
$t->element_exists('#flavor_NET_arch_x86_64');    # Included groupid 1002
```

PR #3799, opened 2021-03-18
Add a switch to enable the symlink feature

Amrysliu:
Compare #3712, this pull request includes a switch to enable the symlink feature.
The related code can be found in the separate commit.

PR #3920, opened 2021-05-26
worker: Fix handling CASEDIR/PRODUCTDIR/NEEDLES_DIR in certain cases

Martchus:
worker: Fix handling CASEDIR/PRODUCTDIR/NEEDLES_DIR in certain cases

This change ensures that the default-assignment for PRODUCTDIR is a
relative directory (unless the feature switch to use absolute paths is
turned on). That is required when a custom CASEDIR has been specified
because in this case it is expected that not only Perl modules but also
`main.pm` itself is loaded from that custom CASEDIR (unless PRODUCTDIR is
set explicitly to some other location). So this should fix
https://progress.opensuse.org/issues/93101.

When only changing the behavior for PRODUCTDIR the needles could not be
loaded anymore (unless NEEDLES_DIR is specified explicitly) because the
checkout of CASEDIR does not necessarily contain needles. To fix this, the
symlinking of the needles directory is done in that case. See the verbose
comments within the code for further explanations.

Note that the absolute PRODUCTDIR has been introduced by the commit
55420c174299d545d654d4c6266415a8e9ee8dd2 to fix two problems. The first
problem won't occur again because now the relative path for PRODUCTDIR is
computed correctly. The second problem is the needle problem mentioned in
the previous paragraph and thus handled as well.

Tested locally with
* no custom CASEDIR, NEEDLES_DIR and PRODUCTDIR
* custom CASEDIR, NEEDLES_DIR, PRODUCTDIR as produced by assigned by
  `openqa-clone-custom-git-refspec`
* only a custom CASEDIR as assigned by `openqa-investigate` (case mentioned
  in https://progress.opensuse.org/issues/93101)
* only a custom NEEDLES_DIR pointing a some Git repository

---

When extending the unit tests I also found a second issue, see the other commit
message.

### DIFF
--- a/lib/OpenQA/Worker/Engines/isotovideo.pm
+++ b/lib/OpenQA/Worker/Engines/isotovideo.pm
@@ -21,12 +21,13 @@ use warnings;
 use Mojo::Base -signatures;
 use OpenQA::Constants qw(WORKER_SR_DONE WORKER_EC_CACHE_FAILURE WORKER_EC_ASSET_FAILURE WORKER_SR_DIED);
 use OpenQA::Log qw(log_error log_info log_debug log_warning get_channel_handle);
-use OpenQA::Utils qw(asset_type_from_setting base_host locate_asset looks_like_url_with_scheme testcasedir productdir);
+use OpenQA::Utils
+  qw(asset_type_from_setting base_host locate_asset looks_like_url_with_scheme testcasedir productdir needledir);
 use POSIX qw(:sys_wait_h strftime uname _exit);
 use Mojo::JSON 'encode_json';    # booleans
 use Cpanel::JSON::XS ();
 use Fcntl;
-use File::Spec::Functions 'catdir';
+use File::Spec::Functions qw(abs2rel catdir file_name_is_absolute);
 use File::Basename 'basename';
 use Errno;
 use Cwd 'abs_path';
@@ -332,6 +333,7 @@ sub engine_workit {
     log_debug('Job settings:');
     log_debug(join("\n", '', map { "    $_=$vars{$_}" } sort keys %vars));
 
+    # cache/locate assets, set ASSETDIR
     my $shared_cache;
     my $assetkeys = detect_asset_keys(\%vars);
     if (my $cache_dir = $global_settings->{CACHEDIRECTORY}) {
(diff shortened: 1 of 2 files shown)


okurz:
regarding tests, https://github.com/os-autoinst/openQA/pull/3799/commits/5b000cdf68548d8375f869892ab3a42b4637069c#diff-831bb8b106b0ca2b8cfaa3dd2dfd6627b1aa8e86349cd67aa279afd922091d0eR236 was were the code

```perl
like $result->{child}->process_id, qr/\d+/, 'don\'t create symlink when CASEDIR is an url address';
```

was introduced that looks unusual to us. We just needed to get the changes merged by force eventually after multiple reverts. The current problem that you try to fix and the unusual test are gladly the only known left issues we see in this area right now :)

lib/OpenQA/Worker/Engines/isotovideo.pm
```diff
@@ -345,33 +347,48 @@ sub engine_workit {
...
+    my $target_name           = path($default_casedir)->basename;
+    my $has_custom_dir        = $vars{CASEDIR} || $vars{PRODUCTDIR};
+    my $casedir               = $vars{CASEDIR} //= $absolute_paths ? $default_casedir : $target_name;
+    if ($casedir eq $target_name) {
+        $vars{PRODUCTDIR} //= substr($default_productdir, rindex($default_casedir, $target_name));
+        if (my $error = _link_repo($default_casedir, $pooldir, $target_name)) { return $error }
+    }
+    else {
+        $vars{PRODUCTDIR} //= $absolute_paths ? $default_productdir : abs2rel($default_productdir, $default_casedir);
+    }
```
okurz, line 366: Instead of this if/else here I would prefer a function with a good name and maybe exception handling instead of error value checking and returning. Could you please give it a try? If it doesn't work keep it as is

t/24-worker-engine.t
```diff
@@ -221,17 +222,19 @@ subtest 'problems when syncing tests' => sub {
...
-    my $settings       = {DISTRI => 'foo'};
-    my $job            = OpenQA::Worker::Job->new($worker, $client, {id => 12, settings => $settings});
     my $pool_directory = tempdir('poolXXXX');
+    my $worker         = Test::FakeWorker->new(pool_directory => $pool_directory);
+    my $client         = Test::FakeClient->new;
     my $casedir        = testcasedir('foo', undef, undef);
-    $worker->pool_directory($pool_directory);
-    my $result = OpenQA::Worker::Engines::isotovideo::engine_workit($job);
+
+    # error case: CASEDIR missing
```
okurz, line 230: looks like you want to define separate subtests for these? As so far we are indented only one level I think we could have that second level :)

okurz (changes requested):

Amrysliu (approved):
The code looks much better and clearer than what I wrote.

lib/OpenQA/Worker/Engines/isotovideo.pm
Martchus: The whole if/else-block is *not* about error handling but about distinguishing different cases (none of the cases is an error case). Changing the error handling of `_link_repo` to use exceptions to get rid of the nested if makes not much sense because then we needed a try-catch here unless we change the error handling of the overall function. But that's out of scope for this PR as I don't want to change too much at a time. And even if that wasn't the case, keep in mind that we return a hashref in the error case which is handled by the outer function specifically so I'd say there's not much to gain.

okurz: I mainly meant to follow https://gist.github.com/wojteklu/73c6914cc446146b8b533c0988cf8d29#comments-rules or https://blog.usejournal.com/stop-writing-code-comments-28fef5272752

okurz (changes requested):

lib/OpenQA/Worker/Engines/isotovideo.pm
Martchus: I'm not sure which points of that lengthy list you're getting at. If you don't like the comment I can remove it. The code is not hard to understand and I can note down the reasoning why this is the desired behavior somewhat else.

okurz: nevermind. I think we will talk about it another day :)

lib/OpenQA/Worker/Engines/isotovideo.pm
```diff
@@ -345,33 +347,48 @@ sub engine_workit {
...
+    # - If a custom CASEDIR/PRODUCTDIR has been specified but no custom NEEDLES_DIR we assume that the default
+    #   needles directory is supposed to be used and make the assignment/symlink accordingly.
+    # - If the specified NEEDLES_DIR is an absolute path or an URL we assume that it points to a custom location
+    #   and keep it as-is.
+    my $default_needles_dir     = needledir(@vars_for_default_dirs);
+    my $needles_dir             = $vars{NEEDLES_DIR};
+    my $need_to_set_needles_dir = !$needles_dir && $has_custom_dir;
+    if ($need_to_set_needles_dir && $absolute_paths) {
+        # simply set the default needles dir when absolute paths are used
+        $vars{NEEDLES_DIR} = $default_needles_dir;
```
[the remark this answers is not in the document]
Martchus: This is covered now. And when I added the test I've noticed that the actual issue (https://progress.opensuse.org/issues/93101) isn't fixed in that case and added the commit "Fix loading main.pm from custom CASEDIR with …". Now all new code should be covered, including different branches of the ternary operator.

Martchus:
The failing style check is unrelated, I've been creating https://github.com/os-autoinst/openQA/pull/3922.

okurz (approved):

PR #3922, opened 2021-05-31
Fix failing stylecheck for tools/test_containers_compose

Martchus:
Prevents

```
In tools/test_containers_compose line 23:
  sudo MOJO_CLIENT_DEBUG=1 docker-compose up -d &&
                                                ^-- SC2015: Note that A && B || C is not if-then-else. C may run when A is true.
```

It doesn't make much sense to use `&&` here anyways because the script uses
`-e`.

### DIFF
--- a/tools/test_containers_compose
+++ b/tools/test_containers_compose
@@ -19,8 +19,8 @@ wait_until() {
 }
 
 setup_containers() {
-  for retry in {2..0}; do sudo docker-compose build && break; echo "Remaining retries $retry"; done &&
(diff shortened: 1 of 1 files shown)


ilausuch (changes requested):
It's important that if some of these lines fail all the function fails, this is the reason of the &&.

Martchus:
@ilausuch But the script has `set -e` so this should now be the case. Note that it actually did *not* work that way before (see the error message from shellcheck).

ilausuch (approved):

PR #3924, opened 2021-06-01
Fix content negotiation with .txt file extension for module source

kraih:
A small regression caused by the same breaking change in Mojolicious that broke all the other routes before.

Progress: https://progress.opensuse.org/issues/93345

### DIFF
--- a/lib/OpenQA/WebAPI.pm
+++ b/lib/OpenQA/WebAPI.pm
@@ -191,7 +191,7 @@ sub startup ($self) {
     my $step_auth = $test_auth->any('/modules/:moduleid/steps/<stepid:step>');
     $step_r->get('/view')->to(action => 'view');
     $step_r->get('/edit')->name('edit_step')->to(action => 'edit');
-    $step_r->get('/src')->name('src_step')->to(action => 'src');
+    $step_r->get('/src', [format => ['txt']])->name('src_step')->to(action => 'src', format => undef);
     $step_auth->post('/')->name('save_needle_ajax')->to('step#save_needle_ajax');
     $step_r->get('/')->name('step')->to(action => 'view');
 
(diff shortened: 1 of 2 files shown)


Martchus:
Will it still work with older Mojolicious version? As we decided to merge https://github.com/os-autoinst/openQA/pull/3898 it would make sense not to destroy the effort again (at least not so soon).

kraih:
Well, considering we don't actually test with older versions of Mojolicious, it's not really supported anyway.

Martchus:
Than please update the version in `dependencies.yaml`. I must also say that a workflow which allows one to deploy the latest openQA but not Mojolicious seems quite flawed and we shouldn't invest effort into making that work.

kraih:
@Martchus As requested, but in a separate PR. #3927

okurz (approved):

AdamWill:
> Than please update the version in `dependencies.yaml`. I must also say that a workflow which allows one to deploy the latest openQA but not Mojolicious seems quite flawed and we shouldn't invest effort into making that work.

I don't really follow the rationale. Mojolicious is a generic web framework, it is not tied to openQA. The main reason I can't just update Mojolicious willy-nilly in Fedora is *because other things use it*. If only openQA used it, there would be no problem in me updating it whenever I liked, but because other things use it and believe in concepts like "major version upgrades are a big deal", I can't.

AdamWill:
Also, no-one actually answered the question. Would this work with older Mojo, or not?

Martchus:
I didn't suggest to update Mojolicious for every Fedora user - we also don't update Mojolicious for every openSUSE Leap user. However, we can update Mojolicious for *our* deployments (using openSUSE Leap and a custom repository which is updated and built automatically on OBS).

AdamWill:
Well, we already discussed this in the other ticket. I *could* do that, but setting it up is a chunk of work I could live without, and technically means I am also on the hook for backporting security fixes to whatever stale snapshot of openQA Fedora 34 winds up with. It was rather easier to just tweak the routing so it still worked with 8. I can do the same for this change if it's actually necessary.

kraih:
No, this does not work with older versions of Mojolicious.

PR #3926, opened 2021-06-01
parser: update IPA parser to support latest img-proof changes

asmorodskyi:
Since img_proof 7.1.0 output file with results change its format. This patch updating code to support
new format. Changes are not backward-compatible which make sense because we always using latest version
of img_proof in all our runs.


Martchus:
Just for the record, the related ticket is https://progress.opensuse.org/issues/93324.

I'm currently working on https://progress.opensuse.org/issues/93342 to improve the error handling generally. Hence I've checked which tests use `parse_extra_log` and there are more occurrences which use the IPA parser. Are you aware of these occurrences? If not, please grep within the test distribution and check whether these would break when we merge this non-backwards compatible change. Maybe it is also just easier to make this change backwards compatible and not having to care where it is used.

lib/OpenQA/Parser/Format/IPA.pm
```diff
@@ -37,7 +37,8 @@ sub parse {
 
     foreach my $res (@{$decoded_json->{tests}}) {
         my $result = {};
-        my $t_name = $res->{name};
+        die "Unexpected file format" unless exists $res->{nodeid};
+        my $t_name = $res->{nodeid};
```
Martchus, line 41: I guess making this backwards compatible shouldn't be a big deal:

```suggestion
        my $t_name = $res->{nodeid} // $res->{name};
        die 'IPA result misses test name / node ID' unless $t_name;
```

And then also just use `$t_name` later on.

I also made the error message more specific.

asmorodskyi: speaking on behalf of team testing Public Cloud in openQA who is **only** consumer of this parser I must state that there is no way we will use older versions of img-proof because this would mean for example lack of support for SLE 15 SP3 images which was introduced in img-proof version `7.2.0` . So there is no good reason to keep it backward-compatible

asmorodskyi: but I will take part of your suggestion ( improved error message and die after assigning variable  ) but I will do that after I will get final build result to see if there some other things needs to be improved ...

Martchus: When grepping I also found tests like `tests/ha/hawk_gui.pm` and `tests/sles4sap/wmp_basic_test.pm` which seem to be unrelated to public could.

asmorodskyi: > When grepping I also found tests like `tests/ha/hawk_gui.pm` and `tests/sles4sap/wmp_basic_test.pm` which seem to be unrelated to public could.

well I am not sure what is proper way to handle that . If someone wrote a parser which explicitly stating it's purpose https://github.com/os-autoinst/openQA/blob/master/lib/OpenQA/Parser/Format/IPA.pm#L19 and someone else using it for different purpose should we really keep supporting both approaches in same parser or maybe we need to migrate `tests/ha/hawk_gui.pm` to use more appropriate parser ? 

@alvarocarvajald  FYI

alvarocarvajald: Hello @asmorodskyi @Martchus. We are using this in 4 tests AFAIA: hawk, WMP basic HA tests, cluster preflight check and pacemaker_cts_cluster_exerciser. All these tests either rely on external tests for which we can modify the code to make the output compatible with this new IPA format or we are generating the JSON file via code in os-autoinst-distri-opensuse modules directly. So either choice is fine by me, if the choice is made to drop backward compatibility we can adapt. Just keep me in the loop. Thanks.

Martchus: Considering that keeping backwards compatibility is so simple in this case I'm wondering why we wouldn't do it. Adapting other code seems like more effort.

asmorodskyi: > Considering that keeping backwards compatibility is so simple in this case I'm wondering why we wouldn't do it. Adapting other code seems like more effort.

ok , I will keep it backward compatible

Martchus (approved):

Martchus:
(I approved but some automatic tests would be nice of course.)

asmorodskyi:
In my opinion PR is ready to be merged . Test data now holds both cases ( nodeid and name ) and all tests are passing

alvarocarvajald (approved):
LGTM

Martchus (approved):

pdostal (approved):

kraih (approved):

PR #3927, opened 2021-06-01
Increase Mojolicious version requirement

kraih:
Thanks to #3898 there are workarounds for old versions of Mojolicious (8.55) in openQA. We do not actually test with versions before Mojolicious 9.11 though, so it's not really supported. And it's starting to get in the way of new fixes, like #3924. Changing the listed dependency would make it clear what we officially support (even if there are still workarounds for older versions in the code).


Martchus (approved):

okurz (approved):
@AdamWill

AdamWill:
I don't really care if the statements change, because I can just patch/ignore them. It's more significant if code that really only works with 9 is added.

okurz:
Because we have different opinions or person, learn something new every day and also are not perfect. I would prefer if we test with older package versions and then adjust our version dependencies as well. Here we just adjust to what is currently tested and supported and maybe even strictly needed, I don't know that. But I will still appreciate and approve any changes that bring in backward compatibility

kraih:
Actually, you really don't want to be using a version before 9.19. There was a very serious security fix in that release (sessions were broken).

AdamWill:
> Actually, you really don't want to be using a version before 9.19. There was a very serious security fix in that release (sessions were broken).

well, I mean, it would arguably be responsible in that case to backport the fix to 8.x and do a point release. Rather than requiring anything that has not yet been ported to 9.x to be ported in order to fix the problem.

PR #3928, opened 2021-06-02
Show results for multiple groups in /test/overview

ilausuch:
The strategy is:
- Modify the query to include the jobs for all the groups_ids
- Modify last_jobs to include the group_id as part of the hash

https://progress.opensuse.org/issues/91650

### DIFF
--- a/lib/OpenQA/Schema/ResultSet/Jobs.pm
+++ b/lib/OpenQA/Schema/ResultSet/Jobs.pm
@@ -1,4 +1,4 @@
-# Copyright (C) 2014-2020 SUSE LLC
+# Copyright (C) 2014-2021 SUSE LLC
 #
 # This program is free software; you can redistribute it and/or modify
 # it under the terms of the GNU General Public License as published by
@@ -98,12 +98,13 @@ the return array.
 
 =cut
 sub latest_jobs {
-    my ($self, $until) = @_;
+    my ($self, $until, $hash_group_id) = @_;
 
     my @jobs = $self->search($until ? {t_created => {'<=' => $until}} : undef, {order_by => ['me.id DESC']});
     my @latest;
     my %seen;
     foreach my $job (@jobs) {
+        my $groupid = $job->group_id;
         my $test    = $job->TEST;
         my $distri  = $job->DISTRI;
         my $version = $job->VERSION;
@@ -112,6 +113,7 @@ sub latest_jobs {
         my $arch    = $job->ARCH;
         my $machine = $job->MACHINE // '';
         my $key     = "$distri-$version-$build-$test-$flavor-$arch-$machine";
+        $key = "$groupid-$key" if ($hash_group_id);
         next if $seen{$key}++;
         push(@latest, $job);
     }
--- a/lib/OpenQA/WebAPI/Controller/Test.pm
+++ b/lib/OpenQA/WebAPI/Controller/Test.pm
@@ -14,7 +14,7 @@
 # with this program; if not, see <http://www.gnu.org/licenses/>.
 
 package OpenQA::WebAPI::Controller::Test;
-use Mojo::Base 'Mojolicious::Controller';
+use Mojo::Base 'Mojolicious::Controller', -signatures;
 
 use OpenQA::Utils;
 use OpenQA::Jobs::Constants;
@@ -642,6 +642,30 @@ sub _add_distri_and_version_to_summary {
     }
 }
 
+sub latest_jobs ($self, $search_args, $until) {
+    my @latest_jobs;
+
+    my $group_ids = $search_args->{groupids};
+    if (defined $group_ids && @$group_ids > 1) {
+        my @composed_search_arch = map {
+            my $local_search_args = {
+                group_id => $_,
+                build    => $self->schema->resultset('Jobs')->latest_build((groupid => $_))};
+            $local_search_args->{distri}  = $search_args->{distri}  if (defined($search_args->{distri}));
+            $local_search_args->{version} = $search_args->{version} if (defined($search_args->{version}));
+            $local_search_args;
+        } @$group_ids;
+
+        @latest_jobs
+          = $self->schema->resultset('Jobs')->search(\@composed_search_arch)->latest_jobs($until, 1);
+    }
+    else {
+        @latest_jobs = $self->schema->resultset('Jobs')->complex_query(%$search_args)->latest_jobs($until);
+    }
+
+    return \@latest_jobs;
+}
+
 # A generic query page showing test results in a configurable matrix
 sub overview {
     my ($self) = @_;
@@ -658,8 +682,9 @@ sub overview {
         groups  => $groups,
         until   => $until,
     );
-    my @latest_jobs = $self->schema->resultset('Jobs')->complex_query(%$search_args)->latest_jobs($until);
-    ($stash{archs}, $stash{results}, $stash{aggregated}) = $self->prepare_job_results(\@latest_jobs);
+
+    my $latest_jobs = $self->latest_jobs($search_args, $until);
+    ($stash{archs}, $stash{results}, $stash{aggregated}) = $self->prepare_job_results($latest_jobs);
 
     # determine distri/version from job results if not explicitely specified via search args
     my @distris = keys %{$stash{results}};
--- a/t/10-tests_overview.t
+++ b/t/10-tests_overview.t
@@ -1,4 +1,4 @@
-# Copyright (C) 2014-2020 SUSE LLC
+# Copyright (C) 2014-2021 SUSE LLC
 #
 # This program is free software; you can redistribute it and/or modify
 # it under the terms of the GNU General Public License as published by
@@ -271,6 +271,13 @@ subtest 'not complete results generally accounted as "Incomplete"' => sub {
     like(get_summary, qr/Passed: 0 Incomplete: 2 Failed: 0 Scheduled: 2 Running: 2 None: 1/i);
     $schema->txn_rollback;
 };
+
+#
+# Test multiple groupid
+#
+$t->get_ok('/tests/overview?groupid=1001&groupid=1002')->status_is(200);
+$t->element_exists('#res_DVD_x86_64_kde');        # Included groupid 1001
+$t->element_exists('#flavor_NET_arch_x86_64');    # Included groupid 1002
 
 #
 # Test filter form


lib/OpenQA/WebAPI/Controller/Test.pm
```diff
@@ -658,7 +658,31 @@ sub overview {
...
+            my $local_search_args;
+            $local_search_args->{group_id} = $groupid;
+            $local_search_args->{build} = $self->schema->resultset('Jobs')->latest_build(%$groupid_search);
+            $local_search_args->{distri} = $search_args->{distri} if (defined($search_args->{distri}));
+            $local_search_args->{version} = $search_args->{version} if (defined($search_args->{version}));
+
+            push(@composed_search_arch, $local_search_args);
+        };
+
+        @latest_jobs = $self->schema->resultset('Jobs')->search(\@composed_search_arch)->latest_jobs_multiple_groups($until);
```
Martchus, line 680: Not using `compex_query` here means we drop all the ~~complex~~fancy features it provides. Maybe that's good enough but we should check which use cases are broken by this simplification.

Martchus (changes requested):

lib/OpenQA/WebAPI/Controller/Test.pm
```diff
@@ -658,7 +658,31 @@ sub overview {
         groups  => $groups,
         until   => $until,
     );
-    my @latest_jobs = $self->schema->resultset('Jobs')->complex_query(%$search_args)->latest_jobs($until);
+
+    my @latest_jobs = [];
```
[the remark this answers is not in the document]
ilausuch: Done

lib/OpenQA/WebAPI/Controller/Test.pm
```diff
@@ -658,7 +658,31 @@ sub overview {
...
+
+    if (@{$search_args->{groupids}} > 1) {
+        my @composed_search_arch;
```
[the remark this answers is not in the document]
ilausuch: Done

ilausuch:
I pushed a commit with a test for this feature

lib/OpenQA/WebAPI/Controller/Test.pm
ilausuch: Remains the investigation of this part

Martchus (changes requested):
I've mentioned that because `complex_query` is not used in the new code path the features it provides are not available when specifying multiple group IDs. I did some tests locally and it seems like the only feature which is actually used from `complex_query` (and not covered by the new code path) is filtering by job modules. So the query parameters `modules` and `modules_result` won't have any effect anymore when used like this: `groupid=1&groupid=100&modules=bootloader_start` - Maybe we can live with that (if the limitation is somehow apparent)?

---

It would also be useful if you could test this with production data (e.g. from o3) and then check whether queries like the ones mentioned under "suggestions" in the related ticket now behave as expected.

ilausuch:
@Martchus Thanks for your advices. This new commit has the small changes you commented

ilausuch:
I added to https://progress.opensuse.org/issues/91650 a comment with the demonstration of this PR with real data

lib/OpenQA/Schema/ResultSet/Jobs.pm
```diff
@@ -118,6 +118,28 @@ sub latest_jobs {
...
+        my $build   = $job->BUILD;
+        my $flavor  = $job->FLAVOR;
+        my $arch    = $job->ARCH;
+        my $machine = $job->MACHINE // '';
+        my $key     = "$groupid-$distri-$version-$build-$test-$flavor-$arch-$machine";
+        next if $seen{$key}++;
+        push(@latest, $job);
+    }
+    return @latest;
+}
```
okurz, line 141: that's good enough for a proof of concept but I would like this duplication to be removed before being able to approve

lib/OpenQA/WebAPI/Controller/Test.pm
```diff
@@ -658,7 +658,26 @@ sub overview {
...
+            $local_search_args->{version} = $search_args->{version} if (defined($search_args->{version}));
+            $local_search_args;
+        } @$group_ids;
+
+        @latest_jobs
+          = $self->schema->resultset('Jobs')->search(\@composed_search_arch)->latest_jobs_multiple_groups($until);
+    }
+    else {
+        @latest_jobs = $self->schema->resultset('Jobs')->complex_query(%$search_args)->latest_jobs($until);
+    }
```
okurz, line 679: Instead of making the sub "overview" which is already big even bigger please extract some functions

t/ui/10-tests_overview.t
```diff
@@ -20,7 +20,7 @@ use lib "$FindBin::Bin/../lib", "$FindBin::Bin/../../external/os-autoinst-common
 use Date::Format 'time2str';
 use Test::Mojo;
 use Test::Warnings ':report_warnings';
-use OpenQA::Test::TimeLimit '20';
+use OpenQA::Test::TimeLimit '200';
```
okurz, line 23: if you just need that for local testing you can also set the environment variable `OPENQA_TEST_TIMEOUT_DISABLE=1`

okurz (changes requested):
@Martchus wrote:
> I've mentioned that because `complex_query` is not used in the new code path the features it provides are not available when specifying multiple group IDs. I did some tests locally and it seems like the only feature which is actually used from `complex_query` (and not covered by the new code path) is filtering by job modules. So the query parameters `modules` and `modules_result` won't have any effect anymore when used like this: `groupid=1&groupid=100&modules=bootloader_start` - Maybe we can live with that (if the limitation is somehow apparent)?

I think we can't live with that because we want teams to show results for multiple groups *plus* filter by job modules. This is a current and realistic use case from multiple SUSE QE teams that want to review for *their* job groups in multiple job groups

okurz:
I suggest you also create screenshots which you can then directly use for documentation purposes as well

lib/OpenQA/WebAPI/Controller/Test.pm
okurz: maybe https://github.com/os-autoinst/openQA/pull/3918 can help to make complex_query a little bit simpler

lib/OpenQA/Schema/ResultSet/Jobs.pm
```diff
@@ -112,6 +113,7 @@ sub latest_jobs {
         my $arch    = $job->ARCH;
         my $machine = $job->MACHINE // '';
         my $key     = "$distri-$version-$build-$test-$flavor-$arch-$machine";
+        $key = "$groupid-" . $key if ($hash_group_id);
```
Martchus, line 116:
```suggestion
        $key = "$groupid-$key" if $hash_group_id;
```

lib/OpenQA/WebAPI/Controller/Test.pm
```diff
@@ -642,6 +642,31 @@ sub _add_distri_and_version_to_summary {
...
+    my @latest_jobs;
+
+    my $group_ids = $search_args->{groupids};
+    if (defined $group_ids && @$group_ids > 1) {
+        my @composed_search_arch = map {
+            my $local_search_args = {
+                group_id => $_,
+                build    => $self->schema->resultset('Jobs')->latest_build((groupid => $_))};
+            $local_search_args->{distri}  = $search_args->{distri}  if (defined($search_args->{distri}));
+            $local_search_args->{version} = $search_args->{version} if (defined($search_args->{version}));
```
Martchus, line 656: I'd save the search args within variables (e.g. `my ($distri, $version) = ($search_args->{distri}, $search_args->{version});`) instead of accessing these twice within each map iteration.

t/ui/10-tests_overview.t
```diff
@@ -382,6 +382,12 @@ subtest "job dependencies displayed on 'Test result overview' page" => sub {
       'parent job was highlighted correctly';
 };
 
+subtest "multiple groups" => sub {
+    $driver->get('/tests/overview?groupid=1001&groupid=1002');
```
Martchus, line 386: I don't think there's much JavaScript involved here so it would better to add this test to `t/10-tests_overview.t` and just check whether the DOM contains these elements.

lib/OpenQA/WebAPI/Controller/Test.pm
```diff
@@ -642,6 +642,31 @@ sub _add_distri_and_version_to_summary {
...
+        } @$group_ids;
+
+        @latest_jobs
+          = $self->schema->resultset('Jobs')->search(\@composed_search_arch)->latest_jobs($until, 1);
+    }
+    else {
+        @latest_jobs = $self->schema->resultset('Jobs')->complex_query(%$search_args)->latest_jobs($until);
+    }
+
+    @latest_jobs;
```
Martchus, line 667: Although it is not required in Perl I'd prefer to add a `return` here. And more importantly: Perl will do a copy of the array elements here (and I believe copy elision isn't really a thing in Perl) so please return a reference here (`return \@latest_jobs`). The calling function needs to expect a scalar then (and can just pass it as-is).

lib/OpenQA/WebAPI/Controller/Test.pm
```diff
@@ -642,6 +642,31 @@ sub _add_distri_and_version_to_summary {
     }
 }
 
+sub latest_jobs {
+    my ($self, $search_args, $until) = @_;
```
Martchus, line 646: When adding new functions or changing arguments it would be nice to make use of signatures.

Martchus (changes requested):

lib/OpenQA/WebAPI/Controller/Test.pm
perlpunk: OTOH it's consistent with the `latest_jobs` method from the schema.
Maybe that one should have been returning an array ref in the first place.

lib/OpenQA/WebAPI/Controller/Test.pm
okurz: I disagree. I don't see the need to copy hash values into variables. That's of course different if you save return values from functions which can be considered more costly

Martchus: I don't think Perl copies Strings, it'll just be a reference internally. This should be faster than the hash lookup. But I also like using variables from the point of DRY as it avoids repeating the key.

okurz:
typo in git commit message: s/stragegy/strategy/

lib/OpenQA/WebAPI/Controller/Test.pm
ilausuch: I don't understand very well I should put that in variables. I need to build the hash to call the search method

Martchus: I mean the `$search_args` this function takes, **not** the hash this function creates (to call the search method).

Martchus: And btw, strings are COW in Perl, indeed: https://perldoc.perl.org/5.20.0/perldelta#Performance-Enhancements

But all of this won't really make much of a difference from the runtime performance so you can also keep your code if you like.

lib/OpenQA/WebAPI/Controller/Test.pm
```diff
@@ -658,8 +682,9 @@ sub overview {
         groups  => $groups,
         until   => $until,
     );
-    my @latest_jobs = $self->schema->resultset('Jobs')->complex_query(%$search_args)->latest_jobs($until);
-    ($stash{archs}, $stash{results}, $stash{aggregated}) = $self->prepare_job_results(\@latest_jobs);
+
+    my @latest_jobs = $self->latest_jobs($search_args, $until);
```
Martchus, line 686:
```suggestion
    my $latest_jobs = $self->latest_jobs($search_args, $until);
```

lib/OpenQA/WebAPI/Controller/Test.pm
```diff
@@ -658,8 +682,9 @@ sub overview {
...
+    ($stash{archs}, $stash{results}, $stash{aggregated}) = $self->prepare_job_results(@latest_jobs);
```
Martchus, line 687:
```suggestion
    ($stash{archs}, $stash{results}, $stash{aggregated}) = $self->prepare_job_results($latest_jobs);
```

Martchus (changes requested):
I believe the way you're passing the jobs around is still wrong, see the inline comments. But of course also take care of the comments from @okurz (which might require some more restructuring anyways).

lib/OpenQA/WebAPI/Controller/Test.pm
ilausuch: Yes! you are right. now is a reference

lib/OpenQA/WebAPI/Controller/Test.pm
ilausuch: done

kalikiana:
Please remember to make it a draft and check all open comments. I see one open thread currently. Otherwise it's not clear if you're waiting for more reviews.

lib/OpenQA/WebAPI/Controller/Test.pm
kalikiana: @ilausuch Is this still open? Or did you address this in your most recent changes?

ilausuch: Waiting for #3918 to test if complex_query could help to do it better

okurz: @ilausuch #3918 is ready to be reviewed

kalikiana:
This can be rebased now that #3918 got merged.

ilausuch:
I close this PR because #3918 solves the problem

Martchus:
When I understand it correctly, #3918 only removes API features and "simplifies" the code. It doesn't solve https://progress.opensuse.org/issues/91650.

ilausuch:
Yes, this is certain

perlpunk:
this can also be closed I guess?

okurz:
yes, obsoleted by https://github.com/os-autoinst/openQA/pull/3973

PR #3934, opened 2021-06-08
Run 10-jobs.t really fast

kraih:
I've added a `OpenQA::Test::Utils::perform_minion_jobs_in_foreground` function for now, but might also add this as an upstream feature to Minion, since there's enough reasonable use cases for it now.

Before:
```
[15:07:34] t/10-jobs.t ........... ok   160051 ms ( 0.06 usr  0.01 sys + 154.81 cusr  4.02 csys = 158.90 CPU)
```

After:
```
[15:39:48] t/10-jobs.t ........... ok    19263 ms ( 0.07 usr  0.01 sys + 18.16 cusr  0.66 csys = 18.90 CPU)
```

Progress: https://progress.opensuse.org/issues/93086

### DIFF
--- a/t/10-jobs.t
+++ b/t/10-jobs.t
@@ -50,8 +50,6 @@ my $t      = Test::Mojo->new('OpenQA::WebAPI');
 my $jobs   = $t->app->schema->resultset("Jobs");
 my $users  = $t->app->schema->resultset("Users");
 
-collect_coverage_of_gru_jobs($t->app);
-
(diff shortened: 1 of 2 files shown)


okurz (approved):
Awesome achievement! And code coverage is ok so all good I think :)

Martchus (approved):
Nice, this looks better than my hack in `t/42-df-based-cleanup.t`. Maybe we want to update this test as well.

PR #3938, opened 2021-06-09
WIP -- codecov.yml: Introduce specific coverage thresholds

okurz:
With path specific coverage targets we can introduce the hard
requirement for 100% coverage within the paths where it is feasible. It
makes sense to have 100% statement coverage within test code itself to
prevent relying on tests that are actually not running.

See https://docs.codecov.com/docs/commit-status#branches

### DIFF
--- a/codecov.yml
+++ b/codecov.yml
@@ -7,9 +7,6 @@ comment:
 fixes:
   - "/home/circleci/openQA::"
 coverage:
-  range: 60..95
-  round: down
-  precision: 2
   status:
     changes: false
     project:
--- a/codecov.yml
+++ b/codecov.yml
@@ -10,9 +10,15 @@ coverage:
   status:
     changes: false
     project:
-      default:
-        target: 97.2
-        threshold: 0.1
+      default: false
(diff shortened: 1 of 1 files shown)


okurz:
So the "codecov" check passes but there is no proper report from codecov. Help wanted to understand how to get codecov to use path-specific coverage targets :)

perlpunk:
Can you link to documentation, or where you found that codecov supports that?

okurz:
> Can you link to documentation, or where you found that codecov supports that?

updated, see https://docs.codecov.com/docs/commit-status#branches

I also changed it a bit more, basically trying to just "change something" to see if this makes it work

okurz:
So the codecov report is updated but I don't see the scopes that I defined additionally, e.g. "tests" or "app". Maybe there is something not supported by the codecovbash module we use

perlpunk:
> Maybe there is something not supported by the codecovbash module we use

codecovbash does nothing more than upload a list of numbers per line for each filename.
Codecov then calculates the coverage.

PR #3939, opened 2021-06-09
Deployment handler: Log error message

perlpunk:


lib/OpenQA/Schema.pm
```diff
@@ -131,6 +131,13 @@ sub _try_deploy_db {
         $version = $dh->version_storage->database_version;
     }
     catch {
+        # If the table does not exist, we want to deploy, and the error
+        # is expected. If we get other errors like "Permission denied" in case
+        # the database is not readable by the current user, we print the
+        # error message
+        unless (m/relation "dbix_class_deploymenthandler_versions" does not exist/) {
+            warn "Error when trying to get the database version: $_";
+        }
```
okurz, line 140:
```suggestion
        warn "Error when trying to get the database version: $_"
            unless m/relation "dbix_class_deploymenthandler_versions" does not exist/;
```

ok?

lib/OpenQA/Schema.pm
```diff
@@ -131,6 +131,13 @@ sub _try_deploy_db {
...
         $dh->install;
         $schema->create_system_user;    # create system user right away
```
okurz, line 142: I am not sure if it makes sense to have this code *and* the warning. Can you explain that in more detail?

perlpunk: You mean it should be a `die` instead?

I am a bit afraid what happens if the error message for the non-existant relation slightly changes in the future.
Then the deploy would always fail.
If it is only a warning, it will be noticed, but it will not make a deployment fail.

okurz: well but warnings are ignored and tend to stay for near-forever and confuse users in case they do not actually need to do something about it, isn't it?

Martchus: It is annoying that we can not distinguish different types of errors here safely. It seems needless to abort here if the error message changes (e.g. due to a differently configured locale or an update of a dependency). I suppose our tests would fail in case of a warning so it won't be unnoticed.

okurz: Right, that's maybe enough

Martchus (approved):

Martchus:
I've just approved but maybe you want to mark the line as uncoverable. (The style suggested by @okurz maybe tricks the coverage analysis so you don't need it then.)

okurz (approved):

PR #3940, opened 2021-06-09
Adapt UI tests to chromedriver changes

Martchus:
Version 91.0.4472.77 of `chromedriver` changed its behavior so
`Selenium::Remote::WebElement` behaves different as well. This change
adapts our tests to the following changes (in a backwards compatible way):

* `Selenium::Remote::WebElement::get_attribute()` now returns `undef`
  instead of an empty string if the attribute is not present at all.
* `Selenium::Remote::WebElement::get_value()` and `is_selected()` sometimes
  no longer return a value. Likely it depends on how these form properties
  are populated. Possible an unwanted side-effect of the `get_attribute()`
  change.

See https://progress.opensuse.org/issues/93453#note-11

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -86,8 +86,6 @@ aliases:
   - &make_test
     name: "Running unit tests"
     command: |
-      # downgrade chromedriver from problematic version to last good version, see https://progress.opensuse.org/issues/93453
-      rpm -q chromedriver | grep -v 91.0.4472.77 || sudo zypper -n in --oldpackage chromedriver-90.0.4430.212-lp152.2.95.1 chromium-90.0.4430.212
       export COVERAGE=1
       export COVERDB_SUFFIX="_${CIRCLE_JOB}"
       make test-$CIRCLE_JOB
--- a/assets/assetpack.def
+++ b/assets/assetpack.def
@@ -161,6 +161,7 @@
 < javascripts/test_result.js
 < javascripts/needlediff.js
 < javascripts/running.js
+< javascripts/disable_status_updates.js [mode==test]
 
 ! job_next_previous.js
 < javascripts/job_next_previous.js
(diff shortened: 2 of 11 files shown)


okurz (approved):

Martchus:
I've also noticed a problem within `t/ui/25-developer_mode.t` which makes the test definitely unstable. (It can be easily reproduced by putting sleeps at the right places.) Hopefully this is the only instability within the test. I'll create another PR later to test this (separately).

okurz (changes requested):
coverage decrease in one test file?

Martchus:
The coverage report looks wrong, as if merging coverage data from many sub processes didn't work. I also cross-checked that the additional mocking for `t/ui/25-developer_mode.t` really only applies to that test.

okurz:
I don't want to approve the PR as long as the coverage mentions that drop from 100% in that file though. The one in t/lib/OpenQA/Test/Utils.pm is the only one we know and track in a currently open ticket https://progress.opensuse.org/issues/80274

perlpunk (approved):

perlpunk:
Just for the record, I tested with a simple 1-element HTML form, and `get_value()` always only returns the original `value` attribute, but not the value that is entered later.

Martchus:
I know that there's already a ticket. So I've just tried to re-run the tests here but it didn't help. But the problem with the report in this PR looks strange and is likely not even related to flaky tests because if the "uncovered" code actually wasn't executed the rest of the tests which are still executed couldn't possibly pass. All the lost coverage is within code which runs in forks so likely collecting coverage from the forks didn't work at some point and now the data is somehow stuck. I can create a new PR if it is necessary.

Martchus:
@perlpunk I guess it might also work if the attribute value is directly populated via JavaScript's DOM-API but it doesn't work if the value is changed somehow else.

okurz:
> […] I can create a new PR if it is necessary.

As I have not seen that specific coverage drop anywhere else I am asking you kindly to find a way within here

Martchus:
Let's see whether a different PR is also affected.

PR #3945, opened 2021-06-11
Adapt UI tests to chromedriver changes, make devel mode test stable

Martchus:
Version 91.0.4472.77 of `chromedriver` changed its behavior so
`Selenium::Remote::WebElement` behaves different as well. This change
adapts our tests to the following changes (in a backwards compatible way):

* `Selenium::Remote::WebElement::get_attribute()` now returns `undef`
  instead of an empty string if the attribute is not present at all.
* `Selenium::Remote::WebElement::get_value()` and `is_selected()` sometimes
  no longer return a value. Likely it depends on how these form properties
  are populated. Possible an unwanted side-effect of the `get_attribute()`
  change.

See https://progress.opensuse.org/issues/93453#note-11 and particular commit messages

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -86,8 +86,6 @@ aliases:
   - &make_test
     name: "Running unit tests"
     command: |
-      # downgrade chromedriver from problematic version to last good version, see https://progress.opensuse.org/issues/93453
-      rpm -q chromedriver | grep -v 91.0.4472.77 || sudo zypper -n in --oldpackage chromedriver-90.0.4430.212-lp152.2.95.1 chromium-90.0.4430.212
       export COVERAGE=1
       export COVERDB_SUFFIX="_${CIRCLE_JOB}"
       make test-$CIRCLE_JOB
--- a/.circleci/unstable_tests.txt
+++ b/.circleci/unstable_tests.txt
@@ -2,4 +2,3 @@ t/25-cache-service.t
 t/ui/01-list.t
 t/ui/26-jobs_restart.t
 t/ui/13-admin.t
-t/ui/25-developer_mode.t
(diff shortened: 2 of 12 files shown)


okurz (changes requested):
Also here I see the coverage drop in

t/ui/13-admin.t | 98.89% &lt;100.00%&gt; (-0.01%)

as problematic

Martchus:
I don't see any unrelated coverage changes. I reduced the number of lines in that file by avoiding the use of extra variable declarations/assignments in some cases. Of course that reduced the percentage of covered lines within the test file as these lines were covered but it is not problematic.

okurz (approved):
hm, ok. I don't know how to read these numbers then. codecov says the first number is "absolute" coverage (see https://docs.codecov.io/docs/codecov-delta#usage) so when that is below 100% I assume that some lines within that file are not covered. The second number is "relative" coverage change so all lines that you changed are covered. But how can it say then that not 100% absolute are covered anymore? Looks like maybe our reference for that comparison is not correct. Well, let's see then.

Martchus:
> But how can it say then that not 100% absolute are covered anymore?

Where does it say that? For the file with the coverage drop it says: `98.89% <100.00%> (-0.01%)`

I assume this means the new coverage is 98.89 % (drop of 0.01 %) but 100 % of the lines I changed within that file have been covered. Considering I removed  covered statements this makes sense.

okurz:
> > But how can it say then that not 100% absolute are covered anymore?
> 
> Where does it say that? For the file with the coverage drop it says: `98.89% <100.00%> (-0.01%)`

Yes ,that line. It says 98.89%. I somehow assumed that the file should have had 100% statement coverage in before. Obviously it had not but I could fix that now with
https://github.com/os-autoinst/openQA/pull/3951
:)

kalikiana (approved):
I won't claim to get everything, but you can consider this a sanity check. Generally looks sensible to me. And I like the added diagnostics in case of failures, assuming they work as I would expect.

PR #3946, opened 2021-06-11
dist: Also save update-cache.sh from OBS

okurz:
After merge we must update https://build.opensuse.org/package/view_file/devel:openQA/openQA/_service?expand=1 to checkout that file as well.

### DIFF
--- a/dist/rpm/update-cache.sh
+++ b/dist/rpm/update-cache.sh
@@ -1,0 +1,21 @@
+#! /bin/sh
+
+set -e
+
+export LC_ALL='en_US.UTF-8'
+export LANG='en_US.UTF-8'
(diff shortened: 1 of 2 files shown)


Martchus (approved):

Martchus (changes requested):
The documentation should be updated as well: https://github.com/os-autoinst/openQA/blob/master/docs/Contributing.asciidoc#update-asset-cache

The script still needs to run within the osc checkout but is not located there anymore (which is currently implied).

okurz:
Updated the documentation

Martchus (approved):

kalikiana (approved):

PR #3947, opened 2021-06-11
Correct volume labels to allow running on SELinux enabled systems

dcermak:
The containers will fail to launch on systems where SELinux is in enforcing mode
as the shared volumes have the wrong labels. Thus we must label all shared
volumens with :z (even /conf as that one is shared by webui replicas and
:Z *will* cause this to fail!).

### DIFF
--- a/container/webui/docker-compose.yaml
+++ b/container/webui/docker-compose.yaml
@@ -4,13 +4,13 @@ services:
   scheduler:
     image: openqa_webui
     volumes: &volumes
-      - ./workdir/data/factory/iso:/data/factory/iso:rw
-      - ./workdir/data/factory/hdd:/data/factory/hdd:rw
-      - ./workdir/data/factory/other:/data/factory/other:rw
-      - ./workdir/data/factory/tmp:/data/factory/tmp:rw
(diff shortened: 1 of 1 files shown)


Martchus (approved):
According to the explanation in https://stackoverflow.com/questions/35218194/what-is-z-flag-in-docker-containers-volumes-from-option I suppose this makes sense.

kalikiana:
> According to the explanation in https://stackoverflow.com/questions/35218194/what-is-z-flag-in-docker-containers-volumes-from-option I suppose this makes sense.

`Two suffixes :z or :Z can be added to the volume mount. These suffixes tell Docker to relabel file objects on the shared volumes. The 'z' option tells Docker that the volume content will be shared between containers. Docker will label the content with a shared content label. Shared volumes labels allow all containers to read/write content. The 'Z' option tells Docker to label the content with a private unshared label.`

This is something I didn't know about. Good find!

Do we have any way to test this? How do we know this works?

dcermak:
> Do we have any way to test this? How do we know this works?

You try a `docker-compose up` on a system with SELinux set to enforcing. I'm afraid there is no other way how this can be tested.

okurz (approved):

PR #3953, opened 2021-06-12
t: Mark uncoverable statements in 01-test-utilities, 14-grutasks, 42-df-based-cleanup

okurz:


kalikiana (approved):
This looks a bit awkward. But I guess it's as good as it gets

Martchus (approved):
It looks awkward, indeed. I assume these ugly comments are the best we can do in Perl.

PR #3954, opened 2021-06-14
Mark openqa-client as deprecated

okurz:
openqa-client is replaced by openqa-cli within the openQA repo and
should be removed eventually.

### DIFF
--- a/script/client
+++ b/script/client
@@ -173,6 +173,9 @@ GetOptions(
     'asset-size-limit:i', 'with-thumbnails', 'accept=s',    'yaml-output',
 ) or usage(1);
 
+warn "WARNING: openqa-client is deprecated and planned to be removed in the future. Please use openqa-cli instead\n"
+  unless $ENV{OPENQA_CLIENT_DISABLE_DEPRECATION_WARNING};
+
 usage(0) if $options{help};
 usage(1) unless @ARGV;
 


script/client
```diff
@@ -173,6 +173,8 @@ GetOptions(
     'asset-size-limit:i', 'with-thumbnails', 'accept=s',    'yaml-output',
 ) or usage(1);
 
+warn 'openqa-client is deprecated and is planned to be removed in the near future. Please use openqa-cli instead';
```
Martchus, line 176:
```suggestion
warn "WARNING: openqa-client is deprecated and is planned to be removed in the near future. Please use openqa-cli instead\n";
```
to suppress the line number at the end (which makes it not really look like a user-facing output) and to raise the attention level.

Maybe it makes also sense to use `print STDERR ` instead.

kalikiana, line 176: I guess we dont have tests here. I suspect this warning will show up everywhere where  openqa-client is being used. Might be good practice to allow disabling it. Some env variable?

okurz: You want to disable the deprecation warning? I don't get it

okurz: Ok, changed the text and the newline, that makes sense. `warn` and `print STDERR` is likely ending up in just the same but I prefer the less characters of `warn` :)

Martchus (approved):

script/client
kalikiana: You can't simple stop using it, especially on behalf of all users. So make it easy to disable the warning in cases where it's still being used. That's generally the only way to have strict tests for bleeding edge software.

okurz: I don't understand. Can you provide examples? Maybe you remember that originally I asked to improve openqa-client and not come up with a rewrite that confuses users but I was overruled so this is what we have. So if anyone can't cope with the warning, how do you imagine they would be able to move to openqa-cli? Because that's what we need them to do

perlpunk: So far users didn't know that it's deprecated.
Letting people disable the warning quickly via env gives them time to properly change their code to use the new client.

kalikiana: > So far users didn't know that it's deprecated.
> Letting people disable the warning quickly via env gives them time to properly change their code to use the new client.

Exactly. And we have seen this from the other perspective, too with e.g. Mojolicious upstream. If somebody pulls in an update with the warning, they need to be able to prepare and adapt uses. "Deal with new warnings or port everything right away" is not a great flow.

okurz: ok. I see. Added an env variable `OPENQA_CLIENT_DISABLE_DEPRECATION_WARNING`. Good enough?

kalikiana (approved):

script/client
kalikiana: Excellent. Thank you!

PR #3958, opened 2021-06-15
Enhance/simplify jobs code4

okurz:


t/10-jobs.t
```diff
@@ -633,13 +633,7 @@ subtest 'job with skipped modules' => sub {
 sub job_is_linked {
     my ($job) = @_;
     $job->discard_changes;
-    my $comments = $job->comments;
```
Martchus, line 636: It would be better not to check every comment for a label and having to create DBIx objects for every comment like the current code does.

Martchus (changes requested):

t/10-jobs.t
perlpunk: Since it's in a test and we don't have thousands of jobs, it's probably ok

okurz: Martchus is still right and code in tests can act as examples for someone to use in the future. We can use the same approach as in 3ccb16bdb though

okurz:
test results:

```
[21:37:11] t/10-tests_overview.t ..................................... 99/? # Premature connection close
[21:37:11] t/10-tests_overview.t ..................................... 109/? 
#   Failed test 'GET /tests/99937/modules/kate/fails'
#   at t/10-tests_overview.t line 315.

#   Failed test 'correct failed needles'
#   at t/10-tests_overview.t line 315.
#          $got = undef
#     $expected = ARRAY(0x560a7a721730)
# Premature connection close
[21:37:11] t/10-tests_overview.t ..................................... 111/? 
#   Failed test 'GET /tests/99937/modules/zypper_up/fails'
#   at t/10-tests_overview.t line 316.

#   Failed test 'failed module: fallback to first step'
#   at t/10-tests_overview.t line 316.
#          got: undef
#     expected: '1'
# Looks like you failed 4 tests of 136.
                                                                              [21:37:11] t/10-tests_overview.t ..................................... Dubious, test returned 4 (wstat 1024, 0x400)
Failed 4/136 subtests
```

perlpunk:
What does "code4" in the subject mean?

okurz:
> What does "code4" in the subject mean?

sorry about that. This was auto-created from the branch name `enhance/simplify_jobs_code4` and the "4" means my fourth iteration of more or less the same approach. Should not really matter for reviewers

PR #3961, opened 2021-06-16
Fix repository used by openqa-bootstrap under Leap 15.3

Martchus:
* Use "https:" URL because the "obs:" one is not resolved correctly
* Tested using
  ```
  docker run -v "$PWD:/tmp/checkout" --rm -it registry.opensuse.org/opensuse/leap:15.3 sh -c 'bash /tmp/checkout/script/openqa-bootstrap'
  ```
* See https://progress.opensuse.org/issues/93609

### DIFF
--- a/script/openqa-bootstrap
+++ b/script/openqa-bootstrap
@@ -10,8 +10,10 @@ dbuser="${dbuser:="geekotest"}"
 # shellcheck disable=SC1091
 . /etc/os-release
 if [ "$NAME" = "openSUSE Leap" ]; then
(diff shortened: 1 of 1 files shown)


okurz:
the obs version looks nice though. Maybe that is a bug in Leap itself. Then please report it as such and reference your change as workaround for the bug

Martchus:
Yes, this looks like a bug in Leap or zypper. I've created a bug and referenced it: https://bugzilla.opensuse.org/show_bug.cgi?id=1187425

okurz (approved):

asdil12 (approved):

PR #3966, opened 2021-06-17
Fix repository used by openqa-bootstrap under Leap 15.3 for internal CA

Martchus:
Same as e4053f76ab364c738d2497c642e7880588ad22c4 but for the repository for
the internal CA. I haven't noticed it in the first place because the ping
dependency was missing which was only fixed by
254d23d0ec92545c28979a50bb6c92a7e7da5031.

See https://progress.opensuse.org/issues/93609

### DIFF
--- a/script/openqa-bootstrap
+++ b/script/openqa-bootstrap
@@ -46,7 +46,12 @@ if ping -c1 download.suse.de. && (! rpm -q ca-certificates-suse) ; then
     if ! zypper info ca-certificates-suse | grep -q ':' ; then
         # add suse ca repo if needed
         # use this way of adding the repo to be distro agnostic
-        zypper -n addrepo obs://SUSE:CA SUSE:CA
+        if [ "$NAME" = "openSUSE Leap" ]; then
+            # avoid using `obs://…` URL to workaround https://bugzilla.opensuse.org/show_bug.cgi?id=1187425
+            zypper -n addrepo "http://download.suse.de/ibs/SUSE:/CA/openSUSE_Leap_${VERSION}" 'SUSE:CA'
+        else
+            zypper -n addrepo obs://SUSE:CA SUSE:CA
+        fi
         sed -i -e 's#download.opensuse.org/repositories#download.suse.de/ibs#' /etc/zypp/repos.d/SUSE:CA.repo
         sed -i -e 's/https/http/' /etc/zypp/repos.d/SUSE:CA.repo
         zypper -n --gpg-auto-import-keys refresh


script/openqa-bootstrap
```diff
@@ -46,7 +46,12 @@ if ping -c1 download.suse.de. && (! rpm -q ca-certificates-suse) ; then
     if ! zypper info ca-certificates-suse | grep -q ':' ; then
         # add suse ca repo if needed
         # use this way of adding the repo to be distro agnostic
-        zypper -n addrepo obs://SUSE:CA SUSE:CA
+        if [ "$NAME" = "openSUSE Leap" ]; then
+            # avoid using `obs://…` URL to workaround https://bugzilla.opensuse.org/show_bug.cgi?id=1187425
+            zypper -n addrepo "https://download.suse.de/ibs/SUSE:/CA/openSUSE_Leap_${VERSION}" 'SUSE:CA'
+        else
+            zypper -n addrepo obs://SUSE:CA SUSE:CA
```
asdil12, line 53: `SUSE:CA` is not in obs but on ibs, so this would just have added the non existent obs repo.
I wonder if this ever worked.

```suggestion
            zypper -n addrepo http://download.suse.de/ibs/SUSE:/CA/openSUSE_Tumbleweed/ SUSE:CA
```

script/openqa-bootstrap
```diff
@@ -46,7 +46,12 @@ if ping -c1 download.suse.de. && (! rpm -q ca-certificates-suse) ; then
     if ! zypper info ca-certificates-suse | grep -q ':' ; then
         # add suse ca repo if needed
         # use this way of adding the repo to be distro agnostic
-        zypper -n addrepo obs://SUSE:CA SUSE:CA
+        if [ "$NAME" = "openSUSE Leap" ]; then
+            # avoid using `obs://…` URL to workaround https://bugzilla.opensuse.org/show_bug.cgi?id=1187425
+            zypper -n addrepo "https://download.suse.de/ibs/SUSE:/CA/openSUSE_Leap_${VERSION}" 'SUSE:CA'
```
asdil12, line 51: We can't use `https` before the ca is added.

```suggestion
            zypper -n addrepo "http://download.suse.de/ibs/SUSE:/CA/openSUSE_Leap_${VERSION}" 'SUSE:CA'
```

asdil12 (changes requested):

script/openqa-bootstrap
Martchus: It should have worked before. The error one gets under Leap 15.3 is:

```
Repository 'SUSE:CA' is invalid.
[SUSE:CA|http://download.suse.de/ibs/SUSE:/CA/Leap_15.3] Valid metadata not found at specified URL
```

So it is actually somehow clever enough to use `download.suse.de` as domain.

script/openqa-bootstrap
Martchus: In my test it worked: https://paste.opensuse.org/36389794

However, I have no idea how. Maybe it is better to use just http instead of relying on some magic I don't understand.

script/openqa-bootstrap
Martchus: Since it is clever enough I'll keep it this way. Keeping the `obs://`-URL in the `else` branch has the advantage that we don't have to handle different distribution names.

asdil12: That's what I thought too - after all I wrote that line initially I guess.

But:
```
# podman run -it --rm opensuse/tumbleweed bash
d6963ac04377:/ # zypper -n addrepo obs://SUSE:CA SUSE:CA
Guessed: platform = openSUSE_Tumbleweed
Adding repository 'SUSE:CA' ................................................................................................................................................................................[done]
Repository 'SUSE:CA' successfully added

URI         : https://download.opensuse.org/repositories/SUSE:/CA/openSUSE_Tumbleweed
Enabled     : Yes
GPG Check   : Yes
Autorefresh : No
Priority    : 99 (default priority)

Repository priorities are without effect. All enabled repositories share the same priority.
```

asdil12: And I can't find code in https://github.com/openSUSE/zypper/blob/master/src/utils/misc.cc#L504 that would do that.
In fact the zypper sourcecode doesn't contain the string `download.suse.de`

asdil12: Well never mind - I knew that back then and replace the url afterwards:
https://github.com/os-autoinst/openQA/blob/1fe2f6888265b99f7b6c4d8d64bb0f7b021320db/script/openqa-bootstrap#L55

asdil12 (approved):

PR #3969, opened 2021-06-18
Optimize search: prefetch job columns

perlpunk:
Currently for every job-module result a `SELECT FROM job` is executed,
when calling `$job_module->job->name`.

The prefetch will save the job data in the object and so the extra
SELECT is avoided.
Additionally this commit reduces the selected columns to what is actually
needed.

A search for 'vlc' `/api/v1/experimental/search?q=vlc` was reduced from 39s to 7s locally.

### DIFF
--- a/lib/OpenQA/WebAPI/Controller/API/V1/Search.pm
+++ b/lib/OpenQA/WebAPI/Controller/API/V1/Search.pm
@@ -129,7 +129,10 @@ sub _search_job_modules {
         {-or => {name => $like}},
         {
             join     => 'job',
-            group_by => ['me.id', 'job.DISTRI', 'job.VERSION', 'job.FLAVOR', 'job.TEST', 'job.ARCH', 'job.MACHINE'],
+            group_by =>
+              ['me.id', 'job.DISTRI', 'job.VERSION', 'job.FLAVOR', 'job.TEST', 'job.ARCH', 'job.MACHINE', 'job.id'],
+            prefetch => [qw(job)],
+            select   => [qw(me.id me.script me.job_id job.DISTRI job.VERSION job.FLAVOR job.ARCH job.BUILD job.TEST)],
             order_by => {-desc => 'job_id'}})->slice(0, $limit);
     while (my $job_module = $job_modules->next) {
         my $contents = $job_module->script;


Martchus (approved):

lib/OpenQA/WebAPI/Controller/API/V1/Search.pm
```diff
@@ -129,7 +129,10 @@ sub _search_job_modules {
         {-or => {name => $like}},
         {
             join     => 'job',
-            group_by => ['me.id', 'job.DISTRI', 'job.VERSION', 'job.FLAVOR', 'job.TEST', 'job.ARCH', 'job.MACHINE'],
+            group_by =>
+              ['me.id', 'job.DISTRI', 'job.VERSION', 'job.FLAVOR', 'job.TEST', 'job.ARCH', 'job.MACHINE', 'job.id'],
+            prefetch => [qw(job)],    # Avoid an extra select in every iteration
```
okurz, line 134: I consider the comment redundant. If a prefetch is done then it's because the job data is accessed from job module objects, that should be clear. In case code changes causing the job to be not accessed anymore then the comment would not be correct anymore

kalikiana: I don't understand the objection.

The comment tells me that a `SELECT` can be avoided. Without the comment, I'd be tempted to drop it and and I wouldn't even see a poblem if I tried. It would still work.

Martchus: I guess the comment has been added to stop people from doing willy-nilly "simplifications" here.

perlpunk: I guess @okurz ' point is that `prefetch` is kind of self-documenting.
If you don't know what it does, you shouldn't remove it without looking it up.
Otherwise we should add such a comment to every prefetch in our code.

okurz (approved):

lib/OpenQA/WebAPI/Controller/API/V1/Search.pm
kraih: Performance optimisations of ORM queries are hard, and the circumstances that led to them are often hard to replicate. So i'm in favour of comments that make the intention obvious to anyone who might be working on the same code in the future.

perlpunk: @kraih too late now ;-)

PR #3971, opened 2021-06-21
Fix show results for multiple groups in /test/overview

ilausuch:
When the query has different groupid and every group
different builds the current operation only returns the jobs
for the last build of the first group.

This PR ensures to return all the jobs for all the last builds for
all the selected groups

https://progress.opensuse.org/issues/91650

### DIFF
--- a/lib/OpenQA/WebAPI/Controller/Test.pm
+++ b/lib/OpenQA/WebAPI/Controller/Test.pm
@@ -673,7 +673,7 @@ sub overview {
     my %stash = (
         # build, version, distri are not mandatory and therefore not
         # necessarily come from the search args so they can be undefined.
-        build   => $search_args->{build},
+        build   => ref $search_args->{build} eq 'ARRAY' ? join(',', @{$search_args->{build}}) : $search_args->{build},
         version => $search_args->{version},
         distri  => $search_args->{distri},
         groups  => $groups,
@@ -681,7 +681,6 @@ sub overview {
     );
     my @latest_jobs = $self->schema->resultset('Jobs')->complex_query(%$search_args)->latest_jobs($until);
     ($stash{archs}, $stash{results}, $stash{aggregated}) = $self->prepare_job_results(\@latest_jobs);
-
     # determine distri/version from job results if not explicitely specified via search args
     my @distris = keys %{$stash{results}};
     my $formatted_distri;
--- a/lib/OpenQA/WebAPI/Plugin/Helpers.pm
+++ b/lib/OpenQA/WebAPI/Plugin/Helpers.pm
@@ -382,11 +382,17 @@ sub _compose_job_overview_search_args {
 
     # determine build number
     if (!$search_args{build}) {
-        # yield the latest build of the first group if (multiple) groups but not build specified
-        # note: the search arg 'groupid' is ignored by complex_query() because we later assign 'groupids'
-        $search_args{groupid} = $groups[0]->id if (@groups);
-
-        $search_args{build} = $schema->resultset('Jobs')->latest_build(%search_args);
+        if (@groups) {
+            my %builds;
+            for my $group (@groups) {
+                my $build = $schema->resultset('Jobs')->latest_build(%search_args, groupid => $group->id) or next;
+                $builds{$build}++;
+            }
+            $search_args{build} = [sort keys %builds];
+        }
+        else {
+            $search_args{build} = $schema->resultset('Jobs')->latest_build(%search_args);
+        }
 
         # print debug output
         if (@groups == 0) {
--- a/t/10-tests_overview.t
+++ b/t/10-tests_overview.t
@@ -233,10 +233,39 @@ $t->get_ok('/tests/overview?distri=opensuse&version=13.1&groupid=1001&groupid=10
 $summary = get_summary;
 like(
     $summary,
-    qr/Summary of opensuse, opensuse test/i,
-    'multiple groups with no build specified yield latest build of first group'
+    qr/Summary of opensuse, opensuse test build 0091 /i,
+    'multiple groups with no build specified yield latest build of every group'
 );
 like($summary, qr/Passed: 2 Failed: 0 Scheduled: 1 Running: 2 None: 1/i);
+
+my $jobGroup = $t->app->schema->resultset('JobGroups')->create(
+    {
+        id         => 1003,
+        sort_order => 0,
+        name       => 'opensuse test 2'
+    });
+
+my $job = $t->app->schema->resultset('Jobs')->create(
+    {
+        id       => 99964,
+        BUILD    => '0092',
+        group_id => 1003,
+        TEST     => 'kde',
+        DISTRI   => 'opensuse',
+        VERSION  => '13.1'
+    });
+
+$t->get_ok('/tests/overview?distri=opensuse&version=13.1&groupid=1001&groupid=1003')->status_is(200);
+$summary = get_summary;
+like(
+    $summary,
+    qr/Summary of opensuse, opensuse test 2 build 0091,0092 /i,
+    'multiple groups with no build specified yield latest build of every group'
+);
+like($summary, qr/Passed: 3 Failed: 0 Scheduled: 2 Running: 1 None: 1/i);
+
+$jobGroup->delete();
+$job->delete();
 
 # overview page searches for all available data with less specified parameters
 $t->get_ok('/tests/overview' => form => {build => '0091', version => '13.1'})->status_is(200);


ilausuch:
Remains the test

lib/OpenQA/WebAPI/Plugin/Helpers.pm
```diff
@@ -382,11 +382,17 @@ sub _compose_job_overview_search_args {
...
     if (!$search_args{build}) {
-        # yield the latest build of the first group if (multiple) groups but not build specified
-        # note: the search arg 'groupid' is ignored by complex_query() because we later assign 'groupids'
-        $search_args{groupid} = $groups[0]->id if (@groups);
-
-        $search_args{build} = $schema->resultset('Jobs')->latest_build(%search_args);
+        if (@groups){
+            my @builds;
+            @builds = map {
+                $search_args{groupid} = $_->{_column_data}->{id};
```
perlpunk, line 388: Please don't use internal data of objects, especially not those which have an underscore in front :)

```suggestion
                $search_args{groupid} = $_->id;
```

lib/OpenQA/WebAPI/Plugin/Helpers.pm
```diff
@@ -382,11 +382,17 @@ sub _compose_job_overview_search_args {
...
+                $schema->resultset('Jobs')->latest_build(%search_args);
+            } @groups;
+            $search_args{build} = \@builds;
```
perlpunk, line 391: Some of the builds (or even all) gathered in the previous `map` might be undef.
```suggestion
            $search_args{build} = [grep { defined $_ } @builds];
```

lib/OpenQA/WebAPI/Plugin/Helpers.pm
ilausuch: I think I tested this before and didn't work. I'll check it again

lib/OpenQA/WebAPI/Plugin/Helpers.pm
perlpunk: In my test I got `[ undef, ':TW.8318' ]` in `@builds`.

Yes, conditionally returning `()` would also work.

e.g.
```perl
$schema->resultset('Jobs')->latest_build(%search_args) || ();
```

One could even save another statement:
```perl
my @builds = map {
    $schema->resultset('Jobs')->latest_build(%search_args, groupid => $_->id) || ();
} @groups;
```

lib/OpenQA/WebAPI/Plugin/Helpers.pm
ilausuch: I am not sure, but this doesn't work for me. the test runs but not in a browser

ilausuch:
An other interesting thing is that in the test now I have the same build duplicated 0091,0091 so I should introduce a "distinct" of build and not include duplicates

lib/OpenQA/WebAPI/Plugin/Helpers.pm
perlpunk: What do you mean with "it doesn't work"?
It works for me, I get the expected result.
Do you get an error message?
Did you trace the SQL statements with DBIC_TRACE?

lib/OpenQA/WebAPI/Plugin/Helpers.pm
perlpunk: To avoid duplicated build names (as you mentioned below) I suggest the following:
```perl
            my %builds;
            for my $group (@groups) {
                my $build = $schema->resultset('Jobs')->latest_build(%search_args, groupid => $group->id) or next;
                $builds{ $build }++;
            }
            $search_args{build} = [sort keys %builds];
```

lib/OpenQA/WebAPI/Controller/Test.pm
```diff
@@ -695,6 +694,8 @@ sub overview {
         }
     }
 
+
+
```
okurz, line 698: unrelated whitespace changes, please revert

lib/OpenQA/WebAPI/Plugin/Helpers.pm
ilausuch: done

lib/OpenQA/WebAPI/Plugin/Helpers.pm
ilausuch: Solved with your next suggestion

lib/OpenQA/WebAPI/Controller/Test.pm
ilausuch: done

t/10-tests_overview.t
```diff
@@ -233,8 +233,8 @@ $t->get_ok('/tests/overview?distri=opensuse&version=13.1&groupid=1001&groupid=10
 $summary = get_summary;
 like(
     $summary,
-    qr/Summary of opensuse, opensuse test/i,
-    'multiple groups with no build specified yield latest build of first group'
+    qr/Summary of opensuse, opensuse test build 0091 /i,
```
ilausuch, line 236: This extra space is to ensure we don't have more than one build in the string

t/10-tests_overview.t
```diff
@@ -233,8 +233,8 @@ $t->get_ok('/tests/overview?distri=opensuse&version=13.1&groupid=1001&groupid=10
...
+    'multiple groups with no build specified yield latest build of every group'
```
perlpunk, line 237: It would probably make sense to add a test with groups that result in two different build names. Right now the result is like before, so the test doesn't prove that the change works.

ilausuch: Done

t/10-tests_overview.t
```diff
@@ -229,14 +229,23 @@ $t->element_exists('#res_DVD_i586_kde',                           'job from grou
...
+# $t->get_ok('/tests/overview?distri=opensuse&version=13.1&groupid=1001&groupid=1002')->status_is(200);
+# $summary = get_summary;
+# like(
+#     $summary,
+#     qr/Summary of opensuse, opensuse test build 0091 /i,
+#     'multiple groups with no build specified yield latest build of every group'
+# );
+# like($summary, qr/Passed: 2 Failed: 0 Scheduled: 1 Running: 2 None: 1/i);
+
 $t->get_ok('/tests/overview?distri=opensuse&version=13.1&groupid=1001&groupid=1002')->status_is(200);
```
okurz, line 241: Instead of adding the job to the lengthy fixture list breaking many other test modules better just create a single job before you do the "get", e.g.

```perl
$t->app->schema->resultset('Jobs')->create({
        id => 99964,
        BUILD => '0092',
        group_id => 1002,
        TEST => 'kde'
    });
```

okurz (changes requested):

t/10-tests_overview.t
kalikiana: Why not use `$` there?

lib/OpenQA/WebAPI/Controller/Test.pm
```diff
@@ -673,15 +673,14 @@ sub overview {
...
-        build   => $search_args->{build},
+        build   => ref $search_args->{build} eq 'ARRAY' ? join(',', @{$search_args->{build}}) : $search_args->{build},
         version => $search_args->{version},
         distri  => $search_args->{distri},
         groups  => $groups,
         until   => $until,
     );
     my @latest_jobs = $self->schema->resultset('Jobs')->complex_query(%$search_args)->latest_jobs($until);
     ($stash{archs}, $stash{results}, $stash{aggregated}) = $self->prepare_job_results(\@latest_jobs);
-
```
okurz, line 684: please remove the unrelated whitespace change

t/10-tests_overview.t
okurz: Because then the regex shouldn't match

okurz (changes requested):
Please squash the commits. The test should be part of the same commit as the implementation

t/10-tests_overview.t
```diff
@@ -233,11 +233,40 @@ $t->get_ok('/tests/overview?distri=opensuse&version=13.1&groupid=1001&groupid=10
...
+        DISTRI   => 'opensuse',
+        VERSION  => '13.1'
+    });
+
+$t->get_ok('/tests/overview?distri=opensuse&version=13.1&groupid=1001&groupid=1003')->status_is(200);
+$summary = get_summary;
+like(
+    $summary,
+    qr/Summary of opensuse, opensuse test 2 build 0091,0092 /i,
+    'multiple groups with no build specified yield latest build of every group'
```
perlpunk, line 263: The test label is now the same as the previous one.
Labels should be different for each test, so that it is easy to find out which test failed.

t/10-tests_overview.t
kalikiana:
```suggestion
    qr/Summary of opensuse, opensuse test build 0091[^,]/i,
```

okurz: yes, that works. I have created https://github.com/os-autoinst/openQA/pull/3973 incorporating the latest review suggestions from here so please approve there as long as ilausuch did not update their PR

t/10-tests_overview.t
okurz: incorporated into https://github.com/os-autoinst/openQA/pull/3973

lib/OpenQA/WebAPI/Controller/Test.pm
okurz: included in https://github.com/os-autoinst/openQA/pull/3973

ilausuch:
incorporated into #3973

PR #3972, opened 2021-06-21
Optimize group_overview: don't fetch bugdetails where not needed

perlpunk:
This will affect /group_overview/ pages and the homepage.

Also select only relevant columns from Comments.

Issue: https://progress.opensuse.org/issues/94354

### DIFF
--- a/lib/OpenQA/Schema/ResultSet/Comments.pm
+++ b/lib/OpenQA/Schema/ResultSet/Comments.pm
@@ -43,17 +43,23 @@ sub referenced_bugs {
 
 =over 4
 
-=item comment_data_for_jobs($jobs)
+=item comment_data_for_jobs($jobs, $args)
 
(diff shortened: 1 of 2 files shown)


okurz (approved):

Martchus (commented):
I'm only wondering why both invocations pass `{bugdetails => 1})`. Isn't at least one supposed to omit the bug details?

perlpunk:
> I'm only wondering why both invocations pass `{bugdetails => 1})`. Isn't at least one supposed to omit the bug details?

Yes, there are three, so the one which I didn't update is the one that is called here:
https://github.com/os-autoinst/openQA/blob/master/lib/OpenQA/BuildResults.pm#L221

I just realized that I should update the pod for this function.

Martchus (approved):

kalikiana (approved):
Neat!

PR #3978, opened 2021-06-25
openqa-label-all: Allow to call without specific build

okurz:
Related progress issue: https://progress.opensuse.org/issues/94714

### DIFF
--- a/script/openqa-label-all
+++ b/script/openqa-label-all
@@ -26,7 +26,7 @@ def parse_args():
     parser = argparse.ArgumentParser(description=__doc__, formatter_class=CustomFormatter)
     parser.add_argument('-v', '--verbose',
                         help="Increase verbosity level, specify multiple times to increase verbosity", action='count', default=1)
-    parser.add_argument('--build', required=True,
+    parser.add_argument('--build',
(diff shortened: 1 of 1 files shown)


kalikiana (changes requested):
Can you please add a test for this in `t/40-script_openqa-label-all.t`?

okurz:
> Can you please add a test for this in `t/40-script_openqa-label-all.t`?

I'd rather not as the test creates a complete process instance, it's not a unit test for a python module or something. In case we decide it's necessary to go deeper then we should rather use a pytest test module. But what I change here is merely parameter parsing which is kinda checking itself with parameter validity checks.

kalikiana (approved):

Martchus (approved):
Btw, this should better use URL encoding.

okurz:
> Btw, this should better use URL encoding.

ok. How could this look like?

Martchus:
I'm normally not using Python but there's likely a utility function within the library you're using to do the query.

perlpunk:
> Btw, this should better use URL encoding.

Maybe even just use the existing variable `request_params` instead? Then the library will do that automatically.

PR #3979, opened 2021-06-25
Use upstream Minion feature to perform gru jobs in tests

kraih:
This is the first followup to #3934, where i've introduced a workaround to speed up a very slow test. The workaround has since been turned into an [upstream feature](https://github.com/mojolicious/minion/compare/de1057fa4607...b8bb86e119c2) we can now use.

And since all our gru jobs run in forked processes in production, there is also a `TEST_FORK_MINION_JOBS=1` environment variable provided by `OpenQA::Test::Utils::perform_minion_jobs`. You can use it for debugging, to quickly rule out issues related to differences in how the job was processed.

I've only converted a few additional tests for now, but will make more PRs when everything goes well.


kraih:
Looks like i broke the git history of my branch...

kraih:
Ok, that's better.

cpanfile
```diff
@@ -49,7 +49,7 @@ requires 'JSON::Validator';
 requires 'JavaScript::Minifier::XS', '>= 0.11';
 requires 'LWP::Protocol::https';
 requires 'LWP::UserAgent';
-requires 'Minion', '>= 10.12';
+requires 'Minion', '>= 10.22';
```
Martchus, line 52: It would actually be just a test requirement but I guess distinguishing this won't bring us an advantage in practice.

Martchus (approved):

cpanfile
kraih: Probably counter productive since we might accidentally end up depending on newer features. And then we wouldn't notice it since we deploy and test with the latest version only. It would only fail for some users who happen to use older versions.

kalikiana (approved):

PR #3981, opened 2021-06-28
Also show "investigation" tabs for incomplete jobs

okurz:
Also in case of incomplete jobs the problem can be caused by something
that the investigation tab could reveal, e.g. invalid test code that
causes the job to abort prematurely with incomplete.

Related progress issue: https://progress.opensuse.org/issues/94792

### DIFF
--- a/assets/javascripts/test_result.js
+++ b/assets/javascripts/test_result.js
@@ -62,7 +62,7 @@ const tabConfiguration = {
   investigation: {
     descriptiveName: 'investigation info',
     conditionForShowingNavItem: function () {
-      return testStatus.state === 'done' && testStatus.result === 'failed';
+      return testStatus.state === 'done' && (testStatus.result === 'failed' || testStatus.result === 'incomplete');
     },
     renderContents: renderInvestigationTab
   },
--- a/lib/OpenQA/Jobs/Constants.pm
+++ b/lib/OpenQA/Jobs/Constants.pm
@@ -140,3 +140,6 @@ my %META_RESULT_MAPPING = (
 );
 sub meta_state ($state) { $META_STATE_MAPPING{$state} // NONE }
 sub meta_result ($result) { $META_RESULT_MAPPING{$result} // NONE }
+sub is_ok_result ($result) {
+    !!grep { $result eq $_ } OK_RESULTS;
+}
--- a/lib/OpenQA/Schema/Result/Jobs.pm
+++ b/lib/OpenQA/Schema/Result/Jobs.pm
@@ -1957,7 +1957,7 @@ sub investigate {
     return {error => 'No result directory available for current job'} unless $self->result_dir();
     my $ignore = OpenQA::App->singleton->config->{global}->{job_investigate_ignore};
     for my $prev (@previous) {
-        if ($prev->result eq 'failed') {
+        if ($prev->should_show_investigation) {
             $inv{first_bad} = {type => 'link', link => '/tests/' . $prev->id, text => $prev->id};
             next;
         }
@@ -2223,7 +2223,8 @@ sub should_show_autoinst_log {
 sub should_show_investigation {
     my ($self) = @_;
 
-    return $self->state ne DONE || $self->result eq FAILED;
+    OpenQA::Jobs::Constants::meta_state($self->state) ne OpenQA::Jobs::Constants::FINAL
+      || !OpenQA::Jobs::Constants::is_ok_result($self->result);
 }
 
 sub status {
--- a/t/ui/18-tests-details.t
+++ b/t/ui/18-tests-details.t
@@ -266,6 +266,9 @@ subtest 'reason and log details on incomplete jobs' => sub {
     my $log_element = $driver->find_element_by_xpath('//*[@id="details"]//pre[string-length(text()) > 0]');
     like($log_element->get_attribute('data-src'), qr/autoinst-log.txt/, 'log file embedded');
     like($log_element->get_text(), qr/Crashed\?/, 'log contents loaded');
+    $driver->find_element_by_link_text('Investigation')->click;
+    $driver->find_element('table#investigation_status_entry')
+      ->text_like(qr/cannot provide hints/, 'investigation status content shown as table');
 };
 
 sub update_status {


lib/OpenQA/Schema/Result/Jobs.pm
```diff
@@ -2235,10 +2235,16 @@ sub should_show_autoinst_log {
     return $self->state eq DONE && !$self->has_modules && $self->has_autoinst_log;
 }
 
+sub _should_show_investigation {
```
Martchus, line 2238: Don't add a new function here which does the same. Just use `$prev->should_show_investigation` in the code above.

Martchus (changes requested):

lib/OpenQA/Schema/Result/Jobs.pm
okurz: of course, "$prev" is a job as well :D

okurz: done

Martchus (changes requested):
```
        #   Failed test 'previous job identified as last good and it is a hash'
        #   at t/10-jobs.t line 473.
        #          got: ''
        #     expected: 'HASH'
        # Looks like you failed 1 test of 4.

    #   Failed test 'additional investigation notes provided on new failed'
    #   at t/10-jobs.t line 489.
    # Looks like you failed 1 test of 10.

#   Failed test 'carry over, including soft-fails'
```

lib/OpenQA/Schema/Result/Jobs.pm
```diff
@@ -1986,7 +1986,7 @@ sub investigate {
     return {error => 'No result directory available for current job'} unless $self->result_dir();
     my $ignore = OpenQA::App->singleton->config->{global}->{job_investigate_ignore};
     for my $prev (@previous) {
-        if ($prev->result eq 'failed') {
+        if ($prev->should_show_investigation) {
```
okurz, line 1989: this change causes tests to fail, don't know why. `$prev->result =~ /^failed$|incomplete/` would work but I would prefer to use the constants. But also I wonder if we should not change the investigate method to only look for a "first bad" which has the same result as the current job. Like, if we are looking for a previous job with the same problem, then it has to have the same result, otherwise it's a different problem, right?

lib/OpenQA/Schema/Result/Jobs.pm
```diff
@@ -1986,7 +1986,7 @@ sub investigate {
     return {error => 'No result directory available for current job'} unless $self->result_dir();
     my $ignore = OpenQA::App->singleton->config->{global}->{job_investigate_ignore};
     for my $prev (@previous) {
-        if ($prev->result eq 'failed') {
+        if ($prev->result =~ /^failed$|incomplete/) {
```
Martchus, line 1989:
```suggestion
        if ($prev->result =~ /^(failed|incomplete)$|/) {
```

Although I would prefer using the constants `FAILED`/`INCOMPLETE`.

okurz: yes, I tried that as well but all of that is not working: https://github.com/os-autoinst/openQA/pull/3981#discussion_r665682365

lib/OpenQA/Schema/Result/Jobs.pm
```diff
@@ -2232,7 +2232,7 @@ sub should_show_autoinst_log {
 sub should_show_investigation {
     my ($self) = @_;
 
-    return $self->state ne DONE || $self->result eq FAILED;
+    return ! grep { /^$self->state$/ } FINAL_STATES || grep { /^$self->result$/ } NOT_OK_RESULTS;
```
Martchus, line 2235: I'd avoid calling `state`/`result` for every state/result we want to consider. And is is really required to turn it into a regex (`$_ eq $state` should work as well)? At least for the states one could also use `meta_state` and check whether it returns `FINAL`. I would also add braces to make sure the precedence is what we want.

Martchus (changes requested):

lib/OpenQA/Schema/Result/Jobs.pm
Martchus: Not sure why it is not working but I added another review comment for `should_show_autoinst_log`.

lib/OpenQA/Schema/Result/Jobs.pm
okurz: tried with

```perl
OpenQA::Jobs::Constants::meta_state($self->state) eq OpenQA::Jobs::Constants::FINAL || grep { /^$self->result$/ } NOT_OK_RE       SULTS;
```

Do you know why the previous code says "done || failed" and not "done && failed"?

Martchus: > Do you know why the previous code says "done || failed" and not "done && failed"?

No, I also think `&&` would make more sense.

lib/OpenQA/Jobs/Constants.pm
```diff
@@ -156,3 +156,8 @@ sub meta_result {
     my ($result) = @_;
     return $META_RESULT_MAPPING{$result} // NONE;
 }
+
+sub is_ok_result {
+    my ($result) = @_;
+    grep { $result eq $_ } NOT_OK_RESULTS;
```
Martchus, line 162:
```suggestion
    grep { $result eq $_ } OK_RESULTS;
```

And maybe use signatures right away (you don't have to convert the whole module).

okurz: of course. But now – as expected – t/10-jobs.t fails

okurz:
Ready for review, waiting for second approval

assets/javascripts/test_result.js
```diff
@@ -62,7 +62,7 @@ const tabConfiguration = {
   investigation: {
     descriptiveName: 'investigation info',
     conditionForShowingNavItem: function () {
-      return testStatus.state === 'done' && testStatus.result === 'failed';
+      return testStatus.state === 'done' && (testStatus.result === 'failed' || testStatus.result === 'incomplete');
```
kalikiana, line 65: Would be nice if we could also make this "if we have inestigation data" rather than adding more states. But I guess we don't really know that here?

kalikiana (changes requested):
Code looks nice.

Would you please also add a test for it?

okurz:
On 12/01/2022 14.10, kalikiana wrote:
> ***@***.**** requested changes on this pull request.
> 
> Code looks nice.
> 
> Would you please also add a test for it?

I don't want to

assets/javascripts/test_result.js
Martchus: I suppose we can always run the investigation. It is just a matter how useful it'll be. I would avoid running the investigation at the time we decide which tabs to show because running the investigation isn't for free.

lib/OpenQA/Schema/Result/Jobs.pm
```diff
@@ -2223,7 +2223,8 @@ sub should_show_autoinst_log {
 sub should_show_investigation {
     my ($self) = @_;
 
-    return $self->state ne DONE || $self->result eq FAILED;
+    OpenQA::Jobs::Constants::meta_state($self->state) eq OpenQA::Jobs::Constants::FINAL
+      && !OpenQA::Jobs::Constants::is_ok_result($self->result);
```
Martchus, line 2227: This changes the behavior. I suppose

```suggestion
    OpenQA::Jobs::Constants::meta_state($self->state) ne OpenQA::Jobs::Constants::FINAL
      || !OpenQA::Jobs::Constants::is_ok_result($self->result);
```

would be correct.

Note that as long as the job state isn't final we need to render the tab server-side (even if it is hidden client-side) because when the job turns out to be a failure the tab is supposed to be shown after all (and thus set visible by the client-side after all).

---

Of course this means you cannot use this function in the way you're using it above. Maybe the function deserves a bitter name or the two slightly different usages should be inlined.

Martchus (changes requested):

lib/OpenQA/Schema/Result/Jobs.pm
Martchus: Looks like I'm contradicting my previous comment (https://github.com/os-autoinst/openQA/pull/3981#discussion_r678298027) but I guess I just haven't thought it trough enough last time.

okurz: Could you please test this?

okurz:
> If this is a joke I don't get it.

It's no joke. We already have sufficient statement coverage and the only risk I see is that the investigation tab does not show up which I covered manually

lib/OpenQA/Schema/Result/Jobs.pm
Martchus: Confirmed. If a test is running and turns into failing the investigation tab is shown correctly (without reloading the page). That relies on the tab being rendered on the server-side for non-final states.

Martchus: By the way, I've tested that by simply setting the job state/result manually via SQL while having the tab open.

okurz: Updated. I think the name of the method is still correct as is

Martchus (approved):
In `sub investigate` your version of `should_show_investigation` would likely make more sense but I suppose it doesn't matter because the client won't show the tab anyways until a final job state is reached.

okurz:
> In `sub investigate` your version of `should_show_investigation` would likely make more sense but I suppose it doesn't matter because the client won't show the tab anyways until a final job state is reached.

Yes. That's why it's called "should show" not "will most definitely do show" ;)

kalikiana:
> It's no joke. We already have sufficient statement coverage and the only risk I see is that the investigation tab does not show up which I covered manually

You didn't mention anything about why tests can't be added, or manual testing, in the description, commit message or ticket.

Is it difficult to add? I would guess `18-tests-details.t` is easy to extend? You can always ask for help.

kalikiana:
> Is it difficult to add? I would guess `18-tests-details.t` is easy to extend? You can always ask for help.

Oli took me up on my offer in chat, so I'll provide one shortly.

okurz:
> > Is it difficult to add? I would guess `18-tests-details.t` is easy to extend? You can always ask for help.
> 
> Oli took me up on my offer in chat, so I'll provide one shortly.

You can just accept this PR as is :)

okurz (commented):
@cdywan your changes LGTM, you can approve again

kalikiana (approved):

PR #3982, opened 2021-06-28
Fix a small mistake where "if (%foo" was used instead of "if (keys %foo"

kraih:
Probably not a fix for the actual issue. But a small mistake i saw while reading the code.

Progress: https://progress.opensuse.org/issues/94531

### DIFF
--- a/lib/OpenQA/Schema/Result/JobGroups.pm
+++ b/lib/OpenQA/Schema/Result/JobGroups.pm
@@ -359,7 +359,7 @@ sub to_template {
                     if ($attr->{machine} eq $default_machine) {
                         delete $attr->{machine} if $test_suites{$arch}{$name} == 1;
                     }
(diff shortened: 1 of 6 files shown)


kalikiana (commented):
I'm guessing we can't test this easily because this works in some of the cases?

kraih:
While i'm at it, i've fixed a few more cases. This is just a question of good Perl style really.

perlpunk:
Btw, the return value of `scalar %hash` changed in perl 5.26 to return the number of keys:
https://perldoc.perl.org/perl5260delta#scalar(%25hash)-return-signature-changed

What minimum perl version do we require?

kraih:
Should be Perl 5.20 for signatures now.

okurz (approved):

Martchus (approved):

PR #3985, opened 2021-06-29
BugLinks: Do not attempt label with incomplete referer URLs

asdil12:
Ticket: https://progress.opensuse.org/issues/94774

All common bugtrackers use decimal bugids so check if
the referer URL is complete by making sure that it contains
a digit.

### DIFF
--- a/lib/OpenQA/Schema/ResultSet/Jobs.pm
+++ b/lib/OpenQA/Schema/ResultSet/Jobs.pm
@@ -442,13 +442,14 @@ sub stale_ones {
 sub mark_job_linked {
     my ($self, $jobid, $referer_url) = @_;
 
-    my $referer = Mojo::URL->new($referer_url)->host;
-    my $app     = OpenQA::App->singleton;
-    return undef unless $referer;
-    return log_debug("Unrecognized referer '$referer'")
-      unless grep { $referer eq $_ } @{$app->config->{global}->{recognized_referers}};
+    my $referer      = Mojo::URL->new($referer_url);
+    my $referer_host = $referer->host;
+    my $app          = OpenQA::App->singleton;
+    return undef unless $referer_host;
+    return log_debug("Unrecognized referer '$referer_host'")
+      unless grep { $referer_host eq $_ } @{$app->config->{global}->{recognized_referers}};
     my $job = $self->find({id => $jobid});
-    return unless $job;
+    return undef if !$job || ($referer->path_query =~ /^\/?$/);
     my $comments = $job->comments;
     return undef if $comments->find({text => {like => 'label:linked%'}});
     my $user = $self->result_source->schema->resultset('Users')->search({username => 'system'})->first;
--- a/t/10-jobs.t
+++ b/t/10-jobs.t
@@ -643,6 +643,7 @@ sub job_is_linked {
 }
 
 subtest 'job is marked as linked if accessed from recognized referal' => sub {
+    my $test_referer = 'http://test.referer.info/foobar';
     $t->app->config->{global}->{recognized_referers}
       = ['test.referer.info', 'test.referer1.info', 'test.referer2.info', 'test.referer3.info'];
     my %_settings = %settings;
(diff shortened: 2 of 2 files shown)


perlpunk (approved):

perlpunk:
There might be bugtrackers having a digit in their hostname, though.
Maybe check if the URI path has a length > 0 (`http://bugtracker.example/bug/123`) or if there is a query string.

okurz (changes requested):
Please look into tests first ensuring what you want to achieve before going into an implementation. I don't think it's a good idea to introduce an additional restricting for "how does a ticket reference look like" only within the "mark_job_linked" function. We have multiple more places where we handle ticket references and we don't want them to apply different rules. But I agree that a bare link to "bugzilla.suse.com" is not helpful and we can handle that.

asdil12:
Updated the PR.
I'm going with Tina's approach here as it is agnostic of the bugtracker used.

lib/OpenQA/Schema/ResultSet/Jobs.pm
```diff
@@ -442,13 +442,14 @@ sub stale_ones {
...
 
-    my $referer = Mojo::URL->new($referer_url)->host;
+    my $referer = Mojo::URL->new($referer_url);
     my $app     = OpenQA::App->singleton;
-    return undef unless $referer;
-    return log_debug("Unrecognized referer '$referer'")
-      unless grep { $referer eq $_ } @{$app->config->{global}->{recognized_referers}};
+    return undef unless $referer->host;
+    return log_debug("Unrecognized referer '".$referer->host."'")
+      unless grep { $referer->host eq $_ } @{$app->config->{global}->{recognized_referers}};
```
Martchus, line 449: Avoid repeated calls of the same function:
```suggestion
    return undef unless my $referer_host = $referer->host;
    return log_debug("Unrecognized referer '$referer_host'")
      unless grep { $referer_host eq $_ } @{$app->config->{global}->{recognized_referers}};
```

lib/OpenQA/Schema/ResultSet/Jobs.pm
```diff
@@ -442,13 +442,14 @@ sub stale_ones {
...
     my $job = $self->find({id => $jobid});
     return unless $job;
+    return undef if $referer->path_query =~ /^\/?$/;
```
Martchus, line 452:
```suggestion
    return undef if !$job || ($referer->path_query =~ /^\/?$/);
```

Martchus (changes requested):

okurz:
I can think of use cases where referral URLs without anything after a slash would make sense. Would the current way not also stumble over URLs like http://bug.tracker/issue/42/ ?

asdil12:
> I can think of use cases where referral URLs without anything after a slash would make sense. Would the current way not also stumble over URLs like http://bug.tracker/issue/42/ ?

```perl
perl> Mojo::URL->new("http://bug.tracker/issue/42/")->path_query;
/issue/42/
```

No - that should be covered, as I'm using the combined path and get parameters.

t/10-jobs.t
```diff
@@ -662,7 +662,8 @@ subtest 'job is marked as linked if accessed from recognized referal' => sub {
     $job->update;
     $linked = job_is_linked($job);
     is($linked, 0, 'new job is not linked');
-    $t->get_ok('/tests/' . $job->id . '/modules/' . $module->id . '/steps/1' => {Referer => 'http://test.referer.info'})
+    $t->get_ok(
+        '/tests/' . $job->id . '/modules/' . $module->id . '/steps/1' => {Referer => 'http://test.referer.info/foobar'})
```
okurz, line 666: To prevent the line split I suggest to pull out the string 'http://test.referer.info/foobar' into a variable which you can reuse

okurz (changes requested):

okurz (approved):

Martchus (approved):

PR #3986, opened 2021-06-29
Revert "Switch to systemd-sysusers instead of getent + useradd"

dcermak:
This reverts commit 72a6d7f87b8773d3ffa8116e02f100966c0064eb.

The sysusers switch will not work until https://build.opensuse.org/request/show/901701 is in Factory.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -110,10 +110,6 @@ install-generic:
 
 	cp -Ra dbicdh "$(DESTDIR)"/usr/share/openqa/dbicdh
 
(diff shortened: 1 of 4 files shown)


okurz (approved):

Martchus (approved):

okurz:
It's not just waiting for https://build.opensuse.org/request/show/901701 . We want to be compatible with openSUSE Leap 15.2 as well. Maybe some maintenance update that we need to wait for as well? Or introduce a background compatible way.

PR #3988, opened 2021-06-29
Fix carry-over only due to one matching bugref in step title

Martchus:
* Fix carry-over only due to one matching bugref in step title
  * Prevent carry-over despite a mismatch of failing test modules like in
    https://progress.opensuse.org/issues/94880
  * Still allow carry-over when the same bug is found via different modules
    (ef5f07c561cb4bb6139d17b98b616af6b81a8fff) but only if the other failing
    test modules match as well
* Consider all modules on carry over, despite missing/broken results
  * If it isn't possible to read the results/details of some (soft)failed
    modules that doesn't mean we should ignore them completely. In addition,
    further (soft)failed modules should not be ignored as well.

---

I've been adding tests now. I've also been testing that tests fail without the change (as there was already a similar test in place which wasn't good enough).

### DIFF
--- a/t/17-labels_carry_over.t
+++ b/t/17-labels_carry_over.t
@@ -19,36 +19,27 @@ use Test::Most;
 
 use FindBin;
 use lib "$FindBin::Bin/lib", "$FindBin::Bin/../external/os-autoinst-common/lib";
+use Mojo::Base -signatures;
 use Test::Mojo;
 use Test::Warnings ':report_warnings';
 use OpenQA::Test::Case;
 use OpenQA::Test::TimeLimit '10';
 use Mojo::JSON qw(decode_json);
 
-my $test_case;
-my $schema;
-my $t;
-my $auth;
-my $rs;
+my $test_case = OpenQA::Test::Case->new;
+my $schema    = $test_case->init_data(fixtures_glob => '01-jobs.pl 03-users.pl 05-job_modules.pl');
+my $t         = Test::Mojo->new('OpenQA::WebAPI');
+my $rs        = $t->app->schema->resultset('Jobs');
+my $auth      = {'X-CSRF-Token' => $t->ua->get('/tests')->res->dom->at('meta[name=csrf-token]')->attr('content')};
+$test_case->login($t, 'percival');
+
 my $comment_must
   = '<a href="https://bugzilla.suse.com/show_bug.cgi?id=1234">bsc#1234</a>(Automatic takeover from <a href="/tests/99962">t#99962</a>)';
-
-sub set_up {
-    $test_case = OpenQA::Test::Case->new;
-    $schema    = $test_case->init_data(fixtures_glob => '01-jobs.pl 03-users.pl 05-job_modules.pl');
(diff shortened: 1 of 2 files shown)


okurz:
I like the code style change :)

Vogtinator (approved):
The algorithm seems sound to me.

The only weakness I could identify is that when a failed module gains a bugref in test code, that blocks carryover. Shouldn't really be an issue in practice, and the previous code had that as well.

Martchus:
> The only weakness I could identify is that when a failed module gains a bugref in test code, that blocks carryover. Shouldn't really be an issue in practice, and the previous code had that as well.

I thought about it as well but thought it wasn't worth it as it would complicate the code a bit. We can still implement it later if needed.

t/17-labels_carry_over.t
```diff
@@ -127,37 +118,47 @@ subtest 'failed->failed labels which are not bugrefs are *not* carried over' =>
...
-    $rs->find(99962)->update_module('aplay',     {result => 'fail', details => [{title => 'bsc#77777'}]});
-    $rs->find(99963)->update_module('yast2_lan', {result => 'fail', details => [{title => 'bsc#77777'}]});
+subtest 'failed in different modules with same bugref in details' => sub {
+    # Fail test in different modules with same bug reference and a 3rd module
+    $prev_job->update_module('aplay',      {result => 'fail', details => [{title => 'bsc#77777'}]});
+    $curr_job->update_module('yast2_lan',  {result => 'fail', details => [{title => 'bsc#77777'}]});
+    $curr_job->update_module('bootloader', {result => 'softfail'});
     $t->post_ok('/api/v1/jobs/99963/set_done', $auth => form => {result => 'failed'})->status_is(200);
+    is @{comments('/tests/99963')}, 0,
+      'no carry-over when 3rd module failes, despite a matching bugref between other modules';
```
okurz, line 150: s/failes/fails/

t/17-labels_carry_over.t
```diff
@@ -154,4 +155,10 @@ subtest 'failed in different modules with same bugref in details' => sub {
     is join('', @{comments('/tests/99963')}), $comment_must, 'label is carried over without other failing modules';
 };
 
+subtest 'failure reason still computed without results' => sub {
+    my $mock = Test::MockModule->new('OpenQA::Schema::Result::JobModules');
+    $mock->redefine(results => undef);
```
okurz, line 160: `$mock->noop('results')`?

okurz (changes requested):
Sorry, I don't understand. What has the "carry-over" have to do with "matching bugref in step title"? I thought carry-over is only looking at failed modules, not steps?

Martchus:
> I thought carry-over is only looking at failed modules, not steps?

That is wrong. With https://github.com/os-autoinst/openQA/commit/ef5f07c561cb4bb6139d17b98b616af6b81a8fff (referenced in the commit message) it has been changed to also take bugrefs within step titles into account.

okurz (approved):

kalikiana (approved):

PR #3990, opened 2021-06-30
overview: Rate limit requests per user and globally

kalikiana:
Related: https://progress.opensuse.org/issues/94901

image

### DIFF
--- a/lib/OpenQA/Setup.pm
+++ b/lib/OpenQA/Setup.pm
@@ -57,7 +57,8 @@ sub read_config {
             auto_clone_regex            => '^(cache failure|terminated prematurely): ',
         },
         rate_limits => {
-            search => 5,
+            overview => 5,
+            search   => 5,
         },
         auth => {
             method => 'OpenID',
--- a/lib/OpenQA/WebAPI/Controller/API/V1/Search.pm
+++ b/lib/OpenQA/WebAPI/Controller/API/V1/Search.pm
@@ -181,11 +181,7 @@ sub query {
     $validation->required('q');
     return $self->reply->validation_error({format => 'json'}) if $validation->has_error;
 
-    # Allow n queries per minute, per user (if logged in)
-    my $lockname = 'webui_query_rate_limit';
-    if (my $user = $self->current_user) { $lockname .= $user->username }
-    return $self->render(json => {error => 'Rate limit exceeded'}, status => 400)
-      unless $self->app->minion->lock($lockname, 60, {limit => $self->app->config->{'rate_limits'}->{'search'}});
+    return undef if $self->rate_limit({route => 'search'});
 
     my $cap = $self->app->config->{'global'}->{'search_results_limit'};
     my @results;
--- a/lib/OpenQA/WebAPI/Controller/Test.pm
+++ b/lib/OpenQA/WebAPI/Controller/Test.pm
@@ -666,6 +666,9 @@ sub _add_distri_and_version_to_summary {
 # A generic query page showing test results in a configurable matrix
 sub overview {
     my ($self) = @_;
+
+    $self->stash(title => 'Test summary');
+    return undef if $self->rate_limit({route => 'overview'});
     my ($search_args, $groups) = $self->compose_job_overview_search_args;
     my $validation = $self->validation;
     $validation->optional('t')->datetime;
--- a/lib/OpenQA/WebAPI/Plugin/Helpers.pm
+++ b/lib/OpenQA/WebAPI/Plugin/Helpers.pm
@@ -20,6 +20,7 @@ use Mojo::ByteStream;
 use OpenQA::Schema;
 use OpenQA::Utils qw(bugurl render_escaped_refs href_to_bugref);
 use OpenQA::Events;
+use Time::Seconds;
 
 sub register ($self, $app, $config) {
     $app->helper(
@@ -333,6 +334,7 @@ sub register ($self, $app, $config) {
 
     $app->helper('reply.validation_error' => \&_validation_error);
 
+    $app->helper(rate_limit                       => \&_rate_limit);
     $app->helper(compose_job_overview_search_args => \&_compose_job_overview_search_args);
     $app->helper(param_hash                       => \&_param_hash);
     $app->helper(
@@ -495,6 +497,21 @@ sub _validation_error {
     my $error  = "Erroneous parameters ($failed)";
     return $c->render(json => {error => $error}, status => 400) if $format eq 'json';
     return $c->render(text => $error,            status => 400);
+}
+
+sub _rate_limit ($c, $args) {
+    my $route = $args->{route};
+    die "Rate limit helper requires route argument" unless defined $route;
+    # Allow n queries per minute, per user (if logged in)
+    my $lockname = "webui_${route}_rate_limit";
+    if (my $user = $c->current_user) { $lockname .= $user->username }
+    my $error = "Rate limit exceeded";
+    my $limit = $c->app->config->{'rate_limits'}->{$route};
+    return $c->respond_to(
+        json => {json     => {error => $error},          status => 429},
+        html => {template => 'layouts/rate_limit_error', limit  => $limit},
+    ) unless $c->app->minion->lock($lockname, ONE_MINUTE, {limit => $limit});
+    return undef;
 }
 
 1;
--- a/t/10-tests_overview.t
+++ b/t/10-tests_overview.t
@@ -27,7 +27,8 @@ use Mojo::Parameters;
 my $test_case = OpenQA::Test::Case->new;
 $test_case->init_data(fixtures_glob => '01-jobs.pl 03-users.pl 05-job_modules.pl');
 
-my $t      = Test::Mojo->new('OpenQA::WebAPI');
+my $t = Test::Mojo->new('OpenQA::WebAPI');
+$t->app->config->{rate_limits}->{overview} = 30;
 my $schema = $t->app->schema;
 
 sub get_summary {
@@ -387,5 +388,13 @@ $t->get_ok(
     })->status_is(200);
 like(get_summary, qr/Passed: 0 Failed: 0/i, 'Job was successful, so failed_modules does not show it');
 $t->element_exists_not('#res-99946', 'no module has failed');
+
+subtest 'Errors' => sub {
+    $t->app->config->{rate_limits}->{overview} = 3;
+    $t->get_ok('/tests/overview', {Accept => 'application/json'})->content_type_is('application/json;charset=UTF-8')
+      for (1 .. 3);
+    $t->status_is(429, 'Rate limit exceeded');
+    $t->json_is('/error' => 'Rate limit exceeded', 'rate limit triggered');
+};
 
 done_testing();
--- a/templates/webapi/layouts/rate_limit_error.html.ep
+++ b/templates/webapi/layouts/rate_limit_error.html.ep
@@ -1,0 +1,12 @@
+% layout 'bootstrap';
+
+<div>
+    <h2>Rate limit exceeded</h2>
+    <div id="summary" class="card">
+      <div class="container" id="content">
+          The rate limit of <%= $limit %> requests per minute was exceeded for this API route. If you're logged in
+          the limit counts against your account, otherwise it's counted globally for everyone. Please reach out
+          to the maintainers of this openQA instance if you feel it's too low.
+      </div>
+    </div>
+</div>
--- a/t/api/15-search.t
+++ b/t/api/15-search.t
@@ -135,6 +135,7 @@ subtest 'Errors' => sub {
 
     $t->app->config->{rate_limits}->{search} = 3;
     $t->get_ok('/api/v1/experimental/search?q=timezone', 'Search succesful') for (1 .. 3);
+    $t->status_is(429, 'Rate limit exceeded');
     $t->json_is('/error' => 'Rate limit exceeded', 'rate limit triggered');
 };
 
--- a/t/config.t
+++ b/t/config.t
@@ -61,7 +61,8 @@ subtest 'Test configuration default modes' => sub {
             worker_timeout              => DEFAULT_WORKER_TIMEOUT,
         },
         rate_limits => {
-            search => 5,
+            overview => 5,
+            search   => 5,
         },
         auth => {
             method => 'Fake',


lib/OpenQA/WebAPI/Controller/Test.pm
```diff
@@ -681,6 +681,12 @@ sub overview {
     );
     my @latest_jobs = $self->schema->resultset('Jobs')->complex_query(%$search_args)->latest_jobs($until);
     ($stash{archs}, $stash{results}, $stash{aggregated}) = $self->prepare_job_results(\@latest_jobs);
+ 
+    # Allow n queries per minute, per user (if logged in)
+    my $lockname = 'webui_overview_rate_limit';
+    if (my $user = $self->current_user) { $lockname .= $user }
```
Martchus, line 687: I would expect that this only happens the address of the user object.

lib/OpenQA/WebAPI/Controller/Test.pm
```diff
@@ -681,6 +681,12 @@ sub overview {
...
+    return $self->render(json => {error => 'Rate limit exceeded'}, status => 400)
+      unless $self->app->minion->lock($lockname, 60, {limit => $self->app->config->{'rate_limits'}->{'overview'}});
```
okurz, line 689: Please replace magic number. I assume 60 means "one minute"? If yes, there is a time date module we can use

lib/OpenQA/WebAPI/Controller/Test.pm
kalikiana: That's a string, like `cdywan`, if logged in

lib/OpenQA/WebAPI/Controller/Test.pm
okurz: How did you resolve this? I don't see a change

lib/OpenQA/WebAPI/Controller/Test.pm
Martchus: I don't think so, it'll likely just be something like `OpenQA::Schema::Result::Users=HASH(0x563a8d87dd40)`.

perlpunk: This is already used here:
https://github.com/os-autoinst/openQA/blob/master/lib/OpenQA/WebAPI/Controller/API/V1/Search.pm#L186

So either this is also wrong, or the user is a string :)

kalikiana: Yeah, I thought it must be sound since I took it from existing code... I guess we never had tests to back it up. I tried to add new unit tests but failed to figure out how to get authenticated API calls working that have a username in the controller.
And I have a fix for the existing code in any case: #3996

lib/OpenQA/WebAPI/Controller/Test.pm
```diff
@@ -682,6 +683,12 @@ sub overview {
     my @latest_jobs = $self->schema->resultset('Jobs')->complex_query(%$search_args)->latest_jobs($until);
     ($stash{archs}, $stash{results}, $stash{aggregated}) = $self->prepare_job_results(\@latest_jobs);
 
+    # Allow n queries per minute, per user (if logged in)
+    my $lockname = 'webui_overview_rate_limit';
+    if (my $user = $self->current_user) { $lockname .= $user->username }
+    return $self->render(json => {error => 'Rate limit exceeded'}, status => 400)
```
okurz, line 689: Why do this *after* querying the db for jobs? Shouldn't it be done before?

lib/OpenQA/Setup.pm
```diff
@@ -57,7 +57,8 @@ sub read_config {
             auto_clone_regex            => '^(cache failure|terminated prematurely): ',
         },
         rate_limits => {
-            search => 5,
+            overview => 5,
```
okurz, line 60: So does this mean that users can not access any test overview page more than 5 times per minute? In that case the limit is in range of 10-100 times below what power users might do

okurz (changes requested):

lib/OpenQA/WebAPI/Controller/Test.pm
Martchus: It would also be nice to make this a helper instead of duplicating code from the search.

lib/OpenQA/Setup.pm
Martchus: And considering this is more of a production tweak, maybe the default should be a very high value (possibly infinity)?

okurz: I could think of 200

lib/OpenQA/WebAPI/Controller/Test.pm
kraih: Agreed, definitely should be a helper.

lib/OpenQA/Setup.pm
coolo: 200 a minute?

okurz: Yes, 200/minute

coolo: Then I must misunderstand something - if this is about the webui, 200 requests a minute don't sound like a user but like a web crawler

okurz: I wouldn't outright deny web crawlers. We could point web crawlers to the API :) thinking of a "power user" having multiple pages open and using automatic page refresh. Not the best approach but considering that we don't provide the users good hints in the error message I opted for this high number. But actually I would favor a more helpful error message

coolo: If someone opts to rely on 200 refreshes a minute, he's a moron trying to sabotage his colleagues.

kalikiana: > But actually I would favor a more helpful error message

If that's your concern: I'm looking into that. But somehow I can't get the negociation for json and html to work. So I pushed my non-working attempt now anyway, and maybe someone can spot where I'm failing.

lib/OpenQA/WebAPI/Controller/API/V1/Search.pm
```diff
@@ -181,11 +181,7 @@ sub query {
     $validation->required('q');
     return $self->reply->validation_error({format => 'json'}) if $validation->has_error;
 
-    # Allow n queries per minute, per user (if logged in)
-    my $lockname = 'webui_query_rate_limit';
-    if (my $user = $self->current_user) { $lockname .= $user->username }
-    return $self->render(json => {error => 'Rate limit exceeded'}, status => 400)
-      unless $self->app->minion->lock($lockname, 60, {limit => $self->app->config->{'rate_limits'}->{'search'}});
+    return undef if ($self->rate_limit({route => 'search'}));
```
okurz, line 184: delete the parentheses on the outer level

lib/OpenQA/WebAPI/Controller/Test.pm
```diff
@@ -666,6 +666,9 @@ sub _add_distri_and_version_to_summary {
 # A generic query page showing test results in a configurable matrix
 sub overview {
     my ($self) = @_;
+
+    return undef if ($self->rate_limit({route => 'overview'}));
```
okurz, line 670: same as above

lib/OpenQA/WebAPI/Controller/Test.pm
```diff
@@ -666,6 +666,9 @@ sub _add_distri_and_version_to_summary {
...
+
```
okurz, line 671: please delete the blank line

lib/OpenQA/WebAPI/Plugin/Helpers.pm
```diff
@@ -497,4 +499,22 @@ sub _validation_error {
...
+sub _rate_limit {
+    my ($c, $args) = @_;
+    my $route = $args->{route};
+    die "Rate limit helper requires route argument" unless defined $route;
+    # Allow n queries per minute, per user (if logged in)
+    my $lockname = "webui_${route}_rate_limit";
+    if (my $user = $c->current_user) { $lockname .= $user->username }
+    my $error = "Rate limit exceeded";
+    return $c->render(json => {error => $error}, status => 400)
+      unless $c->app->minion->lock($lockname, 60, {limit => $c->app->config->{'rate_limits'}->{$route}});
```
okurz, line 511: use time constant

lib/OpenQA/WebAPI/Plugin/Helpers.pm
```diff
@@ -497,4 +499,22 @@ sub _validation_error {
     return $c->render(text => $error,            status => 400);
 }
 
+sub _rate_limit {
+    my ($c, $args) = @_;
```
kraih, line 503: This class already has signatures activated, use them.

lib/OpenQA/WebAPI/Plugin/Helpers.pm
```diff
@@ -497,4 +499,22 @@ sub _validation_error {
...
+    my $route = $args->{route};
+    die "Rate limit helper requires route argument" unless defined $route;
+    # Allow n queries per minute, per user (if logged in)
+    my $lockname = "webui_${route}_rate_limit";
+    if (my $user = $c->current_user) { $lockname .= $user->username }
+    my $error = "Rate limit exceeded";
+    return $c->render(json => {error => $error}, status => 400)
```
kraih, line 510: The rate limit HTTP code is `429 Too Many Requests`.

t/10-tests_overview.t
```diff
@@ -388,4 +389,10 @@ $t->get_ok(
 like(get_summary, qr/Passed: 0 Failed: 0/i, 'Job was successful, so failed_modules does not show it');
 $t->element_exists_not('#res-99946', 'no module has failed');
 
+subtest 'Errors' => sub {
+    $t->app->config->{rate_limits}->{overview} = 3;
+    $t->get_ok('/tests/overview', 'Search succesful') for (1 .. 3);
+    $t->json_is('/error' => 'Rate limit exceeded', 'rate limit triggered');
```
kraih, line 395: The request does not negotiate JSON.

lib/OpenQA/WebAPI/Controller/API/V1/Search.pm
```diff
@@ -181,11 +181,7 @@ sub query {
     $validation->required('q');
     return $self->reply->validation_error({format => 'json'}) if $validation->has_error;
 
-    # Allow n queries per minute, per user (if logged in)
-    my $lockname = 'webui_query_rate_limit';
-    if (my $user = $self->current_user) { $lockname .= $user->username }
-    return $self->render(json => {error => 'Rate limit exceeded'}, status => 400)
-      unless $self->app->minion->lock($lockname, 60, {limit => $self->app->config->{'rate_limits'}->{'search'}});
+    return undef if $self->rate_limit({route => 'search'});
```
kraih, line 184: Actually looks like the return value is insignificant here.

lib/OpenQA/WebAPI/Controller/Test.pm
```diff
@@ -666,6 +666,9 @@ sub _add_distri_and_version_to_summary {
 # A generic query page showing test results in a configurable matrix
 sub overview {
     my ($self) = @_;
+
+    $self->stash(title => 'Test summary');
+    return undef if $self->rate_limit({route => 'overview'});
```
kraih, line 671: Same here, the return value does nothing.

t/10-tests_overview.t
```diff
@@ -388,4 +389,12 @@ $t->get_ok(
 like(get_summary, qr/Passed: 0 Failed: 0/i, 'Job was successful, so failed_modules does not show it');
 $t->element_exists_not('#res-99946', 'no module has failed');
 
+subtest 'Errors' => sub {
+    $t->app->config->{rate_limits}->{overview} = 3;
+    $t->get_ok('/tests/overview', {Accept => 'application/json'})->content_type_is('application/json;charset=UTF-8')
+      for (1 .. 3);
+    $t->status_is(429, 'Rate limit exceeded');
+    $t->json_is('/error' => 'Rate limit exceeded', 'rate limit triggered');
```
kraih, line 397: Looks like the HTML version has no test now?

kalikiana: Ack, thank you for noticing. I overlooked that. Will add

lib/OpenQA/Setup.pm
okurz: I like your error message but I consider 5 too low.

@Martchus you mentioned "possibly infinity", what do you propose?

Martchus: Since this is mainly a production tweak my suggestion is to disable rate limiting by default completely. That's what I meant with infinity here. Of course that is not a useful value if one actually wants to enable that feature.

templates/webapi/layouts/rate_limit_error.html.ep
```diff
@@ -0,0 +1,12 @@
+% layout 'bootstrap';
+
+<div>
+    <h2>Rate limit exceeded</h2>
+    <div id="summary" class="card">
+      <div class="container" id="content">
```
Martchus, line 6: The layout already adds an element with that ID.

templates/webapi/layouts/rate_limit_error.html.ep
```diff
@@ -0,0 +1,12 @@
+% layout 'bootstrap';
+
+<div>
```
Martchus, line 3: The outer `div` seems useless.

Martchus (changes requested):

lib/OpenQA/Setup.pm
okurz: I think we can still propose a good default that we consider makes sense in most cases. So then let me suggest 50

kraih:
Having thought about this some more, i would probably approach the whole issue a bit differently. When and for what we will actually end up needing rate limits is unknown and rather situational. Here we have to specifically add code to every action we  might want to protect in the future. Adding overhead to actions that are already too slow. So it should probably be disabled completely by default (putting unnecessary load on PostgreSQL is a waste of valuable resources).

My preference would probably be to put it on a different layer, like a [before_dispatch](https://metacpan.org/pod/Mojolicious#before_dispatch) hook that checks the path of the incoming request for a regex match from the configuration file. Or an [around_action](https://metacpan.org/pod/Mojolicious#around_action) hook that can check for a configured limit on the current route name. That would give us more flexibility and cost much less resources.

Martchus:
Implementing this in a more generic way seems reasonable. However, I also have nothing against merging this version at this point and generalizing it later.

PR #3992, opened 2021-06-30
Use signatures in OpenQA::Schema::Result::Jobs

Martchus:
Still a draft, let's see what this breaks (based on #3988)


okurz (commented):
This might conflict with https://github.com/os-autoinst/openQA/pull/3958 . Also I would prefer to apply signatures everywhere in one go so that we don't mix it too much with other changes that we introduce, see https://github.com/os-autoinst/os-autoinst/pull/1696

Martchus:
> Also I would prefer to apply signatures everywhere in one go

I wouldn't apply signatures everywhere at once. After all it is a semantic change which possibly breaks stuff.

okurz:
> > Also I would prefer to apply signatures everywhere in one go
> 
> I wouldn't apply signatures everywhere at once. After all it is a semantic change which possibly breaks stuff. (Likely this PR is already too much at once.)

ok, let me rephrase. I would prefer to apply signatures independantly of other changes. After all it is a semantic change which possibly breaks stuff – which is also the reason why the os-autoinst change isn't in yet ;)

okurz (approved):

PR #3993, opened 2021-06-30
Changes how the investigation tab displays the git log output

b10n1k:
This PR unifies the output of the git logs and git log stats into one
https://progress.opensuse.org/issues/91878

The purpose is to have a better view of what files have been changed for each
commit.

Example:
```
test_log

6a44799cb test
 lib/bootbasetest.pm     | 2 +-
 lib/opensusebasetest.pm | 2 +-
 2 files changed, 2 insertions(+), 2 deletions(-)
```

Screenshot_20211109_171700_investigation_git_log_clickable

Signed-off-by: ybonatakis <ybonatakis@suse.com>


Martchus:
* s/Changes/Change/
* The commit contains some unrelated changes.

b10n1k:
image

b10n1k:
> * The commit contains some unrelated changes.

i was wondering if i have to run the tidy before i submit. i can revert those. so tidy?

okurz (changes requested):
So personally as explained in the ticket already I am not convinced that this approach is beneficial in general. Also I would like to be careful changing an existing behaviour unconditionally. I am open to be convinced by multiple power users though if they like your proposed format better :) As an alternative add the new view based on a configurable feature switch. As suggested in the ticket we should go with https://progress.opensuse.org/issues/92731 first before considering changing the format. @b10n1k maybe this is something that you would like to try out first?

b10n1k:
> * s/Changes/Change/

Done

> * The commit contains some unrelated changes.

Reverted

b10n1k:
> So personally as explained in the ticket already I am not convinced that this approach is beneficial in general. Also I would like to be careful changing an existing behaviour unconditionally. I am open to be convinced by multiple power users though if they like your proposed format better :) As an alternative add the new view based on a configurable feature switch. As suggested in the ticket we should go with https://progress.opensuse.org/issues/92731 first before considering changing the format. @b10n1k maybe this is something that you would like to try out first?

https://progress.opensuse.org/issues/92731 is very cool to have but it might take more time for me. So i found some time to implement this as i had it in mind, and go as you had said ["to discuss over code rather just english text"](https://progress.opensuse.org/issues/91878#note-5)

b10n1k:
After the latest commit the results should be shown as this example http://aquarius.suse.cz:7900/tests/56#investigation

image

there are some things still to fixed to become a proper code (and testing apparently). but i would like to get some early feedbacks

assets/javascripts/test_result.js
```diff
@@ -586,6 +586,21 @@ function setCurrentPreviewFromStepLinkIfPossible(stepLink) {
   }
 }
 
+function githash_to_link(value) {
```
Martchus, line 589: That's not Perl, so please use JavaScript naming conventions, e.g.
```suggestion
function githashToLink(value) {
```
But also change the other occurrences.

assets/javascripts/test_result.js
```diff
@@ -586,6 +586,21 @@ function setCurrentPreviewFromStepLinkIfPossible(stepLink) {
...
+    var repo = "https://github.com/os-autoinst/os-autoinst-distri-opensuse/commit/";
+    var tmpvalues = value;
+    var stat_regex = /(\<a.+?)(?=(\<a))/sgm;
+
+    for (let i = 0; i < githashes.length; i++) {
+        tmpvalues = value.replace(githashes[i], githashes[i].link(repo + githashes[i]));
+        value = tmpvalues;
+    }
+    var gitstats = value.match(stat_regex);
+    return gitstats;
```
Martchus, line 601:
```suggestion
    return value.match(stat_regex);
```

assets/javascripts/test_result.js
```diff
@@ -778,60 +793,124 @@ function renderInvestigationTab(response) {
...
+
+        var valueElement = document.createElement('td');
+
+        if (type === "link") {
+            var html = document.createElement("a");
+            if (key === 'first_bad') {
+                alertbox = document.createElement("div");
+                alertbox.appendChild(document.createTextNode('Investigate the first bad test directly: '));
+                html.href = value.link + '#investigation';
+                html.innerHTML = value.text;
```
Martchus, line 834: Does this need escaping? Maybe you can just append a text node as child here?

Martchus (commented):
The style check is failing. You can run `tools/js-tidy` locally to fix this.

b10n1k:
@Martchus i read your comments and i will address. i just have another question. 
I wonder if the [hardcoded repo url](https://github.com/os-autoinst/openQA/pull/3993/files#diff-9a7b8016038c575acb899fc0aa2dd003296bc199c29e13c9ee52c39b835bb62dR592) can be fetched from the controller or something.

Martchus:
> I wonder if the hardcoded repo url can be fetched from the controller or something.

You mean for the sake of making it configurable (via the web UI process' config file)? That would make sense, we do that already in several pages, e.g. templates/webapi/admin/audit_log/productlog.html.ep and templates/webapi/step/edit.html.ep (the block started with `% content_for 'ready_function' => begin`).

assets/javascripts/test_result.js
b10n1k: Actually i havent touched this part at all. this block is never parsed against the `test_log`. i am not sure if i miss something on your comment.

Martchus: Ok, if it is just a reformatting/move of existing code, leave it as-is. I've just noticed it within the diff. (It could be that `value.text` is actually supposed to contain HTML-syntax here.)

assets/javascripts/test_result.js
b10n1k: Done

assets/javascripts/test_result.js
b10n1k: Done

okurz (changes requested):
I still prefer the previous concise list but maybe you can make use of collapsing to keep the concise list and unfold the entries on click? We already use this "more …" label. Maybe this can be a starting point for you

b10n1k:
> I still prefer the previous concise list but maybe you can make use of collapsing to keep the concise list and unfold the entries on click? We already use this "more …" label. Maybe this can be a starting point for you

thats a good idea. thanks

b10n1k:
image

okurz:
can you replace the leading "+" with a "-" if unfolded?

b10n1k:
> can you replace the leading "+" with a "-" if unfolded?

Actually i have. But i have some js strange behavior and i havent found a way through it yet. But the switch of the sign is there

assets/javascripts/test_result.js
```diff
@@ -586,6 +586,22 @@ function setCurrentPreviewFromStepLinkIfPossible(stepLink) {
...
+  var githashes = value.match(hashes_regex);
+  var repo = 'https://github.com/os-autoinst/os-autoinst-distri-opensuse/commit/';
+  var tmpvalues = value;
+  var stat_regex = /(\<a.+?)(?=(\<a))/gms;
+  if (githashes == null) {
+    return;
+  }
+  for (let i = 0; i < githashes.length; i++) {
+    tmpvalues = value.replace(githashes[i], githashes[i].link(repo + githashes[i]));
+    value = tmpvalues;
```
Martchus, line 600:
```suggestion
    value = value.replace(githashes[i], githashes[i].link(repo + githashes[i]));
```

And remove `var tmpvalues…`.

assets/javascripts/test_result.js
```diff
@@ -784,22 +800,67 @@ function renderInvestigationTab(response) {
...
+            logDetailsDiv.className = 'collapse';
+            stats = gitstats[i].split('\n')[0];
+            spanElem.innerHTML = stats;
+            logDetailsDiv.innerHTML = gitstats[i]
+              .split('\n')
+              .slice(1, gitstats.length - 1)
+              .join('\n');
+            statItem.appendChild(collapseSign);
+            statItem.appendChild(spanElem);
+            statItem.appendChild(logDetailsDiv);
```
Martchus, line 834:
```suggestion
            statItem.append(collapseSign, spanElem, logDetailsDiv);
```

assets/javascripts/test_result.js
b10n1k: ready for review

assets/javascripts/test_result.js
b10n1k: ready for review

b10n1k:
@Martchus Practically is ready. However i would like to discuss one or two things that they concern me.

assets/javascripts/test_result.js
```diff
@@ -586,6 +586,20 @@ function setCurrentPreviewFromStepLinkIfPossible(stepLink) {
   }
 }
 
+function githashToLink(value) {
+  var hashes_regex = /(\b[0-9a-f]{9}\b)/gms;
+  var githashes = value.match(hashes_regex);
+  var repo = 'https://github.com/os-autoinst/os-autoinst-distri-opensuse/commit/';
```
Martchus, line 592: I've tested it locally. All the links to GitHub shown in the screenshot lead to a 404 page because it tries to show commits for needles within the test repository. That shows that it is problematic to hard-code the URL here and that needles needed to be distinguished from tests. (See https://github.com/os-autoinst/openQA/pull/3993#issuecomment-885628393.)

assets/javascripts/test_result.js
```diff
@@ -586,6 +586,20 @@ function setCurrentPreviewFromStepLinkIfPossible(stepLink) {
...
+  var stat_regex = /(\<a.+?)(?=(\<a))/gms;
```
Martchus, line 593: That would be our normal coding style (in JavaScript):
```suggestion
  var statRegex = /(\<a.+?)(?=(\<a))/gms;
```

The same counts for the many other occurrences but it is also not *that* important.

assets/javascripts/test_result.js
```diff
@@ -586,6 +586,20 @@ function setCurrentPreviewFromStepLinkIfPossible(stepLink) {
...
+  if (githashes == null) {
```
Martchus, line 594: The `==` operator should be avoided most of the times, maybe just check for an array instead:
```suggestion
  if (!Array.isArray(githashes)) {
```

Martchus (commented):
* It seems to work, that's how it looks like when I tested it locally: screenshot_20210806_144150
* Using a single "+" or "-" sign as button is not the best but that could still be improved later.
* Extending the UI tests would be nice.
* You don't need to use it but I'd like to note that there's `Array.prototype.forEach()` in JavaScript.

assets/javascripts/test_result.js
b10n1k: yup this is one thing i need to improve and i wanted to ask about. i will see what i can do on Monday and i will try to reach you out for recommendations if you dont mind

assets/javascripts/test_result.js
b10n1k: this change is ready for review

assets/javascripts/test_result.js
b10n1k: this change is ready for review

assets/javascripts/test_result.js
b10n1k: this change is ready for review.

lib/OpenQA/Utils.pm
```diff
@@ -164,6 +166,35 @@ sub testcasedir {
...
+    );
+    my $path
+      = $args{repo} eq 'needles'
+      ? needledir($args{distri}, $args{version})
+      : testcasedir($args{distri}, $args{version});
+    my $filename = '';
+    if (-e path($path, '.git')) {
+        $filename = path($path, '.git', 'config');
+    }
+    else { log_warning("$path is not a git directory"); }
```
okurz, line 184: does it make sense to actually call this function if $path is not a git directory?

lib/OpenQA/Utils.pm
```diff
@@ -164,6 +166,35 @@ sub testcasedir {
...
+    my $giturl = '';
+    my $config = Config::Tiny->read($filename, 'utf8');
+    unless (defined $config) {
+        return $config->{'remote "origin"'}{url} if ($config->{'remote "origin"'}{url} =~ /^http(s?)/);
+        my @url_tokenized = split(':', $config->{'remote "origin"'}{url});
+        $url_tokenized[1] = $url_tokenized[1] =~ s/\.git/\/commit\//r;
+        my @githost = split('@', $url_tokenized[0]);
+        $giturl = 'https://' . $githost[1] . '/' . $url_tokenized[1];
+    }
+    return $giturl;
```
okurz, line 195:
```suggestion
    my $config = Config::Tiny->read($filename, 'utf8');
    return '' if defined $config;
    return $config->{'remote "origin"'}{url} if ($config->{'remote "origin"'}{url} =~ /^http(s?)/);
    my @url_tokenized = split(':', $config->{'remote "origin"'}{url});
    $url_tokenized[1] = $url_tokenized[1] =~ s/\.git/\/commit\//r;
    my @githost = split('@', $url_tokenized[0]);
    return 'https://' . $githost[1] . '/' . $url_tokenized[1];
```

but why would you want to return '' if the config is actually defined? Shouldn't it be the other way around?

lib/OpenQA/Utils.pm
```diff
@@ -164,6 +166,35 @@ sub testcasedir {
...
+    my %args = (
+        distri  => '',
+        version => '',
+        repo    => 'tests',
+        @_,
+    );
+    my $path
+      = $args{repo} eq 'needles'
+      ? needledir($args{distri}, $args{version})
+      : testcasedir($args{distri}, $args{version});
```
okurz, line 179: As this is the only use of the "repo" argument I suggest to make the argument a boolean, e.g. `needles => 1` instead of `repo => 'needles'`

templates/webapi/test/result.html.ep
```diff
@@ -1,6 +1,5 @@
 % layout 'bootstrap';
 % title $job->name . " test results";
-
```
okurz, line 3: unrelated change, please revert

templates/webapi/test/result.html.ep
```diff
@@ -16,7 +15,9 @@
 % content_for 'ready_function' => begin
     $('.timeago').timeago();
     setupResultButtons();
-    setupResult(<%= $job->id %>, '<%= $job->state %>', '<%= $job->result %>', '<%= url_for('status', testid => $job->id) %>', '<%= url_for('test_details', testid => $job->id) %>');
+    setupResult(<%= $job->id %>, '<%= $job->state %>', '<%= $job->result %>',
+    '<%= url_for('status', testid => $job->id) %>',
+'<%= url_for('test_details', testid => $job->id) %>');
```
okurz, line 20: looks like this is also just whitespace change and should be reverted unless I overlook the actual functional change

lib/OpenQA/Utils.pm
perlpunk: That one looks a bit weird:
```
$url_tokenized[1] = $url_tokenized[1] =~ s/\.git/\/commit\//r;
```
Why not this:
```
$url_tokenized[1] =~ s/\.git/\/commit\//;
```
or to improve readability:
```
$url_tokenized[1] =~ s{\.git}{/commit/};
```
And I probably would anchor the regex to the string end:
```
$url_tokenized[1] =~ s{\.git$}{/commit/};
```

templates/webapi/test/result.html.ep
b10n1k: reverted. however i think this is produced by the tidy-js. **this change is ready for review**

lib/OpenQA/Utils.pm
b10n1k: *this change is ready for review*

lib/OpenQA/Utils.pm
b10n1k: *this change is ready for review*
thanks @perlpunk @okurz

lib/OpenQA/Utils.pm
b10n1k: you mean the `log_warning`? wouldnt be useful info? maybe for debugging? actually i feel guilty not adding more logs.

okurz: No I mean does it make sense to call the complete function "gitrepodir" when the path "$path/.git" does not even exist when you would end up in this branch

assets/javascripts/test_result.js
```diff
@@ -861,6 +921,19 @@ function renderInvestigationTab(response) {
   tabPanelElement.dataset.initialized = true;
 }
 
+function toggleSign(elem) {
+  if (elem.className === 'collapsed' && elem.innerHTML === '+ ') {
+    elem.innerHTML = '- ';
+  } else {
+    elem.innerHTML = '+ ';
```
okurz, line 928: https://www.javascripttutorial.net/javascript-ternary-operator/ plz

lib/OpenQA/Utils.pm
```diff
@@ -166,6 +168,31 @@ sub testcasedir {
...
+        repo    => 'tests',
+        @_,
+    );
+    my $path
+      = $args{needles}
+      ? needledir($args{distri}, $args{version})
+      : testcasedir($args{distri}, $args{version});
+    my $filename = '';
+    if (-e path($path, '.git')) {
+        $filename = path($path, '.git', 'config');
```
okurz, line 184: ternary please :)

okurz (changes requested):

assets/javascripts/test_result.js
```diff
@@ -811,22 +822,71 @@ function renderInvestigationTab(response) {
...
+              logDetailsDiv.className = 'collapse';
+              stats = gitstats[i].split('\n')[0];
+              spanElem.innerHTML = stats;
+              logDetailsDiv.innerHTML = gitstats[i]
+                .split('\n')
+                .slice(1, gitstats.length - 1)
+                .join('\n');
+              statItem.append(collapseSign, spanElem, logDetailsDiv);
+
+              if (i < 5) {
```
okurz, line 861: magic number. Please define a constant with explanatory name

assets/javascripts/test_result.js
```diff
@@ -811,22 +822,71 @@ function renderInvestigationTab(response) {
...
+                enable_more = true;
+                preElementMore.appendChild(statItem);
+              }
+            }
+          }
+        } else {
+          var textLines = typeof value === 'string' ? value.split('\n') : [value];
+          var textLinesRest;
+
+          var lineLimit = 10;
```
okurz, line 873: same as above, magic number. Please define a constant with explanatory name

okurz (changes requested):

okurz:
@b10n1k I now received notifications about changes in your pull request but seems like you do not address my suggestions so I wonder changes you actually did?

lib/OpenQA/Utils.pm
b10n1k: i see. the function does the check itself. i dont see the problem with that. i tried to keep it simple and practical. this is called with `_show` and the results are assigned to `testgiturl` && `needlegiturl`. from here you can do what ever you want. has no side effect.

Martchus:
Looks like the packaging tests for the client script fail:

```
[   28s] + openqa-client --help
[   29s] Can't locate Config/Tiny.pm in @INC (you may need to install the Config::Tiny module) (@INC contains: /usr/share/openqa/script/../lib /usr/lib/perl5/site_perl/5.26.1/x86_64-linux-thread-multi /usr/lib/perl5/site_perl/5.26.1 /usr/lib/perl5/vendor_perl/5.26.1/x86_64-linux-thread-multi /usr/lib/perl5/vendor_perl/5.26.1 /usr/lib/perl5/5.26.1/x86_64-linux-thread-multi /usr/lib/perl5/5.26.1 /usr/lib/perl5/site_perl) at /usr/share/openqa/script/../lib/OpenQA/Utils.pm line 40.
[   29s] BEGIN failed--compilation aborted at /usr/share/openqa/script/../lib/OpenQA/Utils.pm line 40.
```

You need to add `use Config::Tiny;` as dependency. It should likely go under `common_requires` in `dependencies.yaml` and be applied in other files via `make update-deps`.

assets/javascripts/test_result.js
b10n1k: *this change is ready for review*

lib/OpenQA/Utils.pm
b10n1k: *this change is ready for review*

assets/javascripts/test_result.js
b10n1k: *this change is ready for review*

Martchus (approved):

assets/javascripts/test_result.js
b10n1k: *this change is ready for review*

lib/OpenQA/WebAPI/Controller/Test.pm
```diff
@@ -433,6 +435,8 @@ sub _show {
             show_autoinst_log  => $job->should_show_autoinst_log,
             show_investigation => $job->should_show_investigation,
             show_live_tab      => $job->state ne DONE,
+            testgiturl         => $test_giturl,
+            needlegiturl       => $needle_giturl,
```
Martchus, line 439: Since this is only required for contents *within* the investigation tab it would make more sense to determine these variables only when rendering this tab.

lib/OpenQA/Utils.pm
```diff
@@ -166,6 +168,28 @@ sub testcasedir {
     return $dir;
 }
 
+sub gitrepodir {
+    my %args = (
+        distri  => '',
+        version => '',
+        repo    => 'tests',
+        @_,
+    );
```
Martchus, line 177:
```suggestion
    my %args = (distri  => '', version => '', repo    => 'tests', @_);
```

lib/OpenQA/Utils.pm
```diff
@@ -166,6 +168,28 @@ sub testcasedir {
...
+      : testcasedir($args{distri}, $args{version});
+    my $filename = (-e path($path, '.git')) ? path($path, '.git', 'config') : '';
+    log_warning("$path is not a git directory") if $filename eq '';
+    my $config = Config::Tiny->read($filename, 'utf8');
+    return '' unless defined $config;
+    return $config->{'remote "origin"'}{url} if ($config->{'remote "origin"'}{url} =~ /^http(s?)/);
+    my @url_tokenized = split(':', $config->{'remote "origin"'}{url});
+    $url_tokenized[1] =~ s{\.git$}{/commit/};
+    my @githost = split('@', $url_tokenized[0]);
+    return 'https://' . $githost[1] . '/' . $url_tokenized[1];
```
Martchus, line 190:
```suggestion
    return "https://$githost[1]/$url_tokenized[1]";
```

Martchus (changes requested):

lib/OpenQA/WebAPI/Controller/Test.pm
b10n1k: it makes sense. but i thought that it really not destructing and it can be useful in other cases as well.

lib/OpenQA/Utils.pm
b10n1k: *this change is ready for review*

lib/OpenQA/Utils.pm
b10n1k: *this change is ready for review*

okurz:
please rebase and fix the conflict

lib/OpenQA/Utils.pm
```diff
@@ -166,6 +168,23 @@ sub testcasedir {
...
+      : testcasedir($args{distri}, $args{version});
+    my $filename = (-e path($path, '.git')) ? path($path, '.git', 'config') : '';
+    log_warning("$path is not a git directory") if $filename eq '';
+    my $config = Config::Tiny->read($filename, 'utf8');
+    return '' unless defined $config;
+    return $config->{'remote "origin"'}{url} if ($config->{'remote "origin"'}{url} =~ /^http(s?)/);
+    my @url_tokenized = split(':', $config->{'remote "origin"'}{url});
+    $url_tokenized[1] =~ s{\.git$}{/commit/};
+    my @githost = split('@', $url_tokenized[0]);
+    return "https://$githost[1]/${url_tokenized}[1]";
```
okurz, line 185: CI tests fail on this `${url_tokenized}[1]`, don't know if this is correct syntax

lib/OpenQA/Utils.pm
```diff
@@ -166,6 +168,23 @@ sub testcasedir {
     return $dir;
 }
 
+sub gitrepodir {
```
okurz, line 171: Please add documentation to this function. I would like to know what it is intended to do. Please use POD format

lib/OpenQA/Utils.pm
```diff
@@ -166,6 +168,23 @@ sub testcasedir {
...
+    my %args = (distri => '', version => '', repo => 'tests', @_,);
```
okurz, line 172: this is a bit unusual syntax. What do you want to achieve? supply default values for variables if not supplied from outside? Why should repo default to "tests"?

lib/OpenQA/Utils.pm
```diff
@@ -166,6 +168,23 @@ sub testcasedir {
...
+    my $path
+      = $args{needles}
+      ? needledir($args{distri}, $args{version})
+      : testcasedir($args{distri}, $args{version});
```
okurz, line 176: Can this be put in a single line or does it get too long for tidy to accept?

lib/OpenQA/Utils.pm
b10n1k: fixed

lib/OpenQA/WebAPI/Controller/Test.pm
okurz: I don't know what you mean by that. If I understand Martchus correctly this should say:

```suggestion
            testgiturl         => gitrepodir(distri => $job->DISTRI),
            needlegiturl       => gitrepodir(distri => $job->DISTRI, needles => 1),
```

t/10-jobs.t
```diff
@@ -488,6 +488,25 @@ subtest 'carry over, including soft-fails' => sub {
         is($inv->{test_log}, 'No test changes recorded, test regression unlikely', 'git log with no test changes');
     };
 
+    subtest 'investigation provides test_log with git stats' => sub {
+        path('t/data/last_good.json')->copy_to(path(($job->_previous_scenario_jobs)[1]->result_dir(), 'vars.json'));
+        path('t/data/first_bad.json')->copy_to(path($job->result_dir(),                               'vars.json'));
+        $job->done;
+        is($job->result, OpenQA::Jobs::Constants::FAILED, 'job result is failed');
+        ok(my $inv = $job->investigate, 'job can provide investigation details');
+        ok($inv,                        'job provides failure investigation');
```
okurz, line 497: this is a verbatim copy of the lines in the former subtest. Can you combine both tests to not repeat this? You could nest another subtest within the above. If you really want to keep it then better extract a helper function but then avoid any Test::More commands in here and please use unique test description strings

t/16-utils.t
```diff
@@ -341,25 +341,32 @@ subtest 'base_host' => sub {
...
     is sharedir(),  '/tmp/test/openqa/share',         'right directory';
     is resultdir(), '/tmp/test/openqa/testresults',   'right directory';
     is assetdir(),  '/tmp/test/openqa/share/factory', 'right directory';
     is imagesdir(), '/tmp/test/openqa/images',        'right directory';
-
+    my $distri = 'opensuse';
+    is gitrepodir(distri => $distri), '', 'empty when .git is missing';
+    my $mocked_git = path(sharedir . "/tests/opensuse/.git")->make_path;
+    $mocked_git->child('config')
+      ->spurt(qq{[remote "origin"]\n        url = git\@github.com:b10n1k/os-autoinst-distri-opensuse.git});
```
okurz, line 360: better avoid personal avatar names in the openQA source code. And please don't hardcode to os-autoinst-distri-opensuse, how about github.com/os-autoinst/os-autoinst-distri-example ?

okurz (changes requested):
See the comments. Please be aware that I am still not sure about the approach in general to show the stat only within single commits. The starting point for test writers will likely often be when a test fails "which files within the test distribution have changed". I think in your current approach this is not possible to see anymore with a single click, right?

lib/OpenQA/Utils.pm
b10n1k: i think it was done by tidy. not sure.

t/16-utils.t
b10n1k: *this change is ready for review*

lib/OpenQA/Utils.pm
b10n1k: *this change is ready for review* tidy didnt complain

lib/OpenQA/Utils.pm
b10n1k: *this change is ready for review*

t/10-jobs.t
b10n1k: *this change is ready for review*

lib/OpenQA/Utils.pm
```diff
@@ -166,6 +168,30 @@ sub testcasedir {
     return $dir;
 }
 
+=head2 gitrepodir(distri => DISTRI, version => VERSION)
```
okurz, line 171: I don't this works too put it on the same line

lib/OpenQA/Utils.pm
```diff
@@ -166,6 +168,30 @@ sub testcasedir {
...
+
+I<gitrepodir> reads the F<.git/config> of the projects and returns
+the https address to the git repo. The expected repo used is B<origin>
```
okurz, line 174: Git repos don't need to be https

t/10-jobs.t
okurz: Sorry, I don't see a difference

okurz:
@b10n1k did you see my message in https://github.com/os-autoinst/openQA/pull/3993#pullrequestreview-770265757 ? How about a slight alternative: Keep the current overall diff between "last good" and current and just add your features additionally? Should be an easy change, just revert the deletions you did

b10n1k:
> @b10n1k did you see my message in [#3993 (review)](https://github.com/os-autoinst/openQA/pull/3993#pullrequestreview-770265757) ? 

i missed to reply on this and i was away for some time. Sorry for that. Find my answer below

> How about a slight alternative: Keep the current overall diff between "last good" and current and just add your features additionally? Should be an easy change, just revert the deletions you did

one part would be to restore the deleted lines in `Jobs.pm` and the other would be some adjustment in the javascript. 

> See the comments. Please be aware that I am still not sure about the approach in general to show the stat only within single commits. The starting point for test writers will likely often be when a test fails "which files within the test distribution have changed". I think in your current approach this is not possible to see anymore with a single click, right?

The simple answer is no. And that was my plan from the beginning. The reason was because a change in any file is not connected with a failure necessarily (the change can be unrelated). I think it is more easy to go through the commits (good title and then the description and finally the related modified files) to find the main cause. 
The only way to achieve the optimal solution it would be to have interchangeable links from the files to the commits and from the commits to the files. but this would be tricky when one files is linked to many commits.

templates/webapi/test/result.html.ep
```diff
@@ -89,8 +89,11 @@
                     <div role="tabpanel" class="tab-pane" id="dependencies" data-src="<%= url_for('test_dependencies', testid => $testid) %>" data-current-job-id="<%= $testid %>"></div>
                 % }
                 % if ($show_investigation) {
-                    <div role="tabpanel" class="tab-pane" id="investigation" data-src="<%= url_for('test_investigation', testid => $testid) %>"></div>
-                % }
+                    <div role="tabpanel" class="tab-pane" id="investigation"
+                data-src="<%= url_for('test_investigation', testid => $testid)
+                %>" data-needlegiturl="<%= stash('needlegiturl') %>"
```
Martchus, line 94: Wrapping it this way would look much nicer:
```suggestion
                data-src="<%= url_for('test_investigation', testid => $testid) %>"
                data-needlegiturl="<%= stash('needlegiturl') %>"
```

Martchus (changes requested):
The last commit message is a bit "low-level" and lacks proper casing. Since you've updated your changes it would be nice to know what you've actually changed (also regarding @okurz 's review comments).

b10n1k:
> > The simple answer is no. And that was my plan from the beginning. The reason was because a change in any file is not connected with a failure necessarily (the change can be unrelated). I think it is more easy to go through the commits (good title and then the description and finally the related modified files) to find the main cause.
> 
> Well, I still see it as easier to do a single click to find the complete diff. But my suggestion was that you would simply offer both ways in the separate parts of the investigation tab. So if you bring back the old behaviour for the complete diff and just add the new functionality within the commit list then we can find an agreement :)

i did some changes but i am not completed sure if this what you meant.

okurz (changes requested):
Overall code looks fine so far. Do you have a test instance available where we could try this out?

https://github.com/os-autoinst/openQA/pull/3993/commits/01c5da6f538e1692d85341caf276f1ebfb685edf deletes some parts which you bring back with https://github.com/os-autoinst/openQA/pull/3993/commits/fa11567fba175abdc1d70151c8c289a3ed85af13 . I suggest you combine these changes into one commit but don't squash all commits.

b10n1k:
@okurz i have the local dev one. Let me know when you want to take a look

okurz (approved):
I checked it on @b10n1k's instance and attached the screenshot to the PR description. Looks good. Great work! Approved :)

okurz (approved):

lib/OpenQA/WebAPI/Controller/Test.pm
Martchus: I must still say it would make more sense to render these URLs on the page where they are actually needed (and this is the tab and not the overall page).

lib/OpenQA/WebAPI/Controller/Test.pm
```diff
@@ -421,6 +421,8 @@ sub _show {
             show_autoinst_log => $job->should_show_autoinst_log,
             show_investigation => $job->should_show_investigation,
             show_live_tab => $job->state ne DONE,
+            testgiturl => gitrepodir(distri => $job->DISTRI),
+            needlegiturl => gitrepodir(distri => $job->DISTRI, needles => 1),
```
Martchus, line 425: Likely my last comment got lost. I still think these URLs should be rendered on the tab's page (and not the overall page).

Martchus (changes requested):

Martchus (approved):

PR #3994, opened 2021-06-30
jobs: Limit the maximum number of results

kalikiana:
Related: https://progress.opensuse.org/issues/94901

### DIFF
--- a/lib/OpenQA/Schema/ResultSet/Jobs.pm
+++ b/lib/OpenQA/Schema/ResultSet/Jobs.pm
@@ -94,10 +94,9 @@ MACHINE. For each set of dupes, only the latest job found is included in
 the return array.
 
 =cut
-sub latest_jobs {
-    my ($self, $until) = @_;
-
-    my @jobs = $self->search($until ? {t_created => {'<=' => $until}} : undef, {order_by => ['me.id DESC']});
+sub latest_jobs ($self, $until = undef, $limit = ~0 >> 1) {
+    my @jobs
+      = $self->search($until ? {t_created => {'<=' => $until}} : undef, {order_by => ['me.id DESC'], limit => $limit});
     my @latest;
     my %seen;
     foreach my $job (@jobs) {
--- a/lib/OpenQA/WebAPI/Controller/API/V1/Job.pm
+++ b/lib/OpenQA/WebAPI/Controller/API/V1/Job.pm
@@ -29,8 +29,6 @@ use DBIx::Class::Timestamps 'now';
 use Mojo::Asset::Memory;
 use Mojo::File 'path';
 
-use constant JOB_QUERY_LIMIT => 10000;
-
 =pod
 
 =head1 NAME
@@ -95,9 +93,9 @@ sub list {
     $validation->optional('scope')->in('current', 'relevant');
     $validation->optional('limit')->num(0);
     $validation->optional('latest')->num(1);
-
-    my $limit = $validation->param('limit') // JOB_QUERY_LIMIT;
-    return $self->render(json => {error => 'Limit exceeds maximum'}, status => 400) unless $limit <= JOB_QUERY_LIMIT;
+    my $limit = $validation->param('limit');
+    return $self->render(json => {error => 'Limit exceeds maximum'}, status => 400)
+      unless $limit <= $self->app->config->{'global'}->{'jobs_results_limit'};
     return $self->reply->validation_error({format => 'json'}) if $validation->has_error;
 
     my %args;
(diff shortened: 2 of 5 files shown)


perlpunk (approved):
Maybe the limit should be passed from outside of the db schema.
In `list_ajax` this is used:
https://github.com/os-autoinst/openQA/blob/master/lib/OpenQA/WebAPI/Controller/Test.pm#L80
`limit    => ($self->param('limit') // 500)`

Maybe we could use the same.
Maybe the default *and* a maximum limit should be configurable.
Also it could be possible to only allow authenticated users to specify a higher than default limit.

perlpunk (changes requested):
Sorry, clicked the wrong checkbox

lib/OpenQA/WebAPI/Controller/API/V1/Job.pm
```diff
@@ -95,9 +93,9 @@ sub list {
     $validation->optional('scope')->in('current', 'relevant');
     $validation->optional('limit')->num(0);
     $validation->optional('latest')->num(1);
-
-    my $limit = $validation->param('limit') // JOB_QUERY_LIMIT;
-    return $self->render(json => {error => 'Limit exceeds maximum'}, status => 400) unless $limit <= JOB_QUERY_LIMIT;
+    my $limit = $validation->param('limit');
```
okurz, line 96: why is it not necessary to set a default here anymore?

lib/OpenQA/WebAPI/Controller/Test.pm
```diff
@@ -71,13 +71,19 @@ sub get_match_param {
 sub list_ajax {
     my ($self) = @_;
 
+    my $validation = $self->validation;
+    $validation->optional('limit')->num(0);
+    my $limit = $validation->param('limit');
+    return $self->render(json => {error => 'Limit exceeds maximum'}, status => 400)
+      unless $limit <= $self->app->config->{'global'}->{'jobs_results_limit'};
```
okurz, line 78: these three lines are duplicated from the change in the API endpoint. Better extract into a function

lib/OpenQA/WebAPI/Controller/Test.pm
```diff
@@ -669,6 +675,10 @@ sub overview {
     my ($search_args, $groups) = $self->compose_job_overview_search_args;
     my $validation = $self->validation;
     $validation->optional('t')->datetime;
+    my $limit = $validation->param('limit');
+    $validation->optional('limit')->num(0, $self->app->config->{'global'}->{'jobs_results_limit'});
+    return $self->render(json => {error => 'Limit exceeds maximum'}, status => 400)
+      unless $limit <= $self->app->config->{'global'}->{'jobs_results_limit'};
```
okurz, line 681: same duplication here

okurz (changes requested):

lib/OpenQA/Schema/ResultSet/Jobs.pm
```diff
@@ -94,10 +94,9 @@ MACHINE. For each set of dupes, only the latest job found is included in
 the return array.
 
 =cut
-sub latest_jobs {
-    my ($self, $until) = @_;
-
-    my @jobs = $self->search($until ? {t_created => {'<=' => $until}} : undef, {order_by => ['me.id DESC']});
+sub latest_jobs ($self, $until = undef, $limit = ~0 >> 1) {
```
Martchus, line 97: Just make the default `undef` here. I suppose DBIx can cope with it. Otherwise I'd rather add the `limit` only to the hash if `$limit` is undef.

(Btw, `~0 >> 1` should be the highest possible integer (Perl's internal integer type can represent) divided by 2 (possibly rounded down). This might even be platform-dependent and doesn't look like a very meaningful default to pass to the database.)

Martchus (changes requested):

lib/OpenQA/WebAPI/Controller/Test.pm
kalikiana: I don't know... do we want the same behavior everywhere? We don't have a `limit` everywhere

lib/OpenQA/Schema/ResultSet/Jobs.pm
kalikiana: Yeah, I basically wanted "not specified". Mostly something to test it out with Tina's input. If we agree on an approach I'll see how to do it properly

lib/OpenQA/WebAPI/Controller/API/V1/Job.pm
kalikiana: Do we want each route to have a `limit` argument, a default limit and a maximum? Currently they're not consistent.

lib/OpenQA/WebAPI/Controller/Test.pm
okurz: I don't know if we want the same behaviour everywhere but to my eyes the above looks like a verbatim copy

PR #3995, opened 2021-06-30
Add link to /test/overview from index page

ilausuch:
Provides link to /tests/overview of latest builds of all job groups
that are listed in the index page

https://progress.opensuse.org/issues/94732

Example in the index page

image

Example in the parent group page

image

### DIFF
--- a/assets/stylesheets/test-details.scss
+++ b/assets/stylesheets/test-details.scss
@@ -415,3 +415,16 @@ ul.modcategory {
         opacity: inherit;
     }
 }
+
+.test-result-overview-link {
+    position: relative;
+    top: -0.1em;
+    left: 0.2em;
+    font-size: 75%;
+    i {
+        color: $link-color;
+    }
+    i:hover {
+        color: $link-hover-color;
+    }
+}
--- a/t/22-dashboard.t
+++ b/t/22-dashboard.t
@@ -107,7 +107,6 @@ is_deeply(
     ['opensuse test', 'Test parent', $test_overview_tooltip],
     'parent group shown and opensuse is no more on top-level'
 );
-$t->element_exists('#test_result_overview_link_1', 'parent group has link to tests/overview');
 my $tests_overview_dashboard = $t->tx->res->dom->find("#test_result_overview_link_1")->first;
 is(
     $tests_overview_dashboard->attr('href'),
@@ -210,8 +209,7 @@ for my $url (@urls) {
 # parent group overview
 $t->get_ok('/parent_group_overview/' . $test_parent->id)->status_is(200);
 check_test_parent('expanded');
-$t->element_exists('#test_result_overview_link');
-my $tests_overview = $t->tx->res->dom->find("#test_result_overview_link")->first;
+my $tests_overview = $t->tx->res->dom->find("#test_result_overview_link_1")->first;
 is(
     $tests_overview->attr("href"),
     "/tests/overview?groupid=1001&groupid=1002",
--- a/t/22-dashboard.t
+++ b/t/22-dashboard.t
@@ -100,8 +100,20 @@ like(
 );
 
 $t->get_ok('/dashboard_build_results')->status_is(200);
-@h2 = $t->tx->res->dom->find('h2 a')->map('text')->each;
-is_deeply(\@h2, ['opensuse test', 'Test parent'], 'parent group shown and opensuse is no more on top-level');
+@h2 = $t->tx->res->dom->find('h2 a')->map(sub ($e) { $e->text . ($e->attr('title') // '') })->each;
+my $test_overview_tooltip = 'Shows the latest test results for all job groups within this parent job group';
+is_deeply(
+    \@h2,
+    ['opensuse test', 'Test parent', $test_overview_tooltip],
+    'parent group shown and opensuse is no more on top-level'
+);
+$t->element_exists('#test_result_overview_link_1', 'parent group has link to tests/overview');
+my $tests_overview_dashboard = $t->tx->res->dom->find("#test_result_overview_link_1")->first;
+is(
+    $tests_overview_dashboard->attr('href'),
+    '/tests/overview?groupid=1001',
+    'The "test result overview" anchor href points to /test/overview and includes all the groupids for group 1'
+);
 
 my @h4 = $t->tx->res->dom->find('div.children-collapsed .h4 a')->map('text')->each;
 is_deeply(\@h4, [qw(Build87.5011 Build0048@0815 Build0048)], 'builds on parent-level shown, sorted first by version');
@@ -123,8 +135,12 @@ $opensuse_test_group->update({parent_id => $test_parent->id});
 $opensuse_group->jobs->find({BUILD => '0048@0815'})->comments->create({text => 'poo#1234', user_id => 99901});
 
 $t->get_ok('/dashboard_build_results?limit_builds=20&show_tags=1')->status_is(200);
-@h2 = $t->tx->res->dom->find('h2 a')->map('text')->each;
-is_deeply(\@h2, ['Test parent'], 'only parent shown, no more top-level job groups');
+@h2 = $t->tx->res->dom->find('h2 a')->map(sub ($e) { $e->text . ($e->attr('title') // '') })->each;
+is_deeply(
+    \@h2,
+    ['Test parent', $test_overview_tooltip],
+    'only link to parent (and related overview) shown, no more top-level job groups'
+);
 
 sub check_test_parent {
     my ($default_expanded) = @_;
--- a/templates/webapi/main/dashboard_build_results.html.ep
+++ b/templates/webapi/main/dashboard_build_results.html.ep
@@ -6,6 +6,7 @@
     % if (@{$groupresults->{children}}) {
         <h2>
             %= link_to $group->{name} => url_for('parent_group_overview', groupid => $group->{id})
+            %= include 'main/test_result_overview_link', group_list => $groupresults->{children}, groupid => $group->{id}
         </h2>
         %= include 'main/group_builds', build_results => $build_results, group => $group, max_jobs => $max_jobs, children => $groupresults->{children}, default_expanded => $default_expanded
     % } else {
--- a/templates/webapi/main/parent_group_overview.html.ep
+++ b/templates/webapi/main/parent_group_overview.html.ep
@@ -19,6 +19,7 @@
 
 <h2 class="mb-4">
     Last Builds for <%= $group->{name} %>
+    %= include 'main/test_result_overview_link', group_list => $children, groupid => $group->{id}
 </h2>
 
 <p>
--- a/templates/webapi/main/test_result_overview_link.html.ep
+++ b/templates/webapi/main/test_result_overview_link.html.ep
@@ -1,0 +1,5 @@
+<a href="<%= url_for('tests_overview')->query(map { (groupid => $_->{id}) } @$group_list); %>"
+  title="Shows the latest test results for all job groups within this parent job group"
+  id="test_result_overview_link_<%= $groupid %>"
+  class="test-result-overview-link"
+><i class="fa fa-list"></i></a>


templates/webapi/main/dashboard_build_results.html.ep
```diff
@@ -15,3 +15,14 @@
         %= include 'main/group_builds', build_results => $build_results, group => $group, max_jobs => $max_jobs, children => undef, default_expanded => 0
     % }
 % }
+
+% if (@$results) {
+<p>
+  <a href="<%= url_for('tests_overview')->query(todo => 1, map { (groupid => $_->{group}->{id}) } @$results); %>"
```
Martchus, line 21: Don't duplicate the code from the parent group overview here. You can create a separate template file and include it here and in the parent group overview.

Martchus (changes requested):

templates/webapi/main/dashboard_build_results.html.ep
ilausuch: But the query is not the same, don't you think this adds complexity?

ilausuch: I am thinking, that the url has to be passed from outside

Martchus: The other URL is generated via `url_for('tests_overview')->query(todo => 1, map { (groupid => $_->{id}) } @$children)` so only the name of the array is different. I suggest to simply pass the array as parameter when including the template.

By the way, the title also needs to be different as well (e.g. also using a parameter).

Martchus:
And before merging this it would be good to have a screenshot to see how it looks like. (We should avoid disturbing the index page with too many links.)

perlpunk:
This now creates a link with *all* groups below the dashboard.
I think the ticket means that we want the link only for groups with a parent group, so it could be put somewhere after the `<h2>` with the parent job group title, inside of the already existing loop over $results.

okurz (changes requested):
What Martchus and perlpunk said

ilausuch:
Remains to a common template

t/22-dashboard.t
```diff
@@ -26,6 +26,7 @@ use OpenQA::Test::Case;
 use OpenQA::Test::TimeLimit '10';
 use OpenQA::Jobs::Constants;
 use Mojo::File qw(tempfile);
+use Data::Dumper qw(Dumper);
```
okurz, line 29: should not be checked in

okurz (changes requested):
Please also provide screenshots in the PR description

t/22-dashboard.t
```diff
@@ -100,8 +101,15 @@ like(
 );
 
 $t->get_ok('/dashboard_build_results')->status_is(200);
-@h2 = $t->tx->res->dom->find('h2 a')->map('text')->each;
-is_deeply(\@h2, ['opensuse test', 'Test parent'], 'parent group shown and opensuse is no more on top-level');
+@h2 = $t->tx->res->dom->find('h2 a')->map(sub ($e) { $e->text . ($e->attr('title') // '') })->each;
+is_deeply(\@h2, ['opensuse test', 'Test parent', 'Shows the latest test results for all job groups within this parent job group'], 'parent group shown and opensuse is no more on top-level');
```
Martchus, line 104: The test description here should be changed as well.

t/22-dashboard.t
```diff
@@ -123,8 +131,8 @@ $opensuse_test_group->update({parent_id => $test_parent->id});
 $opensuse_group->jobs->find({BUILD => '0048@0815'})->comments->create({text => 'poo#1234', user_id => 99901});
 
 $t->get_ok('/dashboard_build_results?limit_builds=20&show_tags=1')->status_is(200);
-@h2 = $t->tx->res->dom->find('h2 a')->map('text')->each;
-is_deeply(\@h2, ['Test parent'], 'only parent shown, no more top-level job groups');
+@h2 = $t->tx->res->dom->find('h2 a')->map(sub ($e) { $e->text . ($e->attr('title') // '') })->each;
+is_deeply(\@h2, ['Test parent', 'Shows the latest test results for all job groups within this parent job group'], 'only link to parent (and related overview) shown, no more top-level job groups');
```
Martchus, line 134: Maybe it makes sense to store the long string in a variable, e.g. `my $test_overview_tooltip = …` and use it in both occurrences.

assets/stylesheets/test-details.scss
```diff
@@ -414,4 +414,14 @@ ul.modcategory {
...
+}
+
+.test-result-overview-link {
+    top: -0.1em;
+    position: relative;
+    left: 0.2em;
+    opacity: 0.8;
+    font-size:medium;
+    color: #446E9B;
+    text-decoration: underline dotted;
```
Martchus: This style has no effect but would look bad anyways so just remove the line.

assets/stylesheets/test-details.scss
```diff
@@ -414,4 +414,14 @@ ul.modcategory {
...
         opacity: inherit;
     }
+}
+
+.test-result-overview-link {
+    top: -0.1em;
+    position: relative;
+    left: 0.2em;
+    opacity: 0.8;
+    font-size:medium;
```
Martchus, line 424: Don't set the size to a possibly user-defined default. It might not fit because in all other places we don't use such values. Just make it relative to the size of the heading:

```suggestion
    font-size: font-size: 75%;;
```

Also mind the coding style (the space was missing).

t/22-dashboard.t
ilausuch: ups sorry

assets/stylesheets/test-details.scss
```diff
@@ -414,4 +414,14 @@ ul.modcategory {
     .fa-download:hover {
         opacity: inherit;
     }
+}
+
+.test-result-overview-link {
+    top: -0.1em;
+    position: relative;
+    left: 0.2em;
+    opacity: 0.8;
```
Martchus, line 423: Setting this back to `initial` on hover would likely look good.

assets/stylesheets/test-details.scss
```diff
@@ -414,4 +414,14 @@ ul.modcategory {
     .fa-download:hover {
         opacity: inherit;
     }
+}
+
+.test-result-overview-link {
+    top: -0.1em;
```
Martchus, line 420: Can you move that one down so the relative coordinates are specified consecutively?

assets/stylesheets/test-details.scss
```diff
@@ -414,4 +414,14 @@ ul.modcategory {
...
+    position: relative;
+    left: 0.2em;
+    opacity: 0.8;
+    font-size:medium;
+    color: #446E9B;
```
Martchus: Don't copy & paste code. SCSS allows to avoid that. However, in this case just get rid of the line as it doesn't make a difference anyways.

t/22-dashboard.t
```diff
@@ -100,8 +101,15 @@ like(
 );
 
 $t->get_ok('/dashboard_build_results')->status_is(200);
-@h2 = $t->tx->res->dom->find('h2 a')->map('text')->each;
-is_deeply(\@h2, ['opensuse test', 'Test parent'], 'parent group shown and opensuse is no more on top-level');
+@h2 = $t->tx->res->dom->find('h2 a')->map(sub ($e) { $e->text . ($e->attr('title') // '') })->each;
+is_deeply(\@h2, ['opensuse test', 'Test parent', 'Shows the latest test results for all job groups within this parent job group'], 'parent group shown and opensuse is no more on top-level');
+$t->element_exists('#test_result_overview_link_1');
```
Martchus, line 105: This test has no description but it is useless anyways because if that one fails the next test would fail as well. (It would make more sense to check whether `$tests_overview_dashboard` before blindly calling a method of it.)

t/22-dashboard.t
```diff
@@ -100,8 +101,15 @@ like(
...
+my $tests_overview_dashboard = $t->tx->res->dom->find("#test_result_overview_link_1")->first;
+is(
+    $tests_overview_dashboard->attr("href"),
+    "/tests/overview?todo=1&groupid=1001",
```
Martchus, line 109: Prefer single quotes if no interpolation is needed.

Martchus (changes requested):

templates/webapi/main/dashboard_build_results.html.ep
ilausuch: WIP

assets/stylesheets/test-details.scss
```diff
@@ -414,4 +414,16 @@ ul.modcategory {
...
+
+.test-result-overview-link {
+    top: -0.1em;
+    position: relative;
+    left: 0.2em;
+    opacity: 0.8;
+    font-size:medium;
+    i {
+        color: #446E9B;
+        text-decoration: underline dotted;
```
Martchus, line 427: Ok, now that you've moved these lines here they make a difference. However, I still thinks the text decoration looks bad. You can keep the color if you like (I personally like to keep font awesome's default) but then please don't repeat the color code but use `$link-color` instead. (Btw, these colors are defined in `openqa_theme.scss`.)

Martchus (changes requested):

assets/stylesheets/test-details.scss
ilausuch: I understand this comment is outdated because your last comment. Isn't it?

t/22-dashboard.t
ilausuch: done

assets/stylesheets/test-details.scss
ilausuch: done

assets/stylesheets/test-details.scss
ilausuch: done

t/22-dashboard.t
ilausuch: Do you think this new version is better?

t/22-dashboard.t
ilausuch: done

assets/stylesheets/test-details.scss
ilausuch: Done

templates/webapi/main/dashboard_build_results.html.ep
ilausuch: Done

assets/stylesheets/test-details.scss
ilausuch: Done, as far as I understood

assets/stylesheets/test-details.scss
ilausuch: done

okurz (approved):
I like it :+1:

assets/stylesheets/test-details.scss
okurz: I think it looks good now

t/22-dashboard.t
okurz: New version looks better

t/22-dashboard.t
```diff
@@ -100,8 +100,21 @@ like(
...
-@h2 = $t->tx->res->dom->find('h2 a')->map('text')->each;
-is_deeply(\@h2, ['opensuse test', 'Test parent'], 'parent group shown and opensuse is no more on top-level');
+@h2 = $t->tx->res->dom->find('h2 a')->map(sub ($e) { $e->text . ($e->attr('title') // '') })->each;
+my $test_overview_tooltip = 'Shows the latest test results for all job groups within this parent job group';
+is_deeply(
+    \@h2,
+    ['opensuse test', 'Test parent', $test_overview_tooltip],
+    'parent group shown and opensuse is no more on top-level'
+);
+$t->element_exists('#test_result_overview_link_1', 'The parent group should has the link to tests/overview') and do {
```
kraih, line 110: The `and do {` here doesn't really serve a purpose, `element_exists` always return `$t` again.

t/22-dashboard.t
```diff
@@ -194,13 +211,14 @@ for my $url (@urls) {
...
 $t->get_ok('/parent_group_overview/' . $test_parent->id)->status_is(200);
 check_test_parent('expanded');
-$t->element_exists('#test_result_overview_link');
-my $tests_overview = $t->tx->res->dom->find("#test_result_overview_link")->first;
-is(
-    $tests_overview->attr("href"),
-    "/tests/overview?todo=1&groupid=1001&groupid=1002",
-    "The 'test result overview' anchor href points to /test/overview and includes all the groupids"
-);
+$t->element_exists('#test_result_overview_link_1', 'exists the link to tests/overview') and do {
```
kraih, line 214: Same here for the `and do {`.

t/22-dashboard.t
perlpunk: Ah. that was my fault, as I suggested that code.
I'm more used to Test::More style functions, which return success or failure.

ilausuch: Thanks @perlpunk any way :)
Removed

kalikiana:
FYI fix for `t/ui/12-needle-edit.t` in #4001

okurz:
#4001 is not a fix and it's unlikely to be acceptable quite soon as we have more test failures blocking us.

kalikiana:
> #4001 is not a fix and it's unlikely to be acceptable quite soon as we have more test failures blocking us.

The point is that it addresses failures seen in this PR and you don't have to investigate it from scratch.

templates/webapi/main/test_result_overview_link.html.ep
```diff
@@ -0,0 +1,5 @@
+<a href="<%= url_for('tests_overview')->query(todo => 1, map { (groupid => $_->{id}) } @$group_list); %>"
```
okurz, line 1: I just realized now on OSD in production that this query includes "todo=1" which was unexpected. The link says "test overview", not "all tests needing review" so please delete that query parameter part in its own commit please.

okurz (changes requested):

t/22-dashboard.t
```diff
@@ -100,8 +100,20 @@ like(
...
-@h2 = $t->tx->res->dom->find('h2 a')->map('text')->each;
-is_deeply(\@h2, ['opensuse test', 'Test parent'], 'parent group shown and opensuse is no more on top-level');
+@h2 = $t->tx->res->dom->find('h2 a')->map(sub ($e) { $e->text . ($e->attr('title') // '') })->each;
+my $test_overview_tooltip = 'Shows the latest test results for all job groups within this parent job group';
+is_deeply(
+    \@h2,
+    ['opensuse test', 'Test parent', $test_overview_tooltip],
+    'parent group shown and opensuse is no more on top-level'
+);
+$t->element_exists('#test_result_overview_link_1', 'The parent group should has the link to tests/overview');
```
Martchus, line 110: "should has" is wrong, either rephrase it like
```suggestion
$t->element_exists('#test_result_overview_link_1', 'parent group has link to tests/overview');
```
or remove the check altogether because the subsequent check will fail anyways (as the method `attr` would be called on `undef`).

t/22-dashboard.t
```diff
@@ -194,8 +210,8 @@ for my $url (@urls) {
 # parent group overview
 $t->get_ok('/parent_group_overview/' . $test_parent->id)->status_is(200);
 check_test_parent('expanded');
-$t->element_exists('#test_result_overview_link');
-my $tests_overview = $t->tx->res->dom->find("#test_result_overview_link")->first;
+$t->element_exists('#test_result_overview_link_1', 'exists the link to tests/overview');
```
Martchus, line 213: The same counts as for the previous occurrence. (Here it should be "link to tests/overview exists" but like I said in the other comment, you can remove these checks completely.)

templates/webapi/main/dashboard_build_results.html.ep
```diff
@@ -6,6 +6,7 @@
     % if (@{$groupresults->{children}}) {
         <h2>
             %= link_to $group->{name} => url_for('parent_group_overview', groupid => $group->{id})
+            %= include 'main/test_result_overview_link', group_list => \@{$groupresults->{children}}, groupid => $group->{id}
```
Martchus, line 9: Dereferencing the array ref via `@{…}` just to take the reference again via `\` makes no sense.
```suggestion
            %= include 'main/test_result_overview_link', group_list => $groupresults->{children}, groupid => $group->{id}
```

templates/webapi/main/parent_group_overview.html.ep
```diff
@@ -19,17 +19,9 @@
 
 <h2 class="mb-4">
     Last Builds for <%= $group->{name} %>
+    %= include 'main/test_result_overview_link', group_list => \@$children, groupid => $group->{id}
```
Martchus, line 22: Same as in the previous occurrence:

```suggestion
    %= include 'main/test_result_overview_link', group_list => $children, groupid => $group->{id}
```

templates/webapi/main/test_result_overview_link.html.ep
Martchus, line 1: Looks like you've lost the paragraph here. If that's fine then just ignore/resolve the comment. (I'm just noticing it.)

Martchus (changes requested):

templates/webapi/main/test_result_overview_link.html.ep
ilausuch: Done

t/22-dashboard.t
ilausuch: Done

templates/webapi/main/dashboard_build_results.html.ep
ilausuch: done, thanks

templates/webapi/main/parent_group_overview.html.ep
ilausuch: Done

templates/webapi/main/test_result_overview_link.html.ep
okurz: Please in a separate commit as you effectively fix the problem that is apparent in the already merged and deployed state

okurz: Now handled in https://github.com/os-autoinst/openQA/pull/4013 explicitly to be applied as fix first.

Martchus: I now see myself that having a paragraph makes no sense anymore.

Martchus (approved):

templates/webapi/main/test_result_overview_link.html.ep
ilausuch: Rebased from Master

okurz (approved):

PR #3997, opened 2021-07-01
fullstack test: Stop when first test incompletes

Martchus:
* Also show the `autoinst-log.txt` of the test run.
* Previously it waited until the timeout was exceeded for the test to be
  running although it has already been incompleted anyways.


asdil12 (approved):
Now it fails earlier with the same qemu error as discussed on RC

okurz (approved):

Martchus:
See #3998 for the fix of the actual error.

PR #3999, opened 2021-07-01
Remove uncovered conditions for internal search of job settings

okurz:
Maybe we are just missing tests for this, maybe we do not need it at
all.

I have also tried to trigger the additional query condition over
/tests/overview, e.g.
https://openqa.opensuse.org/tests/overview?distri=opensuse&groupid=1&version=Tumbleweed&HDD_1=opensuse-Tumbleweed-x86_64-20210629-textmode@64bit.qcow2
but to no effect. Tests that do not have that HDD_1 setting still show
up so the filter is not effective.

See https://codecov.io/gh/os-autoinst/openQA/src/master/lib/OpenQA/Schema/ResultSet/JobSettings.pm

### DIFF
--- a/lib/OpenQA/Schema/ResultSet/JobSettings.pm
+++ b/lib/OpenQA/Schema/ResultSet/JobSettings.pm
@@ -1,4 +1,4 @@
-# Copyright (C) 2016 SUSE LLC
+# Copyright (C) 2016-2021 SUSE LLC
 #
 # This program is free software; you can redistribute it and/or modify
 # it under the terms of the GNU General Public License as published by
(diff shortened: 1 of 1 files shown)


lib/OpenQA/Schema/ResultSet/JobSettings.pm
```diff
@@ -39,24 +39,16 @@ sub query_for_settings {
...
-            }
-            my $setting_value = ($args->{$setting} =~ /^:\w+:/) ? {'like', "$&%"} : $args->{$setting};
-            push(
-                @conds,
-                {
-                    "$tname.key"   => $setting,
-                    "$tname.value" => $setting_value
-                });
-        }
+        next unless if $args->{$setting};
```
Martchus, line 42: The `if` is wrong :-)

lib/OpenQA/Schema/ResultSet/JobSettings.pm
```diff
@@ -39,24 +39,16 @@ sub query_for_settings {
...
+        my $tname = 'me';
+        my $setting_value = ($args->{$setting} =~ /^:\w+:/) ? {'like', "$&%"} : $args->{$setting};
+        push(
+            @conds,
+            {
+                "$tname.key"   => $setting,
+                "$tname.value" => $setting_value
+            });
     }
     return $self->search({-and => \@conds}, {join => \@joins});
```
Martchus, line 53: The new code doesn't populate `@joins` at all. Intuitively I would say that cannot be right. (And if it was right than `{join => \@joins}` can be removed completely.)

Martchus (changes requested):

lib/OpenQA/Schema/ResultSet/JobSettings.pm
okurz: Right. I removed all references to `@joins` because according to the coverage analysis we actually never call it in tests. And as explained in https://github.com/os-autoinst/openQA/pull/3999/commits/e0e2e6646bb5900b6c8d375a2ea7e675ce0b0ebe I did a manual test – which might be wrong – to check if it's used. Maybe some of you have an idea how we would trigger the code that adds joins?

lib/OpenQA/Schema/ResultSet/JobSettings.pm
okurz: fixed, thx

Martchus (approved):
I guess it is really unused (see my comment on the other PR which also contains this change).

PR #4000, opened 2021-07-01
Do not fail Minion jobs when asset download fails due to a client error

Martchus:
Otherwise we'll get lots of failed Minion jobs which usually means
something is broken on our side. However, client errors are most likely
caused by users, e.g. by specifying an invalid asset URL which just
returns a 404 page. This change prevents unnecessary alerts in such cases.


okurz:
happy PR #4000 \o/ :fireworks: :partying_face: :fireworks:

okurz (commented):
+1

okurz (approved):
Now reading your comment as reaction to the alert we had for OSD I understand your motivation much better :)

PR #4001, opened 2021-07-01
Mark t/ui/12-needle-edit.t as unstable after recent CI failures

okurz:
Related progress issue: https://progress.opensuse.org/issues/95009

### DIFF
--- a/.circleci/unstable_tests.txt
+++ b/.circleci/unstable_tests.txt
@@ -1,4 +1,5 @@
 t/25-cache-service.t
 t/ui/01-list.t
 t/ui/26-jobs_restart.t
+t/ui/12-needle-edit.t
 t/ui/13-admin.t


perlpunk (approved):

perlpunk:
Looking at the failing job, only `t/ui/26-jobs_restart.t` seems to be executed (and fails) in the unstable now

okurz:
> Looking at the failing job, only `t/ui/26-jobs_restart.t` seems to be executed (and fails) in the unstable now

Right. I have reported https://progress.opensuse.org/issues/95024

kalikiana (approved):

ilausuch (approved):

perlpunk (dismissed):
I think this isn't the right thing to do.

We are seeing timeouts in several tests since yesterday. All seem to be related to timeouts, and an earlier commit pushed has the same problems.

Just marking all of them as unstable wouldn't make sense I believe.

kraih (approved):

okurz:
> I think this isn't the right thing to do.
> 
> We are seeing timeouts in several tests since yesterday. All seem to be related to timeouts, and an earlier commit pushed has the same problems.
> 
> Just marking all of them as unstable wouldn't make sense I believe.

Ok, that means we block all pull requests until we can solve this problem. I am fine with that unless we need to fix another urgent problem.

Martchus (changes requested):
The test failure doesn't look specific to `t/ui/12-needle-edit.t`. It looks more like a general problem of our UI tests and I think I have seen this kind of error already elsewhere. It could even be just a timeout with a very messed-up error message and in the last few days (from last Thursday) CircleCI was just extraordinarily slow.

perlpunk:
This was solved by #4009

PR #4003, opened 2021-07-02
Fix performance issue with `/tests/overview?...&flavor=`

kraih:
As discussed earlier in the weekly meeting, just the most minimal validation (not even actual validation, just filtering of parameters). It prevents `/tests/overview?...&flavor=` from affecting the database query and loading 70k+ jobs on O3, instead of just jobs from the latest build. Right now this is unlikely to happen, but we want to extend the filter form with new options, including `flavor`.

Progress: https://progress.opensuse.org/issues/94901

### DIFF
--- a/lib/OpenQA/WebAPI/Plugin/Helpers.pm
+++ b/lib/OpenQA/WebAPI/Plugin/Helpers.pm
@@ -14,17 +14,14 @@
 # with this program; if not, see <http://www.gnu.org/licenses/>.
 
 package OpenQA::WebAPI::Plugin::Helpers;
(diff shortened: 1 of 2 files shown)


kraih:
Not ready, since Circle CI still has issues and i can't confirm that everything works as it should.

okurz (approved):
Looks all good from the commits and diff. Of course we should await full test results before merging

okurz (approved):

perlpunk (approved):

kalikiana (approved):

PR #4004, opened 2021-07-02
Use a more feasible timeout for t/full-stack.t

asdil12:
I measured 2m 40s for this module so the internal timeout
should be at least 200s instead of 90.
When running in CI the default factor is 3 so the make level
timeout should be 12m to provide some selectivity over the
200s*3=10m internal timeout.

Ticket: https://progress.opensuse.org/issues/93677

### DIFF
--- a/t/full-stack.t
+++ b/t/full-stack.t
@@ -56,7 +56,7 @@ use OpenQA::Test::Utils
   qw(create_websocket_server create_live_view_handler setup_share_dir),
   qw(cache_minion_worker cache_worker_service mock_service_ports setup_fullstack_temp_dir),
   qw(start_worker stop_service wait_for_or_bail_out);
-use OpenQA::Test::TimeLimit '90';
+use OpenQA::Test::TimeLimit '200';
 use OpenQA::Test::FullstackUtils;
 
 plan skip_all => 'set FULLSTACK=1 (be careful)'                                 unless $ENV{FULLSTACK};


okurz:
Wow, I can agree with the change on make level but from 90s to 200s sounds like something else is broken within the full stack test. Could you please conduct multiple runs locally to check? Maybe you don't have a KVM capable local environment which can cause such slow-down?

asdil12:
The test did actually succeed - and kvm seems to actually work on that machine even as a normal user.

But maybe the machine is a bit outdated:

```
Architecture:                    x86_64
CPU op-mode(s):                  32-bit, 64-bit
Byte Order:                      Little Endian
Address sizes:                   36 bits physical, 48 bits virtual
CPU(s):                          8
On-line CPU(s) list:             0-7
Thread(s) per core:              2
Core(s) per socket:              4
Socket(s):                       1
NUMA node(s):                    1
Vendor ID:                       GenuineIntel
CPU family:                      6
Model:                           42
Model name:                      Intel(R) Core(TM) i7-2600 CPU @ 3.40GHz
Stepping:                        7
CPU MHz:                         1600.000
CPU max MHz:                     3800.0000
CPU min MHz:                     1600.0000
BogoMIPS:                        6784.54
Virtualization:                  VT-x
L1d cache:                       128 KiB
L1i cache:                       128 KiB
L2 cache:                        1 MiB
L3 cache:                        8 MiB
NUMA node0 CPU(s):               0-7
Vulnerability Itlb multihit:     KVM: Mitigation: VMX disabled
Vulnerability L1tf:              Mitigation; PTE Inversion; VMX conditional cache flushes, SMT vulnerable
Vulnerability Mds:               Mitigation; Clear CPU buffers; SMT vulnerable
Vulnerability Meltdown:          Mitigation; PTI
Vulnerability Spec store bypass: Mitigation; Speculative Store Bypass disabled via prctl and seccomp
Vulnerability Spectre v1:        Mitigation; usercopy/swapgs barriers and __user pointer sanitization
Vulnerability Spectre v2:        Mitigation; Full generic retpoline, IBPB conditional, IBRS_FW, STIBP conditional, RSB filling
Vulnerability Srbds:             Not affected
Vulnerability Tsx async abort:   Not affected
Flags:                           fpu vme de pse tsc msr pae mce cx8 apic sep mtrr pge mca cmov pat pse36 clflush dts acpi mmx fxsr sse sse2 ht tm pbe syscall nx rdtscp lm constant_tsc arch_perfmon pebs bts rep_good nopl xtopology nonstop_tsc cpuid aperfmperf pni pclmulqdq dtes64 monitor ds_cpl vmx smx est tm2 ssse3 cx16 xtpr pdcm pcid sse4_1 sse4_2 x2apic popcnt tsc_deadline_timer aes xsave avx lahf_lm epb pti ssbd ibrs ibpb stibp tpr_shadow vnmi flexpriority ept vpid xsaveopt dtherm ida arat pln pts md_clear flush_l1d

okurz:
Well, then likely something else is broken within the full stack test. Could you please conduct multiple runs locally to check?

asdil12:
I did another few tests and they always took about 2min 40s.

I uploaded a log [here](https://gist.github.com/asdil12/c8169bd0b284f47b131e8d262070796a) but didn't see any obvious errors that could have slowed things down.

okurz (commented):
Please rebase and remove the Makefile level update after we have https://github.com/os-autoinst/openQA/pull/4009

okurz (approved):

Martchus (commented):
I wouldn't expect the `2600` to be *that* much slower. I have a `6820HQ` here and it runs the test in under 90 seconds. Compared to the `2500k` of my private workstation it generally isn't *that* much faster (and the `2500k` should be slower than your `2600`).

~~In your CPU info output it says `Vulnerability Itlb multihit:     KVM: Mitigation: VMX disabled` which means that hardware virtualization support on that machine is disabled. That's likely the culprit.~~

~~So whether we want to merge this PR or not likely depends on the question of whether we want the timeout to rely on the availability of hardware virtualization. In any case, it would be great if there was a comment in the code stating and explanation (e.g. `# needs to be that high in case hardware virtualization is not available`).~~

Looks like I have `Vulnerability Itlb multihit:     KVM: Mitigation: VMX disabled` on my system as well but I'm pretty sure that hardware virtualization is working here. So likely not the problem after all. But then a comment like `# can be significantly slower on older systems` would make sense. Otherwise someone might be tempted to decrease the timeout again in the future.

okurz:
@Martchus thanks for crosschecking yourself. I tested myself and now unfortunately the full stack test is also slower for me. I guess it's again time for a separate ticket to look into the slowness of this particular test.

Martchus (approved):

PR #4005, opened 2021-07-02
Add filter options for flavor and test

kraih:
Just flavor and test for now, since i'm not sure how to best fit job groups into the form without making it look like a total mess.

form

This should also not be merged before #4003, since the form can create empty flavor values.

Progress: https://progress.opensuse.org/issues/91647

### DIFF
--- a/t/ui/10-tests_overview.t
+++ b/t/ui/10-tests_overview.t
@@ -149,7 +149,8 @@ is(scalar @filtered_out, 0, 'result filter correctly applied');
 
 # Test whether all URL parameter are passed correctly
 my $url_with_escaped_parameters
-  = $baseurl . 'tests/overview?arch=&machine=&modules=&distri=opensuse&build=0091&version=Staging%3AI&groupid=1001';
+  = $baseurl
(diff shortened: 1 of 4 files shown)


okurz (approved):

t/ui/10-tests_overview.t
```diff
@@ -204,17 +205,55 @@ subtest 'filtering by architecture' => sub {
...
+    };
+
+    subtest 'filter for specific test' => sub {
+        $driver->find_element('#filter-panel .card-header')->click();
+        $driver->find_element('#filter-test')->send_keys('textmode');
+        $driver->find_element('#filter-panel .btn-default')->click();
+
+        my @rows = $driver->find_elements('#content tbody tr');
+        is(scalar @rows, 1, 'exactly one row present');
+        like($rows[0]->get_text(), qr/textmode/, 'test is textmode');
```
kalikiana, line 255: This leads to Perl errors if `is` fails and `@rows` is empty iirc. `find_element('#content tbody tr:only-child')` might be better?

kalikiana (approved):
Note to self, hit approve when providing optional suggestions

t/ui/10-tests_overview.t
kraih: Honestly, i just copied it from another test. 😉

PR #4006, opened 2021-07-02
Fix module source content negotiation for Mojo < 9

AdamWill:
The 'fix' in 30571b5 breaks things for Mojo < 9 (which we still
have compatibility with elsewhere in this file).


AdamWill:
I tested this, it works.

okurz (approved):

kraih (commented):
I disagree with applying these patches, since they have no test coverage in our current CI setup.

Martchus (commented):
I'd also prefer not having them on master as we lack test coverage. (Not only for this particular line but openQA generally doesn't support older Mojolicious versions - and we shouldn't add code to leads to the impression that it is.)

okurz:
These are good arguments.

@AdamWill unless you have a good idea how to ensure we have test coverage for the mentioned code paths how about maintaining your patches in patch files and either have them in this repo here or downstream in the Linux distribution package itself?

AdamWill:
Again, I'm kinda confused by the apparent flip-flopping here. When I [first raised this problem](https://github.com/os-autoinst/openQA/pull/3830#issuecomment-834803586), @okurz [filed a ticket](https://github.com/os-autoinst/openQA/issues/3892), I sent [a PR](https://github.com/os-autoinst/openQA/pull/3898) where I outlined the other possible approaches, and the PR got merged.

But ever since then, anything along the same lines gets this kind of response. It doesn't seem at all consistent. If you didn't want this, why not just say so on the first comment or first issue or first PR? It would've been much clearer. Why not remove the other compatibility lines in the file?

Use with Mojo 8 is getting tested by Fedora package builds. They run the test suite. If it fails I'll know (that's exactly how I knew to send this fix). There are other 'uncoverable' lines in the code. I can mark these as uncoverable if that's the issue.

I *can* carry the patches downstream too. It's not a problem exactly. It just would've been nice to come to that conclusion in the first place instead of this sudden seeming u-turn.

okurz:
Yes, would have been nice to know all that in before. Please assume good intent. We are human and we fail. Also different people have different opinions, e.g. kraih has a different opinion here then me which should be ok. I would be ok with "uncoverable" statements.

I still keep my "approval" and both Martchus and kraih only expressed "soft disagreement" with comments, not blocking approval or merge. As soon as a second person would approve *and* tests pass the PR would be merged

kraih:
If we decide compatibility with Mojolicious 8 is important, it should have test coverage.

okurz:
> If we decide compatibility with Mojolicious 8 is important, it should have test coverage.

Well, would be nice to have test coverage for that in our repo but we do not even have test coverage for 4% of other code in our repo. Also as Adam stated they actually have test coverage within Fedora package builds.

I would also prefer to have a cleaner solution which is why I prefer to not break APIs but upstream mojo did so this is why we need to handle that problem downstream in openQA (or decide to push further downstream). IMHO we can accept the limited extent of compatibility code. There are many different ways: Have if/else in the code, maintain patches in a "patches" or "contrib" directory within this project, dynamic patching as part of the install instructions in the Makefile or maintaining downstream in Fedora. For now I am keeping my approval here. I would still prefer to have `# uncoverable statement` comments in the code. But effectively we can solve it democratically and let mergify merge the PR as soon as requirements are fulfilled, i.e. tests fixed (likely by rebasing) and a second approval.

AdamWill:
As I said, it's not really a deal breaker for me to have this merged, if it makes for a harmonious life I can just carry it downstream. It was just difficult trying to understand why different issues/PRs for the same basic problem get different responses. Apparently the answer is: different people have different perspectives and some of them didn't see the initial issue/PR, which is fine.

PR #4007, opened 2021-07-02
Fix skip_all in 27-plugin_obs_rsync_status_details.t

AdamWill:
Where it is is too late; if you actually hit the condition, you
get this error:

Parse errors: Plan (1..0) must be at the beginning or end of the TAP output
              Bad plan.  You planned 0 tests but ran 3.

This fixes that by moving it up a line.

### DIFF
--- a/t/33-developer_mode.t
+++ b/t/33-developer_mode.t
@@ -63,7 +63,7 @@ sub turn_down_stack {
     stop_service($_) for ($worker, $ws, $livehandler, $scheduler);
 }
 
-plan skip_all => $OpenQA::SeleniumTest::drivermissing unless check_driver_modules;
+driver_missing unless check_driver_modules;
 
 # setup directories
 my $tempdir   = setup_fullstack_temp_dir('full-stack.d');
--- a/t/full-stack.t
+++ b/t/full-stack.t
@@ -70,7 +70,7 @@ sub turn_down_stack {
 }
 sub stop_worker { stop_service $worker }
 
-plan skip_all => $OpenQA::SeleniumTest::drivermissing unless check_driver_modules;
+driver_missing unless check_driver_modules;
 
 # setup directories
 my $tempdir  = setup_fullstack_temp_dir('full-stack.d');
--- a/t/lib/OpenQA/SeleniumTest.pm
+++ b/t/lib/OpenQA/SeleniumTest.pm
@@ -7,7 +7,7 @@ use base 'Exporter';
 
 require OpenQA::Test::Database;
 
-our @EXPORT = qw($drivermissing check_driver_modules enable_timeout
+our @EXPORT = qw(driver_missing check_driver_modules enable_timeout
   disable_timeout start_driver
   call_driver kill_driver wait_for_ajax disable_bootstrap_animations
   wait_for_ajax_and_animations
@@ -32,8 +32,7 @@ our $_driver;
 our $webapi;
 our $gru;
 our $mojoport;
-our $startingpid   = 0;
-our $drivermissing = 'Install Selenium::Remote::Driver and Selenium::Chrome to run these tests';
+our $startingpid = 0;
 
 sub _start_app {
     my ($args) = @_;
@@ -397,6 +396,12 @@ sub kill_driver {
 
 sub get_mojoport {
     return $mojoport;
+}
+
+sub driver_missing {
+    diag 'Install Selenium::Remote::Driver and Selenium::Chrome to run these tests';
+    done_testing;
+    exit;
 }
 
 END {
--- a/t/ui/01-list.t
+++ b/t/ui/01-list.t
@@ -139,7 +139,7 @@ sub prepare_database {
 
 prepare_database();
 
-plan skip_all => $OpenQA::SeleniumTest::drivermissing unless my $driver = call_driver;
+driver_missing unless my $driver = call_driver;
 
 $driver->title_is("openQA", "on main page");
 is($driver->get("/results"), 1, "/results gets");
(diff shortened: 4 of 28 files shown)


okurz (approved):

perlpunk:
But this order doesn't work either.

I would suggest to make a function like this:
```perl
sub drivermissing {
    diag $OpenQA::SeleniumTest::drivermissing;
    done_testing;
    exit;
}
# in test
drivermissing() unless my $driver = call_driver({with_gru => 0});
```
instead of the skip_all.
We are using this in all ui tests, so a function would be worth it.

AdamWill:
How do you mean it "doesn't work"? It works fine in the Fedora package building context, at least. This test is now skipped where it wasn't before the change.

perlpunk:
CI is failing:

```
# Starting WebUI service. Port: 39843
# PID of openqa-webapi: 2165
[error] [pid:2165] Stopping openqa-webapi process because a Perl warning occurred: Failed to open /home/squamata/project/t/ui/../etc/openqa/database.ini: No such file or directory at /home/squamata/project/lib/OpenQA/Schema.pm line 56.
not ok 1 - sub process openqa-webapi terminated with exit code 10752
ok 2 - no (unexpected) warnings (via done_testing)
1..2
```
And for me it fails as well locally.

AdamWill:
ah...perhaps it works if a previous test already did the `setup_obs_rsync_test` thing, but fails in isolation? if so, that's awkward. Your proposal seems fine, I'm OK with anything that makes all cases work here. I'm not sure precisely where you envisage putting that, though.

perlpunk:
> perhaps it works if a previous test already did the setup_obs_rsync_test thing, but fails in isolation?

It always fails here if `$driver` is actually containing something true. `setup_obs_rsync_test` is creating a tempdir, so it should not be influenced by previous tests.


> I'm not sure precisely where you envisage putting that, though.

I would put the function in OpenQA::SeleniumTest, where call_driver is already located.
I'm assigned to other issues right now, though.

Maybe there is also a way to split up `setup_obs_rsync_test` into two functions.
Or `call_driver` can be split up. Probably only the `check_driver_modules` check is relevant for the `skip_all`.

AdamWill:
Another option is to just not have `setup_obs_rsync_test` use test-y functions, I guess. Thanks for the pointers, I'll see if I can do something later today.

AdamWill:
OK, I went with more or less what you suggested (except I got rid of the `$drivermissing` string because it's not needed any more if we change all the places that use it). Let's see how this flies.

perlpunk:
Thanks, looks good to me.
I set it to not-ready so that @okurz and @Martchus can re-review.
Tests are failing early right now because the DBD::SQLite issue though.

okurz (approved):

Martchus (approved):

AdamWill:
Ah. Darn. It doesn't work because you can't `exit` from a subroutine (or you can, but it just exits *the subroutine*, it doesn't exit the script).

Unfortunately I don't know of a way around that. You can *die* from a subroutine, but that's not what we want, we just want to exit quietly. We might have to have every test that calls `driver_missing` also `exit` afterwards. :/ Anyone got a better idea?

AdamWill:
Oh hah, that was silly. I still had the order of the lines reversed so it was the same problem as before. I flipped them back to the original order now, let's see if that works.

AdamWill:
I don't think the OBS failures have anything to do with the PR. Four are "unresolvable" (not sure what that means), one was a test failure in a test that this PR does not touch (`43-scheduling-and-worker-scalability.t`).

okurz:
Right now in OBS checks I see a failure in https://build.opensuse.org/package/live_build_log/devel:openQA:TestGithub:PR-4007/openQA/openSUSE_Factory/x86_64 in

```
[ 1081s]     #   Failed test 'all jobs done'
[ 1081s]     #   at ./t/43-scheduling-and-worker-scalability.t line 216.
[ 1081s]     #          got: '4'
[ 1081s]     #     expected: '5'
[ 1081s] 
[ 1081s]     #   Failed test 'all jobs passed'
[ 1081s]     #   at ./t/43-scheduling-and-worker-scalability.t line 217.
[ 1081s]     #          got: '4'
[ 1081s]     #     expected: '5'
[ 1081s]     # All jobs:
[ 1081s]     #  - id: 1, state: done, result: passed, reason: none
[ 1081s]     #  - id: 2, state: done, result: passed, reason: none
[ 1081s]     #  - id: 3, state: assigned, result: none, reason: none
[ 1081s]     #  - id: 4, state: done, result: passed, reason: none
[ 1081s]     #  - id: 5, state: done, result: passed, reason: none
[ 1081s]     # Looks like you failed 2 tests of 3.
[ 1081s] 
[ 1081s] #   Failed test 'assign and run jobs'
[ 1081s] #   at ./t/43-scheduling-and-worker-scalability.t line 219.
[ 1081s] # Looks like you failed 1 test of 4.
[ 1081s] ./t/43-scheduling-and-worker-scalability.t .................. 
[ 1081s] # Running scalability test with 2 worker(s) and 5 job(s).
[ 1081s] # Set SCALABILITY_TEST_WORKER_COUNT/SCALABILITY_TEST_JOB_COUNT to adjust this.
```

I think I have seen this in other cases as well. I have reported https://progress.opensuse.org/issues/96545 to handle that urgently.

perlpunk:
Again, rebasing is necessary to pass all tests, sorry.
The following PRs have been merged;
* https://github.com/os-autoinst/openQA/pull/4106
* https://github.com/os-autoinst/openQA/pull/4105

okurz:
I rebased with `git pull --rebase origin master && git push -f git@github.com:AdamWill/openQA.git HEAD:obs-fix-skip`

AdamWill:
thanks!

PR #4009, opened 2021-07-05
ci: Increase several timeouts

perlpunk:
due to decreased CircleCI performance

* MOJO_INACTIVITY_TIMEOUT
* TIMEOUT_M
* OpenQA::Test::TimeLimit

Issue: https://progress.opensuse.org/issues/95009

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -27,7 +27,12 @@ JSBEAUTIFIER_OPTS ?= --brace-style=collapse,preserve-inline
 TEST_PG_PATH ?= /dev/shm/tpg
 # TIMEOUT_M: Timeout for one retry of tests in minutes
 TIMEOUT_M ?= 60
-TIMEOUT_RETRIES ?= $$((${TIMEOUT_M} * (${RETRY} + 1) ))m
+ifeq ($(CI),)
+SCALE_FACTOR ?= 1
+else
+SCALE_FACTOR ?= 2
+endif
+TIMEOUT_RETRIES ?= $$((${TIMEOUT_M} * ${SCALE_FACTOR} * (${RETRY} + 1) ))m
 mkfile_path := $(abspath $(lastword $(MAKEFILE_LIST)))
 current_dir := $(patsubst %/,%,$(dir $(mkfile_path)))
 docker_env_file := "$(current_dir)/docker.env"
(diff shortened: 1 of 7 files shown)


Makefile
```diff
@@ -148,19 +148,19 @@ test-checkstyle: test-checkstyle-standalone test-tidy-compile
...
+	$(MAKE) test-with-database TIMEOUT_M=30 PROVE_ARGS="$$HARNESS t/*.t" GLOBIGNORE="t/*tidy*:t/*compile*:$(unstables)"
 
 .PHONY: test-heavy
 test-heavy:
 	$(MAKE) test-with-database HEAVY=1 TIMEOUT_M=25 PROVE_ARGS="$$HARNESS $$(grep -l HEAVY=1 t/*.t | tr '\n' ' ')"
 
 .PHONY: test-ui
 test-ui:
-	$(MAKE) test-with-database TIMEOUT_M=25 PROVE_ARGS="$$HARNESS t/ui/*.t" GLOBIGNORE="t/*tidy*:t/*compile*:$(unstables)"
+	$(MAKE) test-with-database MOJO_INACTIVITY_TIMEOUT=120 TIMEOUT_M=35 PROVE_ARGS="$$HARNESS t/ui/*.t" GLOBIGNORE="t/*tidy*:t/*compile*:$(unstables)"
```
okurz, line 159: Can you put the mojo timeout into the test module itself? Or if it's right now really only needed in circleCI you could put it into the circleCI config

perlpunk: Done. I assumed. with test module you mean the test script.

okurz: yes, test script

Makefile
```diff
@@ -148,19 +148,19 @@ test-checkstyle: test-checkstyle-standalone test-tidy-compile
 
 .PHONY: test-t
 test-t:
-	$(MAKE) test-with-database TIMEOUT_M=25 PROVE_ARGS="$$HARNESS t/*.t" GLOBIGNORE="t/*tidy*:t/*compile*:$(unstables)"
+	$(MAKE) test-with-database TIMEOUT_M=30 PROVE_ARGS="$$HARNESS t/*.t" GLOBIGNORE="t/*tidy*:t/*compile*:$(unstables)"
```
okurz, line 151: how about we apply the same time scaling here as in https://github.com/os-autoinst/openQA/blob/master/external/os-autoinst-common/lib/OpenQA/Test/TimeLimit.pm#L26 ? So if `CI` is set, scale timeout in https://github.com/os-autoinst/openQA/pull/4009/files#diff-76ed074a9305c04054cdebb9e9aad2d818052b07091de1f20cad0bbac34ffb52R30 instead of changing here

perlpunk: I assumed that you would be against a global increase, so I spent friday evening to find the right timeout for each of the tests :(

I added a SCALE_FACTOR now in the Makefile

kraih (approved):

okurz (approved):

PR #4010, opened 2021-07-05
Return the error message the Downloader up the stack

dcermak:
We are currently swallowing the error message when anything during the download
& extraction process fails and simply report to the user that:
Downloading "$url" failed because of too many download errors

However, that is potentially completely wrong, as we also support extracting
files in downloader->download() and that process can fail due to a completely
different reason. The actual error message is then only visible in the log file
and needs admin access to the webui.

With this change we expose the error message in the webui as well, thereby
allowing the user to diagnose issues like this easier themselves without having
access to the logs.

This should also help diagnose issues like
https://progress.opensuse.org/issues/94997.


lib/OpenQA/Task/Asset/Download.pm
```diff
@@ -100,8 +100,9 @@ sub _download {
             $ctx->debug(qq{Download of "$assetpath" successful});
         }
     };
-    unless ($downloader->download($url, $assetpath, $options)) {
-        $ctx->error(my $msg = qq{Downloading "$url" failed because of too many download errors});
+    my $err = $downloader->download($url, $assetpath, $options);
+    if ($err) {
```
okurz, line 104: You can combine the two above lines, e.g. `if (my $err = ...`

dcermak: I've added this simplification.

kalikiana (approved):
That looks pretty elegant. Nice.

okurz (approved):

okurz:
@dcermak you could help me with review by explain what changes you did in your last push.

dcermak:
> @dcermak you could help me with review by explain what changes you did in your last push.

I changed the `_get` function to return `($ret, $err)` initially only in the last return and after fixing the tests, I realized that the error message was not correctly returned in some cases. So I made it return the tupple in every return statement in `_get`.

PR #4012, opened 2021-07-06
Fix a few typos

Martchus:
The typos have been determined via `codespell --skip .git -w`.
Unfortunately there are too many false-positives to add this as automatic
check.


okurz (approved):
Very nice. I wonder if we can mark false-positives for exclusion explicitly?

Martchus:
It is likely possible to tweak the `codespell` invocation to avoid the false-positives (e.g. using `--ignore-regex` and `--exclude-file`).

kalikiana (commented):
How did you find these? Is this something we can add to tests/CI?
Edit: I'd not seen the description explaning the change before.

Btw can you make the commit message something more descriptive, like `Fix typos found by codespell`

PR #4014, opened 2021-07-06
Distinguish comment types on /tests

perlpunk:
For finished tests (/tests/list_ajax):
Like in /tests/overview, figure out the type of comment (bug, closed/open,
label) and display them including links.

Issue: https://progress.opensuse.org/issues/94937

Edit: request times are going up from around 570-620ms to 620-720ms

### DIFF
--- a/assets/javascripts/tests.js
+++ b/assets/javascripts/tests.js
@@ -90,9 +90,8 @@ function renderTestName(data, type, row) {
         }
         html += dependencyHtml;
     }
-    if (row.comment_count) {
-        html += ' <a href="/tests/' + row.id + '#comments"><i class="test-label label_comment fa fa-comment" title="' +
-            row.comment_count + (row.comment_count != 1 ? ' comments' : ' comment') + ' available"' + '></i></a>';
+    if (row.comment_data) {
+        html += renderComments(row);
     }
     if (row.clone) {
         html += ' <a href="/tests/' + row.clone + '">(restarted)</a>';
--- a/lib/OpenQA/WebAPI/Controller/Test.pm
+++ b/lib/OpenQA/WebAPI/Controller/Test.pm
@@ -14,7 +14,7 @@
 # with this program; if not, see <http://www.gnu.org/licenses/>.
 
 package OpenQA::WebAPI::Controller::Test;
-use Mojo::Base 'Mojolicious::Controller';
+use Mojo::Base 'Mojolicious::Controller', -signatures;
 
 use OpenQA::Utils;
 use OpenQA::Jobs::Constants;
@@ -42,21 +42,6 @@ sub list {
     my ($self) = @_;
 }
 
-sub prefetch_comment_counts {
-    my ($self, $job_ids) = @_;
-
-    my $comments = $self->schema->resultset("Comments")->search(
-        {'me.job_id' => {in => $job_ids}},
-        {
-            select   => ['job_id', {count => 'job_id', -as => 'count'}],
-            group_by => [qw(job_id)]});
-    my $comment_count;
-    while (my $count = $comments->next) {
-        $comment_count->{$count->job_id} = $count->get_column('count');
-    }
-    return $comment_count;
-}
-
 sub get_match_param {
     my ($self) = @_;
 
@@ -68,9 +53,7 @@ sub get_match_param {
     return $match;
 }
 
-sub list_ajax {
-    my ($self) = @_;
-
+sub list_ajax ($self) {
     my $scope = ($self->param('relevant') ne 'false' ? 'relevant' : '');
     my @jobs  = $self->schema->resultset('Jobs')->complex_query(
         state    => [OpenQA::Jobs::Constants::FINAL_STATES],
@@ -90,33 +73,59 @@ sub list_ajax {
         prefetch => [qw(children parents)],
     )->all;
 
-    # need to use all as the order is too complex for a cursor
-    my $comment_count = $self->prefetch_comment_counts([map { $_->id } @jobs]);
+    my $comment_data = $self->schema->resultset('Comments')->comment_data_for_jobs(\@jobs, {bugdetails => 1});
     my @list;
     for my $job (@jobs) {
-        my $job_id = $job->id;
+        my $job_id        = $job->id;
+        my $rendered_data = 0;
+        if (my $cd = $comment_data->{$job_id}) {
+            $rendered_data = $self->_render_comment_data_for_ajax($job_id, $cd);
+        }
         push(
             @list,
             {
-                DT_RowId      => 'job_' . $job_id,
-                id            => $job_id,
-                result_stats  => $job->result_stats,
-                deps          => $job->dependencies,
-                clone         => $job->clone_id,
-                test          => $job->TEST . '@' . ($job->MACHINE // ''),
-                distri        => $job->DISTRI  // '',
-                version       => $job->VERSION // '',
-                flavor        => $job->FLAVOR  // '',
-                arch          => $job->ARCH    // '',
-                build         => $job->BUILD   // '',
-                testtime      => ($job->t_finished // '') . 'Z',
-                result        => $job->result,
-                group         => $job->group_id,
-                comment_count => $comment_count->{$job_id} // 0,
-                state         => $job->state,
+                DT_RowId     => 'job_' . $job_id,
+                id           => $job_id,
+                result_stats => $job->result_stats,
+                deps         => $job->dependencies,
+                clone        => $job->clone_id,
+                test         => $job->TEST . '@' . ($job->MACHINE // ''),
+                distri       => $job->DISTRI  // '',
+                version      => $job->VERSION // '',
+                flavor       => $job->FLAVOR  // '',
+                arch         => $job->ARCH    // '',
+                build        => $job->BUILD   // '',
+                testtime     => ($job->t_finished // '') . 'Z',
+                result       => $job->result,
+                group        => $job->group_id,
+                comment_data => $rendered_data,
+                state        => $job->state,
             });
     }
     $self->render(json => {data => \@list});
+}
+
+sub _render_comment_data_for_ajax ($self, $job_id, $comment_data) {
+    my %data;
+    $data{comments}     = $comment_data->{comments};
+    $data{reviewed}     = $comment_data->{reviewed};
+    $data{label}        = $comment_data->{label};
+    $data{comment_icon} = $self->comment_icon($job_id, $data{comments});
+    my $bugs = $comment_data->{bugs};
+    $data{bugs} = [
+        map {
+            my $bug = $comment_data->{bugdetails}->{$_};
+            +{
+                id        => $bug->id,
+                content   => $_,
+                closed    => $bug->open,
+                title     => $self->bugtitle_for($_, $bug),
+                url       => $self->bugurl_for($_),
+                css_class => $self->bugicon_for($_, $bug),
+            };
+        } sort keys %$bugs
+    ];
+    return \%data;
 }
 
 sub list_running_ajax {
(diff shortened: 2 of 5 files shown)


perlpunk:
For a screenshot see https://progress.opensuse.org/issues/94937

assets/javascripts/tests.js
```diff
@@ -90,9 +90,28 @@ function renderTestName(data, type, row) {
...
-        html += ' <a href="/tests/' + row.id + '#comments"><i class="test-label label_comment fa fa-comment" title="' +
-            row.comment_count + (row.comment_count != 1 ? ' comments' : ' comment') + ' available"' + '></i></a>';
+    if (row.comment_data) {
+        var bugs = row.comment_data.bugs;
+        for (var i = 0; i < bugs.length; i++) {
+            var bug = bugs[i];
+            var css_class = bug.css_class;
+            var title = bug.title;
+            var url = bug.url;
+            html += '<span id="bug-' + bug.id + '"> ' + '<a href="' + url + '">';
```
Martchus, line 100: It would make sense to apply `htmlEscape()` on the variables when generating HTML code.

assets/javascripts/tests.js
```diff
@@ -90,9 +90,28 @@ function renderTestName(data, type, row) {
         }
         html += dependencyHtml;
     }
-    if (row.comment_count) {
-        html += ' <a href="/tests/' + row.id + '#comments"><i class="test-label label_comment fa fa-comment" title="' +
-            row.comment_count + (row.comment_count != 1 ? ' comments' : ' comment') + ' available"' + '></i></a>';
+    if (row.comment_data) {
+        var bugs = row.comment_data.bugs;
```
Martchus, line 94: You can mark many of the variables `const`, e.g.
```suggestion
        const bugs = row.comment_data.bugs;
```

lib/OpenQA/WebAPI/Controller/Test.pm
```diff
@@ -90,30 +75,53 @@ sub list_ajax {
         prefetch => [qw(children parents)],
     )->all;
 
-    # need to use all as the order is too complex for a cursor
-    my $comment_count = $self->prefetch_comment_counts([map { $_->id } @jobs]);
+    my $comment_data = $self->schema->resultset('Comments')->comment_data_for_jobs(\@jobs, {bugdetails => 1});
     my @list;
     for my $job (@jobs) {
         my $job_id = $job->id;
+        my $cd     = $comment_data->{$job_id} // 0;
```
kalikiana, line 82: What is the `// 0` for here?

kalikiana (approved):
Looks sensible. Just one comment

assets/javascripts/tests.js
```diff
@@ -90,9 +90,28 @@ function renderTestName(data, type, row) {
...
+        for (var i = 0; i < bugs.length; i++) {
```
Martchus, line 95: You could also use `Array.prototype.forEach()` instead of the hand-written loop.

lib/OpenQA/WebAPI/Controller/Test.pm
```diff
@@ -90,30 +75,53 @@ sub list_ajax {
...
+        my %comment_data;
+        if ($cd) {
```
Martchus, line 84:
```suggestion
        my %comment_data;
        if (my $cd = $comment_data->{$job_id}) {
```

lib/OpenQA/WebAPI/Controller/Test.pm
```diff
@@ -90,30 +75,53 @@ sub list_ajax {
...
+                my $bug  = $details->{$bugid};
+                my $item = {
+                    id        => $bug->id,
+                    content   => $bugid,
+                    closed    => $bug->open,
+                    title     => $self->bugtitle_for($bugid, $bug),
+                    url       => $self->bugurl_for($bugid),
+                    css_class => $self->bugicon_for($bugid, $bug),
+                };
+                push @bugs, $item;
```
Martchus, line 102: Could the array be populated using `map`?

t/ui/01-list.t
```diff
@@ -57,6 +57,13 @@ my %job_param = (
 # We do not need job 99981 right now so delete it here just to have a helpful
 # example for customizing the test database
 sub prepare_database {
+    my $bugs     = $schema->resultset('Bugs');
+    my %bug_args = (refreshed => 1, existing => 1);
```
Martchus, line 61: Looks like this isn't really used as a hash so it would be possible to use an array:

```suggestion
    my @bug_args = (refreshed => 1, existing => 1);
```
(Not sure what's generally preferred here.)

lib/OpenQA/WebAPI/Controller/Test.pm
perlpunk: Oh, that's a leftover from a first version of the code

t/ui/01-list.t
perlpunk: Hm, I'm not sure why you think it isn't used as a hash...

lib/OpenQA/WebAPI/Controller/Test.pm
perlpunk: fixed

assets/javascripts/tests.js
perlpunk: done

assets/javascripts/tests.js
perlpunk: done

assets/javascripts/tests.js
perlpunk: done

lib/OpenQA/WebAPI/Controller/Test.pm
perlpunk: done

lib/OpenQA/WebAPI/Controller/Test.pm
perlpunk: ok, I rearranged the code to use map and saved some temp vars

lib/OpenQA/WebAPI/Controller/Test.pm
```diff
@@ -90,30 +75,50 @@ sub list_ajax {
...
+                        id        => $bug->id,
+                        content   => $_,
+                        closed    => $bug->open,
+                        title     => $self->bugtitle_for($_, $bug),
+                        url       => $self->bugurl_for($_),
+                        css_class => $self->bugicon_for($_, $bug),
+                    };
+                } sort keys %$bugs
+            ];
+        }
```
okurz, line 102: I consider this block rather complicated and would favor to have functions with good names extracted for something like this so that there is less indention. My preference is rather simple, similar to the code style in the Linux kernel: not more than two levels of indention

assets/javascripts/tests.js
```diff
@@ -90,9 +90,27 @@ function renderTestName(data, type, row) {
...
+        if (row.comment_data.label) {
+            const label = row.comment_data.label;
+            html += '<span id="test-label-' + row.id + '">';
+            html += ' <i class="test-label label_' + htmlEscape(label) + ' fa fa-bookmark" title="Label: ' + htmlEscape(label) + '"></i>';
+            html += '</span>';
+        } else if (row.comment_data.comments) {
+            html += '<span id="comment-' + row.id + '"> ';
+            html += row.comment_data.comment_icon;
+            html += '</span>';
+        }
```
okurz, line 113: This block looks like it's duplicating what we do as perl code in other cases. So you think it's feasible to use the same approach for deciding about which icon with which link we can add for each job result?

okurz (changes requested):

assets/javascripts/tests.js
perlpunk: > This block looks like it's duplicating what we do as perl code in other cases.

Yes, it'd duplicating some code in templates. Note that parts of it are already generated by perl helper functions.
An alternative would be to run the data through templates here as well and send the complete HTML for this column to Javascript.
Are we doing this already elsewhere?

> So you think it's feasible to use the same approach for deciding about which icon with which link we can add for each job result?

I don't understand this question.
I think it makes sense to use the same approach, because I assume the same result as in /tests/overview is wanted, so a *different* approach would be wrong.
Can you elaborate?

t/ui/01-list.t
Martchus: It isn't used like `$bug_args{some_key}`, just expanded in list context.

perlpunk: Ah, I see.
I don't see a reason to use an array though.
But if at least one more person suggests that, I'll change it :)

assets/javascripts/tests.js
Martchus: Rendering this on the server likely slows down the loading time. I'd rather try to render everything in JavaScript. (It is not nice that we need to generate HTML code here. That's a limitation of the DataTables API as far as I know - so we have to live with it.)

okurz: I think you already provided good suggestions. I would prefer if we use the same approach for both /tests as well as /tests/overview (maybe as well as "Next & previous job results"). Either that approach already exists for /tests/overview (maybe as well "Next & previous job results") or you need to create that new approach and use it everywhere so it will be the *same*

perlpunk: Oh, I totally overlooked Next & previous!
Similar code is used there.
I only looked at /tests/overview.

Now, /tests/overview is generated server side via a template, /tests/list_ajax and job_next_previous_ajax are getting data and render HTML on client side.

What should I do about that (before implementing something and having to change it afterwards)?

Marius votes against rendering throgh a perl template in the Javascript cases.
I personally think that doesn't make much difference performance wise, most load comes from querieng the DB.

Oli requests that the same approach is used. That would mean either
1. changing /tests/overview to use Javascript instead or
2. changing both ajax routes to pregenerate HTML before sending it to Javascript

Martchus: Don't change it to pregenerated HTML. I've already converted several tables over time away from pregenerated HTML because it was too slow. (We have to generate the HTML for *all* rows on the server and then DataTables needs to parse all of this HTML at once and turn it into a functional DataTable.)

For this PR, I would at least share existing JavaScript code. So if we have already JavaScript code for rendering this for "Next & previous" we should reuse it. Streamlining with the test result overview is likely out of scope here.

perlpunk: > We have to generate the HTML for _all_ rows on the server and then DataTables needs to parse all of this HTML

That's not true. I mean, I already *do* generate parts of the HTML here (calling the server side helper function `comment_icon` for example, which returns HTML.
What I meant was to just generate the HTML for the comment icons and do `html += comments`. Not the complete HTML table.

perlpunk: Oh, you mean all rows instead of the only 10/25 that are rendered.
Sorry, I agree.

Martchus: Ah, so you would generate the HTML and pass it as JSON string. That should be better although it would still run the code for all rows on the server upfront. With client-side rendering we can just render the rows we need to render which just makes more sense from a scalability perspective. So I'd still try to unify the JavaScript code we have so far for now. If we'd ever modernize the frontend code (e.g. using some fancy new JavaScript framework) then we'd likely tend to render even more on the client. So moving this code to the server-side seems like a step backwards.

Martchus (approved):

perlpunk:
I will unify the code with the code for `job_next_previous_ajax` which I totally overlooked.

lib/OpenQA/WebAPI/Controller/Test.pm
perlpunk: Now it's in its own function

okurz (approved):
Very nice

Martchus (approved):

okurz:
@perlpunk do you need to keep it as draft?

t/ui/01-list.t
kalikiana: Leave it imho

kalikiana (approved):
Nice revision of the previous approach!

PR #4015, opened 2021-07-06
plugins: Rate limit all routes

kalikiana:
This is a plugin that adds configurable rate limits to all routes.

- This is an alternative to #3990 which doesn't require manual configuration in all routes but rather works for all routes by default  ~~The configuration key `any` is used when nothing more specific was specified~~
- `search` is replaced by `apiv1_search_query` since now the route name is used for the key, and since this is experimental I assume we don't need compatibility code or migration - the default for this route is unchanged.
- Existing search tests are adapted accordingly.
- New tests for the overview route also check the rate limit

### DIFF
--- a/lib/OpenQA/WebAPI/Plugin/RateLimits.pm
+++ b/lib/OpenQA/WebAPI/Plugin/RateLimits.pm
@@ -1,0 +1,42 @@
+# Copyright (C) 2021 SUSE LLC
+#
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation; either version 2 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License along
+# with this program; if not, see <http://www.gnu.org/licenses/>.
+
+package OpenQA::WebAPI::Plugin::RateLimits;
+use Mojo::Base 'Mojolicious::Plugin', -signatures;
+
+
+use Time::Seconds;
+
+sub register ($self, $app, $conf) {
+    $app->hook(
+        around_action => sub ($next, $c, $action, $last) {
+            return $next->() unless $last;
+            # Allow n queries per minute, per user (if logged in)
+            my $route = $c->current_route;
+            # Only handle named routes with an explicit limit
+            if ($route && (my $limit = $c->app->config->{rate_limits}{$route})) {
+                my $lockname = "webui_${route}_rate_limit";
+                if (my $user = $c->current_user) { $lockname .= $user->username }
+                my $error = "Rate limit exceeded";
+                return $c->respond_to(
+                    json => {json     => {error => $error}, status => 429},
+                    html => {template => 'layouts/rate_limit_error', limit => $limit, status => 429},
+                ) unless $c->app->minion->lock($lockname, ONE_MINUTE, {limit => $limit});
+            }
+            return $next->();
+        });
+}
+
+1;
--- a/t/api/15-search.t
+++ b/t/api/15-search.t
@@ -24,7 +24,7 @@ use OpenQA::Test::Case;
 
 OpenQA::Test::Case->new->init_data;
 my $t = Test::Mojo->new('OpenQA::WebAPI');
-$t->app->config->{rate_limits}->{search} = 10;
+$t->app->config->{rate_limits}->{apiv1_search_query} = 10;
 
 subtest 'Perl modules' => sub {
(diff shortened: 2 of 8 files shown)


lib/OpenQA/WebAPI/Plugin/RateLimits.pm
```diff
@@ -0,0 +1,43 @@
...
+# with this program; if not, see <http://www.gnu.org/licenses/>.
+
+package OpenQA::WebAPI::Plugin::RateLimits;
+use Mojo::Base 'Mojolicious::Plugin', -signatures;
+
+
+use Time::Seconds;
+
+sub register {
+    my ($self, $app) = @_;
```
kalikiana, line 23: Note: I actually tried to use signatures here. But it gave me errors from plugin loading code. I guess I would need to fix this on another layer?

perlpunk: https://docs.mojolicious.org/Mojolicious/Plugin
```
sub register ($self, $app, $conf) {
```
:)

lib/OpenQA/WebAPI/Plugin/RateLimits.pm
```diff
@@ -0,0 +1,43 @@
...
+            my $route    = $c->current_route;
+            my $lockname = "webui_${route}_rate_limit";
+            if (my $user = $c->current_user) { $lockname .= $user->username }
+            my $error  = "Rate limit exceeded";
+            my $limits = $c->app->config->{'rate_limits'};
+            my $limit  = $limits->{$route} // $limits->{'any'};
+            return $c->respond_to(
+                json => {json     => {error => $error}, status => 429},
+                html => {template => 'layouts/rate_limit_error', limit => $limit, status => 429},
+            ) unless $c->app->minion->lock($lockname, ONE_MINUTE, {limit => $limit});
```
kraih, line 38: This should really be configurable though and not be enabled indiscriminately for every route. 😅

kraih: Based on route name.

kraih: And not every route has a name.

lib/OpenQA/WebAPI/Plugin/RateLimits.pm
kalikiana: Oooooohhhhhhh 🤦🏾‍♀️

lib/OpenQA/WebAPI/Plugin/RateLimits.pm
kalikiana: It is, though. `$limit` is based on the route name

kalikiana: I'll need to check routes with no names, though. I assumed one can rely on that since mojo docs seemed to imply it

lib/OpenQA/WebAPI/Plugin/RateLimits.pm
```diff
@@ -0,0 +1,43 @@
...
+use Time::Seconds;
+
+sub register {
+    my ($self, $app) = @_;
+
+    $app->hook(
+        around_action => sub ($next, $c, $action, $last) {
+            return $next->() unless $last;
+            # Allow n queries per minute, per user (if logged in)
+            my $route    = $c->current_route;
```
kraih, line 29: Below this like i'd just put an if that short circuits all the other rate limit checks.
```perl
if ($route && (my $limit = $c->app->config->{rate_limits}{$route})) {
  ...
```

lib/OpenQA/WebAPI/Plugin/RateLimits.pm
kalikiana: As discussed in RC, `current_route` can be the empty string. And I'll scrap `any` and stick to explicitly enables all routes.

templates/webapi/layouts/rate_limit_error.html.ep
```diff
@@ -0,0 +1,8 @@
+% layout 'bootstrap';
```
kraih, line 1: Why is this template in the `layouts` directory? Shouldn't it be in the directory above?

t/api/15-search.t
```diff
@@ -133,8 +133,11 @@ subtest 'Errors' => sub {
     $t->get_ok('/api/v1/experimental/search?q=*', 'wildcard is interpreted literally');
     $t->json_is('/data' => [], 'no result for *');
 
-    $t->app->config->{rate_limits}->{search} = 3;
-    $t->get_ok('/api/v1/experimental/search?q=timezone', 'Search succesful') for (1 .. 3);
+    $t->app->config->{rate_limits}->{apiv1_search_query} = 3;
+    $t->get_ok('/api/v1/experimental/search?q=timezone', {Accept => 'application/json'})
+      ->content_type_is('application/json;charset=UTF-8')
+      for (1 .. 3);
```
kraih, line 139: I would check `status_is` on all 3 requests, just for that extra little bit of test coverage. 😁

kalikiana:
Closing this as we've been pursuing other approaches in the meanwhile.

PR #4017, opened 2021-07-07
Simplify script/modify_needle

okurz:


script/modify_needle
```diff
@@ -44,16 +44,15 @@ print help
 use strict;
 use warnings;
 
-use Mojo::JSON;    # booleans
+use Mojo::File qw(path);
+use Mojo::JSON;
```
Martchus, line 48: It doesn't look like that module is used at all.

Martchus (commented):
Unfortunately lots of the `dump_templates` code isn't covered by tests.

script/modify_needle
okurz: fixed

okurz:
> Unfortunately lots of the `dump_templates` code isn't covered by tests.

Split out and moved to https://github.com/os-autoinst/openQA/pull/4023

Martchus (approved):

perlpunk (approved):

PR #4021, opened 2021-07-08
Fix/t jobs restart stable ci

okurz:


okurz:
Got my confirmation, see https://progress.opensuse.org/issues/95024#note-17 for results

PR #4023, opened 2021-07-08
Simplify script/dump_templates and add test coverage

okurz:


okurz:
I didn't introduce more coverage but improved the coverage ratio by simplifying. Hope you can approve nevertheless :)

Martchus (approved):
Yes, you've written uncovered code in a more compact way. However, it means the changes are not tested. I guess we can live with that for `dump_templates` for now.

perlpunk:
This would be a good opportunity to add a test

kalikiana (changes requested):
I see no description or mention of manual tests steps, and the code is not unit-tested afair

okurz:
> I see no description or mention of manual tests steps

correct

> and the code is not unit-tested afair

well, there is 
t/40-script_load_templates.t which also calls `dump_templates`. This is also the reason why the patch coverage is 72%, not 0%. I guess we should rename the test file?

kalikiana:
> > and the code is not unit-tested afair
> 
> well, there is
> t/40-script_load_templates.t which also calls `dump_templates`. This is also the reason why the patch coverage is 72%, not 0%. I guess we should rename the test file?

The report says "not covered" **and** "now covered" in several parts of the code you're changing in `dump_templates`. Other files aren't mentioned.
Looking at the test code itself at least the single vs. multiple groups case doesn't seem to be covered.

okurz:
Seems we have some flaky coverage back. I think I have seen this in before, likely from @Martchus' recent PRs. E should either fix this soon or revert such changes.

okurz:
Updated:
* Changed more to use signatures
* Extracted functions in script/dump_templates
* Added statement coverage to reach 100% (in combination with t/44-scripts.t if that works properly)

PR #4024, opened 2021-07-08
openqa-bootstrap: Check that systemd is available

asdil12:
Ticket: https://progress.opensuse.org/issues/95185

Calling `systemctl` ensures that

* systemd is installed
* systemd is running as PID=1

### DIFF
--- a/script/openqa-bootstrap
+++ b/script/openqa-bootstrap
@@ -6,6 +6,12 @@ set -x
 dbname="${dbname:="openqa"}"
 dbuser="${dbuser:="geekotest"}"
 
+# ensure that systemd is available (which is not the case in some containers)
+if [ -z "$BYPASS_SYSTEMD_CHECK" ] && ! systemctl > /dev/null ; then
+	echo "$0 requires systemd to be installed and running as the init daemon with PID=1"
+	exit 1
+fi
+
 # add extra repos for leap
 # shellcheck disable=SC1091
 . /etc/os-release


script/openqa-bootstrap
```diff
@@ -6,6 +6,14 @@ set -x
 dbname="${dbname:="openqa"}"
 dbuser="${dbuser:="geekotest"}"
 
+
+# ensure that systemd is available (which is not the case in some containers)
+if ! systemctl > /dev/null ; then
```
Martchus, line 11: It can still be useful to run the script within a container. Even if the systemd-specific parts in the end don't work it would make sense to be able to run the first parts. What about:

```suggestion
if [[ -n $BYPASS_SYSTEMD_CHECK ]] && ! systemctl > /dev/null ; then
```

Martchus (changes requested):

Martchus (approved):

kalikiana (approved):
The error looks weird, which I guess is expected. But it works in a container with no systemd.

```
+ systemctl
System has not been booted with systemd as init system (PID 1). Can't operate.
Failed to connect to bus: Host is down
+ echo './script/openqa-bootstrap requires systemd to be installed and running as the init daemon with PID=1'
./script/openqa-bootstrap requires systemd to be installed and running as the init daemon with PID=1
+ exit 1
```

script/openqa-bootstrap
```diff
@@ -6,6 +6,14 @@ set -x
 dbname="${dbname:="openqa"}"
 dbuser="${dbuser:="geekotest"}"
 
+
```
okurz, line 9: I would prefer a single blank line for separation. No need for two.

script/openqa-bootstrap
```diff
@@ -6,6 +6,14 @@ set -x
...
+# ensure that systemd is available (which is not the case in some containers)
+if [[ -z $BYPASS_SYSTEMD_CHECK ]] && ! systemctl > /dev/null ; then
+	echo "$0 requires systemd to be installed and running as the init daemon with PID=1"
+	exit 1
+fi
+
+
```
okurz, line 16: same here

dist/rpm/openQA.spec
```diff
@@ -650,6 +652,8 @@ fi
 %files bootstrap
 %{_datadir}/openqa/script/openqa-bootstrap
 %{_datadir}/openqa/script/openqa-bootstrap-container
+%{_bindir}/openqa-bootstrap
+%{_bindir}/openqa-bootstrap-container
```
okurz, line 656: why this change? Has it anything to do with the other commit? I think we decided that we don't want the script to be within the main bin dir so that users do not trigger it by accident or "curiousity" exploring options of tab-completion :)

okurz (changes requested):

okurz (approved):

okurz:
I guess you need to rebase again

Martchus (approved):

PR #4025, opened 2021-07-08
Fix use of `isnt` in `t/lib/OpenQA/Test/Utils.pm`

Martchus:
I broke this in 71bef7e because it is one of the false-positive changes
done by codespell.


perlpunk:
And it compiled because `isn't` means `isn::t` in perl :]

perlpunk (approved):

okurz (approved):

PR #4026, opened 2021-07-08
Avoid forking in t/24-worker-jobs.t to collect coverage of upload code

Martchus:
* See https://progress.opensuse.org/issues/95170
* This also removes the necessity of using `shared_hash` simplifying the
  test setup

### DIFF
--- a/lib/OpenQA/Worker/Job.pm
+++ b/lib/OpenQA/Worker/Job.pm
@@ -855,64 +855,40 @@ sub _upload_results_step_1_post_status {
     );
 }
 
-sub _upload_results_step_2_upload_images {
-    my ($self, $callback) = @_;
+sub _upload_results_step_2_1_upload_images ($self) {
+    my ($job_id, $client) = ($self->id, $self->client);
+    my $images_to_send = $self->images_to_send;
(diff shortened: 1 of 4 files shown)


perlpunk (approved):

okurz (changes requested):
Looks nice. But your changes include uncovered code section in critical worker code. I would feel saver if we would have this code included in tests. If you plan to do that anyway then I would appreciate if you can do that first or include here. If you see this unfeasible for now then just say so and we can go ahead as is

Martchus:
I've been testing the refactored code locally by running a few jobs and it seems to work. I also extended tests to cover all lines of the refactored code.

Note that the refactoring of actual worker code is only a small part of the diff. The big part of the diff is refactoring of the of the test code.

okurz:
Good enough for an "acceptance-tests". Now to fix the other test problems. Little hint: Just retriggering without handling it properly will make me mad ;)

Martchus:
I haven't realized there were test problems (besides the ones about timeouts which are already handled). The webui-docker-compose check seems to fail everywhere now.

PR #4028, opened 2021-07-08
doc: Document finding tests via the REST-API

Martchus:
See https://progress.opensuse.org/issues/95075

### DIFF
--- a/docs/UsersGuide.asciidoc
+++ b/docs/UsersGuide.asciidoc
@@ -877,6 +877,55 @@ This section focuses on particular API use-cases. Checkout the
 <<Client.asciidoc#client,openQA client>> section for further information about
 the client itself, how authentication works and how plain `curl` can be used.
 
+=== Finding tests
+
+The following example lists all jobs within the job group with the ID `1`
+and the setting `BUILD=20210707` on `openqa.opensuse.org`:
+
+[source,sh]
+----
+curl -s "https://openqa.opensuse.org/api/v1/jobs?groupid=1&build=20210707" | jq
+----
+
+The tool `jq` is used for pretty-printing in this example but it is also useful
+for additional filtering (see js's
+https://stedolan.github.io/jq/tutorial[tutorial]).
+
+However, openQA's API provides many more filters on its own. These can be used
+by adding additional query parameters, e.g.:
+
+* `ids`/`state`/`result`: Return only jobs with matching ID/state/result.
+  Multiple IDs/states/results can be specified by repeating the parameter or
+  by passing comma-separated values.
+* `distri`/`version`/`build`/`test`/`arch`/`machine`
+  /`worker_class`/`iso`/`hdd_1`: Return only jobs where the job settings match
+  the specified values like in the example above. Note that it is not possible
+  to filter by arbitrary job settings although this list might not be complete.
+* `groupid`/`group`: Return only jobs within the job group with the
+  specified ID/name like in the example above. These parameters are mutually
+  exclusive, `groupid` has precedence.
+* `latest=1`: De-duplicates, so that for the same `DISTRI`, `VERSION`, `BUILD`,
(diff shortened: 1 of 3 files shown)


perlpunk (approved):

docs/UsersGuide.asciidoc
```diff
@@ -868,13 +868,46 @@ settings), you have to quote it.
...
+However, openQA's API provides many more filters on its own. These can be used
+by adding additional query parameters, e.g.:
+
+* `ids`/`state`/`result`: Return only jobs with matching ID/state/result.
+  Multiple IDs/states/results can be specified by repeating the parameter or
+  by passing comma-separated values.
+* `distri`/`version`/`build`/`test`/`arch`/`machine`
+  /`worker_class`/`iso`: Return only jobs where the job settings match the
+  specified values like in the example above. Note that it is not possible to
+  filter by arbitrary job settings.
```
okurz, line 903: How about s/arbitrary/any other, not explicitly mentioned/? But on the other hand we might change the implementation and forget to update the documentation. Maybe you have a good idea for a phrasing that explains what is possible, what is not possible while still leaving the option open that more might work :)

docs/UsersGuide.asciidoc
```diff
@@ -868,13 +868,46 @@ settings), you have to quote it.
...
+* `groupid`/`group`: Return only jobs within the job group with the
+  specified ID/name like in the example above.
+* `latest=1`: De-duplicates, so that for the same `DISTRI`, `VERSION`, `BUILD`,
+  `TEST`, `FLAVOR`, `ARCH` and `MACHINE` only the latest job is returned.
+* `limit`/`page`: Limit the number of returned jobs and allow paging.
```
okurz, line 908: The "allow paging" part is not quite clear to me. Like, what values should be specified and what do they mean

docs/UsersGuide.asciidoc
```diff
@@ -868,13 +868,46 @@ settings), you have to quote it.
...
+* `modules`/`modules_result`: Return only jobs which have a test module with the
```
okurz, line 909: Is it the exact name or a substring?

okurz (changes requested):

docs/UsersGuide.asciidoc
Martchus: The exact name (as with every other parameter).

docs/UsersGuide.asciidoc
Martchus: Would it help to use the work pagination and maybe put a link to https://en.wikipedia.org/wiki/Pagination? It seems quite redundant to explain a term here which should be common to everybody who is using databases.

docs/UsersGuide.asciidoc
Martchus: I could append "… although this list might not be complete". I could also add a comment at the place parameters are validated which says one should update the documentation accordingly.

okurz: yes to both

docs/UsersGuide.asciidoc
okurz: Yes, mention "pagination", not even with a link to wikipedia, that should be obvious. But I think here it would help to explain the difference to the default behaviour and also an example would help.

docs/UsersGuide.asciidoc
okurz: fine, keep as is

okurz: actually please replace "name" with "exact name" or corresponding term in the chapter

docs/UsersGuide.asciidoc
kalikiana: Something like `For example page=2&limit=10 would only show results 11-20`. Or the equivalent that actually happens. That's the part I can't guess despite being familiar with the term

okurz (approved):

kalikiana (approved):

PR #4031, opened 2021-07-09
Revert "Distinguish comment types on /tests" due to poo#95281

okurz:
Reverts os-autoinst/openQA#4014 due to
https://progress.opensuse.org/issues/95281


nicksinger (approved):
Since this is just a revert I can give my +1 without understanding what happens here :)

Martchus (approved):
Note that not *all* "Next & Previous" pages are affected. Likely that's also why our tests didn't catch the problem.

Martchus:
Btw, it works again locally after applying the revert. So it is definitely caused by #4014.

PR #3987, opened 2021-06-29
Reenable sysusers

dcermak:
This is essentially a second try at https://github.com/os-autoinst/openQA/pull/3977, however now with additional checks in place to ensure that the user & group creation actually works. If I implemented the checks correctly, then the CI *should* fail on Tumbleweed until https://build.opensuse.org/request/show/901701 is accepted.

PR #4033, opened 2021-07-09
Revert "Reenable sysusers"

okurz:
Reverts os-autoinst/openQA#3987

@dcermak https://build.opensuse.org/package/live_build_log/devel:openQA/openQA/SLE_15_SP3/aarch64 looks like caused by this. I will prepare this revert in case you don't have time to fix that soon

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -115,10 +115,6 @@ install-generic:
 
 	cp -Ra dbicdh "$(DESTDIR)"/usr/share/openqa/dbicdh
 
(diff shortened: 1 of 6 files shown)


okurz:
As dcermak is unavailable for some time *and* because I suspect that https://progress.opensuse.org/issues/95374 might be caused by https://github.com/os-autoinst/openQA/pull/3987 as well I will merge the revert to give this more time to brew :)

PR #4037, opened 2021-07-09
t: Increase waiting time for 43-scheduling-and-worker-scalability.t

perlpunk:
Currently it regularly fails, e.g. https://app.circleci.com/pipelines/github/os-autoinst/openQA/7000/workflows/b4dafb8a-7325-4f81-8e5f-e19dcfc12ab5/jobs/66088

### DIFF
--- a/t/43-scheduling-and-worker-scalability.t
+++ b/t/43-scheduling-and-worker-scalability.t
@@ -136,8 +136,8 @@ my $casedir = testcasedir($distri, $version, undef);
 path($casedir)->make_path unless -d $casedir;
 
 my $seconds_to_wait_per_worker = 5.0;
-my $seconds_to_wait_per_job    = 2.5;
-my $polling_interval           = 0.1;
+my $seconds_to_wait_per_job    = 3.5;
+my $polling_interval           = 0.2;
 my $polling_tries_workers      = $seconds_to_wait_per_worker / $polling_interval * $worker_count;
 my $polling_tries_jobs         = $seconds_to_wait_per_job / $polling_interval * $job_count;
 


t/43-scheduling-and-worker-scalability.t
```diff
@@ -136,8 +136,8 @@ my $casedir = testcasedir($distri, $version, undef);
 path($casedir)->make_path unless -d $casedir;
 
 my $seconds_to_wait_per_worker = 5.0;
-my $seconds_to_wait_per_job    = 2.5;
-my $polling_interval           = 0.1;
+my $seconds_to_wait_per_job    = 3.5;
+my $polling_interval           = 0.2;
```
okurz, line 140: @Martchus 
AFAIR these values were not meant to be tweaked for slow environments but work in a reliable and reproducible manner regardless of the environment and such "timing issues"
Do you have a better idea?

kalikiana: Maybe these values should be defined relative to the total allowed time? Based on the effective scaling factor.

It would be hepful to know where the numbers come from, though. i.e. was it tests that only failed with certain timings?

Martchus: Yes, these values are related to the total allowed time. I'd say they are ideally a bit lower than the total allowed time.

When dealing with a slow environment is would make sense to increase these values (in accordance with the total allowed time). (Note that these timeouts are per job/worker so they are automatically adjusted when setting the `SCALABILITY_TEST_…` environment variables. It is problematic that the total allowed time is not automatically adjusted as well but that's likely not the issue here.)

I must say that the current values are already quite high. However, on a very busy system where the OS scheduler decides not to run certain processes for seconds they're of course nothing. Ideally we'd only account the actual CPU time here.

okurz: ok, can we do that? only account actual CPU time? How about using https://metacpan.org/pod/Timer::CPU ? At least it's in https://build.opensuse.org/package/show/devel:languages:perl:CPAN-T/perl-Timer-CPU ;)

Martchus: Considering https://metacpan.org/pod/Timer::CPU#CAVEATS I don't think we want it: 

> These measurements relate to wall-clock time via an unspecified conversion ratio and can therefore only be used for relative comparisons between code on the same machine (and are subject to other constraints described below)." is p

This is also problem because we need to put some generic threshold in our code.

kalikiana: Since this uses `Time::HiRes::sleep` we could use [the return value](https://perldoc.perl.org/Time::HiRes#sleep-(-$floating_seconds-)) to ensure a more accurate wait 🤔️

Martchus: The accuracy doesn't really matter. The actual CPU time would be interesting (of the process, e.g. worker process, we're waiting for).

okurz: @Martchus of course it can't be deterministic but certainly it's a better estimate than the wall time

okurz:
please check if this is still needed after https://github.com/os-autoinst/openQA/pull/4072

PR #4038, opened 2021-07-12
Show logs when docker compose up fails in the test_containers_compose

ilausuch:
When docker-compose up fails we don't have information to investigate.
This PR show the logs when something wrong happens.

To test the failure case and see the logs. Introduce an "exit 1" as the first line in the script containers/webui/run_openqa.sh. The logs will be showed on the terminal as:

```
Creating webui_db_1 ... done
Creating webui_webui_db_init_1 ... done

ERROR: for webui  Container "XXXXXXXXXXXX" is unhealthy.
ERROR: Encountered errors while bringing up the project.
Attaching to webui_webui_db_init_1, webui_db_1
....
```

https://progress.opensuse.org/issues/95374

### DIFF
--- a/tools/test_containers_compose
+++ b/tools/test_containers_compose
@@ -20,7 +20,13 @@ wait_until() {
 
 setup_containers() {
   for retry in {2..0}; do sudo docker-compose build && break; echo "Remaining retries $retry"; done
-  sudo MOJO_CLIENT_DEBUG=1 docker-compose up -d
+  exit_code=""
+  sudo MOJO_CLIENT_DEBUG=1 docker-compose up -d || exit_code=$?
+  if [[ -n $exit_code ]]; then
+    echo "docker-compose exited with non-zero code $exit_code, showing logs:"
+    docker-compose logs
+    exit "$exit_code"
+  fi
   (docker-compose ps --services --filter status=stopped | grep "^[[:space:]]*$") || (docker-compose logs; sudo docker-compose ps; exit 1)
 }
 


tools/test_containers_compose
```diff
@@ -20,7 +20,7 @@ wait_until() {
 
 setup_containers() {
   for retry in {2..0}; do sudo docker-compose build && break; echo "Remaining retries $retry"; done
-  sudo MOJO_CLIENT_DEBUG=1 docker-compose up -d
+  sudo MOJO_CLIENT_DEBUG=1 docker-compose up -d || docker-compose logs
```
okurz, line 23: I assume that `docker-compose logs` would return 0 as exit code so the failure of the previous command would not abort the script execution

okurz (changes requested):

tools/test_containers_compose
ilausuch: mm yes, you're correct

ilausuch: Fixed

tools/test_containers_compose
```diff
@@ -20,7 +20,7 @@ wait_until() {
 
 setup_containers() {
   for retry in {2..0}; do sudo docker-compose build && break; echo "Remaining retries $retry"; done
-  sudo MOJO_CLIENT_DEBUG=1 docker-compose up -d
+  sudo MOJO_CLIENT_DEBUG=1 docker-compose up -d || docker-compose logs && exit 1
```
Martchus, line 23: This is wrong as well. We don't want to exit with code 1 when `docker-compose up` succeeded. (Mind the operator precedence. Likely it is best to use
```
if ! sudo …; then
    exit_code=$?
    echo "docker-compose exited with non-zero code $exit_code, showing logs:"
    docker-compose logs
    exit "$exit_code"
fi
```
here.)

perlpunk, line 23: This will also execute the `exit 1` if the `docker-compose` command succeeds

Martchus (changes requested):

tools/test_containers_compose
Martchus: Here it looks you were faster (but I only saw your comment after submitting mine) :-)

ilausuch: Thanks fixed.

okurz: and please provide tests for all introduced branches. If not possible automatically then please describe the manual tests done for verification in the git commit message

tools/test_containers_compose
```diff
@@ -20,7 +20,12 @@ wait_until() {
 
 setup_containers() {
   for retry in {2..0}; do sudo docker-compose build && break; echo "Remaining retries $retry"; done
-  sudo MOJO_CLIENT_DEBUG=1 docker-compose up -d
+  if ! sudo MOJO_CLIENT_DEBUG=1 docker-compose up -d; then
```
Martchus, line 23: Looks like you added the code I've suggested. Unfortunately, after reading and testing it myself again, I don't think it works (`exit_code` will be 0). This should work, however:

```suggestion
  sudo MOJO_CLIENT_DEBUG=1 docker-compose up -d || exit_code=$?
  if [[ -n $exit_code ]]; then
```

tools/test_containers_compose
```diff
@@ -20,7 +20,12 @@ wait_until() {
...
+    exit_code=$?
```
Martchus, line 24: Of course you need to remove `exit_code=$?` as well:

```suggestion
  sudo MOJO_CLIENT_DEBUG=1 docker-compose up -d || exit_code=$?
  if [[ -n $exit_code ]]; then
```

tools/test_containers_compose
ilausuch: Thanks, changed

Martchus (approved):

tools/test_containers_compose
```diff
@@ -20,7 +20,12 @@ wait_until() {
 
 setup_containers() {
   for retry in {2..0}; do sudo docker-compose build && break; echo "Remaining retries $retry"; done
-  sudo MOJO_CLIENT_DEBUG=1 docker-compose up -d
+  sudo MOJO_CLIENT_DEBUG=1 docker-compose up -d || exit_code=$?
+  if [[ -n $exit_code ]]; then
```
perlpunk, line 24: Now we're getting this error message:
```
tools/test_containers_compose: line 24: exit_code: unbound variable
```
Since the script does `set -u`, variables have to be set in every case (even if it's just an empty string).
So I suggest a
```
exit_code=0
```
before that line.

okurz (changes requested):
and please provide tests for all introduced branches. If not possible automatically then please describe the manual tests done for verification in the git commit message

tools/test_containers_compose
ilausuch: Done, thanks.

perlpunk (approved):

okurz (approved):

ilausuch:
I rebased from master

PR #4039, opened 2021-07-12
Containers: Fix use of su by providing shell for geekotest

ilausuch:
With this commit we ensure that the geekotest can execute commands,
and therefore we could use su to execute commands with this user.

During the investigation of the issue https://progress.opensuse.org/issues/95374
The execution of all the comands in run_openqa.sh script that were executed by
geekotest user (using su), exited with the error "This account is currently not available."

The reason for that is because that the user geekotest exists but doesn't have a default
shell assigned.

`webui_db_init_1 | geekotest:x:479:479:openQA user:/dev/null:/sbin/nologin`

https://progress.opensuse.org/issues/95374

### DIFF
--- a/container/webui/run_openqa.sh
+++ b/container/webui/run_openqa.sh
@@ -39,6 +39,8 @@ function all_together_apache() {
   su geekotest -c /usr/share/openqa/script/openqa-webui-daemon
 }
 
+usermod --shell /bin/sh geekotest
+
 # run services
 case "$MODE" in
   scheduler ) scheduler;;


container/webui/run_openqa.sh
```diff
@@ -39,6 +39,8 @@ function all_together_apache() {
   su geekotest -c /usr/share/openqa/script/openqa-webui-daemon
 }
 
+usermod --shell /bin/sh geekotest
```
okurz, line 42: I don't understand. If the user "geekotest" does not exist, how does this command help?

okurz (changes requested):

container/webui/run_openqa.sh
ilausuch: During the tests I did geekotest exists but doesn't have a console assigned
`
webui_db_init_1  | geekotest:x:479:479:openQA user:/dev/null:/sbin/nologin
`

ilausuch: This generates the problem: "This account is currently not available."

okurz: hm, ok. Please phrase your git commit message subject line to say what your change fixes, describing the reason for the change, not the "what changed"

ilausuch: Done

Martchus: I guess a commit message like "containers: Fix use of `su` by providing shell for geekotest". Also note that you're changing the *shell* here, not the console.

ilausuch: Changed

okurz (changes requested):
there are some typos in the git commit message. Please also make "progress.opensuse.org/issues/95374" a complete url, i.e. https://…

ilausuch:
I think I fixed all the things and now is in green

okurz (approved):

Martchus (approved):

PR #4042, opened 2021-07-13
ESLint + Prettier

kraih:
This is an alternative approach to ##3900 for updating our JavaScript style. [Prettier](https://prettier.io) itself is purely about code style (like `perltidy`), but it integrates seamlessly into ESLint. That means we can start out with rules that only apply to code style, like consistent quotes and code formatting, and over time add more restrictive rules, such as enforcing the use of `const` instead of `var`. This variant shares the same advantages and disadvantages as the other draft PR.

Locally you need to have `node` installed, and then you can install eslint with `npm`:
```
$ npm install
...
```
With prettier all formatting issues can be resolved automatically (the `fix` script is part of the `package.json` file).
```
npm run fix
...
```
And you can once again use the `lint` script to check for rule violations:
```
npm run lint
...
```

This could be a considered a complete proposal for how we get started with introducing a more strict JavaScript style. Since more ESLint rules can be enabled one by one, or with categories such as `eslint:recommended`.


Makefile
```diff
@@ -264,7 +260,7 @@ prepare-and-launch-docker-to-run-tests-within: docker-test-build launch-docker-t
 
 # all additional checks not called by prove
 .PHONY: test-checkstyle-standalone
-test-checkstyle-standalone: test-shellcheck test-yaml test-critic test-js-style
+test-checkstyle-standalone: test-shellcheck test-yaml test-critic
```
Martchus, line 263: It would be good to have some alternative (not necessary a Makefile target, a script is fine as well) and documentation how to call it.

Martchus (commented):
I guess it is generally good.

kraih:
Actually, for demonstration purposes i've added another commit that enables the [prefer-const](https://eslint.org/docs/rules/prefer-const) rule, that enforces `const` to be used instead of `let` when a value never changes. These additional rules can be enabled as "error" to enforce them, or merely a "warning", to have a more gradual transition.

Makefile
kraih: The main question here is how we want to handle the `node` dependency (and `npm install` call to install the npm modules).

kalikiana: We can bail out if `npm` is not available. We don't have special support for `prove` or other tools in the Makefile either

kraih:
You've already seen what the ESLint annotation look like on GitHub with the previous PR. Here's also a VSCode screenshot to demonstrate IDE integration.
vscode

Makefile
kraih: `prove` ships with Perl though.

Martchus: Maybe let's just add a tidy script under `tools/` which invokes whatever it needs to invoke and prints an error message with installation instructions if it isn't there. It would be good if it would check for locally installed packages first and for globally installed packages second. (So it isn't required to install as root.)

For the CI it might make sense to install the dependency when creating the container.

`npm` itself could be added as dependency of `openQA-devel`.

.github/workflows/javascript.yml
```diff
@@ -0,0 +1,23 @@
...
+    steps:
+    - uses: actions/checkout@v1
+    - name: Use Node ${{ matrix.node-version }}
+      uses: actions/setup-node@v1
+      with:
+        node-version: ${{ matrix.node-version }}
+    - name: npm install
+      run: npm i
+    - name: npm run lint
+      run: npm run lint
```
okurz, line 23: As suggested in https://github.com/os-autoinst/openQA/pull/4042/files#r668812082 please provide something that can be called by users and CI agnostic. We don't want to be dependant on any specific CI service too much, regardless if it's travisCI, circleCI or GHA so better keep github actions as slim as possible.

okurz (changes requested):

perlpunk:
++ for the linebreak at the end of the file.

.github/workflows/javascript.yml
Martchus: The integration would be nice to have but yes, a generic script should be provided in any case.

Makefile
kraih: The script idea makes sense for local development. About CI we don't really have to worry since that's running as a GitHub Action with an Ubuntu container and `actions/setup-node` anyway (for the GitHub annotations).

.github/workflows/javascript.yml
kraih: I'll add a script for local development, but the GitHub Action can't be changed much, since `actions/setup-node` (which is made by GitHub) provides the GitHub annotation features.

kraih: The script will probably be as simple as:
```sh
if ! command -v npm &> /dev/null
then
    echo "Please install node to use the js-tidy script"
    exit
fi
npm install
npm run fix
```

kraih:
I've added a minimal `tools/js-tidy` script that can also run via `make js-tidy`.

docs/Contributing.asciidoc
```diff
@@ -57,6 +57,7 @@ before every commit call:
 [source,sh]
 ----
 make tidy
+make js-tidy
```
okurz, line 60: We don't want everyone to need to call *two* commands. Why not just call `js-tidy` as part of tidy then? Possibly rename the existing "tidy" to "tidy-perl" and have an aggregate make target

tools/js-tidy
```diff
@@ -0,0 +1,30 @@
+#!/bin/bash
+
```
okurz, line 2: Please add `set -euo pipefail`

kraih: Fixed.

docs/Contributing.asciidoc
kraih: Fixed.

Martchus (approved):

okurz (approved):

PR #4049, opened 2021-07-14
docs: Add a test example written in Python

kalikiana:
See: https://progress.opensuse.org/issues/91257

### DIFF
--- a/docs/WritingTests.asciidoc
+++ b/docs/WritingTests.asciidoc
@@ -157,6 +157,37 @@ sub run() {
 }
 --------------------------------------------------------------
 
+In addition to Perl, it is possible to write tests in Python. Although the
+boilerplate is slightly different the function names are the same.
+
+[caption="Example: "]
+.Test for the openQA web UI written in Python
+[source,python]
+--------------------------------------------------------------
+from testapi import *
+
+
+def run(self):
+    assert_screen('openqa-logged-in')
+    assert_and_click('openqa-search')
+    type_string('shutdown.pm')
+    send_key('ret')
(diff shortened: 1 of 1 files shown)


docs/WritingTests.asciidoc
```diff
@@ -157,6 +157,37 @@ sub run() {
 }
 --------------------------------------------------------------
 
+In addition to Perl, it is possible to write tests in Python. Although the
+boilerplate is slightly different the function names are the same.
+
+[caption="Example: "]
+.Test for the openQA web UI written in Python
```
okurz, line 164: hm, I thought of referencing the existing code by URL instead of providing it directly here but I guess the example is actually nice as is

okurz (approved):

okurz:
lib/OpenQA/Scheduler.pm did not show flaky coverage in before and this PR does not touch it. Please report an issue about that

Martchus (commented):
I'd still call it experimental, e.g. I'm wondering how/whether `wait_screen_change` can be called (as it expects a coderef). Just so nobody is disappointed because some functions are not callable in the end.

kalikiana:
> I'd still call it experimental, e.g. I'm wondering how/whether `wait_screen_change` can be called (as it expects a coderef). Just so nobody is disappointed because some functions are not callable in the end.

I intend to publish a little blog post about it as well. Not sure we need to make it extra scary here. 🤔 

I filed [poo#95533](https://progress.opensuse.org/issues/95533), though.

Martchus (approved):
Mh… ok. I don't care that much about it.

PR #4050, opened 2021-07-15
containers: Install ruby and rubygem-sass

ilausuch:
We observed that some openqa tests fails with this error:
`Mojolicious::Plugin::AssetPack::Pipe::Sass requires ruby`

With this commit we ensure that the required libraries are
installed in the container.

We launched this test https://openqa.opensuse.org/tests/1844018 10 times without errors

https://progress.opensuse.org/issues/95179

### DIFF
--- a/container/webui/Dockerfile
+++ b/container/webui/Dockerfile
@@ -7,8 +7,10 @@ LABEL version="0.3"
 RUN zypper ar -p 95 -f http://download.opensuse.org/repositories/devel:openQA/openSUSE_Leap_15.2 devel_openQA && \
     zypper ar -p 90 -f http://download.opensuse.org/repositories/devel:openQA:Leap:15.2/openSUSE_Leap_15.2 devel_openQA_Leap && \
     zypper --gpg-auto-import-keys ref && \
-    zypper in -y ca-certificates-mozilla curl && \
-    zypper in -y --force-resolution openQA-local-db apache2 hostname which w3m && \
+    # openQA might need rubygem:sass to compile assets in this context due to
+    # unknown reasons even though the package should use the precompiled
+    # assets from cache.tar
+    zypper in -y --force-resolution ca-certificates-mozilla curl openQA-local-db apache2 hostname which w3m 'rubygem(sass)' && \
     zypper clean && \
     gensslcert && \
     a2enmod headers && \


container/webui/Dockerfile
```diff
@@ -9,6 +9,7 @@ RUN zypper ar -p 95 -f http://download.opensuse.org/repositories/devel:openQA/op
     zypper --gpg-auto-import-keys ref && \
     zypper in -y ca-certificates-mozilla curl && \
     zypper in -y --force-resolution openQA-local-db apache2 hostname which w3m && \
+    zypper in -y --force-resolution ruby2.5 ruby2.5-rubygem-sass &&\
```
Martchus, line 12: This would be in accordance with our normal dependencies:
```suggestion
    zypper in -y --force-resolution 'rubygem(sass)' &&\
```
(Installing a rubygem should pull-in ruby automatically.)

ilausuch:
Check this test https://openqa.opensuse.org/tests/1842347
I will launch this several times to check the results

ilausuch:
After test this 10 times in openqa without errors, I consider this ready for review.

perlpunk (approved):

container/webui/Dockerfile
```diff
@@ -9,6 +9,7 @@ RUN zypper ar -p 95 -f http://download.opensuse.org/repositories/devel:openQA/op
     zypper --gpg-auto-import-keys ref && \
     zypper in -y ca-certificates-mozilla curl && \
     zypper in -y --force-resolution openQA-local-db apache2 hostname which w3m && \
+    zypper in -y --force-resolution 'rubygem(sass)' &&\
```
okurz, line 12: It shouldn't be necessary to handle these dependencies here manually. Do we need to handle that better in the spec file?

okurz (changes requested):

container/webui/Dockerfile
coolo: the spec file does not require sass as the assets are precompiled as cache.tar

Martchus: Yes, and hence it shouldn't be required for the container setup as it is using rpms. Tests show that it nevertheless sometimes complains about the missing dependency. I'd say that's actually a bug and this PR is merely a workaround.

coolo: My point is, if you accept that the workaround is required, then actually better place it where it's broken (in here)

okurz: Yes, but with a comment explaining the problem and the workaround. If there is "nothing wrong" in the spec file then maybe we can still improve how the situation is handled in our code. As installing in an openSUSE container is for me just one example of what users might do so they could run into the same problem otherwise as well.

ilausuch: I add a comment. Is not too much extend because I tried to summarize what you comment here

perlpunk: typo, it says `rubygetm:sass`

ilausuch: Fixed, I already fixed in my code but I didn't save this version. Thanks

container/webui/Dockerfile
```diff
@@ -9,6 +9,9 @@ RUN zypper ar -p 95 -f http://download.opensuse.org/repositories/devel:openQA/op
     zypper --gpg-auto-import-keys ref && \
     zypper in -y ca-certificates-mozilla curl && \
     zypper in -y --force-resolution openQA-local-db apache2 hostname which w3m && \
+    # The openQA needs rubygem:sass to precompile the assets, otherways there are errors during the tests eventually 
+    # But there is nothing wrong with the spec file because it does not require sass as the assets are precompiled as cache.tar
```
okurz, line 13:
```suggestion
    # openQA might need rubygem:sass to compile assets in this context due to
    # unknown reasons even though the package should use the precompiled
    # assets from cache.tar
```

I also suggest to combine all three `zypper in` calls. Doing all in one step can be a tiny bit faster than three steps of installing

okurz (changes requested):

okurz (approved):

ilausuch:
I did the suggested change in the test and also put the zypper installations in one single line

PR #4051, opened 2021-07-15
search: Search Python filenames and contents

kalikiana:
The code actually hard-codes *.pm to avoid picking up other files but this should include *.py since test modules can be written in Python.

Related: [poo#91257](https://progress.opensuse.org/issues/91257)

### DIFF
--- a/templates/webapi/search/search.html.ep
+++ b/templates/webapi/search/search.html.ep
@@ -9,7 +9,7 @@
     <h2 id="results-heading">Search results</h2>
     <p>The search currently finds <b>job templates</b> by name or description,
        <b>job modules</b> by filename,
-       or Perl modules within the test distributions,
+       or test modules within the test distributions,
        either by <b>filename</b> or <b>source code</b>.</p>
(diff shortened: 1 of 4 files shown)


templates/webapi/search/search.html.ep
```diff
@@ -18,4 +18,4 @@
     </p>
     <div id="results" class="list-group">
     </div>
-</div>
\ No newline at end of file
+</div>
```
Martchus, line 21: Looks unintended (but also not really a problem).

Martchus (approved):
Looks generally good.

okurz (approved):

perlpunk (approved):

templates/webapi/search/search.html.ep
kalikiana: Indentation is unchanged and matches the opener... so not sure there's an issue here

kalikiana:
FYI: Was affected by [poo#96636](https://progress.opensuse.org/issues/96636)

PR #4056, opened 2021-07-19
[WIP] Grant access (rw) to geekotest user to /data subdirectories

ilausuch:
We found that eventually the test single_container_webui fails because the webui
cannot create the needed directories in /root/data/factory

The webui runs with the geekotest user. This commit ensure that the geekotest
has permissions to the working directory /data

https://progress.opensuse.org/issues/95179

### DIFF
--- a/container/webui/run_openqa.sh
+++ b/container/webui/run_openqa.sh
@@ -40,6 +40,7 @@ function all_together_apache() {
 }
 
 usermod --shell /bin/sh geekotest
+[ -d /data ] && chown -R geekotest /data    
 
 # run services
 case "$MODE" in


ilausuch:
Testing....

container/webui/run_openqa.sh
```diff
@@ -40,6 +41,7 @@ function all_together_apache() {
 }
 
 usermod --shell /bin/sh geekotest
+chown -R geekotest /data
```
Martchus, line 44: This directory is shared between (composed) containers, right? Do we somehow ensure that `geekotest` always has the same underlying user ID across all (composed) containers? I suppose this would be required for this line to work.

Note that the spec file alone does *not* seem to ensure that:

```
%pre
if ! getent passwd geekotest > /dev/null; then
    %{_sbindir}/useradd -r -g nogroup -c "openQA user" \
        -d %{_localstatedir}/lib/openqa geekotest 2>/dev/null || :
fi
```

You could of course extend the `usermod` command above:

```
usermod -u 2005 --shell /bin/sh geekotest
```

(I haven't tested the usermod command and it is still a bit hacky since it relies on 2005 not being already used.)

okurz: I don't think we want to hardcode an expected UID in a spec file

container/webui/run_openqa.sh
```diff
@@ -1,5 +1,6 @@
 #!/bin/bash
 set -e
+set -x
```
okurz, line 3: I guess you added that for temporary debugging

okurz (changes requested):

container/webui/run_openqa.sh
ilausuch: yes, I'll remove this

container/webui/run_openqa.sh
Martchus: I would also not hard-code the UID in the spec file (but possibly hard-code it in the container script as suggested unless we have a better idea).

Btw, I've checked some systems with openQA packages installed and the UID indeed varies a little bit between them. However, maybe that's not an issue here because we just spawn multiple instances of the very same container image?

PR #4060, opened 2021-07-20
Generalize CI base container path

okurz:
* Generalize CI base container path


Martchus (approved):

external/os-autoinst-common/tools/update-deps
```diff
@@ -193,7 +193,7 @@ Usage:
     # update cpanfile and dist/rpm/os-autoinst.spec
     $0
     $0 --specfile dist/rpm/os-autoinst.spec
-    $0 --dockerfile container/travis_test/Dockerfile
+    $0 --dockerfile container/ci/Dockerfile
```
perlpunk, line 196: * This is supposed to be an example, so it doesn't have to match what we are using.
* Since it's in the subrepo, can you do the change there first, and we try if a subrepo pull works this time?

okurz: sure, done that in https://github.com/os-autoinst/os-autoinst-common/pull/10

okurz: included a subrepo pull instead of the openQA-local commit

external/os-autoinst-common/tools/update-deps
```diff
@@ -69,8 +69,9 @@ sub update_dockerfile {
     my $end = '# AUTODEPS END';
     my $run = <<"EOM";
 # This part is autogenerated by $scriptname from $yamlfile
+# hadolint ignore=DL3034,DL3037
 RUN zypper in -y -C \\
-$dep   && true
+$dep   && zypper clean
```
okurz, line 74: I thought we had this in already but maybe I got confused and in the end the according subrepo pull maybe has been only applied for os-autoinst so far.

external/os-autoinst-common/tools/update-deps
perlpunk: for some reason in 5d84f326c3caecdd10550f2b17fd2a201a1d86fd the `parent` commit wasn't updated, and I suspect that is the source of the merge conflict.
Here `parent` is updated, so I hope a normal pull will work next time.
Like I mentioned previously, a `clone -f` is usually the easiest way (just use the command which was originally used) in case of any conflicts.

perlpunk (approved):

PR #4062, opened 2021-07-20
Cover wait_with_progress with basic tests

kraih:
The `wait_with_progress` function in `OpenQA::Utils` is only used in one place. That place got test coverage [a few days ago](https://github.com/os-autoinst/openQA/commit/0858fdf42e4482808819c7b90bb202b9e37198f8), but `wait_with_progress` was mocked for the test. So it has no coverage itself currently.

Progress: https://progress.opensuse.org/issues/95170

### DIFF
--- a/lib/OpenQA/Worker/Job.pm
+++ b/lib/OpenQA/Worker/Job.pm
@@ -1068,7 +1068,6 @@ sub _upload_log_file {
     my $regular_upload_failed = 0;
     my $retry_counter         = 5;
     my $retry_limit           = 5;
-    my $tics                  = 5;
     my $res;
     my $client = $self->client;
     my $url    = $client->url;
@@ -1080,12 +1079,8 @@ sub _upload_log_file {
         my $tx = $ua->build_tx(POST => $ua_url => form => $upload_parameter);
 
         if ($regular_upload_failed) {
-            log_warning(
-                sprintf(
-                    'Upload attempts remaining: %s/%s for %s, in %s seconds',
-                    $retry_counter--, $retry_limit, $filename, $tics
-                ));
-            OpenQA::Utils::wait_with_progress($tics);
+            log_warning(sprintf('Upload attempts remaining: %s/%s for %s', $retry_counter--, $retry_limit, $filename));
+            sleep 1;
         }
 
         $res = $ua->start($tx);
--- a/lib/OpenQA/Worker/Job.pm
+++ b/lib/OpenQA/Worker/Job.pm
@@ -1080,7 +1080,7 @@ sub _upload_log_file {
 
         if ($regular_upload_failed) {
             log_warning(sprintf('Upload attempts remaining: %s/%s for %s', $retry_counter--, $retry_limit, $filename));
-            sleep 1;
+            sleep UPLOAD_DELAY;
         }
 
         $res = $ua->start($tx);
--- a/t/24-worker-jobs.t
+++ b/t/24-worker-jobs.t
@@ -15,11 +15,11 @@
 # You should have received a copy of the GNU General Public License along
 # with this program; if not, see <http://www.gnu.org/licenses/>.
 
+use Test::Most;
+
 BEGIN {
     $ENV{OPENQA_UPLOAD_DELAY} = 0;
 }
-
-use Test::Most;
 
 use FindBin;
 use lib "$FindBin::Bin/lib", "$FindBin::Bin/../external/os-autoinst-common/lib";
--- a/t/24-worker-jobs.t
+++ b/t/24-worker-jobs.t
@@ -1369,8 +1369,6 @@ subtest 'asset upload' => sub {
 };
 
 subtest 'log file upload' => sub {
-    my $utils_mock = Test::MockModule->new('OpenQA::Utils');
-    $utils_mock->noop('wait_with_progress');
     my $ua_mock = Test::MockModule->new('Mojo::UserAgent');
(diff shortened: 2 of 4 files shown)


kraih:
If this function needs to exist in the first place is of course a different discussion (and i'd probably say no, it does not), but as long as it does it's better if it has test coverage.

okurz (changes requested):
Interesting findings. Your change looks all fine but I agree that the function does not look like really needed. I support simply removing it.

okurz:
@Martchus

kraih:
Simply removing it probably does not work quite right in production, the `sleep` is most likely still needed there even though the test case mocks it out to speed up the test. This is the downside of mocking too much.

Martchus:
This function has been introduced in e9f2d1c5fffb81893c2a09269dc730695f5afa50 by @foursixnine . I would also remove the function and just put the sleep (for delaying the next upload attempt) where it was called before.

kraih:
Ok, replaced `wait_with_progress` with a `sleep` call.

t/24-worker-jobs.t
```diff
@@ -1369,8 +1369,6 @@ subtest 'asset upload' => sub {
 };
 
 subtest 'log file upload' => sub {
-    my $utils_mock = Test::MockModule->new('OpenQA::Utils');
-    $utils_mock->noop('wait_with_progress');
```
okurz, line 1373: Hm, this likely means unnecessary sleeping in tests unless you mock the sleep call

kraih: The test runs about 4 to 5 seconds slower.

kraih: Since i don't care much for an argument about mocking or not, please just tell me how you want the sleep to be mocked.

kraih: To make this easy I've reused the same ugly global hack we already use in another test. https://github.com/os-autoinst/openQA/pull/4062/commits/bfd335968bd4fb1edb32263acf39637383334878

okurz (approved):

lib/OpenQA/Worker/Job.pm
```diff
@@ -1080,12 +1079,8 @@ sub _upload_log_file {
...
 
         if ($regular_upload_failed) {
-            log_warning(
-                sprintf(
-                    'Upload attempts remaining: %s/%s for %s, in %s seconds',
-                    $retry_counter--, $retry_limit, $filename, $tics
-                ));
-            OpenQA::Utils::wait_with_progress($tics);
+            log_warning(sprintf('Upload attempts remaining: %s/%s for %s', $retry_counter--, $retry_limit, $filename));
+            sleep 1;
```
Martchus, line 1083: This was waiting 5 seconds before. I wouldn't change that. 

It would make sense to make the magic numbers configurable via env variables. There's already `OPENQA_UPLOAD_DELAY` which is read into a constant within the same file so one could just use that constant.

Martchus (changes requested):

lib/OpenQA/Worker/Job.pm
kraih: Resolved.

lib/OpenQA/Worker/Job.pm
```diff
@@ -1080,12 +1079,8 @@ sub _upload_log_file {
...
         if ($regular_upload_failed) {
-            log_warning(
-                sprintf(
-                    'Upload attempts remaining: %s/%s for %s, in %s seconds',
-                    $retry_counter--, $retry_limit, $filename, $tics
-                ));
-            OpenQA::Utils::wait_with_progress($tics);
+            log_warning(sprintf('Upload attempts remaining: %s/%s for %s', $retry_counter--, $retry_limit, $filename));
+            sleep 5;
         }
```
okurz, line 1084: Sorry, but I don't want to accept a sleep with magic number

okurz (changes requested):

lib/OpenQA/Worker/Job.pm
kraih: Please be more specific, we do not have a standard for this.

Martchus: E.g. use the constant `UPLOAD_DELAY` as I suggested in my review comment.

kraih: I don't like constants for this, but ok.

okurz: The constant `UPLOAD_DELAY` is defined in line 49 of the same file

```perl
use constant UPLOAD_DELAY        => $ENV{OPENQA_UPLOAD_DELAY} // 5;
```

I agree with @Martchus that you can just use the same constant so just

```perl
sleep UPLOAD_DELAY;
```

With `UPLOAD_DELAY` being configurable by environment variable you can even simplify the test mocking by just setting the variable `$ENV{OPENQA_UPLOAD_DELAY} = 0;` :)

kraih: Making something configurable only at compile time is still rather clunky.

kraih: Resolved.

Martchus (approved):

okurz (commented):
Hm, I don't see the sleep mocking anymore.

Martchus:
`UPLOAD_DELAY` will be 0 in the test anyways.

okurz (approved):
ah, right. That's good

PR #4063, opened 2021-07-21
worker: Improve error handling of image upload

Martchus:
* See particular commit messages and https://progress.opensuse.org/issues/94531
* Not a full fix for the mentioned issue but a start. So far it merely retries the upload and passes the upload error (after all attempts failed) as `api failure: …` reason but does *not* mark the job as incomplete.


okurz (approved):

Martchus:
Fix for sporadic test failure observed here: https://github.com/os-autoinst/openQA/pull/4068

okurz (approved):

Martchus:
I've been updating #4067 which includes this PR. I don't think it is worth keeping this version.

PR #4066, opened 2021-07-22
Avoid perl warning in string concatenation

perlpunk:
Issue: https://progress.opensuse.org/issues/95730

We try to search for a Job with `$self->id` and no `clone_id`, and update
it with the new `clone_id`.
If no rows are affected then it means that `$self` had already set a
`clone_id` in the database, but we didn't reload the values of self, that's why
it is undef and producing a warning.

Instead we search for the Job again now.

### DIFF
--- a/lib/OpenQA/Schema/Result/Jobs.pm
+++ b/lib/OpenQA/Schema/Result/Jobs.pm
@@ -658,7 +658,7 @@ sub _create_clone {
     # not be created, somebody else was faster at cloning)
     my $orig_id       = $self->id;
     my $affected_rows = $rset->search({id => $orig_id, clone_id => undef})->update({clone_id => $new_job->id});
-    die "Job $orig_id has already been cloned as " . $self->clone_id . "\n" unless $affected_rows == 1;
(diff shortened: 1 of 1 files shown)


lib/OpenQA/Schema/Result/Jobs.pm
```diff
@@ -658,7 +658,10 @@ sub _create_clone {
     # not be created, somebody else was faster at cloning)
     my $orig_id       = $self->id;
     my $affected_rows = $rset->search({id => $orig_id, clone_id => undef})->update({clone_id => $new_job->id});
-    die "Job $orig_id has already been cloned as " . $self->clone_id . "\n" unless $affected_rows == 1;
+    unless ($affected_rows == 1) {
+        $self->discard_changes;
+        die "Job $orig_id has already been cloned as " . $self->clone_id . "\n";
+    }
```
okurz, line 664: Would that be equivalent to

```perl
die "Job $orig_id has already been cloned as " . $rset->search($orig_id)->clone_id . "\n" unless $affected_rows == 1;
```

perlpunk: No, but something similar. I pushed a change.

okurz (approved):

Martchus (approved):

PR #4069, opened 2021-07-23
format_time_duration: Also display days in test duration

ggardet:
Some very long tests, such as https://openqa.opensuse.org/tests/1851295 (very slow because no KVM acceleration is used) can exceed 24 hours. 
Duration is currently shown as `25:23` minutes. With this patch, it will be shown properly as `1 day 00:25 hours`

### DIFF
--- a/lib/OpenQA/WebAPI/Plugin/Helpers.pm
+++ b/lib/OpenQA/WebAPI/Plugin/Helpers.pm
@@ -34,7 +34,10 @@ sub register ($self, $app, $config) {
         format_time_duration => sub {
             my ($c, $timedate) = @_;
             return unless $timedate;
-            if ($timedate->hours() > 0) {
+            if ($timedate->days() > 0) {
+                sprintf("%d days %02d:%02d hours", $timedate->days(), $timedate->hours(), $timedate->minutes());
+            }
+            elsif ($timedate->hours() > 0) {
                 sprintf("%02d:%02d hours", $timedate->hours(), $timedate->minutes());
             }
             else {
--- a/t/ui/18-tests-details.t
+++ b/t/ui/18-tests-details.t
@@ -17,6 +17,7 @@
 
 use Test::Most;
 
+use DateTime;
 use FindBin;
 use lib "$FindBin::Bin/../lib", "$FindBin::Bin/../../external/os-autoinst-common/lib";
 use Test::Mojo;
@@ -671,6 +672,31 @@ subtest 'archived icon' => sub {
     $jobs->find(99947)->update({archived => 1});
     $t->get_ok('/tests/99947/infopanel_ajax')->status_is(200);
     is $t->tx->res->dom->find('#job-archived-badge')->size, 1, 'archived icon shown if job is archived';
+};
+
+subtest 'test duration' => sub {
+    my $start = DateTime->new(
+        year       => 2021,
+        month      => 9,
+        day        => 14,
+        hour       => 15,
+        minute     => 0,
+        second     => 0,
+        nanosecond => 0,
+        time_zone  => 'UTC',
+    );
+    my $end = DateTime->new(
+        year       => 2021,
+        month      => 9,
+        day        => 16,
+        hour       => 17,
+        minute     => 30,
+        second     => 0,
+        nanosecond => 0,
+        time_zone  => 'UTC',
+    );
+    my $duration = $t->app->format_time_duration($end - $start);
+    like $duration, qr/2 days 02:30 hours/, 'duration formatted';
 };
 
 kill_driver();


Martchus (approved):
Using https://metacpan.org/pod/DateTime::Format::Duration would be nicer but it isn't in Factory so I guess this is good enough for now. (The mix of colon-separated values and units is a bit weird here as well but you didn't start it.)

perlpunk (approved):

lib/OpenQA/WebAPI/Plugin/Helpers.pm
```diff
@@ -34,7 +34,10 @@ sub register ($self, $app, $config) {
         format_time_duration => sub {
             my ($c, $timedate) = @_;
             return unless $timedate;
-            if ($timedate->hours() > 0) {
+            if ($timedate->days() > 0) {
+                sprintf("%d days %02d:%02d hours", $timedate->days(), $timedate->hours(), $timedate->minutes());
+            }
+            elsif ($timedate->hours() > 0) {
                 sprintf("%02d:%02d hours", $timedate->hours(), $timedate->minutes());
             }
```
okurz, line 42: hm, maybe we don't need to repeat the formating for hours. How about doing the hours block in both cases and in case of days add that as prefix. Something along the lines of

```suggestion
            if ($timedate->… > 0) {
                my $out = sprintf("%02d:%02d hours", $timedate->hours(), $timedate->minutes());
                $out = $timedate->days() . ' days ' . $out if $timedate->days();
                return $out;
            }
```

hm, ok. maybe not really better ;) Maybe you have a better idea.

okurz (changes requested):

```bash
$ cat $(which git-pr-last )
#!/bin/sh -e
target="${target:-"$USER"}"
```

lib/OpenQA/WebAPI/Plugin/Helpers.pm
Martchus: Then the condition needs to check for `days > 0 || hours > 0`.

ggardet: I do not feel it is really better to make use of a prefix. No code reduction and harder to read.

okurz (approved):

ggardet:
> What you put into the description of the pull request could also be in the commit message details. 

I did not want to overload the commit message with too much details, and wanted to keep it distro agnostic.

okurz:
> I did not want to overload the commit message with too much details, and wanted to keep it distro agnostic.

you mean you want to avoid the o3 link? You can of course leave that out but your example of "before" and "after" is definitely helpful

ggardet:
> > I did not want to overload the commit message with too much details, and wanted to keep it distro agnostic.
> 
> you mean you want to avoid the o3 link? You can of course leave that out but your example of "before" and "after" is definitely helpful

Fixed.

okurz (approved):

kraih:
I'd like to see tests here, helpers that don't access `$c` are super easy to unit test with `$t->app->format_time_duration(...)`.

ggardet:
> I'd like to see tests here, helpers that don't access `$c` are super easy to unit test with `$t->app->format_time_duration(...)`.

Would you have a simple example/guide?

kraih:
> Would you have a simple example/guide?

Not the best example, but [here](https://github.com/os-autoinst/openQA/blob/1a636225d143f36cb0085d28b98491402dd01bc5/t/ui/27-plugin_obs_rsync.t#L86-L89)'s a similar helper test.

ggardet:
> > Would you have a simple example/guide?
> 
> Not the best example, but [here](https://github.com/os-autoinst/openQA/blob/1a636225d143f36cb0085d28b98491402dd01bc5/t/ui/27-plugin_obs_rsync.t#L86-L89)'s a similar helper test.

Shall I add it to an existing file (which one?) or shall I create a new one?

Martchus:
Yes, add it to an existing file. There's always overhead incurred by launching a new test so we shouldn't split tests too much. Since this is about the tests details page, `t/ui/18-tests-details.t` would make sense.

t/ui/18-tests-details.t
```diff
@@ -673,5 +674,30 @@ subtest 'archived icon' => sub {
...
+        month      => 09,
+        day        => 16,
+        hour       => 17,
+        minute     => 30,
+        second     => 0,
+        nanosecond => 0,
+        time_zone  => 'UTC',
+    );
+    my $duration = $t->app->format_time_duration($end - $start);
+    $duration =~ /2 days 02:30 hours/;
```
Martchus, line 699:
```suggestion
    my $duration = $t->app->format_time_duration($end - $start);
    $duration =~ /2 days 02:30 hours/;
    like $duration, qr/2 days 02:30 hours/, 'duration formatted';
```

And maybe the duration can be instantiated in a more direct and compact way.

ggardet: Applied

Martchus: But the line `$duration =~ /2 days 02:30 hours/;` is still there and seems quite useless. And it is really not possible to instantiate the duration object in a more direct and compact way?

ggardet:
circleci seems to hang. Is it expected?

t/ui/18-tests-details.t
ggardet: Fixed.
I guess we could do it in a more compact way, but I fear to loose readability.

okurz:
I don't know since when the CI jobs hang. Please rebase to retrigger CI jobs.

ggardet:
When I want to rebase and push, I get the following error: 
```
To https://github.com/ggardet/openQA
 ! [remote rejected]     fix_duration -> fix_duration (refusing to allow a Personal Access Token to create or update workflow `.github/workflows/commit_message_checker.yml` without `workflow` scope)
error: failed to push some refs to 'https://github.com/ggardet/openQA'
```

okurz:
> When I want to rebase and push, I get the following error:
> 
> ```
> To https://github.com/ggardet/openQA
>  ! [remote rejected]     fix_duration -> fix_duration (refusing to allow a Personal Access Token to create or update workflow `.github/workflows/commit_message_checker.yml` without `workflow` scope)
> error: failed to push some refs to 'https://github.com/ggardet/openQA'
> ```

Hm, weird. Anyway, I did the rebase now. Let's see …

PR #4070, opened 2021-07-23
t: Scale timeouts in scalability test as it is slower when collecting coverage

Martchus:
This should make the test more stable (see
https://progress.opensuse.org/issues/95851).

Avoiding `Devel::Cover::report()` in forked process makes almost no
difference so I'm just applying the normal scaling of timeouts here (which
fits the slowdown observed locally when comparing a `make test …` run with
a `make coverage …` run).

### DIFF
--- a/t/43-scheduling-and-worker-scalability.t
+++ b/t/43-scheduling-and-worker-scalability.t
@@ -135,9 +135,9 @@ $job_ids{$jobs->create({@job_settings, TEST => "dummy-$_"})->id} = 1 for 1 .. $j
 my $casedir = testcasedir($distri, $version, undef);
 path($casedir)->make_path unless -d $casedir;
 
-my $seconds_to_wait_per_worker = 5.0;
-my $seconds_to_wait_per_job    = 2.5;
-my $polling_interval           = 0.1;
+my $seconds_to_wait_per_worker = OpenQA::Test::TimeLimit::scale_timeout 5.0;
+my $seconds_to_wait_per_job    = OpenQA::Test::TimeLimit::scale_timeout 2.5;
+my $polling_interval           = Devel::Cover->can('report') ? 0.2 : 0.1;
 my $polling_tries_workers      = $seconds_to_wait_per_worker / $polling_interval * $worker_count;
 my $polling_tries_jobs         = $seconds_to_wait_per_job / $polling_interval * $job_count;
 


perlpunk:
> Avoiding `Devel::Cover::report()` in forked process makes almost no
> difference

Yeah, reporting should not cost much time, but the process is still running with coverage enabled, and that is what costs time.
So far I don't know if there is a possibility to switch of running in coverage mode in the middle of the process.

Martchus:
Ah, ok. But why does it then often help to reduce the forking completely and run the code instead in the main process? Then the same amount of code is executed - just everything in one process. I really though merging the report from sub processes would be the problem.

By the way, the worker sub processes are started by invoking Perl again (via `IPC::Run`: `start ['perl', $worker_path, "--instance=$instance", @worker_args];`). I suppose in this case no coverage is collected for these processes at all (which would only leave the web UI and web sockets process here anyways).

perlpunk:
> By the way, the worker sub processes are started by invoking Perl again (via `IPC::Run`: `start ['perl', $worker_path, "--instance=$instance", @worker_args];`). I suppose in this case no coverage is collected for these processes at all (which would only leave the web UI and web sockets process here anyways).

Good that you mention it.
Since coverage is started with setting `PERL5OPT`, the variable will also be visible in programs started with a system call.
In this case I would do `local $ENV{PERL5OPT} = "";` before the call, like in 
https://github.com/os-autoinst/openQA/blob/master/lib/OpenQA/Worker/Engines/isotovideo.pm#L444

kraih:
I can say with certainty that many forked processes will block for a long time on `Devel::Cover::report()`. Especially with Minion jobs that is very visible.

okurz (approved):

Martchus:
@perlpunk That worked (see https://github.com/os-autoinst/openQA/pull/4072). And I guess it also worked to avoid `Devel::Cover::report()` as it saved me almost a second. It is just that the main culprit here are the worker processes.

I'm marking this PR as not-ready because if https://github.com/os-autoinst/openQA/pull/4072 works well we likely want to merge it instead of this one.

kraih (approved):

okurz:
please check what from here could still be needed after https://github.com/os-autoinst/openQA/pull/4072 merged

PR #4072, opened 2021-07-23
t: Disable coverage report for forked processes in scalability test

Martchus:
Preferred alternative to https://github.com/os-autoinst/openQA/pull/4070 (as it actually makes the test faster).

---

This speeds up the test significantly when coverage analysis is enabled
and fix https://progress.opensuse.org/issues/95851.

Before:
```
NAME                        TIME        CUMULATIVE      PERCENTAGE
 wait for workers to be idle 3.113       3.113           24.538%
 assign and run jobs         6.884       9.997           54.265%
 stop all workers            2.689       12.686          21.198%
```

After:
```
NAME                        TIME        CUMULATIVE      PERCENTAGE
 wait for workers to be idle 2.530       2.530           57.529%
 assign and run jobs         1.838       4.368           41.803%
 stop all workers            0.029       4.397           0.667%
```

### DIFF
--- a/t/43-scheduling-and-worker-scalability.t
+++ b/t/43-scheduling-and-worker-scalability.t
@@ -73,8 +73,8 @@ my $workers = $schema->resultset('Workers');
 my $jobs    = $schema->resultset('Jobs');
 
 # create web UI and websocket server
-my $web_socket_server = create_websocket_server(undef, 0, 1, 1);
-my $webui             = create_webapi(undef, sub { });
+my $web_socket_server = create_websocket_server(undef, 0, 1, 1, 1);
+my $webui             = create_webapi(undef, 1);
 
 # prepare spawning workers
 my $sharedir        = setup_share_dir($ENV{OPENQA_BASEDIR});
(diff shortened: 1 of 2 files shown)


t/43-scheduling-and-worker-scalability.t
```diff
@@ -99,6 +99,7 @@ note("Spawning $worker_count workers");
 sub spawn_worker {
     my ($instance) = @_;
 
+    local $ENV{PERL5OPT} = '';                                             # uncoverable statement
```
okurz, line 102: I guess we also want a source file comment for *that*. The alternative of course is always an explanatory function name

```suggestion
sub disable coverage_collection_in_spawned_processes {
    local $ENV{PERL5OPT} = '';                                             # uncoverable statement
}
…
sub spawn_worker {
…
    disable coverage_collection_in_spawned_processes;
```

okurz (changes requested):

t/43-scheduling-and-worker-scalability.t
kraih: I like the separate function, makes it more obvious. Especially if we do the same in more places.

okurz (approved):

PR #4073, opened 2021-07-23
Mark most of the unstable worker test code as uncoverable

kraih:
The test is not all that flaky when i run it locally, but in 20 runs i did see 3 cases where the unstable worker code did not run. We'll have to see after a few Circle CI runs if that does the trick.

Progress: https://progress.opensuse.org/issues/80274


okurz (approved):
ok, so I guess the unstable worker is unstable after all ;)

kraih:
After 5 runs on Circle CI i'm not seeing any flakiness anymore. So this probably works.

Amrysliu (approved):

PR #4074, opened 2021-07-26
Increase the waiting time for job results in scheduler-full test

ilausuch:
We observed that t/05-scheduler-full.t fails more often than
it used too (it is in "unstable" already)

In one of the cases we wait for a job to be incompleted. We use
a active wait with a timeout, but seems that in environments with
a high load this timeout is not enough.

This PR increase the timeout from 20 seg. to 30 seg.

https://progress.opensuse.org/issues/95848

### DIFF
--- a/t/05-scheduler-full.t
+++ b/t/05-scheduler-full.t
@@ -213,7 +213,7 @@ subtest 're-scheduling and incompletion of jobs when worker rejects jobs or goes
     note 'waiting for job to be incompleted';
     for (0 .. 100) {
         last if $jobs->find(99982)->state eq OpenQA::Jobs::Constants::DONE;
-        sleep .2;
+        sleep .3;
     }
 
     my $job = $jobs->find(99982);


t/05-scheduler-full.t
```diff
@@ -213,7 +213,7 @@ subtest 're-scheduling and incompletion of jobs when worker rejects jobs or goes
     note 'waiting for job to be incompleted';
     for (0 .. 100) {
         last if $jobs->find(99982)->state eq OpenQA::Jobs::Constants::DONE;
-        sleep .2;
+        sleep .3;
```
okurz, line 216: this is not "20" or "30" seconds but 0.3s, i.e. 300ms. And I don't see how this should fix any instabilities as we run this in a for loop. As this is about a sporadic issue – flaky test as the ticket says – please first get the fail ratio before and then after your fix, not just one change stating "it works"

okurz (changes requested):

t/05-scheduler-full.t
perlpunk: well, if you multiply by 100 (or actually 101), the number of loop iterations, you land at around 20 / 30 seconds.

okurz: Yes, but if that is desired we can also increase the loop range

perlpunk: I was only commenting on the `this is not "20" or "30" seconds`, which is wrong.

ilausuch:
There is a best solution https://github.com/os-autoinst/openQA/pull/4094

PR #4081, opened 2021-07-28
Make message about clone error unique

okurz:
Previously we used the same message in two locations

### DIFF
--- a/lib/OpenQA/Schema/Result/Jobs.pm
+++ b/lib/OpenQA/Schema/Result/Jobs.pm
@@ -658,7 +658,7 @@ sub _create_clone {
     # not be created, somebody else was faster at cloning)
     my $orig_id       = $self->id;
     my $affected_rows = $rset->search({id => $orig_id, clone_id => undef})->update({clone_id => $new_job->id});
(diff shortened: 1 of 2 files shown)


Martchus (commented):
From the user's perspective it is the same problem but I agree that it can be helpful for developers to distinguish the errors. I would remove the `Unable to clone: ` suffix because the context should already be pretty clear here. (The entire purpose of the duplicate function/route is cloning jobs.)

okurz:
@Martchus done

something on a meta-level. I think it's better to more often use the inline review comments because then the reviewee can respond to a message more easily, we can "resolve" individual conversations while other issues in a PR might still be relevant, etc.

Martchus (approved):

perlpunk:
I don't understand how the PR makes something uniq.

> Previously we used the same message in two locations

Where are those two locations?

Martchus:
`git grep 'has already been cloned as'` tells you; the error is indeed duplicated.

perlpunk:
It's possible to link to a certain code location :)
https://github.com/os-autoinst/openQA/blob/master/lib/OpenQA/Schema/Result/Jobs.pm#L930

perlpunk (approved):

PR #4084, opened 2021-07-29
Simplify all test selection switches for all "fullstack" tests into one

okurz:
Within Makefile we already run multiple "fullstack" tests together but
so far we have three variables. At a time this was necessary to control
how we start these tests within the old travis CI infrastructure. By
extending "tools/run-tests-within-container" to use prove and taking
multiple test files as parameter we can also simplify all other places
where the variables SCHEDULER_FULLSTACK or DEVELOPER_FULLSTACK were
defined and combine that into the single variable FULLSTACK. This also
simplifies the setup for developers as only a single variable FULLSTACK
needs to be defined if the developer is ready to run the fullstack tests
within their environment.


Martchus (approved):

okurz:
@Martchus why does the scalability test require the Core-7.2.iso if we only run a fake isotovideo?:
https://app.circleci.com/pipelines/github/os-autoinst/openQA/7214/workflows/8a429f08-bbb7-4b4b-9bc9-c0dbd261a7d0/jobs/68117/steps

Martchus:
I'll have a look. (It shouldn't be required.)

okurz:
Updated:
* Rebased on https://github.com/os-autoinst/openQA/pull/4087
* Removed scalability test from list of fullstack tests to prevent double scheduling
* Moved path of unstable_tests.txt to be independent of circleCI

Martchus (approved):

ilausuch (approved):

PR #4088, opened 2021-07-29
Enable eslint:recommended rules related to syntax and logic errors

kraih:
A first followup to #4042. It enables all rules from the `eslint:recommended` set that are related to syntax and logic errors. We had quite a few violations of the [no-prototype-builtins](https://eslint.org/docs/rules/no-prototype-builtins) rule for example.

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -10,7 +10,44 @@
   "extends": ["prettier"],
   "plugins": ["prettier"],
   "rules": {
+    // Formatting
     "prettier/prettier": "error",
+
+    // Syntax and logic errors
+    "for-direction": "error",
+    "getter-return": "error",
+    "no-async-promise-executor": "error",
+    "no-constant-condition": "error",
+    "no-control-regex": "error",
+    "no-debugger": "error",
+    "no-dupe-args": "error",
+    "no-dupe-else-if": "error",
+    "no-dupe-keys": "error",
+    "no-duplicate-case": "error",
+    "no-empty": "error",
+    "no-empty-character-class": "error",
+    "no-ex-assign": "error",
+    "no-extra-boolean-cast": "error",
+    "no-extra-semi": "error",
+    "no-func-assign": "error",
+    "no-import-assign": "error",
+    "no-inner-declarations": "error",
+    "no-invalid-regexp": "error",
+    "no-irregular-whitespace": "error",
+    "no-misleading-character-class": "error",
+    "no-obj-calls": "error",
+    "no-prototype-builtins": "error",
+    "no-regex-spaces": "error",
+    "no-setter-return": "error",
+    "no-sparse-arrays": "error",
+    "no-unexpected-multiline": "error",
+    "no-unreachable": "error",
+    "no-unsafe-finally": "error",
+    "no-unsafe-negation": "error",
+    "use-isnan": "error",
+    "valid-typeof": "error",
+
+    // Additional rules we find useful
     "prefer-const": "error"
   }
 }
--- a/assets/javascripts/job_templates.js
+++ b/assets/javascripts/job_templates.js
@@ -51,7 +51,7 @@ function templateRemoved(chosen, deselected) {
 
 function addFailed(data) {
   // display something without alert
-  if (data.hasOwnProperty('responseJSON')) {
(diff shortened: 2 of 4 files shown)


assets/javascripts/job_templates.js
```diff
@@ -260,7 +260,10 @@ function buildMediumGroup(group, media) {
       select.attr('id', group + '-' + arch + '-' + test);
       select.attr('data-product-id', archs[arch]['_id']);
       select.addClass('chosen-select');
-      if (archs.hasOwnProperty(arch) && archs[arch].hasOwnProperty(test)) {
+      if (
```
Martchus, line 263: Has it applied these changes programmatically?

Martchus (approved):

assets/javascripts/job_templates.js
kraih: No, i had to fix those manually. But it formatted the code afterwards with prettier rules.

kraih: VSCode does make fixing them very easy though. The popup dialog contains a link to the web site that explains why the rule exists and how to fix the violation.
vscode

kraih: I think most of these rules are too tricky for ESLint to fix on its own.

.eslintrc.json
```diff
@@ -11,6 +11,40 @@
   "plugins": ["prettier"],
   "rules": {
     "prettier/prettier": "error",
+
```
okurz, line 14: why this blank line?

.eslintrc.json
```diff
@@ -11,6 +11,40 @@
...
+    "no-regex-spaces": "error",
+    "no-setter-return": "error",
+    "no-sparse-arrays": "error",
+    "no-unexpected-multiline": "error",
+    "no-unreachable": "error",
+    "no-unsafe-finally": "error",
+    "no-unsafe-negation": "error",
+    "use-isnan": "error",
+    "valid-typeof": "error",
+
```
okurz, line 47: why this blank line?

.eslintrc.json
```diff
@@ -11,6 +11,40 @@
...
+    "no-prototype-builtins": "error",
+    "no-regex-spaces": "error",
+    "no-setter-return": "error",
+    "no-sparse-arrays": "error",
+    "no-unexpected-multiline": "error",
+    "no-unreachable": "error",
+    "no-unsafe-finally": "error",
+    "no-unsafe-negation": "error",
+    "use-isnan": "error",
+    "valid-typeof": "error",
```
okurz, line 46: Is it really necessary to specify the complete list by hand?

assets/javascripts/job_templates.js
Martchus: Ok, I suppose that would be too much to ask for :-)

.eslintrc.json
kraih: To separate the categories. These make up the "syntax and logic errors" category.

kraih: Next group will probably be "general best practices".

.eslintrc.json
Martchus: Seems at least better than including the list from somewhere where it might change and then the CI check fails.

.eslintrc.json
Martchus: Likely to separate style-only rules from syntax/logic error rules.

.eslintrc.json
kraih: Until we have `eslint:recommended` fully covered i believe we have to. Later categories will also have individual rules with so many violations that we'll need to make some warnings for a transition time i imagine.

.eslintrc.json
okurz: can you add a comment? Depending on the JSON parser this should work

kraih: Also the lonely `prefer-const` is not actually part of `eslint:recommended` at all, so it will never be folded into the other rules.

.eslintrc.json
okurz: ok, I understand. That makes sense.

.eslintrc.json
kraih: You're right, ESLint allows comments in the JSON. So, added.

perlpunk (approved):

PR #4089, opened 2021-07-29
Make default test runs less noisy

okurz:
Previously we called prove with the parameter "-v" from within the
Makefile which can make tests quite noisy. As prove already outputs
context to errors in case any happen I consider a better default to run
without "-v" but keep the verbose parameter for CI runs, hence this
commit is removing the "-v" from prove arguments by default but adding
them instead in .circleci/config.yml. With the new Makefile variable
"EXTRA_PROVE_ARGS" one can still select the verbose mode when calling
make without needing to replicate any other necessary prove arguments.


Martchus (approved):
I prefer `-v` by default but I guess I can add the variable to my environment.

perlpunk (approved):

PR #4091, opened 2021-07-29
Add regex and tests for fragment identifier recognition

BillAnastasiadis:
The regex used until now would not recognize fragment identifiers (part of the url that comes after '#'), which is now fixed. 

The accepted characters for fragments were derived from it's definition in https://datatracker.ietf.org/doc/html/rfc3986#section-3.5.

### DIFF
--- a/lib/OpenQA/Constants.pm
+++ b/lib/OpenQA/Constants.pm
@@ -20,6 +20,7 @@ use warnings;
 
 use Time::Seconds;
 use Exporter 'import';
+use Regexp::Common 'URI';
 
 # Minimal worker version that allows them to connect;
 # To be modified manuallly when we want to break compatibility and force workers to update
@@ -107,6 +108,8 @@ use constant DB_TIMESTAMP_ACCURACY => 1;
 use constant VIDEO_FILE_NAME_START => 'video.';
 use constant VIDEO_FILE_NAME_REGEX => qr/^.*\/video\.[^\/]*$/;
 
+use constant FRAGMENT_REGEX => qr'(#([-?/:@.~!$&\'()*+,;=\w]|%[0-9a-fA-F]{2})*)*';
+
 our @EXPORT_OK = qw(
   WEBSOCKET_API_VERSION DEFAULT_WORKER_TIMEOUT
   WORKER_COMMAND_ABORT WORKER_COMMAND_QUIT WORKER_COMMAND_CANCEL WORKER_COMMAND_OBSOLETE WORKER_COMMAND_LIVELOG_STOP
@@ -120,6 +123,7 @@ our @EXPORT_OK = qw(
   DEFAULT_MAX_SETUP_TIME
   DB_TIMESTAMP_ACCURACY
   VIDEO_FILE_NAME_START VIDEO_FILE_NAME_REGEX
+  FRAGMENT_REGEX
 );
 
 1;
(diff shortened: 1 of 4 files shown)


lib/OpenQA/Constants.pm
```diff
@@ -107,6 +108,8 @@ use constant DB_TIMESTAMP_ACCURACY => 1;
 use constant VIDEO_FILE_NAME_START => 'video.';
 use constant VIDEO_FILE_NAME_REGEX => qr/^.*\/video\.[^\/]*$/;
 
+use constant FRAGMENT_REGEX => qr'(#([-?/:@._~!$&\'()*+,;=\w]|%[0-9a-fA-F]{2})*)*';
```
okurz, line 111: please use https://perldoc.perl.org/perlre#Character-Classes-and-other-Special-Escapes

okurz (changes requested):

lib/OpenQA/Constants.pm
BillAnastasiadis: Thanks, I fixed it a bit to remove `_` since it is contained in the `\w` sequence. For the last part, I don't want to match `_` so I refrained from using `\w`, explicitly stating letters and numbers.

I wonder: based on the definition of fragment identifiers, `#` is not in the allowed characters for a fragment (apart from the obvious, at the start). I have read that some browsers are allowing it, though. In the current implementation, `#` is allowed in the fragment identifier body. Should I leave it like this?

okurz: leave it

t/16-markdown.t
```diff
@@ -66,6 +66,16 @@ subtest 'bugrefs' => sub {
 subtest 'openQA additions' => sub {
     is markdown_to_html('https://example.com'),
       qq{<p><a href="https://example.com">https://example.com</a></p>\n}, 'URL turned into a link';
+    is markdown_to_html('https://example.com/#fragment_-'),
+      qq{<p><a href="https://example.com/#fragment_-">https://example.com/#fragment_-</a></p>\n},
+      'URL with fragment turned into a link';
+    is markdown_to_html('https://example.com/#fragment/<script>test</script>'),
+qq{<p><a href="https://example.com/#fragment/">https://example.com/#fragment/</a><!-- raw HTML omitted -->test<!-- raw HTML omitted --></p>\n},
+      'URL with fragment + script turned into a link';
+    is markdown_to_html('https://example.com/#?(.-/\''),
```
okurz, line 75: I suggest to extend this, e.g. write here

```suggestion
    is markdown_to_html('https://example.com/#?(.-/\' and some text'),
```

and ensure that ` and some text` will not become part of the link

okurz (changes requested):
please see inline comments.

Also, please wrap lines in the git commit message

t/16-markdown.t
BillAnastasiadis: Thanks for the suggestion, implemented

okurz (approved):

okurz (changes requested):
Wait. You removed content from the git commit message. I only asked that you would wrap lines

BillAnastasiadis:
> Wait. You removed content from the git commit message. I only asked that you would wrap lines

I did, but not because of your request. I looked at some other merge requests and saw that being less vocal on the message and instead linking to the related tickets is preferred, so I switched to that format. All the previous info can be found in the ticket that is now provided.

okurz:
> I did, but not because of your request. I looked at some other merge requests and saw that being less vocal on the message and instead linking to the related tickets is preferred, so I switched to that format. All the previous info can be found in the ticket that is now provided.

I would not put it like this. @schlad will agree that the git commit message should be as self-sufficient as possible. I only opt to add a ticket for *additional* context but not as replacement for the text. In some cases the git commit subject lines is self-sufficient and then the body might only contain a ticket reference. However this does not mean that the ticket would *replace* any other content.

okurz (approved):

kalikiana (approved):

PR #4096, opened 2021-07-30
Remove `collect_coverage_of_gru_jobs` from `t/api/02-iso-download.t`

Martchus:
It should not be required, see https://progress.opensuse.org/issues/93883

### DIFF
--- a/t/api/02-iso-download.t
+++ b/t/api/02-iso-download.t
@@ -1,5 +1,5 @@
 #!/usr/bin/env perl
-# Copyright (C) 2014-2020 SUSE LLC
+# Copyright (C) 2014-2021 SUSE LLC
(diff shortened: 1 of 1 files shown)


Martchus:
Apparently we lose coverage for two lines (`log_error('Unable to make final image uploads. Maybe the web UI considers this job already dead.');` and `OpenQA::Scheduler::Model::Jobs->singleton->incomplete_and_duplicate_stale_jobs;` and also strangely gain coverage for two lines (`finish => sub ($tx, $code, $reason = undef) {` and `$reason //= 'no reason';` which are actually marked as uncoverable anyways).

I wouldn't have expected that the test has anything to do with the affected code, though. To me it looks like an error.

okurz (approved):

PR #4100, opened 2021-08-02
Dependency cron 2021-08-02

openqabot:


perlpunk:
I triggered this manually because the new DBD::SQLite seems to be there: http://download.opensuse.org/repositories/devel:/openQA:/Leap:/15.2/openSUSE_Leap_15.2/x86_64/
But we still get the same warnings.

perlpunk:
The cached packages show: perl-DBD-SQLite-1.68-lp152.2.1.x86_64.rpm
But we need: perl-DBD-Pg-3.14.2-lp152.2.6.x86_64.rpm
Not sure why it is not picking it up

perlpunk:
Replaced by https://github.com/os-autoinst/openQA/pull/4101

PR #4103, opened 2021-08-03
Simplify loop over results in dump_templates

okurz:


script/dump_templates
```diff
@@ -212,44 +220,35 @@ for my $table (qw(Machines TestSuites Products JobTemplates)) {
...
-        delete $r->{machine}->{id}          if $r->{machine};
-        delete $r->{test_suite}->{id}       if $r->{test_suite};
-        push @r, $r;
-    }
-    $result{$table} = [@r];
+sub handle_table ($table) {
+    $result{$table} = [grep { defined } map { handle_row($table, $_) } @{$result{$table}}];
 }
 
+handle_table($_) for keys %result;
```
okurz, line 250: this looks ok but maybe someone has an even better approach without needing more functions

Martchus (approved):

PR #4107, opened 2021-08-04
fullstack: Return early if stat fails to avoid file errors

kalikiana:
- Mojo::File::path->slurp croaks on error
- This way the test plan won't be broken

### DIFF
--- a/t/full-stack.t
+++ b/t/full-stack.t
@@ -313,7 +313,7 @@ subtest 'Cache tests' => sub {
     ok wait_for_result_panel($driver, qr/Result: passed/), 'test 5 is passed' or show_job_info 5;
     stop_worker;
     $autoinst_log = autoinst_log(5);
-    ok -s $autoinst_log, 'Test 5 autoinst-log.txt file created';
+    ok -s $autoinst_log, 'Test 5 autoinst-log.txt file created' or return;
     my $log_content = $autoinst_log->slurp;
     like $log_content, qr/Downloading Core-7.2.iso/, 'Test 5, downloaded the right iso';
     like $log_content, qr/11116544/,                 'Test 5 Core-7.2.iso size is correct';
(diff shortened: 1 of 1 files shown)


t/full-stack.t
```diff
@@ -191,9 +191,8 @@ $driver->get($job_page_url);
 ok wait_for_result_panel($driver, qr/Result: passed/), 'test 1 is passed' or show_job_info 1;
 my $autoinst_log = autoinst_log(1);
 ok -s $autoinst_log, 'autoinst log file generated';
-my $worker_log = $autoinst_log->dirname->child('worker-log.txt');
-ok -s $worker_log, 'worker log file generated';
-my $log_content = $worker_log->slurp;
+my $worker_log  = $autoinst_log->dirname->child('worker-log.txt');
+my $log_content = eval { $worker_log->slurp };
```
perlpunk, line 195: This way `$log_content` would end up as undef in case of an error, so we would get a unitialized warning in the next test

perlpunk:
I think it could be confusing to remove the `-s` test.
In the future, if that file isn't generated, we will get a lot of test failures for the content checking, but we wouldn't know if the file wasn't generated at all.
If the file wasn't generated at all, the following tests are pointless.
So I don't really understand why that test is removed.

kalikiana:
> I think it could be confusing to remove the `-s` test.
> In the future, if that file isn't generated, we will get a lot of test failures for the content checking, but we wouldn't know if the file wasn't generated at all.
> If the file wasn't generated at all, the following tests are pointless.
> So I don't really understand why that test is removed.

The PR description claims that it's redundant. But you make a good point.

perlpunk:
What I find weird is, how can `-s` succeed, but we still get
```
Can't open file "/tmp/392z3XoTPN/full-stack.d/openqa/testresults/00000/00000005-tinycore-1-flavor-i386-Build1-core@coolone/autoinst-log.txt": No such file or directory at t/full-stack.t line 317.
```
in the next line?

kalikiana:
> What I find weird is, how can `-s` succeed, but we still get
> 
> ```
> Can't open file "/tmp/392z3XoTPN/full-stack.d/openqa/testresults/00000/00000005-tinycore-1-flavor-i386-Build1-core@coolone/autoinst-log.txt": No such file or directory at t/full-stack.t line 317.
> ```
> 
> in the next line?

I would guess it's unreadable or gets deleted between the calls (which is also why usually I would not use a separate check in code like this).

perlpunk:
The message says "No such file or directory", so apparently it got deleted in between.
Either CircleCI is very broken and files just vanish (hopefully not), or maybe something in os-autoinst deletes that file, I guess.

But I don't really see how suppressing the slurp error could help investigating this...

Martchus (commented):
The test plan is broken just because an exception happens during the test run? That shouldn't be the case.

okurz (approved):
I guess I would have bailed but returning is ok as well because we don't want to accept any failing test anyway so the lost test coverage will be caught

PR #4108, opened 2021-08-05
Add yaml file to enable commit message checks

BillAnastasiadis:
see https://progress.opensuse.org/issues/95581

The rules that are checked against are:

- commit subject must begin either with a capital letter (A-Z) or with a tag ([A-Za-z0-9_]:).
- commit subject must be no longer than 72 characters
- commit subject and commit body must be separated by at least one blank line
- each line in commit body must be no longer than 80 characters

### DIFF
--- a/.github/workflows/commit_message_checker.yml
+++ b/.github/workflows/commit_message_checker.yml
@@ -1,0 +1,51 @@
+---
+# https://github.com/marketplace/actions/gs-commit-message-checker
+name: 'Commit Message Check'
+# yamllint disable-line rule:truthy
+on: [push, pull_request]
+
+jobs:
+  check-commit-message:
+    name: Check Commit Message
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check Subject Begining
+        uses: gsactions/commit-message-checker@v1
+        with:
+          pattern: '^([A-Z]|\w+:)'
+          flags: 'g'
+          error: 'The subject does not start with a capital or tag.'
+          excludeDescription: 'true'
+          excludeTitle: 'true'
+          checkAllCommitMessages: 'true'
+          accessToken: ${{ secrets.GITHUB_TOKEN }}
+      - name: Check Subject Line Length
+        uses: gsactions/commit-message-checker@v1
+        with:
+          pattern: '^.{1,72}(\n|$)'
+          flags: 'g'
+          error: 'The maximum subject line length of 72 characters is exceeded.'
+          excludeDescription: 'true'    # excludes the description body of a pull request
+          excludeTitle: 'true'    # excludes the title of a pull request
+          checkAllCommitMessages: 'true'    # checks all commits associated with a pull request
+          accessToken: ${{ secrets.GITHUB_TOKEN }}    # only required if checkAllCommitMessages is true
+      - name: Check Empty Line
+        uses: gsactions/commit-message-checker@v1
+        with:
+          pattern: '^.*(\n\n|$)'
+          flags: 'g'
+          error: 'No newline between title and description.'
+          excludeDescription: 'true'
+          excludeTitle: 'true'
+          checkAllCommitMessages: 'true'
+          accessToken: ${{ secrets.GITHUB_TOKEN }}
+      - name: Check Body Line Length
+        uses: gsactions/commit-message-checker@v1
+        with:
+          pattern: '^((?!.{80,})(.|\n))*$'
+          flags: 'g'
+          error: 'The maximum body line length of 72 characters is exceeded.'
+          excludeDescription: 'true'
+          excludeTitle: 'true'
+          checkAllCommitMessages: 'true'
+          accessToken: ${{ secrets.GITHUB_TOKEN }}


.github/workflows/commit_message_checker.yml
```diff
@@ -0,0 +1,38 @@
...
+        uses: gsactions/commit-message-checker@v1
+        with:
+          pattern: '^.*(\n\n|$)'
+          flags: 'g'
+          error: 'No newline between title and description.'
+          excludeDescription: 'true'
+          excludeTitle: 'true'
+          checkAllCommitMessages: 'true'
+          accessToken: ${{ secrets.GITHUB_TOKEN }}
+      - name: Check	Body Line Length
```
kalikiana, line 29: What's this `	` about?

kalikiana (commented):
I like the separate check/message/regex definitions. What we probably also want is the "foo:" prefix as suggested it in the ticket?

.github/workflows/commit_message_checker.yml
BillAnastasiadis: That was a tab instead of a space, which was completely invisible on vim. Fixed now

BillAnastasiadis:
> I like the separate check/message/regex definitions. What we probably also want is the "foo:" prefix as suggested it in the ticket?

I will check the ticket again as soon as progress is back up. We need to decide whether to enforce some stuff (i.e. the subject could be starting with a `feat|bug|...|...|:` tag from a set of predefined tags OR start with a capital letter. I left it like this to avoid failing commits before the rules are clear to all. Good thing is, due to the separated checks people will know exactly what to add/remove/fix.

.github/workflows/commit_message_checker.yml
```diff
@@ -0,0 +1,38 @@
+name: 'Commit Message Check'
+on: [push, pull_request]
```
perlpunk, line 2: You need
```
# yamllint disable-line rule:truthy
```
before that line.
In YAML 1.1 `on` was considered a boolean, but GutHub Workflows don't use YAML 1.1

BillAnastasiadis: Thanks for the info! Fixed

kalikiana:
> > I like the separate check/message/regex definitions. What we probably also want is the "foo:" prefix as suggested it in the ticket?
> 
> I will check the ticket again as soon as progress is back up. We need to decide whether to enforce some stuff (i.e. the subject could be starting with a `feat|bug|...|...|:` tag from a set of predefined tags OR start with a capital letter. I left it like this to avoid failing commits before the rules are clear to all. Good thing is, due to the separated checks people will know exactly what to add/remove/fix.

The ticket says the following:

```
We could also check that the message starts with an uppercase letter, and optionally with something like t:, ci: etc.
```

I'd be okay to go with what's proposed here and add checks for an optional prefix in a follow-up PR

BillAnastasiadis:
> > > I like the separate check/message/regex definitions. What we probably also want is the "foo:" prefix as suggested it in the ticket?
> > 
> > 
> > I will check the ticket again as soon as progress is back up. We need to decide whether to enforce some stuff (i.e. the subject could be starting with a `feat|bug|...|...|:` tag from a set of predefined tags OR start with a capital letter. I left it like this to avoid failing commits before the rules are clear to all. Good thing is, due to the separated checks people will know exactly what to add/remove/fix.
> 
> The ticket says the following:
> 
> ```
> We could also check that the message starts with an uppercase letter, and optionally with something like t:, ci: etc.
> ```
> 
> I'd be okay to go with what's proposed here and add checks for an optional prefix in a follow-up PR

I'm going for it. I'll push that tomorrow

.github/workflows/commit_message_checker.yml
```diff
@@ -0,0 +1,50 @@
...
+          excludeTitle: 'true'    # excludes the title of a pull request
+          checkAllCommitMessages: 'true'    # checks all commits associated with a pull request
+          accessToken: ${{ secrets.GITHUB_TOKEN }}    # only required if checkAllCommitMessages is true
+      - name: Check Empty Line
+        uses: gsactions/commit-message-checker@v1
+        with:
+          pattern: '^.*(\n\n|$)'
+          flags: 'g'
+          error: 'No newline between title and description.'
+          excludeDescription: 'true'
```
perlpunk, line 37: Have you tested this one?
I'm wondering how `excludeDescription` actually works. What is the regular expression applied to here if there are multiple lines in the commit message?

perlpunk: Oh, or is this really about the github pull request comment and not related to the git commit message?

Martchus (commented):
Hard to tell whether this will do what we expect but I suppose we can always amend it afterwards. It would be nice to add a link to the documentation of the YAML structure somewhere as a comment.

.github/workflows/commit_message_checker.yml
BillAnastasiadis: Yes, this is about the github description. I thought that this should be left open.

I have tested this one, it's basically saying "after the first line, find either the end of the commit msg or two newlines". It will succeed with one-liners, it will succeed with multiple lines where the first line and the rest of the text are separated by an extra newline, but it will fail if the second line doesn't have an extra line between itself and the first line

BillAnastasiadis:
> Hard to tell whether this will do what we expect but I suppose we can always amend it afterwards. It would be nice to add a link to the documentation of the YAML structure somewhere as a comment.

Correct. I have tested all basic cases and everything seems to work fine (no false positives or true negatives so far). Of course, everyone's encouraged to test themselves. If there's any problem, fixing this should be as trivial as changing a regex. 

I based the yaml file on the example given in the action's page here: https://github.com/marketplace/actions/gs-commit-message-checker

The example itself has faulty regexes, but the structure works as expected

.github/workflows/commit_message_checker.yml
BillAnastasiadis: Note that it should also fail if the empty line contains spaces, which can easily be accounted for, but the extra line shouldn't contain any extra spaces anyway so I left it at that

perlpunk (commented):
I'll approve when you put the URL to the docs into the yaml file as @Martchus suggested

kalikiana (approved):

.github/workflows/commit_message_checker.yml
```diff
@@ -0,0 +1,50 @@
+---
```
perlpunk, line 1:
```suggestion
# https://github.com/marketplace/actions/gs-commit-message-checker
---
```

BillAnastasiadis: Thanks, done!

perlpunk (approved):

kalikiana:
FYI: Was affected by [poo#96636](https://progress.opensuse.org/issues/96636)

PR #4109, opened 2021-08-06
Fix "not last X" needle searches to include "never" results

AdamWill:
If you set the "Limit Use:" or "Limit Match:" filter in the
Needles admin UI page to one of the "not last X" options, the
results don't include needles for which the "last use" or "last
match" value is "never" (shown as 'never' in the UI, represented
in the DB as null and in the abstraction layer as undef). This
is a bigger problem than it may seem at first glance because the
value can be 'never' for needles which *have* been seen/matched,
but only a long time ago and all those tests have now been
garbage-collected.

This fixes that, and also enhances the test to check these cases
(using both the 'custom' option and a pre-rolled option, as
those actually hit two different code paths). It also corrects
the existing tests a bit.

### DIFF
--- a/lib/OpenQA/WebAPI/Controller/Admin/Needle.pm
+++ b/lib/OpenQA/WebAPI/Controller/Admin/Needle.pm
@@ -47,13 +47,13 @@ sub _translate_cond($) {
         return {'>=' => _translate_days($1)};
     }
     elsif ($cond =~ m/^max(\d+)$/) {
-        return {'<' => _translate_days($1)};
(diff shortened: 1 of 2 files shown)


AdamWill:
note, I am using CI to run the tests on this, as I can't run the UI tests locally. I probably got something wrong. The actual code change works, I tested it manually, but I'll probably have to tweak the tests a bit somewhere.

Martchus (approved):
Makes sense and the code looks correct.

Amrysliu (approved):
LGTM

PR #4111, opened 2021-08-06
fullstack: Extended wait for result panel

kalikiana:
fullstack test: Extended wait for result panel

- The job is still running and the check times out
-  Bail out whenever the expected result was not found


Martchus (commented):
Maybe the overall time limit needed to be adjusted as well. Not sure whether there's a huge benefit from the bail-out but I guess it cannot hurt.

Martchus (approved):

perlpunk:
huh, that auto-merge didn't wait for circleci build-docs, fullstack and t

okurz:
yes, github auto-merge does not wait for most checks, only mergify does. We would need to change the settings in github project settings

PR #4112, opened 2021-08-09
Log user data received during OAuth2

ssssam:
When the error "User data returned by OAuth2 provider is insufficient"
appears, it's important for admin to see what data the identity provider
actually returned.

This info will only be logged when log level is set to `debug`.

### DIFF
--- a/lib/OpenQA/WebAPI/Auth/OAuth2.pm
+++ b/lib/OpenQA/WebAPI/Auth/OAuth2.pm
@@ -16,6 +16,8 @@
 package OpenQA::WebAPI::Auth::OAuth2;
 use Mojo::Base -base, -signatures;
 use Carp 'croak';
+use Data::Dumper;
+use OpenQA::Log qw(log_debug);
 
 sub auth_setup ($server) {
     my $app    = $server->app;
@@ -79,6 +81,7 @@ sub update_user ($controller, $main_config, $provider_config, $data) {
     }
     my $details = $tx->res->json;
     if (ref $details ne 'HASH' || !$details->{id} || !$details->{$provider_config->{nickname_from}}) {
+        log_debug("OAuth2 user provider returned: " . Dumper($details));
         return $controller->render(text => 'User data returned by OAuth2 provider is insufficient', status => 403);
     }
     my $provider_name = $main_config->{provider};


ssssam:
To give some background to this merge request: we are setting up openqa.gnome.org, and using GNOME Keycloak instance for login which has proved remarkably difficult to configure.

Martchus (commented):
I'm wondering whether it would be bad to have information like this stored in the logs. Note that one might configure the logging verbosity to debug for various reasons even on production systems. One might still not assume that changing the logging verbosity means sensitive data can end up within the logs. Maybe it would be better to hide this behind an explicit config switch (e.g. within the config file or just using an environment variable).

lib/OpenQA/WebAPI/Auth/OAuth2.pm
```diff
@@ -78,6 +80,7 @@ sub update_user ($controller, $main_config, $provider_config, $data) {
         return $controller->render(text => $msg, status => 403);    # return always 403 for consistency
     }
     my $details = $tx->res->json;
+    log_debug("OAuth2 user provider returned: " . Dumper($details));
```
perlpunk, line 83: If this information is only necessary in case of the error message "User data returned by OAuth2 provider is insufficient", then I guess this line should be moved to inside the next `if`

ssssam: Are there error paths later that could be triggered by bad user info? If no, then yes let's move it inside the `if`. That might also help with @Martchus comment (no errors = nothing logged)

ssssam: To answer my own question, the only codepath using $details is in the same function, and has no special error handling so should already log a detailed message. I will make this change.

ssssam:
> I'm wondering whether it would be bad to have information like this stored in the logs. Note that one might configure the logging verbosity to debug for various reasons even on production systems. One might still not assume that changing the logging verbosity means sensitive data can end up within the logs. Maybe it would be better to hide this behind an explicit config switch (e.g. within the config file or just using an environment variable).

I can't give a full security review, but a few thoughts. First, if you have access to the logs, you probably have read access to the database too. Second, the OAuth2 user endpoint will only return data according to the scopes it is given... in our case, this is what gets logged:

```
Details: {
  'sub' => 'f:469edfd7-1d35-4c9d-b69f-b4fd62c417d9:sthursfield',
  'name' => 'Sam Thursfield Thursfield',
  'family_name' => 'Thursfield',
  'email' => 'sam@afuera.me.uk',
  'given_name' => 'Sam Thursfield',
  'preferred_username' => 'sthursfield',
  'email_verified' => bless( do{\(my $o = 0)}, 'JSON::PP::Boolean' )
}
```

I am happy to share the full log because there's nothing secret, just my username and email address which are both already public, and an internal identifier that only makes sense within the identity provider.

Martchus (approved):
I guess I'm convinced :-)

perlpunk (approved):

PR #4113, opened 2021-08-09
ci: Improve hadolint check

perlpunk:
Issues:
* https://progress.opensuse.org/issues/96636
* https://github.com/hadolint/hadolint/issues/475

It seems repeated calls to docker can lead to rare unexpected exits with code
127.
While this commit does not fix the cause, it avoids calling docker repeatedly,
which also makes the test faster.

I tested that the check works by introducing an error in one of the files:
```
container/webui/Dockerfile-lb:2:1 unexpected 'F' expecting '#', '\', ADD, ARG, CMD, COPY, ENTRYPOINT, ENV, EXPOSE, FROM, HEALTHCHECK, LABEL, MAINTAINER, ONBUILD, RUN, SHELL, STOPSIGNAL, USER, VOLUME, WORKDIR, a pragma, at least one space, or end of input
```

### DIFF
--- a/tools/static_check_containers
+++ b/tools/static_check_containers
@@ -9,7 +9,8 @@ else
     img=ghcr.io/hadolint/hadolint
 fi
 cre="${cre:-"docker"}"
-for i in container/*/Dockerfile*; do
-    echo "# Static check of $i"
-    $cre run --rm -i $img < "$i"
-done
+dockerfiles=(container/*/Dockerfile*)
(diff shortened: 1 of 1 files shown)


Martchus (approved):

kalikiana (approved):
I like that we can save the overhead of the container runtime as a side effect. Thank you for tracking down that hadolint bug!

PR #4114, opened 2021-08-09
Allow users to change which group $dbuser belongs to

dcermak:
The `$dbuser` only belongs to `www` on openSUSE, on other distributions this group can be different. With this commit, the group becomes configurable and allows to execute this script on systems where there is no `www` group.

### DIFF
--- a/script/fetchneedles
+++ b/script/fetchneedles
@@ -2,6 +2,7 @@
 [ "$1" = "-h" ] || [ "$1" = "--help" ] && echo "Get openSUSE tests and needles from https://github.com/os-autoinst/os-autoinst-distri-opensuse" && exit
 
 : "${dbuser:="geekotest"}"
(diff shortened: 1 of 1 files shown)


script/fetchneedles
```diff
@@ -21,7 +22,7 @@ dir="/var/lib/openqa/share/tests"
 if [ -w / ]; then
     if [ ! -e "$dir/$dist" ]; then
         mkdir -p "$dir/$dist"
-        chown $dbuser:www "$dir/$dist"
+        chown $dbuser:$dbgroup "$dir/$dist"
```
Martchus, line 25: Since we're already changing this like it would make sense to add quotes btw:
```suggestion
        chown "$dbuser:$dbgroup" "$dir/$dist"
```

perlpunk (approved):

Martchus (approved):

PR #4119, opened 2021-08-11
Improve worker's exception handling and related test

Martchus:
* Turn `try` into `eval` so the contained `return` actually returns from
  the enclosing `sub`
    * No longer log `Another error occurred…` when that's not actually the
      case
    * No longer stop the worker immediately after the first exception
* Log error message from the `eval` block
* Improve tests
    * Check whether the event loop is actually stopped in a controlled way
    * Check whether all error messages are logged
    * Check whether relevant functions are called exactly as many times as
      expected
* See https://progress.opensuse.org/issues/96710

### DIFF
--- a/lib/OpenQA/Worker.pm
+++ b/lib/OpenQA/Worker.pm
@@ -293,8 +293,9 @@ sub init {
             $return_code = 1;
 
             # try to stop gracefully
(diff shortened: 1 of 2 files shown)


asdil12 (approved):
LGTM

PR #4120, opened 2021-08-12
Print stacktrace when worker crashes due to unhandled exception

Martchus:
* Help debugging https://progress.opensuse.org/issues/96710
* Move `Carp::Always` which we're already using from main dependencies to
  common dependencies
* Show the line causing the exception even when it is invoked by the event
  loop
* Keep the normal event loop's error handling as-is
* Example:
  ```
  [info] [pid:24257] Establishing ws connection via ws://localhost:9526/api/v1/ws/20
  Mojo::Reactor::Poll: I/O watcher failed: foobar at /hdd/openqa-devel/repos/openQA/script/../lib/OpenQA/Worker/WebUIConnection.pm line 191.
          …
          OpenQA::Worker::exec(OpenQA::Worker=HASH(0x55809a90e030)) called at script/worker line 140
  [error] [pid:24257] Stopping because a critical error occurred.
  ```


Martchus:
I've already shortened the commit message and I think the body is short enough. So I would ignore the failing commit message check.

perlpunk:
Yeah, we should probably allow longer lines in the body, to be able to paste code

kalikiana:
> Yeah, we should probably allow longer lines in the body, to be able to paste code

Ack. I prepared #4121

Martchus:
It still exceeds the limit but I still think the commit message isn't too long.

kalikiana:
> It still exceeds the limit but I still think the commit message isn't too long.

That was kinda the point of my PR but apparently I can't count or chose the wrong line to check. Maybe we should just drop it... see #4125

kalikiana (approved):

Martchus:
This also leads to back traces in non-fatal cases¹ so I'm not sure whether it makes sense. Having a backtrace of overwhelmingly helpful so I still tend to having it.

---

¹ e.g.:

```
[warn] [pid:9499] Disabling cgroup usage because cgroup creation failed: mkdir /sys/fs/cgroup/systemd: Permission denied at /usr/lib/perl5/vendor_perl/5.32.1/File/Path.pm line 239.
        File::Path::_mkpath(HASH(0x5629e4009750), ARRAY(0x5629e47163a8)) called at /usr/lib/perl5/vendor_perl/5.32.1/File/Path.pm line 173
        File::Path::mkpath(Mojo::File=SCALAR(0x5629e46f2098)) called at /usr/lib/perl5/vendor_perl/5.32.1/Mojo/File.pm line 84
        Mojo::File::make_path(Mojo::File=REF(0x5629e47259d8)) called at /usr/lib/perl5/vendor_perl/5.32.1/Mojo/IOLoop/ReadWriteProcess/CGroup.pm line 40
…
        OpenQA::Worker::exec(OpenQA::Worker=HASH(0x5629e465b3c0)) called at script/worker line 138
```

kalikiana:
> This also leads to back traces in non-fatal cases¹ so I'm not sure whether it makes sense. Having a backtrace of overwhelmingly helpful so I still tend to having it.

It doesn't seem like [Carp::Always](https://metacpan.org/pod/Carp::Always) gives us a choice here. Although the example you gave is something I would rather treat like an error, and which has caused more than one confusing inestigation in the past.

okurz (approved):
hm, I agree that such stack traces can be helpful but if we include "Carp::Always" in the worker code I wonder if it would be consistent to just include it *everywhere*. I don't think it's meant to be used like this. We can enable such functionality when calling the according applications. For example we could change
https://github.com/os-autoinst/openQA/blob/master/systemd/openqa-worker%40.service#L15 to say

```diff
- ExecStart=/usr/share/openqa/script/worker --instance %i
+ ExecStart=/usr/bin/perl -MCarp::Always /usr/share/openqa/script/worker --instance %i
```

does not look that nice but is easier to tweak. Anyway, approving. Maybe we are ok this way.

Martchus:
Just specifying the module on the command-line would have been a good alternative, indeed. I guess we can leave it within the file for now but if it is generally too noisy we can revert the change.

PR #4122, opened 2021-08-13
Let workers declare themselves as broken if asset downloads are piling

asdil12:
Ticket: https://progress.opensuse.org/issues/96623

### DIFF
--- a/lib/OpenQA/CacheService/Response/Info.pm
+++ b/lib/OpenQA/CacheService/Response/Info.pm
@@ -1,4 +1,4 @@
-# Copyright (C) 2019 SUSE LLC
+# Copyright (C) 2019-2021 SUSE LLC
 #
 # This program is free software; you can redistribute it and/or modify
 # it under the terms of the GNU General Public License as published by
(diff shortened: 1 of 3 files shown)


lib/OpenQA/CacheService/Response/Info.pm
```diff
@@ -25,10 +25,18 @@ sub available_workers {
...
+    return undef if $self->max_inactive_jobs < 0;
+    return undef unless my $data = $self->data;
+    return $data->{inactive_jobs} > $self->max_inactive_jobs;
+}
+
 sub availability_error {
     my $self = shift;
     return $self->error if $self->has_error;
     return 'No workers active in the cache service' unless $self->available_workers;
+    return 'Already cache service jobs piling up in queue' if $self->inactive_jobs_exceeded;
```
kalikiana, line 39: How about phrasing it more objectively and mentioning the limit? To avoid the "did the config take effect" question

```suggestion
    return 'Cache service queue is already full ($max_inactive_jobs)' if $self->inactive_jobs_exceeded;
```

lib/OpenQA/CacheService/Response/Info.pm
```diff
@@ -25,10 +25,18 @@ sub available_workers {
     return $data->{active_workers} != 0 || $data->{inactive_workers} != 0;
 }
 
+sub inactive_jobs_exceeded {
+    my $self = shift;
```
Martchus, line 29: It would be nice to use signatures in new code, e.g.
```suggestion
sub inactive_jobs_exceeded ($self) {
```

asdil12: sure - if that works

kalikiana (approved):

PR #4123, opened 2021-08-13
worker: Workaround errors in web socket handling

Martchus:
* Use a normal sleep when waiting for the worker cache instead of running
  the event loop
* Handling `MAX_SETUP_TIME` will no longer work
* Handling job cancellation during the setup will no longer work
* See https://progress.opensuse.org/issues/96710#note-15


perlpunk (approved):
I'll just trust you :)

Martchus:
It is basically already running in production on openqaworker5 and 6. It looks quite good (and hence openqaworker3 and 8 have taken over when it comes to producing many incompletes).

kalikiana (approved):

PR #4124, opened 2021-08-13
worker: Use proper callback instead of `delay` function during setup

Martchus:
* Possibly fix crashes in web socket processing during setup
* Use proper callback instead of `delay` function when waiting for the
  cache service
* See https://progress.opensuse.org/issues/96710#note-15

---

The `delay` function can later be removed from the worker. This is not
included to avoid conflicts with #4123 (which we might want to merge
fast). Removing the delay function would [save use 44 lines of code](https://github.com/os-autoinst/openQA/commit/9b0a11ca499f48d6f760077a59ac260227e1cc82)
so in total the additional number of lines due to this change is < 30 and
it is less hacky.

Before merging this PR I need to do some more testing, possibly even
applying it on one production worker (e.g. openqaworker5).

### DIFF
--- a/lib/OpenQA/Worker.pm
+++ b/lib/OpenQA/Worker.pm
@@ -383,29 +383,9 @@ sub configure_cache_client {
     $client->ua->inactivity_timeout($ENV{OPENQA_WORKER_CACHE_SERVICE_CHECK_INACTIVITY_TIMEOUT} // 10);
 }
 
-# sleeps for the specified number of seconds
-# note: It was supposed to "sleep" for the specified number of seconds while actually running the worker's event
-#       loop started via exec() to keep processing events (like job cancellation). However, this caused problems
-#       (see poo#96710). Hence we're going back to using a normal sleep here.
-sub delay ($self, $delay) { sleep $delay }
-
-sub stop_event_loop ($self) {
-    Mojo::IOLoop->stop;
-    $self->{_resume_loop} = undef;
-}
-
-sub exec {
-    my ($self) = @_;
-
+sub exec ($self) {
     my $return_code = $self->init;
-
-    # start event loop - this will block until stop is called
-    my $loop = Mojo::IOLoop->singleton;
-    do {
-        $self->{_resume_loop} = undef;
-        $loop->start;
-    } while ($self->{_resume_loop});
-
+    Mojo::IOLoop->singleton->start;
     return $return_code;
 }
 
@@ -601,7 +581,7 @@ sub stop {
 
     # stop immediately if there is currently no job
     my $current_job = $self->current_job;
-    return $self->_inform_webuis_before_stopping(sub { $self->stop_event_loop }) unless defined $current_job;
+    return $self->_inform_webuis_before_stopping(sub { Mojo::IOLoop->stop }) unless defined $current_job;
     return undef if $self->{_finishing_off};
 
     # stop job directly during setup because the IO loop is blocked by isotovideo.pm during setup
@@ -621,7 +601,7 @@ sub kill {
     my ($self) = @_;
 
     if (my $current_job = $self->current_job) { $current_job->kill; }
-    $self->stop_event_loop;
+    Mojo::IOLoop->stop;
 }
 
 sub is_stopping {
@@ -707,7 +687,7 @@ sub _handle_client_status_changed {
         }
         if (!defined $self->current_job) {
             log_error('Stopping because registration with all configured web UI hosts failed');
-            return $self->stop_event_loop;
+            return Mojo::IOLoop->stop;
         }
 
         # continue executing the current job even though the registration is not possible anymore; it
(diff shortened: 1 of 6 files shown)


Martchus:
I've been testing it locally, of course with caching and also test pool server enabled. However, I'd like to test it on a production worker on Monday first to see whether this won't introduce the problems we see here (https://progress.opensuse.org/issues/96710) again.

Martchus:
So far it looks good on openqaworker5. I just had to fix the status updates while waiting for the cache service.

Martchus:
I've been extending tests for the so far uncovered lines changed by this PR.

I've been running the code on openqaworker5 for hours and everything looks good. So running the event loop while waiting for the cache service isn't generally a problem; just stopping/starting the event loop in the way the previous delay function did was.

I've also added the commit to remove the `delay` function (and its test) so there shouldn't be any dead code left.

okurz:
can you comment on the coverage changes? Looking at the files with coverage less than 100% there is

lib/OpenQA/Worker.pm
lib/OpenQA/Worker/Engines/isotovideo.pm
lib/OpenQA/Worker/Job.pm
t/24-worker-engine.t

At least for "t/24-worker-engine.t" I thought we already had 100%. Can you please check that?

Martchus:
I can have a look. So far I've only checked that the changed lines are completely covered, not all files being touched.

Martchus:
We haven't had 100 % for any of these files. (PR to improve coverage of the test: https://github.com/os-autoinst/openQA/pull/4128)

lib/OpenQA/Worker.pm
```diff
@@ -383,29 +383,9 @@ sub configure_cache_client {
     $client->ua->inactivity_timeout($ENV{OPENQA_WORKER_CACHE_SERVICE_CHECK_INACTIVITY_TIMEOUT} // 10);
 }
 
-# sleeps for the specified number of seconds
-# note: It was supposed to "sleep" for the specified number of seconds while actually running the worker's event
-#       loop started via exec() to keep processing events (like job cancellation). However, this caused problems
-#       (see poo#96710). Hence we're going back to using a normal sleep here.
-sub delay ($self, $delay) { sleep $delay }
-
-sub stop_event_loop ($self) {
```
kalikiana, line 392: I'm wondering if it would have been useful to keep this helper so it can be "counted" in unit tests? To check that we stop correctly 🤔

lib/OpenQA/Worker/Engines/isotovideo.pm
```diff
@@ -92,76 +94,86 @@ sub detect_asset_keys ($vars) {
...
+    my $asset
+      = $cache_client->asset_exists($webui_host, $asset_uri)
+      ? $cache_client->asset_path($webui_host, $asset_uri)
+      : undef;
+    if ($this_asset eq 'UEFI_PFLASH_VARS' && !defined $asset) {
+        log_error("Failed to download $asset_uri", channels => 'autoinst');
+        # assume that if we have a full path, that's what we should use
+        $vars->{$this_asset} = $asset_uri if -e $asset_uri;
+        # don't kill the job if the asset is not found
+        # TODO: This seems to leave the job stuck in some cases (observed in production on openqaworker3).
```
kalikiana, line 160: I realize you didn't introduce this comment, but I feel like we should have a ticket instead of the comment. It's clearly not useful in getting somebody to fix it.

lib/OpenQA/Worker.pm
Martchus: If we fail to stop the event loop within a test we would actually have a test which just hangs forever in any case. So it is not very useful. The only purpose of this function was to unset `$self->{_resume_loop}` which is now not need anymore (and hence doesn't need to be tested either).

kalikiana (approved):
Looks pretty good to me, besides my two remarks. Sorry for the slow review, I instinctively had to take a break to be sure I was reviewing it properly.

lib/OpenQA/Worker/Engines/isotovideo.pm
Martchus: Yes, this should have a ticket. However, for the sake of this change which is only supposed to turn synchronous functions into asynchronous ones it is out of scope (and I don't like to change multiple unrelated things in one PR).

I'm also afraid that it isn't easily possible to file a workable ticket at this point about it. So likely we can just remove the TODO completely. Without any clues why this would lead to stuck jobs it isn't very useful.

kalikiana: Dunno if that single line is worth a PR. I just assume it'll stay in there and never get fixed.

okurz: Please don't add TODO, same as disabled code as likely nobody else but you will understand this. Better leave it out and keep in your local git repository, either in `git stash` or a temporary "WIP"-commit, a ticket or just delete

okurz (changes requested):

Martchus:
@okurz This code is just moved from one place to another and removing the TODO would be mixing an unrelated change into the PR. I can create an additional commit within this PR if it needs to be.

PR #4129, opened 2021-08-18
fullstack: scale panel waits via TimeLimit

kalikiana:
Rather than increasing the timeouts, we can scale them. This requires tests to use TimeLimit but we should do that anyway.

### DIFF
--- a/t/lib/OpenQA/Test/FullstackUtils.pm
+++ b/t/lib/OpenQA/Test/FullstackUtils.pm
@@ -32,6 +32,7 @@ use Time::HiRes 'sleep';
 use Time::Seconds;
 use OpenQA::SeleniumTest;
 use OpenQA::Scheduler::Model::Jobs;
+use OpenQA::Test::TimeLimit ();
 use OpenQA::Test::Utils 'wait_for_or_bail_out';
 
 my $JOB_SETUP = {
@@ -100,7 +101,8 @@ sub wait_for_result_panel {
     my $looking_for_result = $result_panel =~ qr/Result: /;
     $check_interval //= 0.5;
 
-    for (my $count = 0; $count < (5 * ONE_MINUTE / $check_interval); $count++) {
+    my $timeout = OpenQA::Test::TimeLimit::scale_timeout(ONE_MINUTE * 3) / $check_interval;
+    for (my $count = 0; $count < $timeout; $count++) {
         wait_for_ajax(msg => "result panel shows '$result_panel'");
         my $status_text = find_status_text($driver);
         return 1 if $status_text =~ $result_panel;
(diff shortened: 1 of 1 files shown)


t/lib/OpenQA/Test/FullstackUtils.pm
```diff
@@ -100,7 +100,8 @@ sub wait_for_result_panel {
     my $looking_for_result = $result_panel =~ qr/Result: /;
     $check_interval //= 0.5;
 
-    for (my $count = 0; $count < (5 * ONE_MINUTE / $check_interval); $count++) {
+    my $timeout = OpenQA::Test::TimeLimit::scale_timeout(ONE_MINUTE * 3) / $check_interval;
```
perlpunk, line 103: hm, this relies on the fact that the calling test does a `use OpenQA::Test::TimeLimit 123;`
I try to avoid using modules/functions when the module hasn't been loaded in the same file.
Of course loading `OpenQA::Test::TimeLimit` to setup the `$SCALE_FACTOR` would require calling import, which sets an `alarm`, which we don't want in the FullstackUtils module.

okurz: so what to do?

perlpunk: 1. An easy workaround would be to do `use OpenQA::Test::TimeLimit ()` here.
Then in the worst case `$SCALE_FACTOR` would still be 1 if `import` was never called.

2. For a cleaner solution I would move the scaling code (`$SCALE_FACTOR *= ...`) into its own function or even simply to the top of the module, as we always want to do that when the module is loaded, and we only want to do it once.
3. 
We could go with the first suggestion and followup with the second.

Martchus: About 1.: Then one needs to take care not `use OpenQA::Test::FullstackUtils;` before loading e.g. `use OpenQA::Test::TimeLimit '42';`. If one fails to do so the time limit wouldn't be actually effective as `OpenQA::Test::TimeLimit ()` was already loaded before, right? Maybe it is even better to have this line failing explicitly instead of risking a timeout being silently non-effective.

perlpunk: If you call
```
use OpenQA::Test::TimeLimit ();
use OpenQA::Test::TimeLimit 42;
```
then the module will only be compiled once, but the `import` will still be called for the second line.

Martchus: Ah, I didn't know that. Then also 1. makes totally sense :-)

perlpunk (approved):

okurz (approved):

PR #4130, opened 2021-08-18
Document that needles will not be loaded from a custom CASEDIR

Martchus:
This should clear up the confusion about how we currently load needles if
`CASEDIR` is customized which came up in the ticket
https://progress.opensuse.org/issues/94735.

### DIFF
--- a/docs/WritingTests.asciidoc
+++ b/docs/WritingTests.asciidoc
@@ -1506,7 +1506,6 @@ openqa-clone-custom-git-refspec https://github.com/os-autoinst/os-autoinst-distr
 ----
 
 Keep in mind that if `PRODUCTDIR` is overwritten it might not relate to the
-state of the specified git refspec. For example the asset caching
-functionality will override the `PRODUCTDIR` variable. The above method can
-still be used in this case in the common case that the test schedule does not
-need to be changed or in case a custom schedule is defined by `SCHEDULE`.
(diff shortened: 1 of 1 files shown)


okurz (approved):

docs/WritingTests.asciidoc
```diff
@@ -1506,7 +1507,10 @@ openqa-clone-custom-git-refspec https://github.com/os-autoinst/os-autoinst-distr
 ----
 
 Keep in mind that if `PRODUCTDIR` is overwritten it might not relate to the
-state of the specified git refspec. For example the asset caching
-functionality will override the `PRODUCTDIR` variable. The above method can
-still be used in this case in the common case that the test schedule does not
-need to be changed or in case a custom schedule is defined by `SCHEDULE`.
+state of the specified git refspec. The above method can still be used in this
+case in the common case that the test schedule does not need to be changed or in
+case a custom schedule is defined by `SCHEDULE`.
```
dcermak, line 1512: It looks like either `in this case` or `in the common case` shouldn't be there.

Martchus: I haven't changed this part of the phrasing because I wasn't sure. I suppose "in this case" refers to `PRODUCTDIR` being overridden and "the common case" refers to "the test schedule does not need …".

PR #4131, opened 2021-08-18
Allow configuring the worker cache to keep a certain percentage of free space

Martchus:
This is just a quick draft for the idea mentioned in https://progress.opensuse.org/issues/96554#note-15:

* Add setting `CACHE_MIN_FREE_PERCENTAGE` to configure the minimum free
  disk space the worker cache should leave
* Change the existing setting `CACHELIMIT` so `0` denotes there's no limit
  which is useful to rely only on `CACHE_MIN_FREE_PERCENTAGE`

---

If you like it I can add proper tests.


okurz (commented):
Very nice. I thought about the alternative to reuse the existing configuration parameter but use a negative value or something. But as we use the word "free_percentage" for the asset/result cleanup already I think your choice might be more consistent. But using your way we need to ensure that users do not specify both. What happens if users specify both?

Martchus:
> What happens if users specify both?

Then both limits are effective; the more restrictive one wins. So one actually has to set `CACHELIMIT = 0` if only `CACHE_MIN_FREE_PERCENTAGE` should be used.

okurz:
that approach makes sense. Please include that explanation in https://github.com/os-autoinst/openQA/blob/master/docs/Installing.asciidoc#asset-caching

okurz (approved):

asdil12 (approved):

PR #4133, opened 2021-08-19
Optimize querying progress of running jobs by utilizing module columns

Martchus:


okurz (approved):
that looks too easy :D Would be great if this works.

kraih (approved):
I like less code.

Martchus:
Although the logic for computing the the progress is changed a lot by this change, the percentages for running tests make sense. At least it looks good with production data from OSD. So by simply counting "skipped" and "externally skipped" modules one can indeed figure out the remaining modules and not considering the currently running module doesn't make much of a difference.

Martchus:
By the way, with OSD data (275 running jobs) the loading time for the particular AJAX route goes down from ~ 500 ms to ~ 70 ms to it is around 7 times faster (which of course depends on the number of jobs and modules).

PR #4136, opened 2021-08-20
Prevent assets being cleaned up too early

Martchus:
At this point it is just a draft to see what tests this change breaks.

---

* Associate also non-existent assets with their jobs so they are correctly
  considered pending during the cleanup
* See https://progress.opensuse.org/issues/97304


okurz (approved):

t/17-labels_carry_over.t
```diff
@@ -33,6 +34,7 @@ my $t         = Test::Mojo->new('OpenQA::WebAPI');
 my $rs        = $t->app->schema->resultset('Jobs');
 my $auth      = {'X-CSRF-Token' => $t->ua->get('/tests')->res->dom->at('meta[name=csrf-token]')->attr('content')};
 $test_case->login($t, 'percival');
+assume_all_assets_exist;
```
okurz, line 37: what do these assets have to do with labels carry over? Does the test fail if we don't call this?

t/lib/OpenQA/Test/Utils.pm
```diff
@@ -656,4 +656,6 @@ sub mock_io_loop (%args) {
     return $io_loop_mock;
 }
 
+sub assume_all_assets_exist { OpenQA::Schema->singleton->resultset('Assets')->search({})->update({size => 0}) }
```
okurz, line 659: I am not so sure about the word "assume" as the result of calling this function is not clear. How about "ensure"?

okurz (changes requested):
I find it hard a bit hard to review this PR. Could you maybe split out parts like the "use more Mojo::File" part into separate PRs or commits? I think PR is favourable for easier revertability of the critical changes

t/17-labels_carry_over.t
Martchus: This test restarts jobs and that's prevented by default if assets are missing. So yes, the test fails if we don't call this. I only added this call to tests which would otherwise fail.

t/lib/OpenQA/Test/Utils.pm
Martchus: "ensure" makes it sound like it would possibly create assets if they don't exist. This function merely makes our code assume they exist (by putting *some* size into the database).

Martchus:
Let me see what I can split into other commits.

t/17-labels_carry_over.t
okurz: ok, cool

t/lib/OpenQA/Test/Utils.pm
okurz: Ok, I get that. I would say that effectively assets do exist for the sake of testing just by adding a defined size :) But ok, let's keep "assume"

Martchus:
I was able to split changes of `t/15-assets.t` and a few other changes but most didn't make much sense on their own.

okurz (approved):

Martchus:
I'm really wondering why the fullstack test fails within the CI with no good error message. Locally it passes (also within a clean checkout).

Martchus:
Looks like the introduction of stacktraces fires back. The TAP parser complains that test 66 is missing:

```
t/full-stack.t (Wstat: 0 Tests: 108 Failed: 0)
  Parse errors: Tests out of sequence.  Found (67) but expected (66)
                Tests out of sequence.  Found (68) but expected (67)
                Tests out of sequence.  Found (69) but expected (68)
                Tests out of sequence.  Found (70) but expected (69)
                Tests out of sequence.  Found (71) but expected (70)
Displayed the first 5 of 44 TAP syntax errors.
```

However, it is just meddled into the stacktrace which is printed due to the "usual" cgroup permission error:

```
[warn] [pid:259] Disabling cgroup usage because cgroup creation failed: mkdir /sys/fs/cgroup/systemd/docker: Permission denied at /usr/lib/perl5/vendor_perl/5.26.1/File/Path.pm line 239.
	File::Path::_mkpath(HASH(0x558c28bc0b40), ARRAY(0x558c28bdf9a8)) called at /usr/lib/perl5/vendor_perl/5.26.1/File/Path.pm line 195
	eval {...} called at /home/squamata/project/script/../lib/OpenQA/Worker/Engines/isotovideo.pm line 428
	OpenQA::Worker::Engines::isotovideo::_engine_workit_step_2(OpenQA::Worker::Job=HASH(0x558c28be1820), HASH(0x558c28bc5118), HASH(0x558c28247a90), undef, CODE(0x558c28b251d8)) called at /home/squamata/project/script/../lib/OpenQA/Worker/Engines/isotovideo.pm line 356
	OpenQA::Worker::Engines::isotovideo::engine_workit(OpenQA::Worker::Job=HASH(0x558c28be1820), CODE(0x558c28b251d8)) called at /home/squamata/project/script/../lib/OpenQA/Worker/Job.pm line 275
	OpenQok 66 - Wait for jQuery successful: result panel shows '(?^:State: running)'
A::Worker::Job::start(OpenQA::Worker::Job=HASH(0x558c28be1820)) called at /home/squamata/project/script/../lib/OpenQA/Worker.pm line 726
	OpenQA::Worker::_handle_job_status_changed(OpenQA::Worker=HASH(0x558c28ab3538), OpenQA::Worker::Job=HASH(0x558c28be1820), HASH(0x558c28bdf960)) called at /home/squamata/project/script/../lib/OpenQA/Worker.pm line 421
	OpenQA::Worker::__ANON__(OpenQA::Worker::Job=HASH(0x558c28be1820), HASH(0x558c28bdf960)) called at /usr/lib/perl5/vendor_perl/5.26.1/Mojo/EventEmitter.pm line 15
```

Martchus:
It doesn't work because the `Carp::Always` version in Leap 15.2/15.3 is too old. I've been adding a newer version to our devel repos. One more package cannot hurt anymore at this point and the required version has already been released 2018. I'm bumping the version requirement within this PR.

Martchus:
And now the cache generation doesn't work because it cannot find Leap 15.2 repos: https://app.circleci.com/pipelines/github/os-autoinst/openQA/7478/workflows/c161f046-c329-4e4f-83be-839bd32ca005/jobs/70530/parallel-runs/0/steps/0-108

PR #4138, opened 2021-08-23

okurz:


lib/OpenQA/WebAPI/Controller/Test.pm
```diff
@@ -400,12 +388,8 @@ sub get_current_job {
...
     return $self->reply->not_found if (!defined $self->param('testid'));
 
-    my $job = $self->schema->resultset("Jobs")->search(
-        {
-            id => $self->param('testid')
-        },
-        {prefetch => qw(jobs_assets)})->first;
-    return $job;
+    return $self->schema->resultset("Jobs")->search({id => $self->param('testid')}, {prefetch => qw(jobs_assets)})
+      ->first;
```
Martchus, line 392:
```suggestion
    return $self->schema->resultset('Jobs')->find($self->param('testid'}, {prefetch => qw(jobs_assets)});
```

Martchus (approved):

lib/OpenQA/WebAPI/Controller/Test.pm
okurz: should have tried a multi-line regex :D Fixed and found many more instances

Martchus:
I will review this when https://github.com/os-autoinst/openQA/pull/4202 has been merged because I'm not checking each and every search/find substitution here again.

lib/OpenQA/WebAPI/Controller/Step.pm
```diff
@@ -19,7 +31,7 @@ sub _init {
     my ($self) = @_;
 
     return 0 unless my $job = $self->app->schema->resultset('Jobs')->find($self->param('testid'));
-    my $module = $job->modules->search({name => $self->param('moduleid')})->first;
+    my $module = $job->modules->find({name => $self->param('moduleid')});
```
Martchus, line 34: There's no unique constraint on the name column alone so this query might return multiple results. Better limit it to one row:

```suggestion
    my $module = $job->modules->find({name => $self->param('moduleid')}, {rows => 1});
```

It is a bit unfortunate that the database schema's unique constraints don't match the way API parameters are specified but I suppose improving that is out of scope for this PR. Just ensure we don't end up with Perl warnings in the logs due to this limitation.

---

Please also check whether other changes of that kind fall into the same category. A query via `find` must not return more than one row.

Martchus (changes requested):

lib/OpenQA/Schema/ResultSet/DeveloperSessions.pm
```diff
@@ -14,34 +24,24 @@ use OpenQA::Schema::Result::DeveloperSessions;
 use OpenQA::WebSockets::Client;
 use OpenQA::Log 'log_error';
 
-sub register {
-    my ($self, $job_id, $user_id) = @_;
+sub _register_txn ($self, $job_id, $user_id) {
```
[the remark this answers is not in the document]
okurz: used your suggestion, thxt

lib/OpenQA/WebAPI/Controller/Step.pm
okurz: I moved all "find" related changes to https://github.com/os-autoinst/openQA/pull/4202

PR #4139, opened 2021-08-24
Update documentation to commit 94bf36033

openqabot:


Martchus:
It looks still broken. I assume the code should have been removed completely.

okurz:
> I assume the code should have been removed completely.

What do you mean with that?

okurz:
I will try to generate the documentation manually with the asciidoctor conversion before submitting a new PR for os-autoinst. Previously rendering of direct POD conversion looked fine but it seems Pod::Asciidoctor behaves different

okurz:
https://github.com/os-autoinst/os-autoinst/pull/1747

PR #4141, opened 2021-08-24
Add filter for searching modules by regular expression 

mimi1vx:
poo#96058

Adds 'module_re' parameter to /test/overview which
used as regular expresion for git grep over distri directories.

### DIFF
--- a/lib/OpenQA/Schema/ResultSet/Jobs.pm
+++ b/lib/OpenQA/Schema/ResultSet/Jobs.pm
@@ -21,9 +21,13 @@ use base 'DBIx::Class::ResultSet';
 
 use DBIx::Class::Timestamps 'now';
 use Date::Format 'time2str';
+use File::Basename 'basename';
+use IPC::Run;
 use OpenQA::App;
 use OpenQA::Log qw(log_debug log_warning);
 use OpenQA::Schema::Result::JobDependencies;
+use OpenQA::Utils 'testcasedir';
+use Mojo::File 'path';
 use Mojo::JSON 'encode_json';
 use Mojo::URL;
 use Time::HiRes 'time';
@@ -225,9 +229,30 @@ sub create_from_settings {
     return $job;
 }
 
+sub search_modules ($self, $module_re) {
+    my $distris = path(testcasedir);
+    my @results;
+    for my $distri ($distris->list({dir => 1})->map('realpath')->uniq()->each) {
+        next unless -d $distri;
+
+        my @cmd = ('git', '-C', $distri, 'grep', '--no-index', '-l', $module_re, '--', '*.p[my]');
+        my $stdout;
+        my $stderr;
+        IPC::Run::run(\@cmd, \undef, \$stdout, \$stderr);
+        next if $stderr;
+        push(@results, map { $_ =~ s/\..*$//; basename $_} split(/\n/, $stdout));
+    }
+    return \@results;
+}
+
 sub prepare_complex_query_search_args ($self, $args) {
     my @conds;
     my @joins;
+
+    if ($args->{module_re}) {
+        my $modules = $self->search_modules($args->{module_re});
+        push @{$args->{modules}}, @$modules;
+    }
 
     if ($args->{modules}) {
         push @joins, 'modules';
--- a/t/data/openqa/share/tests/opensuse/tests/openQA/kate.pm
+++ b/t/data/openqa/share/tests/opensuse/tests/openQA/kate.pm
@@ -1,0 +1,23 @@
+# SUSE's openQA tests
+#
+# Copyright © 2009-2013 Bernhard M. Wiedemann
+# Copyright © 2012-2017 SUSE LLC
+#
+# Copying and distribution of this file, with or without modification,
+# are permitted in any medium without royalty provided the copyright
+# notice and this notice are preserved.  This file is offered as-is,
+# without any warranty.
+
+# Package: kate
+# Summary: Test the KDE text editor can be installed, started, typing works
+#   and closed
+# Maintainer: Oliver Kurz <okurz@suse.de>
+
+use strict;
+use warnings;
+
+sub run {
+    my @regexp = qw(test-kate-3 tes-module_re);
+}
+
+1;
--- a/t/ui/10-tests_overview.t
+++ b/t/ui/10-tests_overview.t
@@ -150,7 +150,7 @@ is(scalar @filtered_out, 0, 'result filter correctly applied');
 # Test whether all URL parameter are passed correctly
 my $url_with_escaped_parameters
   = $baseurl
-  . 'tests/overview?arch=&flavor=&machine=&test=&modules=&distri=opensuse&build=0091&version=Staging%3AI&groupid=1001';
+  . 'tests/overview?arch=&flavor=&machine=&test=&modules=&module_re=&distri=opensuse&build=0091&version=Staging%3AI&groupid=1001';
 $driver->get($url_with_escaped_parameters);
 $driver->find_element('#filter-panel .card-header')->click();
 $driver->find_element('#filter-form button')->click();
@@ -352,6 +352,33 @@ subtest 'filtering by module' => sub {
         element_visible('#res_DVD_i586_kde');
         element_visible('#res_DVD_i586_textmode');
     };
+};
+
+subtest 'filtering by module_re' => sub {
+    my $module_re         = 'Maintainer.*okurz';
+    my $job_icon_selector = 'td[id^="res_DVD_"]';
+    my $result            = 'failed';
+
+    subtest "jobs containing the module with any result are present" => sub {
+        my $number_of_found_jobs = 3;
+        $driver->get("/tests/overview?arch=&distri=opensuse&module_re=$module_re");
+        my @jobs = $driver->find_elements($job_icon_selector);
+        # Assert that all the jobs with the specified module are shown in the results
+        is(scalar @jobs, $number_of_found_jobs, "$number_of_found_jobs jobs with \"$module_re\" regexp module found");
+        element_visible('#res_DVD_i586_kde');
+        element_visible('#res_DVD_x86_64_kde');
+        element_visible('#res_DVD_x86_64_doc');
+    };
+
+    subtest "jobs containing the module with the specified result are present" => sub {
+        my $number_of_found_jobs = 1;
+        $driver->get("/tests/overview?arch=&distri=opensuse&module_re=$module_re&modules_result=$result");
+        my @jobs = $driver->find_elements($job_icon_selector);
+        # Assert that all the jobs with the specified module and result are shown in the results
+        is(scalar @jobs, $number_of_found_jobs, "$number_of_found_jobs jobs with \"$module_re\" module regexp found");
+        element_visible('#res_DVD_i586_kde');
+    };
+
 };
 
 subtest "filtering by machine" => sub {
--- a/templates/webapi/test/overview.html.ep
+++ b/templates/webapi/test/overview.html.ep
@@ -126,12 +126,18 @@
                     </div>
                 </div>
                 <div class="row" id="filter-modules">
-                    <div class="col-8">
+                    <div class="col-4">
                         <div class="form-group">
                             <label for="filter-modules"><strong>Module</strong></label>
                             <input name="modules" type="text" id="modules"
                             placeholder="modules, comma separated, e.g. mod1,mod2"
                             class="form-control">
+                        </div>
+                    </div>
+                    <div class="col-4">
+                        <div class="form-group">
+                            <label for="filter-module-re"><strong>Test module</strong></label>
+                            <input type="text" class="form-control" name="module_re" placeholder="regular expression" id="filter-module-re">
                         </div>
                     </div>
                     <div class="col-2">
--- a/assets/javascripts/overview.js
+++ b/assets/javascripts/overview.js
@@ -119,6 +119,9 @@ function setupOverview() {
     } else if (key === 'machine') {
       $('#filter-machine').prop('value', val);
       return val;
+    } else if (key === 'module_re') {
+      $('#filter-module-re').prop('value', val);
+      return val;
     } else if (key === 'modules') {
       $('#modules').prop('value', val);
       return val;
--- a/lib/OpenQA/WebAPI/Plugin/Helpers.pm
+++ b/lib/OpenQA/WebAPI/Plugin/Helpers.pm
@@ -346,6 +346,7 @@ sub _compose_job_overview_search_args ($c) {
     $v->optional('test',           'not_empty');
     $v->optional('modules',        'comma_separated', 'not_empty');
     $v->optional('modules_result', 'not_empty');
+    $v->optional('module_re',      'not_empty');
     $v->optional('group',          'not_empty');
     $v->optional('groupid',        'not_empty');
     $v->optional('id',             'not_empty');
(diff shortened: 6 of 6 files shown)


okurz:
Looks like what you pass as `$args->{regexp}` might be already an array before you pass it into the `rg_files` method. But without a proper test I don't know how you test it anyway. I strongly suggest https://en.wikipedia.org/wiki/Test-driven_development . Or maybe you even want to start with the documentation of how this feature should be used before even starting the implementation.

lib/OpenQA/Schema/ResultSet/Jobs.pm
```diff
@@ -225,10 +227,34 @@ sub create_from_settings {
     return $job;
 }
 
+sub rg_files {
+    my ($self, $dir, $regexp) = @_;
+    my @ret;
+    my $res = run_cmd_with_log_return_error(
+       ['rg', '--type','perl', '--type', 'py', '-l', qq('$regexp'), $dir]);
```
okurz, line 234: In https://progress.opensuse.org/issues/96058 we have considered `git grep` which would work without additional dependencies. I am ok with trying rg as an option but so far I still consider git a better alternative

lib/OpenQA/Schema/ResultSet/Jobs.pm
```diff
@@ -225,10 +227,34 @@ sub create_from_settings {
...
+    }
+    return @ret
+}
+
+
 sub prepare_complex_query_search_args ($self, $args) {
     my @conds;
     my @joins;
 
+    if ($args->{regexp}) {
```
okurz, line 252: A parameter "regexp" might be a bit ambiguous. It only applies to test modules. How about `module_re`?

okurz (changes requested):

lib/OpenQA/Schema/ResultSet/Jobs.pm
```diff
@@ -225,10 +227,34 @@ sub create_from_settings {
     return $job;
 }
 
+sub rg_files {
+    my ($self, $dir, $regexp) = @_;
```
Martchus, line 231: It would be nice to use signatures for new functions.

lib/OpenQA/Schema/ResultSet/Jobs.pm
```diff
@@ -225,10 +227,34 @@ sub create_from_settings {
...
+    my $files = $res->{stderr} if $res->{stderr};
+
+    return @ret unless $files;
+
+    for my $line (split(/\n/, $files)) {
+        $line = basename $line;
+        $line =~ s/\.+$//;
+        push(@ret, $line);
+    }
+    return @ret
```
Martchus, line 244: Returning the array by reference (instead of by copy) would be better:
```suggestion
    return \@ret
```
The same counts for the early-return of course.

templates/webapi/test/overview.html.ep
```diff
@@ -124,6 +124,12 @@
                             <input type="text" class="form-control" name="test" placeholder="any" id="filter-test">
                         </div>
                     </div>
+                    <div class="col-5">
+                        <div class="form-group">
+                            <strong>Module regexp</strong>
```
Martchus, line 129: I would avoid using abbreviations too much. Maybe that would look better:
```suggestion
                            <strong>Test module</strong> (regular expression)
```

lib/OpenQA/Schema/ResultSet/Jobs.pm
```diff
@@ -225,10 +227,34 @@ sub create_from_settings {
...
+}
+
+
```
Martchus, line 247:
```suggestion

```

Martchus (changes requested):
The commit messages should follow https://chris.beams.io/posts/git-commit.

perlpunk:
Do we really need a regular expression, or could simple wildcard matching be sufficient?
A malicious regex can use all CPU, and since it's a form parameter, anyone can submit it.

okurz:
> Do we really need a regular expression, or could simple wildcard matching be sufficient?
> A malicious regex can use all CPU, and since it's a form parameter, anyone can submit it.

Hm, the same would apply for all other search parameters. Also, here I suggested to supply the regex directly into a `git grep` command. Can you provide more details about this? So regex is generally bad or can we apply some "web-safe" limitations? If we opt for "wildcard matching" then I would suggest to use the match parameter "module_content" instead of "module_re"

Martchus:
I suppose we already allow the user to provide a search regex in several other places. I'm wondering whether it is really that bad and whether it would be possible to limit regex usage to a safe subset of features.

> Can you provide more details about this?

We've already noticed that `grep` can be slow on some regexes as it caused the investigation hook to be quite slow. Also this news article comes to my mind https://www.phoronix.com/scan.php?page=news_item&px=GNU-Grep-3.7 although this was clearly just a bug in grep itself.

okurz:
> We've already noticed that `grep` can be slow on some regexes as it caused the investigation hook to be quite slow

yes but I would not compare "git grep" in os-autoinst-distri-opensuse with a multi-line grep on a megabyte long autoinst-log.txt

okurz (commented):
I think a good candidate test to extend could be https://github.com/os-autoinst/openQA/blob/master/t/10-tests_overview.t . So try to run `make test TESTS=t/10-tests_overview.t`, then extend in there. Other candidates to extend could be https://github.com/os-autoinst/openQA/blob/master/t/10-jobs.t or for the UI-only part https://github.com/os-autoinst/openQA/blob/master/t/ui/10-tests_overview.t but I don't think we need to test the UI-only part as it is trivial.

lib/OpenQA/Schema/ResultSet/Jobs.pm
```diff
@@ -227,19 +228,25 @@ sub create_from_settings {
...
 }
 
-sub rg_files ($self, $dir, $module_re) {
-    my $res = run_cmd_with_log_return_error(['rg', '--type','perl', '--type', 'py', '-l', $module_re, $dir]);
-    return [] unless my $files = $res->{stderr};
-    return [map { $_ =~ s/\..*$//; basename $_ } split(/\n/, $files)];
+sub search_modules ($self, $module_re) {
+    my $distris = path(testcasedir);
+    my @results;
+    for my $distri ($distris->list({dir => 1})->map('realpath')->uniq()->each) {
```
okurz, line 234: this looks neat :)

okurz (changes requested):
Sorry, I don't understand the git commit message "git grep based on chris search_modules in api". 

Please also explain why you add the function "run_cmd_with_log_return_all"

okurz (changes requested):
Still, please, the commit message should have a bit more details, e.g. what issue is fixed, why this needs to change, to which versions of which product it applies, link to a bug or a fate entry, the choices you made, etc. Also see https://commit.style/ or http://chris.beams.io/posts/git-commit/ as a helpful guide how to write good commit messages.

Also please explain why you add the new function "run_cmd_with_log_return_all"

lib/OpenQA/Utils.pm
```diff
@@ -257,6 +258,41 @@ sub run_cmd_with_log {
...
+        chomp $stderr;
+        if ($ipc_run_succeeded) {
+            log_debug($stdout);
+            log_info("cmd returned $return_code");
+        }
+        else {
+            log_warning($stderr);
+            log_error("cmd returned $return_code");
+        }
+        return {
```
Martchus, line 278: This will only return from the `try` block because the `{}`-braces of the `try`-block actually enclose a nested function. This only works because the function doesn't to anything after the `try-catch` anyways. However, to avoid future bugs I'd write that as `eval { … }` followed by `if (my $error = $@) { … }`.

lib/OpenQA/Schema/ResultSet/Jobs.pm
mimi1vx: > Sorry, I don't understand the git commit message "git grep based on chris search_modules in api".
> 
I't was temporary commit message
> Please also explain why you add the function "run_cmd_with_log_return_all"

Wanted function which returns correctly stdout and stderr ( not merged ), and I dont like in place solution like in https://github.com/os-autoinst/openQA/blob/7d05a925004d7ec4738568a06c786aebdd1f4bb0/lib/OpenQA/WebAPI/Controller/API/V1/Search.pm#L73-L96

lib/OpenQA/Utils.pm
```diff
@@ -257,6 +258,41 @@ sub run_cmd_with_log {
...
+    log_info("Running cmd: " . join(' ', @$cmd));
+    try {
+        my ($stdout, $stderr);
+        my $ipc_run_succeeded = IPC::Run::run($cmd, \undef, \$stdout, \$stderr);
+        my $return_code       = $?;
+        chomp $stdout;
+        chomp $stderr;
+        if ($ipc_run_succeeded) {
+            log_debug($stdout);
+            log_info("cmd returned $return_code");
```
Martchus, line 272: In the context this code runs (the web UI service) this log message is not very meaningful and raises the question what command it is about. Just printing `$stdout` without context is also not very meaningful. I suppose it is sufficient to return that information back to the user.

lib/OpenQA/Utils.pm
```diff
@@ -257,6 +258,41 @@ sub run_cmd_with_log {
...
+        }
+        else {
+            log_warning($stderr);
```
Martchus, line 275: The lack of context is problematic here as well and maybe even in the error case it is sufficient to report it back to the user.

Martchus (dismissed):

lib/OpenQA/Schema/ResultSet/Jobs.pm
```diff
@@ -225,10 +229,31 @@ sub create_from_settings {
     return $job;
 }
 
+sub search_modules ($self, $module_re) {
```
kalikiana, line 232: Any chance we can re-use [the existing function](https://github.com/os-autoinst/openQA/blob/84e973fa8af13a159ecb68fad82bf0f250aeb34b/lib/OpenQA/WebAPI/Controller/API/V1/Search.pm#L73) rather than copying it here?

kalikiana: Or split out a part of it for the sake of the argument

mimi1vx: isnt same , different command and different results needed

okurz (changes requested):
The UI test is mocking the actual git grep call, that is ok. But now you should still add a test calling the new "search_modules" function. However in there you might be able to mock out the actual git grep call. Our goal for code coverage is 100% statement coverage with explicit exceptions mentioned with "# uncoverable statement" so at the very least we should ensure to not have less coverage than in before, see the codecov comments and result report.

lib/OpenQA/Schema/ResultSet/Jobs.pm
```diff
@@ -225,10 +229,31 @@ sub create_from_settings {
...
+    my $distris = path(testcasedir);
+    my @results;
+    for my $distri ($distris->list({dir => 1})->map('realpath')->uniq()->each) {
+        next unless -d $distri;
+
+        my @cmd = ('git', '-C', $distri, 'grep', '-l', $module_re, '--', '*.p[my]');
+        my $stdout;
+        my $stderr;
+        IPC::Run::run(\@cmd, \undef, \$stdout, \$stderr);
+        next if $stderr;
```
perlpunk, line 242: Do we want to know if the grep failed? We should probably check for the exit code then. It's in `$?`.
E.g. try with the input `foo[z-a]`

mimi1vx: `Invalid range end` ... but still I don't think this is interesting for user

t/data/openqa/share/tests/opensuse/tests/openQA/kate.pm
```diff
@@ -0,0 +1,20 @@
...
+# Copying and distribution of this file, with or without modification,
+# are permitted in any medium without royalty provided the copyright
+# notice and this notice are preserved.  This file is offered as-is,
+# without any warranty.
+
+# Package: kate
+# Summary: Test the KDE text editor can be installed, started, typing works
+#   and closed
+# Maintainer: Oliver Kurz <okurz@suse.de>
+
```
Martchus, line 15: I would avoid the huge comment.

okurz: I guess it's fine as an example

lib/OpenQA/Schema/ResultSet/Jobs.pm
kalikiana: I guess you're passing a regex rather than keywords. Besides that the grep is the same.

t/data/openqa/share/tests/opensuse/tests/openQA/kate.pm
kalikiana: It doesn't seem like the test covers this? If the metadata is searched, I would welcome it

mimi1vx: is used by `t/ui/10-test_overview.t`

t/ui/10-tests_overview.t
```diff
@@ -18,6 +18,7 @@ use Test::Most;
 use FindBin;
 use lib "$FindBin::Bin/../lib", "$FindBin::Bin/../../external/os-autoinst-common/lib";
 use Date::Format 'time2str';
+use Test::MockModule 'strict';
```
kraih, line 21: Doesn't look like you're using the module.

lib/OpenQA/Schema/ResultSet/Jobs.pm
```diff
@@ -225,10 +229,32 @@ sub create_from_settings {
...
+sub search_modules ($self, $module_re) {
+    my $distris = path(testcasedir);
+    my @results;
+    for my $distri ($distris->list({dir => 1})->map('realpath')->uniq()->each) {
+        next unless -d $distri;
+
+        my @cmd = ('git', '-C', $distri, 'grep', '--no-index', '-l', $module_re, '--', '*.p[my]');
+        my $stdout;
+        my $stderr;
+        print("Distri: " . $distri);
```
kraih, line 241: Is the `print` leftover debug code?

templates/webapi/test/overview.html.ep
```diff
@@ -124,6 +124,12 @@
                             <input type="text" class="form-control" name="test" placeholder="any" id="filter-test">
                         </div>
                     </div>
+                    <div class="col-5">
+                        <div class="form-group">
+                            <strong>Test module</strong> (regular expression)
+                            <input type="text" class="form-control" name="module_re" placeholder="any" id="filter-module-re">
+                        </div>
+                    </div>
```
kraih, line 132: A Bootstrap column has 12 rows, you're using 15 here if i'm not mistaken.

lib/OpenQA/Schema/ResultSet/Jobs.pm
mimi1vx: thx, is

templates/webapi/test/overview.html.ep
mimi1vx: thx, learned something new :D

t/ui/10-tests_overview.t
```diff
@@ -354,6 +354,33 @@ subtest 'filtering by module' => sub {
     };
 };
 
+subtest 'filtering by module_re' => sub {
+    my $module_re         = 'test-kate';
```
okurz, line 358: I don't quite get the search approach here. "kate" is the name of the test module but the feature is about searching *within* the module content, so better search for something that does not appear in the filename. A realistic use case would be to search for the maintainer field. That would also make cdywan happy because you would search in the metadata. How about you search for something like `/Maintainer *okurz/`?

mimi1vx: ok, why not

kraih: And maybe add one more test case with a slightly different regex that matches nothing.

okurz (approved):

perlpunk (approved):

PR #4144, opened 2021-08-25
Revert "Let workers declare themselves as broken if asset downloads are piling"

asdil12:
Reverts os-autoinst/openQA#4122

Ticket: https://progress.opensuse.org/issues/97136

### DIFF
--- a/lib/OpenQA/CacheService/Response.pm
+++ b/lib/OpenQA/CacheService/Response.pm
@@ -17,7 +17,6 @@ package OpenQA::CacheService::Response;
 use Mojo::Base -base;
 
 has [qw(data error)];
-has max_inactive_jobs => $ENV{OPENQA_CACHE_MAX_INACTIVE_JOBS} // 5;
 
 sub has_error { !!shift->error }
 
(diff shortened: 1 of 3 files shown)


Martchus:
You could also keep the feature and just configure `OPENQA_CACHE_MAX_INACTIVE_JOBS` to something very high. I've already added an override within salt so it would be quite easy to do.

I suppose the only problem this change creates is the alert which can also have different causes. It would be easy to fix just the alert by simply adjusting the PostgreSQL query to exclude such errors from the broken workers statistics. We already exclude `graceful disconnect` errors (as they are obviously no errors at all) and you can do the same for `Cache service queue already full`. Just search for `graceful disconnect` in salt-states-openqa to find the SQL query to adjust.

okurz (approved):
@Martchus we should look for a solution here in upstream, not just within salt states

Martchus:
Some upstream feature where we distinguish between "overloaded with downloads" and "broken for any other reason"? From the upstream perspective that's already distinguishable because we show that within the web UI. I just think we need to distinguish this within our monitoring as well.

(Maybe "broken" wasn't the best word to chose for a state which can be only temporary but that is what it means also in any other cases, e.g. a QEMU process being stuck for a while.)

asdil12:
https://gitlab.suse.de/openqa/salt-states-openqa/-/merge_requests/565

PR #4145, opened 2021-08-25
Prevent too many jobs returned by queries by default

okurz:
With this limit the default query results on 1000 entries. With the
query parameter "limit" one can still select a higher limit.

Motivated by https://github.com/grisu48/openqa-mon/pull/57

Related progress issue: https://progress.opensuse.org/issues/97190


Martchus (changes requested):
This also limits the tables for running and scheduled/blocked jobs on the "All tests" page as they also use the `complex_query` function. That actually makes sense. However, we should take care to make obvious that not all jobs are displayed (like it used to be). For instance, when the row count of the table is equivalent to the limit the caption could have a "Last …" prefix like we already have for finished jobs.

It should also be tested whether the `limit` parameter specified in the address bar would actually be used for the further AJAX queries. If not, I suppose the limit parameter for the "All tests" page should apply to all tables.

I must also say that 1000 seems a little bit less. We have over 1300 scheduled jobs on OSD right now and loading the table doesn't take very long. For scheduled and running jobs it would also still be nice to see the total number somewhere.

okurz:
Updated the PR to limit the jobs internally but again supply "undef" for the webUI. So the display on the webUI should not be affected.

For potential further steps:

> This also limits the tables for running and scheduled/blocked jobs on the "All tests" page as they also use the `complex_query` function. That actually makes sense. However, we should take care to make obvious that not all jobs are displayed (like it used to be). For instance, when the row count of the table is equivalent to the limit the caption could have a "Last …" prefix like we already have for finished jobs.
>
> I must also say that 1000 seems a little bit less. We have over 1300 scheduled jobs on OSD right now and loading the table doesn't take very long. For scheduled and running jobs it would also still be nice to see the total number somewhere.

ok so I propose we could do the following: In OpenQA::WebAPI::Controller::Test::list_running_ajax next to the array of all running jobs also return the limit parameter and the list of total entries so that we can display both numbers and add a "Last …" if `total > limit`

lib/OpenQA/WebAPI/Controller/Test.pm
```diff
@@ -121,6 +121,7 @@ sub list_running_ajax {
         state => [OpenQA::Jobs::Constants::EXECUTION_STATES],
         match => $self->get_match_param,
         groupid => $self->param('groupid'),
+        limit => ($self->param('limit') // undef),    # show all scheduled by default
```
cfconrad, line 124:
```suggestion
        limit => ($self->param('limit') // undef),    # show all running by default
```

okurz: done, thx

lib/OpenQA/WebAPI/Controller/Test.pm
```diff
@@ -121,6 +121,7 @@ sub list_running_ajax {
         state => [OpenQA::Jobs::Constants::EXECUTION_STATES],
         match => $self->get_match_param,
         groupid => $self->param('groupid'),
+        limit => ($self->param('limit') // undef),    # show all running by default
```
Martchus, line 124: I don't think this will work because if it is `undef` you have the fallback to `OpenQA::App->singleton->config->{global}->{query_limit}` in `prepare_complex_query_search_args`.

Martchus (changes requested):

PR #4156, opened 2021-08-30
Simplify OpenQA::WebAPI::Auth::OpenID

okurz:
* Use signatures in OpenQA::WebAPI::Auth::OpenID
* Simplify OpenQA::WebAPI::Auth::OpenID

### DIFF
--- a/lib/OpenQA/WebAPI/Auth/OpenID.pm
+++ b/lib/OpenQA/WebAPI/Auth/OpenID.pm
@@ -63,6 +63,29 @@ sub auth_login ($self) {
     return (error => $csr->err);
 }
 
+sub _first_last_name ($ax) { join(' ', $ax->{'value.firstname'} // '', $ax->{'value.lastname'} // '') }
+
+sub _create_user ($self, $id, $email, $nickname, $fullname) {
+    $self->schema->resultset('Users')->create_user($id, email => $email, nickname => $nickname, fullname => $fullname);
+}
+
+sub _handle_verified ($self, $vident) {
+    my $sreg = $vident->signed_extension_fields('http://openid.net/extensions/sreg/1.1');
+    my $ax = $vident->signed_extension_fields('http://openid.net/srv/ax/1.0');
+
+    my $email = $sreg->{email} || $ax->{'value.email'} || 'nobody@example.com';
+    my $nickname = $sreg->{nickname} || $ax->{'value.nickname'} || $ax->{'value.firstname'};
+    unless ($nickname) {
+        my @a = split(/\/([^\/]+)$/, $vident->{identity});
+        $nickname = $a[1];
+    }
+
+    my $fullname = $sreg->{fullname} || $ax->{'value.fullname'} || _first_last_name($ax) || $nickname;
+
+    $self->_create_user($vident->{identity}, $email, $nickname, $fullname);
+    $self->session->{user} = $vident->{identity};
+}
+
 sub auth_response ($self) {
     my %params = @{$self->req->params->pairs};
     my $url = $self->app->config->{global}->{base_url} || $self->req->url->base;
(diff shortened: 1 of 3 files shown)


kraih:
I wouldn't put too much effort into this. With OpenID Connect support coming to `Mojolicious::Plugin::OAuth2` and legacy OpenID being deprecated, we should be able to get rid of this code at some point in the not so distant future.

okurz:
I agree. But the above refactoring was really just a 10 minute effort while actively reviewing https://github.com/os-autoinst/openQA/pull/4151

okurz:
Fixed a typo and a compile error. Ran t/03-auth.t locally only. Let's await complete CI results.

Martchus:
Unfortunately the patch coverage doesn't look great.

okurz:
I tested locally with

```
sed -i 's/^# method = .*OpenID/method = OpenID/;s/^#httpsonly = 1/httpsonly = 0/' etc/openqa/openqa.ini
NOT_HEADLESS=1 perl -Ilib -d t/ui/01-list.t
```

and hit

```
Can't locate object method "_handle_verified" via package "OpenQA::Shared::Controller::Session" at lib/OpenQA/WebAPI/Auth/OpenID.pm line 122
```

likely we should try to increase the test coverage with automatic tests.

okurz:
I extended unit tests also for the method _handle_verified but I am still running into

```
Can't locate object method "_handle_verified" via package "OpenQA::Shared::Controller::Session" at lib/OpenQA/WebAPI/Auth/OpenID.pm line 122
```

when running something like `NOT_HEADLESS=1 perl -Ilib -d t/ui/01-list.t`. Some help would be appreciated.

lib/OpenQA/WebAPI/Auth/OpenID.pm
```diff
@@ -9,19 +9,19 @@ use LWP::UserAgent;
 use Net::OpenID::Consumer;
 use MIME::Base64 qw(encode_base64url decode_base64url);
 
-sub auth_login ($self) {
-    my $url = $self->app->config->{global}->{base_url} || $self->req->url->base->to_string;
+sub auth_login ($c) {
```
Martchus, line 12: We actually normally call the controller `$self` in functions of the controller packages. However, calling it `$c` is also fine I suppose.

Martchus (commented):
The patch coverage check is still complaining but the change looks generally good (code is mainly moved around).

kraih:
That reminds me, i've replaced OpenID with OpenID Connect in Cavil now. The change can probably be replicated for openQA to simplify `OpenQA::WebAPI::Auth::OpenID` further. https://github.com/openSUSE/cavil/commit/24b08a5e1eeda5be3cc91ea97e974f1d70cd29b0

okurz:
> That reminds me, i've replaced OpenID with OpenID Connect in Cavil now. The change can probably be replicated for openQA to simplify `OpenQA::WebAPI::Auth::OpenID` further. [openSUSE/cavil@24b08a5](https://github.com/openSUSE/cavil/commit/24b08a5e1eeda5be3cc91ea97e974f1d70cd29b0)

That looks nice. Would be cool if you could propose a PR doing that.

okurz:
> The patch coverage check is still complaining but the change looks generally good (code is mainly moved around).

Yes, we should extend test coverage first. If anyone of you fancies doing so be my guest otherwise a candidate for a mob session.

kraih:
> That looks nice. Would be cool if you could propose a PR doing that.

Sure, once it's in the backlog. 😉 https://progress.opensuse.org/issues/89023

okurz:
Fixed the problem https://github.com/os-autoinst/openQA/pull/4156#issuecomment-1216187921 with help from @Martchus . I shouldn't try to call member methods on the object when those are actually called on the controller. Now properly calling those functions as free functions with `$c` as first argument. Maybe that was all that was necessary

perlpunk (approved):

lib/OpenQA/Setup.pm
```diff
@@ -213,7 +213,7 @@ sub read_config ($app) {
         },
         test => {
             auth => {
-                method => 'Fake',
+                #method => 'Fake',
```
Martchus, line 216: Are these commented-out lines still required? If yes, they should maybe be removed completely - although I'm not sure what this will break.

Martchus (changes requested):

lib/OpenQA/Setup.pm
okurz: thank you. Fixed. I don't know how I managed to include those local testing changes again.

Martchus (approved):

okurz:
Everything good except patch coverage. As agreed in before merging based on manual test results.

Issue #4158, opened 2021-09-01
Consider moving to Fork-Awesome or another free font


phil-hands:
Hi!

While packaging OpenQA for Debian I noticed that OpenQA now uses Font-Awesome v5: os-autoinst/openQA@89a7480671bb326078ba122774f2b81ec7b285a6

I presume that the reason for that change was simply to update to the latest and greatest, but unfortunately it comes at the cost of making OpenQA impossible to build from source.

As of v5, FontAwesome switched to a proprietary build system and it is no longer possible to build the icons. Of course, this was done silently and I am sure you were not aware of this when migrating.

We were in contact with Font-Awesome developers (see [Debian Bug #902981](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=902981#76)) and they responded clearly (and it seems that nothing's changed since):

    the first question to get out of the way is: “Will Font Awesome open source their build tool?”

    And the answer is no for the time being.

This makes it impossible, for example, for Font-Awesome v5 to be included in Debian.

The community reacted to this at the time, and the [Fork-Awesome project](https://forkaweso.me) was created, which is a fork of Font-Awesome v4, from back when the build system was still available.

BTW Fork-Awesome has a v5 compatibility layer (see: ForkAwesome/Fork-Awesome#112) which was introduced when the same issue was noticed in [Syncthing]( https://github.com/syncthing/syncthing/pull/5237).  I don't know if the compatibility layer is really necessary for OpenQA's usage, or whether it would be cleaner to revert the change linked above and just use Fork-Awesome directly.

FWIW I have tested the compatibility layer with OpenQA, and to my (somewhat unschooled eye) it seems to work well.

Would OpenQA consider migrating to Fork-Awesome?

Cheers, Phil.

coolo:
Would you even have noticed if we copied font awesome into our assets folder and put a CC-BY-4.0 license next to it? I assume CC-BY-4.0 works for debian.

kraih:
Mojolicious has files bundled under`CC-BY-SA` license (which is more strict than `CC-BY`), so it should be fine for Debian.

Martchus:
We're not using Font-Awesome's build system but just downloading their released files. However, it would likely be possible to simply adjust the URLs in `assetpack.def` (and then follow https://open.qa/docs/#update-asset-cache).

phil-hands:
[...]

PR #4159, opened 2021-09-01
Migrate from FontAwesome to ForkAwesome

Martchus:
See https://github.com/os-autoinst/openQA/issues/4158

### DIFF
--- a/assets/assetpack.def
+++ b/assets/assetpack.def
@@ -1,19 +1,7 @@
 ! bootstrap.css
 # download the fonts
 < https://cdn.datatables.net/1.10.16/css/dataTables.bootstrap4.css
-< https://use.fontawesome.com/releases/v5.0.10/css/all.css
-<< https://use.fontawesome.com/releases/v5.0.10/webfonts/fa-brands-400.eot
-<< https://use.fontawesome.com/releases/v5.0.10/webfonts/fa-brands-400.woff2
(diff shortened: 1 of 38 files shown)


okurz (changes requested):

kraih:
We could just as well use Font Awesome from a CDN, why switch to Fork Awesome? [Here's](https://github.com/mojolicious/mojo/blob/51d18617866de7cf0743e3603fc8a50a43747850/lib/Mojolicious/resources/templates/mojo/debug.html.ep#L16) how the built-in Mojolicious templates use Font Awesome.

Martchus:
We also just use a pre-built version of FontAwesome. I guess the desire is still to use something which can be build from sources, e.g. if it is needed to do changes.

okurz (approved):

Martchus:
The tests failed downloading the Docker image so I retriggered them. The packaging tests fail with download errors but that's expected. When this PR has been merged I'll update the cache.

kalikiana (approved):

Martchus:
I hope no assets are missing anymore. I've checked for missing icons via:

```
wget https://forkaweso.me/Fork-Awesome/icons
icons=($(git grep -ohP 'fa-[\w\-]+' | uniq))
for icon in ${icons[@]}; do grep -q "class=\"fa $icon\"" icons || echo "$icon"; done
```

PR #4160, opened 2021-09-01
Improve docker-compose setup

Martchus:
* See particular commit messages
* Motivated by https://progress.opensuse.org/issues/92893 and also contains a fix for this particular problem


container/webui/nginx-entrypoint.sh
```diff
@@ -1,14 +1,16 @@
...
+  replicas_cfg="server webui_webui_$i:9526;$replicas_cfg"
 done
 
-reg="s/REPLICAS/$replicas/"
-sed "$reg" /etc/nginx/conf.d/default.conf.template > /etc/nginx/conf.d/default.conf
-
-cat /etc/nginx/conf.d/default.conf
+reg="s/REPLICAS/$replicas_cfg/"
+sed "$reg" /etc/nginx/nginx.conf.template > /etc/nginx/nginx.conf
+echo 'NGINX connfiguration:'
```
okurz, line 11: s/connfiguration/configuration/

okurz (changes requested):

Martchus:
Apparently the test uses the worker docker-compose setup independently of the web UI:

```
Network webui_default declared as external, but could not be found. Please create the network manually using `docker network create webui_default` and try again.
Attaching to 
Network webui_default is external, skipping
```

Maybe I can just drop this part of the configuration again.

container/worker/conf/workers.ini
```diff
@@ -1,3 +1,3 @@
 [global]
 BACKEND = qemu
-HOST = http://openqa_webui
+HOST = http://nginx
```
okurz, line 3: Why do you need to rename that? I would prefer a hostname of "openqa_webui" over nginx

okurz (changes requested):

container/worker/conf/workers.ini
Martchus: I didn't rename anything; I'm just making it connect to `nginx` rather than directly to the web UI. However, that only works when sharing the network with the web UI containers (see my last comment). So I'll have to change that somehow anyways. I suppose it would make sense to use the host IP/domain here as a default.

Martchus: I now changed it so the web UI containers and the worker container don't share the same network and are truly independent from each other. Being able to start the web UI containers + worker container on some machine and having them connect with each other is likely still a good default. So I put `172.17.0.1` here for to refer to the container host. This should be a good out-of-the-box experience. Of course I've been adding an explanatory comment what the IP address means and that one can put any host name there as usual.

By the way, putting `172.17.0.1` and an explaining comment there is definitely better than just `openqa_webui` because it is not at all clear whether `openqa_webui` is just an example or refers to an actual host name which is somehow made available by the setup¹.

¹ It isn't clear to me how it would work to connect via that host name so it was likely just an example.

docs/ContainerizedSetup.asciidoc
```diff
@@ -171,116 +168,121 @@ created for upgrade tests) and createhdds.sh is in `openqa_fedora_tools`
...
+It is also possible to use `openqa-clone-job`, e.g. for the docker-compose setup
+one would call it like this:
 
-After performing the "setup" tasks above - do not forget about tests and ISOs
-- you can schedule a test like this:
+    script/openqa-clone-job \
+        --dir "$PWD/container/webui/workdir/data/factory"
+        --apikey $apikey --apisecret $apisecret \
+        --host http://localhost \
+        https://openqa.opensuse.org/tests/1896520
```
kalikiana, line 281: This should be a command run via the container, no?

Martchus: Not necessarily.

Martchus: I've also added an example to run the script within the web UI container.

docs/ContainerizedSetup.asciidoc
```diff
@@ -11,17 +11,17 @@ and the workers.
...
 
 == Get container images
 
 You can either build the images locally or use Fedora images from Docker Hub.
 
 For the `docker-compose` setup it is required to build the images locally.
-However, it is done automatically so this section can be skipped unless one
-wants to build the containers manually.
+However, it is done via `docker-compose` and explained later so this section
+can should skipped.
```
kalikiana, line 24:
```suggestion
can be skipped.
```

docs/ContainerizedSetup.asciidoc
```diff
@@ -171,116 +168,128 @@ created for upgrade tests) and createhdds.sh is in `openqa_fedora_tools`
...
 
-== Setup with openSUSE-based images and docker-compose
+== Setup openQA with openSUSE-based images and docker-compose
+All relative paths in this section are relative to a checkout of openQA's Git
+repository.
 
-=== Run the data and web UI containers in HA mode
+=== Configuration
+The web UI will be available under http://localhost and https://localhost. So it
+is using default HTTP ports by default. Make sure those ports are not used by
```
kalikiana, line 177:
```suggestion
is using default HTTP(S) ports. Make sure those ports are not used by
```

kalikiana (approved):
Love that we finally have TLS here. Note my two minor points on wording, otherwise looking good to me.

docs/ContainerizedSetup.asciidoc
```diff
@@ -171,116 +168,128 @@ created for upgrade tests) and createhdds.sh is in `openqa_fedora_tools`
...
 
 == Running jobs
+After performing the "setup" tasks above - do not forget about tests and ISOs.
+
+Then you can use `openqa-cli` as usual with the containerized web UI.
+
+It is also possible to use `openqa-clone-job`, e.g. for the docker-compose setup
+one would call it like this:
+
+    script/openqa-clone-job \
```
okurz, line 277: If openQA is installed then it should not be necessary to prefix with "script/", I suggest to just write `openqa-clone-job`

docs/ContainerizedSetup.asciidoc
```diff
@@ -171,116 +168,128 @@ created for upgrade tests) and createhdds.sh is in `openqa_fedora_tools`
...
+        --dir "$PWD/container/webui/workdir/data/factory"
+        --apikey $apikey --apisecret $apisecret \
```
okurz, line 279: Please leave out the `apikey` and `apisecrect` parameters. You don't want to know how often I needed to delete references to internal API keys and secrets and I had to delete leaked user keys a lot. The suggested way to go is to use credentials in ~/.config/openqa/client.conf

docs/ContainerizedSetup.asciidoc
```diff
@@ -171,116 +168,128 @@ created for upgrade tests) and createhdds.sh is in `openqa_fedora_tools`
...
+        https://openqa.opensuse.org/tests/1896520
 
-After performing the "setup" tasks above - do not forget about tests and ISOs
-- you can schedule a test like this:
+Of course it is also possible to run it within a container, e.g.:
 
-    docker exec openqa_webui /var/lib/openqa/script/client isos post ISO=Fedora-Server-netinst-x86_64-22_Beta_RC3.iso DISTRI=fedora VERSION=rawhide FLAVOR=generic_boot ARCH=x86_64 BUILD=22_Beta_RC3
+    cd container/webui
+    docker-compose exec webui openqa-clone-job \
+        --apikey $apikey --apisecret $apisecret \
```
okurz, line 287: delete apikey+secret, same as above

okurz (changes requested):

okurz:
@ilausuch I would appreciate your feedback as well

docs/ContainerizedSetup.asciidoc
okurz: Hm, in this case I am not sure it would even be possible to call the script successfully, think of the dependencies. IMHO would be more consistent to run that script from a container as well, similar to how you do it later with `docker exec`

okurz (approved):

PR #4163, opened 2021-09-03
Dependency cron 2021-09-03

openqabot:


okurz (commented):
hm, seems like this opencv change is a direct consequence of the work on https://progress.opensuse.org/issues/97700 . I will actually not merge this right now, maybe the older openCV version is still in openSUSE Factory as well so we should not run into cache problems.

PR #4170, opened 2021-09-06
Optimize group_overview

perlpunk:
Comments were fetched for every job, although those with keys already seen
were skipped and those comments never processed.

This resulted in response times up to 51 seconds for job groups with builds with many jobs.

I tested with the job group 309 locally, and it went from 4.6 to 0.9s.


mimi1vx (approved):

kalikiana (approved):
Nice find!

Martchus (approved):

PR #4171, opened 2021-09-06
Add documentation for upgrading PostgreSQL database

Martchus:
See https://progress.opensuse.org/issues/76924#note-10

### DIFF
--- a/docs/Installing.asciidoc
+++ b/docs/Installing.asciidoc
@@ -299,6 +299,144 @@ For older versions of openQA, you can migrate from SQLite to PostgreSQL
 according to
 <<Pitfalls.asciidoc#db-migration,DB migration from SQLite to PostgreSQL>>
 
+==== Migrate PostgreSQL database on openSUSE
+The PostgreSQL `data`-directory needs to be migrated in order to switch to a
+newer major version of PostgreSQL. The following instructions are specific to
+openSUSE's PostgreSQL and openQA packaging but with a little adaption they can
+likely be used for other setups as well. These instructions can migrate big
+databases in seconds without requiring additional disk space. However, services
+need to be stopped during the (short) migration.
+
+1. Locate the `data`-directory. Its path is configured in
+`/etc/sysconfig/postgresql` and should be `/var/lib/pgsql/data` by default. The
+paths in the next steps assume the default.
+
+2. To ease migrations, I would recommend making the `data`-directory a symlink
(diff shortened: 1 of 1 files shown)


perlpunk (approved):

kraih (approved):

docs/Installing.asciidoc
```diff
@@ -299,6 +299,144 @@ For older versions of openQA, you can migrate from SQLite to PostgreSQL
...
+openSUSE's PostgreSQL and openQA packaging but with a little adaption they can
+likely be used for other setups as well. These instructions can migrate big
+databases in seconds without requiring additional disk space. However, services
+need to be stopped during the (short) migration.
+
+1. Locate the `data`-directory. Its path is configured in
+`/etc/sysconfig/postgresql` and should be `/var/lib/pgsql/data` by default. The
+paths in the next steps assume the default.
+
+2. To ease migrations, I would recommend making the `data`-directory a symlink
```
okurz, line 314: Who is the *I* here? I suggest to write instead s/I would recommend/it is recommended/

docs/Installing.asciidoc
```diff
@@ -299,6 +299,144 @@ For older versions of openQA, you can migrate from SQLite to PostgreSQL
...
+to a versioned directory. So the file system layout would look like this:
++
+[source,sh]
+----
+martchus@ariel:~> sudo -u postgres ls -l /var/lib/pgsql | grep data
```
okurz, line 319: This could be non-personalized

docs/Installing.asciidoc
```diff
@@ -299,6 +299,144 @@ For older versions of openQA, you can migrate from SQLite to PostgreSQL
 according to
 <<Pitfalls.asciidoc#db-migration,DB migration from SQLite to PostgreSQL>>
 
+==== Migrate PostgreSQL database on openSUSE
```
okurz, line 302: This is now a section in the chapter "Basic configuration", better we move it to "Advanced".

okurz (commented):
Implementation of my suggested changes: https://github.com/os-autoinst/openQA/pull/4193

docs/Installing.asciidoc
Martchus: Ok, point taken. I'd only like to note that the steps are not optional anymore if we decide to use newer PostgreSQL features at some point (unless one doesn't care about the data).

PR #4172, opened 2021-09-06
Fix permission errors when running tidy after creating docker-compose setup

Martchus:
Otherwise one gets the following error messages when running tidy because
the database directory for the docker-compose setup is not accessible:

```
find: ‘./container/webui/workdir/db’: Keine Berechtigung
```

Due to the strict error handling of the script this prevents tidy to work.
This change simply prevents `find` from traversing the directory.

### DIFF
--- a/tools/tidy
+++ b/tools/tidy
@@ -19,9 +19,8 @@ EOF
     exit
 }
 
-cleanup() {
-    find . -name '*.tdy' -delete
-}
+find-tdy() { find . \( -type d -path '*/container/webui/workdir/db' -prune -type f \) -or \( -name '*.tdy' \) ; }
+cleanup()  { find-tdy | xargs -r rm ; }
 
 set -eo pipefail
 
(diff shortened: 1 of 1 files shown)


tools/tidy
```diff
@@ -78,8 +77,7 @@ cleanup
 
 find-files | xargs perltidy --pro=.../.perltidyrc
 
-find . -name "*.tdy" | while read -r file
-do
+find-tdy | while read -r file; do
```
kalikiana, line 80: Can we have a [HEALTHCHECK](https://docs.docker.com/engine/reference/builder/#healthcheck) for this? Or alternatively some way to ensure this continues to work?

Martchus: This is not about running the tidy script *within* the container; it is about running tidy normally after invoking docker-compose (as regular user).

perlpunk (approved):

kraih (approved):

PR #4165, opened 2021-09-03
Speed up 12-needle-edit.t by running all Minion jobs in foreground

kraih:
Not a huge win, but a win.

Before
```
t/ui/12-needle-edit.t .. ok    
All tests successful.
Files=1, Tests=19, 77 wallclock secs ( 0.10 usr  0.04 sys + 190.67 cusr 15.16 csys = 205.97 CPU)
Result: PASS
```

After:
```
t/ui/12-needle-edit.t .. ok    
All tests successful.
Files=1, Tests=19, 42 wallclock secs ( 0.12 usr  0.01 sys + 23.23 cusr  1.20 csys = 24.56 CPU)
Result: PASS
```

Progress: https://progress.opensuse.org/issues/96564

PR #4173, opened 2021-09-06
Speed up 21-admin-needles.t by running all Minion jobs in foreground

kraih:
Followup to #4165.

Progress: Progress: https://progress.opensuse.org/issues/96564

### DIFF
--- a/t/lib/OpenQA/SeleniumTest.pm
+++ b/t/lib/OpenQA/SeleniumTest.pm
@@ -25,12 +25,11 @@ use Time::HiRes qw(time sleep);
 use OpenQA::WebAPI;
 use OpenQA::Log 'log_info';
 use OpenQA::Utils;
-use OpenQA::Test::Utils 'collect_coverage_of_gru_jobs';
+use OpenQA::Test::Utils;
 use POSIX '_exit';
 
 our $_driver;
 our $webapi;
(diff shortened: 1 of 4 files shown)


Martchus (approved):

kalikiana (commented):
Interesting that you're getting rid of the gru part of Selenium. Does that mean those tests no longer use gru? Or is it spawned some other way?
Also, how much speed up should we get from this?

kraih:
> Interesting that you're getting rid of the gru part of Selenium. Does that mean those tests no longer use gru? Or is it spawned some other way?

Gru is Minion.

> Also, how much speed up should we get from this?

Just a few seconds, 35s vs 41s  if i remember correctly. Probably the least out of all converted tests.

kalikiana:
> Gru is Minion.

I guess I should have asked more correctly if there's a drawback to not using the gru service the same way it's used in production. And the answer is probably the same :-D

kalikiana (approved):

kraih:
> > Gru is Minion.
> 
> I guess I should have asked more correctly if there's a drawback to not using the gru service the same way it's used in production. And the answer is probably the same :-D

Yea. 😀 It shouldn't matter, the test code still runs a full Minion worker, just without the fork.

PR #4177, opened 2021-09-07
Dependency cron 2021-09-07

openqabot:


.circleci/ci-packages.txt
```diff
@@ -8,7 +8,6 @@ libaspell15-0.60.8
 libcmark0_28_3-0.28.3
 libjq1-1.6
 libonig4-6.7.0
-libopencv3_4-3.4.9
```
kalikiana, line 11: I don't understand why this is being removed - @Martchus since you were looking into the packaging, can you confirm this is expected?

Martchus (approved):
I suppose that's ok considering https://github.com/os-autoinst/openQA/pull/4163#pullrequestreview-745735485.

PR #4179, opened 2021-09-07
Return errors on artefact creation correctly 

Martchus:
* Return a 400/500 error when creating an artefact fails so the worker
  would log the error and retry in case of the 500 error
* We've seen cases in production where result files are missing but no
  final upload error (after exhausing all attempts) is logged, see
  https://progress.opensuse.org/issues/94531#note-35. This change might
  fix those problems.

### DIFF
--- a/lib/OpenQA/WebAPI/Controller/API/V1/Job.pm
+++ b/lib/OpenQA/WebAPI/Controller/API/V1/Job.pm
@@ -563,10 +563,10 @@ sub create_artefact {
 
     if ($self->param('image')) {
         $job->store_image($validation->param('file'), $validation->param('md5'), $self->param('thumb') // 0);
-        return $self->render(text => "OK");
+        return $self->render(text => 'OK');
     }
     elsif ($self->param('extra_test')) {
-        return $self->render(text => "OK")
+        return $self->render(text => 'OK')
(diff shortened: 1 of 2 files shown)


asdil12 (approved):
LGTM

lib/OpenQA/WebAPI/Controller/API/V1/Job.pm
```diff
@@ -602,10 +603,10 @@ sub create_artefact {
             });
     }
     if ($job->create_artefact($validation->param('file'), $self->param('ulog'))) {
-        $self->render(text => "OK");
+        $self->render(text => 'OK');
     }
     else {
-        $self->render(text => "FAILED");
+        $self->render(json => {error => 'Unable to create artefact'}, status => 500);
```
kalikiana, line 609: Should it worry me that existing tests aren't failing? I would think we want to catch at least one occurence of it

Martchus: Since I wasn't sure where to extend tests I've just been creating the PR in the hope the CI checks will tell me. The API tests are actually failing so you shouldn't be worried :-)

kalikiana (approved):

PR #4181, opened 2021-09-08
Trigger cleanup task when jobs are done

kalikiana:
Fixes: https://progress.opensuse.org/issues/97763

### DIFF
--- a/etc/openqa/openqa.ini
+++ b/etc/openqa/openqa.ini
@@ -161,6 +161,7 @@ blocklist = job_grab job_done
 #untracked_assets_storage_duration = 14
 # Performs the cleanup of results/assets only if the free disk space on the relevant partition is below the specified percentage (and aborts early otherwise)
 #result_cleanup_max_free_percentage = 100
+#trigger_cleanup_on_job_done = 0
 #asset_cleanup_max_free_percentage = 100
 # Specify the screenshot ID range to query at once from the database (reduce to avoid big queries, increase to lower query overhead)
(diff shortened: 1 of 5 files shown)


lib/OpenQA/Task/Job/FinalizeResults.pm
```diff
@@ -55,6 +55,7 @@ sub _finalize_results {
         my $ret          = _done_hook_new_issue($openqa_job, $hook, $timeout, $kill_timeout);
         $minion_job->note(hook_cmd => $hook, hook_result => $ret);
     }
+    $app->minion->enqueue(limit_results_and_logs => []);
```
Martchus, line 58: This needs to be configurable.

etc/openqa/openqa.ini
```diff
@@ -161,6 +161,7 @@ blocklist = job_grab job_done
 #untracked_assets_storage_duration = 14
 # Performs the cleanup of results/assets only if the free disk space on the relevant partition is below the specified percentage (and aborts early otherwise)
 #result_cleanup_max_free_percentage = 100
+#trigger_cleanup_on_job_done = 0
```
Martchus, line 164: This should be in the middle of `result_cleanup_max_free_percentage` and `asset_cleanup_max_free_percentage`. It would make sense to put `trigger_cleanup_on_job_done` under its own explaining comment.

Martchus (changes requested):

PR #4183, opened 2021-09-08
Redirect to original page even on automatic login redirect

asdil12:
Ticket: https://progress.opensuse.org/issues/89710

### DIFF
--- a/lib/OpenQA/Shared/Controller/Session.pm
+++ b/lib/OpenQA/Shared/Controller/Session.pm
@@ -21,13 +21,14 @@ use Carp 'croak';
 sub ensure_user {
     my ($self) = @_;
     return 1 if $self->current_user;
-    $self->redirect_to('login');
+    $self->redirect_to($self->url_for('login')->query(return_page => $self->req->url));
     return undef;
 }
 
 sub ensure_operator {
     my ($self) = @_;
-    $self->redirect_to('login')                       and return undef unless $self->current_user;
+    $self->redirect_to($self->url_for('login')->query(return_page => $self->req->url)) and return undef
+      unless $self->current_user;
     $self->render(text => "Forbidden", status => 403) and return undef unless $self->is_operator;
     return 1 if $self->req->method eq 'GET' || $self->valid_csrf;
     $self->render(text => 'Bad CSRF token!', status => 403);
@@ -36,7 +37,8 @@ sub ensure_operator {
 
 sub ensure_admin {
     my ($self) = @_;
-    $self->redirect_to('login')                       and return undef unless $self->current_user;
+    $self->redirect_to($self->url_for('login')->query(return_page => $self->req->url)) and return undef
+      unless $self->current_user;
     $self->render(text => "Forbidden", status => 403) and return undef unless $self->is_admin;
     return 1 if $self->req->method eq 'GET' || $self->valid_csrf;
     $self->render(text => 'Bad CSRF token!', status => 403);
--- a/lib/OpenQA/WebAPI/Auth/OpenID.pm
+++ b/lib/OpenQA/WebAPI/Auth/OpenID.pm
@@ -60,8 +60,8 @@ sub auth_login {
     );
 
     my $return_url = Mojo::URL->new(qq{$url/response});
-    if ($self->req->headers->referrer) {
-        my $return_page = Mojo::URL->new($self->req->headers->referrer)->path_query;
+    if (my $return_page = $self->param('return_page') || $self->req->headers->referrer) {
+        $return_page = Mojo::URL->new($return_page)->path_query;
         # return_page is encoded using base64 (in a version that avoids / and + symbol)
         # as any special characters like / or ? when urlencoded via % symbols,
         # result in a naive_verify_failed_return error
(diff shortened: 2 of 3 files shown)


lib/OpenQA/WebAPI/Auth/OpenID.pm
```diff
@@ -59,9 +59,10 @@ sub auth_login {
         },
     );
 
-    my $return_url = Mojo::URL->new(qq{$url/response});
-    if ($self->req->headers->referrer) {
-        my $return_page = Mojo::URL->new($self->req->headers->referrer)->path_query;
+    my $return_url  = Mojo::URL->new(qq{$url/response});
+    my $return_page = $self->param('return_page') || $self->req->headers->referrer;
+    if ($return_page) {
```
Martchus, line 64:
```suggestion
    if (my $return_page = $self->param('return_page') || $self->req->headers->referrer) {
```

lib/OpenQA/Shared/Controller/Session.pm
```diff
@@ -21,13 +21,14 @@ use Carp 'croak';
 sub ensure_user {
     my ($self) = @_;
     return 1 if $self->current_user;
-    $self->redirect_to('login');
+    $self->redirect_to($self->url_for('login')->query(return_page => $self->req->url));
```
Martchus, line 24: Because only the "page" within the current openQA instance is relevant here I'd strip the other URL parts here right away:
```suggestion
    $self->redirect_to($self->url_for('login')->query(return_page => $self->req->url->path_query));
```

The same counts for the other occurrences. It would actually make sense to extract this into another function to have only one occurrence of this code.

Martchus (changes requested):

lib/OpenQA/Shared/Controller/Session.pm
asdil12: That will already happen in `lib/OpenQA/WebAPI/Auth/OpenID.pm` so I thought it shouldn't be done twice.
Also when doing some tests I found out that it is not needed - it already redirects you using

```
Location: /login?return_page=%2Fadmin%2Fusers
```

Martchus (approved):

kraih:
The lack of tests for this area of the code is really unfortunate. Passing the URL as a parameter has a lot of potential for percent encoding bugs. Did you make sure that requests with query parameters work too? Storing the URL in the session would probably be less risky.

lib/OpenQA/Shared/Controller/Session.pm
kraih, line 24: The exact stringified format of `$c->req->url` depends on how the server is deployed. In some cases it might only contain parts of the path (and a prefix would be stored in `$c->req->url->base`). I've not seen the webapi deployed that way yet though, so it's a theoretical issue. The safe way would be to use `$c->req->url->to_abs`, which guarantees the absolute URL format.

asdil12: I want a relative way anyway as there is no reason for an absolute url as the redirect from `/response` (or was it `/return`?) to the original page can be relative.
In fact `/login` only takes the `path_query`.

asdil12:
> The lack of tests for this area of the code is really unfortunate. Passing the URL as a parameter has a lot of potential for percent encoding bugs. Did you make sure that requests with query parameters work too? Storing the URL in the session would probably be less risky.

As the encoding of the '/' symbols in the path works fine during my manual tests, I have no reason to believe that encoding `?` or `=` should be a problem here. Mojo seems to correctly encode and decode that. I'm strongly against (ab)using the session for this as it would mess with tabbed browsing.

I just performed another manual test by accessing `http://kazhua.qa.suse.de/admin/users?foo=bar` without being logged in and I was redirected back to that very page after login.

asdil12:
> The lack of tests for this area of the code is really unfortunate.

I added a basic test for this.

t/03-auth.t
```diff
@@ -77,6 +77,16 @@ subtest OpenID => sub {
     is($t->get_ok('/response')->status_is(302)->tx->res->headers->location,
         "/tests/42", "redirect to original papge after login");
 
+    is(
+        $t->get_ok('/api_keys')->status_is(302)->tx->res->headers->location,
+        "/login?return_page=%2Fapi_keys",
+        "remember return_page for ensure_operator"
```
kraih, line 83: You can actually use
```perl
$t->get_ok('/api_keys')->status_is(302)->header_is('Location', '/login?return_page=%2Fapi_keys');
```

t/03-auth.t
```diff
@@ -77,6 +77,16 @@ subtest OpenID => sub {
...
+    );
+    is(
+        $t->get_ok('/admin/users')->status_is(302)->tx->res->headers->location,
+        "/login?return_page=%2Fadmin%2Fusers",
+        "remember return_page for ensure_admin"
```
kraih, line 88: Same here.

Martchus (approved):

kalikiana (approved):
Looks sensible to me. I can't say if this is absolutely correct in Mojo terms, but since this is also covered by tests I'm not worried about perfection.

PR #4187, opened 2021-09-10
Add common error message in pitfalls.asciidoc

FruitFly638:
Add the error message `unable to upgrade ws to command server` in the Pitfalls debug developer mode section
Related ticket: https://progress.opensuse.org/issues/97952

### DIFF
--- a/docs/Pitfalls.asciidoc
+++ b/docs/Pitfalls.asciidoc
@@ -82,7 +82,8 @@ load_templates scripts accordingly.
 
 [id="debugdevelmode"]
 == Steps to debug developer mode setup
-This is basically a checklist to go through in case the developer mode is broken in your setup:
+This is basically a checklist to go through in case the developer mode is broken in your setup
+(e.g. you are getting the error message `unable to upgrade ws to command server`):
 
 . Be sure to have everything up to date. That includes relevant packages on the
   machine hosting the web UI and on the worker.


docs/Pitfalls.asciidoc
```diff
@@ -82,7 +82,7 @@ load_templates scripts accordingly.
 
 [id="debugdevelmode"]
 == Steps to debug developer mode setup
-This is basically a checklist to go through in case the developer mode is broken in your setup:
+This is basically a checklist to go through in case the developer mode is broken in your setup (e.g. you are getting the error message `unable to upgrade ws to command server`):
```
okurz, line 85: IMHO the text itself is fine and at the right location. I simply suggest to wrap the lines at 80 characters

okurz (changes requested):
The static check that fails is about the git commit message subject line. Better phrase the git commit message as "Add common error message …" rather than "added common error message …"

Also see https://commit.style/ or http://chris.beams.io/posts/git-commit/ as a helpful guide how to write good commit messages.

docs/Pitfalls.asciidoc
```diff
@@ -82,7 +82,8 @@ load_templates scripts accordingly.
 
 [id="debugdevelmode"]
 == Steps to debug developer mode setup
-This is basically a checklist to go through in case the developer mode is broken in your setup (e.g. you are getting the error message `unable to upgrade ws to command server`):
+This is basically a checklist to go through in case the developer mode is broken in your setup
+(e.g. you are getting the error message `unable to upgrade ws to command server`):
```
FruitFly638, line 86: Is this better?

Martchus (commented):
The 2nd commit should be squashed into the 1st one and the actual commit message should be adjusted (and not just the PR title). You can do that locally and then push your changes using the `-f`-flag.

Btw, you could have avoid the squashing by using `git commit --amend` to incorporate your latest changes into the last commit.

docs/Pitfalls.asciidoc
okurz: Yes, thx

okurz (approved):

Martchus (approved):

PR #4188, opened 2021-09-10
Fetch and display job status for test urls in comments

perlpunk:
Issue: https://progress.opensuse.org/issues/91773

TODO:

~~- [ ] Make it a feature switch (Don't know yet how to do that from client side)~~
- [x] Write a test

I added a new API route `/experimental/jobs/<id>/status` because it is much cheaper than `/jobs/<id>` which fetches also job_dependencies (2 selects) and assets (1 select per asset).
This is up for discussion, but it could be useful for other tools fetching the status of a job regularly.

Edit: I skip the feature switch as discussed in the daily. We don't expect a high performance impact, if any.

### DIFF
--- a/assets/javascripts/test_result.js
+++ b/assets/javascripts/test_result.js
@@ -729,6 +729,33 @@ function renderCommentsTab(response) {
   const tabPanelElement = this.panelElement;
   tabPanelElement.innerHTML = response;
   $(tabPanelElement).find('[data-toggle="popover"]').popover({html: true});
+  // Add job status icons to /t123 urls
+  const hostname = $(location).attr('host');
+  $(tabPanelElement)
+    .find('a')
+    .each(function (index, element) {
+      const href = $(element).attr('href');
+      if (href === undefined) {
+        return;
+      }
+      const re = new RegExp('^https?://' + hostname + '/t([0-9]+)$');
+      const found = href.match(re);
+      if (!found) {
+        return;
+      }
+      const id = found[1];
+      const url = '/api/v1/experimental/jobs/' + id + '/status';
+      $.ajax(url)
+        .done(function (response) {
+          const i = document.createElement('i');
+          const job = response;
+          const stateHTML = testStateHTML(job);
+          i.className = stateHTML[0];
+          i.title = stateHTML[1];
+          element.appendChild(i);
+        })
+        .fail(function (jqXHR, textStatus, errorThrown) {});
+    });
 }
 
 function renderInvestigationTab(response) {
--- a/lib/OpenQA/WebAPI/Controller/API/V1/Job.pm
+++ b/lib/OpenQA/WebAPI/Controller/API/V1/Job.pm
@@ -14,7 +14,7 @@
 # with this program; if not, see <http://www.gnu.org/licenses/>.
 
 package OpenQA::WebAPI::Controller::API::V1::Job;
-use Mojo::Base 'Mojolicious::Controller';
+use Mojo::Base 'Mojolicious::Controller', -signatures;
 
 use OpenQA::Utils qw(:DEFAULT assetdir);
 use OpenQA::JobSettings;
@@ -460,6 +460,25 @@ sub update_status {
         return $self->render(json => {error => $ret->{error}}, status => $ret->{error_status});
     }
     $self->render(json => $ret);
+}
+
+=over 4
+
+=item get_status()
+
+Retrieve status of a job. Returns id, state, result, blocked_by_id.
+Preferrable over /job/<id> for performance and payload size, if you are only
+interested in the status.
+
+=back
+
+=cut
+
+sub get_status ($self) {
+    my $job_id = int $self->stash('jobid');
+    my @fields = qw(id state result blocked_by_id);
+    my $job    = $self->schema->resultset("Jobs")->find($job_id, {select => [@fields]});
+    $self->render(json => {map { $_ => $job->$_ } @fields});
 }
 
 =over 4
--- a/assets/javascripts/openqa.js
+++ b/assets/javascripts/openqa.js
@@ -335,28 +335,36 @@ function renderSearchResults(query, url) {
   request.send();
 }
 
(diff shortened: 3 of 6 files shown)


lib/OpenQA/WebAPI/Controller/API/V1/Job.pm
```diff
@@ -464,6 +464,27 @@ sub update_status {
...
+
+=cut
+
+sub get_status {
+    my ($self)    = @_;
+    my $job_id    = int $self->stash('jobid');
+    my @fields    = qw(id state result blocked_by_id);
+    my $job       = $self->schema->resultset("Jobs")->find($job_id, {select => [@fields]});
+    my $ret       = {map { $_ => $job->$_ } @fields};
+    $self->render(json => $ret);
```
Martchus, line 483:
```suggestion
    $self->render(json => {map { $_ => $job->$_ } @fields});
```

lib/OpenQA/WebAPI/Controller/API/V1/Job.pm
```diff
@@ -464,6 +464,27 @@ sub update_status {
...
+Retrieve status of a job. Returns id, state, result, blocked_by_id.
+Preferrable over /job/<id> for performance and payload size, if you are only
+interested in the status.
+
+=back
+
+=cut
+
+sub get_status {
+    my ($self)    = @_;
```
Martchus, line 478:
```suggestion
sub get_status ($self) {
```

assets/javascripts/test_result.js
```diff
@@ -729,6 +729,32 @@ function renderCommentsTab(response) {
...
+      url = '/api/v1/jobs/' + id + '/status';
+      $.ajax(url)
+        .done(function (response) {
+          const i = document.createElement('i');
+          const job = response;
+          const stateHTML = testStateHTML(job);
+          const className = stateHTML[0];
+          const title = stateHTML[1];
+          i.className = className;
+          i.title = title;
```
Martchus, line 753:
```suggestion
          i.className = stateHTML[0];
          i.title = stateHTML[1];
```

Martchus (commented):
Looks good and introducing another API route seems reasonable (and is more clear compared to adding a `status_only=1` parameter to the existing route).

perlpunk:
Screenshot: urls-with-job-status

Martchus:
About the feature switch: You can "export" variables from the server to the client's JavaScript by rendering a `data`-attribute in a related element and accessing that via JavaScript. That's likely the cleanest way within our current setup. We also sometimes set a variable within the `ready_function` (e.g. https://github.com/os-autoinst/openQA/blob/master/templates/webapi/step/edit.html.ep#L11).

lib/OpenQA/WebAPI/Controller/API/V1/Job.pm
perlpunk: ok. should I add signatures to the whole module then or only this method?

assets/javascripts/test_result.js
perlpunk: done

lib/OpenQA/WebAPI/Controller/API/V1/Job.pm
perlpunk: done

lib/OpenQA/WebAPI/Controller/API/V1/Job.pm
Martchus: I just like to use them for new functions from the start. Of course you could convert the whole module if you like but that should then be a separate commit.

perlpunk:
I added a test

okurz (approved):
This looks great! IMHO you can mark the PR as ready for review unless you have specific points you want to add first

assets/javascripts/test_result.js
```diff
@@ -729,6 +729,33 @@ function renderCommentsTab(response) {
...
+        return;
+      }
+      const re = new RegExp('^https?://' + hostname + '/t([0-9]+)$');
+      const found = href.match(re);
+      if (!found) {
+        return;
+      }
+      const id = found[1];
+      const url = '/api/v1/jobs/' + id + '/status';
+      $.ajax(url)
```
Martchus, line 748: I suppose it would make sense if there was just one request (for all IDs), e.g. we could allow a list of comma-separated IDs here and return the results in an array.

Martchus (commented):
> IMHO you can mark the PR as ready for review unless you have specific points you want to add first

I think the feature switch is still missing. Not sure whether we actually need it, though.

assets/javascripts/test_result.js
perlpunk: True. The use case this is for will usually have 1-4 such links. We might improve this later and can monitor the Apache response times for now.

okurz (approved):

assets/javascripts/test_result.js
Martchus: Ok. We could allow calling it like `/api/v1/jobs/1,2,3/status` in the future which would still be backwards compatible. However, the response format should better be already changes so we don't have to distinguish the one ID and multiple IDs cases later. Or you just move the route under some experimental prefix for now to we can break compatibility later (e.g. we have already `/api/v1/experimental/search`).

Martchus (changes requested):
As stated in https://github.com/os-autoinst/openQA/pull/4188#discussion_r707221909 I would at least change the API route so we can change it later to support multiple job IDs.

assets/javascripts/test_result.js
perlpunk: Hmm, I'm not really a fan of `/api/v1/jobs/1,2,3/status`.
I vote for going experimental.

perlpunk: done

Martchus (approved):

perlpunk:
Needed to force push because I was following the circleci project and tests did not start

PR #4189, opened 2021-09-10
Speed up scan_for_untracked_assets by using less syscalls

kraih:
This patch gets rid of almost all `stat` syscalls when scanning for untracked files during asset cleanup. I did a benchmark with 60k files on a very very fast SSD, and the result was a performance gain of about 25%.

Progress: https://progress.opensuse.org/issues/97979

### DIFF
--- a/lib/OpenQA/Schema/ResultSet/Assets.pm
+++ b/lib/OpenQA/Schema/ResultSet/Assets.pm
@@ -14,12 +14,7 @@
 # with this program; if not, see <http://www.gnu.org/licenses/>.
 
 package OpenQA::Schema::ResultSet::Assets;
-
-use strict;
-use warnings;
-
-use Mojo::Base -strict, -signatures;
-use base 'DBIx::Class::ResultSet';
+use Mojo::Base 'DBIx::Class::ResultSet', -signatures;
 
 use DBIx::Class::Timestamps 'now';
 use OpenQA::Log qw(log_info log_debug log_warning);
@@ -57,8 +52,7 @@ sub register ($self, $type, $name, $options = {}) {
         });
 }
 
-sub scan_for_untracked_assets {
-    my ($self) = @_;
+sub scan_for_untracked_assets ($self) {
 
     # search for new assets and register them
     my $assetdir = assetdir();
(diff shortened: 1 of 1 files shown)


lib/OpenQA/Schema/ResultSet/Assets.pm
```diff
@@ -57,48 +52,44 @@ sub register ($self, $type, $name, $options = {}) {
...
+            my $name = basename($path);
+            next if $known{$name};
 
-            my $basepath = basename($path);
             # ignore links
             next if -l $path;
 
             # ignore files not owned by us
-            next unless -o $path;
+            next unless -o _;
```
Martchus, line 82: I assume `_` refers to the last internal `stat`. Maybe it would make sense to add an explaining comment nevertheless because the syntax is quite "magic".

It could also be just one line:
```
# ignore links, files not owned by us and non-existing files and folders
# note: `_` refers to the result of the last internal `stat`.
next if -l $path || !-o _ || !-e _;
```

Martchus (commented):
Looks good, I had only one remark.

lib/OpenQA/Schema/ResultSet/Assets.pm
kraih: Using the `_` filehandle is not really magic. It's pretty standard Perl for reusing the result of the previous stat call for `-...` operators and explained in `perldoc -f -e`.

Martchus: I mean not self-explaining and not easy to google for people who don't know Perl very well.

kraih: The idea of explaining Perl features that are in the Perl documentation with comments bugs me a little. Which other Perl features would get the same treatment? Where would we draw the line? Let's hear what others on the team think.

okurz: Maybe reference that part of the documentation without repeating the content? Is it https://perldoc.perl.org/functions/stat that mentions "If stat is passed the special filehandle consisting of an underline, no stat is done, but the current contents of the stat structure from the last stat, lstat, or filetest are returned"?

I like the suggestion of @Martchus to use

```perl
# ignore links, files not owned by us and non-existing files and folders
# note: `_` refers to the result of the last internal `stat`.
next if -l $path || !-o _ || !-e _;
```

lib/OpenQA/Schema/ResultSet/Assets.pm
```diff
@@ -61,44 +61,42 @@ sub scan_for_untracked_assets {
...
+            next unless -e _;
+
+            $paths{basename($path)} = 0;
         }
-        my $assets = $self->search({type => $type});
-        while (my $as = $assets->next) {
+
+        for my $as ($self->search({type => $type})->all) {
             $paths{$as->name} = $as->id;
         }
```
okurz: would it be possible to somehow combine the loops further, e.g.

```patch
- ....
-        for my $as ($self->search({type => $type})->all) {
-            $paths{$as->name} = $as->id;
-        }
+        my %empty_paths = map {basename($_) => 0 } grep {!-l $path && -o _ && -e _} @paths;
+        my %named_paths = map { $_-> name => $_->id } $self->search({type => $type})->all;
+        my %paths = (%empty_paths , %named_paths);
```

?

okurz:
I created https://github.com/os-autoinst/openQA/pull/4191 to make the coverage report easier to read as only a single line is currently uncovered

lib/OpenQA/Schema/ResultSet/Assets.pm
kraih: See https://perldoc.perl.org/functions/-e

mimi1vx: I think better is:
```
# ignore links, files not owned by us and non-existing files and folders
next if  -l -e -o $path;
```

kraih: > I think better is:
> 
> ```perl
> # ignore links, files not owned by us and non-existing files and folders
> next if  -l -e -o $path;
> ```

The logic is wrong.

kraih: Anyway, i think i've addressed all  concerns in the latest revision.

Martchus (approved):

okurz (approved):

PR #4194, opened 2021-09-13
Install os-autoinst-distri-opensuse-deps in worker container

asdil12:
Ticket: https://progress.opensuse.org/issues/97751

### DIFF
--- a/container/worker/run_openqa_worker.sh
+++ b/container/worker/run_openqa_worker.sh
@@ -24,4 +24,7 @@ fi
 qemu-system-x86_64 -S &
 kill $!
 
+# Install test distribution dependencies
+find "/var/lib/openqa/share/tests" -maxdepth 2 -type f -executable -name 'install_deps.*' -exec {} \;
+
 su _openqa-worker -c "/usr/share/openqa/script/worker --verbose --instance \"$OPENQA_WORKER_INSTANCE\""


Martchus (approved):
I also noticed that problem when testing https://github.com/os-autoinst/openQA/pull/4160.

It doesn't seem to be the right approach to tailor our container to a specific test distribution and rather provide a way (e.g. instructions in the documentation) for installing additional dependencies as needed. Obviously, we couldn't accept such a PR for any possible test distribution and the arbitrary dependencies they might need.

However, for the sake of providing a good out-of-the-box experience for openSUSE tests I'm approving the PR nevertheless. There's no harm done for other test distributions.

perlpunk (approved):

okurz (changes requested):
Absolutely not. This is not a "openSUSE openQA worker"

asdil12:
New approach:

The test distribution will get a script into the git root that will install the dependencies. The container entrypoint script will call such a script in the test distribution folder if present.

See also: https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/13244

container/worker/run_openqa_worker.sh
```diff
@@ -24,4 +24,14 @@ fi
 qemu-system-x86_64 -S &
 kill $!
 
+if [[ -e /var/lib/openqa/share/tests ]] ; then
+  pushd /var/lib/openqa/share/tests
+    for dir in $(ls) ; do
```
okurz, line 29: `$(ls)` should be equivalent to `*`, isn't it?

container/worker/run_openqa_worker.sh
```diff
@@ -24,4 +24,14 @@ fi
...
+      pushd "$dir"
+      test -x install_deps.sh && ./install_deps.sh
+      popd
+    done
+  popd
+fi
```
Martchus, line 35: The code is bad for multiple reasons, e.g. it breaks if white spaces are used and produces noisy `pushd`/`popd` output. I'd just use `find`:
```suggestion
find '/hdd/openqa-devel/openqa/share/tests' -maxdepth 2 -type f -executable -name 'install_deps.sh' -exec {} \;
```

Martchus (changes requested):
I suppose this is acceptable although I must say that downloading/installing packages every time when starting a worker is not ideal. I'm only requesting changes because the code should be simplifies (see inline comment).

container/worker/run_openqa_worker.sh
Martchus: `*` would be better because it doesn't break when using white spaces. However, I'd prefer using `find`.

Martchus (approved):

okurz (approved):

PR #4197, opened 2021-09-13
fix broken syntax in logrotate config

phil-hands:
Given that this was introduced in 2014 ( 56dc7e7c441f0cf6f7517959b3eac99901b6ba89 ) I initially assumed that SUSE's logrotate must support the use of `=` but it was causing Debian's logrotate to throw an error.

I found a copy of a [SuSE manpage](https://manpage.me/index.cgi?apropos=0&q=logrotate&sektion=0&manpath=SuSE+Linux%2Fi386+ES+10+SP1&arch=default&format=html) showing an equals sign in use. but if one looks down to the `size` entry, one sees only a space in use:
>        size size
>               Log files are rotated when they grow bigger then size bytes. If
>               size is followed by M, the size if assumed to be in megabytes.
>               If the k is used, the size is in kilobytes. So size 100, size
>               100k, and size 100M are all valid.

Looking for `size=100k` in the upstream, one finds it [being removed in 2008](https://github.com/logrotate/logrotate/commit/82214b44f0a0e7eade089b22d1556262eb875216#diff-722930fb2346414c4bfdcf6a8257563337c7e8a97e2a03ef26125caf76bd239aL90-R92) so perhaps it never was supported after all (in which case the SUSE man page probably ought to be checked)

Anyway, Debian's logrotate is not impressed with the `=` so hopefully this version will work everywhere (not that I'm actually expecting logrotate to be used with our package very often, since it'll use the journal by default, but the presence of the file caused a QA test to fail).

HTH

P.S. I note that the file has the x-bit set, which seems wrong to me, so I've added a commit to fix that too,


okurz (changes requested):
This certainly does look better. Thanks a lot! May I ask you for just one favor: Please start each git commit message subject line with an uppercase letter for consistency.

phil-hands:
done.

perlpunk (approved):

okurz (approved):

PR #4199, opened 2021-09-14
Additions to apparmor profiles

phil-hands:
Here are some additions that I've made to the apparmor config for the Debian package.

I've tried to ensure that I've only added things, so it ought to still be OK for openSUSE systems too.

Please take a look to see which of this seems useful to you, and I'll get rid of any changes you'd prefer not to have upstream.

Here are the descriptions I added to the commit as I made the changes:

- Debian uses dash as its default shell
- Debian has it's git components under /usr/lib/git-core/
- /usr/bin/git is being called directly when scm=git
- unzip is called just 'unzip' on Debian
- git takes a look at /etc/mailname for creating commit messages
  (perhaps it wouldn't if the Author were explicitly set, so maybe this
  is not really needed?)
- git-remote-{http*,rm} get used for pushing changes to needles,
  if enabled
- kvm uses seabios to boot
- needles need to be writable to enable editing
- pwd is in /bin on Debian
- git wants to look at /etc/gitconfig
- need to access OVMF files for UEFI booting VMs


Martchus (commented):
I guess it is ok to reference non-existing files/directories so I'd be fine with that. Covering more distributions here shouldn't harm distributions which are already covered.

phil-hands:
oh, I put the "WIP: " in there thinking that would mark this as a draft, on the basis that you might want not to take the complete patch for whatever reason, but I guess WIP is a gitlab-ism, so never mind -- I've removed it to make sure it doesn't end up in the commit history.

Feel free to ask for changes anyway :-)

Martchus (approved):

okurz (approved):
> oh, I put the "WIP: " in there thinking that would mark this as a draft, on the basis that you might want not to take the complete patch for whatever reason, but I guess WIP is a gitlab-ism, so never mind -- I've removed it to make sure it doesn't end up in the commit history.

We still have an automatic check with the "WIP" github service that prevents merging when WIP is included though in gitlab the support is more streamlined.

profiles/apparmor.d/usr.share.openqa.script.worker
```diff
@@ -145,6 +150,8 @@
   /var/lib/openqa/share/factory/tmp/ rw,
   /var/lib/openqa/share/factory/tmp/** rw,
   /var/lib/openqa/share/tests/** r,
+  /var/lib/openqa/share/tests/*/needles/** rw,
+  /var/lib/openqa/share/tests/*/profiles/*/needles/** rw,
```
marmarek, line 154: Is it worker process that writes them? I'm pretty sure it worked for me before - without this change.

okurz: No, the worker should not write anything in share. Wer have many workers setup mounting share as readonly over NFS

phil-hands: Yes, I agree -- I guess I pasted that into the wrong file at some point when I meant to put it in the openqa one to allow needle editing via the webUI. Sorry about that.

 Should I revert that bit of the patch as a new PR?

okurz: I would appreciate that, yes

PR #4200, opened 2021-09-14
Stop upsetting pod2man

phil-hands:
This fixes a few complaints by pod2man about missing =back's and =items outside an =over etc.

### DIFF
--- a/lib/OpenQA/CLI.pm
+++ b/lib/OpenQA/CLI.pm
@@ -27,6 +27,10 @@ has namespaces => sub { ['OpenQA::CLI'] };
 
 =encoding utf8
 
+=head1 NAME
+
+openqa-cli - provides command-line access to the openQA API
+
 =head1 SYNOPSIS
 
   Usage: openqa-cli COMMAND [OPTIONS]
--- a/script/client
+++ b/script/client
@@ -15,9 +15,15 @@
 # You should have received a copy of the GNU General Public License along
 # with this program; if not, see <http://www.gnu.org/licenses/>.
 
+=head1 NAME
+
+openqa-client - deprecated script for talking to an openQA server
+
 =head1 SYNOPSIS
 
-client [OPTIONS] PATH
+This script is deprecated, use openqa-cli instead
+
+openqa-client [OPTIONS] PATH
 
 =head1 OPTIONS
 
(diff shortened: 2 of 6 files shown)


Martchus (approved):
I trust you've checked that. To me this POD syntax always just looks wrong :-)

phil-hands:
yes, I have :-)

Without this, pod2man (from perl v5.32.1) produces error messages, whereas with the patch it's cheerful.

I've added a patch adding/changing to having a NAME section in each of the pod files, which results in there being a NAME section in the pod2man output (which if one looks at the man(7) page one can see is the only mandatory section).

Of course, it's possible that you don't want this patch, because you never use pod2man, and were relying on the sections being named as they were for something else, but it doesn't really look like that's the case.

phil-hands:
Oh, and I should point out that I added a note to the `client` script, saying it's deprecated in favour of `openqa-cli` (which I believe to be the case).

I also notice that a (possibly) Debian-specific change has crept through, which is that we call that script `openqa-client` -- not that that matters for us much, as I'll not be linking it into /usr/bin to avoid people thinking it's the thing to use, but you may prefer me to change that back to just `client`

kalikiana (approved):
Looks very nice! I wonder if there's a way to check the format in a test 🤔 No idea if that's trivial to add, though, and can be done in a follow-up.

phil-hands:
> Looks very nice!

Glad you like it.

> I wonder if there's a way to check the format in a test

for stuff like the missing `=back` you just need to run pod2man and check for a non-zero exit.

I think the missing NAME sections were picked up by running groff, and looking for output on STDERR, but I am momentarily unable to reproduce that, so cannot immediately tell you the exact command (it was something that was highlighted by `lintian`, debian's package checker, and then I reproduced the error on the command line).

Anyway, I've got another PR on the way that adds a man page, which if accepted it will probably involve someone deciding where to put it, and adding a Makefile clause to install it somewhere, at which point I'll have somewhere to promote the manpage generation/installation bits from `debian/rules` into your `Makefile`, and that might well be the place to add a test too.

okurz:
@phil-hands happy to receive your contributions. I think we should add this check calling pod2man. If you like you can add an according rule to the Makefile. You can also submit Debian specifics. We have some distribution specific data for the openSUSE package in https://github.com/os-autoinst/openQA/tree/master/dist

PR #4202, opened 2021-09-14
Use DBIx method 'find' where applicable rather than search->first

okurz:


lib/OpenQA/Schema/ResultSet/DeveloperSessions.pm
```diff
@@ -34,7 +34,7 @@ sub register {
     my ($result, $worker_id, $is_session_already_existing) = $schema->txn_do(
         sub {
             # refuse if no worker assigned
-            my $worker = $schema->resultset('Workers')->search({job_id => $job_id})->first;
+            my $worker = $schema->resultset('Workers')->first({job_id => $job_id});
```
Amrysliu, line 37: I guess this should be `find`.

Amrysliu:
if the find condition is not primary key or unique constraint, does method `find` work well? I didn't try it.
https://metacpan.org/pod/DBIx::Class::ResultSet#find

docs/ExternalResults.asciidoc
```diff
@@ -96,7 +96,7 @@ $parser_result->tests->each(sub {
 
 my @all = $parser_result->tests->each;
 my @tests = $parser->tests->search(name => qr/1_running_upstream_tests/);
-my $first = $parser->tests->search(name => qr/1_running_upstream_tests/)->first();
+my $first = $parser->tests->find(name => qr/1_running_upstream_tests/);
```
Martchus, line 99: This is an `OpenQA::Parser::Result::OpenQA::Results` so this change is wrong.

okurz:
> if the find condition is not primary key or unique constraint, does method `find` work well? I didn't try it.
> https://metacpan.org/pod/DBIx::Class::ResultSet#find

I guess I don't have enough database experience to properly answer that but maybe the failing tests tell us :)

lib/OpenQA/Schema/ResultSet/Jobs.pm
```diff
@@ -477,7 +477,7 @@ sub mark_job_linked {
     return undef if !$job || ($referer->path_query =~ /^\/?$/);
     my $comments = $job->comments;
     return undef if $comments->find({text => {like => 'label:linked%'}});
-    my $user = $self->result_source->schema->resultset('Users')->search({username => 'system'})->first;
+    my $user = $self->result_source->schema->resultset('Users')->find({username => 'system'});
```
Martchus, line 480: There's no unique constraint (only on a combination of `username` and `provider`) so this should be limited to one row.

lib/OpenQA/WebAPI/Controller/Step.pm
```diff
@@ -31,7 +31,7 @@ sub _init {
     my ($self) = @_;
 
     return 0 unless my $job = $self->app->schema->resultset('Jobs')->find($self->param('testid'));
-    my $module = $job->modules->search({name => $self->param('moduleid')})->first;
+    my $module = $job->modules->find({name => $self->param('moduleid')});
```
Martchus, line 34: There's no unique constraint (only on a combination with other columns) so this should be limited to one row.

t/06-users.t
```diff
@@ -35,7 +35,7 @@ subtest 'new users are not ops and admins' => sub {
 };
 
 subtest 'system user presence' => sub {
-    my $system_user = $t->app->schema->resultset("Users")->search({username => 'system'})->first;
+    my $system_user = $t->app->schema->resultset("Users")->find({username => 'system'});
```
Martchus, line 38: There's no unique constraint (only on a combination with other columns) so this should be limited to one row.

t/30-test_parser.t
```diff
@@ -458,7 +458,7 @@ sub test_xunit_file {
       };
     is $parser->tests->search("name", qr/bacon/)->size, 1;
 
-    is $parser->tests->search("name", qr/bacon/)->first()->name, 'bacon';
+    is $parser->tests->find("name", qr/bacon/)->name,                    'bacon';
```
Martchus, line 461: This is an `OpenQA::Parser::Result::OpenQA::Results` so this change is wrong.

t/30-test_parser.t
```diff
@@ -374,7 +374,7 @@ sub test_junit_file {
         'script'   => undef
       };
 
-    is $parser->tests->search("name", qr/1_running_upstream_tests/)->first()->name, '1_running_upstream_tests';
+    is $parser->tests->find("name", qr/1_running_upstream_tests/)->name,   '1_running_upstream_tests';
```
Martchus, line 377: This is an `OpenQA::Parser::Result::OpenQA::Results` so this change is wrong.

t/api/04-jobs.t
```diff
@@ -914,7 +914,7 @@ subtest 'handle settings when posting job' => sub {
 };
 
 subtest 'do not re-generate settings when cloning job' => sub {
-    my $job_settings = $jobs->search({test => 'autoupgrade'})->first->settings_hash;
+    my $job_settings = $jobs->find({test => 'autoupgrade'})->settings_hash;
```
Martchus, line 917: There's no unique constraint (only on a combination with other columns) so this should be limited to one row.

Martchus (changes requested):

docs/ExternalResults.asciidoc
okurz: I don't understand. Why is it wrong?

lib/OpenQA/Schema/ResultSet/Jobs.pm
okurz: What do you mean with "this should be limited to one row"?

docs/ExternalResults.asciidoc
Martchus: `$parser->tests` returns a `OpenQA::Parser::Result::OpenQA::Results` object which has nothing to do with DBIx and only has a `search` function.

lib/OpenQA/Schema/ResultSet/Jobs.pm
Martchus: You cannot do queries returning more than one row with `find` (see also @Amrysliu's comment). One of the lines you've changed already limits the result to one row: `$last_job_id = $schema->resultset('Jobs')->find(undef, {rows => 1, order_by => {-desc => 'id'}})->id;`

okurz: > You cannot do queries returning more than one row with find

Well, actually one can. https://metacpan.org/pod/DBIx::Class::ResultSet#find says "If no such constraint is found, find currently defaults to a simple search->(\%column_values) which may or may not do what you expect. Note that this fallback behavior may be deprecated in further versions. If you need to search with arbitrary conditions - use "search". If the query resulting from this fallback produces more than one row, a warning to the effect is issued, though only the first row is constructed and returned as $result_object."

so in short: The first match will be returned and a warning is printed.

However, I thought we would only apply search+first when we already know there is only one match. I consider it risky to just take the "first" match when we don't even expected others to be there. I wonder if it wouldn't be better to make "find" fail hard if there are more than one matches unless a force parameter is given or something

lib/OpenQA/WebAPI/Controller/Step.pm
okurz: What do you mean with "this should be limited to one row"? You want me to revert the change or did you misunderstand how "find" behaves when multiple matching rows are found? Because only the first row is returned though a warning is printed, see https://github.com/os-autoinst/openQA/pull/4202#discussion_r711798111 .

In this specific case, why are we ok to just pick the first "moduleid" match?

lib/OpenQA/Schema/ResultSet/Jobs.pm
Martchus: > so in short: The first match will be returned and a warning is printed.

I know, and a warning is not acceptable so I said you cannot use `find` with more than one row.

Likely search+fist was used in context where it didn't matter which row was returned exactly or where it was an accepted limitation that you can only get the first.

In this particular case we want the system user so we should include its provider into the query. I suppose the provider of the system user is just an empty string.

lib/OpenQA/WebAPI/Controller/Step.pm
Martchus: Because in most tests the test module names are unique. However, if there's a test module with the same name in multiple categories this function wouldn't work very well. I have *not* designed this. To me it would make more sense if one would specify all the criteria to identify a test module here (category, name, script) or to actually have a unique constraint for only the name (or having to specify category/name and have a unique constraint only for both).

I would not clutter our logs with warnings about this. If you think this design flaw is worth improving, create a ticket. For the sake of this PR better just use query options like `{rows => 1, order_by => {-desc => 'id'}}` to ensure we get only one row.

t/06-users.t
Martchus: The same counts as for the other occurrence when the system user is wanted. Maybe this could be a function in the `ResultSet/Users.pm`.

t/api/04-jobs.t
Martchus: Here you could specify category and script to make it unique, or just add `{rows => 1, order_by => {-desc => 'id'}}` because simply getting the first makes sense here as well.

Martchus (changes requested):

lib/OpenQA/Schema/ResultSet/Jobs.pm
okurz: I changed the query to look for `provider => ''` additionally now

lib/OpenQA/WebAPI/Controller/Step.pm
okurz: Should I use

```perl
search({name => $self->param('moduleid')}, {rows => 1, order_by => {-desc => 'id'}})->first;
```

https://metacpan.org/pod/DBIx::Class::ResultSet#find states "If no key is specified, the search is carried over all unique constraints which are fully defined by the available condition. If no such constraint is found, find currently defaults to a simple search->(\%column_values) […]. Note that this fallback behavior may be deprecated in further versions". So it would work and I consider it unlikely that DBIx will change that but we can be on the save side with search, right?

t/api/04-jobs.t
```diff
@@ -904,7 +904,8 @@ subtest 'handle settings when posting job' => sub {
 };
 
 subtest 'do not re-generate settings when cloning job' => sub {
-    my $job_settings = $jobs->search({test => 'autoupgrade'})->first->settings_hash;
+    my %attrs = {rows => 1, order_by => {-desc => 'id'}};
```
Martchus, line 907:
```suggestion
    my %attrs = (rows => 1, order_by => {-desc => 'id'});
```

Martchus (changes requested):

lib/OpenQA/Schema/ResultSet/DeveloperSessions.pm
```diff
@@ -22,7 +22,7 @@ sub register {
...
```
perlpunk, line 25: Hm, the documentation says that `first` doesn't expect any arguments
https://metacpan.org/pod/DBIx::Class::ResultSet#first

lib/OpenQA/Schema/ResultSet/Users.pm
```diff
@@ -22,4 +22,6 @@ sub create_user ($self, $id, %attrs) {
     return $user;
 }
 
+sub system ($self) { $self->search({username => 'system', provider => ''}, {rows => 1})->first }
```
perlpunk, line 25: Since there is a unique constraint on username + provider, likely `single` will be enough.
```
$self->single({username => 'system', provider => ''})
```

lib/OpenQA/WebAPI/Controller/API/V1/Job.pm
```diff
@@ -291,7 +291,7 @@ sub show {
     my $self = shift;
     my $job_id = int($self->stash('jobid'));
     my $details = $self->stash('details') || 0;
-    my $job = $self->schema->resultset("Jobs")->search({'me.id' => $job_id}, {prefetch => 'settings'})->first;
+    my $job = $self->schema->resultset("Jobs")->find({'me.id' => $job_id}, {prefetch => 'settings'});
```
perlpunk, line 294: Does this work:
```
my $job = $self->schema->resultset("Jobs")->find($job_id, {prefetch => 'settings'});
```
?

lib/OpenQA/WebAPI/Controller/Test.pm
```diff
@@ -131,27 +130,24 @@ sub list_running_ajax {
...
-                flavor => $job->FLAVOR // '',
-                arch => $job->ARCH // '',
-                build => $job->BUILD // '',
-                testtime => ($job->t_started // '') . 'Z',
-                group => $job->group_id,
-                state => $job->state,
-                progress => $job->progress_info,
-            });
-    }
+    my @running = map {
```
perlpunk, line 133: I don't really see the advantage, and downside is that all DBIC objects are loaded into memory at once.

t/42-df-based-cleanup.t
```diff
@@ -164,7 +164,7 @@ subtest 'unable to make enough room; important job scheduled during the cleanup
     is $job->{result}, 'Unable to cleanup enough results', 'unable to make enough room';
 };
 
-my $new_job_id = $jobs->search({}, {order_by => {-desc => 'id'}})->first->id;
+my $new_job_id = $jobs->find({}, {order_by => {-desc => 'id'}})->id;
```
perlpunk, line 167: `find` without any search attributes?

t/42-screenshots.t
```diff
@@ -69,8 +69,8 @@ is_deeply(
 
 # limit screenshots
 my %args = (
-    min_screenshot_id => $screenshots->search(undef, {rows => 1, order_by => {-asc => 'id'}})->first->id,
-    max_screenshot_id => $screenshots->search(undef, {rows => 1, order_by => {-desc => 'id'}})->first->id,
+    min_screenshot_id => $screenshots->first(undef, {rows => 1, order_by => {-asc => 'id'}})->id,
```
perlpunk, line 72: `first` doesn't expect arguments.

t/ui/26-jobs_restart.t
```diff
@@ -74,7 +74,7 @@ prepare_database;
 
 my $last_job_id;
 sub update_last_job_id {
-    $last_job_id = $schema->resultset('Jobs')->search(undef, {rows => 1, order_by => {-desc => 'id'}})->first->id;
+    $last_job_id = $schema->resultset('Jobs')->find(undef, {rows => 1, order_by => {-desc => 'id'}})->id;
```
perlpunk, line 77: See my previous comment

lib/OpenQA/Schema/ResultSet/DeveloperSessions.pm
okurz: Right. That should be `find` here. Thx

lib/OpenQA/Schema/ResultSet/Users.pm
okurz: So does that mean that

```perl
$self->single({username => 'system', provider => ''})
```

is equivalent to

```perl
$self->search({username => 'system', provider => ''}, {rows => 1})->first
```

and the same in all the other cases as well?

perlpunk: Only if the arguments are a unique constraint. It doesn't add `rows => 1` for example, so the result must return 0 or 1 rows.

I was surprised that you didn't use `find` here.

perlpunk: quote from the docs:
> Note                                                                            
> As of 0.08100, this method enforces the assumption that the preceding query **returns only one row**. If more than one row is returned, **you will receive a warning**:
>
> `Query returned more than one row ` 
> In this case, you should be using "next" or "find" instead, or if you really know what you are doing, use the "rows" attribute to explicitly limit the size of the resultset.

t/42-df-based-cleanup.t
Martchus: This one should definitely be limited to 1 row (as the intend it to get only the "newest" job here).

okurz:
Multiple occurences which I originally changed in this PR have since been done by @perlpunk in https://github.com/os-autoinst/openQA/pull/4556 . I now rebased this PR on top

lib/OpenQA/WebAPI/Controller/Step.pm
```diff
@@ -19,7 +19,8 @@ sub _init {
     my ($self) = @_;
 
     return 0 unless my $job = $self->app->schema->resultset('Jobs')->find($self->param('testid'));
-    my $module = $job->modules->search({name => $self->param('moduleid')})->first;
+    my %attrs = (rows => 1, order_by => {-desc => 'id'});
+    my $module = $job->modules->search({name => $self->param('moduleid')}, \%attrs)->first;
```
Martchus, line 23: Since this is limited to one row I suppose one could also use `find` here.

perlpunk: I looked at this in my PR, too, but the `name` is not a primary or unique key, so theoretically this could return more than one row and issue a warning.

Martchus: But it specifies `rows => 1` as attributes.

lib/OpenQA/Schema/Result/Jobs.pm
```diff
@@ -1903,7 +1903,7 @@ sub _job_stop_cluster {
     return 0 if $job == $self->id;
     my $rset = $self->result_source->resultset;
 
-    $job = $rset->search({id => $job, result => NONE})->first;
+    $job = $rset->find({id => $job, result => NONE});
```
perlpunk, line 1906: I tried this in my PR, but it doesn't work. `find` wants primary/unique keys. any additional search fields will be ignored.

okurz: but @Martchus in https://github.com/os-autoinst/openQA/pull/4202#discussion_r828210781 said we could use find with the `rows => 1` attribute. So what is correct now?

perlpunk: Well, try it out with `DBIC_TRACE=1` and see `WHERE ( me.id = ? )` without mentioning `result`. That's obviously wrong, isn't it?

This is about the jobs table, which has a primary key `id`. *Additional* columns in `find()` will be ignored. (and yes, I think that should die or at least warn instead of silently ignore it.)

https://github.com/os-autoinst/openQA/pull/4202#discussion_r828210781is about the `job_modules` table, where `name` is *part* of a unique key constraint.
So these are two different situations.

Tests are failing, and I haven't had a closer look, but this line here could be the problem.

t/06-users.t
```diff
@@ -23,7 +23,7 @@ subtest 'new users are not ops and admins' => sub {
 };
 
 subtest 'system user presence' => sub {
-    my $system_user = $t->app->schema->resultset("Users")->search({username => 'system'})->first;
+    my $system_user = $t->app->schema->resultset("Users")->system;
```
perlpunk, line 26: that `system` method returns a `Users` object with only the `id` attribute set (because of the select). That's for performance.

The `is_admin` and `is_operator` below check won't work as expected because of this. They might be passing coincidentally.

t/17-build_tagging.t
```diff
@@ -299,7 +299,7 @@ Then "important builds" are skipped from cleanup
 =cut
 subtest 'no cleanup of important builds' => sub {
     # build 0048 has already been tagged as important before
-    my $job = $jobs->search({id => 99938, state => 'done', group_id => 1001, BUILD => '0048'})->first;
+    my $job = $jobs->find({id => 99938, state => 'done', group_id => 1001, BUILD => '0048'});
```
perlpunk, line 302: Maybe the test passes, but only coincidentally. See the other discussion. I will only do `WHERE (me.id = ? )`

t/42-df-based-cleanup.t
```diff
@@ -24,7 +24,7 @@ my $t = Test::Mojo->new('OpenQA::WebAPI');
 my $app = $t->app;
 my $job_groups = $schema->resultset('JobGroups');
 my $jobs = $schema->resultset('Jobs');
-my $user = $schema->resultset('Users')->search({})->first;
+my $user = $schema->resultset('Users')->system;
```
perlpunk, line 27: Possibly the same problem as above.

t/06-users.t
okurz: I see. I changed the "system" method to take an optional attribute hash and only supply that where applicable.

t/42-df-based-cleanup.t
```diff
@@ -164,7 +164,7 @@ subtest 'unable to make enough room; important job scheduled during the cleanup
     is $job->{result}, 'Unable to cleanup enough results', 'unable to make enough room';
 };
 
-my $new_job_id = $jobs->search({}, {order_by => {-desc => 'id'}})->first->id;
+my $new_job_id = $jobs->find({}, {rows => 1, order_by => {-desc => 'id'}})->id;
```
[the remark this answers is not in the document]
okurz: My goal is not to save characters but to make our intention in the code more clear. If it's not more clear then I agree with you that the according code should not be changed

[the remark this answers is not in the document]
perlpunk: To me a `find` here is not clearer. However, the `rows => 1` should certainly be added.
It's just in a test, but we should train ourselves to limit the returned rows whenever we only intend to fetch one.

lib/OpenQA/Schema/ResultSet/Users.pm
```diff
@@ -21,4 +21,6 @@ sub create_user ($self, $id, %attrs) {
     return $user;
 }
 
+sub system ($self, %attrs) { $self->find({username => 'system', provider => ''}, \%attrs) }
```
perlpunk, line 24:
```suggestion
sub system ($self, $attrs) { $self->find({username => 'system', provider => ''}, $attrs) }
```

lib/OpenQA/Schema/Result/Jobs.pm
```diff
@@ -1903,7 +1903,7 @@ sub _job_stop_cluster {
     return 0 if $job == $self->id;
     my $rset = $self->result_source->resultset;
 
-    $job = $rset->search({id => $job, result => NONE})->first;
+    $job = $rset->find({id => $job, result => NONE}, {rows => 1});
```
perlpunk, line 1906: Adding `rows => 1` won't change that `result` is never added to the `WHERE` clause :)

lib/OpenQA/WebAPI/Controller/API/V1/Job.pm
okurz: yes, doen

lib/OpenQA/WebAPI/Controller/Test.pm
okurz: excluded from this PR for now

lib/OpenQA/WebAPI/Controller/Test.pm
```diff
@@ -385,15 +373,8 @@ sub infopanel {
...
-
     return $self->reply->not_found if (!defined $self->param('testid'));
-
-    my $job = $self->schema->resultset("Jobs")->search(
-        {
-            id => $self->param('testid')
-        },
-        {prefetch => qw(jobs_assets)})->first;
-    return $job;
+    return $self->schema->resultset("Jobs")->find({id => $self->param('testid')}, {prefetch => qw(jobs_assets)});
```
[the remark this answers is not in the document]
okurz: done meanwhile

okurz:
I hope I addressed all comments. I also executed the test modules that failed in CI and they passed now for me.

Martchus (approved):

baierjan (approved):

PR #4203, opened 2021-09-14
Fetchneedles improvements

phil-hands:
This is mostly about making it easy to subsequently tune options used by downstream distros, but I think the changes should be of some benefit for OpenSUSE too.

I've not enabled the git-lfs change here (since you don't use it) but can assure you that it's been in use for openqa.debian.net (and workers) for months, and it works well.

### DIFF
--- a/script/fetchneedles
+++ b/script/fetchneedles
@@ -10,6 +10,7 @@
 : "${username:="openQA web UI"}"
 : "${product:="$dist"}"
 
+: "${git_lfs:="0"}"
 : "${needles_separate:="1"}"
 : "${needles_giturl:="git://github.com/os-autoinst/os-autoinst-needles-opensuse.git"}"
 : "${needles_branch:="master"}"
@@ -123,6 +124,9 @@ else
             git clone --depth 1 -b "$needles_branch" "$needles_giturl" .
             git config user.email "$email"
             git config user.name "$username"
+        fi
+        if [ "$git_lfs" = 1 ]; then
+            git lfs install --local
         fi
     else
         do_fetch "$target"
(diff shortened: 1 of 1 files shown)


Martchus (approved):

script/fetchneedles
```diff
@@ -1,23 +1,32 @@
 #!/bin/sh -e
-[ "$1" = "-h" ] || [ "$1" = "--help" ] && echo "Get openSUSE tests and needles from https://github.com/os-autoinst/os-autoinst-distri-opensuse" && exit
-
 : "${dbuser:="geekotest"}"
 : "${dbgroup:="www"}"
 
+: "${dist_name:=${dist:-"OpenSUSE"}}" # the display name, for the help message
```
okurz, line 5: It's actually "openSUSE" with a lowercase leading character.
It's only for the help text but I am sure you understand that we should be consistent about the project's names :)

okurz (changes requested):

script/fetchneedles
okurz: Fixed that with an updated commit

okurz (approved):

Martchus:
Shellcheck has some remarks, see the output of the "ci/circleci: checkstyle` check.

phil-hands:
> Shellcheck has some remarks, see the output of the "ci/circleci: checkstyle` check.

Fair enough -- here's a fix for that:  https://github.com/phil-hands/openQA/commit/3bd81e12067a8ffd7135a98f9aad57b8857283ac

Do you want that as a new pull request, or is that enough as it is?

BTW I've only tested that it produces the expected output, which it did, but not yet tried it out beyond that ... which ought to be fine since this is in a conditional that only runs with -h/--help options.

Cheers, Phil.

okurz:
@phil-hands thanks for the update. With the help of mergify you could see that we are good with the current pull request ;)

PR #4205, opened 2021-09-15
Dependency cron 2021-09-15

openqabot:


Martchus (approved):

okurz:
@Martchus did you see the flaky coverage? Unless I am mistaken we should handle that and did not have a ticket for that so I created https://progress.opensuse.org/issues/98655 for the flaky coverage

kalikiana (approved):

PR #4206, opened 2021-09-16
Comment back to origin job when investigate retry job passes

Amrysliu:
This is another way to implement the comment back requirement.
Compare to https://github.com/os-autoinst/scripts/pull/109, we don't
need to start a minion task for every passed job.

Related: https://progress.opensuse.org/issues/95746

### DIFF
--- a/lib/OpenQA/Schema/Result/Jobs.pm
+++ b/lib/OpenQA/Schema/Result/Jobs.pm
@@ -2068,6 +2068,8 @@ sub done {
     my $carried_over = $self->carry_over_bugrefs;
     $self->enqueue_finalize_job_results($carried_over);
 
+    $self->diagnose_investigate_retry_job;
+
     # stop other jobs in the cluster
     if (defined $new_val{result} && !grep { $result eq $_ } OK_RESULTS) {
         my $jobs = $self->cluster_jobs(cancelmode => 1);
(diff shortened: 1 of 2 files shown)


Martchus (changes requested):
I would rather do this in the finalize Minion task because this additional handling doesn't affect the response of the API request and can therefore easily be offloaded.

It is also a little bit questionable to add code to openQA which is specific to a certain hook script we're running on a particular instance. I'm wondering whether this could be at least a plugin which one would enable within the settings. (This investigation plugin could at some point also contain the logic from the hook script as I suppose the long-term plan is to move it closer to openQA anyways.)

okurz:
I agree with both points that Martchus made. Moving to the finalize minion task should be easy by just looking into the "enqueue" method in https://github.com/os-autoinst/openQA/pull/4206/files#diff-1c4e8d9bc0096026bc67314675374924c7a5da02359b426a1c6401eac9e90abeR2069 but turning this functionality into a full plugin would likely be too much work for the ticket at hand. For now I think it would still be easier to continue with https://github.com/os-autoinst/scripts/pull/109 but for failed jobs.

Amrysliu:
Continue with https://github.com/os-autoinst/scripts/pull/109, close this one.

PR #4208, opened 2021-09-16
Consistently describe "Idle" workers as such

okurz:
Often people were confused finding not the expected number of "Online"
workers on /admin/workers as the meaning was ambiguous. It could
describe a connected worker regardless of if it is working on jobs,
waiting for new jobs or being broken. In the other case "Online" meant
actually workers connected to the webUI but not working on any jobs,
i.e. "idle". The terms "Online" and "Idle" were both already used in the
statistics box on /admin/workers. This commit changes the table
selection control elements as well as the entries in the table to list
"Idle" while keeping the default to show "Idle" workers, same as in
before but with the updated name.

It might be desired to show all "online" or currently connected or alive
workers which would mean changing the default.


Martchus (approved):
The annoyed me as well when I used the workers table the first time.

Amrysliu (approved):

Martchus (approved):

PR #4209, opened 2021-09-16
Refresh asset size after the upload has been completed

Martchus:
Otherwise it looks like the asset is known but does not exist and one gets
an error about the missing asset when trying to restart a job requiring it.

This is basically a follow-up of
https://github.com/os-autoinst/openQA/pull/4136 which introduced the
regression.

---

@mdoucha This fixes the problem you've mentioned in chat.

### DIFF
--- a/lib/OpenQA/Schema/ResultSet/Assets.pm
+++ b/lib/OpenQA/Schema/ResultSet/Assets.pm
@@ -43,6 +43,7 @@ sub register ($self, $type, $name, $options = {}) {
     $self->result_source->schema->txn_do(
         sub {
             my $asset = $self->find_or_create({type => $type, name => $name}, {key => 'assets_type_name'});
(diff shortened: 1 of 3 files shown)


okurz (approved):

Amrysliu (approved):
LGTM

kalikiana:
For the record, `05-scheduler-fullt.` failed with `not ok 1 - Allocated maximum number of jobs that could have been allocated` (this error message is unique in our test suite)

okurz:
> For the record, `t/05-scheduler-full.t` failed with `not ok 1 - Allocated maximum number of jobs that could have been allocated` (this error message is unique in our test suite)

But then it did not fail anymore. I assume someone retriggered CI tests manually? If yes, which ticket covers the failure?

Martchus:
I did not restart any test.

PR #4210, opened 2021-09-17
Explicitly install os-autoinst dependencies in container

asdil12:
As OBS seems to do some black magic zypper commandline parsing
and not supporting recommends.

https://progress.opensuse.org/issues/97751


asdil12:
This should get better with https://github.com/os-autoinst/os-autoinst/pull/1775

Martchus (approved):

PR #4212, opened 2021-09-17
Mark non-existing asset as such in the downloads tab

Martchus:
* It was brought up in the chat that the UI could be more clear here
* Maybe helps with https://progress.opensuse.org/issues/98388 as well

### DIFF
--- a/t/ui/07-file.t
+++ b/t/ui/07-file.t
@@ -24,7 +24,8 @@ use Mojo::File 'path';
 use OpenQA::Test::TimeLimit '8';
 use OpenQA::Test::Case;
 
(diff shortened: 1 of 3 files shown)


kalikiana (approved):

okurz (changes requested):
failure in t/05-scheduler-full.t: https://app.circleci.com/pipelines/github/os-autoinst/openQA/7726/workflows/d9c5ff5f-2d11-4862-9326-d919e334f4b6/jobs/72940/steps . Does not look related but should not go unhandled.

Martchus:
Looks like the problem persists after restarting. It still looks unrelated and locally the test passes. So I'm creating a ticket.

pevik (approved):

PR #4213, opened 2021-09-20
Make clear 'Module' field in test overview page

mimi1vx:
Current 'module' and 'module_re' are confusing for openqa users

### DIFF
--- a/templates/webapi/test/overview.html.ep
+++ b/templates/webapi/test/overview.html.ep
@@ -128,7 +128,9 @@
                 <div class="row" id="filter-modules">
                     <div class="col-4">
                         <div class="form-group">
-                            <label for="filter-modules"><strong>Module</strong></label>
+                            <label for="filter-modules"><strong>Module name</strong></label>
+                            <%= help_popover('Help for the <em>Module name</em> filter' => '
+                            <p>Shows jobs that contain the specified modules</p>') %>
                             <input name="modules" type="text" id="modules"
                             placeholder="modules, comma separated, e.g. mod1,mod2"
                             class="form-control">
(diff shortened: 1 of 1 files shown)


templates/webapi/test/overview.html.ep
```diff
@@ -128,15 +128,15 @@
                 <div class="row" id="filter-modules">
                     <div class="col-4">
                         <div class="form-group">
-                            <label for="filter-modules"><strong>Module</strong></label>
+                            <label for="filter-modules"><strong>Module by name</strong></label>
```
Martchus, line 131: This is not a grouping but just a filtering criteria. So I wouldn't use "by" here:
```suggestion
                            <label for="filter-modules"><strong>Module name</strong></label>
```

templates/webapi/test/overview.html.ep
```diff
@@ -128,15 +128,15 @@
...
                             <input name="modules" type="text" id="modules"
                             placeholder="modules, comma separated, e.g. mod1,mod2"
                             class="form-control">
                         </div>
                     </div>
                     <div class="col-4">
                         <div class="form-group">
-                            <label for="filter-module-re"><strong>Test module</strong></label>
+                            <label for="filter-module-re"><strong>Module by content</strong></label>
```
Martchus, line 139: Same as above and I suppose "code" is more specific here:
```suggestion
                            <label for="filter-module-re"><strong>Module code</strong></label>
```

templates/webapi/test/overview.html.ep
```diff
@@ -152,7 +152,8 @@
                     </div>
                     <div class="col-2 align-self-center">
                          <%= help_popover('Help for the <em>Module</em> filter' => '
-                         <p>Shows jobs that contain the specified modules and allows also to filter them by the results.</p>
+                         <p><strong>Module by name</strong> - Shows jobs that contain the specified modules and allows also to filter them by the results.</p>
```
Martchus, line 155: The "and allows also to filter them by the results" does not relate to "Module by name" and deserves its own paragraph.

templates/webapi/test/overview.html.ep
```diff
@@ -152,7 +152,8 @@
                     </div>
                     <div class="col-2 align-self-center">
                          <%= help_popover('Help for the <em>Module</em> filter' => '
```
Martchus, line 154:
```suggestion
                         <%= help_popover('Help for the module filters' => '
```
Or just add a help icon after each heading. That would likely be the best.

Martchus (changes requested):
For the sake of consistency with my suggestions I'd also rename "Module Result" to "Module result". (This is not a name where one would capitalize each word.)

templates/webapi/test/overview.html.ep
mimi1vx: Good idea ..

templates/webapi/test/overview.html.ep
```diff
@@ -128,33 +128,35 @@
...
                             <input type="text" class="form-control" name="module_re" placeholder="regular expression" id="filter-module-re">
                         </div>
                     </div>
                     <div class="col-2">
                         <div class="form-group">
-                            <label for="modules_result"><strong>Module Result</strong></label>
+                            <label for="modules_result"><strong>Module result</strong></label>
+                            <%= help_popover('Help for the <em>Module result</em> filter' => '
+                            <p>Show job with selected result.</p>
+                            <p>If module result is not selected, shows all the specified modules regardless of the results.</p>') %>
```
Martchus, line 152: I'd use "Show**s** job**s**" for consistency with the rest. The further paragraph can likely be removed as it should be clear anyways that empty boxes mean the filter isn't effective.
```suggestion
                            <p>Shows jobs with selected result.</p>') %>
```

Martchus (approved):

kalikiana (approved):

templates/webapi/test/overview.html.ep
okurz: @Martchus My suggestion was "content" as well. What do you think is more specific about "code" here?

PR #4214, opened 2021-09-20
Add ignore-missing-assets to clone-job

foursixnine:
Avoid problems when some assets are missing (i.e uefi-pvars), which
aren't really needed for specific cases

```
https://openqa.suse.de/tests/7126208/asset/hdd/SLES-15-SP4-x86_64-Build36.1@64bit-gnome-uefi-vars.qcow2 to 
/srv/data/openqa/development_basedir/openqa/share/factory//hdd/SLES-15-SP4-x86_64-Build36.1@64bit-gnome-uefi-vars.qcow2
7126208 failed: SLES-15-SP4-x86_64-Build36.1@64bit-gnome-uefi-vars.qcow2404 Not Found
```

### DIFF
--- a/lib/OpenQA/Script/CloneJob.pm
+++ b/lib/OpenQA/Script/CloneJob.pm
@@ -116,7 +116,11 @@ sub clone_job_download_assets ($jobid, $job, $remote, $remote_url, $ua, $options
 
             print "downloading\n$from\nto\n$dst\n";
             my $r = $ua->mirror($from, $dst);
-            die "$jobid failed: ", $r->status_line, "\n" unless $r->is_success || $r->code == 304;
+            unless ($r->is_success || $r->code == 304) {
+                print "$jobid failed: $file", $r->status_line, "\n";
+                die "Can't clone due to missing assets: ", $r->status_line, " \n"
+                  unless $options->{'ignore-missing-assets'};
+            }
 
             # ensure the asset cleanup preserves the asset the configured amount of days starting from the time
             # it has been cloned (otherwise old assets might be cleaned up directly again after cloning)
--- a/script/openqa-clone-job
+++ b/script/openqa-clone-job
@@ -113,6 +113,10 @@ job again which is of course only possible if --host is the local machine.
 Do NOT download assets. You need to ensure all required assets are provided
 yourself.
 
+=item B<--ignore-missing-assets>
+
+Cloning a job will not fail if an asset is missing.
+
 =item B<--clone-children>
 
 Clone child jobs as well. This is useful if you want to clone a full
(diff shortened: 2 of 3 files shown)


lib/OpenQA/Script/CloneJob.pm
```diff
@@ -116,7 +116,10 @@ sub clone_job_download_assets ($jobid, $job, $remote, $remote_url, $ua, $options
 
             print "downloading\n$from\nto\n$dst\n";
             my $r = $ua->mirror($from, $dst);
-            die "$jobid failed: ", $r->status_line, "\n" unless $r->is_success || $r->code == 304;
+            unless ( $r->is_success || $r->code == 304){
+                print "$jobid failed: $file", $r->status_line, "\n";
+                die "Can't clone due to missing assets: ", $r->status_line," \n" unless $options->{'ignore-missing-assets'};
+            }
```
foursixnine, line 122: I'm not fan of so many unless; I'm open to suggestions :)

Martchus:
It would also make sense to add a permanent exception for `UEFI_PFLASH_VARS` like we already have it in other places (as there are simply too many tests which specify the asset but don't use it after all).

foursixnine:
> `UEFI_PFLASH_VARS`

You mean something like disabling download of UEFI_PFLASH_VARS by default if UEFI variable is not set on the job?. I'm not so sure, also adding the code for that would make the asset management a bit more complex.

Martchus:
I mean it would ignore download errors for `UEFI_PFLASH_VARS` by default. We already have similar handling for `UEFI_PFLASH_VARS` in other places so it would be in-line with that.

foursixnine:
> I mean it would ignore download errors for `UEFI_PFLASH_VARS` by default. We already have similar handling for `UEFI_PFLASH_VARS` in other places so it would be in-line with that.

Hmm I think I'll stick to my proposition here, as it doesn't hide errors.

On a second note, I'm having trouble figuring out how to add a test for this (is it actually needed?)

foursixnine:
@Martchus looking at the `clone_job_download_assets`, adding that behavior by default would make it look a bit uglier, and again, we end up in the place of: if we add a default ignore for `UEFI_PFLASH_VARS` we'd also need to add an option to not ignore it. I think the drawbacks here, outweight any possible benefit :/

Martchus:
Ok, for tests you could extend `t/35-script_clone_job.t`.

foursixnine:
> Ok, for tests you could extend `t/35-script_clone_job.t`.

This I got, what I don't get is how to alter values in Test::FakeLWPUserAgentMirrorResult so that `is_success` and `code` are modifiable without duplicating both mock packages

Martchus:
Should I give extending the tests a try? I could push an additional commit here.

foursixnine:
> Should I give extending the tests a try? I could push an additional commit here.

Please do, I've been caught up in other stuff and hadn't had the chance to come back to this

okurz (approved):
It's nice but in the end we need to find a better solution for all affected users because I don't want to answer 1k+ support questions with "just add --ignore-missing-assets" everytime :)

Martchus (approved):

PR #4215, opened 2021-09-20
Allow configuring concurrent cleanup

Martchus:
* Default behavior remains unchanged (all cleanup tasks block each other)
* For simplicity this setting affects all cleanup tasks (and not just asset
  and result cleanup).
* See https://progress.opensuse.org/issues/98922

### DIFF
--- a/docs/GettingStarted.asciidoc
+++ b/docs/GettingStarted.asciidoc
@@ -295,10 +295,10 @@ inherit properties from the category. The categories are meant to combine job gr
 with common builds so test results for the same build can be shown together on
 the index page.
 
-
+[id="basic_cleanup"]
 === Cleanup
 IMPORTANT: openQA automatically deletes data that it considers "old" based on
-different settings. For example job data is deleted from old jobs by the `gru` task.
+different settings. For example old jobs and assets are deleted at some point.
 
 The following cleanup settings can be done on job-group-level:
 
@@ -321,6 +321,9 @@ be completely removed from the database.
 
 *NOTE* Jobs which do not belong to a job group are currently not affected by
 the mentioned cleanup properties.
+
+*NOTE* For more details, checkout the
+<<Installing.asciidoc#advanced_cleanup,advanced cleanup>> section.
 
 *NOTE* Archiving of important jobs can be enabled. Checkout the related settings
 within the `[archiving]` section of the config file for details.
--- a/etc/openqa/openqa.ini
+++ b/etc/openqa/openqa.ini
@@ -149,6 +149,13 @@ blocklist = job_grab job_done
 #exchange = pubsub
 #topic_prefix = suse
 
+# Cleanup related configuration (besides limits)
+[cleanup]
+# Specifies whether different cleanup tasks can run in parallel. This option
+# makes particularly sense if assets and results are stored on different storage
+# locations or if your common storage solution is performant enough.
+concurrent = 0
+
 # Default limits for cleanup (sizes are in GiB, durations in days, zero denotes infinity)
 [default_group_limits]
 #asset_size_limit = 100 # only used on job group level (parent groups have no default)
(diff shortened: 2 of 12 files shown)


etc/openqa/openqa.ini
```diff
@@ -149,6 +149,13 @@ blocklist = job_grab job_done
 #exchange = pubsub
 #topic_prefix = suse
 
+# Cleanup related configuration (besides limits)
+[cleanup]
+# Specifies whether different cleanup tasks can run in parallel. This makes
+# option particularly sense if a different storage for assets and results is
```
okurz, line 155: s/This makes option/This option makes/ ?

okurz, line 155: s/if a different storage for assets and results is used/if assets and results are stored on different storage locations/

Also I would add "… or if your common storage solution is performant enough".

okurz (changes requested):
I suggest to also add something to the documentation without duplicating the comments from etc.

docs/GettingStarted.asciidoc
```diff
@@ -298,7 +298,12 @@ the index page.
 
 === Cleanup
 IMPORTANT: openQA automatically deletes data that it considers "old" based on
-different settings. For example job data is deleted from old jobs by the `gru` task.
+different settings. For example old jobs and assets are deleted at some point.
+
+*NOTE* Cleanup jobs run within the Minion job queue provided by
```
okurz, line 303: Paragraph is good but does not fit well within the "getting started" guide which targets beginners, not admins. Probably better in "Installing"?

Martchus: I suppose that counts for the whole cleanup section.

okurz: hm, wouldn't say so necessarily. I think it's good to explain the concept of cleanup. Doesn't mean we need to explain all configuration options at once.

Martchus: I created another section under "Advannced configuration" and linked it here.

okurz (approved):
Lovely :)

etc/openqa/openqa.ini
```diff
@@ -149,6 +149,13 @@ blocklist = job_grab job_done
 #exchange = pubsub
 #topic_prefix = suse
 
+# Cleanup related configuration (besides limits)
+[cleanup]
+# Specifies whether different cleanup tasks can run in parallel. This option
+# makes particularly sense if assets and results are stored on different storage
+# locations or if your common storage solution is performant enough.
+concurrent = 0
```
kalikiana, line 157: For some reason I thought this might take a number of tasks to run in parallel. Looking at the implementation clarified that of course... might be clear enough anyway, just wanted to mention it

kalikiana (approved):
I remember wondering why these weren't supposed to run in parallel. Nice to see this straight-forward support for enabling it!

etc/openqa/openqa.ini
Martchus: I hope the word "whether" in the comment makes it clear that this is just a flag.

nicksinger: would it be possible to use bools like "true" or "false" here? This would clarify all ambiguity. If not I think mentioning it in the comment would make sense

lib/OpenQA/Task/Bug/Limit.pm
```diff
@@ -30,9 +31,7 @@ sub _limit {
     return $job->finish('Previous limit_bugs job is still active')
       unless my $guard = $app->minion->guard('limit_bugs_task', ONE_DAY);
 
-    # prevent multiple limit_* tasks to run in parallel
-    return $job->retry({delay => 60})
-      unless my $limit_guard = $app->minion->guard('limit_tasks', ONE_DAY);
+    return undef unless my $limit_guard = acquire_limit_lock_or_retry($job);
```
nicksinger, line 34: It looks like the _limit subroutine has duplicate logic at multiple places. Would it make sense to define a common subroutine for all to which you just pass the argument for `guard()`?

Martchus: Yes, it could likely be de-duplicated further.

etc/openqa/openqa.ini
Martchus: In Perl there's no true/false do this would require extra handling and would be inconsistent with other settings.

PR #4216, opened 2021-09-20
Add OPENQA_HOSTNAME to the job settings

foursixnine:
When a worker is shared for certain jobs, specially when it comes to svirt jobs, can become a bit difficult to figure out in which webUI a job has been executed or triggered from, leaving room for potential collisions (see poo#97532 for instance).

Depends on: https://github.com/os-autoinst/os-autoinst/pull/1777

### DIFF
--- a/lib/OpenQA/Worker/Job.pm
+++ b/lib/OpenQA/Worker/Job.pm
@@ -245,6 +245,14 @@ sub start {
     my $worker                 = $self->worker;
     my $global_worker_settings = $worker->settings->global_settings;
     @{$job_settings}{keys %$global_worker_settings} = values %$global_worker_settings;
+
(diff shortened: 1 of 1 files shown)


okurz (changes requested):
I don't understand how you would find jobs if not over the webUI unless this is about workers rich don't delete it's pool dir. Could you explain again for dummies like me?

foursixnine:
@okurz see https://github.com/os-autoinst/os-autoinst/pull/1777 :)

okurz (approved):

okurz:
@foursixnine please leave merging to mergify by default

foursixnine:
> @foursixnine please leave merging to mergify by default

Right! Force of habit :)

PR #4217, opened 2021-09-20
Worker.pm: Print pool directory on error

pevik:
Printing directory name helps new users to investigate permissions.
Also this is the only related error message which does not print the directory name.

Signed-off-by: Petr Vorel <pvorel@suse.cz>


okurz (approved):

Martchus (approved):

okurz:
fullstack test failed, handled in https://progress.opensuse.org/issues/98952, retriggering failed run.

PR #4220, opened 2021-09-21
Fix errors reported by make test-shellcheck target

pevik:


tools/generate-documentation
```diff
@@ -83,7 +83,8 @@ update_docs() {
         mkdir out || exit 1
         (
         cd out
-        if [[ -z "${PULL_REQUEST_USER}" ]]; then
```
Martchus, line 86: But we're using Bash here and using `[[` avoids spawning a sub process so I wouldn't change that. The same counts for the other occurrences,

tools/generate-packed-assets
```diff
@@ -48,8 +48,7 @@ BUILD_CACHE="plugin AssetPack => { pipes => [qw(${ASSETPACK_PLUGINS})]} and app-
 BUILD_CACHE_OPTS='-Ilib/ -MMojolicious::Lite'
 
 generate() {
-    # shellcheck disable=SC2039
-    for _ in {1..3}; do
+    for _ in $(seq 1 3); do
```
Martchus, line 51: Why not just use Bash here by changing the shebang? Again, it makes sense to avoid spawning a sub process.

Martchus (changes requested):

Martchus (approved):

pevik:
`tools/generate-documentation: Remove useless assignment` is a wrong subject due it starts with lower case. Most projects should consider this as a valid subject, but ok I'll make check script happy and rename it to `Remove useless assignment in ools/generate-documentation` (not only that using file at the beginning it's better readability (there could be just `tools` when changing more files, which would be also wrong), but also using file helps to remove ` in ` thus save some spaces in subject which is required to be <= 80 chars.

Martchus:
I also think the checker should allow prefixes like that.

okurz (commented):
I don't understand why the shell problems that you rightly fix are not reported by shellcheck in our CI. Where did you see failures?

pevik:
Running on Tumbleweed:
```
$ shellcheck -V
ShellCheck - shell script analysis tool
version: 0.7.2

$ make test-shellcheck

In tools/generate-documentation line 86:
        if [[ -z "${PULL_REQUEST_USER}" ]]; then
                  ^------------------^ SC2030: Modification of PULL_REQUEST_USER is local (to subshell caused by (..) group).


In tools/generate-documentation line 117:
            if [[ -n "${PUBLISH}" ]] && [[ "${PUBLISH}" != 0 ]]; then
                      ^--------^ SC2030: Modification of PUBLISH is local (to subshell caused by (..) group).


In tools/generate-documentation line 179:
PUBLISH="${PUBLISH}"
^-----^ SC2269: This variable is assigned to itself, so the assignment does nothing.
         ^--------^ SC2031: PUBLISH was modified in a subshell. That change might be lost.


In tools/generate-documentation line 184:
if [[ -z "${PULL_REQUEST_USER}" ]]; then
          ^------------------^ SC2031: PULL_REQUEST_USER was modified in a subshell. That change might be lost.


In tools/generate-packed-assets line 52:
    for _ in {1..3}; do
             ^----^ SC3009: In POSIX sh, brace expansion is undefined.


In tools/run-tests-within-container line 44:
if [ "x$FULLSTACK" = x1 ]; then
     ^-----------^ SC2268: Avoid x-prefix in comparisons as it no longer serves a purpose.

Did you mean: 
if [ "$FULLSTACK" = 1 ]; then


In tools/run-tests-within-container line 46:
elif [ "x$CHECKSTYLE" = x1 ]; then
       ^------------^ SC2268: Avoid x-prefix in comparisons as it no longer serves a purpose.

Did you mean: 
elif [ "$CHECKSTYLE" = 1 ]; then


In tools/run-tests-within-container line 49:
    if [ "x$UITESTS" = x1 ]; then
         ^---------^ SC2268: Avoid x-prefix in comparisons as it no longer serves a purpose.

Did you mean: 
    if [ "$UITESTS" = 1 ]; then

For more information:
  https://www.shellcheck.net/wiki/SC3009 -- In POSIX sh, brace expansion is u...
  https://www.shellcheck.net/wiki/SC2030 -- Modification of PUBLISH is local ...
  https://www.shellcheck.net/wiki/SC2031 -- PUBLISH was modified in a subshel...
make: *** [Makefile:285: test-shellcheck] Error 1
# exit code 2 quits make test target
```

pevik:
Also except these 2 are the results valid reports and should be fixed regardless CI is complaining or not:
```
In tools/generate-documentation line 86:
        if [[ -z "${PULL_REQUEST_USER}" ]]; then
                  ^------------------^ SC2030: Modification of PULL_REQUEST_USER is local (to subshell caused by (..) group).

In tools/generate-documentation line 117:
            if [[ -n "${PUBLISH}" ]] && [[ "${PUBLISH}" != 0 ]]; then
                      ^--------^ SC2030: Modification of PUBLISH is local (to subshell caused by (..) group).
```

okurz (commented):
Alright. In this case it's fine. The commit message are ok but the PR title is misleading. It's making it sound like the Makefile target is broken or something

okurz (approved):

PR #4222, opened 2021-09-22
Small improvements for container setup

Martchus:
* See particular commit messages
* There's no ticket, these are just fixes for problems I've run into when using the setup with OpenID and no manual CLI parameters for API key/secret.

### DIFF
--- a/container/webui/conf/client.conf
+++ b/container/webui/conf/client.conf
@@ -1,3 +1,7 @@
+[localhost]
+key = 1234567890ABCDEF
+secret = 1234567890ABCDEF
+
 [scheduler]
 key = 1234567890ABCDEF
 secret = 1234567890ABCDEF
--- a/container/webui/conf/openqa.ini
+++ b/container/webui/conf/openqa.ini
@@ -1,3 +1,8 @@
+[global]
+# change to the URL the web UI will be available under so redirection for
+# authentication works
+base_url = https://localhost
+
 [auth]
 method = OpenID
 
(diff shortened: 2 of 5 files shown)


container/webui/conf/client.conf
```diff
@@ -1,3 +1,7 @@
+[localhost]
+key = 1234567890ABCDEF
+secret = 1234567890ABCDEF
+
 [scheduler]
```
okurz, line 5: Maybe you can throw together a quick hack to instead of needing to support to duplicate all these sections to define something like this:

```suggestion
[localhost,scheduler]
```

okurz:
>     * There's no ticket, these are just fixes for problems I've run into when using the setup with OpenID and no manual CLI parameters for API key/secret.

When did you run into this? Last time you had been working on related areas and fixing tickets and these are left-over improvements? Or did you run a docker-compose setup again to try out something different?

container/webui/conf/client.conf
Martchus: Allowing this syntax would be a good idea. It'll likely be easy to achieve so I can give it a try later (or tomorrow). However, I'd like to have it coded in a clean way and see no reason to rush it.

okurz (approved):

Martchus:
Looks like we have yet another unstable test:

```
    #   Failed test 'highlight appears correct'
    #   at t/ui/12-needle-edit.t line 277.
    #                   'Unable to save needle:
    # No areas defined.'
    #     doesn't match '(?^u:Needle logpackages-before-package-selection-\d{8} created/updated - restart job)'

#   Failed test 'Saving needle when "taking matches" selected but no matches present'
#   at t/ui/12-needle-edit.t line 414.
Bailout called.  Further testing stopped:  clickElement: element not interactable at /home/squamata/project/t/ui/../lib/OpenQA/SeleniumTest.pm:81
```

Amrysliu (approved):
LGTM

PR #4223, opened 2021-09-22
Fix build in tests/overview when groupid is defined

mimi1vx:
Fixes https://progress.opensuse.org/issues/98258

### DIFF
--- a/lib/OpenQA/Schema/ResultSet/Jobs.pm
+++ b/lib/OpenQA/Schema/ResultSet/Jobs.pm
@@ -317,8 +317,11 @@ sub prepare_complex_query_search_args ($self, $args) {
             push(@conds, {'me.id' => {-in => $subquery->get_column('job_id')->as_query}});
         }
 
-        for my $key (qw(build distri version flavor arch test machine)) {
+        for my $key (qw(distri version flavor arch test machine)) {
             push(@conds, {"me." . uc($key) => $args->{$key}}) if $args->{$key};
+        }
+        if (my $build = $args->{build}) {
+            push @conds, {'me.BUILD' => ref $build eq 'ARRAY' ? {-in => $build} : $build};
         }
     }
 
--- a/lib/OpenQA/WebAPI/Plugin/Helpers.pm
+++ b/lib/OpenQA/WebAPI/Plugin/Helpers.pm
@@ -356,7 +356,7 @@ sub _compose_job_overview_search_args ($c) {
     $v->optional('limit',          'not_empty')->num(0, undef);
 
     # add simple query params to search args
-    for my $arg (qw(distri version flavor build test limit)) {
+    for my $arg (qw(distri version flavor test limit)) {
         next unless $v->is_valid($arg);
         my $params      = $v->every_param($arg);
         my $param_count = scalar @$params;
@@ -367,6 +367,11 @@ sub _compose_job_overview_search_args ($c) {
             $search_args{$arg} = {-in => $params};
         }
     }
+
+    # handle build separately
+    my $build = $v->every_param('build');
+    $search_args{build} = $build if $build && @$build;
+
     my $modules = $v->every_param('modules');
     $search_args{modules} = $modules if $modules && @$modules;
     my $result = $v->every_param('modules_result');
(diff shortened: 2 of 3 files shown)


okurz (changes requested):
s/wheen/when/

okurz (changes requested):
Please add tests for the issue. I recommend to use https://en.wikipedia.org/wiki/Test-driven_development although for the issue at hand it's obviously too late already :D

lib/OpenQA/Schema/ResultSet/Jobs.pm
```diff
@@ -317,9 +317,12 @@ sub prepare_complex_query_search_args ($self, $args) {
             push(@conds, {'me.id' => {-in => $subquery->get_column('job_id')->as_query}});
         }
 
-        for my $key (qw(build distri version flavor arch test machine)) {
+        for my $key (qw(distri version flavor arch test machine)) {
             push(@conds, {"me." . uc($key) => $args->{$key}}) if $args->{$key};
         }
+        if (defined $args->{build}) {
+            push(@conds, {"me.BUILD" => {-in => $args->{build}}});
```
Martchus, line 324: Ok, so apparently `$args->{build}` is an array ref here. I hope that's always¹ the case and that the values contained by that array ref are sufficiently validated.

¹ This function might be called in multiple places.

mimi1vx: ARRAYREF or Scalar .. it comes from `_compose_job_overview_search_args` if isn't defined in query or from query ..

Martchus: But we call `complex_query` (which calls this function) also in other places, e.g. `OpenQA::WebAPI::Controller::API::V1::Job::list`. Have you checked the other occurrences as well?

mimi1vx: hmm, you're right

mimi1vx: @Martchus probadly overkill but in this case better safe than sorry 🗡️

lib/OpenQA/Schema/ResultSet/Jobs.pm
```diff
@@ -317,9 +317,16 @@ sub prepare_complex_query_search_args ($self, $args) {
...
+        for my $key (qw(distri version flavor arch test machine)) {
             push(@conds, {"me." . uc($key) => $args->{$key}}) if $args->{$key};
         }
+        if (defined $args->{build}) {
+            my $build = $args->{build};
+            if (ref $build eq 'ARRAY') {
+                push(@conds, {"me.BUILD" => {-in => $build}});
+            }
+            else { push(@conds, {"me.BUILD" => $build}) }
+        }
```
okurz, line 329:
```suggestion
        if (my $build = $args->{build}) {
                push @conds, ref $build eq 'ARRAY' ? {"me.BUILD" => {-in => $build}} : {"me.BUILD" => $build};
        }
```

okurz: I *love* the ternary operator and it's something I look for in all programming languages I use :)

okurz (changes requested):
in your second git commit s/overwiew/overview/ but likely you can just squash the git commit into the first one. Test is part of fix.

lib/OpenQA/Schema/ResultSet/Jobs.pm
```diff
@@ -317,9 +317,12 @@ sub prepare_complex_query_search_args ($self, $args) {
             push(@conds, {'me.id' => {-in => $subquery->get_column('job_id')->as_query}});
         }
 
-        for my $key (qw(build distri version flavor arch test machine)) {
+        for my $key (qw(distri version flavor arch test machine)) {
             push(@conds, {"me." . uc($key) => $args->{$key}}) if $args->{$key};
         }
+        if (my $build = $args->{build}) {
+            push @conds, ref $build eq 'ARRAY' ? {"me.BUILD" => {-in => $build}} : {"me.BUILD" => $build};
```
Martchus, line 324: I would not duplicate the common part and we generally use single quotes when not interpolation is required (despite not all old code has been converted to that style):
```suggestion
            push @conds, {'me.BUILD' => (ref $build eq 'ARRAY' ? {-in => $build} : $build)};
```

okurz: my suggestion missed that part, sorry

okurz: Actually not fixed:

```patch
-    push @conds, ref $build eq 'ARRAY' ? {'me.BUILD' => {-in => $build}} : {'me.BUILD' => $build};
+   push @conds, {'me.BUILD' => ref $build eq 'ARRAY' ? {-in => $build} : $build};

okurz (approved):

Martchus (approved):

PR #4225, opened 2021-09-23
templates: Clarify which filter options in /overview apply to the job

okurz:
We have filter options applying to the complete job as well as the
test modules. For modules we already prefix the strings with "Module".
So to be extra-explicit we can add "Job" as prefix for "Result" and
"State".

Related progress issue: https://progress.opensuse.org/issues/98460

### DIFF
--- a/templates/webapi/test/overview.html.ep
+++ b/templates/webapi/test/overview.html.ep
@@ -86,13 +86,13 @@
         <div class="card-body">
             <form action="#" type="get" id="filter-form">
                 <div class="form-group" id="filter-results">
-                    <strong>Result</strong>
+                    <strong>Job result</strong>
                     % for my $result (OpenQA::Jobs::Constants::RESULTS) {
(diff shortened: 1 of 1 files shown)


mimi1vx (approved):

baierjan (approved):

Martchus (approved):

mimi1vx:
Please add tests for the issue. I recommend to use https://en.wikipedia.org/wiki/Test-driven_development although for the issue at hand it's obviously too late already :D

okurz:
> Please add tests for the issue. I recommend to use https://en.wikipedia.org/wiki/Test-driven_development although for the issue at hand it's obviously too late already :D

I like it if you try to use my own tools against me ;)

okurz:
hm, https://app.circleci.com/pipelines/github/os-autoinst/openQA/7796/workflows/e98ecc4b-c331-4c5a-918d-c4a32ba1b4fb/jobs/73710 does not look like t/full-stack.t would actually retry. I am confused.

PR #4226, opened 2021-09-23
Optimize asset cleanup

Martchus:
See particular commit messages and https://progress.opensuse.org/issues/97979#note-13 and further comments.

PR #4227, opened 2021-09-23
Optimize asset cleanup (possibly dangerous)

Martchus:
* See particular commit messages and https://progress.opensuse.org/issues/97979#note-13 and further comments.
* Based on #4226 but contains two additional optimizations which could be dangerous, see the commit messages for details.

### DIFF
--- a/lib/OpenQA/Schema/Result/Assets.pm
+++ b/lib/OpenQA/Schema/Result/Assets.pm
@@ -116,9 +116,17 @@ sub remove_from_disk {
 
     my $type = $self->type;
     my $name = $self->name;
-    my $file = locate_asset($type, $name, mustexist => 1);
+    my $file = $self->{_location} = locate_asset($type, $name, mustexist => 1);
     if (!defined $file) {
         log_info("GRU: skipping removal of $type/$name; it does not exist anyways");
+        return undef;
+    }
+    # prevent removal of fixed assets
+    # note: The asset might have been moved into the "fixed" directory after it
+    #       was initially tracked by openQA. We do not re-check whether an asset is
+    #       fixed on every cleanup.
+    if ($self->is_fixed) {
+        log_info("GRU: skipping removal of $type/$name; it is fixed");
         return undef;
     }
     if (-f $file ? unlink($file) : remove_tree($file)) {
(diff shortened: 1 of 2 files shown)


lib/OpenQA/Schema/Result/Assets.pm
```diff
@@ -101,7 +101,7 @@ sub _getDirSize {
 
 sub disk_file {
     my ($self) = @_;
-    return locate_asset($self->type, $self->name);
+    return $self->{_location} // ($self->{_location} = locate_asset($self->type, $self->name));
```
okurz, line 104: I think it should work to use =//

lib/OpenQA/Schema/Result/Assets.pm
```diff
@@ -116,11 +116,19 @@ sub remove_from_disk {
...
     if (!defined $file) {
         log_info("GRU: skipping removal of $type/$name; it does not exist anyways");
         return undef;
     }
+    # prevent removal of fixed assets
+    # note: The asset might have been moved into the "fixed" directory after it
+    #       was initially tracked by openQA. We do not re-check whether an asset is
+    #       fixed on every cleanup.
+    if ($self->is_fixed) {
+        log_info("GRU: skipping removal of $type/$name; it is fixed");
```
okurz, line 129: Does this info message appear more than once or every time we try to delete the asset in vain? If it reappears I would use debug, not info, but then the question would be if we need to prevent the call of remove at all

lib/OpenQA/Schema/Result/Assets.pm
Martchus: Right, I'll change that in the other PR first.

lib/OpenQA/Schema/Result/Assets.pm
Martchus: It will not re-appear because we will still delete the asset in the database. Then, on the next re-scan it'll be tracked as fixed asset and we won't reach this code. But I admit that this is a bit messy.

Martchus:
One additional problem with "Speed up asset cleanup by only checking fixed status if not seen before" comes to my mind: If a user moves an asset from the "fixed" folder to normal assets then openQA will still keep thinking it is fixed; at least in terms of the asset cleanup. That means those assets will nevertheless never be cleaned up. So maybe that optimization isn't worth it.

PR #4228, opened 2021-09-23
Increase timeout to make 05-scheduler-full.t more stable in CircleCI

asdil12:
Ticket: https://progress.opensuse.org/issues/98898

### DIFF
--- a/t/05-scheduler-full.t
+++ b/t/05-scheduler-full.t
@@ -210,7 +210,9 @@ subtest 're-scheduling and incompletion of jobs when worker rejects jobs or goes
     wait_for_worker($schema, 5);
 
     note 'waiting for job to be incompleted';
-    for (0 .. 100) {
+    for (0 .. 200) {
+        note "JOB 99982 STATE:";
+        note $jobs->find(99982)->state;
         last if $jobs->find(99982)->state eq OpenQA::Jobs::Constants::DONE;
         sleep .2;
     }
--- a/Makefile
+++ b/Makefile
@@ -1,7 +1,7 @@
-RETRY ?= 0
+RETRY ?= 10
 # STABILITY_TEST: Set to 1 to fail as soon as any of the RETRY fails rather
 # than succeed if any of the RETRY succeed
-STABILITY_TEST ?= 0
+STABILITY_TEST ?= 1
 # KEEP_DB: Set to 1 to keep the test database process spawned for tests. This
 # can help with faster re-runs of tests but might yield inconsistent results
 KEEP_DB ?= 0


okurz (changes requested):

Statistics please

t/05-scheduler-full.t
```diff
@@ -210,7 +210,7 @@ subtest 're-scheduling and incompletion of jobs when worker rejects jobs or goes
     wait_for_worker($schema, 5);
 
     note 'waiting for job to be incompleted';
-    for (0 .. 100) {
+    for (0 .. 200) {
```
kalikiana, line 213: Please get rid of the magic number make this fail clearly when the condition isn't met.
```suggestion
    wait_for_or_bail_out { $jobs->find(99982)->state eq OpenQA::Jobs::Constants::DONE } 'Job 99982 is done';
```

asdil12:

Please don't copy and paste this in every PR comment - especially if most of it doesn't apply or is already satisfied.

> Statistics please

I'm unable to reproduce this locally so creating a PR is the only way I know to run this code somewhere where it *might* fail.

t/05-scheduler-full.t
asdil12: What do you suggest?

okurz:
> > The commit message could have a bit more details, e.g. what issue is fixed, why this needs to change, to which versions of which product it applies, link to a bug or a fate entry, the choices you made, etc. Also see https://commit.style/ or http://chris.beams.io/posts/git-commit/ as a helpful guide how to write good commit messages.
> 
> Please don't copy and paste this in every PR comment - especially if most of it doesn't apply or is already satisfied.

Then please put yourself in the shoes of someone needing to review your PRs

> > Statistics please
> 
> I'm unable to reproduce this locally so creating a PR is the only way I know to run this code somewhere where it _might_ fail.

Then state that in the git commit message

t/05-scheduler-full.t
asdil12: This always results in a failed test.
The `wait_for_or_bail_out` doesn't really seem to do anything.

asdil12: I think it is a better idea to leave this in but be more verbose to see if the job is behaving the same as locally in a case that actually results in a sporadic error.

kalikiana: What do you mean by "doesn't do anything"? The function waits until the condition is fulfilled after repeatedly calling it. You can override `timeout` and `interval`, which is 60/.1 by default

asdil12: It just returns immediately which of course breaks the three checks afterwards.

asdil12: But oddly I can't seem to reproduce that now anymore.

asdil12:
I would like to have this "as is" for now to be able to gather mentioned statistics "in production" as there is no other way.

okurz:
> I would like to have this "as is" for now to be able to gather mentioned statistics "in production" as there is no other way.

We don't run t/05-scheduler-full.t "in production". Do you mean CI? But we run the same tests on every PR. You can also set `STABILITY_TEST=1` with a bigger `RETRY`value in the Makefile to gather more statistics from circleCI

asdil12:
> > I would like to have this "as is" for now to be able to gather mentioned statistics "in production" as there is no other way.
> 
> We don't run t/05-scheduler-full.t "in production". Do you mean CI? But we run the same tests on every PR. You can also set `STABILITY_TEST=1` with a bigger `RETRY`value in the Makefile to gather more statistics from circleCI

I did that changes to the makefile - let's see what will happen.

PR #4229, opened 2021-09-23
[draft] Add link to overview of test results to progressbar

FruitFly638:
The whole Progressbar is clickable and leads to the Test overview for
the corresponding build. Works for children as well.
Change according to: https://progress.opensuse.org/issues/98403

This is a draft, feed back is very welcome

### DIFF
--- a/templates/webapi/main/group_builds.html.ep
+++ b/templates/webapi/main/group_builds.html.ep
@@ -25,11 +25,13 @@
                 % }
             </span>
         </div>
-        <div class="px-2 align-self-end flex-grow-1">
+        <a data-target="#<%= $group_id %>" href="<%= url_for('tests_overview')->query(@tests_overview_url_data, map { (groupid => $_->{id}) } @{$children});%>" class="px-2 align-self-end flex-grow-1" style="text-decoration: none">
+        <div>
             % if ($max_jobs) {
                 %= include 'main/build_progressbar', max_jobs => $max_jobs, result => $build_res
             % }
         </div>
+        </a>
     </div>
     % if ($children) {
         <div id="<%= $group_id %>" class="build_group_children <%= $default_expanded ? 'show' : '' %> collapse">
@@ -43,9 +45,11 @@
                                 %= include 'main/review_badge', group_build_id => $group_build_id, build_res => $child_res, id_prefix => 'child-'
                             </span>
                         </div>
-                        <div class="px-2 align-self-end flex-grow-1">
+                        <a href="<%= url_for('tests_overview')->query(distri => [sort keys %{$child_res->{distris}}], version => $child_res->{version}, build => $build, groupid => $child->{id});%>" class="px-2 align-self-end flex-grow-1" style="text-decoration: none">
+                        <div >
                             %= include 'main/build_progressbar', max_jobs => $child_res->{total}, result => $child_res
                         </div>
+                        </a>
                     </div>
                 % }
             % }


templates/webapi/main/group_builds.html.ep
```diff
@@ -25,11 +25,13 @@
                 % }
             </span>
         </div>
-        <div class="px-2 align-self-end flex-grow-1">
+        <a data-target="#<%= $group_id %>" href="<%= url_for('tests_overview')->query(@tests_overview_url_data, map { (groupid => $_->{id}) } @{$children});%>" class="px-2 align-self-end flex-grow-1" style="text-decoration: none">
```
okurz, line 28: hm, so this is the same URL that was previously already generated. If we keep it like, the same URL for the complete bar as for the build label itself then we should reuse a variable and not call helpers with duplicated code. Either that or someone has an idea how to make the individual sections clickable while still rendering correctly visually.

Martchus: The cheapest way to make individual sections clickable without disturbing the rendering is to simply add an `onclick` attribute to the relevant elements and handle it in JavaScript. Not nice from an accessibility point of view but for an unimportant feature like acceptable. (You could still at least style the elements to show a pointer as cursor on hover.)

However, I believe it should be possible to style the `a` elements so that they don't affect the rendering. Having the `a` element on the outer level seems wrong; I don't think links are supposed to contain block elements. I would have to experiment myself with the elements to figure it out, though.

Martchus: For accessibility when using `onclick` you could actually assign an aria role: https://w3c.github.io/aria/#link

okurz: that sounds acceptable as well. So two different options to explore if I understand correctly

FruitFly638: Okay, I wanted to avoid Javascript for accessibility reasons but I saw some ways to make it work with that. Since the individual sections have classes like "progress-bar progress-bar-passed" it should be manageable.
Can you have both at the same time so if you don't have Javascript, it links like it does now but with Javascript it links the individual sections instead? Or is that too complicated/not possible in HTML?

Martchus: I wouldn't implement both ways. If we decide to require JavaScript than that's it. We already use it extensively on the page anyways.

Martchus: The comment I've pushed uses normal links plus some additional CSS so it looks good.

templates/webapi/main/group_builds.html.ep
```diff
@@ -44,7 +45,7 @@
                             </span>
                         </div>
                         <div class="px-2 align-self-end flex-grow-1">
-                            %= include 'main/build_progressbar', max_jobs => $child_res->{total}, result => $child_res
+                            %= include 'main/build_progressbar', max_jobs => $child_res->{total}, url_data => \@tests_overview_url_data, result => $child_res
```
perlpunk, line 48: What's the `url_data` parameter about?

Martchus: Good catch, this should have been

```suggestion
                            %= include 'main/build_progressbar', max_jobs => $child_res->{total}, params => \%progress_bar_params, result => $child_res
```

templates/webapi/main/group_builds_functionality_view.html.ep
```diff
@@ -40,7 +40,7 @@
                 </abbr>
             </div>
             <div class="px-2 align-self-end flex-grow-1">
-                %= include 'main/build_progressbar', max_jobs => $child_res->{total}, result => $child_res, class => "mt-0"
+                %= include 'main/build_progressbar', max_jobs => $child_res->{total}, params => undef, result => $child_res, class => "mt-0"
```
Martchus, line 43: And by the way, params is also not supposed to be `undef` here. Passing the actual parameters is still TODO.

okurz:
@FruitFly638 when updating with new or updated commits I suggest to also write an explicit message about what you changed, e.g.

```
Updated with all comments resolved
```

or

```
* Rebased
* Fixed the CI test failure
```

Martchus (commented):
I've been testing it locally and it looks all good now. So I suppose only the tests are still missing.

FruitFly638:
I changed the 17-build-tagging test that was previously failing by increasing the number of 'a[href^=/tests/]' elements it is looking for. A more elegant (and useful) solution would be to distinguish between the progressbar and the build link somehow.

okurz:
> I changed the 17-build-tagging test that was previously failing by increasing the number of 'a[href^=/tests/]' elements it is looking for. A more elegant (and useful) solution would be to distinguish between the progressbar and the build link somehow.

Well, I guess it's fine. Can you provide an example server or screenshot or generated HTML document somewhere?

Is the pull request still a "draft"?

okurz:
please also see the failing "commit message check": https://github.com/os-autoinst/openQA/pull/4229/checks?check_run_id=3768922506

PR #4230, opened 2021-09-23
templates: Add comma to prevent ambiguous state+timestamp combinations

okurz:
Reading a string like

```
State: uploading started about an hour ago
```

can be confused by users to mean that the upload started about an hour
ago when "started" refers to the overall job. Adding a comma should
prevent that ambiguity or at least make it less likely to misunderstand.


Martchus (commented):
Have you checked that there's no space rendered before the `,` (despite the wrapping)?

okurz:
> Have you checked that there's no space rendered before the `,` (despite the wrapping)?

of course! There is. How do I prevent it in a line like

```
                        , started
```
?

templates/webapi/test/infopanel.html.ep
```diff
@@ -39,28 +39,28 @@
...
                                 % $grutext .= join(",", $grudep->gru_task->id, $grudep->gru_task->taskname);
                                 % $grutext .= " ";
                             % }
-                            waiting for background (GRU) tasks: <%= $grutext %>
+                            , waiting for background (GRU) tasks: <%= $grutext %>
                         % }
                     % }
                     % if ($job->t_finished) {
-                        finished
+                        <%= ', finished' =%>
```
Martchus, line 56: Is it required to render the fix string like that to suppress rendering the white-spaces before? If yes, than that's missing in ", waiting for background …" and otherwise I would avoid it. Btw, I wouldn't consider it bad to simply omit the white-spaces here. Correct indentation isn't everything if it leads to surplus white-spaces being rendered.

Martchus (changes requested):
Yes, the `\` at the end is the right way to do it.

templates/webapi/test/infopanel.html.ep
okurz: Yes, it's required. Yes, starting the string from the beginning of line can work as well. I consider this way nicer here as the indentation is consistent and implicitly it was "template code" in before anyway. Now it's more visible that it's code :)

Martchus: Ok, then keep it this way but apply the approach also to ", waiting for background …" .

okurz: done

templates/webapi/test/infopanel.html.ep
```diff
@@ -39,28 +39,28 @@
...
-                        State: <b><%= $job->state %></b>
+                        State: <b><%= $job->state %></b>\
                         % if (scalar($job->gru_dependencies) > 0) {
                             % my $grutext = '';
                             % for my $grudep ($job->gru_dependencies) {
                                 % $grutext .= join(",", $grudep->gru_task->id, $grudep->gru_task->taskname);
                                 % $grutext .= " ";
                             % }
-                            waiting for background (GRU) tasks: <%= $grutext %>
+                            <%= ", waiting for background (GRU) tasks: $grutext" =%>
```
Martchus, line 52:
```suggestion
                            <%= ", waiting for background (GRU) tasks: $grutext" %>
```
The same counts for the other occurrences.

It apparently doesn't cause any harm but I suppose the correct way (and how we do it everywhere else, and how it is [documented](https://docs.mojolicious.org/Mojo/Template#SYNOPSIS)) is not to have the equal sign at the end.

Martchus (changes requested):

templates/webapi/test/infopanel.html.ep
Martchus: Oh, this *is* for trimming the surrounding white-space characters. So never mind.

Martchus (approved):

PR #4232, opened 2021-09-24
Dependency cron 2021-09-24

openqabot:


.circleci/ci-packages.txt
```diff
@@ -110,11 +110,8 @@ perl-HTTP-Daemon-6.01
 perl-HTTP-Date-6.02
 perl-HTTP-Message-6.14
 perl-HTTP-Negotiate-6.01
-perl-HTTP-Parser-XS-0.17
 perl-Import-Into-1.002005
 perl-indirect-0.38
-perl-Inline-0.80
-perl-Inline-Python-0.56
```
Martchus, line 117: I don't think we should have lost most of these packages and a missing `perl-Inline-Python` would explain why the fullstack test fails here.

Martchus (changes requested):

okurz:
https://github.com/os-autoinst/openQA/pull/4239

PR #4233, opened 2021-09-24
WIP -- check stability of t/full-stack.t

okurz:
Related progress issue: https://progress.opensuse.org/issues/98952

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -176,7 +176,7 @@ test-unstable:
 
 .PHONY: test-fullstack
 test-fullstack:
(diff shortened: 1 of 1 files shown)


kalikiana:
Doesn't seem like this will ever finish.... I proposed #4243 which attempts to improve diagnostics

okurz:
> Doesn't seem like this will ever finish.... I proposed #4243 which attempts to improve diagnostics

At least not *today* :) But you could have likely seen results while tests had been running. But the most important results from my previous change are already recorded in https://progress.opensuse.org/issues/98952 as well.

PR #4236, opened 2021-09-27
Dependency cron 2021-09-27

openqabot:


.circleci/ci-packages.txt
```diff
@@ -110,11 +110,8 @@ perl-HTTP-Daemon-6.01
 perl-HTTP-Date-6.02
 perl-HTTP-Message-6.14
 perl-HTTP-Negotiate-6.01
-perl-HTTP-Parser-XS-0.17
 perl-Import-Into-1.002005
 perl-indirect-0.38
-perl-Inline-0.80
-perl-Inline-Python-0.56
```
Martchus, line 117: This package should definitely be installed by `.circleci/build_dependencies.sh` as it installs `openQA-devel` which depends on `os-autoinst-devel` which depends on `perl(Inline::Python)`. According to the logs (https://app.circleci.com/pipelines/github/os-autoinst/openQA/7836/workflows/a650da46-bc3a-4935-a9ba-a92c9f238c99/jobs/74016)  `openQA-devel` and `os-autoinst-devel` are actually installed but not `perl(Inline::Python)`.

okurz: so is os-autoinst-devel broken now?

Martchus: I should have mentioned that `zypper info --requires os-autoinst-devel` lists `perl(Inline::Python)` on a Leap 15.2 system. That's why I didn't come to that conclusion.

Martchus: I've been creating https://github.com/os-autoinst/openQA/pull/4239 but I'm not sure whether it will help.

okurz:
https://github.com/os-autoinst/openQA/pull/4239

PR #4237, opened 2021-09-27
Don't do 'rpm -i' when there is no package in cache

Amrysliu:
Relate: https://progress.opensuse.org/issues/98496

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -51,7 +51,8 @@ aliases:
         if [ -z "$CIRCLE_WORKFLOW_ID" ]; then
           bash .circleci/build_cache.sh
         else
-          sudo rpm -i -f $(find /var/cache/zypp/packages/ | grep '.rpm$')
+          rpms=$(find /var/cache/zypp/packages/ | grep '.rpm$')
+          [ -z "$rpms" ] || sudo rpm -i -f "$rpms"
         fi
 
   - &test_junit


Amrysliu:
Not sure if it's the right way for [poo#98469](https://progress.opensuse.org/issues/98496). Please help to confirm that.

okurz:
Can you say what you see as benefit of this approach? In the end we need the packages installed anyway.

Amrysliu:
> Can you say what you see as benefit of this approach? In the end we need the packages installed anyway.

According to the ticket's description,  the nightly CI will fail when there is no rpm in `/var/cache/zypp/packages/`.
My thought is to avoid this scenario.
BTW if no rpm in cache directory, how to get them and install them? I checked other nightly CI process, not all rpms (listed in ci-packages.txt) are stored in cache directory.

okurz:
> According to the ticket's description, the nightly CI will fail when there is no rpm in `/var/cache/zypp/packages/`.
> My thought is to avoid this scenario.
> BTW if no rpm in cache directory, how to get them and install them?

I don't know, haven't checked details. Maybe resort to zypper

okurz:
We discussed about this and found that my initial request was no good. We use the cached packages also to define our consistent, version specific set of packages that we want to rely on. Resorting to a generic zypper call on errors might work but defeat the purpose as we would not realize if there are problems on saving and loading the cache so eventually we would end up always downloading different versions with zypper on the fly which we wanted to avoid. For now I don't have any better idea than https://github.com/os-autoinst/openQA/pull/4238 for now, it's really not much.

Our next best bet would be to find the job on circleCI that actually failed to properly save the cache and prevent the error the next time.

PR #4239, opened 2021-09-27
Ensure nightly job for deps prefers openQA:devel

Martchus:
This might fix the problem of missing dependencies (see
https://github.com/os-autoinst/openQA/pull/4236#discussion_r716515707).

It is likely a good idea anyways to make it explicit that we need to
prefer packages from our devel repositories.


.circleci/build_dependencies.sh
```diff
@@ -29,8 +29,8 @@ listdeps() {
 
 listdeps > $DEPS_BEFORE
 
-sudo zypper ar -f https://download.opensuse.org/repositories/devel:openQA/openSUSE_Leap_15.2 devel_openQA
-sudo zypper ar -f https://download.opensuse.org/repositories/devel:/openQA:/Leap:/15.2/openSUSE_Leap_15.2 devel_openQA_Leap
+sudo zypper ar --priority 90 -f https://download.opensuse.org/repositories/devel:openQA/openSUSE_Leap_15.2 devel_openQA
+sudo zypper ar --priority 91 -f https://download.opensuse.org/repositories/devel:/openQA:/Leap:/15.2/openSUSE_Leap_15.2 devel_openQA_Leap
```
okurz, line 33: But this is the wrong way around. The version specific Leap subrepos should be preferred in case they provide version specific overrides

okurz (changes requested):

.circleci/build_dependencies.sh
Martchus: That's true. I'll fix it.

Martchus:
By the way, it looks like this could do the trick. The CI log from the nightly job contains

```
Retrieving: os-autoinst-devel-4.6.1589880721.55a008dd-lp152.1.13.x86_64.rpm [starting]Retrieving: os-autoinst-devel-4.6.1589880721.55a008dd-lp152.1.13.x86_64.rpm [done]
```

and that's actually the "normal" Leap 15.2 version (https://build.opensuse.org/package/view_file/openSUSE:Leap:15.2/os-autoinst/os-autoinst.spec?expand=1).

okurz (approved):

PR #4245, opened 2021-09-28
Fix the needle dir when worker has cache diretory

Amrysliu:
Relate ticket: https://progress.opensuse.org/issues/99378


okurz (changes requested):
Please put more details in the git commit message details (not just the reference to the ticket). And there is a typo s/diretory/directory/

Amrysliu:
> Please put more details in the git commit message details (not just the reference to the ticket). And there is a typo s/diretory/directory/

Thanks. Added more details in commit message.

okurz (commented):
s/diretory/directory/

I did not understand the sentence "This is because we didn't receive the cache directory when count needle_dir." . What do you mean by "count"? Can you rephrase?

Martchus (commented):
Right, we should definitely pass through all the parameters. Maybe it makes also sense to just write it as:
```
sub needledir { productdir(@_) . '/needles' }
```

Amrysliu:
> s/diretory/directory/
> 
> I did not understand the sentence "This is because we didn't receive the cache directory when count needle_dir." . What do you mean by "count"? Can you rephrase?

I simplified the commit message. Hope it would be clear. :)

okurz (commented):
commit message is fine. As you try to fix an actual issue, can you add a test?

Martchus (approved):

Amrysliu:
> commit message is fine. As you try to fix an actual issue, can you add a test?



> commit message is fine. As you try to fix an actual issue, can you add a test?

Added

okurz (approved):

kalikiana (approved):

PR #4247, opened 2021-09-28
ci: Always checkout sha in build_autoinst.sh

kalikiana:
There shouldn't be a case where the sha deviates, and if there's a case for it let's make it explicit.

See #4244 for an example where the sha's inexplicably wrong.


okurz (approved):

.circleci/build_autoinst.sh
```diff
@@ -18,12 +18,12 @@
...
 
 destdir=${1:-../os-autoinst}
-sha=${2}
+sha=${2:-./autoinst.sha}
 
 echo Building os-autoinst $destdir $sha
 git clone https://github.com/os-autoinst/os-autoinst.git "$destdir"
 ( cd "$destdir"
-[ -z "$sha" ] || git checkout $sha
+git checkout $sha
```
Martchus, line 26: Looks like `$sha` is possibly a file path. So I'm not sure whether this will work. I'd also quote the parameter.

perlpunk:
> See #4244 for an example where the sha's inexplicably wrong.

I don't understand where there is something wrong there. Can you link to a CircleCI line where the sha is shown and wrong?

.circleci/build_autoinst.sh
```diff
@@ -18,12 +18,12 @@
 set -ex
 
 destdir=${1:-../os-autoinst}
-sha=${2}
+sha=${2:-./autoinst.sha}
```
perlpunk, line 21: sha is supposed to contain a shasum, not a filename. did you mean:
```
sha=${2:-$(cat ./autoinst.sha)}
```
?

perlpunk (changes requested):

.circleci/build_autoinst.sh
kalikiana: Indeed, that was a thinko on my part. Thanks!

perlpunk (changes requested):
I still don't understand the purpose of this PR

kalikiana:
> I still don't understand the purpose of this PR

The sha in #4244 was wrong. I don't fully understand why. This PR makes specifying the sha mandatory, which rules out at least one potential cause.

perlpunk:
> > I still don't understand the purpose of this PR
> 
> The sha in #4244 was wrong. I don't fully understand why. This PR makes specifying the sha mandatory, which rules out at least one potential cause.

But I asked you for a line showing that "the sha" was wrong.

In #4244 (the dependency PR) apparently the wrong os-autoinst-devel version was installed (and the version has a sha in it), which has nothing to do with the second argument in `build_autoinst.sh`.
Can you show where `build_autoinst.sh` is used in the dependency PR and show where exactly the sha was wrong?

kalikiana:
> rules out at least one potential cause.

No. Like I said, this `rules out at least one potential cause`. The point is that we don't have to doubt this code path next time the question "was the checksum used" is asked.

.circleci/build_autoinst.sh
```diff
@@ -18,12 +18,12 @@
 set -ex
 
 destdir=${1:-../os-autoinst}
-sha=${2}
+sha=${2:-$(cat ./autoinst.sha)}
```
Amrysliu, line 21: should that be `sha=${2:-$(cat .circleci/autoinst.sha)}` ? Not sure if the working directory is in `.circleci` or out of `.circleci/`, I see the config.yml means out it.

kalikiana: I changed it now to not specify it in `.circleci/config.yml` and assume we're at the toplevel.

PR #4248, opened 2021-09-28
OpenQA::Log: Simplify

okurz:


lib/OpenQA/Log.pm
```diff
@@ -52,23 +51,23 @@ my %LOG_DEFAULTS = (LOG_TO_STANDARD_CHANNEL => 1, CHANNELS => []);
 #  log_debug("message");
 #  log_debug("message", channels=>'channel1')
 #  log_debug("message", channels=>'channel1', standard=>0)
-sub log_debug { _log_msg('debug', @_); }
+sub log_debug(@) { _log_msg('debug', @_); }
```
Martchus, line 54: I assume this kind of syntax works (never used it myself before).

Martchus (approved):

lib/OpenQA/Log.pm
okurz: yes, see https://docs.mojolicious.org/perlsub#Signatures about "slurpy parameter"

okurz:
Updated: Fixed "Too few arguments for subroutine 'OpenQA::Log::setup_log' at /home/squamata/project/lib/OpenQA/WebAPI.pm line 55"

perlpunk:
> Updated: Fixed "Too few arguments for subroutine 'OpenQA::Log::setup_log' at /home/squamata/project/lib/OpenQA/WebAPI.pm line 55"

doesn't look like. maybe you forgot to commit something?

perlpunk (approved):

Martchus (approved):

okurz:
oops, I should have asked you if you agree with my later change which is introducing the use of Time::Moment. I would still welcome your feedback after the fact now

kalikiana:
> oops, I should have asked you if you agree with my later change which is introducing the use of Time::Moment. I would still welcome your feedback after the fact now

I had only seen the proof of concept with even worse performance before, So I'm kind of amazed you found an alternative afterall 🤓️ Neither the blog post nor the docs tell me why `Time::Moment` is better or how it is different, except that somehow it's insanely fast. So long as we get the same result, but loads faster I'm okay with that, though.
Actually the hidden bugfix `.0513` vs `.513` might be worth highlighting here.
The two bumps of the timeouts in `25-cache-service.t` and `40-script_openqa-clone-custom-git-refspec.t` would have made me suspicious as a reviewer, considering the Pr promises a 43% performance gain - I assume this was discussed?

okurz:
> The two bumps of the timeouts in `25-cache-service.t` and `40-script_openqa-clone-custom-git-refspec.t` would have made me suspicious as a reviewer, considering the Pr promises a 43% performance gain - I assume this was discussed?

These are only remotely related because after some longer times I ran all openQA tests locally finding these failing. I would say has nothing to do with logging.

PR #4249, opened 2021-09-28
Avoid log syscall overhead with vanilla Mojo::Log format

okurz:
During asset cleanup there are often syscalls accessing /etc/localtime.
We realized that this comes from our custom log format callback.

In
https://github.com/os-autoinst/openQA/blob/master/lib/OpenQA/Log.pm#L178
and
https://github.com/os-autoinst/openQA/blob/master/lib/OpenQA/Log.pm#L180
we call gettimeofday and localtime a lot.

I see no good reason – at least not anymore – why we should need a
custom log format. Instead we should be ok to rely on the vanilla
Mojo::Log format which provides sufficient time resolution as well
output of the log level and PID of the current process.

Related progress issue: https://progress.opensuse.org/issues/99426

### DIFF
--- a/lib/OpenQA/Log.pm
+++ b/lib/OpenQA/Log.pm
@@ -1,4 +1,4 @@
-# Copyright (C) 2020 SUSE LLC
+# Copyright (C) 2020-2021 SUSE LLC
 #
 # This program is free software; you can redistribute it and/or modify
 # it under the terms of the GNU General Public License as published by
@@ -14,8 +14,7 @@
 
 package OpenQA::Log;
 
-use strict;
-use warnings;
+use Mojo::Base -strict, -signatures;
 
 use Carp;
 use Exporter 'import';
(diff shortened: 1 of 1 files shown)


okurz:
I hoped upstream Mojo::Log might be faster but it looks like it's the worst. From my benchmarking:

```
                                      Rate Mojo::Log print with gettimeofday+localtime print with localtime plain print
Mojo::Log                          23810/s        --                              -60%                 -71%        -93%
print with gettimeofday+localtime  58824/s      147%                                --                 -29%        -82%
print with localtime               83333/s      250%                               42%                   --        -75%
plain print                       333333/s     1300%                              467%                 300%          --
```

Martchus:
Good that you've been benchmarking it. Maybe we better optimize our own version then so at least the repeated `stat` syscalls for the timezone config and `getpid` are avoided.

okurz:
obsoleted by https://github.com/os-autoinst/openQA/pull/4248

PR #4250, opened 2021-09-28
Fix SQL error in `refresh_assets` caused by 5b14fe

Martchus:
Otherwise the asset cleanup fails due to an SQL error:

```
DBIx::Class::Storage::DBI::_dbh_execute(): DBI Exception: DBD::Pg::st execute failed: ERROR:  invalid input syntax for type boolean: "" [for Statement "UPDATE assets SET fixed = ?, t_updated = ? WHERE id = ?" with ParamValues: 1='', 2='2021-09-28 15:01:21', 3='27487345'] at /usr/share/openqa/script/../lib/OpenQA/Schema/ResultSet/Assets.pm line 99
```

Apparently DBIx cannot cope with the boolean here so one
has to convert to int manually.

See https://progress.opensuse.org/issues/97979#note-26

### DIFF
--- a/lib/OpenQA/Schema/ResultSet/Assets.pm
+++ b/lib/OpenQA/Schema/ResultSet/Assets.pm
@@ -95,7 +95,7 @@ sub refresh_assets {
     my ($self) = @_;
 
     while (my $asset = $self->next) {
-        my $is_fixed = $asset->is_fixed;
+        my $is_fixed = $asset->is_fixed ? 1 : 0;
         $asset->update({fixed => $is_fixed}) if $is_fixed != $asset->fixed;
 
         $asset->refresh_size;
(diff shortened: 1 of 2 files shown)


lib/OpenQA/Schema/ResultSet/Assets.pm
```diff
@@ -95,7 +95,7 @@ sub refresh_assets {
     my ($self) = @_;
 
     while (my $asset = $self->next) {
-        my $is_fixed = $asset->is_fixed;
+        my $is_fixed = $asset->is_fixed ? 1 : 0;
```
perlpunk, line 98: Maybe this should be changed in the `sub is_fixed` directly?

Martchus: It should only be required here but I can also change `is_fixed` itself.

perlpunk (approved):

Amrysliu (approved):

okurz (approved):

PR #4251, opened 2021-09-29
Dependency cron 2021-09-29

openqabot:


okurz (changes requested):
not fixed, same as in https://github.com/os-autoinst/openQA/pull/4244

okurz:
see https://progress.opensuse.org/issues/98952

PR #4253, opened 2021-09-29
Use general 'container' or 'podman' where applicable

okurz:
* Generalize containerized setup to refer to 'container' instead of docker
* Generalize Makefile and scripts to refer to 'container' instead of docker
* CI: Generalize scripts to refer to 'container' instead of docker
* Use simpler 'podman' as drop-in replacement for our checks


kalikiana (approved):
I like it. Podman is also what I use.

Martchus (approved):
I haven't approved this PR so far because in my opinion `podman` is not more generic than `docker`. But I don't really care and the places where it is replaced by `container` are at least more generic.

PR #4254, opened 2021-09-30
tools/tidy: Disable aligning assignments

pevik:
See https://github.com/os-autoinst/os-autoinst/pull/1787


okurz:
Also here, let's give enough time to be seen by multiple persons

pevik:
Rebased to current master (29ca3aee891b2949837ce7b55d03620bd8e72f93).

okurz:
can you please check why a net 300 lines are added?

perlpunk:
We're dealing with the third problem today (after a broken chromium version and a problem with download.suse.org), fetching images from registry.opensuse.org times out :(
I'm constantly trying to restart jobs to finally get this in

PR #4258, opened 2021-10-01
Upgrade CI container image versions to Leap 15.3

Martchus:
* Base containers for CircleCI jobs on openSUSE Leap 15.3
* Keep other containers as-is
* See https://progress.opensuse.org/issues/99240

### DIFF
--- a/container/ci/Dockerfile
+++ b/container/ci/Dockerfile
@@ -1,11 +1,11 @@
 #!BuildTag: openqa_dev
-FROM opensuse/leap:15.2
+FROM opensuse/leap:15.3
 
 # Define environment variable
 ENV NAME openQA test environment
 ENV LANG en_US.UTF-8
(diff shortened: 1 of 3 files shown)


okurz (commented):
Can work but maybe it's better if we update first the files that are used in OBS and then the ones where we rely on the OBS built images.

okurz (changes requested):
yeah, so "cache" failed because we don't have proper Leap 15.3 base images. I suggest to update those first in a separate PR.

Martchus:
Makes sense. So I reduced this commit to Dockerfiles. (The follow-up commit is still on https://github.com/Martchus/openQA/pull/new/update-ci-2.)

perlpunk (approved):

okurz (approved):

PR #4260, opened 2021-10-02
Add missing service to openQA database migration guide

mimi1vx:


Martchus (approved):

docs/Installing.asciidoc
```diff
@@ -1126,7 +1126,7 @@ sudo ln --force --no-dereference --relative --symbolic /var/lib/pgsql/data.$newv
 [source,sh]
 ----
 sudo systemctl start postgresql
-sudo systemctl start openqa-webui openqa-scheduler openqa-livehandler openqa-gru
+sudo systemctl start openqa-webui openqa-scheduler openqa-livehandler openqa-gru openqa-websockets
```
okurz, line 1129: At least for starting this should be pulled in as dependency

okurz (changes requested):

kalikiana:
Can we have description, please?

docs/Installing.asciidoc
Martchus: True. This would only leave `openqa-webui` and `openqa-scheduler` for starting. However, for stopping it is still required. I've just tested that.

Martchus: By the way, it would be nice if you would put the web socket server directly after `openqa-webui` (for stopping).

kalikiana:
@mimi1vx Are you aware this is waiting on you with open questions?

docs/Installing.asciidoc
okurz: Yesterday I followed the database upgrade guide on o3 and there was no need to stop or start websockets explicitly

mimi1vx: @okurz it is possible, when I upgraded osd from PostgreSQL 10 to 12 , all looked fine, but websockets service started to throws errors. And explicit restart fixed this.

I think there is better to be safe, than wait for result.

okurz: Yes, I agree. Your approach is safer :)

okurz (approved):

Martchus (approved):

okurz:
@mergify rebase

PR #4261, opened 2021-10-04
Prevent useless downloads of uefi-vars assets

Martchus:
See particular commit messages and https://progress.opensuse.org/issues/99672 and https://progress.opensuse.org/issues/98388.


lib/OpenQA/Script/CloneJob.pm
```diff
@@ -100,12 +113,21 @@ sub clone_job_download_assets ($jobid, $job, $remote, $remote_url, $ua, $options
...
             # the generation job or if the only cloned job *is* the generation
             # job.
-            my $nr_parents = @parents;
+            my $nr_parents = @$parents;
             if ((!$options->{'skip-deps'} && !$options->{'skip-chained-deps'}) || ($nr_parents == 0)) {
-                for my $j (@parents, $job) {
-                    next ASSET if $j->{settings}->{PUBLISH_HDD_1} && $file eq $j->{settings}->{PUBLISH_HDD_1};
+                for my $j (@$parents, $job) {
+                    for my $setting (qw(PUBLISH_HDD_1 PUBLISH_PFLASH_VARS)) {
+                        next ASSET if _job_setting_is $j, $setting => $file;
```
okurz, line 120: I suggest you extract some methods to reduce the indentation

lib/OpenQA/Script/CloneJob.pm
```diff
@@ -100,12 +113,21 @@ sub clone_job_download_assets ($jobid, $job, $remote, $remote_url, $ua, $options
...
+                    }
                 }
             }
+            # skip downloading "usef-vars" assets if not actually generated by
```
okurz, line 124: Typo "usef"

okurz (changes requested):

okurz:
Mind the decreased coverage in t/35-script_clone_job.t which should have 100% statement coverage as it's a test

lib/OpenQA/Script/CloneJob.pm
```diff
@@ -90,21 +90,45 @@ sub clone_job_get_job ($jobid, $remote, $remote_url, $options) {
...
+}
+
+sub _is_asset_generated_by_cloned_jobs ($job, $parents, $file, $options) {
+    return 0 if ($options->{'skip-deps'} || $options->{'skip-chained-deps'}) && (scalar @$parents != 0);
+    for my $j (@$parents, $job) {
+        for my $setting (qw(PUBLISH_HDD_1 PUBLISH_PFLASH_VARS)) {
+            return 1 if _job_setting_is $j, $setting => $file;
+        }
+    }
+    return 0;
```
okurz, line 113: Would be nice to combine grep+map here, something like the following maybe?

```suggestion
    grep { _job_setting_is $_, $setting => $file } …
```

lib/OpenQA/Script/CloneJob.pm
```diff
@@ -90,21 +90,45 @@ sub clone_job_get_job ($jobid, $remote, $remote_url, $options) {
...
-                for my $j (@parents, $job) {
-                    next ASSET if $j->{settings}->{PUBLISH_HDD_1} && $file eq $j->{settings}->{PUBLISH_HDD_1};
-                }
+            # skip downloading published assets if we are also cloning the generation job or
+            # if the only cloned job *is* the generation job
+            next ASSET if _is_asset_generated_by_cloned_jobs $job, $parents, $file, $options;
+            # skip downloading "uefi-vars" assets if not actually generated by
+            # any parent
+            if ($file =~ qr/uefi-vars/) {
+                my $parent_publishes_pflash_vars;
```
okurz, line 129: why declare the variable here and not immediately assign? If this is just to prevent line-break then maybe you need to extract more functions to reduce indentation further :)

okurz (changes requested):

lib/OpenQA/Script/CloneJob.pm
Martchus: But the early return makes sense and the for loops are quite easy to read (compared to squashing too much logic into one line).

lib/OpenQA/Script/CloneJob.pm
okurz: I don't understand. To me it looks like you can combine this with the next line as you unconditionally assign a value there, don't you?

okurz (approved):

lib/OpenQA/Script/CloneJob.pm
Martchus: The next line is a for loop and the lifetime of the variable is outside of that loop. Even if I could get away with it in Perl this seems conceptionally wrong to me.

okurz (approved):

mimi1vx (approved):

PR #4263, opened 2021-10-05
[WDYT] Use Number::Bytes::Human instead of own implementation

okurz:


Martchus:
I guess it is ok to add an extra dependency for this. However, judging by https://metacpan.org/pod/Number::Bytes::Human#SYNOPSIS this is using e.g. MB and not MiB. It also lacks the space between the number and the unit which is strictly the correct way of formatting (e.g. also used on Wikipedia and KDE applications).

okurz:
> I guess it is ok to add an extra dependency for this. However, judging by https://metacpan.org/pod/Number::Bytes::Human#SYNOPSIS this is using e.g. MB and not MiB

It uses "MiB" with the "si" option that I applied

> It also lacks the space between the number and the unit which is strictly the correct way of formatting (e.g. also used on Wikipedia and KDE applications).

Right. That should disqualify the use. I have reported https://github.com/aferreira/cpan-Number-Bytes-Human/issues/1 and will close this PR

PR #4265, opened 2021-10-05
Use consistent "Copyright" header without problematic characters

okurz:
Same as done for os-autoinst in https://github.com/os-autoinst/os-autoinst/pull/1799

https://en.wikipedia.org/wiki/Copyright_symbol
explains that (c) can be used as a commonly accepted alternative to the
non-ASCII character previously used which can sometimes (still) cause
problems which we can avoid.

According to https://stackoverflow.com/a/221380 even the character
itself is seen as redundant as a repetition of the word "Copyright".
Also in the United States, the Berne Convention Implementation Act of
1988, effective March 1, 1989, removed the requirement for the copyright
symbol from U.S. copyright law so it is not necessary after all.

So we could have selected "(c)" or treat all files as unicode and
convert any ISO8859-1 copyright characters to the according unicode
character symbol or just get rid of the character at all.

This commit removes all variants of copyright characters in our source
files excluding binary files, full-text license files or external
references, i.e. os-autoinst-common.

I used the command

```
git grep -l 'Copyright' | xargs sed -i 's/Copyright \((C)\|(c)\|©\)/Copyright/g'
```

and manually reverted the changes in places where the change was
unintended as stated above.

This helps to not upset TidyAll and also seems to be a more consistent
standard among source files in multiple projects.

Related progress issue: https://progress.opensuse.org/issues/98616

### DIFF
--- a/.circleci/build_autoinst.sh
+++ b/.circleci/build_autoinst.sh
@@ -1,6 +1,6 @@
 #!/bin/bash
 #
-# Copyright (C) 2019-2020 SUSE LLC
+# Copyright 2019-2020 SUSE LLC
 #
 # This program is free software; you can redistribute it and/or modify
 # it under the terms of the GNU General Public License as published by
--- a/.circleci/build_cache.sh
+++ b/.circleci/build_cache.sh
@@ -1,6 +1,6 @@
 #!/bin/bash
 #
-# Copyright (C) 2019-2020 SUSE LLC
+# Copyright 2019-2020 SUSE LLC
 #
 # This program is free software; you can redistribute it and/or modify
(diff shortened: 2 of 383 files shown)


okurz:
meh, fails now in cache with

Package 'perl-Furl-3.14' not found.
Package 'perl-Selenium-Remote-Driver-1.41' not found.
Package 'perl-Test-Warnings-0.031' not found

I think this is because of the switch from Leap 15.2 to 15.3 . Leap 15.3 has Furl-3.13

Martchus (approved):

mimi1vx (approved):

PR #4268, opened 2021-10-05
Fix paths for `openqa_dev` container image

Martchus:
It was actually using Leap 15.2 before so I'm wondering why the previous
path was not `…/containers15.2/…`.

The previous path is `…/containers/…` which would be for Tumbleweed based
images but judging by 09d035242e2eabf4f8b02d9dae85b696fc35da8f I assume we
actually used Leap 15.2 before.

### DIFF
--- a/container/openqa/Dockerfile
+++ b/container/openqa/Dockerfile
@@ -1,5 +1,5 @@
 # hadolint ignore=DL3007
-FROM registry.opensuse.org/devel/openqa/containers/openqa_dev:latest
+FROM registry.opensuse.org/devel/openqa/containers15.3/openqa_dev:latest
 ENV LANG en_US.UTF-8
 
 COPY entrypoint.sh /usr/bin/entrypoint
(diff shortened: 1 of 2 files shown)


okurz (changes requested):
can we please merge https://github.com/os-autoinst/openQA/pull/4253 first?

okurz (changes requested):
Maybe I made a mistake in devel:openQA by changing the project config recently to fix one of the problems we had with updating the cache with the right packages. I would prefer if we don't need to update the paths in source every year. So better if we either run all that is necessary on Tumbleweed or we update devel:openQA to supply whatever is the most recent released Leap in "containers"

Martchus:
> I would prefer if we don't need to update the paths in source every year. So better if we either run all that is necessary on Tumbleweed or we update devel:openQA to supply whatever is the most recent released Leap in "containers"

This makes sense. So I would choose the last option then.

Martchus:
I've been changing the project config accordingly. (I kept `containers15.2` because it was still used in some home project.)

PR #4271, opened 2021-10-05
Allow more shells in apparmor profile

perlpunk:
Issue: https://progress.opensuse.org/issues/99195

In Leap 15.2, /bin/sh points to /bin/bash, while in 15.3,
it points to /usr/bin/sh -> /usr/bin/bash

Using the opportunity to also allow dash.


okurz:
merging because tests fail due to unrelated problems and apparmor profiles are not tested by CI tests anyway

PR #4273, opened 2021-10-06
Allow hxselect in apparmor profile

perlpunk:
Issue: https://progress.opensuse.org/issues/99195


kalikiana (approved):
I'm gonna assume this works based on conversations in Slack and [poo#99195](https://progress.opensuse.org/issues/99195).

PR #4274, opened 2021-10-06
Run hook script timeout in verbose mode

perlpunk:
So we actually get something in stderr in case it times out

Issues: https://progress.opensuse.org/issues/99519, https://progress.opensuse.org/issues/99741


perlpunk:
Fixed typo in commit message/description

okurz (approved):

PR #4276, opened 2021-10-06
Fix race-condition in `t/05-scheduler-full.t` which can lead to failures

Martchus:
See https://progress.opensuse.org/issues/98898 and particular commit messages


okurz (approved):
After this change you might want to reduce timeouts again maybe?

okurz (changes requested):
the code coverage warning in https://github.com/os-autoinst/openQA/pull/4276/files#diff-150605f30628d39e6c9a7316cd78f4c876fe3be95dee857b9b395100b5d5ea8cR620 should be covered

PR #4280, opened 2021-10-08
.mergify.yml: Make rules stricter requiring patch coverage

okurz:
https://progress.opensuse.org/issues/98448


okurz:
Well, there is https://github.com/os-autoinst/os-autoinst/pull/1788/files to allow the format used here and https://github.com/os-autoinst/os-autoinst-common/pull/13 for os-autoinst-common. We should apply the same here for consistency

PR #4286, opened 2021-10-08
fullstack: Make find_status_text wait for the element

kalikiana:
Fixes the following error:

    findElement: no such element: Unable to locate element: {"method":"css
    selector","selector":"#info_box .card-body"}`


kalikiana:
> Overall the changes look like they do what the git commit message says but I am not convinced that we should do this waiting in all the places. I work prefer if we only do it where we actually understood that we need it. So I suggest to do the replacement only where tests actually fail and if we understood the reason in each case

So after suggesting to rename `find_status_text` you're suggesting we check differently on a case-by-case basis? I guess I will kill the function and see what happens

PR #4288, opened 2021-10-11
Fail on unexpected Javascript console errors

kalikiana:
The function merely returns a truthy value. Let's fail always without relying on the caller. This means we will see errors in all cases, not just when something else happens to fail.

I'm also dropping a comment claiming the function aborts.

### DIFF
--- a/t/full-stack.t
+++ b/t/full-stack.t
@@ -135,7 +135,7 @@ $driver->find_element_by_link_text('core@coolone')->click();
 $driver->title_is("openQA: $job_name test results", 'scheduled test page');
 my $job_page_url = $driver->get_current_url();
 like(status_text, qr/State: scheduled/, 'test 1 is scheduled');
-ok javascript_console_has_no_warnings_or_errors, 'no javascript warnings or errors after test 1 was scheduled';
+ok(javascript_console_has_no_warnings_or_errors(), 'no unexpected js warnings after test 1 was scheduled');
 
 sub assign_jobs ($worker_class = undef) {
     check_scheduled_job_and_wait_for_free_worker $worker_class // 'qemu_i386';
@@ -257,7 +257,7 @@ $job_name = 'tinycore-1-flavor-i386-Build1-core@noassets';
 $driver->title_is("openQA: $job_name test results", 'scheduled test page');
 like status_text, qr/State: scheduled/, 'test 4 is scheduled';
 
-ok javascript_console_has_no_warnings_or_errors, 'no javascript warnings or errors after test 4 was scheduled';
+ok(javascript_console_has_no_warnings_or_errors(), 'no unexpected js warnings after test 4 was scheduled');
 start_worker_and_assign_jobs;
 
 subtest 'incomplete job because of setup failure' => sub {
--- a/t/lib/OpenQA/SeleniumTest.pm
+++ b/t/lib/OpenQA/SeleniumTest.pm
@@ -206,7 +206,16 @@ sub javascript_console_has_no_warnings_or_errors {
 
         my $source = $log_entry->{source};
         my $msg = $log_entry->{message};
-        if ($source eq 'network') {
+        if ($source eq 'javascript' || $source eq 'network') {
+            # ignore when the proxied ws connection is closed; connection errors are tracked via the devel console
+            # anyways and when the test execution is over this kind of error is expected
+            next if ($msg =~ qr/ws\-proxy.*Close received/);
+
+            # ignore "connection establishment" ws errors in ws_console.js; the ws server might just not be running yet
+            # and ws_console.js will retry
+            next if ($msg =~ qr/ws_console.*Error in connection establishment/);    # uncoverable statement
+        }
+        elsif ($source eq 'network') {
             # ignore errors when gravatar not found
             next if ($msg =~ qr/gravatar/);    # uncoverable statement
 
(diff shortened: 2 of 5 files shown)


Martchus:
The advantage of this approach is that we cannot forget to actually use the return value but I also get @okurz 's point. It would be great if Perl had a `[[nodiscard]]` attribute like C++ but it seems like it hasn't.

t/lib/OpenQA/Test/FullstackUtils.pm
```diff
@@ -80,8 +80,8 @@ sub find_status_text { shift->find_element('#info_box .card-body')->get_text() }
 sub _fail_with_result_panel_contents {
     my ($result_panel_contents, $msg) = @_;    # uncoverable statement
     diag("full result panel contents:\n$result_panel_contents");    # uncoverable statement
-    javascript_console_has_no_warnings_or_errors;    # uncoverable statement
-    fail $msg;    # uncoverable statement
+    ok(javascript_console_has_no_warnings_or_errors, 'No unexected js warnings');    # uncoverable statement
```
okurz, line 83: s/unexected/unexpected/

t/lib/OpenQA/Test/FullstackUtils.pm
```diff
@@ -104,7 +104,7 @@ sub wait_for_result_panel {
             diag("stopped waiting for '$result_panel', result turned out to be '$1'");    # uncoverable statement
             return _fail_with_result_panel_contents($status_text, $msg);    # uncoverable statement
         }
-        javascript_console_has_no_warnings_or_errors;
+        ok(javascript_console_has_no_warnings_or_errors, 'No unexected js warnings');
```
okurz, line 107: s/unexected/unexpected/

But also the problem I see that you use `ok` within a helper function scope so we don't know where this came from. I suggest to handle that with a die

okurz (changes requested):

Martchus (approved):

t/lib/OpenQA/Test/FullstackUtils.pm
kalikiana: Considering this fatal here as well seems sensible, I'd rather use `_fail_with_result_panel_contents`, though, since that's how the function fails in all other cases.

t/full-stack.t
```diff
@@ -135,7 +135,7 @@ $driver->find_element_by_link_text('core@coolone')->click();
 $driver->title_is("openQA: $job_name test results", 'scheduled test page');
 my $job_page_url = $driver->get_current_url();
 like(status_text, qr/State: scheduled/, 'test 1 is scheduled');
-ok javascript_console_has_no_warnings_or_errors, 'no javascript warnings or errors after test 1 was scheduled';
+ok(javascript_console_has_no_warnings_or_errors(), 'no unexpected js warnings after test 1 was scheduled');
```
okurz, line 138: are the outer parentheses needed? I prefer the style without

t/lib/OpenQA/Test/FullstackUtils.pm
```diff
@@ -163,7 +163,7 @@ sub wait_for_developer_console_like {
         }
 
         wait_for_ajax(msg => $message_regex);
-        javascript_console_has_no_warnings_or_errors($js_erro_check_suffix) or return;
+        ok(javascript_console_has_no_warnings_or_errors($js_erro_check_suffix), 'No unexpected js warnings') or return;
```
okurz, line 166: here I still see the problem of using Test::More calls in a sub-function where the reporting scope doesn't fit. So better avoid "ok" and other functions like that. Can you just call "die" or simply use the return?

okurz (changes requested):

t/full-stack.t
kalikiana: The lack of parantheses were the reason the code was broken in an ealier version. And since I can't read it properly I'm leaving them

t/lib/OpenQA/Test/FullstackUtils.pm
kalikiana: This function already relies on `pass` and `fail`. That said I could leave this to `return` only

t/lib/OpenQA/SeleniumTest.pm
```diff
@@ -219,12 +219,12 @@ sub javascript_console_has_no_warnings_or_errors {
...
             next if ($msg =~ qr/api\/v1\/exp.*\/job_templates_scheduling\/1003 - Failed.*/);    # uncoverable statement
-        }
-        elsif ($source eq 'javascript') {
+
             # ignore when the proxied ws connection is closed; connection errors are tracked via the devel console
             # anyways and when the test execution is over this kind of error is expected
             next if ($msg =~ qr/ws\-proxy.*Close received/);
-
+        }
+        elsif ($source eq 'javascript') {
```
kalikiana, line 227: I really wish it wasn't so extremely racy. I now managed to reproduce the message with the `javascript` source. So it seems we need it in both afterall. Even if you can easily never see it in a few hundred runs...

Makefile
```diff
@@ -177,7 +177,7 @@ test-unstable:
 
 .PHONY: test-fullstack
 test-fullstack:
-	$(MAKE) test-with-database FULLSTACK=1 TIMEOUT_M=30 RETRY=15 PROVE_ARGS="$$HARNESS t/full-stack.t"
+	$(MAKE) test-with-database FULLSTACK=1 TIMEOUT_M=300000 RETRY=50 STABILITY_TEST=1 PROVE_ARGS="$$HARNESS t/full-stack.t"
```
kalikiana, line 180: Adding `RETRY=50 STABILITY_TEST=1` here since the extra PR was making it more confusing

kalikiana: Well, it seems virtually impossible to get 50 runs without unrelated errors...

Martchus (approved):

t/full-stack.t
okurz: What do you mean with "can't read it properly"?

kalikiana: https://github.com/os-autoinst/openQA/pull/4288#discussion_r732920254

okurz (approved):

PR #4293, opened 2021-10-12
selenium: No timeout for implicit waits by default

kalikiana:
- No timeout by default
- No disabling timeouts in individual tests


Martchus:
Scaling the timeout makes likely sense although we did not need it before. We just have to watch out that the timeout is disabled in all places where we check for the absence of an element. So it would make sense to check that the execution time of tests hasn't increased.

kalikiana:
Closing as per conversation this morning, and mentioned in [poo#98952](https://progress.opensuse.org/issues/98952) because this isn't immediately addressing specific failures

PR #4295, opened 2021-10-12
tests: Use separate databases for unstable tests

kalikiana:
This prevents failing and passing re-runs from being merged and allows individual databases to be inspected or dropped.

See: https://progress.opensuse.org/issues/99594

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -173,7 +173,7 @@ test-api:
 # put unstable tests in tools/unstable_tests.txt and uncomment in circle CI config to handle unstables with retries
 .PHONY: test-unstable
 test-unstable:
-	for f in $$(cat tools/unstable_tests.txt); do $(MAKE) test-with-database TIMEOUT_M=10 PROVE_ARGS="$$HARNESS $$f" RETRY=5 || exit; done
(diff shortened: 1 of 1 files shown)


perlpunk:
The problem is that we don't delete coverage data for failing tests, when we retry.
So, we have to delete coverage **if** we retry.
So far so good. We can do that in the retry script - simply delete the `cover_db_circleci_jobname` folder.
That I will do in my followup.


The problem with **unstable** is that we run `tools/retry` in a for loop for 3 tests.

All those three tests write their coverage in `cover_db_circleci_unstable`.

test1.t suceeds
test2.t fails, then thanks to retry suceeds.
test3.t suceeds

Now, if test2.t would delete `cover_db_circleci_unstable` after the fail, then the coverage data for test1.t would be deleted too. We don't want that.

This PR makes sure that in the case we run tools/retry in such a for loop every test file gets their own directory, e.g. `cover_db_jobname_test1.t` . So if that has to be deleted, the data from the **other** tests are **not** deleted.

Martchus:
Now I understand it but the change suggested by @perlpunk is still required.

PR #4296, opened 2021-10-12
Add possibility to override the job result with special comments

okurz:
When test results are incorporated in other systems and processes the job
result is important and the concept of having a job reviewed with bugref and
labels does not fit all processes. So this is why it can help if the job result
itself can be altered to reflect the actual expectation, for example to mark
false-positives.

Same as for example on o3 with totest-manager we would like to formally
communicate test failures that should be ignored for auto-approval case by
case, e.g. in test failures.

This commit adds parsing for a special comment label "force_result" which can
be used to write in openQA test comments, for example to set a failure to
"softfailed" with a reference to a bug.

The documentation and help text has already been extended in previous commits.

Related progress issue: https://progress.opensuse.org/issues/95479

### DIFF
--- a/lib/OpenQA/Schema/Result/Comments.pm
+++ b/lib/OpenQA/Schema/Result/Comments.pm
@@ -114,7 +114,7 @@ Returns new result value if C<$self> is a special "force_result" label, e.g.
 =cut
 
 sub force_result ($self) {
-    $self->label && $self->text =~ /force_result:(\w+):?(\w*)/;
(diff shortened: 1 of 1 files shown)


kalikiana:
> ok, do you propose alternatives? Prevent deletion or updates of such comments? Change another field on the job to keep that information? revert the job result back if the comment does not apply anymore?

Could we consider any label comment immutable? And if the state needs to be corrected, a new comment can be added. But we wouldn't have to worry about inconsistency.

PR #4299, opened 2021-10-13
t: Add a hook to tools/retry to delete coverage data

perlpunk:
Issue: https://progress.opensuse.org/issues/99594

Depends on #4295 to create individual cover_db directories per test in the `unstable` test


perlpunk:
retriggering CI, hoping to get a failure, to actually see the behaviour in action

perlpunk:
Removed the `-v` from `rm` because it's too noisy, and added an `echo`

PR #4303, opened 2021-10-13
Make text results rendering unified with other

mimi1vx:
https://progress.opensuse.org/issues/93940

example: http://quasar.suse.cz/tests/7288

it changes all *text* results processing from ajax to use `data-` attribute with exception `Soft Failed` result with bug references , because this part is only implemented in template and perl code


asdil12:
Maybe this is out of scope here but navigating text results on your instance, the fadein/fadeout effect is annoying as it *feels slow* and let's the box flicker as if there was an AJAX request in the background - which is not the case - so no reason for the user to feel slow.

mimi1vx:
in the end I decided to drop base64 and use `escape/unescape` conmbo ... it feels much faster

PR #4308, opened 2021-10-14
Add retry for publishing AMQP events

Martchus:
* See https://progress.opensuse.org/issues/99009
* Plus avoiding an escaping issue when making AMQP URL and some tidying, see individual commit message


AdamWill:
thanks a lot, i'll try and get this onto our deployments and keep an eye on it sometime soon. In F35 crunch mode right now...

okurz:
awesome. We actually found the errors in our logs as well so we might be able to verify this in production ourselves but I would be happy to hear if it works for you

PR #4319, opened 2021-10-20
t: Prevent aborting the complete test run when interactively debugging

okurz:
A `BAIL_OUT` would stop the complete test execution in UI tests when an
UI element is not found. This is not necessarily helpful when
interactively debugging with the Perl debugger so instead we can simply
fail the test in this case.


perlpunk:
The codecov report about lib/OpenQA/Worker/WebUIConnection.pm must be a bug.
Looking at https://app.codecov.io/gh/os-autoinst/openQA/compare/4319/changes#D1L185 it displays code which is different in master.

PR #4323, opened 2021-10-21
Update bootstrap and jquery

mimi1vx:
bootstrap to 4.6.1
jQuery to 3.6.0

First step for https://progress.opensuse.org/issues/101382


kalikiana:
Nice. Please mention **bootstrap v4.6.0** here rather than "last supported". And whatever jQuery you're updating to.

Also, did you use any particular commands to do the update? And, do you have any release notes?

mimi1vx:
reopening as blocker `bootstrap-your` was removed

Martchus:
If you'd like you can rebase our changes. Maybe in the meantime already a newer bootstrap version has been released so we might want to bump the version in another commit.

mimi1vx:
@Martchus yep, that's plan :D btw bootstrap5 migration will be more challenging :D

Martchus (approved):
I've gone though all pages locally and didn't find any problems. I also tested the tour, the needle editor, all tabs of the details page, some help popovers, and the developer mode/console. Let's see whether test will pass. (Of course the asset cache needs to be generated manually again.)

Martchus:
I thought on CircleCI it would just pull the new assets from the internet but apparently not. We could run tests just locally instead of course. Not sure whether that's generally acceptable, though.

Martchus:
I've run all tests locally and they've passed.

okurz (approved):

Martchus:
Then I'm going to force-merge now and I'll also update the asset cache for the rpm package.

PR #4336, opened 2021-10-26
Unschedule jobs which are too long in 'SCHEDULED' state

mimi1vx:
https://progress.opensuse.org/issues/100973

### DIFF
--- a/lib/OpenQA/Scheduler/Model/Jobs.pm
+++ b/lib/OpenQA/Scheduler/Model/Jobs.pm
@@ -364,6 +364,9 @@ sub _to_be_scheduled {
 
 sub _update_scheduled_jobs {
     my $self = shift;
+    my $cur_time = DateTime->now(time_zone => 'UTC');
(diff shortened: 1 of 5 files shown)


Martchus (dismissed):

lib/OpenQA/Scheduler/Model/Jobs.pm
```diff
@@ -364,6 +364,9 @@ sub _to_be_scheduled {
 
 sub _update_scheduled_jobs {
     my $self = shift;
+    my $cur_time = DateTime->now(time_zone => 'UTC');
+    my $max_job_scheduled_time = $self->{config}->{scheduler}->{max_job_scheduled_time}
```
perlpunk, line 368: This `$self->{config}` isn't actually working. I found that out as part of https://progress.opensuse.org/issues/129619 when I added a new scheduler config option and used the same code. It needs to be `OpenQA::App->singleton->config`

perlpunk: See #5287

PR #4337, opened 2021-10-26
Fix parent progressbar links

FruitFly638:
Progress bar links for builds with children were previously broken

https://progress.opensuse.org/issues/98403

### DIFF
--- a/t/ui/14-dashboard-parents.t
+++ b/t/ui/14-dashboard-parents.t
@@ -84,6 +84,15 @@ $driver->find_element_by_link_text('Build0091')->click();
 my $element = $driver->find_element_by_link_text('opensuse');
 ok($element->is_displayed(), 'link to child group expanded');
 $driver->find_element_by_link_text('Build0091')->click();
+
+subtest 'progress bar link' => sub {
+    my $first_build_row = $driver->find_elements('.build-row')->[0];
+    my $link = $driver->find_child_element($first_build_row, '.progress-bar-failed a');
+    is($link->get_text, '1 failed', 'link text');
+    my $r = qr|.*/tests/overview\?result=failed&distri=opensuse&version=Factory&build=87\.5011&groupid=1001&groupid=1002|;
+    like($link->get_attribute('href'), $r, 'link href');
+};
+
 # Looking for "is_hidden" does not turn out to be reliable so relying on xpath
 # lookup of collapsed entries instead
 ok($driver->find_element_by_xpath('//div[contains(@class,"children-collapsed")]//a'), 'link to child group collapsed');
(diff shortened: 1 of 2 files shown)


Martchus (commented):
Looks like it would work but tests (or extended checks on an existing test) would still be nice.

FruitFly638:
I am kind of stuck on this, because I still can't figure out how to find a specific URL in a test. I was looking at t/17-build-tagging, where we are doing something like `my @tags = $t->tx->res->dom->find('a[href^=/tests/]')->map('text')->each;` to find all links on the page and I wanted to do something similar in t/ui/14-dashboard-parents to check if there is a URL like `/tests/overview?result=failed&distri=opensuse&version=Factory&build=87.5011&groupid=1001&groupid=1002` and to make sure that it's not duplicated.

Why doesn't this approach work? How do I do this instead?

Martchus (approved):

kalikiana (approved):

t/ui/14-dashboard-parents.t
```diff
@@ -84,6 +84,18 @@ $driver->find_element_by_link_text('Build0091')->click();
...
+
+subtest 'progress bar link' => sub {
+    my $first_build_row = $driver->find_elements('.build-row')->[0];
+    my $link = $driver->find_child_element($first_build_row, '.progress-bar-failed a');
+    is($link->get_text, '1 failed', 'link text');
+    like(
+        $link->get_attribute('href'),
+        qr|.*/tests/overview\?result=failed&distri=opensuse&version=Factory&build=87\.5011&groupid=1001&groupid=1002|,
+        'link href'
+    );
```
okurz, line 96: I would prefer that to be a bit more concise. Also we don't need to check that much, just look for the result and multiple group ids, like so

```suggestion
    like $link->get_attribute('href'), qr|.*/tests/overview\?result=failed&.*&groupid=1001&groupid=1002|, 'link href';
```

okurz (changes requested):

t/ui/14-dashboard-parents.t
Martchus: To save a few lines:
```suggestion
    my $r = qr|.*/tests/overview\?result=failed&distri=opensuse&version=Factory&build=87\.5011&groupid=1001&groupid=1002|;
    like($link->get_attribute('href'), $r, 'link href');
```

okurz: Ok, then use the suggestion from @Martchus

FruitFly638:
Is everything okay now? I incorporated Martchus' changes.

Martchus:
Looks like tidy doesn't like it. Maybe the line is still too long. I needed to check how to break the regex into multiple lines myself. (I personally would have been just fine with the first version.)

perlpunk:
A regex can be broken into multiple lines when using the `/x` flag, e.g.
```
my $r = qr|.*/tests/overview\?
  result=failed&distri=opensuse&version=Factory&build=87\.5011&groupid=1001&groupid=1002|x;
```

okurz (approved):

PR #4345, opened 2021-11-02
Remove workaround for broken Chromium on Leap 15.3

Martchus:
* https://bugzilla.opensuse.org/show_bug.cgi?id=1191429 has been marked as
  fixed so the workaround should no longer be required
* See https://progress.opensuse.org/issues/100850

### DIFF
--- a/tools/ci/run_unit_tests.sh
+++ b/tools/ci/run_unit_tests.sh
@@ -1,9 +1,5 @@
 #!/bin/bash
 set -ex
-# downgrade chromedriver from problematic version to last good version, see https://progress.opensuse.org/issues/100850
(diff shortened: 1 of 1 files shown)


mimi1vx (approved):

kraih (approved):

okurz:
just for completeness: Normally a bug is marked as RESOLVED FIXED as soon as a (potential) bug fix is available as a submit request. This does not mean that at this point a bugfix is already accepted into the affected product. If however you can verify that the bugfix is actually effective in the targeted product please help to set the bug report to VERIFIED FIXED with an according comment

Martchus:
I've been updating the ticket.

PR #4346, opened 2021-11-03
DO NOT MERGE: Test stability of t/full-stack.t

Martchus:


Martchus:
It has already run 13 times consecutively successfully and only got aborted after 30 minutes because of the timeout parameter. So it looks good so far.

Martchus:
It failed on run 34 due to a timeout. Not sure whether that's good enough but it looks definitely better than before (where we had to increase the number of retries to 15).

PR #4349, opened 2021-11-05
clone job: Skip uefi-vars download unless parent job has setting `UEFI=1`

Martchus:
We already skip the download if `PUBLISH_PFLASH_VARS` is not present in any
parent job to avoid download errors for non-existing assets. However, it
seems some jobs have that variable set but still don't have `UEFI=1` and
are therefore not publishing a uefi-vars file.

This change is in accordance with the existing condition in os-autoinst's
function `handle_generated_assets`.


Martchus:
@asdil12 This fixes the download error you've mentioned in the chat.

asdil12 (approved):

okurz (approved):

PR #4351, opened 2021-11-07
Dependency cron 2021-11-07

openqabot:


Martchus:
Not sure whether our tests would find possible problems so maybe it makes sense to test manually whether it works.

okurz:
From OBS:


openSUSE_15.2
x86_64
unresolvable

The build can not begin, because required packages are either missing or not explicitly defined.
Build finished, but repository publishing is disabled
unresolvable: have choice for libpq.so.5()(64bit) needed by perl-DBD-Pg: libpq5 postgresql14-devel-mini, have choice for libpq.so.5()(64bit) needed by postgresql12-server: libpq5 postgresql14-devel-mini, have choice for libpq5 >= 12.6 needed by postgresql12-server: libpq5 postgresql14-devel-mini, have choice for libpq.so.5()(64bit) needed by postgresql12: libpq5 postgresql14-devel-mini, have choice for libpq5 >= 12.6 needed by postgresql12: libpq5 postgresql14-devel-mini

mimi1vx:
@okurz you add this in wrong place :(

kraih:
This is the version of `Mojolicious::Plugin::OAuth2` that brings OpenID Connect support. Which should allow us to retire the old custom OpenID code we currently use.

Martchus (approved):
I've just tested it locally and it still works (tested with GitHub).

PR #4352, opened 2021-11-09
ci: Optional check via #fullstack-stability comment

kalikiana:
- This is not expected to replace or even replicate our CircleCI setup,
  so no setup of coverage or caches is done here.
- This is only triggered if the special comment is used.
- Triggering arbitrary test targets is out of scope.

tl;dr; Add a comment instead of creating a PR like #4346.

I'm not sure a new workflow can be used in the same PR - I'll try anyway:

#fullstack-stability

### DIFF
--- a/.github/workflows/comment-trigger.yml
+++ b/.github/workflows/comment-trigger.yml
@@ -1,0 +1,33 @@
+---
+name: Trigger CI via comment
+on:
+  push:
+  pull_request:
+  workflow_dispatch:
+    inputs:
+      retry:
+        description: 'RETRY'
+        default: 50
+      target:
+        description: 'Makefile target'
+        default: fullstack
+
+jobs:
+  fullstack-stability:
+    runs-on: ubuntu-latest
+    container:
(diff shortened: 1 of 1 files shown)


perlpunk:
It would also be possible to add a workflow_dispatch event to a workflow.
e.g.:
```
on:
  push:
    branches: [ '*' ]
  pull_request:
    branches: [ master ]
  workflow_dispatch:
    branches: [ '*' ]
```
Then the workflow can be triggered in the Actions menu in the webui.

One can choose a branch which it should run on.

kalikiana:
🤔 
```bash
Deleting the contents of '/__w/openQA/openQA'
Initializing the repository
  /usr/bin/git init /__w/openQA/openQA
  /__w/openQA/openQA/.git: Permission denied
  Error: The process '/usr/bin/git' failed with exit code 1
```

okurz (changes requested):
fails with

```
  /usr/bin/git init /__w/openQA/openQA
  /__w/openQA/openQA/.git: Permission denied
  Error: The process '/usr/bin/git' failed with exit code 1
```

can you make this a draft?

PR #4354, opened 2021-11-10
Make logging when stopping command server gracefully less confusing

Martchus:
* Use only the debug level because for the user these messages are not
  important
* Clearly state that this is *not* the reason for the test failure
* Change phrasing to make it sound less severe

### DIFF
--- a/lib/OpenQA/Worker/Isotovideo/Client.pm
+++ b/lib/OpenQA/Worker/Isotovideo/Client.pm
@@ -5,7 +5,7 @@ package OpenQA::Worker::Isotovideo::Client;
 use Mojo::Base -base;
 
 use Mojo::UserAgent;
-use OpenQA::Log qw(log_info log_debug);
+use OpenQA::Log qw(log_debug);
 
 has job => undef, weak => 1;
 has ua => sub { Mojo::UserAgent->new };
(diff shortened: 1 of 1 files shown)


lib/OpenQA/Worker/Isotovideo/Client.pm
```diff
@@ -26,8 +26,8 @@ sub stop_gracefully {
 
             my $res = $tx->res;
             if (!$res->is_success) {
-                log_info('Unable to stop the command server gracefully: ');
-                log_info($res->code ? $res->to_string : 'Command server likely not reachable at all');
+                log_debug('Unable announce job termination (NOT the reason for the job termination):');
```
kalikiana, line 29:
```suggestion
                log_debug('Unable to announce job termination (NOT the reason for the job termination):');
```

lib/OpenQA/Worker/Isotovideo/Client.pm
```diff
@@ -1,11 +1,11 @@
-# Copyright 2019 SUSE LLC
+# Copyright 2019-2021 SUSE LLC
```
okurz, line 1: please leave the old. Our approach should be to not update Copyright anymore at all after the initial year

lib/OpenQA/Worker/Isotovideo/Client.pm
```diff
@@ -26,8 +26,8 @@ sub stop_gracefully {
 
             my $res = $tx->res;
             if (!$res->is_success) {
-                log_info('Unable to stop the command server gracefully: ');
-                log_info($res->code ? $res->to_string : 'Command server likely not reachable at all');
+                log_debug('Unable to announce job termination (NOT the reason for the job termination):');
+                log_debug($res->code ? $res->to_string : 'Command server has likely just already terminated');
```
okurz, line 30: s/has likely just already terminated/is likely finished already/?

okurz (changes requested):

okurz (approved):

perlpunk (approved):

PR #4356, opened 2021-11-11
doc: Consider nftables being used in modern systems

Martchus:
See https://progress.opensuse.org/issues/101950


kalikiana (approved):
Nice. I like it!

baierjan (approved):

PR #4357, opened 2021-11-12
Fix enabling systemd unit `openqa-reload-worker-auto-restart@.path`

Martchus:
Add an `[Install]` section to the unit. Otherwise, trying to enable it only
results in the message "The unit files have no installation config. … This
means they are not meant to be enabled using systemctl.". However, this
unit is actually meant to be enabled as implied by `Installing.asciidoc`.

### DIFF
--- a/systemd/openqa-reload-worker-auto-restart@.path
+++ b/systemd/openqa-reload-worker-auto-restart@.path
@@ -2,3 +2,6 @@
 
 [Path]
 PathChanged=/etc/openqa/workers.ini
+
+[Install]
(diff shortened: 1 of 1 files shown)


baierjan (approved):

Martchus:
The test failure must be completely unrelated but it still looks bad. After 11 retries (exceeding the per-test timeout) the CI run is aborted by CircleCI (due to exceeding the CircleCI timeout).

kalikiana (approved):

PR #4360, opened 2021-11-16
clone-job: Fix accidentally skipping asset downloads

Martchus:
When skipping an asset download, the clone-job script skipped all remaining
assets of the same type as well. This is wrong, e.g. if the uefi-vars asset
is skipped before the actual HDD image is considered the actual HDD image
would be skipped as well. With this change only the uefi-vars asset would
be skipped.

Note that this problem was made worse by e0bb6cb73 which added one more
case to skip assets.


BillAnastasiadis:
LGTM. The commit message fail probably comes from the checker not recognizing the tag as a tag, due to the `-` present. Maybe we should fix that in the checker.

kraih (approved):

BillAnastasiadis (approved):

Martchus:
Yes, I think we should fix it in the checker. Being able to use a prefix like this is quite useful. Especially if one creates multiple commits which are all about the same context it is much clearer to specify the context as prefix instead somehow make it clear within each individual commit message.

okurz (changes requested):
Can you fix the git commit message regex directly? Should be easy for you. Also the full stack test fails. Please look into that as well, bump retries to 1k or something more smart :)

perlpunk:
> Also the full stack test fails. Please look into that as well, bump retries to 1k or something more smart :)

Problem is that the retries aren't even used, it's stopped in the third one because a CircleCI limit itself is reached

perlpunk (approved):

okurz (approved):

PR #4362, opened 2021-11-16
Speed up 25-cache-service.t with a shorter worker timeout

kraih:
Turns out the test spends most of its time in a blocking SQL loop in `Minion::Backend::SQLite`.

Before:
```
All tests successful.
Files=1, Tests=23, 103 wallclock secs ( 0.07 usr  0.01 sys + 15.86 cusr  1.72 csys = 17.66 CPU)
Result: PASS
```

After:
```
All tests successful.
Files=1, Tests=23, 51 wallclock secs ( 0.05 usr  0.01 sys +  5.39 cusr  0.97 csys =  6.42 CPU)
Result: PASS
```

Progress: https://progress.opensuse.org/issues/102221

### DIFF
--- a/t/lib/OpenQA/Test/Utils.pm
+++ b/t/lib/OpenQA/Test/Utils.pm
@@ -93,7 +93,7 @@ sub cache_minion_worker {
             require OpenQA::CacheService;
             local $ENV{MOJO_MODE} = 'test';
             note('Starting cache minion worker');
-            OpenQA::CacheService::run(qw(run));
+            OpenQA::CacheService::run(qw(run --dequeue-timeout 1));
             note('Cache minion worker stopped');
             Devel::Cover::report() if Devel::Cover->can('report');
             _exit(0);
--- a/t/25-cache-service.t
+++ b/t/25-cache-service.t
@@ -37,7 +37,7 @@ use POSIX '_exit';
 use Mojo::IOLoop::ReadWriteProcess qw(queue process);
 use Mojo::IOLoop::ReadWriteProcess::Session 'session';
 use OpenQA::Test::Utils qw(fake_asset_server cache_minion_worker cache_worker_service wait_for_or_bail_out);
-use OpenQA::Test::TimeLimit '110';
+use OpenQA::Test::TimeLimit '90';
 use Mojo::Util qw(md5_sum);
 use OpenQA::CacheService;
 use OpenQA::CacheService::Request;


t/lib/OpenQA/Test/Utils.pm
```diff
@@ -93,7 +93,7 @@ sub cache_minion_worker {
             require OpenQA::CacheService;
             local $ENV{MOJO_MODE} = 'test';
             note('Starting cache minion worker');
-            OpenQA::CacheService::run(qw(run));
+            OpenQA::CacheService::run(qw(run --dequeue-timeout 1));
```
okurz, line 96: wow, what is that? Can we speedup everything by factor 10 by making this `.1`? :) You could go into more details in the git commit message. That would be nice.

kraih: Unfortunately there's a point where performance gets worse again (implementation detail upstream...). 😁

The Minion worker is designed as a daemon with a blocking mainloop that primarily blocks when waiting for new jobs to be dequeued. By default `Minion::Backend::SQLite` blocks for 5 seconds before handling control back to `Minion::Worker`, where it does housekeeping stuff like reaping failed jobs and so on before blocking to dequeue again. That option limits the time it blocks to 1 second. So our workers can stop 4 seconds earlier on average, and this test stops a lot of workers.

kraih:
As a bonus `full-stack.t` gets a little faster too.

Before:
```
All tests successful.
Files=1, Tests=48, 170 wallclock secs ( 0.32 usr  0.00 sys + 52.77 cusr  2.73 csys = 55.82 CPU)
Result: PASS
```

After:
```
All tests successful.
Files=1, Tests=48, 147 wallclock secs ( 0.23 usr  0.04 sys + 52.23 cusr  2.84 csys = 55.34 CPU)
Result: PASS
```

t/lib/OpenQA/Test/Utils.pm
kraih: A `0` timeout might work, but is a little risky since the mainloop could start eating a lot of cpu. 🤔

kraih: But at the cost of almost 5 times the cpu cycles. I think `1` has the best balance between runtime and cpu cost.

kraih: The option `--dequeue-timeout` itself comes from [Minion::Command::worker](https://metacpan.org/pod/Minion::Command::minion::worker#SYNOPSIS), which `::run()` calls indirectly.

Martchus (approved):

okurz:
That's very impressive and a great side-effect. Still, I would appreciate if you could put these details into the git commit message

kraih:
Ok, i've added some more information to the commit message.

PR #4363, opened 2021-11-17
fullstack: Log worker-log.txt on test bail

kalikiana:
See: https://progress.opensuse.org/issues/102578

### DIFF
--- a/t/full-stack.t
+++ b/t/full-stack.t
@@ -160,6 +160,11 @@ sub bail_with_log ($job_id, $message) {
     # uncoverable statement count:2
     my $log = eval { $log_file->slurp };
     diag $@ ? "unable to read $log_file: $@" : "$log_file:\n$log";    # uncoverable statement
+    my $worker_log_file = path("$resultdir/../pool/1/worker-log.txt");    # uncoverable statement
+                                                                          # uncoverable statement count:1
+                                                                          # uncoverable statement count:2
+    my $worker_log = eval { $worker_log_file->slurp };
+    diag $@ ? "unable to read $worker_log_file: $@" : "$worker_log_file:\n$worker_log";    # uncoverable statement
     BAIL_OUT $message;    # uncoverable statement
 }
 


Martchus (approved):

Martchus (commented):
There's actually a problem with this: It uses the worker log from the results directory. However, in the test failures we've observed the test was actually still considered running by the web UI. Hence I doubt we will find the worker log there. Therefore it would make sense to check the pool directory as well (`full-stack.d/openqa/pool/1/worker-log.txt`).

kalikiana:
> There's actually a problem with this: It uses the worker log from the results directory. However, in the test failures we've observed the test was actually still considered running by the web UI. Hence I doubt we will find the worker log there. Therefore it would make sense to check the pool directory as well (`full-stack.d/openqa/pool/1/worker-log.txt`).

Good point. I locally tested it now by removing `wait_for_job_running` so the check fails within the first seconds, and it works in all cases (the file is actually written).

Martchus (commented):
Now it would *only* work if the worker didn't cleanup the job before. I suppose for our case that's sufficient. However, I would have made the pool directly lookup a fallback (to consider both locations).

t/full-stack.t
```diff
@@ -160,6 +160,13 @@ sub bail_with_log ($job_id, $message) {
     # uncoverable statement count:2
     my $log = eval { $log_file->slurp };
     diag $@ ? "unable to read $log_file: $@" : "$log_file:\n$log";    # uncoverable statement
+                                                                      # uncoverable subroutine
```
okurz, line 163: why that?

kalikiana:
> Now it would _only_ work if the worker didn't cleanup the job before. I suppose for our case that's sufficient. However, I would have made the pool directly lookup a fallback (to consider both locations).

So what's the case that should fail? It works here without modifying fullstack.t.

t/full-stack.t
kalikiana: I copied the existing code including all comments.

kalikiana: Let's see what happens if I drop those two "stray" comments. Maybe doesn't mean what I thought it means

t/full-stack.t
```diff
@@ -160,6 +160,11 @@ sub bail_with_log ($job_id, $message) {
     # uncoverable statement count:2
     my $log = eval { $log_file->slurp };
     diag $@ ? "unable to read $log_file: $@" : "$log_file:\n$log";    # uncoverable statement
+    my $worker_log_file = path("$resultdir/../pool/1/worker-log.txt");
```
okurz, line 163: This line should still have an uncoverable statement but not about the sub. I suggest you read about the meaning of the different commands as it's good to know for other cases as well

okurz (changes requested):

t/full-stack.t
kalikiana: Why is it not failing then?

kalikiana: I was assuming `uncoverable subroutine` was for `autoinst_log`; but it seems odd that this line is covered when the rest of the function isn't

kalikiana: I added it back ^^

okurz: > Why is it not failing then?

Hm, interesting. https://app.circleci.com/pipelines/github/os-autoinst/openQA/8420/workflows/87065537-ed48-42ab-bee7-ec7ec818274a/jobs/79591 shows that the full stack test succeeded on the first run. That statement is only for coverage collection. But why is https://app.codecov.io/gh/os-autoinst/openQA/compare/4363/diff?src=pr&el=tree#diff-dC9mdWxsLXN0YWNrLnQ= saying that the line is actually covered? I thought this function would only be called in case of failures?

okurz (approved):

t/full-stack.t
perlpunk: > But why is https://app.codecov.io/gh/os-autoinst/openQA/compare/4363/diff?src=pr&el=tree#diff-dC9mdWxsLXN0YWNrLnQ= saying that the line is actually covered?

Because Devel::Cover reports it as covered. Why? Because we say it's uncoverable, and then marking it as covered in the report which is sent to codecov is the only way to tell codecov that this line is uncoverable.

perlpunk: I now tend to think it would be best to move all those uncoverable helper functions into a separate utility module which is then excluded completely from the report.

PR #4365, opened 2021-11-17
ci: Execute fullstack tests via GHA

kalikiana:
- This is not expected to replace or replicate our CircleCI setup,
  so no setup of coverage or caches is done here.
- Triggering arbitrary test targets is out of scope.

Fixes: https://progress.opensuse.org/issues/102581

### DIFF
--- a/.github/workflows/prove.yml
+++ b/.github/workflows/prove.yml
@@ -20,8 +20,6 @@ jobs:
       - uses: actions/checkout@v1
       - name: Install dependencies
         run: bash -x tools/ci/build_cache.sh
-      - name: Fix permissions of working folder
-        run: sudo chown -R squamata ..
       - name: Build os-autoinst
         run: bash -x tools/ci/build_autoinst.sh
       - name: Make project folder user-writable
--- a/.github/workflows/prove.yml
+++ b/.github/workflows/prove.yml
@@ -1,0 +1,43 @@
+---
+name: Perl tests
+# yamllint disable-line rule:truthy
+on:
+  workflow_dispatch:
+    inputs:
+      stability_test:
+        description: 'Set to 1 to fail if any of the RETRY fails'
+        default: 0
+      target:
+        description: 'Makefile target'
+        default: fullstack
+
+jobs:
+  fullstack:
+    runs-on: ubuntu-latest
+    container:
+      image: registry.opensuse.org/devel/openqa/ci/containers/base:latest
+    steps:
+      - uses: actions/checkout@v1
+      - name: Install dependencies
+        run: bash -x tools/ci/build_cache.sh
+      - name: Fix permissions of working folder
+        run: sudo chown -R squamata ..
+      - name: Build os-autoinst
+        run: bash -x tools/ci/build_autoinst.sh
+      - name: Make project folder user-writable
+        run: sudo chown -R $USER $PWD
+      - name: Running unit tests
+        run: tools/ci/run_unit_tests.sh ${{ github.event.inputs.target }}
+        env:
+          STABILITY_TEST: ${{ github.event.inputs.stability_test }}
+      - name: Merge coverage databases
+        run: cover -write cover_db cover_db*
+      - name: Generate coverage report
+        run: make coverage-report-html
+      - uses: actions/upload-artifact@v2
+        with:
+          name: Upload artifacts
+          path: |
+            test-results
+            cover_db
+        if: always()
--- a/tools/ci/build_autoinst.sh
+++ b/tools/ci/build_autoinst.sh
@@ -8,6 +8,10 @@ set -ex
 destdir=${1:-../os-autoinst}
 sha=${2:-$(cat tools/ci/autoinst.sha)}
 
+[[ -d $destdir ]] || {
+  sudo mkdir "$destdir"
+  sudo chown -R $USER "$destdir"
+}
 echo Building os-autoinst $destdir $sha
 git clone https://github.com/os-autoinst/os-autoinst.git "$destdir"
 ( cd "$destdir"
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -67,8 +67,7 @@ aliases:
         sudo cp -r /var/cache/autoinst ../os-autoinst
         sudo chown -R 1000 ../os-autoinst/t/data
       else # only in local run
-         [ -d ../os-autoinst ] || { sudo mkdir ../os-autoinst && sudo chown -R 1000 ../os-autoinst; }
-         bash tools/ci/build_autoinst.sh ../os-autoinst
+         tools/ci/build_autoinst.sh
        fi
 
   - &make_test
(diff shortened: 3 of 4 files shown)


kalikiana:
Interestingly it seems that `/__w/openQA/openQA/.git: Permission denied` can be worked around by using `actions/checkout@v1` (same issue I saw in #4352).

.github/workflows/prove.yml
```diff
@@ -0,0 +1,19 @@
...
+  fullstack:
+    runs-on: ubuntu-latest
+    container:
+      image: registry.opensuse.org/devel/openqa/ci/containers/base:latest
+    steps:
+      - uses: actions/checkout@v1
+      - name: Install dependencies
+        run: bash -x tools/ci/build_cache.sh
+      - name: Fix permissions of working folder
+        run: sudo chown -R squamata ..
```
okurz, line 15: Is this even necessary for GHA? If it's necessary here as well I suggest to not duplicate but include as part of one of the shell scripts

.github/workflows/prove.yml
```diff
@@ -0,0 +1,19 @@
+---
+name: Perl tests
+# yamllint disable-line rule:truthy
+on: [push, pull_request]
+jobs:
+  fullstack:
+    runs-on: ubuntu-latest
```
okurz, line 7: I think you should add a name in this level. See how it's currently named in the rest results in the main PR page, something about JavaScript, not that helpful

kalikiana: Not sure what you're asking. `fullstack` shows up as the name of this job:

image

.github/workflows/prove.yml
kalikiana: Yes, otherwise the `git clone` in `build_autoinst.sh` fails

kalikiana: Not sure it feels very clean to handle permissions in `build_autoinst.sh`... but I guess it might work

.github/workflows/prove.yml
okurz: I mean here

github_pr_details_name_Screenshot_20211122_101304

where it says "javascript / Node 16.x and ubuntu-latest"

tools/ci/build_autoinst.sh
```diff
@@ -8,6 +8,7 @@ set -ex
 destdir=${1:-../os-autoinst}
 sha=${2:-$(cat tools/ci/autoinst.sha)}
 
+sudo chown -R $USER $destdir/..
```
okurz, line 11: when you understood why we need this would be great to have a comment on top explaining the reason :)

okurz (commented):
I wonder, do you even need this merged or what is the goal? If you just want to show that it works I think you achieved that, haven't you?

.github/workflows/prove.yml
perlpunk: That doesn't have anything to do with this pull request, though.

okurz: oh, sorry. I misread. The new test says "Perl tests / fullstack". Ok, that's fine. Disregard this point

tools/ci/build_autoinst.sh
kalikiana: I'm doing this because I get permission errors. The folder outside the project itself is not owned by the user.

kalikiana:
> I wonder, do you even need this merged or what is the goal? If you just want to show that it works I think you achieved that, haven't you?

Sure. But unless we merge it, it won't be used at all.

> I wonder, do you even need this merged or what is the goal? If you just want to show that it works I think you achieved that, haven't you?

Well, the ticket asks for a proof of concept. So I assume either we consider this useful as-is, or a manually-run version. Otherwise we could just have a follow-up ticket to draw the conclusions for next steps later.

tools/ci/build_autoinst.sh
okurz: hm, ok. Well, leave as is then. But now there is https://github.com/os-autoinst/openQA/runs/4284016039?check_suite_focus=true#step:5:9

okurz:
If we would merge this then we would run full stack tests in parallel to circleCI which is a bit wasteful. The original idea is that we would run the test only on "request". So maybe you can change the line to trigger "on push + pull request" to something like "click of a button"

kalikiana:
Also brought up in the daily. It seems everyone would be okay with 1. manually triggered only 2. with coverage so timing is more comparable (without uploading).
And as discussed I'll move the whole mkdir+chown into build_autoinst.sh so that it's consistent.

The trigger will look like this, but this can only be seen on my fork's default branch for now.

image

perlpunk:
But you could add STABILITY_TEST: as an input field...

okurz (approved):

tools/ci/build_autoinst.sh
```diff
@@ -8,6 +8,10 @@ set -ex
 destdir=${1:-../os-autoinst}
 sha=${2:-$(cat tools/ci/autoinst.sha)}
 
+[ -d $destdir ] || {
```
Martchus, line 11: This lacks quoting but since we use bash just use `[[`:

```suggestion
[[ -d $destdir ]] || {
```

kalikiana:
> But you could add STABILITY_TEST as an input field...

Aye, that's easy.

These obscure permission issues are draining, though:

    mkdir test-results: Permission denied at /usr/lib/perl5/5.26.1/TAP/Harness.pm line 887.

So I added mkdir+chown for that to `prove.yml`.

And also this one now:

    mkdir /__w/openQA/openQA/assets/cache: Permission denied at 

Which prompted me to also add a "make project folder user-writable" step...

[This](https://github.com/kalikiana/openQA/runs/4302035020) is what a full run with errors and retrying looks like on my fork currently - eventually ending here:

    cover -write cover_db cover_db*
    Can't open database /__w/openQA/openQA/cover_db*

perlpunk:
I think cover_db* can't be found because the tests were not run with `COVERAGE=1` (see `tools/ci/run_unit_tests.sh`)

kalikiana:
> I think cover_db* can't be found because the tests were not run with `COVERAGE=1` (see `tools/ci/run_unit_tests.sh`)

Ah, thanks! I guess I can re-use that.. and maybe move the variable setting in there, too, so I can deduplicate a little

Martchus (approved):

Martchus:
But please check whether the error in `cache.fullstack` is related.

okurz (changes requested):
https://app.circleci.com/pipelines/github/os-autoinst/openQA/8453/workflows/50ea8db4-3c82-4d54-8fa6-ef014520f53a/jobs/79882

kalikiana:
> https://app.circleci.com/pipelines/github/os-autoinst/openQA/8453/workflows/50ea8db4-3c82-4d54-8fa6-ef014520f53a/jobs/79882

Unintended changes amended.

okurz:
hm, no coverage report?

perlpunk:
coverage report is here: https://github.com/os-autoinst/openQA/pull/4365/checks?check_run_id=4327231527

okurz (approved):
yes. The report is there and it states that the coverage does not change so maybe that's why there is no comment.

PR #4373, opened 2021-11-25
Fix and improve t/full-stack.t (poo#102578)

okurz:


okurz:
failed in https://app.circleci.com/pipelines/github/os-autoinst/openQA/8463/workflows/cd2a5072-e001-4a19-9e6e-199f78f42340/jobs/79953 with

```
[13:08:01] t/full-stack.t .. 24/? Bailout called.  Further testing stopped:  test 't/full-stack.t' exceeds runtime limit of '1800' seconds
```

in the 3rd rerun. The former 3 runs all passed within 100s. Retriggering once.

t/full-stack.t
```diff
@@ -147,24 +147,27 @@ sub start_worker_and_assign_jobs ($worker_class = undef) {
     assign_jobs $worker_class;
 }
 
-sub autoinst_log ($job_id) { path($resultdir, '00000', sprintf("%08d-$job_name", $job_id))->child('autoinst-log.txt') }
+sub logfile ($job_id, $filename) {
+    my $log = path($resultdir, '00000', sprintf("%08d-$job_name", $job_id))->child($filename);
+    return -e $log ? $log : path("$resultdir/../pool/1/")->child($filename);
```
okurz, line 152: who can explain why these lines are reported to be not covered? We call `logfile` in multiple places that are part of normal test execution, e.g. https://github.com/os-autoinst/openQA/pull/4373/files#diff-b3a7d90ee6870f1e41acae1b9eaeea572deed55bf0b4c32aae8494d86e1fda13R201 . Isn't it?

okurz:
Also https://codecov.io/gh/os-autoinst/openQA/compare/afbb8ee357472fcc37c36c9d80159d67eca19115...71722a9b86411f4f017944ce45020ab04f3d2927/changes says we are not covering L106 and following. Why should that be?

t/full-stack.t
okurz: Discussed with @Martchus @kalikiana and we found that also in https://app.codecov.io/gh/os-autoinst/openQA/blob/master/t/full-stack.t it looks the same and is curious. However also currently https://app.circleci.com/pipelines/github/os-autoinst/openQA/8485/workflows/edaf40b0-a457-4b7c-8308-a433a638ea3b/jobs/80131 fails in master so we decided we should first stabilize t/full-stack.t, then care about coverage more.

kalikiana (approved):

Martchus (approved):

PR #4374, opened 2021-11-26
Fix displaying investigation fields like `diff_to_last_good`

Martchus:
* Remove extra case for displaying errors. This way our tests (which 
  render errors) will automatically cover rendering arbitrary fields.
* Remove check for displaying only certain fields; just use the same code
  path for all the fields as it already handles the case when there is no
  Git repository associated with the diff.
* Otherwise no code has changed. Everything within the previous
  `if ([…].indexOf(key) >= 0) { }`-block is just reindentated.

---

* Fix a few other details, see also further commit messages
* See https://progress.opensuse.org/issues/103032

### DIFF
--- a/assets/javascripts/test_result.js
+++ b/assets/javascripts/test_result.js
@@ -786,6 +786,10 @@ function renderCommentsTab(response) {
 }
 
 function renderInvestigationTab(response) {
+  if (typeof response !== 'object') {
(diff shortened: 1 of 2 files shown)


okurz (approved):

okurz:
Awesome, that as quick. Thank you!

kalikiana (approved):
Nice

PR #4375, opened 2021-11-29
Container setup fixes

dbnicholson:
I was recently experimenting with a container based openQA setup and found a couple issues that had simple fixes.


okurz (approved):
I think this looks great. Thank you!

Martchus (approved):

PR #4376, opened 2021-11-30
AMQP: allow and handle some TLS-related query parameters

AdamWill:
We need to set these query parameters in Fedora infrastructure
in order for message publishing to work. Note they're kind of
odd: they aren't intended for the server to parse, they actually
get consumed on the client before the query is sent, by a rather
indirect route. Mojo::RabbitMQ::Client parses them twice -
first turning 'cacertfile' into 'ca' and 'certfile' into 'cert',
then turning 'certfile' into 'tls_cert' and 'ca' into 'tls_ca'
as options used in a call to `$self->_loop->client`. I didn't
trace it any further than that in detail, but I think these
eventually get used by `Mojo::IOLoop::TLS::negotiate` in
negotiating the TLS connection before the query is sent.

Per its documentation here:
https://metacpan.org/pod/Mojo::RabbitMQ::Client#SUPPORTED-QUERY-PARAMETERS
Mojo::RabbitMQ::Client is following upstream RabbitMQ conventions
in deciding what parameters to honor here, but unlike what's
documented for upstream clients, it does not appear to read the
same options from any configuration file, so it doesn't seem
like we can set these options any other way than jamming them
into the query parameters like this.

### DIFF
--- a/lib/OpenQA/Setup.pm
+++ b/lib/OpenQA/Setup.pm
@@ -113,6 +113,9 @@ sub read_config {
             url => 'amqp://guest:guest@localhost:5672/',
             exchange => 'pubsub',
             topic_prefix => 'suse',
+            cacertfile => '',
+            certfile => '',
+            keyfile => ''
         },
         obs_rsync => {
             home => '',
--- a/lib/OpenQA/WebAPI/Plugin/AMQP.pm
+++ b/lib/OpenQA/WebAPI/Plugin/AMQP.pm
@@ -59,8 +59,15 @@ sub publish_amqp ($self, $topic, $event_data, $headers = {}, $remaining_attempts
     # create publisher and keep reference to avoid early destruction
     log_debug("Sending AMQP event: $topic");
     my $config = $self->{config}->{amqp};
-    my $url = Mojo::URL->new($config->{url})->query({exchange => $config->{exchange}});
-    my $publisher = Mojo::RabbitMQ::Client::Publisher->new(url => $url->to_unsafe_string);
+    my $url = Mojo::URL->new($config->{url});
+    $url->query({exchange => $config->{exchange}});
+    # append optional parameters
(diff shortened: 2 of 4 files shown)


AdamWill:
Until now we've been setting these parameters for Fedora by abusing the sloppy way the URL used to be constructed and just stuffing the extra parameters into the `exchange` config value, but [that got fixed](https://github.com/os-autoinst/openQA/commit/afbd80d15578df3a387384991a7bebdfa0c4a2c3) so our hack didn't work any more.

As the comment mentions, the URL actually winds up being inconsistently escaped. We could make it more consistent by explicitly escaping it at some point during construction, but I'm not sure it's worth the effort. I don't think not escaping the rest of it really causes any problems.

I split out the initial construction of the URL because the old way:

    my $url = Mojo::URL->new($config->{url})->query({exchange => $config->{exchange}});

feels a bit like it's doing too much, and I *think* is actually relying on undocumented behaviour - if I'm reading right, it seems to be relying on the call to `->query()` returning *the `Mojo::URL` instance itself*, but [the docs imply](https://docs.mojolicious.org/Mojo/URL#query) that it should return just the *query*. So it seemed better to avoid any potential problems there. The new way reads better to me anyway.

AdamWill:
oh, and I didn't bother supporting the *other* parameters - `fail_if_no_peer_cert`, `auth_mechanism`, `heartbeat`, `connection_timeout` and `channel_max` - because they just didn't feel like they're likely to be necessary. They could easily be added if needed, though, along the same lines. See https://metacpan.org/pod/Mojo::RabbitMQ::Client#SUPPORTED-QUERY-PARAMETERS .

AdamWill:
The OBS build failure here doesn't look like it's anything to do with the PR.

okurz (approved):
uhm, ok :)

kalikiana (approved):
Looks sensible

PR #4379, opened 2021-12-01
Improve full-stack.t

okurz:
* t: Simplify OpenQA::Test::FullstackUtils
* t: Use signatures in OpenQA::Test::FullstackUtils
* t: Add timeout to developer console waiting status log message

### DIFF
--- a/t/lib/OpenQA/Test/FullstackUtils.pm
+++ b/t/lib/OpenQA/Test/FullstackUtils.pm
@@ -53,22 +53,12 @@ sub get_connect_args () {
     return ["--apikey=1234567890ABCDEF", "--apisecret=1234567890ABCDEF", "--host=http://localhost:$mojoport"];
 }
 
-sub client_output ($args) {
-    my @connect_args = @{get_connect_args()};
-    open(my $client, "-|", "perl ./script/openqa-cli api @connect_args $args");
-    my $out;
-    while (<$client>) {
-        $out .= $_;
-    }
-    close($client);
-    return $out;
-}
+sub client_output ($args) { qx{perl ./script/openqa-cli api @{get_connect_args()} $args} }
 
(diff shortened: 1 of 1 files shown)


t/lib/OpenQA/Test/FullstackUtils.pm
```diff
@@ -44,52 +43,42 @@ my $JOB_SETUP = {
...
+sub wait_for_result_panel (
+    $driver, $result_panel,
+    $context = 'current job',
+    $fail_on_incomplete = undef,
+    $check_interval = 0.5
+  )
+{
     my $looking_for_result = $result_panel =~ qr/Result: /;
-    $context //= 'current job';
     my $msg = "Expected result for $context not found";
```
okurz, line 81: CI tests show a warning about `$context` not defined, see https://app.circleci.com/pipelines/github/os-autoinst/openQA/8499/workflows/0d301351-4632-4b07-bf9f-e2036ff8c24b/jobs/80265?invite=true#step-110-99 , but https://github.com/os-autoinst/openQA/pull/4379/files#diff-523eaf01cc7dd1b0371426b1d310936bc8ae615a6b15cbd107ea0364ccbf8bdaR75 shows that there should be a default value. Did I do it wrong?

Martchus: Likely `undef` (or a variable with `undef` value) is passed explicitly by some invocation. Adding a signature with default value is *not* equivalent to assigning a "default" via `//=`.

okurz: ok, using `undef` as default and setting the string accordingly.

Martchus (approved):
Should be good, only tidy doesn't like it.

kalikiana (approved):

PR #4380, opened 2021-12-01
Improve full-stack.t - take 2

okurz:
* t: Prevent non-deterministic test step calculation using wait_for_ajax
* t: Fix timeout calculation in wait_for_developer_console_like
* t: Prevent trying to wait longer for dev console than overall timeout

### DIFF
--- a/t/lib/OpenQA/SeleniumTest.pm
+++ b/t/lib/OpenQA/SeleniumTest.pm
@@ -173,7 +173,7 @@ sub wait_for_ajax {
         sleep $check_interval;
         $slept = 1;
     }
-    pass("Wait for jQuery successful$msg");
+    note "Wait for jQuery successful$msg";
     return $slept;
 }
 
--- a/t/lib/OpenQA/Test/FullstackUtils.pm
+++ b/t/lib/OpenQA/Test/FullstackUtils.pm
@@ -112,7 +112,7 @@ sub _match_regex_returning_index ($regex, $message, $start_index = 0) {
 
 # waits until the developer console content matches the specified regex
 # note: only considers the console output since the last successful call
-sub wait_for_developer_console_like ($driver, $message_regex, $diag_info) {
+sub wait_for_developer_console_like ($driver, $message_regex, $diag_info, $timeout_s = ONE_MINUTE * 2) {
     my $js_erro_check_suffix = ', waiting for ' . $diag_info;
     ok(javascript_console_has_no_warnings_or_errors($js_erro_check_suffix), 'No unexpected js warnings');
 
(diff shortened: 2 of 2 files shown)


Martchus (approved):

t/lib/OpenQA/SeleniumTest.pm
```diff
@@ -173,7 +173,7 @@ sub wait_for_ajax {
         sleep $check_interval;
         $slept = 1;
     }
-    pass("Wait for jQuery successful$msg");
+    note "Wait for jQuery successful$msg";
```
kalikiana, line 176: I assume this won't show in CircleCI and potentially contribute to a timeout with no output - the comment doesn't mention it, and I think that's potentially important to be aware of

t/lib/OpenQA/Test/FullstackUtils.pm
```diff
@@ -123,7 +123,7 @@ sub wait_for_developer_console_like ($driver, $message_regex, $diag_info) {
     my $previous_log = '';
     # poll less frequently when waiting for paused (might take a minute for the first test module to pass)
     my $check_interval = $diag_info eq 'paused' ? 5 : 1;
-    my $timeout = OpenQA::Test::TimeLimit::scale_timeout(ONE_MINUTE * 5) * $check_interval;
+    my $timeout = OpenQA::Test::TimeLimit::scale_timeout($timeout_s);
```
kalikiana, line 126: I guess based on our collaborative poking we assume that 2 minutes is enough for any successful run.

kalikiana (approved):
No objections, but added comments with my thoughts anyway

t/lib/OpenQA/SeleniumTest.pm
okurz: sorry. I don't know what you mean. However, the logs in circleCI certainly still show a message.

t/lib/OpenQA/Test/FullstackUtils.pm
okurz: yes, should be

t/lib/OpenQA/SeleniumTest.pm
kalikiana: Unlike `diag` a `note` is not included in junit output, see https://perldoc.perl.org/Test::More#note

okurz: sure, that's the junit output which is only about real tests executed

okurz: oh, now I understand what you mean. If there is simply no other test line output for a long enough time then circleCI might abort the job as there was no output received even though there is debugging log that is forwarded into another target. Yes, this might fail tests in those cases. However then something else is broken because no way should there be individual steps that take longer than 10 minutes, the default "no output timeout" in circleCI

PR #4383, opened 2021-12-02
Retry Minion jobs for cleaning job results on SIGTERM/SIGINT

Martchus:
* Prevent `limit_results_and_logs` jobs and follow-up Minion jobs from
  running into "Job terminated unexpectedly" by retrying the jobs instead
* See https://progress.opensuse.org/issues/103416

### DIFF
--- a/lib/OpenQA/Task/Job/Limit.pm
+++ b/lib/OpenQA/Task/Job/Limit.pm
@@ -7,7 +7,8 @@ use Mojo::Base 'Mojolicious::Plugin';
 use OpenQA::Log 'log_debug';
 use OpenQA::ScreenshotDeletion;
 use OpenQA::Utils qw(:DEFAULT resultdir check_df);
-use OpenQA::Task::Utils qw(acquire_limit_lock_or_retry finish_job_if_disk_usage_below_percentage);
+use OpenQA::Task::Utils (qw(acquire_limit_lock_or_retry finish_job_if_disk_usage_below_percentage),
+    qw(enable_retry_on_signals conclude_on_signals));
 use Scalar::Util 'looks_like_number';
 use List::Util 'min';
 use Time::Seconds;
@@ -26,6 +27,7 @@ sub register {
 
 sub _limit {
     my ($job, $args) = @_;
+    enable_retry_on_signals $job;
 
     # prevent multiple limit_results_and_logs tasks and limit_screenshots_task/archive_job_results to run in parallel
     my $app = $job->app;
@@ -65,6 +67,8 @@ sub _limit {
         }
     }
 
+    conclude_on_signals;
+
     # prevent enqueuing new limit_screenshot if there are still inactive/delayed ones
     my $limit_screenshots_jobs
       = $app->minion->jobs({tasks => ['limit_screenshots'], states => ['inactive', 'active']})->total;
@@ -100,6 +104,7 @@ sub _limit {
 
 sub _limit_screenshots {
     my ($job, $args) = @_;
+    enable_retry_on_signals $job;
 
     # prevent multiple limit_screenshots tasks to run in parallel
     my $app = $job->app;
@@ -153,6 +158,7 @@ sub _check_remaining_disk_usage {
 
 sub _ensure_results_below_threshold {
     my ($job, $args) = @_;
+    enable_retry_on_signals $job;
 
     # prevent multiple limit_* tasks to run in parallel
     my $app = $job->app;
--- a/lib/OpenQA/Task/Job/Limit.pm
+++ b/lib/OpenQA/Task/Job/Limit.pm
@@ -7,8 +7,8 @@ use Mojo::Base 'Mojolicious::Plugin';
 use OpenQA::Log 'log_debug';
 use OpenQA::ScreenshotDeletion;
 use OpenQA::Utils qw(:DEFAULT resultdir check_df);
-use OpenQA::Task::Utils (qw(acquire_limit_lock_or_retry finish_job_if_disk_usage_below_percentage),
-    qw(enable_retry_on_signals conclude_on_signals));
+use OpenQA::Task::Utils qw(acquire_limit_lock_or_retry finish_job_if_disk_usage_below_percentage);
+use OpenQA::Task::SignalGuard;
 use Scalar::Util 'looks_like_number';
 use List::Util 'min';
 use Time::Seconds;
@@ -27,7 +27,7 @@ sub register {
 
 sub _limit {
     my ($job, $args) = @_;
-    enable_retry_on_signals $job;
+    my $signal_guard = OpenQA::Task::SignalGuard->new($job);
 
     # prevent multiple limit_results_and_logs tasks and limit_screenshots_task/archive_job_results to run in parallel
     my $app = $job->app;
@@ -67,7 +67,7 @@ sub _limit {
         }
     }
 
-    conclude_on_signals;
+    $signal_guard->retry(0);
 
     # prevent enqueuing new limit_screenshot if there are still inactive/delayed ones
     my $limit_screenshots_jobs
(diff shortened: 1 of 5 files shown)


Martchus:
This turned out to be a bit more complicated than expected because we must not leave dangling locks. However, I could at least add unit tests for all the signal handling code (and the tests aren't too hacky in my opinion).

lib/OpenQA/Task/Utils.pm
```diff
@@ -48,5 +51,53 @@ sub finish_job_if_disk_usage_below_percentage (%args) {
...
+    return $job->note(signal_handler => "Received signal $signal, concluding") if $CONCLUDE_ON_SIGNALS;
+
+    # schedule a retry before stopping the job's execution prematurely
+    $job->note(signal_handler => "Received signal $signal, scheduling retry and releasing locks");
+    $job->retry;
+
+    if (@GUARDS) {
+        # release locks and exit if lock acquisition has been concluded at this point
+        undef $$_ for @GUARDS;
+        exit;
```
kraih, line 69: Are you sure this is necessary? I'd expect `exit` to allow cleanup code in scope guards to run.

kraih: Or is it the timing of destructors resulting in the database connection already being closed when the guard is garbage collected? 🤔

kraih: If every guard really needs to be registered in the future we should probably have a helper that combines creating the guard and registering it.

Martchus: You're right. I must have done something wrong on my first attempt (e.g. sending SIGKILL instead of SIGTERM). That simplifies things greatly. I'll update the PR tomorrow.

Martchus: Ok, I've just pushed it now. That greatly simplifies things :-)

Martchus:
By the way, I've also tested it manually. It works when sending the signal to both processes and when sending it only to the child. Locks are unlocked in any case without special handling (see @kraih 's comment). Not sure what I did wrong on my first test (which lead me to conclude special handling is needed) but the unlocking works actually really reliably.

lib/OpenQA/Task/Utils.pm
```diff
@@ -48,5 +51,30 @@ sub finish_job_if_disk_usage_below_percentage (%args) {
...
+    $job->retry;
+    exit;
+}
+
+# enables retrying the specified Minion job when receiving SIGTERM/SIGINT
+# note: Prevents the job to fail with "Job terminated unexpectedly".
+sub enable_retry_on_signals ($job) {
+    my $handler = sub ($signal) { _handle_signal($job, $signal) };
+    $SIG{TERM} = $handler;
+    $SIG{INT} = $handler;
```
okurz, line 71: I guess this could be

```suggestion
    $SIG{INT} = $SIG{TERM} = sub ($signal) { _handle_signal($job, $signal) };
```

t/16-utils.t
```diff
@@ -4,13 +4,28 @@
...
+our $exit_handler = sub { CORE::exit $_[0] };
+BEGIN {
+    *CORE::GLOBAL::exit = sub (;$) { $exit_handler->(@_ ? 0 + $_[0] : 0) }
+}
+sub exit_code(&) {
+    my $exit_code;
+    local $exit_handler = sub { $exit_code = $_[0] };
+    shift->();
+    return $exit_code;
+}
```
okurz, line 17: that's some fun magic you do here :) Just saying, needs no changes

t/16-utils.t
```diff
@@ -375,10 +392,25 @@ subtest 'change_sec_to_word' => sub {
...
+        $job->called_args_pos_is(0, 3, 'Received signal INT, scheduling retry and releasing locks');
+        $job->clear;
+    };
+    subtest 'signal arrives after `conclude_on_signals` has been called' => sub {
+        conclude_on_signals;
+        is exit_code { kill INT => $$ }, undef, 'not exited (the job is supposed to be concluded at this point)';
+        $job->called_ok('note');
+        $job->called_args_pos_is(0, 3, 'Received signal INT, concluding');
+        ok !$job->called('retry'), 'job not retried';
+    };
```
okurz, line 413: +1

nice use of Test::MockObject :)

okurz (changes requested):

lib/OpenQA/Task/Utils.pm
Martchus: I can try but it might not work because `SIG` is a special variable.

kraih: Should work.

lib/OpenQA/Task/Job/Limit.pm
```diff
@@ -65,6 +67,8 @@ sub _limit {
         }
     }
 
+    conclude_on_signals;
```
kraih, line 70: It would probably be safer if this also used `$job` directly and did a `$job->finish`. Otherwise you'd leave that call to the parent process, and you can't be sure in what state that process is. Less race conditions if the job process updates the status itself.

lib/OpenQA/Task/Job/Limit.pm
```diff
@@ -26,6 +27,7 @@ sub register {
 
 sub _limit {
     my ($job, $args) = @_;
+    enable_retry_on_signals $job;
```
kraih, line 30: I was thinking a bit more about the API, and wouldn't this be nicer as an object? 🤔 
```perl
my $signal_guard = OpenQA::Task::SignalGuard->new($job, {retry => 1});
...
$signal_guard->retry(0);
...
```
On `DESTROY` it could also reset the signal handlers to the previous value. That would probably result in less unintended side effects when jobs run in foreground (without spawning a job process).

lib/OpenQA/Task/Utils.pm
```diff
@@ -7,11 +7,14 @@ use Mojo::Base -signatures;
 use Exporter qw(import);
 use OpenQA::Log qw(log_warning);
 use OpenQA::Utils qw(check_df);
-use Scalar::Util qw(looks_like_number);
+use Scalar::Util qw(looks_like_number weaken);
```
kraih, line 10: You're not actually using `weaken`.

lib/OpenQA/Task/Job/Limit.pm
Martchus: Ok, I'll do that.

kraih (approved):
Code looks pretty clean now.

t/16-utils.t
```diff
@@ -4,14 +4,29 @@
...
 use lib "$FindBin::Bin/lib", "$FindBin::Bin/../external/os-autoinst-common/lib";
 use OpenQA::Utils
   qw(:DEFAULT prjdir sharedir resultdir assetdir imagesdir base_host random_string random_hex download_speed);
+use OpenQA::Task::SignalGuard;
 use OpenQA::Test::Utils 'redirect_output';
 use OpenQA::Test::TimeLimit '10';
 use Scalar::Util 'reftype';
+use Test::MockObject;
 use Mojo::File qw(path tempdir tempfile);
+use Mojo::Util qw(scope_guard);
```
kraih, line 29: I think you're not using this.

okurz (changes requested):
please see the unhandled output in https://app.circleci.com/pipelines/github/os-autoinst/openQA/8524/workflows/995170c8-2cee-450d-9cf9-b0a4e6a17335/jobs/80532

Martchus:
I've been improving the test to avoid the log warnings and the Perl warnings. (None of this is related to my changes in that test module so I've been creating separate commits.)

okurz (approved):

Martchus:
The package build currently fails because it is still relying on Leap 15.2 but RWP is currently unresolvable there. (I removed the *broken* package from our devel repos for now.)

okurz:
please don't merge PRs that can not be fully tested. It does not help to add more untested package changes into the backlog as long as we have a blocking problem

Martchus:
It was just the Leap 15.2 package build. Do we really have to completely stop any kind of change if it is broken?

okurz:
Yes, because we rely on openSUSE Leap 15.2 for now for our production environments and others are allowed to treat it as such until the EOL of Leap 15.2 as well

PR #4384, opened 2021-12-02
Fix running `t/16-utils.t` multiple times

Martchus:
Ensure the `.git` directory is cleaned up before testing the case when it
is missing.


okurz (approved):
I think sometimes I still have the problem that my .git/config in my os-autoinst working copy is changed by one of the tests. I hope that this won't now remove my root directory or something ;)

Martchus:
No, it only affects a nested `.git` directory. I've tested it locally :-)

kalikiana (approved):

PR #4385, opened 2021-12-02
Log the current download speed with every cache service download

kraih:
```
[28029] [i] [#2] Downloading "sle-12-SP3-x86_64-0368-200_123200@64bit.qcow2" from "http://localhost:46603/tests/922756/asset/hdd/sle-12-SP3-x86_64-0368-200_123200@64bit.qcow2"
...
[28029] [i] [#2] Download of "/tmp/Ha0VQsMhaU/t/cache.d/cache/localhost/sle-12-SP3-x86_64-0368-200_123200@64bit.qcow2" successful, new cache size is 123 KiB
```
becomes
```
[28029] [i] [#2] Downloading "sle-12-SP3-x86_64-0368-200_123200@64bit.qcow2" from "http://localhost:46603/tests/922756/asset/hdd/sle-12-SP3-x86_64-0368-200_123200@64bit.qcow2"
...
[28029] [i] [#2] Download of "/tmp/Ha0VQsMhaU/t/cache.d/cache/localhost/sle-12-SP3-x86_64-0368-200_123200@64bit.qcow2" successful (22.13MiB/s), new cache size is 123 KiB
```

Progress: https://progress.opensuse.org/issues/102957


t/16-utils.t
```diff
@@ -61,6 +62,19 @@ subtest 'random number generator' => sub {
     isnt($r, $r2, "random_hex produces different results");
 };
 
+subtest 'download speed' => sub {
+    is download_speed([1638459407, 528237], [1638459408, 628237], 1024), '930.91 Byte/s';
+    is download_speed([1638459407, 528237], [1638459408, 628237], 1024 * 1024), '931 KiB/s';
```
Martchus, line 67: Unless `human_readable_size` lacks test coverage for these cases we likely don't need to test all these cases for the download speed as well.

Martchus (commented):
Looks good, I have just one comment.

t/16-utils.t
kalikiana: I don't agree - but I wonder if this can't be simplified with a hash

kalikiana (approved):
Approving. My comment is only related to style, but no strong opinion there

t/16-utils.t
kraih: In code a hash is usually the right choice, but for tests i really don't like it. Because it makes debugging a bit harder for various reasons. For example you can't as easily rerun one of the cases to replicate a problem, or in the case of an error the line number will point at the test loop, not the actual case that failed. You get the idea. :)

t/25-cache.t
```diff
@@ -329,7 +330,8 @@ subtest 'cache directory is symlink' => sub {
 
     $cache->get_asset($host, {id => 922756}, 'hdd', 'sle-12-SP3-x86_64-0368-200@64bit.qcow2');
     like $cache_log, qr/Downloading "sle-12-SP3-x86_64-0368-200\@64bit.qcow2" from/, 'Asset download attempt';
-    like $cache_log, qr/Download of ".*sle-12-SP3-x86_64-0368-200.*" successful, new cache size is 1024/,
+    like $cache_log,
+      qr/Download of ".*sle-12-SP3-x86_64-0368-200.*" successful \([\d\.]+ \w+\/s\), new cache size is 1024/,
```
okurz, line 334: I would have made it simple and used

```
      qr/Download of ".*sle-12-SP3-x86_64-0368-200.*" successful.*new cache size is 1024/,
```

as I would not care about the *exact* wording for this functional test. As long as we ensure that valid data is there once, that should be enough

kraih: True, testing it once would have been enough.

PR #4388, opened 2021-12-07
Dependency cron 2021-12-07

openqabot:


kalikiana (approved):

okurz (approved):
Tests fail due to perl-Mojo-IOLoop-ReadWriteProcess as discussed in https://progress.opensuse.org/issues/103422 . We need to go forward anyway as we rely on the packages in the current version no matter what

Martchus (changes requested):
I removed `perl-Mojo-IOLoop-ReadWriteProcess` temporarily from our devel repo and it is also not yet in TW. So I'd refrain from updating it right now.

okurz:
> I removed `perl-Mojo-IOLoop-ReadWriteProcess` temporarily from our devel repo and it is also not yet in TW.

Not in Tumbleweed but already in [Factory](https://build.opensuse.org/package/show/openSUSE:Factory/perl-Mojo-IOLoop-ReadWriteProcess). A new Tumbleweed build could not be released mostly due to the problems that the update has caused on openQA already :D

PR #4395, opened 2021-12-10
Add simple gitpod configuration based on the devel:openQA:ci Dockerfile

dcermak:
cc @kalikiana

### DIFF
--- a/lib/OpenQA/Utils.pm
+++ b/lib/OpenQA/Utils.pm
@@ -176,7 +176,11 @@ sub gitrepodir {
     log_warning("$path is not a git directory") if $filename eq '';
     my $config = Config::Tiny->read($filename, 'utf8');
     return '' unless defined $config;
-    return $config->{'remote "origin"'}{url} if ($config->{'remote "origin"'}{url} =~ /^http(s?)/);
+    if ($config->{'remote "origin"'}{url} =~ /^http(s?)/) {
+        my $repo_url = $config->{'remote "origin"'}{url};
+        $repo_url =~ s{\.git$}{/commit/};
+        return $repo_url;
+    }
     my @url_tokenized = split(':', $config->{'remote "origin"'}{url});
     $url_tokenized[1] =~ s{\.git$}{/commit/};
     my @githost = split('@', $url_tokenized[0]);
--- a/t/16-utils.t
+++ b/t/16-utils.t
@@ -348,6 +348,9 @@ subtest 'project directory functions' => sub {
     $mocked_git->child('config')
       ->spurt(qq{[remote "origin"]\n        url = git\@github.com:fakerepo/os-autoinst-distri-opensuse.git});
     is gitrepodir(distri => $distri) =~ /github\.com.+os-autoinst-distri-opensuse\/commit/, 1, 'correct git url';
+    $mocked_git->child('config')
+      ->spurt(qq{[remote "origin"]\n        url = https://github.com/fakerepo/os-autoinst-distri-opensuse.git});
+    is gitrepodir(distri => $distri) =~ /github\.com.+os-autoinst-distri-opensuse\/commit/, 1, 'correct git url';
 
     local $ENV{OPENQA_SHAREDIR} = '/tmp/share';
     is prjdir(), '/tmp/test/openqa', 'right directory';


.gitpod.yml
```diff
@@ -0,0 +1,21 @@
...
+tasks:
+  - name: Lint the Javascript dependencies
+    before: npm i
+    command: npm run lint
+  - name: Build openQA
+    before: |
+      bash -x tools/ci/build_cache.sh
+      bash -x tools/ci/build_autoinst.sh
+      sudo chown -R gitpod $(pwd)
+    command: tools/ci/run_unit_tests.sh fullstack
```
okurz, line 14: openQA does not need to be "built". And what you do is run the fullstack test, but only this one? So what is this for?

container/gitpod/Dockerfile
```diff
@@ -0,0 +1,46 @@
...
+
+# only dependencies for CircleCI to be able to load/save the package cache
+# and fix permissions for the unprivileged user we use
+RUN zypper -n in tar gzip sudo system-group-wheel && zypper clean
+
+# these are autoinst dependencies
+RUN zypper install -y gcc-c++ cmake ninja pkgconfig\(opencv\) pkg-config perl\(Module::CPANfile\) pkgconfig\(fftw3\) pkgconfig\(libpng\) pkgconfig\(sndfile\) pkgconfig\(theoraenc\) npm
+
+# openQA dependencies
+RUN zypper install -y rubygem\(sass\) python3-base python3-requests python3-future git-core rsync curl postgresql-devel postgresql-server qemu qemu-kvm qemu-tools tar xorg-x11-fonts sudo make
```
okurz, line 11: I don't want to maintain this package list manually here as well. You need to find a better solution. At least for gitpod, can you try openQA-devel instead of all packages manually?

container/gitpod/Dockerfile
```diff
@@ -0,0 +1,46 @@
...
+# openQA chromedriver for Selenium tests
+RUN zypper install -y chromedriver
+
+ENV LANG en_US.UTF-8
+
+ENV NORMAL_USER gitpod
+ENV USER gitpod
+
+# RUN echo "$NORMAL_USER ALL=(ALL) NOPASSWD: ALL" >> /etc/sudoers
+# RUN useradd -r -d /home/$NORMAL_USER -m -g users --uid=1000 $NORMAL_USER
```
okurz, line 22: Don't add disabled code as nobody but you will understand why this is not active. Better leave it out and keep in your local git repository, either in `git stash` or a temporary "WIP"-commit

container/gitpod/Dockerfile
```diff
@@ -0,0 +1,46 @@
+FROM opensuse/leap:15.3
```
okurz, line 1: why not inherit from the devel:openQA:ci container instead of copy-pasting content?

okurz (changes requested):

Martchus:
In addition to what @okurz said, the static containers check also failed with a few problems.

container/gitpod/Dockerfile
dcermak: that… was far too obvious…
sorry, fixed now.

.gitpod.yml
dcermak: Mostly to pull in all the dependencies and build os-autoinst, so that the user can start right away testing out their changes.

okurz: that answers the first part and the last question. But what you do is run the fullstack test, but only this one?

container/gitpod/Dockerfile
```diff
@@ -0,0 +1,35 @@
...
+# hadolint ignore=DL3004
+RUN sudo echo "Running 'sudo' for Gitpod: success" && \
+    # create .bashrc.d folder and source it in the bashrc
+    mkdir -p /home/gitpod/.bashrc.d && \
+    (echo; echo "for i in \$(ls -A \$HOME/.bashrc.d/); do source \$HOME/.bashrc.d/\$i; done"; echo) >> /home/gitpod/.bashrc
+
+# explicitly set user/group IDs
+# hadolint ignore=DL3004
+RUN sudo mkdir -p /opt/testing_area && \
+    sudo chown -R "${NORMAL_USER}:users" /opt/testing_area
```
okurz, line 35: IMHO this can be removed as well now as the base container instructions already do the same

container/gitpod/Dockerfile
```diff
@@ -0,0 +1,35 @@
...
+ENV HOME=/home/gitpod
+WORKDIR $HOME
+USER gitpod
+# custom Bash prompt
+RUN { echo && echo ". /etc/bash_completion.d/git-prompt.sh" && echo "PS1='\[\033[01;32m\]\u\[\033[00m\] \[\033[01;34m\]\w\[\033[00m\]\$(__git_ps1 \" (%s)\") $ '" ; } >> .bashrc
+
+### Gitpod user (2) ###
+# use sudo so that user does not get sudo usage info on (the first) login
+# hadolint ignore=DL3004
+RUN sudo echo "Running 'sudo' for Gitpod: success" && \
```
okurz, line 27: If sudo is only necessary to mkdir and chown the dir later on then as the base layer should already do this I wonder if this is still necessary. Or is it necessary because you use here a different user "gitpod" where the base layer uses squamata?

.gitpod.yml
```diff
@@ -0,0 +1,21 @@
...
+tasks:
+  - name: Lint the Javascript dependencies
+    before: npm i
+    command: npm run lint
+  - name: Run the fullstack tests
+    before: |
+      bash -x tools/ci/build_cache.sh
+      bash -x tools/ci/build_autoinst.sh
+      sudo chown -R gitpod $(pwd)
+    command: tools/ci/run_unit_tests.sh fullstack
```
okurz, line 14: this selection of tasks looks rather arbitrary. By default everything should be accessible over the central Makefile. So e.g. `make test` should do something useful after initial dependencies have been installed. I would refrain from calling any scripts from tools/ci as that directory is meant exactly for only CI, the continuous integration systems, e.g. circleCI, or GHA, but not an interactive development environment. However we can make the provisioning of os-autoinst more easily accessible. And maybe it's as easy as renaming tools/ci/build_autoinst.sh to tools/build_autoinst.sh :) But why for example do you need tools/ci/build_cache.sh?

container/gitpod/Dockerfile
dcermak: This is necessary for gitpod to function properly. It needs the user `gitpod` to exist and have passwordless sudo privileges.

container/gitpod/Dockerfile
dcermak: Indeed, that was oversight.

.gitpod.yml
dcermak: Yes, the selection of tasks is very arbitrary: I have simply copy and pasted it from the `ci.yml` worklfow.

Which tasks should be run here?

okurz: I already explained that above: `make test`

container/gitpod/Dockerfile
okurz: ok, fine

.gitpod.yml
kalikiana: > I have simply copy and pasted it from the ci.yml worklfow

@dcermak This is only a subset because our main tests are on CircleCI. And it assumes that you can choose the target via GitHub UI e.g. `checkstyle`, `t`, `ui`, `api`, `fullstack`, `fullstack-unstable`.

If one can select it here as well, that would be really cool - but I'd say all tests is also fine for now.

dcermak:
Huh, interesting, I did not intend to close this PR, but now I also cannot reopen it anymore…

PR #4396, opened 2021-12-10
Retry Minion jobs for cleanup on SIGTERM/SIGINT (e.g. service restarts)

Martchus:
* Use the signal guard introduced by d90f00fd/8b96b710 for all the "limit"
  jobs to avoid them failing with "Job terminated unexpectedly"
* See https://progress.opensuse.org/issues/103416 and
  https://progress.opensuse.org/issues/99831

---

The code is covered by tests. Not sure whether it is necessary to add explicit checks for all these tasks. I did at least a full stack test for all tasks manually.

### DIFF
--- a/lib/OpenQA/Task/Asset/Limit.pm
+++ b/lib/OpenQA/Task/Asset/Limit.pm
@@ -46,7 +46,7 @@ sub _update_exclusively_kept_asset_size {
 
 sub _limit {
     my ($app, $job) = @_;
-    my $signal_guard = OpenQA::Task::SignalGuard->new($job);
+    my $ensure_task_retry_on_termination_signal_guard = OpenQA::Task::SignalGuard->new($job);
 
     # prevent multiple limit_assets tasks to run in parallel
     return $job->finish('Previous limit_assets job is still active')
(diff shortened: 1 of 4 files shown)


lib/OpenQA/Task/Asset/Limit.pm
```diff
@@ -45,6 +46,7 @@ sub _update_exclusively_kept_asset_size {
 
 sub _limit {
     my ($app, $job) = @_;
+    my $signal_guard = OpenQA::Task::SignalGuard->new($job);
```
okurz, line 49: can we make it clearer what the effect of this signal guard is? Maybe just rename the variable to explain what it should ensure. If I understand https://github.com/os-autoinst/openQA/commit/8b96b710 correctly then this could be

```suggestion
    my $ensure_task_retry_on_termination_signal_guard = OpenQA::Task::SignalGuard->new($job);
```

is it?

okurz (changes requested):

okurz (approved):

lib/OpenQA/Task/Asset/Limit.pm
```diff
@@ -45,6 +46,7 @@ sub _update_exclusively_kept_asset_size {
 
 sub _limit {
     my ($app, $job) = @_;
+    my $ensure_task_retry_on_termination_signal_guard = OpenQA::Task::SignalGuard->new($job);
```
kalikiana, line 49: This looks very odd to me, considering the variable will never be used and looks as though it should go out of scope. Except `weaken` and `$SIG` change that.
You're not introducing this API here, so I'm just mentioning it for the record.

kalikiana (approved):

lib/OpenQA/Task/Asset/Limit.pm
Martchus: Having an unused variable is how it works with guards in Perl (and C++ or any language utilizing destructors).

PR #4398, opened 2021-12-10
Improve health-check for `devel:openQA` repository

Martchus:
* Move the code into a separate file and multiple lines
* Log error details in case something goes wrong by adding the curl flag
  `-S` and by printing the OBS response via `trap` to make it clear why
  subsequent actions are skipped
* For https://progress.opensuse.org/issues/103692

### DIFF
--- a/dist/rpm/openQA.spec
+++ b/dist/rpm/openQA.spec
@@ -688,5 +688,6 @@ fi
 %dir %{_unitdir}
 %{_unitdir}/openqa-auto-update.*
 %{_datadir}/openqa/script/openqa-auto-update
+%{_datadir}/openqa/script/openqa-check-devel-repo
 
 %changelog
--- a/script/openqa-check-devel-repo
+++ b/script/openqa-check-devel-repo
@@ -2,6 +2,27 @@
 set -e
 set -o pipefail
 
+usage() {
+    cat << EOF
+Usage: openqa-check-devel-repo
+Checks whether devel:openQA packages are stable
+
+Options:
+ -h, --help         display this help
+EOF
+    exit "$1"
+}
+
+opts=$(getopt -o h --long help -n 'parse-options' -- "$@") || usage 1
+eval set -- "$opts"
+while true; do
+  case "$1" in
+    -h | --help ) usage 0; shift ;;
+    -- ) shift; break ;;
+    * ) break ;;
+  esac
+done
+
 trap 'log_error' ERR
 log_error() {
   echo "devel:openQA looks bad for Leap $VERSION ($ARCH):"
--- a/script/openqa-check-devel-repo
+++ b/script/openqa-check-devel-repo
@@ -1,0 +1,17 @@
+#!/bin/bash
+set -e
+set -o pipefail
+
+trap 'log_error' ERR
+log_error() {
+  echo "devel:openQA looks bad for Leap $VERSION ($ARCH):"
+  echo "${result:-Unable to query OBS}"
+}
+
+[[ $VERSION ]] || . /etc/os-release
+[[ $ARCH ]] || ARCH=$(uname -i)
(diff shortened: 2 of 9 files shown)


dist/rpm/openQA.spec
```diff
@@ -576,6 +577,7 @@ fi
 
 %files common
 %dir %{_datadir}/openqa
+%{_datadir}/openqa/script/openqa-check-devel-repo
```
okurz, line 580: Hm, thinking: Putting this into "common" means that this will be pulled in with the "scripts" package, the webUI package as well as the worker package. So far the "common" package contains what I consider really common for all, but this script has a rather specific purpose. How about just keeping this in auto-update and you would not need the auto-update package to pull in common and then within *our* workers where we want to use this script we also install openQA-auto-update explicitly even though we don't use the systemd file there but only the script explicitly. WDYT?

okurz (changes requested):

dist/rpm/openQA.spec
okurz: By the way. This change is included in the git commit "Improve health-check for devel:openQA repository" which does not sound like a good title for that. "Extract health check …" would be better in that case

Martchus:
Ok, I'll move it to openQA-auto-update and change the commit title. Btw, I intend to use it like this: https://progress.opensuse.org/issues/103692#note-5

okurz (approved):

baierjan (approved):

kalikiana:
[webui-docker-compose](https://github.com/os-autoinst/openQA/runs/4506367796) is failing:
```
Some of the repositories have not been refreshed because of an error.
The command '/bin/sh -c zypper ar -p 95 -f http://download.opensuse.org/repositories/devel:openQA/openSUSE_Leap_15.3 devel_openQA &&     zypper ar -p 90 -f http://download.opensuse.org/repositories/devel:openQA:Leap:15.3/openSUSE_Leap_15.3 devel_openQA_Leap &&     zypper --gpg-auto-import-keys ref &&     zypper in -y --force-resolution ca-certificates-mozilla curl openQA-local-db apache2 hostname which w3m 'rubygem(sass)' &&     zypper clean &&     gensslcert &&     a2enmod headers &&     a2enmod proxy &&     a2enmod proxy_http &&     a2enmod proxy_wstunnel &&     a2enmod ssl &&     a2enmod rewrite &&     a2enflag SSL' returned a non-zero code: 4
Service 'webui_db_init' failed to build : Build failed
Attaching to 
Removing network webui_default
```

script/openqa-check-devel-repo
```diff
@@ -0,0 +1,38 @@
...
+Usage: openqa-check-devel-repo
+Checks whether devel:openQA packages are stable
+
+Options:
+ -h, --help         display this help
+EOF
+    exit "$1"
+}
+
+opts=$(getopt -o h --long help -n 'parse-options' -- "$@") || usage 1
```
perlpunk, line 16: is `parse-options` common here? or should it rather be the name of the script?

script/openqa-check-devel-repo
```diff
@@ -0,0 +1,38 @@
...
+eval set -- "$opts"
```
perlpunk, line 17: I think the `--` are superfluous here as `$opts` is alraedy prefixed with it

okurz:
regarding the failed refresh see https://github.com/openSUSE/zypper/issues/420

script/openqa-check-devel-repo
Martchus: Added another commit to fix it in all scripts.

script/openqa-check-devel-repo
perlpunk: nevermind

perlpunk (approved):

PR #4401, opened 2021-12-13
Add support for new hardware console log

czerw:
Enhancement poo#103903: There is new log hardware-console-log.txt file
to capture console logs on IPMI and PowerVM backends.

Related os-autoinst PR: https://github.com/os-autoinst/os-autoinst/pull/1881

### DIFF
--- a/lib/OpenQA/Schema/Result/Jobs.pm
+++ b/lib/OpenQA/Schema/Result/Jobs.pm
@@ -1190,7 +1190,7 @@ sub delete_logs {
     return undef unless $result_dir;
     my @files = (
         Mojo::Collection->new(
-            map { path($result_dir, $_) } qw(autoinst-log.txt serial0.txt serial_terminal.txt video_time.vtt)
+            map { path($result_dir, $_) } qw(autoinst-log.txt serial0.txt serial_terminal.txt video_time.vtt hardware-console-log.txt)
         ),
         path($result_dir, 'ulogs')->list_tree({hidden => 1}),
         find_video_files($result_dir),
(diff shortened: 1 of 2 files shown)


okurz (changes requested):
Needing to add a file in a manually maintained list in four places is IMHO too much. We need to find a better solution. Anyone has a good suggestion what we could do?

Martchus:
It would make sense to simply extract the list of common log files into a variable (e.g. under utils) and use it everywhere.

Alternatively you could also just avoid the PR altogether and simply put the log under `ulogs`. When I remember correctly, files in that directory are all uploaded implicitly/automatically.

okurz:
> Alternatively you could also just avoid the PR altogether and simply put the log under `ulogs`. When I remember correctly, files in that directory are all uploaded implicitly/automatically.

I would like that as well, yes

czerw:
> It would make sense to simply extract the list of common log files into a variable (e.g. under utils) and use it everywhere.
> 
> Alternatively you could also just avoid the PR altogether and simply put the log under `ulogs`. When I remember correctly, files in that directory are all uploaded implicitly/automatically.

I tried that and files were ignored, placed it into `testresults/ulogs/`.

Martchus:
It is just `ulogs/`, not `testresults/ulogs/.

czerw:
> It is just `ulogs/`, not `testresults/ulogs/`.

Correct, thanks a lot.

czerw:
The downside of `ulogs`: log file is within uploaded logs from SUT, not in `Result files` section, but this is not a big deal.

PR #4407, opened 2021-12-15
Dependency cron 2021-12-15

openqabot:


Martchus (approved):
The new dependencies are due to https://github.com/os-autoinst/os-autoinst/pull/1881 so I guess it is ok.

perlpunk (approved):

PR #4415, opened 2021-12-17
Avoid archiving jobs to fail with "Job terminated unexpectedly"

Martchus:
* Retry archiving jobs when receiving SIGTERM/SIGINT
* Try to conclude cleanup if almost done (should not take long anymore)
* See https://progress.opensuse.org/issues/104116

### DIFF
--- a/lib/OpenQA/Schema/Result/Jobs.pm
+++ b/lib/OpenQA/Schema/Result/Jobs.pm
@@ -260,7 +260,7 @@ sub archivable_result_dir ($self) {
     return $result_dir && -d $result_dir ? $result_dir : undef;
 }
 
-sub archive ($self) {
+sub archive ($self, $signal_guard = undef) {
     return undef unless my $normal_result_dir = $self->archivable_result_dir;
 
     my $archived_result_dir = $self->add_result_dir_prefix($self->remove_result_dir_prefix($normal_result_dir), 1);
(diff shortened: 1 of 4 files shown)


okurz (approved):

lib/OpenQA/Task/Job/ArchiveResults.pm
```diff
@@ -11,6 +11,7 @@ sub register ($self, $app, @args) {
 }
 
 sub _archive_results ($minion_job, @args) {
+    my $ensure_task_retry_on_termination_signal_guard = OpenQA::Task::SignalGuard->new($minion_job);
```
kraih, line 14: That's quite the name. 🥇

lib/OpenQA/Schema/Result/Jobs.pm
```diff
@@ -270,6 +270,7 @@ sub archive ($self) {
         die "Unable to copy '$normal_result_dir' to '$archived_result_dir': $error";
     }
 
+    $signal_guard->retry(0) if $signal_guard;
```
kraih, line 273: What's the purpose of this method call? I only see it used in a test, but `OpenQA::Task::Job::ArchiveResults` doesn't pass along a guard. 🤔

lib/OpenQA/Task/Job/ArchiveResults.pm
Martchus: The name was suggested by @okurz in a previous PR review. I'm just using it consistently here as well.

lib/OpenQA/Schema/Result/Jobs.pm
Martchus: Oh, it should pass it. I'll fix that, good catch.

okurz (approved):

kraih (approved):

PR #4416, opened 2021-12-17
Avoid "File exists" errors when running archiving jobs concurrently

Martchus:
* Create archive destination manually before invoking
  `File::Copy::Recursive::dircopy` to avoid running into race condition
  of the directory creation code of that module
* See https://progress.opensuse.org/issues/104136#note-5


kraih (approved):

kraih:
Wonder if we should maybe just avoid using `File::Copy::Recursive` completely.

okurz (approved):

PR #4417, opened 2021-12-20
Add advanced restart dropdown menu

okurz:
Screenshot of the new button group:

Screenshot_20211220_122717_openqa_restart_button_group

Advanced restart menu shown:

Screenshot_20211220_130544_openqa_advanced_restart_shown

No restart menu shown where not applicable (on jobs without dependencies), preserving old design:

Screenshot_20211220_130639_openqa_no_advanced_restart_on_jobs_without_dependencies_as_in_before

Related progress issue: https://progress.opensuse.org/issues/69979

### DIFF
--- a/t/ui/26-jobs_restart.t
+++ b/t/ui/26-jobs_restart.t
@@ -88,6 +88,7 @@ $driver->find_element_by_link_text('Login')->click();
 subtest 'restart job from info panel in test results' => sub {
     subtest 'parent job shows options for advanced restart' => sub {
         $driver->get_ok('/tests/99900', 'go to job 99900');
+        is @{$driver->find_elements('//a[contains(@id, "restart")]', 'xpath')}, 4, 'multiple restart actions found';
         ok !$driver->find_element('#restart-result-skip-parents')->is_displayed, 'advanced restart entries present';
         $driver->find_element('#restart-button-options')->click;
         my $restart_parent = $driver->find_element('#restart-result-skip-parents');
@@ -108,6 +109,7 @@ subtest 'restart job from info panel in test results' => sub {
     };
     subtest 'assets missing; there is no parent' => sub {
         is($driver->get('/tests/99939'), 1, 'go to job 99939');
+        is @{$driver->find_elements('//a[contains(@id, "restart")]', 'xpath')}, 1, 'only one restart action found';
         $driver->find_element('#restart-result')->click();
         wait_for_ajax(msg => 'fail to start job with missing asset and no parent');
         like($driver->get_current_url, qr|tests/99939|, 'no auto refresh when there are errors/warnings');
--- a/t/ui/26-jobs_restart.t
+++ b/t/ui/26-jobs_restart.t
@@ -83,13 +83,77 @@ sub expected_job_id_regex {
 }
 
 driver_missing unless my $driver = call_driver;
+disable_timeout;
 
 $driver->title_is('openQA', 'on main page');
 $driver->find_element_by_link_text('Login')->click();
-is($driver->get('/tests'), 1, 'get /tests page');
-wait_for_ajax();
+
+subtest 'restart job from info panel in test results' => sub {
+    subtest 'parent job shows options for advanced restart' => sub {
+        $driver->get_ok('/tests/99900', 'go to job 99900');
+        ok !$driver->find_element('#restart-result-skip-parents')->is_displayed, 'advanced restart entries present';
+        $driver->find_element('#restart-button-options')->click;
+        my $restart_parent = $driver->find_element('#restart-result-skip-parents');
+        ok $restart_parent->is_displayed, 'advanced restart entries shown';
+        $restart_parent->attribute_like('href', qr/skip_parents=1/, 'skip parents API URL correct');
+        $driver->find_element('#restart-result-skip-ok-children')
+          ->attribute_like('href', qr/skip_ok_result_children=1/, 'skip OK children API URL correct');
+    };
+    subtest 'child job shows options for advanced restart' => sub {
+        $driver->get_ok('/tests/99901', 'go to job 99901');
+        ok !$driver->find_element('#restart-result-skip-parents')->is_displayed, 'advanced restart present for child';
+        $driver->find_element('#restart-button-options')->click;
+        my $restart_parent = $driver->find_element('#restart-result-skip-parents');
+        ok $restart_parent->is_displayed, 'advanced restart entries shown for child';
+        $restart_parent->attribute_like('href', qr/skip_parents=1/, 'skip parents API URL correct for child');
+        $driver->find_element('#restart-result-skip-ok-children')
+          ->attribute_like('href', qr/skip_ok_result_children=1/, 'skip OK children API URL correct for child');
+    };
+    subtest 'assets missing; there is no parent' => sub {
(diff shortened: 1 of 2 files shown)


lib/OpenQA/Setup.pm
```diff
@@ -373,43 +346,30 @@ sub load_plugins {
     OpenQA::Setup::update_config($server->config, @{$server->plugins->namespaces}, "OpenQA::WebAPI::Auth");
 }
 
-sub set_secure_flag_on_cookies {
-    my ($controller) = @_;
-    if ($controller->req->is_secure) {
-        $controller->app->sessions->secure(1);
-    }
+sub set_secure_flag_on_cookies ($controller) {
```
kraih, line 349: Since this is a big cleanup patch, convention is to call it `$c` instead of `$controller`.

lib/OpenQA/WebAPI/Plugin/REST.pm
```diff
@@ -2,49 +2,33 @@
...
-    my ($self, $app, $config) = @_;
+use Scalar::Util 'blessed';
+use Carp 'croak';
 
+sub register ($self, $app, $config) {
     # special anchor tag with data-method
     $app->helper(
-        action_link => sub {
-            my ($self, $method, $content) = (shift, shift, shift);
+        action_link => sub ($self, $method, $content, @args) {
```
kraih, line 13: This `$self` is incorrect, there is no instance to refer to, it's just `$c` again.

lib/OpenQA/WebAPI/Plugin/REST.pm
```diff
@@ -2,49 +2,33 @@
...
+            return $self->tag('a', href => $url, 'data-method' => $method, @args);
         });
 
     # special anchor tag for post links
-    $app->helper(
-        link_post => sub {
-            my $self = shift;
-            $self->action_link('post', @_);
-        });
+    $app->helper(link_post => sub ($self, @args) { $self->action_link('post', @args) });
```
kraih, line 27: And `$self` instead of `$c` again.

kraih:
The dropdown icon is a bit easy to miss, maybe it would be better as an actual button with outline?

okurz:
> The dropdown icon is a bit easy to miss, maybe it would be better as an actual button with outline?

Well, that would be easy to change. See https://getbootstrap.com/docs/4.1/components/buttons/#examples for all available styles. I am trying to preserve the existing style with small icons and no full-blown buttons. This way we also keep comparable to other symbols, e.g. the cancel button.

lib/OpenQA/Setup.pm
okurz: Thanks. I have included this now and moved to https://github.com/os-autoinst/openQA/pull/4419 as a separate PR (as it can be handled separately and also there is no perfect code coverage yet)

lib/OpenQA/WebAPI/Plugin/REST.pm
okurz: thx, see https://github.com/os-autoinst/openQA/pull/4419

lib/OpenQA/WebAPI/Plugin/REST.pm
okurz: thx, see https://github.com/os-autoinst/openQA/pull/4419

kraih (approved):
I don't like the button, but also not sure what would look better there.

okurz:
> I don't like the button, but also not sure what would look better there.

Thanks for approval then :) As I basically tried to preserve the old design we can still experiment with different designs later on

alice-suse:
> The dropdown icon is a bit easy to miss

@okurz Thanks for the code!  Will notify virtualization team to review.
I also think the dropdown icon is easy to miss. But I do not know how to make it better either. Maybe others have good idea? 
BTW, in the new button group, both the rerun button and dropdown button are there. What happens if people click rerun button without selecting any dropdown item? What happens if selecting the last dropdown item "children"?

okurz:
> BTW, in the new button group, both the rerun button and dropdown button are there. What happens if people click rerun button without selecting any dropdown item? 

That's still the default action so the same things as in before happen. The advanced restart actions in the pull down menu offer the additional behaviour that is available over API, i.e. skip the specified parent or child jobs. Same as over API this allows to save time by not needing to rerun all tests but may be incomplete depending on how the jobs are structured

tbaev:
This is a nice feature for virtualization team, thank you!
I can't think of a better solution then the a drop down button, but I think it works.

kalikiana (approved):
Very nice.

Wrt comments on the button, personally I think least intrusive is best here. Tweaks to the UX could be added as a follow-up even without knowledge of the internals (easy contribution, hint hint)

okurz:
I realized that clicking one of the links in the pull down menu triggers the correct action but the webUI does not update so it is not obvious to the user that the desired action had any effect until manually refreshing the page. Hints how to solve that would be appreciated :)

t/ui/26-jobs_restart.t
```diff
@@ -73,23 +73,86 @@ sub prepare_database {
...
+            expected_job_id_regex, 'warning with link to new job appears');
+    };
+    subtest 'successful restart' => sub {
+        is($driver->get('/tests/99946'), 1, 'go to job 99946');
+        update_last_job_id;
+        $driver->find_element('button > #restart-result')->click();
+        wait_for_ajax(msg => 'successful restart from info panel in test results');
+        like($driver->get_current_url(), expected_job_id_regex, 'auto refresh to restarted job');
+        like($driver->find_element('#info_box .card-header')->get_text(),
+            qr/textmode\@32bit/, 'restarted job is correct');
```
kalikiana, line 149: re https://github.com/os-autoinst/openQA/pull/4417#issuecomment-999355001

I wonder where the gap is - the test looks to cover it all? What's missing?

okurz: I guess what the test is missing that after clicking a restart button we should be automatically moved to the new clone, same as for the default button action

templates/webapi/test/infopanel.html.ep
```diff
@@ -64,20 +64,44 @@
...
-                            </ul>
-                            All rules are applied recursively. Direct use of the REST-API allows excluding parents and (passed/softfailed) children.',
-                            'https://open.qa/docs/#_handling_of_related_jobs_on_failure_cancellation_restart', 'documentation', 'bottom');
-                        %>
+                        % if ($job->has_dependencies) {
+                        <button class="btn btn-link dropdown-toggle" type="button" id="restart-button-options" data-toggle="dropdown">
+                        <span class="caret" title="Additional options for restarting jobs with dependencies"></span></button>
+                          <ul class="dropdown-menu" role="menu" aria-labelledby="restart-button-options" id="restart-menu">
+                            <li role="presentation">
+                              %= link_post url_for('apiv1_restart', jobid => $testid)->query({skip_parents => 1}) => ('data-remote' => 'true', id => 'restart-result-skip-parents', 'data-jobid' => $testid, role => 'menuitem') => begin
```
kraih, line 77: The links might be missing `class="dropdown-item"` for the right styling. 🤔

okurz: Fixed in https://github.com/os-autoinst/openQA/pull/4425

okurz:
follow-up: https://github.com/os-autoinst/openQA/pull/4425

PR #4419, opened 2021-12-20
Use more signatures

okurz:
* Use signatures in OpenQA::WebAPI::Plugin::REST
* Use signatures in OpenQA::Schema
* Use signatures in OpenQA::Setup
* Simplify OpenQA::Setup

### DIFF
--- a/lib/OpenQA/Setup.pm
+++ b/lib/OpenQA/Setup.pm
@@ -2,7 +2,7 @@
 # SPDX-License-Identifier: GPL-2.0-or-later
 
 package OpenQA::Setup;
-use Mojo::Base -strict;
+use Mojo::Base -strict, -signatures;
 
 use Mojo::File 'path';
 use Mojo::Util 'trim';
@@ -20,10 +20,9 @@ use OpenQA::JobGroupDefaults;
 use OpenQA::Task::Job::Limit;
 
 my %CARRY_OVER_DEFAULTS = (lookup_depth => 10, state_changes_limit => 3);
-sub carry_over_defaults { \%CARRY_OVER_DEFAULTS }
+sub carry_over_defaults () { \%CARRY_OVER_DEFAULTS }
 
-sub read_config {
-    my $app = shift;
+sub read_config ($app) {
     my %defaults = (
         global => {
             appname => 'openQA',
@@ -241,8 +240,7 @@ sub read_config {
     _validate_worker_timeout($app);
 }
 
-sub _validate_worker_timeout {
-    my ($app) = @_;
+sub _validate_worker_timeout ($app) {
     my $global_config = $app->config->{global};
     my $configured_worker_timeout = $global_config->{worker_timeout};
     if (!looks_like_number($configured_worker_timeout) || $configured_worker_timeout < MAX_TIMER) {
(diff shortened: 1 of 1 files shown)


kraih (approved):

lib/OpenQA/Schema.pm
```diff
@@ -100,19 +87,12 @@ sub deploy {
 }
 
 # Class attribute used for testing with OpenQA::Test::Database
-sub search_path_for_tests {
-    my $class = shift;
-    state $search_path;
-    $search_path = shift if @_;
-    return $search_path;
-}
+sub search_path_for_tests ($class, $new_search_path = undef) { state $search_path = $new_search_path }
```
perlpunk, line 90: This doesn't work the way you think it is.
Try:
```
$x = search_path_for_tests(x => "abc");
say $x;
$x = search_path_for_tests(x => "abcd");
say $x;
$x = search_path_for_tests("x");
say $x;'
```

```suggestion
sub search_path_for_tests ($class, $new_search_path = undef) {
  state $search_path;
  $search_path = $new_search_path if defined $new_search_path; 
  return $search_path;
}
```

perlpunk (changes requested):

lib/OpenQA/Schema.pm
kraih: Good catch. 🎣

okurz: done

lib/OpenQA/Schema.pm
```diff
@@ -100,19 +87,16 @@ sub deploy {
 }
 
 # Class attribute used for testing with OpenQA::Test::Database
-sub search_path_for_tests {
-    my $class = shift;
+sub search_path_for_tests ($class, $new_search_path = undef) {
     state $search_path;
-    $search_path = shift if @_;
+    $search_path = $new_search_path if defined $new_search_path;
```
perlpunk, line 92: Just an addition to my previous comment:
With this we loose the functionality to set the path to undef.
Previously this was possifle by checking the length of `@_`.
It would still be possible right now and for the next perl versions, but at some point `@_` will be removed for signature subs (there will be an alternative for checking the length of the argument list though).
Context: https://www.nntp.perl.org/group/perl.perl5.porters/2021/12/msg262109.html

So if we need to set it to undef we should check `@_ > 1` here for now.

okurz: so … change it how exactly? Sorry, can you provide a code suggestion for stupid okurz's? :)

kraih: There might be better solutions 🤔 , but one possibility would be:
```perl
sub search_path_for_tests ($class, @new_search_path) {
    state $search_path;
    $search_path = $new_search_path[0] if @new_search_path;
    return $search_path;
}
```

okurz: ok, did that now

kalikiana (approved):

lib/OpenQA/Schema.pm
perlpunk: That's not the code you used.

kalikiana:
```
sub search_path_for_tests ($class, $new_search_path = undef) {
    state $search_path;
    $search_path = $new_search_path[0] if @new_search_path;
    return $search_path;
}
```

okurz: For now I left out the changes in OpenQA::Schema from this PR. Will reserve this for a later change. I noted this conversation to be resolved in the according change first.

okurz:
Moved out the less covered code changes into #4600

okurz:
@perlpunk please

perlpunk (approved):

PR #4420, opened 2021-12-20
Use more signatures - part 2

okurz:
* Use signatures in OpenQA::Shared::Plugin::Gru
* Move 'dsn' function to only place used'
* Use signatures in OpenQA::WebAPI::Plugin::REST
* Use signatures in OpenQA::Schema
* Use signatures in OpenQA::Setup
* Simplify OpenQA::Setup


kalikiana:
@okurz Are you this still waiting for review? There's several CI failures and it's not been updated a while

PR #4422, opened 2021-12-20
Add retry of openQA jobs based on test variables

okurz:
Motivation: Similar as in
https://docs.microsoft.com/en-us/azure/devops/release-notes/2021/sprint-195-update#automatic-retries-for-a-task
and https://docs.gitlab.com/ee/ci/yaml/#retry and
https://github.com/marketplace/actions/retry-step

Acceptance criteria:
* **AC1:** openQA jobs are automatically retriggered with a maximum amount
  specified in test variables
* **AC2:** By default openQA tests apply no automatic retry

Acceptance tests:
* **AT1-1:** *Given* an unfinished openQA test with the special `RETRY`
  variable given with value `N` *When* the job fails *And* the current job is
  not already the N-th clone of an original job *Then* an automatic clone is
  created
* **AT1-2:** *Given* an unfinished openQA test with the special `RETRY`
  variable given with value `N` *When* the job fails *And* the current job is
  the N-th clone of an original job *Then* no automatic clone is created
* **AT2-1:** *Given* an unfinished openQA test with no special retry variable
  given *When* the job fails *Then* no automatic clone is created at this
  point (job done hooks or external events may still trigger clones as well
  as manual triggers)

Review hint: I suggest to review the commits individually. As desired I
can also split the pull request into multiple for easier review.

Related progress issue: https://progress.opensuse.org/issues/104007

### DIFF
--- a/docs/WritingTests.asciidoc
+++ b/docs/WritingTests.asciidoc
@@ -299,6 +299,32 @@ WORKER_CLASS = desktop
 # WORKER_CLASS is not set
 --------------------------------------------------------------------------------
 
+=== Automatic retries of jobs
+
+You might encounter flaky openQA tests that fail sporadically. While the best
+way to address flaky test code is of course to fix the test code itself. For
+example if certain steps rely on external components over network retries
+within the test modules should be applied. However there can still be cases
+where you might want openQA to automatically retrigger jobs. This can be
+achieved by setting the test variable `RETRY` in the format
+`<retries>[:<description>]` to an integer value with the maximum number of
+retries with an optional, additional description string separated by a colon.
+For example triggering an openQA job with the variable `RETRY=2:bug#42` will
+retrigger an openQA test on failure up to 2 totalling to up to 3 jobs. Note
+that the retry jobs are scheduled immediately and will be executed as soon as
+possible depending on available worker slots. Many factors can change in
+retries impacting the reproducibility, e.g. the used worker host and instance,
+any network related content, etc. By default openQA tests do not retry. The
+optional, additional description string is used only for reference and has no
+functional impact.
+
+See <<Installing.asciidoc#automatic_cloning_incomplete_jobs,Automatic cloning of incomplete jobs>>
+for an additional solution intended for administrators handling known issues
+causing incomplete jobs.
+
+<<Installing.asciidoc#custom_hook_scripts_job_done,Custom hook scripts on "job done" based on result>>
+can be used to apply more elaborate issue detection and retriggering of tests.
+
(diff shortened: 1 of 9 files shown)


coolo:
Only looked at the documentation of it, but I think it's a reasonable extension to the ecosystem

perlpunk (approved):

kraih (approved):

docs/WritingTests.asciidoc
```diff
@@ -299,6 +299,32 @@ WORKER_CLASS = desktop
 # WORKER_CLASS is not set
 --------------------------------------------------------------------------------
 
+=== Automatic retries of jobs
+
+You might encounter flaky openQA tests that fail sporadically. While the best
```
Martchus, line 304: The ""While" implies there's a 2nd part within the same sentence but there isn't.

lib/OpenQA/Schema/Result/Jobs.pm
```diff
@@ -2004,6 +2004,20 @@ sub packages_diff {
...
+    return $o ? 1 + $o->ancestors : 0;
+}
+
+sub handle_retry ($self) {
+    return undef unless my $retry = $self->settings_hash->{RETRY};
+    # strip any optional descriptions after a colon
+    $retry =~ s/:.*//;
+    my $ancestors = $self->ancestors;
+    log_debug('handle_retry: Found retry job ' . $self->id . ", try '$ancestors' of up to '$retry' retries");
+    return $ancestors < $retry;
```
Martchus, line 2018: This lacks input validation, e.g. `$retry` might not be a number here.

lib/OpenQA/Schema/Result/Jobs.pm
```diff
@@ -2004,6 +2004,20 @@ sub packages_diff {
...
+    my $o = $self->origin;
+    return $o ? 1 + $o->ancestors : 0;
+}
+
+sub handle_retry ($self) {
+    return undef unless my $retry = $self->settings_hash->{RETRY};
+    # strip any optional descriptions after a colon
+    $retry =~ s/:.*//;
+    my $ancestors = $self->ancestors;
+    log_debug('handle_retry: Found retry job ' . $self->id . ", try '$ancestors' of up to '$retry' retries");
```
Martchus, line 2017: It would be more useful to add this as a comment (visible in the web UI).

Martchus:
This PR should fix all mentioned issues: https://github.com/os-autoinst/openQA/pull/4422#pullrequestreview-922883535=

PR #4425, opened 2021-12-22
templates: Fix redirection to clones on advanced restart actions

okurz:
Related progress issue: https://progress.opensuse.org/issues/69979

### DIFF
--- a/t/ui/26-jobs_restart.t
+++ b/t/ui/26-jobs_restart.t
@@ -142,7 +142,7 @@ subtest 'restart job from info panel in test results' => sub {
     subtest 'successful restart' => sub {
         is($driver->get('/tests/99946'), 1, 'go to job 99946');
         update_last_job_id;
-        $driver->find_element('button > #restart-result')->click();
+        $driver->find_element('#restart-result')->click();
         wait_for_ajax(msg => 'successful restart from info panel in test results');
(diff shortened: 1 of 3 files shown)


kraih:
Some ui tests might be good, considering how JavaScript heavy the code path is.

t/ui/26-jobs_restart.t
```diff
@@ -142,7 +142,7 @@ subtest 'restart job from info panel in test results' => sub {
     subtest 'successful restart' => sub {
         is($driver->get('/tests/99946'), 1, 'go to job 99946');
         update_last_job_id;
-        $driver->find_element('button > #restart-result')->click();
+        $driver->find_element('#restart-result')->click();
```
kalikiana, line 145: You're changing it here. But the test wasn't failing before either. Please convince me why this should work now w/o other changes, or extend the test accordingly

kalikiana (changes requested):

t/ui/26-jobs_restart.t
okurz: Added according tests for the redirection after clicking the advanced menu entries same like on the default action where it's already tested

okurz:
> Some ui tests might be good, considering how JavaScript heavy the code path is.

Extended tests to cover that

kalikiana (approved):

kraih (approved):

PR #4427, opened 2021-12-23
Remove log messages because missing details-*-.json files are expected

kraih:
Same as #4412, but with tests for the regression.

Progress: https://progress.opensuse.org/issues/102332

### DIFF
--- a/lib/OpenQA/Schema/Result/JobModules.pm
+++ b/lib/OpenQA/Schema/Result/JobModules.pm
@@ -9,7 +9,6 @@ use warnings;
 use 5.012;    # so readdir assigns to $_ in a lone while test
 use base 'DBIx::Class::Core';
 
-use OpenQA::Log qw(log_debug);
 use OpenQA::Jobs::Constants;
 use Mojo::JSON qw(decode_json encode_json);
 use Mojo::File qw(path tempfile);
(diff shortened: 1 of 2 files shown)


lib/OpenQA/Schema/Result/JobModules.pm
```diff
@@ -119,14 +113,9 @@ sub results {
...
-            eval { $step->{text_data} = decode('UTF-8', path($dir, $text_file_name)->slurp); };
-            if (my $error = $@) {
-                # try reading the results one more time if the JSON file's size has increased; otherwise render an error
-                # note: Likely a concurrent finalize_job_results Minion job has finished so the separate text file has
-                #       just been incorporated within the JSON file.
-                return $self->results(%options) if (-s $file_name // -1) > ($initial_file_size // -1);
-                $step->{text_data} = "Unable to read $text_file_name.";
-            }
+            my $text_file = path($dir, $text_file_name);
+            my $text_data = -r $text_file ? decode('UTF-8', $text_file->slurp) : undef;
```
kraih, line 117: The readable check that was missing in the previous version of the patch. It's needed because the text file might not have been uploaded yet.

kalikiana (approved):

t/10-jobs-results.t
```diff
@@ -182,4 +182,32 @@ subtest 'saving results' => sub {
     is_deeply(decode_json($details_file->slurp), \%some_test_results, 'overall structure of test results preserved');
 };
 
+subtest 'loading results with missing file in details' => sub {
+    my $some_test_results = {
+        results => [
+            {
+                test_fqn    => 'random:test',
```
okurz, line 189: Needs tidying

okurz (changes requested):

t/10-jobs-results.t
kraih: Looks like i used the wrong version of perltidy.

okurz (approved):

PR #4429, opened 2022-01-05
WIP: use tpage to create the apache templates

phil-hands:
Some of the details of the apache .template files in Debian need to be
different, primarily where they are installed and comments about how SSL
is configured.  That being the case we have a patch to do that.

As an alternative to carrying that patch, I thought I might be able to
come up with a template that can combine the creation of both the
non-SSL and SSL templates (hopefully making it worth accepting upstream)
while also making the template and the Makefile a little more flexible
such that I can define a couple of flags in the debian/rules file, and
have it behave as we need for the Debian package.

This introduces a build dependency on Template::Tools::tpage, which I've
not attempted to add to the dependency information, as I'm not sure how
that's done.

If you don't think this is worth having, that's OK -- I'll just continue
with the patch as I already have it.

FYI the things I define for Debian are:

```
  APACHE_VHOSTS_DIR=sites-available
  APACHE_TEMPLATE_OPTS="--define distri=Debian"
```

The first ensures that the files are installed into Debian's preferred
/etc/apache2/sites-available (rather than .../vhosts.d seen on openSUSE)
and the second mostly gets rid of the stuff about needing SSL in
APACHE_SERVER_FLAGS (which is not how we do it in Debian)

I note that we also need to enable the `expires` module in apache when
on Debian. I don't know if that's a real difference, or whether perhaps
you also need that in openSUSE and simply forgot to include mention of
that here.


Martchus (commented):
`tpage` needs to be added to the dependencies (see package build failures). Likely the spec file needs to be adapted as well. If that's the case we should also test whether the apache setup is still working under openSUSE (so I'm adding the not-ready label).

phil-hands:
> `tpage` needs to be added to the dependencies (see package build failures).

That is correct (as mentioned in the 3rd paragraph: "This introduces a build dependency on Template::Tools::tpage ...")

> Likely the spec file needs to be adapted as well. If that's the case we should also test whether the apache setup is still working under openSUSE (so I'm adding the not-ready label).

As mentioned, I don't really know how to do this, or have the ability to properly test this on openSUSE, so was hoping someone that is able would be kind enough to give it a try.

Martchus:
Dependencies can be added to a YAML file which is then applied to the spec file (see https://open.qa/docs/#_perl_and_other_packages). The package build is checked via the "OBS Package Build" check shown on GitHub. If you've adapted the spec file so the package build passes I can test the package locally.

PR #4430, opened 2022-01-05
WIP: fix for progress bars broken by bootstrap4 update

phil-hands:
In Debian we try to use packaged versions of libraries etc. and also packages are not allowed to download things at build time (and it's generally frowned upon to do it at run-time) so I've got a patch that munges the assetpack.def to use local versions.

I'm therefore using the version of bootstrap4 that is found in debian's [libjs-bootstrap4](https://packages.debian.org/stable/libjs-bootstrap4) package, which is version 4.5.2 rather than the 4.1.1 that's specified in the `assetpack.def`.

The new version works, but the rendering of the text in the scroll bars is wrong, as one can see here:
scrollbar-unpatched

This patch fixes that, thus:
image

although I'm not completely certain if that's the _right_ way to fix that, or whether it also works in 4.1.1 -- hence the WIP

### DIFF
--- a/assets/stylesheets/dashboard.scss
+++ b/assets/stylesheets/dashboard.scss
@@ -24,6 +24,7 @@ td div.progress {
         // don' use flex box because it leads to undesired overflow behavior
         display: inline;
         overflow: hidden;
+        padding-top: 10px;
         a {
             display: block;
             width: 100%;


Martchus:
This may not be the only issue, see https://github.com/os-autoinst/openQA/pull/4323. Preferably use the exact version of the library we're using. These JavaScript/web libraries often have compatibility problems. By the way, for the openSUSE packaging we have similar requirements and therefore create a static archive with all the "web dependencies" (see https://open.qa/docs/#update-asset-cache) to avoid downloading them during the package build.

I'm also wondering whether we can merge your change as it likely breaks with "our" version of Bootstrap.

phil-hands:
> This may not be the only issue, see #4323. Preferably use the exact version of the library we're using.

I'm afraid that until additional issues become apparent (which they've not done so far) I'm not going to be able to justify packaging or embedding an old version of an already packaged library.

>  These JavaScript/web libraries often have compatibility problems.

Yes, I know.  I thought it might be helpful to know that it seems possible that in this case there don't seem to be any immediately obvious issues with changing to this version, in case there was a wish to update at some point.

> By the way, for the openSUSE packaging we have similar requirements and therefore create a static archive with all the "web dependencies" (see https://open.qa/docs/#update-asset-cache) to avoid downloading them during the package build.

I'd not actually noticed `update-cache.sh`, but do also have a patch for `tools/generate-packed-assets` to make it throw an error if it finds a missing asset so that package builds will fail if I've not managed to satisfy all the dependencies before attempting to build the package.  Using a script to populate such dependencies doesn't really fit with the Debian way of doing things, as if anything new turns up I need to document it in the `debian/copyright` file, and should also have a stab at using an already-packaged library if possible.

> I'm also wondering whether we can merge your change as it likely breaks with "our" version of Bootstrap.

That's why I marked this WIP.

I'm not really able to test the openSUSE version, and wouldn't be convinced by my results even if I managed to do so, so I was hoping that someone would give it a go and see what happens (it's only a one-line change, after all :-) ).

mimi1vx:
> This may not be the only issue, see #4323. Preferably use the exact version of the library we're using. These JavaScript/web libraries often have compatibility problems. By the way, for the openSUSE packaging we have similar requirements and therefore create a static archive with all the "web dependencies" (see https://open.qa/docs/#update-asset-cache) to avoid downloading them during the package build.

even better OBS doesn't allow any outside connection during build.

Martchus:
Note that this fix is not required when using Bootstrap 4.6.1, e.g. https://github.com/os-autoinst/openQA/pull/4323 works just fine without it.

Considering this change is only required with Bootstrap versions that are older than the current 4.x version I'm inclined to close this PR as we are unlikely to need the change to move forward.

Martchus:
Looks like we did a mistake in #4323 and we need this after all. I've cherry picked your commit on https://github.com/os-autoinst/openQA/pull/4579 so it'll be merged as part of it.

PR #4431, opened 2022-01-05
Pass the correct asset variable to _handle_asset_processed

perlpunk:
Issue: https://progress.opensuse.org/issues/98577

### DIFF
--- a/lib/OpenQA/Worker/Engines/isotovideo.pm
+++ b/lib/OpenQA/Worker/Engines/isotovideo.pm
@@ -123,8 +123,7 @@ sub cache_assets ($cache_client, $job, $vars, $assets_to_cache, $assetkeys, $web
         CACHE_SERVICE_POLL_DELAY,
         sub ($error, $status) {
             $error
-              = _handle_asset_processed($cache_client, $assets_to_cache, $asset_uri, $status, $vars, $webui_host,
-                $pooldir)
+              = _handle_asset_processed($cache_client, $this_asset, $asset_uri, $status, $vars, $webui_host, $pooldir)
               unless $error;
             return $callback->($error) if $error;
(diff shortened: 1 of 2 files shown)


perlpunk:
The other effect was that this line:
```
$vars->{$this_asset}
      = $vars->{ABSOLUTE_TEST_CONFIG_PATHS} ? $link_result->{absolute_path} : $link_result->{basename};
```
wasn't doing anymore what it was supposed to, writing to `$vars->{"ARRAY(...)"}` instead of the asset name.
But apparently this also didn't raise any problems, so I'm wondering when this is actually needed.

kalikiana (approved):

perlpunk:
@foursixnine maybe you can say something about that, as you authored c1aa4937006e1f80eda38da072bf9a98635a313a ?

perlpunk:
Maybe also @Martchus can help with the open question when he's back next week.

okurz (approved):

Martchus:
We kept `ABSOLUTE_TEST_CONFIG_PATHS` as a feature switch so users can go back to the old behavior of using absolute paths if needed. However, maybe it is time to remove the support for absolute paths at this point. We could check at least our instances whether users still use `ABSOLUTE_TEST_CONFIG_PATHS`.

Martchus (approved):

foursixnine (commented):
I'd add an extra test case to check that defined pflash with undefined uefi, also works, for today and for the future's sake, as there are  no tests verifying that path at all.

Other than that it looks good, hell of a trip down the memory lane

perlpunk:
> I'd add an extra test case to check that defined pflash with undefined uefi, also works

I think I don't understand how such a test would look like.
Can you elaborate?

perlpunk:
I removed the not-ready label to get this merged and will add @foursixnine 's comment to the ticket

PR #4437, opened 2022-01-07
Prevent bugrefs in comments from getting expanded if they are part of a larger string

kraih:
Trying to simplify the comment rules openQA users have to follow for bugrefs.

Progress: https://progress.opensuse.org/issues/104199

### DIFF
--- a/t/16-markdown.t
+++ b/t/16-markdown.t
@@ -50,6 +50,15 @@ subtest 'bugrefs' => sub {
     is markdown_to_html("boo\ntesting boo#123 123\n123"),
       qq{<p>boo\ntesting <a href="https://bugzilla.opensuse.org/show_bug.cgi?id=123">boo#123</a> 123\n123</p>\n},
       'bugref expanded';
+    is markdown_to_html('related issues: boo#123,bsc#1234'),
+      qq{<p>related issues: <a href="https://bugzilla.opensuse.org/show_bug.cgi?id=123">boo#123</a>,}
+      . qq{<a href="https://bugzilla.suse.com/show_bug.cgi?id=1234">bsc#1234</a></p>\n},
+      'bugref expanded';
+    is markdown_to_html('related issue: bsc#1234, yada yada'),
+      qq{<p>related issue: <a href="https://bugzilla.suse.com/show_bug.cgi?id=1234">bsc#1234</a>, yada yada</p>\n},
+      'bugref expanded';
+    is markdown_to_html('label:force_result:passed:bsc#1234'), "<p>label:force_result:passed:bsc#1234</p>\n",
+      'bugref not expanded because part of larger string';
 };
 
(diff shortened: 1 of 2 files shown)


kraih:
Odd, the failed ui test `t/ui/15-admin-workers.t` passes locally.

kraih:
Including the comma in addition to whitespace is arguable. I've included it for backwards compatibility for now since it had a test case.

t/16-markdown.t
```diff
@@ -50,6 +50,15 @@ subtest 'bugrefs' => sub {
...
       'bugref expanded';
+    is markdown_to_html('related issues: boo#123,bsc#1234'),
+      qq{<p>related issues: <a href="https://bugzilla.opensuse.org/show_bug.cgi?id=123">boo#123</a>,}
+      . qq{<a href="https://bugzilla.suse.com/show_bug.cgi?id=1234">bsc#1234</a></p>\n},
+      'bugref expanded';
+    is markdown_to_html('related issue: bsc#1234, yada yada'),
+      qq{<p>related issue: <a href="https://bugzilla.suse.com/show_bug.cgi?id=1234">bsc#1234</a>, yada yada</p>\n},
+      'bugref expanded';
+    is markdown_to_html('label:force_result:passed:bsc#1234'), "<p>label:force_result:passed:bsc#1234</p>\n",
+      'bugref not expanded because part of larger string';
```
okurz, line 61: This will break AC3 of https://progress.opensuse.org/issues/104199 . Do you have an idea how we could still have an `a href` created?

kraih: Your AC3 says `AC3: Any text starting with "label:" is not detected as an openQA bugref`. That's exactly what is happening here. This resolves AC3.

kraih: If you want bugrefs that are part of a label as clickable links please make that a new ticket. And leave it for after the label highlighting has been added. Since it will make that a bit harder too.

kraih: Personally, i'd rather not generate links for substrings of a label. The bugref regex is pretty complex already, but has good test coverage. Label links would require two versions of the regex, which will naturally end up with worse coverage.

Martchus: But doesn't it break AC2? However, I agree that simplicity might be preferable here.

kraih: That's true, it breaks AC2.

Martchus: I've read your messages on Slack so I assume we decided to postpone AC2.

Martchus (approved):

t/16-markdown.t
kalikiana: I saw the note on Slack. I think it's sensible to move out AC2 and make this a separate topic. Even if we want that, it can be looked into on its own

kalikiana (approved):

PR #4438, opened 2022-01-10
Improve documentation for advanced restarting UI

Martchus:
The related ticket https://progress.opensuse.org/issues/69979 is on feedback so here's some in form of a PR. See particular commit messages for details.

### DIFF
--- a/templates/webapi/test/infopanel.html.ep
+++ b/templates/webapi/test/infopanel.html.ep
@@ -92,7 +92,7 @@
                               <li>Any kind of chained child jobs are restarted.</li>
                               <li>Directly chained parent jobs are restarted.</li>
                           </ul>
-                          All rules are applied recursively. Direct use of the REST-API allows excluding parents and (passed/softfailed) children.',
+                          All rules are applied recursively. Advanced restarting options allow changing the behavior, e.g. excluding parents and (passed/softfailed) children.',
                           'https://open.qa/docs/#_handling_of_related_jobs_on_failure_cancellation_restart', 'documentation', 'bottom');
                       %>
                     % }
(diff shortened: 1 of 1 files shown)


templates/webapi/test/infopanel.html.ep
```diff
@@ -79,20 +80,20 @@
                             <span>Skip children</span>
                             %= end
                             %= link_post url_for('apiv1_restart', jobid => $testid)->query({skip_ok_result_children => 1}) => (class => 'dropdown-item restart-result', 'data-remote' => 'true', id => 'restart-result-skip-ok-children', 'data-jobid' => $testid) => begin
-                            <span>Skip restarting "OK" children</span>
+                            <span>Skip restarting "OK" (<%= $ok_results_explanation %>) children</span>
```
Martchus, line 83: Or just
```suggestion
                            <span>Skip restarting <%= $ok_results_explanation %> children</span>
```
which would render to "Skip restarting passed/softfailed children" so we wouldn't bother the users with our internal abstraction/generalization at all?

okurz (approved):

templates/webapi/test/infopanel.html.ep
alice-suse: @kalikiana  Thanks for involving me and delivering the PR. I am not quite familiar with code in this directory, but overall it looks good :-)

kalikiana: This really just documentation/ UX. The important thing is that it makes sense to users - I guess it may not be easy to get from the template files

kraih (approved):

PR #4442, opened 2022-01-11
Replace bugref_regex function with BUGREF_REGEX constant

kraih:
Small cleanup. The `bugref_regex` function should have been a constant.

### DIFF
--- a/lib/OpenQA/Utils.pm
+++ b/lib/OpenQA/Utils.pm
@@ -34,8 +34,46 @@ $Data::Dumper::Terse = 1;
 
 my $FRAG_REGEX = FRAGMENT_REGEX;
 
+my (%BUGREFS, %BUGURLS, $MARKER_REFS, $MARKER_URLS);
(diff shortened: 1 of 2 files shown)


lib/OpenQA/Utils.pm
```diff
@@ -34,8 +34,46 @@ $Data::Dumper::Terse = 1;
 
 my $FRAG_REGEX = FRAGMENT_REGEX;
 
+my (%BUGREFS, %BUGURLS, $MARKER_REFS, $MARKER_URLS);
+BEGIN {
```
okurz, line 38: this begin thingy looks a bit ugly though

kraih: Agreed, no way to avoid it though, constants can only be created at compile time, which requires a `BEGIN` block.

okurz (approved):

Martchus (approved):

PR #4444, opened 2022-01-11
Fix OpenQA::Utils import statements

kraih:
Small cleanup. It seems somewhere along the way the `OpenQA::Utils` imports got split up into two blocks.

### DIFF
--- a/lib/OpenQA/Utils.pm
+++ b/lib/OpenQA/Utils.pm
@@ -2,11 +2,8 @@
 # SPDX-License-Identifier: GPL-2.0-or-later
 
 package OpenQA::Utils;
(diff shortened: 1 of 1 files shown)


okurz (approved):

kalikiana (commented):
Can I ask what the "fix" is here? What difference does it make? Is this fixing warnings or circular dependency issues?

Martchus (approved):

kraih:
> Can I ask what the "fix" is here? What difference does it make? Is this fixing warnings or circular dependency issues?

Just a formatting fix, there should be no differences in behaviour. Unless one of the loaded modules changes global state, in which case this would help to find issues with that.

PR #4445, opened 2022-01-12
Update documentation to commit f1b020200

openqabot:


okurz (changes requested):
the test API content is deleted. This does not look ok

okurz:
See https://github.com/os-autoinst/openQA/pull/4446 for a potential fix

PR #4447, opened 2022-01-12
Try reloading not yet available text results

Martchus:
See https://progress.opensuse.org/issues/102786


kalikiana (approved):
I like the *textdatamissing* method, and also how you're writing the test. Very nice.

PR #4452, opened 2022-01-17
Give default value for uninitialized response code

FruitFly638:
res_code isn't initialized if no openQA server is running.
It now defaults to 0 if there is no connection, the "Connection refused" message still appears.

progress.opensuse.org/issues/104836

### DIFF
--- a/lib/OpenQA/CLI/api.pm
+++ b/lib/OpenQA/CLI/api.pm
@@ -53,7 +53,7 @@ sub command {
     $retries //= $ENV{OPENQA_CLI_RETRIES} // 0;
     do {
         $tx = $client->start($tx);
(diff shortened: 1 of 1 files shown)


lib/OpenQA/CLI/api.pm
```diff
@@ -54,6 +54,7 @@ sub command {
     do {
         $tx = $client->start($tx);
         my $res_code = $tx->res->code;
+        $res_code //= 0;
```
Martchus, line 57:
```suggestion
        my $res_code = $tx->res->code // 0;
```

okurz (approved):

Martchus (approved):

PR #4454, opened 2022-01-17
Fix rendering HTML elements when updating unique flash messages

Martchus:
This especially concerns the flash messages shown within the developer
mode panel, see https://progress.opensuse.org/issues/104917.

### DIFF
--- a/assets/javascripts/openqa.js
+++ b/assets/javascripts/openqa.js
@@ -30,6 +30,10 @@ function setupForAll() {
   });
 }
 
+function makeFlashElement(text) {
+  return typeof text === 'string' ? '<span>' + text + '</span>' : text;
+}
+
 function addFlash(status, text, container) {
   // add flash messages by default on top of the page
   if (!container) {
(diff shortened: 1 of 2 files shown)


okurz (approved):

assets/javascripts/openqa.js
```diff
@@ -58,9 +58,10 @@ function addUniqueFlash(status, id, text, container) {
     window.uniqueFlashMessages = {};
   }
   // update existing flash message
-  var existingFlashMessage = window.uniqueFlashMessages[id];
+  const existingFlashMessage = window.uniqueFlashMessages[id];
   if (existingFlashMessage) {
-    existingFlashMessage.find('span').first().text(text);
+    const element = typeof text === 'string' ? $('<span>' + text + '</span>') : text;
```
kraih, line 63: Why is there a `typeof text === 'string'` check? I went through the calls of this function and didn't see a case where it was not a `string`. Also, why is the element created with `$()`?

Martchus: All is just in consistency with the normal `addFlash`.

kraih: At least the `$(...)` doesn't do anything at all i think.

Martchus: I removed the `$(…)` in both functions.

okurz (approved):

kraih (approved):
Creative solution with `makeFlashElement`, i like it. 😁

PR #4458, opened 2022-01-18
Make openQA labels clearly visible

kraih:
Progress: https://progress.opensuse.org/issues/104616


kraih:
labels

okurz (approved):
just perfect :)

Martchus (approved):
It would be nice not to hard-code colors here (but to use ones from the theme or extend the theme). But it looks very fancy so I'll approve :-)

PR #4460, opened 2022-01-19
DO NOT MERGE: Test stability of t/full-stack.t

Martchus:
For https://progress.opensuse.org/issues/98952

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -156,24 +156,6 @@ jobs:
             fi
       - save_cache: *save_fullstack_cache
 
-  checkstyle: &test-template
-    docker:
-      - <<: *base
-    steps:
-      - checkout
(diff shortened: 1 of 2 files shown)


okurz:
fullstack timed out after 1h after 15 successful runs.

Martchus:
Yes, it ran into the CircleCI time limit. It doesn't look like the test was stuck, it just took that long to execute the (almost) 16 runs. So I'm just restarting the job to get more runs.

okurz:
> Yes, it ran into the CircleCI time limit. It doesn't look like the test was stuck, it just took that long to execute the (almost) 16 runs. So I'm just restarting the job to get more runs.

Or spawn the jobs in parallel, not serial :)

Martchus:
Exact the same result as before, so we have 30 successful runs now.

PR #4463, opened 2022-01-20
Remove retry from fullstack test as it seems stable again

Martchus:
* Conducted 75 runs in the CI which all passed so it seems stable again
* See https://progress.opensuse.org/issues/98952


okurz (approved):

okurz:
kinda trivial, should be ok

PR #4464, opened 2022-01-20
Add retry to rsync in cacheservice

mimi1vx:
https://progress.opensuse.org/issues/103771

### DIFF
--- a/lib/OpenQA/CacheService/Task/Sync.pm
+++ b/lib/OpenQA/CacheService/Task/Sync.pm
@@ -48,7 +48,7 @@ sub _cache_tests {
         $status = $? >> 8;
         $full_output .= "Try $retry:\n" . $output . "\n";
         last unless $status;
-        sleep RSYNC_RETRY_PERIOD;    # uncoverable statement
+        sleep RSYNC_RETRY_PERIOD;
     }
     $job->note(output => "[info] [#$job_id] Calling: $cmd\n$full_output");
     $job->finish("exit code $status");
--- a/lib/OpenQA/CacheService/Task/Sync.pm
+++ b/lib/OpenQA/CacheService/Task/Sync.pm
@@ -8,6 +8,8 @@ use Mojo::URL;
 use Time::Seconds;
 
 use constant RSYNC_TIMEOUT => $ENV{OPENQA_RSYNC_TIMEOUT} // (30 * ONE_MINUTE);
+use constant RSYNC_RETRIES => $ENV{OPENQA_RSYNC_RETRIES} // 3;
+use constant RSYNC_RETRY_PERIOD => $ENV{OPENQA_RSYNC_RETRY_PERIOD} // 3;
 
 sub register {
     my ($self, $app) = @_;
@@ -39,10 +41,17 @@ sub _cache_tests {
     my @cmd = (qw(rsync -avHP --timeout), RSYNC_TIMEOUT, "$from/", qw(--delete), "$to/tests/");
     my $cmd = join ' ', @cmd;
     $ctx->info("Calling: $cmd");
-    my $output = `@cmd`;
-    my $status = $? >> 8;
+    my $status;
+    my $full_output = "";
+    for my $retry (1 .. RSYNC_RETRIES) {
+        my $output = `@cmd`;
+        $status = $? >> 8;
+        $full_output .= "Try $retry:\n" . $output . "\n";
+        last unless $status;
+        sleep RSYNC_RETRY_PERIOD;    # uncoverable statement
+    }
+    $job->note(output => "[info] [#$job_id] Calling: $cmd\n$full_output");
     $job->finish("exit code $status");
-    $job->note(output => "[info] [#$job_id] Calling: $cmd\n$output");
     $ctx->info("Finished sync: $status");
 }
 
--- a/t/25-cache-service.t
+++ b/t/25-cache-service.t
@@ -4,7 +4,7 @@
 
 use Test::Most;
 use Time::Seconds;
-$ENV{MOJO_LOG_LEVEL} = 'trace';
+$ENV{MOJO_LOG_LEVEL} = 'info';
 
 my $tempdir;
 BEGIN {
@@ -144,9 +144,7 @@ sub test_download {
 
     my $status = $cache_client->status($asset_request);
     perform_minion_jobs($t->app->minion);
-    my $worker = $t->app->minion->worker->register;
     $status = $cache_client->status($asset_request) until !$status->is_downloading;
-    $worker->unregister;
 
     # And then goes to PROCESSED state
     ok $status->is_processed, 'only other state is processed';
@@ -366,14 +364,18 @@ subtest 'Default usage' => sub {
     ok(!$cache_client->asset_exists('localhost', $asset), 'Asset absent')
       or die diag "Asset already exists - abort test";
 
+    my @logs;
+    my $log_mock = Test::MockModule->new('Mojo::Log');
+    $log_mock->redefine(info => sub { push @logs, $_[1] });
     BAIL_OUT("Failed enqueuing download") if $cache_client->enqueue($asset_request);
     perform_minion_jobs($t->app->minion);
     wait_for_or_bail_out { $cache_client->status($asset_request)->is_processed } 'asset';
     my $worker = $t->app->minion->repair->worker->register;
+    my $status = $cache_client->status($asset_request);
     my $out = $cache_client->status($asset_request)->output;
     $worker->unregister;
-    ok($out, 'Output should be present') or die diag $out;
-    like $out, qr/Downloading "$asset" from/, "Asset download attempt logged";
+    ok(scalar @logs, 'Output should be present') or die diag $out;
+    like "@logs", qr/Downloading "$asset" from/, "Asset download attempt logged";
     ok(-e path($cachedir, 'localhost')->child($asset), 'Asset downloaded') or die diag "Failed - no asset is there";
 };
 
--- a/t/25-cache-service.t
+++ b/t/25-cache-service.t
@@ -4,6 +4,7 @@
 
 use Test::Most;
 use Time::Seconds;
+$ENV{MOJO_LOG_LEVEL} = 'trace';
 
 my $tempdir;
 BEGIN {
@@ -30,6 +31,8 @@ use FindBin;
 use lib "$FindBin::Bin/lib", "$FindBin::Bin/../external/os-autoinst-common/lib";
 
 use Test::Warnings ':report_warnings';
+use Test::Output qw(stderr_like);
+use Test::Mojo;
 use OpenQA::Utils;
 use IO::Socket::INET;
 use Mojo::Server::Daemon;
@@ -37,7 +40,8 @@ use Mojo::IOLoop::Server;
 use POSIX '_exit';
 use Mojo::IOLoop::ReadWriteProcess qw(queue process);
 use Mojo::IOLoop::ReadWriteProcess::Session 'session';
-use OpenQA::Test::Utils qw(fake_asset_server cache_minion_worker cache_worker_service wait_for_or_bail_out);
+use OpenQA::Test::Utils
+  qw(fake_asset_server cache_minion_worker cache_worker_service wait_for_or_bail_out perform_minion_jobs);
 use OpenQA::Test::TimeLimit '90';
 use Mojo::Util qw(md5_sum);
 use OpenQA::CacheService;
@@ -54,7 +58,7 @@ END { session->clean }
 
 my $daemon;
 my $cache_service = cache_worker_service;
-my $worker_cache_service = cache_minion_worker;
+my $t = Test::Mojo->new('OpenQA::CacheService');
 
 my $server_instance = process sub {
     # Connect application with web server and start accepting connections
@@ -71,7 +75,7 @@ my $server_instance = process sub {
 sub start_server {
     $server_instance->set_pipes(0)->separate_err(0)->blocking_stop(1)->channels(0)->restart;
     $cache_service->set_pipes(0)->separate_err(0)->blocking_stop(1)->channels(0)->restart->restart;
-    $worker_cache_service->restart;
+    perform_minion_jobs($t->app->minion);
     wait_for_or_bail_out { $cache_client->info->available } 'cache service';
 }
 
@@ -80,6 +84,7 @@ sub test_default_usage {
     my $asset_request = $cache_client->asset_request(id => $id, asset => $asset, type => 'hdd', host => $host);
 
     if (!$cache_client->enqueue($asset_request)) {
+        perform_minion_jobs($t->app->minion);
         wait_for_or_bail_out { $cache_client->status($asset_request)->is_processed } 'asset';
     }
     ok($cache_client->asset_exists('localhost', $asset), "Asset $asset downloaded");
@@ -99,6 +104,7 @@ sub test_sync {
 
     ok !$cache_client->enqueue($rsync_request);
 
+    perform_minion_jobs($t->app->minion);
     wait_for_or_bail_out { $cache_client->status($rsync_request)->is_processed } 'rsync';
 
     my $status = $cache_client->status($rsync_request);
@@ -111,10 +117,15 @@ sub test_sync {
     ok -e $expected, "expected file exists, run $run";
     is $expected->slurp, $data, "synced data identical, run $run";
 
-    my $rsync_request2 = $cache_client->rsync_request(from => "empty", to => "nowhewre");
-    ok !$cache_client->enqueue($rsync_request2);
+    my $rsync_request2;
+    stderr_like {
+        $rsync_request2 = $cache_client->rsync_request(from => "empty", to => "nowhewre");
+        ok !$cache_client->enqueue($rsync_request2);
 
-    wait_for_or_bail_out { $cache_client->status($rsync_request2)->is_processed } 'rsync';
+        perform_minion_jobs($t->app->minion);
+        wait_for_or_bail_out { $cache_client->status($rsync_request2)->is_processed } 'rsync';
+    }
+    qr/rsync error:.*/, "rsync error message";
 
     my $status2 = $cache_client->status($rsync_request2);
     is $status2->result, 'exit code 11', "exit code ok, run $run";
@@ -132,7 +143,10 @@ sub test_download {
     ok !$cache_client->enqueue($asset_request), "enqueued id $id, asset $asset";
 
     my $status = $cache_client->status($asset_request);
+    perform_minion_jobs($t->app->minion);
+    my $worker = $t->app->minion->worker->register;
     $status = $cache_client->status($asset_request) until !$status->is_downloading;
+    $worker->unregister;
 
     # And then goes to PROCESSED state
     ok $status->is_processed, 'only other state is processed';
@@ -286,16 +300,18 @@ subtest 'Job progress (guard against parallel downloads of the same file)' => su
 };
 
 subtest 'Client can check if there are available workers' => sub {
-    $worker_cache_service->stop;
     $cache_service->stop;
     ok !$cache_client->info->available, 'Cache server is not available';
     $cache_service->restart;
+    perform_minion_jobs($t->app->minion);
     wait_for_or_bail_out { $cache_client->info->available } 'cache service';
     ok $cache_client->info->available, 'Cache server is available';
     ok !$cache_client->info->available_workers, 'No available workers at the moment';
-    $worker_cache_service->start;
+    perform_minion_jobs($t->app->minion);
+    my $worker = $t->app->minion->worker->register;
     wait_for_or_bail_out { $cache_client->info->available_workers } 'minion_worker';
     ok $cache_client->info->available_workers, 'Workers are available now';
+    $worker->unregister;
 };
 
 subtest 'Asset download' => sub {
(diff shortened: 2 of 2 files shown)


lib/OpenQA/CacheService/Task/Sync.pm
```diff
@@ -39,10 +39,16 @@ sub _cache_tests {
...
-    my $status = $? >> 8;
+    my $retry = 2;
+    while ($retry) {
+        $retry--;
+        my $output = `@cmd`;
+        my $status = $? >> 8;
+        $job->note(output => "[info] [#$job_id] Calling: $cmd\n$output");
+        last unless $status;
+        sleep 3;
+    }
```
okurz, line 50:
```suggestion
use constant RSYNC_RETRIES => $ENV{OPENQA_RSYNC_RETRIES} // 7;
use constant RSYNC_RETRY_PERIOD => $ENV{OPENQA_RSYNC_RETRY_PERIOD} // 3;
…
    for my $retry (1 .. RSYNC_RETRIES) {
        my $output = `@cmd`;
        my $status = $? >> 8;
        $job->note(output => "[info] [#$job_id] Calling try $retry: $cmd\n$output");
        last unless $status;
        sleep RSYNC_RETRY_PERIOD;
    }
```

okurz (changes requested):

lib/OpenQA/CacheService/Task/Sync.pm
mimi1vx: 7 retries ?

okurz: Yes, I always default to 7 retries as a magic number :) I am ok to keep 3 though, then just do the rest of the code suggestions

mimi1vx: j, usage of ENV vars is way better than some random numbers in code

lib/OpenQA/CacheService/Task/Sync.pm
```diff
@@ -39,10 +41,15 @@ sub _cache_tests {
     my @cmd = (qw(rsync -avHP --timeout), RSYNC_TIMEOUT, "$from/", qw(--delete), "$to/tests/");
     my $cmd = join ' ', @cmd;
     $ctx->info("Calling: $cmd");
-    my $output = `@cmd`;
-    my $status = $? >> 8;
+    my $status;
+    for my $retry (1 .. RSYNC_RETRIES) {
+        my $output = `@cmd`;
+        $status = $? >> 8;
+        $job->note(output => "[info] [#$job_id] Calling calling try $retry: $cmd\n$output");
```
kraih, line 48: "Calling calling"?

t/25-cache-service.t
```diff
@@ -104,7 +104,7 @@ sub test_sync {
     is $status->result, 'exit code 0', "exit code ok, run $run";
     ok $status->output, "output ok, run $run";
 
-    like $status->output, qr/Calling: rsync .* --timeout 1800 .*100\%/s, "output correct, run $run"
+    like $status->output, qr/Calling try \d: rsync .* --timeout 1800 .*100\%/s, "output correct, run $run"
```
kraih, line 107: This does not actually test that it tried more than once.

Martchus: Yes, and it would be nice to have a test.

lib/OpenQA/CacheService/Task/Sync.pm
```diff
@@ -39,10 +41,15 @@ sub _cache_tests {
     my @cmd = (qw(rsync -avHP --timeout), RSYNC_TIMEOUT, "$from/", qw(--delete), "$to/tests/");
     my $cmd = join ' ', @cmd;
     $ctx->info("Calling: $cmd");
-    my $output = `@cmd`;
-    my $status = $? >> 8;
+    my $status;
+    for my $retry (1 .. RSYNC_RETRIES) {
+        my $output = `@cmd`;
+        $status = $? >> 8;
+        $job->note(output => "[info] [#$job_id] Calling try $retry: $cmd\n$output");
```
kraih, line 48: Is it intentional that the latest result overwrites the previous one? I think for other retry features we log all attempts.

okurz: I guess it's ok either way

okurz (approved):

lib/OpenQA/CacheService/Task/Sync.pm
mimi1vx: no ,isn't , thx for spotting.

okurz (approved):

Martchus (approved):

lib/OpenQA/CacheService/Task/Sync.pm
```diff
@@ -39,10 +41,17 @@ sub _cache_tests {
...
-    my $output = `@cmd`;
-    my $status = $? >> 8;
+    my $status;
+    my $full_output;
+    for my $retry (1 .. RSYNC_RETRIES) {
+        my $output = `@cmd`;
+        $status = $? >> 8;
+        $full_output = "Try $retry:\n" . $output;
+        last unless $status;
+        sleep RSYNC_RETRY_PERIOD;
```
okurz, line 51: To cover this in tests without needles waiting I suggest to use https://metacpan.org/pod/Test::Mock::Time which we already use in os-autoinst. You just need to specify the package as dependency in openQA but it's already installed in the CI container due to the os-autoinst dependency anyway

Martchus: Or you simply set `OPENQA_RSYNC_RETRY_PERIOD=0` in the test environment. (Needs to be in a `BEGIN` block as it is a constant.)

Martchus (approved):

lib/OpenQA/CacheService/Task/Sync.pm
```diff
@@ -39,10 +41,17 @@ sub _cache_tests {
...
     my $cmd = join ' ', @cmd;
     $ctx->info("Calling: $cmd");
-    my $output = `@cmd`;
-    my $status = $? >> 8;
+    my $status;
+    my $full_output;
+    for my $retry (1 .. RSYNC_RETRIES) {
+        my $output = `@cmd`;
+        $status = $? >> 8;
+        $full_output = "Try $retry:\n" . $output;
```
kraih, line 49: This still throws away previous attempts in `$full_output`.

lib/OpenQA/CacheService/Task/Sync.pm
mimi1vx: > Or you simply set `BEGIN { $ENV{OPENQA_RSYNC_RETRY_PERIOD} = 0 }` in the test environment. (Needs to be in a `BEGIN` block as it is a constant.)

tried, not working:( so time for Test::Mock::Time

Martchus: Strange, it should've been working - especially since there are already similar overrides in the same test file.

mimi1vx: and `Test::More::Time` broke *Subtest: Minion monitoring with InfluxDB* , so probably best option is mark `sleep` line as non-coverable :(

Martchus: So you would want to mark the line as non-coverable because you want to put in some extra condition to not run the code at all? I'm just asking because the issue we want to avoid is that tests take longer due to the sleep.

lib/OpenQA/CacheService/Task/Sync.pm
```diff
@@ -39,10 +41,17 @@ sub _cache_tests {
...
-    my $output = `@cmd`;
-    my $status = $? >> 8;
+    my $status;
+    my $full_output = "";
+    for my $retry (1 .. RSYNC_RETRIES) {
+        my $output = `@cmd`;
+        $status = $? >> 8;
+        $full_output .= "Try $retry:\n" . $output . "\n";
+        last unless $status;
+        sleep RSYNC_RETRY_PERIOD; # uncoverable statement
```
Martchus, line 51: Ok, so this line isn't executed anyways by tests. How do you come to the conclusion that my suggestion to set `OPENQA_RSYNC_RETRY_PERIOD` to 0 isn't working then? Of course if the line isn't executed anyways we don't have to worry much about slowing down tests although it would likely be nice to actually have a test which runs all of the code.

Martchus (dismissed):

lib/OpenQA/CacheService/Task/Sync.pm
mimi1vx: https://github.com/os-autoinst/openQA/pull/4464/files#diff-9a2b1ade12881dfa535761e60a94118eabd51ecdad9a003037400c6cad1df61dR114-R117 this should execute whole for loop include `sleep` statement ...

okurz (changes requested):
.

lib/OpenQA/CacheService/Task/Sync.pm
Martchus: Then we shouldn't need to mark the line as uncoverable statement (unless it is running in a way which doesn't allow us to track its coverage, e.g. in a forked process).

mimi1vx: @kraih how `sleep` behave in minion job?

mimi1vx: > Then we shouldn't need to mark the line as uncoverable statement (unless it is running in a way which doesn't allow us to track its coverage, e.g. in a forked process).

so question is why codecov consider this line not running.. whed I added debug lines before and after they show correctly in test run

Martchus: Oh, when it is within a Minion job that's a forked process and therefore the coverage collection doesn't work. We actually have the utility function `perform_minion_jobs` to avoid forking which might be applicable here as well.

kraih: What @Martchus said. It's a bit tricky, since depending on test strategy Minion jobs might run without forking.

okurz:
https://app.circleci.com/pipelines/github/os-autoinst/openQA/8847/workflows/585504f3-5933-4802-a650-7c2e8b48f245/jobs/83520 shows many error messages that should be avoided

mimi1vx:
> https://app.circleci.com/pipelines/github/os-autoinst/openQA/8847/workflows/585504f3-5933-4802-a650-7c2e8b48f245/jobs/83520 shows many error messages that should be avoided

`rsync: .....` on stderr is excepted 3x per test run , and subtest itself calls `test_sync` 4x

lib/OpenQA/CacheService/Task/Sync.pm
mimi1vx: no idea how to use `perform_minion_jobs` in this case

okurz:
> > https://app.circleci.com/pipelines/github/os-autoinst/openQA/8847/workflows/585504f3-5933-4802-a650-7c2e8b48f245/jobs/83520 shows many error messages that should be avoided
> 
> `rsync: .....` on stderr is excepted 3x per test run , and subtest itself calls `test_sync` 4x

You can use Test::Output to catch the expected error output. We use that in other tests as well. Just grep for `Test::Output`

mimi1vx:
> You can use Test::Output to catch the expected error output. We use that in other tests as well. Just grep for `Test::Output`

I think this can test if `rsync` output is excepted , but not suppress error messages

t/25-cache-service.t
```diff
@@ -109,6 +110,18 @@ sub test_sync {
...
     is $expected->slurp, $data, "synced data identical, run $run";
+
+    my $rsync_request2 = $cache_client->rsync_request(from => "empty", to => "nowhewre");
+    ok !$cache_client->enqueue($rsync_request2);
+
+    wait_for_or_bail_out { $cache_client->status($rsync_request2)->is_processed } 'rsync';
+
+    my $status2 = $cache_client->status($rsync_request2);
+    is $status2->result, 'exit code 11', "exit code ok, run $run";
+    ok $status2->output, "output ok, run $run";
```
perlpunk, line 121: One more thing you could do here is to test the actual output against a regex instead of just testing that it is not empty.

lib/OpenQA/CacheService/Task/Sync.pm
perlpunk: I'm trying to add `perform_minion_jobs`, but this will take a while.

t/25-cache-service.t
mimi1vx: isnt this in next line ?

perlpunk:
> I think this can test if `rsync` output is excepted , but not suppress error messages

I played with `perform_minion_jobs`, and when using that, I can catch the output with `stderr_like`.
However, it is not that trivial to add `perform_minion_jobs` just to one place. We discussed it in the weekly unblock meeting today, that it should be added to the whole test.
I could try that and if I get it working, I'll send you a commit.

t/25-cache-service.t
perlpunk: oh I'm stupid. Ok, but maybe actually add the `Try 1:` stuff to the regex to make sure it actually retries

okurz:
I also suggested an alternative which is to have simple unit-tests for the not covered parts and mock out the lower level code that is executed from minions for the higher-level tasks

t/25-cache-service.t
```diff
@@ -118,8 +142,11 @@ sub test_download {
 
     ok !$cache_client->enqueue($asset_request), "enqueued id $id, asset $asset";
 
+    my $worker = $t->app->minion->worker->register;    # doesn't help
```
perlpunk, line 145: That line and the next `unregister` line below were leftovers from my testing, they are not needed.

t/25-cache-service.t
```diff
@@ -308,9 +337,9 @@ subtest 'Race for same asset' => sub {
 
     my $concurrent_test = sub {
         if (!$cache_client->enqueue($asset_request)) {
+            perform_minion_jobs($t->app->minion);
             wait_for_or_bail_out { $cache_client->status($asset_request)->is_processed } 'asset';
             my $ret = $cache_client->asset_exists('localhost', $asset);
-            Devel::Cover::report() if Devel::Cover->can('report');
```
perlpunk, line 313: I think this line has to be added again

kraih: Since you're running the jobs without forking you might be fine now without it.

perlpunk: I saw in the codecov report that some lines lost their coverage:
https://app.codecov.io/gh/os-autoinst/openQA/compare/4464/changes#D1L339
So I would add it back and see if that changes.
There are still forked processes in the test, right? just not the cacheservice workers?

t/25-cache-service.t
```diff
@@ -4,6 +4,7 @@
 
 use Test::Most;
 use Time::Seconds;
+local $ENV{MOJO_LOG_LEVEL} = 'trace';
```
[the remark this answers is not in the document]
perlpunk: I'm so used to always put a `local` before changing env vars... :)

t/25-cache-service.t
okurz: Yes, please ensure that t/25-cache-service.t has 100% statement coverage. It does not make sense to have lines in tests that are never or not reliably executed – with explicitly marked uncoverable exceptions of course.

t/25-cache-service.t
```diff
@@ -366,7 +400,6 @@ subtest 'Asset download with default usage' => sub {
 };
 
 # The following tests start their own workers
```
Martchus, line 402: All tests now start their own workers, right? So this comment could be removed as well.

okurz (changes requested):
Looks like unfortunately now the test is flaky *and* has unhandled output on stdout, see https://app.circleci.com/pipelines/github/os-autoinst/openQA/8889/workflows/6f71d3ec-2990-441a-bcfc-6d7f4d43e301/jobs/83929

t/25-cache-service.t
perlpunk: hm, this line and the `unregister` line is still in there, just that you removed the comment

perlpunk:
I guess the additional output is from `$ENV{MOJO_LOG_LEVEL} = 'trace';`
Maybe there is a way to only set that locally? (I tried `local` in the subtest, but that didn't work)

perlpunk:
The test failed for me once when I added the `stderr_like` but then I enclosed more commands in it and then the test passed. But that was probably the flaky behaviour that we see in CI now :(

perlpunk:
I pushed a commit that captures the unhandled log messages, while still assuring that the test excepting a log message `ok(scalar @logs, 'Output should be present')` is passing.

I wasn't able to fix the flakyness. The rsync errors sometimes happen outside of the `stderr_like`, no matter where I let it start and end.

Martchus (dismissed):

okurz (approved):
https://app.circleci.com/pipelines/github/os-autoinst/openQA/8910/workflows/260a33d7-fae4-4c4d-be2b-b889e6af4631/jobs/84134 looks good so it seems like you now just try to cover a bigger block with the stderr catching so I guess that's ok

kraih (dismissed):

perlpunk:
> https://app.circleci.com/pipelines/github/os-autoinst/openQA/8910/workflows/260a33d7-fae4-4c4d-be2b-b889e6af4631/jobs/84134 looks good so it seems like you now just try to cover a bigger block with the stderr catching so I guess that's ok

no, i got a failure with this code locally. 
i mentioned this above.

okurz:
@perlpunk so what do you suggest we do? This PR is waiting for your review now

kalikiana:
> @perlpunk so what do you suggest we do? This PR is waiting for your review now

There's two approvals and no open requests for changes. That should be enough

perlpunk:
My review was explicitly requested.
But @okurz was the one who saw the flakyiness https://github.com/os-autoinst/openQA/pull/4464#pullrequestreview-866277438
But if @okurz still approves the PR then I guess it's ok for him.
I still would like to find out what's happening.
Since it's not happening that often we could
* accept this PR
* create a ticket to investigate the problem with the error output

okurz (changes requested):
I might have wrongly assumed that mimi1vx would have addressed your comment so I assumed the sporadic issue would have been addressed. So better we fix it here

kraih:
Guess i'll also invalidate my review then, since i'm a bit confused now. From the looks of it, the code seems reasonable and CI tests passed... 🤷

mimi1vx:
on my instance tests were flaky also before change, but I'm not experienced with test suite enough to decide cause or outcome..

okurz:
yes, it's correct that t/25-cache-service.t is unstable. This is why it's part of tools/unstable_tests.txt. It would be nice to fix the flakyness but here the focus was on changing behaviour in the log output. That's how I understood. @perlpunk can you help to state more explicitly how you currently reproduce failures and which ones in particular?

perlpunk:
I was getting test failures where the rsync error messages were written outside of the `stderr_like` and the `stderr_like` failed because it didn't catch anything.
```
rsync: change_dir "/home/squamata/project//empty" failed: No such file or directory (2)
rsync: mkdir "/home/squamata/project/nowhewre/tests" failed: No such file or directory (2)
rsync error: error in file IO (code 11) at main.c(664) [Receiver=3.1.3]
...
```
It was happening like 1 / 4 times or so.
Now I ran it 25 times without a failure, although the code is the same.

So I guess I'll approve.

perlpunk (approved):

okurz (approved):

perlpunk:
argh. I rebased to master now, and I got the failure at the 6th run. I think last time when I tested it I had it also rebased on master.

I rebase my branches on master as often as possible, because of things like this.

Martchus:
Maybe the instability is unrelated then (e.g. would also occur on master before merging this PR).

perlpunk:
for the record:
```
t/25-cache-service.t .. 17/? rsync: change_dir "/home/tina/openqa-devel/repos/openQA//empty" failed: No such file or directory (2)
rsync: mkdir "/home/tina/openqa-devel/repos/openQA/nowhewre/tests" failed: No such file or directory (2)
rsync error: error in file IO (code 11) at main.c(664) [Receiver=3.1.3]
rsync: change_dir "/home/tina/openqa-devel/repos/openQA//empty" failed: No such file or directory (2)
rsync: mkdir "/home/tina/openqa-devel/repos/openQA/nowhewre/tests" failed: No such file or directory (2)
rsync error: error in file IO (code 11) at main.c(664) [Receiver=3.1.3]
rsync: change_dir "/home/tina/openqa-devel/repos/openQA//empty" failed: No such file or directory (2)
rsync: mkdir "/home/tina/openqa-devel/repos/openQA/nowhewre/tests" failed: No such file or directory (2)
rsync error: error in file IO (code 11) at main.c(664) [Receiver=3.1.3]

    #   Failed test 'rsync error message'
    #   at t/25-cache-service.t line 128.
    # STDERR:
    # 
    # doesn't match:
    # (?^:rsync error:.*)
    # as expected
    # Looks like you failed 1 test of 44.
t/25-cache-service.t .. 22/? 
#   Failed test 'OpenQA::CacheService::Task::Sync'
#   at t/25-cache-service.t line 647.
t/25-cache-service.t .. 23/? # Looks like you failed 1 test of 23.
t/25-cache-service.t .. Dubious, test returned 1 (wstat 256, 0x100)
Failed 1/23 subtests 

```

perlpunk:
Ok, I tried to bisect and apparently it was a coincidence. So it seems to be really rare.

PR #4468, opened 2022-01-21
Use tidyall instead of custom implementation

okurz:
Same as in os-autoinst, see
https://github.com/os-autoinst/os-autoinst/pull/1795

### DIFF
--- a/.tidyallrc
+++ b/.tidyallrc
@@ -1,0 +1,3 @@
+[PerlTidy]
+select = **/*.{pl,pm,t}
+argv = --profile=$ROOT/.perltidyrc
--- a/.gitignore
+++ b/.gitignore
@@ -44,3 +44,4 @@ base_state.json
 job.json
 node_modules
 package-lock.json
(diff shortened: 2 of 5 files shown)


okurz:
@Martchus looks like fullstack isn't stable after all: https://app.circleci.com/pipelines/github/os-autoinst/openQA/8827/workflows/cf2177e1-d1d2-4e7a-a638-db8593f35f8a/jobs/83317 in this really unrelated change

kalikiana (approved):

baierjan (approved):

.tidyallrc
```diff
@@ -0,0 +1,3 @@
+[PerlTidy]
+select = **/*.{pl,pm,t}
```
perlpunk, line 2: This ignores all perl scripts without a file extension, e.g. under `script/`.

okurz: https://github.com/os-autoinst/openQA/pull/4485 should fix that

PR #4470, opened 2022-01-21
Simplify base inheritance statement with Mojo::Base everywhere

okurz:
Wherever we use "Mojo::Base" as well as "base" we can simply combine the
both including some redundant "strict" and "warnings" statements which
are also implicitly included with Mojo::Base.

### DIFF
--- a/t/01-style.t
+++ b/t/01-style.t
@@ -13,4 +13,5 @@ qq{git grep -I -l 'This program is free software.*if not, see <http://www.gnu.or
 ok system(qq{git grep -I -l '[#/ ]*SPDX-License-Identifier ' ':!COPYING' ':!external/' ':!t/01-style.t'}) != 0,
   'SPDX-License-Identifier correctly terminated';
 is qx{git grep -I -L '^use Test::Most' t/**.t}, '', 'All tests use Test::Most';
+is qx{git grep --all-match -e '^use Mojo::Base' -e 'use base'}, '', 'No redundant Mojo::Base+base';
 done_testing;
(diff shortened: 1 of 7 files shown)


baierjan (approved):

Martchus (approved):

t/01-style.t
```diff
@@ -13,4 +13,5 @@ qq{git grep -I -l 'This program is free software.*if not, see <http://www.gnu.or
 ok system(qq{git grep -I -l '[#/ ]*SPDX-License-Identifier ' ':!COPYING' ':!external/' ':!t/01-style.t'}) != 0,
   'SPDX-License-Identifier correctly terminated';
 is qx{git grep -I -L '^use Test::Most' t/**.t}, '', 'All tests use Test::Most';
+is qx{git grep -I -e '^use Mojo::Base' --and -e 'use base'}, '', 'No redundant Mojo::Base+base';
```
perlpunk, line 16: Is this command supposed to output anything in our current master?

perlpunk (changes requested):
See my comment about git-grep

t/01-style.t
okurz: Yes, doesn't it?

perlpunk: no:
```
% git describe
4.5.1528009330.e68ebe2b-6119-g6bb94d7ba
% git grep -I -e '^use Mojo::Base' --and -e 'use base'
% echo $?
1
% git --version
git version 2.26.2
```

okurz: Yes, I will cross-check to have a failing test first

okurz: Updated with a better match rule that lists all the files changed in this PR before the patch is applied. Seems I confused commands when I included the previous one in here without crosschecking after applying if it's still the correct one. Thanks for diligent checking.

perlpunk (approved):

PR #4471, opened 2022-01-21
Add some perl signatures in OpenQA/

FruitFly638:
Some more perl signatures in OpenQA
Also, especially for BuildResults, where does the `-signatures` go? Do I do something like `use feature 'signatures'`?

### DIFF
--- a/lib/OpenQA/App.pm
+++ b/lib/OpenQA/App.pm
@@ -2,14 +2,11 @@
 # SPDX-License-Identifier: GPL-2.0-or-later
 
 package OpenQA::App;
-use Mojo::Base -strict;
+use Mojo::Base -strict, -signatures;
 
 my $SINGLETON;
 
-sub set_singleton {
-    my ($class, $app) = @_;
-    $SINGLETON = $app;
-}
+sub set_singleton ($class, $app) { $SINGLETON = $app; }
 
 sub singleton { $SINGLETON }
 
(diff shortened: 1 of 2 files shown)


lib/OpenQA/App.pm
```diff
@@ -2,12 +2,11 @@
...
 
 package OpenQA::App;
-use Mojo::Base -strict;
+use Mojo::Base -strict, -signatures;
 
 my $SINGLETON;
 
-sub set_singleton {
-    my ($class, $app) = @_;
+sub set_singleton ($class, $app) {
```
perlpunk, line 9: This function is now short enough to be put in one line.

lib/OpenQA/BuildResults.pm
```diff
@@ -13,8 +13,7 @@ use Date::Format;
 use Sort::Versions;
 use Time::Seconds;
 
-sub init_job_figures {
-    my ($job_result) = @_;
+sub init_job_figures ($job_result) {
```
perlpunk, line 16: Just replace line 6 and 7 with
```
```

lib/OpenQA/App.pm
FruitFly638: Okay, I'll do that

Martchus (approved):

perlpunk (approved):

PR #4475, opened 2022-01-24
Show "failed modules" with a coloring resembling an failed state

okurz:
Especially with the new openQA feature of "force_result" it is more
common to see openQA jobs where the result is passed or soft-failed
while there are still failed modules. Right now e.g. in /tests/overview
"failed modules" show up on a green background which could confuse
people thinking of these modules as not failed. We should reconsider
the coloring. This commit changes the styling so that the failed modules
in openQA are rendered with a slightly more alarming design using a
light red background and red border which I made a bit thicker.

**Before:**
Screenshot_20220124_125343_before_failed_modules_green_border

**After:**
Screenshot_20220124_125607_after_failed_modules_red_border

Related progress issue: https://progress.opensuse.org/issues/95479

### DIFF
--- a/assets/stylesheets/tables.scss
+++ b/assets/stylesheets/tables.scss
@@ -129,11 +129,11 @@ table.overview {
     }
 
     .failedmodule {
-        background-color: #E9F7C9;
+        background-color: #F7E1E1;
         padding: 3px;
         margin-right: 1ex;
         border-radius: 3px;
-        border: 1px solid #D2DFB5;
+        border: 2px solid #D98A8A;
         @include media-breakpoint-down(sm) {
             display: none;
         }


baierjan (approved):

okurz:
Adding label "not-ready" to give a chance to more people to provide their feedback.

perlpunk (approved):

Martchus (approved):

dcermak (commented):
I would suggest to make the border less prominent. Right now it looks like these nodes are selected due to the outline.

okurz:
> I would suggest to make the border less prominent. Right now it looks like these nodes are selected due to the outline.

Valid point. How about the middle ground of 2px or do we need to change the color or style differently?

jknphy:
my 2 cents, for soft-failures I would prefer keep the background yellow, it might be a bit distracting. For job groups with few test suites could be useful, but for job groups with huge number of test suite is not something you are going to check when you review it, it is better to trust in the automation that post those failures and group them, so visually doesn't help in those cases to the reviewer.

okurz:
> my 2 cents, for soft-failures I would prefer keep the background yellow, it might be a bit distracting.

The problem is that the background for failed modules was never yellow but green so viewers can be confused about "green == failed"

> For job groups with few test suites could be useful, but for job groups with huge number of test suite is not something you are going to check when you review it, it is better to trust in the automation that post those failures and group them, so visually doesn't help in those cases to the reviewer.

I agree. There it shouldn't make a difference.

jknphy:
> the background for failed modules was never yellow but green so viewers can be confused about "green == failed"

Still I think it is distracting that the soft-failure shows with red background, I would prefer it to see it green that red in this PR, and perhaps update the title to reflect both intentions, those are not 'failed' modules.

dcermak:
> > I would suggest to make the border less prominent. Right now it looks like these nodes are selected due to the outline.
> 
> Valid point. How about the middle ground of 2px or do we need to change the color or style differently?

Honestly, I would consult someone from the design team about this.

okurz:
> Honestly, I would consult someone from the design team about this.

I would like that :) Can we ping someone over github easily?

dcermak:
> > Honestly, I would consult someone from the design team about this.
> 
> I would like that :) Can we ping someone over github easily?

cc @cyntss :point_up:

okurz:
No response within a week. That's fine. I would like to go ahead. After all we can still easily tweak it afterwards. Any feedback afterwards is still appreciated.

PR #4476, opened 2022-01-24
Add some more perl signatures

FruitFly638:
Once more, some more perl signatures
These are all in lib/openQA/CacheService

### DIFF
--- a/lib/OpenQA/CacheService/Command/run.pm
+++ b/lib/OpenQA/CacheService/Command/run.pm
@@ -2,16 +2,14 @@
 # SPDX-License-Identifier: GPL-2.0-or-later
 
 package OpenQA::CacheService::Command::run;
-use Mojo::Base 'Minion::Command::minion::worker';
+use Mojo::Base 'Minion::Command::minion::worker', -signatures;
 
 use Mojo::Util 'getopt';
 
 has description => 'Start Minion worker';
 has usage => sub { shift->extract_usage };
 
-sub run {
-    my ($self, @args) = @_;
-
+sub run ($self, @args) {
     getopt \@args, ['pass_through'], 'reset-locks' => \my $reset_locks;
 
     if ($reset_locks) {
--- a/lib/OpenQA/CacheService/Model/Cache.pm
+++ b/lib/OpenQA/CacheService/Model/Cache.pm
@@ -21,8 +21,7 @@ sub _perform_integrity_check {
     return shift->sqlite->db->query('pragma integrity_check')->arrays->flatten->to_array;
 }
 
-sub _check_database_integrity {
-    my ($self) = @_;
+sub _check_database_integrity ($self) {
     my $integrity_errors = $self->_perform_integrity_check;
     my $log = $self->log;
     if (scalar @$integrity_errors == 1 && ($integrity_errors->[0] // '') eq 'ok') {
@@ -34,15 +33,12 @@ sub _check_database_integrity {
     return $integrity_errors;
 }
 
-sub _kill_db_accessing_processes {
-    my ($self, @db_files) = @_;    # uncoverable statement
+sub _kill_db_accessing_processes ($self, @db_files) {
     qx{fuser -k @db_files; rm -f @db_files};    # uncoverable statement
     die 'Killing DB accessing processes failed when trying to cleanup' unless $? == 0;    # uncoverable statement
 }
 
-sub repair_database {
-    my ($self, $db_file) = @_;
-    $db_file //= $self->_locate_db_file;
+sub repair_database ($self, $db_file = $self->_locate_db_file) {
     return undef unless -e $db_file;
 
     # perform integrity check and test migration; try to provoke an error
@@ -85,9 +81,7 @@ sub init {
 
 sub _realpath { path(shift->location)->realpath }
 
-sub _locate_db_file {
-    my ($self) = @_;
-
+sub _locate_db_file ($self) {
     my $location = $self->_realpath;
     return ($location->child('cache.sqlite'), $location);
 }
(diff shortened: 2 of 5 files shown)


lib/OpenQA/CacheService/Plugin/Helpers.pm
```diff
@@ -5,27 +5,19 @@ package OpenQA::CacheService::Plugin::Helpers;
...
     $app->helper('progress.downloading_job' => \&_progress_downloading_job);
     $app->helper('progress.is_downloading' => \&_progress_is_downloading);
     $app->helper('progress.guard' => \&_progress_guard);
 }
 
-sub _progress_downloading_job {
-    my ($c, $lock) = @_;
-    return $c->downloads->find($lock);
-}
+sub _progress_downloading_job ($c, $lock) { return $c->downloads->find($lock) }
```
perlpunk, line 16: The `return` is not needed, perl will always return the result of the last statement.
Same below

lib/OpenQA/CacheService/Command/run.pm
```diff
@@ -2,15 +2,14 @@
...
+use Mojo::Base 'Minion::Command::minion::worker', -signatures;
 
 use Mojo::Util 'getopt';
 
 has description => 'Start Minion worker';
 has usage => sub { shift->extract_usage };
 
-sub run {
-    my ($self, @args) = @_;
+sub run ($self, @args) {
```
okurz, line 13: Please remove the now blank line in between

lib/OpenQA/CacheService/Model/Cache.pm
```diff
@@ -34,14 +33,12 @@ sub _check_database_integrity {
...
-    my ($self, @db_files) = @_;    # uncoverable statement
+sub _kill_db_accessing_processes ($self, @db_files) {
     qx{fuser -k @db_files; rm -f @db_files};    # uncoverable statement
     die 'Killing DB accessing processes failed when trying to cleanup' unless $? == 0;    # uncoverable statement
 }
 
-sub repair_database {
-    my ($self, $db_file) = @_;
+sub repair_database ($self, $db_file) {
     $db_file //= $self->_locate_db_file;
```
okurz, line 42: You should be able to put this as default value into the line above with

```perl
sub repair_database ($self, $db_file = $self->_locate_db_file) {
```

lib/OpenQA/CacheService/Model/Cache.pm
```diff
@@ -85,8 +82,7 @@ sub init {
 
 sub _realpath { path(shift->location)->realpath }
 
-sub _locate_db_file {
-    my ($self) = @_;
+sub _locate_db_file ($self) {
```
okurz, line 86: Delete blank line

lib/OpenQA/CacheService/Model/Cache.pm
```diff
@@ -105,8 +101,7 @@ sub refresh {
     return $self;
 }
 
-sub get_asset {
-    my ($self, $host, $job, $type, $asset) = @_;
+sub get_asset ($self, $host, $job, $type, $asset) {
```
okurz, line 105: Here as well

lib/OpenQA/CacheService/Model/Cache.pm
```diff
@@ -159,14 +154,12 @@ sub get_asset {
...
-sub asset {
-    my ($self, $asset) = @_;
+sub asset ($self, $asset) {
     my $results = $self->sqlite->db->select('assets', [qw(etag size last_use pending)], {filename => $asset})->hashes;
     return $results->first || {};
 }
 
-sub track_asset {
-    my ($self, $asset) = @_;
+sub track_asset ($self, $asset) {
```
okurz, line 163: Here as well

lib/OpenQA/CacheService/Model/Cache.pm
```diff
@@ -179,8 +172,7 @@ sub track_asset {
     if (my $err = $@) { $self->log->error("Tracking asset failed: $err") }
 }
 
-sub _update_asset_last_use {
-    my ($self, $asset) = @_;
+sub _update_asset_last_use ($self, $asset) {
```
okurz, line 176: And so on ... :)

lib/OpenQA/CacheService/Model/Cache.pm
```diff
@@ -274,8 +263,7 @@ sub _exceeds_limit ($self, $needed) {
     return 0;
 }
 
-sub _check_limits {
-    my ($self, $needed, $to_preserve) = @_;
+sub _check_limits ($self, $needed, $to_preserve) {
```
okurz, line 266:
```suggestion
sub _check_limits ($self, $needed, $to_preserve = undef) {
```

Because it you look for the use of this variable you will see that its existence and truthiness is checked

okurz (changes requested):

lib/OpenQA/CacheService/Model/Cache.pm
FruitFly638: $to_preserve = undef means that it has a default value of undef, right?

okurz: Yes

Martchus (commented):
Seems like `_cache_sync` was left out.

FruitFly638:
Helpers.pm didn't compile because I forgot the -signatures there, lets see if the tests go through now

FruitFly638:
the `register` method in Helpers.pm is getting passed three arguments instead of two and it's failing because of that (I think?) so what do I do about this

okurz:
> the `register` method in Helpers.pm is getting passed three arguments instead of two and it's failing because of that (I think?) so what do I do about this

The most simple solution is to not change it at all and leave it for someone else to finish up. We don't have a hard requirement that all methods within a file need to use signatures (for now).

perlpunk:
> The most simple solution is to not change it at all and leave it for someone else to finish up. We don't have a hard requirement that all methods within a file need to use signatures (for now).

Yeah, and especially with methods that are implementing plugins, or callbacks that get called from another (third party) module, it can be the best to leave off signatures (or always add a `@` at the end), because the code of that module might pass more arguments in the future.

perlpunk (approved):

okurz (approved):

okurz:
All except patch coverage is meeting the goal. As signatures basically introduce static code checks we can merge as is.

PR #4477, opened 2022-01-25
Fix wrong check for number of hash keys

Martchus:


lib/OpenQA/Schema/ResultSet/Jobs.pm
```diff
@@ -208,7 +208,7 @@ sub create_from_settings {
     $job->register_assets_from_settings;
 
     log_warning('Ignoring invalid group ' . encode_json(\%group_args) . ' when creating new job ' . $job->id)
-      if %group_args && !$group;
+      if keys %group_args && !$group;
```
okurz, line 211: I wonder what's the effect of that. I we never hit the warning why not just remove the complete statement? A warning in log is likely not something we would ever look at realistically anyway.

perlpunk: Since perl 5.26 `%hash` in scalar context returns the number of keys:
https://metacpan.org/release/XSAWYERX/perl-5.26.0/changes#scalar(%25hash)-return-signature-changed
So I would say no change necessary.

Martchus: Ah, so this isn't a problem anymore.

Martchus: Note that we *always* hit this warning when creating a job from settings but the group parameters are invalid. This error should most likely be propagated in a different way, though.

PR #4478, opened 2022-01-25
Allow changing job settings via restart API

Martchus:
So `openqa-investigate` can use it to handle jobs with parallel
and directly chained dependencies, see
https://progress.opensuse.org/issues/95783.

### DIFF
--- a/lib/OpenQA/Schema/Result/Jobs.pm
+++ b/lib/OpenQA/Schema/Result/Jobs.pm
@@ -593,6 +593,20 @@ sub is_ok {
     return 0;
 }
 
+sub extract_group_args_from_settings ($settings) {
+    my ($group, %group_args);
+    if (exists $settings->{_GROUP_ID}) {
+        if (my $id = delete $settings->{_GROUP_ID}) { $group_args{id} = $id }
+        else { $group = 0 }
+    }
+    if (exists $settings->{_GROUP}) {
+        if (my $name = delete $settings->{_GROUP}) { $group_args{name} //= $name }
+        else { $group = 0 }
+    }
+    $group = OpenQA::Schema->singleton->resultset('JobGroups')->find(\%group_args) if keys %group_args;
+    return (\%group_args, $group);
+}
+
 =head2 _create_clone
 
 =over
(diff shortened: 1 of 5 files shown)


lib/OpenQA/Schema/Result/Jobs.pm
```diff
@@ -616,24 +629,24 @@ sub _create_clone {
       && $cluster_job_info->{ok};
 
     # Duplicate settings (with exceptions)
+    my %spec_settings = %$settings;
+    my %main_settings
+      = map { (delete $spec_settings{$_}) // $self->$_ } qw(TEST VERSION ARCH FLAVOR MACHINE BUILD DISTRI);
```
okurz, line 634: We already have a list of such parameters in SCENARIO_KEYS and SCENARIO_WITH_MACHINE_KEYS in lib/OpenQA/Schema/Result/Jobs.pm . I suggest you define another constant in there for what you want to use here. I don't have a good name for that ready though

Martchus: I've been thinking the same and there's also a very similar list in `create_from_settings`. Maybe I can at least de-duplicate the common denominator.

lib/OpenQA/Schema/Result/Jobs.pm
```diff
@@ -593,6 +593,21 @@ sub is_ok {
     return 0;
 }
 
+sub extract_group_args_from_settings ($settings) {
+
```
okurz, line 597: remove blank line?

lib/OpenQA/Schema/Result/Jobs.pm
```diff
@@ -43,6 +43,9 @@ use OpenQA::WebSockets::Client;
 use constant SCENARIO_KEYS => (qw(DISTRI VERSION FLAVOR ARCH TEST));
 use constant SCENARIO_WITH_MACHINE_KEYS => (SCENARIO_KEYS, 'MACHINE');
 
+# job settings which are defined directly as column of the jobs table
+use constant MAIN_SETTINGS => (qw(DISTRI VERSION FLAVOR ARCH TEST MACHINE BUILD));
```
kraih, line 47: I think the outer `()` does nothing here.

okurz (approved):
I removed the notice about "still a draft" because obviously it's not anymore :)

kraih (approved):

PR #4481, opened 2022-01-26
Add more perl signatures

FruitFly638:
More perl signatures, lib/OpenQA/CacheService/Task/Asset.pm and Sync.pm


lib/OpenQA/CacheService/Task/Asset.pm
```diff
@@ -2,17 +2,15 @@
 # SPDX-License-Identifier: GPL-2.0-or-later
 
 package OpenQA::CacheService::Task::Asset;
-use Mojo::Base 'Mojolicious::Plugin';
+use Mojo::Base 'Mojolicious::Plugin', -signatures;
 
 sub register {
     my ($self, $app) = @_;
```
FruitFly638, line 8: This method gets passed 3 values, which is why I didn't change it

FruitFly638:
The `register` method in Asset.pm gets passed 3 values, which is why I didn't change it.
I assume the same is true for `register` in Sync.pm I just didn't get any specific error messages, cache_service was just not available.

lib/OpenQA/CacheService/Task/Asset.pm
kraih, line 8: You skipped this one.

lib/OpenQA/CacheService/Task/Sync.pm
```diff
@@ -15,9 +15,7 @@ sub register {
     $app->minion->add_task(cache_tests => \&_cache_tests);
```
kraih, line 15: Skipped the `register` here too.

kraih:
Yes, the signature for `register` should be `($self, $app, $config)`.

lib/OpenQA/CacheService/Task/Asset.pm
FruitFly638: See the comment I made above, I handled this similar to a problem I had in https://github.com/os-autoinst/openQA/pull/4476 in lib/OpenQA/CacheService/Plugin/Helpers.pm

The register method gets passed three values so I either skip it or add a @ at the end of the signature.

FruitFly638:
> Yes, the signature for `register` should be `($self, $app, $config)`.

Ah okay. Is there a reason in Perl to pass unused values? Can they be used somewhere down the line?

lib/OpenQA/CacheService/Task/Asset.pm
kalikiana: The third argument should be `$conf`, no? According to [the docs](https://docs.mojolicious.org/Mojolicious/Plugin#register)?

perlpunk: As I suggested in another PR, for subroutines like this (implementing a third-party API, or a third-party callback), it might be wise to use
`($self, $app, @)`
in this case, because we cannot guarantee that the third-party module will never add arguments.
Opinions?

kraih: It's very unlikely Mojolicious will ever add another argument there. But if you want to be safe `($self, $app, @)` is not a bad choice.

kraih:
In this case it's a standard plugin API (defined upstream by Mojolicious) that always passes a config hash, which will be empty here. Always passing it, even if it is empty, makes it more consistent and easier to write code for (less special cases... and so on).

lib/OpenQA/CacheService/Task/Asset.pm
FruitFly638: I added a $conf parameter for the two subroutines in question and also went back to do the same for the Helpers.pm from before.
I don't like the approach of adding an `·@` because it seems to go against the [reason for changing to signatures](https://progress.opensuse.org/issues/99579)

kalikiana (approved):

kraih (approved):

lib/OpenQA/CacheService/Task/Asset.pm
okurz: It's still adding value for the first parameters that are checked for their type, at least somewhat

PR #4485, opened 2022-01-27
Fix includes of script/ in tidyall config

okurz:
Fix includes of script/ in tidyall config

Within the script/ dir we have Perl scripts and other scripts. Just
asking tidyall to include all in script/ would mean that tidyall tries
to apply a Perl style to shell scripts which will obviously fail. So I
am reincluding parts of the rule we had before in tools/tidy to
explicitly select Perl scripts and Perl modules and tests as file
parameters to tidyall. We still need to include the script directory in
.tidyallrc so that it's not excluded even if tidyall is called with an
explicit file list as arguments.

Related progress issue: https://progress.opensuse.org/issues/99678

### DIFF
--- a/.tidyallrc
+++ b/.tidyallrc
@@ -1,3 +1,6 @@
 [PerlTidy]
-select = **/*.{pl,pm,t}
+; Notice that this does include also non-Perl files in script/ so it is
+; suggested to use tools/tidy to cover all appropriate files. "tidyall -a"
+; will try to cover too much and fail on the non-Perl files
+select = **/*.{pl,pm,t} script/*
 argv = --profile=$ROOT/.perltidyrc
--- a/external/os-autoinst-common/.gitrepo
+++ b/external/os-autoinst-common/.gitrepo
@@ -6,7 +6,7 @@
 [subrepo]
 	remote = git@github.com:os-autoinst/os-autoinst-common.git
 	branch = master
-	commit = a6cbaa31f26302cf30431ec3bc810b8abf94a601
+	commit = d6bc4d5024296520344bbb80add1a345f275ac1a
 	parent = 94bf360336469dd3f6c603860f7a629568adf6b6
 	method = merge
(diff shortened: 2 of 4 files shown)


Martchus (approved):

okurz:
Hmpf, right. Does not work because not all scripts in scripts/ are perl. If no one has a better idea I would likely resort back to processing the file list within tools/tidy and pass the list as parameters to tools/tidy.

okurz:
Ok, this should fix it. I now included a properl tidyall config file as well as an explicitly computed file list as argument in tools/tidy to pass all Perl modules and tests as well as Perl scripts (but not other scripts) to tidyall.

.tidyallrc
```diff
@@ -1,3 +1,6 @@
 [PerlTidy]
-select = **/*.{pl,pm,t}
+; Notice that this does only include Perl modules and tests, not scripts in
+; the scripts/ directory where we also have non-Perl files. Use tools/tidy to
+; cover all files
+select = **/*.{pl,pm,t} script/*
```
Martchus, line 5: The comment says "not scripts in the scripts/ directory" but the code looks pretty much like it would include those scripts.

okurz: ah, right. The comment was outdated. I fixed that. I only updated the comment, no change in functionality

Martchus: So despite `script/*` staying in the code it is still "suggested to use tools/tidy to cover all appropriate files"?

okurz: Yes, unfortunately we need to still mention `script/*` here. I explained that in https://github.com/os-autoinst/openQA/pull/4485/commits/64be93648acd98f190a3c489fdd6b065bc01962c : We still need to include the script directory in
.tidyallrc so that it's not excluded even if tidyall is called with an explicit file list as arguments.

mimi1vx (approved):

PR #4486, opened 2022-01-28
Add more perl signatures in lib:OpenQA

FruitFly638:
Add perl signatures to lib/openQA/ Command.pm, Events.pm and Files.pm

### DIFF
--- a/lib/OpenQA/Command.pm
+++ b/lib/OpenQA/Command.pm
@@ -2,7 +2,7 @@
 # SPDX-License-Identifier: GPL-2.0-or-later
 
 package OpenQA::Command;
-use Mojo::Base 'Mojolicious::Command';
+use Mojo::Base 'Mojolicious::Command', -signatures;
 
 use Cpanel::JSON::XS ();
 use OpenQA::Client;
@@ -20,8 +20,7 @@ has apibase => '/api/v1';
 has [qw(apikey apisecret host)];
 has host => 'http://localhost';
 
-sub client {
-    my ($self, $url) = @_;
+sub client ($self, $url) {
     return OpenQA::Client->new(apikey => $self->apikey, apisecret => $self->apisecret, api => $url->host)
       ->ioloop(Mojo::IOLoop->singleton);
 }
@@ -31,14 +30,11 @@ sub data_from_stdin {
     return !-t STDIN && select($r, undef, undef, 0) ? join '', <STDIN> : '';
 }
 
-sub decode_args {
-    my ($self, @args) = @_;
+sub decode_args ($self, @args) {
     return map { decode 'UTF-8', $_ } @args;
 }
 
-sub handle_result {
-    my ($self, $tx, $options) = @_;
-
+sub handle_result ($self, $tx, $options) {
     my $res = $tx->res;
     my $is_json = ($res->headers->content_type // '') =~ m!application/json!;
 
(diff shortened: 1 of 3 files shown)


lib/OpenQA/File.pm
```diff
@@ -79,8 +76,7 @@ sub get_piece {
     return $self->_chunk($index, $n_chunks, $chunk_size, $residual, $total_cksum);
 }
 
-sub split {
-    my ($self, $chunk_size) = @_;
+sub split ($self, $chunk_size = undef) {
```
baierjan, line 79: Shouldn't this one be `$chunk_size = 10000000`?

perlpunk: Signatures have one problem that can be easily overlooked: the default value will only be assigned if there was no value passed, e.g. if the argument list was empty her e(except `$self`).
So if someone passes undef by passing a variable, e.g.
`$file->split($some_var_that_might_be_undef)`
then the default assignment won't happen.

So passing `undef` here is the safest way.

perlpunk (approved):

lib/OpenQA/Command.pm
```diff
@@ -101,8 +92,7 @@ sub run {
     return $self->command(@args);
 }
 
-sub url_for {
-    my ($self, $path) = @_;
+sub url_for ($self, $path) {
```
perlpunk, line 95: hm, `url_for` https://docs.mojolicious.org/Mojolicious/Controller#url_for might be called with more arguments.
Let's see if tests are failing.

perlpunk: I just tested it locally and replaced
`my $url = $self->url_for($path);`
with
`my $url = $self->url_for($path, { x => 23 });`
in `lib/OpenQA/CLI/api.pm` and calling `perl script/openqa-cli api jobs/1 --o3 ` and it died.
So better add a `@` here.

perlpunk (changes requested):
See comment

lib/OpenQA/Command.pm
kraih: There's no relation to the Mojolicious `url_for` method though.

perlpunk: oh I see. then ignore my comment

perlpunk (approved):

lib/OpenQA/Command.pm
kraih: Bit of a confusing naming choice here, no idea why i did that. 🤷

lib/OpenQA/File.pm
FruitFly638: True, then I could just delete the other line

lib/OpenQA/Command.pm
FruitFly638: Yeah, my local tests didnt fail but I wanted to see if tests fail here

okurz (changes requested):
It was ok for the former PRs but now please really rename the git subject line to be unique. We use the git commit message subject lines for the package changelogs and the duplicate entries "Add more perl signatures" is already not nice, see https://build.opensuse.org/package/view_file/devel:openQA/openQA/_service:obs_scm:openQA.changes?expand=1

FruitFly638:
> It was ok for the former PRs but now please really rename the git subject line to be unique. We use the git commit message subject lines for the package changelogs and the duplicate entries "Add more perl signatures" is already not nice, see https://build.opensuse.org/package/view_file/devel:openQA/openQA/_service:obs_scm:openQA.changes?expand=1

I made some changes, is this better? including the directory makes it more unique

kalikiana (approved):

okurz (commented):
Well, better but nearly all files are in this dir. Better use OpenQA::Commands and only include this one file in one commit and make separate ones for the other file

okurz (approved):

PR #4488, opened 2022-01-28
Show creation time of scheduled jobs in info panel on job details page

Martchus:


perlpunk (approved):

templates/webapi/test/infopanel.html.ep
```diff
@@ -58,11 +58,15 @@
...
                         )
-                    % } elsif ($job->t_started)
-                    % {
+                    % }
+                    % elsif ($job->t_started) {
                         <%= ', started' =%>
                         <abbr class="timeago" title="<%= $job->t_started->datetime() %>Z"><%= format_time($job->t_started) %></abbr>
                     % }
+                    % else {
+                        <%= ', created' =%>
```
okurz, line 67: So it would say "scheduled, created ...". What other combinations would we see, " cancelled, created ..."? I Wonder if it's clearer leaving out the word "created"

okurz (changes requested):

templates/webapi/test/infopanel.html.ep
Martchus: "Scheduled" refers to the whole time the is job being planned for execution here (as it is the job state) so I thought an additional "created" makes sense. We should not get "cancelled, created …" because even if the job has never run it'll be "cancelled, finished …". I suppose we'd also see "assigned, created …" because those jobs haven't a start time yet. That would be ok (and e.g. "assigned, … ago" would be confusing because that would sound like the time refers to the assignment).

okurz: ok, cool

okurz (approved):

PR #4491, opened 2022-02-01
ci: Conditionally install kvm in container

kalikiana:
Notably on aarch64 this is package not available.

See: https://progress.opensuse.org/issues/105432

### DIFF
--- a/container/ci/Dockerfile
+++ b/container/ci/Dockerfile
@@ -32,7 +32,6 @@ RUN zypper in -y -C \
        postgresql-devel \
        qemu \
        qemu-tools \
-       qemu-kvm \
        tar \
        optipng \
        python3-base \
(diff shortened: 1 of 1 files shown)


Martchus (approved):
The same approach works for `openQA_container_worker` which is resolvable so I suppose this will work.

Martchus:
I'm still wondering whether OBS' dependency resolution can actually tell the difference depending on the architecture. Maybe that syntax makes OBS ignore¹ the dependency in *all* architectures? I suppose even if the answer is yes we don't care.

¹Not during the build but when computing dependencies to determine whether the package is unresolvable *before* the build.

perlpunk (approved):

kalikiana:
> Maybe that syntax makes OBS ignore¹ the dependency in _all_ architectures? I suppose even if the answer is yes we don't care.

In case this isn't clear: The [:](https://www.gnu.org/savannah-checkouts/gnu/bash/manual/bash.html#Bourne-Shell-Builtins) is a posix-compliant shell built-in. It's not specific to OBS in that sense.

Martchus:
But I doubt that OBS will determine the dependencies by actually executing the shell script. It will likely have to rely on its own parser. (When I remember correctly that has also already broken in the past by using too fancy shell syntax.)

PR #4494, opened 2022-02-02
WIP - ignore (test obs checks)

perlpunk:


perlpunk:
ok, checks are running in https://build.opensuse.org/project/show/devel:openQA:GitHub:os-autoinst:openQA:PR-4494

perlpunk:
Hm, successful builds are reported, pending aren't (as expected currently), but unresolvable aren't reported either.
Added this to https://github.com/openSUSE/open-build-service/issues/12121

okurz:
so it's tricky to see but we are guarded by mergify here that would wait indefinitely for the Leap build to complete. The unresolvables would need the project config entries from https://build.opensuse.org/projects/devel:openQA/prjconf to resolve the ambiguous choices for ruby. I am not sure if the project config should be inherited for the PR OBS projects or if we can find a better solution in devel:openQA so that we would not need to have this project config. I suggest to crosscheck with the experts in irc://libera.chat/opensuse-factory

PR #4495, opened 2022-02-02
Add new OBS checks to mergify config

perlpunk:
Issue: https://progress.opensuse.org/issues/102464

### DIFF
--- a/.mergify.yml
+++ b/.mergify.yml
@@ -8,9 +8,12 @@ pull_request_rules:
         # safe side even in case of checks not reporting back at all
         - "status-success=codecov/project"
         - "status-success=codecov/patch"
-        # wait for OBS package build as well which is triggered independantly
-        # See https://progress.opensuse.org/issues/55346 for details
-        - "status-success=OBS Package Build"
+        # Multibuild results will not report the "pending" status:
+        # https://openbuildservice.org/help/manuals/obs-user-guide/cha.obs.scm_ci_workflow_integration.html#sec.obs.obs_scm_ci_workflow_integration.setup.status_reporting
+        # "unresolvable" is not reported in general:
+        # https://trello.com/c/0N3jHq5M/2257-report-back-to-scm-when-build-results-arent-failed-and-succeeded
+        # So we need to require the number of tests explicitly:
+        - "#check-success>=30"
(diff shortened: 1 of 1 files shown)


.mergify.yml
```diff
@@ -11,6 +11,21 @@ pull_request_rules:
...
+        - "check-success=OBS: openQA - openSUSE_Tumbleweed/x86_64"
+        - "check-success=OBS: openQA:openQA-client-test - openSUSE_Tumbleweed/x86_64"
+        - "check-success=OBS: openQA:openQA-devel-test - openSUSE_Tumbleweed/x86_64"
+        - "check-success=OBS: openQA:openQA-test - openSUSE_Tumbleweed/x86_64"
+        - "check-success=OBS: openQA:openQA-worker-test - openSUSE_Tumbleweed/x86_64"
+        - "check-success~=OBS: openQA - openSUSE_Leap_.*/x86_64"
+        - "check-success~=OBS: openQA:openQA-client-test - openSUSE_Leap_.*/x86_64"
+        - "check-success~=OBS: openQA:openQA-devel-test - openSUSE_Leap_.*/x86_64"
+        - "check-success~=OBS: openQA:openQA-test - openSUSE_Leap_.*/x86_64"
+        - "check-success~=OBS: openQA:openQA-worker-test - openSUSE_Leap_.*/x86_64"
```
okurz, line 28: As an alternative you could leave out some of them and count, e.g. similar to how I do it in https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/13455/files#diff-6a30e47dd51449847422017b16878889b848f6602ac56d5a2e67ce8a19afade2R16-R17 although because jobs might not show up as pending nor as running you might need something like

```
- "#check-success>=20"
```

perlpunk: done. Requiring 29 checks now which is the number of displayed checks in this PR minus the old OBS check.

perlpunk:
The `OBS SCM/CI Workflow Integration started` check is new, so I might have to increase to 30

okurz (approved):

okurz:
I wonder where codecov has gone. I suggest we merge and observe if mergify works a expected in an independent PR

PR #4498, opened 2022-02-03
Restart failed chained parents "up the chain"

Martchus:
See https://progress.opensuse.org/issues/95783#note-16 for a detailed explanation. I'll update the commit message later. The question within the documentation change is of course also not supposed to stay and I needed to add/adapt tests.


okurz (changes requested):
I find the impact of the change hard to judge. I have read your git commit message. Could you extend it a bit further to explain the starting situation, the problems in the current state and then the impact of this change with both the intended goal as well as any potential further consequences.

Martchus:
Starting situation: You want to restart failing jobs within a dependency tree that has two failing chained root jobs that need both be restarted. That's something the investigate script does. It restarts the two failing chained root jobs individually making two additional dependency trees instead of just one.

Problem: This is almost certainly *not* what's wanted. Having all the child jobs in the dependency tree cloned twice but each time only with one of the parents is pretty useless. Just look at https://openqa.suse.de/tests/8094592#dependencies to see the mess.

Impact: When restarting one of the failed chained root jobs the 2nd one would restarted as well. However, it also affects restarting chained children which would now also restart chained parents. I suppose that's ok because it would only happen if such parents haven't passed or softfailed. Additionally, if one really only wants to restart the child job and not its failed chained parent one can still use `skip_parents=1` (or now the corresponding UI option).

I'm adding the "not-ready" label as it makes likely sense to test it out on a real world dependency tree. I can also share some screenshots then.

okurz:
The mess that Martchus means :)

Screenshot_20220207_132913_messy_openQA_dependency_tree

Martchus:
This is the most basic setup which allows one to reproduce the issue:
screenshot_20220207_163901

Let's say the following happens:
screenshot_20220207_164439

Now you'll have to decide whether to restart `root1` or `root2` first. *With* this change you'll end up exactly at the first screenshot again regardless which job you choose. The full dependency tree is restarted in any case. I would assume that's the behavior most users want/expect. Note that the other root job is only restarted here because it failed as well. If it was ok it wouldn't be restarted (and the behavior remains exactly like before this change). Restarting the child would of course have the same effect. To only restart the child ignoring failed parents one can still use `skip_parents=1`.

*Without* my change the following happens:
screenshot_20220207_165026
So only `root1` and `child1` have been restarted. Restarting `root2` manually actually works but one ends up with another instance of `child1` and is generally a race condition. (In the screenshot from @okurz 's previous comment one can see all these unnecessarily often restarted jobs in the dependency tree at the same time because those jobs are restarted by the investigate script which uses `clone=0` to prevent the "clone relation" which in turn breaks limiting the dependency tree to show only the latest jobs.)

okurz (approved):

kalikiana (approved):
The `not-ready` label made me hesitate yesterday, but I like the test changes. Also the simplificationa that make them a bit more readable.

AdamWill:
This PR actually does more than it claims: it does not only restart "failed" parents, it also restarts scheduled or running ones. This causes a problem for one specific test cluster in Fedora. I've filed https://github.com/os-autoinst/openQA/pull/4962 as a follow-up to make this really only restart *failed* parents, which would avoid our problem case.

PR #4499, opened 2022-02-09
Improve logging errors in case the cache service test fails

Martchus:
So far `bail_with_log 8, 'Job 8 produced the wrong results'` is printed in
case the cache service sub-test fails - regardless whether the failing
checks are actually about job 8. This change introduces sub-tests for
checking the different test results and prints the specific job logs.

Related ticket: https://progress.opensuse.org/issues/98952


okurz (changes requested):
<html><body>


t/full-stack.t | 96.96% <100.00%> (-0.03%)
-- | --



</body>
</html>

says that we don't have 100% anymore.

Martchus:
Is this a test? We haven't had 100 % before, only 96.99 % (see https://app.codecov.io/gh/os-autoinst/openQA/blob/master/t/full-stack.t).

By the way, these figures say we've lost 0.03 % leaving us at 96.96 %. If I understand it correctly, the 100 % mean that 100 % of the *diff* (not the whole file) have been covered. I'm not sure how lines marked as uncovered influence the calculation. Maybe these lines are responsible for the overall drop (but unfortunately I don't see a way to avoid them).

okurz:
ok. But can you just ensure that 100% of our statements in t/full-stack.t are covered? That would be the easiest for checking :)

Martchus:
I'm not sure. The uncovered code is the entire function `logfile` which is executed unconditionally in every successful testrun. I could mark it as uncoverable but it looks more like a bug of the coverage tracking.

okurz:
> before an uncoverable function with signatures is causing this. Maybe there's a better way but I don't know.

how about just adding a "test" for that very function to ensure it's called?

Martchus:
I actually did just that yesterday and it worked. However, there are still some lines left which I can likely just mark as uncoverable.

Martchus:
It seems impossible that the remaining lines are not covered. But there are already lines marked as uncoverable via `# uncoverable statement count:2` after that code. So maybe the same problem/bug as before.

Martchus:
There are only three lines left:
screenshot_20220210_135751

It must still be a bug because there's actually e.g. `ok 17 - at least one free worker with class qemu_i386 registered` in the test logs.

I would leave it at 98.95 % for the sake of this PR. The changes here help me to go on with https://progress.opensuse.org/issues/98952 regardless of the imperfect coverage (and the coverage was at least improved).

okurz (changes requested):

t/full-stack.t
```diff
@@ -106,7 +106,7 @@ sub check_scheduled_job_and_wait_for_free_worker ($worker_class) {
         last;
     }
     Test::More::ok $has_relevant_job, "job with worker class $worker_class scheduled"
-      or diag explain 'scheduled jobs: ', \@scheduled_jobs;
+      or Test::More::diag explain 'scheduled jobs: ', \@scheduled_jobs;
```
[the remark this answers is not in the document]
okurz: ok, fine

okurz (approved):
All looks good now. We don't unfortunately have 100% statement coverage for t/full-stack.t but you managed to increase the coverage which is good in any case.

t/full-stack.t
```diff
@@ -103,31 +105,28 @@ sub check_scheduled_job_and_wait_for_free_worker ($worker_class) {
...
-      or diag explain 'scheduled jobs: ', \@scheduled_jobs;
+    Test::More::ok $has_relevant_job, "job with worker class $worker_class scheduled"
+      or Test::More::diag explain 'scheduled jobs: ', \@scheduled_jobs;
 
     # wait until there's not only a free worker but also one with matching worker properties
     # note: Populating the database is not done atomically so a worker might already show up but relevant
     #       properties (most importantly WEBSOCKET_API_VERSION and WORKER_CLASS) have not been populated yet.
     my ($elapsed, $free_workers) = (0, []);
-    for (; $elapsed <= $setup_timeout; $elapsed += sleep 0.1) {
+    for (; $elapsed <= $setup_timeout; $elapsed += sleep $setup_poll_interval) {
```
kalikiana, line 115: Hrmmm why is this `not covered by tests`? That seems odd. We should have full coverage here? Maybe the variable is the problem?

assets/javascripts/ws_console.js
```diff
@@ -83,6 +83,21 @@ function establishWebSocketConnection() {
...
 }
 
+function submitWebSocketCommand(event) {
+  if (event) {
+    event.preventDefault();
+  }
+  var msg = document.getElementById('msg');
+  var command = msg.value;
+  if (!window.ws || (window.wsUsingProxy && !window.wsProxyConnectionConcluded)) {
+    logStatus("Can't send command, no ws connection opened! Will try to send when connection has been restored.");
```
kalikiana, line 93: Might be good to have the msg in there in case this goes wrong and one wonders which one it was?

```suggestion
    logStatus("Can't send ' + command + ', no ws connection opened! Will try to send when connection has been restored.");
```

Martchus: I only moved this code. Besides, the concrete message doesn't matter at all for this error condition (only the overall connection status).

kalikiana: Fair enough 👍🏾

kalikiana (approved):

t/full-stack.t
perlpunk: I added two diag messages.
This is what I got on CircleCI:
fullstack-coverage-circleci
Test output: https://app.circleci.com/pipelines/github/perlpunk/openQA/258/workflows/06822a3c-9ffb-422f-ab28-9bd425e78080/jobs/1460?invite=true#step-110-128
```
# ================ 0 18 0.1
    # ==== 0
    # ==== 0.100072
    # ==== 0.200194
    # ==== 0.300301
    # ==== 0.400373
    # ==== 0.500481
    # ==== 0.600602
    # ==== 0.700674
```
So the diag lines were both executed.

and this is local:
fullstack-coverage-local
with the only difference of 6 instead of 18:
```
# ================ 0 6 0.1                                                                                                  
# ==== 0                                                                                                                    
# ==== 0.100092                                                                                                             
# ==== 0.20019                                                                                                              
# ==== 0.300283                                                                                                             
# ==== 0.400376                                            
# ==== 0.500496                                            
```

I can't imagine what goes wrong with Devel::Cover here.

PR #4501, opened 2022-02-10
Rename script/{load|dump}_templates to openqa-*-templates

phil-hands:
This renames the scripts in the source in line with the installed names in `/usr/bin`

It also modifies the pod to use the new names.

There is also a commit that provides the old names as symlinks in the script directory, for backwards compatibility (this has not been tested by me, as I don't build RPMs)


okurz:
> There is also a commit that provides the old names as symlinks in the script directory, for backwards compatibility (this has not been tested by me, as I don't build RPMs)

That's fine because this is even be tested automatically by CI. And it  already found a problem in https://build.opensuse.org/package/live_build_log/devel:openQA:GitHub:os-autoinst:openQA:PR-4501/openQA/openSUSE_Leap_15.3/x86_64

```
[   50s] + ln -s openqa-dump-templates /usr/share/openqa/script/dump_templates
[   50s] ln: failed to create symbolic link '/usr/share/openqa/script/dump_templates': File exists
```

phil-hands:
> That's fine because this is even be tested automatically by CI. And it already found a problem in https://build.opensuse.org/package/live_build_log/devel:openQA:GitHub:os-autoinst:openQA:PR-4501/openQA/openSUSE_Leap_15.3/x86_64
> 
> ```
> [   50s] + ln -s openqa-dump-templates /usr/share/openqa/script/dump_templates
> [   50s] ln: failed to create symbolic link '/usr/share/openqa/script/dump_templates': File exists
> ```

Oh -- any idea where that's coming from?

I could just leave the commit that adds the symlinks out, if the scripts exist from somewhere else, and then you could put them in if they're ever needed, or I could check first before calling `ln`, or I guess I could use `ln -sf`(but that seems wrong to me)

dist/rpm/openQA.spec
```diff
@@ -646,8 +648,8 @@ fi
 %dir %{_datadir}/openqa/script
 %{_datadir}/openqa/script/client
 %{_datadir}/openqa/script/clone_job.pl
-%{_datadir}/openqa/script/dump_templates
-%{_datadir}/openqa/script/load_templates
```
Martchus, line 650: The removal looks wrong considering

```
ln -s openqa-dump-templates %{_datadir}/openqa/script/dump_templates
ln -s openqa-load-templates %{_datadir}/openqa/script/load_templates
```

still creates those symlinks. However, you could likely remove *those* symlinks as well (unless we want to keep them for compatibility).

Martchus (dismissed):

okurz:
> I suppose that would be ok (as mentioned in the previous comment). I don't think we need to be too careful about backwards compatibility with the names of those scripts.

Actually considering our userbase I suggest that we don't simply remove a path that was there in before. We should still have symlinks or even better wrapper scripts that output a "DEPRECATED" message and calling the openqa-* variants. But simply removing the files from their original path I consider not so nice.

phil-hands:
> > I suppose that would be ok (as mentioned in the previous comment). I don't think we need to be too careful about backwards compatibility with the names of those scripts.
> 
> Actually considering our userbase I suggest that we don't simply remove a path that was there in before. We should still have symlinks or even better wrapper scripts that output a "DEPRECATED" message and calling the openqa-* variants. But simply removing the files from their original path I consider not so nice.

So, what next?

I don't know where the files that are getting in the way came from, so I've no idea what's needed to get rid of them.

Is this an upgrade test where the contents of the previous version are still present, or some such?

Martchus:
No, it is no update test. As far as I know the packages are built in a "pristine" environment each and every time from scratch. I also find it unexpected that the files are already there (considering the Makefile shouldn't install them anymore with your changes).

phil-hands:
Oh, I see -- one isn't supposed to be writing there at that point in the install - is that it?

I guess it needs the links to be created when the scripts are copied into place, but I don't know how to do that.

I could probably create some symlinks in the source, and then list them somewhere near [here](https://github.com/os-autoinst/openQA/pull/4501/commits/94d52c6d4414d85b2ed90af7fab376dd0689b8ad#diff-0c87909c1233afe071f612c1a8d91703af43b8965059202fb7f075be703b2580L649) but there's presumably a cleaner way to do it.

phil-hands:
> The change generally looks good. Why the OBS build still fails with
> 
> ```
> [  104s] ERROR: link target doesn't exist (neither in build root nor in installed system):
> [  104s]   /usr/share/openqa/script/save_templates -> openqa-save-templates
> ```
> 
> I don't know. But the "cache" and "commit message check" failures can likely be resolved by rebasing onto current master and squashing the commits (or rephrasing to start with a capital letter).

Rebasing seems to have made the tests mostly cheerful (except for the commit message one, which I'm failing on purpose because I don't like that commit much).

I guess the codecov test needs a new test?  Is one supposed to run the same tests against symlinks as well?

I imagine there's a proper way of specifying that you want a symlink in the installed system, without having to clutter the source with symlinks, so if anyone knows how to do that properly, that would be nice ... or alternatively, if you like the cheap hack of having a couple of symlinks in the source, I suppose I could leave it like that and squash the last two commits into one.

okurz:
> I guess the codecov test needs a new test? Is one supposed to run the same tests against symlinks as well?

Don't worry about codecov. This check fails because the scripts which you only change my renaming/linking/moving have a code coverage from our tests that is below our coverage target. But this simply means we need to manually review your changes and we already understood that you don't change them functionally :)

> I imagine there's a proper way of specifying that you want a symlink in the installed system, without having to clutter the source with symlinks, so if anyone knows how to do that properly, that would be nice ... or alternatively, if you like the cheap hack of having a couple of symlinks in the source, I suppose I could leave it like that and squash the last two commits into one.

It's possible to do "just symlinks when installing" by extending the installation rule in the Makefile. However I am ok with the symlinks in source. IMHO we should keep differences in packaging as small as possible vs. the source tree.

phil-hands:
It seems that I managed to type `save_template` when meaning `dump_template`, which didn't help much. :-/

That's fixed now.

dist/rpm/openQA.spec
phil-hands: since @okurz said that backwards compatibility is desired below, I've set this as resolved

okurz (approved):

Martchus:
I suppose you'll have to rebase as the OBS checks are failing because they're looking for old assets (not existing in the current cache file anymore).

okurz:
@Mergifyio rebase

okurz:
@Martchus they do :)

Martchus (approved):

PR #4503, opened 2022-02-11
Make parameters for carry-over configurable

Martchus:
We might want to tweak these settings. It is not clear whether the defaults
are the best for all use-cases.


okurz (approved):

baierjan (approved):

okurz (changes requested):
tests fail

okurz (approved):

PR #4505, opened 2022-02-14
doc: Use `async=1` flag in examples for posting a scheduled product (iso)

Martchus:
See https://progress.opensuse.org/issues/106762

### DIFF
--- a/docs/UsersGuide.asciidoc
+++ b/docs/UsersGuide.asciidoc
@@ -1010,10 +1010,16 @@ Example for `_DEPRIORITIZEBUILD` and `_DEPRIORITIZE_LIMIT`.
 
 [source,sh]
 --------------------------------------------------------------------------------
-openqa-cli api -X POST isos ISO=my_iso.iso DISTRI=my_distri FLAVOR=sweet \
-         ARCH=my_arch VERSION=42 BUILD=1234 \
+openqa-cli api -X POST isos async=0 ISO=my_iso.iso DISTRI=my_distri \
+         FLAVOR=sweet ARCH=my_arch VERSION=42 BUILD=1234 \
          _DEPRIORITIZEBUILD=1 _DEPRIORITIZE_LIMIT=120 \
 --------------------------------------------------------------------------------
+
+*NOTE* By default scheduling products is done synchronously within the requests,
+corresponding to the parameter `async=0`. Use `async=1` to avoid possible
+timeouts by performing the task in background.
+This is recommended on big instances but means that the results (and
+possible errors) need to be polled via `openqa-cli api isos/$scheduled_product_id`.
 
 ==== Remarks ====
 
--- a/lib/OpenQA/CLI/api.pm
+++ b/lib/OpenQA/CLI/api.pm
@@ -99,9 +99,14 @@ sub command ($self, @args) {
     openqa-cli api -X POST jobs ISO=foo.iso DISTRI=my-distri \
       FLAVOR=my-flavor VERSION=42 BUILD=42 TEST=my-test
 
-    # Trigger jobs on ISO "foo.iso"
-    openqa-cli api --o3 -X POST isos ISO=foo.iso DISTRI=my-distri \
-      FLAVOR=my-flavor ARCH=my-arch VERSION=42 BUILD=1234
+    # Trigger jobs on ISO "foo.iso" creating a "scheduled product" (see
+    # https://open.qa/docs/#_spawning_multiple_jobs_based_on_templates_isos_post
+    # for details, e.g for considering to use the `async` flag)
+    openqa-cli api --o3 -X POST isos ISO=foo.iso \
+      DISTRI=my-distri FLAVOR=my-flavor ARCH=my-arch VERSION=42 BUILD=1234
+
+    # Track scheduled product
+    openqa-cli api --o3 isos/1234
 
     # Change group id for job
     openqa-cli api --json --data '{"group_id":1}' -X PUT jobs/639172


okurz (commented):
It doesn't make sense to treat `async=0` the default in code and then only show examples which explicitly state `async=1`. For now  to not go into breaking backwards compatibility I suggest we keep examples as is and provide a strong suggestion to use `async=1` for *big* instances

Martchus:
What about an explicit `async=0` in the examples with an explanation? Then the examples are still good for using it on a small instance and in accordance with the default but we raise awareness of the problem (and that one can use `async=1` to avoid it).

lib/OpenQA/CLI/api.pm
```diff
@@ -99,9 +99,14 @@ sub command ($self, @args) {
     openqa-cli api -X POST jobs ISO=foo.iso DISTRI=my-distri \
       FLAVOR=my-flavor VERSION=42 BUILD=42 TEST=my-test
 
-    # Trigger jobs on ISO "foo.iso"
-    openqa-cli api --o3 -X POST isos ISO=foo.iso DISTRI=my-distri \
-      FLAVOR=my-flavor ARCH=my-arch VERSION=42 BUILD=1234
+    # Trigger jobs on ISO "foo.iso" creating a "scheduled product" (see
+    # https://open.qa/docs/#_spawning_multiple_jobs_based_on_templates_isos_post
+    # for details)
+    openqa-cli api --o3 -X POST isos async=0 ISO=foo.iso \
```
okurz, line 105: The comment says "scheduled product" but then you use again async=0? Did you want to use async=1 here? I guess you should say `async=1` here to show this "advanced" cases whereas on top in lines 99-100 the simple, default cases is shown

docs/UsersGuide.asciidoc
```diff
@@ -1010,11 +1010,15 @@ Example for `_DEPRIORITIZEBUILD` and `_DEPRIORITIZE_LIMIT`.
 
 [source,sh]
 --------------------------------------------------------------------------------
-openqa-cli api -X POST isos ISO=my_iso.iso DISTRI=my_distri FLAVOR=sweet \
-         ARCH=my_arch VERSION=42 BUILD=1234 \
+openqa-cli api -X POST isos async=0 ISO=my_iso.iso DISTRI=my_distri \
+         FLAVOR=sweet ARCH=my_arch VERSION=42 BUILD=1234 \
          _DEPRIORITIZEBUILD=1 _DEPRIORITIZE_LIMIT=120 \
```
okurz, line 1015: Given that we show the most simple examples already in before with - if you follow my suggestion - the default then I would be ok with explicitly repeating the default setting `async=0` as you do here

docs/GettingStarted.asciidoc
```diff
@@ -437,7 +437,7 @@ For Fedora, a sample run might be:
 [source,sh]
 --------------------------------------------------------------------------------
 # Run the first test
-openqa-cli api -X POST isos \
+openqa-cli api -X POST isos async=0 \
```
okurz, line 440: As these seem to be the first and most simple examples I would just stick with the default which for us is async=0 but not explicitly stating it. After this example there is already a reference to the users guide for "more details" so that should ok

docs/Client.asciidoc
```diff
@@ -123,11 +123,11 @@ https://en.wikipedia.org/wiki/JSON[JSON]. Both will be referenced extensively.
 The `--method` option (or `-X` for short) allows you to change the HTTP request
 method from `GET` to something else. In the openQA API you will most commonly
 encounter `POST`, `PUT` and `DELETE`. For example to start testing a new ISO
-image you would use `POST`.
+image you would use `POST`:
 
 [source,sh]
 ----
-openqa-cli api --host http://openqa.example.com -X POST isos \
+openqa-cli api --host http://openqa.example.com -X POST isos async=0 \
```
okurz, line 130: IMHO it's not the job of the client documentation to explain what the server should do or does at default. I would simply not state the async parameter. The users guide already explains (with your changes) where async should be applicable

docs/UsersGuide.asciidoc
```diff
@@ -1010,11 +1010,15 @@ Example for `_DEPRIORITIZEBUILD` and `_DEPRIORITIZE_LIMIT`.
...
 --------------------------------------------------------------------------------
 
+*NOTE* Use `async=1` to avoid possible timeouts by performing the task in background.
+This is recommended on big instances but has the disadvantage that the results (and
+possible errors) need to be polled via `openqa-cli api isos/$scheduled_product_id`.
+
```
okurz, line 1021: Please describe what the default behaviour is, corresponding to `async=0`. After that include your text. E.g.

```suggestion
*NOTE* By default scheduling products is done synchronously within the requests,
corresponding to the parameter `async=0`. Use `async=1` to avoid possible
timeouts by performing the task in background.
This is recommended on big instances but means that the results (and
possible errors) need to be polled via `openqa-cli api isos/$scheduled_product_id`.

```

okurz (changes requested):

lib/OpenQA/CLI/api.pm
Martchus: No, the "scheduled product" is created in any case. I've just mentioned the term here so it is established to be used for the next example. (You can also check the state of scheduled products created with `async=0` so there's also no conflict with that either.)

kraih:
We should probably look for a way to make `async=1` the default.

okurz:
> We should probably look for a way to make `async=1` the default.

Yes, see the ticket, not for this PR

docs/Client.asciidoc
Martchus: So far the example already states all interesting parameters explicitly and it seems consequent to have `async=0` as well. I could of course also just state one parameter, e.g. `ISO=foo.iso …`. We don't really need a full example here.

okurz: well effectively all other parameters specified in this example are actually needed, unlike `async=0` so I would favor that you simply leave this line as-is

lib/OpenQA/CLI/api.pm
Martchus: Removed the parameter.

docs/UsersGuide.asciidoc
Martchus: Removed it in the simpler examples.

okurz (approved):

perlpunk (approved):

PR #4506, opened 2022-02-14
Add support for query parameters to api command

kraih:
Triggered by a Slack discussion. There's really no reason why the path doesn't allow query parameters currently.


Martchus (approved):

mimi1vx (approved):

okurz (commented):
That's ok. It can catch mistakes. Would you say it makes sense if the help text and optionally also documentation state that query parameters also works?

PR #4507, opened 2022-02-15
Show dependency graph also for cloned/restarted jobs

Martchus:
* Do not blindly follow the clone chain to the latest job anymore
    * Display always the current job, even if it is not the latest clone
    * Avoid following the clone chain to the latest job for children if
      the child's clone is not a dependency of the parent job anymore
    * Avoid following the clone chain to the latest job for parents
* Do not display (directly) chained children of jobs when the clone depth
  does not match the one from the current job to avoid dragging too many
  jobs into the tree
* See https://progress.opensuse.org/issues/69976

### DIFF
--- a/lib/OpenQA/WebAPI/Controller/Test.pm
+++ b/lib/OpenQA/WebAPI/Controller/Test.pm
@@ -889,7 +889,7 @@ sub _add_job ($dependency_data, $job, $as_child_of, $preferred_depth) {
     for my $child ($job->children->all) {
         # add chained deps only if we're still on the preferred depth to avoid dragging too many jobs into the tree
         next
-          if ($ancestors //= $job->ancestors) != $preferred_depth
+          if ($ancestors //= $job->ancestors) > $preferred_depth
           && $child->dependency != OpenQA::JobDependencies::Constants::PARALLEL;
         _add_job($dependency_data, $child->child, $job_id, $preferred_depth);
     }
--- a/lib/OpenQA/WebAPI/Controller/Test.pm
+++ b/lib/OpenQA/WebAPI/Controller/Test.pm
@@ -16,6 +16,8 @@ use File::Basename;
 use POSIX 'strftime';
 use Mojo::JSON qw(to_json decode_json);
 
+use constant DEPENDENCY_DEBUG_INFO => $ENV{OPENQA_DEPENDENCY_DEBUG_INFO};
+
 sub referer_check {
     my ($self) = @_;
     return $self->reply->not_found if (!defined $self->param('testid'));
@@ -425,7 +427,7 @@ sub _show {
             scenario => $job->scenario_name,
             worker => $job->worker,
             assigned_worker => $job->assigned_worker,
-            show_dependencies => !defined($job->clone_id) && $job->has_dependencies,
+            show_dependencies => $job->has_dependencies,
             show_autoinst_log => $job->should_show_autoinst_log,
             show_investigation => $job->should_show_investigation,
             show_live_tab => $job->state ne DONE,
@@ -789,8 +791,7 @@ sub module_fails {
         });
 }
 
-sub _add_dependency_to_graph {
-    my ($edges, $cluster, $cluster_by_job, $parent_job_id, $child_job_id, $dependency_type) = @_;
+sub _add_dependency_to_graph ($edges, $cluster, $cluster_by_job, $parent_job_id, $child_job_id, $dependency_type) {
 
     # add edge for chained dependencies
     if (   $dependency_type eq OpenQA::JobDependencies::Constants::CHAINED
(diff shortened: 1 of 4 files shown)


lib/OpenQA/WebAPI/Controller/Test.pm
```diff
@@ -874,26 +884,31 @@ sub _add_job {
     # add parents
     for my $parent ($job->parents->all) {
         my ($parent_job, $dependency_type) = ($parent->parent, $parent->dependency);
-        my $parent_job_id = _add_job($visited, $nodes, $edges, $cluster, $cluster_by_job, $parent_job) or next;
+        my $parent_job_id
+          = _add_job($visited, $nodes, $edges, $cluster, $cluster_by_job, $parent_job, $as_child_of, $preferred_depth)
```
okurz, line 888: if that wraps over multiple lines you know you have too many parameters :) Can you refactor that?

lib/OpenQA/WebAPI/Controller/Test.pm
```diff
@@ -874,26 +884,31 @@ sub _add_job {
...
     }
 
     # add children
     for my $child ($job->children->all) {
-        _add_job($visited, $nodes, $edges, $cluster, $cluster_by_job, $child->child);
+        # add chained deps only if we're still on the preferred depth to avoid dragging too many jobs into the tree
+        next
+          if ($ancestors //= $job->ancestors) != $preferred_depth
+          && $child->dependency != OpenQA::JobDependencies::Constants::PARALLEL;
+        _add_job($visited, $nodes, $edges, $cluster, $cluster_by_job, $child->child, $job_id, $preferred_depth);
```
okurz, line 900: same effect here as in the comment above

lib/OpenQA/WebAPI/Controller/Test.pm
```diff
@@ -874,26 +884,31 @@ sub _add_job {
...
     }
 
     return $job_id;
 }
 
-sub dependencies {
-    my ($self) = @_;
+sub dependencies ($self) {
```
okurz, line 907: to reduce risk of deployment I suggest to split out introducing signatures into a separate commit and even separate PR that we can merge first

okurz (changes requested):

Martchus:
Just a few examples how it looks like. All graphs (except the last example) are real examples from our production instance.

A fully restarted dependency tree looks definitively as it should.
Original tree: grafik
Restarted tree: grafik

---

Partially restarted trees also look most of the time.
Original tree: grafik
Where only the following jobs have been restarted: screenshot_20220216_141734

---

However, partially restarted trees can also look like this: screenshot_20220216_142013

In this case restarted children of another "root" job are dragged into the graph. So far I haven't found a good solution to avoid this (without impairing other cases). However, I don't consider it *that* bad. None of the parallel clusters is only shown half and it wouldn't drag any further chained children of the wrong "restart depth" into the graph.

The graph of the latest jobs is also not affected at all (so there's no regression): screenshot_20220216_142652

---

By the way, a partial restart plus another partial restart as described in https://github.com/os-autoinst/openQA/pull/4498#issuecomment-1031638134 looks like this:

The initial cluster is rendered exactly like one would expect it: screenshot_20220216_143228

This is the graph for the first clone of root1 (after root1 has been restarted without https://github.com/os-autoinst/openQA/pull/4498 and root2 has been restarted separately in addition and then root1 has been restarted again *with* https://github.com/os-autoinst/openQA/pull/4498): screenshot_20220216_143415

It looks a bit confusing but also shows the problem I was getting at in the other PR quite nicely (that child1 has been restarted twice). So I guess having it rendered this way isn't the worst.

The graph for the latest jobs looks still simple (so again, no regression): screenshot_20220216_144127
 
---

Maybe I can still tweak it further to simplify graphs of restarted jobs. However, considering those graphs weren't displayed before at all and graphs for the latest jobs are still as before I already consider this an improvement.

Of course I still need to adapt tests but I wanted to wait for feedback first.

okurz:
The example about "A fully restarted dependency" shows that in one case the "ppc64le-hmc-4disk" is coming first, and the other time at end. Can this be influenced by ordering before the data is rendered or is it really something non-predictable by the graphics rendering?

The part about "In this case restarted children of another "root" job are dragged into the graph" is a bit worrying and I don't really understand it but I would accept it for now. Rest looks fine.

Martchus:
About the first question: This is indeed inconsistent but unrelated to this change. The order in which nodes are added to the tree depends on the current job and I selected a different job in these screenshots. I suppose we could simply sort the nodes in the end, possibly even in JavaScript.

One thing I've forgotten to mention: The code should ensure that no nodes are added to the tree which are not somehow connected to other nodes. So all jobs being dragged into the graph should at least indirectly depend on the current job. (That's what the `$clone->is_child_of($as_child_of)` check is supposed to ensure.)

t/ui/16-tests_dependencies.t
```diff
@@ -219,7 +185,19 @@ subtest 'graph rendering' => sub {
...
-    like(get_tooltip(99961), qr/.*opensuse-13.1-NET-x86_64-Build0091-kde\@64bit<\/p>/, 'tooltip for kde job 99963');
+    like(get_tooltip(99961), qr/.*opensuse-13.1-NET-x86_64-Build0091-kde\@64bit.*<\/p>/, 'tooltip for kde job 99963');
+
+    subtest 'cloned job' => sub {
+        $driver->get('/tests/99945');
+        $driver->find_element_by_link_text('Dependencies')->click();
+        wait_for_ajax();
+        $graph = $driver->find_element_by_id('dependencygraph');
+        $check_element_quandity->(
+            '.cluster', 0, 'no cluster present (as only the latest job has dependency to cluster)'
```
kalikiana, line 196:
```suggestion
            '.cluster', 0, 'no cluster present (as only the latest job depends on the cluster)'
```

kalikiana (approved):

okurz (approved):

PR #4510, opened 2022-02-16
Use signatures in OpenQA::CacheService::Controller

okurz:

PR #4511, opened 2022-02-16
Part 2: Use signatures in OpenQA::CacheService - based on #4510

okurz:
Based on
* #4510


lib/OpenQA/CacheService/Model/Cache.pm
```diff
@@ -113,18 +107,16 @@ sub get_asset ($self, $host, $job, $type, $asset) {
 
     my $start;
     my $options = {
```
Martchus, line 109: Likely one of the nested subs defined here are responsible for the test failure you've mentioned.

Martchus (approved):

baierjan (approved):

PR #4515, opened 2022-02-18
Ensure the dependency graph is sorted consistently

Martchus:
So far the order of elements within the dependency tree could be different
depending on the current job. This change makes it consistent by sorting
the edges before passing them to the rendering library. (Sorting nodes does
not help and also does not seem to be required.)

See https://github.com/os-autoinst/openQA/pull/4507#issuecomment-1041556218


Martchus:
I tested it locally using OSD data to have a tree complex enough to show the difference. In addition, the JavaScript code is covered. If the change would break existing behavior test would fail (currently the order is unspecified).

perlpunk (approved):

PR #4516, opened 2022-02-18
Remove background color and redundant margin in needle editor

kalikiana:
Fixes: https://progress.opensuse.org/issues/105157

Before:

image

After:

image

### DIFF
--- a/templates/webapi/step/edit.html.ep
+++ b/templates/webapi/step/edit.html.ep
@@ -268,7 +268,7 @@
                     % end
                 </div>
             </div>
-            <div style="margin: 0 10px; position: relative; width: 1024px; padding: 1px; border: 0px; background-color: black; margin: 10px auto">
+            <div style="margin: 10px auto; width: 1024px;">
                 <canvas tabindex="1" id="needleeditor_canvas" width="1024" height="768" style="border: 0px;">This text is displayed if your browser does not support HTML5 Canvas.</canvas>
                 % unless (is_operator) {
                     <div style="margin: 0; padding: 0; border: 0; width: 1024px; height: 768px; position: absolute; top: 0; left: 0;"></div>


Martchus:
Ok, so you're simply removing the marging and background "bolor". My brain's HTML/CSS engine says this should work :-)

okurz:
UI tests fail. And please provide screenshots before and after or provide an instance where we can see that working, maybe in cloud or something :)

kalikiana:
> UI tests fail. And please provide screenshots before and after or provide an instance where we can see that working, maybe in cloud or something :)

Even re-running it 20 times I can't see a failure locally, please don't tell me 15-admin-workers.t got unstable 😞

okurz:
The "After" picture shows a problem that had been raised in the ticket already. It seems with no margin configured there is now a black border that is on the positive side consistent but still a black border. Especially in the top part of the screenshot within the needle editor users would be troubled to distinguish between that new black border and the black of the gnome bar. So IMHO this is not yet good enough. Maybe just pick a specific pattern or color to make it explicit?

mimi1vx (approved):

okurz (changes requested):
The "After" picture shows a problem that had been raised in the ticket already. It seems with no margin configured there is now a black border that is on the positive side consistent but still a black border. Especially in the top part of the screenshot within the needle editor users would be troubled to distinguish between that new black border and the black of the gnome bar. So IMHO this is not yet good enough. Maybe just pick a specific pattern or color to make it explicit?

@Vogtinator WDYT?

Vogtinator:
Yeah, the border currently looks like it's part of the VM display.

Martchus:
That's true (and I suppose GNOME doesn't have a black border).

mimi1vx:
yes, but now fulfills one of the acceptance criteria, it looks consistent.

Martchus:
In my opinion we should not show a border here at all (unless it is part of the screenshot which obviously shouldn't be cropped). So can you clarify whether the border is just part of the screenshot in your example? If not I'd be objecting as well - although it is at least an improvement to how it currently looks.

Vogtinator:
> So can you clarify whether the border is just part of the screenshot in your example?

I don't think so. It looks differently in the before/after pictures and the needle PNG doesn't have a black border around firefox, there's no seam at all.

kalikiana:
> In my opinion we should not show a border here at all (unless it is part of the screenshot which obviously shouldn't be cropped). So can you clarify whether the border is just part of the screenshot in your example? If not I'd be objecting as well - although it is at least an improvement to how it currently looks.

For reference, the red here is spacing that comes from the canvas. It's not part of the needle, but it doesn't line up exactly.
My fix is effectively making this transparent:

image https://openqa.opensuse.org/tests/2226366/modules/kontact/steps/18/edit

okurz:
As discussed the result that looks best:

```
<div style="width: 1024px; margin: 10px auto">
```

Martchus (approved):

okurz (approved):
I would have hoped about a bit of back-story in the git commit message after we had these revelations but it's ok :)

PR #4519, opened 2022-02-21
doc: Mention to stop `…-reload-worker-…`-service in auto-restart setup

Martchus:
Not stopping these services will cause them to fail which one likely wants
to avoid, see https://progress.opensuse.org/issues/107152.

### DIFF
--- a/docs/Installing.asciidoc
+++ b/docs/Installing.asciidoc
@@ -663,8 +663,9 @@ interrupting currently running jobs:
 
 [source,sh]
 --------------------------------------------------------------------------------
-# prevent worker services from restarting
-systemctl mask openqa-worker-auto-restart@{1..28}
+# prevent worker services from restarting and being automatically reloaded
+systemctl stop openqa-reload-worker-auto-restart@{1..28}.{service,path}
+systemctl mask openqa-worker-auto-restart@{1..28}.service
 # ensure idling worker services stop now (`--kill-who=main` ensures only the
 # worker receives the signal and *not* isotovideo)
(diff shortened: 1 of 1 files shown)


perlpunk (approved):

okurz (approved):

kraih (approved):

docs/Installing.asciidoc
```diff
@@ -663,8 +663,9 @@ interrupting currently running jobs:
 
 [source,sh]
 --------------------------------------------------------------------------------
-# prevent worker services from restarting
-systemctl mask openqa-worker-auto-restart@{1..28}
+# prevent worker services from restarting and being automatically reloaded
+systemctl stop openqa-reload-worker-auto-restart@{1..28}.{service,path}
```
baierjan, line 667: Shouldn't that be `systemctl disable` or `systemctl disable --now` to make it permanent like the `systemctl mask`?

Martchus: This section isn't about making things permanent. The masking is just done because it will allow the job to continue running but prevents the restarting. Note that in our production setup this actually needed to be `systemctl mask --now` to prevent salt from restarting it (but for that I've created concrete examples in https://gitlab.suse.de/openqa/salt-states-openqa/-/merge_requests/655).

baierjan: Ah, ok. Wasn't sure about this as the "mask" will survive reboot, the "stop" does not. That's why I was talking about permanency.

PR #4520, opened 2022-02-21
Avoid logging user-input related errors in YAML to server log

Martchus:
* Distinguish between errors caused by user-input (1) and unhandled
  errors (2)
* Report only (1) to the user and log only (2) to the server log
* Treat constraint violations as user-input related as they are most likely
  caused by the input as well
* See https://progress.opensuse.org/issues/106880

### DIFF
--- a/lib/OpenQA/Schema/Result/JobGroups.pm
+++ b/lib/OpenQA/Schema/Result/JobGroups.pm
@@ -419,11 +419,12 @@ sub template_data_from_yaml {
                 foreach my $machine_name (@{$machine_names}) {
                     my $job_template_key
                       = $arch . $product_name . $machine_name . ($testsuite_name // '') . ($job_template_name // '');
-                    die "Job template name '"
-                      . ($job_template_name // $testsuite_name)
-                      . "' is defined more than once. "
-                      . "Use a unique name and specify 'testsuite' to re-use test suites in multiple scenarios.\n"
-                      if $job_template_names{$job_template_key};
+                    if ($job_template_names{$job_template_key}) {
+                        my $name = $job_template_name // $testsuite_name;
+                        my $error = "Job template name '$name' is defined more than once. "
+                          . "Use a unique name and specify 'testsuite' to re-use test suites in multiple scenarios.";
+                        return {error => $error};
+                    }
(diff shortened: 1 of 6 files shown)


perlpunk:
The CI problems in the fullstack / unstable tests are likely caused by https://progress.opensuse.org/issues/107002 , I commented there

okurz:
Typo in commit message s/unnecassary/unnecessary/

Martchus:
@perlpunk Indeed. At least now we have the video, see https://progress.opensuse.org/issues/106912.

kalikiana:
So this is failing in unstable:

```
Retry 1 of 5 …
[14:10:29] t/05-scheduler-full.t .. 3/? 
    #   Failed test 'Allocated maximum number of jobs that could have been allocated'
    #   at t/05-scheduler-full.t line 225.
    #          got: '9'
    #     expected: '10'
    # Looks like you failed 1 test of 1.
[14:10:29] t/05-scheduler-full.t .. 4/? 
```

And subsequent runs fail due to issues with retries, see https://progress.opensuse.org/issues/107002 but I think this PR doesn't have to block on that.

kalikiana (approved):

perlpunk (commented):
I think I don't see a test case where an unexpected error happens and is logged, (but we also didn't have that before). Do you think it makes sense to add one?
Other than that all good.

okurz:
@mergifyio rebase

perlpunk:
You could rebase now as the unrelated problems with the unstable tests should be fixed.

okurz:
@perlpunk I just asked mergify to do that. It's just weird that it looks like "kraih" would have rebased.

okurz (changes requested):
codecov/patch failed. As this PR is non-critical I recommend to increase patch coverage as part of this work here.

Martchus:
Ok, so I need to cover:

```
        # Push the exception to the list of errors without the trailing new line
        push @$user_errors, substr($_, 0, -1);
```

and 

```
    if (@server_errors) {
        push @$user_errors, 'Internal server error occurred';
        $self->app->log->error(@server_errors);
```

perlpunk (approved):

PR #4523, opened 2022-02-22
ci: Remove fullstack temp folder between retries

kalikiana:
Because test-unstable executes multiple tests in a loop and overrides COVERDB_SUFFIX the folder needs to be set within the target.

See: https://progress.opensuse.org/issues/107002


tools/delete-artifacts-folder
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
+# In case of a test failure and retry, we want to delete previously recorded
+# artifacts and coverage because it might cover different code paths in case of
+# success, and this leads to flaky coverage reports.
+
+set -e
+
+rm -vfr "cover_db$COVERDB_SUFFIX"
+[ -d "$OPENQA_FULLSTACK_TEMP_DIR" ] && rm -vfr "$OPENQA_FULLSTACK_TEMP_DIR"
```
perlpunk, line 9: hm, this will delete `test-results/fullstack` completely, including result directories for previously running tests.
I don't have a simple idea how to prevent that right now. Maybe add `COVERDB_SUFFIX` to the variable. (And rename `COVERDB_SUFFIX` into something more general.)
Since we're currently only interested in the fullstack results, it might be ok though.

perlpunk (changes requested):
See my comment

tools/delete-artifacts-folder
Martchus: It is still better than having nothing at all or broken retries due to the leftovers. However, one could likely simply move everything into a different directory (or a sub directory).

kalikiana: Let's make the folder $target-specific then

tools/ci/run_unit_tests.sh
```diff
@@ -19,5 +19,5 @@ sudo chown -R $USER test-results
 # circleCI can be particularly slow sometimes since some time around 2021-06
 export OPENQA_TEST_TIMEOUT_SCALE_CI=3
 export EXTRA_PROVE_ARGS="-v"
-export OPENQA_FULLSTACK_TEMP_DIR=$PWD/test-results/fullstack
+export OPENQA_FULLSTACK_TEMP_DIR=$PWD/test-results/fullstack-$target
```
perlpunk, line 22: I think the target is unique per circleci job anyway.
I think we would need the test name, as we use already in `COVERDB_SUFFIX`.
Why not reuse this?

kalikiana: @perlpunk `retry` re-runs everything. The result would be the same:

    RETRY=${RETRY} HOOK=./tools/delete-artifacts-folder timeout -s SIGINT -k 5 -v ${TIMEOUT_RETRIES} tools/retry prove ${PROVE_LIB_ARGS} ${PROVE_ARGS}

kalikiana: Oh, but indeed I overlooked the `unstable` case which is special in that it uses multiple coverdb prefixes within the target!

Makefile
```diff
@@ -197,7 +197,8 @@ test-unit-and-integration:
 	export GLOBIGNORE="$(GLOBIGNORE)";\
 	export DEVEL_COVER_DB_FORMAT=JSON;\
 	export PERL5OPT="$(COVEROPT)$(PERL5OPT) -It/lib -I$(PWD)/t/lib -MOpenQA::Test::PatchDeparse";\
-	RETRY=${RETRY} HOOK=./tools/delete-coverdb-folder timeout -s SIGINT -k 5 -v ${TIMEOUT_RETRIES} tools/retry prove ${PROVE_LIB_ARGS} ${PROVE_ARGS}
+	export OPENQA_FULLSTACK_TEMP_DIR="$PWD/test-results/fullstack$(COVERDB_SUFFIX)"
```
perlpunk, line 200:
```suggestion
	export OPENQA_FULLSTACK_TEMP_DIR="$(PWD)/test-results/fullstack$(COVERDB_SUFFIX)"
```

okurz (changes requested):
Why did you move the variables? I don't see anything against that, just curious.

One phrase the git commit message subject line to state the why rather than the what. Think from changelog perspective

kalikiana:
> Why did you move the variables? I don't see anything against that, just curious.
> 
> One phrase the git commit message subject line to state the why rather than the what. Think from changelog perspective

Quoting myself: `Because test-unstable executes multiple tests in a loop and overrides COVERDB_SUFFIX the folder needs to be set within the target.`

perlpunk:
store_artifacts doesn't support wildcards
https://discuss.circleci.com/t/wildcards-in-path-definition-for-uploading-artifacts/18992/3
(yes, it's weird because other actions in CircleCI *do* support wildcards)

So I'd suggest to create another subfolder, e.g.
OPENQA_FULLSTACK_TEMP_DIR="$(PWD)/test-results/fullstack/tmpdir_$(COVERDB_SUFFIX)" and then keep using `test-results/fullstack` for store_artifacts

kalikiana:
Closing in favor of #4523

PR #4525, opened 2022-02-23
Fix outgoing TLS connections in bootstrap-container

asdil12:
Call update-ca-certificates in bootstrap-container.
The automated execution of update-ca-certificates now seems to
depend on ca-certificates.path running which is not the case
before the container is started.

Ticket: https://progress.opensuse.org/issues/107242


asdil12:
Verified locally.

okurz (changes requested):
The commit message could have a bit more details, e.g. what issue is fixed, why this needs to change, the choices you made, etc. Also see https://commit.style/ or http://chris.beams.io/posts/git-commit/ as a helpful guide how to write good commit messages. You already referenced a ticket, that is good but please also explained in a self-contained manner in the git commit message itself. Also please answer the "Why", not the "What" in the git commit message subject line. Think from the perspective of a changelog reader.

Martchus:
I suppose @okurz was after a commit message *subject* like "Fix outgoing TLS connections in bootstrap-container".

perlpunk (approved):

okurz (approved):

PR #4526, opened 2022-02-23
Expose bandwidth data for worker cache via InfluxDB

kraih:
Progress: https://progress.opensuse.org/issues/106901

### DIFF
--- a/lib/OpenQA/CacheService/Model/Cache.pm
+++ b/lib/OpenQA/CacheService/Model/Cache.pm
@@ -8,7 +8,7 @@ use Carp 'croak';
 use Capture::Tiny 'capture_merged';
 use Mojo::URL;
 use OpenQA::Log qw(log_error);
-use OpenQA::Utils qw(base_host human_readable_size check_df download_speed);
+use OpenQA::Utils qw(base_host human_readable_size check_df download_rate download_speed);
 use OpenQA::Downloader;
 use Mojo::File 'path';
 use Time::HiRes qw(gettimeofday);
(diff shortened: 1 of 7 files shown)


t/25-cache.t
```diff
@@ -78,8 +78,8 @@ END { stop_service($server_instance) }
 
 my $app = OpenQA::CacheService->new(log => $log);
 my $cache = $app->cache;
-is $cache->sqlite->migrations->latest, 4, 'version 4 is the latest version';
-is $cache->sqlite->migrations->active, 4, 'version 4 is the active version';
+is $cache->sqlite->migrations->latest, 5, 'version 4 is the latest version';
+is $cache->sqlite->migrations->active, 5, 'version 4 is the active version';
```
okurz, line 82: Now the description strings are wrong. That's the reason why the description string should not even contain the version number. Better:

```suggestion
is $cache->sqlite->migrations->latest, 5, 'Current version is the latest version';
is $cache->sqlite->migrations->active, 5, 'Current version is the active version';
```

okurz (changes requested):

lib/OpenQA/CacheService/Model/Cache.pm
```diff
@@ -168,6 +169,21 @@ sub track_asset ($self, $asset) {
     if (my $err = $@) { $self->log->error("Tracking asset failed: $err") }
 }
 
+sub metrics ($self) {
+    my $metrics = $self->sqlite->db->query("SELECT * FROM metrics")->hashes;
+    my %metrics;
+    $metrics{$_->{name}} = $_->{value} for @$metrics;
```
Martchus, line 175: I suppose this could use `map`.

t/25-cache-service.t
```diff
@@ -485,45 +485,53 @@ subtest 'Test Minion Sync task' => sub {
 };
 
 subtest 'Minion monitoring with InfluxDB' => sub {
+    my $app = OpenQA::CacheService->new;
+    my $rate
+      = $app->cache->sqlite->db->query("SELECT value FROM metrics WHERE name = 'download_rate' LIMIT 1")->hash->{value};
```
Martchus, line 490: If this would call `sub metrics` you'd directly gain coverage for it.

Martchus (approved):

okurz (approved):

PR #4528, opened 2022-02-23
openqa-clone-job: Prevent cloning jobs accidentally multiple times

Martchus:
The `clone_job` function is called recursively and the recursive call might
already handle the job for which we make the recursive call. In this case
we can simply stop further processing the job to prevent cloning it twice.

This fixes https://progress.opensuse.org/issues/107311.

---

I'll still have to come up with a unit test. So far I only tested locally as
mentioned in https://progress.opensuse.org/issues/107311#note-9.

### DIFF
--- a/lib/OpenQA/Script/CloneJob.pm
+++ b/lib/OpenQA/Script/CloneJob.pm
@@ -239,6 +239,8 @@ sub clone_job ($jobid, $options, $clone_map = {}, $depth = 0, $parent_jobid = 0)
                 # don't clone parallel jobs yet if children job type
                 next if $dependencies == $parallel && $job_type eq 'children';
                 clone_job($_, $options, $clone_map, $depth + 1) for @$dependencies;
+                # abort here if the job has already been cloned as part of the preceding recursive clone_job call
+                return $clone_map->{$jobid} if defined $clone_map->{$jobid};
             }
 
             my @new_chained = map { $clone_map->{$_} } @$chained;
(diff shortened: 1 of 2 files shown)


okurz (approved):

lib/OpenQA/Script/CloneJob.pm
```diff
@@ -237,8 +237,10 @@ sub clone_job ($jobid, $options, $clone_map = {}, $depth = 0, $parent_jobid = 0)
             push @$child_list, @$parallel if (@$parallel && $job_type eq 'children');
             for my $dependencies ($chained, $directly_chained, $parallel) {
                 # don't clone parallel jobs yet if children job type
-                next if (@$dependencies && $dependencies == $parallel && $job_type eq 'children');
+                next if $dependencies == $parallel && $job_type eq 'children';
                 clone_job($_, $options, $clone_map, $depth + 1) for @$dependencies;
+                # abort here if the job has already been cloned as part of the preceding recursive clone_job call
+                return $clone_map->{$jobid} if defined $clone_map->{$jobid};
```
perlpunk, line 243: Hm, isn't this the same as line 225?

Martchus: Yes, but it is required here as well because at this point the check in 225 has already been done but `$clone_map` might have changed so the condition might now be true.

perlpunk: ah ok

perlpunk (approved):

PR #4529, opened 2022-02-23
templates: Clarify which "Arch" we refer to on "Workers" page

okurz:
The table on the "Workers" admin page shows a column "Arch" which is the
detected architecture on the host running a worker instance. This "Arch"
can be misleading if a worker is used to connect to another, potential
remote architecture system under test. This commit changes the wording
to "Host Architecture" to unambiguously refer to the architecture on the
host machine, not what the backend of a specific openQA worker instance
might refer to.

Related progress issue: https://progress.opensuse.org/issues/107197

### DIFF
--- a/templates/webapi/admin/workers/index.html.ep
+++ b/templates/webapi/admin/workers/index.html.ep
@@ -37,7 +37,7 @@
                     <th>Worker</th>
                     <th>Host</th>
                     <th>Class</th>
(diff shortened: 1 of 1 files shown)


Martchus (approved):

Zaoliang:
yes, this is good, thanks!

perlpunk (approved):

PR #4532, opened 2022-02-25
Fix cache service coverage for Minion tasks

kraih:
#4526 showed that coverage is not collected correctly for Minion tasks.

### DIFF
--- a/t/lib/OpenQA/Test/Utils.pm
+++ b/t/lib/OpenQA/Test/Utils.pm
@@ -93,9 +93,21 @@ sub cache_minion_worker {
             require OpenQA::CacheService;
             local $ENV{MOJO_MODE} = 'test';
             note('Starting cache minion worker');
-            OpenQA::CacheService::run(qw(run --dequeue-timeout 1));
+            OpenQA::CacheService::run(
+                ['run', '--dequeue-timeout', '1'],
+                sub ($app) {
+                    $app->minion->on(
+                        worker => sub ($minion, $worker) {
+                            $worker->on(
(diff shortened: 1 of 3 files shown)


t/lib/OpenQA/Test/Utils.pm
```diff
@@ -93,7 +93,23 @@ sub cache_minion_worker {
...
+                            $worker->on(
+                                dequeue => sub {
+                                    my ($worker, $job) = @_;
+                                    $job->on(
+                                        cleanup => sub {
+                                            Devel::Cover::report() if Devel::Cover->can('report');
+                                        });
+                                });
+                        });
+                });
```
okurz, line 112: one has to admire such prettyness :)

Would this work though

```suggestion
            OpenQA::CacheService::run(
                ['run', '--dequeue-timeout', '1'],
                sub ($app) {
                    $app->minion->on(worker => sub ($minion, $worker) {
                            $worker->on(dequeue => sub ($worker, $job) {
                                    $job->on(cleanup => sub { Devel::Cover::report() if Devel::Cover->can('report') });
                                });
                        });
                });
```

kraih: Sure, i don't want to covert the whole file to signatures though. 😉

kraih: Oh, the file is already partially converted.

kraih:
Even after a few iterations, it does not look like the change had the effect i was hoping for. Back to the drawing board i guess.

kraih:
I know the current code is wrong, but i can't prove it with tests yet.

PR #4533, opened 2022-02-25
Update SUSE links for openqa.suse.de branding

okurz:
* branding: Use non-personal daily openQA review links
* branding: Remove outdated reference to caasp staging
* branding: Use new SUSE internal chat reference
* Delete outdated references to novell or microfocus after SUSE split out some time ago

### DIFF
--- a/templates/webapi/branding/openqa.suse.de/docbox.html.ep
+++ b/templates/webapi/branding/openqa.suse.de/docbox.html.ep
@@ -3,8 +3,5 @@
 <p><a href="http://open.qa">» More information regarding openQA</a></p>
 <p><a href="https://confluence.suse.com/display/openqa/openQA">» Description of internal setup and administration</a></p>
 <p><a href="https://chat.suse.de/channel/testing">» Chat</a></p>
-<p>»
-	<a href="https://build.suse.de/project/staging_projects/SUSE:SLE-15:GA">SLE 15</a> and
-	<a href="https://build.suse.de/project/staging_projects/SUSE:SLE-12-SP3:Update:Products:CASP30">CaaSP</a> staging dashboards
-</p>
-<p><a href="https://w3.suse.de/~okurz/openqa_suse_de_status.html">» Daily openQA review</a></p>
+<p><a href="https://build.suse.de/project/staging_projects/SUSE:SLE-15:GA">» SLE 15 staging dashboards</a></p>
+<p><a href="https://openqa.io.suse.de/openqa-review/">» Daily openQA review</a></p>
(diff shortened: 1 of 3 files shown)


perlpunk (approved):

okurz:
Fixed commit message length

templates/webapi/branding/openqa.suse.de/docbox.html.ep
```diff
@@ -2,9 +2,8 @@
 
 <p><a href="http://open.qa">» More information regarding openQA</a></p>
 <p><a href="https://confluence.suse.com/display/openqa/openQA">» Description of internal setup and administration</a></p>
-<p><a href="irc://irc.suse.de/testing">» IRC channel</a></p>
+<p><a href="https://chat.suse.de/channel/testing">» chat</a></p>
```
Martchus, line 5: All points start with a capital letter here so I wouldn't make an exception here.

templates/webapi/branding/openqa.suse.de/docbox.html.ep
```diff
@@ -2,9 +2,8 @@
...
 <p>»
 	<a href="https://build.suse.de/project/staging_projects/SUSE:SLE-15:GA">SLE 15</a> and
-	<a href="https://build.suse.de/project/staging_projects/SUSE:SLE-12-SP3:Update:Products:CASP30">CaaSP</a> staging dashboards
```
Martchus, line 8: I suppose the previous point should then just be "SLE 15 staging dashboards" (and could be a single line).

Martchus (changes requested):
The doc box looks different on openqa.suse.de. I suppose there's a local override in place anyways. Maybe we can completely rely on that override? (It would make more sense to have things like that in the salt repo.)

templates/webapi/branding/openqa.suse.de/docbox.html.ep
okurz: done

templates/webapi/branding/openqa.suse.de/docbox.html.ep
okurz: done

okurz:
> The doc box looks different on openqa.suse.de. I suppose there's a local override in place anyways. Maybe we can completely rely on that override? (It would make more sense to have things like that in the salt repo.)

Yes, we have an override in salt, see https://github.com/os-autoinst/salt-states-openqa/blob/master/openqa/links_docbox.html . This is just retroactively updating the current template to act as "example". But you are right. I now made the docbox near-empty with just the "Welcome"-line

Martchus (approved):

PR #4535, opened 2022-02-25
WIP: Atomic dependency handling for openqa-clone-job

Martchus:
* Very WIP, e.g. the actual handling of dependency parameters is still missing.
* See https://progress.opensuse.org/issues/103971#note-12 for how the API would be used.


Martchus:
I must say this was *way* easier than expected. I could mainly add new code without having to restructure any existing code except the `create` function itself (although most of its code has just been moved to the new function `_create_job` without much changes in the logic). Since the API remains backwards compatible all existing API tests pass unchanged.

lib/OpenQA/WebAPI/Controller/API/V1/Job.pm
```diff
@@ -219,60 +220,174 @@ sub overview {
...
-=item create()
+=item _eval_param_grouping()
 
-Creates a job given a list of settings passed as parameters. TEST setting/parameter
-is mandatory and should be the name of the test.
+Removes job-specific parameters from the specified hash and returns a new hash
+containing the removed parameters (where the keys are job suffixes).
+
+So "{foo => 1, bar:jobA => 2, baz:jobB => 3}" will be turned into "{foo => 1}"
+returnig "{jobA => {bar => 2}, jobB => {baz => 3}}".
```
okurz, line 229: s/returnig/returning/, better yielding

lib/OpenQA/WebAPI/Controller/API/V1/Job.pm
```diff
@@ -219,60 +220,174 @@ sub overview {
...
+sub _eval_param_grouping ($params) {
+    my %grouped_params;
+    for my $param_key (keys %$params) {
+        my $job_suffix_start = rindex($param_key, ':');
+        next unless $job_suffix_start >= 0;
+        my $job_suffix = substr $param_key, $job_suffix_start + 1;
+        my $setting_key = substr $param_key, 0, $job_suffix_start;
+        $grouped_params{$job_suffix}->{$setting_key} = delete $params->{$param_key};
+    }
+    return \%grouped_params;
```
okurz, line 244: Something like this maybe

```suggestion
    map { ($job, $setting) = split /:/; $setting => {$job => $x{$_}} } grep { /:/ } keys %$params;
```

For a test:

```
perl -e '%x = ("foo" => 1, "bar:jobA" => 2, "baz:jobB" => 3); %h = map { ($job, $setting) = split /:/; $setting => {$job => $x{$_}} } grep { /:/ } keys %x; use Data::Dumper; print(Dumper(\%h))'
```

shows

```
$VAR1 = {
          'jobB' => {
                      'baz' => 3
                    },
          'jobA' => {
                      'bar' => 2
                    }
        };
```

lib/OpenQA/WebAPI/Controller/API/V1/Job.pm
```diff
@@ -219,60 +220,174 @@ sub overview {
...
+    return undef unless defined $job_suffixes;
+    my $deps = $self->{_dependencies};
+    for my $parent_job_suffix (split /\s*,\s*/, $job_suffixes) {
+        push @$deps,
+          {
+            child_job_id => $child_job_suffix,
+            parent_job_id => $parent_job_suffix,
+            dependency => $type,
+          };
+    }
```
okurz, line 270:
```suggestion
    my $deps = ($self->{_dependencies}, map { {
            child_job_id => $child_job_suffix,
            parent_job_id => $_,
            dependency => $type } } split /\s*,\s*/, $job_suffixes;
```

okurz (changes requested):

lib/OpenQA/WebAPI/Controller/API/V1/Job.pm
Martchus: I'll change it to returning but yield is mainly used in the context of generators/co-routines so I wouldn't use it here.

lib/OpenQA/JobDependencies/Constants.pm
```diff
@@ -14,6 +16,8 @@ use constant DIRECTLY_CHAINED => 3;
 use constant DEPENDENCIES => (CHAINED, PARALLEL, DIRECTLY_CHAINED);
 use constant CHAINED_DEPENDENCIES => (CHAINED, DIRECTLY_CHAINED);
 
+our @EXPORT = qw(CHAINED PARALLEL DIRECTLY_CHAINED);
```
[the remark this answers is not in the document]
Martchus: I'll check whether it works without it. The only goal here is to avoid `OpenQA::JobDependencies::Constants::…` all the time.

lib/OpenQA/WebAPI/Controller/API/V1/Job.pm
Martchus: * This breaks the deletion.
* This breaks if multiple `:` are present. The splitting should really only occur after the last colon.

But I might be able to make it use `map`.

lib/OpenQA/WebAPI/Controller/API/V1/Job.pm
Martchus: This makes no sense. `$self->{_dependencies}` is supposed to be populated here.

lib/OpenQA/JobDependencies/Constants.pm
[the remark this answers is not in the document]
Martchus: Without it I get

```
"CHAINED" is not exported by the OpenQA::JobDependencies::Constants module
"PARALLEL" is not exported by the OpenQA::JobDependencies::Constants module
"DIRECTLY_CHAINED" is not exported by the OpenQA::JobDependencies::Constants module
Can't continue after import errors at /hdd/openqa-devel/repos/openQA/lib/OpenQA/WebAPI/Controller/API/V1/Job.pm line 10.
```

Do you want me to keep using the `OpenQA::JobDependencies::Constants::…` everywhere?

lib/OpenQA/WebAPI/Controller/API/V1/Job.pm
Martchus: This also breaks if you add `, "bar2:jobA" => 4` (which should result in multiple keys within `jobA`'s nested hash but doesn't.

lib/OpenQA/JobDependencies/Constants.pm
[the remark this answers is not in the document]
okurz: You would need to use

`use OpenQA::JobDependencies::Constants qw(CHAINED PARALLEL DIRECTLY_CHAINED);`

on the imports. I am ok if you use EXPORT but then better export all the symbols that you think are useful, not just the ones you just happen to need in this PR

lib/OpenQA/WebAPI/Controller/API/V1/Job.pm
okurz: I am sure there is a way to use map to create the hash with correct values even if the same key appears multiple times but if we can't find it then leave it

lib/OpenQA/WebAPI/Controller/API/V1/Job.pm
okurz: oh, I see. I misread `my $deps = $self->{_dependencies} ` as deep copy which it isn't. So maybe

```suggestion
$self->{_dependencies} = ($self->{_dependencies}, map { {
            child_job_id => $child_job_suffix,
            parent_job_id => $_,
            dependency => $type } } split /\s*,\s*/, $job_suffixes);
```

because you can merge a second hash (here created by "map") into the first like `%h = (%h, %j)` or here more like `%h = (%h, map { … } @j)`

lib/OpenQA/WebAPI/Controller/API/V1/Job.pm
```diff
@@ -219,60 +220,174 @@ sub overview {
...
-    my %up_params = map { uc $_ => $params->{$_} } keys %$params;
+    my %up_params = map { uc $_ => $global_params->{$_} } keys %$global_params;
+    $up_params{uc $_} = $job_specific_params->{$_} for keys %$job_specific_params;
+
     # restore URL encoded /
     my %params = map { $_ => $up_params{$_} =~ s@%2F@/@gr } keys %up_params;
-    if (!$params{TEST}) {
-        return $self->render(json => {error => 'TEST field mandatory'}, status => 400);
-    }
+    die "TEST field mandatory\n" unless $params{TEST};
```
okurz, line 291: Maybe this just looks weird in the diff but just to double check: Are you changing the error rendering back into a die? Didn't you just recently have a PR doing the opposite?

lib/OpenQA/WebAPI/Controller/API/V1/Job.pm
```diff
@@ -219,60 +220,174 @@ sub overview {
...
-                $self->gru->enqueue_download_jobs($downloads);
-                $job->calculate_blocked_by;
-                OpenQA::Scheduler::Client->singleton->wakeup;
+                if (keys %$grouped_params) {
+                    # create as many jobs as unique ":"-suffixes exist
+                    $self->_create_job($global_params, $_, $grouped_params->{$_}) for keys %$grouped_params;
+                }
+                else {
+                    # create a single job if there are no job-specific parameters
+                    $self->_create_job($global_params);
```
okurz, line 375: That additional level of indendation is a smell for me. How about making the additional parameters mandatory in `_create_job` and adding the single job parameter here as default in case `$grouped_params` is empty. Then you might be able to simplify the condition in https://github.com/os-autoinst/openQA/pull/4535/files#diff-6da53efb4105c5db577df2d5cd890fac0c8f83cd05a4b6a80bf5b067a4597fb2R307 as well

lib/OpenQA/WebAPI/Controller/API/V1/Job.pm
Martchus: But we're not creating "the hash" here. We're creating one hash containing an arbitrary number of nested hashes - and the nested hashes are not populated at once.

lib/OpenQA/WebAPI/Controller/API/V1/Job.pm
Martchus: That's harder to read and unlikely to be more efficient. And it gets even uglier because you needed to make a new hash-ref (by wrapping in `[]` instead of `()`) and you needed to de-ref `$self->{_dependencies}` to retain the flat structure.

lib/OpenQA/WebAPI/Controller/API/V1/Job.pm
Martchus: First, my other PR was about a whole different API. Second, this is still rendered as user-facing error because the controller code simply renders all `die` messages as such. (That isn't changed by my PR. A more fain-grained error handling would be nice but I consider it out of scope for now.)

lib/OpenQA/WebAPI/Controller/API/V1/Job.pm
Martchus: Not sure whether I follow. Should the for-loop go into `_create_job` (which should then likely be named `_create_jobs_internally`)? I also don't see whether this helps simplifying the condition. The condition is about whether `ids` or `id` is populated and the latter is still required for keeping backwards compatibility. Only if there is at least one suffix (even if there's still just a single job¹) the `ids` key should be populated.

¹Otherwise it is more inconsistent in my opinion, e.g. I'd needed to distinguish between `id` and `ids` in the clone script.

lib/OpenQA/JobDependencies/Constants.pm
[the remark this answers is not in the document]
okurz: ok, fine

lib/OpenQA/WebAPI/Controller/API/V1/Job.pm
okurz: I didn't invent it. I thought this is the way how to merge hashes in perl

Martchus: I think of it more as "concatenation". "Merge" means to me something becomes part of something else but here we make something new and throw away the original parts.

okurz: Yes, I meant that `%h = (%h, %j)` is the way to concatenate hashes in perl

Martchus:
By the way, for better backwards compatibility I *could* make the splitting at `:` optional - so it would only happen if an extra parameter is specified (e.g. `multiple_jobs=1`). I refrained from that extra complexity because I think we don't use `:` in keys so existing usage shouldn't be impacted. (Note that this change actually still allows having `:` in keys as the splitting only happens after the last `:`. It just requires using the new syntax, e.g. by just adding another trailing `:`. That means one can still clone jobs with `:` setting keys, e.g. `TEST:FOO=bar` would be passed as `TEST:FOO:42=bar` by the clone script.)

lib/OpenQA/WebAPI/Controller/API/V1/Job.pm
okurz: > Not sure whether I follow. Should the for-loop go into `_create_job` (which should then likely be named `_create_jobs_internally`)?

I think so, yes. I haven't tried this myself yet to be sure this would be better, sorry.

Martchus: Ok, I'll try it tomorrow and see whether it is better or not.

Martchus: Moved into a separate function.

okurz (approved):

Martchus:
I've rebased and added some documentation which hopefully makes the review a bit easier. I've also noticed that `openqa-cli` currently doesn't allow to pass parameters with `:` in it so I extended the regex (and corresponding tests).

kalikiana (approved):
I love the docs.

Martchus:
Thanks for the nice comment :-)

okurz (commented):
I haven't seen the doc changes at time of approval. I recommend to also extend the openqa-cli help correspondingly a bit and point to the users guide for details

PR #4537, opened 2022-02-28
openqa-clone-job: Create jobs atomically to avoid dependency problems

Martchus:
**https://github.com/os-autoinst/openQA/pull/4535 must be deployed first.** (Therefore the not-ready label.)

---

* Non-atomic job creation means that the scheduler might see a half-created
  dependency tree and thus already starts half of it regardless of further
  parallel or directly chained dependencies (which therefore breaks
  parallel and directly chained dependencies).
* This change posts all jobs atomically (using a single API call) to avoid
  the problem described in the previous point.
* See https://progress.opensuse.org/issues/103962


lib/OpenQA/Script/CloneJob.pm
```diff
@@ -204,14 +205,18 @@ sub get_deps ($job, $options, $job_type) {
...
-        print "Created job #$clone_id: $job->{name} -> $base_url/t$clone_id\n";
-        clone_job($_, $options, $clone_map, $depth + 1, $clone_id) for @$child_list;    # clone children if needed
-        return $clone_id;
+    if (!$tx->error && ref $json eq 'HASH' && ref $json->{ids} eq 'HASH') {
+        my $cloned_jobs = $json->{ids};
+        my $base_url = openqa_baseurl($url_handler->{local_url});
+        for my $orig_job_id (keys %$cloned_jobs) {
+            my $orig_job = $jobs->{$orig_job_id};
+            my $cloned_job_id = $cloned_jobs->{$orig_job_id};
+            print "Created job #$cloned_job_id: $orig_job->{name} -> $base_url/t$cloned_job_id\n";
```
okurz, line 217: This will still be the same message as in before, just renamed variables, right?

t/35-script_clone_job.t
```diff
@@ -256,25 +254,30 @@ subtest 'overall cloning with parallel and chained dependencies' => sub {
...
 
-    is_deeply \%clone_map, {41 => 141, 42 => 142, 43 => 143}, 'jobs cloned as expected' or diag explain \%clone_map;
-    is $post_args[0]->[3]->{TEST}, 'parent', 'parent job 41 cloned (as 141)';
-    is $post_args[1]->[3]->{TEST}, 'main', 'main job 42 cloned (as 142)';
-    is $post_args[1]->[3]->{_PARALLEL_JOBS}, '141', 'main job cloned to start parallel with parent job 141';
-    is $post_args[2]->[3]->{TEST}, 'child', 'child job 43 cloned (as 143)';
-    is $post_args[2]->[3]->{_START_AFTER_JOBS}, '142', 'child job cloned to start after main job 142';
-    is scalar @post_args, 3, 'exactly 3 jobs posted';
+    subtest 'post args' => sub {
+        is scalar @post_args, 1, 'exactly onn post call made' or return;
```
okurz, line 269:
```suggestion
        is scalar @post_args, 1, 'exactly one post call made' or return;
```

t/35-script_clone_job.t
```diff
@@ -222,24 +219,25 @@ subtest 'get 2 nodes HA cluster with get_deps' => sub {
...
+    my $test = sub { OpenQA::Script::CloneJob::handle_tx($tx, \%url_handler, {}, {42 => {name => 'testjob'}}) };
     throws_ok { $test->() } qr/Failed to create job, empty response/, 'empty response handled';
     $tx->res->body('foobar');
     throws_ok { $test->() } qr/Failed to create job, server replied: "foobar"/, 'unexpected non-JSON handled';
     $tx->res->code(200)->body('{"foo": "bar"}');
     throws_ok { $test->() } qr/Failed to create job, server replied: \{ foo => "bar" \}/, 'unexpected JSON handled';
-    ($tx = Mojo::Transaction->new)->res->code(200)->body('{"id": 43}');
-    combined_like { $test->() } qr|Created job #43: testjob -> base-url/t43|, 'expected response';
+    ($tx = Mojo::Transaction->new)->res->code(200)->body('{"ids": {"42": 43}}');
+    combined_like { $test->() } qr|Created job #43: testjob -> https://base-url/t43|, 'expected response';
```
okurz, line 230: the output changed here only because of how you use the URL differently within the test, so normally we wouldn't see a difference, right?

lib/OpenQA/Script/CloneJob.pm
Martchus: That's the idea. Of course the ordering of the log messages differs. In the current version these messages are printed after each post call but with my change all of these messages are printed at the end (after the single post call for all jobs and thus after all asset-related messages).

For better machine-readability we might want to add a `--json-output` flag (which would simply print the JSON response from the API as-is). That might be useful for going forward with the investigation script.

t/35-script_clone_job.t
Martchus: Yes, it simply uses the `--host` URL here. Previously the test's fake values didn't include a scheme but now it does so now the test expectation actually looks more like the output would look in reality.

okurz (approved):

okurz:
> **#4535 must be deployed first.** (Therefore the not-ready label.)

If you mention the pull request in a markdown list then we actually see a preview of the status. So instead of linking like this #4535, do this
* #4535

t/api/04-jobs.t
```diff
@@ -979,6 +979,7 @@ subtest 'handle settings when posting job' => sub {
 subtest 'do not re-generate settings when cloning job' => sub {
     my $job_settings = $jobs->search({test => 'autoupgrade'})->first->settings_hash;
     clone_job_apply_settings([qw(BUILD_SDK= ISO_MAXSIZE=)], 0, $job_settings, {});
+    $job_settings->{is_clone_job} = 1;
```
kalikiana, line 982: Why wasn't this set here before? On its own the test change looks wrong because we never checked it

Martchus: I moved the place where we set `is_clone_job` from `clone_job_apply_settings` to a different function¹ which is not executed as part of this test. So we need to set it manually here. The tests in `t/35-script_clone_job.t` still check whether `is_clone_job` is present without faking it. So this doesn't make our testing worse.

---

¹ From `clone_job_apply_settings` to `post_jobs`. This was necessary because the parameter is not an actual job setting and therefore should not get a `:`-suffix.

kalikiana (approved):

okurz:
https://github.com/os-autoinst/openQA/pull/4535 is most definitely already deployed at least for both openqa.suse.de and openqa.opensuse.org. Should we wait longer?

Martchus:
I would have merged it next week on Monday. I suppose waiting much longer doesn't help us much. Users who haven't update their personal instance yet but will nevertheless update `openqa-clone-job` will likely not update their personal instance any time soon so it makes no sense to wait for it.

Note that an old `openqa-clone-job` works with the new deployments and that's the important direction where we need to keep compatibility.

Btw, if one uses the new `openqa-clone-job` with an old deployment no faulty jobs are created but one instead gets an error message that the `TEST` variable is missing. If it comes up we should be able to help users (by simply telling them to upgrade their openQA deployment or to downgrade their openqa-clone-job script).

okurz:
> If one uses the new `openqa-clone-job` with an old deployment no faulty jobs are created but one instead gets an error message that the `TEST` variable is missing. If it comes up we should be able to help users (by simply telling them to upgrade their openQA deployment or to downgrade their openqa-clone-job script). I suppose I can write a mail on the openQA mailing list, on Slack and on our Matrix channel.

Some solution within code is always better and often not more complicated. If you really think the confusion will be very limited then let's leave it as is, otherwise maybe at least put a hint in the error handling of the scripts themselves

Martchus:
Rebased for asset cache changes (after cache has been updated for https://github.com/os-autoinst/openQA/pull/4543).

kalikiana:
for the record:

```
[  998s] ./t/40-script_openqa-clone-custom-git-refspec.t           (Wstat: 0 Tests: 43 Failed: 0)
[  998s]   TODO passed:   30, 32
[  998s] ./t/25-cache-client.t                                     (Wstat: 7424 Tests: 2 Failed: 1)
[  998s]   Failed test:  1
[  998s]   Non-zero exit status: 29
[  998s]   Parse errors: No plan found in TAP output
```

Martchus:
@perlpunk has already created a ticket: https://progress.opensuse.org/issues/108272

So I'm merging the PR as planned as this failure has nothing to do with my changes (and everything was green before).

PR #4538, opened 2022-02-28
Use simpler SQL for has_modules

perlpunk:
We only want to know if there are modules or not, so a COUNT
is sufficient (and will also not confuse postgres' query planner ;-)

Issue: https://progress.opensuse.org/issues/107701


okurz (approved):
Hotpatched OSD with `curl -sS https://patch-diff.githubusercontent.com/raw/os-autoinst/openQA/pull/4538.patch | sudo patch -p1 -d /usr/share/openqa/` and the initial loading of https://openqa.suse.de/tests/8232404 including the info box finishes near-immediate.

kraih (approved):

PR #4539, opened 2022-03-01
Avoid 500 response and error in logs when `filename` param not specified

Martchus:
* This is about the routes for serving the video/logfile pages.
* Prevent log message "Could not render a response …" and 500 response
* See https://progress.opensuse.org/issues/107719


okurz (commented):
Commit message is a bit too long, rest is fine

perlpunk (approved):

kalikiana (approved):

kalikiana:
> Commit message is a bit too long, rest is fine

Looks fine to me.

    Checking commit messages against "^.{1,72}(\n|$)"...
    - OK: "Avoid 500 response and error log when `filename` param not specified

PR #4545, opened 2022-03-04
Optimize jobs API by prefetching data

perlpunk:
For /jobs?... API requests, for each job found, another ~ 4 SQL queries
are executed.

Prefetching data will be faster (50% in my local tests).

Issue: https://progress.opensuse.org/issues/94111

### DIFF
--- a/lib/OpenQA/Schema/Result/Jobs.pm
+++ b/lib/OpenQA/Schema/Result/Jobs.pm
@@ -429,10 +429,11 @@ sub ws_send {
 }
 
 sub settings_hash {
-    my ($self) = @_;
+    my ($self, $prefetched) = @_;
     return $self->{_settings} if defined $self->{_settings};
     my $settings = $self->{_settings} = {};
-    for ($self->settings->all()) {
+    my $all = $prefetched || [$self->settings->all];
+    for (@$all) {
         # handle multi-value WORKER_CLASS
         if (defined $settings->{$_->key}) {
             $settings->{$_->key} .= ',' . $_->value;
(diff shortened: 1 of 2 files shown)


kalikiana (approved):

lib/OpenQA/Schema/Result/Jobs.pm
```diff
@@ -499,21 +502,29 @@ sub to_hash {
...
         push @{$j->{assets}->{$_->type}}, $_->name for @$assets;
     }
     if ($args{deps}) {
-        $j = {%$j, %{$job->dependencies}};
+        if ($dependencies) {
+            $j = {%$j, %{$job->dependencies($dependencies->{children} || [], $dependencies->{parents} || [])}};
+        }
+        else {
+            $j = {%$j, %{$job->dependencies()}};
+        }
```
okurz, line 527:
```suggestion
        $j = {%$j, $dependencies ? %{$job->dependencies($dependencies->{children} || [], $dependencies->{parents} || [])} : %{$job->dependencies()}};
```

but the amount of curly braces is hardly readable in both cases. Maybe better extract a function to return what is needed here

okurz (commented):
Do you think it's feasible to add a test checking for the number of SQL queries conducted and ensuring that it's now less than in before?

Martchus (commented):
Looks good. (I'm not approving to give you a chance to read @okurz 's comments before it would be merged automatically.)

lib/OpenQA/Schema/Result/Jobs.pm
perlpunk: I rearranged it a bit

okurz (commented):
Do you think it's feasible to add a test checking for the number of SQL queries conducted and ensuring that it's now less than in before?

Martchus (approved):

PR #4551, opened 2022-03-08
openqa-clone-job PART 2: Simplify and add --clone-parallel-children

Martchus:
**BASED ON https://github.com/os-autoinst/openQA/pull/4537; additionally https://github.com/os-autoinst/openQA/pull/4535 must be deployed first.** (Therefore the not-ready label.)

---

* See particular commit message for details (first commit is same as https://github.com/os-autoinst/openQA/pull/4537)
* Related ticket: https://progress.opensuse.org/issues/95783#note-33


okurz (changes requested):
This is yet another option to the already complicated set of parameters. You have to understand that users already copy-paste parameters together that don't make sense because many parameters sound alike and the difference is not understandable. I don't even know myself what "parallel children" could be and I don't know what the effective difference to clone-children and other options are. We have to come up with something better than this. What can we do so that "openqa-clone-job" does *the right thing* without the user needing specify this option? Can we replace the behaviour of another, existing option instead?

Martchus:
By the way, I've just noticed that in one of my manual tests I can reproduce a bug on `master` where a parallel cluster isn't completely cloned (although `--clone-children` is specified). That isn't reproducible on this branch anymore because the buggy code path has simply been removed.

> What can we do so that "openqa-clone-job" does the right thing without the user needing specify this option? Can we replace the behaviour of another, existing option instead?

Breaking compatibility. In my opinion `--clone-parallel-children` should be the default. It just doesn't generally make sense to clone only the parent of a parallel cluster. (To only clone a single child plus its parent one can still just clone *that* child.)

> I don't even know myself what "parallel children" could be

Jobs that have the `PARALLEL_WITH` setting. And considering the parent/child-relations don't seem arbitrary in the clusters I see on OSD it looks like testers know how what parallel parents/children are. (The server test is usually a parallel parent and the client tests parallel children.)

okurz:
> And since I'm already talking about defaults: I'd find an infinite max. depth when cloning children more intuitive by default when `--clone-children` is used (instead of cloning only direct children). In fact, that's what I had expected to be the current behavior and maybe even what the one who wrote the code meant to achieve.

I think breaking compatibility within the openqa-clone-job script is a different story than the API of the openQA server so if you think that we need to update and change how openqa-clone-job operates but in the process clean up old cruft then please do so

Martchus:
I've just edited my previous comment. I've asked on the chat about making `--clone-parallel-children` the default.

alvarocarvajald:
> Breaking compatibility. In my opinion `--clone-parallel-children` should be the default. It just doesn't generally make sense to clone only the parent of a parallel cluster. (To only clone a single child plus its parent one can still just clone _that_ child.)

+1 to `--clone-parallel-children` being the default. At least for HA tests, failing to clone all jobs (support server, cluster nodes and clients) would not make sense.

script/openqa-clone-job
```diff
@@ -119,9 +119,19 @@ Cloning a job will not fail if an asset is missing.
 
 =item B<--clone-children>
 
-Clone child jobs as well. This is useful if you want to clone a full
+Clone child jobs as well.
+
+=item B<--clone-parallel-children>
```
b10n1k, line 124: if there is no difference why keep both?

b10n1k: or add usage description. anything above ~1~ instance is multimachine in my mind. So i am not sure when `--clone-children` is suitable.

lib/OpenQA/Script/CloneJob.pm
```diff
@@ -221,54 +226,60 @@ sub handle_tx ($tx, $jobid, $job, $base_url, $options, $clone_map, $depth, $chil
...
-                return $clone_map->{$jobid} if defined $clone_map->{$jobid};
-            }
+sub clone_jobs ($jobid, $options) {
+    my $url_handler = create_url_handler($options);
+    clone_job($jobid, $url_handler, $options, my $post_params = {}, my $jobs = {});
+    my $tx = post_jobs($post_params, $url_handler, $options);
+    handle_tx($tx, $url_handler, $options, $jobs);
+}
+
+sub clone_job ($jobid, $url_handler, $options, $post_params = {}, $jobs = {}, $depth = 1) {
```
b10n1k, line 236: `$depth` is hardcoding here.

b10n1k (changes requested):
What worries me, and after our chat, is that we implement one more feature when there is already some alternatives. I am afraid that it is going to confuse some users. Likewise the double options in some cases (see: `--clone-children` and `--clone-parallel-children`). Which IIUC using `--max-depth` will make the one behaves exactly like the other. i think we can do better here.

However i like that the code is more tidy and clean now. kudos

lib/OpenQA/Script/CloneJob.pm
okurz: that's not hardcoding but just specifying a default value.

b10n1k: my bad. i rush to add the comment. ignore

script/openqa-clone-job
Martchus: But there *is* a difference. Since I'm likely going to make this the default it shouldn't matter much, though.

> So i am not sure when --clone-children is suitable.

This is about all kinds of dependencies (so it is also about any kind of chained dependencies). The specialty of MM tests is that they use parallel dependencies.

lib/OpenQA/Script/CloneJob.pm
```diff
@@ -221,54 +226,60 @@ sub handle_tx ($tx, $jobid, $job, $base_url, $options, $clone_map, $depth, $chil
...
-            for my $dependencies ($chained, $directly_chained, $parallel) {
-                # don't clone parallel jobs yet if children job type
-                next if $dependencies == $parallel && $job_type eq 'children';
-                clone_job($_, $options, $clone_map, $depth + 1) for @$dependencies;
-                # abort here if the job has already been cloned as part of the preceding recursive clone_job call
-                return $clone_map->{$jobid} if defined $clone_map->{$jobid};
-            }
+sub clone_jobs ($jobid, $options) {
+    my $url_handler = create_url_handler($options);
+    clone_job($jobid, $url_handler, $options, my $post_params = {}, my $jobs = {});
```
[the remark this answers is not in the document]
Martchus: No, `max-depth` is passed within `$options` like any other CLI option.

script/openqa-clone-job
b10n1k: thanks for replies. it is still spinning in my head the idea of a better naming but lets not block PR for now just to discuss this.

b10n1k (commented):
All of my questions were answered. the only thing i have is https://github.com/os-autoinst/openQA/pull/4551#pullrequestreview-903348076.

script/openqa-clone-job
Martchus: I now made it default so we don't have to think about a better name anymore.

Martchus:
About docs: I updated the `--help` of `openqa-clone-job` with each commit accordingly. Otherwise there's no explicit documentation for `openqa-clone-job`, it just comes up in examples of other documentation sections. However, I suppose the explanation about dependency handling I've added to `--help` within my last commit should be sufficient. I'll nevertheless review the general documentation we have about dependencies (https://open.qa/docs/#_job_dependencies) and add a sub-section for openQA clone job.

kalikiana (changes requested):
> Unfortunately #4537 broke passing job settings via CLI args. I've pushed a fix (and proper unit tests) to this branch. Hence it would be good to merge this one soon as well.

I hesitated, but since the code path changed a lot I'm fine with including the fix here.

However I think the subject needs to be updated. "part 2" is not descriptive and you're not adding that option afterall. How about

    openqa-clone-job: Hadle parallel children by default

kalikiana (approved):

PR #4552, opened 2022-03-09
doc: Improve documentation about job dependencies

Martchus:
* Simplify the wording while trying to be more accurate at the same time
* Add reference to examples for creating a single set of jobs
* Make a clear distinction between the concept of dependency types and the
  concept of parents/children
* Wrap paragraphs consistently
* Explain the term "cluster" as it is also often used as there are certain
  consequences applying to a "cluster"
* Mention that errors can be checked in the "scheduled product"
* See https://progress.opensuse.org/issues/103962

### DIFF
--- a/docs/WritingTests.asciidoc
+++ b/docs/WritingTests.asciidoc
@@ -346,7 +346,8 @@ In short, writing multi-machine tests adds a few more layers of complexity:
 3. actual technical realization (i.e. <<Networking.asciidoc#networking,custom networking>>)
 
 ==== Job dependencies
-There are three dependency *types*: `CHAINED`, `DIRECTLY_CHAINED` and `PARALLEL`
+There are different dependency *types*, most importantly _chained_ and
+_parallel_ dependencies.
 
 A dependency is always between two jobs where one of the jobs is the _parent_
 and one the _child_. The concept of parent and child jobs is *orthogonal* to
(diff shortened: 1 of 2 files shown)


docs/WritingTests.asciidoc
```diff
@@ -346,34 +346,87 @@ In short, writing multi-machine tests adds a few more layers of complexity:
 3. actual technical realization (i.e. <<Networking.asciidoc#networking,custom networking>>)
 
 ==== Job dependencies
-There are different dependency *types* (see subsequent section). Additionally, dependencies can be machine-specific (see "Inter-machine dependencies" section).
+There are three dependency *types*: `CHAINED`, `DIRECTLY_CHAINED` and `PARALLEL`
 
-===== Dependency types
-There are 3 types of dependencies: `CHAINED`, `DIRECTLY_CHAINED` and `PARALLEL`
+A dependency is always between two jobs where one of the jobs is the _parent_
+and one the _child_. The concept of parent and child jobs is *orthogonal* to
```
okurz, line 352: I consider this sentence "The concept of parent and child jobs is *orthogonal*" hard to read. Maybe we should try to move some explanation further down so that the difference for "chained" and "parallel" appears earlier?

docs/WritingTests.asciidoc
```diff
@@ -346,34 +346,87 @@ In short, writing multi-machine tests adds a few more layers of complexity:
 3. actual technical realization (i.e. <<Networking.asciidoc#networking,custom networking>>)
 
 ==== Job dependencies
-There are different dependency *types* (see subsequent section). Additionally, dependencies can be machine-specific (see "Inter-machine dependencies" section).
+There are three dependency *types*: `CHAINED`, `DIRECTLY_CHAINED` and `PARALLEL`
```
okurz, line 349: I think a reader benefits more and might be less confused if on this level we only distinguish between chained and parallel dependencies and also not use the capital case control words but plain english referring to the section headers below.. The directly-chained dependency is anyway explained below the section "Chained dependencies".

okurz (changes requested):

docs/WritingTests.asciidoc
Martchus: I agree that the meaning of *orthogonal* in this context is hard to understand. Maybe "The concept of parent and child jobs is independent of …". Note that I've been adding that sentence because people seem to mix it up. Therefore I thought before going into specific dependency types it makes sense to clear it up.

docs/WritingTests.asciidoc
Martchus: I was also thinking about that because `CHAINED` and so on are actually "keywords" the user will never have to type literally. Not sure about directly chained. Then I would also go back to "There are different dependency types, i.a. chained and parallel dependencies." (because "chained" and "directly-chained" dependencies are still distinct types).

docs/WritingTests.asciidoc
okurz: ok, independant is good. But also I suggested to simply move this explanation further down.

Martchus: Or maybe "The concept of parent and child jobs applies regardless of the dependency type."?

docs/WritingTests.asciidoc
Martchus: I removed the keywords and mention directly chained dependencies only in the sub section.

docs/WritingTests.asciidoc
okurz: Now I am again confused. Isn't child+parent defined by using the CHAINED dependency?

docs/WritingTests.asciidoc
```diff
@@ -346,7 +346,7 @@ In short, writing multi-machine tests adds a few more layers of complexity:
 3. actual technical realization (i.e. <<Networking.asciidoc#networking,custom networking>>)
 
 ==== Job dependencies
-There are three dependency *types*: `CHAINED`, `DIRECTLY_CHAINED` and `PARALLEL`
+There different dependency *types*, i.a. _chained_ and _parallel_ dependencies.
```
okurz, line 349: missing verb. And please avoid the abbreviation. I don't even know what you mean. There is https://en.wiktionary.org/wiki/i.a. , do you mean that? Again, better avoid

Martchus: https://dict.leo.org/englisch-deutsch/i.a. - but yes, I suppose I should better avoid it.

docs/WritingTests.asciidoc
okurz: So after talking about we realized that we use parent+child and people sometimes wrongly assume that childs always run *after* parents when what we have is that parents more "lead the way" and the dependency type configures how, if parallel or chained

okurz (approved):

kalikiana (approved):

PR #4560, opened 2022-03-16
Return 404 for Jobs not found

perlpunk:
Issue: https://progress.opensuse.org/issues/108401


kalikiana (approved):
Nice fix and nice unit test, too.

perlpunk:
I'll force push. something was wrong yesterday with github, so one check is pending forever

baierjan (approved):

PR #4563, opened 2022-03-17
Let `openqa-clone-custom-git-refspec` handle dependencies better

Martchus:
* Specify the `openqa-clone-job` parameter `--parental-inheritance` so the
  modified variables end up in all jobs which are possibly cloned
    * We specify `--skip-chained-deps` but for parallel dependencies
      parents are cloned as well so it makes a difference here.
* Do *not* override the `TEST` variable. Otherwise all cloned jobs will
  have the same `TEST` value which is very confusing, e.g. only one of the
  jobs will be shown within the test result overview. Instead, the branch
  name is only appended.
* Based on user feedback on https://github.com/os-autoinst/openQA/pull/4537
* Related to the https://progress.opensuse.org/issues/103971 epic
* Tested locally by modifying the script to clone to my local openQA
  instance

### DIFF
--- a/lib/OpenQA/Script/CloneJob.pm
+++ b/lib/OpenQA/Script/CloneJob.pm
@@ -37,7 +37,7 @@ sub clone_job_apply_settings ($argv, $depth, $settings, $options) {
 
     for my $arg (@$argv) {
         # split arg into key and value
-        unless ($arg =~ /([A-Z0-9_]+)=(.*)/) {
+        unless ($arg =~ /([A-Z0-9_]+\+?)=(.*)/) {
             warn "arg '$arg' does not match";
             next;
         }
@@ -49,6 +49,12 @@ sub clone_job_apply_settings ($argv, $depth, $settings, $options) {
         if (!defined $value || $value eq '') {
             delete $settings->{$key};
             next;
+        }
+
+        # allow appending via `+=`
+        if (substr($key, -1) eq '+') {
+            $key = substr $key, 0, -1;
+            $value = ($settings->{$key} // '') . $value;
         }
 
         # assign value to key, delete overrides
--- a/script/openqa-clone-custom-git-refspec
+++ b/script/openqa-clone-custom-git-refspec
@@ -1,6 +1,6 @@
 #!/bin/bash -e
 
-clone_args="${clone_args:-"--skip-chained-deps --within-instance"}"
+clone_args="${clone_args:-"--skip-chained-deps --parental-inheritance --within-instance"}"
 
 usage() {
     cat << EOF
@@ -130,14 +130,14 @@ Please try 'curl $json_url' or select another job, e.g. in the same scenario: $h
         needles_dir="${needles_dir:-"$old_productdir/needles"}"
     fi
     local repo_branch="${repo_branch:-"$repo_name#$branch"}"
-    local test="${test:-"$testsuite@$repo_branch"}"
+    local test_suffix="${test_suffix:-"@$repo_branch"}"
     local build="${build:-"$repo_name#$pr"}"
     local casedir="${casedir:-"$repo#$branch"}"
     local GROUP="${GROUP:-0}"
     local dry_run="${dry_run:-""}"
     local scriptdir
     scriptdir=$(dirname "${BASH_SOURCE[0]}")
-    local cmd="$dry_run $scriptdir/openqa-clone-job $clone_args \"$host\" \"$job\" _GROUP=\"$GROUP\" TEST=\"$test\" BUILD=\"$build\" CASEDIR=\"$casedir\" PRODUCTDIR=\"$productdir\" NEEDLES_DIR=\"$needles_dir\""
+    local cmd="$dry_run $scriptdir/openqa-clone-job $clone_args \"$host\" \"$job\" _GROUP=\"$GROUP\" TEST+=\"$test_suffix\" BUILD=\"$build\" CASEDIR=\"$casedir\" PRODUCTDIR=\"$productdir\" NEEDLES_DIR=\"$needles_dir\""
     [[ ${#args[@]} -ne 0 ]] && cmd=$cmd"$(printf " '%s'" "${args[@]}")"
     if [[ -n "$MARKDOWN" ]]; then
         eval "$cmd" | sed 's/^Created job.*: \([^ ]*\) -> \(.*\)$/* [\1](\2)/'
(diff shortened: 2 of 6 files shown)


Martchus:
Context: https://suse.slack.com/archives/C02CANHLANP/p1647431632743909 (thread on #eng-testing channel)

script/openqa-clone-custom-git-refspec
```diff
@@ -130,14 +130,13 @@ Please try 'curl $json_url' or select another job, e.g. in the same scenario: $h
         needles_dir="${needles_dir:-"$old_productdir/needles"}"
     fi
     local repo_branch="${repo_branch:-"$repo_name#$branch"}"
-    local test="${test:-"$testsuite@$repo_branch"}"
```
b10n1k, line 133: i am thinking that only the expansion needs to be removed. `TEST` can be used to give a custom name (which happens to have use in some cases). Then i assume you have to check somehow if is set and only then add it to the `$cmd`. WDYT?

b10n1k (commented):
i would like to keep the option to have the ability to use `TEST` when i want. 

btw another side effect of the TEST as for now is the deletion. in case you want to delete a job (let say to bring the _overview_ in a previous state). The overview will keep showing the jobs triggered from `openqa-clone-custom-git-refspec` until you find and delete all of them.

script/openqa-clone-custom-git-refspec
Martchus: So `TEST` should not be altered by default but setting the environment variable `test` should still enable one to set it?

> Then i assume you have to check somehow if is set and only then add it to the $cmd. WDYT?

That needed to be done, otherwise we would delete the `TEST` setting by default.

I'd say it makes more sense to simply allow specifying additional, arbitrary parameters to append to the clone invocation (instead of making a special case for `TEST`).

Martchus: And note that this already works, e.g. 
```
script/openqa-clone-custom-git-refspec --dry-run https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/14495 https://openqa.suse.de/tests/8291973 BUILD=b10n1k/os-autoinst-distri-opensuse#14495 TEST=foo SOME=setting```
invokes
```
script/openqa-clone-job --skip-chained-deps --parental-inheritance --within-instance https://openqa.suse.de 8291973 _GROUP=0 BUILD=b10n1k/os-autoinst-distri-opensuse#14495 CASEDIR=https://github.com/b10n1k/os-autoinst-distri-opensuse.git#hpc_scipy PRODUCTDIR=os-autoinst-distri-opensuse/products/sle NEEDLES_DIR=sle/products/sle/needles BUILD=b10n1k/os-autoinst-distri-opensuse#14495 TEST=foo  SOME=setting
```

I think that should be good enough to set arbitrary settings. Of course you still need to keep in mind that *all* jobs will have that setting so it still doesn't make much sense to set e.g. `TEST` for jobs with dependencies.

**Review by okurz — CHANGES_REQUESTED** [review]
>     * Do _not_ override the `TEST` variable. Otherwise all cloned jobs will
>       have the same `TEST` value which is very confusing, e.g. only one of the
>       jobs will be shown within the test result overview. I suppose this is ok
>       because we already include the repo/PR in the `BUILD` variable.

No, that's not a good idea because as long as other parameters stay the same the newly generated job would still show up in the original scenario, e.g. would show up in "Next & Previous" results.

**Comment by Martchus** [comment]
True. Then maybe a `TEST+=` feature for `openqa-clone-job` similar to the one I've added to the restart API (or maybe even more generic to allow appending to all variables)? (It cannot be on API level because in the context of posting a job there's no "previous job" where the variable value could be appended to.)

**Comment by okurz** [comment]
> True. Then maybe a `TEST+=` feature for `openqa-clone-job` similar to the one I've added to the restart API (or maybe even more generic to allow appending to all variables)?

ok

File: script/openqa-clone-custom-git-refspec (hunk 1 above)
Comment (kalikiana) [reply script/openqa-clone-custom-git-refspec:133]: Is there an alternative right now to override the name? I'm honstly not sure it's documented anywhere and I can see your argument with regard to inconsistencies. I would consider it common practice, though, and we should at least think of other options before breaking that use case.

Comment (kalikiana) [reply script/openqa-clone-custom-git-refspec:133]: FTR we discussed it briefly in the daily. The most practical option might be to support +TEST here (maybe not in this PR)

File: lib/OpenQA/Script/CloneJob.pm [hunk 2]
```diff
@@ -51,6 +51,12 @@ sub clone_job_apply_settings ($argv, $depth, $settings, $options) {
             next;
         }
 
+        # allow appending via `+=`
+        if (substr($key, -1) eq '+') {
+            $key = substr $key, 0, -1;
+            $value = ($settings->{$key} // '') . $value;
+        }
+
```
okurz, line 59: I think this is a great feature. Please make sure it's documented.

t/40-script_openqa-clone-custom-git-refspec.t
```diff
@@ -36,36 +36,37 @@ isnt run_once($args), 0, 'without network we fail (without error)';
...
 is $ret, 0, 'exits successfully';
 $prefix .= ' testsuite=new_test needles_dir=/my/needles productdir=my/product';
 $dirs = 'PRODUCTDIR=my/product NEEDLES_DIR=/my/needles';
-my $expected_custom_re = qr{https://openqa.opensuse.org 1234 _GROUP=0 TEST=new_test\@user/repo#my/branch.*${dirs}};
+my $expected_custom_re = qr{https://openqa.opensuse.org 1234 _GROUP=0 .*${dirs}};
 combined_like { $ret = run_once('', $prefix) } $expected_custom_re, 'testsuite and dirs can be overridden';
 is $ret, 0, 'exits successfully';
 my $args_trailing = 'https://github.com/me/repo/pull/1/ https://openqa.opensuse.org/tests/1';
-test_once $args_trailing, qr{TEST=my_test\@user/repo#my/branch.*}, 'trailing slash ignored';
+test_once $args_trailing, qr{BUILD=user/repo#1.*}, 'trailing slash ignored';
```
okurz, line 61: where does this change come from that instead of the fork and branch it would append the pull request number?

okurz (changes requested):

okurz (approved):
Tested manually, output:

```
++(script/openqa-clone-custom-git-refspec:145): clone_job(): script/openqa-clone-job --skip-chained-deps --parental-inheritance --within-instance https://openqa.opensuse.org 2264098 _GROUP=0 TEST+=@okurz/os-autoinst-distri-openQA#feature/next2 BUILD=okurz/os-autoinst-distri-openQA#feature/next2 CASEDIR=https://github.com/okurz/os-autoinst-distri-openQA.git#feature/next2 PRODUCTDIR=os-autoinst-distri-openQA NEEDLES_DIR=openqa/needles FOO=bar
``

Created job #2264154: openqa-Tumbleweed-dev-x86_64-Build:TW.10905-openqa_from_git@64bit-2G -> https://openqa.opensuse.org/t2264154

which looks good.

kalikiana (approved):

PR #4571, opened 2022-03-21
Fix wrongly skipping directly chained jobs

Martchus:
* Consider the result of the relevant directly chained parent job when
  deciding whether to skip the next job (instead of just considering the
  result of the last job that has been executed)
    * Rework queuing logic accordingly
    * In particular, keep track of the currently relevant parent when
      grabbing the next job from the queue
* See https://progress.opensuse.org/issues/108476

---

~~Still a draft because there's another problem to solve: The worker
doesn't really know whether a test has failed or not at this point (except
when the backend process exists with a non-zero exit code or some
other worker-related error happened). Hence the the logic for skipping
directly chained jobs is still flawed.~~

This is solved now by simply taking the web UI's response into account
when setting a job to done.

See the particular commit messages for detail.

### DIFF
--- a/t/24-worker-jobs.t
+++ b/t/24-worker-jobs.t
@@ -101,6 +101,7 @@ sub wait_until_uploading_logs_and_assets_concluded {
     has last_error => undef;
     has fail_job_duplication => 0;
     has configured_retries => 5;
+    has fake_result => undef;
     sub send {
         my ($self, $method, $path, %args) = @_;
         my $params = $args{params};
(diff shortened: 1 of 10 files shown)


t/24-worker-overall.t
```diff
@@ -211,81 +212,126 @@ subtest 'accept or skip next job' => sub {
...
+            [3, [], [0, 1, 2]],
+            [4, [5], [0, 1, 4]],
+            [5, [], [0, 1, 4]],
+            [6, [7, [8, 9, [10, 11]], 12], [0]],
+            [7, [[8, 9, [10, 11]], 12], [0]],
+            [8, [9, [10, 11]], [0, 8]],
+            [9, [[10, 11]], [0, 8]],
+            [10, [11], [0, 8, 10]],
+            [11, [], [0, 8, 10]],
+            [12, [], [0]],
```
Martchus, line 228: Each row now contains the relevant parents for the current iteration step. Otherwise everything is the same.

okurz (changes requested):
The git commit message "api: Return result/reason correctly string when setting job done" is likely wrong. Better rephrase, the word "string" looks like it doesn't belong there

okurz (approved):

t/24-worker-jobs.t
```diff
@@ -114,7 +115,9 @@ sub wait_until_uploading_logs_and_assets_concluded {
             $self->last_error('fake API error');
             return Mojo::IOLoop->next_tick(sub { $args{callback}->(0) });
         }
-        Mojo::IOLoop->next_tick(sub { $args{callback}->({}) }) if $args{callback} && $args{callback} ne 'no';
+        my %a;
```
kraih, line 118: Better not use `a` and `b` as identifiers, they are quite magical.

kraih (approved):

PR #4573, opened 2022-03-21
PART 2: Remove undesired Wants=network in systemd + graceful cacheservice start - after #4577

okurz:
After
* #4577

* Handle cacheservice startup errors gracefully
* systemd: Remove undesired Wants=network.target (boo#1196359)

See: https://progress.opensuse.org/issues/108091

### DIFF
--- a/lib/OpenQA/CacheService.pm
+++ b/lib/OpenQA/CacheService.pm
@@ -130,9 +130,16 @@ sub run {
     $ENV{MOJO_INACTIVITY_TIMEOUT} //= 300;
     $app->log->debug("Starting cache service: $0 @args");
     $app->defaults->{service_pid} = $$;
-
-    my $cmd_return_code = $app->start(@args);
-    return $app->exit_code // $cmd_return_code // 0;
+    my $e;
+    my $retry_interval = $ENV{OPENQA_CACHE_SERVICE_RETRY_INTERVAL} // 10;
+    do {
+        my $cmd_return_code = eval { $app->start(@args) };
+        chomp($e = $@);
+        return $app->exit_code // $cmd_return_code // 0 unless $e;
+        die $e unless $e =~ /Cannot assign requested address/;
+        $app->log->info("cache service failed with '$e', retrying after $retry_interval seconds");
+        sleep $retry_interval;
+    } while (1);
 }
 
 1;


nicksinger:
Cannot judge the perl code but the rest looks fine :)

lib/OpenQA/CacheService.pm
```diff
@@ -130,9 +130,16 @@ sub run {
...
-    my $cmd_return_code = $app->start(@args);
-    return $app->exit_code // $cmd_return_code // 0;
+    my $e;
+    my $cmd_return_code;
+    do {
+        eval { $cmd_return_code = $app->start(@args) };
+        chomp($e = $@);
+        return $app->exit_code // $cmd_return_code // 0 unless $e;
+        my $retry_interval = $ENV{OPENQA_CACHE_SERVICE_RETRY_INTERVAL} // 10;
+        $app->log->info("cache service failed with '$e', retrying after $retry_interval seconds");
```
Martchus, line 140: It'll actually only retry when `$e =~ /Cannot assign requested address/` so logging the message always is a bit confusing.

lib/OpenQA/CacheService.pm
```diff
@@ -130,9 +130,16 @@ sub run {
...
-
-    my $cmd_return_code = $app->start(@args);
-    return $app->exit_code // $cmd_return_code // 0;
+    my $e;
+    my $cmd_return_code;
+    do {
+        eval { $cmd_return_code = $app->start(@args) };
+        chomp($e = $@);
+        return $app->exit_code // $cmd_return_code // 0 unless $e;
+        my $retry_interval = $ENV{OPENQA_CACHE_SERVICE_RETRY_INTERVAL} // 10;
```
Martchus, line 139: I wouldn't define/initialize this variable again in every loop iteration.

lib/OpenQA/CacheService.pm
```diff
@@ -130,9 +130,16 @@ sub run {
     $ENV{MOJO_INACTIVITY_TIMEOUT} //= 300;
     $app->log->debug("Starting cache service: $0 @args");
     $app->defaults->{service_pid} = $$;
-
-    my $cmd_return_code = $app->start(@args);
-    return $app->exit_code // $cmd_return_code // 0;
+    my $e;
+    my $cmd_return_code;
+    do {
+        eval { $cmd_return_code = $app->start(@args) };
```
Martchus, line 136:
```suggestion
        my $cmd_return_code = eval { $cmd_return_code = $app->start(@args) };
```

Martchus (changes requested):

lib/OpenQA/CacheService.pm
kalikiana: Do we know what's expected with async responses here? [Mojolicious::startup](https://docs.mojolicious.org/Mojolicious#startup) doesn't seem to document that?

kraih:
I still believe that this issue should be resolved at the systemd level, and does not require code workarounds.

okurz:
> I still believe that this issue should be resolved at the systemd level, and does not require code workarounds.

I am open to suggestions how you want to handle that better in systemd.

lib/OpenQA/CacheService.pm
Martchus: I think `start` blocks for the whole time the service is running. We *likely* don't need to care about anything happening asynchronously.

lib/OpenQA/CacheService.pm
Martchus: If the error message is *not* matching `Cannot assign requested address` it should likely be logged as an error message.

okurz: Done both I think by calling die in the line above unless we have the specific message

lib/OpenQA/CacheService.pm
okurz: I don't see anything happing asynchronously here.

kraih: It depends on what command is executed by `start`. By default there are no async commands included though.

kalikiana: Should we be using [around_dispatch](https://docs.mojolicious.org/Mojolicious#around_dispatch) instead? That seems to be the "native" approach to exception handling

Martchus: I'm wondering whether invoking `start` a 2nd time will ensure the event loop is reset. If not, I suppose that would be a good idea to do to avoid events from the previous run being processed on the next `start` invocation.

Martchus:
> I am open to suggestions how you want to handle that better in systemd.

When I understand https://www.freedesktop.org/wiki/Software/systemd/NetworkTarget correctly, it should be enough to use `After=network.target` (and `Wants=`/`Requires=` should not be used, indeed).

Apparently this is also relying on support from the networking manager, in our case Wicked. So that could still be a source of error.

Btw, NGINX is using

```
After=network.target network-online.target nss-lookup.target
```

under Arch Linux which is kind of in-line with https://www.freedesktop.org/wiki/Software/systemd/NetworkTarget.

Under Tumbleweed, NGINX is using

```
After=network-online.target remote-fs.target nss-lookup.target
Wants=network-online.target
```

which has `Wants=` which is likely not necessary.

Maybe just go with
```
After=network.target network-online.target nss-lookup.target
```
or even just
```
After=network.target nss-lookup.target
```
?

lib/OpenQA/CacheService.pm
kraih: @kalikiana You might be confusing this with another exception handling issue. This one is during app server startup, when the server can't bind to the socket because the interfaces are not ready yet.

kraih: > I'm wondering whether invoking `start` a 2nd time will ensure the event loop is reset. If not, I suppose that would be a good idea to do to avoid events from the previous run being processed on the next `start` invocation.

Good point, i think this is currently undefined behaviour.

Martchus:
Ok, my last comment is actually just doing https://github.com/os-autoinst/openQA/pull/4577 which I've just approved.

okurz:
> Ok, my last comment is actually just doing #4577 which I've just approved.

You mean #4577 is is doing what you proposed already?

What this PR is doing on top is fixing the error message that can be seen in https://progress.opensuse.org/issues/108091#note-4 for openqa-worker-cacheservice.service

okurz:
As decided in the weekly unblock of SUSE QE Tools we decided that we do not want to solve this within the application code and the majority decided that we should rely on system service management frameworks to ensure the presence of network.

PR #4574, opened 2022-03-22
Prevent error about undefined value in next_previous route

okurz:
Related progress issue: https://progress.opensuse.org/issues/108662

### DIFF
--- a/lib/OpenQA/WebAPI/Controller/Test.pm
+++ b/lib/OpenQA/WebAPI/Controller/Test.pm
@@ -394,7 +394,7 @@ sub infopanel {
     $self->render('test/infopanel');
 }
 
-sub get_current_job ($self, $with_assets = 0) {
+sub _get_current_job ($self, $with_assets = 0) {
     return $self->reply->not_found unless defined $self->param('testid');
 
     my $job = $self->schema->resultset("Jobs")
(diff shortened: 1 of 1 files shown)


perlpunk (approved):

lib/OpenQA/WebAPI/Controller/Test.pm
```diff
@@ -428,7 +428,7 @@ sub _show {
 }
 
 sub job_next_previous_ajax ($self) {
-    my $main_job = $self->get_current_job;
+    return $self->reply->not_found unless my $main_job = $self->_get_current_job;
```
kalikiana, line 431: Would you mind extending our unit tests to cover this?

okurz: I consider this not necessary, it's just basic in-line safety checks which doesn't even touch coverage. Please also see https://github.com/os-autoinst/openQA/pull/4575 where I try to make the overall code simpler with exception handling so that we don't need these checks at all.

kalikiana: I don't follow. I'm asking if we can have a regression test since this issue was only seen in production.

Martchus (approved):

PR #4582, opened 2022-03-28
Add job comment about automatic retry (instead of server log message)

Martchus:
* and [Avoid possible warnings when using RETRY as number when it is not one](https://github.com/os-autoinst/openQA/commit/8ab29bbc0a476e8656680f11016bdc57de7638a8)
* Related ticket: https://progress.opensuse.org/issues/108971


lib/OpenQA/Schema/Result/Jobs.pm
```diff
@@ -1933,9 +1934,13 @@ sub handle_retry ($self) {
     return undef unless my $retry = $self->settings_hash->{RETRY};
     # strip any optional descriptions after a colon
     $retry =~ s/:.*//;
+    return 0 unless looks_like_number $retry;
     my $ancestors = $self->ancestors;
-    log_debug('handle_retry: Found retry job ' . $self->id . ", try '$ancestors' of up to '$retry' retries");
-    return $ancestors < $retry;
+    return 0 if $ancestors >= $retry;
+    my $system_user_id = $self->result_source->schema->resultset('Users')->system->id;
+    my $msg = "Restarting as RETRY set to $retry (and only restarted $ancestors times so far)";
```
kraih, line 1941: `Restarting as RETRY...` doesn't sound quite right to me. Maybe `Restarting because RETRY...`?

baierjan (approved):

PR #4585, opened 2022-03-29
WIP: Fix VLAN assignment on MM jobs created via "jobs post"

cfconrad:
Ticket: poo#109190

If a job was created using "jobs post" and `_PARALLEL_JOBS`, then
the method `Jobs::find_networks()` can retrieves the VLAN from
parent or child, but the current job do not have a entry in
`job_networks` db table.

### DIFF
--- a/lib/OpenQA/Schema/Result/Jobs.pm
+++ b/lib/OpenQA/Schema/Result/Jobs.pm
@@ -1559,7 +1559,10 @@ sub _asset_find ($name, $type, $parents) {
 
 sub allocate_network ($self, $name) {
     my $vlan = $self->_find_network($name);
-    return $vlan if $vlan;
+    if ($vlan) {
+        $self->networks->find_or_create({name => $name, vlan => $vlan});
+        return $vlan;
+    }
     #allocate new
     my @used_rs = $self->result_source->schema->resultset('JobNetworks')->search(
         {},


cfconrad:
> When using the syntax explained [here](https://open.qa/docs/#_spawning_single_new_jobs_jobs_post) this shouldn't be required because then `prepare_for_work`....

Thx for the hint, that is indeed very nice! I will check it out first.

cfconrad:
Hm I'm failing...
```bash
> openqa-cli api --host https://openqa.opensuse.org -X POST jobs TEST:0=first-job TEST:1=second-job _START_AFTER:1=0
400 Bad Request
{"error":"TEST field mandatory","error_status":400}
```

cfconrad:
I needed to update my openQA-Client package from devel repo, the missing patch was https://github.com/os-autoinst/openQA/pull/4535/commits/afc919a7182a0f56d50aee8ba9d2ba8cb0424f60 to get the https://github.com/os-autoinst/openQA/pull/4535 feature working!

Martchus:
We've already resolved the issue in a private chat. For the record, the issue was that `openqa-cli` was too old. (https://github.com/os-autoinst/openQA/pull/4535/commits/afc919a7182a0f56d50aee8ba9d2ba8cb0424f60 is needed, that's *not* in official Leap repos.)

lib/OpenQA/Schema/Result/Jobs.pm
```diff
@@ -1559,7 +1559,10 @@ sub _asset_find ($name, $type, $parents) {
 
 sub allocate_network ($self, $name) {
     my $vlan = $self->_find_network($name);
-    return $vlan if $vlan;
+    if ($vlan) {
+        $self->networks->find_or_create({name => $name, vlan => $vlan});
```
Akemiiii, line 1563: The contents in braces and brackets may be separated by a space.

okurz: I think it's possible to do that. But in general we have a tool called "perltidy" that is also running as part of the continuous integration test suite and ensures that we use a consistent style. So in short: As long as perltidy does not complain we can accept the code as is :)

cfconrad:
@Martchus as your solution how to trigger MM jobs correctly (mentioned here: https://github.com/os-autoinst/openQA/pull/4585#issuecomment-1081970278) is working, I'm not sure if we wanna add this way to fix the NETWORK assignment as well.

From my point of view, we should not use this fix, cause I'm not sure if we have other situations which might not work in the sam way. 
But maybe we should reject such trigger job some how or at least throw a warning?
WDYT?

Feel free to reopen, if you or others think it is worth to continue....

Martchus:
A warning would be nice. However, it makes no sense to place it in openqa-clone-job because it will no longer use the old way anyways. So it would only be shown if the user uses the API to spawn single jobs one-by-one. That's likely not worth the effort.

And yes, I think we don't need this change anymore. (Before merging it we should have tests and then it doesn't seem worth the effort considering the use-case is already covered anyways.)

PR #4589, opened 2022-04-01
Fix race condition during asset registration

Martchus:
* Avoid using DBIx's `find_or_create` function which is not atomic¹ and
  therefore not race-free which can lead to errors like:
  ```
  duplicate key value violates unique constraint \"assets_type_name\"\nDETAIL:  Key (type, name)=(iso, foo.iso) already exists.
  ```
* Use `INSERT INTO … ON CONFLICT DO NOTHING` instead (which unfortunately
  requires the use of raw SQL)
* See https://progress.opensuse.org/issues/109292 although this is not a
  new issue; the oldest scheduled product I found in production was from 3
  years ago

---

¹ https://metacpan.org/dist/DBIx-Class/view/lib/DBIx/Class/ResultSet.pm#find_or_create

### DIFF
--- a/lib/OpenQA/Schema/ResultSet/Assets.pm
+++ b/lib/OpenQA/Schema/ResultSet/Assets.pm
@@ -29,10 +29,10 @@ sub register ($self, $type, $name, $options = {}) {
         return undef;
     }
     my $schema = $self->result_source->schema;
-    my $sth = $schema->storage->dbh->prepare_cached(<<~'END_SQL');
+    my $sth = $schema->storage->dbh->prepare_cached(<<'END_SQL');
         INSERT INTO assets (type, name, t_created, t_updated)
                     VALUES (?,    ?,    now(),     now()    ) ON CONFLICT DO NOTHING
-        END_SQL
+END_SQL
     $schema->txn_do(
         sub {
             $sth->execute($type, $name);    # ensure asset exists
--- a/lib/OpenQA/Schema/ResultSet/Assets.pm
+++ b/lib/OpenQA/Schema/ResultSet/Assets.pm
@@ -28,9 +28,15 @@ sub register ($self, $type, $name, $options = {}) {
         log_warning "no file found for asset '$name' type '$type'";
         return undef;
     }
-    $self->result_source->schema->txn_do(
+    my $schema = $self->result_source->schema;
+    my $sth = $schema->storage->dbh->prepare_cached(<<~'END_SQL');
+        INSERT INTO assets (type, name, t_created, t_updated)
+                    VALUES (?,    ?,    now(),     now()    ) ON CONFLICT DO NOTHING
+        END_SQL
+    $schema->txn_do(
         sub {
-            my $asset = $self->find_or_create({type => $type, name => $name}, {key => 'assets_type_name'});
+            $sth->execute($type, $name);    # ensure asset exists
+            return undef unless my $asset = $self->find({type => $type, name => $name}, {key => 'assets_type_name'});
             $asset->refresh_size if $options->{refresh_size};
             if (my $created_by = $options->{created_by}) {
                 my $scope = $options->{scope} // 'public';


Martchus:
By the way, just making this transactional doesn't work (as previously attempted, see https://progress.opensuse.org/issues/35749 and https://github.com/os-autoinst/openQA/commit/6508af70ed0d98e359e7e75dd233b77332d655bd) is not sufficient. I suppose that's because within the transaction we'll still see changes happening elsewhere unless we would also set the isolation level accordingly (e.g. `SET TRANSACTION ISOLATION LEVEL REPEATABLE READ READ ONLY DEFERRABLE;`). However, even then we'd likely run into a conflict when committing the transaction. So I suppose the best solution is just to use `ON CONFLICT DO NOTHING`.

lib/OpenQA/Schema/ResultSet/Assets.pm
```diff
@@ -21,16 +21,20 @@ sub status_cache_file {
 }
 
 # called when uploading an asset or finding one in scanning
+my $INSERT = 'INSERT INTO assets (type, name, t_created, t_updated) VALUES (?, ?, now(), now()) ON CONFLICT DO NOTHING';
```
kraih, line 24: You don't really gain anything by declaring the query outside the function. Only makes it harder to follow the code imo.

lib/OpenQA/Schema/ResultSet/Assets.pm
```diff
@@ -21,16 +21,20 @@ sub status_cache_file {
...
 sub register ($self, $type, $name, $options = {}) {
     unless ($name) { log_warning 'attempt to register asset with empty name'; return undef }
     unless (grep /^$type$/, TYPES) { log_warning "asset type '$type' invalid"; return undef }
     if (!$options->{missing_ok} && !locate_asset $type, $name, mustexist => 1) {
         log_warning "no file found for asset '$name' type '$type'";
         return undef;
     }
-    $self->result_source->schema->txn_do(
+    my $schema = $self->result_source->schema;
+    my $sth = ($self->{_insert_sth} //= $schema->storage->dbh->prepare($INSERT));
```
kraih, line 33: Maybe you want to call `->prepare_cached` here instead and let DBI deal with the caching?

lib/OpenQA/Schema/ResultSet/Assets.pm
Martchus: I gain that the line doesn't need to be wrapped but I suppose having it inline is generally better, indeed.

lib/OpenQA/Schema/ResultSet/Assets.pm
```diff
@@ -28,9 +28,17 @@ sub register ($self, $type, $name, $options = {}) {
         log_warning "no file found for asset '$name' type '$type'";
         return undef;
     }
-    $self->result_source->schema->txn_do(
+    my $schema = $self->result_source->schema;
+    my $sth = $schema->storage->dbh->prepare_cached(
+        <<'END_SQL'
+        INSERT INTO assets (type, name, t_created, t_updated)
+                    VALUES (?,    ?,    now(),     now()    ) ON CONFLICT DO NOTHING
+END_SQL
```
kraih, line 36: If we can depend on Perl 5.16 this could use an indented heredoc and look much nicer.
```perl
    my $sth = $schema->storage->dbh->prepare_cached(<<~'END_SQL');
        INSERT INTO assets (type, name, t_created, t_updated)
                    VALUES (?,    ?,    now(),     now()    ) ON CONFLICT DO NOTHING
    END_SQL
```

lib/OpenQA/Schema/ResultSet/Assets.pm
```diff
@@ -28,9 +28,17 @@ sub register ($self, $type, $name, $options = {}) {
         log_warning "no file found for asset '$name' type '$type'";
         return undef;
     }
-    $self->result_source->schema->txn_do(
+    my $schema = $self->result_source->schema;
+    my $sth = $schema->storage->dbh->prepare_cached(
+        <<'END_SQL'
```
kraih, line 33: I think you can still do `my $sth = $schema->storage->dbh->prepare_cached(<<'END_SQL');` at least.

kraih (approved):

okurz (approved):

PR #4590, opened 2022-04-01
Ensure consistent cancellation of chained dependent jobs

Martchus:
* Cancel regularly and directly chained dependencies consistently only down
  the chain
    * Keep the behavior for parallel dependencies as-is (so they are still
      canceled up the hierarchy except PARALLEL_CANCEL_WHOLE_CLUSTER is set
      to a falsy value)
    * Cancellation of the entire chain at once is still possible by
      canceling from the root (unless there are multiple roots)
    * The cancellation up the chain was already prevented before in certain
      cases but whether it worked depended on the internal order in which
      dependencies have been traversed
    * Be more in-line with user expectations, at least judging by
      https://progress.opensuse.org/issues/108476 it expected that
      cancellation only affects jobs down the chain
* Track whether the job cluster is computed for cancellation on all levels
  of nested functions calls to ensure a consistent behavior regardless of
  the internal order of dependency traversal

---

~~No tests yet because I hope failing unit tests will give me a hint were to make adjustments.~~

### DIFF
--- a/lib/OpenQA/Schema/Result/Jobs.pm
+++ b/lib/OpenQA/Schema/Result/Jobs.pm
@@ -723,12 +723,13 @@ sub cluster_jobs ($self, @args) {
     my $skip_children = $args{skip_children};
     my $skip_parents = $args{skip_parents};
     my $no_directly_chained_parent = $args{no_directly_chained_parent};
+    my $cancelmode = $args{cancelmode};
 
     # handle re-visiting job
     if (defined $job) {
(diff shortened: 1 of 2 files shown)


okurz (approved):
everything ok maybe except the word "depended" which I am not so sure about. Shouldn't it be "dependant"?

Martchus:
According to the dictionary "dependent" is the most common form: https://www.oxfordlearnersdictionaries.com/definition/american_english/dependent_1

kalikiana (approved):

PR #4592, opened 2022-04-06
Allow adding scheduling settings for informal purposes via `__`-prefix

Martchus:
Those settings will *not* be added scheduled openQA jobs and can be used to
store additional information about the scheduled product itsef, e.g. an URL
to SMELT.

As a bonus I've added a small change to render those setting values as link if
they are HTTP(S)-URLs.

See https://progress.opensuse.org/issues/109554

### DIFF
--- a/assets/javascripts/openqa.js
+++ b/assets/javascripts/openqa.js
@@ -534,3 +534,15 @@ function renderComments(row) {
   }
   return html;
 }
+
+function renderHttpUrlAsLink(value) {
+  const text = document.createTextNode(value);
+  if (!value.match(/^https?:\/\//)) {
+    return text;
+  }
+  const link = document.createElement('a');
+  link.href = value;
+  link.target = 'blank';
+  link.appendChild(text);
+  return link;
+}
--- a/lib/OpenQA/Schema/Result/ScheduledProducts.pm
+++ b/lib/OpenQA/Schema/Result/ScheduledProducts.pm
@@ -4,7 +4,7 @@
 package OpenQA::Schema::Result::ScheduledProducts;
 
 
-use Mojo::Base 'DBIx::Class::Core';
+use Mojo::Base 'DBIx::Class::Core', -signatures;
 
 use DBIx::Class::Timestamps 'now';
 use File::Basename;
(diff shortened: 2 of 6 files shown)


lib/OpenQA/Schema/Result/ScheduledProducts.pm
```diff
@@ -220,6 +220,12 @@ sub _schedule_iso {
         };
     }
 
+    # do not pass any arguments starting with "__" to jobs so these settings can be used to store additional information
+    # about the scheduled product itself
+    for my $arg (keys %$args) {
+        delete $args->{$arg} if index($arg, '__') == 0;
+    }
+
```
okurz, line 228: As the surrounding method is rather big already I suggest to extract that logic into a function with a good name and call it here.

okurz (changes requested):

assets/javascripts/openqa.js
```diff
@@ -534,3 +534,15 @@ function renderComments(row) {
...
+  const text = document.createTextNode(value);
+  if (!value.match(/^https?:\/\//)) {
+    return text;
+  }
+  const link = document.createElement('a');
+  link.href = value;
+  link.target = 'blank';
+  link.appendChild(text);
+  return link;
+}
```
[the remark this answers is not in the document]
Martchus: No, it is no duplicate. The job details settings are rendered in Perl code and this is rendered client-side. Apart from the fact that client-side rendering is the modern approach, it also needs to be client-side because the rendering is needed for a modal dialog we show dynamically when a button in a DataTables row is clicked. I also checked whether we already have such a helper in our existing JavaScript code.

okurz (commented):
Please add documentation for this behaviour, rest looks fine.

kalikiana (approved):
Looks sensible. I agree this should be documented somewhere

baierjan (approved):

lib/OpenQA/Schema/Result/ScheduledProducts.pm
```diff
@@ -180,6 +180,12 @@ sub schedule_iso {
 # make sure that the DISTRI is lowercase
 sub _distri_key { lc(shift->{DISTRI}) }
 
+sub _delete_prefixed_args_storing_info_about_product_itself ($args) {
```
perlpunk, line 183: Signatures have to be enabled

lib/OpenQA/Schema/Result/ScheduledProducts.pm
```diff
@@ -180,6 +180,12 @@ sub schedule_iso {
...
+    for my $arg (keys %$args) {
+        delete $args->{$arg} if index($arg, '__') == 0;
```
perlpunk, line 185: `$arg =~ m/^__/` should be faster than `index()` because the `^` in the regex will only search at the beginning, and `index()` will search the whole string.
alternatively `substr($arg, 0, 2) eq '__'`

okurz:
Please add documentation for this behaviour, rest looks fine.

okurz (approved):

PR #4597, opened 2022-04-08
Avoid corrupted assets on concurrent uploads of same asset

Martchus:
* Write temporary files to a job-specific location so multiple uploads
  don't interfere with each other
* The move in the end should be atomic as it is likely a move within the
  same file system (if not I suppose file corruption can still happen)
* See https://progress.opensuse.org/issues/109319


kalikiana (approved):
image

baierjan (approved):

PR #4599, opened 2022-04-08
Render multiple URLs in scheduled products table correctly

Martchus:
Considering https://github.com/openSUSE/qem-bot/pull/16 and existing code
the same setting value might contain multiple URLs (separated by colons).

This change will render them as individual links.

See https://progress.opensuse.org/issues/109512

### DIFF
--- a/assets/javascripts/openqa.js
+++ b/assets/javascripts/openqa.js
@@ -536,13 +536,17 @@ function renderComments(row) {
 }
 
 function renderHttpUrlAsLink(value) {
-  const text = document.createTextNode(value);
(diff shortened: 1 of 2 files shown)


okurz:
Observed the problem on https://openqa.suse.de/admin/productlog?id=894628 , hotpatched OSD with

```
curl -sS https://patch-diff.githubusercontent.com/raw/os-autoinst/openQA/pull/4599.patch | patch -p1 --directory=/usr/share/openqa
systemctl restart openqa-webui
```

and confirmed it working again on
https://openqa.suse.de/admin/productlog?id=894628

PR #4600, opened 2022-04-09
Use signatures in OpenQA::WebAPI::Plugin::REST

okurz:
After
* #4419

### DIFF
--- a/lib/OpenQA/WebAPI/Plugin/REST.pm
+++ b/lib/OpenQA/WebAPI/Plugin/REST.pm
@@ -13,13 +13,7 @@ sub register ($self, $app, $config) {
         action_link => sub ($c, $method, $content, @args) {
             my $url = $content;
             return '' unless $c->is_operator;
-            # Content
-            unless (ref $args[-1] eq 'CODE') {
-                $url = shift @args;
-                push @args, $content;
-            }
             croak "url is not a url" unless blessed $url && $url->isa('Mojo::URL');
-
             return $c->tag('a', href => $url, 'data-method' => $method, @args);
         });
 
--- a/lib/OpenQA/WebAPI/Plugin/REST.pm
+++ b/lib/OpenQA/WebAPI/Plugin/REST.pm
@@ -2,49 +2,33 @@
 # SPDX-License-Identifier: GPL-2.0-or-later
 
 package OpenQA::WebAPI::Plugin::REST;
-use Mojo::Base 'Mojolicious::Plugin';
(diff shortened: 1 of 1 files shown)


lib/OpenQA/WebAPI/Plugin/REST.pm
```diff
@@ -2,49 +2,33 @@
...
-                Carp::croak "url is not a url"
-                  unless Scalar::Util::blessed $url && $url->isa('Mojo::URL');
-
-                return $self->tag('a', href => $url, 'data-method' => $method, @_);
-            }
-            else {
-                return '';
+            return '' unless $c->is_operator;
+            # Content
+            unless (ref $_[-1] eq 'CODE') {
```
perlpunk, line 17: there is still a reference to `@_`

perlpunk:
```suggestion
            unless (ref $args[-1] eq 'CODE') {
```

lib/OpenQA/WebAPI/Plugin/REST.pm
```diff
@@ -2,49 +2,33 @@
...
-
-                return $self->tag('a', href => $url, 'data-method' => $method, @_);
-            }
-            else {
-                return '';
+            return '' unless $c->is_operator;
+            # Content
+            unless (ref $args[-1] eq 'CODE') {
+                $url = shift @args;
+                push @args, $content;
```
okurz, line 19: this code seems to be uncovered. Anyone has an idea what it could be useful for? Or should we just remove it?

Martchus: Not sure. So Git blame doesn't reveal anything?

okurz: Well, @aaannz has introduced it, we never changed it and it's not called in tests. What behavior could this be for?

aaannz: You mean @ancorgs ( commit https://github.com/os-autoinst/openQA/commit/34d51fb0c3c20d6a6e76bf13f5ae7e6ae52bd919 ). I have no idea what it is of and if it is still valid.

okurz: Well, I meant the refactoring in 04c265d57 but looks like you found a predecessor. I guess I will just remove that code.

kalikiana (approved):

Martchus (approved):

PR #4602, opened 2022-04-11
Add systemd unit to trigger continous deployment from devel:openQA

Martchus:
See https://progress.opensuse.org/issues/105379

---

Deployed on openqaworker7, see https://progress.opensuse.org/issues/105379#note-8

### DIFF
--- a/dist/rpm/openQA.spec
+++ b/dist/rpm/openQA.spec
@@ -257,12 +257,25 @@ Covering both openQA and also os-autoinst test engine.
 %package auto-update
 Summary:        Automatically upgrade and reboot the system when required
 Group:          Development/Tools/Other
+Requires:       %{name}-common
 Requires:       curl
 Requires:       rebootmgr
 
 %description auto-update
 Use this package to install and enable a systemd service for nightly upgrading
 and rebooting the system if devel:openQA packages are stable.
+
+%package continuous-update
+Summary:        Continuously update packages from devel:openQA
+Group:          Development/Tools/Other
+Requires:       %{name}-common
+Requires:       curl
+
+%description continuous-update
+Use this package to install and enable a systemd service for continuously
+upgrading the system if devel:openQA packages are stable and contain updates. It
+is complementary to auto-update which also reboots the system and does updates
+regardless of whether devel:openQA contains updates.
 
 %prep
 %setup -q -a1
@@ -405,6 +418,9 @@ fi
 %pre auto-update
 %service_add_pre openqa-auto-update.timer
 
+%pre continuous-update
+%service_add_pre openqa-continuous-update.timer
+
 %post
 %tmpfiles_create %{_tmpfilesdir}/openqa-webui.conf
 # install empty log file
@@ -438,6 +454,9 @@ fi
 %post auto-update
 %service_add_post openqa-auto-update.timer
 
+%post continuous-update
+%service_add_post openqa-continuous-update.timer
+
 %preun
 %service_del_preun %{openqa_services}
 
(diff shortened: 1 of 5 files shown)


kalikiana (approved):
Nice. Quiter clean

script/openqa-continuous-update
```diff
@@ -0,0 +1,29 @@
+#!/bin/bash
+set -e
+set -o pipefail
+
+usage() {
+    cat << EOF
+Usage: openqa-continous-update
+Trigger automatic system upgrade and reboot if devel:openQA packages are stable
```
okurz, line 8: copy-paste-fail

systemd/opensuse/openqa-continuous-update.timer
```diff
@@ -0,0 +1,9 @@
+[Unit]
+Description=Trigger of continuous openQA deployments, see https://progress.opensuse.org/issues/105379
+
+[Timer]
+OnBootSec=10min
```
okurz, line 5: https://progress.opensuse.org/issues/105379#Suggestions says 5 minutes. I would be ok to accept 5 minutes as continuous, but 10 minutes? I don't know ;) But seriously, I would prefer 5 minutes. Any good reason against?

okurz (changes requested):

systemd/opensuse/openqa-continuous-update.timer
Martchus: No, I have nothing against 5 minutes. It'll usually just exiting very quickly anyways. I suppose if an update takes longer than 5 minutes the systemd timer will be smart enough to handle it.

okurz: Zypper would fail when trying to get the lock

script/openqa-continuous-update
```diff
@@ -0,0 +1,29 @@
...
+while true; do
+  case "$1" in
+    -h | --help ) usage 0; shift ;;
+    -- ) shift; break ;;
+    * ) break ;;
+  esac
+done
+
+repo_name=${DEVEL_OPENQA_REPOSITORY:-devel_openQA}
+if zypper -n ref -r "$repo_name" | grep -q 'is up to date' && "$(dirname "${BASH_SOURCE[0]}")"/openqa-check-devel-repo; then
```
okurz, line 27: Why do you include `openqa-check-devel-repo` in here as well? IMHO if zypper says it has updated packages, which is from download.opensuse.org or any configured mirror, then we should not care querying the API of OBS. Can we just leave that out?

dist/rpm/openQA.spec
```diff
@@ -257,13 +257,24 @@ Covering both openQA and also os-autoinst test engine.
...
 and rebooting the system if devel:openQA packages are stable.
 
+%package continuous-update
+Summary:        Continuously update packages from devel:openQA
+Group:          Development/Tools/Other
+Requires:       %{name}-common
+Requires:       curl
+
+%description continuous-update
+Use this package to install and enable a systemd service for continuously upgrading
```
okurz, line 275: We should describe here the relation to "auto-update".

As auto-update does a `zypper dup` without repo restrictions and then even triggers a full system reboot they are complementary, not alternatives.

Please also extend the documentation. See
https://github.com/os-autoinst/openQA/blob/master/docs/Installing.asciidoc#automatic-system-upgrades-and-reboots-of-openqa-hosts= about the according auto-update section where you could extend.

okurz (changes requested):

dist/rpm/openQA.spec
Martchus: > they are complementary, not alternatives

Yes. I also have both mechanisms still in place on openqaworker7 and also don't see it as part of the initial ticket to remove the existing auto update.

script/openqa-continuous-update
okurz: hm, ok. It won't be a promise that the packages that are actually downloadable are ok but I guess that makes sense to be aligned with our current development approach.

okurz (approved):

PR #4604, opened 2022-04-12
Avoid running into limit when creating job modules

Martchus:
* Migrate job_modules ID to bigint
* Works also if those columns are already bigint (so deployment on manually
  migrated instances should not be a problem)
* See https://progress.opensuse.org/issues/109849


okurz (approved):
Looks straightforward and clean. The migration on OSD took around 5 minutes with 27M job modules so I guess this should be ok for everyone else as well as also upgrading packages and restarting services can be in a similar time range.

perlpunk (approved):

PR #4610, opened 2022-04-14
WIP: Stack parallel children in test results overview

Martchus:
* See https://progress.opensuse.org/issues/109809

---

This kind of works but since this approach is only stacking on row-level
there are problems with multiple archs I still need to figure out.


okurz:
> Expanded (after clicking on the "stack icon"): screenshot_20220414_175625

I suggest to indent the circle icons below the stacked parent job to make it clear which jobs are part of the expanded stack.

Martchus:
It works for multiple archs as well with one caveat: Jobs can be folded even though they are not parallel children. That happens in the case when the job for the same test but a different architecture is a parallel child. For instance, here I've artificially created an aarch64 job with no dependencies at all which has the same test name as an x86_64 job which is a parallel child:

screenshot_20220420_180929

The aarch64 job is not shown by default but can be expanded/collapsed by clicking on the x86_64 parallel parent.

That's not ideal but I'm not sure how to improve this. Maybe by simply giving up on the feature if that case is detected as it would at least be less confusing? Even if there are parallel parents in all archs, it would still be a bit confusing that one has to click on a specific arch's job to expand children of *all* archs. So maybe I add an extra expand/collapse icon just after the test name? Then the expanding/collapsing mechanism would be independent of the concrete jobs. Besides, one can then still navigate to a parallel parent by just clicking its dot/stack-icon (so far the normal link behavior is replaced with expanding/collapsing and one has to middle-click).

Martchus:
I now have it working with one expand/collapse button per row. However, that doesn't solve all problems:

screenshot_20220421_164813

Here it breaks because a parallel child is part of a different flavor table. I suppose it would also break if a parallel parent is for some reason not part of the overview page (e.g. part of a different build). In these cases it would be the best to simply not apply the stacking.

A few visual improvements would likely be nice as well:
* Maybe I should also keep the stacking icon on the specific jobs in the collapsed state using a color representing all stacked jobs best (e.g. the "worst" result wins).
* The indentation @okurz was asking for (only affecting the test name to keep the dots nicely aligned).

Martchus:
Now the example looks like this:
screenshot_20220422_175047

I suppose that's ok although the accumulated state/result is still missing. I also need to add some tests.

okurz (changes requested):
Seems I never sent a pending comment, here you go

Martchus:
That's how it looks like with real data:

grafik

Considering the colored branch icons I suppose I don't need to compute an accumulated state for the dots in the collapsed state.

okurz:
I suggest to introduce the new feature by keeping the expanded state as default and allowing to collapse by a button click

Martchus:
The last two commits implement @okurz 's last comment.

Martchus:
All staging VMs are down and they're likely completely outdated setups as nobody has used them for ages.

kalikiana:
@mergify rebase

okurz:
I would really like and is certainly technically possible is to have webUI instances available for trying out for any pull request.

Martchus:
@okurz I installed the package (directly from the CI package check repo) on staging-1 which I also updated by the way. You can see the feature in action on http://openqa-staging-1.qa.suse.de/tests/overview?distri=sle&version=15-SP4&build=138.1&groupid=262 but feel free to check some more pages.

This is with OSD production data which I have imported as `openqa-osd` on *thincsus*.suse.de which contains common databases for staging-1/2. (I also imported the o3 database as `openqa-o3` as I accidentally imported the wrong dump first.)

okurz:
can you show all jobs by clicking on the top right double-arrow as well? Like simply the opposite of the "hide all" functionality

okurz (approved):

Martchus:
> can you show all jobs by clicking on the top right double-arrow as well?

I could make it do that if all rows are collapsed or I simply add 2nd button. I refrained from implementing it because it wasn't stated as ACs and I suppose I need to add relatively many lines of code for this relatively small addition.

okurz:
> > can you show all jobs by clicking on the top right double-arrow as well?
> 
> I could make it do that if all rows are collapsed or I simply add 2nd button. I refrained from implementing it because it wasn't stated as ACs and I suppose I need to add relatively many lines of code for this relatively small addition.

ok. Leave it out for now. We will see what user feedback brings.

kalikiana (approved):

PR #4612, opened 2022-04-17
Use signatures in Schema::Result::Needles

okurz:


lib/OpenQA/Schema/Result/Needles.pm
```diff
@@ -75,26 +75,17 @@ __PACKAGE__->belongs_to(
...
 }
 
-sub update_needle_cache {
-    my ($needle_cache) = @_;
-
-    for my $needle (values %$needle_cache) {
-        $needle->update;
-    }
-}
+sub update_needle_cache ($needle_cache) { $_->update for (%$needle_cache) }
```
Martchus, line 83:
```suggestion
sub update_needle_cache ($needle_cache) { $_->update for values %$needle_cache }
```

lib/OpenQA/Schema/Result/Needles.pm
```diff
@@ -203,15 +189,9 @@ sub remove {
...
-    my ($self) = @_;
-
-    return $self->file_present(-e $self->path ? 1 : 0);
-}
-
-sub to_json {
-    my ($self, $controller) = @_;
+sub check_file ($self) { $self->file_present(-e $self->path ? 1 : 0) }
 
+sub to_json ($self, $controller) {
```
Martchus, line 194: Looks like `$controller` isn't used. Maybe it makes more sense to removed (removing it from the call site as well).

perlpunk: Yeah, this looks like an accident. It is never called with a second argument, so just remove it.

okurz: done

lib/OpenQA/Schema/Result/Needles.pm
okurz: done

perlpunk (approved):

Martchus (approved):

PR #4613, opened 2022-04-20
Use OpenQA::Test::PatchDeparse from os-autoinst-common

perlpunk:


Martchus (approved):

okurz (approved):

perlpunk:
I restricted tidy to t/, lib/ and script/

.tidyallrc
```diff
@@ -2,5 +2,5 @@
 ; Notice that this does include also non-Perl files in script/ so it is
 ; suggested to use tools/tidy to cover all appropriate files. "tidyall -a"
 ; will try to cover too much and fail on the non-Perl files
-select = **/*.{pl,pm,t} script/*
+select = {lib,t}/**/*.{pl,pm,t} script/*
```
okurz, line 5: This would exclude tools/ at least. I suggest to use

```suggestion
ignore = external/
```

see https://metacpan.org/pod/tidyall

okurz (changes requested):

.tidyallrc
perlpunk: Currently `./tools/tidy -l` does not list anything in tools, but ok.

But I'm a fan of explicitly including things. I hate that I have to use `ignore` in os-autoinst. So I will include `tools` explicitly.

okurz: Well, I prefer to include everything and only exclude so that any new files or directories are not silently ignored by mistake. Having too many files included produces quite explicit errors.

perlpunk: So one against one. Any other opinions?

perlpunk:
Regarding the build-docs failure https://app.circleci.com/pipelines/github/os-autoinst/openQA/9420/workflows/fe97dc24-9382-4abf-b9f9-e1adfdc4a66b/jobs/88871 I created https://progress.opensuse.org/issues/110181

.tidyallrc
perlpunk: How many new directories in the root directory containing perl files are we expecting in the future?

okurz: That's the point, I don't know. To be safe I suggest to include all and specify the exceptions

perlpunk: There are some standard directories where perl files live in a project, and then maybe some custom ones. This config file can serve as an easy way to document in which directories we expect perl files that we want to tidy.
I don't get the argument of "being safe", especially when talking about perltidy.

perlpunk: Just convince someone else of the team to vote for the "ignore" option and I'll do it =)

okurz (approved):

perlpunk:
I will merge this now because the webui-docker-compose failed because of a zypper timeout (ran longer than 1 hour!)
Don't want this to get an endless PR

PR #4616, opened 2022-04-21
Fix documentation

perlpunk:
* Use regular asciidoctor-pdf release
* Fix doc headings and numbers

Issue: https://progress.opensuse.org/issues/110181

### DIFF
--- a/docs/Installing.asciidoc
+++ b/docs/Installing.asciidoc
@@ -1036,7 +1036,7 @@ updates and if it does it will upgrade the whole system. This approach is
 independent of `openQA-auto-update` but can be used complementary. The
 configuration is analogous to `openQA-auto-update`.
 
-==== Migrating from older databases
+== Migrating from older databases
 
 For older versions of openQA, you can migrate from SQLite to PostgreSQL
 according to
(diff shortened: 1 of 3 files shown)


baierjan (approved):

baierjan:
I will leave here the important part from the release notes:
> On the compliance front, Asciidoctor PDF 2 will require Ruby 2.7 and JRuby 9.2 at a minimum

docs/Installing.asciidoc
```diff
@@ -1036,15 +1036,15 @@ updates and if it does it will upgrade the whole system. This approach is
 independent of `openQA-auto-update` but can be used complementary. The
 configuration is analogous to `openQA-auto-update`.
 
-==== Migrating from older databases
+=== Migrating from older databases
```
okurz, line 1039: Level 3 is still not correct as we are still below "Automatic system upgrades and reboots of openQA hosts". Just make it level 2

```suggestion
== Migrating from older databases
```

same further below

okurz (changes requested):

okurz (approved):

Martchus (approved):

PR #4618, opened 2022-04-22
Fix apparmor for job hooks

perlpunk:
Apparmor still complained about python3


Martchus (approved):
I suppose now it is in the right section but I'm also not an apparmor expert :-)

okurz (approved):
I assume you didn't test it. It is possible to use apparmor tools to run services in complain mode and generate profile updates from but let's try this way.

PR #4628, opened 2022-04-28
Deal with cleaned up vars.json on investigation tab

perlpunk:
If vars.json doesn't exist (any more), don't try to open it and leave
an error in the log.

Issue: https://progress.opensuse.org/issues/110392

### DIFF
--- a/lib/OpenQA/Schema/Result/Jobs.pm
+++ b/lib/OpenQA/Schema/Result/Jobs.pm
@@ -1866,13 +1866,17 @@ sub investigate ($self, %args) {
         next unless $prev->result =~ /(?:passed|softfailed)/;
         $inv{last_good} = {type => 'link', link => '/tests/' . $prev->id, text => $prev->id};
         last unless $prev->result_dir;
+        my ($prev_file, $self_file) = map {
+            eval { Mojo::File->new($_->result_dir(), 'vars.json')->slurp }
+              // undef
+        } ($prev, $self);
+        $inv{diff_packages_to_last_good} = $self->packages_diff($prev, $ignore) // 'Diff of packages not available';
+        last unless $prev_file;
         # just ignore any problems on generating the diff with eval, e.g.
         # files missing. This is a best-effort approach.
-        my @files = map { Mojo::File->new($_->result_dir(), 'vars.json')->slurp } ($prev, $self);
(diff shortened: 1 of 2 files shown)


lib/OpenQA/Schema/Result/Jobs.pm
```diff
@@ -1866,26 +1866,31 @@ sub investigate ($self, %args) {
...
-            $dir = needledir($self->DISTRI, $self->VERSION);
-            my $refspec_needles_range = "$before->{NEEDLES_GIT_HASH}..$after->{NEEDLES_GIT_HASH}";
-            $inv{needles_log} = $self->git_log_diff($dir, $refspec_needles_range);
-            $inv{needles_log} ||= 'No needle changes recorded, test regression due to needles unlikely';
-            $inv{needles_diff_stat} = $self->git_diff($dir, $refspec_needles_range) if $inv{needles_log};
+        my ($prev_file, $self_file) = map {
+            eval { Mojo::File->new($_->result_dir(), 'vars.json')->slurp }
+              // undef
+        } ($prev, $self);
+        if ($prev_file) {
```
Martchus, line 1873: `last unless $prev_file;` would avoid having to indent all the code for this case.

Martchus (approved):

lib/OpenQA/Schema/Result/Jobs.pm
okurz: Also I think that at least the line `$inv{diff_packages_to_last_good} = $self->packages_diff($prev, $ignore) // 'Diff of packages not available';` is not relying on being able to read the files. I suggest to move that line up and then use

```suggestion
        last unless $prev_file;
```

okurz (changes requested):

lib/OpenQA/Schema/Result/Jobs.pm
perlpunk: done

okurz (approved):

kalikiana:
FTR [OBS: openQA - openSUSE_Leap_15.3/x86_64](https://build.opensuse.org/package/show/devel:openQA:GitHub:os-autoinst:openQA:PR-4628/openQA) failed in `t/03-auth.t` and I triggered a rebuild.

Martchus (approved):

PR #4636, opened 2022-05-03
cache service: Emit download rate only if known

Martchus:
If no downloads have been done yet the download rate is unknown. We should
not return 0 in that case as it leads to false alerts.

See https://progress.opensuse.org/issues/110497


perlpunk (approved):

okurz (approved):

kraih (approved):

Martchus:
The change is covered by `t/25-cache-service.t` but for some reason the coverage of the forked process is not tracked (although there's a `Devel::Cover::report() if Devel::Cover->can('report');`).

okurz:
yeah, it's ok

PR #4643, opened 2022-05-06
Install rsync to worker container image

baierjan:
To be able to use cache service, one need to have rsync client in the
worker image. As it is only a recommended package, we need to install it
explicitly.

Related issue: https://progress.opensuse.org/issues/110524


okurz (approved):
Hm, the thing is, not everybody needs it but it's not really a big requirement. I guess this should be ok

Martchus (approved):

PR #4646, opened 2022-05-11
Prevent job_done_hooks not knowing that jobs will be retried anyway

okurz:
A common use case for job_done_hooks is to handle failed jobs. However
there can be that case that we already sufficiently handle any such
failures by retrying test cases using the openQA retry feature. So if
test maintainers signal the intent that the right approach to handle
failed jobs is to retry we should ensure that we trigger a retry and
hence put a comment on an openQA job to provide a consistent state
before we trigger any asynchronous job_done_hooks which can evaluate
such information, e.g. as
https://github.com/os-autoinst/scripts/blob/master/openqa-investigate
does to detect if an issue is "known" or "unknown" based on the presence
of a comment.

Related progress issue: https://progress.opensuse.org/issues/110911

### DIFF
--- a/lib/OpenQA/Schema/Result/Jobs.pm
+++ b/lib/OpenQA/Schema/Result/Jobs.pm
@@ -2017,6 +2017,8 @@ sub done ($self, %args) {
         $new_val{reason} = 'no test modules scheduled/uploaded';
     }
     $self->update(\%new_val);
+    $self->unblock;
+    $self->auto_duplicate if $restart || (!$self->is_ok && $self->handle_retry);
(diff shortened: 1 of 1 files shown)


okurz:
I triggered a subset of tests locally and none failed. So maybe we actually benefit from changing the order of events slightly without breaking other workflows. Let's await full CI test results.

okurz:
Hm, really nothing changed. Anyone has a good suggestion what test to extend?

perlpunk (approved):

Martchus (approved):

PR #4650, opened 2022-05-12
Add Helm chart for openQA

baierjan:
Create a simple Helm chart to deploy openQA worker inside Kubernetes.
The chart is created as a sub-chart of openQA chart to allow for more
customization.

Reference: https://progress.opensuse.org/issues/110524


container/helm/Chart.yaml
```diff
@@ -0,0 +1,5 @@
+apiVersion: v2
+name: openQA
+description: A Helm chart for the openQA in Kubernetes
```
okurz, line 3:
```suggestion
description: A Helm chart for openQA in Kubernetes
```

container/helm/README.md
```diff
@@ -0,0 +1,30 @@
...
+
+Prerequisites:
+1. A Kubernetes cluster (for example _k3s_ will do fine)
+2. Installed and configured Helm
+
+For more information please consult corresponding documentation for [k3s](https://rancher.com/docs/k3s/latest/en/) or [Helm](https://helm.sh/docs).
+
+The charts consists of two separate sub-charts: _worker_ and _webui_ (TBD). To install the chart simply execute `helm install openqa .` from this directory. To uninstall and start over, type `helm uninstall openqa`.
+
+It might be necessary to customize the charts by overriding some of the variables inside _values.yaml_ to suite your needs.
```
okurz, line 11:
```suggestion
It might be necessary to customize the charts by overriding some of the variables inside _values.yaml_ to suit your needs.
```

okurz (commented):
looks nice. Any good approach to testing that as part of CI?

baierjan:
> looks nice. Any good approach to testing that as part of CI?

For start we might at least use linting with `helm lint`. But there is a possibility to run Kubernetes inside docker, so we can probably run some tests inside a GitHub actions. I will look at it.

baierjan:
I was also able to add a chart for the webui. In the current form, it depends on https://github.com/os-autoinst/openQA/pull/4653 and there is yet no easy way ho to provide assets/tests/needles, but I do have some ideas though.

okurz:
> there is yet no easy way ho to provide assets/tests/needles, but I do have some ideas though.

That's completely fine because the tests running on workers can just checkout tests (+needles) from git anyway, right?

baierjan:
> That's completely fine because the tests running on workers can just checkout tests (+needles) from git anyway, right?

That would probably be one option (and the same git can be actually checked-out in the webui). I already have a small TODO reminder to add an init container to do exactly that. The second option would be the cache service, which just need a small container with rsyncd service in the webui pod.

container/helm/README.md
```diff
@@ -0,0 +1,44 @@
+### Helm Chart for openQA inside Kubernetes
+
+Prerequisites:
+1. A Kubernetes cluster (for example _k3s_ will do fine)
+2. Installed and configured Helm
+
+For more information please consult corresponding documentation for [k3s](https://rancher.com/docs/k3s/latest/en/) or [Helm](https://helm.sh/docs).
+
+The charts consists of two separate sub-charts: _worker_ and _webui_. To install the chart simply execute `helm install openqa .` from this directory. To uninstall and start over, type `helm uninstall openqa`.
```
Martchus, line 9:
```suggestion
The chart consists of two separate sub-charts: _worker_ and _webui_. To install the chart simply execute `helm install openqa .` from this directory. To uninstall and start over, type `helm uninstall openqa`.
```

container/helm/charts/webui/templates/_helpers.tpl
```diff
@@ -0,0 +1,51 @@
...
+Expand the name of the chart.
+*/}}
+{{- define "webui.name" -}}
+{{- default .Chart.Name .Values.nameOverride | trunc 63 | trimSuffix "-" }}
+{{- end }}
+
+{{/*
+Create a default fully qualified app name.
+We truncate at 63 chars because some Kubernetes name fields are limited to this (by the DNS naming spec).
+If release name contains chart name it will be used as a full name.
```
Martchus, line 11: I can only guess the appropriate articles (the/a):
```suggestion
If the release name contains a chart name it will be used as a full name.
```

Martchus (commented):
That is a lot of code. I cannot say much about it but couldn't spot any obvious mistakes.

container/helm/charts/webui/templates/_helpers.tpl
baierjan: Nice catch, thanks. I did not study this file much (it is from helm template).

Makefile
```diff
@@ -312,6 +312,17 @@ tidy: tidy-js tidy-perl
 test-containers-compose:
 	tools/test_containers_compose
 
+.PHONY: test-helm-chart
+test-helm-chart: test-helm-lint test-helm-install
```
okurz, line 316: this could be called as part of https://github.com/os-autoinst/openQA/pull/4650/files#diff-76ed074a9305c04054cdebb9e9aad2d818052b07091de1f20cad0bbac34ffb52R151?

baierjan: The workflow has two jobs, one with `make test-helm-lint`, the second one with `make test-helm-install`. I just call the sub-targets in parallel and I am not using the `test-helm-chart` target directly.

okurz: Yes, I can see that the targets are used. It just would be great if we have something like a simple `make test` that simply calls every test. Do you think this can be done?

baierjan: I inspired myself in the https://github.com/os-autoinst/openQA/blob/master/.github/workflows/check-containers.yml where the same approach is used.

okurz (approved):

okurz:
> Since we have two targets outside the generic one I think it'd be nice to have a message like `The following targets are not included:\ntest-helm-chart\ntest-containers`. But that can be a follow-up PR

With "test-containers" do you mean "test-containers-compose"? Because that is still called as part of "make test", try with `make -d -n test | grep containers`

okurz:
https://github.com/os-autoinst/openQA/pull/4720

baierjan:
> > Since we have two targets outside the generic one I think it'd be nice to have a message like `The following targets are not included:\ntest-helm-chart\ntest-containers`. But that can be a follow-up PR
> 
> With "test-containers" do you mean "test-containers-compose"? Because that is still called as part of "make test", try with `make -d -n test | grep containers`

I meant the `test-check-containers` target which was not directly part of the test. I totally overlooked that conditional include of yet another target which eventually also does the container check. My bad, I already approved your addition.

PR #4653, opened 2022-05-13
Simplify setup by removing path rewrite

baierjan:
Make websockets server understand the `/api/v1/ws` route which is
currently rewritten by a reverse-proxy to `/ws`. This should remove the
need for having a reverse-proxy as all paths are now handled as are.

There is no need to change anything in the setup as the other route
(i.e. `/ws`) is also handled in the same way.

Prerequisites:
* #4660

### DIFF
--- a/container/webui/nginx.conf
+++ b/container/webui/nginx.conf
@@ -29,8 +29,6 @@ http {
             proxy_set_header Upgrade $http_upgrade;
             proxy_set_header Connection $connection_upgrade;
             proxy_set_header Host $host;
-
-            rewrite /api/v1/ws/(.*) /ws/$1 break;
         }
 
         location /liveviewhandler/ {
@@ -39,8 +37,6 @@ http {
             proxy_set_header Upgrade $http_upgrade;
             proxy_set_header Connection $connection_upgrade;
             proxy_set_header Host $host;
-
-            rewrite /liveviewhandler/(.*) /liveviewhandler/$1 break;
         }
 
         location /assets/ {
--- a/etc/apache2/vhosts.d/openqa-common.inc
+++ b/etc/apache2/vhosts.d/openqa-common.inc
@@ -58,13 +58,13 @@ ProxyPass /error !
 RewriteEngine On
 RewriteCond %{HTTP:Connection} Upgrade [NC]
 RewriteCond %{HTTP:Upgrade} websocket [NC]
-RewriteRule /api/v1/ws/(.*) ws://localhost:9527/ws/$1 [P,L]
+RewriteRule /api/v1/ws/(.*) ws://localhost:9527/api/v1/ws/$1 [P,L]
 RewriteCond %{HTTP:Connection} Upgrade [NC]
 RewriteCond %{HTTP:Upgrade} websocket [NC]
 RewriteRule /liveviewhandler/(.*) ws://localhost:9528/liveviewhandler/$1 [P,L]
 
 # pass websocket server where the worker connects to port 9527
-ProxyPass "/api/v1/ws/" "http://localhost:9527/ws/" keepalive=On
+ProxyPass "/api/v1/ws/" "http://localhost:9527/api/v1/ws/" keepalive=On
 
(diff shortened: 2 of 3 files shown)


container/webui/nginx.conf
```diff
@@ -30,7 +30,7 @@ http {
             proxy_set_header Connection $connection_upgrade;
             proxy_set_header Host $host;
 
-            rewrite /api/v1/ws/(.*) /ws/$1 break;
+            rewrite /api/v1/ws/(.*) /api/v1/ws/$1 break;
```
okurz, line 33: I understand your reasoning in the commit message but to me this looks like `/api/v1/ws/<foo>` is "rewritten" to `/api/v1/ws/<foo>` meaning not rewritten at all. What am I overlooking?

Martchus: Yes, isn't the idea to be able to drop the rewrite completely?

baierjan: Not rewritten at all, that is the intended behavior (or to put it differently, the rewrite is no longer needed). I could probably delete the rewrite completely, but I am not sure if it has another side-effect, as the same style of rewrite can be found in the liveviewhandler section: `rewrite /liveviewhandler/(.*) /liveviewhandler/$1 break;`.

If you are sure, it has no other functionality (like disabling other possible rewrites due to the `break`) I would delete both of them to keep it consistent.

Martchus: I suppose both rewrites can be deleted then.

okurz (commented):
good. And now to testing. Feel free to test on o3+osd as well after trying it out locally including the liveview

baierjan:
> good. And now to testing. Feel free to test on o3+osd as well after trying it out locally including the liveview

Hm, the only real code change is adding a route, that one was already tested during development. The liveview was only changed in the template for nginx, which is as far as I know not used on o3 nor osd. So there is not much added value here. I can probably use the setup from docker compose, where the nginx is actually used (in the load-balancer container which is not build by us).

okurz:
> > good. And now to testing. Feel free to test on o3+osd as well after trying it out locally including the liveview
> 
> Hm, the only real code change is adding a route, that one was already tested during development. The liveview was only changed in the template for nginx, which is as far as I know not used on o3 nor osd. So there is not much added value here.

Well, I mean for example just apply the apache changes on o3 or osd and see that everything still works as expected.

baierjan:
> > > good. And now to testing. Feel free to test on o3+osd as well after trying it out locally including the liveview
> > 
> > 
> > Hm, the only real code change is adding a route, that one was already tested during development. The liveview was only changed in the template for nginx, which is as far as I know not used on o3 nor osd. So there is not much added value here.
> 
> Well, I mean for example just apply the apache changes on o3 or osd and see that everything still works as expected.

I can't apply the apache changes before the code change (and the code change does **not** require anything else, the new route will just not be used in the current setup). Or do you mean I should apply both manually (because I do not see any added value here)?

Martchus:
The best approach is likely to install the packages generated by the OBS checks locally or on a staging instance and see whether it still works after restarting apache2 and the websocket server. (`/etc/apache2/vhosts.d/openqa-common.inc` is contained by the openQA package so simply installing the package from the OBS checks will replace both, the apache config and the websockt server code.)

okurz:
> Or do you mean I should apply both manually (because I do not see any added value here)?

Yes, sorry, then apply both. Or, create a separate pull request with just the code change first, have it merged, then we apply config changes afterwards

okurz:
#4660 merged

@Mergifyio rebase

So, will you test it out on o3/osd or any other instance?

okurz:
@Mergifyio rebase

baierjan:
> #4660 merged
> 
> So, will you test it out on o3/osd or any other instance?

Currently, this is set on osd. So far, no problems detected.

okurz (approved):
Ok, counts as successful test for me then :)

Martchus (approved):

PR #4655, opened 2022-05-16
Link to individual comments from the audit log

kraih:
Alternative approach for #4652. Discussions on Slack about another comment related bug showed that it can be quite useful to have `job_id`/`group_id`/`parent_group_id` in the audit log.

Progress: https://progress.opensuse.org/issues/111060

### DIFF
--- a/t/ui/23-audit-log.t
+++ b/t/ui/23-audit-log.t
@@ -15,7 +15,7 @@ use OpenQA::Client;
 
 use OpenQA::SeleniumTest;
 
-OpenQA::Test::Case->new->init_data;
+OpenQA::Test::Case->new->init_data(fixtures_glob => '01-jobs.pl 04-products.pl');
 driver_missing unless my $driver = call_driver;
 
 sub wait_for_data_table {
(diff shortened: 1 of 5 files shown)


kraih:
Tests will be added too before i merge this PR.

perlpunk (commented):
lgtm

okurz (approved):

Martchus (approved):

kraih:
Now with tests.

t/ui/23-audit-log.t
```diff
@@ -15,7 +15,7 @@ use OpenQA::Client;
 
 use OpenQA::SeleniumTest;
 
-OpenQA::Test::Case->new->init_data;
+OpenQA::Test::Case->new->init_data(fixtures_glob => '01-jobs.pl 04-products.pl');
```
okurz, line 18: Now I realized that you added this which we should avoid to save runtime and reduce inter-dependencies of tests. Well, you were right, we should have discarded review approvals after changes. At least here it would have prevented that ;) I will try to fix that.

PR #4656, opened 2022-05-17
Limit git log in investigation tab

perlpunk:
A `git log --stat` can take a long time for several thousands of
commits, especially if there are many changes.

Issue: https://progress.opensuse.org/issues/110677

This commit limits the log to 200 commits.

Currently it is not configurable.

I have tested the `git log --stat` and `git diff --stat` commands, and the worst is the `git log --stat` on the needles repo, it took over 3 minutes in our case. `git diff --stat` is not that bad, so I left it alone.

### DIFF
--- a/lib/OpenQA/WebAPI/Controller/Test.pm
+++ b/lib/OpenQA/WebAPI/Controller/Test.pm
@@ -902,7 +902,8 @@ sub dependencies ($self) {
 sub investigate {
     my ($self) = @_;
     return $self->reply->not_found unless my $job = $self->_get_current_job;
-    my $investigation = $job->investigate;
+    my $git_limit = OpenQA::App->singleton->config->{global}->{job_investigate_git_log_limit} // 200;
+    my $investigation = $job->investigate(git_limit => $git_limit);
     $self->render(json => $investigation);
 }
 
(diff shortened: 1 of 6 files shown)


lib/OpenQA/WebAPI/Controller/Test.pm
```diff
@@ -902,7 +902,7 @@ sub dependencies ($self) {
 sub investigate {
     my ($self) = @_;
     return $self->reply->not_found unless my $job = $self->_get_current_job;
-    my $investigation = $job->investigate;
+    my $investigation = $job->investigate(git_limit => 200);
```
Martchus, line 905: It would be nice if this value could come from the config file or from an env variable.

perlpunk:
I'm not sure why 10-jobs.t fails in CI; it works locally. It looks like the mocking isn't reset. Will try something.

lib/OpenQA/WebAPI/Controller/Test.pm
perlpunk: It is now configurable

Martchus (approved):
The name of the setting is long and cumbersome enough to fit well into our config file :-)

okurz (approved):

PR #4657, opened 2022-05-17
worker: Avoid logging secret variables

Martchus:
Using the same regex we also use in os-autoinst to filter secrets

### DIFF
--- a/lib/OpenQA/Worker/Engines/isotovideo.pm
+++ b/lib/OpenQA/Worker/Engines/isotovideo.pm
@@ -6,7 +6,7 @@ use Mojo::Base -base, -signatures;
 
 use Mojo::Base -signatures;
 use OpenQA::Constants qw(WORKER_SR_DONE WORKER_EC_CACHE_FAILURE WORKER_EC_ASSET_FAILURE WORKER_SR_DIED);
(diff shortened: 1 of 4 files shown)


kraih:
Considering that this is a security feature, maybe it should be an importable utility function with its own unit tests?

lib/OpenQA/Worker/Engines/isotovideo.pm
```diff
@@ -267,6 +267,11 @@ sub do_asset_caching ($job, $vars, $cache_dir, $assetkeys, $webui_host, $pooldir
         });
 }
 
+sub format_settings ($vars) {
+    my @s = map { $_ !~ qr/(^_SECRET_|_PASSWORD)/ ? ("    $_=$vars->{$_}") : ("    $_=[redacted]") } sort keys %$vars;
+    return join("\n", @s);
+}
```
kraih, line 273: Maybe in `OpenQA::Utils`?

okurz (approved):
I support the idea from kraih but also the current implementation looks ok and a net-benefit because so far I am also just aware of such problem within the scope of the worker engine

kraih (approved):

PR #4659, opened 2022-05-18
containers: Consistently use registry.opensuse.org

kalikiana:
We already use it in some but not in other cases. Being explicit avoids surprises and stops podman from asking to choose one.

See also https://github.com/os-autoinst/os-autoinst-distri-openQA/pull/85

### DIFF
--- a/container/ci/Dockerfile
+++ b/container/ci/Dockerfile
@@ -1,5 +1,5 @@
 #!BuildTag: openqa_dev
-FROM opensuse/leap:15.3
+FROM registry.opensuse.org/opensuse/leap:15.3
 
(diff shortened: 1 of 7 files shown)


okurz (changes requested):
That won't work, OBS internally relies on the relative paths

sqozz:
So an alternative approach would be to tell the "affected" system a default. From a very quick search I stumbled over https://www.redhat.com/sysadmin/manage-container-registries which explains quite a few ways for different scenarios. Would this be more feasible here? (Unfortunately I don't have an overview which "systems" are all involved/affected)

PR #4662, opened 2022-05-19
Add configuration for expiring minion jobs

perlpunk:
Issue: https://progress.opensuse.org/issues/111152

### DIFF
--- a/etc/openqa/openqa.ini
+++ b/etc/openqa/openqa.ini
@@ -195,6 +195,8 @@ concurrent = 0
 # Extends the job result cleanup to ensure the partition results are stored on does not become too full
 # (still experimental, relies on df)
 #results_min_free_disk_space_percentage = 0
+# Duration to keep minion jobs in the database, in seconds
+#minion_job_max_age = 604800
 
 [archiving]
 # Moves logs of jobs which are preserved during the cleanup because they are considered important
--- a/lib/OpenQA/Setup.pm
+++ b/lib/OpenQA/Setup.pm
@@ -14,6 +14,7 @@ use OpenQA::Utils qw(:DEFAULT assetdir random_string);
 use File::Path 'make_path';
 use POSIX 'strftime';
 use Time::HiRes 'gettimeofday';
+use Time::Seconds;
 use Scalar::Util 'looks_like_number';
 use OpenQA::Constants qw(DEFAULT_WORKER_TIMEOUT MAX_TIMER);
 use OpenQA::JobGroupDefaults;
@@ -148,6 +149,7 @@ sub read_config ($app) {
             screenshot_cleanup_batch_size => OpenQA::Task::Job::Limit::DEFAULT_SCREENSHOTS_PER_BATCH,
             screenshot_cleanup_batches_per_minion_job => OpenQA::Task::Job::Limit::DEFAULT_BATCHES_PER_MINION_JOB,
             results_min_free_disk_space_percentage => undef,
+            minion_job_max_age => ONE_WEEK,
         },
         archiving => {
             archive_preserved_important_jobs => 0,
--- a/t/config.t
+++ b/t/config.t
@@ -15,6 +15,7 @@ use OpenQA::Setup;
 use OpenQA::JobGroupDefaults;
 use OpenQA::Task::Job::Limit;
 use Mojo::File 'tempdir';
+use Time::Seconds;
 
 sub read_config {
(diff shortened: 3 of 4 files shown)


lib/OpenQA/Setup.pm
```diff
@@ -148,6 +149,7 @@ sub read_config ($app) {
             screenshot_cleanup_batch_size => OpenQA::Task::Job::Limit::DEFAULT_SCREENSHOTS_PER_BATCH,
             screenshot_cleanup_batches_per_minion_job => OpenQA::Task::Job::Limit::DEFAULT_BATCHES_PER_MINION_JOB,
             results_min_free_disk_space_percentage => undef,
+            expire_minion_jobs => ONE_DAY * 7,
```
okurz, line 152:
```suggestion
            expire_minion_jobs => ONE_WEEK,
```

t/config.t
```diff
@@ -131,6 +132,7 @@ subtest 'Test configuration default modes' => sub {
             asset_cleanup_max_free_percentage => 100,
             screenshot_cleanup_batch_size => OpenQA::Task::Job::Limit::DEFAULT_SCREENSHOTS_PER_BATCH,
             screenshot_cleanup_batches_per_minion_job => OpenQA::Task::Job::Limit::DEFAULT_BATCHES_PER_MINION_JOB,
+            expire_minion_jobs => ONE_DAY * 7,
```
okurz, line 135:
```suggestion
            expire_minion_jobs => ONE_WEEK,
```

okurz (changes requested):

lib/OpenQA/Setup.pm
perlpunk: btw, I was also wondering if one week is a good default or if we should even go for a higher default.

etc/openqa/openqa.ini
```diff
@@ -195,6 +195,8 @@ concurrent = 0
 # Extends the job result cleanup to ensure the partition results are stored on does not become too full
 # (still experimental, relies on df)
 #results_min_free_disk_space_percentage = 0
+##Duration to keep minion jobs in the database (Minion default is 2d)
+#expire_minion_jobs=604800
```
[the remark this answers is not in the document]

okurz (approved):

lib/OpenQA/Setup.pm
perlpunk: done

lib/OpenQA/WebAPI.pm
```diff
@@ -81,6 +81,9 @@ sub startup ($self) {
     # set cookie timeout to 48 hours (will be updated on each request)
     $self->app->sessions->default_expiration(48 * 60 * 60);
 
+    my $expire_minion = OpenQA::App->singleton->config->{misc_limits}->{expire_minion_jobs};
+    $self->app->minion->remove_after($expire_minion) if $expire_minion;
```
Martchus, line 85: I'm wondering whether this would be better set within the gru service. Likely @kraih knows.

kraih: Good point, it probably is cleaner to put the code into the `register` method in `OpenQA::Shared::Plugin::Gru`.

kraih:
The word `expire` is not a good choice here, since it can be confused with the actual expiration feature of Minion (which can be used to delete jobs that have not been processed yet after a certain time).

perlpunk:
> The word `expire` is not a good choice here, [...]

Do you have a better suggestion for a word?

lib/OpenQA/Setup.pm
```diff
@@ -148,6 +149,7 @@ sub read_config ($app) {
             screenshot_cleanup_batch_size => OpenQA::Task::Job::Limit::DEFAULT_SCREENSHOTS_PER_BATCH,
             screenshot_cleanup_batches_per_minion_job => OpenQA::Task::Job::Limit::DEFAULT_BATCHES_PER_MINION_JOB,
             results_min_free_disk_space_percentage => undef,
+            expire_minion_jobs => ONE_WEEK,
```
kalikiana, line 152: For consistency this should start with a noun.
```suggestion
            minion_job_max_age => ONE_WEEK,
```

perlpunk: done

lib/OpenQA/WebAPI.pm
perlpunk: done

kraih (approved):

PR #4663, opened 2022-05-20
Prevent excessive git logs by using log_trace

okurz:


okurz:
@perlpunk like this?

perlpunk:
No, as I said we still should debug stderr IMHO. Otherwise it would have been very easy of course.

Martchus:
I'm also not sure whether it makes sense to basically disable and error logging from Git. And just passing `1` as parameter isn't so nice. Maybe it makes more sense to pass the actual log level here.

okurz:
> No, as I said we still should debug stderr IMHO. Otherwise it would have been very easy of course.

ok, then I misunderstood you.

PR #4665, opened 2022-05-20
Reduce log-level of non-critical issue

okurz:
When users trigger jobs with an invalid job group we already ignore the
setting but so far we reported it as "warning". Well, I do not see what
we could do about this so let's just handle as "info".

Related progress issue: https://progress.opensuse.org/issues/105909

### DIFF
--- a/lib/OpenQA/Schema/ResultSet/Jobs.pm
+++ b/lib/OpenQA/Schema/ResultSet/Jobs.pm
@@ -195,7 +195,7 @@ sub create_from_settings {
     # associate currently available assets with job
     $job->register_assets_from_settings;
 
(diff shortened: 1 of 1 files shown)


Martchus (approved):

kalikiana (approved):

perlpunk:
So here we had a regression because statement coverage says the code is covered, but branch coverage would have told us it is not.

okurz:
Likely, yes. But I would have expected that a "compile" would fail if "log_info" is not defined at all.

perlpunk:
That cannot work in perl because you can even create subroutines at runtime.

`log_info "foo bar"` without the parens would have been a compile time error, though.

PR #4666, opened 2022-05-20
Fix missing import of 'log_info'

okurz:
Fixes regression from daeca39


perlpunk:
> I wonder how our tests in the pull request let that slip by but at least our openQA-in-openQA tests handled it :)

my guess: https://github.com/os-autoinst/openQA/pull/4665#issuecomment-1133338872

PR #4672, opened 2022-05-24
Add a limit for the maximum number of jobs to include in /tests/overview

kraih:
Before:
before

After:
after

The default of 500 jobs is an arbitrary choice, based on the number of jobs i saw clicking around the latest builds in the O3 database. Around 300 was the maximum i saw there. Happy to change it though.

Progress: https://progress.opensuse.org/issues/110680

### DIFF
--- a/lib/OpenQA/Schema/ResultSet/Jobs.pm
+++ b/lib/OpenQA/Schema/ResultSet/Jobs.pm
@@ -84,10 +84,14 @@ MACHINE. For each set of dupes, only the latest job found is included in
 the return array.
 
 =cut
-sub latest_jobs {
-    my ($self, $until) = @_;
+sub latest_jobs ($self, $until = undef) { @{$self->latest_jobs_with_limit($until)->{jobs}} }
 
-    my @jobs = $self->search($until ? {t_created => {'<=' => $until}} : undef, {order_by => ['me.id DESC']});
+sub latest_jobs_with_limit ($self, $until = undef, $limit = undef) {
+    my @jobs = $self->search(
+        $until ? {t_created => {'<=' => $until}} : undef,
+        {order_by => ['me.id DESC'], $limit ? (rows => $limit) : ()});
+    my $limit_exceeded = $limit ? @jobs >= $limit : undef;
+
     my @latest;
     my %seen;
     foreach my $job (@jobs) {
@@ -102,7 +106,7 @@ sub latest_jobs {
         next if $seen{$key}++;
         push(@latest, $job);
     }
-    return @latest;
+    return {jobs => \@latest, limit_exceeded => $limit_exceeded};
 }
 
 sub create_from_settings {
--- a/lib/OpenQA/WebAPI/Controller/Test.pm
+++ b/lib/OpenQA/WebAPI/Controller/Test.pm
@@ -4,6 +4,7 @@
 package OpenQA::WebAPI::Controller::Test;
 use Mojo::Base 'Mojolicious::Controller', -signatures;
 
+use OpenQA::App;
 use OpenQA::Utils;
 use OpenQA::Jobs::Constants;
 use OpenQA::Schema::Result::Jobs;
@@ -500,8 +501,7 @@ sub _calculate_preferred_machines {
 }
 
 # Take an job objects arrayref and prepare data structures for 'overview'
-sub prepare_job_results {
-    my ($self, $jobs) = @_;
+sub _prepare_job_results ($self, $jobs) {
     my %archs;
     my %results;
     my $aggregated = {
@@ -657,8 +657,12 @@ sub overview {
         groups => $groups,
         until => $until,
     );
-    my @latest_jobs = $self->schema->resultset('Jobs')->complex_query(%$search_args)->latest_jobs($until);
-    ($stash{archs}, $stash{results}, $stash{aggregated}) = $self->prepare_job_results(\@latest_jobs);
+    my $limit = OpenQA::App->singleton->config->{misc_limits}->{tests_overview_max_jobs};
+    my $results
+      = $self->schema->resultset('Jobs')->complex_query(%$search_args)->latest_jobs_with_limit($until, $limit);
+
+    my $limit_exceeded = $results->{limit_exceeded};
+    ($stash{archs}, $stash{results}, $stash{aggregated}) = $self->_prepare_job_results($results->{jobs});
 
     # determine distri/version from job results if not explicitly specified via search args
     my @distris = keys %{$stash{results}};
@@ -693,6 +697,7 @@ sub overview {
         %stash,
         summary_parts => \@summary_parts,
         only_distri => $only_distri,
+        limit_exceeded => $limit_exceeded ? $limit : undef
     );
     $self->respond_to(
         json => {json => \%stash},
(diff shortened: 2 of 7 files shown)


kraih:
It's actually really hard to add a notification that will catch the user's attention on this page, because there is already a huge red block present at the very top.

t/config.t
```diff
@@ -133,6 +133,7 @@ subtest 'Test configuration default modes' => sub {
             screenshot_cleanup_batch_size => OpenQA::Task::Job::Limit::DEFAULT_SCREENSHOTS_PER_BATCH,
             screenshot_cleanup_batches_per_minion_job => OpenQA::Task::Job::Limit::DEFAULT_BATCHES_PER_MINION_JOB,
             minion_job_max_age => ONE_WEEK,
+            tests_overview_max_jobs => 500
```
perlpunk, line 136:
```suggestion
            tests_overview_max_jobs => 500,
```

lib/OpenQA/WebAPI/Controller/Test.pm
```diff
@@ -693,6 +697,7 @@ sub overview {
         %stash,
         summary_parts => \@summary_parts,
         only_distri => $only_distri,
+        limit_exceeded => $limit_exceeded
```
perlpunk, line 700:
```suggestion
        limit_exceeded => $limit_exceeded,
```

t/config.t
kraih: Arguable, we don't have a rule for that i believe. 🤔

lib/OpenQA/WebAPI/Controller/Test.pm
```diff
@@ -657,8 +657,12 @@ sub overview {
         groups => $groups,
         until => $until,
     );
-    my @latest_jobs = $self->schema->resultset('Jobs')->complex_query(%$search_args)->latest_jobs($until);
-    ($stash{archs}, $stash{results}, $stash{aggregated}) = $self->prepare_job_results(\@latest_jobs);
+    my $limit = $search_args->{limit} = OpenQA::App->singleton->config->{misc_limits}->{tests_overview_max_jobs};
+    my $jobs = $self->schema->resultset('Jobs')->complex_query(%$search_args);
+    my $limit_exceeded = $jobs->count >= $limit ? $limit : undef;
+    my @latest_jobs = $jobs->latest_jobs($until);
```
Martchus, line 663: Since `latest_jobs` might narrow down the number of jobs further it would make sense to evaluate whether the limit is exceeded only after that:

```suggestion
    my @latest_jobs = $jobs->latest_jobs($until);
    my $limit_exceeded = scalar @latest_jobs >= $limit ? $limit : undef;
```

perlpunk, line 663: The limit might not be exceeded if `$until` is given, but that is only used after doing the job count and check for limit.

t/config.t
kraih: If we want to insist on trailing commas, there's probably a perltidy option we should activate.

lib/OpenQA/WebAPI/Controller/Test.pm
```diff
@@ -657,8 +657,12 @@ sub overview {
         groups => $groups,
         until => $until,
     );
-    my @latest_jobs = $self->schema->resultset('Jobs')->complex_query(%$search_args)->latest_jobs($until);
-    ($stash{archs}, $stash{results}, $stash{aggregated}) = $self->prepare_job_results(\@latest_jobs);
+    my $limit = $search_args->{limit} = OpenQA::App->singleton->config->{misc_limits}->{tests_overview_max_jobs};
+    my $jobs = $self->schema->resultset('Jobs')->complex_query(%$search_args);
+    my $limit_exceeded = $jobs->count >= $limit ? $limit : undef;
```
perlpunk, line 662: Not sure how expensive the count here is (additionally to the real select).
Why not do a select with a limit?

lib/OpenQA/WebAPI/Controller/Test.pm
kraih: I've tried that actually, the limit was much less reliable then and did not work for the case the ticket is about.

lib/OpenQA/WebAPI/Controller/Test.pm
Martchus: I suppose the select *is* limited. That's also the reason for the `>=` (and not just `>`).

t/config.t
perlpunk: If there is one, we should.
Not having trailing commas is one reason why I hate JSON... =)

lib/OpenQA/WebAPI/Controller/Test.pm
kraih: Just moving the check doesn't work for the case the ticket is about, i'll have to revisit the whole logic.

lib/OpenQA/WebAPI/Controller/Test.pm
kraih: Turned out to be really cheap.

okurz (approved):
Next step could be to allow overwriting the limit also by query parameter on request.

lib/OpenQA/WebAPI/Controller/Test.pm
kraih: The problem is the two resultsets being used here, one created by `complex_query` and the other by `latest_jobs`. Without a limit i get 82991 for the first and 67539 for the second. With a limit on the first query i get 500 for the first and 468 for the second. If i apply the limit to the second only i get 82991 for the first and 468 for the second. 😁 Good luck trying to detect when the limit clause actually triggers!

kraih: See other comment for explanation: https://github.com/os-autoinst/openQA/pull/4672#discussion_r880727332

Martchus:
> Next step could be to allow overwriting the limit also by query parameter on request.

But we should only allow to specify a lower limit than what's configured server-side.

kraih:
Ok, i pushed a more "correct" version (explanation in https://github.com/os-autoinst/openQA/pull/4672#discussion_r880727332). But the API for `latest_jobs` is pretty ugly now and i can't think of something cleaner at the moment. Would be happy for suggestions.

kraih:
For something cleaner i would probably have to change the return value and break all other uses of the method.

kraih:
The latest version i don't hate anymore, but ideas for a better API are still welcome.

lib/OpenQA/WebAPI/Controller/Test.pm
kraih: Changed the API a bit to make it work.

okurz (approved):

lib/OpenQA/Schema/ResultSet/Jobs.pm
```diff
@@ -84,10 +84,14 @@ MACHINE. For each set of dupes, only the latest job found is included in
 the return array.
 
 =cut
-sub latest_jobs {
-    my ($self, $until) = @_;
+sub latest_jobs ($self, $until = undef) { @{$self->latest_jobs_with_limit($until)->{jobs}} }
```
Martchus, line 87: It would be nice to avoid copying the array here but I suppose one cannot have everything.

Martchus (approved):

lib/OpenQA/Schema/ResultSet/Jobs.pm
kraih: No big difference to how it was before.

PR #4677, opened 2022-05-27
t: Simplify ui/23-audit-log.t again to use no fixtures

okurz:
Fixtures should be avoided to prevent cross-dependencies and reduce
runtime.

Related progress issue: https://progress.opensuse.org/issues/111539


t/ui/23-audit-log.t
```diff
@@ -130,9 +128,8 @@ subtest 'clickable events' => sub {
     wait_for_data_table;
     $table = $driver->find_element_by_id('audit_log_table');
     @entries = $driver->find_child_elements($table, 'tbody/tr', 'xpath');
-    is(scalar @entries, 2, 'three elements') or return diag $_->get_text for @entries;
+    is(scalar @entries, 1, 'three elements') or return diag $_->get_text for @entries;
```
perlpunk, line 132: I know you didn't introduce the mistakes, but

* `three` is wrong here
* the `return` will result in exactly one `diag` call and will discard the rest of the `@entries`

okurz: On, right :)

okurz: done

perlpunk (approved):

kalikiana (approved):
Nice!

PR #4678, opened 2022-05-30
Prevent dangerous package removal by forcing resolution

okurz:
Whenever zypper repositories are unavailable or incompletely read it can
happen (and did happen) that packages are forcefully removed as zypper
was forced by us to find a resolution which is not what we want.

Related progress issue: https://progress.opensuse.org/issues/111755


baierjan (approved):

Martchus (approved):
Reminds me of problems I previously noticed with openqa-continuous-update (where I first had the option as well). So I suppose removing the option makes generally sense.

PR #4680, opened 2022-05-30
Remove a pointless git warning

kraih:
```
[2022-04-25T09:17:28.805933Z] [warn] [pid:8752] /var/lib/openqa/share/tests/obs/needles is not a git directory
```

The warning does look very pointless in the openQA log. The `gitrepodir` function is written in a way that encourages probing for information, with a proper return value to indicate failure, and the investigate feature uses it that way already.

Progress: https://progress.opensuse.org/issues/110260


kraih:
The warning was originally added [with another feature](https://github.com/os-autoinst/openQA/commit/95f80029575738438ecc4d346397ace926678608) and not even tested. The tests only came later in [an unrelated attempt](https://github.com/os-autoinst/openQA/commit/847ec93974d865890fcadb332713cf6a50a52e5b) to improve coverage.

okurz (approved):
Thank you for also checking why that message was introduced in the first place

perlpunk (approved):

PR #4682, opened 2022-05-31

okurz:
* Simplify OpenQA::WebAPI::Controller::API::V1::Job with early return
* OpenQA::Schema::ResultSet::DeveloperSessions: Extract method
* Simplify database operations in WebAPI::Controller::Test

### DIFF
--- a/lib/OpenQA/Schema/ResultSet/DeveloperSessions.pm
+++ b/lib/OpenQA/Schema/ResultSet/DeveloperSessions.pm
@@ -3,46 +3,34 @@
 
 package OpenQA::Schema::ResultSet::DeveloperSessions;
 
-
-use Mojo::Base 'DBIx::Class::ResultSet';
-
+use Mojo::Base 'DBIx::Class::ResultSet', -signatures;
 use Try::Tiny;
 use OpenQA::Constants qw(WORKER_COMMAND_DEVELOPER_SESSION_START);
 use OpenQA::Schema::Result::DeveloperSessions;
 use OpenQA::WebSockets::Client;
 use OpenQA::Log 'log_error';
 
-sub register {
-    my ($self, $job_id, $user_id) = @_;
+sub _find_or_create_session ($self, $job_id, $user_id) {
+    # refuse if no worker assigned
+    my $worker = $self->result_source->schema->resultset('Workers')->find({job_id => $job_id});
+    return unless ($worker);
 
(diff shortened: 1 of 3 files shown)


lib/OpenQA/Schema/ResultSet/DeveloperSessions.pm
```diff
@@ -3,46 +3,34 @@
...
-                $session = $self->create(
-                    {
-                        job_id => $job_id,
-                        user_id => $user_id,
-                    });
-
-            }
-            return ($session, $worker->id, $is_session_already_existing);
-        });
+    my ($res, $worker_id, $session_exists) = $schema->txn_do(sub { self->find_or_create_session($job_id, $user_id) });
```
perlpunk, line 30: Did you mean `$self`?

lib/OpenQA/WebAPI/Controller/Test.pm
```diff
@@ -138,27 +137,24 @@ sub list_running_ajax {
...
-                flavor => $job->FLAVOR // '',
-                arch => $job->ARCH // '',
-                build => $job->BUILD // '',
-                testtime => ($job->t_started // '') . 'Z',
-                group => $job->group_id,
-                state => $job->state,
-                progress => $job->progress_info,
-            });
-    }
+    my @running = map {
```
perlpunk, line 140: Hm, I remember this change from a previous PR, and I think I already wrote that I don't see the advantage here.
`->next` might actually be preferrable to `->all` which reads all DBIC objects into memory

lib/OpenQA/WebAPI/Controller/Test.pm
```diff
@@ -178,28 +174,25 @@ sub list_scheduled_ajax {
...
-                arch => $job->ARCH // '',
-                build => $job->BUILD // '',
-                testtime => $job->t_created . 'Z',
-                group => $job->group_id,
-                state => $job->state,
-                blocked_by_id => $job->blocked_by_id,
-                prio => $job->priority,
-            });
-    }
+    my @scheduled = map {
```
perlpunk, line 177: same here

lib/OpenQA/Schema/ResultSet/DeveloperSessions.pm
Martchus: And it needed to be `_find_or_create_session`.

PR #4683, opened 2022-05-31
Tiny code improvement

perlpunk:
While looking at that file for another PR, the "monkey patch" caught my eye, which is apparently unnecessary.
Looking further down, the repeated assigning between `$jobs` and `$jobs_to_cancel` confused me, so I generously gave the code two new variables.


kalikiana (approved):
Nice

PR #4681, opened 2022-05-30
jobs: Supply a reason in all cases

kalikiana:
Also allow API calls to specify a reason.

See: https://progress.opensuse.org/issues/111590#note-5

PR #4686, opened 2022-06-02
jobs: Emit cancel_by_settings in all cases

kalikiana:
As opposed to only the case where an API call was used.


kalikiana:
I noticed that #4681 was giving me the new reason but not the audit event for two jobs https://openqa.opensuse.org/tests/2398261 and https://openqa.opensuse.org/tests/2398266 and it turns out there's other cases where we cancel but the log is decoupled

okurz (approved):

PR #4687, opened 2022-06-02
Replace openqa-worker@ with a link to openqa-worker-plain@

baierjan:
Rename generic openqa-worker service to openqa-worker-plain and provide
compatibility symlink. This should not have any impact as both names can
be used in the respective systemctl calls.

In the next phase, we can update spec file to respect the symlink to
allow users to choose the "implementation" of the openqa-worker service.
Currently we recognize two possible options: openqa-worker-plain and
openqa-worker-auto-restart.

Reference: https://progress.opensuse.org/issues/109734


okurz (commented):
This looks promising. I suggest you include a bit of outlook in the git commit message to describe what you intend this change for, basically the transition plan from the ticket

okurz (approved):

perlpunk (approved):

PR #4690, opened 2022-06-03
Fix regression by moving tests/overview limit from SQL back to Perl

kraih:
Every time there are more than 500 jobs in the results we don't know if
the wrong ones have been left out. Because theoretically all 500 could
be sharing the same de-duplication key, taking up the space required by
other more releavant jobs.

The problem is that the de-duplication step needs to happen before jobs
get filtered out by conditions like the result. Unfortunately that
happens in Perl, after the main SQL has already been processed. I've
looked into doing the de-duplication in SQL too, but couldn't find a
solution that was good enough. So for now the best solution is to move
the limit from SQL back to Perl. Which fortunately seems to still be
fast enough for all the cases we care about.

What makes this issue so complicated is that the request and response
data goes through multiple transformations that have continuously grown
in complexity over the past 10 years:

* The `compose_job_overview_search_args` helper reads the HTTP GET
  parameters and transforms them into a simplified data structure
  vaguely representing a `DBIx::Class` query, parts of it are used to
  perform intermediate SQL queries (called in 2 places)

* The `complex_query` resultset method transforms various comma
  separated values into array references, and then uses
  `_prepare_complex_query_search_args` to transform about 20
  simplified parameters into something `DBIx::Class` can understand,
  before returning a resultset using these values as search parameters
  (called in 10 places)

* The `latest_jobs` method is called on the existing resultset, and adds
  more search parameters before retrieving the full list of matching
  jobs from PostgreSQL and doing a de-duplication of jobs based on
  various properties in memory (called in 6 places)

* And this is where things get complex :grin:, because now the
  `overview` action will call the `_prepare_job_results` method to
  prefetch various job related data such as comments (for labels) and
  then pull in HTTP GET parameters again and do the actual filtering of
  jobs for the `failed_modules`, `state`, `result`, `arch`, and `machine`
  conditions (all were previously relevant for de-duplication), and then
  transform the whole data structure into something that can be rendered
  in templates (the action can actually be called in 2 ways, because of
  a backcompat hack)

So, because i moved the limit condition from SQL to Perl, which makes
everything a little bit slower, i have also moved the final filtering
step before the part that prefetches comments. This should result in an
overall performance boost for all `tests/overview` requests.

Progress: https://progress.opensuse.org/issues/111833

### DIFF
--- a/lib/OpenQA/Schema/ResultSet/Jobs.pm
+++ b/lib/OpenQA/Schema/ResultSet/Jobs.pm
@@ -84,13 +84,8 @@ MACHINE. For each set of dupes, only the latest job found is included in
 the return array.
 
 =cut
-sub latest_jobs ($self, $until = undef) { @{$self->latest_jobs_with_limit($until)->{jobs}} }
-
-sub latest_jobs_with_limit ($self, $until = undef, $limit = undef) {
-    my @jobs = $self->search(
-        $until ? {t_created => {'<=' => $until}} : undef,
-        {order_by => ['me.id DESC'], $limit ? (rows => $limit) : ()});
-    my $limit_exceeded = $limit ? @jobs >= $limit : undef;
+sub latest_jobs ($self, $until = undef) {
+    my @jobs = $self->search($until ? {t_created => {'<=' => $until}} : undef, {order_by => ['me.id DESC']});
 
     my @latest;
     my %seen;
@@ -106,7 +101,8 @@ sub latest_jobs_with_limit ($self, $until = undef, $limit = undef) {
         next if $seen{$key}++;
         push(@latest, $job);
     }
-    return {jobs => \@latest, limit_exceeded => $limit_exceeded};
+
+    return @latest;
 }
 
 sub create_from_settings {
@@ -206,7 +202,7 @@ sub create_from_settings {
     return $job;
 }
 
-sub search_modules ($self, $module_re) {
+sub _search_modules ($self, $module_re) {
     my $distris = path(testcasedir);
     my @results;
     for my $distri ($distris->list({dir => 1})->map('realpath')->uniq()->each) {
@@ -222,12 +218,12 @@ sub search_modules ($self, $module_re) {
     return \@results;
 }
 
-sub prepare_complex_query_search_args ($self, $args) {
+sub _prepare_complex_query_search_args ($self, $args) {
     my @conds;
     my @joins;
 
     if ($args->{module_re}) {
-        my $modules = $self->search_modules($args->{module_re});
+        my $modules = $self->_search_modules($args->{module_re});
         push @{$args->{modules}}, @$modules;
     }
 
@@ -320,7 +316,7 @@ sub complex_query ($self, %args) {
         next unless $args{$arg};
         $args{$arg} = [split(',', $args{$arg})] unless (ref($args{$arg}) eq 'ARRAY');
     }
-    my ($conds, $attrs) = $self->prepare_complex_query_search_args(\%args);
+    my ($conds, $attrs) = $self->_prepare_complex_query_search_args(\%args);
     my $jobs = $self->search({-and => $conds}, $attrs);
     return $jobs;
 }
--- a/lib/OpenQA/WebAPI/Controller/Test.pm
+++ b/lib/OpenQA/WebAPI/Controller/Test.pm
@@ -501,7 +501,7 @@ sub _calculate_preferred_machines {
 }
 
 # Take an job objects arrayref and prepare data structures for 'overview'
-sub _prepare_job_results ($self, $jobs) {
+sub _prepare_job_results ($self, $all_jobs, $limit) {
     my %archs;
     my %results;
     my $aggregated = {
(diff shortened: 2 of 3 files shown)


kraih:
TLDR: This is a performance optimisation that coincidentally also fixes the progress ticket. 😉

lib/OpenQA/Schema/ResultSet/Jobs.pm
```diff
@@ -84,13 +84,8 @@ MACHINE. For each set of dupes, only the latest job found is included in
...
 =cut
-sub latest_jobs ($self, $until = undef) { @{$self->latest_jobs_with_limit($until)->{jobs}} }
-
-sub latest_jobs_with_limit ($self, $until = undef, $limit = undef) {
-    my @jobs = $self->search(
-        $until ? {t_created => {'<=' => $until}} : undef,
-        {order_by => ['me.id DESC'], $limit ? (rows => $limit) : ()});
-    my $limit_exceeded = $limit ? @jobs >= $limit : undef;
+sub latest_jobs ($self, $until = undef) {
+    my @jobs = $self->search($until ? {t_created => {'<=' => $until}} : undef, {order_by => ['me.id DESC']});
```
okurz, line 88: Your reasoning makes sense to me and as long as nobody can come up with a way to pass filter+limit parameters directly as according SQL commands and parameters I think the general approach is ok. I just wonder if we can still apply *some* reasonable limit when we actually query the database. Not necessarily to guard against malicious attacks by users but to protect ourselves against implementation bugs as well to prevent unlimited SQL queries, e.g. here, especially as you directly assign to an array so I assume the actual table rows are converted into the Perl array here already.

kraih: Any limit applied here would be arbitrary and bound to cause similar issues. Moving the de-duplication to SQL is the only real solution imo. Fortunately even the worst case examples i came up with on my local O3 clone with many thousands of jobs never took longer than 2 seconds to hit the new 500 job limit in the Perl code.

kraih: The real performance bottleneck here has always been the prefetching of comments. Which is pretty well shielded in this version. Everything else only has a minor impact (< 2s).

kraih:
Long term, we probably have to duplicate most of the code and move it all into one place, and then start optimising the various SQL queries without having to worry about breaking all other places sharing the same code.

lib/OpenQA/Schema/ResultSet/Jobs.pm
kraih: You would fetch all of them, `DBD::Pg` has no support for paging, so it will always fetch all results at once. Here's a sample query generated by `DBIx::Class` for Tumbleweed jobs (82991 results, 1.3s to retrieve all):
```
EXECUTE SELECT me.id, me.result_dir, me.archived, me.state, me.priority, me.result, me.reason, me.clone_id, me.blocked_by_id, me.backend_info, me.TEST, me.DISTRI, me.VERSION, me.FLAVOR, me.ARCH, me.BUILD, me.MACHINE, me.group_id, me.assigned_worker_id, me.t_started, me.t_finished, me.logs_present, me.passed_module_count, me.failed_module_count, me.softfailed_module_count, me.skipped_module_count, me.externally_skipped_module_count, me.scheduled_product_id, me.result_size, me.t_created, me.t_updated FROM jobs me WHERE ( ( me.clone_id IS NULL AND me.VERSION = $1 ) ) ORDER BY me.id DESC (
$1: Tumbleweed
);
```

kraih: So we do still have some headroom, but sooner or later we will have to deal with this and untangle the query generator.

okurz: Hm, hoping for an opinion by other field experts.

perlpunk (approved):
Although I have been working on this code, I wasn't aware that the comment fetching could be moved to a later point.

A configurable limit for the SQL would still be good, I guess.

kraih:
I don't think a limit on the SQL query would gain us anything before we've moved the de-duplicating into SQL. It's not a bottleneck right now and probably won't be anytime soon.

lib/OpenQA/Schema/ResultSet/Jobs.pm
kraih: That would be good, it's a complex issue. Given the commit history of the affected functions, @coolo and @Martchus probably know the code and its history best.

okurz:
@kraih could you please add a test to show that there is no warning about filter limit on combined queries? Should be easy given that https://openqa.opensuse.org/tests/overview?result=incomplete&groupid=1&groupid=3 already shows a warning when it shouldn't

okurz (dismissed):
Please see https://github.com/os-autoinst/openQA/pull/4690#issuecomment-1148381678 and also rephrase the git commit message subject line to point out the important difference how users are impacted, not just the implementation detail of where we do the limit.

kraih:
Test added and commit message updated.

okurz (approved):

kraih:
> @kraih could you please add a test to show that there is no warning about filter limit on combined queries? Should be easy given that https://openqa.opensuse.org/tests/overview?result=incomplete&groupid=1&groupid=3 already shows a warning when it shouldn't

And that warning is gone now. 🥳

PR #4692, opened 2022-06-07
systemd: Depend on time being in sync if possible

kalikiana:
If the system has a unit implementing time-sync the worker service will
not come up until it can rely on the time. This prevents authentication
errors due to a timestamp mismatch.

See: https://progress.opensuse.org/issues/105855

### DIFF
--- a/systemd/openqa-worker-plain@.service
+++ b/systemd/openqa-worker-plain@.service
@@ -4,7 +4,7 @@
 # replace '1' with the instance number you want
 [Unit]
 Description=openQA Worker #%i
-After=openqa-slirpvde.service network.target nss-lookup.target remote-fs.target
+After=openqa-slirpvde.service network.target nss-lookup.target remote-fs.target time-sync.target
 PartOf=openqa-worker.target
 
 [Service]


systemd/openqa-worker.target
```diff
@@ -1,6 +1,7 @@
 [Unit]
 Description=openQA Worker
 Wants=var-lib-openqa-share.mount
+After=time-sync.target
```
okurz, line 4: Hm, not sure if we should put this into `openqa-worker.target` or rather `openqa-worker@.service`?

okurz (changes requested):

okurz (approved):

okurz:
> can we be sure that the time is set when `time-sync.target` is reached?

I don't fully trust it but it's my understanding that the actual *target* means that time is synced, yes. If this does not work then there might be a problem in the time sync problem.

kalikiana:
> This looks like a good start - I wonder, if the ntp sync is done fast enough, though.
> 
> ```
> # /etc/chrony.d/pool.conf
> pool 2.suse.pool.ntp.org iburst
> ```
> 
> The `iburst` is in our default chrony configuration to ensure a quick ntp sync, but can we be sure that the time is set when `time-sync.target` is reached?

It's up to the systemd service to include time-set.target in Before= and Wants=. We rely on that by contract. See systemd.special(7)

Martchus (approved):

PR #4693, opened 2022-06-08
Verify when maintenance update has already been released

DeepthiYV:
Before cloning or restarting a job, verify the maintenance update has already been released.  

- Related ticket: https://progress.opensuse.org/issues/109707

### DIFF
--- a/lib/OpenQA/Script/CloneJob.pm
+++ b/lib/OpenQA/Script/CloneJob.pm
@@ -13,6 +13,7 @@ use OpenQA::Client;
 use Mojo::File 'path';
 use Mojo::URL;
 use Mojo::JSON;    # booleans
+use OpenQA::Script::CloneJobSUSE;
 
 our @EXPORT = qw(
   clone_jobs
@@ -245,8 +246,9 @@ sub clone_job ($jobid, $url_handler, $options, $post_params = {}, $jobs = {}, $d
     return $post_params if defined $post_params->{$jobid};
 
     my $job = $jobs->{$jobid} = clone_job_get_job($jobid, $url_handler, $options);
-    my $settings = $post_params->{$jobid} = {%{$job->{settings}}};
 
+    my $settings = $post_params->{$jobid} = {%{$job->{settings}}};
+    OpenQA::Script::CloneJobSUSE::detect_maintenance_update($jobid, $url_handler, $settings);
     my $clone_children = $options->{'clone-children'};
     my $max_depth = $options->{'max-depth'} // 1;
     for my $job_type (qw(parents children)) {
--- a/lib/OpenQA/Script/CloneJobSUSE.pm
+++ b/lib/OpenQA/Script/CloneJobSUSE.pm
@@ -1,0 +1,41 @@
+# Copyright 2022 SUSE LLC
+# SPDX-License-Identifier: GPL-2.0-or-later
+
+package OpenQA::Script::CloneJobSUSE;
+use Mojo::Base -strict, -signatures;
+use Data::Dump 'pp';
+use Exporter 'import';
+
+our @EXPORT = qw(detect_maintenance_update);
+
+sub collect_incident_repos ($url_handler, $settings) {
+    my @urls;
+    for my $setting (keys %$settings) {
+        return verify_incident_repos($url_handler, $settings->{'INCIDENT_REPO'}) if ($setting =~ /INCIDENT_REPO/);
+        if ($setting =~ /SCC_ADDONS/) {
+            foreach my $SCC_ADDON (split(/,/, $settings->{'SCC_ADDONS'})) {
+                next unless $settings->{uc($SCC_ADDON) . '_TEST_REPOS'};
+                my $incident_urls = verify_incident_repos($url_handler, $settings->{uc($SCC_ADDON) . '_TEST_REPOS'});
+                push @urls, @$incident_urls;
+            }
+        }
+    }
+    return \@urls;
+}
+
+sub verify_incident_repos ($url_handler, $incident_repos) {
+    my @incident_urls;
+    my $ua = $url_handler->{ua};
+    foreach my $incident (split(/,/, $incident_repos)) {
+        push @incident_urls, $incident unless $ua->get($incident)->is_success;
+    }
+    return \@incident_urls;
+}
+
+sub detect_maintenance_update ($jobid, $url_handler, $settings) {
+    my $urls = collect_incident_repos($url_handler, $settings);
+    die "Current job $jobid will fail, because the repositories for the below updates are unavailable\n" . pp($urls)
+      if @$urls;
+}
+
+1;
--- a/t/35-script_clone_job_suse.t
+++ b/t/35-script_clone_job_suse.t
@@ -1,0 +1,54 @@
+# Copyright 2022 SUSE LLC
+# SPDX-License-Identifier: GPL-2.0-or-later
+
+use Test::Most;
+use Test::Warnings ':report_warnings';
+
+use FindBin;
+use lib "$FindBin::Bin/lib", "$FindBin::Bin/../external/os-autoinst-common/lib";
+use OpenQA::Test::TimeLimit '6';
+use Mojo::Base -signatures;
+use Test::MockModule;
+use OpenQA::Script::CloneJobSUSE;
+use Mojo::URL;
+
+# define fake client
+{
+    package Test::FakeLWPUserAgentMirrorResult;
+    use Mojo::Base -base, -signatures;
+    has is_success => 1;
+}
+{
+    package Test::FakeLWPUserAgent;
+    use Mojo::Base -base, -signatures;
+    has is_validrepo => 1;
+    sub get ($self, $url) { Test::FakeLWPUserAgentMirrorResult->new(is_success => $self->is_validrepo) }
+}
+
+subtest 'maintenance update detect' => sub {
+    my $job_id = 1;
+    my %job = (
+        id => $job_id,
+        SCC_ADDONS => "we,sdk",
+        WE_TEST_REPOS => "http://foo/WE_suse/openqa,http://foo/WE_suse_2/openqa",
+        SDK_TEST_REPOS => "http://foo/SDK_suse/openqa,http://foo/SDK_suse_2/openqa",
+    );
+    my %incident_job = (
+        id => $job_id,
+        INCIDENT_REPO => "http://foo/incident_repo/openqa,http://foo/incident_repo_1/openqa",
+    );
+
+    my $fake_ua = Test::FakeLWPUserAgent->new;
+    my %url_handler = (remote_url => Mojo::URL->new('http://foo'), ua => $fake_ua);
+    my $clone_mock = Test::MockModule->new('OpenQA::Script::CloneJobSUSE');
+    $fake_ua->is_validrepo(1);
+    lives_ok { detect_maintenance_update($job_id, \%url_handler, \%job) } 'Maintenance updates are available';
+    lives_ok { detect_maintenance_update($job_id, \%url_handler, \%incident_job) } 'Maintenance updates are available';
+    $fake_ua->is_validrepo(0);
+    throws_ok { detect_maintenance_update($job_id, \%url_handler, \%job) } qr/Current job $job_id will fail/,
+      'Maintenance updates have been released';
+    throws_ok { detect_maintenance_update($job_id, \%url_handler, \%incident_job) } qr/Current job $job_id will fail/,
+      'Maintenance updates have been released';
+};
+
+done_testing();


lib/OpenQA/Script/CloneJob.pm
```diff
@@ -241,10 +241,37 @@ sub clone_jobs ($jobid, $options) {
...
+            my $maintainance_url = $test_template;
+            $maintainance_url =~ s/\@INCIDENTNR\@/$TEST_ISSUE/ ;
+            my $response = $ua->get($maintainance_url);
+            if (!$response->is_success) {
+               push @list_url, $maintainance_url;
+            }
+        }
+    }
+    if (@list_url) {
+      die "Current job will fail, because the below updates has already been released\n" , pp (@list_url);
```
kalikiana, line 263:
```suggestion
      die "Current job will fail, because the below updates have already been released\n" , pp (@list_url);
```

lib/OpenQA/Script/CloneJob.pm
```diff
@@ -241,10 +241,37 @@ sub clone_jobs ($jobid, $options) {
...
+    my @list_url ;
+    for ( keys %test_template) {
+       next  unless ($settings{$_} or  $settings{$test_template{$_}});
+       my $test_template = $settings{$_};
+       my $ua = LWP::UserAgent->new;
+       $ua->timeout(10);
+       foreach my $TEST_ISSUE (split (/,/ ,  $settings{$test_template{$_}})) {
+            my $maintainance_url = $test_template;
+            $maintainance_url =~ s/\@INCIDENTNR\@/$TEST_ISSUE/ ;
+            my $response = $ua->get($maintainance_url);
```
kalikiana, line 256: So this tries to find the URL for the incident and aborts if that is absent. I wonder if we can make this more generic. A simple URL check could be done that's not as specific about incidents.

okurz (changes requested):
The concept of SLE maintenance tests are not known to openQA and seems like you managed to fix a valid problem but the generic openQA clone job script is certainly not the right approach. Maybe think about a downstream script that can be used within the tooling

lib/OpenQA/Script/CloneJob.pm
foursixnine: the clone script has already asset management :) this would apply only to specific issues

DeepthiYV:
> The concept of SLE maintenance tests are not known to openQA and seems like you managed to fix a valid problem but the generic openQA clone job script is certainly not the right approach. Maybe think about a downstream script that can be used within the tooling

Thank you Oliver for the suggestion. Could you please suggest me where can we implement this change. 
I thought of adding this change at the beginning of openQA clone jobs to detect when maintenance updates have been released before downloading the assets for the execution.

okurz (changes requested):
> Thank you Oliver for the suggestion. Could you please suggest me where can we implement this change.

I am thinking of three possible approaches:
1. As openQA already provides a warning if a mandatory "asset" is missing likely the best approach would be to ensure that according repos are available on openQA itself so that a consistent test base is available for the lifetime of a test. As soon as that asset (e.g. a repo) is pruned from openQA by the automatic cleanup then any clone attempt would be prevented with an informative message about the unavailability of the asset. So that would mean there would be no need to do any changes in openQA itself but change how openQA syncs assets and triggers tests, likely in https://github.com/openSUSE/qem-bot/
2. develop the SLE maintenance specific handling as custom "plugin" code for openqa CLI applications. That's likely the hardest but we already have a plugin structure for openQA CLI applications so in the end this might be the best way to go. What likely would be necessary is to rework openqa-clone-job to reuse what openqa-cli has in the backend and then implement custom commands in a separate repo together with the complete deployment structure, e.g. a separate openSUSE package that can be installed along with the package openQA-client
3. Prepare a separate project where you have a new application that internally calls openqa-clone-job and on top does what you do here.

> I thought of adding this change at the beginning of openQA clone jobs to detect when maintenance updates have been released before downloading the assets for the execution.

Sure. Let me ask, why does anybody care about cloning openQA jobs for maintenance updates that have already been released in the first place?

DeepthiYV:
> > Thank you Oliver for the suggestion. Could you please suggest me where can we implement this change.
> 
> I am thinking of three possible approaches:
> 
> 1. As openQA already provides a warning if a mandatory "asset" is missing likely the best approach would be to ensure that according repos are available on openQA itself so that a consistent test base is available for the lifetime of a test. As soon as that asset (e.g. a repo) is pruned from openQA by the automatic cleanup then any clone attempt would be prevented with an informative message about the unavailability of the asset. So that would mean there would be no need to do any changes in openQA itself but change how openQA syncs assets and triggers tests, likely in https://github.com/openSUSE/qem-bot/
> 2. develop the SLE maintenance specific handling as custom "plugin" code for openqa CLI applications. That's likely the hardest but we already have a plugin structure for openQA CLI applications so in the end this might be the best way to go. What likely would be necessary is to rework openqa-clone-job to reuse what openqa-cli has in the backend and then implement custom commands in a separate repo together with the complete deployment structure, e.g. a separate openSUSE package that can be installed along with the package openQA-client
> 3. Prepare a separate project where you have a new application that internally calls openqa-clone-job and on top does what you do here.
> 
> > I thought of adding this change at the beginning of openQA clone jobs to detect when maintenance updates have been released before downloading the assets for the execution.
> 
> Sure. Let me ask, why does anybody care about cloning openQA jobs for maintenance updates that have already been released in the first place?

Sometimes job fails when we restart or clone it while verifying the test, because the updates have already been released. So team decided to add a check before running a job. 
I will discuss with team and come up with a reasonable solution. Thanks Oliver for the feedback and suggestion :)

lib/OpenQA/Script/CloneJob.pm
```diff
@@ -241,10 +241,37 @@ sub clone_jobs ($jobid, $options) {
     handle_tx($tx, $url_handler, $options, $jobs) if $tx;
 }
 
+sub detect_maintaince_update($job) {
+    my %settings = %{$job->{settings}};
```
Martchus, line 245: Making a copy of the hash seems useless.

```suggestion
    my $settings = $job->{settings};
```
And then use it via `$settings->{…}`.

Martchus (changes requested):
What if the URL can just not be reached temporarily? There should definitely be an opt-out.

Please extend the unit tests and ensure the code is tidied.

Martchus:
Also note that the ticket says "we should have a warning". The "die" is a little bit more than a warning.

okurz:
> > Sure. Let me ask, why does anybody care about cloning openQA jobs for maintenance updates that have already been released in the first place?
> 
> Sometimes job fails when we restart or clone it while verifying the test, because the updates have already been released. So team decided to add a check before running a job.

What in particular do you mean with "while verifying the test"?

DeepthiYV:
> > > Sure. Let me ask, why does anybody care about cloning openQA jobs for maintenance updates that have already been released in the first place?
> > 
> > 
> > Sometimes job fails when we restart or clone it while verifying the test, because the updates have already been released. So team decided to add a check before running a job.
> 
> What in particular do you mean with "while verifying the test"?

As per my knowledge we re-execute/clone the failed openQA jobs of particular build(may be 2 days ago build)  to identify the failure reasons.

okurz:
> According to my knowledge, we re-execute/clone the failed openQA jobs of particular build(may be 2 days ago build) to identify the failure reasons.

which wouldn't reproduce a consistent same result if the maintenance update was released meanwhile

DeepthiYV:
> > According to my knowledge, we re-execute/clone the failed openQA jobs of particular build(may be 2 days ago build) to identify the failure reasons.
> 
> which wouldn't reproduce a consistent same result if the maintenance update was released meanwhile

Exactly this is what we want to avoid.

This is something we are facing in maintenance tests but also affects Tumbleweed when snapshot has been released.
The idea of this PR is to reduce the time of failure from possible minutes to seconds by checking the availability of repository earlier and for time being we will start with maintenance then we will add same logic to other products.
May be lets jump on call with @foursixnine  to find the common solution.

One Idea is to add the changes in openqa-clone-custom-git-refspec.

okurz:
> One Idea is to add the changes in openqa-clone-custom-git-refspec.

Well, openqa-clone-custom-git-refspec is using openqa-clone-job and the problem would apply the same for openqa-clone-job so I don't see that as beneficial.

foursixnine:
> > One Idea is to add the changes in openqa-clone-custom-git-refspec.
> 
> Well, openqa-clone-custom-git-refspec is using openqa-clone-job and the problem would apply the same for openqa-clone-job so I don't see that as beneficial.

Yeah, this is just one idea, however in the general sense the changes would need to go one way or the other into the clone job logic, since adding more wrappers or following any of the approaches you mentioned [earlier](https://github.com/os-autoinst/openQA/pull/4693#pullrequestreview-999864594) will result in unnecessary complexity.

I do agree however, that it's not a generic problem, but it's a problem that both, SUSE and openSUSE face, and benefits mostly test developers.

@okurz now if you have an idea of how plugins for clone-job can be written, I'd appreciate direct hints on where to do it, so that other communities can also benefit too ;)

okurz:
> By test developers I mean @os-autoinst/tests-maintainer and others who from time to time have to figure out why a job is failing, when the same job passed minutes earlier... only to figure out that it's due to a released snapshot or a released maintenance update

Why not just gracefully walk over non-existant repos within the test code?

okurz:
Let me go back one step. If you want …
1. openQA to fail quickly or even prevent you from cloning a job that is missing a mandatory asset then make those assets known to openQA, see http://open.qa/docs/#_specifying_assets_required_by_a_job
2. openSUSE/SLE tests to gracefully handle non-existant maintenance update repos then change https://github.com/os-autoinst/os-autoinst-distri-opensuse/blob/master/lib/qam.pm#L76 to either catch zypper failures for non-existant repos or check for the existance of repos before adding. The second might be easier to implement but is not race-free.

foursixnine:
> > By test developers I mean @os-autoinst/tests-maintainer and others who from time to time have to figure out why a job is failing, when the same job passed minutes earlier... only to figure out that it's due to a released snapshot or a released maintenance update
> 
> Why not just gracefully walk over non-existant repos within the test code?

That's a step further, since one might end up hitting the retry button on the webui :)

okurz:
We were discussing in the QE tools team and I am sure we have found a good suggestion now. openQA already has the feature to check on asset existance and provide feedback to users, e.g. preventing a restart but allowing to force a restart when users still want and similar for calls to openqa-clone-job. But here obviously it is about external repo URLs which are not synced to openQA and do not need to be synced because each incident repository can be considered a consistent state at least as long as it exists. If you would like it to stay around for longer of course, same as we do for other assets, you could sync it to openQA. But let's start with making repository URLs probable to check if they exist and are accessible before starting the actual test. For this what one could do is following http://open.qa/docs/#_specifying_assets_required_by_a_job) extending the concept of URL treatment, e.g. introduce a special suffix _CHECK_URL and check each URL (not download it) in the setup phase and only start tests if probing is successful. The probing can not ensure that the repository stays around until the test does not rely on it anymore but your original approach was similar in that regard.

lib/OpenQA/Script/CloneJob.pm
```diff
@@ -241,10 +241,37 @@ sub clone_jobs ($jobid, $options) {
...
+    my %test_template = ('BASE_TEST_TEMPLATE' => 'BASE_TEST_ISSUES', 'DESKTOP_TEST_TEMPLATE' => 'DESKTOP_TEST_ISSUES', 'SDK_TEST_TEMPLATE' => 'SDK_TEST_ISSUES', 'WE_TEST_TEMPLATE' => 'WE_TEST_ISSUES');
+    my @list_url ;
+    for ( keys %test_template) {
+       next  unless ($settings{$_} or  $settings{$test_template{$_}});
+       my $test_template = $settings{$_};
+       my $ua = LWP::UserAgent->new;
```
alvarocarvajald, line 251: I do not think there is a need to create a new UA for each cycle in the loop. I would move L251 and L252 to somewhere between L245 and L247.

okurz:
@DeepthiYV moreover on a general note: Members from the team are happy to help and I support that they also join you and others in more close collaboration where seen more efficient. So it's fine if you just want to research, dig deeper and try yourself. But at any point you can also ask us for something like a chat with rapid feedback time, pair-programming sessions or code walkthrough, e.g. in a video call.

okurz:
https://progress.opensuse.org/issues/112349 about "Provide an option to always clone the latest version of the scenario run by the job" is related to what we suggested here.

lib/OpenQA/Script/CloneJob.pm
```diff
@@ -241,10 +241,38 @@ sub clone_jobs ($jobid, $options) {
...
+                if (!$response->is_success) {
+                    push @maintenance_list, $test_template_url;
+                }
+            }
+        }
+    }
+
+    if (@maintenance_list) {
+        print "Current job will fail, because the below updates have already been released\n", pp(@maintenance_list);
+    }
```
foursixnine, line 265: @DeepthiYV make it die at the end, there's no need to continue if we alrready know it's gonna fail :), update the ticket accordinglly

okurz:
Hi @DeepthiYV, I saw that you made some minor changes. If you apply changes and it's not obvious which parts you addressed and which you still need help with to provide an explicit message stating what you updated since the last review state but also to state the open points. Then we could more easily help you with that as well. Please understand that only patching little things will likely not make the pull request merging. We need a design change. See https://github.com/os-autoinst/openQA/pull/4693#issuecomment-1154964288 . We must not have openQA care about SUSE SLE maintenance updates in particular, that would be bad design. There are several ways to address this.

DeepthiYV:
> Hi @DeepthiYV, I saw that you made some minor changes. If you apply changes and it's not obvious which parts you addressed and which you still need help with to provide an explicit message stating what you updated since the last review state but also to state the open points. Then we could more easily help you with that as well. Please understand that only patching little things will likely not make the pull request merging. We need a design change. See [#4693 (comment)](https://github.com/os-autoinst/openQA/pull/4693#issuecomment-1154964288) . We must not have openQA care about SUSE SLE maintenance updates in particular, that would be bad design. There are several ways to address this.

Thanks for the feedback. I think I’ll need your support for these changes . Can we have a short call to discuss this.

lib/OpenQA/Script/CloneJob.pm
DeepthiYV: correction is done. Thanks for the heads up.

lib/OpenQA/Script/CloneJob.pm
DeepthiYV: Modified the code as suggested. Thanks :)

lib/OpenQA/Script/CloneJob.pm
DeepthiYV: Yes, that’s a good idea. I have fixed it now.

okurz (changes requested):
I think we should not accept SLE specific code into the main functionality flow of the client or clone-job. Also no bigger code blocks without automated tests. Going forward with a call to discuss details is one possibility to do. At the very least although not the best solution could be as simple as a separate file like lib/OpenQA/Script/CloneJobSUSE.pm that is loaded from CloneJob. I would very much prefer what I suggested in https://github.com/os-autoinst/openQA/pull/4693#issuecomment-1154964288. If you want to go ahead with that I suggest to look into the existing code that handles `_DECOMPRESS_URL`. That could be a good starting point to implement something like the `_CHECK_URL` functionality I mentioned.

DeepthiYV:
> I think we should not accept SLE specific code into the main functionality flow of the client or clone-job. Also no bigger code blocks without automated tests. Going forward with a call to discuss details is one possibility to do. At the very least although not the best solution could be as simple as a separate file like lib/OpenQA/Script/CloneJobSUSE.pm that is loaded from CloneJob. I would very much prefer what I suggested in [#4693 (comment)](https://github.com/os-autoinst/openQA/pull/4693#issuecomment-1154964288). If you want to go ahead with that I suggest to look into the existing code that handles `_DECOMPRESS_URL`. That could be a good starting point to implement something like the `_CHECK_URL` functionality I mentioned.

I will start working with the solution: Loading a separate file from Clone Job. 
Other suggested solution on the lines of "_DECOMPRESS_URL" looks complex.

okurz:
ok, I strongly suggest using test driven development to do further implementation

lib/OpenQA/Script/CloneJobSUSE.pm
```diff
@@ -0,0 +1,52 @@
...
+            }
+        }
+
+    }
+    return @incident_urls;
+}
+
+
+sub _verify_maintenance_url {
+    my @urls = @_;
```
okurz, line 41:
```suggestion
sub _verify_maintenance_url (@urls) {
```

lib/OpenQA/Script/CloneJobSUSE.pm
```diff
@@ -0,0 +1,52 @@
...
+use Exporter 'import';
+use LWP::UserAgent;
+
+
+our @EXPORT = qw(
+  collect_incident_repos
+  detect_maintenance_update
+);
+
+sub collect_incident_repos ($settings) {
```
mimi1vx, line 18: This can work correctly now only for aggregate jobs right ?

mimi1vx (changes requested):
This will not work as excepted. It can also kill jobs which without released updates and also incident jobs have different logic.

lib/OpenQA/Script/CloneJobSUSE.pm
mimi1vx: + be aware of https://progress.opensuse.org/issues/111710

lib/OpenQA/Script/CloneJobSUSE.pm
```diff
@@ -0,0 +1,52 @@
...
+    my @incident_urls;
+    my $ua = LWP::UserAgent->new;
```
Martchus, line 20: It would be nice to re-use the existing `$ua` here (instead of creating a new one every time this point is reached), even though it means passing down the object a few functions down. Then you also don't need to care about setting the timeout here.

t/35-script_clone_job_suse.t
```diff
@@ -0,0 +1,40 @@
...
+    has is_success => 1;
+}
+
+subtest 'maintenance update detect' => sub {
+    my $job_id = 1;
+    my %job = (
+        id => $job_id,
+        BASE_TEST_TEMPLATE => 'http://foo/@INCIDENTNR@/SLE-Module-Basesystem/',
+        BASE_TEST_ISSUES => '24266,24314,24379,24392,24404',
+
```
Martchus, line 28:
```suggestion
```

t/35-script_clone_job_suse.t
```diff
@@ -0,0 +1,40 @@
...
+    );
+    my %response = ('is_success' => 1,);
+    my $clone_mock = Test::MockModule->new('LWP::UserAgent');
+    $clone_mock->mock(get => sub { return Test::FakeLWPUserAgentMirrorResult->new(\%response) });
+    ok(detect_maintenance_update(\%job), 'Maintenance updates are available');
+    $response{is_success} = 0;
+    throws_ok { detect_maintenance_update(\%job) } qr/Current job will fail.*/,
+      'Maintenance updates have been released';
+
```
Martchus, line 37:
```suggestion
```

t/35-script_clone_job_suse.t
```diff
@@ -0,0 +1,40 @@
...
+    my $job_id = 1;
+    my %job = (
+        id => $job_id,
+        BASE_TEST_TEMPLATE => 'http://foo/@INCIDENTNR@/SLE-Module-Basesystem/',
+        BASE_TEST_ISSUES => '24266,24314,24379,24392,24404',
+
+    );
+    my %response = ('is_success' => 1,);
+    my $clone_mock = Test::MockModule->new('LWP::UserAgent');
+    $clone_mock->mock(get => sub { return Test::FakeLWPUserAgentMirrorResult->new(\%response) });
```
Martchus, line 32:
```suggestion
    $clone_mock->redefine(get => sub { return Test::FakeLWPUserAgentMirrorResult->new(\%response) });
```

Martchus (changes requested):

lib/OpenQA/Script/CloneJob.pm
```diff
@@ -241,12 +243,15 @@ sub clone_jobs ($jobid, $options) {
...
 
+
 sub clone_job ($jobid, $url_handler, $options, $post_params = {}, $jobs = {}, $depth = 1, $relation = '') {
     return $post_params if defined $post_params->{$jobid};
 
     my $job = $jobs->{$jobid} = clone_job_get_job($jobid, $url_handler, $options);
+
     my $settings = $post_params->{$jobid} = {%{$job->{settings}}};
 
+    OpenQA::Script::CloneJobSUSE::detect_maintenance_update($settings);
```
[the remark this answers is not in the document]
foursixnine: > > ok, I strongly suggest using test driven development to do further implementation
> 
> I had done changes in separate file CloneJobSUSE.pm that is loaded from CloneJob with unit test. I verified the changes on few jobs of O3 and OSD. Kindly review and provide your feedback.

based on https://github.com/os-autoinst/openQA/pull/4693#discussion_r920140743 and the quote, the flag is not needed (also adding a flag defeats the initial purpose, as I would prefer no relying on qa engineers adding a parameter if it's a SLES Maintenance update or not...)

lib/OpenQA/Script/CloneJob.pm
```diff
@@ -140,6 +141,7 @@ sub clone_job_download_assets ($jobid, $job, $url_handler, $options) {
 
             print STDERR "downloading\n$from\nto\n$dst\n";
             my $r = $ua->mirror($from, $dst);
+            print "Debugggg: $r\n";
```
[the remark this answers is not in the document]
DeepthiYV: I removed the line.

lib/OpenQA/Script/CloneJobSUSE.pm
```diff
@@ -0,0 +1,52 @@
...
+
+use Mojo::Base -strict, -signatures;
+
+use Data::Dump 'pp';
+use Exporter 'import';
+use LWP::UserAgent;
+
+
+our @EXPORT = qw(
+  collect_incident_repos
```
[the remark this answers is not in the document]
DeepthiYV: Corrected. Thanks

lib/OpenQA/Script/CloneJobSUSE.pm
```diff
@@ -0,0 +1,52 @@
+# Copyright 2018-2021 SUSE LLC
+# SPDX-License-Identifier: GPL-2.0-or-later
+
+package OpenQA::Script::CloneJobSUSE;
+
+use Mojo::Base -strict, -signatures;
+
+use Data::Dump 'pp';
```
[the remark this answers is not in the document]
DeepthiYV: Here I removed the empty lines. Are you suggesting not  to use modules?

lib/OpenQA/Script/CloneJobSUSE.pm
```diff
@@ -0,0 +1,52 @@
+# Copyright 2018-2021 SUSE LLC
```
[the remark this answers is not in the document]
DeepthiYV: Updated :)

lib/OpenQA/Script/CloneJobSUSE.pm
DeepthiYV: As I understand, I can look up MAINT_TEST_REPO and INCIDENT_REPO only after this [111710](https://progress.opensuse.org/issues/111710) implementation. Instead of killing job I have modified the code to print the list of maintenance updates that have been released.

lib/OpenQA/Script/CloneJobSUSE.pm
DeepthiYV: I Modified the code as suggested. Thanks.

lib/OpenQA/Script/CloneJobSUSE.pm
```diff
@@ -0,0 +1,52 @@
...
+                $test_template_url =~ s/\@INCIDENTNR\@/$incident/;
+                my $response = $ua->get($test_template_url);
+                if (!$response->is_success) {
+                    push @incident_urls, $test_template_url;
+                }
+            }
+        }
+
+    }
+    return @incident_urls;
```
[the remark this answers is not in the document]
DeepthiYV: Updated as suggested.

t/35-script_clone_job_suse.t
DeepthiYV: I have modified the test. Please let me know if this is okay? Thank you.

lib/OpenQA/Script/CloneJobSUSE.pm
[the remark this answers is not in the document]
pdostal: I think the suggestion is to remove just the `Data::Dump` module.

lib/OpenQA/Script/CloneJobSUSE.pm
```diff
@@ -0,0 +1,42 @@
...
+        }
+
+    }
+    return \@incident_urls;
+}
+
+sub detect_maintenance_update ($jobid, $url_handler, $settings) {
+    my $urls = collect_incident_repos($url_handler, $settings);
+    if (@$urls) {
+        print STDERR "Current job $jobid will fail, because the below updates have already been released\n", pp($urls);
```
pdostal, line 36:
```suggestion
        print STDERR "Current job $jobid will fail, because the below updates are unavailable.\n", pp($urls);
```

lib/OpenQA/Script/CloneJob.pm
```diff
@@ -245,8 +246,9 @@ sub clone_job ($jobid, $url_handler, $options, $post_params = {}, $jobs = {}, $d
     return $post_params if defined $post_params->{$jobid};
 
     my $job = $jobs->{$jobid} = clone_job_get_job($jobid, $url_handler, $options);
-    my $settings = $post_params->{$jobid} = {%{$job->{settings}}};
 
+    my $settings = $post_params->{$jobid} = {%{$job->{settings}}};
+    OpenQA::Script::CloneJobSUSE::detect_maintenance_update($jobid, $url_handler, $settings);
```
kalikiana, line 251: I thought it was discussed that we don't want SUSE-specific code and we had a more generic proposal already 🤔

foursixnine: See: https://github.com/os-autoinst/openQA/pull/4693#discussion_r920283603 (save @mimi1vx's comment on which variables to use)

kalikiana: Not sure how that relates to my question. I was asking why we're not solving this in a generic way. I recall a conversation about a concrete implementation. Maybe @Martchus or @okurz knows

foursixnine: @kalikiana can you explain better what generic way you're talking about? comment above states: The code works for osd and o3, without need for any modifications.

lib/OpenQA/Script/CloneJobSUSE.pm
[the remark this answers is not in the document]
okurz: I overlooked that you actually use `pp`, nevermind, keep as is

lib/OpenQA/Script/CloneJobSUSE.pm
okurz: How about

```suggestion
        die "Current job $jobid will fail, because the below updates have already been released\n" . pp($urls);
```

lib/OpenQA/Script/CloneJobSUSE.pm
```diff
@@ -0,0 +1,42 @@
...
+        return 0;
+    }
+    return 1;
```
okurz, line 39: How about

```suggestion
    die "Current job $jobid will fail, because the below updates have already been released\n" . pp($urls) if @$urls;
```

okurz (changes requested):

lib/OpenQA/Script/CloneJob.pm
kalikiana: @foursixnine To be honest I don't see any agreement in this PR, just a period of silence between the generic proposal and what's now implemented. If I was to guess there must have beem an out of band conversation that I was not part of.

> on one hand it's only checking for special test variables which shouldn't bother anyone not using those variables and on the other hand it can serve as a starting point to implement nicer plugins

I agree it doesn't hurt. I don't see the use case for nice plugins, though, if we think we can fully replace it with a generic solution. So if others are fine with it, fine, to me it seems like we'll duplicate the effort and have additional migrations following this.

kalikiana: Btw just because I don't understand the rationale doesn't mean it's a bad idea. I'm just explaining my perspective here because I was surprised.

lib/OpenQA/Script/CloneJobSUSE.pm
alvarocarvajald: Seconded. You are not doing anything with those return codes; better go with `die` as @okurz is suggesting and you also avoid `print STDERR`.

okurz: Well, maybe we need to collect experiences from real-live operations to know for sure. I am ok to leave as is for now.

lib/OpenQA/Script/CloneJob.pm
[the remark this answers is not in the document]
Martchus: Considering the additional API requests are only done if certain test settings are present I suppose we can live with that.

lib/OpenQA/Script/CloneJob.pm
Martchus: > do you agree that we accept this PR after the smaller discussions are resolved?

Ok. I've also seen that most of my review comments have already been resolved. I'll re-review the PR later.

lib/OpenQA/Script/CloneJobSUSE.pm
```diff
@@ -0,0 +1,38 @@
...
+            }
+        }
+
+    }
+    return \@incident_urls;
+}
+
+sub detect_maintenance_update ($jobid, $url_handler, $settings) {
+    my $urls = collect_incident_repos($url_handler, $settings);
+    die "Current job $jobid will fail, because the below updates have already been released\n" . pp($urls) if @$urls;
```
foursixnine, line 35:
```suggestion
    die "Current job $jobid will fail, because the below updates may have already been released or are unavailable\n" . pp($urls) if @$urls;
```

okurz:
@DeepthiYV as soon as you have implemented https://github.com/os-autoinst/openQA/pull/4693#discussion_r924426323 and do not plan to do more changes please mark the PR as ready for review as it's still in draft state

okurz (approved):

t/35-script_clone_job_suse.t
```diff
@@ -0,0 +1,52 @@
...
+    package Test::FakeLWPUserAgent;
+    use Mojo::Base -base, -signatures;
+    has is_validrepo => 1;
+    sub get ($self, $url) {
+        my %res
+          = (!$self->is_validrepo)
+          ? (is_success => 0)
+          : ();
+        Test::FakeLWPUserAgentMirrorResult->new(%res);
+    }
```
Martchus, line 32:
```suggestion
    sub get ($self, $url) { Test::FakeLWPUserAgentMirrorResult->new(is_success => $self->is_validrepo) }
```

lib/OpenQA/Script/CloneJobSUSE.pm
```diff
@@ -0,0 +1,39 @@
...
+            foreach my $incident (split(/,/, $settings->{$1 . 'TEST_ISSUES'})) {
+                my $test_template_url = $settings->{$setting};
+                $test_template_url =~ s/\@INCIDENTNR\@/$incident/;
+                my $response = $ua->get($test_template_url);
+                if (!$response->is_success) {
+                    push @incident_urls, $test_template_url;
+                }
+            }
+        }
+
```
Martchus, line 28:
```suggestion
        next unless $setting =~ /(.*)TEST_TEMPLATE/;
        next unless $settings->{$1 . 'TEST_ISSUES'};
        foreach my $incident (split(/,/, $settings->{$1 . 'TEST_ISSUES'})) {
            my $test_template_url = $settings->{$setting};
            $test_template_url =~ s/\@INCIDENTNR\@/$incident/;
            push @incident_urls, $test_template_url unless $$ua->get($test_template_url)->is_success;
        }
```

lib/OpenQA/Script/CloneJobSUSE.pm
```diff
@@ -0,0 +1,39 @@
...
+# SPDX-License-Identifier: GPL-2.0-or-later
+
+package OpenQA::Script::CloneJobSUSE;
+use Mojo::Base -strict, -signatures;
+use Data::Dump 'pp';
+use Exporter 'import';
+
+our @EXPORT = qw(
+  detect_maintenance_update
+);
```
Martchus, line 11:
```suggestion
our @EXPORT = qw(detect_maintenance_update);
```

t/35-script_clone_job_suse.t
```diff
@@ -0,0 +1,52 @@
...
+use OpenQA::Script::CloneJobSUSE;
+use Mojo::URL;
+
+# define fake client
+{
+    package Test::FakeLWPUserAgentMirrorResult;
+    use Mojo::Base -base, -signatures;
+    has is_success => 1;
+}
+
```
Martchus, line 21:
```suggestion
```

Martchus (changes requested):
I suppose it would make sense to write much of the code in a shorter way, see suggestions.

foursixnine (approved):

t/35-script_clone_job_suse.t
DeepthiYV: Modified the code as suggested. Thank you.

Martchus (approved):

kalikiana:
@mimi1vx I think this is still pending your review

mimi1vx:
still same problem ... using `*_TEST_TEMPLATE` and ` _TEST_ISSUE` variables ... its problem. Really used is only part of this because they aree used to compute `MAINT_TEST_REPO` var using `SCC_ADDONS` but only in aggregate jobs, for incidents there is `INCIDENT_REPO` var

DeepthiYV:
> still same problem ... using `*_TEST_TEMPLATE` and ` _TEST_ISSUE` variables ... its problem. Really used is only part of this because they aree used to compute `MAINT_TEST_REPO` var using `SCC_ADDONS` but only in aggregate jobs, for incidents there is `INCIDENT_REPO` var
@mimi1vx, I think, we can make this changes only after the implementation of  the ticket https://progress.opensuse.org/issues/111710.  @foursixnine and I were already discussed and we will start working on this ticket from next week.

mimi1vx:
> @mimi1vx, I think, we can make this changes only after the implementation of the ticket https://progress.opensuse.org/issues/111710. @foursixnine and I were already discussed and we will start working on this ticket from next week.

no described problem is real now, even without poo#11710 and after implementation of this poo it will be even worse.
by using correct vars:

1) there isn't need to reconstruct URL from ISSUE + template because *MAINT_TEST_REPO* and *INCIDENT_REPO* contains this URL's
2) you will be checking only valid incidents for given job, not all possible incidents
3) it will be  future-proof

current implementation is simply wrong

DeepthiYV:
I discussed this with @mimi1vx . I liked his suggestion. I shall first implement the ticket [111710](https://progress.opensuse.org/issues/111710)  and then I will make changes to this PR.

DeepthiYV:
@mimi1vx  Can you please review my latest changes.

t/35-script_clone_job_suse.t
```diff
@@ -0,0 +1,54 @@
...
+    has is_validrepo => 1;
+    sub get ($self, $url) { Test::FakeLWPUserAgentMirrorResult->new(is_success => $self->is_validrepo) }
+}
+
+subtest 'maintenance update detect' => sub {
+    my $job_id = 1;
+    my %job = (
+        id => $job_id,
+        SCC_ADDONS => "we,sdk",
+        WE_TEST_REPOS => "http://foo/WE_suse/openqa, http://foo/WE_suse_2/openqa",
```
mimi1vx, line 33: in wars there aren't spaces between URL's

mimi1vx (approved):
please check if it works with proper format of vars :D but else LGTM

DeepthiYV:
> please check if it works with proper format of vars :D but else LGTM

I have updated the few VR output in the ticket :)

pdostal (approved):

foursixnine:
Removing the not-ready label as all the concerns have been addressed

PR #4696, opened 2022-06-09
Avoid race condition in Selenium test helper `element_hidden`

Martchus:
The ticket https://progress.opensuse.org/issues/111920 shows that there's
apparently a race condition.

The JavaScript code tested here seems unproblematic. Apparently Selenium's
displayedness check considers the viewport (see
https://w3c.github.io/webdriver/#element-displayedness). I suspect that
this can lead to the test failure we observed because Selenium only waits
until the JavaScript event-loop runs out of work but likely *not* until the
viewport has been updated.


okurz (approved):

Martchus:
> … because Selenium only waits until the JavaScript event-loop runs out of work …

By the way, I've tested that claim via https://github.com/os-autoinst/openQA/commit/beec87da28ab605a34ee3ff56016262613f401a2 and it is true. So there's a 10 seconds delay between "before click" and "after click". That means relying on the JavaScript being processed should be fine. Hence I've figured that it must then be the view port update Selenium's displayedness check relies on.

PR #4700, opened 2022-06-13
Avoid running out of database IDs

Martchus:
* Convert all database ID-columns from integer to bigint
* Convert all related sequences as well (the sequences are already bigint
  on our production instances but that is not necessarily the case, e.g. on
  my local instance they were still integer)
* Adapt code that specifies the types explicitly
* See https://progress.opensuse.org/issues/112265

### DIFF
--- a/dbicdh/PostgreSQL/deploy/95/001-auto-__VERSION.sql
+++ b/dbicdh/PostgreSQL/deploy/95/001-auto-__VERSION.sql
@@ -1,0 +1,18 @@
+--
+-- Created by SQL::Translator::Producer::PostgreSQL
+-- Created on Mon Jun 13 10:16:51 2022
+--
+;
+--
+-- Table: dbix_class_deploymenthandler_versions
+--
+CREATE TABLE dbix_class_deploymenthandler_versions (
+  id serial NOT NULL,
+  version character varying(50) NOT NULL,
+  ddl text,
+  upgrade_sql text,
+  PRIMARY KEY (id),
+  CONSTRAINT dbix_class_deploymenthandler_versions_version UNIQUE (version)
+);
+
+;
--- a/dbicdh/PostgreSQL/deploy/95/001-auto.sql
+++ b/dbicdh/PostgreSQL/deploy/95/001-auto.sql
@@ -1,0 +1,794 @@
+--
+-- Created by SQL::Translator::Producer::PostgreSQL
+-- Created on Mon Jun 13 10:16:51 2022
+--
+;
+--
+-- Table: bugs
+--
+CREATE TABLE bugs (
+  id bigserial NOT NULL,
+  bugid text NOT NULL,
+  title text,
+  priority text,
+  assigned boolean,
+  assignee text,
+  open boolean,
+  status text,
+  resolution text,
+  existing boolean DEFAULT '1' NOT NULL,
+  refreshed boolean DEFAULT '0' NOT NULL,
+  t_created timestamp NOT NULL,
+  t_updated timestamp NOT NULL,
+  PRIMARY KEY (id),
+  CONSTRAINT bugs_bugid UNIQUE (bugid)
+);
+
+;
+--
+-- Table: gru_tasks
+--
+CREATE TABLE gru_tasks (
+  id bigserial NOT NULL,
+  taskname text NOT NULL,
+  args text NOT NULL,
+  run_at timestamp NOT NULL,
+  priority integer NOT NULL,
+  t_created timestamp NOT NULL,
+  t_updated timestamp NOT NULL,
+  PRIMARY KEY (id)
+);
+CREATE INDEX gru_tasks_run_at_reversed on gru_tasks (run_at DESC);
+
+;
+--
+-- Table: job_group_parents
+--
+CREATE TABLE job_group_parents (
+  id bigserial NOT NULL,
+  name text NOT NULL,
+  size_limit_gb integer,
+  exclusively_kept_asset_size bigint,
+  default_keep_logs_in_days integer,
+  default_keep_important_logs_in_days integer,
+  default_keep_results_in_days integer,
+  default_keep_important_results_in_days integer,
+  default_priority integer,
+  sort_order integer,
+  description text,
+  build_version_sort boolean DEFAULT '1' NOT NULL,
+  carry_over_bugrefs boolean,
+  t_created timestamp NOT NULL,
+  t_updated timestamp NOT NULL,
+  PRIMARY KEY (id),
+  CONSTRAINT job_group_parents_name UNIQUE (name)
+);
+
+;
+--
+-- Table: job_modules
+--
+CREATE TABLE job_modules (
+  id bigserial NOT NULL,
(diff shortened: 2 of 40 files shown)


kraih (approved):

Martchus:
Maybe we should think of a suitable timeframe for merging and deploying this change. I've just been starting the migration with OSD's database on my very slow laptop HDD and it has been running for 15 minutes now. It'll likely be much faster on OSD but it would make sense to run it during the night or whenever we have low load.

Martchus:
It has just finished. It utilized one CPU core and considering my HDD was very unresponsive it was likely also utilizing my HDD a lot (with read/write speeds varying a lot but sometimes going over 100 MiB/s). I suppose was effectively rewriting the whole database and that took 2 hours. Since my HDD and overall system is likely slower I don't expect it taking that long on OSD but it will still be a considerable downtime. Otherwise the migrated database seems to work as expected.

I suppose it would be possible to split the PR but I don't think there's much to gain because it cannot be split into very tiny peaces. There would still be the big tables and many tables have foreign keys which need to be converted in one go with the corresponding primary keys. The jobs table is for instance one of the biggest tables and it would pull at least 13 other tables as they're referencing job IDs, for instance the assets table, job settings and job modules tables which are also big ones.

Martchus:
That would be a version with the jobs table and related tables only: https://github.com/os-autoinst/openQA/pull/4701

I've tested it locally and it seems to work. I only need to test how long the OSD migration would take again.

Martchus:
> such a long migration time at once can be problematic, not just for OSD. I suggest to split the PR and do little changes at a time regardless if there are individual much slower migrations.

Since you specifically mention "not just for OSD" but suggest to split I'd like to mention that splitting won't help users that don't update their instance very often (as they will likely still receive all updates at once unless we drag it out very long).

kraih:
I'd just do it all at once on Thursday during the maintenance window (or over the weekend).

perlpunk (approved):

okurz:
> I'd just do it all at once on Thursday during the maintenance window (or over the weekend). Don't see how splitting it up into multiple even slower migrations would help anyone.

Please keep in mind that there are more instances than just the ones we maintain ourselves. And the deployment is continuous and automatic on o3 so it would automatically start as soon as the PR is merged. That's a feature not a bug because it makes us develop more stable code and forces us to think how to conduct such migration gracefully :)

Martchus:
> Please keep in mind that there are more instances than just the ones we maintain ourselves.

Again, splitting won't help with that as most other users (as far as I know) don't update very frequently. Additionally, other instances have hopefully smaller DBs and less users depending on them so the problem shouldn't be so severe. Maybe it is time for a heads-up to @AdamWill and @phil-hands, though.

> And the deployment is continuous and automatic on o3 so it would automatically start as soon as the PR is merged.

We can always temporarily pause that, right?

> and forces us to think how to conduct such migration gracefully

I wouldn't know how to improve this. It seems it needs to re-write almost all rows.

okurz:
> > Please keep in mind that there are more instances than just the ones we maintain ourselves.
> 
> Again, splitting won't help with that as most other users (as far as I know) don't update very frequently. Additionally, other instances have hopefully smaller DBs and less users depending on them so the problem shouldn't be so severe. Maybe it is time for a heads-up to @AdamWill and @phil-hands, though.
> 
> > And the deployment is continuous and automatic on o3 so it would automatically start as soon as the PR is merged.
> 
> We can always temporarily pause that, right?

Yes to both. But we should assume that there is a hug instance out there that we don't know about and we can't inform their admins about the pending migration. And then we should design our applications accordingly.

How about one or multiple of the following ideas:
1. Explicitly failing the migration with an explicit suggestion how to migrate at a convenient time
2. Making the column type configurable defaulting to bigint + a migration script that can be explicitly triggered
3. Dynamically detecting a limit of rows and only conducting an automatic migration if below a certain limit, e.g. if rows < 100k, otherwise go to step 1. or 2.

Martchus:
I'm not sure whether DBIx will behave correctly in all cases if the data type of a column in the database doesn't match the datatype DBIx thinks it has. There's also manual SQL code which needed to handle both cases explicitly. So I don't think we do us a favor with supporting multiple database schemas at the same time (which all of your suggestions boil down to).

Especially 1. seems useless because one can simply postpone updating at all until it is convenient. We can print a warning before starting the migration that is might take long and if one is impatient one can just abort and rollback the openQA version until one has time.

AdamWill:
We (Fedora) have two significant instances, our prod and staging deployments. There's also at least one other team inside Red Hat with a significant deployment, I don't know exactly how busy that one is but I can find out. Fedora's prod instance has run just under 1.3m jobs, the staging instance around 1.86m, if that gives an idea of scale.

AdamWill:
As to the discussion - one can only postpone updating if one is aware of this migration ahead of time. AIUI the value of @okurz 's suggestion is it makes sure an admin makes a conscious decision to trigger the migration. You say "we can print a warning before starting the migration", but are you sure you can? What exactly triggers the beginning migration? What if it's not a human admin visiting a page, but a worker checking in, or a script scheduling some jobs, or something?

Martchus:
> What exactly triggers the migration?

The next restart of any of the openQA services on the web UI host once the new package version is installed.

> What if it's not a human admin visiting a page, but a worker checking in, or a script scheduling some jobs, or something?

If it isn't a human visiting nobody sees the warning until the service's journal is checked (when people are already complaining about the downtime). Then one could decide to abort (stop the service and install the previous version) or continue.

okurz:
> As to the discussion - one can only postpone updating if one is aware of this migration ahead of time. AIUI the value of @okurz 's suggestion is it makes sure an admin makes a conscious decision to trigger the migration

Yes, that would be my intention.

> > What exactly triggers the migration?
> 
> The next restart of any of the openQA services on the web UI host once the new package version is installed.

which is commonly triggered automatically by the package upgrade itself, at least for https://github.com/os-autoinst/openQA/blob/master/dist/rpm/openQA.spec#L448 it is.


@Martchus 
> I'm not sure whether DBIx will behave correctly in all cases if the data type of a column in the database doesn't match the datatype DBIx thinks it has. There's also manual SQL code which needed to handle both cases explicitly. So I don't think we do us a favor with supporting multiple database schemas at the same time (which all of your suggestions boil down to).
> 
> Especially 1. seems useless because one can simply postpone updating at all until it is convenient (or is that what you meant?). We can print a warning before starting the migration that it might take long and if one is impatient one can just abort and rollback the openQA version until one has time.

1. is not useless as AdamWill described as I am assuming the best practice for a continuously released package like openQA to be also continuously deployed hence one would only explicitly stop an upgrade with implicit service restart and migration if they know about it and I am suggest as "communication channel" to use the software itself to only conduct a migration under safe circumstances and ask for manual confirmation in other cases.

Martchus:
I would then merge https://github.com/os-autoinst/openQA/pull/4701 first in the next days and I have added a warning before the migration.

okurz:
Likely obsoleted by https://github.com/os-autoinst/openQA/pull/4705

PR #4701, opened 2022-06-13
Avoid running out of database IDs - jobs table

Martchus:
* Convert all database ID-columns of the jobs table and related tables
  from integer to bigint
* Adapt code that specifies the types explicitly
* See https://progress.opensuse.org/issues/112265

### DIFF
--- a/dbicdh/PostgreSQL/deploy/95/001-auto-__VERSION.sql
+++ b/dbicdh/PostgreSQL/deploy/95/001-auto-__VERSION.sql
@@ -1,0 +1,18 @@
+--
+-- Created by SQL::Translator::Producer::PostgreSQL
+-- Created on Mon Jun 13 14:25:33 2022
+--
+;
+--
+-- Table: dbix_class_deploymenthandler_versions
+--
+CREATE TABLE dbix_class_deploymenthandler_versions (
+  id serial NOT NULL,
+  version character varying(50) NOT NULL,
+  ddl text,
+  upgrade_sql text,
+  PRIMARY KEY (id),
+  CONSTRAINT dbix_class_deploymenthandler_versions_version UNIQUE (version)
+);
+
+;
--- a/dbicdh/PostgreSQL/deploy/95/001-auto.sql
+++ b/dbicdh/PostgreSQL/deploy/95/001-auto.sql
@@ -1,0 +1,794 @@
+--
+-- Created by SQL::Translator::Producer::PostgreSQL
+-- Created on Mon Jun 13 14:25:32 2022
+--
+;
+--
+-- Table: bugs
+--
+CREATE TABLE bugs (
+  id serial NOT NULL,
+  bugid text NOT NULL,
+  title text,
+  priority text,
+  assigned boolean,
+  assignee text,
+  open boolean,
+  status text,
+  resolution text,
+  existing boolean DEFAULT '1' NOT NULL,
+  refreshed boolean DEFAULT '0' NOT NULL,
+  t_created timestamp NOT NULL,
+  t_updated timestamp NOT NULL,
+  PRIMARY KEY (id),
+  CONSTRAINT bugs_bugid UNIQUE (bugid)
+);
+
+;
+--
+-- Table: gru_tasks
+--
+CREATE TABLE gru_tasks (
+  id serial NOT NULL,
+  taskname text NOT NULL,
+  args text NOT NULL,
+  run_at timestamp NOT NULL,
+  priority integer NOT NULL,
+  t_created timestamp NOT NULL,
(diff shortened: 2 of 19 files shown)


kraih (approved):

okurz (approved):

Martchus:
This took now around an hour. So one hour less compared to the full migration. Maybe it makes actually sense to do it in two parts then. At least our own setups would benefit as they're updated very regularly.

kraih:
Still struggling to see the benefit, it's going to require a maintenance window either way.

okurz:
> This took now around an hour. So one hour less compared to the full migration. Maybe it makes actually sense to do it in two parts then.

you mean it makes sense how to solve it how you do it with this separate PR?

okurz:
> Still struggling to see the benefit, it's going to require a maintenance window either way.

I fear so. But we don't have an easy way to communicate to all users and customers how to prepare for such migration. Maybe we can come up with a better idea to make the automatic migration more smooth?

Martchus:
> you mean it makes sense how to solve it how you do it with this separate PR?

I don't care. We can merge either PR, it is just that merging this one first isn't as beneficial as one might have thought but it indeed reduces the time a little bit. Merging this one first means the remaining migrations still need to be done. Maybe it is better to have just one disturbance (and not also a second one - although the second one will hopefully be faster).

> Maybe we can come up with a better idea to make the automatic migration more smooth?

I wouldn't know how. We could maybe allow read-only access while the migration is running but that doesn't seem worth the effort. We could operate on a copy of the database but loosing all changed made during the migration seems counterproductive. Note that the migration will hopefully be faster on the production machine (compared to my slow laptop HDD).

okurz:
So we decided in daily 2022-06-15 that you just add a warning message that at least admins are informed why the service does not come up and why it takes long. Then admins can either wait or abort the migration and reconduct it at a convenient time.

Then we merge on Friday afternoon for two reasons:
1. To pick a time when it would be rolled out automatically on accordingly configured instances when most users wouldn't care about a bit longer downtime
2. To prove that no kittens are harmed during a Friday-deployment

Martchus:
I've been adding the warning. I tested it locally. I have also aborted several times with different delays and it didn't break anything.

okurz:
@Martchus I sent a message in https://matrix.to/#/!dRljORKAiNJcGEDbYA:opensuse.org/$lmq-ywz1qJp65YrYV8MOQwfH1wNjrEwVJWWVd03rfYM and https://suse.slack.com/archives/C02CANHLANP/p1655289571311949 as head's up. You can send the message to various other channels as preparation if you like.

okurz:
Merging after waiting period. We expect that continuously deployed instances can use the weekend with likely lower load and availability expectations for the migration without significant impact

okurz:
Logs from deployment on o3:/var/log/openqa:

```
[2022-06-18T07:10:42.883202Z] [info] Listening at "http://127.0.0.1:9528"
[2022-06-18T07:10:42.884031Z] [info] Listening at "http://[::1]:9528"
[2022-06-18T07:10:42.934304Z] [info] Listening at "http://127.0.0.1:9527"
[2022-06-18T07:10:42.934999Z] [info] Listening at "http://[::1]:9527"
[2022-06-18T07:10:43.019713Z] [warn] [pid:23714] Job IDs will be converted to bigint. That may take multiple hours on big databases.  It is safe to stop the service (and start from scratch on the next startup).
[2022-06-18T07:15:35.981054Z] [debug] attempt to open ws proxy for job alp-0.1-kvm-and-xen-x86_64-Build17.1-alp_default@64bit (2420945) where URL to os-autoinst command server is unknown
[2022-06-18T07:15:35.999919Z] [info] Loading external plugin AMQP
[2022-06-18T07:15:36.105433Z] [info] Loading external plugin ObsRsync
[2022-06-18T07:15:36.853091Z] [info] Listening at "http://127.0.0.1:9526"
[2022-06-18T07:15:36.853696Z] [info] Listening at "http://[::1]:9526"
[2022-06-18T07:15:36.853824Z] [info] Manager 23714 started
```

so it took 5 minutes to migrate on o3.

kraih:
That was fast. 🥳

Martchus:
Yes, but let's see what time it needs to migrate OSD.

okurz:
> Yes, but let's see what time it needs to migrate OSD.

18m

okurz:
After the initial migration we ran into a bigger problem on one of our instances. Problem was reported in detail in https://progress.opensuse.org/issues/112718. A general suggestion is to run a manual `vacuum ANALYZE` on all changed tables, here in particular "jobs", "job_modules" and "job_settings".

PR #4704, opened 2022-06-15
docs: Fix explanation for TIMEOUT_SCALE

okurz:
TIMEOUT_SCALE is not only affecting MAX_JOB_TIME but test API functions
and other backend specific implementations as well.


perlpunk (approved):

Martchus (approved):

okurz:
Fixed the reference to `<<_api,test API>>` after checking local rendering with asciidoctor. Hope that is correct now.

PR #4710, opened 2022-06-21
Fix perl 5.36 warnings on use of @_ in functions with signature

AdamWill:
As with os-autoinst, building openQA with perl 5.36 results in
test failures due to several cases of subs with signatures using
`@_`, which they are not supposed to do. Fortunately the fixes
are fairly straightforward here.

### DIFF
--- a/lib/OpenQA/Worker/Engines/isotovideo.pm
+++ b/lib/OpenQA/Worker/Engines/isotovideo.pm
@@ -98,11 +98,12 @@ sub _poll_cache_service ($job, $cache_client, $request, $delay, $callback) {
 
 sub cache_assets ($cache_client, $job, $vars, $assets_to_cache, $assetkeys, $webui_host, $pooldir, $callback) {
     return $callback->(undef) unless my $this_asset = shift @$assets_to_cache;
-    return cache_assets(@_) unless my $asset_value = $vars->{$this_asset};
+    return cache_assets($cache_client, $job, $vars, $assets_to_cache, $assetkeys, $webui_host, $pooldir, $callback)
+      unless my $asset_value = $vars->{$this_asset};
 
     my $asset_uri = trim($asset_value);
-    # skip UEFI_PFLASH_VARS asset if the job won't use UEFI
-    return cache_assets(@_) if $this_asset eq 'UEFI_PFLASH_VARS' && !$vars->{UEFI};
+    return cache_assets($cache_client, $job, $vars, $assets_to_cache, $assetkeys, $webui_host, $pooldir, $callback)
+      if ($this_asset eq 'UEFI_PFLASH_VARS' && !$vars->{UEFI});
     # check cache availability
     my $error = $cache_client->info->availability_error;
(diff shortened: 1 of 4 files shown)


Martchus (approved):
Ok, except that tidy is failing.

lib/OpenQA/Worker/Engines/isotovideo.pm
```diff
@@ -98,11 +98,16 @@ sub _poll_cache_service ($job, $cache_client, $request, $delay, $callback) {
 
 sub cache_assets ($cache_client, $job, $vars, $assets_to_cache, $assetkeys, $webui_host, $pooldir, $callback) {
     return $callback->(undef) unless my $this_asset = shift @$assets_to_cache;
-    return cache_assets(@_) unless my $asset_value = $vars->{$this_asset};
+    my $asset_value = $vars->{$this_asset};
+    unless ($asset_value) {
+        return cache_assets($cache_client, $job, $vars, $assets_to_cache, $assetkeys, $webui_host, $pooldir, $callback);
+    }
```
okurz, line 104: not sure why you turned that around. Especially with "unless" we commonly prefer the post-condition form

```suggestion
    return cache_assets($cache_client, $job, $vars, $assets_to_cache, $assetkeys, $webui_host, $pooldir, $callback) unless $asset_value;
```

lib/OpenQA/Worker/Engines/isotovideo.pm
```diff
@@ -98,11 +98,16 @@ sub _poll_cache_service ($job, $cache_client, $request, $delay, $callback) {
...
 
     my $asset_uri = trim($asset_value);
     # skip UEFI_PFLASH_VARS asset if the job won't use UEFI
-    return cache_assets(@_) if $this_asset eq 'UEFI_PFLASH_VARS' && !$vars->{UEFI};
+    if ($this_asset eq 'UEFI_PFLASH_VARS' && !$vars->{UEFI}) {
+        return cache_assets($cache_client, $job, $vars, $assets_to_cache, $assetkeys, $webui_host, $pooldir, $callback);
+    }
```
okurz, line 110: same as above, better keep

```suggestion
    return cache_assets($cache_client, $job, $vars, $assets_to_cache, $assetkeys, $webui_host, $pooldir, $callback if $this_asset eq 'UEFI_PFLASH_VARS' && !$vars->{UEFI};
```

okurz (changes requested):

lib/OpenQA/Worker/Engines/isotovideo.pm
AdamWill: Because it makes the lines super long. But if you don't mind that, sure.

AdamWill:
Okay, returned to super long lines and tidied. tidy just splits the long lines before the conditional and indents the second line a bit which reads kinda awkward to me, but hey.

perlpunk (approved):

Martchus (approved):

okurz (approved):

PR #4711, opened 2022-06-21
Fix warning "Query returned more than one row" for linked comments

okurz:
Related progress issue: https://progress.opensuse.org/issues/112718

### DIFF
--- a/lib/OpenQA/Schema/ResultSet/Jobs.pm
+++ b/lib/OpenQA/Schema/ResultSet/Jobs.pm
@@ -452,7 +452,7 @@ sub mark_job_linked {
     my $job = $self->find({id => $jobid});
     return undef if !$job || ($referer->path_query =~ /^\/?$/);
     my $comments = $job->comments;
-    return undef if $comments->find({text => {like => 'label:linked%'}});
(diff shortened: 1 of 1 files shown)


lib/OpenQA/Schema/ResultSet/Jobs.pm
```diff
@@ -452,7 +452,7 @@ sub mark_job_linked {
     my $job = $self->find({id => $jobid});
     return undef if !$job || ($referer->path_query =~ /^\/?$/);
     my $comments = $job->comments;
-    return undef if $comments->find({text => {like => 'label:linked%'}});
+    return undef if $comments->search({text => {like => 'label:linked%'}}, {rows => 1})->first;
```
Martchus, line 455: When limiting to one row it should be possible to continue using find.

kalikiana: `In addition to key, "find" recognizes and applies standard resultset attributes in the same way as "search" does.` [as per docs](https://metacpan.org/pod/DBIx::Class::ResultSet#find)

perlpunk (approved):
I'm fine with using `search` here. We are searching for comments with a certain text.
`find` is typically for finding something by a unique key.

Martchus (approved):

PR #4717, opened 2022-06-23
Only highlight the effective bugref(s) in the comments

kalikiana:
The latest comment with one or more bugrefs is considered for takeover. Others are effectively ignored. This should be clearly visible in the web UI.

### DIFF
--- a/assets/stylesheets/comments.scss
+++ b/assets/stylesheets/comments.scss
@@ -87,7 +87,7 @@
         text-align: right;
     }
 
-    &:last-child .openqa-bugref {
+    &:last-of-type .openqa-bugref {
         font-weight: bold;
     }
 }
--- a/assets/stylesheets/comments.scss
+++ b/assets/stylesheets/comments.scss
@@ -86,6 +86,10 @@
     .col-sm-1 {
         text-align: right;
     }
+
+    &:last-child .openqa-bugref {
+        font-weight: bold;
+    }
 }
 
 .pagination {
--- a/lib/OpenQA/Markdown.pm
+++ b/lib/OpenQA/Markdown.pm
@@ -5,17 +5,11 @@ use Mojo::Base -strict;
 
 use Exporter 'import';
 use Regexp::Common 'URI';
-use OpenQA::Utils qw(BUGREF_REGEX LABEL_REGEX bugurl);
+use OpenQA::Utils qw(BUGREF_REGEX LABEL_REGEX bugurl bugref_to_href);
 use OpenQA::Constants qw(FRAGMENT_REGEX);
 use CommonMark;
 
-our @EXPORT_OK = qw(bugref_to_markdown is_light_color markdown_to_html);
-
-sub bugref_to_markdown {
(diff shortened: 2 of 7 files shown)


okurz:
Nice idea. Please double check that this is *really* the current behavior, not that we have wrong expectations on what actually happens

Martchus:
It would be nice if the other labels would still be formatted but e.g. just be grayed out (instead of having the yellow background).

kalikiana:
> It would be nice if the other labels would still be formatted but e.g. just be grayed out (instead of having the yellow background).

I realized takeover looks for the most recent **bugref**. It doesn't care about labels and doesn't use this CSS class. So I changed it to actually add a new class `openqa-bugref` and make the bugref bold if it's the latest one, in addition to clarifying the docs.

Martchus:
Ok, I suppose that's good. However, it looks like tests needed to be adjusted.

Martchus (approved):

okurz (commented):
please include a screenshot, rest is fine

assets/stylesheets/comments.scss
```diff
@@ -47,6 +47,10 @@
         }
     }
 
+    .openqa-bugref:last-child {
```
okurz, line 50: After discussing we found out that this would only mark one bugref out of potentially multiple within a single comment but we do carry-over of "comments including bugref(s)"

okurz (changes requested):

assets/stylesheets/comments.scss
kalikiana: I also added a unit test for this now. Since we've been unsure what the exact behavior is more than once.

okurz:
please follow https://github.com/os-autoinst/openQA/pull/4717#issuecomment-1164110960, i.e. screenshot please :)

assets/stylesheets/comments.scss
kalikiana: Turns out my fixes for the markdown version of the same code got lost. That's why the tests wouldn't pass. So I thought maybe we can prevent this and merged both code paths.

Martchus (approved):

kalikiana:
> please follow [#4717 (comment)](https://github.com/os-autoinst/openQA/pull/4717#issuecomment-1164110960), i.e. screenshot please :)

To add context to my requesting re-review. I'm not posting a screenshot so long as it doesn't work as we expect. But I need someone to spot the mistake in either the CSS or the Perl code.

Martchus:
Ah, then I'll try it out locally.

Martchus:
I've been pushing an additional commit that should fix it.

It also works when comments are removed or newly created. However, when editing a commit it doesn't work until one reloads the page because comments are not reordered on the fly when editing.

okurz:
@Mergifyio rebase

okurz (changes requested):
please include a screenshot (I already stated that two times), rest is fine

mimi1vx (approved):

PR #4718, opened 2022-06-23
Retry hook scripts that use the exit code 142

kraih:
This patch turns the previous proof of concept into a proper feature for
making hook scripts restartable with a special exit code. The hook
script now runs in a separate Minion job, and all parameters are
configurable with environment variables and job settings.

Progress: https://progress.opensuse.org/issues/112523

### DIFF
--- a/lib/OpenQA/Task/Job/FinalizeResults.pm
+++ b/lib/OpenQA/Task/Job/FinalizeResults.pm
@@ -42,25 +42,38 @@ sub _finalize_results {
 }
 
 sub _run_hook_script ($minion_job, $openqa_job, $app, $guard) {
-    my $trigger_hook = $openqa_job->settings_hash->{_TRIGGER_JOB_DONE_HOOK};
+    my $settings = $openqa_job->settings_hash;
+    my $trigger_hook = $settings->{_TRIGGER_JOB_DONE_HOOK};
+
     return undef if defined $trigger_hook && !$trigger_hook;
     return undef unless my $result = $openqa_job->result;
+
     my $hooks = $app->config->{hooks};
     my $key = "job_done_hook_$result";
     my $hook = $ENV{'OPENQA_' . uc $key} // $hooks->{lc $key};
     $hook = $hooks->{job_done_hook} if !$hook && ($trigger_hook || $hooks->{"job_done_hook_enable_$result"});
     return undef unless $hook;
+
     my $timeout = $ENV{OPENQA_JOB_DONE_HOOK_TIMEOUT} // '5m';
     my $kill_timeout = $ENV{OPENQA_JOB_DONE_HOOK_KILL_TIMEOUT} // '30s';
-    $guard->abort(1);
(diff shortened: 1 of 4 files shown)


lib/OpenQA/Task/Job/FinalizeResults.pm
```diff
@@ -42,25 +42,38 @@ sub _finalize_results {
...
+
     my $hooks = $app->config->{hooks};
     my $key = "job_done_hook_$result";
     my $hook = $ENV{'OPENQA_' . uc $key} // $hooks->{lc $key};
     $hook = $hooks->{job_done_hook} if !$hook && ($trigger_hook || $hooks->{"job_done_hook_enable_$result"});
     return undef unless $hook;
+
     my $timeout = $ENV{OPENQA_JOB_DONE_HOOK_TIMEOUT} // '5m';
     my $kill_timeout = $ENV{OPENQA_JOB_DONE_HOOK_KILL_TIMEOUT} // '30s';
+    my $delay = $settings->{_TRIGGER_JOB_DONE_DELAY} // $ENV{OPENQA_JOB_DONE_HOOK_DELAY} // 60;
```
okurz, line 59:
```suggestion
    my $delay = $settings->{_TRIGGER_JOB_DONE_DELAY} // $ENV{OPENQA_JOB_DONE_HOOK_DELAY} // '1m';
```

right?

okurz (commented):
I ran `count_fail_ratio make test KEEP_DB=1 TESTS=t/10-jobs.t` and got 20/20 stable runs. There might be a slight runtime increase of t/10-jobs.t, not sure

lib/OpenQA/Task/Job/FinalizeResults.pm
kraih: No, it's seconds, that's what the lower level `delay` option uses. Are we already using a module to convert `1m` to `60` somewhere else?

kraih:
Shouldn't be much of a runtime increase, since the only delay tests that actually run multiple times use a `0` delay. The test for the `60` second delay exits after the first iteration because `perform_minion_jobs` does not wait for delayed jobs. 😉

lib/OpenQA/Task/Job/FinalizeResults.pm
okurz: right, sorry. It must be

```suggestion
    my $delay = $settings->{_TRIGGER_JOB_DONE_DELAY} // $ENV{OPENQA_JOB_DONE_HOOK_DELAY} // ONE_MINUTE;
```

using the Time::Seconds module, see https://github.com/os-autoinst/openQA/pull/4718/files#diff-10b9cc653a113467d85dbb952ef075b65f314369090682d3f1094654e2be0b2aR7

kraih: Ah, sure.

okurz (approved):

Martchus (approved):

PR #4720, opened 2022-06-24
Makefile: Ensure "test" includes also helm tests

okurz:
Running all tests as part of the "test" target is the safer approach to
ensure that no tests are overlooked. Many developers will unfortunately
not have all prerequisites to run all the tests necessary locally but
that was already the case in before. Indidvidual test parts can still be
selected explicitly as well as excluded based on variables. So this
commit adds the helm-chart test to the generic "test" target while
configuring it based on the HELM_TEST variable and same as the
CONTAINER_TEST excluding the helm-chart test when the TESTS variable is
specified to select individual test modules.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -90,6 +90,7 @@ images:
       # Exclude container specific tests needing a container runtime
       # environment, using github actions instead
       CONTAINER_TEST: 0
+      HELM_TEST: 0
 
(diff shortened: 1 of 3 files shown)


perlpunk (approved):

baierjan (approved):
Ah, nice.

PR #4723, opened 2022-06-24
Use the workers count from Minion 10.25 to make monitoring more reliable

kraih:
Iy has been suggested that our monitoring is unrelaible because adding
up number of active and inactive workers doesn't quite work right. So
i've added the number of all registered workers upstream to the Minion
stats. The feature will be available with the Minion 10.25 release.

Progress: https://progress.opensuse.org/issues/112898


kraih:
Since Minion 10.25 is not yet in Factory this PR is not ready.

Martchus (approved):

lib/OpenQA/CacheService/Controller/Influxdb.pm
```diff
@@ -12,7 +12,10 @@ sub minion ($self) {
         delayed => $stats->{delayed_jobs},
         failed => $stats->{failed_jobs},
         inactive => $stats->{inactive_jobs}};
-    my $workers = {active => $stats->{active_workers}, inactive => $stats->{inactive_workers}};
+    my $workers = {
+        registered => $stats->{workers},
```
okurz, line 16: what would happen if this happens to run with an older version of the Minion package. Thinking of users of openQA that don't use the openSUSE packages, e.g. Fedora, Debian, etc.?

kraih: 💣

kraih: That's why i bumped the dependency.

perlpunk (approved):

lib/OpenQA/CacheService/Controller/Influxdb.pm
okurz: That's why I asked because the dependency statement may not be used by maintainers.

@AdamWill @phil-hands would the Fedora+Debian packaging cope with this change?

AdamWill: I sync the Fedora package spec with the one in the repo, so we would get the dependency version bump from there.

phil-hands: Thanks for the heads-up -- I'll add a versioned depends for that.

So far I've not been adding versioned depends for every dependency, but I guess I should at least add them when the versions get bumped in your dependencies.

phil-hands: Having just checked, I already either have versioned depends where you do, or the versions being required are older than what is in the current release, so adding them would be a bit redundant.  I may add them anyway, just to be obviously in sync.

okurz: ok, so we should be good. Thank you.

okurz (approved):

okurz:
https://build.opensuse.org/package/show/devel:openQA:Leap:15.4/perl-Minion is 10.25 now so all good.

@Mergifyio rebase

cpanfile
```diff
@@ -50,7 +50,7 @@ requires 'JSON::Validator';
 requires 'JavaScript::Minifier::XS', '>= 0.11';
 requires 'LWP::Protocol::https';
 requires 'LWP::UserAgent';
-requires 'Minion', '>= 10.22';
+requires 'Minion', '>= 10.25';
 requires 'Minion::Backend::SQLite', '>= 5.0.1';
```
okurz, line 54: from `perl -d -Ilib t/25-cache-service.t` I found out what seems to be the problem that in https://github.com/Grinnz/Minion-Backend-SQLite/blob/master/lib/Minion/Backend/SQLite.pm there is only

```
  my $stats = $self->sqlite->db->query(
    q{select
      (select count(*) from minion_jobs where state = 'inactive' and (expires is null or expires > datetime('now')))
        as inactive_jobs,
      (select count(*) from minion_jobs where state = 'active') as active_jobs,
      (select count(*) from minion_jobs where state = 'failed') as failed_jobs,
      (select count(*) from minion_jobs where state = 'finished') as finished_jobs,
      (select count(*) from minion_jobs where state = 'inactive' and delayed > datetime('now')) as delayed_jobs,
      (select count(*) from minion_locks where expires > datetime('now')) as active_locks,
      (select count(distinct(worker)) from minion_jobs where state = 'active') as active_workers,
      ifnull((select seq from sqlite_sequence where name = 'minion_jobs'), 0) as enqueued_jobs,
      (select count(*) from minion_workers) as inactive_workers,
      null as uptime}
  )->hash;
```

so missing the number of "registered workers" corresponding to the Pg backend in https://github.com/mojolicious/minion/commit/c15352a78c8ee2f24dcb4ede25b1884ec83a67de#diff-500e1fc27d944109f9aea9ca0c2d977016e5e2d4691a5ef2a90b04aaf529f254R238

okurz (changes requested):

cpanfile
okurz: https://github.com/Grinnz/Minion-Backend-SQLite/pull/21

okurz:
@Mergifyio rebase

okurz:
```
[16:01:57] t/api/13-influxdb.t ................ 1/? 
#   Failed test 'exact match for content'
#   at t/api/13-influxdb.t line 48.
#          got: 'openqa_minion_jobs,url=http://example.com active=0i,delayed=0i,failed=1i,inactive=1i
# openqa_minion_workers,url=http://example.com active=0i,inactive=1i,registered=1i
# '
#     expected: 'openqa_minion_jobs,url=http://example.com active=0i,delayed=0i,failed=1i,inactive=1i
# openqa_minion_workers,url=http://example.com ,active=0i,inactive=1i,registered=1i
# '
# Looks like you failed 1 test of 17
```

okurz:
https://build.opensuse.org/package/live_build_log/devel:openQA:GitHub:os-autoinst:openQA:PR-4723/openQA/openSUSE_Leap_15.4/x86_64 reproduces the problem. I assume some local packages differ for me which actually makes it *work*

perlpunk:
I can reproduce the failure locally

perlpunk:
test passing now:
```
[09:58:54] t/api/13-influxdb.t ................ ok
```

okurz:
> test passing now:
> 
> ```
> [09:58:54] t/api/13-influxdb.t ................ ok
> ```

\o/ so maybe I fixed it by now and got confused by two months of absence in the middle or something ;)

okurz:
I see the OBS checks fail the same locally

lib/OpenQA/WebAPI/Controller/Admin/Workers.pm
```diff
@@ -46,17 +43,13 @@ sub index {
...
-    $is_admin = 1 if ($self->is_admin);
-
     $self->stash(
         workers_online => $total_online,
         total => $total,
         workers_active_free => $free_active_workers,
         workers_broken_free => $free_broken_workers,
         workers_busy => $busy_workers,
-        is_admin => $is_admin,
+        is_admin => $self->is_admin,
```
perlpunk, line 52:
```suggestion
        is_admin => $self->is_admin // 0,
```

okurz: I used `!!$self->is_admin`, thx

okurz:
IMHO https://github.com/os-autoinst/openQA/blob/master/t/25-cache-service.t#L488 should cover lib/OpenQA/CacheService/Controller/Influxdb.pm but codecov does not see any coverage recorded in https://app.codecov.io/gh/os-autoinst/openQA/blob/master/lib/OpenQA/CacheService/Controller/Influxdb.pm . Looks like the coverage isn't properly recorded. Any hints?

PR #4724, opened 2022-06-27
Add darkmode capability

pdostal:
Hello,

this pull request enables dark mode in openQA. See the screenshots:

image
image
image
image

 * Verification instance: [http://10.100.51.44/](http://10.100.51.44/) (SUSE E&I Network)

## Dark mode in Firefox
Go to `about:config` and set `layout.css.prefers-color-scheme.content-override` to: (dark:0, light:1, system-detected:2, browser-detected:3).

## Dark mode in Chrome
Go to `chrome://flags/` and search for `Dark mode` and see the dropdown menu.

## Sources:
 * [CSS - TRICKS - A Complete Guide to Dark Mode on the Web](https://css-tricks.com/a-complete-guide-to-dark-mode-on-the-web/#os-level)
 * [developer.mozilla.org/references/CSS/@media/prefers-color-scheme](https://developer.mozilla.org/en-US/docs/Web/CSS/@media/prefers-color-scheme)

### DIFF
--- a/assets/stylesheets/ansi-colors.scss
+++ b/assets/stylesheets/ansi-colors.scss
@@ -16,3 +16,11 @@
 .ansi-blue-fg    { color: $blue; }
 .ansi-magenta-fg { color: $pink; }
 .ansi-cyan-fg    { color: $cyan; }
+
+@media (prefers-color-scheme: dark) {
+    .ansi-white-fg   { color: $default-font-color-darktheme; }
+    .ansi-black-fg   { color: $default-font-color-darktheme; }
+    .ansi-red-fg     { color: $red; }
+    .ansi-yellow-fg  { color: $yellow; }
+    .ansi-green-fg   { color: $green; }
+}
--- a/assets/stylesheets/forms.scss
+++ b/assets/stylesheets/forms.scss
@@ -7,12 +7,25 @@
 .btn .fa {
     padding-right: 3pt;
 }
-.form-control {
+.form-control, .form-control:focus {
     padding: 0.4rem;
     line-height: normal;
+
+    color: $default-font-color;
+    background-color: $default-bg-color;
+    border: 1px solid rgba(0,0,0,0.1);
+    @media (prefers-color-scheme: dark) {
+        color: $default-font-color-darktheme;
+        background-color: $default-bg-color-darktheme;
+        border: 1px solid rgba(255,255,255,0.1);
+    }
 }
 .form-control:invalid {
     background: #ffaaaa80;
+
+    @media (prefers-color-scheme: dark) {
+        background: #ff666680;
+    }
 }
 .form-group .control-label {
     font-weight: bold;
@@ -65,6 +78,18 @@ select.form-control {
 @include btn(light,$btn-light-bg);
 @include btn(dark,$btn-dark-bg);
 @include btn(link,#fff);
+@media (prefers-color-scheme: dark) {
+    @include btn(default,$btn-default-bg-darktheme);
+    @include btn(primary,$btn-primary-bg-darktheme);
+    @include btn(secondary,$btn-secondary-bg-darktheme);
+    @include btn(success,$btn-success-bg-darktheme);
+    @include btn(info,$btn-info-bg-darktheme);
+    @include btn(warning,$btn-warning-bg-darktheme);
+    @include btn(danger,$btn-danger-bg-darktheme);
+    @include btn(light,$btn-light-bg-darktheme);
+    @include btn(dark,$btn-dark-bg-darktheme);
+    @include btn(link,#000);
+}
 .btn {
     border: none;
     @include box-shadow(1px 1px 4px rgba(0,0,0,.4));
@@ -103,6 +128,12 @@ select.form-control {
         }
     }
 }
+@media (prefers-color-scheme: dark) {
+    .btn.btn-light {
+        background-color: #070605;
+        border-color: #070605;
+    }
+}
 
 // wells (used for comments)
 .well-lg {
@@ -113,11 +144,16 @@ select.form-control {
     min-height: 20px;
     padding: 19px;
     margin-bottom: 20px;
-    background-color: #fdfdfd;
+    background-color: $default-bg-color;
     border: 1px solid #e3e3e3;
     border-radius: 1px;
     -webkit-box-shadow: inset 0 1px 1px rgba(0, 0, 0, 0.05);
     box-shadow: inset 0 1px 1px rgba(0, 0, 0, 0.05);
+
+    @media (prefers-color-scheme: dark) {
+        background-color: $default-bg-color-darktheme;
+        border-color: #1c1c1c;
+    }
 }
 
 // round images (used for comments) and badges inside summary
@@ -128,4 +164,11 @@ select.form-control {
 .editor-yaml-guide {
     display: none;
     text-align: left !important;
+}
+
+.custom-select {
+    @media (prefers-color-scheme: dark) {
+        color: $default-font-color-darktheme;
+        background-color: $default-bg-color-darktheme;
+    }
 }
--- a/docs/UsersGuide.asciidoc
+++ b/docs/UsersGuide.asciidoc
@@ -543,6 +543,20 @@ the top of the group overview page. However, it only works as operator
 or admin.
 
 
+=== Dark mode
+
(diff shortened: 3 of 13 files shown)


coolo:
I love it - and the implementation appears flawless to me

assets/stylesheets/ansi-colors.scss
```diff
@@ -16,3 +16,11 @@
...
 .ansi-magenta-fg { color: $pink; }
 .ansi-cyan-fg    { color: $cyan; }
+
+@media (prefers-color-scheme: dark) {
+    .ansi-white-fg   { color: $default-font-color-darktheme; }
+    .ansi-black-fg   { color: $default-font-color-darktheme; }
+    .ansi-red-fg     { color: $red; }
+    .ansi-yellow-fg  { color: $yellow; }
+    .ansi-green-fg   { color: $green; }
+}
```
b10n1k, line 26: setup the editor to not produce this "glitches" in the files please

assets/stylesheets/forms.scss
```diff
@@ -7,12 +7,23 @@
 .btn .fa {
     padding-right: 3pt;
 }
-.form-control {
+.form-control,.form-control:focus {
```
b10n1k, line 10: without comma

assets/stylesheets/ansi-colors.scss
pdostal: What do you mean by 'glitches'?

assets/stylesheets/forms.scss
pdostal: Why? It should match both `.form-control` as well as `.form-control:focus`

b10n1k (changes requested):
Very nice but i think there are components which needs some work. For instance now the *Settings* do not provide the strapping between each variable. Which i think it is useful and a good UI practice to represent the table. Another is that now you can distinguish the where the *type to search* textbox starts and where it ends. As another example buttons seems to need some other colour now. For instance if you open filters the _apply_ button could be some shade of grey so it can be visible as a button. Let me know what do you think

assets/stylesheets/ansi-colors.scss
pdostal: I added newlines at the end of every file if that helps.

assets/stylesheets/forms.scss
Martchus: It it should match both a comma is correct. I'd still prefera whitespace after the comma.

Martchus:
I've been adding the acceptance tests needed label because I want to prevent merging this automatically unless one of the team actually tested this. (Our automatic tests aren't very helpful with this.)

Just judging from the screenshots I must also say that this still need some refinement. The backgrounds are now mostly just black and miss details like alternating row backgrounds. Additionally, certain borders seem invisible and some are much more apparent (but likely shouldn't be).

pdostal:
image
image
image
Please review, and point me to other details you don't like.

Martchus:
The new screenshots look definitely better. I don't have much to criticize anymore but like to test it myself locally (going though all pages, including critical ones such as the needle editor).

> and point me to other details you don't like.

From the screenshots I only see a few detail remaining: The text input has no visible border making it basically invisible. The border of the cells of the "Test" column look a bit weird. There's also almost no visible border between the top navbar and the rest of the page (which doesn't look bad per-se but likely a slightly visible border would look better). I also hope the styling of the combo boxes is just something your local browser does (because it looks rather ugly in my opinion).

okurz (changes requested):
Very nice. I suggest you also add your instructions in the documentation or even better allow the user to save such preference although that would certainly be more work because so far we don't have a setting page per user.

pdostal:
> Very nice. I suggest you also add your instructions in the documentation or even better allow the user to save such preference although that would certainly be more work because so far we don't have a setting page per user.

Thank you Oliver, the documentation has been updated. As you can disable the dark mode in your browser I don't think that doubling that functionality in openQA makes much sense.

okurz (approved):

docs/UsersGuide.asciidoc
```diff
@@ -543,6 +543,23 @@ the top of the group overview page. However, it only works as operator
 or admin.
 
 
+=== Dark mode
+
+The dark mode is natively supported by most modern browsers and
+is automatically turned on when dark user interface of either
+the operating system or browser is detected. To disable the dark
+mode please Follow the instructions below:
```
Martchus, line 551:
```suggestion
mode please follow the instructions below:
```

docs/UsersGuide.asciidoc
```diff
@@ -543,6 +543,23 @@ the top of the group overview page. However, it only works as operator
...
+
+ * On Firefox, go to `about:config` and set
+   `layout.css.prefers-color-scheme.content-override`
+   to: (dark:0, light:1, system-detected:2, browser-detected:3).
```
Martchus, line 555: This can also be found in the normal appearance settings. Maybe we can just point users to about:preferences#general and that page is very self-descriptive (like on chrome).

mimi1vx (approved):

Martchus (changes requested):
I've tested it locally and found the following issues in normal mode:

1. Alternating row backgrounds are gone in the normal mode.
2. Background highlighting is broken of top navbar items.

And the following issues in dark mode:

1. The borders on http://localhost:9526/admin/groups look very noisy and the nested groups are still displayed brightly.
3. The background/menu of the top menus is not aligned with the border: screenshot_20220704_110158
 4. Some frames are missing and some things are still bright on the workers page, e.g.
screenshot_20220704_110423
 5. The dependency graph is still bright and tooltips are not set apart from the normal background anymore: screenshot_20220704_110903
 6. The forms are missing frames, e.g.: screenshot_20220704_111121
 7. Also the needle editor doesn't look very nice with missing frames, gray text on white and so on: 

Now I stopped, the list is already long enough. I suppose polishing this will be some work.

okurz:
@Martchus what do you see as absolutely necessary to be able to merge this?

kraih (approved):

kraih (approved):
I don't expect darkmode to be perfect from the start. Once people start to actually use it smaller flaws will get fixed quickly.

Martchus:
I'd still like to fix the problems mentioned in my last comment. That should be the most important ones.

> I don't expect darkmode to be perfect from the start.

But if one enables the dark mode globally in the web browser one is faced with a pretty flawed coloring. That is just a clear regression and not moving forward in small steps. Unless the browser provides a way to disable the dark mode for specific sites one then has to decide between living with bad coloring in openQA or living with dark mode globally disabled again.

> Once people start to actually use it smaller flaws will get fixed quickly.

kraih:
> If we want to do it this way - which would generally make sense indeed - then the dark mode should be an optional feature (and not be enabled unconditionally just due to the general browser preference) to avoid the problem mentioned in the previous paragraph.

That is true. It could work like with GitHub, where you opt-in for automatic light/dark mode.

kraih:
I'll join this effort and add a feature to switch between auto-mode, force-light-mode and force-dark-mode with a cookie to the PR.

kraih (changes requested):
Blocked because i'm adding cookie based theme settings.

kraih:
Unfortunately i don't think the current CSS in this PR is usable for a theme selector. So it will have to be reorganised a bit, possibly with CSS variables.

okurz:
I am a bit worried about "cookie based theme settings". So far we don't use cookies and I would prefer if it stays this way. Otherwise we need cookie approval dialogs which I would like to avoid. We could try to use user settings after login, same as is done by github.

kraih:
> I am a bit worried about "cookie based theme settings". So far we don't use cookies and I would prefer if it stays this way. Otherwise we need cookie approval dialogs which I would like to avoid. We could try to use user settings after login, same as is done by github.

That is how i've implemented it. There's a new "Appearance" menu item for logged in users. And that gives the option to set a theme setting (light, dark, detection) in the session cookie.

Martchus:
> So far we don't use cookies and I would prefer if it stays this way.

We're already using a cookie for the session, don't we?

kraih:
> We're already using a cookie for the session, don't we?

We do, it is only created for logged in users though.

kraih:
Made a followup PR with theme settings added. #4861

PR #4725, opened 2022-06-28
Shprdchris/change UI res

shprdchris:
Display custom screenshot resolution in webUI
  
The previous commit allowed VMs to run with a custom resolution.
However, these changes were not picked up by the webUI.
    
The changes in this commit match the screenshot resolution of a job to
the resolution of its VM.


assets/javascripts/test_result.js
```diff
@@ -96,6 +96,27 @@ function checkPreviewVisible(stepPreviewContainer, preview) {
...
+    var test_file =  "/api/v1/jobs/"+testStatus.jobid;
+    xhr.open("GET", test_file, false);
+    xhr.send(null);
+    let json = JSON.parse(xhr.responseText);
+    x_res = json.job.settings.XRES;
+    y_res = json.job.settings.YRES;
+  }
+  catch(error) {
+    // if {X,Y}RES not defined, will be using the defaults
+    x_res = 1024;
```
Martchus, line 114: Ignoring errors silently is not very nice.

Martchus (changes requested):
Wouldn't it make way more sense to read the resolution from the image instead of doing an additional API request just to query the job settings? Even if it cannot be read from the image, the additional API request could likely be avoided by including the information somewhere in existing requests (e.g. the initial HTML or some other AJAX request that is already done).

assets/javascripts/test_result.js
kraih: Seconded. Only ignore very specific errors.

shprdchris: Hi @Martchus, thanks a lot for all this feedback. This is my first PR on openQA I wouldn't be surprised if my solution isn't an optimal one. RE the original API requests: could you point me to a concrete example?

I also agree that this error handling isn't ideal. Maybe I could catch specific "{X,Y}_RES doesn't exist" errors with this arm, and report other errors as "real" errors?

assets/javascripts/test_result.js
```diff
@@ -96,6 +96,27 @@ function checkPreviewVisible(stepPreviewContainer, preview) {
...
 }
 
+function XYRes() {
+  // the test VM's resolution
+  let x_res;
+  let y_res;
+  try {
+    var xhr = new XMLHttpRequest();
+    var test_file =  "/api/v1/jobs/"+testStatus.jobid;
+    xhr.open("GET", test_file, false);
```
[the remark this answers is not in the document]
shprdchris: The issue I had with the calls need a non-empty return, and I don't know how to do that via the async calls. Is the latter a thing?

assets/javascripts/test_result.js
kraih: It's customary to handle the errors you're interested in and just rethrow what remains.

assets/javascripts/test_result.js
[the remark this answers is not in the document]
Martchus: Let's talk about improving the AJAX code once we have established that there's no way around it. I still think it can be avoided completely, see my other comment about it.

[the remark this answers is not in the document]
shprdchris: I think I've managed to implement your original suggestion. Please let me know what you think :)

Martchus (approved):
I've tested it locally and it also works when needles/screenshot have different resolutions.

okurz:
@Mergifyio rebase

okurz (approved):

Martchus:
Was this really force-pushed by @kraih (and not by @mergify-io)?

Martchus (changes requested):
* The liveview works.
* The needle diff works as mentioned before. However, I have to take back that it works nicely when the screenshot and needles have different resolutions. However, that's likely something we can live with.
* I'm getting `Uncaught TypeError: this.canvasDiv is null` in the needle editor.

Martchus (changes requested):

kraih:
I'm on vacation, that looks like someone is impersonating me. 🤨

assets/javascripts/needleeditor.js
```diff
@@ -2,6 +2,7 @@ function NeedleEditor(needle) {
   this.textarea = document.getElementById('needleeditor_textarea');
   this.tags = document.getElementById('needleeditor_tags');
   this.canvas = document.getElementById('needleeditor_canvas');
+  this.canvasDiv = document.getElementById('needleeditor_div');
```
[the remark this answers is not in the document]
shprdchris: Hi Martchus, thanks for taking another look at this. This line is actually an artefact of a line of reasoning I didn't end up using. I've removed this redundant ID and hopefully this should now run OK without the error message :)

assets/javascripts/needleeditor.js
```diff
@@ -19,6 +19,8 @@ NeedleEditor.prototype.init = function () {
     return;
```
Martchus, line 19: It would be great if the canvas width/height would adjust itself when switching the screenshot via the "Take image from:" combo box. I suppose all you need to do for that is moving your new lines before that if block so it is also executed in case of the early return.

Martchus (changes requested):

assets/javascripts/needleeditor.js
shprdchris: OK, that should be done.

Martchus (approved):

okurz:
Interesting thing in the OBS checks:

```
[   39s] Compressed openQA-change-resolution-4.6.1657095704.6145d8b.tar to openQA-change-resolution-4.6.1657095704.6145d8b.tar.xz
[   39s] -----------------------------------------------------------------
[   39s] I have the following modifications for openQA.spec:
[   39s] 80c80
[   39s] < Release:        0
[   39s] ---
[   39s] > Release:        5174.1
[   39s] -----------------------------------------------------------------
[   39s] ----- building openQA.spec (user abuild)
[   39s] -----------------------------------------------------------------
[   39s] -----------------------------------------------------------------
[   40s] + exec rpmbuild -ba --define '_srcdefattr (-,root,root)' --nosignature --undefine _enable_debug_packages --define 'disturl obs://build.opensuse.org/devel:openQA:GitHub:os-autoinst:openQA:PR-4725/openSUSE_Tumbleweed/7cdfdc2d1bd41af30b1999ae756dfd17-openQA' /home/abuild/rpmbuild/SOURCES/openQA.spec
[   40s] error: Bad source: /home/abuild/rpmbuild/SOURCES/openQA-4.6.1657095704.6145d8b.tar.xz: No such file or directory
```

so apparently a tarball "openQA-change-resolution-4.6.1657095704.6145d8b.tar.xz" is generated with the name part "change-resolution" instead of just "openQA". Why is that?

mimi1vx:
> so apparently a tarball "openQA-change-resolution-4.6.1657095704.6145d8b.tar.xz" is generated with the name part "change-resolution" instead of just "openQA". Why is that?

because fork (source repo ) is named "openQA-change-resolution" ..

PR #4729, opened 2022-07-01
apparmor: Allow grep needed by git

okurz:
See https://github.com/os-autoinst/os-autoinst/pull/2104


baierjan (approved):

mimi1vx (approved):

okurz:
Those changes would be needed for https://github.com/os-autoinst/os-autoinst/pull/2104/ but https://github.com/os-autoinst/os-autoinst/pull/2104/ also needs write permissions for the user `_openqa-worker` to a home directory which currently is /var/lib/empty which does not allow to write. Not sure how to continue.

okurz:
This should not be necessary anymore.

PR #4731, opened 2022-07-04
Allow configuring results to collapse as parallel children on overview

Martchus:
See https://progress.opensuse.org/issues/111470#note-5

@AdamWill

### DIFF
--- a/templates/webapi/test/overview.html.ep
+++ b/templates/webapi/test/overview.html.ep
@@ -5,6 +5,7 @@
 % use List::Util qw(sum);
 
 % content_for 'ready_function' => begin
+  window.overviewParallelChildrenCollapsableResultsSel = '<%= $parallel_children_collapsable_results_sel %>';
   setupOverview();
 % end
 
--- a/assets/javascripts/overview.js
+++ b/assets/javascripts/overview.js
@@ -26,7 +26,7 @@ function toggleAllParallelChildren(expand, table) {
 }
 
 function hasOnlyOkChildren(table, parentID) {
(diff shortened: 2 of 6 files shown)


templates/webapi/test/overview.html.ep
```diff
@@ -5,6 +5,7 @@
 % use List::Util qw(sum);
 
 % content_for 'ready_function' => begin
+  window.overviewOkResultSelector = '<%= $ok_result_sel %>';
```
okurz, line 8: this looks to be inconsistent. In other places you say "…collapsable_results_sel" but here "ok_result_sel" even though it's configurable so not necessarily the same as the real definition of "ok" results that we use elsewhere

okurz (changes requested):

templates/webapi/test/overview.html.ep
Martchus: My thinking was the selector defines what "ok" results are in that particular occasion. However, it is likely better to stick with "collapsable_results_sel".

AdamWill:
Well, it's definitely nice for me to have an option to go back to the old way. The concern is always that adding configurable features adds complexity and can cause less-used paths to break, but in this case it doesn't look too bad.

Martchus:
It is actually not a separate code path. All new lines of code are always executed, just with a different number of array elements.

AdamWill:
Yup, that's why I said in this case it looks good :) I meant that was always the *general* concern when adding configurable options.

okurz (approved):

PR #4735, opened 2022-07-06
Clarify client documentation about --accept option

perlpunk:


Martchus (approved):

kalikiana (approved):

script/client
```diff
@@ -49,8 +49,8 @@ Output YAML instead of Perl structures.
 
 Send Accept header to explicitly tell the API what format is accepted
 
-Note that currently only one endpoint (job_templates_scheduling) is able
-to return YAML, and it's deprecated.
+Returning YAML is deprecated, and currently only one endpoint
```
mimi1vx, line 52: is really deprecated ? Only one endpoint works with YAML but I don't think we going this to change into JSON

okurz: I also still don't really understand the message. Put yourself into the shoes of a newbie wanting to use this. How should they read it?

okurz:
> […]
> But an API response should return the same data, regardless of the content type. That's an opinion, of course.

I agree with that

> […]
> That's why I suggested we mark getting YAML content type as deprecated, and at a later point we can support YAML as a general supported reponse content-type, returning the same data as for JSON.

Understood. How about explaining that the endpoint returning YAML is not returning data consistent with the expected JSON output hence should be treated with care and at best avoided.

> In the new client this is not documented anymore at all.

Honestly I overlooked that part. So it's even "double-deprecated" :D

mimi1vx:
> > […]
> > But an API response should return the same data, regardless of the content type. That's an opinion, of course.
> 
> I agree with that
+1 . but there is problem ... YAML is superset of JSON so data in YAML aren't equivalent to JSON , so here to get full data is used json string instead of proper almost same structure

PR #4737, opened 2022-07-08
Fix unstable `t/ui/23-audit-log.t` by implementing wait condition

Martchus:
* Wait until the expected number of rows show up in the DataTable (with a
  timeout) instead of using `sleep 1` to prevent failures when the one
  second is not enough
* See https://progress.opensuse.org/issues/113138

### DIFF
--- a/t/ui/23-audit-log.t
+++ b/t/ui/23-audit-log.t
@@ -3,6 +3,7 @@
 # SPDX-License-Identifier: GPL-2.0-or-later
 
 use Test::Most;
+use Mojo::Base -base, -signatures;
 
 use FindBin;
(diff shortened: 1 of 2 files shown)


baierjan (approved):

t/ui/23-audit-log.t
```diff
@@ -11,17 +12,29 @@ use Test::Warnings ':report_warnings';
...
 
 use OpenQA::SeleniumTest;
 
 my $schema = OpenQA::Test::Case->new->init_data;
 driver_missing unless my $driver = call_driver;
 
-sub wait_for_data_table {
-    wait_for_ajax;
-    # TODO: add some wait condition for rendering here
-    sleep 1;
```
okurz, line 24: finally :D

PR #4739, opened 2022-07-12
Fix sorting of jobs that have never run on 'All tests' page

Martchus:
If jobs have no `t_finished` timestamp (as they've never run) they are
currently shown before any other jobs. In production that's not a problem
because we hardly cancel jobs before they had a chance to run. However, on
a local system where one experiments with openqa-clone-job that can easily
be the case. This change simply considers `t_updated` if `t_finished` is
not set. That should be good enough because `t_updated` is normally last
set when the job is canceled.

---

Note that this is just something I personally find annoying. I admit that with this
change it is a bit weird that the time of cancellation is shown within the "Finished"
column but I find it better than having the sorting out of place.


okurz (approved):

kalikiana (approved):
Works for me.

PR #4745, opened 2022-07-14
Switch development base to current openSUSE Leap 15.4

okurz:
Related progress issue: https://progress.opensuse.org/issues/111881


Martchus (approved):
Exchanging the 3 with a 4 should be easy enough.

okurz:
Trying first with the CI base container #4747

baierjan:
> "check-helm-chart" says...

Yeah, it is using the containers you are trying to create in this PR so the clean process would be to split that into yet another PR. But as you said, this PR should fix the issues in this PR, so feel free to merge anyway.

okurz:
moved helm chart changes to
* #4748

baierjan (approved):

PR #3112, opened 2020-05-22
Enable NOVIDEO for jobs with increased MAX_JOB_TIME by default

Martchus:
The video contributes significantly to the log size. This is especially the
case for long-running results so it makes sense to disable the video there
by default.

See
* https://progress.opensuse.org/issues/66922
* https://progress.opensuse.org/issues/67087#note-3

PR #4746, opened 2022-07-14
Stop setting NOVIDEO=1 for longer jobs

Vogtinator:
Videos are not that big even if the job is running longer than whatever the
"default" is resp. the maximum time is. They are very valuable for debugging
failed tests and even more useful in the case of those longer running tests,
like upgrade scenarios.

### DIFF
--- a/lib/OpenQA/Worker/Job.pm
+++ b/lib/OpenQA/Worker/Job.pm
@@ -196,8 +196,6 @@ sub _compute_timeouts ($job_settings) {
     my $timeout_scale = $job_settings->{TIMEOUT_SCALE};
     $max_job_time = DEFAULT_MAX_JOB_TIME unless looks_like_number $max_job_time;
     $max_setup_time = DEFAULT_MAX_SETUP_TIME unless looks_like_number $max_setup_time;
-    # disable video for long-running scenarios by default
-    $job_settings->{NOVIDEO} = 1 if !exists $job_settings->{NOVIDEO} && $max_job_time > DEFAULT_MAX_JOB_TIME;
     $max_job_time *= $timeout_scale if looks_like_number $timeout_scale;
     return ($max_job_time, $max_setup_time);
 }
--- a/docs/WritingTests.asciidoc
+++ b/docs/WritingTests.asciidoc
@@ -210,10 +210,6 @@ The download of assets, synchronization of tests and other setup tasks do *not*
 count into `MAX_JOB_TIME`. However, the setup time is limited by default to one
 hour. This can be changed by setting `MAX_SETUP_TIME`.
 
-To save disk space, increasing `MAX_JOB_TIME` beyond the default will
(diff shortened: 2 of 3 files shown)


Martchus (changes requested):
I think we had a good reason for doing this, although it was mainly about OSD. Maybe this should be configurable per instance. Or just add `NOVIDEO=0` on relevant o3 workers.

Vogtinator:
> I think we had a good reason for doing this, although it was mainly about OSD.

Can you elaborate on this?

Martchus:
We were running out of disk space on OSD, see tickets referenced on the PR that introduced the feature (https://github.com/os-autoinst/openQA/pull/3112). That wasn't the first alert of that kind. Before, we frequently deleted old video manually. The videos unfortunately did make a significant different. (On OSD there are unfortunately more long-running tests and some are really running extensively run.)

Vogtinator:
Maybe just manually set `NOVIDEO=0` in those cases if it's specific to certain tests on that instance?

Martchus:
You don't actually think that test maintainers would be disciplined enough to put `NOVIDEO=0` in certain test scenarios appropriately. Having to enable the video manually might be tedious but it forced people to think at least for one second whether they actually need the video (or the long test runtime).

okurz:
I agree that videos are important. But simply reverting doesn't cut it. If you want to help I suggest to do https://progress.opensuse.org/issues/77842 "Use external videoencoder in production on all o3 machines" first to actually have smaller videos, then we can enable videos again.

I also think having this configurable per instance is a good idea. I would keep videos enabled on o3 and disable on osd if above MAX_JOB_TIME.

lib/OpenQA/Worker/Job.pm
```diff
@@ -196,8 +196,6 @@ sub _compute_timeouts ($job_settings) {
     my $timeout_scale = $job_settings->{TIMEOUT_SCALE};
     $max_job_time = DEFAULT_MAX_JOB_TIME unless looks_like_number $max_job_time;
     $max_setup_time = DEFAULT_MAX_SETUP_TIME unless looks_like_number $max_setup_time;
-    # disable video for long-running scenarios by default
-    $job_settings->{NOVIDEO} = 1 if !exists $job_settings->{NOVIDEO} && $max_job_time > DEFAULT_MAX_JOB_TIME;
```
okurz, line 200: I think something like this logic is still beneficial. How about

```suggestion
    # disable video for long-running scenarios by default
    $job_settings->{NOVIDEO} = 1 if !exists $job_settings->{NOVIDEO} && $max_job_time > DEFAULT_MAX_VIDEO_JOB_TIME;
```

with DEFAULT_MAX_VIDEO_JOB_TIME being something like 4*DEFAULT_MAX_JOB_TIME so that there is still some reasonable limit for overly excessive values of MAX_JOB_TIME?

okurz:
included in https://github.com/os-autoinst/openQA/pull/7264

PR #4751, opened 2022-07-15
worker: Retry registration on timestamp mismatch

kalikiana:
Also respect OPENQA_WORKER_CONNECT_INTERVAL like in existing cases.

See: https://progress.opensuse.org/issues/105855

### DIFF
--- a/t/24-worker-overall.t
+++ b/t/24-worker-overall.t
@@ -538,6 +538,14 @@ subtest 'handle client status changes' => sub {
     like($output, qr/Test disabling - ignoring server/s, 'client disabled');
     unlike($output, qr/Stopping.*because registration/s, 'worker not stopped; there are still other clients');
 
+    # see if the expected number of retries is performed
+    $ENV{OPENQA_WORKER_CONNECT_INTERVAL} = 7;
+    combined_like {
+        $worker->_handle_client_status_changed($fake_client,
+            {status => 'failed', reason => 'test', error_message => '500 response: unavailable'})
+    }
+    qr/trying again in 7 seconds/s, 'worker will attempt to retry later';
+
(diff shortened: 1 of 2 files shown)


asdil12 (approved):
LGTM

Martchus (changes requested):
In my opinion it would make much more sense to let the web UI connection go into the "failed" state (instead of the "disabled" state and add additional handling for that state). That would mean a change in `WebUIConnection.pm` where all distinctions according to the HTTP response are already made. (The code in `Worker.pm` isn't supposed to make the distinction anymore.)

t/24-worker-overall.t
```diff
@@ -538,6 +538,15 @@ subtest 'handle client status changes' => sub {
...
     unlike($output, qr/Stopping.*because registration/s, 'worker not stopped; there are still other clients');
 
+    # assume one of the clients is suffering from clock skew
+    $ENV{OPENQA_WORKER_CONNECT_INTERVAL} = 7;
+    $fake_client_2->status('failed');
+    combined_like {
+        $worker->_handle_client_status_changed($fake_client_2,
+            {status => 'failed', reason => 'test', error_message => '403 response: timestamp mismatch'})
+    }
+    qr/trying again in 7 seconds/s, 'worker will attempt to retry later';
```
Martchus, line 548: Looks like we don't test this case so far, at least not with an explicit check for the log message. So I guess it is good to have.

Having that said, this test is *not* covering the new/additional branch. Not sure whether that's necessary, though.

Martchus (approved):

t/24-worker-overall.t
kalikiana: You're right, I updated the test phrasing. This code is not exercising the registration case...

PR #4753, opened 2022-07-16
Dependency cron 2022-07-16

openqabot:


okurz (changes requested):
Looks like os-autoinst deps are removed, tis ain't right

PR #4755, opened 2022-07-18
Dependency cron 2022-07-18

openqabot:


kalikiana (approved):

okurz:
@kalikiana
1. Dependency cron updates do not need to be approved
2. Tests fail and os-autoinst dependencies are removed so this shouldn't be accepted anyway

PR #4768, opened 2022-07-29
Avoid server error if vars.json is missing

perlpunk:
Issue: https://progress.opensuse.org/issues/114829

This can happen if the test failed early, e,g, when downloading an image
failed.

### DIFF
--- a/lib/OpenQA/Schema/Result/Jobs.pm
+++ b/lib/OpenQA/Schema/Result/Jobs.pm
@@ -1884,7 +1884,7 @@ sub investigate ($self, %args) {
               // undef
         } ($prev, $self);
         $inv{diff_packages_to_last_good} = $self->packages_diff($prev, $ignore) // 'Diff of packages not available';
-        last unless $prev_file;
+        last unless $self_file && $prev_file;
         # just ignore any problems on generating the diff with eval, e.g.
         # files missing. This is a best-effort approach.
         my $diff = eval { diff(\$prev_file, \$self_file, {CONTEXT => 0}) };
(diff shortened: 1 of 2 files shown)


kalikiana (approved):

lib/OpenQA/Schema/Result/Jobs.pm
```diff
@@ -1884,7 +1884,7 @@ sub investigate ($self, %args) {
               // undef
         } ($prev, $self);
         $inv{diff_packages_to_last_good} = $self->packages_diff($prev, $ignore) // 'Diff of packages not available';
-        last unless $prev_file;
+        last unless ($self_file and $prev_file);
```
Martchus, line 1887: Why not just
```suggestion
        last unless $self_file && $prev_file;
```

t/10-jobs.t
```diff
@@ -484,6 +484,15 @@ subtest 'carry over, including soft-fails' => sub {
         like $inv->{test_diff_stat}, qr/Invalid range UNKNOWN..c65/, 'test_diff_stat has message about invalid range';
     };
 
+    subtest 'No vars.json' => sub {
+        unlink path($job->result_dir(), 'vars.json');
+        my $last_good_path = path(($job->_previous_scenario_jobs)[1]->result_dir(), 'vars.json');
+        path('t/data/last_good.json')->copy_to(path($last_good_path));
```
Martchus, line 490:
```suggestion
        path('t/data/last_good.json')->copy_to($last_good_path);
```

lib/OpenQA/Schema/Result/Jobs.pm
kalikiana: You say "just". Is there a difference? Maybe this is something tidy should check

Martchus: The operators can make a difference but here we don't so I suggest to stick to our default of using `%%` unless we're chaining commands and avoid braces in simple cases. Maybe tidy can help us with the braces but likely not with the operator.

perlpunk: Changed it. I'm used to using `and` and `or` because I find it more readable.

t/10-jobs.t
perlpunk: done

okurz (approved):

Martchus (approved):

PR #4769, opened 2022-07-29
Post error message similarity

Akemiiii:
This script scans all the autoinst-log.txt in the testresults directory. Then it finds the error messages in the autoinst-log.txt. To avoid double computing, it saves some results in a .pkl file. The file can be converted into a JSON or others if .pkl is not suitable. Next, it uses Levenshtein to calculate the text distance. Finally, it posts the results as comments by OpenQA_Client.
I wrote it in Python because of openqa_client module. I am not sure where it should be, so I put it in the tools/. Thank you.

### DIFF
--- a/tools/post_similarity.py
+++ b/tools/post_similarity.py
@@ -1,0 +1,94 @@
+import os
+import Levenshtein
+import pickle
+from http import client
+from openqa_client.client import OpenQA_Client
+import time
+
+"""
+This script scans all the autoinst-log.txt in the testresults directory.
+Then find the error messages in the autoinst-log.txt.
+Use Levenshtein to calculate the text distance.
+Finally, post the results as comments by OpenQA_Client.
+"""
+
+# gather all the autoinst-log.txt
+logs = []
+for root, dirs, files in os.walk("testresults"):
+    for name in files:
+        if name == "autoinst-log.txt":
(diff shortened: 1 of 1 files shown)


okurz (changes requested):
I think this is a great start. I suggest you just submit it to github.com/os-autoinst/scripts/. There we can more easily accept these scripts which are not directly necessary on the openQA web server. Also it would allow us to use this script in "hook scripts" which we can call on openQA jobs automatically on the server, see http://open.qa/docs/#_enable_custom_hook_scripts_on_job_done_based_on_result . Please also see https://github.com/os-autoinst/scripts/#auto-review---automatically-detect-known-issues-in-openqa-jobs-label-openqa-jobs-with-ticket-references-and-optionally-retrigger . I think "post_similarity" would be a great addition to "auto-review" and "openqa-investigate" for example by posting on any new failures if there are already similar failures elsewhere. Walking over *all* testresults might be a bit too heavy but we can try to tweak that later.

Akemiiii:
Thank you. I will create a pull request on github.com/os-autoinst/scripts/. I have converted the .pkl file into a JSON file for readability. I also changed some codes to minimize the use of walk.

okurz:
The addition in the scripts repo is enough to replace this

PR #4771, opened 2022-08-09
worker: Enable color output explicitly

kalikiana:
This will be necessary with the default becoming `auto`.

See: https://progress.opensuse.org/issues/114878

### DIFF
--- a/lib/OpenQA/Worker/Engines/isotovideo.pm
+++ b/lib/OpenQA/Worker/Engines/isotovideo.pm
@@ -36,6 +36,7 @@ use constant CACHE_SERVICE_TEST_SYNC_ATTEMPTS => $ENV{OPENQA_CACHE_SERVICE_TEST_
 
 my $isotovideo = '/usr/bin/isotovideo';
 my $workerpid;
+my $isotovideo_interface_version = 0;
 
 sub set_engine_exec ($path) {
     if ($path) {
@@ -44,9 +45,9 @@ sub set_engine_exec ($path) {
         $isotovideo = abs_path($path);
     }
     if (-f $isotovideo && qx(perl $isotovideo --version) =~ /interface v(\d+)/) {
-        return $1;
+        $isotovideo_interface_version = $1;
     }
-    return 0;
+    return $isotovideo_interface_version;
 }
 
 sub _save_vars ($pooldir, $vars) {
@@ -466,7 +467,10 @@ sub _engine_workit_step_2 ($job, $job_settings, $vars, $shared_cache, $callback)
             local $ENV{PERL5OPT} = '';
             # Allow to override isotovideo executable with an arbitrary
             # command line based on a config option
-            exec $job_settings->{ISOTOVIDEO} ? $job_settings->{ISOTOVIDEO} : ('perl', $isotovideo, '-d');
+            my $args = ('-d');
+            $isotovideo_interface_version = 30; # XXX Trying to confirm if any test covers this
+            push @{$args}, '--color=yes' if $isotovideo_interface_version > 26;
+            exec $job_settings->{ISOTOVIDEO} ? $job_settings->{ISOTOVIDEO} : ('perl', $isotovideo, $args);
             die "exec failed: $!\n";
         });
     $child->on(


kalikiana:
This is related to os-autoinst/os-autoinst/pull/2143 and **may** help with [no cached packages on CircleCI](https://app.circleci.com/pipelines/github/os-autoinst/openQA/10058/workflows/8547ff2a-ca9e-4af4-9afe-8ffda00e164f/jobs/94681).

lib/OpenQA/Worker/Engines/isotovideo.pm
```diff
@@ -36,6 +36,7 @@ use constant CACHE_SERVICE_TEST_SYNC_ATTEMPTS => $ENV{OPENQA_CACHE_SERVICE_TEST_
 
 my $isotovideo = '/usr/bin/isotovideo';
 my $workerpid;
+my $isotovideo_interface_version = 0;
```
Martchus, line 39: It would be great to avoid introducing further global variables. This version number is already accessible via an attribute of the worker object (which is even already present in a local variable in the function where you need it).

lib/OpenQA/Worker/Engines/isotovideo.pm
```diff
@@ -466,7 +467,9 @@ sub _engine_workit_step_2 ($job, $job_settings, $vars, $shared_cache, $callback)
             local $ENV{PERL5OPT} = '';
             # Allow to override isotovideo executable with an arbitrary
             # command line based on a config option
-            exec $job_settings->{ISOTOVIDEO} ? $job_settings->{ISOTOVIDEO} : ('perl', $isotovideo, '-d');
+            my $args = ('-d');
+            push @{$args}, '--color=yes' if $isotovideo_interface_version > 26;
+            exec $job_settings->{ISOTOVIDEO} ? $job_settings->{ISOTOVIDEO} : ('perl', $isotovideo, $args);
```
perlpunk, line 472:
```suggestion
            exec $job_settings->{ISOTOVIDEO} ? $job_settings->{ISOTOVIDEO} : ('perl', $isotovideo, @$args);
```

kalikiana: I guess I can just declare it as `@args` 🤔 Interestingly the tests are fine as-is

lib/OpenQA/Worker/Engines/isotovideo.pm
```diff
@@ -466,7 +467,9 @@ sub _engine_workit_step_2 ($job, $job_settings, $vars, $shared_cache, $callback)
             local $ENV{PERL5OPT} = '';
             # Allow to override isotovideo executable with an arbitrary
             # command line based on a config option
-            exec $job_settings->{ISOTOVIDEO} ? $job_settings->{ISOTOVIDEO} : ('perl', $isotovideo, '-d');
+            my $args = ('-d');
```
Martchus, line 470: This should've been an array, right?
```suggestion
            my @args = ('-d');
```

lib/OpenQA/Worker/Engines/isotovideo.pm
Martchus, line 472: I'm wondering whether this will work but I suppose if `$args` was actually an array (`@args`) is would.

Martchus (changes requested):

lib/OpenQA/Worker/Engines/isotovideo.pm
kalikiana: See above. I should apparently not trust our unit tests too much...

perlpunk: I guess the test passes because the `push @{$args}` is never executed, so basically it is a string `$args = '-d'` and this is expected to work.
`push @$args` on a string would die under `strict`.

kalikiana: All worker* tests pass for me locally even with an unconditional push

Martchus: I suppose tests should go into `t/24-worker-engine.t` (even though it doesn't cover this particular code yet).

PR #4778, opened 2022-08-15
Improve AutoYaST profile

Martchus:
See particular commit messages and https://progress.opensuse.org/issues/111473


okurz:
The first commit I have in https://github.com/os-autoinst/openQA/pull/4777 as well. About the other two changes I am not so sure. Requiring at least 100GB for a root partition is maybe not something that can be commonly ensured. About the zram change I think it's a good idea but did you see if it already works or do we still need to enable a service or something?

Martchus:
Let's consider this a draft for now.

PR #4780, opened 2022-08-16
Document configuring ports for developer mode with firewalld

Martchus:


mimi1vx (approved):

docs/Pitfalls.asciidoc
```diff
@@ -117,3 +117,18 @@ This is basically a checklist to go through in case the developer mode is broken
   which component needs access to which other components and the ports used by default.
 
 //-
+
+=== Configure firewalld
+As mentioned, certain ports are required to be open on the worker host. If you are
```
okurz, line 122: Better to qualify that this is necessary in case when not using a public zone already which we do in our production cases as the "security measures" happen by segregating the complete networks

baierjan (approved):

PR #4785, opened 2022-08-21
Add support for pagure.io and gitlab.gnome.org trackers

AdamWill:
These have some little quirks: the weird useless /-/ that gitlab
puts in issue URLs, and how pagure uses /issue/ not /issues/ .

### DIFF
--- a/lib/OpenQA/Utils.pm
+++ b/lib/OpenQA/Utils.pm
@@ -49,6 +49,8 @@ BEGIN {
         kde => 'https://bugs.kde.org/show_bug.cgi?id=',
         fdo => 'https://bugs.freedesktop.org/show_bug.cgi?id=',
         jsc => 'https://jira.suse.de/browse/',
+        pio => 'https://pagure.io/',
+        ggo => 'https://gitlab.gnome.org/',
     );
     %BUGURLS = (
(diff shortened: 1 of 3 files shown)


AdamWill:
It's always a little nervous touching someone else's complicated string-handling code, but I *think* these are right, and they pass tests. Note, for gitlab.gnome.org , you can leave the `/-` out of the URL and it'll handle it OK - so we don't need to generate that bit into the URLs we produce - but it will rewrite it to put the `/-` in, and any internal links and so on have the `/-` in it. So URLs people copy-and-paste into openQA are gonna have that bit in them and we need to handle it.

AdamWill:
oh, and pagure.io allows repositories which aren't under a group, e.g. https://pagure.io/releng/issue/10757 , and ones that are, e.g. https://pagure.io/fedora-qa/fedora_openqa/issue/90 . So we need to handle both of those; fortunately it seems to work OK with no changes needed, I included tests to make sure.

For pagure, you can't treat issues and pull requests interchangeably like you can for github, but I don't really need to link to pull requests so I didn't bother handling them specially. For the record, the URL format for pull requests is like this: https://pagure.io/fedora-comps/pull-request/754

mimi1vx (approved):

kalikiana (approved):
Looks sensible.

PR #4791, opened 2022-09-01
Avoid race condition when creating result directories

Martchus:
The `create_result_dir` function might be called concurrently (nobody
prevents API users from doing that) and then we might end up with
`400 response: can't mkdir: … File exists …`. This changes uses
`Mojo::File` which simply doesn't care whether the directories are already
there instead of checking `!-d $dir` before. It also simplifies the code.

I have also encountered a real incomplete on o3 that ran into the issue
(which was actually why I looked into the code):
https://openqa.opensuse.org/tests/2598364

PR #4792, opened 2022-09-01
Fix AJAX/parsing error in needle tables

Martchus:
The date/time picker of Firefox and Chromium passes a timestamp without
seconds when a timestamp on a whole minute is selected. So far this is not
accepted by the controller code leading to an alert box complaining about
an AJAX error. This change fixes the issue by allowing the format on the
controller side. Note that minutes and hours always seem to be present.

This issue has been found due to a test failure when the date/time picker
was randomly on a whole minute (see
https://github.com/os-autoinst/openQA/pull/4791#issuecomment-1234391043 and
further comments).

### DIFF
--- a/lib/OpenQA/WebAPI/Controller/Admin/Needle.pm
+++ b/lib/OpenQA/WebAPI/Controller/Admin/Needle.pm
@@ -20,8 +20,7 @@ sub index {
 sub _translate_days ($days) { time2str('%Y-%m-%d %H:%M:%S', time - $days * ONE_DAY, 'UTC') }
 
 sub _translate_date_format ($datetime) {
-    my $datetime_obj = DateTime::Format::Pg->parse_datetime($datetime);
-    return DateTime::Format::Pg->format_datetime($datetime_obj);
+    DateTime::Format::Pg->format_datetime(DateTime::Format::Pg->parse_datetime($datetime));
 }
 
 sub _translate_cond ($cond) {
--- a/lib/OpenQA/WebAPI/Controller/Admin/Needle.pm
+++ b/lib/OpenQA/WebAPI/Controller/Admin/Needle.pm
@@ -24,19 +24,15 @@ sub _translate_date_format ($datetime) {
 }
 
 sub _translate_cond ($cond) {
-    if ($cond =~ m/^min(\d+)$/) {
-        return {'>=' => _translate_days($1)};
+    my ($operator, @additional_conds) = rindex($cond, 'min', 0) == 0 ? ('>=') : ('<', {'=' => undef});
+    my $translated;
+    if ($cond =~ m/^(min|max)(\d+)$/) {
+        $translated = _translate_days($2);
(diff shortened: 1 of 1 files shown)


lib/OpenQA/WebAPI/Controller/Admin/Needle.pm
```diff
@@ -17,33 +17,22 @@ sub index {
...
-sub _translate_cond($) {
-    my ($cond) = @_;
-
-    if ($cond =~ m/^min(\d+)$/) {
-        return {'>=' => _translate_days($1)};
-    }
-    elsif ($cond =~ m/^max(\d+)$/) {
-        return [{'<' => _translate_days($1)}, {'=' => undef}];
+sub _translate_cond ($cond) {
+    my ($operator, @additional_conds) = rindex($cond, 'min', 0) == 0 ? ('>=') : ('<', {'=' => undef});
```
perlpunk, line 27: Why are you using `rindex` here instead of `index`?

lib/OpenQA/WebAPI/Controller/Admin/Needle.pm
```diff
@@ -86,13 +75,16 @@ sub ajax {
...
+    eval {
+        if ($seen_query && $seen_query ne 'none') {
+            push(@filter_conds, {last_seen_time => _translate_cond($seen_query)});
+        }
+        my $match_query = $self->param('last_match');
+        if ($match_query && $match_query ne 'none') {
+            push(@filter_conds, {last_matched_time => _translate_cond($match_query)});
+        }
+    };
+    return $self->render(json => {error => $@}, status => 400) if $@;
```
perlpunk, line 87: I think we usually remove internal information from error messages that go to the client, so maybe:
```suggestion
    return $self->render(json => {error => $@ =~ s/ at .*//sr}, status => 400) if $@;
```

lib/OpenQA/WebAPI/Controller/Admin/Needle.pm
Martchus: I actually just want a `starts_with`. I suppose neither `index` nor `rindex` are providing that directly but `rindex` would maybe waste less time going though the whole string if `min` in not in the string.

perlpunk: Anchored perl regexes are really fast, plus this code isn't really performance critical.
So I would suggest `$cond =~ m/^min/`

Martchus: I have changed it to use a regex. Maybe that is better readable for more people then. (However, I don't think it is generally more readable. It really depends on whether one knows regex syntax or the rindex function better. And of course what we'd really wanted would be an intuitively named function like `starts_with`.)

perlpunk (approved):

kalikiana (approved):

Issue #4795, opened 2022-09-07
QEMUVGA=virtio does not work in latest contains


ssssam:
The `openqa_worker` container no longer supports QEMU `-vga virtio` option. So tests on openqa.gnome.org have been failing with `qemu-system-x86_64: Virtio VGA not available`.

The GNOME OS images do not work with the standard QEMU vga backend. (That may be fixable, but i don't know how much effort).

It seems like virtio support wasn't deliberately removed, rather that the following snippet in [`scripts/openqa_bootstrap`](https://github.com/os-autoinst/openQA/blob/master/script/openqa-bootstrap#L36) stopped working:

```
# this was split into a separate package on newer dist versions - so install it if available
if zypper -n search -x qemu-hw-display-virtio-gpu-pci ; then
    zypper -n install --no-recommends qemu-hw-display-virtio-gpu qemu-hw-display-virtio-gpu-pci
fi
```

Martchus:
Ok, so this is simply about the bootstrapping script then. Judging by https://build.opensuse.org/package/binaries/openSUSE:Factory/qemu/standard the package is still there under that name on Tumbleweed. Which distribution did you run the bootstrapping script on? It would also be good to have any logs of the invocation of the bootstrapping script which didn't install the required package.

And have you double checked whether the package is really missing in your installation (and that installing it manually helped)? Just to be sure the issue hasn't a whole different cause after all.

ssssam:
Hello, thanks for the quick response!

We are using the OpenQA container images from registry.opensuse.org, specifically  devel/openqa/containers15.3/openqa_worker. Is there a way to get build logs for that container without re-running the build locally myself? (will do that tomorrow if needed)

I have verified that the package is missing and installing it fixes the issue.

ssssam:
I tried a local `cd openQA/container/worker; docker build .` and got a new error, seems that it's building against Leap 15.4 but hasn't been fully updated for that , or something ...

```
STEP 1/8: FROM opensuse/leap:15.4
...
Retrieving: os-autoinst-4.6.1662465492.44da018-lp154.1383.1.x86_64.rpm [not found]
File './x86_64/os-autoinst-4.6.1662465492.44da018-lp154.1383.1.x86_64.rpm' not found on medium 'http://download.opensuse.org/repositories/devel:openQA/15.4'
Abort, retry, ignore? [a/r/i/...? shows all options] (a): a
```

So I'm not sure how to work on solving this locally right now.

[...]

PR #4804, opened 2022-09-13
Include virtio VGA support in the openqa_worker container

ssssam:
This was previously available and got removed somewhere in the 15.3 container lifecycle.

The effect on image size is negligible and it is useful to test images with kernels that only support the 'virtio' graphics driver without requiring a custom worker container.

Fixes https://github.com/os-autoinst/openQA/issues/4795

### DIFF
--- a/container/worker/Dockerfile
+++ b/container/worker/Dockerfile
@@ -10,6 +10,7 @@ RUN zypper ar -p 95 -f http://download.opensuse.org/repositories/devel:openQA/15
     zypper in -yl ca-certificates-mozilla curl gzip rsync && \
     zypper in -yl openQA-worker os-autoinst-s390-deps && \
     zypper in -yl qemu-arm qemu-ppc qemu-x86 qemu-tools && \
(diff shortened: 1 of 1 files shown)


Martchus:
The test fails because the base commit is too old. Let's see whether I can rebase it automatically: @Mergifyio rebase

Otherwise the change looks good. I've also read your explanation (https://github.com/os-autoinst/openQA/issues/4795#issuecomment-1245548729).

ssssam:
Got it, thanks! I can rebase from here if needed

kalikiana:
@Mergifyio rebase

Martchus (approved):

baierjan (approved):

ssssam:
In the registry.opensuse.org/devel/openqa/containers15.4/openqa_worker:latest container as of today (ID 043c0737b4f6708b) the qemu-hw-display-virtio-* packages are still not installed. Is some extra step needed to deploy this change ?

ssssam:
That was quick, thanks!

PR #4808, opened 2022-09-20
Reduce two verbose log messages to trace level

okurz:
On big openQA instances with debug log level enabled I can see multiple
log messages every second which is much more verbose than other type of
messages during normal operation.

### DIFF
--- a/lib/OpenQA/WebAPI/Controller/API/V1.pm
+++ b/lib/OpenQA/WebAPI/Controller/API/V1.pm
@@ -12,7 +12,7 @@ sub auth_jobtoken ($self) {
         {'properties.key' => 'JOBTOKEN', 'properties.value' => $token},
         {columns => ['id'], join => {worker => 'properties'}})->single;
     $self->stash('job_id', $job->id);
-    $self->app->log->debug(sprintf('Found associated job %u', $job->id));
+    $self->app->log->trace('Found associated job ' . $job->id);
     return 1;
 }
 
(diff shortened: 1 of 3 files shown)


Martchus (approved):

lib/OpenQA/WebAPI/Controller/API/V1.pm
```diff
@@ -12,7 +12,7 @@ sub auth_jobtoken ($self) {
         {'properties.key' => 'JOBTOKEN', 'properties.value' => $token},
         {columns => ['id'], join => {worker => 'properties'}})->single;
     $self->stash('job_id', $job->id);
-    $self->app->log->debug(sprintf('Found associated job %u', $job->id));
+    $self->app->log->trace("Found associated job $job->id);
```
Martchus, line 15:
```
Can't find string terminator '"' anywhere before EOF at /home/squamata/project/lib/OpenQA/WebAPI/Controller/API/V1.pm line 15.
```

lib/OpenQA/WebAPI/Controller/API/V1.pm
```diff
@@ -12,7 +12,7 @@ sub auth_jobtoken ($self) {
         {'properties.key' => 'JOBTOKEN', 'properties.value' => $token},
         {columns => ['id'], join => {worker => 'properties'}})->single;
     $self->stash('job_id', $job->id);
-    $self->app->log->debug(sprintf('Found associated job %u', $job->id));
+    $self->app->log->trace("Found associated job $job->id");
```
Martchus, line 15: You cannot call a function within `""`.

Martchus (changes requested):

Martchus (approved):

perlpunk (approved):

PR #4810, opened 2022-09-21
Fix the cloning of job when the corresponding variable is manually set

DeepthiYV:
I modified the verification flow of maintenance releases. First apply the manually set variable to settings and then verify the maintenance, later download the assets.
 
Progress Ticket: https://progress.opensuse.org/issues/116788

### DIFF
--- a/lib/OpenQA/Script/CloneJob.pm
+++ b/lib/OpenQA/Script/CloneJob.pm
@@ -248,7 +248,6 @@ sub clone_job ($jobid, $url_handler, $options, $post_params = {}, $jobs = {}, $d
     my $job = $jobs->{$jobid} = clone_job_get_job($jobid, $url_handler, $options);
 
     my $settings = $post_params->{$jobid} = {%{$job->{settings}}};
-    OpenQA::Script::CloneJobSUSE::detect_maintenance_update($jobid, $url_handler, $settings);
     my $clone_children = $options->{'clone-children'};
     my $max_depth = $options->{'max-depth'} // 1;
     for my $job_type (qw(parents children)) {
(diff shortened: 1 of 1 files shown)


DeepthiYV:
@okurz, @foursixnine   Can you please review.

mimi1vx (approved):

lib/OpenQA/Script/CloneJob.pm
```diff
@@ -271,12 +270,12 @@ sub clone_job ($jobid, $url_handler, $options, $post_params = {}, $jobs = {}, $d
...
         }
     }
-
-    clone_job_download_assets($jobid, $job, $url_handler, $options) unless $options->{'skip-download'};
-
     $settings->{CLONED_FROM} = $url_handler->{remote_url}->clone->path("/tests/$jobid")->to_string;
     if (my $group_id = $job->{group_id}) { $settings->{_GROUP_ID} = $group_id }
     clone_job_apply_settings($options->{args}, $relation eq 'children' ? 0 : $depth, $settings, $options);
+    OpenQA::Script::CloneJobSUSE::detect_maintenance_update($jobid, $url_handler, $settings);
+
```
foursixnine, line 277: please remove this blank line

foursixnine:
Failure `Package 'perl-YAML-LibYAML-0.83' not found.` is not related

foursixnine (approved):

Martchus (commented):
The commit message body should reference the progress ticket. It would also nice to give a bit more context, e.g. by adding "This fix relates to the check for maintenance updates which otherwise does not take those variables into account."

grisu48:
> The commit message body should reference the progress ticket. It would also nice to give a bit more context, e.g. by adding "This fix relates to the check for maintenance updates which otherwise does not take those variables into account."

Indeed, a good commit message explains why you did something, not just what you did. This is a courtesy to the reader and your future self. Example:

Cloning of older jobs with already released jobs was impossible, if at least one of those jobs was already released. This commit allow overwriting of maintenance job test issues and thereby makes it possible to remove already released issues and allows to clone older jobs with proper settings in place.

grisu48 (approved):
Looks good to me so far, but I would like to have someone with more tooling experience approving this as well.

foursixnine:
@Martchus how does one solve the [CI failure](https://app.circleci.com/pipelines/github/os-autoinst/openQA/10189/workflows/be1bfce4-0cc3-481c-baf6-917ad79c1999/jobs/95503?invite=true#step-108-76)? `Package 'perl-YAML-LibYAML-0.83' not found.`

Martchus:
The CI is currently broken in two ways:

* Our bot cannot updated dependencies anymore (https://progress.opensuse.org/issues/116593)
* The result of subtests isn't reported correctly (https://progress.opensuse.org/issues/116596)

By rebasing against master one could at least past the current CI failure (and the pending dependency update is likely not required at this point). Then we'd still need to check on CircleCI manually whether all checks have passed.

Martchus (approved):

Martchus:
It now even reports back that the whole workflow was successful. That would also be ok if there weren't the stuck checks for the job level.

grisu48:
Thank you all!

PR #4811, opened 2022-09-21
Replace references of `openqabot` with new bot user `os-autoinst-bot`

Martchus:
See https://progress.opensuse.org/issues/116593

---

Still a draft as I'm not sure whether that will be sufficient and already work.


Martchus:
Not sure whether this will be sufficient or not but I'd actually just merge it and see what breaks and then fix that (considering that the cron PRs are currently in a broken state anyways).

baierjan (approved):

PR #4813, opened 2022-09-22
Improve performance of 3 api endpoints by about 312% (for O3 data)

kraih:
`/api/v1/products`, `/api/v1/machines`, and `/api/v1/test_suites` share the same backend code for fetching data. The current implementation makes thousands of select queries to fetch settings. That is very inefficient, so this patch switches to a join/collapse based prefetch approach. This is slightly more efficient than a plain prefetch, because it avoids the creation of DateTime objects. See https://metacpan.org/pod/DBIx::Class::ResultSet#collapse for more information about the collapse feature.

Progress: https://progress.opensuse.org/issues/94667

### DIFF
--- a/lib/OpenQA/WebAPI/Controller/API/V1/Table.pm
+++ b/lib/OpenQA/WebAPI/Controller/API/V1/Table.pm
@@ -47,24 +47,27 @@ version, arch and flavor parameters are always required.
 
 =cut
 
-my %tables = (
+my %TABLES = (
     Machines => {
         keys => [['id'], ['name'],],
         cols => ['id', 'name', 'backend', 'description'],
         required => ['name', 'backend'],
         defaults => {description => undef},
+        ref_name => 'machine'
     },
     TestSuites => {
         keys => [['id'], ['name'],],
         cols => ['id', 'name', 'description'],
         required => ['name'],
         defaults => {description => undef},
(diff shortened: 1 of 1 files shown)


mimi1vx (approved):

lib/OpenQA/WebAPI/Controller/API/V1/Table.pm
```diff
@@ -96,15 +96,27 @@ sub list {
...
             }
         }
     }
 
     my @result;
+    my $sql_field = {
+        Machines => 'machine',
+        TestSuites => 'test_suite',
+        Products => 'product',
+    }->{$table};
```
Martchus, line 109: I'm wondering whether we can ask DBIx for the raw table name, e.g. maybe https://metacpan.org/pod/DBIx::Class::ResultSource#name would return it. That would be better than hardcoding the mapping here. Otherwise, it could also be done via a generic CamlCase to snake_case conversion.

Martchus (commented):
It looks generally very good, I have only one remark (see inline comment) that might be worthwhile to look into.

Martchus (approved):

perlpunk:
So the only thing I missed in my [original suggestion](https://progress.opensuse.org/issues/94667#note-7) was the order 🤦

PR #4817, opened 2022-09-27
Delete unused route 'sub export' from Test.pm

r-richardson:
as mentioned in [https://progress.opensuse.org/issues/113189](https://progress.opensuse.org/issues/113189)


perlpunk:
Please rebase to current master to fix the CI failure

Martchus (changes requested):
You also need to delete the registration of the route in `WebAPI.pm` (`$r->get('/tests/export')->name('tests_export')->to('test#export');`).

Martchus:
It would make sense to squash both commits together because they only make sense together. Then you can also add a ticket reference to the commit message body (like in the PR description).

Martchus:
It looks like this is even covered by tests. So you needed to remove `t/19-tests-export.t` as well and also a few checks in `t/22-dashboard.t`.

However, we should maybe double check whether the route is really not needed anymore before. I assumed so because it is not part of the public API and not used internally. However, it is actually possibly used via the `plugin_links` feature. That route has been added by  @coolo via 011e5163d564abd199086476fed121a51892fe03. So, @coolo do we still need that route? I don't see `plugin_links` being used on OSD and o3. Note that I wanted to remove it as shortcut for https://progress.opensuse.org/issues/113189 but we can also keep it and implement a limit.

coolo:
That route served one purpose only: to be able to remove old results because it was possible to archive their outcome. a) meanwhile we have different means to preserve old results and b) it was never done, it was just a checkbox item for management.

perlpunk (approved):

t/22-dashboard.t
```diff
@@ -536,17 +536,4 @@ subtest 'job parent groups with multiple version and builds' => sub {
 
 };
 
-subtest 'extra plugin links' => sub {
```
Martchus, line 539: Don't delete the whole subtest. We still want to test the "plugin_links" feature in general.

Martchus (changes requested):

Martchus (approved):

PR #4820, opened 2022-09-28
Support skip and broken results in LTP parser

acerv:
The current runltp-ng JSON parser is not taking into account broken/skipped tests, which are generated by LTP tests results. In particular TSKIP/TCONF and TBROK.


Martchus (dismissed):
It would also be good to extend tests.

asmorodskyi:
TCONF/TSKIP was in LTP for ages , but initially it was ignored . I want to doable check maybe we missing some idea why it was done like that ? Maybe there some pitfall in also taking into account TCONF/TSKIP ?

@richiejp , @pevik , @foursixnine WDYT ?

acerv:
@asmorodskyi the only usage of `OpenQA::Parser::Format::LTP` is inside `tests/publiccloud/run_ltp.pm` implementation.
The `tests/kernel/run_ltp.pm` doesn't use JSON file because LTP logs are parsed instead, as you can see from the following implementation: https://github.com/os-autoinst/os-autoinst-distri-opensuse/blob/d813b90a08e63eadbf28a622d197a3a2a1e2cc9f/tests/kernel/run_ltp.pm#L211

With this modification we are aligning `tests/kernel/run_ltp.pm` and `tests/publiccloud/run_ltp.pm` to the same results in openQA.

pevik:
> @richiejp , @pevik , @foursixnine WDYT ?

@asmorodskyi Did Andrea answer your question?

I'm not familiar with `tests/publiccloud/run_ltp.pm`, but I changed `tests/kernel/run_ltp.pm` in https://github.com/os-autoinst/os-autoinst-distri-opensuse/commit/a760f7d91721ac159a1c2ffdf0b7fbb40320fb8e to handle TCONF as skipped, i.e. gray instead of green and blue border around box, see: https://openqa.opensuse.org/tests/2763684. Public cloud tests does not have it: https://openqa.suse.de/tests/9615085#step/cacheflush01/1. I'd just recommend to follow this approach for better readability (sometimes changing from TPASS to TCONF shows a problem, i.e. missing module which should be available. But that's a separate problem.

asmorodskyi:
> @asmorodskyi the only usage of `OpenQA::Parser::Format::LTP` is inside `tests/publiccloud/run_ltp.pm` implementation. The `tests/kernel/run_ltp.pm` doesn't use JSON file because LTP logs are parsed instead, as you can see from the following implementation: https://github.com/os-autoinst/os-autoinst-distri-opensuse/blob/d813b90a08e63eadbf28a622d197a3a2a1e2cc9f/tests/kernel/run_ltp.pm#L211
> 
> With this modification we are aligning `tests/kernel/run_ltp.pm` and `tests/publiccloud/run_ltp.pm` to the same results in openQA.

ok I see . I can not say that they truly aligned because we still have same logic in two different places ... but it certainly move into right direction :+1:

asmorodskyi:
> > @richiejp , @pevik , @foursixnine WDYT ?
> 
> @asmorodskyi Did Andrea answer your question?
> 

I think yes

asmorodskyi (approved):
LGTM

perlpunk (approved):

lib/OpenQA/Parser/Format/LTP.pm
```diff
@@ -30,7 +30,13 @@ sub parse {
         $t_name =~ s/:/_/g;
 
         $result->{result} = 'ok';
-        $result->{result} = 'fail' if $result->{test}->{result} !~ /pass/i || $result->{status} !~ /pass/i;
+
+        if ($result->{test}->{result} =~ /(fail|brok)/i || $result->{status} =~ /(fail|brok)/i) {
```
foursixnine, line 34:
```suggestion
        if ($result->{test}->{result} =~ /(fail|broken)/i || $result->{status} =~ /(fail|broken)/i) {
```

We are not being charged per character :)

lib/OpenQA/Parser/Format/LTP.pm
```diff
@@ -30,7 +30,13 @@ sub parse {
...
+            $result->{result} = 'fail';
+        }
+        elsif ($result->{test}->{result} =~ /conf/i || $result->{status} =~ /conf/i) {
+            $result->{result} = 'skip';
```
foursixnine, line 38: What does conf mean?

We are not being charged per character :)

foursixnine (changes requested):

lib/OpenQA/Parser/Format/LTP.pm
pevik: No we're not, but LTP apparently was, because it has `TBROK` in result output:

in both C API:

https://github.com/linux-test-project/ltp/blob/7fe464cd8443a99adabd188e098f34e539f41511/include/tst_res_flags.h#L12-L13

and shell API:

https://github.com/linux-test-project/ltp/blob/7fe464cd8443a99adabd188e098f34e539f41511/testcases/lib/tst_test.sh#L104-L105

matching `broken` would miss these `TBROK` results.

lib/OpenQA/Parser/Format/LTP.pm
pevik: `TCONF`. Remember we're parsing LTP API.

lib/OpenQA/Parser/Format/LTP.pm
pevik: NOTE: the legacy LTP API prints `CONF`, `BROK`, new API prints `TCONF`, `TBROK`, that's why missing leading `T`.

pevik: Believe me, the original code was correct, we don't want to introduce a regression :).

lib/OpenQA/Parser/Format/LTP.pm
richiejp: > What does conf mean?

It meas the test started executing, but exited early because the system's configuration will not allow it to complete (e.g. the kernel feature being tested is missing).

richiejp (approved):

pevik (approved):

lib/OpenQA/Parser/Format/LTP.pm
foursixnine: So, if it's a known bug, make sure there's a comment for this, and the other thingie with conf, and I'm good

kalikiana: @acerv Are you still planning to add a comment here?

acerv: @kalikiana I will do it now

Martchus (approved):

lib/OpenQA/Parser/Format/LTP.pm
acerv: @kalikiana done

m-dati (changes requested):
similar check as in _runltp.pm_ lines **37-38** are needed in lines **51-52** too
https://github.com/os-autoinst/openQA/blob/4fc1d568fd3e2395910ab615fac10d16f8f1f654/lib/OpenQA/Parser/Format/LTP.pm#L51-L52
 for $details update, to resolve the issue of External Results red/fail for skipped tests in Details page, changes like:
```
        my $details = {result => 'ok'};
        if ($result->{test}->{result} =~ /(fail|brok)/i || $result->{status} =~ /(fail|brok)/i ) {
            $details = {result => 'fail'}    
        } 
        elsif ($result->{test}->{result} =~ /conf/i || $result->{status} =~ /conf/i) {
            $details = {result => 'skip'}
        };
```

foursixnine (approved):

acerv:
@m-dati thanks for your suggestion. Please check if code is ok. I added details update in order to have a proper `External results` page in openQA.

acerv:
@m-dati I seen the results and I can consider this task done from my side.
@kalikiana Can we merge this patch?

Martchus:
The change will be merged automatically when tests are passing. Maybe you just need to rebase against master to make them work again.

okurz (changes requested):

acerv:
@okurz done

okurz (approved):

pevik (approved):

okurz:
So the one check failing is "codecov/patch — 33.33% of diff hit (target 100.00%) " meaning that unit test coverage for the changed code sections should be added. @acerv can you look into extending tests accordingly in t/30-test_parser.t?

acerv:
@okurz if you take a look at the latest modifications, I have added the coverage for that part of the code. I'm stucking with tidy atm, but I'll handle it (hopefully)

PR #4827, opened 2022-09-29
Limit listing of job_templates to avoid timeouts

r-richardson:
We already have a limit for the 'All tests' pages, this implementation follows the same approach.
Might need adjustments based on performance tests with production data.

see: https://progress.opensuse.org/issues/114421

### DIFF
--- a/lib/OpenQA/WebAPI/Controller/API/V1/JobTemplate.pm
+++ b/lib/OpenQA/WebAPI/Controller/API/V1/JobTemplate.pm
@@ -4,7 +4,9 @@
 package OpenQA::WebAPI::Controller::API::V1::JobTemplate;
 use Mojo::Base 'Mojolicious::Controller';
 use Try::Tiny;
+use OpenQA::App;
 use OpenQA::YAML qw(load_yaml dump_yaml);
+use List::Util qw(min);
 
 =pod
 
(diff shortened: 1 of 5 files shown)


Martchus (approved):

Martchus:
I've approved but it would still be great to actually do performance tests with OSD or o3 data and possibly tweak the default limits.

lib/OpenQA/WebAPI/Controller/API/V1/JobTemplate.pm
```diff
@@ -61,21 +64,8 @@ sub list {
...
-                my $attrs
-                  = {join => ['machine', 'test_suite', 'product'], prefetch => [qw(machine test_suite product)]};
-                @templates = $schema->resultset("JobTemplates")->search(\%cond, $attrs);
-            }
-            else {
-                @templates
-                  = $schema->resultset("JobTemplates")->search({}, {prefetch => [qw(machine test_suite product)]});
-            }
+            my %attrs = (prefetch => [qw(machine test_suite product)], rows => $limit);
+            @templates = $schema->resultset("JobTemplates")->search(\%cond, \%attrs);
```
perlpunk, line 68: Like I said in the programming session, I double-checked the `$has_query` variable and can't see any conditions when there would be keys with empty values in the `%cond` hash, and if `%cond` itself is completely empty that should also not be a problem

PR #4832, opened 2022-10-06
Add help popover for limiting number of finished jobs

Martchus:
See https://progress.opensuse.org/issues/111770


okurz (approved):

kalikiana (approved):
Maybe an actual GUI option would be nicer? But I guess that's the easiest way for now. So why not.

PR #4835, opened 2022-10-07
Clean stale lockfiles when fetching needles

Martchus:
Git refuses to override an existing lock file, even if no process seems to use it anymore. When our server crashes such a stale lockfile often remains as leftover. This change will remove the stale lockfile automatically. It will of course keep the lock file in-place if it is still owned by at least one process.

See https://progress.opensuse.org/issues/116911

### DIFF
--- a/script/fetchneedles
+++ b/script/fetchneedles
@@ -64,8 +64,17 @@ needlesdir() {
     fi
 }
 
+deal_with_stale_lockfile() {
+    lock_path=.git/index.lock
+    if [ -e "$lock_path" ] && ! fuser --silent "$lock_path"; then  # fuser returns 0 if one or more processes hold the lock
+        echo "removing stale lock $PWD/$lock_path"
+        rm "$lock_path"
+    fi
+}
+
 git_update() {
     branch="${1:-"master"}"
+    deal_with_stale_lockfile
     git gc --auto --quiet
     git fetch -q origin
     # Clear any uncommitted changes that would prevent a rebase


Martchus:
By the way, I've tested this locally:
```
$ flock .git/index.lock sleep infinity
```

Then in another shell:
```
$ LANG=C git add foo
fatal: Unable to create '/hdd/openqa-devel/repos/openQA/.git/index.lock': File exists.

Another git process seems to be running in this repository, e.g.
an editor opened by 'git commit'. Please make sure all processes
are terminated then try again. If it still fails, a git process
may have crashed in this repository earlier:
remove the file manually to continue.
```
```
$ if [ -e "$lock_path" ] && ! fuser --silent "$lock_path"; then echo "removing stale lock $PWD/$lock_path" ; else echo "not touching lockfile" ; fi
not touching lockfile
```

Then, after terminating the first process, Git still refuses to continue but we would remove the lockfile:
```
$ LANG=C git add foo
fatal: Unable to create '/hdd/openqa-devel/repos/openQA/.git/index.lock': File exists.
…
```
$  if [ -e "$lock_path" ] && ! fuser --silent "$lock_path"; then echo "removing stale lock $PWD/$lock_path" ; else echo "not touching lockfile" ; fi
removing stale lock /hdd/openqa-devel/repos/openQA/.git/index.lock
```

baierjan (approved):

okurz (approved):

PR #4836, opened 2022-10-11
Use signatures in lib/OpenQA/LiveHandler.pm

asdil12:
Ticket: https://progress.opensuse.org/issues/100967


okurz (approved):

Martchus (commented):
Let's see whether tests will pass. However, I suspect that `log_name` and `schema` might not actually be called without arguments.

kalikiana:
Too many arguments for subroutine 'OpenQA::LiveHandler::schema' at /home/squamata/project/lib/OpenQA/LiveHandler.pm line 11.

Martchus (approved):

PR #4841, opened 2022-10-12
Avoid test failures due to localized error messages

Martchus:
See particular commit messages

---

Maybe we should merge the different sections within our documentation where we explain how to run openQA's unit tests. For the sake of the PR I've just adjusted both places (as it is not a big change and relevant to both sections).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -35,7 +35,7 @@ endif
 TIMEOUT_RETRIES ?= $$((${TIMEOUT_M} * ${SCALE_FACTOR} * (${RETRY} + 1) ))m
 CRE ?= podman
 # avoid localized error messages (that are matched against in certain cases)
-LANG = C.utf8
+LC_ALL = C.utf8
 LANGUAGE =
 mkfile_path := $(abspath $(lastword $(MAKEFILE_LIST)))
 current_dir := $(patsubst %/,%,$(dir $(mkfile_path)))
--- a/Makefile
+++ b/Makefile
@@ -34,6 +34,9 @@ SCALE_FACTOR ?= 2
 endif
 TIMEOUT_RETRIES ?= $$((${TIMEOUT_M} * ${SCALE_FACTOR} * (${RETRY} + 1) ))m
 CRE ?= podman
+# avoid localized error messages (that are matched against in certain cases)
+LANG = C.utf8
+LANGUAGE =
 mkfile_path := $(abspath $(lastword $(MAKEFILE_LIST)))
 current_dir := $(patsubst %/,%,$(dir $(mkfile_path)))
 container_env_file := "$(current_dir)/container.env"
--- a/docs/Contributing.asciidoc
+++ b/docs/Contributing.asciidoc
@@ -226,7 +226,7 @@ it when running tests. Example:
 
 [source,sh]
 ----
-TEST_PG='DBI:Pg:dbname=openqa_test;host=/dev/shm/tpg' OPENQA_BASEDIR= prove -v t/14-grutasks.t
+TEST_PG='DBI:Pg:dbname=openqa_test;host=/dev/shm/tpg' OPENQA_BASEDIR= LANG=C.utf8 LANGUAGE= prove -v t/14-grutasks.t
 ----
 
 In the case of wanting to tweak the tests as above, to speed up the test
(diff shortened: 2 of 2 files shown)


docs/Contributing.asciidoc
```diff
@@ -585,7 +585,13 @@ temporary PostgreSQL database used for testing. To do this step manually run
 `t/test_postgresql /dev/shm/tpg` to initialize a temporary PostgreSQL database
 and export the environment variable as instructed by that script.
 It is also possible to run a particular test, for example
-`prove t/api/01-workers.t`.
+`prove t/api/01-workers.t`. When using `prove` directly, make sure an English
+locale is set (e.g. `export LANG=C` before initializing the database and running
+`prove`).
+
+To keep the test database running after executing tests with the `Makefile`, add
+`KEEP_DB=1` to the make arguments. To access the test database, use
```
okurz, line 593: we already mention `KEEP_DB` elsewhere but I don't mind

okurz (approved):

docs/Contributing.asciidoc
Martchus: I know. But in this context it is relevant as well. (We likely should really merge the two documentation sections at some point.)

perlpunk (approved):

okurz (dismissed):
I did not follow the complete discussion today but you mentioned "unicode". The test falures look like we should actually use that, WDYT?

Martchus:
Yes. Strangely on my local TW system it works *only* with `LANG=C` but not with the English Unicode locale. Likely a local problem then. (In my helpers I've been previously already working around the problem and had to change from Unicode to C for it to continue to work on my TW system, see https://github.com/Martchus/openQA-helper/commit/f77f4ced9013f4578a2d911f37e076abec24952c. And @asdil12 's TW behaved the same way. I've also tried `en_US.UTF-8` but it didn't work either.)

kalikiana:
> Yes. Strangely on my local TW system it works _only_ with `LANG=C` but not with the English Unicode locale. Likely a local problem then. (In my helpers I've been previously already working around the problem and had to change from Unicode to C for it to continue to work on my TW system, see [Martchus/openQA-helper@f77f4ce](https://github.com/Martchus/openQA-helper/commit/f77f4ced9013f4578a2d911f37e076abec24952c). And @asdil12 's TW behaved the same way. I've also tried `en_US.UTF-8` but it didn't work either.)

There must be something more to it because I use **LANG=en_US.UTF-8** and I didn't even see what Dominik's problem was when I tried to reproduce it yesterday.

perlpunk:
For me tests (t/api/04-jobs.t, t/api/09-comments.t) are currently failing no matter what locale I use. `LANG=C`, `LANG=de_DE.UTF-8`, `LANG=en_US.UTF-8`
It has been working at some point, but I updated a large number of packages a few days ago.

Martchus:
@perlpunk I don't know how you tested, see my messages in slack. That this breaks UTF-8 I actually also can confirm locally (but at the same time it solves the DB language). If you use my helpers, make sure you have reverted https://github.com/Martchus/openQA-helper/commit/f77f4ced9013f4578a2d911f37e076abec24952c locally.

Not sure how to configure a TW system to have a non-English language as a primary language but still allow other locales to be set via `LANG` so I've asked in the openSUSE Factory chat. By the way, the problem can easily reproduced by simply invoking `LANG=en_US.UTF-8 git status` locally on my TW system. It'll ignore `LANG` (and still print German output) except for `LANG=C` (which then breaks Unicode for some applications as we see in the test failures here).

Maybe this should be `C.UTF-8` to *not* break Unicode but to also avoid using `en_US` specifically.

perlpunk:
I'm testing with this:
```
make test TESTS=t/api/04-jobs.t
```

Martchus:
Ok, one needs to clear `LANGUAGE` as well. For some reason that variable is set in my env as well and used by the DB. But if it isn't set than it falls back to `LANG` so that still needs to be set.

Makefile
```diff
@@ -34,6 +34,9 @@ SCALE_FACTOR ?= 2
 endif
 TIMEOUT_RETRIES ?= $$((${TIMEOUT_M} * ${SCALE_FACTOR} * (${RETRY} + 1) ))m
 CRE ?= podman
+# avoid localized error messages (that are matched against in certain cases)
+LANG = C.utf8
+LANGUAGE =
```
okurz, line 39: According to https://www.gnu.org/software/gettext/manual/html_node/Locale-Environment-Variables.html#Locale-Environment-Variables one should be able to use `LC_ALL` instead of needing to specify both LANG+LANGUAGE

okurz (changes requested):

Makefile
Martchus: This doesn't work because `LANGUAGE` is still used when it is set (unless `LANG` or `LC_ALL` is set to `C`). I've tested that so `LANGUAGE` definitely needs to be unset manually. However, I now changed it to `LC_ALL` instead of `LANG` because it has indeed a higher precedence than `LANG` and is also used by autotools.

okurz (approved):

PR #4844, opened 2022-10-18
Avoid alerts by not failing Minion jobs when downloads fail

Martchus:
* This affects only expected download errors (connection errors, error
  responses) but not any kind of unhandled exception or setup related
  errors.
* See https://progress.opensuse.org/issues/118969


Martchus:
Also works for skipped jobs:

screenshot_20221018_125308

perlpunk (approved):

baierjan (approved):

PR #4847, opened 2022-10-19
codecov.yml: Introduce specific coverage thresholds

okurz:
With path specific coverage targets we can introduce the hard
requirement for 100% coverage within the paths where it is feasible. It
makes sense to have 100% statement coverage within test code itself to
prevent relying on tests that are actually not running.

Previously tried in
* https://github.com/os-autoinst/openQA/pull/3938

In os-autoinst we already added according changes in
* https://github.com/os-autoinst/os-autoinst/pull/2016

See https://docs.codecov.com/docs/commit-status#branches

### DIFF
--- a/codecov.yml
+++ b/codecov.yml
@@ -11,8 +11,11 @@ coverage:
     changes: false
     project:
       default:
-        target: 98.0
-        threshold: 0.1
+        target: 98
+      tests:
+        target: 99.48
+        paths:
+          - t/
     patch:
       default:
         target: 100.0
--- a/codecov.yml
+++ b/codecov.yml
@@ -7,9 +7,6 @@ comment:
 fixes:
   - "/home/circleci/openQA::"
 coverage:
-  range: 60..95
-  round: down
-  precision: 2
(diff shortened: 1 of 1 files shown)


okurz:
I don't understand what I am doing wrong. Why isn't there a check for codecov "tests" showing up like we have in https://github.com/os-autoinst/os-autoinst/pull/2016#event-6369295413?

Martchus:
The only difference compared to my change for os-autoinst I can see is that `paths` is no array in your case (but just a single path). Maybe this is also only effective after being merged?

okurz:
Cool. It seems this made a difference. I think it works as expected and it shows that we have currently 99.48% coverage in t/ which is below the target of 100%. I now changed the target level for now to 99.48% to ensure the current state as baseline.

okurz:
Oh wait, one commit is called WIP

Martchus (approved):

Martchus (approved):

baierjan (approved):

Issue #4576, opened 2022-03-22
NEEDLES_DIR documentation and/or behaviour could be improved


phil-hands:
I have needles in the same repo as the tests for the Debian tests (using _git-lfs_ to deal with the bloat, which seems to work well BTW).

If I set the `CASEDIR` to be a branch or commit, it fails with a complaint about not finding the thing it's looking for in the repo.

In that case, it creates a symlink in the pool/N directory for the needles, to the default location for the needles repo, which of course does not match the needles in the `CASEDIR` repo.

If I set `NEEDLES_DIR=products/debian/needles` then it works.

[The Documentation](https://github.com/os-autoinst/openQA/blob/master/docs/WritingTests.asciidoc#triggering-tests-based-on-an-any-remote-git-refspec-or-open-github-pull-request) says:

> Note that customizing CASEDIR does not mean needles will be loaded from there, even if the repository specified as CASEDIR contains needles. To load needles from that repository, it needs to be specified as NEEDLES_DIR as well.

which suggests that one should copy the git URL into both variables, but that doesn't work (and strikes me as a bit pointless TBH).

Given this [comment about `NEEDLES_DIR`, `CASEDIR` & `PRODUCTDIR`](https://github.com/os-autoinst/openQA/blob/master/lib/OpenQA/Worker/Engines/isotovideo.pm#L407) it seems like the current behaviour is intended, and that one _should_ just set the relative path for `NEEDLES_DIR`, in which case I think that should be made more explicit in the docs.

Alternatively, it strikes me that since the thing that one is setting `NEEDLES_DIR` to is just `$PRODUCTDIR/needles` there really ought to be a way of making it do that automatically, without having to specify the `PRODUCTDIR` value twice, but that probably needs to be dependent upon whether the normal repo which the `CASEDIR` is almost certainly forked is also an all-in-one or a needles_separate repo.

okurz:
Now tracked additionally as https://progress.opensuse.org/issues/117133 . Also I created https://progress.opensuse.org/issues/117136 to prevent the stale bot to close issues which are still valid.

Martchus:
The [docs](https://github.com/os-autoinst/openQA/blob/master/docs/WritingTests.asciidoc#triggering-tests-based-on-an-any-remote-git-refspec-or-open-github-pull-request) just say that one is suppose to set `NEEDLES_DIR` but not to what in which use case. That should certainly be improved.

[...]

PR #4851, opened 2022-10-19
Allow specifying `NEEDLES_DIR` to be relative to `CASEDIR`

Martchus:
* Allows prepending `%CASEDIR%` to relative `NEEDLES_DIR` so `%CASEDIR%` is replaced with the path where the custom `CASEDIR` will be checked out
* Enables use-case of specifying a custom `CASEDIR` and a custom `NEEDLES_DIR` where the custom needles dir is a relative path within the custom `CASEDIR` (and *not* a separate repository or a relative path within the default needles directory).
* See https://progress.opensuse.org/issues/117133
* See https://github.com/os-autoinst/openQA/issues/4576

### DIFF
--- a/docs/WritingTests.asciidoc
+++ b/docs/WritingTests.asciidoc
@@ -1619,6 +1619,26 @@ See
 https://github.com/os-autoinst/os-autoinst/blob/master/doc/backend_vars.asciidoc
 for details.
 
+[NOTE]
+====
+The openQA worker normally default-initializes `CASEDIR` and `NEEDLES_DIR` to
+point to default repositories provided by the openQA instance. This behavior
+interacts with specifying a custom `CASEDIR` or `NEEDLES_DIR` in the following
+way:
+
+* If `CASEDIR` or `NEEDLES_DIR` is customized the customized location is used
+  instead of the default repository.
+* If only one of `CASEDIR` or `NEEDLES_DIR` is customized the other variable
+  will still be initialized to point to the default repository.
+* A relative `NEEDLES_DIR` is treated to be relative to the default `CASEDIR`
+  (even if `CASEDIR` is customized). To have it treated to be relative to the
+  custom `CASEDIR`, prefix the relative path with `%CASEDIR%/`. So specifying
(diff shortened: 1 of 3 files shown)


docs/WritingTests.asciidoc
```diff
@@ -1605,20 +1605,39 @@ describes how downloadable assets can be specified. It is important to note
...
-intro a production repository and branch. This works by setting the
+into a production repository and branch. This works by setting the
 `CASEDIR` parameter of os-autoinst to a valid Git repository path including
 an optional branch/refspec specifier. It is also possible to set `NEEDLES_DIR`
 to a valid Git repository path to use custom needles.
 See
 https://github.com/os-autoinst/os-autoinst/blob/master/doc/backend_vars.asciidoc
 for details.
 
+[NOTE]
```
okurz, line 1622: @phil-hands I would appreciate your feedback if the following documentation section would be clear enough to you to call https://github.com/os-autoinst/openQA/issues/4576 fixed

okurz (approved):

docs/WritingTests.asciidoc
phil-hands: _Regarding the PR in general:_  Great! -- I'm very much looking forward to getting it packaged, to be able to use it :-)

_Regarding the docs:_ I'm not sure that for a user who hasn't yet looked at the code, the relevance of what  `NEEDLES_DIR` ends up being set to will immediately reveal what that allows them to do,

How about adding something like:

> ... with the result that the needles that would then be used would be those to be found in `https://github.com/.../tree/master/the-needles`

phil-hands:
I hesitated to word the concluding sentence like that when I was trying to come up with something, because e.g. the `/master/` bit of that URL isn't _definitely_ correct under all circumstances, but other than that I think that's great, and if you are OK with being mostly right-ish rather than absolutely correct, that's fine with me too :-)

Martchus:
You're right. I replaced the "master" part with "…".

kalikiana (approved):

PR #4854, opened 2022-10-21
Add distri-opensuse specific worker container dockerfile

asdil12:
Ticket: https://progress.opensuse.org/issues/117631
VR: https://build.opensuse.org/package/show/devel:openQA/openQA_container_worker_distri_opensuse


Martchus (approved):

container/worker/run_openqa_worker.sh
```diff
@@ -25,6 +25,10 @@ qemu-system-x86_64 -S &
 kill $!
 
 # Install test distribution dependencies
-find -L "/var/lib/openqa/share/tests" -maxdepth 2 -type f -executable -name 'install_deps.*' -exec {} \;
+if [[ -z $TEST_DISTRI_DEPS ]] ; then
```
perlpunk, line 28: if you quote the variable, you can remove the SC2086 again, I think

asdil12: The error was not in this line but in the zypper command line where I didn't quote the var to allow installing multiple packages.

Martchus: Yes. As far as I know you don't need to quote this variable (as `[[` is used) and shellcheck was indeed *only* complaining about the expansion of the variable in the zypper line.

perlpunk: Yeah, I should not do reviews when sitting in a street car ;-)

baierjan (approved):

PR #4856, opened 2022-10-22
Show correct job status in worker job list

perlpunk:
Issue: https://progress.opensuse.org/issues/119110

The renderTestName function expected a running to job to have an undefined or non-existant`result` entry, which was the case for the test lists on the "All tests" page, but not for the job list on a worker page.
That's why it was shown with "Done: none".
Additionally it showed a restart button, which made the test in 15-admin-workers.t restart the first job 99961, although it is running.

Also the cancel_url variable was never set.


lib/OpenQA/WebAPI/Controller/Admin/Workers.pm
```diff
@@ -90,7 +90,7 @@ sub previous_jobs_ajax ($self) {
                         id => $job_id,
                         name => $job->name,
                         deps => $job->dependencies,
-                        result => $job->result,
+                        result => ($job->result // 'none') eq 'none' ? undef : $job->result,
```
Martchus, line 93: You could've use constants here.

perlpunk: Ah right.
I think passing `result` in every case (also in the "All tests" queries) could be even a better solution...

baierjan (approved):

Martchus (approved):

perlpunk:
it's ready now :)

kraih (approved):

PR #4861, opened 2022-10-24
Add darkmode

kraih:
This PR adds theme settings to #4724. The CSS has been reorganised to work with a `.darkmode` class on the `<body>` element. And there is a new "Appearance" menu entry for all logged in users:

menu

The default theme is `light`, same as it used to be. And it can be switched manually to `dark` or browser detection (based on Chrome flag):

appearance

Since this feature is only available for logged in users, we can store the theme setting in the session cookie, together with login information. That means we have no additional privacy considerations regarding cookies.

P.S.: I did not make any changes to the darkmode theme itself. There are still some rough edges, but since it is completely opt-in there should be no disruptions for our users.

Progress: https://progress.opensuse.org/issues/119032

### DIFF
--- a/docs/UsersGuide.asciidoc
+++ b/docs/UsersGuide.asciidoc
@@ -550,6 +550,20 @@ the top of the group overview page. However, it only works as operator
 or admin.
 
 
+=== Dark mode
+
+The dark mode is natively supported by most modern browsers and
+is automatically turned on when dark user interface of either
+the operating system or browser is detected. To disable the dark
+mode please follow the instructions below:
+
+ * On Firefox, go to `about:preferences#general` and search for "Website appearance".
+ * On Chrome, go to `chrome://flags/` and search for "Dark mode".
+
+For more information, see
+https://developer.mozilla.org/en-US/docs/Web/CSS/@media/prefers-color-scheme[developer.mozilla.org/CSS/@media/prefers-color-scheme]
+
+
 === Developer mode ===
 
 The developer mode allows to:
--- a/docs/UsersGuide.asciidoc
+++ b/docs/UsersGuide.asciidoc
@@ -552,10 +552,9 @@ or admin.
 
 === Dark mode
 
-The dark mode is natively supported by most modern browsers and
-is automatically turned on when dark user interface of either
-the operating system or browser is detected. To disable the dark
-mode please follow the instructions below:
+A dark mode theme can be enabled via "Appearance" settings for all logged in users. It can either be forced with
+the "dark mode" setting, or left to browser detection. Switching automatically between light and dark mode is natively
+supported by most modern browsers and can also be controlled manually via flags:
 
  * On Firefox, go to `about:preferences#general` and search for "Website appearance".
  * On Chrome, go to `chrome://flags/` and search for "Dark mode".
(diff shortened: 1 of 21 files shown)


baierjan (approved):

pdostal (approved):
This looks clear to me.

The mechanism will enable 'some' dark theme and users could improve it slowly over time.

Thank you for your help @kraih

docs/UsersGuide.asciidoc
```diff
@@ -550,6 +550,19 @@ the top of the group overview page. However, it only works as operator
...
 
 
+=== Dark mode
+
+A dark mode theme can be enabled via "Appearance" settings for all logged in users. It can either be forced with
+the "dark mode" setting, or left to browser detection. Switching automatically between light and dark mode is natively
+supported by most modern browsers and can also be controlled manually via flags:
+
+ * On Firefox, go to `about:preferences#general` and search for "Website appearance".
+ * On Chrome, go to `chrome://flags/` and search for "Dark mode".
```
Martchus, line 560: I don't think setting chrome://flags will trigger the dark mode you're introducing here. At least on my system it seems to only enforce some Chromium-provided CSS overrides. So maybe it doesn't make sense to point it our here.

lib/OpenQA/WebAPI/Controller/Appearance.pm
```diff
@@ -0,0 +1,23 @@
+# Copyright 2022 SUSE LLC
+# SPDX-License-Identifier: GPL-2.0-or-later
+
+package OpenQA::WebAPI::Controller::Appearance;
+use Mojo::Base 'Mojolicious::Controller', -signatures;
+
+sub index ($self) {
+    $self->render;
+}
```
Martchus, line 9: Is a function that only calls `$self->render` actually required?

templates/webapi/layouts/bootstrap.html.ep
```diff
@@ -34,7 +34,11 @@
             % if (current_user) {
               newFeature(<%= current_user->feature_version %>);
             % }
-          } );
+
+            let theme = '<%= current_theme %>';
+            if (theme === 'detect') theme = window.matchMedia('(prefers-color-scheme: dark)').matches === true ? 'dark' : 'light';
+            if (theme === 'dark') document.querySelector('body').classList.add('darkmode');
```
Martchus, line 40:
```suggestion
            if (theme === 'dark') document.body.classList.add('darkmode');
```

Martchus (commented):
Looks generally good, I have only minor comments.

docs/UsersGuide.asciidoc
pdostal: Did you refresh the page after you changed that settings?

lib/OpenQA/WebAPI/Controller/Appearance.pm
kraih: No, but it's the "correct" way to deal with it. I'm moving away from relying on auto-rendering, since it's a feature Mojolicious is soft-deprecating with the introduction of mojo.js.

docs/UsersGuide.asciidoc
kraih: I've tested this in the Chrome Inspector, it has an override feature for `prefers-color-scheme` where you can toggle back and forth. So i'm very sure that it works as it should.

Martchus: Ah, nice. Then it affects the media query despite also seemingly enforcing its own CSS rules on pages with no darkmode support at all. (Maybe it is also now already different since the time I tested it.)

PR #4862, opened 2022-10-24
Fix cache service host config for IPs != 127.0.0.1

asdil12:
Ticket: https://progress.opensuse.org/issues/117631

### DIFF
--- a/lib/OpenQA/CacheService/Client.pm
+++ b/lib/OpenQA/CacheService/Client.pm
@@ -13,6 +13,7 @@ use OpenQA::Utils qw(base_host service_port);
 use Socket qw(AF_INET IPPROTO_TCP SOCK_STREAM pack_sockaddr_in inet_aton);
 use Mojo::URL;
 use Mojo::File 'path';
+use Regexp::Common qw(net);
 
 # Define sensible defaults to cover even a restarting openQA webUI host being
 # down for up to 5m
@@ -22,6 +23,7 @@ has worker_settings => sub ($self) { OpenQA::Worker::Settings->new };
 has host => sub ($self) {
     $self->worker_settings->global_settings->{CACHESERVICEURL} // ('http://127.0.0.1:' . service_port('cache_service'));
 };
+sub actual_host ($self) { return Mojo::URL->new($self->host)->host }
 has cache_dir => sub ($self) { $ENV{OPENQA_CACHE_DIR} || $self->worker_settings->global_settings->{CACHEDIRECTORY} };
 has ua => sub ($self) {
     my $ua = Mojo::UserAgent->new(inactivity_timeout => 300);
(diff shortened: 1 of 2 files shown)


Martchus (approved):

lib/OpenQA/CacheService/Client.pm
```diff
@@ -71,6 +73,10 @@ sub enqueue ($self, $request) {
     return undef;
 }
 
+sub _is_ipv4_addr ($ip) {
+    return $ip =~ /^(?<!\d)(\d{1,3})\.(\d{1,3})\.(\d{1,3})\.(\d{1,3})(?!\d)$/;
```
perlpunk, line 77: Why the `(?<!\d)` lookbehind? The anchor `^` at the beginning would make that impossible anyway, right?
Or can you give an example?
Same for the `(?!\d)` lookahead.

perlpunk: An alternative:
```perl
use Regexp::Common qw(net);
return $ip =~ m/$RE{net}{IPv4}/;
```

asdil12: fixed.

perlpunk (approved):

kraih (approved):
Not really a fan of `actual_host`, but can't think of something better either.

lib/OpenQA/CacheService/Client.pm
```diff
@@ -31,18 +32,19 @@ has ua => sub ($self) {
...
+        addr => pack_sockaddr_in(service_port('cache_service'), inet_aton($self->actual_host)));
     $ua->socket_options->{PeerAddrInfo} = [\%cache_service_address] if $ua->can('socket_options');
     return $ua;
 };
 
 sub set_port ($self, $port) {
-    $self->ua->socket_options->{PeerAddrInfo}->[0]->{addr} = pack_sockaddr_in($port, inet_aton('127.0.0.1'))
-      if $self->ua->can('socket_options');
+    $self->ua->socket_options->{PeerAddrInfo}->[0]->{addr} = pack_sockaddr_in($port, inet_aton($self->actual_host))
+      if $self->ua->can('socket_options') && _is_ipv4_addr($self->actual_host);
```
kraih, line 47: The `if $self->ua->can('socket_options')` check is obsolete now since we depend on Mojolicious 9.20. You could just remove that.

lib/OpenQA/CacheService/Client.pm
```diff
@@ -71,6 +73,10 @@ sub enqueue ($self, $request) {
     return undef;
 }
 
+sub _is_ipv4_addr ($ip) {
+    return $ip =~ /^(\d{1,3})\.(\d{1,3})\.(\d{1,3})\.(\d{1,3})$/;
+}
```
kraih, line 78: We already use `Regexp::Common`, so you can use https://metacpan.org/pod/Regexp::Common::net.

kraih (changes requested):
No need to use a custom IPv4 regex, we already use `Regexp::Common`.

kraih (approved):

PR #4865, opened 2022-10-25
Bring back the old light theme

kraih:
https://github.com/os-autoinst/openQA/commit/28fa7fdd591fac0a0c41a51b03d9281ebc005426 introduced some changes to the default theme. This patch reverts them all, so the default light theme should look the same again. It does make darkmode look worse, but there is no way around that if we want to fix lightmode quickly.

light

### DIFF
--- a/assets/stylesheets/comments.scss
+++ b/assets/stylesheets/comments.scss
@@ -115,9 +115,6 @@
     align-items: center;
 }
 
-.input-group-text {
-    background-color: $default-bg-color;
-}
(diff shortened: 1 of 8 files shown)


perlpunk (approved):

kraih:
Darkmode looks less good now, no way around that though:

dark

okurz (approved):

Martchus (approved):

pdostal (approved):

PR #4866, opened 2022-10-25
Fix jumbotron and card background in darkmode

kraih:
**Edit: Outdated screenshots, see last post!**

Before:
before

After:
after

I'm starting small. This makes the jumbotron and bootstrap cards visible again.

### DIFF
--- a/assets/stylesheets/overall.scss
+++ b/assets/stylesheets/overall.scss
@@ -104,6 +104,7 @@ h1, h2, h3, h4, h5, h6, .h1, .h2, .h3, .h4, .h5, .h6 {
 .darkmode {
     .card {
         background-color: $default-bg-color-darktheme;
+        border: $table-border-color-darktheme solid 0.5px;
     }
 }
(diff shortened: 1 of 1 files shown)


kraih:
It's very subtle. Anything lighter would clash with the SUSE logo.

kraih:
Appearance page, before:
before

And after:
after

The other card elements need styling too, but it's already more legible.

kraih:
The contrast when combined with tables is a bit high, maybe tables should be toned down a bit:

keys

Martchus (commented):
It is an improvement at least. Maybe it'll be nice to introduce another color definition for this (a variable like `$default-bg-color-darktheme`) instead of repeating the color.

kraih:
Another alternative would be that we go for borders in dark mode:
border1
border2

It's a different design language than light mode, but personally i like it more.

kraih:
I'm changing the proposal a little bit, since i like the jumbotron with a background, and cards with a border.
both1
both2

perlpunk (approved):

Martchus (approved):

PR #4867, opened 2022-10-25
Add button to insert marker for unblocking incident to OSD branding

Martchus:
Related ticket: https://progress.opensuse.org/issues/107923

### DIFF
--- a/templates/webapi/branding/openqa.suse.de/commenting_tools.html.ep
+++ b/templates/webapi/branding/openqa.suse.de/commenting_tools.html.ep
@@ -1,0 +1,5 @@
+% if (!$group_comment && !$job->is_ok) {
+    <a class="help_popover fa fa-unlink" title="Add marker so the specified maintenance incident will be approved despite this job"
+       role="button" data-template="@review:acceptable_for:incident_<incident_id>:<reason>" onclick="insertTemplate(this)"></a>
+    <br>
+% }
--- a/templates/webapi/comments/add_comment_form_groups.html.ep
+++ b/templates/webapi/comments/add_comment_form_groups.html.ep
@@ -54,6 +54,7 @@
                        data-template="label:force_result:new_result[:description_or_bugref]" onclick="insertTemplate(this)"></a>
                 % }
                 <br>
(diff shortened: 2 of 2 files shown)


kalikiana (approved):

okurz (changes requested):
Nice! Can we see a screenshot for that and a live session demo-ing? Is it possible to add some more help text somewhere?

Martchus:
For everyone too lazy to checkout the branch locally:
grafik

Making the help popover extendable (for more help text) would likely be a bit hacky.

okurz:
> Making the help popover extendable (for more help text) would likely be a bit hacky.

Sure. I am just afraid that the text "Add marker to unblock incident" won't be enough for some users to understand that. Maybe just extend that line a bit. I know that providing a link would be tricky because people can't click on the hovering text.

kraih:
Do we not have test coverage for those buttons yet?

Martchus:
For the buttons in general, yes. (But not for all of them as they're just using the same JS function anyways.)

kraih (approved):

templates/webapi/branding/openqa.suse.de/commenting_tools.html.ep
```diff
@@ -0,0 +1,5 @@
+% if (!$group_comment && !$job->is_ok) {
+    <a class="help_popover fa fa-unlink" title="Add marker so the specified maintenance incident will be approved despite this job"
```
okurz, line 2: The comment is not really correct because if other tests still fail the maintenance incident won't be approved but fine for now

okurz (approved):

templates/webapi/branding/openqa.suse.de/commenting_tools.html.ep
Martchus: I thought the "despite this job" would be sufficient.

okurz:
> Do we not have test coverage for those buttons yet?

Learning from https://progress.opensuse.org/issues/119467 it turns out we should have been a bit more persistent regarding test coverage ;)

PR #4868, opened 2022-10-25
Add default and max limits to job_next_previous_ajax

r-richardson:
This limit implementation follows the same approach as the default and maximum limits for the 'All tests' page as well as the limits for the API call listing JobTemplates

see: https://progress.opensuse.org/issues/114421

### DIFF
--- a/etc/openqa/openqa.ini
+++ b/etc/openqa/openqa.ini
@@ -212,6 +212,14 @@ concurrent = 0
 #list_templates_default_limit = 500
 # Maximum number of templates to include in job_templates api response (to prevent performance issues)
 #list_templates_max_limit = 5000
+# Default number of next jobs to include in jobs request (to prevent performance issues)
+#next_jobs_default_limit = 100
+# Maximum number of next jobs to include in jobs request (to prevent performance issues)
+#next_jobs_max_limit = 1000
+# Default number of previous jobs to include in jobs request (to prevent performance issues)
+#previous_jobs_default_limit = 400
+# Maximum number of previous jobs to include in jobs request (to prevent performance issues)
+#previous_jobs_max_limit = 4000
 
 [archiving]
 # Moves logs of jobs which are preserved during the cleanup because they are considered important
--- a/lib/OpenQA/Setup.pm
+++ b/lib/OpenQA/Setup.pm
@@ -156,7 +156,11 @@ sub read_config ($app) {
             all_tests_default_finished_jobs => 500,
             all_tests_max_finished_jobs => 5000,
             list_templates_default_limit => 500,
-            list_templates_max_limit => 5000
+            list_templates_max_limit => 5000,
+            next_jobs_default_limit => 100,
+            next_jobs_max_limit => 1000,
+            previous_jobs_default_limit => 400,
+            previous_jobs_max_limit => 4000
         },
         archiving => {
             archive_preserved_important_jobs => 0,
--- a/lib/OpenQA/WebAPI/Controller/Test.pm
+++ b/lib/OpenQA/WebAPI/Controller/Test.pm
@@ -438,8 +438,11 @@ sub _show {
 sub job_next_previous_ajax ($self) {
     return $self->reply->not_found unless my $main_job = $self->_get_current_job;
     my $main_jobid = $main_job->id;
-    my $p_limit = $self->param('previous_limit') // 400;
-    my $n_limit = $self->param('next_limit') // 100;
+
+    my $limits = OpenQA::App->singleton->config->{misc_limits};
+    my $p_limit = min($limits->{previous_jobs_max_limit},
+        $self->param('previous_limit') // $limits->{previous_jobs_default_limit});
+    my $n_limit = min($limits->{next_jobs_max_limit}, $self->param('next_limit') // $limits->{next_jobs_default_limit});
 
     my $schema = $self->schema;
     my $jobs_rs = $schema->resultset('Jobs')->next_previous_jobs_query(
--- a/t/ui/16-tests_job_next_previous.t
+++ b/t/ui/16-tests_job_next_previous.t
@@ -276,5 +276,39 @@ subtest 'bug reference shown' => sub {
     );
 };
 
+subtest 'server-side limit has precedence over user-specified limit' => sub {
+    my $limits = OpenQA::App->singleton->config->{misc_limits};
+    $limits->{next_jobs_max_limit} = 5;
+    $limits->{previous_jobs_max_limit} = 5;
+    $limits->{next_jobs_default_limit} = 2;
+    $limits->{previous_jobs_default_limit} = 2;
+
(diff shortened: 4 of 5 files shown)


lib/OpenQA/WebAPI/Controller/Test.pm
```diff
@@ -438,8 +438,11 @@ sub _show {
 sub job_next_previous_ajax ($self) {
     return $self->reply->not_found unless my $main_job = $self->_get_current_job;
     my $main_jobid = $main_job->id;
-    my $p_limit = $self->param('previous_limit') // 400;
-    my $n_limit = $self->param('next_limit') // 100;
```
okurz, line 442: I suggest to keep this original values as default so instead of 250/250 use 400/100 in the config file, Setup.pm, etc.

okurz (changes requested):

t/ui/16-tests_job_next_previous.t
```diff
@@ -276,5 +276,38 @@ subtest 'bug reference shown' => sub {
     );
 };
 
+subtest 'server-side limit has precedence over user-specified limit' => sub {
+    my $limits = OpenQA::App->singleton->config->{misc_limits};
+    $limits->{next_ajax_max_limit} = 5;    # set maximum limit for next
+    $limits->{previous_ajax_max_limit} = 5;    # set maximum limit for previous
+    $limits->{next_ajax_default_limit} = 2;    # set default limit for next
+    $limits->{previous_ajax_default_limit} = 2;    # set default limit for previous
```
okurz, line 284: IMHO the comments are redundant and should be removed. If the variable names are not clear enough then the variables should be renamed rather than explained by comments

okurz (changes requested):

t/ui/16-tests_job_next_previous.t
```diff
@@ -276,5 +276,38 @@ subtest 'bug reference shown' => sub {
...
+
+    $t->get_ok('/tests/99910/ajax?previous_limit=0&next_limit=6', 'query with exceeding user-specified limit for next')
+      ->status_is(200);
+    my $jobs = $t->tx->res->json->{data};
+    is ref $jobs, 'ARRAY', 'data returned (1)' and is scalar @$jobs, 7, 'maximum limit for next is effective';
```
kraih, line 289: Tests are much easier to maintain long term when you put each `is` statement on a separate line.

okurz: Here I think it makes sense to combine lines because only if the result is a proper array the second check makes sense, otherwise the error message could be confusing.

lib/OpenQA/WebAPI/Controller/Test.pm
Martchus: Yes, unless you did some performance testing the would show 250/250 is in fact a better choice I'd also keep the current defaults.

t/ui/16-tests_job_next_previous.t
Martchus, line 289: Why are 7 jobs expected here? The limit should max it out at 5 "next" jobs. Then there's supposedly still the current job returned. So 6 jobs would make sense. Is the 7th job a previous job one always gets (despite the previous limit being 0)? It would likely be good to have an explaining comment. I think the same counts for the other checks.

etc/openqa/openqa.ini
```diff
@@ -212,6 +212,14 @@ concurrent = 0
 #list_templates_default_limit = 500
 # Maximum number of templates to include in job_templates api response (to prevent performance issues)
 #list_templates_max_limit = 5000
+# Default number of next jobs to include in ajax request (to prevent performance issues)
+#next_ajax_default_limit = 250
```
Martchus, line 216: Maybe there'd be a better name for this and the other settings. We normally shouldn't care about the "AJAX" part here in this user-facing config file as it is an implementation detail. It would also make sense to have "job" somewhere in these setting keys. So what about:

```suggestion
#next_jobs_default_limit = 250
```

(This suggestion doesn't contain changes for other review comments.)

Martchus (changes requested):

t/ui/16-tests_job_next_previous.t
kraih: A fair point. I would just leave out the first test though and have it die on the second one.

kraih: A good question that leads to this [messy query](https://github.com/os-autoinst/openQA/blob/ec2d97adf17caf5a4aabb4f51a995abca885a86b/lib/OpenQA/Schema/Result/JobNextPrevious.pm#L20), which i've never seen before. Wonder how that's not a method yet.

r-richardson: Yes, by default the latest and the current job are always returned, plus the 5 (if the limit is effective) equals 7

Martchus: Ah, the latest. That makes sense as well. It would still be good to put that information in a comment.

lib/OpenQA/Setup.pm
```diff
@@ -156,7 +156,11 @@ sub read_config ($app) {
             all_tests_default_finished_jobs => 500,
             all_tests_max_finished_jobs => 5000,
             list_templates_default_limit => 500,
-            list_templates_max_limit => 5000
+            list_templates_max_limit => 5000,
+            next_jobs_default_limit => 250,
+            next_jobs_max_limit => 2500,
+            previous_jobs_default_limit => 250,
+            previous_jobs_max_limit => 2500
```
okurz, line 163: these need to be changed as well

lib/OpenQA/Setup.pm
```diff
@@ -156,7 +156,11 @@ sub read_config ($app) {
             all_tests_default_finished_jobs => 500,
             all_tests_max_finished_jobs => 5000,
             list_templates_default_limit => 500,
-            list_templates_max_limit => 5000
+            list_templates_max_limit => 5000,
+            next_jobs_default_limit => 250,
+            next_jobs_max_limit => 2500,
+            previous_jobs_default_limit => 250,
```
kalikiana, line 162: This should match openqa.ini I think. Please make sure they're in sync

kalikiana (approved):

t/ui/16-tests_job_next_previous.t
r-richardson: ok, i removed the comments and changed the naming of the variables 
(replaced 'ajax' with 'jobs' e.g. next_ajax_max_limit -> next_jobs_max_limit, as Marius had suggested)

t/ui/16-tests_job_next_previous.t
r-richardson: i've added `# expected amount is the defined limit plus current and latest job`

Martchus (approved):

kraih (approved):

okurz (approved):

PR #4876, opened 2022-10-28

okurz:


kalikiana (approved):

lib/OpenQA/WebAPI/Controller/Test.pm
```diff
@@ -142,28 +141,25 @@ sub list_running_ajax {
...
-                arch => $job->ARCH // '',
-                build => $job->BUILD // '',
-                testtime => ($job->t_started // '') . 'Z',
-                result => $job->result,
-                group => $job->group_id,
-                state => $job->state,
-                progress => $job->progress_info,
-            });
-    }
+    my @running = map {
```
perlpunk, line 144: Hm, this feels like a dejavu. I don't see the benefit of the change. Haven't you proposed this change before? Maybe it was similar code elsewhere. 
Have you benchmarked this and looked at the memory consumption? The new code will load all database objects into memory.

okurz: Yes, could very well be that I have proposed this change in before. Sorry about that. I have excluded it for now from the commit. Still, I think we should be able to process all objects in memory. I consider this cleaner with such functional rather than procedural code but I have not benchmarked it.

Martchus (approved):
Looks good although it really just changes Perl coding style (and doesn't simplify and database operation, not even changes DBIx usage).

Martchus:
No, I just found the commit message a bit odd.

PR #4880, opened 2022-10-31
Add basic terraform recipe and localstack test

baierjan:
Reference: https://progress.opensuse.org/issues/118660

### DIFF
--- a/container/terraform/variables.tf
+++ b/container/terraform/variables.tf
@@ -38,7 +38,7 @@ variable "type" {
 variable "image_id" {
   type        = string
   description = "AMI (id of machine image) to use"
-  default     = "ami-08199c714a509d3bc"
+  default     = "ami-0083277bb76fc42c3"
 
   validation {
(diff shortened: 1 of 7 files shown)


Martchus (commented):
I know it is just a draft but adding a bit of documentation wouldn't hurt. Nice that you have already validation tests.

baierjan:
> I know it is just a draft but adding a bit of documentation wouldn't hurt. Nice that you have already validation tests.

Yeah, thanks for the note, I also thought about documentation but so far it is just a rewrite of [instructions from the wiki](https://progress.opensuse.org/projects/openqav3/wiki/Wiki#section-71). At the end, I will at least add an explanation how to call the whole thing. But for that, I need to test that first, so more will come.

container/terraform/variables.tf
```diff
@@ -0,0 +1,60 @@
+variable "region" {
+  # preffered region: Frankfurt, Germany
```
perlpunk, line 2:
```suggestion
  # preferred region: Frankfurt, Germany
```

PR #4884, opened 2022-11-03
Fix log upload from openQA jobs in case of read-only cache directory

okurz:
We observed a problem that the os-autoinst command server received log
upload attempts but failed with

```
Mojo::Reactor::Poll: I/O watcher failed: Error in tempfile() using template /var/lib/openqa/cache/tmp/mojo.tmp.XXXXXXXXXXXXXXXX: Could not create temp file /var/lib/openqa/cache/tmp/mojo.tmp.mbmTdPv6NyxTOBTy: Permission denied at /usr/lib/perl5/5.26.1/File/Temp.pm line 1102.
```

Setting the environment variable MOJO_TMPDIR from the worker will mean
that also the os-autoinst command server effectively started by
the openQA worker will inherit

The original change was done at a time when the openQA worker itself
downloaded assets before the introduction of the separate cache service
which already handles that part correctly in OpenQA::Downloader so we
can simply remove the corresponding code without replacement.

Related progress issue: https://progress.opensuse.org/issues/119713

### DIFF
--- a/lib/OpenQA/WebAPI/Controller/Test.pm
+++ b/lib/OpenQA/WebAPI/Controller/Test.pm
@@ -23,9 +23,8 @@ use constant DEPENDENCY_DEBUG_INFO => $ENV{OPENQA_DEPENDENCY_DEBUG_INFO};
 sub referer_check ($self) {
     return $self->reply->not_found if (!defined $self->param('testid'));
     my $referer = $self->req->headers->header('Referer') // '';
-    if ($referer) {
-        $self->schema->resultset('Jobs')->mark_job_linked($self->param('testid'), $referer);
-    }
+    return 1 unless $referer;
+    $self->schema->resultset('Jobs')->mark_job_linked($self->param('testid'), $referer);
     return 1;
 }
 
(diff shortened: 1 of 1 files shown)


lib/OpenQA/Worker.pm
```diff
@@ -579,9 +545,7 @@ sub _inform_webuis_before_stopping {
 }
 
 # stops the current job and (if there is one) and terminates the worker
-sub stop {
-    my ($self, $reason) = @_;
-
+sub stop ($self, $reason) {
```
Martchus, line 548: It should be ok to not pass a `$reason`.

lib/OpenQA/Worker.pm
```diff
@@ -599,31 +563,23 @@ sub stop {
 }
 
 # stops the current job if there's one and it is running
-sub stop_current_job {
-    my ($self, $reason) = @_;
-
+sub stop_current_job ($self, $reason) {
```
Martchus, line 566: It should be ok to not pass a `$reason`.

Martchus (changes requested):

lib/OpenQA/Worker.pm
okurz: ok. I guess I will need to exclude further going signatures changes from this "fix" PR

okurz:
Sorry, I need to open another PR https://github.com/os-autoinst/openQA/pull/4885

PR #4886, opened 2022-11-04
Clean up git repository before rebase

sdclarke:
If there have been previous failures when adding new needles through the needle editor it is possible for the rebase to fail due to uncommitted changes. This commit will ensure that any uncommitted changes are cleaned up before attempting to rebase

### DIFF
--- a/lib/OpenQA/Git.pm
+++ b/lib/OpenQA/Git.pm
@@ -54,6 +54,10 @@ sub set_to_latest_master ($self, $args = undef) {
     }
 
     if (my $update_branch = $self->config->{update_branch}) {
+        if ($self->config->{do_cleanup} eq 'yes') {
+            my $res = run_cmd_with_log_return_error([@git, 'reset', '--hard', 'HEAD']);
+            return _format_git_error($res, 'Unable to reset repository to HEAD') unless $res->{status};
+        }
         my $res = run_cmd_with_log_return_error([@git, 'rebase', $update_branch]);
         return _format_git_error($res, 'Unable to reset repository to origin/master') unless $res->{status};
     }
--- a/t/16-utils-runcmd.t
+++ b/t/16-utils-runcmd.t
@@ -95,13 +95,17 @@ subtest 'git commands with mocked run_cmd_with_log_return_error' => sub {
     # configure update branch and remote
     $git->config->{update_remote} = 'origin';
     $git->config->{update_branch} = 'origin/master';
+    $git->config->{do_cleanup} = 'yes';
     is($git_config->{update_remote}, $git->config->{update_remote}, 'global git config reflects all changes');
 
     # test set_to_latest_master (non-error case)
     is($git->set_to_latest_master, undef, 'no error');
     is_deeply(
         \@executed_commands,
-        [[qw(git -C foo/bar remote update origin)], [qw(git -C foo/bar rebase origin/master)],],
+        [
+            [qw(git -C foo/bar remote update origin)], [qw(git -C foo/bar reset --hard HEAD)],
+            [qw(git -C foo/bar rebase origin/master)],
+        ],
         'git remote update and rebase executed',
     ) or diag explain \@executed_commands;
 
@@ -180,6 +184,7 @@ subtest 'saving needle via Git' => sub {
         \@executed_commands,
         [
             [qw(git -C), $empty_tmp_dir, qw(remote update origin)],
+            [qw(git -C), $empty_tmp_dir, qw(reset --hard HEAD)],
             [qw(git -C), $empty_tmp_dir, qw(rebase origin/master)],
             [qw(git -C), $empty_tmp_dir, qw(add), "foo.json", "foo.png"],
             [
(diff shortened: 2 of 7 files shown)


lib/OpenQA/Git.pm
```diff
@@ -54,7 +54,13 @@ sub set_to_latest_master ($self, $args = undef) {
     }
 
     if (my $update_branch = $self->config->{update_branch}) {
-        my $res = run_cmd_with_log_return_error([@git, 'rebase', $update_branch]);
+        my $res = run_cmd_with_log_return_error([@git, 'restore', '--staged', '.']);
+        return _format_git_error($res, 'Unable to restore staged files') unless $res->{status};
+        $res = run_cmd_with_log_return_error([@git, 'restore', '.']);
+        return _format_git_error($res, 'Unable to restore files') unless $res->{status};
```
Martchus, line 60: I suppose `git reset --hard HEAD` would be simpler than these two restore steps. I suppose then you could even avoid the cleanup.

Martchus (commented):
I have proposed that some years ago and it was rejected. I personally still find it acceptable despite the potentially lost data. If that is still a concern one could also commit all uncommitted changes on a "backup" branch and then switch back to master.

sdclarke:
> I have proposed that some years ago and it was rejected. I personally still find it acceptable despite the potentially lost data. If that is still a concern one could also commit all uncommitted changes on a "backup" branch and then switch back to master.

Maybe we could have it behind an extra config option? Personally I've found myself having to manually clean up the repository too many times but I can understand that for some people this change could cause problems, so maybe requiring it to be explicitly enabled would be a good thing.

lib/OpenQA/Git.pm
sdclarke: Yeah that seems like an easier solution, I'll make that change

Martchus:
> Maybe we could have it behind an extra config option?

That would make sense.

lib/OpenQA/Git.pm
```diff
@@ -54,13 +54,11 @@ sub set_to_latest_master ($self, $args = undef) {
...
 
     if (my $update_branch = $self->config->{update_branch}) {
-        my $res = run_cmd_with_log_return_error([@git, 'restore', '--staged', '.']);
-        return _format_git_error($res, 'Unable to restore staged files') unless $res->{status};
-        $res = run_cmd_with_log_return_error([@git, 'restore', '.']);
-        return _format_git_error($res, 'Unable to restore files') unless $res->{status};
-        $res = run_cmd_with_log_return_error([@git, 'clean', '-f']);
-        return _format_git_error($res, 'Unable to clean untracked files') unless $res->{status};
-        $res = run_cmd_with_log_return_error([@git, 'rebase', $update_branch]);
+        if (($self->config->{do_cleanup} || '') eq 'yes') {
```
Martchus, line 57: The default specified in `Setup.pm` should be effective so there's no need for `|| ''`:
```suggestion
        if (($self->config->{do_cleanup}) eq 'yes') {
```

Martchus, line 57: The default specified in `Setup.pm` should be effective so there's no need for `|| ''`:
```suggestion
        if (($self->config->{do_cleanup}) eq 'yes') {
```

Martchus (commented):
It would be great to have test coverage. Maybe you can simply enable the config in one of the existing tests and possibly also add a check to verify whether the new Git command is actually executed. Otherwise it looks good to merge.

t/16-utils-runcmd.t
```diff
@@ -95,13 +95,14 @@ subtest 'git commands with mocked run_cmd_with_log_return_error' => sub {
...
     $git->config->{update_branch} = 'origin/master';
+    $git->config->{do_cleanup} = 'yes';
     is($git_config->{update_remote}, $git->config->{update_remote}, 'global git config reflects all changes');
 
     # test set_to_latest_master (non-error case)
     is($git->set_to_latest_master, undef, 'no error');
     is_deeply(
         \@executed_commands,
-        [[qw(git -C foo/bar remote update origin)], [qw(git -C foo/bar rebase origin/master)],],
+        [[qw(git -C foo/bar remote update origin)], [qw(git -C foo/bar rebase origin/master)], [qw(git -C foo/bar reset --hard HEAD)],],
```
Martchus, line 105: Shouldn't the reset happen before the rebase?

sdclarke: Ah yes, I didn't read it carefully enough

Martchus:
Looks like you need to run tidy (`tools/tidy`).

Martchus (commented):
Looks good, consider it approved. (I'd only prefer to squash the commits when merging due to the last commit which means it needs to be merged manually and not automatically by approval on GitHub.)

okurz (commented):
I agree with @Martchus. Please squash commits so that every commit is self-contained and passes or would pass CI checks individually, then we can approve

Martchus (approved):

okurz (approved):

PR #4888, opened 2022-11-04
Add an API endpoint to find jobs with certain job settings

kraih:
This PR implements the solution we discussed, which limits the query to the 20000 most recent job ids in the `job_settings` table.

Progress: https://progress.opensuse.org/issues/117655

### DIFF
--- a/lib/OpenQA/Schema/ResultSet/JobSettings.pm
+++ b/lib/OpenQA/Schema/ResultSet/JobSettings.pm
@@ -2,9 +2,41 @@
 # SPDX-License-Identifier: GPL-2.0-or-later
 
 package OpenQA::Schema::ResultSet::JobSettings;
+use Mojo::Base 'DBIx::Class::ResultSet', -signatures;
 
+use OpenQA::App;
 
-use Mojo::Base 'DBIx::Class::ResultSet';
+sub jobs_for_setting ($self, $options) {
+    my $limit = OpenQA::App->singleton->config->{misc_limits}{job_settings_max_recent_jobs};
+
+    my $key_like = $options->{key};
+    $key_like =~ s/\*/\%/g;
+    my $list_value = $options->{list_value};
+    my $list_value_like = "%${list_value}%";
+
+    # Get the highest job id to limit the number of jobs that need to be considered (to improve performance)
+    my $dbh = $self->result_source->schema->storage->dbh;
+    my $sth = $dbh->prepare('SELECT max(job_id) FROM job_settings');
+    $sth->execute;
+
+    my $result = $sth->fetchrow_arrayref;
+    my $max_id = defined $result ? $result->[0] : 0;
+    my $min_id = $max_id > $limit ? $max_id - $limit : 0;
+
+    # Instead of limiting the number of jobs, this query could also use a trigram gin index to achieve even better
+    # performance (at the cost of disk space and setup complexity)
+    $sth = $dbh->prepare(
+        'SELECT job_id, value FROM job_settings WHERE key LIKE ? AND value LIKE ? AND job_id > ? ORDER BY job_id DESC');
+    $sth->execute($key_like, $list_value_like, $min_id);
+
+    # Match list values
+    my @jobs;
+    for my $row (@{$sth->fetchall_arrayref}) {
+        my ($job_id, $value) = @$row;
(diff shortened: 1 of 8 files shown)


Martchus (commented):
I was more thinking about a generic route to query jobs by their settings with keys/values like a certain pattern. It would have the advantage of being useful for other cases and we'd avoid adding qem tooling specific code. Otherwise it looks good to merge.

kraih:
> I was more thinking about a generic route to query jobs by their settings with keys/values like a certain pattern. It would have the advantage of being useful for other cases and we'd avoid adding qem tooling specific code. Otherwise it looks good to merge.

Actually, the value matching would be rather tricky, since the syntax is user defined. In the case of incidents we have a comma separated list of numbers, where we want to match complete numbers (never partials). An API that just supports wildcards at the beginning and end of the search term could not replicate that. We would probably need a more complex query language or regexes.

okurz (dismissed):
Similar as @Martchus mentioned: Our design of openQA does have no meaning of an "incident". This should be generalized.

kraih:
> Similar as @Martchus mentioned: Our design of openQA does have no meaning of an "incident". This should be generalized.

Then how do you want comma separated values to be dealt with in a generic way from the API side?

kraih:
If there are no better suggestions i'll probably go for something that can be extended later, like `/api/v1/job-settings?key=*_TEST_ISSUES&list-value=123456`.

kraih:
> Our design of openQA does have no meaning of an "incident".

Technically this is no longer true, since [this](https://github.com/os-autoinst/openQA/blob/5afc5ed9f1dd59fe481bd9c937da498c793ff2d3/lib/OpenQA/Script/CloneJobSUSE.pm#L11) has been merged, which explicitly references maintenance incidents.

okurz:
> If there are no better suggestions i'll probably go for something that can be extended later, like `/api/v1/jobs-for-settings?key=*_TEST_ISSUES&list-value=123456`.

you mean like globbing? Sounds good to me. I don't know if there are good standards around for how to pass "templated search expressions" or full regex in query URLs. Let's start with something. We have to be brave at some point even if later on we find something better. Then we might just have to choose different formats like "key_re" for regex matches or so.

> > Our design of openQA does have no meaning of an "incident".
> 
> Technically that is no longer true, since [this](https://github.com/os-autoinst/openQA/blob/5afc5ed9f1dd59fe481bd9c937da498c793ff2d3/lib/OpenQA/Script/CloneJobSUSE.pm#L11) has been merged, which explicitly references maintenance incidents.

Yes, you are right. That was a compromise to motivate team-external contributors ;)

kraih:
Ok, i went with two parameters for now, `key` containing a glob like `?key=*_TEST_ISSUES`, and `list_value` containing a substring like `12345` that is part of a comma delimited list of values.

Martchus (approved):

baierjan (approved):

lib/OpenQA/Schema/ResultSet/JobSettings.pm
```diff
@@ -2,9 +2,41 @@
...
+        'SELECT job_id, value FROM job_settings WHERE key LIKE ? AND value LIKE ? AND job_id > ? ORDER BY job_id DESC');
+    $sth->execute($key_like, $list_value_like, $min_id);
+
+    # Match list values
+    my @jobs;
+    for my $row (@{$sth->fetchall_arrayref}) {
+        my ($job_id, $value) = @$row;
+        push @jobs, $job_id if $value =~ /(?:^|,)$list_value(?:$|,)/;
+    }
+    return \@jobs;
```
okurz, line 38: would it be possible to use a grep or map here?

kraih: Actually had that before. The loop is a little more readable though.
```perl
return [map { $_->[1] =~ /(?:^|,)$list_value(?:$|,)/ ? $_->[0] : () } @{$sth->fetchall_arrayref}];
```

kraih: Maybe worth mentioning that the loop version is also a bit faster. The `map` only wins here if you go for a lower line count.

Martchus: Too bad that the `map` has worse performance but I also agree that the loop is more readable here anyways.

PR #4896, opened 2022-11-11
Add distinct set of limits for assets

Martchus:
The generic default limit (that was taken from the "All tests" page for jobs) doesn't make any sense here so this change allows us to configure the limit for asset queries separately.

Related ticket: https://progress.opensuse.org/issues/120315


perlpunk (approved):

Martchus:
I've been increasing the defaults as this can apparently handle a large set of results in reasonable time, tested with OSD where we currently have the biggest number of assets. (The query for assets is rather simple compared to e.g. jobs.)

okurz (approved):

PR #4898, opened 2022-11-15
PART 2: WebAPI::Controller::Running: Remove uncovered close_connection handler - After #5755

okurz:
After:
* https://github.com/os-autoinst/openQA/pull/4961
* https://github.com/os-autoinst/openQA/pull/5755

### DIFF
--- a/lib/OpenQA/Shared/Controller/Running.pm
+++ b/lib/OpenQA/Shared/Controller/Running.pm
@@ -111,12 +111,6 @@ sub streamtext ($self, $file_name, $start_hook = undef, $close_hook = undef) {
     # Check for new lines from the logfile using recurring timer
     # Setup utility function to close the connection if something goes wrong
     my $timer_id;
-    my $close_connection = sub ($self) {
-        Mojo::IOLoop->remove($timer_id);
-        $close_hook->();
-        $self->finish;
-        close $log;
-    };
     $timer_id = Mojo::IOLoop->recurring(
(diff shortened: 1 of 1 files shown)


lib/OpenQA/WebAPI/Controller/Running.pm
```diff
@@ -127,10 +121,9 @@ sub streamtext ($self, $file_name, $start_hook = undef, $close_hook = undef) {
                 $size = -s $logfile;
             }
 
-            # Zero tolerance for any shenanigans with the logfile, such as
-            # truncation, rotation, etc.
             my @st = stat $logfile;
-            return $close_connection->() unless @st && $st[1] == $ino && $st[3] > 0 && $st[7] >= $size;
+            die "Zero tolerance for any shenanigans with the logfile, such as truncation, rotation, etc."
```
Martchus, line 125: We should likely close the connection here as the code previously did. Otherwise we leave the client waiting for further data until a timeout occurs. Placing a `die` here seems generally wrong as it would likely just be caught by Mojolicious' handler. The message has also not enough context to be useful in the server-side logs (where it would end up). If we wanted better error handling, then an error should likely go to the client in form of a bad return code.

Martchus (changes requested):

lib/OpenQA/WebAPI/Controller/Running.pm
okurz: ok, I agree. All CI tests passing show that we are missing test coverage here

PR #4900, opened 2022-11-15
Auto-detect `WORKER_HOSTNAME` on worker startup if not set explicitly

Martchus:
* Make it unnecessary to specify `WORKER_HOSTNAME` in `workers.ini` if `hostname --fqdn` returns the expected name.
* Tested the used functions of the `Net::Domain` module on multiple production worker hosts (using Leap 15.3 and 15.4). They behave as expected (so workers will still register using the short hostname and `WORKER_HOSTNAME` will be set to `hostname --fqdn`).
* Required on worker level for the developer mode (independently of how the variable is handled by os-autoinst)
* See https://progress.opensuse.org/issues/120261#note-8

### DIFF
--- a/lib/OpenQA/Worker.pm
+++ b/lib/OpenQA/Worker.pm
@@ -31,12 +31,12 @@ BEGIN {
     };
 }
 
-use POSIX 'uname';
 use Fcntl;
 use File::Path qw(make_path remove_tree);
 use File::Spec::Functions 'catdir';
 use Mojo::IOLoop;
 use Mojo::File 'path';
+use Net::Domain qw(hostname hostfqdn);
 use Try::Tiny;
 use Scalar::Util 'looks_like_number';
 use OpenQA::Constants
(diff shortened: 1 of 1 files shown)


okurz (approved):
How about you run this for a day on w11/w12 connected to OSD for testing?

Martchus:
I can do that. Of course it will only be worthwhile when also removing `WORKER_HOSTNAME` from `workers.ini` so I'd have to do that manually (temporarily removing the workers from salt).

Martchus:
I have removed `WORKER_HOSTNAME` on both hosts and stopped salt. Then I have deployed the change on both workers:

```
zypper ar --priority 85 https://download.opensuse.org/repositories/devel:/openQA:/GitHub:/os-autoinst:/openQA:/PR-4900/openSUSE_Leap_15.4 pr-4900
zypper in pr-4900:openQA-worker
```

Judging by the worker logs it already looks good.

perlpunk (approved):

perlpunk:
Could it be that this causes this problem? https://progress.opensuse.org/issues/120579
https://openqa.opensuse.org/tests/2879037#step/openqa_worker/9

PR #4901, opened 2022-11-15
Work around openQA-single-instance pulling old openQA on Leap 15.4

a-kpappas:
Using the openQA-bootstrap script on Leap 15.4, without ssh keys in use, causes it to fail, because it pulls an older version of fetchneedles and openQA where poo#104794 has not been fixed.

However by explicitly adding openQA to the list of the packages to be installed, the problems seems to be fixed.

Related ticket: https://progress.opensuse.org/issues/115784

### DIFF
--- a/script/openqa-bootstrap
+++ b/script/openqa-bootstrap
@@ -26,7 +26,7 @@ fi
 
 
 # install packages
-zypper -n install --no-recommends openQA-single-instance qemu-arm qemu-ppc qemu-x86 qemu-tools sudo iputils os-autoinst-distri-opensuse-deps
+zypper -n install --no-recommends openQA openQA-single-instance qemu-arm qemu-ppc qemu-x86 qemu-tools sudo iputils os-autoinst-distri-opensuse-deps
 
 if [ "$(uname -m)" = "aarch64" ]; then
     zypper -n install --no-recommends qemu-uefi-aarch64


perlpunk (approved):

okurz (changes requested):
Hm, so do I understand this correctly that the problem happens when openQA was installed already before starting openqa-bootstrap but from different repositories before openqa-bootstrap adds devel:openQA, right?

I guess if we add extra repositories we should better make sure we call `zypper dup` to update any already installed packages which come from devel:openQA. That can be more than just openQA.

@a-kpappas could you check if instead of explicitly selecting the "openQA" package for install – which effectively upgrades it – call `zypper dup --allow-vendor-change` after line 24 and check if that also helps in your cases.

a-kpappas:
> Hm, so do I understand this correctly that the problem happens when openQA was installed already before starting openqa-bootstrap but from different repositories before openqa-bootstrap adds devel:openQA, right?
> 

openQA is not installed, and the problematic script `fetchneedles` does not exist on the system either. These are installed by openQA-bootstrap. 

```
test-openqa:~ # zypper search -i openQA
...
S  | Name             | Summary                                              | Type
---+------------------+------------------------------------------------------+--------
i+ | openQA-bootstrap | Automated openQA setup                               | package
i  | openQA-client    | Client tools for remote openQA management            | package
i  | openQA-common    | The openQA common tools for web-frontend and workers | package
i+ | openQA-worker    | The openQA worker                                    | package
test-openqa:~ # zypper info openQA-common
```
> I guess if we add extra repositories we should better make sure we call `zypper dup` to update any already installed packages which come from devel:openQA. That can be more than just openQA.
> 
> @a-kpappas could you check if instead of explicitly selecting the "openQA" package for install – which effectively upgrades it – call `zypper dup --allow-vendor-change` after line 24 and check if that also helps in your cases.

`zypper -n dup` does work but it is a much more invasive operation, and should not be done by a random script on a Leap system. I think that just adding the openQA to the initial `zypper in`,  makes more sense, since it is **already** going to be installed

okurz:
> `zypper -n dup` does work but it is a much more invasive operation, and should not be done by a random script on a Leap system. I think that just adding the openQA to the initial `zypper in`, makes more sense, since it is **already** going to be installed

Ok, then try

```
zypper -n dup --allow-vendor-change devel:openQA:Leap:${VERSION}
zypper -n dup --allow-vendor-change devel:openQA
```

a-kpappas:
> Ok, then try
> 
> ```
> zypper -n dup --allow-vendor-change devel:openQA:Leap:${VERSION}
> zypper -n dup --allow-vendor-change devel:openQA
> ```

Those are not valid zypper commands.

a-kpappas:
What are your objections to my approach though?

Martchus:
@okurz If you wanted to use the `--repo` flag in those zypper commands then I'd say it is likely not a good idea. We've seen problems caused by this when using it in the continuous update service.

Maybe openQA-single-instance should simply require a specific openQA version (like openQA requires a specific openQA-common version). So far openQA-single-instance only indirectly depends on openQA which should likely be fixed as well.

okurz:
> Maybe openQA-single-instance should simply require a specific openQA version (like openQA requires a specific openQA-common version). So far openQA-single-instance only indirectly depends on openQA which should likely be fixed as well.

Sounds good. I don't know what you mean exactly but likely best to show in actual code changes

a-kpappas:
@okurz Should I close this then? I don´t mind using my fork in my scripts, until the `openQA-single-instance` package is updated.

Martchus:
Just change this PR to update the package instead. The spec file is in the repo.

a-kpappas:
Updating openQA-single-instance does not work.

okurz:
> Updating openQA-single-instance does not work.

what do you mean in particular?

a-kpappas:
> Just change this PR to update the package instead. The spec file is in the repo.

I assume this comment asks me to `zypper -n install -f openQA-single-instance`. 
This does not work.

Martchus:
No. I suggested to update the spec file for building the openQA-single-instance package so that this package has a dependency to the openQA package with a fixed version (similar like it is already done for other openQA-* packages). The spec file is contained by this repo so you'll just have to amend your commit.

a-kpappas:
I am not familiar enough with spec files  to make a change in that big of a file. I will close this and keep my fork. Hopefully it won't be necessary by the time Leap 15.5 is released next year.

PR #4904, opened 2022-11-16
Fix "Use of uninitialized value" in OpenQA/Worker.pm

okurz:
Fixes regression in 79ab25d32

Related progress issue: https://progress.opensuse.org/issues/120579


kalikiana (approved):

perlpunk (approved):
That fixes the warning, but probably not the actual error in the test.

Martchus:
The change makes sense nevertheless. I'll check what the test is actually testing for (to see what's still missing).

PR #4917, opened 2022-11-24
Add signatures in OpenQA::Task::Bug::Limit and more

okurz:


Martchus:
But considering the CI results there must be a syntax error.

lib/OpenQA/Utils.pm
```diff
@@ -204,7 +201,7 @@ sub testcasedir {
 
 =head2 gitrepodir
 
-  gitrepodir(distri => DISTRI, version => VERSION)
+  gitrepodir({distri => DISTRI, version => VERSION})
```
perlpunk, line 204: Why?

okurz: I guess we will never know …

perlpunk (approved):
I assume you simply fixed the things we commented on and nothing else, I don't want to go through the whole diff of 10 commits again :(

PR #4922, opened 2022-11-24
Use signatures in OpenQA::Shared::Plugin::Gru

okurz:


Martchus (approved):

okurz:
codecov/patch failed, I would go ahead anyway

baierjan (approved):

lib/OpenQA/Schema.pm
```diff
@@ -100,19 +87,16 @@ sub deploy {
 }
 
 # Class attribute used for testing with OpenQA::Test::Database
-sub search_path_for_tests {
-    my $class = shift;
+sub search_path_for_tests ($class, $new_search_path = undef) {
     state $search_path;
-    $search_path = shift if @_;
+    $search_path = $new_search_path if defined $new_search_path;
```
perlpunk, line 90: There's no way to set `$search_path` to undef now.

okurz: thx, I have reverted in the meantime and will look into doing this properly again

okurz: What do you suggest we do here instead?

lib/OpenQA/Shared/Plugin/Gru.pm
```diff
@@ -47,13 +47,14 @@ sub register {
     my $schema = $app->schema;
 
     my $conn = Mojo::Pg->new;
-    if (ref $schema->storage->connect_info->[0] eq 'HASH') {
-        $self->dsn($schema->dsn);
-        $conn->username($schema->storage->connect_info->[0]->{user});
-        $conn->password($schema->storage->connect_info->[0]->{password});
+    my $connect_info = $schema->storage->connect_info->[0];
+    if (ref $connect_info eq 'HASH') {
+        $self->dsn($connect_info->dsn);
```
perlpunk, line 47:
```suggestion
$self->dsn($schema->dsn);
```

lib/OpenQA/Schema.pm
perlpunk:
```perl
sub search_path_for_tests ($class, @args) {
    state $search_path;
    $search_path = shift @args if @args;
```

lib/OpenQA/Shared/Plugin/Gru.pm
okurz: applied in #4928

lib/OpenQA/Schema.pm
okurz: applied in
* #4927

PR #4923, opened 2022-11-24
Relabel the shared volumes for hadolint

dcermak:
This is required to run hadolint on systems with SELinux enabled

### DIFF
--- a/tools/static_check_containers
+++ b/tools/static_check_containers
@@ -12,5 +12,5 @@ cre="${cre:-"podman"}"
 dockerfiles=(container/*/Dockerfile*)
 echo "# Static check of Dockerfiles"
 set -x
-$cre run -e 'HOME=/' --rm -v"$PWD":/repo -w /repo "$img" hadolint "${dockerfiles[@]}"
+$cre run -e 'HOME=/' --rm -v"$PWD":/repo:Z -w /repo "$img" hadolint "${dockerfiles[@]}"
 echo "All files ok"


tools/static_check_containers
```diff
@@ -12,5 +12,5 @@ cre="${cre:-"podman"}"
 dockerfiles=(container/*/Dockerfile*)
 echo "# Static check of Dockerfiles"
 set -x
-$cre run -e 'HOME=/' --rm -v"$PWD":/repo -w /repo "$img" hadolint "${dockerfiles[@]}"
+$cre run -e 'HOME=/' --rm -v"$PWD":/repo:Z -w /repo "$img" hadolint "${dockerfiles[@]}"
```
kalikiana, line 15: Maybe this should have a comment and a check to abort if $PWD is /home? or /usr? [The docs](https://docs.docker.com/storage/bind-mounts/#configure-the-selinux-label) say it renders the host unusable when one of those folders is used.

dcermak: Iirc podman refuses to relabel `/home`

okurz (changes requested):
I suggest you rephrase the git commit message to point out the important effect in the subject line and add more details afterwards. The git subject line is included in the changelog, the details are not. How about:

```
tools: Fix hadolint for systems with SELinux

We need to relabel the shared volumes for hadolint because SELinux ...
```

okurz (approved):

Martchus (approved):

PR #4928, opened 2022-11-25
Remove unused 'dsn' function and its usage

okurz:
Tests did not cover the 'dsn' function anyway.

### DIFF
--- a/lib/OpenQA/Shared/Plugin/Gru.pm
+++ b/lib/OpenQA/Shared/Plugin/Gru.pm
@@ -45,14 +45,8 @@ sub register ($self, $app, $config) {
     my $schema = $app->schema;
 
     my $conn = Mojo::Pg->new;
-    if (ref $schema->storage->connect_info->[0] eq 'HASH') {
-        $self->dsn($schema->dsn);
-        $conn->username($schema->storage->connect_info->[0]->{user});
-        $conn->password($schema->storage->connect_info->[0]->{password});
-    }
-    else {
-        $self->dsn($schema->storage->connect_info->[0]);
-    }
+    my $connect_info = $schema->storage->connect_info->[0];
+    $self->dsn($connect_info);
     $conn->dsn($self->dsn());
 
     # set the search path in accordance with the test setup done in OpenQA::Test::Database
--- a/lib/OpenQA/Schema.pm
+++ b/lib/OpenQA/Schema.pm
@@ -51,11 +51,6 @@ sub disconnect_db () {
     $SINGLETON = undef;
 }
 
-sub dsn {
-    my $self = shift;
-    return $self->storage->connect_info->[0]->{dsn};
-}
-
 sub deploy ($self, $force_overwrite = 0) {
     # lock config file to ensure only one thing will deploy/upgrade DB at once
     # we use a file in prjdir/db as the lock file as the install process and


okurz:
failed with

```
[14:39:07] t/20-stale-job-detection.t ................................ 
    #   Failed test 'minion job to finalize incomplete jobs enqueued'
    #   at t/20-stale-job-detection.t line 65.
    #          got: '96'
    #     expected: '2'
    # Looks like you failed 1 test of 17.
[14:39:07] t/20-stale-job-detection.t ................................ 1/? 
#   Failed test 'worker with job and not updated in last 120s is considered dead'
#   at t/20-stale-job-detection.t line 67.
                                                                            [14:39:07] t/20-stale-job-detection.t ................................ Dubious, test returned 1 (wstat 256, 0x100)
Failed 1/3 subtests 
```

kalikiana (approved):

okurz:
Removed superfluous `'` in the git commit message subject line

lib/OpenQA/Shared/Plugin/Gru.pm
```diff
@@ -45,13 +45,14 @@ sub register ($self, $app, $config) {
     my $schema = $app->schema;
 
     my $conn = Mojo::Pg->new;
-    if (ref $schema->storage->connect_info->[0] eq 'HASH') {
+    my $connect_info = $schema->storage->connect_info->[0];
+    if (ref $connect_info eq 'HASH') {
         $self->dsn($schema->dsn);
```
Martchus, line 50: This change removes `$schema->dsn`, so how is this still supposed to work?

Martchus (changes requested):
This change looks wrong. At least the commit message is misleading (as I only see a function being **re**moved but not moved/inlined).

lib/OpenQA/Shared/Plugin/Gru.pm
okurz: Maybe I lost that change by now by rebasing over the course of two years :) But maybe you have a good idea what we can do to introduce proper test coverage here now?

Martchus: We should probably just invoke the function with `$schema->storage->connect_info` being a hash (as I think this is the missing branch). We might need to mock `Mojo::Pg` and possibly other packages.

okurz: Wanting to try out to call openqa-gru manually I got inspired to draft
* #5796

Martchus: I really meant to invoke this single function.

okurz: I suspect that we can simply delete the if-branch

Martchus (approved):
I guess we don't need this way of specifying credentials anymore then.

baierjan:
There is an issue in production due to this:
```
Aug 01 13:56:01 ariel openqa-webui-daemon[10603]: Can't connect to data source 'HASH(0x5585ac6ef2e8)' because I can't work out what driver to use (it doesn't seem to contain a 'dbi:driver:' prefix and the DBI_DRIVER env var is not set) at /usr/lib/perl5/vendor_perl/5.26.1/Mojo/Pg.pm line 73.
```

okurz:
The good thing is that we saw this in docker-compose tests. I wonder if I can change the docker-compose tests to actually test the branch related to a pull request as we build packages in there - if that's not taking too long.

Martchus:
That would mean the execution of the docker-compose tests needs to happen after the OBS checks have built and published the packages. Not sure whether we can easily chain OBS checks and docker-compose tests as they are triggered/executed in a completely different way.

baierjan:
> That would mean the execution of the docker-compose tests needs to happen after the OBS checks have built and published the packages. Not sure whether we can easily chain OBS checks and docker-compose tests as they are triggered/executed in a completely different way.

In the helm chart tests (which also failed), it pulls the container images from OBS. The docker-compose test does the build on its own. The test is calling `docker compose build` https://github.com/os-autoinst/openQA/blob/438907677f550d782a26e44b57b981713abc30ac/tools/test_containers_compose#L30 then note that there is no OBS image specified in https://github.com/os-autoinst/openQA/blob/438907677f550d782a26e44b57b981713abc30ac/container/webui/docker-compose.yaml#L88 which should use the folder with the same name as the source of the container and build it if not present locally. You can see the build in the test logs. The problem is, that the Dockerfile is using openQA package from the distribution and not from the git as you can see in the failing docker-compose test here: https://github.com/os-autoinst/openQA/actions/runs/10199794743/job/28217764031#step:3:989

Another possible solution could be a new Dockerfile to build openQA container locally from git (but that would be duplicating the work OBS does for us) or as was already mentioned, wait for OBS, but even in that case we need to modify the Dockerfile to use the packages from the correct branch and not the released ones from our devel repo.

okurz:
ok, nevermind the docker-compose ideas. Now second try for the original change but with full test coverage of the changed code:
* #5815

PR #4931, opened 2022-11-28
Avoid `database is locked` errors if SQLite writes are slow

Martchus:
* Set SQLite busy timeout before invoking first query/pragma so the extended busy timeout we already apply counts for the first query/pargma as well. This should avoid the `database is locked` error we see on PowerPC workers when setting the journal mode.
* Tested locally by provoking a timeout via `$db->begin('exclusive')` and then never releasing the lock. Before this change `database is locked` is logged rather quickly on another connection attempt. With this change the error occurs only after 10 minutes. The database is still using the `DELETE` journaling mode.
* See https://progress.opensuse.org/issues/120744


okurz (approved):

kalikiana (approved):
Good catch!

kraih (commented):
Good catch indeed. Makes sense the cases we saw logged were so consistent. 😁

PR #4933, opened 2022-11-29
Ensure openQA-single-instance pulls in same version of openQA

Martchus:
So far openQA-single-instance pulls in openQA itself only indirectly via openQA-local-db package that doesn't require a specific openQA version. However, for openQA-single-instance it makes sense to require a specific version of openQA and the worker to avoid pulling in older versions from other repositories (which is likely not expected/wanted by users).

See https://github.com/os-autoinst/openQA/pull/4901 and https://progress.opensuse.org/issues/115784 for additional context.

### DIFF
--- a/dist/rpm/openQA.spec
+++ b/dist/rpm/openQA.spec
@@ -232,7 +232,8 @@ next to the webui.
 Summary:        Convenience package for a single-instance setup
 Group:          Development/Tools/Other
 Requires:       %{name}-local-db
-Requires:       %{name}-worker
+Requires:       %{name} = %{version}
+Requires:       %{name}-worker = %{version}
 Requires:       apache2
 
 %description single-instance


dist/rpm/openQA.spec
```diff
@@ -232,7 +232,8 @@ next to the webui.
 Summary:        Convenience package for a single-instance setup
 Group:          Development/Tools/Other
 Requires:       %{name}-local-db
-Requires:       %{name}-worker
+Requires:       %{name} = %{version}
```
okurz, line 235: so you think local-db can stay with the version-less requirement but here we should pull in the version specific version. Makes sense. I just suggest you move the explanation from the git commit message into a comment here explaining that. The functional change looks fine.

kraih (approved):

kalikiana (approved):

PR #4935, opened 2022-11-29
Avoid running jobs with undetermined worker address

Martchus:
* Auto-detect worker address as part of availability checks to automatically re-check the FQDN and avoid running jobs without worker address
* See https://progress.opensuse.org/issues/120261

---

Still a draft because tests are missing. I'd just like to show how my idea from the mentioned issue would look like.

Note that for determining the IP address as a fallback there wouldn't be a core Perl module (at least I haven't found one). Supposedly https://metacpan.org/pod/Net::Address::IP::Local would be the best module for that.

### DIFF
--- a/lib/OpenQA/Worker/Settings.pm
+++ b/lib/OpenQA/Worker/Settings.pm
@@ -8,6 +8,7 @@ use Mojo::Util 'trim';
 use Config::IniFiles;
 use Time::Seconds;
 use OpenQA::Log 'setup_log';
+use Net::Domain 'hostfqdn';
 
 has 'global_settings';
 has 'webui_hosts';
@@ -81,6 +82,15 @@ sub new ($class, $instance_number = undef, $cli_options = {}) {
     $self->{_file_path} = $settings_file;
     $self->{_parse_errors} = \@parse_errors;
     return $self;
+}
+
+sub auto_detect_worker_address ($self, $fallback = undef) {
+    my $global_settings = $self->global_settings;
+    return 1 if defined $global_settings->{WORKER_HOSTNAME} && !$self->{_worker_address_auto_detected};
+    $self->{_worker_address_auto_detected} = 1;
+    my $worker_address = hostfqdn() // $fallback;
+    $global_settings->{WORKER_HOSTNAME} = $worker_address if defined $worker_address;
+    return defined $worker_address && index($worker_address, '.') >= 0;
 }
 
 sub apply_to_app ($self, $app) {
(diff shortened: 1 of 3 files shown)


lib/OpenQA/Worker/Settings.pm
```diff
@@ -83,6 +84,15 @@ sub new ($class, $instance_number = undef, $cli_options = {}) {
     return $self;
 }
 
+sub auto_detect_worker_address ($self, $fallback = undef) {
```
kraih, line 87: Considering this is more a computed property, maybe a different name would be more descriptive? Like `auto_detected_worker_address` maybe?

Martchus: If `WORKER_HOSTNAME` was not explicitly set, this function will always auto-detect it (again). So the function name doesn't lie. The idea is to avoid getting stuck on a "bad" value. Note that this function does *not* return the address, only whether the auto-detection worked (or `WORKER_HOSTNAME` was set explicitly).

kraih: Oops, i missed the return value.

t/24-worker-overall.t
```diff
@@ -496,6 +496,20 @@ subtest 'checking and cleaning pool directory' => sub {
...
 
+subtest 'checking worker address' => sub {
+    my $fqdn_lookup_mock = Test::MockModule->new('OpenQA::Worker::Settings');
+    $fqdn_lookup_mock->redefine(hostfqdn => 'foo');    # only a "short" hostname, not an FQDN
+    $global_settings->{WORKER_HOSTNAME} = 'foo';
+    is $worker->check_availability, undef, 'no error if worker address explicitly specified (also if no FQDN)';
+    $global_settings->{WORKER_HOSTNAME} = undef;    # assume WORKER_HOSTNAME has not been explicitly specified
+    like $worker->check_availability, qr/Unable.*worker address/, 'error if no FQDN detected';
+    $fqdn_lookup_mock->redefine(hostfqdn => 'foo.bar');
+    is $worker->check_availability, undef, 'no error if worker address can be auto-detected';
```
kalikiana, line 507: I think this doesn't cover the initial case where the shortname is used? Since `check_availability` never specifies a fallback.

kraih (approved):

kalikiana (approved):

okurz (dismissed):
t/24-worker-overall.t failing in OBS check. Likely because there is no proper network in OBS environment.

okurz:
https://progress.opensuse.org/issues/121567 could be an error due to the change.

Martchus:
Yes, the test likely needs to be adapted. I'll work on it.

PR #4937, opened 2022-11-30
Prefer 'pngquant' over 'optipng' if it is installed

marmarek:
pngquant is significantly faster than optipng. On powerful server it
probably doesn't matter much, but on Raspberry Pi, optipng is unusably
slow (it takes over 15s to process one image). pngquant is significantly
faster - about 1s on the same image.

### DIFF
--- a/dist/rpm/openQA.spec
+++ b/dist/rpm/openQA.spec
@@ -190,6 +190,8 @@ Requires(post): os-autoinst >= 4.6
 Recommends:     qemu
 # Needed for caching - not required if caching not used...
 Recommends:     rsync
+# Optionally enabled with USE_PNGQUANT=1
+Recommends:     pngquant
 %if 0%{?suse_version} >= 1330
 Requires(pre):  group(nogroup)
 %endif
--- a/docs/Pitfalls.asciidoc
+++ b/docs/Pitfalls.asciidoc
@@ -57,6 +57,12 @@ SUT behavior as the execution of the first step after loading a snapshot might
 be delayed. This can lead to problems if the executed tests do not foresee an
 appropriate timeout margin.
 
+On some less powerful systems (like Raspberry Pi), reducing screenshots size
+with optipng may take significant amount of time. In this case, you can switch
+to `pngquant` which is significantly faster, but uses lossy compression. To do
+that, install `pngquant` package and set `USE_PNGQUANT=1` in worker or job
+settings.
+
 [[db-migration]]
 == DB migration from SQlite to postgreSQL
 As a first step to start using postgreSQL, please, configure postgreSQL database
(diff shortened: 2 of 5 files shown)


okurz (commented):
Awesome idea for improvement. We specify optipng in some other files, e.g. https://github.com/os-autoinst/openQA/blob/master/dependencies.yaml . How do you think we could handle that? Maybe with "recommends"?

marmarek:
I'm not sure where else dependencies.yaml is used, but added rpm-specific version there.

Martchus (changes requested):
But does it also produce PNGs of the same size (or even lower)? I we now need significantly more space to store screenshots that would be a regression. So please show that we don't need more disk space than before. Or make it a configurable option keeping `optipng` the default.

marmarek:
```
hal9001:~ # cp orig.png proc-optipng.png
hal9001:~ # cp orig.png proc-pngquant.png 
hal9001:~ # time pngquant --force --output proc-pngquant.png proc-pngquant.png

real	0m1.034s
user	0m1.088s
sys	0m0.028s
hal9001:~ # time optipng -quiet -o2 proc-optipng.png

real	0m20.234s
user	0m20.105s
sys	0m0.012s
hal9001:~ # ls -l orig.png proc-*.png
-rw-r--r-- 1 root root 804070 Nov 29 21:54 orig.png
-rw-r--r-- 1 root root 345446 Nov 30 02:08 proc-optipng.png
-rw-r--r-- 1 root root 269122 Nov 30 02:08 proc-pngquant.png
```

marmarek:
The compression isn't really visible (at least with default options), and also needles generated from such pngs seems to work fine (at least in my limited testing). But sure, I'm fine with adding it as an option. What to do with package dependencies then? Revert to just `R: optipng`?

The `optipng` slowness is rather painful, in many test jobs it takes additional 20+ min after the job completes to process and upload screenshots (it doesn't manage to finish that while the job is still running). This is especially bad when working with developer mode and needle editor...

Martchus:
I think the "or" dependency can be kept.

marmarek:
Is this deterministic? I'd rather avoid situation where one has requested (by not changing the config) `optipng`, but resolver chooses to install `pngquant` _instead_.

Martchus:
Good question. Maybe we'd be on the safe side by always requiring `optipng` then. We could still add `pngquant` as a dependency as well (as it doesn't seem to be a heavy dependency anyways). Alternatively, `pngquant` could be a recommended dependency (like qemu and rsync).

marmarek:
> Supposedly it makes more sense to add this as a worker configuration option.

`$self->worker->settings->global_settings`? Or just `$self->info->{settings}` (aka job settings)?

Martchus:
Likely worker settings (`…->global_settings`) make more sense because that's something one likely want to configure mainly on special/slow workers. It would be similar to the setting for using an external video encoder which are also done on worker-level.

marmarek:
worker settings are merged into job settings, so worker settings would take precedence anyway if I'd use just job settings.

Martchus:
Right. I don't have a strong opinion about it. I suppose you can use either setting then.

okurz (approved):

Martchus (approved):

PR #4939, opened 2022-12-01
Add --links option to openqa-cli for pagination

kraih:
I've been thinking about how to use pagination from `openqa-cli`, and a flag that prints all the links to STDERR seems like a reasonable first step.
pagination
It's a very unobtrusive change, since you have to opt-in with the `--links` flag. The green highlighting helps to separate it from the normal response body on STDOUT.

### DIFF
--- a/lib/OpenQA/CLI/api.pm
+++ b/lib/OpenQA/CLI/api.pm
@@ -26,6 +26,7 @@ sub command ($self, @args) {
       'd|data=s' => \$data,
       'f|form' => \my $form,
       'j|json' => \my $json,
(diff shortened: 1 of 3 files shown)


kraih:
You can also use the existing `--verbose` option to manually parse the `Link` headers, but that's a lot harder:
verbose

Martchus (approved):

kraih:
I've added a second commit that adds support for absolute URLs, so you can paste the full link target and it just works:
absolute

Martchus (approved):

perlpunk (approved):

PR #4942, opened 2022-12-06
Catch error and return 404 for module_fails with non existing test

asdil12:
Ticket: https://progress.opensuse.org/issues/121441

### DIFF
--- a/lib/OpenQA/WebAPI/Controller/Test.pm
+++ b/lib/OpenQA/WebAPI/Controller/Test.pm
@@ -727,6 +727,7 @@ sub module_fails {
             job_id => $self->param('testid'),
             name => $self->param('moduleid'),
         })->first;
+    return $self->reply->not_found unless defined $module;
 
     my @needles;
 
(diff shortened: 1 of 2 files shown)


perlpunk (commented):
++ but maybe add a test? There's already a /modules/.../fail test in t/10-tests_overview.t

lib/OpenQA/WebAPI/Controller/Test.pm
```diff
@@ -727,6 +727,7 @@ sub module_fails {
             job_id => $self->param('testid'),
             name => $self->param('moduleid'),
         })->first;
+    return $self->reply->not_found unless defined $module;
```
kalikiana, line 730: I'm with @perlpunk here. I think this would be good to cover in a unit test

Martchus (approved):
A test would be nice of course.

lib/OpenQA/WebAPI/Controller/Test.pm
asdil12: added test

Martchus (approved):

PR #4943, opened 2022-12-06
Protect one asset size update from possible deadlocks

kraih:
This will probably not fix the actual deadlock issue, but using a `SELECT...FOR UPDATE` before the actual `UPDATE` query while inside a transaction is good defensive programming.

I supsect that the other `$asset->refresh_size` call in `OpenQA::Schema::Result::Jobs` is responsible for the deadlock. This change will prove it.

Progress: https://progress.opensuse.org/issues/120891

### DIFF
--- a/lib/OpenQA/Schema/ResultSet/Assets.pm
+++ b/lib/OpenQA/Schema/ResultSet/Assets.pm
@@ -36,7 +36,9 @@ END_SQL
     $schema->txn_do(
         sub {
             $sth->execute($type, $name);    # ensure asset exists
-            return undef unless my $asset = $self->find({type => $type, name => $name}, {key => 'assets_type_name'});
+            return undef
+              unless my $asset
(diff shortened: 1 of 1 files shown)


kraih:
Also verified with `DBI_TRACE=SQL` that the `FOR UPDATE` works as expected:
```
PREPARE dbdpg_p23368_4f AS SELECT me.id, me.type, me.name, me.size, me.checksum, me.last_use_job_id, me.fixed, me.t_created, me.t_updated FROM assets me WHERE ( ( me.name = $1 AND me.type = $2 ) ) FOR UPDATE;

EXECUTE dbdpg_p23368_4f (
$1: 00099963-hdd_image.qcow2
$2: hdd
);
```

mimi1vx (approved):

kalikiana (approved):

PR #4945, opened 2022-12-06
Make auto-detection of `WORKER_HOSTNAME` an optional feature

Martchus:
The `WORKER_HOSTNAME` variable is not be required everywhere so it makes more sense to make it an opt-in. For this one now need to specify `WORKER_HOSTNAME = auto`.

See https://progress.opensuse.org/issues/121567


etc/openqa/workers.ini
```diff
@@ -14,7 +14,8 @@
 
 # host address (domain name or IP address) the web UI can reach the worker by
 # (required by the web UI's developer mode and backends using jump-hosts to
-#  reach back to os-autoinst; set if the automatically deduced value is not
+#  reach back to os-autoinst; set to "auto" to enable auto-detection or
+#  to an explicit domain/address if the automatically deduced value is not
```
okurz, line 18: so I don't know if I understand. If set to "auto" then WORKER_HOSTNAME will be populated with the FQDN. But if we don't specify the variable, like it is here with line 20 commented out, it will not be populated but the worker will still try to start just that the developer mode is likely not working, right?
If this is true, like for the openQA-in-openQA tests, I am not sure why we do this. Maybe you can clarify the effect better at least in the git commit message.

okurz (changes requested):

etc/openqa/workers.ini
Martchus: > If set to "auto" then WORKER_HOSTNAME will be populated with the FQDN.

Right.

> But if we don't specify the variable, like it is here with line 20 commented out, it will not be populated but the worker will still try to start just that the developer mode is likely not working, right?

Right, like it was before we've been auto-initializing `WORKER_HOSTNAME`. Maybe I should double check whether it really doesn't make things worse for the developer mode.

PR #4948, opened 2022-12-07
Fix race-condition reading job module details files

okurz:
A not readable test module result file is handled and gracefully
returned to the web UI and no error in log. The handling should be
atomic, not "check if readable and then read". Instead of trying to read
and then fail we should handle the read attempt as we handle the
exception in line 96 and return {} in case of error.

Related progress issue: https://progress.opensuse.org/issues/121444

### DIFF
--- a/lib/OpenQA/Schema/Result/JobModules.pm
+++ b/lib/OpenQA/Schema/Result/JobModules.pm
@@ -90,10 +90,9 @@ sub results {
     return {} unless my $dir = $self->job->result_dir;
     my $name = $self->name;
 
-    return {} unless -r (my $file = path($dir, "details-$name.json"));
-
-    my $json_data = $file->slurp;
-    die qq{Malformed JSON file "$file": $@} unless my $json = eval { decode_json($json_data) };
+    my $file = path($dir, "details-$name.json");
+    return {} unless my $json_data = eval { $file->slurp };
+    die qq{Malformed/unreadable JSON file "$file": $@} unless my $json = eval { decode_json($json_data) };
 
     # load detail file which restores all results provided by os-autoinst (with hash-root)
     # support also old format which only restores details information (with array-root)


lib/OpenQA/Schema/Result/JobModules.pm
```diff
@@ -90,10 +90,8 @@ sub results {
     return {} unless my $dir = $self->job->result_dir;
     my $name = $self->name;
 
-    return {} unless -r (my $file = path($dir, "details-$name.json"));
-
-    my $json_data = $file->slurp;
-    die qq{Malformed JSON file "$file": $@} unless my $json = eval { decode_json($json_data) };
+    my $file = path($dir, "details-$name.json");
+    die qq{Malformed/unreadable JSON file "$file": $@} unless my $json = eval { decode_json($file->slurp) };
```
Martchus, line 94: Supposedly slurp and decoding needed to be separated to handle those errors distinctively.

perlpunk (commented):
I thought the conclusion in our meeting was that we do **not** want an exception if the file is not there (return `{}`)

lib/OpenQA/Schema/Result/JobModules.pm
okurz: Why? To have distinct error messages?

okurz:
> I thought the conclusion in our meeting was that we do **not** want an exception if the file is not there (return `{}`)

I wanted to see test results and based on those see if maybe the exceptions are already handled on a higher level and turned into properly returned error messages

asdil12:
> I thought the conclusion in our meeting was that we do **not** want an exception if the file is not there (return `{}`)

I think the idea is that we do want to use exceptions to handle that error because they don't introduce race conditions where we can still run into the issue, but catch those exceptions and return something like 404 to the client.

okurz:
Updated the PR to explicitly return `{}` as originally discussed. This also means that tests pass again as the same path is taken as if the complete result directory is not readable anymore.

perlpunk (approved):

mimi1vx (approved):

lib/OpenQA/Schema/Result/JobModules.pm
```diff
@@ -90,10 +90,9 @@ sub results {
     return {} unless my $dir = $self->job->result_dir;
     my $name = $self->name;
 
-    return {} unless -r (my $file = path($dir, "details-$name.json"));
-
-    my $json_data = $file->slurp;
-    die qq{Malformed JSON file "$file": $@} unless my $json = eval { decode_json($json_data) };
+    my $file = path($dir, "details-$name.json");
+    return {} unless my $json_data = eval { $file->slurp };
```
Martchus, line 94: This would consider an empty file an error. In addition, the previous code only returned {} in case the file wasn't readable and now we treat all errors like this.

Martchus: > Why would this treat an empty file as an error?

I mean it would return an empty hash (which is the error handling here).

I guess the change is generally ok but it slightly changes what errors are handled how (instead of just fixing the race condition).

PR #4949, opened 2022-12-07
Support use-case of "local" worker when detecting worker address

Martchus:
* Allow overriding WORKER_HOSTNAME as before (in which case the
  worker never considers itself broken, even if the override is
  no valid FQDN/IP)
* Assign "localhost" as fallback for local worker setups (where
  all HOSTs are local web UI instances) not requiring an FQDN
* Consider worker still broken for remote worker setups unless
  an FQDN could be detected or WORKER_HOSTNAME has been
  specified explicitly
* See https://progress.opensuse.org/issues/121567
  and https://progress.opensuse.org/issues/120261#note-42

### DIFF
--- a/lib/OpenQA/Worker/Settings.pm
+++ b/lib/OpenQA/Worker/Settings.pm
@@ -4,10 +4,12 @@
 package OpenQA::Worker::Settings;
 use Mojo::Base -base, -signatures;
 
+use Mojo::URL;
 use Mojo::Util 'trim';
 use Config::IniFiles;
 use Time::Seconds;
 use OpenQA::Log 'setup_log';
+use OpenQA::Utils 'is_host_local';
 use Net::Domain 'hostfqdn';
 
 has 'global_settings';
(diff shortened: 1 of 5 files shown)


lib/OpenQA/Worker/Settings.pm
```diff
@@ -86,7 +88,18 @@ sub new ($class, $instance_number = undef, $cli_options = {}) {
...
     my $global_settings = $self->global_settings;
-    return 1 if defined $global_settings->{WORKER_HOSTNAME} && !$self->{_worker_address_auto_detected};
+    my $current_address = $global_settings->{WORKER_HOSTNAME};
+
+    # allow overriding WORKER_HOSTNAME explicitly; no validation is done in this case
+    return 1 if defined $current_address && !$self->{_worker_address_auto_detected};
+
+    # assign "localhost" as WORKER_HOSTNAME so entirely local setups work out of the box
+    if ($self->is_local_worker) {
+        $global_settings->{WORKER_HOSTNAME} = 'localhost';
```
okurz, line 98: How about only defining this when no value is specified in WORKER_HOSTNAME. Like, combine with line 91

Martchus: I have no idea how that would make sense.

mimi1vx (approved):

PR #4951, opened 2022-12-07
Docs: Inform about dependencies before information about running tests

clanig:
In the contribution documentation running tests is mentioned before the information about dependencies is written.
Since running integration tests requires these dependencies as well it might be preferred to have that information earlier.


clanig:
@okurz Thank you for pointing this out. I have updated the commit message.

okurz (approved):
Good idea :)

PR #4955, opened 2022-12-12
Split big transaction in stale job detection

Martchus:
* Using a separate transaction per job is likely good enough and has the advantage that other components see changes faster and there's less potential for deadlocks
* Related issue: https://progress.opensuse.org/issues/121768


baierjan (approved):

kraih:
Good idea. We should try to keep all transaction scopes as small as possible.

PR #4956, opened 2022-12-12
Handle dead job status outside of transactions

kalikiana:
As per https://metacpan.org/pod/DBIx::Class::Storage#txn_do results assuming a transaction was executed should only be logged after the entire coderef was executed.

See: https://progress.opensuse.org/issues/121768

### DIFF
--- a/lib/OpenQA/Scheduler/Model/Jobs.pm
+++ b/lib/OpenQA/Scheduler/Model/Jobs.pm
@@ -259,7 +259,7 @@ sub schedule ($self, $allocated_workers = {}, $allocated_jobs = {}) {
         for my $job (@jobs) {
             try {
                 # remove the associated worker and be sure to be in scheduled state.
-                $schema->txn_do(sub { $job->reschedule_state; });
+                log_warning($schema->txn_do(sub { $job->reschedule_state; }));
             }
             catch {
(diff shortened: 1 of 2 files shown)


lib/OpenQA/Scheduler/Model/Jobs.pm
```diff
@@ -468,24 +468,21 @@ sub incomplete_and_duplicate_stale_jobs ($self) {
...
-                    );
-                    my $res = $job->auto_duplicate;
-                    if (ref $res) {
-                        log_warning(sprintf('Dead job %d aborted and duplicated %d', $job->id, $res->id));
-                    }
-                    else {
-                        log_warning(sprintf('Dead job %d aborted as incomplete', $job->id));
-                    }
-                });
+            log_warning(
```
kalikiana, line 471: Obvious drawback here is that all 3 cases share the same log level. But the message only comes up after the transaction finished successfully if I understand the docs correctly.

Martchus (changes requested):
This only goes one level but log messages from e.g. `auto_duplicate` are still unaffected. Printing some log messages only after the transaction and some during the transaction makes things just more inconsistent in my opinion. Printing log messages only after the transaction makes it also hard to follow what was happening within the transaction.

PR #4957, opened 2022-12-12
Avoid jobs from being stuck in state "running"

Martchus:
* Set jobs to RUNNING/UPLOADING only if they're still ASSIGNED/SETUP (and not when they have been set back to SCHEDULED)
* Set jobs back to SCHEDULED only if they were still just ASSIGNED (and haven't been entering any subsequent states meanwhile)
* Prevent stale job detection from setting jobs back to SCHEDULED when the worker meanwhile showed up
* See https://progress.opensuse.org/issues/121768

### DIFF
--- a/lib/OpenQA/Schema/Result/Jobs.pm
+++ b/lib/OpenQA/Schema/Result/Jobs.pm
@@ -323,24 +323,25 @@ sub scenario_description ($self) {
 sub worker_id ($self) { $self->worker ? $self->worker->id : 0 }
 
 sub reschedule_state ($self, $state = OpenQA::Jobs::Constants::SCHEDULED) {
+    # set job to $state/SCHEDULED if it is still just ASSIGNED (not SETUP/RUNNING yet)
+    # note: As this function is invoked as part of the stale job detection the job might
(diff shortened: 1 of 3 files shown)


lib/OpenQA/Schema/Result/Jobs.pm
```diff
@@ -1410,15 +1411,27 @@ sub failed_modules ($self) {
...
 
+    # set job to RUNNING if it is still ASSIGNED/SETUP
+    my %cond = (id => $self->id, state => {-in => [ASSIGNED, SETUP]});
+    $jobs->search(\%cond)->update({state => RUNNING, t_started => now()});
+
+    # abort any further updates when we couldn't set the job to RUNNING
+    # note: That can be the case if the concurrently running stale job detection
+    #       wins the race updating the job state.
+    $self->discard_changes;
+    return {result => 0} unless $self->state eq RUNNING;
```
kalikiana, line 1433: I don't see anything here that makes use of the result. Previously it was always 1 so it might be good to have a unit test that depends on the difference.

okurz (approved):

baierjan (approved):

PR #4961, opened 2022-12-19
Use signatures in WebAPI::Controller::Running

okurz:
Introduce signatures in all methods which are properly covered by tests and
code coverage collection.

Extracted as part of
* #4898
as I could not yet find out why code is not covered by tests.

### DIFF
--- a/lib/OpenQA/WebAPI/Controller/Running.pm
+++ b/lib/OpenQA/WebAPI/Controller/Running.pm
@@ -2,7 +2,7 @@
 # SPDX-License-Identifier: GPL-2.0-or-later
 
 package OpenQA::WebAPI::Controller::Running;
-use Mojo::Base 'Mojolicious::Controller';
+use Mojo::Base 'Mojolicious::Controller', -signatures;
 
 use Mojo::Util 'b64_encode';
 use Mojo::File 'path';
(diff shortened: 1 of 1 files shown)


kraih (dismissed):

lib/OpenQA/WebAPI/Controller/Running.pm
```diff
@@ -191,7 +183,7 @@ sub streaming {
 
     # setup a function to stop streaming again
     my $timer_id;
-    my $close_connection = sub {
+    my $close_connection = sub ($self, @) {
```
asdil12, line 186: What does the `@` do here?

okurz: It consumes [0;:infinity:[ arguments into an unnamed parameter and ignores them.

asdil12 (approved):

lib/OpenQA/WebAPI/Controller/Running.pm
```diff
@@ -118,7 +112,7 @@ sub streamtext {
     # Check for new lines from the logfile using recurring timer
     # Setup utility function to close the connection if something goes wrong
     my $timer_id;
-    my $close_connection = sub {
+    my $close_connection = sub ($self) {
```
perlpunk, line 115: I think the signature should be empty because below it is called like this without any arguments:
`$close_connection->()`

okurz: done

okurz:
@kraih
> Closure should not have arguments.

do you mean "close_connection" or more in general? If the latter could you give a hint how to better identify all closures?

Martchus (approved):

PR #4962, opened 2022-12-19
Don't restart scheduled or running chained parents

AdamWill:
This is a follow-up to @martchus #4498. The intent of that PR, as the commit message says, is to "Restart failed chained parents "up the chain"". However, it actually does more than this: it also restarts chained parents that are queued or running, because they also do not have "OK" results.

In a way we cannot know whether restarting a chained parent that is still running is the right thing to do, but doing so definitely causes problems in one real-world case:
https://progress.opensuse.org/projects/openqav3/activity?from=2022-12-19 and from the examples backing the original change here, it seems we only really intended to add the restart behaviour when the parent test definitely failed, not when its result was not yet decided. So on the principle of making minimal changes to the previous behaviour, it seems reasonable to make this change, which in fact restores the behaviour from before #4498 for still-running parent jobs. The cases #4498 intended to address with complex investigation jobs should still be addressed.

This also updates the comments for `duplicate` to more accurately reflect current behaviour, and highlight the problem that ignoring children of a parent job can ignore not only other jobs in the same parallel group as the child job, but other children of the same parent that are *not* in the same parallel group as the child job.

Related progress issue: https://progress.opensuse.org/issues/110458 Related progress issue: https://progress.opensuse.org/issues/112256

### DIFF
--- a/lib/OpenQA/Schema/Result/Jobs.pm
+++ b/lib/OpenQA/Schema/Result/Jobs.pm
@@ -665,7 +665,6 @@ sub _create_clones ($self, $jobs, @clone_args) {
                     dependency => OpenQA::JobDependencies::Constants::PARALLEL,
                 });
         }
-        # normally we don't clone chained parents, but you never know
         _create_clone_with_parent($res, \%clones, $_, OpenQA::JobDependencies::Constants::CHAINED)
           for @{$info->{chained_parents}};
         _create_clone_with_parent($res, \%clones, $_, OpenQA::JobDependencies::Constants::DIRECTLY_CHAINED)
@@ -750,7 +749,7 @@ sub cluster_jobs ($self, @args) {
             unless ($skip_parents || $cancelmode) {
                 my $parent_result = $p->result;
                 $p->cluster_jobs(jobs => $jobs, skip_children => 1, cancelmode => $cancelmode)
-                  if !$parent_result || !OpenQA::Jobs::Constants::is_ok_result($parent_result);
+                  if !$parent_result || grep { $parent_result eq $_ } NOT_OK_RESULTS;
             }
             next;
         }
@@ -866,13 +865,13 @@ for PARALLEL dependencies:
  + if child is clone, find the latest clone and clone it
 
 for CHAINED dependencies:
-- do NOT clone parents
+- only clone failed parents, ignoring their children (our siblings, but also potentially our cousins from other child groups)
  + create new dependency - duplicit cloning is prevented by ignorelist, webui will show multiple chained deps though
 - clone children
  + if child is clone, find the latest clone and clone it
 
 for DIRECTLY_CHAINED dependencies:
-- clone parents recursively but ignore their children (our siblings)
+- clone parents recursively but ignore their children (our siblings, but also potentially our cousins from other child groups)
  + if parent is clone, find the latest clone and clone it
 - clone children
  + if child is clone, find the latest clone and clone it
--- a/t/05-scheduler-restart-and-duplicate.t
+++ b/t/05-scheduler-restart-and-duplicate.t
@@ -133,12 +133,13 @@ subtest 'restart with (directly) chained child' => sub {
     );
     my $job_data_99926;
     subtest 'cluster jobs for 99937 which has one chained child and one chained parent' => sub {
+        job_get_rs(99926)->update({result => FAILED});
         is_deeply job_get_rs(99937)->cluster_jobs, \%expected_cluster,
-          'chained parent considered for restarting as its result is not ok';
+          'chained parent considered for restarting as its result is failed';
         job_get_rs(99926)->update({result => SOFTFAILED});
         $job_data_99926 = delete $expected_cluster{99926};
         is_deeply job_get_rs(99937)->cluster_jobs, \%expected_cluster,
-          'only child considered for restarting as the parent result is ok';
+          'only child considered for restarting as the parent result is not failed';
     };
 
     # restart the job


AdamWill:
The problem case here looks like this:

4962-problem

If the test realmd_join_sssd is restarted (this can be because it fails and its `RETRY` setting kicks in, or you can just restart it manually in this scenario, both trigger the bug), then its siblings (`server_role_deploy_domain_controller` and `realmd_join_cockpit`) are also restarted, and so is the parent of `realmd_join_cockpit`, `server_cockpit_default` - because its result is not "ok" (its result at this point is "none" - not an OK result - because it has not completed).

Because the code explicitly does *not* go back "down the chain" and pull in the children of `realmd_join_cockpit`, the other two children of that test - `server_cockpit_updates` and `server_cockpit_basic` - do *not* get restarted with the new parent, they are just left as scheduled. The scheduler realizes that their parent test has finished (because it was cancelled) and runs them, and they immediately abort incomplete, because the hard disk image they needed from the parent isn't there, because it was cancelled before it could reach the point of uploading it.

If you check the dependency graph of `realmd_join_sssd` after restarting it, it looks like this:

4962-after

all the jobs there are new ones, with different IDs from the ones in the initial image. Note that the two 'cousin' jobs, `server_cockpit_updates` and `server_cockpit_basic` , have disappeared from the graph completely; if you go up to the build overview, you'd see that they are now incomplete, and restarting them just causes them to instantly fail incomplete again.

With this PR, after restarting `realmd_join_sssd`, the dependency graph looks exactly the same as it did originally, with all six tests present. The three related child tests - `server_role_deploy_domain_controller`, `realmd_join_sssd`, and `realmd_join_cockpit` - are all new jobs; the parent `server_cockpit_default` and the two 'cousin' tests, `server_cockpit_updates` and `server_cockpit_basic`, are *not* new jobs, they are still the original ones.

In a way, the "real" fix here would be to reconsider the "don't go back 'down the chain' to children of parents when restarting" behaviour, and somehow adjust that so it only skips the original child job's direct siblings - jobs it is parallel with - but includes its cousins - jobs that are also children of a shared parent, but which it is *not* parallel with. But I suspect that would be both harder to write and harder to judge the impact of. For practical purposes, this solves the practical problem we (Fedora) are having with our real-world tests, and it should not regress the multi-parent case @Martchus was trying to solve in the first place, as from the examples it looks like he really did only need to restart *failed* parents in that case, not running ones.

(One small mystery here: I don't actually know why, at the initial state here, `server_role_deploy_domain_controller` is considered blocked by the scheduler, but `realmd_join_sssd` is not. That's a bit odd. Unfortunately the web UI doesn't tell you *why* a job is blocked.)

Martchus (commented):
I guess not restarting scheduled/running chained parents is ok.

Martchus:
I'm wondering why in your example `realmd_join_sssd` is already running while the rest of the parallel cluster is still blocked. I would have expected that the parallel cluster isn't run until `server_cockpit_default` finished. That would make more sense (considering the chained dependency) and you wouldn't run into the problem you're trying to solve now. So do you have an idea why `realmd_join_sssd` is already running?

AdamWill:
I'm not sure, but logically speaking, it's correct, and it works. The expressed dependencies are like this:

"server_cockpit_basic": "+START_AFTER_TEST": "server_cockpit_default"
"server_cockpit_updates": "+START_AFTER_TEST": "server_cockpit_default"
"realmd_join_cockpit": "+START_AFTER_TEST": "server_cockpit_default", "PARALLEL_WITH": "server_role_deploy_domain_controller"
"realmd_join_sssd": "PARALLEL_WITH": "server_role_deploy_domain_controller"

"server_cockpit_default" and "server_role_deploy_domain_controller" have no expressed dependencies (I didn't mark role_deploy_domain_controller as being parallel with its client tests, as just having the client tests say they're parallel with the server was always enough).

My "small mystery" note at the bottom of the original report is basically the same question you're asking. I guess I can kinda see how the scheduler decides server_role_deploy_domain_controller, realmd_join_cockpit and server_cockpit_default are inter-related and so holds the server test until server_cockpit_default is done, but decides realmd_join_sssd is OK because its relationship with server_cockpit_default is just too indirect? Not 100% sure, though. In any case, it's always been like this, and it does save a bit of time as realmd_join_sssd can get all the way through the preparation it does before the server comes up, before the server test starts.

Martchus:
> I guess I can kinda see how the scheduler decides server_role_deploy_domain_controller, realmd_join_cockpit and server_cockpit_default are inter-related and so holds the server test until server_cockpit_default is done, but decides realmd_join_sssd is OK because its relationship with server_cockpit_default is just too indirect?

Maybe. However, I'd consider this more a bug than a feature. No matter how indirect a parallel dependency is; it nevertheless means all those jobs are part of the same parallel cluster (same blue box in the graph). And if only on job in that cluster has a chained dependency (blocking one of those jobs) all should be postponed. You've been creating all of these dependencies in one-go, with one post API call, right?

AdamWill:
Yes.

AdamWill:
that...doesn't look like my problem.

lib/OpenQA/Schema/Result/Jobs.pm
```diff
@@ -866,13 +865,13 @@ for PARALLEL dependencies:
  + if child is clone, find the latest clone and clone it
 
 for CHAINED dependencies:
-- do NOT clone parents
+- only clone failed parents, ignoring their children (our siblings, but also potentially our cousins from other child groups)
```
Martchus, line 868: It would make sense to update `WritingTests.asciidoc` as well.

Martchus (approved):

kraih (approved):

AdamWill:
Ah, I was going to address your note above about updating `WritingTests.asciidoc` but didn't get the roundtuits until you merged it :| I guess I'll send a separate PR for that then.

Martchus:
I can update the documentation accordingly if you like.

Martchus:
I've just created a documentation change as part of https://github.com/os-autoinst/openQA/pull/5106 which I needed to create anyways.

PR #4963, opened 2022-12-20
Avoid deadlock when updating asset size during job creation

Martchus:
* Remove `$asset->ensure_size;` on job creation; this breaks the check for assets that's done when restarting jobs, e.g. `t/15-assets.t` fails
* Determine size of assets when restarting jobs as needed so it still works despite the previous point
* See https://progress.opensuse.org/issues/120891

### DIFF
--- a/lib/OpenQA/Schema/Result/Jobs.pm
+++ b/lib/OpenQA/Schema/Result/Jobs.pm
@@ -550,6 +550,8 @@ sub missing_assets ($self) {
     } @relevant_assets;
     return [] unless @assets_query;
     my $assets = $self->result_source->schema->resultset('Assets');
+    my @undetermined_assets = $assets->search({-or => \@assets_query, size => \'is null'}, {order_by => 'id'});
+    $_->ensure_size for @undetermined_assets;
     my @existing_assets = $assets->search({-or => \@assets_query, size => \'is not null'});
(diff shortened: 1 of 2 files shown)


Martchus:
That PR might mean assets show up with the wrong (or no) size in the web UI's assets tab.

okurz (approved):

kraih (approved):

lib/OpenQA/Schema/Result/Jobs.pm
```diff
@@ -1506,8 +1508,9 @@ sub register_assets_from_settings ($self) {
     for my $asset_info (sort { $a->{name} cmp $b->{name} } values %assets) {
         # avoid plain create or we will get unique constraint problems
         # in case ISO_1 and ISO_2 point to the same ISO
+        # note: Not updating the asset size here as doing it in this big transaction
+        #       would lead to deadlocks (see poo#120891).
         my $asset = $assets->find_or_create($asset_info, {for => 'update'});
```
kraih, line 1513: As discussed in the unblock, this `FOR UPDATE` should not be here anymore.

okurz: Handled in https://github.com/os-autoinst/openQA/pull/4981

PR #4964, opened 2022-12-21
Add pagination for GET /api/v1/jobs

asdil12:
https://progress.opensuse.org/issues/121102

### DIFF
--- a/lib/OpenQA/Schema/ResultSet/Jobs.pm
+++ b/lib/OpenQA/Schema/ResultSet/Jobs.pm
@@ -84,8 +84,11 @@ MACHINE. For each set of dupes, only the latest job found is included in
 the return array.
 
 =cut
-sub latest_jobs ($self, $until = undef) {
-    my @jobs = $self->search($until ? {t_created => {'<=' => $until}} : undef, {order_by => ['me.id DESC']});
+sub latest_jobs ($self, $until = undef, $limit = undef, $offset = undef) {
+    my %attrs = (order_by => ['me.id DESC']);
+    $attrs{rows} = $limit if defined $limit;
+    $attrs{offset} = $offset if defined $offset;
+    my @jobs = $self->search($until ? {t_created => {'<=' => $until}} : undef, \%attrs);
 
     my @latest;
     my %seen;
@@ -303,7 +306,7 @@ sub _prepare_complex_query_search_args ($self, $args) {
     $attrs{columns} = $args->{columns} if $args->{columns};
     $attrs{prefetch} = $args->{prefetch} if $args->{prefetch};
     $attrs{rows} = $args->{limit} if $args->{limit};
-    $attrs{page} = $args->{page} || 0;
+    $attrs{offset} = $args->{offset} if $args->{offset};
     $attrs{order_by} = $args->{order_by} || ['me.id DESC'];
     $attrs{join} = \@joins if @joins;
     return (\@conds, \%attrs);
--- a/lib/OpenQA/WebAPI/Controller/API/V1/Job.pm
+++ b/lib/OpenQA/WebAPI/Controller/API/V1/Job.pm
@@ -82,9 +82,12 @@ sub list ($self) {
     $validation->optional('scope')->in('current', 'relevant');
     $validation->optional('limit')->num(0);
     $validation->optional('latest')->num(1);
+    $validation->optional('limit')->num;
+    $validation->optional('offset')->num;
 
     my $limits = OpenQA::App->singleton->config->{misc_limits};
     my $limit = min($limits->{generic_max_limit}, $validation->param('limit') // $limits->{generic_default_limit});
+    my $offset = $validation->param('offset') // 0;
     return $self->render(json => {error => 'Limit exceeds maximum'}, status => 400) unless $limit;
     return $self->reply->validation_error({format => 'json'}) if $validation->has_error;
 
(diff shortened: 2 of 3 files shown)


kraih (dismissed):
You forgot pagination tests with the `Link` headers, like [these](https://github.com/os-autoinst/openQA/blob/c73687c22fdee12430e1dab26365347517101840/t/api/11-bugs.t#L91-L139).

asdil12:
> You forgot pagination tests with the `Link` headers, like [these](https://github.com/os-autoinst/openQA/blob/c73687c22fdee12430e1dab26365347517101840/t/api/11-bugs.t#L91-L139).

I added tests.
Please have a look at how I don't do the pagination properly for the `latest=1` case which would be hard due to the deduplication breaking the `has_more` check.
Maybe you have an idea how this could be handled.
I'm not sure if pagination in that case is actually needed, though.

kraih:
> Please have a look at how I don't do the pagination properly for the `latest=1` case which would be hard due to the deduplication breaking the `has_more` check. Maybe you have an idea how this could be handled. I'm not sure if pagination in that case is actually needed, though.

If none of the global limits applied to it before either, then it's probably fine. Or it was an oversight. 🤷

asdil12:
well if I understand it correctly, the global limit did apply before as well as the page, but the page could have less than `$limit` entries - in theory even 0, due to the subquery and the deduplication in perl code.

okurz (changes requested):

asdil12:

Not sure what you mean - the commit message has a ticket reference and a descriptive message.
If you want something else, please elaborate.

lib/OpenQA/WebAPI/Controller/API/V1/Job.pm
```diff
@@ -82,18 +82,22 @@ sub list ($self) {
     $validation->optional('scope')->in('current', 'relevant');
     $validation->optional('limit')->num(0);
     $validation->optional('latest')->num(1);
+    $validation->optional('limit')->num;
```
kalikiana, line 85: This is duplicated

kalikiana:
> Not sure what you mean - the commit message has a ticket reference and a descriptive message. If you want something else, please elaborate.

The commit message **only** contains a URL. Generally it's good to have at least a sentence about what is being implemented.

If it was me I'd probably write something like this:

```
By handling `row` and `offset` parameters a subset of jobs can be queried. This relies on on the existing `pagination_links_header` feature.

See: https://progress.opensuse.org/issues/121102
```

okurz:
> Not sure what you mean - the commit message has a ticket reference and a descriptive message. If you want something else, please elaborate.

Well, for example explain the reasoning and what design choices you made. Doesn't have to be anything fancy. Something like "As already done in commit ... adding pagination for /api/v1/jobs to prevent server overloading", etc.

t/api/04-jobs.t
```diff
@@ -144,19 +144,78 @@ subtest 'argument combinations' => sub {
...
+        ok $links->{first}, 'has first page';
+        ok !$links->{next}, 'no next page';
+        ok $links->{prev}, 'has previous page';
+
+        $t->get_ok('/api/v1/jobs?limit=18&latest=1')->status_is(200);
+        my @jids_l = ();
+        my @subjids_l = ();
+        for my $j (@{$t->tx->res->json->{jobs}}) {
+            push(@jids_l, $j->{id});
+        }
```
okurz, line 215: duplicated

t/api/04-jobs.t
```diff
@@ -144,19 +144,78 @@ subtest 'argument combinations' => sub {
...
+        ok $links->{next}, 'has next page';
+        ok $links->{prev}, 'has previous page';
+
+        @subjids = @jids[@jids - 18 .. $#jids - 15];
+        $t->get_ok($links->{next}{link})->status_is(200)->json_is("/jobs/0/id" => $subjids[0])
+          ->json_is("/jobs/1/id" => $subjids[1])->json_is("/jobs/2/id" => $subjids[2])->json_hasnt("/jobs/3");
+        $links = $t->tx->res->headers->links;
+        ok $links->{first}, 'has first page';
+        ok !$links->{next}, 'no next page';
+        ok $links->{prev}, 'has previous page';
```
okurz, line 208: this looks a bit duplicated and too much reliant on specific index position in arrays. That should be cleaned up a bit although functionally looks fine

okurz (changes requested):

Martchus (commented):
Generally ok but consider looking into the points mentioned by @okurz .

kraih:
Ticket has been reassigned to me, guess i'll clean up the PR.

lib/OpenQA/Schema/ResultSet/Jobs.pm
```diff
@@ -84,8 +84,11 @@ MACHINE. For each set of dupes, only the latest job found is included in
 the return array.
 
 =cut
-sub latest_jobs ($self, $until = undef) {
-    my @jobs = $self->search($until ? {t_created => {'<=' => $until}} : undef, {order_by => ['me.id DESC']});
+sub latest_jobs ($self, $until = undef, $limit = undef, $offset = undef) {
+    my %attrs = (order_by => ['me.id DESC']);
+    $attrs{rows} = $limit if defined $limit;
+    $attrs{offset} = $offset if defined $offset;
+    my @jobs = $self->search($until ? {t_created => {'<=' => $until}} : undef, \%attrs);
```
kraih, line 91: This was actually dead code.

PR #4967, opened 2022-12-22
Remove unused 'backend_info'

okurz:
The column "backend_info" seems to have been prepared for use by
backends but I also could not find any reference in os-autoinst so this
commit will remove the table in the database, will remove references and
migrate databases by simply removing the column without replacement.

### DIFF
--- a/lib/OpenQA/Schema/Result/Jobs.pm
+++ b/lib/OpenQA/Schema/Result/Jobs.pm
@@ -89,11 +89,6 @@ __PACKAGE__->add_columns(
         is_foreign_key => 1,
         is_nullable => 1
     },
-    backend_info => {
-        # we store free text JSON here - backends might store random data about the job
-        data_type => 'text',
-        is_nullable => 1,
-    },
     TEST => {
         data_type => 'text'
     },
@@ -1037,10 +1032,7 @@ sub append_log ($self, $log, $file_name) {
 }
 
 sub update_backend ($self, $backend_info) {
-    $self->update(
-        {
-            backend => $backend_info->{backend},
-            backend_info => encode_json($backend_info->{backend_info})});
+    $self->update({backend => $backend_info->{backend}});
 }
 
 sub update_result ($self, $result) {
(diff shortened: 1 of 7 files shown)


okurz:
Thanks to @r-richardson who pointed out that this might not be used. Added the `not-ready` label to keep it open for some weeks for consideration before accepting.

kraih:
The patch should probably remove [this line](https://github.com/os-autoinst/openQA/blob/c9f8b107d3c0b21a4ba8098aaf6d2fe95d0aeba3/lib/OpenQA/Schema/Result/Jobs.pm#L1442) too.

okurz:
> The patch should probably remove [this line](https://github.com/os-autoinst/openQA/blob/c9f8b107d3c0b21a4ba8098aaf6d2fe95d0aeba3/lib/OpenQA/Schema/Result/Jobs.pm#L1442) too.

Well, we still need "backend", I removed too much initially. I updated the code to not read/write backend_info only but still keep backend

lib/OpenQA/Schema/Result/Jobs.pm
```diff
@@ -1040,7 +1035,6 @@ sub update_backend ($self, $backend_info) {
     $self->update(
         {
             backend => $backend_info->{backend},
-            backend_info => encode_json($backend_info->{backend_info})});
```
Martchus, line 1043: You still need the closing braces and semicolon.

Martchus (changes requested):

lib/OpenQA/Schema/Result/Jobs.pm
okurz: Sorry, with stupid mistakes like those I really feel I am not yet 100% healthy again

Martchus (approved):

kalikiana (approved):
Nice find

kraih (approved):

kalikiana:
@mergifyio rebase

PR #4971, opened 2023-01-05
Add postgresql service to requirements

baierjan:
If this unit gets activated, the units listed in Requires= will be activated as well. If one of the other units fails to activate, and an ordering dependency After= on the failing unit is set, this unit will not be started. Besides, with or without specifying After=, this unit will be stopped (or restarted) if one of the other units is explicitly stopped (or restarted).

Reference: https://progress.opensuse.org/issues/122578

### DIFF
--- a/systemd/openqa-webui.service
+++ b/systemd/openqa-webui.service
@@ -3,7 +3,7 @@ Description=The openQA web UI
 Wants=apache2.service openqa-setup-db.service
 Before=apache2.service
 After=postgresql.service openqa-setup-db.service openqa-scheduler.service nss-lookup.target remote-fs.target
-Requires=openqa-livehandler.service openqa-websockets.service openqa-gru.service openqa-enqueue-asset-cleanup.timer openqa-enqueue-result-cleanup.timer openqa-enqueue-bug-cleanup.timer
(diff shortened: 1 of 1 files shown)


perlpunk (approved):

kraih (approved):
After checking the systemd docs, the combination of `After` and `Requires` looks correct to me.

Martchus:
What about the other services needing postgresql? Shouldn't they be treated in the same way?

kraih:
> What about the other services needing postgresql? Shouldn't they be treated in the same way?

Or they have a `Requires` on the webui.

perlpunk:
I retriggered CircleCI jobs, we had various problems there

okurz:
hm, this introduces a strong requirement on PostgreSQL. But what for hosts that don't actually run the database but use a remote database that we support just as well? I am not sure if https://www.freedesktop.org/software/systemd/man/systemd.unit.html#Wants=would also help for stopping services but one could try.

baierjan:
Right, this will require to have the webui on the database server. Maybe we can go around by dropping a symlink into _openqa-webui.service.requires_ directory if _postgresql.service_ is detected; or having multiple service file for each use-case, similar to our multiple worker setup (_openqa-worker-plain_ vs. _openqa-worker-auto-restart_). I am afraid that both use-cases cannot be implemented inside one service file. What about having something like _openqa-webui-standalone.service_?

PR #4975, opened 2023-01-09
Revert "Add postgresql service to requirements"

okurz:
Reverts os-autoinst/openQA#4971 as that would mean any openQA instance using a remote database would fail to start requiring a local postgresql.service running as well.


baierjan (approved):
Yes, we will solve the issue in a cleaner way.

Martchus (approved):

baierjan:
Another attempt to tackle the issue: #4976

PR #4976, opened 2023-01-09
Add postgresql service to requirements

baierjan:
Use openqa-webui.service.requires/ and add the requirement only if openQA-local-db is installed.

If this unit gets activated, the units listed in Requires= will be activated as well. If one of the other units fails to activate, and an ordering dependency After= on the failing unit is set, this unit will not be started. Besides, with or without specifying After=, this unit will be stopped (or restarted) if one of the other units is explicitly stopped (or restarted).

Reference: https://progress.opensuse.org/issues/122578

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -110,6 +110,12 @@ install-generic:
 	install -m 755 systemd/systemd-openqa-generator "$(DESTDIR)"/usr/lib/systemd/system-generators
 	install -m 644 systemd/tmpfiles-openqa.conf "$(DESTDIR)"/usr/lib/tmpfiles.d/openqa.conf
 	install -m 644 systemd/tmpfiles-openqa-webui.conf "$(DESTDIR)"/usr/lib/tmpfiles.d/openqa-webui.conf
+	install -d -m 755 "$(DESTDIR)"/usr/lib/systemd/system/openqa-gru.service.requires
+	ln -s ../postgresql.service "$(DESTDIR)"/usr/lib/systemd/system/openqa-gru.service.requires/postgresql.service
+	install -d -m 755 "$(DESTDIR)"/usr/lib/systemd/system/openqa-scheduler.service.requires
+	ln -s ../postgresql.service "$(DESTDIR)"/usr/lib/systemd/system/openqa-scheduler.service.requires/postgresql.service
+	install -d -m 755 "$(DESTDIR)"/usr/lib/systemd/system/openqa-websockets.service.requires
+	ln -s ../postgresql.service "$(DESTDIR)"/usr/lib/systemd/system/openqa-websockets.service.requires/postgresql.service
 #
 # install openQA apparmor profile
 	install -d -m 755 "$(DESTDIR)"/etc/apparmor.d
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -110,6 +110,10 @@ install-generic:
 	install -m 755 systemd/systemd-openqa-generator "$(DESTDIR)"/usr/lib/systemd/system-generators
 	install -m 644 systemd/tmpfiles-openqa.conf "$(DESTDIR)"/usr/lib/tmpfiles.d/openqa.conf
 	install -m 644 systemd/tmpfiles-openqa-webui.conf "$(DESTDIR)"/usr/lib/tmpfiles.d/openqa-webui.conf
+	install -d -m 755 "$(DESTDIR)"/usr/lib/systemd/system/openqa-webui.service.requires
+	ln -s ../postgresql.service "$(DESTDIR)"/usr/lib/systemd/system/openqa-webui.service.requires/postgresql.service
+	install -d -m 755 "$(DESTDIR)"/usr/lib/systemd/system/openqa-gru.service.requires
+	ln -s ../postgresql.service "$(DESTDIR)"/usr/lib/systemd/system/openqa-gru.service.requires/postgresql.service
```
okurz, line 116: So in the case of remote databases this symlink will not have a target file?

Martchus: Then it won't be installed because the symlink is only part of the openQA-local-db package. So this makes actually sense.

Martchus: However, I'm still wondering about other services. E.g. `openqa-websockets.service` comes in the dependency chain *before* `openqa-webui.service` (`openqa-webui.service` requires `openqa-websockets.service` and not vice verca) and also needs the database. So it should supposedly also be handled by this change (and then openqa-webui.service` *could* be dropped as it then would already indirectly require postgres). The same counts for the other services that currently have `After=postgresql.service`.

Martchus (changes requested):

Makefile
baierjan: The link is added by `openQA-local-db` which should be installed in case of remote database, right? So the link shouldn't be there at all.

baierjan: All services have `After=postgresql.service`, but no service actually `Requires=` it and that the source of the problem. Without the require, the ordering is wrong. But as you said, webui requires other services, so if others requires the database, it should theoretically work. I will try that (I did that for the `openqa-gru.service` already where it nicely prevents errors and restarts).

mimi1vx (approved):

kraih (approved):

Martchus (changes requested):
The scheduler also has `After=postgresql.service` and should likely get the same treatment. (I suppose the livehandler can be left out indeed because it already has `Requires=openqa-webui.service`.)

baierjan:
Ah, right. In my installation, this service was not enabled nor started (most likely because there is no dependency between openqa-webui and openqa-scheduler). Yes, you are correct, the openqa-livehandler is already covered.

Martchus (approved):

ggardet:
This commit breaks build on aarch64: https://build.opensuse.org/packages/openQA/job_history/devel:openQA/openSUSE_Factory_ARM/aarch64

PR #4978, opened 2023-01-10
Avoid race condition by waiting for the #text element

kraih:
It seems there's a race condition where the comment form element is not ready in some cases. I could replicate the issue in about 2 of 300 test runs. After adding the `wait_for_element` call i've had the test pass about 600 times without any issues. So this might be a good enough fix.

Progress: https://progress.opensuse.org/issues/121042


kraih:
This is one of the reasons why i prefer Playwright to Selenium, it just waits 30 seconds for an element to be visible with every action and you don't have to do it manually.

Martchus (approved):

PR #4979, opened 2023-01-10
Add pagination for GET /api/v1/jobs

kraih:
Cleaned up version of #4964.

Progress: https://progress.opensuse.org/issues/121102

### DIFF
--- a/lib/OpenQA/Schema/ResultSet/Jobs.pm
+++ b/lib/OpenQA/Schema/ResultSet/Jobs.pm
@@ -84,8 +84,11 @@ MACHINE. For each set of dupes, only the latest job found is included in
 the return array.
 
 =cut
-sub latest_jobs ($self, $until = undef) {
-    my @jobs = $self->search($until ? {t_created => {'<=' => $until}} : undef, {order_by => ['me.id DESC']});
+sub latest_jobs ($self, $until = undef, $limit = undef, $offset = undef) {
+    my %attrs = (order_by => ['me.id DESC']);
+    $attrs{rows} = $limit if defined $limit;
+    $attrs{offset} = $offset if defined $offset;
+    my @jobs = $self->search($until ? {t_created => {'<=' => $until}} : undef, \%attrs);
 
     my @latest;
     my %seen;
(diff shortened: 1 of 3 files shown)


Martchus:
It would still make sense to extend the commit message a little bit. Otherwise it LGTM.

kraih:
> It would still make sense to extend the commit message a little bit. Otherwise it LGTM.

Editing someone else's commit feels weird. 😬

lib/OpenQA/Schema/ResultSet/Jobs.pm
```diff
@@ -84,11 +84,9 @@ MACHINE. For each set of dupes, only the latest job found is included in
...
 
 =cut
-sub latest_jobs ($self, $until = undef, $limit = undef, $offset = undef) {
-    my %attrs = (order_by => ['me.id DESC']);
-    $attrs{rows} = $limit if defined $limit;
-    $attrs{offset} = $offset if defined $offset;
-    my @jobs = $self->search($until ? {t_created => {'<=' => $until}} : undef, \%attrs);
+
+sub latest_jobs ($self, $until = undef) {
+    my @jobs = $self->search($until ? {t_created => {'<=' => $until}} : undef, {order_by => ['me.id DESC']});
```
perlpunk, line 89: I know that the code was unused, but just mentioning that when using `latest_jobs($until)` we don't ensure an upper limit. It's not mentioned as an AC but I think we wanted to ensure an upper limit for all queries.

kraih:
Ok, commit message updated.

lib/OpenQA/Schema/ResultSet/Jobs.pm
kraih: That's correct, the `limit` case is explicitly excluded. That was already part of the previous implementation.

kraih: We could make a followup ticket for figuring out pagination for the `latest` case as well, but it could be pretty hard.

kraih: Also added a comment to the test related to this.

t/api/04-jobs.t
```diff
@@ -144,19 +144,57 @@ subtest 'argument combinations' => sub {
...
+        ok $links->{first}, 'has first page';
+        ok $links->{next}, 'has next page';
+        ok $links->{prev}, 'has previous page';
+
+        $t->get_ok($links->{first}{link})->status_is(200)->json_is('/jobs/0/id' => 99947)
+          ->json_is('/jobs/3/id' => 99963)->json_is('/jobs/4/id' => 99981)->json_hasnt('/jobs/5');
+        $links = $t->tx->res->headers->links;
+        ok $links->{first}, 'has first page';
+        ok $links->{next}, 'has next page';
+        ok !$links->{prev}, 'no previous page';
```
okurz, line 193: I suggest to make the test description strings unique so that the output relates unambiguously to the corresponding code. I suggest to explain in the strings what explicitly the additional test statements cover so that readers know why the tests are not redundant duplication but what the actual benefits are.

okurz (changes requested):

okurz (approved):

perlpunk (approved):

PR #4980, opened 2023-01-10
Fix asset cache error handling to avoid proceeding in case of failures

Martchus:
So far the cache service doesn't treat download errors as such when the
asset is still present locally. However, that's not a good idea because the
local version might be outdated and problems on the server-side might only
surface later (on the next cache cleanup) which makes the investigation
more difficult (see https://progress.opensuse.org/issues/122776).

This change tracks whether the download has actually been succeeded instead
of only checking whether the cached asset exists locally.

Note assets that are locally cached are *not* re-downloaded unless the
asset has changed on the server. However, to check whether the asset has
changed on the server an HTTP request is done that might fail. With this
change this error case is now not ignored anymore.

---

~~Still a draft as tests maybe need to be adapted/extended.~~ However, I've already
tested it locally having the cache service enabled locally and by provoking
a download error for an already cached asset. Of course I've also tested the case
of not running into any download errors.

### DIFF
--- a/lib/OpenQA/CacheService/Client.pm
+++ b/lib/OpenQA/CacheService/Client.pm
@@ -12,6 +12,7 @@ use OpenQA::CacheService::Response::Status;
 use OpenQA::Utils qw(base_host service_port);
 use Socket qw(AF_INET IPPROTO_TCP SOCK_STREAM pack_sockaddr_in inet_aton);
 use Mojo::URL;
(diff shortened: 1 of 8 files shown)


okurz (commented):
+1

mimi1vx (approved):

okurz (approved):

PR #4982, opened 2023-01-13
Change type of `owner` column to bigint in consistency with jobs table

Martchus:
The `owner` column of the `job_locks` table is referencing a job ID so it should be bigint as we use that type for job IDs.

The `job_locks` table is very small (e.g. less than 50 rows on OSD) so this migration should go very fast.


mimi1vx (approved):

Martchus:
By the way, I ran `ALTER TABLE job_locks ALTER COLUMN owner TYPE bigint;` on a local "OSD database" with currently 81 job locks and it worked fine and took only a second.

okurz (approved):
This will conflict with https://github.com/os-autoinst/openQA/pull/4967 but I guess this can be just merged quicker, let's see :)

PR #4984, opened 2023-01-19
Avoid another race condition by waiting for the #text element

kraih:
Similar to the previous race condition from #4978. Found by Marius this time, but i've been unable to replicate it locally. Applying the same `wait_for_element` to ensure that the comment field is visible should resolve this as well though. The test code is almost exactly the same.

Progress: https://progress.opensuse.org/issues/121042


kalikiana (approved):
Good find!

Martchus (approved):

PR #4988, opened 2023-01-26
Avoid errors about changing the working directory after asset extraction

Martchus:
* Set working directory explicitly before invoking archive extraction via `Archive::Extract`
    * Otherwise `cwd` returns an empty string in the case the working directory doesn't exist anymore which is then used by `Archive::Extract` which then runs into an error due to it.
    * Note that the error would not cause an actual problem as it is ignored anyways; however the log message due to it is not nice and avoided by this change.
* See https://progress.opensuse.org/issues/123649

---

Still a draft as I'll have to check whether it actually works on o3. (I've just noticed that my reproducer wouldn't reproduce the issue anymore anyways as the download is already 404.)

### DIFF
--- a/profiles/apparmor.d/usr.share.openqa.script.openqa
+++ b/profiles/apparmor.d/usr.share.openqa.script.openqa
@@ -19,6 +19,7 @@
   /tmp/** rwk,
   /usr/bin/dash ix,
   /usr/bin/perl ix,
+  /usr/bin/pwd ix,
   /usr/bin/ssh rcx,
   /usr/bin/timeout rix,
   /usr/bin/unzip{,-plain} rix,


Martchus:
This doesn't work at all on production. Looks like nothing is messing with the working directory, we really just cannot read it:

```
martchus@ariel:~> sudo -u geekotest pwd
/home/martchus
martchus@ariel:~> sudo -u geekotest /usr/share/openqa/script/openqa eval -V 'use Cwd; "cwd: " . cwd'
Use of uninitialized value in concatenation (.) or string at (eval 1603) line 1.
"cwd: "
```

Locally I've tested what happens when the working directory really doesn't exist anymore and then one gets a different error that reveals Perl's `cwd` is merely invoking the `pwd` binary under the hood. Maybe it is blocked by apparmor or so.

Martchus:
As mentioned in the previous comment this is really a permission error. I've cross-checked locally via `sudo chmod -x /usr/bin/pwd`. Then Perl's `cwd` really only returns an empty string anymore. So the aparmor change will fix the error we see in production. (Why Perl's `cwd` is invoking `pwd` instead of using `getcwd()` from the C standard library is beyond me. However, the code that uses `cwd` is in a dependency so it is easiest to just allow `pwd`.)

okurz (approved):

mimi1vx (approved):

perlpunk (approved):

PR #4992, opened 2023-02-06
ci: Point out how to fix an empty cache

kalikiana:
The cache is tied to the checksums and may not be generated in time when there is a temporary outage or fewer PRs have been prepared.

See: https://progress.opensuse.org/issues/123867


Martchus (approved):

okurz (approved):
How does one trigger this specific test? Or just state that one can create a pull request with arbitrary content, isn't it?

kalikiana:
> How does one trigger this specific test? Or just state that one can create a pull request with arbitrary content, isn't it?

No PR needed, any existing job will do

PR #4993, opened 2023-02-07
Improve appearance of tables

Vogtinator:
Initially I only noticed that if the last <tr/> in a .table had "display: none" the bottom border would be missing entirely. The only way to address that was to redo the way the table border is laid out. This turned out to be shorter than the original code and fixes other issues like missing borders in dark mode.

Additionally, vertically center table contents explicitly. This requires including openqa.scss after bootstrap so that our declarations are preferred.

Before:

Screenshot_20230207_113147
Screenshot_20230207_113310

After:

Screenshot_20230207_113034
Screenshot_20230207_112935

### DIFF
--- a/assets/assetpack.def
+++ b/assets/assetpack.def
@@ -2,7 +2,6 @@
 # download the fonts
 < https://cdn.datatables.net/1.10.16/css/dataTables.bootstrap4.css
 < https://cdn.jsdelivr.net/npm/fork-awesome@1.2.0/css/fork-awesome.min.css
-< stylesheets/openqa.scss
 < https://raw.githubusercontent.com/bootstrapthemesco/bootstrap-4-multi-dropdown-navbar/beta2.0/css/bootstrap-4-navbar.css
 < https://cdnjs.cloudflare.com/ajax/libs/chosen/1.7.0/chosen.css
 << https://raw.githubusercontent.com/twbs/bootstrap/v4.6.1/scss/bootstrap.scss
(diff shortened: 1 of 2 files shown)


Vogtinator:
I encountered the `display: none` case with collapsed child jobs, but I don't have any of those in my test instance. So I had to fake that by adding `display: none` manually... Any other special cases to check? Where is `.noborder` used?

Martchus (commented):
I guess it looks generally good.

---

> but I don't have any of those in my test instance.

You could run `t/ui/10-tests_overview.t` with the env variable `NOT_HEADLESS=1` set. If you put a sleep in the right place, e.g. somewhere within 'stacking of parallel children' you can test your changes on a page that has collapsed child jobs.

Vogtinator:
> You could run `t/ui/10-tests_overview.t` with the env variable `NOT_HEADLESS=1` set. If you put a sleep in the right place, e.g. somewhere within 'stacking of parallel children' you can test your changes on a page that has collapsed child jobs.

I gave it a quick try. `NOT_HEADLESS=1 make test-ui` just fails with database errors...

Martchus (approved):
I've just tested your branch locally and it works (including collapsed rows).

The only thing I've noticed is that the links in the main menu are now less dark (and become as dark as they were before on hover). This change is likely unintended and due to moving the inclusion order. I personally don't find it bad so I'm nevertheless approving.

Vogtinator:
> The only thing I've noticed is that the links in the main menu are now less dark (and become as dark as they were before on hover). This change is likely unintended and due to moving the inclusion order. I personally don't find it bad so I'm nevertheless approving.

Indeed. If that made a difference, that means that bootstrap previously overwrote some openQA style, so this might technically bring back the originally intended behaviour!

kalikiana (approved):

Martchus:
Might this have broken the openQA-in-openQA test? https://openqa.opensuse.org/tests/3105703#step/dashboard/11

Vogtinator:
Looks like the (un)intentional font color change needs a new needle

Martchus:
Or we restore the old color. Shouldn't be more than a line of CSS. I'm just not sure what we want :-)

PR #4995, opened 2023-02-09
Restore font color of main menu actions

Martchus:
* Revert side-effect of 5856b50143038fb121b90b68557092c721167420
* Include CSS for multi-level dropdown extension after Bootstrap itself; otherwise Bootstrap's default won't be overridden by this Bootstrap extension's CSS which is supposedly intended
* See https://progress.opensuse.org/issues/124143


kalikiana (approved):
Good catch!

okurz (approved):

PR #3848, opened 2021-04-19
[WIP] Load job templates from YAML file when scheduling isos

Amrysliu:
The purpose is reading all configurations from a git file (maybe a YAML
file in github) when doing `isos post`.
The format of the YAML file looks like the `09-schedule_from_file.yaml`.

Relate: https://progress.opensuse.org/issues/71758

PR #4999, opened 2023-02-14
Allow loading scenario definitions for scheduling an ISO/product from templates from YAML file

Martchus:
* Revive and rebase #3848 for https://progress.opensuse.org/issues/92311
* Improve tests

---

Open points for discussion:

1. Where should the YAML file come from? The approach from #3848 only allows for local file system paths. Supposedly we want to allow an HTTP-URL here as well (e.g. pointing to some place on GitHub/GitLab/…).
2. It is not to late to tweak the naming. I think `SCHEDULE_FROM_YAML_FILE` is a bad choice as it can be easily confused with `YAML_SCHEDULE`. What about `SCENARIO_DEFINITIONS_YAML_FILE`? After all, that's what this YAML file is going to contain. I don't think we should overuse the work "schedule" because it tends to become more meaningless the more we use it.
3. The YAML file clearly lacks a version field. To be able to extend it in the future, I suggest to add a `meta` section that will contain the `version` field (and possible other meta info).

### DIFF
--- a/docs/UsersGuide.asciidoc
+++ b/docs/UsersGuide.asciidoc
@@ -1146,29 +1146,11 @@ template like this:
 
 [source,sh]
 --------------------------------------------------------------------------------
-products:
-  example:
-    distri: 'example'
-    version: '1'
-    flavor: 'DVD'
-    arch: 'x86_64'
-    settings:
-      REPO: 'http://…'
-machines:
-  64bit:
-    backend: 'qemu'
-    settings:
-      QEMURAM: '3072'
-      WORKER_CLASS: 'qemu_x86_64'
 job_templates:
   create_hdd:
-    product: 'example'
-    machine: '64bit'
     settings:
       PUBLISH_HDD_1: 'example-%VERSION%-%ARCH%-%BUILD%@%MACHINE%.qcow2'
   boot_from_hdd:
-    product: 'example'
-    machine: '64bit'
     settings:
       HDD_1: 'example-%VERSION%-%ARCH%-%BUILD%@%MACHINE%.qcow2'
       START_AFTER_TEST: 'create_hdd'
--- a/docs/UsersGuide.asciidoc
+++ b/docs/UsersGuide.asciidoc
@@ -1048,12 +1048,16 @@ openqa-cli api -X POST jobs TEST=test KEY:WITH:COLONS:=example
 ```
 
 ==== Spawning multiple jobs based on templates - isos post ====
-
 The most common way of spawning jobs on production instances is using the
-`isos post` API route. Based on previously defined settings for media, job
-groups, machines and test suites jobs are triggered based on template
-matching. The <<GettingStarted.asciidoc#gettingstarted,Getting Started>> guide already
-mentioned examples. Additionally to the necessary template matching parameters
+`isos post` API route. Based on settings for media, job groups, machines and
+test suites jobs are triggered based on template matching. These settings
+need to be defined before on the corresponding pages of the web UI (accessible
+to operators from the user menu). The
+<<UsersGuide.asciidoc#job_templates,section on job templates>> already explains
+details about these tables. Alternatively, these settings can be
+supplied via a <<UsersGuide.asciidoc#scenarios_yaml,YAML document>>.
+
+Additionally to the necessary template matching parameters
 more parameters can be specified which are forwarded to all triggered jobs.
 There are also special parameters which only have an influence on the way the
 triggering itself is done. These parameters all start with a leading
@@ -1108,6 +1112,72 @@ corresponding to the parameter `async=0`. Use `async=1` to avoid possible
 timeouts by performing the task in background.
 This is recommended on big instances but means that the results (and
 possible errors) need to be polled via `openqa-cli api isos/$scheduled_product_id`.
+
+[id="scenarios_yaml"]
+===== Defining test scenarios in YAML =====
+Instead of relying on the tables for machines, mediums/products, test suites and
+job templates of the openQA instance, one can provide these definitions/settings
+also via a YAML document. This YAML document could be specific to a certain test
+distribution and stored in the same repository as those tests (making the versioning
+easier).
+
+WARNING: This feature is still experimental and may change in an incompatible way
+in future versions.
+
+This YAML document can be specified via the scheduling parameter
+`SCENARIO_DEFINITIONS_YAML`:
+
+[source,sh]
+--------------------------------------------------------------------------------
+openqa-cli api … -X POST isos --param-file SCENARIO_DEFINITIONS_YAML=/local/file.yaml …
+--------------------------------------------------------------------------------
+
+This command will upload the contents of the local file `/local/file.yaml` to a
+possibly remote openQA instance. The YAML document will only be used within the
+scope of this particular API request. No settings are stored/altered on the openQA
+instance.
+
+If the YAML document already exists on the openQA host, you can also use
+`SCENARIO_DEFINITIONS_YAML_FILE` which expects the file path of the YAML document
+on the openQA host.
+
+The YAML document itself should define at least one product, machine and job
+template like this:
+
+[source,sh]
+--------------------------------------------------------------------------------
+products:
+  example:
+    distri: 'example'
+    version: '1'
+    flavor: 'DVD'
+    arch: 'x86_64'
+    settings:
+      REPO: 'http://…'
+machines:
+  64bit:
+    backend: 'qemu'
+    settings:
+      QEMURAM: '3072'
+      WORKER_CLASS: 'qemu_x86_64'
+job_templates:
+  create_hdd:
+    product: 'example'
+    machine: '64bit'
+    settings:
+      PUBLISH_HDD_1: 'example-%VERSION%-%ARCH%-%BUILD%@%MACHINE%.qcow2'
+  boot_from_hdd:
+    product: 'example'
+    machine: '64bit'
+    settings:
+      HDD_1: 'example-%VERSION%-%ARCH%-%BUILD%@%MACHINE%.qcow2'
+      START_AFTER_TEST: 'create_hdd'
+      WORKER_CLASS: 'job-specific-class'
+
+--------------------------------------------------------------------------------
+
+These definitions are used like their openQA-instance-wide counterparts (so continue
+reading the next section for more details on job templates).
 
 ==== Remarks ====
 
--- a/lib/OpenQA/Schema/Result/ScheduledProducts.pm
+++ b/lib/OpenQA/Schema/Result/ScheduledProducts.pm
@@ -795,10 +795,10 @@ sub _schedule_from_yaml ($self, $args, $skip_chained_deps, @load_yaml_args) {
         push @job_templates, $settings;
     }
 
-    my %result;
-    $result{settings_result} = _handle_dependency(\@job_templates, \%wanted, $skip_chained_deps);
-    $result{error_message} = $error_msg;
-    return \%result;
+    return {
+        settings_result => _handle_dependency(\@job_templates, \%wanted, $skip_chained_deps),
+        error_message => $error_msg,
+    };
 }
 
 sub _count_wanted_jobs ($args, $settings, $wanted) {
--- a/lib/OpenQA/Schema/Result/ScheduledProducts.pm
+++ b/lib/OpenQA/Schema/Result/ScheduledProducts.pm
@@ -231,7 +231,7 @@ sub _schedule_iso {
     _delete_prefixed_args_storing_info_about_product_itself $args;
 
     my $result;
-    if (my $yaml_file = delete $args->{SCHEDULE_FROM_YAML_FILE}) {
+    if (my $yaml_file = delete $args->{SCENARIO_DEFINITIONS_YAML_FILE}) {
         $result = get_schedule_file($yaml_file);
         $result = $self->_schedule_from_yaml_file($args, $skip_chained_deps, $result->{file})
           unless $result->{error_message};
--- a/lib/OpenQA/Schema/Result/ScheduledProducts.pm
+++ b/lib/OpenQA/Schema/Result/ScheduledProducts.pm
@@ -394,9 +394,9 @@ Return settings key for given job settings. Internal method.
 
 =cut
 
-sub _settings_key {
-    my ($settings) = @_;
-    return "$settings->{TEST}\@$settings->{MACHINE}";
+sub _settings_key ($settings) {
+    my ($test, $machine) = ($settings->{TEST}, $settings->{MACHINE});
+    return $machine ? "$test\@$machine" : $test;
 }
 
 =over 4
@@ -736,26 +736,31 @@ sub _schedule_from_yaml ($self, $args, $skip_chained_deps, @load_yaml_args) {
     return {error_message => "YAML validation failed:\n" . join("\n", @$validation_errors)} if @$validation_errors;
 
     my $products = $data->{products};
-    my $machines = $data->{machines};
+    my $machines = $data->{machines} // {};
     my $job_templates = $data->{job_templates};
     my ($error_msg, %wanted, @job_templates);
-    for my $key (keys %$job_templates) {
+    for my $key (sort keys %$job_templates) {
         my $job_template = $job_templates->{$key};
-        next unless my $product_name = $job_template->{product};
-        next unless my $product = $products->{$product_name};
-        next
-          if ( $product->{distri} ne _distri_key($args)
-            || $product->{flavor} ne $args->{FLAVOR}
-            || $product->{version} ne $args->{VERSION}
-            || $product->{arch} ne $args->{ARCH});
         my $settings = $job_template->{settings} // {};
         $settings->{TEST} = $key;
         my @worker_class;
         push @worker_class, $settings->{WORKER_CLASS} if $settings->{WORKER_CLASS};
 
-        my $product_settings = $product->{settings} // {};
-        _merge_settings_uppercase($product, $settings, 'settings');
-        _merge_settings_and_worker_classes($product_settings, $settings, \@worker_class);
+        # add settings from product (or skip if there is no such product) if a product is specified
+        if (my $product_name = $job_template->{product}) {
+            next unless defined $products;
+            next unless my $product = $products->{$product_name};
+            next
+              if ( $product->{distri} ne _distri_key($args)
+                || $product->{flavor} ne $args->{FLAVOR}
+                || $product->{version} ne $args->{VERSION}
+                || $product->{arch} ne $args->{ARCH});
+            my $product_settings = $product->{settings} // {};
+            _merge_settings_uppercase($product, $settings, 'settings');
+            _merge_settings_and_worker_classes($product_settings, $settings, \@worker_class);
+        }
+
+        # add settings from machine if specified
         if (my $machine = $job_template->{machine}) {
             $settings->{MACHINE} = $machine;
             if (my $mach = $machines->{$machine}) {
@@ -765,10 +770,11 @@ sub _schedule_from_yaml ($self, $args, $skip_chained_deps, @load_yaml_args) {
                 $settings->{PRIO} = $mach->{priority} // DEFAULT_JOB_PRIORITY;
             }
         }
+
+        # handle further settings
         $settings->{WORKER_CLASS} = join ',', sort @worker_class if @worker_class > 0;
         _merge_settings_uppercase($args, $settings, 'TEST');
         $settings->{DISTRI} = _distri_key($settings) if $settings->{DISTRI};
-
         OpenQA::JobSettings::parse_url_settings($settings);
         OpenQA::JobSettings::handle_plus_in_settings($settings);
         my $error = OpenQA::JobSettings::expand_placeholders($settings);
--- a/lib/OpenQA/Schema/Result/ScheduledProducts.pm
+++ b/lib/OpenQA/Schema/Result/ScheduledProducts.pm
@@ -14,6 +14,7 @@ use OpenQA::App;
 use OpenQA::Log qw(log_debug log_warning log_error);
 use OpenQA::Utils;
 use OpenQA::JobSettings;
+use OpenQA::Jobs::Constants;
 use OpenQA::JobDependencies::Constants;
 use OpenQA::Scheduler::Client;
 use Mojo::JSON qw(encode_json decode_json);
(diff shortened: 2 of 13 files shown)


okurz:
> 1. Where should the YAML file come from? The approach from #3848 only allows for local file system paths. Supposedly we want to allow an HTTP-URL here as well (e.g. pointing to some place on GitHub/GitLab/…).

How about defaulting to schedule.yaml in CASEDIR?

> 2. It is not to late to tweak the naming. I think `SCHEDULE_FROM_YAML_FILE` is a bad choice as it can be easily confused with `YAML_SCHEDULE`. What about `SCENARIO_DEFINITIONS_YAML_FILE`? After all, that's what this YAML file is going to contain. I don't think we should overuse the work "schedule" because it tends to become more meaningless the more we use it.

I don't see the need to put the file format into the variable name when the filename should already contain that. How about `SCHEDULE_FILE //= 'schedule.yaml'` defaulting to that file and relative to CASEDIR 
* optionally supporting an absolute path
* or also pattern replacement like we use in other cases, e.g. `%CASEDIR%/my_schedule/foo.yaml`

3. The YAML file clearly lacks a version field. To be able to extend it in the future, I suggest to add a `meta` section that will contain the `version` field (and possible other meta info).

@perlpunk?

Martchus:
> How about defaulting to schedule.yaml in CASEDIR?

If `CASEDIR` is a Git repo that would mean we'd have to checkout the full repo for just one file. Doing that on a possibly often used API request seems not like a good idea.

Considering we're mainly talking about this use case I suppose it makes more sense to just give it a URL to obtain that single file.

Or we could even allow to send the file contents directly with the HTTP request. In the context where this route is called the concrete file is likely available in the local filesystem anyways (e.g. when a GitHub action runs it has all the files of the repo in the right version; when one is triggering this locally then it is usually also for a local checkout). Uploading the file contents with the HTTP request also avoids one further request by the web UI (that can go wrong and maybe needs to be handled asynchronously which complicates things further).

> I don't see the need to put the file format into the variable name when the filename should already contain that.

With the suggestion from my previous paragraph it wouldn't. Note that this is also about giving variables intuitively understandable names. If you're still pondering what value you want to assing to that variable the `_YAML`  suffix give you the helpful hint that only YAML documents are allowed here.

> How about `SCHEDULE_FILE //= 'schedule.yaml'` defaulting to that file and relative to CASEDIR

This suggestion still uses "SCHEDULE" which I'd like to avoid as mentioned before. And like I mentioned, when `CASEDIR` is a Git repo it isn't feasible to treat it related to `CASEDIR`.

> or also pattern replacement like we use in other cases, e.g. `%CASEDIR%/my_schedule/foo.yaml`

We have already two conflicting uses of `%`-sign substitution. Adding one more will not make things worse but I suppose this should actually work in the end, right?

okurz:
>> How about defaulting to schedule.yaml in CASEDIR?
> If `CASEDIR` is a Git repo that would mean we'd have to checkout the full repo for just one file. Doing that on a possibly often used API request seems not like a good idea.

That's not what I meant. I would just say that a relative path tries to read from CASEDIR when that is possible. We don't support checking out git repos in openQA (just os-autoinst, different story) so as long as CASEDIR points to a default as e.g. /var/lib/openqa/share/tests/opensuse then we should try to read a file /var/lib/openqa/share/tests/opensuse/schedule.yaml

> Considering we're mainly talking about this use case I suppose it makes more sense to just give it a URL to obtain that single file.

Yes, that special extra case sounds simple enough additional to what I mentioned above.

> Or we could even allow to send the file contents directly with the HTTP request. In the context where this route is called the concrete file is likely available in the local filesystem anyways (e.g. when a GitHub action runs it has all the files of the repo in the right version; when one is triggering this locally then it is usually also for a local checkout). Uploading the file contents with the HTTP request also avoids one further request by the web UI (that can go wrong and maybe needs to be handled asynchronously which complicates things further).

Sorry, I don't understand that paragraph.

>> I don't see the need to put the file format into the variable name when the filename should already contain that.
> With the suggestion from my previous paragraph it wouldn't. Note that this is also about giving variables intuitively understandable names. If you're still pondering what value you want to assing to that variable the `_YAML`  suffix give you the helpful hint that only YAML documents are allowed here.

yeah, good point. Being explicit is good.

>> How about `SCHEDULE_FILE //= 'schedule.yaml'` defaulting to that file and relative to CASEDIR

> This suggestion still uses "SCHEDULE" which I'd like to avoid as mentioned before. And like I mentioned, when `CASEDIR` is a Git repo it isn't feasible to treat it related to `CASEDIR`.

Yes, the latter would still be unsupported for the time being. Consider git repos in CASEDIR out of scope for the case at hand.

>> or also pattern replacement like we use in other cases, e.g. `%CASEDIR%/my_schedule/foo.yaml`

> We have already two conflicting uses of `%`-sign substitution. Adding one more will not make things worse but I suppose this should actually work in the end, right?

This was merely an outlook. I also would prefer to delay any such implementation until we have sorted out the conflicting use cases regarding %CASEDIR%

Martchus:
> That's not what I meant. I would just say that a relative path tries to read from CASEDIR when that is possible. We don't support checking out git repos in openQA (just os-autoinst, different story) so as long as CASEDIR points to a default as e.g. /var/lib/openqa/share/tests/opensuse then we should try to read a file /var/lib/openqa/share/tests/opensuse/schedule.yaml

Ok - that would be very easy to implement. I just though the benefit of just that would be quite low considering this is part of the version control awareness epic. Considering it would be quite easy to implement it is likely still worth the effort, though.

> Sorry, I don't understand that paragraph.

One would *not* provide a parameter that specifies the location of the file as value. One would provide a parameter that has the contents of the file itself as its value. So there's be no indirect fetching of the file contents (and no need to interpret and resolve any reference). For the GitHub-action use-case this would make the most sense. Within the action the file exists "locally" and can simply be uploaded. This is like any HTTP form upload, e.g. when I'll submit this comment the text is also directly uploaded.

> Consider git repos in CASEDIR out of scope for the case at hand.

I think we should consider what we want in the future and not just what we need for the next step. Otherwise we'd end up with several ways of specifying things (because old ways are kept for compatibility) and bad defaults (also for compatibility). And we have enough of these problems already.

> This was merely an outlook. I also would prefer to delay any such implementation until we have sorted out the conflicting use cases regarding %CASEDIR%

Makes sense.

okurz:
> > That's not what I meant. I would just say that a relative path tries to read from CASEDIR when that is possible. We don't support checking out git repos in openQA (just os-autoinst, different story) so as long as CASEDIR points to a default as e.g. /var/lib/openqa/share/tests/opensuse then we should try to read a file /var/lib/openqa/share/tests/opensuse/schedule.yaml
> 
> Ok - that would be very easy to implement. I just though the benefit of just that would be quite low considering this is part of the version control awareness epic. Considering it would be quite easy to implement it is likely still worth the effort, though.
> 
> > Sorry, I don't understand that paragraph.
> 
> One would _not_ provide a parameter that specifies the location of the file as value. One would provide a parameter that has the contents of the file itself as its value. So there's be no indirect fetching of the file contents (and no need to interpret and resolve any reference). For the GitHub-action use-case this would make the most sense. Within the action the file exists "locally" and can simply be uploaded. This is like any HTTP form upload, e.g. when I'll submit this comment the text is also directly uploaded.

ok, got it. But then I would not call that `..._FILE` but something different in a separate variable.

> > Consider git repos in CASEDIR out of scope for the case at hand.
> 
> I think we should consider what we want in the future and not just what we need for the next step.

What I think we want in the future is: A github action workflow like in https://github.com/os-autoinst/os-autoinst-distri-example/compare/master...okurz:os-autoinst-distri-example:feature/hackweek22_trigger_openqa_in_ci can be used to trigger 1..N openQA tests from this git repo on https://openqa.opensuse.org without having anything pre-defined in the openQA database nor the filesystem. Feel welcome to try that out in a spike solution combining the aspects of checking out a git repo as well as reading a schedule file from that git repo. From these notes I created now https://progress.opensuse.org/issues/124502 . Maybe that makes more sense to you as a first next step?

Martchus:
> ok, got it. But then I would not call that ..._FILE but something different in a separate variable.

Yes, that should be a different variable then.

> Feel welcome to try that out in a spike solution combining the aspects of checking out a git repo as well as reading a schedule file from that git repo. From these notes I created now https://progress.opensuse.org/issues/124502 . Maybe that makes more sense to you as a first next step?

Yes, thanks.

mimi1vx:
isn't this duplication of `os-autoinst-distri-opensuse` -> https://github.com/os-autoinst/os-autoinst-distri-opensuse/blob/master/lib/scheduler.pm ??

Martchus:
> isn't this duplication of `os-autoinst-distri-opensuse` -> https://github.com/os-autoinst/os-autoinst-distri-opensuse/blob/master/lib/scheduler.pm ??

Good question but no. The "scheduler" you're mentioning is as far as I know for compiling the list of the test modules of a specific test will run (as an alternative to "main.pm"). This is "scheduling" on os-autoinst-level.

The change proposed here is about creating openQA jobs (and thus on openQA-level). That's also way I renamed the variable to `SCENARIO_DEFINITIONS_YAML_FILE` because that's what the file contains¹ and it avoid confusion with the `YAML_SCHEDULE` variable used on os-autoinst-level.

---

¹ With "scenario definitions" I basically referring to a list of products, machines and the combinations of them we want to test (similar to the settings tables you already find in openQA).

Martchus:
> Feel welcome to try that out in a spike solution combining the aspects of checking out a git repo as well as reading a schedule file from that git repo.

@okurz This generally works with the changes of this PR. I was able to upload the YAML file directly via `curl … --data-urlencode "SCENARIO_DEFINITIONS_YAML@$scenario_defs_file" …` (see https://github.com/Martchus/os-autoinst-distri-example/pull/1/files for the full code). I don't have a meaningful test scenario yet but the scheduling of jobs themselves works (see https://progress.opensuse.org/issues/124502#note-6).

kalikiana:
> 3. The YAML file clearly lacks a version field. To be able to extend it in the future, I suggest to add a `meta` section that will contain the `version` field (and possible other meta info).

The `$id` contains the version, no? That's what we also have in `JobTemplates-01.yaml`. We were planning to bump this if we needed to make incompatible changes.

Martchus:
> The `$id` contains the version, no?

At the time I was writing this the schema hasn't been added yet.

lib/OpenQA/Schema/Result/ScheduledProducts.pm
```diff
@@ -753,4 +727,122 @@ sub _create_download_lists {
     }
 }
 
+sub _schedule_from_yaml ($self, $args, $skip_chained_deps, @load_yaml_args) {
+    my $data = eval { load_yaml(@load_yaml_args) };
+    if (my $error = $@) { return {error_message => "Unable to load YAML: $error"} }
+    my $app = OpenQA::App->singleton;
+    my $validation_errors = $app->validate_yaml($data, 'JobScenarios-01.yaml', $app->log->level eq 'debug');
+    return {error_message => "YAML validation failed:\n" . join("\n", @$validation_errors)} if @$validation_errors;
```
kalikiana, line 735: This doesn't seem to be caught in the tests?

lib/OpenQA/Schema/Result/ScheduledProducts.pm
```diff
@@ -753,4 +727,122 @@ sub _create_download_lists {
     }
 }
 
+sub _schedule_from_yaml ($self, $args, $skip_chained_deps, @load_yaml_args) {
+    my $data = eval { load_yaml(@load_yaml_args) };
+    if (my $error = $@) { return {error_message => "Unable to load YAML: $error"} }
```
kalikiana, line 732: This doesn't seem to be caught in the tests?

Martchus: It is covered. It is the check for `Could not open 'does-not-exist.yaml' for reading: No such file or directory`.

lib/OpenQA/Schema/Result/ScheduledProducts.pm
Martchus: It is covered. It is the check using `my @expected_errors`.

lib/OpenQA/Schema/Result/ScheduledProducts.pm
```diff
@@ -753,4 +727,122 @@ sub _create_download_lists {
...
+        $settings->{WORKER_CLASS} = join ',', sort @worker_class if @worker_class > 0;
+        foreach (keys %$args) {
+            $settings->{uc $_} = $args->{$_} if $_ ne 'TEST';
+        }
+        $settings->{DISTRI} = lc($settings->{DISTRI}) if $settings->{DISTRI};
+
+        OpenQA::JobSettings::parse_url_settings($settings);
+        OpenQA::JobSettings::handle_plus_in_settings($settings);
+        my $error = OpenQA::JobSettings::expand_placeholders($settings);
+        $error_msg .= $error if defined $error;
```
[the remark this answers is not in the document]
Martchus: I think it is actually useful to present the user with a list of errors, e.g. like a compiler doesn't stop at the first error but shows you all the errors (in the current compilation unit). This way you don't have to invoke it over and over again and can instead tackle multiple problems in one go.

Martchus:
> Feel welcome to try that out in a spike solution combining the aspects of checking out a git repo as well as reading a schedule file from that git repo.

@okurz As mentioned in https://github.com/os-autoinst/openQA/pull/4999#issuecomment-1440347301, this generally works. With https://github.com/os-autoinst/os-autoinst/pull/2263 I was now also able to do a passing run (see https://progress.opensuse.org/issues/124502#note-8 for details).

---

So I guess we could merge this PR and https://progress.opensuse.org/issues/124502 and then I can also re-create https://github.com/Martchus/os-autoinst-distri-example/pull/1 to target the main repo.

lib/OpenQA/Schema/Result/ScheduledProducts.pm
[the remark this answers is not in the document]
kalikiana: Right, that's a fair point. I would still prefer to not schedule any jobs unless all of them are fine, though.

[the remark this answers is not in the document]
Martchus: > I would still prefer to not schedule any jobs unless all of them are fine, though.

That's actually how it works. In the code that is "gathering" different error messages no jobs are created visibly or even persistently yet. That happens only later and if errors occurred here we abort before it comes to this.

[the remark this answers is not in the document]
kalikiana: Okay. I mentioned `schedule_iso` because it is documented to behave the way I describe it. If that's not what is happening here I'm happy to be corrected 😊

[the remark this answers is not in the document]
kalikiana: > Returns a list of job ids from the jobs that were successfully scheduled                    
  and a list of failure reason for the jobs that could not be scheduled.

Specifically is what I was referring to

lib/OpenQA/Schema/Result/ScheduledProducts.pm
Martchus: I added this error-message-prefix now to the checks (in both cases).

lib/OpenQA/Schema/Result/ScheduledProducts.pm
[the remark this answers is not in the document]
Martchus: Yes, but this distinction is not about problems that can happen *at this point*. So it is really out of scope of my changes. (And in case of problems like an unhandled exception one would also just get a 500 response instead of a half set of jobs being created as the actual creation of jobs in the DB happens in a transaction.)

kalikiana (approved):

Martchus:
By the way, documentation is already on https://github.com/os-autoinst/openQA/compare/master...Martchus:openQA:yaml-isos-post-doc. I'd propose this if this as draft when this PR is merged. I would keep this a hidden feature until the pending os-autoinst PR and has been merged and the example distri has been setup.

okurz (commented):
this looks very good. As you are proposing a new file format I would like to be a bit more conservative and do two things:
1. wait some more days for feedback from others if any
2. do a walkthrough together through the code to better understand it. I am sure we will find a timeslot for this the following days.

lib/OpenQA/Schema/Result/ScheduledProducts.pm
```diff
@@ -753,4 +727,122 @@ sub _create_download_lists {
...
+    my $validation_errors = $app->validate_yaml($data, 'JobScenarios-01.yaml', $app->log->level eq 'debug');
+    return {error_message => "YAML validation failed:\n" . join("\n", @$validation_errors)} if @$validation_errors;
+
+    my $products = $data->{products};
+    my $machines = $data->{machines};
+    my $job_templates = $data->{job_templates};
+    my $error_msg;
+    my %wanted;
+    my @job_templates;
+    foreach my $key (keys %$job_templates) {
```
perlpunk, line 743: My usual suggestion to iterate over sorted keys to ensure reproducability:
```suggestion
    foreach my $key (sort keys %$job_templates) {
```

lib/OpenQA/Schema/Result/ScheduledProducts.pm
```diff
@@ -753,4 +727,122 @@ sub _create_download_lists {
...
+                my $machine_settings = $mach->{settings} // {};
+                foreach (keys %$machine_settings) {
+                    if ($_ eq 'WORKER_CLASS') {
+                        push @worker_class, $machine_settings->{WORKER_CLASS};
+                        next;
+                    }
+                    $settings->{$_} = $machine_settings->{$_};
+                }
+                $settings->{BACKEND} = $mach->{backend} if $mach->{backend};
+                $settings->{PRIO} = $mach->{priority} // 50;
```
Martchus, line 780: The `50` should be a constant.

lib/OpenQA/Schema/Result/ScheduledProducts.pm
```diff
@@ -753,4 +727,122 @@ sub _create_download_lists {
     }
 }
 
+sub _schedule_from_yaml ($self, $args, $skip_chained_deps, @load_yaml_args) {
```
Martchus, line 730: This code is similar to `_generate_jobs` but only *similar*. Not sure whether it makes sense to unify it.

lib/OpenQA/Schema/Result/ScheduledProducts.pm
```diff
@@ -753,4 +727,122 @@ sub _create_download_lists {
...
+        if (my $machine = $job_template->{machine}) {
+            $settings->{MACHINE} = $machine;
+            if (my $mach = $machines->{$machine}) {
+                my $machine_settings = $mach->{settings} // {};
+                foreach (keys %$machine_settings) {
+                    if ($_ eq 'WORKER_CLASS') {
+                        push @worker_class, $machine_settings->{WORKER_CLASS};
+                        next;
+                    }
+                    $settings->{$_} = $machine_settings->{$_};
```
Martchus, line 777: We could extract some nested loops like this into a function.

lib/OpenQA/Schema/Result/ScheduledProducts.pm
```diff
@@ -517,9 +529,6 @@ sub _generate_jobs {
 
     my %wanted;    # jobs specified by $args->{TEST} or $args->{MACHINE} or their parents
 
-    # Allow a comma separated list of tests here; whitespaces allowed
-    my @tests = $args->{TEST} ? split(/\s*,\s*/, $args->{TEST}) : ();
-
```
Martchus, line 522: Note that this removal went into `_count_wanted_jobs`.

lib/OpenQA/Schema/Result/ScheduledProducts.pm
Martchus: The "normal" code (using the DB) doesn't sort job templates. I'm not sure whether it is worth sorting this data (as it also has a cost).

perlpunk: The cost of sort can be ignored here. This function will be called rarely enough and the number of items will be small enough.
But I don't insist...

lib/OpenQA/Schema/Result/ScheduledProducts.pm
```diff
@@ -753,4 +727,122 @@ sub _create_download_lists {
...
+        push @job_templates, $settings;
+    }
+
+    return {
+        settings_result => _handle_dependency(\@job_templates, \%wanted, $skip_chained_deps),
+        error_message => $error_msg,
+    };
+}
+
+sub _count_wanted_jobs ($args, $settings, $wanted) {
```
Martchus, line 804: This function is not really counting things. It doesn't return a number but populates the `$wanted` hash instead. So it should be renamed and possibly also have a bit more comments.

lib/OpenQA/Schema/Result/ScheduledProducts.pm
```diff
@@ -753,4 +727,122 @@ sub _create_download_lists {
...
+                if ($test eq $settings->{TEST}) {
+                    $wanted->{_settings_key($settings)} = 1;
+                    last;
+                }
+            }
+        }
+    }
+}
+
+sub _handle_dependency ($jobs, $wanted, $skip_chained_deps) {
```
Martchus, line 821: This function could also have a better name and maybe also deserves better comments (to better convey its function in the bigger picture).  (The function is self is of course nothing new. The code has simply been extracted to avoid duplication.)

lib/OpenQA/Schema/Result/ScheduledProducts.pm
```diff
@@ -753,4 +727,122 @@ sub _create_download_lists {
...
+        }
+        if (my $machine = $job_template->{machine}) {
+            $settings->{MACHINE} = $machine;
+            if (my $mach = $machines->{$machine}) {
+                my $machine_settings = $mach->{settings} // {};
+                foreach (keys %$machine_settings) {
+                    if ($_ eq 'WORKER_CLASS') {
+                        push @worker_class, $machine_settings->{WORKER_CLASS};
+                        next;
+                    }
```
okurz, line 776: duplicating line L762-765

docs/UsersGuide.asciidoc
```diff
@@ -1109,6 +1113,72 @@ timeouts by performing the task in background.
...
+possibly remote openQA instance. The YAML document will only be used within the
+scope of this particular API request. No settings are stored/altered on the openQA
+instance.
+
+If the YAML document already exists on the openQA host, you can also use
+`SCENARIO_DEFINITIONS_YAML_FILE` which expects the file path of the YAML document
+on the openQA host.
+
+The YAML document itself should define at least one product, machine and job
+template like this:
```
okurz, line 1145: As discussed 2023-03-09 we shouldn't see product+machine as mandatory. With *just* job templates defining all variables that effectively makes them "scenarios". Maybe we can have as first example the most simple scenario definition and after that provide the optional products+machines extensions

Martchus: This is not yet possible. If it should already be optional in the first merged version I can work on that.

okurz: I'd prefer that

perlpunk (approved):

PR #5009, opened 2023-02-17
Enable build for PowerPC after openQA packages are no longer noarch

Martchus:
See https://progress.opensuse.org/issues/124670

### DIFF
--- a/dist/rpm/openQA.spec
+++ b/dist/rpm/openQA.spec
@@ -165,7 +165,7 @@ Development package pulling in all build+test dependencies except chromedriver f
 Summary:        Development package pulling in all build+test dependencies
 Group:          Development/Tools/Other
 Requires:       %{devel_requires}
-%ifarch ppc ppc64 ppc64le s390x
+%ifarch s390x
 # missing chromedriver dependency
 ExclusiveArch:  do_not_build
 %endif


dist/rpm/openQA.spec
```diff
@@ -165,7 +165,7 @@ Development package pulling in all build+test dependencies except chromedriver f
 Summary:        Development package pulling in all build+test dependencies
 Group:          Development/Tools/Other
 Requires:       %{devel_requires}
-%ifarch ppc ppc64 ppc64le s390x
+%ifarch s390x
```
okurz, line 168: I don't understand how that should work. If there is a missing chromedriver dependency for ppc and others why should you be able to remove it? It doesn't look you like tried that in a branched OBS project, did you?

Martchus: I would have just enabled it and see what happens. Who knows whether that comment is still relevant. Considering the build is actually attempted (we have a build log, e.g. https://build.opensuse.org/package/live_build_log/devel:openQA/openQA/openSUSE_Leap_15.4/ppc64le) it at least doesn't fail to resolve dependencies on OBS level.

kraih (approved):
Not sure this will fix it, but it seems to at least be the cause for the issue, judging by how the error message is generated.

dist/rpm/openQA.spec
Martchus: Since we don't run chromedriver-based tests in OBS (`rm … t/ui/*.t`) and this is only for the devel-meta-package this will supposedly even just work as-is. Of course one needs to resort to the `no-selenium-devel` on PowerPC then (but likely we don't need the devel package there anyways).

okurz (approved):
ok, let's try

kraih:
In retrospect, this might not have been the smartest decision for a Friday afternoon.

PR #5010, opened 2023-02-17
Fix permission denied and sass application triggered by recent dependency upgrades

okurz:
Please check the individual commits.

I am not sure if the apparmor change is enough as the apparmor profile difference on o3 vs. the version in the git repo is rather big.

Related progress issue: https://progress.opensuse.org/issues/124739


kraih:
I don't mind the change, but the root problem has to be the package building on OBS. If i remember correctly, assets are supposed to be built there and distributed with the package. After deployment the asset pipeline dependencies like Sass should not be necessary.

okurz:
yes, that's correct

PR #5013, opened 2023-02-17
Revert the 'Enable ppc64le build' commit trying to fix asset cache

okurz:


perlpunk:
I think #5009 was trying to fix https://progress.opensuse.org/issues/124670 (openQA does not build on ppcle64 anymore since we removed noarch `error: Architecture is not included: ppc64le`), not the asset cache

okurz:
yes, we don't need this PR anymore. https://github.com/os-autoinst/openQA/pull/5011 was the right revert

PR #5014, opened 2023-02-17
Fix regex to get AssetPack plugins

perlpunk:
Before it would return this:

    % grep 'AssetPack' lib/OpenQA/WebAPI.pm | awk -F '[()]' '{print $3}'
    Sass Css JavaScript Fetch Combine
     if $Mojolicious::Plugin::AssetPack::VERSION > 2.13;

Now the regex is more specific and also restricted to one line

Issue: https://progress.opensuse.org/issues/124739

Also reverts #5011


kraih (approved):
🙈

okurz:
Looks good but I suggest to harden tools/generate-packed-assets first. Make it fail properly when it encounters errors.

perlpunk:
> Looks good but I suggest to harden tools/generate-packed-assets first. Make it fail properly when it encounters errors.

done

okurz (approved):
Very good. Thank you. Although I suggest you only remove the blocking label when you can monitor the impact better, like next working day for you or so :)

PR #5016, opened 2023-02-20
Allow escaping percentage signs in job settings to avoid substitution

Martchus:
Sometimes an expression surrounded by `%`-signs should be preserved and not be substituted, e.g. `NEEDLES_DIR=%CASEDIR%/foo` to express a `NEEDLES_DIR` relative to `CASEDIR`. This change enables the use of `NEEDLES_DIR=%%CASEDIR%%/foo` which will always be substituted to `NEEDLES_DIR=%CASEDIR%/foo` (so no real substitution happens, only the outermost pair of `%`-signs is removed).

The documentation for this specific use has also been extended (including how to pass `%`-signs via `curl`).

Of course this could be useful in other cases as well. Note that this particular use case is from day 4 of Oliver's hackweek (see https://hackweek.opensuse.org/22/projects/trigger-actual-openqa-tests-in-pull-requests).

Related ticket: https://progress.opensuse.org/issues/124502

### DIFF
--- a/docs/WritingTests.asciidoc
+++ b/docs/WritingTests.asciidoc
@@ -1633,10 +1633,12 @@ way:
 * A relative `NEEDLES_DIR` is treated to be relative to the default `CASEDIR`
   (even if `CASEDIR` is customized). To have it treated to be relative to the
   custom `CASEDIR`, prefix the relative path with `%CASEDIR%/`. So specifying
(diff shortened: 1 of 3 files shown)


kalikiana (approved):

mimi1vx (approved):
wow, this .. I needed when substitution was enabled for QEM workflows :D

PR #5017, opened 2023-02-20
Allow re-routing via `Mojolicious::Plugin::RequestBase`

Martchus:
This is useful to put openQA behind a reverse proxy making it not available at the "root" but e.g. `https://my-host/openqa`.

Enabling this plugin by adding `plugins = RequestBase` to `openqa.ini` already works without any code changes. The only thing missing are several AJAX calls which would still just use `/` as route.


kraih:
Instead of `urlBase + '/tests/' + event_data.job_id + '#comments'` everywhere, wouldn't it be nicer to use a function, like `urlWithBase('/tests/' + event_data.job_id + '#comments')`?

assets/javascripts/admin_assets.js
```diff
@@ -69,7 +69,7 @@ function setupAdminAssets() {
           if (!data) {
             return 'none';
           }
-          return '<a href="/tests/' + data + '">' + data + '</a>';
+          return '<a href="' + urlWithBase('/tests/' + data) + '">' + data + '</a>';
```
kalikiana, line 72: Did you look into whether we can test this somehow? Maybe mock it or use a different base in a container test?
I'm prepared to accept that it may be way too complicated, but still wanted to ask.

kraih: At least some basic testing would probably be a good idea. You can fake the URL base with a hook. Something like:
```perl
$t->app->hook(before_dispatch => sub ($c) { $c->req->url->base->path(...) });
```

Martchus: Running all UI tests again with re-routing is likely no worth the effort. Designing a special case that will run all the affected actions with re-routing is likely also too much effort. I could add a basic tests that enables the plugin, somehow adds the HTTP header (not sure how to do this with Selenium) and check whether e.g. tables on "All tests" load. It would still be quite some effort considering the plugin also first needs to be packaged. This would also by no means guarantee that at some  point someone contributes a change that breaks when re-routing is used. I suppose this kind of setup will always be "best effort". So I tend to leave it as-is.

kraih: You don't need the plugin, that hook i showed above does the same thing the plugin would.

Martchus: Ok, I'll add basic testing. And right, for the sake of testing I can actually get away with this one-liner. (The plugin itself doesn't do that much more actually, see https://metacpan.org/dist/Mojolicious-Plugin-RequestBase/source/lib/Mojolicious/Plugin/RequestBase.pm)

perlpunk (approved):

kraih (approved):

PR #5019, opened 2023-02-21
Replace fragile awk script with AssetPack config file

kraih:
Progress: https://progress.opensuse.org/issues/124757


mimi1vx (approved):

lib/OpenQA/WebAPI.pm
```diff
@@ -10,6 +10,7 @@ use OpenQA::Log 'setup_log';
 use OpenQA::Utils qw(detect_current_version service_port);
 use OpenQA::Setup;
 use OpenQA::WebAPI::Description qw(get_pod_from_controllers set_api_desc);
+use OpenQA::YAML qw(load_yaml);
```
perlpunk, line 13: I think you could just use `YAML::PP qw(LoadFile)` here. We don't need the special settings that `OpenQA::YAML` has

kraih: Sure.

perlpunk (approved):

PR #5021, opened 2023-02-23
Do not restart user-cancelled job with RETRY

baierjan:
Job restarts via `RETRY` should skip jobs with OK_RESULTS and also USER_CANCELLED jobs. This means that having retry for complement of `is_ok` is not enough. Introduce `is_ok_to_retry` check to deal with that.

Reference: https://progress.opensuse.org/issues/124961


mimi1vx (approved):

perlpunk (approved):

Martchus (commented):
It would have been nice to extend tests.

PR #5022, opened 2023-02-23
Add openQA status badges

asdil12:
Ticket: https://progress.opensuse.org/issues/124173

The badges can be accessed for any job via:

```
/tests/123/badge
```
```
/tests/latest/badge
```

They support the get parameter `show_build=1` that will
prefix the status with the build number (useful for the `/latest/badge`).

The badge will mostly use the colors as defined in the openQA theme but there are some small changes to ensure readability with the white text.

The width of the badge is automatically aligned to the length of the status text.


asdil12:
blocked
incomplete
passed
running
scheduled
softfailed
timeout_exceeded
uploading
user_cancelled

templates/webapi/test/badge.svg.ep
```diff
@@ -0,0 +1,58 @@
...
+%# now substract weighted for smaller width characters
+% $res_width -= $charlen * (() = $result_text =~ /f/g) * 0.4;
+% $res_width -= $charlen * (() = $result_text =~ /i/g) * 0.6;
+% $res_width -= $charlen * (() = $result_text =~ /j/g) * 0.5;
+% $res_width -= $charlen * (() = $result_text =~ /l/g) * 0.6;
+% $res_width -= $charlen * (() = $result_text =~ /t/g) * 0.5;
+% $res_width -= $charlen * (() = $result_text =~ /1/g) * 0.6;
+% $res_width -= $charlen * (() = $result_text =~ /\./g) * 0.7;
+% $res_width -= $charlen * (() = $result_text =~ /:/g) * 0.7;
+% my $width = $res_width + 75;
```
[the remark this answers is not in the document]
Martchus: However, I would get rid of all of this and let a little bit of JavaScript figure it out. That's how it would look like:

```
<svg height="20" onload="bb = document.querySelector('#text').getBBox(); this.setAttribute('width', bb.x + bb.width + 5)" xmlns:xlink="http://www.w3.org/1999/xlink" xmlns="http://www.w3.org/2000/svg">
  <clipPath id="a">
    <rect rx="3" width="100%" height="100%"/>
  </clipPath>
  <linearGradient id="b" x2="0" y2="100%">
    <stop offset="0" stop-color="#bbb" stop-opacity=".1"/>
    <stop offset="1" stop-opacity=".1"/>
  </linearGradient>
  <g clip-path="url(#a)">
    <rect width="100%" height="100%" fill="#555"/>
    <rect x="85" width="100%" height="100%" fill="#ff6644"/>
    <rect height="20" fill="url(#b)"/>
    <g fill="#fff" font-family="DejaVu Sans,Verdana,Geneva,sans-serif" font-size="11" id="text">
      <text x="30" y="15" fill="#010101" fill-opacity=".3">openQA</text>
      <text x="30" y="14">openQA</text>
      <text x="90" y="15" fill="#010101" fill-opacity=".3">foobar</text>
      <text x="90" y="14">foobar</text>
    </g>
  </g>
  <image height="14" width="14" x="6" xlink:href="data:image/svg+xml;…" y="3" font-family="Verdana,Geneva,DejaVu Sans,sans-serif" text-anchor="middle" text-rendering="geometricPrecision"/>
</svg>
```

I have removed the template substitution for testing purposes here so you'll have to add it back (and as mentioned in the other comment get rid of the base64-encoded part). If you want to play around with it further, you can checkout http://jsfiddle.net/db3fk61e/1/.

Martchus (changes requested):
A unit test is missing. Note that functions of https://docs.mojolicious.org/Test/Mojo (like `element_exists`) should also work with SVG.

templates/webapi/test/badge.svg.ep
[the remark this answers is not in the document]
Martchus: By the way, this JavaScript is not black magic. It is using normal DOM-APIs and `getBBox()` which is [documented](https://developer.mozilla.org/en-US/docs/Web/API/SVGGraphicsElement/getBBox).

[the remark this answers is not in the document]
asdil12: The downside seems to be that only browsers seem to be able to view the svg correctly then.
eom and gimp don't seem to understand the js.

But I guess that is an issue that we can bear with.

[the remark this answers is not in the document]
Martchus: Maybe you want to assign a default with (e.g. `width="200"`) so it isn't completely broken without JavaScript.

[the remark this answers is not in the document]
Martchus: > But I guess that is an issue that we can bear with.

Yes, because it is supposed to be embedded on a web page anyways.

templates/webapi/test/badge.svg.ep
```diff
@@ -0,0 +1,58 @@
...
...A2My4wMjYtMTIuNzkyIDc3LjU5MiA2LjM1OSAxNC41NTIgMTkuMTY1IDE4LjQ0IDM0LjE2IDEzLjcyNCA3NS45NjVsLS4wNDUuMjIyYTg0LjU4NyA4NC41ODcgMCAwIDAgMjEuNjk0IDgzLjgxOGMyMC43MDMgMjAuNjczIDQ3LjA3IDI0LjU0NyA3NS42MSAxOC4wODVsLS42NS0zNi4wODJjLTE2Ljg3MyA2LjY5OS0zMi42MzcgMi43Mi00NS40ODgtMTAuMTE1YTQ1LjUwMiA0NS41MDIgMCAwIDEgNjQuMzI4LTY0LjMyN2M5LjEwOSA5LjEyNCAxMi43MDIgMjEuOTQ1IDEyLjM3NyAyOC4zNjNsLS45NjEgODUuMDNoMjAuMjZsLjE2Mi4xNjNoMzUuNjgzbDExLjc0Mi0zMS43MDYtMzQuNjA0LS4xNDdjLS4xMDQtMjMuMzItLjE0OC01MC4zMzguMjgtNjAuODUzLjcxLTE3LjUyMy0xLjEzOC0zMy41MjQtMTcuMTUzLTQ5LjU2OS0xNi4yOTYtMTYuMjk2LTM3Ljk5LTI0LjEzMy02MC4xODctMjQuMTMzLTEzLjA4NyAwLTI2LjYxOCAzLjE5NC0zOC43NDQgOC45NjEtMS45MjItMjQuNjM2LTEzLjAxMy00NS4xMDMtMjcuODc1LTU4LjA1Ny0yOS41NzYtMjUuNzYtNDcuMjE4LTI3LjE1LTc1LjI0LTI2Ljc2NlptNS4zMzggMTQwLjUzLTI5Ljk2IDc5LjI2MyAyOS4yOCAyMC4wMjIgMjguODA2LTc1LjQ5MnoiIHN0eWxlPSJvcGFjaXR5OjE7ZmlsbDojZWNlY2VjO2ZpbGwtb3BhY2l0eToxO3N0cm9rZTpub25lO3N0cm9rZS13aWR0aDoxNy41OTY5O3N0cm9rZS1saW5lY2FwOnJvdW5kO3N0cm9rZS1saW5lam9pbjpyb3VuZDtzdHJva2UtbWl0ZXJsaW1pdDo0O3N0cm9rZS1kYXNoYXJyYXk6bm9uZTtzdHJva2Utb3BhY2l0eToxIi8+Cjwvc3ZnPgo=" y="3" font-family="Verdana,Geneva,DejaVu Sans,sans-serif" text-anchor="middle" text-rendering="geometricPrecision"/>
```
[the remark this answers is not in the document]
asdil12: This is how OBS does it - and that' what inspired me.
Sure I can try.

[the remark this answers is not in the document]
Martchus: Well, I don't have a strong opinion on it. It just seems not so nice to encode plain text in base64 when we could simply use nested SVG elements.

lib/OpenQA/WebAPI/Controller/Test.pm
```diff
@@ -704,20 +745,28 @@ sub overview {
...
+    my $job = $self->_get_latest_job();
     return $self->render(text => 'No matching job found', status => 404) unless $job;
     $self->stash(testid => $job->id);
     return $self->_show($job);
 }
 
+sub latest_badge ($self) {
+    my $job = $self->_get_latest_job();
+    return $self->_badge($job);
+}
```
Martchus, line 768:
```suggestion
sub latest_badge ($self) { $self->_badge($self->_get_latest_job) }
```

lib/OpenQA/WebAPI/Controller/Test.pm
```diff
@@ -421,6 +440,28 @@ sub _show {
     $self->render('test/result');
 }
 
+sub badge ($self) { $self->_badge($self->_get_current_job(1)) }
+
+sub _badge ($self, $job) {
+    my $badge_text = "Error 404: Job not found!";
```
Martchus, line 446:
```suggestion
    my $badge_text = 'Error 404: Job not found!';
```

lib/OpenQA/WebAPI/Controller/Test.pm
```diff
@@ -20,6 +20,25 @@ use List::Util qw(min);
 
 use constant DEPENDENCY_DEBUG_INFO => $ENV{OPENQA_DEPENDENCY_DEBUG_INFO};
 
+
+# inspired by assets/stylesheets/openqa_theme.scss
+# some colors changed to improve readability of white text
+my %badge_result_colors = (
```
Martchus, line 26: Since it is a global variable it is the convention to use an uppercase name:
```suggestion
my %BADGE_RESULT_COLORS = (
```

Martchus (commented):
My previous remarks are all resolved, except the use of base64.

Martchus (approved):

kalikiana (approved):

Martchus:
This isn't merged automatically because the patch coverage is not 100 %. You're very close so I suppose it makes sense to extend the unit tests just a little bit further.

lib/OpenQA/WebAPI/Controller/Test.pm
```diff
@@ -421,6 +440,28 @@ sub _show {
...
+
+sub _badge ($self, $job) {
+    my $badge_text = 'Error 404: Job not found!';
+    my $badge_color = $BADGE_RESULT_COLORS{cancelled};
+    my $status = 404;
+
+    if ($job) {
+        my $result = ($job->result eq 'none') ? $job->state : $job->result;
+        if ($result eq 'scheduled' && $job->blocked_by_id) {
+            $result = 'blocked';
```
okurz, line 453: Those three lines look like the apply a result based on different factors but in one line you use the ternary operator and then the `if`. I would suggest to stay consistent with one approach. However I think the problem is actually that you are mixing states and results inconsistently with the definition from lib/OpenQA/Jobs/Constants.pm . By the way, another point is that you should use results from there and not the verbatim strings please.

@Martchus what is your take on the mixture for states and results?

okurz (changes requested):
I suggest to carefully check the existing uses of the word "badge". For example there is http://open.qa/docs/#_review_badges from which we should distinguish. I think the route names are fine but in the implementation and whenever a text is visible to users we should use a unique string. How about "test result badge"? But also it looks like you mix states and results.

To be able to help better I would need more explanation about the use cases you want to accomplish.

Please also extend documentation to mention the use. It can be simple, basically what you put into the commit message should be documentation and the commit message should explain your decisions and motivation and design choices.

lib/OpenQA/WebAPI/Controller/Test.pm
Martchus: > @Martchus what is your take on the mixture for states and results?

I guess for the sake of these "badges" I guess it is ok. If there's a result then the user likely cares about the result (and not the state which would only be the distinction between cancelled/done). If there's no result the user cares about the state. It is also fine to take the "blocked"-flag into account.

Maybe these three lines could be extracted into a separate function, though. I'd call it `sub concise_result`. By the way, I'd also change the coding style a little bit, use constants and avoid repeated calls to the same functions (because function calls are slow in Perl):
```
sub concise_result ($self) {
    my $res = $job->result;
    $res = $job->state if $res eq NONE;
    $res = BLOCKED if $res eq SCHEDULED && $job->blocked_by_id;
}
```

asdil12: The mix of state and result comes from the openQA code actually.
You can find it at several places, eg. assets/javascripts/test_result.js line 1007.

Martchus: I know, we already have something similar in JavaScript. It would still be nice to extract it into its own function. Maybe even add a comment that this is in line with some other JavaScript function (and vice versa).

asdil12: We could move this `concise_result` method to the job schema and call it from there.
Then we could remove the extra js code as well.

Martchus: I haven't looked at the JavaScript code yet. Maybe this could be computed server-side as well, maybe it is better to leave it client-side (e.g. when it is computed within a DataTables rendering hook).

Martchus (approved):

lib/OpenQA/WebAPI/Controller/Test.pm
```diff
@@ -20,6 +20,25 @@ use List::Util qw(min);
 
 use constant DEPENDENCY_DEBUG_INFO => $ENV{OPENQA_DEPENDENCY_DEBUG_INFO};
 
+
+# inspired by assets/stylesheets/openqa_theme.scss
+# some colors changed to improve readability of white text
+my %BADGE_RESULT_COLORS = (
```
okurz, line 26: Duplicating color definitions in scss and perl code looks like there should be a better way. What do others think?

okurz (changes requested):
One inline comment where maybe others have a good idea. And please look into my previous review comment, e.g. about documentation

lib/OpenQA/WebAPI/Controller/Test.pm
kalikiana: We agree. But we don't think there is. See https://github.com/os-autoinst/openQA/pull/5022#discussion_r1116904020

asdil12:
> And please look into my previous review comment, e.g. about documentation

I understood your comment about documentation in the way that you wanted to have a more extensive commit message.
So I added the line:

``` The badges can be used to show the current status of an openQA job eg. within a github PR.```

Is this not what you meant?

**Comment by Martchus** [comment]
@asdil12 No, he wants information like what you already had in the commit message to be added *in addition* to the documentation.

File: docs/UsersGuide.asciidoc [hunk 8]
```diff
@@ -540,6 +540,24 @@ image::images/job_group-limit_builds_tagged.png[Show only tagged or all builds]
 
 Related issue: https://progress.opensuse.org/issues/11052[#11052]
 
+
+=== Test result badges https://github.com/os-autoinst/openQA/pull/5022[gh#5022]
+
+For each job result including the latest job result page, there is a corresponding
+route to get an SVG status badge that can eg. be used to build a status dashboard
+or for the showing status within a github comment.
```
Martchus, line 548:
```suggestion
or for showing the status within a GitHub comment.
```

Martchus (changes requested):

Martchus (approved):

okurz (approved):

PR #5025, opened 2023-03-03
Improve error handling when automatically restarting a job

Martchus:
The web UI automatically restarts a job when it is set to done under certain conditions. So far errors when restarting are ignored. This change moves the restarting into a Minion job. An error will now lead to failing Minion jobs with the error message assigned as result. This should allow us to investigate https://progress.opensuse.org/issues/125276. Minion jobs are also retried a few times before ending up failed which should make it a bit more robust as well.

### DIFF
--- a/lib/OpenQA/Schema/Result/Jobs.pm
+++ b/lib/OpenQA/Schema/Result/Jobs.pm
@@ -1942,6 +1942,13 @@ sub handle_retry ($self) {
     return 1;
 }
 
+sub enqueue_restart ($self) {
+    return undef unless my $gru = eval { OpenQA::App->singleton->gru };    # gru might not be present within tests
+    my $openqa_job_id = $self->id;
+    my $minion_job_id = $gru->enqueue(restart_job => [$openqa_job_id])->{minion_id};
+    log_debug "Enqueued restarting openQA job $openqa_job_id via Minion job $minion_job_id";
+}
+
 =head2 done
 
 Finalize job by setting it as DONE.
@@ -2003,7 +2010,7 @@ sub done ($self, %args) {
     }
     $self->update(\%new_val);
     $self->unblock;
-    $self->auto_duplicate if $restart || ($self->is_ok_to_retry && $self->handle_retry);
+    $self->enqueue_restart if $restart || ($self->is_ok_to_retry && $self->handle_retry);
     # bugrefs are there to mark reasons of failure - the function checks itself though
     my $carried_over = $self->carry_over_bugrefs;
     $self->enqueue_finalize_job_results($carried_over);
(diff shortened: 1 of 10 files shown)


t/10-jobs.t
```diff
@@ -875,4 +880,41 @@ subtest '"race" between status updates and stale job detection' => sub {
 
 is $t->app->minion->jobs({states => ['failed']})->total, 0, 'No unexpected failed minion background jobs';
 
+subtest 'special cases when restarting job via Minion task' => sub {
+    $ENV{OPENQA_JOB_RESTART_ATTEMPTS} = 2;
+    $ENV{OPENQA_JOB_RESTART_DELAY} = 1;
```
kraih, line 885: I would always put a `local` in front, to avoid accidentally affecting later tests (when they are added in the future).

lib/OpenQA/Schema/Result/Jobs.pm
```diff
@@ -1942,6 +1942,13 @@ sub handle_retry ($self) {
     return 1;
 }
 
+sub enqueue_restart ($self) {
+    return undef unless my $gru = eval { OpenQA::App->singleton->gru };    # gru might not be present within tests
```
kraih, line 1946: This is rather ugly, does it affect that many tests that its better than a mock?

Martchus: I've actually just copied it from another function. Let's see what breaks without it.

lib/OpenQA/Task/Job/Restart.pm
```diff
@@ -0,0 +1,41 @@
+# Copyright 2021 SUSE LLC
```
okurz, line 1:
```suggestion
# Copyright SUSE LLC
```

okurz (changes requested):
Will this affect
* sending AMQP messages
* label carry-over
* triggering job_done events and things like auto-review jobs

Martchus:
It will affect none of them directly. However, these behaviors might now be executed when the job hasn't been restarted yet (but would be restarted by openQA itself). This might be a problem for the hook script, actually. If it e.g. would return early if the job has been restarted anyways (by openQA) then this would not always work anymore. I suppose it makes sense to delay the execution of hook scripts until the job has been restarted.

Martchus:
> I suppose it makes sense to delay the execution of hook scripts until the job has been restarted.

That should now be the case. Note that I didn't use Minion-level dependencies because we want to invoke the hook script even if restarting fails. So it became not very elegant. However, it is still simple enough: do the first restart as part of the finalizing task, spawn another Minion job in case a retry is required

kraih:
> Note that I didn't use Minion-level dependencies because we want to invoke the hook script even if restarting fails. So it became not very elegant. However, it is still simple enough: do the first restart as part of the finalizing task, spawn another Minion job in case a retry is required

Sounds like Minion dependencies with the `lax` option might be a good choice here. https://docs.mojolicious.org/Minion#enqueue1

kraih:
> When reading the docs, I hesitated because the feature was still experimental.

I've already implemented the feature in Minion.js too, it's not going away. 😁 I'll make it stable in the next Minion release.

perlpunk (approved):

okurz (approved):

PR #5027, opened 2023-03-07
Run cache.fullstack job also in nightly job

perlpunk:
To prevent empty cache errors that we get from time to time.

Issue: https://progress.opensuse.org/issues/123867

I ran the job on my fork, looked good

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -306,5 +306,11 @@ workflows:
               only:
                 - master
     jobs:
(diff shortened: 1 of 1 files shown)


okurz (changes requested):
I suggest you add a description in the workflow file describing why it's necessary to run that in nightly as well. Also IMHO a ticket reference in there would be nice

okurz (approved):

Martchus (approved):

PR #5030, opened 2023-03-09
Add bage for openQA in openQA test (appliance)

foursixnine:
Now that we have badges let's use them!

image


foursixnine:
@mergify squash title+body

Martchus (approved):
This reveals one detail that could still be improved: Currently the badge shows "uploading". I suppose it would be better to show the result of the last job in state "done" instead. (Maybe there could still be an indication that a job is running right now.)

perlpunk (approved):

foursixnine:
> This reveals one detail that could still be improved: Currently the badge shows "uploading". I suppose it would be better to show the result of the last job in state "done" instead. (Maybe there could still be an indication that a job is running right now.)

I think also GH might be doing some extra caching as @asdil12 was mentioning today in some other contex

PR #5036, opened 2023-03-14
Allow scheduling and monitoring a product via `openqa-cli`

Martchus:
Related ticket: https://progress.opensuse.org/issues/125720

---

Example invocation:
```
script/openqa-cli schedule --apikey … --apisecret … --host … DISTRI=example VERSION=0 FLAVOR=DVD ARCH=x86_64 TEST=simple_boot FOO=bar --param-file SCENARIO_DEFINITIONS_YAML=/hdd/openqa-devel/openqa/share/tests/example/scenaio-definitions.yaml -r
```

So one can use CLI-flags similar to `openqa-cli api`.

Still a draft because unit tests are missing and the error handling needs to be improved (it should retry like `openqa-cli api`).

### DIFF
--- a/lib/OpenQA/CLI/schedule.pm
+++ b/lib/OpenQA/CLI/schedule.pm
@@ -14,8 +14,8 @@ has post_url => sub { shift->url_for('isos') };
 
 sub _create_jobs ($self, $client, $args, $param_file, $job_ids) {
     my $params = $self->parse_params($args, $param_file);
-    my $tx = $client->start($client->build_tx(POST => $self->post_url, {}, form => $params));
-    my $res = $self->handle_result($tx, {pretty => 0, quiet => 1, links => 0, verbose => 0});
+    my $tx = $client->build_tx(POST => $self->post_url, {}, form => $params);
+    my $res = $self->retry_tx($client, $tx, {pretty => 0, quiet => 1, links => 0, verbose => 0});
     return $res if $res != 0;
     my $json = $tx->res->json;
     push @$job_ids, $json->{id} if defined $json->{id} && ref $json->{id} eq '';
@@ -28,8 +28,8 @@ sub _create_jobs ($self, $client, $args, $param_file, $job_ids) {
 sub _monitor_jobs ($self, $client, $poll_interval, $job_ids, $job_results) {
     while (@$job_results < @$job_ids) {
         my $job_id = $job_ids->[@$job_results];
-        my $tx = $client->start($client->build_tx(GET => $self->url_for("experimental/jobs/$job_id/status"), {}));
-        my $res = $self->handle_result($tx, {pretty => 0, quiet => 1, links => 0, verbose => 0});
+        my $tx = $client->build_tx(GET => $self->url_for("experimental/jobs/$job_id/status"), {});
+        my $res = $self->retry_tx($client, $tx, {pretty => 0, quiet => 1, links => 0, verbose => 0});
         return $res if $res != 0;
         my $job = $tx->res->json;
         my $job_state = $job->{state} // NONE;
--- a/lib/OpenQA/CLI/schedule.pm
+++ b/lib/OpenQA/CLI/schedule.pm
@@ -1,0 +1,97 @@
+# Copyright SUSE LLC
+# SPDX-License-Identifier: GPL-2.0-or-later
+
+package OpenQA::CLI::schedule;
+use OpenQA::Jobs::Constants;
+use Mojo::Base 'OpenQA::Command', -signatures;
+use Mojo::Util qw(getopt);
+use Term::ANSIColor qw(colored);
+use open qw(:std :encoding(UTF-8));
+
+has description => 'Schedules a set of jobs (via "isos post" creating a schedule product)';
+has usage => sub { shift->extract_usage };
+has post_url => sub { shift->url_for('isos') };
+
+sub _create_jobs ($self, $client, $args, $param_file, $job_ids) {
+    my $params = $self->parse_params($args, $param_file);
+    my $tx = $client->start($client->build_tx(POST => $self->post_url, {}, form => $params));
+    my $res = $self->handle_result($tx, {pretty => 0, quiet => 1, links => 0, verbose => 0});
+    return $res if $res != 0;
+    my $json = $tx->res->json;
+    push @$job_ids, $json->{id} if defined $json->{id} && ref $json->{id} eq '';
+    push @$job_ids, @{$json->{ids}} if ref $json->{ids} eq 'ARRAY';
+    return 0 unless my $error = $json->{error};
+    print STDERR colored(['red'], $error, "\n");
+    return 1;
+}
+
+sub _monitor_jobs ($self, $client, $poll_interval, $job_ids, $job_results) {
+    while (@$job_results < @$job_ids) {
+        my $job_id = $job_ids->[@$job_results];
+        my $tx = $client->start($client->build_tx(GET => $self->url_for("experimental/jobs/$job_id/status"), {}));
+        my $res = $self->handle_result($tx, {pretty => 0, quiet => 1, links => 0, verbose => 0});
+        return $res if $res != 0;
+        my $job = $tx->res->json;
+        my $job_state = $job->{state} // NONE;
+        if (OpenQA::Jobs::Constants::meta_state($job_state) eq OpenQA::Jobs::Constants::FINAL) {
+            push @$job_results, $job->{result} // NONE;
+            next;
+        }
+        print "Job state of job ID $job_id: $job_state, waiting…\n";
+        sleep $poll_interval;
+    }
+}
+
+sub _compute_return_code ($self, $job_results) {
+    for my $job_result (@$job_results) {
+        return 2 unless OpenQA::Jobs::Constants::is_ok_result($job_result);
+    }
+    return 0;
+}
+
+sub command ($self, @args) {
+    die $self->usage
+      unless getopt \@args,
+      'param-file=s' => \my @param_file,
+      'm|monitor' => \my $monitor,
+      'i|poll-interval=i' => \my $poll_interval,
+      'r|set-return-code' => \my $set_return_code;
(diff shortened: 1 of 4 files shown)


lib/OpenQA/CLI/schedule.pm
```diff
@@ -0,0 +1,97 @@
...
+    my $json = $tx->res->json;
+    push @$job_ids, $json->{id} if defined $json->{id} && ref $json->{id} eq '';
+    push @$job_ids, @{$json->{ids}} if ref $json->{ids} eq 'ARRAY';
+    return 0 unless my $error = $json->{error};
+    print STDERR colored(['red'], $error, "\n");
+    return 1;
+}
+
+sub _monitor_jobs ($self, $client, $poll_interval, $job_ids, $job_results) {
+    while (scalar @$job_results < scalar @$job_ids) {
```
perlpunk, line 29:
```suggestion
    while (@$job_results < @$job_ids) {
```

lib/OpenQA/CLI/schedule.pm
```diff
@@ -0,0 +1,97 @@
...
+        my $job_id = $job_ids->[scalar @$job_results];
```
perlpunk, line 30:
```suggestion
        my $job_id = $job_ids->[@$job_results];
```

lib/OpenQA/CLI/schedule.pm
```diff
@@ -0,0 +1,97 @@
...
+        my $tx = $client->start($client->build_tx(GET => $self->url_for("jobs/$job_id"), {}));
```
perlpunk, line 31: We can actually use this cheaper route:
```suggestion
        my $tx = $client->start($client->build_tx(GET => $self->url_for("experimental/jobs/$job_id/status"), {}));
```

lib/OpenQA/CLI/schedule.pm
```diff
@@ -0,0 +1,97 @@
+# Copyright 2023 SUSE LLC
```
okurz, line 1:
```suggestion
# Copyright SUSE LLC
```

lib/OpenQA/CLI/schedule.pm
```diff
@@ -0,0 +1,97 @@
...
+  Usage: openqa-cli schedule [OPTIONS] DISTRI=… VERSION=… FLAVOR=… ARCH=… [ISO=… …]
+
+  Options:
+        --apibase <path>           API base, defaults to /api/v1
+        --apikey <key>             API key
+        --apisecret <secret>       API secret
+        --host <host>              Target host, defaults to http://localhost
+    -h, --help                     Show this summary of available options
+        --osd                      Set target host to http://openqa.suse.de
+        --o3                       Set target host to https://openqa.opensuse.org
```
okurz, line 87: Do we really need to repeat all those parameters?

lib/OpenQA/CLI/schedule.pm
```diff
@@ -0,0 +1,97 @@
...
+        --param-file <param=file>  Load content of params from files instead of
+                                   from command line arguments. Multiple params
+                                   may be specified by adding the option
+                                   multiple times
+    -m, --monitor                  Wait until all jobs are done/cancelled
+    -i, --poll-interval            Specifies the poll interval used with --monitor
+    -r, --set-return-code          Return non-zero exit code if at least one job
+                                   is not passed/softfailed (implies -m)
```
okurz, line 95: Why not make that the default for `-m` anyway?

lib/OpenQA/CLI/schedule.pm
```diff
@@ -0,0 +1,97 @@
...
+    return $self->_compute_return_code(\@job_results);
+}
+
+1;
+
+=encoding utf8
+
+=head1 SYNOPSIS
+
+  Usage: openqa-cli schedule [OPTIONS] DISTRI=… VERSION=… FLAVOR=… ARCH=… [ISO=… …]
```
okurz, line 78: does it still work to specify parameters in lower-case? Then I would prefer that form or at least mention it as a convenience option.

okurz (changes requested):
This looks good!

@grisu48 as the author of openqa-mon, what do you think about this approach? Is it a good complementary solution, something completely different or duplicating behaviour?

lib/OpenQA/CLI/schedule.pm
Martchus: I don't like scripts that do too much at the same time, at least not by default.

It should also be optional so this sub command is usable in more general way. However, I prefer positive arguments over negative ones which speaks also for not making it a default.

It is also an unintuitive default for a client to return a non-zero code if the request generally worked. (If I request a page with the contents "failure" I don't expect curl to return a non-zero code for this.)

lib/OpenQA/CLI/schedule.pm
Martchus: I was asking myself the same question. At least it doesn't look like the Mojolicous command system would allows us to avoid it. I could of course ditch using it (and this weird way of making use of a code-comment that this way becomes more than just a comment after all) and just use a function (using a bunch of string literals) to print this. @kraih What's your take on this?

kraih: Yes, there is currently no way to avoid it.

lib/OpenQA/CLI/schedule.pm
kraih: We could have a flag that auto-uppercases all parameters.

lib/OpenQA/CLI/schedule.pm
okurz: k

lib/OpenQA/CLI/schedule.pm
okurz: yeah but for `monitor` this is something that could be expected and matches our use case if the exit code shows the result of monitored jobs. If you want to avoid negations then maybe cli should return 2 on request errors, not 1 :)

lib/OpenQA/CLI/schedule.pm
okurz: ok, nevermind for now. It's something unrelated to support lowercase arguments automatically turned into uppercase. You are welcome to discuss my proposal in https://github.com/os-autoinst/openQA/pull/5038

lib/OpenQA/CLI/schedule.pm
kalikiana: Can you clarify the use cases of "-r" versus "-rm"? I feel like I would always want the return code. Not sure when it's useful not to have 🤔

kalikiana (approved):

lib/OpenQA/Command.pm
```diff
@@ -107,4 +108,16 @@ sub url_for ($self, $path) {
     return Mojo::URL->new($self->apibase . $path)->to_abs(Mojo::URL->new($self->host));
 }
 
+sub retry_tx ($self, $client, $tx, $handle_args, $retries = undef, $delay = undef) {
+    $delay //= $ENV{OPENQA_CLI_RETRY_SLEEP_TIME_S} // 3;
+    $retries //= $ENV{OPENQA_CLI_RETRIES} // 0;
+    for (;; --$retries) {
```
kraih, line 114: That would probably be easier to read as a while loop.

lib/OpenQA/Command.pm
```diff
@@ -11,6 +11,7 @@ use Mojo::Util qw(decode getopt);
 use Mojo::URL;
 use Mojo::File qw(path);
 use Term::ANSIColor qw(colored);
+use open qw(:std :encoding(UTF-8));
```
kraih, line 14: This will probably cause unintended side effects. In my experience it's better to keep STDIN/STDOUT in binary mode and manually encode/decode when necessary on a case by case basis. Even just printing JSON to STDOUT would now require you to decode it before printing to avoid double encoding issues.

lib/OpenQA/CLI/schedule.pm
```diff
@@ -0,0 +1,97 @@
+# Copyright SUSE LLC
+# SPDX-License-Identifier: GPL-2.0-or-later
+
+package OpenQA::CLI::schedule;
+use OpenQA::Jobs::Constants;
+use Mojo::Base 'OpenQA::Command', -signatures;
+use Mojo::Util qw(getopt);
+use Term::ANSIColor qw(colored);
+use open qw(:std :encoding(UTF-8));
```
kraih, line 9: Same here basically. If you depend on shared functionality with `openqa-cli api` you should maintain the same environment and not change PerlIO globals. Stick with binary mode STDOUT/STDERR.

lib/OpenQA/Command.pm
kraih: I have no strong feelings about this though, if nobody else cares just keep the fancy for loop.

grisu48:
> This looks good!
> 
> @grisu48 as the author of openqa-mon, what do you think about this approach? Is it a good complementary solution, something completely different or duplicating behaviour?

No concerns by me. `openqa-mon` does more than just waiting and having this as complementary option directly in `openqa-cli` is a nice-to-have.

t/43-cli-schedule.t
```diff
@@ -0,0 +1,83 @@
...
+use OpenQA::CLI;
+use OpenQA::CLI::schedule;
+use OpenQA::Jobs::Constants;
+use OpenQA::Test::Case;
+use Mojo::Server::Daemon;
+use Mojo::File qw(tempdir tempfile);
+use Test::Output qw(combined_like);
+use Test::MockModule;
+
+my ($archive, $res) = OpenQA::CLI::schedule->new;
```
kraih, line 21: Is this a typo? A call to `->new` should never return more than one result. 🤔

kraih: And the name `$archive` is inherited from the other command? Probably wanto to name it `$schedule` here.

kraih: It managed to confuse me. 😁

kraih (approved):

PR #5038, opened 2023-03-16
CLI: Support lower-case test argument passing

okurz:
Motivation:
isotovideo from os-autoinst automatically transforms lower-case test
parameter command line arguments into upper-case test arguments. openQA
so far only checked the validity of upper-case parameters in a regex,
failing if lower-case was used.

Design Choices:
We optionally transform lower-case test parameter keys into upper-case
test arguments during `openqa-cli schedule`. The uppercasing applies
only to parameter names (keys), not values, and skips special API keys
like `async` and `scheduled_product_clone_id` which have been
deduplicated. Other commands like `openqa-cli api` continue to
preserve precise casing.

User Benefits:
Users can now conveniently pass lower-case test arguments via
`openqa-cli schedule` without encountering errors, aligning the
experience with `isotovideo`.

### DIFF
--- a/t/43-cli-schedule.t
+++ b/t/43-cli-schedule.t
@@ -12,6 +12,7 @@ use OpenQA::Test::TimeLimit '7';
 use OpenQA::CLI;
 use OpenQA::CLI::monitor;
 use OpenQA::CLI::schedule;
+use OpenQA::CLI::api;
 use OpenQA::Jobs::Constants;
 use OpenQA::Test::Case;
 use OpenQA::Test::Utils qw(perform_minion_jobs);
@@ -28,7 +29,7 @@ my $schedule = OpenQA::CLI::schedule->new;
 
 # change API to simulate job state/result changes
 my $job_controller_mock = Test::MockModule->new('OpenQA::WebAPI::Controller::API::V1::Job');
-my @job_mock_results = (PASSED, SOFTFAILED, PASSED, USER_CANCELLED, PASSED, SOFTFAILED);
+my @job_mock_results = (PASSED, SOFTFAILED, PASSED, USER_CANCELLED, PASSED, SOFTFAILED, PASSED, SOFTFAILED);
 $job_controller_mock->redefine(
     get_status => sub ($self) {
         # reply as usual
@@ -72,6 +73,7 @@ my @scenarios = ('--param-file', "SCENARIO_DEFINITIONS_YAML=$FindBin::Bin/data/0
 my @settings1 = (qw(DISTRI=example VERSION=0 FLAVOR=DVD ARCH=x86_64 TEST=simple_boot));
 my @settings2 = (qw(DISTRI=opensuse VERSION=13.1 FLAVOR=DVD ARCH=i586 BUILD=0091 TEST=autoyast_btrfs));
 my @settings3 = (@settings2, qw(async=1));
+my @settings4 = (qw(distri=opensuse version=13.1 flavor=DVD arch=i586 build=0091 test=autoyast_btrfs));
 
 subtest 'running into error reply' => sub {
     my $res;
@@ -106,6 +108,11 @@ subtest 'scheduling and monitoring set of two jobs' => sub {
     qr|"successful_job_ids":\[\d+,\d+\].*passed.*softfailed|s, 'response logged if async=1 flag was used';
     is $res, 0, 'zero return-code if all jobs are ok with async=1 flag';
 
+    combined_like { $res = $schedule->run(@options, @scenarios, @settings4) }
+    qr|2 jobs have been created.*(http://127.0.0.1.*/tests/\d+.*){2}passed.*softfailed|s,
+      'response logged if jobs created correctly from lower-case arguments';
+    is $res, 0, 'zero return-code if all jobs are ok with lower-case arguments';
+
     $fake_scheduled_product_status = OpenQA::Schema::Result::ScheduledProducts::CANCELLED;
     combined_like { $res = $schedule->run(@options, @scenarios, @settings3) }
     qr|Scheduled product \d+ ended up cancelled|s,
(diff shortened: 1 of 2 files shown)


Martchus (changes requested):
Wouldn't this upper-case any kind of parameter (and not only ones ending up job settings)? Then it would be very wrong.

okurz:
> Wouldn't this upper-case any kind of parameter (and not only ones ending up job settings)? Then it would be very wrong.

Yes, I agree :) I haven't checked that. Tests should tell but apparently my working copy of openQA is outdated.

@mergifyio rebase

perlpunk:
> Wouldn't this upper-case any kind of parameter (and not only ones ending up job settings)? Then it would be very wrong.

Yeah, and it would also uppercase the value of the setting, e.g. `distri=Tumbleweed` => `DISTRI=TUMBLEWEED`

okurz:
just rebased, not more

okurz:
> Wouldn't this upper-case any kind of parameter (and not only ones ending up job settings)? Then it would be very wrong.

Fixed this by excluding special parameters

> Yeah, and it would also uppercase the value of the setting, e.g. `distri=Tumbleweed` => `DISTRI=TUMBLEWEED`

Fixed by only upcasing the key

lib/OpenQA/Command.pm
```diff
@@ -16,6 +16,7 @@ use Term::ANSIColor qw(colored);
 my $JSON = Cpanel::JSON::XS->new->utf8->canonical->allow_nonref->allow_unknown->allow_blessed->convert_blessed
   ->stringify_infnan->escape_slash->allow_dupkeys->pretty;
 my $PARAM_RE = qr/^([[:alnum:]_\[\]\.\:]+)=(.*)$/s;
+my $RESERVED_API_KEYS_RE = qr/^(async|scheduled_product_clone_id)$/;
```
Martchus: Is this really an exhaustive list? How will we keep this in-sync with reality? Note that tests are unlikely to fail because you've implemented the upper-casing within `openqa-cli` which we don't involve in tests of API routes.

lib/OpenQA/Command.pm
```diff
@@ -79,16 +80,20 @@ sub parse_headers ($self, @headers) {
     return {map { /^\s*([^:]+)\s*:\s*(.*+)$/ ? ($1, $2) : () } @headers};
 }
 
-sub parse_params ($self, $args, $param_file) {
+sub parse_params ($self, $args, $param_file, $uc_keys = 0) {
```
Martchus: For what routes is still called? Note that there are many routes that require lower-case parameters, e.g. `distri` (and not `DISTRI`). So this look still very wrong.

Note that there's also a way of specifying parameters via a JSON file. This alternative approach doesn't seem to be altered which means both are now inconsistent.

t/43-cli-schedule.t
```diff
@@ -106,6 +107,11 @@ subtest 'scheduling and monitoring set of two jobs' => sub {
     qr|"successful_job_ids":\[\d+,\d+\].*passed.*softfailed|s, 'response logged if async=1 flag was used';
     is $res, 0, 'zero return-code if all jobs are ok with async=1 flag';
 
+    combined_like { $res = $schedule->run(@options, @scenarios, @settings4) }
+    qr|2 jobs have been created.*(http://127.0.0.1.*/tests/\d+.*){2}passed.*softfailed|s,
+      'response logged if jobs created correctly from lower-case arguments';
+    is $res, 0, 'zero return-code if all jobs are ok with lower-case arguments';
```
Martchus, line 114: This definitely also needs the opposite test: Routes that expect lower-case keys still work.

Martchus (dismissed):
I would implement this in the controller code where we handle such API requests instead of making the case-insensitive behavior client-specific. This would avoid the problems mentioned in inline comments.

lib/OpenQA/Command.pm
okurz: I now moved the functionality into lib/OpenQA/WebAPI/Controller/API/V1/Iso.pm `create` where we already called `uc`. There we know that this list actually *is* exhaustive as no other parameters had been extracted before calling `uc`.

lib/OpenQA/Command.pm
okurz: I now moved the functionality into lib/OpenQA/WebAPI/Controller/API/V1/Iso.pm `create` where we already called `uc` which should means that we handle the transformation more low level so should ensure consistent behaviour.

t/43-cli-schedule.t
okurz: done

okurz:
incorporated review suggestions from @Martchus

t/43-cli-schedule.t
```diff
@@ -132,4 +139,16 @@ subtest 'monitor jobs as a separate command' => sub {
...
+    my $api = OpenQA::CLI::api->new;
+    my @api_options = ('--apikey', 'ARTHURKEY01', '--apisecret', 'EXCALIBUR', '--host', $host);
+    combined_like {
+        $res = $api->run(@api_options, 'isos/job_stats', 'distri=example', 'version=0', 'flavor=DVD', 'arch=x86_64',
+            'build=123')
+    }
+    qr/\{\}/s, 'response logged successfully because lower-case arguments were preserved';
+    is $res, 0, 'zero return-code for successful api call with lower-case arguments';
+};
+
```
Martchus, line 153: Considering you're now not touching `Command.pm` anymore (except for adding a blank line), this test is probably not needed anymore. Of course it can't hurt so you can also keep it.

lib/OpenQA/Command.pm
```diff
@@ -83,6 +83,7 @@ sub parse_headers ($self, @headers) {
 
 sub parse_params ($self, $args, $param_file) {
     my %params;
+
```
Martchus:
```suggestion
```

okurz: done

t/43-cli-schedule.t
okurz: true. I would prefer to keep it as it nicely showcases how one can use the CLI

okurz:
incorporated the suggestion from @Martchus and rebased

Martchus (approved):

d3flex (approved):

PR #5039, opened 2023-03-16
Evaluate comment contents on carry-over

Martchus:
"Evaluate" means that

* the entry for carried-over bugrefs in the bugs table will be refreshed
* carried-over force-result labels are applied

This approach is mentioned in point 3.2 of
https://progress.opensuse.org/issues/123724#note-12 as an alternative to executing the hook script when the carry-over happens.


docs/UsersGuide.asciidoc
```diff
@@ -362,10 +362,10 @@ is `label:force_result:<new_result>[:<description>]`, for example
...
 operator permissions.
 
-NOTE: `force_result`-labels are not evaluated when when a comment is
-<<UsersGuide.asciidoc#carry-over,carried over>>. The carry over will only
-happen when the comment *also* contains a bug reference which is therefore not
-recommended.
+NOTE: `force_result`-labels are evaluated when when a comment is
+<<UsersGuide.asciidoc#carry-over,carried over>>. However, the carry over will
+only happen when the comment *also* contains a bug reference which is therefore
+not recommended.
```
perlpunk, line 368: Hm, I'm not sure how I should read that as a user.
so this feature is generally not recommended? But why?

lib/OpenQA/Schema/Result/Jobs.pm
```diff
@@ -1736,7 +1736,9 @@ sub carry_over_bugrefs ($self) {
           if $text !~ qr/The hook script will not be executed/ && defined $self->hook_script;
         $text .= "\n" unless substr($text, -1, 1) eq "\n";
         my %newone = (text => $text, user_id => $comment->user_id);
-        $self->comments->create_with_event(\%newone, {taken_over_from_job_id => $prev_id});
+        my $comment = $self->comments->create_with_event(\%newone, {taken_over_from_job_id => $prev_id});
+        eval { $comment->handle_special_contents };
+        log_info "Unable to evaluating contents of taken-over comment: $@" if $@;
```
perlpunk, line 1741:
```suggestion
        log_info "Unable to evaluate contents of taken-over comment: $@" if $@;
```

docs/UsersGuide.asciidoc
Martchus: I've dropped the "not recommended" part as it is now actually not the case anymore.

lib/OpenQA/Schema/Result/Comments.pm
```diff
@@ -169,4 +170,29 @@ sub extended_hash ($self) {
...
+    $self->_insert_bugs;
+    $self->_control_job_result($c);
+}
+
+sub _control_job_result ($self, $c) {
+    return undef unless my ($new_result, $description) = $self->force_result;
+    return undef unless $new_result;
+    die "Invalid result '$new_result' for force_result\n"
+      unless grep { $_ eq $new_result } OpenQA::Jobs::Constants::RESULTS;
+    die "force_result labels only allowed for operators\n" unless !$c || $c->is_operator;
```
perlpunk, line 183: I usually avoid `unless` with more than one condition, especially as the first one is a negation. How about:
```suggestion
    die "force_result labels only allowed for operators\n" if $c && !$c->is_operator;
```

perlpunk (approved):

kalikiana (approved):

PR #5041, opened 2023-03-17
t: Add a unit test for latest_jobs until argument

kalikiana:
The optional argument takes a time and is currently uncovered.

### DIFF
--- a/t/10-jobs.t
+++ b/t/10-jobs.t
@@ -16,6 +16,8 @@ use Mojo::Base -signatures;
 use autodie ':all';
 use Encode;
 use File::Copy;
+use Date::Format 'time2str';
+use Time::Seconds;
 use OpenQA::Jobs::Constants;
 use OpenQA::Test::Case;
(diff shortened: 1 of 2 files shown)


t/10-jobs.t
```diff
@@ -81,6 +83,10 @@ subtest 'latest jobs' => sub {
     # These are the later clones, they should appear
     ok(grep(/^99963$/, @ids), 'cloned jobs appear as latest job');
     ok(grep(/^99946$/, @ids), 'cloned jobs appear as latest job (2nd)');
+
+    # Jobs can optionally be filtered via "until"
+    @latest = $jobs->latest_jobs(time2str('%Y-%m-%d %H:%M:%S', time - 5 * ONE_MINUTE, 'UTC'));
```
kalikiana, line 88: I would think this can't work with the broken Jobs.pm code yet it seems to pass?

Martchus (changes requested):

PR #5043, opened 2023-03-17
Optimize database queries for next/previous

perlpunk:
Fetch dependencies and failed modules for all jobs in one go instead of doing up to 500-1000 * 3 queries.

The duration of that route went down locally from 1.25s to 0.2s for a job with more than 50 next/previous pages.

As we do this optimization already for the /tests/overview page I put some code into subroutines.

edit: Looking at osd logs, this route can take up to 20 seconds...


mimi1vx (approved):

Martchus (approved):
Looks good. I'm wondering whether one could do it in one big query but definitely an improvement.

perlpunk:
Somehow the codecov report was missing. I rebased.

PR #5044, opened 2023-03-19
Small change on line 343 and line 345.

debrajnath:
Added a few words for better context to help the reader understand more clearly.

### DIFF
--- a/docs/GettingStarted.asciidoc
+++ b/docs/GettingStarted.asciidoc
@@ -340,9 +340,9 @@ The personal configuration should be stored in a file
 the `client.conf`, i.e. sections for each machine, e.g. `localhost`.
 
 [[get-testing]]
-== Testing openSUSE or Fedora
+== Testing openQA on openSUSE or Fedora
 
-An easy way to start using openQA is to start testing openSUSE or Fedora as they
+An easy way to start using openQA is to start testing on openSUSE or Fedora as they
 have everything setup and prepared to ease the initial deployment. If you want
 to play deeper, you can configure the whole openQA manually from scratch, but
 this document should help you to get started faster.


docs/GettingStarted.asciidoc
```diff
@@ -340,9 +340,9 @@ The personal configuration should be stored in a file
 the `client.conf`, i.e. sections for each machine, e.g. `localhost`.
 
 [[get-testing]]
-== Testing openSUSE or Fedora
+== Testing openQA on openSUSE or Fedora
```
mimi1vx, line 343: This completely reverses meaning of title, 

This isn't about testing openQA on some distribution but about testing distro in openQA

mimi1vx (changes requested):

docs/GettingStarted.asciidoc
debrajnath: Oops! Sorry my fault. 😔

Martchus (changes requested):
Sorry, but I think the intended meaning is really to test openSUSE or Fedora themselves. The idea is to use the existing test distributions for testing those operating systems to get started with running *some* tests on your openQA setup. Maybe it would still make sense to be more explicit about this.

Note that we can also not accept your commits because of the commit message. Start with a verb in imperative and don't put a `.` at the end, e.g. "Improve wording in documentation for getting started".

docs/GettingStarted.asciidoc
kalikiana: If you find it helpful you could phrase it the other way around:

```suggestion
== Testing openSUSE or Fedora with openQA
```

docs/GettingStarted.asciidoc
```diff
@@ -340,9 +340,9 @@ The personal configuration should be stored in a file
...
 
-An easy way to start using openQA is to start testing openSUSE or Fedora as they
+An easy way to start using openQA is to start testing on openSUSE or Fedora as they
```
kalikiana, line 345:
```suggestion
An easy way to start using openQA is to run openSUSE or Fedora tests as they
```

PR #5047, opened 2023-03-21
Unify handling formatting args of openqa-cli sub cmds

Martchus:
Just returning a non-zero exit code is not a good default as I
have noticed when working on
https://progress.opensuse.org/issues/125723. So the first
commit changes `openqa-cli schedule` to no longer assume
`--quiet` by default.

The formatting options are used by the `handle_result` function
which is defined in the command-base-class. So it makes sense to
also parse them there. This way they are automatically supported
by all sub-commands (similar to `--host`, `--apibase` and so on).

Related ticket: https://progress.opensuse.org/issues/125723


lib/OpenQA/Command.pm
```diff
@@ -19,6 +19,7 @@ my $PARAM_RE = qr/^([[:alnum:]_\[\]\.\:]+)=(.*)$/s;
 has apibase => '/api/v1';
 has [qw(apikey apisecret host)];
 has host => 'http://localhost';
+has options => undef;
```
kraih, line 22: I would just assign the defaults here, and use `my $options = $self->options;` in the beginning of the `run` method.

kraih: Also avoids the `$self->options(\%options)` later in the `run` method.

kraih: Just make sure it's `has options => sub {{pretty => 0, quiet => 0, links => 0, verbose => 0}};` with a wrapper function, or all instances of the class would share the same hashref, modifying each other's values. Which could get a bit annoying in tests. 😁

perlpunk (commented):
`$commit_message =~ s/quitet/quiet/`

Martchus:
Fixed the typo in the commit message.

perlpunk (approved):

okurz (approved):

okurz:
@mergifyio rebase

mergify waits for the codecov status check. The result comment is there but no check status. Let's try a rebase if that retriggers?

PR #5051, opened 2023-03-22
Add advice to timestamp mismatch error

mimi1vx:
https://progress.opensuse.org/issues/125372

### DIFF
--- a/lib/OpenQA/Shared/Controller/Auth.pm
+++ b/lib/OpenQA/Shared/Controller/Auth.pm
@@ -159,7 +159,7 @@ sub _key_auth ($self, $reason, $key) {
           $username, $self->tx->remote_address, $api_key->key, $api_key->secret;
         if (!_is_timestamp_valid($build_tx_timestamp, $timestamp)) {
             $log->debug("$reject_msg, $reason");
-            $reason = 'timestamp mismatch';
+            $reason = 'timestamp mismatch - check whether clocks on the local host and the web UI host are in sync';
         }
         elsif (_is_expired($api_key)) {
             $log->debug("$reject_msg, $reason");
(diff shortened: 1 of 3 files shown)


lib/OpenQA/Shared/Controller/Auth.pm
```diff
@@ -159,7 +159,7 @@ sub _key_auth ($self, $reason, $key) {
           $username, $self->tx->remote_address, $api_key->key, $api_key->secret;
         if (!_is_timestamp_valid($build_tx_timestamp, $timestamp)) {
             $log->debug("$reject_msg, $reason");
-            $reason = 'timestamp mismatch';
+            $reason = 'timestamp mismatch - check localtime vs webui';
```
Martchus, line 162: I would avoid the term "localtime" because one might think it refers to the time in your local timezone (as opposed to the timezone used on the web UI host). However, the timezone does not need to match between the hosts. The clocks just need to be generally in sync.

```suggestion
            $reason = 'timestamp mismatch - check whether clocks on the local host and the web UI host are in sync';
```

Martchus (changes requested):

kalikiana (approved):

perlpunk (approved):

Martchus:
The docker compose test failed due to an unrelated issue:

```
 Error: ing: perl-Params-ValidationCompiler-0.26-1.17.noarch.rpm [error]
Download (curl) error for 'http://download.opensuse.org/distribution/leap/15.4/repo/oss/noarch/perl-Params-ValidationCompiler-0.26-1.17.noarch.rpm':
Error code: Curl error 55
Error message: Connection died, tried 5 times before giving up

Abort, retry, ignore? [a/r/i/...? shows all options] (a): a
```

I can download the file just fine so maybe just a temporary networking issue.

PR #5052, opened 2023-03-22
Render web UI URLs for jobs created via `openqa-cli schedule`

Martchus:
* This makes it more convenient to view the jobs manually in the web UI.
* Related ticket: https://progress.opensuse.org/issues/125723

### DIFF
--- a/lib/OpenQA/CLI/schedule.pm
+++ b/lib/OpenQA/CLI/schedule.pm
@@ -19,6 +19,11 @@ sub _create_jobs ($self, $client, $args, $param_file, $job_ids) {
     my $json = $tx->res->json;
     push @$job_ids, $json->{id} if defined $json->{id} && ref $json->{id} eq '';
     push @$job_ids, @{$json->{ids}} if ref $json->{ids} eq 'ARRAY';
+    if (my $job_count = @$job_ids) {
+        my $host_url = Mojo::URL->new($self->host);
+        say $job_count == 1 ? '1 job has been created:' : "$job_count jobs have been created:";
+        say ' - ' . $host_url->clone->path("tests/$_") for @$job_ids;
+    }
     return 0 unless my $error = $json->{error};
     print STDERR colored(['red'], $error, "\n");
     return 1;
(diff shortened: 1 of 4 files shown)


kraih (approved):

okurz (changes requested):
We have that presentation for clone-job. Could you please check that we don't have duplication?

Also how about using the short form

```
say ' - ' . $host_url->clone->path("t$_") for @$job_ids;
```

Martchus:
I can check wether I can de-duplicate code. Note that I would prefer the long form. It is more clear/intuitive and faster as it avoids one redirection.

lib/OpenQA/CLI/schedule.pm
```diff
@@ -19,6 +19,11 @@ sub _create_jobs ($self, $client, $args, $param_file, $job_ids) {
     my $json = $tx->res->json;
     push @$job_ids, $json->{id} if defined $json->{id} && ref $json->{id} eq '';
     push @$job_ids, @{$json->{ids}} if ref $json->{ids} eq 'ARRAY';
+    if (my $job_count = @$job_ids) {
+        my $host_url = Mojo::URL->new($self->host);
+        say $job_count == 1 ? '1 job has been created:' : "$job_count jobs have been created:";
+        say ' - ' . $host_url->clone->path("tests/$_") for @$job_ids;
```
kalikiana, line 25: We may want this to [match the output style used when cloning](https://github.com/os-autoinst/openQA/blob/master/lib/OpenQA/Script/CloneJob.pm#L226)? That's probably what @okurz was referring to.
Currently `print "Created job #$cloned_job_id: $orig_job->{name} -> $base_url/t$cloned_job_id\n";` is how we print newly cloned jobs.

kalikiana: I'd lean towards using the style used here in both cases if we want that since it's less noisy.

kraih: I never liked the clone job output, this is better.

Martchus: But in this context we don't have the name. I wouldn't do an extra request here to print the name. I would also avoid the `/t` route as it is less clear. It is mainly meant for easier typing when entering a URL manually but not useful here.

kalikiana: Yes. We both already said we like yours better. It should just be consistent.

mimi1vx (approved):

PR #5054, opened 2023-03-23
Assign dependencies when cloning jobs with children but without parents

kraih:
Previously the clone job code only assigned dependencies when parents were cloned in addition to children. This patch changes the dependency assignment to work in both directions. The code also gets a little easier to follow by moving the `skip-chained-deps`/`skip-deps` handling into the main loop.

Progress: https://progress.opensuse.org/issues/124493

### DIFF
--- a/lib/OpenQA/Script/CloneJob.pm
+++ b/lib/OpenQA/Script/CloneJob.pm
@@ -200,17 +200,9 @@ sub openqa_baseurl ($local_url) {
     return $local_url->scheme . '://' . $local_url->host . $port;
 }
 
-sub get_deps ($job, $options, $job_type) {
-    my ($chained, $directly_chained, $parallel);
-    unless ($job_type eq 'parents' && $options->{'skip-deps'}) {
-        unless ($job_type eq 'parents' && $options->{'skip-chained-deps'}) {
-            $chained = $job->{$job_type}->{Chained};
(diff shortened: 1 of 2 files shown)


kraih:
This way of assigning dependencies based on if each individual job will actually be cloned also has the advantage that it should just work if we add new ways to filter out specific jobs. Like not cloning any children that contain the word `kernel`.

lib/OpenQA/Script/CloneJob.pm
```diff
@@ -253,21 +245,27 @@ sub clone_job ($jobid, $url_handler, $options, $post_params = {}, $jobs = {}, $d
...
                 next unless $clone_children || $dependencies == $parallel;
             }
+
             clone_job($_, $url_handler, $options, $post_params, $jobs, $depth + 1, $job_type) for @$dependencies;
         }
-        if ($job_type ne 'children') {
-            $settings->{_PARALLEL} = join(',', @$parallel) if @$parallel;
-            $settings->{_START_AFTER} = join(',', @$chained) if @$chained;
-            $settings->{_START_DIRECTLY_AFTER} = join(',', @$directly_chained) if @$directly_chained;
+        if ($job_type eq 'parents') {
```
Martchus, line 265: It is indeed a good idea to make this a positive condition.

Martchus (approved):

mimi1vx (approved):

PR #5056, opened 2023-03-26
Correct formula in _exceeds_limit when CACHE_MIN_FREE_PERCENTAGE is set

rclobus:


rclobus:
If the cache has not been filled yet, and a lot of free space is available, the old formula would return 1 due to the second if statement, causing every asset to be download again and the previous assets to be removed.

Martchus:
I suppose test cases need to be adapted.

It would also be a good idea to cross check with our initial intention behind the formula.

rclobus:
The scenario I had:
* An initial setup, so the cache was still empty (self->{cache_real_size} = 0)
* $self->limit = 50 (the default value of 50 GiB)
* available_bytes = 180 GiB
* needed = 0 (actually the ISO file was 3.5GiB, but that's for another bug report)
* total_bytes = 270 GiB
* CACHE_MIN_FREE_PERCENTAGE in workers.ini was set to 10
* An initial setup, so the cache was still empty

Martchus:
Ok, can you adapt/extend the test in `t/25-cache.t` then? (I get that you're exchanging the `+` with a `-` but not why you're exchanging the `>` with a `<`. Maybe it is still too early in the morning for me.)

rclobus:
I'll add an updated test case soon.

rclobus:
I've updated the test for `_exceeds_limit` extensively.
* The results of previous tests leaked into this test, so I've added `init`
* I've used a more asymmetrical value (8) for `min_free_percentage` to discern between `bavail` and the value for `exceeds_limit`

rclobus:
The test `25-cache.t` is passing the CI now, but fullstack-unstable has `33-developer_mode.t` failing, which does at first glance appears not to be related to my changes...

Martchus (approved):

PR #5057, opened 2023-03-27
Remove obsolete frontend code for sending commands to worker

Martchus:
Since the introduction of the developer mode the web UI does not send worker commands anymore. Note that the API could be removed in the backend as well but since it is strictly an API I would refrain from removing it to keep compatibility.


baierjan (approved):

mimi1vx:
mark deprecated api and after few months remove ?

PR #5058, opened 2023-03-27
Avoid live view getting stuck in a certain case

Martchus:
If the to be appended part of the received log content does not contain a newline, the code for discarding the first probably partial line can get stuck in an endless loop. This change fixes the loop condition to avoid this.


Martchus:
Why not

okurz:
> ~Why not~ One needed to copy the packages from the CI check manually to the host because there's no publishing/repository. So likely not worth the effort.

You can simply enable the publishing flag on the automatically generated OBS repositories and then either add the repo or download the RPM packages from there.

PR #5060, opened 2023-03-27
Fix enabling live stream/log when the backend is not up yet

Martchus:
The current implementation of enabling the live log does nothing at all when the backend hasn't been started yet. This is a problem if asset caching is enabled because then the backend startup is delayed until all assets have been cached and the frontend likely sends the command for enabling the live log before the backend has been started.

This change causes the worker to track the number of live log viewers regardless of whether the backend is running or not. After the backend has been initialized the current number of live log viewers is evaluated and handled accordingly (in the same way as changes in viewership are handled normally).

Related ticket: https://progress.opensuse.org/issues/113318

### DIFF
--- a/lib/OpenQA/Worker/Job.pm
+++ b/lib/OpenQA/Worker/Job.pm
@@ -24,6 +24,7 @@ use Mojo::File 'path';
 use Try::Tiny;
 use Scalar::Util 'looks_like_number';
 use File::Map 'map_file';
+use List::Util 'max';
 
 # define attributes for public properties
 has 'worker';
@@ -75,6 +76,7 @@ sub new {
     $self->{_id} = $job_info->{id};
     $self->{_info} = $job_info;
     $self->{_livelog_viewers} = 0;
+    $self->{_livelog_viewers_before} = 0;
     $self->{_autoinst_log_offset} = 0;
     $self->{_serial_log_offset} = 0;
     $self->{_serial_terminal_offset} = 0;
@@ -308,6 +310,7 @@ sub _handle_engine_startup ($self, $engine, $max_job_time) {
     # stop execution if timeout has been exceeded
     $self->_set_timeout($max_job_time, $engine);
 
+    $self->_add_livelog_viewers(0);
     $self->_set_status(running => {});
 }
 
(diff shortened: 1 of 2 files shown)


Martchus:
Let's see what unit tests this breaks. I've already tested this manually by inserting a delay in the job setup as mentioned in https://progress.opensuse.org/issues/113318#note-12.

lib/OpenQA/Worker/Job.pm
```diff
@@ -653,37 +655,23 @@ sub is_backend_running {
...
+sub start_livelog ($self) { $self->_set_livelog_viewers($self->livelog_viewers + 1) }
+sub stop_livelog ($self) { $self->_set_livelog_viewers(max(0, $self->livelog_viewers - 1)) }
+sub _set_livelog_viewers ($self, $livelog_viewers = undef) {
+    $self->{_livelog_viewers} = $livelog_viewers if defined $livelog_viewers;
     return undef unless $self->is_backend_running;
-
     my $pooldir = $self->worker->pool_directory;
-    my $livelog_viewers = $self->livelog_viewers + 1;
-    if ($livelog_viewers == 1) {
+    if ((!defined $livelog_viewers && $self->{_livelog_viewers} >= 1) || ($self->{_livelog_viewers} == 1)) {
```
kalikiana, line 665: I'm not sure I get what the semantics of an undefined number of viewers are. It's not starting or stopping... but it's somehow used when there's viewers before the live log is enabled? I would suggest to make whatever it does more explicit.

Martchus: Maybe I should add a comment about this. The `$livelog_viewers` variable is suppose to be `undef` if the current number of viewers should be re-evaluated. This is done once the backend is running.

Martchus: This code was even buggy when the number of viewers dropped from 2 to 1 (in which it would try to start the live log but that's not necessary). So I changed this function quite a bit. Now these conditions should make much more sense.

okurz (approved):

kalikiana (approved):

PR #5063, opened 2023-03-29
Make it easier to identify openqa-cli in logs

kraih:
As a result of our adventure today trying to track down the service unintentionally running a DoS attack against OSD. We should do this for all openQA clients to make them easy to identify in logs.


kraih:
> It would also be good to have an option to set a more specific useragent. E.g. when calling openqa-cli from openqa-investigate, it should have both. WDYT?

Good point, in the case of `openqa-cli` that would make a lot of sense.

Martchus (approved):

perlpunk:
We actually had a ticket for this :)
https://progress.opensuse.org/issues/97037

PR #5068, opened 2023-04-03
Update obs_rsync.js

Lokeshsri11:
There are a few bugs that I can identify.
First, the `controlToShow` parameter is not passed correctly to the error callback function of the `fetchValue` function. It should be passed as a separate parameter instead of being included in the callback function's parameter list.

Second, the `postAndRedrawElement` function doesn't handle the case where the `btn.dataset.geturl` value is not defined. It should check if the value exists before calling the `fetchValue` function.

Third, the `postAndRedirect` function doesn't check if the `redir` parameter is defined before redirecting to the URL.

And here also the way to optimize our code :
1.Use `let` and `const` instead of `var` to declare variables. `let` and `const` provide block-level scoping, which can help avoid variable hoisting and improve code readability.

2.Use arrow functions instead of anonymous functions to improve code readability and reduce typing.

3.Use object destructuring to simplify the code when accessing object properties.

4.Use `const` instead of `var` when declaring constant values.

5.Use template literals to concatenate strings instead of using the + operator.

6.Use default function parameters to simplify function calls and provide default values for parameter

### DIFF
--- a/assets/javascripts/obs_rsync.js
+++ b/assets/javascripts/obs_rsync.js
@@ -1,22 +1,19 @@
 function handleObsRsyncAjaxError(xhr, ajaxOptions, thrownError) {
-  var message = xhr.responseJSON.error;
-  if (!message) {
-    message = thrownError ? thrownError : 'no message';
-  }
-  addFlash('danger', 'Error: ' + message);
+  const message = xhr.responseJSON?.error ?? thrownError ?? 'no message';
+  addFlash('danger', `Error: ${message}`);
 }
 
 function fetchValue(url, element, controlToShow) {
   $.ajax({
-    url: url,
+    url,
     method: 'GET',
-    success: function (response) {
-      element.innerText = response.message;
+    success: (response) => {
+      element.innerText = response?.message ?? '';
       if (controlToShow) {
         $(controlToShow).show();
       }
     },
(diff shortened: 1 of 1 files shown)


Martchus (commented):
Not sure whether it is required, but also make sure you run `make tidy-js`. Otherwise the CI checks might complain.

I personally also usually use `let` and `const` and the other JavaScript features you've mentioned and gladly accept the PR. However, we usually don't enforce using these for contributions so that's why our JavaScript code which has been touched by many different people is at it is. This is likely not going to change any time soon unless one is willing to modernize it and enforce the style. (Some of our code also simply predates the introduction of certain modern JavaScript features.)

Lokeshsri11:
Wait @martchus i'll update my code and make it in  `tidy-js` format

Martchus (approved):

Martchus:
Yes, we only need to wait for one more approval from the team. Then I'd merge your changes squashing them into one commit.

kalikiana (approved):
Nice tweaks!

Lokeshsri11:
Thank you @Martchus @kalikiana for this PR review and I will continue to try and contribute more to this project.

Lokeshsri11:
Okay @Martchus , I will pay attention to all these things from now on :)

PR #5071, opened 2023-04-04
Track download count to supress alert if no downloads ongoing

Martchus:
* Return the download count in the same way as we currently the download rate (via the Minion's InfluxDB route)
* Increment the download count when starting one and decrement it when it finished (regardless of the result)
* Reset the download count to zero when starting `openqa-workercache run …` be start counting from scratch after an unclean exit. (still WIP)
* Related ticket: https://progress.opensuse.org/issues/127097

### DIFF
--- a/t/25-cache-service.t
+++ b/t/25-cache-service.t
@@ -481,8 +481,22 @@ subtest 'Test Minion Sync task' => sub {
 
 subtest 'Minion monitoring with InfluxDB' => sub {
     my $app = $t->app;
-    my $rate = $app->cache->metrics->{download_rate};
+    my $cache = $app->cache;
+    my $metrics = $cache->metrics;
+    my $rate = $metrics->{download_rate};
+    my $count = $metrics->{download_count};
     ok $rate > 0, 'download rate is higher than 0 bytes per second';
+
+    subtest 'helpers for tracking download count' => sub {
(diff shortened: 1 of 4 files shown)


t/25-cache-service.t
```diff
@@ -481,15 +481,30 @@ subtest 'Test Minion Sync task' => sub {
...
     my $app = $t->app;
-    my $rate = $app->cache->metrics->{download_rate};
+    my $cache = $app->cache;
+    my $metrics = $cache->metrics;
+    my $rate = $metrics->{download_rate};
+    my $count = $metrics->{download_count};
     ok $rate > 0, 'download rate is higher than 0 bytes per second';
 
+    subtest 'helpers for tracking download count' => sub {
+        my $check_count = sub { $count = ($metrics = $cache->metrics)->{download_count} };
```
kalikiana, line 491: It looks to me like *$count* is updated but never checked? And the following tests don't update it so they just check that it never changes?

Martchus: The variable `$count` is used in checks later on. And yes, not all intermediate values of `$count` are ever used. However, this is a quite compact way of writing it.

kalikiana (approved):

mimi1vx (approved):

PR #5072, opened 2023-04-04
Ensure final status update for cancelled jobs

perlpunk:
Issue: https://progress.opensuse.org/issues/126665

Without this, cancelled jobs would not get their final status update and upload of details-*.json, so the uploaded screenshots could not be viewed.

### DIFF
--- a/lib/OpenQA/Schema/Result/Jobs.pm
+++ b/lib/OpenQA/Schema/Result/Jobs.pm
@@ -1430,7 +1430,7 @@ sub update_status ($self, $status) {
     #       wins the race updating the job state.
     $self->discard_changes;
     my $state = $self->state;
-    return {result => 0} unless $state eq RUNNING || $state eq UPLOADING;
+    return {result => 0} unless $state eq RUNNING || $state eq UPLOADING || $state eq CANCELLED;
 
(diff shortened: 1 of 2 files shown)


kalikiana (approved):

perlpunk:
No codecov today? I'll force push to retrigger

lib/OpenQA/Schema/Result/Jobs.pm
```diff
@@ -1430,7 +1430,7 @@ sub update_status ($self, $status) {
     #       wins the race updating the job state.
     $self->discard_changes;
     my $state = $self->state;
-    return {result => 0} unless $state eq RUNNING || $state eq UPLOADING;
+    return {result => 0} unless $state eq RUNNING || $state eq UPLOADING || $state eq CANCELLED;
```
okurz, line 1433: this feels like we want to combine the three individual states into a named condition

Martchus: Supposedly
```suggestion
    return {result => 0} if $state eq DONE;
```
would work as well. Of course this then includes also scheduled and assigned and one definitely had to test whether this actually works but I guess it wouldn't be a problem.

PR #5073, opened 2023-04-05
Improved build_progressbar.html.ep

Lokeshsri11:
I think we should updated the variable assignment to use `$stash` instead of an empty string (`''`), assuming that `$stash` is an existing hash reference containing a `class` key
and I added quotes around the progress bar section names (`'passed'`, `'unfinished'`, etc.) to make the code more readable and less error-prone
and the last I removed the `%=`, as it is not needed for the `build_progress_bar_section()` function calls.
.
.
and yes if you asking about that what is the benefit of this improvized code so, in first point if we can't use variable assignment as a default value I think it makes the code more robust and also fail if the input data is incomplete and the progress bar section names are now enclosed in quotes, which makes the code more readable and less prone to errors and last the `%= build_progress_bar_section()` calls have been simplified by removing the unnecessary `%=`, which reduces clutter and improves the code's readability.

### DIFF
--- a/templates/webapi/main/build_progressbar.html.ep
+++ b/templates/webapi/main/build_progressbar.html.ep
@@ -1,9 +1,10 @@
-% my $class = stash '';
+# Assign a default value to $class if it's not defined
+my $class = $stash->{class} // '';
 
 <div class="progress build-dashboard <%= $class %>" title="<%= build_progress_bar_title($result) %>">
-    %= build_progress_bar_section(passed => $result->{passed}, $max_jobs, $params)
-    %= build_progress_bar_section(unfinished => $result->{unfinished}, $max_jobs, $params, 'progress-bar-striped')
-    %= build_progress_bar_section(softfailed => $result->{softfailed}, $max_jobs, $params)
-    %= build_progress_bar_section(failed => $result->{failed}, $max_jobs, $params)
-    %= build_progress_bar_section(skipped => $result->{skipped}, $max_jobs, $params)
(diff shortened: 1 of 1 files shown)


templates/webapi/main/build_progressbar.html.ep
```diff
@@ -1,9 +1,10 @@
-% my $class = stash '';
+# Assign a default value to $class if it's not defined
```
Martchus, line 1: This comment is just stating the obvious, I'd remove it.

templates/webapi/main/build_progressbar.html.ep
```diff
@@ -1,9 +1,10 @@
...
+my $class = $stash->{class} // '';
```
Martchus, line 2: Have you tested this at all? I don't think this is working without preceding `% `. As-is, this code would just be rendered literally.

Martchus (commented):
The commit message should start with a verb in imperative, as mentioned in the article I've shared before. Note that "improvized" means "created and performed spontaneously or without preparation". I think you mean "improved". That would be "Improve `build_progressbar.html.ep`" in imperative.

Lokeshsri11:
and @Martchus what about using `for` loop to avoid repetitive code is it possible

Martchus:
One of the bars has `progress-bar-striped` (which complicates using a loop) and it would likely also be a little bit more inefficient (likely neglectable but still seems a bit unnecassary).

PR #5075, opened 2023-04-12
Improve out-of-the-box experience when invoking web UI container

Martchus:
* Configure a default certificate in the container startup
  script (as a fallback)
* Start and configure the database inside the container unless
  the database has been configured by the user manually
* Improve documentation
* See https://progress.opensuse.org/issues/127091
* See https://progress.opensuse.org/issues/127139

### DIFF
--- a/container/webui/Dockerfile
+++ b/container/webui/Dockerfile
@@ -21,8 +21,7 @@ RUN zypper ar -p 95 -f http://download.opensuse.org/repositories/devel:openQA/15
     a2enmod rewrite && \
     a2enflag SSL
 
-COPY openqa-ssl.conf /etc/apache2/vhosts.d/openqa-ssl.conf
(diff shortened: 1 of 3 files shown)


Martchus:
Note that I've simply tested this by building the container image locally via `podman image build --tag openqa_webui_local_build .` and then using `openqa_webui_local_build:latest` instead of `registry.opensuse.org/devel/openqa/containers15.4/openqa_webui:latest`.

We have tests for the docker-compose setup which shouldn't be impacted by this.

I'm still getting the error `Can not create MOJO_TMPDIR : mkdir /var/lib/openqa/share/factory: File exists at /usr/share/openqa/script/../lib/OpenQA/Setup.pm line 337.` which I'm going to look into but this error was there before and shouldn't prevent use to merge this.

mimi1vx (approved):

kalikiana (approved):
Nice. I can confirm that it works now.

PR #5078, opened 2023-04-16
Rclobus/typos

rclobus:
Minor corrections to the manual

### DIFF
--- a/docs/Installing.asciidoc
+++ b/docs/Installing.asciidoc
@@ -312,8 +312,8 @@ a2enmod proxy_wstunnel
 a2enmod rewrite
 --------------------------------------------------------------------------------
 
-For a basic setup, you can copy *openqa.conf.template* to *openqa.conf* and modify the `ServerName` if required
-setting. This will direct all HTTP traffic to openQA.
+For a basic setup, you can copy *openqa.conf.template* to *openqa.conf* and modify the `ServerName` setting if
+required. This will direct all HTTP traffic to openQA.
 
 [source,sh]
 --------------------------------------------------------------------------------
@@ -527,8 +527,8 @@ dnf install openqa-worker
 
 To allow workers to access your instance, you need to log into openQA as
 operator and create a pair of API key and secret. Once you are logged in, in the
-top right corner, is the user menu, follow the link 'manage API keys'.  Click
-the 'create' button to generate `key` and `secret`. There is also a script
+top right corner, is the user menu, follow the link 'Manage API keys'.  Click
+the 'Create' button to generate `key` and `secret`. There is also a script
 available for creating an admin user and an API key+secret pair
 non-interactively, `/usr/share/openqa/script/create_admin`, which can be useful
 for scripted deployments of openQA. Copy and paste the key and secret into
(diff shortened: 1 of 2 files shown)


rclobus:
Please squash them into one commit when merging, the details are IMO not very important.

rclobus:
The changes have been verified in the HTML version of the asciidoc files.

docs/Installing.asciidoc
```diff
@@ -270,8 +270,7 @@ a2enmod proxy_wstunnel
 a2enmod rewrite
 --------------------------------------------------------------------------------
 
-For a basic setup, you can copy *openqa.conf.template* to *openqa.conf* and modify the `ServerName` if required
-setting. This will direct all HTTP traffic to openQA.
+For a basic setup, you can copy *openqa.conf.template* to *openqa.conf* and modify the `ServerName` setting if required. This will direct all HTTP traffic to openQA.
```
Martchus, line 273: It would be nice to keep the wrapping.

docs/Installing.asciidoc
```diff
@@ -797,8 +796,8 @@ directories prior registering with openQA instance:
 4. `/var/lib/openqa`
 5. fail if none of above is available
 
-Once worker registers to openQA instance it checks for available job and starts
-accepting websockets commands. Worker accepts jobs as they will come in, there
+Once a worker registers to an openQA instance it checks for available jobs and starts
+accepting websockets commands. Workers accept jobs as they will come in, there
 is no priority, or other ordering, support at the moment.
```
Martchus, line 801: This whole section is actually completely wrong. A worker does never check for available jobs. It is possible to prioritize jobs.

So this should be: "Once a worker registers to an openQA instance, a scheduled job (of matching worker class) will be assigned to it. Dependencies between jobs will be considered for ordering the job assignment. Generally, jobs with higher priority (lower priority value) will be assigned first." - Although I'm wondering whether we need to go into that much detail here anyways.

Martchus (commented):
It would be nice to start commit message still with a capital letter (*after* the  `documentation: ` prefix). Additionally, I'd prefer "Improve grammar" over just "grammar".

Martchus:
I've just seen you want us to squash the commits. I can do that; so forget the remarks about the commit message. I've only added "not-ready" to prevent mergify from merging this so I can squash the commits instead.

okurz:
Still the individual comments need to be addressed, right? At least I see a need for rewrapping.

docs/Installing.asciidoc
rclobus: I've simplified it to:
"Once a worker registers to an openQA instance, scheduled jobs (of matching
worker class) can be assigned to it. Dependencies between jobs will be
considered for ordering the job assignment."

docs/Installing.asciidoc
rclobus: Done.

rclobus:
I've squashed the commits. Wrapping has been applied to match the original line width. A simplified statement what happens when a worker is registered has been entered.

Martchus (approved):

okurz (approved):

PR #5079, opened 2023-04-17
docs: Add SLE installation instructions

okurz:
Related progress issue: https://progress.opensuse.org/issues/127034

### DIFF
--- a/docs/Installing.asciidoc
+++ b/docs/Installing.asciidoc
@@ -148,9 +148,10 @@ package.
 
 The easiest way to install openQA is from distribution packages.
 
-- For openSUSE, packages are available for Leap and Tumbleweed
-- For Fedora, packages are available in the official repositories for Fedora 23
-and later.
+- For SUSE Linux Enterprise (SLE), openSUSE Leap and Tumbleweed packages are
+  available.
+- For Fedora, packages are available in the official repositories for Fedora
+  23 and later.
 
 === Development version repository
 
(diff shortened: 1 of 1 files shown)


docs/Installing.asciidoc
```diff
@@ -128,7 +128,7 @@ package.
 
 The easiest way to install openQA is from distribution packages.
 
-- For openSUSE, packages are available for Leap and Tumbleweed
+- For SUSE Linux Enterprise (SLE)/openSUSE, packages are available for Leap and Tumbleweed and SLE
```
baierjan, line 131: I see some modules later in the text, so we are talking about SLE 15 and newer, right? But we do not list Leap versions either, so I guess this is ok.

baierjan (approved):

docs/Installing.asciidoc
okurz: yes, I want to avoid mentioning specific versions. There can always be version specific differences.

Martchus, line 131:
```suggestion
- For SUSE Linux Enterprise (SLE), openSUSE Leap and Tumbleweed packages are available.
```

okurz: done

mimi1vx (approved):

docs/Installing.asciidoc
okurz: I added one sentence

```
It is suggested to always use the latest release version of operating systems.
```

That's the best I could find trying to put a statement into the text. I would simply not state anything regarding "support installation".

docs/Installing.asciidoc
```diff
@@ -190,6 +191,11 @@ You can install the main openQA server package using these commands.
 # openSUSE
 zypper in openQA
 
+# SLE
+. /etc/os-release
+SUSEConnect -p sle-module-desktop-applications/$VERSION_ID/$CPU
```
Martchus, line 196: Likely we also need
```
SUSEConnect -p sle-module-development-tools/$VERSION_ID/$CPU
```
there. At least some Perl packages comeing from there are missing in @perlpunk 's VM otherwise. (In my container this repo was already there but supposedly this is not always the case.)

Martchus (changes requested):

docs/Installing.asciidoc
okurz: Yes, I confirmed. I had everything installed but had development tools dependencies fulfilled from an IBS internal repo. Double-checked that now.

Martchus (approved):

perlpunk (approved):

PR #4973, opened 2023-01-06
Fixing virtio console for users

pevik:
Ticket: https://progress.opensuse.org/issues/122935

See also: https://github.com/os-autoinst/os-autoinst/pull/2240, https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/16190

**Verification run**
HPC (hpc_BETA_mpich_mpi_*)
- http://quasar.suse.cz/tests/2058 http://quasar.suse.cz/tests/2059 http://quasar.suse.cz/tests/2056 http://quasar.suse.cz/tests/2057
Log files visible, e.g. 
- http://quasar.suse.cz/tests/2059/file/serial_terminal_user.txt (new file)
- http://quasar.suse.cz/tests/2059/file/serial_terminal.txt

wicked
- http://quasar.suse.cz/tests/2061 http://quasar.suse.cz/tests/2062 (Tumbleweed)
- http://quasar.suse.cz/tests/2068 http://quasar.suse.cz/tests/2069 (SLE15-SP5)

LTP
- http://quasar.suse.cz/tests/2065#downloads (install_ltp Tumbleweed)
- http://quasar.suse.cz/tests/2064#downloads (install_ltp SLE15-SP5)
- http://quasar.suse.cz/tests/2075#downloads http://quasar.suse.cz/tests/2074#downloads (LTP tests SLE15-SP5)
- http://quasar.suse.cz/tests/2072#downloads http://quasar.suse.cz/tests/2073#downloads (LTP tests Tumbleweed)

virtio console test 
- http://quasar.suse.cz/tests/2060 (SLE15-SP5) http://quasar.suse.cz/tests/2063 (Tumbleweed)
- http://quasar.suse.cz/tests/2063#downloads
- http://quasar.suse.cz/tests/2063/logfile?filename=serial_terminal_user.txt (new file)
- http://quasar.suse.cz/tests/2063/logfile?filename=serial_terminal.txt (very big file, loading slowly)

<s>NOTE: live console does not show virtio_console_user.log :(</s>
UPDATE: Fixed - I see `master-node00 login: bernhard` in HPC job http://quasar.suse.cz/tests/2144#live

Deployment: https://progress.opensuse.org/issues/122935#note-7

PR #5085, opened 2023-04-20
Fix non-deleted serial_terminal_user.txt

okurz:
When the file serial_terminal_user.txt was introduced it was overlooked
that we also need to specify it for deletion.

Thanks to Petr Vorel for pointing that out.


pevik:
NOTE: file was introduced in https://github.com/os-autoinst/openQA/pull/4973

foursixnine (approved):

pevik (approved):
LGTM

foursixnine:
@Martchus, @okurz shall we merge manually too? idk what's wrong with Mergify :/

Martchus:
It is not mergify but actually codecov (which doesn't trigger at all and thus the expected number of checks isn't reached).

okurz:
I suggest to not care about some PRs not immediately merged and rather put the effort into fixing the missing codecov checks. For that I created https://progress.opensuse.org/issues/128129

PR #5092, opened 2023-04-21
Add minimalistic "health" check route

okurz:
Related progress issue: https://progress.opensuse.org/issues/127052


kraih:
Seems reasonable, could use a test though.

okurz:
> Seems reasonable, could use a test though.

done

Martchus (approved):

perlpunk (approved):

PR #5095, opened 2023-04-24
Avoid race condition by waiting for the #text element

Martchus:
This is a follow-up of 3266e22545f3a03272995171ac489db55f9efb9d which did not cover all places where the race condition might occur. I'm still not certain why/where we need this workaround. However, we definitely saw the initial problem
(https://progress.opensuse.org/issues/121042) again multiple times (https://progress.opensuse.org/issues/128153) in one of the places that haven't been covered so far but not in the place where the workaround had been added. So it likely makes sense to extend its usage.

### DIFF
--- a/t/ui/15-comments.t
+++ b/t/ui/15-comments.t
@@ -438,6 +438,7 @@ subtest 'commenting on parent group overview' => sub {
     # add another comment to have an equal amount of comments in the parent group and in the job group
     # to unify further checks for pagination between the group overview pages
     # (the job group overview has one more so far because of subtest 'URL auto-replace')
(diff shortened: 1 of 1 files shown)


mimi1vx (approved):

kalikiana (approved):

okurz:
I manually crosschecked Devel::Cover output in https://app.circleci.com/pipelines/github/os-autoinst/openQA/11333/workflows/2ed99f12-e39f-4879-a98a-1be9c48b5dd0/jobs/105960 . And as the changes are only within the linear flow of program execution no change in coverage is expected, merging manually due to codecov problem https://progress.opensuse.org/issues/128129

PR #5096, opened 2023-04-24
Allow partial retrigger

Martchus:
This allows to partially retrigger a test and its children by specifying the test via `TEST`, setting `_SKIP_CHAINED_DEPS=1` to avoid pulling-in chained parents and setting
`_INCLUDE_CHILDREN=1` to pull-in child jobs. Only the last setting is actually a new one.

Related ticket: https://progress.opensuse.org/issues/124469

---

Still a draft because the test case is very limited so far. It *should* at least not break other use-cases because unless I overlooked something the behavior should stay the same if `_INCLUDE_CHILDREN=1` isn't specified. I'm not sure about the performance as well but I made it so additional checks are only performed if `_INCLUDE_CHILDREN=1` is specified.


lib/OpenQA/Schema/Result/ScheduledProducts.pm
```diff
@@ -816,26 +819,41 @@ sub _populate_wanted_jobs_for_test_arg ($args, $settings, $wanted) {
...
 }
 
-sub _populate_wanted_jobs_for_parent_dependencies ($jobs, $wanted, $skip_chained_deps) {
+sub _is_any_parent_wanted ($jobs, $parents, $wanted_list) {
+    for my $parent (@$parents) {
+        my $parent_test_machine = join('@', @$parent);
+        my @parents_job_template = grep { join('@', $_->{TEST}, $_->{MACHINE}) eq $parent_test_machine } @$jobs;
+        for my $parent_job_template (@parents_job_template) {
+            return 1 if $wanted_list->{join('@', $parent_job_template->{TEST}, $parent_job_template->{MACHINE})};
+        }
```
okurz, line 828: I wonder can this be a grep or something as well?

Martchus:
This doesn't work for nested children. I suppose I need to make the lookup for whether the parent is wanted recursive.

mdoucha:
> I think it will consider all job groups (unless you specify the group via `_GROUP` or `_GROUP_ID`). The dependency lookup remains unchanged. So it will take the machine into account as usual.

OK, does `_GROUP_ID=...` interact with `_INCLUDE_CHILDREN=1`? That is, if I limit the `isos post` to just one job group, can I still retrigger nested children in other job groups?

Martchus:
If you filter scheduling via `_GROUP_ID=…` only job templates of that group are considered. The `_INCLUDE_CHILDREN=1` setting will *not* pull in any children defined in job templates of other groups. After limiting the filter via `_GROUP_ID=…` you can of course re-trigger the scheduled product for another group (or even the same group if you wanted¹).

¹ Note that this is not creating "Clone of …" jobs like the restart button would. It will create a new scheduled product with jobs that are independent from the initially scheduled product.

mdoucha:
> After limiting the filter via `_GROUP_ID=…` you can of course re-trigger the scheduled product for another group (or even the same group if you wanted¹).

Yes, but then you lose dependencies between jobs in different groups and you'll have to run the retrigger on the other job group multiple times (once for every cross-group dependency). That's why I've asked for retrigger by job ID in the initial design discussion.

lib/OpenQA/Schema/Result/ScheduledProducts.pm
```diff
@@ -429,16 +430,17 @@ sub _schedule_iso {
...
+Return the "job reference" for the specified job settings. It is used internally as a key for the job in various
+hash maps. It is also used to refer to a job in dependency specifications.
 
 =back
 
 =cut
 
-sub _settings_key ($settings) {
-    my ($test, $machine) = ($settings->{TEST}, $settings->{MACHINE});
+sub _job_ref ($job_settings) {
```
kalikiana, line 442: This rename confused me a bit. Maybe not that important, but "key" is more like what I expect here. This doesn't "refer" to anything.

lib/OpenQA/Schema/Result/ScheduledProducts.pm
```diff
@@ -849,42 +838,54 @@ sub _merge_settings_uppercase ($source_settings, $destination_settings, $excepti
...
-          unless $skip_chained_deps;
-        push @parents, _parse_dep_variable($job->{PARALLEL_WITH}, $job);
+        my $chained_parents = !$skip_chained_deps || $include_children ? _chained_parents($job) : [];
+        my $parents = _parallel_parents($job);
+        push @$parents, @$chained_parents unless $skip_chained_deps;
+
+        # delete unwanted jobs unless the parent is wanted and we include children
+        my $unwanted = !$wanted->{_job_ref($job)};
+        splice @$jobs, $i, 1 and next
+          if $unwanted && (!$include_children || !_is_any_parent_wanted($jobs, $chained_parents, $wanted));
```
kalikiana, line 881: I'd find it slightly more readable as an actual if block

```suggestion
          if $unwanted && (!$include_children || !_is_any_parent_wanted($jobs, $chained_parents, $wanted)) {
            splice @$jobs, $i, 1
            next
          }
```

t/api/02-iso.t
```diff
@@ -831,11 +831,34 @@ subtest '_SKIP_CHAINED_DEPS prevents scheduling parent tests' => sub {
...
+      or diag explain \%created_jobs;
+
+    # schedule with _INCLULDE_CHILDREN as well; now also all nested chained children should be included
+    my %expected_jobs = map { ("child_test_$_" => 1) } 1 .. 7;
+    $res = schedule_iso($t,
+        {%iso, _GROUP => 'opensuse test', TEST => 'child_test_1', _SKIP_CHAINED_DEPS => 1, _INCLUDE_CHILDREN => 1});
+    is $res->json->{count}, 7, '7 jobs scheduled';
+    %created_jobs = map { $jobs->find($_)->settings_hash->{'TEST'} => 1 } @{$res->json->{ids}};
+    is_deeply \%created_jobs, \%expected_jobs, 'only parallel parent and (nested) children scheduled'
+      or diag explain \%created_jobs;
```
kalikiana, line 860: Should/ can we test the behavior with _GROUP? since this came up in discussion?

Martchus: I think it is not necessary. The `_GROUP` filter is applied at a different level (right on the beginning) and the special `_INCLUDE_CHILDREN=1` is not really interacting with it. It only processes the data that has already been filtered or not depending on the `_GROUP` setting. So it would never wrongly pull-in additional children that do not belong to the specified `_GROUP`.

mdoucha:
> If you filter scheduling via `_GROUP_ID=…` only job templates of that group are considered. The `_INCLUDE_CHILDREN=1` setting will _not_ pull in any children defined in job templates of other groups.

Thinking about this some more, child jobs from different job group depending on a job that does not have a unique name in the scheduled product would cause other issues. So either the target job has duplicates but no children, or it has children but no duplicates. In either case, the PR will work.

okurz (approved):

kalikiana (approved):

PR #5103, opened 2023-04-27
Enhance keyboard accessibility for filter and settings button

Lokeshsri11:
This commit adds improved keyboard accessibility for the filter and settings button in the #filter-panel element. The button now has the appropriate role, tabindex, and aria-expanded attributes to ensure it can be accessed and operated using the keyboard. Additionally, a keydown event listener has been implemented to allow expanding/collapsing the button using the "Enter" key. These changes enhance the accessibility of the button for keyboard users and screen reader users, making it easier for them to interact with the interface.

Changes Made:
- Added role, tabindex, and aria-expanded attributes to filter and settings button
- Implemented `keydown` event listener for "Enter" key to trigger button click


assets/javascripts/keyevent.js
```diff
@@ -1,5 +1,31 @@
-// http://stackoverflow.com/questions/5681146/chrome-10-keyevent-or-something-similar-to-firefoxs-keyevent
+//Keyboard Accessibility
+// Wrap the code in a DOMContentLoaded event listener to ensure it runs after the DOM is fully loaded
```
Martchus, line 2: I don't think we need such a comment. At least in my opinion the code speaks for itself. However, does this actually wrap the code into the listener? To me it looks like you're just passing a function with an empty body to it and the rest of the code is executed immediately. I must ask: Have you actually tested this? It would be really appreciated to extend our UI tests.

Martchus (changes requested):

assets/javascripts/keyevent.js
Lokeshsri11: I think anyone beginner who comes to read the code may get some help from the comments, but ok I will delete the comments.

kalikiana: I would assume the following code should go into that function, but that's not the case?

Lokeshsri11:
I think I should closed this PR for now I don't how but that xml file was commit in my PR, no I was firstly Send the demo video of the Keyboard Accessibility with you and then raised the PR

PR #5105, opened 2023-04-27
Add pagination for /api/v1/test_suites, /api/v1/machines, and /api/v1/products

kraih:
They all share the same controller and action.

Progress: https://progress.opensuse.org/issues/121105

### DIFF
--- a/lib/OpenQA/WebAPI/Controller/API/V1/Table.pm
+++ b/lib/OpenQA/WebAPI/Controller/API/V1/Table.pm
@@ -3,7 +3,7 @@
 # SPDX-License-Identifier: GPL-2.0-or-later
 
 package OpenQA::WebAPI::Controller::API::V1::Table;
-use Mojo::Base 'Mojolicious::Controller';
+use Mojo::Base 'Mojolicious::Controller', -signatures;
 
 use Mojo::Util 'trim';
 use OpenQA::App;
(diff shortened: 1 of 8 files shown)


kraih:
The `codecov/path` fail is wrong. It complains about `Influxdb.pm`, which i did not change at all. 🙄

okurz (dismissed):
In the test I see duplicate test messages which make it harder to understand TAP output like which specific test step the message "has first page" refers to. I suggest you either extract test helper functions or make more test messages unique, e.g. "first page of test suites on prev link"

kalikiana:
> In the test I see duplicate test messages which make it harder to understand TAP output like which specific test step the message "has first page" refers to. I suggest you either extract test helper functions or make more test messages unique, e.g. "first page of test suites on prev link"

I guess subtests would help with that

Martchus (approved):
In theory it is a bit problematic that our UI cannot make use of this kind of paging. However, since the default limit is high enough it shouldn't come to this anyways.

okurz (commented):
do we really need to repeat all those pagination tests especially because they all use the same code?

kraih:
> do we really need to repeat all those pagination tests especially because they all use the same code?

Yes, i think so. The tests look very similar, but there are subtle differences and each limit is enforced by a completely different SQL query. The link headers themselves are not so important, but the fact that they represent the correct paginated data does in each case.

okurz (approved):
ok, I hope you checked that there is no significant runtime impact

kraih:
> ok, I hope you checked that there is no significant runtime impact

This is all very cheap and fast.

PR #5106, opened 2023-04-27
Verify and improve restarting behavior in certain case

Martchus:
* Add test for the problematic restart handling we have seen recently on
  o3 (see https://progress.opensuse.org/issues/128267)
    * Add test case to check whether restarting the 2nd root in a
      dependency tree like the one from that issue works as well (after
      the first root has already been restarted)
* Ensure the cluster has been stopped before enqueuing a Minion job that
  possibly restarts it as it would otherwise mark the scheduled jobs in
  that cluster as `parallel_failed`
* This test fails without https://github.com/os-autoinst/openQA/commit/29bcb6cd8bb0aeaf19372ad7a09269cc9557d8ab which
  is actually the main fix (although it intends to improve a slightly
  different situation)


lib/OpenQA/Schema/Result/Jobs.pm
```diff
@@ -2039,6 +2038,10 @@ sub done ($self, %args) {
         }
     }
 
+    # enqueue the finalize job only after stopping the cluster so in case the job should be restarted the cluster
+    # appears cancelled and thus it's jobs in (pre-)execution are not set to PARALLEL_RESTARTED by `auto_duplicate`
```
kalikiana, line 2042:
```suggestion
    # appears cancelled and thus its jobs in (pre-)execution are not set to PARALLEL_RESTARTED by `auto_duplicate`
```

mimi1vx (approved):

PR #5109, opened 2023-05-02
Makefile: Fix manual calls of codecov report

okurz:
Same as when generating reports for generating codecov reports we should
specify the variables which help us with local calls in case of "Sereal"
packages being available locally to prevent errors.


Martchus (approved):
Adding it at the top would be good as well. (Then we don't have to think where exactly we need this. Possibly in more places?)

okurz:
> Adding it at the top would be good as well. (Then we don't have to think where exactly we need this. Possibly in more places?)

I don't think this would do what e

perlpunk (approved):

PR #5111, opened 2023-05-02
Fix several issues with `allocate_network`

Martchus:
* Rewrite code to no longer use `next` in `catch` block (which
  does not work because this block is actually a nested function)
* Use raw SQL query to simplify code
    * A simple insert with `ON CONFLICT DO NOTHING` should be
      sufficient (instead of using a transaction)
    * Using the same query for applying the network to the whole
      job cluster should avoid running into
      "duplicate key value violates unique constraint" (see
      https://progress.opensuse.org/issues/128267#note-13)
* Die in case of a fatal error (instead of possibly provoking the
  error again and again in an endless loop)
* Improve log messages to always include the relevant job ID
* Add explicit unit tests

### DIFF
--- a/lib/OpenQA/Schema/Result/Jobs.pm
+++ b/lib/OpenQA/Schema/Result/Jobs.pm
@@ -1569,25 +1569,21 @@ sub allocate_network ($self, $name) {
 }
 
 sub _find_network ($self, $name, $seen = {}) {
-    return if $seen->{$self->id};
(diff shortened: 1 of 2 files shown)


mimi1vx (approved):

okurz (approved):

okurz:
@Mergifyio rebase

PR #5116, opened 2023-05-04
Add `xz` to worker container

ssssam:
Hello,

I realize the worker container can't be all things to everyone... but it would be handy for GNOME's OpenQA setup if it contained `xz` so we can uncompress disk images on the worker. The increase in container image size is 319KB, which is not very noticeable in a 1.44GB container image :-)

### DIFF
--- a/container/worker/Dockerfile
+++ b/container/worker/Dockerfile
@@ -7,7 +7,7 @@ LABEL version="0.3"
 RUN zypper ar -p 95 -f http://download.opensuse.org/repositories/devel:openQA/15.4 devel_openQA && \
     zypper ar -p 90 -f http://download.opensuse.org/repositories/devel:openQA:Leap:15.4/15.4 devel_openQA_Leap && \
     zypper --gpg-auto-import-keys ref && \
-    zypper in -yl ca-certificates-mozilla curl gzip rsync && \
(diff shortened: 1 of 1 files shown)


okurz (approved):

Martchus (approved):

ssssam:
Could someone trigger a rebuild of the worker container ? Last time [that was needed](https://github.com/os-autoinst/openQA/pull/4804#issuecomment-1270301025), seems the latest available worker container (sha256:85a1f6f8f17dba60923256b1430f9873c0229a24be843a8d0d58273c9985d83c) doesn't have xz inside it yet

okurz:
done, build in https://build.opensuse.org/package/show/devel:openQA/openQA_container_worker is currently ongoing

okurz:
Verified with

```
podman run --pull=always --rm -it --device=/dev/kvm -v /home/okurz/local/os-autoinst/openQA/etc/openqa/:/etc/openqa/:ro --entrypoint='' registry.opensuse.org/devel/openqa/containers15.4/openqa_worker:latest /bin/sh -c 'command -v xz'
```

ssssam:
Fantastic, thank you! Have also confirmed that xz is now present :)

PR #5117, opened 2023-05-04
Fix duplication of Devel::Cover requirements specified

okurz:

PR #5080, opened 2023-04-17
Improve contrast of progress bar showing number of failures

Lokeshsri11:
This commit aims to enhance the contrast of the progress bar used to display the number of failures. The previous implementation had low contrast, making it difficult to discern the actual progress. The updated design uses a more prominent color scheme, improving visibility and making it easier to understand the number of failures at a glance. This change improves the overall user experience and ensures that the progress bar is more accessible to all users.

232456699-4f5afdd7-306c-4399-856b-b8725b85bd3d

PR #5121, opened 2023-05-05
Improve contrast between failed and incomplete color

Martchus:
Commit c3af6a39f2e72559c045666d1bafd749001de538 increased the contrast of progress bar segments showing the number of failures by making the color we generally use for failures darker. However, that decreased the difference between the color of failures and incompletes which makes them harder to distinguish. So this commit reverts this original change. To still have a good contrast in the progress bar segment the text color within the progress bar is set to black. This looks good in bright and dark mode (the mode actually makes no difference for the progress bars).

### DIFF
--- a/assets/stylesheets/dashboard.scss
+++ b/assets/stylesheets/dashboard.scss
@@ -39,10 +39,7 @@ td div.progress {
     .progress-bar-passed, .progress-bar-softfailed, .progress-bar-unfinished {
         a { color: $color-result; }
     }
-    .progress-bar-failed{
-        a { color: white; }
-    }
-    .progress-bar-skipped {
+    .progress-bar-failed, .progress-bar-skipped {
         a { color: black; }
     }
     .progress-bar-passed {
--- a/assets/stylesheets/openqa_theme.scss
+++ b/assets/stylesheets/openqa_theme.scss
@@ -34,7 +34,7 @@ $link-hover-color:            darken($link-color, 15%) !default;
 $color-softfailed:            #EEB560;
 $color-ok:                    #CDFFB9;
 $color-warning:               #ffff99;
-$color-failed:                #cc3333;
+$color-failed:                #FF4E46;
 $color-result-failed:         #FF9E9B;
 $color-incomplete:            #AF1E11;
 $color-result:                #3c3c3c;


Martchus:
The progress bar now looks like this:
image

In the dark mode the colors are exactly the same so I'm saving me the screenshot here. (It is actually not that ideal that the colors are the same but that's a different topic.)

kalikiana:
Since this isn't just a revert, did you discuss it with @Lokeshsri11 yet? And e.g. make sure we know that black on #FF4E46 is actually a good combination.

perlpunk:
I retriggered the OBS checks, they looked like they failed because of some OBS internal problem (e.g. readonly file system)

perlpunk:
Actually that didn't help:
https://build.opensuse.org/package/live_build_log/devel:openQA:GitHub:os-autoinst:openQA:PR-5121/openQA/openSUSE_Tumbleweed/x86_64
```
[   48s] ### VM INTERACTION START ###
[   48s] [0;1;38;5;185mFailed to write wtmp record, ignoring: Read-only file system[0m
[   48s] [0;1;38;5;185mFailed to write utmp record: Read-only file system[0m
[   48s] Powering off.
[   48s] [   43.373792][ T3275] reboot: Power down
[   48s] ### VM INTERACTION END ###
```

perlpunk:
Oh, there's a different error actually:
https://build.opensuse.org/package/live_build_log/devel:openQA:GitHub:os-autoinst:openQA:PR-5121/openQA/openSUSE_Leap_15.4/x86_64
https://build.opensuse.org/package/live_build_log/devel:openQA:GitHub:os-autoinst:openQA:PR-5121/openQA/openSUSE_Tumbleweed/x86_64
```
[   91s] error: line 159: Dependency tokens must begin with alpha-numeric, '_' or '/': Requires:       perl(CSS::Minifier::XS) ...
```

perlpunk:
https://github.com/os-autoinst/openQA/pull/5122 should fix the OBS issue. Not sure why the error didn't popup already for #5117

perlpunk (approved):

okurz:
https://github.com/os-autoinst/openQA/pull/5121#issuecomment-1535944554 looks like the contrast on red is a bit too good now. Do we use some kind of gray text in the other areas?

Lokeshsri11:
> 

yes this one looks good

Martchus:
The other progress bar segments are likely the normal link color. I don't think it would look good on the red background, though. We already use black for skipped progress bar segments so I thought consistency with that would be nice. Of course we could also use the normal text color in all progress bar segments that so far use black and maybe even in all progress bar segments for the best consistency.

Lokeshsri11:
> The other progress bar segments are likely the normal link color. I don't think it would look good on the red background, though. We already use black for skipped progress bar segments so I thought consistency with that would be nice. Of course we could also use the normal text color in all progress bar segments that so far use black and maybe even in all progress bar segments for the best consistency.

Yes, I also feel the same thing, just change the red to normal red, everything else is fine.

okurz (approved):

Vogtinator:
At least for me, the black text on the dark red "failed jobs" bar is really hard to read.

Martchus:
I find it easier to read than the white text¹ although both is actually pretty well readable so I have no strong opinion on that. I also don't find the red very "dark". I suppose colors are a complicated matter :-)

¹ Which failed some contrast test (see https://github.com/os-autoinst/openQA/pull/5080).

okurz:
maybe we can now try to make the red a bit lighter?

Lokeshsri11:
> We can make it a little bit lighter - but not too light. It still needs good contrast with the usually white background.

Yes, if we do more light, then the contrast will not match on the light theme mode.

PR #5123, opened 2023-05-08
Output host of missing api credentials

perlpunk:
If one messes up the commandline options, e.g.

    --within-instance --json-output http://...

then the error message about the missing credentials can be confusing.
Printing out the host (or in this case the misplaced option) will make it clear what's wrong.


okurz (approved):

kalikiana (approved):
Nice catch!

kraih (approved):

perlpunk:
Somehow OBS doesn't report the results for Leap 15.4, that's why mergify is still waiting...

PR #5124, opened 2023-05-08
Increase heartbeat timeout to 15 minutes

kraih:
We have some API endpoints that can run in a blocking manner for 10+ minutes, such as `/api/v1/jobs/:id/set_done`. Which exceeds the 400 second heartbeat timeout we currently use.

Fortunately this timeout only exists so a worker cannot be blocked indefinitely with something like an infinite loop. So increasing it a bit more should not cause any noticeable side effects.

Long term we should consider moving such slow code into Minion background jobs though.

Progress: https://progress.opensuse.org/issues/128345

### DIFF
--- a/script/openqa-webui-daemon
+++ b/script/openqa-webui-daemon
@@ -1,3 +1,3 @@
 #!/bin/sh -e
 # Our API commands are very expensive, so the default timeouts are too tight
-exec "$(dirname "$0")"/openqa prefork -m production --proxy -i 100 -H 400 -w 30 -c 1 -G 800 "$@"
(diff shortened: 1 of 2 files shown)


okurz (approved):

Martchus (approved):
I'm a bit surprised that `set_done` takes *that* long but ok.

kraih:
> I'm a bit surprised that `set_done` takes _that_ long but ok.

It's the `carry_over_bugrefs` code specifically: https://progress.opensuse.org/issues/128345#note-6

Martchus:
That makes unfortunately sense. Supposedly this should go into the finalize Minion job then. We just need to be careful about interactions of this with the hook script or consumers of events.

PR #5125, opened 2023-05-09
Allow datetime in tag's build

mimi1vx:
See: https://progress.opensuse.org/issues/126623

### DIFF
--- a/lib/OpenQA/Schema/Result/Comments.pm
+++ b/lib/OpenQA/Schema/Result/Comments.pm
@@ -133,7 +133,7 @@ e.g. 'tag:0123:important:GM' returns a list of '0123', 'important' and 'GM'.
 =cut
 sub tag ($self) {
     $self->text
-      =~ /\btag:((?<version>[-.@\d\w]+)-)?(?<build>[-.@\d\w]+):(?<type>[-@\d\w]+)(:(?<description>[-.@\d\w]+))?\b/;
+      =~ /\btag:(((?<version>[-.@\d\w]+)-)?(?<build>[-.@\d\w]+)|"((?<version>[-.@\d\w]+)-)?(?<build>[-.@\d\w\s\+:]+)"):(?<type>[-@\d\w]+)(:(?<description>[-.@\d\w]+))?\b/;
     return $+{build}, $+{type}, $+{description}, $+{version};
 }
 
--- a/t/17-build_tagging.t
+++ b/t/17-build_tagging.t
@@ -355,4 +355,26 @@ subtest 'content negotiation' => sub {
       ->content_type_is('application/json;charset=UTF-8');
 };
 
+subtest 'tags with datetime' => sub {
+    create_job_version_build('Factory', '2023-03-24 14:33:39+01:00');
+    post_comment_1001 'tag:"2023-03-24 14:33:39+01:00":important:RC3';
+    $t->get_ok('/group_overview/1001')->status_is(200);
+    my @tags = $t->tx->res->dom->find('.tag')->map('text')->each;
+    my @title = $t->tx->res->dom->find('.tag')->map(attr => 'title')->each;
+    is(scalar @tags, 7, 'seven builds tagged');
+    is(scalar @title, 7, 'seven important titles');
+    is($tags[0], 'RC3', 'tag description shown');
+};
+
+
+subtest 'tags with quoted build works' => sub {
+    create_job_version_build('Factory', '050foobar');
+    post_comment_1001 'tag:"050foobar":important:released';
+    $t->get_ok('/group_overview/1001')->status_is(200);
+    my @tags = $t->tx->res->dom->find('.tag')->map('text')->each;
+    is(scalar @tags, 7, 'seven builds tagged');
+    is($tags[1], 'released', 'tag description shown');
+};
+
+
 done_testing;


lib/OpenQA/Schema/Result/Comments.pm
```diff
@@ -132,9 +132,17 @@ Returns C<build_nr>, C<type> and optionally C<description> if C<$self> is tag,
 e.g. 'tag:0123:important:GM' returns a list of '0123', 'important' and 'GM'.
 =cut
 sub tag ($self) {
-    $self->text
-      =~ /\btag:((?<version>[-.@\d\w]+)-)?(?<build>[-.@\d\w]+):(?<type>[-@\d\w]+)(:(?<description>[-.@\d\w]+))?\b/;
-    return $+{build}, $+{type}, $+{description}, $+{version};
+    my $text = $self->text;
```
kalikiana, line 135: suggest to add at least example in a unit test. Test-driven development will make it very obvious whether this works or not.

mimi1vx: http://quasar.suse.cz/group_overview/1?fullscreen=1 .. the test is a different problem, it looks like `post_comment_1001` don't work as excepted with tag ...

kalikiana: I don't know what you mean by that. I was suggesting TDD because it would tell me 1) what you are testing 2) where it fails. Are we talking about posting comments? Or displaying comments?

Martchus: So it looks indeed good on http://quasar.suse.cz/group_overview/1?fullscreen=1. I assume the remaining problem is that the test helper function `post_comment_1001` doesn't allow you to create recognized tags at all? If you need any help, feel free to ask e.g. in the chat or here.

mimi1vx:
extended regex for `+\s` .. to be able match build based on datetime

lib/OpenQA/Schema/Result/Comments.pm
```diff
@@ -132,9 +132,17 @@ Returns C<build_nr>, C<type> and optionally C<description> if C<$self> is tag,
...
+    if ($text
+        =~ /\btag:((?<version>[-.@\d\w]+)-)?(?<build>[-.@\d\w]+):(?<type>[-@\d\w]+)(:(?<description>[-.@\d\w]+))?\b/)
+    {
+        return $+{build}, $+{type}, $+{description}, $+{version};
+    }
+    elsif ($text
+        =~ /\btag:"((?<version>[-.@\d\w]+)-)?(?<build>[-.@\d\w\s\+:]+)":(?<type>[-@\d\w]+)(:(?<description>[-.@\d\w]+))?\b/)
+    {
+        return $+{build}, $+{type}, $+{description}, $+{version};
+    }
```
Martchus, line 145: It would be nice if the `return` paths could be de-duplicated. Maybe the following works?
```suggestion
    if (($text
        =~ /\btag:((?<version>[-.@\d\w]+)-)?(?<build>[-.@\d\w]+):(?<type>[-@\d\w]+)(:(?<description>[-.@\d\w]+))?\b/)
        || ($text
        =~ /\btag:"((?<version>[-.@\d\w]+)-)?(?<build>[-.@\d\w\s\+:]+)":(?<type>[-@\d\w]+)(:(?<description>[-.@\d\w]+))?\b/))
    {
        return $+{build}, $+{type}, $+{description}, $+{version};
    }
```

t/17-build_tagging.t
```diff
@@ -88,7 +88,7 @@ subtest 'tag icon on group overview on important build' => sub {
 };
 
 subtest 'test whether tags with @ work, too' => sub {
-    post_comment_1001 'tag:0048@0815:important:RC2';
+    post_comment_1001 'tag:"0048@0815:test":important:RC2';
```
kalikiana, line 91: Please add a new subtest w/ `2023-03-24 14:33:39+01:00` in particular

t/17-build_tagging.t
```diff
@@ -16,7 +16,7 @@ use OpenQA::Test::Utils 'perform_minion_jobs';
 use OpenQA::JobGroupDefaults;
 use OpenQA::Schema::Result::JobGroupParents;
 use OpenQA::Jobs::Constants;
-
+use Data::Dumper 'Dumper';
```
perlpunk, line 19: leftover?

t/17-build_tagging.t
```diff
@@ -355,4 +356,25 @@ subtest 'content negotiation' => sub {
...
+};
+
+
+subtest 'test whether tags with quoted build work, too' => sub {
+    create_job_version_build("Factory", "050foobar");
+    post_comment_1001 'tag:"050foobar":important:released';
+    $t->get_ok('/group_overview/1001')->status_is(200);
+    my @tags = $t->tx->res->dom->find('.tag')->map('text')->each;
+    is(scalar @tags, 7, 'seven builds tagged');
+    # this build will sort *above* the first build, so item 0
```
kalikiana, line 375:
```suggestion
```

That's not the case here I think

kalikiana (approved):

t/17-build_tagging.t
```diff
@@ -355,4 +356,24 @@ subtest 'content negotiation' => sub {
       ->content_type_is('application/json;charset=UTF-8');
 };
 
+subtest 'test whether tags with datetime work, too' => sub {
+    create_job_version_build("Factory", "2023-03-24 14:33:39+01:00");
+    post_comment_1001 'tag:"2023-03-24 14:33:39+01:00":important:RC3';
+    $t->get_ok('/group_overview/1001')->status_is(200);
+    my @tags = $t->tx->res->dom->find('.tag')->map('text')->each;
+    is(scalar @tags, 7, 'seven builds tagged');
+    is($tags[0], 'RC3', 'tag description shown');
```
Martchus, line 365: You could also check whether the tooltip (title attribute) is correct (which should be "important"). The same counts for the other subtest.

t/17-build_tagging.t
```diff
@@ -63,6 +63,7 @@ my $job_hash = {
     group_id => 1001
 };
 
+
```
Martchus, line 66:
```suggestion
```

I think generally one new line between subtests is enough - and that's also the usual amount of spacing we use in this repo.

t/17-build_tagging.t
mimi1vx: Isn't this checked in other sub tests?

Martchus: I just thought it wouldn't hurt in that subtest as well.

lib/OpenQA/Schema/Result/Comments.pm
```diff
@@ -132,9 +132,18 @@ Returns C<build_nr>, C<type> and optionally C<description> if C<$self> is tag,
...
+    my $text = $self->text;
+    if (
+        (
+            $text
+            =~ /\btag:((?<version>[-.@\d\w]+)-)?(?<build>[-.@\d\w]+):(?<type>[-@\d\w]+)(:(?<description>[-.@\d\w]+))?\b/
+        )
+        || ($text
+            =~ /\btag:"((?<version>[-.@\d\w]+)-)?(?<build>[-.@\d\w\s\+:]+)":(?<type>[-@\d\w]+)(:(?<description>[-.@\d\w]+))?\b/
+        ))
+    {
```
okurz, line 144: I think having two nearly identical variants of regex strings which already contain optional parts are not that readable. I suggest to combine both. Something like

```suggestion
    if ($text =~ /\btag:(((?<version>[-.@\d\w]+)-)?(?<build>[-.@\d\w]+)|"((?<version>[-.@\d\w]+)-)?(?<build>[-.@\d\w\s\+:]+)"):(?<type>[-@\d\w]+)(:(?<description>[-.@\d\w]+))?\b/) {
```

This is of course also nearly unreadable so I suggest to invest some effort in improving, e.g. with variables.

Also the commit message could have a bit more details, e.g. what issue is fixed, why this needs to change, to which versions of which product it applies, link to a bug or a fate entry, the choices you made, etc. Also see https://commit.style/ or http://chris.beams.io/posts/git-commit/ as a helpful guide how to write good commit messages.

t/17-build_tagging.t
```diff
@@ -355,4 +356,24 @@ subtest 'content negotiation' => sub {
       ->content_type_is('application/json;charset=UTF-8');
 };
 
+subtest 'test whether tags with datetime work, too' => sub {
+    create_job_version_build("Factory", "2023-03-24 14:33:39+01:00");
```
okurz, line 360:
```suggestion
subtest 'tags with datetime' => sub {
    create_job_version_build('Factory', '2023-03-24 14:33:39+01:00');
```

t/17-build_tagging.t
```diff
@@ -355,4 +356,24 @@ subtest 'content negotiation' => sub {
...
+    post_comment_1001 'tag:"2023-03-24 14:33:39+01:00":important:RC3';
+    $t->get_ok('/group_overview/1001')->status_is(200);
+    my @tags = $t->tx->res->dom->find('.tag')->map('text')->each;
+    is(scalar @tags, 7, 'seven builds tagged');
+    is($tags[0], 'RC3', 'tag description shown');
+};
+
+
+subtest 'test whether tags with quoted build work, too' => sub {
```
okurz, line 369:
```suggestion
subtest 'tags with quoted build works too' => sub {
```

t/17-build_tagging.t
```diff
@@ -355,4 +356,24 @@ subtest 'content negotiation' => sub {
...
+    create_job_version_build("Factory", "050foobar");
```
okurz, line 370:
```suggestion
    create_job_version_build('Factory', '050foobar');
```

okurz (changes requested):

lib/OpenQA/Schema/Result/Comments.pm
Martchus: I don't think making one huge regex out of it helps with readability.

okurz: Then variables in the string. But if/else with *two* huge regex strings isn't better

mimi1vx: @okurz in perl re is possible to have two subpatterns with the same name? in that case one ugly regex is still better than two :D

okurz (approved):

Martchus (approved):

PR #5127, opened 2023-05-10
Explicitly match maintenance repo settings variables

kalikiana:
See: https://progress.opensuse.org/issues/128909


perlpunk:
If that's what's really wanted then I suggest to remove the whole for loop. It's a hash, so we can check directly if a setting is there or not:
```perl
sub collect_incident_repos ($url_handler, $settings) {
    my @urls;
    return verify_incident_repos($url_handler, $settings->{INCIDENT_REPO}) if defined $settings->{INCIDENT_REPO};
    if (defined $settings->{SCC_ADDONS}) {
        foreach my $SCC_ADDON (split /,/, $settings->{SCC_ADDONS}) {
            next unless $settings->{uc($SCC_ADDON) . '_TEST_REPOS'};
            my $incident_urls = verify_incident_repos($url_handler, $settings->{uc($SCC_ADDON) . '_TEST_REPOS'});
            push @urls, @$incident_urls;
        }
    }
    return \@urls;
}
```

lib/OpenQA/Script/CloneJobSUSE.pm
```diff
@@ -11,8 +11,8 @@ our @EXPORT = qw(detect_maintenance_update);
 sub collect_incident_repos ($url_handler, $settings) {
     my @urls;
     for my $setting (keys %$settings) {
-        return verify_incident_repos($url_handler, $settings->{'INCIDENT_REPO'}) if ($setting =~ /INCIDENT_REPO/);
-        if ($setting =~ /SCC_ADDONS/) {
+        return verify_incident_repos($url_handler, $settings->{'INCIDENT_REPO'}) if ($setting eq "INCIDENT_REPO");
```
Martchus, line 14:
```suggestion
        return verify_incident_repos($url_handler, $settings->{'INCIDENT_REPO'}) if $setting eq 'INCIDENT_REPO';
```

lib/OpenQA/Script/CloneJobSUSE.pm
```diff
@@ -11,8 +11,8 @@ our @EXPORT = qw(detect_maintenance_update);
...
+        if ($setting eq "SCC_ADDONS") {
```
Martchus, line 15:
```suggestion
        if ($setting eq 'SCC_ADDONS') {
```

Martchus (changes requested):
This shows the outer loop shouldn't be one.

E.g. having
```
if (my $repo =  $settings->{INCIDENT_REPO}) {
    return verify_incident_repos($url_handler, $settings->{INCIDENT_REPO})
}
```
outside of the outer loop will do it. The same can be applied to `SCC_ADDONS`.

And I suppose the "explicit match" as you put it is really what was intended here. Otherwise this code would just be utterly stupid. The case that emits a warning was never working and noone complained about it. (We only found out about the warning in an unrelated way.)

perlpunk (approved):

PR #5128, opened 2023-05-10
Fix sporadic failures of comments test

Martchus:
This will hopefully fix https://progress.opensuse.org/issues/128153 which still sometimes occurs. It explicitly waits until the text form is no longer empty. Maybe this is not sufficient but for now I'll try to keep this workaround as simple as possible. (It is a workaround because `send_keys` should actually only return after the text is present.)

### DIFF
--- a/t/ui/15-comments.t
+++ b/t/ui/15-comments.t
@@ -3,6 +3,7 @@
 # SPDX-License-Identifier: GPL-2.0-or-later
 
 use Test::Most;
+use Mojo::Base -base, -signatures;
 use utf8;
 use FindBin;
 use lib "$FindBin::Bin/../lib", "$FindBin::Bin/../../external/os-autoinst-common/lib";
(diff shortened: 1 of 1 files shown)


t/ui/15-comments.t
```diff
@@ -60,6 +61,19 @@ my $edited_test_message = $test_message . $another_test_message;
 my $description_test_message = "pinned-description ... The description";
 my $user_name = 'Demo';
 
+# fills out the comment form and submits
+sub write_comment ($text, $desc) {
+    wait_for_element selector => '#text', is_displayed => 1, description => 'comment form is displayed';
+    $driver->find_element_by_id('text')->send_keys($text);
```
kalikiana, line 67: Can we combine these?

```suggestion
    wait_for_element(selector => '#text', is_displayed => 1, description => 'comment form is displayed')->send_keys($text);
```

perlpunk:
`$title =~ s/sporatic/sporadic/`

t/ui/15-comments.t
Martchus: good idea

okurz (approved):
very clean improvement, great!

perlpunk (approved):

PR #5129, opened 2023-05-11
Only output JSON when using --json-output

perlpunk:
If --json-output is used, users expect only JSON and no additional text, so the JSON can easily be parsed.
It was printed to stderr, so it can be treated separately, but still, such text output should only appear when for example in a verbose mode.

Issue: https://progress.opensuse.org/issues/128405

### DIFF
--- a/lib/OpenQA/Script/CloneJob.pm
+++ b/lib/OpenQA/Script/CloneJob.pm
@@ -245,7 +245,9 @@ sub clone_job ($jobid, $url_handler, $options, $post_params = {}, $jobs = {}, $d
         next unless $job->{$job_type};
 
         my ($chained, $directly_chained, $parallel) = get_deps($job, $job_type);
-        print STDERR "Cloning $job_type of $job->{name}\n" if @$chained || @$directly_chained || @$parallel;
+        print STDERR "Cloning $job_type of $job->{name}\n"
+          if not $options->{'json-output'} && (@$chained || @$directly_chained || @$parallel);
+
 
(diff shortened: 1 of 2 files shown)


kalikiana (approved):

Martchus (commented):
This this is just about stderr anymore so I'm wondering whether we actually need this. Haven't you fixed the other helper function so it doesn't mix the output streams? I personally find it actually a good practice to *not* suppress error messages and thus would prefer to keep them. Normally the split between stdout and stderr allows that. One could also neatly integrate the error message into the JSON output so it becomes part of the parsable output as well.

perlpunk:
> Haven't you fixed the other helper function so it doesn't mix the output streams?

Yes. It will now print anything it gets on stderr to stderr again with some information about the command that was called.
It still means that anything on stderr will land in our logs, so I thought I remove anything that just looks like an informal message. 

> I personally find it actually a good practice to _not_ suppress error messages and thus would prefer to keep them.

Absolutely. However, In this case I thought this was just an informal message:
```
print STDERR "Cloning $job_type of $job->{name}\n"
```
So if this is an error message, I don't understand it. It certainly doesn't look like an error message to me. It says "Cloning ..." and then it does the actual cloning.

> Normally the split between stdout and stderr allows that. One could also neatly integrate the error message into the JSON output so it becomes part of the parsable output as well.

Just that currently the returned JSON is a hash with the original job id(s) as keys. A new key like `stderr` would of course not conflict wit hthis, just look a bit weird.

Martchus (approved):
You're right, it is just an information message.

PR #5132, opened 2023-05-12
Improve webhook/CI related code

Martchus:
* See individual commit messages
* Related ticket: https://progress.opensuse.org/issues/127949

### DIFF
--- a/lib/OpenQA/WebAPI/Controller/API/V1/Webhook.pm
+++ b/lib/OpenQA/WebAPI/Controller/API/V1/Webhook.pm
@@ -98,16 +98,14 @@ sub product ($self) {
       unless $SUPPORTED_PR_ACTIONS{$action};
     my $pr = $json->{pull_request} // {};
     my $head = $pr->{head} // {};
-    my $sha = $head->{sha};
     my $repo = $head->{repo} // {};
-    my $repo_name = $repo->{full_name};
-    my $clone_url = $repo->{clone_url};
-    my $statuses_url = $pr->{statuses_url};
-    my $html_url = $pr->{html_url};
-    return $self->render(
-        status => 400,
-        text => '"pull_request" lacks "statuses_url", "head/sha" or "head/repo/full_name" or "head/repo/clone_url"'
-    ) unless $sha && $repo_name && $clone_url && $statuses_url;
+    my @missing;
+    push @missing, 'pull_request/id' unless my $pr_id = $pr->{id};
+    push @missing, 'pull_request/statuses_url' unless my $statuses_url = $pr->{statuses_url};
+    push @missing, 'pull_request/head/sha' unless my $sha = $head->{sha};
+    push @missing, 'pull_request/head/repo/full_name' unless my $repo_name = $repo->{full_name};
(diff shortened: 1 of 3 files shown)


lib/OpenQA/WebAPI/Controller/API/V1/Webhook.pm
```diff
@@ -131,9 +129,11 @@ sub product ($self) {
     $params{CI_TARGET_URL} = $base_url if $base_url;
 
     # set GitHub parameters so the Minion job will be able to report the status back to GitHub
+    my $html_url = $pr->{html_url};
     $params{GITHUB_REPO} = $repo_name;
     $params{GITHUB_SHA} = $sha;
     $params{GITHUB_STATUSES_URL} = $statuses_url;
+    $params{GITHUB_PR_ID} = $pr_id;
```
kalikiana, line 136: Would you mind elaborating how this is helpful? It doesn't look like the ID is getting used anywhere.

Martchus: As stated in the commit message:
> * We need some unique ID to associate subsequent updates with
>    an existing scheduled product to be able to implement
>    cancellation and cleanup

I've realized that adding this to the settings might not be the best idea performance-wise and created #5133 for introducing another column.

Regardless how we do this - in this PR alone it'll be not used indeed (except for being displayed) and the code will likely change slightly later. I've been creating this PR mainly to avoid too many changes in one PR but I can also close it if incomplete changes aren't wanted.

lib/OpenQA/WebAPI/Controller/API/V1/Webhook.pm
```diff
@@ -98,16 +98,14 @@ sub product ($self) {
...
-    my $repo_name = $repo->{full_name};
-    my $clone_url = $repo->{clone_url};
-    my $statuses_url = $pr->{statuses_url};
-    my $html_url = $pr->{html_url};
-    return $self->render(
-        status => 400,
-        text => '"pull_request" lacks "statuses_url", "head/sha" or "head/repo/full_name" or "head/repo/clone_url"'
-    ) unless $sha && $repo_name && $clone_url && $statuses_url;
+    my @missing;
+    push @missing, 'pull_request/id' unless my $pr_id = $pr->{id};
```
okurz, line 103: I'd favor less procedural code but I guess you would have applied a "map" approach or something different if you would have known a good way :)

okurz (approved):

lib/OpenQA/WebAPI/Controller/API/V1/Webhook.pm
Martchus: I tried but ended up with that code as the best solution. Here is what I was thinking/considering: This approach has the advantage of being not complicated at all, one has the required fields directly available as local variables and the error message that is computed is also very specific. There is also just one line for each variable which is acceptable. Better than one line for each variable (e.g. a map definition that's too long for one line) *plus* some extra helper code to make it actually do anything. I guess the next best approach would be to define a schema and use the JSON validator but I personally find it more complicated to understand as one then need to understand the abstraction to understand what the code does. It would probably be also more lines of code and definitely less efficient.

mimi1vx (approved):

PR #5134, opened 2023-05-14
Remove codecov token again after problem fixed

Martchus:
According to https://community.codecov.com/t/commit-sha-does-not-match-circle-build/4266/11?u=martchus the problem has been fixed so the token added via
2af754957c56053241a07d05499808ecc236f811 can be removed again.

Related progress issue: https://progress.opensuse.org/issues/128129

### DIFF
--- a/codecov.yml
+++ b/codecov.yml
@@ -1,5 +1,4 @@
 codecov:
-  token: 058d4462-cb3c-455d-bb6d-3e6859f3de9e
   branch: master
 comment:
   behavior: default


Martchus:
I'm curious whether this works now. I'm not sure whether we should actually merge it. Maybe it makes sense to actually keep the token to be on the safe side?

okurz (approved):
I am for removing the token. I doubt that the next problem whatever it will be is manifesting as exactly the same so approving

mimi1vx (approved):

AdamWill:
Did you also change to a *new* token and deactivate the old one? because the old one will now be visible in git history forever...

Martchus:
Maybe we should do that - although it isn't *that* important I suppose. Otherwise we likely shouldn't have made the token public in the first place (not even as a workaround).

PR #5136, opened 2023-05-15
Fix sporadic failure in t/ui/18-tests-details.t

Martchus:
Fixes
```
getElementText: stale element reference: element is not attached to the page document …
Selenium::Remote::WebElement::get_text(Test::Selenium::Remote::WebElement=HASH(0x563e27154330)) called at t/ui/18-tests-details.t line 340
```
which can occur when the element is swapped out by JavaScript at the wrong moment. That the JavaScript code swaps out the element is actually provoked by this test and supposed to happen.

This change simply does the check in JavaScript to avoid running into this condition.

### DIFF
--- a/t/ui/18-tests-details.t
+++ b/t/ui/18-tests-details.t
@@ -336,8 +336,8 @@ subtest 'running job' => sub {
         # pretend the text result has been uploaded
         $aplay_text_result->spurt('some text result');
         update_status sub {
-            $step_detail_element = $driver->find_element('#module_aplay .links');
-            $step_detail_element && $step_detail_element->get_text !~ qr/Unable to read/;
+            my $text = $driver->execute_script('return document.querySelector("#module_aplay .links")?.textContent');
+            ($text !~ qr/Unable to read/) && ($step_detail_element = $driver->find_element('#module_aplay .links'));
         }, 'step detail shows no longer "Unable to read"';
         ok $step_detail_element, 'step detail still present' or return;
         like $step_detail_element->get_text, qr/some text result/, 'text result finally shown';


t/ui/18-tests-details.t
```diff
@@ -336,8 +336,8 @@ subtest 'running job' => sub {
         # pretend the text result has been uploaded
         $aplay_text_result->spurt('some text result');
         update_status sub {
-            $step_detail_element = $driver->find_element('#module_aplay .links');
-            $step_detail_element && $step_detail_element->get_text !~ qr/Unable to read/;
+            my $text = $driver->execute_script('return document.querySelector("#module_aplay .links")?.textContent');
+            ($text !~ qr/Unable to read/) && ($step_detail_element = $driver->find_element('#module_aplay .links'));
         }, 'step detail shows no longer "Unable to read"';
         ok $step_detail_element, 'step detail still present' or return;
         like $step_detail_element->get_text, qr/some text result/, 'text result finally shown';
```
okurz, line 343: this whole block looks overly complicated. Can't we just use

```
$driver->element_text_is('#module_aplay .links', 'some text result', 'text result finally shown';
```

replacing the complete text from "update_status" to the final like?

I don't think we should workaround with javascript. Doesn't look that clean.

perlpunk:
`s/sporatic/sporadic/`

t/ui/18-tests-details.t
Martchus: I'm not sure whether `element_text_is` is race-free. Besides, invoking that function at a time when the element text is not the expected one will lead to a failing test. That's not what we want during this wait-loop. Possibly we also still need the "like" comparison (instead of "is"). Note that there might also be the case when the element doesn't exist at all. Also in this case we just want to wait and not fail the test.

Martchus: Looks like `element_text_is` is internally just calling the same code we had before. So besides from not doing what we want it is also not helping us to get rid of the sporadic failures.

okurz: what about [wait_until](https://github.com/os-autoinst/openQA/blob/master/t/lib/OpenQA/SeleniumTest.pm#L311) or [wait_for_element](https://github.com/os-autoinst/openQA/blob/master/t/lib/OpenQA/SeleniumTest.pm#L342) ?

mimi1vx (approved):

kalikiana (approved):

PR #5137, opened 2023-05-15
t: Fix sporadic error about not logged in user

okurz:
* t: Simplify chromium dir lookup
* t: Fix sporadic error about not logged in user

### DIFF
--- a/t/33-developer_mode.t
+++ b/t/33-developer_mode.t
@@ -258,6 +258,7 @@ subtest 'developer session visible in live view' => sub {
 
 subtest 'status-only route accessible for other users' => sub {
     $driver->get('/logout');
+    $driver->element_text_is('#user-action a', 'Login', 'logged-out before logging in as otherdeveloper');
     $driver->get('/login?user=otherdeveloper');
     $driver->element_text_is('#user-action a', 'Logged in as otherdeveloper', 'otherdeveloper logged-in');
     assert_initial_ui_state();
@@ -303,6 +304,7 @@ subtest 'resume test execution and 2nd tab' => sub {
     # login as demo again
     $driver->switch_to_window($first_tab);
     $driver->get('/logout');
+    $driver->element_text_is('#user-action a', 'Login', 'logged-out before logging in as Demo');
     $driver->get('/login?user=Demo');
 
     # go back to the live view


t/33-developer_mode.t
```diff
@@ -258,8 +258,9 @@ subtest 'developer session visible in live view' => sub {
 
 subtest 'status-only route accessible for other users' => sub {
     $driver->get('/logout');
+    $driver->element_text_is('#user-action a', 'Login', 'logged-out before logging in as otherdeveloper');
     $driver->get('/login?user=otherdeveloper');
-    is $driver->find_element('#user-action a')->get_text(), 'Logged in as otherdeveloper', 'otherdeveloper logged-in';
+    $driver->element_text_is('#user-action a', 'Logged in as otherdeveloper', 'otherdeveloper logged-in');
```
Martchus, line 263: This doesn't change anything. Internally it will just be a sequence of `find_element` + `get_text`. I've just looked at the code:

```
sub _find_element_with_action {
…
    my $element;
    eval { $element = $self->find_element( $locator, $locator_strategy ); };
    if ($@) {
        print "# Error: $@\n";
        return 0;
    }
    return $element->$method( $params, $desc );
}

sub element_text_is {
    my $self   = shift;
    my $method = 'text_is';
    return $self->_find_element_with_action( $method, @_ );
}
```

If your repeated runs show a difference then the sample size is likely too small.

Note that this function also does not introduce a loop or anything in order to wait until the text matches the expected value. Maybe that's what we needed to do - although I'm not quite sure yet.

Martchus (changes requested):

Martchus:
It would actually be interesting to compute the confidence we have that the issue is fixed after executing an experiment like you've just did¹. It would also be interesting to compute the number of test runs required to achieve a certain confidence (e.g. 90 %). I think we did computations like that in the statistics class.

---

```
Before:
count_fail_ratio: Run: 20. Fails: 3. Fail ratio 15.00%

After:
count_fail_ratio: Run: 70. Fails: 0. Fail ratio 0%
```

t/33-developer_mode.t
okurz: The actual fix is line 261 which explicitly checks that the expected login is there. This is what I tried to explain in the git commit message https://github.com/os-autoinst/openQA/pull/5137/commits/dce1b8e510cd6e7faa9f8615b256becd3eeebee7 . The actual effect might just be the same as a small "sleep" call making it more likely to show the login button after time. I am sure that this actually only reduces the likelyhood back to values we had some time ago. The test is still unstable.

lib/OpenQA/Shared/Plugin/SharedHelpers.pm
```diff
@@ -40,44 +37,38 @@ sub _determine_os_autoinst_web_socket_url {
...
+sub _current_user ($c) {
     # If the value is not in the stash
     my $current_user = $c->stash('current_user');
     unless ($current_user && ($current_user->{no_user} || defined $current_user->{user})) {
         my $id = $c->session->{user};
         my $user = $id ? $c->schema->resultset("Users")->find({username => $id}) : undef;
         $c->stash(current_user => $current_user = $user ? {user => $user} : {no_user => 1});
     }
+    use Data::Dumper;
+    $Data::Dumper::Maxdepth = 2;
```
perlpunk, line 54: Please don't set global variables. Use `local` instead.

okurz: that shouldn't even be included, just for debugging, sorry

lib/OpenQA/Shared/Plugin/SharedHelpers.pm
```diff
@@ -61,23 +54,17 @@ sub _current_user {
...
 }
 
-sub _is_admin {
-    my $c = shift;
-    my $user = shift || $c->current_user;
-
+sub _is_admin ($c, $user = undef) {
+    $user //= $c->current_user;
     return ($user && $user->is_admin);
 }
```
Martchus, line 65: I've checked all usages as this may introduce security bugs but it seems fine.

Martchus (approved):

okurz:
I am pulling out https://github.com/os-autoinst/openQA/pull/5200

My experiments:
* Before: count_fail_ratio: Run: 53. Fails: 7. Fail ratio 13.00±0%
* After: count_fail_ratio: Run: 24. Fails: 4. Fail ratio 16.66±14.91%

so no effect

okurz:
@mergifyio rebase

Martchus:
I'm closing this because this commit has been merged as part of #5218.

PR #5141, opened 2023-05-17
Fix logic in openqa-clone-job

perlpunk:
The `&&` took precendence over the `not`.

Issue: https://progress.opensuse.org/issues/128405

### DIFF
--- a/lib/OpenQA/Script/CloneJob.pm
+++ b/lib/OpenQA/Script/CloneJob.pm
@@ -246,7 +246,7 @@ sub clone_job ($jobid, $url_handler, $options, $post_params = {}, $jobs = {}, $d
 
         my ($chained, $directly_chained, $parallel) = get_deps($job, $job_type);
         print STDERR "Cloning $job_type of $job->{name}\n"
-          if not $options->{'json-output'} and (@$chained || @$directly_chained || @$parallel);
+          if !$options->{'json-output'} && (@$chained || @$directly_chained || @$parallel);
 
 
         for my $dependencies ($chained, $directly_chained, $parallel) {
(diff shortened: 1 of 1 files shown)


lib/OpenQA/Script/CloneJob.pm
```diff
@@ -246,7 +246,7 @@ sub clone_job ($jobid, $url_handler, $options, $post_params = {}, $jobs = {}, $d
 
         my ($chained, $directly_chained, $parallel) = get_deps($job, $job_type);
         print STDERR "Cloning $job_type of $job->{name}\n"
-          if not $options->{'json-output'} && (@$chained || @$directly_chained || @$parallel);
+          if not $options->{'json-output'} and (@$chained || @$directly_chained || @$parallel);
```
Martchus, line 249: I prefer using the "normal" operators in conditions exclusively to be more in-line with how it was done in other  programming languages:
```suggestion
          if !$options->{'json-output'} && (@$chained || @$directly_chained || @$parallel);
```
I'd only use the operators with "strange" precedence when chaining commands.

Martchus (changes requested):

lib/OpenQA/Script/CloneJob.pm
kalikiana: Can we cover this in style checks? 🤔

Martchus: Both, `&&` and `and` are short-circuiting. The only difference is the precedence which works out as intended in @perlpunk change and my change. So it is really just the question of what style we prefer in this case and I'd say `&&`.

Martchus (approved):

kalikiana (approved):

PR #5144, opened 2023-05-18
Separate 'Build' from real build display in groups and search results

mimi1vx:
https://progress.opensuse.org/issues/129871
After:
Group overview
Search results

### DIFF
--- a/assets/javascripts/tests.js
+++ b/assets/javascripts/tests.js
@@ -38,7 +38,7 @@ function renderMediumName(data, type, row) {
     link += '&groupid=' + row.group;
   }
 
-  var name = "<a href='" + link + "'>" + 'Build' + row.build + '</a>';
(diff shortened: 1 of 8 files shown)


kalikiana:
@mimi1vx Is there a ticket for this? Maybe we should discuss the motivation first.

Martchus:
Shouldn't this be rather "Build: ..." as normally a colon is used in cases like that? However, maybe we can also get rid of"Build" completely? I'm also wondering about the motivation, though.

Martchus:
If we make it configurable we might as well make the whole prefix configurable. But on which level? This is supposedly not only about the group-related dashboards but also about other tables.

okurz (changes requested):
> If we make it configurable we might as well make the whole prefix configurable. But on which level? This is supposedly not only about the group-related dashboards but also about other tables.

For that I suggest we clarify the use case first

PR #5146, opened 2023-05-19
Consider a tag's version when cleaning up job results

Martchus:
When cleaning up job results, jobs being part of an important build are treated differently. So far really only the BUILD variable is used to decide whether a job belongs to an important build - even if the according tag includes a version number.

With this change the version from the tag is considered as well and matched against the job's VERSION variable.

Related ticket/note: https://progress.opensuse.org/issues/129244#note-18

### DIFF
--- a/lib/OpenQA/Schema/Result/JobGroups.pm
+++ b/lib/OpenQA/Schema/Result/JobGroups.pm
@@ -7,6 +7,7 @@ use Mojo::Base 'DBIx::Class::Core', -signatures;
 use OpenQA::App;
 use OpenQA::Markdown 'markdown_to_html';
 use OpenQA::JobGroupDefaults;
+use OpenQA::Jobs::Constants;
 use Class::Method::Modifiers;
 use OpenQA::Log qw(log_debug);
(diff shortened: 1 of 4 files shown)


lib/OpenQA/Schema/Result/JobGroups.pm
```diff
@@ -172,7 +173,9 @@ sub matches_nested {
 }
 
 # check the group comments for important builds
-sub _important_builds ($self) {
+sub important_builds ($self) {
+    if (my $cached = $self->{_important_builds}) { return $cached }
```
kalikiana, line 177: Should this be invalidated somewhere? At least I don't see it. Or do we assume they don't change after the fact?

Martchus: These result objects are never being invalidated automatically. One has to call discard_changes manually for that. I don't think we ever do that here so I avoided providing an override of that function to cover this.

kalikiana (approved):

kalikiana:
@mergifyio rebase

okurz (commented):
+1 did not find any errors but I would like others to review as well

PR #5148, opened 2023-05-22
Delete worker logs as well when cleaning up job logs

Martchus:
With this change the files `worker-log.txt` and
`worker_packages.txt` will be deleted as well when cleaning up job logs. Those files usually don't take that much disk space which is likely why they haven't been considered for now. However, even just for the sake of consistency it makes sense to delete them - so when looking at job results it is immediately clear that logs have been cleaned up.

Related ticket/note: https://progress.opensuse.org/issues/129244#note-24


mimi1vx (approved):

mimi1vx (approved):

mimi1vx (approved):

mimi1vx (approved):

okurz (approved):
I am not exactly clear about the distinction of the constants "result log", "result", "cleanup log" but the de-duplication of definitions is an obvious improvement even more so because the duplicate definition places were already showing problems in the past

PR #5151, opened 2023-05-23
containers: Rely on images from OBS in compose YAML

kalikiana:
Building the images locally can take a long time.

### DIFF
--- a/container/webui/docker-compose.yaml
+++ b/container/webui/docker-compose.yaml
@@ -2,7 +2,8 @@ version: '3.8'
 
 services:
   scheduler:
-    image: openqa_webui
+    image: &image registry.opensuse.org/devel/openqa/containers15.4/openqa_webui:latest
+    build: .
     volumes: &volumes
       - ./workdir/data/factory/iso:/data/factory/iso:z
       - ./workdir/data/factory/hdd:/data/factory/hdd:z
@@ -27,7 +28,7 @@ services:
       start_period: 10s
 
   websockets:
-    image: openqa_webui
+    image: *image
     build: .
     ports:
       - "9527"  # worker connection
@@ -47,7 +48,7 @@ services:
       start_period: 10s
 
   gru:
-    image: openqa_webui
+    image: *image
     build: .
     volumes: *volumes
     environment:
@@ -64,7 +65,7 @@ services:
       start_period: 10s
 
   livehandler:
-    image: openqa_webui
+    image: *image
     build: .
     ports:
       - "9528"  # handle live view
@@ -85,7 +86,7 @@ services:
       start_period: 5s
 
   webui:
-    image: openqa_webui
+    image: *image
     build: .
     ports:
       - "9526"
@@ -109,7 +110,7 @@ services:
       start_period: 5s
 
   webui_db_init:
-    image: openqa_webui
+    image: *image
     build: .
     ports:
       - "9526"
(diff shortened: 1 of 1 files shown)


kalikiana:
Note that the tests won't pass because of issues with download.opensuse.org - the only change here is specifying images from public registries.

Martchus (changes requested):
For consistency the worker docker compose config should be updated as well.

---

> I don't think it makes sense to have a docker-compose file in this directory and then use upstream containers which are not aligned with the source code of this repository directly

Are they really not aligned? To me it looks like the container image that @kalikiana specifies here is actually built from the very same `Dockerfile` that the so far local build would use. So I'd assume this just works as intended. This counts for all occurrences (except for `postgres` but this change looks also fine to me).

> I would like to not have another point where we need to update the Leap version in the URL

It would indeed be good to be able to avoid it. However, considering that one would normally just do an automatic replacement it likely doesn't really matter in practice.

> .github/workflows/compose.yml would not make sense building in every pull request when we use pre-build images

That's right. One might also still be able to test the container build locally (e.g. with local changes). So I guess that's something we should keep supporting. Supposedly it would be nice if one could make this configurable. Maybe this is possible in the same way as we allow to configure the number of replicas via `OPENQA_WEBUI_REPLICAS` with a default in `.env`? We could add an additional `.env-local` file one could source before invoking compose to do a local build.

okurz:
> For consistency the worker docker compose config should be updated as well.
> 
> > I don't think it makes sense to have a docker-compose file in this directory and then use upstream containers which are not aligned with the source code of this repository directly
> 
> Are they really not aligned? To me it looks like the container image that @kalikiana specifies here is actually built from the very same `Dockerfile` that the so far local build would use. So I'd assume this just works as intended. This counts for all occurrences (except for `postgres` but this change looks also fine to me).

Yes, images are built synchronously for every commit merged to master but that means that
1. the state is not the same as an open pull request refers to
2. the state might lag a bit behind current state of master

kalikiana:
> Yes, images are built synchronously for every commit merged to master but that means that
> 
> 1. the state is not the same as an open pull request refers to
> 2. the state might lag a bit behind current state of master

Perhaps to clarify here. The original motivation came from making it easier to spawn containers via compose. Building is slow even on a good connection.

I do agree with your points. Maybe we should just accept that it's slow by design and not suitable for getting started quickly.

kalikiana:
> It is also possible to provide `docker-compose.override.yaml` next to the `docker-compose.yaml` to achieve the same, the file is picked up automatically if exists.
> 
> ```yaml
> ---
> services:
>   scheduler:
>     image: &openqa_webui registry.opensuse.org/devel/openqa/containers15.4/openqa_webui:latest
> 
>   websockets:
>     image: *openqa_webui
> 
>   gru:
>     image: *openqa_webui
> 
>   livehandler:
>     image: *openqa_webui
> 
>   webui:
>     image: *openqa_webui
> 
>   webui_db_init:
>     image: *openqa_webui
> 
>   nginx:
>     image: &openqa_webui_lb registry.opensuse.org/devel/openqa/containers15.4/openqa_webui_lb:latest
> ```

Well, my point was that we may not need another file :-D

baierjan:
> Well, my point was that we may not need another file :-D

Sure, if the same can be done by adding / removing `--build` flag to the command line, it is the best solution. I am listing this just as an alternative (and it might be helpful, if we need to override something else for development/production).

Martchus (approved):

kalikiana:
I guess we didn't want this

PR #5156, opened 2023-05-24
Improve documentation about containerized setup

Martchus:
* State that the approach using Fedora-based images is not
  supported
* Mention the one-liners that are already present in the
  "Getting started" chapter
* Don't consider the setup proof-of-concept anymore; it
  basically works and users are supposed to report issues with
  it
* Mention how to configure the worker
* Mention where the config files from the example commands come
  from
* Link the "Running openQA workers" and "Triggering tests"
  sections for how to continue once the containers are running
* Mention how to invoke other tools within a container

---

@AdamWill We've also noticed that the documentation about Fedora-based container images and how it is currently interleaved with the documentation about "our" containers is supposedly a bit confusing for newcomers. So I suggest we at least move these parts to the very end of the "Containerized setup" chapter in its own section. Are you ok with that? If it is ok with you we could also remove it entirely from the documentation as it is not regularly updated anymore anyways and might not even work anymore.

### DIFF
--- a/docs/ContainerizedSetup.asciidoc
+++ b/docs/ContainerizedSetup.asciidoc
@@ -5,15 +5,15 @@
 :toclevels: 6
 :author: openQA developers
 
-This section describes two ways to deploy the containers for the openQA web UI
-and the workers.
+The installation guide already contains simple one-liners for starting
+the web UI and workers in the
+<<Installing.asciidoc#container_setup,Container based setup section>>.
 
-The first one describes how to deploy an openQA environment using containers
-with Fedora images or images built locally.
+This chapter describes how to deploy the containers for the openQA web UI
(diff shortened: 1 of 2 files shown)


baierjan (approved):

lruzicka:
@AdamWill I am not sure which documentation is meant here, but there was one written by Josef and Jan from the distant times in the history that probably has never worked correctly. I was unable to containerize openQA because my container experience is limited, I tried however, but I have certainly not updated the github documentation here.

AdamWill:
Ah, yeah, that's right. @Martchus practically speaking, you can probably just drop it for now. If Lukas and I (the current Fedora openQA team) ever do pick up the container stuff we can always re-contribute it.

PR #5158, opened 2023-05-24
container: Remove redundant "openqa" layer in helm chart directory layout

okurz:


kalikiana (approved):

baierjan:
The directories under the `charts` directory correspond to the charts name, ie https://github.com/os-autoinst/openQA/blob/master/container/helm/charts/openqa/Chart.yaml#L2 requires the Chart.yaml to be in the openqa sub-directory (or afaik it is the best practice I understood from https://helm.sh/docs/topics/chart_repository/).

We can remove the `helm` directory and put the charts into `container/charts/...` or even get rid of the `container/helm` and have `charts` beside `container` in the top level.

okurz:
Thx

PR #5162, opened 2023-05-27
Add development container for openQA

mimi1vx:
This container contains everything needed for development of openQA for use with distrobox/toolbox

### DIFF
--- a/container/devel/Dockerfile
+++ b/container/devel/Dockerfile
@@ -1,0 +1,21 @@
+# Specify the license of the container build description (see also the LICENSE file)
+# SPDX-License-Identifier: MIT
+# Define the names/tags of the container
+#!BuildTag: opensuse/openqa_devel:latest opensuse/openqa_devel:%PKG_VERSION% opensuse/openqa_devel:%PKG_VERSION%.%RELEASE%
+
+# hadolint ignore=DL3006
+FROM opensuse/tumbleweed
+
+# Define labels according to https://en.opensuse.org/Building_derived_containers
+# labelprefix=org.opensuse.openqa_devel
+LABEL org.opencontainers.image.title="openQA devel container"
+LABEL org.opencontainers.image.description="openQA development container"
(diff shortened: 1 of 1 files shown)


okurz (approved):

baierjan (approved):

container/devel/Dockerfile
```diff
@@ -0,0 +1,21 @@
...
+LABEL org.opencontainers.image.title="openQA devel container"
+LABEL org.opencontainers.image.description="openQA development container"
+LABEL org.opencontainers.image.version="%PKG_VERSION%.%RELEASE%"
+LABEL org.opensuse.reference="registry.opensuse.org/opensuse/openqa_devel:%PKG_VERSION%.%RELEASE%"
+LABEL org.openbuildservice.disturl="%DISTURL%"
+LABEL org.opencontainers.image.created="%BUILDTIME%"
+# endlabelprefix
+
+# hadolint ignore=DL3037
+RUN zypper in -y os-autoinst-devel openQA-devel vim nodejs-default git mc perl-Devel-REPL perl-Term-ReadKey && \
```
perlpunk, line 20: Why mc and vim?
I also have an [openqa devel container](https://github.com/perlpunk/mytoolboxes/blob/main/Dockerfile.openqa), where I additionally have curl, wget, jq, tmux, tmate, ack, etc., but I never use mc.
Maybe we should leave out such stuff and leave that to the users to build on top.

Martchus: I think including an editor makes sense and vim is likely the best option for a terminal-based one considering the preferences in our team. I assume the "mc" package contains Midnight Commander which I personally also find a bit strange to include.

mimi1vx: hmm, `jq`  and a more powerful variant of `grep` isn't bad idea for this : D

the primary motivation was a container for distrobox I could use without the need to install anything more to it:D

PR #5166, opened 2023-05-30
doc: Improve references for a quick installation

Martchus:
* Fix the broken reference to the bootstrapping section
* Mention the container-based setup section
* Related ticket: https://progress.opensuse.org/issues/129883


mimi1vx (approved):

kalikiana (approved):
Good catch! These are hard to spot sometimes.

PR #5167, opened 2023-05-30
Improve performance of server-side asset upload handling

Martchus:
* Do not use a sub process; this is not useful anymore because we start the web UI service via `… prefork … -c 1 …` to allow only one connection per process (so no other connections are impaired by blocking here)
* Remove event tracking that was only useful for a unit test to track coverage despite using a sub process
* Simplify the error handling and variable passing


Martchus:
Note that there's no particular issue for this. When going though the backlog I was just reminded on our conversation during the last workshop where we established that this code can be simplified. This will also slightly improve the performance although I would not expect wonders.

kraih (approved):

PR #5168, opened 2023-05-30
Simplify JavaScript code for adding/updating comments

Martchus:
* Avoid one level of indentation via an early return
* Use less jQuery, e.g. replace `editorForm.find('[name="text"]')` with just `form.text` which is much simpler
* Use `const`/`let` to make scope and constness of variables more clear
* Related ticket: https://progress.opensuse.org/issues/129946

### DIFF
--- a/assets/javascripts/comments.js
+++ b/assets/javascripts/comments.js
@@ -34,12 +34,16 @@ function renderCommentHeading(comment, commentId) {
   return heading;
 }
 
-function getXhrError(xhr, thrownError) {
+function getXhrError(jqXHR, textStatus, errorThrown) {
   try {
-    return JSON.parse(xhr.responseText).error || thrownError;
+    return JSON.parse(jqXHR.responseText).error || errorThrown || textStatus;
   } catch {
-    return thrownError;
+    return errorThrown || textStatus;
   }
+}
+
+function showXhrError(context, jqXHR, textStatus, errorThrown) {
+  window.alert(context + getXhrError(jqXHR, textStatus, errorThrown));
 }
 
 function updateNumerOfComments() {
@@ -62,9 +66,7 @@ function deleteComment(deleteButton) {
       $(deleteButton).parents('.comment-row, .pinned-comment-row').remove();
       updateNumerOfComments();
     },
-    error: (xhr, ajaxOptions, thrownError) => {
-      window.alert("The comment couldn't be deleted: " + getXhrError(xhr, thrownError));
-    }
+    error: showXhrError.bind(undefined, "The comment couldn't be deleted: ")
   });
 }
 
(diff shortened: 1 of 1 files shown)


Martchus:
Note that with this PR we can exclude jQuery from being the source of the sometimes failing comments test. It isn't likely to be the source of these failures but being able to exclude it is good and the use of jQuery made the code harder to read anyways.

assets/javascripts/comments.js
```diff
@@ -70,102 +70,98 @@ function deleteComment(deleteButton) {
...
+  const editorForm = $(form);
+  const url = form.dataset.putUrl;
+  const headingElement = editorForm.find('h4');
+  const markdownElement = editorForm.find('.markdown');
+  const markdown = markdownElement.html();
+  textElement.style.display = 'none';
+  markdownElement.show();
+  markdownElement.html('<em>Loading…</em>');
+  editorForm.find('[name="applyChanges"]').hide();
+  editorForm.find('[name="discardChanges"]').hide();
```
kalikiana, line 87: I suspect we can simplify those, too?

```suggestion
  form.applyChanges.hide();
  form.discardChanges.hide();
```

assets/javascripts/comments.js
```diff
@@ -70,102 +70,98 @@ function deleteComment(deleteButton) {
...
+    url: url,
+    method: 'PUT',
+    data: editorForm.serialize(),
+    success: function () {
+      // get updated markdown
+      $.ajax({
+        url: url,
+        method: 'GET',
+        success: function (response) {
+          const commentId = headingElement.find('[class="comment-anchor"]')[0].href.split('#comment-')[1];
```
kalikiana, line 98:
```suggestion
          const commentId = headingElement.find('.comment-anchor')[0].href.split('#comment-')[1];
```

assets/javascripts/comments.js
Martchus: Then it needed to be `.style.display = 'none'` instead of `.hide()` because we're not using jQuery anymore, right? I guess even then your suggestion makes sense, though.

assets/javascripts/comments.js
```diff
@@ -70,102 +70,98 @@ function deleteComment(deleteButton) {
...
+  textElement.style.display = 'none';
+  markdownElement.show();
+  markdownElement.html('<em>Loading…</em>');
+  editorForm.find('[name="applyChanges"]').hide();
+  editorForm.find('[name="discardChanges"]').hide();
+  $.ajax({
+    url: url,
+    method: 'PUT',
+    data: editorForm.serialize(),
+    success: function () {
```
kraih, line 92: You can probably use arrow functions here to be a little more efficient.

assets/javascripts/comments.js
Martchus: I've changed it to use the standard API like most other occurrences to avoid mixing this too much.

kalikiana (approved):

PR #5171, opened 2023-06-01
Build only on architectures with chromium

mimi1vx:


baierjan:
`s/aarchitectures/architectures/`

ggardet:
You added `jq` silently (no ref in commit message)

Martchus (changes requested):
I guess adding jq is a good idea. So rebase and change the commit message accordingly.

kalikiana:
Related to https://github.com/os-autoinst/openQA/pull/5171 I assume

perlpunk:
Can we also add these?  `wget curl ack zsh tmux tmate tree perl-doc`

kalikiana:
@mimi1vx With #5172 I think this PR is obsolete? Unless you want to repurpose it to "Make devel container a full development environment"?

> Can we also add these? `wget curl ack zsh tmux tmate tree perl-doc`

Usually I find this tricky since it adds various things most people may never use, and if we all contribute our favorite tools it becomes even bigger.
How about we instead document how to easily build your own container with the tools you like e.g. on OBS?

okurz (changes requested):

PR #5172, opened 2023-06-01
Build openQA devel container only on archs where chromium is built

ggardet:
currently x86_64 aarch64 riscv64


baierjan (approved):
https://github.com/os-autoinst/openQA/pull/5171

Martchus (approved):

PR #5163, opened 2023-05-27
single-instance container (poo#129883)

okurz:
* openqa-bootstrap: Speedup package installation
* openqa-bootstrap: Optionally configure SUSE-specifics and tests initialization
* openqa-bootstrap: Support systemd-less environments
* container: Handle potentially already existing certificates gracefully

PR #5173, opened 2023-06-01
Fix condition evaluation for skip variables

baierjan:
It is not clear from #5163 if skip_suse_tests also means skip openSUSE specific setup. Call fetchneedles by default to avoid test regression.

Reference: https://progress.opensuse.org/issues/130201

### DIFF
--- a/script/openqa-bootstrap
+++ b/script/openqa-bootstrap
@@ -53,7 +53,7 @@ setup=/usr/share/openqa/script/configure-web-proxy
 sed -i -e 's/#*.*method.*=.*$/method = Fake/' /etc/openqa/openqa.ini
 
 
-if [ -n "$skip_suse_specifics" ] && ping -c1 download.suse.de. && (! rpm -q ca-certificates-suse) ; then
+if [ -z "$skip_suse_specifics" ] && ping -c1 download.suse.de. && (! rpm -q ca-certificates-suse) ; then
     # add internal CA if executed within suse network
     if ! zypper info ca-certificates-suse | grep -q ':' ; then
         # add suse ca repo if needed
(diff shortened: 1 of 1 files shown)


script/openqa-bootstrap
```diff
@@ -92,6 +92,8 @@ if [ -n "$skip_suse_tests" ]; then
         fi
         chown -R $dbuser: /var/lib/openqa/tests/opensuse/products/sle/needles
     fi
+else
+    /usr/share/openqa/script/fetchneedles
```
kalikiana, line 96: I actually thought **skip_suse_tests** means both openSUSE and SUSE needles since [poo#129883](https://progress.opensuse.org/issues/129883) is about a generic openQA container that is not specific to any test distribution.

baierjan: Commit message in https://github.com/os-autoinst/openQA/pull/5163/commits/a370919c26ba3aed543227c9bf191b780f188223 suggests, it shoul deal with SUSE specifics...

Martchus (approved):

script/openqa-bootstrap
baierjan: Also, `os-autoinst-distri-opensuse-deps` is installed into the container so I would guess the container was meant to test os-autoinst-distri-opensuse and hence fetching the needles makes sense.

asdil12: I think that `-n` is used wrong.
```
       -n STRING
              the length of STRING is nonzero
```

So currently only of `$skip_suse_tests` is set to eg `1`, the suse tests are run.
So the whole logic seems to be inverted.

asdil12: I think, @okurz meant to use `-z` instead of `-n` in his original PR.

baierjan: Yes, that's also true.

asdil12: Same for this part:
```bash
if [ -n "$skip_suse_specifics" ] && ping -c1 download.suse.de. && (! rpm -q ca-certificates-suse) ; then
```
This should be `-z` as well.

kalikiana: For the record, feel free to go for the most straight-forward fix so we get tests to pass again. I commented on [poo#129883](https://progress.opensuse.org/issues/129883#note-8) regarding the original motivation.

Martchus (approved):

asdil12 (approved):

Martchus:
I've just re-triggered the failing UI tests, it was just https://progress.opensuse.org/issues/129946 again.

PR #5175, opened 2023-06-01
Use needles from correct ref of NEEDLES_DIR

sdclarke:
If the CASEDIR variable is a git repo URL with a fragment specifying the ref of the repository, we attempt to use it as the source of needles when viewing needle diffs in test results.

If the TEST_GIT_SHA variable is set, we use that instead as it should be the most accurate way to determine exactly which commit was used

https://progress.opensuse.org/issues/162035


lib/OpenQA/WebAPI/Controller/Step.pm
```diff
@@ -91,6 +94,45 @@ sub view ($self) {
...
+    my $job = shift;
+    my $tmp_needles_dir = $job->needle_dir;
+    my $real_needle_dir = realpath($tmp_needles_dir) // $tmp_needles_dir;
+    if (my $casedir = $job->settings->single({key => 'CASEDIR'})) {
+        my $casedir_refspec = Mojo::URL->new($casedir->value)->fragment;
+        eval {
+            my $vars_json = Mojo::File->new($job->result_dir(), 'vars.json')->slurp;
+            my $vars = decode_json($vars_json);
+            $casedir_refspec = $vars->{TEST_GIT_HASH};
+        };
```
Martchus, line 107: I suppose the case when we cannot determine `$casedir_refspec` should be handled as soon as possible:

```suggestion
        };
        return $real_needle_dir if !$casedir_refspec;
```

Otherwise there's just no need to invoke `git` at all.

lib/OpenQA/WebAPI/Controller/Step.pm
```diff
@@ -91,6 +94,45 @@ sub view ($self) {
     $self->viewimg;
 }
 
+sub _checkout_needledir_or_default {
+    my $job = shift;
+    my $tmp_needles_dir = $job->needle_dir;
+    my $real_needle_dir = realpath($tmp_needles_dir) // $tmp_needles_dir;
```
Martchus, line 100: `return $tmp_needles_dir unless my $casedir = $job->settings->single({key => 'CASEDIR'});` would save us one level of indentation.

lib/OpenQA/WebAPI/Controller/Step.pm
```diff
@@ -91,6 +94,45 @@ sub view ($self) {
     $self->viewimg;
 }
 
+sub _checkout_needledir_or_default {
+    my $job = shift;
```
Martchus, line 98: Please use signatures in new code:
```suggestion
sub _checkout_needledir_or_default ($job) {
```

lib/OpenQA/WebAPI/Controller/Step.pm
```diff
@@ -91,6 +94,45 @@ sub view ($self) {
     $self->viewimg;
 }
 
+sub _checkout_needledir_or_default {
```
Martchus, line 97: This whole function should likely return early unless this feature is turned on in the settings. I imagine that the performance impact and additional storage usage are not generally neglectable and our openQA instance is at its limits as-is. Besides, the code might not work depending on the state of the Git checkout which might cause problems in production use (e.g. the Git checkout might currently be locked because some other operation is performed or it is in an intermediate state because currently a user saves a needle via the needle editor or simply the hard-coded remote names don't match).

lib/OpenQA/WebAPI/Controller/Step.pm
```diff
@@ -91,6 +94,45 @@ sub view ($self) {
...
+            my $new_path = "/tmp/$needle_dir_basename/worktrees/$casedir_refspec";
+            if (!-d $new_path) {
+                my $git_wt_cmd = "git -C $real_needle_dir worktree";
+                # See how many worktrees we have already checked out, if there are too many, remove them all
+                make_path("/tmp/$needle_dir_basename/worktrees") unless (-d "/tmp/$needle_dir_basename/worktrees");
+                opendir my $dh, "/tmp/$needle_dir_basename/worktrees";
+                my $num_entries = () = readdir($dh);
+                if ($num_entries > 8) {
+                    qx{$git_wt_cmd list | grep /tmp | awk '{ print \$1 }' | xargs -n1 $git_wt_cmd remove &>/dev/null};
+                }
```
Martchus, line 126: That is not going to cut it in our production use. The number 8 needed to be configurable and even then we must really avoid removing worktrees that are still in use. It would also make more sense to do the cleanup as a background task (we use Minion jobs for this) instead in the context of serving an HTTP response (where we can still *trigger* the cleanup).

lib/OpenQA/WebAPI/Controller/Step.pm
```diff
@@ -91,6 +94,45 @@ sub view ($self) {
...
+sub _checkout_needledir_or_default {
+    my $job = shift;
+    my $tmp_needles_dir = $job->needle_dir;
+    my $real_needle_dir = realpath($tmp_needles_dir) // $tmp_needles_dir;
+    if (my $casedir = $job->settings->single({key => 'CASEDIR'})) {
+        my $casedir_refspec = Mojo::URL->new($casedir->value)->fragment;
+        eval {
+            my $vars_json = Mojo::File->new($job->result_dir(), 'vars.json')->slurp;
+            my $vars = decode_json($vars_json);
+            $casedir_refspec = $vars->{TEST_GIT_HASH};
```
Martchus, line 106: The needles might be in a separate repository, e.g. this is the case for the openSUSE test distribution and thus our main production use case. This means the `TEST_GIT_HASH` and the `CASEDIR`-fragment are not relevant. The `NEEDLES_GIT_REFSPEC` and `NEEDLES_DIR`-fragment would be the relevant variables which should be checked before. Note that there is also `TEST_GIT_REFSPEC` which could be checked as well.

Maybe it would make sense to introduce `NEEDLES_GIT_HASH` so we could also do the same lookup for needles.

Martchus (changes requested):
I don't think creating another worktree for every rev we might encounter scales very well. It would make more sense to use `git show …` to obtain a specific needle for a specific rev.

Note that we've been pondering about this problem ourselves and coming up with an approach that scales well enough for being usable on our production instances is hard (e.g. see https://progress.opensuse.org/issues/56789. https://progress.opensuse.org/issues/91905 and other related tickets).

sdclarke:
I appreciate the feedback!

> I don't think creating another worktree for every rev we might encounter scales very well. It would make more sense to use `git show …` to obtain a specific needle for a specific rev.

I agree that it's unlikely to scale well, using `git show` seems like a good idea, I'll take a look into that and see if I can come up with something functional.

Martchus:
By the way, I think the most important point of my review was that this should be put behind a configuration option and be disabled by default - at least until we are confident this works well in all kinds of instances (and currently we are likely pretty far from that).

sdclarke:
> By the way, I think the most important point of my review was that this should be put behind a configuration option and be disabled by default - at least until we are confident this works well in all kinds of instances (and currently we are likely pretty far from that).

Yes, definitely, this patch is very much a hack that was put together to get a quick and nasty solution to the problem for a specific use case, I realise there's an awful lot of change required for it to be anything like production-ready.

sdclarke:
I've changed this to make use of `git show` to get the contents of the needle files from the correct SHA of the repository.

It now also has a config option to prevent the feature being used by default.

The hacky clean up has been removed for now, so another method of cleaning up will be needed.

lib/OpenQA/Task/Needle/RemoveVersions.pm
```diff
@@ -0,0 +1,38 @@
+# Copyright 2019 SUSE LLC
```
okurz, line 1: Please in new files leave out the years, i.e. just

```suggestion
# Copyright SUSE LLC
```

lib/OpenQA/Task/Needle/RemoveVersions.pm
```diff
@@ -0,0 +1,38 @@
...
+use File::stat;
+
+my $minimum_retention_time;
+
+sub register ($self, $app, $job) {
+    $minimum_retention_time = $app->config->{'scm git'}->{minimum_needle_retention_time} * 60 // 30 * 60;
+    $app->minion->add_task(remove_needle_versions => sub ($job) { _remove_needle_versions($app, $job) });
+}
+
+sub _remove_needle_versions ($app, $job) {
```
kalikiana, line 17: Could you add a unit test for this? At least something basic since this is a fair bit of new logic.

Btw it would also be good to mention this in the PR title

sdclarke: I've had a go at adding a unit test for this in `t/14-grutasks.t`, I hope that's sufficient for now as this behaviour won't be active by default anyway.

kalikiana (approved):

lib/OpenQA/WebAPI/Controller/Step.pm
```diff
@@ -91,6 +92,31 @@ sub view ($self) {
...
+    my $needles_ref = $needles_url->fragment;
+    eval {
+        my $vars_json = Mojo::File->new($job->result_dir(), 'vars.json')->slurp;
+        my $vars = decode_json($vars_json);
+        $needles_ref = $vars->{NEEDLES_GIT_HASH};
+    };
+    chomp($needles_ref);
+    return undef unless $needles_ref;
+    qx{git -C "$needle_dir" fetch --depth 1 origin "$needles_ref" &>/dev/null};
+    $needles_ref = qx{git -C "$needle_dir" rev-parse FETCH_HEAD};
```
Martchus, line 113: Errors should be handled here (when Git returns a non-zero exist status).

lib/OpenQA/WebAPI/Controller/Step.pm
```diff
@@ -91,6 +92,31 @@ sub view ($self) {
...
+    my $needle_dir_basename = basename(dirname($needle_dir));
+    my $new_path = "/tmp/needle_dirs/$needle_dir_basename/$needles_ref/needles";
+    make_path($new_path);
```
Martchus, line 116: The error handling with `File::Path::make_path` is rather difficult: https://metacpan.org/pod/File::Path#ERROR-HANDLING

I think we shouldn't just log to stderr here (which would be the default). Maybe it is better to use `Mojo::File::make_path` instead and handle the exception it might throw as needed.

script/openqa-enqueue-needle-versions-cleanup
```diff
@@ -0,0 +1,2 @@
+#!/bin/sh -e
+exec "$(dirname "$0")"/openqa eval -m production -V 'app->gru->enqueue(remove_needle_versions => [], {priority => 5, ttl => 172800, limit => 1})' "$@"
```
Martchus, line 2: We also need systemd service and timer units and `openqa-webui.service` needs to depend on the timer unit.

t/14-grutasks.t
```diff
@@ -401,6 +402,19 @@ subtest 'limit_results_and_logs gru task cleans up logs' => sub {
     ok !-e $log_file_for_groupless_job, 'log file for groupless job got cleaned';
 };
 
+subtest 'remove_needle_versions gru task cleans up needle versions' => sub {
+    # Create a temporary needle file older than the configured expiry time (defaults to 30 minutes)
+    my $temp_needle_path = '/tmp/needle_dirs/test_repo/branch/needles';
+    File::Path->make_path($temp_needle_path);
+    my @needle_files = ($temp_needle_path . 'needle.png', $temp_needle_path . 'needle.json');
+    my $ref = File::Touch->new(atime => time - (30 * 60 + 1));
```
Martchus, line 410: You could also use time constants here.

lib/OpenQA/WebAPI/Controller/File.pm
```diff
@@ -29,11 +29,23 @@ sub needle ($self) {
 
     # make sure the directory of the file parameter is a real subdir of testcasedir before
     # using it to find needle subdirectory, to prevent access outside of the zoo
-    if ($jsonfile && !is_in_tests($jsonfile)) {
+    # Also allow the json file to be under /tmp
```
Martchus, line 32: Again, this path should not be hard-coded.

lib/OpenQA/Utils.pm
```diff
@@ -254,7 +255,27 @@ sub is_in_tests {
...
+        }
+        my $temp_json_path = $temp_needles_dir . '/' . $relative_needle_path;
+        open my $temp_json_fh, '>', $temp_json_path;
+        my $jsonfile = qx{git -C $needles_dir show $needles_ref:./$relative_needle_path};
+        print $temp_json_fh $jsonfile;
+        close $temp_json_fh;
+        my $png_name = dirname($relative_needle_path) . '/' . basename($relative_needle_path, '.json') . '.png';
+        my $temp_png_path = $temp_needles_dir . '/' . $png_name;
+        open my $temp_png_fh, '>', $temp_png_path;
+        my $pngfile = qx{git -C $needles_dir show $needles_ref:./$png_name};
```
Martchus, line 274: We should handle the case when this call fails (Git returns a non-zero exit status or cannot be invoked at all).

Note that error handling for `qx{…}` calls is a bit involved: https://perldoc.perl.org/perlop#Quote-Like-Operators
So you might want to introduce a helper for this (maybe we already have one).

Martchus (changes requested):
The error handling should be improved. I commented on the most important places.

Also, see my other inline-comments.

foursixnine:
@Martchus sorry I thought you hadn't reviewed and clicked again :) (just noticed the thingie, @sdclarke mentioned this during our chat at FOSDEM)

okurz (changes requested):
https://app.circleci.com/pipelines/github/os-autoinst/openQA/12874/workflows/bc1b1c0b-a2fa-4362-9269-22b44265b395/jobs/119993 looks very much related

Martchus (dismissed):
The error handling should be improved as mentioned in my last review.

The cleanup code also needs further improvement, see https://github.com/os-autoinst/openQA/pull/5175/files#r1480067530.

Martchus:
@sdclarke You're not actively working on this, right? I would take over as we'd actually also like to have this feature ourselves (see https://progress.opensuse.org/issues/154783). I would keep your commits as-is (except for rebasing) and add my own commit on top of them.

sdclarke:
> @sdclarke You're not actively working on this, right? I would take over as we'd actually also like to have this feature ourselves (see https://progress.opensuse.org/issues/154783). I would keep your commits as-is (except for rebasing) and add my own commit on top of them.

You're more than welcome to take over, I haven't had time to finish this off recently unfortunately.

Martchus:
I've just pushed an update which should resolve all my previous review comments. This leaves:

* Add UI tests for the test details page
* Make this optional for the needle editor because there we probably also want to be able to create new needles based on the latest version of existing needles
* Add UI tests for the needle editor that cover the displaying of needles/areas as well as saving new needles which should continue to work as before
* Make the cleanup space-aware or generally a bit more clever

Martchus:
See https://progress.opensuse.org/issues/154783#note-10 for the most recent update on this.

neillydun:
Is this currently being worked on, or is there anything that could be done to help get it merged?

okurz:
There are at least three points necessary: 
* Rebase
* Fix rebase/merge conflicts
* Add tests for missing coverage

Martchus:
Also see https://progress.opensuse.org/issues/154783#note-10 and my subsequent comments in that ticket for what's left.

Martchus:
The minimum effort we need to put into this PR to be mergable is:

1. Resolve conflicts
2. Ensure test coverage checks pass (I suppose UI tests are missing)

If anyone wants to work on it (e.g. @sdclarke or someone from our team), drop a comment here for better coordination. We now also have https://progress.opensuse.org/issues/162035 for tracking this.

okurz:
@sdclarke I saw that you pushed an update. Can you state what changes we should expect?

sdclarke:
> @sdclarke I saw that you pushed an update. Can you state what changes we should expect?

This push was just fixing the conflicts, I'm trying to look for the correct places to improve the test coverage but I'm having some trouble working out where the changes are needed. Any pointers in that regard would be appreciated.

Assuming I can deal with the test coverage issues, would that be enough to get this merged for now? From my perspective it would be good to get this over the line as it works well enough for our purposes, and it is disabled by default in the config. We can potentially revisit it later to address any outstanding points.

okurz:
> > @sdclarke I saw that you pushed an update. Can you state what changes we should expect?
> 
> This push was just fixing the conflicts, I'm trying to look for the correct places to improve the test coverage but I'm having some trouble working out where the changes are needed. Any pointers in that regard would be appreciated.

As you are also adding new modules you could create new test modules like
t/15-shared-plugin-gru.t
which are basically unit tests with mocking all external dependencies.

> Assuming I can deal with the test coverage issues, would that be enough to get this merged for now?

Yes. I already asked the team to support for which I created ticket https://progress.opensuse.org/issues/162035 but unfortunately no one picked it up yet.

Martchus:
> Assuming I can deal with the test coverage issues, would that be enough to get this merged for now? From my perspective it would be good to get this over the line as it works well enough for our purposes, and it is disabled by default in the config. We can potentially revisit it later to address any outstanding points.

@sdclarke As stated in https://github.com/os-autoinst/openQA/pull/5175#issuecomment-2165130042 the answer is no - nothing else needs to be done except further problems are found in PR review. However, I'd strongly recommend looking into point 2 of [this note](https://progress.opensuse.org/issues/154783#note-10) because in my opinion it makes not much sense to enable it on a production instance without it.

lib/OpenQA/WebAPI/Controller/Step.pm
```diff
@@ -91,6 +94,34 @@ sub view ($self) {
...
+    my $needles_url = Mojo::URL->new($needles_dir_var->value);
+    return undef unless $needles_url->scheme;
+    my $needles_ref = $needles_url->fragment;
+    eval {
+        my $vars = decode_json(path($job->result_dir, 'vars.json')->slurp);
+        $needles_ref = $vars->{NEEDLES_GIT_HASH} if ref $vars eq 'HASH';
+    };
+    chomp $needles_ref;
+    return undef unless $needles_ref;
+    return undef unless run_cmd_with_log ['git', '-C', $needle_dirs, 'fetch', '--depth', 1, 'origin', $needles_ref];
```
perlpunk, line 116: @sdclarke We have not forgotten about this.

We are currently working on some git features and fixes, and I just wanted to note that we have a `git_auto_clone` feature now that will fetch any requested ref on the webui (branch or sha) in a minion job before the job starts running. (Note that it only works if it is the same origin and not some fork.)

That would make this `fetch` here not necessary anymore and also avoid the need for a lockfile.
The sha should always be in `NEEDLES_GIT_HASH` so the `rev-parse` also seems unnecessary.

Would you consider to use the `git_auto_clone` feature in order to not have to implement that part here?

I'm not sure if there are any cases which `git_auto_clone`  wouldn't cover. @Martchus said there might be.

perlpunk:
Just to clarify a few things, I'm currently trying to understand the feature.

> If the CASEDIR variable is a git repo URL with a fragment specifying the ref of the repository, we attempt to use it as the source of needles when viewing needle diffs in test results.
> 
> If the TEST_GIT_SHA variable is set, we use that instead as it should be the most accurate way to determine exactly which commit was used

The title is talking about NEEDLES_DIR, the code mentions NEEDLES_GIT_SHA, but the PR description talks about CASEDIR and TEST_GIT_SHA.
Of course there can be needles in a CASEDIR if the repo has a `needles` dir itself. I just found it a bit confusing.

I was also wondering: Even if you don't specify CASEDIR/NEEDLES_DIR, the feature is about showing the correct ref from the time the test was running, even if there was a newer version in git meanwhile, right?

Because we have this [ticket](https://progress.opensuse.org/issues/162035), but I can't find a clear description of the story, so no acceptance criteria, so I have to figure out out from the code.

lib/OpenQA/WebAPI/Controller/Step.pm
Martchus: Note that `git_auto_clone` only helps if a Git-URL is specified as `NEEDLES_DIR` (or `CASEDIR` which is relevant if needles are part of the tests).

Even then it is not guaranteed to fetch the refs the test actually runs on. The test might specify a branch name (or no branch which means "default branch") and when the test is finally executed on the worker host the ref that branch points to might differ from the ref checked out by the web UI at the time the test was scheduled. 

Additionally, we probably want to have some cleanup for the fetching done by the `git_auto_clone` feature. So the ref might have also already been cleaned up.

So I think we need that fetch in any case.

---

> That would make this fetch here not necessary anymore and also avoid the need for a lockfile.

I don't think we need a lockfile here. We can probably arrange it so that it would not matter if any other commands are executed in the middle. (The use of `FETCH_HEAD` definitely needs to go way but it can probably be replaced with whatever we specify on the `fetch` command.)

> The sha should always be in NEEDLES_GIT_HASH so the rev-parse also seems unnecessary.

That's true. So we could (and probably should) get rid of it. If one specifies a branch name via `NEEDLES_DIR` the `NEEDLES_GIT_HASH` still has the concrete sha.

perlpunk: > Note that `git_auto_clone` only helps if a Git-URL is specified as `NEEDLES_DIR` (or `CASEDIR` which is relevant if needles are part of the tests).

* The PR description says that it is about CASEDIR/NEEDLES_DIR being set.
* And if it is not set, there is no certain ref to fetch.
* In the near future, `git_auto_clone` will also automatically update clones if CASEDIR/NEEDLES_DIR is empty. It will just update the checkout branch in the checked out directory.

And this is all we need, no?
What else needs to be fetched that git_auto_clone does not fetch?

The only thing could be forks, but as far as I can see this PR also doesn't handle forks.


> Even then it is not guaranteed to fetch the refs the test actually runs on. The test might specify a branch name (or no branch which means "default branch") and when the test is finally executed on the worker host the ref that branch points to might differ from the ref checked out by the web UI at the time the test was scheduled.

What's checked out on the webui does not matter. The important thing is that it has been fetched at some point, so we can do a `git show` on it, right?

> Additionally, we probably want to have some cleanup for the fetching done by the `git_auto_clone` feature. So the ref might have also already been cleaned up.

Yes. I'm not sure what git already offers there and how to know what we can cleanup.

Martchus: > What's checked out on the webui does not matter. The important thing is that it has been fetched at some point, so we can do a git show on it, right?

I meant "fetched by the web UI". And I meant the following sequence of events:

1. A test is scheduled pointing to a branch, e.g. just the default branch.
2. The web UI fetches the latest ref of that branch because `git_auto_clone = 1` is configured.
3. Someone pushes additional commits to that branch or even force-pushes there.
4. The test is finally assigned to a worker and that worker now fetches a different ref than the web UI in step 2 so `NEEDLES_GIT_HASH` points to a ref that is not yet checked out (despite `git_auto_clone = 1`).
5. Someone wants to display the candidate needles for that job. We need to fetch the missing ref.

And this is not very contrived. That someone specifies a branch where many commits happen (e.g. the default branch of some repo with many contributors) is probably a common use case. That it can take between the scheduling of a job and the time it actually gets executed is also something we need to expect.

kalikiana:
> Because we have this [ticket](https://progress.opensuse.org/issues/162035), but I can't find a clear description of the story, so no acceptance criteria, so I have to figure it out from the code.

@sdclarke Hey! 👋 Did you have a chance to look at the ticket description and the acceptance criteria? To be sure we're on the same page now on what's being implemented here.

kalikiana:
> @sdclarke Hey! 👋 Did you have a chance to look at the ticket description and the acceptance criteria? To be sure we're on the same page now on what's being implemented here.

@sdclarke ping? In case you missed the previous one 🙃

sdclarke:
Hi all, sorry for the lack of updates, I haven't had time to follow up on this lately.


> > Because we have this [ticket](https://progress.opensuse.org/issues/162035), but I can't find a clear description of the story, so no acceptance criteria, so I have to figure it out from the code.
> 
> @sdclarke Hey! 👋 Did you have a chance to look at the ticket description and the acceptance criteria? To be sure we're on the same page now on what's being implemented here.

Looking at the ticket, the acceptance criteria look to match what I was trying to achieve.




> The title is talking about NEEDLES_DIR, the code mentions NEEDLES_GIT_SHA, but the PR description talks about CASEDIR and TEST_GIT_SHA. Of course there can be needles in a CASEDIR if the repo has a `needles` dir itself. I just found it a bit confusing.

I think the title became out of sync with the description after an initial round of reviews, when I first wrote it the workflow I was using had the needles directory within the CASEDIR so I didn't initially consider them separately. I subsequently used the NEEDLES_DIR in this PR but must have missed the description.

> I was also wondering: Even if you don't specify CASEDIR/NEEDLES_DIR, the feature is about showing the correct ref from the time the test was running, even if there was a newer version in git meanwhile, right?

The original problem we were having that inspired this was that the web UI might not have any version of a particular needle, as we had a workflow which involved testing new needles in branches of the tests/needles repository before merging them to the main branch.

But ultimately, yes your description of the problem this PR is attempting to solve is correct, I based it on CASEDIR/NEEDLES_DIR as that was how we used it, and it provided the path of least resistance to doing what we needed.

As mentioned I don't have time to work on this for now, once again sorry for my delayed response.

systemd/openqa-enqueue-needle-ref-cleanup.timer
```diff
@@ -0,0 +1,9 @@
+[Unit]
+Description=Enqueues a needle refs task for openQA every day.
+
+[Timer]
+OnCalendar=hourly
```
b10n1k, line 5: i wonder if the description is accurate? OnCalendar says hourly

etc/openqa/openqa.ini
```diff
@@ -117,6 +117,10 @@
 #git_auto_clone = yes
 ## Experimental - Ensure a git update of all test code and needles
 #git_auto_update = no
+# whether openQA should attempt to display needles of the correct version in the web UI
```
b10n1k, line 120: I have a feeling that the comment is a bit _cryptic_ from users' side. What does it mean `attempt`? Also `correct version` I think is not very explanatory. I am talking from the POV of someone who do not know much about the openQA or a new user.

Martchus: I don't think we'll be able to explain this feature within the config file. I think this comment is good enough for people trying to enable this feature after reading documentation about it elsewhere. So we could document this feature as part of the normal openQA documentation. However, I'd be fine keeping this as an almost undocumented "expert feature" for now.

t/14-grutasks.t
```diff
@@ -418,9 +417,16 @@ subtest 'limit_temp_needle_refs task cleans up temp needle refs exceeding retent
     $temp_dir->child($_)->make_path for qw(ref1 ref2);
     my @old_needle_files = ("$temp_dir/ref1/needle_old.png", "$temp_dir/ref1/needle_old.json");
     my @new_needle_files = ("$temp_dir/ref2/needle_new.png", "$temp_dir/ref2/needle_new.json");
-    my $now = time;
-    File::Touch->new(time => $now - (120 * ONE_MINUTE + 1))->touch(@old_needle_files);
-    File::Touch->new(time => $now + ONE_MINUTE)->touch(@new_needle_files);
+    my $old_timestamp = time - (120 * ONE_MINUTE + 1);
+    my $new_timestamp = time - ONE_MINUTE + 1;
```
Martchus, line 421: We could avoid repeated calls to `time`.

etc/openqa/openqa.ini
okurz: Yeah, maybe we can still change the word "correct". Who wouldn't want "correct versions"? :)

kalikiana: How about

```suggestion
# whether openQA should display needles from different branches in the web UI
```

Martchus: This is not about the branch that was used (branches are moving targets) but really about the exact version that was used (not a moving target). This lookup will probably also be best effort. So I find the version of this comment as it is right now much better than the suggested change.

> Who wouldn't want "correct versions"? :)

Someone who doesn't want to pay the price for it. We could state the disadvantages here, especially that the feature is still experimental.

okurz: So rather say "exact version" and state the drawbacks

Wabri:
I've rebased this pr with master #6097 I've setted my pr as `not-ready` for now and wait until I understand what is the current state of this feature/pr!

etc/openqa/openqa.ini
kalikiana: Pondering on this with @Wabri trying to make it clearer. The commit or sha is what we want here (as opposed to branch) most likely?

```suggestion
# whether openQA should attempt to display needles according to the original commits in the web UI
```

etc/openqa/openqa.ini
```diff
@@ -115,12 +115,16 @@
 #do_push = no
 ## whether to clone CASEDIR or NEEDLES_DIR on the web UI if that var points to a Git repo
 #git_auto_clone = yes
-## enable automatic updates of all test code and needles managed via Git (still experimental, currently still breaks scheduling parallel clusters)
-#git_auto_update = no
+## enable automatic updates of all test code and needles managed via Git
+#git_auto_update = yes
```
okurz, line 119: this change does not look like it should be in this PR

okurz (changes requested):

etc/openqa/openqa.ini
Wabri: You can find the original commit here: https://github.com/os-autoinst/openQA/commit/e466e05e548f58bcf2303c2d9c29e6904063eb36#diff-b323be01fab1562900d8df608b80822e7ba6ca5d54dc745ca778315adbabe065R119

Now is committed with https://github.com/os-autoinst/openQA/pull/5175/commits/5ee9dd151e20ce0531d27dfd924d21e243055974

Wabri: It's coming from this #6039

okurz: That is a merged PR so the change shouldn't show up here in a diff. I doubt the default value has actually changed in this PR and moreover it definitely should not. What should a pull request about "use needles from correct ref" have to do with changing the default value of git_auto_update?

Martchus: This change was added to commit https://github.com/os-autoinst/openQA/commit/5ee9dd151e20ce0531d27dfd924d21e243055974 and that shouldn't have been the case because it has nothing to do with Git auto update. Note that on current master this setting is also 'no' and none of the commits of this PR should change that (as this PR is not about the Git auto update feature).

Wabri: I'll do a rebase interactive and replace with the version in the master.

lib/OpenQA/Task/Needle/LimitTempRefs.pm
```diff
@@ -0,0 +1,39 @@
...
+sub register ($self, $app, $job) {
+    $retention = $app->config->{'scm git'}->{temp_needle_refs_retention} * ONE_MINUTE;
+    $app->minion->add_task(limit_temp_needle_refs => sub ($job) { _limit($app, $job) });
+}
+
+sub _limit ($app, $job) {
+    my $ensure_task_retry_on_termination_signal_guard = OpenQA::Task::SignalGuard->new($job);
+
+    return $job->finish({error => 'Another job to remove needle versions is running. Try again later.'})
+      unless my $guard = $app->minion->guard('limit_needle_versions_task', 7200);
```
okurz, line 25: Please use constants from Time::Second here as well

t/14-grutasks.t
```diff
@@ -410,6 +411,38 @@ subtest 'limit_results_and_logs gru task cleans up logs' => sub {
...
 };
 
+subtest 'limit_temp_needle_refs task cleans up temp needle refs exceeding retention' => sub {
+    my $temp_dir = path(OpenQA::Needles::temp_dir);
+    is $temp_dir, 't/data/openqa/webui/cache/needle-refs', 'needle temp dir determined as expected';
+    $temp_dir->child($_)->make_path for qw(ref1 ref2);
+    my @old_needle_files = ("$temp_dir/ref1/needle_old.png", "$temp_dir/ref1/needle_old.json");
+    my @new_needle_files = ("$temp_dir/ref2/needle_new.png", "$temp_dir/ref2/needle_new.json");
+    my $now = time;
+    my $old_timestamp = $now - (120 * ONE_MINUTE + 1);
```
okurz, line 421:
```suggestion
    my $old_timestamp = $now - (2 * ONE_HOUR + 1);
```

t/config.t
```diff
@@ -69,6 +69,8 @@ subtest 'Test configuration default modes' => sub {
             git_auto_clone => 'yes',
             git_auto_update => 'no',
             git_auto_update_method => 'best-effort',
+            checkout_needles_sha => 'no',
+            temp_needle_refs_retention => 120,
```
okurz, line 73:
```suggestion
            temp_needle_refs_retention => 2 * ONE_MINUTE,
```

okurz (changes requested):

lib/OpenQA/Task/Needle/LimitTempRefs.pm
Wabri: Updated with 2*ONE_HOUR

Wabri:
@okurz done, do you want to close this pull request and create more pr with different purposes?

okurz:
> @okurz done, do you want to close this pull request and create more pr with different purposes?

No. I meant you could create another PR with only some parts of this PR which could be merged more easily and then rebase this PR on the other. But only do that if it helps you.

Martchus (changes requested):

t/21-needles.t
```diff
@@ -168,4 +177,42 @@ subtest 'handling relative paths in update_needle' => sub {
...
+    $controller->app->config->{'scm git'} = {checkout_needles_sha => 'yes'};
+    ($needle_dirs, $needles_var) = $controller->_determine_needles_dir_for_job($job);
+    like($needle_dirs, qr|t/data/openqa/share/tests/fedora/needles$|, 'needles dirs is defined');
+    is($needles_var->value, 'https://something#fragment', 'needles var is defined');
+};
+
+subtest '_create_tmpdir_for_needles_refspec' => sub {
+    my $sandbox = tempdir(CLEANUP => 1);
+    my $vars_json = path($sandbox, 'vars.json');
+    $vars_json->spew('{"NEEDLES_GIT_HASH": "0123456789abcdef0123456789abcdef01234567"}');
```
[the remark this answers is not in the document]
Wabri: I prefer this over the persisent data on the repository because it can be less prone to changes in the repository and more general overall

lib/OpenQA/WebAPI/Controller/Step.pm
```diff
@@ -91,6 +118,63 @@ sub view ($self) {
     $self->viewimg;
 }
 
+=head2 _determine_needles_dir_for_job
+
+my ($needle_dirs, $needles_var) = $self->_determine_needles_dir_for_job($job);
+
+Get the directory where needle files are stored for a given job.  
+It checks whether the Git-based needle management is enabled.
+If enabled, it fetches the path to the needles directory and for NEEDLES_DIR or CASEDIR.
```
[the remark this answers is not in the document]
Wabri: Thanks for the feedback. Also, do we want to leave this docs or remove them?

t/21-needles.t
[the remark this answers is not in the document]
Martchus: No, I'm fine with it and have no preference.

t/21-needles.t
```diff
@@ -168,4 +178,53 @@ subtest 'handling relative paths in update_needle' => sub {
...
+    path($vars_json_path)->make_path;
+    my $vars_json = path($vars_json_path, 'vars.json');
+    my $vars_json_content = '{"NEEDLES_GIT_HASH": "0123456789abcdef0123456789abcdef01234567"}';
+    $vars_json->spew($vars_json_content);
+
+    my $controller = OpenQA::WebAPI::Controller::Step->new;
+    $controller->app($t->app);
+    $settings{CASEDIR} = 'https://something#fragment';
+    $job = $schema->resultset('Jobs')->create_from_settings(\%settings);
+    $job->update({result_dir => $testresult_dir});
```
perlpunk, line 221: We discussed that you should rather set OPENQA_BASEDIR to avoid messing with the source git dir at all.
And `setup_fullstack_temp_dir` does that automatically. And also the `db.lock` thing.
https://github.com/os-autoinst/openQA/blob/master/t/lib/OpenQA/Test/Utils.pm#L369
And it does not return the testresult dir, but the tempdir it created.

Why is it necessary to change the dir of only one certain job?

Wabri: If result_dir is not setted the subroutine when call `_determine_needles_dir_for_job` will return undef because `realpath($job->needle_dir)` is undef. Do I need to mock needle_dir?

t/21-needles.t
```diff
@@ -168,4 +180,65 @@ subtest 'handling relative paths in update_needle' => sub {
...
+    };
+};
+
+subtest 'controller->_create_tmpdir_for_needles_refspec' => sub {
+    my $tempdir = setup_fullstack_temp_dir('21-needles');
+    my $testresult_dir = $tempdir . '21-needles/openqa/testresults/00000';
+    path($testresult_dir)->make_path;
+    my $vars_json = path($testresult_dir, 'vars.json');
+    my $vars_json_content = '{"NEEDLES_GIT_HASH": "0123456789abcdef0123456789abcdef01234567"}';
+    $vars_json->spew($vars_json_content);
```
perlpunk, line 230: Remove those lines and check if the test still passes.

lib/OpenQA/WebAPI/Controller/Step.pm
```diff
@@ -91,6 +118,63 @@ sub view ($self) {
...
+sub _create_tmpdir_for_needles_refspec ($self, $job) {
+    my ($needle_dirs, $needles_dir_var) = $self->_determine_needles_dir_for_job($job);
+    return undef unless $needles_dir_var;
+    my $needles_url = Mojo::URL->new($needles_dir_var->value);
+    return undef unless $needles_url->scheme;
+    my $needles_ref = $needles_url->fragment;
+    eval {
+        my $vars = decode_json(path($job->result_dir, 'vars.json')->slurp);
+        $needles_ref = $vars->{NEEDLES_GIT_HASH} if ref $vars eq 'HASH';
+    };
```
perlpunk, line 165: That eval is never checked. The question is, why is it wrapped in an eval? Which part is ok to fail here?
Should it accept a broken `var.json`? Or a non existing one?
I think an eval around that whole part is not a good design.

The thing is, the test below creates the `vars.json` in the wrong directory, so the file here doesn't exist. The test passes because it gets the sha from git git call later.

t/21-needles.t
Wabri: It is, but I was leaving it here in order to use it to cover the eval section of the function

lib/OpenQA/WebAPI/Controller/Step.pm
Wabri: I found it there before I started working on it so I figured there was a reason for, also while I was doing some tests I discover that if the var.json not exists it rightly throw errors.

Wabri: I've removed the eval section and put a check for undef value for job->result_dir

Wabri: This should also let the coverage be happy even if we have test if the job->result_dir is not empty

t/21-needles.t
Wabri: With the change below I can remove this! what do you think?

lib/OpenQA/WebAPI/Controller/Step.pm
perlpunk: I wrote:
> The thing is, the test below creates the `vars.json` in the wrong directory, so the file here doesn't exist. The test passes because it gets the sha from git git call later.

You are still creating the vars.json in the wrong directory. That's not the directory where this file should end up.
And both things should be tested.
The fact that every line is covered now because of `... if $x` does not mean that the statements are actually covered.

We should make sure that the part with the `NEEDLES_GIT_HASH` is working as well.

t/21-needles.t
perlpunk: I did not suggest removing it because it's not necessary. The problem is that the vars.json is in the wrong place

t/21-needles.t
```diff
@@ -168,4 +180,69 @@ subtest 'handling relative paths in update_needle' => sub {
...
+        my ($needle_dirs, $needles_var) = $controller->_determine_needles_dir_for_job($job);
+        like($needle_dirs, qr|t/data/openqa/share/tests/fedora/needles$|, 'needles dirs is defined');
+        is($needles_var->value, 'https://something#fragment', 'needles var is defined');
+    };
+};
+
+subtest 'controller->_create_tmpdir_for_needles_refspec' => sub {
+    my $tempdir = setup_fullstack_temp_dir('21-needles');
+    my $testresult_dir = $tempdir . '21-needles/openqa/testresults/00000';
+    path($testresult_dir)->make_path;
```
perlpunk, line 209: There is already a method that does that for you:
```suggestion
    my $testresult_dir = $job->create_result_dir;
```
If you compare your variable and this one, you will see the important difference.

t/21-needles.t
```diff
@@ -168,4 +180,69 @@ subtest 'handling relative paths in update_needle' => sub {
...
+    my $needle_refs = $controller->_create_tmpdir_for_needles_refspec($job);
+
+    is($needle_refs, '0123456789abcdef0123456789abcdef01234567', 'needles refs is defined with a fragment');
+
+    # TODO: missing line 164
+};
+
+subtest 'needledir set by jsonfile_in_temp_dir' => sub {
+    # Data
+    my $cwd = Cwd::getcwd;
```
perlpunk, line 230: What is the plan here?
1. We have `$FindBin::Bin` for the directory the test is in. Relying on cwd is usually not needed
2. But we agreed that we don't want to mess with the git directory and instead put stuff into a tempdir

t/21-needles.t
Wabri: Yep, thanks!

t/21-needles.t
```diff
@@ -168,4 +179,75 @@ subtest 'handling relative paths in update_needle' => sub {
...
+subtest 'controller->_create_tmpdir_for_needles_refspec' => sub {
+    my $controller = OpenQA::WebAPI::Controller::Step->new;
+    $controller->app($t->app);
+    setup_fullstack_temp_dir('21-needles');
+    $controller->app->config->{'scm git'} = {checkout_needles_sha => 'yes'};
+
+    subtest 'without fragment and needles_git_hash in vars' => sub {
+        my %settings2 = %settings;
+        $settings2{CASEDIR} = 'https://something';
+        my $job = $schema->resultset('Jobs')->create_from_settings(\%settings2);
```
perlpunk, line 214: perl has this feature where you can modify hash entries only for a certain scope:
```suggestion
        local $settings{CASEDIR} = 'https://something';
        my $job = $schema->resultset('Jobs')->create_from_settings(\%settings);
```
It saves just one line, but maybe you like it

t/21-needles.t
Wabri: Fixed

t/21-needles.t
```diff
@@ -168,4 +179,95 @@ subtest 'handling relative paths in update_needle' => sub {
...
+    $controller->param(distri => $settings{DISTRI});
+    $controller->param(jsonfile => $jsonfile);
+
+    # Call
+    my $something = $controller->needle;
+    # $controller->needle;
+
+    # Test
+    is($something, '', 'test should fail');
+    # TODO: is it the only test i can ?
```
Martchus, line 270: No. This test should probably check whether `$controller->static` is initialized as expected. Maybe it makes sense to mock https://docs.mojolicious.org/Mojolicious/Static#serve_asset to see whether it was called correctly.

Otherwise you could of course also use `Test::Mojo` and do a full-blown HTTP request. This shouldn't be too much trouble actually. Then you can just check whether the expected response was returned for that request using suitable check functions of `Test::Mojo`. This is probably the recommended way of testing a controller function that is directly hooked into routing (and not just a helper).

The variable `$t` which is available in this scope already contains a `Test::Mojo` object so you can simply use that, e.g. `$t->get(…)` to do the request and e.g. `$t->status_is(…)` and `$t->content_like(…)` to check the response.

Wabri: I've update the test with mojo http request, but I'm not found anything related to needle endpoint, are there another way to do it?

lib/OpenQA/WebAPI/Controller/Step.pm
kalikiana: @Wabri Do you think this is covered by the tests you added?

kalikiana: Okay, I realize I didn't really understand it before. This still needs to be reworked.

t/21-needles.t
Martchus: When I remember correctly, I answered this on Slack.

okurz:
> This PR can be closed now that #6190 has been merged. This means the exact details of how this works have changed compared to what was proposed here. So if anything is missing for you now, please let us know.

And please open another PR with any delta that you see missing. After all *only* this PR is closed. Your changes are still preserved in https://github.com/sdclarke/openQA/tree/needles_from_casedir

PR #5177, opened 2023-06-01
Reference further documentation for database migrations

Martchus:
* Reference sections explaining how to import production data and how to tackle performance problems as both are very relevant when creating database migrations
* Related ticket: https://progress.opensuse.org/issues/112946

### DIFF
--- a/docs/Contributing.asciidoc
+++ b/docs/Contributing.asciidoc
@@ -481,6 +481,16 @@ createdb -O ownername -T originaldb newdb
 To initialize or update the database manually before restarting the web UI you can run
 either `./script/initdb --init_database` or `./script/upgradedb --upgrade_database`.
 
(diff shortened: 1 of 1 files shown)


docs/Contributing.asciidoc
```diff
@@ -481,6 +481,16 @@ createdb -O ownername -T originaldb newdb
 To initialize or update the database manually before restarting the web UI you can run
 either `./script/initdb --init_database` or `./script/upgradedb --upgrade_database`.
 
+Migrations that affect possibly big tables should be tested against a production
```
baierjan, line 484: Not against production database but rather against production data locally, right?

baierjan (approved):

kalikiana (approved):

PR #5183, opened 2023-06-05
Forbid cyclic references in YAML

perlpunk:
* Forbid cyclic references in YAML
* Remove YAML::PP duplicate_keys option


kalikiana (approved):

t/46.openqa-yaml.t
okurz, line 1: The common naming scheme we follow says this file should be called
`t/46-openqa-yaml.t` but actually I think we don't need this file at all. Every unit test adds an overhead of around 3-10s which is significant. And the instructions within OpenQA::YAML make the code behave exactly in one way so either there is something wrong in upstream which YAML::PP should test itself or we shouldn't care

okurz (changes requested):

t/46.openqa-yaml.t
perlpunk: ok, I'll move it to t/16-utils-job-templates.t

kraih (approved):

t/46.openqa-yaml.t
okurz: Move looks good:

```
t/16-utils-job-templates.t ................................ ok      957 ms
```

Never mind the other part then.

PR #5185, opened 2023-06-05
Add various optimizations and bug fixes for O3 deployment

kraih:
While deploying Nginx in production on O3 we learned a couple of important lessons that led to these improvements:

* Buffer sizes better reflect the requests we are dealing with in production and reduce the need to write bodies to disk.
* Large file downloads without sendfile made the whole webui feel sluggish.
* Increased read and write timeouts are necessary for WebSockets, as well as certain HTTP endpoints that have not yet been optimized.
* The missing `Host` header resulted in OpenID authentication breaking because of localhost redirects.

Progress: https://progress.opensuse.org/issues/129490

### DIFF
--- a/etc/nginx/vhosts.d/openqa.conf
+++ b/etc/nginx/vhosts.d/openqa.conf
@@ -1,12 +1,33 @@
 server {
     listen       80;
     server_name  openqa.example.com;
+
     root /usr/share/openqa/public;
+
     client_max_body_size 0;
+    client_body_buffer_size 64k;
+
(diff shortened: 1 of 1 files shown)


Martchus (approved):

okurz (approved):

Martchus:
I've just restarted the tests. It was https://progress.opensuse.org/issues/129946 again.

okurz:
I am glad to see that we have way more problems with Selenium based tests rather then actual openQA ones :D

PR #5188, opened 2023-06-05
Add re-try in comments test to mitigate alert problem

Martchus:
This should workaround https://progress.opensuse.org/issues/129946 which even still happens after waiting until
`document.forms.commentForm.text.value` is no longer empty. I have only seen this problems in CI runs and never for real so it is supposedly safe to just workaround it in test code.

I've tested this via `$e->send_keys($text) if $attempts > 1;` to simulate the problem locally.


okurz (approved):

kalikiana (approved):
I guess it can't be helped...

PR #5192, opened 2023-06-06
Fix openqa-clone-custom-git-refspec on git-checkout job sources

okurz:
If openqa-clone-custom-git-refspec is called with openQA jobs that already
specify a git repository in CASEDIR the product dir is computed by the
openQA worker as a relative path without the git checkout directory as a
prefix. That so far has led to the problem that the product directory
string that is passed to openqa-clone-job
would be concatenated without separator leading to cases like
"my_distriproducts/foo" when instead "my_distri" and "products/foo"
should be joint with "/" as standard path separator. This commit fixes
this within openqa-clone-custom-git-refspec by always joining the path
components with "/" as path separator while ensuring that no slashes are
duplicated themselves.

Additionally tested by manually scheduling openQA jobs
openqa.opensuse.org with the generated product dir compontents.

Related progress issue: https://progress.opensuse.org/issues/130096

### DIFF
--- a/script/openqa-clone-custom-git-refspec
+++ b/script/openqa-clone-custom-git-refspec
@@ -123,11 +123,14 @@ Please try 'curl $json_url' or select another job, e.g. in the same scenario: $h
         old_productdir=$(echo "$json_data" | jq -r '.PRODUCTDIR') || throw_json_error "$json_url" "$json_data"
         local old_casedir
         old_casedir=$(echo "$json_data" | jq -r '.CASEDIR') || throw_json_error "$json_url" "$json_data"
+        # set a proper productdir pointing to the checked out repository for
+        # absolute+relative paths
         if [[ ${old_productdir:0:1} == "/" ]]; then
-            local productdir="${productdir:-"${repo_name##*/}${old_productdir##"$old_casedir"}"}"
+            local productdir="${productdir:-"${repo_name##*/}/${old_productdir##"$old_casedir"}"}"
         else
-            local productdir="${productdir:-"${repo_name##*/}${old_productdir#*"${old_casedir##*/}"}"}"
+            local productdir="${productdir:-"${repo_name##*/}/${old_productdir#*"${old_casedir##*/}"}"}"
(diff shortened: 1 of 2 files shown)


script/openqa-clone-custom-git-refspec
```diff
@@ -124,10 +124,11 @@ Please try 'curl $json_url' or select another job, e.g. in the same scenario: $h
         local old_casedir
         old_casedir=$(echo "$json_data" | jq -r '.CASEDIR') || throw_json_error "$json_url" "$json_data"
         if [[ ${old_productdir:0:1} == "/" ]]; then
```
Martchus, line 126: It would be useful to add a comment explaining *why* we need to set the product dir here.

script/openqa-clone-custom-git-refspec
```diff
@@ -124,10 +124,11 @@ Please try 'curl $json_url' or select another job, e.g. in the same scenario: $h
...
-            local productdir="${productdir:-"${repo_name##*/}${old_productdir##"$old_casedir"}"}"
+            local productdir="${productdir:-"${repo_name##*/}/${old_productdir##"$old_casedir"}"}"
         else
-            local productdir="${productdir:-"${repo_name##*/}${old_productdir#*"${old_casedir##*/}"}"}"
+            local productdir="${productdir:-"${repo_name##*/}/${old_productdir#*"${old_casedir##*/}"}"}"
         fi
+        productdir="${productdir/\/\//\/}"
```
Martchus, line 131:
```suggestion
        productdir="${productdir/\/\//\/}"  # avoid consecutive slashes
```

okurz: done

script/openqa-clone-custom-git-refspec
okurz: ok. I added a comment "set a proper productdir pointing to the checked out repository for absolute+relative paths"

kalikiana (approved):

PR #5197, opened 2023-06-12
Switch to Leap 15.5 in CI workflow

baierjan:
Reference: https://progress.opensuse.org/issues/130603

### DIFF
--- a/.obs/workflows.yml
+++ b/.obs/workflows.yml
@@ -16,16 +16,12 @@ pr:
         - target_project: devel:openQA
           target_repository: openSUSE_Tumbleweed
         architectures: [ x86_64 ]
-      - name: openSUSE_Leap_15.4
+      - name: openSUSE_Leap_15.5
         paths:
-        # We need openSUSE:Tools until the fixed version is in Leap 15.4
-        # https://github.com/openSUSE/obs-service-tar_scm/issues/446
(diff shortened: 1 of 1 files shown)


perlpunk (commented):
Why should we remove 15.4 already?

baierjan:
We can leave both until November if that is more desirable. The ticket was suggesting to just replace the versions, so I did.

.obs/workflows.yml
```diff
@@ -16,16 +16,12 @@ pr:
         - target_project: devel:openQA
           target_repository: openSUSE_Tumbleweed
         architectures: [ x86_64 ]
-      - name: openSUSE_Leap_15.4
+      - name: openSUSE_Leap_15.5
```
kalikiana, line 19: > To use the most recently released Leap 15.5 as CI container base in os-autoinst and openQA we should adapt that and also run the according OBS checks on Leap 15.4

I would think we want both? Quoting from the ticket.

baierjan: From the very same ticket 😉 
> If it's easier just replace the version

kalikiana: Two votes for just going with 15.5 on the daily so I guess why not

Martchus (approved):

PR #5200, opened 2023-06-14
Use signatures in Shared::Plugin::SharedHelpers

okurz:


perlpunk (approved):

okurz:
OBS checks missing. OBS is currently experiencing severe problems

PR #5201, opened 2023-06-14
nginx: Set if_modified_since before

perlpunk:
We were experiencing repeated downloads of the same assets in openqa-clone-job, because the asset is set to the current timestamp after download, but by default nginx wants the exact timestamp match.


kraih (approved):

Martchus (approved):

okurz:
OBS checks missing. OBS is currently experiencing severe problems

PR #5204, opened 2023-06-16
Fix source view of local case dirs

perlpunk:
When viewing a source of a test module when CASEDIR was set to a local file instead of an url, the code would be stuck and output these warnings:

    Exiting subroutine via last at /lib/OpenQA/WebAPI/Controller/Step.pm line 351.
    Exiting eval via last at /lib/OpenQA/WebAPI/Controller/Step.pm line 351.

### DIFF
--- a/lib/OpenQA/WebAPI/Controller/Step.pm
+++ b/lib/OpenQA/WebAPI/Controller/Step.pm
@@ -347,19 +347,20 @@ sub src ($self) {
         my $casedir_url = Mojo::URL->new($casedir->value);
         # if CASEDIR points to a remote location let's assume it is a git repo
         # that we can reference like gitlab/github
-        last unless $casedir_url->scheme;
-        my $refspec = $casedir_url->fragment;
-        # try to read vars.json from resultdir and replace branch by actual git hash if possible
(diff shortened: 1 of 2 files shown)


okurz (commented):
Strange how this could go unnoticed. Can you reference an openQA job that fails to show the src or something? Because URLs like https://openqa.opensuse.org/tests/3362061/modules/bootloader_uefi/steps/1/src seem to be fine with CASEDIR=opensuse . Do you mean other cases with "local case dirs"?

perlpunk:
It failed when I used `CASEDIR=/home/tina/openqa-toolbox/os-autoinst-distri-example` locally. maybe it's different for an absolute path.
I can try and start a job on o3 with such a setting.

perlpunk:
Yep, reproduced: https://openqa.opensuse.org/tests/3362271

perlpunk:
The test will also reproduce it, btw. I had to kill the test before I added my fix.

okurz (approved):
Yup, https://openqa.opensuse.org/tests/3362271 is real broken :D

Martchus (approved):

PR #5208, opened 2023-06-19
single-instance: remove broken COPY from docker file

mimi1vx:


okurz:
https://github.com/os-autoinst/openQA/pull/5207

PR #5209, opened 2023-06-19
Validate regex for filtering groups on dashboard pages

Martchus:
* Avoid logging errors and warnings for user-specified regexes
* Return the problem to the client like other validation errors
* Related ticket: https://progress.opensuse.org/issues/131012


lib/OpenQA/WebAPI/Plugin/Helpers.pm
```diff
@@ -366,6 +366,33 @@ sub register ($self, $app, $config) {
...
+            state %problematic_regexes;
+            for my $regex_string (@$regexes) {
+                # check whether the regex is already known to be problematic
+                # note: This is not an optimization. It is required because some warnings are only logged for the first
+                #       time they are encountered.
+                last if $regex_problem = $problematic_regexes{$regex_string};
+                # record warnings as some problems are not fatal
+                # note: We should not leave those warnings unhandled as they would end up in the log.
+                #       An example for such a warning is "$* matches null string many times in regex".
+                local $SIG{__WARN__} = sub ($warning, @) { $regex_problem = $warning };
```
perlpunk, line 382: Instead of using a signal handler I would make warnings fatal in this block:
```
use warnings FATAL => "regexp";
```
Then the eval will die also in those cases.

lib/OpenQA/WebAPI/Plugin/Helpers.pm
```diff
@@ -366,6 +366,33 @@ sub register ($self, $app, $config) {
...
+                # test regex compilation
+                my $parsed_regex = eval { qr/$regex_string/ };
+                if ($@) { $regex_problem = $@ }
+                # test regex matching as some problems are only warned about when matching
+                elsif ($parsed_regex) { '' =~ $parsed_regex }
+                next unless $regex_problem;
+                # strip last part of error/warning as it does not contain anything useful for the user
+                $regex_problem =~ s/; marked by.*//;
```
perlpunk, line 390: I would use a `/s` to also delete potential extra lines.
```suggestion
                $regex_problem =~ s/; marked by.*//s;
```
or
```suggestion
                $regex_problem =~ s{/ at .*}{}s;
```

t/22-dashboard.t
```diff
@@ -40,6 +40,9 @@ is_deeply(\@h2, ['opensuse', 'opensuse test'], 'two groups shown (from fixtures)
 my $test_parent = $parent_groups->create({name => 'Test parent', sort_order => 2});
 
 subtest 'Validation errors' => sub {
+    $t->get_ok('/dashboard_build_results/?group=[')->status_is(400)
```
perlpunk, line 43: Should we also have a test for a valid input? I can't find one.

lib/OpenQA/WebAPI/Plugin/Helpers.pm
Martchus: I didn't know that was possible. I suppose I'll do that then.

lib/OpenQA/WebAPI/Plugin/Helpers.pm
```diff
@@ -366,6 +366,33 @@ sub register ($self, $app, $config) {
...
         });
+
+    $app->helper(
+        regex_problem => sub ($c, $regexes, $context = undef) {
+            my $regex_problem;
+            state %problematic_regexes;
+            for my $regex_string (@$regexes) {
+                # check whether the regex is already known to be problematic
+                # note: This is not an optimization. It is required because some warnings are only logged for the first
+                #       time they are encountered.
```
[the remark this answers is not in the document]
Martchus: Basically the same example as the one that only lead to a warning. It is only logged once per line of code where it is used.

t/22-dashboard.t
Martchus: Right, I suppose that would make sense.

lib/OpenQA/WebAPI/Plugin/Helpers.pm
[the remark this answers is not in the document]
perlpunk: Ah I see, it happens sometimes for `$*`, but it doesn't even seem to be deterministic.
However, using fatal warnings will also eliminate this problem, in this case it is always logged.

lib/OpenQA/WebAPI/Plugin/Helpers.pm
```diff
@@ -366,6 +366,27 @@ sub register ($self, $app, $config) {
...
+                #       An example for such a warning is "$* matches null string many times in regex".
+                use warnings FATAL => 'regexp';
+                # test regex compilation
+                my $parsed_regex = eval { qr/$regex_string/ };
+                if ($@) { $regex_problem = $@ }
+                # test regex matching as some problems are only warned about when matching
+                elsif ($parsed_regex) { '' =~ $parsed_regex }
+                next unless $regex_problem;
+                # strip last part of error/warning as it does not contain anything useful for the user
+                $regex_problem =~ s/ at.*//;
```
perlpunk, line 385:
```suggestion
                $regex_problem =~ s/ at.*//s;
```

lib/OpenQA/WebAPI/Plugin/Helpers.pm
```diff
@@ -366,6 +366,27 @@ sub register ($self, $app, $config) {
...
+            for my $regex_string (@$regexes) {
+                # treat regex warnings as fatal as apparently not all problems are fatal errors
+                # note: We should not leave those warnings unhandled as they would end up in the log.
+                #       An example for such a warning is "$* matches null string many times in regex".
+                use warnings FATAL => 'regexp';
+                # test regex compilation
+                my $parsed_regex = eval { qr/$regex_string/ };
+                if ($@) { $regex_problem = $@ }
+                # test regex matching as some problems are only warned about when matching
+                elsif ($parsed_regex) { '' =~ $parsed_regex }
```
perlpunk, line 382: Can you give an example where the warning only appears when matching?
And what would that code here then do? with the fatal warnings, it would die, but it's not in the eval block anymore...

perlpunk: It's probably not possible to catch all possible warnings with having a match on the empty string.
So I just looked through [perldoc perldiag](https://perldoc.perl.org/perldiag), lloking for `(W regexp)`, and I think warnings like `Complex regular subexpression recursion limit (%d) exceeded` would only appear in the real matching.

Doing the eval at the actual match would make the code a bit more complicated, I guess.

Martchus: > Doing the eval at the actual match would make the code a bit more complicated, I guess.

Not sure how to solve this. Maybe this validation is really just good enough as a first sanity check. So the places where we actually use the regex would use some "safe" helper function to do the matching with proper error handling.

t/22-dashboard.t
```diff
@@ -40,6 +40,10 @@ is_deeply(\@h2, ['opensuse', 'opensuse test'], 'two groups shown (from fixtures)
 my $test_parent = $parent_groups->create({name => 'Test parent', sort_order => 2});
 
 subtest 'Validation errors' => sub {
+    $t->get_ok('/dashboard_build_results/?group=valid-regex')->status_is(200);
```
perlpunk, line 43: Maybe it would be good to actually test that we get results back.
```suggestion
    $t->get_ok('/dashboard_build_results/?group=opensuse test')->status_is(200)->content_like(qr{/group_overview/1002});
```

okurz (approved):
If I read it correctly this approach is trying to forward all Perl regex warnings to users without trying to explicitly validate any regex which I think is the right approach

lib/OpenQA/WebAPI/Plugin/Helpers.pm
Martchus: The version I've just pushed fixes all problems. So `'' =~ $parsed_regex` has been moved inside the `eval`-block and regex errors/warnings are also handled when doing actual matching later on.

okurz (approved):

PR #5210, opened 2023-06-20
munin: Make alert thresholds configurable

perlpunk:
Issue: https://progress.opensuse.org/issues/99741

### DIFF
--- a/contrib/munin/minion
+++ b/contrib/munin/minion
@@ -8,7 +8,12 @@ use v5.22;
 # Add this to /etc/munin/plugin-conf.d/munin-node:
 #
 #   [minion_*]
-#   env.url https://openqa.opensuse.org/admin/influxdb/minion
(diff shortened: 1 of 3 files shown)


contrib/munin/minion
```diff
@@ -52,6 +57,8 @@ sub main {
 sub data {
     my ($config, $category) = @_;
     my $url = $ENV{url} || "https://openqa.opensuse.org/admin/influxdb/minion";
+    my $span = 5; # not configurable (yet)
```
Martchus, line 60: Why did you not make this configurable yet? I guess it isn't a necessity but the comment makes one wonder why.

perlpunk: Because I haven't found out if it's possible to adjust the timespan on munin side.
And it would only make sense if we can do that.
The only other way I could think is that we divide the value, e.g. if we configure a timespan of 20 minutes, we would divide the value y 4 to display correct values.

Martchus (approved):

okurz (approved):

PR #5213, opened 2023-06-21
Avoid hard dependency on Apache in systemd file

Martchus:
* Allow using a different HTTP server without having to mask the unit of Apache
* Update documentation to explicitly enable Apache; the bootstrap script already does this anyways
* Keep the spec file as-is because it only lists `apache2` as as recommended dependency
* This is a breaking change for users of Apache that don't have its unit explicitly enabled; I guess it is acceptable as this is easy to notice and fix and not the case when the bootstrap script or our salt states were used to setup
* Related ticket: https://progress.opensuse.org/issues/130477

### DIFF
--- a/docs/Installing.asciidoc
+++ b/docs/Installing.asciidoc
@@ -506,11 +506,13 @@ To start openQA and enable it to run on each boot call
 systemctl enable --now postgresql
 systemctl enable --now openqa-webui
 systemctl enable --now openqa-scheduler
-# openSUSE
+# to use Apache as reverse proxy under openSUSE
+systemctl enable apache2
 systemctl restart apache2
-# Fedora
+# to use Apache as reverse proxy under Fedora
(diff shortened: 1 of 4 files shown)


baierjan (approved):

perlpunk (approved):

Martchus:
> But maybe we can introduce not a breaking change.

I'm open for ideas but I wouldn't over-engineer this. The number of affected users is probably very small and the problem of Apache being not started anymore is very easy to detect and fix.

We could of course create a v2 version of all systemd unit files (except for the worker units) to avoid the breaking change. I personally find that this is more confusing in the long run, though.

Martchus:
I guess they'll get over it and will just C&P the two lines together or use something different anyways. For a more streamlined setup we have the bootstrap script and the containerized setup anyways. So at this point the user is supposedly already aiming for a more customized/manual setup.

PR #5214, opened 2023-06-21
[DISCUSSION] Revert "Avoid hard dependency on Apache in systemd file"

okurz:
Reverts os-autoinst/openQA#5213 in case we need it. See https://github.com/os-autoinst/openQA/pull/5213#pullrequestreview-1491097920 for reasoning


Martchus:
I personally stay by my opinion, see https://github.com/os-autoinst/openQA/pull/5213#issuecomment-1601083032. So I'm not approving this PR (but also won't request changes).

okurz:
Adding the "not-ready" label just to have the PR needed in case we need it or someone finds it and says we broke something and we should merge this

PR #5215, opened 2023-06-21
Add subpackage openQA-munin

perlpunk:
Issue: https://progress.opensuse.org/issues/99741

Tested here:
* OBS https://build.opensuse.org/package/show/home:tinita:branches:devel:openQA/openQA
* Download: https://download.opensuse.org/repositories/home:/tinita:/branches:/devel:/openQA/openSUSE_Tumbleweed/x86_64/

### DIFF
--- a/dist/rpm/openQA.spec
+++ b/dist/rpm/openQA.spec
@@ -288,6 +288,19 @@ upgrading the system if devel:openQA packages are stable and contain updates. It
 is complementary to auto-update which also reboots the system and does updates
 regardless of whether devel:openQA contains updates.
 
+%package munin
+Summary:        Munin scripts
+Group:          Development/Tools/Other
+Requires:       munin
+Requires:       munin-node
+Requires:       curl
+Requires:       perl
+
+%description munin
+Use this package to install munin scripts that allow to monitor some openQA
+statistics.
+
+
 %prep
 %setup -q -a1
 sed -e 's,/bin/env python,/bin/python,' -i script/openqa-label-all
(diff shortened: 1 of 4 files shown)


dist/rpm/openQA.spec
```diff
@@ -288,6 +288,19 @@ upgrading the system if devel:openQA packages are stable and contain updates. It
 is complementary to auto-update which also reboots the system and does updates
 regardless of whether devel:openQA contains updates.
 
+%package munin
+Summary:        Munin scripts
```
Martchus, line 292:
```suggestion
Summary:        Munin scripts for monitoring openQA
```

dist/rpm/openQA.spec
```diff
@@ -742,4 +761,14 @@ fi
 %{_unitdir}/openqa-continuous-update.*
 %{_datadir}/openqa/script/openqa-continuous-update
 
+%files munin
+%defattr(-,root,root)
+%doc contrib/munin/config/minion.config
+%dir %{_prefix}/lib/munin
+%dir %{_prefix}/lib/munin/plugins
```
Martchus, line 768: This is not a usual library. Should it really go under `lib`? Have you checked whether there are already other packages containing Munin plugins and what location those packages use?

perlpunk: At least that's the location where the plugins are located that come with munin itself...
There is one plugin I could find that uses `_datadir`: https://build.opensuse.org/package/view_file/security:Zorp:7:0:7/zorp/zorp-7_0-7.spec?expand=1

okurz (approved):

dist/rpm/openQA.spec
Martchus: Looks like the munin package in Factory uses this path, indeed. I suppose then it is fine.

Martchus (approved):

dist/rpm/openQA.spec
perlpunk: hm, I still wanted to do that before it's merged... :)

PR #5218, opened 2023-06-22
Workaround sporadic failures of developer mode test

Martchus:
* Retry re-login in developer mode test
* Related ticket: https://progress.opensuse.org/issues/128807

### DIFF
--- a/t/33-developer_mode.t
+++ b/t/33-developer_mode.t
@@ -351,7 +351,7 @@ subtest 'quit session' => sub {
 subtest 'test cancelled by quitting the session' => sub {
     $driver->switch_to_window($first_tab);
     $driver->get($job_page_url);
-    ok wait_for_result_panel($driver, qr/(State: cancelled|Result: (user_cancelled|passed))/),
+    ok wait_for_result_panel($driver, qr/Result: (user_cancelled|passed)/),
       'test 1 has been cancelled (if it was fast enough to actually pass that is ok, too)';
     my $log_file_path = path($resultdir, '00000', "00000001-$job_name")->make_path->child('autoinst-log.txt');
     ok -s $log_file_path, "log file generated under $log_file_path";
--- a/t/33-developer_mode.t
+++ b/t/33-developer_mode.t
@@ -80,11 +80,21 @@ $ws = create_websocket_server(undef, 0, 0);
 $scheduler = create_scheduler;
 $livehandler = create_live_view_handler;
 
+# logs out and logs in again as the specified user; tries multiple times to workaround poo#128807
+my $max_login_attempts = $ENV{OPENQA_DEVEL_MODE_TEST_MAX_LOGIN_ATTEMPTS} // 10;
 sub relogin_as ($user) {
-    $driver->get('/logout');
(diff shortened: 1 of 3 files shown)


okurz:
s/sporatic/sporadic/.

Now *I* am interested to know what fail ratios and confidence you found:) I suggest you use https://github.com/okurz/scripts/blob/master/count_fail_ratio for that

t/33-developer_mode.t
```diff
@@ -80,6 +80,23 @@ $ws = create_websocket_server(undef, 0, 0);
 $scheduler = create_scheduler;
 $livehandler = create_live_view_handler;
 
+# logs out and logs in again as the specified user; tries multiple times to workaround poo#128807
+my $max_login_attempts = $ENV{OPENQA_DEVEL_MODE_TEST_MAX_LOGIN_ATTEMPTS} // 10;
+sub relogin_as ($user) {
+    my $login_text = '';
+    my $expected_login_text = 'Logged in as ' . $user;
+    for (my $attempts = 1;; ++$attempts) {
```
Martchus, line 88: I actually need to check for `$max_login_attempts` here.

t/33-developer_mode.t
```diff
@@ -80,6 +80,23 @@ $ws = create_websocket_server(undef, 0, 0);
...
+my $max_login_attempts = $ENV{OPENQA_DEVEL_MODE_TEST_MAX_LOGIN_ATTEMPTS} // 10;
+sub relogin_as ($user) {
+    my $login_text = '';
+    my $expected_login_text = 'Logged in as ' . $user;
+    for (my $attempts = 0; $attempts < $max_login_attempts; ++$attempts) {
+        if ($login_text ne 'Login') {
+            $driver->get('/logout');
+            $driver->element_text_is('#user-action a', 'Login', 'logged-out before logging in as ' . $user);
+        }
+        $driver->get('/login?user=' . $user);
```
okurz, line 93: that's unexpected. Why is not covered now?

Martchus: I also don't understand this. If it was not covered we would never execute `return pass` which would result in a test failure. I suspect the coverage tracking is broken.

Martchus:
With the additional "Fix race condition …" commit which is about a different problem in the same test it seems pretty stable on my local machine:

```
## count_fail_ratio: Run: 35. Fails: 0. Fail ratio 0%
```

I'll keep this running for a little bit longer.

Martchus:
```
## count_fail_ratio: Run: 97. Fails: 0. Fail ratio 0±0%. No fails, computed failure probability < 3.09%
```

Together with the previously mentioned runs this should be good enough.

Martchus:
I cannot reproduce the uncovered line when generating a coverage report locally and it also makes no sense at all. I'll try marking the line as uncoverable. Maybe that helps with the report on codecov.

Martchus:
Looks like `# uncoverable statement (must be bug in coverage tracking)` helped. (I first forced-pushed without it and the missing coverage persisted. So just re-triggering alone didn't help.)

Martchus:
> Please remove it from the list of unstable tests as well now.

done

okurz (approved):

perlpunk (approved):

PR #5220, opened 2023-06-23
Do not hardlink symlink assets

perlpunk:
Issue: https://progress.opensuse.org/issues/129340

Hardlinking a symlink will only work if the link is absolute, and we would have to check that recursively if the symlink target itself is a symlink, so just fall back to symlink then.

### DIFF
--- a/lib/OpenQA/Worker/Engines/isotovideo.pm
+++ b/lib/OpenQA/Worker/Engines/isotovideo.pm
@@ -176,10 +176,16 @@ sub _link_asset ($asset, $pooldir) {
     # Try to use hardlinks first and only fall back to symlinks when that fails,
     # to ensure that assets cannot be purged early from the pool even if the
     # cache service runs out of space
-    eval { link($asset, $target) or die qq{Cannot create link from "$asset" to "$target": $!} };
-    if (my $err = $@) {
+    # If the given asset is a symlink itself, do not hardlink a symlink
+    my $linked = 0;
+    unless (-l $asset) {
+        $linked = eval { link($asset, $target) or die qq{Cannot create link from "$asset" to "$target": $!} };
+        if (my $err = $@) {
+            log_debug(qq{Symlinking asset because hardlink failed: $err});    # uncoverable statement
+        }
+    }
+    unless ($linked) {
         symlink($asset, $target) or die qq{Cannot create symlink from "$asset" to "$target": $!};
-        log_debug(qq{Symlinked asset because hardlink failed: $err});
     }
     log_debug(qq{Linked asset "$asset" to "$target"});
 
--- a/t/24-worker-engine.t
+++ b/t/24-worker-engine.t
@@ -428,21 +428,33 @@ subtest 'behavior with ABSOLUTE_TEST_CONFIG_PATHS=1' => sub {
     };
 };
 
-subtest 'symlink asset' => sub {
+subtest 'link asset' => sub {
     my $pool_directory = tempdir('poolXXXX');
     my $worker = OpenQA::Test::FakeWorker->new(pool_directory => $pool_directory);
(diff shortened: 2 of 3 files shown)


perlpunk:
I'm getting a weird failure on circleci:
https://app.circleci.com/pipelines/github/os-autoinst/openQA/11816/workflows/310c0f76-e852-40f1-b121-ff5829f9d4ff/jobs/110462
```
[11:04:31] t/24-worker-engine.t ...................................... 1/? Can't open file "poolAgHr/vars.json": No such file or directory at t/24-worker-engine.t line 60.
# Tests were run but no plan was declared and done_testing() was not seen.
# Looks like your test exited with 255 just after 11.
[11:04:31] t/24-worker-engine.t ......................................       Dubious, test returned 255 (wstat 65280, 0xff00)
All 11 subtests passed
```
However, the test is passing locally. Not sure what's going on there...

t/24-worker-engine.t
```diff
@@ -428,21 +428,33 @@ subtest 'behavior with ABSOLUTE_TEST_CONFIG_PATHS=1' => sub {
...
+    my $orig_hdd = locate_asset('hdd', 'foo.qcow2');
+    my $orig_iso = locate_asset('iso', 'openSUSE-13.1-DVD-x86_64-Build0091-Media.iso');
+    my $linked_iso = "$pool_directory/openSUSE-13.1-DVD-x86_64-Build0091-Media.iso";
+    my $linked_hdd = "$pool_directory/foo.qcow2";
+    my $linked_hdd2 = "$pool_directory/symlink.qcow2";
+    ok -e $linked_iso, 'the iso is linked to pool directory';
+    ok -e $linked_hdd, 'the hdd is linked to pool directory';
+    ok -l $linked_hdd2, 'the hdd 2 is symlinked to pool directory';
+    is + (stat $linked_hdd)[1], +(stat $orig_hdd)[1], 'hdd is hardlinked';
+    is + (stat $linked_iso)[1], +(stat $orig_iso)[1], 'iso is hardlinked';
```
Martchus, line 454: Why is this "+"-syntax needed? If it is needed it would make sense to be consistent about the spacing around the "+".

Martchus (commented):
Looks generally good.

perlpunk:
> I'm getting a weird failure on circleci: 

Ah, nevermind, I forgot to commit a new file... should be fixed

t/24-worker-engine.t
perlpunk: I wanted to avoid having to use parens. With a simple plus sign one can remove such an ambiguity in perl. The inconsistent spacing is done by perltidy :-/
I guess I go back to parens then, might look too esoteric for most people :)

lib/OpenQA/Worker/Engines/isotovideo.pm
```diff
@@ -176,10 +176,16 @@ sub _link_asset ($asset, $pooldir) {
     # Try to use hardlinks first and only fall back to symlinks when that fails,
     # to ensure that assets cannot be purged early from the pool even if the
     # cache service runs out of space
-    eval { link($asset, $target) or die qq{Cannot create link from "$asset" to "$target": $!} };
-    if (my $err = $@) {
+    # If the given asset is a symlink itself, to not hardlink a symlink
```
okurz, line 179: "to not hardlink a symlink"? I guess s/to/do?

okurz (changes requested):

lib/OpenQA/Worker/Engines/isotovideo.pm
perlpunk: argh, yes. fixed

mimi1vx (approved):

Martchus (approved):

okurz (approved):

PR #5221, opened 2023-06-26
contrib: Add openSUSE Leap autoyast variant with some quirk handling

okurz:
Related progress issue: https://progress.opensuse.org/issues/124221#note-21


okurz:
I made that profile available over http://s.qa.suse.de/oqa-ay-lp which you should be able to use in PXE boot.

mimi1vx (approved):

kalikiana (approved):

okurz:
https://github.com/os-autoinst/openQA/pull/5250 should be a full replacement

PR #5225, opened 2023-06-27
Use a unique template for worker tempdir

perlpunk:
It's easier to identify in case of problems.

Issue: https://progress.opensuse.org/issues/131465

### DIFF
--- a/lib/OpenQA/Scheduler/Model/Jobs.pm
+++ b/lib/OpenQA/Scheduler/Model/Jobs.pm
@@ -458,7 +458,9 @@ sub _assign_multiple_jobs_to_worker ($self, $jobs, $worker, $directly_chained_jo
     if (my $tmpdir = $worker->get_property('WORKER_TMPDIR')) {
         File::Path::rmtree($tmpdir);
     }
-    my %worker_properties = (JOBTOKEN => random_string(), WORKER_TMPDIR => tempdir());
+    my %worker_properties = (
+        JOBTOKEN => random_string(),
+        WORKER_TMPDIR => tempdir(sprintf('webui.worker-%d.XXXXXXXX', $worker_id), TMPDIR => 1));
     $worker->set_property(WORKER_TMPDIR => $worker_properties{WORKER_TMPDIR});
     $job_data{$_->id} = $_->prepare_for_work($worker, \%worker_properties) for @$jobs;
     return OpenQA::WebSockets::Client->singleton->send_jobs(\%job_info);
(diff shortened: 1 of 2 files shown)


okurz (approved):

lib/OpenQA/Scheduler/Model/Jobs.pm
```diff
@@ -458,7 +458,7 @@ sub _assign_multiple_jobs_to_worker ($self, $jobs, $worker, $directly_chained_jo
     if (my $tmpdir = $worker->get_property('WORKER_TMPDIR')) {
         File::Path::rmtree($tmpdir);
     }
-    my %worker_properties = (JOBTOKEN => random_string(), WORKER_TMPDIR => tempdir());
+    my %worker_properties = (JOBTOKEN => random_string(), WORKER_TMPDIR =>  tempdir(sprintf('worker-%d.XXXXXXXX', $worker_id), TMPDIR => 1));
```
kraih, line 461: Maybe it should also mention the origin service, like `webui-worker-%d.XXXXXXXX` and `scheduler-worker-%d.XXXXXXXX`. After all tracking it down took longer because the error looked like it originated on a worker and not the webui host.

perlpunk: done

lib/OpenQA/Scheduler/Model/Jobs.pm
```diff
@@ -458,7 +458,7 @@ sub _assign_multiple_jobs_to_worker ($self, $jobs, $worker, $directly_chained_jo
     if (my $tmpdir = $worker->get_property('WORKER_TMPDIR')) {
         File::Path::rmtree($tmpdir);
     }
-    my %worker_properties = (JOBTOKEN => random_string(), WORKER_TMPDIR => tempdir());
+    my %worker_properties = (JOBTOKEN => random_string(), WORKER_TMPDIR =>  tempdir(sprintf('webui.worker-%d.XXXXXXXX', $worker_id), TMPDIR => 1));
```
kraih, line 461: This model is actually used by the scheduler. 😁

Martchus (commented):
`s/uniqe/unique/`

lib/OpenQA/Scheduler/Model/Jobs.pm
perlpunk: ok, I kept it

Martchus (approved):

lib/OpenQA/Scheduler/Model/Jobs.pm
kraih: I would have preferred unique templates here, because then there would be no guessing which one was responsible if we ever see a related error in the logs again. Now it could be either. 😉

perlpunk: Make a PR :)

PR #5234, opened 2023-07-05
Use indented here-documents

perlpunk:
They are supported since perl 5.26: https://perldoc.perl.org/5.26.0/perldelta#Indented-Here-documents

vim also manages to highlight it correctly :)


Martchus:
Looks like there's at least one mistake, though:
```
Can't find string terminator "END_SQL" anywhere before EOF at /home/squamata/project/lib/OpenQA/Schema/ResultSet/Assets.pm line 151.
```

PR #5235, opened 2023-07-06
Convert script/configure-web-proxy from POSIX sh to bash

asdil12:


script/configure-web-proxy
```diff
@@ -27,11 +27,11 @@ done
 
 sed -i -e 's/^.*httpsonly.*$/httpsonly = 0/g' /etc/openqa/openqa.ini
 
-if [ "$web_proxy" = "nginx" ] ; then
+if [[ $web_proxy == "nginx" ]] ; then
 	echo "Setting up nginx"
 	sed "s/openqa.example.com/$(hostname)/" /etc/nginx/vhosts.d/openqa.conf.template > /etc/nginx/vhosts.d/openqa.conf
 	sed -i -e "s/\(^[^#]*server_name  localhost;\)/#\1/" /etc/nginx/nginx.conf
-elif [ "$web_proxy" = "apache" ] || [ "$web_proxy" = "apache2" ] ; then
+elif [[ $web_proxy == "apache" ]] || [[ $web_proxy == "apache2" ]] ; then
```
perlpunk, line 34:
```suggestion
elif [[ $web_proxy == "apache" || $web_proxy == "apache2" ]] ; then
```

okurz (commented):
Why not keep it POSIX sh? Or was it inconsistent in before?

Martchus:
Because almost all systems have bash installed (even if it is not the shell used by the user) so it makes no sense to limit ourselves to inferior shell syntax.

okurz:
> Because almost all systems have bash installed (even if it is not the shell used by the user) so it makes no sense to limit ourselves to inferior shell syntax.

Ok, that's a good reason why the change can be done without significant impact but is not a reason why the change would provide a benefit

asdil12:
> > Because almost all systems have bash installed (even if it is not the shell used by the user) so it makes no sense to limit ourselves to inferior shell syntax.
> 
> Ok, that's a good reason why the change can be done without significant impact but is not a reason why the change would provide a benefit

I guess we should decide: Make all scripts rely on bash features like `==` or `[[` or be posix sh compliant.
I see no value in the current chaotic situation where we try to enforce `[[` in one file but hesitate to do the same in other files.

Martchus (approved):

perlpunk (approved):

okurz:
Missing codecov feedback

@Mergifyio rebase

okurz:
I think mergify never rebased but it seems the codecov report nevertheless eventually arrived :D

PR #5237, opened 2023-07-07
Fix escaping in test tables

Martchus:
See https://progress.opensuse.org/issues/132413


Martchus:
I've just tested this locally (where I randomly have jobs with such problematic names for some reason). I can extend UI tests if that's wanted.

okurz (approved):

perlpunk (approved):

PR #5238, opened 2023-07-10
Fix openqa-bootstrap script referencing unbound var

asdil12:
https://bugzilla.suse.com/show_bug.cgi?id=1213160
https://progress.opensuse.org/issues/131024

### DIFF
--- a/script/openqa-bootstrap
+++ b/script/openqa-bootstrap
@@ -52,7 +52,7 @@ su postgres -c "/usr/share/openqa/script/setup-db" "$dbuser" "$dbname"
 proxy_args=""
 [[ -n "$setup_web_proxy" ]] && proxy_args="--proxy=$setup_web_proxy"
 setup=/usr/share/openqa/script/configure-web-proxy
-(command -v $setup && sh -ex $setup "$proxy_args") || (curl -s https://raw.githubusercontent.com/os-autoinst/openQA/master/script/configure-web-proxy | bash -ex -s "$proxy_args")
+(command -v $setup && sh -ex $setup "$proxy_args") || (curl -s https://raw.githubusercontent.com/os-autoinst/openQA/master/script/configure-web-proxy | bash -ex -s -- "$proxy_args")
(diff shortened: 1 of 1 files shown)


script/openqa-bootstrap
```diff
@@ -49,7 +49,7 @@ fi
 su postgres -c "/usr/share/openqa/script/setup-db" "$dbuser" "$dbname"
 
 # setup webserver and fake-auth
-[[ -n "$setup_web_proxy" ]] && proxy_args="--proxy=$setup_web_proxy"
+[[ -n "$setup_web_proxy" ]] && proxy_args="--proxy=$setup_web_proxy" || proxy_args=""
```
okurz, line 52: I would say the usual approach is

```suggestion
proxy_args=""
[[ -n "$setup_web_proxy" ]] && proxy_args="--proxy=$setup_web_proxy"
```

perlpunk (approved):

script/openqa-bootstrap
```diff
@@ -32,6 +32,9 @@ pkgs=(openQA-single-instance qemu-arm qemu-ppc qemu-x86 qemu-tools sudo iputils
 if [ "$(uname -m)" = "aarch64" ]; then
     pkgs+=(qemu-uefi-aarch64)
 fi
+if [[ $setup_web_proxy == "nginx" ]] ; then
+    pkgs+=(nginx)
+fi
```
okurz, line 37: I suggest to solve that as part of the package already. Everything we can do in the package we don't need to do in bootstrap, i.e.:
```suggestion
- pkgs=(openQA-single-instance qemu-arm qemu-ppc qemu-x86 qemu-tools sudo iputils os-autoinst-distri-opensuse-deps)
if [[ $setup_web_proxy == "nginx" ]] ; then
    pkgs=(openQA-single-instance-nginx)
else
    pkgs=(openQA-single-instance)
fi
+ pkgs+=(qemu-arm qemu-ppc qemu-x86 qemu-tools sudo iputils os-autoinst-distri-opensuse-deps)
```

okurz (changes requested):

PR #5243, opened 2023-07-11
Add nginx config for tls

nicksinger:
This adds a minimal example to our nginx config to serve openQA over TLS. I used the opportunity to also bind on **all** v6 addresses instead of just loopback on port 80 because I wanted to keep the config consistent with the tls section.


okurz (approved):

kalikiana (approved):
Niceeee

asdil12:
This breaks the tests and openqa bootstrap setup:
https://openqa.opensuse.org/tests/3422958#step/openqa_webui/9

PR #5245, opened 2023-07-12
contrib: Update/extend autoyast profiles

okurz:
* contrib: Add openSUSE Leap specific profile
* contrib: Fix non-interactive 2nd stage autoyast script install
* contrib: Fix the network startup on machines with multiple interfaces
* contrib: Ensure a complete non-interactive autoyast run
* contrib: Use consistent 2-space indention in the autoyast profile

### DIFF
--- a/contrib/ay-openqa-worker-leap.xml
+++ b/contrib/ay-openqa-worker-leap.xml
@@ -1,0 +1,73 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!DOCTYPE profile>
+<profile xmlns="http://www.suse.com/1.0/yast2ns" xmlns:config="http://www.suse.com/1.0/configns">
+  <general>
+    <mode>
+      <confirm config:type="boolean">false</confirm>
+    </mode>
+  </general>
+  <networking>
+    <!-- For multiple network interfaces by default the last is configured, prefer the first -->
+    <interfaces config:type="list">
+      <interface>
+        <bootproto>dhcp</bootproto>
+        <name>eth0</name>
+        <startmode>auto</startmode>
+      </interface>
+    </interfaces>
+    <keep_install_network config:type="boolean">true</keep_install_network>
+  </networking>
+  <software>
+    <!-- It seems unfortunately we need to explicitly specify a product, would have preferred a more generic config -->
+    <products config:type="list">
+      <product>Leap</product>
+    </products>
+    <packages config:type="list">
+      <package>openssh</package>
+      <package>sudo</package>
+      <!-- could not find salt-minion in default Leap repos, disabling, handling in 2nd stage script -->
+      <!--<package>salt-minion</package>-->
+    </packages>
+  </software>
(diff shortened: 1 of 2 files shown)


mimi1vx (approved):

contrib/ay-openqa-worker-leap.xml
```diff
@@ -0,0 +1,72 @@
...
+        <name>eth0</name>
+        <startmode>auto</startmode>
+      </interface>
+    </interfaces>
+    <keep_install_network config:type="boolean">true</keep_install_network>
+  </networking>
+  <software>
+    <!-- It seems unfortunately we need to explicitly specify a product, would have preferred a more generic config -->
+    <products config:type="list">
+      <product>Leap</product>
```
kalikiana, line 23: Can we [use a condition here?](https://doc.opensuse.org/projects/autoyast/#erb-os-release-helper)
```suggestion
      <% if os_release[:id] == 'sle' %>
      <product>SLES</product>
      <% else %>
      <product>Leap</product>
      <% end %>
```

contrib/ay-openqa-worker-leap.xml
```diff
@@ -0,0 +1,72 @@
...
+    <post-scripts config:type="list">
+      <script>
+        <filename>setup.sh</filename>
+        <interpreter>shell</interpreter>
+        <debug config:type="boolean">true</debug>
+        <source><![CDATA[
+            timeout 600 sh -c 'until ! pgrep -a zypper; do sleep 5; done'
+            zypper -n in salt-minion
+            echo 'roles: worker' > /etc/salt/grains
+            systemctl enable --now sshd salt-minion
```
Martchus, line 58: Maybe add a comment that this doesn't work at this point reliably and one might invoke those commands manually instead. It is also useless to start salt-minion without configuring the salt master.

contrib/ay-openqa-worker.xml
```diff
@@ -7,6 +7,14 @@
     </mode>
   </general>
   <networking>
+    <!-- For multiple network interfaces by default the last is configured, prefer the first -->
```
Martchus, line 10: This is just a guess of course. I've just used https://w3.nue.suse.com/~okurz/ay-openqa-worker-leap.xml which has this config as well and it didn't work on https://qesapworker-prg4-mgmt.qa.suse.cz (but I was able to select the right network interface manually and then continue with the automatic installation).

okurz: I don't think it's a "guess" to autoyast but more like a "hint" because I observed on a machine with three ethernet interfaces that it only configured the last, unconnected one until I added this setting when the first but also the other interfaces were configured :shrug:

Martchus: I can just say that it didn't work for me. However, if it increases the likelihood for things to work in general I'm fine with it.

okurz:
obsoleted by https://github.com/os-autoinst/openQA/pull/5250

PR #5248, opened 2023-07-13
Comment out TLS section in nginx config by default as there are no certs

asdil12:
Fixes https://github.com/os-autoinst/openQA/pull/5243

### DIFF
--- a/etc/nginx/vhosts.d/openqa.conf.template
+++ b/etc/nginx/vhosts.d/openqa.conf.template
@@ -8,12 +8,12 @@ server {
     include vhosts.d/openqa-locations.inc;
 }
 
-server {
-    listen       443 ssl;
-    listen       [::]:443 ssl;
-    server_name  openqa.example.com;
-
-    ssl_certificate /etc/dehydrated/certs/openqa.example.com/fullchain.pem;
-    ssl_certificate_key /etc/dehydrated/certs/openqa.example.com/privkey.pem;
-    include vhosts.d/openqa-locations.inc;
-}
+#server {
+#    listen       443 ssl;
+#    listen       [::]:443 ssl;
+#    server_name  openqa.example.com;
+#
+#    ssl_certificate /etc/dehydrated/certs/openqa.example.com/fullchain.pem;
+#    ssl_certificate_key /etc/dehydrated/certs/openqa.example.com/privkey.pem;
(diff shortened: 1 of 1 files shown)


etc/nginx/vhosts.d/openqa.conf.template
```diff
@@ -8,12 +8,12 @@ server {
...
-    ssl_certificate_key /etc/dehydrated/certs/openqa.example.com/privkey.pem;
-    include vhosts.d/openqa-locations.inc;
-}
+#server {
+#    listen       443 ssl;
+#    listen       [::]:443 ssl;
+#    server_name  openqa.example.com;
+#
+#    ssl_certificate /etc/dehydrated/certs/openqa.example.com/fullchain.pem;
+#    ssl_certificate_key /etc/dehydrated/certs/openqa.example.com/privkey.pem;
```
Martchus, line 17: We could just use `/usr/lib/perl5/vendor_perl/5.36.1/Mojo/IOLoop/resources/server.crt` and `/usr/lib/perl5/vendor_perl/5.36.1/Mojo/IOLoop/resources/server.key` by default. The only problem is that these paths contain the Perl version so it is not that easy.

For the containerized setup I therefore wrote code like this:
```
  local mojo_resources=$(perl -e 'use Mojolicious; print(Mojolicious->new->home->child("Mojo/IOLoop/resources"))')
  cp --no-clobber "$mojo_resources"/server.crt /etc/apache2/ssl.crt/server.crt ||:
  cp --no-clobber "$mojo_resources"/server.key /etc/apache2/ssl.key/server.key ||:
  cp --no-clobber "$mojo_resources"/server.crt /etc/apache2/ssl.crt/ca.crt ||:
```

Maybe it makes sense to do something similar in our packaging.

However, I'm also fine with just commenting the code out. That's apparently how the apache2 config template already looks like and it was good enough so far.

Martchus (approved):

etc/nginx/vhosts.d/openqa.conf.template
asdil12: I'd go with the commenting out solution then. A TLS setup with selfsigned certs is not really useful anyway and for SSL setup manual adjusting of the config is required if you use nginx.

Maybe it's time to support caddy webserver which has a builtin ACME client :)

okurz (approved):

PR #5249, opened 2023-07-14
Make the job ID copying easier

pdostal:
A lot of online systems allow user to either copy to clipboard or easily select the job/ticket/case ID.
My intention is to enable the same functionality for openQA, so when user wants to f.e. clone the job it will be faster.

There is [fa-copy](https://fontawesome.com/icons/copy?f=classic&s=solid) icon in All tests and in Job group overview and there is job ID printed in job overview and is clickable.

image
image
image

### DIFF
--- a/assets/javascripts/tests.js
+++ b/assets/javascripts/tests.js
@@ -57,6 +57,8 @@ function renderTestName(data, type, row) {
 
   var html = '';
   if (is_operator) {
+    html += '<a class="copy-jobid" href="#" data-jobid="' + row.id + '">';
+    html += '<i class="action fa fa-fw fa-copy" title="Copy job id"></i></a>';
     if (row.result !== 'none') {
       // allow to restart finished jobs
       if (!row.clone) {
@@ -554,3 +556,8 @@ function showJobDependency(deps) {
   }
   return result;
 }
+
+$(document).on('click', '.copy-jobid', function (event) {
+  event.preventDefault();
+  navigator.clipboard.writeText(this.dataset.jobid);
+});
--- a/templates/webapi/test/infopanel.html.ep
+++ b/templates/webapi/test/infopanel.html.ep
@@ -4,6 +4,7 @@
                 % if (current_route 'latest') {
                     <%= link_to $job->id => url_for ('test', testid => $job->id) %>:
                 % }
+                <i><a class="copy-jobid" href="#" data-jobid="<%= $job->id %>">#<%= $job->id %></a></i>
                 %= $job->name
                 % if ($job->archived) {
                     <span id="job-archived-badge" class="badge badge-pill badge-secondary" title="Job has been archived">
--- a/assets/stylesheets/overall.scss
+++ b/assets/stylesheets/overall.scss
@@ -249,6 +249,11 @@ h1, h2, h3, h4, h5, h6, .h1, .h2, .h3, .h4, .h5, .h6 {
     min-width: 100%;
 }
 
+// Icon for copying job ID
+.copy-jobid {
+    margin-right: 5px;
+}
+
 .darkmode {
     .table {
         color: $default-font-color-darktheme;
--- a/templates/webapi/test/tr_job_result.html.ep
+++ b/templates/webapi/test/tr_job_result.html.ep
@@ -1,6 +1,8 @@
 <td id="res_<%= $resultid %>" name="jobid_td_<%= $jobid %>">
      % if ($res) {
          % if (is_operator) {
+         <a class="copy-jobid" href="#" data-jobid="<%= $jobid %>">
+         <i class="action fa fa-fw fa-copy" title="Copy job id"></i></a>
              % if ($state eq DONE || $state eq CANCELLED) {
                  %= link_post url_for('apiv1_restart', jobid => $jobid) => ('data-remote' => 'true', class => 'restart', 'data-jobid' => $jobid) => begin
                  <i class="action fa fa-undo" title="Restart the job"></i>


assets/javascripts/tests.js
```diff
@@ -554,3 +556,9 @@ function showJobDependency(deps) {
   }
   return result;
 }
+
+$(document).on('click', '.copy-jobid', function (event) {
+  event.preventDefault();
+  var id = $(this).attr('data-jobid');
+  navigator.clipboard.writeText(id);
```
Martchus, line 563: Use of `var` should be avoided in new code, as well as use of jQuery when it is useless anyways.

```suggestion
  navigator.clipboard.writeText(this.dataset.jobid);
```

assets/javascripts/tests.js
```diff
@@ -57,6 +57,8 @@ function renderTestName(data, type, row) {
 
   var html = '';
   if (is_operator) {
+    html += ' <a class="copy-jobid" href data-jobid="' + row.id + '">';
```
Martchus, line 60:
```suggestion
    html += '<a class="copy-jobid" href data-jobid="' + row.id + '">';
```

templates/webapi/test/infopanel.html.ep
```diff
@@ -4,6 +4,7 @@
                 % if (current_route 'latest') {
                     <%= link_to $job->id => url_for ('test', testid => $job->id) %>:
                 % }
+                <i>#<a class="copy-jobid" href data-jobid="<%= $job->id %>"><%= $job->id %></a></i>
```
asdil12, line 7:
```suggestion
                <i><a class="copy-jobid" href data-jobid="<%= $job->id %>">#<%= $job->id %></a></i>
```

Looks better and makes more sense if hash is part of the id.
Also you could set the href to link to the current job.

asdil12 (changes requested):

pdostal:
Resolved. I also added `padding-right: 5px;` so the icons are a bit further from the stop / restart icon. 
image

asmorodskyi:
LGTM!

perlpunk (approved):

assets/javascripts/tests.js
```diff
@@ -57,6 +57,8 @@ function renderTestName(data, type, row) {
 
   var html = '';
   if (is_operator) {
+    html += '<a class="copy-jobid" href="' + urlWithBase('/tests/' + row.id) + '" data-jobidi="' + row.id + '">';
```
b10n1k, line 60: i believe *jobidi* has a typo! dir you mean `jobid`?

b10n1k (changes requested):

assets/javascripts/tests.js
```diff
@@ -57,6 +57,8 @@ function renderTestName(data, type, row) {
...
+    html += '<i class="action fa fa-fw fa-copy" title="Copy job id"></i></a>';
```
okurz, line 61: This is the full "job URL", not just ID, right?

okurz (changes requested):

assets/javascripts/tests.js
pdostal: Yes, the `<a href=` points to the full job id but if you're javascript is enabled then it will not be followed.
That was suggested by @asdil12 [here](https://github.com/os-autoinst/openQA/pull/5249#discussion_r1263890961).

b10n1k (approved):

assets/javascripts/tests.js
okurz: I am not sure I understand. But if the end result is the full job URL then please call it as such in code as well as in the help and git commit message

pdostal: In the `href=""` is the full test URL in case your javascript is off.
But if the javascript is working then only the job id is put to your clipboard.

pdostal: If you need job URL it's two clicks to copy it. But if you f.e. want to clone 10 different jobs, then it is very time consuming to copy the job URL, parse the job ID, store the job ID and repeat 10 times. So this button only gets you the job id.

pdostal: Also people usually reuse commands from their history, so for example `openqa-clone-job https://openqa.opensuse.org/tests/11560397 --host https://pdostal-server.suse.cz --show-progress --skip-chained-deps` and if I want to do 5 verification runs it's usually only the job ID to replace.

grisu48 (approved):

assets/javascripts/tests.js
okurz: Yeah, sorry, I don't think that's a use case I would like feature *more* prominently to just copy the ID instead of the URL as that sounds like a rather personal choice of how you process command lines. Instead what I suggest to do commonly is to iterate over full job URLs, something like:

```bash
for i in https://openqa.opensuse.org/tests/11560397 https://openqa.opensuse.org/tests/11560234 https://openqa.opensuse.org/tests/11560402; do
openqa-clone-job $i --host https://pdostal-server.suse.cz --show-progress --skip-chained-deps
done
```

pdostal: Your command work when you want to clone bunch of jobs at once but if I want to clone only one job and next job in half an hour you end up replacing the URL over and over.

Then let's say I want to monitor a job in my command line: `openqa-cli api --osd --pretty jobs/$job_id`
Oh and then I have review and I want to use the awesome [grisu48/openqa-mon](https://github.com/grisu48/openqa-mon): `openqa-mon http://openqa.opensuse.org -j $ID1,$ID2,$ID3,...`

I've so far received quite positive feedback.
If the icon is in wrong place, we can move it.

Basically every system dealing with jobs or tickets displays the ID somewhere. And I think there's a reason for that. See:
Screenshot 2023-07-18 at 11 48 28
Screenshot 2023-07-18 at 11 46 40
Screenshot 2023-07-18 at 11 44 58
Screenshot 2023-07-18 at 11 44 03
Screenshot 2023-07-18 at 12 01 32

grisu48: I also think this feature might turn out to be useful to more folk than just Pavel :slightly_smiling_face:

pdostal:
I just got rid of the `<a href="JOBID"` in favour of `<a href="#"` so the link is still valid but the code looks less confusing.

mimi1vx (approved):

kalikiana (approved):

assets/javascripts/tests.js
okurz: Ok, convinced :D

asdil12 (approved):

Martchus (approved):

Martchus:
Removed the review request for @b10n1k as the mentioned typo has been fixed and there are more than enough reviews at this point.

PR #5250, opened 2023-07-14
contrib: Update/extend autoyast profile

baierjan:
Based on changed proposed in #5245 I converted both autoyast profiles (Leap and Tumbleweed) into one univeral profile for installing openqa-worker.

If we want, I can invest some more time into it and in next step enhance the profile to also configure and enable salt-minion to automatically connect to a predefined salt-master.


okurz (approved):
Looks good. We can improve later based on that. Maybe a fancy openQA-in-CI test testing the autoyast profile in every PR 🤩

Martchus (approved):

PR #5251, opened 2023-07-19
contrib: Extend and add autoyast profiles

okurz:


Martchus (changes requested):
The PR title doesn't match the commit messages at all.

okurz:
> The PR title doesn't match the commit messages at all.

It does now

contrib/ay-openqa-worker-leap-3-nvme.xml
```diff
@@ -0,0 +1,150 @@
...
+    </drive>
+    <drive>
+      <type t="symbol">CT_DISK</type>
+      <device>/dev/nvme0n1</device>
+      <disklabel>gpt</disklabel>
+      <use>all</use>
+      <partitions t="list">
+        <partition>
+          <size>max</size>
+          <filesystem t="symbol">ext4</filesystem>
```
mimi1vx, line 120: iont this reformatted by service to ext2 ?

mimi1vx (approved):

contrib/ay-openqa-worker-leap-3-nvme.xml
okurz: only if we actually use that service which is AFAIK only the case in OSD with salt

Martchus: I think ext4 is a good default. If it gets overridden that's not a problem.

Martchus (approved):

PR #5252, opened 2023-07-19
Don't use apache specific www group in fetchneedles

asdil12:
https://progress.opensuse.org/issues/131024
Fixes errors like https://openqa.opensuse.org/tests/3439538#step/openqa_bootstrap/14

### DIFF
--- a/script/fetchneedles
+++ b/script/fetchneedles
@@ -1,6 +1,6 @@
 #!/bin/sh -e
 : "${dbuser:="geekotest"}"
-: "${dbgroup:="www"}"
+: "${dbgroup:="nogroup"}"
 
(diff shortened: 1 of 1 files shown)


script/fetchneedles
```diff
@@ -1,6 +1,6 @@
 #!/bin/sh -e
 : "${dbuser:="geekotest"}"
-: "${dbgroup:="www"}"
+: "${dbgroup:="nogroup"}"
```
kalikiana, line 3: I'd say just change it on a worker and see if that works. Then it's fine with me

asdil12: Tested locally in an nspawn container and cloned a job, which worked fine.

perlpunk (approved):

PR #5257, opened 2023-07-20
tests: Improve failure when comparing expected image results fails

kalikiana:
A helpful failure might look something like this:

Using `diag explain` here so we're still checking the whole structure.

See: https://progress.opensuse.org/issues/132410

### DIFF
--- a/t/ui/18-tests-details.t
+++ b/t/ui/18-tests-details.t
@@ -151,7 +151,8 @@ subtest 'show job modules execution time' => sub {
 subtest 'displaying image result with candidates' => sub {
     $driver->find_element('[href="#step/bootloader/1"]')->click();
     wait_for_ajax;
-    is_deeply(find_candidate_needles, {'inst-bootmenu' => []}, 'correct tags displayed');
+    my $needles = find_candidate_needles;
+    is_deeply($needles, {'inst-bootmenu' => []}, 'correct tags displayed') or diag explain $needles;
 };
 
 subtest 'filtering' => sub {


t/ui/18-tests-details.t
```diff
@@ -151,7 +151,8 @@ subtest 'show job modules execution time' => sub {
 subtest 'displaying image result with candidates' => sub {
     $driver->find_element('[href="#step/bootloader/1"]')->click();
     wait_for_ajax;
-    is_deeply(find_candidate_needles, {'inst-bootmenu' => []}, 'correct tags displayed');
+    my $needles = find_candidate_needles;
+    is_deeply(keys %$needles, @{['inst-bootmenu']}, 'correct tags displayed');
```
perlpunk, line 155:
```suggestion
   is_deeply(find_candidate_needles, {'inst-bootmenu' => []}, 'correct tags displayed') or diag explain $needles;
```

perlpunk (changes requested):
`is_deeply` with just two lists won't work as you might expect.
It needs two references.
https://metacpan.org/pod/Test::More#is_deeply

perlpunk:
Btw, if we switched to [Test2](https://metacpan.org/pod/Test2::V0) then the `is` function already provides a deep comparison  which doesn't stop at the first mismatch, but lists all top level mismatches.
```
my $x = { foo => { x => 1 } };
my $y = { bar => [] };
is $x, $y;
__END__
# Failed test at -e line 3.
# +-------+----------------------+---------+------------------+
# | PATH  | GOT                  | OP      | CHECK            |
# +-------+----------------------+---------+------------------+
# | {bar} | <DOES NOT EXIST>     |         | <ARRAY>          |
# | {foo} | HASH(0x55fa58aca840) | !exists | <DOES NOT EXIST> |
# +-------+----------------------+---------+------------------+

```

t/ui/18-tests-details.t
perlpunk: Also not sure what `@{['inst-bootmenu']}` is supposed to achieve. It's creating an array ref and dereferences it immediately.

kalikiana: Well, you can see in the description what the result is when it fails. Without the `@{}` all you get is `ARRAY(0xaaab25f52718`

perlpunk: `@{['inst-bootmenu']}` is just the same as `'inst-bootmenu'`.

perlpunk: What happens if `keys %$needles` returns more than one key? This usage of `is_deeply` wrong.
Say if it has three keys, then your code would be equivalent to:
```
    is_deeply('key1', 'key2', 'key3', 'inst-bootmenu', 'correct tags displayed');
```
What should that do?

t/ui/18-tests-details.t
```diff
@@ -151,7 +151,8 @@ subtest 'show job modules execution time' => sub {
 subtest 'displaying image result with candidates' => sub {
     $driver->find_element('[href="#step/bootloader/1"]')->click();
     wait_for_ajax;
-    is_deeply(find_candidate_needles, {'inst-bootmenu' => []}, 'correct tags displayed');
+    my $needles = find_candidate_needles;
+    is(join(' ', keys %$needles), 'inst-bootmenu', 'correct tags displayed');
```
perlpunk, line 155: I'm still wondering what's wrong with my suggestion?
```suggestion
    is_deeply($needles, {'inst-bootmenu' => []}, 'correct tags displayed') or diag explain $needles;
```
I'm not sure about how important this is for this test, but previously the test did not only check if a certain key exists, but also if the value is an empty array ref.

perlpunk (approved):

perlpunk:
You should rebase
https://app.circleci.com/pipelines/github/os-autoinst/openQA/11973/workflows/686476b4-4317-41cb-a309-10d126910356/jobs/111940

Martchus (approved):

PR #5258, opened 2023-07-20
Display line breaks in data tables

perlpunk:
Some values might contain line breaks, for example the relatively new setting SCENARIO_DEFINITIONS_YAML. Before everything was printed in one line. Line breaks got lost.

## Before:
settings-old
## After:
settings-newlines

### DIFF
--- a/assets/stylesheets/tables.scss
+++ b/assets/stylesheets/tables.scss
@@ -63,7 +63,11 @@
         text-align: center;
     }
 }
-
+table.settings-table tr {
+    td span {
+        white-space: pre;
+    }
+}
 table.dataTable {
     td.name {
         text-align: left;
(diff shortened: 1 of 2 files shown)


assets/stylesheets/tables.scss
```diff
@@ -63,7 +63,11 @@
         text-align: center;
     }
 }
-
+table.dataTable tr {
+    td span {
+        white-space: pre;
```
Martchus, line 68: This rule seems to generic. I don't think we want to match *every* span (which is one of the most generic elements) in *every* data-table.

perlpunk: Yeah, I was kind of expecting that it would be to generic.

perlpunk: I added a settings-table class now

Martchus (approved):

okurz (approved):

PR #5262, opened 2023-07-24
Use job video and result presence markers to improve limiting

AdamWill:
As currently implemented, `_ensure_results_below_threshold` will constantly go over the same old jobs it's already gone over, again and again, until they're garbage collected and entirely gone from the database.

This uses new `videos_present` and `results_present` booleans to let the limit task mark when it's already tried to remove videos and/or results for a job, and then exclude that job from future runs via the 'relevant jobs' db query, so it won't constantly go over the same ground.

### DIFF
--- a/dbicdh/PostgreSQL/deploy/101/001-auto-__VERSION.sql
+++ b/dbicdh/PostgreSQL/deploy/101/001-auto-__VERSION.sql
@@ -1,0 +1,18 @@
+--
+-- Created by SQL::Translator::Producer::PostgreSQL
+-- Created on Mon Jul 24 10:52:38 2023
+--
+;
+--
+-- Table: dbix_class_deploymenthandler_versions
+--
+CREATE TABLE dbix_class_deploymenthandler_versions (
+  id serial NOT NULL,
+  version character varying(50) NOT NULL,
+  ddl text,
+  upgrade_sql text,
+  PRIMARY KEY (id),
+  CONSTRAINT dbix_class_deploymenthandler_versions_version UNIQUE (version)
+);
+
+;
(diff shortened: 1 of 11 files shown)


AdamWill:
@Martchus wdyt of this? I...*think* it makes sense. I started (trying to) use this limit task on Fedora's instances recently as we're getting close to filling up the (huge) shares we use for testresults and images. On our prod instance the task has been running for over a day and only seems to be up to 01496 in testresults so far, going by the directory mtimes (the oldest dir still present is 01329, and the highest is 02030). So...it seems like something like this might be useful? :D

AdamWill:
This PR is affected by https://progress.opensuse.org/issues/133238 , too. It takes the approach I outlined in https://progress.opensuse.org/projects/openqav3/activity?from=2023-07-24 . It adds an sql script to the 92-93 upgrade which adds the necessary columns before the perl script runs. The auto-generated script for the 100-101 upgrade is modified to add the columns only if they don't already exist. Note, this syntax apparently requires postgresql 9.6 or higher. This is the best idea I can come up with, but if anyone has a better one I'm all ears!

Martchus (approved):
This change looks good to me. I suppose the additional database columns are worth it. I have nevertheless added the "not-ready" label to give more people a chance to review.

AdamWill:
For the record, it turned out the space usage on our prod instance which inspired me to turn this cleanup on and propose this improvement turned out to be caused by Netapp snapshotting, so I'm no longer using this cleanup and don't, for now, have any direct need for the optimization. If it's merged and causes any problems I will definitely look at them, but I'm also OK if it's not merged, I'll leave it up to you folks to decide. thanks!

okurz (changes requested):
Ok, with that good reasoning I am also in favor of not merging

Martchus:
I'd concur to not merge this now. This usually big table has already enough columns and we should only add new ones if we have to.

okurz:
Settled then

PR #5274, opened 2023-08-07
Add timezone to SQL query

perlpunk:
DBIx::Class would to the right thing, but we're using DBI directly, and the stringified DateTime object will by default not have the timezone offset appended.

Related issue: https://progress.opensuse.org/issues/133889


kalikiana (approved):
🤦🏾‍♀️

okurz (approved):

PR #5279, opened 2023-08-10
Show max running jobs on /tests page

perlpunk:
See individual commits

~~I think we could also show the actual limit value, e.g. `x jobs are running (limited by server config: 250)`
WDYT?~~

### DIFF
--- a/assets/javascripts/tests.js
+++ b/assets/javascripts/tests.js
@@ -264,7 +264,11 @@ function renderTestLists() {
       data: ajaxQueryParams,
       dataSrc: function (json) {
         // update heading when JSON is available
-        $('#running_jobs_heading').text(json.data.length + ' jobs are running');
+        let text = json.data.length + ' jobs are running';
+        if (json.max_running_jobs !== undefined && json.max_running_jobs >= 0) {
+          text += ' (limited by server config)';
+        }
+        $('#running_jobs_heading').text(text);
         return json.data;
       }
     },
--- a/lib/OpenQA/Scheduler/Model/Jobs.pm
+++ b/lib/OpenQA/Scheduler/Model/Jobs.pm
@@ -43,8 +43,8 @@ sub schedule ($self, $allocated_workers = {}, $allocated_jobs = {}) {
     my $worker_count = $schema->resultset('Workers')->count;
     my $free_worker_count = @$free_workers;
     my $running = $schema->resultset('Jobs')->count({state => [OpenQA::Jobs::Constants::EXECUTION_STATES]});
-    my $limit = $self->{config}->{scheduler}->{max_running_jobs} // '';
-    if (length $limit and $running >= $limit) {
+    my $limit = $self->{config}->{scheduler}->{max_running_jobs} // -1;
+    if ($limit >= 0 && $running >= $limit) {
         log_debug("max_running_jobs ($limit) exceeded, scheduling no additional jobs");
         $self->emit('conclude');
         return;
--- a/lib/OpenQA/WebAPI/Controller/Test.pm
+++ b/lib/OpenQA/WebAPI/Controller/Test.pm
@@ -247,7 +247,10 @@ sub list_running_ajax {
                 progress => $job->progress_info,
             });
     }
-    $self->render(json => {data => \@running});
+    my %response = (data => \@running);
+    my $max_running = OpenQA::App->singleton->config->{scheduler}->{max_running_jobs} // -1;
+    $response{max_running_jobs} = $max_running if $max_running >= 0 && @running >= $max_running;
+    $self->render(json => \%response);
 }
 
 sub list_scheduled_ajax {
--- a/etc/openqa/openqa.ini
+++ b/etc/openqa/openqa.ini
@@ -282,8 +282,8 @@ concurrent = 0
 [scheduler]
 # Specify how many days a job can stay in 'SCHEDULED' state, defaults to 7 days
 # max_job_scheduled_time = 7
-# Specify how many jobs in total can run per webui instance. Defaults to no limit
-# max_running_jobs = 
+# Specify how many jobs in total can run per webui instance. Defaults to -1 (no limit)
+# max_running_jobs = -1
 
 # Configuration of the label/bugref carry-over
 [carry_over]
--- a/lib/OpenQA/Setup.pm
+++ b/lib/OpenQA/Setup.pm
@@ -66,6 +66,7 @@ sub read_config ($app) {
         },
         'scheduler' => {
             max_job_scheduled_time => 7,
+            max_running_jobs => -1,
         },
         logging => {
             level => undef,
--- a/t/config.t
+++ b/t/config.t
@@ -66,6 +66,7 @@ subtest 'Test configuration default modes' => sub {
         },
         'scheduler' => {
             max_job_scheduled_time => 7,
+            max_running_jobs => -1,
         },
         openid => {
             provider => 'https://www.opensuse.org/openid/user/',
--- a/t/ui/01-list.t
+++ b/t/ui/01-list.t
@@ -8,6 +8,7 @@ use FindBin;
 use lib "$FindBin::Bin/../lib", "$FindBin::Bin/../../external/os-autoinst-common/lib";
 use Date::Format 'time2str';
 use Test::Mojo;
+use Mojo::File qw(tempdir);
 use Test::Warnings ':report_warnings';
 use Time::Seconds;
 use OpenQA::Test::TimeLimit '40';
@@ -24,6 +25,9 @@ my $schema = $test_case->init_data(schema_name => $schema_name, fixtures_glob =>
 use OpenQA::SeleniumTest;
 
 my $t = Test::Mojo->new('OpenQA::WebAPI');
+
+$ENV{OPENQA_CONFIG} = my $config_dir = tempdir("$FindBin::Script-XXXX");
+$config_dir->child('openqa.ini')->spurt("[scheduler]\nmax_running_jobs = 3");
 
 my @job_params = (
     group_id => 1002,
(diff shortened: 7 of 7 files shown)


lib/OpenQA/WebAPI/Controller/Test.pm
```diff
@@ -247,7 +247,13 @@ sub list_running_ajax {
                 progress => $job->progress_info,
             });
     }
-    $self->render(json => {data => \@running});
+    my %response = (data => \@running);
+    my $max_running = OpenQA::App->singleton->config->{scheduler}->{max_running_jobs} // -1;
```
kalikiana, line 251: Why is `// -1` needed here?

mimi1vx: because its default if isnt set in config ?

mimi1vx (approved):

lib/OpenQA/WebAPI/Controller/Test.pm
perlpunk: I'm actually not sure what happens if someone sets it to the empty string, then we would get a warning.
But I assume if it is not set at all, it would get our default from Setup.m, so I guess this is not strictly needed...

assets/javascripts/tests.js
```diff
@@ -264,7 +264,11 @@ function renderTestLists() {
       data: ajaxQueryParams,
       dataSrc: function (json) {
         // update heading when JSON is available
-        $('#running_jobs_heading').text(json.data.length + ' jobs are running');
+        let text = json.data.length + ' jobs are running';
+        if (json.max_running_jobs >= 0) {
+          text += ' (limited by server config)';
```
kalikiana, line 269: How about `x/y jobs are running` where **x** is the number of jobs and **y** is the configured maximum

perlpunk:
> I think we could also show the actual limit value, e.g. `x jobs are running (limited by server config: 250)` WDYT?

Oh wait, since this is only shown when the limit is reached I think this is redundant :)

assets/javascripts/tests.js
perlpunk: https://github.com/os-autoinst/openQA/pull/5279#issuecomment-1673322996

lib/OpenQA/WebAPI/Controller/Test.pm
perlpunk: I removed it

lib/OpenQA/WebAPI/Controller/Test.pm
```diff
@@ -247,7 +247,13 @@ sub list_running_ajax {
                 progress => $job->progress_info,
             });
     }
-    $self->render(json => {data => \@running});
+    my %response = (data => \@running);
+    my $max_running = OpenQA::App->singleton->config->{scheduler}->{max_running_jobs};
+    if ($max_running >= 0) {
+        my $running = $self->schema->resultset('Jobs')->count({state => [OpenQA::Jobs::Constants::EXECUTION_STATES]});
+        $response{max_running_jobs} = $max_running if $running >= $max_running;
+    }
```
Martchus, line 255: I'd avoid the additional database query. We have just populated an array of running jobs so let's just use its size:
```suggestion
    $response{max_running_jobs} = @running if $max_running >= 0 && @running >= $max_running;
```

lib/OpenQA/WebAPI/Controller/Test.pm
perlpunk: Hm, some tests are now failing: https://app.circleci.com/pipelines/github/os-autoinst/openQA/12040/workflows/04e65a20-0920-4b4d-ba1b-0d7b7f4e2e3d/jobs/112482
```
Use of uninitialized value $limit in numeric ge (>=) at /home/squamata/project/lib/OpenQA/Scheduler/Model/Jobs.pm line 47.
Use of uninitialized value $limit in concatenation (.) or string at /home/squamata/project/lib/OpenQA/Scheduler/Model/Jobs.pm line 48.
Use of uninitialized value $limit in numeric ge (>=) at /home/squamata/project/lib/OpenQA/Scheduler/Model/Jobs.pm line 47.
Use of uninitialized value $limit in numeric ge (>=) at /home/squamata/project/lib/OpenQA/Scheduler/Model/Jobs.pm line 47.
Use of uninitialized value $limit in concatenation (.) or string at /home/squamata/project/lib/OpenQA/Scheduler/Model/Jobs.pm line 48.
```
I thought OpenQA::Setup would ensure the default value of -1?

Martchus (changes requested):
I'm fine with printing just "… (limited by server config)" in case the limit is effected. Otherwise it would be rather redundant. I'd also avoid a "x / y" format unless we are really sure not to end up with e.g. "201 / 200" in production (which would look rather buggy).

assets/javascripts/tests.js
```diff
@@ -264,7 +264,11 @@ function renderTestLists() {
       data: ajaxQueryParams,
       dataSrc: function (json) {
         // update heading when JSON is available
-        $('#running_jobs_heading').text(json.data.length + ' jobs are running');
+        let text = json.data.length + ' jobs are running';
+        if (json.max_running_jobs >= 0) {
```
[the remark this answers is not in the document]
perlpunk: No, like I wrote here: https://github.com/os-autoinst/openQA/pull/5279#discussion_r1290260950

But I thought Setup.pm would assure a default value.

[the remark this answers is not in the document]
perlpunk: The value is only passed to the client side if it's not undef, so that's ok.

lib/OpenQA/WebAPI/Controller/Test.pm
perlpunk: Like I wrote in the commit message, I can't use that value because you can limit the query, e.g. `/tests?match=extra`.
Otherwise I would also avoid the query.

lib/OpenQA/WebAPI/Controller/Test.pm
perlpunk: So, I added the `//=1` again.

lib/OpenQA/Scheduler/Model/Jobs.pm
```diff
@@ -43,8 +43,8 @@ sub schedule ($self, $allocated_workers = {}, $allocated_jobs = {}) {
     my $worker_count = $schema->resultset('Workers')->count;
     my $free_worker_count = @$free_workers;
     my $running = $schema->resultset('Jobs')->count({state => [OpenQA::Jobs::Constants::EXECUTION_STATES]});
-    my $limit = $self->{config}->{scheduler}->{max_running_jobs} // '';
-    if (length $limit and $running >= $limit) {
+    my $limit = $self->{config}->{scheduler}->{max_running_jobs};
+    if ($limit >= 0 and $running >= $limit) {
```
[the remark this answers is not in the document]
perlpunk: I really like `and` because of it's clear lower precedence, but you won't give up :)

assets/javascripts/tests.js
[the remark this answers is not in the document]
perlpunk: hm, you commented twice on the same code with almost the same message? That's weird.
My answer was targeted at the perl code.

perlpunk:
So I investigated the problem with max_running_jobs not always being set to the default value, causing warnings.
That is actually a problem of the tests, and it wouldn't happen in the productive code.
The problem is that apparemtly OpenQA::Setup::read_config is not always called when we create app instances in tests.

I think that's unfortunate because we have to cover for that in code by ensuring a default value in several places, although that should be done by OpenQA::Setup in just one place.
It would be good to make sure in tests we read the config and fill in defaults.

I will create a ticket because I had a quivk look, and it looks to me like it's not fixable very fast.

lib/OpenQA/Scheduler/Model/Jobs.pm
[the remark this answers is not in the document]
kalikiana: > I'd also avoid a "x / y" format unless we are really sure not to end up with e.g. "201 / 200" in production (which would look rather buggy).

Didn't we discuss that the code path is already behind checks for the number of running jobs? It seems odd to me to avoid it just because we may not know about bugs that break it.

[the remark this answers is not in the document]
perlpunk: We could very well end up with 201 / 200, for example when we introduce the limit or lower it and the number of running jobs is still higher.

Btw., the comment seems misplaced here. This review comment from @Martchus is about `and` vs. `&&`.

okurz (approved):

lib/OpenQA/WebAPI/Controller/Test.pm
Martchus: Mh… that makes sense. It is nevertheless bad that we have to do a count on a huge table and a column we don't have an index for. I guess it is ok considering the code path is only entered when the setting is configured.

assets/javascripts/tests.js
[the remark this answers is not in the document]
Martchus: I know, hence `json.max_running_jobs` might be `undefined` on the client-side. Supposedly the `>=` operator deals with it but it would be good to make it more clear that this object property might be `undefined`.

lib/OpenQA/Scheduler/Model/Jobs.pm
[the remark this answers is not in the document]
kalikiana: You pointed me to this comment ;-)

Anyway feel free to override me if you're on the same page with Marius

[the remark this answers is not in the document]
perlpunk: No, I pointed to https://github.com/os-autoinst/openQA/pull/5279#issuecomment-1673322996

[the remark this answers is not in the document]
perlpunk: To clarify: It would show `250 / 250 jobs running` and possibly sometimes `260 / 250`, but never `200 / 250`, so Marius and I really thing that wouldn't make sense.
Meanwhile in the meeting Marius and I also decided that it's best to only show the notice if the query is not limited by `match` or `group_id` parameters. Because then it would show e.g. `3 jobs running (limited by server config)`, even if there are 250 jobs in total.
Will push an update later today.

assets/javascripts/tests.js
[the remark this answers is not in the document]
perlpunk: done

lib/OpenQA/WebAPI/Controller/Test.pm
perlpunk: We talked about it, and while `state` is indexed, we only want the notice on a page without any filters (otherwise it would be confusing), and in this case using the existing `@running` variable does the job.
Pushed.

Martchus (approved):

PR #5280, opened 2023-08-11
Log number of remaining retry attempts for asset uploads

kraih:
While retries for asset chunks do work correctly, the messages being logged for failures can be a little confusing. By adding the number of remaining retries in a way similar to other log messages, it gets a lot easier to spot what is really going on. And that the "errors" are actually fairly harmless given the context.

Once this has proven itself in production, we might want to redesign the event interface a bit and get rid of a few events. There is a bit of redundancy right now in the logging. But i can see how that might help find issues in the upload logic as it stands right now.

I've also cleaned up the upload code a little bit to bring it closer to our current standards, without making changes to the API (yet).

Progress: https://progress.opensuse.org/issues/132167

### DIFF
--- a/lib/OpenQA/Client/Upload.pm
+++ b/lib/OpenQA/Client/Upload.pm
@@ -9,7 +9,7 @@ use Carp qw(croak);
 use Mojo::Asset::Memory;
 use Mojo::File qw(path);
 
-has max_retrials => 5;
+has max_retries => 5;
 
 sub _upload_asset_fail ($self, $uri, $form) {
     $form->{state} = 'fail';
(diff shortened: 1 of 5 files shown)


kraih:
I also want to increase the retry timeout or the number of attempts, but not sure yet which one would be better here.

kalikiana (approved):

kalikiana:
> I also want to increase the retry timeout or the number of attempts, but not sure yet which one would be better here.

I feel like this may be nice to make configurable. Not necessarily in this branch. I'd leave it to you, though.

Martchus (approved):

PR #5282, opened 2023-08-15
Add timezone to SQL

perlpunk:
For some reason I was only looking at one of the parameters...

Issue: https://progress.opensuse.org/issues/133889

### DIFF
--- a/lib/OpenQA/WebAPI/Controller/Admin/Influxdb.pm
+++ b/lib/OpenQA/WebAPI/Controller/Admin/Influxdb.pm
@@ -110,7 +110,7 @@ sub minion ($self) {
 		  WHERE finished >= ? AND finished < ? AND task = 'hook_script' AND
 		        state = 'finished' AND (notes->'hook_rc')::int != 0}
     );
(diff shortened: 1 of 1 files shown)


lib/OpenQA/WebAPI/Controller/Admin/Influxdb.pm
```diff
@@ -110,7 +110,7 @@ sub minion ($self) {
 		  WHERE finished >= ? AND finished < ? AND task = 'hook_script' AND
 		        state = 'finished' AND (notes->'hook_rc')::int != 0}
     );
-    $sth->execute($rc_fail_timespan_start, "$rc_fail_timespan_end+0");
+    $sth->execute("$rc_fail_timespan_start+0", "$rc_fail_timespan_end+0");
```
kalikiana, line 113: I wonder how we can make unit tests catch this 🤔

mimi1vx (approved):

lib/OpenQA/WebAPI/Controller/Admin/Influxdb.pm
kalikiana: ^^ I'll wonder in another branch. No need to block the PR

PR #5283, opened 2023-08-16
tests: Specify timezone for the test database

kalikiana:
This way we can test that the worker timezone can deviate from the timezone of the database host.

### DIFF
--- a/lib/OpenQA/WebAPI/Controller/Admin/Influxdb.pm
+++ b/lib/OpenQA/WebAPI/Controller/Admin/Influxdb.pm
@@ -110,7 +110,7 @@ sub minion ($self) {
 		  WHERE finished >= ? AND finished < ? AND task = 'hook_script' AND
 		        state = 'finished' AND (notes->'hook_rc')::int != 0}
     );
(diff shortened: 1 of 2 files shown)


Martchus (approved):

kalikiana:
37-limit_assets.t seems to fail with and without the fixes in Influxdb.pm if the tz mismatches

```
Subtest: limit for keeping untracked assets is overridable in settings
    not ok 1 - default is 14 days
    not ok 2 - override works
Subtest: limits based on fine-grained filename-based patterns
    not ok 4 - default without pattern is 14 days
    not ok 6 - simple pattern override works
    not ok 7 - regex pattern matches 12.2
    not ok 8 - regex pattern matches 12.3
    not ok 10 - simple pattern half-way in
    not ok 12 - newer mtime takes precedence
```

`make test TESTS=t/api/13-influxdb.t` seems to fail with timezone=PST8PDT and the fixes in Influxdb **yet** it passed with UTC?

Martchus:
I'm closing this PR because with https://github.com/os-autoinst/openQA/commit/92cc3e3ca9af6240cb2c925b97b6dec7834ffa4b this should not be required anymore.

PR #5285, opened 2023-08-17
Fix scheduler getting max_running_jobs config

perlpunk:
We have to access the config via an OpenQA::App singleton.
$self->{config} simply isn't set, but I was assuming that because
I looked at _update_scheduled_jobs which uses

    my $max_job_scheduled_time = $self->{config}->{scheduler}->{max_job_scheduled_time}

So I relied that this would just work. That's also why I had problems to
set the config in the test.

Issue: https://progress.opensuse.org/issues/129619


Martchus:
It makes sense but it looks like `./t/05-scheduler-capabilities.t` needs to be adapted. Likely the app isn't set in this test.

perlpunk:
I will remove the commit that drops the defaults for now and make a new PR...

kraih (approved):

kalikiana (approved):

PR #5286, opened 2023-08-17
Add usleep_backoff function and use it for asset uploads

kraih:
This change introduces a slowly increasing backoff delay (5-30 seconds) into the asset upload retry logic to replace the hardcoded 5 second delay. Additonally a random 0-1 second padding will be used to decrease the chances of multiple workers retrying asset uploads at the exact same time. The algorithm was chosen fairly arbitrarily, i just wanted something that increases linearly and is easy to understand.

Since O3/OSD service restarts usually take around a minute, i've also increased the number of retries from 5 to 10. That should give all our workers more than enough time to recover from any expected webui downtime by default.

Progress: https://progress.opensuse.org/issues/132167

### DIFF
--- a/lib/OpenQA/Utils.pm
+++ b/lib/OpenQA/Utils.pm
@@ -31,6 +31,7 @@ use File::Spec::Functions qw(catfile catdir);
 use Fcntl;
 use Mojo::JSON qw(encode_json decode_json);
 use Mojo::Util 'xml_escape';
+use List::Util qw(min);
 
 my $FRAG_REGEX = FRAGMENT_REGEX;
 
@@ -79,6 +80,8 @@ BEGIN {
 use constant UNCONSTRAINED_BUGREF_REGEX => $BUGREF_REGEX;
 use constant BUGREF_REGEX => qr{(?:^|(?<=\s|,))$BUGREF_REGEX(?![\w\"])};
 use constant LABEL_REGEX => qr/\blabel:(?<match>([\w:#]+))\b/;
+
+use constant ONE_SECOND_IN_MICROSECONDS => 1_000_000;
 
 our $VERSION = sprintf "%d.%03d", q$Revision: 1.12 $ =~ /(\d+)/g;
 our @EXPORT = qw(
@@ -152,6 +155,7 @@ our @EXPORT_OK = qw(
   random_string
   random_hex
   regex_match
+  usleep_backoff
 );
 
 # override OPENQA_BASEDIR for tests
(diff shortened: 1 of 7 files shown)


mimi1vx (approved):

lib/OpenQA/Utils.pm
```diff
@@ -958,4 +961,18 @@ sub regex_match ($regex_string, $string) {
...
+# Returns a microsecond value suitable for use with "usleep", for the first iteration it returns the minimum value plus
+# random 0-1 second padding, and starting with the second iteration the delay increases by one second plus 0-1 second
+# padding, up to the maximum
+sub usleep_backoff ($iteration, $min_seconds, $max_seconds, $padding = int(rand(ONE_SECOND_IN_MICROSECONDS))) {
+
+    # To allow for backoff to be disabled easily
+    return 0 if $min_seconds == 0;
+
+    my $delay
+      = ($min_seconds * ONE_SECOND_IN_MICROSECONDS) + (ONE_SECOND_IN_MICROSECONDS * ($iteration - 1)) + $padding;
```
Martchus, line 973:
```suggestion
    my $delay = ($min_seconds + $iteration - 1) * ONE_SECOND_IN_MICROSECONDS + $padding;
```

lib/OpenQA/Utils.pm
```diff
@@ -958,4 +961,18 @@ sub regex_match ($regex_string, $string) {
...
+    my $max = $max_seconds * ONE_SECOND_IN_MICROSECONDS;
+    return $delay > $max ? $max : $delay;
```
Martchus, line 975: One could `use List::Util qw(min);` here:
```suggestion
    return min($max, $delay);
```

lib/OpenQA/Utils.pm
```diff
@@ -958,4 +961,18 @@ sub regex_match ($regex_string, $string) {
     return $match;
 }
 
+# Returns a microsecond value suitable for use with "usleep", for the first iteration it returns the minimum value plus
```
Martchus, line 964:
```suggestion
# Returns a microsecond value suitable for use with "usleep". For the first iteration it returns the minimum value plus
```

Martchus (commented):
Makes sense. I only have a few coding style suggestions.

lib/OpenQA/Utils.pm
kraih: I tend to avoid `List::Util` when it doesn't make a big difference for readability, because in many cases it can turn into a performance regression (often because of fake code blocks). For `max` and `min` i'm actually not sure how well they scale.

kraih: I guess it's a little cleaner if i merge it with the previous line too. 🤔

perlpunk (approved):

lib/OpenQA/Utils.pm
perlpunk: Regarding performance, I think in this case we don't need to care because this code isn't called very often.
Also https://metacpan.org/pod/List::Util::XS says that it will use the XS version always.

Martchus: I would have expected that `List::Util` would be faster as it is likely implemented natively (so there's more native code executed and less Perl code). Seems like this is at least true at this point. I guess it really doesn't matter here, though.

kraih: Despite it being implemented in XS there are performance bottlenecks. For example `reduce {...} 1..10` is problematic because the `{...}` blocks here are not real blocks, but functions made to look like blocks. These have a lot more overhead than real blocks that truly native Perl features like `grep {...} 1..10` can use. This is similar to how `for` loops are preferable to iterator APIs like `$collection->for_each(sub ($item) {...})` because loops are so much more efficient.

PR #5288, opened 2023-08-23
Allow using force-result label with all final job states

Martchus:
In particular, it makes sense to allow this not only for done jobs but also for jobs that are canceled (and have never been assigned to a worker).

See https://progress.opensuse.org/issues/127280

### DIFF
--- a/lib/OpenQA/Schema/Result/Jobs.pm
+++ b/lib/OpenQA/Schema/Result/Jobs.pm
@@ -1049,9 +1049,11 @@ sub update_backend ($self, $backend_info) {
     $self->update({backend => $backend_info->{backend}});
 }
 
-sub update_result ($self, $result) {
-    my $res = $self->update({result => $result});
-    OpenQA::App->singleton->emit_event('openqa_job_update_result', {id => $self->id, result => $result}) if $res;
+sub update_result ($self, $result, $state = undef) {
+    my %values = (result => $result);
+    $values{state} = $state if defined $state;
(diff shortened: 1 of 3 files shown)


lib/OpenQA/Schema/Result/Jobs.pm
```diff
@@ -1049,9 +1049,11 @@ sub update_backend ($self, $backend_info) {
...
 }
 
-sub update_result ($self, $result) {
-    my $res = $self->update({result => $result});
-    OpenQA::App->singleton->emit_event('openqa_job_update_result', {id => $self->id, result => $result}) if $res;
+sub update_result ($self, $result, $state = undef) {
+    my @values = (result => $result);
+    push @values, state => $state if defined $state;
+    my $res = $self->update({@values});
+    OpenQA::App->singleton->emit_event('openqa_job_update_result', {id => $self->id, @values}) if $res;
```
perlpunk, line 1056: Not sure why you would use an array here. It would work of course, but a hash seems more natural.
```suggestion
    my %values = (result => $result);
    $values{state} = $state if defined $state;
    my $res = $self->update(\%values);
    OpenQA::App->singleton->emit_event('openqa_job_update_result', {id => $self->id, %values}) if $res;
```

kraih: I vote hash in this context.

kraih (approved):

PR #5289, opened 2023-08-23
Make max_running_jobs a hard limit

perlpunk:
Before the actual limit would have been max_running_jobs + MAX_JOB_ALLOCATION.

Note that any parallel cluster jobs might still get the number of running jobs
above that limit, but those are usually not more than a handful of jobs.

Issue: https://progress.opensuse.org/issues/129619

See individual commits

### DIFF
--- a/lib/OpenQA/Scheduler/Model/Jobs.pm
+++ b/lib/OpenQA/Scheduler/Model/Jobs.pm
@@ -127,8 +127,9 @@ sub schedule ($self, $allocated_workers = {}, $allocated_jobs = {}) {
         }
         # we make sure we schedule clusters no matter what,
         # but we stop if we're over the limit
-        my $busy = scalar(keys %$allocated_workers);
-        last if $busy >= MAX_JOB_ALLOCATION || $busy >= $free_worker_count;
+        my $busy = keys %$allocated_workers;
+        last
+          if $busy >= MAX_JOB_ALLOCATION || $busy >= $free_worker_count || $limit >= 0 && ($running + $busy) >= $limit;
     }
 
     my @successfully_allocated;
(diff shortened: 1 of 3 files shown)


Martchus (commented):
I guess it would still make sense to add tests for this - even if you have to extract some parts into separate functions.

perlpunk:
Ok, good idea, I'm moving the job/worker allocation in its own sub, then I can better test this, and the schedule sub is big enough already to split it anyway.

lib/OpenQA/Scheduler/Model/Jobs.pm
```diff
@@ -127,10 +117,37 @@ sub schedule ($self, $allocated_workers = {}, $allocated_jobs = {}) {
         }
         # we make sure we schedule clusters no matter what,
         # but we stop if we're over the limit
-        my $busy = scalar(keys %$allocated_workers);
-        last if $busy >= MAX_JOB_ALLOCATION || $busy >= $free_worker_count;
+        my $busy = keys %$allocated_workers;
+        if ($busy >= $max_allocate || $busy >= @$free_workers) {
+            my $free_worker_count = @$free_workers;
+            log_debug(
+"limit reached (max_running_jobs=$limit, free workers=$free_worker_count, running=$running, allocated=$busy), scheduling no additional jobs"
```
kraih, line 124: Maybe just split up the string to make it look nicer?

kalikiana (approved):

okurz:
code change looks fine. But the runtime of the test has increased from 5-6s to 7-8s (`count_fail_ratio prove -l t/04-scheduler.t`) . I suggest to re-evaluate if you can decrease the runtime again a bit.

okurz (approved):
As agreed with @perlpunk I will approve the functionally correct PR now and work on the unit test improvement immediately as follow-up as part of the same progress ticket.

PR #5291, opened 2023-08-29
Spike solution for filtering tests by job group

kraih:
Instead of regexes i've used globs, which are a little more user friendly and less prone to security issues. The implementation is a bit inefficient, since it loads all job groups into memory to filter them on the Perl side. But there will probably never be an openQA instance with enough job groups to justify the complexity of filtering in SQL.

Since this is a spike solution, tests and ui enhancements have been left out. The feature can only be used with query parameters in the URL for now, like `/tests/overview?group_glob=*kernel*&not_group_glob=*development*&todo=1`. Multiple `group_glob` and `not_group_glob` values can be used for more complex filtering.

Progress: https://progress.opensuse.org/issues/119746


lib/OpenQA/WebAPI/Plugin/Helpers.pm
```diff
@@ -435,6 +441,17 @@ sub _compose_job_overview_search_args ($c) {
         my @search_terms = (@group_id_search, @group_name_search);
         @groups = $schema->resultset('JobGroups')->search(\@search_terms)->all;
     }
+
+    # use globs to filter job groups, first include all groups that match "group_glob" values, and then exclude those
+    # that also match "not_group_glob" values
+    elsif (@$group_glob || @$not_group_glob) {
+        my @inclusive = map { qr/$_/i } map { glob_to_regex_string($_) } @$group_glob;
```
Martchus, line 448: Maybe avoid the double `map` calls (which in Perl likely result in an intermediate array, right?):
```suggestion
        my @inclusive = map { my $regex = glob_to_regex_string($_); qr/$regex/i } @$group_glob;
```

Martchus (commented):
Looks generally good (for a spike solution, to merge tests would be good).

lib/OpenQA/WebAPI/Plugin/Helpers.pm
kraih: Too cheap to worry about resource usage, the intermediate list costs nothing really. Most readable solution wins here imo.

lib/OpenQA/WebAPI/Plugin/Helpers.pm
```diff
@@ -396,6 +397,8 @@ sub _compose_job_overview_search_args ($c) {
     $v->optional('groupid')->num(0, undef);
     $v->optional('modules', 'comma_separated');
     $v->optional('limit', 'not_empty')->num(0, undef);
+    $v->optional('group_glob');
+    $v->optional('not_group_glob');
```
okurz, line 401: I would leave out the "_glob" suffix.

okurz (changes requested):

lib/OpenQA/WebAPI/Plugin/Helpers.pm
kraih: That would also have been my preference originally too, but we've already set a precedent here with `module_re` for regex. Feels kinda wrong now not to have a special suffix for globs too.

kraih: Perl regexes are very powerful and compiling arbitrary strings from the internet into them has a lot of room for exploitation. For example i could probably use lookahead/behind to create a DoS attack even with small input strings. 🤔 Using globs would be a lot safer and easier to understand by users.

kraih:
> I wonder about the example that you mention about /tests/overview. https://progress.opensuse.org/issues/119746 is about /tests. Does this also work for /tests or was that aspect overlooked?

Yes, that aspect was indeed overlooked after the discussions about the feature. Do we really want the group filtering in the `/tests` context rather than `/tests/overview`? I didn't even think about `/tests` because it fits so well into `/tests/overview`. Unfortunately the code paths are completely separate and probably won't be able to share much.

lib/OpenQA/WebAPI/Plugin/Helpers.pm
Martchus: Yes, unfortunately Perl has no good way of dealing with regexes from users. Unfortunately we already do this. I once wrote the `regex_match` helper so we at least don't get warnings in our server logs but a DoS attach is likely possible. So I guess using globbing is better here, indeed.

(We can actually always make use of native code and e.g. call pcre2. However, that's more work.)

kraih:
> Yes, that aspect was indeed overlooked after the discussions about the feature. Do we really want the group filtering in the `/tests` context rather than `/tests/overview`?

As discussed on Slack, there is interest in having both. So i will look into extending the spike solution.

lib/OpenQA/WebAPI/Plugin/Helpers.pm
kraih: For more information see also https://owasp.org/www-community/attacks/Regular_expression_Denial_of_Service_-_ReDoS.

kraih:
Followup PR has been merged.

PR #5292, opened 2023-09-01
Allow settings to wrap into the next line

kalikiana:
Settings become unreadable or even hidden when they flow outside of the containing element or window.

Fall back to wrapping the text if nothing else.

See: https://progress.opensuse.org/issues/134897


kalikiana:
Screenshot 2023-09-01 at 11 03 28

Martchus (approved):

PR #5293, opened 2023-09-04
Remove one UPDATE query from worker status updates

kraih:
This combines the t_seen and error updates into one. Also removed some dead code in the seen method.

Progress: https://progress.opensuse.org/issues/135122

### DIFF
--- a/lib/OpenQA/Schema/Result/Workers.pm
+++ b/lib/OpenQA/Schema/Result/Workers.pm
@@ -3,9 +3,7 @@
 # SPDX-License-Identifier: GPL-2.0-or-later
 
 package OpenQA::Schema::Result::Workers;
-
-
-use Mojo::Base 'DBIx::Class::Core';
+use Mojo::Base 'DBIx::Class::Core', -signatures;
 
 use DBIx::Class::Timestamps 'now';
 use Try::Tiny;
@@ -70,10 +68,10 @@ sub name {
     return $self->host . ":" . $self->instance;
 }
 
-sub seen {
-    my ($self, $workercaps, $error) = @_;
-    $self->update({t_seen => now()});
-    $self->update_caps($workercaps) if $workercaps;
+sub seen ($self, $options = {}) {
+    my $data = {t_seen => now()};
+    $data->{error} = $options->{error} if exists $options->{error};
+    $self->update($data);
 }
 
 # update worker's capabilities
--- a/lib/OpenQA/WebSockets/Controller/Worker.pm
+++ b/lib/OpenQA/WebSockets/Controller/Worker.pm
@@ -156,8 +156,7 @@ sub _message {
                 sub {
                     return undef unless my $w = $schema->resultset("Workers")->find($worker_id);
                     log_debug("Updating seen of worker $worker_id from worker_status ($current_worker_status)");
-                    $w->seen;
-                    $w->update({error => $current_worker_error});
+                    $w->seen({error => $current_worker_error});
                 });
         }
         catch {


perlpunk (approved):

lib/OpenQA/Schema/Result/Workers.pm
```diff
@@ -71,9 +71,10 @@ sub name {
 }
 
 sub seen {
-    my ($self, $workercaps, $error) = @_;
-    $self->update({t_seen => now()});
-    $self->update_caps($workercaps) if $workercaps;
+    my ($self, $options) = (shift, shift // {});
```
perlpunk, line 74: Maybe we can switch to signatures here?

kraih: I don't see why not.

lib/OpenQA/Schema/Result/Workers.pm
```diff
@@ -70,10 +68,10 @@ sub name {
...
 }
 
-sub seen {
-    my ($self, $workercaps, $error) = @_;
-    $self->update({t_seen => now()});
-    $self->update_caps($workercaps) if $workercaps;
+sub seen ($self, $options = {}) {
+    my $data = {t_seen => now()};
+    $data->{error} = $options->{error} if exists $options->{error};
+    $self->update($data);
```
Martchus, line 74: Supposedly it would be the easiest to just let the caller specify any data that should be updated and only set `t_seen` in addition:

```suggestion
sub seen ($self, $data = {}) {
    $data->{t_seen} = now();
    $self->update($data);
```

This way this function becomes never bigger, even if we want to set more additional data in the future.

Martchus: Do we need an additional abstraction here at all? I personally would try to keep it simple.

perlpunk (approved):

lib/OpenQA/Schema/Result/Workers.pm
kraih: Yes, `$worker->seen` is called in 5 places, copypasting `$worker->update({t_seen => now()})` for each one would be silly.

kraih: But seriously, we've got performance issues to resolve, this is not the time to nitpick on tiny details like this. The only question you should be asking is if this is better than before.

kalikiana (approved):

mimi1vx (approved):

perlpunk:
I think OBS checks were just broken because of the outage: https://build.opensuse.org/package/show/devel:openQA:GitHub:os-autoinst:openQA:PR-5293/openQA
Can't find any unsuccessful webhook deliveries though.
I will merge manually.

lib/OpenQA/Schema/Result/Workers.pm
Martchus: > Do we need an additional abstraction here at all?

With that question I didn't want to ask whether we should drop that function. You said:

> I don't think that would be good abstraction.

And I personally don't see `seen` as an additional layer of abstraction besides avoiding to repeat the code it contains. So I'm only questioning whether the statement "I don't think that would be good abstraction." is relevant.

> But seriously, we've got performance issues to resolve, this is not the time to nitpick on tiny details like this.

That's why I have never requested changes.

kraih: I see, then we had a bit of miscommunication there. My comment about abstraction was a bit too vague indeed.

What i was referring to was specifically replacing a generic options hash `$options = {}` with a `$data = {}` hash tied to the fields of the table. Because it would severely limit how we can expand the method in the future. Instead of being able to add any arbitrary option to the method without changing its signature, we would limit the ways in which we can expand it for no real gain at all.

PR #5294, opened 2023-09-05
Remove redundant worker SELECT query

kraih:
This patch removes the redundant `SELECT` on the workers table, as well as the transaction, which didn't do anything of significance (no `FOR UPDATE`).

Screenshot 2023-09-07 at 14 11 23

Screenshot 2023-09-07 at 13 50 39

### DIFF
--- a/lib/OpenQA/WebSockets/Controller/Worker.pm
+++ b/lib/OpenQA/WebSockets/Controller/Worker.pm
@@ -59,18 +59,18 @@ sub _message {
     my $schema = $app->schema;
 
     # find relevant worker
-    my $worker = OpenQA::WebSockets::Model::Status->singleton->worker_by_transaction->{$self->tx};
-    unless ($worker) {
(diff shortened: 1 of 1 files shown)


lib/OpenQA/WebSockets/Controller/Worker.pm
```diff
@@ -150,22 +150,23 @@ sub _message {
...
+            };
         }
-        catch {
-            log_error("Failed updating seen and error status of worker $worker_id: $_");    # uncoverable statement
-        };
 
         # send worker population
         try {
-            my $workers_population = $schema->resultset("Workers")->count();
+            my $workers_population = $workers->count();
```
Martchus, line 169: And if that whole worker population stuff gets to expensive I suggest to remove it completely (including the receiving side).

kraih: That's exactly the plan for patch 3. 😁

perlpunk (approved):

baierjan (approved):

PR #5296, opened 2023-09-07
Redirect windows image requests

perlpunk:
Issue: https://progress.opensuse.org/issues/135251

### DIFF
--- a/etc/nginx/vhosts.d/openqa-locations.inc
+++ b/etc/nginx/vhosts.d/openqa-locations.inc
@@ -56,4 +56,5 @@ location / {
     proxy_set_header X-Forwarded-Server $host;
     proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
     proxy_set_header X-Forwarded-Proto $scheme;
+    rewrite ^/tests/.*/asset/.*/[wW]in.*$ https://www.microsoft.com/software-download/windows10ISO last;
 }


etc/nginx/vhosts.d/openqa-locations.inc
```diff
@@ -56,4 +56,5 @@ location / {
     proxy_set_header X-Forwarded-Server $host;
     proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
     proxy_set_header X-Forwarded-Proto $scheme;
+    rewrite ^/tests/.*/asset/.*/[wW]in.*$ https://www.microsoft.com/software-download/windows10ISO last;
```
baierjan, line 59: Wasn't that asset**s** in the original rule?

baierjan, line 59:
```suggestion
    rewrite ^(/tests/.*)?/assets?/.*/[wW]in.*$ https://www.microsoft.com/software-download/ last;
```

perlpunk: See my comment in the ticket. The `/assets` url was commented out, so I didn't port that one.

baierjan: Makes sense, I was just looking at the original ticket description.

PR #5297, opened 2023-09-08
Remove worker population broadcasts

kraih:
This patch gets rid of the expensive `SELECT COUNT(*) FROM workers` and replaces the complex status update interval calculation with something simpler.

### DIFF
--- a/lib/OpenQA/Worker/CommandHandler.pm
+++ b/lib/OpenQA/Worker/CommandHandler.pm
@@ -109,18 +109,9 @@ sub handle_command {
             return $worker->skip_job($job_id, $type);
         }
     }
-    log_warning("Ignoring WS message with unknown type $type from $webui_host:\n" . pp($json));
-}
 
-sub _handle_command_info {
-    my ($json, $client, $worker, $webui_host) = @_;
-
-    if (my $population = $json->{population}) {
-        $client->webui_host_population($population);
-    }
-
-    # enfore recalculation of interval for overall worker status update
-    $client->send_status_interval(undef);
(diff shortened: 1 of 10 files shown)


kraih:
I've also been pondering increasing the minimum interval from `20` to `60` seconds and the maximum from `100` seconds to `5` minutes. That should still strike a good enough balance between responsiveness and not exceeding the 30 minute worker timeout.

lib/OpenQA/Worker/CommandHandler.pm
```diff
@@ -109,18 +109,9 @@ sub handle_command {
...
-    my ($json, $client, $worker, $webui_host) = @_;
-
-    if (my $population = $json->{population}) {
-        $client->webui_host_population($population);
-    }
 
-    # enfore recalculation of interval for overall worker status update
-    $client->send_status_interval(undef);
+    # WS messages of type "info" can be sent for debugging
+    log_warning("Ignoring WS message with unknown type $type from $webui_host:\n" . pp($json)) if $type ne 'info';
```
perlpunk, line 114: So currently with your change above we would see this warning for every status update? Maybe make it a `log_debug`?

kraih: That's why i appended the `if $type ne 'info'`. We will not be seeing those in production. 😉

perlpunk: oooh. I reversed the logic in my head...

baierjan (approved):

PR #5298, opened 2023-09-11
Makefile: Make systemd config file replacements fatal on error

okurz:
To prevent any necessary replacement not happening unnoticed we should
exit with explicit exit codes on failed attempts.

Related progress issue: https://progress.opensuse.org/issues/133352


baierjan (changes requested):
This seems to work only for the `s` command:
> t label: If a s/// has done a successful substitution...

baierjan:
I propose an alternative solution in #5300

PR #5299, opened 2023-09-11
Optimize get_property method

kraih:
This one surprised me a little bit, since i did not expect the DBIx::Class overhead to be so significant. At first i wanted to add an index on the `key` field of the `worker_properties` table, but that made no measurable difference at all. And i could confirm with `EXPLAIN ANALYZE` on the production OSD database that the `SELECT` query used here does not benefit from an additional index.

In the end it was all DBIx::Class overhead that made the difference getting 1000 `->get_property('JOB_TOKEN')` calls from `1.34s` down to `0.269s`.

In the greather scheme of things this might not seem like much, but for the hottest code path in the websocket server it can make a big difference.

### DIFF
--- a/lib/OpenQA/Schema/Result/Workers.pm
+++ b/lib/OpenQA/Schema/Result/Workers.pm
@@ -13,6 +13,7 @@ use OpenQA::WebSockets::Client;
 use OpenQA::Constants qw(WORKER_API_COMMANDS DB_TIMESTAMP_ACCURACY);
 use OpenQA::Jobs::Constants;
 use Mojo::JSON qw(encode_json decode_json);
+use DBI qw(:sql_types);
 
 __PACKAGE__->table('workers');
 __PACKAGE__->load_components(qw(InflateColumn::DateTime Timestamps));
@@ -87,8 +88,15 @@ sub update_caps {
 sub get_property {
     my ($self, $key) = @_;
 
-    my $r = $self->properties->find({key => $key});
-    return $r ? $r->value : undef;
+    # Optimized because this is a performance hot spot for the websocket server
+    my $sth = $self->result_source->schema->storage->dbh->prepare(
+        'SELECT value FROM worker_properties WHERE key = ? AND worker_id = ? LIMIT 1');
+    $sth->bind_param(1, $key, SQL_CHAR);
+    $sth->bind_param(2, $self->id, SQL_BIGINT);
+    $sth->execute;
+    my $r = $sth->fetchrow_arrayref;
+
+    return $r ? $r->[0] : undef;
 }
 
 sub delete_properties {


kraih:
For context, this is what the original DBIx::Class query looked like:
```
openqa=# EXPLAIN ANALYZE SELECT me.id, me.key, me.value, me.worker_id, me.t_created, me.t_updated FROM worker_properties me WHERE ( ( me.key = 'JOBTOKEN' AND me.worker_id = 2311 ) );
                                                                QUERY PLAN
------------------------------------------------------------------------------------------------------------------------------------------
 Bitmap Heap Scan on worker_properties me  (cost=4.38..45.86 rows=1 width=130) (actual time=0.100..0.147 rows=1 loops=1)
   Recheck Cond: (worker_id = 2311)
   Filter: (key = 'JOBTOKEN'::text)
   Rows Removed by Filter: 11
   Heap Blocks: exact=5
   ->  Bitmap Index Scan on worker_properties_idx_worker_id  (cost=0.00..4.38 rows=12 width=0) (actual time=0.057..0.057 rows=12 loops=1)
         Index Cond: (worker_id = 2311)
 Planning Time: 0.122 ms
 Execution Time: 0.176 ms
(9 rows)
```

baierjan (approved):

Martchus (approved):
Nice that you've found more room for performance improvements!

perlpunk (changes requested):
Can't this be done with DBIx::Class using the `select` option to only fetch one column?
I would like to avoid falling back to DBI for several reasons.

For the record, this is the currently execured SQL:
```
SELECT me.id, me.key, me.value, me.worker_id, me.t_created, me.t_updated FROM worker_properties me WHERE ( ( me.key = ? AND me.worker_id = ? ) ): 'JOBTOKEN', '604'
```

kraih:
> Can't this be done with DBIx::Class using the `select` option to only fetch one column?

That's not what this patch is about i'm afraid, it makes no measurable difference.

lib/OpenQA/Schema/Result/Workers.pm
```diff
@@ -87,8 +88,15 @@ sub update_caps {
 sub get_property {
     my ($self, $key) = @_;
 
-    my $r = $self->properties->find({key => $key});
-    return $r ? $r->value : undef;
+    # Optimized because this is a performance hot spot for the websocket server
+    my $sth = $self->result_source->schema->storage->dbh->prepare(
```
perlpunk, line 92: Can you try out if the following optimization would be enough?
```
my $r = $self->properties->find({key => $key}, {select => [qw(value)]});
```

kraih: One of the first things i tried, makes literally no difference at all, still `1.34s`.

kraih: Before you ask, I also tried `->search()` with explicit `LIMIT 1`. Also no difference, it's all in the DBIx::Class overhead.

kraih: The `LIMIT 1` doesn't do much work here btw., i only have it in the raw SQL because it's the correct query to use. :grin:

kraih: Here's all the numbers, since you asked for that on Slack:

kraih: So it spends about as much time on the `->proterties` call, as the whole raw SQL version takes performing queries.

perlpunk: Ok.
with DBIx::Class overhead I also associate overhead for creating DateTime objects for columns that are often never used.
I just made a small comparison with Time::Hires, and I also don't really see a big difference.
Maybe DBIx::Class can now lazy-load columns like this?

It wasn't clear to me that you had checked this, so I had to ask.

perlpunk (approved):

lib/OpenQA/Schema/Result/Workers.pm
kraih: Never underestimate the cost of plain old object instantiation in Perl when it comes to performance hot spots.

PR #5301, opened 2023-09-11
Make searching for unfinished jobs faster (Option A)

kraih:
This is option **A**. See #5302 for option B.
 
The `$worker->unfinished_jobs` in the websocket server is the last low hanging fruit shown by my profiling data. And it is an interesting one because it has the potential to be a huge improvement, or to be a bit underwhelming. It is entirely dependent on how many previous jobs the worker sending the status update has in the database.

Currently the SQL query generated by DBIx::Class ends up uing an index for the `assigned_worker_id` field, but not for `t_finished`. A combined index for both fields has shown the best imporvements.

For the benchmark i've used a real world example from OSD with 1420 previous jobs. Due to DBIx::Class overhead there is a performance floor around 1.6s, at that point no further optimizations at the PostgreSQL layer are possible.

1000 status updates **before**:

1000 status updates **after**:

Progress: https://progress.opensuse.org/issues/135362

PR #5302, opened 2023-09-11
Make searching for unfinished jobs faster (Option B)

kraih:
This is option **B**. See #5301 for option A.

The `$worker->unfinished_jobs` in the websocket server is the last low hanging fruit shown by my profiling data. And it is an interesting one because it has the potential to be a huge improvement, or to be a bit underwhelming. It is entirely dependent on how many previous jobs the worker sending the status update has in the database.

Currently the SQL query generated by DBIx::Class ends up using an index for the `assigned_worker_id` field, but not for `t_finished`. A combined index for both fields has shown the best imporvements. But Tina brought up the idea of using the `state` instead of `t_finished`, which is almost as good.

For the benchmark i've used a real world example from OSD with 1420 previous jobs. Due to DBIx::Class overhead there is a performance floor around 1.6s, at that point no further optimizations at the PostgreSQL layer are possible. This patch does not reach the performance floor, but it has the advantage that no new index is required.

### DIFF
--- a/lib/OpenQA/Schema/Result/Workers.pm
+++ b/lib/OpenQA/Schema/Result/Workers.pm
@@ -248,7 +248,7 @@ sub send_command {
 sub unfinished_jobs {
     my ($self) = @_;
 
-    return $self->previous_jobs->search({t_finished => undef});
+    return $self->previous_jobs->search({state => {-in => [OpenQA::Jobs::Constants::PENDING_STATES]}});
 }
 
 sub set_current_job {


Martchus (approved):
Option B is very simple so I guess there's no reason not to approve it (even though we would later decide to use option A after all).

By the way, I was thinking of reasons why I used `t_finished` here overr `state` when writing the code. I couldn't think of any. So this *should* be functionally equivalent.

kraih:
> Option B is very simple so I guess there's no reason not to approve it (even though we would later decide to use option A after all).

One interesting observation i should maybe mention here, a compound index on `(assigned_worker_id, state)` does not appear to improve performance for the `state` solution. So a switch to solution A would also have to go back to using `t_finished` to get the best plan.

kraih:
Closed option A so this one can get reviewed.

okurz (approved):

PR #5303, opened 2023-09-12
Log an info message when worker status updates are too frequent

kraih:
This will give us a hint at when the websocket server is falling behind with handling worker status updates. It should usually happen when multiple updates end up in the websocket buffer and are handled right after another.

In the future we could use this condition to ignore the extra status updates. Which would give the websocket server a much better chance to recover on its own, if this was only a temporary performance issue.

### DIFF
--- a/lib/OpenQA/WebSockets/Controller/Worker.pm
+++ b/lib/OpenQA/WebSockets/Controller/Worker.pm
@@ -7,7 +7,7 @@ use Mojo::Base 'Mojolicious::Controller';
 use DBIx::Class::Timestamps 'now';
 use OpenQA::Schema;
 use OpenQA::Log qw(log_debug log_error log_info log_warning log_trace);
-use OpenQA::Constants qw(WEBSOCKET_API_VERSION);
+use OpenQA::Constants qw(WEBSOCKET_API_VERSION MIN_TIMER);
 use OpenQA::WebSockets::Model::Status;
 use OpenQA::Jobs::Constants;
 use OpenQA::Scheduler::Client;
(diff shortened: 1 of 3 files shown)


lib/OpenQA/WebSockets/Controller/Worker.pm
```diff
@@ -151,6 +152,14 @@ sub _message {
         log_trace "Received from worker $worker_id worker_status message: " . dumper($json)
           if LOG_WORKER_STATUS_MESSAGES;
 
+        # detect if the websocket server is running too close to its limit and falling behind with status updates
+        my ($last_seen, $now) = ($worker_status->{last_seen}, time);
```
Martchus, line 156: I hope getting the time here has only a neglectable performance impact. This just reminds me of 3c42ce17c2b70957215d511166410fbd15a1b04f where it contributed notably to the runtime (but I guess it was more the formatting than getting the time itself).

lib/OpenQA/WebSockets/Controller/Worker.pm
```diff
@@ -47,6 +47,7 @@ sub _finish {
     }
     $reason = ($reason ? ": $reason" : '');
     log_info("Worker $worker->{id} websocket connection closed - $code$reason");
+    delete $worker->{last_seen};
```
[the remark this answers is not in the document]
kraih: Yes, actually is required and i have a test for it.

lib/OpenQA/WebSockets/Controller/Worker.pm
kraih: No measurable impact fortunately.

kraih:
For the record, there is no measurable negative impact on performance by this change, still `3.70s`, same as #5302.

perlpunk (approved):

Martchus (approved):

PR #5305, opened 2023-09-13
Adapt to deprecation of spurt in upstream Mojolicious

okurz:
Related progress issue: https://progress.opensuse.org/issues/135632

PR #5309, opened 2023-09-20
Fix key navigation on test details

Martchus:
When viewing a test live and test module details are added to the page dynamically, then the key navigation breaks. It skips steps instead of just selecting the next/previous step. This is because the event handler is registered multiple times. This change fixes the issue by making sure the event handler is only registered once.

### DIFF
--- a/assets/javascripts/test_result.js
+++ b/assets/javascripts/test_result.js
@@ -644,16 +644,20 @@ function renderTestModules(response) {
     setCurrentPreviewFromStepLinkIfPossible($("[href='" + hash + "'], [data-href='" + hash + "']"));
   }
 
-  // setup keyboard navigation through test details
-  $(window).keydown(handleKeyDownOnTestDetails);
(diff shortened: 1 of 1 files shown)


Martchus:
By the way, I *guess* we have a ticket for that. However, I couldn't find it. (I came across this problem again when working on https://progress.opensuse.org/issues/136028.)

kalikiana:
This is [poo#113468](https://progress.opensuse.org/issues/113468) I think?

kalikiana (approved):
Seems like a sensible fix!

perlpunk (approved):

okurz (approved):

kalikiana:
@mergifyio rebase

okurz:
@kalikiana as long as
* https://github.com/os-autoinst/openQA/pull/5305

is not in rebasing won't help

perlpunk:
@Mergifyio rebase

PR #5310, opened 2023-09-21
Update dependencies

perlpunk:
This is a new PR with the changes of #5305
The old one is still using the package cache in circleci.


okurz:
wow, interesting. Some more failures. I think the version check in tools/tidy is broken, on it

perlpunk:
Yep. fixed the Mojo::File warning, now looking into Perl::Tidy complaints

okurz (approved):
see https://github.com/os-autoinst/openQA/pull/5311 for a fix of the tidy version check in tools/tidy assuming that we actually need to specify the version as 20230909.0.0.

But if all checks in your PR pass then we should merge that

PR #5313, opened 2023-09-26
t: Fix api/13-influxdb.t timezone issue

perlpunk:
We also need to add the timezone offset in the test to tell postgres that the date is meant to be UTC.
Otherwise the test can fail depending on the local timezone.

### DIFF
--- a/lib/OpenQA/WebAPI/Controller/Admin/Influxdb.pm
+++ b/lib/OpenQA/WebAPI/Controller/Admin/Influxdb.pm
@@ -107,8 +107,8 @@ sub minion ($self) {
     # rc means hook script return code
     my $sth = $dbh->prepare(
         q{SELECT COUNT(*) AS rc_failed_count FROM minion_jobs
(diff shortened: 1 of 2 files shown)


okurz (approved):

lib/OpenQA/WebAPI/Controller/Admin/Influxdb.pm
```diff
@@ -107,8 +107,8 @@ sub minion ($self) {
     # rc means hook script return code
     my $sth = $dbh->prepare(
         q{SELECT COUNT(*) AS rc_failed_count FROM minion_jobs
-		  WHERE finished >= ? AND finished < ? AND task = 'hook_script' AND
-		        state = 'finished' AND (notes->'hook_rc')::int != 0}
+                 WHERE finished >= ? AND finished < ? AND task = 'hook_script' AND
+                       state = 'finished' AND (notes->'hook_rc')::int != 0}
```
Martchus, line 111: And tidy didn't complain about this? But it is good that you fixed it of course.

Martchus (approved):

PR #5314, opened 2023-09-26
Require Mojolicious 9.340.0

perlpunk:
Perl modules are getting nomalized version formats

See also https://github.com/openSUSE/cpanspec/issues/47 about the new versions


dependencies.yaml
```diff
@@ -38,7 +38,9 @@ common_requires:
   perl(Filesys::Df):
   perl(Getopt::Long):
   perl(Minion): '>= 10.25'
-  perl(Mojolicious): '>= 9.34'
+  perl(Mojolicious):
+    perl: '>= 9.34'
+    rpm: '>= 9.340.0'
```
kalikiana, line 43: Hrm. This results in `perl(Mojolicious) >= 9.340.` in the spec file. I don't get it.

kalikiana: Ftw I had a typo. We disussed it briefly in Jitsi. So as I understand it the double versions can't be avoided at least for now while rpm packages can have old and new style versions.

kalikiana (approved):

Martchus (approved):

PR #5317, opened 2023-09-28
Avoid displaying buttons on sticky table headers

Martchus:
Raise the z-index of sticky table headers slightly above the default so elements like buttons are hidden under them when scrolling.


okurz (approved):

kalikiana (approved):
Nice!

PR #5311, opened 2023-09-21
Adapt tools/tidy to dotted version

okurz:
* Update code according to new Perl::Tidy
* Fix usage of spew with empty content
* Update tidy to 20230909
* Adapt to deprecation of spurt in upstream Mojolicious
* Dependency cron 2023-09-13
* git subrepo pull (merge) external/os-autoinst-common

PR #5318, opened 2023-10-01
Fix dependency update script

perlpunk:
Perl::Tidy version has two dots now, e.g. 20230909.0.0

The original grep for `[0-9]*` resulted in three lines, so that sed complained:

    sed: -e expression #1, char 55: unterminated `s' command

See https://app.circleci.com/pipelines/github/os-autoinst/openQA/12241/workflows/6a3918da-3db5-402a-9bc3-b616c7d5715d/jobs/114292

### DIFF
--- a/cpanfile
+++ b/cpanfile
@@ -115,7 +115,7 @@ on 'test' => sub {
 };
 
 on 'devel' => sub {
-    requires 'Perl::Tidy', '== 20230909';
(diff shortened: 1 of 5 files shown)


perlpunk:
I added the commit from #5311
I would like to merge this later today to avoid future failures and possibly new dependencies that would make it necessary to amend this PR again.
No objection means approval =)

okurz (approved):

PR #5323, opened 2023-10-06
Apply memory limit for openqa-worker services

baierjan:
All workers are already part of openqa-worker.slice, apply memory limit to that slice to ensure that memory cannot be exhausted exclusively by workers.

Reference: https://progress.opensuse.org/issues/133511

### DIFF
--- a/dist/rpm/openQA.spec
+++ b/dist/rpm/openQA.spec
@@ -693,6 +693,7 @@ fi
 %dir %{_unitdir}
 %{_systemdgeneratordir}
 %{_unitdir}/openqa-worker.target
+%{_unitdir}/openqa-worker.slice
 %{_unitdir}/openqa-worker@.service
 %{_unitdir}/openqa-worker-plain@.service
 %{_unitdir}/openqa-worker-cacheservice-minion.service
--- a/systemd/openqa-worker.slice
+++ b/systemd/openqa-worker.slice
@@ -1,0 +1,7 @@
+[Unit]
+Description=Slice for openqa-worker units
+
+[Slice]
+# Apply memory limit to this slice to ensure that memory cannot be exhausted
+# exclusively by openqa-worker services.
+MemoryMax=90%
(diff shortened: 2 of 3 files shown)


dist/rpm/openQA.spec
```diff
@@ -693,6 +693,7 @@ fi
 %dir %{_unitdir}
 %{_systemdgeneratordir}
 %{_unitdir}/openqa-worker.target
+%{_unitdir}/openqa-worker.slice
```
Martchus, line 696: I see that we have `Slice=openqa-worker.slice` in the `[Service]` section of our unit file which is likely also relevant for this new file to be picked up. I'm wondering where `Slice=` is actually documented. I browsed a lot in the different pages of the systemd documentation but couldn't find it. There's not even a mentioning in which section this setting is supposed to go.

Martchus (commented):
I've read the systemd documentation and I suppose this will work - although I'm not quite sure on the `Slice=` setting (see inline comment).

systemd/openqa-worker.slice
```diff
@@ -0,0 +1,5 @@
+[Unit]
+Description=Slice for openqa-worker units
+
+[Slice]
+MemoryMax=90%
```
okurz, line 5: Please put your commit message description in here as the explanation would serve well for readers of the service file

okurz (changes requested):

Martchus:
The ticket says:

> with some fancy cgroup settings in the openQA worker systemd units depending on available memory **and a fair division for all instances**.

I'm not sure whether that's achieved by this approach. Wouldn't it put all worker instances in the same cgroup? That would preserve some memory for the rest of the system but not achieve a "fair division for all instances". For this I suppose every instance of the template unit needed to have its own slice.

baierjan:
> For this I suppose every instance of the template unit needed to have its own slice.

Actually, it should be just enough to put `MemoryMax=` into our `openqa-worker@.service` template (or as a drop-in), no need to create a separate slice for it.

systemd/openqa-worker.slice
baierjan: I added the comment, is this what you meant?

dist/rpm/openQA.spec
Martchus: Thanks, I must have overlooked it.

okurz (approved):

Martchus (approved):

PR #5325, opened 2023-10-07
Provide convenience CLI shortcuts

okurz:
* dist: Provide convenience CLI shortcuts in package as well
* Add short convenience CLI commands

### DIFF
--- a/dist/rpm/openQA.spec
+++ b/dist/rpm/openQA.spec
@@ -388,6 +388,10 @@ ln -s %{_datadir}/openqa/script/openqa-clone-job %{buildroot}%{_bindir}/openqa-c
 ln -s %{_datadir}/openqa/script/openqa-dump-templates %{buildroot}%{_bindir}/openqa-dump-templates
 ln -s %{_datadir}/openqa/script/openqa-load-templates %{buildroot}%{_bindir}/openqa-load-templates
 ln -s %{_datadir}/openqa/script/openqa-clone-custom-git-refspec %{buildroot}%{_bindir}/openqa-clone-custom-git-refspec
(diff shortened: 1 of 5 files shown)


dist/rpm/openQA.spec
```diff
@@ -388,6 +388,10 @@ ln -s %{_datadir}/openqa/script/openqa-clone-job %{buildroot}%{_bindir}/openqa-c
 ln -s %{_datadir}/openqa/script/openqa-dump-templates %{buildroot}%{_bindir}/openqa-dump-templates
 ln -s %{_datadir}/openqa/script/openqa-load-templates %{buildroot}%{_bindir}/openqa-load-templates
 ln -s %{_datadir}/openqa/script/openqa-clone-custom-git-refspec %{buildroot}%{_bindir}/openqa-clone-custom-git-refspec
+ln -s %{_datadir}/openqa/script/occ %{buildroot}%{_bindir}/occ
+ln -s %{_datadir}/openqa/script/ocj %{buildroot}%{_bindir}/ocj
+ln -s %{_datadir}/openqa/script/ocl %{buildroot}%{_bindir}/ocl
+ln -s %{_datadir}/openqa/script/ola %{buildroot}%{_bindir}/ola
```
kalikiana, line 394: I wonder if these aren't more suited to aliases or at least links setup by users - no super strong opinion but if these also end up in shared scripts or config files it might make it more difficult to follow the code or check where commands are used

foursixnine (changes requested):
I wonder if this is a convenient shorcut for you, or for users who asked for this? however, for the clone-custom-git-refspec, it should have a better name.

okurz:
I only heard the feedback that some commands are long to type but the specific name suggestions are by me. There had been no positive feedback just convincing drawbacks mentioned so closing.

PR #5326, opened 2023-10-09
ci: Use PERL_TEST_WARNINGS_ONLY_REPORT_WARNINGS

perlpunk:
Set to 0 so by default we get fatal warnings.

This makes it easy to quickly set the variable in the workflow settings when necessary.

Issue: https://progress.opensuse.org/issues/137105

### DIFF
--- a/tools/ci/run_unit_tests.sh
+++ b/tools/ci/run_unit_tests.sh
@@ -1,6 +1,8 @@
 #!/bin/bash
 set -ex
 target="${1:?"Specify a makefile target"}"
+# Set to 1 to temporarily ignore warnings
+export PERL_TEST_WARNINGS_ONLY_REPORT_WARNINGS=0
 export COVERAGE=1
 export COVERDB_SUFFIX="_${target}"
 # Enable JUnit-compatible test results


tools/ci/run_unit_tests.sh
```diff
@@ -1,6 +1,8 @@
 #!/bin/bash
 set -ex
 target="${1:?"Specify a makefile target"}"
+# Set to 1 to temporarily ignore warnings
+export PERL_TEST_WARNINGS_ONLY_REPORT_WARNINGS=0
```
kalikiana, line 5: Did you test it? I can't seem to execute it [on your branch](https://github.com/perlpunk/openQA/actions/workflows/prove.yml?query=branch%3Atest-warnings). There's no **Run workflow** button 🤔

okurz: I think that button only appears if the workflow is configured for the manual dispatch. I don't worry about this change though and approved it nevertheless.

perlpunk: I forgot to add the link where I tested it with a warning: https://app.circleci.com/pipelines/github/perlpunk/openQA/306/workflows/da68462a-3df4-404d-9196-89b25e8c8db4/jobs/1851

kalikiana: Yes, `workflow_dispatch` like in `prove.yml`. You're right, though, the intention is to use it with CircleCI specifically

kalikiana (approved):

tools/ci/run_unit_tests.sh
perlpunk: Regarding the button - I think there is no button visible for you because you can't trigger a workflow run in my repo, where you don't have any access. I can see the button.

PR #5331, opened 2023-10-16
Add repeat parameter to clone a job multiple times

ilmanzo:
Reference ticket: https://progress.opensuse.org/issues/122953 

to better support statistical investigation, we want to provide an
immediate way to clone a job multiple times. Till now it has been done
with a for loop similar to

```
for i in {01..50}; do openqa-clone-job … TEST+=-$i; done
```

But we can avoid the API overhead of copying the same parameters and
assets again and again from the original job, simply by cloning it once and
posting the new job(s) N times. The TEST name is automatically numbered, so the
user is not forced to remember any shell syntax.

As the default is `--repeat=1`, the API remains compatible with any shell
script using the `openqa-clone-job` program.

### DIFF
--- a/docs/UsersGuide.asciidoc
+++ b/docs/UsersGuide.asciidoc
@@ -1113,6 +1113,25 @@ timeouts by performing the task in background.
 This is recommended on big instances but means that the results (and
 possible errors) need to be polled via `openqa-cli api isos/$scheduled_product_id`.
 
+
+===== Statistical investigation ===== 
+In case issues appear sporadically and are therefore hard to reproduce it can 
+help to trigger many more jobs on a production instance to gather more data 
+first, for example the failure ratio.
+
+Example of triggering 50 jobs in a development group so that the result of 
+passed/failed jobs is counted by openQA itself on the corresponding overview page:
+[source,sh]
+--------------------------------------------------------------------------------
+openqa-clone-job --skip-chained-deps --repeat=50 --within-instance \
+https://openqa.opensuse.org 123456  BUILD=poo32242_investigation \
+_GROUP="Test Development:openSUSE Tumbleweed" 
+--------------------------------------------------------------------------------
+
+To get an overview about the fail ratio and confidence interval of sporadically 
+failing applications you can also use a script like 
+https://github.com/okurz/scripts/blob/master/count_fail_ratio[this].
+
 [id="scenarios_yaml"]
 ===== Defining test scenarios in YAML =====
 Instead of relying on the tables for machines, mediums/products, test suites and
--- a/lib/OpenQA/Script/CloneJob.pm
+++ b/lib/OpenQA/Script/CloneJob.pm
@@ -226,11 +226,31 @@ sub handle_tx ($tx, $url_handler, $options, $jobs) {
     }
 }
 
+# append a formatted "-NN" to the TEST parameter for each job being posted
+sub append_idx_to_test_name ($n, $digits, $post_params) {
+    my $suffix = sprintf("-%0${digits}d", $n);
+    foreach my $job_key (keys %$post_params) {
+        my $job = $post_params->{$job_key};
+        if ($n == 1) {
+            $job->{TEST} .= $suffix;    # just append at the first time
+        }
+        else {
+            # from the second onwards, replace old with the new number
+            $job->{TEST} =~ s/-\d+$/$suffix/;
+        }
+    }
+}
+
 sub clone_jobs ($jobid, $options) {
     my $url_handler = create_url_handler($options);
+    my $repeat = delete $options->{repeat} || 1;
+    my $digits = length $repeat;
     clone_job($jobid, $url_handler, $options, my $post_params = {}, my $jobs = {});
-    my $tx = post_jobs($post_params, $url_handler, $options);
-    handle_tx($tx, $url_handler, $options, $jobs) if $tx;
+    for my $counter (1 .. $repeat) {
+        append_idx_to_test_name($counter, $digits, $post_params) if $repeat > 1;
+        my $tx = post_jobs($post_params, $url_handler, $options);
+        handle_tx($tx, $url_handler, $options, $jobs) if $tx;
+    }
 }
 
 sub clone_job ($jobid, $url_handler, $options, $post_params = {}, $jobs = {}, $depth = 1, $relation = '') {
--- a/script/openqa-clone-job
+++ b/script/openqa-clone-job
@@ -133,6 +133,17 @@ are cloned. Use C<0> to denote infinity.
 A shortcut for C<--skip-download --from HOST --host HOST> to clone a job within
 a local or remote instance.
 
+=item B<--repeat> N
+
+Do the same clone operation N times; this can be useful for Statistical 
+investigation. Example: 
+
+openqa-clone-job --skip-chained-deps --repeat=50 --within-instance \
+https://openqa.opensuse.org 123456 
+
+Will create 50 clones of the same job.
+
+
 =item B<--show-progress>
 
 Displays a progress bar when downloading assets.
@@ -175,6 +186,7 @@ use FindBin;
 use lib "$FindBin::RealBin/../lib";
 use OpenQA::Script::CloneJob;
 use OpenQA::Utils 'assetdir';
+use Scalar::Util qw(looks_like_number);
 
 my %options;
 my $jobid;
@@ -190,7 +202,7 @@ sub parse_options () {
         "apikey:s", "apisecret:s", "verbose|v", "json-output|j",
         "skip-deps", "skip-chained-deps", "skip-download", "parental-inheritance",
         "help|h", "show-progress", "within-instance|w=s", "clone-children",
-        "max-depth:i", "ignore-missing-assets", "export-command",
+        "max-depth:i", "repeat|r:i", "ignore-missing-assets", "export-command",
     ) or usage(1);
     usage(0) if $options{help};
     usage(1) if $options{help} || ($options{'within-instance'} && $options{from});
@@ -207,6 +219,8 @@ sub parse_options () {
     die "missing job reference, see --help for usage\n" unless $jobid;
     ($options{'from'}, $jobid) = split_jobid($jobid) unless $options{'from'};
     usage(1) unless ($jobid && $options{'from'});
+    my $repeat = $options{'repeat'} //= 1;
+    die "invalid repeat count, should be greater than or equal to 1\n" if !looks_like_number($repeat) || $repeat < 1;
     $options{'dir'} ||= assetdir();
     $options{'host'} ||= 'localhost';
     $options{'args'} = \@ARGV;
--- a/t/35-script_clone_job.t
+++ b/t/35-script_clone_job.t
@@ -244,6 +244,27 @@ subtest 'export command' => sub {
     combined_like { clone_jobs(42, \%options) } $expected_cmd, 'openqa-cli command printed';
 };
 
+subtest 'cloning with repeat count' => sub {
+    my $ua_mock = Test::MockModule->new('Mojo::UserAgent');
+    my $clone_mock = Test::MockModule->new('OpenQA::Script::CloneJob');
+    my @post_args;
+    my $tx_handled = 0;
+    $ua_mock->redefine(post => sub { push @post_args, [@_] });
+    $clone_mock->redefine(handle_tx => sub ($tx, $url_handler, $options, $jobs) { $tx_handled = 1 });
+    my %fake_jobs = (42 => {id => 42, name => 'myjob', settings => {TEST => 'myjob'}});
+    $clone_mock->redefine(clone_job_get_job => sub ($job_id, @args) { $fake_jobs{$job_id} });
+    my %options = (host => 'foo', from => 'bar', repeat => 100, apikey => 'bar', apisecret => 'bar');
+    OpenQA::Script::CloneJob::clone_jobs(42, \%options);
+    is $options{repeat}, undef, 'repeat count has been removed from options';
+    subtest 'post args' => sub {
+        is scalar @post_args, 100, 'exactly 100 post calls made';
+        # check Nth test name ends with -N
+        is $post_args[0]->[3]->{'TEST:42'}, 'myjob-001', 'first index has been appended to test name';
+        is $post_args[41]->[3]->{'TEST:42'}, 'myjob-042', 'random index has been appended to test name';
+        is $post_args[99]->[3]->{'TEST:42'}, 'myjob-100', 'last index has been appended to test name';
+    } or diag explain \@post_args;
+};
+
 subtest 'overall cloning with parallel and chained dependencies' => sub {
     # do not actually post any jobs, assume jobs can be cloned (clone ID = original ID + 100)
     my $ua_mock = Test::MockModule->new('Mojo::UserAgent');


lib/OpenQA/Script/CloneJob.pm
```diff
@@ -226,11 +226,32 @@ sub handle_tx ($tx, $url_handler, $options, $jobs) {
...
+sub append_sequential_number_to_test_name ($n, $post_params) {
+    my $number_suffix = sprintf("-%02d", $n);
+    foreach my $job (keys %$post_params) {
+        my $test_name = $post_params->{$job}->{'TEST'};
+        if ($n == 1) {
+            $test_name .= $number_suffix;    # just append at the first time
+        }
+        else {
+            # from the second onwards, replace the suffix
+            substr($test_name, length($test_name) - length($number_suffix)) = $number_suffix;
```
Martchus, line 239: I guess the following would work as well:
```suggestion
            substr($test_name, -length($number_suffix)) = $number_suffix;
```

So have you actually tested this? I suppose this feature deserves a unit test.

lib/OpenQA/Script/CloneJob.pm
```diff
@@ -226,11 +226,32 @@ sub handle_tx ($tx, $url_handler, $options, $jobs) {
...
+        }
+        $post_params->{$job}->{'TEST'} = $test_name;
+    }
+}
+
 sub clone_jobs ($jobid, $options) {
     my $url_handler = create_url_handler($options);
+    my $how_many_runs = $options->{'repeat'};
+    delete $options->{'repeat'};    # now it's useless, we can remove it
```
Martchus: `$how_many_runs` is a very cumbersome variable name. You can also exploit the return value of `delete`. What about naming it just in accordance with the CLI option?

```suggestion
    my $repeat = delete $options->{'repeat'};
```

Martchus (changes requested):

Martchus:
I suppose this feature deserves a unit test.

lib/OpenQA/Script/CloneJob.pm
```diff
@@ -226,11 +226,32 @@ sub handle_tx ($tx, $url_handler, $options, $jobs) {
     }
 }
 
+# append a formatted "-NN" to the TEST parameter
+sub append_sequential_number_to_test_name ($n, $post_params) {
+    my $number_suffix = sprintf("-%02d", $n);
+    foreach my $job (keys %$post_params) {
+        my $test_name = $post_params->{$job}->{'TEST'};
```
Martchus, line 233: I'd avoid repeated hash access and unnecessary quoting (which tidy will likely complain about as well):
```suggestion
    foreach my $job_key (keys %$post_params) {
        my $job = $post_params->{$job_key};
        my $test_name = $job->{TEST};
```
Then use `$job` in further code instead of repeating `$post_params->{$job}`.

Martchus:
By the way, what prevents you from doing simply `for i in {01..50}; do openqa-clone-job … TEST+=-$i; done`? I always use a command similar like supposedly most people.

ilmanzo:
> By the way, what prevents you from doing simply `for i in {01..50}; do openqa-clone-job … TEST+=-$i; done`? I always use a command similar to that and I'm sure people on the team sometimes do the same.

No one prevents you, I guess it's just a matter of 'user experience', as discussed in the [ticket](https://progress.opensuse.org/issues/122953). You can always wrap the clone-job inside a for loop, but if you feel lazy just use the `--repeat` parameter (or do both) :)

foursixnine:
Looks pretty good

docs/UsersGuide.asciidoc
```diff
@@ -1113,6 +1113,19 @@ timeouts by performing the task in background.
 This is recommended on big instances but means that the results (and
 possible errors) need to be polled via `openqa-cli api isos/$scheduled_product_id`.
 
+
+===== Statistical investigation ===== 
+In case issues appear sporadically and are therefore hard to reproduce it can help to trigger many more jobs on a production instance to gather more data first, for example the failure ratio.
```
okurz, line 1118: please wrap all lines consistently at 80 characters

docs/UsersGuide.asciidoc
```diff
@@ -1113,6 +1113,19 @@ timeouts by performing the task in background.
...
+
+Example of triggering 50 jobs in the development group so that the result of passed/failed jobs is counted by openQA itself on the corresponding overview page:
```
okurz, line 1120:
```suggestion
Example of triggering 50 jobs in a development group so that the result of passed/failed jobs is counted by openQA itself on the corresponding overview page:
```

lib/OpenQA/Script/CloneJob.pm
```diff
@@ -226,11 +226,32 @@ sub handle_tx ($tx, $url_handler, $options, $jobs) {
...
+}
+
 sub clone_jobs ($jobid, $options) {
     my $url_handler = create_url_handler($options);
+    my $repeat = delete $options->{'repeat'} || 1;
     clone_job($jobid, $url_handler, $options, my $post_params = {}, my $jobs = {});
-    my $tx = post_jobs($post_params, $url_handler, $options);
-    handle_tx($tx, $url_handler, $options, $jobs) if $tx;
+    for my $counter (1 .. $repeat) {
+        append_sequential_number_to_test_name($counter, $post_params) if $repeat > 1;
```
okurz, line 251: How about

```suggestion
        append_idx_to_test_name($counter, $post_params) if $repeat > 1;
```

script/openqa-clone-job
```diff
@@ -133,6 +133,10 @@ are cloned. Use C<0> to denote infinity.
 A shortcut for C<--skip-download --from HOST --host HOST> to clone a job within
 a local or remote instance.
 
+=item B<--repeat> N
+
+Do the same operation N times. Allowed values are from 1 to 99
```
okurz, line 138: Why that limit? I see good use cases for much higher numbers and besides no one stops you from calling the command with repeat like 1000 times except for a rate limit on the server which is out of scope for the addition to the client scripts. I would just not limit here at all

okurz (changes requested):
Sounds like a nice idea but please provide much more detail into your git commit message to be able to follow your motivation and design choices

script/openqa-clone-job
foursixnine: > Why that limit? I see good use cases for much higher numbers, and besides, no one stops you from calling the command with repeat like 1000 times except for a rate limit on the server which is out of scope for the addition to the client scripts. I would not limit myself here at all

This basically invites a mistake of scheduling 1k jobs :D and more

okurz: Still any limit is arbitrary and I see valid use cases to schedule more than thousand jobs. Please just don't limit here at all. openqa-clone-job is just not the place for it to decide. If you really want to prevent mistakes then make it a soft-limit of let's say 300 and allow to override with another parameter, e.g. "--force-repeat". I would just go with no limit for now. After all the iteration would still be done client-side so scheduling 1k jobs takes possibly some minutes and the user can still abort at any time with ^C

Martchus: If you don't limit at all, then make also sure the code for appending the index can handle a variable number of digit. The code in its current form would not be able to handle that because it replaces the previous index using the number of digits of the current index which would not work when the number of digits actually differs.

ilmanzo: I'd second point from @Martchus; of course, it's possible to handle variable digit numbers, but IMHO It's not really worth it because adding more digits makes the code more complicated for the (small?) percentage of use cases when is really needed to clone a job over 1000 times, which BTW can be accomplished if the user just press up arrow and repeat the previous command. 
So my proposal is to keep it to 3 digits and see if there is a need, we can always improve on that.

lib/OpenQA/Script/CloneJob.pm
```diff
@@ -226,11 +226,32 @@ sub handle_tx ($tx, $url_handler, $options, $jobs) {
...
+    my $number_suffix = sprintf("-%03d", $n);
+    foreach my $job_key (keys %$post_params) {
+        my $job = $post_params->{$job_key};
+        my $test_name = $job->{TEST};
+        if ($n == 1) {
+            $test_name .= $number_suffix;    # just append at the first time
+        }
+        else {
+            # from the second onwards, replace the suffix
+            substr($test_name, -length($number_suffix)) = $number_suffix;
```
perlpunk, line 240:
```suggestion
            $test_name =~ s/-\d+$/$number_suffix/;
```
That doesn't rely on the length of the previous number.

okurz:
I saw that you updated your PR but the suggestions from comments haven't been implemented yet. Can you implement those or comment on which different path you want to take?

script/openqa-clone-job
kalikiana: I regularly schedule 300 jobs at once and if you added a hard limit you'd only force me to make mistakes by having to use shell syntax to work around it ;-)

kalikiana: Also, thanks for adding this. I'll definitely use it

lib/OpenQA/Script/CloneJob.pm
ilmanzo: this is a good idea if we want to handle variable number of digits (eg. leave unlimited the `--repeat` count). We also need to properly format the `$number_suffix`, e.g. if the user wants 32768 jobs cloned, he will get 00001, 00002 etc

ilmanzo:
> I suppose this feature deserves a unit test.

Agree; added a new unit test

script/openqa-clone-job
ilmanzo: Thanks, 300 is well under the current (updated) 999 limit, so seems it's good also for you :)

script/openqa-clone-job
```diff
@@ -207,6 +211,8 @@ sub parse_options () {
     die "missing job reference, see --help for usage\n" unless $jobid;
     ($options{'from'}, $jobid) = split_jobid($jobid) unless $options{'from'};
     usage(1) unless ($jobid && $options{'from'});
+    $options{'repeat'} ||= 1;
+    die "invalid repeat count, should be between 1 and 999" if ($options{'repeat'} < 1 or $options{'repeat'} > 999);
```
Martchus, line 215: This should probably check whether the user-specified value is actually a number:
```suggestion
    my $repeat = $options{'repeat'} //= 1;
    die "invalid repeat count, should be between 1 and 999\n" if !looks_like_number($repeat) || $repeat < 1 || $repeat > 999;
```

I would also prefer `||` over `or` except for command-chaining. Adding an `"\n"` at the end of the error message prevents printing the line number making this user error more look like one.

You also have to use `//=` instead of `||=` because otherwise the error message would not be shown in case the user specifies `0`.

And as others have already suggested it would make sense to remove the rather arbitrary limit and use @perlpunk 's suggestion to handle it.

Martchus (changes requested):

script/openqa-clone-job
ilmanzo: good point, thanks for the suggestion. I will change the code and adjust unit test accordingly.

kalikiana (approved):

Martchus (approved):

lib/OpenQA/Script/CloneJob.pm
```diff
@@ -226,11 +226,31 @@ sub handle_tx ($tx, $url_handler, $options, $jobs) {
...
+        my $job = $post_params->{$job_key};
+        my $test_name = $job->{TEST};
+        if ($n == 1) {
+            $test_name .= '-1';    # just append at the first time
+        }
+        else {
+            # from the second onwards, replace with the number
+            $test_name =~ s/-\d+$/\-$n/;
+        }
+        $job->{TEST} = $test_name;
```
okurz, line 241:
```suggestion
        # just append at the first time. From the second onwards, replace with the number
        $job->{TEST} = $n == 1 ? '-1' : $job->{TEST} =~ s/-\d+$/\-$n/r;
```

script/openqa-clone-job
```diff
@@ -133,6 +133,10 @@ are cloned. Use C<0> to denote infinity.
 A shortcut for C<--skip-download --from HOST --host HOST> to clone a job within
 a local or remote instance.
 
+=item B<--repeat> N
+
+Do the same clone operation N times.
```
okurz, line 138: I had been thinking about the name and the default. I think a more common and better name would be "count". For "repeat" with default "1" one might assume that a job is cloned *and* repeated one time equaling two jobs. For "count=1" the case is clear. Also I tink multiple other commands use a command line option with "-c/--count" so how about renaming? The logic can stay the same

okurz (changes requested):

lib/OpenQA/Script/CloneJob.pm
ilmanzo: Won't this replace the whole test name with `-1` on first pass ? I think we should keep the original name and just append the number

okurz: ah, yes of course. Corrected my suggestion

script/openqa-clone-job
ilmanzo: yeah, naming things + off by one in the same topic :) 
I get your point, this is already a 'clone' operation so if you want to do it only once you just avoid to use repeat. The documentation seems clear, do it N times when N by default is 1 . Should we give an error if the user specifies `repeat=1` ? 

Still, 'count' to me sounds more like a verb, like "please count and tell me how many jobs I cloned" . How about `times` ?

lib/OpenQA/Script/CloneJob.pm
ilmanzo: Honestly, I am a bit against this kind of oneliners. I know it's cool to "compress" 9 lines in one, but you gain nothing in terms of performance or feature, and just make the code more difficult to understand for a contributor. I'm in favor of clean, expressive code, even if it's Perl :)

script/openqa-clone-job
okurz: hm, no, `times` is too confusing with either `time` or as in multiplying. I wanted to look up examples where other commands do something similar. I only found `dd` which says

```
  count=N         copy only N input blocks
```

If you don't like count then we can just stick with `repeat`

lib/OpenQA/Script/CloneJob.pm
okurz: This is not about oneliners but about conditional assignments using the ternary operator. if/else should be used when different branches in execution are taken but here it's not really about a procedure but in the end you want to have a value assigned to `$job->{TEST}`. If you think this is not clear enough then maybe we should find a way where we don't need to replace the detected number but instead make sure that we always apply a sensible suffix

okurz: oh, I just realized that the suffix is not selected in a good way because "-2" would be sorted after "-19". So instead you should look into a way how the numbers of necessary digits can be derived to be equivalent to the bash for loop:

```
for i in {001..100} ; …
```

which yields numbers

```
001
002
…
010
…
099
100
```

with the zeroes prefixed.

foursixnine: The way it is right now looks clear enough to me, and I see no benefit over using a ternary operator instead, also maybe I'm not following the code correctly but, shouldn't `$job->{TEST} = $test_name."-$n"` be enough here? or am I missing something?

ilmanzo: @foursixnine If we simply append the number each iteration, we will end up with jobs named foobar-1-2-3-4-5-6 and so on :) 

My first proposal to have a fixed amount of digits was intended to obtain clean formatting with zeroes prefixed; so let's think about a way to format it properly with a variable number of digits.

foursixnine: for the padding, `sprintf("%04d", $n)` should solve it, but from what you are saying, each job is a clone of the previous one and not the original one?

foursixnine: I wonder if getting the test name or passing it to the `append_idx_to_test_name` is something that can be done, after all.. by the time we enter in the for loop, `post_params` is populated already by clone_job, so extracting the original name of the test is/should be possible..

ilmanzo: job is cloned once, and we only change the TEST name in the parameters hash by appending a number. Each iteration keeps the hash, so if we just append, it will be appended each time. As was requested to have no limit in `repeat` , maybe 04 is not enough, so we should do something like

```Perl
sub clone_jobs ($jobid, $options) {
    my $url_handler = create_url_handler($options);
    my $repeat = delete $options->{repeat} || 1;
    my $digits = length $repeat;
    clone_job($jobid, $url_handler, $options, my $post_params = {}, my $jobs = {});
    for my $counter (1 .. $repeat) {
        if ($repeat > 1) {
            my $number = sprintf("-%0${digits}d", $counter);
            append_idx_to_test_name($number, $post_params);
        }
        my $tx = post_jobs($post_params, $url_handler, $options);
        handle_tx($tx, $url_handler, $options, $jobs) if $tx;
    }
}
```

ilmanzo: And BTW while I understand the "no limit" argument, I still wonder how many use cases we have to clone a job more than thousands times :D

ilmanzo: > I wonder if getting the test name or passing it to the `append_idx_to_test_name` is something that can be done, after all.. by the time we enter in the for loop, `post_params` is populated already by clone_job, so extracting the original name of the test is/should be possible..

we should really talk about test `names`, because a clone operation can lead to parents/children/parallel jobs also getting cloned, each with its own name to modify. So we need an iteration over the nested hash 

```Perl
# append a formatted "-NN" to the TEST parameter
sub append_idx_to_test_name ($n, $post_params) {
    foreach my $job_key (keys %$post_params) {
        my $job = $post_params->{$job_key};
        my $test_name = $job->{TEST};
        ...
    }
}
```

foursixnine: > And BTW while I understand the "no limit" argument, I still wonder how many use cases we have to clone a job more than 9 thousands times :D

The limit here, is me trying to workaround openQA not having api-rate limiting, nothing limits the user to circumvent this, but at least lets make them write a `for loop` or hack their way across 🙃

script/openqa-clone-job
foursixnine: > I had been thinking about the name and the default. I think a more common and better name would be "count". For "repeat" with default "1" one might assume that a job is cloned _and_ repeated one time equaling two jobs. For "count=1" the case is clear. Also I tink multiple other commands use a command line option with "-c/--count" so how about renaming? The logic can stay the same

looking at the documented example, 

```
openqa-clone-job --skip-chained-deps --repeat=50 --within-instance \
https://openqa.opensuse.org 123456  BUILD=poo32242_investigation \
_GROUP="Test Development:openSUSE Tumbleweed"
```

I read: I want to clone this job, repeat the operation 50 times and group it onto a build X, inside group Y.

@ilmanzo can you add a "This can be useful for Statistical investigation.."? or a similar comment, to supplement line 138:

> Clone job A, repeat the operation N times and group it onto a build X, inside group Y.
> 

And add an example command line.

foursixnine (approved):

lib/OpenQA/Script/CloneJob.pm
foursixnine: > about test names for this use case a good way to start is
> 
> ```
> openqa-clone-job --skip-chained-deps --parental-inheritance --within-instance …
> ```
> 
> with the "--skip-chained-deps" no parents in upper layers, e.g. "image creation jobs", are cloned and with "--parental-inheritance" test parameters affect cloned parents+children.

The dependencies will end up or should end up with their dependencies where they should be, if the interest is 10 tries of X job, the parents only become relevant when they fail, and if, the children will have a reference. 

If somebody wishes also extra changes to the whole chain, they should use isos post, in any case, I'd leave handling of dependencies (parent jobs et al) to a separate PR.

ilmanzo: With the latest change, I addressed the padding issue

ilmanzo: > about test names for this use case a good way to start is
> 
> ```
> openqa-clone-job --skip-chained-deps --parental-inheritance --within-instance …
> ```
> 
> with the "--skip-chained-deps" no parents in upper layers, e.g. "image creation jobs", are cloned and with "--parental-inheritance" test parameters affect cloned parents+children.

Sorry but I don't understand what this has to do with the PR itself ? With the actual implementation, all cloned tests (parents, siblings and so on) gets the name numbered. If you skip cloning parents, only childrens will have the name modified.

okurz: Yes, that's fine. I wanted to provide that users can still specify `--parental-inheritance` if they want custom test settings overridden in all created jobs.

okurz:
From https://app.circleci.com/pipelines/github/os-autoinst/openQA/12328/workflows/dec7ae53-201a-46cc-9558-0a06c041ce8a/jobs/115056 it says "Cloning children of myjob". Please make sure that all output is caught by Test::Output functions or other means. Possibly this test or others already provide good examples.

lib/OpenQA/Script/CloneJob.pm
okurz: I still https://github.com/os-autoinst/openQA/pull/5331#discussion_r1362558824 as open. I suggest to use the ternary operator for the alternatives in assignment. As an alternative maybe we can find an approach so that we don't need to replace the value in all. How about reading "TEST" from the first job and just overwriting the test variable in the for-loop with an iterating value. WDYT?

foursixnine: > I still [#5331 (comment)](https://github.com/os-autoinst/openQA/pull/5331#discussion_r1362558824) as open. I suggest to use the ternary operator for the alternatives in assignment. 

This sounds like a strawman argument over something that has zero value.

Suggesting usage of a ternary operator, over readability doesn't make much sense... otherwise I'd suggest to come up with a tidy/perlcritic rule that enforces ternary operators

okurz: > This sounds like a strawman argument over something that has zero value.

Well, I stated it because I do see the value. Sorry to see that you think I would provide something with no value.

I wouldn't know how to put that into a perlcritic rule. And for everything that we can't or do not know how to put into a static check we have the human review process, right? So this issue here right now is the only open point that I know of. I see the following possibilities:
1. Use a ternary operator for the assignment
2. Follow my idea about reading TEST on the outer scope and iterate over that
3. Ask me to do provide a working implementation – I would be happy to try – and add to your branch
4. State explicitly that you will not change it and you definitely think this is better. Then I would need to think about.

kalikiana: Can I ask a very naive question? Why can't we simply print the current iterator always without any conditionals? The bash equivalent does that.

ilmanzo: Thanks @okurz for the review. I already expressed my opinion on ternary operators and I would like to avoid the usage when possible, like in this case.
About the second option, it was my first implementation but it requirese to keep the "original" set of test names into an hash `{ job_key => name }` and assign them to the proper numbered value just before posting, or change it on the fly. I assumed this would complicate the code too much, because you also need to have the "original" hash in the `append_idx` function; so I came up with another proposal, which I just submitted including the fix for unit test output.  

```Perl
# append a formatted "-NN" to the TEST parameter for each job being posted
sub append_idx_to_test_name ($n, $digits, $post_params) {
    my $suffix = sprintf("-%0${digits}d", $n);
    foreach my $job_key (keys %$post_params) {
        my $job = $post_params->{$job_key};
        if ($n == 1) {
            $job->{TEST} .= $suffix;    # just append at the first time
        }
        else {
            # from the second onwards, replace old with the new number
            $job->{TEST} =~ s/-\d+$/$suffix/;
        }
    }
}
```

of course the owner will have the last word :)

ilmanzo:
> From https://app.circleci.com/pipelines/github/os-autoinst/openQA/12328/workflows/dec7ae53-201a-46cc-9558-0a06c041ce8a/jobs/115056 it says "Cloning children of myjob". Please make sure that all output is caught by Test::Output functions or other means. Possibly this test or others already provide good examples.

good point, fixed with latest commit-push

lib/OpenQA/Script/CloneJob.pm
ilmanzo: > Can I ask a very naive question? Why can't we simply print the current iterator always without any conditionals? The bash equivalent does that.

If we want to do exactly the same thing the bash for..loop does, we can just move the `clone_job` inside the `for` loop and you will get each iteration a "clean" TEST name. 
At this point, it's just a matter of appending the -00NN number, and job done. 

My proposal sounds to me like an occasion to use fewer resources, avoiding repeating N times the same `clone` operation and doing it only once, then `POST` N times the cloned job(s) data structure.


```Perl
sub clone_jobs ($jobid, $options) {
    my $url_handler = create_url_handler($options);
    my $repeat = delete $options->{repeat} || 1;
    my $digits = length $repeat;
    clone_job($jobid, $url_handler, $options, my $post_params = {}, my $jobs = {});
    for my $counter (1 .. $repeat) {
        append_idx_to_test_name($counter, $digits, $post_params) if $repeat > 1;
        my $tx = post_jobs($post_params, $url_handler, $options);
        handle_tx($tx, $url_handler, $options, $jobs) if $tx;
    }
}
```

okurz: Ok. That sounds similar to something that @perlpunk and me have just come up with. Please take a look into https://github.com/os-autoinst/openQA/pull/5336 where I added a second commit with the refactoring on top of your branch. If you like the approach you can cherry-pick my commit https://github.com/os-autoinst/openQA/pull/5336/commits/6341433ee715ccdde940baf97adc02c4c87e04b8 into your branch or we merge mine.

perlpunk: But `clone_job` is actually doing some requests to get the dependencies, so I think @ilmanzo is right that this is maybe using too many resources. I wouldn't care about 100 loops of some perl code, but if it is doing 100 requests, then it's not that nice.

ilmanzo: NP, feel free to merge yours; thanks again because I learned a lot from this exchange of views!

ilmanzo: @perlpunk Yes, my proposal is to avoid what the bash loop does:

```
- clone
- post
- clone
- post
- clone
- post
- clone
- post
...
```
but do a more efficient  
```
- clone
- post
- post 
- post
- post
... 
```
even at the cost of the code being (slightly) more complex :)

ilmanzo: said in another way, for a `--repeat N` this implementation will spare N-1 `GET` calls to the API server, which are needed to download job dependencies on each clone operation.

okurz (approved):

lib/OpenQA/Script/CloneJob.pm
okurz: Got it. I plan to follow up with rework after that while of course keeping the approach regarding avoiding unneeded requests. At the very least we should avoid rename functions as we have identified in #5336 . That of course is something that was already in the codebase so not at all any critic about this PR here

okurz:
@mergifyio rebase

okurz:
thank you. With the rebase request my intent was the sam but there was nothing to rebase so that did not retrigger anything

Martchus:
I've just used this parameter to create https://openqa.suse.de/tests/overview?version=5.5&build=Martchus%2Fos-autoinst-distri-opensuse%2318034-m2048&distri=sle-micro and admittedly, it is more convenient than the for loop :-)

PR #5332, opened 2023-10-17
Update Multi Machine docs to reference only os-autoinst-setup-multi-machine

asdil12:
Ticket: https://progress.opensuse.org/issues/133025


docs/Networking.asciidoc
```diff
@@ -82,299 +70,52 @@ firewall is not covered in all necessary details in this documentation.
...
-zypper -n in openvswitch
-systemctl enable --now openvswitch
+# specify the number of test VMs to run on this host
+instances=30 bash -x $(which os-autoinst-setup-multi-machine)
 ----
 
-*  Install and configure _os-autoinst-openvswitch.service_:
+==== What os-autoinst-setup-multi-machine does
+
+===== Set Up Open vSwitch
```
Martchus, line 88:
```suggestion
===== Set up Open vSwitch
```

docs/Networking.asciidoc
```diff
@@ -82,299 +70,52 @@ firewall is not covered in all necessary details in this documentation.
...
+NOTE: The bridge device will also call a script at
+`/etc/wicked/scripts/gre_tunnel_preup.sh` on _PRE_UP_.
+This script needs *manual* touch if you want to set up multiple
+multi-machine worker hosts. Refer to the *GRE tunnels* section below
+for further information.
 
-* Add a tap interface for every multi-machine worker instance:
+===== Configure NAT with firewalld
+The required firewall rules, zone configuration and NAT will be set up.
+IP-Forwarding will be enabled.
```
Martchus, line 114: Maybe it is worth stating how the zone is called.

docs/Networking.asciidoc
```diff
@@ -82,299 +70,52 @@ firewall is not covered in all necessary details in this documentation.
...
-
-==== Grant CAP_NET_ADMIN Capabilities to QEMU
-In order to let QEMU create TAP devices on demand it is required to set
-CAP_NET_ADMIN capability on QEMU binary file:
-
-[source,sh]
-----
-zypper -n in libcap-progs
-setcap CAP_NET_ADMIN=ep /usr/bin/qemu-system-x86_64
-----
```
Martchus, line 313: It doesn't look like this is done by the new script at all so we should rather preserve this section (as we'd otherwise lose this information completely). Or is it really completely outdated or never wanted to consider this kind of setup?

docs/Networking.asciidoc
```diff
@@ -409,21 +150,36 @@ PRE_UP_SCRIPT="wicked:gre_tunnel_preup.sh"
...
+[source,sh]
+----
+# /etc/openqa/workers.ini
+[global]
+WORKER_CLASS = qemu_x86_64,tap
+----
+
+NOTE: The number of tap devices should correspond to the number of the running
+worker instances. For example, if you have set up 3 tap devices, the same
+number of worker instances should be configured.
```
Martchus, line 171: Swapping the order here would be more natural to me:
```suggestion
worker instances. For example, if you have set up 3 worker instances, the same
number of tab devices should be configured.
```
Or just remove the example because the preceding sentence should be clear enough on its own:
```suggestion
worker instances.
```

docs/Networking.asciidoc
```diff
@@ -409,21 +150,36 @@ PRE_UP_SCRIPT="wicked:gre_tunnel_preup.sh"
...
+
+Enable worker instances to be started on system boot:
+
+[source,sh]
+----
+systemctl enable openqa-worker@1
+systemctl enable openqa-worker@2
+systemctl enable openqa-worker@3
```
Martchus, line 179: It would make sense to demonstrate doing this in a less manual way (even if the syntax might not be supported by all shells):
```suggestion
systemctl enable openqa-worker@{1..3}
```

Martchus (changes requested):

docs/Networking.asciidoc
```diff
@@ -82,299 +70,52 @@ firewall is not covered in all necessary details in this documentation.
 NOTE: Another way to setup the environment with _iptables_ and _firewalld_ is described
 on the link:https://fedoraproject.org/wiki/OpenQA_advanced_network_guide[Fedora wiki].
 
-NOTE: The salt states contained by
+NOTE: Yet another way to setup multi machine tap setup is by using the salt states.
 https://github.com/os-autoinst/salt-states-openqa[salt-states-openqa] contain all the
```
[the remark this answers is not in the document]
kalikiana: Perhaps it can be a lot simpler without repeating the same phrasing?

```suggestion
NOTE: Alternatively https://github.com/os-autoinst/salt-states-openqa[salt-states-openqa] contains
```

docs/Networking.asciidoc
```diff
@@ -82,299 +70,52 @@ firewall is not covered in all necessary details in this documentation.
...
-IPADDR='10.0.2.2/15'
-STARTMODE='auto'
-ZONE='trusted'
-OVS_BRIDGE='yes'
-OVS_BRIDGE_PORT_DEVICE_0='tap0'
-----
+NOTE: The bridge device will also call a script at
+`/etc/wicked/scripts/gre_tunnel_preup.sh` on _PRE_UP_.
+This script needs *manual* touch if you want to set up multiple
+multi-machine worker hosts. Refer to the *GRE tunnels* section below
```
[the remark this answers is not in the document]
kalikiana:
```suggestion
multi-machine worker hosts. Refer to the <<GRE tunnels>> section below
```

docs/Networking.asciidoc
okurz: apparently outdated and not needed anymore, see https://github.com/os-autoinst/os-autoinst/pull/2361

docs/Networking.asciidoc
```diff
@@ -82,299 +70,52 @@ firewall is not covered in all necessary details in this documentation.
...
+# specify the number of test VMs to run on this host
+instances=30 bash -x $(which os-autoinst-setup-multi-machine)
 ----
 
-*  Install and configure _os-autoinst-openvswitch.service_:
+==== What os-autoinst-setup-multi-machine does
+
+===== Set Up Open vSwitch
+
+The script will install and configure openvswitch as well as
```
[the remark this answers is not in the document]
asdil12: I spelled it like that, because this is the name of the RPM.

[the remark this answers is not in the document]
kalikiana: That is the spelling used e.g. on their website, though. So if you're not saying `... package` I'd lean towards agreeing with Marius

docs/Networking.asciidoc
```diff
@@ -82,299 +70,59 @@ firewall is not covered in all necessary details in this documentation.
...
 
-==== Configure virtual interfaces
-
-* Add the bridge config including the part for all the tap devices that should
-  be connected to it. The file has to be located in the
-  `/etc/sysconfig/network/` directory. File name is `ifcfg-br<N>`, where `<N>`
-  is the id of the bridge (e.g. `1`):
+===== Configure NAT with firewalld
+The required firewall rules for masquerading (NAT) and zone configuration
+for the trusted zonewill be set up. The bridge devices will be added to
```
Martchus, line 113:
```suggestion
for the trusted zone will be set up. The bridge devices will be added to
```

PR #5333, opened 2023-10-17
Extend Test Module Documentation & Explain Python integration

josegomezr:
- Extended the OpenQA Test Module documentation with improved structure and examples.

  The module interface is documented (all the possible routines and their purpose are explained & trackable in the sidebar).

- Explained how the python support works in openQA.

  Added a small section on how python support is provided by openqa, as well as the limitations that the FFI Layer brings.

- Reorganized the example test modules.

### DIFF
--- a/docs/WritingTests.asciidoc
+++ b/docs/WritingTests.asciidoc
@@ -354,7 +354,7 @@ implementing the interface described above.
 --------------------------------------------------------------
 from testapi import *
 
-def run():
+def run(self):
     assert_screen('openqa-logged-in')
     assert_and_click('openqa-search')
     type_string('shutdown.pm')
@@ -373,11 +373,11 @@ def run():
 def switch_to_root_console():
     send_key('ctrl-alt-f3')
 
-def post_fail_hook():
+def post_fail_hook(self):
     switch_to_root_console()
     assert_script_run('openqa-cli api experimental/search q=shutdown.pm')
 
-def test_flags():
+def test_flags(self):
     return {'fatal': 1}
 --------------------------------------------------------------
 
--- a/docs/WritingTests.asciidoc
+++ b/docs/WritingTests.asciidoc
@@ -1,6 +1,6 @@
 
 [[writingtests]]
-= openQA tests developer guide
+= openQA test developer guide
 :toc: left
 :toclevels: 6
 :author: openQA Team
@@ -22,52 +22,230 @@ https://github.com/os-autoinst/openQA[official repository].
 == Basic
 [id="basic"]
 
-This section explains the basic layout of openQA tests and the API available in tests.
-openQA tests are written in the *Perl* programming language. Some basic but no
-in-depth knowledge of Perl is needed. This document assumes that the reader
-is already familiar with Perl.
+This section explains the basic layout of an openQA test and the API available.
+Tests are written in the *Perl* programming language. However there is support
+for the *Python* programming language (through the Perl module
+`Inline::Python`).
 
-== API
+Some basic but no in-depth knowledge of Perl or Python is needed. This document
+assumes that the reader is already familiar with Perl or Python.
+
+== Test API
 [id="api"]
 :testapi: https://github.com/os-autoinst/os-autoinst/blob/master/testapi.pm[os-autoinst]
 
-{testapi} provides the API for the tests using the os-autoinst backend, you can
-take a look to the published documentation at http://open.qa/api/testapi/. This test API
-is sometimes also referred to as an openQA DSL, because in some contexts it can look like
-a domain specific language.
+{testapi} provides the API for the test using the os-autoinst backend. Take a
+look at the http://open.qa/api/testapi[test API documentation] for further
+information. Note that this test API is sometimes also referred to as an openQA
+DSL, because in some contexts it can look like a domain specific language.
 
 == How to write tests
 
-openQA tests need to implement at least the *run* subroutine to
-contain the actual test code and the test needs to be loaded in the distribution's
-main.pm.
+=== Test module interface
 
-The *test_flags* subroutine specifies what should happen when test execution of the
-current test module is finished depending on the result.
-If we should skip execution of the following test modules if current one failed, or it
-should be used to create a snapshot of the SUT to rollback to. The following flags
-are supported:
+An openQA test needs to implement at least the `run` subroutine containing the
+actual test code and the test needs to be loaded in the distribution's
+`main.pm`.
 
-* *fatal*: The whole test suite is aborted if the test module fails. The overall state
-  is set to `failed`.
-* *ignore_failure*: If this module fails, it will not affect the overall result at all.
-* *milestone*: After this test succeeds, update the 'lastgood' snapshot.
-* *no_rollback*: Don't roll back to the 'lastgood' snapshot if the test module fails.
-* *always_rollback*: Roll back to the 'lastgood' snapshot even if test was successful.
+Here is an example in Perl:
 
-See the example below for how to enable a test flag.
-Note that snapshots are only supported by the QEMU backend. When using other backends
-`fatal` is therefore enabled by default. One can explicitly set it to `0` to disable
-the behavior for all backends even though it is not possible to roll back.
+[source,perl]
+-------------------------------------------------------------------
+use Mojo::Base "basetest";
+use testapi;
 
-There are several callbacks defined:
+sub run () {
+    # write in this block the code for your test.
+}
+-------------------------------------------------------------------
 
-* *post_fail_hook* is called to upload log files or determine the state of
-  the machine
-* *pre_run_hook* is called before the run function - mainly useful for a whole
-  group of tests
-* *post_run_hook* is run after successful run function - mainly useful for a whole
-  group of tests
+And here is an example in Python:
+
+[source,perl]
+-------------------------------------------------------------------
+from testapi import *
+
+def run():
+    # write in this block the code for your test.
+-------------------------------------------------------------------
+
+There are more optional subroutines that can be defined to extend the behavior
+of a test. A test must comply with the interface defined below.
+_Please note that the subroutine marked with `*1` are optional._
+
+[source,python]
+-------------------------------------------------------------------
+# Written in type-hinted python to indicate explicitly return types
+def run(): -> None
+def test_flags(): -> dict # *1
+def post_fail_hook(): -> None # *1
+def pre_run_hook(): -> None # *1
+def post_run_hook(): -> None # *1
+-------------------------------------------------------------------
+
+==== `run`
+Defines the actual steps to be performed during the module execution.
+
+An example usage:
+
+[source,perl]
+-------------------------------------------------------------------
+sub run () {
+    # wait for bootloader to appear
+    # with a timeout explicitly lower than the default because
+    # the bootloader screen will timeout itself
+    assert_screen "bootloader", 15;
+
+    # press enter to boot right away
+    send_key "ret";
+
+    # wait for the desktop to appear
+    assert_screen "desktop", 300;
+}
+-------------------------------------------------------------------
+
+`assert_screen` & `send_key` are provided by {testapi}.
+
+==== `test_flags`
+
+Specifies what should happen when test execution of the current test module is
+finished depending on the result.
+
+Each flag is defined with a hash key, the possible hash keys are:
+
+* *fatal*: When set to `1` the whole test suite is aborted if the test module
+   fails. The overall state is set to `failed`.
+* *ignore_failure*: When set to `1` and the test module fails, it will not
+   affect the overall result at all.
+* *milestone*: After this test succeeds, update the 'lastgood' snapshot of the
+   SUT.
+* *no_rollback*: Don't roll back to the 'lastgood' snapshot of the SUT if the
+   test module fails.
+* *always_rollback*: Roll back to the 'lastgood' snapshot of the SUT even if
+   test was successful.
+
+See the example below for how to enable a test flag. Note that snapshots are
+only supported by the QEMU backend. When using other backends `fatal` is
+therefore enabled by default. One can explicitly set it to `0` to disable the
+behavior for all backends even though it is not possible to roll back.
+
+An example usage:
+
+[source,perl]
+-------------------------------------------------------------------
+sub test_flags () {
+    return {fatal => 1};
+}
+-------------------------------------------------------------------
+
+==== `pre_run_hook`
+
+It is called before the run function - mainly useful for a whole group of tests.
+It is useful to setup the start point of the test.
+
+An example usage:
+
+[source,perl]
+-------------------------------------------------------------------
+sub pre_run_hook () {
+    # Make sure to begin the test in the root console.
+    select_console 'root-console';
+}
+-------------------------------------------------------------------
+
+==== `post_fail_hook`
+
+It is called after `run()` failed. It is useful to upload log files
+or to determine the state of the machine.
+
+An example usage:
+
+[source,perl]
+-------------------------------------------------------------------
+sub post_fail_hook () {
+    # Take an screenshot when the test failed
+    save_screenshot;
+}
+-------------------------------------------------------------------
+
+==== `post_run_hook`
+
+It is called after `run()` regardless of the result of the test run.
+
+An example usage:
+
+[source,perl]
+-------------------------------------------------------------------
+sub post_fail_hook () {
+    send_key 'ctrl-alt-f3';
+
+    assert_script_run 'openqa-cli api experimental/search q=shutdown.pm' ;    
+}
+-------------------------------------------------------------------
+
+=== Notes on the Python API
+[id="notes-python-api"]
+
+The Python integration that OpenQA offers through `Inline::Python` also allows
+the test modules to import other Perl modules with the usage of the `perl`
+virtual package provided by `Inline::Python`.
+
+Because of the way `Inline::Python` binds Perl functions to Python it is not
+possible to use keywords arguments from Python to Perl functions. They must be
+passed as positional arguments, for example `"key", "value"`.
+
+See the following snippet of Perl code
+
+[source,perl]
+-------------------------------------------------------------------
+use x11utils;
+
+# [...] omitted for brevity
+
+sub run () {
+    # [...] omitted for brevity
+
+    # Start vncviewer - notice the named arguments
+    x11_start_program('vncviewer :0',
+        target_match => 'virtman-gnome_virt-install',
+        match_timeout => 100
+    );
+    # [...] omitted for brevity
+}
+-------------------------------------------------------------------
+
+versus the equivalent python code:
+
+[source,python]
+-------------------------------------------------------------------
+from testapi import *
+
+# [...] omitted for brevity
+
+def run():
+    perl.require('x11utils')
+
+    # Start vncviewer - notice the named arguments passed as positional arguments
+    # Formatted in pairs for better visibility.
+
+    perl.x11utils.x11_start_program('vncviewer :0',
+        'target_match', 'virtman-gnome_virt-install',
+        'match_timeout', 100
+    )
+    # [...] omitted for brevity
+-------------------------------------------------------------------
+
+=== Example Perl test modules
+[id="testmodule_perl_examples"]
+
+The following examples are short complete test modules written in Perl
+implementing the interface described above.
+
+==== Boot to desktop
+[id="testmodule_perl_boot"]
+
+[caption="Example: "]
+.Boots into desktop when pressing enter at the boot loader screen.
 
 The following example is a basic test that assumes some live image
 that boots into the desktop when pressing enter at the boot loader:
@@ -77,7 +255,7 @@ that boots into the desktop when pressing enter at the boot loader:
 use Mojo::Base "basetest";
 use testapi;
 
-sub run {
+sub run () {
     # wait for bootloader to appear
     # with a timeout explicitly lower than the default because
     # the bootloader screen will timeout itself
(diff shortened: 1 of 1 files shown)


okurz (changes requested):
There is a bit of over-capitalization in the git commit message subject line :)

* Also s/openqa/openQA/
* Please explain "FFI Layer"

josegomezr:
> Please explain "FFI Layer"

`Inline::Python` is a wrapper of the python API providing perl capabilities to eval python snippets. It wraps the CPython C API.

It's then used in `os-autoinst/autotest.pm` circa line 120 ther's a self evaled string block that imports `Inline::Python`:

```perl
#[...]

use Inline::Python qw(py_eval py_bind_func py_study_package);
py_eval(<<'END_OF_PYTHON_CODE');
#[...]
END_OF_PYTHON_CODE
```

With `py_eval` the provided python test module is injected into the perl context. The `py_study_package` and `py_bind_func` are there to couple the python context with the perl context so they can call each other out.

Plus some `sys.path` manipulation to make everything available "like magic".

The important part is:
1. This is not a python binary invocation. All of this python code is being executed by Perl directly with the Python C API.
2. The Python version used will depend on which python devel headers are available to `Inline::Python` on installation time.
3. The python code has an amount of context switches between python and perl, but I don't think they are _that_ dramatic tbh.
4. I can't confirm it 100% yet (it's been quite complex to get this working on my local), but this python binding is independent from the system's python binary since it's going through the CPython C API. This could be problematic if these sections of the code are running in an environment with multiple python versions, and there it can be confusing to know what python version is actually running.

docs/WritingTests.asciidoc
```diff
@@ -156,19 +317,27 @@ sub run() {
...
 
-In addition to Perl, it is possible to write tests in Python. Although the
-boilerplate is slightly different the function names are the same.
+=== Example Python Test Modules
+[id="testmodule_python_examples"]
+
+The following examples are short complete test modules written in Python
+implementing the interface described above.
+
+==== OpenQA Web UI Sample test
```
Martchus, line 330:
```suggestion
==== openQA web UI sample test
```

Martchus (changes requested):

docs/WritingTests.asciidoc
```diff
@@ -97,14 +252,15 @@ sub test_flags {
...
+==== Install software via `zypper`
+[id="testmodule_perl_zypper"]
 
 [caption="Example: "]
 .Console test that installs software from remote repository via zypper command
+
 [source,perl]
 ----------------------------------------------------------------------------------------------------------
-sub run() {
+sub run {
```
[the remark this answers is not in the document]
josegomezr: I don't mind bringing back the parentheses, I read the tests in distri opensuse and most of them didn't have the empty parentheses.

If there desired, I'll bring them back!

docs/WritingTests.asciidoc
```diff
@@ -156,19 +317,27 @@ sub run() {
...
-In addition to Perl, it is possible to write tests in Python. Although the
-boilerplate is slightly different the function names are the same.
+=== Example Python Test Modules
+[id="testmodule_python_examples"]
+
+The following examples are short complete test modules written in Python
+implementing the interface described above.
+
+==== OpenQA Web UI Sample test
+[id="testmodule_python_webui"]
```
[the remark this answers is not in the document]
josegomezr: They can be handy for bookmarking/hotlinking but no idea if there used or not.

docs/WritingTests.asciidoc
```diff
@@ -22,52 +22,207 @@ https://github.com/os-autoinst/openQA[official repository].
...
+-------------------------------------------------------------------
+sub post_fail_hook {
+    send_key 'ctrl-alt-f3';
+
+    assert_script_run 'openqa-cli api experimental/search q=shutdown.pm' ;    
+}
+-------------------------------------------------------------------
+
+=== Example Perl Test Modules
+[id="testmodule_perl_examples"]
```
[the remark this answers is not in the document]
josegomezr: Hm, I switched the lines for the `[id=` and the `===` and didn't affect at all the id generation. it marks the h3 with that ID regardless of the position.

Also the headers above do it like that too.

What do you mean by "following a reference"?

docs/WritingTests.asciidoc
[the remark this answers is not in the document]
foursixnine: better to have something direct to bookmark

docs/WritingTests.asciidoc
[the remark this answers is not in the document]
okurz: Please keep the parentheses. And tests in os-autoinst-distri-opensuse are often not following any consistent pattern nor modern standards. The immense value of os-autoinst-distri-opensuse is in what tests it provides, not the code quality.

okurz:
> > Please explain "FFI Layer"
> 
> `Inline::Python` is a wrapper of the python API providing perl capabilities to eval python snippets. It wraps the CPython C API.

Sorry, I meant make your code and git commit messages explain that, not explain to me on github :)

josegomezr:
> Sorry, I meant make your code and git commit messages explain that, not explain to me on github :)

My bad :sweat_smile: but hope that the latest changes (thanks to @Martchus & @kalikiana) explain better the integration.

docs/WritingTests.asciidoc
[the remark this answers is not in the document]
josegomezr: Addressed in ebe9a6f8b

docs/WritingTests.asciidoc
```diff
@@ -90,18 +246,19 @@ sub run {
     assert_screen "desktop", 300;
 }
 
-sub test_flags {
+sub test_flags() {
```
okurz, line 249:
```suggestion
sub test_flag () {
```

docs/WritingTests.asciidoc
```diff
@@ -90,18 +246,19 @@ sub run {
...
     return {fatal => 1};
 }
 
 1;
 -------------------------------------------------------------------
 
-=== Test Case Examples
-[id="testcase_examples"]
+==== Install software via `zypper`
+[id="testmodule_perl_zypper"]
```
okurz, line 257: I don't know why you changed that and, sorry, that does not make sense. Keep in mind that openQA is a generic product so not specific to SLE/openSUSE hence "zypper" is merely an example here, not something that is important to explain in the context of openQA

docs/WritingTests.asciidoc
```diff
@@ -123,10 +280,15 @@ sub run() {
     # capture a screenshot and compare with needle 'test-zypper_in'
     assert_screen 'test-zypper_in';
 }
+
+1;
```
okurz, line 284: Please leave this out. The example above is not a complete file content, just the implementation of one method

docs/WritingTests.asciidoc
```diff
@@ -156,19 +318,27 @@ sub run() {
     # make sure kate was closed
     assert_screen 'desktop';
 }
+
+1;
```
okurz, line 322: see above

docs/WritingTests.asciidoc
```diff
@@ -261,7 +431,7 @@ $testapi::distri->set_expected_serial_failures([
 .Defining serial exception capture in the test
 [source,perl]
 --------------------------------------------------------------
-sub run {
+sub run() {
```
okurz, line 434:
```suggestion
sub run () {
```

docs/WritingTests.asciidoc
```diff
@@ -22,52 +22,208 @@ https://github.com/os-autoinst/openQA[official repository].
...
+-------------------------------------------------------------------
+
+==== `run`
+Defines the actual steps to be performed during the module execution.
+
+An example usage:
+
+[source,perl]
+-------------------------------------------------------------------
+sub run() {
```
okurz, line 121: I don't understand your change here. First you explain the interface that python test modules must implement and then the perl run method is shown as example? Also the python test modules do *not* need to implement those functions, only `run` is required. For the rest the baseclass already provide implementations

docs/WritingTests.asciidoc
```diff
@@ -22,52 +22,208 @@ https://github.com/os-autoinst/openQA[official repository].
...
+Take a look at the http://open.qa/api/testapi[test API documentation] for further
+information. Note that this test API is sometimes also referred to as an
+openQA DSL, because in some contexts it can look like a domain specific
+language.
+
+=== Notes on the Python API
+[id="notes-python-api"]
+
+The Python integration that OpenQA offers through `Inline::Python` also allows
+the test modules to import other Perl modules with the usage of the `perl`
```
okurz, line 46: This whole block should be much further below. First a basic example should be featured as we had in before, only much later you should add this note

okurz (changes requested):

docs/WritingTests.asciidoc
josegomezr: Addressed in d5a519bab

docs/WritingTests.asciidoc
josegomezr: Addressed in d5a519bab

docs/WritingTests.asciidoc
josegomezr: The desire is to layout the complete interface of a test module in a language/dialect/markup that allows to indicate: signatures & return types.

The interface is not "defined" in python per se, but rather explained in type-hinted python terms (so we can get signatures & return types) on what's actually required for the test.

I could layout the base class and what can be overwritten there. But there's no "class" in the actual test module, or is it? (I still get lost in perl, please correct me if my assumption is wrong)

docs/WritingTests.asciidoc
josegomezr: Agreed, I didn't quite feel a good place for it, but wanted the notes to be somewhere close. I'll move it down the first basic explanation, maybe there makes more sense.

docs/WritingTests.asciidoc
```diff
@@ -22,52 +22,208 @@ https://github.com/os-autoinst/openQA[official repository].
 == Basic
 [id="basic"]
 
-This section explains the basic layout of openQA tests and the API available in tests.
-openQA tests are written in the *Perl* programming language. Some basic but no
-in-depth knowledge of Perl is needed. This document assumes that the reader
-is already familiar with Perl.
+This section explains the basic layout of openQA test modules and the API
+available. Test modules are written in the *Perl* programming language.
```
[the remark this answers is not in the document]
foursixnine: If so, that might also needed then to be extended to also consider python, either you live enough to become generic, or you die a specialized document.

docs/WritingTests.asciidoc
```diff
@@ -22,52 +22,208 @@ https://github.com/os-autoinst/openQA[official repository].
...
+However there is support for the *Python* programming language (through the Perl module
+`Inline::Python`).
 
-== API
+This document assumes that the reader is already familiar with Perl or Python.
+
+== Test module API
```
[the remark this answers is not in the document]
foursixnine: Lets keep it as `Test API`, seldomly we refer to it as the "test module api".

docs/WritingTests.asciidoc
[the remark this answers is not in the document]
Martchus: It is not required for bookmarking as you'll get an auto-generated ID in any case (e.g. https://open.qa/docs/#_sending_new_lines_and_continuation_characters).

docs/WritingTests.asciidoc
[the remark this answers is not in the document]
Martchus: > Also the headers above do it like that too.

We're trying to improve the documentation for years but it isn't perfect yet.

> What do you mean by "following a reference"?

Clicking on an internal cross reference within the document or following a URL with hash like https://open.qa/docs/#snapshots. Just have a look at this URL. When loading this page you'll end up below the actual heading. It would be much nicer if the heading was still visible which is achieved by moving the ID before the heading. Of course one could also just use the auto-generated ID (and only that is the ID that marks the h3 element, see e.g. https://open.qa/docs/#_using_snapshots_to_speed_up_development_of_tests).

docs/WritingTests.asciidoc
```diff
@@ -1,6 +1,6 @@
 
 [[writingtests]]
-= openQA tests developer guide
+= openQA test module developer guide
```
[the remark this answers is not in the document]
Martchus: You're right. There's no need to be over-specific here.

docs/WritingTests.asciidoc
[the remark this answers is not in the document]
Martchus: Yes, "Test API" is really more in-line with how people generally name it.

docs/WritingTests.asciidoc
```diff
@@ -22,52 +22,205 @@ https://github.com/os-autoinst/openQA[official repository].
...
+    # Make sure to begin the test in the root console.
+    select_console 'root-console';
+}
+-------------------------------------------------------------------
+
+==== `post_fail_hook`
 
-== How to write tests
+It is called only when the test execution failed, it is useful to upload log files
+or determine the state of the machine.
```
Martchus, line 133: In accordance with my other suggestion:
```suggestion
It is called after `run()` failed. It is useful to upload log files
or to determine the state of the machine.
```
Maybe also go over the other similar sections and try to streamline them.

Martchus (changes requested):

docs/WritingTests.asciidoc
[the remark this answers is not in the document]
josegomezr: Huh, I see now.

It looks like a deeper problem with the CSS Layout, normally the referred id should be the first item in the viewport. I don't quite know rn how to improve it without going way off-track with CSS changes or adding workarounds in the asciidocs.

docs/WritingTests.asciidoc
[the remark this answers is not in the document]
foursixnine: > It is not required for bookmarking as you'll get an auto-generated ID in any case (e.g. https://open.qa/docs/#_sending_new_lines_and_continuation_characters).

i'd still prefer something intentional over autogenerated but it isn't the end of the world

docs/WritingTests.asciidoc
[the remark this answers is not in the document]
josegomezr: Addressed in 835fe9e91

docs/WritingTests.asciidoc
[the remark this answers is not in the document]
Martchus: > normally the referred id should be the first item in the viewport.

It should be but that's also not the problem here. If you use the `[id="…"]` you'll get the ID on the next paragraph. So the paragraph ends up as the first item in the viewport and not the heading. Maybe the HTML generator behaves not like that in all cases (e.g. when the element after the heading is no paragraph but something else) but it makes still generally sense to put the ID before the heading.

docs/WritingTests.asciidoc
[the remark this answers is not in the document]
Martchus: Right, if it is very important that the ID stays stable then you should keep the explicit ID. I guess that's what they're for. But then still place it before the heading.

docs/WritingTests.asciidoc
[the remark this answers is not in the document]
Martchus: I guess if you wanted to change the ID of the heading element itself then you would write it like this:

```
[[usersguide]]
= openQA users guide
```

docs/WritingTests.asciidoc
okurz: > But there's no "class" in the actual test module, or is it?

There is, https://github.com/os-autoinst/os-autoinst/blob/master/basetest.pm

foursixnine: > > But there's no "class" in the actual test module, or is it?
> 
> There is, https://github.com/os-autoinst/os-autoinst/blob/master/basetest.pm

Basetest is a poor example, as it is tightly coupled with code that should be on the backend baseclass, autotest or testapi (i.e, anything that calls query_isotovideo): https://github.com/os-autoinst/os-autoinst/blob/9a99cd728b32edc918b2d170e947eecd6638656d/basetest.pm#L624

docs/WritingTests.asciidoc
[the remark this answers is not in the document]
foursixnine: > They do get used, but there's no agreement apparently.
> 
> see a sample of files having `#open.qa/docs#`
> 
> ```
> assets/javascripts/running.js
> container/helm/README.md
> docs/UsersGuide.asciidoc
> lib/OpenQA/CLI/api.pm
> lib/OpenQA/Schema/Result/Jobs.pm
> lib/OpenQA/WebAPI/Plugin/Helpers.pm
> ```
> 
> I don't have a strong opinion on keeping or removing. Adding them keeps the ID's stable (and in turn the bookmark) even if the title changes. Removing it means the links will vary with the title.

lets leave the follow ups for when we have a technical writer.

docs/WritingTests.asciidoc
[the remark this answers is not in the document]
josegomezr: > It should be but that's also not the problem here. If you use the [id="…"] you'll get the ID on the next paragraph.

Just checked this, and it doesn't render like that. IIUC this:

```asciidoc
=== Example Perl Test Modules
[id="testmodule_perl_examples"]
```

is expected to render:

```html
<!-- [...] -->
<h3>Example Perl test modules</h3>
<!-- the next element in line has the ID -->
<div id="_example_perl_test_modules"> class="paragraph">
<!-- [...] -->
```

But it actually renders:

```html
<!-- [...] -->
<h3 id="_example_perl_test_modules">Example Perl test modules</h3>
<div id="testmodule_perl_examples" class="paragraph">
<!-- [...] -->
```

HTML seems to be rendered correctly with `[id=...]` below the heading.

docs/WritingTests.asciidoc
josegomezr: After a deep dive into understanding "classes" in perl. Any sub defined in [basetest.pm][basetest.pm] could potentially be overwritten by the test module and it's allowed.

[basetest.pm][basetest.pm] is an Abstract-ish Base "Class", a concrete "class" and the test context at the same time :sweat_smile:.

The thing is now that the way these classes are used is *very* perl, they can't be called classes in python because Python classes are a completely different beast.

Right now I can't think of another way to markup the signature of what's expected for a developer to provide in a test module that what's in the docs in 995dbad4f, but I'm open to anything neutral that indicates:

* Function/Subroutine Signature
* Return Types
* What symbols are optional

[basetest.pm]: https://github.com/os-autoinst/os-autoinst/blob/master/basetest.pm

okurz: Well, the issue I see right now is that your additions state that a test module would need to implement an interface. As a software designer I understand it and it's correct but the target audience are testers and QA engineers with little to no programming experience. Hence we should be careful to use mostly plain English even under the risk of not completely technically correct statements. Formerly the text stated something along the lines of "To write a test module just put a run-function and you are good". That's what I would like to keep.

foursixnine: >  As a software designer I understand it and it's correct but the target audience are testers and QA engineers with little to no programming experience.

This assumption is incorrect (and a bit condescending), I would be ok if `interface` links to a definition on I.E Wikipedia, while we build a glossary of terms that we can refer to later. Also something missing is to say that the only required method is the run method. I'd buy (barely) Perl, but not a general statement such as that

The target audience is QA Engineers, who regardless of their programming experience have decided to look into test automation using openQA.

josegomezr: > Hence we should be careful to use mostly plain English even under the risk of not completely technically correct statements.

I think in this case we can keep both, check 8cd76802f.

Brought back the former writing focusing on `run` and elaborated below so it's clear that only a single function is needed.

okurz (approved):

Martchus:
Should we squash the commit messages? Not all would satisfy our normal criteria and I guess in this case not much is lost from squashing.

okurz:
> Should we squash the commit messages? Not all would satisfy our normal criteria and I guess in this case not much is lost from squashing.

Yes, thank you @Martchus . Fixed meanwhile so I guess you can approve now.

docs/WritingTests.asciidoc
```diff
@@ -158,17 +338,23 @@ sub run() {
...
 [caption="Example: "]
 .Test for the openQA web UI written in Python
+
 [source,python]
 --------------------------------------------------------------
 from testapi import *
 
-
-def run(self):
+def run():
```
perlpunk, line 357:
```suggestion
def run(self):
```
Otherwise we'll get
```
TypeError: run() takes 0 positional arguments but 1 was given
```

docs/WritingTests.asciidoc
```diff
@@ -180,18 +366,18 @@ def run(self):
...
+    #
+    # See the explanation in the "Notes on the Python API" section.
     perl.x11utils.x11_start_program('flatpak run com.obsproject.Studio', 'target_match', 'obsproject-wizard')
 
 def switch_to_root_console():
     send_key('ctrl-alt-f3')
 
-
-def post_fail_hook(self):
+def post_fail_hook():
```
perlpunk, line 376:
```suggestion
def post_fail_hook(self):
```

docs/WritingTests.asciidoc
```diff
@@ -180,18 +366,18 @@ def run(self):
...
     switch_to_root_console()
     assert_script_run('openqa-cli api experimental/search q=shutdown.pm')
 
-
-def test_flags(self):
+def test_flags():
```
perlpunk, line 380:
```suggestion
def test_flags(self):
```

docs/WritingTests.asciidoc
josegomezr: Changing it! 

just out of my own curiosity, can you:

```
def run(self):
  print(type(self), dir(self), self)
```

I wanna know how the self looks like :thinking:

docs/WritingTests.asciidoc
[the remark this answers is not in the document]
perlpunk: Actually, the parentheses are superfluous.
A little background: Without activating the newer `signatures` feature, parentheses after subroutine names mean [prototypes](https://perldoc.perl.org/perlsub#Prototypes). An empty prototype means, this subroutine does not accept any argument.
However, object orientation was added to Perl later, and when calling a subroutine as a method (which we do for test modules), prototypes are ignored.
So I guess it's ok to keep the parentheses, but it would be more correct to remove them :)

Martchus (approved):

PR #5335, opened 2023-10-18
RFC: Follow redirects when downloading assets for cloning

kalikiana:
Note: Untested by me as I couldn't reproduce the issue.

### DIFF
--- a/lib/OpenQA/Script/CloneJob.pm
+++ b/lib/OpenQA/Script/CloneJob.pm
@@ -141,6 +141,7 @@ sub clone_job_download_assets ($jobid, $job, $url_handler, $options) {
 
             print STDERR "downloading\n$from\nto\n$dst\n";
             my $r = $ua->mirror($from, $dst);
+            $r = $ua->mirror($r->header('Location'), $dst) if ($r->code == 308);
             unless ($r->is_success || $r->code == 304) {
                 print STDERR "$jobid failed: $file, ", $r->status_line, "\n";
                 die "Can't clone due to missing assets: ", $r->status_line, " \n"


lib/OpenQA/Script/CloneJob.pm
```diff
@@ -141,6 +141,7 @@ sub clone_job_download_assets ($jobid, $job, $url_handler, $options) {
 
             print STDERR "downloading\n$from\nto\n$dst\n";
             my $r = $ua->mirror($from, $dst);
+            $r = $ua->mirror($r->header('Location'), $dst) if ($r->code == 308);
```
Martchus, line 144: What if `Location` is `undef`? What if there's yet another redirection? We should also not follow a location from https to http.

Martchus (changes requested):

lib/OpenQA/Script/CloneJob.pm
Martchus: But generally a good idea of course.

pevik (approved):
Works as expected, thanks! (I reported the issue and also test this fix).

lib/OpenQA/Script/CloneJob.pm
pevik: @Martchus FYI in my case it was the redirect from http to https on o3.

Martchus:
> Note: Untested by me as I couldn't reproduce the issue.

Good point. I've just tried to reproduce it here as well and also wasn't able to:

```
script/openqa-clone-job --dir /hdd/openqa-devel/openqa/share/factory --show-progress --host http://localhost:9526 http://openqa.opensuse.org/tests/3655171
downloading
http://openqa.opensuse.org/tests/3655171/asset/hdd/obs-server.x86_64-2.10.51-qcow2-Build26.586.qcow2
to
/hdd/openqa-devel/openqa/share/factory/hdd/obs-server.x86_64-2.10.51-qcow2-Build26.586.qcow2
** GET http://openqa.opensuse.org/tests/3655171/asset/hdd/obs-server.x86_64-2.10.51-qcow2-Build26.586.qcow2 ==> 308 Permanent Redirect
** GET https://openqa.opensuse.org/tests/3655171/asset/hdd/obs-server.x86_64-2.10.51-qcow2-Build26.586.qcow2 ==> 302 Found
** GET https://openqa.opensuse.org/assets/hdd/obs-server.x86_64-2.10.51-qcow2-Build26.586.qcow2 ==> 200 OK (159s)
1 job has been created:
 - obs-Unstable-Appliance-x86_64-Build2.10.51-qcow2-26.586-obs_appliance@64bit-4G -> http://localhost:9526/tests/3757
```

I only used http here and as shown in the logs it follows the redirect. It also works without asset downloads.

perlpunk:
I was able to reproduce it on Leap 15.4 with LWP::UserAgent 6.31:
```
./script/openqa-clone-job --host http://localhost:9526 --skip-chained-deps --max-depth 0 --parental-inheritance --from http://openqa.opensuse.org/tests/3650601
Cloning parents of opensuse-Tumbleweed-DVD-x86_64-Build20231016-ltp_cve@64bit
downloading
http://openqa.opensuse.org/tests/3650601/asset/hdd/opensuse-Tumbleweed-x86_64-20231016-DVD@64bit-with-ltp.qcow2
to
/home/tina/openqa-toolbox/openqa/share/factory/hdd/opensuse-Tumbleweed-x86_64-20231016-DVD@64bit-with-ltp.qcow2
3650601 failed: opensuse-Tumbleweed-x86_64-20231016-DVD@64bit-with-ltp.qcow2, 308 Permanent Redirect
Can't clone due to missing assets: 308 Permanent Redirect 
```
It was fixed in 6.48: https://metacpan.org/dist/libwww-perl/changes#L98
```
6.48      2020-09-20 15:25:51Z
    - Support 308 Permanent Redirect (GH#349) (Galen Huntington)
```

lib/OpenQA/Script/CloneJob.pm
perlpunk: Maybe add a comment that this is a temporary fix until LWP 6.48 is available on all our supported platforms?

Martchus:
Then I suggest to add `libwww-perl` to our Leap 15.4 specific devel repository instead of introducing any extra handling on our side.

perlpunk:
Also 15.5, as we still have the same version there...

okurz:
> Then I suggest to add libwww-perl to our Leap 15.4
As I don't see the issue as critical how about just submitting to Leap as maintenance update and wait for that?

lib/OpenQA/Script/CloneJob.pm
pevik: @kalikiana I suppose this is expected you add before merge, right @okurz ?

okurz: yes, or anyone else who can add a commit to this PR

Martchus:
I linked the package via `osc linkpac openSUSE:Factory perl-libwww-perl devel:openQA:Leap:15.5` but it is unresolvable: https://build.opensuse.org/package/show/devel:openQA:Leap:15.5/perl-libwww-perl

> nothing provides perl(HTTP::Daemon) >= 6.12, (got version 6.01 provided by perl-HTTP-Daemon), nothing provides perl(HTTP::Request) >= 6.18, (got version 6.14 provided by perl-HTTP-Message), nothing provides perl(HTTP::Request::Common) >= 6.18, (got version 6.14 provided by perl-HTTP-Message), nothing provides perl(HTTP::Response) >= 6.18, (got version 6.14 provided by perl-HTTP-Message), nothing provides perl(HTTP::Status) >= 6.18, (got version 6.14 provided by perl-HTTP-Message), nothing provides perl(Net::HTTP) >= 6.18, (got version 6.17 provided by perl-Net-HTTP)

That are a lot of dependencies. So should we go down that path or just keep things as-is (which would make sense for a "stable" distribution like Leap).

okurz:
> I linked the package via `osc linkpac openSUSE:Factory perl-libwww-perl devel:openQA:Leap:15.5` but it is unresolvable: https://build.opensuse.org/package/show/devel:openQA:Leap:15.5/perl-libwww-perl

It is interesting to see that according to
https://software.opensuse.org/package/perl-libwww-perl?search_term=perl-libwww-perl

perl-libwww-perl is part of Tumbleweed but not in Leap although it was part of Leap until 15.2. I wonder if there is a better way to do it then without needing to pull this in again. But with such requirements and dependencies missing the original approach is obviously not straight forward. Maybe we can think of alternative approaches then?

Martchus:
We could also just move forward with the PR as an alternative approach. It only needs minor improvements to be mergable:

1. Handle the case when `Location` is `undef`
2. Handle the case of multiple redirections but with a limit to avoid getting stuck in a redirection loop
3. Disallow redirections from https to http

okurz:
So, is this really only a workaround for OSs like Leap 15.4 which goes EOL in a couple of weeks if not already? Current is Leap 15.5 but so far I did not understand which packages of Leap 15.5 this would involve

Martchus:
According to @pevik Leap 15.5 is also affected:

> It saves me adding it manually all the time because perl-libwww-perl got newer updated in Leap 15.5.

(I think "newer" is supposed to be "never".)

That's also why I invoked the `linkpac` command for Leap 15.5 and it involves the package I tried to link (and the unresolved dependencies).

okurz:
So why not fix it at the root and submit an update to Leap?

pevik:
> So why not fix it at the root and submit an update to Leap?

I finally submitted the package to Leap 15.6: https://build.opensuse.org/request/show/1181525. Feel free to comment.

lib/OpenQA/Script/CloneJob.pm
okurz: Apparently we can't achieve with a reasonable effort to update LWP in SLE/Leap hence we need to keep this workaround in place until eventually we use only more recent versions in newer OS versions


```suggestion
            # Workaround for behaviour LWP < 6.48, see https://github.com/libwww-perl/libwww-perl/pull/349
            $r = $ua->mirror($r->header('Location'), $dst) if $r->code == 308;
```

pevik:
@kalikiana OK, the update is unlikely to be accepted and even if it is accepted @okurz proposed to have this as a temporary fix until not needed. Could you please rebase? I'll retest it for you. FYI I've been using this code in my instance, because it's needed.

Martchus:
As mentioned in the chat: If you really need this, please make sure to not allow redirections from https to http and make sure it cannot be stuck in an endless loop.

lib/OpenQA/Script/CloneJob.pm
kalikiana: @pevik Maybe it's best if you take over from here? I don't even remember what issue this was addressing.

pevik:
FYI I keep using this patch (it's still needed). And I'm not sure if submitting package to openSUSE:Leap:15.6:Update would be accepted (suggested at rejected SR https://build.opensuse.org/request/show/1181525).

okurz:
I don't see a problem why the patch shouldn't be acceptable for SLE or Leap. https://build.opensuse.org/request/show/1181525 looks like you revoked it even though nobody rejected, or is it?

kalikiana:
> I don't see a problem why the patch shouldn't be acceptable for SLE or Leap. https://build.opensuse.org/request/show/1181525 looks like you revoked it even though nobody rejected, or is it?

https://build.opensuse.org/request/show/1181525#comment-1950196

lib/OpenQA/Script/CloneJob.pm
pevik: I'll do, because I still have to use this workaround.

pevik: New PR based on this one: https://github.com/os-autoinst/openQA/pull/6618.

PR #5336, opened 2023-10-19
Refactor repeat parameter with internal loop

okurz:
Additional refactoring done with @perlpunk based on
* #5331


Martchus (approved):
This is much better than first appending and then replacing the suffix.

perlpunk (changes requested):
Like I wrote in https://github.com/os-autoinst/openQA/pull/5331#discussion_r1365531892 this is doing `$repeat` requests inside the `clone_job` (now called `prepare_job`).
So I think we shouldn't do that if we can avoid it.

PR #5339, opened 2023-10-20
Remove redundant handling of test database

kalikiana:
The makefile should take advantage of what the script does without duplicating the logic.

### DIFF
--- a/t/test_postgresql
+++ b/t/test_postgresql
@@ -1,16 +1,20 @@
 #!/bin/sh -e
 
+KEEP_DB="${KEEP_DB:-0}"
 DIR=$1
 if test -z "$DIR"; then
     DIR=$(mktemp -d)
 else
     DIR=$(readlink -f "$DIR")
-fi
-if test -e "$DIR"/postmaster.pid; then
-  pg_ctl -D "$DIR" stop
-fi
-if test -d "$DIR"; then
-  rm -r "$DIR"
+    if test "$KEEP_DB" = "0"; then
+        if test -d "$DIR"; then
+            test -e "$DIR"/postmaster.pid && pg_ctl -D "$DIR" stop
+            rm -r "$DIR"
+        fi
+    else
+        echo "Re-using existing database in $DIR because KEEP_DB is set to 1."
+        exit 0
+    fi
 fi
 initdb --auth-local=peer -N "$DIR" -U "$(id -u -n)"
 (echo "listen_addresses=''"
--- a/Makefile
+++ b/Makefile
@@ -212,7 +212,7 @@ test-fullstack-unstable:
 # only works with one and Fedora's build system only works with the other
 .PHONY: test-with-database
 test-with-database:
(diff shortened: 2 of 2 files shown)


Martchus (approved):

t/test_postgresql
```diff
@@ -1,16 +1,14 @@
...
 else
+    mkdir -p $DIR
     DIR=$(readlink -f "$DIR")
-fi
-if test -e "$DIR"/postmaster.pid; then
-  pg_ctl -D "$DIR" stop
-fi
-if test -d "$DIR"; then
-  rm -r "$DIR"
+    test -e "$DIR"/postmaster.pid && pg_ctl -D "$DIR" stop
```
okurz, line 10: so if the database exists then now you would always stop it and re-initialize? That's not preserving the behaviour from before.

okurz (changes requested):

t/test_postgresql
```diff
@@ -1,16 +1,20 @@
...
-fi
-if test -d "$DIR"; then
-  rm -r "$DIR"
+    if test "$KEEP_DB" = "0"; then
+        if test -d "$DIR"; then
+            test -e "$DIR"/postmaster.pid && pg_ctl -D "$DIR" stop
+            rm -r "$DIR"
+        fi
+    else
+        echo "Database in $DIR will be re-used because KEEP_DB is set to 1."
```
okurz, line 15:
```suggestion
        echo "Reusing existing database in $DIR because KEEP_DB is set to 1."
```

would read a bit less dangerous. The sentence "database will be re-used" sounds like more bad things could happen as in. "we warned you, this will be going bad" ;)

kalikiana: Well... bad things did happen to Tina but fine by me

perlpunk:
With this PR I'm getting:
```
% make test TESTS=t/api/10-jobgroups.t KEEP_DB=1 
./t/test_postgresql /dev/shm/tpg
Re-using existing database in /dev/shm/tpg because KEEP_DB is set to 1.
make test-unit-and-integration TEST_PG="DBI:Pg:dbname=openqa_test;host=/dev/shm/tpg"
make[1]: Entering directory '/home/tina/openqa-toolbox/repos/openQA'
export GLOBIGNORE="";\
export DEVEL_COVER_DB_FORMAT=JSON;\
export PERL5OPT=" -It/lib -I/home/tina/openqa-toolbox/repos/openQA/t/lib -I/home/tina/openqa-toolbox/repos/openQA/external/os-autoinst-common/lib -MOpenQA::Test::PatchDeparse";\
RETRY=0 HOOK=./tools/delete-coverdb-folder timeout -s SIGINT -k 5 -v $((60 * 1 * (0 + 1) ))m tools/retry prove -l --trap  t/api/10-jobgroups.t
t/api/10-jobgroups.t .. DBIx::Class::Storage::DBI::catch {...} (): DBI Connection failed: DBI connect('dbname=openqa_test;host=/dev/shm/tpg','',...) failed: connection to server on socket "/dev/shm/tpg/.s.PGSQL.5432" failed: No such file or directory
        Is the server running locally and accepting connections on that socket? at /usr/lib/perl5/vendor_perl/5.26.1/DBIx/Class/Storage/DBI.pm line 1517. at /home/tina/openqa-toolbox/repos/openQA/t/api/../lib/OpenQA/Test/Database.pm line 35
```

PR #5340, opened 2023-10-20
Document `run_args` limitation on Python Modules.

josegomezr:
Arguments sent from Perl to Python can't be used (due to blessed references not properly translated to Python into usable objects).

A note of caution is added to warn against using any argument in the `run` function and updated examples.


docs/WritingTests.asciidoc
```diff
@@ -65,7 +65,7 @@ And here is an example in Python:
 -------------------------------------------------------------------
 from testapi import *
 
-def run():
+def run(*_):
```
kalikiana, line 68: What's the point of this?

josegomezr: `def fn(*_)` means ignore all parameters passed to the function.

> We could still document def run(self), as it is possible to call methods on self, e.g. it would be possible to do something like:

That can be done, but I personally wouldn't encourage it, using `self.method_defined_in_py` does a roundtrip to perl and python unnecessarily.

See the following example:

```py
from testapi import *

def run(self):
  diag("output: {}".format(self.helper()))

def helper():
  return "value"
```

it's completely valid and runs (unless `run_args` is not passed of course). But `self.helper` goes to perl to get a handler of the python package to call the helper method FROM perl and return the result in python.

Whereas:

```py
from testapi import *

def run(self):
  diag("output: {}".format(helper()))

def helper():
  return "value"
```

Would stay in python realm for the same operation.

josegomezr: They're not affected because they're called from python to perl and not the other way around. I guess that's good news? :sweat_smile:

josegomezr: > and the test inherits from it.

Inherits in the perl sense of things, to python... I'm not sure. the "self" argument that gets to the method only has four symbols:

```py
['py_bind_func', 'py_eval', 'py_study_package', 'run'],
```

that `run` is a Perl reference to the Python `run` (inception).

kalikiana: I guess I'm getting confused because on the one hand you're saying arguments can't be parsed, and on the other hand *_ would allow any arguments to be parsed.

As a user I would expect it's either no arguments ever or known restrictions e.g. self only or whatever the restrictions are.

josegomezr: > I guess I'm getting confused because on the one hand you're saying arguments can't be parsed, and on the other hand *_ would allow any arguments to be parsed.

Yeah, my bad with the terminology (it's very confusing).

Complex objects (Perl blessed References in this case) do not fly properly to python-land. They are accessible but unusable. `run_args` gets to python as: `OpenQA::Test::RunArgs=HASH(0x555970652750)` that can't be:

1. Called (`run_args()`) 
2. Scripted (`run_args['key']`)
3. Iterated (`list(run_args)`)
4. Members called (`run_args.has('abc')`).

Hope it explains better the main reason why dropping the parameters of `run`.

kalikiana: So... can we just make it `run(self)` or `run(self, *kwargs)`? My question was aiming for what works rather than what doesn't to be honest

josegomezr: I would vote for the latter: `run(self, *_):`. And encourage to keep inside python as much as possible.

docs/WritingTests.asciidoc
```diff
@@ -235,6 +235,11 @@ def run():
     # [...] omitted for brevity
 -------------------------------------------------------------------
 
+Additionally, the `run` method in Python is not able to handle any argument
```
Martchus, line 238: According to @perlpunk it is possible to run Perl-defined baseclass-functions on the first argument (`self`), see https://github.com/os-autoinst/openQA/pull/5340/files#r1366928714. So I guess this statement is not true then. The argument is at least handled in the way that it allows calling back Perl-defined functions - and being able to call those functions is likely also generally wanted.

docs/WritingTests.asciidoc
kalikiana: So what is _* for then?

josegomezr: It's being passed if it's defined, the `*_` will drop it in case is passed.

perlpunk: It's like `@` in perl signatures, a catchall throwaway basically.

kalikiana: Yes... I know what it does. I don't know why we want it here.

josegomezr: @kalikiana I'll test what you suggested in os-autoinst.

If it works as it looks like, then we can keep self and be certain that functions in Python have a stable signature.

docs/WritingTests.asciidoc
josegomezr: Agreed, Will change the wording to reflect better that `self` is the only argument accepted and `run_args` is not supported for Python.

docs/WritingTests.asciidoc
```diff
@@ -235,10 +235,12 @@ def run(*_):
     # [...] omitted for brevity
 -------------------------------------------------------------------
 
-Additionally, the `run` method in Python is not able to handle any argument
-passed. Due to how `Inline::Python` handles arguments passing between Perl <->
-Python, references to complex Perl objects do not reach python properly and
-they can't be used.
+Additionally, Python tests do not support `run_args`. An error will be present
+when a python test detects the presence of `run_args` on schedule.
```
Martchus, line 239:
```suggestion
when a Python test detects the presence of `run_args` on schedule.
```

docs/WritingTests.asciidoc
```diff
@@ -235,10 +235,12 @@ def run(*_):
...
+
+This is because of the way `Inline::Python` handles arguments passing between
```
Martchus, line 241:
```suggestion
This is because of the way `Inline::Python` handles argument passing between
```

docs/WritingTests.asciidoc
```diff
@@ -235,10 +235,12 @@ def run(*_):
...
+Perl <-> Python, references to complex Perl objects do not reach python
```
Martchus, line 242:
```suggestion
Perl <-> Python, references to complex Perl objects do not reach Python
```

Martchus (changes requested):
I'd also remove the `.` at the end of the commit message.

kalikiana (approved):

okurz (changes requested):
Code looks fine. Please observe the comment from Martchus about the git commit message subject line

Martchus (approved):

PR #5348, opened 2023-10-25
Unify perl linting rules

josegomezr:
Testing the perltidy config from os-autoinst/os-autoinst-common#30

### DIFF
--- a/.perltidyrc
+++ b/.perltidyrc
@@ -2,8 +2,9 @@
 # # See <https://github.com/houseabsolute/perl-code-tidyall/issues/84>.
 --character-encoding=none
 --no-valign
--l=120   # 120 characters per line
+-l=160
 -fbl     # don't change blank lines
+-fnl     # don't remove new lines
 -nsfs    # no spaces before semicolons
 -baao    # space after operators
 -bbao    # space before operators
--- a/lib/OpenQA/Constants.pm
+++ b/lib/OpenQA/Constants.pm
@@ -69,7 +69,7 @@ use constant WORKER_STOP_REASONS => (
 # Define error categories used alongside the reasons defined above for finer error handling where needed
 use constant {
     WORKER_EC_CACHE_FAILURE => 'cache failure',    # the cache service made problems
-    WORKER_EC_ASSET_FAILURE => 'asset failure',   # a problem occurred when handling assets, e.g. an asset was not found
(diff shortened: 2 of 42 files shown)


lib/OpenQA/CacheService/Model/Cache.pm
```diff
@@ -160,8 +159,7 @@ sub track_asset ($self, $asset) {
     eval {
         my $db = $self->sqlite->db;
         my $tx = $db->begin('exclusive');
-        my $sql = "INSERT INTO assets (filename, size, last_use) VALUES (?, 0, strftime('%s','now'))"
-          . "ON CONFLICT (filename) DO UPDATE SET pending=1";
+        my $sql = "INSERT INTO assets (filename, size, last_use) VALUES (?, 0, strftime('%s','now'))" . "ON CONFLICT (filename) DO UPDATE SET pending=1";
```
perlpunk, line 162: I guess we could live with that, but ideally we would manually remove the `" . "` for such cases. But that could be a LOT :)

josegomezr: Did a quick lookup, not so many, 8811a7ac1 tackles them.

.perltidyrc
```diff
@@ -2,8 +2,9 @@
 # # See <https://github.com/houseabsolute/perl-code-tidyall/issues/84>.
 --character-encoding=none
 --no-valign
--l=120   # 120 characters per line
+-l=160
 -fbl     # don't change blank lines
+-fnl     # don't remove new lines
```
okurz, line 7: I agree to include this line but I do not want to loosen the line-length rule. We must find a way to support project specific rule additions and exceptions.

okurz (changes requested):

.perltidyrc
kalikiana: As I remember the conversation from the other PR there's no easy way to add exceptions per project. Could 120 be used instead?

kalikiana: Okay. I guess I'm fine with 160 then if it's the only deviation.

kalikiana (approved):
Works for me.

Martchus (approved):

.perltidyrc
perlpunk: I think we can live with a .perltidyrc per project. If the common project has a config, existing and new projects *can* use this by symlinking, but they don't have to, so openQA would keep is own.
That's at least what @josegomezr agreed on when we chatted today

PR #5349, opened 2023-10-25
Fix skipping Selenium tests

AdamWill:
Not sure what changed, but on current Fedora Rawhide, just 'exit'ing like this now results in a 255 exit code and a failed test. Before:

 Install Selenium::Remote::Driver and Selenium::Chrome to run these tests
./t/ui/27-plugin_obs_rsync_status_details.t ................. ok
 Install Selenium::Remote::Driver and Selenium::Chrome to run these tests
./t/ui/28-keys_to_render_as_links.t ......................... ok All tests successful.

after:

 Install Selenium::Remote::Driver and Selenium::Chrome to run these tests
./t/ui/27-plugin_obs_rsync_status_details.t ................. ok
 Install Selenium::Remote::Driver and Selenium::Chrome to run these tests
./t/ui/28-keys_to_render_as_links.t ......................... skipped: (no reason given) ...
./t/ui/28-keys_to_render_as_links.t                       (Wstat: 65280 (exited 255) Tests: 0 Failed: 0)
  Non-zero exit status: 255

This uses `plan skip_all` to skip these tests correctly. We also have to reorder one test slightly so we skip before running any tests.

### DIFF
--- a/t/lib/OpenQA/SeleniumTest.pm
+++ b/t/lib/OpenQA/SeleniumTest.pm
@@ -352,9 +352,7 @@ sub get_mojoport () { $mojoport }
 # uncoverable subroutine
 # uncoverable statement
 sub driver_missing () {
-    diag 'Install Selenium::Remote::Driver and Selenium::Chrome to run these tests';    # uncoverable statement
-    done_testing;    # uncoverable statement
-    exit;    # uncoverable statement
+    plan skip_all => 'Install Selenium::Remote::Driver and Selenium::Chrome to run these tests'; # uncoverable statement
 }
 
 END {
--- a/t/ui/27-plugin_obs_rsync_status_details.t
+++ b/t/ui/27-plugin_obs_rsync_status_details.t
@@ -11,8 +11,9 @@ use OpenQA::Test::TimeLimit '60';
 use OpenQA::SeleniumTest;
 use OpenQA::Test::ObsRsync 'setup_obs_rsync_test';
 
+driver_missing unless check_driver_modules;
 my ($t, $tempdir) = setup_obs_rsync_test(fixtures_glob => '01-jobs.pl 03-users.pl');
-driver_missing unless my $driver = call_driver();
+my $driver = call_driver();
 $driver->find_element_by_class('navbar-brand')->click;
 $driver->find_element_by_link_text('Login')->click;
 


AdamWill:
Huh. Well, the failure is in a file touched in the PR, but it doesn't immediately make any sense to me...

    [23:19:23] t/ui/27-plugin_obs_rsync_status_details.t .. 1/? 
    #   Failed test 'sub process openqa-webapi terminated with exit code 10752'
    #   at /home/squamata/project/t/ui/../lib/OpenQA/Test/Utils.pm line 234.
    # Looks like your test exited with 10752 just after 2.

Martchus:
The more detailed logs (visible via the artifacts tab in CircleCI) are only slightly more helpful:

```
# Starting WebUI service. Port: 38945
# PID of openqa-webapi: 2843
[error] [pid:2843] Stopping openqa-webapi process because a Perl warning occurred: Failed to open /home/squamata/project/t/ui/../etc/openqa/database.ini: No such file or directory at /home/squamata/project/lib/OpenQA/Schema.pm line 40.
not ok 1 - sub process openqa-webapi terminated with exit code 10752
ok 2 - no (unexpected) warnings (via done_testing)
1..2
```

So with your change it would still attempt to start the web UI process. Maybe a combination of `plan skip_all …` and `exit` works?

But it is very strange because this test doesn't even seem to use Selenium so it shouldn't be affected by your changes at all. So I retriggered the test. Maybe the failure is a sporatic one and not related to your changes.

AdamWill:
That test *does* use selenium, the next line right after the driver_missing call is `$driver->find_element_by_class('navbar-brand')->click;`, for instance.

It's the test I had to reorder slightly in the PR - I had to make the `setup_obs_rsync_test` call come *after* the `driver_missing` call, because `setup_obs_rsync_test` uses a couple of test functions and so it's considered to have "run two tests". You can't use `skip_all` after running any tests.

Possibly the problem is with doing `setup_obs_rsync_test` after `call_driver`, though I don't know how. I guess I'll try rejigging it - I'll put the order back as it was, but have just this test *not* use `driver_missing` but instead do its own custom skipping (along the lines of how the old driver_missing worked) to account for the fact that two 'tests' have run already.

okurz (approved):

t/ui/27-plugin_obs_rsync_status_details.t
```diff
@@ -12,7 +12,15 @@ use OpenQA::SeleniumTest;
 use OpenQA::Test::ObsRsync 'setup_obs_rsync_test';
 
 my ($t, $tempdir) = setup_obs_rsync_test(fixtures_glob => '01-jobs.pl 03-users.pl');
-driver_missing unless my $driver = call_driver();
+my $driver = call_driver();
+unless ($driver) {
+    # can't use driver_missing here as it uses skip_all but we have
+    # already run two 'tests' in setup_obs_rsync_test, so we'll skip
+    # the old-fashioned way
+    diag 'Install Selenium::Remote::Driver and Selenium::Chrome to run these tests';    # uncoverable statement
```
kalikiana, line 20: Please keep this within `driver_missing` and make it configurable

t/lib/OpenQA/SeleniumTest.pm
```diff
@@ -352,9 +352,7 @@ sub get_mojoport () { $mojoport }
 # uncoverable subroutine
 # uncoverable statement
 sub driver_missing () {
-    diag 'Install Selenium::Remote::Driver and Selenium::Chrome to run these tests';    # uncoverable statement
-    done_testing;    # uncoverable statement
-    exit;    # uncoverable statement
+    plan skip_all => 'Install Selenium::Remote::Driver and Selenium::Chrome to run these tests'; # uncoverable statement
```
kalikiana, line 355:
```suggestion
    my $msg = 'Install Selenium::Remote::Driver and Selenium::Chrome to run these tests';
    plan skip_all => $msg unless $exit; # uncoverable statement
    diag $msg;    # uncoverable statement
    done_testing;    # uncoverable statement
    exit;    # uncoverable statement
```

kalikiana (changes requested):

t/lib/OpenQA/SeleniumTest.pm
AdamWill: I don't really love that idea. `skip_all` already handles printing the message and skipping; it's a one-stop shop. Doing `diag $msg` after `plan skip_all => $msg` would duplicate the message. Calling `done_testing` after `skip_all` is of questionable validity (it might even cause an error, I haven't checked); it's certainly redundant.

We could just make it a big if/else:

    sub driver_missing {
        my $exit = shift;
        my $msg = 'Install Selenium::Remote::Driver and Selenium::Chrome to run these tests';
        if ($exit) {
            diag $msg;
            done_testing;
            exit;
        }
        else {
            plan skip_all => $msg;
        }
    }

but since we only need the differing behaviour in one place, is there actually any point in factoring it out?

AdamWill: (this would, of course, be a lot simpler if `Test::More` had a `skip_remaining`, but AFAICT it doesn't. sigh. why is this test framework so weird about this stuff?)

perlpunk (approved):

t/lib/OpenQA/SeleniumTest.pm
kalikiana: Huh? You are right now duplicating the message, that was not my idea :-D I only suggested to handle it once in driver_missing so that there is no need for that

kalikiana: > We could just make it a big if/else:

Ah. Yes. I guess there's a typo in my snippet. I just meant to handle both in one place, no change in the logic.
If you're spelling out the same $msg twice it's likely to be changed inconsistently or copied if needed for another case. It could be as simple as a second unit test to avoid making this one too big which we sometimes do.

AdamWill: oh, I see. then yes, if you think not duplicating the message is important, we can do it that way. there are some pointers in my upstream issue to a way we could have driver_missing discover which path it needs to use - I'm going to look at that when I get a minute.

AdamWill:
OK, so when I came back to this after the discussion at https://github.com/Test-More/test-more/issues/919 , I came up with this idea: bail early in the troublesome test after all, but do it by using `check_driver_modules` instead of `call_driver`. I'm hoping this will be a nice still-simple-ish way to deal with the issue.

haarg's suggestion to [use Test::Needs](https://github.com/Test-More/test-more/issues/919#issuecomment-1783933575) is an interesting one - we could look at dropping `driver_missing` and `check_driver_modules` and just using `Test::Needs` in all the relevant tests instead. but that seems like a bigger change that can be a separate PR.

t/lib/OpenQA/SeleniumTest.pm
AdamWill: new version of the PR avoids these concerns, I think.

kalikiana (approved):

t/lib/OpenQA/SeleniumTest.pm
kalikiana: This looks way slicker. Thank you!

PR #5352, opened 2023-10-27
Display the correct failing host on worker boot

josegomezr:
When a worker cannot establish a websocket connection the log line reports the URL of the REST API, and not the websocket endpoint where the connection failed.

Old output:
```
[info] Registering with openQA 127.0.0.1:9526
[info] Establishing ws connection via ws://127.0.0.1:9526/api/v1/ws/1
[warn] Unable to upgrade to ws connection via http://127.0.0.1:9526/api/v1/ws/1 - trying again in 10 seconds
```

New output:
```
[info] Registering with openQA 127.0.0.1:9526
[info] Establishing ws connection via ws://127.0.0.1:9526/api/v1/ws/1
[warn] Unable to upgrade to ws connection via http://127.0.0.1:9526/api/v1/ws/1, Connection refused in http://localhost:9527/ws/1 - trying again in 10 seconds
```

Notice the addition of the network error & the final request url where the failre happened.

### DIFF
--- a/lib/OpenQA/Worker/WebUIConnection.pm
+++ b/lib/OpenQA/Worker/WebUIConnection.pm
@@ -157,6 +157,11 @@ sub _setup_websocket_connection ($self, $websocket_url = undef) {
 
                 my $error = $tx->error;
                 my $error_message = "Unable to upgrade to ws connection via $websocket_url";
+
+                if ($tx->res->error && $tx->res->error->{message}) {
+                    $error_message .= sprintf(", %s in %s", $tx->res->error->{message}, $tx->req->url);
+                }
+
                 $error_message .= ", code $error->{code}" if ($error && $error->{code});
                 $self->_set_status(failed => {error_message => $error_message});
                 return undef;


okurz:
So the next learning opportunity for you: Extend the tests to make the coverage check https://github.com/os-autoinst/openQA/pull/5352/checks?check_run_id=18141921282 happy :)

josegomezr:
> So the next learning opportunity for you: Extend the tests to make the coverage check https://github.com/os-autoinst/openQA/pull/5352/checks?check_run_id=18141921282 happy :)

I tried but got stuck on getting connection refused and not the problem at hand 😅

lib/OpenQA/Worker/WebUIConnection.pm
```diff
@@ -157,6 +157,11 @@ sub _setup_websocket_connection ($self, $websocket_url = undef) {
 
                 my $error = $tx->error;
                 my $error_message = "Unable to upgrade to ws connection via $websocket_url";
+
+                if ($tx->res->error && $tx->res->error->{message}) {
```
okurz, line 161: Please use the post-if to be consistent with other lines

lib/OpenQA/Worker/WebUIConnection.pm
```diff
@@ -157,6 +157,11 @@ sub _setup_websocket_connection ($self, $websocket_url = undef) {
...
+                    $error_message .= sprintf(", %s in %s", $tx->res->error->{message}, $tx->req->url);
+                }
+
                 $error_message .= ", code $error->{code}" if ($error && $error->{code});
```
Martchus, line 165: You can just use `$error` for simplicity and there should always be a message: 
```suggestion
                if ($error) {
                    $error_message .= ", $error->{message}";
                    $error_message .= ", code $error->{code}" if $error->{code};
                }
```
I have also removed the URL because it is already part of the error message and I would just change that other part. However, if you disagree and want the original URL still being logged that's fine as well.

lib/OpenQA/Worker/WebUIConnection.pm
```diff
@@ -157,6 +157,11 @@ sub _setup_websocket_connection ($self, $websocket_url = undef) {
 
                 my $error = $tx->error;
                 my $error_message = "Unable to upgrade to ws connection via $websocket_url";
```
Martchus, line 159: I would just alter the URL here:
```suggestion
                my $error_message = 'Unable to upgrade to ws connection via ' . $tx->req->url;
```
The original URL is still logged in the "Establishing …" line so there's nothing lost in my opinion.

Martchus (changes requested):

PR #5356, opened 2023-11-03
templates: Clarify how issue reference carry over can be prevented

okurz:
Related progress issue: https://progress.opensuse.org/issues/139055

### DIFF
--- a/templates/webapi/comments/add_comment_form_groups.html.ep
+++ b/templates/webapi/comments/add_comment_form_groups.html.ep
@@ -5,7 +5,7 @@
     </p>
     <ul>
         <li>
-            For bugreferences write <code><i>bugtracker_shortname</i>#<i>bug_nr</i></code> in a comment, e.g. <code>bsc#1234</code>.
+            For issue references write <code><i>bugtracker_shortname</i>#<i>bug_nr</i></code> in a comment, e.g. <code>bsc#1234</code> anywhere in a sentence.
         </li>
         <li>
             For generic labels use <code>label:<i>keyword</i></code> where <i>keyword</i> can be any valid character up to the next whitespace, e.g. "false_positive".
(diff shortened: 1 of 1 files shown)


Martchus (approved):

templates/webapi/comments/add_comment_form_groups.html.ep
```diff
@@ -5,7 +5,7 @@
     </p>
     <ul>
         <li>
-            For bugreferences write <code><i>bugtracker_shortname</i>#<i>bug_nr</i></code> in a comment, e.g. <code>bsc#1234</code>.
+            For issue references write <code><i>bugtracker_shortname</i>#<i>bug_nr</i></code> in a comment, e.g. <code>bsc#1234</code> anywhere in a sentence
```
kalikiana, line 8: Mixing issue and bug here seems inconsistent. If you want to make it more generic I'd rather refer to reference and tracker here, or issue or bug references even.
I would keep the period unless you're also dropping it from the next bullet point.

```suggestion
            For issue or bug references write <code><i>tracker_shortname</i>#<i>nr</i></code> in a comment, e.g. <code>bsc#1234</code> anywhere in a sentence.
```

kalikiana (changes requested):

templates/webapi/comments/add_comment_form_groups.html.ep
okurz: I don't intend to mix bug and issue references but use issue references consistently in this help text, see the other neighboring sentences

kalikiana: Well, issue and bugtracker in the same sentence is mixing to my mind... I don't care to argue, it can be fixed in another PR.

kalikiana (approved):

PR #5357, opened 2023-11-03
Add formatted bugref and carryover flag for comments

asdil12:
Ticket: https://progress.opensuse.org/issues/136244

image

### DIFF
--- a/assets/stylesheets/comments.scss
+++ b/assets/stylesheets/comments.scss
@@ -47,17 +47,37 @@
         }
     }
 
-    .openqa-label {
+    .openqa-bugref, .openqa-label, .openqa-flag {
         position: relative;
+        padding: .3em;
+        padding-left: .25em;
+        border-radius: .25rem;
+        white-space: nowrap;
+    }
+
+    .openqa-bugref {
+        background-color: #eee;
+        color: #666666;
+        border: 1px solid #ddd;
+    }
+
+    .openqa-label {
         background-color: #fdfd96;
         color: #666666;
-        padding: .3em;
-        padding-left: 1.25em;
-        border-radius: .25rem;
         border: 1px solid #ecec85;
     }
 
-    .openqa-label::before {
+    .openqa-flag {
+        background-color: #fdd596;
+        color: #666666;
+        border: 1px solid #ecd785;
+    }
+
+    .openqa-label, .openqa-flag {
+        padding-left: 1.25em;
+    }
+
+    .openqa-label::before, .openqa-flag::before {
         display: inline-block;
         font-style: normal;
         font-variant: normal;
@@ -65,7 +85,6 @@
         font-family: "ForkAwesome";
         font-weight: 900;
         font-size: .8em;
-        content: "\f02e";
         position: absolute;
         left: 0;
         top: 50%;
@@ -73,6 +92,15 @@
         width: 1.25em;
         text-align: center;
         padding-left: .25em;
+    }
+
+    .openqa-label::before {
+        content: "\f02e";
+    }
+
+
+    .openqa-flag::before {
+        content: "\f024";
     }
 
     .well {
--- a/docs/UsersGuide.asciidoc
+++ b/docs/UsersGuide.asciidoc
@@ -307,7 +307,7 @@ image::images/review_badges.png[Review badges]
 
 (To simplify, checking for false-negatives is not considered here.)
 
-=== Bug references and labels
+=== Bug references, labels and flags
 ==== Bug references
 It is possible to reference a bug by writing `<bugtracker_shortname>#<bug_nr>`
 in a comment, e.g. `bsc#1234`. It is also possible to spell out the full URL,
@@ -316,6 +316,8 @@ shortened automatically. A bug reference is rendered as link and a bug icon
 is displayed for the job in various places as shown in the figure below.
 A comment containing a bug reference will also be
 <<UsersGuide.asciidoc#carry-over,carried over>> to reduce manual review work.
+Refer to the Flags section below for other ways to trigger automated comment
+carryover.
 
 WARNING: If you want to reference a bug without making it count as a bug
 reference you need to wrap it into a label (see subsequent section), e.g.
@@ -364,7 +366,15 @@ operator permissions.
 
 NOTE: `force_result`-labels are evaluated when when a comment is
 <<UsersGuide.asciidoc#carry-over,carried over>>. However, the carry over will
-only happen when the comment *also* contains a bug reference.
+only happen when the comment *also* contains a bug reference or `flag:carryover`.
+
+==== Flags
+Currently there is only one flag for job comments supported.
+
+===== flag:carryover
+Adding `flag:carryover` to a comment, will result in this comment being
+<<UsersGuide.asciidoc#carry-over,carried over>> to a new job failing for
+the same reason, without a bugref required.
 
 === Distinguish product and test issues bugref https://github.com/os-autoinst/openQA/pull/708[gh#708]
 
@@ -563,7 +573,9 @@ prefix the status with the build number.
 Many test failures within the same scenario might be due to the same reason.
 To avoid human reviewers having to add the same bug references again and
 again, bug references are carried over from previous failures in the same
-scenario if a job fails. This idea is inspired by the
+scenario if a job fails. The same behaviour can be achieved by adding
+`flag:carryover` to a comment.
+This idea is inspired by the
 https://wiki.jenkins-ci.org/display/JENKINS/Claim+plugin[Claim plugin] for
 Jenkins.
 
--- a/lib/OpenQA/Markdown.pm
+++ b/lib/OpenQA/Markdown.pm
@@ -2,20 +2,16 @@
 # SPDX-License-Identifier: GPL-2.0-or-later
 package OpenQA::Markdown;
 use Mojo::Base -strict, -signatures;
+use Mojo::Util 'xml_escape';
+use OpenQA::App;
 
 use Exporter 'import';
 use Regexp::Common 'URI';
-use OpenQA::Utils qw(BUGREF_REGEX UNCONSTRAINED_BUGREF_REGEX LABEL_REGEX bugurl);
+use OpenQA::Utils qw(BUGREF_REGEX UNCONSTRAINED_BUGREF_REGEX LABEL_REGEX FLAG_REGEX);
 use OpenQA::Constants qw(FRAGMENT_REGEX);
 use CommonMark;
 
-our @EXPORT_OK = qw(bugref_to_markdown is_light_color markdown_to_html);
-
-sub bugref_to_markdown {
-    my $text = shift;
-    $text =~ s/${\BUGREF_REGEX}/"[$+{match}](" . bugurl($+{match}) . ')'/geio;
-    return $text;
-}
+our @EXPORT_OK = qw(bugref_to_html is_light_color markdown_to_html);
 
 sub is_light_color {
     my $color = shift;
@@ -25,20 +21,30 @@ sub is_light_color {
     return $sum > 380;
 }
 
-sub _bugref_to_html ($bugref) {
-    my $bugurl = bugurl($bugref);
-    return "<a href=\"$bugurl\">$bugref</a>";
+sub bugref_to_html ($bugref, $fancy = 0) {
+    my $app = OpenQA::App->singleton;
+    my $bugs = $app->schema->resultset('Bugs');
+    my $bug = $bugs->get_bug($bugref);
+    my $bugurl = $app->bugurl_for($bugref);
+    my $bugtitle = xml_escape($app->bugtitle_for($bugref, $bug));
+    my $bugicon = $app->bugicon_for($bugref, $bug);
+    return
+qq{<span title="$bugtitle" class="openqa-bugref"><a href="$bugurl"><i class="test-label $bugicon"></i>&nbsp;$bugref</a></span>}
+      if ($fancy);
+    return qq{<a href="$bugurl" title="$bugtitle">$bugref</a>};
 }
 
 sub _label_to_html ($label_text) {
-    $label_text =~ s/${\UNCONSTRAINED_BUGREF_REGEX}/_bugref_to_html($+{match})/ge;
+    $label_text =~ s/${\UNCONSTRAINED_BUGREF_REGEX}/bugref_to_html($+{match})/ge;
     return "<span class=\"openqa-label\">label:$label_text<\/span>";
 }
 
-sub markdown_to_html ($text) {
-    $text = bugref_to_markdown($text);
+sub _flag_to_html ($flag_text) {
+    return "<span class=\"openqa-flag\">flag:$flag_text<\/span>";
+}
 
-    # Turn all remaining URLs into links
+sub markdown_to_html ($text) {
+    # Turn all URLs into links
     $text =~ s/(?<!['"(<>])($RE{URI}${\FRAGMENT_REGEX})/<$1>/gio;
 
     # Turn references to test modules and needling steps into links
@@ -46,8 +52,14 @@ sub markdown_to_html ($text) {
 
     my $html = CommonMark->markdown_to_html($text);
 
-    # Make labels easy to highlight
+    # Turn all bugrefs into fancy bugref links
+    $html =~ s/${\BUGREF_REGEX}/bugref_to_html($+{match}, 1)/geio;
+
+    # Highlight labels
     $html =~ s/${\LABEL_REGEX}/_label_to_html($+{match})/ge;
+
+    # Highlight flags (e.g. carryover)
+    $html =~ s/${\FLAG_REGEX}/_flag_to_html($+{match})/ge;
 
     # Custom markup "{{color:#ff0000|Some text}}"
     $html =~ s/(\{\{([^|]+?)\|(.*?)\}\})/_custom($1, $2, $3)/ge;
--- a/lib/OpenQA/Schema/Result/Comments.pm
+++ b/lib/OpenQA/Schema/Result/Comments.pm
@@ -169,11 +169,11 @@ sub event_data ($self) {
     return $data;
 }
 
-sub extended_hash ($self) {
+sub extended_hash ($self, $render_markdown = 1) {
     return {
         id => $self->id,
         text => $self->text,
-        renderedMarkdown => $self->rendered_markdown->to_string,
+        renderedMarkdown => ($render_markdown) ? $self->rendered_markdown->to_string : undef,
         bugrefs => $self->bugrefs,
         created => $self->t_created->strftime("%Y-%m-%d %H:%M:%S %z"),
         updated => $self->t_updated->strftime("%Y-%m-%d %H:%M:%S %z"),
--- a/lib/OpenQA/Schema/Result/Comments.pm
+++ b/lib/OpenQA/Schema/Result/Comments.pm
@@ -5,7 +5,7 @@ package OpenQA::Schema::Result::Comments;
 use Mojo::Base 'DBIx::Class::Core', -signatures;
 
 use OpenQA::Jobs::Constants;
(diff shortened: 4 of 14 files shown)


templates/webapi/comments/add_comment_form_groups.html.ep
```diff
@@ -52,6 +52,9 @@
                 % } else {
                     <a class="help_popover fa fa-ticket" title="Add force result label" role="button"
                        data-template="label:force_result:new_result[:description_or_bugref]" onclick="insertTemplate(this)"></a>
+                    <br>
+                    <a class="help_popover fa fa-flag" title="Flag this comment for carryover without bugref (or bugref only in force-result label)" role="button"
```
okurz, line 56:
```suggestion
                    <a class="help_popover fa fa-flag" title="Flag this comment for carryover without issue reference (or issue reference only in force-result label)" role="button"
```

But I don't understand the part "(or issue reference only in force-result label)". I suggest you make it a separate complete sentence.

okurz (changes requested):
I don't know if it's a good idea to introduce "flags". To be able to answer that I suggest you start with updating the documentation to explain flags. Maybe that can show us if this too confusing or helpful.

lib/OpenQA/Markdown.pm
```diff
@@ -25,30 +19,46 @@ sub is_light_color {
     return $sum > 380;
 }
 
-sub _bugref_to_html ($bugref) {
-    my $bugurl = bugurl($bugref);
-    return "<a href=\"$bugurl\">$bugref</a>";
+sub bugref_to_html ($bugref, $fancy = 0) {
+	my $app = OpenQA::App->singleton;
```
Martchus, line 23: You'll have to run tidy.

lib/OpenQA/Markdown.pm
```diff
@@ -25,30 +19,46 @@ sub is_light_color {
...
+    my $bugs = $app->schema->resultset('Bugs');
+	my $bug = $bugs->get_bug($bugref);
+    my $bugurl = $app->bugurl_for($bugref);
+    my $bugtitle = $app->bugtitle_for($bugref, $bug);
+    my $bugicon = $app->bugicon_for($bugref, $bug);
+	if ($fancy) {
+		return qq{<span title="$bugtitle" class="openqa-bugref"><a href="$bugurl"><i class="test-label $bugicon"></i>&nbsp;$bugref</a></span>};
+	}
```
Martchus, line 31: Maybe using a suffix if would make sense. Or combine it with the next statement using the ternary operator.

templates/webapi/comments/add_comment_form_groups.html.ep
Martchus, line 56:
```suggestion
                    <a class="help_popover fa fa-flag" title="Flag this comment for carryover (despite not referencing an issue)" role="button"
```

Martchus (commented):
I guess a generic flagging mechanism makes sense. Can you update the documentation?

I suppose https://progress.opensuse.org/issues/136244 should be mentioned in the commit message.

lib/OpenQA/Markdown.pm
Martchus: Looks like a `use OpenQA::App;` is missing in that file. (And despite the update this still needs tidying. We don't use tabs here.)

lib/OpenQA/Markdown.pm
```diff
@@ -25,30 +19,46 @@ sub is_light_color {
...
     my $html = CommonMark->markdown_to_html($text);
 
-    # Make labels easy to highlight
+	# Turn all bugrefs into fancy bugref links
+    $html =~ s/${\BUGREF_REGEX}/bugref_to_html($+{match}, 1)/geio;
+
+    # Highlight labels
     $html =~ s/${\LABEL_REGEX}/_label_to_html($+{match})/ge;
 
+    # Highlight flags (eg carryover)
```
Martchus, line 59:
```suggestion
    # Highlight flags (e.g. carryover)
```

lib/OpenQA/Markdown.pm
```diff
@@ -25,30 +19,47 @@ sub is_light_color {
...
+    my $bugs = $app->schema->resultset('Bugs');
+    my $bug = $bugs->get_bug($bugref);
+    my $bugurl = $app->bugurl_for($bugref);
+    my $bugtitle = $app->bugtitle_for($bugref, $bug);
+    my $bugicon = $app->bugicon_for($bugref, $bug);
+    if ($fancy) {
+        return
+qq{<span title="$bugtitle" class="openqa-bugref"><a href="$bugurl"><i class="test-label $bugicon"></i>&nbsp;$bugref</a></span>};
+    }
+    return "<a href=\"$bugurl\" title=\"$bugtitle\">$bugref</a>";
```
Martchus: Making a database query is generally expensive. It would be better if we could avoid it in the non-fancy case. I think these should really be two distinct functions. Keep the old function as-is and introduce a new one which does the database lookup.

Martchus: It is also not so nice in general that the Markdown module now depends on database operations. Maybe this new function would also better live at another place, e.g. the database schema (to make it explicit that this is formatting based on information from the database).

Martchus (changes requested):

templates/webapi/comments/add_comment_form_groups.html.ep
asdil12: Updated with a more detailed description as suggested in Oli's comment.

lib/OpenQA/Markdown.pm
Martchus: You've updated the changes so this is no longer irrelevant; see my other comments regarding this problematic dependency structuring.

lib/OpenQA/Markdown.pm
asdil12: In both cases information from the database (e.g. the title in the non-fancy case) is shown.
I actually think that this is the right place, as the comment rendering is the logical location for this.
Also it was here before.

Martchus: Really? You even had to introduce `my $t = Test::Mojo->new('OpenQA::WebAPI');` to make the corresponding test work - so I would assume this worked before without Mojo app and database.

And the non-fancy case is really only using the bugref and bugurl variables. Nothing that would require a database lookup. The old code was also just using the free function bugurl to compute the bug URL so maybe that's how it should be kept (and all other places just keep using that simple function directly or via some helper if it needs to be).

asdil12: yes - but now it is used.
The non-fancy case uses `return qq{<a href="$bugurl" title="$bugtitle">$bugref</a>};` - the title is used and comes from the database.
The redundant bugurl function was removed for good - we really shouldn't duplicate that.

lib/OpenQA/Schema/Result/Comments.pm
```diff
@@ -107,6 +107,17 @@ sub label ($self) {
     return find_labels($self->text)->[0];
 }
 
+=head2 flags
+
+Returns flag values if C<$self> is has flags, e.g. 'flag:carryover flag:foobar' returns a hashref with the keys 'carryover' and 'foobar'
```
kalikiana, line 112:
```suggestion
Returns flag values if C<$self> has flags, e.g. 'flag:carryover flag:foobar' returns a hashref with the keys 'carryover' and 'foobar'
```

lib/OpenQA/Utils.pm
```diff
@@ -429,6 +432,13 @@ sub find_labels {
     return \@labels;
 }
 
+sub find_flags {
+    my $text = shift // '';
```
kalikiana, line 436: Just wondering if you can use a signature here? If not ignore me. I don't remember if this would need more changes here.

```suggestion
sub find_flags($text) {
```

lib/OpenQA/Markdown.pm
kalikiana: Just to chime in purely from a user's perspective. I find it invaluable to see the title of a ticket in Redmine, and I often fix links that aren't linked correctly because opening those tickets takes a lot more time.

perlpunk: Note that this is also called when retrieving the comments of a job via the API, e.g.
```
% openqa-cli api --o3  jobs/3702970/comments | jq
[
  {
    "bugrefs": [],
    "created": "2023-11-04 23:56:34 +0000",
    "id": 508341,
    "renderedMarkdown": "<p>Automatic investigation jobs...
...
```
osc, qem-bot and our tooling are requesting this route, so a new database query should be carefully considered.
I agree though that seeing the issue title is really nice.
Maybe we can change the api route to not render the markdown at all, because I believe in most cases they are not interested in that. For our tooling I'm sure, we can check for the others. It could be done via an extra query string parameter `renderHTML=1` if someone is really interested to get the HTML.

t/ui/14-dashboard.t
```diff
@@ -72,7 +72,7 @@ log_debug('build_url: ' . $build_url);
 is(scalar @{$driver->find_elements('.h4', 'css')}, 5, 'number of builds for opensuse');
 is(
     $driver->find_element_by_id('group_description')->get_text(),
-    "Test description\nwith bugref bsc#1234",
+    "Test description\nwith bugref  bsc#1234",    # second space comes nbsp used in bugref span
```
Martchus, line 75:
```suggestion
    "Test description\nwith bugref  bsc#1234",    # second space comes from using non-breaking space in bugref span
```

lib/OpenQA/Utils.pm
okurz: please use signatures here

assets/stylesheets/comments.scss
```diff
@@ -47,6 +47,16 @@
         }
     }
 
+    .openqa-bugref {
+        position: relative;
+        background-color: #eee;
+        color: #666666;
```
okurz, line 53: this looks like some duplication. We already had colors in before. So why should we need to define those settings here again?

okurz (changes requested):

assets/stylesheets/comments.scss
kalikiana: We don't currently have styling for bugrefs

assets/stylesheets/comments.scss
```diff
@@ -75,6 +85,34 @@
...
 
+    .openqa-flag {
+        position: relative;
+        background-color: #fdd596;
+        color: #666666;
+        padding: .3em;
+        padding-left: 1.25em;
+        border-radius: .25rem;
+        border: 1px solid #ecd785;
+    }
```
Martchus, line 96: 4 of these properties are identical to `.openqa-bugref` and `.openqa-label`. Maybe it makes sense to not repeat them and put the common properties into a `.openqa-bugref, .openqa-label, .openqa-flag` block.

assets/stylesheets/comments.scss
```diff
@@ -75,6 +85,34 @@
...
+        font-size: .8em;
+        content: "\f024";
+        position: absolute;
+        left: 0;
+        top: 50%;
+        transform: translateY(-50%);
+        width: 1.25em;
+        text-align: center;
+        padding-left: .25em;
+    }
```
Martchus, line 114: This looks identical to `.openqa-label::before` except for `content` so I also suggest you make a common block for them instead of C&P.

Martchus (changes requested):
Looks generally good. I'd only avoid repetition in the CSS code.

assets/stylesheets/comments.scss
okurz: I don't see how this was resolved. Can you clarify?

asdil12: @kalikiana's comment explained it.

Martchus (changes requested):
I *guess* the test changes are good. I only have a few remarks and also see the comments from my previous review.

assets/stylesheets/comments.scss
asdil12: done

perlpunk:
Was my comment https://github.com/os-autoinst/openQA/pull/5357#discussion_r1383228902 considered?

t/17-labels_carry_over.t
```diff
@@ -73,56 +78,72 @@ subtest '"happy path": failed->failed carries over last issue reference' => sub
...
 subtest 'passed->failed does not carry over old labels' => sub {
-    my $res = restart_with_result($job, 'failed');
-    $old_job = $job;
-    $job = $res->{result}[0]->{$job};
-    my @comments_new = @{comments($res->{test_url}[0]->{$old_job})};
+    $t->post_ok("/api/v1/jobs/$old_job/comments", $auth => form => {text => 'bsc#1234'})->status_is(200);
+    $t->post_ok("/api/v1/jobs/$old_job/set_done", $auth => form => {result => 'passed'})->status_is(200);
+    foreach my $m ($schema->resultset('Jobs')->find($old_job)->modules->all) {
+        $m->update({result => 'passed'});
+    }
```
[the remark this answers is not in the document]
Martchus: It doesn't look like you have applied my suggestion but also not provided an explanation why.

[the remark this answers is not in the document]
asdil12: Sorry - forgot to add the files to the commit.

lib/OpenQA/Markdown.pm
okurz: As discussed we should change to *not* deliver "renderedMarkdown" by default but allow with a flag, e.g. `openqa-cli api --o3  jobs/3702970/comments rendered_markdown=1`

okurz:
> Was my comment [#5357 (comment)](https://github.com/os-autoinst/openQA/pull/5357#discussion_r1383228902) considered?

done that now with https://github.com/os-autoinst/openQA/pull/5357#discussion_r1395434009

t/16-markdown.t
```diff
@@ -3,12 +3,19 @@
...
 
 use FindBin;
 use lib "$FindBin::Bin/lib", "$FindBin::Bin/../external/os-autoinst-common/lib";
+use OpenQA::Test::Case;
 use OpenQA::Test::TimeLimit '10';
-use OpenQA::Markdown qw(bugref_to_markdown is_light_color markdown_to_html);
+use OpenQA::Markdown qw(bugref_to_html is_light_color markdown_to_html);
+
+my $test_case = OpenQA::Test::Case->new;
+my $schema = $test_case->init_data(fixtures_glob => '01-jobs.pl 03-users.pl 05-job_modules.pl');
```
okurz, line 16: We did not need fixtures in this test in before. I am worried about the additional requirement and performance impact. What is the impact on runtime before and after this change?

t/16-markdown.t
```diff
@@ -128,33 +136,57 @@ subtest 'unsafe HTML filtered out' => sub {
...
-    is bugref_to_markdown('gfs#flatpak/fedora-flatpaks#26'),
-      '[gfs#flatpak/fedora-flatpaks#26](https://gitlab.com/fedora/sigs/flatpak/fedora-flatpaks/issues/26)',
+    is bugref_to_html('fdo#9876'),
+      '<a href="https://bugs.freedesktop.org/show_bug.cgi?id=9876" title="Bug referenced: fdo#9876">fdo#9876</a>',
       'right markdown';
-    is bugref_to_markdown("boo#9876\n\ntest boo#211\n"),
-      "[boo#9876](https://bugzilla.opensuse.org/show_bug.cgi?id=9876)\n\n"
-      . "test [boo#211](https://bugzilla.opensuse.org/show_bug.cgi?id=211)\n",
+    is bugref_to_html('jsc#9876'),
+      '<a href="https://jira.suse.de/browse/9876" title="Bug referenced: jsc#9876">jsc#9876</a>', 'right markdown';
```
okurz, line 170: please use unique test description strings explicit about the individual checks you done.

templates/webapi/comments/add_comment_form_groups.html.ep
```diff
@@ -52,6 +52,9 @@
                 % } else {
                     <a class="help_popover fa fa-ticket" title="Add force result label" role="button"
                        data-template="label:force_result:new_result[:description_or_bugref]" onclick="insertTemplate(this)"></a>
+                    <br>
+                    <a class="help_popover fa fa-flag" title="Flag this comment for carryover even without issue reference. This is useful for comments containing only a force-result label that won't trigger automatic carryover despite an issue reference being a part of the label." role="button"
```
okurz, line 56: Please also include a corresponding entry in our documentation

okurz (changes requested):

t/16-markdown.t
asdil12: Removed that line as it was not needed.

okurz: perfect :)

lib/OpenQA/Markdown.pm
asdil12: done

lib/OpenQA/WebAPI/Controller/API/V1/Comment.pm
```diff
@@ -82,7 +82,8 @@ text, date of update and the user name that created the comment.
 sub list ($self) {
     my $comments = $self->_comments();
     return unless $comments;
-    $self->render(json => [map { $_->extended_hash } $comments->all]);
+    my $render_markdown = $self->param('render_markdown') // 0;
+    $self->render(json => [map { $_->extended_hash($render_markdown) } $comments->all]);
```
okurz, line 86: Please add explanation for the parameter in the POD string above starting in line 74. That string is rendered into API help, e.g. see https://openqa.opensuse.org/not_found

okurz (changes requested):

okurz (changes requested):

docs/UsersGuide.asciidoc
```diff
@@ -307,7 +307,7 @@ image::images/review_badges.png[Review badges]
 
 (To simplify, checking for false-negatives is not considered here.)
 
-=== Bug references and labels
+=== Bug references, flags and labels
```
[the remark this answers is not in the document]
asdil12: I actually put the flags next to the bugrefs, as carryover flags have the same effect as bugrefs.

lib/OpenQA/WebAPI/Controller/API/V1/Comment.pm
```diff
@@ -72,17 +72,21 @@ sub _comments ($self) {
 =item list()
 
 Returns a list of comments for a job or a job group given its id. For each comment the
-list includes its bug references, date of creation, comment id, rendered markdown text,
+list includes its bug references, date of creation, comment id,
 text, date of update and the user name that created the comment.
 
+The optional parameter "render_markdown" if set to "1" will result in the
+rendered markdown text to be included for each comment.
+
```
Martchus, line 80: I guess that could be expressed much simpler:
```suggestion
Add the "render_markdown=1" parameter to include the rendered markdown text
for each comment.

```

lib/OpenQA/WebAPI/Controller/API/V1/Comment.pm
```diff
@@ -72,17 +72,21 @@ sub _comments ($self) {
...
 =back
 
 =cut
 
 sub list ($self) {
     my $comments = $self->_comments();
     return unless $comments;
-    $self->render(json => [map { $_->extended_hash } $comments->all]);
+    my $render_markdown = $self->param('render_markdown') // 0;
+    $self->render(json => [map { $_->extended_hash($render_markdown) } $comments->all]);
```
Martchus, line 89: Should we do the same in `sub text` for consistency?

asdil12: As this means only selecting a single comment I left it out, as the performance impact is not that problematic there.

Martchus (approved):

docs/UsersGuide.asciidoc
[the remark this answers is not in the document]
asdil12: but I updated it anyway.

okurz (approved):

PR #5358, opened 2023-11-07
Add headers to search results

perlpunk:
Now one can easily jump to the results by type and see the individual number of results

Slightly related issue: https://progress.opensuse.org/issues/137243

Screenshot:
openqa-searchresults

### DIFF
--- a/assets/javascripts/openqa.js
+++ b/assets/javascripts/openqa.js
@@ -290,27 +290,44 @@ function renderSearchResults(query, url) {
     }
     spinner.style.display = 'none';
     var heading = document.getElementById('results-heading');
-    heading.appendChild(document.createTextNode(': ' + json.data.length + ' matches found'));
+    heading.appendChild(document.createTextNode(': ' + json.data.total_count + ' matches found'));
     var results = document.createElement('div');
     results.id = 'results';
     results.className = 'list-group';
-    json.data.forEach(function (value, index) {
-      var item = document.createElement('div');
-      item.className = 'list-group-item';
-      var header = document.createElement('div');
-      header.className = 'd-flex w-100 justify-content-between';
-      var title = document.createElement('h5');
-      title.className = 'occurrence mb-1';
-      title.appendChild(document.createTextNode(value.occurrence));
-      header.appendChild(title);
-      item.appendChild(header);
-      if (value.contents) {
-        var contents = document.createElement('pre');
-        contents.className = 'contents mb-1';
-        contents.appendChild(document.createTextNode(value.contents));
-        item.appendChild(contents);
+    const types = {code: 'Test modules', modules: 'Job modules', templates: 'Job Templates'};
+
+    Object.keys(types).forEach(function (searchtype) {
+      var searchresults = json.data.results[searchtype];
+      if (searchresults.length > 0) {
+        const item = document.createElement('div');
+        item.className = 'list-group-item';
+        const header = document.createElement('h3');
+        item.appendChild(header);
+        header.id = searchtype;
+        const bold = document.createElement('strong');
+        const textnode = document.createTextNode(types[searchtype] + ': ' + searchresults.length);
+        bold.appendChild(textnode);
+        header.appendChild(bold);
+        results.append(item);
       }
-      results.append(item);
+      searchresults.forEach(function (value, index) {
+        const item = document.createElement('div');
+        item.className = 'list-group-item';
+        const header = document.createElement('div');
+        header.className = 'd-flex w-100 justify-content-between';
+        const title = document.createElement('h5');
+        title.className = 'occurrence mb-1';
+        title.appendChild(document.createTextNode(value.occurrence));
+        header.appendChild(title);
+        item.appendChild(header);
+        if (value.contents) {
+          const contents = document.createElement('pre');
+          contents.className = 'contents mb-1';
+          contents.appendChild(document.createTextNode(value.contents));
+          item.appendChild(contents);
+        }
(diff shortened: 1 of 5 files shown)


perlpunk:
I know that the headers don't look very nice yet, but at least the information is there, and the data structure returned makes it easier now to deal with different kinds of results, so I would leave it to other pull requests to make it look nicer. I just wanted this for a long time and now did it :)

kalikiana (approved):
> I know that the headers don't look very nice yet, but at least the information is there, and the data structure returned makes it easier now to deal with different kinds of results, so I would leave it to other pull requests to make it look nicer. I just wanted this for a long time and now did it :)

Awesome. Simple but effective.

I agree styling doesn't need to be added here, and I like that you're going for the smallest step here. Not to mention this will also make it easier to add more types when we don't need to solve the whole filtering UX thing in one go.

lib/OpenQA/WebAPI/Controller/API/V1/Search.pm
```diff
@@ -176,22 +176,27 @@ sub query {
...
-    $cap -= scalar @{$job_module_results};
-    push @results, @{$job_module_results};
-    return $self->render(json => {data => \@results}) unless $cap > 0;
+    $json{total_count} += @{$job_module_results};
+    $cap -= @{$job_module_results};
+    $results{modules} = $job_module_results;
+    return $self->render(json => {data => \%json}) unless $cap > 0;
 
-    push @results, @{$self->_search_job_templates($keywords, $cap)};
+    my $job_template_resuts = $self->_search_job_templates($keywords, $cap);
```
okurz, line 195: Typo resuts

okurz (changes requested):

lib/OpenQA/WebAPI/Controller/API/V1/Search.pm
perlpunk: fixed

mimi1vx (approved):

assets/javascripts/openqa.js
```diff
@@ -290,28 +290,45 @@ function renderSearchResults(query, url) {
...
+        bold.appendChild(textnode);
+
+        header.className = 'list-group-item';
+        header.appendChild(bold);
+        results.append(header);
       }
-      results.append(item);
-    });
+      searchresults.forEach(function (value, index) {
+        var item = document.createElement('div');
```
kraih, line 314: Can't many of these `var` uses be replaced with `const`?

assets/javascripts/openqa.js
```diff
@@ -290,28 +290,45 @@ function renderSearchResults(query, url) {
...
-      header.appendChild(title);
-      item.appendChild(header);
-      if (value.contents) {
-        var contents = document.createElement('pre');
-        contents.className = 'contents mb-1';
-        contents.appendChild(document.createTextNode(value.contents));
-        item.appendChild(contents);
+    const types = ['code', 'modules', 'templates'];
+
+    for (var i = 0; i < types.length; i++) {
```
kraih, line 299: And this `var` should be a `let`.

assets/javascripts/openqa.js
perlpunk: Ok, I replaced them, also in the existing loop

assets/javascripts/openqa.js
perlpunk: I changed it into an iterator

perlpunk:
The headers are now `h3`, easier to read now

assets/javascripts/openqa.js
kraih: I find loops cleaner looking personally. Arrow functions are now also preferable to use whenever possible, instead of the `keyword` function.

perlpunk: I changed the variable into an object and didn't know how to do it in a for loop :)

assets/javascripts/openqa.js
```diff
@@ -290,27 +290,44 @@ function renderSearchResults(query, url) {
...
-      item.appendChild(header);
-      if (value.contents) {
-        var contents = document.createElement('pre');
-        contents.className = 'contents mb-1';
-        contents.appendChild(document.createTextNode(value.contents));
-        item.appendChild(contents);
+    const types = {code: 'Test modules', modules: 'Job modules', templates: 'Job Templates'};
+
+    Object.keys(types).forEach(function (searchtype) {
+      var searchresults = json.data.results[searchtype];
```
Martchus, line 300: This is not Perl so we can do better: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/entries

Martchus (approved):

assets/javascripts/openqa.js
```diff
@@ -290,27 +290,44 @@ function renderSearchResults(query, url) {
...
-      header.appendChild(title);
-      item.appendChild(header);
-      if (value.contents) {
-        var contents = document.createElement('pre');
-        contents.className = 'contents mb-1';
-        contents.appendChild(document.createTextNode(value.contents));
-        item.appendChild(contents);
+    const types = {code: 'Test modules', modules: 'Job modules', templates: 'Job Templates'};
+
+    Object.keys(types).forEach(function (searchtype) {
```
kraih, line 299:
```js
for (const searchtype of Object.keys(types)) {
```

perlpunk:
Please create followup-PRs for any improvements. I don't have a ticket for that and just wanted it solved. The kind of for loop is just a matter of taste.

PR #5359, opened 2023-11-13
Make openqa-cli retry also on connection errors

perlpunk:
There may be temporary network problems or an overloaded webui. In those cases we also want to retry.

Issue: https://progress.opensuse.org/issues/138545

### DIFF
--- a/t/43-cli-api.t
+++ b/t/43-cli-api.t
@@ -451,6 +451,11 @@ EOF
 
     ($stdout, $stderr, @result) = capture sub { $api->run('--retries', '1', @params) };
     like $stdout, qr/failed.*retrying/, 'requests are retried on error if requested';
+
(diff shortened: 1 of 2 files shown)


t/43-cli-api.t
```diff
@@ -60,6 +60,7 @@ my @host = ('--host', $host);
 # Default options for authentication tests
 my @auth = ('--apikey', 'ARTHURKEY01', '--apisecret', 'EXCALIBUR', @host);
 
+$ENV{MOJO_CONNECT_TIMEOUT} = 1;
```
okurz, line 63: does this mean waiting 1s in tests additionally?

perlpunk: Actually we don't need it. I think because "Connection refused" will return immediately in this case. So I removed it again.

kalikiana (approved):

PR #5362, opened 2023-11-15
Update documentation about manual setup of PostgreSQL

asmorodskyi:
1. Remove mentioning of version from which PostgreSQL is supported ( this was introduced
so many years ago so no really find version which does not support it)
2. Explicitly state requirements of PostgreSQL setup

### DIFF
--- a/docs/Installing.asciidoc
+++ b/docs/Installing.asciidoc
@@ -387,15 +387,12 @@ httpsonly = 0
 [[database]]
 === Database
 
-Since version _4.5.1512500474.437cc1c7_ of openQA, PostgreSQL is used as the
-database.
-
-To configure access to the database in openQA, edit `/etc/openqa/database.ini`
-and change the settings in the `[production]` section.
-
-The `dsn` value format technically depends on the database type and is
-documented for PostgreSQL at
-https://metacpan.org/pod/DBD::Pg#DBI-Class-Methods[DBD::Pg]
+openQA uses PostgreSQL as database. By default, a database with name `openqa`
+and `geekotest` user as owner is used. An automatic setup of a freshly 
+installed PostgreSQL instance can be done using https://github.com/os-autoinst/openQA/blob/master/script/setup-db[this script].
+The database connection can be configured in `/etc/openqa/database.ini`
+(normally the `[production]` section is relevant). More info about the `dsn`
+value format can be found in the https://metacpan.org/pod/DBD::Pg#DBI-Class-Methods[DBD::Pg documentation].
 
 ==== Example for connecting to local PostgreSQL database
 


docs/Installing.asciidoc
```diff
@@ -387,15 +387,11 @@ httpsonly = 0
...
-and change the settings in the `[production]` section.
-
-The `dsn` value format technically depends on the database type and is
-documented for PostgreSQL at
-https://metacpan.org/pod/DBD::Pg#DBI-Class-Methods[DBD::Pg]
+PostgreSQL used by openQA suppose to have database with name `openqa` and
+`geekotest` user as owner of this database. Automatic setup of freshly installed
+PostgreSQL instance can be done using https://github.com/os-autoinst/openQA/blob/master/script/setup-db[this script].
+Connection to database configured in `/etc/openqa/database.ini` in `[production]`
+section. More info about `dsn` value format you can find in https://metacpan.org/pod/DBD::Pg#DBI-Class-Methods[DBD::Pg].
```
Martchus, line 394: That really depends on what you put into the database config. Additionally, you have to use more articles in English :-)

```suggestion
openQA uses PostgreSQL as database. By default, a database with name `openqa` and
`geekotest` user as owner is used. An automatic setup of a freshly installed PostgreSQL
instance can be done using https://github.com/os-autoinst/openQA/blob/master/script/setup-db[this script].
The database connection can be configured in `/etc/openqa/database.ini` (normally the `[production]`
section is relevant). More info about the `dsn` value format can be found in the
https://metacpan.org/pod/DBD::Pg#DBI-Class-Methods[DBD::Pg documentation].
```

Martchus (changes requested):

Martchus (approved):

docs/Installing.asciidoc
```diff
@@ -387,15 +387,12 @@ httpsonly = 0
...
-
-The `dsn` value format technically depends on the database type and is
-documented for PostgreSQL at
-https://metacpan.org/pod/DBD::Pg#DBI-Class-Methods[DBD::Pg]
+openQA uses PostgreSQL as database. By default, a database with name `openqa` and
+`geekotest` user as owner is used. An automatic setup of a freshly installed PostgreSQL
+instance can be done using https://github.com/os-autoinst/openQA/blob/master/script/setup-db[this script].
+The database connection can be configured in `/etc/openqa/database.ini` (normally the `[production]`
+section is relevant). More info about the `dsn` value format can be found in the
+https://metacpan.org/pod/DBD::Pg#DBI-Class-Methods[DBD::Pg documentation].
```
okurz, line 395: please wrap consistently at 80 characters per line

okurz (changes requested):

docs/Installing.asciidoc
asmorodskyi: I think `https://github.com/os-autoinst/openQA/blob/master/script/setup-db[this script]` and `https://metacpan.org/pod/DBD::Pg#DBI-Class-Methods[DBD::Pg documentation]` may break it because major part of the text will not be displayed , right ?

asmorodskyi: I updated everything else

okurz: looks like you wrapped manually? I suggest to trust editors to do that. Likely they would put the URLs in a separate line. But let's keep it like it is.

okurz (approved):

PR #5363, opened 2023-11-15
Increase MOJO_CONNECT_TIMEOUT in openqa-cli

perlpunk:
Issue: https://progress.opensuse.org/issues/138545

### DIFF
--- a/lib/OpenQA/Command.pm
+++ b/lib/OpenQA/Command.pm
@@ -119,6 +119,7 @@ sub url_for ($self, $path) {
 }
 
 sub retry_tx ($self, $client, $tx, $retries = undef, $delay = undef) {
+    $client->connect_timeout($ENV{MOJO_CONNECT_TIMEOUT} // 30);
     $delay //= $ENV{OPENQA_CLI_RETRY_SLEEP_TIME_S} // 3;
     $retries //= $ENV{OPENQA_CLI_RETRIES} // 0;
     for (;; --$retries) {


lib/OpenQA/Command.pm
```diff
@@ -119,6 +119,7 @@ sub url_for ($self, $path) {
 }
 
 sub retry_tx ($self, $client, $tx, $retries = undef, $delay = undef) {
+    local $ENV{MOJO_CONNECT_TIMEOUT} = 30;
```
Martchus, line 122: This might also be a decrease, e.g. if the user had explicitly exported `MOJO_CONNECT_TIMEOUT=60`.

```suggestion
    local $ENV{MOJO_CONNECT_TIMEOUT} = 30 unless defined $ENV{MOJO_CONNECT_TIMEOUT};
```

Or maybe just (but I'm not sure how it'll behave together with `local`):
```suggestion
    local $ENV{MOJO_CONNECT_TIMEOUT} //= 30;
```

perlpunk: done. `//=` works with `local`

lib/OpenQA/Command.pm
```diff
@@ -119,6 +119,7 @@ sub url_for ($self, $path) {
 }
 
 sub retry_tx ($self, $client, $tx, $retries = undef, $delay = undef) {
+    local $ENV{MOJO_CONNECT_TIMEOUT} //= 30;
```
kraih, line 122: That shouldn't work. It's a `Mojo::UserAgent` attribute with lazy initialization.

perlpunk: oh, right :)
fixed

Martchus (approved):

kraih (approved):

PR #5374, opened 2023-11-23
infopanel: render scenario_description as markdown

cfconrad:
I would like to have clickable links in the `JOB_DESCRIPTION`. With this, I could use the `JOB_DESCRIPTION` linking to a corresponding trigger source (e.g. a gitlab pipeline).

### DIFF
--- a/templates/webapi/test/infopanel.html.ep
+++ b/templates/webapi/test/infopanel.html.ep
@@ -178,7 +178,7 @@
                         </span>
                     </div>
                 % }
-                % if (my $scenario_description = $job->scenario_description) {
+                % if (my $scenario_description = $job->rendered_scenario_description) {
                     <div id="scenario-description">
                             %= $scenario_description
                     </div>
--- a/lib/OpenQA/Schema/Result/Jobs.pm
+++ b/lib/OpenQA/Schema/Result/Jobs.pm
@@ -18,6 +18,7 @@ use OpenQA::Utils (
 use OpenQA::App;
 use OpenQA::Jobs::Constants;
 use OpenQA::JobDependencies::Constants;
+use OpenQA::Markdown 'markdown_to_html';
 use OpenQA::Setup;
 use OpenQA::ScreenshotDeletion;
 use File::Basename qw(basename dirname);
@@ -313,6 +314,11 @@ sub scenario_description ($self) {
     return $description if defined $description;
     my $scenario = $self->scenario or return undef;
     return $scenario->description;
+}
+
+sub rendered_scenario_description ($self) {
+    return undef unless my $desc = $self->scenario_description;
+    return Mojo::ByteStream->new(markdown_to_html($desc));
 }
 
 # return 0 if we have no worker


templates/webapi/test/infopanel.html.ep
```diff
@@ -166,7 +166,7 @@
                 % }
                 % if (my $scenario_description = $job->scenario_description) {
                     <div id="scenario-description">
-                            %= $scenario_description
+                            %= Mojo::ByteStream->new(OpenQA::Markdown::markdown_to_html($scenario_description))
```
okurz, line 169: Why do we need the outer Mojo::ByteStream?

Martchus: Because otherwise the HTML code would be escaped. However, there's a more elegant way of doing that:

```suggestion
                            %== OpenQA::Markdown::markdown_to_html($scenario_description)
```

But it would also make sense to add a `rendered_scenario_description` function on the job that does this (similar to the `rendered_description` function we already have for job groups).

cfconrad: thx for the pointer I will add the `rendered_description`.

cfconrad:
Try it with:
```
openqa-cli api --host "http://openqa-mt.wicked.suse.de" -X POST jobs SCHEDULE="tests/foo/i_do_not_exists" TEST=test_infopanel_markdown JOB_DESCRIPTION='Mark: [Down](http://rocks)'
 ```
 
I think, this change might also affect the view of scenario descriptions used in Testsuites https://github.com/os-autoinst/openQA/pull/1094. But I think using Markdown here as well, is a benefit. WDYT?

**Review by okurz — APPROVED** [review]

**Review by Martchus — APPROVED** [review]

**Review by okurz — CHANGES_REQUESTED** [review]
Now only CI checks need to be fixed

**Comment by cfconrad** [comment]
> Now only CI checks need to be fixed

Are the failures related to my changes. I didn't connect 
```
Loading repository data...
Reading installed packages...
Package 'alsa-1.2.6.1' not found.
Package 'alsa-ucm-conf-1.2.6.3' not found.
Package 'alsa-utils-1.2.6' not found.
Package 'gdk-pixbuf-loader-rsvg-2.52.6' not found.
Package 'hwdata-0.360' not found.
Package 'libXvnc1-1.10.1' not found.
Package 'libatopology2-1.2.6.1' not found.
Package 'libepoxy0-1.5.9' not found.
Package 'librsvg-2-2-2.52.6' not found.
```
to my changes (https://app.circleci.com/pipelines/github/os-autoinst/openQA/12567/workflows/ed531370-3d37-4398-8814-57826dac6c12/jobs/117403)

and with https://build.opensuse.org/package/show/devel:openQA:GitHub:os-autoinst:openQA:PR-5374/openQA
```
 unresolvable: nothing provides openQA-client = 4.6.1700835293.acf9c7f, (got version 4.6.1700951825.872b397-lp155.6187.1) 
```
I'm not sure where this comes from.

Maybe you can give me a hint how to fix it?

Martchus:
None of these failures seem related. The cache creation might be fixed by ensuring your commit is based on the latest commit from the master branch. The checks on OBS also look like something we've already been fixed so the same applies here.

okurz (approved):

PR #5376, opened 2023-11-24
Fix testrefs in lists

asdil12:
image


perlpunk (approved):

okurz (commented):
removing code sounds good but I would need more context to be able to approve that. Maybe somebody else can approve nevertheless

Martchus (approved):

PR #5381, opened 2023-11-30
Pass VNC argument for developer mode dynamically when test has started

Martchus:
* The `WORKER_HOSTNAME` variable is only available once the test is running
  and its value has been passed by the worker via a status update. To use
  it in the developer mode's VNC instructions we must pass the VNC argument
  dynamically once the web socket connection has been established.
  Otherwise one had to reload the page manually.
* See https://progress.opensuse.org/issues/151576

### DIFF
--- a/assets/javascripts/running.js
+++ b/assets/javascripts/running.js
@@ -506,10 +506,12 @@ var developerMode = {
     );
   },
 
-  // returns the specified property evaluating possibly assigned functions
-  prop: function (propertyName) {
-    var prop = this[propertyName];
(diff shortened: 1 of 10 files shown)


Martchus:
This is a follow-up of https://github.com/os-autoinst/openQA/pull/5379. I have now also tested it on a tab that had been opened before the job was started. This change has gotten a bit involved because introducing span-elements for an easier dynamic replacement required CSS changes.

okurz (approved):

mimi1vx (approved):

PR #5385, opened 2023-12-05
Allow passing settings only for specific test when cloning jobs

Martchus:
This is especially useful when cloning MM clusters and specific jobs are supposed to run on specific hosts (e.g. to test connectivity between specific hosts).

Related ticket: https://progress.opensuse.org/issues/151310


lib/OpenQA/Script/CloneJob.pm
```diff
@@ -39,13 +39,13 @@ sub clone_job_apply_settings ($argv, $depth, $settings, $options) {
 
     for my $arg (@$argv) {
         # split arg into key and value
-        unless ($arg =~ /([A-Z0-9_]+\+?)=(.*)/) {
-            warn "arg '$arg' does not match";
+        unless ($arg =~ /(([a-zA-Z0-9_]+):)?([A-Z0-9_]+\+?)=(.*)/) {
```
okurz, line 42: how about

```suggestion
        unless ($arg =~ /(([[:word:]]+):)?([A-Z0-9_]+\+?)=(.*)/) {
```

see https://perldoc.perl.org/perlrecharclass

script/openqa-clone-job
```diff
@@ -68,8 +68,14 @@ this:
...
 C<--skip-chained-deps> is specified. If C<--skip-chained-deps> is specified
 published assets generated by parent jobs are downloaded to be directly used
-instead of generated. Keep in mind that by default any additionally specified
-job settings are NOT added to the also cloned parent jobs.
+instead of generated.
+
+Keep in mind that by default any additionally specified job settings are NOT
+added to the also cloned parent jobs. Specify C<--parental-inheritance> if this
+is wanted. It is also possible to specify the job a setting should be added to
+explicitly, e.g. C<create_hpc:WORKER_CLASS=special-worker>. In this example the
```
okurz, line 76: I wonder about the relation to http://open.qa/docs/#_further_examples_for_advanced_dependency_handling . They kinda do the equivalent from user perspective although the implementation is independant. So should we try to simulate that? Then likely `C<WORKER_CLASS:create_hpc=special-worker>` would be better. If you want to make it explicit that this is not the same feature then I would also use a different syntax with at least the separator missing if not more.

okurz (changes requested):

script/openqa-clone-job
Martchus: I only put it at the front because it made the handling for the `+` (for the `+=` syntax) easier. However, then I actually liked it because this way it kind of looks like a namespace for the setting (which is in-line with what the syntax does).

> I wonder about the relation to http://open.qa/docs/#_further_examples_for_advanced_dependency_handling . 

This is also really a different feature so trying to make them look the same is likely not useful. The feature this PR introduces defines a scope for the variable (that is matched against `TEST`). The other feature adds a tag to the variable (not matched against anything but used for grouping settings and to refer to them in other variables like `_START_AFTER`). But I guess it doesn't really matter; generally nothing speaks against having the extra part at the end in both cases. So I can change this PR if you like (code will likely get a bit more complicated).

> I would also use a different syntax with at least the separator missing if not more.

The separator *missing*? I guess we need some kind of separator. Do you mean a different separator should be used? I would stick to `:` because otherwise one always has to remember which separator to use in what situation even if there's really no need to explicitly distinguish.

okurz: Yeah, I meant *different* separator. But after considering what you wrote I guess we can just keep everything as is, as you have currently implemented

Martchus: I had another look at the code and I suppose when re-arranging the regex a little bit putting the additional part would not make things more complicated. Thinking about my own argument about the separator again I must also say that being consistent about the order actually had the advantage of not having to think about the order depending on the place (regardless of the exact meaning/feature). So now I'm torn what order would be better.

Martchus: I've just pushed another commit. It was really not complicating things. Not sure why I was under the impression it would yesterday. Sometimes it is best not to rush things :-)

PR #5388, opened 2023-12-06
Add job group filter options for /tests/overview

kraih:
This change introduces Text::Glob as a new dependency. Because globs are much safer for user provided input than regular expressions, which are prone to ReDoS attacks.


kraih:
This is what the new include and exclude filter options look like, both are comma separated lists of globs:
overview

templates/webapi/test/overview.html.ep
```diff
@@ -147,6 +147,27 @@
...
                     </div>
                 </div>
+                <div class="row" id="filter-job-groups">
+                    <div class="col-5">
+                        <div class="form-group">
+                            <label for="group-glob"><strong>Include job groups</strong></label>
+                            <%= help_popover('Help for the <em>Include job group</em> filter' => '
+                            <p>Shows jobs from job groups matching these globs</p>') %>
+                            <input name="group_glob" type="text" id="group-glob"
+                            placeholder="globs, comma separated, e.g. *kernel*,*Tumbleweed*"
```
kraih, line 157: The default globs are not very creative, there's probably better ones to show. 🤔

kraih:
While i've only really used the zero or more character syntax in tests, there's a bit more: https://metacpan.org/pod/Text::Glob#SYNTAX

lib/OpenQA/WebAPI/Plugin/Helpers.pm
```diff
@@ -481,6 +496,14 @@ sub _compose_job_overview_search_args ($c) {
     return (\%search_args, \@groups);
 }
 
+sub _match_group ($regexes, $group) {
+    my $name = $group->name;
+    for my $regex (@$regexes) {
+        return 1 if $name =~ $regex;
+    }
+    return 0;
```
okurz, line 504: how about https://perldoc.perl.org/List::Util#any

```suggestion
    return any { $name =~ $regex } @$regexes;
```

kraih: Loops are faster than `List::Util` functions usually, but that's not really a factor here. So i'll change it to the shorter version.

PR #5389, opened 2023-12-07
Simplify generation of documentation

josegomezr:
- Introducing two `Makefile` targets:
  * `generate-docs`: Installs all dependencies and compiles the docs.
  * `serve-docs`: Spins a local webserver to inspect the compiled docs.

~~Once this is merged it may be possible to delete ruby dependencies from the OpenQA Docker image :thinking:~~

The documentation cron does not depend on this script, it runs isolated in its own way to run directly asciidoctor.

As a first step, for developers this would suffice to easily build docs. In a follow-up the documentation publisher script could be improved.

### DIFF
--- a/tools/generate-docs
+++ b/tools/generate-docs
@@ -8,10 +8,16 @@ if [[ ! -e ./build/images ]]; then
     ln -s ../images build/images
 fi
 
-export BUNDLE_PATH=./vendor/
+asciidoctor_bin=$(command -v asciidoctor) || true
 
-# Make sure dependencies exist
-bundle install
+if [[ -z "$asciidoctor_bin" ]]; then
+    echo "asciidoc not found in PATH."
+    echo "Fallback to Rubygem's Asciidoc"
+    export BUNDLE_PATH=./vendor/
+    # Make sure dependencies exist
+    bundle install
+    asciidoctor_bin="bundle exec asciidoctor"
+fi
 
 # Run asciidoc
-bundle exec asciidoctor -o build/index.html ./index.asciidoc -d book
+$asciidoctor_bin -o build/index.html ./index.asciidoc -d book
--- a/tools/generate-docs
+++ b/tools/generate-docs
@@ -1,0 +1,13 @@
+#!/bin/bash -e
+
+# Navigate to the docs directory
+cd $(git rev-parse --show-toplevel)/docs
+
+# Make sure dependencies exist
+bundle install
+
+# Run asciidoc
+bundle exec asciidoctor -o openqa-documentation.html ./index.asciidoc -d book
+
+# Create a symlink to images (in case it doesn't exist).
+test -e build/images || ln -s ../images build/images
--- a/tools/generate-docs
+++ b/tools/generate-docs
@@ -1,13 +1,17 @@
 #!/bin/bash -e
 
 # Navigate to the docs directory
-cd $(git rev-parse --show-toplevel)/docs
+cd "$(dirname "${BASH_SOURCE[0]}")/../docs"
+mkdir -p ./build/
+
+if [[ ! -e ./build/images ]]; then
+    ln -s ../images build/images
+fi
+
+export BUNDLE_PATH=./vendor/
 
 # Make sure dependencies exist
 bundle install
 
 # Run asciidoc
-bundle exec asciidoctor -o openqa-documentation.html ./index.asciidoc -d book
-
-# Create a symlink to images (in case it doesn't exist).
-test -e build/images || ln -s ../images build/images
+bundle exec asciidoctor -o build/index.html ./index.asciidoc -d book
--- a/tools/generate-htmldoc
+++ b/tools/generate-htmldoc
@@ -1,4 +1,0 @@
-#!/bin/bash -e
-
-cre=${cre:-"podman"}
-$cre run -it --rm -v"$PWD"/:/openqa --workdir /openqa --env GEM_HOME=/home/squamata/.gem registry.opensuse.org/devel/openqa/ci/containers/base:latest tools/generate-htmldoc-in-container
--- a/.github/checklist.yml
+++ b/.github/checklist.yml
@@ -4,6 +4,6 @@ paths:
   "assets/stylesheets/**":
     - Consider providing screenshots in a PR comment to show the difference visually
   "docs/*.asciidoc":
-    - Consider generating documentation locally to verify it is rendered correctly using `tools/generate-htmldoc`
+    - Consider generating documentation locally to verify it is rendered correctly using `tools/generate-docs`
   "external/**":
     - Consider doing this change in the upstream repository directly, see external/os-autoinst-common/README.md
--- a/.gitignore
+++ b/.gitignore
@@ -46,3 +46,5 @@ node_modules
 package-lock.json
 .tidyall.d/
 openqa-documentation.html
+docs/build/
+
--- a/.gitignore
+++ b/.gitignore
@@ -47,4 +47,4 @@ package-lock.json
 .tidyall.d/
 openqa-documentation.html
 docs/build/
-
+docs/vendor/
--- a/Makefile
+++ b/Makefile
@@ -350,3 +350,11 @@ test-helm-install:
 .PHONY: update-deps
 update-deps:
 	tools/update-deps --specfile dist/rpm/openQA.spec
+
+.PHONY: generate-docs
+generate-docs:
+	tools/generate-docs
+
+.PHONY: serve-docs
+serve-docs: generate-docs
+	(cd docs/build/; python3 -m http.server)
--- a/docs/Gemfile
+++ b/docs/Gemfile
@@ -1,0 +1,4 @@
+source 'https://rubygems.org'
+
+gem 'asciidoctor'
+gem 'pygments.rb'
--- a/docs/Gemfile.lock
+++ b/docs/Gemfile.lock
@@ -1,0 +1,15 @@
+GEM
+  remote: https://rubygems.org/
+  specs:
+    asciidoctor (2.0.20)
+    pygments.rb (2.4.1)
+
+PLATFORMS
+  x86_64-linux
+
+DEPENDENCIES
+  asciidoctor
+  pygments.rb
+
+BUNDLED WITH
+   2.4.10
--- a/tools/generate-htmldoc-in-container
+++ b/tools/generate-htmldoc-in-container
@@ -1,9 +1,0 @@
-#!/bin/bash -e
-
-gem install asciidoctor pygments.rb
-
-asciidoctor_bin=$(command -v asciidoctor) || true
-[[ -n "$asciidoctor_bin" ]] || \
-    for asciidoctor_bin in "${GEM_HOME}"/bin/asciidoctor.ruby*; do :; done
-set -x
-"$asciidoctor_bin" -o openqa-documentation.html docs/index.asciidoc -d book


tools/generate-docs
```diff
@@ -0,0 +1,13 @@
+#!/bin/bash -e
+
+# Navigate to the docs directory
+cd $(git rev-parse --show-toplevel)/docs
+
+# Make sure dependencies exist
+bundle install
```
baierjan, line 7: This requires something like `ruby3.2-rubygem-bundler` to be installed in the execution environment, right?

baierjan: I guess that's why we use the container 😄 

Anyways, that dependency should be documented somewhere if we are going to drop the equivalent from the Dockerfile.

Also this is likely no longer true then: https://github.com/os-autoinst/openQA/blob/db07886e61b59ce822501c5ee439edbc21724e6d/.github/checklist.yml#L7

Martchus:
I tried it locally but it failed with:

```
Bundler::PermissionError: There was an error while trying to write to `/usr/lib64/ruby/gems/3.2.0/cache/pygments.rb-2.4.1.gem`. It is likely that you need to grant write permissions for that path.
```

I don't want to give it permissions to write there because I don't think our Makefile should write to that location. It should only write to the current working directory and maybe (if there's no other way) install packages for the current user.

Not sure whether the target for serving these files is needed. These are static web pages so you can just open `index.html` in any web browser from the file system.

baierjan:
> I don't want to give it permissions to write there because I don't think our Makefile should write to that location. It should only write to the current working directory and maybe (if there's no other way) install packages for the current user.

bundler should be probably invoked with `--path .`, or better use the non-deprecated way to configure it to put the dependencies inside some non-versioned folder inside the project?

Martchus:
The updated version is better but still doesn't seem to fully work. Now it failed with `ln: failed to create symbolic link 'build/images': No such file or directory` but `mkdir docs/build` fixed that. With that it seems to create some files but not the full documentation. The make target exists with a zero-return code but there's no `index.html` are a similar document. I'm wondering what I do wrong here.

josegomezr:
@Martchus Fixed the logic order of the script and should work properly now.

tools/generate-htmldoc
baierjan, line 1: `.github/checklist.yml` mentions this file and needs to be updated

baierjan (dismissed):

docs/Gemfile.ruby2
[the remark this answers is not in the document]
josegomezr: it's a symlink

josegomezr:
> the original script had 9 lines and one podman call

It's the same level of complexity, ~~sweeping it down the bed~~ putting it in a container does not solve the complexity in itself. I should phrase it as: Easier & Explicit generation of documentation.

>[...] I see that a half of it is just looking for the correct directory and the ruby+gem magic.

Yes, precisely for the last response in this post.

> I don't know, can we make it somehow even more elegant? 

The goal is make it understandable and easy to run. 

> can we also provide a Dockerfile with those doc dependencies? 
This can be a good idea, can I get access to `devel:openQa:ci` ? I make the image there and provide the two facets: For developers like me who run the command on the machine to use the makefile, and for container-based developers to use the container image with all provided.

> With that, we can maybe just use ruby 3?

Sure enough, but that will force people using leap to import ruby3 from `d-l-ruby` and deal with that can of worms. The "magic" is there because in opensuse packaging, ruby2 comes with bundler v1 and ruby3 comes with bundler v2. Bundler v2 will do the best to adapt and install dependencies, bundler v1 wont. Since that's the default, the script is accommodating for both scenarios.

If you have a ruby rvm/rbenv setup then you can get away with all of this for free, just adjust the appropriate version and call it a day.

What I wanna achieve is: an easier path for people to see how the docs will look in their machine. With the minimum amount of obscure techniques and the most explicit code so people in the future can review the tools and not just ignore them.

baierjan:
> > the original script had 9 lines and one podman call
> 
> It's the same level of complexity, ~sweeping it down the bed~ putting it in a container does not solve the complexity in itself. I should phrase it as: Easier & Explicit generation of documentation.

That's exactly my point, running almost identical script outside (or inside) a container doesn't make a difference (and is not simpler).

> > I don't know, can we make it somehow even more elegant?
> 
> The goal is make it understandable and easy to run.

It's still the same, right? Just call `make something-something-docs`...

> > can we also provide a Dockerfile with those doc dependencies?
> 
> This can be a good idea, can I get access to `devel:openQa:ci` ? I make the image there and provide the two facets: For developers like me who run the command on the machine to use the makefile, and for container-based developers to use the container image with all provided.

I am sure you can. Not sure if the `ci` subproject is the correct place though (I think that is meant just for images which are running as a part of CI tests). Anyway, all the dockerfiles lives in the container subdirectory in this project. 

> > With that, we can maybe just use ruby 3?
> 
> Sure enough, but that will force people using leap to import ruby3 from `d-l-ruby` and deal with that can of worms. The "magic" is there because in opensuse packaging, ruby2 comes with bundler v1 and ruby3 comes with bundler v2. Bundler v2 will do the best to adapt and install dependencies, bundler v1 wont. Since that's the default, the script is accommodating for both scenarios.
> 
> If you have a ruby rvm/rbenv setup then you can get away with all of this for free, just adjust the appropriate version and call it a day.

Afaik bundler v2 is supported since Ruby 2.3 (and you can just `gem install bundler` to get rid of bundler v1). But more importantly, the support for Ruby < 3.0 ended already in March. I would guess that supporting old Ruby versions will be problematic also within the used gems (see https://progress.opensuse.org/issues/122584 for an example). So I am not sure if it is worth the effort (and code lines) here. I know that the ruby situation in Leap is tragic (not sure if that will improve in 15.6), but that is outside of our scope. That's probably also a reason why we provide [openQA devel container](https://build.opensuse.org/package/show/devel:openQA/openQA-devel-container). But that is just my personal opinion.

> What I wanna achieve is: an easier path for people to see how the docs will look in their machine. With the minimum amount of obscure techniques and the most explicit code so people in the future can review the tools and not just ignore them.

Definitely agree on this (but would not call podman/containers obscure techniques 😄 ).

josegomezr:
@baierjan ok, then I'll drop support for ruby2 (less obscure magic :tada:).

okurz (changes requested):
I don't see how this simplifies things. In before at least in some point in time it was possible to run tools/generate-htmldoc which would run the thing in a container so quite portable. Now it's relying on a ruby runtime environment?

josegomezr:
> Now it's relying on a ruby runtime environment?

it was *always* relying on a ruby runtime.

baierjan:
> > Now it's relying on a ruby runtime environment?
> 
> it was _always_ relying on a ruby runtime.

It was relying on a ruby runtime provided inside the container (registry.opensuse.org/devel/openqa/ci/containers/base:latest)

josegomezr:
Why is throwing things off a container more favorable than explicitness? It's a fake sense portability if you ask me.

Sure it can "run", but who understands what's happening? If the source is behind 7 indirection layers it defeats the purpose of being "open", it's just as obscure as closed source would be .-.

perlpunk:
Providing a tool that runs the command in a container makes it easier for contributors that don't want to install things for all the projects they are contributing to. And it's not that uncommon to provide something like that.
I don't see why running a container would create 7 indirection layers.

Personally I run everything in a toolbox container anyway, and ruby is installed in there as we need it elsewhere too, so for me it doesn't matter that much.

tools/generate-docs
```diff
@@ -0,0 +1,38 @@
...
+
+while [ -h "$SCRIPT_PATH" ];
+do
+    cd "$( dirname -- "$SCRIPT_PATH"; )";
+    SCRIPT_PATH="$( readlink -f -- "$SCRIPT_PATH"; )";
+done
+
+cd "$( dirname -- "$SCRIPT_PATH"; )" > '/dev/null';
+SCRIPT_PATH="$( pwd; )";
+popd  > '/dev/null';
```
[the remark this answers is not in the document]
josegomezr: It tackles a lof of edgecases: `../script`, `script/abc`, symlink, symlink of a symlnk and more. I've taken it from this reply since some years:

https://stackoverflow.com/a/12197518

baierjan (dismissed):

tools/generate-docs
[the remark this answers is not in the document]
perlpunk: ok, `readlink` might not be installed everywhere. But OTOH we don't have a use case for symlinking this script here anyway, right?
I wouldn't add code for a use case that we might never have, so maybe just drop the symlink stuff and do `cd "$( dirname "$( dirname "$SCRIPT_PATH" )" )"`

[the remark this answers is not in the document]
baierjan: Please note, that in the original script, the `cd` is only used to store the real path to the script in to the variable and it is immediately followed by the `popd`.

As seen on line 19 (`cd ./docs/`), the rest of the script expects the project root path as a current working directory (not the `$SCRIPT_PATH` aka `tools/`).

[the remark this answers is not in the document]
josegomezr: Left a comment on the script, and written it so that it assumes it's being called from the makefile or the project's root.

That does make it shorter and simpler :+1:

okurz (changes requested):

tools/generate-docs
```diff
@@ -0,0 +1,27 @@
+#!/bin/bash -e
+
+# Ensure this script must be executed on the project's root.
+SCRIPT_PATH="${BASH_SOURCE[0]:-$0}";
+cd "$( dirname "$( dirname -- "$( realpath "$SCRIPT_PATH" )" )" )"
+
+# Navigate to the docs directory
+cd ./docs/
+mkdir -p ./build/
```
baierjan, line 9: Just out of curiosity, what's wrong on something like?
```suggestion
# Navigate to the docs directory
cd "$(dirname ${BASH_SOURCE[0]})/../docs"
mkdir -p ./build/
```

tools/generate-docs
```diff
@@ -0,0 +1,27 @@
+#!/bin/bash -e
+
+# Ensure this script must be executed on the project's root.
+SCRIPT_PATH="${BASH_SOURCE[0]:-$0}";
+cd "$( dirname "$( dirname -- "$( realpath "$SCRIPT_PATH" )" )" )"
```
okurz, line 5: why not just `cd "$SCRIPT_PATH/.."`?

okurz (commented):
just one minor question out of curiosity regarding https://github.com/os-autoinst/openQA/pull/5389/files#r1423621744 . Besides that everything looks fine. Tested it locally and works fine for me.

tools/generate-docs
josegomezr: I'll re-quote-link myself on this one:

https://github.com/os-autoinst/openQA/pull/5389/files#r1422496164

okurz: I am not convinced why tools/generate-docs should use a different implementation in relation to other tools just based on personal preference to make something "more explicit". Either that's the way to go for all cases or we stick to the existing way.

foursixnine: > I am not convinced why tools/generate-docs should use a different implementation in relation to other tools just based on personal preference to make something "more explicit". Either that's the way to go for all cases or we stick to the existing way.

What do you mean by different implementation in relation to other tools? Can you be more explicit (no joke intended)

> Either that's the way to go for all cases or we stick to the existing way.

Lets be explicit :) overall

okurz: I mean other uses of `BASH_SOURCE` in the scope of https://github.com/os-autoinst

josegomezr: That will mean I'll have to go over *all of os-autoinst* to enforce explicitness on bash scripts.

That's too much effort, sorry. Applied @baierjan suggestion.

foursixnine: > I mean other uses of `BASH_SOURCE` in the scope of https://github.com/os-autoinst

@okurz please add a style check of such things and/or add it to a contributing guide, it can't be that conversations and contributions are stalled for days, because of arbitrary standards or preferences (i.e the Ternay operator discussion et al)

okurz (commented):
diff looks fine, commits need reworking

tools/generate-docs
```diff
@@ -0,0 +1,23 @@
+#!/bin/bash -e
+
+# Navigate to the docs directory
+cd "$(dirname ${BASH_SOURCE[0]})/../docs"
```
perlpunk, line 4:
```suggestion
cd "$(dirname "${BASH_SOURCE[0]}")/../docs"
```

okurz (approved):

PR #5390, opened 2023-12-07
Parse error messages from scheduling failures

baierjan:
openqa-cli schedule will only fail if there is error key in the json
root. In our case, the error message is inside the failed job list. Make
openqa-cli fail on that as well.

Reference: https://progress.opensuse.org/issues/138203


Martchus:
The change looks generally good. However, it seems it broke the developer mode and the fullstack test. Both tests are very similar so that failures are likely related. Maybe they relied on being able to schedule jobs at all despite those errors?

baierjan:
And the test failures are gone...

Martchus (approved):

okurz (approved):

PR #5393, opened 2023-12-11
DO NOT MERGE: Test stability of t/ui/15-comments.t

Martchus:
https://progress.opensuse.org/issues/152077


okurz:
This couldn't reproduce the issue. Unless we see that reappearing I assume that this was a temporary fluke that needs no action from our side.

PR #5394, opened 2023-12-11
Improve instructions for setting up VM to debug MM setup manually

Martchus:
* Use a MAC address in the example that can actually be used but that also does not conflict with the range we assign in os-autoinst for SUT VMs
* Ensure that the device is up upfront and combine that with setting the MTU
* Make it clear that the tap device must not be used by a SUT-VM at the same time
* See https://progress.opensuse.org/issues/152389

### DIFF
--- a/docs/Networking.asciidoc
+++ b/docs/Networking.asciidoc
@@ -209,31 +209,37 @@ Also be sure to reboot the worker host to make sure the setup is actually persis
 ==== Start test VMs manually
 You may also start VMs manually to verify the setup.
 
-First, download a suitable image and launch a VM in the same way `os-autoinst` would do
-for MM jobs:
+First, download a suitable image and launch a VM in the same way `os-autoinst`
+would do for MM jobs:
 ```
 wget http://download.opensuse.org/tumbleweed/appliances/openSUSE-Tumbleweed-Minimal-VM.x86_64-Cloud.qcow2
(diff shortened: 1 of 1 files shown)


docs/Networking.asciidoc
```diff
@@ -214,17 +214,21 @@ for MM jobs:
...
 ```
 
-The image used here is of course just an example. Within the VM configure the network
-*like* this (you may need to adjust concrete IP addresses, subnets and interface names):
+The image used here is of course just an example. You need to make sure to assign a unique
+MAC address (e.g. by adjusting the last two figures in the example; this will not conflict with
+MAC addresses used by os-autoinst) and use a tap device not used at the same time by a SUT-VM.
+
+Within the VM configure the network *like* this (you may need to adjust concrete IP addresses,
+subnets and interface names):
```
okurz, line 226: please wrap consistently at 80 characters per line. Your editor should already cover that :)

okurz (changes requested):

docs/Networking.asciidoc
Martchus: I inferred our maximum for this document from what we use elsewhere in that document. I now wrapped everything (except code blocks) at 80 characters.

mimi1vx (approved):

okurz (approved):

PR #5397, opened 2023-12-14
Limit number of auto_clone restarts

perlpunk:
It can happen that a job consistently fails with the same error. We want to prevent an endless cloning loop here.

Issue: https://progress.opensuse.org/issues/152569

### DIFF
--- a/etc/openqa/openqa.ini
+++ b/etc/openqa/openqa.ini
@@ -78,6 +78,8 @@
 ## Causes jobs reported as incomplete by the worker to be cloned automatically when the
 ## reason matches; set to 0 to disable
 #auto_clone_regex = ^(cache failure: |terminated prematurely: |api failure: Failed to register .* 503|backend died: .*VNC.*(timeout|timed out|refused)|QEMU terminated: Failed to allocate KVM HPT of order 25.* Cannot allocate memory)
+## The maximum amount of times a job will be restarted if the auto_clone_regex matches
+#auto_clone_limit = 20
 
 ## A regex pattern that a "force_result" label description in job comments
 ## must match to be accepted. If undefined and by default no rules are applied
--- a/lib/OpenQA/Schema/Result/Jobs.pm
+++ b/lib/OpenQA/Schema/Result/Jobs.pm
@@ -1955,6 +1955,29 @@ sub descendants ($self, $limit = -1) {
     $self->{_descendants} = $sth->fetchrow_array;
 }
 
+sub incomplete_ancestors ($self, $limit = -1) {
+    my $ancestors = $self->{_incomplete_ancestors};
+    return $ancestors if defined $ancestors;
+    my $sth = $self->result_source->schema->storage->dbh->prepare(<<~'EOM');
+    with recursive orig_id as (
+      select ? as orig_id, 0 as level, ? as orig_result, ? as orig_reason
+        union all
+      select id as orig_id, orig_id.level + 1 as level, result as orig_result, reason as orig_reason
+        from jobs join orig_id on orig_id.orig_id = jobs.clone_id
+        where result = 'incomplete' and (? < 0 or level < ?))
+      select orig_id, level, orig_reason from orig_id order by level asc;
+    EOM
+    $sth->bind_param(1, $self->id, SQL_BIGINT);
+    $sth->bind_param(2, $self->result, SQL_VARCHAR);
+    $sth->bind_param(3, $self->reason, SQL_VARCHAR);
+    $sth->bind_param(4, $limit, SQL_BIGINT);
+    $sth->bind_param(5, $limit, SQL_BIGINT);
+    $sth->execute;
+    $ancestors = $sth->fetchall_arrayref;
+    shift @$ancestors;
+    $self->{_incomplete_ancestors} = $ancestors;
+}
+
 sub latest_job ($self) {
     return $self unless my $clone = $self->clone;
     return $clone->latest_job;
@@ -2065,28 +2088,12 @@ sub done ($self, %args) {
         $worker->update({job_id => undef});
     }
 
+    my %new_val = (state => DONE);
     # update result unless already known (it is already known for CANCELLED jobs)
     # update the reason if updating the result or if there is no reason yet
     my $reason = $args{reason};
-    my $result_unknown = $self->result eq NONE;
-    my $reason_unknown = !$self->reason;
-    my %new_val = (state => DONE);
     my $restart = 0;
-    $new_val{result} = $result if $result_unknown;
-    if (($result_unknown || $reason_unknown) && defined $reason) {
-        # restart incompletes when the reason matches the configured regex
-        my $auto_clone_regex = OpenQA::App->singleton->config->{global}->{auto_clone_regex};
-        $restart = 1 if $result eq INCOMPLETE && $auto_clone_regex && $reason =~ $auto_clone_regex;
-
-        # limit length of the reason
-        # note: The reason can be anything the worker picked up as useful information so better cut it at a
-        #       reasonable, human-readable length. This also avoids growing the database too big.
-        $reason = substr($reason, 0, 300) . '…' if defined $reason && length $reason > 300;
-        $new_val{reason} = $reason;
-    }
-    elsif ($reason_unknown && !defined $reason && $result eq INCOMPLETE) {
-        $new_val{reason} = 'no test modules scheduled/uploaded';
-    }
+    $self->_compute_result_and_reason(\%new_val, $result, $reason, \$restart);
     $self->update(\%new_val);
     $self->unblock;
     my %finalize_opts = (lax => 1);
@@ -2107,6 +2114,41 @@ sub done ($self, %args) {
     $self->enqueue_finalize_job_results([$carried_over], \%finalize_opts);
 
     return $new_val{result} // $self->result;
+}
+
+sub _compute_result_and_reason ($self, $new_val, $result, $reason, $restart) {
+    my $result_unknown = $self->result eq NONE;
+    my $reason_unknown = !$self->reason;
+    $new_val->{result} = $result if $result_unknown;
+    if (($result_unknown || $reason_unknown) && defined $reason) {
+        # restart incompletes when the reason matches the configured regex
+        my $append_reason = '';
+        my $auto_clone_regex = OpenQA::App->singleton->config->{global}->{auto_clone_regex};
+        if ($result eq INCOMPLETE and $auto_clone_regex and $reason =~ $auto_clone_regex) {
+            my $limit = OpenQA::App->singleton->config->{global}{auto_clone_limit};
+            my $ancestors = $self->incomplete_ancestors($limit + 1);
+            # how many of those incomplete ancestors had a reason not matching auto_clone?
+            my $unrelated = grep { $_->[2] !~ m/$auto_clone_regex/ } @$ancestors;
+            if (@$ancestors < $limit || $unrelated > 0) {
+                $append_reason = ' [Auto-restarting because reason matches /$auto_clone_regex/]';
+                $$restart = 1;
+            }
+            else {
+                $append_reason
+                  = ' [Not restarting because job has been restarted already $auto_clone_limit times and failed with /$auto_clone_regex/]';
+            }
+        }
+
+        # limit length of the reason
+        # note: The reason can be anything the worker picked up as useful information so better cut it at a
+        # reasonable, human-readable length. This also avoids growing the database too big.
+        $reason = substr($reason, 0, 300) . '…' if length $reason > 300;
+        $reason .= $append_reason;
+        $new_val->{reason} = $reason;
+    }
+    elsif ($reason_unknown && !defined $reason && $result eq INCOMPLETE) {
+        $new_val->{reason} = 'no test modules scheduled/uploaded';
+    }
 }
 
 sub cancel ($self, $result, $reason = undef) {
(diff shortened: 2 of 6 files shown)


etc/openqa/openqa.ini
```diff
@@ -78,6 +78,8 @@
 ## Causes jobs reported as incomplete by the worker to be cloned automatically when the
 ## reason matches; set to 0 to disable
 #auto_clone_regex = ^(cache failure: |terminated prematurely: |api failure: Failed to register .* 503|backend died: .*VNC.*(timeout|timed out|refused)|QEMU terminated: Failed to allocate KVM HPT of order 25.* Cannot allocate memory)
+## How many times should a job be restarted
+#auto_clone_limit = 20;
```
okurz, line 82: I don't think there should be semicolon here

okurz (changes requested):
The problem I see with this is that apparently there is no clear communication to test reviewers that tests are restarted so many times. If the auto-cloning would stop in the realistic case any test reviewers if they ever stumble across the scenario again would likely just hit the retrigger button anyway. So, how to communicate the stop of auto-cloning to test reviewers?

perlpunk:
> The problem I see with this is that apparently there is no clear communication to test reviewers that tests are restarted so many times

I don't really understand.
Is there any clear communication to test reviewers currently about when a job is auto cloned at all? And "so many times" - well, it's basically endlessly if the cloned jobs also fail.

asdil12:
Stop the clonging!

s/clonging/cloning/  ;)

etc/openqa/openqa.ini
perlpunk: semicolon removed

Martchus:
> The problem I see with this is that apparently there is no clear communication to test reviewers that tests are restarted so many times

There is no clear communication with and without this PR. We have accumulated over 50 pages of jobs in the Next & Previous tab in relevant scenarios and apparently no reviewers took notice of it. If we now only have say 10 pages this will not change anything for reviewers (except that why might not be wondering anymore why the heck openQA is endlessly restarting these jobs if they would care about this scenario anyways which they apparently don't). So I don't see how this PR makes things worse.

lib/OpenQA/Schema/Result/Jobs.pm
```diff
@@ -2065,7 +2065,11 @@ sub done ($self, %args) {
     if (($result_unknown || $reason_unknown) && defined $reason) {
         # restart incompletes when the reason matches the configured regex
         my $auto_clone_regex = OpenQA::App->singleton->config->{global}->{auto_clone_regex};
-        $restart = 1 if $result eq INCOMPLETE && $auto_clone_regex && $reason =~ $auto_clone_regex;
+        if ($result eq INCOMPLETE and $auto_clone_regex && $reason =~ $auto_clone_regex) {
```
Martchus, line 2068: Mixing `and` and `&&` in the same line of code isn't nice. The code will do what it is supposed to do but I guess a bit more consistency would be better:

```suggestion
        if ($result eq INCOMPLETE && $auto_clone_regex && $reason =~ $auto_clone_regex) {
```

Martchus (commented):
Looks generally good. Maybe we could add a comment stating that the maximum number of retries are exhausted instead of just doing nothing. (I guess that wouldn't be too much work.)

perlpunk:
> Maybe we could add a comment stating that the maximum number of retries are exhausted instead of just doing nothing. (I guess that wouldn't be too much work.)

I'm wondering how much work we should put into this, given we also have the investigation tools, that can also do retries and add comments.
Maybe this feature should be moved there instead?

Martchus (approved):

kalikiana (approved):

etc/openqa/openqa.ini
```diff
@@ -78,6 +78,8 @@
 ## Causes jobs reported as incomplete by the worker to be cloned automatically when the
 ## reason matches; set to 0 to disable
 #auto_clone_regex = ^(cache failure: |terminated prematurely: |api failure: Failed to register .* 503|backend died: .*VNC.*(timeout|timed out|refused)|QEMU terminated: Failed to allocate KVM HPT of order 25.* Cannot allocate memory)
+## How many times should a job be restarted
```
okurz, line 81: I guess that should say "maximum amount of times"?

okurz (changes requested):
We should rethink the overall approach. Obviously it's wasteful to restart jobs over and over again if nobody cares about the results. However the auto cloning triggered by the worker was always intended to trigger in the case when a need for a retry would arise that's not in the responsibility of any test maintainer like the bugs we have in the cache service or terminating worker processes.

Your question if we should move the retries to scripts makes me think that we need to clarify the original goal, maybe talking in person next week?

Instead of stopping the automatic retries I would rather think about preventing new incompletes to happen at all, here maybe with not even starting jobs in continuously failing scenarios?

etc/openqa/openqa.ini
kalikiana: `How many times` works, but it shouldn't be inverted:

```suggestion
## How many times a job should be restarted
```

lib/OpenQA/Schema/Result/Jobs.pm
```diff
@@ -2065,7 +2065,11 @@ sub done ($self, %args) {
     if (($result_unknown || $reason_unknown) && defined $reason) {
         # restart incompletes when the reason matches the configured regex
         my $auto_clone_regex = OpenQA::App->singleton->config->{global}->{auto_clone_regex};
-        $restart = 1 if $result eq INCOMPLETE && $auto_clone_regex && $reason =~ $auto_clone_regex;
+        if ($result eq INCOMPLETE and $auto_clone_regex and $reason =~ $auto_clone_regex) {
+            my $limit = OpenQA::App->singleton->config->{global}{auto_clone_limit};
+            my $ancestors = $self->ancestors($limit + 1);
+            $restart = 1 if $ancestors < $limit;
```
okurz, line 2071: please check if each ancestor is *at least* incomplete if not also matching the very same reason for all. I would prefer the latter to be sure we are looking for the very same reason.


```suggestion
            if {$ancestors < $limit) {
                $restart = 1 ;
                $reason = "$reason: auto-cloning based on server setting 'auto_clone_regex'";
            }
            else {
                $reason = "$reason: Not auto-cloning due to server setting 'auto_clone_limit'";
            }
```

okurz (changes requested):

lib/OpenQA/Schema/Result/Jobs.pm
kalikiana: Wouldn't that make the limit ineffective? Say a job is cloned 19 times and then fails with a different reason.

Maybe this should be considered for a separate change.

okurz: Well, we could just apply that additional explanation in the case of aborting the restart but IIRC @perlpunk suggested herself to make also the auto-cloning explicit.

lib/OpenQA/Schema/Result/Jobs.pm
```diff
@@ -1955,6 +1955,29 @@ sub descendants ($self, $limit = -1) {
...
+    $sth->bind_param(2, $self->result, SQL_VARCHAR);
+    $sth->bind_param(3, $self->reason, SQL_VARCHAR);
+    $sth->bind_param(4, $limit, SQL_BIGINT);
+    $sth->bind_param(5, $limit, SQL_BIGINT);
+    $sth->execute;
+    $ancestors = $sth->fetchall_arrayref;
+    shift @$ancestors;
+    $self->{_incomplete_ancestors} = $ancestors;
+}
+
```
okurz, line 1980: I must say this becomes rather complicated and the file lib/OpenQA/Schema/Result/Jobs.pm is already unbearably large so I hope somebody has good hints for refactoring. Not necessarily this method but at least somewhere in this file.

etc/openqa/openqa.ini
```diff
@@ -78,6 +78,8 @@
...
+#auto_clone_limit = 20
```
okurz, line 82: As your new approach is now not looking into the actual reason to determine if we should limit I suggest to consider a new variable name if you have a good idea. And also I suggest to increase the default value then, let's say 50?

okurz (changes requested):

etc/openqa/openqa.ini
okurz: nevermind, you are still looking for any auto_clone regex match so all good

t/api/04-jobs.t
```diff
@@ -1482,7 +1482,57 @@ subtest 'marking job as done' => sub {
...
-        $t->json_like('/job/clone_id' => qr/\d+/, 'job cloned when reason does matches configured regex');
+        $t->json_like('/job/clone_id' => qr/\d+/, 'job cloned when reason does match configured regex');
+
+        subtest 'auto_clone limits' => sub {
+            $schema->txn_begin;
+            local $t->app->config->{global}{auto_clone_limit} = 5;
+            my @jobs;
+            my $backend_reason = 'backend died: VNC Connection refused';
+            for my $i (reverse 10 .. 20) {
+                my $clone_id = $i < 20 ? $i + 1 : undef;
```
okurz, line 1493: I suggest for the sake of efficient testing to limit this to "auto_clone_limit+1" and the auto_clone_limit could be the most minimal. I guess that would be 3+1 considering you want to build a chain of "same, unrelated, same"

t/api/04-jobs.t
```diff
@@ -1482,7 +1482,57 @@ subtest 'marking job as done' => sub {
...
+            is $restart, 0, '_reason_and_result: more than 5 auto_clone incompletes, no restart';
+
+            $jobs[3]->update({reason => 'unrelated'});
+            delete $last->{_incomplete_ancestors};
+            $last_incompletes = $last->incomplete_ancestors(5);
+            is scalar @$last_incompletes, 5,
+              'incomplete_ancestors returns a row of 5 incompletes with different reasons';
+            $restart = 0;
+            %new = ();
+            $last->_reason_and_result(\%new, 'incomplete', 'backend died: VNC Connection refused', \$restart);
```
okurz, line 1522: this can use `$backend_reason` as well, right?

t/api/04-jobs.t
okurz, line 1: As this file is now beyond 1.6k lines I suggest to move new tests elsewhere or better split the test module

okurz (changes requested):

t/api/04-jobs.t
perlpunk: I moved the new tests to `t/09-job_clone.t` as it doesn't have to do with the api anymore anyway.

etc/openqa/openqa.ini

lib/OpenQA/Schema/Result/Jobs.pm
perlpunk: At least I extracted the `_reason_and_result` method out of `done`, although I'm quite unhappy with the name.

The module has some methods that don't necessarily have to live in Result::Jobs, e.g. `create_asset` which only needs the job id. But the above function is one that clearly belong there IMHO.

No idea how to refactor yet.

Martchus: Not sure how to split a package (acting as a "class") into multiple files in Perl. If that was possible we could e.g. move all code related to cloning/duplicating into a separate file. (Although I must say that I find the file size manageable and I don't think that splitting makes things much easier to find.)

lib/OpenQA/Schema/Result/Jobs.pm
```diff
@@ -2109,6 +2116,41 @@ sub done ($self, %args) {
     return $new_val{result} // $self->result;
 }
 
+sub _reason_and_result ($self, $new_val, $result, $reason, $restart) {
```
Martchus, line 2119:
```suggestion
sub _compute_result_and_reason ($self, $new_val, $result, $reason, $restart) {
```

lib/OpenQA/Schema/Result/Jobs.pm
```diff
@@ -2109,6 +2116,41 @@ sub done ($self, %args) {
...
+        # restart incompletes when the reason matches the configured regex
+        my $append_reason = '';
+        my $auto_clone_regex = OpenQA::App->singleton->config->{global}->{auto_clone_regex};
+        if ($result eq INCOMPLETE and $auto_clone_regex and $reason =~ $auto_clone_regex) {
+            my $limit = OpenQA::App->singleton->config->{global}{auto_clone_limit};
+            my $ancestors = $self->incomplete_ancestors($limit + 1);
+            # how many of those incomplete ancestors had a reason not matching auto_clone?
+            my $unrelated = grep { $_->[2] !~ m/$auto_clone_regex/ } @$ancestors;
+            if (@$ancestors < $limit || $unrelated > 0) {
+                $append_reason = ' [Auto-restarting because reason matches auto_clone_regex]';
```
okurz, line 2133:
```suggestion
                $append_reason = ' [Auto-restarting because reason matches $auto_clone_regex]';
```

lib/OpenQA/Schema/Result/Jobs.pm
```diff
@@ -2109,6 +2116,41 @@ sub done ($self, %args) {
...
+                $$restart = 1;
+            }
+            else {
+                $append_reason
+                  = ' [Not restarting because job has been restarted already auto_clone_limit times and failed with /$auto_clone_regex/]';
```
okurz, line 2138:
```suggestion
                  = ' [Not restarting because job has been restarted already $auto_clone_limit times and failed with /$auto_clone_regex/]';
```

as discussed

okurz (approved):

PR #5398, opened 2023-12-14
Use postgresql recursive feature for getting scheduled product

perlpunk:
For longer chains of clones it's more efficient. It also avoids a warning:

    Deep recursion on subroutine "OpenQA::Schema::Result::Jobs::related_scheduled_product_id"

Issue: https://progress.opensuse.org/issues/152569

### DIFF
--- a/lib/OpenQA/Schema/Result/Jobs.pm
+++ b/lib/OpenQA/Schema/Result/Jobs.pm
@@ -2014,7 +2014,18 @@ sub cancel_whole_clone_chain ($self, @args) {
 # returns the related scheduled product; if the job has not been created by one the origin job is checked
 sub related_scheduled_product_id ($self) {
     if (my $sp_id = $self->scheduled_product_id) { return $sp_id }
-    if (my $origin = $self->origin) { return $origin->related_scheduled_product_id }
+    return $self->{_orig_scheduled_product_id} if exists $self->{_orig_scheduled_product_id};
+    my $sth = $self->result_source->schema->storage->dbh->prepare(<<~'EOM');
+        with recursive orig_id as (
+            select ? as orig_id, ? as prod
+            union all
+            select id as orig_id, scheduled_product_id as prod from jobs join orig_id on orig_id.orig_id = jobs.clone_id)
+        select prod from orig_id where prod is not null
+        EOM
+    $sth->bind_param(1, $self->id, SQL_BIGINT);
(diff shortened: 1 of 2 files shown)


kalikiana:
> For longer chains of clones it's more efiicient. It also avoids a warning:

Typo: efficient

The SQL looks correct, but I'm not sure how to tell if it's more or less efficient.

perlpunk:
Regarding efficiency:
```
openqa=> explain analyze with recursive orig_id as (
            select (13062217)::bigint as orig_id, (null)::bigint as prod
            union all
select id as orig_id, scheduled_product_id as prod from jobs join orig_id on orig_id.orig_id = jobs.clone_id)
select orig_id, prod from orig_id where prod is not null order by orig_id desc;
                                                                   QUERY PLAN                                                                   
------------------------------------------------------------------------------------------------------------------------------------------------
 Sort  (cost=854.62..854.87 rows=100 width=16) (actual time=10.769..10.770 rows=1 loops=1)
   Sort Key: orig_id.orig_id DESC
   Sort Method: quicksort  Memory: 25kB
   CTE orig_id
     ->  Recursive Union  (cost=0.00..849.28 rows=101 width=16) (actual time=0.002..10.420 rows=1099 loops=1)
           ->  Result  (cost=0.00..0.01 rows=1 width=16) (actual time=0.001..0.001 rows=1 loops=1)
           ->  Nested Loop  (cost=0.42..84.72 rows=10 width=16) (actual time=0.009..0.009 rows=1 loops=1099)
                 ->  WorkTable Scan on orig_id orig_id_1  (cost=0.00..0.20 rows=10 width=8) (actual time=0.000..0.000 rows=1 loops=1099)
                 ->  Index Scan using jobs_idx_clone_id on jobs  (cost=0.42..8.44 rows=1 width=24) (actual time=0.008..0.008 rows=1 loops=1099)
                       Index Cond: (clone_id = orig_id_1.orig_id)
   ->  CTE Scan on orig_id  (cost=0.00..2.02 rows=100 width=16) (actual time=10.757..10.764 rows=1 loops=1)
         Filter: (prod IS NOT NULL)
         Rows Removed by Filter: 1098
 Planning Time: 0.187 ms
 Execution Time: 10.811 ms
(15 rows)
```
And that example is a job which has about 1100 ancestors.

Martchus (approved):

kalikiana (approved):

PR #5400, opened 2023-12-18
Fix dependencies not showing up in graph under certain conditions

Martchus:
When computing the dependencies to show in the dependency graph we prefer to show the latest child of a parent instead of the child we immediately found when traversing the list of children. This is implemented as an early return. In case of this early return we must *not* count the immediate child as visited. Otherwise we might miss it when traversing the dependency graph further (and it is the latest child from that perspective).

Related ticket: https://progress.opensuse.org/issues/136154

---

I have tested it manually ~~and it works but I'll keep it a draft because I need to extend unit tests~~.

### DIFF
--- a/t/ui/16-tests_dependencies.t
+++ b/t/ui/16-tests_dependencies.t
@@ -15,6 +15,7 @@ use OpenQA::Test::Case;
 use OpenQA::Client;
 use OpenQA::SeleniumTest;
 use OpenQA::Schema::Result::JobDependencies;
+use OpenQA::JobDependencies::Constants qw(CHAINED DIRECTLY_CHAINED);
 use OpenQA::Jobs::Constants;
 
 my $test_case = OpenQA::Test::Case->new;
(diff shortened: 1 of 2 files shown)


okurz (changes requested):
I suggest some before and after screenshots of the dependency trees

Martchus:
Before:
screenshot_20231219_112450

After:
screenshot_20231219_112538

(Note that in production the "skipped" job would probably be "parallel_failed" instead. Here it is only skipped because I've never let it execute on a worker.)

t/ui/16-tests_dependencies.t
```diff
@@ -148,6 +129,54 @@ subtest 'dependency json' => sub {
...
+        my @expected_nodes = (
+            99937,    # because it has not been duplicated
+            99938,    # because it has not been duplicated
+            @new_jobs    # because those are the duplicated jobs
+        );
+        is_deeply \@rendered_nodes, \@expected_nodes, 'only the latest jobs are shown'
+          or diag explain \@rendered_nodes;
+    };
+
+    $schema->txn_rollback;
```
kalikiana, line 177: > I have tested it manually and it works but I'll keep it a draft because I need to extend unit tests.

I think the description of the PR is lying 😉

okurz: solved

okurz (approved):

kalikiana (approved):

PR #5401, opened 2023-12-19
Add group filter options for /tests

kraih:
Based on #5388, and reusing most of the same logic, to bring job group filtering to `/tests` from `/tests/overview`. Similar to the `?match=...` option, this does not have a UI element yet and can only be used by manually entering the query parameter. Since the question of how to extend the filter form with text fields is still open, i consider it out of scope for this PR.

### DIFF
--- a/assets/javascripts/tests.js
+++ b/assets/javascripts/tests.js
@@ -245,7 +245,7 @@ function renderTestLists() {
       this[paramName] = paramValues[0];
     }
   };
-  jQuery.each(['limit', 'groupid', 'match'], function (index, paramName) {
+  jQuery.each(['limit', 'groupid', 'match', 'group_glob', 'not_group_glob'], (index, paramName) => {
     ajaxQueryParams.addFirstParam(paramName);
   });
   delete ajaxQueryParams.addFirstParam;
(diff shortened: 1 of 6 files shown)


kalikiana (approved):
The naming seems... awkward, but it looks pretty good and I honestly don't have a better suggestion so I guess I'm fine with it

kraih:
Naming is at least consistent with the other PR.

okurz (changes requested):
Care to add some documentation or help text?

kraih:
> Care to add some documentation or help text?

Where and in what form? I believe the `?match=...` and `?groupid=...` query parameters are also undocumented for `/tests`.

kraih:
> > > Care to add some documentation or help text?
> > 
> > 
> > Where and in what form? I believe the `?match=...` and `?groupid=...` query parameters are also undocumented for `/tests`.
> 
> That is correct. How about to mention them all as hover text on the "Search:" label like visible on https://openqa.opensuse.org/tests/ ?

Sure, but not in this PR. I do not want to touch unrelated features here and create a scope creep situation.

kraih:
I've expanded the tooltip at the bottom of `/tests` for `group_glob` and `not_group_glob`:
popup

okurz (approved):
perfect!

PR #5404, opened 2023-12-20
Add a note for `Mojo::Base` in Documentation for Tests

josegomezr:
- Explain that it includes default best practices for perl (warnings, strict, utf8, etc.).

- Add a link to `Mojo::Base` upstream docs.

### DIFF
--- a/docs/WritingTests.asciidoc
+++ b/docs/WritingTests.asciidoc
@@ -58,6 +58,7 @@ sub run () {
     # write in this block the code for your test.
 }
 -------------------------------------------------------------------
+* **Note:** `Mojo::Base` _automatically enables: strict, warnings, utf8, perl5.16. See https://docs.mosjolicious.org/Mojo/Base#DESCRIPTION[Mojo::Base Description]_
 
 And here is an example in Python:
 


docs/WritingTests.asciidoc
```diff
@@ -51,13 +51,15 @@ Here is an example in Perl:
 
 [source,perl]
 -------------------------------------------------------------------
-use Mojo::Base "basetest";
+use Mojo::Base "basetest"; # See note <1>
```
okurz, line 54: Hm, keeping that in the code example means that people either copy the comment verbatim with the dangling reference or they need to remove that comment. How about we just keep the note below in line 62 and you don't change the code example itself?

docs/WritingTests.asciidoc
```diff
@@ -51,13 +51,15 @@ Here is an example in Perl:
...
+
```
okurz, line 55: unrelated change. I would prefer to not add that blank line.

okurz (changes requested):

josegomezr:
Removed the blank line & the `<>` callout in favor of a note after the code sample.

docs/WritingTests.asciidoc
kalikiana: Can we lint that?

kalikiana: Not in this branch, just wondering since it's a very common point that comes up

kalikiana (approved):
Nice!

PR #5408, opened 2023-12-21
Prevent ambiguous PRJDIR ending up in vars.json

okurz:
We observed that the variable PRJDIR ends up with inconsistent values in
vars.json files which is then rendered as a diff from the openQA
investigation tab even though likely the effective value used within
openQA never differed. The variable PRJDIR is actually never used itself
so to prevent confusion we can simply remove all according references.
The internal function "prjdir()" is untouched.

This change effectively prevents pre-2020 versions of os-autoinst, in
particular before commit 9eb2f564 removing the last use of PRJDIR, to
work which should not be our concern because for multiple reasons it
would be a bad idea if not outright impossible to have a current openQA
working with such old versions of os-autoinst.

Related progress issue: https://progress.opensuse.org/issues/152392


okurz:
No CI checks failed so we can just assume that this will be fine. Undrafted.

perlpunk (approved):

kalikiana (approved):
🤞🏾

PR #5409, opened 2023-12-21
ObsRsync: Include URL in GRU task

kalikiana:
This fascilitates debugging of issues with sync.

See: https://progress.opensuse.org/issues/112871

### DIFF
--- a/lib/OpenQA/WebAPI/Plugin/ObsRsync/Controller/Gru.pm
+++ b/lib/OpenQA/WebAPI/Plugin/ObsRsync/Controller/Gru.pm
@@ -63,7 +63,7 @@ sub _extend_job_info {
 
     my $args = $job->{args};
     $args = $args->[0] if (ref $args eq 'ARRAY' && scalar(@$args) == 1);
-    if (ref $args eq 'HASH' && scalar(%$args) == 1 && $args->{project}) {
+    if (ref $args eq 'HASH' && $args->{project}) {
         $args = $args->{project};
     }
     else { $args = dump($args) }    # uncoverable statement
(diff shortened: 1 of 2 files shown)


perlpunk (approved):

okurz:
t/ui/27-plugin_obs_rsync_obs_status.t failed. I suggest you run that locally

lib/OpenQA/WebAPI/Plugin/ObsRsync/Controller/Gru.pm
```diff
@@ -128,7 +128,7 @@ sub run {
     return $self->render(json => {message => 'queue full'}, status => QUEUE_FULL)
       if ($results->{total} >= $queue_limit);
 
-    $app->gru->enqueue('obs_rsync_run', {project => $project}, {priority => 100});
+    $app->gru->enqueue('obs_rsync_run', {project => $project, url => $helper->project_status_url}, {priority => 100});
```
kalikiana, line 131: I can't make sense of the unit test failure. Adding the `url` argument somehow causes the list of inactive projects to be different. Changing `_extend_job_info` doesn't seem relevant. There must be something else?

perlpunk: The adjustment needs to be done here as well: https://github.com/os-autoinst/openQA/blob/master/lib/OpenQA/WebAPI/Plugin/ObsRsync/Controller/Folders.pm#L75
```
            next unless (ref $args eq 'HASH' && scalar(%$args) == 1 && $args->{project});
```

kalikiana: That was it! Thank you so much!

PR #5411, opened 2023-12-22
Handle no job group filter matches gracefully

kraih:
This is a followup for #5388 and #5401. When there were no matches for job group globs, no query condition would be generated previously. Giving the false impression that all job groups were matching. So now we just generate an impossible query that cannot match anything with the group id `0`.

### DIFF
--- a/lib/OpenQA/WebAPI/Controller/Test.pm
+++ b/lib/OpenQA/WebAPI/Controller/Test.pm
@@ -719,9 +719,8 @@ sub _prepare_job_results ($self, $all_jobs, $limit) {
 }
 
 sub _prepare_groupids ($self) {
-    if (my @groups = $self->groups_for_globs) {
-        return [map { $_->id } @groups];
-    }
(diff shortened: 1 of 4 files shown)


perlpunk:
`s/Hanlde/Handle/` :)

kalikiana (approved):

perlpunk (approved):

lib/OpenQA/WebAPI/Controller/Test.pm
```diff
@@ -719,9 +719,8 @@ sub _prepare_job_results ($self, $all_jobs, $limit) {
 }
 
 sub _prepare_groupids ($self) {
-    if (my @groups = $self->groups_for_globs) {
-        return [map { $_->id } @groups];
-    }
+    return [0] unless my $groups = $self->groups_for_globs;
+    return [map { $_->id } @$groups] if @$groups;
```
okurz, line 724: looks even cleaner than before, nice! :)

PR #5412, opened 2023-12-28
CI: Try to increase timeout for "cache" job

okurz:
As seen in recent failures download from registry.opensuse.org seems to
either take longer than the usual timeouts or actually gets stuck for
too long. This commit should increase the timeout from 20m to 50m, just
short of the hard overall job timeout of 1h in our "free" plan of using
circleCI. With this we should be able to either fix the problems we
recently observed or it's more obvious if the download which should
normally only take a couple of minutes is instead getting stuck for
good.

Related progress issue: https://progress.opensuse.org/issues/152941

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -122,6 +122,7 @@ jobs:
           command: tools/ci/prepare_dependency_pr.sh
 
   cache:
+    no_output_timeout: 50m
     docker:
       - <<: *base
     steps:


perlpunk (approved):

okurz:
the config key is applied on the wrong level and can not be applied for the failing step "Spin up environment". This won't work

PR #5413, opened 2024-01-03
CI: Shift nightly schedule further trying to avoid download problems

okurz:
Related progress issue: https://progress.opensuse.org/issues/152941


kalikiana (approved):
Sounds fine

PR #5418, opened 2024-01-05
Download shepherd.js from cdn.jsdelivr.net

perlpunk:
Issue: https://progress.opensuse.org/issues/153145

### DIFF
--- a/dist/rpm/update-cache.sh
+++ b/dist/rpm/update-cache.sh
@@ -1,5 +1,8 @@
 #! /bin/sh
 
+# Note: this might not work correctly anymore because the 'lr' command
+# was obsoleted.
+
 set -e
 
 export LC_ALL='en_US.UTF-8'
@@ -9,7 +12,7 @@ rm -f _service\:*
 rm -f *.tar *.cpio
 osc service lr
 # special call for tar buildtime service
-osc service lr tar
+osc service run tar
 
 SD=$PWD
 cd openQA
--- a/assets/assetpack.def
+++ b/assets/assetpack.def
@@ -149,7 +149,7 @@
 < https://cdnjs.cloudflare.com/ajax/libs/chosen/1.7.0/chosen.jquery.js
 < https://cdnjs.cloudflare.com/ajax/libs/jquery-ujs/1.2.1/rails.js
 < javascripts/disable_animations.js [mode==test]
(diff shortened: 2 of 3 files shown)


perlpunk:
I had problems with running update-cache.sh, I executed the steps manually:
```
% osc service lr
WARNING: Command 'lr' is obsolete
Already up to date.



bf59e0109f6296a26001991c924db2698af58fa5
% osc service lr tar

WARNING: Command 'lr' is obsolete

# Then I did this instead:
% osc service run tar
```
but after that `_servicedata` and `openQA.changes` were modified with a comment by Oli.
So it might not work correctly in the future.

baierjan:
> Since the osc docs don't tell me what `lr` was supposed to do, I don't have a replacement for it. It's just obsolete :(

According to `osc help service`:
>   localrun    lr the same as "run" but services with mode "serveronly" are also executed

dist/rpm/update-cache.sh
```diff
@@ -1,5 +1,8 @@
 #! /bin/sh
 
+# Note: this might not work correctly anymore because the 'lr' command
+# was obsoleted.
```
kalikiana, line 4: Is there documented anywhere? I can't find any mention of it in https://github.com/search?q=repo%3AopenSUSE%2Fobs-docu+localrun

perlpunk:
We can find out about update-cache later. It would be good to merge this to fix our build and not disturb other waiting PRs

dist/rpm/update-cache.sh
asdil12: If you call `osc service lr`, it will tell you.

asdil12 (approved):

baierjan (approved):

dist/rpm/update-cache.sh
perlpunk: Like @baierjan wrote, `osc help service` tells me, but I don't know what I can use as a replacement.
I don't know what the update-cache.sh script is supposed to do. When I ran it, more files than expected were changed.

Also, in this case the script cannot be used because it would do `cd openQA` and generate the tar file from the existing asset urls from the version of openQA in OBS, but I needed the changed version from my local update.

asdil12:
```
$ osc service lr
WARNING: Command 'lr' is obsolete, please use 'run' instead.

PR #5419, opened 2024-01-10
Reload dbus in post install script

asdil12:
This is needed so that newly created users are taken into account
for dbus permission handling.

### DIFF
--- a/dist/rpm/openQA.spec
+++ b/dist/rpm/openQA.spec
@@ -495,10 +495,12 @@ else
 fi
 
 %service_add_post %{openqa_services}
+systemctl reload dbus || :
 
 %post worker
 %tmpfiles_create %{_tmpfilesdir}/openqa.conf
 %service_add_post %{openqa_worker_services}
+systemctl reload dbus || :
 
 %post auto-update
 %service_add_post openqa-auto-update.timer


kalikiana (approved):

okurz (changes requested):
* Are you sure this actually works?
* Have you tried it out?
* Is there a macro to be used for systemd or should this be unconditionally included?
* Please add a ticket reference
* I suggest to rewrite the git commit message to make the effect clear in the subject line, i.e. turn around the subject and first line
* Isn't this about os-autoinst where script/os-autoinst-setup-multi-machine and systemd/os-autoinst-openvswitch.service.in live?

asdil12:
> * Are you sure this actually works?

No

> * Have you tried it out?

I tried but we seem to do too much black magic so I wasn't able to quickly build the rpm locally using `osc build`.

> * Is there a macro to be used for systemd or should this be unconditionally included?

I don't know - if there is one, factory reviewers might tell us. Documentation in the wiki doesn't state that anything needs to be used.
I added '|| :' so that it should still work in docker containers.

> * Please add a ticket reference

https://progress.opensuse.org/issues/138302

> * I suggest to rewrite the git commit message to make the effect clear in the subject line, i.e. turn around the subject and first line

I think it makes more sence like it is now.

> * Isn't this about os-autoinst where script/os-autoinst-setup-multi-machine and systemd/os-autoinst-openvswitch.service.in live?

The user is created in this spec file (actually in the %pre script - see https://en.opensuse.org/openSUSE:Packaging_guidelines#Users_and_Groups) - so it makes sence to reload dbus after the user got created.

But you are right - maybe it is not about the user but about the dbus config:


```
[dheidler@midas tmp]$ bsdtar xvf os-autoinst-openvswitch-4.6.1704884885.e3cee62-1735.1.x86_64.rpm
x ./etc/dbus-1/system.d/org.opensuse.os_autoinst.switch.conf
x ./usr/lib/os-autoinst/os-autoinst-openvswitch
x ./usr/lib/systemd/system/os-autoinst-openvswitch.service
x ./usr/sbin/rcos-autoinst-openvswitch
[dheidler@midas tmp]$ cat ./etc/dbus-1/system.d/org.opensuse.os_autoinst.switch.conf
<?xml version="1.0"?> <!--*-nxml-*-->
<!DOCTYPE busconfig PUBLIC "-//freedesktop//DTD D-BUS Bus Configuration 1.0//EN"
        "http://www.freedesktop.org/standards/dbus/1.0/busconfig.dtd">

<!--
  This file is part of os-autoinst.
-->

<busconfig>

        <policy user="root">
                <allow own="org.opensuse.os_autoinst.switch"/>
                <allow send_destination="org.opensuse.os_autoinst.switch"/>
                <allow receive_sender="org.opensuse.os_autoinst.switch"/>
        </policy>

        <policy user="_openqa-worker">
                <allow send_destination="org.opensuse.os_autoinst.switch"/>
        </policy>

</busconfig>
```

asdil12:
Maybe we need to do it like some other packages do:


https://build.opensuse.org/package/view_file/openSUSE:Factory/certmonger/certmonger.spec?expand=1
```spec
%post
if test $1 -eq 1 ; then
  %{_bindir}/dbus-send --system --type=method_call --dest=org.freedesktop.DBus / org.freedesktop.DBus.ReloadConfig 2>&1 || :
fi
```

This is essentially the same as `systemctl reload dbus`:
```
ExecReload=/usr/bin/dbus-send --print-reply --system --type=method_call --dest=org.freedesktop.DBus / org.freedesktop.DBus.ReloadConfig
```

okurz:
> > * Are you sure this actually works?
> 
> No
> 
> > * Have you tried it out?
> 
> I tried but we seem to do too much black magic so I wasn't able to quickly build the rpm locally using `osc build`.

The OBS CI checks actually generate the packages so after fixing the OBS build failing you can install projects from those repos – maybe need to be set to published – and then try it out.

> > * Please add a ticket reference
> 
> https://progress.opensuse.org/issues/138302

in the git commit message.

> > * Isn't this about os-autoinst where script/os-autoinst-setup-multi-machine and systemd/os-autoinst-openvswitch.service.in live?
> 
> The user is created in this spec file (actually in the %pre script - see https://en.opensuse.org/openSUSE:Packaging_guidelines#Users_and_Groups) - so it makes sence to reload dbus after the user got created.
> 
> But you are right - maybe it is not about the user but about the dbus config:
> 
> ```
> [dheidler@midas tmp]$ bsdtar xvf os-autoinst-openvswitch-4.6.1704884885.e3cee62-1735.1.x86_64.rpm
> x ./etc/dbus-1/system.d/org.opensuse.os_autoinst.switch.conf
> x ./usr/lib/os-autoinst/os-autoinst-openvswitch
> x ./usr/lib/systemd/system/os-autoinst-openvswitch.service
> x ./usr/sbin/rcos-autoinst-openvswitch
> [dheidler@midas tmp]$ cat ./etc/dbus-1/system.d/org.opensuse.os_autoinst.switch.conf
> <?xml version="1.0"?> 
> <!DOCTYPE busconfig PUBLIC "-//freedesktop//DTD D-BUS Bus Configuration 1.0//EN"
>         "http://www.freedesktop.org/standards/dbus/1.0/busconfig.dtd">
> 
> 
> 
> <busconfig>
> 
>         <policy user="root">
>                 <allow own="org.opensuse.os_autoinst.switch"/>
>                 <allow send_destination="org.opensuse.os_autoinst.switch"/>
>                 <allow receive_sender="org.opensuse.os_autoinst.switch"/>
>         </policy>
> 
>         <policy user="_openqa-worker">
>                 <allow send_destination="org.opensuse.os_autoinst.switch"/>
>         </policy>
> 
> </busconfig>
> ```

yes, so it looks better suited to do this in os-autoinst.

> Maybe we need to do it like some other packages do:
> 
> https://build.opensuse.org/package/view_file/openSUSE:Factory/certmonger/certmonger.spec?expand=1
> 
> ```python
> %post
> if test $1 -eq 1 ; then
>   %{_bindir}/dbus-send --system --type=method_call --dest=org.freedesktop.DBus / org.freedesktop.DBus.ReloadConfig 2>&1 || :
> fi
> ```


looks good.

Martchus:
> > * Are you sure this actually works?
> 
> No
> 
> > * Have you tried it out?
> 
> I tried but we seem to do too much black magic so I wasn't able to quickly build the rpm locally using `osc build`.
> 

Good that we have CI checks for that. Unfortunately they failed before due to an unrelated problem. Make sure to be based on the current master because you might be missing https://github.com/os-autoinst/openQA/pull/5418.

> > * I suggest to rewrite the git commit message to make the effect clear in the subject line, i.e. turn around the subject and first line
> 
> I think it makes more sence like it is now.
> 

I think "Ensure D-Bus config is effective after installing package" would make sense.

asdil12:
--> https://github.com/os-autoinst/os-autoinst/pull/2430

PR #5425, opened 2024-01-15
Allow searching for jobs by comment text

Martchus:
See particular commit messages and https://progress.opensuse.org/issues/151402

### DIFF
--- a/t/10-tests_overview.t
+++ b/t/10-tests_overview.t
@@ -23,6 +23,7 @@ sub get_summary { OpenQA::Test::Case::trim_whitespace($t->tx->res->dom->at('#sum
 
 my $jobs = $schema->resultset('Jobs');
 $jobs->find(99928)->update({blocked_by_id => 99927});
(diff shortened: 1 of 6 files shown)


t/10-tests_overview.t
```diff
@@ -23,6 +23,7 @@ sub get_summary { OpenQA::Test::Case::trim_whitespace($t->tx->res->dom->at('#sum
 
 my $jobs = $schema->resultset('Jobs');
 $jobs->find(99928)->update({blocked_by_id => 99927});
+$jobs->find($_)->comments->create({text => 'test1', user_id => 99901}) for 99946, 99963;
```
kraih, line 26: Maybe don't make it an exact match for the search string? So you can be sure substring match works.

perlpunk (approved):

okurz (approved):
LGTM as is. You can consider adding a small help text like in https://github.com/os-autoinst/openQA/blob/69e8e802aadb9b4d330fb30e0701c8f6a84f5945/templates/webapi/test/list.html.ep#L48

PR #5426, opened 2024-01-16
Fix "duplicate key value violates unique constraint" on needle updates

Martchus:
* Use raw SQL queries with "ON CONFLICT DO NOTHING" to avoid running into
  those errors
* Follow example from `OpenQA::Schema::ResultSet::Assets::register` where
  we already use raw queries like that followed by an DBIx call to get the
  newly inserted row
* Not using `{for => 'update'}` here to locking as we likely don't need it
  here as we don't mind the *exact* last module
* Avoid repeated function calls simplifying the code a little bit
* See https://progress.opensuse.org/issues/153616

### DIFF
--- a/lib/OpenQA/Schema/Result/Needles.pm
+++ b/lib/OpenQA/Schema/Result/Needles.pm
@@ -75,6 +75,8 @@ __PACKAGE__->belongs_to(
 __PACKAGE__->belongs_to(directory => 'OpenQA::Schema::Result::NeedleDirs', 'dir_id');
 
 # override insert to ensure new needles have 'last_updated' set
+# note: Not used by the function update_needle but still required for unit tests that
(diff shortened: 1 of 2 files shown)


okurz (approved):

asdil12:
> However, I would have to find out how to bring this change into the test setup first.

This is not trivial.
You would have to build RPMs and modify line 16 in `tests/install/openqa_webui.pm` to use that repo instead of `devel:openQA`.

asdil12 (approved):

asdil12:
Looks very good: https://openqa.opensuse.org/tests/overview?version=Tumbleweed&build=dheidler_mm3_ping&distri=openqa&groupid=38

The first 10 out of 10 all passed.

PR #5427, opened 2024-01-16
Improve UI for comments filter

Martchus:
See particular commit messages and https://progress.opensuse.org/issues/151402#note-9


okurz (commented):
+1

t/ui/10-tests_overview.t fails

okurz (approved):

t/ui/10-tests_overview.t
```diff
@@ -139,7 +139,7 @@ is(scalar @filtered_out, 0, 'result filter correctly applied');
 # Test whether all URL parameter are passed correctly
 my $url_with_escaped_parameters
   = $baseurl
-  . 'tests/overview?arch=&flavor=&machine=&test=&modules=&module_re=&group_glob=&not_group_glob=&distri=opensuse&build=0091&version=Staging%3AI&groupid=1001';
+  . 'tests/overview?arch=&flavor=&machine=&test=&modules=&module_re=&group_glob=&not_group_glob=&comment=&distri=opensuse&build=0091&version=Staging%3AI&groupid=1001';
```
kalikiana, line 142: Can/should we add an actual comment here or in an additional case? It looks like it would be trivial?

kalikiana: Answering my own q. The logic is tested in the previously added test. Here we just confirm the arg is in the URL.

PR #5428, opened 2024-01-17
Preserve assets that are symlinks to the fixed directory

Martchus:
See https://progress.opensuse.org/issues/152377

### DIFF
--- a/lib/OpenQA/Schema/Result/Assets.pm
+++ b/lib/OpenQA/Schema/Result/Assets.pm
@@ -4,7 +4,7 @@
 package OpenQA::Schema::Result::Assets;
 
 
-use Mojo::Base 'DBIx::Class::Core';
+use Mojo::Base 'DBIx::Class::Core', -signatures;
 
 use OpenQA::App;
 use OpenQA::Jobs::Constants;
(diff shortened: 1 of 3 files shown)


lib/OpenQA/Schema/Result/Assets.pm
```diff
@@ -106,6 +106,12 @@ sub remove_from_disk {
         log_info("GRU: skipping removal of $type/$name; it does not exist anyways");
         return undef;
     }
+    if (my $target = readlink $file) {
```
Martchus, line 109: The preceding if-block let's me ponder: Will we ever delete dangling symlinks? It doesn't look like it because we call `locate_asset` with `mustexit => 1` which does a `-e` check which requires not only the symlink to exist but also its target. So maybe there's also room for improvement when it comes to deleting dangling symlinks.

lib/OpenQA/Schema/Result/Assets.pm
```diff
@@ -106,6 +106,12 @@ sub remove_from_disk {
...
+        if ($target =~ qr{^fixed/} && -e $target) {
```
Martchus, line 110: This would of course only work for relative symlinks but looking at the list of symlinks on OSD it would be good enough.

lib/OpenQA/Schema/Result/Assets.pm
Martchus: My newly pushed changes will address that problem.

lib/OpenQA/Schema/Result/Assets.pm
```diff
@@ -96,22 +96,18 @@ sub is_fixed {
...
-    else {
-        log_error("GRU: unable to remove $file");    # uncoverable statement trivial error report
+sub remove_from_disk ($self) {
+    my $file = locate_asset($self->type, $self->name);
+    if (my $target = readlink $file) {
+        return _unlink $file, "GRU: removed dangling symlink '$file' pointing to '$target'" unless -e $file;
+        return log_info "GRU: skipping removal of '$file' as it points to '$target' in the fixed directory"
+          if $target =~ qr{^(\./)?fixed/};
     }
+    return log_info "GRU: skipping removal of '$file'; it does not exist anyway" unless -e $file;
```
okurz, line 108:
```suggestion
    return log_info "GRU: skipping removal of non-existing '$file'" unless -e $file;
```

Martchus:
Removing the draft-state because I think this is an improvement regardless of the concrete use case the ticket is about.

okurz (approved):

kalikiana (approved):

PR #5433, opened 2024-01-23
Add /job_groups/id/build_results API route

perlpunk:
Issue: https://progress.opensuse.org/issues/152939
    
I moved `_map_tags_into_build` into the `compute_build_results()` function.
    
The command lime to get the last "published" build would look like this:
    
        openqa-cli api job_groups/1/build_results only_tagged=1 limit_builds=1 \
          | jq -r '[.build_results[] | select(.tag.description=="published") | select(.version=="Tumbleweed") | .build ][0]'
    
To only get the latest build:
    
        openqa-cli api job_groups/1/build_results  limit_builds=1| jq -r '[.build_results[] | .build ][0]'

Only normal groups are supported, not parent groups, to keep it simple. I can go though our logs to see if the group_overview.json is ever called for parent groups.

I am seeing a problem with the documentation display of the new function when getting a 404. The documentation for the new function is displayed for `+/job_groups/<group_id:num>/jobs GET` as well as ` +/job_groups/<group_id:num>/build_results`.
Maybe a bug in how Mojolicious renders the pod. I can't see anything wrong there.

Also I found a misdocumented route. I will open a seperate PR.

Still need to document this somewhere.

### DIFF
--- a/lib/OpenQA/WebAPI/Controller/API/V1/JobGroup.pm
+++ b/lib/OpenQA/WebAPI/Controller/API/V1/JobGroup.pm
@@ -326,4 +326,45 @@ sub delete ($self) {
     $self->render(json => $event_data);
 }
 
+=over 4
+
+=item build_results()
+
+Shows build results for a job group, similar to what the group_overview page
+provides.
+
+Currently it does not support parent job groups.
+
+Use limit_builds=n to limit the number of returned builds. Default is 10.
+
+Use time_limit_days=n to only go back n days.
+
+Use only_tagged=1 to only return tagged builds.
+
+Use show_tags=1 to show tags for each build. only_tagged implies show_tags.
+
+=back
+
+=cut
+
+sub build_results ($self) {
+    my $group = $self->find_group() or return;
+    my $validation = $self->validation;
+    $validation->optional('limit_builds')->num;
+    $validation->optional('time_limit_days')->like(qr/^[0-9.]+$/);
+    $validation->optional('only_tagged');
+    $validation->optional('show_tags');
+    my $limit_builds = $validation->param('limit_builds') // 10;
+    my $time_limit_days = $validation->param('time_limit_days') // 0;
+    my $only_tagged = $validation->param('only_tagged') // 0;
+    my $show_tags = $validation->param('show_tags') // $only_tagged;
+
+    my $tags = $show_tags ? $group->tags : undef;
+    my $cbr
+      = OpenQA::BuildResults::compute_build_results($group, $limit_builds,
+        $time_limit_days, $only_tagged ? $tags : undef,
+        [], $tags);
+    $self->render(json => $cbr);
+}
+
 1;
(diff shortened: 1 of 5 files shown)


okurz (commented):
+1

lib/OpenQA/WebAPI/Controller/API/V1/JobGroup.pm
```diff
@@ -344,4 +326,45 @@ sub delete {
...
 }
 
+=over 4
+
+=item build_results()
+
+Shows build results for a job group, similar to what the group_overview page
+provides.
+
+Currently it does not support parent job groups.
```
foursixnine, line 336: Because it is unclear to me, will the user get an error message? For sake's simplicity, I guess, making it explicit is better in this case.

Martchus:
> `openqa-cli api job_groups/1/build_results | jq -r '[.build_results[] | .build ][0]'`

So the `limit_builds` parameter is not supported (as on the UI route like this: https://openqa.suse.de/group_overview/110?limit_builds=1)? It would make sense to support the same set of parameters:

`openqa-cli api job_groups/1/build_results?limit_builds=1`

Martchus (approved):

lib/OpenQA/WebAPI/Controller/API/V1/JobGroup.pm
perlpunk: The parent job groups are in a different table and the ids are overlapping with job group ids.
For the WebUI we have `/parent_group_overview/x` and `/group_overview/x`, so we would need also a new route for parent groups, if people want to have that.

foursixnine: Got it, thanks!

perlpunk:
> `openqa-cli api job_groups/1/build_results?limit_builds=1`
> 
> EDIT: Looks like this would actually already work.

I edited the commit message to include `limit_builds`

okurz (approved):

AdamWill:
One test from this is failing in Fedora, for some reason. `$t->json_is('/build_results/4/all_passed' => '');` fails because it's coming out as "0", not "". I've no idea why. We have perl 5.40.0-509.fc41 and Mojolicious 9.37; no idea what else would be involved in the JSONification. I can just patch the expected result, I guess - functionally, "0" should behave the same as "", they both evaluate false in `if` checks, which is what this is used for - but it's weird.

AdamWill:
Huh, this "fixes" it:
```
-    $build_res->{all_passed} = $build_res->{passed} + $build_res->{softfailed} >= $build_res->{total};
+    $build_res->{all_passed} = $build_res->{passed} + $build_res->{softfailed} >= $build_res->{total} ? "1" : "";
```
...weird.

AdamWill:
Hmm, this feels very fuzzy. If I make the code into this:
    $build_res->{all_passed} = $build_res->{passed} + $build_res->{softfailed} >= $build_res->{total} ? 1 : 0;
and then make the check this:
    $t->json_is('/build_results/4/all_passed' => 0);
it passes, not surprisingly. But it *also* passes if I make the check this:
    $t->json_is('/build_results/4/all_passed' => "0");
so it "is" both 0 and "0"? Hmm.

AdamWill:
https://github.com/os-autoinst/openQA/pull/5731

PR #5434, opened 2024-01-23
Update os-autoinst-common and yaml files

perlpunk:
Currently we cannot use .yamllint from os-autoinst-common as
the yamllint version in Leap does not support ignore-from-file.
    
I ran all files which yamllint complained about through yamltidy,
and all passes now.

    git subrepo clone (merge) --force git@github.com:os-autoinst/os-autoinst-common.git external/os-autoinst-common

    subrepo:
      subdir:   "external/os-autoinst-common"
      merged:   "a94b58b23"
    upstream:
      origin:   "git@github.com:os-autoinst/os-autoinst-common.git"
      branch:   "master"
      commit:   "a94b58b23"
    git-subrepo:
      version:  "0.4.6"
      origin:   "git@github.com:ingydotnet/git-subrepo"
      commit:   "110b9eb"


okurz (approved):

t/data/job-templates/openqa-null.yaml
```diff
@@ -16,19 +16,19 @@ products:
...
-        testsuite: null
-        settings:
-          OPENQA_FROM_GIT: '1'
-        description: >-
-          Maintainer: okurz@suse.de Test for running openQA itself from git. To be used with "openqa"
-          distri.
+      - openqa_install+publish:
+          testsuite: null
+          settings:
+            PUBLISH_HDD_1: opensuse-Tumbleweed-%ARCH%@%MACHINE%-%BUILD%.qcow2
```
b10n1k, line 22: just making sure. are the variable propagated without quotes?

Martchus (approved):

t/data/job-templates/openqa-null.yaml
perlpunk: The quotes here don't make a difference. In YAML you need to quote something if it would otherwise be a number, boolean etc., or if it contains special characters at the beginning, or `: `.

perlpunk:
I think this needs to be merged manually because .mergify.yml changed (even if it's just a formatting change. Theoretically mergify would be able to detect that...)

PR #5438, opened 2024-01-25
Ensure only the Git cache directory from the local worker config is used

Martchus:
* Prevent users from cluttering directories when accidentally specifying a wrong `GIT_CACHE_DIR`
* Avoid the Git cache directory configured on a production worker to be used when cloning a job to the local instance (where that directory might not even exists or another cache directory is wanted)
* See https://progress.opensuse.org/issues/154240

---

Note that I haven't extended tests because we already have a test for the changed line and we likely don't need to test the same behavior for all the relevant keys.

### DIFF
--- a/lib/OpenQA/Worker/Job.pm
+++ b/lib/OpenQA/Worker/Job.pm
@@ -237,7 +237,7 @@ sub start {
     # delete settings we better not allow to be set on job-level (and instead should only be set within the
     # worker config)
     my $job_settings = $self->info->{settings} // {};
-    delete $job_settings->{GENERAL_HW_CMD_DIR};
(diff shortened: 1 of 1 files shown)


baierjan (approved):

perlpunk (commented):
If the setting isn't used at all, why is it there?
Can we first discuss this please?

Martchus:
It might be there because a user created a job explicitly specifying it but in production we still want to use what's configured by us. (You have mentioned yourself that users might mistakes which can be dangerous.)

I'm actually not sure anymore whether `openqa-clone-job` would actually pull that setting in if it is just in `vars.json` and not in the openQA database. However, I think it makes sense to just teat it as a "local setting" like we do for other settings already regardless of that.

kalikiana (approved):
As discussed the change makes sense.

PR #5443, opened 2024-01-26
Avoid deprecation warnings in GitHub action for JavaScript linting

Martchus:
* Update Node version to 20
* Update the version of the actions
* Improve naming of the action/steps

### DIFF
--- a/.github/workflows/javascript.yml
+++ b/.github/workflows/javascript.yml
@@ -4,17 +4,17 @@ name: javascript
 on: [push, pull_request]
 jobs:
   test:
-    name: Node ${{ matrix.node-version }} and ${{ matrix.os }}
+    name: Lint JavaScript code (using Node version ${{ matrix.node-version }})
     runs-on: ${{ matrix.os }}
     strategy:
       fail-fast: false
       matrix:
-        node-version: [16.x]
+        node-version: [20]
         os: [ubuntu-latest]
     steps:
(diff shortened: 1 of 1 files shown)


baierjan (approved):

.github/workflows/javascript.yml
```diff
@@ -4,17 +4,17 @@ name: javascript
...
 jobs:
   test:
-    name: Node ${{ matrix.node-version }} and ${{ matrix.os }}
+    name: Lint JavaScript code (using Node version ${{ matrix.node-version }})
     runs-on: ${{ matrix.os }}
     strategy:
       fail-fast: false
       matrix:
-        node-version: [16.x]
+        node-version: [18]
```
baierjan, line 12: This is 18, not 20 😉

baierjan (changes requested):
You might want to include additional value, change this one or edit the commit message.

.github/workflows/javascript.yml
Martchus: Oh, I forgot to actually stage that change in the end. Let's see whether 20 works as well because I suppose the newer the better.

By the way, I would have tried to avoid specifying a version at all but it is recommended by https://github.com/actions/setup-node.

baierjan (approved):

kraih (approved):

okurz (approved):

PR #5440, opened 2024-01-25
Add editorconfig and update bash files accordingly

b10n1k:
https://progress.opensuse.org/issues/153793

PR #5446, opened 2024-01-30
Add editorconfig and update bash files accordingly

b10n1k:
https://progress.opensuse.org/issues/153793
Pull latest changes of .editorconfig from os-autoinst-common, shfmt to the
dependencies and enable CI checks.

recreate https://github.com/os-autoinst/openQA/pull/5440

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -293,7 +293,7 @@ prepare-and-launch-container-to-run-tests-within: container-test-build launch-co
 
 # all additional checks not called by prove
 .PHONY: test-checkstyle-standalone
-test-checkstyle-standalone: test-shellcheck test-yaml test-critic
+test-checkstyle-standalone: test-shellcheck test-yaml test-critic test-shfmt
 ifeq ($(CONTAINER_TEST),1)
 test-checkstyle-standalone: test-check-containers
 endif
@@ -316,6 +316,11 @@ test-yaml:
 	@which yamllint >/dev/null 2>&1 || (echo "Command 'yamllint' not found, can not execute YAML syntax checks" && false)
 	@# Fall back to find if there is no git, e.g. in package builds
 	yamllint --strict $$((git ls-files "*.yml" "*.yaml" 2>/dev/null || find -name '*.y*ml') | grep -v ^dbicdh)
+
+.PHONY: test-shfmt
+test-shfmt:
+	@which shfmt >/dev/null 2>&1 || (echo "Command 'shfmt' not found, can not execute bash script syntax checks" && false)
+	shfmt -d -i 4 -bn -ci -sr $$(shfmt -f .)
 
 .PHONY: test-check-containers
 test-check-containers:
--- a/external/os-autoinst-common/.editorconfig
+++ b/external/os-autoinst-common/.editorconfig
@@ -1,0 +1,12 @@
+[*]
+indent_style = space
+indent_size = 4
+binary_next_line = true
+switch_case_indent = true
+space_redirects = true
+
+[.bpan/**]
+ignore = true
+
+[*~]
+ignore = true
--- a/tools/ci/ci-packages.txt
+++ b/tools/ci/ci-packages.txt
@@ -325,6 +325,7 @@ python3-Pillow-tk-8.4.0
 python3-PyYAML-5.4.1
 python3-tk-3.6.15
 python3-yamllint-1.22.1
+shfmt-3.5.1
 setxkbmap-1.3.1
 ShellCheck-0.8.0
 shfmt-3.5.1
--- a/container/openqa/entrypoint.sh
+++ b/container/openqa/entrypoint.sh
@@ -10,19 +10,19 @@ export LC_ALL="en_US.UTF-8"
 
 # First, try to upgrade all container dependencies (or we won't catch bugs
 # until a new container image is built)
-[ "$UPGRADE_FROM_ZYPPER" -eq 1 ] && \
-  sudo zypper --gpg-auto-import-keys -n ref --force && \
-  sudo zypper up -l -y
+[ "$UPGRADE_FROM_ZYPPER" -eq 1 ] \
+    && sudo zypper --gpg-auto-import-keys -n ref --force \
+    && sudo zypper up -l -y
 
 cp -rd /opt/openqa /opt/testing_area
 cd /opt/testing_area/openqa
 
 run_as_normal_user() {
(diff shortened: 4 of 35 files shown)


okurz:
@Martchus why are javascript issues still showing up like that?

And
https://app.circleci.com/pipelines/github/os-autoinst/openQA/12852/workflows/e01e4366-9228-4b2b-829a-825f2c36f18a/jobs/119783

says
"dependencies.yaml
  189:3     error    duplication of key "shfmt" in mapping  (key-duplicates)"

tools/ci/ci-packages.txt
```diff
@@ -325,6 +325,7 @@ python3-Pillow-tk-8.4.0
 python3-PyYAML-5.4.1
 python3-tk-3.6.15
 python3-yamllint-1.22.1
+shfmt-3.5.1
 setxkbmap-1.3.1
 ShellCheck-0.8.0
 shfmt-3.5.1
```
baierjan, line 331:
```suggestion
setxkbmap-1.3.1
ShellCheck-0.8.0
shfmt-3.5.1
```

b10n1k:
eslint starts failing once the .editorconfig is created

tools/ci/ci-packages.txt
b10n1k: have been fixed

okurz:
> eslint starts failing once the .editorconfig is created. The `"prettier/prettier": "error"` in `.eslintrc.json` is causing the errors

Ok, so as we don't have plans to merge PRs with failing checks what are your intentions about it?

external/os-autoinst-common/.editorconfig
```diff
@@ -0,0 +1,12 @@
+[*]
```
perlpunk, line 1: I just stumbled upon this issue: https://github.com/mvdan/sh/issues/664
And it mentions `[[shell]]`. I can't find that in the [shfmt documentation](https://manpages.debian.org/testing/shfmt/shfmt.1.en.html) nor in the [EditorConfig documentation](https://editorconfig.org/) but it seems to work :)

The problem seems to be that eslint is using the editorconfig as well, and as we are using `*` we can't blame it :)

perlpunk: Oh, nevermind, I misread the diff.
It was apparently touching the right files when i used [[shell]], but then it was using the default formatting rules, using tabs etc.

perlpunk: I tried `[script/**]` instead, but it still touched files elsewhere (and did not touch anything under script). Weird. Have to work on other tasks now though...

perlpunk: Apparently this works, but we can't use the config file then:
```
shfmt -w -i 4 -bn -ci -sr $(shfmt -f .)
```
e.g.
```
shfmt -w $(shfmt -f .)
```
does not use .editorconfig

So -f is able to find the right files.

Makefile
```diff
@@ -317,6 +317,11 @@ test-yaml:
 	@# Fall back to find if there is no git, e.g. in package builds
 	yamllint --strict $$((git ls-files "*.yml" "*.yaml" 2>/dev/null || find -name '*.y*ml') | grep -v ^dbicdh)
 
+.PHONY: test-shfmt
+test-shfmt:
+	@which shfmt >/dev/null 2>&1 || (echo "Command 'shfmt' not found, can not execute bash script syntax checks" && false)
+	shfmt -d .
```
perlpunk, line 323: I suggest to do the following right now:
```suggestion
	shfmt -w -i 4 -bn -ci -sr $(shfmt -f .)
```
and remove the .editorconfig symlink.
We can think of any improvements later and get this big PR merged.

Martchus: Considering we were not able to make eslint/prettier ignore the editor config (after reading https://github.com/prettier/eslint-plugin-prettier, https://prettier.io/docs/en/configuration.html, https://prettier.io/docs/en/options.html and https://eslint.org/docs/latest/use/configure/configuration-files#using-configuration-files and trying several variants) that might be the best way to move this forward for now, indeed.

perlpunk: And apparently 
```
shfmt -d -i 4 -bn -ci -sr .
```
also works. Not sure why we ruled that out when we experimented with it in the scripts repo.

perlpunk: oh right, because of the `.bpan` directory which we wanted to ignore.
So we might need something like this, depending on the repo:
```
shfmt -d -i 4 -bn -ci -sr  $(shfmt -f . | grep -v '^.bpan' )
```

external/os-autoinst-common/.editorconfig
baierjan: What about the other way around?
```
[*.js]
ignore = true
```

Makefile
```diff
@@ -317,6 +317,11 @@ test-yaml:
 	@# Fall back to find if there is no git, e.g. in package builds
 	yamllint --strict $$((git ls-files "*.yml" "*.yaml" 2>/dev/null || find -name '*.y*ml') | grep -v ^dbicdh)
 
+.PHONY: test-shfmt
+test-shfmt:
+	@which shfmt >/dev/null 2>&1 || (echo "Command 'shfmt' not found, can not execute bash script syntax checks" && false)
+	shfmt -d -d -i 4 -bn -ci -sr $(shfmt -f . | grep -v '^.bpan' )
```
perlpunk, line 323: I think this doesn't work as it is a Makefile, where shell syntax doesn't work.
Also we don't have .bpan here in this repo, so maybe don't use it.
Try this:
```suggestion
	shfmt -d -d -i 4 -bn -ci -sr $$(shfmt -f .)
```
With the double dollar it should work.

Martchus (approved):
Looks like it ran successfully:

```
lib/OpenQA/Worker/Isotovideo/Client.pm source OK
lib/perlcritic/Perl/Critic/Policy/HashKeyQuotes.pm source OK
shfmt -d -i 4 -bn -ci -sr $(shfmt -f .)
make test-unit-and-integration TIMEOUT_M=20 PROVE_ARGS="$HARNESS t/*{tidy,compile}*.t" GLOBIGNORE="t/25-cache-service.t:t/43-scheduling-and-worker-scalability.t:t/ui/01-list.t:t/ui/26-jobs_restart.t:t/ui/13-admin.t:"
make[1]: Entering directory '/home/squamata/project'
```

And the JavaScript check didn't complain anymore as well.

perlpunk (approved):

PR #5450, opened 2024-02-02
Refactor methods which violate the deeply nested loops

b10n1k:
Simply extract parts of the subroutines which causes perlcritic to complain about deeply nested loops.
https://progress.opensuse.org/issues/138416

### DIFF
--- a/lib/OpenQA/Scheduler/Model/Jobs.pm
+++ b/lib/OpenQA/Scheduler/Model/Jobs.pm
@@ -101,21 +101,7 @@ sub _allocate_jobs ($self, $free_workers) {
                 my $prio = $j->{priority};    # we only consider the priority of the main job
                 for my $worker (keys %taken) {
                     my $ji = $taken{$worker};
-                    if ($prio > 0) {
-                        # this means we will by default increase the offset per half-assigned job,
-                        # so if we miss 1/25 jobs, we'll bump by +24
-                        log_debug "Discarding job $ji->{id} (with priority $prio) due to incomplete parallel cluster"
-                          . ', reducing priority by '
-                          . STARVATION_PROTECTION_PRIORITY_OFFSET;
-                        $j->{priority_offset} += STARVATION_PROTECTION_PRIORITY_OFFSET;
-                    }
-                    else {
-                        # don't "take" the worker, but make sure it's not
-                        # used for another job and stays around
-                        log_debug "Holding worker $worker for job $ji->{id} to avoid starvation";
-                        $allocated_workers->{$worker} = $ji->{id};
-                    }
-
+                    _allocate_worker_with_priority($prio, $ji, $j, $allocated_workers, $worker);
                 }
                 %taken = ();
                 last;
@@ -141,6 +127,22 @@ sub _allocate_jobs ($self, $free_workers) {
     return ($allocated_workers, $allocated_jobs);
 }
 
+sub _allocate_worker_with_priority ($prio, $ji, $j, $allocated_workers, $worker) {
+    if ($prio > 0) {
+        # this means we will by default increase the offset per half-assigned job,
+        # so if we miss 1/25 jobs, we'll bump by +24
+        log_debug "Discarding job $ji->{id} (with priority $prio) due to incomplete parallel cluster"
+          . ', reducing priority by '
+          . STARVATION_PROTECTION_PRIORITY_OFFSET;
+        $j->{priority_offset} += STARVATION_PROTECTION_PRIORITY_OFFSET;
+    }
+    else {
+        # don't "take" the worker, but make sure it's not
+        # used for another job and stays around
+        log_debug "Holding worker $worker for job $ji->{id} to avoid starvation";
+        $allocated_workers->{$worker} = $ji->{id};
+    }
+}
 
 sub schedule ($self) {
     my $start_time = time;
--- a/lib/OpenQA/Schema/Result/JobGroups.pm
+++ b/lib/OpenQA/Schema/Result/JobGroups.pm
@@ -338,26 +338,30 @@ sub to_template {
 
         foreach my $product (sort keys %{$group{scenarios}->{$arch}}) {
             my @scenarios;
-            foreach my $test_suite (@{$group{scenarios}->{$arch}->{$product}}) {
-                foreach my $name (sort keys %$test_suite) {
-                    my $attr = $test_suite->{$name};
-                    if ($attr->{machine} eq $default_machine) {
-                        delete $attr->{machine} if $test_suites{$arch}{$name} == 1;
-                    }
-                    if (keys %$attr) {
-                        $test_suite->{$name} = $attr;
-                        push @scenarios, $test_suite;
-                    }
-                    else {
-                        push @scenarios, $name;
-                    }
-                }
-            }
-            $group{scenarios}{$arch}{$product} = \@scenarios;
+            _remove_test_suite_defaults($product, $default_machine, $arch, \%group, \%test_suites, \@scenarios);
         }
     }
 
     return \%group;
+}
+
+sub _remove_test_suite_defaults ($product, $default_machine, $arch, $group, $test_suites, $scenarios) {
+    foreach my $test_suite (@{$group->{scenarios}->{$arch}->{$product}}) {
+        foreach my $name (sort keys %$test_suite) {
+            my $attr = $test_suite->{$name};
+            if ($attr->{machine} eq $default_machine) {
+                delete $attr->{machine} if $test_suites->{$arch}{$name} == 1;
+            }
+            if (keys %$attr) {
+                $test_suite->{$name} = $attr;
+                push @$scenarios, $test_suite;
+            }
+            else {
+                push @$scenarios, $name;
+            }
+        }
+    }
+    $group->{scenarios}{$arch}{$product} = $scenarios;
 }
 
 sub to_yaml {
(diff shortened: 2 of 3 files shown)


lib/OpenQA/Scheduler/Model/Jobs.pm
```diff
@@ -141,6 +127,23 @@ sub _allocate_jobs ($self, $free_workers) {
     return ($allocated_workers, $allocated_jobs);
 }
 
+sub _allocate_worker_with_priority {
+    my ($prio, $ji, $j, $allocated_workers, $worker) = @_;
```
Martchus, line 131: Please use signatures in new functions:
```suggestion
sub _allocate_worker_with_priority ($prio, $ji, $j, $allocated_workers, $worker) {
```

lib/OpenQA/Schema/Result/JobGroups.pm
```diff
@@ -338,28 +338,33 @@ sub to_template {
...
-            $group{scenarios}{$arch}{$product} = \@scenarios;
+            _remove_test_suite_defaults($product, $default_machine, $arch, \%group, \%test_suites, @scenarios);
         }
     }
 
     return \%group;
 }
 
+sub _remove_test_suite_defaults {
+    my ($product, $default_machine, $arch, %group, %test_suites, @scenarios) = @_;
```
Martchus, line 349: That's not going to work. You cannot initialize multiple hashes and arrays from the argument list like that. Also, use signatures.

```suggestion
sub _remove_test_suite_defaults ($product, $default_machine, $arch, $group, $test_suites, $scenarios) {
```

Of course this means you need to adjust the calling site to pass the hashes/arrays not directly but only a reference. The code inside the function also needs to be adjusted accordingly.

lib/OpenQA/Schema/Result/JobGroups.pm
```diff
@@ -382,68 +387,92 @@ sub template_data_from_yaml {
...
+                $testsuite_name,
+                $settings,
+                $description);
         }
     }
+}
 
-    return \%job_template_names;
+sub _parse_job_template_machines {
+    my ($machine_names, %job_template_names, $prio, $arch, $product_name, $yaml_products, $job_template_name, $testsuite_name, $settings, $description) = @_;
```
Martchus, line 454: Same problem as mentioned for the other occurrence.

Martchus (changes requested):

lib/OpenQA/Schema/Result/JobGroups.pm
```diff
@@ -382,68 +386,91 @@ sub template_data_from_yaml {
...
     }
+}
 
-    return \%job_template_names;
+sub _parse_job_template_machines ($machine_names, $job_template_names, $prio, $arch, $product_name, $yaml_products, $job_template_name, $testsuite_name, $settings, $description) {
+    # my ($machine_names, %job_template_names, $prio, $arch, $product_name, $yaml_products, $job_template_name, $testsuite_name, $settings, $description) = @_;
+    foreach my $machine_name (@{$machine_names}) {
+        my $job_template_key
+          = $arch . $product_name . $machine_name . ($testsuite_name // '') . ($job_template_name // '');
+        if ($$job_template_names{$job_template_key}) {
```
Martchus, line 456:
```suggestion
        if ($job_template_names->{$job_template_key}) {
```

lib/OpenQA/Schema/ResultSet/Jobs.pm
```diff
@@ -203,6 +193,20 @@ sub create_from_settings {
     return $job;
 }
 
+sub _handle_directly_chained_dep($parent_worker_class, $id, $settings) {
+    if ($parent_worker_class = $parent_worker_class ? $parent_worker_class->value : '') {
+        if (!$$settings->{WORKER_CLASS}) {
```
Martchus, line 198:
```suggestion
        if (!$settings->{WORKER_CLASS}) {
```

lib/OpenQA/Schema/ResultSet/Jobs.pm
```diff
@@ -203,6 +193,20 @@ sub create_from_settings {
...
+            # assume we want to use the worker class from the parent here (and not the default which
+            # is otherwise assumed)
+            $$settings->{WORKER_CLASS} = $parent_worker_class;
+        }
+        elsif ($$settings->{WORKER_CLASS} ne $parent_worker_class) {
```
Martchus, line 203:
```suggestion
        elsif ($settings->{WORKER_CLASS} ne $parent_worker_class) {
```

lib/OpenQA/Schema/ResultSet/Jobs.pm
```diff
@@ -203,6 +193,20 @@ sub create_from_settings {
...
+            die "Specified WORKER_CLASS ($$settings->{WORKER_CLASS}) does not match the one from"
```
Martchus, line 204:
```suggestion
            die "Specified WORKER_CLASS ($settings->{WORKER_CLASS}) does not match the one from"
```

lib/OpenQA/Scheduler/Model/Jobs.pm
```diff
@@ -101,21 +101,7 @@ sub _allocate_jobs ($self, $free_workers) {
...
-                    }
-                    else {
-                        # don't "take" the worker, but make sure it's not
-                        # used for another job and stays around
-                        log_debug "Holding worker $worker for job $ji->{id} to avoid starvation";
-                        $allocated_workers->{$worker} = $ji->{id};
-                    }
-
+                    _allocate_worker_with_priority($prio, $ji, $j, $allocated_workers, $worker);
                 }
```
okurz, line 105: Now maybe we can go one step further

```perl
_allocate_worker_with_priority($prio, $taken{$_}, $j, $allocated_workers, $_) for keys %taken;
```

lib/OpenQA/Schema/ResultSet/Jobs.pm
```diff
@@ -203,6 +193,20 @@ sub create_from_settings {
...
+        if (!$settings->{WORKER_CLASS}) {
+            # assume we want to use the worker class from the parent here (and not the default which
+            # is otherwise assumed)
+            $settings->{WORKER_CLASS} = $parent_worker_class;
+        }
+        elsif ($settings->{WORKER_CLASS} ne $parent_worker_class) {
+            die "Specified WORKER_CLASS ($settings->{WORKER_CLASS}) does not match the one from"
+              . " directly chained parent $id ($parent_worker_class)";
+        }
+    }
```
okurz, line 207:
```suggestion
    return undef unless $parent_worker_class = $parent_worker_class ? $parent_worker_class->value : '';
    if (!$settings->{WORKER_CLASS}) {
        # assume we want to use the worker class from the parent here (and not the default which
        # is otherwise assumed)
        $settings->{WORKER_CLASS} = $parent_worker_class;
    }
    elsif ($settings->{WORKER_CLASS} ne $parent_worker_class) {
        die "Specified WORKER_CLASS ($settings->{WORKER_CLASS}) does not match the one from"
          . " directly chained parent $id ($parent_worker_class)";
    }
```

though I am not sure if the `… unless $foo = $foo ? … : '';` works syntax-wise.

Martchus: I guess it'll work but maybe it makes sense to do additional changes like this one separately.

okurz: sure, @b10n1k you can leave that out at your discretion

Martchus (approved):

lib/OpenQA/Scheduler/Model/Jobs.pm
b10n1k: no sure. I see `$ji` in use below in the code

lib/OpenQA/Schema/ResultSet/Jobs.pm
b10n1k: I could try that. But i am afraid that it will need additional tests.

perlpunk: More important than syntax wise if if it is readable, and I think it's not.

okurz: nevermind, leave it out for now.

perlpunk:
It would be useful to introduce changes step by step and rerun the failing test, to see which change is the culprit.

lib/OpenQA/Schema/Result/JobGroups.pm
```diff
@@ -382,68 +386,91 @@ sub template_data_from_yaml {
...
 
-    return \%job_template_names;
+sub _parse_job_template_machines ($machine_names, $job_template_names, $prio, $arch, $product_name, $yaml_products, $job_template_name, $testsuite_name, $settings, $description) {
+    # my ($machine_names, %job_template_names, $prio, $arch, $product_name, $yaml_products, $job_template_name, $testsuite_name, $settings, $description) = @_;
+    foreach my $machine_name (@{$machine_names}) {
+        my $job_template_key
+          = $arch . $product_name . $machine_name . ($testsuite_name // '') . ($job_template_name // '');
+        if ($job_template_names->{$job_template_key}) {
+            my $name = $job_template_name // $testsuite_name;
+            my $error = "Job template name '$name' is defined more than once. "
```
perlpunk, line 458: The `to_yaml` method was previously returning this error.
Now you are returning this in a nested function, so you would need to catch it where you call it and pass it on.
For that exceptions are good., e.g. `die` and then catch it with `eval` or `try/catch`.
You could move this change out of the PR, so that the rest can be approved and merged.

perlpunk: You just pushed a change a minute after my comment that fixes it. Since this is always returning a hash ref and passing it on, the `die` approach is not needed.

okurz:
still failed tidy. Please tidy and consider to mark your PR as ready for review

Martchus (approved):

Martchus (approved):

perlpunk (approved):

okurz (approved):

PR #5452, opened 2024-02-06
docs: Explain combination of worker class values and more examples

okurz:


docs/WritingTests.asciidoc
```diff
@@ -497,6 +501,21 @@ WORKER_CLASS = desktop
...
+Worker class values can also be set to additionally qualify workers or worker
+instances for informational purposes, for example region and location tags
+based on company conventions:
+
+[caption="File: "]
+.workers.ini
+[source,ini]
+--------------------------------------------------------------------------------
+[global]
+WORKER_CLASS = planent-earth,continent-antarctica,location-my_station
```
perlpunk, line 518:
```suggestion
WORKER_CLASS = planet-earth,continent-antarctica,location-my_station
```

foursixnine:
Nice example! :)... now we need to come back to the federation, and have either quadrants or galaxy superclusters (not kidding)

lgtm

docs/WritingTests.asciidoc
```diff
@@ -477,13 +477,17 @@ sub run () {
 
 By default, any worker can get any job with the matching architecture.
 
-This behavior can be changed by setting job variable `WORKER_CLASS`. Jobs with
-this variable set (typically via machines or test suites configuration) are
-assigned only to workers, which have the same variable in the configuration
-file.
+This behavior can be changed by setting the job variable `WORKER_CLASS` taking
+a comma-separated list of worker class values. The values from multiple places
+of definition are combined. Typically machines and test suite configurations
```
schlad, line 482: 'of definition'? 
How about something like this instead: 
If the variable `WORKER_CLASS`, for a given job, is set in multiple places, the values are combined.

Is that what you try to say here? Sorry if I get this wrong; the sentence "The values from multiple places
of definition are combined." confuses me somehow.

docs/WritingTests.asciidoc
```diff
@@ -477,13 +477,17 @@ sub run () {
...
+set the worker class. Jobs with this variable set are assigned only to
+workers, which have all corresponding worker class values in their
+configuration (and-combination).
 
 For example, the following configuration ensures, that jobs with
 `WORKER_CLASS=desktop` can be assigned _only_ to worker instances 1 and 2.
```
mdoucha, line 488: Job with `WORKER_CLASS=desktop` can now be assigned to workers 1, 2 and 4.

docs/WritingTests.asciidoc
mdoucha: While it may be difficult to understand, "values from multiple places of definition" is a correct wording for what the sentence is trying to say.

Martchus (changes requested):
Just fixed the typo mentioned by @perlpunk.

docs/WritingTests.asciidoc
okurz: Hm, we could say "The values are combined from multiple places where defined". Would that be better?

docs/WritingTests.asciidoc
okurz: I deleted instance 4 again and extended instance 2 to have "desktop,foo,bar" so the first sentence of explanation can stay as in before

Martchus (approved):

schlad (approved):

foursixnine (approved):

PR #5454, opened 2024-02-06
RFC: Simplify the workflow for dealing with external assets

Martchus:
* Add all external assets directly to our Git repository so no additional downloads are necessary when starting the web application for the first time
* Adjust `.gitignore` accordingly
* Remove/update relevant documentation
* Clarify licensing of our code vs. third-party code in the README
* WIP: Add licenses for third-party code now present in our repository * Specify the files the LICENSE applies to at the beginning of those files * Add MIT licenses individually because each MIT license is technically a distinct license * SO FAR ONLY TWO LICENSES HAVE BEEN ADDED; this should be enough for demonstrational purposes
* See https://progress.opensuse.org/issues/153427


docs/Contributing.asciidoc
```diff
@@ -380,10 +380,13 @@ Also find more details in
 == Handling of dependencies
 
 === Javascript and CSS
-Add 3rd party JavaScript and CSS file to `assets/assetpack.def`. When
-restarting the web server the new/updated files are pulled automatically. Also
-take care to <<Contributing.asciidoc#update-asset-cache,update the asset cache
-for the openSUSE RPM package>>.
+Add thrid-party JavaScript and CSS file to `assets/assetpack.def`. When
```
okurz, line 383: s/thrid/third/

docs/Contributing.asciidoc
```diff
@@ -380,10 +380,13 @@ Also find more details in
...
+restarting the web server the new/updated files are downloaded automatically.
+
+Make sure to add those files to Git when committing your changes. Additionally,
+add the license(s) for the newly added thrid-party code to the root directory of
```
okurz, line 387: Same

okurz:
Seeing this PR I wonder about two points:
1. So when starting openQA from git then the assets are downloaded automatically?
2. Can we use npm to at least cover some parts of the asset management and then only store the non-npm controlled JavaScript assets manually directly in git?

Martchus:
> So when starting openQA from git then the assets are downloaded automatically?

With this change asset pack will find all assets already in place and thus skip the download.

> Can we use npm to at least cover some parts of the asset management and then only store the non-npm controlled JavaScript assets manually directly in git?

Maybe that's a good idea, indeed. Of course npm will not put files where we need them but a simple script could copy files to the right location.

Martchus:
Closing in favor of #5463 which is more promising.

PR #5456, opened 2024-02-06
Adapt usage of tools/update-deps

perlpunk:
The cpanfile now needs to be added on the commandline explicitly

Also fix the develop cpanfile target

The script had been fixed, but not our yaml file yet.

### DIFF
--- a/cpanfile
+++ b/cpanfile
@@ -115,7 +115,7 @@ on 'test' => sub {
 
 };
 
-on 'devel' => sub {
+on 'develop' => sub {
     requires 'Perl::Tidy', '== 20230912.0.0';
 
 };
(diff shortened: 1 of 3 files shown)


Martchus (approved):

cpanfile
```diff
@@ -115,8 +115,7 @@ on 'test' => sub {
 
 };
 
-on 'devel' => sub {
-    requires 'Perl::Tidy', '== 20230912.0.0';
```
perlpunk, line 119: I don't know why it removed Perl::Tidy here.
I really hope next time we can do everything in smaller steps.
The usage of update-deps changed, but I wasn't aware of this.
I will debug now

perlpunk: Ah, I forgot to rename the target to develop here. Should be fixed now

okurz (approved):

perlpunk:
Merging manually. Somehow OBS checks are missing

PR #5458, opened 2024-02-07
Ensure the worker can call `du` for space aware cleanup of Git caching

Martchus:
See https://progress.opensuse.org/issues/154156


okurz (approved):

okurz:
I recorded the UI unit test failure in https://progress.opensuse.org/issues/155068

PR #5463, opened 2024-02-08
Install assets via npm where possible

Martchus:
* See https://progress.opensuse.org/issues/153427 and the particular commit messages for details.
* This adds `bootstrap-4-multi-dropdown-navbar` to our local checkout as this is the only dependency not available via npm (it is not a big dependency).
* ~~I still have to try out whether I can not use https://github.com/openSUSE/obs-service-node_modules. Maybe I can test this in the OBS project automatically setup for CI checks.~~
* This is an alternative to https://github.com/os-autoinst/openQA/pull/5454.


Martchus:
It doesn't look like that would help. The service fails on OBS with:

```
service node_modules failed:
time="2024-02-08T14:58:59Z" level=warning msg="Path \"/etc/SUSEConnect\" from \"/etc/containers/mounts.conf\" doesn't exist, skipping"
time="2024-02-08T14:58:59Z" level=warning msg="Path \"/etc/zypp/credentials.d/SCCcredentials\" from \"/etc/containers/mounts.conf\" doesn't exist, skipping"
Running /usr/lib/obs/service//node_modules --cpio node_modules.obscpio --output node_modules.spec.inc --source-offset 1000 --outdir /var/cache/obs/rTcaV1NuWAUs/out
Traceback (most recent call last):
  File "/usr/lib/obs/service//node_modules", line 624, in <module>
    sys.exit(main(args))
  File "/usr/lib/obs/service//node_modules", line 372, in main
    with open(input_file) as fh:
TypeError: expected str, bytes or os.PathLike object, not NoneType
```
(https://build.opensuse.org/package/live_build_log/devel:openQA:GitHub:os-autoinst:openQA:PR-5463/openQA/openSUSE_Tumbleweed/x86_64)

I copied the relevant section in the `_service` file from cockpid. The only adjustment I made was removing `mode="manual"` because otherwise the service wouldn't actually run on OBS. So it seems like cockpit is also not really downloading these files on the fly on OBS.

That means this whole approach doesn't solve the original problem of making assets available to build the rpm package. (It is generally still nice to use npm.)

assets/3rdparty/bootstrap-4-multi-dropdown-navbar/beta2.0/css/bootstrap-4-navbar.css
okurz, line 1: does including this file mean that it's not available from a proper package management system?

okurz (commented):
The current state looks very clean. I would be ok if you can get this to work with the combined approach to reference node modules while also putting non-node-module code directly into the repo. For this proof of concept it's ok but maybe it's already easier for you to create a separate PR only using node modules additionally to the still present assets cache?

assets/3rdparty/bootstrap-4-multi-dropdown-navbar/beta2.0/css/bootstrap-4-navbar.css
Martchus: yes

Martchus:
> For this proof of concept it's ok but maybe it's already easier for you to create a separate PR only using node modules additionally to the still present assets cache?

It is either getting the dependency via npm or via a normal download. One cannot specify multiple sources for one file in asset pack (at least I wouldn't know how). Of course it would be possible to create a script which would make some symlinks in the asset cache directory to files installed via npm but that might be a bit hacky.

Makefile
```diff
@@ -320,7 +324,7 @@ test-yaml:
 .PHONY: test-shfmt
 test-shfmt:
 	@which shfmt >/dev/null 2>&1 || (echo "Command 'shfmt' not found, can not execute bash script syntax checks" && false)
-	shfmt -d -i 4 -bn -ci -sr $$(shfmt -f .)
+	shfmt -d -i 4 -bn -ci -sr $$(shfmt -f . | grep --invert-match node_modules)
```
okurz, line 327: How about the same approach as for "test-shellcheck" to explicitly select shell files by type

Martchus: I didn't like the `shfmt -f .`-approach anyway, e.g. in my unclean repo it just stopped after a permission error and thus didn't check much. I made it a separate commit.

okurz (approved):

Martchus:
I tested the Tumbleweed package for commit 07b3e6b locally and it works. (No more messages about missing files at startup and the UI looks as expected.)

I also pushed another commit to temporarily use https://build.opensuse.org/package/show/devel:openQA:Leap:15.6/local-npm-registry for the SLE-15-SP6 build.

With that I hope this is now good to merge. Once it has been merged I can re-trigger the build of our CI container and remove the manual installation of npm for the caching step on CircleCI. Then I'll also have a look on our Tumbleweed submissions.

README.asciidoc
```diff
@@ -56,3 +56,9 @@ http://open.qa/contact/[our contact page].
 openQA is developed on a continuous base where every commit in the git master
 branch is considered stable and a valid and installable version. The old tags
 on github are therefore misleading.
+
+== License
+All code is licensed under link:COPYING[GPL-2-or-later] unless stated otherwise
+in particular files. This does *not* apply to code found in the
+`assets/3rdparty` directory. It contains third-party code and the relevant
+licenses can be found in the root directory of this code repository.
```
kalikiana, line 64: I wouldn't add it here, but just for the future [REUSE](https://reuse.software/) would help with documenting and checking licenses for the whole repo

kalikiana (approved):

PR #5464, opened 2024-02-08
Ensure all hidden form fields in test overview filter

perlpunk:
Issue: https://progress.opensuse.org/issues/154537

The code was skipping elements that already existed in the form, and this resulted in elements with multiple values to only keep one value. Now the elements to be added are collected and only added to the form after the loop.

Bug was introduced in 2f6b7e0fd

### DIFF
--- a/assets/javascripts/filter_form.js
+++ b/assets/javascripts/filter_form.js
@@ -31,6 +31,7 @@ function setupFilterForm(options) {
 function parseFilterArguments(paramHandler) {
   var varPairs = window.location.search.substring(1).split('&');
   var filterLabels = [];
+  var hiddenInputs = [];
   for (var j = 0; j < varPairs.length; ++j) {
     var pair = varPairs[j].split('=');
     if (pair.length > 1) {
@@ -47,9 +48,12 @@ function parseFilterArguments(paramHandler) {
         input.attr('value', val);
         input.attr('name', key);
         input.attr('hidden', true);
-        $('#filter-form').append(input);
+        hiddenInputs.push(input);
       }
     }
+  }
+  for (var i = 0; i < hiddenInputs.length; i++) {
+    $('#filter-form').append(hiddenInputs[i]);
   }
   if (filterLabels.length > 0) {
     $('#filter-panel .card-header')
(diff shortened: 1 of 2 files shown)


assets/javascripts/filter_form.js
```diff
@@ -47,10 +48,13 @@ function parseFilterArguments(paramHandler) {
...
         input.attr('name', key);
         input.attr('hidden', true);
-        $('#filter-form').append(input);
+        hiddenInputs.push(input);
       }
     }
   }
+  for (var i = 0; i < hiddenInputs.length; i++) {
+    $('#filter-form').append(hiddenInputs[i]);
+  }
```
Martchus, line 57: I would avoid the repeated calls to `$('#filter-form')` in the loop and stick to standard JavaScript (as we probably shouldn't introduce more jQuery code at this point):
```suggestion
  const filterForm = document.getElementById('filter-form');
  hiddenInputs.forEach(i => filterForm.appendChild(i));
```

Probably the code that creates the input elements should also just use `document.createElement('input')` and normal DOM APIs.

It is not obvious at all why your change makes a difference but it is because of what the code passed via `paramHandler` does. That means you could probably also just change *that* code and make it skip those elements, e.g. replace `if (formElement) {` with `if (formElement && !formElement.hidden) {`. This would avoid an extra array and loop.

perlpunk: I tried now:
```
if (formElement && !formElement.hidden)
```
but then it is always skipping the last distri mentioned in the query parameters...
I need to debug this.

perlpunk: So apparently in the first call of the `paramHandler` `form[key]` is undefined. In the second it''s returning `HTMLInputElement` (hidden). After that it returns a `RadioNodeList`. Possibly because there are now two elements already. Not sure yet how I should handle that.

perlpunk: Also I would like to add that I didn't introduce any new jquery code. I just mvoed it from one loop to the other.

okurz (approved):

assets/javascripts/filter_form.js
Martchus: Ok, then your approach might be simpler after all. And it is true that the code was just moved.

Martchus (approved):

PR #5471, opened 2024-02-13
Make developer mode work in single-container setup

Martchus:
* Configure the service port delta to zero so the connection to the live view handler is still made via the reverse proxy despite the reverse proxy being exposed on a non-standard port
* Keep docker-compose-based setup as-is because it uses standard ports and also exposes the ports of the individual services
* See https://progress.opensuse.org/issues/153499

---

I tested this manually by building the image locally and then invoking the command from our documentation with my local image. It works (and with the current image the issue was reproducible).

### DIFF
--- a/container/single-instance/Dockerfile
+++ b/container/single-instance/Dockerfile
@@ -17,7 +17,8 @@ LABEL org.opencontainers.image.created="%BUILDTIME%"
 # hadolint ignore=DL3037
 RUN zypper in -y openQA-single-instance openQA-bootstrap \
     qemu-arm qemu-ppc qemu-x86 qemu-tools sudo iputils os-autoinst-distri-opensuse-deps && \
-    zypper clean -a
+    zypper clean -a && \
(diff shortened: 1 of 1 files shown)


kalikiana (approved):

okurz (approved):
ok. We should add a test though

okurz (changes requested):
Actually, sorry, need to rethink. Why should we patch the port in the Dockerfile when the Dockerfile never sets any custom port? Somehow this feels like not the right place. Anyone building a custom container just wrapping openqa-bootstrap would be affected the same, isn't it?

okurz (approved):
I understand this better now and https://github.com/os-autoinst/openQA/pull/5435 is strongly related and also not that recent bringing in service_port_delta. So this is just the logical consequential change and I agree. We can also consider changing the actual low-level default to use a service_port_delta of 0 but we should make that another explicit pull request even if the change would only affect openQA developers.

PR #5472, opened 2024-02-13
Improve out-of-box experience with single-instance container

Martchus:
* Give the container a name for easier reference in further explanations
* Add KVM support; this is very useful as many jobs in our instance seem
  to use `QEMUCPU=host` and would therefore otherwise fail because this
  setting requires KVM support
* Provide an example command for cloning a job and for making an API query;
  otherwise it isn't clear how to proceed with a containerized setup like
  that after exploring the web UI
* Explain how to provide test code in a containerized setup like this
* See https://progress.opensuse.org/issues/153499

### DIFF
--- a/docs/Installing.asciidoc
+++ b/docs/Installing.asciidoc
@@ -22,7 +22,7 @@ repository].
 
 Continue with the next section <<#container_setup,Container based setup>> to
 setup a simple, ready-to-use container based openQA instance which is useful
-for a single user setup. For a quick boostrapping under openSUSE go to
+for a single user setup. For a quick bootstrapping under openSUSE go to
 <<#bootstrapping,Quick bootstrapping>>. Else, continue with the more
 advanced section about <<#custom_installation,Custom installation>>. For a
 setup suitable to develop openQA itself, have a look at the
--- a/docs/Installing.asciidoc
+++ b/docs/Installing.asciidoc
@@ -40,7 +40,8 @@ single command line using the `podman` container engine is the following:
 
 [source,sh]
 ----
-podman run -p 1080:80 -p 1443:443 --rm -it registry.opensuse.org/devel/openqa/containers/openqa-single-instance
+podman run --name openqa --device /dev/kvm -p 1080:80 -p 1443:443 --rm -it \
+  registry.opensuse.org/devel/openqa/containers/openqa-single-instance
 ----
 
 Once the startup has finished, the web UI is accessible via http://localhost:1080


kalikiana (dismissed):

docs/Installing.asciidoc
```diff
@@ -40,11 +40,34 @@ single command line using the `podman` container engine is the following:
...
 [source,sh]
 ----
-podman run -p 1080:80 -p 1443:443 --rm -it registry.opensuse.org/devel/openqa/containers/openqa-single-instance
+podman run --name openqa --device /dev/kvm -p 1080:80 -p 1443:443 --rm -it \
+  registry.opensuse.org/devel/openqa/containers/openqa-single-instance
 ----
 
 Once the startup has finished, the web UI is accessible via http://localhost:1080
-or https://localhost:1443.
+or https://localhost:1443. You can login without authentication and create a API
```
okurz, line 48:
```suggestion
or https://localhost:1443. You can login without authentication and create an API
```

docs/Installing.asciidoc
```diff
@@ -40,11 +40,34 @@ single command line using the `podman` container engine is the following:
...
+  CASEDIR=https://….git NEEDLES_DIR=https://….git
+# make an API query
+podman exec openqa openqa-cli api --pretty jobs/1
+----
+
+NOTE: You need to provide test code to run tests. This example suggests you
+provide the URL to a Git repository via the `CASEDIR` variable (and possibly
+via `NEEDLES_DIR` if needles are in a separate repository). For development it
+may be easier to specify a local directory and make it available within the
+container via the `-v`-option.
```
okurz, line 70: I understand why this would be convenient to have here but https://github.com/os-autoinst/openQA/pull/5472/files#diff-71734bf4d91118f92743c7d0db9218c4f24729410e020110884606b5ad6a3878R132 states

```
For creating a first test job, checkout the
<<UsersGuide.asciidoc#_triggering_tests,Triggering tests section>>. Note that the
commands mentioned there can also be invoked within a container, e.g.:
[source,sh]
----
podman run \
   --rm -it registry.opensuse.org/devel/openqa/containers15.5/openqa_webui:latest \
   openqa-cli --help
----
```

With your section I feel this is becoming a too weird mix. Maybe the section you add fits better elsewhere or is better split. As yet another alternative we can consider a completely new "quickstart guide". Then I think it would be ok to duplicate content a bit.

okurz (changes requested):

docs/Installing.asciidoc
```diff
@@ -40,11 +40,34 @@ single command line using the `podman` container engine is the following:
...
 ----
-podman run -p 1080:80 -p 1443:443 --rm -it registry.opensuse.org/devel/openqa/containers/openqa-single-instance
+podman run --name openqa --device /dev/kvm -p 1080:80 -p 1443:443 --rm -it \
+  registry.opensuse.org/devel/openqa/containers/openqa-single-instance
 ----
 
 Once the startup has finished, the web UI is accessible via http://localhost:1080
-or https://localhost:1443.
+or https://localhost:1443. You can login without authentication and create a API
+key and secret on http://localhost:1080/api_keys.
```
okurz, line 49: To prevent stating only 1080 and not 1443 I suggest

```suggestion
key and secret on the available web UI.
```

docs/Installing.asciidoc
```diff
@@ -40,11 +40,34 @@ single command line using the `podman` container engine is the following:
...
+
+You can execute further tools like `openqa-cli` and `openqa-clone-job` via the
+same container. To clone a job from another openQA instance like
+https://openqa.opensuse.org[o3] you can run a command like this:
+[source,sh]
+----
+# make API key and secret available to subsequent commands
+podman exec openqa bash -c \
+  'echo -e "[localhost]\nkey = 1234567890ABCDEF\nsecret = 1234567890ABCDEF" > /etc/openqa/client.conf'
```
okurz, line 58: the bootstrap script already puts key+secret in the config for the worker so we shouldn't overwrite. If in doubt then leave out this code section for now

docs/Installing.asciidoc
```diff
@@ -40,11 +40,34 @@ single command line using the `podman` container engine is the following:
 
 [source,sh]
 ----
-podman run -p 1080:80 -p 1443:443 --rm -it registry.opensuse.org/devel/openqa/containers/openqa-single-instance
+podman run --name openqa --device /dev/kvm -p 1080:80 -p 1443:443 --rm -it \
```
okurz, line 43: sounds good. keep this as is

docs/Installing.asciidoc
okurz: After discussing with you in person I have the following suggestions:
1. I suggest to ditch the "NEEDLES_DIR" mention as that is an advanced use case and explained elsewhere. A standard distribution like os-autoinst-distri-opensuse should not need it
2. The `-v` part could go into another paragraph further down below after "Custom configuration for containers", e.g. "Persistent database and asset storage" (or leave it out for now)

okurz (changes requested):

docs/Installing.asciidoc
Martchus: I'll briefly look into https://progress.opensuse.org/issues/155473 because yes, this should not be required (but it was in my tests yesterday which is very weird).

Martchus:
I shortened this to the parts we hopefully can all agree on.

kalikiana (approved):

PR #5474, opened 2024-02-13
Remove manual installation of npm when caching assets

Martchus:
Just a draft because we need to merge #5473 first; only then the next paragraph is actually true.

---

This was only required for bootstrapping the switch to using npm. Now the container image we use for the CI has been rebuilt and contains the npm package.

Related ticket: https://progress.opensuse.org/issues/153427


okurz:
+1

kalikiana (approved):

PR #5478, opened 2024-02-14
Fix select dropdown arrow floating over text

asdil12:
Eg. by https://openqa.opensuse.org/tests/3937974#next_previous

Before:

After:

### DIFF
--- a/assets/stylesheets/forms.scss
+++ b/assets/stylesheets/forms.scss
@@ -172,3 +172,9 @@ select.form-control {
         background-color: $default-bg-color-darktheme;
     }
 }
+
+// fix drop-down arrow floating over text despite `with: auto;` from DataTable's upstream CSS
+div.dataTables_wrapper div.dataTables_length select {
+    background-position: right 0.2rem center;
+    padding: 0 1em 0 0.5em;
+}


assets/stylesheets/forms.scss
```diff
@@ -172,3 +172,9 @@ select.form-control {
         background-color: $default-bg-color-darktheme;
     }
 }
+
+// fix dropdown arrow floating over text
+.custom-select-sm {
+    background-position: right 0.2rem center;
+    padding: 0 1em 0 0.5em;
+}
```
Martchus, line 180: I was wondering why both - Firefox and Chromium - need such a hack. It looks like the problematic code comes from DataTable's bootstrap-specific CSS:

```
div.dataTables_wrapper div.dataTables_length select {
  width: auto;
  display: inline-block;
}
```

The `width: auto;` line causes the problem. It makes nevertheless sense to keep it because without it the box is too long. However, it would make sense to use the same selector as the other stylings and mention it in the comment:

```suggestion
// fix drop-down arrow floating over text despite `with: auto;` from DataTable's upstream CSS
div.dataTables_wrapper div.dataTables_length select {
    background-position: right 0.2rem center;
    padding: 0 1em 0 0.5em;
}
```

Martchus:
You probably also need to rebase. Or we currently have a more severe problem with the caching.

asdil12:
I guess the CI fails are unrelated due to some issues with the CI, as rebasing didn't help.

kalikiana (approved):

kalikiana:
@mergifyio rebase

Martchus (approved):

Martchus:
@Mergifyio rebase

PR #5481, opened 2024-02-15
Update subrepo external/os-autoinst-common

perlpunk:


Martchus (approved):

Martchus:
Annoyingly the commit message check is still not coping with subrepo's commit messages. For now we can force-merge if all other checks pass but we should probably revisit that. (And maybe there's a way to prevent subrepo from creating such horrible commit messages in the first place? The check is actually right to complain…)

perlpunk:
I edited the commit message for now.
I guess git-subrepo would need a configuration file to know when it should split a commit message title

okurz (approved):

PR #5482, opened 2024-02-20
Split monitoring of jobs via `openqa-cli` into a separate command

Martchus:
This allows to monitor a set of job which will be useful for further CI-related use cases like cloning existing jobs and monitoring them (see https://progress.opensuse.org/issues/130934).

Of course the existing `openqa-cli schedule` command will continue to support the `--monitor` parameter.

### DIFF
--- a/lib/OpenQA/CLI/monitor.pm
+++ b/lib/OpenQA/CLI/monitor.pm
@@ -13,7 +13,7 @@ has usage => sub { shift->extract_usage };
 sub _monitor_jobs ($self, $client, $poll_interval, $job_ids, $job_results) {
     while (@$job_results < @$job_ids) {
         my $job_id = $job_ids->[@$job_results];
-        my $tx = $client->build_tx(GET => $self->url_for("experimental/jobs/$job_id/status"), {});
+        my $tx = $client->build_tx(GET => $self->url_for("experimental/jobs/$job_id/status"));
         my $res = $self->retry_tx($client, $tx);
         return $res if $res != 0;
         my $job = $tx->res->json;
--- a/lib/OpenQA/CLI/monitor.pm
+++ b/lib/OpenQA/CLI/monitor.pm
@@ -1,0 +1,74 @@
+# Copyright SUSE LLC
+# SPDX-License-Identifier: GPL-2.0-or-later
+
+package OpenQA::CLI::monitor;
+use Mojo::Base 'OpenQA::Command', -signatures;
+
+use OpenQA::Jobs::Constants;
+use Mojo::Util qw(encode getopt);
+
+has description => 'Monitors a set of jobs';
+has usage => sub { shift->extract_usage };
+
+sub _monitor_jobs ($self, $client, $poll_interval, $job_ids, $job_results) {
+    while (@$job_results < @$job_ids) {
+        my $job_id = $job_ids->[@$job_results];
+        my $tx = $client->build_tx(GET => $self->url_for("experimental/jobs/$job_id/status"), {});
+        my $res = $self->retry_tx($client, $tx);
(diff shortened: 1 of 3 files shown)


lib/OpenQA/CLI/monitor.pm
```diff
@@ -0,0 +1,73 @@
...
+    $self->_monitor_and_return($self->client($self->url_for('tests')), $poll_interval, \@args);
+}
+
+1;
+
+=encoding utf8
+
+=head1 SYNOPSIS
+
+  Usage: openqa-cli monitor [OPTIONS] [job_id…]
```
kraih, line 55: Shouldn't that look more like `[OPTIONS] [JOB_IDS]`?

lib/OpenQA/CLI/monitor.pm
```diff
@@ -0,0 +1,73 @@
...
+        sleep $poll_interval;
+    }
+}
+
+sub _compute_return_code ($self, $job_results) {
+    for my $job_result (@$job_results) {
+        return 2 unless OpenQA::Jobs::Constants::is_ok_result($job_result);
+    }
+    return 0;
+}
```
Martchus, line 34: These functions are exactly as they were before in `schedule.pm`.

lib/OpenQA/CLI/monitor.pm
```diff
@@ -0,0 +1,73 @@
+# Copyright SUSE LLC
+# SPDX-License-Identifier: GPL-2.0-or-later
+
+package OpenQA::CLI::monitor;
+use OpenQA::Jobs::Constants;
+use Mojo::Base 'OpenQA::Command', -signatures;
```
kraih, line 6: Wrong order. Constants are imported before strictures are activated with `Mojo::Base`.

lib/OpenQA/CLI/monitor.pm
```diff
@@ -0,0 +1,73 @@
...
+        --apikey <key>             API key
+        --apisecret <secret>       API secret
+        --host <host>              Target host, defaults to http://localhost
+    -h, --help                     Show this summary of available options
+        --osd                      Set target host to http://openqa.suse.de
+        --o3                       Set target host to https://openqa.opensuse.org
+        --name <name>              Name of this client, used by openQA to
+                                   identify different clients via User-Agent
+                                   header, defaults to "openqa-cli"
+    -i, --poll-interval            Specifies the poll interval
```
kraih, line 68: Should probably be `--poll-interval <seconds>` or so.

lib/OpenQA/CLI/monitor.pm
kraih: Fair enough, i'm still reviewing this as if it was a new feature.

kraih (approved):

lib/OpenQA/CLI/monitor.pm
```diff
@@ -0,0 +1,74 @@
...
+        print encode('UTF-8', "Job state of job ID $job_id: $job_state, waiting …\n");
+        sleep $poll_interval;
+    }
+}
+
+sub _compute_return_code ($self, $job_results) {
+    for my $job_result (@$job_results) {
+        return 2 unless OpenQA::Jobs::Constants::is_ok_result($job_result);
+    }
+    return 0;
```
okurz, line 34:
```suggestion
use List::Util qw(all);
…
    all { OpenQA::Jobs::Constants::is_ok_result($_) } @$job_results ? 0 : 2;
```

okurz: But I will move that to a separate PR as you are just moving the code

okurz (approved):

PR #5492, opened 2024-02-28
Update Autoyast profile for nvme to include chrony

b10n1k:
https://progress.opensuse.org/issues/155305

### DIFF
--- a/contrib/ay-openqa-worker-leap-3-nvme.xml
+++ b/contrib/ay-openqa-worker-leap-3-nvme.xml
@@ -54,6 +54,7 @@
     </patterns>
     <packages t="list">
       <package>openssh</package>
+      <package>chrony</package>
       <package>sudo</package>
     </packages>
   </software>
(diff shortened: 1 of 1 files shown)


contrib/ay-openqa-worker-leap-3-nvme.xml
```diff
@@ -54,6 +54,7 @@
     </patterns>
     <packages t="list">
       <package>openssh</package>
+      <package>chrony</package>
```
kalikiana, line 57: So was this tested? Did you check if it is linked and documented somewhere?

okurz: @kalikiana I would trust that this is a trivial change. We have seen a validation for the .xml.erb profile already. Automatic validation will be covered as part of https://progress.opensuse.org/issues/156169 so out of scope for here.

okurz (approved):

PR #5497, opened 2024-03-01
Filter settings by result and groupid

d3flex:
This commit provides additional arguments to the job_settings API call in
order to get the ids of the jobs which, in additional the key and the its
value, it will match the job result per job_group. job_group is intended to be
a list for further grouping.
    
The proposal doesnt demostrate a complete solution.
    
Obviously the if statement is under further discussion.
An alternative approach could be a separate function, potential named
`blocked` but I am not sure for that particular name as it comes from
internal SUSE concept and it is not assisiated with openQA.

image

https://progress.opensuse.org/issues/131279

### DIFF
--- a/lib/OpenQA/Schema/ResultSet/JobSettings.pm
+++ b/lib/OpenQA/Schema/ResultSet/JobSettings.pm
@@ -13,6 +13,8 @@ sub jobs_for_setting ($self, $options) {
     $key_like =~ s/\*/\%/g;
     my $list_value = $options->{list_value};
     my $list_value_like = "%${list_value}%";
+    my $filter_results = $options->{filter_results} if $options->{filter_results};
+    my $groupids = $options->{gids} if $options->{gids};
 
     # Get the highest job id to limit the number of jobs that need to be considered (to improve performance)
     my $dbh = $self->result_source->schema->storage->dbh;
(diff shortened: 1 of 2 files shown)


lib/OpenQA/Schema/ResultSet/JobSettings.pm
```diff
@@ -25,10 +27,15 @@ sub jobs_for_setting ($self, $options) {
...
+	$sth = $dbh->prepare(
+        'SELECT job_id, value FROM job_settings INNER JOIN jobs ON jobs.id=job_settings.job_id WHERE key LIKE ? AND value LIKE ? AND jobs.result = ? AND jobs.group_id = ? AND job_id > ? ORDER BY job_settings.job_id DESC');
+	$sth->execute($key_like, $list_value_like, $filter_results, $groupids, $min_id);
+    } else {
+	$sth = $dbh->prepare(
         'SELECT job_id, value FROM job_settings WHERE key LIKE ? AND value LIKE ? AND job_id > ? ORDER BY job_id DESC');
-    $sth->execute($key_like, $list_value_like, $min_id);
-
+	$sth->execute($key_like, $list_value_like, $min_id);
+    }
```
foursixnine, line 38: The jobgroup id is irrelevant at this stage, if a job matches the search, but it is pre-filtered from the list of results, in the end will hide the data from the user.

Also be aware of the discussion on slack about this route [not returning all of the information](https://suse.slack.com/archives/C02CANHLANP/p1708342424627099?thread_ts=1708020678.269079&cid=C02CANHLANP)

lib/OpenQA/WebAPI/Controller/API/V1/JobSettings.pm
```diff
@@ -8,11 +8,18 @@ sub jobs ($self) {
     my $validation = $self->validation;
     $validation->required('key')->like(qr/^[\w\*]+$/);
     $validation->required('list_value')->like(qr/^\w+$/);
+    $validation->optional('results')->like(qr/^\w+$/);
+    $validation->optional('list_groupids')->like(qr/^[\w\*]+$/);
```
okurz, line 12: For other queries we support "group_glob" and "not_group_globs". Introduing "list_groupids" here looks like a new, inconsistent approach. Also see the second suggestion in https://progress.opensuse.org/issues/131279#Suggestions How about extending the current group_glob support into here?

As alternative, potentially even simpler, see https://progress.opensuse.org/issues/131279#Suggestions number 3 stating `"squads" could be mapped into openQA for example with special job settings, e.g. QE Core ensures to trigger all their tests with _SQUAD='QE Core' and then be able to filter by that`

okurz (changes requested):

lib/OpenQA/WebAPI/Controller/API/V1/JobSettings.pm
b10n1k: I dont see a straight way to invoke group_glob inside the lib/OpenQA/WebAPI/Controller/API/V1/JobSettings.pm. That is because the variables are tight inside [_compose_job_overview_search_args](https://github.com/os-autoinst/openQA/blob/69e8e802aadb9b4d330fb30e0701c8f6a84f5945/lib/OpenQA/WebAPI/Plugin/Helpers.pm#L394C5-L394C38).

Martchus: If the code cannot be reused it would at least make sense to keep the design/usage consistent from a user perspective. (And I think this is also mainly what @okurz was asking for.)

PR #5499, opened 2024-03-04
Avoid syncing tests via the cache service when using Git anyway

Martchus:
* Skip syncing the tests directory (usually `/var/lib/openqa/share/tests`) via the cache service when the job is using Git anyway for its `CASEDIR` (and if separate its `NEEDLES_DIR`).
* Keep the sync enabled in the absence of `NEEDLES_DIR` as we don't know whether needles are part of `CASEDIR` or in a separate repository then
    * Note that in the first attempt (commit e45bc7716f) it was assumed that the absence of `NEEDLES_DIR` means needles are part of `CASEDIR` but that's not true in practice, e.g. investigation jobs are only overriding `CASEDIR` and not `NEEDLES_DIR` for the openSUSE test distribution and without syncing would run into `/var/lib/openqa/share/tests/opensuse/needles does not exist `.
    * This is what the 2nd test case is testing. In contrast to the first attempt the check now asserts that the sync is done (instead of being prevented).
* See https://progress.opensuse.org/issues/154237

### DIFF
--- a/t/24-worker-engine.t
+++ b/t/24-worker-engine.t
@@ -32,6 +32,7 @@ use OpenQA::Utils qw(testcasedir productdir needledir locate_asset base_host);
     use Mojo::Base -base;
     has id => 42;
     has worker => undef;
+    has client => sub { Test::FakeClient->new };
     sub post_setup_status { 1 }
     sub is_stopped_or_stopping { 0 }
 }
@@ -52,6 +53,7 @@ use OpenQA::Utils qw(testcasedir productdir needledir locate_asset base_host);
     has worker_id => 1;
     has webui_host => 'localhost';
     has service_port_delta => 2;
+    has testpool_server => 'fake-testpool-server';
 }
 
 $ENV{OPENQA_CONFIG} = "$FindBin::Bin/data/24-worker-overall";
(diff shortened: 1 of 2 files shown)


okurz (approved):
LGTM. Added "not-ready". After second approval please remove the "not-ready" label at a time convenient to you and monitor the effect after deployment on o3 to catch cases like in the original PR quickly.

t/24-worker-engine.t
```diff
@@ -227,6 +230,34 @@ subtest 'problems when caching assets' => sub {
...
+        $isotovideo_mock->redefine(sync_tests => sub (@) { $sync_tests_called = 1 });
+
+        OpenQA::Worker::Engines::isotovideo::do_asset_caching(@args);
+        is $error, undef, 'callback called without error (1)';
+        is $sync_tests_called, undef, 'sync tests was not called for Git-only job (separate needles repo)';
+
+        delete $vars{NEEDLES_DIR};    # no NEEDLES_DIR means needles are within CASEDIR or separate
+        OpenQA::Worker::Engines::isotovideo::do_asset_caching(@args);
+        is $error, undef, 'callback called without error (2)';
+        is $sync_tests_called, 1, 'sync tests was not called for job where needles might be separate';
```
perlpunk, line 247: Here sync tests *is* actually called.
```suggestion
        is $sync_tests_called, 1, 'sync tests was called for job where needles might be separate';
```

perlpunk (approved):

PR #5500, opened 2024-03-04
Make it less likely for jobs incompleting with `Cache … queue … full`

Martchus:
* Introduce a hard-limit for inactive Minion jobs
* Make the hard-limit slightly higher than the normal limit
* Enforce only the hard-limit on openQA job setup; so when the normal Minion queue limit was reached after a job had already been picked-up it can still continue to create Minion jobs instead of ending up as incomplete
* See https://progress.opensuse.org/issues/128276

---

Still a draft because tests are missing.

### DIFF
--- a/lib/OpenQA/CacheService/Response.pm
+++ b/lib/OpenQA/CacheService/Response.pm
@@ -5,7 +5,14 @@ package OpenQA::CacheService::Response;
 use Mojo::Base -base, -signatures;
 
 has [qw(data error)];
+
+# define soft-limit for inactive Minion jobs (enforced when determining idle worker slots availability)
 has max_inactive_jobs => $ENV{OPENQA_CACHE_MAX_INACTIVE_JOBS} // 5;
+
+# define hard-limit for inactive Minion jobs (enforced on setup of already started openQA job)
+has max_inactive_jobs_hard_limit => sub ($self) {
+    $ENV{OPENQA_CACHE_MAX_INACTIVE_JOBS_HARD_LIMIT} // ($self->max_inactive_jobs + 40);
+};
 
 sub has_error ($self) { !!$self->error }
 
--- a/lib/OpenQA/CacheService/Response/Info.pm
+++ b/lib/OpenQA/CacheService/Response/Info.pm
@@ -12,17 +12,19 @@ sub available_workers ($self) {
     return $data->{active_workers} != 0 || $data->{inactive_workers} != 0;
 }
 
-sub inactive_jobs_exceeded ($self) {
-    return undef if $self->max_inactive_jobs < 0;
+sub inactive_jobs_exceeded ($self, $options = {}) {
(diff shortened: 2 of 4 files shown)


lib/OpenQA/CacheService/Response.pm
```diff
@@ -5,8 +5,15 @@ package OpenQA::CacheService::Response;
 use Mojo::Base -base, -signatures;
 
 has [qw(data error)];
+
+# define soft-limit for inactive Minion jobs (enforced when determining idle worker slots availability)
 has max_inactive_jobs => $ENV{OPENQA_CACHE_MAX_INACTIVE_JOBS} // 5;
 
+# define hard-limit for inactive Minion jobs (enforced on setup of already started openQA job)
+has max_inactive_jobs_hard_limit => sub ($self) {
+    $ENV{OPENQA_CACHE_MAX_INACTIVE_JOBS_HARD_LIMIT} // ($self->max_inactive_jobs + 10);
```
okurz, line 14: In https://progress.opensuse.org/issues/128276#Suggestions we suggested "cache_queue_limit_concurrent" = 3, "cache_queue_limit_max" = 50, so should we make the above 5 a 3 and then here make the 10 a 40?

okurz (changes requested):
There might be more points. Just had one question for a start. Will look into more later.

lib/OpenQA/CacheService/Response.pm
Martchus: I don't think we should go as low as 3; also the limit of 5 is quite low. Making the hard limit a 40 would be ok, though.

okurz (approved):

lib/OpenQA/CacheService/Response.pm
```diff
@@ -5,8 +5,15 @@ package OpenQA::CacheService::Response;
 use Mojo::Base -base, -signatures;
 
 has [qw(data error)];
+
+# define soft-limit for inactive Minion jobs (enforced when determining idle worker slots availability)
 has max_inactive_jobs => $ENV{OPENQA_CACHE_MAX_INACTIVE_JOBS} // 5;
 
+# define hard-limit for inactive Minion jobs (enforced on setup of already started openQA job)
+has max_inactive_jobs_hard_limit => sub ($self) {
+    $ENV{OPENQA_CACHE_MAX_INACTIVE_JOBS_HARD_LIMIT} // ($self->max_inactive_jobs + 40);
```
kalikiana, line 14: I would find it more intuitive to default the hard limit 50 rather than depending on the soft limit. And it should be documented accordingly

Martchus: But this way one can increase `OPENQA_CACHE_MAX_INACTIVE_JOBS` and then `OPENQA_CACHE_MAX_INACTIVE_JOBS_HARD_LIMIT` will increase accordingly automatically (which is probably wanted). Not sure whether it makes sense to document is somewhere at this point.

kalikiana (approved):

PR #5501, opened 2024-03-06
CI: Add now needed 'ruby-devel' in base container

okurz:
circleCI nightly documentation jobs fail with

```
mkmf.rb can't find header files for ruby at /usr/lib64/ruby/include/ruby.h
```

For reasons yet unknown it seems we now need to specify ruby-devel as
dependency to install explicitly.

I reproduced the error locally with

```
podman run --rm -it registry.opensuse.org/devel/openqa/ci/containers/base:latest
```

and then verified that after instally "ruby-devel" I could install the
other ruby gems successfully and finally call asciidoctor-pdf.

Related progress issue: https://progress.opensuse.org/issues/156769


perlpunk (approved):

foursixnine (dismissed):
See my comment on https://progress.opensuse.org/issues/156769#note-4

baierjan:
Although I agree that adding `ruby-devel` is a correct solution to the linked problem, it will not solve the entire issue; the next issue in the line is Ruby 2.7 requirement. I propose an alternative solution in #5503

PR #5506, opened 2024-03-12
Use exponential increased delay on retry_tx

b10n1k:
Increase the delay in each retry to handle slow connections. delay is                                                                                                                                                                        
increased exponentially with a factor which by default is 1, meaning                                                                                                                                                                         
exponential backoff is not used.

### DIFF
--- a/lib/OpenQA/Command.pm
+++ b/lib/OpenQA/Command.pm
@@ -122,6 +122,7 @@ sub retry_tx ($self, $client, $tx, $retries = undef, $delay = undef) {
     $client->connect_timeout($ENV{MOJO_CONNECT_TIMEOUT} // 30);
     $delay //= $ENV{OPENQA_CLI_RETRY_SLEEP_TIME_S} // 3;
     $retries //= $ENV{OPENQA_CLI_RETRIES} // 0;
+    my $factor = $ENV{OPENQA_CLI_RETRY_FACTOR} // 1;
     my $start = time;
     for (;; --$retries) {
         $tx = $client->start($tx);
@@ -132,6 +133,7 @@ sub retry_tx ($self, $client, $tx, $retries = undef, $delay = undef) {
 "Request failed, hit error $res_code, retrying up to $retries more times after waiting … (delay: $delay; waited ${waited}s)\n"
         );
         sleep $delay;
+        $delay *= $factor;
     }
 }
 


b10n1k:
> I would suggest to how it is done in https://github.com/okurz/retry/blob/main/retry The ticket says incremental back-off, but probably we want exponential? Also it should be optional via another environment variable.

I thought that exponential can be crazy because the number can grow very fast for a request. what happens if someone gives a big number of retries? Also i dont know about the environment variable. should be enough from user perspective to define only the retries according his needs?

okurz:
> > I would suggest to how it is done in https://github.com/okurz/retry/blob/main/retry The ticket says incremental back-off, but probably we want exponential? Also it should be optional via another environment variable.
> 
> I thought that exponential can be crazy because the number can grow very fast for a request. what happens if someone gives a big number of retries?

That's their choice. Why not let them?

> Also i dont know about the environment variable. should be enough from user perspective to define only the retries according his needs?

Same as in https://github.com/okurz/retry/blob/main/retry and same for whenever any default value is selected: Select a sane default but allow to override it. In this case allow to specify the initial sleep interval, the number of retries and the increase factor.

b10n1k:
As discussed, the increase should happen to the delay, exponentially.

perlpunk (commented):
I think the exponential backoff should be optional, I suggest to just use an environment variable like for the other variables, e.g.
```
my $retry_factor = $ENV{OPENQA_CLI_RETRY_FACTOR} // 1;
```
edit: make the default 1 so no exponential backoff by default

lib/OpenQA/Command.pm
```diff
@@ -121,18 +121,20 @@ sub url_for ($self, $path) {
...
         my $res_code = $tx->res->code // 0;
         return $self->handle_result($tx) unless $res_code =~ /^(50[23]|0)$/ && $retries > 0;
         my $waited = time - $start;
         print encode('UTF-8',
-"Request failed, hit error $res_code, retrying up to $retries more times after waiting … (delay: $delay; waited ${waited}s)\n"
+            "Request failed, hit error $res_code, retrying up to $retries more times after waiting … (delay: $delay; waited ${waited}s)\n"
         );
         sleep $delay;
-    }
+        $delay = $delay << 1 if $delay > $max_delay;
```
perlpunk, line 136: I think we don't need a maximum.
If people deliberately specify 10 retries plus a factor it's their choice, and I don't see a big problem with it.

perlpunk: Also here you only increase delay if it is bigger then the maximum. Which will never happen unless it is bigger than the maximum from the beginning.
Also we said a normal multiplication would be preferrable.
```
$delay *= $factor;
```

Martchus: And even if we wanted a maximum the condition would probably we wrong. So better just remove it.

b10n1k: I have to apologize in advanced. Apparently i dont get it completely. I assume that retry_tx is running in each `openqa-cli` invocation. I try to be responsible of its requests and unlimited ones without think the coincidences sound wrong. Lets say is not a big deal, but to tackle requests which keep failing after retries is not the best design. 
if 5 retries with $delay=3 never reach the max_delay. you need 8 retries. 600sec==10min. Does it really make sense to keep trying or do we need something additional here?

  > my $retry_factor = $ENV{OPENQA_CLI_RETRY_FACTOR} // 1;
  
  Also i think this complicates things. Why to disable it at all? (i think this is open for discussion which i cant win :) But glad to have a productive debate)

b10n1k: > And even if we wanted a maximum the condition would probably we wrong. So better just remove it.

if i dont refresh the page i miss comments. However i have updated the statement

lib/OpenQA/Command.pm
```diff
@@ -132,7 +133,8 @@ sub retry_tx ($self, $client, $tx, $retries = undef, $delay = undef) {
 "Request failed, hit error $res_code, retrying up to $retries more times after waiting … (delay: $delay; waited ${waited}s)\n"
         );
         sleep $delay;
-    }
+        $delay = $delay << 1 unless $delay > $max_delay;
```
okurz, line 136: This doesn't seem to include the suggestions yet to use a simple multiplication, ditch the check for maximum, use a configuration value, only exponential on request

okurz (changes requested):

lib/OpenQA/Command.pm
b10n1k: it was not my intention but i guess i cant fight it.

okurz (approved):

lib/OpenQA/Command.pm
```diff
@@ -132,7 +133,8 @@ sub retry_tx ($self, $client, $tx, $retries = undef, $delay = undef) {
 "Request failed, hit error $res_code, retrying up to $retries more times after waiting … (delay: $delay; waited ${waited}s)\n"
         );
         sleep $delay;
-    }
+        $delay *= $factor;
+    } while ($retries--);
```
kalikiana, line 137: Can we cover this in a unit test e.g. in `t/43-cli-api.t`?

And in case that's not clear. Your answer can be that it is not feasible or there is manual test steps that verify this better. Or maybe other reviewers don't find it necessary.

b10n1k: a small test is added

t/43-cli-api.t
```diff
@@ -456,6 +456,11 @@ EOF
     ($stdout, $stderr, @result) = capture sub { $api->run(@params) };
     like $stderr, qr/Connection refused/, 'aborts on connection refused';
     like $stdout, qr/failed.*retrying/, 'requests are retried on error if requested';
+    $ENV{OPENQA_CLI_RETRY_FACTOR} = 2;
+    $ENV{OPENQA_CLI_RETRY_SLEEP_TIME_S} = 3;
+    @params = ('--host', 'http://localhost:123456', '--retries', 2, 'api', 'test');
+    ($stdout, $stderr, @result) = capture sub { $api->run(@params) };
+    like $stdout, qr/delay: 6; waited 3s/, 'delay is increasing exponentially';
```
okurz, line 463: I tried that locally and that test timed out. I assume you did not mock the sleep time correctly. Tests should never designed to wait for multiple seconds in realtime.

okurz (changes requested):

t/43-cli-api.t
kalikiana: I assume this retries with a delay and eventually fails. Which should be fine. Except it will increase the total runtime.

I thought we had some mocking for sleep calls, but I can't find other uses of that 🤔

kalikiana: Ah. I guess you just add `use Test::Mock::Time;`

b10n1k: I havent managed to make it shorter. I can only reduce the values but i dont see significant improvement

perlpunk: I tried to use Test::Mock::Time (at the top of the test file), but then `retry_tx` just hangs in the second iteration when doing `$tx = $client->start($tx)`.

okurz: @b10n1k I meant "significant" as in the statistical sense, not colloquial. An individual unit test that developers would like to execute many times during development, especially with proper TDD, must not take longer then a couple of seconds. Hence we must be very careful to not unintentionally extend the runtime of tests by some seconds. Otherwise the test becomes ineffective as a developer help and developers find other, inefficient means to develop while the test then still stays and is often only executed in CI runs, likely failing, needing multiple adaptions, taking much more time until a feature can be considered complete. That's also another reason why I consider TDD really important. I am sure if you would have added this test in the beginning you would have realized that it's taking sufficiently long to be annoying already. Only then we can take a look for improvements.

So I suggest to mock whatever is necessary to provide efficient test runs. If that is not possible then we can consider the alternative to not add a test at all as the code is at least covered by already existing tests.

perlpunk: Maybe we can go without the test for now.
But it would be good to find out why Test::Mock::Time isn't working as expected.

kalikiana (approved):

t/43-cli-api.t
b10n1k: Test was removed to move things forward. I might take another look in the first chance

lib/OpenQA/Command.pm
```diff
@@ -122,8 +122,9 @@ sub retry_tx ($self, $client, $tx, $retries = undef, $delay = undef) {
     $client->connect_timeout($ENV{MOJO_CONNECT_TIMEOUT} // 30);
     $delay //= $ENV{OPENQA_CLI_RETRY_SLEEP_TIME_S} // 3;
     $retries //= $ENV{OPENQA_CLI_RETRIES} // 0;
+    my $factor = $ENV{OPENQA_CLI_RETRY_FACTOR} // 1;
     my $start = time;
-    for (;; --$retries) {
+    do {
```
Martchus, line 127: I'd prefer to keep the for-loop. It just made more sense. We do not need to do any checking in the loop-statement itself because our exit-condition is the ` && $retries > 0`-part in the code below which probably should stay (and also stays after your commit). Making this a while-loop with exit-condition-check means we now have two exit-conditions which makes the code harder to understand overall.

Martchus (changes requested):

lib/OpenQA/Command.pm
Martchus: You could make it a `for (;;) { … }`-loop and change ` && $retries > 0` to ` && $retries-- > 0` if you want to keep the decrement and condition in one place. This would possibly be the cleanest solution and wouldn't introduce another exit-condition. However, changing how this loop works is probably out of scope for your commit anyway so I suggest to simply take that change back and really only add the exponential backoff in your commit.

b10n1k: you are right in the sense that two exit conditions are present. But i see more logic to just remove the `&& $retries > `. do..while ensures blocks runs at least one and keep iterate for all the retries. WDYT?

kalikiana: I would prefer to leave the original loop logic untouched at this point.

b10n1k: seems there is a conflict of opinions and I am not convinced. I encouraged to go on with this after an offline comment while we examined some problem. And from my POV `for` implies a certain number of iteration, while the `do--while` says run at least one and then try again under true condition. it is more clear. What;s the actual reason to keep the `for` other than personal preference?

okurz: @b10n1k I would prefer if you can please speed this up just getting this over with. This trivial change is now taking quite long and becoming costly. There should not be unrelated changes in pull requests just "for the sake of it". If you like to introduce a do-while then do that in a separate pull request with an according explicit explanation

Martchus: > What;s the actual reason to keep the for other than personal preference?

I have stated that before:

> Making this a while-loop with exit-condition-check means we now have two exit-conditions which makes the code harder to understand overall.

So I just want to keep the number of conditions for this loop's control flow at one and your change increases them to two.

b10n1k: yea but that means that i have missed to remove the `&& $retries > 0`. it doesnt explain why to switch back to the `for`. I hope my concern is clear.

kalikiana: I understand that

    for (;; --$retries) {
    ...
    }

does what we want and is easier to read.

Or you think there is a bug in the existing code?

b10n1k: https://github.com/os-autoinst/openQA/pull/5506#discussion_r1530148325

removing the `&& $retries > 0` you stay with a clear block which says run once and then repeat while condition is satisfied. I dont think is any bug. it is how you read it. Also i dont remember the argument from the one who suggest that we shouldnt use `for` here, so i cant comment on that

lib/OpenQA/Command.pm
```diff
@@ -122,17 +122,19 @@ sub retry_tx ($self, $client, $tx, $retries = undef, $delay = undef) {
...
         $tx = $client->start($tx);
         my $res_code = $tx->res->code // 0;
-        return $self->handle_result($tx) unless $res_code =~ /^(50[23]|0)$/ && $retries > 0;
+        return $self->handle_result($tx) unless $res_code =~ /^(50[23]|0)$/;
```
perlpunk, line 130: If you remove the `$retries > 0` then an additional line
```
Request failed, hit error 0, retrying up to 0 more times after waiting …
```
will be printed

perlpunk (approved):

lib/OpenQA/Command.pm
b10n1k: Lets forget it. I pushed a changed and as @perlpunk mentioned is not good. I reverted and put back the `for` loop

okurz (approved):

Martchus:
Yes :-)

PR #5507, opened 2024-03-12
Fix showing current result/state filters on overview page

Martchus:
There's no ticket, just something I noticed when working on https://progress.opensuse.org/issues/154261.

### DIFF
--- a/t/ui/10-tests_overview.t
+++ b/t/ui/10-tests_overview.t
@@ -2,6 +2,7 @@
 # SPDX-License-Identifier: GPL-2.0-or-later
 
 use Test::Most;
+use Mojo::Base -signatures;
 
 use FindBin;
 use lib "$FindBin::Bin/../lib", "$FindBin::Bin/../../external/os-autoinst-common/lib";
@@ -497,8 +498,8 @@ subtest 'filtering by job group' => sub {
             name => 'SLE 15 SP5 development'
         });
 
-    my $get_text = sub {
-        $driver->get(shift);
+    my $get_text = sub ($url) {
+        $driver->get($url);
         my @el = $driver->find_element('.card-header');
         return $el[0]->get_text;
     };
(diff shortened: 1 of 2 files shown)


t/ui/10-tests_overview.t
```diff
@@ -497,8 +498,8 @@ subtest 'filtering by job group' => sub {
             name => 'SLE 15 SP5 development'
         });
 
-    my $get_text = sub {
-        $driver->get(shift);
+    my $get_text = sub ($url) {
```
kalikiana, line 501: Isn't this just a style change? I don't see anything here that changes behavior

Martchus: Yes, this is just a style change. I first thought I could reuse this function so I changed it to our current style as I had to move it around. As it turned out that I couldn't reuse it I still kept the style change (as separate commit).

kalikiana: I suggest renaming the PR then

kalikiana: Hrm. Looking at the commits `return filter.replace(/_/g, ' ');` is a functional fix, though?

kalikiana: Okay, so I guess the part I was missing is this the `return` is required if you also use curly parens

kalikiana (approved):

t/ui/10-tests_overview.t
Martchus: Yes. The functional and the style changes are in separate commits so this should be clear.

okurz (approved):

t/ui/10-tests_overview.t
kalikiana: Well, it wasn't 😅 I mentioned it because it stopped me approving, and sometimes you ask about unreviewed PR's

PR #5509, opened 2024-03-12
Allow batch-commenting on test results overview

Martchus:
* Based on #5507
* See https://progress.opensuse.org/issues/154261

screenshot_20240313_163345

---

~~Still a draft because I need to add UI tests and maybe replace the alert boxes with something better.~~


lib/OpenQA/WebAPI.pm
```diff
@@ -501,6 +501,7 @@ sub startup ($self) {
       ->name('apiv1_put_parent_group_comment')->to('comment#update');
     $api_ra->delete('/parent_groups/<parent_group_id:num>/comments/<comment_id:num>')
       ->name('apiv1_delete_parent_group_comment')->to('comment#delete');
+    $api_ru->post('/comments')->name('apiv1_post_comments')->to('comment#create_many');
```
okurz, line 504: I would be careful for two reasons:
1. We should consider complete CRUD
2. "comment" and "comments" are quite similar

Should we consider a different name?

Martchus: > We should consider complete CRUD

We have complete CRUD for comments. Of course I could add routes to also get, update and delete many comments at once by just specifying multiple comment IDs. However, that wouldn't be required for the ticket. I could nevertheless add those routes if you like. They'd only be usable via the CLI of course and one had to use some jq/bash magic to make e.g. delete previously created comments again in one go (based on the audit event¹).

Note that another missing part of this route is the possibility to specify group and parent group IDs. So far one can only create multiple comments for jobs. I also didn't implement group comments because it is not required for the ticket.

Note that with APIs we can always add new stuff but never remove old stuff. This speaks for only adding what we know we need.

> Should we consider a different name?

Maybe `apiv1_post_many_comments`? But I don't think the name we chose here really matters and I'd rather keep it short.

---

¹ I suppose I could also add a button in the audit log table that would actually make use of the delete route to undo a `openqa_comments_create` event.

assets/stylesheets/comments.scss
```diff
@@ -11,25 +25,6 @@
         color: #000;
     }
 
-    .comment-buttons {
-        margin-top: -10px;
-        margin-bottom: 15px;
-    }
-
```
Martchus, line 18: Note that this class doesn't seem to be use anywhere and the comment buttons still look good without it. So the removal is intentional.

lib/OpenQA/WebAPI.pm
okurz: The latter makes sense I guess

templates/webapi/test/overview.html.ep
```diff
@@ -206,3 +211,33 @@
...
+                    <h4 class="modal-title">Add comment on all currently shown jobs</h4>
+                    <button type="button" class="close" data-dismiss="modal">&times;</button>
+                </div>
+                <div class="modal-body">
+                    %= include 'comments/add_comment_form_groups', group_comment => 0, nosubmit => 1
+                </div>
+                <div class="modal-footer">
+                    <div id="batch-commenting-progress-indication" style="width: 100%; display: none">
+                        <span class="mr-auto"><i class="fa fa-cog fa-spin fa-fw"></i> Saving…</span>
+                        <button type="button" class="btn btn" data-dismiss="modal">Close</button>
```
b10n1k, line 230: Why btn is given twice here?

templates/webapi/test/overview.html.ep
```diff
@@ -206,3 +211,33 @@
...
+                    </div>
+                    <div id="batch-commenting-controls">
+                        <button type="submit" class="btn btn-danger">
+                            <i class="fa fa-comment"></i> Submit comment on all <%= scalar @$job_ids %> jobs
+                        </button>
+                        <button type="button" class="btn btn" data-dismiss="modal">Discard</button>
```
b10n1k, line 236: and here

t/api/09-comments.t
```diff
@@ -156,6 +156,18 @@ subtest 'parent group comments' => sub {
...
 };
 
+subtest 'create many comments' => sub {
+    $t->post_ok('/api/v1/comments?job_id=42&job_id=foo')->status_is(400);
+    $t->json_is('/error' => 'Erroneous parameters (job_id invalid, text missing)');
+    $t->post_ok('/api/v1/comments?job_id=42&job_id=80000&text=batch-comment')->status_is(400);
+    $t->json_is('/error' => 'Not all comments could be created.', 'error returned');
+    $t->json_is('/failed' => [{job_id => 42}], 'failed IDs returned');
+    $t->json_is('/created' => [{job_id => 80000, id => 5}], 'created comment returned');
+    ok my $comment = $t->app->schema->resultset('Comments')->find(5), 'comment created' or return;
```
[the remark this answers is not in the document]
okurz: that means if the "ok"-test did not pass then the code would return from here as continuing would lead to misleading errors when trying to work with non-existant comments.

Martchus:
@foursixnine Would this feature be helpful to you? I'm asking because you left a [note](https://progress.opensuse.org/issues/154261#note-6) in the corresponding ticket.

Martchus:
I added an "Undo" button in the audit event table. For this I also added an API route for batch deletion (in the simplest way possible).

Note that even if this PR is not 100 % what's needed (e.g. a way to select individual jobs is required after all) then I believe that this PR nevertheless provides the groundwork for future improvements.

t/ui/23-audit-log.t
```diff
@@ -64,7 +68,7 @@ subtest 'audit log entries' => sub {
     my $search = $driver->find_element('#audit_log_table_filter input.form-control');
     ok($search, 'search box found');
 
-    check_data_table_entries $table, 4, 'four rows without filter';
+    check_data_table_entries $table, 5, 'four rows without filter';
```
okurz, line 71:
```suggestion
    check_data_table_entries $table, 5, 'five rows without filter';
```

?

Although the test description isn't really good. Why should we care that we have exactly four?

t/ui/23-audit-log.t
```diff
@@ -80,7 +84,7 @@ subtest 'audit log entries' => sub {
     $search->clear;
 
     $search->send_keys('newer:today');
-    check_data_table_entries $table, 4, 'again 4 rows when filtering for only newer than today';
+    check_data_table_entries $table, 5, 'again 4 rows when filtering for only newer than today';
```
okurz, line 87: 5 vs. 4?

t/ui/23-audit-log.t
```diff
@@ -111,7 +115,17 @@ subtest 'clickable events' => sub {
     $driver->refresh;
     wait_for_ajax msg => 'DataTable ready';
     my $table = $driver->find_element_by_id('audit_log_table');
-    check_data_table_entries $table, 7, 'seven rows without filter and before posting job/comment';
+    check_data_table_entries $table, 8, 'seven rows without filter and before posting job/comment';
```
okurz, line 118: similar here

okurz (changes requested):

lib/OpenQA/WebAPI/Controller/API/V1/Comment.pm
```diff
@@ -147,6 +147,52 @@ sub create ($self) {
...
+new comment ids or 400 if no text is specified for the comment.
+
+=back
+
+=cut
+
+sub create_many ($self) {
+    my $validation = $self->validation;
+    $validation->required('text')->like(qr/^(?!\s*$).+/);
+    $validation->required('job_id')->num(0, undef);
```
foursixnine, line 162: Please add the parameters to the documentation. In some places through the API, we use jobid, in others job_id, testid, etc.

Same for the delete_many. (A ticket would also suffice)

lib/OpenQA/WebAPI.pm
```diff
@@ -501,6 +501,8 @@ sub startup ($self) {
...
```
foursixnine, line 504: `/jobs/comments` to keep it consistent with what it does.

foursixnine:
> @foursixnine Would this feature be helpful to you? I'm asking because you left a [note](https://progress.opensuse.org/issues/154261#note-6) in the corresponding ticket.

Absolutely :), I do have questions on the performance of multiple jobs on a heavy day, however, exited that this particular one is moving forward!

lib/OpenQA/WebAPI.pm
Martchus: I wouldn't do that because in the future we might want to allow specifying group IDs as well. Then it makes most sense to still have only one route for posting comments in general.

The path `/jobs/comments` is also in conflict with `/jobs/<job_id>/…`. And yes, this is doable. However, I find it not ideal if "comments" and a job ID can occur at the same path element.

Martchus:
> Absolutely :), I do have questions on the performance of multiple jobs on a heavy day, however, exited that this particular one is moving forward!

I doubt there will be performance problems. Of course this could take a few seconds if many IDs have been specified and the server is also otherwise busy. If it really takes too long in extreme cases and thus runs into timeouts we can still make the UI split the requests (like we do on the needle admin page). For now would keep it simple, though.

lib/OpenQA/WebAPI.pm
foursixnine: fair enough

okurz (approved):

kalikiana (approved):

PR #5510, opened 2024-03-13
Fix race conditions when handling barriers

Martchus:
* Avoid race condition when creating a barrier by using `ON CONFLICT DO
  NOTHING` like we already do for job networks
* Avoid race conditions when waiting on and destroying barriers by calling
  update/delete on the search results (and not a single search result) so
  the update/deletion is just not affecting any rows instead of running
  into `row not found`
* See https://progress.opensuse.org/issues/156754

---

The handling of normal locks should probably be improved as well but is not covered here.


lib/OpenQA/Resource/Locks.pm
```diff
@@ -2,6 +2,7 @@
 # SPDX-License-Identifier: GPL-2.0-or-later
 
 package OpenQA::Resource::Locks;
+use Mojo::Base -signatures;
```
perlpunk, line 5:
```suggestion
use Mojo::Base -strict, -signatures;
```
and then remove the strict and warnings lines below

okurz (approved):

perlpunk (approved):

PR #5513, opened 2024-03-15
Enable dependabot updates

perlpunk:
Issue: https://progress.opensuse.org/issues/155410


okurz (approved):
Cool. Would you like to include the rule from https://github.com/openSUSE/qem-dashboard/blob/main/.mergify.yml#L27 as well in this PR or later?

perlpunk:
I would like to make a seperate PR and first see if this works

Martchus:
We shouldn't add a rule to merge these updates automatically because for CSS changes we don't have good tests/checks. For JavaScript it is better but also problematic.

PR #5525, opened 2024-03-15
Add dependabot rules to mergify config

perlpunk:
Issue: https://progress.opensuse.org/issues/155410

### DIFF
--- a/.mergify.yml
+++ b/.mergify.yml
@@ -43,3 +43,24 @@ pull_request_rules:
     actions:
       comment:
         message: This pull request is now in conflicts. Could you fix it? 🙏
+  - name: automatic merge for dependabot updates
+    conditions:
+      - and: *base_checks
+      - and:
+          # mergify config checks needs at least two rules in "and" so we repeat
+          # one from the base checks
+          - base=main
+          - author=dependabot[bot]
+          - label=waited
+    actions:
+      merge:
+        method: squash
+  - name: Wait for 2 days before validating merge
+    actions:
+      label:
+        add: [ waited ]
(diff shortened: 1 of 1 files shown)


Martchus (changes requested):
I wouldn't trust our tests on this because this mainly affects CSS (where we usually don't have good tests/checks for) and JavaScript (where our test coverage is unknown and thanks to use of jQuery APIs many errors are silently ignored).

perlpunk:
Then AC2 https://progress.opensuse.org/issues/155410 cannot be fulfilled.

okurz (approved):
@Martchus I don't see an alternative to that. We can't possibly expect people to test manually every dependabot PR. We can add more tests and openQA-in-openQA tests where we see a lack.

Martchus:
The alternative is what we did before: Not updated this part of the stack all the time just because we can. I think the bot won't create new PRs if there are already pending ones so this shouldn't be a problem.

okurz:
> The alternative is what we did before: Not updated this part of the stack all the time just because we can.

True, not "all the time" but "eventually" unless we completely replace components by something else

> I think the bot won't create new PRs if there are already pending ones so this shouldn't be a problem.

Yes, I expect the same. We will have to collect experiences with that in our context

perlpunk:
So what do we do here?

okurz:
I am sticking with my approval. @Martchus what actual changes would you like to see before we can merge this?

Martchus:
Although I'd still prefer the more conservative approach of testing things out manually as needed and then approve the PR manually.

PR #5531, opened 2024-03-18
Handle special releases in openqa-boostrap (like "Leap 15.6 Beta")

Martchus:
See https://progress.opensuse.org/issues/156907

### DIFF
--- a/script/openqa-bootstrap
+++ b/script/openqa-bootstrap
@@ -20,8 +20,10 @@ fi
 if [ "$NAME" = "openSUSE Leap" ]; then
     # avoid using `obs://…` URL to workaround https://bugzilla.opensuse.org/show_bug.cgi?id=1187425
     repobase=https://download.opensuse.org/repositories/devel:/openQA
-    zypper -n addrepo -p 95 "$repobase/${VERSION}" 'devel:openQA'
-    zypper -n addrepo -p 90 "$repobase:/Leap:/${VERSION}/${VERSION}" "devel:openQA:Leap:${VERSION}"
+    # remove suffixes like " Beta" so e.g. "15.6 Beta" becomes just "15.6"
+    version=${VERSION%% *}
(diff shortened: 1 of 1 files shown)


perlpunk (approved):

okurz (commented):
+1

Martchus:
I created https://openqa.opensuse.org/tests/4021968 to test this. It is the scenario we try to fix but I changed the test code to use the bootstrap script from this PR (see https://github.com/Martchus/os-autoinst-distri-opensuse/blob/bootstrap/tests/openqa/install/openqa_bootstrap.pm). That test has just passed to I guess this PR can be merged.

I can try to create a SR for the openQA-bootstrap package in Leap 15.6. If it isn't accepted in time then this shouldn't be a problem because as soon as Leap 15.6 is actually released the problem would go away on its own. So maybe I should just adjust the test code to use `wget` conditionally?

okurz:
> I created https://openqa.opensuse.org/tests/4021968 to test this. It is the scenario we try to fix but I changed the test code to use the bootstrap script from this PR (see https://github.com/Martchus/os-autoinst-distri-opensuse/blob/bootstrap/tests/openqa/install/openqa_bootstrap.pm). That test has just passed to I guess this PR can be merged.
> 
> I can try to create a SR for the openQA-bootstrap package in Leap 15.6. If it isn't accepted in time then this shouldn't be a problem because as soon as Leap 15.6 is actually released the problem would go away on its own. So maybe I should just adjust the test code to use `wget` conditionally?

I'd try the SR first. If it becomes unlikely that such SR would be quickly accepted then go for workarounds in the test code. I wouldn't mind if you *also* enhance the test code though.

okurz (approved):

Martchus:
Ok, I guess I'll have to wait until this version is in Factory and then create a SR from Factory to https://build.opensuse.org/package/show/openSUSE:Leap:15.6/openQA.

PR #5533, opened 2024-03-19
Update d3 dependencies

perlpunk:
Issue: https://progress.opensuse.org/issues/157462


okurz (approved):

perlpunk:
I think this might be the wrong way. we don't want those as direct dependencies. Maybe it'd better after all to wait for dependabotupdates.

okurz:
How does this relate to https://github.com/os-autoinst/openQA/pull/5548, still necessary?

PR #5536, opened 2024-03-19
Allow pinning parallel clusters on one host

Martchus:
* See contained documentation change
* See https://progress.opensuse.org/issues/135035

---

~~A draft because I still need to extend unit tests. Possibly the big function should also be split. It got a little bit involved but I think the algorithm works without being a full-blown SAT solver. You might want to read the test code first (before the implementation) to see what the difficulty here actually is.~~

### DIFF
--- a/docs/WritingTests.asciidoc
+++ b/docs/WritingTests.asciidoc
@@ -637,6 +637,13 @@ only assign these jobs if there is a sufficient number of free worker-slots. To
 avoid a parallel cluster from starvation its priority is increased gradually and
 eventually workers can be held back for the cluster.
 
+====== Dependency pinning
+It is possible to ensure that all jobs within the same _parallel_ cluster are
+executed on the same worker host. This is useful for connecting the SUTs without
+having to connect the physical worker hosts. Use `PARALLEL_ONE_HOST_ONLY=1` to
+enable this. Of course this means that you need to provide enough worker slots
+on single worker hosts to fit an entire cluster.
+
 ==== Inter-machine dependencies
 Those dependencies make it possible to create job dependencies between tests
 which are supposed to run on different machines.
--- a/docs/WritingTests.asciidoc
+++ b/docs/WritingTests.asciidoc
@@ -641,8 +641,14 @@ eventually workers can be held back for the cluster.
 It is possible to ensure that all jobs within the same _parallel_ cluster are
 executed on the same worker host. This is useful for connecting the SUTs without
 having to connect the physical worker hosts. Use `PARALLEL_ONE_HOST_ONLY=1` to
-enable this. Of course this means that you need to provide enough worker slots
-on single worker hosts to fit an entire cluster.
+enable this. Note that adding this setting in `workers.ini` has currently *no*
+effect.
+
+WARNING: You need to provide enough worker slots on single worker hosts to fit
+an entire cluster. So this feature is mainly intended to workaround situations
+where establishing a physical connection between worker hosts is problematic and
+should not be used needlessly. This feature is also still subject to change as
+we explore ways to make it more flexible.
 
 ==== Inter-machine dependencies
 Those dependencies make it possible to create job dependencies between tests
--- a/lib/OpenQA/Scheduler/Model/Jobs.pm
+++ b/lib/OpenQA/Scheduler/Model/Jobs.pm
@@ -15,7 +15,7 @@ use OpenQA::Constants qw(WEBSOCKET_API_VERSION);
 use OpenQA::Schema;
 use OpenQA::Scheduler::Model::WorkerSlotPicker;
 use Time::HiRes 'time';
-use List::Util qw(all shuffle min sum);
+use List::Util qw(all any shuffle min sum);
 
 # How many jobs to allocate in one tick. Defaults to 80 ( set it to 0 for as much as possible)
 use constant MAX_JOB_ALLOCATION => $ENV{OPENQA_SCHEDULER_MAX_JOB_ALLOCATION} // 80;
@@ -432,16 +432,14 @@ sub _update_scheduled_jobs ($self) {
             push(@missing_worker_class, $job->id);
             $info->{worker_classes} = [];
         }
-        $info->{cluster_jobs} ||= $cluster_infos{$job->id};
-        $info->{one_host_only} = $job->settings_hash->{PARALLEL_ONE_HOST_ONLY};
 
-        if (!$info->{cluster_jobs}) {
-            $info->{cluster_jobs} = $job->cluster_jobs;
+        my $cluster_jobs = $info->{cluster_jobs} ||= $cluster_infos{$job->id};
+        if (!$cluster_jobs) {
+            $cluster_jobs = $info->{cluster_jobs} = $job->cluster_jobs;
             # it's the same cluster for all, so share
-            for my $j (keys %{$info->{cluster_jobs}}) {
-                $cluster_infos{$j} = $info->{cluster_jobs};
-            }
+            $cluster_infos{$_} = $cluster_jobs for keys %$cluster_jobs;
         }
+        $info->{one_host_only} = any { $_->{one_host_only} } values %$cluster_jobs;
         $scheduled_jobs->{$job->id} = $info;
     }
     # fetch worker classes
--- a/lib/OpenQA/Scheduler/Model/Jobs.pm
+++ b/lib/OpenQA/Scheduler/Model/Jobs.pm
@@ -13,7 +13,7 @@ use OpenQA::Log qw(log_debug log_info log_warning);
 use OpenQA::Utils 'random_string';
 use OpenQA::Constants qw(WEBSOCKET_API_VERSION);
 use OpenQA::Schema;
-use OpenQA::Scheduler::Model::WorkerSlotPicker;
+use OpenQA::Scheduler::WorkerSlotPicker;
 use Time::HiRes 'time';
 use List::Util qw(all any shuffle min sum);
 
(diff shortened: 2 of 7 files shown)


t/04-scheduler.t
```diff
@@ -567,4 +567,98 @@ subtest 'allocating network' => sub {
...
+    my $ws1 = $workers->create({host => 'host1', instance => 1}) or return fail 'unable to create worker1';
+    my $ws2 = $workers->create({host => 'host2', instance => 1}) or return fail 'unable to create worker2';
+    my $ws3 = $workers->create({host => 'host2', instance => 2}) or return fail 'unable to create worker3';
+    my $ws4 = $workers->create({host => 'host2', instance => 3}) or return fail 'unable to create worker4';
+
+    my $slots;
+    my $explain_slots = sub () {
+        diag 'slots: ' . join ', ', map { $_->name } @$slots;
+    };
+    my @to_be_schduled = ({id => 1, matching_workers => [$ws1, $ws2]}, {id => 2, matching_workers => [$ws1, $ws2]});
```
okurz, line 581:
```suggestion
    my @to_be_scheduled = ({id => 1, matching_workers => [$ws1, $ws2]}, {id => 2, matching_workers => [$ws1, $ws2]});
```

okurz (changes requested):
oh geez, I have no idea what your very elaborate algorithm is doing :D

only trivial questions or corrections. The majority of the change I did not really want to dive into … yet

Martchus:
> oh geez, I have no idea what your very elaborate algorithm is doing :D

After thinking about it further I must admit that the algorithm is probably also not able to sort out all sorts of inputs. I guess for that we'd really needed a full SAT solver. However, I hope it is sufficient for our purposes. Actually I suppose we don't even have cases as complicated as the one in the unit test in production. The worst case is that jobs won't be scheduled although they could be scheduled if one mixes up worker classes within a scenario too wildely and then we can still improve the algorithm (with TDD by first adding unit tests according to that production scenaio).

lib/OpenQA/Scheduler/Model/Jobs.pm
```diff
@@ -37,6 +37,82 @@ sub determine_scheduled_jobs ($self) {
...
+            next if $visited_worker_slots_by_host{$host}++;  # skip to pick only one slot per host
+            my $matching_worker_slots = ($matching_worker_slots_by_host{$host} //= []);
+            push @$matching_worker_slots, $matching_worker;
+            $visited_worker_slots_by_id{$id} = $job;
+            next if @$matching_worker_slots < $cluster_size;
+            $picked_matching_worker_slots = $matching_worker_slots;
+            last;
+        }
+
+        # go tough the list of matching workers again to re-visit hosts we could not pick a slot on
```
okurz, line 67: Sounds like maybe a recursive call could help here?

lib/OpenQA/Scheduler/Model/Jobs.pm
```diff
@@ -37,6 +37,82 @@ sub determine_scheduled_jobs ($self) {
...
+        for my $matching_worker (@{$job->{matching_workers}}) {
+            my $host = $matching_worker->host;
+            next if $visited_worker_slots_by_host{$host};  # skip hosts we were able to pick a slot on
+
+            # check the job we are competing with for this slot and see whether we might be able to swap picks by finding
```
okurz, line 72: I assume this is only when the candidate host already has other jobs planned occupying jobs? In such case I would be ok if the pinned cluster simply doesn't get a chance to run until enough free slots are available. Can this help to keep the code simpler?

okurz (changes requested):

lib/OpenQA/Scheduler/Model/Jobs.pm
Martchus: I was also thinking about making this recursive. In fact, to be able to solve this in all cases, it probably needed to be recursive. However, maybe it would be better to keep it simple (and thus also not increase the complexity of the algorithm making it scale even more badly¹). What I'm going to do for now is to extract redundant code to separate functions without introducing further recursions.

¹ I think right now it'll scale still well enough while also still be good enough. It doesn't try too badly to solve weird constellations of worker classes but is also not completely stupid about it.

lib/OpenQA/Scheduler/Model/Jobs.pm
Martchus: > In such case I would be ok if the pinned cluster simply doesn't get a chance to run until enough free slots are available.

But it might never run at all then. I think it would be nice if one has a setup with a limited number of slots and parallelism at the same time when assignments would work at all (at least in most cases).

Martchus:
I added/extended unit tests where needed and refactored the code into smaller functions and moved it into its own file.

I also gave it a real test on my local machine with two worker slots and faking the worker host (by manually editing the worker code before starting the particular worker instances). I tested with and without `PARALLEL_ONE_HOST_ONLY=1` and with and without mixed hosts. It seemed to work as expected in all cases.

docs/WritingTests.asciidoc
```diff
@@ -637,6 +637,13 @@ only assign these jobs if there is a sufficient number of free worker-slots. To
 avoid a parallel cluster from starvation its priority is increased gradually and
 eventually workers can be held back for the cluster.
 
+====== Dependency pinning
+It is possible to ensure that all jobs within the same _parallel_ cluster are
+executed on the same worker host. This is useful for connecting the SUTs without
+having to connect the physical worker hosts. Use `PARALLEL_ONE_HOST_ONLY=1` to
+enable this. Of course this means that you need to provide enough worker slots
+on single worker hosts to fit an entire cluster.
```
okurz, line 645: If I am asking vim to consistently wrap that I get

```suggestion
It is possible to ensure that all jobs within the same _parallel_ cluster are
executed on the same worker host. This is useful for connecting the SUTs
without having to connect the physical worker hosts. Use
`PARALLEL_ONE_HOST_ONLY=1` to enable this. Of course this means that you need
to provide enough worker slots on single worker hosts to fit an entire cluster.
```

okurz:
```
set textwidth=78
set wrapmargin=78
```

anyway, I didn't want to block the approval on my comment. I will approve.

okurz (approved):

lib/OpenQA/Scheduler/Model/Jobs.pm
```diff
@@ -79,6 +80,7 @@ sub _allocate_jobs ($self, $free_workers) {
         my $tobescheduled = _to_be_scheduled($j, $scheduled_jobs);
         next if defined $allocated_jobs->{$j->{id}};
         next unless $tobescheduled;
+        OpenQA::Scheduler::Model::WorkerSlotPicker->new($tobescheduled)->pick_slots_with_common_worker_host;
```
kraih, line 83: This should probably be a utility function instead of a model class, since it doesn't rerally represent anything stored longer term.

kraih:
It's a very complex feature, but i like how well encapsulated it is. That will help a lot with debugging in the future.

okurz (changes requested):
After thinking about this and potential follow-up work I came up with one open point that I would like to clarify before merging: The openQA scheduler already has access to worker classes and possibly further worker properties from workers. So would it be possible to define a special worker class like "tap-parallel_one_host_only" resembling the new test setting and read that from the worker to have the same effect as the test variable which is to not schedule any across-host clusters on that host?

Martchus:
We discussed more features that involve worker classes and dependencies and how they are intertwined. We need to make sure that this PR is in-line with those ideas for future features:

1. Allow specifying negative worker class matching. So a job with e.g. `WORKER_CLASS=foo,-bar` would work on slots that provide "foo" unless they also provide "bar".
2. Allow grouping worker classes using tags, e.g. `WORKER_CLASS=host:worker1,region:prg1` would specify the worker classes `host:worker1` and `region:prg1` as usual. However, openQA would also recognize `host` and `region` as tags that come into play when implementing further features.
    1. Allow pinning parallel clusters based on tags. This would be the same as the feature proposed by this PR. However, not the host but a list of tags would be specified. So we would add a variable `PARALLEL_ONE_TAG_ONLY` that could be assigned like `PARALLEL_ONE_TAG_ONLY=region` and then a cluster of parallel jobs would only be assigned to a set worker slots with a common region (so those parallel jobs would *not* be assigned across regions).
        1. Allow specifying the same without extra setting directly as part of the `WORKER_CLASS` setting by introducing a special syntax like `WORKER_CLASS=qemu_x86_64,region:prg1,#region:parallel-one-host`.
    2. Allow pinning other sets of jobs (like a chain of jobs) based on tags. This is probably not needed in the foreseeable future. Having to introduce another variable for this would be ok.
3. Allow the settings introduced by this PR or future feature ideas to be specified in `workers.ini`. This would mean communicating any additional settings as "property" like we already do for e.g. `WORKER_CLASS` (which would be very easy to do). It would also mean that we need to consider those settings in our scheduling algorithm (which is the hard part).

Note that the last point (3.) is maybe the most interesting one for us because in practice we probably really want to put the setting in `workers.ini`.

Here's the relation of those feature ideas with this PR (how I see it):

1. I don't think this PR is in conflict with introducing such a feature. We wouldn't need to change the code I'm introducing here to implement such a feature.
2. I don't think this PR is in conflict with introducing such a feature but there is of course some overlap. We would extend the code I'm introducing here (and/or we introduce for 3.) to implement it. That's fine. It would however leave us with two settings `PARALLEL_ONE_HOST_ONLY` and `PARALLEL_ONE_TAG_ONLY` (and thus basically two ways to achieve the same). I can change this PR to introduce tagging right away to only have `PARALLEL_ONE_TAG_ONLY`. I could also implement 2.1.1 but than it would probably become a bit too much and the syntax is questionable. (The other point 2.2 is probably not relevant.)
3. This changes things. It is again not in conflict with this PR but I probably need to adjust extend the algorithm further because it actually adds another dimension to the problem when the setting come from worker properties. (When the setting comes from the job it is very easy to handle because then the information is part of the set of jobs we want to schedule which we know upfront. When it comes from the worker properties then it is harder to handle because we only build the list of assigned/allocated workers during the process of scheduling.)

I would tend to merge this PR right now as is (plus maybe a note in the documentation that this feature is still subject to change). Then we have something to work with at all for tickets like https://progress.opensuse.org/issues/157414 and can already gather experience with scheduler changes like this in general. Otherwise the PR might also become even more complicated to review.

okurz (commented):
I also agree with this PR as-is except for fixing CI failures and please just add one line in the documentation that you added. "This feature and the naming of the variable is subject to change."

Martchus:
I extended the documentation.

kraih:
You forgot to add the new file it seems.

okurz:
The CI failure was due to https://progress.opensuse.org/issues/157540, added a comment there, bumped prio and retriggered the CI job here.

okurz (approved):

kraih (approved):

PR #5540, opened 2024-03-21
Avoid warning about use of uninitialized value

Martchus:
The variable `$testsuite_name` might be `undef` if the spec is an empty hash leading to:

```
Use of uninitialized value $testsuite_name in hash element at /usr/share/openqa/script/../lib/OpenQA/Schema/Result/JobGroups.pm line 413.
```

We can just skip empty specs here avoiding the warning.

### DIFF
--- a/lib/OpenQA/Schema/Result/JobGroups.pm
+++ b/lib/OpenQA/Schema/Result/JobGroups.pm
@@ -409,7 +409,7 @@ sub _parse_job_template_products ($yaml_products_for_arch, $yaml_defaults_for_ar
             my $description = '';
             if (ref $spec eq 'HASH') {
                 # We only have one key. Asserted by schema
-                $testsuite_name = (keys %$spec)[0];
+                next unless $testsuite_name = (keys %$spec)[0];
                 my $attr = $spec->{$testsuite_name};
                 if ($attr->{priority}) {
                     $prio = $attr->{priority};


okurz (approved):

perlpunk:
where did you see this error? do you have the according yaml file?

Martchus:
In the log on OSD and no, I don't have the yaml file.

lib/OpenQA/Schema/Result/JobGroups.pm
```diff
@@ -409,7 +409,7 @@ sub _parse_job_template_products ($yaml_products_for_arch, $yaml_defaults_for_ar
             my $description = '';
             if (ref $spec eq 'HASH') {
                 # We only have one key. Asserted by schema
-                $testsuite_name = (keys %$spec)[0];
+                next unless $testsuite_name = (keys %$spec)[0];
```
kalikiana, line 412: What does this case look like? As the comment even says, this should have been validated beforehand.

perlpunk: I had a look in the log:
journal:
```
Mar 21 14:49:58 openqa openqa[29349]: Use of uninitialized value $testsuite_name in hash element at /usr/share/openqa/script/../lib/OpenQA/Schema/Result/JobGroups.pm line 413.
```
accesslog:
```
[21/Mar/2024:14:49:58 +0100] "POST /api/v1/job_templates_scheduling/510
[21/Mar/2024:14:49:58 +0100] "POST /api/v1/job_templates_scheduling/319
[21/Mar/2024:14:49:58 +0100] "POST /api/v1/job_templates_scheduling/456
```
Looking at
https://openqa.suse.de/admin/job_templates/319, it has:
```
scenarios:
# ...
  s390x:
    sle-15-SP6-Migration-from-SLE15-SPx-s390x:
      - {}
 ```

Comment (Martchus) [reply lib/OpenQA/Schema/Result/JobGroups.pm:412]: I suppose an empty hash is considered valid here. It probably might make even sense as a "null value" (and this change would to the right thing then).

Comment (perlpunk) [reply lib/OpenQA/Schema/Result/JobGroups.pm:412]: Currently it would still call ` _parse_job_template_machines()` below with an undefined value.
If this is skipped, then the result might be different.
I just had a look at the job template editor and clicked on "Show expanded version and got this:
```
scenarios:
# ...
  s390x:
    sle-15-SP6-Migration-from-SLE15-SPx-s390x:
    - '':
        machine: s390x-kvm
        priority: 50
        settings: {}
```
I think with this PR this entry wouldn't be there anymore.
But also an empty string for the testsuite name doesn't make much sense.
I also looked into the database and we have a lot of entries like this with an empty name:
```
select id, product_id, test_suite_id, t_updated, group_id, name from job_templates where group_id=319 limit 20;
   id   | product_id | test_suite_id |      t_updated      | group_id |                               name                                
--------+------------+---------------+---------------------+----------+-------------------------------------------------------------------
...
 255410 |       2690 |          5576 | 2023-10-19 06:27:17 |      319 | 
```

perlpunk (changes requested):
See my comments. I'm not completely sure if it might break things

lib/OpenQA/Schema/Result/JobGroups.pm
kalikiana: #5583

Martchus:
Superseded by https://github.com/os-autoinst/openQA/pull/5583.

PR #5543, opened 2024-03-22
Display full footer on pages

b10n1k:
Added align-content to the .footer element to make the element stick to the
end. align-content takes effect with flex position variable. The scroll bar
disappears but the stack still do not show pretty when the size of the
window is too small.
    
https://progress.opensuse.org/issues/157615

### DIFF
--- a/assets/stylesheets/overall.scss
+++ b/assets/stylesheets/overall.scss
@@ -35,6 +35,8 @@ body {
     width: 100%;
     // set the fixed height of the footer here
     height: 30px;
+    display: flex;
+    align-content: end;
     background-color: #f5f5f5;
 }
 .darkmode {


okurz:
How does it look like when there is no footer?

Martchus (changes requested):
Can we completely avoid specifying a fixed height here? The problem with a fixed height is that the footer actually has a variable height depending on its contents and probably also depending on whether line wrapping occurs.

b10n1k:
> with a fixed height is that the footer actually has a variable height depending on its contents and probably also depending on whether line wrapping occurs.

So I experimented with bottom variable and percentages values on height variable. I think removing the fixed value altogether solves the problem. please review

assets/stylesheets/overall.scss
```diff
@@ -33,8 +33,6 @@ body {
     position: absolute;
     bottom: 0;
     width: 100%;
-    // set the fixed height of the footer here
-    height: 30px;
```
kalikiana, line 37: For me this causes the footer to overlap the **Submit comment** button unless the page is wide enough:

image

On the overview the message box at the bottom is also partially obscured:

image

b10n1k: you are right

Martchus: I guess it would be ok because also on those screenshots it looks like one had to shrink the screen a lot (and also increase the zoom factor).

Maybe it would still be beneficial to make the footer a normal element (not overlapping with other elements but just taking its space on the page) in this case (e.g. using media queries or some [positioning magic](https://developer.mozilla.org/en-US/docs/Web/CSS/position) perhaps following this [guide](https://developer.mozilla.org/en-US/docs/Web/CSS/Layout_cookbook/Sticky_footers)).

assets/stylesheets/overall.scss
```diff
@@ -1,6 +1,6 @@
 body {
     font-family: $default-font;
-    line-height: 1.428571429;
+    line-height: normal;
```
okurz, line 3: If that fixes something why was it set to that specific number in before? Did you crosscheck that it still looks fine even when there is no custom_links page adding the second line?

Martchus (changes requested):
Note that this is about https://progress.opensuse.org/issues/157576 which should be referenced.

Having to change the line height globally to fix just the footer seems wrong, though.

~~I tried to reproduce the issue locally but I couldn't with the latest Chromium or Firefox. Even with the smallest possible window size and zooming the footer was never cropped like in the screenshot of the ticket. Of course I tested with the OSD footer and on various pages, including the test results overview page.~~

~~In the screenshot of the ticket it looks like the scrollbar is not at the very bottom. But then it would make sense that the very bottom of the page is not visible. That the footer behaves like a normal page element in this case is expected.~~

~~Considering my findings I would tend to reject the ticket and avoid any global changes or otherwise unnecessary tweaks.~~

Ok, this is reproducible on pages that have not much content so there has to be space between the end of the normal page contents and the footer.

---

(The way the additional links on OSD's footer are placed at the left side is not very nice but that's a different story and probably needed changes in the HTML structure.)

assets/stylesheets/overall.scss
kalikiana: image

This is what I'm seeing

b10n1k: no it doesnt. I will have to revert it

assets/stylesheets/overall.scss
```diff
@@ -35,6 +35,7 @@ body {
     width: 100%;
     // set the fixed height of the footer here
     height: 30px;
+    align-content: end;
```
okurz, line 38: my firefox developer tools say about this
> align-content has no effect on this element since it’s neither a flex container nor a grid container.
> Try adding display:grid or display:flex.

assets/stylesheets/overall.scss
```diff
@@ -57,6 +58,9 @@ body {
         margin-right: 15px;
     }
 }
+#footer-legal {
+    font-size: x-small;
+}
```
okurz, line 63: I seems we don't have a common understanding what the issue is. I don't want the text to be smaller.

I tried to make that clear in https://progress.opensuse.org/issues/157576 with the screenshot and the explaining text. The issue is not that the existing content takes too much space but that on a page with a lot of white space still there is a scrollbar when there shouldn't be one at all. Please see  again

okurz (changes requested):
Please see my alternative attempt in https://github.com/os-autoinst/openQA/pull/5546

assets/stylesheets/overall.scss
b10n1k: yes i know, i forgot to add this variable

assets/stylesheets/overall.scss
b10n1k: I get that part but I didnt like how it arrange the two div. I thought that the size wouldnt be a big problem. Anyway i remove it and left it as is

okurz: Sure, you can introduce such changes but then please make them explicit with an according explaining commit message. Also I suggest in pull requests describe the changes that you do in updates. E.g. "I added the missing `display:flex` and added another commit to make the div arrangement look nice by reducing the text size which I think is better but tell me what you think"

assets/stylesheets/overall.scss
okurz: And does this actually something for you? For me it makes no difference in display

okurz (changes requested):

b10n1k:
image

Martchus:
Maybe that's actually a good solution. Does it fix the problem mentioned by @kalikiana?

okurz:
Maybe it does. I didn't understand why @kalikiana stated in https://github.com/os-autoinst/openQA/pull/5546#discussion_r1539049620 that everything is becoming worse

PR #5546, opened 2024-03-25
style: Fix custom footer overflow

okurz:
On an openQA page using a custom links_footer_left and/or
links_footer_right, e.g.
https://openqa.suse.de/tests/overview?distri=sle&build=20240318-1&result=incomplete
the version string in the footer is cropped despite enough white space
being available. The full version string is readable by scrolling down
but the scroll bar should not be necessary. So no cropping and no scroll
bar should appear if there is enough blank space in both the case of
with or without a custom footer_link content used or not.

This problem was originally introduced with 58747a5ef setting a fixed
height which offsets the version string from the bottom to make it more
readable but does not account for the optional custom_footer inclusion.

This commit fixes this by using a bottom padding instead of prescribing
the overall height of the complete footer box.

Before:
Screenshot_20240325_221909_before_fixed_height

After:
Screenshot_20240325_221758_after_padding-bottom-10px

Related progress issue: https://progress.opensuse.org/issues/157576

### DIFF
--- a/assets/stylesheets/overall.scss
+++ b/assets/stylesheets/overall.scss
@@ -35,6 +35,7 @@ body {
     width: 100%;
     padding-bottom: 10px;
     background-color: #f5f5f5;
+    z-index: -1;
 }
 .darkmode {
     .footer {
--- a/assets/stylesheets/overall.scss
+++ b/assets/stylesheets/overall.scss
@@ -33,8 +33,7 @@ body {
     position: absolute;
     bottom: 0;
     width: 100%;
-    // set the fixed height of the footer here
-    height: 30px;
+    padding-bottom: 10px;
     background-color: #f5f5f5;
 }
 .darkmode {


perlpunk (approved):

Martchus (commented):
This has the issue pointed out by @kalikiana in the other PR but I'd be ok with it. Maybe also add padding to the top for the text being vertically centered? And I think just 5 px would look better (but top and bottom).

okurz:
> This has the issue pointed out by @kalikiana in the other PR but I'd be ok with it. Maybe also add padding to the top for the text being vertically centered? And I think just 5 px would look better (but top and bottom).

The single text line for me would be 10px high so I merely replaced the 30px height with (automatic line height)+10px padding. I guess the original idea was to not squeeze the text on the lower border, not to center it in the grey box. But I guess in the process we could also change it though my personal preference would be to stay as much true to the previous approach. So, should I still change it? What do others think?

assets/stylesheets/overall.scss
```diff
@@ -33,8 +33,7 @@ body {
     position: absolute;
     bottom: 0;
     width: 100%;
-    // set the fixed height of the footer here
-    height: 30px;
+    padding-bottom: 10px;
```
kalikiana, line 36: For me the result is worse than before.

Before:

image


After:

image

okurz: it seems you want to address a quite different problem. Your screenshots look like a tiny resolution.

If this is about the footer hiding the button we could give the footer a `z-index: -1;` so the footer would overlap other content on the screen

perlpunk: Yes, I also thought about using z-index. so the footer would **not** overlap content.

Martchus: > Your screenshots look like a tiny resolution.

Well, let's say the *effective* width is very small. And I'd argue that this use case is not very important because with such a small width the rest of the page is not very usable anymore anyway. So I'd still be fine with this change.

> If this is about the footer hiding the button we could give the footer a `z-index: -1;` so the footer would overlap other content on the screen

You can try that.

okurz: ok. I added another commit with `z-index: -1;` which fixes the issue that @kalikiana pointed out at least in my tests. Acceptable?

Martchus (approved):
I guess I would have preferred `padding-top: 5px; padding-bottom: 5px;` but I guess this is fine as well (especially considering your explanation).

assets/stylesheets/overall.scss
kalikiana: As my previous message, and in particular the last screenshot, was supposed to clarify: **I don't see any issue besides the background color unless I make the page tiny**. The z-index just breaks it differently by hiding the footer.

You can argue that you don't care about browser other than whatever you tested. For me all of these suggestions worsen the situation.

PR #5559, opened 2024-04-02
Simplify+signatures

okurz:
* Simplify Resource::Locks with direct return of booleans
* Use signatures in Resource::Locks
* Use signatures in Resource::Jobs
* Slightly simplify script/openqa-validate-yaml

### DIFF
--- a/lib/OpenQA/Resource/Locks.pm
+++ b/lib/OpenQA/Resource/Locks.pm
@@ -55,13 +55,9 @@ sub lock ($name, $jobid, $where) {
     return 0 unless $lock;
 
     # lock is locked and not by us
-    if ($lock->locked_by) {
-        return 0 if ($lock->locked_by != $jobid);
-        return 1;
-    }
+    return $lock->locked_by == $jobid if $lock->locked_by;
     # we're using optimistic locking, if this succeeded, we were first
-    return 1 if ($locks->update({locked_by => $jobid}) > 0);
-    return 0;
+    return $locks->update({locked_by => $jobid}) > 0;
 }
 
 sub unlock ($name, $jobid, $where) {
@@ -121,15 +117,11 @@ sub barrier_wait ($name = undef, $jobid = undef, $where = undef, $check_dead_job
         }
     }
 
-    if (grep { $_ eq $jobid } @jobs) {
-        return 1 if @jobs == $barrier->count;
-        return 0;
-    }
+    return @jobs == $barrier->count if grep { $_ eq $jobid } @jobs;
 
     push @jobs, $jobid;
     return -1 unless $barriers->update({locked_by => join(',', @jobs)}) > 0;
-    return 1 if @jobs == $barrier->count;
-    return 0;
+    return @jobs == $barrier->count;
 }
 
 sub barrier_destroy ($name = undef, $jobid = undef, $where = undef) {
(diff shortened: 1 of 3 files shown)


lib/OpenQA/Resource/Locks.pm
```diff
@@ -19,8 +19,7 @@ my %final_states = map { $_ => 1 } OpenQA::Jobs::Constants::NOT_OK_RESULTS();
 # So we have to specify, which child job is supposed to create the lock
 # and watch its state.
 #
-sub _get_lock {
-    my ($name, $jobid, $where) = @_;
+sub _get_lock ($name = undef, $jobid = undef, $where = undef) {
```
perlpunk, line 22: I think `_get_lock` is always called with 3 arguments, so
```suggestion
sub _get_lock ($name, $jobid, $where) {
```

Martchus: The same counts probably for `unlock` as well.

lib/OpenQA/Resource/Locks.pm
```diff
@@ -126,18 +116,14 @@ sub barrier_wait ($name, $jobid, $where, $check_dead_job) {
         }
     }
 
-    if (grep { $_ eq $jobid } @jobs) {
-        return 1 if @jobs == $barrier->count;
-        return 0;
-    }
+    return @jobs == $barrier->count if grep { $_ eq $jobid } @jobs;
```
Martchus, line 119: The code invoking this function actually does a numerical comparison with the return value. We shouldn't just return a boolean. The same counts for the other occurrences.

Martchus (dismissed):

lib/OpenQA/Resource/Locks.pm
perlpunk: A boolean created like this will already behave like expected in numeric context without a warning.
```
$b = 23 == 24; # magic boolean that just looks like an empty string
$s = ""; # just an empty string
say "\$b:";
say $b + 1; 
say "\$s:";
say $s + 1;
__END__
$b:
1
$s:
Argument "" isn't numeric in addition (+) at -e line 7.
1
```

lib/OpenQA/Resource/Locks.pm
okurz: ok, removed the `= undef` and also the obsolete safety checks for variables after that then

lib/OpenQA/Resource/Locks.pm
```diff
@@ -97,7 +91,7 @@ sub create {
 ## Barriers
 # barriers are created with number of expected jobs. Then wait call waits until the expected number of jobs is waiting
 
-sub barrier_create ($name, $jobid, $expected_jobs) {
+sub barrier_create ($name = undef, $jobid = undef, $expected_jobs = undef) {
```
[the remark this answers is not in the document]
okurz: hm, why do we have `return 0 unless $name && $jobid && $expected_jobs;` in here then?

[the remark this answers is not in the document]
Martchus: Because they might be `undef`. There's a difference between an argument being not passed at all (not the case here) and `undef` being passed as argument (possibly the case here).

okurz:
please review again and explain what's missing. I didn't see any open points.

kalikiana (approved):

okurz:
CI fails due to problems on rebasing, on it …

Martchus (approved):

perlpunk (approved):

PR #5560, opened 2024-04-02
Switch to docker compose subcommand

perlpunk:
The docker-compose tool isn't maintained anymore and should be replaced by the docker compose subcommand:
https://docs.docker.com/compose/migrate/

    docker-compose: command not found

### DIFF
--- a/tools/test_containers_compose
+++ b/tools/test_containers_compose
@@ -19,19 +19,19 @@ wait_until() {
 
 setup_containers() {
     for retry in {2..0}; do
-        sudo docker-compose build && break
+        sudo docker compose --compatibility build && break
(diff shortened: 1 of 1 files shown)


okurz (approved):

perlpunk:
I'm debugging this

perlpunk:
`export COMPOSE_COMPATIBILITY=1` at the top of `tools/test_containers_compose` did not help

Martchus (approved):

perlpunk:
https://docs.docker.com/compose/migrate/#service-container-names suggests that the changed container names (`-` instead of `_`) are the problem. I currently don't see a place where we use the container names explicitly though.

PR #5567, opened 2024-04-08
worker: Provide a sensible default for "critical load threshold"

okurz:
Based on our experiences various issues with tests can appear if the system load is exceeding certain sane levels, for example typing issues such as mistyping, hanging keys, lost characters, but also various timeouts and even network outages.

This commit introduces a new default value enabling the recently introduced check for "critical load" on an openQA worker to prevent an openQA worker from starting new jobs until the system load is again below a threshold.

The value selection is of course based on empirical data gathered so far and can and should be tweaked if we know more.

Related progress issue: https://progress.opensuse.org/issues/158125


Martchus:
Looks like tests need to be adjusted.

PR #5568, opened 2024-04-08
t: Fix sporadic test failure in ui/13-admin.t

okurz:
There was a sporadic failure originally observed in CI runs with very low
failure ratio looking like this

```
t/ui/13-admin.t .. 14/?
        #   Failed test 'exactly one key present'
        #   at t/ui/13-admin.t line 681.
        #          got: '0'
        #     expected: '1'
```

I reproduced the problem locally with

```
runs=200 OPENQA_TEST_TIMEOUT_DISABLE=1 count-fail-ratio make test KEEP_DB=1
TESTS=t/ui/13-admin.t
```

with results fails: 4. Fail ratio 2.00±1.94%, mean runtime: 39.0±3.0 s.

With `make coverage` the results were fails 7. Fail ratio 3.50±2.54%,
mean runtime: 80 s so slightly higher fail ratio with double runtime.

This commit uses "wait_for_element" combining the submit action with the lookup
of the resulting table rows to properly synchronize the action and waiting for
result. This fixes the problem. Verified with 400 runs with a computed failure
probability of less than .75%.

Previously I tried to just wait for "#api-keys-tbody > tr" in
"wait_for_element" but running into sporadic issues with similar failure rate
as in before. Apparently the key creation never yields a valid table
representation in this case without retriggering the action.

Related progress issue: https://progress.opensuse.org/issues/158422

### DIFF
--- a/t/ui/13-admin.t
+++ b/t/ui/13-admin.t
@@ -677,17 +677,24 @@ subtest 'Manage API keys' => sub {
     };
 
     subtest 'create key with expiration date' => sub {
-        $driver->find_element('#api-keys-form input[type=submit]')->click;
-        is(scalar @{$driver->find_child_elements($tbody = api_keys_tbody, 'tr')}, 1, 'exactly one key present');
-        like($tbody->get_text, qr/\d{4}-\d{2}-\d{2} \d{2}:\d{2}:\d{2}/, 'new key has expiration date');
+        my $row = wait_for_element(
+            trigger_function => sub { $driver->find_element('#api-keys-form input[type=submit]')->click },
+            selector => '#api-keys-tbody > tr',
+            description => 'key row present'
+        ) or return;
+        like($row->get_text, qr/\d{4}-\d{2}-\d{2} \d{2}:\d{2}:\d{2}/, 'new key has expiration date');
     };
 
     subtest 'create key without expiration date' => sub {
+        $tbody = api_keys_tbody;
         unlike($tbody->get_text, qr/never/, 'no key without expiration date present so far');
         $driver->find_element_by_id('expiration')->click;
-        $driver->find_element('#api-keys-form input[type=submit]')->click;
-        is(scalar @{$driver->find_child_elements($tbody = api_keys_tbody, 'tr')}, 2, 'two keys present now');
-        like($tbody->get_text, qr/never/, 'new key has expiration date');
+        my $row = wait_for_element(
+            trigger_function => sub { $driver->find_element('#api-keys-form input[type=submit]')->click },
+            selector => '#api-keys-tbody tr:nth-child(2)',
+            description => 'key row present (without expiration)'
+        ) or return;
+        like($row->get_text, qr/never/, 'new key has expiration date');
     };
 };
 


t/ui/13-admin.t
```diff
@@ -677,17 +677,25 @@ subtest 'Manage API keys' => sub {
...
     };
 
     subtest 'create key without expiration date' => sub {
+        $tbody = api_keys_tbody;
         unlike($tbody->get_text, qr/never/, 'no key without expiration date present so far');
         $driver->find_element_by_id('expiration')->click;
-        $driver->find_element('#api-keys-form input[type=submit]')->click;
-        is(scalar @{$driver->find_child_elements($tbody = api_keys_tbody, 'tr')}, 2, 'two keys present now');
-        like($tbody->get_text, qr/never/, 'new key has expiration date');
+        wait_for_element(
```
Martchus, line 692: Considering there's already one element from the previous subtest this `wait_for_element` will *not* wait for an *additional* element to show up. So using it like this is a bit pointless. Maybe the function `wait_for_element` should have a parameter to wait for the n-th element.

t/ui/13-admin.t
```diff
@@ -677,17 +677,25 @@ subtest 'Manage API keys' => sub {
...
+            trigger_function => sub { $driver->find_element('#api-keys-form input[type=submit]')->click },
+            selector => '#api-keys-tbody > tr', description => 'key row present (without expiration)'
+        );
+        my @rows = $driver->find_elements('#api-keys-tbody > tr');
```
Martchus, line 696: With my previous suggestion this repetition could be avoided (and the return value of `wait_for_element` be used instead).

t/ui/13-admin.t
```diff
@@ -677,17 +677,25 @@ subtest 'Manage API keys' => sub {
...
     subtest 'create key with expiration date' => sub {
-        $driver->find_element('#api-keys-form input[type=submit]')->click;
-        is(scalar @{$driver->find_child_elements($tbody = api_keys_tbody, 'tr')}, 1, 'exactly one key present');
-        like($tbody->get_text, qr/\d{4}-\d{2}-\d{2} \d{2}:\d{2}:\d{2}/, 'new key has expiration date');
+        my @rows = wait_for_element(
+            trigger_function => sub { $driver->find_element('#api-keys-form input[type=submit]')->click },
+            selector => '#api-keys-tbody > tr', description => 'key row present'
+        );
+        is(scalar @rows, 1, 'exactly one key present') or diag explain @rows;
+        like($rows[0]->get_text, qr/\d{4}-\d{2}-\d{2} \d{2}:\d{2}:\d{2}/, 'new key has expiration date');
```
Martchus, line 685: `wait_for_element` will always return one value so this check doesn't make much sense and the variable name `@rows` is misleading. The `diag explain @rows` call will likely not print anything useful as well.

```suggestion
        my $row = wait_for_element(
            trigger_function => sub { $driver->find_element('#api-keys-form input[type=submit]')->click },
            selector => '#api-keys-tbody > tr', description => 'key row present'
        );
        ok defined $row, 'row with key present' or return;
        like $row->get_text, qr/\d{4}-\d{2}-\d{2} \d{2}:\d{2}:\d{2}/, 'new key has expiration date';
```

Martchus (changes requested):

t/ui/13-admin.t
okurz: Ok, I combined the `wait_for_element` directly with `return` as the `wait_for_element` would already fail if it hasn't found an element.

okurz:
Updated
* `wait_for_element … or return` instead of redundant array size checks
* Using `#api-keys-tbody tr:nth-child(2)` to select the actual needed second element

PR #5573, opened 2024-04-11
Create unit files to auto restart services after minion update

baierjan:
Reference: https://progress.opensuse.org/issues/158814


systemd/openqa-minion-restart.service
```diff
@@ -0,0 +1,6 @@
+[Unit]
+Description=Restarts services which are using Minion
+
+[Service]
+Type=oneshot
+ExecStart=/usr/bin/systemctl try-restart openqa-webui.service openqa-gru.service openqa-worker-cacheservice-minion.service
```
Martchus, line 6: We probably also need to restart `openqa-worker-cachservice.service` as it is basically the Minion dashboard on the workers:
```suggestion
ExecStart=/usr/bin/systemctl try-restart openqa-webui.service openqa-gru.service openqa-worker-cacheservice.service openqa-worker-cacheservice-minion.service
```

Martchus (commented):
Looks generally good.

systemd/openqa-minion-restart.service
baierjan: Added.

okurz (approved):

Martchus (approved):

PR #5574, opened 2024-04-11
Fix misc errors and warnings showing up in web UI log

Martchus:
See particular commit messages and https://progress.opensuse.org/issues/158808


Martchus:
Note that this will *not* fix https://progress.opensuse.org/issues/158859. It will just avoid two related errors from showing up in our logs as described in the commit message. The error will of course still be displayed to the user and maybe we will need to tweak the way we invoke grep (as part of https://progress.opensuse.org/issues/158859).

kalikiana (approved):

okurz (approved):
Great opportunity to fix the test coverage :)

PR #5575, opened 2024-04-12
Typos in live.html.ep

tbertels:


Martchus:
Looks good but it would be nice if the commit message was "Fix typos in live.html.ep" (in-line with our normal commit message style).

tbertels:
https://github.com/os-autoinst/openQA/pull/5577

PR #5577, opened 2024-04-12
Fix typos in live.html.ep

tbertels:


Martchus:
Thanks for the fix and updating the commit message. Note that you can simply update a commit without creating a new PR by amending and force-pushing it (so you can save yourself some effort next time).

okurz (approved):

Martchus (approved):

PR #5579, opened 2024-04-13
Fix needle tags display for check_screen and assert_screen

tbertels:
`data.check_screen` returns `false` for `check_screen` and `assert_screen` so the needle tags aren't displayed. `data.check_screen.mustmatch` could be checked instead, but this check happens already in testapi.pm `_check_or_assert`

### DIFF
--- a/assets/javascripts/running.js
+++ b/assets/javascripts/running.js
@@ -1100,7 +1100,7 @@ var messageToStatusVariable = [
     statusVar: 'currentApiFunction',
     action: function (value, data) {
       developerMode.currentApiFunctionArgs = '';
-      if ((value === 'assert_screen' || value === 'check_screen') && data.check_screen) {
+      if (value === 'assert_screen' || value === 'check_screen') {
         developerMode.currentApiFunctionArgs = data.check_screen.mustmatch;
       }
     }


assets/javascripts/running.js
```diff
@@ -1100,7 +1100,7 @@ var messageToStatusVariable = [
     statusVar: 'currentApiFunction',
     action: function (value, data) {
       developerMode.currentApiFunctionArgs = '';
-      if ((value === 'assert_screen' || value === 'check_screen') && data.check_screen) {
+      if (value === 'assert_screen' || value === 'check_screen') {
```
Martchus, line 1103: Did you actually test this? If `data.check_screen` is null/undefined then it is not possible to access `data.check_screen.mustmatch` and we'd get an error. So to me it makes sense to check whether `data.check_screen` is a truthy value.

Martchus (changes requested):

assets/javascripts/running.js
tbertels: After more tests I figured out that this only happens for the first call of both `check_screen` and `assert_screen`: they have an `undefined` `data.check_screen`.
When restarting the job, `data.check_screen` does exist during the first call of both `check_screen` and `assert_screen`. This is what confused me.

PR #5589, opened 2024-04-23
Make CRITICAL_LOAD_AVG_THRESHOLD work with AppArmor

Martchus:
The code otherwise runs into a permission error leading to the following warning:

```
Apr 23 14:46:30 aarch64-o3 worker[39373]: [warn] Unable to determine average load: Can't open file "/proc/loadavg": Permission denied at /usr/lib/perl5/vendor_perl/5.26.1/Mojo/File.pm line 133.
```


kalikiana (approved):
Good catch!

PR #5592, opened 2024-04-25
Use the latest version of AceEditor instead of CodeMirror

Martchus:
The version of CodeMirror we are using is outdated. The latest version of AceEditor seems easier to use than the latest version of CodeMirror (as no bundling is required) and supports everything we need. So it makes sense to switch to AceEditor.

Related ticket: https://progress.opensuse.org/issues/159411

### DIFF
--- a/assets/assetpack.def
+++ b/assets/assetpack.def
@@ -102,15 +102,15 @@
 ! dependency_graph.css
 < stylesheets/dependency_graph.scss
 
-! codemirror.css
-< ../node_modules/codemirror/lib/codemirror.css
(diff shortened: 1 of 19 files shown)


Martchus:
I tested it locally on Chromium and Firefox. The behavior regarding line-wrapping and scrolling is the same as before. Syntax highlighting still works as well. Now the YAML editor also supports code-folding.

screenshot_%Y%M%D_%H%m%S

baierjan (approved):

okurz (approved):

PR #5593, opened 2024-04-26
Revert "Use the latest version of AceEditor instead of CodeMirror"

baierjan:
Reverts os-autoinst/openQA#5592

```
Apr 26 10:11:36 new-ariel openqa-webui-daemon[21685]: AssetPack was unable to locate related asset '../node_modules/ace-builds/css/main-1.png' at /usr/share/openqa/script/../lib/Mojolicious/Plugin/AssetPack/Pipe/FetchForNode.pm line 25.
```

PR #5594, opened 2024-04-26
Fix missing installation of further assets required by Ace

Martchus:
See https://progress.opensuse.org/issues/159411

---

Reproduced/tested locally by installing it using some `DESTDIR` and using the installed `node_modules` directory instead of the one populated by NPM. I'll come up with another change to test this, e.g. as part of our package builds. It would also be possible to avoid having to specify those files manually - although I'm not sure whether it is worth the effort.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -65,7 +65,7 @@ install-generic:
 		cp -a $$i/* "$(DESTDIR)"/usr/share/openqa/$$i ;\
 	done
 	for f in $(shell grep --perl-regexp '\.\.\/node_modules\/.*\.*+' assets/assetpack.def | sed -e 's|<* ../||') \
-		node_modules/fork-awesome/fonts/* node_modules/chosen-js/*.png; do \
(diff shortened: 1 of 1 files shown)


okurz (approved):
please look into the docker-compose issue and merge manually as necessary

Martchus:
It fails due to

```
 webui-db-init-1  | AssetPack was unable to locate related asset '../node_modules/ace-builds/css/main-1.png' at /usr/share/openqa/script/../lib/Mojolicious/Plugin/AssetPack/Pipe/FetchForNode.pm line 25.
```

This is related but I see this error also on #5593 which means this check simply only runs on master. This is also why it couldn't catch the issue in the first place. (The docker-compose check is pretty useless for finding errors except in the docker-compose setup/config itself.)

baierjan:
> The docker-compose check is pretty useless for finding errors except in the docker-compose setup/config itself.

Yes, that's annoying.

PR #5595, opened 2024-04-26
Add check for loading assets via openQA-test package

Martchus:
See https://progress.opensuse.org/issues/159411

### DIFF
--- a/dist/rpm/openQA-test.spec
+++ b/dist/rpm/openQA-test.spec
@@ -23,6 +23,11 @@ touch %{_sourcedir}/%{short_name}
 # call one of the components but not openqa itself which would need a valid
 # configuration
 /usr/share/openqa/script/initdb --help
+
+# verify whether assets can be loaded
+perl -I/usr/share/openqa/lib -mMojolicious -mMojo::Home -mOpenQA::Setup \
+    -e 'OpenQA::Setup::setup_asset_pack(Mojolicious->new(home => Mojo::Home->new("/usr/share/openqa")))'
+
 getent passwd geekotest
 
 %install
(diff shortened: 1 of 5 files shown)


dist/rpm/openQA-test.spec
```diff
@@ -23,6 +23,11 @@ touch %{_sourcedir}/%{short_name}
 # call one of the components but not openqa itself which would need a valid
 # configuration
 /usr/share/openqa/script/initdb --help
+
+# verify whether assets can be loaded
+perl -I/usr/share/openqa/lib -mMojolicious -mMojo::Home -mOpenQA::Setup \
+    -e 'OpenQA::Setup::setup_asset_pack(Mojolicious->new(home => Mojo::Home->new("/usr/share/openqa")))'
```
okurz, line 29:
```suggestion
perl -I/usr/share/openqa/lib -mMojolicious -mMojo::Home -mOpenQA::Setup \
    -e 'OpenQA::Setup::setup_asset_pack(Mojolicious->new(home => Mojo::Home->new("/usr/share/openqa")))' || \
     { echo "assets could not be loaded"; exit $ERRCODE; }
```

Martchus: If this command fails it will already print a useful error message. So I wouldn't add another one.

okurz: sure but I was trying to get rid of the comment with executable code :)

okurz (approved):

Martchus:
I now additionally added another commit to make this generally more reliable and less manual.

I created https://github.com/os-autoinst/openQA/pull/5596 to see whether this check would now actually fail.

baierjan (approved):

PR #5597, opened 2024-04-29
Fix build requirements in spec file when tests disabled after fc99325b

Martchus:
The make target now needs to load `Setup.pm` so its dependencies need to be present at build time.

### DIFF
--- a/dist/rpm/openQA.spec
+++ b/dist/rpm/openQA.spec
@@ -125,6 +125,8 @@ ExcludeArch:    %{ix86}
 %{?systemd_requires}
 %if %{with tests}
 BuildRequires:  %{test_requires}
+%else
+BuildRequires:  %{common_requires} %{main_requires}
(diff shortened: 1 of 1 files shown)


baierjan (approved):

dist/rpm/openQA.spec
```diff
@@ -125,6 +125,8 @@ ExcludeArch:    %{ix86}
 %{?systemd_requires}
 %if %{with tests}
 BuildRequires:  %{test_requires}
+%else
+BuildRequires:  %{common_requires} %{main_requires}
```
okurz, line 129: should be handled in dependencies.yaml as part of "build_requires"

okurz (changes requested):

Martchus:
Probably a better alternative: https://github.com/os-autoinst/openQA/pull/5598

PR #5598, opened 2024-04-29
Move asset-related setup into its own file to avoid extra build requires

Martchus:
Otherwise we need many additional dependencies at build time. This is notable when those dependencies are not already pulled-in anyways for running tests and thus will fix builds on architectures where we have tests disabled.

### DIFF
--- a/lib/OpenQA/Assets.pm
+++ b/lib/OpenQA/Assets.pm
@@ -1,0 +1,48 @@
+# Copyright SUSE LLC
+# SPDX-License-Identifier: GPL-2.0-or-later
+
+package OpenQA::Assets;
+use Mojo::Base -strict, -signatures;
+
+# This file contains helpers to setup handling of assets of the web UI. The list function is used at install-time.
+
+use Mojolicious;
+use Mojo::File qw(path);
+use Mojo::Home;
+use Mojolicious::Plugin::AssetPack;
+use YAML::PP qw(LoadFile);
+
+sub setup ($server) {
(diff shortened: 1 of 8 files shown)


okurz (approved):
This looks very clean though might fail tests still :)

baierjan (approved):

lib/OpenQA/Assets.pm
```diff
@@ -0,0 +1,48 @@
+# Copyright SUSE LLC
+# SPDX-License-Identifier: GPL-2.0-or-later
+
+package OpenQA::Assets;
+use Mojo::Base -strict, -signatures;
+
+# This file contains helpers to setup handling of assets of the web UI. The list function is used at install-time.
+
+use Mojolicious;
```
okurz, line 9: as you directly use Mojolicious you should also need to explicitly specify that

okurz (approved):

okurz:
From https://build.opensuse.org/package/live_build_log/devel:openQA:GitHub:os-autoinst:openQA:PR-5598/openQA:openQA-test/openSUSE_Tumbleweed/x86_64

[   20s] + perl -I/usr/share/openqa/lib -mMojolicious -mMojo::Home -mOpenQA::Setup -e 'OpenQA::Assets::setup(Mojolicious->new(home => Mojo::Home->new("/usr/share/openqa")))'
[   21s] Undefined subroutine &OpenQA::Assets::setup called at -e line 1.

Martchus:
Builds for all archs look good again.

PR #5605, opened 2024-04-29
Remove obsoleted and non-working makefile targets and outdated documentation

baierjan:
Remove obsoleted and non-working makefile targets, those targets were either pulling an obsoleted container or failed to build a custom one. Use-case of those targets is covered by pulling openqa_devel container and running make run-tests-within-container within that container.

Remove out of date documentation about tests in non-CI containers. Most of this is already outdated and the relevant parts are covered in the circleci-local-container section and were mostly redundant.

Reference: https://progress.opensuse.org/issues/159747

### DIFF
--- a/docs/Contributing.asciidoc
+++ b/docs/Contributing.asciidoc
@@ -435,7 +435,7 @@ and run `make update-deps`.  This will generate the `cpanfile` and
 `dist/rpm/openQA.spec` files.
 
 The same applies to `os-autoinst` where `make update-deps` will generate the
-`cpanfile`, `os-autoinst.spec` and `container/ci/Dockerfile`.
+`cpanfile`, `os-autoinst.spec` and `container/os-autoinst_dev/Dockerfile`.
 
 If changing any package dependencies make sure packages and updated packages
 are available in openSUSE Factory and whatever current Leap version is in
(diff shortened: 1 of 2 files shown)


docs/Contributing.asciidoc
```diff
@@ -606,92 +606,6 @@ To keep the test database running after executing tests with the `Makefile`, add
 
 To watch the execution of the UI tests, set the environment variable `NOT_HEADLESS`.
 
-=== Run tests within a container
```
okurz, line 609: That's quite a big section you remove so I think there is some value in it which I think we should keep. Can we instead slim it down and keep something? Just because nobody in our team uses those tests that might still provide an easy entry point for some, isn't it?

okurz (changes requested):

docs/Contributing.asciidoc
baierjan: Yes, I also wasn't sure if I should delete so much. This can be probably kept, because it is not directly tied to the old removed container.

docs/Contributing.asciidoc
```diff
@@ -656,39 +645,17 @@ So by replacing VAR1 and VAR2 with those values one can trigger the different te
...
+  podman run --rm -ti --name openqa-testsuite -v /tmp/.X11-unix:/tmp/.X11-unix:rw -e DISPLAY="$DISPLAY" -e NOT_HEADLESS=1 openqa_devel:latest prove -v t/ui/14-dashboard.t
   xhost -local:root
 
 It is also possible to use a custom os-autoinst checkout using the following arguments:
 
   podman run … -e CUSTOM_OS_AUTOINST=1 -v /path/to/your/os-autoinst:/opt/os-autoinst make run-tests-within-container
 
-By default, `configure` and `make` are still executed (so a clean checkout is expected). If your checkout is already prepared to use,
-set `CUSTOM_OS_AUTOINST_SKIP_BUILD` to prevent this. Be aware that the build produced outside of the container might not work inside the
-container if both environments provide different, incompatible library versions (eg. OpenCV).
```
Martchus, line 687: The  `run-tests-within-container` script still contains the logic regarding `CUSTOM_OS_AUTOINST_SKIP_BUILD`. The remark about builds being possibly incompatible is probably also still valid.

Martchus (changes requested):

docs/Contributing.asciidoc
baierjan: Ah, ok. So this is not the entrypoint script feature but `run-tests-within-container` feature. I will check if that is applicable for the new image as well and adapt/keep it then.

baierjan:
So I guess now I am removing only the non-relevant parts of the documentation.

Martchus (approved):

okurz (approved):
Well, let's try :)

PR #5608, opened 2024-04-30
Add ShellCheck directive for SC2153

b10n1k:
Details of the error https://www.shellcheck.net/wiki/SC2153. Hits on `${VERSION%% *}` variable since ShellCheck version 0.9.0.


Martchus:
Sorry, but I was faster :-)

(See https://github.com/os-autoinst/openQA/pull/5607, and I also mentioned that PR in the chat immediately after creating it.)

PR #5610, opened 2024-05-02
Remove containers not mentioned anywhere

baierjan:
Those containers are not mentioned anywhere anymore. The ci container is already replaced by the devel:openQA:ci/base container, the other was supposed to act as a base for locally build openqa_dev container which is already superseded by openqa_devel.

Reference: https://progress.opensuse.org/issues/159747

The above will be true right after #5605 is merged.


Martchus (approved):
I thought this might still be used somewhere but couldn't find it. So you're probably right :-)

baierjan:
> I thought this might still be used somewhere but couldn't find it. So you're probably right :-)

I hope I deleted every evidence 😄

kalikiana (approved):

PR #5611, opened 2024-05-02
Add recording how to run openQA test in 5 minutes

baierjan:
Add scenario for asciinema recording and include it into documentation.

Add Dockerfile and a short Makefile to allow generating GIFs out of autocast yaml files with the asciicast + agg combo without installing anything. The cast will be done inside the container as well, so the recording is done isolated from the rest of the system.

Reference: https://progress.opensuse.org/issues/76978

### DIFF
--- a/docs/asciinema/openqa-in-5-minutes.yaml
+++ b/docs/asciinema/openqa-in-5-minutes.yaml
@@ -5,11 +5,21 @@ settings:
   title: How to run an openQA test in 5 minutes
   timeout: 300s
   type_speed: 50ms
-  environment:
-    - name: BAT_THEME
-      value: Solarized (light)
+  environment_capture:
+    - BAT_THEME
 instructions:
   - !Command
+    command: "# Install podman"
+  - !Command
+    command:
+      - zypper in -y podman
+  - !Command
+    # Workaround to hide some warnings due to having podman inside podman
+    command:
+      - alias podman='podman --log-level error'
+    hidden: true
+  - !Command
+    # Make sure the image will be downloaded, because that looks nice in the video
     command:
       - podman rmi
       - registry.opensuse.org/devel/openqa/containers/openqa-single-instance
@@ -23,6 +33,11 @@ instructions:
       - --env skip_suse_tests="" --env skip_suse_specifics=""
       - --device /dev/kvm -p 1080:80 -p 1443:443
       - registry.opensuse.org/devel/openqa/containers/openqa-single-instance
+  - !Command
+    # Workaround for healtchecks not working without systemd inside recording container
+    command: >-
+      ( nohup retry -r 20 -s 10 -- podman healthcheck run openqa &> /dev/null & )
+    hidden: true
   - !Command
     command: podman wait --condition healthy openqa
   - !Marker Clone job from openqa.opensuse.org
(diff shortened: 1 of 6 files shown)


kalikiana (approved):
Neat

docs/asciinema/openqa-in-5-minutes.yaml
```diff
@@ -0,0 +1,102 @@
...
+    command: "# Schedule the product"
+  - !Command
+    command:
+      - podman exec -ti openqa
+      - openqa-cli schedule --monitor
+      - --param-file SCENARIO_DEFINITIONS_YAML=scenario-definitions.yaml
+      - DISTRI=example VERSION=0 FLAVOR=DVD ARCH=x86_64
+      - TEST=simple_boot _GROUP_ID=0 BUILD=test
+      - CASEDIR=https://github.com/os-autoinst/os-autoinst-distri-example.git
+      - NEEDLES_DIR=%%CASEDIR%%/needles
```
okurz, line 79: Is this actually necessary? I thought this simple is actually the default?

baierjan: I don't know, I have it directly from https://github.com/os-autoinst/os-autoinst-distri-openQA/blob/master/tests/osautoinst/start_test.pm#L39 where I would expect the "most correct" way of doing that.

Martchus: > I thought this simple is actually the default?

I don't think this is the default and if we would make it the default it would maybe break some user's workflow.

Martchus (changes requested):
Adding 5.5 MiB of binary data to our repo's master branch is maybe not so nice. It would be nice if we could offload the binary to the gh-pages branch (which has a completely separate chain of commits).

(I'm only requesting changes here so we have at least some time to discuss this. If you all think having 5.5 MiB on the master branch is ok then just dismiss my review. Otherwise the change looks good to merge.)

baierjan:
> Adding 5.5 MiB of binary data to our repo's master branch is maybe not so nice. It would be nice if we could offload the binary to the gh-pages branch (which has a completely separate chain of commits).
> 
> (I'm only requesting changes here so we have at least some time to discuss this. If you all think having 5.5 MiB on the master branch is ok then just dismiss my review. Otherwise the change looks good to merge.)

Yes, that is actually a very good point. Having a (rather large) gif file is more portable, but it makes the repo and also the page (which then loads a little bit slower). I was also thinking about using the `.cast` file and embed the asciicast javascript player into the documentation, but that would on the other hand introduce external dependency.

As we might be adding more of this later, we might want to have the optimal way to handle it. Do we have any convenient way how to use the gh-pages branch? I mean I can easily push the recording there but it would be then a little bit awkward to keep track if the version there is "in-sync" with the source in the main branch.

docs/asciinema/openqa-in-5-minutes.yaml
okurz: K, fine

baierjan:
I created https://github.com/os-autoinst/openQA/pull/5616 which should include the recording directly into pages. The GIF is removed from this pull request and instead a gitignore is introduced to not accidentally push the recordings here. Will that work for you @Martchus ?

Martchus (approved):

docs/asciinema/openqa-in-5-minutes.yaml
```diff
@@ -0,0 +1,102 @@
+---
```
perlpunk, line 1: Where can I find a documentation of this file format? I searched https://docs.asciinema.org/ but could not find anything

baierjan: And this is where I was able to find it initially: https://docs.asciinema.org/integrations/?h=autocast

PR #5615, opened 2024-05-06
Fix regular expression in grep

baierjan:
Fix "grep: warning: stray \ before @"; as `@` is not a special character in regex it does not need to be escaped.


kalikiana (approved):
Nice catch!

perlpunk (approved):

Martchus (approved):

PR #5616, opened 2024-05-06
Add recording into documentation

baierjan:
Recording based source from commit 5cb881b9e

Reference: https://progress.opensuse.org/issues/76978


kalikiana:
So this is only the .gif from #5611? Any reason we can't generate the files like other docs and push to gh-pages instead of manually adding it?

baierjan:
> So this is only the .gif from #5611?

Yes

> Any reason we can't generate the files like other docs and push to gh-pages instead of manually adding it?

Generate where? Generating it involves running a podman container, doing some stuff inside and the creating a gif file from the result. And you want to see the gif before pushing as it might contains stuff you do not want (in case of errors and such).

In theory it is possible to generate the gif from the intermediate cast file, as that is no longer "live". But then we would need to keep in-sync the cast file and introduce more dependencies into the documentation pipeline. Not sure if that is a win.

Martchus (approved):
I think it is totally fine that the Gif is generated and tested locally as needed  (in some containerized environment).

kalikiana (approved):
> Generate where? Generating it involves running a podman container, doing some stuff inside and the creating a gif file from the result. And you want to see the gif before pushing as it might contains stuff you do not want (in case of errors and such).

We run various things in containers in GHA and CircleCI so it doesn't seem that far-fetched to me. But I get that setting that up right now might not be worth the hassle. So no strong objection.

PR #5621, opened 2024-05-06
Adjust bootstrap to 5.x

b10n1k:
Remove asset files not in current version and adjust some CSS variables as per documentation.


kalikiana:
https://progress.opensuse.org/issues/159408 I'm guessing

kalikiana:
Alternative to #5514? In the Unblock the q came up if we can't re-use the existing PR

PR #5633, opened 2024-05-10
CI: Workaround broken uploader validation after upgrade to Leap 15.6

okurz:
See
* #5632

Alternative to
* #5634

Related progress issue: https://progress.opensuse.org/issues/157984


.circleci/config.yml
```diff
@@ -263,7 +263,13 @@ jobs:
...
       - codecov/upload:
           file: cover_db/codecov.json
-          xtra_args: -v -Z
+          cli_args: -v -Z
+          # Workaround broken uploader validation after upgrade to Leap 15.6
+          # base
+          # See https://github.com/os-autoinst/openQA/pull/5632
+          # and
+          # https://progress.opensuse.org/issues/157984
+          validate: false
```
baierjan, line 272: https://github.com/codecov/codecov-circleci-orb/issues/157#issue-1692263650 suggests we might be just missing a `~/.gnupg` folder and something like

```diff
       - run: make coverage-report-codecov
+      - run:
+          name: Quick fix for GPG error in Codecov
+          command: mkdir -p ~/.gnupg
       - codecov/upload:
```

should help.

okurz: hm, doubt that. https://app.circleci.com/pipelines/github/os-autoinst/openQA/13512/workflows/ba9f5742-b89d-460f-93a6-38b49cdbef17/jobs/126791?invite=true#step-114-17 says

```
gpg: directory '/home/squamata/.gnupg' created
gpg: /home/squamata/.gnupg/trustdb.gpg: trustdb created
```

before the error so the directory exists.

baierjan: I know, but the same applies for the comment I linked.

baierjan: To express myself better, the directory needs to exists before `codecov/upload` creates it itself because then it will fallback to use `keyboxd` and fails.

okurz: The "last good" https://app.circleci.com/pipelines/github/os-autoinst/openQA/13507/workflows/57e98e70-3b5d-46d0-bcfd-52de4009e9ce/jobs/126714?invite=true#step-114-4 shows that `gpg --no-default-keyring --keyring trustedkeys.gpg --import` is called with the parameter `--keyring` which is interesting because on Leap 15.5 with gpg 2.2.27 `gpg --help` doesn't mention `--keyring` as an allowed command or option but the man page does. On Leap 15.6 we have gpg 2.4.4 where the manpage still mentions `--keyring` but the command seems to be have different as if the `--keyring` parameter would not be effective.

baierjan: Hm, yes... The man page also mentions

> Note that if the option use-keyboxd is enabled in ‘common.conf’, no keyrings are used at all and keys are all maintained by the keyboxd process in its own database.

Maybe the `--keyring` option is not effective because gpg decides to use keyboxd, which might be actually the default as documentation says:

> common.conf
> This is an optional configuration file read by gpg on startup. It may contain options pertaining to all components of GnuPG. Its current main use is for the "use-keyboxd" option. If the default home directory ~/.gnupg does not exist, GnuPG creates this directory and a common.conf file with "use_keyboxd".

okurz: anyway, trying https://github.com/os-autoinst/openQA/pull/5634 nevertheless as I could not easily reproduce the problem with like `for i in 15.5 15.6; do podman run --rm -it registry.opensuse.org/opensuse/leap:$i sh -c 'gpg --version; gpg --no-default-keyring --keyring trustedkeys.kbx --import < /dev/null ; ls /root/.gnupg/'; done`

baierjan: Beware, `/root/.gnupg` already exists in the container. Try 
```
for i in 15.5 15.6; do podman run --rm -it registry.opensuse.org/opensuse/leap:$i sh -c 'rm -r /root/.gnupg; gpg --version; gpg --no-default-keyring --keyring trustedkeys.kbx --import < /dev/null ; ls /root/.gnupg/'; done
```

Than the different behavior is quite obvious:
```
gpg (GnuPG) 2.2.27
libgcrypt 1.9.4-unknown
Copyright (C) 2021 Free Software Foundation, Inc.
License GNU GPL-3.0-or-later <https://gnu.org/licenses/gpl.html>
This is free software: you are free to change and redistribute it.
There is NO WARRANTY, to the extent permitted by law.

Home: /root/.gnupg
Supported algorithms:
Pubkey: RSA, ELG, DSA, ECDH, ECDSA, EDDSA
Cipher: IDEA, 3DES, CAST5, BLOWFISH, AES, AES192, AES256, TWOFISH,
        CAMELLIA128, CAMELLIA192, CAMELLIA256
Hash: SHA1, RIPEMD160, SHA256, SHA384, SHA512, SHA224
Compression: Uncompressed, ZIP, ZLIB, BZIP2
gpg: directory '/root/.gnupg' created
gpg: keybox '/root/.gnupg/trustedkeys.kbx' created
gpg: no valid OpenPGP data found.
gpg: Total number processed: 0
trustedkeys.kbx
gpg (GnuPG) 2.4.4
libgcrypt 1.10.3
Copyright (C) 2024 g10 Code GmbH
License GNU GPL-3.0-or-later <https://gnu.org/licenses/gpl.html>
This is free software: you are free to change and redistribute it.
There is NO WARRANTY, to the extent permitted by law.

Home: /root/.gnupg
Supported algorithms:
Pubkey: RSA, ELG, DSA, ECDH, ECDSA, EDDSA
Cipher: IDEA, 3DES, CAST5, BLOWFISH, AES, AES192, AES256, TWOFISH,
        CAMELLIA128, CAMELLIA192, CAMELLIA256
Hash: SHA1, RIPEMD160, SHA256, SHA384, SHA512, SHA224
Compression: Uncompressed, ZIP, ZLIB, BZIP2
gpg: directory '/root/.gnupg' created
gpg: no valid OpenPGP data found.
gpg: Total number processed: 0
common.conf
```

baierjan: Hard to label that a product bug when the different (new) behavior is documented and affects new installations only. But maybe we can adapt and incorporate the fix directly into our image: https://github.com/os-autoinst/openQA/pull/5635

PR #5634, opened 2024-05-10
CI: Fix broken uploader validation after upgrade to Leap 15.6

okurz:
See https://github.com/os-autoinst/openQA/pull/5632
and https://github.com/codecov/codecov-circleci-orb/issues/157

Related progress issue: https://progress.opensuse.org/issues/157984


.circleci/config.yml
```diff
@@ -261,9 +261,19 @@ jobs:
...
       - codecov/upload:
           file: cover_db/codecov.json
-          xtra_args: -v -Z
+          cli_args: -v
+          # Workaround broken uploader validation after upgrade to Leap 15.6
+          # base
+          # See https://github.com/os-autoinst/openQA/pull/5632
+          # and
+          # https://progress.opensuse.org/issues/157984
+          validate: false
```
baierjan, line 276: Not sure if you already tried that (in that case please disregard this suggestion), but if the linked comment is not wrong, we might be able to keep the validation with the the `mkdir` workaround.

```suggestion
```

okurz: of course, that was my idea with calling this PR "fixed". Got confused here as well :/ Thx.

baierjan (approved):
Feel free to merge as is. The workaround can be incorporated into the base image at any point as you have `-p` here.

PR #5647, opened 2024-05-17
Add verification run scheduler

ByteOtter:
# What?

Add a script which wraps the standard use cases for `openqa-clone-custom-git-refspec` into a more usable form. Instead of requiring the user to type out the URLs of the `openqa.opensuse.org` test they want to use it injects all information - like the PR number, base test ID, test modules to load, etc. - into a `openqa-clone-custom-git-refspec` call.

This script itself does not implement new functionality or behaviour, but is simply a different "front-end" - with a more descriptive name - to use for quickly scheduling verification runs for PRs against e.g `os-autoinst`.

This is just a thought experiment and could theoretically be integrated into other components without the need for a separate script.

# How?

Several parameters are mandatory for the script to function:

- `$SOURCE` - Which specifies whether to use `openqa.opensuse.org` or `openqa.suse.de` (with custom targets potentially supported via config file)
- `$PR_ID` - The ID the GitHub PR
- `$BASE_TEST_ID` - The ID of the test the user wants to use as a base
- `$[test/module,...]` - A comma sperated list of test modules to load
- `$TEST_NAME` - The `TEST` parameter for the call

The GitHub access token is provided via a config file at `~/.config/gh.conf` (name temporary).

# Why?

This PR aims to lower the barrier of entry for new contributors and make general test writing easier by having a easy-to-use, descriptively named script a user can call to provide a run for their PR. Instead of going around trying to find the right command.


foursixnine:
Following the discussion slack: 

- This could be integrated inside [clone-custom-git-refspec](https://github.com/os-autoinst/openQA/blob/master/script/openqa-clone-custom-git-refspec) directly. That’s more due to how we already work, and I fear introducing a new script might not be 100% the best idea, but its a start already
- The handling of github tokens is a great adition!
- on the spec file, perhaps once we figure out what to do, it could be aliased to `openqa-verify-pr` or a nice short name :D (because `openqa-clone-custom-git-refspec` is... well, a bit long

script/openqa-verify_pr
```diff
@@ -0,0 +1,126 @@
...
+if [ "$source" = "opensuse" ]; then
+  echo "Dispatching verification run for opensuse..."
+  openqa-clone-custom-git-refspec "https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/$pr_id" \
+    "https://openqa.opensuse.org/tests/$test_id" \
+    SCHEDULE="tests/installation/bootloader_start,tests/boot/boot_to_desktop,$modules" \
+    TEST="$test_name"
+elif [ "$source" = "suse" ]; then
+  echo "Dispatching verification run for suse..."
+  openqa-clonse-custom-git-refspec "https://github.com/os-autoinst-distri-opensuse/pull/$pr_id" \
+    "https://openqa.suse.de/tests/$test_id" \
```
kalikiana, line 119: I wonder, are you aware of the existing support for job URL's within the PR description? See https://github.com/os-autoinst/openQA/blob/master/script/openqa-clone-custom-git-refspec#L52

ByteOtter: I actually wasn't aware of this, thanks for making me aware. That'd make integrating parts of the ideas into `openqa-clone-custom-git-refspec` even more alluring to take advantage of this feature. @foursixnine, what do you think?

foursixnine: > I actually wasn't aware of this, thanks for making me aware. That'd make integrating parts of the ideas into `openqa-clone-custom-git-refspec` even more alluring to take advantage of this feature. @foursixnine, what do you think?

Yep its the first point here: https://github.com/os-autoinst/openQA/pull/5647#issuecomment-2117728546

Martchus:
Please capitalize the first letter of the commit message subject lines. It would also be much nicer if `openqa-verify_pr` was `openqa-verify-pr`. The formatting of the help text and general coding style is unfortunately not in-line with what we generally use in our scripts. I would at least add a `set -e` or perhaps even `set -eux -o pipefail` at the beginning.

Note that additional scripts could also be added to https://github.com/os-autoinst/scripts - although if it is supposed to be distributed via an RPM package keeping it here might be the better option.

ByteOtter:
Thanks for your feedback, I have changed the name and added the `set -o pipefail` line. Could you elaborate a bit on what you mean with "general coding style"? What do I need to do differently?
Thanks :)

script/openqa-verify-pr
```diff
@@ -0,0 +1,130 @@
...
+
+source="$1"
+pr_id="$2"
+test_id="$3"
+modules="$4"
+test_name="$5"
+
+config_file="$HOME/.config/openqa/verification.conf"
+config=""
+custom_target=""
```
Martchus, line 81: Unless shellcheck is enforcing them I'd get rid of these useless quotes and write everything in two lines:
```suggestion
source=$1 pr_id=$2 test_id=$3 modules=$4 test_name=$5
config_file=$HOME/.config/openqa/verification.conf config= custom_target=
```

script/openqa-verify-pr
```diff
@@ -0,0 +1,130 @@
...
+
+# See help text.
+if [ "$1" = "-h" ] || [ "$1" = "--help" ]; then
```
Martchus, line 84: In Bash scripts it makes sense to use `[[` which also means that the left-hand-side doesn't need to be quoted. We usually also prefer single-quotes if no variable-substitution is used:
```suggestion
if [[ $1 = '-h' ] || [ $1 = '--help' ]; then
```

The same counts for the other occurrences as well.

script/openqa-verify-pr
```diff
@@ -0,0 +1,130 @@
...
+  config=$(grep "^GH_ACCESS_TOKEN=" "$config_file" | cut -d'"' -f2)
+  custom_target=$(grep "^CUSTOM_TARGET=" "$config_file" | cut -d'"' -f2)
+else
+  echo "No config file found! Run with '-h' for more info."
+  exit 1
+fi
+
+GITHUB_TOKEN=$config
+
+export GITHUB_TOKEN
```
Martchus, line 106:
```suggestion
export GITHUB_TOKEN=$config
```

script/openqa-verify-pr
```diff
@@ -0,0 +1,130 @@
...
+# 
+# Example
+#
+# Pass arguments to this script like this:
+#
+# schedule_verification_run opensuse TOKEN 1234 12345 tests/console/rustup,tests/console/cargo RUST
+
+
+help_text=$(cat <<'EOF'
+OpenQA verification run scheduler.
```
Martchus, line 39:
```suggestion
openQA verification run scheduler
```

script/openqa-verify-pr
```diff
@@ -0,0 +1,130 @@
...
+
+Usage:
+ schedule_verification_run $SOURCE $PR_ID $BASE_TEST_ID $[test/module, test/module,...] $TEST_NAME
```
Martchus, line 42:
```suggestion
 openqa-verify-pr $SOURCE $PR_ID $BASE_TEST_ID $[test/module, test/module,...] $TEST_NAME
```

script/openqa-verify-pr
```diff
@@ -0,0 +1,130 @@
+#!/bin/bash
+# Copyright 2024 (c) SUSE LLC
+# SPDX-License-Identifier: GPL-2.0-or-later
+
+# Schedule a verification run for a os-autoinst PR
+# USAGE:
+#
+# schedule_verification_run $SOURCE $PR_ID $BASE_TEST_ID $[test/module, test/module, ...] $TEST_NAME
```
Martchus, line 8:
```suggestion
# openqa-verify-pr $SOURCE $PR_ID $BASE_TEST_ID $[test/module, test/module, ...] $TEST_NAME
```

But considering this is redundant to the help text I would just remove this lengthy comment-block completely.

script/openqa-verify-pr
```diff
@@ -0,0 +1,130 @@
+#!/bin/bash
+# Copyright 2024 (c) SUSE LLC
+# SPDX-License-Identifier: GPL-2.0-or-later
```
Martchus, line 3:
```suggestion
# Copyright SUSE LLC
# SPDX-License-Identifier: GPL-2.0-or-later
```

script/openqa-verify-pr
```diff
@@ -0,0 +1,130 @@
...
+
+help_text=$(cat <<'EOF'
+OpenQA verification run scheduler.
+
+Usage:
+ schedule_verification_run $SOURCE $PR_ID $BASE_TEST_ID $[test/module, test/module,...] $TEST_NAME
+
+Parameters:
+ SOURCE - The source you want to schedule a run for. Can be either opensuse or suse.
+ PR_ID - The ID of your GitHub Pull Request.
```
Martchus, line 46:
```suggestion
 PR_ID - The ID of your GitHub pull request.
```
(in accordance with https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/proposing-changes-to-your-work-with-pull-requests/about-pull-requests)

script/openqa-verify-pr
```diff
@@ -0,0 +1,130 @@
...
+
+
+help_text=$(cat <<'EOF'
+OpenQA verification run scheduler.
+
+Usage:
+ schedule_verification_run $SOURCE $PR_ID $BASE_TEST_ID $[test/module, test/module,...] $TEST_NAME
+
+Parameters:
+ SOURCE - The source you want to schedule a run for. Can be either opensuse or suse.
```
Martchus, line 45:
```suggestion
 SOURCE - The source you want to schedule a run for. Can be either openSUSE or SUSE.
```

script/openqa-verify-pr
```diff
@@ -0,0 +1,130 @@
...
+Specify your GitHub access token in '~/.config/openqa/verification.conf' like this:
+
+GH_ACCESS_TOKEN="YOUR_TOKEN_HERE"
+
+To specify the URL to your own openQA instance specify it in the same config file like this:
+
+CUSTOM_TARGET="https://openqa.yourdomain.com"
+
+Default test modules:
+The following modules are automatically selected to load for your run:
```
Martchus, line 60: Two consecutive phrases ending with a colon are a bit weird.

script/openqa-verify-pr
```diff
@@ -0,0 +1,130 @@
...
+pr_id="$2"
+test_id="$3"
+modules="$4"
+test_name="$5"
+
+config_file="$HOME/.config/openqa/verification.conf"
+config=""
+custom_target=""
+
+# See help text.
```
Martchus, line 83:
```suggestion
```

script/openqa-verify-pr
```diff
@@ -0,0 +1,130 @@
...
+if [ "$1" = "-h" ] || [ "$1" = "--help" ]; then
+  echo "$help_text"
+  exit 0
+fi
+
+# If too few or too many arguments are supplied -> die.
```
Martchus, line 89:
```suggestion
# Validate argument count
```

Martchus (commented):
Generally we use a more compact coding style. I gave a few concrete suggestions in inline-comments.

script/openqa-verify-pr
ByteOtter: In this case I wanted to print the actual values you can pass to the script. but it's outdated anyway so I'll change it

script/openqa-verify-pr
ByteOtter: These quotes were warnings of my LSP but your solution seems more compact. I will change it

script/openqa-verify-pr
ByteOtter: fixed it

script/openqa-verify-pr
ByteOtter: fixed it 👍🏼

script/openqa-verify-pr
ByteOtter: Leftover from what I had `Default test modules` as a heading for that section. I'll change that.

script/openqa-verify-pr
ByteOtter: fixed it 👍🏼

script/openqa-verify-pr
Martchus: Those quotes are really not required. The only annoyance can be that shellcheck will complain (needlessly). So I would check that first.

script/openqa-verify-pr
ByteOtter: fixed 👍🏼

script/openqa-verify-pr
ByteOtter: fixed 👍🏼

script/openqa-verify-pr
ByteOtter: Marked `opensuse` and `suse` with single quotes to underline them being values to pass rather than simple name listing

script/openqa-verify-pr
ByteOtter: fixed it

script/openqa-clone-custom-git-refspec
```diff
@@ -32,6 +32,9 @@ set -o pipefail
 if [[ -n "$GITHUB_TOKEN" ]]; then
     AUTHENTICATED_REQUEST=" -u $GITHUB_TOKEN:x-oauth-basic"
     echo "Github oauth token provided, performing authenticated requests"
+elif [[ -f "$HOME/.config/openqa/verification.conf" ]]; then
```
foursixnine, line 35:
```suggestion
elif [[ -f "$HOME/.config/openqa/github-token.conf" ]]; then
```

script/openqa-clone-custom-git-refspec
```diff
@@ -55,6 +58,16 @@ extract_urls_from_pr() {
     fi
 }
 
+# Get the Github Token from a config file
+# located at ~/.config/openqa/verification.conf
+get_token_from_file() {
+  local config_file="$HOME/.config/openqa/verification.conf"
```
foursixnine, line 64: See above, also provide an example of the format in the help file.

script/openqa-verify-pr
```diff
@@ -0,0 +1,119 @@
...
+  exit 1
+fi
+
+export GITHUB_TOKEN=$config
+
+if [ $source = 'opensuse' ]; then
+  echo "Dispatching verification run for opensuse..."
+  openqa-clone-custom-git-refspec "https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/$pr_id" \
+    "https://openqa.opensuse.org/tests/$test_id" \
+    SCHEDULE="tests/installation/bootloader_start,tests/boot/boot_to_desktop,$modules" \
```
foursixnine, line 101: Remember to move this to the  [openqa-clone-custom-git-refspec](https://github.com/os-autoinst/openQA/pull/5647/files#diff-1da53224de38473695839b4d4c7efee7a2add51a1aa9a3c259e872a5663b9498) behind a switch i.e `--modules a,b,c` so it gets appended via the for loop, before calling cone job.

and then there's the consideration that if the caller is passing:
- YAML_SCHEDULE + SCHEDULE, die (as one of them has to be overwritten.
- if the switch is passed, and so is either (yaml_schedule or schedule) die, as options aren't compatible

ByteOtter: Yep, currently looking into that :)

script/openqa-clone-custom-git-refspec
ByteOtter: fixed 👍🏼

script/openqa-clone-custom-git-refspec
```diff
@@ -157,6 +182,12 @@ Please try 'curl $json_url' or select another job, e.g. in the same scenario: $h
     local scriptdir
     scriptdir=$(dirname "${BASH_SOURCE[0]}")
     local cmd="$dry_run $scriptdir/openqa-clone-job $clone_args \"$host\" \"$job\" _GROUP=\"$GROUP\" TEST+=\"$test_suffix\" BUILD=\"$build\" CASEDIR=\"$casedir\" PRODUCTDIR=\"$productdir\" NEEDLES_DIR=\"$needles_dir\""
+    if [[ $cmd =~ (SCHEDULE|YAML_SCHEDULE) ]]; then
```
foursixnine, line 185: is this an and? or an or?

ByteOtter: it's an or, my bad. Though changing it to
```bash
if [[ $cmd =~ SCHEDULE ]] && [[ $cmd =~ YAML_SCHEDULE ]]; then
...
```
doesnt seem to change anythin rn so my problem may be that I am using `cmd`?

okurz (changes requested):
As suggested in chat originally I appreciate that you try to bring in new ideas with pull requests but we should be careful about introducing new scripts. And the same probably goes for the modules command line parameter which is redundant considering the support for generic openQA test variables. Before investing more effort into that I recommend you look into this variables concept in general. Maybe for now you want to focus only on the aspect of reading the github token from a config file which seems independent of the other changes

script/openqa-clone-custom-git-refspec
```diff
@@ -12,6 +12,7 @@ Options:
  -v, --verbose          execute with verbose output
  -h, -?, --help         display this help
  -n, --dry-run          execute in dry-run mode, do not clone any openQA jobs
+ -m, --modules          specify the tets modules you want to load as a comma-separated list of paths to your test files
```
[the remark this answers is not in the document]
ByteOtter: To separate the concerns of the `--modules` flag and the token handling, I extracted the token handling into its own PR https://github.com/os-autoinst/openQA/pull/5666 for now.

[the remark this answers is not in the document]
foursixnine: This is added as a comodity command, the schedule, include_modules and exclude modules are like a shotgun: You can blow your foot if you want to.

`--modules` is more like a hammer, you can still smash your fingers, but won't have you hunting for reasons why your forced schedule is not being taken into account. See line 200.

PR #5648, opened 2024-05-17
Add devcontainer for Codespaces

perlpunk:
Issue: https://progress.opensuse.org/issues/87695

This PR also restructures openqa-bootstrap a bit.

To try it out, go to https://codespaces.new/os-autoinst/openQA/tree/codespace?quickstart=1 and create a new (or resume an existing) codespace. That can take a while.

## The current usage

openqa-bootstrap will automatically start in the background and write to `/var/log/openqa-bootstrap.log`.

Open a terminal:
```
tail -f /var/log/openqa-bootstrap.log
```

After the webserver has started at port 80, you should be able to open the url in the browser.
Now you can start another codespaces terminal and type
```
openqa-clone-job https://openqa.opensuse.org/tests/4191283
```
for example.

When your codespace is stopped after inactivity, you can resume it (that will only take a few seconds), but then you have to start the daemons (database, webui etc.) again with:
```
/usr/share/openqa/script/openqa-bootstrap start
```

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -1,0 +1,26 @@
+{
+    "name": "openQA Single Instance",
+    "image": "registry.opensuse.org/devel/openqa/containers15.5/openqa-single-instance",
+    "runArgs": [ "--privileged", "--device", "/dev/kvm", "--entrypoint", "bash" ],
+    "postCreateCommand": "chown root:kvm /dev/kvm",
+    "containerEnv": {
+      "VNCPORT_OFFSET": "100"
+    },
+    "postStartCommand": "/usr/share/openqa/script/openqa-bootstrap >/var/log/openqa-bootstrap.log 2>&1"
+
+
+    // Features to add to the dev container. More info: https://containers.dev/features.
+    // "features": {},
+
+    // Use 'forwardPorts' to make a list of ports inside the container available locally.
+    // "forwardPorts": [],
+
+    // Use 'postCreateCommand' to run commands after the container is created.
+    //"postCreateCommand": "pip3 install --user -r requirements.txt"
+
+    // Configure tool-specific properties.
+    // "customizations": {},
+
+    // Uncomment to connect as root instead. More info: https://aka.ms/dev-containers-non-root.
+    // "remoteUser": "root"
+}
--- a/README.asciidoc
+++ b/README.asciidoc
@@ -33,6 +33,12 @@ link:docs/UsersGuide.asciidoc[Users Guide].
 If you are interested in writing tests using openQA read the
 link:docs/WritingTests.asciidoc[Tests Developer Guide].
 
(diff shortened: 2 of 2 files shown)


perlpunk:
Waiting for #5654

.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,26 @@
+{
+    "name": "openQA Single Instance",
+    "build": {
+        "dockerfile": "Dockerfile"
```
kalikiana, line 4: Can we use an image here? Since it takes a while to install everything from scratch. Something like this:

```suggestion
        "image": "registry.opensuse.org/devel/openqa/containers/openqa-single-instance"
```

.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,26 @@
...
+    // "forwardPorts": [],
+
+    // Use 'postCreateCommand' to run commands after the container is created.
+    //"postCreateCommand": "pip3 install --user -r requirements.txt"
+
+    // Configure tool-specific properties.
+    // "customizations": {},
+
+    // Uncomment to connect as root instead. More info: https://aka.ms/dev-containers-non-root.
+    // "remoteUser": "root"
```
kalikiana, line 25: Can we try and use a regular user instead of root?

.devcontainer/devcontainer.json
perlpunk: Sure, that's one of the first things I tried.
Detailed answer in the ticket:
https://progress.opensuse.org/issues/87695?issue_count=54&issue_position=32&next_issue_id=58184&prev_issue_id=159408#note-18

perlpunk: What I also find a bit suboptimal with the single instance container is:
https://build.opensuse.org/projects/devel:openQA/packages/openQA-single-instance-container/files/_service:download_url:Dockerfile?expand=1
It installs openQA packages in the Dockerfile fron the tumbleweed repo:
```
RUN zypper in -y openQA-single-instance openQA-bootstrap \
    qemu-arm qemu-ppc qemu-x86 qemu-tools sudo iputils os-autoinst-distri-opensuse-deps
```
And when I later call openqa-bootstrap it adds the devel:openQA repo and installs this stuff again. Ok, it's quite fast because most dependencies are already there. But why was it done this way?

perlpunk:
#5664 for the Dockerfile change.
After that I can adjust the path to the image (As I don't know yet where it will land)

perlpunk:
You can still try it out here: https://codespaces.new/os-autoinst/openQA/tree/codespace?quickstart=1
openqa-bootstrap will start in the background and write to `/var/log/openqa-bootstrap.log`. open a terminal, look at the log, clone a job from o3, open the browser

Martchus (approved):

kalikiana (approved):

PR #5657, opened 2024-05-23
Allow seq in apparmor scripts needed by openqa-label-known-issues

asdil12:


perlpunk (approved):

baierjan (changes requested):
there already is ` /usr/bin/sed mrix,` on line 155

profiles/apparmor.d/usr.share.openqa.script.openqa
```diff
@@ -156,6 +156,7 @@
   /usr/bin/tac mrix,
   /usr/bin/tail mrix,
   /usr/bin/wget ix,
+  /usr/bin/sed rix,
```
perlpunk, line 159:
```suggestion
  /usr/bin/seq rix,
```

perlpunk (changes requested):

profiles/apparmor.d/usr.share.openqa.script.openqa
baierjan: Ah, so the commit title is right 😄 , please also sort the lines so the sed and seq are next to each other.

perlpunk (approved):

PR #5662, opened 2024-05-27
Overhaul info panel on job details page

Martchus:
* Remove `… (0)` for jobs with undetermined runtime
* Use proper buttons to make it less likely for users to click on the wrong button
* Improve appearance of help popovers next to buttons
* State explicitly that re-scheduling is not limited to the current job group
* Move buttons to their own section at the right
* See https://progress.opensuse.org/issues/138593

---

The buttons are no actually proper bootstrap buttons. The help-popovers are also proper bootstrap buttons (part of the same button group). When focusing/hovering over the button it also becomes clear that the help-popover is actually a separate button.

screenshot_%Y%M%D_%H%m%S

Of course not always all actions are displayed and then the actions section at the right takes less space:
screenshot_%Y%M%D_%H%m%S-1

The menu with additional options for dependencies is also still present (if relevant):
screenshot_%Y%M%D_%H%m%S-2

On very small screen sizes actions are displayed below:
screenshot_%Y%M%D_%H%m%S-3

It also looks acceptable when darkmode is enabled (but not the rest of the page so this is a bit pointless at the moment).

---

~~Addition a confirmation dialog before re-triggering the related scheduled product (as of this job) is still to be done.~~ A confirmation dialog was already present. I made the warning more clear (mentioning that it will not be limited to the current job group).

### DIFF
--- a/templates/webapi/test/infopanel.html.ep
+++ b/templates/webapi/test/infopanel.html.ep
@@ -12,14 +12,15 @@
                     </span>
                 % }
             </div>
-            <div id="info-box-content" class="card-body"
+            <div id="info-box-content" class="card-body row"
                 % if ($additional_data) {
                     data-favicon-url-16="<%= favicon_url '-16.png' %>" data-favicon-url-svg="<%= favicon_url '.svg' %>"
                 % }
             >
+                <div class="col-xl">
                 <div>
                     % if ($job->state eq DONE) {
-                        <div class="position-absolute top-right pt-2-and-half pr-3">
+                        <div class="position-absolute top-right badge-pill badge-light">
                             %= $job->passed_module_count
                             <i class='fa module_passed fa-star' title='modules passed'></i>
                             % if ($job->softfailed_module_count) {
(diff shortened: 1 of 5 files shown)


templates/webapi/test/infopanel.html.ep
```diff
@@ -12,14 +12,15 @@
...
+            <div id="info-box-content" class="card-body row"
                 % if ($additional_data) {
                     data-favicon-url-16="<%= favicon_url '-16.png' %>" data-favicon-url-svg="<%= favicon_url '.svg' %>"
                 % }
             >
+                <div class="col-xl">
                 <div>
                     % if ($job->state eq DONE) {
-                        <div class="position-absolute top-right pt-2-and-half pr-3">
+                        <div class="position-absolute top-right badge-pill badge-light">
```
b10n1k, line 23: keep in mind that `badge-pill` is dropped from BS5

b10n1k (approved):

kalikiana (approved):
Looks very cool. Much clearer than before

templates/webapi/test/infopanel.html.ep
Martchus: Ok. I guess we can very easily replace that with our own rules, though.

PR #5663, opened 2024-05-27
Avoid long build numbers from taking too much space on dashboard

Martchus:
Long build numbers might be assigned e.g. to incorporate Git commit hashes. So far they can take a huge amount of the available width so there is not much left for the progress bars next to them. This change limits the maximum width of the build numbers leaving room for the progress bars.

Related ticket: https://progress.opensuse.org/issues/130940

### DIFF
--- a/assets/javascripts/openqa.js
+++ b/assets/javascripts/openqa.js
@@ -119,11 +119,10 @@ function renderDataSize(sizeInByte) {
 }
 
 function alignBuildLabels() {
-  var values = $.map($('.build-label'), function (el, index) {
-    return parseInt($(el).css('width'));
-  });
-  var max = Math.max.apply(null, values);
-  $('.build-label').css('min-width', max + 'px');
+  const max = Math.max(...Array.from(document.getElementsByClassName('build-label')).map(e => e.offsetWidth));
+  const style = document.createElement('style');
+  document.head.appendChild(style);
+  style.sheet.insertRule(`@media (min-width: 1000px) { .build-label { width: ${max}px; } }`);
 }
 
 // reloads the page - this wrapper exists to be able to disable the reload during tests
--- a/assets/stylesheets/dashboard.scss
+++ b/assets/stylesheets/dashboard.scss
@@ -79,6 +79,11 @@ td div.progress {
         margin-top: 2px;
     }
 }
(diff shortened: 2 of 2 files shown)


assets/javascripts/openqa.js
```diff
@@ -119,11 +119,10 @@ function renderDataSize(sizeInByte) {
...
 function alignBuildLabels() {
-  var values = $.map($('.build-label'), function (el, index) {
-    return parseInt($(el).css('width'));
-  });
-  var max = Math.max.apply(null, values);
-  $('.build-label').css('min-width', max + 'px');
+  const max = Math.max(...Array.from(document.getElementsByClassName('build-label')).map(e => e.offsetWidth));
+  const style = document.createElement('style');
+  document.head.appendChild(style);
+  style.sheet.insertRule(`@media (min-width: 1000px) { .build-label { width: ${max}px; } }`);
```
kalikiana, line 125: I guess you're modifying existing logic and there's a good reason to do this in javascript... just strikes me as odd that you're getting values from css to then calculate a width to set as a new rule?

Martchus: Yes, it is odd. It is what the code did before, though. The main thing I change here is introducing a media query that is in-line with the one I added on the CSS-side (so the assigned with doesn't get in the way when the layout is collapsed into one column).

Note that we could probably get rid of this JavaScript code completely by *not* using the layouting/grid of Bootstrap and using instead plain CSS. (At least I wouldn't know how to avoid it when using Bootstrap classes but I have a rough idea how to avoid it when using modern CSS.) However, that would be a greater refactoring so I'm not changing the approach as part of this PR.

Martchus:
Before (not much room for progress bars, and there's a vertical scroll bar):

grafik

After (everything fits and the with used for the build label and progress bars is more balanced):

grafik

(The black bars are just from my terminal in the background. I kept it so you get a better sense of the viewport with which is the same in both screenshots.)

assets/javascripts/openqa.js
kalikiana: Ack. Maybe it's best anyway to see what we need for the new bs upgrade. Was just wondering if this might have been a "actually there is an easy alternative".

kalikiana (approved):

PR #5664, opened 2024-05-28
Downgrade single-instance-container to Leap

perlpunk:
Issue: https://progress.opensuse.org/issues/87695

We talked about the above ticket, and that codespaces has a problem with tumbleweed containers.

Buildung a new container on every codespaces start was considered too slow, so it was suggested to downgrade the single-instance-container so that we can use it in codespace.

This is needed before #5648

### DIFF
--- a/container/single-instance/Dockerfile
+++ b/container/single-instance/Dockerfile
@@ -2,7 +2,7 @@
 #!BuildTag: openqa-single-instance:latest opensuse/openqa-single-instance:latest opensuse/openqa-single-instance:%PKG_VERSION% opensuse/openqa-single-instance:%PKG_VERSION%.%RELEASE%
 
 # hadolint ignore=DL3006
-FROM opensuse/tumbleweed
+FROM opensuse/leap:15.5
 
 # labelprefix=org.opensuse.openqa-single-instance
 LABEL org.opencontainers.image.title="openQA single-instance container"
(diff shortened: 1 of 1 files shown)


Martchus (approved):

kalikiana (approved):

container/single-instance/Dockerfile
```diff
@@ -2,7 +2,7 @@
 #!BuildTag: openqa-single-instance:latest opensuse/openqa-single-instance:latest opensuse/openqa-single-instance:%PKG_VERSION% opensuse/openqa-single-instance:%PKG_VERSION%.%RELEASE%
 
 # hadolint ignore=DL3006
-FROM opensuse/tumbleweed
+FROM opensuse/leap:15.5
```
okurz, line 5: We already upgraded the other containers to 15.6 . I would prefer Tumbleweed as we wouldn't need to care about upgrading those versions but if there are specific problems on Tumbleweed right now then please upgrade to 15.6 and state here why we can't use Tumbleweed

perlpunk: 1. Tumbleweed doesn't work with Github codespaces.
That reminds me that I wanted to create a bugreport for this.
Try it out in this example repo: https://github.com/perlpunk/test-codespace-opensuse-tumbleweed
See also https://progress.opensuse.org/issues/87695#note-18

2. 15.6 doesn't work because of other problems

> I'm now trying out Leap 1.5, because 15.6 did not work:
> During building in OBS, it has a newer version of qemu available than later when the container is running. This seems to be special to Leap 15.6.

https://progress.opensuse.org/issues/87695#note-22

Try it out yourself; build the single instance container in a branch for 15.6 and then run it and run openqa-bootstrap on it. The zypper install will run into a conflict.

PR #5666, opened 2024-05-29
Read GitHub token from config file

ByteOtter:
## What?

This PR adds  a feature to `openqa-clone-custom-git-refspec` to read the GitHub access token from a file in addition to attempting to find a environment variable.

### DIFF
--- a/script/openqa-clone-custom-git-refspec
+++ b/script/openqa-clone-custom-git-refspec
@@ -21,21 +21,19 @@ Examples:
  openqa-clone-custom-git-refspec -n -c '--show-progress' https://github.com/coolgw/os-autoinst-distri-opensuse/tree/nfs https://openqa.opensuse.org/tests/835060 DESKTOP=textmode
  openqa-clone-custom-git-refspec https://github.com/foursixnine/os-autoinst-distri-opensuse/tree/oopsitsbrokenagain https://openqa.opensuse.org/tests/3128467 NEEDLES_DIR='https://github.com/foursixnine/os-autoinst-needles-opensuse.git#oopsitsbrokenagain'
 
-You need to set the environment variable GITHUB_TOKEN to use authenticated requests.
+To use authenticated requests, you need to either provide a GitHub token by
+adding 'GITHUB_TOKEN=...' to a configuration file at
+'github-token.conf' in your $OPENQA_CONFIG folder or set the GITHUB_TOKEN environment variable.
+
+Use this snippet in your 'github-token.conf' to do so:
+
+ GITHUB_TOKEN="YourTokenHere12345"
 
 EOF
     exit "$1"
 }
 
 set -o pipefail
-
-if [[ -n "$GITHUB_TOKEN" ]]; then
-    AUTHENTICATED_REQUEST=" -u $GITHUB_TOKEN:x-oauth-basic"
-    echo "Github oauth token provided, performing authenticated requests"
-fi
-
-curl_github="${curl_github:-"curl${AUTHENTICATED_REQUEST}"}"
-curl_openqa="${curl_openqa:-"curl"}"
 
 fail() {
     echo "$*" >&2
@@ -54,6 +52,22 @@ extract_urls_from_pr() {
         echo "$urls"
     fi
 }
+
+config_file="${OPENQA_CONFIG}/github-token.conf"
+get_token_from_file() {
+    grep "^GITHUB_TOKEN=" "$config_file" | cut -d'"' -f2
+}
+
+if [[ -n "$GITHUB_TOKEN" ]]; then
+    AUTHENTICATED_REQUEST=" -u $GITHUB_TOKEN:x-oauth-basic"
+    echo "GitHub OAuth token provided in environment variable, performing authenticated requests."
+elif [[ -f $config_file ]]; then
+    AUTHENTICATED_REQUEST=" -u $(get_token_from_file):x-oauth-basic"
+    echo "GitHub OAuth token found in config file, performing authenticated requests."
+fi
+
+curl_github="${curl_github:-"curl${AUTHENTICATED_REQUEST}"}"
+curl_openqa="${curl_openqa:-"curl"}"
 
 opts=$(getopt -o vhnc: --long verbose,dry-run,help,clone-job-args: -n "$0" -- "$@") || usage 1
 eval set -- "$opts"


script/openqa-clone-custom-git-refspec
```diff
@@ -55,6 +52,25 @@ extract_urls_from_pr() {
     fi
 }
 
+config_file="$HOME/.config/openqa/github-token.conf"
+get_token_from_file() {
+    if [[ -f $config_file ]]; then
```
dzedro, line 57: I think this is not needed, because it's duplicate of line 66.

script/openqa-clone-custom-git-refspec
```diff
@@ -21,22 +21,19 @@ Examples:
  openqa-clone-custom-git-refspec -n -c '--show-progress' https://github.com/coolgw/os-autoinst-distri-opensuse/tree/nfs https://openqa.opensuse.org/tests/835060 DESKTOP=textmode
  openqa-clone-custom-git-refspec https://github.com/foursixnine/os-autoinst-distri-opensuse/tree/oopsitsbrokenagain https://openqa.opensuse.org/tests/3128467 NEEDLES_DIR='https://github.com/foursixnine/os-autoinst-needles-opensuse.git#oopsitsbrokenagain'
 
-You need to set the environment variable GITHUB_TOKEN to use authenticated requests.
+To use authenticated requests, you need to either provide a GitHub token via a configuration file at '~/.config/openqa/github-token.conf' or set
```
okurz, line 24: Please wrap at 80 characters. Also can you provide the format of the configuration file?

script/openqa-clone-custom-git-refspec
```diff
@@ -21,22 +21,19 @@ Examples:
...
+the GITHUB_TOKEN environment variable.
+
+Use this snippet to do so:
+
+ GH_ACCESS_TOKEN="YourTokenHere12345"
```
Martchus, line 30: This probably needs an `export` or should clarify that it is about the config file. However, I suggest to leave the "snippet" out completely as people probably know how they want to set this environment variable for their purposes and with my other suggestion the use of the config file should be clear enough.

```suggestion
```

okurz (changes requested):

script/openqa-clone-custom-git-refspec
```diff
@@ -21,22 +21,19 @@ Examples:
  openqa-clone-custom-git-refspec -n -c '--show-progress' https://github.com/coolgw/os-autoinst-distri-opensuse/tree/nfs https://openqa.opensuse.org/tests/835060 DESKTOP=textmode
  openqa-clone-custom-git-refspec https://github.com/foursixnine/os-autoinst-distri-opensuse/tree/oopsitsbrokenagain https://openqa.opensuse.org/tests/3128467 NEEDLES_DIR='https://github.com/foursixnine/os-autoinst-needles-opensuse.git#oopsitsbrokenagain'
 
-You need to set the environment variable GITHUB_TOKEN to use authenticated requests.
+To use authenticated requests, you need to either provide a GitHub token via a configuration file at '~/.config/openqa/github-token.conf' or set
+the GITHUB_TOKEN environment variable.
```
Martchus:
```suggestion
To use authenticated requests, you need to either provide a GitHub token by adding 'GH_ACCESS_TOKEN=…' to a configuration file at '~/.config/openqa/github-token.conf' or set
the GITHUB_TOKEN environment variable.
```

script/openqa-clone-custom-git-refspec
ByteOtter: Not sure what else I should provide as the example snippet is all that's necessary 🤔 
Now that I think of it, we could include this into an existing config file like `cllient.conf` tbh. I'll look into that...

Martchus (commented):
It would maybe make sense to also use the key `GITHUB_TOKEN` because that's already used for the env variable and having two very similar keys in different places might be confusing.

script/openqa-clone-custom-git-refspec
ByteOtter: This snippet refers to the sentence above. I have changed the wording to make it more clear that the snippet goes into the `github-token.conf`

script/openqa-clone-custom-git-refspec
```diff
@@ -55,6 +53,23 @@ extract_urls_from_pr() {
...
 
+config_file="$HOME/.config/openqa/github-token.conf"
+get_token_from_file() {
+    token=$(grep "^GITHUB_TOKEN=" "$config_file" | cut -d'"' -f2)
+    echo "$token"
+}
+
+if [[ -n "$GITHUB_TOKEN" ]]; then
+    AUTHENTICATED_REQUEST=" -u $GITHUB_TOKEN:x-oauth-basic"
+    echo "Github oauth token provided in environment variable, performing authenticated requests."
```
Martchus, line 64:
```suggestion
    echo "GitHub OAuth token provided in environment variable, performing authenticated requests."
```

script/openqa-clone-custom-git-refspec
```diff
@@ -55,6 +53,23 @@ extract_urls_from_pr() {
...
+elif [[ -f $config_file ]]; then
+    AUTHENTICATED_REQUEST=" -u $(get_token_from_file):x-oauth-basic"
+    echo "GitHub oauth token found in config file, performing authenticated requests."
```
Martchus, line 67:
```suggestion
    echo "GitHub OAuth token found in config file, performing authenticated requests."
```

script/openqa-clone-custom-git-refspec
```diff
@@ -55,6 +53,23 @@ extract_urls_from_pr() {
     fi
 }
 
+config_file="$HOME/.config/openqa/github-token.conf"
+get_token_from_file() {
+    token=$(grep "^GITHUB_TOKEN=" "$config_file" | cut -d'"' -f2)
+    echo "$token"
```
Martchus, line 59: This should be equivalent, right?
```suggestion
    grep "^GITHUB_TOKEN=" "$config_file" | cut -d'"' -f2
```

script/openqa-clone-custom-git-refspec
ByteOtter: Afaik this is the official spelling of OAuth.
https://en.m.wikipedia.org/wiki/OAuth

script/openqa-clone-custom-git-refspec
ByteOtter: Not sure about if its equivalent, but I personally prefer a explicit return to make the code more expressive. As long as this is not breaking things I would prefer to keep it

okurz: I don't think it makes sense to first assign to a variable and then expand the variable content within a string that only contains the variable and then output that. So I would also prefer the variant that @Martchus proposed

ByteOtter: Applied it 👍🏼

script/openqa-clone-custom-git-refspec
```diff
@@ -55,6 +53,22 @@ extract_urls_from_pr() {
     fi
 }
 
+config_file="$HOME/.config/openqa/github-token.conf"
```
kalikiana, line 56: Please respect `OPENQA_CONFIG` here which is where all other files are stored. See [Customize configuration directory](https://open.qa/docs/#_customize_configuration_directory).

```suggestion
config_file="${OPENQA_CONFIG}/github-token.conf"
```

kalikiana (changes requested):

Martchus (approved):

script/openqa-clone-custom-git-refspec
```diff
@@ -21,22 +21,20 @@ Examples:
  openqa-clone-custom-git-refspec -n -c '--show-progress' https://github.com/coolgw/os-autoinst-distri-opensuse/tree/nfs https://openqa.opensuse.org/tests/835060 DESKTOP=textmode
  openqa-clone-custom-git-refspec https://github.com/foursixnine/os-autoinst-distri-opensuse/tree/oopsitsbrokenagain https://openqa.opensuse.org/tests/3128467 NEEDLES_DIR='https://github.com/foursixnine/os-autoinst-needles-opensuse.git#oopsitsbrokenagain'
 
-You need to set the environment variable GITHUB_TOKEN to use authenticated requests.
+To use authenticated requests, you need to either provide a GitHub token by
+adding 'GITHUB_TOKEN=...' to a configuration file at
+'~/.config/openqa/github-token.conf' or set the GITHUB_TOKEN environment variable.
```
kalikiana, line 26:
```suggestion
'github-token.conf' in your $OPENQA_CONFIG folder or set the GITHUB_TOKEN environment variable.
```

Mirroring the implementation, this shouldn't assume a specific folder. Might also be worth referring to the docs.

kalikiana (changes requested):

script/openqa-clone-custom-git-refspec
kalikiana: https://open.qa/docs/#_customize_configuration_directory

Martchus: Right, it would be great it that feature would also support our usual way to customize the config directory.

kalikiana: It does. It's just documented wrongly now 😁

kalikiana (approved):

PR #5678, opened 2024-06-05
Migrate to Bootstrap 5 with minor UI tweaks and improvements of the darkmode

Martchus:
I kept the first three commits separate because those are from @b10n1k. However, it makes most sense to review the overall diff and checkout the particular commits only for the commit messages.

This PR supersedes https://github.com/os-autoinst/openQA/pull/5621.

Related ticket: https://progress.opensuse.org/issues/159408

PR #5681, opened 2024-06-07
Update version requirement of `rubygem(sass)` to 3.7.4 for Bootstrap 5

Martchus:
Version 3.5.3 provided by Leap cannot handle certain CSS syntax used by Bootstrap 5 leading to errors like:

```
Error: $color: "var(--bs-emphasis-color-rgb)" is not a color for `rgba'
```

Version 3.7.4 from Tumbleweed works. It also works when building the spec file from Tumbleweed against Leap 15.5 or Leap 15.6 so I am setting 3.7.4 as minimum version requirement. Testing out whether 3.6 would be recent enough as well is not worth it.

Related ticket: https://progress.opensuse.org/issues/159408


okurz (approved):

kalikiana (approved):

perlpunk (approved):

perlpunk (commented):
Is that available on Leap 15.5?

Martchus:
I linked it in our devel repos, see https://github.com/os-autoinst/openQA/pull/5678#issuecomment-2154563627. (If it was already in Leap 15.5/15.6 it would already be in the CI container and we would not have noticed the version requirement.)

baierjan (approved):

PR #5685, opened 2024-06-10
Prevent ambiguous 'localhost' worker connection

okurz:
Related progress issue: https://progress.opensuse.org/issues/161327

### DIFF
--- a/lib/OpenQA/Worker/WebUIConnection.pm
+++ b/lib/OpenQA/Worker/WebUIConnection.pm
@@ -83,6 +83,11 @@ sub register ($self) {
     my $ua = $self->ua;
     my $url = $self->url->clone;
     die 'client not correctly initialized before registration' unless ($webui_host && $working_dir && $ua && $url);
+    if ($worker->worker_hostname eq 'localhost' && !is_host_local($webui_host)) {
+        $self->_set_status(disabled =>
+              {error_message => "Rejecting registration as ambiguous 'localhost' to remote web UI host '$webui_host'"});
+        return undef;
+    }
 
     # finish any existing websocket connection
     $self->finish_websocket_connection;
--- a/t/24-worker-webui-connection.t
+++ b/t/24-worker-webui-connection.t
@@ -584,4 +584,15 @@ subtest 'destruction' => sub {
     is_deeply \@removed_timers, [42], 'timer removed on destruction';
 };
 
+subtest 'hostname checks' => sub {
+    $client->worker->worker_hostname('localhost');
+    $client->register;
+    is $client->status, 'disabled', 'ambiguous localhost on remote worker prevented';
+    like $happened_events[-1]->{error_message}, qr/Rejecting.*localhost/, 'clear message about ambiguous localhost';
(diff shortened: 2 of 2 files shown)


lib/OpenQA/Worker/WebUIConnection.pm
```diff
@@ -83,6 +83,8 @@ sub register ($self) {
     my $ua = $self->ua;
     my $url = $self->url->clone;
     die 'client not correctly initialized before registration' unless ($webui_host && $working_dir && $ua && $url);
+    die "Rejecting registration as ambiguous 'localhost' to remote webui_host '$webui_host'"
+      if $worker->worker_hostname eq 'localhost' && $webui_host !~ qr/localhost/;
```
Martchus, line 87: This breaks if one uses e.g. 127.0.0.1. We already have a utility to cover this somewhere so please use it here.

Martchus: Don't just die here; this will not lead to the desired error handling. The line above is ok because it acts as an internal assert (and if it would ever fire we have a programming mistake to fix). However, your line is supposed to be logged as a normal error message and the worker is supposed to keep running (as there might be other web UI connections that are not broken). So do the error handling like in the `!defined $worker_id` case below.

t/24-worker-webui-connection.t
```diff
@@ -584,4 +584,8 @@ subtest 'destruction' => sub {
     is_deeply \@removed_timers, [42], 'timer removed on destruction';
 };
 
+$client->worker->worker_hostname('localhost');
+like exception { $client->register }, qr/Rejecting.*localhost/, 'ambiguous localhost on remote worker prevented';
+
+
```
Martchus, line 590:
```suggestion
```

Martchus (changes requested):

lib/OpenQA/Worker/WebUIConnection.pm
```diff
@@ -83,6 +83,11 @@ sub register ($self) {
     my $ua = $self->ua;
     my $url = $self->url->clone;
     die 'client not correctly initialized before registration' unless ($webui_host && $working_dir && $ua && $url);
+    if ($worker->worker_hostname eq 'localhost' && !is_host_local($webui_host)) {
+        $self->_set_status(disabled =>
+              {error_message => "Rejecting registration as ambiguous 'localhost' to remote webui_host '$webui_host'"});
```
Martchus, line 88:
```suggestion
              {error_message => "Rejecting registration as ambiguous 'localhost' to remote web UI host '$webui_host'"});
```

Martchus (commented):
Looks good, I just have one minor nitpick.

lib/OpenQA/Worker/WebUIConnection.pm
okurz: ok, fine. I used `webui_host` because that's the name of the variable but granted it does not ready as proper english text anymore then.

Martchus (approved):

perlpunk (approved):

PR #5687, opened 2024-06-11
Enable automatic cloning of Git repos in openqa-bootstrap setups

Martchus:
This enables automatic cloning of Git repositories by the web UI when tests using Git sources are scheduled. We should not yet remove the automatic cloning of openSUSE tests and needles in `openqa-boostrap` itself because we still want to support cloning tests from non-Git sources for now. Additionally, the download in advance might still be desirable.

I plan to add an according check in openQA-in-openQA tests as part of the related ticket (https://progress.opensuse.org/issues/161771).

### DIFF
--- a/script/openqa-bootstrap
+++ b/script/openqa-bootstrap
@@ -108,7 +108,7 @@ if command -v $setup; then
 else
     curl -s https://raw.githubusercontent.com/os-autoinst/openQA/master/script/configure-web-proxy | bash -ex -s -- "$proxy_args"
 fi
-sed -i -e 's/#*.*method.*=.*$/method = Fake/' /etc/openqa/openqa.ini
+sed -i -e 's/#*.*method.*=.*$/method = Fake/' -e 's/#git_auto_clone =.*$/git_auto_clone = yes' /etc/openqa/openqa.ini
 
 if [[ -z $skip_suse_specifics ]] && ping -c1 download.suse.de. && (! rpm -q ca-certificates-suse); then
     # add internal CA if executed within suse network


script/openqa-bootstrap
```diff
@@ -108,7 +108,7 @@ if command -v $setup; then
 else
     curl -s https://raw.githubusercontent.com/os-autoinst/openQA/master/script/configure-web-proxy | bash -ex -s -- "$proxy_args"
 fi
-sed -i -e 's/#*.*method.*=.*$/method = Fake/' /etc/openqa/openqa.ini
+sed -i -e 's/#*.*method.*=.*$/method = Fake/' -e 's/#git_auto_clone =.*$/git_auto_clone = yes' /etc/openqa/openqa.ini
```
okurz, line 111: why is this not already enabled by default in openQA?

asdil12: Because it is new. And this way it was easier to merge the feature it without fearing to break anything right away.

asdil12 (approved):

script/openqa-bootstrap
Martchus: To avoid breaking existing production instances in case the feature misbehaves. The feature might also have security implications that could be relevant for instances used by the public. And because there's no cleanup yet.

(The last point makes enabling this actually generally questionable but I hope it is ok for new instances setup by the bootstrap script.)

okurz (approved):

script/openqa-bootstrap
okurz: ok. I created https://progress.opensuse.org/issues/162077 to follow up with that

ggardet:
openQA found this breaks on sed expression (missing final `/` ?) https://openqa.opensuse.org/tests/4269840#step/openqa_bootstrap/13

Martchus:
A C&P mistake. I created https://github.com/os-autoinst/openQA/pull/5694 to fix this.

PR #5688, opened 2024-06-11
Improve actions on test details page

Martchus:
See particular commit messages and https://progress.opensuse.org/issues/138593


kalikiana (approved):
Nice. I like the added explanations.

Also cool that you found a way to deal with the "empty" actions case.

okurz (approved):

PR #5691, opened 2024-06-11
feat: support configurable secrets hiding in UI and API

okurz:
Motivation:
Some test settings might contain sensitive information that doesn't follow
the default redaction pattern (^_SECRET_ or _PASSWORD). Users should be able
to specify custom patterns for redaction to prevent leaking secrets in logs,
test settings files, and the web interface.

Design Choices:
- Added _HIDE_SECRETS_REGEX job setting to allow custom redaction patterns.
- Optimized regex compilation in OpenQA::Log::redact_settings.
- Added redacted_settings_hash to OpenQA::Schema::Result::Jobs for centralized
  redaction.
- Updated to_hash in Jobs result class to support optional redaction.
- Updated Web UI templates and API controllers to use redacted settings.

Benefits:
- Improved security by preventing accidental leakage of custom secrets.
- Greater flexibility for users to protect sensitive data.
- Consistent redaction across logs, worker files, web UI, and API.

Related progress issue: https://progress.opensuse.org/issues/162086

### DIFF
--- a/lib/OpenQA/Log.pm
+++ b/lib/OpenQA/Log.pm
@@ -29,6 +29,7 @@ our @EXPORT_OK = qw(
   log_format_callback
   get_channel_handle
   setup_log
+  redact_settings
   redact_settings_in_file
   format_settings
 );
@@ -210,8 +211,22 @@ sub setup_log ($app, $logfile = undef, $logdir = undef, $level = undef) {
     }
 }
 
+# same approach as in os-autoinst bmwqemu.pm
 sub redact_settings ($vars) {
-    return {map { m/(^_SECRET_|_PASSWORD)/ ? ($_ => '[redacted]') : ($_ => $vars->{$_}) } keys %$vars};
+    my $hide_re = qr/^_SECRET_|_PASSWORD/;
+    my $error;
+
+    if (my $custom = $vars->{_HIDE_SECRETS_REGEX}) {
+        try { $hide_re = qr/$hide_re|$custom/ }
+        catch ($e) {
+            ($error = $e) =~ s/ at .*//s;
+            $error =~ s/\s+$//;
+        }
+    }
+
+    my $redacted = {map { $_ =~ $hide_re ? ($_ => '[redacted]') : ($_ => $vars->{$_}) } keys %$vars};
+    $redacted->{_HIDE_SECRETS_REGEX} = "(invalid regex specified: $error)" if $error;
+    return $redacted;
 }
 
 sub redact_settings_in_file ($file) {
--- a/lib/OpenQA/Schema/Result/Jobs.pm
+++ b/lib/OpenQA/Schema/Result/Jobs.pm
@@ -7,7 +7,7 @@ use Mojo::Base 'DBIx::Class::Core', -signatures;
 use Fcntl;
 use DateTime;
 use OpenQA::Constants qw(WORKER_COMMAND_ABORT WORKER_COMMAND_CANCEL);
-use OpenQA::Log qw(log_trace log_debug log_info log_warning log_error);
+use OpenQA::Log qw(log_trace log_debug log_info log_warning log_error redact_settings);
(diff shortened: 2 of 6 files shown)


Martchus (changes requested):

lib/OpenQA/Log.pm
```diff
@@ -208,8 +208,11 @@ sub setup_log ($app, $logfile = undef, $logdir = undef, $level = undef) {
     OpenQA::App->set_singleton($app);
 }
 
+# same approach as in os-autoinst bmwqemu.pm
 sub redact_settings ($vars) {
-    return {map { $_ !~ qr/(^_SECRET_|_PASSWORD)/ ? ($_ => $vars->{$_}) : ($_ => '[redacted]') } keys %$vars};
+    my $hide_re = '^_SECRET_|_PASSWORD';
+    $hide_re .= "|$vars->{_HIDE_MATCH_RE}" if $vars->{_HIDE_MATCH_RE};
```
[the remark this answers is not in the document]
okurz: Both suggestions covered including the fallback to ignore the additional custom stuff and still apply the default. I am still doing a concatenation to form the final regex. With more procedural code we could also use `if /$default_re/ || /$custom_re/` but IMHO it wouldn't be better. WDYT?

lib/OpenQA/Log.pm
```diff
@@ -208,8 +208,13 @@ sub setup_log ($app, $logfile = undef, $logdir = undef, $level = undef) {
...
 }
 
+# same approach as in os-autoinst bmwqemu.pm
 sub redact_settings ($vars) {
-    return {map { $_ !~ qr/(^_SECRET_|_PASSWORD)/ ? ($_ => $vars->{$_}) : ($_ => '[redacted]') } keys %$vars};
+    my $hide_re = '^_SECRET_|_PASSWORD';
+    my $custom_hide_re = eval { qr/$vars->{_HIDE_SECRETS_REGEX}/ } if $vars->{_HIDE_SECRETS_REGEX};
+    $vars->{_HIDE_SECRETS_REGEX} = "(invalid regex specified: $@)" if $@;
+    $hide_re .= "|$custom_hide_re" if $custom_hide_re;
+    return {map { /($hide_re)/ ? ($_ => '[redacted]') : ($_ => $vars->{$_}) } keys %$vars};
```
Martchus, line 217: It would probably make sense to compile `$hide_re` into a regex only once outside the `map` call:

```suggestion
    $hide_re = $custom_hide_re ? qr/($hide_re|$custom_hide_re)/ : qr/($hide_re)/;
    return {map { ($_ =~ $hide_re) ? ($_ => '[redacted]') : ($_ => $vars->{$_}) } keys %$vars};
```

Martchus (changes requested):

lib/OpenQA/Log.pm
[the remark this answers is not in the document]
okurz: done

okurz:
all suggestions incorporated and covered redacting in UI+API

lib/OpenQA/Schema/Result/Jobs.pm
```diff
@@ -432,6 +432,11 @@ sub settings_hash ($self, $prefetched = undef) {
     return $settings;
 }
 
+sub redacted_settings_hash ($self, $prefetched = undef) {
+    require OpenQA::Log;
```
Martchus: Why can't we just use `OpenQA::Log` normally here?

Martchus (changes requested):

lib/OpenQA/Schema/Result/Jobs.pm
okurz: yes, we can. Seems like an old left-over for localized testing. fixed

Martchus (approved):

perlpunk (approved):

PR #5701, opened 2024-06-14
Fix some UI issues after Bootstrap 5 migration

Martchus:
See particular commit messages and https://progress.opensuse.org/issues/159408

### DIFF
--- a/templates/webapi/admin/audit_log/index.html.ep
+++ b/templates/webapi/admin/audit_log/index.html.ep
@@ -55,7 +55,7 @@
         %>
     </span>
 
-    <table id="audit_log_table" class="table table-striped">
+    <table id="audit_log_table" class="table table-striped" style="width: 100%">
             <thead>
                 <tr>
                     <th>ID</th>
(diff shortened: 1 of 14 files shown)


perlpunk (approved):

templates/webapi/admin/audit_log/index.html.ep
```diff
@@ -55,7 +55,7 @@
         %>
     </span>
 
-    <table id="audit_log_table" class="table table-striped">
+    <table id="audit_log_table" class="table table-striped" style="width: 100%">
```
baierjan, line 58: Isn't that equivalent of something like this?
```suggestion
    <table id="audit_log_table" class="table table-striped w-100">
```

Martchus: This will not prevent DataTables from assigning a width via JavaScript as an element-specific style. It would actually work because this class uses `!important` but I think it is cleaner to prevent DataTables from setting the element-specific style in the first place. Besides, `w-100` is not really a readable class/style name in my opinion. `style="width: …`" is on the other hand self-explaining.

baierjan (approved):

PR #5706, opened 2024-06-17
Improve partial re-triggering of products

Martchus:
See particular commit messages and https://progress.opensuse.org/issues/138593

---

Note that I found more bugs when testing this locally with the production tests mentioned in the ticket. In particular, the dependencies where not correctly created. I'm looking into it but the changes proposed here are definitely already a step in the right direction.


b10n1k (approved):
LGTM

PR #5712, opened 2024-06-19
templates: Add group title names to title

pevik:
<title> tag should be always specific for various reasons (it's used as a default name when creating bookmark, it is a link description when link shared in a smartphone, used in os browser panel and tab name in a browser, maybe still SEO relevant, ...).

Therefore add group title names to the <title> (e.g. 'openSUSE Leap 15.6 Maintenance / openSUSE Leap 15.6 Backports') instead of generic 'openQA'.

To keep <title> reasonable long, don't include "Last Builds for ... group", but just a group name.

@Martchus @kalikiana @okurz FYI

Tested on my fork (restart web via: `systemctl restart openqa-webui`):
- http://quasar.suse.cz/group_overview/3
- http://quasar.suse.cz/group_overview/5
- http://quasar.suse.cz/parent_group_overview/2#grouped_by_build

### DIFF
--- a/templates/webapi/main/group_overview.html.ep
+++ b/templates/webapi/main/group_overview.html.ep
@@ -1,5 +1,5 @@
 % layout 'bootstrap';
-% title '';
+% title $group->{full_name};
 
 %= include 'layouts/info'
 
--- a/lib/OpenQA/WebAPI/Controller/Main.pm
+++ b/lib/OpenQA/WebAPI/Controller/Main.pm
@@ -140,12 +140,14 @@ sub _group_overview ($self, $resultset, $template) {
     my $group_hash = {
         id => $group->id,
         name => $group->name,
+        full_name => $group->name,
         is_parent => $is_parent_group,
         rendered_description => $group->rendered_description
     };
     if (!$is_parent_group && (my $parent = $group->parent)) {
         $group_hash->{parent_id} = $parent->id;
         $group_hash->{parent_name} = $parent->name;
+        $group_hash->{full_name} = $group->full_name;
     }
     $self->stash(
         group => $group_hash,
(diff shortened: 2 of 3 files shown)


kalikiana (approved):
I like it

templates/webapi/main/group_overview.html.ep
```diff
@@ -1,5 +1,5 @@
 % layout 'bootstrap';
-% title '';
+% title ((defined $group->{parent_name} ? $group->{parent_name}  . " / " : "") . $group->{name});
```
okurz, line 2: I wonder if we use this functionality already elsewhere and shouldn't duplicate but haven't found it yet. Maybe also there is a nicer way to write this without needing the logic?

Martchus, line 2:
```suggestion
% title $group->{full_name};
```
And populate that within the controller. (In `lib/OpenQA/WebAPI/Controller/Main.pm`, extend the `$group_hash` variable, e.g. in the subsequent if-block:

```
if (!$is_parent_group && (my $parent = $group->parent)) {
        $group_hash->{parent_id} = $parent->id;
        $group_hash->{parent_name} = $parent->name;
        $group_hash->{full_name} = $group->full_name;
    }
```
)

Martchus (changes requested):

templates/webapi/main/group_overview.html.ep
pevik: Maybe `parent_group_overview.html.ep` could now use `% title $group->{full_name};` as well

okurz (approved):

perlpunk (approved):

Martchus (approved):

PR #5714, opened 2024-06-19
Fix `openqa-cli` getting stuck on retrying

Martchus:
We need to create a new transaction after the previous transaction failed before retrying. Otherwise errors from the previous transaction will still be present even if the new transaction succeeds. Since the callers of `retry_tx` will still look at the initial transaction to get the result we need to assign it to the original transaction.

Related ticket: https://progress.opensuse.org/issues/160877


perlpunk (approved):

okurz:
s/it is no longer gets stuck/it no longer gets stuck/

Martchus:
The bad phrasing was just in the PR title.

okurz (approved):

PR #5722, opened 2024-06-25
Final fix v2 now for real

okurz:


asdil12:
Please separate the unrelated style changes into another git commit

PR #5725, opened 2024-06-26
apparmor: Allow 'sleep' as needed in os-autoinst scripts

okurz:
Related progress issue: https://progress.opensuse.org/issues/162752


baierjan (approved):
`r` might be unnecessary but probably will not hurt.

kalikiana (approved):

PR #5731, opened 2024-06-28
Be more explicit about the booleans in build_results

AdamWill:
There seems to be some ambiguity about exactly what these wind up being in the hash and then the JSON output generated from that hash; on Fedora, the tests fail because when the check result is false, the value seems to be showing up as 0 or '0', not ''.

To try and make this as explicit as possible, let's force the result of the check to be 0 or 1, and expect the JSON to contain the same integer. Although in actual fact, integer or string type doesn't seem to matter to the checks. If we force the result to be an integer and use strings in the checks - both json_is and is_deeply - they pass, and ditto if we force the result to be a string and use integers in the checks. Still, this seems to be the best we can do.


lib/OpenQA/BuildResults.pm
```diff
@@ -69,9 +69,9 @@ sub count_job ($job, $jr, $labels) {
 
 sub add_review_badge ($build_res) {
 
-    $build_res->{all_passed} = $build_res->{passed} + $build_res->{softfailed} >= $build_res->{total};
-    $build_res->{reviewed} = $build_res->{labeled} >= $build_res->{failed};
-    $build_res->{commented} = $build_res->{comments} >= $build_res->{failed};
+    $build_res->{all_passed} = $build_res->{passed} + $build_res->{softfailed} >= $build_res->{total} ? 1 : 0;
```
okurz, line 72: I think you want

```suggestion
    $build_res->{all_passed} = !!($build_res->{passed} + $build_res->{softfailed} >= $build_res->{total} ? 1 : 0);
```

And accordingly for the other lines.

okurz (changes requested):

okurz:
seems the issue you are trying to fix also just popped up in openSUSE Factory tests. Observed in https://build.opensuse.org/package/live_build_log/devel:openQA/openQA/openSUSE_Factory/x86_64

```
[  420s] t/api/10-jobgroups.t ...................................... 
[  420s] [INFO] using database schema "tmp_phs6WB_LYibHSqD2"
[  420s] [info] [pid:6545] setting database search path to tmp_phs6WB_LYibHSqD2 when registering Minion plugin
[  420s] # Subtest: build results
[  420s]     # Subtest: default
[  420s]         ok 1 - GET /api/v1/job_groups/1001/build_results?time_limit_days=99999
[  420s]         ok 2 - 200 OK
[  420s]         not ok 3 - exact match for JSON Pointer "/build_results/4/all_passed"
[  420s]         ok 4 - exact match for JSON Pointer "/build_results/4/build"

```

So good that you started your change.

lib/OpenQA/BuildResults.pm
AdamWill: I do? What does that do?

AdamWill: per https://stackoverflow.com/questions/2168406/what-does-double-exclamation-point-mean that seems to be redundant, but less good (it says of the `!!` version "Actually it returns an empty string which numifys to 0", which sounds like the thing we're trying to avoid here). `? 1 : 0` is already a ternary that attempts to ensure we get 1 if the check passes and 0 if it fails; that's the main point of this PR. I don't see what wrapping it in a `!!` does.

okurz: Sorry, I meant `!!($build_res->{passed} + $build_res->{softfailed} >= $build_res->{total})` as you wrote. The good thing is that we now clearly see the issue that you are trying to fix, e.g. https://build.opensuse.org/package/live_build_log/devel:openQA:GitHub:os-autoinst:openQA:PR-5735/openQA/openSUSE_Tumbleweed/x86_64
So I suggest we try with the numifying approach

perlpunk:
I guess the failure is caused by the latest release of Cpanel::JSON::XS, which we also see in our OBS build: perl-Cpanel-JSON-XS-4.380.0-1.1
https://metacpan.org/dist/Cpanel-JSON-XS/changes#L6
`- Encode real core booleans as boolean notation (PR #224 leonerd)`

```
% perl -wE'                           
use Cpanel::JSON::XS;
use Data::Dumper;
say Cpanel::JSON::XS->VERSION;
my $json = encode_json [1 > 2, !!0];
say $json;
my $data = decode_json $json;
say Dumper $data;'
4.37
["",""]
$VAR1 = [
          '',
          ''
        ];

% export PERL5LIB=$PWD/local/lib/perl5                      
% perl -wE'                           
use Cpanel::JSON::XS;
use Data::Dumper;
say Cpanel::JSON::XS->VERSION;
my $json = encode_json [1 > 2, !!0];
say $json;
my $data = decode_json $json;
say Dumper $data;'
4.38
[false,false]
$VAR1 = [
          bless( do{\(my $o = 0)}, 'JSON::PP::Boolean' ),
          $VAR1->[0]
        ];
```


Background: `!!` and comparisons like `$x > $y` both create a boolean context and will return a "boolean", a scalar with a special flag.
Since perl 5.36 the boolean "types" can now be tracked:
https://perldoc.perl.org/perl5360delta#Stable-boolean-tracking
The value didn't change, though, it's still an empty string for false and 1 for true.
But with the new Cpanel::JSON::XS release this will now be passed to the client as `false`.

Now, when decoding again, this gets loaded as `bless( do{\(my $o = 0)}, 'JSON::PP::Boolean' )`. Although I would expect it to load it as the new 5.36 booleans, the default seems to be to return blessed objects https://metacpan.org/pod/Cpanel::JSON::XS#$json-=-$json-%3Eunblessed_bool-([$enable]) , which IMHO is a bit inconsistent.

Compare YAML::PP which is consistent:
```
% perl -wE'
use YAML::PP;
use Data::Dumper;
my $yaml = YAML::PP::Dump [1 > 2, !!0];
say $yaml;
my $d = YAML::PP::Load($yaml);
say Dumper $d;'
---
- false
- false

$VAR1 = [
          !!0,
          !!0
        ];
```
So, with that said, explicitly numifying is the safest.

lib/OpenQA/BuildResults.pm
perlpunk: Yeah, as long as Cpanel::JSON::XS does return blessed objects as the default (or Mojolicious doesn't call it with different options), only numifying will help.

perlpunk:
I wonder when Mojo::JSON will encode the new booleans. @kraih, any plans to do this? I guess it would be good to be consistent with JSON::PP and Cpanel::JSON::XS.

perlpunk (approved):

kraih:
> I wonder when Mojo::JSON will encode the new booleans. @kraih, any plans to do this? I guess it would be good to be consistent with JSON::PP and Cpanel::JSON::XS.

Nobody has opened a PR yet.

AdamWill:
@perlpunk thanks a lot for tracking down the cause! I hate it when I don't quite know why something is happening. :)

AdamWill:
Even when the new behaviour is "correct" (encoding to JSON booleans) it'll be different between older and newer versions, right? So we'll either have to accept both possibilities, stick with numifying, or set a minimum version bound on the relevant components...

@okurz can you remove your requested change if we're now all in agreement that numifying is needed?

okurz (approved):
Well, I thought that also @perlpunk and @kraih would prefer the `!!` but the ternary is fine as well then

perlpunk:
> Well, I thought that also @perlpunk and @kraih would prefer the `!!` but the ternary is fine as well then

But that wouldn't work. `$x > $y` and `!!($x > $y)` create the same thing, as I explained above:
```
% perl -wE'
use Devel::Peek;
$x = 1 > 2;
$y = !!0;
Dump $x;
Dump $y;
say "($x, $y)"'
SV = PVNV(0x5575039b52a0) at 0x557503a00be0
  REFCNT = 1
  FLAGS = (IOK,NOK,POK,IsCOW,pIOK,pNOK,pPOK)
  IV = 0
  NV = 0
  PV = 0x557502804383 "" [BOOL PL_No]
  CUR = 0
  LEN = 0
SV = PVNV(0x5575039b52c0) at 0x5575039f20e0
  REFCNT = 1
  FLAGS = (IOK,NOK,POK,IsCOW,pIOK,pNOK,pPOK)
  IV = 0
  NV = 0
  PV = 0x557502804383 "" [BOOL PL_No]
  CUR = 0
  LEN = 0
(, )

```
It's 1 or the empty string.

PR #5734, opened 2024-06-29
Add link to monthly QA testing call

okurz:


okurz:
@foursixnine good like this?

Martchus (approved):

PR #5746, opened 2024-07-04
Fix URL rendering in job tests' comment

b10n1k:
Add a missing character to the BUGREF_REGEX which is used to render URL links. As such repos with an underscore in its name can be displayed as a link instead of a plain text in the comment section.

https://progress.opensuse.org/issues/163058

### DIFF
--- a/lib/OpenQA/Utils.pm
+++ b/lib/OpenQA/Utils.pm
@@ -76,7 +76,7 @@ BEGIN {
     $MARKER_URLS = join('|', keys %BUGURLS);
 
     # <marker>[#<project/repo>]#<id>
-    $BUGREF_REGEX = qr{(?<match>(?<marker>$MARKER_REFS)\#?(?<repo>[a-zA-Z/_-]+)?\#(?<id>([A-Z]+-)?\d+))};
+    $BUGREF_REGEX = qr{(?<match>(?<marker>$MARKER_REFS)\#?(?<repo>[^#\s<>,]+)?\#(?<id>([A-Z]+-)?\d+))};
 }
 
 use constant UNCONSTRAINED_BUGREF_REGEX => $BUGREF_REGEX;
--- a/lib/OpenQA/Utils.pm
+++ b/lib/OpenQA/Utils.pm
@@ -76,7 +76,7 @@ BEGIN {
     $MARKER_URLS = join('|', keys %BUGURLS);
 
     # <marker>[#<project/repo>]#<id>
-    $BUGREF_REGEX = qr{(?<match>(?<marker>$MARKER_REFS)\#?(?<repo>[a-zA-Z/-]+)?\#(?<id>([A-Z]+-)?\d+))};
+    $BUGREF_REGEX = qr{(?<match>(?<marker>$MARKER_REFS)\#?(?<repo>[a-zA-Z/_-]+)?\#(?<id>([A-Z]+-)?\d+))};
 }
 
 use constant UNCONSTRAINED_BUGREF_REGEX => $BUGREF_REGEX;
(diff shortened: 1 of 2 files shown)


kalikiana (approved):
Good catch! I assume the unit test should tell us if it works. Nice.

perlpunk (approved):

lib/OpenQA/Utils.pm
```diff
@@ -76,7 +76,7 @@ BEGIN {
     $MARKER_URLS = join('|', keys %BUGURLS);
 
     # <marker>[#<project/repo>]#<id>
-    $BUGREF_REGEX = qr{(?<match>(?<marker>$MARKER_REFS)\#?(?<repo>[a-zA-Z/-]+)?\#(?<id>([A-Z]+-)?\d+))};
+    $BUGREF_REGEX = qr{(?<match>(?<marker>$MARKER_REFS)\#?(?<repo>[a-zA-Z/_-]+)?\#(?<id>([A-Z]+-)?\d+))};
```
Martchus, line 79: Maybe it makes more sense to use a negative list of characters here to specify only the characters that should *not* match. Otherwise we'll have the same problem again if someone e.g. puts a `.` in the URL.

Martchus (changes requested):

lib/OpenQA/Utils.pm
b10n1k: As there are clear requirements of what should be in that exclusion I suggest to match everything for now, to move forward. What do you think?

Martchus: I guess the `#` needs to be in the exclusion.

t/16-utils.t
```diff
@@ -118,6 +118,9 @@ is href_to_bugref('https://progress.opensuse.org/issues/1234'), 'poo#1234';
 is bugref_to_href('boo#9876'), '<a href="https://bugzilla.opensuse.org/show_bug.cgi?id=9876">boo#9876</a>';
 is href_to_bugref('https://github.com/foo/bar/issues/1234'), 'gh#foo/bar#1234';
 is href_to_bugref('https://github.com/foo/bar_bar/issues/1234'), 'gh#foo/bar_bar#1234';
+is href_to_bugref('https://github.com/foo6/bar666/issues/1234'), 'gh#foo6/bar666#1234';
+# invalid URL but it is still generated
```
Martchus, line 122: The URL on the next line is valid. (It leads to a 404 page but the URL itself is valid.) So just remove that comment.

```suggestion
```

Martchus (changes requested):

lib/OpenQA/Utils.pm
Martchus: > Otherwise we'll have the same problem again if someone e.g. puts a `.` in the URL.

That was maybe a very bad example on my side - because on second though we probably want punctuation characters to act as terminators when parsing URLs. So maybe add the `.` to the range (which is now an exclude list).

b10n1k:
for some reason `t/ui/15-admin-workers.t` passes on my machine but fails here. checking

b10n1k:
> for some reason `t/ui/15-admin-workers.t` passes on my machine but fails here. checking

just a glitch in the matrix i guess. Now it passes

kalikiana:
> > for some reason `t/ui/15-admin-workers.t` passes on my machine but fails here. checking
> 
> just a glitch in the matrix i guess. Now it passes

[poo#163127](https://progress.opensuse.org/issues/163127)

Martchus (approved):

lib/OpenQA/Utils.pm
okurz: Please take a look into https://stackoverflow.com/a/59082561/22022448 . For security reasons @perlpunk and me would prefer a passlist but I guess we are ok with how it is now

PR #5749, opened 2024-07-08
Fix synchronization in fullstack test

Martchus:
Fix synchronization in fullstack test

The fullstack tests so far only waits until the worker has registered but
not until it has established a web socket connection. This means the job
assignment might fail if the web socket connection has not been established
quick enough by chance.

It is not easily possible to observe whether the web socket connection has
been established (one needed to parse the worker logs or create an API to
query the web socket server). Therefore this change simply re-tries the
scheduling similar to how it would work in production (where the scheduling
logic also runs in a loop). It will use the same re-try timeout/logic as
the preceding check for the worker registration.

Related ticket: https://progress.opensuse.org/issues/163127

---

I also did two other small changes in code I had to go though for this. They
should not affect the behavior. See the particular commit messages for
details.


kalikiana (approved):

perlpunk (approved):

Martchus:
I think the coverage is good (despite rendering issues in the HTML coverage report).

PR #5750, opened 2024-07-08
Use Codecov-related env variables from orga context

Martchus:
* Make configured token available by specifying context
* See https://progress.opensuse.org/issues/163187

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -317,6 +317,7 @@ workflows:
       - fullstack-unstable:
           <<: *requires_fullstack
       - codecov:
+          context: Codecov
           requires:
             - t
             - heavy


okurz (approved):

Martchus:
This unfortunately didn't work. As mentioned in https://progress.opensuse.org/issues/163187#note-8 I'm really wondering how OBS does it.

Martchus:
It seems like the report cannot be generated in the first place when running only a reduced number of tests. So I'll have to run the full test suite again to see whether it now works after adding security groups on CircleCI.

Martchus:
My tests revealed that the `CODECOV_TOKEN` environment variable is simply empty and that the Orb then passes an empty token leading to the the 401 auth error we see. This can be worked around by not using the token at all by also not using the Orb, see https://github.com/os-autoinst/openQA/pull/5751.

This means the culprit is the environment variable passing of CircleCI. I haven't found anything useful about it [on the forums](https://discuss.circleci.com/search?q=environment). So maybe I'll ask there myself considering we definitely followed the [documentation](https://circleci.com/docs/env-vars).

Martchus:
Looks like all I had to do was enabling the setting "Pass secrets to builds from forked pull requests", see https://github.com/os-autoinst/openQA/pull/5751. I tested that on this PR which showed the correct checksum.

We may still merge the initial change of this PR. I removed the additional commits again and reworded the first commit as "[Use Codecov-related env variables from orga context](https://github.com/os-autoinst/openQA/pull/5750/commits/6c665edd877c5e671b88f9b3886f969512f9f466)".

okurz (approved):

kalikiana (approved):

PR #5753, opened 2024-07-10
Avoid relying on project-level environment variables in CircleCI

Martchus:
We need to make some variables available on forks but some should better not be available on forks. The security settings of CircleCI contexts on organization-level allow that so it makes sense to avoid relying on project-level variables.

Related ticket: https://progress.opensuse.org/issues/163187


Martchus:
When this is merged I'll remove the variable from project-level. I guess the only way to find out whether nightly jobs are still able to run without the variable on project-level is to merge this PR.

Martchus:
I re-triggered a nightly run and it succeeded. However, it also had nothing to do so I guess we'll only see whether this worked on the run when there's an actual change.

PR #5754, opened 2024-07-10
Prevent live view viewers from making openQA unresponsive

Martchus:
* Avoid live view viewers from occupying too many Mojolicious workers by moving the image streaming of the live mode to the live handler daemon
* Make sure the frontend code can still connect to the streaming route in a development setup by taking the different port into account and by setting CORS headers
* Tested by running the main web UI service allowing only a limited number of connections via `script/openqa prefork -c 1 -w 1`. This causes it to be unresponsive if there is at least one live view tab open. With this change this is no longer a problem and multiple live tabs can be opened and are updated simultaneously.
* Tested without and with reverse proxy.

### DIFF
--- a/assets/javascripts/openqa.js
+++ b/assets/javascripts/openqa.js
@@ -121,16 +121,9 @@ function reloadPage() {
   location.reload();
 }
 
-// returns an absolute "ws://" URL for the specified URL which might be relative
-function makeWsUrlAbsolute(url, servicePortDelta) {
-  // don't adjust URLs which are already absolute
-  if (url.indexOf('ws:') === 0) {
-    return url;
-  }
-
-  // read port from the page's current URL
-  var location = window.location;
-  var port = Number.parseInt(location.port);
+function makeUrlPort(servicePortDelta) {
+  // read port from the location of the current page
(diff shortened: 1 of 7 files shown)


okurz (approved):

perlpunk:
```
[  310s] Can't locate object method "is_local_request" via package "OpenQA::Shared::Controller::Running" at /home/abuild/rpmbuild/BUILD/openQA-4.6.1720629109.ab1aa5c/lib/OpenQA/Shared/Controller/Running.pm line 186.
[  310s] # Tests were run but no plan was declared and done_testing() was not seen.
[  310s] # Looks like your test exited with 255 just after 1.
[  310s] t/26-controllerrunning.t .................................. 
```

perlpunk (approved):

perlpunk:
This might be related: https://progress.opensuse.org/issues/163931
`Can't locate object method "render_specific_not_found" via package "OpenQA::Shared::Controller::Running"`

Martchus:
Right, well have to move that helper to the shared helpers. I assume you'll take care of it considering you've already assigned yourself to the ticket.

PR #5757, opened 2024-07-11
[RFC] Create a hook placeholder

b10n1k:
`perl -I lib script/openqa-cli api --hook-scripts twdata --pretty --host http://127.0.0.1:9526 -X POST isos FOO=BARRR DISTRI=Awesome`

will get hook/twdata.json and will merge the variables with the ones from stdin. 
ofc location and dir name is under consideration. And also I thinkit is possible to extend it to run an actual script if the source is some program

### DIFF
--- a/lib/OpenQA/CLI.pm
+++ b/lib/OpenQA/CLI.pm
@@ -35,6 +35,9 @@ openqa-cli - provides command-line access to the openQA API
     # Show details for OSD job (prettified JSON)
     openqa-cli api --osd --pretty jobs/4160811
 
+    # Trigger a product using a hook script
+    openqa-cli api --o3 --hook-script foo -X POST isos
+
     # Archive job from O3
     openqa-cli archive --o3 408 /tmp/job_408
 
--- a/lib/OpenQA/CLI/api.pm
+++ b/lib/OpenQA/CLI/api.pm
@@ -7,13 +7,18 @@ use Mojo::Base 'OpenQA::Command', -signatures;
 use Mojo::File 'path';
 use Mojo::JSON qw(decode_json);
 use Mojo::Util qw(getopt);
+use File::stat;
+use File::Basename;
 use OpenQA::Constants qw(JOBS_OVERVIEW_SEARCH_CRITERIA);
+use OpenQA::Log qw(log_info log_debug log_warning log_error);
+use IPC::Run3 qw(run3);
 
 has description => 'Issue an arbitrary request to the API';
 has usage => sub {
     my $search_criteria = join(', ', JOBS_OVERVIEW_SEARCH_CRITERIA);
     shift->extract_usage =~ s/\$search_criteria/$search_criteria/r;
 };
+
 
 sub command ($self, @args) {
 
@@ -26,7 +31,8 @@ sub command ($self, @args) {
       'd|data=s' => \$data,
       'f|form' => \my $form,
       'j|json' => \my $json,
-      'H|hook-scripts=s' => \my $hooksrc,
+      'pre-script-hook=s' => \my $prehooksrc,
+      'post-script-hook=s' => \my $posthooksrc,
       'param-file=s' => \my @param_file,
       'r|retries=i' => \my $retries,
       'X|method=s' => \(my $method = 'GET');
@@ -36,32 +42,61 @@ sub command ($self, @args) {
 
     my $params;
     my @data;
-    if ($hooksrc) {
-	# if $hooksrc is a json  
-        my $hookdata = path("hook/$hooksrc.json")->slurp;
-	$data = ($hookdata);
-	my $hookparams = decode_json($data);
-	my $argsparams = $self->parse_params(\@args, \@param_file);
-	$params = {%$hookparams, %$argsparams};
-	# otherwise try to run a script
-	# here
-    } else {
-	$data = path($data_file)->slurp if $data_file;
-	@data = ($data);
-	$params = $form ? decode_json($data) : $self->parse_params(\@args, \@param_file);
-    }
-    @data = (form => $params) if keys %$params;
+    if ($prehooksrc) {
 
+	$self->_run_script ($_) for path($posthooksrc)->list->grep(qr/post_/)->each;
+	#my $hookdata = path($prehooksrc)->slurp;
+	#$data = ($hookdata);
+	#my $hookparams = decode_json($data);
+	#my $argsparams = $self->parse_params(\@args, \@param_file);
+	#$params = {%$hookparams, %$argsparams};
+    }
+    
+    $data = path($data_file)->slurp if $data_file;
+    @data = ($data);
+    $params = $form ? decode_json($data) : $self->parse_params(\@args, \@param_file);
+    use Data::Dumper;
+    @data = (form => $params) if keys %$params;
+    print Dumper(@data[1]);
+    if ($posthooksrc) {
+	use Data::Dumper;
+	# say for path('/home/sri/myapp')->list->each;
+	# run_script check exec permissions
+	$self->_run_script ($_, $data[1]) for path($posthooksrc)->list->grep(qr/post_/)->each;
+	say '_run_script returns ' . Dumper @data;
+    }
+    die "we got to the end";
     my $headers = $self->parse_headers(@headers);
     $headers->{Accept} //= 'application/json';
     $headers->{'Content-Type'} = 'application/json' if $json;
-
     my $url = $self->url_for($path);
     my $client = $self->client($url);
     my $tx = $client->build_tx($method, $url, $headers, @data);
     $self->retry_tx($client, $tx, $retries);
 }
 
+
+sub _run_script ($self, $hook_path, $data) {
+    my $hook_fl = basename($hook_path);
+    return unless ($hook_fl =~ /^(pre|post)_/);
+    my $info    = stat($hook_path) or die "path doesnt exist $hook_path : $!";
+    unless (-x $info) {
+	log_info $_ . ' is not executable';
+	return;
+    }
+    log_info $_ . ' is called';
+    eval {
+	print 'TST ';
+	# https://stackoverflow.com/a/54500161/1462096
+	#my @out = qx($hook_path);
+	#print '@out ';
+	my ($out, $err);
+	my @out = run3([$hook_path, $data], \$out, \$err);
+	print 'This is out ' . Dumper $out;
+	print 'This is err ' . Dumper $err;
+    };
+    log_error "$@" if $@;
+}
 1;
 
 =encoding utf8
--- a/lib/OpenQA/CLI/api.pm
+++ b/lib/OpenQA/CLI/api.pm
@@ -26,6 +26,7 @@ sub command ($self, @args) {
       'd|data=s' => \$data,
       'f|form' => \my $form,
       'j|json' => \my $json,
+      'H|hook-scripts=s' => \my $hooksrc,
       'param-file=s' => \my @param_file,
       'r|retries=i' => \my $retries,
       'X|method=s' => \(my $method = 'GET');
@@ -33,9 +34,22 @@ sub command ($self, @args) {
     @args = $self->decode_args(@args);
     die $self->usage unless my $path = shift @args;
 
-    $data = path($data_file)->slurp if $data_file;
-    my @data = ($data);
-    my $params = $form ? decode_json($data) : $self->parse_params(\@args, \@param_file);
+    my $params;
+    my @data;
+    if ($hooksrc) {
+	# if $hooksrc is a json  
+        my $hookdata = path("hook/$hooksrc.json")->slurp;
+	$data = ($hookdata);
+	my $hookparams = decode_json($data);
+	my $argsparams = $self->parse_params(\@args, \@param_file);
+	$params = {%$hookparams, %$argsparams};
+	# otherwise try to run a script
+	# here
+    } else {
+	$data = path($data_file)->slurp if $data_file;
+	@data = ($data);
+	$params = $form ? decode_json($data) : $self->parse_params(\@args, \@param_file);
+    }
     @data = (form => $params) if keys %$params;
 
     my $headers = $self->parse_headers(@headers);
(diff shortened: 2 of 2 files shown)


lib/OpenQA/CLI/api.pm
```diff
@@ -26,16 +26,30 @@ sub command ($self, @args) {
       'd|data=s' => \$data,
       'f|form' => \my $form,
       'j|json' => \my $json,
+      'H|hook-scripts=s' => \my $hooksrc,
```
Martchus, line 29: Here it is `--hook-scripts` and in the example before just `--hook-script`. I guess it should be the latter because being able to invoke one script must be enough.

lib/OpenQA/CLI/api.pm
```diff
@@ -26,16 +26,30 @@ sub command ($self, @args) {
...
 
     @args = $self->decode_args(@args);
     die $self->usage unless my $path = shift @args;
 
-    $data = path($data_file)->slurp if $data_file;
-    my @data = ($data);
-    my $params = $form ? decode_json($data) : $self->parse_params(\@args, \@param_file);
+    my $params;
+    my @data;
+    if ($hooksrc) {
```
Martchus, line 39: I don't think the presence of a hook script should cause the tool to completely ignore the presence of a data file. If both options are not combinable (but probably they should be combinable) this should lead to a clear error/warning.

lib/OpenQA/CLI/api.pm
```diff
@@ -26,16 +26,30 @@ sub command ($self, @args) {
...
+	# if $hooksrc is a json  
+        my $hookdata = path("hook/$hooksrc.json")->slurp;
```
Martchus, line 41: This is very unintuitive. Let's keep it simple and not "surround" the specified path with `hook/`…`.json`.

Just invoke the specified hook script reading its stdout. Abort with a clear error if it exited with a non-zero return code. Otherwise, parse the output as JSON. There's *no* reason to involve an intermediate `.json` file that must be placed according to some conventions.

Note that you can use the quote like operator `qx(…)` in Perl to invoke a script and get its output. When using it, have a close look at [the documentation](https://perldoc.perl.org/perlop#Quote-Like-Operators) because error handling is a bit involved (but should be done correctly to get a clear error message in any case).

lib/OpenQA/CLI/api.pm
```diff
@@ -26,16 +26,30 @@ sub command ($self, @args) {
...
+	$data = ($hookdata);
```
Martchus, line 42: Please avoid needless copies. The use of parenthesis here makes no sense as well.

lib/OpenQA/CLI/api.pm
```diff
@@ -114,18 +128,20 @@ sub command ($self, @args) {
...
         --apisecret <secret>      API secret
     -a, --header <name:value>     One or more additional HTTP headers
-    -D, --data-file <path>        Load content to send with request from file
+    -D, --data-file <path>        Load content to send with request from file.
+                                  Works with when -f is provided
     -d, --data <string>           Content to send with request, alternatively
                                   you can also pipe data to openqa-cli
     -f, --form                    Turn JSON object into form parameters
         --host <host>             Target host, defaults to http://localhost
+    -H, --hook-scripts            Use a hook script!! TDB
```
Martchus, line 144: I guess you mean TBD but more importantly the parameter name should most likely be:

```suggestion
    -H, --hook-script             Use a hook script!! TBD
```

Martchus (changes requested):
The commit message should have details (similar to what you have in the PR description) and a ticket reference. But don't just re-use the PR description. It is full of casing and punctuation issues and too informal (avoid abbreviations like "ofc"). Please generally put a bit more though into your wording. The commit message subject should also state the why (and not the what). **I must say it would be nice if we would not have to repeat how to write commit messages that often.**

lib/OpenQA/CLI/api.pm
```diff
@@ -36,32 +42,61 @@ sub command ($self, @args) {
...
+    my $info    = stat($hook_path) or die "path doesnt exist $hook_path : $!";
+    unless (-x $info) {
+	log_info $_ . ' is not executable';
+	return;
+    }
+    log_info $_ . ' is called';
+    eval {
+	print 'TST ';
+	# https://stackoverflow.com/a/54500161/1462096
+	#my @out = qx($hook_path);
```
Martchus, line 91: I'm not a fan of `IPC::Run`. You may use it but you can still also just keep using Perl built-ins, e.g. like I did in one of our scripts: https://github.com/os-autoinst/scripts/blob/b4f88b41c51ef5571e3f96764a99acd43252243c/openqa-clone-and-monitor-job-from-pr#L154

Martchus (changes requested):
The code is too chaotic at this point to do a real review.

lib/OpenQA/CLI/api.pm
perlpunk: That's IPC::Run3 though (I guss you meant to comment 3 lines below?)

lib/OpenQA/CLI/api.pm
```diff
@@ -26,28 +31,72 @@ sub command ($self, @args) {
...
+	#print '@out ';
+	my ($out, $err);
+	my @out = run3([$hook_path, $data], \$out, \$err);
```
perlpunk, line 94:
```suggestion
	my @out = run3([$hook_path, $data], \my $out, \my $err);
```

okurz (changes requested):
ok, good start. How would discovering and reading hook scripts from an outside place like https://github.com/os-autoinst/openqa-trigger-from-obs look like and what would the hook script accomplish what is not already possible?

okurz:
> [...] So this is just for convenience - so that we don't have to do bigger changes to the architecture of https://github.com/os-autoinst/openqa-trigger-from-obs.

Sorry, I don't see the point in this. In the end we expect that some power users are able to tweak the test schedule based on logic that can be maintained without needing direct changes to os-autoinst nor openQA. I don't yet see how this openqa-cli extension enables that

lib/OpenQA/CLI/api.pm
```diff
@@ -26,7 +31,8 @@ sub command ($self, @args) {
       'd|data=s' => \$data,
       'f|form' => \my $form,
       'j|json' => \my $json,
-      'H|hook-scripts=s' => \my $hooksrc,
+      'pre-script-hook=s' => \my $prehooksrc,
+      'post-script-hook=s' => \my $posthooksrc,
```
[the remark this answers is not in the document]
foursixnine: See motivation part of https://progress.opensuse.org/issues/161168

lib/OpenQA/CLI/api.pm
foursixnine: This parameter is not needed

```
      'pre-script-hook=s' => \my $prehooksrc,
      'post-script-hook=s' => \my $posthooksrc,
```

during the session we also spoke about not having short options for these.

lib/OpenQA/CLI/api.pm
```diff
@@ -114,18 +163,20 @@ sub command ($self, @args) {
...
+    openqa-cli api -f --data-file form.json -X POST job_templates_scheduling/1
 
   Options:
         --apibase <path>          API base, defaults to /api/v1
         --apikey <key>            API key
         --apisecret <secret>      API secret
     -a, --header <name:value>     One or more additional HTTP headers
-    -D, --data-file <path>        Load content to send with request from file
+    -D, --data-file <path>        Load content to send with request from file.
+                                  Works with when -f is provided
```
foursixnine, line 174: `-f` is needed for `-D` to work? if so, make it clear (i guess is missing doc?)
```suggestion
                                  Works with when -f is provided
```

lib/OpenQA/CLI/api.pm
```diff
@@ -26,28 +31,72 @@ sub command ($self, @args) {
...
+    $params = $form ? decode_json($data) : $self->parse_params(\@args, \@param_file);
+    use Data::Dumper;
     @data = (form => $params) if keys %$params;
-
+    print Dumper(@data[1]);
+    if ($posthooksrc) {
+	use Data::Dumper;
+	# say for path('/home/sri/myapp')->list->each;
+	# run_script check exec permissions
+	$self->_run_script ($_, $data[1]) for path($posthooksrc)->list->grep(qr/post_/)->each;
```
foursixnine, line 65: something important here, is to guarantee execution order, and document it 

```
for i in 1 9 10 19 29 21 01 09 20 100 200 101 110 109 199;do touch {pre_$i,post_$i}; done;
﻿foursixnine@pakhet test-sort % perl -Mojo -E 'say $_ for f->path->list->each'
/Users/foursixnine/tmp/test-sort/post_01
/Users/foursixnine/tmp/test-sort/post_09
/Users/foursixnine/tmp/test-sort/post_1
/Users/foursixnine/tmp/test-sort/post_10
/Users/foursixnine/tmp/test-sort/post_100
/Users/foursixnine/tmp/test-sort/post_101
/Users/foursixnine/tmp/test-sort/post_109
/Users/foursixnine/tmp/test-sort/post_110
/Users/foursixnine/tmp/test-sort/post_19
/Users/foursixnine/tmp/test-sort/post_199
/Users/foursixnine/tmp/test-sort/post_20
/Users/foursixnine/tmp/test-sort/post_200
/Users/foursixnine/tmp/test-sort/post_21
/Users/foursixnine/tmp/test-sort/post_29
/Users/foursixnine/tmp/test-sort/post_9
/Users/foursixnine/tmp/test-sort/pre_01
/Users/foursixnine/tmp/test-sort/pre_09
/Users/foursixnine/tmp/test-sort/pre_1
/Users/foursixnine/tmp/test-sort/pre_10
/Users/foursixnine/tmp/test-sort/pre_100
/Users/foursixnine/tmp/test-sort/pre_101
/Users/foursixnine/tmp/test-sort/pre_109
/Users/foursixnine/tmp/test-sort/pre_110
/Users/foursixnine/tmp/test-sort/pre_19
/Users/foursixnine/tmp/test-sort/pre_199
/Users/foursixnine/tmp/test-sort/pre_20
/Users/foursixnine/tmp/test-sort/pre_200
/Users/foursixnine/tmp/test-sort/pre_21
/Users/foursixnine/tmp/test-sort/pre_29
/Users/foursixnine/tmp/test-sort/pre_9
```

lib/OpenQA/CLI/api.pm
```diff
@@ -26,28 +31,72 @@ sub command ($self, @args) {
...
       'param-file=s' => \my @param_file,
       'r|retries=i' => \my $retries,
       'X|method=s' => \(my $method = 'GET');
 
     @args = $self->decode_args(@args);
     die $self->usage unless my $path = shift @args;
 
+    my $params;
+    my @data;
+    if ($prehooksrc) {
```
foursixnine, line 45:
```suggestion
    if ($pre_hooks_dir) {
```

lib/OpenQA/CLI/api.pm
```diff
@@ -26,28 +31,72 @@ sub command ($self, @args) {
       'd|data=s' => \$data,
       'f|form' => \my $form,
       'j|json' => \my $json,
+      'pre-script-hook=s' => \my $prehooksrc,
+      'post-script-hook=s' => \my $posthooksrc,
```
foursixnine, line 35:
```suggestion
      'pre-script-hook=s' => \my $pre_hooks_dir,
      'post-script-hook=s' => \my $post_hooks_dir,
```

lib/OpenQA/CLI.pm
```diff
@@ -35,6 +35,9 @@ openqa-cli - provides command-line access to the openQA API
     # Show details for OSD job (prettified JSON)
     openqa-cli api --osd --pretty jobs/4160811
 
+    # Trigger a product using a hook script
+    openqa-cli api --o3 --hook-script foo -X POST isos
```
foursixnine, line 39: needs to be updated too

lib/OpenQA/CLI/api.pm
foursixnine: Why not pick one implementation and run with it? (also to remove the dependency on `Mojo::IOLoop::ReadWriteProcess` perhaps moving that `_run_cmd_capturing_output` (and similar ones) can be moved to a single place in some other PR. 

Looks like the same thing is being implemented over and over...

lib/OpenQA/CLI/api.pm
Martchus: I'm also wondering whether it makes sense to run multiple scripts. I guess I would not even have separate pre/post scripts. Or can you explain what the use case of those will be?

foursixnine: > Or can you explain what the use case of those will be?

pre_hook: Add very_expensive_testsuite if repodata contains $_SPECIAL_STUFF
post_hook: turn on very_expensive_machine if repodata contains $_SPECIAL_STUFF

one instance of such actions _could be_ spinning up a Grace Hopper machine, or making an api call to a third service

b10n1k: I would expect that the scripts are atomic and independent. I can think of some implications. What happens if one script overwrites previous variables for example.

perlpunk: regarding the order: `list()` does a sort, so the order is guaranteed. it's alphanumeric, so if you want to order by number it probably makes sense to have the same number of digits, e.g. `post_001`, `post_002` etc.

Martchus:
> Sorry, I don't see the point in this. In the end we expect that some power users are able to tweak the test schedule based on logic that can be maintained without needing direct changes to os-autoinst nor openQA. I don't yet see how this openqa-cli extension enables that

The logic would not live in os-autoinst or openQA and no direct changes to those repos would be needed. Only this change to generally allow running a hook script would be required. The point was to make things easier on the openqa-trigger-from-obs side of things - especially because we as tools team were asked to help here.

However, with the pre/post-distinction and otherwise too convoluted logic I also don't see this PR being mergeable. The initial idea was to call a single, explicitly specified script. This PR is nothing like that, though. So I tend to close it and do whatever is needed in openqa-trigger-from-obs after all. At least https://github.com/os-autoinst/openqa-trigger-from-obs/pull/255 looks more promising than this change - and also simpler again.

PR #5758, opened 2024-07-11
Support parameters `arch` and `machine` in job overview API

Martchus:
The job overview API already supports most filter parameters the overview
web page supports. This change adds support for the two remaining filter
parameters. (The `t` parameter is still not supported.)

### DIFF
--- a/t/api/04-jobs.t
+++ b/t/api/04-jobs.t
@@ -222,47 +222,38 @@ subtest 'validation of IDs' => sub {
 };
 
 subtest 'job overview' => sub {
+    # define expected jobs
+    my %job_99940 = (id => 99940, name => 'opensuse-Factory-DVD-x86_64-Build0048@0815-doc@64bit');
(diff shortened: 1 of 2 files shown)


t/api/04-jobs.t
```diff
@@ -222,47 +222,38 @@ subtest 'validation of IDs' => sub {
...
-        ],
-        'latest build present'
-    );
+    subtest 'specific build without additional parameters' => sub {
+        $query->query(build => '0048');
+        $t->get_ok($query->path_query)->status_is(200);
+        $t->json_is('' => [\%job_99939, \%job_99938, \%job_99936], 'latest build present');
+    };
+
+    subtest 'additional parameters' => sub {
```
b10n1k, line 246: Better like `query with specific build and additional parameters`?

Martchus: Not all queries within this sub test have a specific build - and it shouldn't really matter. (I mainly specified the build in some of these queries to generally limit the set of jobs but not because I think it is worthwhile to test the `build` parameter together with these other parameters.)

okurz (approved):

PR #5761, opened 2024-07-12
Prevent failure when restarting Minion-related services

Martchus:
Not all services we are trying to restart might be installed, e.g. in production setups workers are on separate hosts and therefore no cache service units are present. This change treats the restart still as success if not all units can be found.

Related ticket: https://progress.opensuse.org/issues/163013


Martchus:
I tested this by modifying the unit locally to include units that don't exist. I checked whether the units that do exist are still restarted when some don't and they are. I also tested the case when all units exists and it works (so exit code 0 doesn't need to be mentioned explicitly).

b10n1k (approved):

perlpunk (approved):

PR #5768, opened 2024-07-16
Retry compose workflow to accomodate zypper errors

kalikiana:
See: https://progress.opensuse.org/issues/162848

### DIFF
--- a/.github/workflows/compose.yml
+++ b/.github/workflows/compose.yml
@@ -5,8 +5,12 @@ on: [push, pull_request]
 jobs:
   webui-docker-compose:
     runs-on: ubuntu-latest
-    timeout-minutes: 60
+    timeout-minutes: 650
     steps:
       - uses: actions/checkout@v4
-      - name: Verify that containers can be composed
-        run: make test-containers-compose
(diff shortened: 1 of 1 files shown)


.github/workflows/compose.yml
```diff
@@ -5,8 +5,12 @@ on: [push, pull_request]
 jobs:
   webui-docker-compose:
     runs-on: ubuntu-latest
-    timeout-minutes: 60
+    timeout-minutes: 650
     steps:
       - uses: actions/checkout@v4
-      - name: Verify that containers can be composed
-        run: make test-containers-compose
+      - uses: nick-fields/retry@v3
```
okurz, line 11: Why not a retry within the code where it applies? Consider to explain that in your commit rather than (just) here

okurz (changes requested):

.github/workflows/compose.yml
kalikiana: #5772 is an alternative retrying from within the script

okurz: #5772 looks better. merged meanwhile. Close here?

PR #5770, opened 2024-07-16
Add test for render_specific_not_found

perlpunk:
Issue: https://progress.opensuse.org/issues/163931

It now returns a plain text, because we don't have all helpers and variables
for a fully fledged openQA page in the livehandler.
That's probably ok for such pages that you normally don't link to
directly.

### DIFF
--- a/lib/OpenQA/Shared/Plugin/SharedHelpers.pm
+++ b/lib/OpenQA/Shared/Plugin/SharedHelpers.pm
@@ -75,8 +75,8 @@ sub _is_local_request ($c) {
 }
 
 sub _render_specific_not_found ($c, $title, $error_message) {
-    $c->stash(title => $title, error_message => $error_message);
-    return $c->render(template => 'main/specific_not_found', status => 404);
+    $c->res->headers->content_type('text/plain');
+    return $c->render(status => 404, text => "$title - $error_message");
 }
 
 1;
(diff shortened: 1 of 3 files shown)


perlpunk:
I thought that maybe it couldn't find the template, so I added 
```
$self->renderer->paths->[0] = path($self->renderer->paths->[0])->child('webapi')->to_string;
```
to the live handler app, but no success.

So currently I'm wondering if this actually works in production, if the LiveHandler doesn't have the path to the templates.

Martchus:
I guess it doesn't work in production: https://openqa.opensuse.org/liveviewhandler/tests/4345159/streaming

I also tested this locally: http://localhost:9528/liveviewhandler/tests/1234/streaming
Locally I'm getting the same error you mentioned in the PR description.

perlpunk:
Like discussed I made the specific template a plaintext page.
I guess it still has to be mocked.
It would be nice if the error message from Mojolicious would tell us more details, though. In the log I see `[trace] [mIY3-s_V52Eq] Template "main/specific_not_found.html.ep" not found`. Why is that a trace message only?

perlpunk:
Before I can continue to think about what to do I first need to know how I get to a `/tests/id/edit` page in the webui.
t/ui/12-needle-edit.t is failing, and I don't see the actual problem when running in non headless mode.
I don't see any "edit test" link. So is this page supposed to be seen by a human or not?

Martchus:
When I remember correctly, this route can be accessed via the developer mode when the test is paused on a needle mismatch.

perlpunk:
The function now returns a plain text (although that's probably not strictly what you are supposed to return for text/html.
I should probably do `$c->res->headers->content_type('text/plain');` as well?

kalikiana (approved):

okurz (approved):

PR #5774, opened 2024-07-17
Use nginx zone size large enough for 16k pages

kalikiana:
Minimum zone size is page size * 8.

See: https://progress.opensuse.org/issues/162611

### DIFF
--- a/docs/Installing.asciidoc
+++ b/docs/Installing.asciidoc
@@ -425,6 +425,10 @@ when accessing the server via its IP address. This means that the default vhost
 for `localhost` in *nginx.conf* will take precedence when accessing the server
 via `localhost`. You might want to disable it.
 
+If you use the openqa-upstreams.inc which is included with the upstream
+sources and openQA packages, you may want to customize the size of the shared
+memory segment according to the formula: page_size * 8
+
 For openQA you need to set `httpsonly = 0` as described in the TLS/SSL section
 below, if you do not setup NGINX for SSL.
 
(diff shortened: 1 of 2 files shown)


nicksinger:
very good catch! Glad you finally found a reasonable explanation for that. Should we extend openqa setup documentation as well? (not particularly in this PR)

okurz (approved):

docs/Installing.asciidoc
```diff
@@ -425,6 +425,10 @@ when accessing the server via its IP address. This means that the default vhost
 for `localhost` in *nginx.conf* will take precedence when accessing the server
 via `localhost`. You might want to disable it.
 
+If you use the openqa-upstreams.inc which is included with the upstream
+sources and openQA packages you may want to customize the use of **zones**
+and use different zones or change the size of the shared memory.
```
nicksinger, line 430:
```suggestion
sources and openQA packages, you may want to customize the size of the shared memory segment according to the formula: page_size * 8
```

I'm not sure about the comma between packages and you - it feels right but I might be completely wrong :) Using different zones might be an advanced config but not necessary to get openQA running (in contrast to the zone size). I would leave it out for now and only add it once we actually make use of that zone (e.g. for rate limiting)

kalikiana: Well, if you don't use zones it's totally irrelevant. so there's nothing to document then.

kalikiana: And by **it** I mean documenting the size as per your suggestion

nicksinger: What do you mean with "If you use no zones"? We include it in our upstream and therefore it will get used - you encountered this first hand.
I mainly wanted to clear confusion that the user/admin is not supposed to change `upstream_webui` but might be *required* to update the size.

But anyway, I'm not rejecting your changes because it honestly is just a minor detail and nitpick from me :)

nicksinger (approved):

perlpunk (approved):

PR #5781, opened 2024-07-20
Avoid warning about clashing imported subroutine

perlpunk:
Test::Most exports everything from several other modules, including Test::Deep, which exports 'any'.
This lead to:

    Prototype mismatch: sub main::any: none vs (&@)

I saw the warning in the reported issue
https://progress.opensuse.org/issues/164239

### DIFF
--- a/t/full-stack.t
+++ b/t/full-stack.t
@@ -19,12 +19,12 @@ BEGIN {
     $ENV{MOJO_INACTIVITY_TIMEOUT} = 10 * 60;
 }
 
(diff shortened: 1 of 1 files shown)


okurz (approved):
Nice! I was working on the same thing. See https://github.com/os-autoinst/openQA/commit/f28fcd7206e1a7ab722f7fd1ac6c88136779ffcc . My commit is actually 3h older than yours but you were the first to open a PR ;)

PR #5789, opened 2024-07-24
CI: Cover repo errors with exponential retries

okurz:
Related progress issue: https://progress.opensuse.org/issues/162848

### DIFF
--- a/tools/retry
+++ b/tools/retry
@@ -5,6 +5,8 @@ RETRY="${RETRY:-3}"
 STABILITY_TEST="${STABILITY_TEST:-0}"
 # Pass a hook command which is executed if a retry occurs
 HOOK="${HOOK:-}"
+EXPONENTIAL=${EXPONENTIAL:-1}
+SLEEP="${SLEEP:-0}"
 
 run_once() {
     "$@"
@@ -29,5 +31,8 @@ else
             RETRY_ATTEMPT=$n "$HOOK"
         fi
         n=$((n + 1))
+        echo "Retrying after sleeping ${SLEEP}s …" >&2
+        sleep "$SLEEP"
+        [ -n "$EXPONENTIAL" ] && SLEEP=$((SLEEP * EXPONENTIAL))
     done
 fi
--- a/tools/test_containers_compose
+++ b/tools/test_containers_compose
@@ -4,6 +4,13 @@ set -euo pipefail
 
 thisdir="$(cd "$(dirname "${BASH_SOURCE[0]}")" > /dev/null 2>&1 && pwd)"
 
+# longer, exponential sleep between retries to retry on infrastructure issues
+EXPONENTIAL="${EXPONENTIAL:-5}"
+SLEEP="${SLEEP:-30}"
+
+export EXPONENTIAL
+export SLEEP
+
 wait_until() {
     command=$1
     timeout=$2


perlpunk:
https://app.circleci.com/pipelines/github/os-autoinst/openQA/14088/workflows/38f48aa9-d43f-4d94-aa34-125ce4857a7a/jobs/132716
```

In tools/test_containers_compose line 5:
thisdir="$(cd "$(dirname "${BASH_SOURCE[0]}")" > /dev/null 2>&1 && pwd)"
^-----^ SC2034 (warning): thisdir appears unused. Verify use (or export if used externally).

For more information:
  https://www.shellcheck.net/wiki/SC2034 -- thisdir appears unused. Verify us...
make: *** [Makefile:302: test-shellcheck] Error 1
```

perlpunk (approved):

tools/test_containers_compose
```diff
@@ -20,9 +18,16 @@ wait_until() {
 }
 
 setup_containers() {
-    "$thisdir"/retry sudo docker compose build
+    if command -v retry > /dev/null; then
+        retry="retry -e -s 30 -r 7"
+    else
+        curl -sS https://raw.githubusercontent.com/okurz/retry/main/retry > retry
```
kalikiana, line 24: I would prefer we didn't download scripts into the local checkout of the repository

kalikiana (changes requested):

tools/test_containers_compose
okurz: > I would prefer we didn't download scripts into the local checkout of the repository

what should we do instead?

kalikiana: To be honest using "retry" instead of "retry" here also raises more questions. We also rely on this script for zypper issues in `build-docs` and `build_cache`.
So the download ideally happens either in a wrapper script or as part of Makefile.

We only have a "build" folder for docs I guess.

okurz: > To be honest using "retry" instead of "retry" here also raises more questions.

Do you mean it's confusing if we don't rename the application because it happens to be called the same? Or do you confuse github.com/okurz/retry with tools/retry which are two different things?

> We also rely on this script for zypper issues in `build-docs` and `build_cache`. So the download ideally happens either in a wrapper script or as part of Makefile.

We rely on tools/retry so far and not github.com/okurz/retry . Maybe you want to suggest that we should use github.com/okurz/retry everywhere. That's actually what I would plan but didn't want to try everything in one go. So what do you suggest how should we proceed?

kalikiana: > Do you mean it's confusing if we don't rename the application because it happens to be called the same? Or do you confuse github.com/okurz/retry with tools/retry which are two different things?

You are replacing the second with the first here. And we use tools/retry for the same use case so far. So if we use it here, we should probably use it everywhere?

okurz: I mean in theory we could do that but it's not that easy as we need to ensure we have the application either downloaded or available from packages which is not the case on all OSs. And also tools/retry supports a stability test. Ok, I re-implemented the exponential sleep in a corresponding fashion now in tools/retry. Better?

tools/retry
```diff
@@ -29,5 +31,8 @@ else
             RETRY_ATTEMPT=$n "$HOOK"
         fi
         n=$((n + 1))
+    echo "Retrying after sleeping ${SLEEP}s …" >&2
+    sleep "$SLEEP"
+    [ -n "$EXPONENTIAL" ] && SLEEP=$((SLEEP*EXPONENTIAL))
```
kalikiana, line 36:
```
-    echo "Retrying after sleeping ${SLEEP}s …" >&2
-    sleep "$SLEEP"
-    [ -n "$EXPONENTIAL" ] && SLEEP=$((SLEEP*EXPONENTIAL))
+        echo "Retrying after sleeping ${SLEEP}s …" >&2
+        sleep "$SLEEP"
+        [ -n "$EXPONENTIAL" ] && SLEEP=$((SLEEP * EXPONENTIAL))
```

I guess the indentation is off.

kalikiana, line 36:
```suggestion
       echo "Retrying after sleeping ${SLEEP}s …" >&2
       sleep "$SLEEP"
       [ -n "$EXPONENTIAL" ] && SLEEP=$((SLEEP * EXPONENTIAL))
```

okurz: done

Martchus (approved):

kalikiana (approved):

PR #5796, opened 2024-07-29
Add development service convenience runner "openqa-run"

okurz:


d3flex (changes requested):
thanks first of all, but just no. IMO this is incomplete and make it even more confusing. Sorry.

tools/openqa-run
```diff
@@ -0,0 +1,38 @@
...
+        --)
+            shift
+            break
+            ;;
+        *) break ;;
+    esac
+done
+
+main() {
+    sudo -u geekotest OPENQA_BASEDIR="${OPENQA_BASEDIR}" "$@"
```
Martchus, line 35: I really don't like the idea of calling the Mojo app as `geekotest`.

This script intends to use a sub directory of the Git checkout as `OPENQA_BASEDIR` to it should invoke everything as the user this checkout was made by. That's probably just the current user and not `geekotest`:

```suggestion
    "$@"
```

Note that if one just checks out the Git repo (for development, without having any packages installed) the user `geekotest` might not even exist.

tools/openqa-run
```diff
@@ -0,0 +1,38 @@
...
+Example:
+    openqa-run openqa-webui-daemon
+
+Options:
+ -h, --help         display this help
+EOF
+    exit "$1"
+}
+
+OPENQA_BASEDIR="${OPENQA_BASEDIR:-"$(dirname "$0")/../t/data"}"
```
Martchus, line 19: Not sure whether that's the best idea. I get the idea of reusing test fixtures. However, it is also a bit problematic as this might lead to accidental cluttering and changes in that directory. This might be annoying because those changes might show up in Git (and we cannot just ignore them because files in that directory are supposed to be under version control) and might also lead to failures when running unit tests (which might not be obvious to debug).

tools/openqa-run
```diff
@@ -0,0 +1,38 @@
...
+set -euo pipefail
+
+usage() {
+    cat << EOF
+Usage: openqa-run [OPTIONS...] OPENQA_SERVICE
+Call an openQA service OPENQA_SERVICE with appropriate user account and the
+base-dir set to the local working copy.
+
+Example:
+    openqa-run openqa-webui-daemon
```
Martchus, line 11: This kind of usage is very annoying without auto completion. In my helper is worked around this via shortcuts.

The script doesn't seem to ensure that e.g. `openqa-webui-daemon` is actually in the `PATH` so I guess this example also won't lead to a great out-of-the-box experience even if copied verbatimly. (The example you added to the docs does make sense, though.)

docs/Contributing.asciidoc
```diff
@@ -293,6 +293,17 @@ assets can need a big amount of disk space.
 
 WARNING: Be sure to *clear* that variable when running unit tests locally.
 
+For convenience you can use the `openqa-run` wrapper in to call openQA
+services with `OPENQA_BASEDIR` set to your local development environment with
+the appropriate service account. For example to start the openQA GRU service
```
Martchus, line 298: I disagree. The service account this script chooses is not appropriate for local development. I recommend *not* to mix the production/packaged setup with the development environment.

This description also makes it not clear at all to what `OPENQA_BASEDIR` is actually set (because "your local development environment" is not clear in this context, *apparently* it means test files from the current working dir but also *only* if `OPENQA_BASEDIR` is not set anyway).

Martchus (changes requested):
All in all I'd say: no
This is causing more confusion than it is helping.

tools/openqa-run
kalikiana: Why is `geekotest` even used here? I always considered that a detail of the production setup on e.g. o3 and not something I generally have for development

okurz: > I really don't like the idea of calling the Mojo app as `geekotest`.
> 
> This script intends to use a sub directory of the Git checkout as `OPENQA_BASEDIR` to it should invoke everything as the user this checkout was made by. That's probably just the current user and not `geekotest`:
> 
> Note that if one just checks out the Git repo (for development, without having any packages installed) the user `geekotest` might not even exist.

Yes, correct. But we need to provide a database with the according role. So should we also set the database connection to use an automatically started test database from t/test_postgresql?

> Why is `geekotest` even used here? I always considered that a detail of the production setup on e.g. o3 and not something I generally have for development

no, geekotest is also the user used within the packaged version and hence also the user that would be used in the initial database setup which one might want to reuse when running from the development directory.

tools/openqa-run
okurz: We could create a temporary directory and use that.

tools/openqa-run
Martchus: I would not recommend to reuse the same database.

okurz: So do you recommend to use an automatically started test database from t/test_postgresql?

okurz: Now rewritten in Python using typer. Starting without any user change and also added an option to start a test database

tools/openqa-run
okurz: Using the same as `make run-webui-test-env` and others accordingly. We could consider to use `openqa-run` within those Makefile targets

tools/openqa-run
okurz: Now rewritten in Python using typer which automatically allows to install auto completions plus looking up commands from the `scripts/` repo

docs/Contributing.asciidoc
okurz: now reverting to the current user and defaulting to t/data as the local test "basedir", same as `run-webui-test-env` and others

Martchus (approved):

perlpunk (approved):

PR #5797, opened 2024-07-29
Do not retry jobs that were obsoleted

AdamWill:
The 'obsoleted' result indicates the job was cancelled because a new schedule request for the same scenario came in. In this case it probably doesn't make sense to auto-retry the cancelled job because the new request indicates the admin wants the job run with the new settings.

See: https://progress.opensuse.org/issues/164613

### DIFF
--- a/lib/OpenQA/Schema/Result/Jobs.pm
+++ b/lib/OpenQA/Schema/Result/Jobs.pm
@@ -575,7 +575,8 @@ sub is_ok ($self) {
 sub is_ok_to_retry ($self) {
     return 1 unless my $result = $self->result;
     return 0 if grep { $_ eq $result } OK_RESULTS;    # retry is not needed if job is ok
-    return 0 if $result eq USER_CANCELLED;    # retry is not needed if job is user-cancelled
+    return 0 if $result eq USER_CANCELLED;    # retry is not needed if job is user-cancelled...
+    return 0 if $result eq OBSOLETED;    # ...or obsoleted
     return 1;
 }
 
--- a/t/10-jobs.t
+++ b/t/10-jobs.t
@@ -831,6 +831,10 @@ subtest 'job setting based retriggering' => sub {
     $job->done(result => USER_CANCELLED);
     perform_minion_jobs($minion);
     is $jobs->count, $jobs_nr + 1, 'no additional job retriggered if USER_CANCELLED (with retry)';
+    $job->update({state => SCHEDULED, result => NONE});
+    $job->done(result => OBSOLETED);
+    perform_minion_jobs($minion);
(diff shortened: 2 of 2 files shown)


lib/OpenQA/Schema/Result/Jobs.pm
```diff
@@ -575,7 +575,8 @@ sub is_ok ($self) {
 sub is_ok_to_retry ($self) {
     return 1 unless my $result = $self->result;
     return 0 if grep { $_ eq $result } OK_RESULTS;    # retry is not needed if job is ok
-    return 0 if $result eq USER_CANCELLED;    # retry is not needed if job is user-cancelled
+    return 0 if $result eq USER_CANCELLED;    # retry is not needed if job is user-cancelled...
+    return 0 if $result eq OBSOLETED;    # ...or obsoleted
```
b10n1k, line 579: The only consideration is whether this statement should come before the `USER_CANCELLED`. probably it doesnt make big difference though. Just as a sequence I think it is more intuitive. Other than that, LGTM

AdamWill: i really don't think it matters. in *theory* the one most likely to be encountered should come first, I guess. I could have made them into a two-item array and changed the check into a `grep`, but it just didn't feel worth it.

okurz: Did you consider to use `ABORTED_RESULTS` from lib/OpenQA/Jobs/Constants.pm? That would be

```
use constant ABORTED_RESULTS =>
  (SKIPPED, OBSOLETED, PARALLEL_FAILED, PARALLEL_RESTARTED, USER_CANCELLED, USER_RESTARTED);
```

I would assume in all those cases we can also skip the retry. @Martchus WDYT?

okurz (changes requested):

lib/OpenQA/Schema/Result/Jobs.pm
AdamWill: I specifically decided against using it because of `PARALLEL_FAILED`; we definitely want retries in that case (if job A and job B are parallel, and A fails, we definitely want B to be retried). But if, as you say, such a job actually would still be retried in this case, we can reconsider it. I think we should investigate exactly when each of those states can happen first, though. Sometimes words lie. In any case, it seems like a rather more significant change than this one (I already checked that `OBSOLETED` really is *only* used when a job is eaten by a new schedule request).

okurz: fine, your choice is the safe one

kalikiana (approved):

lib/OpenQA/Schema/Result/Jobs.pm
okurz: I dared to create
* #5806

PR #5803, opened 2024-07-30
Avoid incomplete jobs when Git cloning fails

Martchus:
* Retry Git processing instead of immediately failing the Minion job (and
  thus also setting the related openQA job to incomplete)
* Improve error messages in git_clone tasks (see separate commit message)
* See https://progress.opensuse.org/issues/163784

### DIFF
--- a/lib/OpenQA/Task/Git/Clone.pm
+++ b/lib/OpenQA/Task/Git/Clone.pm
@@ -27,10 +27,11 @@ sub _git_clone_all ($job, $clones) {
 
     # Prevent multiple git clone tasks for the same path to run in parallel
     my @guards;
+    my $retry_delay = {delay => 30 + int(rand(10))};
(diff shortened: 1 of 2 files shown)


lib/OpenQA/Task/Git/Clone.pm
```diff
@@ -27,10 +27,11 @@ sub _git_clone_all ($job, $clones) {
 
     # Prevent multiple git clone tasks for the same path to run in parallel
     my @guards;
+    my $retry_dely = {delay => 30 + int(rand(10))};
```
okurz, line 30:
```suggestion
    my $retry_delay = {delay => 30 + int(rand(10))};
```

okurz (changes requested):

okurz (approved):

PR #5808, opened 2024-07-31
Proof of Concept: Support git_auto_clone for empty CASDIR/NEEDLES_DIR

perlpunk:
Issue: https://progress.opensuse.org/issues/162632


Martchus (commented):
This looks generally good. I suppose the related ticket is https://progress.opensuse.org/issues/162125 (and not https://progress.opensuse.org/issues/162632).

lib/OpenQA/Task/Git/Clone.pm
```diff
@@ -83,35 +83,67 @@ sub _git_fetch ($path, $branch_arg) {
...
+    }
+    if (my $is_dirty = _git_diff_index($path)) {
+        $ctx->warn("Clone at '$path' has dirty status, not updating: $is_dirty");
         return;
     }
 
+    _git_gc($path);
     my $current_branch = _get_current_branch($path);
+    # TODO deal with conflicts because of save_needle
+    # see fetchneedles
```
Martchus, line 146: Maybe we can disable running fetchneedles and periodically enqueue a Minion job that runs your new code. Then we wouldn't have to worry about a conflict.

The `save_needle` task could simply try to lock the same Minion locks as this task to avoid conflicts. Note that the same counts for the `delete_needles` task.

okurz: As the git operation locking is pretty good I would prefer to rely on that so actually call git commands and let them handle the conflicts. WDYT?

okurz: ok, synchronizing on the complete sequence of commands with a minion lock sounds good then

perlpunk: This is now done already by adding a minion guard for `limit_needle_task` in `git_clone`

okurz:
s/CASDIR/CASEDIR/

PR #5809, opened 2024-07-31
Fix sporadic failures in test of candidates menu

Martchus:
`wait_for_ajax_and_animations` does not fully cover the asynchronous behaviors of the code and also doesn't help with animations. (The relevant animation is already globally disabled via `$.fx.off = true;` but this does not mean code becomes synchronous.)

With this change we simply wait for the element to show up instead.

Related ticket: https://progress.opensuse.org/issues/164745


perlpunk (approved):

okurz (approved):
That's not only fixing but IMHO better and can even speed up the test as it's only explicitly waiting for what is actually required :)

PR #5811, opened 2024-08-01
Workaround candidates menu not opening sometimes in UI tests

Martchus:
See https://progress.opensuse.org/issues/164745

---

To check whether this actually works I added an additional commit to run the relevant test 200 times on CircleCI. So this is still a draft.

### DIFF
--- a/t/ui/18-tests-details.t
+++ b/t/ui/18-tests-details.t
@@ -610,8 +610,16 @@ subtest 'test candidate list' => sub {
         \%expected_candidates, 'needles appear twice, each time under different tag');
 
     $driver->get('/tests/99946#step/installer_timezone/1');
-    wait_for_element(selector => '#candidatesMenu', is_displayed => 1)->click();
-    wait_for_element(selector => '#needlediff_selector .show-needle-info', is_displayed => 1)->click();
+    my $clicks = 0;
+    wait_for_element(
+        selector => '#needlediff_selector .show-needle-info',
+        is_displayed => 1,
+        trigger_function => sub () {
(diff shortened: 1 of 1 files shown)


Martchus:
It only failed because the CircleCI deadline was exceeded. So I'm re-triggering to give it more runs (but it'll of course exceed the deadline again).

okurz (approved):

t/ui/18-tests-details.t
```diff
@@ -610,8 +610,16 @@ subtest 'test candidate list' => sub {
...
-    wait_for_element(selector => '#candidatesMenu', is_displayed => 1)->click();
-    wait_for_element(selector => '#needlediff_selector .show-needle-info', is_displayed => 1)->click();
+    my $clicks = 0;
+    wait_for_element(
+        selector => '#needlediff_selector .show-needle-info',
+        is_displayed => 1,
+        trigger_function => sub () {
+            # open the candidates menu at the beginning and try again before every second check
+            return if $clicks != 0 && ($clicks % 2) == 1;
+            wait_for_element(selector => '#candidatesMenu', is_displayed => 1)->click;
```
kalikiana, line 620: Was wondering if we'll still see failures when this is legitimately not showing up, but I guess it doesn't help anyway unless we can reproduce it.

Martchus: We still require the elements to be present, visible and clickable. And the subsequent check would also fail because then `.needle-info-table` wouldn't be in the DOM.

kalikiana (approved):

PR #5812, opened 2024-08-01
Ease setup for Git-based workflows

Martchus:
Enable the `git_auto_clone` feature by default as it seems to work well enough in openQA-in-openQA tests.

Related tickets: https://progress.opensuse.org/issues/162077

### DIFF
--- a/docs/GettingStarted.asciidoc
+++ b/docs/GettingStarted.asciidoc
@@ -366,9 +366,8 @@ to play deeper, you can configure the whole openQA manually from scratch, but
 this document should help you to get started faster.
 
 === Getting tests
-Set `git_auto_clone = yes` in `openqa.ini`. Then you can point `CASEDIR` and
-`NEEDLES_DIR` to Git repositories. openQA will checkout those repositories
-automatically and no manual setup is needed.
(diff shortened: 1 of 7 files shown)


docs/GettingStarted.asciidoc
```diff
@@ -366,9 +366,8 @@ to play deeper, you can configure the whole openQA manually from scratch, but
 this document should help you to get started faster.
 
 === Getting tests
-Set `git_auto_clone = yes` in `openqa.ini`. Then you can point `CASEDIR` and
-`NEEDLES_DIR` to Git repositories. openQA will checkout those repositories
-automatically and no manual setup is needed.
+You can point `CASEDIR` and `NEEDLES_DIR` to Git repositories. openQA will
+checkout those repositories automatically and no manual setup is needed.
```
okurz, line 370: How about mentioning how to disable it then?

okurz (approved):

docs/GettingStarted.asciidoc
Martchus: I decided against it because we shouldn't bother people with this in the "Getting started" section. Anybody who is interested in checking out what can be tweaked will find the setting with an explanation in the default config template.

kalikiana (approved):

PR #5813, opened 2024-08-01
Revert "Remove unused 'dsn' function and its usage"

Martchus:
Reverts os-autoinst/openQA#4928 as it is actually used in production, e.g. o3

### DIFF
--- a/lib/OpenQA/Schema.pm
+++ b/lib/OpenQA/Schema.pm
@@ -51,6 +51,11 @@ sub disconnect_db () {
     $SINGLETON = undef;
 }
 
+sub dsn {
+    my $self = shift;
+    return $self->storage->connect_info->[0]->{dsn};
+}
+
 sub deploy ($self, $force_overwrite = 0) {
     # lock config file to ensure only one thing will deploy/upgrade DB at once
     # we use a file in prjdir/db as the lock file as the install process and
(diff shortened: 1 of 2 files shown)


baierjan (approved):

Martchus:
Note that the error message was:

```
Aug 01 13:57:08 ariel systemd[1]: Started The openQA web UI.
Aug 01 13:57:09 ariel openqa-webui-daemon[11005]: Can't connect to data source 'HASH(0x55d82c060598)' because I can't work out what driver to use (it doesn't seem to contain a 'dbi:driver:' prefix and the DBI_DRIVER env var is not set)>
Aug 01 13:57:09 ariel systemd[1]: openqa-webui.service: Main process exited, code=exited, status=255/EXCEPTION
Aug 01 13:57:09 ariel systemd[1]: openqa-webui.service: Failed with result 'exit-code'.
```

And it didn't help to simply remove the line `user = geekotest` from `database.ini`.

baierjan:
And of course, now the issue is in our docker compose tests:
```
webui-db-init-1  | Can't connect to data source 'HASH(0x563a5c882088)' because I can't work out what driver to use (it doesn't seem to contain a 'dbi:driver:' prefix and the DBI_DRIVER env var is not set) at /usr/lib/perl5/vendor_perl/5.26.1/Mojo/Pg.pm line 73.
```

PR #5814, opened 2024-08-01
Allow openqa-bootstrap to clone job in non-systemd environments

baierjan:
Run openqa-worker on background to allow script to continue with openqa-clone-job invocation, then wait for the worker process.

Reference: https://progress.opensuse.org/issues/164595

### DIFF
--- a/script/openqa-bootstrap
+++ b/script/openqa-bootstrap
@@ -26,7 +26,7 @@ start-database() {
 start-worker() {
     if [[ -z $running_systemd ]]; then
         /usr/bin/install -d -m 0755 -o _openqa-worker /var/lib/openqa/pool/1
-        su _openqa-worker -c '/usr/share/openqa/script/worker --instance 1'
+        su _openqa-worker -c '/usr/share/openqa/script/worker --instance 1' &
     else
         systemctl enable --now openqa-worker@1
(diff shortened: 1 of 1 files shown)


script/openqa-bootstrap
```diff
@@ -182,3 +182,5 @@ start-worker
 if [[ $# -ne 0 ]]; then
     openqa-clone-job "$@"
 fi
+
+[[ -n $running_systemd ]] || wait
```
nicksinger, line 186: we know from the current problem that `worker` will never stop by itself - wouldn't that mean `openqa-bootstrap` would again never finish but this time waiting at the end?

baierjan: Yes and that is intended, see https://progress.opensuse.org/issues/164595?#note-5

TLDR: entrypoint for openqa-single-instance container

nicksinger: ah, that explains it. Would you mind adding a short comment to that line explaining it? Maybe even referencing the comment?

nicksinger: awesome, thanks!

nicksinger (approved):

Martchus (approved):

PR #5815, opened 2024-08-01
Move 'dsn' functionality into only place where used

okurz:
In b1da48eac I removed the "dsn" function and its usage as it looked
like that would not be necessary anymore at all and we did not have any
test coverage for that. This turned out to be a false assumption leading
to a production openQA instance to not start anymore with the error
message

```
openqa-webui-daemon: Can't connect to data source 'HASH(0x55d82c060598)' because I can't work out what driver to use (it doesn't seem to contain a 'dbi:driver:' prefix and the DBI_DRIVER env var is not set)>
```

so the change was quickly reverted.

This commit now moves the 'dsn' functionality to the only place where
it is needed, simplifies the call and adds complete test coverage for
the according method with a separate, new test module
"t/15-shared-plugin-gru.t".

Tests did not cover the 'dsn' function anyway.

Verified by calling

```
env OPENQA_BASEDIR=$PWD/t/data OPENQA_DATABASE=test script/openqa-gru
```

locally.

### DIFF
--- a/lib/OpenQA/Shared/Plugin/Gru.pm
+++ b/lib/OpenQA/Shared/Plugin/Gru.pm
@@ -40,18 +40,31 @@ sub register_tasks ($self) {
       );
 }
 
+# allow the continuously polled stats to be available on an
+# unauthenticated route to prevent recurring broken requests to the login
+# provider if not logged in
+#
+# hard/impossible to mock due to name collision of "remove" method on
+# Test::MockObject, hence marking as
+# uncoverable statement
+sub _allow_unauthenticated_minion_stats ($app) {
+    my $route = $app->routes->find('minion_stats')->remove;    # uncoverable statement
+    $app->routes->any('/minion')->add_child($route);    # uncoverable statement
+}
+
 sub register ($self, $app, $config) {
     $self->app($app) unless $self->app;
     my $schema = $app->schema;
 
     my $conn = Mojo::Pg->new;
-    if (ref $schema->storage->connect_info->[0] eq 'HASH') {
-        $self->dsn($schema->dsn);
-        $conn->username($schema->storage->connect_info->[0]->{user});
-        $conn->password($schema->storage->connect_info->[0]->{password});
+    my $connect_info = $schema->storage->connect_info->[0];
+    if (ref $connect_info eq 'HASH') {
+        $self->dsn($connect_info->{dsn});
+        $conn->username($connect_info->{user});
+        $conn->password($connect_info->{password});
     }
     else {
-        $self->dsn($schema->storage->connect_info->[0]);
+        $self->dsn($connect_info);
     }
     $conn->dsn($self->dsn());
 
@@ -82,12 +95,7 @@ sub register ($self, $app, $config) {
     # Enable the Minion Admin interface under /minion
     my $auth = $app->routes->under('/minion')->to('session#ensure_admin');
     $app->plugin('Minion::Admin' => {route => $auth});
-    # allow the continuously polled stats to be available on an
-    # unauthenticated route to prevent recurring broken requests to the login
-    # provider if not logged in
-    my $route = $app->routes->find('minion_stats')->remove;
-    $app->routes->any('/minion')->add_child($route);
-
+    _allow_unauthenticated_minion_stats($app);
     my $gru = OpenQA::Shared::Plugin::Gru->new($app);
     $app->helper(gru => sub ($c) { $gru });
 }
--- a/t/15-shared-plugin-gru.t
+++ b/t/15-shared-plugin-gru.t
@@ -1,0 +1,41 @@
+#!/usr/bin/env perl
+# Copyright SUSE LLC
+# SPDX-License-Identifier: GPL-2.0-or-later
+
+use Test::Most;
+
+use Test::Warnings ':report_warnings';
+use Test::MockModule;
+use Test::MockObject;
+use FindBin;
+use lib "$FindBin::Bin/lib", "$FindBin::Bin/../external/os-autoinst-common/lib";
+use OpenQA::Test::TimeLimit '5';
+use OpenQA::App;
+use OpenQA::Shared::Plugin::Gru;    # SUT
+
+ok my $gru = OpenQA::Shared::Plugin::Gru->new, 'can instantiate gru plugin';
+
+my $connect_info = {dsn => 'my_dsn', user => 'user', password => 'foo'};
+my $storage = Test::MockObject->new->set_always(connect_info => [$connect_info]);
+my $schema = Test::MockObject->new->set_always(storage => $storage)->set_true('search_path_for_tests');
+my %config = (misc_limits => {minion_job_max_age => 0});
+my $minion = Test::MockObject->new->set_true('on');
+my $log = Test::MockObject->new->set_true('level', 'info');
+my $under = Test::MockObject->new->set_true('to');
+my $routes = Test::MockObject->new->set_always(under => $under);
+my $app
+  = Test::MockObject->new->set_always(schema => $schema)->set_always(config => \%config)->set_always(minion => $minion)
+  ->set_always(log => $log)->set_always(routes => $routes)->set_true('plugin', 'helper');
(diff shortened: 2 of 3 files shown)


baierjan (approved):

t/15-shared-plugin-gru.t
```diff
@@ -0,0 +1,35 @@
...
+my $schema = Test::MockObject->new->set_always(storage => $storage)->set_true('search_path_for_tests');
+my %config = (misc_limits => {minion_job_max_age => 0});
+my $minion = Test::MockObject->new->set_true('on');
+my $log = Test::MockObject->new->set_true('level', 'info');
+my $under = Test::MockObject->new->set_true('to');
+my $routes = Test::MockObject->new->set_always(under => $under);
+my $app
+  = Test::MockObject->new->set_always(schema => $schema)->set_always(config => \%config)->set_always(minion => $minion)
+  ->set_always(log => $log)->set_always(routes => $routes)->set_true('plugin', 'helper');
+OpenQA::App->set_singleton($app);
```
Martchus, line 30: It wouldn't hurt to add `use OpenQA::App;` even though this is apparently indirectly pulled in.

t/15-shared-plugin-gru.t
```diff
@@ -0,0 +1,35 @@
...
+my $mock = Test::MockModule->new('OpenQA::Shared::Plugin::Gru');
+$mock->noop('_allow_unauthenticated_minion_stats');
+ok $gru->register($app, undef), 'can register gru plugin';
+
```
Martchus, line 34: This test should also check whether the Minion plugin and its PostgreSQL connection were setup correctly (with the expected credentials). You can also create small test objects with sub packages like in the worker tests if `Test::MockObject` is not sufficient for this.

t/15-shared-plugin-gru.t
```diff
@@ -0,0 +1,35 @@
...
+use OpenQA::Shared::Plugin::Gru;    # SUT
+
+ok my $gru = OpenQA::Shared::Plugin::Gru->new, 'can instantiate gru plugin';
+
+my %connect_info = (dsn => 'my_dsn', user => 'user', password => 'foo');
+my $storage = Test::MockObject->new->set_always(connect_info => [\%connect_info]);
+my $schema = Test::MockObject->new->set_always(storage => $storage)->set_true('search_path_for_tests');
+my %config = (misc_limits => {minion_job_max_age => 0});
+my $minion = Test::MockObject->new->set_true('on');
+my $log = Test::MockObject->new->set_true('level', 'info');
```
Martchus, line 24: Just instantiating a real object via `Mojo::Log->new` would probably not hurt (but considering we only need two methods here mocking it is ok I suppose).

t/15-shared-plugin-gru.t
```diff
@@ -0,0 +1,35 @@
+#!/usr/bin/env perl
+# Copyright SUSE LLC
+# SPDX-License-Identifier: GPL-2.0-or-later
+
+use Test::Most;
+
+use Test::Warnings ':report_warnings';
+use Test::MockModule;
+use Test::MockObject;
+use Test::Mojo;
```
Martchus, line 10: This is not used.

t/15-shared-plugin-gru.t
```diff
@@ -0,0 +1,35 @@
...
+use FindBin;
+use lib "$FindBin::Bin/lib", "$FindBin::Bin/../external/os-autoinst-common/lib";
+use OpenQA::Test::TimeLimit '5';
+use OpenQA::Test::Case;
```
Martchus, line 14: This is not used.

lib/OpenQA/Shared/Plugin/Gru.pm
```diff
@@ -40,18 +40,31 @@ sub register_tasks ($self) {
       );
 }
 
+# allow the continuously polled stats to be available on an
+# unauthenticated route to prevent recurring broken requests to the login
+# provider if not logged in
+#
+# hard/impossible to mock due to name collision of "remove" method on
+# Test::MockObject, hence marking as
```
Martchus, line 48: You could just use a sub package like in the worker tests instead of `Test::MockObject`. Or you just use a real app by creating one via `Mojolicious->new` (which is not that expensive).

Martchus (changes requested):

t/15-shared-plugin-gru.t
okurz: Yes, and it's the most efficient for this unit-level test.

lib/OpenQA/Shared/Plugin/Gru.pm
okurz: True but would it be better than this approach? I am not sure and I would prefer to use proper mocking modules where possible.

t/15-shared-plugin-gru.t
okurz: With the current design of OpenQA::Shared::Plugin::Gru the PostgreSQL connection object is created within the method `register` which is IMHO bad design and makes testing not directly possible. We would need to move the object creation outside to inject dependencies. Then we would be able to inject a testable dependency. But I can mock the Mojo::Pg module (not object) with Test::MockModule.

I think checking anything more than just the current unit is overkill but I will do it. Updated.

t/15-shared-plugin-gru.t
okurz: done

t/15-shared-plugin-gru.t
okurz: done

t/15-shared-plugin-gru.t
```diff
@@ -0,0 +1,41 @@
...
+  ->set_always(log => $log)->set_always(routes => $routes)->set_true('plugin', 'helper');
+OpenQA::App->set_singleton($app);
+my $mock = Test::MockModule->new('OpenQA::Shared::Plugin::Gru');
+$mock->noop('_allow_unauthenticated_minion_stats');
+my $pg = Test::MockObject->new->set_true('dsn', 'username', 'password', 'search_path', 'find');
+my $pg_module = Test::MockModule->new('Mojo::Pg')->redefine(new => $pg);
+ok $gru->register($app, undef), 'can register gru plugin';
+$pg->called_ok('username', 'pg connection initialized with username');
+is(($pg->call_args(0))[1], $connect_info->{user}, 'pg connection username is correct');
+is(($pg->call_args(2))[1], $connect_info->{password}, 'pg connection password is correct');
```
Martchus, line 37: I'm wondering about this jump from 0 to 2 in the indexes as password is called directly after user but I guess it is ok.

lib/OpenQA/Shared/Plugin/Gru.pm
Martchus: Not mocking things always increases the changes that we actually test what we need to test. I'd normally only mock things if not mocking has a too big performance impact. I guess in this case it doesn't really matter, though.

I would not consider `Test::MockModule` more "proper" than just creating sub packages. If you make the sub packages inherit from `Mojo::Base` there's very little boilerplate code and code is actually easier to read. Just because someone wrote a Perl module it doesn't mean it is automatically the better approach to use it - and the API of `Test::MockModule` is not *that* intuitive.

t/15-shared-plugin-gru.t
Martchus: This is also where I think the API of `Test::MockModule` is overshooting the goal. Well have to deal with these indices here when we actually just wanted to check whether properties of a module have eventually been set to some values.

okurz: Do you have a specific code suggestion you can provide to improve? I wouldn't know otherwise.

Martchus: You can keep it. I just wanted to mention this downside of `Test::MockModule` because you said `Test::MockModule` is the "proper" way compared to just using a sub package.

okurz: We can also create a local variable and write to that in a mocked module with MockObject/MockModule

Martchus (approved):

PR #5818, opened 2024-08-05
Define single-instance container for Codespaces

baierjan:
Reference: https://progress.opensuse.org/issues/161399

This is part 1.

### DIFF
--- a/container/single-instance-codespaces/Dockerfile
+++ b/container/single-instance-codespaces/Dockerfile
@@ -1,0 +1,19 @@
+# SPDX-License-Identifier: MIT
+#!BuildTag: openqa-single-instance-codespaces:latest opensuse/openqa-single-instance-codespaces:latest opensuse/openqa-single-instance-codespaces:%PKG_VERSION% opensuse/openqa-single-instance-codespaces:%PKG_VERSION%.%RELEASE%
+
+# hadolint ignore=DL3006
+FROM openqa-single-instance
+
+# labelprefix=org.opensuse.openqa-single-instance-codespaces
+LABEL org.opencontainers.image.title="openQA single-instance container for Codespaces"
+LABEL org.opencontainers.image.description="A complete openQA instance composed of all necessary components to execute openQA tests including an openQA worker"
+LABEL org.opencontainers.image.version="%PKG_VERSION%.%RELEASE%"
+LABEL org.opensuse.reference="registry.opensuse.org/opensuse/openqa-single-instance-codespaces:%PKG_VERSION%.%RELEASE%"
+LABEL org.openbuildservice.disturl="%DISTURL%"
+LABEL org.opencontainers.image.created="%BUILDTIME%"
+# endlabelprefix
+
+# install Codespaces requirements
+# hadolint ignore=DL3037
+RUN zypper in -y awk tar && \
+    zypper clean -a


perlpunk:
Why don't you change https://github.com/os-autoinst/openQA/blob/master/container/single-instance/Dockerfile ?

b10n1k:
> Why don't you change https://github.com/os-autoinst/openQA/blob/master/container/single-instance/Dockerfile ?

it is gonna be in part 2?! :)

baierjan:
> Why don't you change https://github.com/os-autoinst/openQA/blob/master/container/single-instance/Dockerfile ?

Not sure what exactly you mean, but that's likely against *AC2* in the referenced ticket.

container/single-instance-codespaces/Dockerfile
```diff
@@ -0,0 +1,19 @@
+# SPDX-License-Identifier: MIT
+#!BuildTag: openqa-single-instance-codespaces:latest opensuse/openqa-single-instance-codespaces:latest opensuse/openqa-single-instance-codespaces:%PKG_VERSION% opensuse/openqa-single-instance-codespaces:%PKG_VERSION%.%RELEASE%
+
+# hadolint ignore=DL3006
+FROM openqa-single-instance
```
b10n1k, line 5: isnt that a Leap? I thought that it should be a TW. should we have a comment? wdyt?

baierjan: It is Leap right now, but it will be Tumbleweed in a near future, see #5819.

b10n1k: I just saw https://github.com/os-autoinst/openQA/pull/5819/files#diff-cd62a72c6c09ba757d9d23c56722b50b7c1b2e2da6a7a0c496c8a8e6e3ce8e9fR5. Still it would be nice to have a link to the container/single-instance/Dockerfile here?

perlpunk (approved):
I think having tar and awk in the "pure openqa-single-instance" container would be fine, but ok :)

container/single-instance-codespaces/Dockerfile
```diff
@@ -0,0 +1,19 @@
...
+LABEL org.opencontainers.image.description="A complete openQA instance composed of all necessary components to execute openQA tests including an openQA worker"
+LABEL org.opencontainers.image.version="%PKG_VERSION%.%RELEASE%"
+LABEL org.opensuse.reference="registry.opensuse.org/opensuse/openqa-single-instance-codespaces:%PKG_VERSION%.%RELEASE%"
+LABEL org.openbuildservice.disturl="%DISTURL%"
+LABEL org.opencontainers.image.created="%BUILDTIME%"
+# endlabelprefix
+
+# install Codespaces requirements
+# hadolint ignore=DL3037
+RUN zypper in -y awk tar && \
```
kalikiana, line 18: Why not add those to the existing container? Seems like they would be useful anyway and we wouldn't need yet another container to maintain - and I wonder where to test this? openQA-in-openQA tests or somewhere else?

baierjan: See https://github.com/os-autoinst/openQA/pull/5818#issuecomment-2269059690

It violated *AC2* as discussed during the estimation (yes, I would also think adding just those basic tools wouldn't hurt, but the ticket says otherwise).

I do not know if codespaces containers can be tested automatically, for manual testing I guess just start the environment should do the trick (although it is unclear to me if I can do that without changing the devcontainer definition).

perlpunk: testing codespaces is a bit cumbersome. you would need to branch the obs project (and enable publishing) and then in your github fork point to the container from your fork.

perlpunk: regarding awk and tar - The downside of having two containers is that it makes it even harder to test any changes, so I personally would vote against AC2.

okurz (approved):

PR #5819, opened 2024-08-05
Switch codespaces devcontainer to single-instance-codespaces

baierjan:
This goes after:
* #5818


okurz (commented):
+1

baierjan:
The newly build Tumbleweed based codespaces container is already present in the OBS.

okurz (approved):

b10n1k (approved):

PR #5820, opened 2024-08-05
Allow restarting `openqa-webui-daemon` without downtime

Martchus:
* Allow restarting `openqa-webui-daemon` without downtime by sending SIGHUP to the process or reloading the systemd unit `openqa-webui.service`
* Start the Mojolicious application with `reuse=1` as mentioned on https://docs.mojolicious.org/Mojolicious/Guides/Cookbook#Zero-downtime-software-upgrades
* Note that other services are not covered but those are also not user facing or retried and thus not required
* See https://progress.opensuse.org/issues/162533

### DIFF
--- a/dist/rpm/openQA.spec
+++ b/dist/rpm/openQA.spec
@@ -17,7 +17,9 @@
 
 
 # can't use linebreaks here!
-%define openqa_services openqa-webui.service openqa-gru.service openqa-websockets.service openqa-scheduler.service openqa-enqueue-audit-event-cleanup.service openqa-enqueue-audit-event-cleanup.timer openqa-enqueue-asset-cleanup.service openqa-enqueue-asset-cleanup.timer openqa-enqueue-result-cleanup.service openqa-enqueue-result-cleanup.timer openqa-enqueue-bug-cleanup.service openqa-enqueue-bug-cleanup.timer
+%define openqa_main_service openqa-webui.service
+%define openqa_extra_services openqa-gru.service openqa-websockets.service openqa-scheduler.service openqa-enqueue-audit-event-cleanup.service openqa-enqueue-audit-event-cleanup.timer openqa-enqueue-asset-cleanup.service openqa-enqueue-asset-cleanup.timer openqa-enqueue-result-cleanup.service openqa-enqueue-result-cleanup.timer openqa-enqueue-bug-cleanup.service openqa-enqueue-bug-cleanup.timer
+%define openqa_services %{openqa_main_service} %{openqa_extra_services}
 %define openqa_worker_services openqa-worker.target openqa-slirpvde.service openqa-vde_switch.service openqa-worker-cacheservice.service openqa-worker-cacheservice-minion.service
 %if %{undefined tmpfiles_create}
 %define tmpfiles_create() \
@@ -504,7 +506,12 @@ fi
 %service_del_preun openqa-continuous-update.timer
 
 %postun
-%service_del_postun %{openqa_services}
+# reload main service (but do not restart it via service_del_postun to minimize downtimes)
+if [ -x /usr/bin/systemctl ] && [ $1 -ge 1 ]; then
+    /usr/bin/systemctl reload %{openqa_main_service} || :
+fi
+# restart other services
+%service_del_postun %{openqa_extra_services}
 %restart_on_update apparmor
 
 %postun worker
--- a/script/openqa-webui-daemon
+++ b/script/openqa-webui-daemon
@@ -1,3 +1,38 @@
-#!/bin/sh -e
-# Our API commands are very expensive, so the default timeouts are too tight
-exec "$(dirname "$0")"/openqa prefork -m production --proxy -i 100 -H 900 -w 30 -c 1 -G 800 "$@"
+#!/bin/bash
+set -e
+
+pid=
+pid_idx=0
+pid_dir=${OPENQA_BASEDIR:-/var/lib}/openqa/webui
+openqa_args=("$@")
+openqa_dir=$(dirname "$0")
+
+function start_service {
+    # keep track of the previous and next PID
+    pid_last=$pid
+    pid_file=$pid_dir/prefork-$pid_idx.pid
+    pid_idx=$(((pid_idx + 1) % 2))
+    rm -f "$pid_file"
+
+    # start openQA in the background
+    # note: Our API commands are very expensive, so the default timeouts are too tight.
+    "$openqa_dir"/openqa prefork -m production --proxy -i 100 -H 900 -w 30 -c 1 -G 800 -P "$pid_file" "${openqa_args[@]}" &
+    pid=$!
+
+    # wait until openQA is ready to accept requests by waiting for its PID file
+    while [[ ! -e $pid_file ]] && [[ -e /proc/$pid ]]; do sleep 1; done
+
+    # terminate a previously started openQA instance
+    [[ $pid_last ]] && kill -s TERM "$pid_last"
+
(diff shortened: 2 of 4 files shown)


okurz (commented):
+1

Martchus:
It works locally when installing the packages from the OBS check. (If you want to reproduce, be sure to also install `openQA-common` because the `reuse=1` change is part of that sub package.)

I tested this by hammering the F5 key in the web browser why reloading the service via `sleep 5 && sudo systemctl reload openqa-webui`. Without this change there's a time window of around a second where one gets no connection and with the change this doesn't happen.

The journal also looks good - so the old service is really only stopped once the new one is starting:

```
Aug 05 17:20:33 linux-9lzf systemd[1]: Reloaded The openQA web UI.
Aug 05 17:20:35 linux-9lzf openqa-webui-daemon[52867]: [info] Listening at "http://127.0.0.1:9526?reuse=1"
Aug 05 17:20:35 linux-9lzf openqa-webui-daemon[52867]: Web application available at http://127.0.0.1:9526
Aug 05 17:20:35 linux-9lzf openqa-webui-daemon[52867]: [info] Listening at "http://[::1]:9526?reuse=1"
Aug 05 17:20:35 linux-9lzf openqa-webui-daemon[52867]: Web application available at http://[::1]:9526
Aug 05 17:20:35 linux-9lzf openqa-webui-daemon[52867]: [info] Manager 52867 started
…
Aug 05 17:20:35 linux-9lzf openqa-webui-daemon[52908]: [info] Worker 52908 started
Aug 05 17:20:35 linux-9lzf openqa-webui-daemon[52867]: [info] Creating process id file "/var/lib/openqa/webui/prefork-1.pid"
Aug 05 17:20:35 linux-9lzf openqa-webui-daemon[52909]: [info] Worker 52909 started
Aug 05 17:20:36 linux-9lzf openqa-webui-daemon[52799]: [warn] Stopping worker 52833 immediately
Aug 05 17:20:36 linux-9lzf openqa-webui-daemon[52799]: [warn] Stopping worker 52831 immediately
…
```

I still need to ensure that the service is not restarted on updates via the rpm scripts and that other services being restarted don't trigger a restart of the web UI. And I also need to add `reload: True` in our salt states (according to https://docs.saltproject.io/en/latest/ref/states/all/salt.states.service.html).

systemd/openqa-webui.service
```diff
@@ -1,12 +1,12 @@
 [Unit]
 Description=The openQA web UI
-Wants=openqa-setup-db.service
+Wants=openqa-setup-db.service openqa-livehandler.service openqa-websockets.service openqa-gru.service openqa-enqueue-asset-cleanup.timer openqa-enqueue-result-cleanup.timer openqa-enqueue-bug-cleanup.timer openqa-minion-restart.path
```
nicksinger, line 3: Why is this needed? Would the unit otherwise restart all the required units as well?

dist/rpm/openQA.spec
```diff
@@ -504,7 +506,12 @@ fi
 %service_del_preun openqa-continuous-update.timer
 
 %postun
-%service_del_postun %{openqa_services}
+# reload main service (but do not restart it via service_del_postun to minimize downtimes)
+if [ -x /usr/bin/systemctl ] && [ $1 -ge 1 ]; then
+    /usr/bin/systemctl reload %{openqa_main_service} || :
+fi
```
nicksinger, line 512: should `openqa_main_service` be in `service_del_postrun` on systems where systemd might not be present? (e.g. containers)

script/openqa-webui-daemon
```diff
@@ -1,3 +1,38 @@
...
+pid=
+pid_idx=0
+pid_dir=${OPENQA_BASEDIR:-/var/lib}/openqa/webui
+openqa_args=("$@")
+openqa_dir=$(dirname "$0")
+
+function start_service {
+    # keep track of the previous and next PID
+    pid_last=$pid
+    pid_file=$pid_dir/prefork-$pid_idx.pid
```
nicksinger, line 13: this would produce `prefork-_idx.pid` on first invocation, or do I miss something?

systemd/openqa-webui.service
Martchus: No, but the unit itself would be restarted if any of the other units are restarted.

script/openqa-webui-daemon
```diff
@@ -1,3 +1,38 @@
...
-# Our API commands are very expensive, so the default timeouts are too tight
-exec "$(dirname "$0")"/openqa prefork -m production --proxy -i 100 -H 900 -w 30 -c 1 -G 800 "$@"
+#!/bin/bash
+set -e
+
+pid=
+pid_idx=0
+pid_dir=${OPENQA_BASEDIR:-/var/lib}/openqa/webui
+openqa_args=("$@")
+openqa_dir=$(dirname "$0")
```
nicksinger, line 8: strictly speaking this is the script dir and not openqa dir ;) Of course both can be the same if the package installs it later into the openqa directory.

dist/rpm/openQA.spec
Martchus: I don't think it is relevant at all without systemd. We also already use the same approach for worker services where it is good enough.

script/openqa-webui-daemon
Martchus: I called it that way because it is the dir used for the openQA invocation in consistency with `openqa_args`.

systemd/openqa-webui.service
nicksinger: I understand. But having the other services in Wants would not fail openqa-webui.service if any of the other services fails. We might want to note this down somehow.

OTOH docs state "This is the recommended way to hook start-up of one unit to the start-up of another unit." - so maybe it is just the right thing to do. Feel free to resolve

script/openqa-webui-daemon
Martchus: No, all the variables are default initialized.

systemd/openqa-webui.service
Martchus: I started that this is acceptable behavior in the commit message. And I also read the systemd documentation which is in-line which what makes sense here specifically.

script/openqa-webui-daemon
nicksinger: Initialized, yes but empty: https://github.com/os-autoinst/openQA/pull/5820/files/a1f44e49621ea7f8bae63853c11400fa079069a3#diff-88ceba577fe4e4d895b021004b958cac9c30cb780fc1f443d6bcdc54f362377eR4 - It just contains a value the second the function is called

Martchus: But this is fine. `pid_last=$pid` makes sense even if `pid` is empty because of the check `[[ $pid_last ]] && …` below. Note that I also tested this and the PID files always had the expected name and `kill` was never executed accidentally or wrong arguments (which would lead to a notable failure via `set -e`).

nicksinger: yes you're right. I actually mis-clicked and was more concerned about the pid_file above but it uses `pid_idx` and not just `pid` - that is why I usually like ${} a little bit more. Anyhow, no problem then.

nicksinger (approved):

okurz (approved):

Martchus:
I tested the package from OBS checks locally and it works. So reinstalling/updating the package now causes the main service to reload and other services are still restarted.

I also didn't run into any limits regarding PostgreSQL connections. However, in production we might have other limits so I'll check whether I can run two prefork instances (which all the usual settings) in parallel on OSD and o3.

PR #5822, opened 2024-08-05
Get parent class as one row to avoid warnings

kalikiana:
There should only ever be one match.

See: https://progress.opensuse.org/issues/154735

### DIFF
--- a/lib/OpenQA/Schema/ResultSet/JobSettings.pm
+++ b/lib/OpenQA/Schema/ResultSet/JobSettings.pm
@@ -68,4 +68,14 @@ sub query_for_settings ($self, $args) {
     return $self->search({-and => \@conds});
 }
 
+sub parent_worker_class ($self, $parent_job_id, $worker_class) {
+    my $result = $self->search({parent_job_id => $job_id, key => 'WORKER_CLASS'})->next;
+    my $parent_worker_class = $result ? $result->value : '';
+    die
+      "Worker class $worker_class does not match the worker class of its directly chained parent ($parent_worker_class)"
+      unless $worker_class eq $parent_worker_class;
+    die "Ambiguous job settings for dependent jobs" if $result->next;
+    return $parent_worker_class;
+}
+
 1;
--- a/lib/OpenQA/Schema/ResultSet/Jobs.pm
+++ b/lib/OpenQA/Schema/ResultSet/Jobs.pm
@@ -154,8 +154,9 @@ sub create_from_settings {
         my $dependency_type = $dependency_definition->{dependency_type};
         for my $id (@$ids) {
             if ($dependency_type eq OpenQA::JobDependencies::Constants::DIRECTLY_CHAINED) {
-                my $parent_worker_class = $job_settings->find({job_id => $id, key => 'WORKER_CLASS'});
-                _handle_directly_chained_dep($parent_worker_class, $id, \%settings);
+                my $parent_worker_class = $job_settings->parent_worker_class($id, $settings->{WORKER_CLASS});
+          # assume we want to use the worker class from the parent here (and not the default which is otherwise assumed)
+                $settings->{WORKER_CLASS} = $parent_worker_class;
             }
             push(@{$new_job_args{parents}}, {parent_job_id => $id, dependency => $dependency_type});
         }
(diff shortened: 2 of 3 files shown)


lib/OpenQA/Schema/ResultSet/JobSettings.pm
```diff
@@ -68,4 +68,14 @@ sub query_for_settings ($self, $args) {
     return $self->search({-and => \@conds});
 }
 
+sub parent_worker_class ($self, $parent_job_id, $worker_class) {
+    my $result = $self->search({parent_job_id => $job_id, key => 'WORKER_CLASS'})->next;
```
perlpunk, line 72: if we only want one row, it's better to tell that the database

```suggestion
    my $result = $self->search({parent_job_id => $job_id, key => 'WORKER_CLASS'}, {rows => 1})->single;
```
since there can be multiple WORKER_CLASS entries, we're still relying on the internal order of postgres here I guess.

lib/OpenQA/Schema/ResultSet/JobSettings.pm
```diff
@@ -68,4 +68,14 @@ sub query_for_settings ($self, $args) {
...
+    my $parent_worker_class = $result ? $result->value : '';
+    die
+      "Worker class $worker_class does not match the worker class of its directly chained parent ($parent_worker_class)"
+      unless $worker_class eq $parent_worker_class;
+    die "Ambiguous job settings for dependent jobs" if $result->next;
```
perlpunk, line 77: There can be multiple entries

lib/OpenQA/Schema/ResultSet/Jobs.pm
```diff
@@ -154,8 +154,9 @@ sub create_from_settings {
         my $dependency_type = $dependency_definition->{dependency_type};
         for my $id (@$ids) {
             if ($dependency_type eq OpenQA::JobDependencies::Constants::DIRECTLY_CHAINED) {
-                my $parent_worker_class = $job_settings->find({job_id => $id, key => 'WORKER_CLASS'});
-                _handle_directly_chained_dep($parent_worker_class, $id, \%settings);
+                my $parent_worker_class = $job_settings->parent_worker_class($id, $settings->{WORKER_CLASS});
+          # assume we want to use the worker class from the parent here (and not the default which is otherwise assumed)
+                $settings->{WORKER_CLASS} = $parent_worker_class;
```
perlpunk, line 159:
```suggestion
                $settings->{WORKER_CLASS} = $parent_worker_class->value;
```
Also, before the code also handled the situation when the job had no WORKER_CLASS at all

perlpunk: ah, nevermind, that's done in the other function

perlpunk:
Maybe we even want to copy all WORKER_CLASS entries instead of just the first?

perlpunk:
OTOH I'm wondering why we don't see this warning more often.

okurz (changes requested):
https://github.com/os-autoinst/openQA/pull/5822/files#r1704371011

lib/OpenQA/Schema/ResultSet/JobSettings.pm
Martchus, line 77: We may have multiple `WORKER_CLASS` settings per job and that's perfectly fine. This code should actually get all of them and then compare whether the full set of worker classes is identical to the full set of worker classes of the parent job (the order doesn't matter).

Note that this means that already the line `$job->settings->find({key => 'WORKER_CLASS'});` is problematic. Also in this case we need to handle that there can be multiple worker classes (and then compare it to the multiple worker classes the call you're trying to fix might return).

Martchus (changes requested):

lib/OpenQA/Schema/ResultSet/JobSettings.pm
perlpunk: Right, like I wrote. And in the case of `ResultSet::Jobs::create_from_settings` we probably want to copy every WORKER_CLASS to the child

kalikiana:
Closing in favor of #5936

PR #5825, opened 2024-08-06
WIP Allow excluding groupless jobs through filtering

r-richardson:
Adds the ability to filter out (exclude) groupless jobs, as well as jobs of specific groups

Related ticket
[155398](https://progress.opensuse.org/issues/155398)

### DIFF
--- a/lib/OpenQA/WebAPI/Controller/API/V1/Job.pm
+++ b/lib/OpenQA/WebAPI/Controller/API/V1/Job.pm
@@ -85,6 +85,7 @@ sub list ($self) {
     $validation->optional('limit')->num;
     $validation->optional('offset')->num;
     $validation->optional('groupid')->num;
+    $validation->optional('not_groupid')->num; 
 
     my $limits = OpenQA::App->singleton->config->{misc_limits};
     my $limit = min($limits->{generic_max_limit}, $validation->param('limit') // $limits->{generic_default_limit});
@@ -124,6 +125,11 @@ sub list ($self) {
     my $latest = $validation->param('latest');
     my $schema = $self->schema;
     my $rs = $schema->resultset('Jobs')->complex_query(%args);
+
+    if (defined(my $not_groupid = $self->param('not_groupid'))) {
+        $rs = $rs->search({group_id => {-not_in => [$not_groupid]}});
+    }
+
     my @jobarray = defined $latest ? $rs->latest_jobs : $rs->all;
 
     # Pagination
--- a/lib/OpenQA/WebAPI/Controller/Test.pm
+++ b/lib/OpenQA/WebAPI/Controller/Test.pm
@@ -727,12 +727,21 @@ sub _prepare_job_results ($self, $all_jobs, $limit) {
 }
 
 sub _prepare_groupids ($self) {
-    return [0] unless my $groups = $self->groups_for_globs;
-    return [map { $_->id } @$groups] if @$groups;
-
     my $v = $self->validation;
     $v->optional('groupid')->num(0, undef);
-    return $v->is_valid('groupid') ? $self->every_param('groupid') : undef;
+    $v->optional('not_groupid')->num(0, undef);
+
+    my $groupids = $v->is_valid('groupid') ? $self->every_param('groupid') : [];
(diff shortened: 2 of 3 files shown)


lib/OpenQA/WebAPI/Controller/API/V1/Job.pm
```diff
@@ -124,6 +125,11 @@ sub list ($self) {
     my $latest = $validation->param('latest');
     my $schema = $self->schema;
     my $rs = $schema->resultset('Jobs')->complex_query(%args);
+
+    if (defined(my $not_groupid = $self->param('not_groupid'))) {
+        $rs = $rs->search({group_id => {-not_in => [$not_groupid]}});
+    }
+
```
okurz, line 132:
```suggestion
    if (defined(my $not_groupid = $self->param('not_groupid'))) {
        $rs = $rs->search({group_id => {-not_in => [$not_groupid]}});
    }
```

lib/OpenQA/WebAPI/Controller/Test.pm
```diff
@@ -727,12 +727,21 @@ sub _prepare_job_results ($self, $all_jobs, $limit) {
...
+    my $not_groupids = $v->is_valid('not_groupid') ? $self->every_param('not_groupid') : [];
+
+    if (@$not_groupids) {
+        my %not_groupid_hash = map { $_ => 1 } @$not_groupids;
+        $groupids = [grep { !$not_groupid_hash{$_} } @$groupids];
+    }
+
+    return $groupids if @$groupids;
+    return [0] unless my $groups = $self->groups_for_globs;
+    return [map { $_->id } @$groups];
```
Martchus, line 744: I'm wondering what the point of this change is. It is just filtering one set of query parameters by another set of query parameters.

lib/OpenQA/WebAPI/Plugin/Helpers.pm
```diff
@@ -480,6 +481,15 @@ sub _compose_job_overview_search_args ($c) {
     # allow filtering by group ID or group name
     $search_args{groupids} = [map { $_->id } @groups] if @groups;
 
+    # allow excluding jobs by group ID (0 to exclude groupless jobs)
+    if ($v->is_valid('not_groupid')) {
+        my @not_group_ids = @{$v->every_param('not_groupid')};
```
Martchus, line 486: It makes sense to avoid a copy in cases like this:
```suggestion
        my $not_group_ids = $v->every_param('not_groupid');
```
Of course you'll need to change the following lines to use `$not_group_ids` instead of `@not_group_ids`.

lib/OpenQA/WebAPI/Plugin/Helpers.pm
```diff
@@ -480,6 +481,15 @@ sub _compose_job_overview_search_args ($c) {
...
+        $search_args{groupid} = {-not_in => \@not_group_ids};
+        if (grep { $_ == 0 } @not_group_ids) {
+            $search_args{groupid} = {-not_in => \@not_group_ids, '!=' => undef};
```
Martchus, line 489: This is just overriding the previous assignment. I'm not sure whether this is going to work as intended but we'll see when you've added tests.

lib/OpenQA/WebAPI/Controller/API/V1/Job.pm
```diff
@@ -124,6 +125,11 @@ sub list ($self) {
     my $latest = $validation->param('latest');
     my $schema = $self->schema;
     my $rs = $schema->resultset('Jobs')->complex_query(%args);
+
+    if (defined(my $not_groupid = $self->param('not_groupid'))) {
```
Martchus, line 129: Not sure how other parameters are handled here but this should probably use the value from the validation object:
```suggestion
    if (defined(my $not_groupid = $validation->param('not_groupid'))) {
```

Martchus (changes requested):

r-richardson:
As this is no longer in scope of [155398](https://progress.opensuse.org/issues/155398), i'm closing this PR for now.
(The exclusion of groupless jobs is already possible and instructions [have been added in form of a help text.](https://github.com/os-autoinst/openQA/commit/4e79e5673be3e3c1a1d7260478c83eeecab6dad9))

PR #5837, opened 2024-08-07
Add filter to exclude groupless jobs to API/V1

r-richardson:
Related Ticket: https://progress.opensuse.org/issues/155398

### DIFF
--- a/lib/OpenQA/Schema/ResultSet/Jobs.pm
+++ b/lib/OpenQA/Schema/ResultSet/Jobs.pm
@@ -284,6 +284,15 @@ sub _prepare_complex_query_search_args ($self, $args) {
         push @conds, {'me.group_id' => {-in => $subquery}};
     }
 
+    if (defined $args->{not_groupid}) {
+        my $id = $args->{not_groupid};
+        if ($id) {
+            push @conds, {-or => [{'me.group_id' => {-not_in => $id}}, {'me.group_id' => undef},]};
+        }
+        else {
+            push @conds, {'me.group_id' => {-not => undef}};
+        }
+    }
     if ($args->{ids}) {
         push @conds, {'me.id' => {-in => $args->{ids}}};
     }
--- a/t/api/04-jobs.t
+++ b/t/api/04-jobs.t
@@ -123,6 +123,21 @@ subtest 'check job group' => sub {
     is(scalar(@{$t->tx->res->json->{jobs}}), 0);
 };
 
+subtest 'exclude groupless jobs' => sub {
+    my %jobs = map { $_->{id} => $_ } @jobs;
+    is($jobs{99928}->{state}, 'scheduled', 'groupless job is listed');
+    $t->get_ok('/api/v1/jobs?not_groupid=0')->status_is(200);
+    @jobs = @{$t->tx->res->json->{jobs}};
+    is scalar @jobs, 15, 'groupless jobs are excluded';
+};
+
+subtest 'exclude specific group' => sub {
+    my %jobs = map { $_->{id} => $_ } @jobs;
+    $t->get_ok('/api/v1/jobs?not_groupid=1001')->status_is(200);
+    @jobs = @{$t->tx->res->json->{jobs}};
+    is scalar @jobs, 4, 'jobs of specified groups are excluded';
+};
+
 subtest 'restricted query' => sub {
     $t->get_ok('/api/v1/jobs?iso=openSUSE-13.1-DVD-i586-Build0091-Media.iso');
     is(scalar(@{$t->tx->res->json->{jobs}}), 6, 'query for existing jobs by iso');
--- a/lib/OpenQA/WebAPI/Controller/API/V1/Job.pm
+++ b/lib/OpenQA/WebAPI/Controller/API/V1/Job.pm
@@ -85,6 +85,7 @@ sub list ($self) {
     $validation->optional('limit')->num;
     $validation->optional('offset')->num;
     $validation->optional('groupid')->num;
+    $validation->optional('not_groupid')->num;
 
     my $limits = OpenQA::App->singleton->config->{misc_limits};
     my $limit = min($limits->{generic_max_limit}, $validation->param('limit') // $limits->{generic_default_limit});
@@ -99,7 +100,7 @@ sub list ($self) {
     $args{limit} = $limit + 1;
     $args{offset} = $offset;
     my @args = qw(build iso distri version flavor scope group groupid
-      before after arch hdd_1 test machine worker_class
+      not_groupid before after arch hdd_1 test machine worker_class
       modules modules_result);
     for my $arg (@args) {
         next unless defined(my $value = $self->param($arg));


t/api/04-jobs.t
```diff
@@ -123,6 +123,14 @@ subtest 'check job group' => sub {
     is(scalar(@{$t->tx->res->json->{jobs}}), 0);
 };
 
+subtest 'exclude groupless jobs' => sub { 
+    my %jobs = map { $_->{id} => $_ } @jobs;
+    is($jobs{99928}->{state}, 'scheduled', 'groupless job is listed');
```
Martchus, line 128:
```suggestion
    is($jobs{99928}->{state}, SCHEDULED, 'groupless job is listed');
```
(Maybe you'll need to add a `use …;` to use the constant.)

t/ui/02-list-group.t
```diff
@@ -17,7 +17,21 @@ use OpenQA::SeleniumTest;
...
+    DISTRI => 'opensuse',
+    BUILD => '0091',
+    VERSION => '13.1',
+    MACHINE => '32bit',
+    ARCH => 'i586'
+};
+
 subtest 'groupid' => sub {
+    pass 'skip';
+    return;
```
Martchus, line 34: Looks like a leftover.
```suggestion
```

t/ui/02-list-group.t
```diff
@@ -87,6 +90,8 @@ subtest 'group_glob and not_group_glob' => sub {
             id => 99954,
             group_id => 1006
         });
+    pass 'skip';
+    return;
```
Martchus, line 94: Looks like a leftover.
```suggestion
```

t/ui/02-list-group.t
```diff
@@ -153,5 +158,19 @@ subtest 'group_glob and not_group_glob' => sub {
     };
 };
 
+subtest 'filter for not_groupid' => sub {
+    $schema->resultset('Jobs')->create({
+            %$template,
+            id => 99955,
+            group_id => undef
+          });
```
Martchus, line 166:
```suggestion
    $schema->resultset('Jobs')->create({%$template, id => 99955, group_id => undef});
```
(Unless tidy force you. You could also just modify `$template` which would probably be faster.)

t/ui/02-list-group.t
```diff
@@ -153,5 +158,19 @@ subtest 'group_glob and not_group_glob' => sub {
...
+
+    ok $driver->get('/tests?not_groupid=0'), 'list jobs excluding groupless';
+        wait_for_ajax(msg => 'wait for test list excluding groupless');
```
Martchus, line 169: The indentation is broken here.

t/ui/02-list-group.t
```diff
@@ -153,5 +158,19 @@ subtest 'group_glob and not_group_glob' => sub {
...
+        my @rows = $driver->find_child_elements($driver->find_element('#results tbody'), 'tr');
+        ok @rows > 0, 'grouped jobs still show up';
+        ok ! $driver->find_element('#results #job_99955'), '99955 not listed, as expected';
```
Martchus, line 172: One ID is enough (as IDs must be unique within the document):
```suggestion
        ok ! $driver->find_element('#job_99955'), '99955 not listed, as expected';
```

Martchus (dismissed):

lib/OpenQA/Schema/ResultSet/Jobs.pm
```diff
@@ -273,6 +273,9 @@ sub _prepare_complex_query_search_args ($self, $args) {
         my $subquery = $schema->resultset('JobGroups')->search({name => $args->{group}})->get_column('id')->as_query;
         push @conds, {'me.group_id' => {-in => $subquery}};
     }
+    if (defined $args->{not_groupid}) {
+      push @conds, {'me.group_id' => {-not_in => $args->{not_groupid}}};
```
perlpunk, line 277: Just a note, as we do not have groups with id 0, but null: this is turned into
```
WHERE ( me.group_id NOT IN ( 0 ) )
```
which works because in SQL, comparing something with `null` is always false, so any random number would work.

t/ui/02-list-group.t
```diff
@@ -17,7 +17,21 @@ use OpenQA::SeleniumTest;
 my $t = Test::Mojo->new('OpenQA::WebAPI');
 driver_missing unless my $driver = call_driver;
 
+my $template = {
```
perlpunk, line 20: All the changes in this test seem to be leftovers from your previous attempt, right?
Didn't we say that `group_glob=*` should do it? If there's not yet a test for that, you could add one, but I suggest in a different PR.

r-richardson:
When trying to only exclude a specific group, it will still exclude all groupless jobs (this is why the second test i added still fails).
Is this maybe due to the comparison behaviour tina mentioned ?

> Just a note, as we do not have groups with id 0, but null: this is turned into
> 
> ```
> WHERE ( me.group_id NOT IN ( 0 ) )
> ```
> 
> which works because in SQL, comparing something with `null` is always false, so any random number would work.

In the test data are 14 Jobs with the groupid 1001, one with groupid 1002 and three groupless jobs.

perlpunk:
Comparing somthing with `null`, even if it is a negative comparison, will always be false:
```
openqa=> select id, group_id from jobs where id = 4389101;
   id    | group_id 
---------+----------
 4389101 |         
(1 row)

openqa=> select id, group_id from jobs where id = 4389101 and group_id not in (23);
 id | group_id 
----+----------
(0 rows)

openqa=> select id, group_id from jobs where id = 4389101 and group_id != 23;
 id | group_id 
----+----------
(0 rows)
```
An additional `or group_id is null` would be necessary.

okurz (approved):

lib/OpenQA/Schema/ResultSet/Jobs.pm
```diff
@@ -273,6 +273,9 @@ sub _prepare_complex_query_search_args ($self, $args) {
         my $subquery = $schema->resultset('JobGroups')->search({name => $args->{group}})->get_column('id')->as_query;
         push @conds, {'me.group_id' => {-in => $subquery}};
     }
+    if (defined $args->{not_groupid}) {
+        push @conds, {'me.group_id' => {-not_in => $args->{not_groupid}}};
+    }
```
perlpunk, line 278: How about this:
```suggestion
    if (defined $args->{not_groupid}) {
        my $id = $args->{not_groupid;
        if ($id) {
            push @conds, {-or => [
                {'me.group_id' => {-not_in => $args->{not_groupid}}},
                {'me.group_id' => {-not => undef}},
                ]};
        }
        else {
            push @conds, {'me.group_id' => {-not => undef}};
        }
    }
```
not tested, so check the generated SQL.

perlpunk (changes requested):

lib/OpenQA/Schema/ResultSet/Jobs.pm
r-richardson: thank you, this lets my test pass :) :+1: 
Not sure if i'm doing this right, i'm trying to have a look at the queries using `export DBI_TRACE=1;...` but it does not show a statement with the group_id value in the condition anywhere. I only find it after a INSERT and some SELECT statements

perlpunk: DBI_TRACE should show you everything. 
mind that the shown SQL statements are using placeholders, with the actual values listed after it

r-richardson: `DBIC_TRACE` is alot better readable, thanks. Tried again to [find the query](https://drive.zoopa.dev/code/#/2/code/view/2BZBotsLtFFFE-ibOMjz3fBJOLTXNWFgUtNu4LzaK94/embed/present/) with grep but havent found it yet

perlpunk: I can see it:
```
SELECT me.id, me.result_dir, me.archived, me.state, me.priority, me.result, me.reason, me.clone_id, me.blocked_
by_id, me.TEST, me.DISTRI, me.VERSION, me.FLAVOR, me.ARCH, me.BUILD, me.MACHINE, me.group_id, me.assigned_worke
r_id, me.t_started, me.t_finished, me.logs_present, me.passed_module_count, me.failed_module_count, me.softfail
ed_module_count, me.skipped_module_count, me.externally_skipped_module_count, me.scheduled_product_id, me.resul
t_size, me.t_created, me.t_updated FROM jobs me WHERE ( ( me.group_id NOT IN ( ? ) OR me.group_id IS NULL ) ) O
RDER BY me.id DESC LIMIT ?: '1001', '10001'
```

perlpunk (approved):

perlpunk:
The circleci codecov problem still exists, it just doesn't run, and it's not possible to restart the workflow.
https://app.circleci.com/pipelines/github/os-autoinst/openQA/14269/workflows/c0550356-11a0-4dd0-a7a3-3b856edbdd63
I will merge manually now.

PR #5838, opened 2024-08-08
Use new perltidy wrapper tools/tidyall

perlpunk:
Issue: https://progress.opensuse.org/issues/155188


kalikiana (approved):

perlpunk:
I now added `shebang = perl` to `.tidyallrc`
https://metacpan.org/dist/Code-TidyAll/view/bin/tidyall#shebang

perlpunk:
I think I now found a configuration that works for all.
https://metacpan.org/dist/Code-TidyAll/view/bin/tidyall#[class]-or-[class-description]
We can have more than one section. In one we specify the files with a file extension, in the other we specify the scripts and use the shebang option.

external/os-autoinst-common/.github/workflows/base-commit-message-checker.yml
```diff
@@ -26,7 +26,7 @@ jobs:
       - name: Check subject line length
         uses: gsactions/commit-message-checker@v2
         with:
-          pattern: '^.{1,72}(\n|$)'
+          pattern: '^(Revert "?)?.{1,72}(\n|$)'
```
b10n1k, line 29: Does this count `Revert`+72chars right? which exceeds the limit of 72 IIUC! And it makes the error msg incorrect. Just making sure that this is what we want.

.tidyallrc
```diff
@@ -1,6 +1,11 @@
...
-select = {tools,lib,t}/**/*.{pl,pm,t} script/*
+; We split into modules and scripts because the scripts have no file
+; extension, so we detect them by their shebang
+
+[PerlTidy modules]
+select = {lib,t}/**/*.{pl,pm,t}
 argv = --profile=$ROOT/.perltidyrc
+
+[PerlTidy scripts]
+select = tools/* script/*
```
b10n1k, line 9: s/tools/*/tools/**/*/? 
or doesnt it matter?

external/os-autoinst-common/.github/workflows/base-commit-message-checker.yml
perlpunk: This is an update of the external repo. If it's wrong, then it needs to be fixed upstream, so please file an issue there

.tidyallrc
perlpunk: We don't have subdirectories in there

baierjan (approved):

okurz (approved):

perlpunk:
We don't get codecov patch info, but that's not relevant here, so merging

PR #5839, opened 2024-08-08
Fix firewall configuration in Autoyast profile

b10n1k:
Configure the firewall properties as SuSEfirewall2 has been replaced by firewalld and the validation produce a warning. FW_CONFIGURATIONS_EXT keeps working so far because is supported but it is deprecated.

* Avoid deprecation warning about using firewall property
  from SuSEfirewall2 times
* Use firewalld configuration instead

https://progress.opensuse.org/issues/156169

### DIFF
--- a/contrib/ay-openqa-worker-leap-3-nvme.xml
+++ b/contrib/ay-openqa-worker-leap-3-nvme.xml
@@ -139,7 +139,15 @@
   <firewall>
     <enable_firewall t="boolean">true</enable_firewall>
     <start_firewall t="boolean">true</start_firewall>
-    <FW_CONFIGURATIONS_EXT>sshd</FW_CONFIGURATIONS_EXT>
+    <default_zone>public</default_zone>
+    <zones config:type="list">
+      <zone>
+	<name>public</name>
+	<services config:type="list">
+          <service>ssh</service>
+	</services>
+      </zone>
+    </zones>
   </firewall>
   <services-manager>
     <services>
--- a/contrib/ay-openqa-worker.xml.erb
+++ b/contrib/ay-openqa-worker.xml.erb
@@ -98,7 +98,15 @@
   <firewall>
     <enable_firewall config:type="boolean">true</enable_firewall>
     <start_firewall config:type="boolean">true</start_firewall>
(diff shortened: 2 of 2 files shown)


okurz (commented):
Can you clarify how you tested that? As we don't have automatic tests for autoyast yet that matters :)

b10n1k:
> Can you clarify how you tested that? As we don't have automatic tests for autoyast yet that matters :)

with `yast autoyast check-profile`. I didnt try to perform an installation in some VM.

okurz:
> > Can you clarify how you tested that? As we don't have automatic tests for autoyast yet that matters :)
> 
> with `yast autoyast check-profile`. I didnt try to perform an installation in some VM.

hm, I don't know if `yast autoyast check-profile` is depending on the OS version on which it's called or if it can be trusted to not produce false positives so please verify with an actual installation if this works as expected.

Martchus:
The commit message is not very easy to read and contains a few grammar mistakes. This is also technically not fixing anything at this point. What about:

```
Avoid using deprecated Autoyast config for firewall

* Avoid deprecation warning about using firewall property
  from SuSEfirewall2 times
* Use firewalld configuration instead

Related ticket: https://progress.opensuse.org/issues/156169
```

And yes, it would make sense to test this, e.g. by installing it in a VM.

contrib/ay-openqa-worker-leap-3-nvme.xml
```diff
@@ -139,7 +139,15 @@
   <firewall>
     <enable_firewall t="boolean">true</enable_firewall>
     <start_firewall t="boolean">true</start_firewall>
-    <FW_CONFIGURATIONS_EXT>sshd</FW_CONFIGURATIONS_EXT>
+    <default_zone>public</default_zone>
+    <zones config:type="list">
+      <zone>
+	<name>public</name>
+	<services config:type="list">
+          <service>ssh</service>
```
kalikiana, line 147: I guess you've not add any CI so we still need a manual validation run on o3?

kalikiana: https://openqa.opensuse.org/tests/4415391#settings

kalikiana (approved):

perlpunk (approved):

Martchus (approved):

PR #5840, opened 2024-08-09
openqa-bootstrap: Refresh and retry in case of package changes

kalikiana:
Packages may be deleted before they can be installed.

See: https://progress.opensuse.org/issues/165024

### DIFF
--- a/script/openqa-bootstrap
+++ b/script/openqa-bootstrap
@@ -76,6 +76,9 @@ if [[ $NAME = "openSUSE Leap" ]]; then
     zypper -n --gpg-auto-import-keys refresh
 fi
 
+# Ensure we have retry available to handle temporary package issues later
+command -v retry > /dev/null || zypper -n --gpg-auto-import-keys in retry
+
 # install packages
 pkgs=(qemu-arm qemu-ppc qemu-x86 qemu-tools sudo iputils os-autoinst-distri-opensuse-deps)
 
@@ -92,7 +95,8 @@ fi
 if zypper -n search -x qemu-hw-display-virtio-gpu-pci; then
     pkgs+=(qemu-hw-display-virtio-gpu qemu-hw-display-virtio-gpu-pci)
 fi
-zypper -n install --no-recommends "${pkgs[@]}"
+packages="${pkgs[*]}"
+retry -e -s 30 -r 7 -- sh -c "zypper -n --gpg-auto-import-keys ref && zypper -n in --no-recommends $packages"
 
 # setup database
 chown -R postgres: /var/lib/pgsql/ # fix broken postgres working dir permissions in the nspawn container


script/openqa-bootstrap
```diff
@@ -92,7 +95,8 @@ fi
 if zypper -n search -x qemu-hw-display-virtio-gpu-pci; then
     pkgs+=(qemu-hw-display-virtio-gpu qemu-hw-display-virtio-gpu-pci)
 fi
-zypper -n install --no-recommends "${pkgs[@]}"
+packages="${pkgs[@]}"
+retry -e -s 30 -r 7 -- sh -c "zypper -n --gpg-auto-import-keys ref && zypper -n in--no-recommends $packages"
```
okurz, line 99: With that change we would do

```
zypper ref
zypper in retry
retry zypper ref
retry zypper in
```

which is quite inefficient in the happy path when retry is not needed. I suggest the following

```
command -v retry >/dev/null || zypper --n --gpg-auto-import-keys in retry
retry -e -s 30 -r 7 -- sh -c "zypper -n --gpg-auto-import-keys ref && zypper -n in--no-recommends $packages"
```

this way in the best case retry already exists and we only call `zypper ref`. In case retry needs to be installed the initial install request for `retry` would refresh as desired but at this point we don't need the additional repositories we added so we do not necessarily need a `ref`.

okurz (changes requested):

b10n1k (approved):

okurz:
From CI checks:

```
+ make test-checkstyle

In script/openqa-bootstrap line 98:
packages="${pkgs[@]}"
         ^----------^ SC2124 (warning): Assigning an array to a string! Assign as array, or use * instead of @ to concatenate.

For more information:
  https://www.shellcheck.net/wiki/SC2124 -- Assigning an array to a string! A...
make: *** [Makefile:302: test-shellcheck] Error 1
```

okurz (approved):

script/openqa-bootstrap
```diff
@@ -76,6 +76,9 @@ if [[ $NAME = "openSUSE Leap" ]]; then
     zypper -n --gpg-auto-import-keys refresh
 fi
 
+# Ensure we have retry available to handle temporary package issues later
+command -v retry >/dev/null || zypper -n --gpg-auto-import-keys in retry
```
okurz, line 80:
```suggestion
command -v retry > /dev/null || zypper -n --gpg-auto-import-keys in retry
```

okurz (changes requested):
Another failed style check but trivial to fix

script/openqa-bootstrap
kalikiana: Interesting. Didn't come up here when I ran the style check locally.

kalikiana: Hum. Maybe it just got drowned in verbose output.

okurz (approved):

PR #5841, opened 2024-08-09
Use perlcritic wrapper and plugins from common repo

perlpunk:
Issue: https://progress.opensuse.org/issues/155188

This will currently fail because
* the HashKeyQuotes.pm doesn't handle all kinds of hash keys and needs a fix
* some files need to be adjusted to the new plugins

edit: or actually it won't fail because so far we don't run perlcritic in CI...


b10n1k (approved):

perlpunk:
Something is odd with circleci. will close and create a new PR.

PR #5851, opened 2024-08-13
WIP Extend test coverage of openQA::Git

r-richardson:
I'm trying to get openQA::Git test coverage to 100% though its not working yet. My test is currently still running commit and push on my actual git dev branch even though i mocked `run_cmd_with_log_return_error` (which is used by openQA::Git) to simply return success or failures. I've also tried using set_series for the MockObject instead but that did not help so far


t/16-utils-git.t
```diff
@@ -0,0 +1,44 @@
...
+use Test::Warnings ':report_warnings';
+use Test::MockModule;
+use Test::MockObject;
+use FindBin;
+use lib "$FindBin::Bin/lib", "$FindBin::Bin/../external/os-autoinst-common/lib";
+use OpenQA::Test::TimeLimit '10';
+use OpenQA::Git;
+
+my $mock = Test::MockModule->new('OpenQA::Utils');
+$mock->redefine('run_cmd_with_log_return_error', sub {
```
perlpunk, line 14: Have a look at t/16-utils-runcmd.t
`run_cmd_with_log_return_error` is an imported function, so it is already defined in OpenQA::Git when you create the OpenQA::Util mock.
Just do it like in t/16-utils-runcmd.t

Maybe the tests from there should be moved to your new test?
It would be a bit weird to have them in separate files.

PR #5852, opened 2024-08-13
ci: Ensure tests pass even under high load

perlpunk:
Workers don't accept jobs if the load is too high.

Issua: https://progress.opensuse.org/issues/165198


okurz (approved):
I would prefer less duplication. Could be a helper file called from both locations

kalikiana (approved):

PR #5861, opened 2024-08-15
Remove tour to drop problematic dependency

Martchus:
See particular commit messages and https://progress.opensuse.org/issues/163004

---

I recommend to look at the individual commits.


t/api/15-users.t
```diff
@@ -31,4 +31,7 @@ $t->ua(OpenQA::Client->new(apikey => 'LANCELOTKEY01', apisecret => 'MANYPEOPLEKN
 $t->app($app);
 $t->delete_ok('/api/v1/user/99904')->status_is(403, 'non-admins cannot delete users');
 
+$t->post_ok('/api/v1/feature?version=42')->status_is(200, 'can set the feature version of current user');
+is $app->schema->resultset('Users')->find(99902)->feature_version, 42, 'feature version was updated';
```
b10n1k, line 35: kinda nitpicking but I think the test checks the value in the DB. it doesnt update  it. I mean I understand that it reffers to the previous action but maybe rephrase it. 
Something like `Schema returns correct feature version` or `Correct feature version value can be found in database`. wdyt?

templates/webapi/layouts/navbar.html.ep
```diff
@@ -42,10 +44,12 @@
...
         % if (current_user) {
             <li class="nav-item dropdown" id="user-action">
-                <a href="#" class="nav-link dropdown-toggle" data-bs-toggle="dropdown" role="button" aria-haspopup="true" aria-expanded="false" >Logged in as <%= current_user->name %></a>
+                <a href="#" class="nav-link dropdown-toggle" data-bs-toggle="dropdown" role="button"
+                   aria-haspopup="true" aria-expanded="false"
+                   title="Contains the Activity View and various configuration pages">Logged in as <%= current_user->name %></a>
                 <div class="dropdown-menu">
                   %= tag 'h3' => class => 'dropdown-header' => 'Operators Menu'
-                  %= link_to 'Activity View' => url_for('activity_view') => class => 'dropdown-item' => id => 'activity_view'
+                  %= link_to 'Activity View' => url_for('activity_view') => class => 'dropdown-item' => id => 'activity_view', title => 'Gives you an overview of your current jobs'
```
b10n1k, line 52: Maybe `Provides an overview of your current jobs` sounds better? Feel free to resolve this if you think it does not.

baierjan (commented):
Looks good. Just to be sure, it is expected this should **not** even break openQA-in-openQA tests because we are not testing the tour, right?

t/api/15-users.t
kalikiana: It is a post, though? And it checks the new version

kalikiana (dismissed):

okurz (changes requested):
I think you misunderstood our goals for https://progress.opensuse.org/issues/163004. I am against removing the tour. You can replace it with bootstrap hints as suggested. If I am reading your changes correctly then you are just putting help text in kinda hidden places. The ticket subject says "a simple bootstrap hint pointing to first steps"

Martchus:
So I don't misunderstand this again: What do you mean with "boostrap hint" and "first steps"?

Note that the tour so far just explained some UI elements that are in my opinion rather self-explanatory - especially with the newly added tooltips. It did *not* tell one first steps (like what we have in the "Getting started" documentation). The tour was also in general not about first steps but to notify users about new features (except that actually extending the tour was not really feasible in most cases and if we would have done it consequently we would now have a tour that is *way* to lengthy to click though from the beginning).

t/api/15-users.t
Martchus: I would not mention the database explicitly here. If we did this in similar cases we would lengthen a lot of lines in our code. Note that the test names are also generally supposed to explain the check on a higher level instead of just repeating what the code does.

I think using the past here makes it clear enough that the updating was done before the check and the check is just verifying what happened before (in this case the post).

Martchus:
> Just to be sure, it is expected this should not even break openQA-in-openQA tests because we are not testing the tour, right?

It may break these tests because we probably dismiss the tour at some point.

templates/webapi/layouts/navbar.html.ep
Martchus: After a quick research I think both verbs are fine and *give* and overview is probably even the more commonly used verb (with overview).

templates/webapi/layouts/navbar.html.ep
```diff
@@ -7,11 +7,13 @@
...
          <li class='nav-item' id="all_tests">
-            %= link_to 'All Tests' => url_for('tests') => class => 'nav-link'
+            %= link_to 'All Tests' => url_for('tests') => class => 'nav-link', title => 'Lists all tests grouped by state'
          </li>
 
          <li class="nav-item dropdown" id="job_groups">
-            <a href="#" class="nav-link dropdown-toggle" data-bs-toggle="dropdown" role="button" aria-haspopup="true" aria-expanded="false" data-submenu>Job Groups</a>
+            <a href="#" class="nav-link dropdown-toggle" data-bs-toggle="dropdown" role="button"
+               aria-haspopup="true" aria-expanded="false" data-submenu
+               title="Shows test results for a specific group (usually a specific distribution/product)">Job Groups</a>
```
okurz, line 16:
```suggestion
               title="Shows test results for a specific group (usually a specific product)">Job Groups</a>
```

I would avoid the word "distribution" as in the context of openQA it is mostly used to mean "test distribution", e.g. os-autoinst-distri-example, which does not usually correspond to job groups.

okurz:
> So I don't misunderstand this again: What do you mean with "boostrap hint" and "first steps"?

I actually meant a "bootstrap popover" to point to a single menu entry where newly logged in users probably want to start. But please read on for my reconsideration :)

> Note that the tour so far just explained some UI elements that are in my opinion rather self-explanatory - especially with the newly added tooltips. It did _not_ tell one first steps (like what we have in the "Getting started" documentation). The tour was also in general not about first steps but to notify users about new features (except that actually extending the tour was not really feasible in most cases and if we would have done it consequently we would now have a tour that is _way_ to lengthy to click though from the beginning).

Correct. Just from the point of view of a beginner one should be presented with unobstrusive hints where to get started. Previously we had the four steps "All tests area", "Job group menu", "User menu", "Activity View". I like how you moved some text already into help for the entries without needing those to be tour-popups. This shows how we are really liking good "title" attributes in many more places :)

I was thinking about maybe a single popup for new users where to start, like click on "All Tests" for a start. But then I figured that depending on the instance the index page or the job groups might be more likely what new users would look for. So I guess I am ok with your approach.

Just I would prefer to have the git commit message subject and PR title rephrased to be user-facing. How about "Use better help on menu items instead of obtrusive tour". And in the details you could mention the secondary benefit for us maintainers that by this we are dropping the problematic dependency :)

okurz (changes requested):
> So I don't misunderstand this again: What do you mean with "boostrap hint" and "first steps"?

I actually meant a "bootstrap popover" to point to a single menu entry where newly logged in users probably want to start. But please read on for my reconsideration :)

> Note that the tour so far just explained some UI elements that are in my opinion rather self-explanatory - especially with the newly added tooltips. It did _not_ tell one first steps (like what we have in the "Getting started" documentation). The tour was also in general not about first steps but to notify users about new features (except that actually extending the tour was not really feasible in most cases and if we would have done it consequently we would now have a tour that is _way_ to lengthy to click though from the beginning).

Correct. Just from the point of view of a beginner one should be presented with unobstrusive hints where to get started. Previously we had the four steps "All tests area", "Job group menu", "User menu", "Activity View". I like how you moved some text already into help for the entries without needing those to be tour-popups. This shows how we are really liking good "title" attributes in many more places :)

I was thinking about maybe a single popup for new users where to start, like click on "All Tests" for a start. But then I figured that depending on the instance the index page or the job groups might be more likely what new users would look for. So I guess I am ok with your approach.

Just I would prefer to have the git commit message subject and PR title rephrased to be user-facing. How about "Use better help on menu items instead of obtrusive tour". And in the details you could mention the secondary benefit for us maintainers that by this we are dropping the problematic dependency :)

Martchus:
After @okurz 's last comment I hope we have found an agreement. So please re-review my latest changes.

baierjan (approved):

okurz (approved):

PR #5862, opened 2024-08-15
Prevent git_clone to interfere with needle tasks

perlpunk:
Issue: https://progress.opensuse.org/issues/164883

### DIFF
--- a/lib/OpenQA/Task/Git/Clone.pm
+++ b/lib/OpenQA/Task/Git/Clone.pm
@@ -25,14 +25,17 @@ sub _git_clone_all ($job, $clones) {
     my $app = $job->app;
     my $job_id = $job->id;
 
-    # Prevent multiple git clone tasks for the same path to run in parallel
-    my @guards;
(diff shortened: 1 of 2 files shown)


lib/OpenQA/Task/Git/Clone.pm
```diff
@@ -25,14 +25,17 @@ sub _git_clone_all ($job, $clones) {
     my $app = $job->app;
     my $job_id = $job->id;
 
-    # Prevent multiple git clone tasks for the same path to run in parallel
-    my @guards;
     my $retry_delay = {delay => 30 + int(rand(10))};
+    # Don't interfere with any needle task
+    return $job->retry($retry_delay)
```
okurz, line 30: Each needle task shouldn't take that long. I would use much shorter delays here. How about 3s?

perlpunk: The delete_needles task can take longer I guess. But I reduced it to 5s now.

kalikiana (approved):

okurz (approved):

PR #5863, opened 2024-08-15
Refactor generic git functions into git module

r-richardson:
Refactor generic git related functions from OpenQA::Task::Git::Clone into the OpenQA::Git module for easier re-usability.
Related Ticket: [164886](https://progress.opensuse.org/issues/164886)

### DIFF
--- a/lib/OpenQA/Git.pm
+++ b/lib/OpenQA/Git.pm
@@ -1,12 +1,12 @@
 # Copyright 2019 SUSE LLC
 # SPDX-License-Identifier: GPL-2.0-or-later
+
 package OpenQA::Git;
 
 use Mojo::Base -base, -signatures;
 use Mojo::Util 'trim';
 use Cwd 'abs_path';
 use OpenQA::Utils qw(run_cmd_with_log_return_error);
-use Mojo::File;
 
 has 'app';
 has 'dir';
@@ -22,8 +22,23 @@ sub config ($self, $args = undef) {
     return $app->config->{'scm git'};
 }
 
-sub _prepare_git_command ($self, $args = undef) {
-    my $dir = $args->{dir} // $self->dir;
+sub _validate_attributes ($self) {
+    for my $mandatory_property (qw(app dir user)) {
+        die "no $mandatory_property specified" unless $self->$mandatory_property();
+    }
+}
+
+sub _run_cmd ($self, $args) {
+    run_cmd_with_log_return_error([$self->_prepare_git_command, @$args]);
+}
+
+sub _ssh_git_cmd ($self, $args) {
+    run_cmd_with_log_return_error(['env', 'GIT_SSH_COMMAND=ssh -oBatchMode=yes', $self->_prepare_git_command, @$args]);
+}
+
+sub _prepare_git_command ($self) {
+    my $dir = $self->dir;
+    die 'no valid directory was found during git preparation' unless $dir;
     if ($dir !~ /^\//) {
         my $absolute_path = abs_path($dir);
         $dir = $absolute_path if ($absolute_path);
@@ -32,36 +47,27 @@ sub _prepare_git_command ($self, $args = undef) {
 }
 
 sub _format_git_error ($self, $result, $error_message) {
-
-    my $path = $self->dir;
+    my $dir = $self->dir;
     if ($result->{stderr} or $result->{stdout}) {
-        $error_message .= "($path): " . $result->{stdout} . $result->{stderr};
+        $error_message .= " ($dir): " . $result->{stdout} . $result->{stderr};
     }
     return $error_message;
-}
-
-sub _validate_attributes ($self) {
-    for my $mandatory_property (qw(app dir user)) {
-        die "no $mandatory_property specified" unless $self->$mandatory_property();
-    }
 }
 
 sub set_to_latest_master ($self, $args = undef) {
     $self->_validate_attributes;
 
-    my @git = $self->_prepare_git_command($args);
-
     if (my $update_remote = $self->config->{update_remote}) {
-        my $res = run_cmd_with_log_return_error([@git, 'remote', 'update', $update_remote]);
+        my $res = $self->_run_cmd(['remote', 'update', $update_remote]);
         return $self->_format_git_error($res, 'Unable to fetch from origin master') unless $res->{status};
     }
 
     if (my $update_branch = $self->config->{update_branch}) {
         if ($self->config->{do_cleanup} eq 'yes') {
-            my $res = run_cmd_with_log_return_error([@git, 'reset', '--hard', 'HEAD']);
+            my $res = $self->_run_cmd(['reset', '--hard', 'HEAD']);
             return $self->_format_git_error($res, 'Unable to reset repository to HEAD') unless $res->{status};
         }
-        my $res = run_cmd_with_log_return_error([@git, 'rebase', $update_branch]);
+        my $res = $self->_run_cmd(['rebase', $update_branch]);
         return $self->_format_git_error($res, 'Unable to reset repository to origin/master') unless $res->{status};
     }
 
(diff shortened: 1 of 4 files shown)


okurz (commented):
If this is a prerequisite step for using the functions in other places then this looks good. If it stays like this then I am against at as functions that are only used once should be as close as possible where needed. We already discussed the possibility to keep the state within the object so that not every command needs to re-initialize or something

perlpunk:
We discussed a few things.
OpenQA::Git so far returns undef for success and the error message for a failure.
The new functions use `die`, and I like that more, as failures cannot be accidentally ignored, but must be catched with an `eval`.
We should decide on one or the other.

As another improvement I suggest another helper. Instead of
```perl
sub get_current_branch ($self) {
    my @git = $self->_prepare_git_command();
    my $r = run_cmd_with_log_return_error([@git, 'branch', '--show-current']);
```
we could do
```perl
sub _run_cmd ($self, $args) {
    my @git = $self->_prepare_git_command();
    return run_cmd_with_log_return_error([@git, @$args]);
}
...
sub get_current_branch ($self) {
    my $r = $self->_run_cmd(['branch', '--show-current']);
```

perlpunk:
>  If it stays like this then I am against at as functions that are only used once should be as close as possible where needed.

But the point of the ticket is to move all git functions to OpenQA::Git. if you are against this, we should have discussed this in the estimation.

t/14-grutasks.t
```diff
@@ -644,10 +645,12 @@ subtest 'handling dying GRU task' => sub {
     like $associated_job->reason, qr/^preparation failed: Thrown fail at/, 'reason of associated job set';
 };
 
+
+
```
okurz, line 649: please review your changes for such whitespace noise before submitting.

okurz (changes requested):

t/14-grutasks.t
kalikiana: Keep in mind that this is a draft

lib/OpenQA/Git.pm
```diff
@@ -22,15 +23,24 @@ sub config ($self, $args = undef) {
 }
 
 sub _prepare_git_command ($self, $args = undef) {
-    my $dir = $args->{dir} // $self->dir;
+    my $dir;
```
perlpunk, line 26: why is this change necessary?

r-richardson: I've been running into `t/16-utils-runcmd.t .. 1/? Not a HASH reference at /home/rrichardson/Code/oqa/openQA/lib/
OpenQA/Git.pm line 26.`earlier, which is why i added the check. I think however the `args` argument may not be needed at all here.

lib/OpenQA/Git.pm
```diff
@@ -43,23 +46,26 @@ sub _validate_attributes ($self) {
     }
 }
 
+sub _run_cmd ($self, $args = undef) {
+    my @git = $self->_prepare_git_command();
+    return run_cmd_with_log_return_error([@git, @$args]);
+}
```
Martchus, line 52: It makes no sense to assume `undef` as default for `$args` when the subsequent code would print a warning when `$args` is `undef`. I suppose creating the temporary array `@git` can also be avoided.

```suggestion
sub _run_cmd ($self, $args) { run_cmd_with_log_return_error([$self->_prepare_git_command, @$args]) }
```

lib/OpenQA/Git.pm
```diff
@@ -1,11 +1,12 @@
 # Copyright 2019 SUSE LLC
 # SPDX-License-Identifier: GPL-2.0-or-later
-
 package OpenQA::Git;
 
 use Mojo::Base -base, -signatures;
+use Mojo::Util 'trim';
 use Cwd 'abs_path';
 use OpenQA::Utils qw(run_cmd_with_log_return_error);
+use Mojo::File;
```
Martchus, line 9: This isn't actually used.

lib/OpenQA/Git.pm
```diff
@@ -68,30 +70,73 @@ sub set_to_latest_master ($self, $args = undef) {
...
-        return _format_git_error($res, 'Unable to push Git commit') unless $res->{status};
+        $res = $self->_run_cmd(['push']);
+        return $self->_format_git_error($res, 'Unable to push Git commit') unless $res->{status};
     }
 
     return undef;
 }
 
+sub get_current_branch ($self) {
+    my @git = $self->_prepare_git_command();
```
perlpunk, line 99: That's not used

lib/OpenQA/Git.pm
```diff
@@ -68,30 +70,73 @@ sub set_to_latest_master ($self, $args = undef) {
...
+sub get_remote_default_branch ($self, $url) {
+    my @git = $self->_prepare_git_command();
+    my $r = run_cmd_with_log_return_error(_ssh_git_cmd(['ls-remote', '--symref', $url, 'HEAD']));
+    die $self->_format_git_error($r, "Error detecting remote default branch name for '$url'")
+      unless $r->{status} && $r->{stdout} =~ /refs\/heads\/(\S+)\s+HEAD/;
+    return $1;
+}
+
+sub git_clone_url_to_path ($self, $url, $path) {
+    my @git = $self->_prepare_git_command();
```
perlpunk, line 118: Not used

lib/OpenQA/Git.pm
```diff
@@ -68,30 +70,73 @@ sub set_to_latest_master ($self, $args = undef) {
...
+    my $r = $self->_run_cmd(_ssh_git_cmd(['clone', $url, $path]));
+    die $self->_format_git_error($r, "Failed to clone $url into '$path'") unless $r->{status};
+}
+
+sub git_get_origin_url ($self) {
+    my @git = $self->_prepare_git_command();
```
perlpunk, line 124: Not used

lib/OpenQA/Git.pm
```diff
@@ -68,30 +70,73 @@ sub set_to_latest_master ($self, $args = undef) {
...
+    my $r = $self->_run_cmd(['remote', 'get-url', 'origin']);
+    die $self->_format_git_error($r, 'Failed to get origin url') unless $r->{status};
+    return Mojo::Util::trim($r->{stdout});
+}
+
+sub git_fetch ($self, $branch_arg) {
+    my @git = $self->_prepare_git_command();
```
perlpunk, line 131: Not used

lib/OpenQA/Git.pm
```diff
@@ -68,30 +70,73 @@ sub set_to_latest_master ($self, $args = undef) {
...
+    my $r = $self->_run_cmd(_ssh_git_cmd([@git, 'fetch', 'origin', $branch_arg]));
+    die $self->_format_git_error($r, "Failed to fetch from '$branch_arg'") unless $r->{status};
+}
+
+sub git_reset_hard ($self, $branch) {
+    my @git = $self->_prepare_git_command();
```
perlpunk, line 137: Not used

perlpunk (changes requested):

lib/OpenQA/Git.pm
```diff
@@ -1,9 +1,9 @@
 # Copyright 2019 SUSE LLC
 # SPDX-License-Identifier: GPL-2.0-or-later
-
```
okurz, line 3: unrelated whitespace change, please revert

lib/OpenQA/Git.pm
```diff
@@ -21,16 +21,18 @@ sub config ($self, $args = undef) {
     return $app->config->{'scm git'};
 }
 
-sub _prepare_git_command ($self, $args = undef) {
-    my $dir = $args->{dir} // $self->dir;
+sub _prepare_git_command ($self) {
+    my $dir = $self->dir;
+
```
okurz, line 26: unrelated whitespace change, please revert

lib/OpenQA/Git.pm
```diff
@@ -21,16 +21,18 @@ sub config ($self, $args = undef) {
...
     if ($dir !~ /^\//) {
         my $absolute_path = abs_path($dir);
         $dir = $absolute_path if ($absolute_path);
     }
     return ('git', '-C', $dir);
 }
 
-sub _format_git_error ($result, $error_message) {
+sub _format_git_error ($self, $result, $error_message) {
+    my $path = $self->dir;
```
okurz, line 35: I don't see this variable being used. What did you try to achieve here?

okurz (dismissed):

lib/OpenQA/Git.pm
perlpunk: The path was supposed to go into the error message, I saw it before, but at some point it was removed

lib/OpenQA/Git.pm
```diff
@@ -68,30 +70,68 @@ sub set_to_latest_master ($self, $args = undef) {
...
+
+sub get_remote_default_branch ($self, $url) {
+    my $r = run_cmd_with_log_return_error(_ssh_git_cmd(['ls-remote', '--symref', $url, 'HEAD']));
+    die $self->_format_git_error($r, "Error detecting remote default branch name for '$url'")
+      unless $r->{status} && $r->{stdout} =~ /refs\/heads\/(\S+)\s+HEAD/;
+    return $1;
+}
+
+sub git_clone_url_to_path ($self, $url, $path) {
+    my $r = $self->_run_cmd(_ssh_git_cmd(['clone', $url, $path]));
```
perlpunk, line 116:
```suggestion
sub git_clone_url_to_path ($self, $url) {
    my $r = $self->_run_cmd(_ssh_git_cmd(['clone', $url, $self->dir]));
```

lib/OpenQA/Task/Git/Clone.pm
```diff
@@ -13,14 +13,6 @@ sub register ($self, $app, @) {
...
 }
 
-
-# $clones is a hashref with paths as keys and urls to git repos as values.
-# The urls may also refer to a branch via the url fragment.
-# If no branch is set, the default branch of the remote (if target path doesn't exist yet)
-# or local checkout is used.
-# If the target path already exists, an update of the specified branch is performed,
-# if the url matches the origin remote.
-
```
perlpunk, line 23: Why did you remove those comments?

lib/OpenQA/Git.pm
r-richardson: thanks, this is a leftover

r-richardson: for context, we originally wanted to improve the format_git_error method output, however as this message has to match the git output exactly it makes no sense to add or remove any elements to the output

lib/OpenQA/Task/Git/Clone.pm
```diff
@@ -40,86 +39,49 @@ sub _git_clone_all ($job, $clones) {
 
     # iterate clones sorted by path length to ensure that a needle dir is always cloned after the corresponding casedir
     for my $path (sort { length($a) <=> length($b) } keys %$clones) {
+        my $git = OpenQA::Git->new(app => $app, dir => $path);
         my $url = $clones->{$path};
         die "Don't even think about putting '..' into '$path'." if $path =~ /\.\./;
-        eval { _git_clone($job, $ctx, $path, $url) };
+        eval { _git_clone($git, $job, $ctx, $path, $url) };
```
perlpunk, line 45: I think we can create the `$git` object in `_git_clone` because we only need it there and not in the loop here.
One less parameter to pass around.

lib/OpenQA/Task/Git/Clone.pm
r-richardson: restored

lib/OpenQA/Git.pm
```diff
@@ -21,16 +22,16 @@ sub config ($self, $args = undef) {
...
         $dir = $absolute_path if ($absolute_path);
     }
     return ('git', '-C', $dir);
 }
 
-sub _format_git_error ($result, $error_message) {
+sub _format_git_error ($self, $result, $error_message) {
     if ($result->{stderr} or $result->{stdout}) {
         $error_message .= ': ' . $result->{stdout} . $result->{stderr};
     }
```
perlpunk, line 37: Why don't we have the path in the error message anymore now?
We did this together, and I think if we get an error it would be good to know which path it was happening in.
The original code has the path in the error message.

lib/OpenQA/Git.pm
perlpunk: What do you mean with "this message has to match the git output exactly"?
Where is the problem?

lib/OpenQA/Git.pm
perlpunk: Just for information: there should not be an error message like this when `$args` is undef. Only if it is not undef and not a hash ref.
At that point you were probably accidentally passing some other argument.

lib/OpenQA/Git.pm
r-richardson: Ah, i misunderstood something before, in [16-utils-runcmd.t#L51](https://github.com/r-richardson/openQA/blob/301acc6ceb68654f17caee1de9438ad318504762/t/16-utils-runcmd.t#L51), can i just replace the `tempdir()` call with e.g. `./temp-dir` ? Because currently the paths contain random strings which make it impossible to prepare the string for comparison in the tests. Or should i only display the path conditionally instead ?

```
t/16-utils-runcmd.t .. 1/? 
    #   Failed test 'Git error message returned'
    #   at t/16-utils-runcmd.t line 56.
    #                   'Unable to commit via Git /tmp/_gssEGO8wE: fatal: not a git repository (or any parent up to mount point /)
    # Stopping at filesystem boundary (GIT_DISCOVERY_ACROSS_FILESYSTEM not set).'
    #     doesn't match '(?^u:^Unable to commit via Git: fatal: (N|n)ot a git repository \(or any)'
```

r-richardson: You're right. I misunderstood the way the git_format output is tested and thought it does not make sense here. Dir output is back now and tests adjusted

lib/OpenQA/Git.pm
r-richardson: Misunderstanding on my side, sorry. `$dir` is now being used within the output.

lib/OpenQA/Git.pm
perlpunk: We might even rename the method to just `git_clone_url`

perlpunk: I can't reproduce that error. If I do my suggested change, and I replace `$path` in the message one line below with `$self->dir` as well, and in the calling code I replace
```diff
- $git->git_clone_url_to_path($url, $path);
+ $git->git_clone_url_to_path($url);
```
the test passes

perlpunk: The only thing that might be something to improve: the `-C dir` is actually not necessary and confusing when we do the initial clone.
Also we don't need it when doing the `get_remote_default_branch`. So maybe the `_prepare_git_command` should not be done in those cases.

lib/OpenQA/Git.pm
perlpunk: Actually, this now adds `@git` to the arguments, but `_run_cmd` does call `_prepare_git_command` itself. I wonder how that works.

perlpunk: When I add debugging, I see that we call this command:
```
git -C /root/ env GIT_SSH_COMMAND=ssh -oBatchMode=yes git git -C /root/ fetch origin foobranch:foobranch
```
The test doesn't fail though, but that will fail in production I guess

perlpunk: It would probably be good if we compare the whole `@mocked_git_calls` to what we expect, not only selected items.

lib/OpenQA/Git.pm
```diff
@@ -68,30 +73,68 @@ sub set_to_latest_master ($self, $args = undef) {
...
     return undef;
 }
 
+sub get_current_branch ($self) {
+    my $r = $self->_run_cmd(['branch', '--show-current']);
+    die $self->_format_git_error($r, 'Error detecting current branch') unless $r->{status};
+    return Mojo::Util::trim($r->{stdout});
+}
+
+sub _ssh_git_cmd ($self, $args) {
```
perlpunk, line 107: This should be moved up to the other helpers.

lib/OpenQA/Git.pm
```diff
@@ -43,23 +45,26 @@ sub _validate_attributes ($self) {
     }
 }
 
+sub _run_cmd ($self, $args, $wrapper = []) {
```
perlpunk, line 48: Leftover
```suggestion
sub _run_cmd ($self, $args) {
```

perlpunk (commented):
Somehow git thinks I have started and never finished a review, so it's pending.
So I just try to submit it although I don't know what is still pending.

lib/OpenQA/Git.pm
```diff
@@ -21,18 +22,19 @@ sub config ($self, $args = undef) {
     return $app->config->{'scm git'};
 }
 
-sub _prepare_git_command ($self, $args = undef) {
-    my $dir = $args->{dir} // $self->dir;
+sub _prepare_git_command ($self) {
+    my $dir = $self->dir;
```
b10n1k, line 26: Maybe add a comment here as this function has some caveats and some things to improve.  for instance should it die or something if `$dir` is undef?

perlpunk (approved):

lib/OpenQA/Git.pm
r-richardson: I've moved the method right below `_run_cmd` and `_ssh_git_cmd`, in order to make it more obvious where it is called. Also added a check if `$dir` is set. I can also add a comment, do you have a suggestion ?

Martchus (approved):

lib/OpenQA/Git.pm
b10n1k: Lets move on. looks good

b10n1k (approved):

PR #5864, opened 2024-08-16
Update codecov.yml with current coverage in t/

okurz:


codecov.yml
```diff
@@ -13,7 +13,7 @@ coverage:
       default:
         target: 98
       tests:
-        target: 99.48
+        target: 99.6
```
kalikiana, line 16: Seems in line with https://app.codecov.io/gh/os-autoinst/openQA (just for reference as I was looking to see where the 99.6 was coming from)

kalikiana (approved):

perlpunk (approved):

PR #5871, opened 2024-08-20
Allow openqa-cli to export commands without an API

kalikiana:
The --export-command feature by design has no need for credentials.

### DIFF
--- a/lib/OpenQA/Script/CloneJob.pm
+++ b/lib/OpenQA/Script/CloneJob.pm
@@ -194,7 +194,7 @@ sub create_url_handler ($options) {
         apikey => $options->{apikey},
         apisecret => $options->{apisecret});
     die "API key/secret for '$options->{host}' missing. Checkout '$0 --help' for the config file syntax/lookup.\n"
-      unless $local->apikey && $local->apisecret;
+      if !$options->{'export-command'} && !($local->apikey && $local->apisecret);
 
     my $remote_url = OpenQA::Client::url_from_host($options->{from});
     $remote_url->path('/api/v1/jobs');


lib/OpenQA/Script/CloneJob.pm
```diff
@@ -193,8 +193,10 @@ sub create_url_handler ($options) {
         api => $local_url->host,
         apikey => $options->{apikey},
         apisecret => $options->{apisecret});
-    die "API key/secret for '$options->{host}' missing. Checkout '$0 --help' for the config file syntax/lookup.\n"
-      unless $local->apikey && $local->apisecret;
+    if (!$options->{'export-command'}) {
+        die "API key/secret for '$options->{host}' missing. Checkout '$0 --help' for the config file syntax/lookup.\n"
+          unless $local->apikey && $local->apisecret;
+    }
```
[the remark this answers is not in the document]
kalikiana: They would not be part of the command one way or the other 🤔 Or maybe I misunderstand your question

lib/OpenQA/Script/CloneJob.pm
```diff
@@ -193,8 +193,10 @@ sub create_url_handler ($options) {
         api => $local_url->host,
         apikey => $options->{apikey},
         apisecret => $options->{apisecret});
-    die "API key/secret for '$options->{host}' missing. Checkout '$0 --help' for the config file syntax/lookup.\n"
-      unless $local->apikey && $local->apisecret;
+    if (!$options->{'export-command'}) {
```
b10n1k, line 196: `unless ($options->{'export-command'})` looks more suitable here. wdyt?

lib/OpenQA/Script/CloneJob.pm
Martchus, line 199:
```suggestion
    die "API key/secret for '$options->{host}' missing. Checkout '$0 --help' for the config file syntax/lookup.\n"
          unless $options->{'export-command'} || ($local->apikey && $local->apisecret);
```

Martchus (commented):
I'd avoid the additional level of nesting.

lib/OpenQA/Script/CloneJob.pm
perlpunk: I personally never use `unless` with multiple conditions because it makes my brain sick.

lib/OpenQA/Script/CloneJob.pm
kalikiana: I thought it might become less readable since there's already an unless here... no strong opinion, though

kalikiana: https://github.com/os-autoinst/openQA/pull/5871#discussion_r1724980157

PR #5872, opened 2024-08-20
ci: Add helper for reporting OBS url

perlpunk:
Sometimes the OBS webhook fails and we need to find out the URL to the broken OBS project manually.
This will write a comment onthe PR with the URL.

Issue: https://progress.opensuse.org/issues/165144

Demo: https://github.com/perlpunk/os-autoinst/pull/15#issuecomment-2299076322

### DIFF
--- a/.github/workflows/obs-helper.yaml
+++ b/.github/workflows/obs-helper.yaml
@@ -1,0 +1,22 @@
+---
+# yamllint disable rule:line-length
+name: OBS Helper
+
+on: [pull_request_target]
+
+jobs:
+  obs-url:
+    name: Report OBS Url
+    runs-on: ubuntu-latest
+    permissions:
+      issues: write
+    steps:
+      - uses: actions/checkout@v4
+      - name: Comment
+        # Only run once on pull request creation
+        if: github.run_number == 1
+        env:
+          GH_TOKEN: ${{ github.token }}
+        run: |
+          gh issue comment ${{ github.event.pull_request.number }} \
+            --body "You might want to check the OBS project created for this PR: https://build.opensuse.org/project/show/devel:openQA:GitHub:os-autoinst:openQA:${{ github.event.pull_request.number }}"$'\n'"Here you can find all pull requests: https://build.opensuse.org/project/subprojects/devel:openQA:GitHub"$'\n'"_(This is just a little helper in case the OBS workflow fails and does not report the URL yet.)_"


perlpunk:
Hm, this does not work:
```
GraphQL: Resource not accessible by integration (addComment)
```

perlpunk:
I changed it to `pull_request_target` now. We can't see if it works before merging then, though. But the permissions should be fine then.

.github/workflows/obs-helper.yaml
```diff
@@ -0,0 +1,22 @@
...
+  obs-url:
+    name: Report OBS Url
+    runs-on: ubuntu-latest
+    permissions:
+      issues: write
+    steps:
+      - uses: actions/checkout@v4
+      - name: Comment
+        # Only run once on pull request creation
+        if: github.run_number == 1
```
kalikiana, line 17: Could we also make this depend on whether the OBS hook failed?

kalikiana: Although I'm not sure why we can't fix the original OBS check to always show the URL. It **seems** like that should be trivial as the URL is predictable.

perlpunk: > Could we also make this depend on whether the OBS hook failed?

I wasn't successful to query webhook deliveries via command line so far.
https://docs.github.com/en/rest/apps/webhooks?apiVersion=2022-11-28

perlpunk: > Although I'm not sure why we can't fix the original OBS check to always show the URL. It **seems** like that should be trivial as the URL is predictable.

Wanna make a PR? https://github.com/openSUSE/open-build-service/blob/7d122807c183ef66a4b8ad92f03de98ca06a2a2b/src/api/app/services/github_status_reporter.rb#L49
:)

perlpunk: But I guess it might not be trivial. I think it's possible to configure several different workflows in such a file, so there is not only one URL with the results.
And as far as I know, already reporting the various states as pending from the beginning is not possible currently due to a limitation in OBS.

b10n1k:
I tried to open the link from https://github.com/perlpunk/os-autoinst/pull/15#issuecomment-2299076322 and it redirects me to the github.com. If I copy the link to the url bar I have a not found project `Project not found: devel:openQA:GitHub:os-autoinst:os-autoinst:15` is the link correct or...

perlpunk:
> I tried to open the link from [perlpunk/os-autoinst#15 (comment)](https://github.com/perlpunk/os-autoinst/pull/15#issuecomment-2299076322) and it redirects me to the github.com. If I copy the link to the url bar I have a not found project `Project not found: devel:openQA:GitHub:os-autoinst:os-autoinst:15` is the link correct or...

That link is not supposed to exist.
The point is how the git comment looks like.

kalikiana (approved):

perlpunk:
#5879 for a better alternative using the status API

PR #5873, opened 2024-08-20
Add line numbers with anchoring in log viewer

Martchus:
Related ticket: https://progress.opensuse.org/issues/165402

---

screenshot_20240820_173655

---

screenshot_20240820_173927

### DIFF
--- a/assets/javascripts/test_result.js
+++ b/assets/javascripts/test_result.js
@@ -358,7 +358,7 @@ function handleKeyDownOnTestDetails(e) {
 function setPageHashAccordingToCurrentTab(tabNameOrHash, replace) {
   // don't mess with #step hashes within details tab
   const currentHash = window.location.hash;
-  if (tabNameOrHash === 'details' && currentHash.search('#step/') === 0) {
+  if (tabNameOrHash === 'details' && (currentHash.startsWith('#step/') || currentHash.startsWith('#line-'))) {
     return;
   }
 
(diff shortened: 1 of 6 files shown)


perlpunk:
Can you change it so it does not select the line numbers when selecting the text?

Martchus:
Note that I changed the filter to use query parameters. I mainly did this so it doesn't interfere with the normal anchoring behavior of the web browser that is now utilized to jump to log lines. Otherwise I would have to parse the fragment in JavaScript and implement the jumping manually. However, I don't think this is a bad change. In fact, it is more consistent with other places where we provide filtering parameters. (And I could actually reduce the required code by simply reusing helper functions we already have.)

Martchus:
> Can you change it so it does not select the line numbers when selecting the text?

It does *not* select the line numbers when selecting text. It only selects the line numbers when clicking on one of the newly added line numbers on the left explicitly.

perlpunk:
A minor bug:
* Select some line number
* Search for something else that doesn't appear in that line, so the selected line disappears as expected. Don't unfocus the search field
* Then select one of the filtered line numbers. It does not work. Only when you click it the second time

When unfocusing the search field first, it works.

perlpunk:
Is it possible to select multiple line numbers?

perlpunk:
> It does _not_ select the line numbers when selecting text.

It does select them for me.

perlpunk:
log-view-line-numbers
Same behaviour in chromium and firefox.

Martchus:
> A minor bug:

I'll fix that tomorrow.

> Is it possible to select multiple line numbers?

No. I guess implementing that would be a bit more work.

> It does select them for me.

I misunderstood. I thought you were talking about the hash/highlighting being updated. I'll change it tomorrow so line numbers are not selected :-)

assets/javascripts/test_result.js
```diff
@@ -461,13 +461,16 @@ function activateTabAccordingToHashChange() {
...
+    link = $("[href='#details']");
+    tabName = 'details';
+  } else if (hash.startsWith('#step/')) {
     setCurrentPreviewFromStepLinkIfPossible(link);
     link = $("[href='#details']");
     tabName = 'details';
     // note: It is not a problem if the details haven't been loaded so far. Once the details become available the hash
     //       is checked again and the exact step preview will be shown.
-  } else if (hash.search('#comment-') === 0) {
+  } else if (hash.startsWith('#comment-')) {
```
b10n1k, line 473: Can this merged somehow?

Martchus:
The latest version should fix the mentioned problems. I haven't implemented anchoring a range of lines because that would be a bit more involved.

perlpunk (approved):

kalikiana (approved):

b10n1k (approved):

PR #5874, opened 2024-08-20
Remove git fetch with destination refspec feature

b10n1k:
As per git documentation the refspec can formed as <src>:<dst> but dst cannot
use arbitrary expressions (SHA) but an actual ref must be named. Its usage
in the `_git_clone` mistakely adds a SHA ref as destination creating a branch
with its name, which is not supported by git.
   
As far as now the feature is not used in OpenQA so it can be safely removed.

https://progress.opensuse.org/issues/164937

### DIFF
--- a/lib/OpenQA/Task/Git/Clone.pm
+++ b/lib/OpenQA/Task/Git/Clone.pm
@@ -115,15 +115,9 @@ sub _git_clone ($job, $ctx, $path, $url) {
     }
 
     my $current_branch = _get_current_branch($path);
-    if ($requested_branch eq $current_branch) {
-        # updating default branch (including checkout)
-        _git_fetch($path, $requested_branch);
-        _git_reset_hard($path, $requested_branch);
-    }
-    else {
-        # updating local branch to latest remote branch version
-        _git_fetch($path, "$requested_branch:$requested_branch");
-    }
+    # updating default branch (including checkout)
+    _git_fetch($path, $requested_branch);
+    _git_reset_hard($path, $requested_branch) if ($requested_branch eq $current_branch);
 }
 
 1;
--- a/t/14-grutasks.t
+++ b/t/14-grutasks.t
@@ -671,7 +671,7 @@ subtest 'git clone' => sub {
     is $res->{result}, 'Job successfully executed', 'minion job result indicates success';
     like $mocked_git_calls[3], qr'git -C /etc/ fetch origin master', 'fetch origin master for /etc/';
     like $mocked_git_calls[4], qr'reset --hard origin/master', 'reset origin/master for /etc/';
-    like $mocked_git_calls[8], qr'git -C /root/ fetch origin foobranch:foobranch', 'fetch non-default branch';
+    like $mocked_git_calls[8], qr'git -C /root/ fetch origin foobranch', 'fetch non-default branch';
     like $mocked_git_calls[10], qr'git clone http://localhost/bar.git /this_directory_does_not_exist/',
       'clone to /this_directory_does_not_exist/';
 


perlpunk (commented):
We talked about it earlier today and decided that we don't need the local branches at all.

lib/OpenQA/Task/Git/Clone.pm
```diff
@@ -119,7 +119,8 @@ sub _git_clone ($job, $ctx, $path, $url) {
     }
     else {
         # updating local branch to latest remote branch version
-        _git_fetch($path, "$requested_branch:$requested_branch");
+        my $dst = 'clone_' . substr($requested_branch, 0, 7);
```
perlpunk, line 122: This could result in duplicates if branch names have common prefixes

perlpunk (changes requested):
See comments

lib/OpenQA/Task/Git/Clone.pm
```diff
@@ -119,7 +119,8 @@ sub _git_clone ($job, $ctx, $path, $url) {
...
+        _git_fetch($path, "$requested_branch:$dst");
```
perlpunk, line 123: This should do it:
```suggestion
        _git_fetch($path, $requested_branch);
```

perlpunk: Why did you ignore this suggestion?

t/14-grutasks.t
```diff
@@ -671,7 +671,6 @@ subtest 'git clone' => sub {
     is $res->{result}, 'Job successfully executed', 'minion job result indicates success';
     like $mocked_git_calls[3], qr'git -C /etc/ fetch origin master', 'fetch origin master for /etc/';
     like $mocked_git_calls[4], qr'reset --hard origin/master', 'reset origin/master for /etc/';
-    like $mocked_git_calls[8], qr'git -C /root/ fetch origin foobranch:foobranch', 'fetch non-default branch';
```
perlpunk, line 674: Why did you remove this test? Don't we still want to test that command? Just that it changed?

b10n1k: We do not use any more. there is not such call with refspec in this format. Unless you want to test that it fails!

perlpunk: The call is still there. I can see it if I checkout your PR.
It's just not `git -C /root/ fetch origin foobranch:foobranch` anymore, but `git -C /root/ fetch origin foobranch`.

lib/OpenQA/Task/Git/Clone.pm
perlpunk: Ok, I think your change is actually the same as my suggestion, just that you shortened it because the fetch is now the same call for both cases.

t/14-grutasks.t
b10n1k: bring it back on f8e3adb2b. I had remove it as I dont think is much different from the test 2 lines above.

perlpunk: Sure, but it's a different use case.
And as we saw in [the other pull request](https://github.com/os-autoinst/openQA/pull/5863#discussion_r1725013852), we actually would have produced an error in production because we didn't test all elements in that array.
Removing an existing test is rarely a good idea.

perlpunk (commented):
Will review again when conflicts are resolved

perlpunk (approved):

perlpunk:
Added not-ready label to avoid having trouble reverting #5880

perlpunk:
Sorry, we had to revert #5880 so I guess you need to resolve conflicts again...

Martchus:
Or wait until we can reinstate the refactoring. Otherwise it'll be a back and forth which is probably not worth it.

perlpunk:
The refactoring will include some more changes which will cause more conflicts with this PR, so conflicts will appear in both cases possiibly

perlpunk (approved):

PR #5876, opened 2024-08-21
Redeliver failed webhooks to account for temporary errors

kalikiana:
See: https://progress.opensuse.org/issues/165144

### DIFF
--- a/.github/workflows/redeliver.yaml
+++ b/.github/workflows/redeliver.yaml
@@ -1,0 +1,38 @@
+---
+name: Redeliver failed webhooks
+
+# This workflow runs every 6 hours or when manually triggered.
+on:
+  schedule:
+    - cron: '20 */6 * * *'
+  workflow_dispatch:
+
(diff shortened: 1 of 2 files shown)


Martchus:
I'm not sure whether that's the best idea. The context for which the hooks have been triggered might have changed.

.github/workflows/redeliver.yaml
```diff
@@ -0,0 +1,36 @@
...
+      - uses: actions/setup-node@v4
+        with:
+          node-version: '18.x'
+      - run: npm install octokit
+      - name: Run script
+        env:
+          TOKEN: ${{ secrets.GH_TOKEN_FOR_ACTIONS }}
+          REPO_OWNER: 'os-autoinst'
+          REPO_NAME: 'openQA'
+          HOOK_ID: 'https://build.opensuse.org/trigger/workflow?id=5857'
```
perlpunk, line 29: I think we need the id of the hook here, not the url.

kalikiana:
This has meanwhile been addressed upstream. See https://github.com/openSUSE/open-build-service/pull/16834

PR #5877, opened 2024-08-21
Redeliver failed webhooks to account for temporary errors

kalikiana:
See: https://progress.opensuse.org/issues/165144

### DIFF
--- a/.github/workflows/redeliver.yaml
+++ b/.github/workflows/redeliver.yaml
@@ -1,0 +1,28 @@
+---
+name: Redeliver failed webhooks
+
+# This workflow runs every 6 hours or when manually triggered.
+on:
+  schedule:
+    - cron: '20 */6 * * *'
+  workflow_dispatch:
+
+# This workflow will use the built in `GITHUB_TOKEN` to check out the repository contents. This grants `GITHUB_TOKEN` permission to do that.
+permissions:
+  contents: read
+
+jobs:
+  redeliver:
+  - name: Redeliver failed webhooks
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v4
+    # Using a commit as the action isn't currently versioned
+    - uses: gateixeira/retrigger-webhook-action@9383944
+      env:
+        token: ${{ secrets.GH_TOKEN_FOR_ACTIONS }}
+        owner: ${{ github.repository_owner }}
+        repo: ${{ github.event.repository.name }}
+        # This variable will be updated with the last timestamp
+        # to avoid running the same hooks again
+        last_redelivery_variable_name: 'LAST_WEBHOOK_REDELIVERY'


kalikiana:
This is an alternative to #5876 which I found after preparing the draft based on the boilerplate from official GitHub docs, and probably what we would prefer as we don't need to maintain this in the repo **plus** it already supports multiple hooks out of the box.

.github/workflows/redeliver.yaml
```diff
@@ -0,0 +1,25 @@
...
+permissions:
+  contents: read
+
+jobs:
+  redeliver:
+  - name: Redeliver failed webhooks
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v4
+    - uses: gateixeira/retrigger-webhook-action@main
```
perlpunk, line 20: This action doesn't seem to be an official github action.
Also I think we should never use the `main` version of a third party action we don't control.
And also the usage
https://github.com/gateixeira/retrigger-webhook-action
seems to be totally different. It should not be a scheduled workflow on its own. It says "Add the following snippet to an existing workflow file".
And I think this should really be tested somehow, maybe you can add a dummy workflow in your fork.

.github/workflows/redeliver.yaml
```diff
@@ -0,0 +1,25 @@
...
+      env:
+        token: ${{ secrets.GH_TOKEN_FOR_ACTIONS }}
+        owner: ${{ github.repository_owner }}
+        repo: ${{ github.event.repository.name }}
+        last_redelivery_variable_name: 'LAST_WEBHOOK_REDELIVERY'
```
perlpunk, line 25: Where does that come from? I don't really understand the documentation of the action.

perlpunk: Also, which webhook should it attempt to redeliver? we have more than one and probably need to give it an id, e.g. like documented `webhook_id: 'YOUR_HOOK_ID'`

.github/workflows/redeliver.yaml
kalikiana: The alternative is #5876 where we implement it from scratch.

I agree on adding a version, though.

.github/workflows/redeliver.yaml
kalikiana: Ah. I forgot to add the comment here. This is to store the last timestamp in a variable.

.github/workflows/redeliver.yaml
perlpunk: > What existing workflow would this be if not this one? I read **existing** as any workflow where this action is useful, and the schedule is what GitHub docs recommend.

I don't know. I don't find the documentation of that action very useful and also not very reassuring :)

kalikiana:
> Did you have a look at the code this executes? I haven't and thus cannot really approve this PR. As mentioned in the other PR we should probably test what implications this can have for the various webhooks we have. (We don't want things to happen twice or e.g. re-trigger CI checks based on an old state.)

https://github.com/gateixeira/retrigger-webhook-action/blob/f606bae35fe73db18b10fe1ce6b5e52a734b6062/src/main.ts

It is based on [the officially documented boilerplate](https://docs.github.com/en/webhooks/using-webhooks/automatically-redelivering-failed-deliveries-for-a-repository-webhook) and uses the same pattern to avoid retriggering hooks more than once.

.github/workflows/redeliver.yaml
perlpunk: I never said that I don't like the idea.

And for a proof of concept we still need the proof :)

okurz:
With https://github.com/openSUSE/open-build-service/pull/16834 this shouldn't be necessary anymore, right?

kalikiana:
> With [openSUSE/open-build-service#16834](https://github.com/openSUSE/open-build-service/pull/16834) this shouldn't be necessary anymore, right?

Ack.

PR #5879, opened 2024-08-21
Test status API

perlpunk:


perlpunk:
You can see the new check at the bottom of the list of checks as "continuous-integration/obs-helper — Check OBS SCM URL".

kalikiana (approved):

Martchus (approved):

.mergify.yml
```diff
@@ -13,7 +13,7 @@ pull_request_rules:
           # "unresolvable" is not reported in general:
           # https://trello.com/c/0N3jHq5M/2257-report-back-to-scm-when-build-results-arent-failed-and-succeeded
           # So we need to require the number of tests explicitly:
-          - "#check-success>=35"
+          - "#check-success>=50"
```
Martchus, line 16: I suppose 45 would have been the correct number. At least that's how many check I got on https://github.com/os-autoinst/openQA/pull/5889 and that PR looks complete.

perlpunk: Yeah, feel free to set it to 45 in your PR.
In this PR you can see 52 checks actually, but I'm not sure why it's different.

perlpunk: In this PR I see some checks twice, I guess that explains it. Maybe that's the case when you create a PR on the origin repo.

PR #5883, opened 2024-08-22
Fix AppArmor rules for git

baierjan:
The change was tested within openQA-in-openQA test, the issue should be visible in https://openqa.opensuse.org/tests/4423875


perlpunk:
How can it be explained that the openqa-in-openqa test only failed sporadically?

baierjan:
You mean in context of https://progress.opensuse.org/issues/165408? For me in my setup, I can reproduce it every time. Maybe we have workers where AppArmor is not enforced (succeeds)  and where it is enforced (fails)?

baierjan:
> Maybe we have workers where AppArmor is not enforced (succeeds) and where it is enforced (fails)?

Although that does not make sense as the AppArmor needs to be enforced inside to fail at that stage...

perlpunk:
Yes, I mean https://progress.opensuse.org/issues/165408
I still don't understand it.
Maybe I'm confusing what's running on the o3 worker and what's running inside of qemu in the openqa-in-openqa test.

So in https://github.com/os-autoinst/os-autoinst-distri-openQA/pull/199 you are enabling apparmor inside of the test.

Can an apparmor installation on an o3 worker influence whether the service is enabled inside the test?

perlpunk:
And I thought we do actually have apparmor enabled already:
https://openqa.opensuse.org/tests/4422620/logfile?filename=test_running-journal.log.txt#filter=apparmor
```
Aug 22 06:54:31 susetest apparmor.systemd[690]: Restarting AppArmor
Aug 22 06:54:31 susetest apparmor.systemd[690]: Reloading AppArmor profiles
```

baierjan:
AppArmor in the inner test will be enabled after https://github.com/os-autoinst/os-autoinst-distri-openQA/pull/199. The AppArmor itself is installed and enabled from the boot, but after the profile is installed (which happens during the test itself) the newly added profiles are not loaded as AppArmor needs reload and the binaries needs to be restarted (there might be maybe the room for some sporadic behaviour?).

perlpunk:
right, I also thought about it being an issue with timing...

kalikiana (approved):

perlpunk (approved):

baierjan:
As mergify already merged the other one, I will proceed here; otherwise we will see https://openqa.opensuse.org/tests/4423875#step/test_running/8 more often.

PR #5884, opened 2024-08-22
Improve git commands in git_clone job

perlpunk:
I rearranged and got rid of some calls.

In case of the initial clone, we don't actually need to do a fetch afterwards, because an initial clone without any restrictions fetches all refs.
Also, we don't need to retrieve the default remote.

For an update we return early if the remote url doesn't match. And then only if there is no url fragment with a ref, we fetch the default remote.

Finally, I removed the second die in case we couldn't get a default remote. _get_remote_default_branch already dies, the outer die would have never been reached. This case is now also tested in a new subtest.

(And I improved the regex to avoid having to escape slashes: `\/ \/`)

The full list of commands is now tested, and better readable what commands are called for which use case.

Related Issue: https://progress.opensuse.org/issues/164886

### DIFF
--- a/lib/OpenQA/Task/Git/Clone.pm
+++ b/lib/OpenQA/Task/Git/Clone.pm
@@ -65,8 +65,8 @@ sub _ssh_git_cmd ($git_args) {
 
 sub _get_remote_default_branch ($url) {
     my $r = run_cmd_with_log_return_error(_ssh_git_cmd(['ls-remote', '--symref', $url, 'HEAD']));
-    die "Error detecting remote default branch name for '$url': $r->{stderr}"
-      unless $r->{status} && $r->{stdout} =~ /refs\/heads\/(\S+)\s+HEAD/;
+    die "Error detecting remote default branch name for '$url': $r->{stdout} $r->{stderr}"
+      unless $r->{status} && $r->{stdout} =~ m{refs/heads/(\S+)\s+HEAD};
     return $1;
 }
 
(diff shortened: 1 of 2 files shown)


perlpunk:
Ah, the OBS helper workflow still needs to run as pull_request_target.
https://github.com/os-autoinst/openQA/actions/runs/10515519332/job/29135748986
Otherwise it can't read the token.
Will prepare a PR.

t/14-grutasks.t
```diff
@@ -647,12 +647,18 @@ subtest 'handling dying GRU task' => sub {
...
+        '/etc/' => 'http://localhost/foo.git',
+        '/root/' => 'http://localhost/foo.git#foobranch',
+        '/this_directory_does_not_exist/' => 'http://localhost/bar.git',
+    };
     $openqa_utils->redefine(
         run_cmd_with_log_return_error => sub ($cmd) {
             my $stdout = '';
             $stdout = 'ref: refs/heads/master	HEAD' if $cmd->[3] eq 'ls-remote';
-            $stdout = 'http://localhost/foo.git' if $cmd->[4] eq 'get-url';
+            $stdout = $clone_dirs->{$cmd->[2]} =~ s/#.*//r if $cmd->[4] eq 'get-url';
```
kalikiana, line 659: How about making this a little more explicit to avoid having to trace what the numbers mean?

```suggestion
            my dir = $cmd->[2];
            my $action = $cmd->[4];
            $stdout = $clone_dirs->{$dir} =~ s/#.*//r if $action eq 'get-url';
```

perlpunk: One problem is that the dir is not always item number 2, and action is not always number 4. When we call the `env SSH... git ...` wrapper the numbers are different.

kalikiana: But then how does this work in the first place? The code applies to all of the commands, no?

kalikiana: Maybe it would make sense to split the hash into something like label/command/action?

Martchus (commented):
`s/ efault/ default/`
`s/alraedy/already/`

t/14-grutasks.t
perlpunk: > But then how does this work in the first place? The code applies to all of the commands, no?

Yep.
When the action is `ls-remote`, it's on position 3, and for `get-url` it's on position 4.

perlpunk: > Maybe it would make sense to split the hash into something like label/command/action?

I can't imagine right now which hash you mean and how it would look like? Maybe a short example?

lib/OpenQA/Task/Git/Clone.pm
```diff
@@ -96,24 +96,24 @@ sub _git_clone ($job, $ctx, $path, $url) {
...
 
-    if (!-d $path) {
+    unless (-d $path) {
         _git_clone_url_to_path($url, $path);
-        # update local branch to latest remote branch version
-        _git_fetch($path, "$requested_branch:$requested_branch")
-          if ($requested_branch ne $remote_default_branch);
+        # Fetches all refs, we are done
+        return;
     }
```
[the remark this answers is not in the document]
perlpunk: I used a different wording, but I hope it's ok now. For me `refs` simply means branches (and tags and shas).

t/14-grutasks.t
perlpunk: I changed it a bit. I hope it's clearer now.

kalikiana: Something like this:

```
    my $expected_calls = [
        # /etc/
        [action => 'get-url', cmd => 'git -C /etc/ remote get-url origin'],
        [action => 'ls-remote', cmd => 'env GIT_SSH_COMMAND="ssh -oBatchMode=yes" git ls-remote --symref http://localhost/foo.git HEAD'],
```

Martchus (approved):

kalikiana (approved):

PR #5886, opened 2024-08-23
Ensure full test coverage of test code

Martchus:
* Remove dead code
* Mark lines only executed in case tests are failing or in case tests are executed with verbose logging as uncoverable
* Mark lines executed in sub processes where coverage cannot be collected as uncoverable
* Fix attempts to mark lines as uncoverable
* See https://progress.opensuse.org/issues/165396


kalikiana:
```
Run gh api \
gh: To use GitHub CLI in a GitHub Actions workflow, set the GH_TOKEN environment variable. Example:
  env:
    GH_TOKEN: ${{ github.token }}
Error: Process completed with exit code 4.
```

perlpunk:
Fix for the OBS helper check: #5888

t/ui/01-list.t
```diff
@@ -256,12 +256,12 @@ subtest 'available comments shown' => sub {
             $driver->find_element('#job_99963 .fa-comment')->get_attribute('title'),
             '2 comments available',
             'available comments shown for running jobs'
-        );
+        );    # uncoverable statement
```
perlpunk, line 259: Hm, why are we unconditionally skipping tests here anyway?

perlpunk:
fullstack failure: https://progress.opensuse.org/issues/165764

t/ui/01-list.t
kalikiana: Good question. It looks like it's never run and there's no way to enable it?

t/api/14-plugin_obs_rsync_async.t
```diff
@@ -19,10 +19,10 @@ $t->app($app);
...
 sub start_gru {
     start sub {
-        note('starting gru');
-        $0 = 'openqa-gru';
-        $ENV{MOJO_MODE} = 'test';
-        Mojolicious::Commands->start_app('OpenQA::WebAPI', 'gru', 'run', '-m', 'test');
+        note('starting gru');    # uncoverable statement
+        $0 = 'openqa-gru';    # uncoverable statement
+        $ENV{MOJO_MODE} = 'test';    # uncoverable statement
+        Mojolicious::Commands->start_app('OpenQA::WebAPI', 'gru', 'run', '-m', 'test');    # uncoverable statement
```
kalikiana, line 25: Where is this used? I see `my $gru = start_gru();` below so I would think this is always covered/executed

t/ui/01-list.t
Martchus: The comment says "to imporove performance". I can try enabling those tests.

t/api/14-plugin_obs_rsync_async.t
kalikiana: Ah. I guess that makes sense. Maybe add a comment at the top to indicate that.

t/ui/01-list.t
Martchus: I changed the checks to be actually executed and to verify the current behavior.

t/api/14-plugin_obs_rsync_async.t
Martchus: This is only one of many of such situations. I would not add an extra comment about it.

Martchus: Added the comment nevertheless. At this point sprinkling comments doesn't hurt anymore anyway.

perlpunk (approved):

kalikiana (approved):

perlpunk:
Now it complains about: https://app.codecov.io/gh/os-autoinst/openQA/pull/5886/blob/t/lib/OpenQA/Test/Utils.pm#L209

PR #5887, opened 2024-08-23
Avoid having RPM build depend on packages only needed for style checks

Martchus:
The RPM builds do not execute style checks (also when `--with-tests` is used) so dependencies like ShellCheck should not be required for the build.

This avoids pulling in a lot of transitive dependencies into the build and thus also greatly reduces the number of packages we need submit to SLE to build openQA on/for SLE.

Related ticket: https://progress.opensuse.org/issues/165683

### DIFF
--- a/cpanfile
+++ b/cpanfile
@@ -98,8 +98,6 @@ requires 'warnings';
 
 on 'test' => sub {
     requires 'App::cpanminus';
-    requires 'Perl::Critic';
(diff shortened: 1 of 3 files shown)


kalikiana:
[poo#165764](https://progress.opensuse.org/issues/165764)

kalikiana (approved):

perlpunk (approved):

PR #5889, opened 2024-08-23
Fix invalid HTML leading to fullstack test failures

Martchus:
Remove use of `code` tag within `select`/`option` to avoid new Chromium warning `A code tag was parsed inside of a <select> which was not inserted into the document.` that interferes with the fullstack test.

Related ticket: https://progress.opensuse.org/issues/165764


baierjan (approved):

perlpunk (approved):
Ah, there it was. I looked on the wrong page and couldn't find a code in a select...

PR #5891, opened 2024-08-23
ci: Decrease the number of required checks

perlpunk:
50 was a wrong guess. I saw 52 checks in my PR, but that might have been because it was created on the origin repo, so we see push and pull_request checks.


perlpunk:
@Martchus was a few seconds faster: #5890

PR #5894, opened 2024-08-26
openqa-cli: Avoid reading unexpectedly from STDIN

Martchus:
* Avoid skipping the rest of a script when that script is invoked via `sh < …` and internally invokes openqa-cli (which in this case must not consume all remaining lines in that script)
* Read only from STDIN if explicitly requested via `--data-file -` in consistency with curl (which also allows using `-` where a file can be specified to use STDIN)
* See https://progress.opensuse.org/issues/160820

### DIFF
--- a/lib/OpenQA/CLI/api.pm
+++ b/lib/OpenQA/CLI/api.pm
@@ -16,14 +16,11 @@ has usage => sub {
 };
 
 sub command ($self, @args) {
-
-    my $data = $self->data_from_stdin;
-
     die $self->usage
       unless getopt \@args,
       'a|header=s' => \my @headers,
       'D|data-file=s' => \my $data_file,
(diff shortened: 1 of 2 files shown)


b10n1k (approved):

kalikiana (approved):

lib/OpenQA/CLI/api.pm
```diff
@@ -32,11 +29,10 @@ sub command ($self, @args) {
 
     @args = $self->decode_args(@args);
     die $self->usage unless my $path = shift @args;
+    $data = $data_file eq '-' ? $self->data_from_stdin : path($data_file)->slurp if $data_file;
```
perlpunk, line 32: Maybe this should also be added to the documentation in line 124

PR #5900, opened 2024-08-28
Refactor git functionality into OpenQA::Git

r-richardson:
Refactor git related functions from OpenQA::Task::Git::Clone into the OpenQA::Git module for easier re-usability.
Related Ticket: [164886](https://progress.opensuse.org/issues/164886)

- Turn git functions into git module methods
- Avoid array recreation with _run_cmd helper method and optional use_ssh and use_path parameters
- Improve and adjust and test cases
- Fix merge conflicts after rebase to master branch

### DIFF
--- a/lib/OpenQA/Git.pm
+++ b/lib/OpenQA/Git.pm
@@ -4,6 +4,7 @@
 package OpenQA::Git;
 
 use Mojo::Base -base, -signatures;
+use Mojo::Util 'trim';
 use Cwd 'abs_path';
 use OpenQA::Utils qw(run_cmd_with_log_return_error);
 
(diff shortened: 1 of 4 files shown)


perlpunk (commented):
Regarding the method names:
In non object orient code, it can be good to have functions with a common prefix, so it made sense to have `git` prepended in the current version.
Now for a module that is called `OpenQA::Git` and the object is called `$git`, I think we can remove those prefixes.
e.g.
```diff
- my $origin_url = $git->git_get_origin_url;
+ my $origin_url = $git->get_origin_url;
```

lib/OpenQA/Git.pm
```diff
@@ -21,45 +22,69 @@ sub config ($self, $args = undef) {
...
+    if ($use_ssh) {
+        push @cmd, 'env', 'GIT_SSH_COMMAND="ssh -oBatchMode=yes"';
     }
-    return $error_message;
+
+    push @cmd, $self->_prepare_git_command($use_dir), @$args;
+
+    $self->app->log->debug("Executing Git command: @cmd");
+    my $result = run_cmd_with_log_return_error(\@cmd);
+    if (!$result->{status}) {
```
perlpunk, line 45: `run_cmd_with_log_return_error` is already logging in case of a failure

lib/OpenQA/Git.pm
```diff
@@ -68,30 +93,63 @@ sub set_to_latest_master ($self, $args = undef) {
...
+
+sub clone_url ($self, $url) {
+    my $r = $self->_run_cmd(['clone', $url, $self->dir], {use_dir => 0, use_ssh => 1});
+    die $self->_format_git_error($r, "Failed to clone $url") unless $r->{status};
+}
+
+sub get_origin_url ($self) {
+    my $r = $self->_run_cmd(['remote', 'get-url', 'origin']);
+    die $self->_format_git_error($r, 'Failed to get origin url') unless $r->{status};
+    return Mojo::Util::trim($r->{stdout});
```
perlpunk, line 142:
```suggestion
    trim($r->{stdout});
```

lib/OpenQA/Git.pm
```diff
@@ -21,45 +22,69 @@ sub config ($self, $args = undef) {
...
+    if (!$result->{status}) {
+        $self->app->log->error("Git command failed: @cmd - Error: $result->{stderr}");
+    }
+    return $result;
 }
 
-sub _validate_attributes ($self) {
-    for my $mandatory_property (qw(app dir user)) {
-        die "no $mandatory_property specified" unless $self->$mandatory_property();
+sub _prepare_git_command ($self, $use_dir) {
```
b10n1k, line 51: maybe a little nitpicked but what about `expand_git_path` in place of `use_dir`?

perlpunk: expand means from a short form to a long form. IMHO doesn't fit here

lib/OpenQA/Git.pm
```diff
@@ -21,45 +22,68 @@ sub config ($self, $args = undef) {
...
+    if (!$result->{status}) {
+        $self->app->log->error("Git command failed: @cmd - Error: $result->{stderr}");
+    }
+    return $result;
 }
 
-sub _validate_attributes ($self) {
-    for my $mandatory_property (qw(app dir user)) {
-        die "no $mandatory_property specified" unless $self->$mandatory_property();
+sub _prepare_git_command ($self, $expand_git_path) {
```
perlpunk, line 50: I think expand is really the wrong word here, sorry

perlpunk (approved):
Approving anyway to get things forward

lib/OpenQA/Git.pm
```diff
@@ -21,45 +22,68 @@ sub config ($self, $args = undef) {
...
     }
-    return ('git', '-C', $dir);
 }
 
-sub _format_git_error ($result, $error_message) {
-    if ($result->{stderr} or $result->{stdout}) {
-        $error_message .= ': ' . $result->{stdout} . $result->{stderr};
+sub _run_cmd ($self, $args, $options = {}) {
+    my $expand_git_path = $options->{expand_git_path} // 1;
+    my $use_ssh = $options->{use_ssh} // 0;
```
perlpunk, line 33: I think `use_ssh` is the wrong word. It's not abuot using ssh, but to tell shh to use BatchMode. But that should be done in a followup, so that we don't block other tickets.

lib/OpenQA/Git.pm
r-richardson: changed to `include_git_path`

lib/OpenQA/Git.pm
r-richardson: Changed to `ssh_batchmode`, is this ok for now ? Now sure how this change would block other tickets ?

perlpunk: Yeah.
I just meant the longer we wait with this PR, the longer Dominik has to wait with his PR touching the same code.

b10n1k (approved):

PR #5901, opened 2024-08-29
Only update clean git checkouts

asdil12:
Ticket: https://progress.opensuse.org/issues/165066

### DIFF
--- a/lib/OpenQA/Task/Git/Clone.pm
+++ b/lib/OpenQA/Task/Git/Clone.pm
@@ -68,6 +68,8 @@ sub _git_clone ($app, $job, $ctx, $path, $url) {
         return;
     }
 
+    die "NOT updating dirty git checkout at $path" if !$git->is_workdir_clean();
+
     unless ($requested_branch) {
         my $remote_default = $git->get_remote_default_branch($url);
         $requested_branch = $remote_default;
--- a/lib/OpenQA/Git.pm
+++ b/lib/OpenQA/Git.pm
@@ -150,4 +150,11 @@ sub reset_hard ($self, $branch) {
     die $self->_format_git_error($r, "Failed to reset to 'origin/$branch'") unless $r->{status};
 }
 
+sub is_workdir_clean ($self) {
+    my $r = $self->_run_cmd(['diff-index', 'HEAD', '--exit-code']);
+    die $self->_format_git_error($r, 'Internal Git error: Unexpected exit code ' . $r->{return_code})
+      if $r->{return_code} > 1;
+    return $r->{status};
+}
+
 1;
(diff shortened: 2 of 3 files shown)


lib/OpenQA/Task/Git/Clone.pm
```diff
@@ -112,10 +118,15 @@ sub _git_clone ($job, $ctx, $path, $url) {
         $ctx->debug(qq{Remote default branch $remote_default});
     }
 
-    my $current_branch = _get_current_branch($path);
-    # updating default branch (including checkout)
-    _git_fetch($path, $requested_branch);
-    _git_reset_hard($path, $requested_branch) if ($requested_branch eq $current_branch);
+    if (_git_is_workdir_clean($path)) {
```
perlpunk, line 121: We should try to do things as early as possible to avoid unnecessary git calls.
I would move that up before the `_get_remote_default_branch` and just return. Then we also save one nesting level.

lib/OpenQA/Task/Git/Clone.pm
```diff
@@ -91,6 +91,12 @@ sub _git_reset_hard ($path, $branch) {
     die "Failed to reset to 'origin/$branch': $r->{stderr}" unless $r->{status};
 }
 
+sub _git_is_workdir_clean ($path) {
+    my $r = run_cmd_with_log_return_error(['git', '-C', $path, 'diff-index', 'HEAD', '--exit-code']);
+    return $r->{status};
```
perlpunk, line 96: Should we also check if there is a different kind of error?
e.g. if there is a diff, the exit code is 1, but if the command fails for some other reason, we should log this.
Also maybe we could log the actual output in case there is a diff.

asdil12: `run_cmd_with_log_return_error` is already handling this and logging any errors.

perlpunk:
As a general note: I would like to wait with this until #5900 is merged, because we have several followup tickets that touch this code, and @r-richardson  would have to rebase and resolve again and again.

lib/OpenQA/Task/Git/Clone.pm
```diff
@@ -106,6 +112,11 @@ sub _git_clone ($job, $ctx, $path, $url) {
         return;
     }
 
+    unless (_git_is_workdir_clean($path)) {
+        $ctx->warn("NOT updating dirty git checkout at $path");
```
b10n1k, line 116: would be nice to add a prompt. I assume that the checkout is skipped and need cleaning, right?

perlpunk (commented):
Ok, now you can rebase after #5900 :)

Tests needed

lib/OpenQA/Task/Git/Clone.pm
perlpunk: What do you mean with a prompt here?

Martchus (approved):

perlpunk (changes requested):
This needs tests

lib/OpenQA/Task/Git/Clone.pm
b10n1k: I mean add something like "Need to clean $path." or something

perlpunk (approved):

lib/OpenQA/Task/Git/Clone.pm
perlpunk: To me the error message is clear enoufg

PR #5902, opened 2024-08-29
Refactor ScheduledProducts

perlpunk:
Issue: https://progress.opensuse.org/issues/165866

This doesn't solve the problem, but I found `_schedule_iso` hard to read.

### DIFF
--- a/lib/OpenQA/Schema/Result/ScheduledProducts.pm
+++ b/lib/OpenQA/Schema/Result/ScheduledProducts.pm
@@ -232,6 +232,78 @@ sub _delete_prefixed_args_storing_info_about_product_itself ($args) {
     }
 }
 
+sub _create_jobs_in_database ($self, $jobs, $failed_job_info, $skip_chained_deps, $include_children,
+    $successful_job_ids)
+{
+    my $schema = $self->result_source->schema;
+    my $jobs_resultset = $schema->resultset('Jobs');
+    my @created_jobs;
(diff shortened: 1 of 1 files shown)


b10n1k (approved):

lib/OpenQA/Shared/GruJob.pm
```diff
@@ -41,13 +41,15 @@ sub execute {
 
 sub _delete_gru {
     my ($self, $id) = @_;
-    my $gru = $self->minion->app->schema->resultset('GruTasks')->find($id);
+    my $gru = $self->minion->app->schema->resultset('GruTasks')->find($id)
+      or $self->app->log->error("Could not find GruTask '$id' in database");
```
Martchus, line 45: Wouldn't we get an error in our logs with this every time someone restarts a Minion job via the Minion dashboard or when a Minion job is enqueued using the command API (like our cleanup tasks)? Or are those cases irrelevant because then this custom class/package is not used at all?

Of course the same counts for the other function.

Martchus (commented):
I have just one question. The code looks generally good.

lib/OpenQA/Shared/GruJob.pm
perlpunk: I removed that commit.
That was before I found out that it might be valid to restart a job even if it had been running successfully and deleted the gru task entry.

Martchus (approved):

PR #5905, opened 2024-08-30
Require GruTask in database before running a minion job

perlpunk:
It can happen that a GruTasks entry is added in a transaction and the minion
 job already enqueued and started before the transaction is committed.
    
  In that case we shouldn't start the minion job (the transaction could have
  been rollbacked even).
    
  The task needs to be deleted by the minion job in order to go on with
  processing a job, and it can't delete it if it doesn't see the entry in the
  database.
    
  But requiring the entry before the start of the job is even more correct IMHO.
    
  Issue: https://progress.opensuse.org/issues/165866

### DIFF
--- a/lib/OpenQA/Shared/GruJob.pm
+++ b/lib/OpenQA/Shared/GruJob.pm
@@ -10,6 +10,12 @@ sub execute {
     my $self = shift;
 
     my $gru_id = $self->info->{notes}{gru_id};
+    if ($gru_id and not $self->info->{finished}) {
+        # We have a gru_id and this is the first run of the job
+        my $gru = $self->minion->app->schema->resultset('GruTasks')->find($gru_id);
+        # GruTask might not yet have landed in database due to open transaction
+        return $self->retry({delay => 2}) unless $gru;
+    }
     my $err = $self->SUPER::execute;
 
     # Non-Gru tasks
--- a/t/14-grutasks.t
+++ b/t/14-grutasks.t
@@ -127,6 +127,7 @@ $t->app->minion->add_task(
         return $job->retry({delay => 30})
           unless my $guard = $job->minion->guard('limit_gru_retry_task', ONE_HOUR);
     });
+$t->app->minion->add_task(wait_for_grutask => sub (@) { undef });
 
 # Gru task that reached failed/finished manually
 # uncoverable statement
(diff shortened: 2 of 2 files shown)


perlpunk:
t/api/16-webhooks.t is failing. it is apparently running the same minion job from a previous test, although it had been finished.
Not sure if that is valid.
Trying to adjust the test.
https://app.circleci.com/pipelines/github/os-autoinst/openQA/14492/workflows/2863290b-e984-41b0-ae0c-da689eb88e7d/jobs/137419

perlpunk:
If it's valid to rerun a finished minion job, I could check for the `finished` attribute and in that case not require the GruTask entry, as it proves it was there for the first run...

lib/OpenQA/Shared/GruJob.pm
```diff
@@ -10,6 +10,12 @@ sub execute {
     my $self = shift;
 
     my $gru_id = $self->info->{notes}{gru_id};
+    if ($gru_id and not $self->info->{finished}) {
+        # We have a gru_id and this is the first run of the job
+        my $gru = $self->minion->app->schema->resultset('GruTasks')->find($gru_id);
+        # GruTask might not yet have landed in database due to open transaction
+        return $self->retry({delay => 2}) unless $gru;
+    }
```
b10n1k, line 18: What `retry` exactly does here? does it cancel current scheduled job and create a new one?

perlpunk: It aborts the current minion job and schedules to run the same minion  job again after 2 seconds.

b10n1k: And IIUC jobs history will ends with those canceled jobs. I just wonder how much acceptable is this from the qa perspective. Maybe not so serious impact but I dont really know how to avoid it.

b10n1k: thanks @perlpunk for explanation.resolved

perlpunk: Discussed in chat:
We are talking about **minion** jobs here (not openQA jobs), and those are also not cancelled, but retried, and that will run the very same minion job with the same id, and the `retries` field will be incremented.

The related openQA job will wait in the `scheduled` state until the minion job has run successfully and deleted the entry in the GruTasks table. Only then it will be assigned and start running at some point.

perlpunk:
I changed it so that it will only lookup a GruTask in the database if the minion job hadn't been finished yet, so it is the first run, or it had been retried. In that case the `finished` field is not set.
I hope that should work for all cases. Jobs that don't have a gru task entry won't be considered.

kalikiana (approved):

b10n1k (approved):

PR #5908, opened 2024-09-03
Add clear Add comments button label to make it discoverable

kalikiana:
Also mention that this adds comments to all jobs.

image

### DIFF
--- a/assets/stylesheets/comments.scss
+++ b/assets/stylesheets/comments.scss
@@ -3,8 +3,6 @@
     margin-top: -2px;
     margin-left: 4px;
     border-radius: 50rem;
-    width: 2.2rem;
-    height: 2.2rem;
     opacity: 0.6;
 
     &:hover {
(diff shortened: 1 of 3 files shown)


okurz:
Your commit message starts with "Add clear Add comments ..."?

Martchus (commented):
I find this feature has the right amount of discoverability as-is. Considering it is an advanced feature that should be used with care it make sense to use an icon-only button in my opinion (and that's also why I initially designed it this way).

okurz:
We might want to reconsider the design anyway with https://progress.opensuse.org/issues/166556. Maybe you want to incorporate that into your ideas?

kalikiana:
> I find this feature has the right amount of discoverability as-is. Considering it is an advanced feature that should be used with care it make sense to use an icon-only button in my opinion (and that's also why I initially designed it this way).

See https://suse.slack.com/archives/C02CANHLANP/p1724915110449239?thread_ts=1724833412.133459&cid=C02CANHLANP for the original feedback suggesting people don't find the feature at all.

PR #5910, opened 2024-09-04
Automatically update git for jobs without CASEDIR/NEEDLES_DIR

perlpunk:

This replaces the update part of `fetchneedles`. It doesn't replace the initial cloning yet.
git is updated when
* scheduling a product / job -> tests and needles of the relevant jobs are updated
* scheduled hourly via service (or cronjob) -> all repos under `OPENQA_BASEDIR` are updated

Note: `fetchneedles` has a `force` option. We set it to true in our cronjob currently.
This can be added later as a config option.

### DIFF
--- a/etc/openqa/openqa.ini
+++ b/etc/openqa/openqa.ini
@@ -115,6 +115,8 @@
 #do_push = no
 ## whether to clone CASEDIR or NEEDLES_DIR on the web UI if that var points to a Git repo
 #git_auto_clone = yes
+## Experimental - Ensure a git update of all test code and needles
+#git_auto_update = no
 
 ## Authentication method to use for user management
 [auth]
--- a/lib/OpenQA/Setup.pm
+++ b/lib/OpenQA/Setup.pm
@@ -68,6 +68,7 @@ sub read_config ($app) {
             do_push => 'no',
             do_cleanup => 'no',
             git_auto_clone => 'yes',
+            git_auto_update => 'no',
         },
         scheduler => {
             max_job_scheduled_time => 7,
--- a/lib/OpenQA/Shared/Plugin/Gru.pm
+++ b/lib/OpenQA/Shared/Plugin/Gru.pm
@@ -10,8 +10,10 @@ use DBIx::Class::Timestamps 'now';
 use OpenQA::Schema;
 use OpenQA::Shared::GruJob;
 use OpenQA::Log 'log_info';
+use OpenQA::Utils qw(sharedir);
 use Mojo::Pg;
 use Mojo::Promise;
+use Mojo::File qw(path);
 
 has app => undef, weak => 1;
 has 'dsn';
@@ -163,6 +165,32 @@ sub enqueue_download_jobs ($self, $downloads) {
         my ($path, $do_extract, $block_job_ids) = @{$downloads->{$url}};
         $self->enqueue('download_asset', [$url, $path, $do_extract], {priority => 10}, $block_job_ids);
     }
+}
+
+sub enqueue_git_update_all ($self) {
+    my $conf = OpenQA::App->singleton->config->{'scm git'};
+    return if $conf->{git_auto_clone} ne 'yes' || $conf->{git_auto_update} ne 'yes';
+    my %clones;
+    my $testdir = path(sharedir() . '/tests');
+    for my $distri ($testdir->list({dir => 1})->each) {
+        next unless -d $distri;    # no symlinks
+        next unless -e $distri->child('.git');
+        $clones{$distri} = undef;
+        if (-e $distri->child('products')) {
+            for my $product ($distri->child('products')->list({dir => 1})->each) {
+                next unless -d $product;    # no symlinks
+                my $needle = $product->child('needles');
+                next unless -e $needle->child('.git');
+                $clones{$needle} = undef;
+            }
+        }
+        else {
+            my $needle = $distri->child('needles');
+            next unless -e $needle->child('.git');
+            $clones{$needle} = undef;
+        }
+    }
+    $self->enqueue('git_clone', \%clones, {priority => 10});
 }
 
 sub enqueue_git_clones ($self, $clones, $job_ids) {
--- a/lib/OpenQA/Shared/Plugin/Gru.pm
+++ b/lib/OpenQA/Shared/Plugin/Gru.pm
@@ -23,21 +23,20 @@ sub new ($class, $app = undef) {
 
 sub register_tasks ($self) {
     my $app = $self->app;
-    $app->plugin($_)
-      for (
-        qw(OpenQA::Task::AuditEvents::Limit),
-        qw(OpenQA::Task::Asset::Download),
-        qw(OpenQA::Task::Asset::Limit),
-        qw(OpenQA::Task::Git::Clone),
-        qw(OpenQA::Task::Needle::Scan OpenQA::Task::Needle::Save OpenQA::Task::Needle::Delete),
-        qw(OpenQA::Task::Job::Limit),
-        qw(OpenQA::Task::Job::ArchiveResults),
-        qw(OpenQA::Task::Job::FinalizeResults),
-        qw(OpenQA::Task::Job::HookScript),
-        qw(OpenQA::Task::Job::Restart),
-        qw(OpenQA::Task::Iso::Schedule),
-        qw(OpenQA::Task::Bug::Limit),
-      );
+    $app->plugin($_) for qw(
+      OpenQA::Task::AuditEvents::Limit
+      OpenQA::Task::Asset::Download
+      OpenQA::Task::Asset::Limit
+      OpenQA::Task::Git::Clone
+      OpenQA::Task::Needle::Scan OpenQA::Task::Needle::Save OpenQA::Task::Needle::Delete
+      OpenQA::Task::Job::Limit
+      OpenQA::Task::Job::ArchiveResults
+      OpenQA::Task::Job::FinalizeResults
+      OpenQA::Task::Job::HookScript
+      OpenQA::Task::Job::Restart
+      OpenQA::Task::Iso::Schedule
+      OpenQA::Task::Bug::Limit
+    );
 }
 
 # allow the continuously polled stats to be available on an
--- a/t/14-grutasks-git.t
+++ b/t/14-grutasks-git.t
@@ -23,6 +23,8 @@ my $workdir = tempdir("$FindBin::Script-XXXX", TMPDIR => 1);
 my $guard = scope_guard sub { chdir $FindBin::Bin };
 chdir $workdir;
 path('t/data/openqa/db')->make_path;
+my $git_clones = "$workdir/git-clones";
+mkdir $git_clones;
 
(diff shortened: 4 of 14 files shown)


Martchus (commented):
Looks good so far.

okurz (changes requested):
Why do we need the hourly update when there is already an update when new jobs are triggered?

perlpunk:
> Why do we need the hourly update when there is already an update when new jobs are triggered?

To avoid very big updates when there were many commits but no according test scheduled for a while

perlpunk:
Note that the perltidy check is currently not working due to
* `tidyall` is not installed in our test container apparently
* the new tools/tidyall wrapper doesn't check if the exec was successful

perlpunk:
https://app.circleci.com/pipelines/github/os-autoinst/openQA/14588/workflows/550adc1a-6b94-4b80-9bc9-553406a07205/jobs/138447
```
[21:00:46] t/ui/10-tests_overview.t ................... 18/? 
        #   Failed test '#flavor_DVD_arch_x86_64 present exactly once'
        #   at /home/squamata/project/t/ui/../lib/OpenQA/SeleniumTest.pm line 237.
        #          got: '0'
        #     expected: '1'

        #   Failed test '#flavor_DVD_arch_x86_64 exists'
        #   at /home/squamata/project/t/ui/../lib/OpenQA/SeleniumTest.pm line 240.
        # Looks like you failed 2 tests of 18.

    #   Failed test 'filter for specific machine'
    #   at t/ui/10-tests_overview.t line 474.
[21:00:46] t/ui/10-tests_overview.t ................... 20/? 
#   Failed test 'filtering by machine'
#   at t/ui/10-tests_overview.t line 475.
[21:00:46] t/ui/10-tests_overview.t ................... 26/? # Looks like you failed 1 test of 26.
[21:00:46] t/ui/10-tests_overview.t ................... Dubious, test returned 1 (wstat 256, 0x100)
Failed 1/26 subtests 
```
I've seen that elsewhere today, not related I guess.
I'll retrigger and create a ticket

okurz:
> > Why do we need the hourly update when there is already an update when new jobs are triggered?
> 
> To avoid very big updates when there were many commits but no according test scheduled for a while

How about covering that in documentation like http://open.qa/docs/#_timers_and_triggers ? But then probably we want to rephrase at least the chapter title which currently says [Cleanup of assets, results and other data](http://open.qa/docs/#cleanup) . It could be "Maintenance of …"

perlpunk:
Regarding documentation: I would rather try this feature out for a while before documenting it. That's why I didn't add it to openqa.ini as well. Should have stated that in the description.
Also I'm still not completely happy with the feature name.

lib/OpenQA/Setup.pm
```diff
@@ -68,6 +68,7 @@ sub read_config ($app) {
             do_push => 'no',
             do_cleanup => 'no',
             git_auto_clone => 'yes',
+            git_auto_clone_all => 'no',
```
Martchus, line 71: Maybe it would make sense to emphasize on the fact that this will *update* existing checkouts (and actually not clone new ones):
```suggestion
            git_auto_update => 'no',
```

I would avoid `_all`. I know what you're getting at here - with this setting more jobs are affected than with just `git_auto_clone` - but to a user `_all` is probably not saying much.

(I think we'll eventually need to describe under what conditions which settings apply and what Git commands they'll execute in a comment in the ini file - but this can be done in a follow-up.)

perlpunk: I changed it to `git_auto_update`

perlpunk:
I'm also skipping symlinks now, like in fetchneedles

perlpunk:
https://app.circleci.com/pipelines/github/os-autoinst/openQA/14646/workflows/4dc1a2ed-d4b8-4120-b075-cefe02b94746/jobs/139066
```
[20:25:07] t/ui/10-tests_overview.t ................... 12/? 
        #   Failed test 'exactly one row present'
        #   at t/ui/10-tests_overview.t line 308.
        #          got: '0'
        #     expected: '1'
        # Looks like you failed 1 test of 1.

    #   Failed test 'filter for specific test'
    #   at t/ui/10-tests_overview.t line 310.
[20:25:07] t/ui/10-tests_overview.t ................... 14/? 
#   Failed test 'filtering by test'
#   at t/ui/10-tests_overview.t line 311.
Can't call method "get_text" on an undefined value at t/ui/10-tests_overview.t line 309.
[20:25:07] t/ui/10-tests_overview.t ................... 15/? # Tests were run but no plan was declared and done_testing() was not seen.
[20:25:07] t/ui/10-tests_overview.t ................... Dubious, test returned 1 (wstat 256, 0x100)
Failed 1/15 subtests 
```
Will retrigger and create a ticket. Seems to be a different issue than the previous failure: https://github.com/os-autoinst/openQA/pull/5910#issuecomment-2350252387

systemd/openqa-enqueue-git-update-all.service
```diff
@@ -0,0 +1,9 @@
+[Unit]
+Description=Enqueues a git update of all tests/needles every hour.
```
okurz, line 2: This needs to be corrected as the service does not "enqueue … every hour". When it is started this still enqueues a git update but for immediate execution. Also maybe better make it clear what is the effect. And better avoid the ambiguous term "tests". How about

```suggestion
Description=Ensure a git update of all test code and needles regardless of test execution.
```

systemd/openqa-enqueue-git-update-all.timer
```diff
@@ -0,0 +1,9 @@
+[Unit]
+Description=Enqueues a git update of all tests/needles every hour.
```
okurz, line 2: Similar as commented above

```suggestion
Description=Enqueues a git update of all test code and needles every hour.
```

t/14-grutasks-git.t
```diff
@@ -95,8 +114,11 @@ subtest 'git clone' => sub {
 
     subtest 'no default remote branch' => sub {
         $ENV{OPENQA_GIT_CLONE_RETRIES} = 0;
-        $clone_dirs = {'/tmp/' => 'http://localhost/nodefault.git'};
-        my $res = run_gru_job($t->app, 'git_clone', $clone_dirs, {priority => 10});
+        $clone_dirs = {"$git_clones/nodefault" => 'http://localhost/nodefault.git'};
+        stderr_like {
+            $res = run_gru_job($t->app, 'git_clone', $clone_dirs, {priority => 10})
+        }
+        qr(Error detecting remote default), 'error on stderr';
```
okurz, line 121: As a similar block repeats very often I suggest to shorten it so that the repeated line is only a single line. How about


```suggestion
@args = ($t->app, 'git_clone', $clone_dirs, {priority => 10});
…
        stderr_like { $res = run_gru_job(@args) } qr(Error detecting remote default), 'error on stderr';
```

if that's valid syntax :)

okurz:
> Regarding documentation: I would rather try this feature out for a while before documenting it. That's why I didn't add it to openqa.ini as well. Should have stated that in the description. Also I'm still not completely happy with the feature name.

For the sake of completeness please still include it in openqa.ini as well as we already have everything else included like systemd services and timers, etc. You may still comment it as something experiment if you prefer

okurz (changes requested):

systemd/openqa-enqueue-git-update-all.service
Martchus: The "enqueuing" part is actually correct. This service does nothing more (and the gru service is actually doing the real job). So maybe:

```suggestion
Description=Enqueues a git update of all test code and needles regardless of test execution.
```

t/14-grutasks-git.t
perlpunk: Ok. I had to make some adjustments for the changing `$clone_dirs` values.

systemd/openqa-enqueue-git-update-all.service
perlpunk: I thought so as well, but still changed it to @okurz' suggestion.
I think it's ok, and the script name has `enqueue` in it.

lib/OpenQA/Shared/Plugin/Gru.pm
```diff
@@ -166,6 +166,33 @@ sub enqueue_download_jobs ($self, $downloads) {
     }
 }
 
+sub enqueue_git_update_all ($self) {
+    my %clones;
+    my $conf = OpenQA::App->singleton->config->{'scm git'};
+    return unless $conf->{git_auto_clone} eq 'yes';
+    return unless $conf->{git_auto_update} eq 'yes';
```
Martchus, line 173:
```suggestion
    return unless $conf->{git_auto_clone} eq 'yes' && $conf->{git_auto_update} eq 'yes';
```
(Not important and keep your version if tidy would wrap it.)

Martchus (changes requested):

etc/openqa/openqa.ini
```diff
@@ -115,6 +115,8 @@
 #do_push = no
 ## whether to clone CASEDIR or NEEDLES_DIR on the web UI if that var points to a Git repo
 #git_auto_clone = yes
+## Experimental - Ensure a git update of all test code and needles regardless of test execution
```
okurz, line 118: Ensure a *periodic* git update?

perlpunk: I changed it now and made it shorter.
Can we talk about doc improvements later please?
We should really get the feature done and deployed in production and see how it performs.

lib/OpenQA/Shared/Plugin/Gru.pm
perlpunk: done

lib/OpenQA/Shared/Plugin/Gru.pm
```diff
@@ -166,6 +166,33 @@ sub enqueue_download_jobs ($self, $downloads) {
...
+    my $testdir = path("$ENV{OPENQA_BASEDIR}/openqa/share/tests");
```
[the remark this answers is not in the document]
perlpunk: done

okurz (commented):
Looks good with the exception of the failing full-stack test https://app.circleci.com/pipelines/github/os-autoinst/openQA/14687/workflows/5a12e57a-aea7-4f71-9bce-f47809a264c0/jobs/139517 which in doubt I am assuming to be related until proven otherwise. so far it did not look like an obvious, unrelated, otherwise known problem to me

Martchus (approved):

okurz (approved):

perlpunk:
I retriggered the fullstack test and it passed. I also executed it locally a couple of times and it passed. Will create a ticket.

Now also the test coverage for t fails:
https://app.codecov.io/gh/os-autoinst/openQA/pull/5910/indirect-changes
But I have seen this in https://github.com/os-autoinst/openQA/pull/5943 as well, so also unrelated.
Will check locally what happens.

perlpunk:
https://progress.opensuse.org/issues/167272 for the coverage issue
https://progress.opensuse.org/issues/167269 for the fullstack issue

I can't fix the coverage issue right now, and it seems to happen in every PR.
Force merge?

PR #5913, opened 2024-09-09
WIP Evaluate viability of using HTMX

r-richardson:
Related: https://progress.opensuse.org/issues/156055

Draft PR

### DIFF
--- a/assets/assetpack.def
+++ b/assets/assetpack.def
@@ -119,6 +119,9 @@
 < javascripts/needleeditor.js
 < javascripts/shapes.js
 
+! htmx.js
+< ../node_modules/htmx.org/dist/htmx.min.js
+
 ! bootstrap.js
 < ../node_modules/jquery/dist/jquery.min.js
 < javascripts/openqa.js
--- a/assets/javascripts/comments.js
+++ b/assets/javascripts/comments.js
@@ -38,7 +38,7 @@ function showXhrError(context, jqXHR, textStatus, errorThrown) {
   window.alert(context + getXhrError(jqXHR, textStatus, errorThrown));
 }
 
-function updateNumerOfComments() {
+function updateNumberOfComments() {
   const commentsLink = document.querySelector('a[href="#comments"]');
   if (commentsLink) {
     const linkText = 'Comments (' + document.getElementsByClassName('comment-row').length + ')';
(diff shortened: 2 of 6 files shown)


assets/assetpack.def
```diff
@@ -164,6 +164,9 @@
 ! job_next_previous.js
 < javascripts/job_next_previous.js
 
+! htmx.js
+< ../node_modules/htmx.org/dist/htmx.js
```
Martchus, line 168: What's the size of that file?

templates/webapi/comments/comment_row.html.ep
```diff
@@ -15,27 +15,33 @@
...
-                            <a href="#comment-<%= $comment_id %>" class="comment-anchor"><abbr class="timeago" title="<%= $comment->t_created->datetime() %>Z"><%= format_time($comment->t_created) %></abbr></a>
-                            % if ($comment->t_updated->subtract(seconds => OpenQA::Constants::DB_TIMESTAMP_ACCURACY) >= $comment->t_created) {
-                                (last edited <abbr class="timeago" title="<%= $comment->t_updated->datetime() %>Z"><%= format_time($comment->t_updated) %></abbr>)
-                            % }
+                      <button 
+                        class="btn btn-danger btn-circle btn-sm remove-edit-button" 
+                        hx-delete="<%= url_for($delete_action => ($context->{type}.'_id' => $context->{id}, comment_id => $comment_id)) %>" 
+                        hx-confirm="Do you really want to delete the comment written by <%= $user->name %>?"
+                        hx-target="closest .comment-row, .pinned-comment-row" 
+                        hx-swap="outerHTML"
```
Martchus, line 23: I don't think we want to replace the target element with the response. We want to delete the target element - but of course only if the response was a 200 return code. We also need to decrement the number of comments shown in the tab header (if the comment is part of the test details page).

I guess you've chosen a good example here. That *is* the case kind of simple case we might be able to use htmx for. However, if it turns out that even this simple case it too complicated we know that we probably don't want to use htmx.

assets/javascripts/comments.js
```diff
@@ -46,21 +46,21 @@ function updateNumerOfComments() {
...
+//  $.ajax({
+//    url: deleteButton.dataset.deleteUrl,
+//    method: 'DELETE',
+//    success: () => {
+//      $(deleteButton).parents('.comment-row, .pinned-comment-row').remove();
+//      updateNumerOfComments();
+//    },
+//    error: showXhrError.bind(undefined, "The comment couldn't be deleted: ")
+//  });
+//}
```
Martchus, line 63: Feel free to delete the code completely. This way the diff statistics tell use how many lines of custom code we avoided.

Martchus (commented):
I guess this is a good showcase. We can discuss it in one of the next meetings.

assets/assetpack.def
r-richardson:
```
160K	./htmx.js
52K	./htmx.min.js
16K	./htmx.min.js.gz
```

t/16-utils-vcs-provider.t
```diff
@@ -7,7 +7,7 @@ use Test::Most;
 use FindBin;
 use lib "$FindBin::Bin/lib", "$FindBin::Bin/../external/os-autoinst-common/lib";
 require OpenQA::Test::Database;
-use OpenQA::Test::TimeLimit '5';
+use OpenQA::Test::TimeLimit '25';
```
kalikiana, line 10: As you bumped the timeout without any other changes it might be worth checking if/why the test is slow for you. Generally it should work unmodified

PR #5914, opened 2024-09-09
[POC] Trigger os-autoinst-distri-example tests from fresh openQA

b10n1k:
With this PR a logged in user can trigger a job via a form, providing
settings, git repo and a scenarion definition. User doesnt need to have
special role. Keep in mind that NEEDLES_DIR doesnt need for
os-autoinst-distri-example.
Things which stayed out of this PR are:
- The UI design
- Posibility to upload the yaml definition
- Some way to return data to the webUI, like the job id
- Input validation
- Additional error handling (maybe)

https://progress.opensuse.org/issues/162899

### DIFF
--- a/templates/webapi/NewJob/create.html.ep
+++ b/templates/webapi/NewJob/create.html.ep
@@ -1,0 +1,126 @@
+% layout 'bootstrap';
+% content_for 'head' => begin
+  %= asset 'ace.js'
+  %= asset 'ace.css'
+% end
+
+% title 'New job';
+<div>
+  <h2><%= title %></h2>
+  
+  %= include 'layouts/info'
+  
+  <form class="row g-3" id="form">
+    <div class="col-md-6">
+      <label for="distri" class="form-label">Distri</label>
+      <input type="text" class="form-control" id="distri"
+	     placeholder="example" name="distri">
+    </div>
+    <div class="col-md-6">
+      <label for="version" class="form-label">Version</label>
+      <input type="text" class="form-control" id="version" placeholder="0"  name="version">
+    </div>
+    <div class="col-6">
+      <label for="flavor" class="form-label">Flavor</label>
+      <input type="text" class="form-control" id="flavor" placeholder="DVD"  name="Flavor">
+    </div>
+    <div class="col-12">
+      <label for="arch" class="form-label">Arch</label>
+      <select id="arch" class="form-select"  name="arch">
+	<option selected>Choose...</option>
+	<option>x86_64</option>
+	<option>arch64</option>
+      </select>
+    </div>
+    <div class="col-md-6">
+      <label for="test" class="form-label">Test Name</label>
+      <input type="text" class="form-control" id="test" placeholder="simple_boot" name="TEST">
+      </div>
+    <div class="col-md-6">
+      <label for="build" class="form-label">Build</label>
+      <input type="text" class="form-control" id="build" name="BUILD">
+    </div>
+    <div class="col-md-6">
+      <label for="casedir" class="form-label">CASEDIR</label>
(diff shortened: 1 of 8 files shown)


b10n1k:
https://github.com/user-attachments/assets/b65bf5ed-d37d-49df-9257-0e48d98246cb

Due to the allowed size of upload the footage is as small as possible

templates/webapi/NewJob/create.html.ep
```diff
@@ -0,0 +1,122 @@
+% layout 'bootstrap';
```
Martchus, line 1: We don't use camel-cased path elements here. So please move this file under `test`. (Is it really that hard to follow the established conventions of a projects. Please don't do everything your way. If everyone would do that this project would end up as a total mess of inconsistencies. **And it becomes really annoying having to state this so often.**)

templates/webapi/NewJob/create.html.ep
```diff
@@ -0,0 +1,122 @@
...
+% end
+
+% title 'Run a Job';
+
+<div>
+  <h2><%= title %></h2>
+  
+  %= include 'layouts/info'
+  <p><b>Hint:</b> Experimental form to trigger a test E.g. see <%= link_to 'this
+								      link' => url_for('not_found') %>.
```
Martchus, line 14: This way of wrapping it is weird:

```suggestion
  <p><b>Hint:</b> Experimental form to trigger a test E.g. see <%= link_to 'this link' => url_for('not_found') %>.
```

templates/webapi/NewJob/create.html.ep
```diff
@@ -0,0 +1,122 @@
...
+  </p>
+  <div class="container-md">
+  <form class="row g-3" onsubmit="runExample(this);" data-post-url="/api/v1/isos">
```
Martchus, line 17: The wrapping in the subsequent HTML code is completely inconsistent.

Martchus, line 17: We should not hardcode paths like `/api/v1/isos`. Use the Mojolicious helper `url_for` instead.

Martchus (changes requested):
I suggest you clean up this PR as mentioned in the inline-comments. Even though it is just a POC we probably want something another person could actually continue to work on. This is not even in a state where it is fun to look at for the sake of reviewing.

I added a few remarks about the modal. However, I suggest you remove it completely and show a flash message instead. That is much easier than making the modal actually work and the modal is not used/working in its current form anyway.

templates/webapi/NewJob/create.html.ep
```diff
@@ -0,0 +1,122 @@
...
+machines:
+  64bit:
+    backend: "qemu"
+job_templates:
+  simple_boot:
+    product: "example"
+    machine: "64bit"
+      </textarea>
+    </div>
+
```
okurz, line 92: I would leave this whole block out for now. openQA should read the scenario definitions from the example test distribution itself.

okurz (changes requested):

templates/webapi/NewJob/create.html.ep
perlpunk: But having this in the PoC makes it easy to test

templates/webapi/NewJob/create.html.ep
b10n1k: Now I noticed. believe me if I say I didnt try to do anything on my way here. rather trying to find my way. and what happened briefly I create this somewhere else and then I copied and remained as such because I didnt noticed

b10n1k:
this has been implemented https://github.com/os-autoinst/openQA/pull/5933

PR #5921, opened 2024-09-11
Slightly simplify Schema::Result::ScheduledProducts

okurz:
This commit extracts a function instead of a code block with comment and
uses post-for calls for two loops.

Inspired by https://github.com/os-autoinst/openQA/pull/5902


okurz:
@perlpunk do you also agree with the change?

PR #5923, opened 2024-09-12
Exclude sub packages with many additional dependencies on SLE

Martchus:
As stated in https://build.suse.de/request/show/343857#comment-4023273 not all sub packages are installable on SLE. This only concerns the `devel` and `munin` sub packages which are really optional. So it makes sense to avoid submitting a huge number of Haskell and Perl packages to SLE and exclude those packages instead.

Related ticket: https://progress.opensuse.org/issues/165683

### DIFF
--- a/dist/rpm/openQA.spec
+++ b/dist/rpm/openQA.spec
@@ -42,6 +42,14 @@
 %else
 %bcond_without python_scripts
 %endif
+# exclude additional sub packages that would pull in a lot of extra dependencies on SLE
+%if 0%{?sle_version} && !0%{?is_opensuse}
+%bcond_with devel_package
+%bcond_with munin_package
+%else
+%bcond_without devel_package
+%bcond_without munin_package
+%endif
 # runtime requirements that also the testsuite needs
 %if %{with python_scripts}
 %define python_scripts_requires python3-base python3-requests openQA-client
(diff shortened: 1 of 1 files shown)


Martchus:
I'm currently testing this on https://build.opensuse.org/package/show/home:mkittler:branches:devel:openQA/openQA.

okurz (commented):
Why would it make sense to exclude a devel package on SLE when we still want to build and test the package on SLE?

Martchus:
The devel package is not required to build and test the package. It only contains further dependencies such as Selenium/chromedriver and ShellCheck to one can install them conveniently to setup a local development environment. Note that during the package build Selenium and developer-facing tests using ShellCheck are *not* executed and this is not SLE-specific.

okurz (commented):
But with package hub it would be possible to build those packages. I am not convinced we want to disable the build completely based on those conditions. Would it be possible to set/unset the condition variable in ibs project config? Or as a completely different alternative disable the build of selected sub packages in the according ibs projects?

Martchus:
> Would it be possible to set/unset the condition variable in ibs project config? Or as a completely different alternative disable the build of selected sub packages in the according ibs projects?

I don't know. I would generally say that this is not possible because we are not in control of the project config where the final package will end up.

I think we should treat SLE as "pure SLE" in the spec file and not as "SLE plus package hub". If one wants to build those packages with package hub they can still do so by adding the following to their project config:

```
Macros:
%_with_devel_package 1
%_with_munin_package 1
:Macros
```

Martchus:
Note that I also generally don't think these kinds of packages (or rather the dependencies they pull in) belong to something we should generally support on SLE. For instance, just because we want to add a cool new linter we shouldn't need to support that on SLE as well. Also supporting the Haskell stack (of which we as a team have probably very little experience with) is not the best idea.

kalikiana (approved):

PR #5924, opened 2024-09-12
Add missing flash message box to /tests page

asdil12:
This page can have flash messages created by ajax events.

### DIFF
--- a/templates/webapi/test/list.html.ep
+++ b/templates/webapi/test/list.html.ep
@@ -8,6 +8,10 @@
   renderTestLists();
 % end
 
+<div id="flash-messages">
+    %= include 'layouts/flash_messages';
+</div>
+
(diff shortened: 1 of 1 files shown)


kalikiana (approved):

Martchus:
> This page can have flash messages *created by ajax events*.

The last part let's me wonder. For flash messages created by ajax handlers we don't need to include the layout that will add flash messages from server-side stash variables. So maybe we can just add the div here without including the layout.

asdil12:
Sure we could do that. But I prefer to not have two different flash snippets around where some pages then support serverside flash messages and some only support client side messages.
Actually flash messages should be part of the basic layout and fully work on every page.

okurz (approved):

PR #5925, opened 2024-09-13
Replace $.ajax() with fetch()

asdil12:
Ticket: https://progress.opensuse.org/issues/166310

### DIFF
--- a/assets/javascripts/admin_assets.js
+++ b/assets/javascripts/admin_assets.js
@@ -121,37 +121,42 @@ function reloadAssetsTable() {
 }
 
 function deleteAsset(assetId) {
-  $.ajax({
-    url: urlWithBase('/api/v1/assets/' + assetId),
-    method: 'DELETE',
-    dataType: 'json',
-    success: function () {
+  fetchWithCSRF(urlWithBase(`/api/v1/assets/${assetId}`), {method: 'DELETE'})
+    .then(response => {
+      // not checking for status code as 404 case also returns proper json
+      return response.json();
+    })
+    .then(response => {
+      if (response.error) throw response.error;
       addFlash(
         'info',
-        'The asset was deleted successfully. The asset table\'s contents are cached. Hence the removal is not immediately visible. To update the view use the "Trigger asset cleanup" button. Note that this is an expensive operation which might take a while.'
+        "The asset was deleted successfully. The asset table's contents are cached." +
+          'Hence the removal is not immediately visible. To update the view use the "Trigger asset cleanup" button.' +
+          'Note that this is an expensive operation which might take a while.'
       );
-    },
-    error: function (xhr, ajaxOptions, thrownError) {
-      var error_message = xhr.responseJSON.error;
-      addFlash('danger', error_message);
-    }
-  });
+    })
+    .catch(error => {
+      console.error(error);
+      addFlash('danger', `Error deleting asset: ${error}`);
+    });
 }
 
 function triggerAssetCleanup(form) {
-  $.ajax({
-    url: form.action,
-    method: form.method,
-    success: function () {
+  fetchWithCSRF(form.action, {method: form.method})
+    .then(response => {
+      if (!response.ok) throw `Server returned ${response.status}: ${response.statusText}`;
+      return response.json();
+    })
+    .then(response => {
       addFlash(
         'info',
-        'Asset cleanup has been triggered. Open the <a href="/minion/jobs?task=limit_assets">Minion dashboard</a> to keep track of the task.'
+        `Asset cleanup has been triggered. Open the <a href="/minion/jobs?task=limit_assets">Minion dashboard</a> to keep track of the task (gru_id #${response.gru_id}).`
       );
-    },
-    error: function (xhr, ajaxOptions, thrownError) {
-      addFlash('danger', 'Unable to trigger the asset cleanup: ' + thrownError);
-    }
-  });
+    })
+    .catch(error => {
+      console.error(error);
+      addFlash('danger', `Unable to trigger the asset cleanup: ${error}`);
+    });
 }
 
 function showLastAssetStatusUpdate(assetStatus) {
--- a/assets/javascripts/admin_groups.js
+++ b/assets/javascripts/admin_groups.js
@@ -37,21 +37,23 @@ function showError(message) {
 }
 
 function fetchHtmlEntry(url, targetElement) {
-  $.ajax({
-    url: url,
-    method: 'GET',
-    success: function (response) {
+  fetch(url)
+    .then(response => {
+      if (!response.ok) throw `Server returned ${response.status}: ${response.statusText}`;
+      return response.text();
+    })
+    .then(response => {
       var element = $(response);
       element.hide();
       targetElement.prepend(element);
       $('#new_group_creating').hide();
       $('#add_group_modal').modal('hide');
       element.fadeIn('slow');
-    },
-    error: function (xhr, ajaxOptions, thrownError) {
-      showError(thrownError + ' (requesting entry HTML, group probably added though! - reload page to find out)');
-    }
-  });
+    })
+    .catch(error => {
+      console.error(error);
+      showError(`${error} (requesting entry HTML, group probably added though! - reload page to find out)`);
+    });
 }
 
 function countEmptyInputs(form) {
@@ -87,7 +89,7 @@ function createGroup(form) {
   $('#new_group_error').hide();
   $('#new_group_creating').show();
 
-  let data = $(form).serialize();
+  let data = new FormData(form);
   let postUrl, rowUrl, targetElement;
   if (form.dataset.createParent !== 'false') {
     postUrl = form.dataset.postParentGroupUrl;
@@ -99,36 +101,26 @@ function createGroup(form) {
     const parentId = form.dataset.parentId;
     if (parentId !== 'none') {
       targetElement = $('#parent_group_' + parentId).find('ul');
-      data += '&parent_id=' + parentId;
+      data.set('parent_id', parentId);
     } else {
       targetElement = $('#job_group_list');
     }
   }
 
-  $.ajax({
-    url: postUrl,
-    method: 'POST',
-    data: data,
-    success: function (response) {
-      if (!response) {
-        showError('Server returned no response');
-        return;
-      }
-      var id = response.id;
-      if (!id) {
-        showError('Server returned no ID');
-        return;
-      }
+  fetchWithCSRF(postUrl, {method: 'POST', body: data})
+    .then(response => {
+      if (!response.ok) throw `Server returned ${response.status}: ${response.statusText}`;
+      return response.json();
+    })
+    .then(response => {
+      if (!response) throw 'Server returned no response';
+      if (!response.id) throw 'Server returned no ID';
       fetchHtmlEntry(rowUrl + response.id, targetElement);
-    },
-    error: function (xhr, ajaxOptions, thrownError) {
-      if (xhr.responseJSON.error) {
-        showError(xhr.responseJSON.error);
-      } else {
-        showError(thrownError);
-      }
-    }
-  });
+    })
+    .catch(error => {
+      console.error(error);
+      showError(error);
+    });
 
   return false;
 }
@@ -262,24 +254,25 @@ function saveReorganizedGroups() {
   var updateJobGroupUrl = jobGroupList.data('put-job-group-url');
 
   // event handlers for AJAX queries
-  var handleError = function (xhr, ajaxOptions, thrownError) {
+  var handleError = function (error) {
+    console.error(error);
     $('#reorganize_groups_panel').show();
     $('#reorganize_groups_error').show();
     $('#reorganize_groups_progress').hide();
-    $('#reorganize_groups_error_message').text(thrownError ? thrownError : 'something went wrong');
+    $('#reorganize_groups_error_message').text(error ? error : 'something went wrong');
     $('html, body').animate({scrollTop: 0}, 1000);
   };
 
   var handleSuccess = function (response, groupLi, index, parentId) {
     if (!response) {
-      handleError(undefined, undefined, 'Server returned nothing');
+      handleError('Server returned nothing');
       return;
     }
 
     if (!response.nothingToDo) {
       var id = response.id;
       if (!id) {
-        handleError(undefined, undefined, 'Server returned no ID');
+        handleError('Server returned no ID');
         return;
       }
 
@@ -292,7 +285,7 @@ function saveReorganizedGroups() {
 
     if (ajaxQueries.length) {
       // do next query
-      $.ajax(ajaxQueries.shift());
+      handleQuery(ajaxQueries.shift());
     } else {
       // all queries done
       if (showPanelTimeout) {
@@ -325,7 +318,7 @@ function saveReorganizedGroups() {
       ajaxQueries.push({
         url: updateGroupUrl + groupId,
         method: 'PUT',
-        data: {
+        body: {
           sort_order: groupIndex,
           parent_id: 'none',
           drag: 1
@@ -350,7 +343,7 @@ function saveReorganizedGroups() {
             ajaxQueries.push({
               url: updateJobGroupUrl + jobGroupId,
               method: 'PUT',
-              data: {
+              body: {
                 sort_order: childGroupIndex,
                 parent_id: groupId,
                 drag: 1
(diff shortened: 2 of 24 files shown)


assets/javascripts/openqa.js
```diff
@@ -23,6 +23,16 @@ function setupForAll() {
   });
 }
 
+function getCSRFToken() {
+  return document.querySelector('meta[name="csrf-token"]').content;
+}
+
+function fetchWithCSRF(resource, options) {
+  if (!('headers' in options)) options['headers'] = {};
+  options['headers']['X-CSRF-TOKEN'] = getCSRFToken();
```
Martchus, line 32:
```suggestion
  options.headers ??= {};
  options.headers['X-CSRF-TOKEN'] ??= getCSRFToken();
```

assets/javascripts/tests.js
```diff
@@ -191,16 +191,17 @@ function changeJobPrio(jobId, delta, linkElement) {
...
-    method: 'POST',
-    success: function (result) {
-      prioValueElement.text(newPrio);
-    },
-    error: function (xhr, ajaxOptions, thrownError) {
-      addFlash('danger', 'Unable to set the priority value of job ' + jobId + '.');
-    }
-  });
+
+  fetchWithCSRF(urlWithBase('/api/v1/jobs/' + jobId + '/prio?prio=' + newPrio), {method: 'POST'})
```
Martchus, line 195: If you like you can also change the code like this (but of course this is actually independent and could be done as a separate improvement):
```suggestion
  fetchWithCSRF(urlWithBase(`/api/v1/jobs/${jobId}/prio?prio=${newPrio}`), {method: 'POST'})
```

Martchus:
Looks generally good so from my side you could go ahead and change more places this way.

assets/javascripts/comments.js
```diff
@@ -51,15 +47,15 @@ function deleteComment(deleteButton) {
   if (!window.confirm('Do you really want to delete the comment written by ' + author + '?')) {
     return;
   }
-  $.ajax({
-    url: deleteButton.dataset.deleteUrl,
-    method: 'DELETE',
-    success: () => {
+  fetchWithCSRF(deleteButton.dataset.deleteUrl, {method: 'DELETE'})
+    .then(response => {
+      if (response.status != 200) throw `Server returned ${response.status}: ${response.statusText}`;
```
b10n1k, line 52: Why `response.status` is limited on 200. I wonder if `response.ok` is more appropiate in those if https://developer.mozilla.org/en-US/docs/Web/API/Response/ok

assets/javascripts/openqa.js
```diff
@@ -261,13 +272,12 @@ function restartJob(ajaxUrl, jobId) {
       if (newJobUrl) {
         window.location.replace(newJobUrl);
       } else {
-        showError('URL for new job not available');
+        throw 'URL for new job not available';
```
b10n1k, line 275: is there any reason to not stay with `showError` here?

asdil12: We actually do.
The throw will jump to the catch.
And the catch executes `showError`.

lib/OpenQA/WebAPI/Controller/API/V1/Table.pm
```diff
@@ -352,6 +353,23 @@ Use by both B<create()> and B<update()> method.
 sub _prepare_settings {
     my ($self, $table, $entry) = @_;
     my $validation = $self->validation;
+    my $hp;
+    # accept both traditional application/x-www-form-urlencoded parameters
+    # with hash entries having key names encoded like settings[value1]
+    # (see doc at the end of HashedParams.pm)
+    # as well as modern application/json encoded hashes
+    if ($self->req->headers->content_type =~ /^application\/json/) {
+        $hp = decode_json($self->req->body);
```
Martchus, line 362: We should return a 400 error (and not die returning a 500 error) in case the JSON cannot be decoded.

lib/OpenQA/WebAPI/Controller/API/V1/Table.pm
```diff
@@ -352,6 +353,23 @@ Use by both B<create()> and B<update()> method.
 sub _prepare_settings {
     my ($self, $table, $entry) = @_;
     my $validation = $self->validation;
+    my $hp;
+    # accept both traditional application/x-www-form-urlencoded parameters
+    # with hash entries having key names encoded like settings[value1]
+    # (see doc at the end of HashedParams.pm)
+    # as well as modern application/json encoded hashes
+    if ($self->req->headers->content_type =~ /^application\/json/) {
```
Martchus, line 361: It would make sense to store the return value of `$self->req` in a variable instead of invoking this function over and over again (especially because it is even used in a loop).

assets/javascripts/admin_groups.js
```diff
@@ -366,9 +359,30 @@ function saveReorganizedGroups() {
...
 }
+
+function handleQuery(query) {
+  const url = query.url;
+  delete query.url;
+  const success = query.success;
+  delete query.success;
+  const error = query.error;
+  delete query.error;
+  var body = new FormData();
```
Martchus, line 376:
```suggestion
  const body = new FormData();
```

I think there are also some other variables which could be const or at least use let.

assets/javascripts/needleeditor.js
```diff
@@ -545,12 +500,49 @@ function saveNeedle(overwrite) {
...
+          response.success +=
+            " - <a href='" +
+            urlWithBase('/tests/' + response.developer_session_job_id) +
+            "#live'>back to live view</a>";
+        } else if (response.restart) {
+          response.success += " - <a href='#' data-url='" + response.restart + "' class='restart-link'>restart job</a>";
+        }
+        addFlash('info', response.success);
+      } else {
+        throw `<b>Unknown Error:</b><code>$response</code>`;
```
Martchus, line 531: In JavaScript you have to use curly brackets:
```suggestion
        throw `<b>Unknown Error:</b><code>${response}</code>`;
```

assets/javascripts/needleeditor.js
```diff
@@ -545,12 +500,49 @@ function saveNeedle(overwrite) {
...
+      }
+    })
+    .catch(error => {
+      console.error(error);
+      // add context to the error message unless it starts with an HTML tag (and is therefore assumed to be
+      // already nicely formatted)
+      if (`$error`.indexOf('<') !== 0) error = `<b>Fatal error when saving needle.</b><br>$error`;
```
Martchus, line 538: In JavaScript you have to use curly brackets. Additionally, `error` is a string anyway and you can use `startsWith`:
```suggestion
      if (!error.startsWith('<')) error = `<b>Fatal error when saving needle.</b><br>${error}`;
```

assets/javascripts/test_result.js
```diff
@@ -497,27 +497,30 @@ function loadTabPanelElement(tabName, tabConfig) {
...
+      console.error(error);
       const customRenderer = tabConfig.renderError;
       if (customRenderer) {
-        return customRenderer.call(tabConfig, xhr, ajaxOptions, thrownError);
+        return customRenderer.call(tabConfig, error);
       }
       tabPanelElement.innerHTML = '';
       tabPanelElement.appendChild(
-        document.createTextNode('Unable to load ' + (tabConfig.descriptiveName || tabName) + '.')
+        document.createTextNode('Unable to load ' + (tabConfig.descriptiveName || tabName) + `: ${error}`)
```
Martchus, line 521: It would be nicer to just use one quoting style at a time:
```suggestion
        document.createTextNode(`Unable to load ${tabConfig.descriptiveName || tabName}: ${error}`)
```

Martchus (changes requested):
I added a few inline comments. Also see my inline comments from yesterday.

This PR will also break `wait_for_ajax` in `SeleniumTest.pm`. Maybe we can create a wrapper around fetch that pushes the promises to some global array. Then `wait_for_ajax` could call `Promise.allSettled()` on this array. We could add the wrapper via `[mode==test]` and otherwise just a set it to the actual `window.fetch` function to avoid any overhead when we're not running tests.

lib/OpenQA/WebAPI/Controller/API/V1/Table.pm
asdil12: This is not in a loop.

assets/javascripts/needleeditor.js
asdil12: Error is not always a string. It can be an exception object (or whatever that is called in js). Therefore I'm converting it to a string as that exception doesn't have a `indexOf()` or `startsWith()` method

asdil12: Updated to use curly brackets and `startsWith`

lib/OpenQA/WebAPI/Controller/API/V1/Table.pm
Martchus: I see a loop below using `$self->req`.

assets/javascripts/admin_groups.js
Martchus: This is not what const means in JavaScript. You can "write to" a const variable as much as you like (unless it is an imitable type). const just means that the variable won't be reassigned.

assets/javascripts/admin_assets.js
```diff
@@ -121,37 +121,42 @@ function reloadAssetsTable() {
...
+      return response.json();
+    })
+    .then(response => {
+      if (response.error) throw response.error;
       addFlash(
         'info',
-        'The asset was deleted successfully. The asset table\'s contents are cached. Hence the removal is not immediately visible. To update the view use the "Trigger asset cleanup" button. Note that this is an expensive operation which might take a while.'
+        "The asset was deleted successfully. The asset table's contents are cached." +
+          'Hence the removal is not immediately visible. To update the view use the "Trigger asset cleanup" button.' +
+          'Note that this is an expensive operation which might take a while.'
```
b10n1k, line 135:
```suggestion
        `The asset was deleted successfully. The asset table's contents are cached.
          Hence the removal is not immediately visible. To update the view use the "Trigger asset cleanup" button.
          Note that this is an expensive operation which might take a while.`
```

assets/javascripts/admin_user.js
```diff
@@ -22,46 +22,43 @@ function setup_admin_user() {
...
     if (!confirm('Are you sure you want to delete this user?')) return;
 
-    $.ajax({
-      url: urlWithBase('/api/v1/user/' + id),
-      method: 'DELETE',
-      dataType: 'json',
-      success: function () {
+    fetchWithCSRF(urlWithBase('/api/v1/user/' + id), {method: 'DELETE'})
+      .then(response => {
+        if (response.status == 500)
```
b10n1k, line 46:
```suggestion
        if (response.status >= 500 && response.status < 600)
```

assets/javascripts/audit_log.js
```diff
@@ -56,16 +56,25 @@ function undoComments(undoButton) {
...
   undoButton.style.display = 'none';
-  $.ajax({
-    url: urlWithBase('/api/v1/comments'),
-    method: 'DELETE',
-    data: ids.map(id => `id=${id}`).join('&'),
-    success: () => addFlash('info', 'The comments have been deleted.'),
-    error: (jqXHR, textStatus, errorThrown) => {
+  var data = new FormData();
+  console.log('ids:');
+  console.log(ids);
```
b10n1k, line 61:
```suggestion
  console.log('ids: %s', ids);
```

assets/javascripts/comments.js
```diff
@@ -75,31 +71,30 @@ function updateComment(form) {
...
-    method: 'PUT',
-    data: $(form).serialize(),
-    success: () => {
-      $.ajax({
-        url: url,
-        method: 'GET',
-        success: response => {
+  fetchWithCSRF(url, {method: 'PUT', body: new FormData(form)})
+    .then(response => {
+      if (!response.ok) throw `Server returned ${response.status}: ${response.statusText}`;
```
b10n1k, line 76: To be honest I would rather use a more specific error. like `Request failed to update comment ${response.status}: ${response.statusText}`.

kalikiana: I agree. For debugging context will be useful to preserve here.

asdil12: But we do have this here. You have to look at the `.catch()`:
```js
    .catch(error => {
      window.alert(`The comment couldn't be updated : ${error}`);
    });

Comment (Martchus) [reply assets/javascripts/comments.js:76]: Right, then we really don't need it. I suppose we have to get used to this `throw ->carch(…)` pattern.

File: assets/javascripts/admin_assets.js (hunk 11 above)
Comment (asdil12) [reply assets/javascripts/admin_assets.js:135]: this would actually change the contents of the string. Also no reason to use ` over ' unless expanding vars.

File: t/lib/OpenQA/SeleniumTest.pm [hunk 15]
```diff
@@ -144,8 +144,14 @@ sub wait_for_ajax (%args) {
...
 
-    while (!$_driver->execute_script('return window.jQuery && jQuery.active === 0')) {
+    while (!$_driver->execute_script('return window.jQuery && jQuery.active === 0 && !window.runningFetchRequests')) {
         if ($timeout <= 0) {
+            $msg .= ' (jQuery.active='    # uncoverable statement
+              . $_driver->execute_script('window.jQuery && jQuery.active');    # uncoverable statement
+            $msg    # uncoverable statement
+              .= ',runningFetchRequests='
+              . $_driver->execute_script('window.runningFetchRequests')
+              . ')';    # uncoverable statement
```
Martchus, line 154: I'd avoid the separate script calls:
```suggestion
            my $s = '` (jQuery: ${window.jQuery && jQuery.active}, fetch: ${window.runningFetchRequests})`';  # uncoverable statement
            $msg .= $_driver->execute_script(s);  # uncoverable statement
```

Martchus (commented):
There is still one open inline-comment: https://github.com/os-autoinst/openQA/pull/5925/files#r1775493947

Otherwise this change is good to merge if tests are passing.

Martchus:
I know it's bad but this line needs to be marked as uncoverable (despite you have already marked the statement as uncoverable): https://app.codecov.io/gh/os-autoinst/openQA/pull/5925/blob/t/lib/OpenQA/SeleniumTest.pm#L149

Maybe it makes also sense to use `#<<< no perltidy` (and a closing `#>>> no perltidy`) in this block to prevent tidy from doing this stupid wrapping in the first place.

asdil12:
@Martchus done - let's hope that I don't have to add an `# uncoverable statement` to the `#<<< no perltidy` line as well (:

Martchus (approved):

okurz:
@mergifyio rebase

okurz:
Please merge when you can monitor

assets/javascripts/running.js
```diff
@@ -171,8 +175,9 @@ function updateTestStatus(newStatus) {
         reloadBrokenThumbnails(true);
       }
     })
-    .fail(function () {
+    .catch(error => {
       console.log('ERROR: modlist fail');
```
b10n1k, line 179: Do you need this as there is a console.error in the next line?

asdil12: It is good to see where it failed and afterwards also what the error was.

asdil12:
Rebased

lib/OpenQA/WebAPI/Controller/API/V1/Table.pm
```diff
@@ -352,6 +353,32 @@ Use by both B<create()> and B<update()> method.
...
+    my $error;
+    if ($self->req->headers->content_type =~ /^application\/json/) {
+        try {
+            $hp = decode_json($self->req->body);
+        }
+        catch {
+            $error = $_;
+        };
+        # of course stupid perl doesn't let you return directly from catch
+        # as that would actually lead to readable code
```
perlpunk, line 370: Totally unnecessary comment!

asdil12: Well it explains why the code needs an odd workaround around a missing language feature at this locaton.

But at least I didn't add a 20 line rant like the person writing the open source PSD implementation :P
https://github.com/zepouet/Xee-xCode-4.5/blob/master/XeePhotoshopLoader.m#L108

perlpunk: It could have explained that without saying "stupid perl"

kraih: This looks very unprofessional to me.

PR #5928, opened 2024-09-16
Ensure tidyall is available in CI after 28c5dd1

Martchus:
* Ensure `openQA-devel` depends on `perl(Code::TidyAll)` so that package is
  available in the CI
* Add the package only to style check dependencies (and not test
  dependencies) so the package build does not depend on it
* See https://progress.opensuse.org/issues/166817

---

Note that I have already prepared a [branch](https://github.com/os-autoinst/openQA/compare/master...Martchus:openQA:tidy?expand=1) to merge https://github.com/os-autoinst/os-autoinst-common/pull/55. For now I excluded this commit because we'll have to wait for the OBS package build on `devel:openQA` and the nightly job on CircleCI for this to become effective.

### DIFF
--- a/cpanfile
+++ b/cpanfile
@@ -114,6 +114,7 @@ on 'test' => sub {
 };
 
 on 'develop' => sub {
+    requires 'Code::TidyAll';
     requires 'Perl::Critic';
     requires 'Perl::Critic::Freenode';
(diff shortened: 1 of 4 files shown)


kalikiana (approved):
Good catch!

baierjan (approved):

PR #5933, opened 2024-09-17
Allow scheduling a (example) product via the web UI

Martchus:
Related ticket: https://progress.opensuse.org/issues/166658

---

~~Still a draft because I need to add UI tests and move the so far hard-coded data into `openqa.ini`. It already reads scenario definitions and certain parameters directly from the example distribution, though.~~ Of course this only works if there's a checkout of the example distribution. Maybe we should offer the user to create that checkout in case it is missing (which wouldn't be very hard to do; we'd just add a link on the web UI that'll internally spawn a checkout Minion job we recently implemented anyway).

---

This is how the menu, the form and success/error messages look like in one screenshot:

grafik

### DIFF
--- a/etc/openqa/openqa.ini
+++ b/etc/openqa/openqa.ini
@@ -325,3 +325,11 @@ concurrent = 0
 # home =
 # username = openqa-user
 # ssh_key_file = ~/.ssh/id_rsa
+
+# Override form values for creating example test
+#[test_preset example]
+#title = Create example test
+#info = Some info that will show up on the "Create … -> Example test" page
+#casedir = https://github.com/os-autoinst/os-autoinst-distri-example.git
+#distri = example
+#build = openqa
--- a/lib/OpenQA/Setup.pm
+++ b/lib/OpenQA/Setup.pm
@@ -247,10 +247,7 @@ sub read_config ($app) {
         }
         for my $k (@known_keys) {
             my $v = $cfg && $cfg->val($section, $k);
-            $v
-              //= exists $mode_defaults{$app->mode}{$section}->{$k}
-              ? $mode_defaults{$app->mode}{$section}->{$k}
-              : $defaults{$section}->{$k};
+            $v //= $mode_defaults{$app->mode}{$section}->{$k} // $defaults{$section}->{$k};
             $config->{$section}->{$k} = trim $v if defined $v;
         }
     }
--- a/lib/OpenQA/Setup.pm
+++ b/lib/OpenQA/Setup.pm
@@ -198,8 +198,15 @@ sub read_config ($app) {
         influxdb => {
             ignored_failed_minion_jobs => '',
         },
-        carry_over => \%CARRY_OVER_DEFAULTS
-    );
+        carry_over => \%CARRY_OVER_DEFAULTS,
+        'test_preset example' => {
+            title => 'Create example test',
+            info => 'Parameters to create an example test have been pre-filled in the following form. '
+              . 'You can simply submit the form as-is to test your openQA setup.',
+            casedir => 'https://github.com/os-autoinst/os-autoinst-distri-example.git',
+            distri => 'example',
+            build => 'openqa',
+        });
 
     # in development mode we use fake auth and log to stderr
     my %mode_defaults = (
--- a/lib/OpenQA/WebAPI/Controller/Test.pm
+++ b/lib/OpenQA/WebAPI/Controller/Test.pm
@@ -135,14 +135,21 @@ sub _load_test_preset ($self, $preset_key) {
 
 sub _load_scenario_definitions ($self, $preset) {
     return undef if exists $preset->{scenario_definitions};
-    return undef unless my $casedir = testcasedir($preset->{distri}, $preset->{version});
+    return undef unless my $distri = $preset->{distri};
+    return undef unless my $casedir = testcasedir($distri, $preset->{version});
     my $defs_yaml = eval { path($casedir, 'scenario-definitions.yaml')->slurp('UTF-8') };
-    $preset->{scenario_definitions} = $defs_yaml;
-    return $self->stash(flash_error => "Unable to read scenario definitions for the specified preset: $@") if $@;
+    if (my $error = $@) {
+        return $self->stash(flash_error => "Unable to read scenario definitions for the specified preset: $error")
+          unless $error =~ /no.*file/i;
+        my $info = Mojo::ByteStream->new(
+            qq(You first need to <a href="#" onclick="cloneTests(this)">clone the $distri test distribution</a>.));
+        return $self->stash(flash_info => $info);
+    }
     my $defs = eval { load_yaml(string => $defs_yaml) };
     return $self->stash(flash_error => "Unable to parse scenario definitions for the specified preset: $@") if $@;
     my $e = join("\n", @{$self->app->validate_yaml($defs, 'JobScenarios-01.yaml')});
     return $self->stash(flash_error => "Unable to validate scenarios definitions of the specified preset:\n$e") if $e;
+    $preset->{scenario_definitions} = $defs_yaml;
     return undef unless my @products = values %{$defs->{products}};
     return undef unless my @job_templates = keys %{$defs->{job_templates}};
     $preset->{$_} //= $products[0]->{$_} for qw(distri version flavor arch);
@@ -160,6 +167,19 @@ sub create ($self) {
         $self->stash(flash_error => "The specified preset '$preset_key' does not exist.");
     }
     $self->stash(preset => ($preset // {}));
+}
+
+sub clone ($self) {
+    my $preset = $self->_load_test_preset($self->param('preset'));
+    return $self->render(status => 400, text => 'unable to find preset') unless defined $preset;
+    return $self->render(status => 400, text => 'preset has no distri') unless my $distri = $preset->{distri};
+    return $self->render(status => 400, text => 'preset has no casedir') unless my $casedir = $preset->{casedir};
+    $self->gru->enqueue_and_keep_track(
+        task_name => 'git_clone',
+        task_description => 'cloning test distribution',
+        task_args => {testcasedir($distri, $preset->{version}) => $casedir}    # uncoverable statement
+      )->then(sub ($result) { $self->render(json => $result) })    # uncoverable statement
+      ->catch(sub ($error, @) { $self->reply->gru_result($error, 400) });
 }
 
 sub get_match_param {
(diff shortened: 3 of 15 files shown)


okurz:
overall this looks quite good. I was thinking of alternatives where to place the action but your proposal so far looks like the best to me. Just with the "+" I was thinking of a bigger button, possibly blue, with a frame around to distinguish clearly from the caret

Martchus:
I'm pondering about configurability. Reading the structure that has currently the FIXME comment from `openqa.ini` doesn't really work because it needs nesting but we only use INI. I suppose I could simply read those values from an additional file (e.g. `test_presets.ini`). Otherwise I could check for sections like `[test_presets/example]` in `openqa.ini` but it'll be an additional loop.

Martchus:
I now made it configurable. So in this screenshot I overrode the title by putting a `[test_presets/example]` section to my `openqa.ini`:

screenshot_20240918_143341

This screenshot also shows that it is now also possible to add additional settings. I think this is important if we want to offer this as a generic form (and the implementation is also quite simple).

This leaves tests and the possibility to clone the example distribution automatically (for the scenario definitions).

perlpunk:
> I'm pondering about configurability. Reading the structure that has currently the FIXME comment from `openqa.ini` doesn't really work because it needs nesting but we only use INI. I suppose I could simply read those values from an additional file (e.g. `test_presets.ini`). Otherwise I could check for sections like `[test_presets/example]` in `openqa.ini` but it'll be an additional loop.
> 
> EDIT: I guess I'll go for `[test_presets/example]` because I suppose the additional loop can be avoided.

Config::IniFiles allows groups. You separate the group from the section with a space.
https://metacpan.org/pod/Config::IniFiles#Groups

Martchus:
> Config::IniFiles allows groups. You separate the group from the section with a space.
> https://metacpan.org/pod/Config::IniFiles#Groups

I now changed it to use e.g. `[test_preset foo]`. This way we could use the grouping mechanism of the INI parser - even though this is not necessary for the PR in its current form. With this we could use https://metacpan.org/pod/Config::IniFiles#GroupMembers-($group) to show the list of available presets in e.g. the navbar menu. (Not implemented yet. For now I stick to what we have planned in the ticket. So cloning the example repo automatically is probably more important.)

kalikiana:
> Still a draft because I need to add UI tests and move the so far hard-coded data into openqa.ini. It already reads scenario definitions and certain parameters directly from the example distribution, though. Of course this only works if there's a checkout of the example distribution. Maybe we should offer the user to create that checkout in case it is missing (which wouldn't be very hard to do; we'd just add a link on the web UI that'll internally spawn a checkout Minion job we recently implemented anyway).

Is this obsolete? As this is not a draft anymore 😉

kalikiana (approved):
Looks really nice. Especially with the help texts. And I appreciate your making it extensible in the future.

Martchus:
I crossed it the obsolete part.

lib/OpenQA/WebAPI/Controller/Test.pm
```diff
@@ -116,6 +117,51 @@ sub list {
...
+    # avoid reading INI file again on subsequent calls
+    state %presets;
+    return $presets{$preset_key} if exists $presets{$preset_key};
+    $presets{$preset_key} = undef;
+    # read preset from an INI section [test_presets/…] or fallback to defaults assigned on setup
+    my $config = $self->app->config;
+    return undef unless my $ini_config = $config->{ini_config};
+    my $ini_key = "test_preset $preset_key";
+    return $presets{$preset_key}
+      = $ini_config->SectionExists($ini_key)
```
perlpunk, line 131: I still don't like that we are creating some extra code for this specific feature to handle nested data, while we could do this in a more generic way in Setup.pm with the group feature.
Also it moves any validation we might want to do to runtime instead of server startup.
Since it's something internal we can probably change the implementation later, but just saying.
Not sure what others think.

okurz:
> I crossed out the obsolete part.
> 
> Considering there's remaining work to do I'm still considering this as not ready. We don't want to show a new option prominently that'll not work out of the box.

Consider a feature switch which should be off by default until we stabilized the feature

Martchus:
I now implemented cloning the example test distribution. So when the user enters the page and the repo hasn't been cloned yet it says "You have to **clone the example test distribution** first." where the bold part is a link, When clicking on the link one gets a loading indication and after a few seconds (it really doesn't take long to checkout the example distribution) the form reloads with all values filled in. (And in any error case the error is shown and one can retry.) This didn't require adding much code because we already have Minion tasks for cloning from Git now.

perlpunk:
Btw, we have to be careful to not do this on o3 or osd until fetchneedles is replaced. The example distri has a default branch `main`, and fetchneedles can't handle different default branches.

etc/openqa/openqa.ini
```diff
@@ -325,3 +325,8 @@ concurrent = 0
 # home =
 # username = openqa-user
 # ssh_key_file = ~/.ssh/id_rsa
+
+[test_preset bar]
+title = Some preset
+distri = does-not-exist
+casedir = http://foo.git
```
okurz, line 332: Those should be commented by default. And can you add some explanation comments here? I wouldn't know why I would want to set those settings. And please use the actual default values

lib/OpenQA/WebAPI/Controller/Test.pm
```diff
@@ -116,6 +117,71 @@ sub list {
...
+sub create ($self) {
+    my $preset_key = $self->param('preset');
+    my $preset = $self->_load_test_preset($preset_key);
+    if (defined $preset) {
+        $self->stash(flash_info => $preset->{info});
+        $self->_load_scenario_definitions($preset);
+    }
+    elsif (defined $preset_key) {
+        $self->stash(flash_error => "The specified preset '$preset_key' does not exist.");
+    }
```
okurz, line 168: Can this be a

```suggestion
    return $self->stash(flash_error => "The specified preset '$preset_key' does not exist.") if defined $preset_key;
    if (defined $preset) {
        $self->stash(flash_info => $preset->{info});
        $self->_load_scenario_definitions($preset);
    }
```

although I wonder why should we error out if `$preset_key` is defined and not if it's *not* defined?

t/ui/29-create_tests.t
```diff
@@ -0,0 +1,89 @@
+#!/usr/bin/env perl
+# Copyright 2024 SUSE LLC
```
okurz, line 2:
```suggestion
# Copyright SUSE LLC
```

okurz (changes requested):
Some inline comments.

And have you considered https://github.com/os-autoinst/openQA/pull/5933#issuecomment-2361040754 ?

etc/openqa/openqa.ini
Martchus: This should have been added to the config file we use for tests. I didn't intend to touch the normal config template at all for now.

Martchus:
> And have you considered https://github.com/os-autoinst/openQA/pull/5933#issuecomment-2361040754 ?

I only read that comment after I implemented the most recent changes I pushed to this PR yesterday. I don't think it is useful to merge this feature without enabling it by default because it is about the out of the box experience.

okurz:
> > And have you considered [#5933 (comment)](https://github.com/os-autoinst/openQA/pull/5933#issuecomment-2361040754) ?
> 
> I only read that comment after I implemented the most recent changes I pushed to this PR yesterday. I don't think it is useful to merge this feature without enabling it by default because it is about the out of the box experience.

I know but do you trust it enough to be enabled on all production instances? After all the primary goal was to show that easy starting point on *clean jobless* instances

Martchus:
> I know but do you trust it enough to be enabled on all production instances?

The feature itself I'd generally consider stable. It has full test coverage. The checkout of the example distribution relies on the Git cloning we implemented a while ago and I don't think it is generally problematic.

 I suppose the only problem is fetchneedles cannot handle the use of `main` as main branch (as opposed to `master`). At least according to @perlpunk. So I'll have a look at the situation. Maybe that's easy to fix or we can ignore it until fetchneedles is replaced or I can do a manual checkout on our production instances that uses and alias for master.

Martchus:
I added another commit to avoid problems with `fetchneedles`. See the commit message for details. The short story is that it'll work on our production instances and I also did local test runs of the entire script. There's only one caveat which I don't find problematic but I added a comment in the `fetchneedles` script documenting it.

kalikiana (approved):

script/fetchneedles
```diff
@@ -72,24 +72,40 @@ deal_with_stale_lockfile() {
     fi
 }
 
+# Determine the "main branch" for the remote origin unless a branch has been specified explicitly
+# note: This relies on a file called `refs/remotes/origin/HEAD` with contents like `ref: refs/remotes/origin/master` being
+#       present under the `.git` direcgtory. This should be the case for all recently cloned Git repositories. Otherwise
```
kalikiana, line 77:
```suggestion
#       present under the `.git` directory. This should be the case for all recently cloned Git repositories. Otherwise
```

t/data/openqa.ini
```diff
@@ -1 +0,0 @@
-../../etc/openqa/openqa.ini
```
okurz, line 1: why?

okurz (approved):
Looks very nice now. Fix the code style issues and I am sure I can approve.

okurz (changes requested):
Looks very nice now. Fix the code style issues and I am sure I can approve.

t/data/openqa.ini
Martchus: Because I had to add config values for testing instead of just testing with the default config. (We want to cover the part of the code that allows the users/admins to override values so we obviously need to supply this kind of configuration when running the test.)

lib/OpenQA/Setup.pm
```diff
@@ -198,8 +198,15 @@ sub read_config ($app) {
...
-    );
+        carry_over => \%CARRY_OVER_DEFAULTS,
+        'test_preset example' => {
+            title => 'Create example test',
+            info => 'Parameters to create an example test have been pre-filled in the following form. '
+              . 'You can simply submit the form as-is to test your openQA setup.',
+            casedir => 'https://github.com/os-autoinst/os-autoinst-distri-example.git',
+            distri => 'example',
+            build => 'openqa',
+        });
```
okurz, line 209: Where did the according example changes to the openqa.ini file go? We should still have the according sections pre-filled and commented out for all possible options

okurz (changes requested):

t/data/openqa.ini
perlpunk: In other tests we just change the config on the fly to test specific features

okurz (approved):

PR #5936, opened 2024-09-23
Fix checks for worker classes in directly chained clusters

Martchus:
See particular commit messages and https://progress.opensuse.org/issues/154735

### DIFF
--- a/lib/OpenQA/Schema/ResultSet/JobSettings.pm
+++ b/lib/OpenQA/Schema/ResultSet/JobSettings.pm
@@ -6,6 +6,11 @@ use Mojo::Base 'DBIx::Class::ResultSet', -signatures;
 
 use OpenQA::App;
 
+sub all_values_sorted ($self, $job_id, $key) {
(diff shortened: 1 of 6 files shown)


lib/OpenQA/Schema/ResultSet/JobSettings.pm
```diff
@@ -6,6 +6,11 @@ use Mojo::Base 'DBIx::Class::ResultSet', -signatures;
 
 use OpenQA::App;
 
+sub all_values_sorted ($self, $job_id, $key) {
+    state $options = {distinct => 1, columns => 'value', order_by => 'value'};
```
perlpunk, line 10: Why are you using `state` here?

Martchus: Because this hash never changes, so no need to re-initialize it on every function call.

perlpunk: It's a bit uncommon for a simple, small hashref, but I won't complain

kalikiana (approved):

PR #5941, opened 2024-09-23
Remove deprecated legacy job group view

kalikiana:
The old view has been deprecated 5 years ago.

Related: https://progress.opensuse.org/issues/166310#note-14


okurz (changes requested):
See comments by Martchus

kalikiana:
> Did you check whether any of the job groups on our production instances still use the old approach? If that would be the case, will they be force-migrated?

I [asked for feedback in Slack](https://suse.slack.com/archives/C02CANHLANP/p1727103681707079), as mentioned in the ticket.

okurz:
I found https://openqa.fedoraproject.org/admin/job_templates/1 and https://openqa.debian.net/admin/job_templates/10 still using the table based view. @phil-hands @AdamWill @marmarek WDYT about the proposed change and enforcing the YAML editor for job group settings?

marmarek:
I'm fine with that. I think Qubes uses YAML everywhere already (if not, that's an oversight).

asdil12:
As discussed today, I'm fine with merging this anyways if it will simply switch over existing instances seamlessly to using the yaml editor without losing data.

AdamWill:
What is the "old view" and what is the "new view"?

We don't typically use any kind of web UI editing of the job templates in Fedora; we work on them as text in my FIF interchange format in https://pagure.io/fedora-qa/os-autoinst-distri-fedora - see https://pagure.io/fedora-qa/os-autoinst-distri-fedora/blob/main/f/templates.fif.json and https://pagure.io/fedora-qa/os-autoinst-distri-fedora/blob/main/f/templates-updates.fif.json . https://pagure.io/fedora-qa/os-autoinst-distri-fedora/blob/main/f/fifloader.py converts to the format expected by load_templates and calls load_templates on it.

phil-hands:
Adam Williamson ***@***.***> writes:

> What is the "old view" and what is the "new view"?
>
> We don't typically use any kind of web UI editing of the job templates in Fedora; we work on them as text in my FIF interchange format in https://pagure.io/fedora-qa/os-autoinst-distri-fedora - see https://pagure.io/fedora-qa/os-autoinst-distri-fedora/blob/main/f/templates.fif.json and https://pagure.io/fedora-qa/os-autoinst-distri-fedora/blob/main/f/templates-updates.fif.json . https://pagure.io/fedora-qa/os-autoinst-distri-fedora/blob/main/f/fifloader.py converts to the format expected by load_templates and calls load_templates on it.

We (debian) currently also use fifloader. I've not yet experimented with
moving our config into the git repo (which seems to be the way forward).

Once I manage to update to a version that includes the caching of git
repos, I'll be able to experiment.  At present if I use CASEDIR more than
a little salsa.debian.org (our gitlab instance) decides I'm conducting a
DoS and blocks the workers. :-/

I have an update that's mostly ready, so hopefully will be able to sort
that out shortly.

Cheers, Phil.
-- 
Philip Hands -- https://hands.com/~phil

kalikiana:
> What is the "old view" and what is the "new view"?

This is the old view: https://openqa.fedoraproject.org/admin/job_templates/1

image

As opposed to the YAML editor:

image

Martchus:
The script mentioned by @AdamWill will maybe break if we proceed with this (because if you have once submitted YAML you cannot update job templates anymore; the script uses `load_templates` which has *some* support for using YAML but I'm not sure whether this use of it will just work without modifications).

AdamWill:
to be clear, what would be easiest for us is if we can continue to just use `load_templates` to completely re-load the templates when they change (which is what we do, we don't bother with the 'updating' features). if that works, we do not care what the web UI edit interface looks like. if it doesn't, we would need to figure out how to change to the new thing.

marmarek:
FWIW I used to do something similar (didn't used the web ui for job templates, and instead re-imported generated input each time), but nowadays I'm using YAML version with web UI and I'd say it's better. But also, I have a bit less templates, so maybe there is some duplication that could be avoided by keeping external generator.

Anyway, with the removal of legacy job group, I think any of such generator probably needs to be changed to generate YAML content instead of json dict for job groups.

AdamWill:
I wrote the intermediate format and parser before the current YAML stuff existed, because working with the previous format in text was painful. If it's necessary to adapt it we can, it's just work for no obvious benefit, which always sucks.

marmarek:
> because working with the previous format in text was painful

On this I agree...

PR #5946, opened 2024-09-25
Add option to restart jobs upon comment submission

b10n1k:
This commit enhances the comment submission functionality with an integrated job restart option, improving workflow efficiency for users managing multiple jobs.

For this `restart_job` helper added in `Helpers.pm` and `_restart` in `Job.pm` had to be refactored to encapsulate the logic and the reusability of the feature(restart_jobs) in multiple places within the application.

https://progress.opensuse.org/issues/166559


assets/javascripts/overview.js
```diff
@@ -289,6 +289,7 @@ function showAddCommentsDialog() {
 
 function addComments(form) {
   const text = form.elements.text.value;
+  let wannaRestart = document.getElementById('wannaRestartCheck').checked ? 1 : 0;
```
okurz, line 292:
```suggestion
  let restartRequested = document.getElementById('restartRequestedCheck').checked ? 1 : 0;
```

and all other occurences in this PR accordingly including the Perl variables please

lib/OpenQA/WebAPI/Controller/API/V1/Comment.pm
```diff
@@ -191,12 +193,24 @@ sub create_many ($self) {
...
 
+    if ($wanna_restart && $wanna_restart == 1) {
+        for my $job_id (@$job_ids) {
+            my ($res, $jobs, $auto, $single_job_id, $dup_route);
+            my %args = (jobs => $job_id);
+            $self->param('jobid', $job_id);
+            eval { ($res, $jobs, $auto, $single_job_id, $dup_route) = $self->restart_job(\%args); };
+            if ($@) {
+                push @failed_restart, {job_id => $job_id, error => "Failed to restart job: $@"};
+            }
```
okurz, line 204:
```suggestion
            push @failed_restart, {job_id => $job_id, error => "Failed to restart job: $@"} if $@;
```

templates/webapi/test/overview.html.ep
```diff
@@ -231,10 +231,17 @@
...
                     <div id="add-comments-controls">
-                        <button type="submit" class="btn btn-danger">
-                            <i class="fa fa-comment"></i> Submit comment on all <%= scalar @$job_ids %> jobs
-                        </button>
-                        <button type="button" class="btn btn-secondary" data-bs-dismiss="modal">Discard</button>
+		      <div class="form-check my-1">
+		      <input class="form-check-input"
+			     type="checkbox" value="" id="wannaRestartCheck">
+		      <label class="form-check-label" for="wannaRestartCheck">
+			Restart on Submit
```
okurz, line 238:
```suggestion
			Also restart job(s) on submit
```

okurz (changes requested):

templates/webapi/test/overview.html.ep
b10n1k: Can I avoid `also`? `Restart job(s) on submit` sound sufficient, no?

assets/javascripts/overview.js
b10n1k: Done https://github.com/os-autoinst/openQA/commit/2c24af9fbb696e60daaca9c2e4b11ffe7298dbc4

lib/OpenQA/WebAPI/Controller/API/V1/Comment.pm
b10n1k: Done https://github.com/os-autoinst/openQA/commit/2c24af9fbb696e60daaca9c2e4b11ffe7298dbc4

lib/OpenQA/WebAPI/Controller/API/V1/Comment.pm
```diff
@@ -191,12 +193,22 @@ sub create_many ($self) {
         push @failed, {job_id => $job_id} if $@;
     }
 
+    if ($wanna_restart && $wanna_restart == 1) {
+        for my $job_id (@$job_ids) {
+            my ($res, $jobs, $auto, $single_job_id, $dup_route);
+            my %args = (jobs => $job_id);
+            $self->param('jobid', $job_id);
+            eval { ($res, $jobs, $auto, $single_job_id, $dup_route) = $self->restart_job(\%args); };
```
perlpunk, line 201: The `_restart` method already supports multiple jobs as parameters.
The documentation for the `/restart` route says "Restart job(s)."

Calling this for every single job and validating the same request parameters again and again is not necessary.
Also you are passing something in `%args` but you never read that in the function.

And there is already one function that restarts one job, that could also be called, instead of calling the whole `_restart` thing:
```
my $res = OpenQA::Resource::Jobs::job_restart($jobs, @params);
```

perlpunk: @Martchus and that's actually also what I meant earlier today: `_restart` is a complete endpoint:
```
sub restart ($self) { $self->_restart }
```
making a helper that implements a complete endpoint, doesn't make much sense to me.

templates/webapi/test/overview.html.ep
okurz: I see the "also" as necessary to make it clear that the checkbox wouldn't "restart instead of commenting" but "restart *and* comment". But after the discussion we just had I suggest to actually have two buttons:
1. Left, less-alerting button for "comment", e.g. in light orange
2. Right, more-alerting button for "Comment and restart", e.g. the current red color

kalikiana:
Obosoleted by #5971 I guess?

PR #5947, opened 2024-09-25
Fix quoting of git ssh arguments

perlpunk:
Error detecting remote default branch name for "some git":
        "ssh -oBatchMode=yes": ssh -oBatchMode=yes: command not found

Since we are passing an array ref to IPC::Run::run, we should not add quotes. git is trying to call `"ssh ..."` including the quotes.

Issues:
* https://progress.opensuse.org/issues/164898
* https://progress.opensuse.org/issues/164886

### DIFF
--- a/lib/OpenQA/Git.pm
+++ b/lib/OpenQA/Git.pm
@@ -32,7 +32,7 @@ sub _run_cmd ($self, $args, $options = {}) {
     my $include_git_path = $options->{include_git_path} // 1;
     my $ssh_batchmode = $options->{ssh_batchmode} // 0;
     my @cmd;
-    push @cmd, 'env', 'GIT_SSH_COMMAND="ssh -oBatchMode=yes"' if $ssh_batchmode;
+    push @cmd, 'env', 'GIT_SSH_COMMAND=ssh -oBatchMode=yes' if $ssh_batchmode;
     push @cmd, $self->_prepare_git_command($include_git_path), @$args;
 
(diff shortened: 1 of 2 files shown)


perlpunk:
I tested this locally

okurz (approved):

lib/OpenQA/Git.pm
```diff
@@ -32,7 +32,7 @@ sub _run_cmd ($self, $args, $options = {}) {
     my $include_git_path = $options->{include_git_path} // 1;
     my $ssh_batchmode = $options->{ssh_batchmode} // 0;
     my @cmd;
-    push @cmd, 'env', 'GIT_SSH_COMMAND="ssh -oBatchMode=yes"' if $ssh_batchmode;
+    push @cmd, 'env', 'GIT_SSH_COMMAND=ssh -oBatchMode=yes' if $ssh_batchmode;
```
kalikiana, line 35: Can we have unit test coverage for this?

perlpunk: I thought about it, but no good idea yet.
e.g. if we actually ran git on temporary git repositories created during unit tests, that would be an improvement, but we would use local git repositories in paths, not remote ones, I guess, so the `ssh ...` stuff would never be called.
Ideas welcome.
In any case, we should deploy this fix.

perlpunk:
ok, I need to fix the tests of course

Martchus (approved):

PR #5948, opened 2024-09-25
t: Improve mocked commands

perlpunk:
To be able to have valid git commands in our test (e.g. for copy/paste) we now at least quote any argument that has a space in it.

Related issue: https://progress.opensuse.org/issues/164898


okurz (approved):
Good. No need to force merge. I would prefer to have https://progress.opensuse.org/issues/167272 fixed first

PR #5951, opened 2024-09-26
Skip symlinks in enqueue_git_update_all

perlpunk:
I forgot that `-d` is actually true for symlinks.

It resulted in the gru task processing over the same realpaths, trying
to get a guard for the same path multiple times, which failed and ran
int an endless retry.

I tested it locally with symlinks and could now reproduce the problem. It tries to get a guard for the same directory multiple times and runs into an endless retry.


okurz (approved):
I guess you need to extend tests according to the codecov report

perlpunk:
> I guess you need to extend tests according to the codecov report

No, codecov just didn't run on my force push yet

perlpunk:
I reduced this PR to only the first commit about using `-l`.

okurz:
@mergifyio rebase

PR #5952, opened 2024-09-26
Improve OpenID auth handling on not_openid response

b10n1k:
Enhanced error handling by including OpenID provider URI in error messages for
invalid data. The error seems sporadic and as part of the research on
https://progress.opensuse.org/issues/167266 we can consider this not a severe
problem. Because of the sensitivity of the arguments, it is not feasible to
add any info from the available data for later debbugging. But at least now
we can tell which server cause problems.

### DIFF
--- a/lib/OpenQA/WebAPI/Auth/OpenID.pm
+++ b/lib/OpenQA/WebAPI/Auth/OpenID.pm
@@ -108,8 +108,10 @@ sub auth_response ($c) {
     };
 
     $csr->handle_server_response(
-        not_openid =>
-          sub () { $err_handler->('Failed to login', 'OpenID provider returned invalid data. Please retry again') },
+        not_openid => sub () {
+            my $op_uri = $params{'openid.op_endpoint'} // '';
+            $err_handler->('Failed to login', "OpenID provider '$op_uri' returned invalid data. Please retry again");
+        },
         setup_needed => sub ($setup_url) {
             # Redirect the user to $setup_url
             $setup_url = URI::Escape::uri_unescape($setup_url);
@@ -117,9 +119,10 @@ sub auth_response ($c) {
 
             return (redirect => $setup_url, error => 0);
         },
-        cancelled => sub () { },    # Do something appropriate when the user hits "cancel" at the OP
-        verified => sub ($vident) { _handle_verified($c, $vident) },
-        error => sub (@args) { $err_handler->(@args) },
+        # Do something appropriate when the user hits "cancel" at the OP
+        cancelled => sub () { },    # uncoverable statement
+        verified => sub ($vident) { _handle_verified($c, $vident) },    # uncoverable statement
+        error => sub (@args) { $err_handler->(@args) },    # uncoverable statement
     );
 
     return (redirect => decode_base64url($csr->args('return_page'), error => 0)) if $csr->args('return_page');
--- a/t/03-auth-openid.t
+++ b/t/03-auth-openid.t
@@ -2,6 +2,7 @@
 # SPDX-License-Identifier: GPL-2.0-or-later
 
 use Test::Most;
+use Mojo::Base -signatures;
 use Test::Warnings ':report_warnings';
 use Test::MockModule;
 use Test::MockObject;
@@ -26,11 +27,30 @@ ok OpenQA::WebAPI::Auth::OpenID::_handle_verified($c, $vident), 'can call _handl
 $users->called_ok('create_user', 'new user is created for initial login');
 is(($users->call_args(2))[1], 'mordred', 'new user created with details');
 $c->set_always(
-    req => Test::MockObject->new->set_always(params => Test::MockObject->new->set_always(pairs => [1, 2]))
-      ->set_always(url => Test::MockObject->new->set_always(base => 'openqa')))
-  ->set_always(app => Test::MockObject->new->set_always(config => {})
-      ->set_always(log => Test::MockObject->new->set_true('error', 'debug')))->set_true('flash');
+    req => Test::MockObject->new->set_always(
+        params => Test::MockObject->new->set_always(pairs => ['openid.op_endpoint', 'https://www.opensuse.org/openid/'])
+    )->set_always(url => Test::MockObject->new->set_always(base => 'openqa')))
+  ->set_always(
+    app => Test::MockObject->new->set_always(config => {})->set_always(log => Test::MockObject->new->set_true('error')))
+  ->set_true('flash');
 is OpenQA::WebAPI::Auth::OpenID::auth_response($c), 0, 'can call auth_response';
 $c->app->log->called_ok('error', 'an error was logged for call without proper config');
 
+my $mock_openid_consumer = Test::MockModule->new('Net::OpenID::Consumer');
+$mock_openid_consumer->redefine(
+    'handle_server_response',
+    sub ($self, %res_handlers) {
+        return $res_handlers{setup_needed}
+          ? $res_handlers{setup_needed}->("https://www.opensuse.org/openid/setup")
+          : undef;
+    });
+$c->set_always(
+    req => Test::MockObject->new->set_always(
+        params => Test::MockObject->new->set_always(pairs => ['openid.op_endpoint', 'https://www.opensuse.org/openid/'])
+    )->set_always(url => Test::MockObject->new->set_always(base => 'openqa')))
+  ->set_always(
+    app => Test::MockObject->new->set_always(config => {})->set_always(log => Test::MockObject->new->set_true('debug')))
+  ->set_true('flash');
+is OpenQA::WebAPI::Auth::OpenID::auth_response($c), 0, 'can handle setup_needed response';
+$c->app->log->called_ok('debug', 'a debug messgae is logged when setup_needed respond');
 done_testing;


lib/OpenQA/WebAPI/Auth/OpenID.pm
```diff
@@ -108,8 +108,10 @@ sub auth_response ($c) {
     };
 
     $csr->handle_server_response(
-        not_openid =>
-          sub () { $err_handler->('Failed to login', 'OpenID provider returned invalid data. Please retry again') },
+        not_openid => sub () {
+            my $txt = 'Please retry again. Likely, OpenID message is not considered a valid object. It is missing, at least, "openqa.mode" value';
```
nicksinger, line 112: I assume we cannot give further suggestions to server operators reading this message? Should they get in touch with somebody? Just ignore it?

lib/OpenQA/WebAPI/Auth/OpenID.pm
```diff
@@ -108,8 +108,10 @@ sub auth_response ($c) {
...
+            log_warning("Invalid OpenID provided. $txt");
```
perlpunk, line 113: Before `$err_handler` was called, which calls `$c->flash(error => "$err: $txt")`.
I'm not sure if that's something the user would see.
But if it is, now the user wouldn't even see an error anymore?

lib/OpenQA/WebAPI/Auth/OpenID.pm
kalikiana: The API doesn't seem to provide anything else here: [Net::OpenID::Consumer](https://metacpan.org/pod/Net::OpenID::Consumer).

lib/OpenQA/WebAPI/Auth/OpenID.pm
b10n1k: I dont know as well. But I guess it is fine to add this line to be consistent?

lib/OpenQA/WebAPI/Auth/OpenID.pm
b10n1k: it is not very clear what cause the invalid args. Usually retry helps, right? 
As @okurz suggested on slack[0] maybe we can include some data from controller.

[0] https://suse.slack.com/archives/C02AJ1E568M/p1727358701821639?thread_ts=1727357210.018199&cid=C02AJ1E568M

lib/OpenQA/WebAPI/Auth/OpenID.pm
Martchus: I think the flash message is something the user would see in certain cases. (It probably depends on the page the user is getting back to.)

I'd probably try to keep the flash message, and maybe also as an error - because this is in fact why the login attempt failed.

Martchus (changes requested):
The test still needs to be implemented and we cannot accept commented-out code.

lib/OpenQA/WebAPI/Auth/OpenID.pm
b10n1k: I added `flash` on 06f453ea6

b10n1k: But I cant see any popup for warn or even debug. So I reverted it

b10n1k:
> The test still needs to be implemented and we cannot accept commented-out code.

obviously. I just forgot it like that before submit

okurz (changes requested):
In the commit message: 
Decrease not_openid log level and provide an better explaination
s/an/a/ and s/explaination/explanation/

b10n1k:
> > The test still needs to be implemented and we cannot accept commented-out code.
> 
> obviously. I just forgot it like that before submit

Test fixed

b10n1k:
> In the commit message: Decrease not_openid log level and provide an better explaination s/an/a/ and s/explaination/explanation/

Done

b10n1k:
I have tried to fix the coverage but so far I cant make it work. I follow https://openid.net/specs/openid-authentication-1_1.html#mode_checkid_immediate to find what it needs.

lib/OpenQA/WebAPI/Auth/OpenID.pm
b10n1k: > I don't think the retry hint is useful in that logs as it's not directed to users

removed

lib/OpenQA/WebAPI/Auth/OpenID.pm
Martchus: I would say the flash message should stay at error-level. As I said, for the single request this is a fatal error. Just from the point of view of the server admin it is nothing one can do much about so a warning makes more sense.

b10n1k: > I would say the flash message should be kept and stay at error-level. As I said, for the single request this is a fatal error. Just from the point of view of the server admin it is nothing one can do much about so a warning makes more sense.

Do you ask for a change or is it just a confirmation statement?

Martchus: I ask for adding the flash message back (as error, while logging only a warning/info message on the server logs). That's at least what I think is best, not sure what the rest of the team thinks.

lib/OpenQA/WebAPI/Auth/OpenID.pm
```diff
@@ -108,18 +108,22 @@ sub auth_response ($c) {
     };
 
     $csr->handle_server_response(
-        not_openid =>
-          sub () { $err_handler->('Failed to login', 'OpenID provider returned invalid data. Please retry again') },
+        not_openid => sub () {
+            my $txt
+              = 'Likely, OpenID message is not considered a valid object. It is missing, at least, "openqa.mode" value';
```
okurz, line 113: I don't understand the "Likely". What do you mean by that?

b10n1k: the `not_openid` is set undef in different conditions. In the `IndirectMessage.pm` you will find that return undef in a single `$self->get('mode')` check. Then in other places checks other keys of the object. To me this says that the object might be just incomplete for some reason, and based on the keys the object is assumed invalid. But the problem might also occur if there is a problem on the OpenID Provider respond or the URI is wrong.

lib/OpenQA/WebAPI/Auth/OpenID.pm
b10n1k: done c6cc29a9b

lib/OpenQA/WebAPI/Auth/OpenID.pm
okurz: Ok, but regardless the handler is seeing invalid data and can not process the request, right? So I don't think the "Likely" helps. And the second part is also adding details which can be considered good but then it does not really help with the issue that the data can not be processed. Imho the more important part that should be made explicit is what the effect of the situation is, i.e. you explained the "what". Still missing is the "so what"

b10n1k: Could be `so what` a prompt to retry but we remove it because this is not useful to the end user. I can think only a suggestion to check the OP. or a link to the documentation. 
The effect is not disastrous and it is not happen that often. Not sure what info makes sense

t/03-auth-openid.t
```diff
@@ -25,12 +26,36 @@ $c->set_always(session => \%session);
 ok OpenQA::WebAPI::Auth::OpenID::_handle_verified($c, $vident), 'can call _handle_verified';
 $users->called_ok('create_user', 'new user is created for initial login');
 is(($users->call_args(2))[1], 'mordred', 'new user created with details');
+
```
okurz, line 29: unrelated whitespace change.

t/03-auth-openid.t
```diff
@@ -25,12 +26,36 @@ $c->set_always(session => \%session);
...
+stderr_like {
+    my $ret = OpenQA::WebAPI::Auth::OpenID::auth_response($c);
+    is($ret, 0, 'can handle not_openid response on invalid OpenID message');
+}
+qr/Invalid OpenID/, 'warning about invalid OpenID is displayed';
+
+$c->set_always(
+    req => Test::MockObject->new->set_always(params => Test::MockObject->new->set_always(pairs => [1, 2]))
+      ->set_always(url => Test::MockObject->new->set_always(base => 'openqa'))
+      ->set_always(log => Test::MockObject->new->set_true('error', 'debug'))->set_true('flash'));
```
okurz, line 44: this looks like mostly duplicated from lines L28-32. Better reuse the original mocked controller and only overwrite the function you need.

t/03-auth-openid.t
```diff
@@ -25,12 +26,36 @@ $c->set_always(session => \%session);
...
+my $mock_openid_consumer = Test::MockModule->new('Net::OpenID::Consumer');
+$mock_openid_consumer->redefine(
+    'handle_server_response',
+    sub {
+        my ($self, %res_handlers) = @_;
+        if ($res_handlers{setup_needed}) {
+            return $res_handlers{setup_needed}->("https://www.opensuse.org/openid/setup");
+        }
```
okurz, line 52: can signatures be used here?

```suggestion
    sub ($self, %res_handlers) { return $res_handlers{setup_needed)? $res_handlers{setup_needed}->("https://www.opensuse.org/openid/setup") : undef }
```

lib/OpenQA/WebAPI/Auth/OpenID.pm
okurz: But this text is also provided to users. What should it tell them?

okurz (changes requested):

lib/OpenQA/WebAPI/Auth/OpenID.pm
b10n1k: *login failed*

t/03-auth-openid.t
b10n1k: done. 11e55fac4

t/03-auth-openid.t
b10n1k: done. 11e55fac4

t/03-auth-openid.t
b10n1k: done. 11e55fac4

okurz:
please rebase on top of https://github.com/os-autoinst/openQA/pull/5968

lib/OpenQA/WebAPI/Auth/OpenID.pm
kalikiana: @b10n1k Did you look into providing more details based on [Net::OpenID::Consumer](https://metacpan.org/pod/Net::OpenID::Consumer) as was suggested [in the original question about this message](https://github.com/os-autoinst/openQA/pull/5952#discussion_r1777294685)?

lib/OpenQA/WebAPI/Auth/OpenID.pm
```diff
@@ -108,18 +108,23 @@ sub auth_response ($c) {
     };
 
     $csr->handle_server_response(
-        not_openid =>
-          sub () { $err_handler->('Failed to login', 'OpenID provider returned invalid data. Please retry again') },
```
okurz, line 112: Just add the detail which openid server we connected, e.g.

```patch
- 'OpenID provider returned invalid data. Please retry again'
+ "OpenID provider '$url' returned invalid data. Please retry again"
```

and revert the rest

b10n1k: Done

t/03-auth-openid.t
```diff
@@ -26,11 +27,30 @@ ok OpenQA::WebAPI::Auth::OpenID::_handle_verified($c, $vident), 'can call _handl
...
+          ? $res_handlers{setup_needed}->("https://www.opensuse.org/openid/setup")
+          : undef;
+    });
+$c->set_always(
+    req => Test::MockObject->new->set_always(
+        params => Test::MockObject->new->set_always(pairs => ['openid.op_endpoint', 'https://www.opensuse.org/openid/'])
+    )->set_always(url => Test::MockObject->new->set_always(base => 'openqa')))
+  ->set_always(
+    app => Test::MockObject->new->set_always(config => {})->set_always(log => Test::MockObject->new->set_true('debug')))
+  ->set_true('flash');
```
okurz, line 53: can you reduce the duplication here?

okurz (changes requested):

t/03-auth-openid.t
b10n1k: I tried. but I didnt find a way. I couldnt just change the `set_always(log => Test::MockObject->new->set_true('debug'))`

b10n1k: I can leave it as it was but this was confusing to me. tests produces positive negatives. for instance:
if I replace `error` with `debug` in https://github.com/os-autoinst/openQA/pull/5952/files#diff-dc3d35214db1365be81906f2eabbaae8a531b3ab984c06eff8fd425f3cb40f2dR37 still works, it was before.

okurz: Ok, then leave this for now

okurz (approved):

baierjan (approved):

PR #5956, opened 2024-09-27
Decrease log level from warn to info for git clone reject

asdil12:
Ticket: https://progress.opensuse.org/issues/167386

### DIFF
--- a/lib/OpenQA/Task/Git/Clone.pm
+++ b/lib/OpenQA/Task/Git/Clone.pm
@@ -65,7 +65,9 @@ sub _git_clone ($app, $job, $ctx, $path, $url) {
 
         my $origin_url = $git->get_origin_url;
         if ($url ne $origin_url) {
-            $ctx->warn("Local checkout at $path has origin $origin_url but requesting to clone from $url");
+            $ctx->info(
+"Local checkout at $path has origin $origin_url but requesting to clone from $url. Be aware that the requested URL will not be cloned. This is fine.🔥"
+            );
             return;
         }
     }
--- a/t/14-grutasks-git.t
+++ b/t/14-grutasks-git.t
@@ -36,7 +36,7 @@ my $mojo_port = Mojo::IOLoop::Server->generate_port;
 my $webapi = OpenQA::Test::Utils::create_webapi($mojo_port, sub { });
 
 # prevent writing to a log file to enable use of combined_like in the following tests
-$t->app->log(Mojo::Log->new(level => 'warn'));
+$t->app->log(Mojo::Log->new(level => 'info'));
 
 subtest 'git clone' => sub {
     my $openqa_git = Test::MockModule->new('OpenQA::Git');


lib/OpenQA/Task/Git/Clone.pm
```diff
@@ -65,7 +65,7 @@ sub _git_clone ($app, $job, $ctx, $path, $url) {
 
         my $origin_url = $git->get_origin_url;
         if ($url ne $origin_url) {
-            $ctx->warn("Local checkout at $path has origin $origin_url but requesting to clone from $url");
+            $ctx->info("Local checkout at $path has origin $origin_url but requesting to clone from $url");
```
okurz, line 68: Can you rephrase the message to make the "so what" clear?

okurz (changes requested):

perlpunk:
webui-docker-compose fail is due to https://progress.opensuse.org/issues/167524

lib/OpenQA/Task/Git/Clone.pm
kalikiana: Is this actionable? My understanding from the ticket is this works as expected. So I would actually say remove it.

Martchus: This can be helpful without being actionable, e.g. when debugging and trying to make sense of what our Git code does or skips.

lib/OpenQA/Task/Git/Clone.pm
```diff
@@ -65,7 +65,9 @@ sub _git_clone ($app, $job, $ctx, $path, $url) {
 
         my $origin_url = $git->get_origin_url;
         if ($url ne $origin_url) {
-            $ctx->warn("Local checkout at $path has origin $origin_url but requesting to clone from $url");
+            $ctx->info(
+"Local checkout at $path has origin $origin_url but requesting to clone from $url. This is not that bad."
```
Martchus, line 69:
```suggestion
"Local checkout at $path has origin $origin_url but requesting to clone from $url. Be aware that the requested URL will not be cloned."
```

perlpunk, line 69: Suggestion by @okurz ;-)
```suggestion
"Local checkout at $path has origin $origin_url but requesting to clone from $url. This is fine.🔥"

```

okurz (approved):
Well, let's try :)

lib/OpenQA/Task/Git/Clone.pm
perlpunk: Sorry, that was just meant to be a joke, not a real suggestion

Martchus (approved):

lib/OpenQA/Task/Git/Clone.pm
asdil12: Pssst. Don't prevent the opinion to see what we can break with UTF-8 characters on the logfile ;)

PR #5961, opened 2024-09-30
Improve minion guards for needle tasks

perlpunk:
In OpenQA::Task::Needle::Save, OpenQA::Task::Needle::Delete and
OpenQA::Task::Git::Clone we are now using a guard for the individual directory
`git_clone_${needledir}_task` instead of a global guard limit_needle_task.

### DIFF
--- a/lib/OpenQA/Task/Git/Clone.pm
+++ b/lib/OpenQA/Task/Git/Clone.pm
@@ -6,6 +6,7 @@ use Mojo::Base 'Mojolicious::Plugin', -signatures;
 use Mojo::Util 'trim';
 
 use Mojo::File;
+use OpenQA::Log qw(log_debug);
 use Time::Seconds 'ONE_HOUR';
 
 sub register ($self, $app, @) {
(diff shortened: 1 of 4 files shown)


Martchus (changes requested):
We probably need the same for the delete needle task.

perlpunk:
> We probably need the same for the delete needle task.

why?

perlpunk:
> Because it also invokes Git commands on a needle checkout (so far only using the `limit_needle_task` lock).

Oh, the git operation is hidden in OpenQA::Schema::Result::Needles. I only looked at OpenQA::Task::Needle::Delete and didn't expect it elsewhere :(

okurz:
@mergifyio rebase

perlpunk:
I updated all git gru tasks to use a guard per directory instead of the global one

perlpunk:
There seems to be flaky coverage in `lib/OpenQA/Shared/Controller/Running.pm`. We don't see it because the coverage decrease is so low, but going to the report I can see it in the indirect changes ta often:
```
     # setup a function to stop streaming again
     my $timer_id;
     my $close_connection = sub ($self, @) {     # line 204-206 reported as uncovered
         $self->finish;
     };
 
     # setup a recurring timer to send the last screenshot to the client
```

okurz (approved):

b10n1k (approved):

PR #5968, opened 2024-10-01
dist: Add required "group(kvm)" for openQA-worker

okurz:
openSUSE Factory since about 20240928 seems to need an explicit addition
of the "kvm" group. This commit tries to fix that same as we do with
"nogroup".

Related progress issue: https://progress.opensuse.org/issues/167587

### DIFF
--- a/dist/rpm/openQA.spec
+++ b/dist/rpm/openQA.spec
@@ -205,6 +205,7 @@ Recommends:     pngquant
 Recommends:     openssh-common
 %if 0%{?suse_version} >= 1330
 Requires(pre):  group(nogroup)
+Requires(pre):  group(kvm)
 %endif
 
 %description worker


okurz:
@Vogtinator @DimStar77 can you say if this is the right approach to fix

```
usermod: group 'kvm' does not exist
```

in the openSUSE Factory build or do you have another suggestion?

Vogtinator:
sysusers.d says: `Add a user to a group. If the user or group do not exist yet, they will be implicitly created.`

DimStar77:
Ensuring the group is properly pulled in when using it does sound correct to me

Vogtinator:
So arguably a bug in whatever reads the sysusers.d file there, but IMO requiring the proper group is the better option.

We can't use the RPM native sysusers.d support and sysuser-tools doesn't generate requires, so I think this is the correct fix.

PR #4402, opened 2021-12-13
cli: List supported search criteria for jobs/overview route

okurz:
Output in the openqa-cli help:

```
  # List the latest jobs matching the search criteria
  # supported search criteria: distri, version, flavor, build, test, modules, modules_result, module_re, group, groupid, id
  openqa-cli api --osd jobs/overview groupid=135 distri=caasp version=3.0
```

Related progress issue: https://progress.opensuse.org/issues/103864

PR #5974, opened 2024-10-03
cli: List 'arch' and 'machine' as supported in API help

rbmarliere:
Since commit [f19726339710](https://github.com/os-autoinst/openQA/commit/f19726339710e9d9deaf55c0ea2b5a84426f9aad) ("Support parameters `arch` and `machine` in job overview API"), user can filter using those. Add them to the constant shown to the user in `openqa-cli api -h`.

Links:
https://github.com/os-autoinst/openQA/pull/4402
https://progress.opensuse.org/issues/103864

### DIFF
--- a/lib/OpenQA/Constants.pm
+++ b/lib/OpenQA/Constants.pm
@@ -99,7 +99,7 @@ use constant VIDEO_FILE_NAME_REGEX => qr/^.*\/video\.[^\/]*$/;
 use constant FRAGMENT_REGEX => qr'(#([-?/:@.~!$&\'()*+,;=\w]|%[0-9a-fA-F]{2})*)*';
 
 use constant JOBS_OVERVIEW_SEARCH_CRITERIA =>
-  (qw(distri version flavor build test modules modules_result module_re group groupid id));
+  (qw(distri version flavor arch build test machine modules modules_result module_re group groupid id));
 
 our @EXPORT_OK = qw(
   WEBSOCKET_API_VERSION DEFAULT_WORKER_TIMEOUT


lib/OpenQA/Constants.pm
```diff
@@ -99,7 +99,7 @@ use constant VIDEO_FILE_NAME_REGEX => qr/^.*\/video\.[^\/]*$/;
 use constant FRAGMENT_REGEX => qr'(#([-?/:@.~!$&\'()*+,;=\w]|%[0-9a-fA-F]{2})*)*';
 
 use constant JOBS_OVERVIEW_SEARCH_CRITERIA =>
-  (qw(distri version flavor build test modules modules_result module_re group groupid id));
+  (qw(arch distri version flavor build test modules modules_result module_re group groupid id machine));
```
okurz, line 102: We commonly order it like this so please write like this

```suggestion
  (qw(distri version flavor arch build test machine modules modules_result module_re group groupid id));
```

t/43-cli-api.t
```diff
@@ -70,7 +70,7 @@ my $api = OpenQA::CLI::api->new;
 subtest 'Help' => sub {
     my ($stdout, $stderr, @result) = capture sub { $cli->run('help', 'api') };
     like $stdout, qr/Usage: openqa-cli api/, 'help';
-    like $stdout, qr/supported search criteria: distri, version.*id/, 'search criteria listed';
+    like $stdout, qr/supported search criteria: arch, distri, version.*machine/, 'search criteria listed';
```
okurz, line 73: Keeping the old should be fine with my above suggestion

lib/OpenQA/Constants.pm
rbmarliere: Thanks for the review, makes sense to me!

okurz:
@Martchus as the author of f19726339710 can you confirm this makes sense?

okurz (approved):
@Martchus as the author of f19726339710 can you confirm this makes sense?

Martchus (changes requested):
The commit title should be on a higher level and not just describe the change itself (which is already obvious from the diff).

I suggest `cli: List 'arch' and 'machine' as supported in API help`.

You could also append one more note to the commit message body:

> Note that as a side-effect these fields will now also be explicitly validated as optional non-empty parameters. This is not expected to have an impact on the API usage.

(But the change itself looks good.)

PR #5976, opened 2024-10-04
Remove scripted waiting for db in favor of healthcheck

baierjan:
Reference: https://progress.opensuse.org/issues/167524

### DIFF
--- a/container/webui/docker-compose.yaml
+++ b/container/webui/docker-compose.yaml
@@ -126,8 +126,7 @@ services:
       start_period: 5s
 
   db:
-    # Teporary fix version https://progress.opensuse.org/issues/167524
-    image: postgres:16
+    image: postgres
     restart: always
     environment:
       POSTGRES_PASSWORD: openqa
@@ -136,7 +135,12 @@ services:
     volumes:
       - ./workdir/db:/var/lib/postgresql/data:Z
     healthcheck:
-      test: ["CMD", "sh", "-c", "echo 'select * from api_keys;' | psql -U openqa -v 'ON_ERROR_STOP=1' openqa"]
+      test:
+        - CMD-SHELL
+        - >-
+          pg_isready -U openqa -d openqa
+          && echo 'select * from api_keys;'
+          | psql -U openqa -v 'ON_ERROR_STOP=1' openqa
       interval: 10s
       timeout: 10s
       retries: 3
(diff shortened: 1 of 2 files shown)


container/webui/docker-compose.yaml
```diff
@@ -136,7 +135,12 @@ services:
     volumes:
       - ./workdir/db:/var/lib/postgresql/data:Z
     healthcheck:
-      test: ["CMD", "sh", "-c", "echo 'select * from api_keys;' | psql -U openqa -v 'ON_ERROR_STOP=1' openqa"]
+      test:
+        - CMD-SHELL
+        - >-
+          pg_isready -U openqa -d openqa
```
b10n1k, line 141: shouldnt use a retry here?

baierjan: No, why? This is a healthcheck probe: https://docs.docker.com/reference/dockerfile/#healthcheck, the retry is "built-in".

b10n1k (approved):

kalikiana (approved):
Nice!

container/webui/docker-compose.yaml
```diff
@@ -136,7 +135,12 @@ services:
     volumes:
       - ./workdir/db:/var/lib/postgresql/data:Z
     healthcheck:
-      test: ["CMD", "sh", "-c", "echo 'select * from api_keys;' | psql -U openqa -v 'ON_ERROR_STOP=1' openqa"]
+      test:
```
perlpunk, line 138: What's the functional difference here? Does this healthcheck somehow work different than before?

baierjan: > so this is just nicer

Yes, correct.

> is this change here needed for the PR?

Probably not, I guess the former test would work as well. This (`pg_isready`) is a recommended way to check, so I added it.

From my understanding the presence of
```
    depends_on:
      db:
        condition: service_healthy
```
together with an appropriate health-check in the db container should be enough. And that's precisely what the linked commit adds. The check in the script I removed was added before that..

> Otherwise the checks in the script would have never been necessary

If I am reading the history correctly, it was introduced in the opposite order and the script wasn't necessary (unless of course docker didn't know proper healthchecks back then).

We should also note, that the podman/docker was itself also improved in the meantime and maybe it was also not behaving correctly at that time. It could be also that I am utterly wrong and we will see the race condition in the near feature.

PR #5978, opened 2024-10-04
Fix race condition in `t/ui/18-tests-details.t`

Martchus:
* The synchronization via `wait_for_ajax_and_animations` did likely not work because there is probably still asynchronicity going on despite the attempt to disable animations explicitly. Note that the relevant Ajax code still uses jQuery so this is *not* a regression of 87a7e36.
* Use `wait_for_element` like in many other cases.
* Avoid running into `Can't call method "is_enabled" on an undefined value` in case the test fails.
* Remove code that attempted to find more than one element by ID. (Even if the ID was not unique the test would not be able to find this violation.)
* See https://progress.opensuse.org/issues/167656

### DIFF
--- a/t/ui/18-tests-details.t
+++ b/t/ui/18-tests-details.t
@@ -77,13 +77,12 @@ sub current_tab { $driver->find_element('.nav.nav-tabs .active')->get_text }
 # returns the contents of the candidates combo box as hash (key: tag, value: array of needle names)
 sub find_candidate_needles {
     # ensure the candidates menu is visible
-    my @candidates_menus = $driver->find_elements('#candidatesMenu');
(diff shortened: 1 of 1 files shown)


b10n1k (approved):

b10n1k:
just a note. Codecov shows a small decrease but I dont get it. The commit doesnt look like removing  something in this sense.

perlpunk:
That's because https://app.codecov.io/gh/os-autoinst/openQA/pull/5978/indirect-changes
`lib/OpenQA/Shared/Controller/Running.pm` has flaky coverage.
I noticed this already before. I forgot to create a ticket.

PR #5979, opened 2024-10-07
Add munin alert email wrapper

perlpunk:
Issue: https://progress.opensuse.org/issues/166739

The only question is to which path this should be deployed as part of the openQA-munin package

### DIFF
--- a/contrib/munin/utils/munin-mail
+++ b/contrib/munin/utils/munin-mail
@@ -1,0 +1,22 @@
+#!/usr/bin/perl
+# script to be used for sending munin alert emails. This lets us add
+# useful information and links for certain alerts.
+# configuration in munin.conf:
+# contact.o3admins.command /path/to/munin-mail "${var:group} ${var:host} ${var:plugin} ${var:graph_category} '${var:graph_title}'" email-address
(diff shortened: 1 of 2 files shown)


contrib/munin/utils/munin-mail
```diff
@@ -0,0 +1,22 @@
...
+my $content = '';
+while (<STDIN>) {
+    $content .= $_;
+}
+if ($subject =~ m/systemd_(units|status)/) {
+    $content .= "\n" . qx{systemctl --failed};
+}
+
+open my $pipe, '|-', 'mail', '-s', $subject, '-r', $email, $email or die "Could not open pipe: $?";
+print $pipe $content;
```
Martchus, line 20: This line probably also needs `… or die …`.

Martchus (approved):

Martchus:
It could probably just go under `/usr/share/openqa/script`.

perlpunk:
> It could probably just go under `/usr/share/openqa/script`.

Ok, I hope I added the correct instructions to the spec file

okurz (approved):

PR #5981, opened 2024-10-07
Allow adding a comment when restarting jobs

Martchus:
See https://progress.opensuse.org/issues/167827

### DIFF
--- a/lib/OpenQA/Schema/Result/Jobs.pm
+++ b/lib/OpenQA/Schema/Result/Jobs.pm
@@ -657,7 +657,7 @@ sub _create_clone_with_child ($res, $clones, $c, $dependency) {
     $res->children->find_or_create({child_job_id => $c, dependency => $dependency});
 }
 
-sub _create_clones ($self, $jobs, $comment, $comment_user_id, @clone_args) {
+sub _create_clones ($self, $jobs, $comments, $comment_text, $comment_user_id, @clone_args) {
     # create the clones
     my $result_source = $self->result_source;
     my $rset = $result_source->resultset;
@@ -703,8 +703,9 @@ sub _create_clones ($self, $jobs, $comment, $comment_user_id, @clone_args) {
     }
 
     # create comments on original jobs
-    $result_source->schema->resultset('Comments')->create_for_jobs(\@original_job_ids, $comment, $comment_user_id)
-      if defined $comment;
+    $result_source->schema->resultset('Comments')
+      ->create_for_jobs(\@original_job_ids, $comment_text, $comment_user_id, $comments)
+      if defined $comment_text;
 }
 
 # internal (recursive) function for duplicate - returns hash of all jobs in the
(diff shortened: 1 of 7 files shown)


lib/OpenQA/Schema/Result/Jobs.pm
```diff
@@ -641,6 +641,8 @@ sub _create_clone ($self, $cluster_job_info, $clone, $prio, $skip_ok_result_chil
         die "Job $orig_id already has clone " . $rset->find($orig_id)->clone_id . "\n" unless $affected_rows == 1;
     }
 
+
+
     # Needed to load default values from DB
```
okurz, line 646: Unrelated w/s change. Please revert

lib/OpenQA/Schema/Result/Jobs.pm
```diff
@@ -657,13 +659,15 @@ sub _create_clone_with_child ($res, $clones, $c, $dependency) {
     $res->children->find_or_create({child_job_id => $c, dependency => $dependency});
 }
 
-sub _create_clones ($self, $jobs, @clone_args) {
+sub _create_clones ($self, $jobs, $comment, $comment_user_id, @clone_args) {
```
okurz, line 662: Shouldn't that be

```suggestion
sub _create_clones ($self, $jobs, $comment = undef, $comment_user_id = undef, @clone_args) {
```

lib/OpenQA/Schema/Result/Jobs.pm
```diff
@@ -692,13 +696,26 @@ sub _create_clones ($self, $jobs, @clone_args) {
...
+    $clones{$_}->calculate_blocked_by for @original_job_ids;
 
     # add a reference to the clone within $jobs
-    for my $job (keys %clones) {
+    for my $job (@original_job_ids) {
         my $clone = $clones{$job};
         $jobs->{$job}->{clone} = $clone->id if $clone;
     }
+
+    # create comments on original jobs
```
okurz, line 707: This comment sounds like the following code wants to be it's own function

lib/OpenQA/WebAPI/Controller/API/V1/Job.pm
```diff
@@ -848,7 +854,9 @@ sub _restart ($self, %args) {
     my @urls;
     for (my $i = 0; $i < @$duplicates; $i++) {
         my $result = $duplicates->[$i];
-        $self->emit_event(openqa_job_restart => {id => $jobs->[$i], result => $result, auto => $auto});
+        my %event_data = (id => $jobs->[$i], result => $result, auto => $auto);
+        $event_data{comment} = $comment if defined $comment;
+        $self->emit_event(openqa_job_restart => \%event_data);
```
okurz, line 859: So this would emit the restart event with a comment attribute but no comment creation event would be triggered. Should we instead emit both events?

okurz (changes requested):

lib/OpenQA/Schema/Result/Jobs.pm
Martchus: No, those arguments will always be passed.

perlpunk: Then the first two additional `@job_args` would land in those variables.

lib/OpenQA/WebAPI/Controller/API/V1/Job.pm
Martchus: Correct. For the sake of simplicity I just made it one event. However, I can change it to also emit single events for comments. (Will be a bit more code changes to pass around all the information.)

asdil12 (approved):

PR #5988, opened 2024-10-08
Allow limiting number of worker websocket connections

Martchus:
* Allow configuring the limit server-side in `openqa.ini`
* Let workers wait with the next connection attempt until a configurable
  amount of time has passed (instead of the normal 10 seconds)
* Show such workers as "Offline (limited)" in the web UI excluding them
  from assignments from the scheduler
* See https://progress.opensuse.org/issues/157690

---

I also tested manually whether the worker, scheduler and table in the UI
behave correctly. I could not find any problems with it.

This change simply sets the "error" field of the database similar to what
we already do for graceful disconnects. This way we don't need to invent
yet another state a worker can be in for the scheduler to avoid assigning
jobs to it (and for the admin table to show it as offline despite a successful
initial registration).


etc/openqa/openqa.ini
```diff
@@ -263,6 +263,10 @@ concurrent = 0
 #assets_default_limit = 100000
 ## Maximum number of next jobs to include asset listing request (to prevent performance issues)
 #assets_max_limit = 200000
+## Maximum number of online workers (to prevent performance issues)
+#max_online_workers =
```
okurz, line 267: We have sensible defaults for many if not most other cases so I suggest to do the same here. This is also given the rather explicit feedback about that limitation that you built in. So how about we use 1000.

okurz (changes requested):
One note regarding a default value.

And there is missing patch coverage which would be great if you can look into.

etc/openqa/openqa.ini
Martchus: I'm not sure about it. The 1000 is probably specific to our setup, e.g. if our VM was backed by a different physical machine (e.g. with a larger CPU cache or whatever creates this performance threshold) the limit might be higher.

okurz: the same applies to all the other numbers for which we provide a default. The thing is: We know that openQA will misbehave if there are too many worker instances connected so that's why we should limit by default. If you want to argue that 2k or 10k is a better limit then fine, do that. But I think the system specs of OSD are not too low and also not that special so that's why I think 1k is a good value. We could even try to go lower if that's what you prefer

okurz (approved):

perlpunk (approved):

PR #5973, opened 2024-10-02
Drop OpenQA::WebAPI::Plugin::HashedParams

asdil12:
Ticket: https://progress.opensuse.org/issues/64412

This will drop the legacy support for `application/x-www-form-urlencoded` requests handled via the AdminTables code on the server side.
Instead only the new `application/json` requests will be supported.
This is only affecting `POST` and `PUT` requests.
`GET` requests can still filter using query parameters.

This is mostly due to the odd way that nested hashes for settings variables were encoded in the flat `x-www-form-urlencoded` structure (e.g. `settings[FOO]=bar&settings[baz]=abc`).

The webui is already moved to the new json requests.

This change will remove backwards compatibility with the old requests for these API routes:

```
/api/v1/products
```

```
/api/v1/machines
```

```
/api/v1/test_suites
```

Importing job templates should continue to work fine as this PR also updates `openqa-load-templates` to json.

@AdamWill @phil-hands @marmarek FYI

PR #5992, opened 2024-10-09
Revert "Drop OpenQA::WebAPI::Plugin::HashedParams"

perlpunk:
Reverts os-autoinst/openQA#5973

The problem is that it can take while until client side packages are updated, for example in tumbleweed.
I suggest to split this up and make the change in openqa-load-templates first, and only disable the feature in the webui after a while.
There is no urgency to get rid of that old feature.


asdil12:
Let's only revert the changes on the server side.

asdil12:
--> https://github.com/os-autoinst/openQA/pull/5993

okurz:
https://github.com/os-autoinst/openQA/pull/5993 merged. @asdil12 can you prepare the diff that is left for server-side as another PR which we will mark as "not-ready" until tooling is available for more downstream users?

PR #5995, opened 2024-10-09
Make activity view accessible to operators

Martchus:
* Allow anyone to access the activity view but show the menu item only to
  operators because only those users will have any activity (that the
  activity view supports, so comments don't count)
* Add additional (internal) route to query only audit events for the current
  user to avoid having to expose all audit events to more users
* See https://progress.opensuse.org/issues/167818

### DIFF
--- a/lib/OpenQA/WebAPI.pm
+++ b/lib/OpenQA/WebAPI.pm
@@ -245,7 +245,8 @@ sub startup ($self) {
     my $pub_admin_r = $admin->any('/')->to(namespace => 'OpenQA::WebAPI::Controller::Admin');
 
     # operators accessible tables
-    $admin_r->get('/activity_view')->name('activity_view')->to('activity_view#user');
+    $pub_admin_r->get('/activity_view')->name('activity_view')->to('activity_view#user');
+    $pub_admin_r->get('/activity_view/ajax')->name('activity_view_ajax')->to('audit_log#ajax_current_user');
     $pub_admin_r->get('/products')->name('admin_products')->to('product#index');
     $pub_admin_r->get('/machines')->name('admin_machines')->to('machine#index');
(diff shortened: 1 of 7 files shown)


perlpunk (dismissed):

lib/OpenQA/WebAPI.pm
```diff
@@ -286,8 +286,8 @@ sub startup ($self) {
     $admin_r->get('/needles/:module_id/:needle_id')->name('admin_needle_module')->to('needle#module');
     $admin_r->get('/needles/ajax')->name('admin_needle_ajax')->to('needle#ajax');
     $admin_r->delete('/needles/delete')->name('admin_needle_delete')->to('needle#delete');
-    $admin_r->get('/auditlog')->name('audit_log')->to('audit_log#index');
-    $admin_r->get('/auditlog/ajax')->name('audit_ajax')->to('audit_log#ajax');
+    $op_r->get('/auditlog')->name('audit_log')->to('audit_log#index');
+    $op_r->get('/auditlog/ajax')->name('audit_ajax')->to('audit_log#ajax');
```
okurz, line 290: I don't think this is a good idea. The audit log can reveal personal information and enable something like "tracking employee performance" hence it is on purpose limited to admins. We should not allow that for the big group of operators. We need to think of a better solution here.

okurz (changes requested):

lib/OpenQA/WebAPI.pm
Martchus: Ok, so I guess I'll implement a special "limited" audit log route where the user parameter is always the current user. That I could actually make accessible to all users so the audit log would no longer have to be operator-only.

okurz: That sounds perfect!

okurz (approved):

PR #5997, opened 2024-10-09
Stabilize `t/ui/21-admin-needles.t`

Martchus:
See individual commit messages and https://progress.opensuse.org/issues/167611.


okurz (approved):

t/lib/OpenQA/SeleniumTest.pm
```diff
@@ -345,6 +345,18 @@ sub wait_for_element (%args) {
     return $element;
 }
 
+sub wait_for_data_table_entries ($table, $expected_entry_count) {
```
asdil12, line 348: Is `$expected_entry_count` actually referring to the expected number of rows in the table?
If yes, then maybe a better phrasing could be used to name this variable.

baierjan (approved):

PR #5998, opened 2024-10-10
Show error when admintable settings key contains invalid chars

asdil12:
Ticket: https://progress.opensuse.org/issues/104631

### DIFF
--- a/lib/OpenQA/WebAPI/Controller/API/V1/Table.pm
+++ b/lib/OpenQA/WebAPI/Controller/API/V1/Table.pm
@@ -5,7 +5,7 @@
 package OpenQA::WebAPI::Controller::API::V1::Table;
 use Mojo::Base 'Mojolicious::Controller', -signatures;
 
-use Mojo::Util 'trim';
+use Mojo::Util qw(trim xml_escape);
 use OpenQA::App;
 use OpenQA::Log 'log_debug';
 use List::Util qw(min);
@@ -397,9 +397,16 @@ sub _prepare_settings {
     my @keys;
     if ($hp->{settings}) {
         for my $k (keys %{$hp->{settings}}) {
-            $k = trim $k;
             my $value = trim $hp->{settings}->{$k};
-            $k =~ s/[^\]\[0-9a-zA-Z_\+]//g;
+            $k = trim $k;
+            my $invalid_chars_in_key = $k;
+            # Remove any allowed chars so that only invalid chars remain
+            $invalid_chars_in_key =~ s/[\]\[0-9a-zA-Z_\+]//g;
+            $invalid_chars_in_key =~ s/(.)(?=.*\1)//g;    # Remove any chars that occurs more than once
+            if ($invalid_chars_in_key ne '') {
+                my $eick = join(', ', map { '<b>' . xml_escape($_) . '</b>' } split('', $invalid_chars_in_key));
+                return sprintf('Invalid characters %s in settings key <b>%s</b>', $eick, xml_escape($k));
+            }
             push @settings, {key => $k, value => $value};
             push @keys, $k;
         }
--- a/lib/OpenQA/WebAPI/Controller/API/V1/Table.pm
+++ b/lib/OpenQA/WebAPI/Controller/API/V1/Table.pm
@@ -399,12 +399,10 @@ sub _prepare_settings {
         for my $k (keys %{$hp->{settings}}) {
             my $value = trim $hp->{settings}->{$k};
             $k = trim $k;
-            my $invalid_chars_in_key = $k;
-            # Remove any allowed chars so that only invalid chars remain
-            $invalid_chars_in_key =~ s/[\]\[0-9a-zA-Z_\+]//g;
-            $invalid_chars_in_key =~ s/(.)(?=.*\1)//g;    # Remove any chars that occurs more than once
-            if ($invalid_chars_in_key ne '') {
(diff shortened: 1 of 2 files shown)


lib/OpenQA/WebAPI/Controller/API/V1/Table.pm
```diff
@@ -397,9 +397,17 @@ sub _prepare_settings {
     my @keys;
     if ($hp->{settings}) {
         for my $k (keys %{$hp->{settings}}) {
-            $k = trim $k;
             my $value = trim $hp->{settings}->{$k};
-            $k =~ s/[^\]\[0-9a-zA-Z_\+]//g;
+            $k = trim $k;
+            my $invalid_chars_in_key = $k;
+            $invalid_chars_in_key =~ s/[\]\[0-9a-zA-Z_\+]//g;
```
Martchus, line 403: This line probably deserves a comment as well.

Martchus (commented):
I suppose tests should be extended.

okurz (approved):

lib/OpenQA/WebAPI/Controller/API/V1/Table.pm
```diff
@@ -397,9 +397,16 @@ sub _prepare_settings {
...
-            $k =~ s/[^\]\[0-9a-zA-Z_\+]//g;
+            $k = trim $k;
+            my $invalid_chars_in_key = $k;
+            # Remove any allowed chars so that only invalid chars remain
+            $invalid_chars_in_key =~ s/[\]\[0-9a-zA-Z_\+]//g;
+            $invalid_chars_in_key =~ s/(.)(?=.*\1)//g;    # Remove any chars that occurs more than once
+            if ($invalid_chars_in_key ne '') {
+                my $eick = join(', ', map { '<b>' . xml_escape($_) . '</b>' } split('', $invalid_chars_in_key));
+                return sprintf('Invalid characters %s in settings key <b>%s</b>', $eick, xml_escape($k));
+            }
```
perlpunk, line 409:
```suggestion
            my %invalid;
            $k =~ s/([^\]\[0-9a-zA-Z_\+])/$inv{$1} = 1/eg;
            delete $invalid{' '}; # silently remove spaces
            if (keys %invalid) {
                my $eick = join ', ', map { '<b>' . xml_escape($_) . '</b>' } sort keys %invalid;
                return sprintf('Invalid characters %s in settings key <b>%s</b>', $eick, xml_escape($k));
            }
```

perlpunk: Also, if you want to keep your code, as an alternative to a regex:
```
my $invalid_chars_in_key = $k = tr/[]0-9a-zA-Z_+//dr;
```

asdil12: hm - this doesn't work as expected as it modifies `$k`.

asdil12: also I'm not sure if we want to `delete $invalid{' '}; # silently remove spaces`

asdil12: I updated it accordingly.

lib/OpenQA/WebAPI/Controller/API/V1/Table.pm
```diff
@@ -397,9 +397,15 @@ sub _prepare_settings {
     my @keys;
     if ($hp->{settings}) {
         for my $k (keys %{$hp->{settings}}) {
-            $k = trim $k;
             my $value = trim $hp->{settings}->{$k};
-            $k =~ s/[^\]\[0-9a-zA-Z_\+]//g;
+            $k = trim $k;
+            my $ktmp = $k;
+            my %invalid;
+            $ktmp =~ s/([^\]\[0-9a-zA-Z_\+])/$invalid{$1} = 1; ''/eg;
```
perlpunk, line 404: You could save one line:
```suggestion
            my %invalid;
            my $ktmp = $k =~ s/([^\]\[0-9a-zA-Z_\+])/$invalid{$1} = 1; ''/egr;
```

asdil12: No this doesn't work. When doing this, `$ktmp` and `$k` have both the same value afterwards - I just tested it.

asdil12: Edit: When using the `/r` flag, it actually works.

lib/OpenQA/WebAPI/Controller/API/V1/Table.pm
```diff
@@ -397,9 +397,14 @@ sub _prepare_settings {
     my @keys;
     if ($hp->{settings}) {
         for my $k (keys %{$hp->{settings}}) {
-            $k = trim $k;
             my $value = trim $hp->{settings}->{$k};
-            $k =~ s/[^\]\[0-9a-zA-Z_\+]//g;
+            $k = trim $k;
+            my %invalid;
+            () = $k =~ s/([^\]\[0-9a-zA-Z_\+])/$invalid{$1} = 1; ''/egr;
```
perlpunk, line 403: Ok, I didn't see that you would not need the returned string.
Clever :) but maybe then just use a match:
```suggestion
            @inv{$foo =~ m/([^\]\[0-9a-zA-Z_\+])/g} = ();
```

perlpunk (approved):

PR #5999, opened 2024-10-10
Avoid incomplete jobs when restarting Minion while syncing tests

Martchus:
* Use a signal guard like we already do in the `cache_asset` task (also part of the cache service)
* Tested by sending SIGTERM to a running `cache_tests` tasks; it went temporarily into the "inactive" state and the openQA job could still run after the Minion job was eventually finished
* See https://progress.opensuse.org/issues/167797

### DIFF
--- a/lib/OpenQA/CacheService/Task/Sync.pm
+++ b/lib/OpenQA/CacheService/Task/Sync.pm
@@ -3,6 +3,7 @@
 
 package OpenQA::CacheService::Task::Sync;
 use Mojo::Base 'Mojolicious::Plugin', -signatures;
+use OpenQA::Task::SignalGuard;
 
 use Mojo::URL;
 use Time::Seconds;
@@ -14,6 +15,8 @@ use constant RSYNC_RETRY_PERIOD => $ENV{OPENQA_RSYNC_RETRY_PERIOD} // 3;
 sub register ($self, $app, $conf) { $app->minion->add_task(cache_tests => \&_cache_tests) }
 
 sub _cache_tests ($job, $from = undef, $to = undef) {
+    my $ensure_task_retry_on_termination_signal_guard = OpenQA::Task::SignalGuard->new($job);
+
     my $app = $job->app;
     my $job_id = $job->id;
     my $lock = $job->info->{notes}{lock};


lib/OpenQA/CacheService/Task/Sync.pm
```diff
@@ -14,6 +15,8 @@ use constant RSYNC_RETRY_PERIOD => $ENV{OPENQA_RSYNC_RETRY_PERIOD} // 3;
 sub register ($self, $app, $conf) { $app->minion->add_task(cache_tests => \&_cache_tests) }
 
 sub _cache_tests ($job, $from = undef, $to = undef) {
+    my $ensure_task_retry_on_termination_signal_guard = OpenQA::Task::SignalGuard->new($job);
```
baierjan, line 18: Just to be sure I understand correctly as this practically looks like an unused variable (object) but it is actually doing important "magic". This just use the object constructor to setup things and then the destructor to some more stuff when the variable goes out-of-scope at the end of the function, right?

baierjan: I wonder if and how static code analysis tools are able to understand that (maybe by inspecting the class?). Or if that is considered a side-effect shoudn't be best practice to at least do some assignment (at least something like `$a = undef`) to the variable at the end to "use" the variable? I mean, I have nothing against the code (after all it is Perl, so I am expecting such magic from every other line), just trying to think if there is an alternative way without using the "write-only" variable if you know what I mean.

baierjan (approved):

lib/OpenQA/CacheService/Task/Sync.pm
asdil12, line 18:
```suggestion
    # this relies on constructor and destructor magic when the variable is getting out of scope
    my $ensure_task_retry_on_termination_signal_guard = OpenQA::Task::SignalGuard->new($job);
```
maybe add something like this as a comment?

okurz: Such comment shouldn't be needed. The same is done in other places and the variable suffix "guard" should be sufficient to hint to that

okurz (approved):

lib/OpenQA/CacheService/Task/Sync.pm
baierjan: I guess I found my answer in https://metacpan.org/pod/Perl::Critic::Policy::Variables::ProhibitUnusedVariables
> Right now, this only looks for simply declared, uninitialized lexical variables.

PR #6001, opened 2024-10-10
Reuse existing GruTasks for same git_clone task

perlpunk:
To avoid calling `git fetch` too often and having too many minion locks,
we skip a task if it has been done less than one minute ago.

https://progress.opensuse.org/issues/164898

### DIFF
--- a/lib/OpenQA/Shared/Plugin/Gru.pm
+++ b/lib/OpenQA/Shared/Plugin/Gru.pm
@@ -9,11 +9,12 @@ use Minion;
 use DBIx::Class::Timestamps 'now';
 use OpenQA::Schema;
 use OpenQA::Shared::GruJob;
-use OpenQA::Log 'log_info';
+use OpenQA::Log qw(log_debug log_info);
 use OpenQA::Utils qw(sharedir);
 use Mojo::Pg;
 use Mojo::Promise;
 use Mojo::File qw(path);
+use Mojo::JSON qw(decode_json);
 
 has app => undef, weak => 1;
 has 'dsn';
@@ -115,6 +116,52 @@ sub is_task_active ($self, $task) {
 # checks if there are worker registered
 sub has_workers ($self) { !!$self->app->minion->backend->list_workers(0, 1)->{total} }
 
+# For some tasks with the same args we don't need to repeat them if they were
+# enqueued less than a minute ago, like 'git fetch'
+sub _find_existing_minion_job ($self, $task, $args, $job_ids) {
+    my $schema = OpenQA::Schema->singleton;
+    $args = [$args] if ref $args eq 'HASH';
+    my $dtf = $schema->storage->datetime_parser;
+    my $dbh = $schema->storage->dbh;
+    my $sql = q{SELECT id, args, created, state, retries, notes, result FROM minion_jobs
+                WHERE state IN ('inactive', 'active', 'finished')
+                AND created >= ? AND task = ? AND args = ?
+                ORDER BY array_position(array['finished'::varchar, 'inactive'::varchar, 'active'::varchar], state::varchar)
+                LIMIT 1};
+    my $sth = $dbh->prepare($sql);
+    my @args = (
+        $dtf->format_datetime(DateTime->now()->subtract(minutes => 1)),
+        'git_clone', OpenQA::Schema::Result::GruTasks->encode_json_to_db($args));
+    $sth->execute(@args);
+    return unless my $job = $sth->fetchrow_hashref;
+    my $notes = decode_json $job->{notes};
+    if ($job->{state} eq 'finished') {
+        # same task was run less than 1 minute ago and finished, nothing to do
+        @$job_ids = ();
+        return;
+    }
+    $self->_add_jobs_to_gru_task($notes->{gru_id}, $job_ids);
+}
+
+sub _add_jobs_to_gru_task ($self, $gru_id, $job_ids) {
+    my $schema = OpenQA::Schema->singleton;
+    for my $id (@$job_ids) {
+        # Add job to existing gru task with the same args
+        my $gru_dep = eval { $schema->resultset('GruDependencies')->create({job_id => $id, gru_task_id => $gru_id}); };
+        unless ($gru_dep) {
+            my $error = $@;
+            die $error
+              unless $error
+              =~ m/insert or update on table "gru_dependencies" violates foreign key constraint "gru_dependencies_fk_gru_task_id"/i;
+            # if the GruTask was already deleted meanwhile, we can skip
+            # the rest of the jobs, since the wanted task was done
+            log_debug("GruTask $gru_id already gone, skip assigning jobs (message: $error)");
+            last;
+        }
+    }
+    @$job_ids = ();
+}
+
 sub enqueue ($self, $task, $args = [], $options = {}, $jobs = []) {
     my $ttl = $options->{ttl};
     my $limit = $options->{limit} ? $options->{limit} : undef;
@@ -199,7 +246,8 @@ sub enqueue_git_clones ($self, $clones, $job_ids) {
     return unless OpenQA::App->singleton->config->{'scm git'}->{git_auto_clone} eq 'yes';
     # $clones is a hashref with paths as keys and git urls as values
     # $job_id is used to create entries in a related table (gru_dependencies)
-    $self->enqueue('git_clone', $clones, {priority => 10}, $job_ids);
+    $self->_find_existing_minion_job('git_clone', $clones, $job_ids);
+    $self->enqueue('git_clone', $clones, {priority => 10}, $job_ids) if @$job_ids;
 }
 
 sub enqueue_and_keep_track {
--- a/t/14-grutasks-git.t
+++ b/t/14-grutasks-git.t
@@ -231,6 +231,61 @@ subtest 'git_update_all' => sub {
     is_deeply [sort keys %$args], \@clones, 'job args as expected';
 };
 
+subtest 'enqueue_git_clones' => sub {
+    my $minion = $t->app->minion;
+    my $clones = {x => 'y'};
+    my @j = map { $schema->resultset('Jobs')->create({state => 'scheduled', TEST => "t$_"}); } 1 .. 5;
+    my $jobs = [$j[0]->id, $j[1]->id];
+    my $result = $t->app->gru->enqueue_git_clones($clones, $jobs);
+    my $minion_job = $minion->job($result->{minion_id});
+    my $job_id = $minion_job->id;
+    my $task = $schema->resultset('GruTasks')->find($result->{gru_id});
+    my @deps = $task->jobs;
+    is_deeply [map $_->job_id, @deps], $jobs, 'expected GruDependencies created';
+
+    subtest 'add to existing GruTask' => sub {
+        my $enq = 0;
+        my $mocked_gru = Test::MockModule->new('OpenQA::Shared::Plugin::Gru');
+        $mocked_gru->redefine(enqueue => sub (@) { $enq++; });
+        my $jobs = [$j[2]->id];
+        my $result = $t->app->gru->enqueue_git_clones($clones, $jobs);
+        is_deeply $jobs, [], 'job array was emptied';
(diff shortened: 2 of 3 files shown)


perlpunk:
Maybe I could also try to select from GruTasks directly without searching for Minion jobs

Martchus:
> Maybe I could also try to select from GruTasks directly without searching for Minion jobs

That would probably be possible.

perlpunk:
Maybe it is even better to look for Minion jobs including finished ones, e.g. if there is a finished minion job less than 1 minute ago, we don't need to add anything at all.
Because as it is right now, searching for GruTasks, it still might lead to a lot of tasks, because they get deleted immediately when finished.

Another thought: When specifying a certain sha, the task could first check if that ref is already there (which is the case for investigation jobs), and only when it is not, acquire a lock and do the fetch.
Unfortunately the arguments usually include the casedir and the needledir, and even if the casedir has a sha, the needledir usually doesn't, so it would need a lock in any case.
But shortening the locks to a minimal scope could help in general.

lib/OpenQA/Shared/Plugin/Gru.pm
```diff
@@ -115,7 +116,66 @@ sub is_task_active ($self, $task) {
...
+                =~ m/insert or update on table "gru_dependencies" violates foreign key constraint "gru_dependencies_fk_gru_task_id"/i
+              )
+            {
+                # if the GruTask was already deleted meanwhile, we can skip
+                # the rest of the jobs, since the wanted task was done
+                log_debug("GruTask $gru_id already gone, skip assigning jobs (message: $@)");
+                @$jobs = ();
+                last;
+            }
+            die $error;
```
okurz, line 171:
```suggestion
            die $error;  # uncoverable statement
```

lib/OpenQA/Shared/Plugin/Gru.pm
```diff
@@ -115,7 +116,66 @@ sub is_task_active ($self, $task) {
...
+        }
+        shift @$jobs;
```
okurz, line 173:
```suggestion
            my $error = $@;
            die $error unless $error
                =~ m/insert or update on table "gru_dependencies" violates foreign key constraint "gru_dependencies_fk_gru_task_id"/i;
            # if the GruTask was already deleted meanwhile, we can skip
            # the rest of the jobs, since the wanted task was done
            log_debug("GruTask $gru_id already gone, skip assigning jobs (message: $@)");
            @$jobs = ();
            last;
        }
        shift @$jobs;
```

lib/OpenQA/Shared/Plugin/Gru.pm
```diff
@@ -115,7 +116,66 @@ sub is_task_active ($self, $task) {
...
+# enqueued less than a minute ago, like 'git fetch'
+sub _find_existing_gru_task ($self, $task, $args, $jobs) {
+    # warn __PACKAGE__ . ':' . __LINE__ . ": ========================== _find_existing_gru_task($task)\n";
+    my $schema = OpenQA::Schema->singleton;
+    $args = [$args] if ref $args eq 'HASH';
+    my $dtf = $schema->storage->datetime_parser;
+    my $dbh = $schema->storage->dbh;
+    my $sql = q{SELECT id, args, created, state, retries, notes, result FROM minion_jobs
+                WHERE state IN ('inactive', 'active', 'finished')
+                AND created >= ? AND task = ? AND args = ? limit 10};
```
okurz, line 129: As you suggested consider to use "array_position" from https://www.postgresql.org/docs/9.5/functions-array.html

perlpunk:
I'm trying to cover the case where a finished minion job exists.
But I'm having some trouble creating a finished minion job.
I tried this code on a previously enqueued job:
```
$minion_job->execute;
$minion_job->finish;
```
It runs fine, and I see the (mocked) code for the task being executed, but the job is still is marked as `inactive`.
Maybe I have to manipulate the job via SQL, but this seems wrong.

perlpunk:
https://app.circleci.com/pipelines/github/os-autoinst/openQA/15009/workflows/db61e5e1-73ac-4214-a22c-dfdb4650a557/jobs/143274
```
[17:37:15] t/api/04-jobs.t .................... 63/? 
    #   Failed test '200 OK'
    #   at t/api/04-jobs.t line 1627.
    #          got: '400'
    #     expected: '200'
DBIx::Class::ResultSource::_minimal_valueset_satisfying_constraint(): NULL/undef values supplied for requested unique constraint 'primary' (NULL values in column(s): 'id'). This is almost certainly not what you wanted, though you can set DBIC_NULLABLE_KEY_NOWARN to disable this warning. at t/api/04-jobs.t line 1630

#   Failed test 'handle git_clone with CASEDIR'
#   at t/api/04-jobs.t line 1643.
Can't call method "settings_hash" on an undefined value at t/api/04-jobs.t line 1630.

#   Failed test 'no (unexpected) warnings (via END block)'
#   at /usr/lib/perl5/5.26.1/Test/Builder.pm line 135.
# Got the following unexpected warnings:
#   1: DBIx::Class::ResultSource::_minimal_valueset_satisfying_constraint(): NULL/undef values supplied for requested unique constraint 'primary' (NULL values in column(s): 'id'). This is almost certainly not what you wanted, though you can set DBIC_NULLABLE_KEY_NOWARN to disable this warning. at t/api/04-jobs.t line 1630
# Tests were run but no plan was declared and done_testing() was not seen.
# Looks like your test exited with 255 just after 66.
[17:37:15] t/api/04-jobs.t .................... Dubious, test returned 255 (wstat 65280, 0xff00)
Failed 2/66 subtests 
```
this doesn't fail for me locally. investigating...

lib/OpenQA/Shared/Plugin/Gru.pm
```diff
@@ -115,6 +116,54 @@ sub is_task_active ($self, $task) {
...
+sub _find_existing_minion_job ($self, $task, $args, $jobs) {
+    my $schema = OpenQA::Schema->singleton;
+    $args = [$args] if ref $args eq 'HASH';
+    my $dtf = $schema->storage->datetime_parser;
+    my $dbh = $schema->storage->dbh;
+    my $sql = q{SELECT id, args, created, state, retries, notes, result FROM minion_jobs
+                WHERE state IN ('inactive', 'active', 'finished')
+                AND created >= ? AND task = ? AND args = ?
+                ORDER BY array_position(array['finished', 'inactive', 'active'], state::varchar)
+                limiT 1};
```
okurz, line 130:
```suggestion
                limit 1};
```

t/14-grutasks-git.t
```diff
@@ -231,6 +231,61 @@ subtest 'git_update_all' => sub {
...
+
+    subtest 'add to existing GruTask' => sub {
+        my $enq = 0;
+        my $mocked_gru = Test::MockModule->new('OpenQA::Shared::Plugin::Gru');
+        $mocked_gru->redefine(enqueue => sub (@) { $enq++; });
+        my $jobs = [$j[2]->id];
+        my $result = $t->app->gru->enqueue_git_clones($clones, $jobs);
+        is_deeply $jobs, [], 'job array was emptied';
+        my @deps = $task->jobs;
+        is scalar @deps, 3, 'job was aded to GruTask';
```
okurz, line 254:
```suggestion
        is scalar @deps, 3, 'job was added to GruTask';
```

t/14-grutasks-git.t
```diff
@@ -231,6 +231,61 @@ subtest 'git_update_all' => sub {
...
+        $sth->execute($job_id);
+        my $enq = 0;
+        my $mocked_gru = Test::MockModule->new('OpenQA::Shared::Plugin::Gru');
+        $mocked_gru->redefine(enqueue => sub (@) { $enq++; });
+        my $jobs = [$j[3]->id];
+        my $result = $t->app->gru->enqueue_git_clones($clones, $jobs);
+        is_deeply $jobs, [], 'job array was emptied';
+        my @deps = $task->jobs;
+        is scalar @deps, 3, 'no job was aded to GruTask';
+        is $enq, 0, 'enqueue was not called';
```
okurz, line 272: better make those test descriptions unique, e.g.

```suggestion
        is scalar @deps, 3, 'no job was added to GruTask for finished minion job';
        is $enq, 0, 'enqueue was not called for finished minion job';
```

t/14-grutasks-git.t
```diff
@@ -231,6 +231,61 @@ subtest 'git_update_all' => sub {
...
+
+    subtest 'skip GruTask because task done during assigning' => sub {
+        $t->app->log(Mojo::Log->new(level => 'debug'));
+        my $jobs = [$j[4]->id];
+        stderr_like { $t->app->gru->_add_jobs_to_gru_task(999, $jobs) }
+qr{GruTask 999 already gone.*insert or update on table "gru_dependencies" violates foreign key constraint "gru_dependencies_fk_gru_task_id"},
+          'expected log output if GruTask deleted in between';
+        is_deeply $jobs, [], 'job array was emptied';
+        my @deps = $task->jobs;
+        is scalar @deps, 3, 'no job was aded to GruTask';
```
okurz, line 283:
```suggestion
        is_deeply $jobs, [], 'job array was emptied for done task';
        my @deps = $task->jobs;
        is scalar @deps, 3, 'no job was added to GruTask for done task';
```

perlpunk:
https://app.circleci.com/pipelines/github/os-autoinst/openQA/15011/workflows/8c1211b5-b338-4d8f-bec8-a31b671f08b2/jobs/143298
```
    # main:1630 $json = \{
    #             'error' => 'OpenQA::Shared::Plugin::Gru::_find_existing_minion_job(): DBI Exception: DBD::Pg::st execute failed: ERROR:  function array_position(text[], character varying) does not exist
    # LINE 4:                 ORDER BY array_position(array[\'finished\', \'i...
    #                                  ^
    # HINT:  No function matches the given name and argument types. You might need to add explicit type casts. [for Statement "SELECT id, args, created, state, retries, notes, result FROM minion_jobs
    #                 WHERE state IN (\'inactive\', \'active\', \'finished\')
    #                 AND created >= ? AND task = ? AND args = ?
    #                 ORDER BY array_position(array[\'finished\', \'inactive\', \'active\'], state::varchar)
    #                 limiT 1" with ParamValues: 1=\'2024-10-15 19:03:39+0000\', 2=\'git_clone\', 3=\'[{"t\\/data\\/openqa\\/share\\/tests\\/foobar":"http:\\/\\/localhost\\/foo.git","t\\/data\\/openqa\\/share\\/tests\\/foobar\\/needles":"http:\\/\\/localhost\\/bar.git"}]\'] at /home/squamata/project/lib/OpenQA/Shared/Plugin/Gru.pm line 251',
    #             'error_status' => 400
    #           };
```
not sure why it fails in CircleCI and not locally.

perlpunk:
So we need to typecase every array element as well :-/
```
ORDER BY array_position(array['finished'::varchar, 'inactive'::varchar, 'active'::varchar], state::varchar)
```
Maybe the postgres version is different.

t/14-grutasks-git.t
perlpunk: I disagree; that's why we have subtest names. I don't want to repeat myself

okurz: yeah but the verbose output of test is IMHO more readable with unique messages

perlpunk: I still disagree.

okurz (approved):

lib/OpenQA/Shared/Plugin/Gru.pm
```diff
@@ -115,6 +116,54 @@ sub is_task_active ($self, $task) {
...
+        @$jobs = ();
+        return;
+    }
+    $self->_add_jobs_to_gru_task($notes->{gru_id}, $jobs);
+}
+
+sub _add_jobs_to_gru_task ($self, $gru_id, $jobs) {
+    my $schema = OpenQA::Schema->singleton;
+    while (@$jobs) {
+        my $id = $jobs->[0];
```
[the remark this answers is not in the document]
perlpunk: Right, that was actually a leftover from a previous version, where I thought I have to keep the rest of the ids for enqueuing. Updated.

Martchus (approved):

PR #6006, opened 2024-10-14
etc: Fix wrong default value in comment

okurz:
Regression from 2f3ee128f


perlpunk (approved):

baierjan (approved):

etc/openqa/openqa.ini
```diff
@@ -266,7 +266,7 @@ concurrent = 0
 ## Maximum number of online workers (to prevent performance issues)
 #max_online_workers = 1000
 ## Retry delay for limited workers in seconds
-#worker_limit_retry_delay = 3600
+#worker_limit_retry_delay = 900
```
nicksinger, line 269: I like 666 more and I think it is more correct than 900 :)

PR #6010, opened 2024-10-15
Improve worker load threshold detection

okurz:
With https://github.com/os-autoinst/openQA/commit/03b301d1df74857f3b3457477857364255a1671a we have a worker load limit which helps but there can still
be cases like happened on a PowerNV machine with the load going way above
the configured load limit. The reason was that when the worker was idle
within a short time frame of roughly one minute multiple jobs were
assigned to individual worker instances on the machine. As we only
looked at load15 which was still low at that time all jobs were picked
up by the machine leading to overload only about one minute later.
Further reducing the load limit would not realistically prevent this
situation but only delay until load15 decays sufficiently enough so that
new jobs will be picked up again. Instead this commit changes the
evaluation to look at all three system load values, load1, load5 and
load15, but considering the load evolution over time to react quickly
enough if the load rises but still accept a falling edge to allow to
pick up jobs again when the load decays.

Related progress issue: https://progress.opensuse.org/issues/168244

### DIFF
--- a/lib/OpenQA/Worker.pm
+++ b/lib/OpenQA/Worker.pm
@@ -34,6 +34,7 @@ BEGIN {
 use Fcntl;
 use File::Path qw(make_path remove_tree);
 use File::Spec::Functions 'catdir';
+use List::Util qw(all max min);
 use Mojo::IOLoop;
 use Mojo::File 'path';
 use POSIX;
@@ -750,18 +751,24 @@ sub _handle_job_status_changed ($self, $job, $event_data) {
     }
 }
 
-sub _load_avg ($field = 2) {
-    my $value = eval { (split(' ', path($ENV{OPENQA_LOAD_AVG_FILE} // '/proc/loadavg')->slurp))[$field] };
+sub _load_avg ($path = $ENV{OPENQA_LOAD_AVG_FILE} // '/proc/loadavg') {
+    my @load = eval { split(' ', path($path)->slurp) };
     log_warning "Unable to determine average load: $@" if $@;
-    return looks_like_number($value) ? $value : undef;
+    splice @load, 3;    # remove non-load numbers
+    log_error "Unable to parse system load from file '$path'" and return [] unless all { looks_like_number $_ } @load;
+    return \@load;
 }
 
-sub _check_system_utilization ($self) {
-    my $settings = $self->settings->global_settings;
-    return undef unless my $threshold = $settings->{CRITICAL_LOAD_AVG_THRESHOLD};
-    my $load_avg = _load_avg;
-    return "The average load $load_avg is exceeding the configured threshold of $threshold."
-      if defined $load_avg && $load_avg >= $threshold;
+sub _check_system_utilization (
+    $self,
+    $threshold = $self->settings->global_settings->{CRITICAL_LOAD_AVG_THRESHOLD},
+    $load = _load_avg())
+{
+    return undef unless $threshold && @$load >= 3;
+    # look at the load evolution over time to react quick enough if the load
+    # rises but accept a falling edge
+    return "The average load (@$load) is exceeding the configured threshold of $threshold."
+      if max(@$load) > $threshold && ($load->[0] > $load->[1] || $load->[0] > $load->[2] || min(@$load) > $threshold);
     return undef;
 }
 
--- a/t/24-worker-overall.t
+++ b/t/24-worker-overall.t
@@ -30,8 +30,6 @@ $ENV{OPENQA_CONFIG} = "$FindBin::Bin/data/24-worker-overall";
 #       file specified via OPENQA_LOGFILE instead of stdout/stderr.
 $ENV{OPENQA_LOGFILE} = undef;
 
-my $load_avg_file = simulate_load('0.93 0.95 10.25 2/2207 1212', 'worker-overall-load-avg');
-
 # define fake isotovideo
 {
     package Test::FakeProcess;    # uncoverable statement count:1
@@ -90,6 +88,7 @@ my $dbus_mock = Test::MockModule->new('Net::DBus', no_auto => 1);
 $dbus_mock->define(system => sub (@) { Test::FakeDBus->new });
 my $cache_service_client_mock = Test::MockModule->new('OpenQA::CacheService::Client');
 $cache_service_client_mock->redefine(info => sub { Test::FakeCacheServiceClientInfo->new });
+my $load_avg_file = simulate_load('10.93 10.91 10.25 2/2207 1212', 'worker-overall-load-avg');
 
 like(
     exception {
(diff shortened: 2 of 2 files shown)


lib/OpenQA/Worker.pm
```diff
@@ -750,18 +751,22 @@ sub _handle_job_status_changed ($self, $job, $event_data) {
...
 }
 
-sub _load_avg ($field = 2) {
-    my $value = eval { (split(' ', path($ENV{OPENQA_LOAD_AVG_FILE} // '/proc/loadavg')->slurp))[$field] };
+sub _load_avg () {
+    my @load = eval { split(' ', path($ENV{OPENQA_LOAD_AVG_FILE} // '/proc/loadavg')->slurp) };
+    splice @load, 3;    # remove non-load numbers
     log_warning "Unable to determine average load: $@" if $@;
-    return looks_like_number($value) ? $value : undef;
+    return undef unless all { *looks_like_number } @load;
```
perlpunk, line 758: `*looks_like_number`?
```suggestion
    return undef unless all { looks_like_number $_ } @load;
```

lib/OpenQA/Worker.pm
```diff
@@ -750,18 +751,22 @@ sub _handle_job_status_changed ($self, $job, $event_data) {
...
 }
 
 sub _check_system_utilization ($self) {
     my $settings = $self->settings->global_settings;
     return undef unless my $threshold = $settings->{CRITICAL_LOAD_AVG_THRESHOLD};
-    my $load_avg = _load_avg;
-    return "The average load $load_avg is exceeding the configured threshold of $threshold."
-      if defined $load_avg && $load_avg >= $threshold;
+    my @load = _load_avg;
+    return undef unless @load;
```
perlpunk, line 766: if you `return undef` explicitly in `_load_avg`, `@load` will always contain at least 1 element here, so this will never return here.

lib/OpenQA/Worker.pm
perlpunk:
```suggestion
    return () unless all { looks_like_number $_ } @load;
```

okurz: > `*looks_like_number`?

yeah, that's https://perldoc.perl.org/perlapi#looks_like_number , seems to be working, or is this wrong?

perlpunk: the function name is the right one, but not the star in front of it

okurz:
1. Incorporated @perlpunk's suggestions
2. Extended test cases with a dedicated load detection subtest which is significantly faster than extending the worker restarting based tests
3. Extended the algorithm to also handle short-term load dips and plateaus

lib/OpenQA/Worker.pm
```diff
@@ -750,18 +751,25 @@ sub _handle_job_status_changed ($self, $job, $event_data) {
     }
 }
 
-sub _load_avg ($field = 2) {
-    my $value = eval { (split(' ', path($ENV{OPENQA_LOAD_AVG_FILE} // '/proc/loadavg')->slurp))[$field] };
+sub _load_avg () {
+    my @load = eval { split(' ', path($ENV{OPENQA_LOAD_AVG_FILE} // '/proc/loadavg')->slurp) };
+    splice @load, 3;    # remove non-load numbers
     log_warning "Unable to determine average load: $@" if $@;
```
Martchus, line 757: This line using `$@` should preferably run immediately after the eval.

lib/OpenQA/Worker.pm
```diff
@@ -750,18 +751,25 @@ sub _handle_job_status_changed ($self, $job, $event_data) {
...
+    return \@load;
+}
+
+sub _check_system_utilization (
+    $self,
+    $threshold = $self->settings->global_settings->{CRITICAL_LOAD_AVG_THRESHOLD},
+    $load = _load_avg())
+{
+    return undef unless $threshold;
+    return undef unless @$load;
```
Martchus, line 768: I'd combine these conditions and also check that load has at least 3 elements because the subsequent code relies on that:
```suggestion
    return undef unless $threshold && @$load >= 3;
```

Martchus (changes requested):
Looks good, I only have two small nitpicks.

lib/OpenQA/Worker.pm
okurz: Do we need to check if load is defined before checking it's length to prevent perl warnings? Like

```suggestion
    return undef unless $threshold && $load && @$load >= 3;
```

perlpunk: No, you always return an array ref

okurz:
> Looks good, I only have two small nitpicks.

done

lib/OpenQA/Worker.pm
```diff
@@ -750,18 +751,24 @@ sub _handle_job_status_changed ($self, $job, $event_data) {
...
-}
-
-sub _check_system_utilization ($self) {
-    my $settings = $self->settings->global_settings;
-    return undef unless my $threshold = $settings->{CRITICAL_LOAD_AVG_THRESHOLD};
-    my $load_avg = _load_avg;
-    return "The average load $load_avg is exceeding the configured threshold of $threshold."
-      if defined $load_avg && $load_avg >= $threshold;
+    splice @load, 3;    # remove non-load numbers
+    return [] unless all { looks_like_number $_ } @load;
```
nicksinger, line 758: IMHO the user/admin should be warned if the parser failed as well. Potentially pointing to a wrongly configured `OPENQA_LOAD_AVG_FILE`-variable

okurz: done

lib/OpenQA/Worker.pm
```diff
@@ -750,18 +751,24 @@ sub _handle_job_status_changed ($self, $job, $event_data) {
...
+    splice @load, 3;    # remove non-load numbers
+    log_error "Unable to parse system load from file '$path'" and return [] unless all { looks_like_number $_ } @load;
+    return \@load;
+}
+
+sub _check_system_utilization (
+    $self,
+    $threshold = $self->settings->global_settings->{CRITICAL_LOAD_AVG_THRESHOLD},
+    $load = _load_avg())
+{
```
nicksinger, line 766: bad copy&paste?

nicksinger (changes requested):

lib/OpenQA/Worker.pm
okurz: This might look a bit unusual but what is actually happening here is defining the method with three parameters and defining default values plus a bit of unfortunate line wrapping dictated by perl tidy with explicit rules

Martchus: It runs you Perl golfing score but is at least quite readable:-)

asdil12 (approved):

nicksinger (approved):

PR #6011, opened 2024-10-16
Consistently handle settings with multiple values

r-richardson:
As mentioned in: https://progress.opensuse.org/issues/166154

Screenshot From 2024-10-15 23-51-35

### DIFF
--- a/lib/OpenQA/Schema/ResultSet/Jobs.pm
+++ b/lib/OpenQA/Schema/ResultSet/Jobs.pm
@@ -6,6 +6,7 @@ package OpenQA::Schema::ResultSet::Jobs;
 use Mojo::Base 'DBIx::Class::ResultSet', -signatures;
 use DBIx::Class::Timestamps 'now';
 use Date::Format 'time2str';
+use Encode qw(decode_utf8);
 use File::Basename 'basename';
 use IPC::Run;
 use OpenQA::App;
@@ -112,7 +113,7 @@ sub create_from_settings {
     my ($self, $settings, $scheduled_product_id) = @_;
 
     my %settings = %$settings;
-    my %new_job_args = (TEST => $settings{TEST});
+    my %new_job_args;
 
     my $result_source = $self->result_source;
     my $schema = $result_source->schema;
@@ -170,6 +171,12 @@ sub create_from_settings {
             push(@{$new_job_args{parents}}, {parent_job_id => $id, dependency => $dependency_type});
         }
     }
+
+    for my $key (keys %settings) {
+        my $value = $settings{$key};
+        $settings{$key} = decode_utf8 encode_json $value if (ref $value eq 'ARRAY' || ref $value eq 'HASH');
+    }
+    $new_job_args{TEST} = $settings{TEST};
 
     # move important keys from the settings directly to the job
     for my $key (OpenQA::Schema::Result::Jobs::MAIN_SETTINGS) {
--- a/t/10-jobs.t
+++ b/t/10-jobs.t
@@ -14,7 +14,6 @@ use FindBin;
 use lib "$FindBin::Bin/lib", "$FindBin::Bin/../external/os-autoinst-common/lib";
 use Mojo::Base -signatures;
 use autodie ':all';
-use Encode;
 use File::Copy;
 use OpenQA::Jobs::Constants;
 use OpenQA::Test::Case;
--- a/t/10-jobs.t
+++ b/t/10-jobs.t
@@ -25,6 +25,8 @@ use Mojo::JSON qw(decode_json encode_json);
 use OpenQA::Test::Utils qw(perform_minion_jobs);
 use OpenQA::Test::TimeLimit '30';
 
+binmode(STDOUT, ":encoding(UTF-8)");
+
 my $schema_name = OpenQA::Test::Database::generate_schema_name;
 my $schema = OpenQA::Test::Case->new->init_data(
     fixtures_glob => '01-jobs.pl 05-job_modules.pl 06-job_dependencies.pl',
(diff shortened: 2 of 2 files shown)


lib/OpenQA/Schema/ResultSet/Jobs.pm
```diff
@@ -112,6 +112,13 @@ sub create_from_settings {
     my ($self, $settings, $scheduled_product_id) = @_;
 
     my %settings = %$settings;
+    foreach my $key (keys %settings) {
+        my $value = $settings{$key};
+
+        if (ref $value eq 'ARRAY' || ref $value eq 'HASH') {
+            $settings{$key} = encode_json($value);
+        }
+    }
```
okurz, line 121: please try if this works

```suggestion
    my %settings = map { ref $settings->{$_} eq 'ARRAY' || ref $settings->{$_} eq 'HASH' ? encode_json($settings->{$_}) : $settings->{$_} } %$settings;
```

or if you want to use a helper variable for the value you could try

```suggestion
    my %settings = map { my $v = $settings->{$_}; ref $v eq 'ARRAY' || ref $v eq 'HASH' ? encode_json($v) : $v } %$settings;
```

okurz (changes requested):

lib/OpenQA/Schema/ResultSet/Jobs.pm
```diff
@@ -112,6 +112,13 @@ sub create_from_settings {
     my ($self, $settings, $scheduled_product_id) = @_;
 
     my %settings = %$settings;
+    foreach my $key (keys %settings) {
+        my $value = $settings{$key};
+
+        if (ref $value eq 'ARRAY' || ref $value eq 'HASH') {
+            $settings{$key} = encode_json($value);
```
Martchus, line 119: Does this work with non-ASCII characters like umlauts? If not you might need to encode this to UTF-8.

lib/OpenQA/Schema/ResultSet/Jobs.pm
Martchus: I would keep using the helper variable. However, I also think the for-loop you wrote initially is fine.

okurz: we need to think less procedural programming :)

lib/OpenQA/Schema/ResultSet/Jobs.pm
r-richardson: you're right, they are not printed correctly, i'll do that :+1:

lib/OpenQA/Schema/ResultSet/Jobs.pm
r-richardson: getting a `400 Bad Request` for either one of the two suggestions, shouldnt i be modifying `$settings` directly ?

perlpunk: yes, I would work on the `%settings` hash, not the `$settings` hashref parameter directly

okurz (changes requested):
TDD plz

lib/OpenQA/Schema/ResultSet/Jobs.pm
```diff
@@ -171,6 +171,12 @@ sub create_from_settings {
         }
     }
 
+    for my $key (keys %settings) {
+        my $value = $settings{$key};
+        $settings{$key} = encode_json($value) if (ref $value eq 'ARRAY' || ref $value eq 'HASH');
+    }
```
okurz, line 177:
```suggestion
    %settings = map { my $v = $settings{$_}; ref $v eq 'ARRAY' || ref $v eq 'HASH' ? encode_json($v) : $v } %settings;
```

perlpunk: That doesn't work.
You need to iterate over the **keys** of the settings hash, and also you need to return the key **and** the value inside of the map.
We tried your suggestion together and found it not working.
I vote for keeping it like it is.

okurz: so what is a better way to write code then? Using a loop is fine if we need to abort at some point but here we want an altered copy of the complete hash.

okurz: I prefer the for loop over your one-liner suggestion as that looks like not the intended use case for map

lib/OpenQA/Schema/ResultSet/Jobs.pm
```diff
@@ -171,6 +171,12 @@ sub create_from_settings {
         }
     }
 
+    for my $key (keys %settings) {
+        my $value = $settings{$key};
+        $settings{$key} = encode_json($value) if (ref $value eq 'ARRAY' || ref $value eq 'HASH');
```
kalikiana, line 176: Would you mind adding some test coverage? The fact that this doesn't break anything suggests this is untested 😶

okurz (approved):

lib/OpenQA/Schema/ResultSet/Jobs.pm
```diff
@@ -6,6 +6,7 @@ package OpenQA::Schema::ResultSet::Jobs;
 use Mojo::Base 'DBIx::Class::ResultSet', -signatures;
 use DBIx::Class::Timestamps 'now';
 use Date::Format 'time2str';
+use Encode;
```
perlpunk, line 9: I prefer explicitly imported functions:
```suggestion
use Encode qw(decode_utf8);
```

t/10-jobs.t
```diff
@@ -14,7 +14,7 @@ use FindBin;
 use lib "$FindBin::Bin/lib", "$FindBin::Bin/../external/os-autoinst-common/lib";
 use Mojo::Base -signatures;
 use autodie ':all';
-use Encode;
+use Encode qw(decode_utf8);
```
okurz, line 17:
```suggestion
```

okurz (approved):

perlpunk (approved):

kalikiana:
Note this is still affected by [poo#167482](https://progress.opensuse.org/issues/167482) - maybe it didn't have the update yet and a rebase is enough

kalikiana:
@mergifyio rebase

PR #6016, opened 2024-10-17
Avoid inconsistent coverage of drain callback in streaming code

Martchus:
* Ensure code section mentioned in https://progress.opensuse.org/issues/167803 is consistently covered
* See further commit messages for other improvements; this PR will generally improve the coverage of streaming-related code


okurz (approved):

kalikiana (approved):
I guess the finish/tick bits are what makes it reliable. Good catch!

Martchus:
> I guess the finish/tick bits are what makes it reliable.

No, the code section mentioned in the ticket was simply not covered at all by this unit test. It was only covered by the fullstack test and that was not reliable. Finishing the controller is for extending coverage (see codecov report).

The one-tick bits are so that all events are handled before leaving the subtest. Otherwise they'd be handled in the next subtest where certain objects have already been destroyed leading to Perl warnings.

PR #6017, opened 2024-10-17
Do not git fetch if requested sha is already present

perlpunk:
Issue: https://progress.opensuse.org/issues/168400

### DIFF
--- a/lib/OpenQA/Task/Git/Clone.pm
+++ b/lib/OpenQA/Task/Git/Clone.pm
@@ -42,7 +42,7 @@ sub _git_clone_all ($job, $clones) {
     my $ctx = $log->context("[#$job_id]");
 
     # iterate clones sorted by path length to ensure that a needle dir is always cloned after the corresponding casedir
-    for my $path (sort { length($a) <=> length($b) } keys %$clones) {
+    for my $path (sort { length($a) <=> length($b) || $a cmp $b } keys %$clones) {
         my $url = $clones->{$path};
         die "Don't even think about putting '..' into '$path'." if $path =~ /\.\./;
         eval { _git_clone($app, $job, $ctx, $path, $url) };
--- a/lib/OpenQA/Task/Git/Clone.pm
+++ b/lib/OpenQA/Task/Git/Clone.pm
@@ -77,6 +77,8 @@ sub _git_clone ($app, $job, $ctx, $path, $url) {
         $url = $git->get_origin_url;
     }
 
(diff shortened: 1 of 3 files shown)


lib/OpenQA/Task/Git/Clone.pm
```diff
@@ -77,6 +77,8 @@ sub _git_clone ($app, $job, $ctx, $path, $url) {
         $url = $git->get_origin_url;
     }
 
+    return if ($requested_branch and $requested_branch !~ tr/a-f0-9//c and $git->check_sha($requested_branch));
```
perlpunk, line 80: Just don't create branches named `coffee` or `beef` ;-)

perlpunk: I just saw an example `https://github.com/rbmarliere/os-autoinst-distri-opensuse.git#162791` on o3.
So it's probably better to doublecheck the output of rev-parse. if it returns the same string, it's a sha, otherwise it's a branch name.

Martchus: I guess we can live with that limitation - although putting in a regex that would not only check the characters but also the length probably wouldn't be too much of an effort.

kalikiana: If by limitation you mean breaking branch names that could be sha's I vote against it. That would be very confusing in practice.

Martchus: I haven't seen @perlpunk 's last message when I wrote mine. So you can disregard my last message.

perlpunk: I'm now checking if the given sha/branch matches the beginning of the returned sha from rev-parse

kalikiana (approved):

t/14-grutasks-git.t
```diff
@@ -81,6 +85,18 @@ subtest 'git clone' => sub {
                 $return_code = 2
                   if $path =~ m/dirty-error/;
             }
+            elsif ($action eq 'rev-parse') {
+                if ($path =~ m/sha1/) {
+                    $return_code = 0;
+                    $stdout = 'abcdef123456';
+                }
+                if ($path =~ m/sha-branchname/) {
```
Martchus, line 93:
```suggestion
                }
                elsif ($path =~ m/sha-branchname/) {
```

Martchus (approved):

PR #6022, opened 2024-10-22
Prevent big empty liveview in early state of running

r-richardson:
- This PR makes the box significantly smaller while its empty, i've tested displaying it without fixed size and it was still displayed correctly

Related issue: https://progress.opensuse.org/issues/134840

### DIFF
--- a/templates/webapi/test/live.html.ep
+++ b/templates/webapi/test/live.html.ep
@@ -193,7 +193,7 @@
 </div>
 
 <div id="canholder">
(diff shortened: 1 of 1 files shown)


okurz (approved):
This is how it looks in action

[live_view_running_init_pr6022.webm](https://github.com/user-attachments/assets/161c0dfb-6bd2-49fc-bfbc-d1b0bd545284)

PR #6024, opened 2024-10-24
Fix pattern for needle tag

perlpunk:
Developer console showed me this error:

    Pattern attribute value ^[A-Za-z0-9-_]+$ is not a valid regular expression:
    Uncaught SyntaxError: Invalid regular expression: /^[A-Za-z0-9-_]+$/v:
    Invalid character class

Apparently dashes have to be escaped. Moving it to the end of the character class (this would have fixed it for a perl regex) did not help.

### DIFF
--- a/t/ui/12-needle-edit.t
+++ b/t/ui/12-needle-edit.t
@@ -334,6 +334,18 @@ my $needlename = 'test-newneedle';
 my $xoffset = my $yoffset = 200;
 
 subtest 'Create new needle' => sub {
+    subtest 'invalid needle tag is rejected' => sub {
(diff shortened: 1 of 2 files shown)


perlpunk:
I guess a test should be possible. The "Add" button should not be clickable if the pattern doesn't match.

Martchus (commented):
I guess it would make sense to change the title as well but the change itself looks good. Testing could be done in the Selenium-based test of the needle editor but is likely not required for this trivial change.

perlpunk:
Since we actually never tested this, I added a test for it.

kalikiana (approved):

Martchus (approved):

PR #6028, opened 2024-10-28
Handle foreign key violation with savepoint

perlpunk:
Apparently wrapping a potential database error in an eval doesn't prevent DBIx::Class from aborting the transaction.

But with a SAVEPOINT we can wiork around that.
svp_rollback will set back the transaction to the sate from svp_begin.

Issue: https://progress.opensuse.org/issues/124487


Martchus (approved):
It looks like this will work. Did you test whether this actually has an effect in the relevant case? About the unit test: Maybe we can just wrap the call within a transaction within the unit test? Otherwise a nested transaction is probably not too bad - maybe a nested transaction *instead* of the savepoint would also do the trick¹.

---

¹ The [PostgreSQL documentation](https://www.postgresql.org/docs/current/subxacts.html) reads a little bit like that was the case.

perlpunk:
> It looks like this will work. Did you test whether this actually has an effect in the relevant case?

Yes. It's easy to reproduce by doing a `$gru_id+100` in `_add_jobs_to_gru_task` and then cloning two jobs.

> About the unit test: Maybe we can just wrap the call within a transaction within the unit test? Otherwise a nested transaction is probably not too bad - maybe a nested transaction _instead_ of the savepoint would also do the trick¹.

I couldn't find out how to fix it with a nested transaction and without the savepoint. The DBIC docs say it won't do anything if it's nested.

perlpunk:
I wrapped it in an inner `txn_do` now

perlpunk:
According to that https://www.postgresql.org/docs/current/subxacts.html a subtransaction is actually a SAVEPOINT.

kalikiana (approved):

perlpunk:
It seems it is missing codecov checks, so probably a problem with the uploading again.
I'll force push

PR #6038, opened 2024-10-30
Resolve symlinks before enqueuing git_clone GRU task

asdil12:
Ticket: https://progress.opensuse.org/issues/168400

### DIFF
--- a/lib/OpenQA/Shared/Plugin/Gru.pm
+++ b/lib/OpenQA/Shared/Plugin/Gru.pm
@@ -253,8 +253,16 @@ sub enqueue_git_clones ($self, $clones, $job_ids) {
     return unless OpenQA::App->singleton->config->{'scm git'}->{git_auto_clone} eq 'yes';
     # $clones is a hashref with paths as keys and git urls as values
     # $job_id is used to create entries in a related table (gru_dependencies)
-    my $found = $self->_find_existing_minion_job('git_clone', $clones, $job_ids);
-    $self->enqueue('git_clone', $clones, {priority => 10}, $job_ids) unless $found;
+
+    # resolve all symlinks in keys of $clones to allow _find_existing_minion_job find and skip identical jobs
+    my $clones_sr = {};
+    for my $path (keys %$clones) {
+        my $path_sr = eval { path($path)->realpath } // $path;
+        $clones_sr->{$path_sr} = $clones->{$path};
+    }
+
+    my $found = $self->_find_existing_minion_job('git_clone', $clones_sr, $job_ids);
+    $self->enqueue('git_clone', $clones_sr, {priority => 10}, $job_ids) unless $found;
 }
 
 sub enqueue_and_keep_track {
--- a/t/api/04-jobs.t
+++ b/t/api/04-jobs.t
@@ -1636,7 +1636,7 @@ subtest 'handle git_clone with CASEDIR' => sub {
     my $gru_task_values = shift @gru_args;
     is_deeply $gru_task_values,
       {
-        't/data/openqa/share/tests/foobar' => 'http://localhost/foo.git',
+        path('t/data/openqa/share/tests/foobar')->realpath => 'http://localhost/foo.git',
         't/data/openqa/share/tests/foobar/needles' => 'http://localhost/bar.git'
       },
       'the git clone gru tasks were created correctly';
@@ -1671,8 +1671,8 @@ subtest 'handle git_clone without CASEDIR' => sub {
       [
         [
             {
-                't/data/openqa/share/tests/needles' => undef,
-                't/data/openqa/share/tests' => undef,
+                path('t/data/openqa/share/tests')->realpath => undef,
+                path('t/data/openqa/share/tests/needles')->realpath => undef,
             }]
       ],
       'the git_clone gru tasks was created correctly';


lib/OpenQA/Shared/Plugin/Gru.pm
```diff
@@ -253,8 +253,16 @@ sub enqueue_git_clones ($self, $clones, $job_ids) {
     return unless OpenQA::App->singleton->config->{'scm git'}->{git_auto_clone} eq 'yes';
     # $clones is a hashref with paths as keys and git urls as values
     # $job_id is used to create entries in a related table (gru_dependencies)
-    my $found = $self->_find_existing_minion_job('git_clone', $clones, $job_ids);
-    $self->enqueue('git_clone', $clones, {priority => 10}, $job_ids) unless $found;
+
+    # resolve all symlinks in keys of $clones to allow _find_existing_minion_job find and skip itentical jobs
```
perlpunk, line 257:
```suggestion
    # resolve all symlinks in keys of $clones to allow _find_existing_minion_job find and skip identical jobs
```

lib/OpenQA/Shared/Plugin/Gru.pm
```diff
@@ -253,8 +253,16 @@ sub enqueue_git_clones ($self, $clones, $job_ids) {
...
+    my $clones_sr = {};
+    for my $path (keys %$clones) {
+        my $path_sr = (-e $path) ? Mojo::File->new($path)->realpath : $path;    # resolve symlinks
```
Martchus, line 260: This way there's no extra stat for `-e` and also no race condition (as the existence of the file might change between `-e` and `->realpath`):
```suggestion
        my $path_sr = eval { path($path)->realpath } // $path;
```

I'd also get rid of the comment because it is quite redundant.

Martchus (commented):
Looks good. Of course it would be nice to have tests for it.

EDIT: Judging by the CI failures you can just adapt existing tests.

lib/OpenQA/Shared/Plugin/Gru.pm
asdil12: Hm - if there is the possibility of race conditions because multiple calls of `enqueue_git_clones` could run at the samt time, there would need to be a mutex guard for the whole block:

```
  257     # resolve all symlinks in keys of $clones to allow _find_existing_minion_job find and skip identical jobs
  258     my $clones_sr = {};
  262     for my $path (keys %$clones) {
  263         my $path_sr = eval { path($path)->realpath } // $path;
  264         $clones_sr->{$path_sr} = $clones->{$path};
  266     }
  267
  268     my $found = $self->_find_existing_minion_job('git_clone', $clones_sr, $job_ids);
  269     $self->enqueue('git_clone', $clones_sr, {priority => 10}, $job_ids) unless $found;
```

otherwise the symlink resolving could change behaviour depending on paths existing or not and also the `_find_existing_minion_job` method might race against `enqueue` of another thread.

Is this a scenario that we need to cover?

perlpunk: We still have the minion job locking per directory, so two minion jobs with the same args won't be a problem.

Minimizing the number of minion jobs with the same  arguments is just best effort - it can always happen because we don't check for that in the transaction. In fact, I've seen multiple identical jobs on OSD several times in the last days. That seems to happen when jobs are quickly created after each other and all of them don't find an existing job in parallel and then create one.

t/api/04-jobs.t
```diff
@@ -1636,7 +1636,7 @@ subtest 'handle git_clone with CASEDIR' => sub {
     my $gru_task_values = shift @gru_args;
     is_deeply $gru_task_values,
       {
-        't/data/openqa/share/tests/foobar' => 'http://localhost/foo.git',
+        path('t/data/openqa/share/tests/foobar')->realpath => 'http://localhost/foo.git',
         't/data/openqa/share/tests/foobar/needles' => 'http://localhost/bar.git'
```
asdil12, line 1640: If anyone wonders why I don't expand the path for needles,
let me explain that `Mojo::File` behaves a bit odd.
Actually neither of those paths exist. But if at least the path up to the last path element (e.g. `t/data/openqa/share/tests/`) exists, `path()` will not raise an error when being called with an unexisting path like `t/data/openqa/share/tests/foobar`.

```
perl> use Mojo::File qw(path);
perl> path("t/data/openqa/share/tests/foobar")->realpath;p
/home/dheidler/openqa/repos/openQA/t/data/openqa/share/tests/foobar
perl> path("t/data/openqa/share/tests/foobar/needles")->realpath;
Invalid path at (eval 36) line 1.
```

```
$ ls -l t/data/openqa/share/tests/
insgesamt 0
-rw-r--r-- 1 dheidler users  0 30. Okt 2023  archive.tar.xz
drwxr-xr-x 1 dheidler users 14 12. Sep 2022  archlinux
drwxr-xr-x 1 dheidler users 50 30. Sep 11:38 example
drwxr-xr-x 1 dheidler users 14 12. Sep 2022  fedora
drwxr-xr-x 1 dheidler users 56 18. Okt 16:41 opensuse
drwxr-xr-x 1 dheidler users 18 30. Okt 2023  Unicorn

perlpunk (approved):

Martchus (approved):

PR #6039, opened 2024-10-30
Enable automatic updates of test code and needles by default

Martchus:
See individual commit messages and https://progress.opensuse.org/issues/168379

### DIFF
--- a/docs/GettingStarted.asciidoc
+++ b/docs/GettingStarted.asciidoc
@@ -111,6 +111,10 @@ distri:: a test distribution but also sometimes referring to a _product_
 multiple _test modules_ in a folder structure that compose _test suites_, e.g.
 "opensuse" (test distribution, short for "os-autoinst-distri-opensuse")
 
+needles:: reference images to assert whether what is on the screen matches
+expectations and to locate elements on the screen the tests needs to interact
+with (e.g. to locate a button to click on it)
+
 product:: the main "system under test" (SUT), e.g. "openSUSE", also called
 "Medium Types" in the web interface of openQA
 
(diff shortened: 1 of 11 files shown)


okurz (approved):

Martchus:
@phil-hands @AdamWill @marmarek @sdclarke In case you so far have e.g. a cron job to update Git repositories you might want to disable that after this PR has been merged. openQA will now take care of it by default. If you encounter problems or simply don't want to change your setup you can always disable the auto update feature (and maybe create a ticket/PR to tell us why it isn't working for you).

perlpunk (approved):

AdamWill:
we'd definitely disable this on our staging/lab instance (though I think it got reverted, right?), as we frequently deploy a dirty work git branch on that instance; these are typical developer work branches that get force-pushed a lot and need to get updated with `git fetch` / `git reset --hard`, `git pull` will just fail on them. also we sometimes just test stuff out by doing live edits on that instance.

we, er, do both of those things on prod too, sometimes. yes, I know, terrible practice. :P but yeah, for those reasons I'll probably turn this off for both Fedora instances actually.

AdamWill:
oh, it got reverted then reapplied in february. then yeah, we'll disable this I think.

AdamWill:
Hmm. Now I dig deeper into this, I see what looks like a bigger problem.

The `git_auto_update` setting is within the `[scm git]` config section, which is *meant* to be configuration specific to the case where `scm = git` is set. This, to me, implies that none of this automatic updating stuff should happen if `scm = git` is *not* set. But AFAICS nothing ever checks that. `create_git_clone_list` doesn't check it, the things that call that sub don't check it, and AFAICS nothing in the chain down from there to ultimately creating a gru task that will *do* the git clone checks it.

The only things that actually check it seem to be the older needle-saving-related code - `lib/OpenQA/Schema/Result/Needles.pm` and `lib/OpenQA/Task/Needle/Save.pm`. These both do `my $git = OpenQA::Git->new(app => $app, dir => $path);` and then wrap subsequent work in a `if ($git->enabled)` check. But `lib/OpenQA/Task/Git/Clone.pm` `_git_clone_all`, which is what ultimately backs the gru `git_clone` job and so all of *waves hands* this, never checks `$git->enabled`.

I'd think that a) `_git_clone_all` *should* check whether git is actually "enabled" and bail if not, and probably b) there should be checks much earlier in this whole flow to bail out if git is not "enabled".

AdamWill:
hmm, but then that's awkward, because you want `git_auto_update` to be enabled by default, but `scm` is *not* set to `'git'` by default, and doing so would *also* enable the 'commit needles created in the editor to git' functionality by default when that is not currently the case.

this feels like a bit of a mess :/

Martchus:
I think you're right. We should probably fix this mess so none of these settings are used unless git support is enabled in general. A PR is welcome of course.

AdamWill:
yeah, I'm working on that now. It involves quite a bit of detail, though. Hoping to have it done by end of day.

AdamWill:
ugh, this keeps getting messier. I'm wondering about the intersection of `NEEDLES_DIR` with this older automatic-needle-commit behaviour. What happens if you create a needle from a test with `NEEDLES_DIR` set and `scm = git` also set? Does the needle get committed back to...a temporary directory? How is this supposed to work?

PR #6040, opened 2024-10-30
t: Fix sporadic stale element failures in t/ui/18-tests-details.t

perlpunk:
Issue: https://progress.opensuse.org/issues/169096

Calling a URL with a module step makes an ajax call, so we need `wait_for_ajax` here.

It passed 100 times in a row now.
Maybe @r-richardson  can try it out too.

### DIFF
--- a/t/ui/18-tests-details.t
+++ b/t/ui/18-tests-details.t
@@ -78,10 +78,11 @@ sub current_tab { $driver->find_element('.nav.nav-tabs .active')->get_text }
 sub find_candidate_needles {
     # ensure the candidates menu is visible
     my $candidates_menu = wait_for_element(selector => '#candidatesMenu', is_displayed => 1) or return {};
+    return {} unless $candidates_menu->is_enabled;
+
     # save implicit waiting time as long as we are only looking for elements
     # that should be visible already
     disable_timeout;
-    return {} unless $candidates_menu->is_enabled;
     $candidates_menu->click;
 
     # read the tags/needles from the HTML structure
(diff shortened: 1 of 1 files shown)


t/ui/18-tests-details.t
```diff
@@ -81,7 +81,10 @@ sub find_candidate_needles {
     # save implicit waiting time as long as we are only looking for elements
     # that should be visible already
     disable_timeout;
-    return {} unless $candidates_menu->is_enabled;
+    unless ($candidates_menu->is_enabled) {
+        enable_timeout;
+        return {};
+    }
```
Martchus, line 87: Good catch.

I don't think disabling the timeout has any effect on `is_enabled` here. So we can probably just move the early return before `disable_timeout`.

Martchus (commented):
I suppose the expectation was that `my $candidates_menu = wait_for_element(selector => '#candidatesMenu', is_displayed => 1) or return {};` will wait until the ajax call is done because only then the element will be there. So there's no need to explicitly wait for the ajax call.

If this PR makes a difference that means the element might appear before the ajax call has finished and when the ajax call finishes the element might be replaced again. I suppose that's a valid theory so this PR might actually fix the problem.

> It passed 100 times in a row now.

And could you reproduce the issue before (in the environment where it now passes 100 times)?

t/ui/18-tests-details.t
perlpunk: Yes, I could reproduce it before, usually after less than 20 runs.

Then I added the `enable_timeout` thing before the retun, and from then on it started to fail even more often, inside of the `wait_for_element`, which is when I started looking for other places where we call a module step and saw that we always do a `wait_for_ajax` there.

Martchus (approved):

t/ui/18-tests-details.t
perlpunk: I changed it as you suggested and moved the `disable_timeout` below

r-richardson (approved):
Looks good, i also just ran it 100 times without a failure :+1: Thank you!

PR #6044, opened 2024-11-04
Have filter options only affect finished jobs

r-richardson:
This PR ensures the result filter options only affects the finished jobs table, meaning the scheduled and running jobs tables are no longer affected.

Related Issue: https://progress.opensuse.org/issues/65205

example_screenshot

### DIFF
--- a/assets/javascripts/tests.js
+++ b/assets/javascripts/tests.js
@@ -402,6 +402,10 @@ function renderTestLists() {
 
   // add a handler for the actual filtering
   $.fn.dataTable.ext.search.push(function (settings, data, dataIndex) {
+    if (settings.nTable.getAttribute('id') !== 'results') {
+      return true; // Do not filter other tables
+    }
+
     var selectedResults = finishedJobsResultFilter.find('option:selected');
     // don't apply filter if no result is selected
     if (!selectedResults.length) {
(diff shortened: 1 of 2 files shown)


assets/javascripts/tests.js
```diff
@@ -402,6 +402,10 @@ function renderTestLists() {
 
   // add a handler for the actual filtering
   $.fn.dataTable.ext.search.push(function (settings, data, dataIndex) {
+    if (settings.nTable.getAttribute('id') !== 'results') {
```
perlpunk, line 405: I'm wondering whether this could be improved. This function is still called for every entry of each table. Can it be registered only for the finished table?
I don't know myself right now how to do it, and it's not really a problem, as it returns immediately, but maybe you could have a look?

perlpunk (approved):

assets/javascripts/tests.js
r-richardson: I had a look but didn't yet find a way to register a such a function exclusively to a specific table, maybe `afnFiltering` might work

okurz (approved):

t/ui/10-tests_overview.t
```diff
@@ -607,6 +607,19 @@ subtest "job dependencies displayed on 'Test result overview' page" => sub {
       'parent job was highlighted correctly';
 };
 
+subtest 'result filter does not affect scheduled and running jobs' => sub {
```
perlpunk, line 610: I didn't look close enough - I think this test belongs in `t/ui/01-list.t`.
In `t/ui/10-tests_overview.t` we are testing `/tests/overview`, and your added test is the only one testing the `/tests` "all tests" page.

PR #6047, opened 2024-11-05
Extend variable expansion to consider variables from worker config

Martchus:
* Allow substituting placeholders that refer to variables that are only
  defined in `workers.ini` for the particular worker slot that picked up
  the job
* Keep placeholders of non-existing keys when doing the variable
  substitution during job creation on the web UI side (instead of removing
  them) so the worker still sees the placeholder and can do a final
  substitution (removing placeholders of non-existing keys, so the overall
  behavior does not change)
* Do *not* propagate settings from the final substitution back to the web
  UI as the job might be restarted and then run on a different worker where
  settings might be different
* Extend relevant documentation
* See https://progress.opensuse.org/issues/169159

---

@foursixnine @okurz  I have not implemented a default value as suggested in https://progress.opensuse.org/issues/169159 to avoid further complexity. This has already turned out to be a little bit more involved than I expected without it. However, let me know if you need it. Since we're talking about OSD here where all workers are Salt-controlled it is not a big deal to simply set the variable on all workers, though. Alternatively one could also set a default within OSADO unless this is also about ISO/HDD assets (because for those the cache service already needed to know the default).

### DIFF
--- a/docs/UsersGuide.asciidoc
+++ b/docs/UsersGuide.asciidoc
@@ -193,19 +193,22 @@ by different means:
 
 
 === Variable expansion
+Any job setting can refer to another variable using this syntax: `%NAME%`. When
+the test job is created, the string will be substituted with the value of the
+specified variable at that time.
 
-Any variable defined in Test Suite, Machine, Product or Job Template table can
-refer to another variable using this syntax: `%NAME%`. When the test job is created,
-the string will be substituted with the value of the specified variable at that time.
+The variable expansion applies to job settings defined in test suites, machines,
+products and job templates. It also applies to job settings specified when
+creating a single set of jobs.
 
-For example this variable defined for Test Suite:
+Consider this example where a variable is defined within a test suite:
 
 [source,sh]
 --------------------------------------------------------------------------------
 PUBLISH_HDD_1 = %DISTRI%-%VERSION%-%ARCH%-%DESKTOP%.qcow2
 --------------------------------------------------------------------------------
 
-may be expanded to this job variable:
+It may expanded to this job setting:
 
 [source,sh]
 --------------------------------------------------------------------------------
--- a/docs/UsersGuide.asciidoc
+++ b/docs/UsersGuide.asciidoc
@@ -199,7 +199,7 @@ specified variable at that time.
 
 The variable expansion applies to job settings defined in test suites, machines,
 products and job templates. It also applies to job settings specified when
-creating a single set of jobs.
+creating a single set of jobs and to variables specified in the worker config.
 
 Consider this example where a variable is defined within a test suite:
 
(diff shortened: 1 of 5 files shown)


docs/UsersGuide.asciidoc
```diff
@@ -193,28 +193,36 @@ by different means:
...
+Any job setting can refer to another variable using this syntax: `%NAME%`. When
+the test job is created, the string will be substituted with the value of the
+specified variable at that time.
 
-Any variable defined in Test Suite, Machine, Product or Job Template table can
-refer to another variable using this syntax: `%NAME%`. When the test job is created,
-the string will be substituted with the value of the specified variable at that time.
+The variable expansion applies to job settings defined in test suites, machines,
+products and job templates. It also applies to job settings specified when
+creating a single set of jobs and to variables specified in the worker config.
```
foursixnine, line 202:
```suggestion
creating a single set of jobs and to variables specified in the worker config at the time of job exectution.
```

okurz (approved):
Awesome 😎

docs/UsersGuide.asciidoc
```diff
@@ -193,28 +193,36 @@ by different means:
 
 
 === Variable expansion
+Any job setting can refer to another variable using this syntax: `%NAME%`. When
+the test job is created, the string will be substituted with the value of the
+specified variable at that time.
```
foursixnine, line 198:
```suggestion
openQA supports variable expansion, surrounding a test variable with percentage signs (`%`), e.g `%NAME%`.

Variable expansion can happen only at two stages:

1. When the job is created the string will be substituted with the value of the
specified variable at that time. This is useful to handle static information, like a directory or a repository.
1. Before the download state of the worker, making it a runtime value, that will not be shown in the openQA webUI, but is shown in the logs. i.e runtime provisioning of ephemeral values, or using different values defined at the worker level.
```

docs/UsersGuide.asciidoc
foursixnine: not sure if this applies, if the above is added.

docs/UsersGuide.asciidoc
foursixnine: I would go as far as to say that also protected values like `SECRET` or `_PASSWORD` would also be a great case for https://progress.opensuse.org/issues/109019 and https://progress.opensuse.org/issues/105624

but for now that's a random crazy idea

lib/OpenQA/JobSettings.pm
```diff
@@ -52,29 +52,32 @@ sub generate_settings ($params) {
...
+        };
         return "Error: $@" if $@;
     }
     return undef;
 }
 
-sub _expand_placeholder ($settings, $key, $start, $end, $visited_placeholders_in_parent_scope) {
-    return '' unless defined $settings->{$key};
+sub _expand_placeholder ($settings, $key, $start, $end, $visited_placeholders_in_parent_scope, $on_web_ui) {
+    # handle %CASEDIR% only on web UI; on the worker it is handled by _engine_workit_step_2 separately
```
foursixnine, line 68: This should be added in the documentation above

foursixnine:
@Martchus I won't block on docs, feel free to create follow ups if you see fit.

lib/OpenQA/JobSettings.pm
Martchus: No, it is already documented in a more fitting section.

docs/UsersGuide.asciidoc
Martchus: I wouldn't make this so long and avoid generic expressions like "the string". However, maybe I'll incorporate some parts of it.

docs/UsersGuide.asciidoc
Martchus: I wanted to keep it short. Our documentation is long enough and the addition seems redundant.

b10n1k (approved):

PR #6048, opened 2024-11-06
Fix enqueueing of Minion jobs breaking `PARALLEL_ONE_HOST_ONLY=1`

Martchus:
When restarting openQA jobs, additional Minion jobs are enqueued (of
`git_clone` task, this is happening especially often when `git_auto_clone`
is enabled). So far this didn't happen in a transaction. So the scheduler
might see the openQA jobs but not the Minion jobs they are blocked by. This
is problematic because the scheduler might assign jobs too soon to a
worker. This is especially problematic if it does not happen consistently
across a parallel job cluster as it then breaks the
`PARALLEL_ONE_HOST_ONLY=1` feature.

Related ticket: https://progress.opensuse.org/issues/169342

### DIFF
--- a/lib/OpenQA/Schema/Result/Jobs.pm
+++ b/lib/OpenQA/Schema/Result/Jobs.pm
@@ -11,7 +11,7 @@ use DateTime;
 use OpenQA::Constants qw(WORKER_COMMAND_ABORT WORKER_COMMAND_CANCEL);
 use OpenQA::Log qw(log_trace log_debug log_info log_warning log_error);
 use OpenQA::Utils (
-    qw(parse_assets_from_settings locate_asset),
+    qw(create_git_clone_list parse_assets_from_settings locate_asset),
     qw(resultdir assetdir read_test_modules find_bugref random_string),
     qw(run_cmd_with_log_return_error needledir testcasedir gitrepodir find_video_files)
 );
@@ -693,18 +693,26 @@ sub _create_clones ($self, $jobs, $comments, $comment_text, $comment_user_id, @c
         $res->register_assets_from_settings;
     }
 
+    my $app = OpenQA::App->singleton;
+    my %git_clones;
+    my @clone_ids;
     for my $original_job_id (@original_job_ids) {
         my $cloned_job = $clones{$original_job_id};
         # calculate blocked_by
         $cloned_job->calculate_blocked_by;
         # add a reference to the clone within $jobs
-        $jobs->{$original_job_id}->{clone} = $cloned_job->id;
+        push @clone_ids, $jobs->{$original_job_id}->{clone} = $cloned_job->id;
+        # add Git repositories to clone
+        create_git_clone_list($cloned_job->settings_hash, \%git_clones) if $app;
     }
 
     # create comments on original jobs
     $result_source->schema->resultset('Comments')
       ->create_for_jobs(\@original_job_ids, $comment_text, $comment_user_id, $comments)
       if defined $comment_text;
+
+    # enqueue Minion jobs to clone required Git repositories
+    $app->gru->enqueue_git_clones(\%git_clones, \@clone_ids) if $app;
 }
 
 # internal (recursive) function for duplicate - returns hash of all jobs in the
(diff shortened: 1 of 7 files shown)


lib/OpenQA/Schema/Result/Jobs.pm
```diff
@@ -693,18 +693,25 @@ sub _create_clones ($self, $jobs, $comments, $comment_text, $comment_user_id, @c
...
+        create_git_clone_list($cloned_job->settings_hash, \%git_clones);
     }
 
     # create comments on original jobs
     $result_source->schema->resultset('Comments')
       ->create_for_jobs(\@original_job_ids, $comment_text, $comment_user_id, $comments)
       if defined $comment_text;
+
+    # enqueue Minion jobs to clone required Git repositories
+    OpenQA::App->singleton->gru->enqueue_git_clones(\%clones, \@clone_ids);
```
perlpunk, line 714: This would create a minion job for every job, only cluster jobs would be grouped together, right?
That would be way too many jobs in some cases.
The detection for identical git_clone tasks is not perfect, especially if they are created quickly after each other.

Martchus: Otherwise we really needed a "preparing" state.

I suppose it would work like this:

1. We create jobs and now the initial state is "preparing" instead of "scheduled" keeping track of the job IDs.
2. We enqueue Minion jobs.
3. We set all jobs we haven't created Minion jobs for in step 2 to "scheduled" immediately.
4. Before deleting GRU tasks we set related jobs to "preparing". (This would still not fix the scheduling problem when there's a different set of e.g. download jobs within a parallel cluster - unless we take job dependencies into account here.)

Step 2 and 3 would happen in one transaction. Step 4 would also happen in one transaction.

Martchus: For the sake of simplicity we could also reduce the number of Minion jobs on job restarts by avoiding Minion jobs for `git_auto_update`. Of course then `git_auto_update` would rely only on the periodic updates for restarted jobs.

perlpunk: I don't understand step 4. Maybe you can explain tomorrow after the daily?

OTOH we could try this PR and see how it works out in practice.

Theoretically we could find out if any new minion job is required by keeping track of the git directories in the `%git_clones` hash while iterating over the jobs to restart. I guess in most cases we will only have one CASEDIR/NEEDLES_DIR or DISTRI.
Just that we would need to pass an additional `%git_clones_not_yet_enqueued` hash through from the toplevel job_restart, which doesn't sound nice.

perlpunk: > For the sake of simplicity we could also reduce the number of Minion jobs on job restarts by avoiding Minion jobs for `git_auto_update`. Of course then `git_auto_update` would rely only on the periodic updates for restarted jobs.

That was a requirement though. We did have a complain from someone who restarted a job and was wondering why it wasn't using the updated code.

perlpunk: btw, I guess this should be:
```suggestion
    OpenQA::App->singleton->gru->enqueue_git_clones(\%git_clones, \@clone_ids);
```
Probably the reason for the test failures

Martchus: We discussed introducing a new initial state as alternative.

I guess there would be another alternative: Restart all jobs in one transaction (and also create all the Minion jobs in one transaction). Since we'd just append to the job, job settings and minion jobs tables I don't think this would be problematic (so I don't think it would cause deadlocks/conflicts). Then this route would also get a nice "all or nothing" behavior.

Doing all in one transaction would probably be simpler than introducing a new state and it is essentially what we already do when scheduling products (where also many jobs are created). Of course the scheduling has an async mode where it runs as Minion job so it is a bit different.

Creating a new state means changing many places in the code. It is probably the less fragile approach, though (if implemented correctly / all bugs are ironed out).

So I'm a bit torn here.

okurz (approved):

perlpunk (approved):

perlpunk:
https://app.circleci.com/pipelines/github/os-autoinst/openQA/15641/workflows/85530f36-690e-4ee8-81dc-e37118fa559a/jobs/150163
```
[16:31:45] t/ui/18-tests-details.t .................... [WARN] Duplication rolled back after error: Can't call method "config" on an undefined value
Can't use string ("An internal error occurred when "...) as a HASH ref while "strict refs" in use at t/ui/18-tests-details.t line 50.
[16:31:45] t/ui/18-tests-details.t .................... Dubious, test returned 255 (wstat 65280, 0xff00)
No subtests run 
```

PR #6049, opened 2024-11-06
Avoid scheduling jobs if not all parallel jobs are ready

Martchus:
Some jobs of a parallel cluster can be blocked (by a chained parent or by a pending Gru task) while some jobs can be scheduled. Before this change the scheduler assigns the jobs that can be scheduled which creates a half- scheduled parallel cluster.

This is particularly problematic when `PARALLEL_ONE_HOST_ONLY=1` and `git_auto_update = yes` are used because then repairing half-scheduled clusters is more challenging and the likeliness that a cluster is partially blocked is higher. In theory this is also problematic when jobs within the parallel cluster depend on different asset downloads.

With this change the scheduler skips the whole cluster until all jobs can be scheduled to avoid half-scheduled clusters. Judging by the previous code and its comments this was already intended but the code didn't actually work correctly.

I tested the rewritten version of the code by creating a half-blocked cluster manually and observed the behavior of the scheduler before and after this commit.

Related ticket: https://progress.opensuse.org/issues/169342

### DIFF
--- a/lib/OpenQA/Scheduler/Model/Jobs.pm
+++ b/lib/OpenQA/Scheduler/Model/Jobs.pm
@@ -95,8 +95,11 @@ sub _allocate_jobs ($self, $free_workers) {
         next if $checked_jobs{$j->{id}};
         next unless @{$j->{matching_workers}};
         my $tobescheduled = _to_be_scheduled($j, $scheduled_jobs);
+        if (!defined $tobescheduled) {
+            log_debug "Skipping job $j->{id} because dependent jobs are not ready";
+            next;
+        }
         next if defined $allocated_jobs->{$j->{id}};
-        next unless $tobescheduled;
         OpenQA::Scheduler::WorkerSlotPicker->new($tobescheduled)->pick_slots_with_common_worker_host;
         my @tobescheduled = grep { $_->{id} } @$tobescheduled;
         my $parallel_count = scalar(@tobescheduled);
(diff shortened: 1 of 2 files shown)


t/05-scheduler-dependencies.t
```diff
@@ -1278,4 +1278,18 @@ subtest 'starvation of parallel jobs prevented' => sub {
...
+subtest 'partially blocked clusters are not scheduled' => sub {
+    # assume parallel child with ID 2 is blocked by removing it from the set of mocked jobs
+    delete $mocked_jobs{2};
+
+    my ($allocated_workers, $allocated_jobs);
+    my $free_workers = OpenQA::Scheduler::Model::Jobs::determine_free_workers();
+    combined_like {
+        ($allocated_workers, $allocated_jobs) = OpenQA::Scheduler::Model::Jobs->singleton->_allocate_jobs($free_workers)
+    }
+    qr/Skipping job .* because dependent jobs are not ready/, 'skippinng job if dependent jobs not ready';
```
okurz, line 1290:
```suggestion
    qr/Skipping job .* because dependent jobs are not ready/, 'skipping job if dependent jobs not ready';
```

lib/OpenQA/Scheduler/Model/Jobs.pm
```diff
@@ -95,8 +95,11 @@ sub _allocate_jobs ($self, $free_workers) {
         next if $checked_jobs{$j->{id}};
         next unless @{$j->{matching_workers}};
         my $tobescheduled = _to_be_scheduled($j, $scheduled_jobs);
+        if (!defined $tobescheduled) {
```
b10n1k, line 98: correct me but I think `_to_be_scheduled_recurse` doesnt return undef. shouldnt that be just `!$tobescheduled`?

b10n1k (approved):

PR #6055, opened 2024-11-12
Avoid blocking openQA jobs on Git updates

Martchus:
* Unblock openQA jobs if Git updates failed for two times
    * Change the behavior with `git_auto_update = yes` to be more like it
      was with the `fetchneedles` approach (that also did not block jobs in
      case of failures)
    * Avoid running into timeouts in openQA-in-openQA tests which expect
      jobs to be scheduled in a more timely manner, see
      https://progress.opensuse.org/issues/169204
* Keep blocking openQA jobs if any of the repositories required by the
  scheduled jobs was explicitly specified

---

~~Still a draft as I still need to restrict it to network issues.~~

### DIFF
--- a/etc/openqa/openqa.ini
+++ b/etc/openqa/openqa.ini
@@ -117,6 +117,10 @@
 #git_auto_clone = yes
 ## enable automatic updates of all test code and needles managed via Git (still experimental, currently still breaks scheduling parallel clusters)
 #git_auto_update = no
+## specifies how to handle errors on automatic updates via git_auto_update
+## - when set to "best-effort" openQA jobs are started even if the update failed
+## - when set to "strict" openQA jobs will be blocked until the update succeeded or set to incomplete when the retries for updating are exhausted
+#git_auto_update_method = best-effort
 
 ## Authentication method to use for user management
 [auth]
--- a/etc/openqa/openqa.ini
+++ b/etc/openqa/openqa.ini
@@ -53,9 +53,9 @@
 ## web UI. Default is 'repo'
 #hide_asset_types = repo iso
 
-## Recognized referers contains list of hostnames separated by space. When
-## openQA detects (via 'Referer' header) that test job was accessed from
-## this domain, this job is labeled as linked and considered as important
+## space-separated list of domains recognized by job labeling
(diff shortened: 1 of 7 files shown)


etc/openqa/openqa.ini
```diff
@@ -117,6 +117,8 @@
 #git_auto_clone = yes
 ## enable automatic updates of all test code and needles managed via Git (still experimental, currently still breaks scheduling parallel clusters)
 #git_auto_update = no
+## whether automatic updates via git_auto_update are considered best effort, meaning openQA jobs can be scheduled even if fetching is not possible
+#git_auto_update_best_effort = yes
```
okurz, line 121: How about

```suggestion
git_auto_update_method = best-effort
```

okurz (changes requested):

okurz (approved):

t/14-grutasks-git.t
```diff
@@ -266,11 +272,36 @@ subtest 'git_update_all' => sub {
...
+
+        # retry the job now
+        $minion->foreground($job->id);
+        is $job->info->{state}, 'inactive', 'job failed but again set to inactive to be retried';
+        is $gru_task->jobs->count, 0, 'openQA job unblocked after second try';
+
+        combined_like { $minion->foreground($job->id) } qr/fake disconnect/, 'error logged';
+        is $job->info->{state}, 'failed', 'job failed for real after retries exhausted';
+      }
+      if $gru_task;
```
kalikiana, line 304: Looks like the unit test never sets `git_auto_update_method` so `strict` is never tested?

Martchus: That's true. Just like we never test for "no" in "yes"/"no" settings. Everything that is not "yes" is simply "no". And here it is similar, so everything that is not "best-effort" is "strict". I could of course add some kind of validation in the config parser and print a warning for invalid values. (That's not how we usually handle this, though.)

kalikiana: Ah. I wasn't thinking about validation. Just to cover the case where it is set to `strict` and doesn't use the new logic.

kalikiana (approved):

PR #6056, opened 2024-11-14
webapi: allow dots in needle tags

marmarek:
And also put dash at the end instead of escaping.


Martchus (approved):

marmarek:
> And also put dash at the end instead of escaping.

Uhm, apparently it isn't how it works in js... Reverted to escaping.

okurz (approved):

PR #6063, opened 2024-11-19
Add bash and zsh completion scripts for openqa-cli

d3flex:
* Allow static completion of the openqa-cli
* Provide full cli options for each subcommand
* Group options for each subcommand
* Not help text in bash due to `compgen` function doesn't support descriptions
* Options are generated from public/openqa-cli.yaml introduced in #6316
* no CI check enforces that committed scripts match the YAML. I am not sure how the public/openqa-cli.yaml is updated atm


openqa-cli-completion.bash
```diff
@@ -0,0 +1,37 @@
+#!/usr/bin/env bash
+
+_openqa_cli_completions() {
```
kalikiana, line 3: How is this file installed? I would guess something needs to be added to the build and packaging?

b10n1k: good point. we could add documentation. because what the user needs to do is to add it to the bashrc or zsh script and source it

perlpunk: We could also create sub packages, like I did it here for git-subrepo:
https://build.opensuse.org/projects/devel:tools:scm/packages/git-subrepo/files/git-subrepo.spec?expand=1
So we would have `openqa-client-{bash,zsh}-completion`

b10n1k: I dont know if I need to do anything in the spec file tho?! Can we have a chat tomorrow?

okurz: Give it a try. Look what https://build.opensuse.org/projects/devel:tools:scm/packages/git-subrepo/files/git-subrepo.spec?expand=1 is doing for "bash-completion" and "zsh-completion" respectively. Then extend https://github.com/os-autoinst/openQA/blob/master/dist/rpm/openQA.spec by doing the equivalent.

perlpunk: If it's in the spec, people only need to install one of the subpackages, rather than having to manually add something to their rc files

perlpunk (commented):
The only thing I don't like is the duplication.  We would have the parameters now in four places:
* https://github.com/os-autoinst/openQA/blob/master/lib/OpenQA/CLI/api.pm#L21
* https://github.com/os-autoinst/openQA/blob/master/lib/OpenQA/CLI/api.pm#L119
* And now in the two completion scripts

There is a solution for that, but it is a bigger change, not sure if we want that here.

I wrote a [program](https://github.com/perlpunk/App-AppSpec-p5) that can create pod and completion scripts from a YAML file.
The openqa-cli code could also read the YAML file, so we would only have the parameters in one place.
But it would be a new dependency, at least for developers.
(One cool thing is that it can generate completion with descriptions also for bash, with a trick)

I could at least post a proof of concept.

openqa-cli-completion.bash
baierjan: Look at something like for example alacritty: https://build.opensuse.org/projects/openSUSE:Factory/packages/alacritty/files/alacritty.spec

There is a subpackage for the completion which supplements the main package and bash (so it is automatically installed if you have the main program and bash).
```
%package bash-completion
Summary:        Bash Completion for %{name}
Requires:       bash-completion
Supplements:    (%{name} and bash)
BuildArch:      noarch
```

and the subpackage just needs to drop the completion file into `%{buildroot}/%{_datadir}/bash-completion/completions/%{name}`

perlpunk:
> I could at least post a proof of concept.

https://gist.github.com/perlpunk/b0d9fe14f3c877f28d8af3db6be28c70
Here is the demo with the bash completion including descriptions:
https://gist.github.com/perlpunk/b0d9fe14f3c877f28d8af3db6be28c70#file-demo-txt

b10n1k:
> The only thing I don't like is the duplication. We would have the parameters now in four places:
> 
> * https://github.com/os-autoinst/openQA/blob/master/lib/OpenQA/CLI/api.pm#L21
> * https://github.com/os-autoinst/openQA/blob/master/lib/OpenQA/CLI/api.pm#L119
> * And now in the two completion scripts
> 
> There is a solution for that, but it is a bigger change, not sure if we want that here.
> 
> I wrote a [program](https://github.com/perlpunk/App-AppSpec-p5) that can create pod and completion scripts from a YAML file. The openqa-cli code could also read the YAML file, so we would only have the parameters in one place. But it would be a new dependency, at least for developers. (One cool thing is that it can generate completion with descriptions also for bash, with a trick)
> 
> I could at least post a proof of concept.

awesome and thanks for taking the time for the proof of concept.
in general, this was one idea for the hackweek, and learning stuff for me. I know that ideally it should be automatic in some way, but it was easier for me to provide something within the hackweek, and thats why I keep it as simple as possible. Not sure I am going to do the automation, as I believe is on another level for me and my plan is to look at another small project I am interested in. I hope we are not gonna be too restrict with this PR for now. It is so cool to have it and I have also additional ideas for it. 
So manybe we could open some backlog tickets for the rest? wdyt?

Martchus (commented):
It is not nice that we not only introduce one more place where we specify arguments but even two. Ideally we have a single source of truth. That would of course complicate things. Maybe we can make it so there's at least just one additional place where we specify arguments.

The code that does the argument parsing definitely needs to state in a comment that one also has to change these files when adding new parameters or changing help texts.

okurz (changes requested):
Considering that @perlpunk already provided a proof of concept to reduce duplication with https://github.com/os-autoinst/openQA/pull/6063#issuecomment-2488349236 we should follow that approach

dist/rpm/openQA.spec
```diff
@@ -314,6 +314,13 @@ Use this package to install munin scripts that allow to monitor some openQA
 statistics.
 %endif
 
+%package openqa-cli-bash-completion.bash
+Summary:        Bash Completion for %{name}
+Requires:       bash-completion
+Supplements:    (%{name} and bash)
```
perlpunk, line 320: I'm not completely sure, but this should probably be:
```suggestion
%package bash-completion
Summary:        Bash Completion for %{name}
Requires:       bash-completion
Supplements:    (openQA-client and bash)
```

baierjan: `%{name}` is "openQA" and for subpackage, the name is automatically prepended so we will end up with openQA-bash-completion package which maybe not what we want so we might want to be even more crazy

```suggestion
%package client-bash-completion
Summary:        Bash Completion for %{name}-client
Requires:       bash-completion
Supplements:    (%{name}-client and bash)
```

dist/rpm/openQA.spec
```diff
@@ -314,6 +314,13 @@ Use this package to install munin scripts that allow to monitor some openQA
...
+
+%description openqa-cli-bash-completion
```
baierjan, line 322: and this also needs to be in-sync with the package definition above, i.e.
```suggestion
%description client-bash-completion
```

Note: Although the completion is for openqa-cli, I would go for the package name openQA-client-bash-completion as the openqa-cli is delivered by openQA-client package.

dist/rpm/openQA.spec
```diff
@@ -314,6 +314,22 @@ Use this package to install munin scripts that allow to monitor some openQA
...
+Requires:       bash-completion
+Supplements:    (%{name}-client and bash)
+
+%description client-bash-completion
+The official bash completion script for openqa-cli.
+
+
+%package client-zsh-completion
+Summary:        Zsh Completion for %{name}
+Requires:       bash-completion
```
perlpunk, line 328:
```suggestion
```

dist/rpm/openQA.spec
```diff
@@ -769,6 +786,9 @@ fi
 %{_bindir}/openqa-load-templates
 %{_bindir}/openqa-clone-custom-git-refspec
 %{_bindir}/openqa-validate-yaml
+%dir %{_datadir}/openqa/completion
```
perlpunk, line 789: That's not going to work. People would still have to manually add these files into their rc files.
I showed you how git-subrepo is doing it.
Just look at `rpm -ql git-subrepo-zsh-completion git-subrepo-bash-completion`.

What we need in the end is this (and with those exact filenames):
```
/usr/share/zsh/site-functions/_openqa-cli
/usr/share/bash-completion/completions/openqa-cli
```

completion/openqa-cli-completion.bash
```diff
@@ -0,0 +1,37 @@
+#!/usr/bin/env bash
```
perlpunk, line 1: I would still prefer a more generic directory name. We don't want to create a top level subfolder for every kind of feature we add.
I suggest either `share/completion` , or we use the existing `contrib` directory, so `contrib/completion`.
Other opinions?

dist/rpm/openQA.spec
perlpunk: Also you are listing those files as part of the client package (see `%files client` above), but it should go to `%files client-bash-completion` and `%files client-zsh-completion` respectively.

dist/rpm/openQA.spec
```diff
@@ -419,6 +434,8 @@ done
 #
 install -D -m 644 /dev/null %{buildroot}%{_localstatedir}/log/openqa
 install -m 0644 %{_sourcedir}/openQA.changes %{buildroot}%{_datadir}/openqa/public/Changelog
+install -Dm 0644 completion/openqa-cli-completion.bash %{buildroot}%{_datadir}/openqa/completion/openqa-cli-completion.bash
+install -Dm 0644 completion/openqa-cli-completion.zsh %{buildroot}%{_datadir}/openqa/completion/openqa-cli-completion.zsh
```
perlpunk, line 438: I think this should do it, haven't tested it yet:
```suggestion
install -Dm0644 %{_sourcedir}/contrib/completion/openqa-cli-completion.bash %{buildroot}%{_datadir}/bash-completion/completions/openqa-cli
install -Dm0644 %{_sourcedir}/contrib/completion/openqa-cli-completion.zsh %{buildroot}%{_datadir}/zsh/site-functions/_openqa-cli


# ...
# later

%files client-bash-completion
%{_datadir}/bash-completion/completions/openqa-cli

%files client-zsh-completion
%{_datadir}/zsh

```

completion/openqa-cli-completion.bash
baierjan: `contrib/completion` seems like a nice place

dist/rpm/openQA.spec
b10n1k: As always I noticed this a bit late. and once I realise that the path should be different. I try to make it work as you suggest. I will let you know

dist/rpm/openQA.spec
```diff
@@ -419,6 +434,8 @@ done
 #
 install -D -m 644 /dev/null %{buildroot}%{_localstatedir}/log/openqa
 install -m 0644 %{_sourcedir}/openQA.changes %{buildroot}%{_datadir}/openqa/public/Changelog
+install -Dm 0644 completion/openqa-cli-completion.bash %{buildroot}%{_datadir}/bash-completion/completions/openqa-cli
+install -Dm 0644 completion/openqa-cli-completion.zsh %{buildroot}%{_datadir}/usr/share/zsh/site-functions/_openqa-cli
```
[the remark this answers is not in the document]
b10n1k: sorry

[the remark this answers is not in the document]
perlpunk: Apart from missing a slash, now I think _sourcedir might not be the directory we want here.
Try it out by adding the slash.
Otherwise you can just move those statements up before the `cd` statement.

dist/rpm/openQA.spec
```diff
@@ -314,6 +314,24 @@ Use this package to install munin scripts that allow to monitor some openQA
...
+install -Dm 0644 completion/openqa-cli-completion.bash %{buildroot}%{_datadir}/bash-completion/completions/openqa-cli
+
+%package client-zsh-completion
+Summary:        Zsh Completion for %{name}
+Supplements:    (%{name}-client and zsh)
+
+%description client-zsh-completion
+The official zsh completion script for openqa-cli.
+
+install -Dm 0644 completion/openqa-cli-completion.zsh %{buildroot}%{_datadir}/usr/share/zsh/site-functions/_openqa-cli
```
perlpunk, line 334: Now that string is part of the description :)

It needs to be in the `%install` section.

b10n1k: I am working on that.

dist/rpm/openQA.spec
```diff
@@ -399,6 +423,8 @@ ln -s %{_datadir}/openqa/script/setup-db %{buildroot}%{_bindir}/openqa-setup-db
 ln -s %{_datadir}/openqa/script/openqa-label-all %{buildroot}%{_bindir}/openqa-label-all
 %endif
 
+# completion
+
```
okurz, line 427: what is this for?

docs/UsersGuide.asciidoc
```diff
@@ -963,6 +963,10 @@ without needing to install openQA itself. Call `openqa-cli --help` for help.
 The sub-commands provide further help, e.g. `openqa-cli api --help` contains
 a lot of examples.
 
+`openqa-cli` supports auto completion for `bash` and `zsh`. We distribute two
+packages `openQA-client-bash-completion` and `openQA-client-zsh-completion` as
+part of openQA-client.
+
```
okurz, line 969: That's only true for openSUSE/SLE

okurz (changes requested):

completion/openqa-cli-completion.zsh
```diff
@@ -0,0 +1,78 @@
...
+    local -a api_options
+    local -a archive_options
+    local -a monitor_options
+    local -a schedule_options
+    subcommands=("api" "archive" "monitor" "schedule")
+    main_options=(
+        '--apibase[Base URL for the API]'
+        '--apikey[API key for authentication]'
+        '--apisecret[API secret for authentication]'
+        '--help[Show help message]'
```
Wabri, line 18:
```suggestion
        '--help[Display help with available subcommands and options]'
```

completion/openqa-cli-completion.zsh
```diff
@@ -0,0 +1,78 @@
...
+        '--host[Target host, defaults to http://localhost]'
+        '--o3[Set target host to https://openqa.opensuse.org]'
+        '--osd[Set target host to http://openqa.suse.de]'
```
Wabri, line 21:
```suggestion
        '--osd[Set target host to https://openqa.suse.de]'
```

completion/openqa-cli-completion.zsh
```diff
@@ -0,0 +1,78 @@
...
+        '--asset-size-limit[Asset size limit in bytes]'
+        '--name[Name of this client, used by openQA to identify different clients via User-Agent header, defaults to "openqa-cli"]'
+        '--with-thumbnails[Download thumbnails as well]'
+    )
+    monitor_options=(
+        '--name[Name of this client, used by openQA to identify different clients via User-Agent header, defaults to "openqa-cli"]'
+        '--poll-interval[Specifies the poll interval in seconds]'
+    )
+    schedule_options=(
+        '--monitor[it until all jobs are done/cancelled and return non-zero exit code if at least on job has not passed/softfailed]'
```
Wabri, line 49: Is the help message imply Monitor? Like this:
```suggestion
        '--monitor[Monitor it until all jobs are done/cancelled and return non-zero exit code if at least on job has not passed/softfailed]'
```

completion/openqa-cli-completion.zsh
perlpunk: The options and its descriptions are copied from https://github.com/os-autoinst/openQA/blob/master/lib/OpenQA/CLI/api.pm#L115 , https://github.com/os-autoinst/openQA/blob/master/lib/OpenQA/CLI/schedule.pm#L55 etc.
Apparently for some the content was changed, or a mistake happened when copying & pasting.

We discussed already above that we want some kind of deduplication first, so that we don't have that content in two more places.

okurz:
I tried out the sub-packages. From within `podman run --rm -it registry.opensuse.org/opensuse/tumbleweed` I did

```
zypper ar https://download.opensuse.org/repositories/devel:/openQA:/GitHub:/os-autoinst:/openQA:/PR-6063/openSUSE_Tumbleweed/ pr
zypper in openQA-client-bash-completion openQA-client-zsh-completion
rpm -ql openQA-client-zsh-completion-4.6.1733298771.703167ed-7471.1.x86_64
rpm -ql openQA-client-bash-completion-4.6.1733298771.703167ed-7471.1.x86_64
```

and found
/usr/share/zsh/site-functions/_openqa-cli
/usr/share/zsh/site-functions/_openqa-cli/openqa-cli-completion.zsh
/usr/share/bash-completion/completions/openqa-cli
/usr/share/bash-completion/completions/openqa-cli/openqa-cli-completion.bash

perlpunk:
I wrote in https://github.com/os-autoinst/openQA/pull/6063#discussion_r1855273100:

> What we need in the end is this (and with those exact filenames):
```
/usr/share/zsh/site-functions/_openqa-cli
/usr/share/bash-completion/completions/openqa-cli
```
Those files need to be there, as files, not directories.

When you hit tab in bash, it looks into `/usr/share/bash-completion/completions` for a file with the same filename as the command. you can just have a look in that directory. You only see filenames there, right? No directories.
The same for zsh, just that the convention there is to prepend the file with an underscore.

Also I showed you the example for git-subrepo before. We looked at the output for `rpm -ql` together.
```
% rpm -ql git-subrepo-zsh-completion-0.4.6-16.4.noarch.rpm
/usr/share/zsh
/usr/share/zsh/site-functions
/usr/share/zsh/site-functions/_git-subrepo
% rpm -ql git-subrepo-bash-completion-0.4.6-16.4.noarch.rpm
/usr/share/bash-completion/completions/git-subrepo
```
Simply look at those files and think about what the files should look like for `openqa-cli`. Replace `git-subrepo` with `openqa-cli`.

completion/openqa-cli-completion.zsh
```diff
@@ -0,0 +1,78 @@
...
+                _arguments '*:option:->options' $main_options $api_options
+                ;;
+            archive)
+                _arguments '*:option:->options' $main_options $archive_options
+                ;;
+            monitor)
+                _arguments '*:option:->options' $main_options $archive_options
+            ;;
+            schedule)
+                _arguments '*:option:->options' $main_options $archive_options
```
perlpunk, line 68: they all three have `$archive_options`

completion/openqa-cli-completion.bash
perlpunk, line 1: Replace that with:
```suggestion
#compdef openqa-cli
```
so that automatic completion works

perlpunk: oops, sorry, that was meant for zsh only.
shouldn't harm in the bash file, but might be confusing :)

b10n1k: I removed the shebang from the bash. I checked other files and it seems it works either way. zsh works also with compdef.

b10n1k: done

docs/UsersGuide.asciidoc
```diff
@@ -963,6 +963,10 @@ without needing to install openQA itself. Call `openqa-cli --help` for help.
 The sub-commands provide further help, e.g. `openqa-cli api --help` contains
 a lot of examples.
 
+`openqa-cli` supports auto completion for `bash` and `zsh` with openSUSE/SLE.
+We distribute two packages `openQA-client-bash-completion` and
+`openQA-client-zsh-completion` as part of openQA-client.
```
okurz, line 968: Now you got it the wrong way around. With your contrib files there is auto-completion regardless of openSUSE/SLE but the packages are only available when using the spec file. How about installing within Makefile "install:" instead

b10n1k: Can you check again? 

Now `make install` should setup autocompletion according to the $SHELL

okurz (changes requested):
Latest changes look good but please keep in mind that the major issue is still https://github.com/os-autoinst/openQA/pull/6063#pullrequestreview-2451185603

Makefile
```diff
@@ -125,6 +125,17 @@ install-generic:
 	ln -s ../postgresql.service "$(DESTDIR)"/usr/lib/systemd/system/openqa-scheduler.service.requires/postgresql.service
 	install -d -m 755 "$(DESTDIR)"/usr/lib/systemd/system/openqa-websockets.service.requires
 	ln -s ../postgresql.service "$(DESTDIR)"/usr/lib/systemd/system/openqa-websockets.service.requires/postgresql.service
+	install -D -m 644 contrib/completions/openqa-cli-completion.bash "$(DESTDIR)"/usr/share//bash-completion/completions/openqa-cli
+	install -D -m 644 contrib/completions/openqa-cli-completion.zsh "$(DESTDIR)"/usr/share/zsh/site-functions/_openqa-cli
+	if [ "$(SHELL)" = "/bin/bash" ] || [ "$(SHELL)" = "/usr/bin/bash" ]; then \
```
perlpunk, line 130: I wouldn't add anything to a user's bashrc/zshrc by default when doing `make install`.
And if they have bash-completion installed, it shouldn't be necessary anyway. Same for zsh.

tools/generate-cli-completions
```diff
@@ -0,0 +1,160 @@
...
+use YAML::PP;
+
+my $spec_file = 'public/openqa-cli.yaml';
+my $bash_out = 'contrib/completions/openqa-cli-completion.bash';
+my $zsh_out = 'contrib/completions/openqa-cli-completion.zsh';
+
+my $spec = YAML::PP->new->load_file($spec_file);
+
+sub parse_option {
+    my ($entry) = @_;
```
Martchus: Use signatures in new code.

contrib/completions/openqa-cli-completion.bash
```diff
@@ -0,0 +1,36 @@
+# Generated by tools/generate-cli-completions — do not edit manually.
```
Martchus: I would avoid having these generated scripts in Git. One can simply invoke the Makefile target to build them if needed and the install target can also depend on that.

dist/rpm/openQA.spec
```diff
@@ -326,6 +326,20 @@ Use this package to install munin scripts that allow to monitor some openQA
 statistics.
 %endif
 
+%package client-bash-completion
+Summary:        Bash Completion for %{name}
+Requires:       bash-completion
+Supplements:    (%{name}-client and bash)
+
+%description client-bash-completion
+The official bash completion script for openqa-cli.
```
Martchus, line 340: I would avoid the "official" (as official vs. unofficial is probably not a useful distinction here). The same counts for Zsh.

```suggestion
Provides Bash completion for %{name}
```

dist/rpm/openQA.spec
```diff
@@ -422,6 +436,10 @@ install -m 644 contrib/munin/config/minion.config %{buildroot}/%{_sysconfdir}/mu
 install -m 755 contrib/munin/utils/munin-mail %{buildroot}/%{_datadir}/openqa/script/munin-mail
 %endif
 
+# completion
+install -Dm 0644 contrib/completions/openqa-cli-completion.bash %{buildroot}%{_datadir}/bash-completion/completions/openqa-cli
+install -Dm 0644 contrib/completions/openqa-cli-completion.zsh %{buildroot}%{_datadir}/zsh/site-functions/_openqa-cli
```
Martchus: Use the install target instead of repeating that here.

dist/rpm/openQA.spec
```diff
@@ -696,6 +714,11 @@ fi
 %files devel
 %endif
 
+%files client-bash-completion
+%{_datadir}/bash-completion/completions/openqa-cli
+%files client-zsh-completion
```
Martchus, line 721:
```suggestion

%files client-zsh-completion
```

Martchus (changes requested):

Martchus:
It would be good to avoid introducing the generated files at all (and *not* having a commit adding these files and removing them in the same PR).

tools/generate-cli-completions
perlpunk, line 1: Why a script when we can use my framework for that?
I mentioned this already here: https://github.com/os-autoinst/openQA/pull/6063#pullrequestreview-2448096604
* https://github.com/perlpunk/App-AppSpec-p5
This can create pod and completion automatically from the yaml file.
No need to invent something new.
We would only need to add this to Factory and Leap.

Have you tried that out at all?

perlpunk (dismissed):

tools/generate-cli-completions
d3flex: yes I did - I test it only in zsh tho. 
I missed the https://github.com/perlpunk/App-AppSpec-p5. I had completely forgot about that. but either way, the current PR works and no need for another dependency. I would like to have the completion and I am good to get it refactor if you later want to integrate your script.

perlpunk: I submitted two modules to Factory:
* https://build.opensuse.org/requests/1350776 perl-App-AppSpec
* https://build.opensuse.org/requests/1350775 perl-App-Spec

d3flex: Eitherway if you do not see something wrong, I think you still can accept my PR. it will be a single change after that

tools/generate-cli-completions
```diff
@@ -0,0 +1,157 @@
...
+
+print $zsh_fh <<'FOOTER';
+            *)
+                _values 'subcommand' api archive monitor schedule
+                ;;
+        esac
+    fi
+}
+
+compdef _openqa_cli_completions openqa-cli
```
perlpunk: This line can and should be left out when putting it into a standard zsh completion location.
This is because of performance.
It avoids compiling all zsh completion scripts at shell startup, but it will only be compiled the first time you use a program (that's the reason why you have to hit TAB twice in a shell when you want completion for a command that you have not yet used in this session).
If you add this line, it will compile it at shell startup.
So this line is something the user can do if they want to try out a completion script that is not (yet) installed in a standard location.

kalikiana: Ack. Please remove this part

tools/generate-cli-completions
kalikiana: I appreciate you're updating the PR, but to be honest after 2 years, maybe we can wait 2 weeks for packaging?

tools/generate-cli-completions
d3flex: that's super feedback IMO🙏. manual tested as well

dist/rpm/openQA.spec
perlpunk: There is still no empty line here, as @Martchus suggested. Why is it resolved?

tools/generate-cli-completions
kalikiana: Just to clarify, we discussed this and this is not blocking the PR

tools/generate-cli-completions
```diff
@@ -149,8 +149,6 @@ print $zsh_fh <<'FOOTER';
         esac
     fi
 }
-
-compdef _openqa_cli_completions openqa-cli
```
Martchus: The documentation commit shouldn't contain this removal. It shouldn't be part of the 1st commit in the first place.

Martchus (commented):
I just have a small nitpck. Otherwise it looks good to merge.

tools/generate-cli-completions
```diff
@@ -0,0 +1,155 @@
...
+        push @result, [parse_option($entry)];
+    }
+    return @result;
+}
+
+my @global_opts = collect_options($spec->{options});
+my %subcommands = %{$spec->{subcommands} // {}};
+my @subcmds = sort keys %subcommands;
+
+# bash
```
okurz, line 46: ?

tools/generate-cli-completions
```diff
@@ -0,0 +1,155 @@
+#!/usr/bin/env perl
```
okurz, line 1: Please add copyright+license

tools/generate-cli-completions
```diff
@@ -0,0 +1,155 @@
...
+    return ($name, $desc);
+}
+
+sub collect_options ($option_list) {
+    my @result;
+    for my $entry (@{$option_list // []}) {
+        push @result, [parse_option($entry)];
+    }
+    return @result;
+}
```
okurz, line 40: Use `map`

tools/generate-cli-completions
```diff
@@ -0,0 +1,155 @@
...
+my @global_opts = collect_options($spec->{options});
+my %subcommands = %{$spec->{subcommands} // {}};
+my @subcmds = sort keys %subcommands;
+
+# bash
+
+sub bash_option_list (@opts) {
+    return join ' ', map { "--$_->[0]" } @opts;
+}
+
```
okurz, line 51:
```suggestion
sub bash_option_list (@opts) { join ' ', map { "--$_->[0]" } @opts }

```

tools/generate-cli-completions
```diff
@@ -0,0 +1,155 @@
...
+open my $bash_fh, '>', $bash_out or die "Cannot write $bash_out: $!";
+print $bash_fh <<'HEADER';
+# Generated by tools/generate-cli-completions — do not edit manually.
+# Regenerate with: make generate-completions
+_openqa_cli_completions() {
+    local cur subcommands main_options api_options archive_options monitor_options schedule_options
+    cur="${COMP_WORDS[COMP_CWORD]}"
+HEADER
```
okurz:
```suggestion
use Mojo::File qw(path);
…
path($bash_out)->spew(<<'HEADER');
# Generated by tools/generate-cli-completions — do not edit manually.
# Regenerate with: make generate-completions
_openqa_cli_completions() {
    local cur subcommands main_options api_options archive_options monitor_options schedule_options
    cur="${COMP_WORDS[COMP_CWORD]}"
HEADER
```

and correspondingly further on

tools/generate-cli-completions
okurz, line 1: You managed to create an elaborate custom generation script for what looks like Perl does not otherwise offer. In comparison in Python using the "typer" library brings bash completions out of the box. This makes me wonder if it wouldn't be better for us to migrate to that.

okurz (changes requested):

tools/generate-cli-completions
perlpunk: > You managed to create an elaborate custom generation script for what looks like Perl does not otherwise offer.

There is a perl module for it, like I mentioned. See the earlier discussion.
https://github.com/os-autoinst/openQA/pull/6063#discussion_r3182997064
It's in Factory now, and I will submit it to Leap 16.
Should I also submit it to 15.6, considering we only need it for the generation at build time?

perlpunk: btw, the App::Spec module does not only generate completion, it's a complete commandline framework we could move to. We already have the yaml file with the options that the module can consume.
Though I would have to check if there are any features we use from `Mojolicious::Commands` that mine does not offer.

perlpunk: btw, the completion generation with appspec would be as simple as that:
```
appspec completion --bash public/openqa-cli.yaml >openqa-cli.bash
appspec completion --zsh public/openqa-cli.yaml >_openqa-cli
```

perlpunk: Submit to Leap 16: https://code.opensuse.org/leap/features/issue/302

okurz: > Should I also submit it to 15.6, considering we only need it for the generation at build time?

no, Leap 15.6 is EOL

tools/generate-cli-completions
d3flex: separation of  the bash/zsh generation parts. gonna leave it

tools/generate-cli-completions
d3flex: done and revert it by tidy. ignore

okurz (approved):

dist/rpm/openQA.spec
Martchus: This still hasn't been fixed.

Martchus (approved):

PR #6070, opened 2024-11-26
Add endpoint for listing all WebAPI routes

perlpunk:
```
% openqa-cli api routes | jq
{                 
  "routes": [                     
    {             
      "methods": [
        "GET"                                     
      ],                                 
      "path": "/"
    },                                              
...
    {
      "methods": [
        "DELETE",                 
        "GET"     
      ],          
      "path": "/api/v1/assets/<id:num>"           
    },                                   
```
I find it cumbersome to find out routes currently. The webpage 404 page, e.g. https://openqa.opensuse.org/foo , is very long and I lose the overview under which top level route I am.

This could also be useful later for creating dynamic completion, e.g. `openqa-cli api j<TAB>`

### DIFF
--- a/lib/OpenQA/WebAPI/Controller/API/V1/Routes.pm
+++ b/lib/OpenQA/WebAPI/Controller/API/V1/Routes.pm
@@ -1,0 +1,54 @@
+# Copyright SUSE LLC
+# SPDX-License-Identifier: GPL-2.0-or-later
+
+package OpenQA::WebAPI::Controller::API::V1::Routes;
+use Mojo::Base 'Mojolicious::Controller', -signatures;
+
+=pod
+
+=head1 NAME
+
+OpenQA::WebAPI::Controller::API::V1::Routes
+
+=head1 SYNOPSIS
+
+  use OpenQA::WebAPI::Controller::API::V1::Routes;
+
+=head1 DESCRIPTION
+
+
+=head1 METHODS
+
+=over 4
+
+=item list()
+
+Lists all WebAPI routes
+
+=back
+
+=cut
+
+sub list ($self) {
+    my $routes = $self->app->routes->children;
+    my %hash;
+    _walk($_, \%hash, '') for @$routes;
+    my @list;
+    for my $route (sort keys %hash) {
+        my $methods = $hash{$route};
+        push @list, {path => $route, methods => [sort keys %$methods]};
+    }
+    $self->render(json => {routes => \@list});
+}
+
+sub _walk ($route, $hash, $path) {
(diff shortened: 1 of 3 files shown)


Martchus (approved):
Good idea and it might indeed be interesting for dynamic completion.

perlpunk:
No codecov result. I'll force push. I would expect that the module appears as uncovered, as I didn't add any tests so far

Martchus:
Still no results. Recent webhook deliveries look good but there's probably some problem on the Codecov side right now.

okurz:
… and now there is the report :)

lib/OpenQA/WebAPI/Controller/API/V1/Routes.pm
```diff
@@ -0,0 +1,60 @@
...
+        @methods = @{($route->methods) || []};
+    }
+    else {
+        @methods = @{($route->via) || []};
+    }
+    my $endpoint = "$path$pattern" || '/';
+    $hash->{$endpoint}->{$_} = 1 for @methods;
+    for my $child (@{$route->children || []}) {
+        _walk($child, $hash, "$path$pattern");
+    }
```
kalikiana, line 57: I guess at least a small unit test would be good. Something like `$t->get_ok('/api/v1/routes')->json_has('XXX')->->json_is('AAA' => BBB);` to check some known API routes as included (c.f. `t/api/04-jobs.t`)

perlpunk: yeah, the difficult part so far was to decide in which existing test file this should go :)

lib/OpenQA/WebAPI/Controller/API/V1/Routes.pm
```diff
@@ -0,0 +1,60 @@
...
+        push @list, {path => $route, methods => [sort keys %$methods]};
+    }
+    $self->render(json => {routes => \@list});
+}
+
+sub _walk ($route, $hash, $path) {
+    my $pattern = $route->pattern->unparsed || '';
+    my @methods;
+    if ($route->can('methods')) {
+        @methods = @{($route->methods) || []};
```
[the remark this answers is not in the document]
perlpunk: I don't understand. An empty list is empty.
`methods` might return undef, in which case `@{ somehing undef }`  will error, so  `@{ somehing undef || []}` is needed.
btw, I copied this code from https://github.com/os-autoinst/openQA/blob/master/templates/webapi/not_found.html.ep#L26

perlpunk:
I added a test and simplified the _walk method a bit

PR #6073, opened 2024-12-02
Make 'Broken' workers show as 'Unavailable' instead

r-richardson:
This PR displays "Broken" workers as instead "Unavailable", as the previous wording was e.g. not including workers which where unavailable due to 'too high load'.
It also replaces the 'Error' title of the worker states hover popup with 'Details' and enhances the 'too high load' output.

Example:
Screenshot From 2024-12-02 21-17-21

Related Ticket: https://progress.opensuse.org/issues/164084

### DIFF
--- a/lib/OpenQA/Worker.pm
+++ b/lib/OpenQA/Worker.pm
@@ -767,7 +767,8 @@ sub _check_system_utilization (
     return undef unless $threshold && @$load >= 3;
     # look at the load evolution over time to react quick enough if the load
     # rises but accept a falling edge
-    return "The average load (@$load) is exceeding the configured threshold of $threshold."
+    return
+"The average load (@$load) is exceeding the configured threshold of $threshold. The worker will temporarily not accept new jobs until the load is lower again."
       if max(@$load) > $threshold && ($load->[0] > $load->[1] || $load->[0] > $load->[2] || min(@$load) > $threshold);
     return undef;
 }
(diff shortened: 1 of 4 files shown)


lib/OpenQA/Worker.pm
```diff
@@ -767,7 +767,8 @@ sub _check_system_utilization (
     return undef unless $threshold && @$load >= 3;
     # look at the load evolution over time to react quick enough if the load
     # rises but accept a falling edge
-    return "The average load (@$load) is exceeding the configured threshold of $threshold."
+    return
+"The average load (@$load) is exceeding the configured threshold of $threshold. The worker will temporarily not accept new jobs until the load is lower again."
```
b10n1k, line 771: one comment quite off the PR. this seems to be displayed as a info next to the status. I wonder why we do not display the load as another key:value field in the page.

b10n1k: not sure. Maybe check with others too. for me makes sense but maybe there is a reason

perlpunk: The load is not checked every time the worker details page is shown. The worker sends this info via websocket if the load is too high, and it gets saved into the worker table in the database.

perlpunk (commented):
The "Broken" in the dropdown in the worker list also needs to be changed.

perlpunk (approved):

PR #6074, opened 2024-12-02
Update SUSE Jira url

mpagot:
Update url from .de to .com
Update unit testing accordingly.


lib/OpenQA/Utils.pm
```diff
@@ -49,7 +49,7 @@ BEGIN {
         gh => 'https://github.com/',
         kde => 'https://bugs.kde.org/show_bug.cgi?id=',
         fdo => 'https://bugs.freedesktop.org/show_bug.cgi?id=',
-        jsc => 'https://jira.suse.de/browse/',
+        jsc => 'https://jira.suse.com/browse/',
```
Martchus, line 52: Maybe the `%BUGURLS` hash below for the reverse direction should be extended (like it was done for bsc) so old URLs are still resolved.

okurz (approved):

PR #6076, opened 2024-12-04
Add config option to restrict asset downloads to logged-in users

Martchus:
This PR is now only covering the web app and probably not useful on its own. You can still see how the approach using basic auth/htpasswd would look like on https://github.com/Martchus/openQA/tree/asset-auth-nginx-basic-auth.

---

See https://progress.opensuse.org/issues/170380

### DIFF
--- a/etc/openqa/openqa.ini
+++ b/etc/openqa/openqa.ini
@@ -125,6 +125,8 @@
 ## Authentication method to use for user management
 [auth]
 #method = Fake|OpenID|OAuth2
+## whether authentication is required to access assets; does NOT affect assets made available directly via e.g. Apache or NGINX
+#require_for_assets = 0
 
 ## for GitHub, salsa.debian.org and providers listed on https://metacpan.org/pod/Mojolicious::Plugin::OAuth2#Configuration
 ## one can use:
--- a/lib/OpenQA/Setup.pm
+++ b/lib/OpenQA/Setup.pm
@@ -61,6 +61,7 @@ sub read_config ($app) {
         },
         auth => {
             method => 'OpenID',
+            require_for_assets => 0,
         },
         'scm git' => {
             update_remote => '',
--- a/lib/OpenQA/WebAPI.pm
+++ b/lib/OpenQA/WebAPI.pm
@@ -141,11 +141,15 @@ sub startup ($self) {
     $r->get('/tests/list_scheduled_ajax')->name('tests_ajax')->to('test#list_scheduled_ajax');
 
     # only provide a URL helper - this is overtaken by apache
-    $r->get('/assets/*assetpath')->name('download_asset')->to('file#download_asset');
+    my $config = $app->config;
+    my $require_auth_for_assets = $config->{auth}->{require_for_assets};
+    my $assets_r = $require_auth_for_assets ? $logged_in : $r;
+    $assets_r->get('/assets/*assetpath')->name('download_asset')->to('file#download_asset');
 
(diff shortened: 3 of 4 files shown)


okurz (commented):
Seems like you try to incorporate changes applicable to pure mojo+nginx+apache at the same time. Wouldn't it be easier if we focus for a start on mojo especially to avoid the tricky authentication synchronisation between htpasswd and the database?

Martchus:
Maybe we can use https://docs.nginx.com/nginx/admin-guide/security-controls/configuring-subrequest-authentication instead of maintaining an htpasswd file.

Martchus:
It makes not much sense to focus on the web app as a start because this change would be useless if one can just access assets via NGINX anyway. However, I now split the commits because the changes to the web app are of course independent and I probably want to try out https://docs.nginx.com/nginx/admin-guide/security-controls/configuring-subrequest-authentication for NGINX.

So this PR is now only covering the web app. You can still see how the approach using basic auth/htpasswd would look like on https://github.com/Martchus/openQA/tree/asset-auth-nginx-basic-auth.

Martchus:
Using a subquery for authentication would actually be quite easy: https://github.com/Martchus/openQA/tree/asset-auth-nginx-subrequest

With this we still need to adapt other code, e.g. so assets are still accessible via a logged-in web UI session. With Chromium this just works but with Firefox it doesn't and I cannot get it to work using tokenauth (e.g. via https://github.com/Martchus/openQA/tree/asset-auth-nginx-subrequest-with-redirect).

foursixnine:
@okurz @Martchus we have much longer time to work on this (6 months) we can take a step back now and rethink a bit, what is the actual problem we're trying to solve.

I believe I have provided you with all of the tools and information for you to do your work efficiently, when in doubt, I work under the priorities set by qe-core product owner.

Martchus:
It looks like the cache service continues to work without modification:

```
notes:
  has_download_error: false
  lock: openSUSE-Tumbleweed-DVD-x86_64-Snapshot20241203-Media.iso.http://localhost
  output: |
    [info] [#1120] Cache size of "/hdd/openqa-devel/openqa/worker/cache" is 46 GiB, with limit 50 GiB
    [info] [#1120] Downloading "openSUSE-Tumbleweed-DVD-x86_64-Snapshot20241203-Media.iso" from "http://localhost/tests/4401/asset/iso/openSUSE-Tumbleweed-DVD-x86_64-Snapshot20241203-Media.iso"
    [info] [#1120] Size of "/hdd/openqa-devel/openqa/worker/cache/localhost/openSUSE-Tumbleweed-DVD-x86_64-Snapshot20241203-Media.iso" is 33 KiB, with ETag ""
    [info] [#1120] Download of "/hdd/openqa-devel/openqa/worker/cache/localhost/openSUSE-Tumbleweed-DVD-x86_64-Snapshot20241203-Media.iso" successful (169 KiB/s), new cache size is 46 GiB
```

This works despite C&P-ing `http://localhost/tests/4401/asset/iso/openSUSE-Tumbleweed-DVD-x86_64-Snapshot20241203-Media.iso` into Firefox gives a 403 response.

Unfortunately Firefox also gives a 403 response with a logged-in session under `http://localhost` which also affects asset downloads from the web UI as mentioned before. Now it also doesn't work with Chromium. Probably it still had basic auth credentials cached in my test on Thursday.

Martchus:
I summarized my findings so far in https://progress.opensuse.org/issues/174154.

PR #6082, opened 2024-12-12
Add config option to restrict asset downloads to logged-in users

Martchus:
* Disallow unauthenticated access to assets in web application routes
* Does NOT cover access served via Apache/NGINX
* Does NOT cover adjusting the cache service and additional tooling like the `openqa-clone-job` script
* See https://progress.opensuse.org/issues/170380

### DIFF
--- a/etc/openqa/openqa.ini
+++ b/etc/openqa/openqa.ini
@@ -125,6 +125,10 @@
 ## Authentication method to use for user management
 [auth]
 #method = Fake|OpenID|OAuth2
+## whether authentication is required to access assets; caveats:
+## - Does NOT affect assets made available outside the scope of the openQA service (e.g. via Apache, NGINX or NFS).
+## - Might not work well with other features like openqa-clone-job and the cache service.
+#require_for_assets = 0
 
 ## for GitHub, salsa.debian.org and providers listed on https://metacpan.org/pod/Mojolicious::Plugin::OAuth2#Configuration
 ## one can use:
--- a/lib/OpenQA/Setup.pm
+++ b/lib/OpenQA/Setup.pm
@@ -61,6 +61,7 @@ sub read_config ($app) {
         },
         auth => {
             method => 'OpenID',
+            require_for_assets => 0,
         },
         'scm git' => {
             update_remote => '',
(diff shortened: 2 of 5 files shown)


etc/openqa/openqa.ini
```diff
@@ -125,6 +125,8 @@
 ## Authentication method to use for user management
 [auth]
 #method = Fake|OpenID|OAuth2
+## whether authentication is required to access assets; does NOT affect assets made available directly via e.g. Apache or NGINX
```
okurz, line 128:
```suggestion
## whether authentication is required to access assets; does NOT affect assets made available directly via a web proxy, e.g. Apache/NGINX
```

okurz (changes requested):
Looks pretty clean. Can openQA jobs on workers still access assets during job runs, e.g. to read a iso or qcow file needed for test execution?

Martchus:
> Can openQA jobs on workers still access assets during job runs, e.g. to read a iso or qcow file needed for test execution?

No, they cannot (except via NFS which is probably completely out of scope). The cache service is one of many open points I added on https://progress.opensuse.org/issues/174154. With https://github.com/Martchus/openQA/commit/32e1b30434f09ff71d242d5f8f51b68e0090d05a (which is referenced in the progress issue) this would likely work.

okurz:
> > Can openQA jobs on workers still access assets during job runs, e.g. to read a iso or qcow file needed for test execution?
> 
> No, they cannot (except via NFS which is probably completely out of scope). The cache service is one of many open points I added on https://progress.opensuse.org/issues/174154. With [Martchus@32e1b30](https://github.com/Martchus/openQA/commit/32e1b30434f09ff71d242d5f8f51b68e0090d05a) (which is referenced in the progress issue) this would likely work.

Well, worker instances having direct access to the openQA base dir still work which includes cases like worker instances running on the same host as the webUI – which can even be considered the default for tiny instances – as well as when having access over NFS. How about you state that in the explanatory comment in openqa.ini. Then I guess we can merge this limited-scope feature

perlpunk (approved):

PR #6090, opened 2024-12-20
Prevent SIGCHLD handler override within tests

baierjan:
This should prevent _setup_sigchld_handler overriding the handler for IPC::Run3 inside AssetPack. Without this, the sass process is not handled correctly inside the tests and the assets are not correctly compiled by the AssetPack.

Reference: https://progress.opensuse.org/issues/174607


perlpunk (approved):

Martchus (approved):

perlpunk:
codecov report missing, merging manually

PR #6101, opened 2025-01-09
Allow monitor command to skip aborted jobs

kalikiana:
Aborted includes cancelled and parallel failed.

See: https://progress.opensuse.org/issues/174583

### DIFF
--- a/lib/OpenQA/CLI/monitor.pm
+++ b/lib/OpenQA/CLI/monitor.pm
@@ -23,6 +23,11 @@ sub _monitor_jobs ($self, $client, $poll_interval, $job_ids, $job_results) {
         if (OpenQA::Jobs::Constants::meta_state($job_state) eq OpenQA::Jobs::Constants::FINAL) {
             push @$job_results, $job->{result} // NONE;
             next;
+        } elsif (OpenQA::Jobs::Constants::meta_state($job_state) eq OpenQA::Jobs::Constants::ABORTED_RESULTS) {
+            print encode('UTF-8',
+                "Job $job_id was aborted but did not fail ($job_state), hence skipping)\n");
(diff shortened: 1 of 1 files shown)


lib/OpenQA/CLI/monitor.pm
```diff
@@ -23,6 +23,11 @@ sub _monitor_jobs ($self, $client, $poll_interval, $job_ids, $job_results) {
         if (OpenQA::Jobs::Constants::meta_state($job_state) eq OpenQA::Jobs::Constants::FINAL) {
             push @$job_results, $job->{result} // NONE;
             next;
+        } elsif (OpenQA::Jobs::Constants::meta_state($job_state) eq OpenQA::Jobs::Constants::ABORTED_RESULTS) {
```
perlpunk, line 26: It's still a draft, but I would mention that this should probably be configurable. An aborted result usually not considered as successful.

perlpunk: Also, `obsoleted` is a result and not a state. Any job will sooner or later be in a `FINAL` state.

Martchus: This condition doesn't make much sense. `meta_state` returns meta-state for the specified concrete state but this check compares it to a meta-*result* which is a different set of constants.

But yes, as @perlpunk has already mentioned, this should be configurable anyway.

So this should probably ignore the *results* `skipped` and `obsoleted` and maybe for the sake of completeness the the meta-*result* `ABORTED_RESULTS`. Not sure how much sense it makes to deal with the meta constants here when making this configurable anyway.

perlpunk: And the check should be done in `_compute_return_code`, not here

PR #6104, opened 2025-01-10
Allow cleaning up openQA devel packages from the zypper cache directory

Martchus:
* Add a script `openqa-clean-repo-cache` that allows to clean RPM cache repositories keeping a certain number of packages per version; this is similar to the `paccache` script from Arch Linux
* Allow running this script as part of `openqa-auto-update`
* See https://progress.opensuse.org/issues/174313

---

There are no automatic tests but I tested this locally a lot. The interesting part is just 80 lines of code anyway (and that is *with* the help text and argument parsing boiler plate).

### DIFF
--- a/dist/rpm/openQA.spec
+++ b/dist/rpm/openQA.spec
@@ -694,6 +694,7 @@ fi
 %{_localstatedir}/lib/openqa/script
 %{_localstatedir}/lib/openqa/tests
 %{_datadir}/openqa/script/openqa-check-devel-repo
+%{_datadir}/openqa/script/openqa-clean-repo-cache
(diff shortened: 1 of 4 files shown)


perlpunk:
> I'm wondering whether introducing shfmt was a good idea but for now I'd keep it despite its limitations.

I'm fine with it because its advantages overweigh this IMHO.

We could use [my bash app generator](https://metacpan.org/dist/App-Spec-Bash/view/lib/appspec-bash.pod) to hide all the argument parsing away and get rid of the manual usage print ;-)
But probably overkill for such a simple script...

perlpunk (approved):

PR #6107, opened 2025-01-13
Apply t/full-stack.t as unstable to mitigate the failing test

Wabri:
https://progress.opensuse.org/issues/175060


Martchus (approved):
This probably won't help because it just handles the lack of stability with retries which is problematic here. However, I'm also not against this change as it also doesn't make things worse. Let's consider it a short-term mitigation that *might* help (but probably doesn't).

okurz:
I don't think this will work as t/full-stack.t has its extra make target

PR #6110, opened 2025-01-13
Avoid calling `is_running` unnecessarily in `kill`

Martchus:
The `stop` function does the same check internally so it isn't required.

Related ticket: https://progress.opensuse.org/issues/170209


Martchus:
Note that this doesn't fix the related issue. For this we need https://github.com/openSUSE/Mojo-IOLoop-ReadWriteProcess/pull/61.

perlpunk (approved):

PR #6111, opened 2025-01-14
t: Fix hidden output of 'diag explain'

okurz:
The common way to provide additional detailed output about internal data
structures in case of test failures is 'diag explain' from Test::More.
However in Test::Most that behaves different and causes unexpected
issues like a bare '# 0' instead of array content.
https://metacpan.org/pod/Test::Most#explain explains that difference in
behaviour.

This commit changes all according occurences to 'always_explain'.

I used the following command:

```
git grep -l 'diag explain' | xargs sed -i 's/diag explain/always_explain/g'
```


t/full-stack.t
```diff
@@ -118,13 +118,13 @@ sub check_scheduled_job_and_wait_for_free_worker ($worker_class) {
...
     Test::More::fail "no worker with class $worker_class showed up after $elapsed seconds";
-    Test::More::diag explain 'free workers: ', [map { $_->info } @$free_workers];
+    Test::More::always_explain 'free workers: ', [map { $_->info } @$free_workers];
     return ($relevant_jobs, $elapsed);
 }
 
 sub show_job_info ($job_id) {
     my $job = $schema->resultset('Jobs')->find($job_id);
-    Test::More::diag explain 'job info: ', $job ? $job->to_hash : undef;
+    Test::More::always_explain 'job info: ', $job ? $job->to_hash : undef;
```
Wabri, line 127:
```suggestion
    Test::More::diag always_explain 'job info: ', $job ? $job->to_hash : undef;
```

t/full-stack.t
```diff
@@ -118,13 +118,13 @@ sub check_scheduled_job_and_wait_for_free_worker ($worker_class) {
         }
     }
     Test::More::fail "no worker with class $worker_class showed up after $elapsed seconds";
-    Test::More::diag explain 'free workers: ', [map { $_->info } @$free_workers];
+    Test::More::always_explain 'free workers: ', [map { $_->info } @$free_workers];
```
Wabri, line 121:
```suggestion
    Test::More::diag always_explain 'free workers: ', [map { $_->info } @$free_workers];
```

Wabri (changes requested):
By looking at https://app.circleci.com/pipelines/github/os-autoinst/openQA/15474/workflows/f3c78440-a144-4fee-bc4b-50077e30fe86/jobs/148306 it seems that something not working. Can it be the diag missing for always_explain?

Martchus (approved):

t/full-stack.t
okurz: actually I removed the module prefix is `always_explain` is unique now

okurz:
@mergifyio rebase

Wabri (approved):

PR #6114, opened 2025-01-14
Add restricting asset downloads via NGINX using subrequest auth

Martchus:
See the individual commit messages and https://progress.opensuse.org/issues/174301 for details.

I found out the problem I previously struggled with (see problematic points on https://progress.opensuse.org/issues/174154). With that out of the way there was actually not that much I had to change. With this PR all ATs in https://progress.opensuse.org/issues/174301 work. (I also slightly extended them along the way to cover both relevant routes.)

I haven't tested the cache service yet.

### DIFF
--- a/lib/OpenQA/Shared/Controller/Auth.pm
+++ b/lib/OpenQA/Shared/Controller/Auth.pm
@@ -49,7 +49,7 @@ sub auth ($self) {
     # Browser with a logged in user
     my ($user, $reason) = (undef, 'Not authorized');
     if ($user = $self->current_user) {
-        ($user, $reason) = (undef, 'Bad CSRF token!') unless $self->valid_csrf;
+        ($user, $reason) = (undef, 'Bad CSRF token!') unless $self->req->method eq 'GET' || $self->valid_csrf;
     }
 
     # No session (probably not a browser)
--- a/lib/OpenQA/WebAPI.pm
+++ b/lib/OpenQA/WebAPI.pm
@@ -53,6 +53,7 @@ sub startup ($self) {
 
     # register basic routes
     my $logged_in = $r->under('/')->to('session#ensure_user');
+    my $auth_any_user = $r->under('/')->to('Auth#auth');
     my $auth = $r->under('/')->to('session#ensure_operator');
 
     # Routes used by plugins (UI and API)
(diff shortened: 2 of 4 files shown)


okurz (approved):

lib/OpenQA/Shared/Controller/Auth.pm
```diff
@@ -49,7 +49,7 @@ sub auth ($self) {
     # Browser with a logged in user
     my ($user, $reason) = (undef, 'Not authorized');
     if ($user = $self->current_user) {
-        ($user, $reason) = (undef, 'Bad CSRF token!') unless $self->valid_csrf;
+        ($user, $reason) = (undef, 'Bad CSRF token!') unless $self->req->method eq 'GET' || $self->valid_csrf;
```
b10n1k, line 52: Why do you need this change? there is a comment in the valid_csrf which made me think that if this restriction has to be explicit it has to be better in the valid_csrf, no?

lib/OpenQA/WebAPI.pm
```diff
@@ -53,6 +53,7 @@ sub startup ($self) {
 
     # register basic routes
     my $logged_in = $r->under('/')->to('session#ensure_user');
+    my $auth_any_user = $r->under('/')->to('Auth#auth');
```
b10n1k, line 56: I notice that we have now two equal routes. one here `$aut_any_user` and one in L68 `$api_auth_any_user`. I dont know for sure but I have a feeling that they can be ust one. wdys?

lib/OpenQA/Shared/Controller/Auth.pm
Martchus: Your first question is answered in the last paragraph of the commit message of the first commit.

I don't understand the second question completely. I can only say that from a security standpoint is is of course generally better to be restrictive. However, I suppose here we have to make a change to be less restrictive. The reasoning why I think this is fine is also explained in the commit message.

lib/OpenQA/WebAPI.pm
Martchus: The other route is under `$r->any('/api/v1')`. So no, they are no equivalent. It would maybe be possible to reduce redundancy. However, I don't think it is worthwhile to change how the routing is structured for the sake of this.

perlpunk (approved):

PR #6118, opened 2025-01-15
perltidy: Prevent excessive line-wrapping

okurz:
Also see discussion in https://github.com/os-autoinst/openQA/pull/6116#issuecomment-2592760425


Martchus:
It would make sense to base this on the tidy update itself to see how it behaves with the new tidy version. And you of course have to run tidy :-)

okurz:
> It would make sense to base this on the tidy update itself to see how it behaves with the new tidy version. And you of course have to run tidy :-)

Done. Looks better

perlpunk:
See #6119

PR #6120, opened 2025-01-15
Make cache service work with authenticated asset downloads

Martchus:
This PR incorporates https://github.com/os-autoinst/openQA/pull/6114 which should probably be merged first.

It will tweak the cache service and the openQA user agent in general so both work when authenticated asset downloads are enabled. See https://progress.opensuse.org/issues/170380 and individual commit messages for details.

If wanted I can also extend tests. (Maybe I'll have to adapt tests anyway.) I tested this locally. Before this change it didn't work and with it it works when connecting directly and when using NGINX.


okurz (approved):

Martchus:
I pushed one more commit because I forgot to include these changes into the commit yesterday.

okurz (approved):

Wabri (approved):

PR #6126, opened 2025-01-17
Add jobs API option "follow" to return latest clone of queried job

asdil12:
Ticket: https://progress.opensuse.org/issues/175305

### DIFF
--- a/lib/OpenQA/WebAPI/Controller/API/V1/Job.pm
+++ b/lib/OpenQA/WebAPI/Controller/API/V1/Job.pm
@@ -444,6 +444,7 @@ sub create ($self) {
 Shows details for a specific job, such as the assets associated, assigned worker id,
 children and parents, job id, group id, name, parent group id and name, priority, result,
 settings, state and times of startup and finish of the job.
+Pass follow=1 as query param to follow job clones and report most recent result for given id.
 
 =back
 
@@ -453,9 +454,11 @@ sub show ($self) {
     my $job_id = int($self->stash('jobid'));
     my $details = $self->stash('details') || 0;
     my $check_assets = !!$self->param('check_assets');
-    my $job = $self->schema->resultset('Jobs')->find($job_id, {prefetch => 'settings'});
-    return $self->reply->not_found unless $job;
+    my $follow = $self->param('follow');
+    return unless my $job = $self->find_job_or_render_not_found($job_id, $follow ? {prefetch => 'settings'} : {});
+    $job = $job->latest_job if $follow;
     $job = $job->to_hash(assets => 1, check_assets => $check_assets, deps => 1, details => $details, parent_group => 1);
+    $job->{followed_id} = $job_id if ($job_id != $job->{id});
     $self->render(json => {job => $job});
 }
 
@@ -591,15 +594,21 @@ sub update_status ($self) {
 Retrieve status of a job. Returns id, state, result, blocked_by_id.
 Preferable over /job/<id> for performance and payload size, if you are only
 interested in the status.
+Pass follow=1 as query param to follow job clones and report most recent result for given id.
 
 =back
 
 =cut
 
 sub get_status ($self) {
+    my $follow = $self->param('follow');    # follow job clones and report most recent result for given id
     my @fields = qw(id state result blocked_by_id);
-    return unless my $job = $self->find_job_or_render_not_found($self->stash('jobid'));
-    $self->render(json => {map { $_ => $job->$_ } @fields});
+    my $jobid = $self->stash('jobid');
+    return unless my $job = $self->find_job_or_render_not_found($jobid);
+    $job = $job->latest_job if $follow;
+    my $json = {map { $_ => $job->$_ } @fields};
+    $json->{followed_id} = $jobid if $jobid != $job->id;
+    $self->render(json => $json);
 }
 
 =over 4
--- a/lib/OpenQA/CLI/monitor.pm
+++ b/lib/OpenQA/CLI/monitor.pm
@@ -11,11 +11,13 @@ use Mojo::Util qw(encode getopt);
 has description => 'Monitors a set of jobs';
 has usage => sub { shift->extract_usage };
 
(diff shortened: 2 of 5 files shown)


lib/OpenQA/WebAPI/Controller/API/V1/Job.pm
```diff
@@ -453,9 +453,25 @@ sub show ($self) {
...
     my $details = $self->stash('details') || 0;
     my $check_assets = !!$self->param('check_assets');
-    my $job = $self->schema->resultset('Jobs')->find($job_id, {prefetch => 'settings'});
-    return $self->reply->not_found unless $job;
+    my $follow = $self->param('follow');    # follow job clones and report most recent result for given id
+
+    my @cloned_as;
+    my $job;
+    while (1) {
+        return unless $job = $self->find_job_or_render_not_found($job_id, {prefetch => 'settings'});
```
perlpunk, line 461:
```suggestion
        return $self->reply->not_found unless $job = $self->find_job_or_render_not_found($job_id, {prefetch => 'settings'});
```

perlpunk: Actuallu, nevermind. `find_job_or_render_not_found` handles that already

lib/OpenQA/WebAPI/Controller/API/V1/Job.pm
```diff
@@ -453,9 +453,25 @@ sub show ($self) {
...
+        if ($follow && $job->clone_id) {
+            push(@cloned_as, $job->clone_id);
```
perlpunk, line 463:
```suggestion
            push @cloned_as, $job->clone_id;
```

lib/OpenQA/WebAPI/Controller/API/V1/Job.pm
```diff
@@ -597,9 +613,28 @@ interested in the status.
...
     my @fields = qw(id state result blocked_by_id);
-    return unless my $job = $self->find_job_or_render_not_found($self->stash('jobid'));
-    $self->render(json => {map { $_ => $job->$_ } @fields});
+    my $query_jobid = $self->stash('jobid');
+    my @cloned_as;
+    my $job;
+    while (1) {
+        return unless $job = $self->find_job_or_render_not_found($query_jobid);
+        if ($follow && $job->clone_id) {
+            push(@cloned_as, $job->clone_id);
```
perlpunk, line 624:
```suggestion
            push @cloned_as, $job->clone_id;
```

lib/OpenQA/WebAPI/Controller/API/V1/Job.pm
```diff
@@ -453,9 +453,25 @@ sub show ($self) {
...
+    my $follow = $self->param('follow');    # follow job clones and report most recent result for given id
+
+    my @cloned_as;
+    my $job;
+    while (1) {
+        return unless $job = $self->find_job_or_render_not_found($job_id, {prefetch => 'settings'});
+        if ($follow && $job->clone_id) {
+            push(@cloned_as, $job->clone_id);
+            $job_id = $job->clone_id;
+            next;
```
Martchus, line 465: This should be its own function, like the `latest_job` function we already have (but tracking the whole chain of IDs). Especially since you're duplicating it later on. We could also simplify this to use the existing `latest_job` function and not return the whole chain of IDs.

You could also use *one* recursive SQL query instead of doing one per clone. Search for `sub descendants` or `sub incomplete_ancestors` for functions where we already do this. This would just be an optimization but maybe a worthwhile one and the SQL code for this is actually quite interesting.

lib/OpenQA/WebAPI/Controller/API/V1/Job.pm
```diff
@@ -453,9 +453,25 @@ sub show ($self) {
...
+        }
+        last;
+    }
     $job = $job->to_hash(assets => 1, check_assets => $check_assets, deps => 1, details => $details, parent_group => 1);
+    if ($job_id != $self->stash('jobid')) {
+        # if follow=1 return cloned job with original id
+        # but indicate clone chain in cloned_as
+        $job->{cloned_as} = \@cloned_as;
+    }
```
Martchus, line 474: I'd avoid the repeated querying of the job ID from the stash and write it as one-liner:

```suggestion
    $job->{cloned_as} = \@cloned_as if @cloned_as;
```

There's probably also no need for the comment.

In this short form the duplication you have in the other function is probably even acceptable.

lib/OpenQA/WebAPI/Controller/API/V1/Job.pm
```diff
@@ -597,9 +613,28 @@ interested in the status.
...
+            $query_jobid = $job->clone_id;
+            next;
+        }
+        last;
+    }
+    my $json = {map { $_ => $job->$_ } @fields};
+    if ($query_jobid != $self->stash('jobid')) {
+        # if follow=1 return cloned job with original id
+        # but indicate clone chain in cloned_as
+        $json->{id} = $self->stash('jobid');
```
Martchus, line 634: This shouldn't return the original job ID anymore, or should it? I would say that with `--follow` one expects all data to be returned for the latest job including the ID. One can still read `cloned_as` to get the original ID.

Martchus (changes requested):

lib/OpenQA/WebAPI/Controller/API/V1/Job.pm
asdil12: If it doesn't, it won't work for the API methods for `openqa-cli monitoring` mentioned explicitly in the ticket. And there would be no way for the user calling openqa-cli to know which job id corresponds to which id they passed as a parameter.

okurz (changes requested):
Needs tidying with potential rebase first: https://app.circleci.com/pipelines/github/os-autoinst/openQA/15527/workflows/fcaa6b03-d81f-4642-bda5-2a9db29a57d9/jobs/148927

Martchus:
Looks like you need to run tidy and also be sure everything is based on the latest master.

okurz (approved):

PR #6128, opened 2025-01-20
Improve help texts of `openqa-cli`

Martchus:
* Fix documentation of syntax for displaying help text of nested commands
* Add a short description of what the schedule and monitor commands do on their help texts

### DIFF
--- a/lib/OpenQA/CLI/monitor.pm
+++ b/lib/OpenQA/CLI/monitor.pm
@@ -56,6 +56,9 @@ sub command ($self, @args) {
 
   Usage: openqa-cli monitor [OPTIONS] [JOB_IDS]
 
+  Wait until all specified jobs have reached a final state and return a non-zero
+  exit code if at least one job is not ok (is not passed or softfailed).
+
   Options:
         --apibase <path>           API base, defaults to /api/v1
         --apikey <key>             API key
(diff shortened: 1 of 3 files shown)


perlpunk (approved):

lib/OpenQA/CLI/monitor.pm
```diff
@@ -56,6 +56,9 @@ sub command ($self, @args) {
 
   Usage: openqa-cli monitor [OPTIONS] [JOB_IDS]
 
+  Wait until all specified jobs are done/cancelled and return non-zero exit code
```
okurz, line 59:
```suggestion
  Wait until all specified jobs have reached a final state and return non-zero exit code
```

that's according to lib/OpenQA/Jobs/Constants.pm

lib/OpenQA/CLI/monitor.pm
```diff
@@ -56,6 +56,9 @@ sub command ($self, @args) {
...
+  if at least on job has not passed/softfailed.
```
okurz, line 60:
```suggestion
  if at least one job is not ok.
```

so s/on/one/ and the second also according to lib/OpenQA/Jobs/Constants.pm

okurz (changes requested):

okurz (approved):

PR #6131, opened 2025-01-21
Add SYNC_ASSETS_HOOK support

marmarek:
This is useful when having site-local cache shared among several worker
hosts. See added documentation for scenario details.

Question: Should this setting be also exclusive to workers.ini, similar to `GENERAL_HW_CMD_DIR` and few others?

### DIFF
--- a/lib/OpenQA/Worker/Engines/isotovideo.pm
+++ b/lib/OpenQA/Worker/Engines/isotovideo.pm
@@ -326,6 +326,12 @@ sub engine_workit ($job, $callback) {
 
     log_debug "Job settings:\n" . format_settings(\%vars);
 
+    # start pre-job hook if any
+    if ($job_settings->{SYNC_ASSETS_HOOK}) {
+        log_debug 'Running SYNC_ASSETS_HOOK';
+        system($job_settings->{SYNC_ASSETS_HOOK});
+    }
+
     # cache/locate assets, set ASSETDIR
     my $assetkeys = detect_asset_keys(\%vars);
     if (my $cache_dir = $global_settings->{CACHEDIRECTORY}) {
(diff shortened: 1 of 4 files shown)


lib/OpenQA/Worker/Engines/isotovideo.pm
```diff
@@ -326,6 +326,12 @@ sub engine_workit ($job, $callback) {
 
     log_debug "Job settings:\n" . format_settings(\%vars);
 
+    # start pre-job hook if any
+    if ($job_settings->{SYNC_ASSETS_HOOK}) {
+        log_debug 'Running SYNC_ASSETS_HOOK';
+        system($job_settings->{SYNC_ASSETS_HOOK});
+    }
```
Martchus, line 333: Since this is essentially just a pre-job hook as stated by your own comment it would probably make sense to call the setting `JOB_PRE_EXECUTION_HOOK` as well.

marmarek: I was considering this, but then it turned out to be needed at developer mode unpause too...

marmarek:
ping?

Martchus (approved):

kalikiana (approved):

perlpunk:
I would suggest to add that new variable here: https://github.com/os-autoinst/openQA/blob/master/lib/OpenQA/Worker/Job.pm#L240
We don't want people to add that setting to a job; it's supposed to be only set by the workers.ini, right?

perlpunk:
It's possible that there are more variables that should be only allowed via workers.ini.
I already talked about that in the past, and I think it's a bit unfortunate that all are ending up in the same variable.
Anything that can mess with the host system running qemu should probably forbidden.
Scheduling jobs cannot be done by any random person, so there is some protection, but another layer of protection can't hurt, especially since we can schedule jobs via PR comments now.

perlpunk:
I created #6160

PR #6135, opened 2025-01-24
Allow openQA worker to execute check_qemu_oom script in apparmor

asdil12:
Ticket: https://progress.opensuse.org/issues/176049

### DIFF
--- a/profiles/apparmor.d/usr.share.openqa.script.worker
+++ b/profiles/apparmor.d/usr.share.openqa.script.worker
@@ -132,7 +132,7 @@
   /usr/lib{,exec}/git{,-core}/git-remote-https rix,
   /usr/lib{,exec}/git{,-core}/git-write-tree rix,
   /usr/lib/os-autoinst/videoencoder rix,
-  /usr/lib/os-autoinst/check_qemu_oom arCx,
+  /usr/lib/os-autoinst/script/check_qemu_oom arCx,
   /usr/bin/ffmpeg rix,
   /usr/lib/utempter/utempter rix,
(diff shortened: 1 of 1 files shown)


okurz (approved):

profiles/apparmor.d/usr.share.openqa.script.worker
```diff
@@ -133,6 +133,7 @@
   /usr/lib{,exec}/git{,-core}/git-write-tree rix,
   /usr/lib/os-autoinst/videoencoder rix,
   /usr/lib/os-autoinst/check_qemu_oom arCx,
```
Martchus, line 135:
```suggestion
```

Martchus (changes requested):
The line `profile /usr/lib/os-autoinst/check_qemu_oom {` probably needs to be adjusted as well.

asdil12:
Seems like this is some outfall of https://github.com/os-autoinst/os-autoinst/pull/1978

okurz:
> Seems like this is some outfall of [os-autoinst/os-autoinst#1978](https://github.com/os-autoinst/os-autoinst/pull/1978)

yes, very likely. I wonder why we only see the problem now then

ggardet:
> > Seems like this is some outfall of [os-autoinst/os-autoinst#1978](https://github.com/os-autoinst/os-autoinst/pull/1978)
> 
> yes, very likely. I wonder why we only see the problem now then

I have seen it in the past sporadically (cannot tell since when), but never took time to report it properly until recently.

Martchus (approved):

PR #6140, opened 2025-01-28
Prevent git from hanging a process

b10n1k:
* Use of a timeout when IPC::Run invokes
* Enhance error message and notify user for likely solution

I tried to catch the prompt but I assume that the process is not running in a tty and IPC::Run could not get the output.

https://progress.opensuse.org/issues/174592

### DIFF
--- a/lib/OpenQA/Git.pm
+++ b/lib/OpenQA/Git.pm
@@ -30,9 +30,11 @@ sub _validate_attributes ($self) {
 
 sub _run_cmd ($self, $args, $options = {}) {
     my $include_git_path = $options->{include_git_path} // 1;
-    my $ssh_batchmode = $options->{ssh_batchmode} // 0;
+    #warn(">>>> $options");
+    my $batchmode = $options->{batchmode} // 0;
     my @cmd;
-    push @cmd, 'env', 'GIT_SSH_COMMAND=ssh -oBatchMode=yes' if $ssh_batchmode;
+    push @cmd, 'env', 'GIT_SSH_COMMAND=ssh -oBatchMode=yes' if $batchmode;
+    #push @cmd, 'env', $options->{nopass} if $options->{nopass};
     push @cmd, $self->_prepare_git_command($include_git_path), @$args;
 
     my $result = run_cmd_with_log_return_error(\@cmd);
@@ -97,6 +99,7 @@ sub commit ($self, $args = undef) {
 
     # push changes
     if (($self->config->{do_push} || '') eq 'yes') {
+        #my $options->{nopass} = "GIT_TERMINAL_PROMPT=false";
         $res = $self->_run_cmd(['push']);
         return $self->_format_git_error($res, 'Unable to push Git commit') unless $res->{status};
     }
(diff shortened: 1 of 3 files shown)


lib/OpenQA/Utils.pm
```diff
@@ -349,20 +353,30 @@ sub run_cmd_with_log_return_error ($cmd, %args) {
...
             stderr => $stderr,
         };
     }
     catch {
-        return {
-            status => 0,
-            return_code => undef,
-            stderr => 'an internal error occurred',
-            stdout => '',
+        my $additional_info = "If you use http for CASEDIR or NEEDLES_DIR is adviced to
```
perlpunk, line 363: `run_cmd_with_log_return_error` is a general purpose function not only used for git commands.
It doesn't make sense to add a message about git here.

Also, we shouldn't add lengthy directions into the error message, but link to the documentation, as it was suggested in the ticket

lib/OpenQA/Utils.pm
```diff
@@ -349,20 +353,30 @@ sub run_cmd_with_log_return_error ($cmd, %args) {
...
+        configure git with the insteadof option in the .gitconfig file.
+        Run the following command to:
+        git config [--global] url.\"git://github.com\".insteadOf https://github.com";
```
perlpunk, line 366: `url.insteadOf` is **not** what we want. it is `url.pushInsteadOf`, which is an important difference.
We still want the **fetch** to happen with http.

lib/OpenQA/Git.pm
```diff
@@ -30,9 +30,11 @@ sub _validate_attributes ($self) {
 
 sub _run_cmd ($self, $args, $options = {}) {
     my $include_git_path = $options->{include_git_path} // 1;
-    my $ssh_batchmode = $options->{ssh_batchmode} // 0;
+    #warn(">>>> $options");
+    my $batchmode = $options->{batchmode} // 0;
     my @cmd;
-    push @cmd, 'env', 'GIT_SSH_COMMAND=ssh -oBatchMode=yes' if $ssh_batchmode;
+    push @cmd, 'env', 'GIT_SSH_COMMAND=ssh -oBatchMode=yes' if $batchmode;
```
perlpunk, line 36: Why did you not use `GIT_ASKPASS` and `GIT_TERMINAL_PROMPT` like suggested in the ticket?
Waiting for a timing out command seems way more complicated than just telling git to not prompt in the first place.

perlpunk (commented):
See my alternative suggestion: #6141

okurz (changes requested):
rebase on top of https://github.com/os-autoinst/openQA/pull/6141

lib/OpenQA/Git.pm
b10n1k: I was thinking that with the `timeout` we gonna have a general solution when the _run_cmd (or more specifically  run_cmd_with_log_return_error) encounter similar issue. 
GIT_ASKPASS and GIT_TERMINAL_PROMPT make the logic more complicated IMO but also provide a better error as I didnt find a way to catch the prompt in the $stdout somehow. 
Maybe both can be applied at the end.

b10n1k:
> rebase on top of #6141

I didnt notice and I didnt know about this. Again two ppl work on the same thing

b10n1k:
> > rebase on top of #6141
> 
> I didnt notice and I didnt know about this. Again two ppl work on the same thing

This is not merged tho. what do you mean?

lib/OpenQA/Git.pm
perlpunk: I'm not sure what's complicated about using two env vars. It's as simple as it gets?

b10n1k:
closed in favor of  #6141

PR #6153, opened 2025-02-03
Improve Git-related log messages

Martchus:
See https://progress.opensuse.org/issues/169522 and particular commit messages

### DIFF
--- a/lib/OpenQA/Git.pm
+++ b/lib/OpenQA/Git.pm
@@ -118,7 +118,7 @@ sub get_current_branch ($self) {
 
 sub check_sha ($self, $sha) {
     my $r = $self->_run_cmd(['rev-parse', '--verify', '-q', $sha]);
-    return 0 if $r->{return_code} == 128;
+    return 0 if $r->{return_code} == 1;
     if ($r->{return_code} == 0) {
         # rev-parse returns a sha, $sha could be a branchname. if $sha matches
         # the beginning of the returned sha, we're good
--- a/lib/OpenQA/Git.pm
+++ b/lib/OpenQA/Git.pm
@@ -12,6 +12,8 @@ has 'app';
 has 'dir';
 has 'user';
 
+my %CHECK_OPTIONS = (expected_return_codes => {0 => 1, 1 => 1});
+
 sub enabled ($self, $args = undef) {
     die 'no app assigned' unless my $app = $self->app;
     return ($app->config->{global}->{scm} || '') eq 'git';
@@ -35,7 +37,7 @@ sub _run_cmd ($self, $args, $options = {}) {
     push @cmd, 'env', 'GIT_SSH_COMMAND=ssh -oBatchMode=yes', 'GIT_ASKPASS=', 'GIT_TERMINAL_PROMPT=false' if $batchmode;
     push @cmd, $self->_prepare_git_command($include_git_path), @$args;
 
-    my $result = run_cmd_with_log_return_error(\@cmd);
+    my $result = run_cmd_with_log_return_error(\@cmd, expected_return_codes => $options->{expected_return_codes});
     $self->app->log->error("Git command failed: @cmd - Error: $result->{stderr}") unless $result->{status};
     return $result;
 }
@@ -116,8 +118,13 @@ sub get_current_branch ($self) {
     return trim($r->{stdout});
 }
 
+sub invoke_command ($self, $command_and_args) {
+    my $r = $self->_run_cmd($command_and_args);
+    $r->{status} or die $self->_format_git_error($r, "Failed to invoke Git command $command_and_args->[0]");
+}
+
 sub check_sha ($self, $sha) {
-    my $r = $self->_run_cmd(['rev-parse', '--verify', '-q', $sha]);
+    my $r = $self->_run_cmd(['rev-parse', '--verify', '-q', $sha], \%CHECK_OPTIONS);
     return 0 if $r->{return_code} == 1;
     if ($r->{return_code} == 0) {
         # rev-parse returns a sha, $sha could be a branchname. if $sha matches
@@ -157,7 +164,7 @@ sub reset_hard ($self, $branch) {
 }
 
 sub is_workdir_clean ($self) {
-    my $r = $self->_run_cmd(['diff-index', 'HEAD', '--exit-code']);
+    my $r = $self->_run_cmd(['diff-index', 'HEAD', '--exit-code'], \%CHECK_OPTIONS);
     die $self->_format_git_error($r, 'Internal Git error: Unexpected exit code ' . $r->{return_code})
       if $r->{return_code} > 1;
     return $r->{status};
--- a/lib/OpenQA/Task/Git/Clone.pm
+++ b/lib/OpenQA/Task/Git/Clone.pm
@@ -85,9 +85,9 @@ sub _git_clone ($app, $job, $ctx, $path, $url) {
 
         my $origin_url = $git->get_origin_url;
         if ($url ne $origin_url) {
-            $ctx->info(
-"Local checkout at $path has origin $origin_url but requesting to clone from $url. Be aware that the requested URL will not be cloned. This is fine.🔥"
-            );
+            $ctx->info(<<~"END_OF_MESSAGE");
+                Local checkout at $path has origin $origin_url but requesting to clone from $url. The requested URL will not be cloned.
+                END_OF_MESSAGE
             return;
         }
     }
--- a/lib/OpenQA/Task/Git/Clone.pm
+++ b/lib/OpenQA/Task/Git/Clone.pm
@@ -97,7 +97,11 @@ sub _git_clone ($app, $job, $ctx, $path, $url) {
 
     return if ($requested_branch and $requested_branch !~ tr/a-f0-9//c and $git->check_sha($requested_branch));
 
-    die "NOT updating dirty git checkout at $path" if !$git->is_workdir_clean();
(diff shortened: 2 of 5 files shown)


lib/OpenQA/Task/Git/Clone.pm
```diff
@@ -97,7 +97,10 @@ sub _git_clone ($app, $job, $ctx, $path, $url) {
 
     return if ($requested_branch and $requested_branch !~ tr/a-f0-9//c and $git->check_sha($requested_branch));
 
-    die "NOT updating dirty git checkout at $path" if !$git->is_workdir_clean();
+    die <<~"END_OF_ERROR_MESSAGE" unless $git->is_workdir_clean;
+        NOT updating dirty Git checkout at '$path'.
+        You may disable auto-updating by setting 'git_auto_update = no' under '[scm git]' in the config.
```
okurz, line 102: That sounds like "just always disable git_auto_update, it only makes problems" which is not what we want. How about pointing to http://open.qa/docs/#_getting_tests for details and potentially also state the bad consequence

okurz (changes requested):

perlpunk (changes requested):
You have to also update the tests in
* t/14-grutasks-git.t where we mock the return code
* lib/OpenQA/Git.pm where we check the return_code
* possible other places where we use `run_cmd_with_log_return_error`

ci passes because we mock the return code.

okurz (approved):

lib/OpenQA/Task/Git/Clone.pm
```diff
@@ -97,7 +97,11 @@ sub _git_clone ($app, $job, $ctx, $path, $url) {
 
     return if ($requested_branch and $requested_branch !~ tr/a-f0-9//c and $git->check_sha($requested_branch));
 
-    die "NOT updating dirty git checkout at $path" if !$git->is_workdir_clean();
+    die <<~"END_OF_ERROR_MESSAGE" unless $git->is_workdir_clean;
```
perlpunk, line 100: I think one problem might still be that this error will not appear in the logfile, only as the reason for the failed minion job. At least the original question was about the logfile not printing a helpful message

Martchus:
I don't think I have to update `t/14-grutasks-git.t` because there we already mock the return code as it is supposed to be. I'm only changing the underlying function here and adjusted/extended tests accordingly where this function is tested.

I didn't find an invocation of `run_cmd_with_log_return_error` where we expect the old behavior so I don't think there's anything else to be adjusted.

The original question came up [on Slack](https://suse.slack.com/archives/C02CANHLANP/p1730970100931999) where the user said "Hi colleagues, I get an error from openqa-gru when I try to start a test." followed by an excerpt of the Gru logs (probably from their local openQA instance). The changed log message would end up in these logs. So I think the user would have seen them. When I remember correctly, the error/exception from a failing Gru job might even end up on the relevant openQA job but I needed to verify that if it is important.

perlpunk:
We're setting `$return_code` here in several places: https://github.com/os-autoinst/openQA/blob/master/t/14-grutasks-git.t#L67
Here we set it to 128: https://github.com/os-autoinst/openQA/blob/master/t/14-grutasks-git.t#L100

Here we check for code 128:
* https://github.com/os-autoinst/openQA/blob/master/lib/OpenQA/Git.pm#L104
* https://github.com/os-autoinst/openQA/blob/master/lib/OpenQA/Git.pm#L121

perlpunk:
Ok, I think I get it now.
Then there is only one question left:
https://github.com/os-autoinst/openQA/blob/master/lib/OpenQA/Git.pm#L39
Here we check the `status` entry, so for the `is_workdir_clean` this would still print the output into the log because `status` is only true if the return code was zero.

t/16-utils-runcmd.t
```diff
@@ -32,11 +32,22 @@ subtest 'run (arbitrary) command' => sub {
...
-    ok(!$res->{status}, 'status not ok (non-zero status returned)');
+    is $res->{return_code}, 1, 'correct non-zero exit code returned ($? >> 8)';
+    ok !$res->{status}, 'status not ok (non-zero status returned)';
+
+    stdout_unlike { $res = run_cmd_with_log_return_error([qw(falße)]) } qr/.*cmd returned [1-9][0-9]*/i;
+    is $res->{return_code}, undef, 'no exit code returned if command cannot be executed';
+    is $res->{stderr}, 'an internal error occurred', 'error message returned as stderr';
+    ok !$res->{status}, 'status not ok if command cannot be executed';
+
+    stdout_like { $res = run_cmd_with_log_return_error([qw(bash -c), 'kill -s KILL $$']) } qr/.*cmd died with signal 9\n.*/i;
```
Wabri, line 48: Not a proposal, I just want to understand why not use something like this:

`run_cmd_with_log_return_error(['bash', '-c', 'kill -s KILL $$'])`

t/16-utils-runcmd.t
```diff
@@ -32,11 +32,22 @@ subtest 'run (arbitrary) command' => sub {
     stdout_like { is(run_cmd_with_log([qw(false)]), '') } qr/[WARN].*[ERROR]/i;
 
     my $res = run_cmd_with_log_return_error([qw(echo Hallo Welt)]);
-    ok($res->{status}, 'status ok');
-    is($res->{stdout}, "Hallo Welt\n", 'cmd output returned');
+    is $res->{return_code}, 0, 'correct zero exit code returned ($? >> 8)';
+    ok $res->{status}, 'status ok';
+    is $res->{stdout}, "Hallo Welt\n", 'cmd output returned';
```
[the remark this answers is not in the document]
Martchus: I'm just reformatting this line. I guess the reason is that several Germans work and used to work on this project.

t/16-utils-runcmd.t
Martchus: I guess you version is better. I just started out with qw and the only moved out what I had to.

t/16-utils-runcmd.t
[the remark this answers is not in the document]
okurz: @Wabri we shouldn't point out things in code review that was already there and only review the changes. What you can do is create your own draft PR for some ideas on old code, like in this case.

Martchus:
@perlpunk 

> We're setting `$return_code` here in several places: https://github.com/os-autoinst/openQA/blob/master/t/14-grutasks-git.t#L67 Here we set it to 128: https://github.com/os-autoinst/openQA/blob/master/t/14-grutasks-git.t#L100
> 
> Here we check for code 128:
> 
>     * https://github.com/os-autoinst/openQA/blob/master/lib/OpenQA/Git.pm#L104
> 
>     * https://github.com/os-autoinst/openQA/blob/master/lib/OpenQA/Git.pm#L121

Right, I checked these occurrences now. It seems that the real return code we have to check for is `1` and with my change we also need to adjust that code. So I did that and also extended the unit tests to test with real Git and not just rely on mocking. (The test runtime is still acceptable.)



> Ok, I think I get it now. Then there is only one question left: https://github.com/os-autoinst/openQA/blob/master/lib/OpenQA/Git.pm#L39 Here we check the `status` entry, so for the `is_workdir_clean` this would still print the output into the log because `status` is only true if the return code was zero.

I fixed this in my new commit. I also fixed the `check_sha` function because both are just checks.

okurz (approved):

lib/OpenQA/Git.pm
```diff
@@ -101,7 +103,7 @@ sub commit ($self, $args = undef) {
         return undef if $res->{status};
 
         my $msg = 'Unable to push Git commit';
-        if ($res->{return_code} == 128 and $res->{stderr} =~ m/Authentication failed for .http/) {
+        if ($res->{return_code} == 1 and $res->{stderr} =~ m/Authentication failed for .http/) {
```
perlpunk, line 106: I'm sorry that I previously was confusing. But the return code here is 32768, so that would be 128 with `>> 8`

lib/OpenQA/Git.pm
```diff
@@ -116,9 +118,14 @@ sub get_current_branch ($self) {
...
+sub init ($self) {
+    my $r = $self->_run_cmd(['init']);
+    $r->{status} or die $self->_format_git_error($r, 'Failed to init repo');
+}
+
 sub check_sha ($self, $sha) {
-    my $r = $self->_run_cmd(['rev-parse', '--verify', '-q', $sha]);
-    return 0 if $r->{return_code} == 128;
+    my $r = $self->_run_cmd(['rev-parse', '--verify', '-q', $sha], \%CHECK_OPTIONS);
+    return 0 if $r->{return_code} == 1;
```
perlpunk, line 128: I also checked again this one. This is also 32768, so 128 is the right value

lib/OpenQA/Git.pm
Martchus: Interesting. I provoked an auth error via SSH and got a `1` return code. However, now I provoked one via HTTP and got `128`. So apparently it depends on the protocol. I'll change it back then.

lib/OpenQA/Git.pm
Martchus: For this one I always get `0` in case the sha exists and `1` in case it doesn't and something else if the check didn't work. So I guess `1` is correct here.

perlpunk: oh, interesting. the `-q` parameter makes a difference. yeah, it's 1 then.

perlpunk (approved):

t/16-utils-runcmd.t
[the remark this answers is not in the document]
Wabri: @okurz I asked, not requested!

PR #6154, opened 2025-02-04
Support OpenID Connect better in the OAuth2 'custom' provider.

ssssam:
The OpenID Connect standard specifies that `sub` contains the user ID. Until now, openQA only looked for a field named `id`.

When using Keycloak as identity provider there is no such field, it returns only `sub` as expected in OpenID Connect.

To avoid breaking any existing configs, this patch adds a new `id_from` config field which defaults to `id`, so existing behaviour is preserved. Set `id_from = sub` in the OAuth2 provider config to get the new behaviour.

### DIFF
--- a/lib/OpenQA/Setup.pm
+++ b/lib/OpenQA/Setup.pm
@@ -96,6 +96,7 @@ sub read_config ($app) {
             token_label => '',
             nickname_from => '',
             unique_name => '',
+            id_from => 'id',
         },
         hypnotoad => {
             listen => ['http://localhost:9526/'],
--- a/etc/openqa/openqa.ini
+++ b/etc/openqa/openqa.ini
@@ -148,7 +148,11 @@
 #user_url  = https://salsa.debian.org/api/v4/user
 #token_scope = read_user
 #token_label = Bearer
+#id_from = id
 #nickname_from = username
+#
+# The default value for `id_from` is "id" for backwards compatibility.
+# For OpenID Connect, the value should be "sub", following the OpenID Connect ID Token standard.
 
 [logging]
 ## logging is to stderr (so systemd journal) by default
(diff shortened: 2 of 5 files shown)


lib/OpenQA/Setup.pm
```diff
@@ -96,6 +96,7 @@ sub read_config ($app) {
             token_label => '',
             nickname_from => '',
             unique_name => '',
+            id_from => 'id',
```
okurz, line 99: can you add the according config option also to etc/openqa.ini with an explanatory comment from your git commit message?

okurz (changes requested):

lib/OpenQA/Setup.pm
Martchus: It looks like `t/config.t` also needs to be adjusted.

ssssam: All the test issues should be resolved now. Also added the comment to `etc/openqa.ini`. Thanks for the reviews

Martchus (approved):

Martchus:
Looks like the Codecov check couldn't be triggered. The related CircleCI job ran for "0 s" and if one clicks on the job one gets only "Something Unexpected Happened". All buttons for re-running are grayed out and the workflow status is also "Error" (and  not e.g. "Failed").

So I suppose the easiest way to get green checkmarks on the CI is to amend the commit and force-push again.

okurz (approved):

Martchus:
The CI checks failed (or rather "errored") in the same way as before despite the commit being based on a recent commit from master. I also tried to re-deliver the webhook to CircleCI but it didn't have any effect.

Maybe we can merge this forcefully considering all steps of the CircleCI pipeline worked except Codecov? Otherwise you would probably need to create a new PR.

okurz:
I created https://github.com/os-autoinst/openQA/pull/6158 as a re-submit which might help to trigger the codecov report generation

okurz:
Merged as part of
* #6158

thank you for the contribution!

PR #6161, opened 2025-02-07
Propagate json errors to user on fetch requests with non-200 status code

asdil12:
Ticket: https://progress.opensuse.org/issues/176214

### DIFF
--- a/assets/javascripts/admin_needle.js
+++ b/assets/javascripts/admin_needle.js
@@ -160,8 +160,21 @@ function setupAdminNeedles() {
 
       fetchWithCSRF(url + nextIDs.join('&id='), {method: 'DELETE'})
         .then(response => {
-          if (!response.ok) throw `Server returned ${response.status}: ${response.statusText}`;
-          return response.json();
+          return response
+            .json()
+            .then(json => {
+              // Attach the parsed JSON to the response object for further use
+              return {response, json};
+            })
+            .catch(() => {
+              // If parsing fails, handle it as a non-JSON response
+              throw `Server returned ${response.status}: ${response.statusText}`;
+            });
+        })
+        .then(({response, json}) => {
+          // we got parsable json and this api method has pretty custom errors
+          // so let's ignore the http status in this case and use json based error handling below
+          return json;
         })
         .then(response => {
           // add error affecting all deletions
--- a/assets/javascripts/fetch_status.js
+++ b/assets/javascripts/fetch_status.js
@@ -1,3 +1,6 @@
+// propagate window.runningFetchRequests variable which
+// is used in unit tests to see if there are still fetch
+// requests running
 window.runningFetchRequests = 0;
 
 window.originalFetch = window.fetch;
--- a/lib/OpenQA/Shared/Controller/Session.pm
+++ b/lib/OpenQA/Shared/Controller/Session.pm
@@ -25,8 +25,14 @@ sub ensure_operator {
 
 sub ensure_admin {
     my ($self) = @_;
-    $self->redirect_to($self->url_for('login')->query(return_page => $self->req->url)) and return undef
-      unless $self->current_user;
+    unless ($self->current_user) {
+        if (($self->tx->req->headers->accept // '') eq 'application/json') {
+            $self->render(json => {'error' => 'No valid user session'}, status => 401);
+        } else {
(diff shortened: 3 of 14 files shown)


lib/OpenQA/WebAPI/Controller/Admin/User.pm
```diff
@@ -27,17 +27,25 @@ sub update {
...
     }
     else {
         $user->update({is_admin => $is_admin, is_operator => $is_operator});
-        $self->flash('info', 'User ' . $user->nickname . ' updated');
+        $msg = 'User ' . $user->nickname . ' updated';
         $self->emit_event('user_update_res', {nickname => $user->nickname, role => $role});
     }
 
-    $self->redirect_to($self->url_for('admin_users'));
+    if ($self->tx->req->headers->header('Accept') eq 'application/json') {
```
kalikiana, line 44: `Use of uninitialized value in string eq at /home/squamata/project/lib/OpenQA/WebAPI/Controller/Admin/User.pm line 44.` in t/ui/09-users-list.t

lib/OpenQA/Shared/Controller/Session.pm
```diff
@@ -25,8 +25,14 @@ sub ensure_operator {
 
 sub ensure_admin {
     my ($self) = @_;
-    $self->redirect_to($self->url_for('login')->query(return_page => $self->req->url)) and return undef
-      unless $self->current_user;
+    unless ($self->current_user) {
+        if ($self->tx->req->headers->header('Accept') eq 'application/json') {
```
kalikiana, line 29: `Use of uninitialized value in string eq at /home/squamata/project/lib/OpenQA/Shared/Controller/Session.pm line 29` from t/03-auth.t

Martchus:
```suggestion
        if (($self->tx->req->headers->accept // '') eq 'application/json') {
```

lib/OpenQA/WebAPI/Controller/Admin/User.pm
Martchus:
```suggestion
    if (($self->tx->req->headers->accept // '') eq 'application/json') {
```

assets/javascripts/fetch_status.js
```diff
@@ -1,3 +1,6 @@
+// propagate window.runningFetchRequests variable which
+// is used in unittests to see if there are still fetch
```
Martchus, line 2:
```suggestion
// is used in unit tests to see if there are still fetch
```

assets/javascripts/admin_needle.js
```diff
@@ -160,8 +160,21 @@ function setupAdminNeedles() {
...
+              // Attach the parsed JSON to the response object for further use
+              return {response, json};
+            })
+            .catch(() => {
+              // If parsing fails, handle it as a non-JSON response
+              throw `Server returned ${response.status}: ${response.statusText}`;
+            });
+        })
+        .then(({response, json}) => {
+          // we got parsable json and this api method has pretty custom error(s) handling
```
Martchus, line 175:
```suggestion
          // we got parsable json and this api method has pretty custom errors
```

assets/javascripts/admin_needle.js
```diff
@@ -160,8 +160,21 @@ function setupAdminNeedles() {
...
+          // so let's ignore http status in this case and use json based error handling below
```
Martchus, line 176:
```suggestion
          // so let's ignore the http status in this case and use json based error handling below
```

Martchus (changes requested):
It would be nice if all of this repetitive code could be moved into a separate function (maybe as a separate commit or a follow-up PR).

okurz (changes requested):
Awaiting CI results. After that at least some git commit squashing needs to be done

asdil12:
> It would be nice if all of this repetitive code could be moved into a separate function (maybe as a separate commit or a follow-up PR).

It is not that repetitive always.
Sometimes the response code isn't checked at all and also the error handling differs.

Martchus (approved):

asdil12:
That should be fixed now.

Martchus (approved):

okurz (approved):

PR #6162, opened 2025-02-07
Allow reading web UI config from INI files under `openqa.d`

Martchus:
Related ticket: https://progress.opensuse.org/issues/176421

### DIFF
--- a/lib/OpenQA/Setup.pm
+++ b/lib/OpenQA/Setup.pm
@@ -48,22 +48,24 @@ sub _load_config ($app, $defaults, $mode_specific_defaults) {
     my $mode_defaults = $mode_specific_defaults->{$app->mode} // {};
     my $config_path = $ENV{OPENQA_CONFIG} ? path($ENV{OPENQA_CONFIG}) : $app->home->child('etc', 'openqa');
     my $main_config_file = $config_path->child('openqa.ini');
-    my @config_file_paths = @{$config_path->child('openqa.d')->list->grep(qr/\.ini$/)->sort(sub { $b cmp $a })};
-    push @config_file_paths, $main_config_file if -e $main_config_file;
+    my @config_file_paths = -e $main_config_file ? ($main_config_file) : ();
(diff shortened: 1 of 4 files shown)


lib/OpenQA/Setup.pm
```diff
@@ -24,6 +24,49 @@ use OpenQA::Task::Job::Limit;
...
+    my @config_file_paths = @{$config_path->child('openqa.d')->list->grep(qr/\.ini$/)->sort(sub { $b cmp $a })};
+    push @config_file_paths, $main_config_file if -e $main_config_file;
+
+    # read config files
+    my @config_files;
+    for my $config_file_path (@config_file_paths) {
+        my $config_file = Config::IniFiles->new(-file => $config_file_path->to_string);
+        _read_config_file($config, $config_file, $defaults, $mode_defaults);
+        push @config_files, $config_file if $config_file;
+    }
```
okurz, line 60: https://metacpan.org/pod/Config::IniFiles#IMPORT-/-DELTA-FEATURES looks like it should help here to do some things which you might have implemented yourself

okurz (changes requested):

lib/OpenQA/Setup.pm
Martchus: I was searching for such a feature but once I noticed that one can only specify one file at a time I thought this wouldn't be possible.

I guess I can use this import/overlay feature. It looks like one still has to instantiate multiple Config::IniFiles objects building them like a linked list. So this won't be much of an improvement and this loop might even become a little bit more complex (as one now needs to pass the previous Config::IniFiles object to the next one, handling the case of creating the first Config::IniFiles object separately). The only actual benefit would be avoiding the loops in the other functions where `$config->{ini_configs}` is used (which is no real benefit, one of those loops is even just a `List::Util::first`).

kalikiana (approved):

okurz (approved):

PR #6170, opened 2025-02-11
Prevent download prompt on plain text files with Chromium

Martchus:
* Add `X-Content-Type-Options: nosniff` to avoid a download prompt on plain text files with Chromium
* See https://stackoverflow.com/questions/9660514/content-type-of-text-plain-causes-browser-to-download-of-file/9660578#9660578


pevik (approved):
LGTM, thank you!

kalikiana (approved):

PR #6179, opened 2025-02-14
Remove incorrect signatures from documentation

baierjan:


docs/WritingTests.asciidoc
```diff
@@ -133,7 +133,7 @@ An example usage:
 
 [source,perl]
 -------------------------------------------------------------------
-sub test_flags () {
+sub test_flags ($) {
```
okurz, line 136: Are those actually needed? AFAIK the modules are not loaded with full "signatures" support so this would be detected as prototype, not signature, isn't it?

baierjan: I guess it depends, I create my new custom class as
```
  package mybase;
  use Mojo::Base 'basetest', -signatures;
```

and then was surprised with

`Too many arguments for subroutine 'mybase::test_flags' (got 1; expected 0) at /usr/lib/os-autoinst/autotest.pm line 239.`

baierjan: Also it is difference between no signatures

```
sub test_flags { {fatal => 1} }
```

and what we have in documentation
```
sub test_flags () { {fatal => 1} }
```

right?

perlpunk: Note that prototypes are ignored for method calls, and we call it as `$self->test_flags();`. So normally it doesn't make any difference what we specify there.

If you use signatures, that's of course a different thing. But then a `($)` signature I wouldn't consider helpful, it should rather be `($basetest)` or something.

Maybe we should rather go with no prototype / signature in the documentaion by default, and add a section about signatures.

baierjan:
I guess we should rather go the other way around as suggested in https://github.com/os-autoinst/openQA/pull/6179#discussion_r1956597812. It feels to me that without deeper understanding one can quickly get burnt by just copy&pasting the code from our distri-openQA (which has signature enabled) and documentation (where the signature does not match the reality, as noted in https://github.com/os-autoinst/openQA/pull/6179#discussion_r1956587905)

okurz (approved):

PR #6166, opened 2025-02-10
WIP: Use Feature::Compat::Try, remove Try::Tiny, ensure consistent Test::Exception style

okurz:
Only style and dependency changes so far. This shows all places needing a code update.

* wip -- -- Remove obsolete dependency on Try::Tiny
* Add new dependency Feature::Compat::Try
* Use Feature::Compat::Try in place of Try::Tiny
* t: Consistently use Test::Exception method instead of manual eval checks
* t: Use more simple prototyped Test::Exceptions functions
* t: Focus on using Test::Exception only instead of Test::Fatal
* t: Ensure we use implicit Test::Exception in all places

Related progress issues:
* https://progress.opensuse.org/issues/176475
* https://progress.opensuse.org/issues/176862

PR #6189, opened 2025-02-19
t: Use implicit Test::Exception in all places

b10n1k:

### DIFF
--- a/t/01-style.t
+++ b/t/01-style.t
@@ -17,6 +17,7 @@ is qx{git grep --all-match -e '^use Mojo::Base' -e 'use base'}, '', 'No redundan
 is qx{git grep -I --all-match -e '^use Mojo::Base' -e 'use \\(strict\\|warnings\\);' ':!docs'}, '',
   'Only combined Mojo::Base+strict+warnings';
 is qx{git grep -I -L '^use Test::Warnings' t/**.t ':!t/01-style.t'}, '', 'All tests use Test::Warnings';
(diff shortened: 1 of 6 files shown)


okurz (changes requested):
Please focus on https://github.com/os-autoinst/openQA/pull/6166/commits/fc4b0edc053c2e86130a6da2965cd66f25705b99 only, i.e. only remove mentions of Test::Exceptions, and leave out https://github.com/os-autoinst/openQA/pull/6166/commits/eec1d0e33a1b43fc536f2873519dcdb2e44c406d – t: Focus on using Test::Exception only instead of Test::Fatal](https://github.com/os-autoinst/openQA/pull/6166/commits/eec1d0e33a1b43fc536f2873519dcdb2e44c406d) - to a separate PR

b10n1k:
> LGTM, but the title is now wrong and the ticket reference is missing

title updated. But I am not sure about the ticket

perlpunk:
Suggestion:
The first commit title is "t: Ensure we use implicit Test::Exception in all places". That's ok. Because we add a style check that ensures that.
In the second commit you do the change to satisfy that check. So it should be "Use implicit Test::Exception in all places".
The PR title should be "Use implicit Test::Exception in all places" as well.
At least the PR description and one of the commits should reference the ticket, so https://progress.opensuse.org/issues/176862

perlpunk (approved):

okurz (approved):

PR #6191, opened 2025-02-19
t: Use more simple prototyped Test::Exceptions functions

b10n1k:


perlpunk (approved):

okurz (approved):
Well, the message says we would use "more ..." something when in the end this just implements the style function but fine :)

PR #6192, opened 2025-02-19
Improve needle editor tests

Martchus:
See particular commit messages and https://progress.opensuse.org/issues/177564

### DIFF
--- a/t/ui/12-needle-edit.t
+++ b/t/ui/12-needle-edit.t
@@ -328,12 +328,26 @@ subtest 'Needle editor layout' => sub {
     is @{decode_utf8_json(element_prop('needleeditor_textarea'))->{area}}, 2, 'exclude areas always present';
     $driver->find_element_by_xpath('//input[@value="inst-timezone"]')->click();
     is @{decode_utf8_json(element_prop('needleeditor_textarea'))->{area}}, 2, 'no duplicated exclude areas present';
+
+    # put oddly sized png in place for the needle inst-timezone-text to be able to verify it was correctly loaded
+    my $src_dir = path('t/testresults/00099/00099963-opensuse-13.1-DVD-x86_64-Build0091-kde');
+    $src_dir->child('reboot_after_install-1-reboot_after_install-131M4-diff2.png')->copy_to($dir->child('inst-timezone-text.png'));
+
+    # select needle to take image from
+    $driver->execute_script("document.getElementById('image_select').selectedIndex = 1; loadBackground();");
+    is element_prop('image_select'), 'inst-timezone-text', 'selected needle to take image from';
+    sleep 0.01 until $driver->execute_script('return nEditor.bgImage.complete;');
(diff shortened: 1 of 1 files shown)


t/ui/12-needle-edit.t
```diff
@@ -328,12 +333,36 @@ subtest 'Needle editor layout' => sub {
     is @{decode_utf8_json(element_prop('needleeditor_textarea'))->{area}}, 2, 'exclude areas always present';
     $driver->find_element_by_xpath('//input[@value="inst-timezone"]')->click();
     is @{decode_utf8_json(element_prop('needleeditor_textarea'))->{area}}, 2, 'no duplicated exclude areas present';
+
+    # select needle to take image from
+    $driver->execute_script("document.getElementById('image_select').selectedIndex = 1; loadBackground();");
+    is element_prop('image_select'), 'inst-timezone-text', 'selected needle to take image from';
+    sleep 0.1 until $driver->execute_script('return nEditor.bgImage.complete;');
```
okurz, line 340: I consider 100ms quite long for unit tests. Can we use .01?

t/ui/12-needle-edit.t
```diff
@@ -270,7 +271,11 @@ sub check_flash_for_saving_logpackages {
     $driver->find_element('#flash-messages .btn-close')->click();
 }
 
-
+# put oddly sized png in place for the needle inst-timezone-text to be able to verify it was correctly loaded
+my $src_dir = path('t/testresults/00099/00099963-opensuse-13.1-DVD-x86_64-Build0091-kde');
+my $src_needle = $src_dir->child('reboot_after_install-1-reboot_after_install-131M4-diff2.png');
+my $src_needle_sha = sha1_sum $src_needle->slurp;
+$src_needle->copy_to($dir->child('inst-timezone-text.png'));
```
okurz, line 278: Does this do the action in a temp directory or pollute the working directory? How about mocking the hashing as we don't really need to test that hashing files works, right?

okurz (changes requested):

t/ui/12-needle-edit.t
Martchus: It pollutes the working directory like the whole test already does. The test cleans up the directory it pollutes before it runs so potential leftovers are not a problem. The particular directory is also ignored by Git. So this is not a big deal. Changing this would mean changing the whole test. 

Note that this does not invoke the hashing function to test hashing (as no hashing is done by the code we test here). I only use it to compare whether the expected file has been created when the needle is saved (and not a different file due to a regression).

b10n1k (approved):

t/ui/12-needle-edit.t
Martchus: Ok, I can change that.

perlpunk (approved):

okurz (approved):

PR #6193, opened 2025-02-19
t: Focus on using Test::Exception only instead of Test::Fatal

b10n1k:


Martchus (commented):
I'd replace "Focus on using" with just "Use" as it is simpler and conveys the same meaning.

Martchus (approved):

perlpunk (approved):

PR #6194, opened 2025-02-20
t: Use Test::Exception method instead of manual eval checks

b10n1k:


t/15-assets.t
```diff
@@ -102,11 +102,9 @@ $jobA->set_prio(1);
 # register worker
 my $c = OpenQA::WebAPI::Controller::API::V1::Worker->new;
 my $w;
-eval { $w = $c->_register($schema, 'host', '1', $workercaps); };
-like($@, qr/Incompatible websocket API version/, 'Worker no version - incompatible version exception');
+throws_ok { $w = $c->_register($schema, 'host', '1', $workercaps); } qr/Incompatible websocket API version/, 'Worker no version - incompatible version exception';
```
okurz, line 105: That should work without the inner ;

t/43-cli-archive.t
```diff
@@ -1,3 +1,4 @@
+
```
okurz, line 1: I guess that was an accident?

okurz (changes requested):

t/15-assets.t
Martchus: And the same counts for all the other occurrences. I also suggest to omit the inner `;`.

b10n1k: yea. I did take care of the other situation. Except one which the test failed once I removed it

Martchus (approved):

t/16-utils-job-templates.t
```diff
@@ -28,9 +28,7 @@ my $template = {
 };
 my $errors = validate_data(%default_args, data => $template,);
 is scalar @$errors, 0, "Empty template - no errors";
-
-eval { my $errors = validate_data(schema_file => $invalid_schema, data => $template); };
-like($@, qr{JSON::Validator}, "Invalid schema file");
+throws_ok { my $errors = validate_data(schema_file => $invalid_schema, data => $template); } qr{JSON::Validator}, "Invalid schema file");
```
okurz, line 31: Why is $errors set to the return value but then overwritten in the next statement as well?

okurz (changes requested):

okurz (approved):

t/35-script_clone_job.t
```diff
@@ -155,8 +155,7 @@ subtest 'asset download' => sub {
         qr|downloading|s, 'download not attempted if openQA reports missing assets';
     }
     qr|assets are missing.*iso/foo.iso|s, 'error about missing asset shown';
-    unlike $@, qr|hdd/some.qcow2|, 'no error about qcow2 image which is generated by parent job';
```
okurz, line 158: You shouldn't just delete this test

okurz (changes requested):

t/35-script_clone_job.t
b10n1k: I was waiting to see the test results.  but I was about to ask to review this. So if you say no I will revert it. but then I have another problem which I am not sure how to approach. Any ideas?

b10n1k: would it make sense to negate the throws_ok?

t/35-script_clone_job.t
```diff
@@ -155,7 +155,8 @@ subtest 'asset download' => sub {
...
+    my $ret = $@;
```
perlpunk, line 158: `$ret` is a bit confusing
```suggestion
    my $e = $@;
```

okurz (approved):

PR #6183, opened 2025-02-17
Replace Try::Tiny with Feature::Compat::Try

b10n1k:
In general straight forward refactoring.
However there are some points which needs some additional attention:
- when return try..catch
- the return value (either $_, or other complex structure)
- when return from try block
  
Two modules which has `return` inside _try_ have been left out and
they will be worked in another PR to enphasize to the quality and
proper refactoring.
    

https://progress.opensuse.org/issues/176862

PR #6199, opened 2025-02-21
Use Feature::Compat::Try in JobTemplate controller

perlpunk:
This one is a bit more complicated because it uses several `return` statements, but just for an early return of the control flow. With Try::Tiny it would just leave the `try` block.
In that case we can use an inner extra block and use `last` to leave the `try` without returning from the sub.

Issue: https://progress.opensuse.org/issues/176862

See also #6183

### DIFF
--- a/lib/OpenQA/WebAPI/Controller/API/V1/JobTemplate.pm
+++ b/lib/OpenQA/WebAPI/Controller/API/V1/JobTemplate.pm
@@ -3,7 +3,7 @@
 
 package OpenQA::WebAPI::Controller::API::V1::JobTemplate;
 use Mojo::Base 'Mojolicious::Controller', -signatures;
-use Try::Tiny;
+use Feature::Compat::Try;
 use OpenQA::App;
 use OpenQA::YAML qw(load_yaml dump_yaml);
 use List::Util qw(min);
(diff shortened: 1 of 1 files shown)


Martchus (approved):

perlpunk:
oh no, perltidy doesn't like to nested blocks like that. It will get one level of indentation more.
In that case I can put it on its own line and add some comment; maybe that's a good idea anyway

okurz:
Typo in commit message
"Try::Tny"

perlpunk:
> Due to the increased level of indentation I suggest to extract methods.

Ok. Then let's @b10n1k do that after #6183. This PR was meant for showing how to do early flow control returns with Feature::Compat::Try

kalikiana:
> > Due to the increased level of indentation I suggest to extract methods.
> 
> Ok. Then let's @b10n1k do that after #6183. This PR was meant for showing how to do early flow control returns with Feature::Compat::Try

Did you decide if this should be merged as-is? It looks fine to me, but I figure you might want to coordinate.

perlpunk:
I set it back to draft.
I would expect a refactoring into one or more methods.

lib/OpenQA/WebAPI/Controller/API/V1/JobTemplate.pm
```diff
@@ -247,51 +247,54 @@ sub update ($self) {
...
+            # Backwards compatibility: ID used to mean group ID on this route
+            $json->{id} = $group_id;
+
+            if (my $reference = $validation->param('reference')) {
+                my $template = $job_group->to_yaml;
+                $json->{template} = $template;
+                # Compare with no regard for trailing whitespace
+                chomp $template;
+                chomp $reference;
+                push @$user_errors, 'Template was modified' and last unless $template eq $reference;
```
okurz, line 278: so this is still four levels of indentation which I consider more than desired. I am ok with the change but would still be interested if we can do better. I might try as a "evening fun project" so don't wait for me :)

PR #6201, opened 2025-02-23
openqa-dump-templates: Reproducible output

rclobus:
The output file in JSON format is now reproducible, because all keys will be sorted before the JSON file is written.
The JSON spec states that objects are unordered, but for the purpose of storing the JSON version of the dump in a VCS it helps that the difference between consecutive commits is minimal.

### DIFF
--- a/script/openqa-dump-templates
+++ b/script/openqa-dump-templates
@@ -240,7 +240,7 @@ handle_table($_) for keys %result;
 if ($options{json}) {
     use Mojo::JSON;    # booleans
     use Cpanel::JSON::XS;
-    print Cpanel::JSON::XS->new->ascii->pretty->encode(\%result);
+    print Cpanel::JSON::XS->new->ascii->pretty->canonical(1)->encode(\%result);
 }
 else {
     dd \%result;


Martchus (approved):

perlpunk (approved):

script/openqa-dump-templates
```diff
@@ -240,7 +240,7 @@ handle_table($_) for keys %result;
 if ($options{json}) {
     use Mojo::JSON;    # booleans
     use Cpanel::JSON::XS;
-    print Cpanel::JSON::XS->new->ascii->pretty->encode(\%result);
+    print Cpanel::JSON::XS->new->ascii->pretty->canonical(1)->encode(\%result);
```
perlpunk, line 243: I think this would be enough though:
```suggestion
    print Cpanel::JSON::XS->new->ascii->pretty->canonical->encode(\%result);
```

rclobus: Indeed. How will such cosmetic change be merged (given that this PR has already been merged)?
Will you make the change, should I update this PR, should I create a new PR, ...?

PR #6207, opened 2025-02-24
Reduce resource usage of fullstack scheduler test in CI

Martchus:
This might help with workers being forcefully terminated leading to test failing with:
```
not ok 36 - sub process openqa-worker-unstable terminated with exit code 9
```

Related ticket: https://progress.opensuse.org/issues/177775


okurz (approved):
If it helps the urgent problem go for it but I suggest to keep the changes from the "CI" environment to the manual test environment as little as possible. Otherwise "manual reproduction" is full of bad surprises :)

perlpunk (approved):

PR #6209, opened 2025-02-25
Update documentation about post_run_hook behavior

gauravpathak:
Related to: https://progress.opensuse.org/issues/176823


Martchus (changes requested):
It would be good to state what successful means. I suppose it means that no exception occurs and that the return value is not relevant. Can you brush that and be explicit about it in the documentation?

docs/WritingTests.asciidoc
```diff
@@ -170,7 +170,7 @@ sub post_fail_hook {
 
 ==== `post_run_hook`
 
-It is called after `run` regardless of the result of the test run.
+It is called after `run` only if the `run` subroutine completes successfully, without any failures.
```
okurz, line 173: Did you find out what code change changed behaviour? Is a post_run_hook not intended to be run regardless of the test result?

okurz (changes requested):

docs/WritingTests.asciidoc
Martchus: I guess we should not change the existing behavior as people might rely on it - especially because in practice it probably makes more sense if either the post run hook *or* the post fail hook is executed (but not both).

gauravpathak:
> It would be good to state what successful means. I suppose it means that no exception occurs and that the return value is not relevant. Can you verify that and be explicit about it in the documentation?

@Martchus You are absolutely correct! I verified it by running some more tests, return value from `run` subroutine is not relevant. In case of an exception the `post_run_hook` is skipped.

docs/WritingTests.asciidoc
gauravpathak: > Did you find out what code change changed behaviour? Is a post_run_hook not intended to be run regardless of the test result?

@okurz I ran `git blame` on `basetest.pm`  in os-autoinst but couldn't find any change that should've changed the behavior. I think this is the first time @baierjan noticed and reported this behavior. But I am unsure about it.

okurz (approved):

Martchus (approved):

docs/WritingTests.asciidoc
baierjan: > I guess we should not change the existing behavior as people might rely on it - especially because in practice it probably makes more sense if either the post run hook _or_ the post fail hook is executed (but not both).

afaik you can even achieve both, if you use `die` or fail to assert something inside `post_run_hook` 😉  but that is probably not a practical use.

PR #6210, opened 2025-02-25
Proof of Conect - Link to githubusercontent for image urls

perlpunk:
* This would need mapping for other providers, so the regex should also be configurable
* Avoid doing this if test used the latest sha?

### DIFF
--- a/lib/OpenQA/WebAPI/Controller/File.pm
+++ b/lib/OpenQA/WebAPI/Controller/File.pm
@@ -16,6 +16,10 @@ use Scalar::Util qw(blessed);
 
 has static => sub { Mojolicious::Static->new };
 
+my %urlmap = (
+    'github.com' => 'https://raw.githubusercontent.com/$USER/$REPO/$SHA/$FILE',
+);
+
 sub needle ($self) {
     # do the format splitting ourselves instead of using mojo to restrict the suffixes
     # 13.1.png would be format 1.png otherwise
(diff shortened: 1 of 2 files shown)


perlpunk:
@sdclarke maybe you could try this alternative out as well. this is only for viewing needle matches in the test details steps view. needle candidates and needle editor are not changed at all.

asdil12:
Note how `Mojo::URL` has some issues with parsing `git@host:path/to/repo.git` urls - the lesser used `ssh://git@host/path/to/repo.git` style should be fine, though.

Note how `git@github.com` is detected as a scheme and host is actually empty.
```
perl> $u = Mojo::URL->new('git@github.com:asdil12/opi.git');
git@github.com:asdil12/opi.git
perl> print Dumper($u);
$VAR1 = bless( {
                 'scheme' => 'git@github.com',
                 'query' => bless( {
                                     'charset' => 'UTF-8',
                                     'pairs' => []
                                   }, 'Mojo::Parameters' ),
                 'path' => bless( {
                                    'leading_slash' => !!0,
                                    'trailing_slash' => !!0,
                                    'charset' => 'UTF-8',
                                    'parts' => [
                                                 'asdil12',
                                                 'opi.git'
                                               ]
                                  }, 'Mojo::Path' )
               }, 'Mojo::URL' );
1

perl> $u->path
asdil12/opi.git
perl> $u->host

perl> $u->scheme
git@github.com

Martchus:
In my opinion `Mojo::URL` is parsing `git@github.com:asdil12/opi.git` correctly because when thinking of that blindly as a generic URL it kind of makes sense to consider `git@github.com` the scheme. It looks like the JavaScript URL parser behaves differently and just treats everything as a path - but that is also not really correct.

I guess the problem is not in any of those parsers but that `git@github.com:asdil12/opi.git` is just not really meant to be interpreted as a URL. I guess converting that using a regex makes sense.

PR #6212, opened 2025-02-25
Prevent Unitialized value on is_deadlock

b10n1k:
Avoid having unintialized `$error` value.

### DIFF
--- a/lib/OpenQA/Schema.pm
+++ b/lib/OpenQA/Schema.pm
@@ -151,7 +151,7 @@ sub read_application_secrets ($self) {
     return [map { $_->secret } @secrets];
 }
 
-sub is_deadlock ($self, $error) { $error =~ DEADLOCK_REGEX }
+sub is_deadlock ($self, $error) { defined $error ? $error =~ DEADLOCK_REGEX : undef }
 
 sub txn_do_retry_on_deadlock ($self, $sub, $deadlock_cb = undef) {
     for (my $tries = 0;; ++$tries) {


lib/OpenQA/Schema.pm
```diff
@@ -151,7 +151,7 @@ sub read_application_secrets ($self) {
     return [map { $_->secret } @secrets];
 }
 
-sub is_deadlock ($self, $error) { $error =~ DEADLOCK_REGEX }
+sub is_deadlock ($self, $error) { defined $error ? $error =~ DEADLOCK_REGEX : undef }
```
perlpunk, line 154: This is not necessary. `is_deadlock` is called from here with `$_` which contains the exception:
https://github.com/os-autoinst/openQA/blob/master/lib/OpenQA/Schema/Result/ScheduledProducts.pm#L275
```
            die $_ if $schema->is_deadlock($_);    # uncoverable statement
```
`$_` is always defined if the `catch` blocked is called. So we don't need to check for definedness.

perlpunk (changes requested):

lib/OpenQA/Schema.pm
perlpunk: And if this should be really useful, it should rather die, making the user aware that the code is passing an undefined variable. Otherwise it will just be silently returning `undef`, and possibly noone will ever know that they are passing the wrong variable.

perlpunk: And think of the other perspective - we only saw your mistake of replacing `$_` with `$e` here only once because we got a warning.
If we had this PR here merged before, we would have never noticed that mistake

perlpunk: If some coder adds a call to that function with a variable that is undef, it is very probably a mistake.

Martchus (changes requested):
That the error is not undef is a precondition of this function. I suggest you rather fix the call site.

lib/OpenQA/Schema.pm
okurz: So ... how do you stand regarding making warnings fatal?

perlpunk:
> That the error is not undef is a precondition of this function. I suggest you rather fix the call site.

That's not completely true. It only ensures that an argument is passed, but the value can be undef.

Martchus:
I meant a precondition in the sense that calling functions have to pass something that is not `undef`. Of course this is not enforced by the signature but merely how I intended this function to be called.

lib/OpenQA/Schema.pm
Martchus: That would be a bad idea. Making warnings fatal makes only sense for build-time warnings of development builds or maybe at runtime during development.

Making the warnings we are talking about here fatal in general would lead to many unnecessary troubles, e.g. one can not have deprecation periods (emitting a warning about something being deprecated) without causing immediate breakage.

okurz: well, fine. We already treat warnings as fatal during tests anyway and within o3 we are alerted about warnings reported in the logfiles so we should be good. With that I suggest @b10n1k you close the PR again

b10n1k:
I guess it doesnt matter at that point

PR #6216, opened 2025-02-26
Document traceability and reproducibility of tests

Martchus:
* Explain relevant variables
* Improve formatting of list in section about logging package versions
* Mention the behavior of the restart button with regard to reproducibility
* Mention `openqa-investigate`
* See https://progress.opensuse.org/issues/173878

### DIFF
--- a/docs/WritingTests.asciidoc
+++ b/docs/WritingTests.asciidoc
@@ -474,9 +474,34 @@ sub run {
 }
 --------------------------------------------------------------
 
-=== Logging package versions used for test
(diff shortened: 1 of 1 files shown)


okurz (changes requested):
one minor remark. Rest is perfect

EDIT: One more thought about additions. How about hints how to make a test fully reproducible, e.g. avoid relying on any external resource like online repositories not controlled by openQA test variables

okurz (approved):

perlpunk (approved):

PR #6217, opened 2025-02-26
Bidirectional links for /admin/productlog

rclobus:
Additionally to the JSON pre-formatted dump, the hyperlinks to the tests are shown. This allows for a bi-directional linking from job to product_log.

### DIFF
--- a/assets/javascripts/audit_log.js
+++ b/assets/javascripts/audit_log.js
@@ -310,6 +310,53 @@ function loadProductLogTable(dataTableUrl, rescheduleUrlTemplate, showActions) {
       },
       {targets: 2, orderable: false},
       {
+        targets: 6,
+        render: function (data, type, row) {
+          const rowData = dataForLink(this);
+          const settings = rowData.settings;
+          const version = settings.VERSION;
+          const build = settings.BUILD;
+          const distri = settings.DISTRI;
+          if (version !== undefined && build !== undefined && distri !== undefined) {
+            const link = urlWithBase(
+              '/tests/overview?build=' +
+                encodeURIComponent(build) +
+                '&distri=' +
+                encodeURIComponent(distri) +
+                '&version=' +
+                encodeURIComponent(version) +
+                '&flavor=' +
+                encodeURIComponent(data)
+            );
+            return `<a href='${link}'>${data}</a>`;
+          } else {
+            return data;
+          }
+        }
+      },
+      {
+        targets: 8,
+        render: function (data, type, row) {
+          const rowData = dataForLink(this);
+          const settings = rowData.settings;
+          const version = settings.VERSION;
+          const distri = settings.DISTRI;
+          if (version !== undefined && distri !== undefined) {
+            const link = urlWithBase(
+              '/tests/overview?build=' +
+                encodeURIComponent(data) +
+                '&distri=' +
+                encodeURIComponent(distri) +
+                '&version=' +
+                encodeURIComponent(version)
+            );
+            return `<a href='${link}'>${data}</a>`;
+          } else {
+            return data;
+          }
+        }
+      },
+      {
         targets: 10,
         orderable: false,
         render: function (data, type, row) {


rclobus:
I apparently can't paste an image here, so try this URL: https://pioneers.game-host.org/Sample.png

okurz:
ok, nice start. But as your example image already shows: Is it really that useful to show links to tens or even hundreds to thousand numbers? Why not just link to the according /tests/overview page for the triggered build?

assets/javascripts/audit_log.js
```diff
@@ -210,8 +210,26 @@ function showScheduledProductSettings(link) {
...
-  element.textContent = results ? JSON.stringify(results, undefined, 4) : 'No results available.';
+  const element = document.createElement('p');
+  if (results) {
+    const jsondump = document.createElement('pre');
+    jsondump.textContent = JSON.stringify(results, undefined, 4);
+    element.append(jsondump);
+    if (results.successful_job_ids) {
+      const urllist = document.createElement('p');
+      element.append('Successful job ids: ');
+      for (i = 0; i < results.successful_job_ids.length; i++) {
```
Martchus, line 221: Please avoid using global variables. There are also better ways of iterating an array than using an index variable (even in JavaScript).

Martchus (changes requested):
The commit message title should state what feature is added and not just link to an issue. (The commit message body should of course still reference to the issue.)

assets/javascripts/audit_log.js
kalikiana: I'm kinda surprised the "global" i is not caught in a style check.

Martchus:
> Perhaps instead of modifying the /admin/productlog page, the line for 'Scheduled product' could point to the URL with distri, flavor, version and build.

But please only in *addition*. It is often very useful to know what exact scheduled product is responsible for a certain job so the link to the product log is still needed. Note that there is already a similar link so this would also be a bit redundant, e.g. on https://openqa.opensuse.org/tests/4887814 one can access the relevant test results overview page by clicking on "Job Groups" -> "Build 20250226".

rclobus:
In the second commit, I am adding hyperlinks in the table. This is actually more to my taste (instead of the code from the first commit).

Note that I marked it as draft, because I think there would need to be some hardening, i.e. when one of the fields is somehow undefined, the 'a href' tag should not be generated. However, since this is a draft, we can further discuss, whether this approach looks OK to the openQA team.

From the table, I can jump back to the build of the product, and even to a more finely filtered view, which shows only the flavor.

rclobus:
I've added some hardening, the hyperlink will only show up if the data is present.
The hyperlinks show up in the table and allows one to jump back to the same build or the same build+flavor (the latter is quiet useful e.g. for Debian, where many flavors share the same build number: https://openqa.debian.net/tests/overview?distri=debian-live&version=sid&build=20250313&groupid=17)

assets/javascripts/audit_log.js
rclobus: I've removed this part of the PR.

rclobus:
Friendly ping. Is there still something to do from my side?

assets/javascripts/audit_log.js
```diff
@@ -309,6 +309,35 @@ function loadProductLogTable(dataTableUrl, rescheduleUrlTemplate, showActions) {
...
+      {
+        targets: 6,
+        render: function (data, type, row) {
+          const rowData = dataForLink(this);
+          const settings = rowData.settings;
+          const version = settings['VERSION'];
+          const build = settings['BUILD'];
+          const distri = settings['DISTRI'];
+          if (version !== undefined && build !== undefined && distri !== undefined) {
+            return `<a href=/tests/overview?version=${version}&build=${build}&distri=${distri}&flavor=${data}>${data}</a>`;
```
Martchus, line 321: The attribute value should be quoted. I'm also wondering whether we need to do proper escaping here. (It is very annoying that one cannot just return a DOM element here.)

The same counts for the other occurrence.

assets/javascripts/audit_log.js
```diff
@@ -309,6 +309,35 @@ function loadProductLogTable(dataTableUrl, rescheduleUrlTemplate, showActions) {
...
       },
       {targets: 2, orderable: false},
+      {
+        targets: 6,
+        render: function (data, type, row) {
+          const rowData = dataForLink(this);
+          const settings = rowData.settings;
+          const version = settings['VERSION'];
+          const build = settings['BUILD'];
+          const distri = settings['DISTRI'];
```
Martchus, line 319:
```suggestion
          const version = settings.VERSION;
          const build = settings.BUILD;
          const distri = settings.DISTRI;
```
The same counts the the other definition.

Martchus (changes requested):
I have a few remarks regarding the code. I also still need to test it on my machine which I'll do tomorrow.

okurz:
@rclobus please rebase to prevent the merge in your feature branch

rclobus:
The hyperlinks are showing up nicely now.

However, upon further testing (for the version without parameters http://localhost/admin/productlog), the hyperlinks are not correct. All additional parameters in the newly created hyperlink point to the values from the first row, except for the value matching the column.
For the typical scenario where only a single entry is shown in the table, everything works as intended.

So it needs a final bit of tweaking:
* Either the hyperlink will be added when there is one row of data
* Or I'll need your help to point me to the right data source for the row (or you'll implement the final bit)

See this screenshot: the mouse is hovering over entry #205:
image

Martchus (approved):

Martchus:
I am approving because this is already an improvement.

Note that I personally found the first approach much more useful as it didn't just link to a test overview page but rendered the links of the actual jobs that were created by the scheduled product. The test results overview might show something different entirely or some other additional jobs that were scheduled in a different way.

PR #6225, opened 2025-02-27
Allow `WORKER_CLASS`-specific settings in worker config

Martchus:
See https://progress.opensuse.org/issues/169492


t/data/24-worker-settings/workers.ini
```diff
@@ -21,3 +21,9 @@ HOST_SPECIFIC = setting (localhost)
 
 [https://remotehost]
 HOST_SPECIFIC = specific setting (remotehost)
+
+[class:special-hardware]
+FOO = not used because setting from flot has precedence
```
nicksinger, line 26: s/flot/slot

okurz (approved):

docs/WritingTests.asciidoc
```diff
@@ -588,6 +588,9 @@ WORKER_CLASS = desktop,foo,bar
 
 [3]
 # WORKER_CLASS is not set
+
+[class:foo]
+FOO = this value is present on all instances where WORKER_CLASS contains foo
```
kalikiana, line 593:
```suggestion
# will only be used if there's no slot-specific override for FOO
FOO = this value is present on all instances where WORKER_CLASS contains foo
```

b10n1k: Actually I think it is better if the whole section could improve here. But maybe in another PR to not block this

lib/OpenQA/Worker/Settings.pm
```diff
@@ -81,6 +82,10 @@ sub _read_section ($cfg, $section, $out) {
     $out->{uc $_} = trim $cfg->val($section, $_) for $cfg->Parameters($section);
 }
 
+sub _read_section_keeping_default ($cfg, $section, $out) {
+    $out->{uc $_} //= trim $cfg->val($section, $_) for $cfg->Parameters($section);
```
b10n1k, line 86: Could we change the name of the `$out` param?  why not `$settings`?

b10n1k (changes requested):

lib/OpenQA/Worker/Settings.pm
okurz: I think `$out` is better as that shows it's the resulting outgoing value. `$setting` would be ambiguous as that could be the unparsed value

docs/WritingTests.asciidoc
Martchus: Yes, I added a small paragraph with an additional explanation and extended the example.

lib/OpenQA/Worker/Settings.pm
```diff
@@ -31,14 +31,12 @@ sub new ($class, $instance_number = undef, $cli_options = {}) {
...
-    for my $section ('global', $instance_number) {
-        if ($cfg && $cfg->SectionExists($section)) {
-            for my $set ($cfg->Parameters($section)) {
-                $global_settings{uc $set} = trim $cfg->val($section, $set);
-            }
-        }
+    if ($cfg) {
+        _read_section($cfg, 'global', \%global_settings);
+        _read_section($cfg, $instance_number, \%global_settings);
+        _read_section($cfg, "class:$_", \%global_settings) for split(',', $global_settings{WORKER_CLASS} // '');
```
perlpunk: Note that [Config::IniFiles](https://metacpan.org/pod/Config::IniFiles#Groups) implements subsections / Groups with a space, so it is a builtin feature. Using a colon is not wrong per se, as the ini format doesn't specify that feature, but maybe it would be nice to use the format that is supported by the module we use.

b10n1k (approved):

lib/OpenQA/Worker/Settings.pm
perlpunk: I guess I should have requested changes here to at least discuss it...

Martchus: I just went for the first idea that was mentioned on the ticket (probably by myself). I guess it doesn't really matter, though. All we need to achieve "grouping" here is to concatenate a string - and it really didn't come to my mind to check for a library feature to help with such a basic thing.

perlpunk: We had the same discussion before though for a different feature where I mentioned the group feature.

PR #6227, opened 2025-02-28
Add softfail to XUnit parser

pdostal:
Hello,

we use `XUnit` parser to parse `pytest` output and we would like openQA to softfail in situations the `pytest` testcase encounters known bug (which is documented and referenced).
Currently those testcases are reported as `skipped` which makes them less visible than `softfailures`.

- Related ticket: [poo#177321](https://progress.opensuse.org/issues/177321)
- Related pull request: os-autoinst/os-autoinst-distri-opensuse#21304

### DIFF
--- a/lib/OpenQA/Parser/Format/XUnit.pm
+++ b/lib/OpenQA/Parser/Format/XUnit.pm
@@ -11,6 +11,8 @@ sub _add_single_result { shift->results->add(OpenQA::Parser::Result::XUnit->new(
 
 sub addproperty { shift->{properties}->add(OpenQA::Parser::Result::XUnit::Property->new(shift)) }
 
+my %TC_RESULT_BY_TAG = (softfailure => 'softfail', failure => 'fail', error => 'fail');
+
 sub parse {
     my ($self, $xml) = @_;
     confess 'No XML given/loaded' unless $xml;
@@ -30,6 +32,7 @@ sub parse {
 
         $result->{errors} = exists $ts->{errors} ? $ts->{errors} : undef;
         $result->{tests} = exists $ts->{tests} ? $ts->{tests} : undef;
+        $result->{softfailures} = exists $ts->{softfailures} ? $ts->{softfailures} : undef;
         $result->{failures} = exists $ts->{failures} ? $ts->{failures} : undef;
         $result->{time} = exists $ts->{time} ? $ts->{time} : undef;
 
@@ -56,6 +59,7 @@ sub parse {
             });
 
         my $ts_result = 'ok';
+        $ts_result = 'softfail' if ($ts->{softfailures} && $ts->{softfailures} > 0);
         $ts_result = 'fail' if ($ts->{failures} && $ts->{failures} > 0) || ($ts->{errors} && $ts->{errors} > 0);
         $result->{result} = $ts_result;
         $result->{dents} = 0;
@@ -67,6 +71,7 @@ sub parse {
             sub {
                 my $tc = shift;
                 my $tc_result = 'ok';
+                $tc_result = 'softfail' if ($tc->{softfailures} && $tc->{softfailures} > 0);
                 $tc_result = 'fail'
                   if ($tc->{failures} && $tc->{failures} > 0) || ($tc->{errors} && $tc->{errors} > 0);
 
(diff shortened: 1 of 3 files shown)


asdil12:
Could you maybe give some example how this would be used?
Like the changes when calling XUnit from the test module code or what the different in the to be parsed unit test output file would be - essentially how does XUnit get to know what should softfail?

Martchus (changes requested):
Wouldn't it make more sense to assign `softfail` before assigning `fail` so the worse result "wins" in case there are failures and softfailures at the same time?

It would also make sense to extend unit tests, e.g. the xunit-related subtest in `t/30-test_parser.t`.

okurz (changes requested):
The commit message should have a bit more details, e.g. why this needs to change, link to a ticket, the alternatives you considered, reasoning, etc. Adding "softfail" might sound like a small thing but it can have significant impact on consequences which not all been foreseen.

Also can you comment how that relates to the handling of softfailures in lib/OpenQA/Parser/Format/JUnit.pm ?

pdostal:
@asdil12 please see the modifications in `t/data/xunit_format_example.xml` file. Currently the test cases can be marked as `skipped, passed, error, failure` and I add `softfailure` as another option.

pdostal:
@okurz please check the commit message now.

> Also can you comment how that relates to the handling of softfailures in lib/OpenQA/Parser/Format/JUnit.pm ?

It has no relations as far as I know.

lib/OpenQA/Parser/Format/XUnit.pm
```diff
@@ -76,7 +79,8 @@ sub parse {
                 my $content = '# Test messages ';
                 $content .= "# $tc->{name}\n" if $tc->{name};
 
-                for my $out ($tc->children('skipped, passed, error, failure')->each) {
+                for my $out ($tc->children('skipped, passed, error, failure, softfailure')->each) {
+                    $tc_result = 'softfail' if ($out->tag =~ m/softfailure/);
                     $tc_result = 'fail' if ($out->tag =~ m/failure|error/);
```
Martchus, line 84: The tag `softfailure` contains `failure`. So won't this assign `fail` even if just a `softfailure` tag is present due to the lack of anchoring in this regex?

I would at some point avoid the repetitive use of `$out->tag` in this loop and store that value a variable. Then one could also use interpolation in the line below.

lib/OpenQA/Parser/Format/XUnit.pm
```diff
@@ -76,7 +79,8 @@ sub parse {
                 my $content = '# Test messages ';
                 $content .= "# $tc->{name}\n" if $tc->{name};
 
-                for my $out ($tc->children('skipped, passed, error, failure')->each) {
+                for my $out ($tc->children('skipped, passed, error, failure, softfailure')->each) {
+                    $tc_result = 'softfail' if ($out->tag =~ m/softfailure/);
```
Martchus, line 83: There's probably no need to use a regex here:
```suggestion
                    $tc_result = 'softfail' if $out->tag eq 'softfailure';
```
See also my other comment for the use of `$out->tag`.

Martchus (changes requested):

okurz:
+1

okurz (approved):

lib/OpenQA/Parser/Format/XUnit.pm
```diff
@@ -76,8 +79,9 @@ sub parse {
                 my $content = '# Test messages ';
                 $content .= "# $tc->{name}\n" if $tc->{name};
 
-                for my $out ($tc->children('skipped, passed, error, failure')->each) {
-                    $tc_result = 'fail' if ($out->tag =~ m/failure|error/);
+                my %TC_RESULT_BY_TAG = (softfailure => 'softfail', failure => 'fail', error => 'fail');
```
Martchus, line 82: Please move this outside of the function. There's no need to initialize this hash each and every time this function is invoked. It would make sense to move it even outside the whole function.

Martchus (changes requested):
I have just one minor nitpick. Otherwise it looks like this will work now.

lib/OpenQA/Parser/Format/XUnit.pm
```diff
@@ -19,6 +19,7 @@ sub parse {
     my @tests;
     my %t_names;
     my $i = 1;
+    my %TC_RESULT_BY_TAG = (softfailure => 'softfail', failure => 'fail', error => 'fail');
```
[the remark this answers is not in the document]
pdostal: Ah, you mean outside the `run` subroutine, I thought to just move it outside the `for` loop. Now it's done.

pdostal:
Here is [verification run](https://pdostal-server.suse.cz/tests/8524).

Martchus (approved):

pdostal:
Can I ask you when should I expect this in openqa.suse.de?

Martchus:
I think it has already been deployed, see https://mailman.suse.de/mlarch/SuSE/openqa/2025/openqa.2025.03/msg00002.html.

alvarocarvajald:
For reference, go to https://openqa.suse.de/tests/16818158#external (ran 17 days ago) and click on `Show only failures`. You'll see the same failures marked as failed.

PR #6228, opened 2025-02-28
Replace Try::Tiny from remaining OpenQA modules

b10n1k:

### DIFF
--- a/lib/OpenQA/WebSockets/Controller/Worker.pm
+++ b/lib/OpenQA/WebSockets/Controller/Worker.pm
@@ -13,7 +13,7 @@ use OpenQA::Jobs::Constants;
 use OpenQA::Scheduler::Client;
 use DateTime;
 use Data::Dump 'pp';
-use Try::Tiny;
+use Feature::Compat::Try;
 use Mojo::Util 'dumper';
 
 use constant LOG_WORKER_STATUS_MESSAGES => $ENV{OPENQA_LOG_WORKER_STATUS_MESSAGES} // 0;
@@ -105,10 +105,10 @@ sub _message ($self, $json) {
                     $_->reschedule_state for @jobs;
                 });
         }
-        catch {
+        catch ($e) {
             # uncoverable statement
-            log_warning("Unable to re-schedule job(s) $job_ids_str rejected by worker $worker_id: $_");
-        };
+            log_warning("Unable to re-schedule job(s) $job_ids_str rejected by worker $worker_id: $e");
+        }
 
         # log that we 'saw' the worker
         $worker_db->seen;
@@ -167,9 +167,9 @@ sub _message ($self, $json) {
             # Tell the worker that we saw it (used for tests and debugging)
             $tx->send({json => {type => 'info', seen => 1}});
         }
-        catch {
-            log_error("Failed updating seen and error status of worker $worker_id: $_");    # uncoverable statement
-        };
+        catch ($e) {
+            log_error("Failed updating seen and error status of worker $worker_id: $e");    # uncoverable statement
+        }
 
         # find the job currently associated with that worker and check whether the worker still
         # executes the job it is supposed to
@@ -206,11 +206,10 @@ sub _message ($self, $json) {
                 $worker->reschedule_assigned_jobs([$current_job, @unfinished_jobs]);
             }
         }
-        catch {
+        catch ($e) {
             # uncoverable statement
-            log_warning("Unable to verify whether worker $worker_id runs its job(s) as expected: $_");
-        };
-
+            log_warning("Unable to verify whether worker $worker_id runs its job(s) as expected: $e");
+        }
         # consider the worker idle unless it claims to be broken or work on a job
         $worker_status->{idle_despite_job_assignment} = !$worker_is_broken && !defined $job_id;
     }
--- a/lib/OpenQA/WebAPI/Controller/API/V1/JobTemplate.pm
+++ b/lib/OpenQA/WebAPI/Controller/API/V1/JobTemplate.pm
@@ -3,7 +3,7 @@
 
 package OpenQA::WebAPI::Controller::API::V1::JobTemplate;
 use Mojo::Base 'Mojolicious::Controller', -signatures;
-use Try::Tiny;
+use Feature::Compat::Try;
 use OpenQA::App;
 use OpenQA::YAML qw(load_yaml dump_yaml);
 use List::Util qw(min);
@@ -251,10 +251,10 @@ sub update ($self) {
         $user_errors
           = $self->app->validate_yaml($data, $validation->param('schema'), $self->app->log->level eq 'debug');
     }
-    catch {
+    catch ($e) {
         # Push the exception to the list of errors without the trailing new line
-        push @$user_errors, substr($_, 0, -1);
-    };
+        push @$user_errors, substr($e, 0, -1);
+    }
     return $self->respond_to(json => {json => {error => $user_errors}, status => 400}) if @$user_errors;
 
     my $json = {};
@@ -262,12 +262,12 @@ sub update ($self) {
     try {
         $self->_perform_update($id, $name, $json, $data, $yaml, $template_reference, $to_expand, $user_errors);
     }
-    catch {
+    catch ($e) {
         # Push the exception to the list of errors without the trailing new line
-        my $error = substr($_, 0, -1);
+        my $error = substr($e, 0, -1);
         my $error_type = ($error =~ qr/unique constraint/) ? $user_errors : \@server_errors;
         push @$error_type, $error unless $error eq 'abort transaction';
-    };
+    }
 
     if (@server_errors) {
         push @$user_errors, 'Internal server error occurred';


lib/OpenQA/WebSockets/Controller/Worker.pm
```diff
@@ -172,62 +168,25 @@ sub _message {
...
-              && (scalar @unfinished_jobs <= 1);
-
-            # give worker a second chance to process the job assignment
-            # possible situation on the worker: The worker might be sending a status update claiming it is
-            # idle (or has doing that task piled up on the event loop). At the same time a job arrives. The
-            # message regarding that job will be processed after sending the idle status. So let's give the
-            # worker another change to process the message about its assigned job.
-            return undef unless $worker_previously_idle;
-
+            my ($current_job, $unfinished_jobs)
```
perlpunk, line 180: At the first glance that method is not returning those two values. Have to have a closer look

perlpunk (dismissed):

lib/OpenQA/WebSockets/Controller/Worker.pm
b10n1k: true. I just rushed. moment. I will mark `not-ready` and look at it after meeting

perlpunk: Previously the code would not call `reschedule_assigned_jobs` in certain cases. Now you call `_check_assignment_process` but you don't do anything depending on what the method returns, so `reschedule_assigned_jobs`  is always called.

lib/OpenQA/WebSockets/Controller/Worker.pm
```diff
@@ -238,4 +197,48 @@ sub _message {
...
+        && OpenQA::Jobs::Constants::meta_state($current_job_state) eq OpenQA::Jobs::Constants::EXECUTION)
+      && (scalar @unfinished_jobs <= 1);
+
+    # give worker a second chance to process the job assignment
+    # possible situation on the worker: The worker might be sending a status update claiming it is
+    # idle (or has doing that task piled up on the event loop). At the same time a job arrives. The
+    # message regarding that job will be processed after sending the idle status. So let's give the
+    # worker another change to process the message about its assigned job.
+    return undef unless $worker_previously_idle;
+    return (($current_job, \@unfinished_jobs));
```
Martchus, line 241:
```suggestion
    return ($current_job, \@unfinished_jobs);
```

lib/OpenQA/WebSockets/Controller/Worker.pm
```diff
@@ -238,4 +197,48 @@ sub _message {
     }
 }
 
+sub _check_assignment_process($self, $worker, $current_job_id, $worker_id, $job_id, $job_token, $worker_previously_idle)
```
Martchus, line 200:
```suggestion
sub _should_assign_jobs ($self, $worker, $current_job_id, $worker_id, $job_id, $job_token, $worker_previously_idle)
```

lib/OpenQA/WebSockets/Controller/Worker.pm
```diff
@@ -172,62 +168,25 @@ sub _message {
...
-            # message regarding that job will be processed after sending the idle status. So let's give the
-            # worker another change to process the message about its assigned job.
-            return undef unless $worker_previously_idle;
-
+            my ($current_job, $unfinished_jobs)
+              = $self->_check_assignment_process($worker, $current_job_id, $worker_id, $job_id, $job_token,
+                $worker_previously_idle);
             log_debug("Rescheduling jobs assigned to worker $worker_id");
-            $worker->reschedule_assigned_jobs([$current_job, @unfinished_jobs]);
+            $worker->reschedule_assigned_jobs([$current_job, @$unfinished_jobs]);
```
Martchus, line 184: This should only be called conditionally depending on the outcome of the code you have extracted.

Martchus (changes requested):

okurz:
Due to the multiple comments I suggest to create a separate PR for just https://github.com/os-autoinst/openQA/pull/6228/commits/979e2d11796534a5a533bb101664d8b090a684ac which should be mergeable easily

lib/OpenQA/WebSockets/Controller/Worker.pm
b10n1k: I think this is solved as part of https://github.com/os-autoinst/openQA/pull/6228#discussion_r1975244731. Can you check again?

lib/OpenQA/WebSockets/Controller/Worker.pm
```diff
@@ -172,62 +168,28 @@ sub _message {
...
-            my $current_job = $worker->job // $unfinished_jobs[0];
-            if ($current_job) {
-                $current_job_id = $current_job->id;
-                $current_job_state = $current_job->state;
+            my ($current_job, $unfinished_jobs)
+              = $self->_should_assign_jobs($worker, $current_job_id, $worker_id, $job_id, $job_token,
+					   $worker_previously_idle);
+	    log_debug("Rescheduling jobs assigned to worker $worker_id");
+                
+            if (defined $unfinished_jobs) {
```
Martchus, line 185:
```suggestion
            if (defined $unfinished_jobs) {
                log_debug("Rescheduling jobs assigned to worker $worker_id");
```

Martchus (changes requested):

lib/OpenQA/WebSockets/Controller/Worker.pm
b10n1k: I tried this because of the CI. which I couldnt reproduce local. I will revert

b10n1k:
> Due to the multiple comments I suggest to create a separate PR for just [979e2d1](https://github.com/os-autoinst/openQA/commit/979e2d11796534a5a533bb101664d8b090a684ac) which should be mergeable easily

https://github.com/os-autoinst/openQA/pull/6233
I thought I could resolve the problem with the CI faster and due to the fast reaction to the comments. but I had no luck

lib/OpenQA/WebSockets/Controller/Worker.pm
```diff
@@ -234,4 +198,47 @@ sub _message ($self, $json) {
     }
 }
 
+sub _should_assign_jobs ($self, $worker, $worker_id, $current_job_id, $job_id, $job_token, $worker_previously_idle) {
+    return undef unless $worker;
```
b10n1k, line 202: I think I can remove this due to https://github.com/os-autoinst/openQA/pull/6228/files#diff-51176bd675ee4037fef74049369705b9259134ad87070a88e59501936761fee1R163. What do you think?

b10n1k: or even refactor further this part and combine it with the try block of that reference

b10n1k:
> See #6238 for an alternative. I think most of the return statements in the try are actually intended to return from the sub.

is this as a reply to https://github.com/os-autoinst/openQA/pull/6228#discussion_r1976362817? 
Maybe it better to close this PR first and proceed to close the task on the scope of the ticket and then improve those concerns.

okurz:
I already approved https://github.com/os-autoinst/openQA/pull/6238 so if you approve as well you can just rebase on top of it. However other files are not affected so you can go ahead and change all files and bring in the style check

b10n1k:
> I already approved #6238 so if you approve as well you can just rebase on top of it. However other files are not affected so you can go ahead and change all files and bring in the style check

Why? this is out of scope of the ticket and if I accept it feels another iterations of changes and reviews? When I comment Ionly wanted to bring it up to attention. Can you finish with one task before force further changes?

perlpunk:
> Maybe it better to close this PR first and proceed to close the task on the scope of the ticket and then improve those concerns. 

Well, I see mx PR in scope of the ticket.
The worker_status code is a bit complicated, and I am even not 100% sure myself if I didn't break anything with my PR.
your suggestions for further changes seemed to go in the wrong direction.
now you can concentrate on the try/catch only.

lib/OpenQA/WebSockets/Controller/Worker.pm
```diff
@@ -234,4 +198,47 @@ sub _message ($self, $json) {
...
+        && OpenQA::Jobs::Constants::meta_state($current_job_state) eq OpenQA::Jobs::Constants::EXECUTION)
+      && (scalar @unfinished_jobs <= 1);
+
+    # give worker a second chance to process the job assignment
+    # possible situation on the worker: The worker might be sending a status update claiming it is
+    # idle (or has doing that task piled up on the event loop). At the same time a job arrives. The
+    # message regarding that job will be processed after sending the idle status. So let's give the
+    # worker another change to process the message about its assigned job.
+    return undef unless $worker_previously_idle;
+    return ($current_job, $job_id, \@unfinished_jobs);
```
perlpunk, line 241: Why are you returning `$job_id` here? it's not changed in this function.

lib/OpenQA/WebSockets/Controller/Worker.pm
perlpunk: Why do you think it can be removed due to the line you linked to?
That's a 
```
if ($worker) {
...
}
```
block above

okurz (approved):
LGTM. There is at least one point that @perlpunk brought up and already requested changes on. Assuming that any related changes would be trivial we can keep my approval.

lib/OpenQA/WebSockets/Controller/Worker.pm
b10n1k: the idea was that the there is an earlier check of the $worker. The simplest move will be to move the code inside this `if` block. or as you did I think in your PR return earlier.

perlpunk: And combining the two try blocks to one would also not work, as they are catching two different problems.

okurz:
#6238 is merged so please rebase on top. @b10n1k I am sure @perlpunk is happy to help to resolve any non-trivial conflicts

lib/OpenQA/WebSockets/Controller/Worker.pm
b10n1k: can this resolved as the https://github.com/os-autoinst/openQA/pull/6238 merged and it might be different situation now

perlpunk: Well, sure. You opened this, but I' can resolve it as well

lib/OpenQA/WebSockets/Controller/Worker.pm
```diff
@@ -167,8 +167,8 @@ sub _message ($self, $json) {
             # Tell the worker that we saw it (used for tests and debugging)
             $tx->send({json => {type => 'info', seen => 1}});
         }
-        catch {
-            log_error("Failed updating seen and error status of worker $worker_id: $_");    # uncoverable statement
+        catch ($e) {
+            log_error("Failed updating seen and error status of worker $worker_id: $e");    # uncoverable statement
         };
```
perlpunk, line 172:
```suggestion
        }
```

perlpunk: Other than that, LGTM

Martchus (approved):

PR #6235, opened 2025-02-28
Add signatures in OpenQA::Utils and others

okurz:


lib/OpenQA/Utils.pm
```diff
@@ -361,33 +342,20 @@ sub run_cmd_with_log_return_error ($cmd, %args) {
...
-    if ($setting eq 'UEFI_PFLASH_VARS' && $value !~ m,^/,) {
-        return 'hdd';
-    }
-    if ($setting =~ /^REPO_\d+$/) {
-        return 'repo';
-    }
-    if ($setting =~ /^ASSET_\d+$/ || $setting eq 'KERNEL' || $setting eq 'INITRD') {
-        return 'other';
-    }
+    return 'hdd' if $setting eq 'UEFI_PFLASH_VARS' && $value !~ m,^/,;
```
b10n1k, line 351: This seems it could be one line. But not big problem

b10n1k:
```suggestion
    # non-absolute-path value of UEFI_PFLASH_VARS treated as HDD asset
    return 'hdd' if $setting =~ /^HDD_\d+$/ || ($setting eq 'UEFI_PFLASH_VARS' && $value !~ m,^/,);
```

b10n1k, line 351:
```suggestion
    # non-absolute-path value of UEFI_PFLASH_VARS treated as HDD asset
    return 'hdd' if $setting =~ /^HDD_\d+$/ || ($setting eq 'UEFI_PFLASH_VARS' && $value !~ m,^/,);
```
The preview doesnt include the line 349 above but this is what I meant

okurz: then the comment wouldn't comment the right statement anymore. I prefer to keep the old style due to the comment

lib/OpenQA/Utils.pm
```diff
@@ -425,28 +393,18 @@ sub find_labels {
     return \@labels;
 }
 
-sub find_flags ($text) {
```
perlpunk, line 428: Did you delete this function on purpose?

lib/OpenQA/Utils.pm
```diff
@@ -425,28 +393,18 @@ sub find_labels {
...
-    my ($text) = @_;
-    $text //= '';
+sub find_bugref ($text = '') {
     $text =~ BUGREF_REGEX;
     return $+{match};
 }
 
-sub find_bugrefs {
-    my $text = shift // '';
+sub find_bugrefs ($text = '') {
```
perlpunk, line 401: That would allow calling `find_bugrefs` without an argument. That doesn't make sense to me.

lib/OpenQA/Utils.pm
okurz: oops. That seems to be a rebase problem. Thx, fixed

lib/OpenQA/Utils.pm
```diff
@@ -865,16 +794,11 @@ sub service_port {
     return $base + $offsets->{$service};
 }
 
-sub random_string {
-    my ($length, $chars) = @_;
-    $length //= 16;
-    $chars //= ['a' .. 'z', 'A' .. 'Z', '0' .. '9', '_'];
+sub random_string ($length = 16, $chars = ['a' .. 'z', 'A' .. 'Z', '0' .. '9', '_']) {
```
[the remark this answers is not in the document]
okurz: Sure, why not. But that is not related to the change of this PR which focuses on adding signatures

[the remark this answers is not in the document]
b10n1k: feel free to do in a separate commit or PR. and you can resolve this thread if you wish

lib/OpenQA/Utils.pm
Wabri: even before the subroutine could be called without arguments

perlpunk: Yes, because we didn't use signatures, so there was no check. But now with signatures we can check that, so calls without arguments will get a clear error message. That's one of the advantages of signatures.

lib/OpenQA/WebAPI/Controller/Admin/Needle.pm
```diff
@@ -26,8 +26,8 @@ sub _translate_date_format ($datetime) {
 sub _translate_cond ($cond) {
     my ($operator, @additional_conds) = ($cond =~ m/^min/) ? ('>=') : ('<', {'=' => undef});
     my $translated;
-    if ($cond =~ m/^(min|max)(\d+)$/) {
-        $translated = _translate_days($2);
+    if ($cond =~ m/^min(\d+)$/) {
+        return {'>=' => _translate_days($1)};
```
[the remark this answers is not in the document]
okurz: yeah, dosen't make sense to me. Excluded. Just focusing  on [lib/OpenQA/Utils.pm](https://github.com/os-autoinst/openQA/pull/6235/files#diff-22808b520763018397277d64e2a9af02da10fcf85f481e9c2b559033d4e0658b) which has enough comments :)

lib/OpenQA/Utils.pm
okurz: so previously when code would have called `find_bugrefs $undefined_object` that would just return falso from the fanction as there is no match. How can we keep that functionality? Do I need to do

```perl
sub find_bugrefs ($text = undef) {
    $text //= '';
```

lib/OpenQA/Utils.pm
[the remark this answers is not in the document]
okurz: Guys I don't know what you mean. @b10n1k you said "static variable". Do you mean a constant for the default value like

```perl
use constant RANDOM_STRING_DEFAULT_LENGTH => 16;
…
sub random_string ($length = RANDOM_STRING_DEFAULT_LENGTH, …
```

?

And @Martchus "Considering how we call this" means what?

lib/OpenQA/Utils.pm
okurz: @perlpunk should I do what I suggested in the above code example?

lib/OpenQA/Utils.pm
```diff
@@ -361,33 +342,20 @@ sub run_cmd_with_log_return_error ($cmd, %args) {
     }
 }
 
-sub asset_type_from_setting {
-    # passing $value is optional but makes the result more accurate
-    # (it affects only UEFI_PFLASH_VARS currently).
-    my ($setting, $value) = @_;
-    $value //= '';
```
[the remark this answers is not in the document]
okurz: Well, from `git grep` I can't be sure as sometimes the secondary argument is passed on from elsewhere. So I guess I should do

```perl
sub asset_type_from_setting ($setting, $value = undef) {
    $value //= '';
```

?

lib/OpenQA/Utils.pm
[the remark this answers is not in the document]
Martchus: Yes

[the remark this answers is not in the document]
b10n1k: yes I meant exactly what you understand
```use constant RANDOM_STRING_DEFAULT_LENGTH => 16;```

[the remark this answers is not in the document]
Comment (okurz) [reply lib/OpenQA/Utils.pm]: done

File: lib/OpenQA/Utils.pm [hunk 7]
```diff
@@ -606,20 +553,10 @@ sub create_git_clone_list ($job_settings, $clones = {}) {
...
-        # give it one digit
-        return int($size * 10 + .5) / 10.;
-    }
-
-    return int($size + .5);
-}
-
-sub human_readable_size {
-    my ($size) = @_;
+# give it one digit
```
Wabri, line 556: Is this comment really necessary?

lib/OpenQA/Utils.pm
```diff
@@ -762,14 +695,13 @@ sub loaded_modules {
 
 # Fallback to loaded_modules if no arguments are given.
 # Accepts namespaces as arguments. If supplied, it will filter by them
-sub loaded_plugins {
-    my $ns = join('|', map { quotemeta } @_);
-    return @_ ? grep { /^$ns/ } loaded_modules() : loaded_modules();
+sub loaded_plugins (@ns) {
```
Wabri, line 698: Is `@ns` a namespace? I think is something more unrpocessed, should it deserve better name?

lib/OpenQA/Utils.pm
```diff
@@ -762,14 +695,13 @@ sub loaded_modules {
...
+    my $ns = join('|', map { quotemeta } @ns);
+    return @ns ? grep { /^$ns/ } loaded_modules() : loaded_modules();
 }
 
 # Walks a hash keeping keys as a stack
-sub hashwalker {
-    my ($hash, $callback, $keys) = @_;
+sub hashwalker ($hash, $callback, $keys = undef) {
     $keys = [] if !$keys;
```
Wabri, line 705: Why not:
```suggestion
sub hashwalker ($hash, $callback, $keys = []) {
```

Wabri (changes requested):
Some comments about the code and I think the commit description deserve more care!

lib/OpenQA/Utils.pm
Martchus: Because that would be an incompatible change, e.g. calls like `hashwalker($hash, $callback, $keys)` would cease to work if `$keys` is `undef`. Of course we *could* change it like you suggest but only after checking that no calls like in my example are made.

lib/OpenQA/Utils.pm
okurz: Please be aware that I did not introduce it. I just moved the comment according to the integration of signatures. I recommend to not change this as part of one commit as that is unrelated to introducing signatures

lib/OpenQA/Utils.pm
okurz: Again, not related to introducing signatures. The variable was called `@ns` and is still called `@ns` in the signature

lib/OpenQA/Utils.pm
okurz: Which again are not directly related to introducing signatures

Wabri (approved):

Martchus (approved):

PR #6238, opened 2025-03-02
Simplify WebSockets::Worker::Controller

perlpunk:
If `$worker` is undef, the rest of the statements in the try block don't get executed and the final return value will be false anyway.

`$current_job_id` will be only set if there is a `$current_job`, and all statements after that will only be true if `$current_job_id` is defined, expect some `log_trace` statements, which we can probably consider not important.

We can move most of the statements out of the try block because they are not changing anything in the database or can fail in an expected way. The "skip any further actions" if statement is only true when `$job_id` is defined, and the final statement is only true if `$job_id` is not defined, so it can return from the sub directly.

The only code left in the try is `reschedule_assigned_jobs()`.

### DIFF
--- a/lib/OpenQA/WebSockets/Controller/Worker.pm
+++ b/lib/OpenQA/WebSockets/Controller/Worker.pm
@@ -157,68 +157,54 @@ sub _message ($self, $json) {
         $worker_status->{last_seen} = $now;
 
         my $workers = $schema->resultset('Workers');
-        my $worker = $workers->find($worker_id);
+        return undef unless my $worker = $workers->find($worker_id);
 
         # log that we 'saw' the worker
-        if ($worker) {
-            try {
-                log_debug("Updating seen of worker $worker_id from worker_status ($current_worker_status)");
-                $worker->seen({error => $current_worker_error});
+        try {
+            log_debug("Updating seen of worker $worker_id from worker_status ($current_worker_status)");
+            $worker->seen({error => $current_worker_error});
 
-                # Tell the worker that we saw it (used for tests and debugging)
-                $tx->send({json => {type => 'info', seen => 1}});
-            }
-            catch {
-                log_error("Failed updating seen and error status of worker $worker_id: $_");    # uncoverable statement
-            };
+            # Tell the worker that we saw it (used for tests and debugging)
+            $tx->send({json => {type => 'info', seen => 1}});
         }
+        catch {
+            log_error("Failed updating seen and error status of worker $worker_id: $_");    # uncoverable statement
+        };
 
         # find the job currently associated with that worker and check whether the worker still
         # executes the job it is supposed to
-        my $current_job_id;
+        my @unfinished_jobs = $worker->unfinished_jobs;
+        my $current_job = $worker->job // $unfinished_jobs[0]
+          or return undef;
+        my $current_job_id = $current_job->id;
+        my $current_job_state = $current_job->state;
+
+        log_trace("Found job $current_job_id in DB from worker_status update sent by worker $worker_id");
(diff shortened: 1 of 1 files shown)


okurz (approved):

perlpunk:
> I suggest waiting for #6228 to not intefere with an ongoing task. I am not blocking but it would be nice

Doing my changes on top of yours would include undoing your changes first.
I wanted to mske it easier, so after this you can just do the try/catch change without having to refactor further.

perlpunk:
Somehow both my PRs I opened yesterday have the "WIP" check pending, and I don't know why.

lib/OpenQA/WebSockets/Controller/Worker.pm
```diff
@@ -158,76 +158,63 @@ sub _message ($self, $json) {
...
+            }
         }
         catch {
             # uncoverable statement
             log_warning("Unable to verify whether worker $worker_id runs its job(s) as expected: $_");
         };
 
         # consider the worker idle unless it claims to be broken or work on a job
-        $worker_status->{idle_despite_job_assignment}
-          = !$worker_is_broken && !defined $job_id && defined $current_job_id;
```
Martchus, line 230: This variable is no longer set to a false value in some cases. This leaves it completely missing in some cases considering we initially delete it (via `delete $worker_status->{idle_despite_job_assignment}`). However, it shouldn't make a difference because we never distinguish between the absence of this and a false value.

lib/OpenQA/WebSockets/Controller/Worker.pm
```diff
@@ -158,76 +158,63 @@ sub _message ($self, $json) {
...
+            $tx->send({json => {type => 'info', seen => 1}});
         }
+        catch {
+            log_error("Failed updating seen and error status of worker $worker_id: $_");    # uncoverable statement
+        };
 
         # find the job currently associated with that worker and check whether the worker still
         # executes the job it is supposed to
-        my $current_job_id;
+        my $registered_job_token;
```
Martchus, line 177:
```suggestion
```

lib/OpenQA/WebSockets/Controller/Worker.pm
```diff
@@ -158,76 +158,63 @@ sub _message ($self, $json) {
...
+        my @unfinished_jobs = $worker->unfinished_jobs;
+        my $current_job = $worker->job // $unfinished_jobs[0]
+          or return undef;
+        my $current_job_id = $current_job->id;
+        my $current_job_state = $current_job->state;
+
+        log_trace("Found job $current_job_id in DB from worker_status update sent by worker $worker_id");
+        log_trace("Received request has job id: $job_id") if defined $job_id;
+        $registered_job_token = $worker->get_property('JOBTOKEN');
```
Martchus, line 186:
```suggestion
        my $registered_job_token = $worker->get_property('JOBTOKEN');
```

lib/OpenQA/WebSockets/Controller/Worker.pm
```diff
@@ -158,76 +158,63 @@ sub _message ($self, $json) {
 
         my $workers = $schema->resultset('Workers');
         my $worker = $workers->find($worker_id);
+        return undef unless $worker;
```
Martchus, line 161:
```suggestion
        return undef unless my $worker = $workers->find($worker_id);
```

Martchus (commented):
Looks good to merge. The inline comments are just suggestions and further explanations why I think this change doesn't change the behavior.

perlpunk:
Thanks, I did your three suggested changes

okurz (approved):

Martchus (approved):

PR #6242, opened 2025-03-03
Remove any Try::Tiny reference and add check

b10n1k:
Refactor the deploy.t to remove the Try::Tiny dependency in order to use the explicit function from Test::Exception.

### DIFF
--- a/dbicdh/_common/upgrade/76-77/001-migrate-audit-events-to-scheduled-products.pl
+++ b/dbicdh/_common/upgrade/76-77/001-migrate-audit-events-to-scheduled-products.pl
@@ -10,7 +10,7 @@ use OpenQA::Schema::Result::ScheduledProducts;
 use OpenQA::Log qw(log_info log_warning);
 use OpenQA::Utils;
 use Mojo::JSON qw(decode_json encode_json);
-use Try::Tiny;
+use Feature::Compat::Try;
 
 sub {
     my ($schema) = @_;
@@ -26,8 +26,8 @@ sub {
         my $settings;
         try {
             $settings = decode_json($event->event_data);
-        };
-        if (!$settings) {
+        }
+        catch ($e) {
             log_warning(
                 "Unable to read settings from 'iso_create' audit event with ID $event_id. Skipping its migration.");
             next;
--- a/script/openqa-load-templates
+++ b/script/openqa-load-templates
@@ -54,7 +54,7 @@ use Mojo::Base -strict, -signatures;
 use FindBin;
 use lib "$FindBin::RealBin/../lib";
 use File::Basename qw(dirname);
-use Try::Tiny;
+use Feature::Compat::Try;
 use Data::Dump 'dd';
 use Mojo::Util qw(decamelize);
 use Mojo::URL;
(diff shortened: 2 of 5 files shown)


perlpunk (changes requested):
The title is wrong. We use it implicitly, not explicitly.

Also there are still other files using Try::Tiny.

dbicdh/_common/upgrade/76-77/001-migrate-audit-events-to-scheduled-products.pl
```diff
@@ -26,11 +26,10 @@
         my $settings;
         try {
             $settings = decode_json($event->event_data);
-        };
-        if (!$settings) {
+        }
+	catch ($e) {
             log_warning(
                 "Unable to read settings from 'iso_create' audit event with ID $event_id. Skipping its migration.");
-            next;
```
perlpunk, line 33: It's not clear to me why the `next` is not needed anymore

Wabri: I think he removed accidentally, otherwise it will create the scheduled_products and update the event.

script/openqa-load-templates
```diff
@@ -86,10 +86,10 @@ try {
     $info = Cpanel::JSON::XS->new->relaxed->decode($datafile);
     dd $info;
 }
-catch {
-    $info = eval $datafile;
-    if (my $error = $@) { die "Error in data file: $error\n" }
-};
+catch ($e) {
+    try { $info = Cpanel::JSON::XS->new->relaxed->decode($datafile) }
+    catch ($e) { die "Error in data file: $e\n" }
```
perlpunk, line 91: That should be left an `eval` for now. This is a "string eval", which evaluates the given string as perl code. It's quite different from the "block eval", which is more like `try`.

perlpunk (changes requested):

script/openqa-load-templates
perlpunk: plus, you are basically doing the same thing as in the try block here, so why should this succeed the second time, if it fails the first time.

dbicdh/_common/upgrade/76-77/001-migrate-audit-events-to-scheduled-products.pl
```diff
@@ -26,11 +26,11 @@
...
+	    next;
```
okurz, line 33: please revert this change

okurz (changes requested):

perlpunk:
> Refactor openqa-load-templates script with Feature::Compat::Try

I wouldn't request changes because of this but replacing Try::Tiny with Feature::Compat::Try is not a refactoring, especially if it's a simple replacement without any other changes.

perlpunk:
> Use excplicit Test::Exception on deploy.t

I commented before: we are not using Test::Exception explicitly, but implicitly. You can leave the word out and simply "Use Test::Exception in deploy.t". Notice the "in" instead of "on" as well.

okurz (approved):

okurz:
Rename s/Use Test::Exception on deploy.t/Use Test::Exception on deploy.t/ and likely perlpunk can approve as well :)

b10n1k:
> Rename s/Use Test::Exception on deploy.t/Use Test::Exception on deploy.t/ and likely perlpunk can approve as well :)

What is the difference on the string. they are the same

okurz:
> > Rename s/Use Test::Exception on deploy.t/Use Test::Exception on deploy.t/ and likely perlpunk can approve as well :)
> 
> What is the difference on the string. they are the same

Sorry, corrected, s/on/in/

perlpunk (approved):

perlpunk:
merging manually because the WIP app is broken currently

PR #6244, opened 2025-03-04
Apply mitigation for failing circleci with 7 retry attempts

Wabri:
reference: https://progress.opensuse.org/issues/175060

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -214,7 +214,7 @@ test-unstable: node_modules
 
 .PHONY: test-fullstack
 test-fullstack: node_modules
-	$(MAKE) test-with-database FULLSTACK=1 TIMEOUT_M=30 PROVE_ARGS="$$HARNESS t/full-stack.t t/33-developer_mode.t" RETRY=2
+	$(MAKE) test-with-database FULLSTACK=1 TIMEOUT_M=30 PROVE_ARGS="$$HARNESS t/full-stack.t t/33-developer_mode.t" RETRY=7
 
 .PHONY: test-fullstack-unstable
 test-fullstack-unstable: node_modules


Makefile
```diff
@@ -214,7 +214,7 @@ test-unstable: node_modules
 
 .PHONY: test-fullstack
 test-fullstack: node_modules
-	$(MAKE) test-with-database FULLSTACK=1 TIMEOUT_M=30 PROVE_ARGS="$$HARNESS t/full-stack.t t/33-developer_mode.t" RETRY=2
+	$(MAKE) test-with-database FULLSTACK=1 TIMEOUT_M=30 PROVE_ARGS="$$HARNESS t/full-stack.t t/33-developer_mode.t"
```
Martchus, line 217: How is that setting the number of retry attempts to 7?

Wabri: Ups :smile: 
image

Updated!

Makefile
```diff
@@ -214,7 +214,7 @@ test-unstable: node_modules
 
 .PHONY: test-fullstack
 test-fullstack: node_modules
-	$(MAKE) test-with-database FULLSTACK=1 TIMEOUT_M=30 PROVE_ARGS="$$HARNESS t/full-stack.t t/33-developer_mode.t" RETRY=2
+	$(MAKE) test-with-database FULLSTACK=1 TIMEOUT_M=30 PROVE_ARGS="$$HARNESS t/full-stack.t t/33-developer_mode.t" RETRY=7
```
okurz, line 217: But wouldn't that run into the circleCI timeout as we discussed in the daily today?

okurz (changes requested):

Makefile
Wabri: I've done a research about the circleci timeout and it turns out that it have only 1 timeout that is related to the `no-output` and is setted to 10min by default. We have setted to 50min in the [config.yml](https://github.com/os-autoinst/openQA/blob/1f5cdabdb0a0191ff1eea7b1030a6cd481fc4331/.circleci/config.yml#L87), do you think increasing it would be better?

perlpunk: It's not about the no-output time.
https://support.circleci.com/hc/en-us/articles/16616033407131-Max-Runtime-in-CircleCI-Server
> Build timeouts are enforced at the plan level. Jobs have a maximum runtime of 1 (Free), 3 (Performance), or 5 (Scale) hours depending on pricing plan.

Wabri: Since we have the free tier with strange credit calculation (see above) we cannot increase the build max runtime. 

image

I saw that opensuse org have the performance tier, since os-autoinst is part of it, does it make sense to have this separated plan?

perlpunk: The problem is
https://app.circleci.com/pipelines/github/os-autoinst/openQA/16047/workflows/be893671-47af-4c11-a2a7-a69140b42c57/jobs/154592/artifacts
```
...
# Wait for AJAX successful: wait_for_result_panel: waiting for '(?^u:State: running)' (count 3239 of 3240)
```

https://github.com/os-autoinst/openQA/blob/master/t/lib/OpenQA/Test/FullstackUtils.pm#L85
```
my $timeout = OpenQA::Test::TimeLimit::scale_timeout(ONE_MINUTE * 3) / $check_interval;
```
`$check_interval is 0.5. 3240 times with a 0.5s delay means 27 minutes alone for waiting.

The default timeout here would be 360s if SCALE_FACTOR is 1. Not sure what value we have for that in CI.
Maybe we can use `ONE_SECOND * 30` here instead and see if this is enough for normal tests, and then we can also increase the retries.

perlpunk: * #6248

Wabri: Awesome, I didn't consider making the fullstack test more efficient. do we still want to increase the retries to 7?

perlpunk: yes, that was the point of my pr. it's now failing faster, so we can have more retriies :)

Wabri: thank you :+1: I've request a new review from @okurz in order to merge it

perlpunk (approved):

okurz (approved):

PR #6248, opened 2025-03-04
ci: Reduce timeout for wait_for_result_panel

perlpunk:
Issue: https://progress.opensuse.org/issues/175060

See https://github.com/os-autoinst/openQA/pull/6244#discussion_r1979983862


perlpunk:
The usual runtime of the whole `t/fullstack.t` is around 4.5m, so I assume 27 / 6 which is 4.5 for only one call of `wait_for_result_panel` should be enough time.

okurz (approved):

Martchus (approved):

PR #6252, opened 2025-03-05
Simplify Controller::Test+Plugin::AMQP

okurz:


lib/OpenQA/WebAPI/Controller/Test.pm
```diff
@@ -179,12 +179,8 @@ sub clone ($self) {
...
 
 sub get_match_param ($self) {
-    my $match;
-    if (defined($self->param('match'))) {
-        $match = $self->param('match');
-        $match =~ s/[^\w\[\]\{\}\(\),:.+*?\\\$^|-]//g;    # sanitize
-    }
-    return $match;
+    return undef unless defined $self->param('match');
+    return $self->param('match') =~ s/[^\w\[\]\{\}\(\),:.+*?\\\$^|-]//gr;    # sanitize
```
Martchus, line 183:
```suggestion
    return undef unless defined(my $match = $self->param('match'));
    return $match =~ s/[^\w\[\]\{\}\(\),:.+*?\\\$^|-]//gr;    # sanitize
```

Martchus (commented):
Looks good. I only have a minor nitpick.

lib/OpenQA/WebAPI/Controller/Test.pm
okurz: done

perlpunk (approved):

PR #6253, opened 2025-03-05
Fix unstable UI test of result filter

Martchus:
* Use `wait_for_ajax` to wait for the server-side table like in other places of the same test
* See https://progress.opensuse.org/issues/178315

### DIFF
--- a/t/ui/01-list.t
+++ b/t/ui/01-list.t
@@ -428,6 +428,7 @@ subtest 'filter-finished' => sub {
 
 subtest 'result filter does not affect scheduled and running jobs' => sub {
     $driver->get('/tests?resultfilter=Failed');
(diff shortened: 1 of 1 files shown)


mimi1vx (approved):

perlpunk (commented):
Can it be moved out of unstable then?

Martchus:
> Can it be moved out of unstable then?

I'm not sure. The test wasn't considered unstable specifically due to the issue that is fixed here. The commit message from back then was just "Add back t/ui/01-list.t to "unstable" as not reliable" (1b6aa5dcc476bab467e10f7c529c7ca5e3ac9281) which isn't very helpful.

perlpunk (approved):
I guess we can have a look into the test results after a week and check if it still fails sometimes

okurz (approved):

PR #6258, opened 2025-03-06
Api v1 comment issue demo

b10n1k:


lib/OpenQA/WebAPI/Controller/API/V1/Comment.pm
```diff
@@ -8,8 +8,7 @@ use Date::Format;
 use OpenQA::App;
 use OpenQA::Utils qw(:DEFAULT href_to_bugref);
 use List::Util qw(min);
-
-=pod
+use Feature::Compat::Try=pod
```
Martchus, line 11: This change will probably break more than just `t/ui/15-comment.t`.

Martchus (changes requested):

PR #6262, opened 2025-03-06
Also use try/catch in Controller::API::V1::Comment

okurz:
The txn_guard needs an explicit undef in catch

Related progress issue: https://progress.opensuse.org/issues/176862

### DIFF
--- a/lib/OpenQA/WebAPI/Controller/API/V1/Comment.pm
+++ b/lib/OpenQA/WebAPI/Controller/API/V1/Comment.pm
@@ -5,6 +5,7 @@ package OpenQA::WebAPI::Controller::API::V1::Comment;
 use Mojo::Base 'Mojolicious::Controller', -signatures;
 
 use Date::Format;
+use Feature::Compat::Try;
 use OpenQA::App;
 use OpenQA::Utils qw(:DEFAULT href_to_bugref);
 use List::Util qw(min);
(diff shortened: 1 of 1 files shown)


lib/OpenQA/WebAPI/Controller/API/V1/Comment.pm
```diff
@@ -138,8 +139,11 @@ sub create ($self) {
             user_id => $self->current_user->id
         });
 
-    eval { $comment->handle_special_contents($self) };
-    return $self->render(json => {error => $@}, status => 400) if $@;
+    try { $comment->handle_special_contents($self) }
+    catch ($e) {
+        undef $txn_guard;
+        return $self->render(json => {error => $e}, status => 400);
+    }
```
Martchus, line 146:
```suggestion
    eval { $comment->handle_special_contents($self) };  # using eval as the txn guard interacts badly with try-catch
    return $self->render(json => {error => $@}, status => 400) if $@;
```

Martchus (commented):
Not really an improvement.

perlpunk (approved):
The advantage would be that we can filter for `$@` usage, only leaving `eval` where we are not using the error message.

lib/OpenQA/WebAPI/Controller/API/V1/Comment.pm
okurz: > The advantage would be that we can filter for `$@` usage, only leaving `eval` where we are not using the error message.

Let me demonstrate that in
* #6303

how do you like that?

Martchus: The case where we need to undefine the txn guard is still worse.

I also don't see it as necessity to enforce use of try-catch and thus filtering doesn't seem very useful. As long as we don't use `Try::Tiny` anymore which has counter-intuitive behavior for everyone coming from different programming languages we're good.

okurz:
included with https://github.com/os-autoinst/openQA/pull/6303

PR #6267, opened 2025-03-07
Remove Try::Tiny from ci-packages

b10n1k:

### DIFF
--- a/tools/ci/ci-packages.txt
+++ b/tools/ci/ci-packages.txt
@@ -231,7 +231,6 @@ perl-TimeDate-2.30
 perl-Time-Duration-Parse-0.16
 perl-Time-modules-2013.0912
 perl-Time-Moment-0.44
-perl-Try-Tiny-0.30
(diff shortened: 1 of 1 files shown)


okurz (approved):

Martchus (changes requested):
This is still used by important packages in our dependency chain. Therefore it would make sense to list it here explicitly which (when I remember correctly) allows us to track version updates.

(So this package falls into the same category as e.g. `perl-URI-Nested` which is also listed in this file even though it is not a direct dependency.)

perlpunk (commented):
additionally to what @Martchus wrote: that file is automatically updated by our bot, if necessary. no need to update it.

b10n1k:
> This is still used by important packages in our dependency chain. Therefore it would make sense to list it here explicitly which (when I remember correctly) allows us to track version updates.
> 
> (So this package falls into the same category as e.g. `perl-URI-Nested` which is also listed in this file even though it is not a direct dependency.)

I will close it then

PR #6270, opened 2025-03-07
Mention use of CRITICAL_LOAD_AVG_THRESHOLD

kalikiana:
See: https://progress.opensuse.org/issues/175060

### DIFF
--- a/docs/WritingTests.asciidoc
+++ b/docs/WritingTests.asciidoc
@@ -667,6 +667,16 @@ based on company conventions:
 WORKER_CLASS = planet-earth,continent-antarctica,location-my_station
 --------------------------------------------------------------------------------
 
+The load of the system can affect test execution and cause failures due to delays of command execution and https://en.wikipedia.org/wiki/Thrashing_(computer_science)[thrashing]. If the average over a period of 15 minutes exceeds the specified value the worker will not accept new jobs.
+
+[caption="File: "]
+.workers.ini
+[source,ini]
+--------------------------------------------------------------------------------
+[global]
+# Set to 0 to disable.
+#CRITICAL_LOAD_AVG_THRESHOLD = 40
+--------------------------------------------------------------------------------
 
 === Running a custom worker engine
 


okurz (approved):

docs/WritingTests.asciidoc
```diff
@@ -611,6 +611,19 @@ based on company conventions:
...
+[caption="File: "]
+.workers.ini
+[source,ini]
+--------------------------------------------------------------------------------
+[global]
+# Set to 0 to disable.
+#CRITICAL_LOAD_AVG_THRESHOLD = 40
+--------------------------------------------------------------------------------
+
+Note that the value can also be overridden by setting `OPENQA_LOAD_AVG_FILE` in the environment, such as in the systemd service.
```
Martchus, line 625: The expression "the value" makes it sound like the env variable would allow overriding the threshold value set via CRITICAL_LOAD_AVG_THRESHOLD as shown above. However, the env variable really replaces the "file" the load is read from. This is only meant for testing/development. So I'd remove this whole paragraph about the env variable from this user-facing documentation.

Martchus (changes requested):

docs/WritingTests.asciidoc
d3flex: Not sure. As testers/developers we still use the documentation from time to time. However it would better to make that clear in the text. that it is used for development purposes.

okurz: We don't have all variables mentioned in documentation and that should be seen as ok as our documentation in config files or code are usually sufficient. Here this is about an environment file and I guess the best way to find all is by grepping for `$ENV`. The only better way I could see is to create a separate documentation document explaining all environment variables and ensuring with an automatic static check that all environment variables are also listed in such document.

Wabri (approved):

docs/WritingTests.asciidoc
kalikiana: So I just realized this was still open...

I don't have a super strong opinion on where this should go, but ideally we document these variables **somewhere**. If you have a suggestion where to put it, let me know, otherwise I'd prefer to go with this instead of not documenting them.

okurz: As I wrote the only better way I could see is to create a separate documentation document explaining all environment variables and ensuring with an automatic static check that all environment variables are also listed in such document, whatever the name or location which is secondary

Martchus: I'd also create a separate documentation for that.

Of course the only thing that really prevents me from hitting the approval button is that the documentation you added about the env variable is simply wrong as explained in my comment on March. There I suggested to simply remove it completely as an easy way out but you might as well also fix the wording. (But I think I really only introduced this variable to make the unit test work and not so that users can change this file path.)

kalikiana: Oh. Re-reading it now I get what you mean it's wrong. I only got the part about it being meant for testing before 😅

kalikiana: I tried to come up with a simple check for missing variables, but that suggests way more than I was planning to document here.

So I decided to just remove the environment variable for now.

PR #6279, opened 2025-03-10
Avoid race condition in `t/ui/21-admin-needles.t`

Martchus:
* Wait for intermediate state when selecting options to minimize the chance for a race condition leading to sometimes observing the intermediate state in further checks (and thus failing the test)
* See https://progress.opensuse.org/issues/167611#note-6


r-richardson (approved):

perlpunk:
codecov is sleeping, merging manually

PR #6283, opened 2025-03-11
Ensure fullstack tests run under high system load

Wabri:
https://progress.opensuse.org/issues/175060

### DIFF
--- a/t/data/workers.ini
+++ b/t/data/workers.ini
@@ -3,6 +3,9 @@
 [global]
 HOST = http://localhost:9526
 WORKER_HOSTNAME = 127.0.0.1
+# ensure fullstack tests run even under high load
+# the value set to 0 prevents jobs from being blocked in ci
+CRITICAL_LOAD_AVG_THRESHOLD = 0
 
 [1]
 WORKER_CLASS = qemu_i386,qemu_x86_64


Martchus (changes requested):
I don't think this will work because this is a configuration setting (and not an environment variable).

You probably need to add this to https://github.com/os-autoinst/openQA/blob/master/t/data/workers.ini or adjust the [setup function](https://github.com/os-autoinst/openQA/blob/8eee9517ff72daae713b6e5295d1357b78d93848/t/lib/OpenQA/Test/Utils.pm#L375) where this config file is copied.

Wabri:
Yes, you are right! We already had this conversation with the same outcome, I'm updating the pr!

t/data/workers.ini
```diff
@@ -3,6 +3,7 @@
 [global]
 HOST = http://localhost:9526
 WORKER_HOSTNAME = 127.0.0.1
+CRITICAL_LOAD_AVG_THRESHOLD = 0;
```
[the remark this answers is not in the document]
Wabri: I guess that was the problem, I went into more complicated way to force that within the setup_fullstack_temp_dir

Wabri:
The compose/webui-docker-compose will fail for known problem with `(download|cdn).opensuse.org`, but it should work now!

t/lib/OpenQA/Test/Utils.pm
```diff
@@ -388,6 +388,12 @@ sub setup_fullstack_temp_dir {
     $config =~ s/^#git_auto_clone = .*/git_auto_clone = no/m;
     $config =~ s/^#git_auto_update = .*/git_auto_update = no/m;
     $configdir->child('openqa.ini')->spew($config);
+
+    my $workers_ini_path = $configdir->child('workers.ini')->to_string;
+    my $workers_ini = Config::IniFiles->new(-file => $workers_ini_path);
+    $workers_ini->setval('global', 'CRITICAL_LOAD_AVG_THRESHOLD', 0);
+    $workers_ini->RewriteConfig;
```
Martchus, line 395: Why is this required? And doesn't this just change the file to be like it is anyway (as you modified `t/data/workers.ini` directly)?

Wabri: I thought that other tests use this subroutine and they might be overwright the value of it so I'm forcing the value. Do you think is not necessary? is this a better way to handle the main problem or change the workers.ini is enough?

perlpunk: But you don't need both changes. Either change the original file, or change it in `setup_fullstack_temp_dir`.
And if a test is explicitly overwriting that setting, it probably does it for a reason.

Wabri: Yes, make sense. Updating

t/lib/OpenQA/Test/Utils.pm
```diff
@@ -388,6 +388,7 @@ sub setup_fullstack_temp_dir {
     $config =~ s/^#git_auto_clone = .*/git_auto_clone = no/m;
     $config =~ s/^#git_auto_update = .*/git_auto_update = no/m;
     $configdir->child('openqa.ini')->spew($config);
+
```
okurz, line 391: unrelated

t/data/workers.ini
```diff
@@ -3,6 +3,7 @@
 [global]
 HOST = http://localhost:9526
 WORKER_HOSTNAME = 127.0.0.1
+CRITICAL_LOAD_AVG_THRESHOLD = 0
```
okurz, line 6: I suggest you add a comment here explaining context and the reason

okurz (changes requested):
In your commit+PR please explain the "Why" rather than the "What"

t/data/workers.ini
Wabri: I've added a comment above the variable, what do you think?

Wabri:
Update pr with better commit description, title text and add a comment in the newest worker.ini configuration

t/data/workers.ini
okurz: nice!

Wabri:
@okurz since this is a resolution for high load during ci I'm not sure if a local test is a good check to do. The only thing I've done is create a new test case with this:

```
my $configuration = "t/data/workers.ini";
my $cfg = Config::IniFiles->new(-file => $configuration);

my $treshold = $cfg->val('global', 'CRITICAL_LOAD_AVG_THRESHOLD');
is $treshold, 40, 'CRITICAL_LOAD_AVG_THRESHOLD is 0';
```

and it results in error due to the value in workers.ini set to 0 and not 40.

okurz (approved):

PR #6285, opened 2025-03-11
Improve reading of text_data in module results

perlpunk:
Issue: https://progress.opensuse.org/issues/178597

### DIFF
--- a/lib/OpenQA/Schema/Result/JobModules.pm
+++ b/lib/OpenQA/Schema/Result/JobModules.pm
@@ -3,8 +3,7 @@
 
 package OpenQA::Schema::Result::JobModules;
 
-
-use Mojo::Base 'DBIx::Class::Core';
+use Mojo::Base 'DBIx::Class::Core', -signatures;
 
 use OpenQA::Jobs::Constants;
 use Mojo::JSON qw(decode_json encode_json);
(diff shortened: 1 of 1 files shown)


lib/OpenQA/Schema/Result/JobModules.pm
```diff
@@ -251,4 +232,12 @@ sub finalize_results {
     }
 }
 
+sub _read_text_data ($self, $dir, $filename) {
+    my $text_data = do {
+        try { path($dir, $filename)->slurp }
+        catch ($e) { "Unable to read $filename." }
+    };
+    return decode('UTF-8', $text_data) // "Unable to decode $filename.";
```
perlpunk, line 240: The question is, should `finalize_results` actually modify the content of the `details-*.json` like this, or should the fallback "Unable to ..." only be presented on the fly when viewing job details?

perlpunk: I changed it and left `finalize_results` as it was

okurz (changes requested):

lib/OpenQA/Schema/Result/JobModules.pm
```diff
@@ -109,11 +109,11 @@ sub results ($self, %options) {
...
         my $text_file_name = $step->{text};
         if (!$skip_text_data && $text_file_name && !defined $step->{text_data}) {
-            my $text_file = path($dir, $text_file_name);
-            $step->{text_data} = do {
-                try { decode('UTF-8', $text_file->slurp) }
+            my $text_data = do {
+                try { path($dir, $text_file_name)->slurp }
                 catch ($e) { "Unable to read $text_file_name." }
             };
+            $step->{text_data} = decode('UTF-8', $text_data) // "Unable to decode $text_file_name.";
```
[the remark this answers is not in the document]
perlpunk: Hm, I like to keep try statements short to focus on the code that possibly can die, but this suggestion seems to be readable enough.

Martchus (approved):

PR #6287, opened 2025-03-12
Link a job only with existing tracking issues

d3flex:
Assume that the referer has a issue_id otherwise do not link the URL in the comments. Once a ticket has been created the referer should contain an ID despite where it comes from.
It could go one step further to send a request and check the response but it doesn't seem to be necessary.

https://progress.opensuse.org/issues/176826

### DIFF
--- a/lib/OpenQA/Schema/ResultSet/Jobs.pm
+++ b/lib/OpenQA/Schema/ResultSet/Jobs.pm
@@ -472,12 +472,20 @@ sub mark_job_linked ($self, $jobid, $referer_url) {
     return undef unless $referer_host;
     return log_trace("Unrecognized referer '$referer_host'")
       unless grep { $referer_host eq $_ } @{$app->config->{global}->{recognized_referers}};
+    return undef unless _specific_issue_exists($referer);
     my $job = $self->find({id => $jobid});
     return undef if !$job || ($referer->path_query =~ /^\/?$/);
     my $comments = $job->comments;
     return undef if $comments->search({text => {like => 'label:linked%'}}, {rows => 1})->first;
     my $user = $self->result_source->schema->resultset('Users')->system({select => ['id']});
     $comments->create_with_event({text => "label:linked Job mentioned in $referer_url", user_id => $user->id});
+}
+
+sub _specific_issue_exists ($url) {
+    my $path = $url->path();
+    my ($issue_id) = $path =~ m/(?:id=|\/)(\d+)/;
+    # Assume that the url is valid if there is issue_id
+    return defined $issue_id;
 }
 
 1;
--- a/t/10-jobs-referal.t
+++ b/t/10-jobs-referal.t
@@ -32,7 +32,7 @@ sub job_is_linked ($job) {
 }
 
 subtest 'job is marked as linked if accessed from recognized referal' => sub {
-    my $test_referer = 'http://test.referer.info/foobar';
+    my $test_referer = 'http://test.referer.info/foobar/123';
     $t->app->config->{global}->{recognized_referers}
       = ['test.referer.info', 'test.referer1.info', 'test.referer2.info', 'test.referer3.info'];
     my %_settings = %settings;
@@ -57,6 +57,14 @@ subtest 'job is marked as linked if accessed from recognized referal' => sub {
     $t->get_ok('/tests/' . $job->id . '/modules/' . $module->id . '/steps/1' => {Referer => $test_referer})
       ->status_is(302);
     is job_is_linked($job), 1, 'job linked after accessed from known referer';
+
+    subtest 'do not link not existing tickets from recognized referal' => sub {
+        $test_referer = 'http://test.referer.info/foobar/new';
+        my $job = _job_create(\%_settings);
+        is job_is_linked($job), 0, 'new job is not linked';
+        $t->get_ok('/tests/' . $job->id => {Referer => $test_referer})->status_is(200);
+        is job_is_linked($job), 0, 'job is not linked from known referer without an issue_id';
+    };
 };
 
 subtest 'job is not marked as linked if accessed from unrecognized referal' => sub {
--- a/t/10-jobs-referal.t
+++ b/t/10-jobs-referal.t
@@ -41,12 +41,10 @@ subtest 'job is marked as linked if accessed from recognized referal' => sub {
     my $cb = $openqa_events->on(openqa_comment_create => sub ($events, $data) { push @comment_events, $data });
     $_settings{TEST} = 'refJobTest';
     my $job = _job_create(\%_settings);
-    my $linked = job_is_linked($job);
-    is($linked, 0, 'new job is not linked');
+    is job_is_linked($job), 0, 'new job is not linked';
     $t->get_ok('/tests/' . $job->id => {Referer => $test_referer})->status_is(200);
-    $linked = job_is_linked($job);
-    is($linked, 1, 'job linked after accessed from known referer');
-    is(scalar @comment_events, 1, 'exactly one comment event emitted') or always_explain \@comment_events;
+    is job_is_linked($job), 1, 'job linked after accessed from known referer';
+    is scalar @comment_events, 1, 'exactly one comment event emitted' or always_explain \@comment_events;
     $openqa_events->unsubscribe($cb);
 
     $_settings{TEST} = 'refJobTest-step';
@@ -55,12 +53,10 @@ subtest 'job is marked as linked if accessed from recognized referal' => sub {
     $job->insert_module({name => 'a', category => 'a', script => 'a', flags => {}});
     my $module = $job->modules->find({name => 'a'});
     $job->update;
-    $linked = job_is_linked($job);
-    is($linked, 0, 'new job is not linked');
+    is job_is_linked($job), 0, 'new job is not linked';
     $t->get_ok('/tests/' . $job->id . '/modules/' . $module->id . '/steps/1' => {Referer => $test_referer})
       ->status_is(302);
-    $linked = job_is_linked($job);
-    is($linked, 1, 'job linked after accessed from known referer');
+    is job_is_linked($job), 1, 'job linked after accessed from known referer';
 };
 
 subtest 'job is not marked as linked if accessed from unrecognized referal' => sub {
@@ -69,14 +65,11 @@ subtest 'job is not marked as linked if accessed from unrecognized referal' => s
     my %_settings = %settings;
     $_settings{TEST} = 'refJobTest2';
     my $job = _job_create(\%_settings);
-    my $linked = job_is_linked($job);
-    is($linked, 0, 'new job is not linked');
+    is job_is_linked($job), 0, 'new job is not linked';
     $t->get_ok('/tests/' . $job->id => {Referer => 'http://unknown.referer.info'})->status_is(200);
-    $linked = job_is_linked($job);
-    is($linked, 0, 'job not linked after accessed from unknown referer');
+    is job_is_linked($job), 0, 'job not linked after accessed from unknown referer';
     $t->get_ok('/tests/' . $job->id => {Referer => 'http://test.referer.info/'})->status_is(200);
-    $linked = job_is_linked($job);
-    is($linked, 0, 'job not linked after accessed from referer with empty query_path');
+    is job_is_linked($job), 0, 'job not linked after accessed from referer with empty query_path';
 };
 
 done_testing;


lib/OpenQA/Schema/ResultSet/Jobs.pm
```diff
@@ -485,8 +485,19 @@ sub mark_job_linked {
...
     my $comments = $job->comments;
     return undef if $comments->search({text => {like => 'label:linked%'}}, {rows => 1})->first;
+    return undef unless _specific_issue_exists($referer);
     my $user = $self->result_source->schema->resultset('Users')->system({select => ['id']});
     $comments->create_with_event({text => "label:linked Job mentioned in $referer_url", user_id => $user->id});
 }
 
+sub _specific_issue_exists {
+    my ($referer_url) = @_;
+    my $url = Mojo::URL->new($referer_url);
```
perlpunk, line 495: The passed variable is already a Mojo::URL object

t/10-jobs-referal.t
```diff
@@ -61,6 +61,17 @@ subtest 'job is marked as linked if accessed from recognized referal' => sub {
...
+
+
+    subtest 'do not link not existing tickets from recognized referal' => sub {
+        $test_referer = 'http://test.referer.info/foobar/new';
+        my $job = _job_create(\%_settings);
+        my $linked = job_is_linked($job);
+        is($linked, 0, 'new job is not linked');
+        $t->get_ok('/tests/' . $job->id => {Referer => $test_referer})->status_is(200);
+        $linked = job_is_linked($job);
+        is($linked, 0, 'job is not linked from known referer without an issue_id');
```
okurz, line 73:
```suggestion
        is job_is_linked($job), 0, 'job is not linked from known referer without an issue_id';
```

perlpunk (changes requested):
Checking if the url path contains a number is a neat idea.
Now, this wouldn't work with bugzilla urls `https://bugzilla.suse.com/show_bug.cgi?id=123` because the number is in the query string.
We know that bugzilla somehow doesn't pass the full referrer anymore anyway, but this should ideally work for all kinds of urls, even if the issue id is in the query string.

What do others think?

okurz (changes requested):
@baierjan I don't see how that solves the original problem from https://progress.opensuse.org/issues/176826 . What is your opinion on this?

perlpunk:
> @baierjan I don't see how that solves the original problem from https://progress.opensuse.org/issues/176826 . What is your opinion on this?

I think it would solve the original problem, at least if the code does what it states (check if there is a number in the path). But it would be good to show this in a test.

Why do you think it doesn't solve the problem?

lib/OpenQA/Schema/ResultSet/Jobs.pm
```diff
@@ -485,8 +485,19 @@ sub mark_job_linked {
     return undef if !$job || ($referer->path_query =~ /^\/?$/);
     my $comments = $job->comments;
     return undef if $comments->search({text => {like => 'label:linked%'}}, {rows => 1})->first;
+    return undef unless _specific_issue_exists($referer);
     my $user = $self->result_source->schema->resultset('Users')->system({select => ['id']});
     $comments->create_with_event({text => "label:linked Job mentioned in $referer_url", user_id => $user->id});
 }
 
+sub _specific_issue_exists {
+    my ($referer_url) = @_;
```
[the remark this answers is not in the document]
b10n1k: this is not yet done. I noticed but I decided to make this PR first. Would you like to resolve and continue the signature out of here?

lib/OpenQA/Schema/ResultSet/Jobs.pm
```diff
@@ -485,8 +485,19 @@ sub mark_job_linked {
...
+    my $url = Mojo::URL->new($referer_url);
+    my $path = $url->path();
+    my ($issue_id) = $path =~ m{\d+};
```
okurz, line 497: > Checking if the url path contains a number is a neat idea. Now, this wouldn't work with bugzilla urls `https://bugzilla.suse.com/show_bug.cgi?id=123` because the number is in the query string. We know that bugzilla somehow doesn't pass the full referrer anymore anyway, but this should ideally work for all kinds of urls, even if the issue id is in the query string.
> 
> What do others think?

I think it's a bad design to have such strict requirements for issue references in this code point. Please look into https://github.com/os-autoinst/openQA/blob/master/lib/OpenQA/Utils.pm#L41

okurz (changes requested):

t/10-jobs-referal.t
```diff
@@ -61,6 +61,17 @@ subtest 'job is marked as linked if accessed from recognized referal' => sub {
       ->status_is(302);
     $linked = job_is_linked($job);
     is($linked, 1, 'job linked after accessed from known referer');
+
+
+    subtest 'do not link not existing tickets from recognized referal' => sub {
+        $test_referer = 'http://test.referer.info/foobar/new';
+        my $job = _job_create(\%_settings);
+        my $linked = job_is_linked($job);
+        is($linked, 0, 'new job is not linked');
```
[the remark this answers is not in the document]
b10n1k: I am amazed sometimes that I get suggestions for things which they aligned with the existing approved stuff. And if I had done it different, then someone would complain that I am not concistent. And I am not sure whether is one or the other policy. 
Having said that, I will create another commit in the same PR for all the checks. would this cause other problems?

lib/OpenQA/Schema/ResultSet/Jobs.pm
perlpunk: Yes, I think I would also go for a mapping like the one for the bugrefs. That would of course change the way it currently works, so it should be done in a way it doesn't break existing installations.

perlpunk: the link @okurz gave.
but i think we wouldn't even need a mapping, just a list.
maybe we can support both host names like before plus regexes that would match the whole url?
maybe if the string starts and ends with a `/` it is supposed to be uses as a regex? something like that...

okurz (changes requested):

t/10-jobs-referal.t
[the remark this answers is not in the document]
perlpunk: > I am amazed sometimes that I get suggestions for things which they aligned with the existing approved stuff. And if I had done it different, then someone would complain that I am not concistent. And I am not sure whether is one or the other policy.

Don't take it personal. It's just a suggestion to improve the code. Reviewers don't always look at existing surrounding code.
You used the same style as existing tests, and that's fine.
You can now use the suggestionto imprrove the existing code as well.

lib/OpenQA/Schema/ResultSet/Jobs.pm
```diff
@@ -485,8 +485,17 @@ sub mark_job_linked {
     return undef if !$job || ($referer->path_query =~ /^\/?$/);
     my $comments = $job->comments;
     return undef if $comments->search({text => {like => 'label:linked%'}}, {rows => 1})->first;
+    return undef unless _specific_issue_exists($referer);
     my $user = $self->result_source->schema->resultset('Users')->system({select => ['id']});
     $comments->create_with_event({text => "label:linked Job mentioned in $referer_url", user_id => $user->id});
 }
 
+sub _specific_issue_exists {
+    my ($url) = @_;
```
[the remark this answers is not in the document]
b10n1k: I work on another PR for that. and I am dumping into problems. so I think it was a good idea to separate those two concepts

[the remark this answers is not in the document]
perlpunk: Why a separate PR? The change here is really easy, as this file already has the `signatures` feature enabled.
Does the following not work?
```suggestion
sub _specific_issue_exists ($url) {
```

b10n1k:
https://github.com/os-autoinst/openQA/pull/6287#discussion_r1991701660

if I do, that means I would have to change all the existing references as you see for instance https://github.com/os-autoinst/openQA/blob/master/t/10-jobs-referal.t#L45

lib/OpenQA/Schema/ResultSet/Jobs.pm
b10n1k: Still not sure what is the advantage of that mapping (or even list). We could achieve the same enhancing the regex (as it just checking for ID now, and as you said might not be sufficient for all, but still). it is simplier and you dont have to maintain another structure of data. 
Should I provide an alternative regex or feel free to implement another solution.

perlpunk: This is what we have in our default config file:
```
#recognized_referers = bugzilla.suse.com bugzilla.opensuse.org progress.opensuse.org github.com
```
That's the list I mean. We can still use this, but I think we have to support regexes there.

b10n1k: i take it as `provide an alternative regex`

lib/OpenQA/Schema/ResultSet/Jobs.pm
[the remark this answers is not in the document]
b10n1k: there is alwasys the fear to make things complicated. smaller PR in a hope of faster results. there is no reason for rush AFAIT. 
 anyway maybe you want to close this thread in favor of https://github.com/os-autoinst/openQA/pull/6288/files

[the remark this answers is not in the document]
perlpunk: Sure, a PR with changing all subs in this file is fine, but it wouldn't have been necessary. You could have just used a signature for the new sub you introduced :)

lib/OpenQA/Schema/ResultSet/Jobs.pm
perlpunk: Rather "Make it possible to provide a regex in the configuration"

lib/OpenQA/Schema/ResultSet/Jobs.pm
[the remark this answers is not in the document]
b10n1k: fact. i just didnt notice that signatures are already used.

lib/OpenQA/Schema/ResultSet/Jobs.pm
perlpunk: Maybe we can simply use the `%BUGURLS` mapping. But that's hardcoded.
What do others think?

b10n1k: i just saw the last two messages. 
the only thing I did is in https://github.com/os-autoinst/openQA/commit/71faa24c23161335b8c7a4a4a2d17305a406c099 but dont expect anything fancy.

Anyway I am waiting the existinting discussion to close to apply the other improvements. 

> Rather "Make it possible to provide a regex in the configuration"

if we go with this, I would also wish another PR for that

lib/OpenQA/Schema/ResultSet/Jobs.pm
```diff
@@ -476,8 +476,16 @@ sub mark_job_linked ($self, $jobid, $referer_url) {
     return undef if !$job || ($referer->path_query =~ /^\/?$/);
     my $comments = $job->comments;
     return undef if $comments->search({text => {like => 'label:linked%'}}, {rows => 1})->first;
+    return undef unless _specific_issue_exists($referer);
```
baierjan, line 479: As this is a referer-related check which only needs the url, would be better to place it just after checking the `$referer_host` (i.e. line 475) to spare ourselves some potentially costly database lookups (finding the job and searching comments)?

perlpunk: It should not be its own check anyway. We do already check the `recognized_referers` in line 473, so I don't know why this was put in a different location.

lib/OpenQA/Schema/ResultSet/Jobs.pm
perlpunk: Why a different PR? What would this PR here then be about?
Simply checking for digits is not solving any problem in general, it only fixes it specifically for `progress.opensuse.org`. But people can configure multiple referer hosts with different issue trackers, and for some your digit check would even break it.

I don't have a complete solution myself, but you need to think about the existing `recognized_referers` and should not develop something independent from that.

baierjan:
> > @baierjan I don't see how that solves the original problem from https://progress.opensuse.org/issues/176826 . What is your opinion on this?
> 
> I think it would solve the original problem, at least if the code does what it states (check if there is a number in the path). But it would be good to show this in a test.
> 
> Why do you think it doesn't solve the problem?

I agree that checking for issue id (i.e. number) in the HTTP referer should solve the original issue (as I was in the new issue page). This is (of course) definitely not an universal solution, but should work at least for progress and bugzilla. Just as an example, it will not work for Jira tickets (which are luckily currently not configured on o3 nor osd) as they use `PROJECT-123` as an issue identifier; that was also the reason why one of the suggestion is mentioning a regex map.

lib/OpenQA/Schema/ResultSet/Jobs.pm
perlpunk: @b10n1k were you actually aware of `recognized_referers` before creating this PR?

lib/OpenQA/Schema/ResultSet/Jobs.pm
baierjan: > Maybe we can simply use the `%BUGURLS` mapping. But that's hardcoded. What do others think?

I can live with having the regexes in hardcoded in `%BUGURLS` and at the same time having an option to provide a custom "universal" regex in the configuration to cover a generic issue tracker we didn't cover.

Of course, the ideal option would be to put everything in the configuration. Maybe this is also big enough for its own section?

Instead of:
```ini
recognized_referers = bugzilla.suse.com progress.opensuse.org
```

something like:
```ini
[recognized_referers]
bugzilla.suse.com = id=(\d+)
progress.opensuse.org = issues/(\d+)
jira.example.com = (\w+-\d+)
```

Or is that too much?

lib/OpenQA/Schema/ResultSet/Jobs.pm
```diff
@@ -476,8 +476,16 @@ sub mark_job_linked ($self, $jobid, $referer_url) {
...
     my $user = $self->result_source->schema->resultset('Users')->system({select => ['id']});
     $comments->create_with_event({text => "label:linked Job mentioned in $referer_url", user_id => $user->id});
 }
 
+sub _specific_issue_exists ($url) {
+    my $path = $url->path();
+    my ($issue_id) = $path =~ m/(?:id=|\/)(\d+)/;
```
perlpunk, line 486: This is wrong in several regards.
1. `$path` is only the path of the url and does not include the `query_string` (the thing after the `?`).

2. If you would do that pattern matching on the path including the `query_string`, then it would still not solve the problem in general.

Please think about the user perspective. We allow to configure **any** bugtracker here. How the id of a certain issue is coded in a url depends on that bugtracker.
It doesn't make sense to let people configure bugtracker hosts, but then have a hardcoded regex to look for a number somewhere specific.

lib/OpenQA/Schema/ResultSet/Jobs.pm
perlpunk: The other solution I was thinking of was
```
recognized_referers = /bugzilla.suse.com/show_bug.cgi?id=\d+/ some.other.host /https://progress.opensuse.org/issues/\d+/
```

But putting our bugref mapping into the configuration could be even better, and we don't actually need regexes because we already have everything we need:
```
bmo => 'https://bugzilla.mozilla.org/show_bug.cgi?id='
```

baierjan: On progress, it is `https://progress.opensuse.org/issues/new` vs `https://progress.opensuse.org/issues/123456` and I guess it could be similar for other issue tracker, so you definitely need more than a static prefix.

perlpunk: After discussing with @baierjan I created a ticket to reimplement bugrefs via configuration: https://progress.opensuse.org/issues/178771
@AdamWill this might also be interesting for you and others, as you added some bugtracker urls in the past

AdamWill: oh, yeah, that'd be great. i ran across another one that isn't implemented just the other day...qt bug tracker, I think.

t/10-jobs-referal.t
```diff
@@ -61,6 +61,17 @@ subtest 'job is marked as linked if accessed from recognized referal' => sub {
       ->status_is(302);
     $linked = job_is_linked($job);
     is($linked, 1, 'job linked after accessed from known referer');
+
```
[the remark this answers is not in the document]
b10n1k: I wonder why this is still unresolved. there were two blanklines. I removed the one.

PR #6288, opened 2025-03-12
Use signatures on lib/OpenQA/Schema/ResultSet/Jobs.pm

b10n1k:
And refactor the test in order to pass to the create_from_settings the expected missing params for `scheduled_product_id`.


okurz:
same as last please s/on/in/ :)

t/10-jobs-referal.t
```diff
@@ -31,6 +31,9 @@ sub job_is_linked ($job) {
     $job->comments->find({text => {like => 'label:linked%'}}) ? 1 : 0;
 }
 
+$schema->txn_begin;
+my $scheduled_products_rs = $schema->resultset('ScheduledProducts')->create({id => 30, settings => []});
+
```
okurz, line 36: unrelated

okurz (changes requested):

t/10-jobs-referal.t
b10n1k: should the beginning of the subtest be after a blankline for visibility or..?

t/10-jobs-referal.t
perlpunk, line 1: Leave the test as it is

perlpunk (changes requested):

PR #6294, opened 2025-03-14
Ignore short Git service outages

r-richardson:
Introduce git service outage handler module
    
This commit creates a outage handling mechanism which allows for smaller gitlab outages to be ignored while prolonged outages still cause a failure as they did before.
    
Related Issue: https://progress.opensuse.org/issues/179038

### DIFF
--- a/lib/OpenQA/Git/ServerAvailability.pm
+++ b/lib/OpenQA/Git/ServerAvailability.pm
@@ -1,0 +1,40 @@
+# Copyright SUSE LLC
+# SPDX-License-Identifier: GPL-2.0-or-later
+
+package OpenQA::Git::ServerAvailability;
+use Mojo::Base -strict, -signatures;
+use Mojo::File qw(path);
+use Time::Seconds qw(ONE_HOUR);
+use OpenQA::Utils qw(prjdir);
+use Exporter qw(import);
+
+our @EXPORT_OK = qw(report_server_unavailable report_server_available SKIP FAIL);
+
+use constant MAX_DURATION_SECONDS => ($ENV{OPENQA_GIT_SERVER_OUTAGE_DURATION} || (ONE_HOUR / 2));
+
+use constant {
+    SKIP => 'SKIP',
+    FAIL => 'FAIL',
+};
+
+sub _state_file ($app, $server_name) {
+    return path(($ENV{OPENQA_GIT_SERVER_OUTAGE_FILE} // prjdir . '/git_server_outage') . ".$server_name.flag");
+}
+
+sub report_server_available ($app, $server_name) {
+    _state_file($app, $server_name)->remove;
+}
+
+sub report_server_unavailable ($app, $server_name) {
+    my $file = _state_file($app, $server_name);
+    my $stat = $file->stat;
+    unless (defined $stat) {
+        $file->dirname->make_path;
+        $file->touch;
+        return SKIP;
+    }
+    my $file_age = time - $stat->mtime;
+    return ($file_age >= MAX_DURATION_SECONDS) ? FAIL : SKIP;
+}
+
+1;
--- a/lib/OpenQA/Task/Git/Clone.pm
+++ b/lib/OpenQA/Task/Git/Clone.pm
@@ -9,7 +9,8 @@ use Mojo::File;
 use Feature::Compat::Try;
 use List::Util qw(min);
 use OpenQA::Log qw(log_debug);
-use Time::Seconds 'ONE_HOUR';
+use OpenQA::Git::ServerAvailability qw(report_server_unavailable report_server_available SKIP FAIL);
+use Time::Seconds qw(ONE_HOUR);
 
 sub register ($self, $app, @) {
     $app->minion->add_task(git_clone => \&_git_clone_all);
@@ -47,11 +48,30 @@ sub _git_clone_all ($job, $clones) {
 
     # iterate clones sorted by path length to ensure that a needle dir is always cloned after the corresponding casedir
     for my $path (sort { length($a) <=> length($b) || $a cmp $b } keys %$clones) {
-        my $url = $clones->{$path};
         die "Don't even think about putting '..' into '$path'." if $path =~ /\.\./;
+
+        my $git = OpenQA::Git->new(app => $app, dir => $path);
+        my $origin_url = -d $path ? $git->get_origin_url : undef;
+        my $url = $origin_url // $clones->{$path};
+        my $server_host = _extract_server_host($url);
         my $error;
-        try { _git_clone($app, $job, $ctx, $path, $url); next; }
+        try {
+            _git_clone($git, $ctx, $path, $clones->{$path}, $origin_url);
+            report_server_available($app, $server_host);
+            next;
+        }
         catch ($e) { $error = $e }
+        if ($error =~ /Internal API unreachable/) {
+            my $outcome = report_server_unavailable($app, $server_host);
+            if ($outcome eq 'SKIP') {
+                $ctx->info('Git server outage likely, skipping clone job.');
+                return;
+            }
+            elsif ($outcome eq 'FAIL') {
+                $ctx->info('Prolonged Git server outage, failing the job.');
+                return $job->fail($error);
+            }
+        }
 
         # unblock openQA jobs despite network errors under best-effort configuration
         my $retries = $job->retries;
@@ -73,8 +93,7 @@ sub _git_clone_all ($job, $clones) {
     }
 }
 
-sub _git_clone ($app, $job, $ctx, $path, $url) {
-    my $git = OpenQA::Git->new(app => $app, dir => $path);
+sub _git_clone ($git, $ctx, $path, $url, $origin_url) {
     $ctx->debug(sprintf q{Updating '%s' to '%s'}, $path, ($url // 'n/a'));
     my $requested_branch;
     if ($url) {
@@ -85,7 +104,6 @@ sub _git_clone ($app, $job, $ctx, $path, $url) {
         # An initial clone fetches all refs, we are done
         return $git->clone_url($url) unless -d $path;
 
-        my $origin_url = $git->get_origin_url;
         if ($url ne $origin_url) {
             $ctx->info(<<~"END_OF_MESSAGE");
                 Local checkout at $path has origin $origin_url but requesting to clone from $url. The requested URL will not be cloned.
@@ -94,7 +112,7 @@ sub _git_clone ($app, $job, $ctx, $path, $url) {
         }
     }
     else {
-        $url = $git->get_origin_url;
+        $url = $origin_url;
     }
 
     return if ($requested_branch and $requested_branch !~ tr/a-f0-9//c and $git->check_sha($requested_branch));
@@ -115,6 +133,17 @@ sub _git_clone ($app, $job, $ctx, $path, $url) {
     # updating default branch (including checkout)
     $git->fetch($requested_branch);
     $git->reset_hard($requested_branch) if ($requested_branch eq $current_branch);
+}
+
+sub _extract_server_host ($url) {
+    if ($url =~ m{^[^@]+@([^:]+):}) {
+        # e.g. "git@gitlab.suse.de:qa/foo", "user@host:repo"
+        return $1;
+    }
+    else {
+        my $mojo_url = Mojo::URL->new($url);
+        return $mojo_url->host // 'other';
+    }
 }
 
 1;
--- a/t/14-grutasks-git.t
+++ b/t/14-grutasks-git.t
@@ -8,6 +8,7 @@ use Mojo::Base -signatures;
 use FindBin;
 use lib "$FindBin::Bin/lib", "$FindBin::Bin/../external/os-autoinst-common/lib";
 use OpenQA::Task::Git::Clone;
+use OpenQA::Git::ServerAvailability qw(report_server_unavailable report_server_available SKIP FAIL);
 require OpenQA::Test::Database;
 use OpenQA::Test::Utils qw(run_gru_job perform_minion_jobs);
 use OpenQA::Test::TimeLimit '20';
@@ -30,8 +31,9 @@ chdir $workdir;
 path('t/data/openqa/db')->make_path;
 my $git_clones = "$workdir/git-clones";
 mkdir $git_clones;
-mkdir "$git_clones/$_"
-  for qw(default branch dirty-error dirty-status nodefault wrong-url sha1 sha2 sha-error sha-branchname);
+path("$git_clones/$_")->make_path
+  for
+  qw(default branch dirty-error dirty-status nodefault wrong-url sha1 sha2 sha-error sha-branchname opensuse opensuse/needles);
 mkdir 't';
 dircopy "$FindBin::Bin/$_", "$workdir/t/$_" or BAIL_OUT($!) for qw(data);
 
@@ -266,6 +268,8 @@ subtest 'git clone' => sub {
 subtest 'git_update_all' => sub {
     my $clone_mock = Test::MockModule->new('OpenQA::Task::Git::Clone');
     $clone_mock->redefine(_git_clone => sub (@args) { die 'fake disconnect' });
+    my $openqa_git = Test::MockModule->new('OpenQA::Git');
+    $openqa_git->redefine(get_origin_url => 'foo');
 
     my $git_config = $t->app->config->{'scm git'};
     $git_config->{git_auto_update} = 'yes';
@@ -448,11 +452,103 @@ subtest 'delete_needles' => sub {
     like $error->{message}, qr{Unable to rm via Git}, 'expected error from git';
 
     subtest 'minion guard' => sub {
-        my $guard = $t->app->minion->guard("git_clone_t/data/openqa/share/tests/fedora/needles_task", ONE_HOUR);
+        my $guard = $t->app->minion->guard('git_clone_t/data/openqa/share/tests/fedora/needles_task', ONE_HOUR);
         $res = run_gru_job(@gru_args);
         is $res->{state}, 'finished', 'job finished';
         $error = $res->{result}->{errors}->[0];
         like $error->{message}, qr{Another git task for.*fedora.*is ongoing}, 'expected error message';
+    };
+};
+
+subtest ServerAvailability => sub {
+    my $tmpdir = tempdir();
+    local $ENV{OPENQA_GIT_SERVER_OUTAGE_FILE} = "$tmpdir/foo/mock";
+
+    subtest 'no flag file => skip' => sub {
+        my $outcome = report_server_unavailable($t->app, 'gitlab');
+        is $outcome, 'SKIP', 'Skips when flag file for gitlab is absent';
+    };
+
+    subtest 'file present but younger than 1800 => skip' => sub {
+        my $mock_file = $tmpdir->child('foo/mock.gitlab.flag')->touch;
+        my $outcome = report_server_unavailable($t->app, 'gitlab');
+        is $outcome, 'SKIP', 'Skips the job for an mtime < 1800';
+    };
+
+    subtest 'file present and older => fail' => sub {
+        my $mock_file = $tmpdir->child('foo/mock.gitlab.flag')->touch;
+        my $old_time = time - 2000;
+        utime($old_time, $old_time, $mock_file)
+          or die "Couldn't change mtime on file: $mock_file - $!";
+        my $outcome = report_server_unavailable($t->app, 'gitlab');
+        is $outcome, 'FAIL', 'Fails the job when mtime >= 1800';
+    };
+
+    subtest 'report_server_available removes existing flag file' => sub {
+        my $mock_file = $tmpdir->child('foo/mock.gitlab.flag')->touch;
+        ok -f $mock_file, 'mock.gitlab.flag exists initially';
+        report_server_available($t->app, 'gitlab');
+        ok !-f $mock_file, 'gitlab flag file was removed';
+    };
+
+    subtest 'multiple servers do not conflict' => sub {
+        my $outcome_gitlab = report_server_unavailable($t->app, 'gitlab');
+        is $outcome_gitlab, 'SKIP', 'No gitlab flag => skip';
+        my $gitlab_file = "$tmpdir/foo/mock.gitlab.flag";
+        ok -f $gitlab_file, 'Created mock.gitlab.flag';
+        my $outcome_github = report_server_unavailable($t->app, 'github');
+        is $outcome_github, 'SKIP', 'No github flag => skip';
+        my $github_file = "$tmpdir/foo/mock.github.flag";
+        ok -f $github_file, 'Created mock.github.flag';
+
+        my $old_time = time - 2000;
+        utime($old_time, $old_time, $github_file);
+        # Re-check
+        $outcome_gitlab = report_server_unavailable($t->app, 'gitlab');
+        $outcome_github = report_server_unavailable($t->app, 'github');
+        is $outcome_gitlab, 'SKIP', 'GitLab file still fresh => skip';
+        is $outcome_github, 'FAIL', 'GitHub file older => fail';
+    };
+
+    subtest Clone => sub {
+        my $clone_mock = Test::MockModule->new('OpenQA::Task::Git::Clone');
+        $clone_mock->redefine(_git_clone => sub { die "Internal API unreachable\n" });
+        subtest 'Internal API unreachable => skip' => sub {
+            my @gru_args = (
+                $t->app,
+                'git_clone',
+                {
+                    '/my/mock/path' => 'git@gitlab.suse.de:qa/foo',
+                });
+            my $res;
+            combined_like {
+                $res = run_gru_job(@gru_args);
+            }
+            qr/Git server outage likely, skipping clone job/i,
+              'Minion clone job skipped due to short Git server outage';
+            is $res->{state}, 'finished', 'Job ended in "finished" state';
+            is $res->{result}, 'Job successfully executed', 'Job result indicates success (skip)';
+        };
+
+        subtest 'Internal API unreachable => fail' => sub {
+            $tmpdir->child('foo/mock.gitlab.suse.de.flag')->touch;
+            my $old_time = time - 2000;    # "older" than 1800s
+            utime($old_time, $old_time, "$tmpdir/foo/mock.gitlab.suse.de.flag");
+            my @gru_args = (
+                $t->app,
+                'git_clone',
+                {
+                    '/my/other/mock/path' => 'https://gitlab.suse.de/qa/foo',
+                });
+            my $res;
+            combined_like {
+                $res = run_gru_job(@gru_args);
+            }
+            qr/Prolonged Git server outage, failing the job/i,
+              'Minion clone job failed due to prolonged Git server outage';
+            is $res->{state}, 'failed', 'Job ended in "failed" state';
+            like $res->{result}, qr/Internal API unreachable/, 'Job result includes original error message';
+        };
     };
 };
 


perlpunk (changes requested):
I'm not sure if this is the right place. We are already calling this from minion jobs which are doing their own retries.

So it should at least default to no retries.
What is the related ticket?

r-richardson:
> I'm not sure if this is the right place. We are already calling this from minion jobs which are doing their own retries.

ok thanks, i wasnt sure, so i should move it into  `lib/OpenQA/Task/Git/Clone.pm`  instead ? Or where would you suggest ?

> So it should at least default to no retries. What is the related ticket?

Sorry i had i amended the ticket link to the commit and forgot to update the PR description. Updated.

perlpunk:
Ah, that ticket.
If you look at the latest failed git_clone job:
https://openqa.suse.de/minion/jobs?id=14843257
```
result: |
  Error detecting remote default branch name for "https://gitlab.suse.de/suse-liberty-linux/openqa-tests-sll":  fatal: unable to access 'https://gitlab.suse.de/suse-liberty-linux/openqa-tests-sll/': The requested URL returned error: 502 at /usr/share/openqa/script/../lib/OpenQA/Git.pm line 147.
retried: 2025-03-12T00:05:56.497198Z
retries: 10
```
you can see that it retried 10 times.
So we already have a retry for that. And that's using Task:Git::Clone (which is using OpenQA::Git)

I think what was discussed for the ticket is that we get that specific error message from the ticket `Internal API unreachable` then we know it is very likely gitlab maintenance, which means that we can give up and mark it as success, because we know people also can't push at that time, so it will be fine if we don't update in this case.

But there's also a suggestion to use a state file to check for how long this is going on already.
So it's a bit more complicated.

lib/OpenQA/Git/MaintenanceOutage.pm
```diff
@@ -0,0 +1,33 @@
...
+        };
+        return {skip => 1};
+    }
+
+    my $content = path($STATE_FILE)->slurp;
+    my $outage_t = Time::Piece->new($content);
+    my $now = Time::Piece->new;
+
+    my $diff_hours = ($now - $outage_t) / (60 * 60);
+    if ($diff_hours >= $MAX_DURATION_HOURS) {
```
okurz, line 24:
```suggestion
    return $diff_hours >= $MAX_DURATION_HOURS ? {fail => 1} : {skip => 1};
```

lib/OpenQA/Git/MaintenanceOutage.pm
```diff
@@ -0,0 +1,33 @@
...
+            path($STATE_FILE)->spurt(time . "\n");    # time of intial failure
+        };
+        return {skip => 1};
+    }
+
+    my $content = path($STATE_FILE)->slurp;
+    my $outage_t = Time::Piece->new($content);
+    my $now = Time::Piece->new;
+
+    my $diff_hours = ($now - $outage_t) / (60 * 60);
```
okurz, line 23: Use Time::Seconds constants

lib/OpenQA/Git/MaintenanceOutage.pm
```diff
@@ -0,0 +1,33 @@
+package OpenQA::Git::MaintenanceOutage;
+use Mojo::Base -strict;
+use Mojo::File 'path';
+use Time::Piece;
+
+my $STATE_FILE = "/var/lib/openqa/share/gitlab_maintenance.lock";
+my $MAX_DURATION_HOURS = 24;    #TODO make this a config var ?
+
+sub decide_outcome {
```
okurz, line 9: Use signatures

okurz (changes requested):
Tests?

lib/OpenQA/Git/MaintenanceOutage.pm
```diff
@@ -0,0 +1,33 @@
...
+    unless (-f $STATE_FILE) {
+        eval {
+            path($STATE_FILE)->spurt(time . "\n");    # time of intial failure
+        };
+        return {skip => 1};
+    }
+
+    my $content = path($STATE_FILE)->slurp;
+    my $outage_t = Time::Piece->new($content);
+    my $now = Time::Piece->new;
```
perlpunk, line 21: If you just want to subtract two epoch second integers I think you don't need to create a Time::Piece object

perlpunk (commented):
I think 24 hours as a default is a bit long. I would rather suggest something like 30 minutes.

Now, looking at the logic of the code and say we configure 30 minutes. Let's say we have the following git operations:
* 15:03 error (Internal API unreachable)
* 15:05 success
* ... more cases of success
* 16:00 error (Internal API unreachable)

What would the code do for the failre at 16:00?

What would the code do

lib/OpenQA/Task/Git/Clone.pm
```diff
@@ -53,6 +53,18 @@ sub _git_clone_all ($job, $clones) {
...
 
+        if ($error =~ /Internal API unreachable/) {
+            my $outage_handling = OpenQA::Git::MaintenanceOutage::decide_outcome($app, $ctx, $error);
+            if ($outage_handling->{skip}) {
+                $ctx->info("Git clone: GitLab outage likely, skipping clone job).");
+                return;
+            }
+            elsif ($outage_handling->{fail}) {
+                $ctx->info("Prolonged GitLab outage; failing the job.");
+                die $error;
```
perlpunk, line 64:
```suggestion
                $job->fail($error);
```

lib/OpenQA/Git/MaintenanceOutage.pm
```diff
@@ -0,0 +1,33 @@
...
+
+my $STATE_FILE = "/var/lib/openqa/share/gitlab_maintenance.lock";
+my $MAX_DURATION_HOURS = 24;    #TODO make this a config var ?
+
+sub decide_outcome {
+    my ($app, $ctx, $error_string) = @_;
+
+    unless (-f $STATE_FILE) {
+        eval {
+            path($STATE_FILE)->spurt(time . "\n");    # time of intial failure
```
perlpunk, line 14: (`spurt` is deprecated and should be replaced by `spew`.)
But `spew` isn't  an atomic operation, so there is a race condition, and the file could end up with partial content.
We would either need file locking, or use a file per repository, because the minion job guards the job with a lock for all paths.

perlpunk: Maybe we can just use `touch` and use the file mtime for the timestamp. So we wouldn't need locking.
The race condition between the `-f` and the `touch` wouldn't matter in any case.

lib/OpenQA/Git/MaintenanceOutage.pm
```diff
@@ -0,0 +1,31 @@
...
+sub _state_file {
+    return $ENV{OPENQA_GITLAB_MAINTENANCE_FILE} || $DEFAULT_STATE_FILE;
+}
+
+sub decide_outcome {
+    my $state_file = _state_file();
+
+    unless (-f $state_file) {
+        path($state_file)->touch;
+        return {skip => 1};
```
Martchus, line 18: Maybe it would make more sense if this function returned a single value from a set of previously defined constants.

lib/OpenQA/Git/MaintenanceOutage.pm
```diff
@@ -0,0 +1,31 @@
+package OpenQA::Git::MaintenanceOutage;
+use Mojo::Base -strict, -signatures;
+use Mojo::File 'path';
+use Time::Seconds;
+
+my $DEFAULT_STATE_FILE = "/var/lib/openqa/share/gitlab_maintenance.lock";
+my $MAX_DURATION_SECONDS = 1800;
```
Martchus, line 7: This should probably be a constant and be overridable by an env variable (like we do in many places).

lib/OpenQA/Git/MaintenanceOutage.pm
```diff
@@ -0,0 +1,31 @@
+package OpenQA::Git::MaintenanceOutage;
+use Mojo::Base -strict, -signatures;
+use Mojo::File 'path';
+use Time::Seconds;
+
+my $DEFAULT_STATE_FILE = "/var/lib/openqa/share/gitlab_maintenance.lock";
```
Martchus, line 6: This should not be under `…/share/…` and the `/var/lib` part should not be hard-coded. I suggest you follow the same pattern as in `lib/OpenQA/Schema/ResultSet/Assets.pm` (except for using the `cache` subdirectory).

lib/OpenQA/Git/MaintenanceOutage.pm
```diff
@@ -0,0 +1,44 @@
...
+}
+
+sub decide_outcome($app) {
+    my $file = _state_file($app);
+    unless (-f $file) {
+        $file->touch;
+        return SKIP;
+    }
+    my $file_age = time - _state_file($app)->stat->mtime;
+    if ($file_age >= ($ENV{OPENQA_GITLAB_MAINTENANCE_DURATION} || MAX_DURATION_SECONDS)) {
```
okurz, line 35: Use a ternary operator in the return

lib/OpenQA/Git/MaintenanceOutage.pm
```diff
@@ -0,0 +1,44 @@
...
+        return FAIL;
+    }
+    else {
+        return SKIP;
+    }
+}
+
+1;
+
```
okurz, line 44: Remove the blank line

lib/OpenQA/Task/Git/Clone.pm
```diff
@@ -115,6 +131,7 @@ sub _git_clone ($app, $job, $ctx, $path, $url) {
     # updating default branch (including checkout)
     $git->fetch($requested_branch);
     $git->reset_hard($requested_branch) if ($requested_branch eq $current_branch);
+
```
okurz, line 134: Why?

okurz (changes requested):

lib/OpenQA/Git/MaintenanceOutage.pm
```diff
@@ -0,0 +1,37 @@
...
+# Remove or rename these "constants" in favor of a dynamic path
+use constant MAX_DURATION_SECONDS => 1800;
+
+use constant {
+    SKIP => 'SKIP',
+    FAIL => 'FAIL',
+};
+
+sub _state_file ($app) {
+    path($ENV{OPENQA_GITLAB_MAINTENANCE_FILE} // ($app->config->{global}->{cache_dir} // '/var/lib/openqa/cache')
```
okurz, line 17: The package has a generic name which I like but here you say "gitlab". Please keep it generic. And I don't think you need to set the complete path when the config isn't set. I assume we have other variables which define by default /var/lib/openqa?

lib/OpenQA/Task/Git/Clone.pm
```diff
@@ -50,9 +50,25 @@ sub _git_clone_all ($job, $clones) {
...
+            _git_clone($app, $job, $ctx, $path, $url);
+            OpenQA::Git::MaintenanceOutage::remove_state_file($app);
+            next;
+        }
         catch ($e) { $error = $e }
 
+        if ($error =~ /Internal API unreachable/) {
+            my $outage_handling = OpenQA::Git::MaintenanceOutage::decide_outcome($app, $ctx, $error);
+            if ($outage_handling->{skip}) {
+                $ctx->info('Git clone: GitLab outage likely, skipping clone job.');
```
okurz, line 63: Please keep it generic. There should be nothing about gitlab here

okurz (changes requested):

lib/OpenQA/Git/MaintenanceOutage.pm
```diff
@@ -0,0 +1,31 @@
...
+use Mojo::Base -strict, -signatures;
+use Mojo::File 'path';
+use Time::Seconds;
+
+my $DEFAULT_STATE_FILE = "/var/lib/openqa/share/gitlab_maintenance.lock";
+my $MAX_DURATION_SECONDS = 1800;
+
+sub _state_file {
+    return $ENV{OPENQA_GITLAB_MAINTENANCE_FILE} || $DEFAULT_STATE_FILE;
+}
```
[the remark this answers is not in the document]
r-richardson: removed the return statement, why break indentation though ?

lib/OpenQA/Git/MaintenanceOutage.pm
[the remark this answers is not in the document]
r-richardson: Variable not used anymore, but i replaced double with single quotes for all literal strings

lib/OpenQA/Git/MaintenanceOutage.pm
r-richardson: Yes i agree, though i'm not sure on how to word it, any better idea then `Source Control Management System` ? :sweat_smile:

lib/OpenQA/Task/Git/Clone.pm
r-richardson: Yes i agree, though i'm not sure on how to word it, `Source Control Management System Outage` seemed rather long/weird to me, any better idea  ? :sweat_smile:

lib/OpenQA/Git/MaintenanceOutage.pm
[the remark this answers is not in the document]
okurz: Maybe it looks weird in your browser. What this does is not change indentation but actually put everything into one line which is a common style in particular this kind of function that only directly returns a value

lib/OpenQA/Git/MaintenanceOutage.pm
okurz: Well, we are within "Git::MaintenanceOutage". How about "git server outage"?

okurz: By the way thinking about it I don't think we should call the package "MaintenanceOutage". We don't know why a remote server is not reachable. We can't assume it's due to a maintenance

lib/OpenQA/Task/Git/Clone.pm
okurz: Well, we are within "Git::MaintenanceOutage". How about "git server outage"?

t/14-grutasks-git.t
```diff
@@ -454,6 +455,54 @@ subtest 'delete_needles' => sub {
         $error = $res->{result}->{errors}->[0];
         like $error->{message}, qr{Another git task for.*fedora.*is ongoing}, 'expected error message';
     };
+
+    subtest 'MaintenanceOutage: no lock file => skip' => sub {
+        my $tmpdir = tempdir(CLEANUP => 1);
```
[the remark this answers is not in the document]
r-richardson: yes you're right. I put it in the no needles subtest temoprarily for now while working on the final subtest, since it seemed easier to retrigger a git job in there to see if the state file was successfully removed.

lib/OpenQA/Git/MaintenanceOutage.pm
r-richardson: Changed to `OPENQA_GIT_SERVER_MAINTENANCE_FILE` and `OPENQA_GIT_SERVER_MAINTENANCE_DURATION`, is that better ?

lib/OpenQA/Task/Git/Clone.pm
r-richardson: Renamed to `...Git server outage...` for both info texts

lib/OpenQA/Git/MaintenanceOutage.pm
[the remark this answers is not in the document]
r-richardson: Ah you're right i should've tried zooming out :D Tried a oneliner like this as well, but tidy doesnt seem to like it and will make it a three-liner again

lib/OpenQA/Git/MaintenanceOutage.pm
```diff
@@ -0,0 +1,37 @@
...
+use constant MAX_DURATION_SECONDS => 1800;
+
+use constant {
+    SKIP => 'SKIP',
+    FAIL => 'FAIL',
+};
+
+sub _state_file ($app) {
+    path($ENV{OPENQA_GIT_SERVER_MAINTENANCE_FILE} // ($app->config->{global}->{cache_dir} // '/var/lib/openqa/cache')
+          . ($ENV{OPENQA_GIT_SERVER_MAINTENANCE_FILE} ? '' : '/gitlab_maintenance.lock'));
```
Martchus, line 18: I'd avoid the repetition of the lengthy `$ENV{OPENQA_GIT_SERVER_MAINTENANCE_FILE}`.

lib/OpenQA/Git/MaintenanceOutage.pm
```diff
@@ -0,0 +1,37 @@
...
+}
+
+sub remove_state_file ($app) {
+    my $file = _state_file($app);
+    return unless -f $file;
+    $file->remove;
+}
```
Martchus, line 25: `Mojo::File::remove` doesn't fail if the file doesn't exist so the additional condition/check isn't necessary:

```suggestion
sub remove_state_file ($app) { _stat_file($app)->remove }
```

lib/OpenQA/Task/Git/Clone.pm
```diff
@@ -50,9 +50,25 @@ sub _git_clone_all ($job, $clones) {
...
+        try {
+            _git_clone($app, $job, $ctx, $path, $url);
+            OpenQA::Git::MaintenanceOutage::remove_state_file($app);
+            next;
+        }
         catch ($e) { $error = $e }
 
+        if ($error =~ /Internal API unreachable/) {
+            my $outage_handling = OpenQA::Git::MaintenanceOutage::decide_outcome($app, $ctx, $error);
+            if ($outage_handling->{skip}) {
```
Martchus, line 62: This needs to be changed to be in-line with what `decide_outcome` returns.

lib/OpenQA/Git/MaintenanceOutage.pm
```diff
@@ -0,0 +1,33 @@
+# Copyright SUSE LLC
+# SPDX-License-Identifier: GPL-2.0-or-later
+
+package OpenQA::Git::MaintenanceOutage;
+use Mojo::Base -strict, -signatures;
+use Mojo::File 'path';
+
+# Remove or rename these "constants" in favor of a dynamic path
+use constant MAX_DURATION_SECONDS => 1800;
```
okurz, line 9: Please use Time::Seconds here

okurz (changes requested):

t/14-grutasks-git.t
```diff
@@ -448,12 +449,93 @@ subtest 'delete_needles' => sub {
...
         $res = run_gru_job(@gru_args);
         is $res->{state}, 'finished', 'job finished';
         $error = $res->{result}->{errors}->[0];
         like $error->{message}, qr{Another git task for.*fedora.*is ongoing}, 'expected error message';
     };
+
+};
+
+subtest 'MaintenanceOutage.pm: no lock file => skip' => sub {
+    my $tmpdir = tempdir(CLEANUP => 1);
```
perlpunk, line 462: CLEANUP is the default.
```suggestion
    my $tmpdir = tempdir();
```

lib/OpenQA/Git/MaintenanceOutage.pm
r-richardson: changed it to `path($ENV{OPENQA_GIT_SERVER_MAINTENANCE_FILE} // ($app->config->{global}->{cache_dir} // '/var/lib/openqa/cache') . '/gitlab_maintenance.lock');`
I wanted to make sure that '/gitlab_maintenance.lock' is not appended if a path was provided via a env variable, but i guess this still works due to operator precedence.. ?

okurz (approved):

lib/OpenQA/Git/MaintenanceOutage.pm
```diff
@@ -0,0 +1,33 @@
...
+
+sub remove_state_file ($app) { _state_file($app)->remove }
+
+sub decide_outcome ($app) {
+    my $file = _state_file($app);
+    unless (-f $file) {
+        $file->touch;
+        return SKIP;
+    }
+    my $file_age = time - _state_file($app)->stat->mtime;
```
perlpunk, line 29: This repeats `_state_file($app)` instead of reusing `$file`;

But also, there is a race condition.
If the file was removed between the `-f` check and this line, it will die with `Can't call method "mtime" on an undefined value`.

To fix this, you could just use `my $stat = _state_file($app)->stat` and check if that is defined or not for deciding whether the file exists; and then just call mtime on it

perlpunk (changes requested):

lib/OpenQA/Git/MaintenanceOutage.pm
```diff
@@ -0,0 +1,33 @@
...
+    FAIL => 'FAIL',
+};
+
+sub _state_file ($app) {
+    path($ENV{OPENQA_GIT_SERVER_MAINTENANCE_FILE} // ($app->config->{global}->{cache_dir} // '/var/lib/openqa/cache') . '/gitlab_maintenance.lock');
+}
+
+sub remove_state_file ($app) { _state_file($app)->remove }
+
+sub decide_outcome ($app) {
```
perlpunk, line 23: I find the subroutine name a bit generic. Most subroutines are kind of deciding on an outcome :)
Maybe `check_availability` or something?

lib/OpenQA/Git/MaintenanceOutage.pm
```diff
@@ -0,0 +1,33 @@
...
+# Remove or rename these "constants" in favor of a dynamic path
+use constant MAX_DURATION_SECONDS => ONE_HOUR / 2;
+
+use constant {
+    SKIP => 'SKIP',
+    FAIL => 'FAIL',
+};
+
+sub _state_file ($app) {
+    path($ENV{OPENQA_GIT_SERVER_MAINTENANCE_FILE} // ($app->config->{global}->{cache_dir} // '/var/lib/openqa/cache') . '/gitlab_maintenance.lock');
```
perlpunk, line 18: Instead of a hardcoded `/var/lib/openqa` I would use
```
use OpenQA::Utils qw(prjdir);
```
Also, is this `cache` a new directory? It doesn't exist on o3, for example.
As a result, `$file->touch` below will fail because it needs to have that directory first.

Looking in our source code, `/var/lib/openqa/cache` is the default directory for the cache service (which runs on workers).

It might be confusing to use the same directory here.

lib/OpenQA/Git/MaintenanceOutage.pm
okurz, line 1: You still have the name "MaintenanceOutage" included but as I stated in before you don't know if a remote service is down for maintenance or other reasons

okurz (changes requested):

perlpunk (commented):
My original suggestion was to have such a state file per repository (inside the repository clone itself).
But the suggestion to have one global file sounded easier.
But thinking about it again, it might not scale well regarding the different remote servers.
We currently use our own gitlab server and github. So it's fine for our current usage, as the error message is only known for gitlab.
But I think the file should be per git server name. One server might be down, the other not.

edit: actually, it is not even fine with our current use case.
One git_clone from gitlab fails, state file is created. Next git_clone from github works, state file is removed.

okurz (approved):

lib/OpenQA/Git/GitServerOutage.pm
```diff
@@ -0,0 +1,39 @@
...
+    return path(
+        ($ENV{OPENQA_GIT_SERVER_OUTAGE_FILE} // ($app->config->{global}->{base_dir} // prjdir) . '/git_server_outage')
+        . ".$server_name.lock");
+}
+
+sub remove_state_file ($app, $server_name) {
+    _state_file($app, $server_name)->remove;
+}
+
+sub check_availability ($app, $server_name) {
```
Martchus, line 28: This is not really checking the availability. It is meant to report that the server is unavailable and returning whether to fail/skip:
```suggestion
sub report_server_unavailable ($app, $server_name) {
```

lib/OpenQA/Git/GitServerOutage.pm
```diff
@@ -0,0 +1,39 @@
...
+    FAIL => 'FAIL',
+};
+
+sub _state_file ($app, $server_name) {
+    return path(
+        ($ENV{OPENQA_GIT_SERVER_OUTAGE_FILE} // ($app->config->{global}->{base_dir} // prjdir) . '/git_server_outage')
+        . ".$server_name.lock");
+}
+
+sub remove_state_file ($app, $server_name) {
```
Martchus, line 24: This function is supposed to be called when the server is available, right?
```suggestion
sub report_server_available ($app, $server_name) {
```

lib/OpenQA/Git/GitServerOutage.pm
```diff
@@ -0,0 +1,39 @@
+
+# Copyright SUSE LLC
+# SPDX-License-Identifier: GPL-2.0-or-later
+
+package OpenQA::Git::GitServerOutage;
```
Martchus, line 5: The package name makes it sound like this represents a concrete outage but it is rather for tracking availability in general. I'd also remove the `Git…` prefix because it is within `OpenQA::Git` anyway.

```suggestion
package OpenQA::Git::ServerAvailability;
```

lib/OpenQA/Task/Git/Clone.pm
```diff
@@ -47,12 +47,34 @@ sub _git_clone_all ($job, $clones) {
...
     # iterate clones sorted by path length to ensure that a needle dir is always cloned after the corresponding casedir
     for my $path (sort { length($a) <=> length($b) || $a cmp $b } keys %$clones) {
-        my $url = $clones->{$path};
+        my $url = $clones->{$path} // '';
         die "Don't even think about putting '..' into '$path'." if $path =~ /\.\./;
+
+        my $server;
+        if ($url =~ /gitlab/) { $server = 'gitlab'; }
+        elsif ($url =~ /github/) { $server = 'github'; }
+        else { $server = 'other'; }
```
Martchus, line 56:
```suggestion
        if ($url =~ /gitlab/) { $server = 'gitlab' }
        elsif ($url =~ /github/) { $server = 'github' }
        else { $server = 'other' }
```

Martchus, line 56: Maybe this could even be one line (not tested):
```suggestion
        my $server = ($url =~ m/(github|gitlab)/) ? $1 : 'other';
```

t/14-grutasks-git.t
```diff
@@ -448,12 +449,114 @@ subtest 'delete_needles' => sub {
...
+};
+
+subtest 'Clone.pm: Internal API unreachable => skip' => sub {
+    my $tmpdir = tempdir();
+    local $ENV{OPENQA_GIT_SERVER_OUTAGE_FILE} = "$tmpdir/mock.other.lock";
+    my $clone_mock = Test::MockModule->new('OpenQA::Task::Git::Clone');
+    $clone_mock->redefine(
+        _git_clone => sub {
+            die "Internal API unreachable\n";
+        });
```
Martchus, line 527:
```suggestion
    $clone_mock->redefine(_git_clone => sub { die "Internal API unreachable\n" });
```
The same counts for the other occurrence and a few other cases where things probably just fit in one line.

Martchus (changes requested):

lib/OpenQA/Task/Git/Clone.pm
```diff
@@ -47,12 +47,34 @@ sub _git_clone_all ($job, $clones) {
...
+        try {
+            _git_clone($app, $job, $ctx, $path, $url);
+            OpenQA::Git::GitServerOutage::remove_state_file($app, $server);
+            next;
+        }
         catch ($e) { $error = $e }
 
+        if ($error =~ /Internal API unreachable/) {
+            my $outcome = OpenQA::Git::GitServerOutage::check_availability($app, $server);
+            if ($outcome eq 'SKIP') {
```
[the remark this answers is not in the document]
r-richardson: getting `Bareword "OpenQA::Git::ServerAvailability::SKIP" not allowed while "strict subs" in use` tried explicitly exporting (`our @EXPORT_OK = qw(SKIP FAIL);`) , though same result. What am i missing ?

lib/OpenQA/Task/Git/Clone.pm
r-richardson: went with the oneliner you suggested later on

lib/OpenQA/Task/Git/Clone.pm
```diff
@@ -47,12 +47,31 @@ sub _git_clone_all ($job, $clones) {
 
     # iterate clones sorted by path length to ensure that a needle dir is always cloned after the corresponding casedir
     for my $path (sort { length($a) <=> length($b) || $a cmp $b } keys %$clones) {
-        my $url = $clones->{$path};
+        my $url = $clones->{$path} // '';
         die "Don't even think about putting '..' into '$path'." if $path =~ /\.\./;
+
+        my $server = ($url =~ m/(github|gitlab)/) ? $1 : 'other';
```
perlpunk, line 53: I would not use hardcoded server "types". Just extract the host name from the url.
Theoretically there can be more than one gitlab server, and possible even more than one github server, and also more than one other server.

Also, the url is not always set here. So it would have to be retrieved from the git directory first.

perlpunk: e.g. here is a typical git_clone job: https://openqa.opensuse.org/minion/jobs?id=5087197

perlpunk:
```perl
my $git = OpenQA::Git->new(app => $app, dir => $path);
$url //= $git->get_origin_url;

...
# later
_git_clone($git, $job, $ctx, $url);
```

and then remove the `get_origin_url` from `_git_clone`

t/14-grutasks-git.t
```diff
@@ -448,12 +449,108 @@ subtest 'delete_needles' => sub {
...
         $error = $res->{result}->{errors}->[0];
         like $error->{message}, qr{Another git task for.*fedora.*is ongoing}, 'expected error message';
     };
+
+};
+
+subtest 'ServerAvailability.pm: no lock file => skip' => sub {
+    my $tmpdir = tempdir();
+    local $ENV{OPENQA_GIT_SERVER_OUTAGE_FILE} = "$tmpdir/mock";
+    my $outcome = OpenQA::Git::ServerAvailability::report_server_unavailable($t->app, 'gitlab');
```
b10n1k, line 464: Maybe you can export `report_server_unavailable` from `OpenQA::Git::ServerAvailability`?

lib/OpenQA/Task/Git/Clone.pm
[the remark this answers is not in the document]
r-richardson: yes i had that there during my first test though, i found out was i was missing was `use Exporter qw(import);
`

okurz (changes requested):
Missing patch coverage

lib/OpenQA/Task/Git/Clone.pm
```diff
@@ -47,11 +48,42 @@ sub _git_clone_all ($job, $clones) {
 
     # iterate clones sorted by path length to ensure that a needle dir is always cloned after the corresponding casedir
     for my $path (sort { length($a) <=> length($b) || $a cmp $b } keys %$clones) {
-        my $url = $clones->{$path};
+        my $url = $clones->{$path} // '';
         die "Don't even think about putting '..' into '$path'." if $path =~ /\.\./;
+
+        my $git = OpenQA::Git->new(app => $app, dir => $path);
```
perlpunk, line 54: You create a `$git` object here, and you retrieve the url.

Then you can pass that object to `_git_clone` and don't recreate it there again.
Also the url is now set in every case `_git_clone` is called, so I would remove the `if ($url)` there.

lib/OpenQA/Task/Git/Clone.pm
```diff
@@ -47,11 +48,42 @@ sub _git_clone_all ($job, $clones) {
 
     # iterate clones sorted by path length to ensure that a needle dir is always cloned after the corresponding casedir
     for my $path (sort { length($a) <=> length($b) || $a cmp $b } keys %$clones) {
-        my $url = $clones->{$path};
+        my $url = $clones->{$path} // '';
```
perlpunk, line 51: That will ensure that `$url` is always defined (but might be the empty string).

Below you use
```
$url //= $git->get_origin_url;
```
which evaluates the right hand side only if `$url` is **not defined**, but it is never undefined because of this line here.

perlpunk (changes requested):

lib/OpenQA/Git/ServerAvailability.pm
```diff
@@ -0,0 +1,42 @@
...
+use constant MAX_DURATION_SECONDS => ($ENV{OPENQA_GIT_SERVER_OUTAGE_DURATION} || (ONE_HOUR / 2));
+
+use constant {
+    SKIP => 'SKIP',
+    FAIL => 'FAIL',
+};
+
+sub _state_file ($app, $server_name) {
+    return path(
+        ($ENV{OPENQA_GIT_SERVER_OUTAGE_FILE} // ($app->config->{global}->{base_dir} // prjdir) . '/git_server_outage')
```
perlpunk, line 23: Where does `$app->config->{global}->{base_dir}` come from? I can't find such a setting anywhere

perlpunk: What does happen if the directory does not exist on the server?

okurz, line 23: And use `path($foo, $bar)` to concatenate a valid path. See https://docs.mojolicious.org/Mojo/File#path

lib/OpenQA/Task/Git/Clone.pm
```diff
@@ -47,11 +48,42 @@ sub _git_clone_all ($job, $clones) {
...
-        try { _git_clone($app, $job, $ctx, $path, $url); next; }
+        try {
+            _git_clone($app, $job, $ctx, $path, $url);
+            report_server_available($app, $server_host);
+            next;
+        }
         catch ($e) { $error = $e }
+        if ($error =~ /Internal API unreachable/) {
+            my $outcome = report_server_unavailable($app, $server_host);
+
```
okurz, line 77:
```suggestion
```

lib/OpenQA/Git/ServerAvailability.pm
r-richardson: oh yes i wanted to remove this, that was a guess on my side, not sure is basedir also a ini setting, if yes whats the easiest way to retrieve it? Changing this to `$ENV{OPENQA_BASEDIR}` for now.

perlpunk:
for the tests I various suggestions to make it more compact. As it doesn't feel feasible to make several small suggestions, here is a diff:
```
diff --git a/t/14-grutasks-git.t b/t/14-grutasks-git.t
index 8d46a7037..5a2deadfb 100644
--- a/t/14-grutasks-git.t
+++ b/t/14-grutasks-git.t
@@ -457,99 +457,87 @@ subtest 'delete_needles' => sub {
     };
 };
 
-subtest 'ServerAvailability.pm: no lock file => skip' => sub {
+subtest ServerAvailability => sub {
     my $tmpdir = tempdir();
     local $ENV{OPENQA_GIT_SERVER_OUTAGE_FILE} = "$tmpdir/mock";
-    my $outcome = report_server_unavailable($t->app, 'gitlab');
-    is($outcome, 'SKIP', 'Skips when lock file for gitlab is absent');
-};
 
-subtest 'ServerAvailability.pm: file present but younger than 1800 => skip' => sub {
-    my $tmpdir = tempdir();
-    local $ENV{OPENQA_GIT_SERVER_OUTAGE_FILE} = "$tmpdir/mock";
-    my $mock_file = "$tmpdir/mock.gitlab.lock";
-    path($mock_file)->touch;
-    my $outcome = report_server_unavailable($t->app, 'gitlab');
-    is($outcome, 'SKIP', 'Skips the job for an mtime < 1800');
-};
+    subtest 'no lock file => skip' => sub {
+        my $outcome = report_server_unavailable($t->app, 'gitlab');
+        is $outcome, 'SKIP', 'Skips when lock file for gitlab is absent';
+    };
 
-subtest 'ServerAvailability.pm: file present and older => fail' => sub {
-    my $tmpdir = tempdir();
-    local $ENV{OPENQA_GIT_SERVER_OUTAGE_FILE} = "$tmpdir/mock";
-    my $mock_file = "$tmpdir/mock.gitlab.lock";
-    path($mock_file)->touch;
-    my $old_time = time - 2000;
-    utime($old_time, $old_time, $mock_file)
-      or die "Couldn't change mtime on file: $mock_file - $!";
-    my $outcome = report_server_unavailable($t->app, 'gitlab');
-    is($outcome, 'FAIL', 'Fails the job when mtime >= 1800');
-};
+    subtest 'file present but younger than 1800 => skip' => sub {
+        my $mock_file = $tmpdir->child('mock.gitlab.lock')->touch;
+        my $outcome = report_server_unavailable($t->app, 'gitlab');
+        is $outcome, 'SKIP', 'Skips the job for an mtime < 1800';
+    };
 
-subtest 'ServerAvailability.pm: report_server_available removes existing lock file' => sub {
-    my $tmpdir = tempdir();
-    local $ENV{OPENQA_GIT_SERVER_OUTAGE_FILE} = "$tmpdir/mock";
-    my $mock_file = "$tmpdir/mock.gitlab.lock";
-    path($mock_file)->touch;
-    ok(-f $mock_file, 'mock.gitlab.lock exists initially');
-    report_server_available($t->app, 'gitlab');
-    ok(!-f $mock_file, 'gitlab lock file was removed');
-};
+    subtest 'file present and older => fail' => sub {
+        my $mock_file = $tmpdir->child('mock.gitlab.lock')->touch;
+        my $old_time = time - 2000;
+        utime($old_time, $old_time, $mock_file)
+          or die "Couldn't change mtime on file: $mock_file - $!";
+        my $outcome = report_server_unavailable($t->app, 'gitlab');
+        is $outcome, 'FAIL', 'Fails the job when mtime >= 1800';
+    };
 
-subtest 'ServerAvailability.pm: multiple servers do not conflict' => sub {
-    my $tmpdir = tempdir();
-    local $ENV{OPENQA_GIT_SERVER_OUTAGE_FILE} = "$tmpdir/mock";
-    my $outcome_gitlab = report_server_unavailable($t->app, 'gitlab');
-    is($outcome_gitlab, 'SKIP', 'No gitlab lock => skip');
-    my $gitlab_file = "$tmpdir/mock.gitlab.lock";
-    ok(-f $gitlab_file, 'Created mock.gitlab.lock');
-    my $outcome_github = report_server_unavailable($t->app, 'github');
-    is($outcome_github, 'SKIP', 'No github lock => skip');
-    my $github_file = "$tmpdir/mock.github.lock";
-    ok(-f $github_file, 'Created mock.github.lock');
-
-    my $old_time = time - 2000;
-    utime($old_time, $old_time, $github_file);
-    # Re-check
-    $outcome_gitlab = report_server_unavailable($t->app, 'gitlab');
-    $outcome_github = report_server_unavailable($t->app, 'github');
-    is($outcome_gitlab, 'SKIP', 'GitLab file still fresh => skip');
-    is($outcome_github, 'FAIL', 'GitHub file older => fail');
-};
+    subtest 'report_server_available removes existing lock file' => sub {
+        my $mock_file = $tmpdir->child('mock.gitlab.lock')->touch;
+        ok -f $mock_file, 'mock.gitlab.lock exists initially';
+        report_server_available($t->app, 'gitlab');
+        ok !-f $mock_file, 'gitlab lock file was removed';
+    };
 
-subtest 'Clone.pm: Internal API unreachable => skip' => sub {
-    my $tmpdir = tempdir();
-    local $ENV{OPENQA_GIT_SERVER_OUTAGE_FILE} = "$tmpdir/mock";
-    my $clone_mock = Test::MockModule->new('OpenQA::Task::Git::Clone');
-    $clone_mock->redefine(_git_clone => sub { die "Internal API unreachable\n" });
-    my $t = Test::Mojo->new('OpenQA::WebAPI');
-    my @gru_args = ($t->app, 'git_clone', {priority => 10});
-    my $res;
-    combined_like {
-        $res = run_gru_job(@gru_args);
-    }
-    qr/Git server outage likely, skipping clone job/i, 'Minion clone job skipped due to short Git server outage';
-    is($res->{state}, 'finished', 'Job ended in "finished" state');
-    is($res->{result}, 'Job successfully executed', 'Job result indicates success (skip)');
-    done_testing();
-};
+    subtest 'multiple servers do not conflict' => sub {
+        my $outcome_gitlab = report_server_unavailable($t->app, 'gitlab');
+        is $outcome_gitlab, 'SKIP', 'No gitlab lock => skip';
+        my $gitlab_file = "$tmpdir/mock.gitlab.lock";
+        ok -f $gitlab_file, 'Created mock.gitlab.lock';
+        my $outcome_github = report_server_unavailable($t->app, 'github');
+        is $outcome_github, 'SKIP', 'No github lock => skip';
+        my $github_file = "$tmpdir/mock.github.lock";
+        ok -f $github_file, 'Created mock.github.lock';
+
+        my $old_time = time - 2000;
+        utime($old_time, $old_time, $github_file);
+        # Re-check
+        $outcome_gitlab = report_server_unavailable($t->app, 'gitlab');
+        $outcome_github = report_server_unavailable($t->app, 'github');
+        is $outcome_gitlab, 'SKIP', 'GitLab file still fresh => skip';
+        is $outcome_github, 'FAIL', 'GitHub file older => fail';
+    };
 
-subtest 'Clone.pm: Internal API unreachable => fail' => sub {
-    my $tmpdir = tempdir();
-    local $ENV{OPENQA_GIT_SERVER_OUTAGE_FILE} = "$tmpdir/mock";
-    path("$tmpdir/mock.other.lock")->touch;
-    my $old_time = time - 2000;    # "older" than 1800s
-    utime($old_time, $old_time, "$tmpdir/mock.other.lock");
-    my $clone_mock = Test::MockModule->new('OpenQA::Task::Git::Clone');
-    $clone_mock->redefine(_git_clone => sub { die "Internal API unreachable\n" });
-    my $t = Test::Mojo->new('OpenQA::WebAPI');
-    my @gru_args = ($t->app, 'git_clone', {priority => 10});
-    my $res;
-    combined_like {
-        $res = run_gru_job(@gru_args);
-    }
-    qr/Prolonged Git server outage, failing the job/i, 'Minion clone job failed due to prolonged Git server outage';
-    is($res->{state}, 'failed', 'Job ended in "failed" state');
-    like($res->{result}, qr/Internal API unreachable/, 'Job result includes original error message');
+    subtest Clone => sub {
+        my $clone_mock = Test::MockModule->new('OpenQA::Task::Git::Clone');
+        $clone_mock->redefine(_git_clone => sub { die "Internal API unreachable\n" });
+        subtest 'Internal API unreachable => skip' => sub {
+            my $t = Test::Mojo->new('OpenQA::WebAPI');
+            my @gru_args = ($t->app, 'git_clone', {priority => 10});
+            my $res;
+            combined_like {
+                $res = run_gru_job(@gru_args);
+            }
+            qr/Git server outage likely, skipping clone job/i, 'Minion clone job skipped due to short Git server outage';
+            is $res->{state}, 'finished', 'Job ended in "finished" state';
+            is $res->{result}, 'Job successfully executed', 'Job result indicates success (skip)';
+            done_testing();
+        };
+
+        subtest 'Internal API unreachable => fail' => sub {
+            $tmpdir->child('mock.other.lock')->touch;
+            my $old_time = time - 2000;    # "older" than 1800s
+            utime($old_time, $old_time, "$tmpdir/mock.other.lock");
+            my $t = Test::Mojo->new('OpenQA::WebAPI');
+            my @gru_args = ($t->app, 'git_clone', {priority => 10});
+            my $res;
+            combined_like {
+                $res = run_gru_job(@gru_args);
+            }
+            qr/Prolonged Git server outage, failing the job/i, 'Minion clone job failed due to prolonged Git server outage';
+            is $res->{state}, 'failed', 'Job ended in "failed" state';
+            like $res->{result}, qr/Internal API unreachable/, 'Job result includes original error message';
+        };
+    };
 };
 
 done_testing();
```

lib/OpenQA/Git/ServerAvailability.pm
perlpunk: My second question is: what happens if this is merged and deployed to o3 and osd and the outage happens, and the `git_server_outage` does not yet exist?

Martchus (commented):
A more concise Git commit message title would be "Ignore short Git service outages".

lib/OpenQA/Git/ServerAvailability.pm
r-richardson: fixed earlier using `$file->dirname->make_path;` thanks again for the help :)

lib/OpenQA/Task/Git/Clone.pm
r-richardson: yes i'm trying to refactor this, however it'll always mess up the `update clones without CASEDIR` commands, it seems to not only be the order but also somewhat different commands then expected (there is for example no clone command in the list of expected commands)

```
    # Subtest: update clones without CASEDIR
        ok 1 - minion job finished
        ok 2 - minion job result indicates success
        ok 3 - 0: get-url
        not ok 4 - 1: check dirty

        #   Failed test '1: check dirty'
        #   at t/14-grutasks-git.t line 254.
        #          got: 'env 'GIT_SSH_COMMAND=ssh -oBatchMode=yes' GIT_ASKPASS= GIT_TERMINAL_PROMPT=false git clone http://osado /opensuse'
        #     expected: 'git -C /opensuse diff-index HEAD --exit-code'
        not ok 5 - 2: default remote

        #   Failed test '2: default remote'
        #   at t/14-grutasks-git.t line 254.
        #          got: 'git -C /opensuse/needles remote get-url origin'
        #     expected: 'env 'GIT_SSH_COMMAND=ssh -oBatchMode=yes' GIT_ASKPASS= GIT_TERMINAL_PROMPT=false git ls-remote --symref http://osado HEAD'
        not ok 6 - 3: current branch

        #   Failed test '3: current branch'
        #   at t/14-grutasks-git.t line 254.
        #          got: 'env 'GIT_SSH_COMMAND=ssh -oBatchMode=yes' GIT_ASKPASS= GIT_TERMINAL_PROMPT=false git clone http://osado /opensuse/needles'
        #     expected: 'git -C /opensuse branch --show-current'
        not ok 7 - 4: fetch default 
```

okurz:
@r-richardson patch is still incomplete. Did you run tests locally to ensure the necessary lines are covered?

lib/OpenQA/Task/Git/Clone.pm
```diff
@@ -47,11 +48,41 @@ sub _git_clone_all ($job, $clones) {
...
     # iterate clones sorted by path length to ensure that a needle dir is always cloned after the corresponding casedir
     for my $path (sort { length($a) <=> length($b) || $a cmp $b } keys %$clones) {
-        my $url = $clones->{$path};
+        my $url = $clones->{$path} // '';
         die "Don't even think about putting '..' into '$path'." if $path =~ /\.\./;
+
+        my $git = OpenQA::Git->new(app => $app, dir => $path);
+        $url //= $git->get_origin_url;
+        my $server_host;
+        if ($url =~ m{^(?:\w+://)?([^/\@:\s]+)(?::\d+)?/}) {
```
b10n1k, line 57: to me, looks like you dont need the `(?::\d+)?/`. or is there any particular reason?

lib/OpenQA/Task/Git/Clone.pm
```diff
@@ -47,11 +48,41 @@ sub _git_clone_all ($job, $clones) {
...
+            # e.g. "https://gitlab.suse.de/qa/foo", "ssh://git.foo.org/repo"
+            $server_host = $1;
+        }
+        elsif ($url =~ m{^(?:[^@]+@)?([^:]+):}) {
+            # e.g. "git@gitlab.suse.de:qa/foo", "user@host:repo"
+            $server_host = $1;
```
b10n1k, line 63: Can this be in one regex?

perlpunk:
Viewing diffs in the codecov report is currently broken for me in firefox and chromium.
When clicking on the expanding arrow, it takes several seconds, and then I get "Bad Request".
However, we still have the Devel::Cover report. You can click on the "codecov" github action linking to CircleCI, and there you go to "Artifacts" and then `coverage.html`

lib/OpenQA/Task/Git/Clone.pm
r-richardson: yes :+1:  now using `Mojo::URL->host` for the http/ssh part and only one regex just for scp-like urls, as previously suggested by tina

lib/OpenQA/Task/Git/Clone.pm
r-richardson: Yeah not sure it was really needed here but it was meant to optionally match a port number. Now this regex is replaced by `Mojo::URL->host` though

lib/OpenQA/Git/ServerAvailability.pm
```diff
@@ -0,0 +1,39 @@
...
+
+use constant MAX_DURATION_SECONDS => ($ENV{OPENQA_GIT_SERVER_OUTAGE_DURATION} || (ONE_HOUR / 2));
+
+use constant {
+    SKIP => 'SKIP',
+    FAIL => 'FAIL',
+};
+
+sub _state_file ($app, $server_name) {
+    return path(($ENV{OPENQA_GIT_SERVER_OUTAGE_FILE} // prjdir . '/git_server_outage') . ".$server_name.lock");
```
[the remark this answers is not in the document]
r-richardson: now using `.flag`

t/14-grutasks-git.t
```diff
@@ -448,14 +449,99 @@ subtest 'delete_needles' => sub {
...
+        $outcome_github = report_server_unavailable($t->app, 'github');
+        is $outcome_gitlab, 'SKIP', 'GitLab file still fresh => skip';
+        is $outcome_github, 'FAIL', 'GitHub file older => fail';
+    };
+
+    subtest Clone => sub {
+        my $clone_mock = Test::MockModule->new('OpenQA::Task::Git::Clone');
+        $clone_mock->redefine(_git_clone => sub { die "Internal API unreachable\n" });
+        subtest 'Internal API unreachable => skip' => sub {
+            my $t = Test::Mojo->new('OpenQA::WebAPI');
```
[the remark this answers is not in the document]
r-richardson: removed

t/14-grutasks-git.t
```diff
@@ -448,14 +449,99 @@ subtest 'delete_needles' => sub {
...
+            is $res->{state}, 'finished', 'Job ended in "finished" state';
+            is $res->{result}, 'Job successfully executed', 'Job result indicates success (skip)';
+            done_testing();
+        };
+
+        subtest 'Internal API unreachable => fail' => sub {
+            $tmpdir->child('mock.other.lock')->touch;
+            my $old_time = time - 2000;    # "older" than 1800s
+            utime($old_time, $old_time, "$tmpdir/mock.other.lock");
+            my $t = Test::Mojo->new('OpenQA::WebAPI');
```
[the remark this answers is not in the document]
r-richardson: removed

lib/OpenQA/Git/ServerAvailability.pm
```diff
@@ -0,0 +1,40 @@
...
+
+sub report_server_available ($app, $server_name) {
+    _state_file($app, $server_name)->remove;
+}
+
+sub report_server_unavailable ($app, $server_name) {
+    my $file = _state_file($app, $server_name);
+    my $stat = $file->stat;
+    unless (defined $stat) {
+        #$file->dirname->make_path;
```
Martchus, line 32:
```suggestion
```

Martchus (changes requested):
It would still make sense to change the commit message like I suggested before.

okurz:
the patch coverage is still incomplete. I don't see how it makes sense to even push updates then.

lib/OpenQA/Git/ServerAvailability.pm
perlpunk: Why? Did I miss something?
I mean, why was it commented out and why is it removed completely now?
@r-richardson did you adjust the test, like I suggested, so that it will actually reproduce that the `make_path` is necessary?

t/14-grutasks-git.t
```diff
@@ -448,14 +452,102 @@ subtest 'delete_needles' => sub {
...
         $res = run_gru_job(@gru_args);
         is $res->{state}, 'finished', 'job finished';
         $error = $res->{result}->{errors}->[0];
         like $error->{message}, qr{Another git task for.*fedora.*is ongoing}, 'expected error message';
     };
 };
 
+subtest ServerAvailability => sub {
+    my $tmpdir = tempdir();
+    local $ENV{OPENQA_GIT_SERVER_OUTAGE_FILE} = "$tmpdir/mock";
```
perlpunk, line 465: Make that a nested subdir to actually prove that the make_path is necessary

perlpunk (dismissed):

lib/OpenQA/Git/ServerAvailability.pm
perlpunk: > What does happen if the directory does not exist on the server?

Why is this resolved if the `make_path` is still not there?

lib/OpenQA/Git/ServerAvailability.pm
Martchus: I just suggested removing it because it looked like a leftover. If it is needed then enable the code and adjust tests to show that it is necessary.

t/14-grutasks-git.t
```diff
@@ -448,14 +452,102 @@ subtest 'delete_needles' => sub {
...
+
+    subtest Clone => sub {
+        my $clone_mock = Test::MockModule->new('OpenQA::Task::Git::Clone');
+        $clone_mock->redefine(_git_clone => sub { die "Internal API unreachable\n" });
+        subtest 'Internal API unreachable => skip' => sub {
+            my @gru_args = (
+                $t->app,
+                'git_clone',
+                {
+                    priority => 10,
```
perlpunk, line 521: What does the priority have to do here?
This should be a hash with directory names as keys and urls (or undef) as values.

t/14-grutasks-git.t
```diff
@@ -448,14 +452,102 @@ subtest 'delete_needles' => sub {
...
+              'Minion clone job skipped due to short Git server outage';
+            is $res->{state}, 'finished', 'Job ended in "finished" state';
+            is $res->{result}, 'Job successfully executed', 'Job result indicates success (skip)';
+        };
+
+        subtest 'Internal API unreachable => fail' => sub {
+            $tmpdir->child('mock.other.flag')->touch;
+            my $old_time = time - 2000;    # "older" than 1800s
+            utime($old_time, $old_time, "$tmpdir/mock.other.flag");
+            my @gru_args = ($t->app, 'git_clone', {priority => 10});
```
perlpunk, line 538: `priority` is a weird path, and `10` is a weird git url.
It works because of the fallback to `other`, if it is not http nor ssh, but we should not write confusing tests.

t/14-grutasks-git.t
```diff
@@ -448,14 +452,102 @@ subtest 'delete_needles' => sub {
...
+    subtest Clone => sub {
+        my $clone_mock = Test::MockModule->new('OpenQA::Task::Git::Clone');
+        $clone_mock->redefine(_git_clone => sub { die "Internal API unreachable\n" });
+        subtest 'Internal API unreachable => skip' => sub {
+            my @gru_args = (
+                $t->app,
+                'git_clone',
+                {
+                    priority => 10,
+                    clones => 'git@gitlab.suse.de:qa/foo',
```
perlpunk, line 522: Please use something that looks like an actual directory name, preferrable an absolute path.

t/14-grutasks-git.t
```diff
@@ -448,14 +452,101 @@ subtest 'delete_needles' => sub {
...
+
+    subtest Clone => sub {
+        my $clone_mock = Test::MockModule->new('OpenQA::Task::Git::Clone');
+        $clone_mock->redefine(_git_clone => sub { die "Internal API unreachable\n" });
+        subtest 'Internal API unreachable => skip' => sub {
+            my @gru_args = (
+                $t->app,
+                'git_clone',
+                {
+                    clones => {'/my/mock/path' => 'git@gitlab.suse.de:qa/foo'},
```
perlpunk, line 521:
```suggestion
                    '/my/mock/path' => 'git@gitlab.suse.de:qa/foo',
```

lib/OpenQA/Task/Git/Clone.pm
```diff
@@ -47,11 +48,39 @@ sub _git_clone_all ($job, $clones) {
...
+        my $server_host;
+
+        if ($url =~ m{^[^@]+@([^:]+):}) {
+            # e.g. "git@gitlab.suse.de:qa/foo", "user@host:repo"
+            $server_host = $1;
+        }
+        else {
+            my $mojo_url = Mojo::URL->new($url);
+            $server_host = $mojo_url->host // 'other';
+        }
```
Martchus, line 65: This code is not fully covered and should probably be moved into its own function, e.g. `_server_host_from_git_remote`.

t/14-grutasks-git.t
r-richardson: sorry i assumed this was needed as all other gru_args defintions within the file had this set as well.

lib/OpenQA/Task/Git/Clone.pm
```diff
@@ -85,16 +104,16 @@ sub _git_clone ($app, $job, $ctx, $path, $url) {
         # An initial clone fetches all refs, we are done
         return $git->clone_url($url) unless -d $path;
 
-        my $origin_url = $git->get_origin_url;
         if ($url ne $origin_url) {
+            #my $git_path = $git->dirname;
```
okurz, line 108: remove

okurz (changes requested):

okurz (approved):

Martchus (approved):

PR #6303, opened 2025-03-19
Consistently use try/catch instead of 'if ' everywhere

okurz:


perlpunk:
Most of those files were already done...

okurz:
> Most of those files were already done...

I know, I crosschecked the style check against an older state of the repo. Standby for the remaining changes to fulfill the newly added style check

script/modify_needle
```diff
@@ -31,14 +31,15 @@ print help
...
 use Mojo::Base -strict, -signatures;
 use Mojo::File qw(path);
 use Cpanel::JSON::XS ();
 use Getopt::Long;
 
 sub usage ($r) {
-    eval "use Pod::Usage; pod2usage($r);";
-    die "cannot display help, install perl(Pod::Usage)\n" if $@;    # uncoverable statement
+    try { use Pod::Usage; pod2usage($r) }
+    catch ($e) { die "cannot display help, install perl(Pod::Usage)\n" }
```
perlpunk, line 42: Actually, we don't need the try/catch or eval here, see a0a409dbbcc73646f3d9d9551bd71de940b7bf0f
I don't know why I missed that single script

lib/OpenQA/Worker/Job.pm
```diff
@@ -1300,13 +1300,11 @@ sub _ignore_known_files {
...
 sub _parse_log_file ($self, $match_string) {
-    my $msg = '';
     try {
         map_file my $log_content, path($self->worker->pool_directory, 'autoinst-log.txt'), '<';
-        $msg = ": $1" if ($log_content =~ /($match_string)/);
+        return ": $1" if ($log_content =~ /($match_string)/);
     }
     catch ($e) { log_warning($e) }
-    return $msg;
 }
```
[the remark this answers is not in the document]
okurz: done

script/modify_needle
okurz: done

lib/OpenQA/Worker/Job.pm
```diff
@@ -1300,13 +1300,12 @@ sub _ignore_known_files {
...
 sub _parse_log_file ($self, $match_string) {
-    my $msg = '';
     try {
         map_file my $log_content, path($self->worker->pool_directory, 'autoinst-log.txt'), '<';
-        $msg = ": $1" if ($log_content =~ /($match_string)/);
+        return ": $1" if ($log_content =~ /($match_string)/);
     }
     catch ($e) { log_warning($e) }
-    return $msg;
+    return undef;
```
perlpunk, line 1308: Before it returned the empty string, now it's returning undef. That's what I meant with "This might now return undef".
There are cases where the caller expects a defined string.
```suggestion
    return '';
```

perlpunk (changes requested):

script/modify_needle
```diff
@@ -31,15 +31,13 @@ print help
 
 =cut
 
+use Feature::Compat::Try;
```
perlpunk, line 34:
```suggestion
```

t/lib/OpenQA/Test/Database.pm
```diff
@@ -63,9 +63,9 @@ sub insert_fixtures ($self, $schema, $fixtures_glob = '*.pl') {
     my %ids;
     foreach my $fixture (glob "$fixtures_glob") {
 
-        my $info = eval path($fixture)->slurp;    ## no critic
```
perlpunk, line 66: That's a string eval and needs to stay.

okurz: Do you have an idea how we could get rid of the `if $@` in the next line then? Something like combining eval+try together?

lib/OpenQA/WebAPI/Controller/API/V1/Comment.pm
```diff
@@ -228,8 +235,11 @@ sub update ($self) {
       unless ($comment->user_id == $self->current_user->id);
     my $txn_guard = $self->schema->txn_scope_guard;
     my $res = $comment->update({text => href_to_bugref($text)});
-    eval { $res->handle_special_contents($self) };
-    return $self->render(json => {error => $@}, status => 400) if $@;
+    try { $res->handle_special_contents($self) }
+    catch ($e) {
+        undef $txn_guard;
+        return $self->render(json => {error => $e}, status => 400) if $e;
```
b10n1k, line 241: I guess no need for `if` here!

t/01-style.t
```diff
@@ -22,6 +22,7 @@ is qx{git grep -I -l '^use Test::\\(Exception\\|Fatal\\)' t/**.t}, '',
 is qx{git grep -I -l '^\\(throws\\|dies\\|lives\\)_ok.*\<sub\>' t/**.t}, '',
   'Only use simplified prototyped Test::Exception functions';
 is qx{git grep -I -l 'like.*\$\@' t/**.t}, '', 'Use throws_ok instead of manual checks of exceptions';
+is qx{git grep -I -l ' \\(if\\|unless\\) \$\@'}, '', 'Use try/catch instead of manual \$\@ checks';
```
b10n1k, line 25: is there an empty space in the beginning of ' \\(if\\|unless\\) \$\@' intentionally or...?

b10n1k (changes requested):

t/lib/OpenQA/Test/Database.pm
perlpunk: Two ideas:

```
chdir $cwd, croak "Could not insert fixture $fixture: $@" if ${@};   # ;-)
```

```
use Englsh;

chdir $cwd, croak "Could not insert fixture $fixture: $EVAL_ERROR" if $EVAL_ERROR;
```

t/01-style.t
okurz: yes. Alternatively we could use `\<` to look for a word beginning but I wanted to look for the post-conditions only. I will consider if we generalize that to `\(if\|unless\).*$@`

t/lib/OpenQA/Test/Database.pm
okurz: Hm. How about a comment `# nostyle:bare-eval` and in our style check we only complain about `if $@` with a negative look-ahead for that special comment?

perlpunk: I would be ok with that.
I just thought that using `$EVAL_ERROR` would be a form of "I know what I'm doing, I cannot use try/catch here".

okurz: yeah, I guess that works. Updated

lib/OpenQA/WebAPI/Controller/Admin/Needle.pm
```diff
@@ -72,16 +73,16 @@ sub ajax ($self) {
...
             push(@filter_conds, {last_seen_time => _translate_cond($seen_query)});
         }
         my $match_query = $self->param('last_match');
         if ($match_query && $match_query ne 'none') {
             push(@filter_conds, {last_matched_time => _translate_cond($match_query)});
         }
-    };
-    return $self->render(json => {error => ($@ =~ s/ at .*//sr)}, status => 400) if $@;
+    }
+    catch ($e) { return $self->render(json => {error => ($@ =~ s/ at .*//sr)}, status => 400) }
```
perlpunk, line 85:
```suggestion
    catch ($e) { return $self->render(json => {error => ($e =~ s/ at .*//sr)}, status => 400) }
```

t/02-pod.t
```diff
@@ -1,8 +1,9 @@
 use Test::Most;
 use Test::Warnings qw(:no_end_test :report_warnings);
+use Feature::Compat::Try;
 # no OpenQA::Test::TimeLimit for this trivial test
 
-eval 'use Test::Pod';
-plan skip_all => "Test::Pod required for testing POD" if $@;
+try { use Test::Pod }
```
perlpunk, line 6: That will execute `use Test::Pod` at compile time.
```suggestion
try { require Test::Pod; import Test::Pod }
```
Alternatively leave the `import` out and call `Test::Pod::all_pod_files_ok()` below

perlpunk: Another alternative:
```
eval 'use Test::Pod; 1' or plan skip_all => "Test::Pod required for testing POD";
```

t/lib/OpenQA/Test/Database.pm
```diff
@@ -63,9 +64,8 @@ sub insert_fixtures ($self, $schema, $fixtures_glob = '*.pl') {
...
-        chdir $cwd, croak "Could not insert fixture $fixture: $@" if $@;
-
+        my $info = eval path($fixture)->slurp;
```
b10n1k, line 67: should the `## no critic` stays?

okurz: No. Not needed anymore (apparently)

t/02-pod.t
```diff
@@ -1,8 +1,8 @@
...
 use Test::Warnings qw(:no_end_test :report_warnings);
+use Feature::Compat::Try;
 # no OpenQA::Test::TimeLimit for this trivial test
 
-eval 'use Test::Pod';
-plan skip_all => "Test::Pod required for testing POD" if $@;
+eval 'use Test::Pod; 1' or plan skip_all => "Test::Pod required for testing POD";
 
-all_pod_files_ok();
+Test::Pod::all_pod_files_ok();
```
perlpunk, line 8: If you use `use` above, then the function is imported, so no need to use the full name:
```suggestion
all_pod_files_ok();
```

perlpunk (approved):

PR #6313, opened 2025-03-21
Allow web proxy to reach openQA on SELinux systems

asdil12:
Ticket: https://progress.opensuse.org/issues/178822


asdil12:
image

Martchus:
But the style check fails:

```
-if which semanage &>/dev/null; then
+if which semanage &> /dev/null; then
```

PR #6314, opened 2025-03-21
Fix config drop-in path to upstream specifications

okurz:
See
* https://uapi-group.org/specifications/specs/configuration_files_specification/
* https://en.opensuse.org/openSUSE:Packaging_UsrEtc

The .d/ directory must be in the same directory as the .ini/.conf file
and use the complete name of the overridden config file path including
extension.

As the drop-in config directory is not documented yet and only has
implementation we should be ok to just change that :)

Related progress issue: https://progress.opensuse.org/issues/179359


okurz:
@Vogtinator your feedback would be appreciated

Vogtinator (approved):
LGTM

PR #6315, opened 2025-03-21
Update url query only with the selected option

b10n1k:
dashboard provides filter for limits and tags.
`url_with` extends the query when options are selected. Which creates a long queries like `?limit_builds=20&only_tagged=1&only_tagged=0&limit_builds=10`. Ideally it should have unique entries with clear a clear query as described in the documentation https://docs.mojolicious.org/Mojolicious/Controller#url_for and seems like it needs a hash instead of a list. This updates the current request as expected.

### DIFF
--- a/t/ui/14-dashboard.t
+++ b/t/ui/14-dashboard.t
@@ -94,13 +94,22 @@ is(
 
 $driver->find_element_by_link_text('400')->click();
 is($driver->find_element('#more_builds b')->get_text(), 400, 'limited to the selected number');
+like($driver->get_current_url(), qr/1001\?limit_builds=400/, 'url shows the selected build_limits');
+$driver->find_element_by_link_text('50')->click();
+my $actual_url = $driver->get_current_url();
+unlike($actual_url, qr/limit_builds=400/, 'url query doesnt show previous limit_builds');
+like($actual_url, qr/1001\?limit_builds=50/, 'url query has unique limit_builds entry');
 $driver->find_element_by_link_text('tagged')->click();
 is(scalar @{$driver->find_elements('.h4', 'css')}, 0, 'no tagged builds exist');
+$driver->find_element_by_link_text('400')->click();
+$actual_url = $driver->get_current_url();
+unlike($actual_url, qr/limit_builds=50/, 'url query updates only limit_builds entry and old selection is removed');
+like($actual_url, qr/limit_builds=400/, 'url query updates only limit_builds entry and shows new selection');
+like($actual_url, qr/only_tagged=1/, 'url query contains tag settings');
 
 $driver->get('/group_overview/1001');
 my $res = OpenQA::Test::Case::trim_whitespace($driver->find_element_by_id('more_builds')->get_text);
 is($res, q{Limit to 10 / 20 / 50 / 100 / 400 builds, only tagged / all}, 'more builds can be requested');
-
 is($driver->get('/?group=opensuse'), 1, 'group parameter is not exact by default');
 wait_for_ajax;
 is(scalar @{$driver->find_elements('h2', 'css')}, 2, 'both job groups shown');
--- a/t/ui/14-dashboard.t
+++ b/t/ui/14-dashboard.t
@@ -28,7 +28,7 @@ plan skip_all => 'Install Selenium::Remote::WDKeys to run this test'
 #
 $driver->title_is("openQA", "on main page");
 my @build_headings = $driver->find_elements('.h4', 'css');
-is(scalar @build_headings, 4, '4 builds shown');
+is scalar @build_headings, 4, '4 builds shown';
 
 subtest 'Back to top button' => sub {
     my $check_visibility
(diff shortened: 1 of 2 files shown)


okurz (changes requested):
Where is the test showing the effect?

b10n1k:
> Where is the test showing the effect?

it should have been there. 😇

t/ui/14-dashboard.t
```diff
@@ -94,13 +94,25 @@ is(
 
 $driver->find_element_by_link_text('400')->click();
 is($driver->find_element('#more_builds b')->get_text(), 400, 'limited to the selected number');
+like($driver->get_current_url(), qr/1001\?limit_builds=400/, 'url shows the selected build_limits');
+$driver->find_element_by_link_text('50')->click();
+like(
+    $driver->get_current_url(),
+    qr/1001\?(?!.*limit_builds&)limit_builds=50/,
```
Martchus, line 101: I *guess* this is the correct syntax. When I would have to write this check myself I'd probably just used `unlike` and `like` (so two separate checks) for what's supposed to be absent and what's supposed to be present. That would be much clearer as it wouldn't involve so much regex black-magic.

It would probably also make sense to check whether a previously selected option (e.g. `only_tagged=1`) is still present after using the pagination because your change *possibly* breaks that.

b10n1k: Starting from the second part: is this covered https://github.com/os-autoinst/openQA/pull/6315/files#diff-9ad9a55405b1ec79d25f3b32f354896690e22a296b04a1e42b240a1b48a72763R109?

`1001\?(?!.*limit_builds&)limit_builds=50` says 

`1001\?limit_builds=50` is present and it is query is not extended. the combination of unlike+like sounds more complicated to me. in additional, more check, less faster (even is so trivial in general). it would look like, right?

```
unlike $driver->get_current_url(), qr/1001\?limit_builds=400&limit_builds=50/;
like $driver->get_current_url(), qr/1001\?limit_builds=50/;
```

Martchus: Right, the check for another option like `only_tagged=1` is indeed covered by the diff/line you've mentioned.

---

Yes, it would look like that except that I'd remove the `limit_builds=50` part from the `unlike` check.

If you don't invoke `$driver->get_current_url()` twice (but save that in a variable) the shouldn't be much of a performance overhead due to checking this twice. And in any case the performance difference is negligible.

perlpunk: Why `limit_builds&`? `limit_builds` without a `=value` would not appear in the query anyway, would it?

perlpunk:
```perl
my @urls = (
    "/1001?limit_builds=0&limit_builds=400&limit_builds=50",
    "/1001?limit_builds=50",
    "/1001?xxx=yyy&limit_builds=50",
);
for (@urls) {
    say "==== $_ ====";
    say m/1001\?(?!.*limit_builds&)limit_builds=50/ ? "match" : "no match"
}
__END__
==== /1001?limit_builds=0&limit_builds=400&limit_builds=50 ====
no match
==== /1001?limit_builds=50 ====
match
==== /1001?xxx=yyy&limit_builds=50 ====
no match
```
So basically your regex is ensuring the same as `m/1001\?limit_builds=50/`

b10n1k: nice catch @perlpunk . Lets do it as @Martchus suggested anyway

t/ui/14-dashboard.t
```diff
@@ -94,13 +94,23 @@ is(
...
+$driver->find_element_by_link_text('50')->click();
+my $actual_url = $driver->get_current_url();
+unlike($actual_url, qr/limit_builds=400/, 'url query doesnt show previous limit_builds');
+like($actual_url, qr/1001\?limit_builds=50/, 'url query has unique limit_builds entry');
 $driver->find_element_by_link_text('tagged')->click();
 is(scalar @{$driver->find_elements('.h4', 'css')}, 0, 'no tagged builds exist');
+$driver->find_element_by_link_text('400')->click();
+like(
+    $driver->get_current_url(),
+    qr/1001\?only_tagged=1&(?!.*limit_builds&)limit_builds=400/,
```
Martchus, line 107: I'm wondering about the middle part here (`(?!.*limit_builds&)`). This is again a bit too much regex magic for my taste and I would use `unlike` as before.

It would also make sense test for `only_tagged=1` separately to make this check not assert the order in which these parameters are present.

t/ui/14-dashboard.t
```diff
@@ -94,13 +94,22 @@ is(
 
 $driver->find_element_by_link_text('400')->click();
 is($driver->find_element('#more_builds b')->get_text(), 400, 'limited to the selected number');
+like($driver->get_current_url(), qr/1001\?limit_builds=400/, 'url shows the selected build_limits');
+$driver->find_element_by_link_text('50')->click();
+my $actual_url = $driver->get_current_url();
+unlike($actual_url, qr/limit_builds=400/, 'url query doesnt show previous limit_builds');
+like($actual_url, qr/1001\?limit_builds=50/, 'url query has unique limit_builds entry');
```
okurz, line 101:
```suggestion
like $driver->get_current_url(), qr/1001\?limit_builds=400/, 'url shows the selected build_limits';
$driver->find_element_by_link_text('50')->click();
my $actual_url = $driver->get_current_url();
unlike $actual_url, qr/limit_builds=400/, 'url query doesnt show previous limit_builds';
like $actual_url, qr/1001\?limit_builds=50/, 'url query has unique limit_builds entry';
```

t/ui/14-dashboard.t
```diff
@@ -94,13 +94,22 @@ is(
...
 $driver->find_element_by_link_text('tagged')->click();
 is(scalar @{$driver->find_elements('.h4', 'css')}, 0, 'no tagged builds exist');
+$driver->find_element_by_link_text('400')->click();
+$actual_url = $driver->get_current_url();
+unlike($actual_url, qr/limit_builds=50/, 'url query updates only limit_builds entry and old selection is removed');
+like($actual_url, qr/limit_builds=400/, 'url query updates only limit_builds entry and shows new selection');
+like($actual_url, qr/only_tagged=1/, 'url query updates only limit_builds entry and shows new selection');
```
okurz, line 108:
```suggestion
$driver->find_element_by_link_text('400')->click();
$actual_url = $driver->get_current_url();
unlike $actual_url, qr/limit_builds=50/, 'url query updates only limit_builds entry and old selection is removed';
like $actual_url, qr/limit_builds=400/, 'url query updates only limit_builds entry and shows new selection';
like $actual_url, qr/only_tagged=1/, 'url query updates only limit_builds entry and shows new selection';
```

and the last two test strings are identical

okurz (changes requested):

t/ui/14-dashboard.t
```diff
@@ -10,7 +10,7 @@ use lib "$FindBin::Bin/../lib", "$FindBin::Bin/../../external/os-autoinst-common
 use Test::Mojo;
 use Test::Warnings ':report_warnings';
 use OpenQA::Log 'log_debug';
-use OpenQA::Test::TimeLimit '30';
+#use OpenQA::Test::TimeLimit '30';
```
okurz, line 13: Revert

okurz (changes requested):
One trivial issue. Rest looks fine to me

t/ui/14-dashboard.t
b10n1k: obviously. sorry

okurz (approved):

Martchus (approved):

PR #6316, opened 2025-03-23
Deduplicate openqa-cli options and help

perlpunk:
Use a spec file to deduplicate the options and help output for openqa-cli.

### Only one place, no duplicated information that can diverge over time

Currently the global options are repeated in several places:
* https://github.com/os-autoinst/openQA/blob/master/lib/OpenQA/Command.pm#L100-L110
* https://github.com/os-autoinst/openQA/blob/master/lib/OpenQA/CLI/api.pm#L116 ff
* https://github.com/os-autoinst/openQA/blob/master/lib/OpenQA/CLI/schedule.pm#L59 ff
* https://github.com/os-autoinst/openQA/blob/master/lib/OpenQA/CLI/monitor.pm#L67 ff
* https://github.com/os-autoinst/openQA/blob/master/lib/OpenQA/CLI/archive.pm#L50 ff
* 

And the options for the subcommands are also duplicated:
* https://github.com/os-autoinst/openQA/blob/master/lib/OpenQA/CLI/api.pm#L21-L28
* https://github.com/os-autoinst/openQA/blob/master/lib/OpenQA/CLI/api.pm#L123 ff
* https://github.com/os-autoinst/openQA/blob/master/lib/OpenQA/CLI/schedule.pm#L34-L37
* https://github.com/os-autoinst/openQA/blob/master/lib/OpenQA/CLI/schedule.pm#L66
* https://github.com/os-autoinst/openQA/blob/master/lib/OpenQA/CLI/monitor.pm#L49-L50
* https://github.com/os-autoinst/openQA/blob/master/lib/OpenQA/CLI/monitor.pm#L77
* https://github.com/os-autoinst/openQA/blob/master/lib/OpenQA/CLI/archive.pm#L16-L17
* https://github.com/os-autoinst/openQA/blob/master/lib/OpenQA/CLI/archive.pm#L55

Some of the options were forgotten in the pod, e.g. archive does not list `--pretty`, `--quiet` and `--verbose`

### Easy shell completion

@b10n1k had the nice idea to add shell completion for openqa-cli (#6063).

However, manually written completion scripts have several disadvantages. Typos, and the content can diverge over time. Duplicated information again.

With the spec file we can get bas/zsh completion practically for free with [appspec](https://metacpan.org/dist/App-AppSpec/view/lib/appspec.pod), including completion descriptions for bash.
We only have to regenerate it when an option gets added/changed, and a test can ensure that we don't forget regeneration.
So appspec would only be a dependency for us openQA authors.


lib/OpenQA/CLI.pm
```diff
@@ -2,62 +2,87 @@
...
+        $help .= "$cmd->{description}\n" if $cmd->{description};
+        $help .= _print_options('Options:', _get_options($name));
+    }
+    return encode 'UTF-8' => $help;
+}
+
+sub _print_options ($label, $options) {
+    my $help = "$label\n";
+    my @rows;
+    my $max = 15;
```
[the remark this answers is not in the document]
okurz: ok. how about a constant called accordingly?

perlpunk:
Not sure how to solve this error:
https://build.opensuse.org/package/live_build_log/devel:openQA:GitHub:os-autoinst:openQA:PR-6316/openQA:openQA-client-test/openSUSE_Tumbleweed/x86_64
```
Could not open '/usr/share/openqa/script/../public/openqa-cli.yaml' for reading: No such file or directory
```
Where should I put a file that the client needs to read?

kalikiana:
> Not sure how to solve this error: https://build.opensuse.org/package/live_build_log/devel:openQA:GitHub:os-autoinst:openQA:PR-6316/openQA:openQA-client-test/openSUSE_Tumbleweed/x86_64
> 
> ```
> Could not open '/usr/share/openqa/script/../public/openqa-cli.yaml' for reading: No such file or directory
> ```
> 
> Where should I put a file that the client needs to read?

How about datadir/openqa/client

b10n1k:
> Not sure how to solve this error: https://build.opensuse.org/package/live_build_log/devel:openQA:GitHub:os-autoinst:openQA:PR-6316/openQA:openQA-client-test/openSUSE_Tumbleweed/x86_64
> 
> ```
> Could not open '/usr/share/openqa/script/../public/openqa-cli.yaml' for reading: No such file or directory
> ```
> 
> Where should I put a file that the client needs to read?

I guess the easiest would be somewhere in lib/openQA as under `/usr/share/openqa` there are only _lib_ and _script_ dirs.. even so _public_ seems it should  be able to be found in `/usr/share/openqa/`!!

perlpunk:
The file will now be installed to `/usr/share/openqa/client/openqa-cli.yaml`. In git it's still under `public/` for now. That's why the code has to check for both locations.
Alternatively we could create a `client/` folder in git.

lib/OpenQA/CLI.pm
```diff
@@ -2,63 +2,94 @@
 # SPDX-License-Identifier: GPL-2.0-or-later
 
 package OpenQA::CLI;
-use Mojo::Base 'Mojolicious::Commands';
+use Mojo::Base 'Mojolicious::Commands', -signatures;
+use Mojo::Util qw(getopt);
+use YAML::PP;
+use FindBin '$RealBin';
+use Mojo::Util qw(getopt);
+use Mojo::Util qw(encode);
```
Martchus:
```suggestion
use Mojo::Util qw(getopt encode);
use YAML::PP;
use FindBin '$RealBin';
```

Martchus (commented):
Looks good I guess. I only have a small nitpick on the `use` lines.

lib/OpenQA/CLI.pm
perlpunk: done

PR #6323, opened 2025-03-24
Make loading nested config files work with our RPM packaging

Martchus:
Make the full `/etc/openqa` directory (where config files are installed to by the `Makefile` install target) available to the Mojo app home directory (where the Mojo app actually looks for the config) like it is done in Debian packaging.

### DIFF
--- a/dist/rpm/openQA.spec
+++ b/dist/rpm/openQA.spec
@@ -382,9 +382,8 @@ export LANG=en_US.UTF-8
 rm -rf %{buildroot}/%{_sysusersdir}
 %endif
 
-mkdir -p %{buildroot}%{_datadir}/openqa%{_sysconfdir}/openqa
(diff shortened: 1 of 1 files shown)


okurz (approved):

Martchus:
I tested installing the package from the OBS checks locally but it looks like the new package cannot be updated to smoothly:

```
error: unpacking of archive failed on file /usr/share/openqa/etc/openqa;67e25f0d: cpio: File from package already exists as a directory in system
error: openQA-5.1742832312.e44eedb2-7769.1.x86_64: install failed
error: openQA-5.1741698447.0e363abf-7721.1.x86_64: erase skipped
```

So perhaps the other PR is the better alternative.

Martchus:
The alternative https://github.com/os-autoinst/openQA/pull/6324 has been merged.

PR #6324, opened 2025-03-24
Make loading nested config files work with our RPM packaging (no symlink)

Martchus:
Alternative to https://github.com/os-autoinst/openQA/pull/6323 that avoids the symlinking completely.

---

* Read config files directly from `/etc/openqa` if the config directory
  doesn't exist under `OPENQA_CONFIG` or the app home
* Keep reading from `OPENQA_CONFIG` and the app home if a config directory
  exists under those locations as this is useful for development setups and
  running unit tests
* Read config files from `/usr/etc/openqa` if no config directory exists
  under any of the other locations to support
  https://en.opensuse.org/openSUSE:Packaging_UsrEtc#Variant_1_(ideal_case)
* Remove symlinks in our packaging so an installation from packages will in
  fact use all the files from `/etc/openqa` or `/usr/etc/openqa` and not be
  stuck with an only partially symlinked config
* See https://progress.opensuse.org/issues/179425

### DIFF
--- a/lib/OpenQA/Setup.pm
+++ b/lib/OpenQA/Setup.pm
@@ -25,6 +25,20 @@ use Feature::Compat::Try;
 my %CARRY_OVER_DEFAULTS = (lookup_depth => 10, state_changes_limit => 3);
 sub carry_over_defaults () { \%CARRY_OVER_DEFAULTS }
 
+sub _lookup_config_files ($app) {
+    my @config_file_paths;
+    for my $path ($ENV{OPENQA_CONFIG}, $app->home->child('etc', 'openqa'), '/etc/openqa', '/usr/etc/openqa') {
+        next unless defined $path && -e $path;
+        my $config_path = path($path);
+        my $main_config_file = $config_path->child('openqa.ini');
+        push @config_file_paths, $main_config_file if -e $main_config_file;
+        push @config_file_paths, @{$config_path->child('openqa.ini.d')->list->grep(qr/\.ini$/)->sort};
+        last;
+    }
+    if (my $log = $app->log) { $app->log->info("Reading config from: @config_file_paths") }
+    return \@config_file_paths;
+}
+
 sub _read_config_file ($config, $config_file, $defaults, $mode_defaults) {
     for my $section (sort keys %$defaults) {
         my $section_defaults = $defaults->{$section};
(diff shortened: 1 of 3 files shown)


kalikiana (approved):
I slightly prefer this over #6323 as it should work consistently without relying on packaging to get this right

lib/OpenQA/Setup.pm
```diff
@@ -48,7 +48,7 @@ sub _read_config_file ($config, $config_file, $defaults, $mode_defaults) {
 sub _load_config ($app, $defaults, $mode_specific_defaults) {
     my $config = $app->config;
     my $mode_defaults = $mode_specific_defaults->{$app->mode} // {};
-    my $config_path = $ENV{OPENQA_CONFIG} ? path($ENV{OPENQA_CONFIG}) : $app->home->child('etc', 'openqa');
+    my $config_path = path($ENV{OPENQA_CONFIG} || '/etc/openqa');
```
okurz, line 51: Looks like the former at least in theory was platform agnostic, e.g. / vs. also \. What would be the more portable approach?

Martchus: Maybe there's a good/maintained library like [QStandardPaths](https://doc.qt.io/qt-6/qstandardpaths.html) for Perl we could use. Otherwise, one can always set `OPENQA_CONFIG`.

Note that our install script / Makefile is also platform-specific. Despite that, this code and the Makefile would probably even work on Windows when using MSYS2.

okurz: Have you found out why originally the change was made to read from $app->home?

Also please keep https://progress.opensuse.org/issues/179359 in mind. I assume it's easy to be solved just looking for additional config files in multiple directories but maybe you see your change as related nevertheless.

Martchus: Ok, I'll keep https://progress.opensuse.org/issues/179359 in mind. So I assigned myself to that ticket as I'll have to do the read-up on that topic anyway.

Martchus: Perhaps the reason to read it indirectly is also the reason why two tests fail now: Tests now cannot find a config file anymore. The same counts for developer setups which no longer just read the config from the Git checkout. This is not a big deal for most tests but the two failing ones fail because logging is not initialized so the warning of the absence of the config file cannot be printed.

I think it makes most sense to do the following:

1. Read from the env variable. It is meant as an override so it should have precedence.
2. Read from the app home directory as this is useful for development setups and unit tests. (I will remove the symlinking from the spec file so the packaged version will not find anything in its home directory and proceed with the next lookup steps.)
3. Read from `/etc/openqa`.
4. Read from `/usr/etc/openqa` to support https://en.opensuse.org/openSUSE:Packaging_UsrEtc#Variant_1_(ideal_case).

Considering we now need a lookup mechanism that checks multiple directories I can implement https://en.opensuse.org/openSUSE:Packaging_UsrEtc#Variant_1_(ideal_case) right away. Additional remarks:

* The lookup stops at the first step where it can find any config file in accordance with the linked Wiki page.
* I think we don't *need* to package anything under `/usr/etc/` because our upstream default config should also be the downstream/packaging default config.
    * Maybe it makes nevertheless sense to do it as we could package the "empty" config file (only containing comments) so far installed under `/etc/openqa` under `/usr/etc/openqa`. Admins would then copy the "empty" config file from `/usr/etc/openqa` to `/etc/openqa` (either fully or partially using `/etc/openqa/openqa.ini.d`).
    * Maybe it makes sense to install the config file we so far installed under `/etc/openqa` under `/usr/etc/skel`.

Martchus: This now takes `/usr/etc` into account. The paths `/etc/…` and `/usr/etc/…` are now mere fallbacks in addition to the lookup we had before. Hence I don't think we need to worry about making this cross-platform (which we don't support anyway).

I haven't changed the location where our default config file is installed as this can be separated and might be something we need to discuss first.

okurz (approved):

nicksinger:
I think this broke openqa.opensuse.org with an automatic upgrade. Afterwards openQA refuses to start because it cannot find its database.ini:

```
Mar 25 10:40:43 ariel systemd[1]: Reloading The openQA web UI...
Mar 25 10:40:43 ariel systemd[1]: Reloaded The openQA web UI.
Mar 25 10:40:45 ariel openqa-webui-daemon[23615]: [2025-03-25 10:40:45.49969] [23615] [info] Reading config from: /etc/openqa/openqa.ini
Mar 25 10:40:45 ariel openqa-webui-daemon[23615]: Failed to open /usr/share/openqa/script/../etc/openqa/database.ini: No such file or directory at /usr/share/openqa/script/../lib/OpenQA/Schema.pm line 44.
Mar 25 10:40:45 ariel openqa-webui-daemon[23615]: Could not find database section 'production' in /usr/share/openqa/script/../etc/openqa/database.ini at /usr/share/openqa/script/../lib/OpenQA/Schema.pm line 45.
Mar 25 10:40:45 ariel systemd[1]: openqa-webui.service: Main process exited, code=exited, status=2/INVALIDARGUMENT
Mar 25 10:40:46 ariel systemd[1]: openqa-webui.service: Failed with result 'exit-code'.

```

Martchus:
I guess the problem was simply that the change also needs to be applied to the database config lookup. Next time I'll add "not-ready" to be able to test the rpm package before merging.

PR #6328, opened 2025-03-25
Make loading nested config files work with RPM packaging (2nd attempt)

Martchus:
* Read config files directly from `/etc/openqa` if the config directory
    doesn't exist under `OPENQA_CONFIG` or the app home
* Keep reading from `OPENQA_CONFIG` and the app home if a config directory
    exists under those locations as this is useful for development setups and
    running unit tests
* Read config files from `/usr/etc/openqa` if no config directory exists
    under any of the other locations to support
    https://en.opensuse.org/openSUSE:Packaging_UsrEtc#Variant_1_(ideal_case)
* Remove symlinks in our packaging so an installation from packages will in
    fact use all the files from `/etc/openqa` or `/usr/etc/openqa` and not be
    stuck with an only partially symlinked config
* See https://progress.opensuse.org/issues/179425

### DIFF
--- a/dist/rpm/openQA.spec
+++ b/dist/rpm/openQA.spec
@@ -382,9 +382,6 @@ export LANG=en_US.UTF-8
 rm -rf %{buildroot}/%{_sysusersdir}
 %endif
 
-mkdir -p %{buildroot}%{_datadir}/openqa%{_sysconfdir}/openqa
-ln -s %{_sysconfdir}/openqa/openqa.ini %{buildroot}%{_datadir}/openqa%{_sysconfdir}/openqa/openqa.ini
-ln -s %{_sysconfdir}/openqa/database.ini %{buildroot}%{_datadir}/openqa%{_sysconfdir}/openqa/database.ini
 mkdir -p %{buildroot}%{_bindir}
 ln -s %{_datadir}/openqa/script/client %{buildroot}%{_bindir}/openqa-client
 ln -s %{_datadir}/openqa/script/openqa-cli %{buildroot}%{_bindir}/openqa-cli
@@ -569,10 +566,6 @@ fi
 %config(noreplace) %attr(-,geekotest,root) %{_sysconfdir}/openqa/openqa.ini
 %config(noreplace) %attr(-,geekotest,root) %{_sysconfdir}/openqa/database.ini
 %dir %{_datadir}/openqa
-%dir %{_datadir}/openqa/etc
-%dir %{_datadir}/openqa%{_sysconfdir}/openqa
-%{_datadir}/openqa%{_sysconfdir}/openqa/openqa.ini
-%{_datadir}/openqa%{_sysconfdir}/openqa/database.ini
 %config %{_sysconfdir}/logrotate.d
 # apache vhost
 %dir %{_sysconfdir}/apache2
--- a/docs/Contributing.asciidoc
+++ b/docs/Contributing.asciidoc
@@ -294,6 +294,10 @@ assets can need a big amount of disk space.
 WARNING: Be sure to *clear* that variable when running unit tests locally.
 
 === Customize configuration directory
+When running openQA from a Git checkout it will find configuration files from
+that checkout under `etc/openqa` and not use any system provided config files
+under e.g. `/etc/openqa`.
+
 It can be necessary during development to change the configuration.
 For example you have to edit `etc/openqa/database.ini` to use another database.
 It can also be useful to set the authentication method to `Fake` and increase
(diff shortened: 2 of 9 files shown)


lib/OpenQA/Config.pm
```diff
@@ -0,0 +1,49 @@
...
+sub lookup_config_files ($home, $name, $silent = 0) {
+    my @config_file_paths;
+    for my $path ($ENV{OPENQA_CONFIG}, $home->child('etc', 'openqa'), '/etc/openqa', '/usr/etc/openqa') {
+        next unless defined $path;
+        my $config_path = path($path);
+        my $main_config_file = $config_path->child($name);
+        my $extension = $main_config_file->extname;
+        my $config_name = $main_config_file->basename(".$extension");
+        push @config_file_paths, $main_config_file if -r $main_config_file;
+        push @config_file_paths, @{$config_path->child("$name.d")->list->grep(qr/\.$extension$/)->sort};
```
okurz, line 23: Does this mean that if I do for testing `OPENQA_CONFIG=/my/dir script/openqa` then still settings from /usr/etc/openqa/openqa.ini.d/secret-hidden-file.ini would be effective if not set in /my/dir/openqa.ini?

Martchus: If you do `OPENQA_CONFIG=/my/dir script/openqa` it depends on the contents of `/my/dir`.

If it does not contain any config file (`openqa.ini` or `openqa.ini.d/*.ini`) the directory is ignored and the next one tried (which could be `/usr/etc/…` if none of the other directories contain any config file).

If it contains any config file then the directory is used and none of the other directories (including `/usr/etc/…`) are used at all.

dist/rpm/openQA.spec
```diff
@@ -569,10 +566,6 @@ fi
 %config(noreplace) %attr(-,geekotest,root) %{_sysconfdir}/openqa/openqa.ini
```
okurz, line 566: How about a comment in this file explaining how to use/override or point to the documentation

dist/rpm/openQA.spec
```diff
@@ -382,9 +382,6 @@ export LANG=en_US.UTF-8
 rm -rf %{buildroot}/%{_sysusersdir}
 %endif
 
-mkdir -p %{buildroot}%{_datadir}/openqa%{_sysconfdir}/openqa
-ln -s %{_sysconfdir}/openqa/openqa.ini %{buildroot}%{_datadir}/openqa%{_sysconfdir}/openqa/openqa.ini
-ln -s %{_sysconfdir}/openqa/database.ini %{buildroot}%{_datadir}/openqa%{_sysconfdir}/openqa/database.ini
```
okurz, line 387: Please test based on RPM packages from https://download.opensuse.org/repositories/devel:/openQA:/GitHub:/os-autoinst:/openQA:/PR-6328/openSUSE_Tumbleweed/devel:openQA:GitHub:os-autoinst:openQA:PR-6328.repo regarding still being able to start the webUI, gru, database connection, worker, client, etc.

docs/Contributing.asciidoc
```diff
@@ -294,6 +294,10 @@ assets can need a big amount of disk space.
 WARNING: Be sure to *clear* that variable when running unit tests locally.
 
 === Customize configuration directory
+When running openQA from a Git checkout it will find configuration files from
```
okurz, line 297: Please add documentation for admins, not just developers. Maybe a simple sentence is enough but in the according section

okurz (changes requested):

dist/rpm/openQA.spec
Martchus: I have already done that so I'm removing the "not-ready" label now. The packages are installable and used the correct config files (including the database and client). The worker also still works (although it is only affected for the client config anyway).

docs/Contributing.asciidoc
Martchus: So far nothing has changed for admins. The config files are still under `/etc`. I will extend the config for admins as soon as we I make changes to install/package config files under `/usr/etc`.

Martchus:
Note that I want to merge this change first before doing any admin facing changes. I hope it will work better then the last attempt. If it works and it is what we want I will move on with changing the install script and packaging to use `/usr/etc` and only then update the documentation for admins. (I'm also still wondering how to change the documentation. There are many explicit mentionings of `/etc/openqa/openqa.ini` which would not exist anymore by default and also should not be created anymore in favor of using a drop-in config. Stating this explicitly on all those occasions would be quite verbose, though.)

Martchus:
I'm adding back the "not-ready" label so this is only deployed on o3 when I can monitor what happens (but I have already tested the packaging locally).

okurz (approved):

kalikiana (approved):

PR #6330, opened 2025-03-26
Update tempdir path to /tmp to cleanup worktree for workers tests

Wabri:
related: https://progress.opensuse.org/issues/178207

### DIFF
--- a/t/24-worker-overall.t
+++ b/t/24-worker-overall.t
@@ -29,6 +29,10 @@ $ENV{OPENQA_CONFIG} = "$FindBin::Bin/data/24-worker-overall";
 #       file specified via OPENQA_LOGFILE instead of stdout/stderr.
 $ENV{OPENQA_LOGFILE} = undef;
 
+my $workdir = tempdir("/tmp/$FindBin::Script-XXXX");
+chdir $workdir;
+my $guard = scope_guard sub { chdir $FindBin::Bin };
+
 # define fake isotovideo
 {
     package Test::FakeProcess;    # uncoverable statement count:1
@@ -492,7 +496,7 @@ subtest 'is_qemu_running' => sub {
     ok OpenQA::Worker::is_qemu('/usr/bin/qemu-system-x86_64'), 'QEMU executable considered to be QEMU';
     ok !OpenQA::Worker::is_qemu('/usr/bin/true'), 'other executable not considered to be QEMU';
 
-    my $pool_directory = tempdir('poolXXXX');
+    my $pool_directory = tempdir('/tmp/poolXXXX');
     $worker->pool_directory($pool_directory);
 
     $worker->no_cleanup(0);
@@ -519,8 +523,8 @@ subtest 'checking and cleaning pool directory' => sub {
 
     # assign temporary pool dir
     # note: Using scope guard to "get out" of pool directory again so we can delete the tempdir.
-    my $pool_directory = tempdir('poolXXXX');
-    my $guard = scope_guard sub { chdir $FindBin::Bin };
+    my $pool_directory = tempdir('/tmp/poolXXXX');
+    my $guard = scope_guard sub { chdir $workdir };
     $worker->pool_directory($pool_directory);
 
     # pretend QEMU is still running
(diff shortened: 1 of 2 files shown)


perlpunk (commented):
Running
```
make test TESTS=t/24-worker-overall.t
```
still creates `t/poolE5kM` and does not delete it.

Read the suggestions in the ticket:
> "creates a tempdir 3 times, but one of them is not cleaned up. I believe this is because the worker is still holding the pool dir lock when the cleanup is called."

you can solve this first to satisfy AC2 and then work on AC1

Wabri:
Yes, I was already working on it after the push, I thought it was from the previous run, but no... On it!

Wabri:
There is still a leftover `worker-log.txt` inside the `t/` directory, also I think it needs to be inside a pool... I'll leave the resolve for tomorrow!

Wabri:
Maybe I should simply create them on `/tmp` and call it the day!

Martchus (approved):
The style check and commit message check are failing but it looks generally good.

lib/OpenQA/Worker.pm
```diff
@@ -814,8 +814,7 @@ sub _clean_pool_directory ($self) {
         next if $excludes{$file};
         if (-d $file) {
             remove_tree($file);
-        }
-        else {
+        } else {
```
okurz, line 817: ?

perlpunk:
This leaves a `t/worker-log.txt` after running `t/24-worker-overall.t`

perlpunk:
> I thought we would check for leftover files after tests, so I am wondering if this check is still there

Ah, so far we only have this for os-autoinst: t/lib/OpenQA/Test/CheckGitStatus.pm

t/24-worker-overall.t
```diff
@@ -917,7 +923,7 @@ subtest 'resolving 127.0.0.1 without relying on getaddrinfo()' => sub {
 };
 
 subtest 'storing package list' => sub {
-    $worker->pool_directory(tempdir('pool-dir-XXXXX'));
+    $worker->pool_directory(tempdir('pool-dir-XXXXX', CLEANUP => 1));
```
perlpunk, line 926: Why did you not use `/tmp` here?

t/24-worker-overall.t
```diff
@@ -492,7 +492,7 @@ subtest 'is_qemu_running' => sub {
     ok OpenQA::Worker::is_qemu('/usr/bin/qemu-system-x86_64'), 'QEMU executable considered to be QEMU';
     ok !OpenQA::Worker::is_qemu('/usr/bin/true'), 'other executable not considered to be QEMU';
 
-    my $pool_directory = tempdir('poolXXXX');
+    my $pool_directory = tempdir('/tmp/poolXXXX', CLEANUP => 1);
```
perlpunk, line 495: I think the `CLEANUP` is not needed. The code calls `File::Temp->newdir()` which defaults to `CLEANUP => 1`

lib/OpenQA/Worker.pm
Wabri: I was wondering what that function did, I think I went into autopilot and changed that ugly else under the parenthesis. Do you want me to revert the change?

Wabri: Perl tidy don't like this, so I'm reverting!

Wabri:
I've add an unlink to worker-log.txt at the end of the test to clean the worktree.

t/24-worker-overall.t
```diff
@@ -113,6 +113,7 @@ combined_like { $worker->log_setup_info }
 qr/.*http:\/\/localhost:9527,https:\/\/remotehost.*qemu_i386,qemu_x86_64.*Errors occurred.*foo.*bar.*/s,
   'setup info with parse errors';
 
+
```
okurz, line 116: ?

t/24-worker-overall.t
```diff
@@ -941,4 +948,6 @@ subtest 'storing package list' => sub {
     qr/doesn't return any data/, 'log message about no data';
 };
 
+unlink 'worker-log.txt';
```
okurz, line 951: why this? And just calling unlink here means in case of fatal exits the file will not be removed. That would be better solved with

```suggestion
END { unlink 'worker-log.txt' }
```

but the change looks unrelated to tempdir related problems

okurz (changes requested):

t/24-worker-overall.t
perlpunk: And that's what we should do in every test. Just use a temp directory for the whole test, then we don't have to care abotu individual files.
That's what we did for os-autoinst a while ago.

In some tests that might require copying over some stuff from the git dir.

okurz (approved):

PR #6333, opened 2025-03-27
Avoid interrupting/skipping directly chained jobs due to high load

Martchus:
Just like we do not abort running jobs due to high load we should also avoid interrupting directly chained jobs due to high load as this will be rather disturbing.

Note that directly chained dependencies were added to support bare-metal backends like IPMI where one cannot easily upload/reuse use HDD images. The check for the load was on the other hand introduced specifically for VM-based backends where high load leads to typing and other issues. So considering the high load check non-fatal for directly chained dependencies is not a problem.

Related ticket: https://progress.opensuse.org/issues/179545

### DIFF
--- a/lib/OpenQA/Worker.pm
+++ b/lib/OpenQA/Worker.pm
@@ -59,6 +59,7 @@ has 'clients_by_webui_host';
 has 'current_webui_host';
 has 'current_job';
 has 'current_error';
+has 'current_error_is_fatal';
 has 'worker_hostname';
 has 'isotovideo_interface_version';
 
@@ -225,13 +226,12 @@ sub status ($self) {
         $status{current_webui_host} = $self->current_webui_host;
         $status{job} = $current_job->info;
     }
-    elsif (my $availability_error = $self->check_availability) {
+    elsif (my $availability_reason = $self->set_current_error_based_on_availability) {
         $status{status} = 'broken';
-        $self->current_error($status{reason} = $availability_error);
+        $status{reason} = $availability_reason;
     }
     else {
         $status{status} = 'free';
-        $self->current_error(undef);
     }
     if (my $queue = $self->{_queue}) {
         $status{pending_job_ids} = $queue->{pending_job_ids} if keys %{$queue->{pending_job_ids}};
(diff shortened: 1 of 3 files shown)


okurz (approved):

lib/OpenQA/Worker.pm
```diff
@@ -611,36 +616,48 @@ sub is_ovs_dbus_service_running ($self) {
...
     # check org.opensuse.os_autoinst.switch if it is a MM-capable worker slot
-    return "D-Bus service '$self->{_ovs_dbus_service_name}' is not running"
+    return (0, "D-Bus service '$self->{_ovs_dbus_service_name}' is not running")
       if $settings->has_class('tap') && !$self->is_ovs_dbus_service_running;
 
-    return undef;
+    # avoid running jobs if system utilization is critical
+    if (my $error = $self->_check_system_utilization) { return (1, $error) }
+
+    return (1, undef);
```
okurz, line 653: why not just

```suggestion
    return (1, $self->_check_system_utilization);
```

okurz (changes requested):

lib/OpenQA/Worker.pm
```diff
@@ -611,36 +616,46 @@ sub is_ovs_dbus_service_running ($self) {
...
+    return (0, 'Unable to determine worker address (WORKER_HOSTNAME)') unless $settings->auto_detect_worker_address;
 
     # check org.opensuse.os_autoinst.switch if it is a MM-capable worker slot
-    return "D-Bus service '$self->{_ovs_dbus_service_name}' is not running"
+    return (0, "D-Bus service '$self->{_ovs_dbus_service_name}' is not running")
       if $settings->has_class('tap') && !$self->is_ovs_dbus_service_running;
 
-    return undef;
+    # avoid picking up new jobs if system utilization is critical
+    return (1, $self->_check_system_utilization);
```
perlpunk, line 651: this sounds a bit confusing to me. It says "avoid picking up new jobs" but it is returning a 1 which means success, right?

Martchus: The "1" is so it still continues with the execution of already enqueued jobs. It will just avoid picking up new jobs and that's the 2nd return value. I changed the phrasing. This should be good enough considering I also added general explanation at the top of the function.

okurz (approved):

perlpunk (approved):

PR #6335, opened 2025-03-28
oauth2: use openQA base_url to create redirect URI if set

AdamWill:
I'm setting up oauth2 auth for Fedora ATM (up till now we used OpenID auth, but we're shutting that off). One problem I hit was that this redirect URI generation used the wrong hostname. Our staging instance is at https://openqa.stg.fedoraproject.org. The machine that serves it has the hostname
openqa-lab01.iad2.fedoraproject.org. The redirect_uri was using the machine's hostname, so it was coming out as
https://openqa-lab01.iad2.fedoraproject.org/login , which is not going to work.

This tweaks the redirect URI generation to use the hostname from the openQA config base_url, if it's set. If not, it falls back to the hostname from the request.

### DIFF
--- a/lib/OpenQA/WebAPI/Auth/OAuth2.pm
+++ b/lib/OpenQA/WebAPI/Auth/OAuth2.pm
@@ -93,7 +93,9 @@ sub auth_login ($controller) {
     croak 'Config was not parsed' unless my $main_config = $controller->app->config->{oauth2};
     croak 'Setup was not called' unless my $provider_config = $main_config->{provider_config};
 
-    my $get_token_args = {redirect_uri => $controller->url_for('login')->userinfo(undef)->to_abs};
+    my $base_url = $controller->app->config->{global}->{base_url};
+    my $host = $base_url ? Mojo::URL->new($base_url)->host : $controller->req->url->host;
+    my $get_token_args = {redirect_uri => $controller->url_for('login')->userinfo(undef)->host($host)->to_abs};
     $get_token_args->{scope} = $provider_config->{token_scope};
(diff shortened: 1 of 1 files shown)


AdamWill:
I'm not at all sure this is correct, but I'm sending it as a PR so I have it on record, and to see if e.g. @kraih or @marmarek can suggest a better way to do this.

lib/OpenQA/WebAPI/Auth/OAuth2.pm
```diff
@@ -93,7 +93,8 @@ sub auth_login ($controller) {
     croak 'Config was not parsed' unless my $main_config = $controller->app->config->{oauth2};
     croak 'Setup was not called' unless my $provider_config = $main_config->{provider_config};
 
-    my $get_token_args = {redirect_uri => $controller->url_for('login')->userinfo(undef)->to_abs};
+    my $url = $controller->app->config->{global}->{base_url} || $controller->req->url->base;
+    my $get_token_args = {redirect_uri => "${url}/login"};
```
Martchus, line 97: Maybe the following works and would preserve the use of `url_for` to find the URL for that named route:
```suggestion
    my $base_url = $controller->app->config->{global}->{base_url};
    my $host = ($base_url ? Mojo::URL->new($base_url) : $controller->req->url)->host;
    my $get_token_args = {redirect_uri => "$controller->url_for('login')->host($host)->userinfo(undef)};
```

AdamWill: that had a few errors in it, but with those fixed, yeah, seems to work.

okurz (approved):

Wabri (approved):

PR #6338, opened 2025-03-28
Check if the taskname exists if not avoid undefined value

Wabri:
related: https://progress.opensuse.org/issues/179395

### DIFF
--- a/templates/webapi/test/infopanel.html.ep
+++ b/templates/webapi/test/infopanel.html.ep
@@ -46,7 +46,9 @@
                     % else {
                         % my $state_detail = '';
                         % if (my @gru_deps = $job->gru_dependencies) {
-                            % $state_detail = join(';', map { 'id: ' . $_->gru_task_id . ', name: ' . $_->gru_task->taskname } @gru_deps);
+                            % $state_detail = join(';', map {
+                            %   'id: ' . $_->id . ', name: ' . $_->taskname
+                            %  } map { $_->gru_task // () } @gru_deps);
                             % $state_detail = ", waiting for background tasks ($state_detail)";
                         % }
                         State: <b><%= $job->state %></b><%= $state_detail %>\


templates/webapi/test/infopanel.html.ep
```diff
@@ -46,7 +46,10 @@
                     % else {
                         % my $state_detail = '';
                         % if (my @gru_deps = $job->gru_dependencies) {
-                            % $state_detail = join(';', map { 'id: ' . $_->gru_task_id . ', name: ' . $_->gru_task->taskname } @gru_deps);
+                            % my $task = $_->gru_task;
+                            % $state_detail = join(';', map {
+                            %   'id: ' . $_->gru_task_id . ', name: ' . ($task ? $task-taskname : '')
```
kalikiana, line 51:
```suggestion
                            %   'id: ' . $_->gru_task_id . ', name: ' . ($task ? $task->taskname : '')
```

Wabri: Autocompletion in these files is nonexistent, thanks for the catch!

templates/webapi/test/infopanel.html.ep
```diff
@@ -201,4 +204,4 @@
                     % }
                 </div>
                 % }
-            </div>
\ No newline at end of file
+            </div>
```
okurz, line 207: please revert

templates/webapi/test/infopanel.html.ep
```diff
@@ -46,7 +46,10 @@
                     % else {
                         % my $state_detail = '';
                         % if (my @gru_deps = $job->gru_dependencies) {
-                            % $state_detail = join(';', map { 'id: ' . $_->gru_task_id . ', name: ' . $_->gru_task->taskname } @gru_deps);
+                            % my $task = $_->gru_task;
```
okurz, line 49: I don't see how this is supposed to work. Did you test this?

okurz (changes requested):

templates/webapi/test/infopanel.html.ep
Wabri: why not?

Wabri: Oh you are right, I wanted to put it inside the map not outside...

templates/webapi/test/infopanel.html.ep
```diff
@@ -46,7 +46,10 @@
                     % else {
                         % my $state_detail = '';
                         % if (my @gru_deps = $job->gru_dependencies) {
-                            % $state_detail = join(';', map { 'id: ' . $_->gru_task_id . ', name: ' . $_->gru_task->taskname } @gru_deps);
+                            % $state_detail = join(';', map {
+                            %   my $task = $_->gru_task;
+                            %   'id: ' . $_->gru_task_id . ', name: ' . ($task ? $task->taskname : '')
+                            % } @gru_deps);
```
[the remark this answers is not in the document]
Wabri: I didn't got the point of the race condition problem, can you be more clear about that?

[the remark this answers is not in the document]
perlpunk: When calling `my @gru_deps = $job->gru_dependencies` one database query is executed.
Then when iterating over `@gru_deps`  `$_->task` will execute another database query. At this point (between the two SELECTs) the original entry from `gru_dependencies` and the related entry from `gru_tasks` might have already been deleted, so the second SELECT will return nothing.
But that's ok, because if those entries have been deleted meanwhile, we don't need to show anything to the user.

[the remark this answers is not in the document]
Wabri: Ok, thanks a lot, now it make sense!

Resolve the conversation!

perlpunk (approved):

okurz (approved):

kalikiana (approved):

Martchus (approved):

PR #6339, opened 2025-03-28
Support all config features for `workers.ini` as well

Martchus:
* Locate the worker config just like the web UI and client config files to support
    * Locating the config under `/usr/etc`
    * Drop-in configuration files
    * Reading the config file from the Git checkout for development without having to set `OPENQA_CONFIG`

---

There's no ticket for this but when working on https://progress.opensuse.org/issues/179359 I noticed that I would have to weirdly exclude the worker config from Makefile/specfile/documentation changes. So it make sense to bring the worker in-line with everything else first.

### DIFF
--- a/lib/OpenQA/Worker/Settings.pm
+++ b/lib/OpenQA/Worker/Settings.pm
@@ -4,10 +4,13 @@
 package OpenQA::Worker::Settings;
 use Mojo::Base -base, -signatures;
 
+use FindBin '$Bin';
+use Mojo::File 'path';
 use Mojo::URL;
 use Mojo::Util 'trim';
 use Config::IniFiles;
 use Time::Seconds;
+use OpenQA::Config;
 use OpenQA::Log 'setup_log';
 use OpenQA::Utils 'is_host_local';
 use Net::Domain 'hostfqdn';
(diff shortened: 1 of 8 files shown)


lib/OpenQA/Worker/Settings.pm
```diff
@@ -19,17 +22,9 @@ has 'webui_host_specific_settings';
...
-    my @parse_errors;
-    if (-e $settings_file) {
-        $cfg = Config::IniFiles->new(-file => $settings_file);
-        push(@parse_errors, @Config::IniFiles::errors) unless $cfg;
-    }
-    else {
-        push(@parse_errors, "Config file not found at '$settings_file'.");
-        $settings_file = undef;
-    }
+    my $config_paths = lookup_config_files(path($Bin)->dirname->child('etc', 'openqa'), 'workers.ini', 1);
```
kalikiana, line 25: Passing the home again here seems redundant? I guess I should have noticed this in the original PR. Shouldn't home be the same in all cases?

okurz: Where is home passed again? What is passed here is the complete path to workers.ini

okurz (approved):

lib/OpenQA/Worker/Settings.pm
kalikiana: The first argument is "home" in terms of lookup_config_files. Shouldn't the filename be enough when the API knows other locations, too?

Martchus: The `path($Bin)->dirname->child('etc', 'openqa')` part is redundant, indeed (as it is the same for the database initialization). I can improve that.

I guess name of the argument of `lookup_config_files` also shouldn't be called `$home` as it is actually a subdirectory within the `$home` directory which needs to be passed.

Martchus: Now I'm also wondering whether this `path($Bin)->dirname->child('etc', 'openqa')` approach is a good idea at all.

If we start packaging `/usr/etc/…` then the case of finding the config directory within the Git checkout will actually be accidentally "triggered" for a packaged installation (where binaries are under `/usr/bin/` and `/usr/bin/.../etc/` will be valid). Maybe this should use the home directory of the Mojo app in any case.

okurz (approved):

kalikiana (approved):

PR #6342, opened 2025-03-31
Use consistent temporary database path

okurz:
As we already use /dev/shm/tpg mostly everywhere we can simplify the
call to t/test_postgresql by creating the temporary database in the same
path by default. This way to manually execute tests relying on a
database one can just call "t/test_postgresql && prove -l …".

### DIFF
--- a/t/test_postgresql
+++ b/t/test_postgresql
@@ -1,11 +1,7 @@
 #!/bin/sh -e
 
-DIR=$1
-if test -z "$DIR"; then
-    DIR=$(mktemp -d)
-else
-    DIR=$(readlink -f "$DIR")
-fi
+DIR=${1:-/dev/shm/tpg}
+DIR=$(readlink -f "$DIR")
 if test -e "$DIR"/postmaster.pid; then
     pg_ctl -D "$DIR" stop
 fi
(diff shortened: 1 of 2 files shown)


t/test_postgresql
```diff
@@ -1,11 +1,7 @@
 #!/bin/sh -e
 
-DIR=$1
-if test -z "$DIR"; then
-    DIR=$(mktemp -d)
-else
-    DIR=$(readlink -f "$DIR")
-fi
+DIR=${1:-/dev/shm/tpg}
```
nicksinger, line 3: I don't recommend storing bigger data here. If really the only way, I recommend checking the available size of this path first before using it. A system with high utilization and low memory could yield out-of-space issues quite easily.

perlpunk (approved):

Martchus (approved):

t/test_postgresql
okurz: This is only for tests and our usual CI systems already use the same path everywhere as that is the default used in the Makefile which is used by both developers as well as CI. This only makes a difference for "expert developers" calling the script manually so I don't think this is a relevant problem.

Issue #6309, opened 2025-03-20
View: Parent group with limit confusing


rclobus:
Hello,

On the Debian openQA instance, I'm currently using a parent group, to group all live images together.
Some of them are tested on a daily basis, others weekly or even monthly.

The view is rather confusing:
https://openqa.debian.net/parent_group_overview/1?limit_builds=20#grouped_by_group

![Image](https://github.com/user-attachments/assets/01f1f142-b7c2-49d9-a08c-aef7f0a56a74)

For the last 2 groups, no builds are shown.

When I extend the view to 50 items, the last 2 groups show builds, some of them only 4 days old.

![Image](https://github.com/user-attachments/assets/6eaa4639-44bb-4384-9915-c4ebc1f44084)

With the view limited to 20 builds, I would expect:
* Preferably the last 20 builds, independent of the group (as judged be either the build number or the date)
* Alternatively 5 builds for each of the 4 groups

In all cases, I would have expected the builds from 2025-03-16 to show up.

b10n1k:
https://progress.opensuse.org/issues/179296

b10n1k:
There is an open draft https://github.com/os-autoinst/openQA/pull/6343. welcome to take a look, provide an idea or suggestion

kalikiana:
> There is an open draft [#6343](https://github.com/os-autoinst/openQA/pull/6343). welcome to take a look, provide an idea or suggestion

[poo#180851](https://progress.opensuse.org/issues/180851)

PR #6343, opened 2025-03-31
Do not show empty groups when has no builds on parent_group_overview

b10n1k:
The problem with parent_group_overview is that offers two ways of grouping a set of data. The design of the query seems to serve the purpose of the grouped_by_build interface. When it comes to the grouped_by_group, though, there should be another query to achieve the expected output in the view. That would likely be, the latest build of each group within the build_limits.

The solution presented in this commit is to visualuze only the groups which carry some builds from the query, but let user know that more builds are provided in that parent groups.
To address this problem, another solution would be to add another endpoint in the API and invoke the resultset from this query.
The idea of modifying the message left out because it would not provide a better user experience at the end, leaving still a unsatisfied feeling.

The reason for the `clearFlash` is that the panel remains between the change from one group to another, and otherwise would need to be closed manually.

https://progress.opensuse.org/issues/179296

### DIFF
--- a/assets/javascripts/openqa.js
+++ b/assets/javascripts/openqa.js
@@ -52,6 +52,10 @@ function addFlash(status, text, container) {
   return div;
 }
 
+function clearFlash() {
+  document.querySelectorAll('#flash-messages .alert.alert-primary.alert-dismissible').forEach(e => e.remove());
+}
+
 function addUniqueFlash(status, id, text, container) {
   // add hash to store present flash messages
   if (!window.uniqueFlashMessages) {
--- a/assets/javascripts/parent_group_overview.js
+++ b/assets/javascripts/parent_group_overview.js
@@ -31,6 +31,7 @@ function setupParentGroupOverviewAssets(group_id) {
   }
 
   function updateGroupedByClasses() {
+    clearFlash(); // Do not let the notification messages hanging when group is updated
     if (window.location.hash === undefined || window.location.hash === '#grouped_by_build') {
       document
         .getElementById('grouped_by_group_tab')
@@ -41,6 +42,16 @@ function setupParentGroupOverviewAssets(group_id) {
         .getElementById('grouped_by_build_tab')
         .classList.remove('active', 'show', 'parent_group_overview_grouping_active');
       document.getElementById('grouped_by_group_tab').classList.add('active', 'parent_group_overview_grouping_active');
+      hiddenBuilds = document.querySelectorAll('.no-build-data');
+      if (hiddenBuilds.length > 0) {
+        hiddenBuilds.forEach(elem => {
+          console.log(elem.dataset.buildName);
+        });
+        addFlash(
+          'info',
+          `Parent group has ${hiddenBuilds.length} more builds with no results in the current build limits.`
+        );
+      }
     }
   }
 
--- a/templates/webapi/main/group_builds_functionality_view.html.ep
+++ b/templates/webapi/main/group_builds_functionality_view.html.ep
@@ -9,6 +9,10 @@
     % }
   % }
 
+  % if (!$has_jobs) {
+      <div class="no-build-data" data-build-name="<%= $child->{name} %>" hidden></div>
+  % }
+  % else {
   <div class="mb-2">
     <div class="h4 <%= $title_class_mod %> mb-2">
       %= link_to $child->{name} => url_for('group_overview', groupid => $child->{id})
@@ -45,10 +49,7 @@
             </div>
         </div>
     % }
-    % if ($has_jobs eq 0) {
-      <div class="ml-3 text-body-secondary">
-        No builds
-      </div>
-    % }
+  
   </div>
+  % }
 % }
--- a/t/ui/14-dashboard-parents.t
+++ b/t/ui/14-dashboard-parents.t
@@ -142,6 +142,7 @@ subtest 'filtering subgroups' => sub {
 
 subtest 'View grouped by group' => sub {
     $driver->get('/parent_group_overview/' . $parent_groups->find({name => 'Test parent'})->id);
+    my $grouped_builds = $driver->find_elements('.d-xl-flex flex-row build-row');
     $driver->find_element_by_id('grouped_by_group_tab')->click();
     is(
         $driver->find_element_by_id('grouped_by_group_tab')->get_attribute('class'),
@@ -153,7 +154,19 @@ subtest 'View grouped by group' => sub {
         'active parent_group_overview_grouping_active',
         'grouped by group link remains active'
     );
+    my $grouped_groups = $driver->find_elements('.d-xl-flex flex-row build-row');
     $driver->find_element_by_id('grouped_by_group')->is_displayed();
+    $driver->get('parent_group_overview/'
+          . $parent_groups->find({name => 'Test parent'})->id
+          . '?limit_builds=2#grouped_by_group');
+    my $groups_without_jobs = $driver->find_elements('.no-build-data');
+    is(scalar @$groups_without_jobs, 1, 'correct amount of groups without jobs found');
+    is(scalar @$grouped_groups, scalar @$grouped_builds, 'both groups display the same amount of builds');
+    is(
+        $driver->find_element('#flash-messages .alert-primary span')->get_text(),
+        'Parent group has 1 more builds with no results in the current build limits.',
+        'Notification is shown for jobless groups'
+    );
 };
 
 kill_driver();


perlpunk:
There is something wrong with the layout in the group overview now, but I assume that you are working on it, as it is a draft.

assets/javascripts/parent_group_overview.js
```diff
@@ -41,6 +42,17 @@ function setupParentGroupOverviewAssets(group_id) {
         .getElementById('grouped_by_build_tab')
         .classList.remove('active', 'show', 'parent_group_overview_grouping_active');
       document.getElementById('grouped_by_group_tab').classList.add('active', 'parent_group_overview_grouping_active');
+      hiddenBuilds = document.querySelectorAll('.noBuildData');
+      if (hiddenBuilds.length > 0) {
+        hiddenBuilds.forEach(elem => {
+          console.log(elem.dataset.buildName);
+        });
+        addFlash(
```
okurz, line 50: As I suggested I would just extend the "No builds" text, not a flash

okurz (changes requested):

assets/javascripts/parent_group_overview.js
b10n1k: nevertheless we agree that doing so it doesnt solve the problem from the users perspective. Whats wrong with the flash?

b10n1k:
image

image

b10n1k:
> There is something wrong with the layout in the group overview now, but I assume that you are working on it, as it is a draft.

if you have spotted something, point me to it. what looks broken exactly?

perlpunk:
master: parent-group-before
PR: parent-group-after

b10n1k:
thanks @perlpunk

b10n1k:
working on the tests. I pushed something but I will try to extend it in the evening.

assets/javascripts/parent_group_overview.js
```diff
@@ -31,6 +31,7 @@ function setupParentGroupOverviewAssets(group_id) {
   }
 
   function updateGroupedByClasses() {
+    clearFlash();
```
Martchus, line 34: What is this expected to clear? A comment in the code would help.

assets/javascripts/openqa.js
```diff
@@ -52,6 +52,13 @@ function addFlash(status, text, container) {
   return div;
 }
 
+function clearFlash() {
+  const flashMessages = $('#flash-messages').find('.alert.alert-primary.alert-dismissible');
+  flashMessages.each(function () {
+    $(this).remove();
+  });
```
Martchus, line 59: This can be achieved in a single line as jQuery always operates on lists:

```suggestion
  $('#flash-messages .alert.alert-primary.alert-dismissible').remove();
```

The design of jQuery is so unintuitive (and thus probably bad) that one can even easily miss cases where the design is useful.

**This is also easily doable without jQuery and I'd avoid using it in new code. So I suggest to use the following:**

```suggestion
  document.querySelectorAll('#flash-messages .alert.alert-primary.alert-dismissible').forEach(e => e.remove());
```

This code uses proper function names instead of "magic" ones and is thus much more self-explaining.

templates/webapi/main/group_builds_functionality_view.html.ep
```diff
@@ -9,6 +9,10 @@
     % }
   % }
 
+  % if ($has_jobs eq 0) {
+      <div class="noBuildData" data-build-name="<%= $child->{name} %>" hidden></div>
```
Martchus, line 13: We normally use dashes in CSS classes:
```suggestion
      <div class="no-build-data" data-build-name="<%= $child->{name} %>" hidden></div>
```

Martchus (changes requested):

assets/javascripts/parent_group_overview.js
b10n1k: the results passed to the template are indeed limited by the filter. each template parses the same data. 

Now I read your comment repeately ... are you saying, that if we keep the notification should be more clear that the view is limited by the filter? or how should I interpret the last sentence?

b10n1k:
Very cool review @Martchus. thanks a lot

assets/javascripts/parent_group_overview.js
okurz: I would just keep the group title and replace/extend the "No builds" text depending on if there are no builds at all or just no builds shown due to the filtering

okurz:
@rclobus this should fix #6309 . Your feedback would be appreciated in particular regarding the discussion in https://github.com/os-autoinst/openQA/pull/6343#discussion_r2022151792

rclobus:
> @rclobus this should fix #6309 . Your feedback would be appreciated

I've applied this patch (4ab9ade1d6d9b697789b9411abf22c0cfa75ee75) as hotfix to our production server :-) Feel free to browse around.
I can also prepare a database dump, if you would like to reproduce the scenario locally.

As you can see with https://openqa.debian.net/parent_group_overview/1?limit_builds=50#grouped_by_group, there is an entry for 'Debian live stable' with build 20250402, which I would definitely have expected to see when only 20 builds are shown, however that build is not shown.

image

I also tried the view with '20':
I find the flash text confusing, it refers to 'builds', whereas I would expect 'job groups' (because many more builds are not shown when the limit is only 20):
`Parent group has 2 more builds with no results in the current build limits.`
->
`The parent group has 2 more job groups with no results in the current build limits.`

Then I tried the view with '10':
https://openqa.debian.net/parent_group_overview/1?limit_builds=10#grouped_by_group

There is no flash text, and only one of the sub groups is listed. The other 3 sub groups are not mentioned...

The latest 5 builds (at the moment I write this) are:
* 20250403 (unstable, testing)
* 20250402 (unstable, stable)
* 20250401 (unstable, stable)
* 20250331 (unstable)
* 20250331-w (testing)

I would expect these 5 to be shown in all cases

b10n1k:
> > @rclobus this should fix #6309 . Your feedback would be appreciated
> 
> I've applied this patch ([4ab9ade](https://github.com/os-autoinst/openQA/commit/4ab9ade1d6d9b697789b9411abf22c0cfa75ee75)) as hotfix to our production server :-) Feel free to browse around. I can also prepare a database dump, if you would like to reproduce the scenario locally.
> 

Please share the dump so we test over the same context

> As you can see with https://openqa.debian.net/parent_group_overview/1?limit_builds=50#grouped_by_group, there is an entry for 'Debian live stable' with build 20250402, which I would definitely have expected to see when only 20 builds are shown, however that build is not shown.
> 

From one perspective you are right. I guess the confussion comes from the fact that the `by group` will parse only the data that you will see in `by build`. If not there, the `by group` wont do a new query, and as such will reformat the builds to present as `job groups`.

> image
> 
> I also tried the view with '20': I find the flash text confusing, it refers to 'builds', whereas I would expect 'job groups' (because many more builds are not shown when the limit is only 20): `Parent group has 2 more builds with no results in the current build limits.` -> `The parent group has 2 more job groups with no results in the current build limits.`
> 
> Then I tried the view with '10': https://openqa.debian.net/parent_group_overview/1?limit_builds=10#grouped_by_group
> 
> There is no flash text, and only one of the sub groups is listed. The other 3 sub groups are not mentioned...
> 

Actually I can see the flash on https://openqa.debian.net/parent_group_overview/1?limit_builds=10#grouped_by_group

However one question is whether the flash make the view less confusing, or as a user you would like to have another text under each group, replacing the `no builds` which is misleading.

> The latest 5 builds (at the moment I write this) are:
> 
> * 20250403 (unstable, testing)
> * 20250402 (unstable, stable)
> * 20250401 (unstable, stable)
> * 20250331 (unstable)
> * 20250331-w (testing)
> 
> I would expect these 5 to be shown in all cases

I understand the expectations here but to achieve this would be bigger change. We cant change the query, as it will break the `by build` view. We will have to restruct a few things to make another call to the database in order to make it right. Something to keep for the future as for this commit the goal is to provide a view with less confusion for the user.

rclobus:
>> I can also prepare a database dump, if you would like to reproduce the scenario locally.
> 
> Please share the dump so we test over the same context

I've created the dump as geekotest with `pg_dump -d openqa --clean --create --blobs --file openqa.debian.net.20250403.sql`, you can restore it with `openqa.debian.net.20250403.sql | psql` (CAUTION: It will drop the existing database!)

The compressed dump has been uploaded at (about 60MB): http://pioneers.game-host.org/openqa.debian.net.20250403.sql.gz

> > As you can see with https://openqa.debian.net/parent_group_overview/1?limit_builds=50#grouped_by_group, there is an entry for 'Debian live stable' with build 20250402, which I would definitely have expected to see when only 20 builds are shown, however that build is not shown.
> 
> From one perspective you are right. I guess the confussion comes from the fact that the `by group` will parse only the data that you will see in `by build`. If not there, the `by group` wont do a new query, and as such will reformat the builds to present as `job groups`.

But even the 'by build' query results in a rather random selection of builds (but I didn't look at the 'by build' results in detail before, I only now looked at it).

> > I also tried the view with '20': I find the flash text confusing, it refers to 'builds', whereas I would expect 'job groups' (because many more builds are not shown when the limit is only 20): `Parent group has 2 more builds with no results in the current build limits.` -> `The parent group has 2 more job groups with no results in the current build limits.`
> > Then I tried the view with '10': https://openqa.debian.net/parent_group_overview/1?limit_builds=10#grouped_by_group
> > There is no flash text, and only one of the sub groups is listed. The other 3 sub groups are not mentioned...
> 
> Actually I can see the flash on https://openqa.debian.net/parent_group_overview/1?limit_builds=10#grouped_by_group

Indeed. I can see it now. I must have scrolled down and missed the message at the top.

> However one question is whether the flash make the view less confusing, or as a user you would like to have another text under each group, replacing the `no builds` which is misleading.

I don't like leaving out the sub groups in the view, even if nothing is shown for them (at the moment).
The flash at the top is OK.

...
> I understand the expectations here but to achieve this would be bigger change. We cant change the query, as it will break the `by build` view. We will have to restruct a few things to make another call to the database in order to make it right. Something to keep for the future as for this commit the goal is to provide a view with less confusion for the user.

As it currently is, the 'by build' view also shows a result set that doesn't contain the most recent builds, so I guess that adjusting the query should be handled before the adjustment of the view.

image

Here you can also see some of the more recent builds from the sub groups are missing.

b10n1k:
@rclobus I couldnt use the dump to test it but I created another PR. can you take a look https://github.com/os-autoinst/openQA/pull/6367

b10n1k:
https://progress.opensuse.org/issues/180851

PR #6346, opened 2025-04-01
Trigger ci for https://progress.opensuse.org/issues/176319

aylen384:
https://progress.opensuse.org/issues/176319

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,8 +1,5 @@
 version: 2.1
 
-orbs:
-  codecov: codecov/codecov@5.3.0
-
 aliases:
   - &restore_cache
     keys:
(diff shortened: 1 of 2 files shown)


asdil12:
meh - the /home/squamata/os-autoinst is from 2025-03-26 - which doesn't contain my log stuff yet. Why is it not up to date?

okurz:
> meh - the /home/squamata/os-autoinst in the CI is from 2025-03-26 which doesn't contain my log stuff yet. Why is it not up to date?

Possibly because https://app.circleci.com/pipelines/github/os-autoinst/openQA/16467/workflows/41310445-fc05-4644-8bdb-340afbee537a/jobs/159305 only updates the os-autoinst git hash when also any other dependency changes

PR #6347, opened 2025-04-01
Install default config under `/usr/share/doc/openqa/examples`

Martchus:
See particular commit messages and https://progress.opensuse.org/issues/179359.

### DIFF
--- a/dist/rpm/openQA.spec
+++ b/dist/rpm/openQA.spec
@@ -566,6 +566,8 @@ fi
 %{_sbindir}/rcopenqa-websockets
 %{_sbindir}/rcopenqa-webui
 %{_sbindir}/rcopenqa-livehandler
+%ghost %config(noreplace) %attr(0644,geekotest,root) %{_sysconfdir}/openqa/openqa.ini
+%ghost %config(noreplace) %attr(0640,geekotest,root) %{_sysconfdir}/openqa/database.ini
 %dir %{_sysconfdir}/openqa
 %dir %{_sysconfdir}/openqa/openqa.ini.d
 %dir %{_sysconfdir}/openqa/database.ini.d
@@ -713,6 +715,8 @@ fi
 %{_sbindir}/rcopenqa-slirpvde
 %{_sbindir}/rcopenqa-vde_switch
 %{_sbindir}/rcopenqa-worker
+%ghost %config(noreplace) %{_sysconfdir}/openqa/workers.ini
+%ghost %config(noreplace) %attr(0400,_openqa-worker,root) %{_sysconfdir}/openqa/client.conf
 %dir %{_sysconfdir}/openqa/workers.ini.d
 %dir %{_sysconfdir}/openqa/client.conf.d
 %dir %{_distconfdir}/openqa/workers.ini.d
--- a/dist/rpm/openQA.spec
+++ b/dist/rpm/openQA.spec
@@ -15,6 +15,10 @@
 # Please submit bugfixes or comments via http://bugs.opensuse.org/
 #
 
+# define _distconfdir for openSUSE Leap 15 and SLE 15 (like other packages such as libvpl do)
+%if 0%{?suse_version} < 1550
+%define _distconfdir %{_prefix}%{_sysconfdir}
+%endif
 
 # can't use linebreaks here!
 %define openqa_main_service openqa-webui.service
@@ -563,8 +567,13 @@ fi
 %{_sbindir}/rcopenqa-webui
 %{_sbindir}/rcopenqa-livehandler
 %dir %{_sysconfdir}/openqa
-%config(noreplace) %attr(-,geekotest,root) %{_sysconfdir}/openqa/openqa.ini
-%config(noreplace) %attr(-,geekotest,root) %{_sysconfdir}/openqa/database.ini
+%dir %{_sysconfdir}/openqa/openqa.ini.d
+%dir %{_sysconfdir}/openqa/database.ini.d
+%dir %{_distconfdir}/openqa
+%dir %{_distconfdir}/openqa/openqa.ini.d
+%dir %{_distconfdir}/openqa/database.ini.d
+%{_datadir}/doc/openqa/examples/openqa.ini
+%{_datadir}/doc/openqa/examples/database.ini
 %dir %{_datadir}/openqa
 %config %{_sysconfdir}/logrotate.d
 # apache vhost
@@ -669,6 +678,11 @@ fi
 %endif
 
 %files common
+%if 0%{?suse_version} < 1550
+%dir %{_distconfdir}
+%endif
+%dir %{_datadir}/doc/openqa
+%dir %{_datadir}/doc/openqa/examples
 %dir %{_datadir}/openqa
 %{_datadir}/openqa/lib
 %exclude %{_datadir}/openqa/lib/OpenQA/CacheService/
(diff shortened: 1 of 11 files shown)


okurz (changes requested):
Needs at least spec-file changes, see OBS build check errors

```
[  547s] Processing files: openQA-worker-5.1743516815.a621d572-7793.1.x86_64
[  547s] error: File not found: /home/abuild/rpmbuild/BUILD/openQA-5.1743516815.a621d572-build/BUILDROOT/usr/etc/openqa/workers.ini
[  547s] 
[  547s] RPM build warnings:
[  547s]     File listed twice: /usr/share/openqa
```

And I thought the database.ini file so far is mandatory, isn't it? I guess you would need to fix the build errors first and then verify the functionality based on the actual packages

dist/rpm/openQA.spec
```diff
@@ -563,8 +563,13 @@ fi
...
 %{_sbindir}/rcopenqa-livehandler
 %dir %{_sysconfdir}/openqa
-%config(noreplace) %attr(-,geekotest,root) %{_sysconfdir}/openqa/openqa.ini
-%config(noreplace) %attr(-,geekotest,root) %{_sysconfdir}/openqa/database.ini
+%dir %{_sysconfdir}/openqa/openqa.ini.d
+%dir %{_sysconfdir}/openqa/database.ini.d
+%dir %{_distconfdir}/openqa
+%dir %{_distconfdir}/openqa/openqa.ini.d
+%dir %{_distconfdir}/openqa/database.ini.d
+%config(noreplace) %attr(-,geekotest,root) %{_datadir}/doc/openqa/examples/openqa.ini
```
perlpunk, line 571: Why `noreplace` for the example configs? I guess we would never edit them

Martchus: Right, this is a leftover I need to get rid of.

Martchus:
> And I thought the database.ini file so far is mandatory, isn't it? I guess you would need to fix the build errors first and then verify the functionality based on the actual packages

I'll look into the packaging errors and will of course check whether the packages work. Note that `database.ini` is *not* mandatory anymore after my recent changes.

Martchus:
I tested it locally and it works. There's only one caveat: The config files I had under `/etc/openqa`, e.g. `/etc/openqa/openqa.ini` were renamed to e.g. `/etc/openqa/openqa.ini.rpmsave`. Not a big deal but it means one cannot just update, restart and everything still works. Not sure how to avoid this behavior.

docs/GettingStarted.asciidoc
```diff
@@ -264,6 +264,52 @@ Each click point can have an `id`, and if a needle contains multiple click point
 you must pass it to `testapi::assert_and_click` to select which click point
 to use.
 
+=== Configuration
```
okurz, line 267: I wonder why you put this section into "GettingStarted" and not "Installing". Especially because we now by default do not need files like openqa.ini and such anymore for default operation I think it's better suited to move this section to Installing and keep this documented tailored for users which might not even need to maintain their own openQA installation.

docs/GettingStarted.asciidoc
```diff
@@ -264,6 +264,52 @@ Each click point can have an `id`, and if a needle contains multiple click point
...
+The different components of openQA read their configuration from the following
+files:
+
+* `openqa.ini`: This is the "web UI" config as services providing the web
+  interface and related services such as the scheduler are configured via that
+  file.
```
okurz, line 273: This makes it sounds like this file would actually be necessary but by default we would not even need it anymore. We could try to start with the  "Locations" section and then put this list together with drop-in configurations but I guess you wanted to explain first configurations applicable for the individual components. How about leaving out the extension and not taking about files yet but call it "configuration components"? Something like



```suggestion
The different components of openQA use the following configuration components:

* `openqa`: This is the "web UI" configuration for services providing the web
  interface and related services such as the scheduler.
```

okurz (changes requested):

docs/GettingStarted.asciidoc
Martchus: I added it to "GettingStarted" because this document is the first and already contains mentionings of `/etc/openqa/openqa.ini`. So I thought it makes most sense to explain the concept of configuration upfront as it is a prerequisite for following sections.

I must also say that I find the configuration in general a very fundamental thing to know about when using software. It seems actually more important than some of the other topics we mention on "GettingStarted".

I can nevertheless move it to "Installing" if you like that better.

docs/GettingStarted.asciidoc
Martchus: That's bad because there is no "component" that is literally called `openqa`. I would rather explicitly name these configuration files. When I read documentation I usually don't want any abstract bla-bla but concrete information.

I can add a note that the files are not required to exist.

docs/GettingStarted.asciidoc
okurz: Open to hear feedback from others. As this is about documentation also "beginners" can provide helpful insight

kalikiana: Maybe @realcharmer could provide some useful feedback here?

realcharmer: Whilst looking at it as a relative newcomer, I prefer the way @Martchus structures it. I'd like to know what each configuration file does before diving into installation itself. I might also already be using an existing instance without having to install it myself from scratch. Having an initial overview of the config file is good in my opinion, although i do understand what @okurz is trying to point out.

docs/GettingStarted.asciidoc
Martchus: For now I only fixed the phrasing and added an explicit note stating that defaults are used if the files are not present.

docs/GettingStarted.asciidoc
```diff
@@ -264,6 +264,54 @@ Each click point can have an `id`, and if a needle contains multiple click point
...
+https://en.opensuse.org/openSUSE:Packaging_UsrEtc#Variant_1_(ideal_case)[package maintainer]
+can place default values deviating from upstream defaults.
+
+The client configuration can also be put under `~/.config/openqa/client.conf`.
+
+For development, checkout the section about
+<<Contributing.asciidoc#_customize_configuration_directory,customizing the configuration directory>>.
+
+==== Drop-in configurations
+It is possible to split the configuration into multiple files and store these
```
okurz, line 307: Can we emphasize that drop-in configurations are preferred to the complete overrides? I am worried because now you *first* describe the complete ini files and then mention drop-in that people assume they would need or should use openqa.ini. WDYT?

kalikiana: I would lean towards just mentioning the legacy config files which we continue to support at the end. That's what they are to my mind.

Martchus:
I've just tested the newly built packages locally. For that I first reverted my local setup installed the latest RPMs before this change. Then I updated again using this change. It looks like the use of `%ghost` prevents existing config files to be renamed/removed. So I keep this change as it really seems to be the simplest solution to the problem. (Note that `rpm -ql …` still lists e.g. `/etc/openqa/openqa.ini` but the rpm doesn't actually contain that file anymore and our install scripts also don't install them anymore.)

okurz (approved):

docs/GettingStarted.asciidoc
```diff
@@ -264,6 +264,57 @@ Each click point can have an `id`, and if a needle contains multiple click point
 you must pass it to `testapi::assert_and_click` to select which click point
 to use.
 
+=== Configuration
+The different components of openQA read their configuration from the following
+files:
+
+* `openqa.ini`, `openqa.ini.d/*.ini`: These files are the "web UI" config.
+  Services providing the web interface and related services such as the
+  scheduler are configured via these files.
```
b10n1k, line 273: no problem for now but I found it too generic. And later it says `If these files are not present, defaults are used.` but which is the default if files appears in more than one path

b10n1k (approved):

ggardet:
Some files are now now owned by `geekotest` anymore. ALso, on an existing installation, I needed to change owner of `openqa.ini.d/` to `geekotest`, otherwise the openqa worker did not start with permission denied.
Or maybe, https://github.com/os-autoinst/openQA/pull/6366 would be enough?

Martchus:
I think https://github.com/os-autoinst/openQA/pull/6366 should be enough.

PR #6348, opened 2025-04-01
Improve job priority display

realcharmer:
https://progress.opensuse.org/issues/113495

Improve the job priority section when job is scheduled. Implemented changes are as follows:

- Do not display the + and - buttons when user is not an operator
- Add ? icon with a tooltip containing description of how priority works
- The ? icon is a hyperlink for the relevant documentation
- Existing tooltips for the + and - buttons have been removed

### DIFF
--- a/templates/webapi/test/infopanel.html.ep
+++ b/templates/webapi/test/infopanel.html.ep
@@ -110,13 +110,19 @@
                 </div>
                 % if ($job->state eq SCHEDULED) {
                     <div>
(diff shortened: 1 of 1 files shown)


okurz:
Care to share a screenshot? That would help users reviewing this

realcharmer:
The + and - buttons are shown only when the user is an operator:

priority-operator

The ? help icon is shown regardless of user status. It provides a tooltip (native bootstrap component) with a short description of how priority works. The ? symbol itself is a hyperlink for the relevant documentation. This example shows a user which is not logged it and is hovering their cursor over the help icon:

priority-tooltip

okurz (approved):

perlpunk (approved):

PR #6354, opened 2025-04-03
Change default branch handling to avoid git conflicts

gauravpathak:
Related: https://progress.opensuse.org/issues/179314


lib/OpenQA/Git.pm
```diff
@@ -81,6 +81,20 @@ sub set_to_latest_master ($self, $args = undef) {
         return $self->_format_git_error($res, 'Unable to reset repository to origin/master') unless $res->{status};
     }
 
+    if ((!$self->config->{update_remote}) && (!$self->config->{update_branch})) {
+        eval {
```
Martchus, line 85: We probably want to use `Feature::Compat::Try` in new code.

lib/OpenQA/Git.pm
```diff
@@ -81,6 +81,20 @@ sub set_to_latest_master ($self, $args = undef) {
         return $self->_format_git_error($res, 'Unable to reset repository to origin/master') unless $res->{status};
     }
 
+    if ((!$self->config->{update_remote}) && (!$self->config->{update_branch})) {
```
Martchus, line 84: It would be good to add a comment explaining what this code does and more importantly *why*. (The "what" can also be covered by extracting this into a separate function with a good name.)

gauravpathak: Moving this to a new function can cause significant changes, since `$self->config->{update_remote}` is already being used in the same function `set_to_latest_master()`.
I prefer adding a comment to explain what this code does (maybe it can also cover _why_ as well)  ;-)

lib/OpenQA/Git.pm
gauravpathak: The file `Git.pm` already using it https://github.com/os-autoinst/openQA/blob/master/lib/OpenQA/Git.pm#L12

lib/OpenQA/Git.pm
```diff
@@ -81,6 +81,22 @@ sub set_to_latest_master ($self, $args = undef) {
         return $self->_format_git_error($res, 'Unable to reset repository to origin/master') unless $res->{status};
     }
 
+=comment
```
perlpunk, line 84: That's usually not how pod looks like.
We can decide to document the whole function, but a lonely `=comment` directive somewhere is very uncommon. Also a `=comment` directive does not exist, as far as I know.
There is `=for comment`, but that's actually for being **not** displayed when using perldoc.

also a pod directive should have an empty line before and after.

Does `perldoc lib/OpenQA/Git.pm` show something with this change?

perlpunk: https://perldoc.perl.org/perlpod

lib/OpenQA/Git.pm
perlpunk: The above reset is only done when `update_branch` and `do_cleanup` is set.
Should we also check for `do_cleanup` here?

gauravpathak:
I will look into the review comments

lib/OpenQA/Git.pm
```diff
@@ -65,22 +65,50 @@ sub _format_git_error ($self, $result, $error_message) {
...
 }
 
+=over 4
+
+=item set_to_latest_master()
+
+This function is used to checkout the latest commit from a git branch provided via update_branch
+and update_remote configuration variables in openqa.ini. In case if both the configuration variables
+are not provided then this function sets the default git branch by first extracting the git URL by using
+get_origin_url function and then setting the default git branch from the extracted git URL.
```
Martchus, line 75:
```suggestion
get_origin_url and then setting the default git branch from the extracted git URL.
```

lib/OpenQA/Git.pm
```diff
@@ -65,22 +65,50 @@ sub _format_git_error ($self, $result, $error_message) {
...
 sub set_to_latest_master ($self, $args = undef) {
     $self->_validate_attributes;
     if (my $update_remote = $self->config->{update_remote}) {
-        my $res = $self->_run_cmd(['remote', 'update', $update_remote]);
-        return $self->_format_git_error($res, 'Unable to fetch from origin master') unless $res->{status};
+        try {
+            my $current_branch = $self->get_current_branch;
+            $self->fetch($current_branch);
+        }
+        catch ($e) { return $e; }
```
Martchus, line 88:
```suggestion
        try { $self->fetch($self->get_current_branch) }
        catch ($e) { return $e }
```

Martchus (approved):
Don't have merge commits on you PR branch. Our workflow is to *rebase* PR branch.

lib/OpenQA/Git.pm
```diff
@@ -65,22 +65,47 @@ sub _format_git_error ($self, $result, $error_message) {
...
+=back
+
+=cut
+
 sub set_to_latest_master ($self, $args = undef) {
     $self->_validate_attributes;
     if (my $update_remote = $self->config->{update_remote}) {
-        my $res = $self->_run_cmd(['remote', 'update', $update_remote]);
-        return $self->_format_git_error($res, 'Unable to fetch from origin master') unless $res->{status};
+        try { $self->fetch($self->get_current_branch) }
```
perlpunk, line 84: I'm sorry, but in hindsight I think we shouldn't change the behaviour here, because I think it's behaving a little bit different from before.
So my suggestion was wrong.
And I think those settings should be deprecated anyway at some point.
So I would suggest to revert this and the change in the `update_branch` block.

lib/OpenQA/Git.pm
```diff
@@ -65,22 +65,47 @@ sub _format_git_error ($self, $result, $error_message) {
...
+            try {
+                $self->reset_hard('HEAD');
+            }
+            catch ($e) { return $e; }
         }
         my $res = $self->_run_cmd(['rebase', $update_branch]);
         return $self->_format_git_error($res, 'Unable to reset repository to origin/master') unless $res->{status};
     }
 
+    if ((length($self->config->{update_remote}) == 0) && (length($self->config->{update_branch}) == 0)) {
```
perlpunk, line 99:
```suggestion
    if (! $self->config->{update_remote} && ! $self->config->{update_branch}) {
```

PR #6367, opened 2025-04-07
Adjust limit_builds in parent_group_overview results

b10n1k:
With this commit, the limit_builds check is moved into the responsible template. This gives some flexibility over the resultset but it lets the compute_build_results to parse, prepare and return all of them. By default the function query the database for the first 400 jobs unless a bigger value is requested.

What change is
- *before*: when limit_builds is 10, by_builds lists the 10 latest builds. the by_group lists the same %result filtering by group. That was ignoring the latest jobs of the children group.
- *now": when limit_builds is 10, by_builds lists the 10 latest builds but the by_group lists the latest jobs within the limit_builds which have results for all the groups.

relative ticket: https://progress.opensuse.org/issues/179296

### DIFF
--- a/lib/OpenQA/BuildResults.pm
+++ b/lib/OpenQA/BuildResults.pm
@@ -181,7 +181,7 @@ sub compute_build_results ($group, $limit, $time_limit_days, $tags, $subgroup_fi
     my $max_jobs = 0;
     my $now = DateTime->now;
     for my $build (@builds) {
-        last if defined($limit) && (--$limit < 0);
+        #last if defined($limit) && (--$limit < 0);
 
         my ($version, $buildnr) = ($build->VERSION, $build->BUILD);
         my $jobs = $jobs_resultset->search(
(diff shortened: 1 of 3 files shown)


lib/OpenQA/BuildResults.pm
```diff
@@ -181,7 +181,7 @@ sub compute_build_results ($group, $limit, $time_limit_days, $tags, $subgroup_fi
     my $max_jobs = 0;
     my $now = DateTime->now;
     for my $build (@builds) {
-        last if defined($limit) && (--$limit < 0);
+        #last if defined($limit) && (--$limit < 0);
```
Martchus, line 184: This is normally limited to 3 or 10 depending on the page. I'm wondering on the performance impact of running this loop on all the builds on a production database like the OSD database. I can imagine that it becomes notably slower, especially on the index page.

Martchus (changes requested):

lib/OpenQA/BuildResults.pm
b10n1k: yea I guess that might be a concern. My thought was that the default 400 limit is not that bad.

b10n1k: btw this line can stay and return immediately if $limit == 0

b10n1k:
https://progress.opensuse.org/issues/180851

PR #6368, opened 2025-04-07
Consider all config files in unit for auto-reloading workers

Martchus:
With this the worker is also reloaded when drop-in config files are changed, when config files under `/usr/etc` are changed and when the client config is changed.

I tested locally whether this works for directories and it does. So the worker is reloaded when drop-in config files are added, changed or removed.


kalikiana (approved):
Good catch!

okurz (approved):

PR #6293, opened 2025-03-14
Rollback retries applied for full-stack test preventing ci from failing

Wabri:
The retry commit is now possible to rever thanks to the merged pr of Dominik Heidler #2676

This will revert the changes made by prs that mitigate the problem: #6244 and #6115

related: https://progress.opensuse.org/issues/175060

PR #6374, opened 2025-04-07
Add STABILITY_ALLOW_FAILURE option for script/retry

Wabri:
This could be use in stability tests in order to run over and over again a job to check if is stable and if not in what percentage. This is an example of output you can get if the option is enabled `Rerun 59 of 200 … with failures: 16 (27%)`.

Followup of #6293

### DIFF
--- a/tools/retry
+++ b/tools/retry
@@ -2,6 +2,7 @@
 # shellcheck disable=SC2048,SC2086
 
 RETRY="${RETRY:-3}"
+STABILITY_ALLOW_FAILURE="${STABILITY_ALLOW_FAILURE:-0}"
 STABILITY_TEST="${STABILITY_TEST:-0}"
 # Pass a hook command which is executed if a retry occurs
 HOOK="${HOOK:-}"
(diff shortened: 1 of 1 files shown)


okurz (commented):
Could be useful but as stated before why rewrite https://github.com/okurz/retry/blob/main/count-fail-ratio?

Wabri:
What happens if one of these days you decide to stop working on that personal tool?

okurz:
> What happens if one of these days you decide to stop working on that personal tool?

Fork it. It's free software with a proper license :) But we can consider to move it to a non-personal scope as well if that's important

Wabri:
Yes, you are right, we can fork it of course! I didn't know about that is in the official opensuse repository :star2:

I leave the decision to you, just want to move this into a proper PR and open the discussion.

okurz:
> Yes, you are right, we can fork it of course! I didn't know about that is in the official opensuse repository 🌟
> 
> I leave the decision to you, just want to move this into a proper PR and open the discussion.

Well, given that we have a very similar tool used in like 10 other places I think it's confusing to still maintain a slightly different variant in openQA hence I recommend to replace tools/retry completely by the use of "retry" and "count-fail-ratio" where needed. How do others think about that?

PR #6376, opened 2025-04-08
Ensure `…-restart@.path` units don't run into inotify limit

Martchus:
After recent changes to the path units more inotify watches are requires. The default limit is quite low (at least on openSUSE) so we need to put a higher limit in place.


Martchus:
Note that `/usr/lib/sysctl.d/50-default.conf` provided by `aaa_base` already sets `fs.inotify.max_user_watches = 65536` which is also effective on our systems. However, we also need to increase `fs.inotify.max_user_instances` as we otherwise still run into https://github.com/systemd/systemd/issues/4246.

perlpunk (approved):

kalikiana (approved):

PR #6381, opened 2025-04-09
Add signal guard for schedule_iso minion tasks

b10n1k:
Similar to cd99074 to prevent `Job terminated unexpectedly` on results. This ensures that the minion job will retry in case of interruptions, like network glitch in the system, without kill the process.

https://progress.opensuse.org/issues/108983

### DIFF
--- a/lib/OpenQA/Schema/Result/ScheduledProducts.pm
+++ b/lib/OpenQA/Schema/Result/ScheduledProducts.pm
@@ -192,11 +192,16 @@ from B<_generate_jobs()>. Returns a list of job ids from the jobs that were succ
 and a list of failure reason for the jobs that could not be scheduled. Internal function, not
 exported - but called by B<create()>.
 
+=item $guard
+
+Expects an C<OpenQA::Task::SignalGuard> object. If the related Minion job is aborting after the database
+transaction has completed, a restart of the Minion job is prevented so jobs aren't created twice.
+
 =back
 
 =cut
 
-sub schedule_iso ($self, $args) {
+sub schedule_iso ($self, $args, $guard) {
     # update status to SCHEDULING or just return if the job was updated otherwise
     return undef unless $self->_update_status_if(SCHEDULING, status => ADDED);
     $self->{_settings} = $args;
@@ -204,7 +209,7 @@ sub schedule_iso ($self, $args) {
     # schedule the ISO
     $self->discard_changes;
     my $result = do {
-        try { $self->_schedule_iso($args) }
+        try { $self->_schedule_iso($args, $guard) }
         catch ($e) { {error => $e} }
     };
     $self->set_done($result);
@@ -316,9 +321,7 @@ Internal function to actually schedule the ISO, see schedule_iso().
 
 =cut
 
-sub _schedule_iso {
-    my ($self, $args) = @_;
-
+sub _schedule_iso ($self, $args, $guard) {
     my @notes;
     my $schema = $self->result_source->schema;
     my $user_id = $self->user_id;
@@ -411,6 +414,7 @@ sub _schedule_iso {
         @successful_job_ids = ();
     }
 
+    $guard->retry(0) if defined($guard);
     # emit events
     for my $succjob (@successful_job_ids) {
         OpenQA::Events->singleton->emit_event('openqa_job_create', data => {id => $succjob}, user_id => $user_id);
--- a/lib/OpenQA/Task/Iso/Schedule.pm
+++ b/lib/OpenQA/Task/Iso/Schedule.pm
@@ -5,6 +5,7 @@ package OpenQA::Task::Iso::Schedule;
 use Mojo::Base 'Mojolicious::Plugin', -signatures;
 
 use OpenQA::Utils 'format_tx_error';
+use OpenQA::Task::SignalGuard;
 use Mojo::URL;
 use Mojo::UserAgent;
 
@@ -37,12 +38,14 @@ sub _download_scenario_definitions ($minion_job, $scheduled_product, $scheduling
 }
 
 sub _schedule_iso ($app, $minion_job, $args, @) {
+    my $ensure_task_retry_on_termination_signal_guard = OpenQA::Task::SignalGuard->new($minion_job);
     my $scheduled_product_id = $args->{scheduled_product_id};
     my $scheduling_params = $args->{scheduling_params};
     return $minion_job->fail({error => "Scheduled product with ID $scheduled_product_id does not exist."})
       unless my $scheduled_product = $app->schema->resultset('ScheduledProducts')->find($scheduled_product_id);
     return undef unless _download_scenario_definitions($minion_job, $scheduled_product, $scheduling_params);
-    $minion_job->finish($scheduled_product->schedule_iso($scheduling_params));
+    $minion_job->finish(
+        $scheduled_product->schedule_iso($scheduling_params, $ensure_task_retry_on_termination_signal_guard));
 }
 
 1;
--- a/t/39-scheduled_products-table.t
+++ b/t/39-scheduled_products-table.t
@@ -9,6 +9,7 @@ use lib "$FindBin::Bin/lib", "$FindBin::Bin/../external/os-autoinst-common/lib";
 use Test::Mojo;
 use Test::Warnings ':report_warnings';
 use Test::MockModule;
+use Test::MockObject;
 use Test::Output qw(combined_like);
 use OpenQA::Jobs::Constants;
 use OpenQA::Schema::Result::ScheduledProducts qw(ADDED SCHEDULED SCHEDULING CANCELLING CANCELLED);
@@ -33,23 +34,22 @@ my %settings = (
     user_id => $user->id,
 );
 
-# prevent job creation
+my $signal_guard = Test::MockObject->new->mock(retry => sub { undef });
 my $scheduled_products_mock = Test::MockModule->new('OpenQA::Schema::Result::ScheduledProducts');
-$scheduled_products_mock->redefine(_generate_jobs => sub { return undef; });
+$scheduled_products_mock->redefine(_generate_jobs => undef);    # prevent job creation
 
 my $scheduled_product;
 subtest 'handling assets with invalid name' => sub {
     $scheduled_product = $scheduled_products->create(\%settings);
-
     $schema->txn_begin;
     is_deeply(
-        $scheduled_product->schedule_iso({REPO_0 => ''}),
+        $scheduled_product->schedule_iso({REPO_0 => ''}, $signal_guard),
         {error => 'Asset type and name must not be empty.'},
         'schedule_iso prevents adding assets with empty name',
     );
     $schema->txn_rollback;
 
-    like $scheduled_product->schedule_iso({_DEPRIORITIZEBUILD => 1, TEST => 'foo'})->{error},
+    like $scheduled_product->schedule_iso({_DEPRIORITIZEBUILD => 1, TEST => 'foo'}, undef)->{error},
       qr/One must not specify TEST and _DEPRIORITIZEBUILD.*/,
       'schedule_iso prevents deprioritization/obsoletion when scheduling single scenario';
 
@@ -58,7 +58,7 @@ subtest 'handling assets with invalid name' => sub {
 
     $scheduled_product = $scheduled_products->create(\%settings);
     is_deeply(
-        $scheduled_product->schedule_iso({REPO_0 => 'invalid'}),
+        $scheduled_product->schedule_iso({REPO_0 => 'invalid'}, $signal_guard),
         {
             successful_job_ids => [],
             failed_job_info => [],
@@ -70,7 +70,7 @@ subtest 'handling assets with invalid name' => sub {
     is $scheduled_product->status, SCHEDULED, 'product marked as scheduled, though';
 };
 
-is $scheduled_product->schedule_iso(\%settings), undef, 'scheduling the same product again prevented';
+is $scheduled_product->schedule_iso(\%settings, $signal_guard), undef, 'scheduling the same product again prevented';
 
 subtest 'asset registration on scheduling' => sub {
     my $assets = $schema->resultset('Assets');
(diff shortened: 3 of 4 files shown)


Martchus (changes requested):
This change is actually trivial enough that you don't have to test it locally. (When I introduced `OpenQA::Task::SignalGuard` I tested its behavior locally and added unit tests.)

The only problem that comes to mind is receiving `SIGTERM` right after the database transaction creating all the jobs has been concluded. Then we would retry the Minion job and schedule all those jobs again.

It would make sense to call `$ensure_task_retry_on_termination_signal_guard->retry(0)` after the transaction so the Minion job can just finish in this case (which is hopefully not too time consuming). Unfortunately, "after the transaction" means within `_create_jobs_in_database` so one needed to pass down the guard object there.

If everything that comes after the transaction seems to be too time consuming for going on with it we could also call `$ensure_task_retry_on_termination_signal_guard->abort(1)` after the transaction instead. This would leave the remaining work undone (e.g. emitting events) but would guarantee a quick exit without scheduling jobs twice.

b10n1k:
> This change is actually trivial enough that you don't have to test it locally. (When I introduced `OpenQA::Task::SignalGuard` I tested its behavior locally and added unit tests.)
> 
> The only problem that comes to mind is receiving `SIGTERM` right after the database transaction creating all the jobs has been concluded. Then we would retry the Minion job and schedule all those jobs again.
> 
> It would make sense to call `$ensure_task_retry_on_termination_signal_guard->retry(0)` after the transaction so the Minion job can just finish in this case (which is hopefully not too time consuming). Unfortunately, "after the transaction" means within `_create_jobs_in_database` so one needed to pass down the guard object there.
> 
> If everything that comes after the transaction seems to be too time consuming for going on with it we could also call `$ensure_task_retry_on_termination_signal_guard->abort(1)` after the transaction instead. This would leave the remaining work undone (e.g. emitting events) but would guarantee a quick exit without scheduling jobs twice.

check https://github.com/os-autoinst/openQA/commit/cbe26b94889c19ac0edf44d53819e7c7e5a00b3c
I will expand some thoughts here for reviewer:

💡likely the change in t/lib/OpenQA/Test/Utils.pm is not mandatory. I thought it will just not make it confusing when you compare with the original function. But other than that it is not really useful

💡I dont know if I should go the next step to add the guard also in API:V1:Iso. it could be another commit but maybe the smaller this one is the better

t/39-scheduled_products-table.t
```diff
@@ -9,8 +9,10 @@ use lib "$FindBin::Bin/lib", "$FindBin::Bin/../external/os-autoinst-common/lib";
 use Test::Mojo;
 use Test::Warnings ':report_warnings';
 use Test::MockModule;
+use Test::MockObject;
 use OpenQA::Jobs::Constants;
 use OpenQA::Schema::Result::ScheduledProducts qw(ADDED SCHEDULED SCHEDULING CANCELLING CANCELLED);
+use OpenQA::Task::SignalGuard;
```
Martchus, line 15: Please use some kind of fake object in *all* these tests. We probably don't want to influence the signal handling of unit tests.

lib/OpenQA/Schema/Result/ScheduledProducts.pm
```diff
@@ -319,7 +319,7 @@ Internal function to actually schedule the ISO, see schedule_iso().
 =cut
 
 sub _schedule_iso {
-    my ($self, $args) = @_;
+    my ($self, $args, $guard) = @_;
```
Martchus, line 322: I recommend using signatures in new code.

Martchus (changes requested):

lib/OpenQA/Schema/Result/ScheduledProducts.pm
```diff
@@ -319,7 +319,7 @@ Internal function to actually schedule the ISO, see schedule_iso().
 =cut
 
 sub _schedule_iso {
```
Wabri, line 321:
```suggestion
sub _schedule_iso ($self, $args, $guard) {
```

lib/OpenQA/Schema/Result/ScheduledProducts.pm
Wabri:
```suggestion
```

lib/OpenQA/Task/Iso/Schedule.pm
```diff
@@ -37,12 +38,14 @@ sub _download_scenario_definitions ($minion_job, $scheduled_product, $scheduling
...
 sub _schedule_iso ($app, $minion_job, $args, @) {
+    my $ensure_task_retry_on_termination_signal_guard = OpenQA::Task::SignalGuard->new($minion_job);
     my $scheduled_product_id = $args->{scheduled_product_id};
     my $scheduling_params = $args->{scheduling_params};
     return $minion_job->fail({error => "Scheduled product with ID $scheduled_product_id does not exist."})
       unless my $scheduled_product = $app->schema->resultset('ScheduledProducts')->find($scheduled_product_id);
     return undef unless _download_scenario_definitions($minion_job, $scheduled_product, $scheduling_params);
-    $minion_job->finish($scheduled_product->schedule_iso($scheduling_params));
+    $minion_job->finish(
+        $scheduled_product->schedule_iso($scheduling_params, $ensure_task_retry_on_termination_signal_guard));
```
Wabri, line 48:
```suggestion
    $minion_job->finish($scheduled_product->schedule_iso($scheduling_params, $ensure_task_retry_on_termination_signal_guard));
```

t/api/02-iso-download.t
```diff
@@ -67,30 +72,31 @@ sub job_gru ($job_id) { $gru_dependencies->search({job_id => $job_id})->single->
 my $expected_job_count = 10;
 
 # Schedule download of an existing ISO
-$rsp = schedule_iso($t, {%iso, ISO_URL => 'http://localhost/openSUSE-13.1-DVD-i586-Build0091-Media.iso'});
+$rsp
+  = schedule_iso($t, {%iso, ISO_URL => 'http://localhost/openSUSE-13.1-DVD-i586-Build0091-Media.iso'}, $signal_guard);
```
Wabri, line 76:
```suggestion
$rsp = schedule_iso($t, {%iso, ISO_URL => 'http://localhost/openSUSE-13.1-DVD-i586-Build0091-Media.iso'}, $signal_guard);
```

t/api/02-iso.t
```diff
@@ -87,11 +92,12 @@ subtest 'group filter and priority override' => sub {
...
 
-    my $res = schedule_iso($t, {%iso, PRECEDENCE => 'original', _GROUP => 'invalid group name'});
+    my $res = schedule_iso($t, {%iso, PRECEDENCE => 'original', _GROUP => 'invalid group name'}, $signal_guard);
     is($res->json->{count}, 0, 'no jobs created if group invalid');
 
-    $res = schedule_iso($t,
-        {%iso, PRECEDENCE => 'original', _GROUP => 'opensuse test', _FOO => 'bar', __FOO_URL => 'bar'});
+    $res
+      = schedule_iso($t, {%iso, PRECEDENCE => 'original', _GROUP => 'opensuse test', _FOO => 'bar', __FOO_URL => 'bar'},
+        $signal_guard);
```
Wabri, line 100:
```suggestion
    $res = schedule_iso($t, {%iso, PRECEDENCE => 'original', _GROUP => 'opensuse test', _FOO => 'bar', __FOO_URL => 'bar'}, $signal_guard);
```

t/api/02-iso.t
```diff
@@ -332,12 +340,13 @@ $t->post_ok("/api/v1/isos/$iso/cancel")->status_is(200);
...
-schedule_iso($t, {%iso, FLAVOR => 'cherry'}, 200, {}, 'no product found');
-schedule_iso($t, {%iso, _GROUP_ID => 12345}, 404, {}, 'no templates found');
+schedule_iso($t, {iso => $iso, tests => "kde/usb"}, $signal_guard, 400, {}, 'invalid parameters');
+schedule_iso($t, {%iso, FLAVOR => 'cherry'}, $signal_guard, 200, {}, 'no product found');
+schedule_iso($t, {%iso, _GROUP_ID => 12345}, $signal_guard, 404, {}, 'no templates found');
 
 # handle list of tests
-$res = schedule_iso($t, {%iso, TEST => 'server,kde,textmode', _OBSOLETE => 1, _FORCE_OBSOLETE => 1}, 200);
+$res
+  = schedule_iso($t, {%iso, TEST => 'server,kde,textmode', _OBSOLETE => 1, _FORCE_OBSOLETE => 1}, $signal_guard, 200);
```
Wabri, line 349:
```suggestion
$res = schedule_iso($t, {%iso, TEST => 'server,kde,textmode', _OBSOLETE => 1, _FORCE_OBSOLETE => 1}, $signal_guard, 200);
```

t/api/02-iso.t
```diff
@@ -927,15 +937,16 @@ subtest 'schedule tests correctly when changing TEST to job template name' => su
     );
     like($failed_message->{job_id}, qr/\d+/, 'child_test was scheduled');
 
-    $res = schedule_iso($t, {%iso, _GROUP_ID => '1002', TEST => 'child_variant1'});
+    $res = schedule_iso($t, {%iso, _GROUP_ID => '1002', TEST => 'child_variant1'}, $signal_guard);
     is($res->json->{count}, 2, 'both child and parent jobs were triggered successfully');
 
-    $res = schedule_iso($t, {%iso, _GROUP_ID => '1002', TEST => 'child_variant1', _SKIP_CHAINED_DEPS => 1});
+    $res
+      = schedule_iso($t, {%iso, _GROUP_ID => '1002', TEST => 'child_variant1', _SKIP_CHAINED_DEPS => 1}, $signal_guard);
```
Wabri, line 944:
```suggestion
    $res = schedule_iso($t, {%iso, _GROUP_ID => '1002', TEST => 'child_variant1', _SKIP_CHAINED_DEPS => 1}, $signal_guard);
```

Wabri (changes requested):

Wabri:
I didn't run tidy locally, but for me have declaration and assignation in separate lines are really awful. I'm going to resolve all the conversation related to that, but still for me is not tidy at all.

t/api/02-iso.t
Wabri: Not necessary

t/api/02-iso.t
Wabri: Not necessary

t/api/02-iso.t
Wabri: Not necessary

lib/OpenQA/Task/Iso/Schedule.pm
Wabri: Not necessary

b10n1k:
💡Mock the `OpenQA::Task::SignalGuard` for all the tests. However I found that there are other tests which seems to use the real object.
💡Update schedule_iso to reflect the new parameter
💡schedule_iso with signature now (there are other functions without the new signature definition but I didnt touch them)

t/39-scheduled_products-table.t
```diff
@@ -32,23 +33,23 @@ my %settings = (
     user_id => $user->id,
 );
 
-# prevent job creation
+my $signal_guard = Test::MockObject->new->mock(retry => sub { undef });
+
 my $scheduled_products_mock = Test::MockModule->new('OpenQA::Schema::Result::ScheduledProducts');
 $scheduled_products_mock->redefine(_generate_jobs => sub { return undef; });
```
Martchus, line 39: I'd preserve the comment and avoid the blank line.

```suggestion
my $signal_guard = Test::MockObject->new->mock(retry => sub { undef });
my $scheduled_products_mock = Test::MockModule->new('OpenQA::Schema::Result::ScheduledProducts');
$scheduled_products_mock->redefine(_generate_jobs => undef);  # prevent job creation
```

t/lib/OpenQA/Test/Utils.pm
```diff
@@ -607,7 +607,7 @@ sub mock_io_loop (%args) {
 
 sub assume_all_assets_exist { OpenQA::Schema->singleton->resultset('Assets')->search({})->update({size => 0}) }
 
-sub schedule_iso ($t, $args, $status = 200, $query_params = {}, $msg = undef) {
```
Martchus, line 610: You're adding a new parameter here but it isn't used. So I suggest you leave that change out.

lib/OpenQA/Schema/Result/ScheduledProducts.pm
```diff
@@ -192,19 +192,24 @@ from B<_generate_jobs()>. Returns a list of job ids from the jobs that were succ
 and a list of failure reason for the jobs that could not be scheduled. Internal function, not
 exported - but called by B<create()>.
 
+=item $guard
+
+Expects an C<OpenQA::Task::SignalGuard> object. If database transcation fails during the scheduling,
+this ensures that the minion job will be restarted.
```
Martchus, line 198: First a spelling/grammar nitpick:

```suggestion
Expects an C<OpenQA::Task::SignalGuard> object. If the database transaction fails during the scheduling,
this ensures that the Minion job will be restarted.
```

But this explanation is also completely the opposite of what passing the `$guard` will achieve here. It achieves the following:

```suggestion
Expects an C<OpenQA::Task::SignalGuard> object. If the related Minion job is aborting after the database
transaction has completed, a restart of the Minion job is prevented so jobs aren't created twice.
```

Martchus (changes requested):

b10n1k:
💡Revert all the changes in the tests which use the helper schedule_iso

t/39-scheduled_products-table.t
```diff
@@ -32,23 +33,22 @@ my %settings = (
     user_id => $user->id,
 );
 
-# prevent job creation
+my $signal_guard = Test::MockObject->new->mock(retry => sub { undef });
 my $scheduled_products_mock = Test::MockModule->new('OpenQA::Schema::Result::ScheduledProducts');
-$scheduled_products_mock->redefine(_generate_jobs => sub { return undef; });
+$scheduled_products_mock->redefine(_generate_jobs => sub { return undef; });    # prevent job creation
```
Martchus, line 38:
```suggestion
$scheduled_products_mock->redefine(_generate_jobs => undef);    # prevent job creation
```

Martchus (approved):
You'll need to resolve conflicts, though. I also have one small improvement you might want to add since you'll have to push again anyway.

Martchus:
I guess now only tidy complains.

Wabri (approved):

PR #6382, opened 2025-04-09
Wrap prove to prevent unhandled output

Wabri:
related: https://progress.opensuse.org/issues/178117

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -16,6 +16,8 @@ HELM_TEST ?= 1
 TESTS ?=
 # EXTRA_PROVE_ARGS: Additional prove arguments to pass
 EXTRA_PROVE_ARGS ?=
+# PROVE: Test application for Perl tests
+PROVE ?= prove
 ifeq ($(TESTS),)
 PROVE_ARGS ?= --trap -r ${EXTRA_PROVE_ARGS} t
 else
@@ -240,7 +242,7 @@ test-unit-and-integration: node_modules
 	export GLOBIGNORE="$(GLOBIGNORE)";\
 	export DEVEL_COVER_DB_FORMAT=JSON;\
 	export PERL5OPT="$(COVEROPT)$(PERL5OPT) -It/lib -I$(PWD)/t/lib -I$(PWD)/external/os-autoinst-common/lib -MOpenQA::Test::PatchDeparse";\
-	RETRY=${RETRY} HOOK=./tools/delete-coverdb-folder timeout -s SIGINT -k 5 -v ${TIMEOUT_RETRIES} tools/retry prove ${PROVE_LIB_ARGS} ${PROVE_ARGS}
+	RETRY=${RETRY} HOOK=./tools/delete-coverdb-folder timeout -s SIGINT -k 5 -v ${TIMEOUT_RETRIES} tools/retry "${PROVE}" ${PROVE_LIB_ARGS} ${PROVE_ARGS}
 
 .PHONY: setup-database
 setup-database:
--- a/docs/Contributing.asciidoc
+++ b/docs/Contributing.asciidoc
@@ -267,6 +267,18 @@ We use annotations in some places to mark "uncoverable" code such as this:
 
 See the docs for details https://metacpan.org/pod/Devel::Cover
 
+To run the tests with a different command than `prove` you can set the
+`PROVE` variable in the Makefile. For example, in our CircleCI tests we
+use `tools/prove_wrapper` to run the tests. This wrapper is a simple script that
+will run `prove` and fail the test if there is unhandled output. This is useful
+to avoid having unhandled output in tests. To run the tests with
+`tools/prove_wrapper` you can do:
+
+[source,sh]
+----
+PROVE=tools/prove_wrapper make test TESTS=t/foo.t
+----
+
 There are some ways to save some time when executing local tests:
 
 * One option is selecting individual tests to run as explained above
--- a/tools/ci/run_unit_tests.sh
+++ b/tools/ci/run_unit_tests.sh
@@ -22,4 +22,5 @@ sudo chown -R $USER test-results
 export OPENQA_TEST_TIMEOUT_SCALE_CI=3
 export EXTRA_PROVE_ARGS="-v"
 export OPENQA_FULLSTACK_TEMP_DIR=$PWD/test-results/fullstack
+export PROVE="tools/prove_wrapper"
 make test-$target
--- a/tools/prove_wrapper
+++ b/tools/prove_wrapper
@@ -1,0 +1,31 @@
+#!/bin/bash
+
+set -euo pipefail
+
+OUTPUT=$(mktemp)
+
+echo "Running prove with TAP output check ..."
+
+prove "$@" 2>&1 | tee "$OUTPUT"
+STATUS=${PIPESTATUS[0]}
+
+if [ "$STATUS" -ne 0 ]; then
+    echo "not ok - prove failed"
+    exit "$STATUS"
+fi
+
+sed -E '
+  s/^\[[0-9]{2}:[0-9]{2}:[0-9]{2}\][[:space:]]*//
+  /^All tests successful\./,$d
+' "$OUTPUT" > "$OUTPUT.processed"
+
+UNHANDLED=$(grep -vE '^t/.*([0-9]{2}-|)([[:alpha:]]|_|-)+\.t \.+' "$OUTPUT.processed" || true)
+
+if [ -n "$UNHANDLED" ]; then
+    echo "not ok - unhandled output found"
+    echo "Run with PROVE_COMMAND=tools/prove_wrapper to reproduce locally"
+    exit 1
+else
+    echo "ok - no unhandled output found"
+    exit 0
+fi


tools/prove_wrapper
```diff
@@ -0,0 +1,25 @@
...
+OUTPUT=$(prove "$@" 2>&1)
+STATUS=$?
+
+echo "Output:"
+echo "$OUTPUT"
+
+if [ "$STATUS" -ne 0 ]; then
+    echo "Failed with status $STATUS"
+    exit $STATUS
+fi
```
okurz, line 14: As @perlpunk and me suggested this looks like you want to use `tee` instead. Also probably it's not a good idea to do `2>&1` because I assume one of the two is already properly handled? Also consider [Test::Warnings here](https://metacpan.org/pod/Test::Warnings)

okurz (changes requested):

perlpunk:
I think the acceptance criteria should be extended to say that we still want immediate output, not only after the last test has finished. If there is a circleci job that runs for 15 minutes, I don't want to wait until all tests have finished to see that already the first failed.
In many cases we will be fine with a delayed output, but sometimes it would be really annoying to have to wait.

tools/prove_wrapper
```diff
@@ -0,0 +1,25 @@
+#!/bin/sh
+# shellcheck disable=SC2048,SC2086
+
+echo "Running prove with TAP output check"
+OUTPUT=$(prove "$@" 2>&1)
```
Martchus, line 5: This seems bad as we would lose the ability to follow the output as it is written.

tools/prove_wrapper
```diff
@@ -0,0 +1,25 @@
...
+fi
+
+grep -vE '^(\[[0-9]{2}:[0-9]{2}:[0-9]{2})' "$OUTPUT" | grep -v '^\s*$' > "$UNHANDLED"
+
+if [ -s "$UNHANDLED" ]; then
+    echo "Unhandled output found:"
+    cat "$UNHANDLED"
+    exit 1
+else
+    echo "Check passed!"
```
Martchus, line 23:
```suggestion
    echo "No unhandled output found"
```

Or maybe even produce something that is in-line with tap output in all those `echo` calls.

Wabri: What do you mean?

Martchus: Something like:
```suggestion
    echo "ok - no unhandled output found"
```

perlpunk (commented):
Please make sure we can still call `make test` locally successfully. Currently it fails even if there is no unhandled output. That's because the output looks different.

Also, IMHO we should only do this in CircleCI and not when running locally, at least not by default. Other opinions? It might be annoying during developing.

Martchus:
I would prefer to have a local Makefile target (or some other means) to be able to invoke this check before creating a PR. However, it would probably make sense to not run it by default as part of the normal `test` target. Especially if it influences behavior negatively like cancelling a test via Ctrl+C it would make sense to avoid running this by default.

Wabri:
Good idea, I've temporarily create a new variable in the makefile, but I'm wondering if could it be something like "enable/disable" wrapper or choose the command

Wabri:
@Martchus did you mean to create a new make target just to run again the unit test and check the output from it?

Wabri:
Otherwise we can let circleci upload the artifact of the tests and fetch it using api in another job where we run the checks.

perlpunk:
> Otherwise we can let circleci upload the artifact of the tests and fetch it using api in another job where we run the checks.

I wouldn't make it that complicated. I don't see why it should be moved to a different workflow step.

Let's use the simple approach for now and use the variable ~PROVE_COMMAND` you introduced. We can improve it later if necessary.

tools/prove_wrapper
perlpunk: > Well, it catches warnings so I assume it tapped into `warn`.

Test::Warnings set `$SIG{__WARN__}`, but it does not override the `warn` function in any way.

> Maybe we can tap into "usual output" functions like "printf"? and catch output this way?

How do you want to differentiate between print statements we need and those which we don't need?
I don't think that's going to work.

perlpunk: > I found https://metacpan.org/pod/Capture::Tiny#tee which looks like it could be useful

Test::Output is using Capture::Tiny behind the scenes.

Martchus:
A different variable is perhaps better than a new target.

tools/prove_wrapper
Wabri: I'm not sure what both of you are talking about, but just for context i need to say that I've stopped, not even push, the experiment I was doing using perl to do this by now. Do you think I need to switch to it? I think is not necessary due to the not that complexity of the check we need, but critiques are welcome.

tools/prove_wrapper
```diff
@@ -0,0 +1,28 @@
...
+
+prove "$@" 2>&1 | tee "$OUTPUT"
+STATUS=${PIPESTATUS[0]}
+
+if [ "$STATUS" -ne 0 ]; then
+    echo "not ok - prove failed"
+    exit "$STATUS"
+fi
+
+sed -E '
```
Martchus, line 15: Does this already cover skipped tests as well? If not than that's probably needed.

Wabri: skipped test is considered as successful, so yes

Martchus (commented):
Looks good. Now only the "heavy" tests fail:

```
RETRY=0 HOOK=./tools/delete-coverdb-folder timeout -s SIGINT -k 5 -v $((25 * 2 * (0 + 1) ))m tools/retry tools/prove_wrapper -l --harness TAP::Harness::JUnit --timer t/14-grutasks.t t/32-openqa_client.t t/44-scripts.t 
Running prove with TAP output check ...
[2025-04-16 12:49:56.05844] [181] [debug] [pid:181] _carry_over_candidate(99983): _failure_reason=b:failed
[12:50:04] t/14-grutasks.t ....... ok    31095 ms ( 0.18 usr  0.01 sys + 28.69 cusr  1.36 csys = 30.24 CPU)
[12:50:23] t/32-openqa_client.t .. ok    19490 ms ( 0.02 usr  0.00 sys + 17.57 cusr  1.70 csys = 19.29 CPU)
[12:54:03] t/44-scripts.t ........ ok   219818 ms ( 0.02 usr  0.00 sys + 206.02 cusr 11.39 csys = 217.43 CPU)
[12:54:03]
All tests successful.
Files=3, Tests=136, 270.428 wallclock secs ( 0.23 usr  0.01 sys + 252.28 cusr 14.45 csys = 266.97 CPU)
Result: PASS
not ok - unhandled output found
make[2]: *** [Makefile:242: test-unit-and-integration] Error 1
```

I guess that's good in the way that it shows that this new check actually works. Maybe you can fix this instance of unhandled output so we can merge this PR cleanly?

Wabri:
I've open a new pr to check if the change I made comes from that change or not. For now I've included the commits coming from this pr, once we confirm that the unhandled output comes from that I'll remove them.

Makefile
```diff
@@ -16,6 +16,8 @@ HELM_TEST ?= 1
 TESTS ?=
 # EXTRA_PROVE_ARGS: Additional prove arguments to pass
 EXTRA_PROVE_ARGS ?=
+# PROVE_COMMAND: If you want to use a different command than prove
+PROVE_COMMAND ?= prove
```
kalikiana, line 20: So by default the new feature is not used, right? It would only be enabled in CI?
Could you add a sentence to the README on how to enable it?

tools/prove_wrapper
```diff
@@ -0,0 +1,28 @@
...
+  s/^\[[0-9]{2}:[0-9]{2}:[0-9]{2}\][[:space:]]*//
+  /^All tests successful\./,$d
+' "$OUTPUT" > "$OUTPUT.processed"
+
+UNHANDLED=$(grep -vE '^t/.*([0-9]{2}-|)([[:alpha:]]|_|-)+\.t \.+' "$OUTPUT.processed")
+
+if [ -n "$UNHANDLED" ]; then
+    echo "not ok - unhandled output found"
```
kalikiana, line 23:
```suggestion
    echo "not ok - unhandled output found"
    echo "Run with PROVE_COMMAND='tools/prove_wrapper' to reproduce locally" "
```

Makefile
Wabri: Yes, good idea :)

tools/prove_wrapper
perlpunk: It seems like you added this to the successful case now?

Wabri: Moved into the unsuccessful case :grin:

tools/prove_wrapper
```diff
@@ -0,0 +1,29 @@
+#!/bin/bash
+
```
okurz, line 2: By default for every bash script I recommend `set -euo pipefail` to prevent nasty surprises

okurz (changes requested):

tools/prove_wrapper
Wabri: Done

tools/prove_wrapper
```diff
@@ -0,0 +1,31 @@
...
+set -euo pipefail
+
+OUTPUT=$(mktemp)
+
+echo "Running prove with TAP output check ..."
```
b10n1k, line 7: small think but could we add some distinctive mark in the beginning of the string?

b10n1k (commented):
My test run with `❯ PROVE_COMMAND=tools/prove_wrapper make test HEAVY=1 TESTS="t/14-grutasks.t t/14-grutasks-git.t"`

it seems to work and tests pass but I missed the output from ` echo "ok - no unhandled output found"`. is that right?

tools/prove_wrapper
Wabri: Do you have any suggestion?

Wabri:
@b10n1k you are right, to emulate what has been done in CircleCI locally is necessary to run for example `timeout -s SIGINT -k 5 -v $((25 * 2 * (0 + 1) ))m tools/prove_wrapper -l --timer t/21-needles.t`:

```
Running prove with TAP output check ...
[10:17:53] t/21-needles.t .. ok     1808 ms ( 0.01 usr  0.00 sys +  1.34 cusr  0.32 csys =  1.67 CPU)
[10:17:53]
All tests successful.
Files=1, Tests=9,  2 wallclock secs ( 0.03 usr  0.00 sys +  1.34 cusr  0.32 csys =  1.69 CPU)
Result: PASS
ok - no unhandled output found
```

I've update the documentation accordingly.

tools/ci/run_unit_tests.sh
```diff
@@ -22,4 +22,5 @@ sudo chown -R $USER test-results
 export OPENQA_TEST_TIMEOUT_SCALE_CI=3
 export EXTRA_PROVE_ARGS="-v"
 export OPENQA_FULLSTACK_TEMP_DIR=$PWD/test-results/fullstack
+export PROVE_COMMAND="tools/prove_wrapper"
```
Martchus, line 25:
```suggestion
export PROVE=tools/prove_wrapper
```

Martchus (changes requested):

tools/ci/run_unit_tests.sh
Wabri: Ups, you are right! Updated!

Martchus (approved):

Makefile
```diff
@@ -16,6 +16,8 @@ HELM_TEST ?= 1
 TESTS ?=
 # EXTRA_PROVE_ARGS: Additional prove arguments to pass
 EXTRA_PROVE_ARGS ?=
+# PROVE: If you want to use a different command than prove
```
okurz, line 19: I would write

```suggestion
# PROVE: Test application for Perl tests
```

the string "If you want to use a different command than prove" is a bit incomplete

okurz (approved):

Makefile
```diff
@@ -16,6 +16,8 @@ HELM_TEST ?= 1
 TESTS ?=
 # EXTRA_PROVE_ARGS: Additional prove arguments to pass
 EXTRA_PROVE_ARGS ?=
+# PROVE: Test applicatiohn for Perl tests
```
okurz, line 19:
```suggestion
# PROVE: Test application for Perl tests
```

okurz (changes requested):

okurz (approved):

docs/Contributing.asciidoc
```diff
@@ -267,6 +267,17 @@ We use annotations in some places to mark "uncoverable" code such as this:
...
 
+To run the tests with a different command than `prove` you can set the
+`PROVE` variable in the Makefile. For example, in our CircleCI tests we
+use `tools/prove_wrapper` to run the tests. This wrapper is a simple script that
+will run `prove` and fail the test if there is unhandled output. This is useful to
+avoid having unhandled output in tests. To run the tests with `tools/prove_wrapper` you can do:
+
+[source,sh]
+----
+timeout -s SIGINT -k 5 -v $((25 * 2 * (0 + 1) ))m tools/prove_wrapper -l --timer t/foo.t
```
perlpunk, line 278: this looks a bit weird to me. why can't we use `make test` here?

Wabri: It's the fastest way to test the unhandled output, but to proper emulate the circleci and use make we can use:

```
RETRY=0 TIMEOUT_M=25 PROVE_ARGS='--timer t/foo.t' PROVE=tools/prove_wrapper make test-unit-and-integration
```

Wabri: without the `--timer` is failing due to `set -e` suggested to add here https://github.com/os-autoinst/openQA/pull/6382#discussion_r2054335351

Wabri: It's failing due to the grep that define `UNHANDLED` variable, I can suppress by adding a false positive `|| true` after the `grep`.

Wabri: I've update the script with `UNHANDLED=$(grep -vE '^t/.*([0-9]{2}-|)([[:alpha:]]|_|-)+\.t \.+' "$OUTPUT.processed" || true)` and the documentation accordingly `PROVE=tools/prove_wrapper make test TESTS=t/foo.t`

tools/prove_wrapper
b10n1k: Maybe `... Running prove with TAP output check` or `[CUSTOM_PROVE] Running prove with TAP output check` or something like that. but I guess you can merge either way for now as is and think about that (if that matters later)

docs/Contributing.asciidoc
```diff
@@ -267,6 +267,17 @@ We use annotations in some places to mark "uncoverable" code such as this:
 
 See the docs for details https://metacpan.org/pod/Devel::Cover
 
+To run the tests with a different command than `prove` you can set the
+`PROVE` variable in the Makefile. For example, in our CircleCI tests we
+use `tools/prove_wrapper` to run the tests. This wrapper is a simple script that
+will run `prove` and fail the test if there is unhandled output. This is useful to
+avoid having unhandled output in tests. To run the tests with `tools/prove_wrapper` you can do:
```
okurz, line 274: I suggest you configure your editor to consistently wrap at 80 characters. This line is too long

okurz (changes requested):

docs/Contributing.asciidoc
Wabri: I saw a lot more lines in the file that goes over the 80 character, does it still relevant in 2025?

okurz: Given how big some people scale their fonts when sharing their screen, yes :)

okurz (approved):

docs/Contributing.asciidoc
Wabri: anyway I updated the doc accordingly

r-richardson:
LGTM

PR #6390, opened 2025-04-14
Avoid security review by not packaging sysctl config for Tumbleweed

Martchus:
The config is not required because Tumbleweed has it anyway. So this file only needs to be packaged for Leap where no security review is required.

Security review/ticket: https://bugzilla.suse.com/show_bug.cgi?id=1241146

### DIFF
--- a/dist/rpm/openQA.spec
+++ b/dist/rpm/openQA.spec
@@ -387,6 +387,11 @@ export LANG=en_US.UTF-8
 rm -rf %{buildroot}/%{_sysusersdir}
 %endif
 
+%if 0%{?suse_version} > 1560
+# the limit set via sysctl is already present on Tumbleweed and packaging sysctl config would require a security review
+rm -rf %{buildroot}/%{_prefix}/lib/sysctl.d/01-openqa-reload-worker-auto-restart.conf
+%endif
+
 mkdir -p %{buildroot}%{_bindir}
 ln -s %{_datadir}/openqa/script/client %{buildroot}%{_bindir}/openqa-client
 ln -s %{_datadir}/openqa/script/openqa-cli %{buildroot}%{_bindir}/openqa-cli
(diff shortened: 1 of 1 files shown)


perlpunk (approved):

dist/rpm/openQA.spec
```diff
@@ -385,6 +385,9 @@ export LANG=en_US.UTF-8
 %if 0%{?suse_version} <= 1500
 # we only use sysusers on Tumbleweed
 rm -rf %{buildroot}/%{_sysusersdir}
+%else
+# the limit set via sysctl is already present on Tumbleweed and packaging sysctl config would require a security review
+rm -rf %{buildroot}/%{_prefix}/lib/sysctl.d/01-openqa-reload-worker-auto-restart.conf
```
wfrisch, line 390: IMHO, for the sysctl drop-in rm -rf, it should be `0%{?suse_version} > 1560`. Otherwise it would be deleted on Leap 15.6, where you still need it.

wfrisch (changes requested):
See comment

dist/rpm/openQA.spec
wfrisch: Maybe better:

```
#if 0%{?suse_version} > 1560
# the limit set via sysctl is already present on Tumbleweed and packaging sysctl config would require a security review
rm -rf %{buildroot}/%{_prefix}/lib/sysctl.d/01-openqa-reload-worker-auto-restart.conf
%endif
```

and the revert the modifiction to the sysusers part.

dist/rpm/openQA.spec
```diff
@@ -387,6 +387,11 @@ export LANG=en_US.UTF-8
 rm -rf %{buildroot}/%{_sysusersdir}
 %endif
 
+%if 0%{?suse_version} > 1560
```
b10n1k, line 390: what happens in other distros? this will be always false, right?

Martchus: I think it'll be true on the corresponding SLE but on other distros like Fedora it will be always false. That's not a problem, we just don't package that file then which is the conservative approach.

PR #6391, opened 2025-04-15
Fix jobs sometimes ending up without group after retry on deadlocks

Martchus:
After adding a retry on deadlocks the function that inserts the jobs into the database might run multiple times. Hence it must not delete the group ID setting (or any other settings) from the previously computed hash. This change deletes the group ID only within `create_from_settings` which only works on a copy of the settings.

Related ticket: https://progress.opensuse.org/issues/180020


realcharmer:
Are the CI fails expected in this case?

Martchus:
I was actually hoping that it fails *somewhere* so I can use that as starting point for changing tests :-)
(Maybe I broke it completely, though.)

Wabri (approved):
LGTM

b10n1k (approved):
I tested local and approved. But I was wondering if it is worth a test on retry

PR #6396, opened 2025-04-16
RFC: Show changelog on web UI via GitHub

Martchus:
* Avoid parsing the RPM changelog
* Avoid having our own page for displaying the changelog (which does not handle non-ASCII characters well at this point)
* Simplify/remove our code
* See https://progress.opensuse.org/issues/180068

---

Not tested yet; this is just a draft so you can see how it would look like in code. There's probably more code which can be removed, e.g. tests of the changelog parsing.

### DIFF
--- a/lib/OpenQA/Utils.pm
+++ b/lib/OpenQA/Utils.pm
@@ -652,19 +652,17 @@ sub parse_tags_from_comments ($group, $res) {
 }
 
 sub detect_current_version ($path) {
-    # Get application version
     my $current_version = undef;
-    my $changelog_file = path($path, 'public', 'Changelog');
+    my $version_file = $path->child('version.txt');
+    if (-e $version_file) {
+        $current_version = $version_file->slurp;
+        chomp $current_version;
+        return $current_version;
+    }
+
     my $head_file = path($path, '.git', 'refs', 'heads', 'master');
     my $refs_file = path($path, '.git', 'packed-refs');
-
-    if (-e $changelog_file) {
-        my $changelog = $changelog_file->slurp;
-        if ($changelog && $changelog =~ /Update to version (\d+\.\d+\.(\b[0-9a-f]{5,40}\b))\:/mi) {
-            $current_version = $1;
-        }
-    }
-    elsif (-e $head_file && -e $refs_file) {
+    if (-e $head_file && -e $refs_file) {
         my $master_head = $head_file->slurp;
         my $packed_refs = $refs_file->slurp;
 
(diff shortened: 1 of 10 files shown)


lib/OpenQA/Utils.pm
```diff
@@ -652,19 +652,17 @@ sub parse_tags_from_comments ($group, $res) {
 }
 
 sub detect_current_version ($path) {
-    # Get application version
     my $current_version = undef;
-    my $changelog_file = path($path, 'public', 'Changelog');
+    my $version_file = $path->child('version.txt');
+    if (-e $version_file) {
```
b10n1k, line 657: Since you go with `version.txt` the `$current_version` should be always available, unless it is an older openQA version. So I believe the condition needs something more.

templates/webapi/layouts/navbar.html.ep
```diff
@@ -97,7 +97,9 @@
                   % }
                   %= link_to "Appearance" => url_for('appearance') => class => 'dropdown-item'
                   %= link_to "API help" => url_for('api_help') => class => 'dropdown-item'
-                  %= link_to 'Changelog' => url_for('changelog') => class => 'dropdown-item'
+                  % if ($current_version) {
```
b10n1k, line 100: Taking https://github.com/os-autoinst/openQA/pull/6396/files#r2048260791 in mind, I would say that this element should be always displayed.

kalikiana: Only from rpm packaging. The .spec is where version.txt is created.

I wonder if the version.txt could be prepared as part of the build system, though?

Martchus: I created it in the .spec file because there we have the version available as rpm macro/variable. We could of course pass it from the .spec file to the Makefile via an env variable. I find it a bit of an overcomplication, though.

lib/OpenQA/WebAPI/Plugin/Helpers.pm
```diff
@@ -381,6 +381,11 @@ sub register ($self, $app, $config) {
             }
             return $regex_problem && $context ? "$context: $regex_problem" : $regex_problem;
         });
+
+    $app->helper(url_for_changelog => sub ($c, $v) {
+        $v =~ s/.*-//;
+        return "https://github.com/os-autoinst/openQA/commits/$v";
```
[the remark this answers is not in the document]
Martchus: No, we just render a link to GitHub. Only if a user clicks on the link GitHub is accessed (by them).

PR #6399, opened 2025-04-19
Fix URLs in admin/productlog

rclobus:
Fixes the URLs for each row in the table

### DIFF
--- a/assets/javascripts/audit_log.js
+++ b/assets/javascripts/audit_log.js
@@ -312,48 +312,41 @@ function loadProductLogTable(dataTableUrl, rescheduleUrlTemplate, showActions) {
       {
         targets: 6,
         render: function (data, type, row) {
-          const rowData = dataForLink(this);
-          const settings = rowData.settings;
-          const version = settings.VERSION;
-          const build = settings.BUILD;
-          const distri = settings.DISTRI;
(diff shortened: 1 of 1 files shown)


rclobus:
This is a follow-up for https://github.com/os-autoinst/openQA/pull/6217#issuecomment-2763343627, where I noted that in the unfiltered auditlog the URLs were not correct.

rclobus:
Set to draft, until I've fixed the automated checks

rclobus:
All automated tests passed now.

I've removed the 'if'-statement, because the values for distri, version, flavor etc. come directly from the row in the table, instead of via another function.

b10n1k:
> All automated tests passed now.
> 
> I've removed the 'if'-statement, because the values for distri, version, flavor etc. come directly from the row in the table, instead of via another function.

I would like to have this context in the commit message. Other than that LGTM

Martchus (approved):

rclobus:
> > I've removed the 'if'-statement, because the values for distri, version, flavor etc. come directly from the row in the table, instead of via another function.
> 
> I would like to have this context in the commit message. Other than that LGTM

Done.

okurz (approved):

PR #6405, opened 2025-04-23
Move database dump script

nicksinger:
Carbon copy of https://github.com/os-autoinst/sync-and-trigger/blob/2b891a67d7509aa589ac108f594d3235826f5ba9/dump-psql

Related issue: https://progress.opensuse.org/issues/181301

It would make sense to include this script here ~(and later package it as part of the [`openQA-local-db`-package](https://build.opensuse.org/projects/devel:openQA/packages/openQA/files/_service:obs_scm:openQA.spec?expand=1))~. I know there is a lot of improvement possible but for now I went with a pure copy to better trace the history of this script in the future.

### DIFF
--- a/script/dump-db
+++ b/script/dump-db
@@ -1,5 +1,6 @@
 #!/bin/bash
-# must run as geekotest
+set -euo pipefail
 
-pg_dump -Fc -c openqa  -f ~/SQL-DUMPS/`date -Idate`.dump
-find ~/SQL-DUMPS/ -mtime +7 -print0 | xargs --no-run-if-empty -0 rm
+DUMP_FOLDER="/var/lib/openqa/SQL-DUMPS"
+pg_dump -Fc -c openqa -f "${DUMP_FOLDER}/$(date -Idate).dump"
+find "${DUMP_FOLDER}" -mtime +7 -print0 -delete
--- a/script/dump-db
+++ b/script/dump-db
@@ -1,8 +1,8 @@
 #!/bin/bash
 set -euo pipefail
-DUMP_FOLDER="/var/lib/openqa/SQL-DUMPS"
-DAYS_TO_KEEP=7
-[ "$1" = "-h" ] || [ "$1" = "--help" ] && echo "Dump the openQA postgresql database to ${DUMP_FOLDER} and clean-up backups older than ${DAYS_TO_KEEP} days." && exit
+DUMP_FOLDER=${DUMP_FOLDER:-"/var/lib/openqa/SQL-DUMPS"}
+DAYS_TO_KEEP=${DAYS_TO_KEEP:-"7"}
+[[ $1 = '-h' ]] || [[ $1 = '--help' ]] && echo "Dump the openQA postgresql database to ${DUMP_FOLDER} and clean-up backups older than ${DAYS_TO_KEEP} days." && exit
 
 pg_dump -Fc -c openqa -f "${DUMP_FOLDER}/$(date -Idate).dump"
-find "${DUMP_FOLDER}" -mtime +${DAYS_TO_KEEP} -print0 -delete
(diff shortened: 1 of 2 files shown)


nicksinger:
seems like the tests don't like some styles in this script (understandable) but also call all scripts in this directory unconditionally. So either another adjustment is needed to create the backup directory first or make the folder available in the test beforehand.

I wait for some feedback first before moving forward with any of these actions.

script/dump-db
```diff
@@ -0,0 +1,5 @@
+#!/bin/bash
+# must run as geekotest
+
+pg_dump -Fc -c openqa  -f ~/SQL-DUMPS/`date -Idate`.dump
+find ~/SQL-DUMPS/ -mtime +7 -print0 | xargs --no-run-if-empty -0 rm
```
Martchus, line 5: Having this in an actually maintained repo will only increase the chance that somebody accidentally copies this.

Maybe this makes it at least less dangerous:

```suggestion
find /var/lib/openqa/SQL-DUMPS -mtime +7 -print0 -delete
```

It would probably still make more sense to keep a certain number of dumps.

The line above should probably use the full path as well.

Martchus (changes requested):

script/dump-db
```diff
@@ -1,6 +1,8 @@
 #!/bin/bash
 set -euo pipefail
-
 DUMP_FOLDER="/var/lib/openqa/SQL-DUMPS"
+DAYS_TO_KEEP=7
+[ "$1" = "-h" ] || [ "$1" = "--help" ] && echo "Dump the openQA postgresql database to ${DUMP_FOLDER} and clean-up backups older than ${DAYS_TO_KEEP} days." && exit
```
Martchus, line 5: Since we're using Bash here:
```suggestion
[[ $1 = '-h' ] || [[ $1 = '--help' ]] && echo "Dump the openQA PostgreSQL database to ${DUMP_FOLDER} and clean-up backups older than ${DAYS_TO_KEEP} days." && exit
```

script/dump-db
```diff
@@ -1,6 +1,8 @@
 #!/bin/bash
 set -euo pipefail
-
 DUMP_FOLDER="/var/lib/openqa/SQL-DUMPS"
```
Martchus, line 3: Normally I would suggest to replace `/var/lib` with `${OPENQA_BASEDIR:-/var/lib}`. Maybe we want to keep it explicit here, though (due to the danger of deleting too much files).

script/dump-db
```diff
@@ -1,6 +1,8 @@
...
+DAYS_TO_KEEP=7
+[ "$1" = "-h" ] || [ "$1" = "--help" ] && echo "Dump the openQA postgresql database to ${DUMP_FOLDER} and clean-up backups older than ${DAYS_TO_KEEP} days." && exit
+
 pg_dump -Fc -c openqa -f "${DUMP_FOLDER}/$(date -Idate).dump"
-find "${DUMP_FOLDER}" -mtime +7 -print0 -delete
+find "${DUMP_FOLDER}" -mtime +${DAYS_TO_KEEP} -print0 -delete
```
Martchus, line 8:
```suggestion
find "${DUMP_FOLDER}" -mtime "+${DAYS_TO_KEEP}" -print0 -delete
```

script/dump-db
nicksinger: adjusted. I assume you missed to double `]` at the first test? (I used you version and bash complained ;) )

script/dump-db
nicksinger: I would keep it explicit but I also modified the script that the caller can at least adjust these values.

okurz (approved):

Martchus (approved):

PR #6406, opened 2025-04-23
Avoid leftover Minion jobs after failing txn when scheduling jobs

Martchus:
Otherwise the Minion jobs will be retried forever. It would still make sense to limit the retry of those jobs in general.

Note that this is particularly problematic after a retry for the txn was introduced to deal with deadlocks.


Martchus:
The extended test fails only most of the time without the fix. However, we at least do have a unit test that reproduces the issue we see in production. We should probably merge https://github.com/os-autoinst/openQA/pull/6407 first as it is the simpler change and will also already help. I'll rebase my change after that.

okurz (approved):

kalikiana (approved):

PR #6412, opened 2025-04-28
Follow symlinks but only delete files with dump-db script

nicksinger:
This allows `$backup_dir` to be a symlink to a different location and should also increase stability a bit by only considering files to delete on cleanup.

Opposing to the suggestion in https://progress.opensuse.org/issues/181301#note-20 I decided *against* using a trailing `backup_dir/` to avoid a similar problem like we faced previously (variable being empty and then only having `/` as path remaining)


okurz (approved):

realcharmer (approved):

perlpunk (commented):
I had thought about `-L` as well, but I hesitated, as having a command that deletes files and follows symlinks is something I avoid :)

PR #6414, opened 2025-05-02
Block all distri git operations unless `scm = git` is set, add `git_auto_commit` setting, disentangle `git_auto_clone` and `git_auto_update`

AdamWill:
This is my attempt to resolve the issues I found diving into the `git_auto_update` and `git_auto_clone` stuff, which is new to me (haven't updated Fedora's openQA since last year). See https://github.com/os-autoinst/openQA/pull/6039#issuecomment-2845218269 for the notes that led me to this.

Most importantly, this attempts to disable all git cloning shenanigans if the global config section does not have `scm = git`, because the older automatic-needle-commit feature kinda establishes that expectation. Beyond that, we add a `git_auto_commit` setting so you can have autoclone and/or autoupdate enabled without autocommit, and try to disentangle `git_auto_clone` and `git_auto_update` so you can have `auto_update` working fully without `auto_clone` (which I don't think is currently possible).

### DIFF
--- a/lib/OpenQA/Setup.pm
+++ b/lib/OpenQA/Setup.pm
@@ -100,6 +100,7 @@ sub read_config ($app) {
             require_for_assets => 0,
         },
         'scm git' => {
+            git_auto_commit => 'yes',
             update_remote => '',
             update_branch => '',
             do_push => 'no',
--- a/lib/OpenQA/Task/Git/Clone.pm
+++ b/lib/OpenQA/Task/Git/Clone.pm
@@ -26,6 +26,9 @@ sub register ($self, $app, @) {
 sub _git_clone_all ($job, $clones) {
     my $app = $job->app;
     my $job_id = $job->id;
+    # failsafe check - we should never actually get here with git
+    # disabled, but just in case
+    return if (($app->config->{global}->{scm} // '') ne 'git');
 
     my $retry_delay = {delay => 30 + int(rand(10))};
     # Prevent multiple git_clone tasks for the same path to run in parallel
@@ -75,14 +78,13 @@ sub _git_clone_all ($job, $clones) {
 
         # unblock openQA jobs despite network errors under best-effort configuration
         my $retries = $job->retries;
-        my $git_config = $app->config->{'scm git'};
         my $max_retries = $ENV{OPENQA_GIT_CLONE_RETRIES} // 10;
         my $max_best_effort_retries = min($max_retries, $ENV{OPENQA_GIT_CLONE_RETRIES_BEST_EFFORT} // 2);
         my $gru_task_id = $job->info->{notes}->{gru_id};
         if (   $is_path_only
             && defined($gru_task_id)
             && ($error =~ m/disconnect|curl|stream.*closed|/i)
-            && $git_config->{git_auto_update_method} eq 'best-effort'
+            && $git->config->{git_auto_update_method} eq 'best-effort'
             && $retries >= $max_best_effort_retries)
         {
             $app->schema->resultset('GruDependencies')->search({gru_task_id => $gru_task_id})->delete;
--- a/lib/OpenQA/Utils.pm
+++ b/lib/OpenQA/Utils.pm
@@ -536,20 +536,24 @@ sub create_downloads_list ($job_settings) {
 sub create_git_clone_list ($job_settings, $clones = {}) {
     my $distri = $job_settings->{DISTRI};
     my $config = OpenQA::App->singleton->config;
-    if (($config->{global}->{scm} // '') eq 'git' && $config->{'scm git'}->{git_auto_update} eq 'yes') {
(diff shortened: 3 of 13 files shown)


okurz (approved):
Looks very clean

AdamWill:
so far this is all "on paper", I'll deploy it on Fedora staging instance tomorrow and make sure it at least does what's intended there (i.e., well, nothing). Obviously also want the tests to pass, I hope I got that finally this time, except this weird fullstack failure which AFAICT isn't actually related to the PR?

lib/OpenQA/Setup.pm
```diff
@@ -100,6 +100,7 @@ sub read_config ($app) {
             require_for_assets => 0,
         },
         'scm git' => {
+            git_auto_commit => 'yes',
```
b10n1k, line 103: would you mind to move this along with the other git_auto_* down in the L108?

AdamWill: I intentionally put it here because it is the precursor to the following options - `update_remote`, `update_branch`, `do_push` and `do_cleanup` all configure specific aspects of the `git_auto_commit` behaviour. To me it makes more sense to organize the options by what they do than by what they're called.

AdamWill:
sigh, I'll deal with codecov tomorrow.

b10n1k (approved):
LGTM

b10n1k:
set `not-ready` due to the codecov.

lib/OpenQA/Task/Git/Clone.pm
```diff
@@ -26,6 +26,12 @@ sub register ($self, $app, @) {
 sub _git_clone_all ($job, $clones) {
     my $app = $job->app;
     my $job_id = $job->id;
+    my $log = $app->log;
+    my $ctx = $log->context("[#$job_id]");
+    if (($app->config->{global}->{scm} // '') ne 'git') {
+        $ctx->info('Git not enabled in config, skipping all clones.');
+        return;
```
Martchus, line 33: I think it would be sufficient to avoid the creation of those tasks. Removing this case would also avoid the failing coverage check.

lib/OpenQA/Utils.pm
```diff
@@ -535,20 +535,25 @@ sub create_downloads_list ($job_settings) {
 
 sub create_git_clone_list ($job_settings, $clones = {}) {
     my $distri = $job_settings->{DISTRI};
-    if (OpenQA::App->singleton->config->{'scm git'}->{git_auto_update} eq 'yes') {
+    my $git_config = OpenQA::App->singleton->config;
```
Martchus, line 538: This variable could just be called `$config`.

Martchus (commented):
Looks good, I just have a few nitpicks.

lib/OpenQA/Task/Git/Clone.pm
AdamWill: ok. at this point I think it should not be possible to reach this anyway, as I think all paths to it are guarded by an earlier check, so it's just a kind of failsafe. I'll adjust it and add a comment.

Martchus (approved):

Martchus:
> set `not-ready` due to the codecov.

There is no need to add this label do to missing coverage as the failing checks will prevent the automatic merge. 

However, the label is still useful as we should keep it until Adam tested the changes in production.

AdamWill:
I've deployed to [Fedora lab](https://openqa.stg.fedoraproject.org/) now and things seem fine, though note all this really tests is "nothing happens if scm = git is not set", because that's what we want. But I don't have tests blowing up or anything like that.

okurz:
This has caused https://progress.opensuse.org/issues/181772

The git auto clone+update operations should still be default so probably we need to set the scm option to enable git by default

okurz:
> I did flag up as loudly as possible that this PR would cause that, so I figured you'd already have `scm = git` in all your ansible/salt/whatever setups...:D

Well, I still did not get that. However regardless of what we have in ansible/salt/whatever the decision must be about users of openQA that don't read pull requests and we shouldn't introduce breaking changes unless we have to. I will merge the revert to prevent the regression to stay and to give us more time to come up with a better solution together

AdamWill:
> the decision must be about users of openQA that don't read pull requests and we shouldn't introduce breaking changes unless we have to

I agree, but that's exactly the concern I have in the first place. As a 'user of openQA' who has an instance where `scm` is very deliberately *not* set to `'git'`, if I had not read the changelog, I would have been *very* surprised at an updated openQA suddenly starting to try and do git pulls all the time. Especially when it turns out that this behaviour is controlled by settings in the `[git scm]` config section.

okurz:
> I agreed merging this change because I thought it would be ok to change the default and our production instances have the setting configured explicitly anyway. The only one on the team strongly advocating for Git cloning/updating by default is @okurz and he had already approved the PR so I thought it was ok.

@Martchus can you explain how openQA should behave relying on git repos without checking them out automatically by default?

PR #6415, opened 2025-05-02
Add linear backoof in hook script

b10n1k:
https://progress.opensuse.org/issues/181427


t/10-jobs.t
```diff
@@ -695,6 +697,46 @@ subtest 'carry over, including soft-fails' => sub {
             like($notes->{hook_result}, qr/delayed/, 'real hook cmd from config called if result matches (5)');
             is $notes->{hook_rc}, 142, 'exit code of the hook cmd is as expected';
             is $job_info->{retries}, 1, 'hook script has been retried once because of delay';
+
+            subtest "Retry hook script with linear backoff..." => sub {
```
okurz, line 701: This high level of indentation makes it hard to read. Please move this subtest out onto the global scope

lib/OpenQA/Task/Job/HookScript.pm
```diff
@@ -1,28 +1,40 @@
...
-    my $kill_timeout = $options->{kill_timeout};
-    my $delay = $options->{delay};
-    my $retries = $options->{retries};
-    my $skip_rc = $options->{skip_rc};
+    my $timeout = $options->{timeout};    # 5
+    my $kill_timeout = $options->{kill_timeout};    # 30
+
+    #my $delay = $options->{delay};    # 1min
+    my $retries = $options->{retries};    # 1440
+    my $delay = ($options->{delay} == 0) ? $options->{delay} : (LINEAR_BACKOFF_BASE_SECONDS * ($job->retries + 1));
```
okurz, line 24: If the delay is 0 you want to use the delay which is 0?

okurz (changes requested):

lib/OpenQA/Task/Job/HookScript.pm
b10n1k: yes. 
`$options->{delay}` seems it comes from `$settings->{_TRIGGER_JOB_DONE_DELAY} // $ENV{OPENQA_JOB_DONE_HOOK_DELAY} // ONE_MINUTE`, so there are cases where the $ENV{OPENQA_JOB_DONE_HOOK_DELAY} is set to 0 to trigger a retry immediately afait from the tests.

lib/OpenQA/Task/Job/HookScript.pm
```diff
@@ -1,28 +1,40 @@
...
 # SPDX-License-Identifier: GPL-2.0-or-later
 
 package OpenQA::Task::Job::HookScript;
 use Mojo::Base 'Mojolicious::Plugin', -signatures;
 use OpenQA::Task::SignalGuard;
+use Data::Dumper;
+$Data::Dumper::Indent = 1;
+$Data::Dumper::Sortkeys = 1;
+
+use constant LINEAR_BACKOFF_BASE_SECONDS => 40;
```
perlpunk, line 11: So currently it's possible to influence the delay via an environment variable, but you want to make it a hardcoded constant now, and the delay passed via the `$options` is only used when it's 0?

My suggestion would be: The delay of one minute is ok for the initial retry, and we can simply use the existing `$options->{delay}`, and then just multiply it with the `$job->retries` as you have already done.

lib/OpenQA/Task/Job/HookScript.pm
```diff
@@ -1,28 +1,40 @@
-# Copyright 2022 SUSE LLC
+# Copyright 2022 SUSE LLC3
```
perlpunk, line 1: ?

b10n1k:
💡 I just managed to make it retry but I cant validate the delay. Needs a better review, so still not usable

t/10-jobs.t
```diff
@@ -699,6 +699,73 @@ subtest 'carry over, including soft-fails' => sub {
...
+    perform_minion_jobs($t->app->minion);
+    my $job_info = {};
+    $job_info = $t->app->minion->jobs({tasks => ['hook_script']})->next;
+    is_deeply($job_info->{args}[2],
+              {delay => 60, retries => 4, skip_rc => 142, kill_timeout => '30s', timeout => '5m'},
+        "Initially delay is set to whatever comes from _run_hook_script");
+    my $expected_linear_delay = $job_info->{args}[2]->{delay};
+    for my $retry_iter (1 .. $max_retries_linear - 1) {
+        my $attempt_num = $retry_iter + 1;
+        my $current_retries_before_run = $retry_iter;
```
Martchus, line 733: You might as well remove this line and use `$retry_iter`.

t/10-jobs.t
```diff
@@ -699,6 +699,73 @@ subtest 'carry over, including soft-fails' => sub {
...
+        note "Linear Backoff: Running attempt $attempt_num (retries=$current_retries_before_run)";
+        perform_minion_jobs($t->app->minion);
+        note "calculate the delay";
+        $expected_linear_delay *= $attempt_num;
+        warn "**** $expected_linear_delay";
```
Martchus, line 738: This should be removed before merging.

lib/OpenQA/Task/Job/HookScript.pm
perlpunk: Currently there is a delay of 1 minute, and it is retrying 1440 times, waiting 1 minute in between.
One hook script run takes around 10 seconds, meaning that 6 hook script jobs all retrying every minute make a whole minion worker slot busy (10 seconds * 6 jobs retrying every minute).

By retrying less often, e.g. increasing the delay linearly, we make the minion server less busy, so it can pick up jobs that would otherwise pile up.

The initial delay of 1 minute is totally fine.

perlpunk: And if you wanted to change that initial delay, you could do so by setting the environment variable or job setting, so I don't know why you want to use a hard coded constant now instead.

lib/OpenQA/Task/Job/FinalizeResults.pm
```diff
@@ -49,7 +49,8 @@ sub _run_hook_script ($minion_job, $openqa_job, $app, $guard) {
     my $timeout = $ENV{OPENQA_JOB_DONE_HOOK_TIMEOUT} // '5m';
     my $kill_timeout = $ENV{OPENQA_JOB_DONE_HOOK_KILL_TIMEOUT} // '30s';
     my $settings = $openqa_job->settings_hash;
-    my $delay = $settings->{_TRIGGER_JOB_DONE_DELAY} // $ENV{OPENQA_JOB_DONE_HOOK_DELAY} // ONE_MINUTE;
+    my $delay = $settings->{_TRIGGER_JOB_DONE_DELAY} // $ENV{OPENQA_JOB_DONE_HOOK_DELAY}
+      // ONE_MINUTE;    # it might make sense to replace the ONE_MINITE
```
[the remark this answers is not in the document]
b10n1k: this and some other are intent to keep track of thoughts as far I have not clear plan (especially as long I keep this as draft) and give thoughts to the reviewer but they are not meant to stay.

perlpunk:
The tests for this can be tricky.
My suggestion for the tests would be (apply that to the master branch):
```diff
diff --git a/t/10-jobs.t b/t/10-jobs.t
index 5d05b9ede..ccd9e2936 100644
--- a/t/10-jobs.t
+++ b/t/10-jobs.t
@@ -685,6 +685,7 @@ subtest 'carry over, including soft-fails' => sub {
             $job->update({state => UPLOADING});
             $job->discard_changes;
             $job->done;
+            my $start = time;
             perform_minion_jobs($t->app->minion);
             $job_info = $t->app->minion->jobs({tasks => ['hook_script']})->next;
             is $job_info->{state}, 'inactive', 'hook script has been retried with long delay';
@@ -695,11 +696,24 @@ subtest 'carry over, including soft-fails' => sub {
             like($notes->{hook_result}, qr/delayed/, 'real hook cmd from config called if result matches (5)');
             is $notes->{hook_rc}, 142, 'exit code of the hook cmd is as expected';
             is $job_info->{retries}, 1, 'hook script has been retried once because of delay';
+            cmp_ok $job_info->{delayed} - $start, '>', 60, 'default delay for first retry';
+
+            # force some retries with delay 0 to avoid having to wait in the test
+            $t->app->minion->job($job_info->{id})->retry({ delay => 0 });
+            $t->app->minion->job($job_info->{id})->retry({ delay => 0 });
+
+            $start = time;
+            perform_minion_jobs($t->app->minion);
+            $job_info = $t->app->minion->jobs({tasks => ['hook_script']})->next;
+
+            is $job_info->{retries}, 4, 'hook script has been retried once because of delay';
+            cmp_ok $job_info->{delayed} - $start, '>', 240, '4 times default delay for 4th retry';
         };
     };
 };
```

lib/OpenQA/Task/Job/HookScript.pm
```diff
@@ -4,25 +4,37 @@
 package OpenQA::Task::Job::HookScript;
 use Mojo::Base 'Mojolicious::Plugin', -signatures;
 use OpenQA::Task::SignalGuard;
+use Data::Dumper;
+$Data::Dumper::Indent = 1;
+$Data::Dumper::Sortkeys = 1;
+
+use constant LINEAR_BACKOFF_BASE_SECONDS => 90;
```
perlpunk, line 11: Please use the delay set in FinalizeResults.
```suggestion
```

lib/OpenQA/Task/Job/HookScript.pm
```diff
@@ -4,25 +4,37 @@
...
-    my $kill_timeout = $options->{kill_timeout};
-    my $delay = $options->{delay};
-    my $retries = $options->{retries};
-    my $skip_rc = $options->{skip_rc};
+    my $timeout = $options->{timeout};    # 5
+    my $kill_timeout = $options->{kill_timeout};    # 30
+
+    #my $delay = $options->{delay};    # 1min
+    my $retries = $options->{retries};    # 1440
+    my $delay = ($options->{delay} == 0) ? $options->{delay} : (LINEAR_BACKOFF_BASE_SECONDS * ($job->retries + 1));
```
perlpunk, line 24:
```suggestion
    my $delay = $options->{delay} * ($job->retries + 1);
```

perlpunk (changes requested):
The default number of retries is still 1440, which would be too much. Please reduce it as I suggested.
Also there should be a test making sure that after the number of retries is exceeded, that the job finishes (without failure).
I think this is currently not tested.

t/10-jobs.t
```diff
@@ -695,10 +696,22 @@ subtest 'carry over, including soft-fails' => sub {
             like($notes->{hook_result}, qr/delayed/, 'real hook cmd from config called if result matches (5)');
             is $notes->{hook_rc}, 142, 'exit code of the hook cmd is as expected';
             is $job_info->{retries}, 1, 'hook script has been retried once because of delay';
+
+            note "There is a linear delay for each hook script which gets exit 142";
+            cmp_ok $job_info->{delayed} - $start, '>', 60, 'default delay for first retry applied correctly';
```
perlpunk, line 701: Maybe it would make sense to also test that the delay is below 120 (and in the below test `> 240` and `< 300`.
And I think we can use the `$job_info->{started}` for the comparison to get a more accurate value, which is less likely to deviate in case of slow test execution.

```diff
diff --git a/t/10-jobs.t b/t/10-jobs.t
index 488c206b2..ac24021dc 100644
--- a/t/10-jobs.t
+++ b/t/10-jobs.t
@@ -685,7 +685,6 @@ subtest 'carry over, including soft-fails' => sub {
             $job->update({state => UPLOADING});
             $job->discard_changes;
             $job->done;
-            my $start = time;
             perform_minion_jobs($t->app->minion);
             $job_info = $t->app->minion->jobs({tasks => ['hook_script']})->next;
             is $job_info->{state}, 'inactive', 'hook script has been retried with long delay';
@@ -698,19 +697,21 @@ subtest 'carry over, including soft-fails' => sub {
             is $job_info->{retries}, 1, 'hook script has been retried once because of delay';
 
             note "There is a linear delay for each hook script which gets exit 142";
-            cmp_ok $job_info->{delayed} - $start, '>', 60, 'default delay for first retry applied correctly';
+            my $delay = $job_info->{delayed} - $job_info->{started};
+            ok $delay > 60 && $delay < 120, 'default delay for first retry applied correctly';
             # force some retries with delay 0 to avoid having to wait in the test
             $t->app->minion->job($job_info->{id})->retry({delay => 0});
             $t->app->minion->job($job_info->{id})->retry({delay => 0});
-            $start = time;
             perform_minion_jobs($t->app->minion);
             $job_info = $t->app->minion->jobs({tasks => ['hook_script']})->next;
             is $job_info->{retries}, 4, 'hook script has been retried 4 times';
-            cmp_ok $job_info->{delayed} - $start, '>', 240, 'linear delay calculated correctly for 4th retry';
+            $delay = $job_info->{delayed} - $job_info->{started};
+            ok $delay > 240 && $delay < 300, 'linear delay calculated correctly for 4th retry';
         };
     };
 };
```

t/10-jobs.t
```diff
@@ -695,10 +696,22 @@ subtest 'carry over, including soft-fails' => sub {
...
+            $start = time;
+            perform_minion_jobs($t->app->minion);
+            $job_info = $t->app->minion->jobs({tasks => ['hook_script']})->next;
+            is $job_info->{retries}, 4, 'hook script has been retried 4 times';
+            cmp_ok $job_info->{delayed} - $start, '>', 240, 'linear delay calculated correctly for 4th retry';
         };
     };
 };
 
+
```
perlpunk, line 714:
```suggestion
```

t/10-jobs.t
perlpunk: Or actually, using the `retried` field gives us the exact delay:
```diff
diff --git a/t/10-jobs.t b/t/10-jobs.t
index 488c206b2..b385e2326 100644
--- a/t/10-jobs.t
+++ b/t/10-jobs.t
@@ -685,7 +685,6 @@ subtest 'carry over, including soft-fails' => sub {
             $job->update({state => UPLOADING});
             $job->discard_changes;
             $job->done;
-            my $start = time;
             perform_minion_jobs($t->app->minion);
             $job_info = $t->app->minion->jobs({tasks => ['hook_script']})->next;
             is $job_info->{state}, 'inactive', 'hook script has been retried with long delay';
@@ -698,19 +697,21 @@ subtest 'carry over, including soft-fails' => sub {
             is $job_info->{retries}, 1, 'hook script has been retried once because of delay';
 
             note "There is a linear delay for each hook script which gets exit 142";
-            cmp_ok $job_info->{delayed} - $start, '>', 60, 'default delay for first retry applied correctly';
+            my $delay = $job_info->{delayed} - $job_info->{retried};
+            is $delay, 60 , 'default delay for first retry applied correctly';
             # force some retries with delay 0 to avoid having to wait in the test
             $t->app->minion->job($job_info->{id})->retry({delay => 0});
             $t->app->minion->job($job_info->{id})->retry({delay => 0});
-            $start = time;
             perform_minion_jobs($t->app->minion);
             $job_info = $t->app->minion->jobs({tasks => ['hook_script']})->next;
             is $job_info->{retries}, 4, 'hook script has been retried 4 times';
-            cmp_ok $job_info->{delayed} - $start, '>', 240, 'linear delay calculated correctly for 4th retry';
+            $delay = $job_info->{delayed} - $job_info->{retried};
+            is $delay, 240, 'linear delay calculated correctly for 4th retry';
         };
     };
 };
```

b10n1k:
> The default number of retries is still 1440, which would be too much. Please reduce it as I suggested. Also there should be a test making sure that after the number of retries is exceeded, that the job finishes (without failure). I think this is currently not tested.

Actually i did that manual. the state comes to {finished}. I didnt check but I would expect an surprises from this change on how it behaves at the end of retries. Anyway lets reduce the retries.

t/10-jobs.t
b10n1k: when I tested the test the delay was not accurate. `$job_info->{delayed} - $job_info->{retried}` looks much better approach. thanks

lib/OpenQA/Task/Job/HookScript.pm
```diff
@@ -13,16 +13,17 @@ sub _hook_script ($job, $hook, $openqa_job_id, $options) {
...
     my $skip_rc = $options->{skip_rc};
 
     $ensure_task_retry_on_termination_signal_guard->abort(1);
     my ($rc, $out) = _run_hook($hook, $openqa_job_id, $timeout, $kill_timeout);
     $job->note(hook_cmd => $hook, hook_result => $out, hook_rc => $rc);
 
-    if ($retries && $skip_rc) {
-        $job->retry($delay ? {delay => $delay} : {}) if defined $rc && $rc == $skip_rc && $job->retries < $retries;
+    if ($retries && $skip_rc && defined $rc && $rc == $skip_rc && $job->retries < $retries) {
+        $job->app->log->debug("Job $job->id: Retrying $job->retries with linear delay: $delay");
```
okurz, line 25: I assume that's seconds so

```suggestion
        $job->app->log->debug("Job $job->id: Retrying $job->retries with linear delay: ${delay}s");
```

Also can you give an example of a log line from actual execution? I don't think any of the three used variables are ensured to be defined

t/10-jobs.t
```diff
@@ -689,12 +688,29 @@ subtest 'carry over, including soft-fails' => sub {
...
             like($notes->{hook_result}, qr/delayed/, 'real hook cmd from config called if result matches (5)');
             is $notes->{hook_rc}, 142, 'exit code of the hook cmd is as expected';
             is $job_info->{retries}, 1, 'hook script has been retried once because of delay';
+
+            note "There is a linear delay for each hook script which gets exit 142";
+            my $linear_delay = $job_info->{delayed} - $job_info->{retried};
+            is $linear_delay, 60, 'default delay for first retry applied correctly';
+            # force some retries with delay 0 to avoid having to wait in the test
+            $t->app->minion->job($job_info->{id})->retry({delay => 0});
+            $t->app->minion->job($job_info->{id})->retry({delay => 0});
```
okurz, line 703: In os-autoinst we use Test::Mock::Time so try if that helps here as well

t/10-jobs.t
```diff
@@ -689,12 +688,29 @@ subtest 'carry over, including soft-fails' => sub {
...
+            perform_minion_jobs($t->app->minion);
+            $job_info = $t->app->minion->jobs({tasks => ['hook_script']})->next;
+            is $job_info->{retries}, 4, 'hook script has been retried 4 times';
```
okurz, line 706: the test label stating an explicit number of retries tends to be outdated sooner or later. Better just state

```suggestion
            is $job_info->{retries}, 4, 'hook script has been retried the expected times';
```

t/10-jobs.t
```diff
@@ -679,7 +679,6 @@ subtest 'carry over, including soft-fails' => sub {
 
             # Defaults (retry with delay)
             $job->settings->search({key => '_TRIGGER_JOB_DONE_DELAY'})->delete;
-            $job->settings->search({key => '_TRIGGER_JOB_DONE_RETRIES'})->delete;
```
okurz, line 682: why?

lib/OpenQA/Task/Job/HookScript.pm
b10n1k: I tried but gru doesnt work properly local. However all of them are expected to be available, and if I remember correctly I could print them and find them

t/10-jobs.t
okurz: https://github.com/os-autoinst/openQA/pull/6415#pullrequestreview-2830835533 states that you should reduce the default which this PR so far does not do. And it states that you should add a test to test the exhaustion of retries. Instead what you do in the test that states that it would test the defaults is now actually not testing defaults anymore

b10n1k: Correct. I oversaw that this was checking the default, so I thought it is simplier. I extracted to a subtest.

t/10-jobs.t
Martchus: And in any case, this line seems to be duplicated.

b10n1k:
💡Address  https://github.com/os-autoinst/openQA/pull/6415#discussion_r2084421775 and removed the test in another subtest

t/10-jobs.t
```diff
@@ -696,6 +696,40 @@ subtest 'carry over, including soft-fails' => sub {
...
+            perform_minion_jobs($t->app->minion);
+            $job_info = $t->app->minion->jobs({tasks => ['hook_script']})->next;
+            is_deeply(
+                $job_info->{args}[2],
+                {delay => 60, retries => 4, skip_rc => 142, kill_timeout => '30s', timeout => '5m'},
+                'Confirm that args key contains expected parameters'
+            );
+            is $job_info->{retries}, 1, 'hook script has been retried once because of delay';
+            is $job_info->{state}, 'inactive', 'and it ends up inactive due to retry';
+            note "we expect (4 retries * the default one minute delay";
```
okurz, line 715: What happened with the `(4` here?

b10n1k: yea sorry.

t/10-jobs.t
kalikiana: This is still missing?

Actually I think the suggested change is missing a couple words.

```suggestion
            is $job_info->{retries}, 4, 'hook script has been retried the expected number of times';
```

kalikiana: @b10n1k Please be sure to check open suggestions

t/10-jobs.t
kalikiana: @b10n1k Please remove the duplicate line

Martchus: Or use a loop to avoid the duplication.

kalikiana:
```
Can't call method "get_attribute" on an undefined value at t/ui/15-comments.t line 328.
```

b10n1k:
> ```
> Can't call method "get_attribute" on an undefined value at t/ui/15-comments.t line 328.
> ```

yes but this seems had nothing likely with my changes. they passed the last few times. yesterday I have the impression I rebase and fix a typo. So no sure if I dont take a deeper look where this error comes from

kalikiana:
> >     Can't call method "get_attribute" on an undefined value at t/ui/15-comments.t line 328.
> 
> yes but this seems had nothing likely with my changes. they passed the last few times. yesterday I have the impression I rebase and fix a typo. So no sure if I dont take a deeper look where this error comes from

I was thinking that, too. Tho I can't find this in other PR's 🤔 Lemme try and reproduce with a trivial PR

okurz (approved):

kalikiana (approved):

perlpunk:
To verify "AC3: The minion job is still eventually giving up after a reasonable time", can you give us a calculation of the total wait time?

b10n1k:
> To verify "AC3: The minion job is still eventually giving up after a reasonable time", can you give us a calculation of the total wait time?

I think @perlpunk already answers this in https://github.com/os-autoinst/openQA/pull/6415#discussion_r2075868523

lib/OpenQA/Task/Job/HookScript.pm
```diff
@@ -13,16 +13,17 @@ sub _hook_script ($job, $hook, $openqa_job_id, $options) {
...
     my $skip_rc = $options->{skip_rc};
 
     $ensure_task_retry_on_termination_signal_guard->abort(1);
     my ($rc, $out) = _run_hook($hook, $openqa_job_id, $timeout, $kill_timeout);
     $job->note(hook_cmd => $hook, hook_result => $out, hook_rc => $rc);
 
-    if ($retries && $skip_rc) {
-        $job->retry($delay ? {delay => $delay} : {}) if defined $rc && $rc == $skip_rc && $job->retries < $retries;
+    if ($retries && $skip_rc && defined $rc && $rc == $skip_rc && $job->retries < $retries) {
+        $job->app->log->debug("Job $job->id: Retrying $job->retries with linear delay: ${delay}s");
```
Martchus, line 25: I don't think you can make a function call like `$job->retries` in double-quotes expecting interpolation to happen. Have you even tested whether this is logged correctly?

Martchus (changes requested):

t/10-jobs.t
```diff
@@ -696,6 +696,39 @@ subtest 'carry over, including soft-fails' => sub {
...
+            perform_minion_jobs($t->app->minion);
+            $job_info = $t->app->minion->jobs({tasks => ['hook_script']})->next;
+            is_deeply(
+                $job_info->{args}[2],
+                {delay => 60, retries => 4, skip_rc => 142, kill_timeout => '30s', timeout => '5m'},
+                'Confirm that args key contains expected parameters'
+            );
+            is $job_info->{retries}, 1, 'hook script has been retried once because of delay';
+            is $job_info->{state}, 'inactive', 'and it ends up inactive due to retry';
+            note "we expect 4 retries * the default one minute delay";
```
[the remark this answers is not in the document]
b10n1k: I am planning to use `note` more in the future to drive some logic and to not getting lost between changes from one test case to another when the assertion provides simply some description but is not exactly clear what or why is "OK". Hope is that it will help others too when reviews tests or want to debug. So i am planning to keep it, or I could think to rephrase it later, when I have even better understanding

lib/OpenQA/Task/Job/HookScript.pm
b10n1k: I think @okurz asked me the same. I couldnt reproduce it local (problems with the gru). I only remember that I used print statements during implementation. I need to check again and back back to you

t/10-jobs.t
```diff
@@ -696,6 +696,39 @@ subtest 'carry over, including soft-fails' => sub {
...
+            $job->discard_changes;
+            $job->done;
+            perform_minion_jobs($t->app->minion);
+            $job_info = $t->app->minion->jobs({tasks => ['hook_script']})->next;
+            is_deeply(
+                $job_info->{args}[2],
+                {delay => 60, retries => 4, skip_rc => 142, kill_timeout => '30s', timeout => '5m'},
+                'Confirm that args key contains expected parameters'
+            );
+            is $job_info->{retries}, 1, 'hook script has been retried once because of delay';
```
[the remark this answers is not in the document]
Martchus: Since *it* has been queued?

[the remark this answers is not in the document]
Martchus: The word order is also wrong. And the correct tense for the last part would probably be past simple:

```
hook script has already been tried once since it was queued
```

("Since" is a signal word for the present perfect but *only* for the part that happened since then.)

lib/OpenQA/Task/Job/HookScript.pm
Martchus: You definitely need to test this *somehow*. Shouldn't the unit tests run this line of code? Maybe you can temporarily raise the log level (if needed) and see what's printed when executing the tests.

perlpunk:
> Simple answer: it depends from the value of the retries.

Exactly. But there is a default, and that is 1440.

## If this PR is merged and deployed, how long will hook_script jobs wait in total?

It's easy to calculate something like `1 + 2 + 3 + 4 + ... + x`:
It's `(1 + x) * x / 2`.

e.g. `1+2+3+4+5+6+7+8+9+10 == (10+1) * 10 / 2 == 110 / 2 == 55`

 We can even calculate our case in our head, if we are not interested in a very accurate number:

So, x is 1440 in our case:
```
(1440 + 1) * 1440 / 2
# 1440 stands for 1440 minutes, so 24 hours or 1 day.
# Let's forget about the extra 1 and say 1440+1 is roughly 1 day:
~> 1d * 1440 / 2
-> 1d * 720
-> 720 days
~> 2 years
```
(My calculator says `(1440 + 1) * 1440 / 2 = 1037520minutes == 17292hours == 720.5days`)

### Answer: 2 years.

I think we can agree that this is not a reasonable time to wait :)

So please reduce the default number of retries, as suggested in https://github.com/os-autoinst/openQA/pull/6415#pullrequestreview-2830835533


> googles says 53 retries for 60 sec delay 60+ 53 = 86220

That seems wrong. Don't use google for calculating (without double checking)  :)

lib/OpenQA/Task/Job/HookScript.pm
b10n1k: OK. I couldnt get something useful from my openqa instance. 
So the interpolation in double quotes can work as I found in https://perldoc.perl.org/perlref#Miscellaneous-Usage. i test this. 
or I can assign this to a variable to stay save. what do you say?

perlpunk: > OK. I couldnt get something useful from my openqa instance.

What do you mean? Did you run the test with `-v` as suggested?
```
% make test EXTRA_PROVE_ARGS="-v" TESTS=t/10-jobs.t
[debug] [pid:21982] Job 100015 duplicated as 100016
[debug] Job OpenQA::Shared::GruJob=HASH(0x55ca17808c90)->id: Retrying OpenQA::Shared::GruJob=HASH(0x55ca17808c90)->retries with linear delay: 60s
    ok 15 - clone exists for retry nr. 2
[debug] [pid:21982] _carry_over_candidate(100016): _failure_reason=GOOD
[debug] Job OpenQA::Shared::GruJob=HASH(0x55ca178de770)->id: Retrying OpenQA::Shared::GruJob=HASH(0x55ca178de770)->retries with linear delay: 60s
```
You can interpolate scalar variables, hash or array elements like this:
```
say "$number, $hash->{key}, $array->[23]";
```
but you cannot interpolate method calls. As you can see, the output is not helpful.

b10n1k:
> > Simple answer: it depends from the value of the retries.
> 
> Exactly. But there is a default, and that is 1440.
> 
> ## If this PR is merged and deployed, how long will hook_script jobs wait in total?
> It's easy to calculate something like `1 + 2 + 3 + 4 + ... + x`: It's `(1 + x) * x / 2`.
> 
> e.g. `1+2+3+4+5+6+7+8+9+10 == (10+1) * 10 / 2 == 110 / 2 == 55`
> 
> We can even calculate our case in our head, if we are not interested in a very accurate number:
> 
> So, x is 1440 in our case:
> 
> ```
> (1440 + 1) * 1440 / 2
> # 1440 stands for 1440 minutes, so 24 hours or 1 day.
> # Let's forget about the extra 1 and say 1440+1 is roughly 1 day:
> ~> 1d * 1440 / 2
> -> 1d * 720
> -> 720 days
> ~> 2 years
> ```
> 
> (My calculator says `(1440 + 1) * 1440 / 2 = 1037520minutes == 17292hours == 720.5days`)
> 
> ### Answer: 2 years.
> I think we can agree that this is not a reasonable time to wait :)
> 
> So please reduce the default number of retries, as suggested in [#6415 (review)](https://github.com/os-autoinst/openQA/pull/6415#pullrequestreview-2830835533)
> 
> > googles says 53 retries for 60 sec delay 60+ 53 = 86220
> 
> That seems wrong. Don't use google for calculating (without double checking) :)

But can I just reduce the retries? I am afraid that it is not that easy and i would like to ask for some clarifications before change anything further. so lets take a look tomorrow in the collab.

perlpunk:
> But can I just reduce the retries? I am afraid that it is not that easy and i would like to ask for some clarifications before change anything further. so lets take a look tomorrow in the collab.

I would like to see this finished soon, so:

```diff
diff --git a/lib/OpenQA/Task/Job/FinalizeResults.pm b/lib/OpenQA/Task/Job/FinalizeResults.pm
index c0895ff83..ca0cf34c4 100644
--- a/lib/OpenQA/Task/Job/FinalizeResults.pm
+++ b/lib/OpenQA/Task/Job/FinalizeResults.pm
@@ -50,7 +50,8 @@ sub _run_hook_script ($minion_job, $openqa_job, $app, $guard) {
     my $kill_timeout = $ENV{OPENQA_JOB_DONE_HOOK_KILL_TIMEOUT} // '30s';
     my $settings = $openqa_job->settings_hash;
     my $delay = $settings->{_TRIGGER_JOB_DONE_DELAY} // $ENV{OPENQA_JOB_DONE_HOOK_DELAY} // ONE_MINUTE;
-    my $retries = $settings->{_TRIGGER_JOB_DONE_RETRIES} // $ENV{OPENQA_JOB_DONE_HOOK_RETRIES} // 1440;
+    # Linear backoff with 53 retries ~ 1 day
+    my $retries = $settings->{_TRIGGER_JOB_DONE_RETRIES} // $ENV{OPENQA_JOB_DONE_HOOK_RETRIES} // 53;
     my $skip_rc = $settings->{_TRIGGER_JOB_DONE_SKIP_RC} // $ENV{OPENQA_JOB_DONE_HOOK_SKIP_RC} // 142;
     $guard->retry(0);
     my $id = $app->minion->enqueue(
```
Leftover questions can be clarified tomorrow.

lib/OpenQA/Task/Job/HookScript.pm
perlpunk: > or I can assign this to a variable to stay save. what do you say?

The preferred way is to use [`sprintf`](https://perldoc.perl.org/functions/sprintf) to print variables in a string, if interpolation does not work or it would look to messy.

t/10-jobs.t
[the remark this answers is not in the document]
okurz: I suggest to be very careful adding more "note". Sooner or later any such comments can outdate and be more confusing than helpful. Here also I am with Marius and suggest to remove it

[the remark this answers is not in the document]
b10n1k: thats also true. I really want only to provide some _help_ here for the future but as two of you requested it, I will remove

lib/OpenQA/Task/Job/HookScript.pm
b10n1k: ok. something new to learn. and thanks @perlpunk for hint
 
So you all are right. but `$job->app->log->debug("Job @{[$job->id]}: Retrying @{[$job->retries]} with linear delay: ${delay}s");`
seems ok 
``` ] Job 15: Retrying 0 with linear delay: 60s
            ok 1 - args contain expected parameters
            ok 2 - hook script has been already tried once since has been queued
            ok 3 - job ends up inactive due to retry
            ok 4 - default delay for first retry applied correctly
[debug] Job 15: Retrying 3 with linear delay: 240s
```

So.. what do you want me to do? remove the debug message or change the string or...?

b10n1k:
💡change the default number of retries in lib/OpenQA/Task/Job/FinalizeResults.pm

okurz (approved):

t/10-jobs.t
```diff
@@ -689,13 +690,45 @@ subtest 'carry over, including soft-fails' => sub {
...
-                {delay => 60, retries => 1440, skip_rc => 142, kill_timeout => '30s', timeout => '5m'});
+                {delay => 60, retries => $default_retries, skip_rc => 142, kill_timeout => '30s', timeout => '5m'});
             $notes = $job_info->{notes};
             is($notes->{hook_cmd}, 'echo delayed && exit 142;', 'real hook cmd in notes if result matches (5)');
             like($notes->{hook_result}, qr/delayed/, 'real hook cmd from config called if result matches (5)');
             is $notes->{hook_rc}, 142, 'exit code of the hook cmd is as expected';
             is $job_info->{retries}, 1, 'hook script has been retried once because of delay';
         };
+
+        subtest 'There is a linear delay for each hook script which exits with 142' => sub {
```
perlpunk, line 701: Not requesting a change, but for the future: whole sentences for subtests titles are usually too much IMHO.

t/10-jobs.t
```diff
@@ -689,13 +690,45 @@ subtest 'carry over, including soft-fails' => sub {
...
+            my $linear_delay = $job_info->{delayed} - $job_info->{retried};
+            is $linear_delay, 60, 'default delay for first retry applied correctly';
+            # force some retries with delay 0 to avoid having to wait in the test
+            $t->app->minion->job($job_info->{id})->retry({delay => 0}) for 1, 2;
+            perform_minion_jobs($t->app->minion);
+            $job_info = $t->app->minion->jobs({tasks => ['hook_script']})->next;
+            is $job_info->{retries}, 4, 'hook script has been retried the expected number of times';
+            $linear_delay = $job_info->{delayed} - $job_info->{retried};
+            is $linear_delay, 240, 'linear delay calculated correctly for 4th retry';
+            note "Retry one last time to validate the final state";
```
perlpunk, line 725: I also wouldn't use `note` here, just a comment

perlpunk (approved):

t/10-jobs.t
b10n1k: I will remove but I would like to just ask everyone why is this a problem when those are acting as comments? How can i achieve what I intend to?

okurz: What does "those are acting as comments" mean?

lib/OpenQA/Task/Job/HookScript.pm
```diff
@@ -13,16 +13,17 @@ sub _hook_script ($job, $hook, $openqa_job_id, $options) {
...
     my $skip_rc = $options->{skip_rc};
 
     $ensure_task_retry_on_termination_signal_guard->abort(1);
     my ($rc, $out) = _run_hook($hook, $openqa_job_id, $timeout, $kill_timeout);
     $job->note(hook_cmd => $hook, hook_result => $out, hook_rc => $rc);
 
-    if ($retries && $skip_rc) {
-        $job->retry($delay ? {delay => $delay} : {}) if defined $rc && $rc == $skip_rc && $job->retries < $retries;
+    if ($retries && $skip_rc && defined $rc && $rc == $skip_rc && $job->retries < $retries) {
+        $job->app->log->debug("Job @{[$job->id]}: Retrying @{[$job->retries]} with linear delay: ${delay}s");
```
Martchus, line 25: This is an interesting hack to make interpolation work and I trust you tested that it actually works. So I have only have one nitpick:

```suggestion
        $job->app->log->debug("Job @{[$job->id]}: Retrying @{[$job->retries]} times with linear delay: ${delay}s");
```

t/10-jobs.t
```diff
@@ -696,6 +696,39 @@ subtest 'carry over, including soft-fails' => sub {
...
             is $job_info->{retries}, 1, 'hook script has been retried once because of delay';
         };
+
+        subtest 'There is a linear delay for each hook script which gets exit 142' => sub {
+            $job->settings->create({key => '_TRIGGER_JOB_DONE_RETRIES', value => '4'});
+            $hooks->{job_done_hook_failed} = 'echo delayed && exit 142;';
+            $job->update({state => UPLOADING});
+            $job->discard_changes;
+            $job->done;
+            perform_minion_jobs($t->app->minion);
```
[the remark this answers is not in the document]
Martchus: Doing it only in this subtest would be good enough. If you like you can do a search & replace in the whole test and provide this change as a separate commit. The last option would be best.

t/10-jobs.t
```diff
@@ -689,13 +690,44 @@ subtest 'carry over, including soft-fails' => sub {
...
-                {delay => 60, retries => 1440, skip_rc => 142, kill_timeout => '30s', timeout => '5m'});
+                {delay => 60, retries => $default_retries, skip_rc => 142, kill_timeout => '30s', timeout => '5m'});
             $notes = $job_info->{notes};
             is($notes->{hook_cmd}, 'echo delayed && exit 142;', 'real hook cmd in notes if result matches (5)');
             like($notes->{hook_result}, qr/delayed/, 'real hook cmd from config called if result matches (5)');
             is $notes->{hook_rc}, 142, 'exit code of the hook cmd is as expected';
             is $job_info->{retries}, 1, 'hook script has been retried once because of delay';
         };
+
+        subtest 'There is a linear delay for each hook script which exits with 142' => sub {
```
Martchus, line 701: I agree with @perlpunk. To make this shorter you can e.g. remove the "There is a" part which is not adding anything useful and rephrase it a little bit:
```suggestion
        subtest 'linear delay of each hook scripts exiting with 142' => sub {
```

Martchus (changes requested):

t/10-jobs.t
b10n1k: there are not visible, unless test run with verbose flag

b10n1k:
💡remove `note` in test
💡use variable in place of $t->app->minion
💡change subtest description and some other wordings

t/10-jobs.t
```diff
@@ -625,14 +625,15 @@ subtest 'carry over, including soft-fails' => sub {
 
         subtest 'Retry hook script with exit code 142' => sub {
             # Defaults (no retry)
+            my $default_retries = 53;
```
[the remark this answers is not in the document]
perlpunk: No, it's not duplicating. This test simply checks if we get the default value in line 636.
If you change the default in FinalizeResults, the test will fail, so all good

okurz (approved):

Martchus (approved):

PR #6419, opened 2025-05-02
Makefile: Create config folder template only where to be used by users

okurz:
Only etc/openqa is intended to be used by users. If distributors want
to put things into usr/etc then the according directories should only be
created in the according places but not during generic installation to
prevent problems like distribution specific requirements for such
folders.

This fixes a problem with "ostree" as reported by AdamWill.

Related to
https://progress.opensuse.org/issues/179359#note-33

### DIFF
--- a/dist/rpm/openQA.spec
+++ b/dist/rpm/openQA.spec
@@ -15,11 +15,6 @@
 # Please submit bugfixes or comments via http://bugs.opensuse.org/
 #
 
-# define _distconfdir for openSUSE Leap 15 and SLE 15 (like other packages such as libvpl do)
-%if 0%{?suse_version} < 1550
-%define _distconfdir %{_prefix}%{_sysconfdir}
-%endif
-
 # can't use linebreaks here!
 %define openqa_main_service openqa-webui.service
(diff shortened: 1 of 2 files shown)


AdamWill:
I'd say maybe "local administrators" rather than "users", but yeah. https://en.opensuse.org/openSUSE:Packaging_UsrEtc does say "/usr/etc/example.conf or /usr/etc/example.conf.d always exist and will be provided by the distributor", but I think the implication is that they're expected to "exist" *to provide the default configuration*. If the default configuration is kept in code (as it is in openQA), providing just `/usr/etc/example.conf.d` as an empty directory is kind of meaningless.

AdamWill:
I guess as an alternative we could *stop* keeping the default config in code and instead install it as a `/usr`-level config file, but if we do that, it'd really be nice if we could put it in `/usr/lib` (as systemd does) or `/usr/share` (as various other things do).

https://github.com/uapi-group/specifications/issues/76#issuecomment-2379123095 from a SUSE person says "We follow
https://uapi-group.org/specifications/specs/configuration_files_specification/ if the format of the configuration file allows it. If not, we only provide a fallback config if there is none in /etc. In which case admins have to manually merge changes. /usr/etc is only used for configuration files where no upstream really exists, like "services" or" protocols" or so."

which seems to suggest the SUSE policy isn't really meant to imply that *upstreams* should put distributor-level config in `/usr/etc`. So I *think* it'd be fine to have it in `/usr/lib` or `/usr/share` and the SUSE package could keep it there too?

AdamWill:
and yeah, this does at least require changes to the RPM spec in the tree, it needs to be changed to not include these files.

dist/rpm/openQA.spec
```diff
@@ -583,9 +583,6 @@ fi
 %dir %{_sysconfdir}/openqa
 %dir %{_sysconfdir}/openqa/openqa.ini.d
 %dir %{_sysconfdir}/openqa/database.ini.d
-%dir %{_distconfdir}/openqa
```
Martchus, line 586: The worker and client config need a similar change.

okurz: done

Martchus (approved):

Wabri (approved):

PR #6420, opened 2025-05-02
Improve regex to match flexible spacing

realcharmer:
https://progress.opensuse.org/issues/180002

The current regex expect exactly two whitespaces for spacing. This change makes the regex tolerant to any separation.

### DIFF
--- a/script/configure-web-proxy
+++ b/script/configure-web-proxy
@@ -46,7 +46,7 @@ if [[ $web_proxy == "nginx" ]]; then
     if [[ -n "$web_port" ]]; then
         sed -Ei "s@(.*listen.*)80@\1${web_port}@g" /etc/nginx/vhosts.d/openqa.conf /etc/nginx/nginx.conf
     fi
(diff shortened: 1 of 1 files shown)


okurz (approved):

perlpunk (commented):
just a nitpick, but the cureent regex expects two spaces, not whitespaces.
and the term whitespace is actually already a kind of plural, meaning multiple occurrences of spaces, tabs, newlines, etc.

Martchus (approved):

realcharmer:
> just a nitpick, but the cureent regex expects two spaces, not whitespaces. and the term whitespace is actually already a kind of plural, meaning multiple occurrences of spaces, tabs, newlines, etc.

I see. Though `[[:space:]]` includes tabs, spaces and so on, so it should be ready for possible future changes in the config.

perlpunk:
> The current regex expect exactly two whitespaces

this is the sentence from the pr description.

PR #6421, opened 2025-05-05
Ensure git features are again enabled by default

okurz:
Related progress issue: https://progress.opensuse.org/issues/181772


okurz:
@Martchus does this go in the right direction regarding reverting the decision of

* #6414

while still allowing to completely disable git support?

@AdamWill please review as well

Martchus:
Keeping possibly disturbing features disabled by default (and thus not merge this PR and enable Git support manually in openQA-in-openQA tests) would be the conservative approach. This would also be the better option for local development.

Enabling those Git features means a better out-of-the-box experience for some use cases (we probably consider more important than others and also test in openQA-in-openQA tests).

I don't care whether we enable/disable Git by default. If it was my personal project I'd probably pick the conservative approach unless is is known for a fact that the majority of users would actually enable those Git features.

lib/OpenQA/Setup.pm
```diff
@@ -68,7 +68,7 @@ sub read_config ($app) {
             branding => 'openSUSE',
             download_domains => undef,
             suse_mirror => undef,
-            scm => undef,
+            scm => 'git',
```
Martchus, line 71: `t/config.t` needed to be adjusted as well.

okurz:
> Keeping possibly disturbing features disabled by default (and thus not merge this PR and enable Git support manually in openQA-in-openQA tests) would be the conservative approach. This would also be the better option for local development.
> 
> Enabling those Git features means a better out-of-the-box experience for some use cases (we probably consider more important than others and also test in openQA-in-openQA tests).
> 
> I don't care whether we enable/disable Git by default. If it was my personal project I'd probably pick the conservative approach unless is is known for a fact that the majority of users would actually enable those Git features.

Yes, git support in general must be enabled by default, see https://progress.opensuse.org/issues/168379 . But we might need to reconsider the defaults of individual sub-features

okurz:
Fixed Perl tidy style and extended git commit message details

Martchus:
> Yes, git support in general must be enabled by default, see https://progress.opensuse.org/issues/168379 .

I didn't say it must be enabled by default - only the ticket implies that.

Maybe we should disable auto commit and auto push by default to retain the previous behavior. On the other hand, those features are probably the one's that get the least in the way and might even be wanted when doing local development.

AdamWill:
Well, I'm kinda with @Martchus . I understand that you need git behaviour enabled on all the official SUSE deployments of openQA (or whatever exactly the situation is there), but I don't see why that means it needs to be the default *as openQA is shipped as an upstream project*. Imagining myself as an entire outsider to openQA, I think I would assume features like this would be disabled by default.

Is it really necessary to enable these features by default upstream in order to resolve the situation with SUSE deployments? Can you not, for instance, just make sure `scm = git` is in all your ansible/salt/whatever configs? Or add a config snippet with `scm = git` to a package that is reliably deployed on SUSE openQA deployments (but not necessarily on non-SUSE-infra deployments?)

okurz:
> Well, I'm kinda with @Martchus . I understand that you need git behaviour enabled on all the official SUSE deployments of openQA (or whatever exactly the situation is there), but I don't see why that means it needs to be the default _as openQA is shipped as an upstream project_. Imagining myself as an entire outsider to openQA, I think I would assume features like this would be disabled by default.

I think we have a misunderstanding regarding the intentions on what we envision for newcomers to openQA: The idea was that anyone deploying openQA without any configuration can just trigger tests pointing to a git repository including test definitions and then the repo and test definitions within are automatically checked out and also kept up to date. For that we need auto-clone+update.

> Is it really necessary to enable these features by default upstream in order to resolve the situation with SUSE deployments? Can you not, for instance, just make sure `scm = git` is in all your ansible/salt/whatever configs? Or add a config snippet with `scm = git` to a package that is reliably deployed on SUSE openQA deployments (but not necessarily on non-SUSE-infra deployments?)

No, this is not about any SUSE deployment. This is purely an "upstream" decision. It would not be a problem of enabling that where necessary for the relevant deployments. As explained and decided *and implemented* in https://progress.opensuse.org/issues/168379 we want auto-clone+update enabled by default. But it seems now that auto-submit+push are also enabled which might be too far. I see two possibilties so that we don't keep this regression open for too long: Merge this PR or revert https://github.com/os-autoinst/openQA/pull/6414 and then reconsider.

Martchus:
> Maybe we should disable auto commit and auto push by default to retain the previous behavior. On the other hand, those features are probably the ones that get the least in the way and might even be wanted when doing local development.

Thinking about it further I'm not sure whether automatically pushing (and maybe not even committing) is useful for local development.

> Is it really necessary to enable these features by default upstream in order to resolve the situation with SUSE deployments?

This is already enabled in our deployments. The problem is just that openQA-in-openQA tests (which are testing openQA with itself) currently expect that Git features are enabled. Normally we prefer testing the default config in that test but we also want to test the Git features there. So there's a bit of an conflict. However, this descicion really shouldn't be about the openQA-in-openQA test but what we think is the best default.

> The idea was that anyone deploying openQA without any configuration can just trigger tests pointing to a git repository including test definitions and then the repo and test definitions within are automatically checked out and also kept up to date. For that we need auto-clone+update.

Yes, we need it for *that*. However, it is not a necassity but a desicion to consider this use case more important than other use cases.

AdamWill:
ah, I see. well, yeah, it's definitely a judgment call in that case. I can adapt our downstream config easily enough to whatever is decided.

okurz:
I added the new additionally needed change to disable auto_commit by default to resemble what we had before this PR and before #6414 if I understood it correctly. I reworked the according documentation section now thoroughly as by default we really only keep auto-clone+update but not auto-commit+push.

@Martchus with this change I also reverted our now not anymore necessary changes to t/ui/21-admin-needles.t . I hope you both see this as acceptable now.

AdamWill:
another way to do it would be to get rid of the whole 'scm git' concept entirely, or deprecate it. have the enabling options at the top level of the config - so we have `git_auto_*` in the `general` section. treat `scm = git` as a deprecated synonym for `git_auto_commit = yes`. have the detail configuration options either in general or in separate sections per topic area, and treat the `scm git` section as a deprecated synonym for whatever the new name for the settings related to `git_auto_commit` are, with special handling for `git_auto_push = no` and `git_auto_update = no` in that section treated as a backwards-compatibility feature...

so something like:

```
[general]
git_auto_commit = yes|no
git_auto_clone = yes|no
git_auto_update = yes|no
scm = git | * ['git' treated as deprecated synonym for `git_auto_commit = yes`]

[git_auto_commit_options]
update_remote = origin
update_branch = origin/master
do_cleanup = no
do_push = no

[git_auto_clone_options]

[git_auto_update_options]
checkout_needles_sha = no
allow_arbitrary_url_fetch = no
temp_needle_refs_retention = 120

[scm git]
git_auto_push = yes|no [treated as deprecated synonym for general.git_auto_push]
git_auto_update = yes|no [treated as deprecated synonym for general.git_auto_update]
[all other settings treated as deprecated synonyms for the setting in the new section]

then we could default `general.git_auto_push` and `general.git_auto_update` to yes, but `general.git_auto_commit` to no. That *should* both match the previous OOTB defaults and avoid changing behavior for an existing deployment with `scm = git` in config...

**Comment by AdamWill** [comment]
we could also finesse this by having a hidden default setting for `scm` which is *not* `git`. call it `git-but-no-ac`. Then:

1. if `scm = git`, `git_auto_*` are treated as yes *unless* they are set to no
2. if `scm = git-but-no-ac`, `git_auto_commit` is treated as no *unless* it's set to yes, `git_auto_push` and `git_auto_update` are treated as yes unless they're set to no

and the default values for each would be empty string.

The effect of this would be that if you have an explicit `scm = git` in your config, all three auto behaviours would be on unless you explicitly set them to 'no'. But if you don't have an explicit `scm` setting in your config, auto_push and auto_update would be on unless you set them to 'no', but auto_commit would be off unless you set it to 'yes'. If `scm` is explicitly set to anything but 'git', all three would be off.

This would be a bit weird to implement and document, though, I guess.

**Comment by AdamWill** [comment]
alternatively we could do that but treat *all three* auto options as 'default no' if scm is not *explicitly* set to 'git', but 'default yes' if scm is *explicitly* set to 'git'. I kinda like that option actually. let me see if I can draft it up.

File: etc/openqa/openqa.ini [hunk 2]
```diff
@@ -108,7 +108,7 @@
 ## if 'scm' is set to 'git' in [global]
 #[scm git]
 ## whether to automatically commit needles created in the editor back to git
-#git_auto_commit = yes
+#git_auto_commit = no
```
b10n1k, line 111: if git features are enabled by default why is that changed?

AdamWill: what we're trying to achieve (AIUI) is for `git_auto_push` and `git_auto_update` to be default-enabled, but `git_auto_commit` to be default-disabled, because that's how things were historically and also what we think is appropriate (because we perceive `git_auto_commit` to be a more intrusive thing than the other two).

the tricky part is implementing this in such a way that existing config files maintain their meaning (i.e. `scm = git` turns `git_auto_commit` on, but without `scm = git`, it's off).

again, AIUI.

okurz: > what we're trying to achieve (AIUI) is for `git_auto_push` and `git_auto_update` to be default-enabled, but `git_auto_commit` to be default-disabled, because that's how things were historically and also what we think is appropriate (because we perceive `git_auto_commit` to be a more intrusive thing than the other two).

I think you mean `git_auto_clone` instead of `git_auto_push` which does not exist, only `do_push`. But yes to the other points

> the tricky part is implementing this in such a way that existing config files maintain their meaning (i.e. `scm = git` turns `git_auto_commit` on _even without an explicit `git_auto_commit` setting_, but without `scm = git`, it's off).
> 
> again, AIUI.

Yes, correct

AdamWill: edited, thanks.

b10n1k: got it

Wabri (approved):

AdamWill:
Note there was also a lot of discussion in https://github.com/os-autoinst/openQA/pull/6430#issuecomment-2855908321 , and I'm on the hook to propose something based on that discussion and this PR.

AdamWill:
oh, sorry, in case anyone wondered: I've been distracted by https://github.com/os-autoinst/openQA/pull/6446 , which is why I didn't get back to this yet.

AdamWill:
OK, https://github.com/os-autoinst/openQA/pull/6472 is my new effort here.

PR #6426, opened 2025-05-05
Revert "Block all distri git operations unless `scm = git` is set" due to unexpected change of default behaviour

okurz:
Reverts os-autoinst/openQA#6414 due to unexpected change of default behaviour. I tried to ensure that git auto-clone+update are still enabled by default with #6421 by setting the global config key `scm` by default to `git` however that also means that automatic git commits of needles are enabled. #6414 introduced a new config for automatic git commits but disabling that by default also breaks compatibility with existing instances so I see no better option than to fully revert #6414

### DIFF
--- a/docs/Installing.asciidoc
+++ b/docs/Installing.asciidoc
@@ -725,14 +725,8 @@ section.
 
 === Setting up git support
 
-If your tests and needles are stored in git, openQA can perform various operations:
-
-* Automatically commit needles created in the web UI editor back to the repository
-* Automatically update the repository when scheduling tests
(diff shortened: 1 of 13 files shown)


okurz:
> It's no longer possible to avoid _some_ kind of "breaking change" for _somebody_, here. It has not been since #6039 was merged.

Yes, that is correct. I would like to have the git auto update clone handled better to not cause problems

AdamWill:
Not quite sure what you mean by "handled better to not cause problems"?

PR #6171, opened 2025-02-11
Enable auto updates of tests/needles by default again

Martchus:
* Reapply "Enable automatic updates of test code and needles by default" (commit 40ae05a38ed2ee6c686eba7fde715a04cb00e33e) again after fixing problems with scheduling and restarting of parallel clusters
* See https://progress.opensuse.org/issues/168379

PR #5622, opened 2024-05-06
Add GRU task to update DISTRI and NEEDLES_DIR from git if needed

asdil12:
Ticket: https://progress.opensuse.org/issues/159171

PR #6430, opened 2025-05-06
Reapply "Block all distri git operations unless `scm = git` is set, a…

AdamWill:
…dd `git_auto_commit` setting, disentangle `git_auto_clone` and `git_auto_update`"

This reverts commit d4a551b0440a518ba45ee4f62b6f619821dd73db - i.e. this is a re-run of #6414 . I still believe implementing the changes from #6414 leaves things in an overall better state.

### DIFF
--- a/docs/Installing.asciidoc
+++ b/docs/Installing.asciidoc
@@ -725,8 +725,14 @@ section.
 
 === Setting up git support
 
-Editing needles from web can optionally commit new or changed needles
-automatically to git. To do so, you need to enable git support by setting
+If your tests and needles are stored in git, openQA can perform various operations:
+
+* Automatically commit needles created in the web UI editor back to the repository
+* Automatically update the repository when scheduling tests
+* Clone tests and needles from git repos specified in a job's `CASEDIR` and `NEEDLES_DIR`
+variables
+
+To enable these features, you need to enable git support by setting
 
 [source,ini]
 --------------------------------------------------------------------------------
@@ -734,8 +740,22 @@ automatically to git. To do so, you need to enable git support by setting
 scm = git
 --------------------------------------------------------------------------------
 in <<GettingStarted.asciidoc#_configuration,the web UI configuration>>.
-Once you do so and restart the web interface, openQA will automatically commit
-new needles to the git repository.
+Once you do so and restart the web interface, all of the above features will be enabled.
+
+To disable individual features, you can use these config settings:
+
+[source,ini]
+--------------------------------------------------------------------------------
+[scm git]
+git_auto_commit = no
+git_auto_clone = no
+git_auto_update = no
+--------------------------------------------------------------------------------
+
+`git_auto_commit` controls whether needles created or edited in the webUI editor are
+automatically committed. `git_auto_update` controls automatic test/needle updating
+when scheduling tests. `git_auto_clone` controls the automatic cloning of repos
+referenced by `CASEDIR` and `NEEDLES_DIR`.
 
 You may want to add some description to automatic commits coming from the web
 UI.
--- a/etc/openqa/openqa.ini
+++ b/etc/openqa/openqa.ini
@@ -104,7 +104,11 @@
 ## can cause jobs to incomplete with "timestamp mismatch" error messages.
 #api_hmac_time_tolerance = 300
 
+## Configuration related to actions openQA may perform on a git distri
+## if 'scm' is set to 'git' in [global]
 #[scm git]
+## whether to automatically commit needles created in the editor back to git
+#git_auto_commit = yes
 ## name of remote to get updates from before committing changes (e.g. origin, leave out-commented to disable remote update)
 #update_remote = origin
 ## name of branch to rebase against before committing changes (e.g. origin/master, leave out-commented to disable rebase)
--- a/lib/OpenQA/Schema/Result/Needles.pm
+++ b/lib/OpenQA/Schema/Result/Needles.pm
@@ -176,7 +176,7 @@ sub remove ($self, $user) {
     $app->log->debug("Remove needle $fname and $screenshot");
 
     my $git = OpenQA::Git->new({app => $app, dir => $self->directory->path, user => $user});
-    if ($git->enabled) {
+    if ($git->enabled && $git->config->{git_auto_commit} eq 'yes') {
         my $directory = $self->directory;
         my $error = $git->commit(
             {
--- a/lib/OpenQA/Setup.pm
+++ b/lib/OpenQA/Setup.pm
@@ -100,6 +100,7 @@ sub read_config ($app) {
             require_for_assets => 0,
         },
         'scm git' => {
+            git_auto_commit => 'yes',
             update_remote => '',
             update_branch => '',
             do_push => 'no',
--- a/lib/OpenQA/Shared/Plugin/Gru.pm
+++ b/lib/OpenQA/Shared/Plugin/Gru.pm
@@ -235,8 +235,9 @@ sub enqueue_download_jobs ($self, $downloads, $minion_ids = undef) {
 }
 
 sub enqueue_git_update_all ($self) {
-    my $conf = OpenQA::App->singleton->config->{'scm git'};
-    return if $conf->{git_auto_clone} ne 'yes' || $conf->{git_auto_update} ne 'yes';
+    my $conf = OpenQA::App->singleton->config;
+    return if ($conf->{global}->{scm} // '') ne 'git';
+    return if $conf->{'scm git'}->{git_auto_update} ne 'yes';
     my %clones;
     my $testdir = path(sharedir() . '/tests');
     for my $distri ($testdir->list({dir => 1})->each) {
@@ -263,7 +264,8 @@ sub enqueue_git_update_all ($self) {
 
 sub enqueue_git_clones ($self, $clones, $job_ids, $minion_ids = undef) {
     return unless keys %$clones;
-    return unless OpenQA::App->singleton->config->{'scm git'}->{git_auto_clone} eq 'yes';
+    my $conf = OpenQA::App->singleton->config;
+    return unless ($conf->{global}->{scm} // '') eq 'git';
     # $clones is a hashref with paths as keys and git urls as values
     # $job_id is used to create entries in a related table (gru_dependencies)
 
--- a/lib/OpenQA/Task/Git/Clone.pm
+++ b/lib/OpenQA/Task/Git/Clone.pm
@@ -26,6 +26,9 @@ sub register ($self, $app, @) {
 sub _git_clone_all ($job, $clones) {
     my $app = $job->app;
     my $job_id = $job->id;
+    # failsafe check - we should never actually get here with git
+    # disabled, but just in case
+    return if (($app->config->{global}->{scm} // '') ne 'git');
 
     my $retry_delay = {delay => 30 + int(rand(10))};
     # Prevent multiple git_clone tasks for the same path to run in parallel
@@ -75,14 +78,13 @@ sub _git_clone_all ($job, $clones) {
 
         # unblock openQA jobs despite network errors under best-effort configuration
         my $retries = $job->retries;
-        my $git_config = $app->config->{'scm git'};
         my $max_retries = $ENV{OPENQA_GIT_CLONE_RETRIES} // 10;
         my $max_best_effort_retries = min($max_retries, $ENV{OPENQA_GIT_CLONE_RETRIES_BEST_EFFORT} // 2);
         my $gru_task_id = $job->info->{notes}->{gru_id};
         if (   $is_path_only
             && defined($gru_task_id)
             && ($error =~ m/disconnect|curl|stream.*closed|/i)
-            && $git_config->{git_auto_update_method} eq 'best-effort'
+            && $git->config->{git_auto_update_method} eq 'best-effort'
             && $retries >= $max_best_effort_retries)
         {
             $app->schema->resultset('GruDependencies')->search({gru_task_id => $gru_task_id})->delete;
--- a/lib/OpenQA/Task/Needle/Save.pm
+++ b/lib/OpenQA/Task/Needle/Save.pm
@@ -101,7 +101,7 @@ sub _save_needle {
 
     # ensure needle dir is up-to-date
     my $git = OpenQA::Git->new({app => $app, dir => $needledir, user => $user});
-    if ($git->enabled) {
+    if ($git->enabled && $git->config->{git_auto_commit} eq 'yes') {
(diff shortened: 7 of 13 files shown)


AdamWill:
@okurz reverted #6414 in https://github.com/os-autoinst/openQA/pull/6426 . Having considered it, I disagree with that, and so I'm opening this PR to restore #6414.

Let me try to summarize the situation as best I can. As a convention, I'll refer to the config setting `bar` in the config section `[foo]` as `[foo].bar` sometimes.

The `[scm git]` config section was introduced in 0d26009147f41a33cda1ff954bb40a3e3ea7198d in 2014. In that commit and in all subsequent commits until recently, the settings in the `[scm git]` config section were only relevant if `[general].scm` section was set to `git`. At that time the default value of `scm` was `git`, but it was changed to `undef` in 973a41a13755c868c0f2031c6e084be5ba447514 in 2015. From that time onwards, until last year, this was the state:

1. `[general].scm` controlled whether needle creation / deletion was automatically committed to git; this happened if it was set to `git`, otherwise it did not happen. No other automatic git operations existed. This setting defaulted to `undef` (i.e. off).
2. The `[scm git]` section contained only options that fine-tuned the automatic git commit behaviour. No option in that section did anything meaningful unless `[general].scm` was set to `git`.

This changed in https://github.com/os-autoinst/openQA/pull/5622 . That introduced a `[scm git].git_auto_clone` config section which controls whether openQA will assume the server's distri dir is a git repo and attempt to update it sometimes. However, this setting was not made to depend on `[general].scm`. Regardless of what `[general].scm` was set to, auto-clone behavior happened if `[scm git].git_auto_clone` was 'yes', otherwise it did not happen.

In a later commit, `[scm git].git_auto_update` was added with similar behaviour. In https://github.com/os-autoinst/openQA/pull/6039 , `[scm git].git_auto_update` and `[scm git].git_auto_clone` were set on by default. This was reverted soon after, but reapplied in https://github.com/os-autoinst/openQA/pull/6171 .

So as of #6171 , this was the state:

1. `[general].scm` controlled whether needle creation / deletion was automatically committed to git; this happened if it was set to `git`, otherwise it did not happen. This setting defaulted to `undef` (i.e. off).
2. Most options in `[scm git]` still fine-tuned the behavior of needle auto commit. However, the settings `[scm git].git_auto_clone` and `[scm git].git_auto_update` controlled new automatic git behaviors, and these settings took effect regardless of what `[general].scm` was set to. These options defaulted to on.

For me, this has three obvious drawbacks:

1. The semantics are awful. The plain words of the config file *and* past behaviour imply that settings in `[scm git]` will not do anything unless `[general].scm` is set to `git`, but this is not true.
2. The enablement of `git_auto_clone` and `git_auto_update` by default is a significant and potentially breaking change that is not obvious to a sysadmin who did not closely read the relevant PRs/git commit messages.
3. The mechanisms for configuring three similar behaviors (git automatic clone, git automatic update, and git automatic commit) are inconsistent. The former two are consistent with each other, but the latter one is completely different.

So, in #6414, I sent a series of changes that, together, resulted in the following state:

1. `[general].scm` is a 'master switch' for all automatic git behaviours. If it's set to `git`, they may happen, depending on the settings in `[scm git]`. If it's set to anything else, or unset, they will not. It still defaults to `undef`, so all automatic git behaviors are disabled out of the box.
2. There are three consistently-named options in `[scm git]` that control the similar 'automatic' behaviors - `git_auto_clone`, `git_auto_update` and `git_auto_commit`. All three default to `yes`, so *if you set* `[general].scm = git`, all three turn on.

This was merged. However, it [caused failures in the openQA-in-openQA test](https://progress.opensuse.org/issues/181772). @okurz argued that auto-clone and auto-update behavior should still be on by default, so sent https://github.com/os-autoinst/openQA/pull/6421 , which attempted to implement that while keeping my overall design. However, through a couple of revisions he kept running into problems achieving his exact desired behavior, so he instead reverted my #6414 entirely in #6426. This means that the current state is the same as the state "as of #6171" which I described above, and still has all the same drawbacks, which to me are significant ones.

In https://github.com/os-autoinst/openQA/pull/6421#issuecomment-2852860308 and https://github.com/os-autoinst/openQA/pull/6421#issuecomment-2852895184 I outlined a couple of ways we could, I believe, maintain a semantically-consistent and sane config file layout, but achieve the behavior @okurz wants.

But having thought it through for a while, I think I first want to re-propose that we just go ahead with #6414 as-was, and achieve the desired behavior for SUSE's instances and the openQA-in-openQA test by setting the desired configuration there. I don't think that enabling auto-clone and auto-push out of the box is a good idea, given that there was, prior to 2024, nine years of precedent that openQA would not assume it could treat the server's tests/needles as a git repo it could mess around with unless `[general].scm` was set to `git`. *Additionally*, but only as a relatively minor point, I will note that it's quite difficult to achieve @okurz 's desired state while having a configuration file format that makes any kind of *sense*, AFAICS. Both the approaches I proposed are a bit subtle and require careful implementation and documentation. Neither of them is what you'd design if you were designing this from scratch.

I'm gonna argue the fallout from changing that expectation is *worse* than requiring a single explicit config value for new, modern deployments that want to point to a git repo for tests/needles and have it automatically updated. It's pretty easy to just write into the instructions "to do this, you need to set `scm = git` in the config file".

If I lose this argument, I'm willing to work on one of the options I outlined in comments on #6421 , because I do think either of those is *preferable* to the current state. But first I want to propose this.

AdamWill:
I guess it would be fair to note that, with my plan, if you have a pre-2024 instance *with `[general].scm` set to `git`*, you still get the new auto-clone and auto-update behaviors enabled by default, which might surprise you. If anyone thinks that's a problem, we could set their defaults to no, I guess. That's easy enough to do.

AdamWill:
So reading through those, what it looks like to me is that we kinda have two possible workflows. There's the legacy workflow, where there is a set of distris stored in a filesystem on the server which is canonical. At the time a test is created, it runs against the current state of the given distri on the server.

This workflow is pretty simple and clear, but it makes development somewhat difficult, because you can't really run a single job against development code very easily. To do development you need an instance where it's okay for *all* jobs to use your development code, at least temporarily. And if multiple people are trying to develop different things at once, they need separate 'development servers' or very careful co-ordination of a shared development branch, or something.

So that's the impetus (AIUI) for the new workflow we're kinda stumbling towards, where you can cause specific jobs to be run against different test/needle codebases. It seems *to me* like the way this should work is that it should be a completely different workflow which, at least in intent, does not depend on the state of a directory on the server *at all*. In this workflow, the job variables define a git repo where the test code and needles should be taken from, and the repo state that should be used. This should be kept as the canonical test/needle state for that job forever; the worker should execute the test using that state of the tests and needles, and the web UI should display the test code and needles from that state of that repository. (Presumably, even if the requested state is something like 'repo foo, current master branch', we should show the test code and needles from master *at the time the test ran*, even if they're changed later?)

Also, it should be possible to use the web UI needle editor to create new needles that are committed back to that repository (and I guess to the relevant branch, if the defined repo state included a branch name? not sure about this)

Anyhow - it seems to me some of the awkwardness here is caused by the details of how we're *implementing* the new workflow. I think the way I'd want it to work is that, for *any* job which specifies to get the needles/test code from a git repo state, auto-clone/auto-update (at least) should be assumed - but this should *never* apply to the `/var/lib/openqa/share/tests` (or whatever) folder on the server. That should *only* be used for jobs which do *not* have variables specifying a git repo to get tests/needles from (i.e. jobs that follow the 'legacy' workflow). And in consistency with past behavior, those directories should *never* be assumed to be git repos and automatically updated by default.

I guess I'm saying we should never *use* the single magic tree of distris on the server *in the new, git-y workflow*. We should never check a folder under that tree out to be some arbitrary git state for a job that just got scheduled. That tree should always be under the sysadmin's control. But that tree should also be *irrelevant* to jobs which want to use the new, git-y workflow. Wherever we actually check out the tests and needles that any given job's variables point to, it should *NOT* ever be `/var/lib/openqa/share/tests/[distri]`. if it wasn't, that might help solve a lot of the problems here? We could still have the new, more development-friendly, git-based workflow available, but without behaviors that mess with established expectations for the legacy workflow.

Am I misunderstanding things here?

AdamWill:
hmm, to put it another way, I guess, my biggest problem here is with https://progress.opensuse.org/issues/164898 / 101f9b9d134dc9b176843e419f26acc5ca2c1163 , which seems to be a really different thing from the whole 'how do we do test development?' question. It's really the `git_auto_update` behavior I disagree with, not the `git_auto_clone` behavior, and the only justification for `git_auto_update` was "we want to replace this `fetchneedles` script we have on the SUSE instances".

So...could we get somewhere interesting here by treating `git_auto_clone` as something very different to `git_auto_update` and `git_auto_commit`? The more I look at it, the more that seems to be the true state of things. `git_auto_commit` and `git_auto_update` seem very much to do with the 'legacy workflow', and I'd argue it'd be appropriate for them to be controlled by `[general].scm = git` and neither to be enabled by default.

`git_auto_clone` seems much more to be related to the whole 'we need a good development workflow' question. You could almost view it as a setting which controls whether the New Git Workflow is 'enabled' for a given instance, but which still doesn't *do* anything unless you schedule jobs with the relevant settings. I can see the argument for having it enabled by default - but maybe it shouldn't be in `[scm git]`, because it's really a different *thing*, it's not about 'having openQA control the canonical distris used by the legacy workflow as git repos', but 'enabling the new different-distri-state-per-job-controlled-by-git workflow'.

okurz:
> Let's say in theory we agree:
>     1. Git Auto Commit should be disabled by default
>     2. Git Auto Update should be disabled by default
>     3. Git Auto Clone should be enabled by default
> 
> then that's great...however, it still leaves us with the config file mess to sort out.

Yes and I very much appreciate even your first approach in https://github.com/os-autoinst/openQA/pull/6414 to sort this out. I just think it's a bit easier if we first and very explicitly in its own pull request do `git_auto_update` which I assume is what you found as first offender.

> Yes, you're right I wrote 'general' instead of 'global', sorry (edited). I understand why you'd say we can "consider `[global].scm = git` a legacy and badly named option to enable what you called `git_auto_commit` and that only" in theory, but it's...kinda awkward in practice, because there's _also_ the whole config file section called `[scm git]`, which is clearly a reference to that option. If we only go as far as treating `[global].scm` as a legacy version of `[scm git].git_auto_commit`, it raises the awkward question of why there's a config file section called `[scm git]` at all, and what that means. But if we go further and try to move all the things that are in `[scm git]` out of it but maintain backwards compatibility, that gets kinda big and messy in a hurry.
>
> I can _try_ and draft up something that moves the three `git_auto` options to the global level, treats global `scm` as a deprecated equivalent of global `git_auto_commit`, and somehow moves the sub-options around while maintaining backward compatibility, though. And we would default `git_auto_commit` and `git_auto_update` to `no`, but `git_auto_clone` to `yes`. It's probably going to need some helper functions in `Setup.pm` that do all the heavy lifting.

No wait. I am in support for your approach in #6414 and the config options all look nice and dandy. I am merely looking for an approach how to ensure that new clean openQA instances do `git_auto_clone`.

okurz:
One idea I had was together with the changes in this PR moving and renaming `[scm git].git_auto_clone` to `[global].auto_clone_scm` so that we can treat it independently of all other git options. But reading through the older tickets and pull requests where we introduced `git_auto_update` I still see this as a good default, at least for *new* instances. But let me answer to an earlier paragraph by you

> I'm gonna argue the fallout from changing that expectation is _worse_ than requiring a single explicit config value for new, modern deployments that want to point to a git repo for tests/needles and have it automatically updated. It's pretty easy to just write into the instructions "to do this, you need to set `scm = git` in the config file".

The point should be that new instances work by default because people likely only read instructions if they encounter problems on setup. And consider short-lived throw-away instances or container based bootstrapped openQA instances like in http://open.qa/docs/#_how_to_run_a_test_with_single_instance_container_in_5_minutes . We don't want users to need to specify any setting.

With that and reading more of your comments in other pull requests I think I like your approach from https://github.com/os-autoinst/openQA/pull/6421#issuecomment-2852895184 . Did you already think about implementing that? Would you be ok to look into that?

AdamWill:
Well, the problem is the design from #6414 is nice and clean, but we can't keep that design *and* keep pre-5622 behavior for existing instances *and* have any of the auto options enabled out of the box :(

If we keep that design, we can only keep pre-5622 default behavior by keeping the default value of `[global].scm` as undefined (or, anything other than `git`). If we have it default to `git`, we cannot preserve existing auto-commit behavior, because:

* If we set `[scm git].git_auto_commit` off by default, then autocommit would be disabled for existing instances that have `[global].scm` set to `git`, which is a change.

* If we set `[scm git].git_auto_commit` on by default, autocommit would be enabled for existing instances that have `[global].scm` *not* set to `git`, which is a change.

So we can only preserve existing auto-commit behavior by having `[global].scm` default to off, but *that* means we can't have auto-clone default to on (in that design).

We can keep my design but only use it for auto-commit and auto-update - that's what I was suggesting above. That would look *something* like this, as a mock config file showing default values:

```
[global]
scm = undef
git_auto_clone = yes
checkout_needles_sha = no
allow_arbitrary_url_fetch = no
temp_needle_refs_retention = 120

[scm git]
git_auto_commit = yes
update_remote = origin
update_branch = origin/master
do_cleanup = no
do_push = no
git_auto_update = yes
```

That I think achieves all the goals, but I don't really *love* it. It's fun to imagine how you'd write the comment for the `scm` setting, in this case - "this controls whether openQA treats the canonical distri dirs as git repos, except for git_auto_clone which is its own setting for stupid reasons"? :D

Now I think about it, `checkout_needles_sha` and the other two related options are pretty awkward too. They're *kinda* related to `git_auto_clone` but kinda not - they still will do stuff if `git_auto_clone` is disabled, in fact you can almost look at them as *alternative implementations* of how to deal with needles as per-job git repos. `git_auto_clone` just fetches the repo in the 'canonical' location at job creation time, and updates to the current head of the branch if the requested branch is the one that's currently checked out. `checkout_needles_sha` is doing something more sophisticated, like what I envisaged, where it actually does a temporary unique git checkout that's *specific to the job you're trying to view* in the web UI (AIUI).

AdamWill:
I guess in my "ideal world" `git_auto_clone` would not exist, but `checkout_needles_sha` would default to 'yes', and would also work for test code, not just needles. And any bugs/performance problems/security issues in it would be fixed, of course. Heh.

AdamWill:
oh, sorry, in case anyone wondered: I've been distracted by https://github.com/os-autoinst/openQA/pull/6446 , which is why I didn't get back to this yet.

Martchus:
I wasn't wondering :-)

AdamWill:
OK, https://github.com/os-autoinst/openQA/pull/6472 is my new effort here.

PR #6435, opened 2025-05-08
On shutdown keep openqa-webui and web server running until gru stopped

asdil12:
Ticket: https://progress.opensuse.org/issues/181370

### DIFF
--- a/systemd/openqa-webui.service
+++ b/systemd/openqa-webui.service
@@ -1,7 +1,8 @@
 [Unit]
 Description=The openQA web UI
 Wants=openqa-setup-db.service openqa-livehandler.service openqa-websockets.service openqa-gru.service openqa-enqueue-asset-cleanup.timer openqa-enqueue-result-cleanup.timer openqa-enqueue-bug-cleanup.timer openqa-minion-restart.path
-After=postgresql.service openqa-setup-db.service openqa-scheduler.service nss-lookup.target remote-fs.target
+After=postgresql.service openqa-setup-db.service openqa-scheduler.service nss-lookup.target remote-fs.target apache2.service nginx.service caddy.service
+Before=openqa-gru.service
 
 [Service]
 User=geekotest
--- a/systemd/openqa-webui.service
+++ b/systemd/openqa-webui.service
@@ -1,7 +1,7 @@
 [Unit]
 Description=The openQA web UI
 Wants=openqa-setup-db.service openqa-livehandler.service openqa-websockets.service openqa-gru.service openqa-enqueue-asset-cleanup.timer openqa-enqueue-result-cleanup.timer openqa-enqueue-bug-cleanup.timer openqa-minion-restart.path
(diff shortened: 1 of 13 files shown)


systemd/openqa-webui.service
```diff
@@ -1,7 +1,8 @@
 [Unit]
 Description=The openQA web UI
 Wants=openqa-setup-db.service openqa-livehandler.service openqa-websockets.service openqa-gru.service openqa-enqueue-asset-cleanup.timer openqa-enqueue-result-cleanup.timer openqa-enqueue-bug-cleanup.timer openqa-minion-restart.path
-After=postgresql.service openqa-setup-db.service openqa-scheduler.service nss-lookup.target remote-fs.target
+After=postgresql.service openqa-setup-db.service openqa-scheduler.service nss-lookup.target remote-fs.target apache2.service nginx.service caddy.service
```
nicksinger, line 4: Wouldn't this allow a short period on startup where a webserver is running but cannot serve requests because the backend is still starting up?

nicksinger: Given startup seems to be *much* quicker than shutdown, I think it is an acceptable trade-off. But it would still be nice to have both I guess?

okurz (approved):
I would have preferred two separate commits for reverse proxy services and openQA services only but couldn't convince @asdil12

I plan to add another commit on top to remove `After=openqa-scheduler`

okurz (approved):

asdil12:
Manually tested. The services are started and stopped in the expected order.
Disabled services (on my machine: apache2) are not affected and do not interfere.
Neither do not installed services (on my machine: caddy).

okurz:
As we have OBS repositories triggered with https://download.opensuse.org/repositories/devel:/openQA:/GitHub:/os-autoinst:/openQA:/PR-6435/openSUSE_Tumbleweed/devel:openQA:GitHub:os-autoinst:openQA:PR-6435.repo I suggest to trigger openQA-in-openQA tests from https://openqa.opensuse.org/tests/overview?distri=openqa&version=Tumbleweed&build=%3ATW.36424&groupid=24, in particular "openqa_install+publish" and "openqa_install_nginx". You just need to set the test variable OPENQA_REPO_URL, see https://github.com/os-autoinst/os-autoinst-distri-openQA/blob/74ab68383985389dc41a5f2ce06ba5d0581c11a5/tests/install/openqa_webui.pm#L17C30-L17C44 so something like

```
for i in 5048623 5048625; do
    openqa-clone-job --within-instance https://openqa.opensuse.org/$i _GROUP=0 {TEST,BUILD}+=-gh6435 OPENQA_REPO_URL=https://download.opensuse.org/repositories/devel:/openQA:/GitHub:/os-autoinst:/openQA:/PR-6435/openSUSE_Tumbleweed/
```

asdil12:
https://openqa.opensuse.org/tests/5048785
https://openqa.opensuse.org/tests/5048786

okurz:
Apparently container checks are still missing even though they should have been directly triggered by github. Other recent PRs like https://github.com/os-autoinst/openQA/pull/6437 have those checks so I guess this is a one-off so far and I will merge manually.

PR #6437, opened 2025-05-08
Create new systemd service and timer for database dump and cleanup

gauravpathak:
Related to: https://progress.opensuse.org/issues/181304


systemd/openqa-dump-db.service
```diff
@@ -0,0 +1,9 @@
+[Unit]
+Description=Executes openQA database dump and cleanup for openQA.
+After=postgresql.service openqa-setup-db.service
+Wants=openqa-setup-db.service
```
okurz, line 4: Why would we need openqa-setup-db?

gauravpathak: Ok, understood. This should get executed as soon as `/var/lib/openqa` mount is available. `postgres` user is created during postgres db installation and not by our `openqa-setup-db.service`

systemd/openqa-dump-db.timer
```diff
@@ -0,0 +1,9 @@
+[Unit]
+Description=Schedules openQA database dump and cleanup for openQA every day.
```
asdil12, line 2:
```suggestion
Description=Daily openQA database dump and cleanup task
```

asdil12 (changes requested):

systemd/openqa-dump-db.service
gauravpathak: This is needed to ensure starting the timer unit on a fresh installation and every time at boot.

`openqa-setup-db.service` is dependent on `postgresql.service`, in our documentation, there's an instruction to start and enable `postgresql.service` https://open.qa/docs/#_run_the_web_ui

asdil12 (approved):

Makefile
```diff
@@ -71,9 +71,10 @@ install-generic:
 		install -m 644 -D --target-directory="$(DESTDIR)/usr/share/openqa/$${f%/*}" "$$f";\
 	done
 
-	for i in db images testresults pool/1 cache webui/cache; do \
+	for i in db images testresults pool/1 cache webui/cache backup; do \
 		mkdir -p "$(DESTDIR)"/var/lib/openqa/$$i ;\
 	done
+	ln -sfn /var/lib/openqa/backup "$(DESTDIR)"/var/lib/openqa/SQL-DUMPS
```
Martchus, line 77: Do we really need this symlink?

gauravpathak: If we remove this symlink we have to modify the database backup script as well.

gauravpathak: The default [DUMP_FOLDER](https://github.com/os-autoinst/openQA/blob/master/script/dump-db#L3) is the symlink to `/var/lib/openqa/backup`, both of which aren't created by default at present during installation.

okurz (changes requested):
With this change by default database backup is enabled which might pose problems for some setups so I don't know if we need to enable it by default. Please also check where the documentation needs to be extended

gauravpathak:
> With this change by default database backup is enabled which might pose problems for some setups so I don't know if we need to enable it by default. Please also check where the documentation needs to be extended

Yeah sure, we can make it optional and mention it in the document.
I think it should be mentioned in or after: https://open.qa/docs/#_run_the_web_ui

Makefile
gauravpathak: I think it won't break anything even if we change the script to use the `backup` directory and completely ignore `SQL-DUMPS` in the script.
 `SQL-DUMPS` will keep following what's inside `backup` as it is a symlink.

okurz:
@nicksinger WDYT?

systemd/openqa-dump-db.service
```diff
@@ -0,0 +1,9 @@
+[Unit]
+Description=openQA database dump and cleanup task
+ConditionPathIsReadWrite=/var/lib/openqa/backup
+ConditionPathIsSymbolicLink=/var/lib/openqa/SQL-DUMPS
```
Martchus, line 4: This line is no longer required after modifying the script.

```suggestion
```

systemd/openqa-dump-db.service
kalikiana: I guess this was removed, hence resolving the thread

docs/Installing.asciidoc
```diff
@@ -633,6 +633,16 @@ The openQA web UI should be available on http://localhost/ now. To simply
...
+== Enable automatic database backup
+An optional systemd service, `openqa-dump-db.service`, can be enabled to perform daily database backups. This service is triggered by the `openqa-dump-db.timer`.
+To enable automatic database backup, run:
+
+[source,sh]
+--------------------------------------------------------------------------------
+systemctl enable --now openqa-dump-db.timer
+--------------------------------------------------------------------------------
+Backups are stored at `/var/lib/openqa/backup`
+
```
nicksinger, line 645: I haven't read the full section but should we mention that `openQA-local-db` needs to be installed before these services/timers can even work?

nicksinger (approved):

docs/Installing.asciidoc
gauravpathak: I think it can create confusion. `openQA-local-db` package get automatically pulled with openQA: https://github.com/os-autoinst/openQA/blob/master/dist/rpm/openQA.spec#L118

nicksinger: Agreed. If you think the user cannot easily end up with the service/timer installed but without the script, we're fine.

docs/Installing.asciidoc
```diff
@@ -633,6 +633,16 @@ The openQA web UI should be available on http://localhost/ now. To simply
 start openQA without enabling it permanently one can simply use `systemctl
 start` instead.
 
+== Enable automatic database backup
+An optional systemd service, `openqa-dump-db.service`, can be enabled to perform daily database backups. This service is triggered by the `openqa-dump-db.timer`.
```
Martchus, line 637: Please break this line at a similar length as it is done in surrounding paragraphs.

gauravpathak: Done

dist/rpm/openQA.spec
```diff
@@ -669,6 +671,8 @@ fi
 %dir %{_localstatedir}/lib/openqa/share/factory/hdd
 %dir %{_localstatedir}/lib/openqa/share/factory/repo
 %dir %{_localstatedir}/lib/openqa/share/factory/other
+%dir %{_localstatedir}/lib/openqa/backup
```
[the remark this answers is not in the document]
gauravpathak: @kalikiana yes, I made changes accordingly

Martchus (approved):

dist/rpm/openQA.spec
```diff
@@ -553,13 +554,13 @@ fi
...
-%service_add_post openqa-setup-db.service
+%service_add_post openqa_localdb_services
 
 %preun local-db
-%service_del_preun openqa-setup-db.service
+%service_del_preun openqa_localdb_services
 
 %postun local-db
-%service_del_postun openqa-setup-db.service
+%service_del_postun openqa_localdb_services
```
tacerus, line 563: Sorry if I'm missing something, but shouldn't these references be expanded to the variable defined earlier (i.e. something like `%{openqa_localdb_services}`)?

script/dump-db
```diff
@@ -1,6 +1,6 @@
 #!/bin/bash
 set -euo pipefail
-DUMP_FOLDER=${DUMP_FOLDER:-"/var/lib/openqa/SQL-DUMPS"}
+DUMP_FOLDER=${DUMP_FOLDER:-"/var/lib/openqa/backup"}
```
okurz, line 3: This breaks expectation of how o3 and osd handle backups. As I expect at this point it is only o3 and osd using this newly added script I think it's ok if we change it but we should be explicit and prepare for that. So please put this into a separate commit and adapt o3 and osd backup at least in https://gitlab.suse.de/qa-sle/backup-server-salt/-/blob/master/rsnapshot/rsnapshot.conf?ref_type=heads#L28 and https://gitlab.suse.de/openqa/salt-states-openqa/-/blob/master/etc/backup/rsnapshot_generic.conf?ref_type=heads#L28 and accordingly for OSD, etc.

okurz (changes requested):

script/dump-db
gauravpathak: I think it won't break it. We are not deleting existing symlink. So, `SQL-DUMPS` will continue to follow `backup` directory. If backup directory gets updated the same will be reflected in the `SQL-DUMPS`

gauravpathak: I have created merge requests.
- https://gitlab.suse.de/qa-sle/backup-server-salt/-/merge_requests/25/diffs
- https://gitlab.suse.de/openqa/salt-states-openqa/-/merge_requests/1457/diffs

docs/Installing.asciidoc
```diff
@@ -633,6 +633,17 @@ The openQA web UI should be available on http://localhost/ now. To simply
...
+An optional systemd service, `openqa-dump-db.service`, can be enabled to
+perform daily database backups. This service is triggered by the
+`openqa-dump-db.timer`. To enable automatic database backup, run:
+
+[source,sh]
+--------------------------------------------------------------------------------
+systemctl enable --now openqa-dump-db.timer
+--------------------------------------------------------------------------------
+Backups are stored at `/var/lib/openqa/backup`.
+
```
okurz, line 646: Please move this section further down, e.g. into "Advanced configuration". This section focusses on getting new instances up and running in the "basic configuration" and an optional database backup doesn't fit here.

okurz (changes requested):

okurz (approved):

Martchus (approved):

PR #6438, opened 2025-05-09
Allow comma-separated filtering of flavor

realcharmer:
https://progress.opensuse.org/issues/113432

### DIFF
--- a/lib/OpenQA/WebAPI/Plugin/Helpers.pm
+++ b/lib/OpenQA/WebAPI/Plugin/Helpers.pm
@@ -398,14 +398,8 @@ sub _compose_job_overview_search_args ($c) {
     # add simple query params to search args
     for my $arg (qw(distri version flavor test limit)) {
         next unless $v->is_valid($arg);
-        my $params = $v->every_param($arg);
-        my $param_count = scalar @$params;
-        if ($param_count == 1) {
-            $search_args{$arg} = $params->[0];
-        }
-        elsif ($param_count > 1) {
-            $search_args{$arg} = {-in => $params};
-        }
+        my @params = @{$v->every_param($arg) // []};
+        $search_args{$arg} = @params == 1 ? $params[0] : {-in => \@params} if @params;
     }
 
     # handle build separately
--- a/lib/OpenQA/WebAPI/Plugin/Helpers.pm
+++ b/lib/OpenQA/WebAPI/Plugin/Helpers.pm
@@ -392,6 +392,7 @@ sub _compose_job_overview_search_args ($c) {
     $v->optional('comment');
     $v->optional('groupid')->num(0, undef);
     $v->optional('modules', 'comma_separated');
+    $v->optional('flavor', 'comma_separated');
     $v->optional('limit', 'not_empty')->num(0, undef);
 
     # add simple query params to search args
--- a/t/ui/10-tests_overview.t
+++ b/t/ui/10-tests_overview.t
@@ -283,11 +283,18 @@ subtest 'filtering by flavor' => sub {
         $driver->find_element('#filter-flavor')->send_keys('DVD');
         apply_filter('DVD');
         like $driver->get_current_url, qr/flavor=DVD/, 'url contains single flavor';
-
         element_visible('#flavor_DVD_arch_i586', qr/i586/);
         element_visible('#flavor_DVD_arch_x86_64', qr/x86_64/);
         element_not_present('#flavor_GNOME-Live_arch_i686');
         element_not_present('#flavor_NET_arch_x86_64');
+        $driver->find_element('#filter-panel .card-header')->click();
+        $driver->find_element('#filter-flavor')->send_keys(',NET');
+        apply_filter('DVD,NET');
+        like $driver->get_current_url, qr/flavor=DVD.*NET/, 'url contains both flavors';
+        element_visible('#flavor_DVD_arch_i586', qr/i586/);
+        element_visible('#flavor_DVD_arch_x86_64', qr/x86_64/);
+        element_not_present('#flavor_GNOME-Live_arch_i686');
+        element_visible('#flavor_NET_arch_x86_64');
     };
 };
 
--- a/t/ui/10-tests_overview.t
+++ b/t/ui/10-tests_overview.t
@@ -296,6 +296,23 @@ subtest 'filtering by flavor' => sub {
         element_not_present('#flavor_GNOME-Live_arch_i686');
         element_visible('#flavor_NET_arch_x86_64');
     };
+
+    subtest 'options from url are carried over to the form' => sub {
+        $driver->get('/tests/overview?distri=opensuse&version=13.1&build=0091&flavor=DVD%2CNET');
+        like $driver->find_element('#filter-flavor')->get_attribute('value'), qr/DVD,NET/,
+          'comma-separated flavor options from the URL are present in the form';
+        element_visible('#flavor_DVD_arch_i586', qr/i586/);
+        element_visible('#flavor_DVD_arch_x86_64', qr/x86_64/);
+        element_not_present('#flavor_GNOME-Live_arch_i686');
+        element_visible('#flavor_NET_arch_x86_64');
+        $driver->get('/tests/overview?distri=opensuse&version=13.1&build=0091&flavor=DVD&flavor=NET');
+        like $driver->find_element('#filter-flavor')->get_attribute('value'), qr/DVD,NET/,
+          'multiple flavor options from the URL are present in the form';
+        element_visible('#flavor_DVD_arch_i586', qr/i586/);
+        element_visible('#flavor_DVD_arch_x86_64', qr/x86_64/);
+        element_not_present('#flavor_GNOME-Live_arch_i686');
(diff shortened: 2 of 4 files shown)


realcharmer:
Currently there is a regression where `&flavor=DVD&flavor=MicroOS-Image` doesn't return any results, but `&flavor=DVD%2CMicroOS-Image` does. Work in progress.

Martchus:
Looks good so far. Probably the invocation of `complex_query` needs to be adjusted to actually pass an array if multiple parameters are present. (This would maybe mean that as soon as one specifies a second `flavor` parameter the splitting by comma would not be done anymore within `complex_query`. That is probably not a problem but perhaps also not really intuitive. So maybe we always want to split by comma in `complex_query`, even if an array was passed.)

t/10-tests_overview.t
```diff
@@ -225,6 +225,15 @@ subtest 'todo-flag on test overview' => sub {
     $schema->txn_rollback;
 };
 
+subtest 'comma-separated filtering of flavor in query param' => sub {
```
b10n1k, line 228: I have the impression that this doesnt cover the poo expectations. Likely a UI test is needed to show that tht filter preserve the flavors.. or do I understand it wrong? 

this test it seems that validates basically that the controller can handle comma-separated values

realcharmer: Indeed, I am working on more test coverage now :)

okurz (dismissed):

t/ui/10-tests_overview.t
```diff
@@ -282,12 +282,37 @@ subtest 'filtering by flavor' => sub {
...
-
+        like $driver->get_current_url, qr/flavor=DVD/, 'url contains single flavor';
         element_visible('#flavor_DVD_arch_i586', qr/i586/);
         element_visible('#flavor_DVD_arch_x86_64', qr/x86_64/);
         element_not_present('#flavor_GNOME-Live_arch_i686');
         element_not_present('#flavor_NET_arch_x86_64');
+        $driver->find_element('#filter-panel .card-header')->click();
+        $driver->find_element('#filter-flavor')->send_keys(',NET');
+        apply_filter('DVD,NET');
+        like $driver->get_current_url, qr/flavor=DVD.*NET/, 'url contains both flavors';
```
b10n1k, line 293: same here. think of `/tests/overview?distri=opensuse&version=13.1&build=0091&flavor=DVD&HDD=NET`

t/ui/10-tests_overview.t
```diff
@@ -282,12 +282,37 @@ subtest 'filtering by flavor' => sub {
...
+        $driver->get('/tests/overview?distri=opensuse&version=13.1&build=0091&flavor=DVD%2CNET');
+        like $driver->find_element('#filter-flavor')->get_attribute('value'), qr/DVD,NET/,
+          'options are present in the form';
+        element_visible('#flavor_DVD_arch_i586', qr/i586/);
+        element_visible('#flavor_DVD_arch_x86_64', qr/x86_64/);
+        element_not_present('#flavor_GNOME-Live_arch_i686');
+        element_visible('#flavor_NET_arch_x86_64');
+        $driver->get('/tests/overview?distri=opensuse&version=13.1&build=0091&flavor=DVD&flavor=NET');
+        like $driver->find_element('#filter-flavor')->get_attribute('value'), qr/DVD,NET/,
+          'options are present in the form';
```
okurz, line 310: Please make that string unique so different from L303

perlpunk:
Please rebase to get rid of the dependency error

lib/OpenQA/WebAPI/Plugin/Helpers.pm
```diff
@@ -397,14 +397,8 @@ sub _compose_job_overview_search_args ($c) {
...
         next unless $v->is_valid($arg);
-        my $params = $v->every_param($arg);
-        my $param_count = scalar @$params;
-        if ($param_count == 1) {
-            $search_args{$arg} = $params->[0];
-        }
-        elsif ($param_count > 1) {
-            $search_args{$arg} = {-in => $params};
-        }
+        my @values = grep { length } map { split /\s*,\s*/ } @{$v->every_param($arg) // []};
```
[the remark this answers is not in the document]
perlpunk: I suggest to take a look at line 394:
```
$v->optional('modules', 'comma_separated');
```
This should save you from having to make any changes here in the loop. Just do the same for `flavor`.
The shortening with grep/map can be done in a separate commit, I would say.

perlpunk:
It seems that the javascript code filling out the form cannot deal with multiple parameters.
My suggestion would be to add that to `assets/javascripts/overview.js` around line 237.

templates/webapi/test/overview.html.ep
```diff
@@ -106,7 +106,9 @@
                     <div class="col-5">
                         <div class="mb-3">
                             <strong>Flavor</strong>
-                            <input type="text" class="form-control" name="flavor" placeholder="any" id="filter-flavor">
+                            <input type="text" class="form-control" name="flavor" id="filter-flavor"
+                            placeholder="flavors, comma separated, e.g. foo,bar"
+                            >
```
Wabri, line 111:
```suggestion
                            <input type="text" class="form-control" name="flavor" id="filter-flavor"
                            placeholder="flavors, comma separated, e.g. foo,bar">
```

Wabri (changes requested):

perlpunk (changes requested):
I tried your PR locally, and noticed that `modules_result` gets "lost" if I select more than one.
e.g. select "failed" and "softfailed", click apply and then open the filter form. `modules_result` is now empty.
The fact that this does not make any test fail, also means that we don't test this yet.

realcharmer:
> I tried your PR locally, and noticed that `modules_result` gets "lost" if I select more than one. e.g. select "failed" and "softfailed", click apply and then open the filter form. `modules_result` is now empty. The fact that this does not make any test fail, also means that we don't test this yet.

Thanks for catching this 🙏
I'll fix the JS and add more test coverage for this case.

> My suggestion would be to add that to assets/javascripts/overview.js around line 237.

As far as I understand it, we have to adjust how `parseFilterArguments` is parsing the options from URL to be able to combine weird cases such as `&flavor=foo%2Cbar&flavor=boo` to correctly unfold as `foo,bar,boo`. The `parseFilterArguments` function sits in `assets/javascripts/filter_form.js`. Or am I overthinking this?

perlpunk:
> . The parseFilterArguments function sits in assets/javascripts/filter_form.js.

yes, but it is called with a handler, which is defined in overview.js.

Wabri (approved):
I would change the first 3 commit description with the common practice of the repository, but approving if everyone are ok with them

perlpunk:
> I would change the first 3 commit description with the common practice of the repository [...]

what do you mean? In case it's the `t: ` at the beginning - for commits that are about tests that's fine

assets/javascripts/filter_form.js
```diff
@@ -55,10 +55,11 @@ function parseFilterArguments(paramHandler) {
   for (var i = 0; i < hiddenInputs.length; i++) {
     $('#filter-form').append(hiddenInputs[i]);
   }
+  filterLabels = [...new Set(filterLabels.flatMap(label => label.split(',')))];
   if (filterLabels.length > 0) {
     $('#filter-panel .card-header')
       .find('span')
-      .text('current: ' + filterLabels.join(', '));
+      .text('current: ' + filterLabels.join(','));
```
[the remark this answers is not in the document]
realcharmer: The original behavior (also expected by tests) is the form header contains options listed without a space, such as `foo,bar` and spaces were only put between different option groups, such as job result, job state and so on.
I now have to process each option separately, even comma-sepparated options, otherwise they would get printed multiple times, if multiple flavor attributes were set, eg. `&flavor=foo&flavor=bar` or even mixed `&flavor=foo&flavor=bar%2Cboo`. Unit tests were failing with the added space on line 61.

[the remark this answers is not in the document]
realcharmer: After implementing your latest suggestions, this code is not needed and has been removed. Thanks for the review.

Wabri:
> what do you mean? In case it's the t:  at the beginning - for commits that are about tests that's fine

Since we are following the https://cbea.ms/git-commit [Contributing#rules-for-commits](https://github.com/os-autoinst/openQA/blob/master/docs/Contributing.asciidoc#rules-for-commits), `t:` doesn't ensure rule 3 and 5.

realcharmer:
> > what do you mean? In case it's the t:  at the beginning - for commits that are about tests that's fine
> 
> Since we are following the https://cbea.ms/git-commit find in [Contributing#rules-for-commits](https://github.com/os-autoinst/openQA/blob/master/docs/Contributing.asciidoc#rules-for-commits), `t:` doesn't ensure rule 3 and 5.

The `t:` prefix was suggested by Oliver to have a clearer changelog where the prefix helps to describe what the change actually covers, since we list commit messages in there.

assets/javascripts/filter_form.js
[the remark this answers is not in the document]
perlpunk: oh, nice :)

PR #6439, opened 2025-05-09
Distribute openqa_devel container with podman

d3flex:
Running the openqa_devel as it is described, there is a complication of running scripts, which whatever their intention it should not fail. one of them is the `static_check_containers` which runs by default with podman but the container lacks the executable.

### DIFF
--- a/container/devel/Dockerfile
+++ b/container/devel/Dockerfile
@@ -20,7 +20,7 @@ LABEL org.opencontainers.image.created="%BUILDTIME%"
 # endlabelprefix
 
 # hadolint ignore=DL3037
-RUN zypper in -y os-autoinst-devel openQA-devel vim nodejs-default git mc perl-Devel-REPL perl-Term-ReadKey && \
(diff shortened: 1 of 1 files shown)


asdil12:
Could you give an example which error is to be fixed here?
Maybe some steps to reproduce the issue?
Installing podman within the container doesn't look right to me.

Martchus (commented):
I don't like this kind of nesting. It would probably make more sense to invoke tests/scripts that require containerized tooling from the outside.

That tests/scripts really on containerized tooling is probably the real problem here. We should probably provide on container that has everything and not really on different ones.

b10n1k:
> Could you give an example which error is to be fixed here? Maybe some steps to reproduce the issue? Installing podman within the container doesn't look right to me.

Running `podman run --rm -v /home/iob/openqadev/openQA/:/opt/openqa -e CHECKSTYLE=1  openqa_devel:latest make run-tests-within-container` it attempts to run `tools/static_check_containers` which expects a container engine in the environment.
the error comes, simply from the fact that CONTAINER_TEST is enabled and the environment doesnt have podman.

However it was discussed and this changes seems is minor and possible not needed. Unless you decide otherwise

PR #6445, opened 2025-05-12
Allow showing finished jobs blocking an incident on "All tests" page

Martchus:
See particular commit messages and https://progress.opensuse.org/issues/156553

---

It is perhaps not ideal to pass the regex unsanitized to to the database. Hence this is still a draft but I suppose it is sufficient as spike solution. (I first tried with a simple glob but I don't think this would be sufficient here, see the 2nd commit message.)

The performance is also problematic but this was of course expected. At least the performance is only impacted when the query parameter is used and when I tested it locally with OSD production data is was returning results after several seconds.

### DIFF
--- a/assets/javascripts/tests.js
+++ b/assets/javascripts/tests.js
@@ -257,9 +257,12 @@ function renderTestLists() {
       this[paramName] = paramValues[0];
     }
   };
-  jQuery.each(['limit', 'groupid', 'match', 'group_glob', 'not_group_glob', 'comment'], (index, paramName) => {
-    ajaxQueryParams.addFirstParam(paramName);
-  });
+  jQuery.each(
+    ['limit', 'groupid', 'match', 'group_glob', 'not_group_glob', 'job_setting', 'comment'],
+    (index, paramName) => {
+      ajaxQueryParams.addFirstParam(paramName);
+    }
+  );
   delete ajaxQueryParams.addFirstParam;
   filters.forEach(filter => {
     const param = pageQueryParams[filter];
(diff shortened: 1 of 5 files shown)


b10n1k (commented):
Why should it be one parameter `job_setting` and not two for passing key and searching value. This approach smells a bit despite the intelligent approach. I think it would be easier to understand and maintain a simplier solution. We really have to write something like `job_setting=_TEST_ISSUES=(^|,)(33135)(,|$)` in the URL bar. this can go wrong in many ways. 
btw it was cool PoC and clean PR IMO

Martchus:
I though about using different parameters for key and value (as this was also suggested in the ticket). However, I found this even more complicated. The problem is that one can always specify an arbitrary number of any kind of parameter. By having just one parameter the covers key and value at the same time this is easy to handle. With separate parameters you need to handle cases like an unequal number of keys and values being specified which I found more complicated.

I must also say that having to specify regexes here is not nice. That's not how I as a user would like to use this. However, we explicitly wanted this kind of "stupid but simple" approach in the ticket.

I'm not sure how we should proceed with this. I noted down [my thoughts on progress](https://progress.opensuse.org/issues/156553#note-23).

okurz:
Have you considered using the /job_settings API or at least the internal implementation which considers performance? See #4888

Martchus:
It makes no sense to use an external facing API when implementing another external facing API. I also don't think I can re-use any existing implementation details because there is already code to do the queries on the "All tests" and "Test results overview" pages doing a rather complex SQL query. I cannot just combine this with a different piece of code doing a different SQL query. I could try to limit the search depth for job settings like the other route does (when I remember correctly) but this would of course be yet another trade-off making this feature probably way less useful.

PR #6449, opened 2025-05-13
Avoid workers from getting stuck with old jobs

Martchus:
* Continue marking the current job as incomplete if the job duplication
  fails (job duplication is not essential but complicated enough to
  potentially fail)
* Continue setting the result of a job if carrying over bugrefs fails (as
  carrying over bugrefs relies on parsing JSON data from the job and thus
  might fail but we can nevertheless set the job result)
* Continue with marking the worker as free if the job cannot be set to
  incomplete (as setting a worker free is more important and those jobs end
  up canceled at some point in practice anyway)
* See https://progress.opensuse.org/issues/182303

---

Still a draft because I'll have to extend tests for all these error cases. This shouldn't be hard to do with some mocking.


kalikiana:
Just a minor point, but I'd suggest either `Avoid workers getting stuck` or `Prevent workers from getting stuck`

asdil12 (approved):

Martchus:
All recent Codecov webhooks returned `{"error":"Permission denied"}` right now. That's why the checks are missing. I was also not able to login on the Codecov webpage before so this is probably generally broken right now.

realcharmer (approved):

okurz:
> All recent Codecov webhooks returned `{"error":"Permission denied"}` right now. That's why the checks are missing. I was also not able to login on the Codecov webpage before so this is probably generally broken right now.

Another PR just received their codecov report. Maybe we just need to be patient

Martchus:
Recent deliveries of the webhook still look bad. So those are probably a red herring then.

PR #6450, opened 2025-05-14
Use encrypted session cookies when available

okurz:


lib/OpenQA/Setup.pm
```diff
@@ -396,7 +396,7 @@ sub load_plugins ($server, $monitoring_root_route = undef, %options) {
 }
 
 sub set_secure_flag_on_cookies ($c) {
-    $c->app->sessions->secure(1) if $c->req->is_secure;
+    $c->app->sessions->encrypted(1) if $c->req->is_secure;
```
kraih, line 399: These two are not related, `secure` makes the sessions cookie require HTTPS and the `is_secure` check ensures HTTPS was used for the request.

kraih:
The feature was added in Mojolicious 9.39 from last year.

Martchus:
> The feature was added in Mojolicious 9.39 from last year.

So I guess we either make this conditional (using `if …->can('encrypted')`) or we bump the minimum version (from 9.34 to 3.39).

lib/OpenQA/Setup.pm
okurz: Changed to `$app->sessions->encrypted(1) if $app->sessions->can('encrypted');`

okurz:
> > The feature was added in Mojolicious 9.39 from last year.
> 
> So I guess we either make this conditional (using `if …->can('encrypted')`) or we bump the minimum version (from 9.34 to 3.39).

Done the conditional

lib/OpenQA/LiveHandler.pm
```diff
@@ -18,6 +18,7 @@ has [qw(cmd_srv_transactions_by_job devel_java_script_transactions_by_job status
 
 # This method will run once at server start
 sub startup ($self) {
+    $self->moniker('open_q_a-web_a_p_i');
```
Martchus: Is this change needed? I assume you added it for consistency with the main web UI. It would probably make sense to use this in both places (instead of a very weirdly spelled version of "openQA-WebAPI"):

```suggestion
    $self->moniker('openqa_webapi');
```

Or just:

```suggestion
    $self->moniker('openqa');
```

For consistency we should probably set this up in all Mojolicious services.

Martchus (changes requested):

lib/OpenQA/LiveHandler.pm
okurz: > Is this change needed? I assume you added it for consistency with the main web UI.

Not just for consistency but to actually fix existing live-handler tests we run.

> It would probably make sense to use this in both places (instead of a very weirdly spelled version of "openQA-WebAPI"):
> 
> Or just:
> 
> For consistency we should probably set this up in all Mojolicious services.

Yes, that can be done. Done

Martchus (approved):

lib/OpenQA/LiveHandler.pm
Martchus: I approved. I see that you now used different names in the different services (except for the main and livehandler services) but I guess that's ok.

lib/OpenQA/Setup.pm
Wabri: Since we require https://metacpan.org/pod/CryptX can we simply remove the if and set the session encrypted by default?

perlpunk (approved):

PR #6451, opened 2025-05-14
Add API route to show all not-ok jobs

d3flex:
The PR is not ready, but I had to share where I am heading. I was just experiment so far. 

What you will find currently:
- an API endpoint and a smoke test. The name must change, but I couldnt think anything for now. 
- modified the sql query, in order to _join_ jobs to filter for the result, without change/break the way it work at the moment 
- adjustment in the test because I added another Job in fixtures. I should have been able to avoid to have some of the test impacted but I didnt manage to do so. need to improve)

So later I am planning the following changes:
- the endpoint reuse the `/job_settings/jobs` passing an optional parameter for the result value (failed, incomplete, etc). I think the end target is that the endpoint should return this by default.

https://progress.opensuse.org/issues/156547

### DIFF
--- a/lib/OpenQA/Schema/ResultSet/JobSettings.pm
+++ b/lib/OpenQA/Schema/ResultSet/JobSettings.pm
@@ -18,6 +18,7 @@ sub jobs_for_setting ($self, $options) {
     $key_like =~ s/\*/\%/g;
     my $list_value = $options->{list_value};
     my $list_value_like = "%${list_value}%";
+    my $list_result = $options->{list_result};
 
     # Get the highest job id to limit the number of jobs that need to be considered (to improve performance)
     my $dbh = $self->result_source->schema->storage->dbh;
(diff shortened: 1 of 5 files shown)


lib/OpenQA/Schema/ResultSet/JobSettings.pm
```diff
@@ -28,11 +29,28 @@ sub jobs_for_setting ($self, $options) {
...
-        'SELECT job_id, value FROM job_settings WHERE key LIKE ? AND value LIKE ? AND job_id > ? ORDER BY job_id DESC');
-    $sth->execute($key_like, $list_value_like, $min_id);
+    my $where_queries = 'WHERE ' . join(' AND ', @where_queries);
+    my $sql = "SELECT js.job_id, js.value
+               FROM job_settings js
+               $join_part
+               $where_queries
+               ORDER BY js.job_id DESC";
+    $sth = $dbh->prepare($sql);
+    $sth->execute(@where_values);
```
Martchus, line 50: It looks like we only have two cases here (`$list_result` is defined or not). So maybe it would be clearer to just spell out the entire SQL query in one piece for each case (despite the repetition). It would be easier to read and less likely contain any chances for SQL injections.

perlpunk: But I'm wondering why this is using raw SQL instead of DBIx::Class features? Is there anything in the query not supported by DBIC?

Martchus: When trying optimizations to search in a comma-separated list (the real optimization would of course be to use a proper database schema) I noticed that DBIx gets more in the way than it helps. So maybe this question is not leading in a useful direction.

But to answer the question: I would say DBIx is also not making things really better here, too. Otherwise the question probably goes to @kraih who wrote this query.

perlpunk: Please note that the common abbreviation for DBIx::Class is "DBIC". "DBIx" is a namespace that contains many other DBI related modules that have nothing to do with ORMs.

kraih: Afraid i don't remember anymore why this is a raw query. But there probably was a reason. Maybe DBIx::Class generated a less optimal query? I do remember performance being a big concern here.

d3flex:
not relevant anymore due to work on other PRs

PR #6455, opened 2025-05-14
load-templates: with --clean, empty job group YAML templates

AdamWill:
When running load-templates --clean, for any job group with a non-empty YAML template string, we should write an empty one (one that defines no scenarios or products). This is most in line with the intended behavior of --clean - to wipe all existing job configuration before loading the data.

We should also clean JobGroups before everything else, otherwise we will often choke on "must be updated through the YAML template" when trying to delete something that openQA realizes is referenced in an existing YAML template string.

This is different from existing behavior if job groups exist that are not included in the data being loaded. Previously we would leave those alone (and probably choke on the error, if they defined any templates). Now, we will 'empty' those groups, and load the data successfully.


AdamWill:
test failure is a timeout that doesn't look related to the PR.

Martchus (approved):

okurz (approved):

PR #6466, opened 2025-05-19
Fix quoting issue in run-tests-within-container script

b10n1k:
the script expands the ${PROVE_ARGS} to an empty string after --lib which breaks the prove execution.

### DIFF
--- a/tools/run-tests-within-container
+++ b/tools/run-tests-within-container
@@ -36,7 +36,8 @@ run_fullstack_test() {
     prepare_fullstack_test
 
     export PERL5OPT="$PERL5OPT $HARNESS_PERL_SWITCHES"
-    if ! prove -l "${PROVE_ARGS}" "$@"; then
+    # shellcheck disable=SC2086
+    if ! prove -l ${PROVE_ARGS} "$@"; then
         TESTS_FAILED_FLAG=1
         find '/tmp' -path '*/t/full-stack.d/openqa/testresults/*/autoinst-log.txt' \
             -exec echo 'contents of' {} \; -exec cat {} \; 2> /dev/null > '/opt/openqa/autoinst-log.txt'


perlpunk (approved):

baierjan (commented):
I guess this is the same as on https://github.com/os-autoinst/openQA/pull/6466/files#diff-e494b75dd148815614c47e2c93d7a10accc462dd143b3a0f90e019db89ea2e6aR10-R11

tools/run-tests-within-container
```diff
@@ -36,7 +36,7 @@ run_fullstack_test() {
     prepare_fullstack_test
 
     export PERL5OPT="$PERL5OPT $HARNESS_PERL_SWITCHES"
-    if ! prove -l "${PROVE_ARGS}" "$@"; then
+    if ! prove -l ${PROVE_ARGS:+} "$@"; then
```
perlpunk, line 39: please make sure line 11 and this line are doing the same thing.

perlpunk (changes requested):
see comment

tools/run-tests-within-container
```diff
@@ -8,7 +8,7 @@ TESTS_FAILED_FLAG=0
 run_test() {
     # Allow command line options without quotes
     # shellcheck disable=SC2086
-    prove -l ${PROVE_ARGS} "$@" || TESTS_FAILED_FLAG=1
+    prove -l ${PROVE_ARGS:+} "$@" || TESTS_FAILED_FLAG=1
```
perlpunk, line 11: What about the line above now?

b10n1k:
> Have you verified that this is working when `PROVE_ARGS` actually has a value?

No I didnt checked that. I updated because it didnt seem to work, which in my understanding of the expansion seems weird not to

tools/run-tests-within-container
```diff
@@ -8,7 +8,7 @@ TESTS_FAILED_FLAG=0
 run_test() {
     # Allow command line options without quotes
     # shellcheck disable=SC2086
-    prove -l ${PROVE_ARGS} "$@" || TESTS_FAILED_FLAG=1
+    prove -l ${PROVE_ARGS:+$PROVE_ARGS} "$@" || TESTS_FAILED_FLAG=1
```
Martchus, line 11: What is the point of the `+$PROVE_ARGS` part?

b10n1k: it ensures that expands to nothing at all without quotes. for some reason the `${PROVE_ARGS:+}` when `$PROVE_ARGS=foo` still evaluates to nothing. 

What I am looking for is
- PROVE_ARGS="-v" make run-tests-within-container expands to ` prove -l -v -r`
- make run-tests-within-container expands to ` prove -l -r`
- PROVE_ARGS="" make run-tests-within-container expands to ` prove -l -r`
- and in no case expansion to `prove -l ' ' -r`


I used `${PROVE_ARGS:-}` also before but this complains about `SC2086 (info): Double quote to prevent globbing and word splitting.` 

how would you do it?

tools/run-tests-within-container
```diff
@@ -36,7 +36,7 @@ run_fullstack_test() {
     prepare_fullstack_test
 
     export PERL5OPT="$PERL5OPT $HARNESS_PERL_SWITCHES"
-    if ! prove -l "${PROVE_ARGS}" "$@"; then
+    if ! prove -l ${PROVE_ARGS:+$PROVE_ARGS} "$@"; then
```
perlpunk, line 39: My suggestion would be:
```diff
diff --git a/tools/run-tests-within-container b/tools/run-tests-within-container
index a15a1b123..bff35680b 100755
--- a/tools/run-tests-within-container
+++ b/tools/run-tests-within-container
@@ -36,7 +36,8 @@ run_fullstack_test() {
     prepare_fullstack_test
 
     export PERL5OPT="$PERL5OPT $HARNESS_PERL_SWITCHES"
-    if ! prove -l "${PROVE_ARGS}" "$@"; then
+    # shellcheck disable=SC2086
+    if ! prove -l ${PROVE_ARGS} "$@"; then
         TESTS_FAILED_FLAG=1
         find '/tmp' -path '*/t/full-stack.d/openqa/testresults/*/autoinst-log.txt' \
             -exec echo 'contents of' {} \; -exec cat {} \; 2> /dev/null > '/opt/openqa/autoinst-log.txt'
```

tools/run-tests-within-container
perlpunk: Just look one line above!
I even mentioned this line yesterday.
You just need to copy this line to the one you changed.

tools/run-tests-within-container
perlpunk: Here I asked you about the line above, the `# shellcheck disable=SC2086`.

tools/run-tests-within-container
perlpunk: `# shellcheck disable=SC2086`

tools/run-tests-within-container
b10n1k: as line above I thought you meant the other `prove` command. checking again

perlpunk (approved):

Wabri (approved):

PR #6468, opened 2025-05-20
Deprecate skip-checks and add check-repos option

asmorodskyi:
--skip-checks option was disabling opt out checks of maintenance updates repositories.
    This has several flaws:
    1. It is questionable how much common situation from which check is trying to protect,
    so it make sense to invert logic to opt in allowing user to decide whether they need it
    or not
    2. Only check currently covered by this option is SUSE specific and it is another reason
    to switch to opt in
    3. skip-checks is confusing name because in a fact there is one and only check of maintenance
    repositories
    To address this problems two changes were done:
    - invert logic from opt out to opt in
    - deprecate skip-checks option because it is not needed anymore ( not to do check
      from now on is default behavior
    - introduce new command line option check-repos which has more accurate
      description and trigger check on user demand

### DIFF
--- a/lib/OpenQA/Script/CloneJob.pm
+++ b/lib/OpenQA/Script/CloneJob.pm
@@ -327,7 +327,7 @@ sub clone_job ($jobid, $url_handler, $options, $post_params = {}, $jobs = {}, $d
     if (my $group_id = $job->{group_id}) { $settings->{_GROUP_ID} = $group_id }
     clone_job_apply_settings($options->{args}, $relation eq 'children' ? 0 : $depth, $settings, $options);
     OpenQA::Script::CloneJobSUSE::detect_maintenance_update($jobid, $url_handler, $settings)
-      unless $options->{'skip-checks'};
+      if $options->{'check-repos'};
     clone_job_download_assets($jobid, $job, $url_handler, $options) unless $options->{'skip-download'};
 }
 
--- a/script/openqa-clone-job
+++ b/script/openqa-clone-job
@@ -112,7 +112,12 @@ $OPENQA_SHAREDIR/factory).
 
 =item B<--skip-checks>
 
-Skips additional checks like maintenance update availability.
+Deprecated option. The logic was inverted, so currently the check is skipped by default.
+If you were using it, just remove it and the behavior will remain the same.
+
+=item B<--check-repos>
+
+Enables additional checks for availability of maintenance update repositories (SUSE specific check).
 
 =item B<--skip-deps>
 
@@ -215,10 +220,13 @@ sub parse_options () {
         "skip-deps", "skip-chained-deps", "skip-download", "parental-inheritance",
         "help|h", "show-progress", "within-instance|w=s", "clone-children",
         "max-depth:i", "repeat|r:i", "ignore-missing-assets", "export-command",
-        "skip-checks",
+        "skip-checks", "check-repos",
     ) or usage(1);
     usage(0) if $options{help};
     usage(1) if $options{help} || ($options{'within-instance'} && $options{from});
+    warn
+"--skip-checks is deprecated and will be dropped soon. Just remove it because checks are not executed by default anymore"
+      if $options{'skip-checks'};
     if ($options{'within-instance'}) {
         ($options{'within-instance'}, $jobid) = split_jobid($options{'within-instance'});
         $options{'skip-download'} = 1;


script/openqa-clone-job
```diff
@@ -110,9 +110,9 @@ from JOBREF if it is a URL).
 Specifies the directory to store test assets (defaults to
 $OPENQA_SHAREDIR/factory).
 
-=item B<--skip-checks>
+=item B<--check-repos>
 
-Skips additional checks like maintenance update availability.
+Additionally check maintenance update repositories availability(SUSE specific check).
 
 =item B<--skip-deps>
```
b10n1k, line 117: missed

asmorodskyi: I don't understand . what does it mean ? what exactly missed ?

script/openqa-clone-job
```diff
@@ -110,9 +110,9 @@ from JOBREF if it is a URL).
 Specifies the directory to store test assets (defaults to
 $OPENQA_SHAREDIR/factory).
 
-=item B<--skip-checks>
+=item B<--check-repos>
 
-Skips additional checks like maintenance update availability.
+Additionally check maintenance update repositories availability(SUSE specific check).
```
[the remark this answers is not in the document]
asmorodskyi: LGTM ( if someone else will confirm )

script/openqa-clone-job
b10n1k: ignore. my mistake

script/openqa-clone-job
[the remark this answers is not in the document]
okurz: Please stay consistent with other options which use present tense describing what the option does not imperative mood asking to do something

lib/OpenQA/Script/CloneJob.pm
```diff
@@ -327,7 +327,7 @@ sub clone_job ($jobid, $url_handler, $options, $post_params = {}, $jobs = {}, $d
     if (my $group_id = $job->{group_id}) { $settings->{_GROUP_ID} = $group_id }
     clone_job_apply_settings($options->{args}, $relation eq 'children' ? 0 : $depth, $settings, $options);
     OpenQA::Script::CloneJobSUSE::detect_maintenance_update($jobid, $url_handler, $settings)
-      unless $options->{'skip-checks'};
+      if $options->{'check-repos'};
```
okurz, line 330: @grisu48 you first introduced this in e0ebe7648. I assume you wanted to introduce a generic option to skip *any* checks including any potential future additions? In that case we could keep the old option and only introduce a new option `check-repos`. The old option would then of course not do anything useful so I guess @asmorodskyi's change is fine but I would just like to give you the opportunity to intervene

okurz (changes requested):

Martchus (commented):
This of course breaks compatibility removing an existing option. I'd be fine with it even though it probably breaks a few scripts.

script/openqa-clone-job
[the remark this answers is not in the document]
asmorodskyi: `Additional check of maintenance update repositories availability(SUSE specific check)`  will work ?

lib/OpenQA/Script/CloneJob.pm
asmorodskyi: Like I mention in commit message having option `--skip-checks`  which is doing single check is confusing for **current** users . When I read parameter name and its description I really like to know what it is doing **currently** not that someone will potentially implement in it in 5 years

kalikiana: I suggest keeping `--skip-checks` for compatibility but with a deprecation message.

The user story is not fully clear to me, though, so I'm a little worried about introducing a new option for a feature that isn't widely used (anymore)... do we actually want to keep it? Can we please confirm the use case before adding more new options here?

kalikiana (changes requested):

lib/OpenQA/Script/CloneJob.pm
asmorodskyi: well my opinion that you are OVERestimating that . We are not breaking kernel API or glibc . It is script which is in reality used in 5-10 different places in non-interactive way . And half of this places are under control of people who currently reviewing this PR . Also whoever will be affected by this break will fix it in 5 minutes :man_shrugging:  So even effort of this discussion around this deprecation is bigger then actual impact.
But let's stop here I would just do whatever you will tell me

script/openqa-clone-job
```diff
@@ -215,10 +220,11 @@ sub parse_options () {
         "skip-deps", "skip-chained-deps", "skip-download", "parental-inheritance",
         "help|h", "show-progress", "within-instance|w=s", "clone-children",
         "max-depth:i", "repeat|r:i", "ignore-missing-assets", "export-command",
-        "skip-checks",
+        "skip-checks","check-repos",
     ) or usage(1);
     usage(0) if $options{help};
     usage(1) if $options{help} || ($options{'within-instance'} && $options{from});
+    print STDERR "--skip-checks is deprecated and will be dropped soon. Just remove it because checks not executed by default anymore" if $options{'skip-checks'};
```
okurz, line 227: this probably needs tidying. Please call `make tidy-perl`

okurz (changes requested):

script/openqa-clone-job
Martchus, line 227:
```suggestion
    print STDERR "--skip-checks is deprecated and will be dropped soon. Just remove it because checks are not executed by default anymore" if $options{'skip-checks'};
```

script/openqa-clone-job
```diff
@@ -215,10 +220,13 @@ sub parse_options () {
         "skip-deps", "skip-chained-deps", "skip-download", "parental-inheritance",
         "help|h", "show-progress", "within-instance|w=s", "clone-children",
         "max-depth:i", "repeat|r:i", "ignore-missing-assets", "export-command",
-        "skip-checks",
+        "skip-checks", "check-repos",
     ) or usage(1);
     usage(0) if $options{help};
     usage(1) if $options{help} || ($options{'within-instance'} && $options{from});
+    print STDERR
+"--skip-checks is deprecated and will be dropped soon. Just remove it because checks are not executed by default anymore"
```
perlpunk, line 228: perl has a `warn` function for that:
```suggestion
    warn
"--skip-checks is deprecated and will be dropped soon. Just remove it because checks are not executed by default anymore\n"
```

perlpunk (changes requested):
see comment

lib/OpenQA/Script/CloneJob.pm
perlpunk: It might only take 5 minutes for the actual fix, but it might break pipelines and things are blocked until someone can actually do the fix.
When I recently did the perltidy change, there was a big complaint that this was blocking osado PRs, although the fix was just running perltidy.
If you are in the position of the person whose pipelines breaks / is blocked you might think differently.

asmorodskyi: just to clarify : all the noise around perltidy was about the setup which lead to flow where change in one project may break another one and not about change itself .
About estimation how much disruptive this change may be I just share my view on that . Didn't want to convince you . Anyway you are "gatekeepers"  of this project so I will just follow whatever you saying

okurz (approved):

script/openqa-clone-job
```diff
@@ -112,7 +112,12 @@ $OPENQA_SHAREDIR/factory).
 
 =item B<--skip-checks>
 
-Skips additional checks like maintenance update availability.
+Deptrecated option. Logic was inverted so currently check skipped by default. So if you
```
Wabri, line 115: @okurz I'm not sure if we have a ticket or something that keep track of this deprecation.

Wabri (changes requested):
I think the title of the pr and the commit description should be change!

script/openqa-clone-job
kalikiana: We [do now](https://progress.opensuse.org/issues/182834).

asmorodskyi:
> I think the title of the pr and the commit description should be change!

updated PR label and description AND git commit  label and description

script/openqa-clone-job
```diff
@@ -112,7 +112,12 @@ $OPENQA_SHAREDIR/factory).
...
+were using it just remove it and behavior will remain the same.
```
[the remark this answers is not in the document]
asmorodskyi: done

lib/OpenQA/Script/CloneJob.pm
kalikiana: > do we actually want to keep it? Can we please confirm the use case before adding more new options here?

Note I filed [poo#182834 regarding follow-up work](https://progress.opensuse.org/issues/182834), so no strong objection to introducing `--check-repos` here.

kalikiana (approved):

perlpunk (approved):

perlpunk:
check-helm-chart failed
https://github.com/os-autoinst/openQA/actions/runs/15164432729/job/42638410230?pr=6468
Can't really see why. I'll retrigger

PR #6471, opened 2025-05-21
Stop flagging obs_rsync errors as failures

kalikiana:
Logs remain available, but are not considered actionable in most cases.

See: https://progress.opensuse.org/issues/180962

### DIFF
--- a/lib/OpenQA/WebAPI/Plugin/ObsRsync/Task.pm
+++ b/lib/OpenQA/WebAPI/Plugin/ObsRsync/Task.pm
@@ -76,7 +76,7 @@ sub run {
     $error =~ s/\s+$//;
     $app->log->error('ObsRsync#_run failed (' . $exit_code . '): ' . $error);
     $helper->log_failure($project, $job->id);
-    return $job->fail({code => $exit_code, message => $error});
+    return $job->finish({code => $exit_code, message => $error});
 }
 
 sub update_dirty_status {
@@ -123,7 +123,7 @@ sub update_obs_builds_text {
 
     my ($exit_code, $error) = $helper->for_every_batch($alias, $sub);
 
-    return $job->fail({code => $exit_code, message => $error}) if $exit_code;
+    return $job->finish({code => $exit_code, message => $error}) if $exit_code;
     return $job->finish('Success');
 }
 
--- a/t/api/14-plugin_obs_rsync_async.t
+++ b/t/api/14-plugin_obs_rsync_async.t
@@ -148,38 +148,14 @@ $t->put_ok('/api/v1/obs_rsync/MockProjectError/runs')->status_is(201, 'Start ano
 
 sleep_until_all_jobs_finished($t);
 
-# MockProjectError will fail so number of finished jobs should remain, but one job must be failed
-is(_jobs_cnt('finished'), 10, 'Number of finished jobs');
+# MockProjectError should not be raised because errors are ignored
+is(_jobs_cnt('finished'), 11, 'Number of finished jobs');
 my ($cnt, $jobs) = _jobs('failed');
-is($cnt, 1, 'Number of finished jobs');
-is($jobs->[0]->{result}->{message}, 'Mock Error', 'Correct error message') if $cnt;
+is($cnt, 0, 'Number of failed jobs');
 
-subtest 'test max retry count' => sub {
-    # use all concurrency slots to reach concurrency limit
-    my @guards = map { $t->app->obs_rsync->concurrency_guard() } (1 .. $config{queue_limit});
-    # put request and make sure it succeeded within 5 sec
-    $t->put_ok('/api/v1/obs_rsync/Proj1/runs')->status_is(201, 'trigger rsync');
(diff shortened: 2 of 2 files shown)


okurz:
(obviously) the test is expecting failed jobs. Please adjust the test expectations accordingly

t/data/openqa-trigger-from-obs/script/rsync.sh
```diff
@@ -13,6 +13,6 @@ function sleep_until_file_created {
 [[ "$1" =~ MockProjectLongProcessing* ]] && { sleep_until_file_created "$1" || exit 1; }
 [ "$1" != MockProjectError ] || {
     echo >&2 "Mock Error"
-    exit 1
+    exit 777
```
kalikiana, line 16: Exit code never seems to be propagated? I tried TDD as usual to confirm the behavior, but this only helped me understand the retries. I don't know how errors would be reported. [Docs](https://docs.mojolicious.org/Minion/Job#retry) provide no clues either.

lib/OpenQA/WebAPI/Plugin/ObsRsync/Task.pm
perlpunk, line 1: I think the change for `rsync.sh` should be done in the `run` method. This is what I came up with after a quick glance:
```diff
diff --git a/lib/OpenQA/WebAPI/Plugin/ObsRsync/Task.pm b/lib/OpenQA/WebAPI/Plugin/ObsRsync/Task.pm
index 5417e0abb..fc0849953 100644
--- a/lib/OpenQA/WebAPI/Plugin/ObsRsync/Task.pm
+++ b/lib/OpenQA/WebAPI/Plugin/ObsRsync/Task.pm
@@ -76,7 +76,7 @@ sub run {
     $error =~ s/\s+$//;
     $app->log->error('ObsRsync#_run failed (' . $exit_code . '): ' . $error);
     $helper->log_failure($project, $job->id);
-    return $job->fail({code => $exit_code, message => $error});
+    return $job->finish({code => $exit_code, message => $error});
 }
 
 sub update_dirty_status {
diff --git a/t/api/14-plugin_obs_rsync_async.t b/t/api/14-plugin_obs_rsync_async.t
index 3822f236e..ebbf80583 100644
--- a/t/api/14-plugin_obs_rsync_async.t
+++ b/t/api/14-plugin_obs_rsync_async.t
@@ -148,12 +148,18 @@ $t->put_ok('/api/v1/obs_rsync/MockProjectError/runs')->status_is(201, 'Start ano
 
 sleep_until_all_jobs_finished($t);
 
-# MockProjectError will fail so number of finished jobs should remain, but one job must be failed
-is(_jobs_cnt('finished'), 10, 'Number of finished jobs');
+# MockProjectError should finish because we want to ignore failures
+is(_jobs_cnt('finished'), 11, 'Number of finished jobs');
 my ($cnt, $jobs) = _jobs('failed');
+is($cnt, 0, 'Number of failed jobs');
+
+($cnt, $jobs) = _jobs('finished');
 is($cnt, 1, 'Number of finished jobs');
-is($jobs->[0]->{result}->{message}, 'Mock Error', 'Correct error message') if $cnt;
+is($jobs->[0]->{result}->{message}, 'Mock Error', 'Correct error message');
+is($jobs->[0]->{result}->{code}, 256, 'Correct exit code');
```
But the other tasks should also be handled, not only `obs_rsync_run`.

t/data/openqa-trigger-from-obs/script/rsync.sh
perlpunk: exit codes should be between 0 and 255

t/api/14-plugin_obs_rsync_async.t
```diff
@@ -153,6 +153,7 @@ is(_jobs_cnt('finished'), 10, 'Number of finished jobs');
 my ($cnt, $jobs) = _jobs('failed');
 is($cnt, 1, 'Number of finished jobs');
 is($jobs->[0]->{result}->{message}, 'Mock Error', 'Correct error message') if $cnt;
+is($jobs->[0]->{result}->{code}, 777, 'Correct return code') if $cnt;
```
perlpunk, line 156: Note that here you would get `$exit_code << 8`

lib/OpenQA/WebAPI/Plugin/ObsRsync/Task.pm
kalikiana: Seems like the retry test no longer makes sense now. Because the cases no longer fail - retries still apply  for dirty and lock checks, but not sure if that makes sense to test here, or maybe move it to t/ui/27-plugin_obs_rsync_obs_status.t

kalikiana: > But the other tasks should also be handled, not only obs_rsync_run.

This branch covers **update_builds_text**. *update_dirty_status* never fails afair.

perlpunk (approved):

aylen384 (approved):

PR #6474, opened 2025-05-22
systemd: Increase availability of openqa-webui with ordering

okurz:
By starting openqa-webui before scheduler+livehandler which are less
time-critical we can increase the availability of openqa-webui which will be
effective on both startup of systems as well as during shutdown when the
openqa-webui is kept around longer after scheduler+livehandler have been
stopped first.

Verified manually on openqa.suse.de

Related progress issue: https://progress.opensuse.org/issues/181370

### DIFF
--- a/systemd/openqa-webui.service
+++ b/systemd/openqa-webui.service
@@ -2,7 +2,7 @@
 Description=The openQA web UI
 Wants=openqa-setup-db.service openqa-livehandler.service openqa-websockets.service openqa-gru.service openqa-enqueue-asset-cleanup.timer openqa-enqueue-result-cleanup.timer openqa-enqueue-bug-cleanup.timer openqa-minion-restart.path
 After=postgresql.service openqa-setup-db.service nss-lookup.target remote-fs.target apache2.service nginx.service caddy.service
-Before=openqa-gru.service
+Before=openqa-gru.service openqa-scheduler.service openqa-livehandler.service
 
 [Service]
 User=geekotest
--- a/systemd/openqa-websockets.service
+++ b/systemd/openqa-websockets.service
@@ -2,7 +2,7 @@
 Description=The openQA WebSockets server
 Wants=openqa-setup-db.service
 Before=openqa-webui.service
-After=openqa-scheduler.service postgresql.service openqa-setup-db.service network.target nss-lookup.target remote-fs.target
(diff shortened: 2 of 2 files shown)


systemd/openqa-webui.service
```diff
@@ -2,7 +2,7 @@
 Description=The openQA web UI
 Wants=openqa-setup-db.service openqa-livehandler.service openqa-websockets.service openqa-gru.service openqa-enqueue-asset-cleanup.timer openqa-enqueue-result-cleanup.timer openqa-enqueue-bug-cleanup.timer openqa-minion-restart.path
 After=postgresql.service openqa-setup-db.service nss-lookup.target remote-fs.target apache2.service nginx.service caddy.service
-Before=openqa-gru.service
+Before=openqa-gru.service openqa-scheduler.service openqa-livehandler.service
```
Martchus, line 5: I suppose we could leave the livehandler out of this as it is actually used to serve certain web UI routes and only for technical reasons separated. I also wouldn't expect the livehandler to take long to terminate. However, in practice it won't matter.

okurz: I don't understand your comment. My thinking is that the openQA webUI is more time critical than both scheduler as well as livehandler. By ordering we ensure that more system resources are available to bring up the webUI before even trying to bring up the livehandler as well as on shutdown we can bring down the livehandler and do necessary cleanup which again might take some system resources so that we keep the webUI around for longer for users or connected external services. Also we workaround any stalling of the livehandler to impact the openQA webUI as observed in https://progress.opensuse.org/issues/181370#note-18 . So in my observation this is really helpful.

nicksinger (approved):

systemd/openqa-webui.service
Martchus: Ok, so if the livehandler takes some time to shutdown (which I thought wouldn't be the case) this change makes sense - although it is nevertheless part of the web UI and shutting it down before that means the live view will not be available.

Martchus (approved):

systemd/openqa-webui.service
okurz: Yes, that's correct. But would shutting down the webUI *before* the liveviewhandler make sense? In before we just didn't have any ordering and now we have explicit ordering which I think is an improvement anyway. In the best case all services startup/shutdown so swiftly that all doesn't matter :)

PR #6477, opened 2025-05-23
CI: Enable Leap 16.0 OBS build checks

okurz:
Related progress issue: https://progress.opensuse.org/issues/180737


perlpunk (approved):
I think this will only have an effect after being merged

b10n1k (approved):

okurz:
Ok, let's see. I would be surprised if the builds succeed.

PR #6478, opened 2025-05-23
Revert "CI: Enable Leap 16.0 OBS build checks" as build results are not yet stable

okurz:
Reverts os-autoinst/openQA#6477 as build results are not yet stable.
We need to fix build issues already apparent in https://build.opensuse.org/project/monitor/devel:openQA?arch_x86_64=1&defaults=0&repo_16_0=1&unresolvable=1 before re-introducing this.

### DIFF
--- a/.obs/workflows.yml
+++ b/.obs/workflows.yml
@@ -30,14 +30,6 @@ pr:
               - target_project: devel:openQA:Leap:15.6
                 target_repository: '15.6'
             architectures: [ x86_64 ]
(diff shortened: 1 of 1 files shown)


perlpunk:
Hm, it's surprising that it doesn't report the leap 16 results at all, although they are appearing in the [branched project](https://build.opensuse.org/package/show/devel:openQA:GitHub:os-autoinst:openQA:PR-6478/openQA)

PR #6479, opened 2025-05-23
CI: Enable Leap 16.0 OBS build checks

okurz:
Related progress issue: https://progress.opensuse.org/issues/180737


okurz:
We first need to fix build issues as observed in https://build.opensuse.org/project/monitor/devel:openQA?arch_x86_64=1&defaults=0&repo_16_0=1&unresolvable=1

PR #6481, opened 2025-05-27
add support for OIDC

cschneemann:
Add support for OIDC, based on the OAuth2 auth_method, but gets all information from the token and does not contact userinfo url.
Defining an access_group is mandatory for getting access.

### DIFF
--- a/lib/OpenQA/WebAPI/Auth/OIDC.pm
+++ b/lib/OpenQA/WebAPI/Auth/OIDC.pm
@@ -1,0 +1,85 @@
+# SPDX-License-Identifier: GPL-2.0-or-later
+
+package OpenQA::WebAPI::Auth::OIDC;
+use Mojo::Base -base, -signatures;
+
+use Carp 'croak';
+use Mojo::Util qw(dumper);
+use OpenQA::Log qw(log_debug);
+use MIME::Base64 qw(decode_base64url);
+use Mojo::JSON qw(decode_json);
+
+sub auth_setup ($server) {
+    my $app = $server->app;
+    my $config = $app->config->{oidc};
+
+    my $params = {
+        args => [
+            authorize_url => $config->{authorize_url},
+            token_url => $config->{token_url},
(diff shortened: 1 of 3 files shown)


lib/OpenQA/WebAPI/Auth/OIDC.pm
```diff
@@ -0,0 +1,83 @@
...
+sub auth_login ($controller) {
+    croak 'Config was not parsed' unless my $main_config = $controller->app->config->{oidc};
+    croak 'Setup was not called' unless my $provider_config = $main_config->{provider_config};
+
+    my $base_url = $controller->app->config->{global}->{base_url};
+    my $host = $base_url ? Mojo::URL->new($base_url)->host : $controller->req->url->host;
+    my $get_token_args = {redirect_uri => $controller->url_for('login')->userinfo(undef)->host($host)->to_abs,
+                          scope => "openid"};
+    $controller->oauth2->get_token_p(oidc => $get_token_args)
+      ->then(sub { update_user($controller, $main_config, $provider_config, shift) })
```
Martchus, line 78:
```suggestion
      ->then(sub ($data) { update_user($controller, $main_config, $provider_config, $data) })
```

Martchus (changes requested):
Looks like style checks and tests are failing. For the style checks, run `make tidy` and for the tests you probably only need to adjust `t/config.t`. Then the test coverage check will probably complain as your newly added file doesn't seem to have any test coverage at all. Perhaps you can create tests similar to how the other auth modules are tested (in e.g. `03-auth.t`).

We generally also don't have merge commits on feature branches. So please avoid the merge commit and rebase against master.

Otherwise I suppose the change looks good - I don't know much about oidc to be able to really tell, though.

Martchus (changes requested):
See my previous comment.

cschneemann:
rebased with latest master and had run `make tidy-perl`. Tests still missing..

PR #6483, opened 2025-05-27
Avoid job terminated unexpectedly by add signal handler

Wabri:
Related to: https://progress.opensuse.org/issues/108989

### DIFF
--- a/t/16-utils-runcmd.t
+++ b/t/16-utils-runcmd.t
@@ -30,6 +30,17 @@ my $schema = OpenQA::Test::Database->new->create(fixtures_glob => $fixtures);
 my $first_user = $schema->resultset('Users')->first;
 my $t = Test::Mojo->new('OpenQA::WebAPI');
 
+my $empty_tmp_dir = tempdir();
+my $fake_needle = {
+    job_id => 99926,
+    user_id => 99903,
+    needle_json => '{"area":[{"xpos":0,"ypos":0,"width":0,"height":0,"type":"match"}],"tags":["foo"]}',
+    needlename => 'foo',
+    needledir => $empty_tmp_dir,
+    imagedir => 't/data/openqa/share/tests/archlinux/needles',
+    imagename => 'test-rootneedle.png',
+};
+
 subtest 'run (arbitrary) command' => sub {
     ok(run_cmd_with_log([qw(echo Hallo Welt)]), 'run simple command');
     stdout_like { is(run_cmd_with_log([qw(false)]), '') } qr/[WARN].*[ERROR]/i;
@@ -248,33 +259,20 @@ subtest 'git commands with mocked run_cmd_with_log_return_error' => sub {
 subtest 'saving needle via Git' => sub {
     my $utils_mock = Test::MockModule->new('OpenQA::Git');
     $utils_mock->redefine(run_cmd_with_log_return_error => \&_run_cmd_mock);
-    {
-        package Test::FakeMinionJob;    # uncoverable statement
+    package Test::FakeMinionJob {
         sub finish { }
         sub fail {
             Test::Most::fail("Minion job shouldn't have failed.");    # uncoverable statement
             Test::Most::note(Data::Dumper::Dumper(\@_));    # uncoverable statement
         }
-    }
+    }    # uncoverable statement
 
     # configure use of Git
     $t->app->config->{global}->{scm} = 'git';
-    my $empty_tmp_dir = tempdir();
 
     # trigger saving needles like Minion would do
     @executed_commands = ();
-    OpenQA::Task::Needle::Save::_save_needle(
-        $t->app,
-        bless({} => 'Test::FakeMinionJob'),
-        {
-            job_id => 99926,
-            user_id => 99903,
-            needle_json => '{"area":[{"xpos":0,"ypos":0,"width":0,"height":0,"type":"match"}],"tags":["foo"]}',
-            needlename => 'foo',
-            needledir => $empty_tmp_dir,
-            imagedir => 't/data/openqa/share/tests/archlinux/needles',
-            imagename => 'test-rootneedle.png',
-        });
+    OpenQA::Task::Needle::Save::_save_needle($t->app, bless({} => 'Test::FakeMinionJob'), $fake_needle);
     is_deeply(
         \@executed_commands,
         [
@@ -311,21 +309,32 @@ subtest 'signal guard aborts when git is disabled and do_cleanup is "no"' => sub
 
     # trigger saving needles like Minion would do
     @executed_commands = ();
-    OpenQA::Task::Needle::Save::_save_needle(
-        $t->app,
-        bless({} => 'Test::FakeMinionJob'),
-        {
-            job_id => 99926,
-            user_id => 99903,
-            needle_json => '{"area":[{"xpos":0,"ypos":0,"width":0,"height":0,"type":"match"}],"tags":["foo"]}',
-            needlename => 'foo',
-            needledir => tempdir(),
-            imagedir => 't/data/openqa/share/tests/archlinux/needles',
-            imagename => 'test-rootneedle.png',
-        });
+    OpenQA::Task::Needle::Save::_save_needle($t->app, bless({} => 'Test::FakeMinionJob'), $fake_needle);
 
     isa_ok $signal_guard, 'My::FakeSignalGuard', 'signal guard has been created with the right class';
     ok $signal_guard->abort, 'signal guard is set to abort';
+};
+
+subtest 'save_needle returns and logs error when set_to_latest_master fails' => sub {
+    package Test::FailingMinionJob {
+        sub finish { }
+        sub fail($self, $args) { $self->{fail_message} = $args }
+    }    # uncoverable statement
+
+    my $git_mock = Test::MockModule->new('OpenQA::Git');
+    $t->app->config->{global}->{scm} = 'git';    # enable Git
+    $t->app->config->{do_cleanup} = 'yes';    # enable cleanup
+    $git_mock->redefine(set_to_latest_master => 'Unable to fetch from origin master: mocked error');
+
+    my @log_errors;
+    my $log_mock = Test::MockModule->new(ref $t->app->log);
+    $log_mock->redefine(error => sub ($self, $message) { push @log_errors, $message; });
+    my $job = bless({} => 'Test::FailingMinionJob');
+    OpenQA::Task::Needle::Save::_save_needle($t->app, $job, $fake_needle);
+
+    like $log_errors[0], qr/Unable to fetch.*mocked/, 'error logged on fail';
+    like $job->{fail_message}->{error}, qr/<strong>Failed to save.*<\/strong>.*<pre>Unable to fetch.*mocked.*<\/pre>/,
+      'error message in fail';
 };
 
 done_testing();


Martchus (changes requested):

lib/OpenQA/Task/Needle/Save.pm
```diff
@@ -102,6 +104,10 @@ sub _save_needle {
     # ensure needle dir is up-to-date
     my $git = OpenQA::Git->new({app => $app, dir => $needledir, user => $user});
     if ($git->enabled) {
+        if (!$git->is_workdir_clean && $git->config->{do_cleanup} eq 'no') {
+            $signal_guard->abort(1);
```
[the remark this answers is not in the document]
Wabri: I've move the signal abort setup and the log.

Wabri:
Since I've not change enough things to let the tests fail, I'm working to force tdd by let the test failing on `16-utils-runcmd.t` and than let them pass again.

Wabri:
@perlpunk definitely, 133 commit missing :rofl:

Wabri:
I left a lot of commented code, I'll cleanup tomorrow!

lib/OpenQA/Git.pm
```diff
@@ -81,7 +81,7 @@ sub set_to_latest_master ($self, $args = undef) {
         return $self->_format_git_error($res, 'Unable to reset repository to origin/master') unless $res->{status};
     }
 
-    return undef;
+    return;
```
okurz, line 84: why this change? It's good style to make a return always return a valid type explicitly hence the `return undef`

lib/OpenQA/Task/Needle/Save.pm
```diff
@@ -52,8 +53,12 @@ sub _format_git_error {
     return "<strong>Failed to save $name.</strong><br><pre>$error</pre>";
 }
 
-sub _save_needle {
-    my ($app, $minion_job, $args) = @_;
+sub _create_signal_guard ($job) {
+    return OpenQA::Task::SignalGuard->new($job);
+}
```
okurz, line 58:
```suggestion
sub _create_signal_guard ($job) { OpenQA::Task::SignalGuard->new($job) }
```

but I actually I suggest to not define this function as just using the object creation in place should be clear enough. You can still mock the module OpenQA::Task::SignalGuard for testing purposes

t/16-utils-runcmd.t
```diff
@@ -296,4 +297,79 @@ subtest 'saving needle via Git' => sub {
...
 };
 
+subtest 'signal guard aborts when git is disabled and do_cleanup is no' => sub {
+    my $save_mock = Test::MockModule->new('OpenQA::Task::Needle::Save');
+    my $signal_guard = OpenQA::Task::SignalGuard->new(bless({} => 'Test::FakeMinionJob'));
+    $save_mock->redefine(_create_signal_guard => sub { return $signal_guard; });
+
+    my $git_mock = Test::MockModule->new('OpenQA::Git');
+    $git_mock->redefine(enabled => sub { 0; });
+    $git_mock->redefine(config => sub { {do_cleanup => 'no'} });
```
okurz, line 307:
```suggestion
    $git_mock->redefine(enabled => 0);
    $git_mock->redefine(config => {do_cleanup => 'no'});
```

t/16-utils-runcmd.t
```diff
@@ -296,4 +297,79 @@ subtest 'saving needle via Git' => sub {
...
+subtest 'save_needle return log error when set_to_latest_master fail' => sub {
+    {
+        package Test::FailingMinionJob;    # uncoverable statement
+        sub finish { }
+        sub fail {
+            Test::Most::note("Minion job failed with: " . Data::Dumper::Dumper(\@_));    # uncoverable statement
+        }
+    }
+    my $git_mock = Test::MockModule->new('OpenQA::Git');
+    $git_mock->redefine(enabled => sub { 1; });
```
okurz, line 336: see above for this and following redefine methods

okurz (changes requested):

lib/OpenQA/Git.pm
Wabri: Coming from other programming languages, e.g. python, most of the time is preferred to use the return without explicitly a value. I'll update it with `return undef`.

lib/OpenQA/Task/Needle/Save.pm
```diff
@@ -101,10 +106,14 @@ sub _save_needle {
...
     my $git = OpenQA::Git->new({app => $app, dir => $needledir, user => $user});
+    # avoid retrying if cleanup via Git is not configured (then an admin might need to
+    # manually cleanup the repo first)
+    $signal_guard->abort(1) if !$git->enabled || $git->config->{do_cleanup} eq 'no';
     if ($git->enabled) {
         my $error = $git->set_to_latest_master;
         if ($error) {
-            $app->log->error($error);
+            $app->log->error(
+                "Dirty repo detected with do_cleanup disabled. A manual cleanup is required. Error: $error");
```
perlpunk, line 116: Is it known at this point that `do_cleanup` is disabled?

perlpunk (commented):
One thing about `set_to_latest_master` - it is only called if `update_branch` is set (and `do_cleanup` enabled), and `update_branch` is not set on o3 and is empty by default.
There is a ticket about that: https://progress.opensuse.org/issues/179314

lib/OpenQA/Task/Needle/Save.pm
Wabri: Without the object creation how can I check if abort is reset to 1? I don't have the object used in the save needle!

Wabri: I think I found how to do it!

t/16-utils-runcmd.t
```diff
@@ -296,4 +297,79 @@ subtest 'saving needle via Git' => sub {
     #       it is tested here explicitly.
 };
 
+subtest 'signal guard aborts when git is disabled and do_cleanup is no' => sub {
+    my $save_mock = Test::MockModule->new('OpenQA::Task::Needle::Save');
+    my $signal_guard = OpenQA::Task::SignalGuard->new(bless({} => 'Test::FakeMinionJob'));
+    $save_mock->redefine(_create_signal_guard => sub { return $signal_guard; });
+
+    my $git_mock = Test::MockModule->new('OpenQA::Git');
+    $git_mock->redefine(enabled => sub { 0; });
```
perlpunk, line 306: Please see line 151 ff. on how to change the config in tests.
This works but I find it a bit unusual.

t/16-utils-runcmd.t
```diff
@@ -296,4 +297,79 @@ subtest 'saving needle via Git' => sub {
     #       it is tested here explicitly.
 };
 
+subtest 'signal guard aborts when git is disabled and do_cleanup is no' => sub {
+    my $signal_guard;
+    my $signal_guard_mock = Test::MockModule->new('OpenQA::Task::SignalGuard');
+    $signal_guard_mock->redefine(
+        new => sub {
+            my ($class, @args) = @_;
+            $signal_guard = $class->Mojo::Base::new(@args);
```
perlpunk, line 306:
```suggestion
            $signal_guard = $signal_guard_mock->original('new')->($class, @args);
```

t/16-utils-runcmd.t
```diff
@@ -296,4 +297,79 @@ subtest 'saving needle via Git' => sub {
     #       it is tested here explicitly.
 };
 
+subtest 'signal guard aborts when git is disabled and do_cleanup is no' => sub {
+    my $signal_guard;
+    my $signal_guard_mock = Test::MockModule->new('OpenQA::Task::SignalGuard');
+    $signal_guard_mock->redefine(
+        new => sub {
+            my ($class, @args) = @_;
```
perlpunk, line 305:
```suggestion
        new => sub ($class, @args) {
```

t/16-utils-runcmd.t
```diff
@@ -296,4 +297,79 @@ subtest 'saving needle via Git' => sub {
...
+            imagedir => 't/data/openqa/share/tests/archlinux/needles',
+            imagename => 'test-rootneedle.png',
+        });
+
+    is($signal_guard->abort, 1, 'signal guard is not aborted initially');
+};
+
+subtest 'save_needle return log error when set_to_latest_master fail' => sub {
+    {
+        package Test::FailingMinionJob;    # uncoverable statement
```
perlpunk, line 334:
```suggestion
    package Test::FailingMinionJob {
```

lib/OpenQA/Task/Needle/Save.pm
Wabri: Since in the test that is covering this block `do_cleanup` is set to yes, I think we can confirm that do_cleanup is not disabled.

Wabri: I've changed the error message.

t/16-utils-runcmd.t
Wabri: What is this magic? :eyes:

t/16-utils-runcmd.t
```diff
@@ -296,4 +297,79 @@ subtest 'saving needle via Git' => sub {
...
+        sub finish { }
+        sub fail {
+            Test::Most::note("Minion job failed with: " . Data::Dumper::Dumper(\@_));    # uncoverable statement
+        }
+    }
+    my $git_mock = Test::MockModule->new('OpenQA::Git');
+    $t->app->config->{global}->{scm} = 'git'; #enable Git
+    $git_mock->redefine(config => {do_cleanup => 'yes'} );
```
perlpunk, line 342: Please use `$t->app->config` here as well

t/16-utils-runcmd.t
```diff
@@ -296,4 +297,79 @@ subtest 'saving needle via Git' => sub {
...
+    $signal_guard_mock->redefine(
+        new => sub {
+            my ($class, @args) = @_;
+            $signal_guard = $class->Mojo::Base::new(@args);
+            return $signal_guard;
+        });
+
+    my $git_mock = Test::MockModule->new('OpenQA::Git');
+    $git_mock->redefine(enabled => 0 );
+    $git_mock->redefine(config => {do_cleanup => 'no'} );
```
perlpunk, line 312: Please use `$t->app->config` here as well

lib/OpenQA/Task/Needle/Save.pm
```diff
@@ -101,10 +102,13 @@ sub _save_needle {
...
     # ensure needle dir is up-to-date
     my $git = OpenQA::Git->new({app => $app, dir => $needledir, user => $user});
+    # avoid retrying if cleanup via Git is not configured (then an admin might need to
+    # manually cleanup the repo first)
+    $signal_guard->abort(1) if !$git->enabled || $git->config->{do_cleanup} eq 'no';
     if ($git->enabled) {
         my $error = $git->set_to_latest_master;
         if ($error) {
-            $app->log->error($error);
+            $app->log->error("Dirty repo detected, a manual cleanup is required. Error: $error");
```
Martchus: The previous call might have failed due to a different reason. I suggest you simply take this one line back.

lib/OpenQA/Task/Needle/Save.pm
```diff
@@ -101,10 +102,13 @@ sub _save_needle {
 
     # ensure needle dir is up-to-date
     my $git = OpenQA::Git->new({app => $app, dir => $needledir, user => $user});
+    # avoid retrying if cleanup via Git is not configured (then an admin might need to
+    # manually cleanup the repo first)
+    $signal_guard->abort(1) if !$git->enabled || $git->config->{do_cleanup} eq 'no';
```
Martchus, line 107: The only caveat with this is that the job will just be marked as normally finished in case of a signal. A failing job might be better to get an admins attention. On the other hand, on a setup without cleanup enabled the admin is probably also just the user clicking the save button and working otherwise closely with the repo (in a manual way). So I guess this is fine.

Martchus (changes requested):

lib/OpenQA/Task/Needle/Save.pm
Wabri: If we consider the previous error message useful ok, otherwise the restore of the line will break the AC3

perlpunk: AC3 is more about the error message that the user will see in the web browser.
And like @Martchus said, we don't know here if the error is actually about a dirty git status.

The error message the user would see is the error passed when calling `$minion_job->fail(...)`.
We discussed yesterday that setting `abort` to true will make the minion job finish successfully, so right now AC3 cannot be fulfilled. I guess the signalhandler class would have to support that.

t/16-utils-runcmd.t
```diff
@@ -296,4 +296,74 @@ subtest 'saving needle via Git' => sub {
...
+            job_id => 99926,
+            user_id => 99903,
+            needle_json => '{"area":[{"xpos":0,"ypos":0,"width":0,"height":0,"type":"match"}],"tags":["foo"]}',
+            needlename => 'foo',
+            needledir => tempdir(),
+            imagedir => 't/data/openqa/share/tests/archlinux/needles',
+            imagename => 'test-rootneedle.png',
+        });
+
+    is($signal_guard->abort, 1, 'signal guard is not aborted initially');
```
perlpunk:
```suggestion
    is $signal_guard->abort, 1, 'signal guard is set to abort';
```

t/16-utils-runcmd.t
```diff
@@ -296,4 +296,74 @@ subtest 'saving needle via Git' => sub {
...
+};
+
+subtest 'save_needle return log error when set_to_latest_master fail' => sub {
+    package Test::FailingMinionJob {
+        sub finish { }
+        sub fail { Test::Most::note("Minion job failed with: " . Data::Dumper::Dumper(\@_)); }
```
perlpunk: I'm wondering why this is uncoverable.
If it's not covered, maybe this can be removed?

t/16-utils-runcmd.t
```diff
@@ -296,4 +296,74 @@ subtest 'saving needle via Git' => sub {
...
+    }    # uncoverable statement
+
+    my $git_mock = Test::MockModule->new('OpenQA::Git');
+    $t->app->config->{global}->{scm} = 'git';    # enable Git
+    $t->app->config->{do_cleanup} = 'yes';    # enable cleanup
+    $git_mock->redefine(
+        set_to_latest_master => sub {
+            return 'Unable to fetch from origin master: mocked error';
+        });
```
perlpunk:
```suggestion
    $git_mock->redefine(set_to_latest_master => 'Unable to fetch from origin master: mocked error');
```

t/16-utils-runcmd.t
```diff
@@ -296,4 +296,74 @@ subtest 'saving needle via Git' => sub {
...
+
+    my @log_errors;
+    my $log_mock = Test::MockModule->new(ref $t->app->log);
+    $log_mock->redefine(
+        error => sub {
+            my ($self, $message) = @_;
+            push @log_errors, $message;
+        });
```
perlpunk:
```suggestion
    $log_mock->redefine(
        error => sub ($self, $message) {
            push @log_errors, $message;
        });
```

t/16-utils-runcmd.t
Wabri: There isn't any uncoverable statement on this line!

lib/OpenQA/Task/Needle/Save.pm
Wabri: Ok this change the way I thought about the AC3, maybe change the line 40 of signal guard into something like `job->fail({error => "<strong>Abort signal received.</strong>"});` could it be more useful for the web user?

Wabri: https://github.com/os-autoinst/openQA/blob/70e849c4db2319951bbd6f894ff67c13418a3b3a/lib/OpenQA/Task/SignalGuard.pm#L39

t/16-utils-runcmd.t
```diff
@@ -296,4 +296,70 @@ subtest 'saving needle via Git' => sub {
...
+            job_id => 99926,
+            user_id => 99903,
+            needle_json => '{"area":[{"xpos":0,"ypos":0,"width":0,"height":0,"type":"match"}],"tags":["foo"]}',
+            needlename => 'foo',
+            needledir => tempdir(),
+            imagedir => 't/data/openqa/share/tests/archlinux/needles',
+            imagename => 'test-rootneedle.png',
+        });
+
+    like $log_errors[0], qr/Unable to fetch.*mocked/, 'error logged on fail';
```
okurz:
```suggestion
    my @log_errors;
    my $log_mock = Test::MockModule->new(ref $t->app->log);
    $log_mock->redefine(error => sub ($self, $message) { push @log_errors, $message });
    my $job = bless({} => 'Test::FailingMinionJob');
    my $needle = {{
            job_id => 99926,
            user_id => 99903,
            needle_json => '{"area":[{"xpos":0,"ypos":0,"width":0,"height":0,"type":"match"}],"tags":["foo"]}',
            needlename => 'foo',
            needledir => tempdir(),
            imagedir => 't/data/openqa/share/tests/archlinux/needles',
            imagename => 'test-rootneedle.png',
        };
    my $result = OpenQA::Task::Needle::Save::_save_needle($t->app, $job, $needle);
    like $log_errors[0], qr/Unable to fetch.*mocked/, 'error logged on fail';
```

t/16-utils-runcmd.t
```diff
@@ -296,4 +296,65 @@ subtest 'saving needle via Git' => sub {
...
+subtest 'save_needle return log error when set_to_latest_master fail' => sub {
+    package Test::FailingMinionJob {
+        sub finish { }
+        sub fail { Test::Most::note("Minion job failed with: " . Data::Dumper::Dumper(\@_)); }
+    }    # uncoverable statement
+
+    my $git_mock = Test::MockModule->new('OpenQA::Git');
+    $t->app->config->{global}->{scm} = 'git';    # enable Git
+    $t->app->config->{do_cleanup} = 'yes';    # enable cleanup
+    $git_mock->redefine(set_to_latest_master => sub { return 'Unable to fetch from origin master: mocked error'; });
```
okurz:
```suggestion
    $git_mock->redefine(set_to_latest_master => 'Unable to fetch from origin master: mocked error');
```

t/16-utils-runcmd.t
```diff
@@ -296,4 +296,65 @@ subtest 'saving needle via Git' => sub {
...
+
+    my @log_errors;
+    my $log_mock = Test::MockModule->new(ref $t->app->log);
+    $log_mock->redefine(
+        error => sub ($self, $message) {
+            push @log_errors, $message;
+        });
```
okurz:
```suggestion
        error => sub ($self, $message) { push @log_errors, $message });
```

t/16-utils-runcmd.t
okurz: seems like you didn't use my suggestion or at least not completely. I will add a suggestion again

t/16-utils-runcmd.t
```diff
@@ -296,4 +296,65 @@ subtest 'saving needle via Git' => sub {
...
 };
 
+subtest 'signal guard aborts when git is disabled and do_cleanup is no' => sub {
+    my $signal_guard;
+    my $signal_guard_mock = Test::MockModule->new('OpenQA::Task::SignalGuard');
+    $signal_guard_mock->redefine(
+        new => sub ($class, @args) {
+            $signal_guard = $signal_guard_mock->original('new')->($class, @args);
+            return $signal_guard;
+        });
```
okurz:
```suggestion
    $signal_guard_mock->redefine(
        new => sub ($class, @args) { $signal_guard_mock->original('new')->($class, @args) });
```

t/16-utils-runcmd.t
```diff
@@ -296,4 +296,65 @@ subtest 'saving needle via Git' => sub {
...
+            job_id => 99926,
+            user_id => 99903,
+            needle_json => '{"area":[{"xpos":0,"ypos":0,"width":0,"height":0,"type":"match"}],"tags":["foo"]}',
+            needlename => 'foo',
+            needledir => tempdir(),
+            imagedir => 't/data/openqa/share/tests/archlinux/needles',
+            imagename => 'test-rootneedle.png',
+        });
+
+    is $signal_guard->abort, 1, 'signal guard is set to abort';
```
okurz:
```suggestion
    ok $signal_guard->abort, 'signal guard is set to abort';
```

t/16-utils-runcmd.t
perlpunk: ah sorry, I guess I got confused by linewrapping in the webui

t/16-utils-runcmd.t
```diff
@@ -296,4 +296,59 @@ subtest 'saving needle via Git' => sub {
...
+            needlename => 'foo',
+            needledir => tempdir(),
+            imagedir => 't/data/openqa/share/tests/archlinux/needles',
+            imagename => 'test-rootneedle.png',
+        });
+
+    ok $signal_guard->abort, 'signal guard is set to abort';
+};
+
+subtest 'save_needle return log error when set_to_latest_master fail' => sub {
```
Martchus:
```suggestion
subtest 'save_needle returns and logs error when set_to_latest_master fails' => sub {
```

t/16-utils-runcmd.t
```diff
@@ -296,4 +296,59 @@ subtest 'saving needle via Git' => sub {
...
+    package Test::FailingMinionJob {
+        sub finish { }
+        sub fail { Test::Most::note("Minion job failed with: " . Data::Dumper::Dumper(\@_)); }
```
Martchus: This should probably not just print a note but also check that the job fails with the expected result (and that the `fail` method is called at all).

t/16-utils-runcmd.t
```diff
@@ -296,4 +296,59 @@ subtest 'saving needle via Git' => sub {
     #       it is tested here explicitly.
 };
 
+subtest 'signal guard aborts when git is disabled and do_cleanup is no' => sub {
```
Martchus:
```suggestion
subtest 'signal guard aborts when git is disabled and do_cleanup is "no"' => sub {
```

t/16-utils-runcmd.t
```diff
@@ -296,4 +296,59 @@ subtest 'saving needle via Git' => sub {
...
+    my $job = bless({} => 'Test::FailingMinionJob');
+    my $needle = {
+        job_id => 99926,
+        user_id => 99903,
+        needle_json => '{"area":[{"xpos":0,"ypos":0,"width":0,"height":0,"type":"match"}],"tags":["foo"]}',
+        needlename => 'foo',
+        needledir => tempdir(),
+        imagedir => 't/data/openqa/share/tests/archlinux/needles',
+        imagename => 'test-rootneedle.png',
+    };
```
Martchus: It would be good to avoid duplicating this data as it seems identical to the data used in the previous subtest.

t/16-utils-runcmd.t
```diff
@@ -296,4 +296,59 @@ subtest 'saving needle via Git' => sub {
     #       it is tested here explicitly.
 };
 
+subtest 'signal guard aborts when git is disabled and do_cleanup is no' => sub {
+    my $signal_guard;
+    my $signal_guard_mock = Test::MockModule->new('OpenQA::Task::SignalGuard');
+    $signal_guard_mock->redefine(
+        new => sub ($class, @args) { $signal_guard = $signal_guard_mock->original('new')->($class, @args); });
```
Martchus: Maybe it makes more sense to return a `Test::MockObject` here (or a fake object of *some* kind) to avoid the signal guard messing with signal handling during test execution.

t/16-utils-runcmd.t
```diff
@@ -296,4 +295,49 @@ subtest 'saving needle via Git' => sub {
     #       it is tested here explicitly.
 };
 
+subtest 'signal guard aborts when git is disabled and do_cleanup is "no"' => sub {
+    package My::FakeSignalGuard {
+        use Mojo::Base -base, -signatures;
+
+        has 'abort' => sub { 0 };
+
+        sub new($class, @args) { bless {abort => 0}, $class; }
```
Martchus: There's no need for an extra constructor if you use `Mojo::Base -base;`. The attribute can be written simpler, too:
```suggestion
        use Mojo::Base -base, -signatures;
        has abort => 0;
```

t/16-utils-runcmd.t
```diff
@@ -296,4 +295,49 @@ subtest 'saving needle via Git' => sub {
...
+
+    # trigger saving needles like Minion would do
+    @executed_commands = ();
+    OpenQA::Task::Needle::Save::_save_needle($t->app, bless({} => 'Test::FakeMinionJob'), $fake_needle);
+
+    isa_ok $signal_guard, 'My::FakeSignalGuard', 'signal guard has been created with the right class';
+    ok $signal_guard->abort, 'signal guard is set to abort';
+};
+
+subtest 'save_needle return and logs error when set_to_latest_master fails' => sub {
```
Martchus:
```suggestion
subtest 'save_needle returns and logs error when set_to_latest_master fails' => sub {
```

Martchus (changes requested):

Martchus (approved):

t/16-utils-runcmd.t
```diff
@@ -247,8 +259,7 @@ subtest 'git commands with mocked run_cmd_with_log_return_error' => sub {
 subtest 'saving needle via Git' => sub {
     my $utils_mock = Test::MockModule->new('OpenQA::Git');
     $utils_mock->redefine(run_cmd_with_log_return_error => \&_run_cmd_mock);
-    {
-        package Test::FakeMinionJob;    # uncoverable statement
+    package Test::FakeMinionJob {
```
perlpunk, line 262: Now the closing `}` needs to get the `# uncoverable statement`.

perlpunk:
>  or your changes influenced the tracking of this line indirectly (which of course shouldn't be the case).

Unfortunately it is:
https://github.com/os-autoinst/openQA/pull/6483/files#r2137609530

AdamWill:
The test changes here set `do_cleanup` in two different wrong places:
```
$t->app->config->{global}->{do_cleanup} = 'no';    # disable cleanup
...
$t->app->config->{do_cleanup} = 'yes';    # enable cleanup
```
neither of those is right. They should both use `$t->app->config->{'scm git'}->{do_cleanup}`.

perlpunk:
Thanks, I reopened the ticket

Wabri:
Thanks @AdamWill I've open a new pr https://github.com/os-autoinst/openQA/pull/6517

PR #6484, opened 2025-05-28
Avoid database error when more than one limit parameter is specified

Martchus:
* Do not treat the limit parameter as a column (where searching for more than one value is possible)
* See https://progress.opensuse.org/issues/183026


Martchus:
Maybe it is simply slow, indeed. I have already commented in the ticket we have for this.

PR #6485, opened 2025-05-29
Add permissions warnings in openQA build

b10n1k:
Just apply the suggestions from the build's output to prevent those warnings.


dist/rpm/openQA.spec
```diff
@@ -796,6 +796,8 @@ fi
 
 %files doc
 %doc docs/*
+%exclude %{_pkgdocdir}/asciinema/.gitignore
+%exclude %{_pkgdocdir}/images/.gitignore
```
Martchus: We should probably fix the install script to avoid installing those files.

Martchus (changes requested):
This doesn't seem to work:

```
[  454s] RPM build errors:
[  454s]     File listed twice: /usr/share/openqa
[  454s]     File must begin with "/": %{_pkgdocdir}/asciinema/.gitignore
[  454s]     File must begin with "/": %{_pkgdocdir}/images/.gitignore
```

dist/rpm/openQA.spec
d3flex: check latest changes. I looked at the scripts in tools/ but it looks there is no need to touch those.

okurz:
> > > Also why were the files added in the first place when they are not needed?
> > 
> > 
> > this might not be the right question. I would like to avoid to answer this for now. @baierjan has been added and he might shed light without me in the middle
> 
> See [#5611 (comment)](https://github.com/os-autoinst/openQA/pull/5611#issuecomment-2095555093)

ok. That reason still applies so the .gitignore files shouldn't be deleted

baierjan:
> > instead a gitignore is introduced to not accidentally push the recordings here
> 
> here refers to where? openQA, right?

here refers to master branch, see https://github.com/os-autoinst/openQA/pull/5616 where the file is actually added in the gh-pages branch instead

.gitignore
```diff
@@ -46,3 +46,5 @@ node_modules
 openqa-documentation.html
 docs/build/
 docs/vendor/
+# docs
+**/openqa-in-5-minutes.gif
```
Martchus, line 50: I'm wondering whether we need this at all. It doesn't seem to be generated in my checkout when generating the documentation and I'm not sure when else it would be generated. I guess we can keep the entry to be sure, though.

baierjan: The generation is "hidden" inside the nested [Makefile](https://github.com/os-autoinst/openQA/blob/master/docs/asciinema/Makefile) and shouldn't be called during normal doc generating process, you need to call it explicitly when you need to regenerate the "video".

d3flex: to be honest, i am not sure what should I do. is this discussion open because I need to change something or ...?

Martchus: No, from my point of view you can keep this as-is.

okurz (approved):

PR #6493, opened 2025-06-03
dist: Fix "Finding LWP::Protocol::https on cpanmetadb failed." on Leap 16.0 build

okurz:
Related progress issue: https://progress.opensuse.org/issues/180803

### DIFF
--- a/dist/rpm/openQA.spec
+++ b/dist/rpm/openQA.spec
@@ -99,6 +99,11 @@ Source0:        %{name}-%{version}.tar.xz
 Source1:        openQA-rpmlintrc
 Source2:        node_modules.spec.inc
 %include        %{_sourcedir}/node_modules.spec.inc
+# Indirect dependency first observed to be necessary on Leap 16.0 (but not
+# <16.0 nor Tumbleweed)
+%if 0%{?suse_version} >= 1600 && 0%{?suse_version} < 1699
+BuildRequires:  perl-LWP-Protocol-https
+%endif
 BuildRequires:  fdupes
 # for install-opensuse in Makefile
 %if 0%{?is_opensuse}


Wabri (approved):

dist/rpm/openQA.spec
```diff
@@ -99,6 +99,11 @@ Source0:        %{name}-%{version}.tar.xz
 Source1:        openQA-rpmlintrc
 Source2:        node_modules.spec.inc
 %include        %{_sourcedir}/node_modules.spec.inc
+# Indirect dependency first observed to be necessary on Leap 16.0 (but not
+# <16.0 nor Tumbleweed)
+%if 0%{?suse_version} >= 1600 && 0%{?suse_version} < 1699
+BuildRequires:  perl-LWP-Protocol-https
```
Martchus, line 105: I suppose this is an optional dependency for an optional feature of a package we use and we always need this optional feature (probably https support). Then this makes sense but the comment should state we need this for otherwise optional https support. We could also make https support optional on our end and only add this package as recommended.

If that's not the case the dependency should be added in the package where it is actually needed.

okurz: > If that's not the case the dependency should be added in the package where it is actually needed.

sure. But how can I find that out?

Martchus: I guess by checking the upstream documentation or relevant spec files for clues.

okurz: so far we only have a dependency on `perl(LWP::UserAgent)` which is provided by perl-libwww-perl. `perl(LWP::Protocol::https)` is referenced in https://build.opensuse.org/projects/openSUSE:Factory/packages/perl-libwww-perl/files/perl-libwww-perl.spec?expand=1 in line 103 . We need to follow the rabbit hole further from there.

PR #6496, opened 2025-06-04
Proof of Concept: Use CSS::Sass

perlpunk:
For now installing from CPAN, as the module is not in Factory.

As soon as CSS::Sass is installed, Mojolicious::Plugin::AssetPack will use it automatically.
It's not possible to configure it.

Issue: https://progress.opensuse.org/issues/162500


okurz:
> For now installing from CPAN, as the module is not in Factory.

As workaround you can submit to devel:openQA (not :Leap) until it's in Factory

perlpunk:
#6501

PR #6497, opened 2025-06-05
Restore behavior in case of empty details JSON after ca86aec6

Martchus:
* Do not return a 500 response but simply assume there are no details
* See https://progress.opensuse.org/issues/183575


lib/OpenQA/Schema/Result/JobModules.pm
```diff
@@ -87,9 +87,11 @@ sub results ($self, %options) {
...
 
     my $file = path($dir, "details-$name.json");
-    my $json_data;
-    try { $json_data = $file->slurp }
-    catch ($e) { return {} }
+    my $json_data = do {
+        try { $file->slurp }
+        catch ($e) { }
+      }
+      or return {};
```
Martchus: Note that this could be a one-liner. However, this is the lowest number of lines that tidy accepted.

(And Perl itself complains if you leave out the `do` or the `catch` block. So it really needs to be that wordy unless we want to go back to `eval`.)

t/10-jobs.t
```diff
@@ -1069,4 +1069,19 @@ subtest 'handling of array and unicode settings' => sub {
...
+    my $empty_module = $modules->create({name => 'empty', script => 'foo', category => 'unittests'});
+    my $broken_module = $modules->create({name => 'broken', script => 'foo', category => 'unittests'});
+
+    my $results = $valid_module->results;
+    is scalar @{$results->{details}}, 1, 'details returned if JSON file exists' or always_explain $results;
+
+    $results = $empty_module->results;
+    is scalar @{$results->{details}}, 0, 'empty details returned if JSON file empty' or always_explain $results;
+
+    throws_ok { $broken_module->results } qr/Malformed.*JSON/i, 'exception if JSON is malformed';
```
Martchus: This is actually questionable as it leads to a 500 response. We should probably also return an empty list of details here and in addition provide some warning that test results contain invalid data. (The warning should probably be rendered on the web page and not go into the server logs.)

However, this is asserting the behavior before ca86aec6 for now. Let me know if I should improve the error handling in general as described as part of https://progress.opensuse.org/issues/183575.

lib/OpenQA/Schema/Result/JobModules.pm
Martchus: The empty `catch` block is of course also a sign of bad/lacking error handling. Let me know if I should improve this as part of https://progress.opensuse.org/issues/183575 in the way I described in the comment about the test.

lib/OpenQA/Schema/Result/JobModules.pm
```diff
@@ -82,20 +82,22 @@ sub sqlt_deploy_hook ($self, $sqlt_table) {
...
     my $file = path($dir, "details-$name.json");
     my $json_data = do {
         try { $file->slurp }
-        catch ($e) { }
+        catch ($e) { push @$errors, qq{Unable to read "$file": $e} if $errors }
+      }
+      or return {};
+    my $json = do {
+        try { decode_json($json_data) }
+        catch ($e) { push @$errors, qq{Malformed JSON file "$file": $e} if $errors }
```
perlpunk: Hm, in the error case `$json` would end up with the result of the `push` command, usually `1`.
It seems that it doesn't make the tests fail, because if `$json == 1` then the
```
    if (ref $details ne 'ARRAY') {                                              
        $results->{details} = [];                                               
        return $results;                                                        
    }                                                                           
```
will return, but it would be nice if that would be handled earlier.

lib/OpenQA/Schema/Result/JobModules.pm
```diff
@@ -82,18 +82,22 @@ sub sqlt_deploy_hook ($self, $sqlt_table) {
...
-    try { $json = decode_json($json_data) }
-    catch ($e) { die qq{Malformed/unreadable JSON file "$file": $e} }
+    my $json_data = do {
+        try { $file->slurp }
+        catch ($e) { push @$errors, qq{Unable to read "$file": $e} if $errors }
+      }
+      or return {};
+    my $json = do {
+        try { decode_json($json_data) }
+        catch ($e) { push @$errors, qq{Malformed JSON file "$file": $e} if $errors }
```
perlpunk:
```suggestion
        catch ($e) { push @$errors, qq{Unable to read "$file": $e} if $errors; {} }
      }
      or return {};
    my $json = do {
        try { decode_json($json_data) }
        catch ($e) { push @$errors, qq{Malformed JSON file "$file": $e} if $errors; {} }
```

Martchus: I made it return `undef` so we don't specify the fallback value twice.

perlpunk (approved):

okurz (approved):

Martchus:
This now shows too many disturbing errors when tests are still running. I'll create a follow-up to suppress errors in this case after all.

Martchus:
Follow-up: https://github.com/os-autoinst/openQA/pull/6509

PR #6498, opened 2025-06-05
WIP

Wabri:
Related: https://progress.opensuse.org/issues/183056


.github/workflows/check-helm-chart.yml
```diff
@@ -10,7 +10,7 @@ jobs:
         uses: actions/checkout@v4
 
       - name: Install Helm
-        uses: azure/setup-helm@v4
+        uses: azure/setup-helm@v4.3.0
```
d3flex: I think that those changes do not bump the version. the difference is that now you target a specific release tag. if you compare the output likely you wont find a big difference because i believe when you use just major version, it will use the latest stable. Look at https://docs.github.com/en/actions/sharing-automations/creating-actions/about-custom-actions

Martchus: I suppose this would indeed just use a different tag. Whether it'll bump the version therefore depends on the usage of those tags. This should probably be double-checked.

Martchus (commented):
I guess a change like that is generally acceptable. I lack the knowledge about the helm ecosystem to know whether the new repository you put in is acceptable.

.github/workflows/check-helm-chart.yml
Wabri: Yes, you are right. Thanks for pointing out!

Wabri:
I'm running 200 stability test locally in order to provide a good statistic of how this could solve the issue we have sporadically.

Wabri:
200 stability test run locally: 100% success, 0% failure!

container/helm/charts/openqa/Chart.yaml
```diff
@@ -10,8 +10,8 @@ dependencies:
     version: ~0.1.0
     condition: webui.enabled
   - name: postgresql
-    version: 6.5.0
-    repository: https://charts.helm.sh/stable
+    version: 15.5.38
+    repository: https://charts.bitnami.com/bitnami
```
Martchus, line 14: So we were on a *very* old version of PostgreSQL. I guess it makes sense to upgrade. I'm wondering how the migration is handled for anyone who previously used this helm chart. Probably it isn't handled at all and breaks unless one follows our migration guide where the commands can probably not copied and pasted directly for this containerized environment. Having this said, I suppose I also don't care.

This is not only upgrading the version but also the domain and thus were this comes from. I don't know enough about this ecosystem to know whether this is a trustworthy source. I guess again I don't *really* care but it would be great if that wouldn't go completely unnoticed in a PR/commit that only says it updates the version.

Martchus (approved):

container/helm/charts/openqa/Chart.yaml
Wabri: That isn't a postgresql version, is the **helm chart** version. This change remove the usage of the deprecated helm repository and replace with the maintained bitnami one.

The new postgresql version is `16.4.0` not `15.5.38`, the latest stable release of the old chart was using `11.7.0`.

Reference:
- https://github.com/helm/charts/blob/master/stable/postgresql/README.md#this-helm-chart-is-deprecated
- https://helm.sh/blog/charts-repo-deprecation/

PR #6499, opened 2025-06-06
Use CSS::Sass instead of rubygem-sass

perlpunk:
rubygem-sass is not supported anymore

Issue: https://progress.opensuse.org/issues/162500

### DIFF
--- a/container/devel:openQA:ci/base/Dockerfile
+++ b/container/devel:openQA:ci/base/Dockerfile
@@ -14,7 +14,7 @@ RUN zypper -n in tar gzip sudo
 RUN zypper install -y gcc-c++ cmake ninja pkgconfig\(opencv4\) pkg-config perl\(Module::CPANfile\) pkgconfig\(fftw3\) pkgconfig\(libpng\) pkgconfig\(sndfile\) pkgconfig\(theoraenc\) tesseract-ocr
 
 # openQA dependencies
-RUN zypper install -y 'rubygem(sass)>=3.7.4' ruby-devel npm python3-base python3-requests git-core rsync curl postgresql-devel postgresql-server qemu qemu-tools tar xorg-x11-fonts sudo make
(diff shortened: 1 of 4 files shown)


container/devel:openQA:ci/base/Dockerfile
```diff
@@ -14,7 +14,7 @@ RUN zypper -n in tar gzip sudo
 RUN zypper install -y gcc-c++ cmake ninja pkgconfig\(opencv4\) pkg-config perl\(Module::CPANfile\) pkgconfig\(fftw3\) pkgconfig\(libpng\) pkgconfig\(sndfile\) pkgconfig\(theoraenc\) tesseract-ocr
 
 # openQA dependencies
-RUN zypper install -y 'rubygem(sass)>=3.7.4' ruby-devel npm python3-base python3-requests git-core rsync curl postgresql-devel postgresql-server qemu qemu-tools tar xorg-x11-fonts sudo make
+RUN zypper install -y 'perl(CSS::Sass)' npm python3-base python3-requests git-core rsync curl postgresql-devel postgresql-server qemu qemu-tools tar xorg-x11-fonts sudo make
```
perlpunk, line 17: Not sure about deleting `ruby-devel`. Do we need ruby for something else? for asciidoctor?

Martchus:
> … and stick with rubygem-sass in Leap 15.

Maybe we should do that. There's probably no immediate harm in using different sass backends. When Leap 15 goes out of support we can also fully remove the dependency on rubygem-sass.

container/devel:openQA:ci/base/Dockerfile
okurz: Yes, see tools/generate-documentation and any reference for the command line tool `gem`

okurz (changes requested):

PR #6500, opened 2025-06-06
Allow using list and range specifiers in worker config file

Martchus:
Related ticket: https://progress.opensuse.org/issues/183101

---

This is so far only introducing list and range specifiers. It doesn't change the overriding behavior. So the behavior is still as follows:

The section `global` is read first. Other sections are read in the order they are defined and values from previous sections are overridden. This is what we want except that it should be possible to concatenate worker classes (instead of overriding). ~~For this I'd go for a `+=` syntax because concatenating by default and only clearing when an empty value occurs would not be backwards compatible/consistent with our current behavior of overriding (even though it only happened between the global and instance-specific section (and perhaps the web UI specific).~~ The `+=` syntax has now been implemented.

### DIFF
--- a/lib/OpenQA/Worker/Settings.pm
+++ b/lib/OpenQA/Worker/Settings.pm
@@ -70,23 +70,12 @@ sub new ($class, $instance_number = undef, $cli_options = {}) {
     return $self;
 }
 
-sub _is_uint ($value) { $value =~ m/^\d+$/ }
-
 sub _is_section_relevant ($section_name, $instance_number) {
     return 1 if $section_name eq $instance_number;
-    return 0 unless _is_uint $instance_number;
-    my @section_bounds = map {
-        [map { trim $_ } split '-', $_, 2]
-    } split ',', $section_name;
-    for my $bounds (@section_bounds) {
-        my ($lower_bound, $upper_bound) = @$bounds;
-        return 1
-          if defined $upper_bound
-          && _is_uint($lower_bound)
-          && _is_uint($upper_bound)
-          && $instance_number >= $lower_bound
-          && $instance_number <= $upper_bound;
-        return 1 if _is_uint($lower_bound) && $instance_number == $lower_bound;
+    return 0 unless $instance_number =~ m/^\d+$/;
+    for my $range (split m/ *, */, $section_name) {
+        next unless $range =~ m/^(\d+) *(?:- *(\d+))?$/;
+        return 1 if $instance_number >= $1 && $instance_number <= ($2 // $1);
     }
     return 0;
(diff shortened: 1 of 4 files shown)


t/24-worker-settings.t
```diff
@@ -71,11 +65,7 @@ subtest 'check for local worker' => sub {
 subtest 'apply settings to app' => sub {
     my ($setup_log_called, $setup_log_app);
     my $mock = Test::MockModule->new('OpenQA::Worker::Settings');
-    $mock->redefine(
-        setup_log => sub {
-            $setup_log_app = shift;
-            $setup_log_called = 1;
-        });
+    $mock->redefine(setup_log => sub ($setup_log_app) { $setup_log_called = 1 });
```
perlpunk: You need to enable signatures for that

lib/OpenQA/Worker/Settings.pm
```diff
@@ -71,6 +72,30 @@ sub new ($class, $instance_number = undef, $cli_options = {}) {
     return $self;
 }
 
+sub _is_section_relevant ($section_name, $instance_number) {
+    return 1 if $section_name eq $instance_number;
+    return 0 unless looks_like_number $instance_number;
```
perlpunk: `looks_like_number` also returns true for floats and Infinity. Maybe we can be a bit more strict here?

lib/OpenQA/Worker/Settings.pm
```diff
@@ -71,6 +71,32 @@ sub new ($class, $instance_number = undef, $cli_options = {}) {
...
+    for my $bounds (@section_bounds) {
+        my ($lower_bound, $upper_bound) = @$bounds;
+        return 1
+          if defined $upper_bound
+          && _is_uint($lower_bound)
+          && _is_uint($upper_bound)
+          && $instance_number >= $lower_bound
+          && $instance_number <= $upper_bound;
+        return 1 if _is_uint($lower_bound) && $instance_number == $lower_bound;
+    }
```
perlpunk: This might be less readable, but I thought I'd still suggest it:
```suggestion
    return 0 unless $instance_number =~ m/^\d+$/;
    for my $range (split m/ *, */, $section_name) {
        next unless $range =~ m/^(\d+) *(?:- *(\d+))?$/;
        return 1 if $instance_number >= $1 && $instance_number <= ($2 // $1);
```
edit: simplified the regex

Martchus: It is less readable in my opinion but nevertheless also more elegant as the validation and trimming is happens "along the way". So I took your change.

Issue #6505, opened 2025-06-08
Print one url per line for setting links (only if it is an array of space separated urls)


alimirjamali:
While working with `openqa.qubes-os.org`, it is hard to follow white-space separated links in test settings (e.g. [this test](https://openqa.qubes-os.org/tests/142048#settings)). It looks like the screenshot below:

![Image](https://github.com/user-attachments/assets/463763bf-85bd-49ac-a0fc-ae765c60921b)

I already made a patch to make it like this:

![Image](https://github.com/user-attachments/assets/af7e5c25-c4bb-4585-8f8b-c93ee8a44b88)

PR #6506, opened 2025-06-08
Print one url per line in test settings

alimirjamali:
resolves: https://github.com/os-autoinst/openQA/issues/6505

### DIFF
--- a/assets/javascripts/openqa.js
+++ b/assets/javascripts/openqa.js
@@ -608,7 +608,7 @@ function renderComments(row) {
   return html;
 }
 
-function renderHttpUrlAsLink(value) {
+function renderHttpUrlAsLink(value, oneUrlPerLine = false) {
   if (!value) {
     return document.createTextNode('');
   }
@@ -626,6 +626,11 @@ function renderHttpUrlAsLink(value) {
     value = String(value);
   }
   const urlRegex = /https?:\/\/[^\s,]*/g;
+  if (oneUrlPerLine) {
+    // Regex to match an array of space-separated URLs (to print one per line)
+    const arrayOfUrlsRegex = /^(https?:\/\/[^\s,]*)(\s+https?:\/\/[^\s,]*)+\s*$/g;
+    oneUrlPerLine = arrayOfUrlsRegex.exec(value) !== null;
+  }
   let lastIndex = 0;
   let match;
   while ((match = urlRegex.exec(value)) !== null) {
@@ -634,6 +639,9 @@ function renderHttpUrlAsLink(value) {
     }
     const a = document.createElement('a');
     a.href = a.textContent = match[0];
+    if (oneUrlPerLine) {
+      a.style.display = 'block';
+    }
     fragment.appendChild(a);
     lastIndex = urlRegex.lastIndex;
   }
--- a/assets/javascripts/test_result.js
+++ b/assets/javascripts/test_result.js
@@ -1116,7 +1116,7 @@ function renderSettingsTab(response) {
   Array.from(tabPanelElement.getElementsByClassName('settings-value')).forEach(settingsLink => {
     const url = settingsLink.textContent;
     settingsLink.innerHTML = null;
-    settingsLink.appendChild(renderHttpUrlAsLink(url));
+    settingsLink.appendChild(renderHttpUrlAsLink(url, true));
   });
 }
 
--- a/t/ui/28-keys_to_render_as_links.t
+++ b/t/ui/28-keys_to_render_as_links.t
@@ -28,7 +28,8 @@ my $job_id = 99938;
 my $foo_path = "foo/foo.txt";
 my $uri_path_from_root_dir = "/tests/$job_id/settings/$foo_path";
 my $uri_path_from_default_data_dir = "/tests/$job_id/settings/bar/foo.txt";
-$schema->resultset('Jobs')->find($job_id)->settings->create({key => 'SOME_URLS', value => 'https://foo,http://bar'});
+$schema->resultset('Jobs')->find($job_id)
+  ->settings->create({key => 'SOME_URLS', value => 'https://foo http://bar http://baz'});
 
 driver_missing unless my $driver = call_driver;
 my $url = 'http://localhost:' . OpenQA::SeleniumTest::get_mojoport;
@@ -42,10 +43,14 @@ $driver->get("/tests/$job_id#settings");
 note 'Finding link associated with keys_to_render_as_links';
 ok $driver->find_element_by_link($foo_path), 'configured setting key rendered as link';
 ok $driver->find_element_by_link('https://foo'), 'multi URL rendered as link (1)';
+is $driver->find_element_by_link('https://foo')->get_css_attribute('display'), 'block';
 ok $driver->find_element_by_link('http://bar'), 'multi URL rendered as link (2)';
+is $driver->find_element_by_link('http://bar')->get_css_attribute('display'), 'block';
+ok $driver->find_element_by_link('http://baz'), 'multi URL rendered as link (3)';
+is $driver->find_element_by_link('http://baz')->get_css_attribute('display'), 'block';
 note 'Making sure that no other settings are rendered as links';
 my @number_of_elem = $driver->find_elements('#settings_box a');
-is(scalar @number_of_elem, 3, 'only configured setting keys and URLs render as links');
+is(scalar @number_of_elem, 4, 'only configured setting keys and URLs render as links');
 note 'Checking link navigation to the source';
 $driver->find_element_by_link($foo_path)->click();
 is($driver->get_current_url(), "$url$uri_path_from_root_dir", 'Link is accessed with correct URI');


assets/javascripts/openqa.js
```diff
@@ -626,15 +626,21 @@ function renderHttpUrlAsLink(value) {
     value = String(value);
   }
   const urlRegex = /https?:\/\/[^\s,]*/g;
+  const arrayOfUrlsRegex = /^(https?:\/\/[^\s,]*)(\s+https?:\/\/[^\s,]*)+$/g;
```
d3flex: seems to work. would you mind to add a comment that this is intent to match arrays of URL *with space*. And I would appreciate it if you could write this also in the commit description.

assets/javascripts/openqa.js
```diff
@@ -626,15 +626,21 @@ function renderHttpUrlAsLink(value) {
...
   let match;
+  let arrayOfUrls = arrayOfUrlsRegex.exec(value) !== null;
   while ((match = urlRegex.exec(value)) !== null) {
     if (match.index > lastIndex) {
       fragment.appendChild(document.createTextNode(value.substring(lastIndex, match.index)));
     }
     const a = document.createElement('a');
     a.href = a.textContent = match[0];
     fragment.appendChild(a);
+    if (oneUrlPerLine && arrayOfUrls) {
```
d3flex: small thing but I guess `oneUrlPerLine` should be checked once `arrayOfUrls` is true. Not the other way around.

b10n1k (changes requested):
codecov is missing. it would be nice to add some test anyways. wdyt?

assets/javascripts/openqa.js
alimirjamali: This is resolved

assets/javascripts/openqa.js
alimirjamali: This is resolved

alimirjamali:
> codecov is missing. it would be nice to add some test anyways. wdyt?

I have a problem with unittests. I tried to implement it (without having prior Perl knowledge). But One of the unittests fail and I have no idea. I appreciate if you could take a look

assets/javascripts/openqa.js
```diff
@@ -626,15 +626,22 @@ function renderHttpUrlAsLink(value) {
...
   let match;
+  let arrayOfUrls = arrayOfUrlsRegex.exec(value) !== null;
   while ((match = urlRegex.exec(value)) !== null) {
     if (match.index > lastIndex) {
       fragment.appendChild(document.createTextNode(value.substring(lastIndex, match.index)));
     }
     const a = document.createElement('a');
     a.href = a.textContent = match[0];
     fragment.appendChild(a);
+    if (arrayOfUrls && oneUrlPerLine) {
```
d3flex: here it would be better to use Array.isArray 

Something like, but then you should also change the variable name. (I think it is a little weird that _array_ appears so many times)) 

```suggestion
    if (Array.isArray(arrayOfUrls) && oneUrlPerLine) {
```

b10n1k:
> > > codecov is missing. it would be nice to add some test anyways. wdyt?
> > 
> > 
> > I have a problem with unittests. I tried to implement it (without having prior Perl knowledge). But One of the unittests fail and I have no idea. I appreciate if you could take a look
> 
> first thanks for the test. there is no rush. I will take a look and I will let you know

image

I dont know whether you have seen this or how you run the test.. This is the view when I run your test. I think this is not what you expect so I would say, something is wrong.
But I noticed that this is not the test's settings page but the `/admin/productlog?id=xxx`

assets/javascripts/openqa.js
alimirjamali: OK. I added this. This is resoved

alimirjamali:
> I dont know whether you have seen this or how you run the test.. This is the view when I run your test. I think this is not what you expect so I would say, something is wrong.
> But I noticed that this is not the test's settings page but the `/admin/productlog?id=xxx`

I fixed the unittest. I had to install all requirements and do the tests locally (instead of relying on intuition and abusing online CI/CD facilities). But it was a lot of of Perl modules to install.

assets/javascripts/openqa.js
d3flex: I am gonna resolve this as comment added + commit was updated. However, I'm not sure about the commit wording...

alimirjamali:
I tried `make test-ui` locally on `master` branch. It fails even without my patch.

assets/javascripts/openqa.js
```diff
@@ -626,15 +626,22 @@ function renderHttpUrlAsLink(value) {
...
   let match;
+  let arrayOfUrls = arrayOfUrlsRegex.exec(value) !== null;
   while ((match = urlRegex.exec(value)) !== null) {
     if (match.index > lastIndex) {
       fragment.appendChild(document.createTextNode(value.substring(lastIndex, match.index)));
     }
     const a = document.createElement('a');
     a.href = a.textContent = match[0];
     fragment.appendChild(a);
+    if (Array.isArray(arrayOfUrls) && oneUrlPerLine) {
```
Martchus: The check `Array.isArray(arrayOfUrls)` makes not much sense because it will always be false because a boolean is assigned previously to that variable via `let arrayOfUrls = arrayOfUrlsRegex.exec(value) !== null;`.

assets/javascripts/openqa.js
```diff
@@ -626,15 +626,22 @@ function renderHttpUrlAsLink(value) {
     value = String(value);
   }
   const urlRegex = /https?:\/\/[^\s,]*/g;
+  // Regex to match an array of space-separated URLs (to print one per line)
+  const arrayOfUrlsRegex = /^(https?:\/\/[^\s,]*)(\s+https?:\/\/[^\s,]*)+\s*$/g;
   let lastIndex = 0;
   let match;
+  let arrayOfUrls = arrayOfUrlsRegex.exec(value) !== null;
```
Martchus: It would be nicer to avoid executing the regex if `oneUrlPerLine` is false anyway.

assets/javascripts/openqa.js
```diff
@@ -626,15 +626,22 @@ function renderHttpUrlAsLink(value) {
...
   while ((match = urlRegex.exec(value)) !== null) {
     if (match.index > lastIndex) {
       fragment.appendChild(document.createTextNode(value.substring(lastIndex, match.index)));
     }
     const a = document.createElement('a');
     a.href = a.textContent = match[0];
     fragment.appendChild(a);
+    if (Array.isArray(arrayOfUrls) && oneUrlPerLine) {
+      const br = document.createElement('br');
+      fragment.appendChild(br);
```
Martchus: Wouldn't this insert a trailing surplus `br` element?

assets/javascripts/openqa.js
```diff
@@ -626,15 +626,22 @@ function renderHttpUrlAsLink(value) {
...
+  // Regex to match an array of space-separated URLs (to print one per line)
+  const arrayOfUrlsRegex = /^(https?:\/\/[^\s,]*)(\s+https?:\/\/[^\s,]*)+\s*$/g;
   let lastIndex = 0;
   let match;
+  let arrayOfUrls = arrayOfUrlsRegex.exec(value) !== null;
   while ((match = urlRegex.exec(value)) !== null) {
     if (match.index > lastIndex) {
       fragment.appendChild(document.createTextNode(value.substring(lastIndex, match.index)));
     }
     const a = document.createElement('a');
```
Martchus, line 640: Considering my other comments the proposed implementation is not ideal. Maybe we can get away with something much simpler instead:

```suggestion
    const a = document.createElement('a');
    if (oneUrlPerLine) {
        a.style.display = 'block';
    }
```

Martchus (changes requested):

assets/javascripts/openqa.js
alimirjamali: Yes. It does. I found out about it while writing the unittest

assets/javascripts/openqa.js
alimirjamali: This is resolved

assets/javascripts/openqa.js
alimirjamali: This is resolved

assets/javascripts/openqa.js
alimirjamali: I resolved the issue with surplus `br` element

assets/javascripts/openqa.js
alimirjamali: I axed `arrayOfUrls` and reused the existing `oneUrlPerLine` boolean variable. I appreciae if you could take another look

Martchus: So this suggestion which just relies on `display: block` and doesn't need any additional regexes doesn't work?

alimirjamali: Oh. I missed this one suggestion. I will check it now (but at least, the CI/CD tests are green this time)

alimirjamali: This solution works like a charm. The only problem is I do not know how to check for it in the unittests (my current knowledge of Perl is very limited).

alimirjamali: @Martchus This is also done. I updated the unittests to check if the `display: block` style is applied.

b10n1k (approved):

PR #6512, opened 2025-06-11
apparmor: Support CASEDIR with the needles included

rclobus:
Fixes a hanging 'git clone' when CASEDIR is used and the git repository includes the needles

### DIFF
--- a/profiles/apparmor.d/usr.share.openqa.script.worker
+++ b/profiles/apparmor.d/usr.share.openqa.script.worker
@@ -81,6 +81,7 @@
   /sys/firmware/devicetree/base/ r,
   /sys/fs/cgroup/systemd/openqa.slice/** rw,
   /sys/kernel/mm/transparent_hugepage/enabled r,
+  /sys/kernel/mm/transparent_hugepage/hpage_pmd_size r,
   /tmp/** rwk,
   /usr/bin/Xvnc rCx,
   /{usr/,}bin/cat rix,


rclobus:
I killed the 'git lfs' process in https://openqa.debian.net/tests/405567. A git clone was working before, see https://openqa.debian.net/tests/405565.

So the failing git clone might be an issue with the git server. However, this entry is shown in the dmesg log, and it appears harmless, so it could be included in the apparmor profile.

profiles/apparmor.d/usr.share.openqa.script.worker
```diff
@@ -81,6 +81,7 @@
   /sys/firmware/devicetree/base/ r,
   /sys/fs/cgroup/systemd/openqa.slice/** rw,
   /sys/kernel/mm/transparent_hugepage/enabled r,
+  /sys/kernel/mm/transparent_hugepage/hpage_pmd_size r,
```
Martchus, line 84: Out of curiosity: What Git feature exactly requires this? Is having needles implicitly enabling that feature? I suppose the answer to the last question is no (even though the commit message kind of implies that the answer is yes).

Your additional comment actually reads like this is not really related to cloning getting stuck and just something you found while investigating. Then we should maybe rephrase the commit message like "Allow access to hpage_pmd_size which is used by Git under certain conditions" (although maybe we need to shorten that).

rclobus: After starting the test
```
openqa-cli api -X POST isos DESKTOP=windows DISTRI=windows FLAVOR=DVD VERSION=11 BUILD=24H2 ARCH=x86_64 CASEDIR=https://salsa.debian.org/rclobus-guest/openqa-tests-debian.git#rclobus/dualboot_windows11
```

This was the process tree:
```
$ ps -edaf | grep git
_openqa+   25553   25535  0 12:23 ?        00:00:00 sh -c env GIT_SSH_COMMAND="ssh -oBatchMode=yes" git clone --depth='1' https://salsa.debian.org/rclobus-guest/openqa-tests-debian.git 2>&1
_openqa+   25554   25553  3 12:23 ?        00:00:00 git clone --depth=1 https://salsa.debian.org/rclobus-guest/openqa-tests-debian.git
_openqa+   25570   25554  0 12:23 ?        00:00:00 /bin/sh -c git-lfs filter-process git-lfs filter-process
_openqa+   25571   25570 24 12:23 ?        00:00:00 git-lfs filter-process
user       25611    4368  0 12:23 pts/1    00:00:00 grep git
```
After 10 minutes, the `git-lfs filter-process` was still active.

In dmesg, an entry similar to the following was logged:
```
[  640.484623] audit: type=1400 audit(1749453420.608:58): apparmor="DENIED" operation="open" profile="/usr/share/openqa/script/worker" name="/sys/kernel/mm/transparent_hugepage/hpage_pmd_size" pid=3473 comm="git-lfs" requested_mask="r" denied_mask="r" fsuid=995 ouid=0
```

Martchus: Then I guess the Git feature/extension this is required by is https://git-lfs.com. I suppose it makes sense to use this when dealing with needles but of course it makes sense to refer to it explicitly in the commit message. So I guess the commit message should be e.g. "Allow using git-lfs". (The commit message is important here as this is so far the only way we document *why* we need a certain line in the profile.)

rclobus: I've updated the commit message. I hope it is now sufficiently clear.

rclobus:
@Martchus Some CI checks failed, but their result should have been unaffected by this change. I cannot find a way to restart them. Is that a blocker for the merge?

asdil12 (approved):

PR #6522, opened 2025-06-12
Prevent keep logs in days being higher than keep results in days

Wabri:
Reference: https://progress.opensuse.org/issues/64830

### DIFF
--- a/lib/OpenQA/WebAPI/Controller/API/V1/JobGroup.pm
+++ b/lib/OpenQA/WebAPI/Controller/API/V1/JobGroup.pm
@@ -272,6 +272,14 @@ sub update ($self) {
     }
 
     my $properties = $self->load_properties;
+
+    if ($properties->{default_keep_logs_in_days} > $properties->{default_keep_results_in_days}) {
+        return $self->render(
+            json => {error => '`keep_logs_in_days` must be grater than `keep_results_in_days`'},
+            status => 400
+        );
+    }
+
     my $id;
     try { $id = $group->update($properties)->id }
     catch ($e) { return $self->render(json => {error => $e}, status => 400) }


perlpunk (commented):
You could also start with test driven development in t/api/10-jobgroups.t

Wabri:
> You could also start with test driven development in t/api/10-jobgroups.t

Yes, the tests were failing due to undefined values used in the condition of if I've added, so first I fix that, now I'm adding test that will cover those new lines!

lib/OpenQA/WebAPI/Controller/API/V1/JobGroup.pm
```diff
@@ -272,6 +272,17 @@ sub update ($self) {
     }
 
     my $properties = $self->load_properties;
+
+    if (   defined $properties->{default_keep_logs_in_days}
+        && $properties->{default_keep_results_in_days}
```
Wabri: Maybe a defined is necessary here

```suggestion
        && defined $properties->{default_keep_results_in_days}
```

t/api/10-jobgroups.t
```diff
@@ -479,4 +479,47 @@ subtest 'prevent create/update duplicate job group on same parent group' => sub(
...
+    )->status_is(400, 'if keep_logs_in_days greater than keep_results_in_days update should fail')
+     ->json_is(
+        '/error' => '`keep_logs_in_days` must be grater than `keep_results_in_days`',
+        'error returned succesfuly'
+     );
+
+
+    # TODO: is it necessary?
+    return always_explain $t->tx->res->json unless $t->success;
+};
```
Wabri: Remove the `TODO` from tests

Wabri (commented):
Notes for me

lib/OpenQA/WebAPI/Controller/API/V1/JobGroup.pm
```diff
@@ -272,6 +272,17 @@ sub update ($self) {
...
+        && $properties->{default_keep_logs_in_days} > $properties->{default_keep_results_in_days})
```
perlpunk: perl has a `defined-or` operator `//`:
```suggestion
    if ($properties->{default_keep_logs_in_days} // 0 > $properties->{default_keep_results_in_days} // 0)
```

perlpunk: But it seems to be possible to just update one setting via the API, without passing all the other values.
So someone could just update the value of `default_keep_logs_in_days`. In that case we would have to retrieve the existing value for `default_keep_results_in_days` from the database.

lib/OpenQA/WebAPI/Controller/API/V1/JobGroup.pm
```diff
@@ -272,6 +272,17 @@ sub update ($self) {
...
+    {
+        return $self->render(
+            json => {error => '`keep_logs_in_days` must be grater than `keep_results_in_days`'},
```
perlpunk:
```suggestion
            json => {error => '`default_keep_logs_in_days` must be grater than `default_keep_results_in_days`'},
```

lib/OpenQA/WebAPI/Controller/API/V1/JobGroup.pm
Martchus: `defined` is probably either required for both or none of those variables. So you should check which one it is.

t/api/10-jobgroups.t
```diff
@@ -479,4 +479,47 @@ subtest 'prevent create/update duplicate job group on same parent group' => sub(
...
+        }
+    )->status_is(400, 'if keep_logs_in_days greater than keep_results_in_days update should fail')
+     ->json_is(
+        '/error' => '`keep_logs_in_days` must be grater than `keep_results_in_days`',
+        'error returned succesfuly'
+     );
+
+
+    # TODO: is it necessary?
+    return always_explain $t->tx->res->json unless $t->success;
```
Martchus: I sometimes like to keep this kind of debugging code but feel free to remove it.

lib/OpenQA/WebAPI/Controller/API/V1/JobGroup.pm
Wabri: Interesting, changed and add parenthesis otherwise will complain for uninitialized value.

t/api/10-jobgroups.t
Wabri: Ok, I was asking because not sure about the usefullnes!

lib/OpenQA/WebAPI/Controller/API/V1/JobGroup.pm
```diff
@@ -5,6 +5,7 @@ package OpenQA::WebAPI::Controller::API::V1::JobGroup;
 use Mojo::Base 'Mojolicious::Controller', -signatures;
 use Mojo::JSON;
 use Feature::Compat::Try;
+use Data::Dumper;
```
Wabri: TODO: remove this

lib/OpenQA/WebAPI/Controller/API/V1/JobGroup.pm
```diff
@@ -187,6 +188,28 @@ sub _validate_common_properties ($self) {
...
+    } else {
+        if ( ($properties->{keep_logs_in_days} // 0) > ($properties->{keep_results_in_days} // 0)) {
+          $self->render(
+            json => {error => '`keep_logs_in_days` must be greater than `keep_results_in_days`'},
+            status => 400
+          );
+          return 1;
+        }
+    }
+    return undef;
```
Wabri: Why undef? 🤔

okurz: That's correct style in Perl as the fallback.

lib/OpenQA/WebAPI/Controller/API/V1/JobGroup.pm
```diff
@@ -187,6 +188,28 @@ sub _validate_common_properties ($self) {
...
+# TODO: this exceed runtime limit for some reason
+sub _check_keep_logs_and_results ($self, $properties) {
+    if ($self->is_parent) {
+        if ( ($properties->{default_keep_logs_in_days} // 0) > ($properties->{default_keep_results_in_days} // 0)) {
+          $self->render(
+            json => {error => '`default_keep_logs_in_days` must be greater than `default_keep_results_in_days`'},
+            status => 400
+          );
+          return 1;
+        }
```
okurz: Can this be

```suggestion
    return $self->render(
            json => {error => '`default_keep_logs_in_days` must be greater than `default_keep_results_in_days`'},
            status => 400
          ) if $self->is_parent && ($properties->{default_keep_logs_in_days} // 0) > ($properties->{default_keep_results_in_days} // 0);
```

and accordingly for the else branch

t/api/10-jobgroups.t
```diff
@@ -479,4 +479,91 @@ subtest 'prevent create/update duplicate job group on same parent group' => sub(
...
+    $t->post_ok(
+        '/api/v1/parent_groups',
+        form => {
+            name => 'test_parent_group',
+            default_keep_logs_in_days => 888,
+            default_keep_results_in_days => 777,
+        }
+    )->status_is(400, 'if default_keep_logs_in_days greater than default_keep_results_in_days create should fail')
+     ->json_is(
+        '/error' => '`default_keep_logs_in_days` must be grater than `default_keep_results_in_days`',
```
okurz: Not sure about the choice of
<p><a href="https://commons.wikimedia.org/wiki/File:Kuechenreibe_(fcm).jpg#/media/File:Kuechenreibe_(fcm).jpg"></a><br>By Frank C. Müller, <a href="https://creativecommons.org/licenses/by-sa/4.0" title="Creative Commons Attribution-Share Alike 4.0">CC BY-SA 4.0</a>, <a href="https://commons.wikimedia.org/w/index.php?curid=374762">Link</a></p>
here :)

t/api/10-jobgroups.t
```diff
@@ -479,4 +479,91 @@ subtest 'prevent create/update duplicate job group on same parent group' => sub(
...
+        'error returned succesfuly'
```
okurz: also no :)

t/api/10-jobgroups.t
```diff
@@ -479,4 +479,91 @@ subtest 'prevent create/update duplicate job group on same parent group' => sub(
...
+    # TODO: refactor this test by extracting the common code from the next test
+    # parent group test
+    $t->post_ok(
+        '/api/v1/parent_groups',
+        form => {
+            name => 'test_parent_group',
+            default_keep_logs_in_days => 888,
+            default_keep_results_in_days => 777,
+        }
+    )->status_is(400, 'if default_keep_logs_in_days greater than default_keep_results_in_days create should fail')
```
okurz: I guess if you solve your TODO in https://github.com/os-autoinst/openQA/pull/6522/files#diff-c1af4e5c4c843b331bde9bcfcb93d638a6fe8bc375927237d286c24b2cf5e128R318 by extracting helper constants/functions then also here you don't need to redo complete (heavy) requests over and over again but simply check function calls which should make this code easier as well

okurz (changes requested):

lib/OpenQA/WebAPI/Controller/API/V1/JobGroup.pm
```diff
@@ -187,6 +188,28 @@ sub _validate_common_properties ($self) {
     $validation->optional('description');
 }
 
+# TODO: this exceed runtime limit for some reason
```
perlpunk: Did you use:
```
return undef if $self->_check_keep_logs_and_results($properties);
```
in `create` and `update`? I see a very small increase in the runtime.
However, if I use
```
$self->_check_keep_logs_and_results($properties);
```
without returning then it runs into the timeout with the error:
```
[error] [BTyZmHNypvRy] A response has already been rendered at /usr/lib/perl5/vendor_perl/5.26.1/Mojolicious/Controller.pm line 189.
Mojo::Reactor::Poll: I/O watcher failed: A response has already been rendered at /usr/lib/perl5/vendor_perl/5.26.1/Mojolicious/Controller.pm line 189.
Bailout called.  Further testing stopped:  test 't/api/10-jobgroups.t' exceeds runtime limit of '8' seconds
```

lib/OpenQA/WebAPI/Controller/API/V1/JobGroup.pm
perlpunk: I guess you can also `return 0;`

perlpunk: But I think the method name suggests the reverse. A name with `check` should return something true if the check passed.

lib/OpenQA/WebAPI/Controller/API/V1/JobGroup.pm
```diff
@@ -187,6 +188,28 @@ sub _validate_common_properties ($self) {
...
+sub _check_keep_logs_and_results ($self, $properties) {
+    if ($self->is_parent) {
+        if ( ($properties->{default_keep_logs_in_days} // 0) > ($properties->{default_keep_results_in_days} // 0)) {
+          $self->render(
+            json => {error => '`default_keep_logs_in_days` must be greater than `default_keep_results_in_days`'},
```
perlpunk: The ticket's AC says:
> prevent the value of `keep_logs_in_days` being higher than `keep_results_in_days` in UI

So `keep_logs` must be lower than `keep_results'.
In the `if` you check if `keep_logs` is greater than `keep_results`, but in the error message you say the opposite.

t/api/10-jobgroups.t
Wabri: This takes me a while 🤣

lib/OpenQA/WebAPI/Controller/API/V1/JobGroup.pm
Wabri: Fixed 🙂

lib/OpenQA/WebAPI/Controller/API/V1/JobGroup.pm
Wabri: Ups 😅

lib/OpenQA/WebAPI/Controller/API/V1/JobGroup.pm
Wabri: I'm not sure if could it work like this, I try but `Mojo::Reactor::Poll: I/O watcher failed: A response has already been rendered at /usr/lib/perl5/vendor_perl/5.40.2/Mojolicious/Controller.pm line 189.`

lib/OpenQA/WebAPI/Controller/API/V1/JobGroup.pm
```diff
@@ -187,6 +188,40 @@ sub _validate_common_properties ($self) {
     $validation->optional('description');
 }
 
+sub _check_keep_logs_and_results ($self, $properties) {
+
+# TODO: try to understand why this throw `Mojo::Reactor::Poll: I/O watcher failed: A response has already been rendered`
```
perlpunk: The return value must indicate to the calling method if it should return itself, so we need to explicitly return something true or false. I currently wouldn't know by heart if `render` would return true or false, so I think an explicit return statement is better

Wabri: Looking at the documentation https://docs.mojolicious.org/Mojolicious/Controller#render it should return a bool so is already a true false return.

Maybe there are something more inside the render subroutine that I don't know!

perlpunk: Repeating myself:
The return value must indicate to the calling method if it should return itself, so we need to explicitly return something true or false.
The method is called like this:
```
return undef unless $self->_check_keep_logs_and_results($properties);
```
So let's assume, `render` returns true.
Now that would signal the calling method that the check passed. So it doesn't return itself, but goes on.
`render` is called a second time, which leads to the error message.

That's why I suggested to explicitly return 1 or 0.

lib/OpenQA/WebAPI/Controller/API/V1/JobGroup.pm
Wabri: Discussion resolved with https://github.com/os-autoinst/openQA/pull/6522#discussion_r2154795326

templates/webapi/admin/group/group_property_editor.html.ep
```diff
@@ -128,6 +128,24 @@
                     <label for="editor-keep-logs-in-days" class="form-label col-sm-2 control-label" data-bs-toggle="tooltip" title="Number of days to keep logs of jobs">Keep logs for</label>
                     <div class="col-sm-10">
                         <input type="number" min="0" class="form-control" id="editor-keep-logs-in-days" name="default_keep_logs_in_days" value="<%= $group->default_keep_logs_in_days %>"> days
+                          <%= help_popover 'Log vs. Result retention' => '
```
perlpunk: This  popover is only showing in the form for parent groups

templates/webapi/admin/group/group_property_editor.html.ep
```diff
@@ -128,6 +128,24 @@
...
+                            <p>
+                                Specifies the default retention time for logs and results stored by this job group.
+                            </p>
+                            <ul>
+                                <li>
+                                    <code>keep_logs_in_days</code> (or <code>default_keep_logs_in_days</code> for parent groups)
```
perlpunk: I'm wondering if we should rather use the label text here instead of the form names.

templates/webapi/admin/group/group_property_editor.html.ep
```diff
@@ -128,6 +128,24 @@
...
+                                    must be <strong>less than or equal to</strong>
+                                    <code>keep_results_in_days</code> (or <code>default_keep_results_in_days</code>).
+                                </li>
+                                <li>
+                                    This ensures logs are not retained longer than the test results themselves.
+                                </li>
+                                <li>
+                                    A validation error will be shown if this constraint is not satisfied during group creation or update.
```
perlpunk: I think this sentence can be removed. People will see the error when it happens. We don't need to tell them before.

t/api/10-jobgroups.t
```diff
@@ -479,4 +479,65 @@ subtest 'prevent create/update duplicate job group on same parent group' => sub(
     );
 };
 
+subtest 'update default_keep_logs_in_days and default_keep_results_in_days' => sub() {
+    my %tests_paramenters = (
+        'parent group' => {
+            endpoint => '/api/v1/parent_groups',
+            form_name => 'test_parent_group',
+            keep_logs_name => 'default_keep_logs_in_days',
+            keep_return_name => 'default_keep_results_in_days',
```
perlpunk: `return_name`?

perlpunk (changes requested):

lib/OpenQA/WebAPI/Controller/API/V1/JobGroup.pm
```diff
@@ -187,6 +187,30 @@ sub _validate_common_properties ($self) {
     $validation->optional('description');
 }
 
+sub _check_keep_logs_and_results ($self, $properties) {
```
perlpunk: This is still not checking the case when only one of the values is passed. Then we need to compare it to the value already stored in the database.
e.g. if the form  / post parameters only contains `keep_logs_in_days`, we need to get `keep_results_in_days` from the database

templates/webapi/admin/group/group_property_editor.html.ep
Wabri: Yes, I think is more userfriendly. I'll update this.

t/api/10-jobgroups.t
Wabri: Typo :)

lib/OpenQA/WebAPI/Controller/API/V1/JobGroup.pm
Wabri: Yep, you are right! I'll update it!

lib/OpenQA/WebAPI/Controller/API/V1/JobGroup.pm
```diff
@@ -187,6 +187,64 @@ sub _validate_common_properties ($self) {
...
+    #        $self->render(
+    #            json => {error => '`keep_logs_in_days` must be lower than or equal to `keep_results_in_days`'},
+    #            status => 400
+    #        );
+    #        return 0;
+    #    }
+    #}
+    #return 1;
+}
+
```
Wabri: TODO: need to cleanup this mess :)

lib/OpenQA/WebAPI/Controller/API/V1/JobGroup.pm
```diff
@@ -230,6 +288,8 @@ sub create ($self) {
 
     my $properties = $self->load_properties;
 
+    return undef unless $self->_check_keep_logs_and_results($properties);
```
Wabri, line 246: @perlpunk I think since the group is not yet present the `group` argument is not necessary here, am I wrong?

lib/OpenQA/WebAPI/Controller/API/V1/JobGroup.pm
```diff
@@ -187,6 +187,37 @@ sub _validate_common_properties ($self) {
...
+        ($log_key, $result_key) = ('default_keep_logs_in_days', 'default_keep_results_in_days');
+    }
+    else {
+        ($log_key, $result_key) = ('keep_logs_in_days', 'keep_results_in_days');
+    }
+
+    my $log_value
+      = defined $properties->{$log_key}
+      ? $properties->{$log_key}
+      : ($group ? $group->$log_key : 0);
```
perlpunk: Nice, but can be shortened:
```suggestion
    my $log_value = $properties->{$log_key} // ($group ? $group->$log_key : 0);
```

lib/OpenQA/WebAPI/Controller/API/V1/JobGroup.pm
```diff
@@ -187,6 +187,37 @@ sub _validate_common_properties ($self) {
...
 
+sub _check_keep_logs_and_results ($self, $properties, $group = undef) {
+    my ($log_key, $result_key);
+
+    if ($self->is_parent) {
+        ($log_key, $result_key) = ('default_keep_logs_in_days', 'default_keep_results_in_days');
+    }
+    else {
+        ($log_key, $result_key) = ('keep_logs_in_days', 'keep_results_in_days');
+    }
```
perlpunk: Could maybe be shortened like that?
```suggestion
    my $prefix = $self->is_parent ? 'default_' : '';
    my $log_key = $prefix . 'keep_logs_in_days';
    my $result_key = $prefix . 'keep_results_in_days';
```

lib/OpenQA/WebAPI/Controller/API/V1/JobGroup.pm
perlpunk: correct :)

t/api/10-jobgroups.t
```diff
@@ -479,4 +479,65 @@ subtest 'prevent create/update duplicate job group on same parent group' => sub(
...
+            keep_results_name => 'default_keep_results_in_days',
+        },
+        'job group' => {
+            endpoint => '/api/v1/job_groups',
+            form_name => 'test_job_group',
+            keep_logs_name => 'keep_logs_in_days',
+            keep_results_name => 'keep_results_in_days',
+        });
+
+    for my $test (keys %tests_paramenters) {
```
perlpunk: I recommend to always sort hash keys, at least when running tests. It will just make the test output more predictable.
```suggestion
    for my $test (sort keys %tests_paramenters) {
```

t/api/10-jobgroups.t
```diff
@@ -479,4 +479,65 @@ subtest 'prevent create/update duplicate job group on same parent group' => sub(
     );
 };
 
+subtest 'update default_keep_logs_in_days and default_keep_results_in_days' => sub() {
+    my %tests_paramenters = (
+        'parent group' => {
+            endpoint => '/api/v1/parent_groups',
+            form_name => 'test_parent_group',
```
perlpunk: This field is the name of the job group, so I suggest:
```suggestion
            group_name => 'test_parent_group',
```
or even
```suggestion
            name => 'test_parent_group',
```

templates/webapi/admin/group/group_property_editor.html.ep
```diff
@@ -99,6 +99,21 @@
                     <label for="editor-keep-logs-in-days" class="form-label col-sm-2 control-label" data-bs-toggle="tooltip" title="Number of days to keep logs of jobs">Keep logs for</label>
                     <div class="col-sm-10">
                         <input type="number" min="0" class="form-control" id="editor-keep-logs-in-days" name="keep_logs_in_days" value="<%= $group->keep_logs_in_days %>"> days
+                          <%= help_popover 'Log vs. Result retention' => '
```
perlpunk: Since this has the same text in all four places, it can be made a [reusable block](https://docs.mojolicious.org/Mojolicious/Guides/Rendering#Reusable-template-blocks):

Define this before the `if (!$is_parent)`:
```
              <% my $log_retention = begin %>
              <%= help_popover 'Log vs. Result retention' => '
                <p>
                    Specifies the default retention time for logs and results stored by this job group.
                </p>
                <ul>
                    <li>
                        <code>Keep logs for</code>
                        must be <strong>less than or equal to</strong>
                        <code>Keep results for</code>
                    </li>
                    <li>
                        This ensures logs are not retained longer than the test results themselves.
                    </li>
                </ul>
              '; %>
              <% end %>
```
and use it like this:
```
<%= $log_retention->() %>
```

Wabri: Thanks :100:

templates/webapi/admin/group/group_property_editor.html.ep
```diff
@@ -128,6 +147,21 @@
...
+                                <li>
+                                    <code>Keep logs for</code>
+                                    must be <strong>less than or equal to</strong>
+                                    <code>Keep results for</code>
+                                </li>
+                                <li>
+                                    This ensures logs are not retained longer than the test results themselves.
+                                </li>
+                            </ul>
+                        '; %>
```
perlpunk:
```suggestion
                        <%= $log_retention->() %>
```

templates/webapi/admin/group/group_property_editor.html.ep
```diff
@@ -142,6 +176,21 @@
...
```
perlpunk:
```suggestion
                        <%= $log_retention->() %>
```

templates/webapi/admin/group/group_property_editor.html.ep
Wabri: I thought I got them all, fixed!

lib/OpenQA/WebAPI/Controller/API/V1/JobGroup.pm
```diff
@@ -187,6 +187,24 @@ sub _validate_common_properties ($self) {
...
+
+    if ($log_value > $result_value) {
+        $self->render(
+            json => {error => "`$log_key` must be lower than or equal to `$result_key`"},
+            status => 400
+        );
+        return 0;
+    }
+
+    return 1;
```
okurz:
```suggestion
    my $result_value = $properties->{$result_key} // ($group ? $group->$result_key : 0);
    return 1 if $log_value <= $result_value;
    $self->render(json => {error => "`$log_key` must be lower than or equal to `$result_key`"}, status => 400);
    return 0;
```

lib/OpenQA/WebAPI/Controller/API/V1/JobGroup.pm
```diff
@@ -272,6 +292,9 @@ sub update ($self) {
     }
 
     my $properties = $self->load_properties;
+
+    return undef unless $self->_check_keep_logs_and_results($properties, $group);
+
```
okurz:
```suggestion
    return undef unless $self->_check_keep_logs_and_results($properties, $group);
```

lib/OpenQA/WebAPI/Controller/API/V1/JobGroup.pm
```diff
@@ -230,6 +248,8 @@ sub create ($self) {
 
     my $properties = $self->load_properties;
 
+    return undef unless $self->_check_keep_logs_and_results($properties);
+
```
okurz:
```suggestion
```

okurz (changes requested):

t/api/10-jobgroups.t
```diff
@@ -479,4 +479,65 @@ subtest 'prevent create/update duplicate job group on same parent group' => sub(
     );
 };
 
+subtest 'update default_keep_logs_in_days and default_keep_results_in_days' => sub() {
+    my %tests_paramenters = (
```
okurz: s/paramenters/parameters/

t/api/10-jobgroups.t
```diff
@@ -479,4 +479,65 @@ subtest 'prevent create/update duplicate job group on same parent group' => sub(
...
+                form => {
+                    name => $params->{name},
+                    $params->{keep_logs_name} => 2000,
+                    $params->{keep_results_name} => 1,
+                })
+              ->status_is(400,
+                "if $params->{keep_logs_name} lower than or equal to $params->{keep_results_name} create should fail")
+              ->json_is(
+                '/error' => "`$params->{keep_logs_name}` must be lower than or equal to `$params->{keep_results_name}`",
+                'error returned succesfuly'
```
okurz: s/succesfuly/successfully/

t/api/10-jobgroups.t
```diff
@@ -479,4 +479,65 @@ subtest 'prevent create/update duplicate job group on same parent group' => sub(
...
+                form => {
+                    name => $params->{name},
+                    $params->{keep_logs_name} => 2000,
+                    $params->{keep_results_name} => 1,
+                })
+              ->status_is(400,
+                "if $params->{keep_logs_name} lower than or equal to $params->{keep_results_name} update should fail")
+              ->json_is(
+                '/error' => "`$params->{keep_logs_name}` must be lower than or equal to `$params->{keep_results_name}`",
+                'error returned succesfuly'
```
okurz: same as above. But at best write unique test description strings, e.g. "error returned for update"

okurz: You used "error returned for update" but used it also for the "create" case. Please use unique test description strings

Wabri: I've added more context in order to be more clear :+1:

okurz (approved):

PR #6523, opened 2025-06-13
Docs: Convert images/openqa-in-5-minutes.gif to webm

aylen384:
Ticket: https://progress.opensuse.org/issues/183779

### DIFF
--- a/docs/asciinema/Makefile
+++ b/docs/asciinema/Makefile
@@ -1,5 +1,5 @@
 YAML_FILES=$(wildcard *.yaml)
-GIF_FILES=$(YAML_FILES:.yaml=.gif)
+WEBM_FILES=$(YAML_FILES:.yaml=.webm)
 THEME=monokai
 BAT_THEME=Monokai Extended
 
@@ -7,13 +7,16 @@ BAT_THEME=Monokai Extended
 all: build run
 
 .PHONY: cast
-cast: $(GIF_FILES)
+cast: $(WEBM_FILES)
 
 %.cast:%.yaml
 	autocast --overwrite $< $@
 
 %.gif: %.cast
 	agg --theme $(THEME) $< $@
+
+%.webm: %.gif
+	ffmpeg -i $< -c:v libvpx-vp9 -b:v 0 -crf 30 -pix_fmt yuv420p $@
 
 .PHONY: build
 build:
--- a/.gitignore
+++ b/.gitignore
@@ -48,3 +48,4 @@ docs/build/
 docs/vendor/
 # docs
 **/openqa-in-5-minutes.gif
+**/openqa-in-5-minutes.webm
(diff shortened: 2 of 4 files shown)


perlpunk (commented):
How did you convert it? Is there a command that you can share for future images like this?

baierjan (approved):
Yes, it is generated by that makefile. It is not automatically called though (as it would be pointless and expensive to regenerate the "video" at every documentation change. So make sure you generate the referenced webm video and commit it directly to the gh-pages branch as it was done for the gif in #5616

asdil12:
Okay - looks like while there is code (run `make` in `docs/asciinema`) to automatically generate the gif, that code doesn't seem to run as part of the CI.
This was already discussed in https://github.com/os-autoinst/openQA/pull/5616.

So creating a separate PR to manually update the gif to webm on the gh-pages branch:
https://github.com/os-autoinst/openQA/pull/6524

docs/asciinema/Makefile
```diff
@@ -1,20 +1,23 @@
 YAML_FILES=$(wildcard *.yaml)
-GIF_FILES=$(YAML_FILES:.yaml=.gif)
+WEBM_FILES=$(YAML_FILES:.yaml=.webm)
```
perlpunk, line 2: I'm wondering if `.gif` needs to stay there.
It still needs to be generated.

I tried to build the asciinema container locally, but it needs a file that I don't have, so I have no clue how to test that.
It is run in the nightly job on circleci.
```
STEP 9/16: COPY <<EOF /root/.config/browsh/config.toml
Error: building at STEP "COPY <<EOF /root/.config/browsh/config.toml":
 checking on sources under "/path/to/openQA/docs/asciinema":
 copier: stat: "/<<EOF": no such file or directory
```

aylen384: `make cast` is called (by the makefile itself via `make run` via `make all` via `make`.

`make cast` depents on `$(WEBM_FILES)`.
And that depends on `%.gif` for each file and that depends on `%.cast` for each file.
So the gif will still be generated on the way.

baierjan: Yep, there should be a chain .webm <- .gif <- .cast <- .yaml and `make cast` is now supposed to want the webm and start the chain.

baierjan: > Error: building at STEP "COPY <<EOF /root/.config/browsh/config.toml":

Does you "builder" has a support for heredocs in Dockerfile? Builds fine on my TW.

asdil12:
Btw this is one reason why we don't do it nightly:
image

perlpunk:
> Yes, it is generated by that makefile. It is not automatically called though (as it would be pointless and expensive to regenerate the "video" at every documentation change). So make sure you generate the referenced webm video and commit it directly to the gh-pages branch as it was done for the gif in #5616

That should be documented somewhere

docs/asciinema/Makefile
perlpunk: `podman version 4.7.1`

baierjan: Looks like you need at least 4.8 according to https://blog.podman.io/2024/08/heredoc-is-here-for-containerfiles-in-podman/

PR #6529, opened 2025-06-16
Extend single-instance container documentation

realcharmer:
https://progress.opensuse.org/issues/183617

### DIFF
--- a/docs/Installing.asciidoc
+++ b/docs/Installing.asciidoc
@@ -74,11 +74,27 @@ openqa-cli schedule --monitor \
     CASEDIR=https://github.com/os-autoinst/os-autoinst-distri-example.git \
     NEEDLES_DIR=%%CASEDIR%%/needles
 
-openqa-clone-job https://openqa.opensuse.org/tests/1896520
+openqa-clone-job https://openqa.opensuse.org/tests/1
 ----
 
 More details on triggering tests can also be found in the
 <<UsersGuide.asciidoc#usersguide,Users Guide>>.
+
+You can also run the container and directly clone tests, e.g.:
+
+[source,sh]
+----
+podman run -e skip_suse_specifics= -e skip_suse_tests= \
+ --name openqa --device /dev/kvm -p 1080:80 -p 1443:443 --rm \
+ -it registry.opensuse.org/devel/openqa/containers/openqa-single-instance \
+ https://openqa.opensuse.org/tests/1
+----
+
+The `skip_suse_specifics=` and `skip_suse_tests=` options ensure the test files
+and needles required for openSUSE/SUSE tests are downloaded. Refer to the
+<<Installing.asciidoc#_how_to_run_a_test_with_single_instance_container_in_5_minutes,example video above>>
+and the <<GettingStarted.asciidoc#get-testing,get testing>> section for more
+details.
 
 === Separate web UI and worker containers
 As an alternative also separate containers are provided for both the web UI


docs/Installing.asciidoc
```diff
@@ -63,7 +63,21 @@ single-instance container. You can spawn a new shell via:
...
 
-From there, you can trigger a new job or clone an existing one, e.g.:
+From there, you can clone the openSUSE test files:
+
+[source,sh]
+----
+git clone https://github.com/os-autoinst/os-autoinst-distri-opensuse.git /var/lib/openqa/tests/opensuse
+----
+
+Afterwards you can clone an existing job, e.g.:
```
baierjan: Did something changed lately? I am pretty sure you do not need to clone the test files before you clone an existing job, because that's what `openqa-bootstrap` does in the Dockerfile.

realcharmer: `/var/lib/openqa/tests/` is definitely empty after spawning a fresh single-instance container.

realcharmer: Do you think this is a regression?

baierjan: That's why I am asking if something changed.

Looking at the Dockerfile: https://github.com/os-autoinst/openQA/blob/6b21a3a03898709d67fa5ebde8e114b2aa0f26ee/container/single-instance/Dockerfile#L29

and then at the openqa-bootstrap: https://github.com/os-autoinst/openQA/blob/6b21a3a03898709d67fa5ebde8e114b2aa0f26ee/script/openqa-bootstrap#L143

The code is still there, so it might be a change in default settings or a regression.

Martchus: Maybe `openqa-bootstrap` as a generic tool only clones openSUSE tests under certain conditions (e.g. when some environment variable is set). Also our containers are meant to be generic so the absence of openSUSE-specific tests makes actually sense.

By the way, doesn't this also need needles? Maybe it would make sense to add a command for cloning those, too.

Maybe it makes sense to mention the command to start a new job first and only mention openSUSE-specific workflows after that.

baierjan: > when some environment variable is set

I am blind, yes 😄 

https://github.com/os-autoinst/openQA/blob/6b21a3a03898709d67fa5ebde8e114b2aa0f26ee/container/single-instance/Dockerfile#L25

Martchus:
It would also be nice to start the commit message title with a verb in imperative. Otherwise one might be wondering "What about Detailed single-inst…?". Here the verb would probably be "Tweak …" or "Extend …".

docs/Installing.asciidoc
perlpunk: For cloning tests which have a git url in CASEDIR, it is not necessary, as git will be called automatically by openqa.
For os-autoinst-distri-opensuse tests, they need to be cloned (also the needles by the way), but one can influence that in openqa-bootstrap with passing `skip_suse_tests=` (and `skip_suse_specifics=`) so it will be done automatically:

```
% podman run -e skip_suse_specifics= -e skip_suse_tests= --name openqa --device /dev/kvm \
 -p 1080:80 -p 1443:443 --rm -it registry.opensuse.org/devel/openqa/containers/openqa-single-instance
```

baierjan: Just wanted to point out you can use `-e skip_suse_tests=` to enable that feature in the openqa-bootstrap and we have the auto-clone feature but Tina was faster 😄

perlpunk:
https://cbea.ms/git-commit/

docs/Installing.asciidoc
baierjan: The point is, you should be able to schedule a test with a one-liner like (where the env variables might not be necessary depending on the settings of the job you are cloning):

```
podman run --device /dev/kvm -p 1080:80 -p 1443:443 --rm -it \
registry.opensuse.org/devel/openqa/containers/openqa-single-instance \
https://openqa.opensuse.org/tests/5109561
```

realcharmer: I am not following. So do we want to mention to clone the tests and needles in the documentation, or do we want to avoid it?

docs/Installing.asciidoc
```diff
@@ -63,7 +63,21 @@ single-instance container. You can spawn a new shell via:
 podman exec -ti openqa /bin/bash
 ----
 
-From there, you can trigger a new job or clone an existing one, e.g.:
+From there, you can clone the openSUSE test files:
+
+[source,sh]
+----
+git clone https://github.com/os-autoinst/os-autoinst-distri-opensuse.git /var/lib/openqa/tests/opensuse
```
okurz: This is for sure too much openSUSE specific. Here please think as an openQA developer, not SUSE QE engineer.

okurz (changes requested):
The commit message should have more details, e.g. what issue is fixed, why this needs to change, to which versions of which product it applies, link to a bug, the choices you made, etc. Also see https://commit.style/ or http://chris.beams.io/posts/git-commit/ or what perlpunk referenced as a helpful guide how to write good commit messages.

docs/Installing.asciidoc
realcharmer: Indeed that's what I am thinking about. But if we don't mention it there, the provided clone example will result in an error. On the one hand, I am wondering whether we want to change anything  or rather leave it as it is now. On the other hand, leaving it without mentioning the cloning of osado will result in people arriving the the error of missing files.
What would you recommend in this case? I am starting to lean more towards keeping it as-is and having the example cloning really as an *example*.

Martchus: @okurz Would you be ok with my previous propsoal:

> Maybe it makes sense to mention the command to start a new job first and only mention openSUSE-specific workflows after that.

---

I think *if* we add a command for cloning an openSUSE tests we should indeed also add a command for cloning the test code in the first place (and the required needles **which this PR is so far lacking**) as this would not happen automatically.

So I'd say either remove the commands for cloning openSUSE tests completely (1) or add commands for cloning test code and needles (2).

I'm in favor of (2) because leaving out parts many people are surely interested in promotes creating more downstream documentation which is perhaps not what we want.

realcharmer:
I've re-done the documentation to clone both test files and needles using a script `/usr/share/openqa/script/fetchneedles`. The commit message and description was also changed to include more detail.
Please check whether we could improve the wording or if it's fine.

docs/Installing.asciidoc
```diff
@@ -63,7 +63,21 @@ single-instance container. You can spawn a new shell via:
 podman exec -ti openqa /bin/bash
 ----
 
-From there, you can trigger a new job or clone an existing one, e.g.:
+From there, you can schedule or clone jobs. If you wish to test openSUSE, you have to clone the https://github.com/os-autoinst/os-autoinst-distri-opensuse[openSUSE test data]. This can be done easily by running a script, which will clone both test files and needles:
```
okurz: Please wrap consistently at 80 characters

Also I suggest for this section to not mention anything related to osado at all. For that there is http://open.qa/docs/#get-testing

docs/Installing.asciidoc
```diff
@@ -63,7 +63,21 @@ single-instance container. You can spawn a new shell via:
...
+
+[source,sh]
+----
+/usr/share/openqa/script/fetchneedles
+----
```
okurz: That's duplicating http://open.qa/docs/#_getting_tests and shouldn't be done

okurz (changes requested):

docs/Installing.asciidoc
okurz: We already have a separate section for "openSUSE specific tests". I suggest for this section to not mention anything related to osado at all. For that there is http://open.qa/docs/#get-testing . But as necessary you can point to that section

docs/Installing.asciidoc
```diff
@@ -63,7 +63,21 @@ single-instance container. You can spawn a new shell via:
...
+
+Afterwards you can clone an existing job, e.g.:
+
+[source,sh]
+----
+openqa-clone-job https://openqa.opensuse.org/tests/5107001
```
perlpunk: If you refer to a osado test here, then this won't work without manual cloning.
So removing the mention of fetchneedles will result in this example not working.

docs/Installing.asciidoc
realcharmer: I don't see an issue with repeating instructions. Do we expect users to figure everything themselves? I hope not, because that's what omitting this would effectively lead to. This is a separate self-contained topic and should have it's own full and simple instructions on how to run an example job, which is ultimately what most users would try to accomplish, as proven by the [support ticket](https://progress.opensuse.org/issues/183617), where the user would fail precisely because the test data cloning was not mentioned.

I understand you want a strictly structured documentation, but that doesn't mean *useful* or *easy to use* documentation. Not everyone wants to read the entire documentation just to figure out the basics. Giving full information at relevant points, even if repeated, is better than leaving users to fend for themselves.

So the only options are:
1. Leave it as as is; people will get errors when trying out the provided example
2. Remove examples ans osado references entirely
3. Include explicitly both how to clone test data and how to clone a test from o3

I assume Tina's following comment is also related to your suggestion:

>If you refer to a osado test here, then this won't work without manual cloning.
>So removing the mention of fetchneedles will result in this example not working.

realcharmer: As for referring to https://open.qa/docs/#get-testing, this doesn't contain examples of test cloning. This would mean a reader will have to jump to a different section, learn how to clone test data several ways (look for the `fetchneedles` script), come back and run the example clone. Utterly deranged, nobody wants to do this this. Mentioning `fetchneedles` here makes sense, because it's an integral part for the clone example.

realcharmer: If anything, I'd remove example with `openqa-cli schedule`, only include `openqa-clone-job` and refer to https://open.qa/docs/#get-testing at the end of the single-instance container section. So the structure would be:

- Run container
- Get shell access to the container
- Clone test data with `fetchneedles`
- Run a simple `openqa-clone-job`
- Refer to https://open.qa/docs/#get-testing for more info

baierjan: `openqa-cli schedule` is the most useful thing if you just want to run something. For cloning a job you don't even need a shell access at all. If the distri you are using does have `CASEDIR` in git-compatible format (like distri-openqa), it is supposed to work out of the box. If the distri uses just a folder name (like osado), it does need some other way to get the tests/needles and for osado specifically we do have the fetchneedles script which will do the job and which can be executed as a part of the openqa-bootstrap process so you do *not* need to run in manually. Just spawn the podman with the additional environment settings, i.e. with `-e skip_suse_specifics= -e skip_suse_tests=`. Btw, the entrypoint does allow to specify additional parameters which will be used as openqa-clone-job arguments in the bootstrap script, so you can directly clone the job within the podman invocation.

So
```
podman run -e skip_suse_specifics= -e skip_suse_tests= --name openqa --device /dev/kvm \
 -p 1080:80 -p 1443:443 --rm -it registry.opensuse.org/devel/openqa/containers/openqa-single-instance https://openqa.opensuse.org/tests/5107001
```
will prepare the webui and worker, fetch osado tests and needles and clone job from o3 within the one single command.

baierjan: Also this is what we demonstrate in the "Running openQA job within 5 minutes" video in the documentation. No fetchneedles needed there (edit: and fetchneedles is osado specific, so we want to avoid that).

realcharmer: Ok, that makes sense. Thanks @baierjan.
So is this example something that everyone would be happy with? Seems reasonable to me.

okurz: yes, do that

docs/Installing.asciidoc
```diff
@@ -74,12 +74,22 @@ openqa-cli schedule --monitor \
     CASEDIR=https://github.com/os-autoinst/os-autoinst-distri-example.git \
     NEEDLES_DIR=%%CASEDIR%%/needles
 
-openqa-clone-job https://openqa.opensuse.org/tests/1896520
+openqa-clone-job https://openqa.opensuse.org/tests/5107001
```
okurz: all those jobs will not stay forever anyway. We could also make that more clear with obvious "example" URLs like

```suggestion
openqa-clone-job https://openqa.opensuse.org/tests/1
```

docs/Installing.asciidoc
```diff
@@ -74,12 +74,22 @@ openqa-cli schedule --monitor \
...
 ----
 
 More details on triggering tests can also be found in the
 <<UsersGuide.asciidoc#usersguide,Users Guide>>.
 
+You can also run the container and directly clone tests, e.g.:
```
okurz, line 83: Isn't this again a repetition but from http://open.qa/docs/#_how_to_run_a_test_with_single_instance_container_in_5_minutes ? I am ok to mention this but then please explain the "suse" part and refer to http://open.qa/docs/#_how_to_run_a_test_with_single_instance_container_in_5_minutes for details

realcharmer:
Ok, I tried explaining the options and referencing other sections, including the video, without going into detail about osado or other specific topics.

perlpunk (approved):

docs/Installing.asciidoc
```diff
@@ -74,12 +74,28 @@ openqa-cli schedule --monitor \
...
+
+[source,sh]
+----
+podman run -e skip_suse_specifics= -e skip_suse_tests= \
+ --name openqa --device /dev/kvm -p 1080:80 -p 1443:443 --rm \
+ -it registry.opensuse.org/devel/openqa/containers/openqa-single-instance \
+ https://openqa.opensuse.org/tests/1
+----
+
+The `skip_suse_specifics=` and `skip_suse_tests=` options ensure the test files
```
okurz, line 93: How about saying that those are only needed for openSUSE/SUSE tests?

Wabri (approved):

PR #6532, opened 2025-06-16
Avoid adding non-fatal API errors as reason of incomplete jobs

Martchus:
* Track web socket error individually to be able to clear them in case the
  web socket connection could be re-established after all (improving what
  the previous commit has already begun)
* Record only failing API errors of critical requests with no retry
  attempts remaining as last error (which would later potentially be added
  as reason)
* See https://progress.opensuse.org/issues/184022


Martchus:
I'm adding the "not-ready" label because I want to put some more though into whether this is the right approach. Maybe this can and should be solved more elegantly.

okurz (approved):

PR #6536, opened 2025-06-19
Refactor existing Helm Charts and README

realcharmer:
https://progress.opensuse.org/issues/183608

This change improves existing Helm Charts, mainly:

- Improve configuration options and structure
- Update README.md
- Add missing API keys

### DIFF
--- a/container/helm/README.md
+++ b/container/helm/README.md
@@ -1,46 +1,35 @@
-### Helm Chart for openQA inside Kubernetes
+# Helm Chart for openQA inside Kubernetes
 
 Prerequisites:
-1. A Kubernetes cluster (for example _k3s_ will do fine)
+
+1. A Kubernetes cluster (for example [k3s](https://docs.k3s.io/)
 2. Installed and configured Helm
 
-For more information please consult corresponding documentation for [k3s](https://rancher.com/docs/k3s/latest/en/) or [Helm](https://helm.sh/docs).
+For more information, please consult the [Helm
+documentation](https://helm.sh/docs).
 
-The chart consists of two separate sub-charts: _worker_ and _webui_. To install the chart simply execute `helm install openqa .` from this directory. To uninstall and start over, type `helm uninstall openqa`.
+The chart consists of two separate sub-charts, _worker_ and _webui_, and
+a parent chart, _openqa_.
 
-It might be necessary to customize the charts by overriding some of the variables inside _values.yaml_ to suit your needs.
+## Installation
 
-#### Worker
-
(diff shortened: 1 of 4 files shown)


container/helm/charts/worker/values.yaml
```diff
@@ -1,15 +1,12 @@
...
   host: openqa.host
-  key: 123456789ABCDEF
-  secret: 123456789ABCDEF
-
+  key: 1234567890ABCDEF
+  secret: 1234567890ABCDEF
 cacheService: false
-cacheLimit: 50
+cacheLimit: 20
+workerClass: qemu_x86_64,kubernetes
```
Wabri, line 12: Do we need this?

realcharmer: Well, it's a valid option, since it's used in `charts/worker/templates/configmap.yaml` to set `WORKER_CLASS`. I think having this here as a default is at least a good example.

Wabri (approved):

okurz:
@baierjan as former author of the changed files I would appreciate your review feedback

container/helm/README.md
```diff
@@ -1,46 +1,35 @@
-### Helm Chart for openQA inside Kubernetes
+# Helm Chart for openQA inside Kubernetes
 
 Prerequisites:
-1. A Kubernetes cluster (for example _k3s_ will do fine)
-2. Installed and configured Helm
-
-For more information please consult corresponding documentation for [k3s](https://rancher.com/docs/k3s/latest/en/) or [Helm](https://helm.sh/docs).
```
okurz, line 7: Why remove the helpful pointers to k3s?

okurz (changes requested):

container/helm/README.md
realcharmer: I thought it'd be better if it didn't mention specific technologies, but be more general instead, as the charts are independent of the cluster implementation. But we can keep it.

baierjan: btw., k3s is mentioned not as a randomly chosen technology but as "our" technology (or our product if you wish) so in my opinion the "for example k3s will do fine" statement should be read in the same way as saying "you can install openQA on Tumbleweed"

baierjan (commented):
I am fine with those changes (I left a comment about one particular detail but do not have a strong opinion on that).

I believe Emil that those changes are in sync with the new features and behavior in the current Helm version.

container/helm/README.md
realcharmer: I see, that makes sense. I'll add a mention of *k3s* back in.

okurz (approved):

PR #6546, opened 2025-06-25
configure_web_proxy: deal with Debian configuration

phil-hands:
Debian uses `.../sites-available` paths rather than `.../vhosts.d` for storing virtual host configurations.

This PR adds the ability to the `configure_web_proxy` script to use the location of the template file to detect which style of distro. the script is dealing with, and then to fit in with the local standards.

This means that the script behaves _slightly_ differently on Debian systems, since creating the configuration in `.../sites-available` doesn't actually activate it, so the user will also need to do e.g. `a2ensite openqa.conf` -- that's something that we'll want to document for Debian, but I've not quite worked out the best place to do that yet.


script/configure-web-proxy
```diff
@@ -39,20 +39,32 @@ done
 
 echo -e "[openid]\nhttpsonly = 0" > /etc/openqa/openqa.ini.d/01-enable-http-auth.ini
 
+etc_proxy=/etc/${web_proxy/apache/apache2}
+for v in vhosts.d sites-available; do
+    if [[ -e "${etc_proxy}/${v}/openqa.conf.template" ]]; then
+        vhosts_dir="${etc_proxy}/${v}"
```
Martchus, line 45:
```suggestion
        vhosts_dir="${etc_proxy}/${v}"
        break
```

phil-hands: Ah, I see there's a "Commit Suggestion" button -- never mind, did that by hand locally instead :-)

PR #6553, opened 2025-06-30
Fix pagination in server-side tables

Martchus:
Ensure that rows are always sorted by ID, at least as the last ordering criteria. Otherwise, when the main ordering criteria is ambiguous across different pages¹ it is up to chance on what page those rows will appear. So the same row might appear on e.g. page 1 and then page 2 again and some rows might not appear on any page.

So this fixes the second problem mentioned in
https://progress.opensuse.org/issues/184765#note-6 (but not the whole issue).

---

¹ e.g. when sorting needles by `last_seen_time` and a bunch of needles have
  the same value for this column


b10n1k:
Can you update the commit title? it has `in` twice

okurz (approved):

PR #6555, opened 2025-06-30
Avoid creating stale needle entries in needle editor when symlinks used

Martchus:
When the path of a needle directory contains a symlink we generally still always refer to the needle under its real path when updating the needles database table.

The only exception (I am aware of so far) is the needle editor which will create needle entries in the database for symlinked locations. Those will never be updated (e.g. `last_seen_time` will never be set) as the code updating needles always updates the entries for the needles under the real path. This change removes this exception and thus avoids creating entries which are never updated.

This is related to https://progress.opensuse.org/issues/184765 as those entries disturb the sort order.

### DIFF
--- a/t/16-utils-runcmd.t
+++ b/t/16-utils-runcmd.t
@@ -286,12 +286,12 @@ subtest 'saving needle via Git' => sub {
     is_deeply(
         \@executed_commands,
         [
-            [qw(git -C), $empty_tmp_dir, qw(remote update origin)],
-            [qw(git -C), $empty_tmp_dir, qw(reset --hard HEAD)],
-            [qw(git -C), $empty_tmp_dir, qw(rebase origin/master)],
-            [qw(git -C), $empty_tmp_dir, qw(add), "foo.json", "foo.png"],
+            [qw(git -C), "$empty_tmp_dir", qw(remote update origin)],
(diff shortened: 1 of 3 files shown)


okurz (approved):

kalikiana (approved):

t/16-utils-runcmd.t
```diff
@@ -286,12 +286,12 @@ subtest 'saving needle via Git' => sub {
     is_deeply(
         \@executed_commands,
         [
-            [qw(git -C), $empty_tmp_dir, qw(remote update origin)],
-            [qw(git -C), $empty_tmp_dir, qw(reset --hard HEAD)],
-            [qw(git -C), $empty_tmp_dir, qw(rebase origin/master)],
-            [qw(git -C), $empty_tmp_dir, qw(add), "foo.json", "foo.png"],
+            [qw(git -C), "$empty_tmp_dir", qw(remote update origin)],
```
perlpunk, line 289: why did you add quotes?

Martchus: Because this argument was the passed-though object returned by `tempdir` before my change and now the argument is a string. Apparently the check doesn't automatically convert this object to a string and instead fails.

PR #6556, opened 2025-06-30
Avoid showing confusing data in the needles table

Martchus:
* Do not "fill in" information that is missing in the database on the fly just before displaying the needles table as this breaks sorting and filtering which is done before and only based on data from the database
* Simplify the controller code greatly

The disadvantage is that needle entries under symlinked locations will not have their "Last Use" and "Last Match" shown. Ideally those entries don't exist in the first place, though. (And we should make sure those entries also don't exist for real.)

Related ticket: https://progress.opensuse.org/issues/184765


Martchus:
It makes perhaps sense to merge https://github.com/os-autoinst/openQA/pull/6555 first to avoid one source of "needle entries under symlinked locations" as mentioned in the commit message. I am also thinking of creating a PR that would get rid of existing entries under symlinked locations. Then we could deploy both and if no further entries are created again also merge this PR.

We could of course also merge this PR now. The only disadvantage is the one mentioned in the commit message. Not sure whether the current state or the state with this PR is less confusing.

okurz:
#6555 is merged

okurz:
> We are back to the problem of https://progress.opensuse.org/issues/53762.
> 
> > or did you change behaviour to just not show anything in the table rows for needles from symlinked folders?
> 
> I cannot easily do that because for this we need to filter out rows for symlinked folders which isn't possible because this information isn't in the database.
> 
> So the solution is to avoid these entries in the first place. With #6555 we avoid one source of these entries and I'm currently not sure where else they would be created. I am currently also working on a change to remove existing symlinked entries. Maybe with those two changes we would be good regarding https://progress.opensuse.org/issues/53762 in the sense that there simply won't be those problematic rows anymore to begin with.
> 
> EDIT: This is the PR to remove existing symlinked entries: #6561

I approved #6561 . I am not sure about conflicting changes. Anyway I suggest to have #6561 merged first and then continue here … which I have also already approved :)

PR #6558, opened 2025-06-30
openqa-bootstrap: Allow curl to follow redirects (boo#1245379)

nicksinger:
We're mainly interested in the http response code but want to follow eventual redirects. To follow these redirects we can use curls `--location`/`-L`-option but this will cause multiple responses (until no redirect is left). Since `--head`/`-I` will always print all headers for all responses, we now need to redirect *everything* to /dev/null and print the actual, last response code with `--write-out`/`-w` only.

### DIFF
--- a/script/openqa-bootstrap
+++ b/script/openqa-bootstrap
@@ -163,7 +163,7 @@ grep -q "$(hostname)" /etc/hosts || echo "127.0.0.1 $(hostname)" >> /etc/hosts
 start-daemons
 
 # wait for webui to become available
-while ! curl -sI http://localhost/ | grep 200; do
+while ! curl -o /dev/null -w "%{http_code}" -sIL http://localhost/ | grep 200; do
     sleep 3
 done
 


okurz (approved):

nicksinger:
I see `[warn] [pid:19] Failed to register at openqa.host - connection error: Can't connect: Name or service not known - trying again in 10 seconds` in https://github.com/os-autoinst/openQA/actions/runs/15977674139/job/45064266862?pr=6558#step:7:1171 and many others described in https://progress.opensuse.org/issues/183056#note-14. Also found https://progress.opensuse.org/issues/183608 - is this fail here related to these tickets or did I break something? Do we cover the bootstrap-script in any other tests?

okurz:
> I see `[warn] [pid:19] Failed to register at openqa.host - connection error: Can't connect: Name or service not known - trying again in 10 seconds` in https://github.com/os-autoinst/openQA/actions/runs/15977674139/job/45064266862?pr=6558#step:7:1171 and many others described in https://progress.opensuse.org/issues/183056#note-14. Also found https://progress.opensuse.org/issues/183608 - is this fail here related to these tickets or did I break something?

how did you break something before merge?

PR #6559, opened 2025-07-01
create_admin: fix for POO #179359 changes

AdamWill:
Since ff06373beb I don't think create_admin can ever work. It does not specify `from_script` when getting an OpenQA::Schema instance, so Schema's `connect_db` will try and do `OpenQA::App->singleton->home`, and that's never going to work because we don't really have an app singleton (nothing is ever going to have called `set_singleton` on this path, I don't think).

As a script, create_admin should clearly be using the from_script path, so let's change the way it gets a Schema instance to match all the other scripts and fix the problem.

### DIFF
--- a/script/create_admin
+++ b/script/create_admin
@@ -63,7 +63,7 @@ unless ($key) {
     print "Secret: $secret\n";
 }
 
-my $schema = OpenQA::Schema->singleton;
+my $schema = OpenQA::Schema::connect_db(deploy => 0, silent => 1, from_script => 1);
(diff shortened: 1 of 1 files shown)


AdamWill:
I tested this works on our staging server, where it does not work without this change. There doesn't appear to be any test coverage, which is probably why nobody noticed this had got broken.

Martchus:
No need to block this on a review from @kraih. I think this change makes sense. Automatic merging is prevented due to lack of test coverage. In my opinion we can merge this trivial change manually, though.

PR #6560, opened 2025-07-01
Convert build version sorting type from bool to int

aylen384:
This allows adding additional sorting options
(e.g. by oldest job per build)

Ticket: https://progress.opensuse.org/issues/90545


templates/webapi/admin/group/group_property_editor.html.ep
```diff
@@ -83,15 +83,21 @@
...
-                        <label class="form-label" for="editor-build-version-sort-no" data-bs-toggle="tooltip" title="Sort builds by the time a job was most recently created for each build (choose this if the build values do not sort properly as version numbers)">Sort by time job most recently created</label>
+                        <label class="form-label" for="editor-build-version-sort-recent" data-bs-toggle="tooltip" title="Sort builds by the time a job was most recently created for each build (choose this if the build values do not sort properly as version numbers)">Sort by time job most recently created</label>
+
+                        <br>
+                        <input type="radio" id="editor-build-version-sort-oldest" name="build_version_sort" value="2"
+                            %= 'checked' if $group->build_version_sort == 2
+                            >
+                        <label class="form-label" for="editor-build-version-sort-oldest" data-bs-toggle="tooltip" title="Sort builds by the time the oldest job was created for each build (choose this if the build values do not sort properly as version numbers)">Sort by time job least recently created</label>
```
Martchus: I assume this third option hasn't actually been implemented yet. So I suggest you leave this change out for now (and maybe also change the range in the validation to 0 and 1 only). This should instead be part of the commit that actually implements this option.

Martchus (changes requested):

lib/OpenQA/BuildResults.pm
```diff
@@ -131,13 +131,24 @@ sub compute_build_results ($group, $limit, $time_limit_days, $tags, $subgroup_fi
         return \%result;
     }
 
+    # build sorting
+    # 1: sort by build name (default)
+    # 0: sort by newest job in build
+    # 2: sort by oldest job in build
+    my $buildver_sort_mode = 1;
+    $buildver_sort_mode = $group->build_version_sort if $group->can('build_version_sort');
```
perlpunk, line 137: I thin the `can` can actually be removed (in lib/OpenQA/BuildResults.pm as well) as `build_version_sort` was added to parent groups in bca3b925c16bf0542a28e6b725b973d5a05fd32d and the original code was added before that in 68aee72211c63425eec826ce297c6ad27d3a14d9

But maybe it's better to do that in a separate PR.

lib/OpenQA/BuildResults.pm
```diff
@@ -131,13 +131,24 @@ sub compute_build_results ($group, $limit, $time_limit_days, $tags, $subgroup_fi
...
+    my $sort_column = 'lasted_job';
```
perlpunk: Can we fix the typo as well? `s/lasted_job/latest_job/`

lib/OpenQA/BuildResults.pm
```diff
@@ -131,13 +131,24 @@ sub compute_build_results ($group, $limit, $time_limit_days, $tags, $subgroup_fi
...
+    if ($buildver_sort_mode == 2) {
+        $sort_column = 'oldest_job';
+    }
+
     # 400 is the max. limit selectable in the group overview
     my $row_limit = (defined($limit) && $limit > 400) ? $limit : 400;
     my @search_cols = qw(VERSION BUILD);
     my %search_opts = (
-        select => [@search_cols, {max => 'id', -as => 'lasted_job'}],
+        select => [@search_cols, {max => 'id', -as => 'lasted_job'}, {min => 'id', -as => 'oldest_job'}],
```
Martchus: "lasted" is probably a typo and should be "latest". And "latest" would go with "earliest" and not "oldest".

We could also use the pair "newest_job" and "oldest_job".

lib/OpenQA/BuildResults.pm
```diff
@@ -131,13 +131,24 @@ sub compute_build_results ($group, $limit, $time_limit_days, $tags, $subgroup_fi
         return \%result;
     }
 
+    # build sorting
+    # 1: sort by build name (default)
+    # 0: sort by newest job in build
+    # 2: sort by oldest job in build
```
Martchus: These numbers should be constants and with those constants you can remove this comment.

lib/OpenQA/BuildResults.pm
```diff
@@ -131,13 +131,24 @@ sub compute_build_results ($group, $limit, $time_limit_days, $tags, $subgroup_fi
...
+    my $buildver_sort_mode = 1;
+    $buildver_sort_mode = $group->build_version_sort if $group->can('build_version_sort');
+    my $sort_column = 'lasted_job';
+    if ($buildver_sort_mode == 2) {
+        $sort_column = 'oldest_job';
+    }
```
Martchus: This could be more nicely assigned using the ternary operator.

```suggestion
    my $sort_column = $buildver_sort_mode == 2 ? 'oldest_job' : 'lasted_job';
```

lib/OpenQA/BuildResults.pm
```diff
@@ -169,12 +180,8 @@ sub compute_build_results ($group, $limit, $time_limit_days, $tags, $subgroup_fi
         $build->{key} = join('-', $version, $buildnr);
         $versions_per_build{$buildnr}->{$version} = 1;
     }
-    # sort by treating the key as a version number, if job group
-    # indicates this is OK (the default). otherwise, list remains
-    # sorted on the most recent job for each build
-    my $versort = 1;
-    $versort = $group->build_version_sort if $group->can('build_version_sort');
-    if ($versort) {
+    # build sorting by name
```
Martchus: If you use a constant then this comment can be removed.

templates/webapi/admin/group/group_property_editor.html.ep
```diff
@@ -83,15 +83,21 @@
...
-                            %= 'checked' if $group->build_version_sort
+                        <input type="radio" id="editor-build-version-sort-name" name="build_version_sort" value="1"
+                            %= 'checked' if $group->build_version_sort == 1
                             >
-                        <label class="form-label" for="editor-build-version-sort-yes" data-bs-toggle="tooltip" title="Treat the build values as version numbers and sort based on those">Sort by build value (as a version)</label>
+                        <label class="form-label" for="editor-build-version-sort-name" data-bs-toggle="tooltip" title="Treat the build values as version numbers and sort based on those">Sort by build value (as a version)</label>
                         <br>
-                        <input type="radio" id="editor-build-version-sort-no" name="build_version_sort" value="0"
-                            %= 'checked' if !$group->build_version_sort
+                        <input type="radio" id="editor-build-version-sort-recent" name="build_version_sort" value="0"
```
Martchus: "recent" should be replaced with "latest" or "newest" in consistency with how you decide on the wording in controller code.

lib/OpenQA/BuildResults.pm
Martchus: I'm wondering whether we should only specify on of these hashes depending on which one we need.

aylen384: We already get them even if we later re-sort by name.

t/22-dashboard.t
```diff
@@ -501,6 +501,11 @@ subtest 'proper build sorting for dotted build number' => sub {
     $group->update({build_version_sort => 0});
     @build_names = reverse @build_names;
     check_builds(\@build_names, $group, 'builds shown sorted by dotted number');
+
+    # sorted by oldest job per build (which is the same here as there is no overlap)
+    $group->update({build_version_sort => 2});
+    check_builds(\@build_names, $group, 'builds shown sorted by dotted number');
+
```
Martchus: Those tests should use the newly introduced constants as well.

t/api/10-jobgroups.t
```diff
@@ -286,7 +286,7 @@ subtest 'update job group' => sub() {
         form => {
             name => 'opensuse',
             size_limit_gb => 101,
-            build_version_sort => 0,
+            build_version_sort => 2,
```
Martchus: Also this should use the constant.

lib/OpenQA/Schema/Result/JobGroupParents.pm
```diff
@@ -61,7 +61,7 @@ __PACKAGE__->add_columns(
         is_nullable => 1,
     },
     build_version_sort => {
-        data_type => 'boolean',
+        data_type => 'integer',
         default_value => 1,
```
Martchus: Also this should use the constant.

lib/OpenQA/Schema/Result/JobGroups.pm
```diff
@@ -75,7 +75,7 @@ __PACKAGE__->add_columns(
         is_nullable => 1,
     },
     build_version_sort => {
-        data_type => 'boolean',
+        data_type => 'integer',
         default_value => 1,
```
Martchus: Also this should use the constant.

t/22-dashboard.t
```diff
@@ -501,6 +501,11 @@ subtest 'proper build sorting for dotted build number' => sub {
     $group->update({build_version_sort => 0});
     @build_names = reverse @build_names;
     check_builds(\@build_names, $group, 'builds shown sorted by dotted number');
+
+    # sorted by oldest job per build (which is the same here as there is no overlap)
+    $group->update({build_version_sort => 2});
+    check_builds(\@build_names, $group, 'builds shown sorted by dotted number');
```
Martchus: It is not nice to repeat the same check-description (even though this is already the case with the previous tests).

It is also looks wrong that the expected `@build_names` are the same as for the previous case.

Martchus (changes requested):

t/22-dashboard.t
okurz: > It is not nice to repeat the same check-description (even though this is already the case with the previous tests).
means we clean this up as soon as we see it and always leave the campground a bit cleaner than we found it :)

Martchus (approved):

PR #6566, opened 2025-07-04
Append worker.ini configuration to avoid warning

Wabri:
This is more a mitigation than a real resolution, I know not the most beautiful resolution of the problem, but it work!

Related to: https://progress.opensuse.org/issues/181409

This is the output without the warning:

```
+ su _openqa-worker -c '/usr/share/openqa/script/worker --instance 1'
[info] worker 1:
 - config file:                      /etc/openqa/workers.ini
 - name used to register:            4419550c1aab
 - web UI hosts:                     http://localhost:9526
[info] Project dir for host http://localhost:9526 is /var/lib/openqa/share
[info] Registering with openQA http://localhost:9526
[info] Establishing ws connection via ws://localhost:9526/api/v1/ws/1
[info] Registered and connected via websockets with openQA host http://localhost:9526 and worker ID 1
```

This solve both AC1 and AC2 so the ticket could be closed, but I'm looking to find a better resolution.

### DIFF
--- a/container/single-instance/Dockerfile
+++ b/container/single-instance/Dockerfile
@@ -20,6 +20,11 @@ RUN zypper in -y openQA-single-instance openQA-bootstrap \
     qemu-arm qemu-ppc qemu-x86 qemu-tools sudo iputils os-autoinst-distri-opensuse-deps \
     qemu-hw-display-virtio-gpu qemu-hw-display-virtio-gpu-pci && \
     zypper clean -a
+RUN cat > /etc/openqa/workers.ini <<EOF
+[global]
+HOST = http://localhost:9526
+EOF
+
 EXPOSE 80 443 9526
 ENV skip_suse_specifics=1
 ENV skip_suse_tests=1


Martchus (commented):
We can make it so at least one config file is expected. This has the advantage that errors trying to supply a config file can be spotted easily and we probably don't have to change much except of what you're doing here. 

We can also make the error go away at its source. The code change for this would probably be fairly simple, too. The disadvantage is of course worse feedback for a configuration (or rather lack of configuration) that is probably not sufficient/wanted by most users. Probably not such a big deal so maybe actually the way to go forward. 

We can also keep the existing behavior but log it only add info message and perhaps also change the wording a bit. This would probably be the best solution but likely also requires the most code changes.

okurz:
> We can make it so at least one config file is expected. This has the advantage that errors trying to supply a config file can be spotted easily and we probably don't have to change much except of what you're doing here.
> 
> We can also make the error go away at its source. The code change for this would probably be fairly simple, too. The disadvantage is of course worse feedback for a configuration (or rather lack of configuration) that is probably not sufficient/wanted by most users. Probably not such a big deal so maybe actually the way to go forward.
> 
> We can also keep the existing behavior but log it only as info message and perhaps also change the wording a bit. This would probably be the best solution but likely also requires the most code changes.

Yes, I prefer your suggestion to log only as "info". Why that should be the "most code changes" I don't understand.

container/single-instance/Dockerfile
```diff
@@ -20,6 +20,11 @@ RUN zypper in -y openQA-single-instance openQA-bootstrap \
     qemu-arm qemu-ppc qemu-x86 qemu-tools sudo iputils os-autoinst-distri-opensuse-deps \
     qemu-hw-display-virtio-gpu qemu-hw-display-virtio-gpu-pci && \
     zypper clean -a
+RUN cat > /etc/openqa/workers.ini <<EOF
```
d3flex, line 23: why not provide a workers.ini which you CP in the container?

okurz: Let's come to a conclusion in #6568 first. Added back "not-ready" so that we know where to target the discussion

PR #6567, opened 2025-07-04
Fix redundant bash flags and clean up shebang

Wabri:


script/openqa-bootstrap
```diff
@@ -1,4 +1,4 @@
-#!/bin/bash -e
+#!/usr/bin/env bash
```
baierjan: Afaik `/usr/bin/env` should be avoided and rpmlint will complain about this, see https://lists.opensuse.org/archives/list/packaging@lists.opensuse.org/thread/52FEF7ZZGAVXOOAPY2BKRVMC76JR4A7N/

Wabri: Oh seriously? I'm using only that in my scripts 🤣 I read the emails, but still not understand the reason why not use for bash. Python is a completely different environment that can have multiple versions and executable, but bash is bash. Maybe we should avoid for python or perl, even if I saw other scripts that have the same shebang `/usr/bin/env perl` for example.

baierjan: From what I understand, some tools cannot properly detect the interpreter in this way. Also if you use `/bin/bash` it is clear what you want; `/usr/bin/env bash` can execute something completely different as it takes `$PATH` into account, that's probably the biggest difference.

Martchus (changes requested):
The commit message and the actual change are not in line with each other. The actual change is questionable.

okurz (changes requested):
I think it's still a safer and actually common approach to keep `-e` hence I suggest to not change the current script

Wabri:
@okurz safer for what? we already have literally the same setting in line 4

Martchus:
I would also say having the setting in line 4 is enough. Otherwise we have two places where we specify the same flag which calls for introducing inconsistencies.

Martchus (approved):

b10n1k (approved):

Martchus:
Maybe then remove `-e` in the shebang of other files, too. (And add it within the actual script if not present yet.)

perlpunk:
Note that `-e` in the shebang has no effect if you are sourcing the script (something we do in unit tests for example).
So we probably always should use `set -e` instead anyway.

Wabri:
@perlpunk based on this we should change all scripts by moving the assignment of `-e` from the shebang

PR #6568, opened 2025-07-04
Prevent error when workers config not found, throw warning instead

Wabri:
Related: https://progress.opensuse.org/issues/181409

This is still a work in progress, the purpose of this is to replace the error with a warning since the instance even without the definition of the configuration is working.


lib/OpenQA/Worker/Settings.pm
```diff
@@ -23,7 +23,8 @@ use constant VNCPORT_OFFSET => $ENV{VNCPORT_OFFSET} // 90;
 sub new ($class, $instance_number = undef, $cli_options = {}) {
     my $config_paths = lookup_config_files(undef, 'workers.ini', 1);
     my $cfg = parse_config_files($config_paths);
-    my @parse_errors = @$config_paths ? (@Config::IniFiles::errors) : ('No config file found.');
+    my @parse_errors = @$config_paths ? (@Config::IniFiles::errors) : ();
+    log_warning "[OpenQA::Worker::Settings] workers.ini not found, using default settings\n" unless @$config_paths;
```
okurz: I would use level "info" as in: nothing is wrong but maybe noteworthy

Martchus:
```suggestion
    log_info "workers.ini not found, using default settings" unless @$config_paths;
```

Martchus:
I'm afraid you'll need to restructure the corresponding test:

```
    #   Failed test 'error logged'
    #   at t/24-worker-settings.t line 130.
    #          $got->[0] = Does not exist
    #     $expected->[0] = 'No config file found.'
    # []
```

Martchus:
And how does this actually look like as a whole (together with the other startup log messages)? I'm just wondering because you're effectively moving the place where we log this and therefore possibly change the order of startup log messages.

Wabri:
@Martchus finally find the way to spin a single-instance container with local changes and this is the result:

```
+ su _openqa-worker -c '/opt/openqa/script/worker --instance 1'
+ [[ -n '' ]]
+ wait
[INFO] workers.ini not found, using default settings
[info] worker 1:
 - config file:
 - name used to register:            1e039ebbb480
 - web UI hosts:                     localhost
[info] Project dir for host localhost is /var/lib/openqa/share
[info] Registering with openQA localhost
```

The log info is shown as expected!

Martchus:
The inconsistent capitalization of the info messages is not nice but I guess it is acceptable. (I guess I would have made it so it would still log everything were it was logged before and just make the parameter passing between the function calls a bit more involved. I guess your simple approach is ok, too.)

PR #6575, opened 2025-07-08
Allow running single instance container with local changes

Wabri:
This change adds support for running the single-instance container
using local source tree instead of the system zypper package.

this can help testing changes to openQA code inside the container
without requiring package installation.

## TODOs

- [ ] fix error related to not authorized and registration web ui failure
```
   + su _openqa-worker -c '/opt/openqa/script/worker --instance 1'
[INFO] workers.ini not found, using default settings
[info] worker 1:
 - config file:
 - name used to register:            7f7cc3de60bf
 - web UI hosts:                     localhost
[info] Project dir for host localhost is /var/lib/openqa/share
[info] Registering with openQA localhost
[error] Failed to register at localhost - 403 response: Not authorized - ignoring server
[error] Stopping because registration with all configured web UI hosts failed
```
- [ ] documentation

### DIFF
--- a/container/single-instance-devel/Dockerfile
+++ b/container/single-instance-devel/Dockerfile
@@ -1,0 +1,27 @@
+# SPDX-License-Identifier: MIT
+#!BuildTag: openqa-single-instance-devel:latest opensuse/openqa-single-instance-devel:latest opensuse/openqa-single-instance-devel:%PKG_VERSION% opensuse/openqa-single-instance-devel:%PKG_VERSION%.%RELEASE%
+
+# hadolint ignore=DL3006
+FROM opensuse/openqa_devel
+
+# labelprefix=org.opensuse.openqa-single-instance-devel
+LABEL org.opencontainers.image.title="openQA developer single-instance container"
+LABEL org.opencontainers.image.description="A complete openQA developer instance composed of all necessary components to execute openQA tests including an openQA worker"
+LABEL org.opencontainers.image.version="%PKG_VERSION%.%RELEASE%"
+LABEL org.opensuse.reference="registry.opensuse.org/opensuse/openqa-single-instance-devel:%PKG_VERSION%.%RELEASE%"
+LABEL org.openbuildservice.disturl="%DISTURL%"
+LABEL org.opencontainers.image.created="%BUILDTIME%"
+# endlabelprefix
+
+COPY . /opt/openqa
+WORKDIR /opt/openqa
+
+RUN make node_modules
+
+EXPOSE 80 443 9526
+ENV skip_suse_specifics=1
+ENV skip_suse_tests=1
+ENV OPENQA_SERVICE_PORT_DELTA=0
+ENV LC_ALL C.UTF-8
+
+ENTRYPOINT ["/opt/openqa/script/openqa-bootstrap"]


script/openqa-bootstrap
```diff
@@ -3,6 +3,8 @@
 
 set -xeuo pipefail
 
+OPENQA_DIR="${OPENQA_DIR:="/usr/share/openqa"}"
```
Martchus: None of these quotes make a difference.
```suggestion
OPENQA_DIR=${OPENQA_DIR:=/usr/share/openqa}
```
(Although admittedly the subsequent assignments use the same style so you can also leave it as-is. But then maybe remove the blank line so these assignments are together.)

---

You need quotes when use *use* the variable, though. And those quotes are missing in the lines where `$OPENQA_DIR` is used below. So please add them in the lines you changed below.

container/single-instance-local/Dockerfile
```diff
@@ -0,0 +1,40 @@
...
+LABEL org.opencontainers.image.created="%BUILDTIME%"
+# endlabelprefix
+
+# installing more of packages that are selected in openqa-bootstrap.  Combining here saves installation time
+# hadolint ignore=DL3037
+RUN zypper in -y retry git make gcc perl \
+    openQA-bootstrap openQA-devel os-autoinst-distri-opensuse-deps \
+    qemu-arm qemu-ppc qemu-x86 qemu-tools sudo iputils\
+    qemu-hw-display-virtio-gpu qemu-hw-display-virtio-gpu-pci
+RUN zypper clean -a
```
Martchus: Multiple consecutive `RUN` commands should be combined. The same counts for the `RUN` commands below.

container/single-instance-local/Dockerfile
```diff
@@ -0,0 +1,40 @@
+# SPDX-License-Identifier: MIT
```
Martchus: There should be some documentation how this container file is meant to be used.

container/single-instance-local/Dockerfile
```diff
@@ -0,0 +1,40 @@
...
+LABEL org.openbuildservice.disturl="%DISTURL%"
+LABEL org.opencontainers.image.created="%BUILDTIME%"
+# endlabelprefix
+
+# installing more of packages that are selected in openqa-bootstrap.  Combining here saves installation time
+# hadolint ignore=DL3037
+RUN zypper in -y retry git make gcc perl \
+    openQA-bootstrap openQA-devel os-autoinst-distri-opensuse-deps \
+    qemu-arm qemu-ppc qemu-x86 qemu-tools sudo iputils\
+    qemu-hw-display-virtio-gpu qemu-hw-display-virtio-gpu-pci
```
Martchus: I'm wondering whether this list of packages is copied from somewhere and now duplicated.

Martchus (changes requested):

container/single-instance-local/Dockerfile
okurz: Well, mostly it is and I consider this bad as it makes the container file less maintainable.

container/single-instance-local/Dockerfile
```diff
@@ -0,0 +1,40 @@
+# SPDX-License-Identifier: MIT
+#!BuildTag: openqa-single-instance:latest opensuse/openqa-single-instance:latest opensuse/openqa-single-instance:%PKG_VERSION% opensuse/openqa-single-instance:%PKG_VERSION%.%RELEASE%
+
+# hadolint ignore=DL3006
+FROM opensuse/tumbleweed
```
baierjan:
```suggestion
FROM opensuse/openqa_devel
```

container/single-instance-local/Dockerfile
```diff
@@ -0,0 +1,40 @@
...
+
+# installing more of packages that are selected in openqa-bootstrap.  Combining here saves installation time
+# hadolint ignore=DL3037
+RUN zypper in -y retry git make gcc perl \
+    openQA-bootstrap openQA-devel os-autoinst-distri-opensuse-deps \
+    qemu-arm qemu-ppc qemu-x86 qemu-tools sudo iputils\
+    qemu-hw-display-virtio-gpu qemu-hw-display-virtio-gpu-pci
+RUN zypper clean -a
+
+COPY . /opt/openqa
```
baierjan: Wouldn't mounting it as a volume make more sense? That way you can use your standard editing tools to change the code from the outside.

Wabri: Using VOLUME instead of COPY break the ENTRYPOINT because without the script the container cant run due to missing path:

```
$ podman run --rm -it -v .:/opt/openqa -p 9526:80 openqa-test:latest

/bin/bash: /opt/openqa/script/openqa-bootstrap: Permission denied
```

In order to fix this we should use the bootstrap script from zypper package and not the one from root.

container/single-instance-local/Dockerfile
Wabri: Fixed thanks!

container/single-instance-local/Dockerfile
Wabri: Fixed thanks!

container/single-instance-local/Dockerfile
Wabri: Fixed thanks!

script/openqa-bootstrap
Wabri: Fixed thanks!

container/single-instance-local/Dockerfile
Wabri: I'll take a look!

Wabri:
@okurz I've fixed typos on the commit message and for now due to discussion in our slack channel (https://suse.slack.com/archives/C02AJ1E568M/p1752069661892629) I'm using the ticket https://progress.opensuse.org/issues/181409

script/openqa-bootstrap
Martchus: The usages of the variable are still not quoted.

Wabri: Oh yeah miss that point :pray: Now fixed

script/openqa-bootstrap
```diff
@@ -101,12 +103,12 @@ retry -e -s 30 -r 7 -- sh -c "zypper -n --gpg-auto-import-keys ref && zypper -n
...
 chown -R postgres: /var/lib/pgsql/ # fix broken postgres working dir permissions in the nspawn container
 start-database
-su postgres -c "/usr/share/openqa/script/setup-db" "$dbuser" "$dbname"
+su postgres -c "$OPENQA_DIR/script/setup-db" "$dbuser" "$dbname"
 
 # setup webserver and fake-auth
 proxy_args=""
 [[ -n "$setup_web_proxy" ]] && proxy_args="--proxy=$setup_web_proxy"
-setup=/usr/share/openqa/script/configure-web-proxy
+setup="$OPENQA_DIR/script/configure-web-proxy"
```
Martchus: You didn't have to use quotes on *that* use of the variable. (Expansion is not happening on assignments, so quoting is not needed here.)

You don't have to change this single line now, I'm just mentioning it for a better understanding.

script/openqa-bootstrap
okurz, line 1: The changes in this file make sense independently of the Dockerfile changes. I suggest you create a separate commit out of it, possibly a separate PR which we can merge easily before we are able to finish the discussion on the Dockerfile :)

Wabri: I'm going to split it 👍

Wabri: Moved here #6582

container/single-instance-devel/Dockerfile
```diff
@@ -0,0 +1,27 @@
...
+# labelprefix=org.opensuse.openqa-single-instance-devel
+LABEL org.opencontainers.image.title="openQA developer single-instance container"
+LABEL org.opencontainers.image.description="A complete openQA developer instance composed of all necessary components to execute openQA tests including an openQA worker"
+LABEL org.opencontainers.image.version="%PKG_VERSION%.%RELEASE%"
+LABEL org.opensuse.reference="registry.opensuse.org/opensuse/openqa-single-instance-devel:%PKG_VERSION%.%RELEASE%"
+LABEL org.openbuildservice.disturl="%DISTURL%"
+LABEL org.opencontainers.image.created="%BUILDTIME%"
+# endlabelprefix
+
+COPY . /opt/openqa
```
d3flex, line 16: does it make sense to be explicit here? I guess what you want to copy is everything from the root openqa tree of the repo. But the Dockerfile is under container/

container/single-instance-local/Dockerfile
okurz: I don't see the purpose of COPY here for development purposes. As Jan said either a volume or just using `-v` should suffice not needing any additional container definition

container/single-instance-devel/Dockerfile
perlpunk: I think when this is built in OBS, there might not even be any files from git available. Did you test this in a branch in OBS?

Wabri: This dockerfile doesn't meant to be build in obs, I'm thinking that it would be better to create a docker compose instead of this one. something like:

```
services:
  db:
    image: postgres:...

  webui:
    build:
      context: .
      dockerfile: .../webui/Dockerfile
    ports:
      - "9526:9526"
    environment:
      OPENQA_DIR: /opt/openqa
    volumes:
      - .:/opt/openqa
    depends_on:
      - db
    entrypoint: /opt/openqa/script/openqa-bootstrap

# Same for worker
```

container/single-instance-local/Dockerfile
Wabri: I've try to not use the COPY but for how I set up the dockerfile is not possible to avoid it if using the current openqa-bootstrap script as entrypoint.

As said in the comment above (https://github.com/os-autoinst/openQA/pull/6575#discussion_r2198200429) I think to be more clear a docker compose would be better for this case, I just wanted to try to make it work like this.

container/single-instance-devel/Dockerfile
Wabri: now that I think about it, it would be exactly the right thing to create a compose and i could have invested my time better, but at least we/I have the proof that what I wanted to achieve is not doable via dockerfile

PR #6576, opened 2025-07-09
Fix race cond when date ref job got deleted while serving build list

aylen384:
Ticket: https://progress.opensuse.org/issues/185494

### DIFF
--- a/lib/OpenQA/BuildResults.pm
+++ b/lib/OpenQA/BuildResults.pm
@@ -232,7 +232,7 @@ sub compute_build_results ($group, $limit, $time_limit_days, $tags, $subgroup_fi
         unless (defined $jr{date}) {
             # job was not in @jobs - so fetch it from db
             my $job = $jobs_resultset->find($date_ref_job);
-            $jr{date} = $job->t_created;
+            $jr{date} = (defined $job) ? $job->t_created : DateTime->from_epoch(0);
         }
         $jr{escaped_version} = $jr{version};
         $jr{escaped_version} =~ s/\W/_/g;


lib/OpenQA/BuildResults.pm
```diff
@@ -146,6 +146,13 @@ sub compute_build_results ($group, $limit, $time_limit_days, $tags, $subgroup_fi
         order_by => {-desc => $sort_column},
         rows => $row_limit
     );
+    # SELECT VERSION, BUILD, MAX(id) as newest_job, MIN(id) as oldest_job, nj.t_created, oj.t_created FROM jobs
+    # WHERE $search_filter
+    # GROUP BY VERSION, BUILD
+    # ORDER BY $sort_column DESC
+    # LEFT JOIN jobs as nj ON nj.id = newest_job
+    # LEFT JOIN jobs as oj ON oj.id = oldest_job
+    # LIMIT $row_limit
```
Martchus: I also think that SQL is easier to read then the DBIx query. However, I would still avoid duplicated and commented-out code. This SQL is also a bit of a simplification and I think the order in which `WHERE`, `LEFT JOIN` and `ORDER BY` occur here is actually wrong.

Martchus (commented):
The change itself makes sense but I wouldn't add the SQL pseudo code.

lib/OpenQA/BuildResults.pm
aylen384: that comment was actually my mockup of what would could be done and wasn't ment to be included in this PR

PR #6577, opened 2025-07-09
Skip upload of test modules and extra tests with no name

Martchus:
Trying to upload those will lead to errors as openQA requires a name. It is probably better to avoid any fatal errors here so the rest of the results can still be uploaded.

Related ticket: https://progress.opensuse.org/issues/185497

---

Still a draft because I might want to test this for real first.


Martchus:
Tests have changed and therefore this is not reproducible. We should probably nevertheless merge this.

okurz (approved):

PR #6578, opened 2025-07-09
worker: disconnect dbus from NameOwnerChanged signal (POO #183833)

AdamWill:
This avoids us piling up signals we never handle and eventually exceeding a dbus quota limit and causing all subsequent tap jobs to fail.

New Net::DBus instances are automatically connected to this signal on creation, you cannot change this. If the instance is kept around but nothing iterates the dbus event loop once in a while, that signal will just keep piling up messages indefinitely. We encountered this once before in os-autoinst and "solved" it by not persisting the dbus connection, but in this case it doesn't seem a good idea to re-initialize that connection every time we want to check on the status of a tap worker. So instead let's disconnect it from the signal.

The 1 here is a bit magical. The Net::DBus connect_to_signal method issues signal numbers on a first-come, first-served basis, starting at 1 (0 is never used), and returns the number it issued. When Net::DBus _new calls it, though, it doesn't record the return value. So we have to know/guess that this will always be the first signal connected.

Related ticket: https://progress.opensuse.org/issues/183833

### DIFF
--- a/lib/OpenQA/Worker.pm
+++ b/lib/OpenQA/Worker.pm
@@ -611,8 +611,12 @@ sub is_qemu_running ($self) {
 sub is_ovs_dbus_service_running ($self) {
     try { defined &Net::DBus::system or require Net::DBus }
     catch ($e) { return 0 }
-    my $bus = ($self->{_system_dbus} //= Net::DBus->system(nomainloop => 1));
-    try { return defined $bus->get_service('org.opensuse.os_autoinst.switch') }
+    unless (defined $self->{_system_dbus}) {
+        $self->{_system_dbus} = Net::DBus->system(nomainloop => 1);
+        # this avoids piling up signals we never do anything with - POO #183833
+        $self->{_system_dbus}->get_bus_object->disconnect_from_signal('NameOwnerChanged', 1);
+    }
+    try { return defined $self->{_system_dbus}->get_service('org.opensuse.os_autoinst.switch') }
     catch ($e) { return 0 }
 }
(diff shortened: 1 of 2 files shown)


AdamWill:
I'm not 100% sure this is un-hacky enough to upstream, but figured I'd submit it at least. It'd be great if @berrange could comment. Relevant perl-net-dbus code areas:

* [the signal connection in _new](https://gitlab.com/berrange/perl-net-dbus/-/blob/4b700810b6a924da14420f4e3c12e109a7a8228b/lib/Net/DBus.pm#L294)
* [connect_to_signal's handling of signal IDs](https://gitlab.com/berrange/perl-net-dbus/-/blob/4b700810b6a924da14420f4e3c12e109a7a8228b/lib/Net/DBus/RemoteObject.pm#L298) - note nothing else in the codebase touches this `signal_id` property AFAICT, which makes me feel like the assumption that the ID here will always be 1 is pretty solid, unless I'm missing something

berrange:
> I'm not 100% sure this is un-hacky enough to upstream, but figured I'd submit it at least. It'd be great if @berrange could comment. Relevant perl-net-dbus code areas:

Yes it is a bit hacky, but at the same time I think it is likely the least worst option, given the constraints imposed by the perl-net-dbus implementation. Since the perl-net-dbus project is largely dormant, it is unlikely to see new changes that would invalidate the assumption made here.

Martchus (approved):

aylen384 (approved):

PR #6579, opened 2025-07-10
Schema::Result::Jobs: Catch undefined parent jobs

okurz:
Related progress issue: https://progress.opensuse.org/issues/185515

### DIFF
--- a/lib/OpenQA/Schema/Result/Jobs.pm
+++ b/lib/OpenQA/Schema/Result/Jobs.pm
@@ -2237,10 +2237,10 @@ sub dependencies ($self, $children_list = undef, $parents_list = undef) {
     $parents_list ||= [$self->parents->all];
     for my $s (@$parents_list) {
         push(@{$parents{$s->to_string}}, $s->parent_job_id);
-        $has_parents = 1;
         my $jobs = $self->result_source->schema->resultset('Jobs');
-        $parents_ok = $jobs->find($s->parent_job_id, {select => ['result']})->is_ok
-          if $is_final && $parents_ok;
+        next unless my $parent = $jobs->find($s->parent_job_id, {select => ['result']});
+        $has_parents = 1;
+        $parents_ok &&= $parent->is_ok if $is_final;
     }
     $children_list ||= [$self->children->all];
     for my $s (@$children_list) {


aylen384 (approved):

Martchus (changes requested):
I'm wondering whether we should only set `$has_parents = 1;` if we are able to find the parent. Judging from its use in JavaScript we probably should. So maybe do the following:

```
next unless my $parent = $self->result_source->schema->resultset('Jobs')->find($s->parent_job_id, {select => ['result']});
$has_parents = 1;
$parents_ok = $parent->is_ok if $is_final && $parents_ok;

```

okurz:
Changed to use the suggestion my @Martchus in https://github.com/os-autoinst/openQA/pull/6579#pullrequestreview-3004820827

lib/OpenQA/Schema/Result/Jobs.pm
```diff
@@ -2238,9 +2238,8 @@ sub dependencies ($self, $children_list = undef, $parents_list = undef) {
     for my $s (@$parents_list) {
         push(@{$parents{$s->to_string}}, $s->parent_job_id);
         $has_parents = 1;
-        my $jobs = $self->result_source->schema->resultset('Jobs');
-        $parents_ok = $jobs->find($s->parent_job_id, {select => ['result']})->is_ok
-          if $is_final && $parents_ok;
+        my $parent = $self->result_source->schema->resultset('Jobs')->find($s->parent_job_id, {select => ['result']});
+        $parents_ok = defined $parent && $parent->is_ok if $is_final && $parents_ok;
```
d3flex: not sure but the `$parents_ok` seems redundant here.
IIUC $parents_ok is initialized and it is assumed `is_ok`. then for each parent you check if the result still `is_ok`. 
the variable then it should be updated only and when if $is_final.
Why does it need to check again `$parents_ok`?

perlpunk: Unrelated, but I think the intention of considering the old value of `$parents_ok` can be expressed better like this:
```suggestion
        $parents_ok &&= defined $parent && $parent->is_ok if $is_final;
```

Martchus:
It doesn't look like this contains the suggestion from my review.

perlpunk:
> Changed to use the suggestion my @Martchus in [#6579 (review)](https://github.com/os-autoinst/openQA/pull/6579#pullrequestreview-3004820827)

Actually I don't see that change...

b10n1k (approved):

lib/OpenQA/Schema/Result/Jobs.pm
okurz: I didn't question this logic, just introduced a check for definedness

lib/OpenQA/Schema/Result/Jobs.pm
```diff
@@ -2237,10 +2237,10 @@ sub dependencies ($self, $children_list = undef, $parents_list = undef) {
     $parents_list ||= [$self->parents->all];
     for my $s (@$parents_list) {
         push(@{$parents{$s->to_string}}, $s->parent_job_id);
-        $has_parents = 1;
         my $jobs = $self->result_source->schema->resultset('Jobs');
-        $parents_ok = $jobs->find($s->parent_job_id, {select => ['result']})->is_ok
-          if $is_final && $parents_ok;
+        next unless my $parent = $jobs->find($s->parent_job_id, {select => ['result']});
```
perlpunk, line 2241: Should the `push` above happen if the parent is undefined?

perlpunk (changes requested):

lib/OpenQA/Schema/Result/Jobs.pm
okurz: :shrug:

perlpunk (approved):

lib/OpenQA/Schema/Result/Jobs.pm
Martchus: Not sure. It probably doesn't hurt because it'll just not have any effect on the JavaScript side but it would probably make more sense to also avoid the `push`.

perlpunk: I was thinking about it and it could also happen after that loop, that such a job gets deleted, which would have the same effect on the client side, so I guess it's ok

PR #6581, opened 2025-07-10
Improve error handling when handling status updates from worker

Martchus:
The code is already mainly just skipping invalid bits. These changes just make the checks for what's considered invalid a bit stricter to avoid Perl warnings and database errors. See the particular commit messages and https://progress.opensuse.org/issues/185497 for details.


kalikiana (approved):
I guess the change is really making things consistent. Treating incomplete data as errors would be a different discussion (just for completeness since I raised this concern before).

PR #6588, opened 2025-07-15
Make hint about cloning example repo more prominent

perlpunk:
Related issue: https://progress.opensuse.org/issues/184459

I tried to create an example test in the webui, but the existing hint did not make clear to me that it would do something, I just thought it would be a link to the example distribution repo.


perlpunk:
before:
create-example-test
after:
create-example-test-new

perlpunk:
Arguably "CLICK HERE" in caps is not the nicest thing, but I didn't have an idea how to improve it otherwise.
One thing could be to color links in blue boxes something other than blue in general.

perlpunk:
> > Arguably "CLICK HERE" in caps is not the nicest thing, but I didn't have an idea how to improve it otherwise.
> 
> I had the same though. Considering we can render an `a` element here we can probably also render a `button` with the text "Clone". So I would do that. Maybe you can actually keep using an `a` element and just add `class="btn btn-primary"`.

A button, excellent!
Done that

okurz (changes requested):
I don't quite understand the current message. Why is it necessary to click any button to clone a distribution when openQA should clone the test distribution automatically on start? Under which circumstances do we get here?

Martchus:
Maybe "example distribution" could now be a link to GitHub. This way users who want to know more before hitting the "Clone" button could have a look beforehand. Otherwise it is actually still quite unclear what exactly will be cloned.

---

@okurz I think your questions lead to far for a PR that is merely attempting to improve the UI (and not the logic behind). So I wouldn't block the merging of this PR on this discussion.

> … when openQA should clone the test distribution automatically on start?

On the start of *what*? I don't think it is desirable that any openQA deployment automatically/unasked clones some additional repository from GitHub on first launch (if you mean that). We could trigger this automatically when one enters the specific page, though. I personally find it better if the cloning is triggered semi-automatic so users can follow better what's going on and the context of a potential error showing up is clearer.

> Under which circumstances do we get here?

By clicking on "Create … Example tests" on a freshly installed openQA instance where our example distribution hasn't been cloned yet. It is required to load the scenario definitions from there.

okurz:
> > … when openQA should clone the test distribution automatically on start?
> 
> On the start of _what_? I don't think it is desirable that any openQA deployment automatically/unasked clones some additional repository from GitHub on first launch (if you mean that). We could trigger this automatically when one enters the specific page, though. I personally find it better if the cloning is triggered semi-automatic so users can follow better what's going on and the context of a potential error showing up is clearer.
> 
> > Under which circumstances do we get here?
> 
> By clicking on "Create … Example tests" on a freshly installed openQA instance where our example distribution hasn't been cloned yet. It is required to load the scenario definitions from there.

I see. The document is expanded. How about a button inside the scenario definitions text edit field to load the file (not the complete repo) remotely on demand? Like "Click here to fetch example scenario definitions"

PR #6590, opened 2025-07-15
Add instructions for running a openQA/tools dev environment

perlpunk:
Issue: https://progress.opensuse.org/issues/184459

Preview: https://github.com/perlpunk/openQA/blob/devcontainer/docs/Contributing.asciidoc

### DIFF
--- a/docs/Contributing.asciidoc
+++ b/docs/Contributing.asciidoc
@@ -131,10 +131,9 @@ additional tools are required. The easiest way to install all needed
 dependencies is using the available os-autoinst and openQA packages, as
 described in the Installation Guide.
 
-In the case of os-autoinst, only a few http://www.cpan.org/[CPAN] modules are
-required. Basically `Carp::Always`, `Data::Dump`. `JSON` and `YAML`. On the other
-hand, several external tools are needed including
-http://wiki.qemu.org/Main_Page[QEMU],
+In the case of https://github.com/os-autoinst/os-autoinst[os-autoinst], some
+http://www.cpan.org/[CPAN] modules are required. Additionally, several external
+tools are needed including http://wiki.qemu.org/Main_Page[QEMU],
 https://code.google.com/p/tesseract-ocr/[Tesseract] and
 http://optipng.sourceforge.net/[OptiPNG]. Last but not least, the
 http://opencv.org/[OpenCV] library is the core of the openQA image matching
@@ -143,7 +142,7 @@ mechanism, so it must be available on the system.
 The openQA package is built on top of Mojolicious, an excellent Perl framework
 for web development that will be extremely familiar to developers coming from
 other modern web frameworks like Sinatra and that have nice and comprehensive
-documentation available at its http://mojolicio.us[home page].
+documentation available at its https://mojolicious.org/[home page].
 
 In addition to Mojolicious and its dependencies, several other CPAN modules are
 required by the openQA package. See
@@ -155,10 +154,7 @@ but that is no longer possible.
 As stated in the previous section, every feature implemented in both packages
 should be backed by proper tests.
 http://perldoc.perl.org/Test/Most.html[Test::Most] is used to implement those
-tests. As usual, tests are located under the `/t/` directory. In the openQA
-package, one of the tests consists of a call to
-http://perltidy.sourceforge.net/[Perltidy] to ensure that the contributed code
-follows the most common Perl style conventions.
+tests. As usual, tests are located under the `t/` directory.
 
 == Folder structure
 
@@ -179,25 +175,85 @@ systemd:: systemd service definitions
 templates:: HTML templates delivered by web UI
 tools:: Development tools
 
+[[dependencies]]
+== Dependencies
+
+Here we differentiate between these types of dependencies:
+
+* Runtime: run openQA
+* Test: run the test suite of openQA itself
+* For developers: style checks, code coverage etc.
+
+You can get those dependencies in several ways:
+
+* Packaged in openSUSE and other distributions
+* Manually install from source
+* Perl modules via CPAN client
+* Use a prebuilt container
+
+We manage our dependencies via a central `dependencies.yaml` file per
+repository. We generate the `.spec` files for `rpm` as well as a `cpanfile`
+for quicker installation of all Perl modules.
+
+[[runtime-dependencies]]
+=== Runtime Dependencies
+
+For openSUSE you need the packages `openQA` and `os-autoinst`.
+That should be everything necessary to run an openQA instance. For more
+details check <<Installing.asciidoc#_installation,Installation>.
+
+If you have to manually install packages, look into the `dependencies.yaml`
+file or the spec file, e.g. `dist/rpm/openQA.spec` (or `dist/rpm/*.spec` in
+general). You can leave out sections starting with `test_requires` for example.
+
+A CPAN client will by default also leave out modules for testing and
+development.
+
+As alternative you can try out the
+<<Installing.asciidoc#single-instance-container,Single-Instance Container>> as
+a minimal runtime environment.
+
+[[development-dependencies]]
+=== Development Dependencies
+
+For openSUSE:
+```
+zypper install openQA-devel
+```
+
+For distributions without those packages, look into `dependencies.yaml` or
+`dist/rpm/openQA.spec` like above. The `devel_*requires` sections/definitions
+list dependencies for tests and development.
+
+If you just want to run the test suite to see if it passes on your platform,
+you would not need the developer's dependencies, like `Perl::Tidy` and
+`Devel::Cover`. They are recommended as soon as you plan a contribution.
+
+For the UI tests you need `perl-Selenium-Remote-Driver` as well as
+`chromedriver` and either `chrome` or `chromium`.
+
+See <<Contributing.asciidoc#run-tests-in-container,Run tests in a container>>
+for a container with all dependencies.
 
 [[development-setup]]
 == Development setup
+
+The following explains in detail what is necessary to develop openQA code.
+
+For a setup with containers you can checkout the
+<<Contributing.asciidoc#quick-container-development-setup,Quick container
+development setup>> below with a complete list of instructions.
+
+Otherwise follow the detailed steps below. This setup is applicable to a
+container environment as well as without.
+
 For developing openQA and os-autoinst itself it makes sense to checkout the
 <<Contributing.asciidoc#repo-urls,Git repositories>> and either execute
 existing tests or start the daemons manually.
 
-[[dependencies]]
-=== Dependencies
-Have a look at the packaged version (e.g. `dist/rpm/openQA.spec` within the
-root of the openQA repository) for all required dependencies. For development
-build time dependencies need to be installed as well. Recommended
-dependencies such as logrotate can be ignored. For openSUSE there is also the
-`openQA-devel` meta-package which pulls all required dependencies for
-development.
-
-You can find all required Perl modules in form of a `cpanfile` that enables
-you to install them with a CPAN client. They are also defined in
-`dist/rpm/openQA.spec`.
+Install the necessary
+<<Contributing.asciidoc#development-dependencies,Development Dependencies>>
+first.
 
 [[testing]]
 === Conducting tests
@@ -431,10 +487,240 @@ Also find more details in
   look at the
   https://github.com/Martchus/openQA-helper[openQA-helper repository].
 
+[[quick-container-setup]]
+=== Quick container development setup
+
+==== Requirements
+
+In these examples we use the following tools which are all in openSUSE Leap:
+
+* https://podman.io/[`podman`]
+* https://en.opensuse.org/Distrobox[`distrobox`]
+
+You can use similar tools like `docker`, `toolbox` etc. to achieve the same.
+
+==== Intro
+
+These instructions are ready to use without any changes. You only might want to
+change the `OPENQA_BASEDIR` location.
+
+You will run two containers:
+
+* A separate postgres container (so you don't have to deal with the database
+  when updating your dev container)
+* The openQA environment in an openSUSE Leap or Tumbleweed container, which
+  both can use the same postgres container.
+  In this example we will be setting up and using the Leap container.
+
+`distrobox`, for example, can be used to take care of network, home directory,
+permissions, and graphical environment.
+
+Some common settings we are using here:
+
+* `OPENQA_BASEDIR=$HOME/openqadev.leap`
+* Database name: `openqa-local`
+* Database user: `geekotest`
+* Database container name: `postgres-openqa`
+
+==== Environment Variables
+
+Create a file that you can `source` whenever you want to work in this
+environment:
+
+File: `~/localopenqa.leap.sh`:
+[source,sh]
+----
+#!/bin/bash
+
+# The directory where openQA operates in and creates everything in:
+# * git repos
+# * openQA assets, pool directory, testresults
+# * configuration
+
+export OPENQA_BASEDIR=$HOME/openqadev.leap
+
+# The rest of the variables can likely stay like this
+
+# If you want to import database dumps from elsewhere
+#export SQL_DUMP_DIR=$HOME/sqldumps
+
+export OPENQA_CONFIG=$OPENQA_BASEDIR/config
+export OPENQA_REPOS="$OPENQA_BASEDIR/git"
+PATH=$OPENQA_REPOS/openQA/scripts:$PATH
+
+# ----------- CPAN ------------
+# Whenever necessary, you can install additional modules from CPAN like:
+# cpanm -l $OPENQA_BASEDIR/perl5 Some::Module@1.23
+export PERL5LIB=$OPENQA_BASEDIR/perl5/lib/perl5
+PATH=$OPENQA_BASEDIR/perl5/bin:$PATH
+
+export OPENQA_KEY=1234567890ABCDEF
+export OPENQA_SECRET=1234567890ABCDEF
+----
+
+==== Postgres Container
+
+To be able to reuse the database later when stopping the container, we create a
+named volume.
+We are using a Debian postgres container here, but it doesn't matter what
+you use.
+[source,sh]
+----
+### host ###
+podman volume create mypostgres
+podman run -it --rm --name postgres-openqa \
+    -v mypostgres:/var/lib/postgresql/data \
+    -p 5432:5432 \
+    -e POSTGRES_PASSWORD="x" \
+    -d \
+    docker.io/library/postgres:17
+
+# Optionally add this to mount a directory for sql dumps to import
+#    -v "$SQL_DUMP_DIR:/sqldumps" \
+----
+
+For the next step you need to execute commands in the container:
+
+[source,sh]
+----
+# Create the role and database
+podman exec -it postgres-openqa su - postgres
+
+### postgres container ###
+createuser --no-createdb --pwprompt geekotest
+# enter password ('x' in this example)
+createdb -O geekotest openqa-local
+----
+
+==== Distrobox
+
+You can use a Leap container and install the requirements, but you can also
+build your own container with the requirements already installed.
+
+[source,sh]
+----
+### host ###
+distrobox create -i registry.opensuse.org/opensuse/leap:15.6 -n box-openqadev-leap
+distrobox enter box-openqadev-leap
+# Tumbleweed:
+#distrobox create -i registry.opensuse.org/opensuse/tumbleweed:latest -n box-openqadev-tw
+
+### distrobox container ###
+
+# Leap 15.6
+zypper addrepo -G -p 95 https://download.opensuse.org/repositories/devel:/openQA/15.6 devel-openqa \
+    && zypper addrepo -G -p 95 https://download.opensuse.org/repositories/devel:/openQA:/Leap:/15.6/15.6/ devel-openqa-leap
+
+# Tumbleweed
+#zypper addrepo -G p 95 https://download.opensuse.org/repositories/devel:/openQA/openSUSE_Tumbleweed devel-openqa \
+#        && zypper addrepo -G -p 95 https://download.opensuse.org/repositories/SUSE:/CA/openSUSE_Tumbleweed/SUSE:CA.repo
+
+zypper -n ref && zypper -n install ca-certificates-suse \
+        os-autoinst-devel openQA-devel \
+        os-autoinst-scripts-deps \
+        os-autoinst-scripts-deps-devel \
+        crudini
+# plus your favourite dev tools, e.g. tmux
+----
+Note: https://github.com/pixelb/crudini[`crudini`] is a tool for manipulating
+values in ini files.
+
+==== Git
+
+[source,sh]
+----
+### distrobox container ###
+# Remember to source this file
+source ~/localopenqa.leap.sh
+mkdir $OPENQA_BASEDIR
+mkdir $OPENQA_REPOS
+mkdir $OPENQA_BASEDIR/openqa $OPENQA_CONFIG $OPENQA_BASEDIR/perl5 $OPENQA_BASEDIR/openqa/db
+
+git clone git@github.com:os-autoinst/openQA $OPENQA_REPOS/openQA
+git clone git@github.com:os-autoinst/os-autoinst $OPENQA_REPOS/os-autoinst
+git clone git@github.com:os-autoinst/scripts $OPENQA_REPOS/scripts
+----
+
+==== Config Files
+
+[source,sh]
+----
+### distrobox container ###
+echo <<< EOM
+[localhost]
+key = 1234567890ABCDEF
+secret = 1234567890ABCDEF
+EOM >> $OPENQA_CONFIG/client.conf
+
+echo <<< EOM
+[production]
+dsn = dbi:Pg:dbname=openqa-local;host=127.0.0.1;port=5432
+user = geekotest
+password = x
+EOM >> $OPENQA_CONFIG/database.ini
+
+cp $OPENQA_REPOS/openQA/etc/openqa/workers.ini $OPENQA_CONFIG/workers.ini
+crudini --set $OPENQA_CONFIG/workers.ini global HOST http://localhost:9526
+crudini --set $OPENQA_CONFIG/workers.ini global WORKER_HOSTNAME 127.0.0.1
+
+cp $OPENQA_REPOS/openQA/etc/openqa/openqa.ini $OPENQA_CONFIG/openqa.ini
+crudini --set $OPENQA_CONFIG/openqa.ini auth method Fake
+----
+
+==== Build openQA
+
+[source,sh]
+----
+### distrobox container ###
+# Build steps
+cd $OPENQA_REPOS/os-autoinst
+make
+cd $OPENQA_REPOS/openQA
+make node_modules
+
+# Optionally check if you can access the database
+#psql -U geekotest -h 127.0.0.1 -p 5432 openqa-local
+
+# Initialize the database schema. This is using the database.ini you
+# created above.
+$OPENQA_REPOS/openQA/script/initdb --init_database
+----
+
+==== Run openQA
+
+Ready!
+
+For playing with the web UI, you just need to run the `openqa-webui-daemon`.
+For running openQA jobs, you need all daemons.
+
+[source,sh]
+----
+### distrobox container ###
+# Main web UI (required for web interface and web API)
+openqa-webui-daemon
+
+# Websockets (for workers to connect so they can receive job assignments)
+openqa-websockets daemon
+
+# Scheduler (assigns jobs to workers)
+openqa-scheduler daemon
+
+# Gru (for running background tasks as Minion jobs)
+openqa gru run
+
+# Livehandler (for the developer mode on the web UI)
+openqa-livehandler daemon
+
+# Worker
+worker --isotovideo "$OPENQA_REPOS/os-autoinst/isotovideo" --instance 1 --verbose --apikey $OPENQA_KEY --apisecret $OPENQA_SECRET
+----
+
+For running unit tests, see <<Contributing.asciidoc#testing,Conducting tests>>.
+
 [[dependency-handling]]
 == Handling of dependencies
 
-=== Javascript and CSS
+=== JavaScript and CSS
 Install third-party JavaScript and CSS files via their corresponding npm
 packages and add the paths of those files to `assets/assetpack.def`.
 
@@ -604,10 +890,8 @@ specific to that environment.
 Be sure to install all required dependencies. The package `openQA-devel` will
 provide them.
 
-If the package is not available the dependencies can also be found in the file
-`dist/rpm/openQA.spec` in the openQA repository. In this case also the package
-`perl-Selenium-Remote-Driver` is required to run UI tests. You also need to
-install chromedriver and either chrome or chromium for the UI tests.
+For more information look into
+<<Contributing.asciidoc#development-dependencies,Development Dependencies>>.
 
 To execute the testsuite use `make test`. This will also initialize a
 temporary PostgreSQL database used for testing. To do this step manually run
@@ -623,6 +907,7 @@ To keep the test database running after executing tests with the `Makefile`, add
 
 To watch the execution of the UI tests, set the environment variable `NOT_HEADLESS`.
 
+[[run-tests-in-container]]
 === Run tests within a container
 The container used in this section of the documentation is not identical with the container used
 within the CI. To run tests within the CI environment locally, checkout the
--- a/docs/Installing.asciidoc
+++ b/docs/Installing.asciidoc
@@ -33,6 +33,7 @@ setup suitable to develop openQA itself, have a look at the
 
 openQA is provided in containers. Multiple variants exist.
 
+[[single-instance-container]]
 === Single-instance container
 
 The easiest and quickest way to spawn a single instance of openQA with a


container/devel/Dockerfile.openqa.leap
```diff
@@ -0,0 +1,34 @@
+FROM registry.opensuse.org/opensuse/leap:15.6
+
+RUN zypper ar -G https://download.opensuse.org/repositories/devel:/openQA/15.6 devel-openqa \
+        && zypper ar -G https://download.opensuse.org/repositories/devel:/openQA:/Leap:/15.6/15.6/ devel-openqa-leap \
```
okurz: I suggest to use priorities as explained in http://open.qa/docs/#_development_version_repository

container/devel/Dockerfile.openqa.leap
```diff
@@ -0,0 +1,34 @@
...
+        && zypper ar -G https://download.opensuse.org/repositories/SUSE:/CA/15.6/SUSE:CA.repo \
+        && zypper -n ref \
+    && true
```
okurz: What's the purpose of this?

container/devel/Dockerfile.openqa.leap
```diff
@@ -0,0 +1,34 @@
...
+        openssh-clients \
+    && true
+
+# scripts
+RUN zypper -n install \
+        os-autoinst-scripts-deps \
+        os-autoinst-scripts-deps-devel \
+        python3-pytest \
+        python3-virtualenv \
+    && true
```
okurz: How about

```suggestion
RUN zypper -n ref && zypper -n install \
        os-autoinst-devel openQA-devel \
        expect \
        nodejs \
        openssh-clients \
        ca-certificates-suse \
        os-autoinst-scripts-deps \
        os-autoinst-scripts-deps-devel \
        python3-pytest \
        python3-virtualenv \
    && zypper clean
```

and I think this shows a flaw in https://github.com/os-autoinst/scripts/blob/master/dist/rpm/os-autoinst-scripts-deps.spec#L32 compared to https://github.com/os-autoinst/os-autoinst/blob/master/dist/rpm/os-autoinst.spec#L122 that os-autoinst-scripts development requirements should include main+test requires. Maybe better fix that there? And the other dependencies: When they are actually needed for development then they should be part of os-autoinst-devel or openQA-devel

container/devel/Dockerfile.openqa.leap
```diff
@@ -0,0 +1,34 @@
...
+
+# Run custom scripts
+# Put your scripts under ./scripts. Filename should start with custom
+# to be ignored by git
+RUN mkdir /tmp/userscripts
+COPY etc/runuserscripts.sh /tmp/
+COPY scripts/*.sh /tmp/userscripts
+RUN /tmp/runuserscripts.sh
```
okurz: Why not just mount in a volume with `-v` instead of copying?

container/devel/README.md
```diff
@@ -0,0 +1,171 @@
...
+These examples use `podman` and
+[`distrobox`](https://en.opensuse.org/Distrobox). You can use similar tools
+like `docker`, `toolbox` etc. to achieve the same.
+
+You will run two containers:
+* A separate postgres container (so you don't have to deal with the database
+  when updating your dev container)
+* The openQA environment in a openSUSE Leap or Tumbleweed container, which
+  both can use the same postgres container.
+  In this example we will be settting up and using the Leap container.
```
okurz: I suggest to link to http://open.qa/docs/#contributing if not rather move the documentation there. Right now unfortunately it seems this documentation is a complete alternative to http://open.qa/docs/#development-setup confusing users which instructions they should follow.

container/devel/Dockerfile.openqa.leap
```diff
@@ -0,0 +1,34 @@
...
+RUN zypper -n install \
+        os-autoinst-scripts-deps \
+        os-autoinst-scripts-deps-devel \
+        python3-pytest \
+        python3-virtualenv \
+    && true
+
+# Run custom scripts
+# Put your scripts under ./scripts. Filename should start with custom
+# to be ignored by git
```
okurz: "should start with custom"? What does that mean?

container/devel/Dockerfile.openqa.leap
okurz, line 1: Please make it clear in documentation how this relates to [container/devel/Dockerfile](https://github.com/os-autoinst/openQA/blob/master/container/devel/Dockerfile)

okurz (changes requested):

container/devel/README.md
perlpunk: That's why it's a draft

container/devel/Dockerfile.openqa.leap
perlpunk: The `&& true`? Just in my muscle memory to always end multiline commands with this, so it is easy to add a command without having to add a `\` at the previous line :)
I guess I can remove it here

container/devel/Dockerfile.openqa.leap
perlpunk: done

container/devel/Dockerfile.openqa.leap
perlpunk: it should start with "custom". I added quotes now.

container/devel/Dockerfile.openqa.leap
perlpunk: done

container/devel/Dockerfile.openqa.leap
perlpunk: I had these lines in there for a while, I think even before we had a package for the scripts repo.
I now created
* https://github.com/os-autoinst/scripts/pull/434

perlpunk: I can't remember why I needed `expect`. It was a long time ago. Maybe it's not necessary anymore.
I'll also remove nodejs. If we don't have it in the rpm spec, it's not needed.

docs/Contributing.asciidoc
```diff
@@ -431,6 +435,235 @@ Also find more details in
...
+
+# Tumbleweed
+#zypper addrepo -G p 95 https://download.opensuse.org/repositories/devel:/openQA/openSUSE_Tumbleweed devel-openqa \
+#        && zypper addrepo -G -p 95 https://download.opensuse.org/repositories/SUSE:/CA/openSUSE_Tumbleweed/SUSE:CA.repo \
+
+zypper -n ref && zypper -n install ca-certificates-suse \
+        os-autoinst-devel openQA-devel \
+        os-autoinst-scripts-deps \
+        os-autoinst-scripts-deps-devel \
+        openssh-clients \
```
perlpunk: openssh-clients is necessary to do `git clone git@...`

Should we add it to os-autoinst and openQA requirements?

docs/Contributing.asciidoc
```diff
@@ -182,6 +182,10 @@ tools:: Development tools
 
 [[development-setup]]
 == Development setup
+
+For the lazy, check out the <<Contributing.asciidoc#quick-container-setup,Quick
+container setup>> with instructions you can mostly copy & paste.
```
d3flex: This is the very first line of the `Development setup` and it would be more enlightening to describe the options or the differences from other sections. Something like

```In order to contribute, your development environment needs the following steps [blabla].. In order to setup a containerized env follow these instructions....[]```

File: docs/Contributing.asciidoc (hunk 7 above)
Comment (okurz) [reply docs/Contributing.asciidoc]: yes, makes sense. So part of os-autoinst-devel and openQA-devel you mean I assume?

File: docs/Contributing.asciidoc [hunk 9]
```diff
@@ -431,6 +435,235 @@ Also find more details in
...
+# Initialize the database schema. This is using the database.ini you
+# created above.
+$OPENQA_REPOS/openQA/script/initdb --init_database
+----
+
+==== Run openQA
+
+Ready!
+
+For playing with the Webui, you just need to run this one daemon.
```
okurz:
```suggestion
For playing with the web UI, you just need to run this one daemon.
```

At least we use "web UI" many times more in documentation rather than webUI. And no other place mentions "Webui" so we should avoid that :)

docs/Contributing.asciidoc
okurz: @b10n1k good point. The next paragraph covers parts of it. I just suggest to rephrase this.

Instead of "For the lazy" please write for which use case the "quick container setup" is actually suitable in comparison to the next paragraph which says "For developing openQA […] itself […] execute existing tests or start the daemons manually".

So how you mention the new approach as alternative to provide a container based development environment?

And then for completeness I suggest to link to https://github.com/os-autoinst/openQA/pull/6590/files#diff-4d6d1050121df87969e32770abd50500a1fbda9f136db4fbb374fe05a236f2c2R859 as well and make the different use cases clear

docs/Contributing.asciidoc
```diff
@@ -431,6 +435,235 @@ Also find more details in
...
+You can use a leap container and install the requirements, but you can also
+build your own container with the requirements already installed.
+
+[source,sh]
+----
+### host ###
+distrobox create -i registry.opensuse.org/opensuse/leap:15.6 -n box-openqadev
+distrobox enter box-openqadev
+
+### distrobox ###
```
Wabri:
```suggestion
### distrobox container ###
```

docs/Contributing.asciidoc
```diff
@@ -431,6 +435,235 @@ Also find more details in
...
+
+==== Distrobox
+
+You can use a leap container and install the requirements, but you can also
+build your own container with the requirements already installed.
+
+[source,sh]
+----
+### host ###
+distrobox create -i registry.opensuse.org/opensuse/leap:15.6 -n box-openqadev
```
Wabri: Since below you show also the tumbleweed zypper setup might be worth to add tumbleweed here too?

docs/Contributing.asciidoc
```diff
@@ -431,6 +435,235 @@ Also find more details in
...
+### distrobox ###
+# Remember to source this file
+source ~/localopenqa.leap.sh
+mkdir $OPENQA_BASEDIR
+mkdir $OPENQA_REPOS
+mkdir $OPENQA_BASEDIR/openqa $OPENQA_CONFIG $OPENQA_BASEDIR/perl5 $OPENQA_BASEDIR/openqa/db
+
+git clone git@github.com:os-autoinst/openQA $OPENQA_REPOS/openQA
+git clone git@github.com:os-autoinst/os-autoinst $OPENQA_REPOS/os-autoinst
+git clone git@github.com:os-autoinst/scripts $OPENQA_REPOS/scripts
```
Wabri, line 641:
```
📦[wabri@openqa ~]$ git clone git@github.com:os-autoinst/openQA ~/openqa-sandbox/openqa-repo
Cloning into '/home/wabri/openqa-sandbox/openqa-repo'...
git@github.com: Permission denied (publickey).
fatal: Could not read from remote repository.

Please make sure you have the correct access rights
and the repository exists.
```

Maybe is necessary to also add in the localopenqa.leap.sh something like this:

```
export GIT_SSH_COMMAND='ssh -i ~/.ssh/<yourkey> -o IdentitiesOnly=yes'
```

Or maybe let the user clone from the outside of the container since is a developer we can suppose that have git and ssh enabled. We can also suppose that a clone of the repositories have already been done. Thinking out loud we could consider to add the clone setup on the requirements section, this is the solution I like the most.

docs/Contributing.asciidoc
```diff
@@ -431,6 +435,235 @@ Also find more details in
...
+openqa-websockets daemon
+
+# Scheduler
+openqa-scheduler daemon
+
+# Gru (Minion)
+openqa gru run
+
+# Lievehandler
+openqa-livehandler daemon
```
Wabri: I would add the `$OPENQA_REPOS/openQA/script/` path to PATH and use the one on the repository and not the one coming from zypper, I had a necessity to do it while I was working with `openqa-bootstrap`.

docs/Contributing.asciidoc
perlpunk: I am wondering why this doesn't work for you. distrobox mounts your home directory, so `~/.ssh` should be there already...

docs/Contributing.asciidoc
perlpunk: > I would add the `$OPENQA_REPOS/openQA/script/` path to PATH

That is already set in `localopenqa.leap.sh`

docs/Contributing.asciidoc
perlpunk: added

docs/Contributing.asciidoc
perlpunk: done, although it seems a bit verbose to me

perlpunk:
I rearranged several things now. "Dependencies" is now a top level section that is referred to from several places.

docs/Contributing.asciidoc
```diff
@@ -155,10 +163,7 @@ but that is no longer possible.
 As stated in the previous section, every feature implemented in both packages
 should be backed by proper tests.
 http://perldoc.perl.org/Test/Most.html[Test::Most] is used to implement those
-tests. As usual, tests are located under the `/t/` directory. In the openQA
-package, one of the tests consists of a call to
-http://perltidy.sourceforge.net/[Perltidy] to ensure that the contributed code
-follows the most common Perl style conventions.
+tests. As usual, tests are located under the `/t/` directory.
```
d3flex: I think it should be `t/`

docs/Contributing.asciidoc
```diff
@@ -179,25 +184,90 @@ systemd:: systemd service definitions
...
+See <<Contributing.asciidoc#run-tests-in-container,Run tests in a container>>
+for a container with all dependencies.
+
 
 [[development-setup]]
 == Development setup
+
+The following explains in detail what is necessary to develop openQA code.
+
+For something to mostly just copy & paste, check out the
```
d3flex: > For something to mostly just copy & paste

I would rather say something like

`The simplest approach. if you are in hurry...`

docs/Contributing.asciidoc
perlpunk: Yeah, I was also wondering about the leading slash. This sentence I only moved around. I will remove the slash.

docs/Contributing.asciidoc
Martchus: Please don't add the phrasing "if you are in hurry" in our documentation. This just doesn't read very professional :-)

docs/Contributing.asciidoc
```diff
@@ -97,6 +97,15 @@ requests for consideration or create an issue with a code change proposal.
...
 *** Do *not* combine sub attributes with signatures (requires Perl 5.28+), e.g. `sub foo :lvalue ($first) {`
 
+* General
+
+** Let http://perltidy.sourceforge.net/[Perltidy] tidy perl files
+automatically:
+
+    tools/tidyall --git # tidy changed files
+    tools/tidyall --all # tidy all files
+    make tidy # to tidy also Javascript files in the openQA project
```
Martchus:
```suggestion
    make tidy # to tidy also JavaScript files in the openQA project
```

docs/Contributing.asciidoc
```diff
@@ -179,25 +184,90 @@ systemd:: systemd service definitions
...
 
+[[dependencies]]
+== Dependencies
+
+Here we differentiate between these types of dependencies:
+
+* Runtime: Run openQA
+* Development:
+** Test: Run openQA's test suite
+** For developers: style checks, code coverage etc.
```
Martchus: I would be consistent with the casing after the colon. I also think the list can be simplified a little bit:

```suggestion
* Runtime: run openQA
* Testing: run the test suite of openQA itself
* For developers: style checks, code coverage etc.
```

docs/Contributing.asciidoc
```diff
@@ -179,25 +184,90 @@ systemd:: systemd service definitions
...
+
+You can get those dependencies in several ways:
+
+* packaged in openSUSE and other distributions
```
Martchus:
```suggestion
* Packaged in openSUSE and other distributions
```

docs/Contributing.asciidoc
```diff
@@ -179,25 +184,90 @@ systemd:: systemd service definitions
...
+* Manually install from source
+* Perl modules via CPAN client
+* Use a prebuilt container
+
+We manage our dependencies via a central `dependencies.yaml` file per
+repository. We generate the `.spec` files for `rpm` as well as a `cpanfile`
+for quicker installation of all perl modules.
```
Martchus:
```suggestion
for quicker installation of all Perl modules.
```

docs/Contributing.asciidoc
```diff
@@ -179,25 +184,90 @@ systemd:: systemd service definitions
...
+
+For openSUSE:
+```
+zypper install os-autoinst-devel openQA-devel
+```
+
+For distributions without those packages, look into `dependencies.yaml` or
+`dist/rpm/openQA.spec` like above.
+
+The `-devel` packages will install dependencies for tests and development.
```
Martchus: This could probably be merged with the previous paragraph. Maybe it would also make more sense to mention `devel_requires` instead of `-devel` packages:

```suggestion
For distributions without those packages, look into `dependencies.yaml` or
`dist/rpm/openQA.spec` like above. The `devel_*requires` sections/definitions
list dependencies for tests and development.
```

docs/Contributing.asciidoc
```diff
@@ -179,25 +184,90 @@ systemd:: systemd service definitions
...
+
+If you just want to run the test suite to see if it passes on your platform,
+you would not need the developer's dependencies, like `Perl::Tidy` and
+`Devel::Cover`. They are recommended as soon as you plan a contribution.
```
Martchus, line 230:
```suggestion
`Devel::Cover`. They are only required as soon as you plan a contribution.
```

docs/Contributing.asciidoc
```diff
@@ -431,6 +501,241 @@ Also find more details in
...
+
+==== Intro
+
+These instructions are ready to use without any changes. You only might want to
+change the `OPENQA_BASEDIR` location.
+
+You will run two containers:
+
+* A separate postgres container (so you don't have to deal with the database
+  when updating your dev container)
```
Martchus, line 510: Although you probably still need to do a manual database migration in case you update the postgres container. And for this one can probably no longer follow our instructions for this as-is. (Because those instructions are maybe openSUSE specific or things are generally more involved when using a container.)

You can keep these instructions but if you could shed some light on how database migrations work in this setup it would probably be beneficial.

docs/Contributing.asciidoc
```diff
@@ -431,6 +501,241 @@ Also find more details in
...
+* The openQA environment in a openSUSE Leap or Tumbleweed container, which
```
Martchus:
```suggestion
* The openQA environment is an openSUSE Leap or Tumbleweed container, which
```

docs/Contributing.asciidoc
```diff
@@ -431,6 +501,241 @@ Also find more details in
...
+
+Create a file that you can `source` whenever you want to work in this
+environment:
+
+File: `~/localopenqa.leap.sh`:
+[source,sh]
+----
+#!/bin/bash
+
+# The directory where openQA operates in and which creates everything:
```
Martchus:
```suggestion
# The directory where openQA operates in and creates everything in:
```
(Otherwise it sounds like the directory itself would create everything.)

docs/Contributing.asciidoc
```diff
@@ -431,6 +501,241 @@ Also find more details in
...
+
+[source,sh]
+----
+# Create the role and database
+podman exec -it postgres-openqa su - postgres
+
+### postgres container ###
+createuser --no-createdb --pwprompt geekotest
+# enter password ('x' in this example)
+createdb -O geekotest openqa-local
```
Martchus, line 592: Doesn't this need to be prefixed by a `podman exec …` to actually run within the posgres container?

docs/Contributing.asciidoc
```diff
@@ -431,6 +501,241 @@ Also find more details in
...
+----
+
+==== Distrobox
+
+You can use a leap container and install the requirements, but you can also
```
Martchus:
```suggestion
You can use a Leap container and install the requirements, but you can also
```

docs/Contributing.asciidoc
Martchus: I would say we should if Git is also already a requirement.

docs/Contributing.asciidoc
```diff
@@ -431,6 +501,241 @@ Also find more details in
...
+        os-autoinst-scripts-deps \
+        os-autoinst-scripts-deps-devel \
+        openssh-clients \
+        crudini
+# plus your favourite dev tools, e.g. tmux
+----
+Note: https://github.com/pixelb/crudini[`crudini`] is a tool for manipulating
+values in ini files.
+
+Time to grab a hot beverage while installing...
```
Martchus:
```suggestion
Time to grab a hot beverage while installing…
```

docs/Contributing.asciidoc
```diff
@@ -431,6 +501,241 @@ Also find more details in
...
+# Initialize the database schema. This is using the database.ini you
+# created above.
+$OPENQA_REPOS/openQA/script/initdb --init_database
+----
+
+==== Run openQA
+
+Ready!
+
+For playing with the Webui, you just need to run this one daemon.
```
Martchus:
```suggestion
For playing with the web UI, you just need to run this one daemon.
```

docs/Contributing.asciidoc
```diff
@@ -431,6 +501,241 @@ Also find more details in
...
+For running openQA jobs, you need all.
+
+[source,sh]
+----
+### distrobox container ###
+# WebUI
```
Martchus:
```suggestion
# Main web UI (required for web interface and web API)
```

docs/Contributing.asciidoc
```diff
@@ -431,6 +501,241 @@ Also find more details in
...
+$OPENQA_REPOS/openQA/script/openqa-webui-daemon
+
+# Worker
+worker --isotovideo "$OPENQA_REPOS/os-autoinst/isotovideo" --instance 1 --verbose --apikey $OPENQA_KEY --apisecret $OPENQA_SECRET
```
Martchus: I would put the worker last as it is the only service that isn't part of the web UI and therefore shouldn't be interleaved with the web UI services. (Except for the worker cache services which should go very last but you don't mention them here anyway.)

docs/Contributing.asciidoc
```diff
@@ -431,6 +501,241 @@ Also find more details in
...
+
+# Websocket
```
Martchus:
```suggestion
# Websockets (for workers to connect so they can receive job assignments)
```

docs/Contributing.asciidoc
```diff
@@ -431,6 +501,241 @@ Also find more details in
...
+openqa-websockets daemon
+
+# Scheduler
```
Martchus:
```suggestion
# Scheduler (assigns jobs to workers)
```

docs/Contributing.asciidoc
```diff
@@ -431,6 +501,241 @@ Also find more details in
...
+openqa-scheduler daemon
+
+# Gru (Minion)
```
Martchus:
```suggestion
# Gru (for running background tasks as Minion jobs)
```

docs/Contributing.asciidoc
```diff
@@ -431,6 +501,241 @@ Also find more details in
...
+openqa gru run
+
+# Lievehandler
```
Martchus:
```suggestion
# Livehandler (for the developer mode on the web UI)
```

Martchus (commented):
Looks good but I also have a lot of nitpicks :-)

docs/Contributing.asciidoc
d3flex: if you need the job done...
Alternative, ...


In general I dont expect _copy paste_ of my words, as they are only for inspiration or demostrate intention. So even without `if you are in hurry` is also ok with me

docs/Contributing.asciidoc
perlpunk: The `podman exec` is right above this

docs/Contributing.asciidoc
Martchus: The many times are probably mostly due to me. The reason I think UI should be separated is that it is an abbreviation and abbreviations are generally treated just like any other word and thus need to be preceded by a space here. Note that "webUI" is also not a name like "openQA".

I think Wikipedia is a good example, e.g. https://en.wikipedia.org/wiki/User_interface always precedes "UI" with a space if used in the middle of a sentence. The same counts for abbreviations that are units, e.g. https://en.wikipedia.org/wiki/Millisecond.

docs/Contributing.asciidoc
perlpunk: @okurz  This comment is already outdated, please check the updated content

perlpunk: > And then for completeness I suggest to link to https://github.com/os-autoinst/openQA/pull/6590/files#diff-4d6d1050121df87969e32770abd50500a1fbda9f136db4fbb374fe05a236f2c2R859 as well and make the different use cases clear

That link does not make it clear to me what I should link to. This is just pointing to Contributing.asciidoc

docs/Contributing.asciidoc
perlpunk: The question is, should it be only a development requirement?
Even in normal instances people would want to clone via ssh. Or should it be just a `Recommends`?

docs/Contributing.asciidoc
perlpunk: Strictly speaking, they are not really required

docs/Contributing.asciidoc
perlpunk: I actually never had to do it for my development environment.
I work with a fresh database or a database dump.
I assume a simple `pg_dump` and `pg_restore` wouldn't work?
But I would say, I consider this out of scope for this ticket.

docs/Contributing.asciidoc
Martchus: I would put it wherever we put Git itself. (Since we do a lot with Git at this point it is probably a normal mandatory runtime dependency.)

docs/Contributing.asciidoc
```diff
@@ -97,6 +97,15 @@ requests for consideration or create an issue with a code change proposal.
...
 
+* General
+
+** Let http://perltidy.sourceforge.net/[Perltidy] tidy perl files
+automatically:
+
+    tools/tidyall --git # tidy changed files
+    tools/tidyall --all # tidy all files
+    make tidy # to tidy also JavaScript files in the openQA project
+
```
okurz: This looks inconsistent and redundant. The first sentence says "tidy perl" but then "make tidy" mentions JavaScript files. Our Makefile is well structured. Please don't pick details out of it. Just calling `make tidy` is what we should do and that is what https://github.com/os-autoinst/openQA/pull/6590/files#diff-4d6d1050121df87969e32770abd50500a1fbda9f136db4fbb374fe05a236f2c2R59 already mentions. so why this point "General" at all?

docs/Contributing.asciidoc
```diff
@@ -179,25 +184,88 @@ systemd:: systemd service definitions
...
+[[runtime-dependencies]]
+=== Runtime Dependencies
+
+For openSUSE there is `openQA` and `os-autoinst`, including several subpackages.
+That should be everything necessary to run an openQA instance.
+
+```
+zypper install openQA os-autoinst
+```
```
okurz: That is not correct as it leaves out the worker package and hence it does not make sense to pull in os-autoinst. Instead of repeating (incorrectly) I suggest to just point to our installation instructions.

docs/Contributing.asciidoc
okurz: Instead of "For something to mostly just copy & paste" please explain the specific use case this is useful for in comparison to the other existing way. But I guess we should look into https://github.com/os-autoinst/openQA/pull/6590#discussion_r2231851548 first

docs/Contributing.asciidoc
```diff
@@ -179,25 +184,88 @@ systemd:: systemd service definitions
...
 == Development setup
+
+The following explains in detail what is necessary to develop openQA code.
+
+For something to mostly just copy & paste, check out the
+<<Contributing.asciidoc#quick-container-development-setup,Quick container
+development setup>>.
+
+Otherwise follow the detailed steps below. This setup leaves it up to
+you if you run this in a container or not.
```
okurz:
```suggestion
Otherwise follow the detailed steps below. This setup is applicable to
a container environment as well as without.
```

docs/Contributing.asciidoc
```diff
@@ -431,10 +499,246 @@ Also find more details in
...
+These instructions are ready to use without any changes. You only might want to
+change the `OPENQA_BASEDIR` location.
+
+You will run two containers:
+
+* A separate postgres container (so you don't have to deal with the database
+  when updating your dev container)
+* The openQA environment in an openSUSE Leap or Tumbleweed container, which
+  both can use the same postgres container.
+  In this example we will be settting up and using the Leap container.
```
okurz:
```suggestion
  In this example we will be setting up and using the Leap container.
```

docs/Contributing.asciidoc
okurz: the dependency was introduced with https://github.com/os-autoinst/openQA/pull/6611 so I guess you can remove this line

docs/Contributing.asciidoc
```diff
@@ -431,10 +499,246 @@ Also find more details in
...
+
+### distrobox container ###
+
+# Leap 15.6
+zypper addrepo -G -p 95 https://download.opensuse.org/repositories/devel:/openQA/15.6 devel-openqa \
+    && zypper addrepo -G -p 95 https://download.opensuse.org/repositories/devel:/openQA:/Leap:/15.6/15.6/ devel-openqa-leap
+
+# Tumbleweed
+#zypper addrepo -G p 95 https://download.opensuse.org/repositories/devel:/openQA/openSUSE_Tumbleweed devel-openqa \
+#        && zypper addrepo -G -p 95 https://download.opensuse.org/repositories/SUSE:/CA/openSUSE_Tumbleweed/SUSE:CA.repo
```
okurz, line 616: Instead of repeating how about pointing to https://github.com/os-autoinst/openQA/blob/master/docs/Installing.asciidoc#development-version-repository instead. This way we also don't need to change the version reference every year

docs/Contributing.asciidoc
okurz: > That link does not make it clear to me what I should link to. This is just pointing to Contributing.asciidoc

Sorry, it seems I prepared comments and then published them later when you already updated so the links don't match anymore. My suggestion is in this section to include a sentence to "read on" in the following section https://github.com/os-autoinst/openQA/blob/master/docs/Contributing.asciidoc#testing

okurz (changes requested):

docs/Contributing.asciidoc
Wabri: Confirm that is only my problem due to the fact that I'm using different ssh key for different projects and I'm using direnv to load them and in distrobox to use them I need to source the file where they are defined.

docs/Contributing.asciidoc
Wabri: since this is a 5min development environment setup I would prefer to not ping pong over documentations and have everything in the same section/file/page

Wabri (changes requested):

docs/Contributing.asciidoc
Martchus: I would actually mention `tools/js-tidy` for the sake of consistency. It makes sense to explain the raw commands here as only this way one can easily add more arguments, e.g. `--force` is sometimes useful. (If we really wanted to use the Makefile here then the first two commands should probably be replaced with `make tidy-perl` but then we couldn't explain the useful `--git` option.)

perlpunk: @okurz The link https://github.com/os-autoinst/openQA/pull/6590/files#diff-4d6d1050121df87969e32770abd50500a1fbda9f136db4fbb374fe05a236f2c2R59 is just pointing to Contributing.asciidoc. I don't know what you mean.

I removed the General section now.

Note that I moved stuff around. In the current version Perltidy is mentioned explicitly with a link to their website. That link is now gone. The Technologies section is also inconsistent in a way, that it only mentions Perltidy, but not what we use for javascript.

docs/Contributing.asciidoc
perlpunk: this link https://github.com/os-autoinst/openQA/pull/6590#discussion_r2231851548 is pointing to itself...

perlpunk: The new section is useful if you want instructions that you can follow without jumping around and trying to find what you actually have to do, but instead something that you can mostly just copy & paste.

docs/Contributing.asciidoc
perlpunk: done

docs/Contributing.asciidoc
perlpunk: done

docs/Contributing.asciidoc
perlpunk: I changed it a bit now.

docs/Contributing.asciidoc
okurz: > @okurz The link https://github.com/os-autoinst/openQA/pull/6590/files#diff-4d6d1050121df87969e32770abd50500a1fbda9f136db4fbb374fe05a236f2c2R59 is just pointing to Contributing.asciidoc. I don't know what you mean.

I am sorry, that is weird. The URL contains the suffix "R59" which should point to line 59 in the "post-patch" view, i.e. in the file including your changes. This is how Firefox handles the link for me

Screenshot_20250731_220356_make_tidy

> I removed the General section now.
> 
> Note that I moved stuff around. In the current version Perltidy is mentioned explicitly with a link to their website. That link is now gone. The Technologies section is also inconsistent in a way, that it only mentions Perltidy, but not what we use for javascript.

docs/Contributing.asciidoc
```diff
@@ -179,25 +175,86 @@ systemd:: systemd service definitions
...
+We manage our dependencies via a central `dependencies.yaml` file per
+repository. We generate the `.spec` files for `rpm` as well as a `cpanfile`
+for quicker installation of all Perl modules.
+
+[[runtime-dependencies]]
+=== Runtime Dependencies
+
+For openSUSE you need the packages `openQA` and `os-autoinst`.
+That should be everything necessary to run an openQA instance. For more
+detais check  <<Installing.asciidoc#_installation,Installation>.
```
okurz:
```suggestion
details check  <<Installing.asciidoc#_installation,Installation>.
```

docs/Contributing.asciidoc
```diff
@@ -179,25 +175,86 @@ systemd:: systemd service definitions
...
+If you have to manually install packages, look into the `dependencies.yaml`
+file or the spec file, e.g. `dist/rpm/openQA.spec` (or `dist/rpm/*.spec` in
+general). You can leave out sections starting with `test_requires` for example.
+
+A CPAN client will by default also leave out modules for testing and
+development.
+
+There is a <<Installing.asciidoc#single-instance-container,Single-Instance Container>>
+which can be tried out as a minimal installation.
+
```
okurz: Running a container is likely not called an "installation". How about

```suggestion
As alternative you can try out the <<Installing.asciidoc#single-instance-container,Single-Instance Container>>
as a minimal runtime environment.
```

docs/Contributing.asciidoc
```diff
@@ -179,25 +175,86 @@ systemd:: systemd service definitions
...
+[[development-dependencies]]
+=== Development Dependencies
+
+For openSUSE:
+```
+zypper install os-autoinst-devel openQA-devel
```
okurz: Why is openQA-devel not enough? I just tried

```
podman run --rm -it --pull=newer registry.opensuse.org/opensuse/tumbleweed
```

and in there `zypper -n in openQA-devel` and that pulls in also os-autoinst. If I don't want to develop os-autoinst then I don't need os-autoinst-devel. By the way the difference of os-autoinst-devel dependencies that are not in openQA-devel is 239 additional packages.

okurz (dismissed):

docs/Contributing.asciidoc
okurz: I like the current state :+1:

docs/Contributing.asciidoc
okurz: I couldn't find this in the changes anymore. If you removed the relevant part I guess we are good and you can resolve this conversation

docs/Contributing.asciidoc
```diff
@@ -431,10 +489,245 @@ Also find more details in
...
+export PERL5LIB=$OPENQA_BASEDIR/perl5/lib/perl5
+PATH=$OPENQA_BASEDIR/perl5/bin:$PATH
+
+export OPENQA_KEY=1234567890ABCDEF
+export OPENQA_SECRET=1234567890ABCDEF
+----
+
+==== Postgres Container
+
+To be able to keep the database when stopping the container, we create a
```
d3flex: niptick: would it be better if you say `To be able to reuse the database..`?

docs/Contributing.asciidoc
```diff
@@ -431,10 +489,245 @@ Also find more details in
...
+podman volume create mypostgres
+podman run -it --rm --name postgres-openqa \
+    -v mypostgres:/var/lib/postgresql/data \
+    -p 5432:5432 \
+    -e POSTGRES_PASSWORD="x" \
+    -d \
+    docker.io/library/postgres:17
+
+# Optionally add this to mount a directory for sql dumps to import
+#    -v "$SQL_DUMP_DIR:/sqldumps" \
```
d3flex, line 579: Can you make it a Note outside the example?

```
NOTE: Optionally add `-v "$SQL_DUMP_DIR:/sqldumps"` to mount a directory for sql dumps to import
```

docs/Contributing.asciidoc
```diff
@@ -431,10 +489,245 @@ Also find more details in
...
+For running openQA jobs, you need all.
+
+[source,sh]
+----
+### distrobox container ###
+# Main web UI (required for web interface and web API)
+$OPENQA_REPOS/openQA/script/openqa-webui-daemon
+
+# Websockets (for workers to connect so they can receive job assignments)
+openqa-websockets daemon
```
d3flex, line 703: Maybe add `$OPENQA_REPOS/openQA/script/` to the other services as well. All those hae to run from the script/ dir

docs/Contributing.asciidoc
kalikiana: Not sure I ever ran into it either. I would leave this for a follow-up.

docs/Contributing.asciidoc
perlpunk: Actually I removed the path from this one.
Because` $OPENQA_REPOS/openQA/script` is already in `$PATH` if you source the development script.

docs/Contributing.asciidoc
perlpunk: I rephrased it

docs/Contributing.asciidoc
kalikiana: I think it's nice to have these steps all in one place. Install docs are the full reference, though - why not stick one setup here, for example Leap & devel repos, postgres from Debian. And then link "for other options...".

perlpunk: I think I was asked to add Tumbleweed here in a previous review.
I can remove it again...

docs/Contributing.asciidoc
perlpunk: I like it the way it is now for easy moving into the above command

docs/Contributing.asciidoc
kalikiana: Leave it if it helps getting this PR merged 😉

docs/Contributing.asciidoc
```diff
@@ -143,7 +142,7 @@ mechanism, so it must be available on the system.
 The openQA package is built on top of Mojolicious, an excellent Perl framework
 for web development that will be extremely familiar to developers coming from
 other modern web frameworks like Sinatra and that have nice and comprehensive
-documentation available at its http://mojolicio.us[home page].
+documentation available at its https://mojolicious.org/[home page].
```
Wabri, line 145:
```suggestion
documentation available at its https://mojolicious.org/[mojolicious home page].
```

docs/Contributing.asciidoc
```diff
@@ -131,9 +131,8 @@ additional tools are required. The easiest way to install all needed
 dependencies is using the available os-autoinst and openQA packages, as
 described in the Installation Guide.
 
-In the case of os-autoinst, only a few http://www.cpan.org/[CPAN] modules are
-required. Basically `Carp::Always`, `Data::Dump`. `JSON` and `YAML`. On the other
-hand, several external tools are needed including
+In the case of os-autoinst, some http://www.cpan.org/[CPAN] modules are
```
Wabri:
```suggestion
In the case of https://github.com/os-autoinst/os-autoinst[os-autoinst], some http://www.cpan.org/[CPAN] modules are
```

docs/Contributing.asciidoc
```diff
@@ -179,25 +175,87 @@ systemd:: systemd service definitions
...
+repository. We generate the `.spec` files for `rpm` as well as a `cpanfile`
+for quicker installation of all Perl modules.
+
+[[runtime-dependencies]]
+=== Runtime Dependencies
+
+For openSUSE you need the packages `openQA` and `os-autoinst`.
+That should be everything necessary to run an openQA instance. For more
+details check <<Installing.asciidoc#_installation,Installation>.
+
```
Wabri, line 204:
```suggestion
```

docs/Contributing.asciidoc
```diff
@@ -179,25 +175,87 @@ systemd:: systemd service definitions
...
+If you just want to run the test suite to see if it passes on your platform,
+you would not need the developer's dependencies, like `Perl::Tidy` and
+`Devel::Cover`. They are recommended as soon as you plan a contribution.
+
+For the UI tests you need `perl-Selenium-Remote-Driver` as well as
+`chromedriver` and either `chrome` or `chromium`.
+
+See <<Contributing.asciidoc#run-tests-in-container,Run tests in a container>>
+for a container with all dependencies.
+
```
Wabri, line 237:
```suggestion
```

docs/Contributing.asciidoc
```diff
@@ -431,10 +489,245 @@ Also find more details in
...
+You will run two containers:
+
+* A separate postgres container (so you don't have to deal with the database
+  when updating your dev container)
+* The openQA environment in an openSUSE Leap or Tumbleweed container, which
+  both can use the same postgres container.
+  In this example we will be setting up and using the Leap container.
+
+`distrobox`, for example, can be used, to take care of network, home directory,
+permissions, and graphical environment.
```
Wabri: the punctuation in this sentence doesn't make much sense to me, maybe:

```suggestion
For example `distrobox` can be used to take care of: network, home directory,
permissions, and graphical environment.
```

docs/Contributing.asciidoc
```diff
@@ -431,10 +489,245 @@ Also find more details in
...
+We are using a Debian postgres container here, but it doesn't matter what
+you use.
+[source,sh]
+----
+### host ###
+podman volume create mypostgres
+podman run -it --rm --name postgres-openqa \
+    -v mypostgres:/var/lib/postgresql/data \
+    -p 5432:5432 \
+    -e POSTGRES_PASSWORD="x" \
```
Wabri, line 574: Is `POSTGRES_USER` not necessary because of the step below?

docs/Contributing.asciidoc
```diff
@@ -431,10 +489,245 @@ Also find more details in
...
+
+You can use a Leap container and install the requirements, but you can also
+build your own container with the requirements already installed.
+
+[source,sh]
+----
+### host ###
+distrobox create -i registry.opensuse.org/opensuse/leap:15.6 -n box-openqadev-leap
+distrobox enter box-openqadev-leap
+# Tumbleweed:
```
Wabri, line 605:
```suggestion
# Or for Tumbleweed:
```

docs/Contributing.asciidoc
```diff
@@ -431,10 +489,245 @@ Also find more details in
...
+#distrobox create -i registry.opensuse.org/opensuse/tumbleweed:latest -n box-openqadev-tw
+
+### distrobox container ###
+
+# Leap 15.6
+zypper addrepo -G -p 95 https://download.opensuse.org/repositories/devel:/openQA/15.6 devel-openqa \
+    && zypper addrepo -G -p 95 https://download.opensuse.org/repositories/devel:/openQA:/Leap:/15.6/15.6/ devel-openqa-leap
+
+# Tumbleweed
```
Wabri, line 614:
```suggestion
# If you choose Tumbleweed:
```

docs/Contributing.asciidoc
```diff
@@ -431,10 +489,245 @@ Also find more details in
...
+#zypper addrepo -G p 95 https://download.opensuse.org/repositories/devel:/openQA/openSUSE_Tumbleweed devel-openqa \
+#        && zypper addrepo -G -p 95 https://download.opensuse.org/repositories/SUSE:/CA/openSUSE_Tumbleweed/SUSE:CA.repo
+
+zypper -n ref && zypper -n install ca-certificates-suse \
```
Wabri, line 618: why are `ca-certificates-suse` necessary?

docs/Contributing.asciidoc
```diff
@@ -431,10 +489,245 @@ Also find more details in
...
+        os-autoinst-scripts-deps \
+        os-autoinst-scripts-deps-devel \
+        crudini
+# plus your favourite dev tools, e.g. tmux
+----
+Note: https://github.com/pixelb/crudini[`crudini`] is a tool for manipulating
+values in ini files.
+
+Time to grab a hot beverage while installing…
+
```
Wabri: Do we have something similar in other part of the documentation?
I would avoid it otherwise, just to be consistent:

```suggestion
```

Or maybe replace with something like this:

```suggestion
This installations will take time.
```

docs/Contributing.asciidoc
```diff
@@ -431,10 +489,245 @@ Also find more details in
...
+EOM >> $OPENQA_CONFIG/database.ini
+
+cp $OPENQA_REPOS/openQA/etc/openqa/workers.ini $OPENQA_CONFIG/workers.ini
+crudini --set $OPENQA_CONFIG/workers.ini global HOST http://localhost:9526
+crudini --set $OPENQA_CONFIG/workers.ini global WORKER_HOSTNAME 127.0.0.1
+
+cp $OPENQA_REPOS/openQA/etc/openqa/openqa.ini $OPENQA_CONFIG/openqa.ini
+crudini --set $OPENQA_CONFIG/openqa.ini auth method Fake
+----
+
```
Wabri, line 669:
```suggestion
```

docs/Contributing.asciidoc
```diff
@@ -431,10 +489,245 @@ Also find more details in
...
+# Initialize the database schema. This is using the database.ini you
+# created above.
+$OPENQA_REPOS/openQA/script/initdb --init_database
+----
+
+==== Run openQA
+
+Ready!
+
+For playing with the web UI, you just need to run this one daemon.
```
Wabri: > this one daemon

Which one?

docs/Contributing.asciidoc
```diff
@@ -431,10 +489,245 @@ Also find more details in
...
+For running openQA jobs, you need all.
```
Wabri:
```suggestion
For running openQA jobs, you need to run all of this daemons:
```

docs/Contributing.asciidoc
```diff
@@ -431,10 +489,245 @@ Also find more details in
...
+
+# Gru (for running background tasks as Minion jobs)
+openqa gru run
+
+# Livehandler (for the developer mode on the web UI)
+openqa-livehandler daemon
+
+# Worker
+worker --isotovideo "$OPENQA_REPOS/os-autoinst/isotovideo" --instance 1 --verbose --apikey $OPENQA_KEY --apisecret $OPENQA_SECRET
+
```
Wabri:
```suggestion
```

docs/Contributing.asciidoc
```diff
@@ -431,10 +489,245 @@ Also find more details in
...
+----
+
+For running unit tests, see <<Contributing.asciidoc#testing,Conducting tests>>.
+
```
Wabri, line 719:
```suggestion
```

Wabri (changes requested):

docs/Contributing.asciidoc
perlpunk: It's using the default user `postgres`

docs/Contributing.asciidoc
perlpunk: This is getting endless now...

docs/Contributing.asciidoc
perlpunk: to access urls like osd

docs/Contributing.asciidoc
Wabri: It was just a suggestion... you can keep like it is if there were already a discussion over this!

docs/Contributing.asciidoc
perlpunk: If we are talking about Mojolicious in this paragraph, then `its Mojolicious home page` sounds weird.

docs/Contributing.asciidoc
Wabri: So it is mostly for internals suse employees, aren't we mixing too many things? Maybe we can create a section just for it? but maybe it is too much. Just wondering if it make sense for peoples that are not suse employees that want to contribute to the project.

perlpunk: I wanted something for people joining the tools team to get up and running as fast as possible.

Maybe I should close this PR and put it into an extra repo.

kalikiana: It's a publicly accessible repo fwiw

docs/Contributing.asciidoc
```diff
@@ -431,10 +487,240 @@ Also find more details in
...
+# Initialize the database schema. This is using the database.ini you
+# created above.
+$OPENQA_REPOS/openQA/script/initdb --init_database
+----
+
+==== Run openQA
+
+Ready!
+
+For playing with the web UI, you just need to run the openqa-webui-daemon.
```
Wabri:
```suggestion
For playing with the web UI, you just need to run the `openqa-webui-daemon`.
```

Wabri (approved):

perlpunk:
helm chart test is failing in multiple PRs now:
https://github.com/os-autoinst/openQA/actions/runs/16933683466/job/47985069408?pr=6590
```
Error: searching log query: Post "https://rekor.sigstore.dev/api/v1/log/entries/retrieve": POST https://rekor.sigstore.dev/api/v1/log/entries/retrieve giving up after 4 attempt(s): Post "https://rekor.sigstore.dev/api/v1/log/entries/retrieve": dial tcp 34.36.47.134:443: i/o timeout
```

docs/Contributing.asciidoc
Martchus: Considering it is a public repo and one has to add additional repos anyway I don't think that makes it worse for anybody. So I'm not against having this here.

PR #6591, opened 2025-07-16
Add openqa-config script for showing/changing config values

perlpunk:
Related issue: https://progress.opensuse.org/issues/184459

I was tired of either having to add instructions to manually edit config files, or to have to use some `sed` statement to manipulate a config file.

I think there can be made some improvements regarding the output.
Right now it outputs human readable YAML, but I think it could also output JSON when requested, so you could process a section or the whole config file with `jq` for example.

### DIFF
--- a/lib/OpenQA/Config.pm
+++ b/lib/OpenQA/Config.pm
@@ -8,8 +8,13 @@ use Config::IniFiles;
 use Exporter qw(import);
 use OpenQA::Log qw(log_info);
 use Mojo::File qw(path);
+use File::Copy qw(copy);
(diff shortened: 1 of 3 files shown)


perlpunk:
Note that there are some general ini manipulation scripts out there. I haven't tried them out yet.
The advantage here could be that we could also show default values from OpenQA::Setup on request.

perlpunk:
There is https://github.com/pixelb/crudini which is in Leap 15.6.
It doesn't have an option for JSON output unfortunately. But maybe they accept pull requests...

PR #6603, opened 2025-07-22
Support keeping only jobs in database (while removing all logs and test results sooner)

Martchus:
See https://progress.opensuse.org/issues/179263 and particular commit messages


Martchus:
Oh, and I would really like to avoid repeating the approach we took for archiving. Here we archive jobs which are "preserved because they are important". While it works for archiving I must say that adding this kind of "condition" does add more complexity and makes things therefore harder to understand. Adding just another configurable duration is on the other hand just more of the same and doesn't add complexity and thus doesn't make things harder to understand.

okurz:
> I think this feature was intended with the opposite thought in mind: The database entry of a job needs not much disk space but results potentially a lot. So we can store much more database entries than results. It seems unnecessary to delete jobs from the database (and thus not having them anymore on dashboard pages) when we really only run out of result storage. This change helps preserving jobs in the database longer. It is not meant to delete data more aggressively.

Exactly. But I think a "job group owner" neither understands this distinction nor would be able to derive what that could mean for their job group. Obviously with the configurations for "keep results" and "keep jobs" next to each other I would assume that users compare those values and try to select something sensible but they can't know what effect on available database storage that actually has.

> Oh, and I would really like to avoid repeating the approach we took for archiving. Here we archive jobs which are "preserved because they are important". While it works for archiving I must say that adding this kind of "condition" does add more complexity and makes things therefore harder to understand. Adding just another configurable duration is on the other hand just more of the same and doesn't add complexity and thus doesn't make things harder to understand.

I think "archiving" is a good example of leaving the decision to the administrators of an instance which are also the ones need to ensure that storage space is not depleted. How about a simpler approach but still leaving the choice to the administrator: Similar to our implementation for "space aware cleanup" only clean jobs from the database when it's actually necessary. As you explained it can be expected that the database are residing on different storage than results so how about two config settings for number of jobs as well as a threshold to keep free storage configured in the ini file with sane default and only remove jobs when any of those two limits are reached?

lib/OpenQA/JobGroupDefaults.pm
```diff
@@ -12,6 +12,8 @@ use constant {
     KEEP_IMPORTANT_LOGS_IN_DAYS => 120,
     KEEP_RESULTS_IN_DAYS => 365,
     KEEP_IMPORTANT_RESULTS_IN_DAYS => 0,
+    KEEP_JOBS_IN_DAYS => 365,
```
okurz, line 15: I assume you selected 365 so that the behaviour (by default) is effectively the same as in before, right?
I think that is a good idea.

So for later how about making use of the new feature by default by selecting an according number higher than "KEEP_RESULTS_IN_DAYS"

```suggestion
    KEEP_JOBS_IN_DAYS => 730,
```

If you think this is too much then I suggest to lower KEEP_RESULTS_IN_DAYS  to something like 200 and set KEEP_JOBS_IN_DAYS to 365 then.

Martchus:
> Yes, that is the current approach and I find it suboptimal as it is a constant back and forth between users and admins. I would like to find a way where users and admins don't need to argue over the same values but all use different parameters that they care about.

Ok - although I don't see how adding another parameter makes things worse now.

> Think of a scenario where we would have a ten times bigger openQA instance or compare to github. There are also no settings in github projects that you and me set and sometimes github admins would contact us and say "hey, we need to lower those values you set".

For 10x bigger we would probably need a different database setup and probably a different setup overall. Having everything served by a single host probably won't cut it anymore. I think GitHub and other services can grow in such an unlimited way because they can just add more servers.  I don't think we should make very advanced/big/expensive setups like these the next step for openQA, though. We can still help users/admins with simple setups and more limited resources.

By the way, I was recently curious how much disk space one has on GitHub as I upload assets for my releases there. It turns out GitHub actually [does ask their users to stay within limits](https://docs.github.com/en/repositories/working-with-files/managing-large-files/about-large-files-on-github) to avoid performance problems.

And another by the way, many other big platforms directly confront users with limits simply to ask them to pay at some point. I'm not sure whether big platforms are generally a good comparison but this just shows that users often are confronted with limits. So it is not like there's "a perfect world out there" and we need to reach the same level - as it seems you want me to believe.

> Please take a look into recent problems we had with overuse …

Ok, I'm not reading these different threads right now. However, I think I know some of them anyway. And that is also why I propose to improve the help popover here. (And because I simply couldn't add anything to the existing popover without breaking its currently terrible structure.)

> ok. So as this PR is for a spike solution how about doing the "important" clarification first and then think about how to do "keeping only jobs" in a better way.

Ok, I can create a separate PR for UI improvements first. It'll be easier and more straight forward it I was able to base further changes on this PR, though.

> Yes. I am aware. That's why I created the ticket assuming that we have many building blocks in place already.

But considering we already delete only results first as part of the space aware cleanup *and* you are against the additions I am proposing here I'm not really sure what you are asking for. Some "magic solution" to make things better probably. Sorry, but all your suggestions so far seem impractical and really leading much further than anyone reading the ticket you've created could have guessed. (At least we didn't guess that when moving the ticked into the backlog. At this point @kalikiana and @perlpunk  and me actually briefly talked about this.)

> But I am still concerned about asking users to make choices which would be more suitable for admins.

I think collapsing these settings into an advanced/admin section will be good enough. Of course it isn't ideal that all these settings are mixed on a single page even though different user groups probably care about them to varying degrees. There is probably no sharp line that can be drawn between admins and users in our case so it also makes kind of sense. (And it also makes sense because distributing these settings would make it much harder to find them in general.)

> If after reading all of the above you still think that openQA would scale well - "think ten times bigger" - with this current implementation including the UI integration then I will approve.

I think openQA will not scale 10x bigger regardless of this particular change. However, I don't think this change will get in the way of scalability. As I mentioned before, this is really not adding anything new. Just another setting on a level where we already have very similar settings.

> We could simply leave out all job group attributes and UI changes for now and only keep the database rows and default values as they can still be set over the config and should have a good default. Then we can collect experience, improve the definition of "important" and then come back to deciding about which UI integration is useful.

Having to change this in the database manually (and having to use the database just to check the current values) seems very cumbersome. Would it be ok for you if I collapse these two new settings as mentioned before? If not, would it be ok for you if I hid these settings by default but show them only when a certain query parameter is specified? If the answer to any of those questions is yes than I would implement that and also additional improvements to the help popover only top of that (because it is easier).

okurz:
> Would it be ok for you if I collapse these two new settings as mentioned before?

Yes, ok. After you provided your counter points on all previous statements and still think it's good to bring in the UI changes I can also accept the current state of this PR as is without additional folding considering that by default all the job group properties are anyway already "folded away".

> If not, would it be ok for you if I hid these settings by default but show them only when a certain query parameter is specified?

Not necessary, see above

> If the answer to any of those questions is yes than I would implement that and also additional improvements to the help popover only top of that (because it is easier).

Fine with the order of implementing that. As soon as conflicts and CI checks are resolved I will approve the PR without needing further changes to the UI

docs/GettingStarted.asciidoc
```diff
@@ -371,36 +371,10 @@ the index page.
 
 [id="basic_cleanup"]
 === Cleanup
-IMPORTANT: openQA automatically deletes data that it considers "old" based on
+openQA automatically archives and deletes data that it considers "old" based on
```
okurz, line 374: "archives and deletes data" at the same time? More like "archives or deletes", right?

okurz (changes requested):

docs/GettingStarted.asciidoc
Martchus: Archiving can only happen when deletion/cleanup is enabled at the same time. So the features are not exclusive options but "or" makes it sound a bit like that. My thinking with "and" was: So some data gets archived *and* some deleted. Of course a specific piece of data cannot be archived and deleted at the same time but this sentence talks about data in a very generic sense (hence also the generic term "data").

okurz: Ok, good. Keep it

lib/OpenQA/WebAPI/Controller/API/V1/JobGroup.pm
```diff
@@ -182,27 +182,34 @@ sub _validate_common_properties ($self) {
...
-    return 0;
+    for my $important ('', '_important') {
+        my $log_key = "${prefix}keep${important}_logs_in_days";
+        my $result_key = "${prefix}keep${important}_results_in_days";
+        my $job_key = "${prefix}keep${important}_jobs_in_days";
+        my $log_value = $properties->{$log_key} // ($group ? $group->$log_key : 0);
+        my $result_value = $properties->{$result_key} // ($group ? $group->$result_key : 0);
+        my $job_value = $properties->{$job_key} // ($group ? $group->$job_key : 0);
+        push @errors, "'$log_key' must be <= '$result_key'" if $result_value != 0 && $log_value > $result_value;
+        push @errors, "'$result_key' must be <= '$job_key'" if $job_value != 0 && $result_value > $job_value;
```
Martchus, line 201: By the way, depending on user feedback we could also simply extend the too low values here. This way users would be confronted with fewer errors and openQA would just do what they had *probably* in mind anyway. We could still show a reworded  version of these messages as information to make it clear what happened.

okurz:
> > So for later how about making use of the new feature by default by selecting an according number higher than "KEEP_RESULTS_IN_DAYS"
> 
> I think putting in a higher number as default would make sense. I assume with "later" you mean in a follow-up PR?

Yes, please.

Martchus:
Ok, I'll create a follow up once this PR has been merged.

okurz (approved):

lib/OpenQA/WebAPI/Controller/API/V1/JobGroup.pm
```diff
@@ -182,10 +182,14 @@ sub _validate_common_properties ($self) {
...
     $validation->optional('default_keep_results_in_days')->num(0);
     $validation->optional('default_keep_important_results_in_days')->num(0);
+    $validation->optional('default_keep_jobs_in_days')->num(0);
+    $validation->optional('default_keep_important_jobs_in_days')->num(0);
     $validation->optional('keep_logs_in_days')->num(0);
     $validation->optional('keep_important_logs_in_days')->num(0);
     $validation->optional('keep_results_in_days')->num(0);
     $validation->optional('keep_important_results_in_days')->num(0);
+    $validation->optional('keep_jobs_in_days')->num(0);
+    $validation->optional('keep_important_jobs_in_days')->num(0);
```
perlpunk: I guess at this point it could make sense to shorten this:
```perl
for my $field (qw(keep_logs_in_days keep_important_logs_in_days
  keep_results_in_days keep_important_results_in_days
  keep_jobs_in_days keep_important_jobs_in_days)) {
    $validation->optional($field)->num(0);
    $validation->optional("default_$field")->num(0);
}
```

perlpunk: or even
```perl
for my $field (qw(logs important_logs results important_results jobs important_jobs)) {
    $validation->optional("keep_${field}_in_days")->num(0);
    $validation->optional("default_keep_${field}_in_days")->num(0);
}
```

lib/OpenQA/Setup.pm
```diff
@@ -298,6 +304,23 @@ sub _validate_worker_timeout ($app) {
...
+sub _set_default_storage_durations ($limits) {
+    # assign according constant as default but never assign a lower value than the one for the smaller cleanup
+    $limits->{result_storage_duration}
+      //= _max_limit($limits->{log_storage_duration}, OpenQA::JobGroupDefaults::KEEP_RESULTS_IN_DAYS);
+    $limits->{important_result_storage_duration} //= _max_limit($limits->{important_log_storage_duration},
+        OpenQA::JobGroupDefaults::KEEP_IMPORTANT_RESULTS_IN_DAYS);
+    $limits->{job_storage_duration}
+      //= _max_limit($limits->{result_storage_duration}, OpenQA::JobGroupDefaults::KEEP_JOBS_IN_DAYS);
+    $limits->{important_job_storage_duration} //= _max_limit($limits->{important_result_storage_duration},
+        OpenQA::JobGroupDefaults::KEEP_IMPORTANT_JOBS_IN_DAYS);
```
d3flex: This looks like it could be in a for loop. Can we write something like

```suggestion
my @limit_defs = (
    ['result_storage_duration',  'log_storage_duration', OpenQA::JobGroupDefaults::KEEP_RESULTS_IN_DAYS],
    ['important_result_storage_duration', 'important_log_storage_duration', OpenQA::JobGroupDefaults::KEEP_IMPORTANT_RESULTS_IN_DAYS],
    ['job_storage_duration', 'result_storage_duration', penQA::JobGroupDefaults::KEEP_JOBS_IN_DAYS],
    ['important_job_storage_duration', 'important_result_storage_duration', OpenQA::JobGroupDefaults::KEEP_IMPORTANT_JOBS_IN_DAYS],
);

for my $limit (@limit_defs) {
    my ($key, $settings, $default) = @$limit;
    $limits->{$key} //= _max_limit($limits->{$settings}, $default);
}
```

lib/OpenQA/Schema/Result/JobGroups.pm
```diff
@@ -275,9 +280,12 @@ sub find_jobs_with_expired_logs ($self, $preserved_important_jobs_out = undef) {
 
 # helper function for cleanup task
 sub limit_results_and_logs ($self, $preserved_important_jobs_out = undef) {
-    my $expired_jobs = $self->find_jobs_with_expired_results;
+    my $expired_jobs = $self->find_expired_jobs;
     $_->delete for @$expired_jobs;
 
+    my $expired_results_jobs = $self->find_jobs_with_expired_results;
+    $_->delete_results for @$expired_results_jobs;
```
perlpunk, line 287: We are currently also calling `delete_results` as part of the `ensure_results_below_threshold` task:
https://github.com/os-autoinst/openQA/blob/master/lib/OpenQA/Task/Job/Limit.pm#L237
https://github.com/os-autoinst/openQA/blob/master/lib/OpenQA/Task/Job/Limit.pm#L253

Is it intentional that we have two places for that now?

Martchus: Yes, that's intentional. That's now a 2nd place - besides the space-aware cleanup - where we delete results but keep the job in the database.

The space-aware cleanup is for deleting results more aggressively so they are deleted *sooner* as needed when free disk space is falling below a threshold. It doesn't consider configured retentions anymore so it is really more a last measure. The new property allows to be more explicit about this with configured retentions so jobs can be kept in the database *longer*. So we're tackling the same problem from two different angles here.

lib/OpenQA/Setup.pm
Martchus: This is a good idea. I changed it but didn't make the `_storage_duration` suffix part of the arrays because otherwise tidy would reformat them not so nicely.

Martchus:
Actually, with the database migration in place that would assign the current value of `keep_results_in_days` to `keep_jobs_in_days` we would not see any difference if we later increase the constant of `KEEP_JOBS_IN_DAYS` for existing groups. Only new groups relying on defaults would be affected.

I'm not sure whether that's what we want. I could change the assignment in that migration to happen only if `keep_results_in_days` is already bigger than `KEEP_JOBS_IN_DAYS`. Then we would change the cleanup behavior for existing groups to store job in the database longer (365 days) if `keep_results_in_days` was reduced (to be below 365 days).

Note that if we merge this as-is we can still easily keep jobs longer in the database for all groups afterwards. Just bumping `KEEP_JOBS_IN_DAYS`  won't be enough but with a simple database query we can make these groups use the default again:

```
UPDATE job_groups SET keep_jobs_in_days = null WHERE keep_jobs_in_days <= 365;
UPDATE job_group_parents SET default_keep_jobs_in_days = null WHERE default_keep_jobs_in_days <= 365;
```

lib/OpenQA/Setup.pm
perlpunk: It seems to result in even more lines, though, according to the diff

PR #6610, opened 2025-07-24
Allow empty values in openqa-cli parameters

perlpunk:
openqa-cli --apibase ''

did not set the parameter as expected

Issue: https://progress.opensuse.org/issues/186146

Bug introduced in 47c3c7bec

### DIFF
--- a/t/43-cli-api.t
+++ b/t/43-cli-api.t
@@ -110,6 +110,10 @@ subtest 'API' => sub {
     throws_ok { $api->run(@auth) } qr/Usage: openqa-cli api/, 'usage';
     is $api->apikey, 'ARTHURKEY01', 'apikey';
     is $api->apisecret, 'EXCALIBUR', 'apisecret';
+
+    $api = OpenQA::CLI::api->new;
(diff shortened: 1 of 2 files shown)


t/43-cli-api.t
```diff
@@ -110,6 +110,10 @@ subtest 'API' => sub {
     throws_ok { $api->run(@auth) } qr/Usage: openqa-cli api/, 'usage';
     is $api->apikey, 'ARTHURKEY01', 'apikey';
     is $api->apisecret, 'EXCALIBUR', 'apisecret';
+
+    $api = OpenQA::CLI::api->new;
+    throws_ok { $api->run('--apibase', '') } qr/Usage: openqa-cli api/, 'usage';
```
kalikiana, line 115: This looks very strange to me... but I guess this is a regression and this time we have proper test coverage.

kalikiana (approved):

t/43-cli-api.t
perlpunk: these tests are missing actual parameters (they only have an option as parameter), that's why they fail. but still it can be tested if the apibase parameter was set correctly

Martchus (approved):

PR #6616, opened 2025-07-28
Show note about sorting on group overview pages

Martchus:
Now with three different sorting options and one of them affecting time stamps it makes sense to add an according note on the group overview page.

Related ticket: https://progress.opensuse.org/issues/185566

---

This is how it looks like:

screenshot_20250728_162824

This screenshot shows that the sorting doesn't fully work as one would expect, though. At least I would expect the build from 2 hours ago to be at the top. Possibly the version is also taken into account somehow in this case on my local instance.

### DIFF
--- a/lib/OpenQA/WebAPI/Controller/Main.pm
+++ b/lib/OpenQA/WebAPI/Controller/Main.pm
@@ -6,6 +6,7 @@ use Mojo::Base 'Mojolicious::Controller', -signatures;
 use Feature::Compat::Try;
 
 use Date::Format;
+use OpenQA::Constants qw(BUILD_SORT_BY_NAME BUILD_SORT_BY_NEWEST_JOB BUILD_SORT_BY_OLDEST_JOB);
 use OpenQA::Jobs::Constants;
 use OpenQA::Schema::Result::Jobs;
 use OpenQA::BuildResults;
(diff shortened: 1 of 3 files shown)


Martchus:
> If you really chose sorting by newest, then this is a bug.

I really chose the newest. And I think this also works in simpler cases. I'm just not sure yet what makes my test jobs problematic yet.

lib/OpenQA/WebAPI/Controller/Main.pm
```diff
@@ -72,6 +73,18 @@ sub _respond_error_for_group_overview ($self, $error) {
     );
 }
 
+sub _sort_info_time ($new_old) { "Builds are sorted by the creation time of the <em>$new_old</em> job." }
+
+sub _sort_help_timestamps ($new_old) {
+"This means the timestamps are the creation time of the <em>$new_old</em> job in that build (accross all architectures).";
```
Wabri, line 79: Are there any reasons for 0 indentation?

```suggestion
    "This means the timestamps are the creation time of the <em>$new_old</em> job in that build (accross all architectures).";
```

Martchus: Yes, tidy prefers it this way.

PR #559, opened 2016-02-18
t: Catch expected warnings in 05-scheduler-iso

okurz:
Cleanup test to not spill expected warnings. Also use better test comparison
methods with better feedback on the expected and actual values when a test
fails.

PR #6619, opened 2025-07-29
Double default storage duration for jobs in database

Martchus:
Let's try to keep jobs longer in the database by default now that this can be configured independently of job results.

Related ticket: https://progress.opensuse.org/issues/179263


perlpunk (approved):

Martchus:
I would wait with this until the previous PR has been deployed and we can assume that there are no regressions.

Then, to make this effective we need to let existing job groups follow this new default:

```
UPDATE job_groups SET keep_jobs_in_days = null WHERE keep_jobs_in_days <= 720 and keep_jobs_in_days != 0;
UPDATE job_group_parents SET default_keep_jobs_in_days = null WHERE default_keep_jobs_in_days <= 720 and default_keep_jobs_in_days != 0;
```

By the way, the `WHERE … <= 720` check is mainly for the sake of completeness because this would update almost all groups in practice anyway. (Out of curiosity I checked what groups exceed the current default for results on OSD and o3 via  `keep_results_in_days >= 365 or keep_results_in_days = 0` and the corresponding parent group setting. There is only one job group¹ that exceeds this value with its setting of 1095 days.)

---

¹ That is the deprecated group "Maintenance: Virtualization - VMware + Hyper-V -- deprecated by group #559".

Martchus:
The cleanup and UI still looks good on o3 and OSD after my recent PR. (On o3 the cleanup is skipped because we're under the configured threshold, though.)

So I'm removing the "not-ready" label here. I can also execute the SQL commands if that's wanted *now*.

PR #6627, opened 2025-07-31
Add perlpod for job_settings/jobs

d3flex:
Provide also an example in the openqa-cli help command.

poo: https://progress.opensuse.org/issues/180296

### DIFF
--- a/lib/OpenQA/WebAPI/Controller/API/V1/JobSettings.pm
+++ b/lib/OpenQA/WebAPI/Controller/API/V1/JobSettings.pm
@@ -4,6 +4,38 @@
 package OpenQA::WebAPI::Controller::API::V1::JobSettings;
 use Mojo::Base 'Mojolicious::Controller', -signatures;
 
+=head2 jobs
+
+Filters jobs based on a single setting key/value pair.
+
+  openqa-cli api --osd job_settings/jobs key="*_TEST_ISSUES" list_value=39911
+  openqa-cli api --osd job_settings/jobs key="*" list_value=39911
+
+=head3 Arguments
+
+=over 4
+
+=item C<key>
+
+The setting key to filter by. It accepts a string or a glob of a job variable.
+Some variables are not stored and they can not be used.
+Examples are DISTRI, BUILD and ARCH
+
+=item C<list_value>
+
+The value to match for the given key. This can be a string without special characters.
+
+=back
+
+=head3 Returns
+
+On success, returns an array of the matched job ids. The results rely on
+some openqa settings C<{misc_limits}{job_settings_max_recent_jobs}>.
+
+On failure, returns a 400 code on invalid input of key or value.
+
+=cut
+
 sub jobs ($self) {
     my $validation = $self->validation;
     $validation->required('key')->like(qr/^[\w\*]+$/);
--- a/public/openqa-cli.yaml
+++ b/public/openqa-cli.yaml
@@ -119,6 +119,9 @@ subcommands:
         # Post job template (from JSON file)
         openqa-cli api --data-file form.json -X POST job_templates_scheduling/1
 
+        # Get jobid(s) filtered by key/value job settings
+        openqa-cli api --osd job_settings/jobs key="*_TEST_ISSUES" list_value=39911
+
     options:
       - header|a=s@           --One or more additional HTTP headers (<name:value>)
       - name: data-file


lib/OpenQA/WebAPI/Controller/API/V1/JobSettings.pm
```diff
@@ -4,6 +4,38 @@
 package OpenQA::WebAPI::Controller::API::V1::JobSettings;
 use Mojo::Base 'Mojolicious::Controller', -signatures;
 
+=head2 jobs
```
perlpunk: This doesn't show up as a description on https://openqa.opensuse.org/404 btw.

For that it has to be a single `=item`.
The doc generator does not support advanced formatting.

Just have a look at the other API routes and their pod entry.

lib/OpenQA/WebAPI/Controller/API/V1/JobSettings.pm
```diff
@@ -4,6 +4,38 @@
...
+
+=head3 Arguments
+
+=over 4
+
+=item C<key>
+
+The setting key to filter by. It accepts a string or a glob of a job variable.
+Some variables are not stored and they can not be used.
+Examples are DISTRI, BUILD and ARCH
```
perlpunk: I wouldn't add too much details here.

docs/UsersGuide.asciidoc
```diff
@@ -1021,6 +1021,28 @@ by adding additional query parameters, e.g.:
...
   as well.
 
+=== Finding tests by quering particular job settings
+
+In case you are looking for a way to get a list of all the tests based on their
+settings, here two ways to achieve it.
+
+[source,sh]
+----
+curl -s "https://openqa.suse.de/api/v1/job_settings/jobs?key=*_TEST_ISSUES&list_value=39911"
```
perlpunk: You are using a lot of `--osd` / openqa.suse.de in the examples you added.
We should prefer to use public urls.
Also I don't see why we should add both the curl and the openqa-cli usage here.
If we do that for every API call, the docs will get noisy.
Just describe the `openqa-cli` call. And leave off the `--osd` switch.

docs/UsersGuide.asciidoc
```diff
@@ -1021,6 +1021,28 @@ by adding additional query parameters, e.g.:
...
+----
+
+or with the power of the `openqa-cli api`
+
+[source,sh]
+----
+openqa-cli api --osd job_settings/jobs key="*_TEST_ISSUES" list_value=39911
+----
+
+It is possible to use a literal string or a glob for C<key> as is shown in the
```
perlpunk: This is a bit of duplication to the pod section in the code.
This will improve when we switch to OpenAPI at some point.

lib/OpenQA/WebAPI/Controller/API/V1/JobSettings.pm
```diff
@@ -4,6 +4,38 @@
...
+The value to match for the given key. This can be a string without special characters.
+
+=back
+
+=head3 Returns
+
+On success, returns an array of the matched job ids. The results rely on
+some openqa settings C<{misc_limits}{job_settings_max_recent_jobs}>.
+
+On failure, returns a 400 code on invalid input of key or value.
```
perlpunk: A 400 Bad Request is a standard response for API calls in case of invalid parameters. We should not add this for every API call.
Also "On failure" and "on invalid input" in the same sentence is confusing.
A 400 Bad Request is not really a failure.
Just remove this.

docs/UsersGuide.asciidoc
d3flex: I didnt change this one. I think this is fine and part of the documentation. or it should be

perlpunk (changes requested):
Almost none of my comments were addressed

docs/UsersGuide.asciidoc
```diff
@@ -1021,6 +1021,21 @@ by adding additional query parameters, e.g.:
   `openSUSE Leap 15 ARM`. This applies to parameters for filtering job settings
   as well.
 
+=== Finding tests by quering particular job settings
```
Martchus:
```suggestion
=== Finding tests by querying particular job settings
```

docs/UsersGuide.asciidoc
```diff
@@ -1021,6 +1021,21 @@ by adding additional query parameters, e.g.:
...
+
+In case you are looking for a way to get a list of all the tests based on their
+settings, here one way to achieve with the power of the `openqa-cli api`
```
Martchus:
```suggestion
settings, here is one way to achieve this with `openqa-cli api`:
```

docs/UsersGuide.asciidoc
```diff
@@ -1021,6 +1021,21 @@ by adding additional query parameters, e.g.:
...
+
+[source,sh]
+----
+openqa-cli api job_settings/jobs key="*_TEST_ISSUES" list_value=39911
+----
+
+It is possible to use a literal string or a glob for C<key> as is shown in the
+examples. We cannot use glob for the C<list_value>.
+OpenQA configuration sets C<job_settings_max_recent_jobs> value which limits
```
Martchus:
```suggestion
openQA configuration sets C<job_settings_max_recent_jobs> value which limits
```

docs/UsersGuide.asciidoc
```diff
@@ -1021,6 +1021,21 @@ by adding additional query parameters, e.g.:
...
+
+In case you are looking for a way to get a list of all the tests based on their
+settings, here one way to achieve with the power of the `openqa-cli api`
+
+[source,sh]
+----
+openqa-cli api job_settings/jobs key="*_TEST_ISSUES" list_value=39911
+----
+
+It is possible to use a literal string or a glob for C<key> as is shown in the
```
Martchus:
```suggestion
It is possible to use a literal string or a glob for C<key> as shown in the
```

docs/UsersGuide.asciidoc
```diff
@@ -1021,6 +1021,21 @@ by adding additional query parameters, e.g.:
...
+examples. We cannot use glob for the C<list_value>.
```
Martchus:
```suggestion
examples. We cannot use a glob in the C<list_value>.
```

I would also avoid using phrasings like "We …" in documentation.

```suggestion
examples. C<list_value> does not support globbing, though.
```

lib/OpenQA/WebAPI/Controller/API/V1/JobSettings.pm
```diff
@@ -4,6 +4,33 @@
...
+=item C<list_value>
+
+The value to match for the given key. This can be a string without special characters.
+
+=back
+
+=head3 Returns
+
+On success, returns an array of the matched job ids. The results rely on
+some openqa settings C<{misc_limits}{job_settings_max_recent_jobs}>.
```
Martchus: It makes no sense to end this sentence with "some openqa settings" and the mentioning of a concrete setting at the same time.

public/openqa-cli.yaml
```diff
@@ -119,6 +119,9 @@ subcommands:
         # Post job template (from JSON file)
         openqa-cli api --data-file form.json -X POST job_templates_scheduling/1
 
+        # Get jobid(s) filtered by key/value job settings
```
Martchus:
```suggestion
        # Get job ID(s) filtered by job settings
```
(I would omit the "key/value" part because the next line makes it clear how to specify the job settings anyway.)

Martchus (changes requested):

lib/OpenQA/WebAPI/Controller/API/V1/JobSettings.pm
```diff
@@ -4,6 +4,33 @@
...
+=over 4
+
+=item jobs
+
+Filters jobs based on a single setting key/value pair.
+
+  openqa-cli api --osd job_settings/jobs key="*_TEST_ISSUES" list_value=39911
+  openqa-cli api --osd job_settings/jobs key="*" list_value=39911
+
+=item C<key>
```
perlpunk, line 16: The only caveat here is that now the `jobs` item is at the same indentation level as the `key` and `list_value` items. But I guess we can live we that, and we can hopefully move to openAPI "soon".

lib/OpenQA/WebAPI/Controller/API/V1/JobSettings.pm
```diff
@@ -4,6 +4,33 @@
...
+The setting key to filter by. It accepts a string or a glob of a job variable.
+Some variables are not stored and they can not be used.
+
+=item C<list_value>
+
+The value to match for the given key. This can be a string without special characters.
+
+=back
+
+=head3 Returns
```
perlpunk: This is introducing a new header, basically separating the previous items from the content after the header.
you probably wanted an `item` here as well...

docs/UsersGuide.asciidoc
```diff
@@ -1021,6 +1021,21 @@ by adding additional query parameters, e.g.:
...
+settings, settings, here is one way to achieve this with `openqa-cli api`:
+
+[source,sh]
+----
+openqa-cli api job_settings/jobs key="*_TEST_ISSUES" list_value=39911
+----
+
+It is possible to use a literal string or a glob for C<key> as shown in the
+examples. C<list_value> does not support globbing, though.
+openQA configuration sets C<job_settings_max_recent_jobs> value which limits
```
Martchus:
```suggestion
openQA configuration sets C<job_settings_max_recent_jobs> which limits
```
(If you wanted to keep "value" you needed a "the" before the variable name.)

Martchus (changes requested):

perlpunk:
https://app.circleci.com/pipelines/github/os-autoinst/openQA/17854/workflows/e90ef135-2d80-4aa3-86c2-bfc5dfb9b0c4/jobs/175051
```
#   Failed test 'POD test for lib/OpenQA/WebAPI/Controller/API/V1/JobSettings.pm'
#   at /usr/lib/perl5/vendor_perl/5.26.1/Test/Pod.pm line 187.
# lib/OpenQA/WebAPI/Controller/API/V1/JobSettings.pm (27): '=item' outside of any '=over'
# Looks like you failed 1 test of 226.
[08:59:47] t/02-pod.t ................................................ 
Failed 1/226 subtests 
```

lib/OpenQA/WebAPI/Controller/API/V1/JobSettings.pm
```diff
@@ -4,6 +4,35 @@
...
+
+The setting key to filter by. It accepts a string or a glob of a job variable.
+Some variables are not stored and they can not be used.
+
+=item C<list_value>
+
+The value to match for the given key. This can be a string without special characters.
+
+=back
+
```
perlpunk:
```suggestion
```

Martchus (approved):

lib/OpenQA/WebAPI/Controller/API/V1/JobSettings.pm
perlpunk: You removed the wrong one

perlpunk: Note that you can run the 02-pod.t test locally

perlpunk (approved):

perlpunk:
@mergifyio rebase

PR #6630, opened 2025-08-03
Proof of Concept - Demonstrate OpenAPI::Modern plugin

perlpunk:
Issue: https://progress.opensuse.org/issues/186675

Notes:
When porting only single routes, they have to be specified explicitly. Once all routes are ported, there only needs to be one route per controller:

    $api_public_r->any('/jobs*')->to('job_group#catchall');

Contra:
* Does not yet support application/x-www-form-urlencoded, so PUT/POST won't work. We would have to move to application/json requests
* Some dependencies not yet in Tumbleweed, even more not in 15.6


lib/OpenQA/Shared/Controller/OpenAPI.pm
```diff
@@ -0,0 +1,23 @@
+# Copyright 2019 SUSE LLC
```
Martchus, line 1:
```suggestion
# Copyright SUSE LLC
```

lib/OpenQA/WebAPI.pm
```diff
@@ -298,7 +306,7 @@ sub startup ($self) {
 
     # api/v1/job_groups
     $api_public_r->get('/job_groups')->name('apiv1_list_job_groups')->to('job_group#list');
-    $api_public_r->get('/job_groups/<group_id:num>')->name('apiv1_get_job_group')->to('job_group#list');
```
Martchus, line 301: I guess the rout name (`apiv1_get_job_group`) could stay.

lib/OpenQA/WebAPI/Controller/API/V1/JobGroup.pm
```diff
@@ -7,6 +7,8 @@ use Mojo::JSON;
 use Feature::Compat::Try;
 
 use OpenQA::Schema::Result::JobGroups;
+use OpenQA::Shared::Controller::OpenAPI;
+*catchall = \&OpenQA::Shared::Controller::OpenAPI::catchall;
```
Martchus, line 11: Maybe we can just inherit from `OpenQA::Shared::Controller::OpenAPI`.

lib/OpenQA/Shared/Controller/OpenAPI.pm
```diff
@@ -0,0 +1,23 @@
...
+    unless ($result->valid) {
+        return $c->render(status => 400, json => {result => $result});
+    }
+    my $VALIDATE = $c->stash('openapi');
+    my $path_captures = $VALIDATE->{path_captures};
+    $c->stash($_, $path_captures->{$_}) for keys %$path_captures;
+    my $op = $VALIDATE->{operation_id};
+    if ($op =~ s/^operation_(jobs|groups)_//) {
+        return $c->$op();
+    }
```
Martchus, line 19: I'm wondering whether Mojolicious provides some kind of hook so we can still use its logic to call the right function instead of having to reimplement it. Then we could probably also avoid changes to the route registrations.

perlpunk: The point is that the openAPI document provides the logic for the right function. We don't need our route registrations anymore, as it would be duplicated logic.

lib/OpenQA/WebAPI.pm
perlpunk: This is just a proof of concept with three routes. We don't need detailed routes likr this anymore when we move all API routes to openAPI

lib/OpenQA/Shared/Controller/OpenAPI.pm
Martchus: It is just that this regex and dynamic function call feel kind of like we're adding additional logic here as well.

lib/OpenQA/WebAPI.pm
Martchus: Ok, it is just that we use these names with the `url_for` helper all over the place. I'm wondering how this would look like then.

lib/OpenQA/Shared/Controller/OpenAPI.pm
perlpunk: The only thing to figure out is how to delegate this to the right controller. so it won't get more complicated than that, and maybe even easier.
Not sure yet how it would look like with the other plugin.

lib/OpenQA/WebAPI/Controller/API/V1/JobGroup.pm
perlpunk: I tried, but somehow it didn't work, but I don't know the error message any more

lib/OpenQA/WebAPI.pm
perlpunk: It could be that operationId is what's used for the route name. I would have to check.

PR #6633, opened 2025-08-04
Add reference to another lua example

aylen384:


aylen384:
It is - it shows how to call functions like `check_screen()` (https://open.qa/api/testapi/#_check_screen) that are using keyword arguments from lua.

docs/WritingTests.asciidoc
```diff
@@ -430,6 +430,8 @@ function test_flags(self)
 end
 --------------------------------------------------------------
 
+Another Lua example can be found https://github.com/os-autoinst/os-autoinst/blob/master/t/data/tests/luatests/pre_boot_lua.lua[here].
```
kalikiana: Maybe it would be nicer to add an example inline and add a couple sentences explaining it? Rather than linking to a file used for unit tests

aylen384: Done

docs/WritingTests.asciidoc
```diff
@@ -260,6 +260,72 @@ Additionally, Python tests do not support `run_args` and treat the presence of
...
+
+NOTE: Lua libraries are - other than perl libraries - imported into their own scope.
+
+[source,lua]
+-------------------------------------------------------------------
+require 'luatestlib'
+luatestlib.testfunc3() -- returns 44
+-------------------------------------------------------------------
+
+Because of the way `Inline::Lua` binds Perl functions to Python it is not
```
d3flex: Did you mean `Inline::Python`, instead of Lua here?

aylen384: no - but I meant Lua instead of Python

d3flex (approved):

PR #6634, opened 2025-08-04
Link minion job from openQA job with waiting task

Wabri:
related to: https://progress.opensuse.org/issues/169144

### DIFF
--- a/lib/OpenQA/Shared/Plugin/Gru.pm
+++ b/lib/OpenQA/Shared/Plugin/Gru.pm
@@ -211,7 +211,7 @@ sub enqueue ($self, $task, $args = [], $options = {}, $jobs = []) {
         @ttl,
         priority => $priority,
         delay => $delay,
-        notes => {gru_id => $gru_id, @notes},
+        notes => {gru_id => $gru_id, "gru_id_$gru_id" => 1, @notes},
         defined $lax ? (lax => $lax) : (),
         defined $parents ? (parents => $parents) : (),
     );
--- a/lib/OpenQA/WebAPI/Controller/Test.pm
+++ b/lib/OpenQA/WebAPI/Controller/Test.pm
@@ -522,7 +522,7 @@ sub _stash_clone_info ($self, $job) {
         });
 }
 
-sub gru_tasks_items ($self, $job) {
+sub _gru_tasks_items ($self, $job) {
     return [
         map {
             my $task = $_->gru_task;
@@ -538,6 +538,7 @@ sub gru_tasks_items ($self, $job) {
 
 sub infopanel ($self) {
     my $job = $self->_stash_job or return $self->reply->not_found;
+    $self->stash(gru_dependencies => $self->_gru_tasks_items($job));
     $self->stash({worker => $job->assigned_worker, additional_data => 1});
     $self->_stash_clone_info($job);
     $self->render('test/infopanel');
@@ -571,6 +572,7 @@ sub _show ($self, $job = undef) {
             show_investigation => $job->should_show_investigation,
             show_live_tab => $job->state ne DONE,
         });
+    $self->stash(gru_dependencies => $self->_gru_tasks_items($job));
     $self->_stash_clone_info($job);
     $self->render('test/result');
 }
--- a/lib/OpenQA/WebAPI/Controller/Test.pm
+++ b/lib/OpenQA/WebAPI/Controller/Test.pm
@@ -522,6 +522,20 @@ sub _stash_clone_info ($self, $job) {
         });
 }
 
+sub gru_tasks_items ($self, $job) {
+    return [
+        map {
+            my $task = $_->gru_task;
+            my $label = 'id: ' . $task->id . ', name: ' . $task->taskname;
+            my $href = '';
+            if ($self->is_operator) {
+                $href = $self->url_for('/minion/jobs')->query(note => 'gru_id_' . $task->id, task => $task->taskname);
+            }
+            {label => $label, href => $href}
+        } $job->gru_dependencies
+    ];
+}
+
 sub infopanel ($self) {
     my $job = $self->_stash_job or return $self->reply->not_found;
     $self->stash({worker => $job->assigned_worker, additional_data => 1});
--- a/t/ui/18-tests-details.t
+++ b/t/ui/18-tests-details.t
@@ -563,6 +563,49 @@ subtest ansiToHtml => sub {
 
 my $t = Test::Mojo->new('OpenQA::WebAPI');
 
+subtest 'GRU dependency shown in infopanel' => sub {
+    my $job = $jobs->find(99927);
+    $job->update({state => SCHEDULED, result => NONE});
+
+    my $task = $schema->resultset('GruTasks')->create(
+        {
+            taskname => 'git_clone',
+            args => {},
+            run_at => DateTime->now(),
+            priority => 0,
+        });
+    my $gru_dep = $schema->resultset('GruDependencies')->create(
+        {
+            job_id => $job->id,
+            gru_task_id => $task->id,
+        });
+
+    $driver->get('/tests/' . $job->id);
+    wait_for_ajax msg => 'job with GRU dependency loaded';
+
+    my $gru_task_id = $gru_dep->gru_task_id;
+
+    subtest 'operator sees link and it works' => sub {
+        ok my $link = $driver->find_element_by_link_text("id: $gru_task_id, name: git_clone"), 'found link';
+        like $link->get_attribute('href'), qr{minion/jobs\?note=gru_id_$gru_task_id&task=git_clone},
+          'link to minion job shown';
+    };
+
+    subtest 'guest user sees no link' => sub {
+        $driver->get('/logout');
+        wait_for_ajax msg => 'job with GRU dependency loaded';
+        $driver->get('/tests/' . $job->id);
+        wait_for_ajax msg => 'job with GRU dependency loaded';
+
+        my $txt = $driver->find_element('#info_box')->get_text;
+        like $txt, qr{State: .*, waiting for background tasks \(id: $gru_task_id, name: git_clone\)},
+          'The infobox is properly formatted';
+
+        my $links = $driver->find_elements('#info_box a[href*="/minion/jobs"]');
+        is(scalar(@{$links}), 0, 'Expect 0 link to minion jobs as guest');
+    };
+};
+
 subtest 'scheduled job' => sub {
     $t->get_ok('/tests/99927/infopanel_ajax')->status_is(200);
     $t->content_like(qr/scheduled.*, created.*\d\d\d\d-\d\d-\d\d \d\d:\d\d:\d\d/s, 'creation date displayed');
--- a/templates/webapi/test/infopanel.html.ep
+++ b/templates/webapi/test/infopanel.html.ep
@@ -44,16 +44,12 @@
                         Result: <b><%= $job->result %></b>\
                     % }
                     % else {
+                        % my $gru_tasks_items = $self->gru_tasks_items($job);
                         % my $state_detail = '';
-                        % if (my @gru_deps = $job->gru_dependencies) {
+                        % if (@$gru_tasks_items) {
                             % $state_detail = join(';', map {
-                            %   my $state = 'id: ' . $_->id . ', name: ' . $_->taskname;
-                            %   if (is_operator) {
-                            %      my $link = url_for('/minion/jobs')->query(note => 'gru_id_' . $_->id, task => $_->taskname);
-                            %      $state = link_to $state => $link
-                            %   }
-                            %   $state
-                            % } map { $_->gru_task // () } @gru_deps);
+                            %   $_->{href} ? link_to($_->{label} => $_->{href}) : $_->{label}
+                            % } @$gru_tasks_items);
                             % $state_detail = qq{, waiting for background tasks ($state_detail)};
                         % }
                         State: <b><%= $job->state %></b><%== $state_detail %>\
--- a/templates/webapi/test/infopanel.html.ep
+++ b/templates/webapi/test/infopanel.html.ep
@@ -47,11 +47,16 @@
                         % my $state_detail = '';
                         % if (my @gru_deps = $job->gru_dependencies) {
                             % $state_detail = join(';', map {
-                            %   'id: ' . $_->id . ', name: ' . $_->taskname
-                            %  } map { $_->gru_task // () } @gru_deps);
-                            % $state_detail = ", waiting for background tasks ($state_detail)";
+                            %   my $state = 'id: ' . $_->id . ', name: ' . $_->taskname;
+                            %   if (is_operator) {
+                            %      my $link = url_for('/minion/jobs')->query(note => 'gru_id_' . $_->id, task => $_->taskname);
+                            %      $state = link_to $state => $link
+                            %   }
+                            %   $state
+                            % } map { $_->gru_task // () } @gru_deps);
+                            % $state_detail = qq{, waiting for background tasks ($state_detail)};
                         % }
-                        State: <b><%= $job->state %></b><%= $state_detail %>\
+                        State: <b><%= $job->state %></b><%== $state_detail %>\
                     % }
                     % if ($job->t_finished) {
                         <%= ', finished' =%>
--- a/templates/webapi/test/infopanel.html.ep
+++ b/templates/webapi/test/infopanel.html.ep
@@ -44,12 +44,11 @@
                         Result: <b><%= $job->result %></b>\
                     % }
                     % else {
-                        % my $gru_tasks_items = $self->gru_tasks_items($job);
                         % my $state_detail = '';
-                        % if (@$gru_tasks_items) {
+                        % if (@$gru_dependencies) {
                             % $state_detail = join(';', map {
                             %   $_->{href} ? link_to($_->{label} => $_->{href}) : $_->{label}
-                            % } @$gru_tasks_items);
+                            % } @$gru_dependencies);
                             % $state_detail = qq{, waiting for background tasks ($state_detail)};
                         % }
                         State: <b><%= $job->state %></b><%== $state_detail %>\


lib/OpenQA/Schema/Result/Jobs.pm
```diff
@@ -2408,4 +2408,8 @@ sub video_file_paths ($self) {
     return find_video_files($testresdir);
 }
 
+sub minion_jobs ($self) {
+    return map {$_->gru_task->minion_job } $self->gru_dependencies->all; # don't remember why
```
perlpunk: There is no `minion_job` method for GruTasks...

templates/webapi/test/infopanel.html.ep
```diff
@@ -50,6 +50,13 @@
                             %   'id: ' . $_->id . ', name: ' . $_->taskname
                             %  } map { $_->gru_task // () } @gru_deps);
                             % $state_detail = ", waiting for background tasks ($state_detail)";
+                            % my $minion_job = $gru_deps[0]->gru_task->minion_job;
+                            % if ($minion_job) {
+                                % my $link = url_for('/minion/jobs')->query(id => "$job_id");
```
perlpunk: The suggestion in the ticket is to just link a query for searching for the gru id, not to calculate the minion job id on the server side every time a job is viewed.

lib/OpenQA/Shared/Plugin/Gru.pm
```diff
@@ -212,6 +215,7 @@ sub enqueue ($self, $task, $args = [], $options = {}, $jobs = []) {
         priority => $priority,
         delay => $delay,
         notes => {gru_id => $gru_id, @notes},
+        #notes => {gru_id => $gru_id, "gru_id_$gru_id" => 1, @notes},
```
perlpunk: This looks like the right approach. With this you should be able to use the link suggested in the ticket, e.g. `https://server/minion/jobs?task=...&note=gru_id_123`

lib/OpenQA/Schema/Result/Jobs.pm
Wabri: Yes, I think that was an implementation idea that I never finished to do

templates/webapi/test/infopanel.html.ep
Wabri: Yes this is the fourth attempts of the resolution :) I guess i go back to the `gru_id_123` attempt :+1:

lib/OpenQA/Shared/Plugin/Gru.pm
Wabri: Why note and not notes? I think there is something I didn't understand or maybe missing in the minion/jobs querying since even with `note=gru_id_$minion_id` (or with `s/note/notes`) the query return the list of all the jobs and not the one requested within the notes.

perlpunk: Have a look at
https://openqa.opensuse.org/minion/jobs?task=git_clone&state=&queue=&note=gru_id
That is how you search for minion jobs with a key `gru_id` in notes.

Wabri: Ok in order to let it works I should have reload the webui daemon :grimacing:

templates/webapi/test/infopanel.html.ep
```diff
@@ -46,12 +46,14 @@
                     % else {
                         % my $state_detail = '';
                         % if (my @gru_deps = $job->gru_dependencies) {
+                            % my $minion_id = $gru_deps[0] ? $gru_deps[0]->gru_task_id : "";
+                            % my $link = $gru_deps[0] ? url_for('/minion/jobs?note=gru_id_' . $minion_id) : "";
                             % $state_detail = join(';', map {
                             %   'id: ' . $_->id . ', name: ' . $_->taskname
                             %  } map { $_->gru_task // () } @gru_deps);
-                            % $state_detail = ", waiting for background tasks ($state_detail)";
+                            % $state_detail = ', waiting for background tasks (<a href="' . $link . '">' . $state_detail . '</a>)';
```
perlpunk:
```suggestion
                            % $state_detail = qq{, waiting for background tasks (<a href="$link">$state_detail</a>)};
```

templates/webapi/test/infopanel.html.ep
```diff
@@ -46,12 +46,14 @@
                     % else {
                         % my $state_detail = '';
                         % if (my @gru_deps = $job->gru_dependencies) {
+                            % my $minion_id = $gru_deps[0] ? $gru_deps[0]->gru_task_id : "";
```
perlpunk: There could be more than one gru_task theoretically. Why do you only add a link for the first?

templates/webapi/test/infopanel.html.ep
```diff
@@ -46,12 +46,14 @@
...
+                            % my $link = $gru_deps[0] ? url_for('/minion/jobs?note=gru_id_' . $minion_id) : "";
```
perlpunk: Please use the `query` method to add query parameters:
```suggestion
                            % my $link = $gru_deps[0] ? url_for('/minion/jobs')->query(note => "gru_id_$minion_id") : "";
```
But I would add the link for every gru task and also add the task name in the query.

templates/webapi/test/infopanel.html.ep
```diff
@@ -47,11 +47,12 @@
                         % my $state_detail = '';
                         % if (my @gru_deps = $job->gru_dependencies) {
                             % $state_detail = join(';', map {
-                            %   'id: ' . $_->id . ', name: ' . $_->taskname
+                            %   my $link = url_for('/minion/jobs')->query(note => 'gru_id_' . $_->id);
```
perlpunk:
```suggestion
                            %   my $link = url_for('/minion/jobs')->query(note => 'gru_id_' . $_->id, task => $_->taskname);
```

Wabri: Could it be more minion jobs with the same gru_id?

perlpunk: no

Wabri: ok, so why querying also the taskname?

perlpunk: I don't know how Minion is indexing its database, so I thought it could be useful.
You can leave it

Wabri: I've added it, mine was just a question to understand if I misunderstand something :)

templates/webapi/test/infopanel.html.ep
```diff
@@ -47,11 +47,12 @@
                         % my $state_detail = '';
                         % if (my @gru_deps = $job->gru_dependencies) {
                             % $state_detail = join(';', map {
-                            %   'id: ' . $_->id . ', name: ' . $_->taskname
+                            %   my $link = url_for('/minion/jobs')->query(note => 'gru_id_' . $_->id, task => $_->taskname);
```
perlpunk: Now I'm wondering - not everyone can access the minion dashboard. Maybe the link could only be printed `if (is_operator)`. Presenting a link that one cannot access is maybe not so nice.

Wabri: Could it be better to put it before doing the mapping? maybe in the if statement above to prevent useless computation? or maybe we want to explicitly tell that there is a minion job working in background? looking at the UX prospective I think we should present that there are minions working without linking to them. I'm going to add the is_operator flag inside the map to prevent the link creation if not an operator.

templates/webapi/test/infopanel.html.ep
```diff
@@ -47,11 +47,19 @@
...
-                            %   'id: ' . $_->id . ', name: ' . $_->taskname
+                            %   my $prefix = '';
+                            %   my $state = 'id: ' . $_->id . ', name: ' . $_->taskname;
+                            %   my $suffix = '';
+                            %   if (is_operator) {
+                            %      my $link = url_for('/minion/jobs')->query(note => 'gru_id_' . $_->id, task => $_->taskname);
+                            %      $prefix = '<a href="' . $link . '">';
+                            %      $suffix = '</a>';
+                            %   }
+                            %   $prefix . $state . $suffix
```
perlpunk:
```suggestion
                            %   my $state = 'id: ' . $_->id . ', name: ' . $_->taskname;
                            %   if (is_operator) {
                            %      my $link = url_for('/minion/jobs')->query(note => 'gru_id_' . $_->id, task => $_->taskname);
                            %      $state = qq{<a href="$link">$state</a>};
                            %   }
                            %   $state
```

Wabri: Ok you didn't like my verbose string formatting 🤣 I'm going to fix it!

kalikiana (approved):

t/ui/18-tests-details.t
```diff
@@ -563,6 +563,40 @@ subtest ansiToHtml => sub {
 
 my $t = Test::Mojo->new('OpenQA::WebAPI');
 
+subtest 'GRU dependency shown in infopanel' => sub {
```
perlpunk, line 566: Ok, but when I run this test on current master, it also succeeds :)

So maybe you can additionally check for the link with the correct gru_task id?

t/ui/18-tests-details.t
```diff
@@ -563,6 +563,41 @@ subtest ansiToHtml => sub {
...
+    $test_case->login($t, 'arthur');
```
Wabri: TODO: find why the login is not working here

t/ui/18-tests-details.t
```diff
@@ -563,6 +563,41 @@ subtest ansiToHtml => sub {
...
+
+    is(current_tab, 'Settings', 'Waiting for gru task start from Details tab');
+
+    like(
+        $driver->find_element('#info_box')->get_text(),
+        qr{State: scheduled, waiting for background tasks (<a href},
+        'Waiting shown'
+    );
+    $t->get_ok('/tests/' . $job->id . '/infopanel_ajax')->status_is(200)
+      ->content_like(qr/waiting for background tasks *href/s, 'waiting text present');
```
Wabri: TODO: improve the regex

t/ui/18-tests-details.t
```diff
@@ -563,6 +563,41 @@ subtest ansiToHtml => sub {
...
+
+    $driver->get('/tests/' . $job->id);
+    disable_bootstrap_animations;
+    wait_for_ajax msg => 'job with GRU dependency loaded';
+
+    is(current_tab, 'Settings', 'Waiting for gru task start from Details tab');
+
+    like(
+        $driver->find_element('#info_box')->get_text(),
+        qr{State: scheduled, waiting for background tasks (<a href},
```
perlpunk:
```suggestion
        qr{State: scheduled, waiting for background tasks \(<a href},
```

t/ui/18-tests-details.t
perlpunk: you are already logged in.
in NON_HEADLESS mode I can see the link.

Wabri: As you mention in the daily the problem was related to the get text, I've replaced it with innerHTML :)

t/ui/18-tests-details.t
```diff
@@ -563,6 +563,40 @@ subtest ansiToHtml => sub {
...
+
+    my $gru_task_id = $gru_dep->gru_task_id;
+    my $innerHTML = $driver->find_element('#info_box')->get_attribute('innerHTML');
+    my $state_regex = qr{State: <b>.*</b>, waiting for background tasks \(<a href=.*note=gru_id_*$gru_task_id};
+    like (
+        $innerHTML,
+        $state_regex,
+        'GRU tasks are linked properly'
+    );
+};
```
perlpunk:
```suggestion
    like $driver->find_element('#info_box')->get_text(), qr{State: scheduled, waiting for background tasks},
      'waiting shown';
    ok my $link = $driver->find_element_by_link_text('id: 1, name: git_clone'), 'found link';
    like $link->get_attribute('href'), qr{minion/jobs\?note=gru_id_1&task=git_clone}, 'link to minion job shown';
```

Wabri: Sorry I've already change the test :D I didn't notice this!

t/ui/18-tests-details.t
```diff
@@ -563,6 +563,48 @@ subtest ansiToHtml => sub {
...
+        {
+            job_id => $job->id,
+            gru_task_id => $task->id,
+        });
+
+    $driver->get('/tests/' . $job->id);
+    disable_bootstrap_animations;
+    wait_for_ajax msg => 'job with GRU dependency loaded';
+
+    my $state_regex = qr{State: <b>.*</b>, waiting for background tasks \(<a href=.*note=gru_id_1};
```
perlpunk: I would avoid matching literal HTML if not necessary. If a small thing changes, you have to adapt all the tests.
Also, so far I can't find a usage of `innerHTML` in any of our tests.

And I would actually use the `$task->id` instead of the hardcoded 1, it was just a quick code suggestion where I used it.

t/ui/18-tests-details.t
```diff
@@ -563,6 +563,48 @@ subtest ansiToHtml => sub {
...
+        });
+
+    my $gru_dep = $schema->resultset('GruDependencies')->create(
+        {
+            job_id => $job->id,
+            gru_task_id => $task->id,
+        });
+
+    $driver->get('/tests/' . $job->id);
+    disable_bootstrap_animations;
```
Martchus: I'm not sure whether `disable_bootstrap_animations` is still required at all. So did you notice that it is required in this new piece of test code? If not (and you just copied it blindly from elsewhere) then I would leave it out.

t/ui/18-tests-details.t
```diff
@@ -563,6 +563,48 @@ subtest ansiToHtml => sub {
...
+    wait_for_ajax msg => 'job with GRU dependency loaded';
+
+    my $state_regex = qr{State: <b>.*</b>, waiting for background tasks \(<a href=.*note=gru_id_1};
+
+    subtest 'operator sees link and it works' => sub {
+        my $innerHTML = $driver->find_element('#info_box')->get_attribute('innerHTML');
```
Martchus: Just using `get_text()` doesn't work?
```suggestion
        my $innerHTML = $driver->find_element('#info_box')->get_text();
```

t/ui/18-tests-details.t
```diff
@@ -563,6 +563,48 @@ subtest ansiToHtml => sub {
...
+        like($innerHTML, $state_regex, 'GRU tasks are linked properly');
+
+        $driver->find_element_by_link_text('id: 1, name: git_clone')->click;
+        like($driver->get_current_url(), qr{/minion/jobs\?note=gru_id_1}, 'Minion page');
+        $driver->go_back();
```
Martchus: No need to go back considering the next subtest starts on a fresh page anyway. So this just slows down the test execution.

t/ui/18-tests-details.t
```diff
@@ -563,6 +563,48 @@ subtest ansiToHtml => sub {
...
+    };
+
+    subtest 'guest user sees no link' => sub {
+        $driver->get('/logout');
+        $driver->get('/tests/' . $job->id);
+        my $innerHTML = $driver->find_element('#info_box')->get_attribute('innerHTML');
+        unlike($innerHTML, $state_regex, 'GRU tasks are not linked when no auths');
+        $driver->get('/login');
```
Martchus: Maybe this should be moved to the next test that actually requires the user to be logged in.

templates/webapi/test/infopanel.html.ep
```diff
@@ -47,11 +47,16 @@
...
-                            %  } map { $_->gru_task // () } @gru_deps);
-                            % $state_detail = ", waiting for background tasks ($state_detail)";
+                            %   my $state = 'id: ' . $_->id . ', name: ' . $_->taskname;
+                            %   if (is_operator) {
+                            %      my $link = url_for('/minion/jobs')->query(note => 'gru_id_' . $_->id, task => $_->taskname);
+                            %      $state = qq{<a href="$link">$state</a>};
+                            %   }
+                            %   $state
+                            % } map { $_->gru_task // () } @gru_deps);
+                            % $state_detail = qq{, waiting for background tasks ($state_detail)};
```
Martchus: This should now probably be moved to the controller. (There you can still use helpers like `url_for` and `is_operator` via the controller object (usually `$self` or `$c`).)

templates/webapi/test/infopanel.html.ep
```diff
@@ -47,11 +47,16 @@
                         % my $state_detail = '';
                         % if (my @gru_deps = $job->gru_dependencies) {
                             % $state_detail = join(';', map {
-                            %   'id: ' . $_->id . ', name: ' . $_->taskname
-                            %  } map { $_->gru_task // () } @gru_deps);
-                            % $state_detail = ", waiting for background tasks ($state_detail)";
+                            %   my $state = 'id: ' . $_->id . ', name: ' . $_->taskname;
+                            %   if (is_operator) {
+                            %      my $link = url_for('/minion/jobs')->query(note => 'gru_id_' . $_->id, task => $_->taskname);
+                            %      $state = qq{<a href="$link">$state</a>};
```
Martchus: This code lacks proper escaping. You should probably use the `link_to` helper here.

t/ui/18-tests-details.t
```diff
@@ -563,6 +563,48 @@ subtest ansiToHtml => sub {
...
+    $job->update({state => SCHEDULED, result => NONE});
+
+    my $task = $schema->resultset('GruTasks')->create(
+        {
+            taskname => 'git_clone',
+            args => {},
+            run_at => DateTime->now(),
+            priority => 0,
+        });
+
```
Martchus:
```suggestion
```

t/ui/18-tests-details.t
Wabri: innerHTML is a property of an element in the dom, should I use stuff only that has already been used?

I changed from the variable to hardcoded because I saw that you didn't use in your proposal, I'll put it back once again!

t/ui/18-tests-details.t
Wabri: Yep, I just copy from other test and forgot there while editing. may I ask you what is that about? I don't find any reference to the declaration of it!

t/ui/18-tests-details.t
Wabri: Nope, because I need to check the `<a href=blabla> ... </a>` and `get_text` only take the text.

t/ui/18-tests-details.t
Wabri: It's a refuse of when all the gru tests were in the same test.

t/ui/18-tests-details.t
Wabri: Actually they didn't care about the login, I'm removing it!

t/ui/18-tests-details.t
perlpunk: > innerHTML is a property of an element in the dom, should I use stuff only that has already been used?

No, but that's not the point.
My point is, there is maybe a reason, why we don't use innerHTML in tests yet, and I stated my reason:
> I would avoid matching literal HTML if not necessary. If a small thing changes, you have to adapt all the tests.

t/ui/18-tests-details.t
```diff
@@ -563,6 +563,48 @@ subtest ansiToHtml => sub {
...
+    wait_for_ajax msg => 'job with GRU dependency loaded';
+
+    my $state_regex = qr{State: <b>.*</b>, waiting for background tasks \(<a href=.*note=gru_id_1};
+
+    subtest 'operator sees link and it works' => sub {
+        my $innerHTML = $driver->find_element('#info_box')->get_attribute('innerHTML');
+        like($innerHTML, $state_regex, 'GRU tasks are linked properly');
+
+        $driver->find_element_by_link_text('id: 1, name: git_clone')->click;
+        like($driver->get_current_url(), qr{/minion/jobs\?note=gru_id_1}, 'Minion page');
```
perlpunk:
```suggestion
    my $gru_task_id = $task->id;
    like $driver->find_element('#info_box')->get_text(), qr{State: scheduled, waiting for background tasks},
      'waiting shown';
    ok my $link = $driver->find_element_by_link_text("id: $gru_task_id, name: git_clone"), 'found link';
    like $link->get_attribute('href'), qr{minion/jobs\?note=gru_id_$gru_task_id&task=git_clone}, 'link to minion job shown';
```

Wabri: Why remove the subtest?

perlpunk: I'm just showing how to avoid matching literal HTML.
You can keep the subtest

t/ui/18-tests-details.t
Martchus: Then it would make sense to use a selector that returns that `a` element and get its href attribute. To make it easier you can add a class to the `a` element and then use e.g. `#info_box .state-link` as selector.

t/ui/18-tests-details.t
Martchus: And @perlpunk's suggestion is also in-line with my review comment where I also suggested alternatives.

t/ui/18-tests-details.t
Martchus: > I don't find any reference to the declaration of it!

Then try to `grep` harder and when you've found it use `git blame`. I don't remember exactly anymore why this is still necessary so I would have to check the Git history myself. Perhaps none of these function calls (and hence also the function itself) are required anymore but figuring this out is of course beyond the scope of this PR. (I'm just asking to avoid spreading the use of this function further unless you say it is actually necessary here.)

Martchus: (The function is in `t/lib/OpenQA/SeleniumTest.pm`.)

t/ui/18-tests-details.t
kalikiana: So why not `get_attribute('href')` then? 🤔

Wabri: Right now locally it's:

```
        ok(my $link = $driver->find_element_by_link_text("id: $gru_task_id, name: git_clone"), 'found link');
        like($link->get_attribute('href'), qr{minion/jobs\?note=gru_id_$gru_task_id&task=git_clone}, 'link to minion job shown');

        $driver->find_element_by_link_text("id: $gru_task_id, name: git_clone")->click;
        like($driver->get_current_url(), qr{/minion/jobs\?note=gru_id_$gru_task_id}, 'Minion page');
```

as suggested by @perlpunk here https://github.com/os-autoinst/openQA/pull/6634/files#r2270151137

t/ui/18-tests-details.t
Wabri: the only problem is for the guest user where the `find_element_by_link_text` would not work

templates/webapi/test/infopanel.html.ep
Wabri: do you mean the creation of the state detail? is it a job for the controller?

t/ui/18-tests-details.t
Wabri: commit done by you on 2018:
```
commit f1b52b7a4a9418d30e690d1ac98166aba6003803
Author: Marius Kittler <mkittler@suse.de>
Date:   Thu Aug 2 17:05:33 2018 +0200

    Unify disable_bootstrap_animations and disable_bootstrap_fade_animation

diff --git a/t/lib/OpenQA/SeleniumTest.pm b/t/lib/OpenQA/SeleniumTest.pm
index 1c50a1d17..7f04468c1 100644
--- a/t/lib/OpenQA/SeleniumTest.pm
+++ b/t/lib/OpenQA/SeleniumTest.pm
@@ -9,8 +9,9 @@ require Mojolicious::Commands;
 
 our @EXPORT = qw($drivermissing check_driver_modules start_driver
-  call_driver kill_driver wait_for_ajax open_new_tab mock_js_functions
-  element_visible element_hidden javascript_console_has_no_warnings_or_errors);
+  call_driver kill_driver wait_for_ajax disable_bootstrap_animations
+  open_new_tab mock_js_functions element_visible element_hidden
+  javascript_console_has_no_warnings_or_errors);
 
 use Data::Dump 'pp';
 use Test::More;
@@ -177,6 +178,16 @@ sub wait_for_ajax {
     }
 }
 
+sub disable_bootstrap_animations {
+    my @rules = (
+        "'.fade', '-webkit-transition: none !important; transition: none !important;'",
+        "'.collapsing', '-webkit-transition: none !important; transition: none !important;'",
+    );
+    for my $rule (@rules) {
+        $_driver->execute_script("document.styleSheets[0].addRule($rule, 1);");
+    }
+}
+
 sub javascript_console_has_no_warnings_or_errors {
     my ($test_name_suffix) = @_;
     $test_name_suffix //= '';
diff --git a/t/ui/14-dashboard-parents.t b/t/ui/14-dashboard-parents.t
index 3f5f539b5..0da4cf879 100644
--- a/t/ui/14-dashboard-parents.t
+++ b/t/ui/14-dashboard-parents.t
@@ -1,6 +1,6 @@
 #! /usr/bin/perl
 
-# Copyright (C) 2016-2017 SUSE LLC
+# Copyright (C) 2016-2018 SUSE LLC
 #
 # This program is free software; you can redistribute it and/or modify
 # it under the terms of the GNU General Public License as published by
@@ -97,12 +97,6 @@ unless (can_load(modules => {'Selenium::Remote::WDKeys' => undef,})) {
 
 # don't tests basics here again, this is already done in t/22-dashboard.t and t/ui/14-dashboard.t
 
-sub disable_bootstrap_animations {
-    $driver->execute_script(
-"document.styleSheets[0].addRule('.collapsing', '-webkit-transition: none !important; transition: none !important;', 1);"
-    );
-}
-
 $driver->title_is('openQA', 'on main page');
 my $baseurl = $driver->get_current_url();
 
diff --git a/t/ui/18-tests-details.t b/t/ui/18-tests-details.t
index a14581cd9..9ba9a1c03 100644
--- a/t/ui/18-tests-details.t
+++ b/t/ui/18-tests-details.t
@@ -1,6 +1,6 @@
 #! /usr/bin/perl
 
-# Copyright (C) 2014-2017 SUSE LLC
+# Copyright (C) 2014-2018 SUSE LLC
 #
 # This program is free software; you can redistribute it and/or modify
 # it under the terms of the GNU General Public License as published by
@@ -55,12 +55,6 @@ unless ($driver) {
 }
 my $baseurl = $driver->get_current_url;
 
-sub disable_bootstrap_fade_animation {
-    $driver->execute_script(
-"document.styleSheets[0].addRule('.fade', '-webkit-transition: none !important; transition: none !important;', 1);"
-    );
-}
-
 # returns the contents of the candidates combo box as hash (key: tag, value: array of needle names)
 sub find_candidate_needles {
     # ensure the candidates menu is visible
@@ -120,7 +114,7 @@ like(
 is($driver->find_element('.cm-comment')->get_text(), '#!/usr/bin/perl -w', "we have a perl comment");
 
 $driver->get("/tests/99937");
-disable_bootstrap_fade_animation;
+disable_bootstrap_animations;
 sub current_tab {
     return $driver->find_element('.nav.nav-tabs .active')->get_text;
 }
@@ -324,7 +318,7 @@ sub test_with_error {
     # check whether candidates are displayed as expected
     my $random_number = int(rand(100000));
     $driver->get("/tests/99946?prevent_caching=$random_number#step/yast2_lan/1");
-    disable_bootstrap_fade_animation;
+    disable_bootstrap_animations;
     wait_for_ajax;
     is_deeply(find_candidate_needles, $expect, $test_name // 'candidates displayed as expected');
 }
```

Thanks for the explanation! I removed from the test!

t/ui/18-tests-details.t
```diff
@@ -563,6 +563,45 @@ subtest ansiToHtml => sub {
...
+        like($driver->get_current_url(), qr{/minion/jobs\?note=gru_id_$gru_task_id}, 'Minion page');
+    };
+
+    subtest 'guest user sees no link' => sub {
+        $driver->get('/logout');
+        $driver->get('/tests/' . $job->id);
+        my $state_regex = qr{State: <b>.*</b>, waiting for background tasks \(<a href=.*note=gru_id_$gru_task_id};
+        my $innerHTML = $driver->find_element('#info_box')->get_attribute('innerHTML');
+        unlike($innerHTML, $state_regex, 'GRU tasks are not linked when no auths');
+    };
```
Wabri: TODO: refactor

t/ui/18-tests-details.t
```diff
@@ -563,6 +563,45 @@ subtest ansiToHtml => sub {
...
+    wait_for_ajax msg => 'job with GRU dependency loaded';
+
+    my $gru_task_id = $gru_dep->gru_task_id;
+
+    subtest 'operator sees link and it works' => sub {
+        ok my $link = $driver->find_element_by_link_text("id: $gru_task_id, name: git_clone"), 'found link';
+        like $link->get_attribute('href'), qr{minion/jobs\?note=gru_id_$gru_task_id&task=git_clone}, 'link to minion job shown';
+
+        $driver->find_element_by_link_text("id: $gru_task_id, name: git_clone")->click;
+        like($driver->get_current_url(), qr{/minion/jobs\?note=gru_id_$gru_task_id}, 'Minion page');
```
Martchus: Considering we don't do anything on the Minion page I suggest you remove the click and subsequent check. It is probably not worth checking that the link itself works as this functionality is basically provided by the browser. 

Without the clicking it would make sense to make the check on the href attribute more strict, though - e.g. use anchoring in the regex to make sure the path starts as expected.

t/ui/18-tests-details.t
Wabri: Solved using the `find_elements`

t/ui/18-tests-details.t
Wabri: done

t/ui/18-tests-details.t
Wabri: :+1:

t/ui/18-tests-details.t
```diff
@@ -563,6 +563,50 @@ subtest ansiToHtml => sub {
...
+    wait_for_ajax msg => 'job with GRU dependency loaded';
+
+    my $gru_task_id = $gru_dep->gru_task_id;
+
+    subtest 'operator sees link and it works' => sub {
+        ok my $link = $driver->find_element_by_link_text("id: $gru_task_id, name: git_clone"), 'found link';
+        like $link->get_attribute('href'), qr{minion/jobs\?note=gru_id_$gru_task_id&task=git_clone}, 'link to minion job shown';
+
+        my $links = $driver->find_elements('#info_box a[href*="/minion/jobs"]');
+        is(scalar(@{$links}), 1, 'Expect 1 link to minion jobs as operator');
```
Wabri: @Martchus can something like this work instead of click the link itself?

templates/webapi/test/infopanel.html.ep
perlpunk: I guess I would create an array reference in the controller with links and labels (for each gru task) and save it in the stash, so that you only have to iterate over the array here

t/ui/18-tests-details.t
Martchus: This seems redundant. Doesn't it do the same as the check "link to minion job shown" which would only need anchoring in the regex?

templates/webapi/test/infopanel.html.ep
```diff
@@ -44,14 +44,13 @@
...
-                            %   'id: ' . $_->id . ', name: ' . $_->taskname
-                            %  } map { $_->gru_task // () } @gru_deps);
-                            % $state_detail = ", waiting for background tasks ($state_detail)";
-                        % }
-                        State: <b><%= $job->state %></b><%= $state_detail %>\
+                        % $state_detail = join(';', map {
+                        %   $_->{href} ? link_to($_->{label} => $_->{href}) : $_->{label}
+                        % } @$gru_tasks_items);
+                        % $state_detail = qq{, waiting for background tasks ($state_detail)};
+                        State: <b><%= $job->state %></b><%== $state_detail %>\
```
Wabri: TODO: I'm not sure about this without any checks to the $gru_tasks_items, should I check for not empty and if it is skip the detail creation?

lib/OpenQA/WebAPI/Controller/Test.pm
```diff
@@ -522,6 +522,21 @@ sub _stash_clone_info ($self, $job) {
...
+        my $label = 'id: ' . $task->id . ', name: ' . $task->taskname;
+        my $href = '';
+        if($self->is_operator){
+            $href = $self->url_for('/minion/jobs')
+                ->query(note=>'gru_id_'.$task->id,task=>$task->taskname);
+        }
+        push @items, { label=>$label, href=>$href};
+    }
+    return \@items
+}
```
Wabri: TODO: find the place to put this took me a while, is it the right place? there are some refactor to do 100%

templates/webapi/test/infopanel.html.ep
Wabri: I try to use the `stash` I failed, so I went for a more easy solution (for me at least)

perlpunk: > I try to use the `stash` I failed, so I went for a more easy solution (for me at least)

```
$self->stash(gru_dependencies => $self->gru_tasks_items);
```
and then use `$gru_dependencies` in the template.

lib/OpenQA/WebAPI/Controller/Test.pm
```diff
@@ -522,6 +522,21 @@ sub _stash_clone_info ($self, $job) {
         });
 }
 
+sub gru_tasks_items($self, $job){
```
perlpunk: If it's not supposed to be called from anywhere else, name it `_gru_tasks_items`.

templates/webapi/test/infopanel.html.ep
perlpunk: I see "State: scheduled, waiting for background tasks (), created 3 minutes ago" for a scheduled job that doesn't have a gru task, so yes, check it

lib/OpenQA/WebAPI/Controller/Test.pm
```diff
@@ -522,6 +522,21 @@ sub _stash_clone_info ($self, $job) {
...
+        my $task = $dep->gru_task;
+        my $label = 'id: ' . $task->id . ', name: ' . $task->taskname;
+        my $href = '';
+        if($self->is_operator){
+            $href = $self->url_for('/minion/jobs')
+                ->query(note=>'gru_id_'.$task->id,task=>$task->taskname);
+        }
+        push @items, { label=>$label, href=>$href};
+    }
+    return \@items
```
okurz: This might work

```suggestion
    return [map {
        my $task = $_->gru_task;
        my $label = 'id: ' . $task->id . ', name: ' . $task->taskname;
        my $href = '';
        if($self->is_operator) {
            $href = $self->url_for('/minion/jobs')
                ->query(note=>'gru_id_'.$task->id,task=>$task->taskname);
        }
        { label=>$label, href=>$href}
     } $job->gru_dependencies];
```

or using another function to create one entity


```suggestion
sub gru_task_entry ($self, $dep) {
        my $task = $dep->gru_task;
        my $label = 'id: ' . $task->id . ', name: ' . $task->taskname;
        my $href = '';
        if($self->is_operator) {
            $href = $self->url_for('/minion/jobs')
                ->query(note=>'gru_id_'.$task->id,task=>$task->taskname);
        }
        return { label=>$label, href=>$href};
}
…

    return [map { $self->gru_task_entry($_) } $job->gru_dependencies];
```

lib/OpenQA/WebAPI/Controller/Test.pm
Wabri: It's used in the infopanel!

templates/webapi/test/infopanel.html.ep
```diff
@@ -44,14 +44,15 @@
                         Result: <b><%= $job->result %></b>\
                     % }
                     % else {
+                        % my $gru_tasks_items = $self->gru_tasks_items($job);
```
perlpunk: I would still vote for using the stash.

Martchus: I wasn't even aware (or don't remember) that it is possible to refer to the controller via `$self` in templates. The stash would indeed be the more conventional way. (Or adding a helper function but here the stash makes more sense.)

templates/webapi/test/infopanel.html.ep
Wabri: The problem was not in the controller itself, but in the page that seems that gru_dependencies cannot be used as for job variable.

templates/webapi/test/infopanel.html.ep
perlpunk: Yeah. I think it's a good idea in general to let the controller do the logic work, and have all data ready for the template to consume.
Helpers are exceptions that are sometimes needed, but I don't think we need one here, as this is used only in one place right now.

templates/webapi/test/infopanel.html.ep
Wabri: Fixed, commit incoming!

lib/OpenQA/WebAPI/Controller/Test.pm
```diff
@@ -522,8 +522,27 @@ sub _stash_clone_info ($self, $job) {
...
+    ];
+}
+
+sub _stash_gru_dependencies ($self, $job) {
+    $self->stash(gru_dependencies => $self->_gru_tasks_items($job));
+}
+
 sub infopanel ($self) {
     my $job = $self->_stash_job or return $self->reply->not_found;
+    $self->_stash_gru_dependencies($job);
```
perlpunk:
```suggestion
    $self->stash(gru_dependencies => $self->_gru_tasks_items($job));
```

Wabri: Yep this is better, I was reproducing the things made for $job and it started to worked only when I've add the stash into the `_show` subroutine, I guess is used to setup the stash for the template.

Wabri: I've also removed from the infopanel subroutine and it still works as long as it is in `_show`.

@Martchus does it make sense?

Wabri: (In the meantime, I have updated the pr)

perlpunk: I'm wondering btw, where do we even use `infopanel_ajax`?
```
<div id="info_box" data-src="<%= url_for('test_infopanel', testid => $testid) %>">
%= include 'test/infopanel', worker => $assigned_worker, additional_data => 0
```
So the div has a data-src, but we include the infopanel template as well, and `infopanel_ajax` is never actually accessed?

Wabri: Maybe is a regressions of some sort

perlpunk: Oh I see, when the job status changes, `infopanel_ajax` is retrieved...

Wabri: Yes, removing it from the `infopanel` subroutine break the test `scheduled job`.

templates/webapi/test/infopanel.html.ep
```diff
@@ -44,14 +44,15 @@
                         Result: <b><%= $job->result %></b>\
                     % }
                     % else {
+                        % my $gru_tasks_items = $gru_dependencies;
```
Martchus: Looks like `$gru_tasks_items` is no longer necessary as you can use `$gru_dependencies` directly.

Wabri: You are right... Updated!

Martchus (approved):

PR #6623, opened 2025-07-29
Obsolete script/client after 4 year deprecation period

okurz:
script/client was declared deprecated in 755a7a256 2021-07-14. It is
time to obsolete this script but point to openqa-cli as replacement.

Related progress issue: https://progress.opensuse.org/issues/173881

PR #6637, opened 2025-08-05
Feat/173881 okurz fork

Wabri:
Superseded of #6623  

Related progress issue: https://progress.opensuse.org/issues/173881


perlpunk:
There is another test: `t/31-client.t`

script/client
```diff
@@ -1,188 +1,15 @@
...
-if ($options{form} && $options{'json-data'}) {
-    print STDERR "ERROR: The options --form and --json-data can not be combined.\n";
-    exit(2);
-}
-
-my $ret = run(\%options, @ARGV);
-exit(defined $ret ? 0 : 1);
-
-1;
+echo "NOTICE: openqa-client is obsolete. Use openqa-cli instead"
```
perlpunk:
```suggestion
echo "NOTICE: openqa-client is obsolete. Use openqa-cli instead" >&2
```
And maybe you an just add a line to satisfy the 44-scripts test, e.g.
```bash
if [[ $1 == --help ]]; then exit; fi
```

dist/rpm/openQA.spec
```diff
@@ -391,7 +391,6 @@ rm -rf %{buildroot}/%{_prefix}/lib/sysctl.d/01-openqa-reload-worker-auto-restart
 %endif
 
 mkdir -p %{buildroot}%{_bindir}
-ln -s %{_datadir}/openqa/script/client %{buildroot}%{_bindir}/openqa-client
```
perlpunk: I think it should still be part of the package. Otherwise we could have just deleted it instead of letting it print an error message.

Wabri: I was thinking about where `openqa-client` was declared on the specfile :rofl: thanks, you save me 10min at least :grin:

perlpunk (approved):

kalikiana (approved):

t/31-client.t
```diff
@@ -1,96 +0,0 @@
...
-use Test::Warnings ':report_warnings';
-use OpenQA::WebAPI;
-use OpenQA::Test::Case;
-use OpenQA::Script::Client;
-use OpenQA::Scheduler::Client;
-use OpenQA::WebSockets::Client;
-use Mojo::File qw(tempdir);
-
-subtest 'hostnames configurable' => sub {
-    my $config_dir = tempdir;
```
perlpunk, line 23: We discussed this morning that these tests shouldn't actually be deleted...
Can you recreate this (without the OpenQA::Script::Client tests)?

Wabri: Yes, I can. I'm going to prepare a new pr for this!

PR #6638, opened 2025-08-05
Use MCP to connect openQA with AI

kraih:
This patch adds the basic infrastructure for native MCP support in openQA. Most MCP clients use bearer token authentication at the moment, so i've added that to the existing token authentication mechanism that already existed in openQA. Future versions of the MCP specification will most likely require more sophisticated authentication mechanisms.

For clients that use the standard `mcp.json` format, a configuration should look something like this:
```
{
  "mcpServers": {
    "openqa": {
      "url": "http://127.0.0.1:9526/experimental/mcp",
      "headers": {
        "Authorization": "Bearer USER:KEY:SECRET"
      }
    }
  }
}
```

For starters i have only added a single tool called `current_user`, which an LLM can use to identify the openQA user.

Here is an example conversation with Gemma 3 where it makes a tool call to identify me:
```
You:
Who am i?

gemma-3-12b-it:
current_user({})
Arguments
{}
Result
[
  {
    "type": "text",
    "text": "name: Demo, id: 2"
  }
]

You are Demo.

You:
Do i have a user id too?

gemma-3-12b-it:
Yes, your user ID is 2.
```

There are many alternative ways in which MCP support could be added to openQA. But i believe this one will perform best.

Choosing what other tools to add will be the most important next step. And not an easy one, since the LLM context window size is a finite resource, we need to be picky about which tools are really necessary, and which information they will include in their responses.

### DIFF
--- a/lib/OpenQA/Shared/Controller/Auth.pm
+++ b/lib/OpenQA/Shared/Controller/Auth.pm
@@ -56,7 +56,10 @@ sub auth ($self) {
     else {
 
         # Personal access token
-        if (my $userinfo = $self->req->url->to_abs->userinfo) {
+        if (($self->req->headers->authorization // '') =~ /^Bearer\s+(.+)$/) {
+            ($user, $reason) = $self->_token_auth($reason, $1);
+        }
+        elsif (my $userinfo = $self->req->url->to_abs->userinfo) {
             ($user, $reason) = $self->_token_auth($reason, $userinfo);
         }
 
--- a/lib/OpenQA/WebAPI.pm
+++ b/lib/OpenQA/WebAPI.pm
@@ -11,6 +11,7 @@ use OpenQA::Log qw(setup_log log_debug log_error);
 use OpenQA::Utils qw(detect_current_version service_port);
 use OpenQA::Setup;
 use OpenQA::WebAPI::Description qw(get_pod_from_controllers set_api_desc);
+use OpenQA::WebAPI::MCP;
 use Mojo::File 'path';
 use Mojo::IOLoop;
 use Mojo::Util 'trim';
@@ -51,6 +52,9 @@ sub startup ($self) {
     # take care of DB deployment or migration before starting the main app
     OpenQA::Schema->singleton;
 
+    # MCP server
+    my $mcp = OpenQA::MCP->new;
+
     # Some controllers are shared between openQA micro services
     my $r = $self->routes->namespaces(['OpenQA::Shared::Controller', 'OpenQA::WebAPI::Controller', 'OpenQA::WebAPI']);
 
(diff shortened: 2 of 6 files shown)


kraih:
Tests won't pass due to the new dependency on the `MCP` module.

perlpunk:
> Tests won't pass due to the new dependency on the `MCP` module.

I can submit it to d:l:perl if you want

kraih:
> I can submit it to d:l:perl if you want

 That would be much appreciated, thanks.

kraih:
For the first two tools to actually do something useful, i was thinking of one that returns information about a specific job, and one that allows the LLM to add labels to a job.

kraih:
Seeing the chain of thought in Qwen 3 with tool calling is also quite interesting:
```
You
Who am i?

qwen3-14b
Thought for 2.01 seconds

Okay, the user is asking "Who am I?" in the context of openQA. I need to figure out which function to use here. The available tool is the current_user function, which retrieves the openQA name and id of the current user. Since the user wants to know their identity, calling this function makes sense. It doesn't require any parameters, so the arguments object is empty. I'll structure the tool call accordingly.

current_user({})
Arguments
{}
Result
[
  {
    "type": "text",
    "text": "name: Demo, id: 2"
  }
]
Thought for 2.58 seconds

Okay, let me process this. The user asked "Who am I?" and the system called the current_user function. The response from the tool says the name is Demo and the ID is 2. So, the user's openQA name is Demo, and their ID is 2. I should present this information clearly. Maybe start by answering the question directly, then provide the details. Make sure it's friendly and straightforward. Check if there's anything else needed, but since the user only asked for their identity, this should suffice.

You are a user in the openQA system. Your name is Demo and your user ID is 2. Let me know if you need further assistance!
```

lib/OpenQA/Shared/Controller/Auth.pm
```diff
@@ -56,7 +56,10 @@ sub auth ($self) {
     else {
 
         # Personal access token
-        if (my $userinfo = $self->req->url->to_abs->userinfo) {
+        if (($self->req->headers->authorization // '') =~ /^Bearer\s+(.+)$/) {
+            ($user, $reason) = $self->_token_auth($reason, $1);
+        }
+        elsif (my $userinfo = $self->req->url->to_abs->userinfo) {
```
Martchus, line 62: I guess this change and its corresponding unit tests could be split into a separate commit and already be merged.

lib/OpenQA/WebAPI.pm
```diff
@@ -51,6 +52,9 @@ sub startup ($self) {
     # take care of DB deployment or migration before starting the main app
     OpenQA::Schema->singleton;
 
+    # MCP server
+    my $mcp = OpenQA::MCP->new;
```
Martchus, line 56: The use of this should probably be configurable. I'm wondering whether moving this into a Mojolicious plugin would make sense. This would also avoid the hard dependency on the MCP module.

kraih: Agreed about making it configurable. But i think moving it into a plugin might make authentication awkward.

PR #6642, opened 2025-08-07
Fix/173881/client tests

Wabri:
Related to: https://progress.opensuse.org/issues/173881


script/client
perlpunk, line 1: Why did you change the script back to perl and added back the pod? Was a test failing?

Wabri: Since yesterday you told that probably the fact that the script was in perl was not the problem, I try to revert that back to perl and it works. if we prefer it in bash i can leave in bash no problem 👍

kalikiana: Maybe you could add a simple unit test to check that the script runs and shows the obsoletion message?

perlpunk: I think the script is simple enough to leave it at that. It has no dependencies, it works now, and noone should change it

t/31-client-obsolete.t
```diff
@@ -0,0 +1,34 @@
+# Copyright 2014-2025 SUSE LLC
+# SPDX-License-Identifier: GPL-2.0-or-later
+
+use Mojo::Base -signatures;
+use Test::Most;
+use Test::Warnings ':report_warnings';
+use Capture::Tiny qw(capture);
+use Test::Output qw(combined_like combined_unlike stderr_like);
```
Wabri:
```suggestion
```

t/31-client-obsolete.t
```diff
@@ -0,0 +1,34 @@
+# Copyright 2014-2025 SUSE LLC
+# SPDX-License-Identifier: GPL-2.0-or-later
+
+use Mojo::Base -signatures;
+use Test::Most;
+use Test::Warnings ':report_warnings';
```
Wabri:
```suggestion
```

t/31-client-obsolete.t
```diff
@@ -0,0 +1,34 @@
...
+use FindBin;
+my $script = path($FindBin::Bin, '..', 'script', 'client')->to_string;
+
+my $message = "NOTICE: openqa-client is obsolete. Use openqa-cli instead\n";
+
+sub run_client (@args) {
+    # https://perldoc.perl.org/functions/system
+    my ($out, $err) = capture { system $script, @args };
+    my $code = ${^CHILD_ERROR_NATIVE} % 255; # if exit 1 than 256 
+    return ($out, $err, $code);
```
Wabri:
```suggestion
    my ($out, $err) = capture { system $script, @args };
    my $code = ${^CHILD_ERROR_NATIVE} % 255;
    return ($out, $err, $code);
```

Wabri (commented):
Remove this notes and debug

t/31-client-obsolete.t
perlpunk: why do you want to remove this?

t/31-client-obsolete.t
perlpunk, line 1: I vote against this, see my other comment.

perlpunk (changes requested):

t/31-client-obsolete.t
Wabri: Because it's never used in this test

t/31-client-obsolete.t
Wabri: I'm ok with both of the decision, but I also want to present both of the solution! So we will wait for @kalikiana's answer to this thread https://github.com/os-autoinst/openQA/pull/6642#discussion_r2262370007 to continue the discussion

t/31-client-obsolete.t
perlpunk: `:report_warnings` enables the feature to report all received unexpected warnings at the end of the test. So this line uses the feature, so it is used in the test. We use it in every test.

t/31-client-obsolete.t
kalikiana: No strong opinion in this case.

Wabri: Ok let me remove this commit :cry:

perlpunk (approved):

PR #6649, opened 2025-08-11
Fix Too few arguments for subroutine error on close_connection

d3flex:
To prevent `Timer failed: Too few arguments for subroutine`, the `$self` can be removed from the close_connection function because the variable is still accessible within the closure, so it is not needed.

Not directly connected to the issue but related
poo: https://progress.opensuse.org/issues/187014

### DIFF
--- a/lib/OpenQA/Shared/Controller/Running.pm
+++ b/lib/OpenQA/Shared/Controller/Running.pm
@@ -113,8 +113,9 @@ sub streamtext ($self, $file_name, $start_hook = undef, $close_hook = undef) {
     # Setup utility function to close the connection if something goes wrong
     my $timer_id;
     my $close_connection = sub ($self) {
+        my $worker_id = $worker->id;
+        log_debug "Connection to worker with ID $worker_id will be closed because $logfile changed";
         Mojo::IOLoop->remove($timer_id);
-        $close_hook->();
         $self->finish;
         close $log;
     };
--- a/lib/OpenQA/Shared/Controller/Running.pm
+++ b/lib/OpenQA/Shared/Controller/Running.pm
@@ -130,7 +130,7 @@ sub streamtext ($self, $file_name, $start_hook = undef, $close_hook = undef) {
             # Zero tolerance for any shenanigans with the logfile, such as
             # truncation, rotation, etc.
             my @st = stat $logfile;
-            return $close_connection->() unless @st && $st[1] == $ino && $st[3] > 0 && $st[7] >= $size;
+            return $close_connection->($self) unless @st && $st[1] == $ino && $st[3] > 0 && $st[7] >= $size;
 
             # If there's new data, read it all and send it out. Then
             # seek to the current position to reset EOF.
--- a/t/26-controllerrunning.t
+++ b/t/26-controllerrunning.t
@@ -85,9 +85,15 @@ subtest streaming => sub {
         ok $size > (10 * 1024), 'test file size is greater than 10 * 1024';
         like $controller->tx->res->content->{body_buffer}, qr/data: \["A\\n"\]/, 'body buffer contains "A"';
         Mojo::IOLoop->one_tick;
+
+        $controller->streamtext('test.txt');
+        $t_file->remove();
+        Mojo::IOLoop->one_tick;
+        ok $c_finished, 'controller has closed after file removed';
     };
 
     subtest image => sub {
+        $c_finished = 0;
         $app->attr(schema => sub { FakeSchema->new() });
         my $controller = OpenQA::Shared::Controller::Running->new(app => $app);
         my $faketx = Mojo::Transaction::Fake->new(fakestream => $id);
--- a/t/26-controllerrunning.t
+++ b/t/26-controllerrunning.t
@@ -88,7 +88,9 @@ subtest streaming => sub {
 
         $controller->streamtext('test.txt');
         $t_file->remove();
-        Mojo::IOLoop->one_tick;
+        combined_like { Mojo::IOLoop->one_tick }
+        qr{Connection to worker .* will be closed because /tmp/.+/test.txt changed},
+          'connection termination is logged when log is truncated';
         ok $c_finished, 'controller has closed after file removed';
     };
 


lib/OpenQA/Shared/Controller/Running.pm
```diff
@@ -112,7 +112,7 @@ sub streamtext ($self, $file_name, $start_hook = undef, $close_hook = undef) {
     # Check for new lines from the logfile using recurring timer
     # Setup utility function to close the connection if something goes wrong
     my $timer_id;
-    my $close_connection = sub ($self) {
+    my $close_connection = sub {
```
perlpunk: The suggestion in the ticket was the opposite. Call `$close_connection` with `$self`.

A subroutine that uses an outer variable is problematic in the way that it only uses the instance of that variable when it was first called:
https://perldoc.perl.org/perlsub#Persistent-variables-with-closures

The same problem we have with `$close_hook`, I think.
The second problem with `$close_hook` is that below in line 152 we call it with 2 arguments, and here in line 117 it's called without arguments.
I think that part was never tested, and the lines are not covered.

You can ignore the `$close_hook` for now. I can take a look later. Since we don't use that feature anyway, I think it's not that important.

So please
* add `$self` as an argument
* Add a test in `t/26-controllerrunning.t` by reproducing the condition under which this `$close_connction` is called (e.g. a shrinked file)

d3flex:
> > Not directly connected to the issue but related
> > poo: https://progress.opensuse.org/issues/187014
> 
> Why is it not directly related to this issue?

from my POV the ticket #187014 is about the repeating alerts. Not because we have that specific error.

> Or were there simple many different live views at that time? Did you have a log at the access logs?

no idea. I dont know how can I tell whether `autoinst-log-live.txt` where open at the time to be honest.

> Also try to reproduce the error in a test running in your local instance. What happens if the logfile shrinks during the live view. Can it explain why we saw the error a thousand times?

Also this sounds challenging to me. I will need a lot of time to come up with results. And I dont think I will spend time on it at the moment.

perlpunk:
> Also this sounds challenging to me. I will need a lot of time to come up with results. And I don't think I will spend time on it at the moment.

Run a local instance. Watch the logs of the livehandler specifically.
Start a test and open the log viewer in the webui.
Look into the `/tmp` directory for the according `/tmp/webui.worker-...` directory, and do 
```
echo foo >/tmp/webui.worker-.../autoinst-log-live.txt
```
Check the livehandler logs.

I did this now, and the message is repeated, because the connection is never closed.
So that's why we got over 1000 errors in the log. This should be fixed with calling the subroutine in the correct way.

I think the fix could include at least a debug message that reports that the file was shrunk, or the inode changed. This shouldn't happen often, but if something is going wrong, we would have that in the logs.

Do you think you are able to add the unit test at least? If not, you can also unassign.

lib/OpenQA/Shared/Controller/Running.pm
```diff
@@ -112,9 +112,8 @@ sub streamtext ($self, $file_name, $start_hook = undef, $close_hook = undef) {
...
```
perlpunk: Please do not remove the signature.

perlpunk (changes requested):

lib/OpenQA/Shared/Controller/Running.pm
d3flex: I thought we agreed to revert this 8e91645.

perlpunk (approved):
btw, the PR title is still wrong

perlpunk:
> I think the fix could include at least a debug message that reports that the file was shrunk, or the inode changed. This shouldn't happen often, but if something is going wrong, we would have that in the logs.

btw, did you disagree with this?

d3flex:
> > I think the fix could include at least a debug message that reports that the file was shrunk, or the inode changed. This shouldn't happen often, but if something is going wrong, we would have that in the logs.
> 
> btw, did you disagree with this?

Sure. wait for it

lib/OpenQA/Shared/Controller/Running.pm
```diff
@@ -113,6 +113,7 @@ sub streamtext ($self, $file_name, $start_hook = undef, $close_hook = undef) {
     # Setup utility function to close the connection if something goes wrong
     my $timer_id;
     my $close_connection = sub ($self) {
+        log_debug "Connection to $worker will be terminated because $logfile changed";
```
perlpunk: Can you log the worker id instead? `$worker` as a string will only log the internal memory address

perlpunk (changes requested):

lib/OpenQA/Shared/Controller/Running.pm
```diff
@@ -113,8 +113,8 @@ sub streamtext ($self, $file_name, $start_hook = undef, $close_hook = undef) {
...
         Mojo::IOLoop->remove($timer_id);
-        $close_hook->();
```
Martchus, line 117: Why do you remove this? If you remove this because it is called via `$self->on(finish => …` anyway then `Mojo::IOLoop->remove($timer_id);` should *perhaps* be removed as well. It would make sense to think this though and state the reasoning in the commit message. Otherwise I'd refrain from removing this line.

If the close hook cannot handle the situation of sometimes being called with 2 arguments and sometimes being called with 0 arguments than the hook should be fixed or we always call it with 2 arguments.

lib/OpenQA/Shared/Controller/Running.pm
```diff
@@ -113,8 +113,8 @@ sub streamtext ($self, $file_name, $start_hook = undef, $close_hook = undef) {
     # Setup utility function to close the connection if something goes wrong
     my $timer_id;
     my $close_connection = sub ($self) {
+        log_debug "Connection to $worker will be terminated because $logfile changed";
```
Martchus: This looks wrong as `$worker` is not containing a merge ID/name. I also doubt that this log message is generally very useful during operation so either reduce the level to `trace` or remove it.

Martchus (changes requested):

lib/OpenQA/Shared/Controller/Running.pm
perlpunk: I suggested this log message above in https://github.com/os-autoinst/openQA/pull/6649#issuecomment-3174414209
> I think the fix could include at least a debug message that reports that the file was shrunk, or the inode changed. This shouldn't happen often, but if something is going wrong, we would have that in the logs.

So maybe you could comment on that. If we give @d3flex different suggestions, what should he do?

lib/OpenQA/Shared/Controller/Running.pm
perlpunk: To be fair, we don't use the `$close_hook`, and it's not really documented. So why would it be called with no arguments and later with two arguments?
What's the purpose of it and in which situations would it be used?

In the case of removing the timer I guess calling it twice can't hurt because the second time it's a no-op

lib/OpenQA/Shared/Controller/Running.pm
Martchus: This sounds more the use case of temporary debug printing than a debug message that should stay.

perlpunk: Well, apparently it doesn't happen very often, otherwise we would have seen the bug earlier (I remember seeing it once a few weeks ago in a logreport email), and it's something that normally shouldn't happen, right?
But I am ok with not adding it, if you think it's not helpful.

lib/OpenQA/Shared/Controller/Running.pm
Martchus: If we don't use it at all I suggest to remove all of its invocations consistently. (And of course also the function argument so we don't have an unused argument.) It would make sense to do this in a separate commit.

lib/OpenQA/Shared/Controller/Running.pm
```diff
@@ -113,6 +113,10 @@ sub streamtext ($self, $file_name, $start_hook = undef, $close_hook = undef) {
     # Setup utility function to close the connection if something goes wrong
     my $timer_id;
     my $close_connection = sub ($self) {
+        my $worker_id = $worker->id;
+        my $worker_instance = $worker->instance;
+        log_debug
+          "Connection to instance $worker_instance of worker $worker_id will be terminated because $logfile changed";
```
Martchus: This will print something like "Connection to instance 42 of worker 3563 will be …" which makes no sense.

The instance number is only meaningful when logging the host as well. However, the ID is already as good as it gets when identifying a worker so it is good enough to just log the ID:

```suggestion
        my $worker_id = $worker->id;
        log_debug "Connection to worker with ID $worker_id will be terminated because $logfile changed";
```

lib/OpenQA/Shared/Controller/Running.pm
```diff
@@ -113,8 +113,10 @@ sub streamtext ($self, $file_name, $start_hook = undef, $close_hook = undef) {
     # Setup utility function to close the connection if something goes wrong
     my $timer_id;
     my $close_connection = sub ($self) {
+        my $worker_id = $worker->id;
+        log_debug
+          "Connection to worker with ID $worker_id will be terminated because $logfile changed";
```
Martchus: Connections are usually *closed*:
```suggestion
        log_debug "Connection to worker with ID $worker_id will be closed because $logfile changed";
```

Not a big deal but I guess you'll have to push again anyway for the style checks.

okurz (changes requested):
There are two commits with the same subject. Please make 'em unique

perlpunk (approved):

t/26-controllerrunning.t
```diff
@@ -85,9 +85,17 @@ subtest streaming => sub {
         ok $size > (10 * 1024), 'test file size is greater than 10 * 1024';
         like $controller->tx->res->content->{body_buffer}, qr/data: \["A\\n"\]/, 'body buffer contains "A"';
         Mojo::IOLoop->one_tick;
+
+        $controller->streamtext('test.txt');
+        $t_file->remove();
+        combined_like { Mojo::IOLoop->one_tick }
+        qr/Connection to worker .* will be closed because \/tmp\/.+\/test.txt changed/,
```
perlpunk: I encourage everyone to use different regex delimiters to avoid backslash nightmares, but not blocking merge
```suggestion
        qr{Connection to worker .* will be closed because /tmp/.+/test.txt changed},
```

d3flex: nice one. I keep forgeting about those

okurz (approved):

PR #6651, opened 2025-08-12
WIP Add webhook functionality for Gitea

perlpunk:
Issue: https://progress.opensuse.org/issues/185491

This is just to push what I have so far.

I'm using the review_requested action from gitea, which leads to some differences in the payload. (There would also be differences in the pull_request open action, but less).
Currently Webhook.pm looks a bit chaotic because of this.
I'm thinking about how to restructure it.


lib/OpenQA/VcsHook.pm
d3flex, line 1: 👍

PR #6654, opened 2025-08-14
Add early die on unexpanded variables in clone attempt

okurz:
Related progress issue: https://progress.opensuse.org/issues/182645


t/35-script_clone_job_suse.t
```diff
@@ -24,6 +24,12 @@ package Test::FakeLWPUserAgent {
     sub get ($self, $url) { Test::FakeLWPUserAgentMirrorResult->new(is_success => $self->is_validrepo) }
 }    # uncoverable statement
 
+sub create_mock () {
+    my $fake_ua = Test::FakeLWPUserAgent->new;
+    my %url_handler = (remote_url => Mojo::URL->new('http://foo'), ua => $fake_ua);
+    return ($fake_ua, %url_handler);
```
Martchus, line 30: I don't think you can return a hash map. This code just expands the hash map again. So this is equivalent to:

```suggestion
    return ($fake_ua, remote_url => Mojo::URL->new('http://foo'), ua => $fake_ua);
```

Probably the following would be better:

```suggestion
    my %url_handler = (remote_url => Mojo::URL->new('http://foo'), ua => $fake_ua);
    return ($fake_ua, \%url_handler);
```

Of course you'll need to change the call-sites to expect a hash ref.

lib/OpenQA/Script/CloneJobSUSE.pm
```diff
@@ -28,6 +28,8 @@ sub verify_incident_repos ($url_handler, $incident_repos) {
     my @incident_urls;
     my $ua = $url_handler->{ua};
     foreach my $incident (split(/,/, $incident_repos)) {
+        die "URL '$incident' contains an unexpanded variable. See https://open.qa/docs/#_variable_expansion for details"
+          if $incident =~ /%/;
```
Martchus, line 32: The `%` sign is commonly used in [URL encoding](https://en.wikipedia.org/wiki/Percent-encoding) so this check seems quite wrong.

Martchus (changes requested):

t/35-script_clone_job_suse.t
perlpunk: > I don't think you can return a hash map.

You can, kind of.
It's correct that it will just be expanded to a list, but in the caller we have:
```
my ($fake_ua, %url_handler) = create_mock;
```
so the returned list will be assigned to a hash again.
But I would also prefer a hash ref in such cases.

Martchus: Yes, but really only "kind of". Try to pass 2 hash maps this way and the house of cards falls apart :-)

okurz:
Closing in favor of #6655 . We might not be able to easily introduce a check for unexpanded variables without actually trying to expand variables :)

PR #6655, opened 2025-08-14
Support expanding variables in openqa-clone-job

okurz:
Related progress issue: https://progress.opensuse.org/issues/182645

### DIFF
--- a/lib/OpenQA/JobSettings.pm
+++ b/lib/OpenQA/JobSettings.pm
@@ -12,6 +12,8 @@ use OpenQA::Log 'log_debug';
 use OpenQA::Utils qw(asset_type_from_setting get_url_short);
 use Feature::Compat::Try;
 
+my $PLACEHOLDER_RE = qr/(%+)(\w+)(%+)/;
+
 sub generate_settings ($params) {
     my $settings = $params->{settings};
     my @worker_class;
@@ -49,23 +51,27 @@ sub generate_settings ($params) {
 
     parse_url_settings($settings);
     handle_plus_in_settings($settings);
-    return expand_placeholders($settings);
+    my ($error) = expand_placeholders($settings);
+    return $error;
 }
 
 # replace %NAME% with $settings{NAME} (but not %%NAME%%)
 sub expand_placeholders ($settings, $on_web_ui = 1) {
-    for my $value (values %$settings) {
-        next unless defined $value;
-        my %visited_placeholders;
-        try {
-            $value =~ s/(%+)(\w+)(%+)/_expand_placeholder($settings, $2, $1, $3, \%visited_placeholders, $on_web_ui)/eg;
+    my %unexpanded;
+    my $err;
+    try {
+        for (grep { defined } values %$settings) {
+            my %visited;
+            s/$PLACEHOLDER_RE/_expand_placeholder($settings, $2, $1, $3, \%visited, $on_web_ui, \%unexpanded)/eg;
         }
-        catch ($e) { return "Error: $e" }
     }
-    return undef;
+    catch ($e) {
+        $err = "Error: $e";
+    }
+    return ($err, [sort keys %unexpanded]);
 }
 
-sub _expand_placeholder ($settings, $key, $start, $end, $visited_placeholders_in_parent_scope, $on_web_ui) {
+sub _expand_placeholder ($settings, $key, $start, $end, $visited, $on_web_ui, $unexpanded) {
     # handle %CASEDIR% only on web UI; on the worker it is handled by _engine_workit_step_2 separately
     return "$start$key$end" if !$on_web_ui && $key eq 'CASEDIR';
 
@@ -73,12 +79,17 @@ sub _expand_placeholder ($settings, $key, $start, $end, $visited_placeholders_in
     return (substr $start, 1) . ($key) . (substr $end, 0, -1) unless $start eq '%' && $end eq '%';
 
     # do not replace non-existing keys on web UI level to leave them to the worker
-    return $on_web_ui ? "$start$key$end" : '' unless defined(my $value = $settings->{$key});
+    my $value = $settings->{$key};
+    unless (defined $value) {
+        $unexpanded->{$key}++;
+        return $on_web_ui ? "$start$key$end" : '';
+    }
 
     # substitute the whole %…% expression with the value of the other setting
-    my %visited_placeholders = %$visited_placeholders_in_parent_scope;
-    die "The key $key contains a circular reference, its value is $value.\n" if $visited_placeholders{$key}++;
-    $value =~ s/(%+)(\w+)(%+)/_expand_placeholder($settings, $2, $1, $3, \%visited_placeholders, $on_web_ui)/eg;
+    my %visited_branch = %$visited;
+    die "The key $key contains a circular reference, its value is $value.\n" if $visited_branch{$key}++;
+    $value
+      =~ s/$PLACEHOLDER_RE/_expand_placeholder($settings, $2, $1, $3, \%visited_branch, $on_web_ui, $unexpanded)/eg;
     return $value;
 }
 
--- a/lib/OpenQA/Script/CloneJobSUSE.pm
+++ b/lib/OpenQA/Script/CloneJobSUSE.pm
@@ -10,16 +10,16 @@ use OpenQA::JobSettings;
 our @EXPORT_OK = qw(detect_maintenance_update);
 
 sub collect_incident_repos ($url_handler, $settings) {
-    my $error = OpenQA::JobSettings::expand_placeholders($settings);
-    die "Expanding variables failed: $error" if $error;
+    my ($error, $unexpanded) = OpenQA::JobSettings::expand_placeholders($settings);
+    die "Expanding variables failed: $error\n" if $error;
     if (my $repo = $settings->{INCIDENT_REPO}) {
-        return verify_incident_repos($url_handler, $repo);
+        return verify_incident_repos($url_handler, $repo, $unexpanded);
     }
     my @urls;
     if (my $addons = $settings->{SCC_ADDONS}) {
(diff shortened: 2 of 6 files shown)


lib/OpenQA/Script/CloneJobSUSE.pm
```diff
@@ -28,6 +30,9 @@ sub verify_incident_repos ($url_handler, $incident_repos) {
     my @incident_urls;
     my $ua = $url_handler->{ua};
     foreach my $incident (split(/,/, $incident_repos)) {
+        die
```
asmorodskyi: I don't think it is good idea , maybe just print warning instead ? It reminds me story around https://github.com/os-autoinst/openQA/pull/6468 which showed us that it is not always good thing to decide on behalf of caller if it is make sense to continue . I still see some cases where no matter that incident repos are not expanded run still make sense . For example some VRs where I simply skip use of this variables

lib/OpenQA/Script/CloneJobSUSE.pm
```diff
@@ -28,6 +30,9 @@ sub verify_incident_repos ($url_handler, $incident_repos) {
...
+"URL '$incident' contains an unexpanded variable. Specify the necessary variables at the command line for expansion. See https://open.qa/docs/#_variable_expansion for details"
+          if $incident =~ /%/;
```
Martchus: The `%` sign is commonly used in [URL encoding](https://en.wikipedia.org/wiki/Percent-encoding) so this check seems quite wrong.

lib/OpenQA/Script/CloneJobSUSE.pm
```diff
@@ -5,10 +5,12 @@ package OpenQA::Script::CloneJobSUSE;
 use Mojo::Base -strict, -signatures;
 use Data::Dump 'pp';
 use Exporter 'import';
+use OpenQA::JobSettings;
 
 our @EXPORT = qw(detect_maintenance_update);
 
 sub collect_incident_repos ($url_handler, $settings) {
+    OpenQA::JobSettings::expand_placeholders($settings);
```
Martchus: This function should probably return whether some placeholders could not be expanded. Don't do this by checking for a  `%` as mentioned in the other comment.

lib/OpenQA/Script/CloneJobSUSE.pm
Martchus: I wouldn't mind the `die`. If we continue here we would probably just run into a hard failure later so failing fast is probably the best thing to do here.

lib/OpenQA/Script/CloneJobSUSE.pm
```diff
@@ -28,6 +30,9 @@ sub verify_incident_repos ($url_handler, $incident_repos) {
     my @incident_urls;
     my $ua = $url_handler->{ua};
     foreach my $incident (split(/,/, $incident_repos)) {
+        die
+"URL '$incident' contains an unexpanded variable. Specify the necessary variables at the command line for expansion. See https://open.qa/docs/#_variable_expansion for details"
```
Martchus: This needs a trailing `\n`, otherwise Perl prints a line number and it looks like a mistake in the program itself:
```suggestion
"URL '$incident' contains an unexpanded variable. Specify the necessary variables at the command line for expansion. See https://open.qa/docs/#_variable_expansion for details\n"
```

Martchus (changes requested):
Not repeating my comments from https://github.com/os-autoinst/openQA/pull/6654. In cases like this it would probably make sense to just open on PR.

lib/OpenQA/Script/CloneJobSUSE.pm
asmorodskyi: @Martchus  what about the case when I clone job for some VR and simply skip install of updates ? We doing it quite often and just not bother about content of this variables

Martchus: And then you still use `--check-repos` and the checks pass despite the incomplete content of the variables? If that's the case then it should probably be a warning. Maybe you can give a concrete example of how you invoke `openqa-clone-job` so we're sure we're talking about the same thing.

asmorodskyi: > And then you still use --check-repos and the checks pass despite the incomplete content of the variables?

ah sorry I didn't realize that . so it will die only if `--check-repos` flag will be used ?

Martchus: I've just double-checked and code in `CloneJobSUSE.pm` is only executed if `--check-repos` is present.

asmorodskyi: ok than I am agree that we need to die here

lib/OpenQA/Script/CloneJobSUSE.pm
okurz: Ok. I suggest we do
* #6662

first to ensure that the worker has no problems with the return value of expanding and then I can try to incorporate that here.

lib/OpenQA/Script/CloneJobSUSE.pm
okurz: ok, done. Please check the updated additional commit.

lib/OpenQA/Script/CloneJobSUSE.pm
okurz: done, please check

lib/OpenQA/Script/CloneJobSUSE.pm
okurz: done

lib/OpenQA/Script/CloneJobSUSE.pm
```diff
@@ -5,10 +5,18 @@ package OpenQA::Script::CloneJobSUSE;
...
 use Exporter 'import';
+use OpenQA::JobSettings;
 
 our @EXPORT = qw(detect_maintenance_update);
 
 sub collect_incident_repos ($url_handler, $settings) {
+    my ($error, $unexpanded) = OpenQA::JobSettings::expand_placeholders($settings);
+    die "Expanding variables failed: $error\n" if $error;
+    if ($unexpanded && @$unexpanded) {
+        my $vars = join(', ', @$unexpanded);
```
perlpunk:
```suggestion
        my $vars = join ', ', @$unexpanded;
```

okurz: there should be no need to repeatedly point this out in human review when we have an automatic check. What do we need to enable such check?

lib/OpenQA/JobSettings.pm
```diff
@@ -54,31 +56,39 @@ sub generate_settings ($params) {
...
+        for (grep { defined } values %$settings) {
+            my %visited;
+            s/$PLACEHOLDER_RE/_expand_placeholder($settings, $2, $1, $3, \%visited, $on_web_ui, \%unexpanded)/eg;
         }
-        catch ($e) { return "Error: $e" }
+        return wantarray ? (undef, [sort keys %unexpanded]) : undef;
+    }
+    catch ($e) {
+        my $err = "Error: $e";
+        return wantarray ? ($err, [sort keys %unexpanded]) : $err;
```
Martchus: I'd rather refrain from using `wantarray`. It is probably not a big deal to update the other call-sites and less messy.

It would probably also make sense to simply always return `($err, [sort keys %unexpanded])`. If there was no error, `$err` will just be `undef` anyway. You only have to declare it outside the `catch`.

Martchus (changes requested):

lib/OpenQA/Script/CloneJobSUSE.pm
perlpunk: There you go: https://app.circleci.com/pipelines/github/os-autoinst/openQA/20138/workflows/8a55182e-630f-49ba-bbe5-a9eeff2096ee/jobs/202066

lib/OpenQA/JobSettings.pm
okurz: done

okurz:
ready to review again

lib/OpenQA/Script/CloneJobSUSE.pm
```diff
@@ -5,10 +5,18 @@ package OpenQA::Script::CloneJobSUSE;
 use Mojo::Base -strict, -signatures;
 use Data::Dump 'pp';
 use Exporter 'import';
+use OpenQA::JobSettings;
 
 our @EXPORT_OK = qw(detect_maintenance_update);
 
 sub collect_incident_repos ($url_handler, $settings) {
+    my ($error, $unexpanded) = OpenQA::JobSettings::expand_placeholders($settings);
```
Martchus, line 13: Maybe a stupid question, but how will this magically fill in `%REPO_MIRROR_HOST%` (the variable mentioned in https://progress.opensuse.org/issues/182645)? I thought this variable comes from the worker config and is therefore not resolved right until the job is executed.

Martchus: And for variables that can be resolved here this change is also not very useful because those variable have already been resolved at the time of scheduling (in the call you adjust in `ScheduledProducts.pm`). Or do I miss something here?

okurz: Honestly I don't remember from my 2 year old changes what I fully intended but I guess you are right. openqa-clone-job running on the client side cannot expand variables that are evaluated on the worker side but we can still add support to expand *if* specified, e.g. for cloning to local workers that don't have those settings.

The previous implementation in this branch was too aggressive and would fail the clone job entirely if any variables remained unexpanded. I just pushed a change that corrects this behavior by warning instead of die.

Martchus: I guess with "Specify the necessary variables at the command line for expansion." this change makes sense.

Martchus (approved):

aylen384 (approved):

PR #6656, opened 2025-08-14
Improve error handling of worker

Martchus:
See https://progress.opensuse.org/issues/186308 and particular commit messages

---

I tested this by inserting `die`s in various places in the callback-chains of the upload and stop code.

Without this change the problem mentioned in the ticket is easily reproducible. So the worker stays responsive (e.g. maintains its ws connection) but it waits forever for the upload/stopping to complete and sending `SIGHUP` or `SIGINT`/`SIGTERM` doesn't help.

With this change the worker is actually stopped after `Stopping because a critical error occurred.` is logged and the job is marked as "incomplete" with reason "exception". If the `die` is inserted at a place that's also critical for stopping one gets also `Another error occurred when trying to stop gracefully due to an error` and `Trying to kill ourself forcefully now` and the job is not marked as "incomplete" if the exception happened before. The worker is terminated in any case, though.

With the 2nd commit also a non-zero return code is returned in the error case.

### DIFF
--- a/lib/OpenQA/Worker.pm
+++ b/lib/OpenQA/Worker.pm
@@ -351,7 +351,7 @@ sub init ($self) {
             });
     }
 
-    return $return_code;
+    return \$return_code;
 }
 
 sub configure_cache_client ($self) {
@@ -367,7 +367,7 @@ sub configure_cache_client ($self) {
 sub exec ($self) {
     my $return_code = $self->init;
     Mojo::IOLoop->singleton->start;
-    return $return_code;
+    return $$return_code;
 }
 
 sub _prepare_cache_directory ($webui_host, $cachedirectory) {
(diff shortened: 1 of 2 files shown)


lib/OpenQA/Worker.pm
```diff
@@ -347,7 +351,7 @@ sub init ($self) {
             });
     }
 
-    return $return_code;
+    return \$return_code;
```
perlpunk, line 354: Why this change?

perlpunk (approved):

lib/OpenQA/Worker.pm
d3flex, line 1: Not directly connected to the PR but once you touch this code,  would it make sense to use signatures for the `error` callback https://github.com/os-autoinst/openQA/pull/6656/files#diff-a503e62ccc687387da869355470f852d01c022345aab88b13f1f8618a6c05bf4R288

```
error => sub ($reactor, $err) {
```

d3flex: nevermind. just saw https://github.com/os-autoinst/openQA/pull/6657. 👍👍

PR #6657, opened 2025-08-14
Use signatures in worker job code and related mocking

Martchus:
Based on #6656 as I noticed that many of the functions in that area don't use signatures yet. We should probably merge #6656 first.


lib/OpenQA/Worker.pm
```diff
@@ -287,11 +287,15 @@ sub init ($self) {
     Mojo::IOLoop->singleton->reactor->on(
         error => sub {
```
d3flex: this was missed

d3flex (changes requested):

lib/OpenQA/Worker.pm
Martchus: With the signatures change I was focusing on `Job.pm` and none of the commit messages say that 100 % of a certain file is covered.

Martchus: I covered this function now nevertheless.

PR #6659, opened 2025-08-15
Worker::Engines::isotovideo: Simplify using more Mojo::File

okurz:


lib/OpenQA/Worker/Engines/isotovideo.pm
```diff
@@ -51,14 +51,7 @@ sub set_engine_exec ($path) {
...
     die 'cannot get environment variables!\n' unless $vars;
-    my $fn = $pooldir . '/vars.json';
-    unlink "$pooldir/vars.json" if -e "$pooldir/vars.json";
-    open(my $fd, '>', $fn) or die "can not write vars.json: $!\n";
-    fcntl($fd, F_SETLKW, pack('ssqql', F_WRLCK, 0, 0, 0, $$)) or die "cannot lock vars.json: $!\n";
-    truncate($fd, 0) or die "cannot truncate vars.json: $!\n";
-
-    print $fd Cpanel::JSON::XS->new->pretty(1)->encode(\%$vars);
-    close($fd);
+    path($pooldir, 'vars.json')->spew(Cpanel::JSON::XS->new->pretty(1)->encode(\%$vars));
```
Martchus, line 54: The previous code tried to handle the case when this function is executed concurrently. Is this also handled by `spew`? (This is of course irrelevant when the function is not actually invoked concurrently.)

But to be fair, I guess the previous code didn't handle this in a good way, too. It actually tried to unlink the file before acquiring the lock and I am not sure whether that's a good idea.

perlpunk: I was wondering about that, too.
But at least the code was assuring that the file contained valid JSON, I think.
`spew` won't guarantee that.
But i'm wondering as well when our code would concurrently try to write this file.
To be safe, maybe the easiest thing here would be to first write to `"$pooldir/vars.json.$$"` and then `rename` it to its target.

Martchus: Yes, renaming is probably the best way.

okurz: Can you elaborate under which circumstances we would need to first write to a temp file and then rename to the target?

Martchus: > under which circumstances …

> … when this function is executed concurrently

okurz: > > … when this function is executed concurrently

but when would this code be executed concurrently for one individual worker instance?

Martchus: Looks like this function is only executed once and that's before creating the isotovideo process. So I guess it is ok not to worry about concurrent execution then.

Martchus (approved):

lib/OpenQA/Worker/Engines/isotovideo.pm
```diff
@@ -294,20 +287,10 @@ sub engine_workit ($job, $callback) {
...
     my $job_settings = $job_info->{settings};
-    for my $i (qw(QEMUPORT VNC OPENQA_HOSTNAME)) {
-        $job_settings->{$i} = $ENV{$i};
-    }
-    if (open(my $fh, '>', 'job.json')) {
-        print $fh Cpanel::JSON::XS->new->pretty(1)->encode($job_info);
-        close $fh;
-    }
+    $job_settings->{$_} = $ENV{$_} for qw(QEMUPORT VNC OPENQA_HOSTNAME);
+    path('job.json')->spew(Cpanel::JSON::XS->new->pretty(1)->encode($job_info));
```
perlpunk, line 293: Before this was silently ignoring when the file could not be opened, now it will die.
But I guess it's ok.

lib/OpenQA/Worker/Engines/isotovideo.pm
perlpunk: So either there was a possiblity in the past that this could be running concurrently, or it was overengineering :)

perlpunk (approved):

lib/OpenQA/Worker/Engines/isotovideo.pm
okurz: Yes, one of those

lib/OpenQA/Worker/Engines/isotovideo.pm
okurz: Yes, we should see if this makes problem from production

PR #1019, opened 2016-11-18
Add serial terminal output to live view

richiejp:
This is my initial attempt at adding serial console/terminal output to the live view in the same vein as Travis CI's output although much more simple.

Requires: https://github.com/os-autoinst/os-autoinst/pull/644

Feature request: https://progress.opensuse.org/issues/14690

PR #6661, opened 2025-08-15
Clean up streamtext signature from unused params

d3flex:
The signature introduced in https://github.com/os-autoinst/openQA/pull/1019 which was also implemented the underline functions and seems had different design, until https://github.com/os-autoinst/openQA/pull/5060. The functions used websockets to send to the worker a $command. Both were replaced in some point and the corresponding functions can be found as `start_livelog` and `stop_livelog` in the Jobs.pm. They are assigned to the some contant variables and mainly are triggered by `streaming` function via `_send_livestream_command_to_worker`.

All this time the `$start_hook` and `$close_hook` were noop apparently, without any defect, AFAIT. Removing it comes with a small tidy and moving everything from close_connection to finish.

### DIFF
--- a/lib/OpenQA/Shared/Controller/Running.pm
+++ b/lib/OpenQA/Shared/Controller/Running.pm
@@ -75,14 +75,10 @@ sub edit ($self) {
     $self->redirect_to('edit_step', moduleid => $running_module->name(), stepid => $stepid);
 }
 
-sub streamtext ($self, $file_name, $start_hook = undef, $close_hook = undef) {
+sub streamtext ($self, $file_name) {
     my $job = $self->stash('job');
     my $worker = $job->worker;
-    $start_hook ||= sub (@) { };
-    $close_hook ||= sub (@) { };
     my $logfile = $worker->get_property('WORKER_TMPDIR') . "/$file_name";
-
-    $start_hook->($worker, $job);
     $self->render_later;
     Mojo::IOLoop->stream($self->tx->connection)->timeout(900);
     my $res = $self->res;
@@ -113,10 +109,7 @@ sub streamtext ($self, $file_name, $start_hook = undef, $close_hook = undef) {
     # Setup utility function to close the connection if something goes wrong
     my $timer_id;
     my $close_connection = sub ($self) {
-        Mojo::IOLoop->remove($timer_id);
-        $close_hook->();
         $self->finish;
-        close $log;
     };
     $timer_id = Mojo::IOLoop->recurring(
         TEXT_STREAMING_INTERVAL() => sub (@) {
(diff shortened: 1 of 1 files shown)


lib/OpenQA/Shared/Controller/Running.pm
```diff
@@ -113,10 +109,7 @@ sub streamtext ($self, $file_name, $start_hook = undef, $close_hook = undef) {
     # Setup utility function to close the connection if something goes wrong
     my $timer_id;
     my $close_connection = sub ($self) {
-        Mojo::IOLoop->remove($timer_id);
```
kalikiana: Why are you removing the timer? I see no new code covering this here, and the timer is still being used 🤔

d3flex: `close_connection` calls finish which is taking care of this. that was exactly the objective.. to handle it consistently and not call it twice

Martchus: With the given explanation I also think it makes sense to remove the removal of the timer here.

lib/OpenQA/Shared/Controller/Running.pm
```diff
@@ -113,10 +109,7 @@ sub streamtext ($self, $file_name, $start_hook = undef, $close_hook = undef) {
...
-        $close_hook->();
         $self->finish;
-        close $log;
     };
```
Martchus: I think we no longer need `$close_connection` then. The code below can just invoke `$self->finish` directly.

lib/OpenQA/Shared/Controller/Running.pm
```diff
@@ -149,7 +142,7 @@ sub streamtext ($self, $file_name, $start_hook = undef, $close_hook = undef) {
     $self->on(
         finish => sub (@) {
             Mojo::IOLoop->remove($timer_id);
-            $close_hook->($worker, $job);
```
Martchus: This function could be reformatted to be a one-liner.

d3flex: No change here.

lib/OpenQA/Shared/Controller/Running.pm
```diff
@@ -107,11 +107,7 @@ sub streamtext ($self, $file_name) {
 
     # Check for new lines from the logfile using recurring timer
     # Setup utility function to close the connection if something goes wrong
-    my $timer_id;
```
Martchus, line 110: The removal of `$timer_id` looks wrong. It is still used in `$self->on(finish …`.

d3flex: but it wasnt removed, was it?

Martchus: Right, somehow I was under the impression it was.

Martchus (approved):

PR #6662, opened 2025-08-16
Handle 'expand_placeholders' return value consistently

okurz:


lib/OpenQA/Worker/Engines/isotovideo.pm
```diff
@@ -321,8 +321,8 @@ sub engine_workit ($job, $callback) {
     );
 
     # do final variable expansion so placeholders of variables defined in worker config are replaced as well
-    OpenQA::JobSettings::expand_placeholders(\%vars, 0);
-
+    my $error = OpenQA::JobSettings::expand_placeholders(\%vars, 0);
```
d3flex, line 324: I just wonder why not throw and exception or that  `log_error` in the `catch ($e)` of `expand_placeholders`!

Martchus: @d3flex If you wanted `expand_placeholders` to throw an exception like this you would probably just remove the `catch ($e)` there entirely. However, I think here a return value makes sense. Either the immediate caller handlers the error or not - one would probably not want to propagate the error even further.

d3flex: https://github.com/os-autoinst/openQA/blob/131c0f0a9e01ed7989f34a757d4e9801bbc5b430/lib/OpenQA/JobSettings.pm#L63

just returns `"Error: $e"`. my question is why not handle the error there. for instance 

``` lib/OpenQA/JobSettings.pm
catch ($e) { log_error("Expanding placeholders failed: $e") }
```

perlpunk: Because it's better in this case to let the caller decide what to do with the error message.
The function might not be called in a context where we would want to log the error, but to further return the message up the call stack.

d3flex: Still.. 

```openQA/lib/OpenQA/JobSettings.pm
catch ($e) { log_error("Expanding placeholders failed: $e"); return "$e"; }
```

perlpunk: But sometimes we don't want to log this error. E.g. when called from `schedule_iso` from the Iso API, the error is supposed to be returned as a result of the API call.
Why should there be something in our logs if we can't do anything about it, and it's not "our fault"?

perlpunk: It's true that the alternative would be to just throw an exception and then let the callers handle it, instead of returning a hashref and let callers chec if there is an `error` key in the hashref.
It's even possible to throw exceptions in perl which are hashrefs or objects, so the caller would not just get a string, but could pick the error data from the exception and decide how to return it to the according API user, or in this case in this PR, just log it.

okurz: > It's true that the alternative would be to just throw an exception

that's correct and I considered to do so. However then found that the function already returns a value but that we don't use it everywhere hence this PR

PR #6665, opened 2025-08-18
Fix example in documentation of openqa-setup-db

d3flex:


docs/Contributing.asciidoc
```diff
@@ -408,7 +408,7 @@ specify the user and database name and run it as user `postgres`:
 
 [source,sh]
 ----
-sudo sudo -u postgres openqa-setup-db your_username openqa-local`
+sudo su -u postgres openqa-setup-db your_username openqa-local
```
baierjan, line 411:
```suggestion
sudo -u postgres openqa-setup-db your_username openqa-local
```
or (if really needed)

```suggestion
sudo su - postgres openqa-setup-db your_username openqa-local
```

`su` does not have `-u` option

d3flex: personally I use `sudo -u postgres` and I would have remove the su but it is another reference which give the command as `sudo su -u postgres`. https://github.com/os-autoinst/openQA/pull/6665/files#diff-4d6d1050121df87969e32770abd50500a1fbda9f136db4fbb374fe05a236f2c2R393

baierjan: the reference you are mentioning gives `sudo su - postgres`, the missing `u` is not a typo 😉

Wabri: why not just using the sudo -u postgres?

Martchus: The `sudo sudo -u postgres` command is useful if you're not already root. If you just run `sudo -u postgres` as your normal user you might be prompted for the password of the postgres user (depending on the sudo configuration) which you probably wouldn't know.

PR #6667, opened 2025-08-20
Disallow browsing HTML assets/results/logs by default

Martchus:
* Browsing these files can be dangerous as the browser session could be used for malicious actions
* Enforce the download of HTML files in the web application as secure default (can still be configured otherwise if convenience is preferred)
* Enforce the download of all assets in the default NGINX config (can of course be commented out if convenience is preferred)
* Add a non-binary config option to allow extending it in the future with a policy like `subdomain:…` to be able to have both - security and convenience - with a little bit more configuration effort
* See https://progress.opensuse.org/issues/184456

### DIFF
--- a/lib/OpenQA/Setup.pm
+++ b/lib/OpenQA/Setup.pm
@@ -78,6 +78,7 @@ sub read_config ($app) {
             monitoring_enabled => 0,
             plugins => undef,
             hide_asset_types => 'repo',
+            file_security_policy => 'download-prompt',
             recognized_referers => '',
             changelog_file => '/usr/share/openqa/public/Changelog',
             job_investigate_ignore => '"(JOBTOKEN|NAME)"',
@@ -288,6 +289,7 @@ sub read_config ($app) {
     $global_config->{parallel_children_collapsable_results_sel}
       = ' .status' . join('', map { ":not(.result_$_)" } split(/\s+/, $results));
     _validate_worker_timeout($app);
+    _validate_security_policy($app, $global_config);
     _set_default_storage_durations($_) for $config->{default_group_limits}, $config->{no_group_limits};
     return $config;
 }
@@ -301,6 +303,13 @@ sub _validate_worker_timeout ($app) {
             'The specified worker_timeout is invalid and will be ignored. The timeout must be an integer greater than '
               . MAX_TIMER
               . '.');
+    }
+}
+
+sub _validate_security_policy ($app, $config) {
+    if ($config->{file_security_policy} !~ m/^(download-prompt|insecure-browsing)$/) {
+        $config->{file_security_policy} = 'download-prompt';
+        $app->log->warn('Invalid file_security_policy specified, defaulting to "download-prompt"');
     }
 }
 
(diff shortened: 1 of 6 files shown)


okurz (approved):

lib/OpenQA/Setup.pm
```diff
@@ -304,6 +306,13 @@ sub _validate_worker_timeout ($app) {
     }
 }
 
+sub _validate_security_policy ($app, $config) {
+    if ($config->{file_security_policy} !~ m/^(download-prompt|insecure-browsing)$/) {
```
kalikiana, line 310: I would rather not include `insecure-browsing` at all. What's the use case for this?

kalikiana (changes requested):

lib/OpenQA/Setup.pm
Martchus: A setup only used by trusted users (no test scheduling based on PRs or the like) where the admin doesn't want to take the effort of setting up a subdomain. The subdomain feature is only present in the next PR anyway - so for this first step this is the *only* way to restore the current convenient behavior.

Besides, without this AC1 of the related ticket would not be fulfilled.

I also think the `insecure-...` makes it quite obvious that this is, well, insecure. So I wouldn't consider it problematic. (And also because it is/was the default for years now and good enough all that time.)

d3flex: The only thing I have to ask in regards, is whether would make sense to _highlight_ somehow that the app is "insecure" when the `insecure-browsing` is set. I am not asking for a change but maybe something to keep in mind for later

kalikiana: > it is/was the default for years now

The ticket was filed because that's a terrible idea. Maybe we were lucky. Let's not use that as an argument 🙃

I don't get how you would argue convenience over security. But you have three people on your side, so that's completely fair.

kalikiana (approved):
I don't understand the reasoning but I'm okay to be outvoted.

lib/OpenQA/WebAPI/Controller/File.pm
```diff
@@ -143,6 +144,23 @@ sub test_asset ($self) {
...
+    # guess content type from extension
+    my $headers = $self->res->headers;
+    return $headers->content_type('application/octet-stream') unless $filename =~ m/\.([^\.]+)$/;
+    my $ext = $1;
+    my $as_attachment = 1;
+    if (my $filetype = $self->app->types->type($ext)) {
+        $headers->content_type($filetype);
+        $headers->header('X-Content-Type-Options', 'nosniff') if $filetype =~ qr|^text/plain;?|;
+        my $allow_insecure = $self->app->config->{global}->{file_security_policy} ne 'download-prompt';
+        $as_attachment = 0 if ($allow_insecure || $filetype !~ m|html|) && $ext ne 'iso';
```
aylen384, line 158: It might not be sufficient to only disallow html.
It is also possible to execute javascript from `.svg` files.

aylen384: https://github.com/os-autoinst/openQA/pull/6752

PR #6668, opened 2025-08-20
Allow configuring subdomain for serving logs/assets more securely

Martchus:
* Based on https://github.com/os-autoinst/openQA/pull/6667 so it makes
   most sense to review only the 2nd commit here (after reviewing #6667)
* Allow configuring a redirection to a configurable subdomain
* Update NINGX templates to cover files served directly via NGINX as well
    * Move most of `openqa-locations.inc` to a separate file to be able to
      include it separately (without asset locations); none of the code in
      `openqa-endpoints.inc` is therefore new
    * Keep asset location in `openqa-locations.inc` to avoid breaking
      too much; this way one can still include `openqa-locations.inc` as
      before covering assets and endpoints
* See https://progress.opensuse.org/issues/184456

### DIFF
--- a/etc/nginx/vhosts.d/openqa-locations.inc
+++ b/etc/nginx/vhosts.d/openqa-locations.inc
@@ -1,82 +1,10 @@
-root /usr/share/openqa/public;
-
-client_max_body_size 0;
-
-# The "client_body_buffer_size" value should usually be larger
-# than the UPLOAD_CHUNK_SIZE used by openQA workers, so there is
-# no excessive buffering to disk
-client_body_buffer_size 2m;
-
-# Default is exact which would need an exact match of Last-Modified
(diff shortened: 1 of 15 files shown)


kalikiana (approved):
Looks sensible. I didn't wanna rush reviewing this, so only hitting the button now that I (think I) get the logic.

etc/nginx/vhosts.d/openqa-locations.inc
perlpunk, line 1: https://progress.opensuse.org/issues/188208
```
Sep 03 07:56:59 susetest nginx[3481]: nginx: [emerg] unexpected end of file, expecting ";" or "}" in /etc/nginx/vhosts.d/openqa-locations.inc:11
```

PR #6670, opened 2025-08-20
Document semantics of low limits for important data

kalikiana:
As per the implementation, important limits are only applied if they are higher than regular limits.

See: https://progress.opensuse.org/issues/187590

### DIFF
--- a/docs/UsersGuide.asciidoc
+++ b/docs/UsersGuide.asciidoc
@@ -1408,13 +1408,14 @@ Keep logs for:: How long *logs* of non-important jobs are retained after
   they finished. *Logs* are files listed under "Result files" in the
   "Logs & Assets" tab on the job details page, e.g. `autoinst-log.txt`.
 Keep important logs for:: How long logs of an *important* job are retained after
-  it finished.
+  it finished. This limit only applies if it's higher than regular logs.
 Keep results for:: How long *results* of non-important job are retained after
   they finished. *Results* are all additional files stored on disk for a job
   (excluding assets) such as screenshots. This also includes logs and therefore
   this retention must be longer (or equal) than the retention of *logs*.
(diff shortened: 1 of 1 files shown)


kalikiana:
Btw the updated docs [look like this](https://github.com/kalikiana/openQA/blob/docs_important_limits/docs/UsersGuide.asciidoc#cleanup-of-assets-results-and-other-data)

Wabri (approved):

docs/UsersGuide.asciidoc
```diff
@@ -1408,21 +1408,21 @@ Keep logs for:: How long *logs* of non-important jobs are retained after
...
 Keep important logs for:: How long logs of an *important* job are retained after
-  it finished.
+  it finished. This limit only applies if it's higher than regular logs.
 Keep results for:: How long *results* of non-important job are retained after
   they finished. *Results* are all additional files stored on disk for a job
   (excluding assets) such as screenshots. This also includes logs and therefore
   this retention must be longer (or equal) than the retention of *logs*.
 Keep important results for:: How long results of *important* jobs are retained
-  after they finished.
+  after they finished. This limit only applies if it's higher than regular results.
```
okurz: sorry, you need to wrap the lines :)

okurz (changes requested):

docs/UsersGuide.asciidoc
kalikiana: Do I? Ci is not failing 🙃

d3flex (commented):
withdraw request for changes 🙏

docs/UsersGuide.asciidoc
kalikiana: I guess it's the only 83 character line in this particular paragraph? So hopefully 80 is what you were asking for? I can wrap this line then.

PR #1571, opened 2018-01-30
Allow download of a full job data from the client

foursixnine:
This is meant to add support for the openQA::Client to support download of all the assets and generated data from a job. 

In theory this is another stepping stone for the worker_bridge/federation

https://progress.opensuse.org/issues/14580

PR #6672, opened 2025-08-21
Use dedicated option to download iso and hdd with archive subcommand

d3flex:
Based on the history this did not meant to download other assets than 'other'
assets and it is implemeted using the `asset-size-limit` parameter, which was
set in a quite small size.
https://github.com/os-autoinst/openQA/pull/1571#discussion_r165847808
    
With this size in place, the error is confusing and not intuitive. It makes
more sense to remove the limit in case you want to get the full assets.
    
- Introduce ~`--all`~ `no-limit` to skip _asset-size-limit_. short `-u` as unlimit
- keep _asset-size-limit_ for the 'other' assets

### DIFF
--- a/lib/OpenQA/CLI/archive.pm
+++ b/lib/OpenQA/CLI/archive.pm
@@ -21,7 +21,8 @@ sub command ($self, @args) {
             url => $url,
             archive => $path,
             'with-thumbnails' => $options{'with-thumbnails'},
-            'asset-size-limit' => $options{'asset-size-limit'}});
+            'asset-size-limit' => $options{'asset-size-limit'},
+            'no-limit' => $options{'no-limit'}});
 
     return 0;
 }
--- a/lib/OpenQA/Client/Archive.pm
+++ b/lib/OpenQA/Client/Archive.pm
@@ -18,7 +18,8 @@ sub run ($self, $options) {
     my $res = $req->res;
     my $default_max_asset_size = 1024 * 1024 * 200;
     $options->{'asset-size-limit'} //= $default_max_asset_size;
-    $self->client->max_response_size($options->{'asset-size-limit'});
+    $options->{'no-limit'} //= 0;
+    $self->client->max_response_size($options->{'no-limit'} ? 0 : $options->{'asset-size-limit'});
 
     my $code = $res->code;
     die "There's an error openQA client returned $code" unless $code eq 200;
@@ -32,7 +33,7 @@ sub run ($self, $options) {
     delete($job->{assets}->{repo});
 
     $path->child('testresults', 'thumbnails')->make_path if $options->{'with-thumbnails'};
-    $self->_download_assets($url, $job, $path);
+    $self->_download_assets($url, $job, $path) if $options->{'no-limit'};
     $self->_download_test_results($url, $job, $path, $options);
 }
 
--- a/public/openqa-cli.yaml
+++ b/public/openqa-cli.yaml
@@ -163,6 +163,7 @@ subcommands:
     options:
       - asset-size-limit|l=i   --Asset size limit in bytes
       - with-thumbnails|t      --Download thumbnails as well
+      - no-limit|u             --Download all assets without size limits
 
   schedule:
     summary: Schedules a set of jobs (via "isos post" creating a schedule product)
--- a/t/43-cli-archive.t
+++ b/t/43-cli-archive.t
@@ -15,6 +15,7 @@ use OpenQA::CLI;
 use OpenQA::CLI::archive;
 use OpenQA::Test::Case;
 use OpenQA::Test::TimeLimit '10';
+use OpenQA::Utils qw(assetdir);
 
 OpenQA::Test::Case->new->init_data(fixtures_glob => '01-jobs.pl 03-users.pl 05-job_modules.pl');
 
@@ -144,6 +145,17 @@ subtest 'Archive job with thumbnails' => sub {
     ok -f $results->child('details-isosize.json'), 'details-isosize.json exists';
     ok -f $results->child('autoinst-log.txt'), 'autoinst-log.txt exists';
     ok -d $results->child('thumbnails'), 'thumbnails';
+};
+
+subtest 'Archive job with assets' => sub {
+    note('Asset directory: ' . assetdir());
+    my $target = $dir->child('archive-with-assets')->make_path;
+    my ($stdout, $stderr, @result)
+      = capture_stdout sub { $cli->run('archive', @host, '--no-limit', 99963, $target->to_string) };
+    like $stdout, qr{Downloading.*/iso/openSUSE-13.1-DVD-x86_64-Build0091-Media.iso}, 'download iso without size limit';
+    my $expected_iso = 'openSUSE-13.1-DVD-x86_64-Build0091-Media.iso';
+    my $iso_path = $target->child('iso', $expected_iso);
+    ok(-f $iso_path, "Asset was downloaded");
 };
 
 done_testing();


lib/OpenQA/CLI/archive.pm
```diff
@@ -21,7 +21,8 @@ sub command ($self, @args) {
             url => $url,
             archive => $path,
             'with-thumbnails' => $options{'with-thumbnails'},
-            'asset-size-limit' => $options{'asset-size-limit'}});
+            'asset-size-limit' => $options{'asset-size-limit'},
+            'all' => $options{'all'}});
```
Wabri:
```suggestion
            'all' => $options{'all'},
        });
```

Or maybe:

```suggestion
            'all' => $options{'all'},
        }
     );
```

what `./tools/tidyall` do?

d3flex: `tidyall` is fine with that. and I didnt alter it from its original. So lets skip this, if you dont mind.

public/openqa-cli.yaml
```diff
@@ -163,6 +163,7 @@ subcommands:
     options:
       - asset-size-limit|l=i   --Asset size limit in bytes
       - with-thumbnails|t      --Download thumbnails as well
+      - all|a      --Download including iso, hdd as well
```
perlpunk: Like discussed in the daily, your change is, strictly speaking, not about the type of files to download. It's just that iso and hdd are typically bigger than the default limit.
Maybe call the option `--no-limit`.
And as another improvement, when an asset is bigger than the limit, then print its size so one can easily specify the required limit.
What would also be cool, if one could specify a limit in a format like `100M` or `2G`, but I would suggest doing a single PR for each change.

d3flex: > Like discussed in the daily, your change is, strictly speaking, not about the type of files to download. It's just that iso and hdd are typically bigger than the default limit. 

I didnt say that. I guess is how you see it. if you go back in time, at least in my understanding, that was meant mainly for the test results, and thats why they set a small limit. 

> Maybe call the option `--no-limit`. 

`--no-limit` sounded right in first place. in a second thought, it is not exactly what the change achieves. 
That means that a simple rename of the parameter will not reflect the implementation. I will have to remove the limit completely within the `--no-limit` flow. just keep that in mind.

> And as another improvement, when an asset is bigger than the limit, then print its size so one can easily specify the required limit. What would also be cool, if one could specify a limit in a format like `100M` or `2G`, but I would suggest doing a single PR for each change.

💡

kalikiana: The PR description says `Introduce --all to skip asset-size-limit`. That aligns with this change:

    - $self->client->max_response_size($options->{'asset-size-limit'});
    + $self->client->max_response_size($options->{all} ? 0 : $options->{'asset-size-limit'});

However there's another change:

    - $self->_download_assets($url, $job, $path);
    + $self->_download_assets($url, $job, $path) if $options->{all};

So the code no longer downloads all assets by default. Instead it will only download results.

I'm not sure that behavioral change is something we want? If scripts rely on this it might break stuff.

d3flex: that's correct. first it keeps the same flow for the test results, including the limits. it just prevents to even try the `_download_assets` unless is explicitly asked. that prevents the errors and simplify the flow for images which will never fit the size limit. 
secondly, I will point you to #1571 which implies that the subcommand meant for the test results and the size limit, was introduced to enable largest assets, like isos, i guess

d3flex: `--all` rename `--no-limit`

kalikiana: > I will point you to https://github.com/os-autoinst/openQA/pull/1571 which implies that the subcommand meant for the test results and the size limit

That was suggested but not implemented, though. So as it is right now you're breaking compatibility.

lib/OpenQA/Client/Archive.pm
```diff
@@ -32,7 +33,7 @@ sub run ($self, $options) {
     delete($job->{assets}->{repo});
 
     $path->child('testresults', 'thumbnails')->make_path if $options->{'with-thumbnails'};
-    $self->_download_assets($url, $job, $path);
+    $self->_download_assets($url, $job, $path) if $options->{'no-limit'};
```
kalikiana, line 36: This is very misleading. `--no-limit` implies this is about the size. But the code no longer downloads assets by default.

kalikiana (changes requested):

lib/OpenQA/Client/Archive.pm
kalikiana: I suggest to clarify what behavior we want. The PR description is not clear right now.

For instance "Only download test results by default unless --download-assets is specified".

And maybe `--no-limit` should really only set `asset-size-limit=0`? That's what I would expect as a user without digging through the code.

kalikiana: And if this ends up breaking compatibility it needs to be announced to a wider audience. Assuming people use this command and rely on it 🙃

public/openqa-cli.yaml
d3flex: ok. it is not clear what you try to say. I dont understand where is the confusion or what is not clear. 
- by default download only testresults with limits.
- download assets only with an explicit parameter which overrides the size limits. Side-effects will reset size limits for testresults as well.

lib/OpenQA/Client/Archive.pm
d3flex: just copied the answer from another comment.

- by default download only testresults with limits.
- download assets only with an explicit parameter which overrides the size limits. Side-effects will reset size limits for testresults as well.

> it needs to be announced to a wider audience.

I will ask for reviews in the eng-testing, if this is good enough

public/openqa-cli.yaml
kalikiana: That is much clearer. I suggest putting that in the PR description.

public/openqa-cli.yaml
```diff
@@ -163,6 +163,7 @@ subcommands:
     options:
       - asset-size-limit|l=i   --Asset size limit in bytes
       - with-thumbnails|t      --Download thumbnails as well
+      - no-limit|u      --Download all assets without size limits
```
perlpunk: Please align with the other options.

t/43-cli-archive.t
```diff
@@ -15,6 +15,7 @@ use OpenQA::CLI;
 use OpenQA::CLI::archive;
 use OpenQA::Test::Case;
 use OpenQA::Test::TimeLimit '10';
+use OpenQA::Utils qw(:DEFAULT assetdir);
```
perlpunk: Why are you importing everything with `:DEFAULT`? I think you only need `assetdir()`, right?

t/43-cli-archive.t
```diff
@@ -146,4 +147,16 @@ subtest 'Archive job with thumbnails' => sub {
     ok -d $results->child('thumbnails'), 'thumbnails';
 };
 
+subtest 'Archive job with assets' => sub {
+    note('Asset directory: ' . assetdir());
+    my $target = $dir->child('archive-with-assets')->make_path;
+    my ($stdout, $stderr, @result)
+      = capture_stdout sub { $cli->run('archive', @host, '--no-limit', 99963, $target->to_string) };
+    like $stdout, qr{Downloading.*\/iso\/openSUSE-13.1-DVD-x86_64-Build0091-Media.iso},
```
perlpunk:
```suggestion
    like $stdout, qr{Downloading.*/iso/openSUSE-13.1-DVD-x86_64-Build0091-Media.iso},
```

t/43-cli-archive.t
d3flex: yea I just copy paste but forgot to remove it. will fix

public/openqa-cli.yaml
d3flex: > That is much clearer. I suggest putting that in the PR description.

Done

lib/OpenQA/Client/Archive.pm
```diff
@@ -18,7 +18,8 @@ sub run ($self, $options) {
     my $res = $req->res;
     my $default_max_asset_size = 1024 * 1024 * 200;
     $options->{'asset-size-limit'} //= $default_max_asset_size;
-    $self->client->max_response_size($options->{'asset-size-limit'});
+    $options->{'no-limit'} //= 0;
+    $self->client->max_response_size($options->{all} ? 0 : $options->{'asset-size-limit'});
```
Martchus: Where is `$options->{all}` set? It doesn't show up on `openqa-cli archive -h` and it also doesn't look like this PR is introducing it. And wouldn't it make sense to let the user simply specify `--asset-size-limit 0` instead of having this special case?

d3flex: > Where is $options->{all} set?

variable renamed and I missed to update this. fixed.

> And wouldn't it make sense to let the user simply specify --asset-size-limit 0 instead of having this special case?

Actually it makes sense, even if there is nothing wrong having two variables. To me, `--asset-size-limit 0` looks unclear and confusing. Initially I was looking for a name that indicates download of isos and hdds. I went with the suggestions at the end.

lib/OpenQA/Client/Archive.pm
kalikiana: Did this happen? 🤔 I don't see any related comments here

lib/OpenQA/Client/Archive.pm
```diff
@@ -18,7 +18,8 @@ sub run ($self, $options) {
     my $res = $req->res;
     my $default_max_asset_size = 1024 * 1024 * 200;
     $options->{'asset-size-limit'} //= $default_max_asset_size;
-    $self->client->max_response_size($options->{'asset-size-limit'});
+    $options->{'no-limit'} //= 0;
+    $self->client->max_response_size($options->{'no-limit'} ? 0 : $options->{'asset-size-limit'});
```
Martchus, line 22: Ok, so this was really just mean to be `no-limit`. I must still say I don't like the extra flag that makes the code ignore `--asset-size-limit`. It is always better to minimize interactions between flags. So I'd suggest to either:

* Make this at least more explicit.
    1. Call the flag `--no-asset-size-limit` to make it clear that it relates to the `--asset-size-limit`.
    2. Perhaps check whether both arguments are specified and print a hard error in this case (as it makes never sense to specify both of these arguments at the same time).
* Follow my previous suggestion and just allow `--asset-size-limit 0`. The help text could mention the special meaning of `0`. That `0` means "unlimited" is conventional and we also already follow this convention in other places. Mojolicious follows this convention as well hence this way we needed zero code changes to make this actually work.
    * If you don't like `0` you could add code like `$options->{'asset-size-limit'} eq 'infinity' ? 0 : $options->{'asset-size-limit'}` and make this way instead `infinity` a special value.

lib/OpenQA/Client/Archive.pm
Martchus: I also find this misleading. It would make sense to have an additional flag like `$options->{'no-download'}`. If you really want/need to change the default behavior than you could turn the flag around. Or you make it behave like this in case of `--asset-size-limit 0` as suggested by @kalikiana (which could still be combined with my suggestion for `--asset-size-limit infinity`).

Martchus (changes requested):

kalikiana:
I wonder if at this point this should be closed. Or maybe it is worth asking for feedback from a wider audience since there's no consensus on the changes?

okurz:
No reaction for multiple weeks. Closing.

PR #6673, opened 2025-08-22
Do not invoke Mojo::IOLoop->remove twice

d3flex:
Seems that one of the calls can be removed. When `close_connection` is called, jumps into `$self->finish` which does exactly the same.

### DIFF
--- a/lib/OpenQA/Shared/Controller/Running.pm
+++ b/lib/OpenQA/Shared/Controller/Running.pm
@@ -111,9 +111,7 @@ sub streamtext ($self, $file_name) {
     my $close_connection = sub ($self) {
         my $worker_id = $worker->id;
         log_debug "Connection to worker with ID $worker_id will be closed because $logfile changed";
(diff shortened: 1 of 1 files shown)


okurz (changes requested):
Can you add an according test for this change?

d3flex:
> Can you add an according test for this change?

https://app.codecov.io/gh/os-autoinst/openQA/pull/6673/indirect-changes seems like it covers it. Mostly i guess on t/26-controllerrunning.t. what sort of test you would like, if you dont mind.

.readthedocs.yaml
perlpunk, line 1: That seems very unrelated. It is in a separate commit, but this should be in its own PR, with an explanation what this is about.

d3flex: I know. I just refactored it and pushed again

Martchus (approved):

okurz (approved):

PR #6674, opened 2025-08-22
Cleanup unused parameter in streamtext

d3flex:
The parameters introduced in ffdf7fa1191d49aa but since it seems like the design changed and the `start_hook` and `stop_hook` have been deprecated and are not used anymore. `livelog_start` and `livelog_stop` are used as fallback through `_send_livestream_command_to_worker`.


d3flex:
> This might cover certain behaviour that is not fully covered with automatic test. Can you give it a manual verification test to ensure that the text is still streamed in the live view?

image

PR #6679, opened 2025-08-26
MCP: Add support for Bearer token authentication

kraih:
This patch adds bearer token authentication to the existing personal access token mechanism. It is meant to prepare openQA for the addition of MCP support. Almost all MCP clients use bearer token authentication at the moment. Future versions of the MCP specification will most likely require more sophisticated authentication mechanisms.

For clients that use the standard `mcp.json` format, a configuration using bearer token authentication should look something like this:
```
{
  "mcpServers": {
    "openqa": {
      "url": "http://127.0.0.1:9526/experimental/mcp",
      "headers": {
        "Authorization": "Bearer USER:KEY:SECRET"
      }
    }
  }
}
```

More patches with the actual MCP functionality will follow.

### DIFF
--- a/docs/Client.asciidoc
+++ b/docs/Client.asciidoc
@@ -96,9 +96,9 @@ openqa-cli api --host http://openqa.example.com --apikey 1234567890ABCDEF \
 The authentication mechanism used by `openqa-cli` was specifically designed to
 allow secure access to the REST API even via unencrypted HTTP connections. But
 when your openQA server has been deployed with HTTPS (and for HTTP connections
-originating from localhost) you can also use plain old Basic authentication
-with a personal access token. That allows for almost any HTTP client to be used
-with openQA.
+originating from localhost) you can also use plain old Basic or Bearer token
+authentication with a personal access token. That allows for almost any HTTP
+client to be used with openQA.
 
 This access token is made up of your username, and the same key/secret combo
 the `openqa-cli` authentication mechanism uses. All you have to do is combine
@@ -109,6 +109,16 @@ admin REST endpoints (depending on user privileges of course).
 ----
 curl -u arthur:1234567890ABCDEF:ABCDEF1234567890 -X DELETE \
     https://openqa.example.com/api/v1/assets/1
+----
+
+And for HTTP clients that don't support Basic authentication or where the use
+of plain HTTP headers might be more convenient, you can also send the
+personal access token in the form of a Bearer token.
+
+[source,sh]
+----
+curl -H 'Authorization: Bearer arthur:1234567890ABCDEF:ABCDEF1234567890'\
+    -X DELETE https://openqa.example.com/api/v1/assets/1
 ----
 
 == Features
--- a/lib/OpenQA/Shared/Controller/Auth.pm
+++ b/lib/OpenQA/Shared/Controller/Auth.pm
@@ -56,7 +56,10 @@ sub auth ($self) {
     else {
 
         # Personal access token
-        if (my $userinfo = $self->req->url->to_abs->userinfo) {
+        if (($self->req->headers->authorization // '') =~ /^Bearer\s+(.+)$/) {
+            ($user, $reason) = $self->_token_auth($reason, $1);
+        }
+        elsif (my $userinfo = $self->req->url->to_abs->userinfo) {
             ($user, $reason) = $self->_token_auth($reason, $userinfo);
         }
 
--- a/t/api/03-auth.t
+++ b/t/api/03-auth.t
@@ -213,6 +213,28 @@ subtest 'personal access token' => sub {
 
     # Valid access token (again)
     $t->$userinfo('artie:ARTHURKEY01:EXCALIBUR')->delete_ok('/api/v1/assets/1')->status_is(404);
+
+    subtest 'Bearer token' => sub {
+        subtest 'Valid token' => sub {
+            $t->post_ok('/api/v1/feature' => {Authorization => 'Bearer lance:LANCELOTKEY01:MANYPEOPLEKNOW'} => form =>
+                  {version => 100})->status_is(200);
+        };
+
+        subtest 'Invalid username' => sub {
+            $t->post_ok('/api/v1/feature' => {Authorization => 'Bearer invalid:LANCELOTKEY01:MANYPEOPLEKNOW'} => form =>
+                  {version => 100})->status_is(403)->json_is({error => 'invalid personal access token'});
+        };
+
+        subtest 'Invalid key' => sub {
+            $t->post_ok('/api/v1/feature' => {Authorization => 'Bearer lance:LANCELOTKEY02:MANYPEOPLEKNOW'} => form =>
+                  {version => 100})->status_is(403)->json_is({error => 'invalid personal access token'});
+        };
+
+        subtest 'Invalid secret' => sub {
+            $t->post_ok('/api/v1/feature' => {Authorization => 'Bearer lance:LANCELOTKEY01:MANYPEOPLEKNOWS'} => form =>
+                  {version => 100})->status_is(403)->json_is({error => 'invalid personal access token'});
+        };
+    };
 };
 
 subtest 'personal access token (with reverse proxy)' => sub {


docs/Client.asciidoc
```diff
@@ -111,6 +111,16 @@ curl -u arthur:1234567890ABCDEF:ABCDEF1234567890 -X DELETE \
     https://openqa.example.com/api/v1/assets/1
 ----
 
+And for HTTP clients that don't support Basic authentication or where the use
```
d3flex, line 114: Could I suggest `s/And for HTTP clients/Additionally, regarding HTTP clients/`

I just think gives more emphasis to the alternative than a `And...`

lib/OpenQA/Shared/Controller/Auth.pm
```diff
@@ -56,7 +56,10 @@ sub auth ($self) {
     else {
 
         # Personal access token
-        if (my $userinfo = $self->req->url->to_abs->userinfo) {
+        if (($self->req->headers->authorization // '') =~ /^Bearer\s+(.+)$/) {
+            ($user, $reason) = $self->_token_auth($reason, $1);
```
d3flex, line 60: One question: if authentication is achieved here, shouldnt stop `_token_auth` repeating the steps for different token?
It seems that there is a possibility to run twice the ` _token_auth` with the `elsif`

kraih: Please elaborate, i don't understand what you mean.

d3flex: not sure I can but what do you think about the following (not sure if this actually works) but it should make sense

```
my $userinfo = (($self->req->headers->authorization) =~ /^Bearer\s+(.+)$/) ? $1 : $self->req->url->to_abs->userinfo;
($user, $reason) = $self->_token_auth($reason, $userinfo);
```

but not sure if it is more readable, to behonest

kraih: What is the problem you are solving here? I still don't get it.

t/api/03-auth.t
```diff
@@ -213,6 +213,24 @@ subtest 'personal access token' => sub {
...
+        $t->post_ok('/api/v1/feature' => {Authorization => 'Bearer invalid:LANCELOTKEY01:MANYPEOPLEKNOW'} => form =>
+              {version => 100})->status_is(403)->json_is({error => 'invalid personal access token'});
+
+        # Invalid key
+        $t->post_ok('/api/v1/feature' => {Authorization => 'Bearer lance:LANCELOTKEY02:MANYPEOPLEKNOW'} => form =>
+              {version => 100})->status_is(403)->json_is({error => 'invalid personal access token'});
+
+        # Invalid secret
+        $t->post_ok('/api/v1/feature' => {Authorization => 'Bearer lance:LANCELOTKEY01:MANYPEOPLEKNOWS'} => form =>
+              {version => 100})->status_is(403)->json_is({error => 'invalid personal access token'});
```
okurz: Both `status_is` and `json_is` support a test description string as second argument so we should use that instead of a comment

```suggestion
        $t->post_ok('/api/v1/feature' => {Authorization => 'Bearer lance:LANCELOTKEY01:MANYPEOPLEKNOW'} => form =>
              {version => 100})->status_is(200, 'valid token accepted');
        $t->post_ok('/api/v1/feature' => {Authorization => 'Bearer invalid:LANCELOTKEY01:MANYPEOPLEKNOW'} => form =>
              {version => 100})->status_is(403)->json_is({error => 'invalid personal access token'}, undef, 'invalid username rejected');
        $t->post_ok('/api/v1/feature' => {Authorization => 'Bearer lance:LANCELOTKEY02:MANYPEOPLEKNOW'} => form =>
              {version => 100})->status_is(403)->json_is({error => 'invalid personal access token'}, undef, 'invalid key rejected');
        $t->post_ok('/api/v1/feature' => {Authorization => 'Bearer lance:LANCELOTKEY01:MANYPEOPLEKNOWS'} => form =>
              {version => 100})->status_is(403)->json_is({error => 'invalid personal access token'}, undef, 'invalid secret rejected');
```

I can prepare an according change to update the other test calls in this file.

EDIT:  See https://github.com/os-autoinst/openQA/pull/6681

kraih: Those test methods already produce pretty good descriptions by default. If you want to get rid of comments may i suggest turning them into more subtests? Then the TAP output is much more readable.

lib/OpenQA/Shared/Controller/Auth.pm
d3flex: discuss on Slack. I though it could be a small improvement. not a big thing, so I will resolve it

t/api/03-auth.t
kraih: I've updated the PR with subtests, so you can see for yourself, looks pretty nice now.
```
    # Subtest: Bearer token
        # Subtest: Valid token
            ok 1 - POST /api/v1/feature
            ok 2 - 200 OK
            1..2
        ok 1 - Valid token
        # Subtest: Invalid username
            ok 1 - POST /api/v1/feature
            ok 2 - 403 Forbidden
            ok 3 - exact match for JSON Pointer ""
            1..3
        ok 2 - Invalid username
        # Subtest: Invalid key
            ok 1 - POST /api/v1/feature
            ok 2 - 403 Forbidden
            ok 3 - exact match for JSON Pointer ""
            1..3
        ok 3 - Invalid key
        # Subtest: Invalid secret
            ok 1 - POST /api/v1/feature
            ok 2 - 403 Forbidden
            ok 3 - exact match for JSON Pointer ""
            1..3
        ok 4 - Invalid secret
        1..4
    ok 25 - Bearer token
```

lib/OpenQA/Shared/Controller/Auth.pm
perlpunk: > It seems that there is a possibility to run twice the ` _token_auth` with the `elsif`

It's an `elsif`, not an `andif` :)
It will only execute one of both.

t/api/03-auth.t
```diff
@@ -213,6 +213,28 @@ subtest 'personal access token' => sub {
...
+        subtest 'Invalid key' => sub {
+            $t->post_ok('/api/v1/feature' => {Authorization => 'Bearer lance:LANCELOTKEY02:MANYPEOPLEKNOW'} => form =>
+                  {version => 100})->status_is(403)->json_is({error => 'invalid personal access token'});
+        };
+
+        subtest 'Invalid secret' => sub {
+            $t->post_ok('/api/v1/feature' => {Authorization => 'Bearer lance:LANCELOTKEY01:MANYPEOPLEKNOWS'} => form =>
+                  {version => 100})->status_is(403)->json_is({error => 'invalid personal access token'});
+        };
+    };
```
kraih, line 237: Since it has been brought up in a comment before, i really like how the TAP output of these tests looks like now:
```
   # Subtest: Bearer token
        # Subtest: Valid token
            ok 1 - POST /api/v1/feature
            ok 2 - 200 OK
            1..2
        ok 1 - Valid token
        # Subtest: Invalid username
            ok 1 - POST /api/v1/feature
            ok 2 - 403 Forbidden
            ok 3 - exact match for JSON Pointer ""
            1..3
        ok 2 - Invalid username
        # Subtest: Invalid key
            ok 1 - POST /api/v1/feature
            ok 2 - 403 Forbidden
            ok 3 - exact match for JSON Pointer ""
            1..3
        ok 3 - Invalid key
        # Subtest: Invalid secret
            ok 1 - POST /api/v1/feature
            ok 2 - 403 Forbidden
            ok 3 - exact match for JSON Pointer ""
            1..3
        ok 4 - Invalid secret
        1..4
    ok 25 - Bearer token
```

d3flex (approved):
I dont know if you were planning to do the change in the docs/Client.asciidoc but I think this whole thing will be needed, so lets move on and approve

t/api/03-auth.t
okurz: yes, grouping the multiple combined test expressions into subtests can help. I would prefer the more condensed code by using test description strings which your version does not use. I am fine with your version if you can delete the blank lines in between. the separate subtest commands ending with  semicolon should suffice for separation :)

PR #6680, opened 2025-08-26
WIP

Wabri:


t/31-api_descriptions.t
```diff
@@ -2,13 +2,40 @@
...
+my $API_path = path($app_home . '/lib/OpenQA/WebAPI/Controller/API/V1/');
+my $route_child = $app->routes->any('/child')->to(controller => 'Main');
+my $route = $app->routes->any('/*wathever')->add_child($route_child);
+
+make_path($API_path);
+$API_path->child('Main.pm');
+$app->home($app_home);
+my $new_file = "$API_path/Main.pm";
+open my $fh, '>', $new_file or die "open($new_file): $!";
+close $fh;
```
perlpunk:
```suggestion
path($API_path)->child('Main.pm')->touch;
```

t/31-api_descriptions.t
```diff
@@ -2,13 +2,40 @@
...
+
+combined_like {
+    get_pod_from_controllers($app, $route)
+}
+qr/[WARN].*get_pod_from_controllers/, 'Catch warning when file not exists';
```
perlpunk: Either:
```suggestion
qr/\[WARN\].*get_pod_from_controllers/, 'warning when file does not exists';
```
or
```suggestion
qr/WARN.*get_pod_from_controllers/, 'warning when file does not exists';
```

perlpunk: ... because `[WARN]` is a character class and will just match one character that is either W, A, R or N

t/31-api_descriptions.t
```diff
@@ -2,13 +2,40 @@
...
+
 use Mojo::Base 'Mojolicious', -signatures;
 
 use_ok('OpenQA::WebAPI::Description', qw(get_pod_from_controllers set_api_desc));
 my $app = Mojolicious->new;
 get_pod_from_controllers($app);
 
+# mock the parse_file
+my $mock = Test::MockModule->new('Pod::POM');
+$mock->redefine('parse_file', sub { return });
```
okurz: Can this be

```suggestion
$mock->noop('parse_file');
```

t/31-api_descriptions.t
```diff
@@ -2,13 +2,40 @@
...
+
+my $app_home = tempdir();
+my $API_path = path($app_home . '/lib/OpenQA/WebAPI/Controller/API/V1/');
+my $route_child = $app->routes->any('/child')->to(controller => 'Main');
+my $route = $app->routes->any('/*wathever')->add_child($route_child);
+
+make_path($API_path);
+$API_path->child('Main.pm');
```
okurz: Why do you call this and not use the return value? Maybe you wanted to use it further down instead of the manual `open` call?

t/31-api_descriptions.t
Wabri: No it can't, because it gets warning from the `_itemize` function in line 92.

t/31-api_descriptions.t
Wabri: What do you mean?

perlpunk: Well, you were doing
```
my $new_file = "$API_path/Main.pm";
```
below, which is basically the same as `$API_path->child('Main.pm');`

but I commented on that already: https://github.com/os-autoinst/openQA/pull/6680#pullrequestreview-3172960379

t/31-api_descriptions.t
```diff
@@ -2,13 +2,36 @@
...
+my $route = $app->routes->any('/*wathever')->add_child($app->routes->any('/child')->to(controller => 'Main'));
+
+make_path($API_path);
+$API_path->child('Main.pm')->touch;
+$app->home($app_home);
+
+combined_like {
+    get_pod_from_controllers($app, $route)
+}
+qr/\[WARN\].*get_pod_from_controllers/, 'Warning when file does\'t exists';
```
perlpunk:
```suggestion
qr/\[WARN\].*get_pod_from_controllers/, 'Warning when file does not exist';
```
Avoid things like "isn't" or "doesn't"

t/31-api_descriptions.t
Wabri: Got it now thanks tina! With that: yes, you are right oli. I've change the test and I think now It's much better!

t/31-api_descriptions.t
perlpunk: ~~The suggested change works for me~~

perlpunk: Sorry, that isn't true, I ran the wrong test

perlpunk: but this should work:
```suggestion
$mock->redefine(parse_file => undef);
```

t/31-api_descriptions.t
```diff
@@ -2,13 +2,36 @@
...
 
+# mock the parse_file
+my $mock = Test::MockModule->new('Pod::POM');
+$mock->redefine('parse_file', sub { return });
+
+my $app_home = tempdir();
+my $API_path = path($app_home . '/lib/OpenQA/WebAPI/Controller/API/V1/');
+my $route = $app->routes->any('/*wathever')->add_child($app->routes->any('/child')->to(controller => 'Main'));
+
+make_path($API_path);
```
perlpunk:
```suggestion
$API_path->make_path;
```
And remove the `use File::Path qw(make_path);` above

t/31-api_descriptions.t
```diff
@@ -2,13 +2,36 @@
...
+$API_path->child('Main.pm')->touch;
+$app->home($app_home);
+
+combined_like {
+    get_pod_from_controllers($app, $route)
+}
+qr/\[WARN\].*get_pod_from_controllers/, 'Warning when file does not exists';
```
perlpunk:
```suggestion
qr/\[WARN\].*get_pod_from_controllers/, 'Warning when file does not exist';
```

t/31-api_descriptions.t
```diff
@@ -2,13 +2,33 @@
...
+
 use Mojo::Base 'Mojolicious', -signatures;
 
 use_ok('OpenQA::WebAPI::Description', qw(get_pod_from_controllers set_api_desc));
 my $app = Mojolicious->new;
 get_pod_from_controllers($app);
 
+# mock the parse_file
+my $mock = Test::MockModule->new('Pod::POM');
+$mock->redefine('parse_file' => undef);
```
perlpunk: I think we don't run perlcritic on test files yet, otherwise it would have complained that they key is uselessly quoted:
```suggestion
$mock->redefine(parse_file => undef);
```

okurz: @perlpunk can you check if we are missing perlcritic on tests? If we don't have that please create a ticket and/or draft PR for that

perlpunk: There is a ticket: https://progress.opensuse.org/issues/179975

okurz: Oh, I see

perlpunk (approved):

okurz (approved):

PR #6681, opened 2025-08-27
t: Use proper test description strings in 03-auth.t

okurz:


t/api/03-auth.t
```diff
@@ -175,40 +175,24 @@ subtest 'personal access token' => sub {
...
-    # No access token
     my $t = Test::Mojo->new('OpenQA::WebAPI');
-    $t->delete_ok('/api/v1/assets/1')->status_is(403)->json_is({error => 'no api key'});
-
-    # Valid access token
-    $t->$userinfo('artie:ARTHURKEY01:EXCALIBUR')->delete_ok('/api/v1/assets/1')->status_is(404);
-
-    # Valid access token (OpenID user)
+    $t->delete_ok('/api/v1/assets/1')->status_is(403)
+      ->json_is({error => 'no api key'}, undef, 'no access token denied');
```
d3flex: maybe the description can say explicitly `token is required` or something similar. 'no access token denied' is hard to read IMO

okurz: sure, done

PR #6689, opened 2025-09-01
Replace argument assignment with signatures on ObsRsync 

Wabri:

### DIFF
--- a/lib/OpenQA/WebAPI/Plugin/ObsRsync.pm
+++ b/lib/OpenQA/WebAPI/Plugin/ObsRsync.pm
@@ -23,8 +23,7 @@ my $lock_timeout = 360000;
 # non-privileged routes - will be accessible with curl without authentication
 # ensure_operator - privileged routes for using in UI
 # api_ensure_operator - privileged routes for API access
-sub register_common_routes {
-    my ($self, $r, $suffix) = @_;
+sub register_common_routes ($self, $r, $suffix = undef) {
     my $prefix = 'plugin_obs_rsync_';
     $prefix .= $suffix . '_' if $suffix;
 
@@ -35,8 +34,7 @@ sub register_common_routes {
       ->to('Plugin::ObsRsync::Controller::Folders#test_result');
 }
 
-sub register {
-    my ($self, $app, $config) = @_;
+sub register ($self, $app, $config) {
     my $plugin_r = $app->routes->find('ensure_operator');
     my $plugin_api_r = $app->routes->find('api_ensure_operator');
 
@@ -83,8 +81,7 @@ sub register {
         $app->helper('obs_rsync.get_test_result' => \&_get_test_result);
         $app->helper('obs_rsync.get_run_last_info' => \&_get_run_last_info);
         $app->helper(
-            'obs_rsync.get_fail_last_info' => sub {
-                my ($c, $project) = @_;
+            'obs_rsync.get_fail_last_info' => sub ($c, $project) {
                 return _get_last_failed_job($c, $project, 1);
             });
         $app->helper('obs_rsync.get_api_repo' => \&_get_api_repo);
@@ -171,8 +168,7 @@ sub register {
 
 # try to determine whether project is dirty
 # undef means status is unknown
-sub _is_obs_project_status_dirty {
-    my ($self, $url, $project, $repo, $helper) = @_;
+sub _is_obs_project_status_dirty ($self, $url, $project, $repo, $helper) {
     return undef unless $url;
 
     # Use only one UserAgent
@@ -194,8 +190,7 @@ sub _is_obs_project_status_dirty {
     return _parse_obs_response_dirty($res, $repo);
 }
 
-sub _parse_obs_response_dirty {
-    my ($res, $repo) = @_;
+sub _parse_obs_response_dirty ($res, $repo = undef) {
     $repo = 'images' unless $repo;
 
     my $results = $res->dom('result');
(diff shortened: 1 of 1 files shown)


perlpunk (commented):
Small thing about the title: "mojo signatures standard" - there is no such thing.
Signatures are a perl builtin feature since perl 5.20.

You can enable signatures by using
```
use experimental 'signatures';
```
That avoids warnings for versions earlier than 5.38 (where signatures aren't experimental anymore).
However, since we use Mojolicious, several things including strict and warnings and signatures can be enabled in one `use Mojo ...` line, that's why we use it this way.

Once we can rely on perl 5.36, we can just use
```
use v5.36; # enables strict, warnings and signatures
```
to have the same effect.
https://perldoc.perl.org/5.36.0/perldelta#Core-Enhancements

lib/OpenQA/WebAPI/Plugin/ObsRsync.pm
```diff
@@ -147,7 +144,7 @@ sub register {
         $plugin_r->post('/obs_rsync/#alias/run_last')->name('plugin_obs_rsync_forget_run_last')
           ->to('Plugin::ObsRsync::Controller::Folders#forget_run_last');
 
-        $self->register_common_routes($plugin_r);
+        $self->register_common_routes($plugin_r, '');
```
perlpunk: Please revert this

Wabri: If I revert this it will throw error due to unset arg of register_common_routes

perlpunk: yes, but there's a different way to fix it :)
https://perldoc.perl.org/perlsub#Signatures
look for "optional"

Wabri: Oh mh you are right :D

Wabri: I've update in the other pr #6690

lib/OpenQA/WebAPI/Plugin/ObsRsync.pm
```diff
@@ -219,7 +214,7 @@ sub _parse_obs_response_dirty {
 # if input doesn't have '::' character -
 # returned pair is ($project, '')
 sub _split_repo {
-    my (undef, $alias) = @_;
+    my (undef, $alias) = @_; # TODO: confused around the undef as first argument
```
Martchus: If you lookup the usage you'll find this is a helper so the first argument is a controller object. So this should be `_split_repo ($c, $alias) {`. The same counts for the other function. This is then also consistent with `_get_batches` below which you've already handled.

perlpunk: If `$c` isn't used, one can also just use `$` to define a parameter that's not used.
```perl
sub _split_repo ($, $alias) {
```
https://perldoc.perl.org/perlsub#Signatures
Look for "ignored"

perlpunk:
The title is still wrong, btw

okurz (changes requested):
In the subject s/assignation/assignment/

lib/OpenQA/WebAPI/Plugin/ObsRsync.pm
Martchus: I still find it quite nice to use a proper name here. It gives a hint what is passed here and we might use it in the future. Using a proper name will also *not* lead to warnings about unused variables.

lib/OpenQA/WebAPI/Plugin/ObsRsync.pm
```diff
@@ -454,53 +432,44 @@ sub _get_obs_builds_text {
     return join ', ', @builds;
 }
 
-sub _concurrency_guard {
-    my $app = shift->app;
+sub _concurrency_guard ($app) {
```
perlpunk: I'm wondering why that works.
Before the first argument was the controller, so it was like:
```
my ($c) = @_;
my $app = $c->app;
```
Now you are assigning the controller directly to `$app`.
Maybe that works because both the controller and app have access to the helper function? (_concurrency_guard is used as a helper).

The line where that is used is covered, but can you please check that it does the right thing?
If it is covered by tests then it might be enough to add some debugging here and run the according test.

perlpunk (changes requested):

lib/OpenQA/WebAPI/Plugin/ObsRsync.pm
Wabri: I think, since we didn't have proper unit tests for this specific function nor integration test that can cover all the interaction we have, this kind of think can happen and we wont get errors as long as the function in consideration is not actually used.

I've used a find and replace within a vim query and I only check after if the changes made sense. I've fixed the assignation now and I'm going to check the correct assignation during the morning.

lib/OpenQA/WebAPI/Plugin/ObsRsync.pm
```diff
@@ -454,53 +431,44 @@ sub _get_obs_builds_text {
     return join ', ', @builds;
 }
 
-sub _concurrency_guard {
-    my $app = shift->app;
```
perlpunk, line 462: Now that's also wrong.
Before the sub retrieved one argument from the argument list, the controller, and called `->app` on that and assigned it to `$app`.
Now you are assuming that `$app` is given as a second argument.
```suggestion
sub _concurrency_guard ($c) {
    return $c->app->minion->guard('obs_rsync_run_guard', $lock_timeout, {limit => $app->obs_rsync->concurrency});
```

Wabri: That should not work `$app` is never declared, maybe:

```
sub _concurrency_guard ($c) {
    return $c->app->minion->guard('obs_rsync_run_guard', $lock_timeout, {limit => $c->app->obs_rsync->concurrency});
```

Martchus: I would just store app in a local variable like the previous version did. There's no need to change that when introducing use of signatures.

perlpunk (approved):

PR #6690, opened 2025-09-01
Mark unconfigured api route log as uncoverable statement

Wabri:
Related: https://progress.opensuse.org/issues/178948

### DIFF
--- a/lib/OpenQA/WebAPI/Plugin/ObsRsync.pm
+++ b/lib/OpenQA/WebAPI/Plugin/ObsRsync.pm
@@ -44,7 +44,9 @@ sub register {
     die("ssh-keygen is not available. Aborting.\n") unless which('ssh-keygen');
 
     if (!$plugin_r) {
-        $app->log->error('Routes not configured, plugin ObsRsync will be disabled') unless $plugin_r;
+        # uncoverable statement
+        $app->log->error('Routes not configured, plugin ObsRsync will be disabled')
+          unless $plugin_r;
     }
     else {
         $app->helper('obs_rsync.home' => sub { shift->app->config->{obs_rsync}->{home} });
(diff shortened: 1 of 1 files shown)


lib/OpenQA/WebAPI/Plugin/ObsRsync.pm
```diff
@@ -23,8 +23,7 @@ my $lock_timeout = 360000;
 # non-privileged routes - will be accessible with curl without authentication
 # ensure_operator - privileged routes for using in UI
 # api_ensure_operator - privileged routes for API access
-sub register_common_routes {
-    my ($self, $r, $suffix) = @_;
+sub register_common_routes ($self, $r, $suffix = '') {
```
perlpunk: I would rather use `undef` instead, although also the empty string is evaluating to false
```suggestion
sub register_common_routes ($self, $r, $suffix = undef) {
```

t/api/14-plugin_obs_rsync.t
```diff
@@ -9,6 +9,9 @@ use OpenQA::Test::TimeLimit '50';
 use Mojo::IOLoop;
 use OpenQA::Test::Utils 'perform_minion_jobs';
 use OpenQA::Test::ObsRsync 'setup_obs_rsync_test';
+use OpenQA::WebAPI::Plugin::ObsRsync qw(register);
+use File::Temp qw(tempdir);
+use Mojo::File qw(path);
```
Martchus: These additions seem unrelated as no other lines in that file were changed.

Martchus (changes requested):
It doesn't look like this change covers anything. It adds signatures to some functions and marks one line as uncoverable. The commit message should state that. Perhaps these should then be separate commits (like "Use signatures …" and "Mark … as uncoverable").

t/api/14-plugin_obs_rsync.t
Wabri: I forgot to remove the imports when replace the test with the uncoverable statement

perlpunk (approved):

Martchus (approved):
Not sure why you introduced extra wrapping instead of just adding the comments but I guess it is ok.

Wabri:
> introduced extra wrapping
@Martchus I've only added that 2 uncoverable statement comment and run the tidyall and this is what came out

PR #6387, opened 2025-04-12
ci: Ensure clean git status with Test::CheckGitStatus

perlpunk:
Issue: https://progress.opensuse.org/issues/179708

This is using cpanm for now to install the trial version `Test::CheckGitStatus@v0.0.1`

This will currently fail, because there are a couple of tests are creating files without cleaning up.

PR #6692, opened 2025-09-02
Check git status

d3flex:


t/api/04-jobs.t
```diff
@@ -57,6 +58,13 @@ sub calculate_file_md5 ($file) {
     return $md5->hexdigest;
 }
 
+sub ensure_clean_git () {
+    return scope_guard {
+        my $cmd = 'git clean -fd';
```
perlpunk: That's a rather invasive thing to do and I hope this is only temporary for testing :)

perlpunk (commented):
The ideal solution would be that the test does not modify anything in the git directory at all.
That can mean
* Creating a temp dir
* Copy files from git into that dir

That's what at some point every test should do to avoid having any modified/untracked files by tests, so we can also remove existing entries from `.gitignore`.

E.g. see what I did in `t/37-limit_assets.t` or  `t/43-scheduling-and-worker-scalability.t` (which were admittedly pretty easy).

t/api/04-jobs.t
d3flex: for testing but also a potential solution.

perlpunk:
> > E.g. see what I did in `t/37-limit_assets.t` or `t/43-scheduling-and-worker-scalability.t` (which were admittedly pretty easy).
> 
> Despite what you have done, I cant really immediately tell, if we have the same situation. The 04-jobs.t fails on untracked files

The tests I fixed also failed because of untracked files, and I fixed it by making sure the test is in the temp directory, where it can happily create and modify as many files as it needs :)

perlpunk:
And we can work on t/32-openqa_client.t together tomorrow.

perlpunk:
> > And we can work on t/32-openqa_client.t together tomorrow.
> 
> I am also fine if you want to take over. it is yours anyway

Well, you were trying out `scope_guard` which is not a bad idea in general, I just think it can be much easier in this case, and I offered pair programming.

t/lib/OpenQA/Test/Utils.pm
```diff
@@ -623,4 +623,10 @@ sub simulate_load ($load, $from) {
     return $load_avg_file;
 }
 
+sub cleanup_on_exit ($file_path) {
```
perlpunk: That is a nice idea.
But avoiding to write to the git directory in the first place seems to be an easier solution, as we don't have to take care of individual files.

t/32-openqa_client.t
perlpunk, line 1: And for this file the solution is also easy.
```
use Mojo::Util qw(scope_guard);
my $tempdir = tempdir("$FindBin::Script-XXXX", TMPDIR => 1);
chdir $tempdir;
my $guard = scope_guard sub { chdir $FindBin::Bin };
```

t/32-openqa_client.t
```diff
@@ -38,6 +39,9 @@ $fh->syswrite('X');
 undef $fh;
 my $filename = $tempfile->to_string;
 my $sum = OpenQA::File->file_digest($filename);
+my $tempdir = tempdir("$FindBin::Script-XXXX", TMPDIR => 1);
```
perlpunk: Can you move this up before line 26 please?

perlpunk: I mean the whole 3 lines

perlpunk:
Ok. Can you now reorganize your added changes like I suggested in https://github.com/os-autoinst/openQA/pull/6692#pullrequestreview-3180019540 ?
Your changes to the tests were added to my WIP commit which was about enabling Test::CheckGitStatus, and should be separate from the test changes.
After that I think it will be good to merge

d3flex:
> Ok. Can you now reorganize your added changes like I suggested in [#6692 (review)](https://github.com/os-autoinst/openQA/pull/6692#pullrequestreview-3180019540) ? Your changes to the tests were added to my WIP commit which was about enabling Test::CheckGitStatus, which should be separate from the test changes. After that I think it will be good to merge

lets wait for https://github.com/os-autoinst/openQA/pull/6387.

perlpunk:
But adding your changes on top of my updated commits should be an easy and quick thing to do

perlpunk:
The commit message of the third commit now is "Activating the Test::CheckGitStatus module", but it is about your changes to the tests...

perlpunk (approved):

perlpunk:
I set it to "Ready for review" for you, I guess that you just forgot that

Martchus (approved):

tools/ci/run_unit_tests.sh
```diff
@@ -16,6 +16,8 @@ export JUNIT_OUTPUT_FILE=test-results/junit/$target.xml
 # that context in the prove calling context.
 export PERL_TEST_HARNESS_DUMP_TAP=test-results
 export HARNESS='--harness TAP::Harness::JUnit --timer'
+# Activate Test::CheckGitStatus
+export CHECK_GIT_STATUS=1
```
okurz, line 20: why not enable that by default in Makefile? After all the checks are only called if one of the higher level make targets are called which is hardly done by developers unless on purpose, is it?

perlpunk: One thing we should make sure is that the check doesn't run in OBS, because there we have no git (and I think the module is not installed in OBS).
If you have an idea how to add it to the Makefile, can you create a PR?

okurz: Set the default value in Makefile and overwrite it in obs

PR #6693, opened 2025-09-02
Fully Cover OpenQA::WebAPI::Plugin::MemoryLimit

Wabri:

### DIFF
--- a/lib/OpenQA/WebAPI/Plugin/MemoryLimit.pm
+++ b/lib/OpenQA/WebAPI/Plugin/MemoryLimit.pm
@@ -2,23 +2,22 @@
 # SPDX-License-Identifier: GPL-2.0-or-later
 
 package OpenQA::WebAPI::Plugin::MemoryLimit;
-use Mojo::Base 'Mojolicious::Plugin';
+use Mojo::Base 'Mojolicious::Plugin', -signatures;
 
 use BSD::Resource 'getrusage';
 use Mojo::IOLoop;
 
 # Stop prefork workers gracefully once they reach a certain size
-sub register {
-    my ($self, $app) = @_;
-
+sub register ($self, $app, $conf) {
     my $max = $app->config->{global}{max_rss_limit};
+    my $interval = $ENV{OPENQA_RSS_CHECK_INTERVAL} // 5;
     return unless $max && $max > 0;
 
     my $parent = $$;
     Mojo::IOLoop->next_tick(
         sub {
             Mojo::IOLoop->recurring(
-                5 => sub {
+                $interval => sub {
                     my $rss = (getrusage())[2];
                     # RSS is in KB under Linux
                     return unless $rss > $max;
--- a/t/api/14-plugin_memory_limit.t
+++ b/t/api/14-plugin_memory_limit.t
@@ -1,0 +1,26 @@
+# Copyright SUSE LLC
+# SPDX-License-Identifier: GPL-2.0-or-later
+
+use Test::Most;
+use Capture::Tiny qw(capture);
+use Mojolicious;
+use Mojo::IOLoop;
+
+use_ok 'OpenQA::WebAPI::Plugin::MemoryLimit';
+
+my $app = Mojolicious->new;
+$app->config->{global}{max_rss_limit} = 42;
+local $ENV{OPENQA_RSS_CHECK_INTERVAL} = 0;
+
+OpenQA::WebAPI::Plugin::MemoryLimit->register($app, undef);
+
+# change the pid of the test
+local $$ = $$ + 1;
+
+# capture the loop
+my ($out, $err) = capture { Mojo::IOLoop->start };
+
+like $err, qr/Worker exceeded RSS limit "\d+ > 42", restarting/,
+  'Debug error should be captured when value of worker rss is more than max';
+
+done_testing;


t/api/14-plugin_memory_limit.t
```diff
@@ -0,0 +1,12 @@
+# Copyright 2017-2025 SUSE LLC
```
Martchus:
```suggestion
# Copyright SUSE LLC
```

lib/OpenQA/WebAPI/Plugin/MemoryLimit.pm
```diff
@@ -2,15 +2,13 @@
...
+use Mojo::Base 'Mojolicious::Plugin', -signatures;
 
 use BSD::Resource 'getrusage';
 use Mojo::IOLoop;
 
 # Stop prefork workers gracefully once they reach a certain size
-sub register {
-    my ($self, $app) = @_;
-
+sub register ($self, $app, $) {
```
Martchus:
```suggestion
sub register ($self, $app, $conf) {
```

t/api/14-plugin_memory_limit.t
```diff
@@ -0,0 +1,12 @@
...
+# SPDX-License-Identifier: GPL-2.0-or-later
+
+use Test::Most;
+use Mojolicious;
+
+use_ok('OpenQA::WebAPI::Plugin::MemoryLimit');
+
+my $app = Mojolicious->new;
+OpenQA::WebAPI::Plugin::MemoryLimit->register($app);
```
Martchus: This code is still not setting `max_rss_limit` to something > 0 and even if it would the relevant code wouldn't be executed as the test needed to run the event loop.

Wabri: Mh you are right, but I was sure to have it at one point :confused:

lib/OpenQA/WebAPI/Plugin/MemoryLimit.pm
Wabri: It's not even used, so why bother to add to the args list?

Martchus: I just think it is nice to know what the 3rd argument would be.

t/api/14-plugin_memory_limit.t
Martchus: Your new version still doesn't run the event loop. Probably you'll just have to start it and the code under test will then stop it on its own. You probably want to make the hard-coded timer interval configurable so the test doesn't waste 5 seconds.

Martchus: It would also make sense to check whether `Worker exceeded RSS limit` is printed.

Wabri: Yas, you are right... I've try to test it, but I guess I miss some knowledge on https://docs.mojolicious.org/Mojo/IOLoop

Martchus: Well, you've already found the documentation :-)

t/api/14-plugin_memory_limit.t
```diff
@@ -0,0 +1,13 @@
...
+
+use Test::Most;
+use Mojolicious;
+
+use_ok('OpenQA::WebAPI::Plugin::MemoryLimit');
+
+my $app = Mojolicious->new;
+$app->config->{global}{max_rss_limit} = 42;
+OpenQA::WebAPI::Plugin::MemoryLimit->register($app, undef);
+
```
Martchus, line 25:
```suggestion
Mojo::IOLoop->start;

```

I guess you can set the conditions so the code under test will stop the loop automatically. And you should make the interval configurable to avoid wasting 5 seconds until that would happen.

Wabri: It took me more than I thought to make it work, but I think I've handle it now!

lib/OpenQA/WebAPI/Plugin/MemoryLimit.pm
```diff
@@ -2,23 +2,21 @@
...
+use Mojo::Base 'Mojolicious::Plugin', -signatures;
 
 use BSD::Resource 'getrusage';
 use Mojo::IOLoop;
 
 # Stop prefork workers gracefully once they reach a certain size
-sub register {
-    my ($self, $app) = @_;
-
+sub register ($self, $app, $conf, $interval = 5) {
```
perlpunk: I find it a bit unorthodox to add a parameter here which is never used in reality, because that is not the way the register function is called.
Maybe an environment variable could be used instead.
What do others think?

t/api/14-plugin_memory_limit.t
```diff
@@ -0,0 +1,27 @@
...
+use_ok 'OpenQA::WebAPI::Plugin::MemoryLimit';
+
+my $app = Mojolicious->new;
+$app->config->{global}{max_rss_limit} = 42;
+
+OpenQA::WebAPI::Plugin::MemoryLimit->register($app, undef, 0.01);
+
+# change the pid of the test
+local $$ = 42;
```
perlpunk: Why is this necessary?

Wabri: This takes me a while too and the answer is that if not changed the condition in the MemoryLimit package `if $parent ne $$;` inside the `Mojo::IOLoop->recurring` it will be false and the block will not be run.

lib/OpenQA/WebAPI/Plugin/MemoryLimit.pm
Wabri: I've followed @Martchus's comment https://github.com/os-autoinst/openQA/pull/6693#discussion_r2325326111 but I think move the configuration to an environmental variable could be a solution.

t/api/14-plugin_memory_limit.t
```diff
@@ -0,0 +1,27 @@
...
+
+use Test::Most;
+use Capture::Tiny qw(capture);
+use Mojolicious;
+use Mojo::IOLoop;
+
+use_ok 'OpenQA::WebAPI::Plugin::MemoryLimit';
+
+my $app = Mojolicious->new;
+$app->config->{global}{max_rss_limit} = 42;
```
Wabri, line 12: question(non-blocking): does it make sense to extract this value into a support variable in order to recall it also on line 24 in the like test?

Martchus (approved):

PR #6695, opened 2025-09-02
Revert "Allow configuring subdomain for serving logs/assets more securely"

perlpunk:
Reverts os-autoinst/openQA#6668
```
[2025-09-02T17:08:00.452369Z] [error] [wDBjssAQ5_9i] Can't locate object method "via_subdomain" via package "OpenQA::Shared::Controller::Auth" at
/usr/share/openqa/script/../lib/OpenQA/Shared/Controller/Auth.pm line 50.
```


perlpunk:
> You mean a process was restarting while rpm was updating the file system so the process loaded an inconsistent set of modules? This would make sense - even though it would be the first time we run into an issue like this.

yep. but I predicted it a while ago :)

PR #6696, opened 2025-09-03
Prevent abort while doing database interaction in Delete needles

Wabri:
Reference: https://progress.opensuse.org/issues/183077

### DIFF
--- a/lib/OpenQA/Task/Needle/Delete.pm
+++ b/lib/OpenQA/Task/Needle/Delete.pm
@@ -40,6 +40,8 @@ sub _delete_needles ($app, $minion_job, $args) {
         push @{$to_remove{$needle->directory->path}}, $needle;
     }
 
+    $signal_guard->retry(0);
+
     for my $dir (sort keys %to_remove) {
         my $needles = $to_remove{$dir};
         # prevent multiple git tasks to run in parallel


lib/OpenQA/Task/Needle/Delete.pm
```diff
@@ -40,6 +41,7 @@ sub _delete_needles ($app, $minion_job, $args) {
         push @{$to_remove{$needle->directory->path}}, $needle;
     }
 
+    $signal_guard->abort(0);
```
perlpunk: Shouldn't that be `retry`?

lib/OpenQA/Task/Needle/Delete.pm
```diff
@@ -19,6 +19,7 @@ sub _delete_needles ($app, $minion_job, $args) {
     # instead will retry once the gru server returned up and running. The popup
     # on the frontend will wait until the retried job finished.
     my $signal_guard = OpenQA::Task::SignalGuard->new($minion_job);
+    $signal_guard->abort(1);
```
Martchus:
```suggestion
    my $signal_guard = OpenQA::Task::SignalGuard->new($minion_job, abort => 1);
```

lib/OpenQA/Task/Needle/Delete.pm
Martchus: I think it would make sense to abort these jobs at any time so users can just attempt the deletion again manually from the UI. Having to do things manually and being confronted with intermediate errors is of course not nice but it is probably better than just a very long waiting time (or even running into a timeout).

It would probably make sense to set `abort => 1` in the beginning and not change anything else later on.

Of course we could also use `retry => 0` but then the job might take too long to complete and will be killed by systemd.

Martchus (changes requested):

lib/OpenQA/Task/Needle/Delete.pm
perlpunk: But we discussed in the collab session that the Task is doing changes in the database *and* in the filesystem, so when it is aborted, it could end up in an inconsistent state.
One thing we could do is check if the state would actually repair itself. I would try to do that if the ticket was assigned to me, but I guess it's a bit too much for @Wabri

perlpunk: And we also created a followup ticket which would allow us to use callbacks with the singal guard, so we could ensure that one needle is handled atomically in the db and filesystem, but it could abort the loop after each needle.

perlpunk: And the reason for changing it later is, that in the beginning the job is only fetching information from the database, at which point it is totally fine to abort.
But when it starts changing things, it should not abort.

Wabri: For the last comment I've put `abort(0)` once again!

lib/OpenQA/Task/Needle/Delete.pm
```diff
@@ -18,7 +18,7 @@ sub _delete_needles ($app, $minion_job, $args) {
     # SignalGuard will prevent the delete task to interrupt with no recovery,
     # instead will retry once the gru server returned up and running. The popup
     # on the frontend will wait until the retried job finished.
-    my $signal_guard = OpenQA::Task::SignalGuard->new($minion_job);
+    my $signal_guard = OpenQA::Task::SignalGuard->new($minion_job)->abort(1);
```
Martchus: The following is supposed to work as well and should be faster (minimizing the race condition we have here):
```suggestion
    my $signal_guard = OpenQA::Task::SignalGuard->new($minion_job, abort => 1);
```

kalikiana (approved):

lib/OpenQA/Task/Needle/Delete.pm
```diff
@@ -18,7 +18,7 @@ sub _delete_needles ($app, $minion_job, $args) {
     # SignalGuard will prevent the delete task to interrupt with no recovery,
     # instead will retry once the gru server returned up and running. The popup
     # on the frontend will wait until the retried job finished.
-    my $signal_guard = OpenQA::Task::SignalGuard->new($minion_job);
+    my $signal_guard = OpenQA::Task::SignalGuard->new($minion_job, abort => 1);
```
perlpunk: I don't know. I said we should set `retry` to 0 below and leave `abort` alone. In my opinion the value of `abort` is fine.
`retry` is 1 as a default, and `abort` is 0.
So now we would set `abort` to 0 below, but `retry` is still set to 1, and retry is also aborting the job, only that it retries it, so it would still have the same problem.

But @Martchus seems to have a different opinion.

Martchus: @perlpunk Sounds like you looked up how these attributes interact in detail. So I guess I trust your suggestion then :-)

Wabri: @perlpunk I admit that I read your message at least three times, but I'm still not sure what you're saying. From what I understand, you're saying that we agreed to set retry to 0 so that there would be no retries on the database? Remove `abort => 1` mean that the job is re-executed in the event of a sigint/term, which is fine in the first part, but not when we modify the database which once it has captured a sigint/term should not re-execute.

Should be this:
```suggestion
    my $signal_guard = OpenQA::Task::SignalGuard->new($minion_job);
```

And after populate the `to_remove` array:
```
    $signal_guard->retry(0);
```
?

perlpunk: @Wabri Yes, that's what I would suggest

perlpunk (approved):

Wabri:
Opened https://progress.opensuse.org/issues/188982

PR #6697, opened 2025-09-03
Fix syntax error in nginx config

nicksinger:
This should fix https://progress.opensuse.org/issues/188208 and the [currently failing openqa-in-openqa tests](https://openqa.opensuse.org/tests/5284271/#step/openqa_webui/15).


perlpunk (approved):

Martchus:
I was just going to create a PR after running into this myself when updating the config of o3.

PR #6698, opened 2025-09-03
Fully cover WebAPI::Plugin::ObsRsync::Controller::Folders

Wabri:

### DIFF
--- a/lib/OpenQA/WebAPI/Plugin/ObsRsync/Controller/Folders.pm
+++ b/lib/OpenQA/WebAPI/Plugin/ObsRsync/Controller/Folders.pm
@@ -2,14 +2,13 @@
 # SPDX-License-Identifier: GPL-2.0-or-later
 
 package OpenQA::WebAPI::Plugin::ObsRsync::Controller::Folders;
-use Mojo::Base 'Mojolicious::Controller';
+use Mojo::Base 'Mojolicious::Controller', -signatures;
 use Mojo::File;
 use POSIX 'strftime';
 
 my $non_project_folders = '^(t|profiles|script|xml)$';
 
-sub list {
-    my $self = shift;
+sub list ($self) {
     my $helper = $self->obs_rsync;
     my $folders = Mojo::File->new($helper->home)->list({dir => 1})->grep(sub { -d $_ })->map('basename')->to_array;
     my @res;
@@ -26,8 +25,7 @@ sub list {
     return $self->render(json => \@res, status => 200);
 }
 
-sub index {
-    my ($self, $obs_project, $folders) = @_;
+sub index ($self, $obs_project, $folders) {
     my $helper = $self->obs_rsync;
     my %folder_info_by_name;
     if (!$folders) {
@@ -85,8 +83,7 @@ sub index {
     $self->render('ObsRsync_index', folder_info_by_name => \%folder_info_by_name, project => $obs_project);
 }
 
-sub folder {
-    my $self = shift;
+sub folder ($self) {
     my $alias = $self->param('alias');
     my $helper = $self->obs_rsync;
     return undef if $helper->check_and_render_error($alias);
@@ -113,8 +110,7 @@ sub folder {
         openqa_sh => $files->grep(qr/print_openqa\.sh/)->join());
 }
 
-sub runs {
-    my $self = shift;
+sub runs ($self) {
     my $alias = $self->param('alias');
     my $helper = $self->obs_rsync;
     return undef if $helper->check_and_render_error($alias);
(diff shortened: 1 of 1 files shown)


lib/OpenQA/WebAPI/Plugin/ObsRsync/Controller/Folders.pm
```diff
@@ -189,6 +189,7 @@ sub forget_run_last {
     if (unlink($dest)) {
         return $self->render(json => {message => 'success'}, status => 200);
     }
+    # uncoverable statement
     return $self->render(json => {message => "error $!"}, status => 500);
```
Martchus: It is probably possible to write a test covering this line.

Wabri: Probably, I just follow the suggestion by tina to mark most of them uncoverable, but I'll work on testing it if you want :+1:

Martchus: I would rather just leave it uncovered if writing a test exceeds the effort we currently want to spend here instead of marking the line wrongly as uncoverable. This will just fake the statistics making it harder to achieve true 100 % test coverage.

Wabri: Giving this:

```
❯ grep "uncoverable statement" -R | wc -l
347
```

Question: Do we want to open a ticket to cover all the uncoverable statement we already have?

Thought: In order to make `unlink($dest)` fail, I need to not have a lock for `$dest`. To have that the first think that came up to my mind is a mock of `unlink`, but I'm not sure if it's possible to do since is a buildin function. Maybe create a symlink could be the way, but I'm not sure... I'm going to try the second way.

okurz: In general mocking also builtin functions is possible. If not feasible then it can be easier to extract a function in the same module to override. We have examples in the existing codebase os-autoinst+openQA for both, overriding builtin functions as well as the second approach

Wabri: > In general mocking also builtin functions is possible.

I remember that I've spent a lot of time doing it without succeed, but I can try it once again. 

>  If not feasible then it can be easier to extract a function in the same module to override.

I've done it twice, maybe more on os-autoinst, but remember a discussion during a collab around it that came up with the fact that we should avoid it. That's why I didn't even propose that way.

okurz (changes requested):
In your commit message s/assignation/assignment/

lib/OpenQA/WebAPI/Plugin/ObsRsync/Controller/Folders.pm
```diff
@@ -189,11 +181,11 @@ sub forget_run_last {
...
```
okurz: I would just write it like this

```suggestion
return $self->render(json => {message => "error $!"}, status => 500) unless unlink($dest);
return $self->render(json => {message => 'success'}, status => 200);
```

or maybe even cleaner

```suggestion
unlink($dest) or return $self->render(json => {message => "error $!"}, status => 500);
return $self->render(json => {message => 'success'}, status => 200);
```


that's regardless if you want to write a test for it or not. I think it's just cleaner to show the expected happy path flow with error handling for the unexpected case. Here you wouldn't need a comment for the "uncovered statement" although branch coverage might still be incomplete.

lib/OpenQA/WebAPI/Plugin/ObsRsync/Controller/Folders.pm
```diff
@@ -186,14 +178,11 @@ sub forget_run_last {
     my $project_lock = Mojo::File->new($helper->home, $project, 'rsync.lock');
     -f $project_lock and return $self->render(json => {message => 'Project lock exists'}, status => 423);
 
-    if (unlink($dest)) {
-        return $self->render(json => {message => 'success'}, status => 200);
-    }
-    return $self->render(json => {message => "error $!"}, status => 500);
+    unlink($dest) or return $self->render(json => {message => "error $!"}, status => 500);
```
perlpunk, line 181: For the future:
```perl
my $MOCK = 0;
sub mocked_unlink {
    return CORE::unlink(@_) unless $MOCK;
    # mock something
}
BEGIN {
    *CORE::GLOBAL::unlink = \&mocked_unlink;
}

...
$MOCK = 1;
# test
```

perlpunk (approved):

lib/OpenQA/WebAPI/Plugin/ObsRsync/Controller/Folders.pm
Wabri: Thanks for the tips :)

PR #6699, opened 2025-09-03
Ensure destination dir for asset downloads exists when cloning jobs

Martchus:
* Avoid running into errors like `can't write /hdd/openqa-devel/openqa/share/factory/hdd` when this directory just doesn't exist; try to create the directory instead
* Keep the `-w $dst_dir` check so one still gets the usual error message when trying to download assets under e.g. `/var/lib/openqa` with a user that cannot write there
* See https://progress.opensuse.org/issues/184909

---

I tested this locally by running `openqa-clone-job` with different users and different `--dir` parameters.

### DIFF
--- a/lib/OpenQA/Script/CloneJob.pm
+++ b/lib/OpenQA/Script/CloneJob.pm
@@ -158,7 +158,7 @@ sub clone_job_download_assets ($jobid, $job, $url_handler, $options) {
             $from->path(sprintf '/tests/%d/asset/%s/%s', $jobid, $type, $file);
             $from = $from->to_string;
 
-            die "can't write $dst_dir\n" unless -w $dst_dir;
+            die "Cannot write $dst_dir\n" unless -w $dst_dir;
 
             print STDERR "downloading\n$from\nto\n$dst\n";
             my $r = $ua->mirror($from, $dst);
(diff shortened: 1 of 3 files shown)


lib/OpenQA/Script/CloneJob.pm
```diff
@@ -152,12 +152,13 @@ sub clone_job_download_assets ($jobid, $job, $url_handler, $options) {
...
             $dst =~ s,.*/,,;
-            $dst = join('/', $options->{dir}, $type, $dst);
+            my $dst_dir = path($options->{dir}, $type)->make_path;
+            $dst = $dst_dir->child($dst)->to_string;
             my $from = $remote_url->clone;
             $from->path(sprintf '/tests/%d/asset/%s/%s', $jobid, $type, $file);
             $from = $from->to_string;
 
-            die "can't write $options->{dir}/$type\n" unless -w "$options->{dir}/$type";
+            die "can't write $dst_dir\n" unless -w $dst_dir;
```
perlpunk: Maybe we can get rid of the contraction here:

```suggestion
            die "Cannot write $dst_dir\n" unless -w $dst_dir;
```
otherwise LGTM

perlpunk (approved):

PR #6703, opened 2025-09-04
Use autoyast ERB helper to detect disk dynamically

gauravpathak:
Related to: https://progress.opensuse.org/issues/186948

This commit introduces an ERB helper template to dynamically generate the AutoYaST profile during OS installation.

Key changes include:
- Automatically selects the smallest disk for OS installation
- Allocates remaining disk(s) for openQA data storage
- Creates a RAID array for openQA data if more than two disks are available

### DIFF
--- a/contrib/ay-openqa-worker.xml.erb
+++ b/contrib/ay-openqa-worker.xml.erb
@@ -69,6 +69,8 @@
   </timezone>
   <partitioning config:type="list">
     <drive>
+      <% disk = disks.sort_by { |d| d[:size] }.first %> <%# find the [smallest] disk for main OS installation %>
+      <device><%= disk[:udev_names][0] %></device> <%# print the disk device name %>
       <initialize config:type="boolean">true</initialize>
       <partitions config:type="list">
         <partition>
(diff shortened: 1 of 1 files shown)


contrib/ay-openqa-worker.xml.erb
```diff
@@ -81,7 +83,60 @@
...
+      <type t="symbol">CT_MD</type>
+      <device>/dev/md/openqa</device>
+      <partitions config:type="list">
+        <partition>
+          <mount>/var</mount>
+          <size>max</size>
+        </partition>
+      </partitions>
+      <raid_options>
+        <raid_type>raid0</raid_type>
```
nicksinger, line 114: do you remember why we have RAID1 in the current AY-profiles and not RAID0 any longer?

contrib/ay-openqa-worker.xml.erb
```diff
@@ -81,7 +83,60 @@
           <size>auto</size>
         </partition>
       </partitions>
+      <use>all</use>
     </drive>
+    <% raidarray = disks.sort_by { |d| d[:size] }.slice(1..-1) %> <%# remove first [smallest] disk from array %>
+    <% if raidarray.length > 1 %> <%# if there are more than two disks on the system, combine them in raid array%>
```
nicksinger:
```suggestion
    <% if raidarray.length > 1 %> <%# if there is more than one disk remaining on the system, combine them into a raid array %>
```

Martchus (commented):
> Automatically selects the smallest disk for OS installation

But hopefully not a randomly plugged USB device or the ESP.

I guess it looks generally good, though. How did you test it?

gauravpathak:
> But hopefully not a randomly plugged USB device or the ESP.
> 
I am afraid it's **not** right now, filtering out plugged in USB calls for some additional ruby code. It is doable, but not sure how complex it will be.
What's ESP? Some on-board "Embedded Storage" thingy?

> I guess it looks generally good, though. How did you test it?

I tested it by installing OS on `worker37`.

Martchus:
I guess we don't need to care much about e.g. USB drives - it just came to mind :-)

For ESP, see https://en.wikipedia.org/wiki/EFI_system_partition.

contrib/ay-openqa-worker.xml.erb
gauravpathak: I think, @okurz can give some insights here.

We saw `raid1` is used here https://github.com/os-autoinst/openQA/blob/master/contrib/ay-openqa-worker-leap-3-nvme.xml#L94 and got confused. @nicksinger couldn't recall why we switched to `raid1`.

okurz: Isn't https://github.com/os-autoinst/openQA/blob/ec726f9c5dd194ebc31c8601d42b1c2d50a17260/contrib/ay-openqa-worker-leap-3-nvme.xml#L94 used for the root partition? AFAIR RAID0 for all devices attached to /var/lib/openqa is a good choice, even if it's just one device

gauravpathak: Got it, that’s for the root partition. Our main question was really about whether to go with RAID 0 or RAID 1 for openqa data and cache. Thanks for clearing that up!

contrib/ay-openqa-worker.xml.erb
```diff
@@ -81,7 +83,60 @@
...
+      </partitions>
+      <use>all</use>
+    </drive>
+    <% end %> <%# end for loop %>
+    <drive>
+      <type t="symbol">CT_MD</type>
+      <device>/dev/md/openqa</device>
+      <partitions config:type="list">
+        <partition>
+          <mount>/var</mount>
```
gauravpathak: @okurz @nicksinger 
Should this be changed to `/var/lib/openqa`? Just like how it's done at [line 134](https://github.com/os-autoinst/openQA/pull/6703/files#diff-305e1a4949974dba9f27cdcc4524eb07bdd899c1370c3eab2aa5762000eaf97cR134)

Martchus: Yes, it should probably be changed.

nicksinger: Yes, keep it similar to the existing profile (I guess using `/var` might even conflict with our default subvolume configuration)

PR #6704, opened 2025-09-04
Allow specifying a full domain via `file_security_policy`

Martchus:
With this, files can be served from any additional domain and not just from domains with some additional prefix. This of course means that one now also has to specify the full domain always. However, that probably not a big
deal and we have other settings such as `download_domains` which are already similar.

This replaces the previous `subdomain:…` syntax to avoid providing too many options and having too complex code.

Related ticket: https://progress.opensuse.org/issues/184456

### DIFF
--- a/lib/OpenQA/Setup.pm
+++ b/lib/OpenQA/Setup.pm
@@ -308,8 +308,8 @@ sub _validate_worker_timeout ($app) {
 }
 
 sub _validate_security_policy ($app, $config) {
-    if ($config->{file_security_policy} =~ m/^(download-prompt|insecure-browsing|subdomain:(.+))$/) {
-        if (defined(my $subdomain = $2)) { $config->{file_subdomain} = $subdomain }
+    if ($config->{file_security_policy} =~ m/^(download-prompt|insecure-browsing|domain:(.+))$/) {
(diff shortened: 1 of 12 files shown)


lib/OpenQA/Setup.pm
```diff
@@ -309,7 +309,7 @@ sub _validate_worker_timeout ($app) {
 
 sub _validate_security_policy ($app, $config) {
     if ($config->{file_security_policy} =~ m/^(download-prompt|insecure-browsing|subdomain:(.+))$/) {
-        if (defined(my $subdomain = $2)) { $config->{file_subdomain} = "$subdomain." }
+        if (defined(my $subdomain = $2)) { $config->{file_subdomain} = $subdomain }
```
kalikiana: I guess that's annoying now, but considering the option was just introduced... now might be the best chance to actually make it `file_domain` and allow any full domain 😶

Martchus: I guess that would make sense.

Martchus: But I'll replace the current approach with that. As you said, the option was just introduced. So we can still break it and for the sake of simplicity make it always the full domain.

kalikiana (approved):

d3flex (approved):

PR #6705, opened 2025-09-04
parser: ktap: Don't add skipped subtests to output

rbmarliere:
Verification runs:
- Before: https://rmarliere-openqa.qe.prg2.suse.org/tests/1555
- After: https://rmarliere-openqa.qe.prg2.suse.org/tests/1556

### DIFF
--- a/lib/OpenQA/Parser/Format/KTAP.pm
+++ b/lib/OpenQA/Parser/Format/KTAP.pm
@@ -70,6 +70,7 @@ sub _parse_subtest ($self, $result) {
         $steps->{result} = 'softfail';
     }
 
+    return if $line =~ /#\s*SKIP\b/i;
     $self->_add_output({file => $filename, content => $line});
     $self->state->{m} = $m + 1;
 }


lib/OpenQA/Parser/Format/KTAP.pm
```diff
@@ -70,6 +70,9 @@ sub _parse_subtest ($self, $result) {
         $steps->{result} = 'softfail';
     }
 
+    if ($line =~ /#\s*SKIP\b/i) {
+        return;
+    }
```
Martchus: To fool the otherwise failing check for test coverage:
```suggestion
    return if $line =~ /#\s*SKIP\b/i;
```

rbmarliere: Thank you, it seems to have resolved in https://github.com/os-autoinst/openQA/pull/6705/commits/d222601a2998902a4aeb9a7fc96183c3ea54781c :)

PR #6707, opened 2025-09-04
Worker: Detect IPMI config and shutdown SUT when unused

aylen384:
Ticket: https://progress.opensuse.org/issues/185992

### DIFF
--- a/lib/OpenQA/Worker.pm
+++ b/lib/OpenQA/Worker.pm
@@ -3,6 +3,7 @@
 
 package OpenQA::Worker;
 use Mojo::Base -base, -signatures;
+use IPC::Run ();
 
 BEGIN {
     use Socket;
@@ -350,7 +351,33 @@ sub init ($self) {
             });
     }
 
+    my $interval = $global_settings->{IPMI_AUTOSHUTDOWN_INTERVAL} // 300;
+    if (   $global_settings->{IPMI_HOSTNAME}
+        && $global_settings->{IPMI_USER}
+        && $global_settings->{IPMI_PASSWORD}
+        && $interval ne '0')
+    {
+        log_info 'IPMI config present -> will periodically make sure SUT is powered off when unused.';
+        Mojo::IOLoop->recurring($interval => sub { $self->shutdown_ipmi_sut() });
+    }
+
     return \$return_code;
+}
+
+sub shutdown_ipmi_sut ($self) {
+    return if $self->current_job;
+    my $sgs = $self->settings->global_settings;
+    my $ipmi_host = $sgs->{IPMI_HOSTNAME};
+    my $ipmi_user = $sgs->{IPMI_USER};
+    my $ipmi_pass = $sgs->{IPMI_PASSWORD};
+    my $ipmi_opts = $sgs->{IPMI_OPTIONS} // '-I lanplus';
+    my @cmd = (
+        'ipmitool', split(' ', $ipmi_opts),
+        '-H', $ipmi_host, '-U', $ipmi_user, '-P', $ipmi_pass, 'chassis', 'power', 'off'
+    );
+    my $ret = IPC::Run::run(\@cmd, \my $stdin, \my $stdout, \my $stderr);
+    chomp $stderr;
+    log_warning(join(' ', map { $_ eq $ipmi_pass ? '[masked]' : $_ } @cmd) . ": $stderr") unless $ret;
 }
 
 sub configure_cache_client ($self) {
--- a/t/24-worker-overall.t
+++ b/t/24-worker-overall.t
@@ -961,4 +961,31 @@ subtest 'storing package list' => sub {
     qr/doesn't return any data/, 'log message about no data';
 };
 
+subtest 'worker ipmi' => sub {
+    my $ioloop_mock = Test::MockModule->new('Mojo::IOLoop');
+    my $sched = 0;
+    $ioloop_mock->redefine(recurring => sub { $sched = 1; });
+    my $global_settings = $settings->global_settings;
+    delete $global_settings->{LOG_DIR};
+    $global_settings->{IPMI_HOSTNAME} = "ipmi.host";
+    $global_settings->{IPMI_USER} = "username";
+    $global_settings->{IPMI_PASSWORD} = "password";
+    $global_settings->{IPMI_AUTOSHUTDOWN_INTERVAL} = "1";
+    combined_like { $worker->init }
+    qr/IPMI config present/, 'IPMI config detected';
+    ok $sched, 'ipmi job scheduled';
+    $global_settings->{IPMI_AUTOSHUTDOWN_INTERVAL} = '0';
+    $ioloop_mock->unmock('recurring');
+
+    $worker->current_job(undef);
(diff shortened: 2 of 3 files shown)


lib/OpenQA/Worker.pm
```diff
@@ -350,9 +351,32 @@ sub init ($self) {
...
+    my $ipmi_host = $sgs->{IPMI_HOSTNAME};
+    my $ipmi_user = $sgs->{IPMI_USER};
+    my $ipmi_pass = $sgs->{IPMI_PASSWORD};
+    my $ipmi_opts = $sgs->{IPMI_OPTIONS} // '-I lanplus';
+    my @cmd = (
+        'ipmitool', split(' ', $ipmi_opts),
+        '-H', $ipmi_host, '-U', $ipmi_user, '-P', $ipmi_pass, 'chassis', 'power', 'off'
+    );
+    my ($stdin, $stdout, $stderr, $ret);
+    $ret = IPC::Run::run(\@cmd, \$stdin, \$stdout, \$stderr);
```
perlpunk:
```suggestion
    my $ret = IPC::Run::run(\@cmd, \my $stdin, \my $stdout, \my $stderr);
```

lib/OpenQA/Worker.pm
```diff
@@ -350,9 +351,35 @@ sub init ($self) {
             });
     }
 
+    if (   $global_settings->{IPMI_HOSTNAME}
+        && $global_settings->{IPMI_USER}
+        && $global_settings->{IPMI_PASSWORD}
+        && $global_settings->{IPMI_AUTOSHUTDOWN} ne '0')
```
Martchus:
```suggestion
        && $global_settings->{IPMI_AUTOSHUTDOWN_INTERVAL} ne '0')
```

lib/OpenQA/Worker.pm
```diff
@@ -350,9 +351,35 @@ sub init ($self) {
...
+    {
+        log_info 'IPMI config present -> will periodically make sure SUT is powered off when unused.';
```
d3flex, line 360: Just a personal opipion. that should be better a debug log

aylen384: It's only printed once so better let the user know.
Also it's too important for debug.

t/24-worker-overall.t
```diff
@@ -961,4 +961,31 @@ subtest 'storing package list' => sub {
...
+    combined_like { $worker->init }
+    qr/IPMI config present/, 'IPMI config detected';
+    ok $sched, 'ipmi job scheduled';
+    $global_settings->{IPMI_AUTOSHUTDOWN_INTERVAL} = '0';
+    $ioloop_mock->unmock('recurring');
+
+    $worker->current_job(undef);
+    my $ipc_mock = Test::MockModule->new('IPC::Run');
+    my $c = [];
+    $ipc_mock->redefine('run' => sub { $c = shift; shift; shift; ${(shift)} = ''; return 1; });
```
okurz: Does it work to use signatures here instead of the quadruple shift?

lib/OpenQA/Worker.pm
```diff
@@ -350,9 +351,35 @@ sub init ($self) {
             });
     }
 
+    if (   $global_settings->{IPMI_HOSTNAME}
+        && $global_settings->{IPMI_USER}
+        && $global_settings->{IPMI_PASSWORD}
+        && $global_settings->{IPMI_AUTOSHUTDOWN_INTERVAL} ne '0')
```
okurz: I think you should document those settings. Probably best to do in /etc/openqa/workers.ini

okurz:
> > Going that deep and backend specific in the openQA worker is questionable. … calling ipmitool which in before was reserved for the os-autoinst backend.
> 
> We already do something similar with the checks whether QEMU is still running. Some things make more sense on the worker level as the worker is a long-running service. Adding this kind of code in os-autoinst makes not much sense because when the test fails due to a broken IPMI connection we cannot expect to be able to instantly shutdown the machine via IPMI.

Not immediately but after reasonable waiting time with retries.

Martchus:
> Going that deep and backend specific in the openQA worker is questionable.

Where are we going *deep* here, btw? This code is just reading test variables (like other worker code already does; we sometimes even write test variables) and invoking the generic `ipmitool` without relying on knowledge of the innerworkings of the backend.

Martchus:
I guess it wouldn't hurt to go though a layer of isotovideo code here.

aylen384:
> I guess it wouldn't hurt to go though a layer of isotovideo code here.

I'm strongly against this as it would add way too much overhead.
We do this with qemu so we can do it as well with ipmitool.
Also the work is already done.

t/24-worker-overall.t
```diff
@@ -961,4 +961,31 @@ subtest 'storing package list' => sub {
...
+subtest 'worker ipmi' => sub {
+    my $ioloop_mock = Test::MockModule->new('Mojo::IOLoop');
+    my $sched = 0;
+    $ioloop_mock->redefine(recurring => sub { $sched = 1; });
+    my $global_settings = $settings->global_settings;
+    delete $global_settings->{LOG_DIR};
+    $global_settings->{IPMI_HOSTNAME} = "ipmi.host";
+    $global_settings->{IPMI_USER} = "username";
+    $global_settings->{IPMI_PASSWORD} = "password";
+    $global_settings->{IPMI_AUTOSHUTDOWN_INTERVAL} = "1";
```
d3flex, line 973: Maybe this can be assigned to a smaller value.

lib/OpenQA/Worker.pm
Martchus: This will print a warning if `IPMI_AUTOSHUTDOWN_INTERVAL` is undef.

You should probably define `$interval` outside of the condition and use that here as well.

t/24-worker-overall.t
Martchus: It doesn't matter because the recurring function is mocked here anyway.

lib/OpenQA/Worker.pm
```diff
@@ -350,9 +351,35 @@ sub init ($self) {
...
+    return if $self->current_job;
+    my $sgs = $self->settings->global_settings;
+    my $ipmi_host = $sgs->{IPMI_HOSTNAME};
+    my $ipmi_user = $sgs->{IPMI_USER};
+    my $ipmi_pass = $sgs->{IPMI_PASSWORD};
+    my $ipmi_opts = $sgs->{IPMI_OPTIONS} // '-I lanplus';
+    my @cmd = (
+        'ipmitool', split(' ', $ipmi_opts),
+        '-H', $ipmi_host, '-U', $ipmi_user, '-P', $ipmi_pass, 'chassis', 'power', 'off'
+    );
```
kalikiana, line 377: Reading `IPMI_HOSTNAME` and friends here feels a bit fragile. These are implemented in a backend.

To my mind it would be cleaner to make this part of testapi and implement it in the backend.

okurz (approved):
Considering the last assessment of @Martchus as we agreed approving

kalikiana (approved):

PR #6708, opened 2025-09-04
Add perl(MCP) dependency in preparation for AI support

okurz:


dist/rpm/openQA.spec
```diff
@@ -215,6 +217,13 @@ Requires(pre):  group(kvm)
 %description worker
 The openQA worker manages test engine (provided by os-autoinst package).
 
+%package mcp
+Summary:        Additional MCP package for AI support in openQA
+Requires:       %{mcp_requires}
+
+%description mcp
+This package contains additional resources for AI support in openQA.
```
Martchus, line 225: If we merge the actual plugin we can rewrite this:
```suggestion
This package contains a plugin for AI support in openQA.
```

Martchus (approved):

dist/rpm/openQA.spec
okurz: Good point. I included that in https://github.com/os-autoinst/openQA/pull/6711

perlpunk (approved):

PR #6685, opened 2025-08-29
MCP: Add MCP support as an optional feature

kraih:
Followup to #6679, which should be merged first.

This patch adds MCP support with three tools as an optional feature that needs to be enabled in the `openqa.ini`:
```
[global]
mcp_enabled = read-only
```

For starters i have added three read-only tools. The first (`openqa_get_info`)
provides information about the openQA server and the current user. The second
(`openqa_get_job_info`) information about a specific job. And the third
(`openqa_get_log_file`) returns the content of a job specific log file. The
returned information is formatted with templates and very easy to modify.
Current gen LLMs prefer plain text information, but that can of course change
as the technology evolves.

Example conversation with `openqa_get_info` tool:
```
You
Who am i?

gemma-3-12b-it
openqa_get_info({})
Arguments: {}
Result
[...]

You are a demo user in openQA with ID 1092. You have both administrator and operator privileges.
```

Example conversation with `openqa_get_job_info` tool:
```
You
What can you tell me about openQA job 5266406

gemma-3-12b-it
openqa_get_job_info({"job_id":5266406}…})
Arguments: {"job_id":5266406}
Result
[...]

OpenQA job 5266406 is a test of the
opensuse-Tumbleweed-JeOS-for-kvm-and-xen-sdboot-aarch64 build, specifically
targeting an ARM architecture. It ran on August 26, 2025, and ultimately
failed. The failure appears to be related to verify_setup, with a bug
reference (poo#184444) indicating the system booted into an emergency shell.
Several other tests also failed including image_checks, record_machine_id,
and force_scheduled_tasks. The job used qemu as its backend and was
configured with several specific settings related to disk size,
repositories, and UEFI boot.

You
Is this a known error?

gemma-3-12b-it
Yes, according to the comments associated with the job result, it's a
known issue referenced by bug poo#184444. The comment explicitly states
"test fails in verify_setup - system boots into emergency shell" and
links to the bug report on progress.opensuse.org. It also notes that
this is a carryover from a previous test (t#5191155).
```

Of course this is just a starting point to allow openQA users to experiment with LLM driven openQA workflows. And i expect this feature to change a lot before it leaves experimental status.

Progress: https://progress.opensuse.org/issues/184798

PR #6711, opened 2025-09-04
Add MCP support as an optional feature

okurz:
This is a recreated pull request based on
* #6685

as well as
* #6708

which brings in the dependency but as a non-problematic sub-package so that tests not relying on MCP can continue unchanged.

This patch adds MCP support with three tools as an optional feature that
needs to be enabled in the `openqa.ini`:
```
[global]
mcp_enabled = read-only
```

For starters i have added three read-only tools. The first (`openqa_get_info`)
provides information about the openQA server and the current user. The second
(`openqa_get_job_info`) information about a specific job. And the third
(`openqa_get_log_file`) returns the content of a job specific log file. The
returned information is formatted with templates and very easy to modify.
Current gen LLMs prefer plain text information, but that can of course change
as the technology evolves.

Example conversation with `openqa_get_info` tool:
```
You
Who am i?

gemma-3-12b-it
openqa_get_info({})
Arguments: {}
Result
[...]

You are a demo user in openQA with ID 1092. You have both administrator and operator privileges.
```

Example conversation with `openqa_get_job_info` tool:
```
You
What can you tell me about openQA job 5266406?

gemma-3-12b-it
openqa_get_job_info({"job_id":5266406}…})
Arguments: {"job_id":5266406}
Result
[...]

OpenQA job 5266406 is a test of the
opensuse-Tumbleweed-JeOS-for-kvm-and-xen-sdboot-aarch64 build, specifically
targeting an ARM architecture. It ran on August 26, 2025, and ultimately
failed. The failure appears to be related to verify_setup, with a bug
reference (poo#184444) indicating the system booted into an emergency shell.
Several other tests also failed including image_checks, record_machine_id,
and force_scheduled_tasks. The job used qemu as its backend and was
configured with several specific settings related to disk size,
repositories, and UEFI boot.

You
Is this a known error?

gemma-3-12b-it
Yes, according to the comments associated with the job result, it's a
known issue referenced by bug poo#184444. The comment explicitly states
"test fails in verify_setup - system boots into emergency shell" and
links to the bug report on progress.opensuse.org. It also notes that
this is a carryover from a previous test (t#5191155).
```

Of course this is just a starting point to allow openQA users to
experiment with LLM driven openQA workflows. And i expect this feature
to change a lot before it leaves experimental status.

### DIFF
--- a/lib/OpenQA/Schema/ResultSet/Workers.pm
+++ b/lib/OpenQA/Schema/ResultSet/Workers.pm
@@ -1,0 +1,23 @@
+# Copyright SUSE LLC
+# SPDX-License-Identifier: GPL-2.0-or-later
+
+package OpenQA::Schema::ResultSet::Workers;
+use Mojo::Base 'DBIx::Class::ResultSet', -signatures;
+
+sub stats ($self) {
+    my $total = $self->count;
+    my $total_online = grep { !$_->dead } $self->all();
+    my $free_active_workers = grep { !$_->dead } $self->search({job_id => undef, error => undef})->all();
+    my $free_broken_workers = grep { !$_->dead } $self->search({job_id => undef, error => {'!=' => undef}})->all();
+    my $busy_workers = grep { !$_->dead } $self->search({job_id => {'!=' => undef}})->all();
+
+    return {
+        total => $total,
+        total_online => $total_online,
+        free_active_workers => $free_active_workers,
+        free_broken_workers => $free_broken_workers,
+        busy_workers => $busy_workers,
+    };
+}
+
+1;
--- a/lib/OpenQA/Setup.pm
+++ b/lib/OpenQA/Setup.pm
@@ -77,6 +77,7 @@ sub read_config ($app) {
             max_rss_limit => 0,
             profiling_enabled => 0,
             monitoring_enabled => 0,
+            mcp_enabled => 'no',
             plugins => undef,
             hide_asset_types => 'repo',
             file_security_policy => 'download-prompt',
@@ -236,6 +237,7 @@ sub read_config ($app) {
             max_online_workers => 1000,
             wait_for_grutask_retries => 6,    # exponential, ~4 minutes
             worker_limit_retry_delay => ONE_HOUR / 4,
+            mcp_max_result_size => 500000,
         },
         archiving => {
             archive_preserved_important_jobs => 0,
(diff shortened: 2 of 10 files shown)


lib/OpenQA/Schema/ResultSet/Workers.pm
```diff
@@ -0,0 +1,23 @@
+# Copyright SUSE LLC
+# SPDX-License-Identifier: GPL-2.0-or-later
+
+package OpenQA::Schema::ResultSet::Workers;
+use Mojo::Base 'DBIx::Class::ResultSet', -signatures;
+
+sub stats ($self) {
+    my $total = $self->count;
+    my $total_online = grep { !$_->dead } $self->all();
+    my $free_active_workers = grep { !$_->dead } $self->search({job_id => undef, error => undef})->all();
```
d3flex, line 10: It seems you can reuse the `$total_online`. I havent tested but I guess just `$total_online->search(<whatever>)` will do.
As such it wont have to do this grep all the time.

lib/OpenQA/WebAPI/Controller/Admin/Workers.pm
```diff
@@ -29,24 +29,19 @@ sub _extend_info ($w) {
 
 sub index ($self) {
     my $workers_db = $self->schema->resultset('Workers');
-    my $total_online = grep { !$_->dead } $workers_db->all();
-    my $total = $workers_db->count;
-    my $free_active_workers = grep { !$_->dead } $workers_db->search({job_id => undef, error => undef})->all();
-    my $free_broken_workers
-      = grep { !$_->dead } $workers_db->search({job_id => undef, error => {'!=' => undef}})->all();
-    my $busy_workers = grep { !$_->dead } $workers_db->search({job_id => {'!=' => undef}})->all();
```
d3flex, line 37: now i see that comes from here. So it's up to you

lib/OpenQA/Setup.pm
```diff
@@ -236,6 +237,7 @@ sub read_config ($app) {
             max_online_workers => 1000,
             wait_for_grutask_retries => 6,    # exponential, ~4 minutes
             worker_limit_retry_delay => ONE_HOUR / 4,
+            mcp_max_result_size => 500000,
```
d3flex, line 240: Maybe that could be a constant

kalikiana (approved):

lib/OpenQA/Setup.pm
okurz: Look above for more than 10 other examples where the verbatim value is set and that is ok. We define the default configuration values in here.

lib/OpenQA/Schema/ResultSet/Workers.pm
okurz: Be aware that this PR brings back @kraih's original commit from
* #6685

but just with adapted handling of the perl-MCP dependency in a sub-package. I did not change any other code and can not take credit for it nor do I want to change it at this point as the original PR was approved and merge. The only exception is if we find real flaws that should have prevented approval in the original PR as well.

PR #6713, opened 2025-09-04
Make ping available inside CI container

nicksinger:
I hope this helps fixing the failing CI in https://github.com/os-autoinst/os-autoinst/pull/2743


nicksinger:
https://github.com/os-autoinst/os-autoinst/pull/2744 should be a better place to add this dependency

PR #6714, opened 2025-09-04
RFC: Allow multiple machines per job template

perlpunk:
Issue: https://progress.opensuse.org/issues/188310
This is just a proof of concept on how we could allow multple machines per job template in SCENARIO_DEFINITIONs.

It came up because we possibly want to run openqa-in-openqa jobs for both machines, and it's a useful feature in general.

Here it how our scenario file would look like:
https://github.com/perlpunk/os-autoinst-distri-openQA/blob/multiple-machines/scenario-definitions.yaml

Two possibilities to define machines per job template:
```
job_templates:

  # define traditionally as mapping key, as a string or an array of strings
  openqa_from_git:
    machine: [64bit-2G , uefi-2G]
    <<: *common
    settings:
      DESKTOP: minimalx
      OPENQA_FROM_GIT: "1"

  # define as template suffix
  openqa_from_containers@(64bit-2G,uefi2G):
    <<: *common
    settings:
      DESKTOP: minimalx
      LOAD_PYTHON_TEST_MODULES: "0"
      OPENQA_CONTAINERS: "1"
      OPENQA_FROM_GIT: "1"
```

### DIFF
--- a/lib/OpenQA/Schema/Result/ScheduledProducts.pm
+++ b/lib/OpenQA/Schema/Result/ScheduledProducts.pm
@@ -763,9 +763,34 @@ sub _schedule_from_yaml ($self, $args, $skip_chained_deps, $include_children, @l
     my $machines = $data->{machines} // {};
     my $job_templates = $data->{job_templates};
     my ($error_msg, %wanted, @job_templates);
+    my @flattened;
     for my $key (sort keys %$job_templates) {
-        my $job_template = $job_templates->{$key};
-        my $settings = $job_template->{settings} // {};
+        my $template = $job_templates->{$key};
+        my %new_template = (%$template, name => $key);
+        my @machines;
+        if ($key =~ s/\@\(([\w,-]+)\)$//) {
+            @machines = grep { length } split /,/, $1;
+        }
+        elsif ($template->{machine}) {
+            my $machine = delete $template->{machine};
+            if (ref $machine eq 'ARRAY') {
+                @machines = @$machine;
+            }
+            else {
+                @machines = $machine;
+            }
+        }
+        else {
+            push @flattened, {%new_template};
+            next;
+        }
+        for my $machine (@machines) {
+            push @flattened, {%new_template, machine => $machine};
+        }
+    }
+    for my $job_template (@flattened) {
+        my $key = $job_template->{name};
+        my $settings = {%{$job_template->{settings} // {}}};
         $settings->{TEST} = $key;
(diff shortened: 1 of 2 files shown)


lib/OpenQA/Schema/Result/ScheduledProducts.pm
```diff
@@ -763,9 +763,34 @@ sub _schedule_from_yaml ($self, $args, $skip_chained_deps, $include_children, @l
...
+        my $template = $job_templates->{$key};
+        my %new_template = (%$template, name => $key);
+        my @machines;
+        if ($key =~ s/\@\(([\w,-]+)\)$//) {
+            @machines = grep { length } split /,/, $1;
+        }
+        elsif ($template->{machine}) {
+            my $machine = delete $template->{machine};
+            if (ref $machine eq 'ARRAY') {
+                @machines = @$machine;
```
Martchus, line 777: Maybe also remove empty machines here for the sake of consistency:
```suggestion
                @machines = grep { length } @$machine;
```

lib/OpenQA/Schema/Result/ScheduledProducts.pm
```diff
@@ -763,9 +763,34 @@ sub _schedule_from_yaml ($self, $args, $skip_chained_deps, $include_children, @l
     my $machines = $data->{machines} // {};
     my $job_templates = $data->{job_templates};
     my ($error_msg, %wanted, @job_templates);
+    my @flattened;
     for my $key (sort keys %$job_templates) {
-        my $job_template = $job_templates->{$key};
-        my $settings = $job_template->{settings} // {};
+        my $template = $job_templates->{$key};
+        my %new_template = (%$template, name => $key);
```
Martchus, line 769: This is never used as a hash so we could avoid hashing:

```suggestion
        my @new_template = (%$template, name => $key);
```

Martchus (commented):
I guess this makes sense. The array syntax would be enough so maybe it would be best to only allow that. But I also don't really care as the code for allowing `@(…)` isn't really complicated.

okurz:
> Issue: https://progress.opensuse.org/issues/188310 This is just a proof of concept on how we could allow multple machines per job template in SCENARIO_DEFINITIONs.

With "allow" do you mean to add the array syntax to schedule all specified variants?

perlpunk:
I think it would also be useful to have a default like in jobgroup templates, so that we don't have to specify any machine if it's the default.

PR #6717, opened 2025-09-05
Stop logging confusing exception when GruTask is gone

kalikiana:
The exception is misleading as this is expected.

See: https://progress.opensuse.org/issues/188007

### DIFF
--- a/t/14-grutasks-git.t
+++ b/t/14-grutasks-git.t
@@ -378,8 +378,7 @@ subtest 'enqueue_git_clones' => sub {
         $t->app->log(Mojo::Log->new(level => 'debug'));
         my $jobs = [$j[4]->id];
         stderr_like { $t->app->gru->_add_jobs_to_gru_task(999, $jobs) }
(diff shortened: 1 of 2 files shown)


okurz (approved):

t/14-grutasks-git.t
```diff
@@ -378,8 +378,7 @@ subtest 'enqueue_git_clones' => sub {
         $t->app->log(Mojo::Log->new(level => 'debug'));
         my $jobs = [$j[4]->id];
         stderr_like { $t->app->gru->_add_jobs_to_gru_task(999, $jobs) }
-qr{GruTask 999 already gone.*insert or update on table "gru_dependencies" violates foreign key constraint "gru_dependencies_fk_gru_task_id"},
-          'expected log output if GruTask deleted in between';
+        qr{GruTask 999 already gone, skip assigning jobs"}, 'expected log output if GruTask deleted in between';
```
perlpunk:
```suggestion
        qr{GruTask 999 already gone, skip assigning jobs}, 'expected log output if GruTask deleted in between';
```

kalikiana: Thanks 😅

PR #6718, opened 2025-09-05
Use find_or_create when registering assets

kalikiana:
This is to avoid a race condition where the asset was already registered.

See: https://progress.opensuse.org/issues/188007

### DIFF
--- a/lib/OpenQA/Schema/ResultSet/Assets.pm
+++ b/lib/OpenQA/Schema/ResultSet/Assets.pm
@@ -42,7 +42,7 @@ sub register ($self, $type, $name, $options = {}) {
             $asset->refresh_size if $options->{refresh_size};
             if (my $created_by = $options->{created_by}) {
                 my $scope = $options->{scope} // 'public';
-                $created_by->jobs_assets->create({asset_id => $asset->id, created_by => 1});
+                $created_by->jobs_assets->find_or_create({asset_id => $asset->id, created_by => 1});
                 $created_by->reevaluate_children_asset_settings if $scope ne 'public';
             }
             return $asset;


okurz (approved):

Martchus (approved):
The commit message is a bit misleading but the change will probably avoid the situation where we run into an error here.

lib/OpenQA/Schema/ResultSet/Assets.pm
```diff
@@ -42,7 +42,7 @@ sub register ($self, $type, $name, $options = {}) {
             $asset->refresh_size if $options->{refresh_size};
             if (my $created_by = $options->{created_by}) {
                 my $scope = $options->{scope} // 'public';
-                $created_by->jobs_assets->create({asset_id => $asset->id, created_by => 1});
+                $created_by->jobs_assets->find_or_create({asset_id => $asset->id, created_by => 1});
```
d3flex, line 45: I am kinda afraid to accept this change. And I need to dig deeper to see whats going on here. But the `find_or_create` doesnt prevent `duplicate key value violates unique constraint` and the documentation have some caveats. 

```Note: Because find_or_create() reads from the database and then possibly inserts based on the result, this method is subject to a race condition. Another process could create a record in the table after the find has completed and before the create has started. To avoid this problem, use find_or_create() inside a transaction.```

In additional I found commit which replace the `find_or_create` exactly because it was not atomic. But I am struggling to understand the intention of this _if_ block.  
----
So far I think:
- the transaction insert row in the assets and ignores conflict
- then refresh the size of the asset!
- retrieves the asset from db and 
- reevaluate asset settings

What I do not understand is that the `$created_by->jobs_assets` seems to refer to another table than assets

perlpunk: Just linking to the commit you were referring to: 95408f8f55561d280104ec1115fa243c946e44fc

kalikiana: Awesome find. I'll check it out later.

Martchus, line 45: Maybe it would have been better to use `update_or_create` here or raw SQL, see https://progress.opensuse.org/issues/188667#note-7.

Martchus: Perhaps there already exists a row with `created_by == false` for the `job_id` and `asset_id`. This would explain the error that PR is trying to fix. However, when fixing it via `find_or_create` the `created_by` column wouldn't be set to `true` so we probably need `update_or_create` here. I still don't see how it would cause https://progress.opensuse.org/issues/188667#note-7, though.

PR #6719, opened 2025-09-06
AMQP: include list of failed modules in job.done messages

AdamWill:
This was requested by a Fedora maintainer who wrote a listener for failed jobs:
https://github.com/edsantiago/containertools/tree/main/openqa-listener
Seems like a useful piece of data, but if we don't want to put it in here I can tell Ed he has to go get it out of the API.


lib/OpenQA/WebAPI/Plugin/AMQP.pm
```diff
@@ -104,6 +105,12 @@ sub on_job_event ($self, $args) {
         if ($event_data->{bugref} = $bugref) {
             $event_data->{bugurl} = OpenQA::Utils::bugurl($bugref);
         }
+        # include failed test modules
+        my $tmo = read_test_modules($job);
+        my $test_modules = $tmo ? $tmo->{modules} : [];
+        my @faileds = grep { $_->{result} eq OpenQA::Jobs::Constants::FAILED } @$test_modules;
```
okurz: I wonder if we don't have that already available to not need to compute for failed modules. In the test overview and "next and previous" page we also show "failed modules", maybe available from there?

lib/OpenQA/WebAPI/Plugin/AMQP.pm
```diff
@@ -104,6 +105,12 @@ sub on_job_event ($self, $args) {
...
+        my @fnames = map { $_->{name} } @faileds;
+        $event_data->{failedmodules} = @faileds ? \@fnames : [];
```
Martchus:
```suggestion
        $event_data->{failed_modules} = $job->failed_modules;
```

Martchus (changes requested):

lib/OpenQA/WebAPI/Plugin/AMQP.pm
okurz: https://github.com/os-autoinst/openQA/pull/6719#pullrequestreview-3195506015 pretty much answers that :)

lib/OpenQA/WebAPI/Plugin/AMQP.pm
AdamWill: thanks. I was away yesterday so didn't get to look at it, will verify and update the PR tomorrow.

AdamWill: btw, I put no underscore in the name because that seems to be the style for other fields in the message - there's `bugref` and `newbuild`. There's also `group_id`, but it's still 2 vs. 1. And underscores are a pain to type. :D

okurz: > And underscores are a pain to type. :D

For that I recommend https://neo-layout.org/, works well for German and English and Perl and Python :)

AdamWill: thanks, done.

lib/OpenQA/WebAPI/Plugin/AMQP.pm
```diff
@@ -104,6 +105,8 @@ sub on_job_event ($self, $args) {
         if ($event_data->{bugref} = $bugref) {
             $event_data->{bugurl} = OpenQA::Utils::bugurl($bugref);
         }
+        # include failed test modules
```
okurz:
```suggestion
        # include failed test modules
```

I guess the next line is really clear enough about what it does so that we don't need a comment :)

AdamWill: hmm, yeah, I guess with the revision that's true.

PR #6731, opened 2025-09-10
Improve validation/UI of cleanup configuration

Martchus:
See individual commit messages and https://progress.opensuse.org/issues/187698 for details

This change does *not* yet tweak the behavior. It only introduces a warning and better displaying of warnings/errors. I'm leaving further improvements like automatically adjusting values for another PR because those are probably more controversial.

---

Feedback in case of warnings:

screenshot_20250910_132134

---

Feedback in case of errors:

screenshot_20250910_141632

So the rendering of the error message hasn't really changed except that field names are now displayed using the same labels as in the UI.

### DIFF
--- a/assets/javascripts/job_templates.js
+++ b/assets/javascripts/job_templates.js
@@ -568,8 +568,8 @@ function submitProperties(form) {
         collapse.click();
       }
       const overallError = updateValidation(form, json);
-      if (!response.ok || overallError)
-        throw `Server returned ${response.status}: ${response.statusText}\n${overallError || ''}`;
+      if (overallError) throw overallError;
+      if (!response.ok) throw `Server returned ${response.status}: ${response.statusText}`;
       const warnings = json?.warnings_by_field;
       const remark =
         typeof warnings === 'object' && Object.keys(warnings).length > 0


lib/OpenQA/WebAPI/Controller/API/V1/JobGroup.pm
```diff
@@ -197,13 +199,42 @@ sub _check_keep_logs_and_results ($self, $properties, $group = undef) {
         my $log_value = $properties->{$log_key} // ($group ? $group->$log_key : 0);
         my $result_value = $properties->{$result_key} // ($group ? $group->$result_key : 0);
         my $job_value = $properties->{$job_key} // ($group ? $group->$job_key : 0);
-        push @errors, "'$log_key' must be <= '$result_key'" if $result_value != 0 && $log_value > $result_value;
-        push @errors, "'$result_key' must be <= '$job_key'" if $job_value != 0 && $result_value > $job_value;
+        if ($result_value != 0 && $log_value > $result_value) {
+            push @{$errors_by_field{$log_key}}, "must be <= '$result_key'";
+            push @errors, "'$log_key' must be <= '$result_key'";
```
kalikiana, line 204: It would be cool if `<=` could be avoided and it reads more like a sentence e.g. instead of `must be <= "Keeps results for"` make it `The limit for logs must not be higher than the limit for results`. Tho I can be convinced it's not worth the effort.

assets/javascripts/job_templates.js
```diff
@@ -523,9 +563,19 @@ function submitProperties(form) {
...
-      if (!response.ok || json.error)
-        throw `Server returned ${response.status}: ${response.statusText}\n${json.error || ''}`;
-      showSubmitResults(editorForm, '<i class="fa fa-save"></i> Changes applied');
+      const collapse = document.getElementById('show-advanced-cleanup-settings-button');
+      if (json?.error?.includes('_jobs') && collapse.getAttribute('aria-expanded') !== 'true') {
+        collapse.click();
+      }
+      const overallError = updateValidation(form, json);
+      if (!response.ok || overallError)
+        throw `Server returned ${response.status}: ${response.statusText}\n${overallError || ''}`;
```
kalikiana: Maybe leave out the technical bits? We have inline validation now, and the bad request is handled in code already.

t/ui/13-admin.t
```diff
@@ -428,17 +428,25 @@ subtest 'job property editor' => sub() {
...
+        $ele->send_keys('501');
         is($driver->find_element('#properties p.buttons button.btn-primary')->get_attribute('disabled'),
             undef, 'group properties save button is enabled');
         $driver->find_element_by_id('editor-carry-over-bugrefs')->click();
         $driver->find_element('#properties p.buttons button.btn-primary')->click();
         wait_for_ajax(msg => 'ensure there is no race condition, even though the page is reloaded');
+        my $status = $driver->find_element('.properties-status')->get_text;
+        my $feedback = $driver->find_element('#editor-keep-important-results-in-days ~ .invalid-feedback')->get_text;
+        like $status, qr/Changes applied.*warnings/i, 'remark about warnings';
+        like $feedback, qr/should be >= "Keep results for"/i, 'warning shown';
```
kalikiana: Please also add a case involving `.is-invalid-non-critical` here

lib/OpenQA/WebAPI/Controller/API/V1/JobGroup.pm
Martchus: Making it a sentence like this would be more effort and I find the use of `<=` actually much more concise. (It would be nice to use the UTF-8 "≤", though. But I also haven't done this here because I don't want to do too many changes in one PR.)

I would leave changes like these for a subsequent PR in case we don't want to make this auto-adjusting anyways. (I'm going to provide a PR implementing auto-adjusting. This PR will reword these errors/warnings anyway. So in case this future PR is merged the phrasing as-is becomes irrelevant.)

assets/javascripts/job_templates.js
Martchus: I can do this but would prefer doing this in a separate commit/PR. We shouldn't make too many/big changes at a time. Note that we only have inline validation if we have it. If the server responds with e.g. a 403 or 500 error we still need to deal with that the usual way.

t/ui/13-admin.t
Martchus: I added a check for that and also that the warning is shown in a badge of the expected type.

assets/javascripts/job_templates.js
Martchus: I created a separate commit. I kept the "Server returned …" message in case we don't have any more specific error to show but normally it should just show a specific error now.

lib/OpenQA/WebAPI/Controller/API/V1/JobGroup.pm
```diff
@@ -197,13 +199,42 @@ sub _check_keep_logs_and_results ($self, $properties, $group = undef) {
...
     }
-    $self->render(json => {error => join(', ', @errors)}, status => 400) if @errors;
+    $self->{_errors} = \@errors;
+    $self->{_errors_by_field} = \%errors_by_field;
+    $self->{_warnings_by_field} = \%warnings_by_field;
     return @errors == 0;
 }
 
+sub _render_json ($self, $id = undef) {
+    my $errors = $self->{_errors};
```
perlpunk, line 227: I don't like bypassing perl's object this way, but I won't request changes

PR #6732, opened 2025-09-10
Add usage examples of netrc to the access tokens

grisu48:
Adds concrete usage examples of storing personal access tokens using a netrc file for curl and for wget.


docs/Client.asciidoc
```diff
@@ -121,6 +121,19 @@ curl -H 'Authorization: Bearer arthur:1234567890ABCDEF:ABCDEF1234567890'\
     -X DELETE https://openqa.example.com/api/v1/assets/1
 ----
 
+If you are using `curl` or `wget` one can utilise the `~/.netrc` file to
```
Martchus: This sentence should use either you or one but not a mix of both.

grisu48: Thanks, updated the whole paragraph to use the "one" style and be less direct.

PR #6733, opened 2025-09-11
Openapi validator

d3flex:
There are two commits with the very basic changes.
which provides:
- validation on requests
- validation on responses

Two packages are required `cpanm Mojolicious::Plugin::OpenAPI` and 
`cpanm Mojolicious::Plugin::SwaggerUI`. Remove the lines in the Setup for the second if you do not need it for now. 
If not, when you start the server you should see the API documentation with the Swagger UI (go to http://127.0.0.1/api/docs)

My understanding is that there are two ways to perform the validation. this is the not-automatic way. 
You can ignore the components section as well. it is not used.

### DIFF
--- a/lib/OpenQA/WebAPI.pm
+++ b/lib/OpenQA/WebAPI.pm
@@ -52,7 +52,11 @@ sub startup ($self) {
     OpenQA::Schema->singleton;
 
     # Some controllers are shared between openQA micro services
-    my $r = $self->routes->namespaces(['OpenQA::Shared::Controller', 'OpenQA::WebAPI::Controller', 'OpenQA::WebAPI', 'OpenQA::WebAPI::Controller::API::V1']);
+    my $r = $self->routes->namespaces(
+        [
+            'OpenQA::Shared::Controller', 'OpenQA::WebAPI::Controller',
+            'OpenQA::WebAPI', 'OpenQA::WebAPI::Controller::API::V1'
+        ]);
 
     # register basic routes
     my $logged_in = $r->under('/')->to('session#ensure_user');
--- a/lib/OpenQA/WebAPI.pm
+++ b/lib/OpenQA/WebAPI.pm
@@ -52,7 +52,7 @@ sub startup ($self) {
     OpenQA::Schema->singleton;
 
     # Some controllers are shared between openQA micro services
-    my $r = $self->routes->namespaces(['OpenQA::Shared::Controller', 'OpenQA::WebAPI::Controller', 'OpenQA::WebAPI']);
+    my $r = $self->routes->namespaces(['OpenQA::Shared::Controller', 'OpenQA::WebAPI::Controller', 'OpenQA::WebAPI', 'OpenQA::WebAPI::Controller::API::V1']);
 
     # register basic routes
     my $logged_in = $r->under('/')->to('session#ensure_user');
@@ -325,12 +325,12 @@ sub startup ($self) {
 
     my $job_r = $api_ro->any('/jobs/<jobid:num>');
     push @api_routes, $job_r;
-    $api_public_r->any('/jobs/<jobid:num>')->name('apiv1_job')->to('job#show');
+    #$api_public_r->any('/jobs/<jobid:num>')->name('apiv1_job')->to('job#show');
     $api_public_r->get('/experimental/jobs/<jobid:num>/status')->name('apiv1_get_status')->to('job#get_status');
     $api_public_r->any('/jobs/<jobid:num>/details')->name('apiv1_job')->to('job#show', details => 1);
     $job_r->put('/')->name('apiv1_put_job')->to('job#update');
     $job_r->delete('/')->name('apiv1_delete_job')->to('job#destroy');
-    $job_r->post('/prio')->name('apiv1_job_prio')->to('job#prio');
+    #$job_r->post('/prio')->name('apiv1_job_prio')->to('job#prio');
     $job_r->post('/set_done')->name('apiv1_set_done')->to('job#done');
     $job_r->post('/status')->name('apiv1_update_status')->to('job#update_status');
     $job_r->post('/artefact')->name('apiv1_create_artefact')->to('job#create_artefact');
--- a/lib/OpenQA/WebAPI/Controller/API/V1/Job.pm
+++ b/lib/OpenQA/WebAPI/Controller/API/V1/Job.pm
@@ -80,6 +80,8 @@ Limit the number of jobs.
 
 sub list ($self) {
     my $validation = $self->validation;
+    #$self->openapi->valid_input or return;
+    #my $validation = $self->openapi->validator;
     $validation->optional('scope')->in('current', 'relevant');
     $validation->optional('latest')->num(1);
     $validation->optional('limit')->num;
@@ -455,6 +457,8 @@ Pass follow=1 as query param to follow job clones and report most recent result 
 =cut
 
 sub show ($self) {
+    # manually triggers validation - unclear if this is how it should be
+    $self->openapi->valid_input or return;
     my $job_id = int($self->stash('jobid'));
     my $details = $self->stash('details') || 0;
     my $check_assets = !!$self->param('check_assets');
@@ -463,7 +467,7 @@ sub show ($self) {
     $job = $job->latest_job if $follow;
     $job = $job->to_hash(assets => 1, check_assets => $check_assets, deps => 1, details => $details, parent_group => 1);
     $job->{followed_id} = $job_id if ($job_id != $job->{id});
-    $self->render(json => {job => $job});
+    $self->render(openapi => {job => $job});
 }
 
 =over 4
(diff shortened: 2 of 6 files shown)


lib/OpenQA/WebAPI/Controller/API/V1/Job.pm
```diff
@@ -80,6 +80,8 @@ Limit the number of jobs.
 
 sub list ($self) {
     my $validation = $self->validation;
+    #$self->openapi->valid_input or return;
+    #my $validation = $self->openapi->validator;
```
perlpunk, line 84: Please remove

lib/OpenQA/WebAPI/Controller/API/V1/Job.pm
```diff
@@ -455,6 +457,8 @@ Pass follow=1 as query param to follow job clones and report most recent result
 =cut
 
 sub show ($self) {
+    # manually triggers validation - unclear if this is how it should be
```
perlpunk, line 460: You can remove the comment. That's how it should work.

lib/OpenQA/WebAPI/Controller/API/V1/Job.pm
```diff
@@ -463,7 +467,7 @@ sub show ($self) {
     $job = $job->latest_job if $follow;
     $job = $job->to_hash(assets => 1, check_assets => $check_assets, deps => 1, details => $details, parent_group => 1);
     $job->{followed_id} = $job_id if ($job_id != $job->{id});
-    $self->render(json => {job => $job});
+    $self->render(openapi => {job => $job});
```
perlpunk, line 470: With this you are telling the plugin to validate the **response**.
We said that we don't want this for now (we can always add it later).
If you remove this change, you can also remove the `responses` parts from the yaml file.

lib/OpenQA/WebAPI/Controller/API/V1/Job.pm
```diff
@@ -494,13 +498,17 @@ Sets priority for a given job.
 =cut
 
 sub prio ($self) {
+    
+    # seems this is the default behavior
```
perlpunk, line 502: ?

lib/OpenQA/WebAPI/Controller/API/V1/Job.pm
```diff
@@ -494,13 +498,17 @@ Sets priority for a given job.
...
+    $self->openapi->valid_input or return;
     return unless my $job = $self->find_job_or_render_not_found($self->stash('jobid'));
-    my $v = $self->validation;
-    my $prio = $v->required('prio')->num->param;
-    return $self->reply->validation_error({format => 'json'}) if $v->has_error;
+    #my $v = $self->validation;
+    #my $prio = $v->required('prio')->num->param;
```
perlpunk, line 506: Please remove

lib/OpenQA/WebAPI/Controller/API/V1/Job.pm
```diff
@@ -494,13 +498,17 @@ Sets priority for a given job.
...
+    my $prio = $self->param('prio');
+    #return $self->reply->validation_error({format => 'json'}) if $v->has_error;
```
perlpunk, line 508: Please remove

lib/OpenQA/WebAPI/Controller/API/V1/Job.pm
```diff
@@ -494,13 +498,17 @@ Sets priority for a given job.
...
 
     # Referencing the scalar will result in true or false (see http://mojolicio.us/perldoc/Mojo/JSON)
-    $self->render(json => {result => \$job->set_prio($prio)});
+    $self->render(openapi => {result => \$job->set_prio($prio)});
```
perlpunk, line 511: Please remove, see above

lib/OpenQA/WebAPI/Plugin/Helpers.pm
```diff
@@ -523,7 +523,7 @@ sub _param_hash ($c, $name) {
 sub _find_job_or_render_not_found ($c, $job_id, $query_settings = {}) {
     my $job = $c->schema->resultset('Jobs')->find(int($job_id), $query_settings);
     return $job if $job;
-    $c->render(json => {error => 'Job does not exist'}, status => 404);
+    $c->render(openapi => {error => 'Job does not exist'}, status => 404);
```
perlpunk, line 526: Please remove, see above

public/openapi.yaml
```diff
@@ -0,0 +1,230 @@
...
+externalDocs:
+  description: The official OpenQA doc
+  url: https://open.qa/docs/
+servers:
+  - url: /api/v1
+tags:
+  - name: public
+    description: Those API routes which are public (no auth)
+  - name: ro
+    description: read-only routes
```
perlpunk, line 14: I'm not sure `/jobs/$id/prio` is really a read only route :)

public/openapi.yaml
```diff
@@ -0,0 +1,230 @@
...
+security:
+  - basicAuth: []
```
perlpunk, line 16: Please remove for now

public/openapi.yaml
```diff
@@ -0,0 +1,230 @@
...
+paths:
+  /jobs/{jobid}:
+    get:
+      tags:
+        - public
+      summary: Filters jobs based on a single setting key/value pair.
+      operationId: apiv1_job
+      x-mojo-name: apiv1_job
```
perlpunk, line 24: I think we need only one of those two

public/openapi.yaml
```diff
@@ -0,0 +1,230 @@
...
+      #           properties:
+      #             error_status:
+      #               type: integer
+      #             error:
+      #               type: string
+      #               example: Erroneous parameters (key invalid, list_value invalid)
+      #       text/plain:
+      #         schema:
+      #           type: string
+      #           example: Erroneous parameters (key invalid, list_value invalid)
```
perlpunk, line 69: Please remove

public/openapi.yaml
```diff
@@ -0,0 +1,230 @@
...
+          in: path
+          required: true
+          description: 'The ID of the job to update.'
+          schema:
+            type: integer
+        # - $ref: '#/components/parameters/CSRFHeader'
+        # - in: cookie
+        #   name: csrftoken
+        #   schema:
+        #     type: string
```
perlpunk, line 91: Please remove

public/openapi.yaml
```diff
@@ -0,0 +1,230 @@
...
+        #       schema:
+        #         type: string
+        #         example: 'Specified scheduled_product_id is invalid.'
+        # '404':
+        #   description: 'Not Found.'
+        #   content:
+        #     text/plain:
+        #       schema:
+        #         type: string
+        #         example: 'Scheduled product to clone settings from not found.'
```
perlpunk, line 138: Please remove

public/openapi.yaml
```diff
@@ -0,0 +1,230 @@
...
+          type: object
+          additionalProperties: true
+
+    IsoAsyncResponse:
+      type: object
+      properties:
+        job_id:
+          type: integer
+        scheduled_product_id:
+          type: integer
```
perlpunk, line 230: Please remove

perlpunk (commented):
Ok, good :)

Now please put yourself in the shoes of the person picking up the followup ticket and looking at all the commented out stuff and also the not commented out stuff that is not needed.

A proof of concept doesn't need to pass all tests, and it can also have commented out parts, but please clean up the not needed parts that don't serve any purpose for what we want. I think it will look much cleaner then, and then we can concentrate on the things that are left.

See individual comments.

lib/OpenQA/WebAPI.pm
```diff
@@ -52,7 +52,11 @@ sub startup ($self) {
     OpenQA::Schema->singleton;
 
     # Some controllers are shared between openQA micro services
-    my $r = $self->routes->namespaces(['OpenQA::Shared::Controller', 'OpenQA::WebAPI::Controller', 'OpenQA::WebAPI']);
+    my $r = $self->routes->namespaces(
+        [
+            'OpenQA::Shared::Controller', 'OpenQA::WebAPI::Controller',
+            'OpenQA::WebAPI', 'OpenQA::WebAPI::Controller::API::V1'
```
perlpunk, line 58: I'm wondering about this one. Cc @kraih @Martchus
We need this so that the OpenAPI plugin can find the controllers.
But could this have any other implications to add this as a "global" namespace?
Currently we explicitly route to that namespace in WebAPI.pm for the api routes.
I couldn't find any configuration option in the OpenAPI plugin to configure a namespace.

public/openapi.yaml
d3flex: As far as I can tell, operationId is not required. is not one or the other. operationId can be used for some more advanced stuff (like links), but it has to be unique. here I use the same as `x-mojo-name` but I think it is not unique in WebAPI.pm

lib/OpenQA/WebAPI.pm
Martchus: If we cannot specify the namespace (and just specify the controller/method via e.g. `x-mojo-to: 'job#show'`) then we have to make sure that the names of the controller packages are unique. Probably not a big deal. Probably it wouldn't be hard to extend the plugin to allow specifying e.g. `x-mojo-ns: 'OpenQA::WebAPI::Controller::API::V1'` to avoid this.

d3flex:
not relevant anymore due to work on other PRs

PR #6736, opened 2025-09-12
Avoid partitioning in raid0 device

gauravpathak:
Related to: https://progress.opensuse.org/issues/186948

This commit skips partitioning of raid0 device /dev/md/openqa by
- removing `<disklabel>none</disklabel>` and letting AutoYaST decide what to choose during installation
- removing mountpoint `/var/lib/openqa` and filesystem type, since this is being taken care by systemd automount service(s)

### DIFF
--- a/contrib/ay-openqa-worker.xml.erb
+++ b/contrib/ay-openqa-worker.xml.erb
@@ -3,19 +3,19 @@
 <profile xmlns="http://www.suse.com/1.0/yast2ns" xmlns:config="http://www.suse.com/1.0/configns">
   <general>
     <mode>
-      <confirm config:type="boolean">false</confirm>
+      <confirm t="boolean">false</confirm>
     </mode>
   </general>
   <networking>
     <!-- okurz: 2023-06-13: For multiple network interfaces by default the last is configured, prefer the first -->
-    <interfaces config:type="list">
+    <interfaces t="list">
       <interface>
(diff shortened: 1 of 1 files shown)


contrib/ay-openqa-worker.xml.erb
```diff
@@ -80,7 +80,7 @@
         </partition>
         <partition>
           <mount>swap</mount>
-          <size>auto</size>
+          <size>2GiB</size>
```
okurz: that's unrelated to the commit. Shouldn't it be kept on "auto"?

gauravpathak: Keeping it auto assigns only 1GiB swap space.

okurz: which should also be enough. I would trust the "auto" here

contrib/ay-openqa-worker.xml.erb
```diff
@@ -123,17 +121,31 @@
     <% else %> <%# maybe we have only one disk left on the system %>
     <% if raidarray.length != 0 %>
     <drive>
```
d3flex, line 119: tiny thing. The new attribute is `t`. it replaces the `config:type`. I wouldnt ask for a change but I noticed that the document is mixed with both. What about keep it consistenced and use only one out of the two. if yes maybe that should be `t`

PR #6738, opened 2025-09-13
Demo Mojolicious::Plugin::OpenAPI - With custom validation helper

perlpunk:
Based on:
* #6734 

Issue: https://progress.opensuse.org/issues/188742

Adds a helper for a custom validation error response.


lib/OpenQA/WebAPI/Controller/API/V1/Job.pm
```diff
@@ -452,6 +440,11 @@ sub show ($self) {
     $self->render(json => {job => $job});
 }
 
+sub show_details ($self) {
+    $self->stash(details => 1);
+    show($self);
```
Martchus, line 445: Why not:
```suggestion
    $self->show;
```

templates/webapi/swagger.html.ep
```diff
@@ -0,0 +1,20 @@
+<!-- HTML for static distribution bundle build -->
+<!DOCTYPE html>
+<html lang="en">
+  <head>
+    <meta charset="UTF-8">
+    <title>Swagger UI</title>
+    <link rel="stylesheet" type="text/css" href="/swagger/swagger-ui.css" />
+    <link rel="stylesheet" type="text/css" href="index.css" />
```
Martchus, line 8: The use of absolute and relative paths is inconsistent. If it is possible I'd prefer using relative paths everywhere so one can easily put this under a sub directory. The same applies to the `script` tags below.

lib/OpenQA/WebAPI/Controller/API/V1/Worker.pm
```diff
@@ -30,27 +30,14 @@ Implements API methods relating to OpenQA Workers.
 
 =head1 METHODS
```
d3flex, line 31: should this also removed?

perlpunk: why?

PR #6739, opened 2025-09-15
Automatically increase important durations to match regular durations

Martchus:
This makes it easier to configure storage durations; instead of having to fix a probable configuration mistake manually it happens automatically with this change. The disadvantage is that one can now no longer configure a lower retention for important jobs than for regular jobs but this is most likely not what anybody wants anyway.

Related ticket: https://progress.opensuse.org/issues/187698

---

screenshot_20250915_102138

(See https://github.com/os-autoinst/openQA/pull/6731 for how it looked before this change.)

### DIFF
--- a/assets/javascripts/job_templates.js
+++ b/assets/javascripts/job_templates.js
@@ -511,6 +511,7 @@ function updateValidation(form, response) {
   const E = createElement;
   const errorsByField = response?.errors_by_field ?? {};
   const warningsByField = response?.warnings_by_field ?? {};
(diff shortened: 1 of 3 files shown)


perlpunk (approved):

okurz:
> code lgtm but Maybe it is a good idea to 'promote' it in the #eng-testing to get feedback on it as a feature.

this isn't SUSE specific so https://app.element.io/#/room/#openqa:opensuse.org would be more suitable

PR #6740, opened 2025-09-15
Improve code of job asset registration

Martchus:
See https://progress.opensuse.org/issues/188007 and individual commit messages.

### DIFF
--- a/lib/OpenQA/Schema/ResultSet/Assets.pm
+++ b/lib/OpenQA/Schema/ResultSet/Assets.pm
@@ -29,9 +29,15 @@ sub register ($self, $type, $name, $options = {}) {
         return undef;
     }
     my $schema = $self->result_source->schema;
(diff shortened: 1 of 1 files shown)


t/15-assets.t
```diff
@@ -345,6 +345,22 @@ subtest 'check for missing assets' => sub {
     };
 };
 
+subtest 'job asset creation' => sub {
+    my $jobs = $schema->resultset('Jobs');
+    my $job_assets = $schema->resultset('JobsAssets');
+    ok my $job = $jobs->first;
+    my $assets = $schema->resultset('Assets');
+    my $asset = $assets->register(repo => 'foo', {missing_ok => 1, created_by => $job});
+    ok $asset, 'asset created' or return;
```
d3flex, line 354: Shouldnt that be `explain` instead of return?

Martchus: If `$asset` is `undef` then there's nothing we could explain. So it should be `return`.

PR #6741, opened 2025-09-15
Improve client-side validation in group property editor

Martchus:
See https://progress.opensuse.org/issues/187698 and individual commit messages.

### DIFF
--- a/assets/javascripts/job_templates.js
+++ b/assets/javascripts/job_templates.js
@@ -339,7 +339,7 @@ function alignCols() {
 
 function toggleEdit() {
   $('#properties').toggle(250);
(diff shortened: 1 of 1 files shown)


assets/javascripts/job_templates.js
```diff
@@ -550,6 +550,20 @@ function updateValidation(form, response) {
   return overallError;
 }
 
+function showAdvancedFieldsIfJsonReferrsToThem(response) {
```
d3flex: likely its should be `showAdvancedFieldsIfJsonRefersToThem`, right? notice the double `r` in referrs

PR #6742, opened 2025-09-15
RFC: Automatically increase durations to match next bigger category

Martchus:
* Based on https://github.com/os-autoinst/openQA/pull/6741 so review that one first
* See the commit message of the last WIP commit (still WIP because I haven't updated tests yet)

### DIFF
--- a/lib/OpenQA/WebAPI/Controller/API/V1/JobGroup.pm
+++ b/lib/OpenQA/WebAPI/Controller/API/V1/JobGroup.pm
@@ -193,22 +193,6 @@ sub _check_keep_logs_and_results ($self, $properties, $group = undef) {
     my %warnings_by_field;
     my %changed_fields;
     my $prefix = $self->is_parent ? 'default_' : '';
-    for my $important ('', '_important') {
-        my $log_key = "${prefix}keep${important}_logs_in_days";
-        my $result_key = "${prefix}keep${important}_results_in_days";
-        my $job_key = "${prefix}keep${important}_jobs_in_days";
-        my $log_value = $properties->{$log_key} // ($group ? $group->$log_key : 0);
-        my $result_value = $properties->{$result_key} // ($group ? $group->$result_key : 0);
-        my $job_value = $properties->{$job_key} // ($group ? $group->$job_key : 0);
-        if ($result_value != 0 && $log_value > $result_value) {
-            push @{$errors_by_field{$log_key}}, "must be <= '$result_key'";
-            push @errors, "'$log_key' must be <= '$result_key'";
-        }
-        if ($job_value != 0 && $result_value > $job_value) {
-            push @{$errors_by_field{$result_key}}, "must be <= '$job_key'";
-            push @errors, "'$result_key' must be <= '$job_key'";
-        }
-    }
     for my $field (qw(logs results jobs)) {
         my $important_key = "${prefix}keep_important_${field}_in_days";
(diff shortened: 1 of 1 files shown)


lib/OpenQA/WebAPI/Controller/API/V1/JobGroup.pm
```diff
@@ -201,22 +211,12 @@ sub _check_keep_logs_and_results ($self, $properties, $group = undef) {
         my $result_value = $properties->{$result_key} // ($group ? $group->$result_key : 0);
         my $job_value = $properties->{$job_key} // ($group ? $group->$job_key : 0);
         if ($result_value != 0 && $log_value > $result_value) {
-            push @{$errors_by_field{$log_key}}, "must be <= '$result_key'";
-            push @errors, "'$log_key' must be <= '$result_key'";
+            $changed_fields{$result_key} = $properties->{$result_key} = $log_value;
+            push @{$warnings_by_field{$result_key}}, "automatically increased to match '$log_key'";
```
Martchus, line 215: This is the relevant error which this PR turns into a warning automatically increasing the value.

perlpunk:
https://app.circleci.com/pipelines/github/os-autoinst/openQA/18280/workflows/b00c05aa-b1f4-4797-9649-5b3232441e9a/jobs/180091
```
...
# Looks like you failed 1 test of 13.
[08:25:06] t/api/10-jobgroups.t ............... 
Failed 1/13 subtests 
```

Martchus:
> Will this fix the values and automatically store, or just redisplay the fixed values so that you have to click save again?

 Just like the previously merged PR https://github.com/os-autoinst/openQA/pull/6739 it will fix the values automatically and store then immediately. (It will of course *also* redisplay the fixed values to avoid confusion.)

The failing tests and WIP are because I'm only adjusting tests if this is actually wanted. I'm a bit torn with this myself. It makes things more convenient but people will also have "less opportunity/incentive" to get a better understanding of how the cleanup works.

lib/OpenQA/WebAPI/Controller/API/V1/JobGroup.pm
kalikiana: I'm on the fence. This might cause limits to be adjusted when they were only accidentally changed.

perlpunk:
> > Will this fix the values and automatically store, or just redisplay the fixed values so that you have to click save again?
> 
> Just like the previously merged PR #6739 it will fix the values automatically and store then immediately. (It will of course _also_ redisplay the fixed values to avoid confusion.)

I have to say, I didn't pay enough attention then with the previous PR.
I thought it would just display the error(s) and do nothing. Saving something and then afterwards displaying errors is a bit unexpected.
IMHO, display errors and don't save would be enough.

> The failing tests and WIP are because I'm only adjusting tests if this is actually wanted.

I see. You set it to "Ready to review" without a "Not ready" label, that's why I thought it's ready :)

lib/OpenQA/WebAPI/Controller/API/V1/JobGroup.pm
perlpunk: Yeah. When automatically adjusting, there should be the chance to just reload to redisplay the old values, so don't actually save it.

Martchus:
> I'm on the fence. This might cause limits to be adjusted when they were only accidentally changed.

> IMHO, display errors and don't save would be enough.

Ok, then I'm closing this PR and will resolve the ticket. (We are already displaying errors.)

PR #6744, opened 2025-09-15
Use more Mojo::File in OpenQA::Worker

okurz:
* Use Mojo::File in Worker for is_qemu_running
* Use Mojo::File in Worker for meminfo


lib/OpenQA/Worker.pm
```diff
@@ -593,11 +591,8 @@ sub is_qemu ($executable_path) { defined $executable_path && $executable_path =~
 # checks whether a qemu instance using the current pool directory is running and returns its PID if that's the case
 sub is_qemu_running ($self) {
     return undef unless my $pool_directory = $self->pool_directory;
-    return undef unless open(my $fh, '<', my $pid_file = "$pool_directory/qemu.pid");
-
-    my $pid = <$fh>;
-    chomp($pid);
-    close($fh);
+    my $pid_file = "$pool_directory/qemu.pid";
+    my $pid = path($pid_file)->slurp;
```
Martchus: This must not throw in case the file just doesn't exist.

This probably shouldn't throw in any error case but rather propagate the error in some other way.

lib/OpenQA/Worker.pm
```diff
@@ -180,15 +180,13 @@ sub capabilities ($self) {
     }
 
     # determine memory limit
-    open(my $MEMINFO, '<', '/proc/meminfo');
```
Martchus, line 184: I'm not sure whether throwing an exception makes sense here. (The previous code probably just returned the capabilities without `mem_max`.)

lib/OpenQA/Worker.pm
```diff
@@ -181,15 +181,13 @@ sub capabilities ($self) {
...
-    for my $line (<$MEMINFO>) {
-        chomp $line;
+    my @meminfo = split /\n/, path('/proc/meminfo')->slurp;
```
d3flex: Should this be `$meminfo = path('/proc/meminfo')->open('<');`

okurz: No. We don't need to keep any file handle open. We just read the complete file in one go.

lib/OpenQA/Worker.pm
okurz: done with `eval`

Martchus (approved):

perlpunk (approved):

PR #6748, opened 2025-09-16
Validate needle timestamp upon saving

kalikiana:
The suffix should always be - and a timestamp.

See: https://progress.opensuse.org/issues/188865


okurz (commented):
Have you checked that one can actually create a needle without timestamp?

Also this might not necessarily be something that we need/should always enforce for every use of openQA

kalikiana:
> Have you checked that one can actually create a needle without timestamp?

Yes. That's the problem I'm fixing 😛 

I'm filing a follow-up ticket as discussed in the estimation. This is only a proof of concept.

PR #6753, opened 2025-09-19
Redirect test assets to file_domain if configured

aylen384:
This will ensure all assets are served via file_domain if it is set.
Ticket: https://progress.opensuse.org/issues/188226

### DIFF
--- a/lib/OpenQA/WebAPI/Controller/File.pm
+++ b/lib/OpenQA/WebAPI/Controller/File.pm
@@ -138,11 +138,16 @@ sub test_asset ($self) {
       if ($path =~ /\/\.\./ || $path =~ /\.\.\//);
 
     # map to URL - mojo will canonicalize
-    $path = $self->url_for('download_asset', assetpath => $path);
-    $self->app->log->debug("redirect to $path");
+    my $url = $self->url_for('download_asset', assetpath => $path);
+
+    # redirect to file domain if configured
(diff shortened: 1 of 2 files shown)


d3flex:
> The `_redirect_if_configured` method does this based on the current url and is only called by `_serve_static`. But the `test_asset` method will always generate a redirect (not based on the current url) because even before the security related changes here, we would redirect e.g.
> 
> * from `/tests/5323439/asset/iso/openSUSE-Tumbleweed-DVD-ppc64le-Snapshot20250918-Media.iso`
> * to `/tests/5323439/asset/iso/openSUSE-Tumbleweed-DVD-ppc64le-Snapshot20250918-Media.iso`
> 
> And now we change that to redirect to
> 
> * `//openqa-assets.opensuse.org/tests/5323439/asset/iso/openSUSE-Tumbleweed-DVD-ppc64le-Snapshot20250918-Media.iso`

Not sure. I am pretty sure from https://github.com/os-autoinst/openQA/pull/6668/files and https://github.com/os-autoinst/openQA/pull/6704 that there is redirection to a domain like openqa-assets.opensuse.org/tests/5323439/asset/iso/openSUSE-Tumbleweed-DVD-ppc64le-Snapshot20250918-Media.iso. I might be wrong but it seems that `_redirect_if_configured` is reusable here.

PR #6759, opened 2025-09-22
Don't redirect screenshots, thumbs and needles to files_domain

aylen384:
Ticket: https://progress.opensuse.org/issues/188226


Martchus (approved):
Tests would be nice, though.

PR #6760, opened 2025-09-22
Document archiving of important jobs

kalikiana:
See: https://progress.opensuse.org/issues/187998

### DIFF
--- a/docs/Client.asciidoc
+++ b/docs/Client.asciidoc
@@ -377,7 +377,7 @@ Just be aware that this is a feature the openQA team does not control, and the
 exact output as well as how it escapes control characters will change a bit over
 time.
 
-== Archive mode
+== Archiving individual job results
 
 With the `archive` subcommand of `openqa-cli` you can download all the assets
 and test results of a job for archiving or debugging purposes.
@@ -393,3 +393,6 @@ Thumbnails are not included by default, but can be added with the
 openqa-cli archive --host http://openqa.example.com --with-thumbnails \
     408 ~/openqa_job_408
 ----
+
+See also <<UsersGuide.asciidoc#archiving,Archiving>> for automatic archiving
+of old jobs.
--- a/docs/UsersGuide.asciidoc
+++ b/docs/UsersGuide.asciidoc
@@ -1435,10 +1435,25 @@ Further remarks:
   section of <<GettingStarted.asciidoc#_configuration,the web UI configuration>>.
   Jobs outside of any group use the limits configured in the `[no_group_limits]`
   section of the web UI configuration.
-* Archiving of important jobs can be enabled. Checkout the related settings
-  within the `[archiving]` section of the config file for details.
 * The <<Installing.asciidoc#auditing,Auditing>> section explains the cleanup of
   the audit log.
+
+[id="archiving"]
+=== Automatic archiving of old jobs ===
+
+Archiving of important jobs can be enabled:
+[source,ini]
+--------------------------------------------------------------------------------
+[archiving]
+## Moves results of important jobs which are preserved during the cleanup
+## to OPENQA_ARCHIVEDIR (/var/lib/openqa/archive/ by default)
+#archive_preserved_important_jobs = 0
+--------------------------------------------------------------------------------
+
+When archiving is enabled important job results are moved to the archive
+directory once they reach the expiration for regular job logs. This
+happens as part of the normal cleanup. In the web UI an icon indicates that
+a job was archived.
 
 [id="asset_cleanup"]
 === Cleanup strategy for assets ===
--- a/docs/Installing.asciidoc
+++ b/docs/Installing.asciidoc
@@ -1653,6 +1653,9 @@ the workers create their own instance directories.
 * the "project directory"
     - defined as `$basedir/openqa`, by default `/var/lib/openqa`
     - referred as `$prjdir` within openQA
+* the "archive directory"
+    - defined as `$prjdir/openqa`, by default `/var/lib/openqa/archive`
+    - referred as `$archivedir` within openQA
 * the "share directory": contains directories shared between web UI and (remote) workers
     - defined as `$prjdir/share`, by default `/var/lib/openqa/share`
     - referred as `$sharedir` within openQA


docs/Client.asciidoc
```diff
@@ -377,7 +377,7 @@ Just be aware that this is a feature the openQA team does not control, and the
 exact output as well as how it escapes control characters will change a bit over
 time.
 
-== Archive mode
+== Archiving of jobs
```
okurz: To better distinguish from the "Automatic archiving of old jobs" how about "Archiving individual job results"

docs/UsersGuide.asciidoc
```diff
@@ -1435,11 +1435,28 @@ Further remarks:
...
+## to OPENQA_ARCHIVEDIR (/var/lib/openqa/archive/ by default)
+#archive_preserved_important_jobs = 0
+--------------------------------------------------------------------------------
+
+Currently the primary use case is keeping important jobs in a dedicated
+directory once they expire. Results of such jobs are periodically moved to
+`OPENQA_ARCHIVEDIR`. In the web UI an icon indicates that a job was archived.
+
+Note that retention periods are the same as for important logs and archiving
+is not reversible.
```
okurz: The above comment in line 1448 reads nicely but here it's a bit convoluted. The first sentence says "keeping important jobs […] once they expire" which is contradicting. And results are not really moved periodically but when the according cleanup is triggered. How about the following to make it clear that the alternatives are deletion or archiving:

```suggestion
When archiving is enabled then job results that reach their expiration but are
considered important are moved to the archive. In the web UI an icon
indicates that a job was archived. As only important jobs are considered
for archiving the retention periods for important job results and logs also
apply here.
```

okurz (changes requested):

docs/UsersGuide.asciidoc
```diff
@@ -1435,11 +1435,27 @@ Further remarks:
...
+## Moves results of important jobs which are preserved during the cleanup
+## to OPENQA_ARCHIVEDIR (/var/lib/openqa/archive/ by default)
+#archive_preserved_important_jobs = 0
+--------------------------------------------------------------------------------
+
+When archiving is enabled important job results are moved to the archive
+directory when they reach their expiration (instead of being deleted). This
+happens as part of the normal cleanup. In the web UI an icon indicates that
+a job was archived. As only important jobs are archived the same retention
+periods apply.
```
Martchus: You should double-check the code whether that's really how it works. I think it works more like the comment about the config setting already insinuates. Out of my head I think this is how archiving behaves:

```suggestion
When enabled, archiving applies to important jobs. More specifically, when an
important job reaches the expiration configured for regular job results - but is
not deleted because it is important - it will be archived. This means moving its
results to `OPENQA_ARCHIVEDIR`. This allows configuring high storage durations
for important job results provided there is enough disk space on
`OPENQA_ARCHIVEDIR`.
```

Martchus (changes requested):

docs/UsersGuide.asciidoc
Martchus: And if I'm correct we should probably split the sentence with the `,` and `-` into two smaller sentences.

kalikiana: I made an effort to trace the code in order to document it. If normal and important retention limits are mixed up I don't understand where the bug is. Maybe best to file a ticket and fix it first, because this will be confusing for users of the feature.

kalikiana: For reference I was checking especially t/14-grutasks.t, lib/OpenA/Task/Job/Limit.pm and lib/OpenQA/Schema/Result/JobGroups.pm - maybe extending the unit test would be the easiest to confirm the behavior.

Martchus: I checked the code again. To me it looks like it would behave as in my description except that the cleanup/retention of logs is relevant (and not the cleanup/retention of results).

The limit task calls `limit_results_and_logs` which returns jobs to be archived via `$preserved_important_jobs`. This variable is populated by `_find_expired_jobs` which uses the condition `-and => [{t_finished => $timecond}, {t_finished => {'>=' => $important_timestamp}}]` to assign jobs to that variable that *newer* than the expiration date of important jobs (but still older than the expiration date of regular jobs).

Emphasis is on *newer* because this contradicts the wording "instead of being deleted" in your change because jobs "being deleted" would be *older* than the expiration date. Or am I right now confused on how these queries with timestamps work?

---

> If normal and important retention limits are mixed up I don't understand where the bug is.

I wouldn't consider this a bug, just unintuitive/complex behavior. (And this *is* how we (actually mainly @okurz) wanted this to work avoiding the introduction of yet another distinct retention value for archiving (which is how I would have done it for the sake of simplicity). This is at least how I remember it. I also wouldn't say limits are "mixed up". Using the cleanup thresholds in archiving (which is obviously a different action than cleanup) is of course counterintuitive and leads to cleanup and archiving being entangled with each other and therefore more complexity. But I don't think we should go into this discussion from the past again.)

---

> For reference I was checking especially t/14-grutasks.t, lib/OpenA/Task/Job/Limit.pm and lib/OpenQA/Schema/Result/JobGroups.pm - maybe extending the unit test would be the easiest to confirm the behavior.

The check

```
ok -e $filename, 'results of important job preserved if only exceeding normal retention period';
```

followed by

```
is $job->archived, 0, 'job not archived yet';
```

does *not* mean it is like in your description. The job is just asserted to be not archived because archiving is not enabled at this point. And in a subsequent test which actually runs with archiving enabled we have the check

```
is $job->archived, 1, 'job flagged as archived';
```

for the same job. That makes it look like my description is in-line with the test.

kalikiana: > I checked the code again. To me it looks like it would behave as in my description except that the cleanup/retention of logs is relevant (and not the cleanup/retention of results).

Right! My original phrasing was this:

    Note that retention periods are the same as for important logs and archiving is not reversible.

This got lost when I tried to simplify the wording.

Martchus: The original phrasing is definitely better :-)

kalikiana: > Emphasis is on newer because this contradicts the wording "instead of being deleted" in your change because jobs "being deleted" would be older than the expiration date. Or am I right now confused on how these queries with timestamps work?

Ack. I just tried to keep the explanation simple.

kalikiana: > I wouldn't consider this a bug, just unintuitive/complex behavior.

Maybe I can phrase it more objectively. My user story is [keeping job results of certain job groups for multiple years, important and not important](https://progress.opensuse.org/issues/185374). At this point I assume we will need a distinct value.

kalikiana:
```
target livehandler: failed to solve: process "/bin/sh -c zypper ar -p 95 -f http://download.opensuse.org/repositories/devel:openQA/15.6 devel_openQA &&     zypper ar -p 90 -f http://download.opensuse.org/repositories/devel:openQA:Leap:15.6/15.6 devel_openQA_Leap &&     zypper --gpg-auto-import-keys ref &&     zypper in -y ca-certificates-mozilla curl openQA-local-db apache2 apache2-utils hostname which w3m 'rubygem(sass)>=3.7.4' &&     zypper clean &&     gensslcert &&     a2enmod headers &&     a2enmod proxy &&     a2enmod proxy_http &&     a2enmod proxy_wstunnel &&     a2enmod ssl &&     a2enmod rewrite &&     a2enflag SSL" did not complete successfully: exit code: 8
[...]
52.06 Retrieving: openQA-common-5.1758538159.eb335552-lp156.8382.1.x86_64.rpm [.error]
52.07 Download (curl) error for 'http://cdn.opensuse.org/repositories/devel:openQA/15.6/x86_64/openQA-common-5.1758538159.eb335552-lp156.8382.1.x86_64.rpm':
52.07 Error code: Curl error 56
52.07 Error message: Recv failure: Connection reset by peer
```

docs/UsersGuide.asciidoc
Martchus: Yes, to cover the "not important" part we need a distinct value for archiving. (Although now, with the distinct retention for having the job just in the database we might come up with a clever reuse of the result retention for archiving. However, I'm not even thinking how this would be like. It would make phrasing the documentation even more challenging and add even more complexity.)

okurz: @kalikiana I don't think we will enable archiving for not-important jobs and I don't see the need. The original user story is independant of any openQA implementation of "important jobs". To cover the user story we can just ensure that each SLE maintenance test is "important" and then archiving is an option

kalikiana: Does that mean I can also document that as the supported setup? I don't want to have to explain the real logic to users 😛

docs/UsersGuide.asciidoc
```diff
@@ -1435,11 +1435,28 @@ Further remarks:
...
+[source,ini]
+--------------------------------------------------------------------------------
+[archiving]
+## Moves results of important jobs which are preserved during the cleanup
+## to OPENQA_ARCHIVEDIR (/var/lib/openqa/archive/ by default)
+#archive_preserved_important_jobs = 0
+--------------------------------------------------------------------------------
+
+When archiving is enabled important job results are moved to the archive
+directory once they reach the expiration for regular job results. This
```
Martchus:
```suggestion
directory once they reach the expiration for regular job logs. This
```

docs/UsersGuide.asciidoc
```diff
@@ -1435,11 +1435,28 @@ Further remarks:
...
+happens as part of the normal cleanup. In the web UI an icon indicates that
+a job was archived.
+
+Note that retention periods are the same as for important logs.
```
Martchus: This sentence doesn't add much value here (as retentions are covered in the cleanup section) and it can also be misinterpreted.

```suggestion
```

Martchus (approved):

PR #6762, opened 2025-09-22
Avoid installing NPM dev deps unnecessarily in `Makefile`

Martchus:
This reduces the number of NPM packages pulled in when running regular tests so `eslint` and its dependencies are only installed as needed. This helps reducing the attack surface from malicious NPM packages a little bit.

Note that the `install` target already only installs files from the `node_modules` directory that are really needed to run the web application so there are no adjustments needed here.

Those dependencies are still generally made available unnecessarily during RPM builds on OBS. This is something we would need to change in `obs-service-node_modules` although it really doesn't matter much as this script is only downloading and archiving the code and with this commit we do not install these packages.

Related tickets: https://progress.opensuse.org/issues/188877


perlpunk (approved):

kalikiana (approved):

Makefile
```diff
@@ -207,8 +207,11 @@ endif
 
 # Ensure npm packages are installed and up-to-date (unless local-npm-registry is installed; in this case we can
 # assume installing npm packages is taken care of separately, e.g. in builds on OBS)
+# note: Excluding dev dependencies like `eslint` via `--only=prod` to pull in only dependencies needed at runtime (and for
+#       regular tests). Development tests/tooling like `js-tidy` will invoke `npm install …` to install missing dependencies
+#       on its own anyway.
 node_modules: package-lock.json
-	@which local-npm-registry >/dev/null 2>&1 || npm install --no-audit --no-fund --ignore-scripts
+	@which local-npm-registry >/dev/null 2>&1 || npm install --no-audit --no-fund --ignore-scripts --only=prod
```
okurz: My `npm help install` says this would need to be

```suggestion
	@which local-npm-registry >/dev/null 2>&1 || npm install --no-audit --no-fund --ignore-scripts --production
```

Maybe `--only=prod` is some undocumented less common variant of the same?

Martchus: With `--production` npm says one should use `--omit=dev` instead. To specify it more positively I then decided to use `--only=prod`. I've just noticed that one then actually gets the same warning as with `--production` so I guess we should change this to `--omit=dev`.

PR #6763, opened 2025-09-22
Cleanup .gitignore files

perlpunk:
and order by topic

There are still some files generated by tests, which we could remove in the future.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,3 @@
-logs/*.tar.*
 etc/mine/
 # Devel::Cover
 /cover_db*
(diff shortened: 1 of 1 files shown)


.gitignore
```diff
@@ -1,56 +1,35 @@
-logs/*.tar.*
 etc/mine/
```
perlpunk, line 1: This folder is referenced here: https://open.qa/docs/#_customize_configuration_directory
So I left it in.

okurz (approved):

Martchus (approved):

PR #6764, opened 2025-09-23
Check for consistent solver.dupAllowVendorChange for auto/continuous-update

okurz:
The error message would look like this

```
devel:openQA looks bad for Leap  (x86_64):
/etc/zypp/zypp.conf needs to have '^solver.dupAllowVendorChange = true' for devel:openQA repositories. See https://progress.opensuse.org/issues/188235 for details
```

Related progress issue: https://progress.opensuse.org/issues/188235

### DIFF
--- a/script/openqa-check-devel-repo
+++ b/script/openqa-check-devel-repo
@@ -34,6 +34,10 @@ log_error() {
 
 [[ $VERSION ]] || . /etc/os-release
 [[ $ARCH ]] || ARCH=$(uname -i)
+if ! grep -q '^solver.dupAllowVendorChange\s*=\s*true' /etc/zypp/zypp.conf; then
+    result="/etc/zypp/zypp.conf needs to have '^solver.dupAllowVendorChange = true' for devel:openQA repositories. See https://progress.opensuse.org/issues/188235 for details"
+    false
+fi
 result="$(curl -sS "${REPO_URL:-https://api.opensuse.org/public/build/devel:openQA}/_result?repository=${VERSION}&arch=${ARCH}")"
 
 echo -e "$result" | grep -q "project.*state=.published."


script/openqa-check-devel-repo
```diff
@@ -34,6 +34,10 @@ log_error() {
 
 [[ $VERSION ]] || . /etc/os-release
 [[ $ARCH ]] || ARCH=$(uname -i)
+if ! grep -q '^solver.dupAllowVendorChange = true' /etc/zypp/zypp.conf; then
+   result="/etc/zypp/zypp.conf needs to have '^solver.dupAllowVendorChange = true' for devel:openQA repositories. See https://progress.opensuse.org/issues/188235 for details"
+   false
```
Martchus: Using the builtin `exit` function would be much more explicit:
```suggestion
   exit 1
```

script/openqa-check-devel-repo
```diff
@@ -34,6 +34,10 @@ log_error() {
 
 [[ $VERSION ]] || . /etc/os-release
 [[ $ARCH ]] || ARCH=$(uname -i)
+if ! grep -q '^solver.dupAllowVendorChange = true' /etc/zypp/zypp.conf; then
```
Martchus: The regex should probably allow different kinds or absence of white-spaces around the `=`.

okurz: So `\s*=\s*`?

Martchus: I suppose. I'm not sure about using regex in grep, though. (With `--perl-regexp` it is supposed to work, though.)

Maybe a trailing anchor (`$`) would make sense as well.

okurz: Well, I tested the regex manually on my system with grep and it matches as expected.

> Maybe a trailing anchor ($) would make sense as well.

I would not do that. Being more resilient than zypper regarding the whitespace is fine but being potentially more strict I see as out of responsibility of this check. If zypper parses

```
solver.dupAllowVendorChange   =truebutmaybenot  # …and a comment
```

as truthy value so should we :)

PR #6766, opened 2025-09-24
script: Allow to configure openQA mode

okurz:


script/openqa-webui-daemon
```diff
@@ -1,6 +1,7 @@
 #!/bin/bash
 set -e
 
+OPENQA_WEBUI_MODE="${OPENQA_WEBUI_MODE:-"production"}"
```
perlpunk:
```suggestion
OPENQA_WEBUI_MODE=${OPENQA_WEBUI_MODE:-"production"}
```

script/openqa-webui-daemon
```diff
@@ -16,7 +17,7 @@ function start_service {
 
     # start openQA in the background
     # note: Our API commands are very expensive, so the default timeouts are too tight.
-    "$openqa_dir"/openqa prefork -m production --proxy -i 100 -H 900 -w 30 -c 1 -G 800 -P "$pid_file" "${openqa_args[@]}" &
+    "$openqa_dir"/openqa prefork -m $OPENQA_WEBUI_MODE --proxy -i 100 -H 900 -w 30 -c 1 -G 800 -P "$pid_file" "${openqa_args[@]}" &
```
perlpunk:
```suggestion
    "$openqa_dir"/openqa prefork -m "$OPENQA_WEBUI_MODE" --proxy -i 100 -H 900 -w 30 -c 1 -G 800 -P "$pid_file" "${openqa_args[@]}" &
```

kalikiana (approved):

Martchus:
It would make sense to add this to all web UI services, e.g. also to the ws server, livehandler and scheduler (and perhaps even gru if applicable). Otherwise one ends up with a mixup of modes.

okurz:
> I mean one variable for all. Considering this affects authentication it could lead to authentication issues just affecting e.g. the ws server of livehandler. And I wouldn't want to debug issues like that. As-is this PR just introduced another way to create a likely misconfiguration (or at least an inconsistent configuration).

I can introduce the flag in the other services. But nothing ensures that the variable would be available in the scope of all services.

PR #6767, opened 2025-09-24
Allow to configure the worker host+key+secret as env variables

okurz:


okurz:
Changed to set the webui_host, not the host which is part of the specified webUI URL. Thx to @d3flex for pointing that out

Martchus (changes requested):
Have you tested this? I'm asking because I'm wondering whether this covers the websocket connection as well. It likely does but it would be good to test that before merging.

okurz:
> Have you tested this? I'm asking because I'm wondering whether this covers the websocket connection as well. It likely does but it would be good to test that before merging.

I have tested that the settings are passed into the correct module but I have not covered the complete connection flow. I plan to do that. But also I might reconsider the place where the read the env variable because other env variables are also read in a different place. Maybe better if we put those together

Wabri (approved):

PR #6768, opened 2025-09-24
Add support for worker authentication tokens

okurz:
Follow-up to
* #6767

TODO: Cover in documentation

### DIFF
--- a/lib/OpenQA/Worker/Settings.pm
+++ b/lib/OpenQA/Worker/Settings.pm
@@ -9,7 +9,7 @@ use Mojo::URL;
 use Mojo::Util 'trim';
 use Config::IniFiles;
 use Feature::Compat::Try;
-use MIME::Base64 qw(encode_base64url decode_base64url);
+use MIME::Base64 qw(decode_base64url);
 use Time::Seconds;
 use OpenQA::Config;
 use OpenQA::Constants qw(VNCPORT_OFFSET);
@@ -155,13 +155,6 @@ sub is_local_worker ($self) {
 sub has_class ($self, $worker_class) {
     my $c = $self->{_worker_classes} //= {map { $_ => 1 } split /,/, $self->global_settings->{WORKER_CLASS} // ''};
     return exists $c->{$worker_class};
-}
-
-# Generate a length-optimized, URL escaped, base64 encoded worker token string
-# with a prefix abbreviated for "openQA worker token"
-sub encode_token ($host, $key, $secret) {
-    my ($key_bin, $secret_bin) = map { pack 'H*', $_ } ($key, $secret);
-    return 'oqwt-' . encode_base64url join "\x00", $host, $key_bin, $secret_bin;
 }
 
 sub decode_token ($token) {
--- a/lib/OpenQA/Worker/Settings.pm
+++ b/lib/OpenQA/Worker/Settings.pm
@@ -8,6 +8,8 @@ use Mojo::File 'path';
 use Mojo::URL;
 use Mojo::Util 'trim';
 use Config::IniFiles;
+use Feature::Compat::Try;
+use MIME::Base64 qw(encode_base64url decode_base64url);
 use Time::Seconds;
 use OpenQA::Config;
 use OpenQA::Constants qw(VNCPORT_OFFSET);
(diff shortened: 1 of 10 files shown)


okurz:
@Martchus had the idea to put the env variables into the client so the deepest level. I think we can try to put the authentication token support into the client as well

lib/OpenQA/Worker/Settings.pm
```diff
@@ -45,6 +47,15 @@ sub new ($class, $instance_number = undef, $cli_options = {}) {
     my @hosts = split(' ', $cli_options->{host} || $global_settings{HOST} || 'localhost');
     delete $global_settings{HOST};
     _read_section($cfg, $_, $webui_host_specific_settings{$_} = {}) for @hosts;
+    my %tokens = map {
+        my ($host, $key, $secret) = decode_token($_);
+        $host => {key => $key, secret => $secret}
+    } split /,/, $ENV{OPENQA_WORKER_TOKEN} // $cli_options->{token} // '';
+    for (keys %tokens) {
+        $webui_host_specific_settings{$_} ||= {};
```
perlpunk, line 57:
```suggestion
```

lib/OpenQA/Worker/WebUIConnection.pm
```diff
@@ -25,12 +25,19 @@ has 'service_port_delta';    # delta from web UI port on which to directly conne
 # the websocket connection to receive commands from the web UI and send the status (Mojo::Transaction::WebSockets instance)
 has 'websocket_connection';
 
-sub new ($class, $webui_host, $cli_options) {
+
+
+sub new ($class, $webui_host, $args) {
+    $webui_host = $ENV{OPENQA_WORKER_WEBUI_HOST} // $webui_host;
     my $url = $webui_host !~ '/' ? Mojo::URL->new->scheme('http')->host_port($webui_host) : Mojo::URL->new($webui_host);
+    my ($host, $key, $secret) = split /|/, $ENV{OPENQA_WORKER_TOKEN} if $ENV{OPENQA_WORKER_TOKEN};
```
perlpunk:
```
perl -wE'@a = split /|/, "abc|de"; say for @a'
a
b
c
|
d
e
```
You want `split m/\|/`

script/worker
```diff
@@ -97,11 +107,13 @@ my %options;
...
 GetOptions(
-    \%options, "no-cleanup", "instance=i", "isotovideo=s", "host=s", "apikey:s",
-    "apisecret:s", "verbose|v|debug|d", "help|h",
+    \%options, "no-cleanup", "instance=i", "isotovideo=s",
+    "host=s", "apikey:s", "apisecret:s", "token:s",
+    "encode-token:s", "verbose|v|debug|d", "help|h",
 ) or usage(1);
 
 usage(0) if ($options{help});
+exit !OpenQA::Worker::encode_token($options{encode-token}) if $options{encode-token};
```
perlpunk:
```
Unquoted string "encode" may clash with future reserved word at -e line 1.
Argument "token" isn't numeric in subtraction (-) at -e line 1.
Argument "encode" isn't numeric in subtraction (-) at -e line 1.
```

lib/OpenQA/Worker/Settings.pm
okurz: so you think this line isn't necessary? Isn't that needed for auto-vivification so that we can write values into the inner hash?

script/worker
okurz: sorry, I missed some local changes for that

lib/OpenQA/Worker/Settings.pm
perlpunk: Auto-vivifaction is exactly why it isn't needed, that's why it's called "auto" :)

lib/OpenQA/Worker/WebUIConnection.pm
d3flex: Why not reuse the `decode_token` here?

script/worker
```diff
@@ -30,6 +30,16 @@ specify the public key needed for API authentication
 
 specify the secret key needed for API authentication
 
+=item B<--token> <value>
+
+specify a comma-separated list of tokens for C<host>@C<key>@C<secret>
+combinations as alternative to the individual options. Alternatively use the
+environment variable C<OPENQA_WORKER_TOKEN>.
+
+=item B<--encode-token> HOST@KEY@SECRET
```
d3flex, line 43: I think the advantage of using a token is also secure the `secret`. wouldnt make it more "secure" if you do not have to expose the secret in the terminal?

okurz: Well, this is only for `encode-token` so on the side of the client so that the complete token can be handled securily everywhere else. But keep in mind that this is only encoding the token, not encrypting it. So the token string is not "more secure" than the secret string

d3flex: yea thats why I quote the term `secure`. thanks for reply
I will resolve the thread and I see your point but I would like to bring the topic up again in the future.

PR #6771, opened 2025-09-25
Consider all jobs for cleanup, not just jobs that were executed

Martchus:
* Use `t_created` instead of `t_finished` to find expired jobs so e.g.
  scheduled, assigned and canceled jobs are also cleaned up if they are
  older than configured retentions
* This means jobs are now cleaned up slightly sooner than before but in
  practice it won't be a big difference, e.g. on OSD the median job age is
  20 minutes and the "long tail" (min value of worst case machine classes)
  is 3 hours. I don't think this difference is significant as we normally
  measure the cleanup retentions in days and not hours and we have a
  de-scheduling after 7 days (by default and also on OSD).
* See https://progress.opensuse.org/issues/189174


okurz:
now failures in "ui" and "unstable", possibly as consequence of your change to fixtures

okurz (approved):

lib/OpenQA/Schema/Result/JobGroups.pm
```diff
@@ -230,14 +230,14 @@ sub _find_expired_jobs ($self, $keep_in_days, $keep_important_in_days, $preserve
         {
             @not_important_cond,
             text => {like => 'label:linked%'},
-            t_finished => $timecond,
+            'me.t_created' => $timecond,
```
Wabri, line 233: question(non-blocking): who is 'me'?

Martchus: `me` is referring to the resultset (basically the table) being searched. (DBIx::Class generates SQL like `SELECT … from jobs AS me`.) This is used to disambiguate with the `t_created` columns from joined tables (in this case from the `comments` table).

perlpunk (approved):

PR #6772, opened 2025-09-25
Add requirement dependency via symlinks to webui service

d3flex:
Use a symlinks to set dependency to the openqa_webui as it was done in https://github.com/os-autoinst/openQA/pull/4976/files in order to restart the webui when the database is restarted.

It seems that something changed since
https://github.com/os-autoinst/openQA/pull/4976 which makes the webui to not behave as it was at the time of that commit and adding it now ensures that the sequence of restarts will take place as expected.

poo: https://progress.opensuse.org/issues/183302

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -172,6 +172,8 @@ install-generic: generate-assets
 	ln -s ../postgresql.service "$(DESTDIR)"/usr/lib/systemd/system/openqa-scheduler.service.requires/postgresql.service
 	install -d -m 755 "$(DESTDIR)"/usr/lib/systemd/system/openqa-websockets.service.requires
 	ln -s ../postgresql.service "$(DESTDIR)"/usr/lib/systemd/system/openqa-websockets.service.requires/postgresql.service
+	install -d -m 755 "$(DESTDIR)"/usr/lib/systemd/system/openqa-webui.service.requires
+	ln -s ../postgresql.service "$(DESTDIR)"/usr/lib/systemd/system/openqa-webui.service.requires/postgresql.service
 	install -D -m 644 usr/lib/sysctl.d/01-openqa-reload-worker-auto-restart.conf "$(DESTDIR)"/usr/lib/sysctl.d/01-openqa-reload-worker-auto-restart.conf
 #
 # install openQA apparmor profile
--- a/dist/rpm/openQA.spec
+++ b/dist/rpm/openQA.spec
@@ -614,6 +614,7 @@ fi
 # init
 %dir %{_unitdir}
 %{_unitdir}/openqa-webui.service
+%dir %{_unitdir}/openqa-webui.service.requires
 %{_unitdir}/openqa-livehandler.service
 %{_unitdir}/openqa-gru.service
 %dir %{_unitdir}/openqa-gru.service.requires
@@ -823,6 +824,7 @@ fi
 %{_unitdir}/openqa-gru.service.requires/postgresql.service
 %{_unitdir}/openqa-scheduler.service.requires/postgresql.service
 %{_unitdir}/openqa-websockets.service.requires/postgresql.service
+%{_unitdir}/openqa-webui.service.requires/postgresql.service
 %{_datadir}/openqa/script/setup-db
 %{_datadir}/openqa/script/dump-db
 %{_bindir}/openqa-setup-db


d3flex:
keep in mind that the change is already in o3 which is used to verify the behavior on postgresql restart

Martchus:
> The webUI needs to stay up as much as possible and needs to handle an unavailable database as gracefully as possible.

This change won't make it that much worse, though. Instead of getting a 500 error because the web UI ran into a database error it'll be a 503 or 504 because the web UI isn't running at all. Of course the restarting means that the time period in which the 503/504 errors will happen is slightly longer.

> For zero downtime the webui needs to: …

I'm wondering whether we could just `SIGSTOP` and `SIGCONT` all the web UI processes while PostgreSQL is unavailable to achieve this in a simpler way. (E.g. by adding `systemctl --kill-whom=all --signal=SIGSTOP openqa-webui.service` in `ExecStopPost=` of `postgresql.service` and the reverse for the start command.)

> Just a quick idea, this kind of sounds like you need a highly-available postgres cluster, i.e. handle postgres restarts outside of the openQA scope and decouple database and the rest of the stack.

Sounds like that would be the proper solution.

Martchus:
Actually, we don't even need SIGSTOP/SIGCONT for this. If I just restart the database during the sleep (and it probably doesn't behave differently when Perl code is executed) I get the same behavior. So a re-connect is logged but DBI is handling it gracefully under the hood. So considering that I'd say it even already behaves like what @okurz is asking for. I'm still not sure what happens if this occurs in the middle of a transaction, though.

So not sure under which conditions we'll run into  `failed: FATAL:  the database system is shutting down`. Locally, when shutting down the database while a request is processed I run into `failed: connection to server on socket`. The former is probably harder to reproduce locally because it likely only happens if the DB shutdown is dragged out for long enough so a query reaches the DB during the shutdown.

>> I suggest to first understand the impact of the original issue before we make it worse.
>
> I count 316 500 Server errors for the restart yesterday on osd.

What errors exactly? The `failed: no connection to the server` don't necessarily have a user-visible impact according to the tests I've just did locally.

perlpunk:
> What errors exactly? The `failed: no connection to the server` don't necessarily have a user-visible impact according to the tests I've just did locally.

Should I attach the according access and error log parts to the ticket? Do you think we simply had some unrelated server errors?

perlpunk:
For example I'm seeing those:
```
Sep 25 13:03:55 openqa openqa-webui-daemon[22426]:         (in cleanup) DBIx::Class::Storage::DBI::_populate_dbh(): DBI Exception: DBD::Pg::st DESTROY failed: no connection to the server [for Statement "INSERT INTO audit_events ( connection_id, event, event_data, t_created, t_updated, user_id) VALUES ( ?, ?, ?, ?, ?, ? ) RETURNING id"] at /usr/share/openqa/script/../lib/OpenQA/Shared/Controller/Auth.pm line 164
Sep 25 13:03:55 openqa openqa-webui-daemon[25588]:         (in cleanup) DBIx::Class::Storage::DBI::_populate_dbh(): DBI Exception: DBD::Pg::st DESTROY failed: no connection to the server [for Statement "INSERT INTO audit_events ( connection_id, event, event_data, t_created, t_updated, user_id) VALUES ( ?, ?, ?, ?, ?, ? ) RETURNING id"] at /usr/share/openqa/script/../lib/OpenQA/Shared/Controller/Auth.pm line 164
Sep 25 13:03:55 openqa openqa-webui-daemon[24293]:         (in cleanup) DBIx::Class::Storage::DBI::_populate_dbh(): DBI Exception: DBD::Pg::st DESTROY failed: no connection to the server [for Statement "INSERT INTO audit_events ( connection_id, event, event_data, t_created, t_updated, user_id) VALUES ( ?, ?, ?, ?, ?, ? ) RETURNING id"] at /usr/share/openqa/script/../lib/OpenQA/Shared/Controller/Auth.pm line 164
Sep 25 13:03:55 openqa openqa-webui-daemon[13900]:         (in cleanup) DBIx::Class::Storage::DBI::_populate_dbh(): DBI Exception: DBD::Pg::st DESTROY failed: no connection to the server [for Statement "SELECT me.name FROM job_modules me WHERE ( ( me.job_id = ? AND result = ? ) ) ORDER BY t_updated"] at /usr/share/openqa/script/../lib/OpenQA/Shared/Controller/Auth.pm line 164
Sep 25 13:03:55 openqa openqa-webui-daemon[11774]:         (in cleanup) DBIx::Class::Storage::DBI::_populate_dbh(): DBI Exception: DBD::Pg::st DESTROY failed: no connection to the server [for Statement "SELECT me.name FROM job_modules me WHERE ( ( me.job_id = ? AND result = ? ) ) ORDER BY t_updated"] at /usr/share/openqa/script/../lib/OpenQA/Shared/Controller/Auth.pm line 165
Sep 25 13:03:55 openqa openqa-webui-daemon[22861]:         (in cleanup) DBIx::Class::Storage::DBI::_populate_dbh(): DBI Exception: DBD::Pg::st DESTROY failed: no connection to the server [for Statement "INSERT INTO audit_events ( connection_id, event, event_data, t_created, t_updated, user_id) VALUES ( ?, ?, ?, ?, ?, ? ) RETURNING id"] at /usr/share/openqa/script/../lib/OpenQA/Schema/Result/Jobs.pm line 411
```
which look to me like those statements weren't committed, so we would lose data. It doesn't log the values for the placeholders, so it's hard to check if that's actually the case.

d3flex:
> > > I count 316 500 Server errors for the restart yesterday on osd.
> > 
> > 
> > restart of what? postgresql or openqa-webui?
> 
> postgresql restart. Between 25/Sep/2025:13:03:55 and 25/Sep/2025:13:03:57 there were 509 500 Server Errors in the webui.

That's also likely as many because the restart too longer time to come back. in o3 it was much faster

perlpunk:
> That's also likely as many because the restart too longer time to come back. in o3 it was much faster

How many 500s were there on o3 then?

okurz:
We have established by now that we don't have any proof for negative user-facing problems due to the database restarts and it might be just the error message in the log files while on higher levels DBIx or the openQA web UI handle database restarts already gracefully including potential HTTP errors returned to outside callers but error codes that are usually treated as a "try again later" hint. So we decided that this is not a good change as if we would forcefully restart the openQA web UI user facing services would be guaranteed to have a more severe effect.

PR #6773, opened 2025-09-25
Fully cover WebAPI::Auth::OpenID

Wabri:

### DIFF
--- a/lib/OpenQA/WebAPI/Auth/OpenID.pm
+++ b/lib/OpenQA/WebAPI/Auth/OpenID.pm
@@ -22,10 +22,8 @@ sub auth_login ($c) {
     );
 
     my $claimed_id = $csr->claimed_identity($c->config->{openid}->{provider});
-    if (!defined $claimed_id) {
-        log_error("Claiming OpenID identity for URL '$url' failed: " . $csr->err);
-        return;
-    }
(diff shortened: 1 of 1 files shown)


lib/OpenQA/WebAPI/Auth/OpenID.pm
```diff
@@ -93,6 +91,7 @@ sub auth_response ($c) {
       if ($c->app->config->{openid}->{httpsonly} && $url !~ /^https:\/\//);
     %params = map { $_ => URI::Escape::uri_unescape($params{$_}) } keys %params;
 
+    # TODO: why this is not cover? Since the other assignation is covered this should too.
```
Wabri: praise(blocking): debug sub is not covered

lib/OpenQA/WebAPI/Auth/OpenID.pm
```diff
@@ -22,10 +22,8 @@ sub auth_login ($c) {
     );
 
     my $claimed_id = $csr->claimed_identity($c->config->{openid}->{provider});
-    if (!defined $claimed_id) {
-        log_error("Claiming OpenID identity for URL '$url' failed: " . $csr->err);
-        return;
-    }
+    !defined $claimed_id and do { log_error("Claiming OpenID identity for URL '$url' failed: " . $csr->err); return; };
```
perlpunk, line 25: The `return` will only return from the `do` block and not from the subroutine

PR #6774, opened 2025-09-25
PoC: Report status to Gitea when scheduled product finishes

perlpunk:
Issue: https://progress.opensuse.org/issues/185491

So far only rewrote existing code to be able to deal with more than just github

### DIFF
--- a/lib/OpenQA/VcsProvider/Base.pm
+++ b/lib/OpenQA/VcsProvider/Base.pm
@@ -1,0 +1,49 @@
+# Copyright SUSE LLC
+# SPDX-License-Identifier: GPL-2.0-or-later
+
+package OpenQA::VcsProvider::Base;
+
+use Mojo::Base -base, -signatures;
+use Mojo::JSON qw(encode_json);
+use Mojo::URL;
+
+has 'app';
+has 'base_url';
+has 'statuses_url';
+
+sub add_params ($self, $params, $scheduled_product_id) {
+    $params->{context} //= 'openqa';
+    $params->{description} //= 'openQA test run';
+    my $base_url = $self->base_url;
+    $params->{target_url} //= "$base_url/admin/productlog?id=$scheduled_product_id"
+      if $scheduled_product_id && $base_url;
+}
+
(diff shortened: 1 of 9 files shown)


Martchus (commented):
Looks generally good. If cleaned up we could merge it as this is generally a good first step to supporting more than on VCS provider.

lib/OpenQA/VcsProvider/Base.pm
```diff
@@ -0,0 +1,49 @@
...
+      if $scheduled_product_id && $base_url;
+}
+
+sub create_request ($self, $params) {
+    my $app = $self->app;
+    my $ua = $app->ua;
+    # TODO Note that anyone who can create an openQA job can set settings
+    # with a webhook id and a statuses URL. Maybe we should configure the
+    # base url for each git provider and double check the url, because
+    # we are making an API request with a token to an otherwise unchecked URL
```
perlpunk, line 28: Just commenting to be able to link to this more easily

lib/OpenQA/VcsProvider/Gitea.pm
```diff
@@ -0,0 +1,27 @@
...
+    $self->base_url($settings->{CI_TARGET_URL});
+    return undef unless $self->statuses_url;
+    return 1;
+}
+
+sub create_request ($self, $params) {
+    my $tx = $self->SUPER::create_request($params);
+
+    my $headers = $tx->req->headers;
+    # TODO there might be more than one gitea server -> add sections?
```
perlpunk, line 19: Just commenting to be able to link to this more easily

lib/OpenQA/VcsProvider/Base.pm
Martchus: We should definitely add such a check.

lib/OpenQA/VcsProvider/Gitea.pm
Martchus: We probably need to be able to configure one token per instance, so yes we probably need multiple sections.

PR #6776, opened 2025-09-26
Enable automatic dependabot updates again after improvements

okurz:
This reverts commit 56709e76a5b156c7284596498aea3161ff5229e4.

To me it seems the situation has cooled down. We have reduced our attack
surface with --omit=dev for openQA and have reviewed other practices for
handling updates. In my understanding our previous approach of automated but
slightly delayed updates is still fine.

### DIFF
--- a/.mergify.yml
+++ b/.mergify.yml
@@ -19,7 +19,7 @@ pull_request_rules:
           - "#check-failure=0"
           - "#check-pending=0"
           - linear-history
-      - and: &manual_review
+      - and:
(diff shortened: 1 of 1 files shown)


Wabri (approved):
LGTM

Martchus (approved):
> In my understanding our previous approach of automated but
> slightly delayed updates is still fine.

Ok, then let's not manually approve dependabot PRs anymore by default.

PR #6777, opened 2025-09-26
Partial cover OpenQA::WebAPI::Plugin::ObsRsync::Task

Wabri:

### DIFF
--- a/t/api/14-plugin_obs_rsync.t
+++ b/t/api/14-plugin_obs_rsync.t
@@ -7,6 +7,8 @@ use FindBin;
 use lib "$FindBin::Bin/../lib", "$FindBin::Bin/../../external/os-autoinst-common/lib";
 use OpenQA::Test::TimeLimit '50';
 use Mojo::IOLoop;
+use Test::MockObject;
+
 use OpenQA::Test::Utils 'perform_minion_jobs';
 use OpenQA::Test::ObsRsync 'setup_obs_rsync_test';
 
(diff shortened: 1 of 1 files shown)


Wabri:
Failing due to: https://progress.opensuse.org/issues/189384

t/api/14-plugin_obs_rsync.t
```diff
@@ -124,4 +126,14 @@ subtest 'test lock after failure' => sub {
     lock_test();
 };
 
+subtest 'Finish job when retry count exceeded' => sub {
```
d3flex, line 129: sorry
if you check the other descriptions, all start with low case. Could you adjust this as well?

t/api/14-plugin_obs_rsync.t
```diff
@@ -124,4 +126,14 @@ subtest 'test lock after failure' => sub {
...
+    use_ok('OpenQA::WebAPI::Plugin::ObsRsync::Task');
+    my $finish_return;
+    my $job = Test::MockObject->new->mock('retries', sub { 42 })->set_true('retry')->set_true('finish')
+      ->mock('finish', sub { shift; $finish_return = [@_]; return 1 });
```
d3flex, line 133: I think this doesnt make sense. Because you set `finish` to true and right after you mock it again. 
In additional I do not think you need the `set_true('retry')`. AFAIU `$job->retry` it wont run based on the arguments ($job, undef, undef, 1, 1). `$retry_max_count` is true (aka false due to negation) and I think `$job->retries < $retry_max_count` is also false with those values.

PR #6779, opened 2025-09-28
Slightly simplify BuildResults

okurz:


lib/OpenQA/BuildResults.pm
```diff
@@ -175,16 +156,13 @@ sub compute_build_results ($group, $limit, $time_limit_days, $tags, $subgroup_fi
         $build->{key} = join('-', $version, $buildnr);
         $versions_per_build{$buildnr}->{$version} = 1;
     }
-    if ($buildver_sort_mode == BUILD_SORT_BY_NAME) {
-        @builds = reverse sort { versioncmp($a->{key}, $b->{key}); } @builds;
-    }
-
+    @builds = reverse sort { versioncmp($a->{key}, $b->{key}); } @builds
+      if $buildver_sort_mode == BUILD_SORT_BY_NAME;
```
d3flex, line 160: I have the impression that you could use https://github.com/okurz/openQA/blob/8dbe6d13612776e4c121845d1ae37125991995e4/lib/OpenQA/BuildResults.pm#L163 here. it seems like sort `unless $newest`

perlpunk (approved):

lib/OpenQA/BuildResults.pm
d3flex: in general looks good but this might be still valid.

PR #6780, opened 2025-09-28
README: Fix openQA badge after switch to UEFI

okurz:
With
https://github.com/os-autoinst/os-autoinst-distri-openQA/pull/254
the openQA-in-openQA tests switched to UEFI due to the changed default
in openSUSE Tumbleweed. Hence the badge must also not reference the
obsolete 64bit-2G. Instead of hardcoding uefi we can just leave out the
machine definition so that we look up whatever is "last".


Martchus (approved):

perlpunk (approved):

Wabri:
Reported https://progress.opensuse.org/issues/189384 about the docker-compose problem. Merging manually if there are no other checks missing looking at the mergify state.

PR #6781, opened 2025-09-29
Update openqa containers to Leap16

d3flex:
As a side effect this should fix also the problem with the postgres incompatibility.

### DIFF
--- a/container/devel:openQA:ci/base/Dockerfile
+++ b/container/devel:openQA:ci/base/Dockerfile
@@ -4,7 +4,7 @@
 # in one go
 # see https://progress.opensuse.org/issues/67855
 #!NoSquash
-FROM opensuse/leap:15.6
+FROM opensuse/leap:16.0
 
 # only dependencies for CircleCI to be able to load/save the package cache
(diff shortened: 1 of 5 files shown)


Martchus:
> Some packages are still missing in leap16, apparently

How did you figure this? All tests are passing. I'd almost approve your change :-)

container/devel:openQA:ci/base/Dockerfile
perlpunk, line 17: We're not using `rubygem(sass)` anymore in Leap 16.0, but `perl-CSS-Sass`.
I'm pretty sure we already have an according ticket to move to Leap 16.0, btw.

okurz (changes requested):
by the way https://progress.opensuse.org/issues/180716 is the epic about "support Leap 16.0" which also covers the work you and me mentioned about change to sass and such.  https://progress.opensuse.org/issues/180731 would be the specific ticket about "Upgrade CI container image versions to Leap 16.0"

I don't think we should move to 16.0 yet as we would loose coverage for what we consider the primary version also used in production.

PR #6785, opened 2025-09-29
Fix Docker Compose checks after updating to PostgreSQL 18

Martchus:
Related ticket: https://progress.opensuse.org/issues/189384

### DIFF
--- a/container/webui/docker-compose.yaml
+++ b/container/webui/docker-compose.yaml
@@ -124,14 +124,14 @@ services:
       start_period: 5s
 
   db:
-    image: docker.io/library/postgres:18
+    image: docker.io/library/postgres
     restart: always
     environment:
       POSTGRES_PASSWORD: openqa
       POSTGRES_USER: openqa
       POSTGRES_DB: openqa
     volumes:
-      - ./workdir/db:/var/lib/postgresql/18/docker:Z
+      - ./workdir/db:/var/lib/postgresql:Z
     healthcheck:
       test:
         - CMD-SHELL
--- a/docs/Contributing.asciidoc
+++ b/docs/Contributing.asciidoc
@@ -569,11 +569,11 @@ you use.
 ### host ###
 podman volume create mypostgres
 podman run -it --rm --name postgres-openqa \
-    -v mypostgres:/var/lib/postgresql/18/docker \
+    -v mypostgres:/var/lib/postgresql \
     -p 5432:5432 \
     -e POSTGRES_PASSWORD="x" \
     -d \
-    docker.io/library/postgres:18
+    docker.io/library/postgres
 
 # Optionally add this to mount a directory for sql dumps to import
 #    -v "$SQL_DUMP_DIR:/sqldumps" \


container/webui/docker-compose.yaml
```diff
@@ -124,14 +124,14 @@ services:
       start_period: 5s
 
   db:
-    image: postgres
+    image: docker.io/library/postgres:18
```
okurz: As we don't have automated version bums I prefer to keep the non-versioned reference. We just need to adapt this as necessary whenever a new version changes behaviour

Martchus: But we will still have to change the data path then. And note that the data path has been changed by upstream to be versioned now for a good reason: A migration of the data is necessary when moving from one major version to the next. We probably don't have good documentation covering this for any containerized setups (only for using openSUSE packaging) but it doesn't change the fact that it is necessary.

So I don't think it is useful to avoid specifying the version here. We'd just be lying about being able to use whatever version is the latest when this is not possible in practice. And when something breaks again you'll again create a ticket like https://progress.opensuse.org/issues/189384 which shows that the cause of these kinds of breakages are not always completely obvious. (If I manually bump the version beforehand and it then breaks the reason for the breakage is rather obvious, though.)

kalikiana: Maybe we can use a variable? Also for the actual path since 18 in the path is whatever version we use. We'll need to change that everywhere

kalikiana: Alternatively force the old name e.g. `PGDATA=/var/lib/postgresql/data` and use that instead

docs/Contributing.asciidoc
```diff
@@ -569,11 +569,11 @@ you use.
 ### host ###
 podman volume create mypostgres
 podman run -it --rm --name postgres-openqa \
-    -v mypostgres:/var/lib/postgresql/data \
+    -v mypostgres:/var/lib/postgresql/18/docker \
     -p 5432:5432 \
     -e POSTGRES_PASSWORD="x" \
     -d \
-    docker.io/library/postgres:17
+    docker.io/library/postgres:18
```
okurz: Also here I suggest to avoid hardcoding a version. As you already encountered this ages and outdates way too quickly. I wonder if we can find a version-agnostic mount approach

container/webui/docker-compose.yaml
okurz: The upstream change makes sense. But we should be able to just mount the parent directory. That would also ease migration because when mounting /var/lib/postgresql then container version $i saves in a subdir $i and a different container version $j saves in $j preserved in the data mounted into the container also allowing the migration from $i to $j

okurz (changes requested):

docs/Contributing.asciidoc
Martchus: I guess it is easier to get rid of the versions and mount the parent directory than arguing with you. In any case we don't have a good solution¹ for migrating the database when using a containerized setup.

¹ Or at least nothing is documented as far as I'm aware.

container/webui/docker-compose.yaml
Martchus: > Alternatively force the old name e.g. …

I would avoid that because the upstream change makes sense.

> But we should be able to just mount the parent directory.

I changed the setup to do that. And yes, maybe it makes migrations easier. Not sure how e.g. @perlpunk or the other team members who actually use a containerized PostgreSQL setup deal with this. (Feel free to create a PR extending the documentation or point me to existing documentation as this might need adjustment after my changes here.)

docs/Contributing.asciidoc
okurz: Well I see the two options: Either we run always the latest and live with the need to fix our setup as soon as it breaks due to upstream changes or we ensure to update to newer versions explicitly but triggered by automation as I don't see any manual "oh, we forgot to update this since years"-process working so I hope you don't see any argument with me "unreasonable".

> In any case we don't have a good solution¹ for migrating the database when using a containerized setup.

Yes we don't have that documented, only https://open.qa/docs/#_migrating_postgresql_database_on_opensuse . But considering that https://hub.docker.com/_/postgres/ just moved to a versioned directory also within the container *now* makes it seem like this is not something commonly done using a live migration. But now it's possible so the same approach as before is possible, the only part that differs is that one would not stop/start any local systemd services but stop/start the containers.

container/webui/docker-compose.yaml
perlpunk: > I changed the setup to do that. And yes, maybe it makes migrations easier. Not sure how e.g. @perlpunk or the other team members who actually use a containerized PostgreSQL setup deal with this.

If you mean how I deal with postgres upgrades, I said before that I so far never had the need to do this in my development setup. I either use a fresh database, or I use a database dump.
And otherwise, I guess a postgres dump and import would work fine.
I don't understand the problem.

Martchus:
The compose checks are working now so I'm merging this manually (with only the helm chart tests failing).

container/webui/docker-compose.yaml
okurz: > I don't understand the problem.

Well, the problem is merely that the upstream postgres container changed paths "to ease migration", regardless if one uses any migration or just uses a fresh database. Migration between database versions was of course possible in before, just different.

PR #6786, opened 2025-09-29
Update PostgreSQL version used in Helm charts

Martchus:
Maybe this helps with https://progress.opensuse.org/issues/188982.

### DIFF
--- a/container/helm/charts/openqa/Chart.lock
+++ b/container/helm/charts/openqa/Chart.lock
@@ -4,9 +4,9 @@ dependencies:
   version: 0.1.0
 - name: postgresql
   repository: https://charts.bitnami.com/bitnami
-  version: 15.5.38
+  version: 16.7.27
 - name: worker
   repository: file://../worker/
   version: 0.1.0
-digest: sha256:b8d33fe53bfb3cba746a17cbd13c29167fae37b44414b8769de3182c6a69adc6
-generated: "2025-06-10T15:13:53.917940726+02:00"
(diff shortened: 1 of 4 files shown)


perlpunk:
If I understand https://github.com/bitnami/containers/issues/83267 correctly, we can only use the latest version, not a specific version anymore.
But it seems to work, maybe because it is the latest version right now.

Martchus:
> But it seems to work, maybe because it is the latest version right now.

That was also a bit the idea behind this change and thus the version I put there is the latest¹. I don't know how to specify "latest" in these YAML files but that would probably be the best solution.

---

¹ Or at least the version mentioned on https://artifacthub.io/packages/helm/bitnami/postgresql?modal=install but that's probably the latest.

perlpunk:
hm, it did not work after all :(
https://github.com/os-autoinst/openQA/actions/runs/18099813187/job/51499580830?pr=6786#step:7:211
```
Failed to pull image "docker.io/bitnami/postgresql:17.6.0-debian-12-r4": rpc error: code = NotFound
```

perlpunk:
> I don't know how to specify "latest" in these YAML files

Maybe just `version: latest`?

Martchus:
I'm getting the same with the legacy repo now. I guess finding an exact constellation of Char and image that work together is taking too long now. So I'll create a PR disabling the check for now (so other PRs can at least be merged automatically again).

Martchus:
It still shows the same errors as in https://github.com/os-autoinst/openQA/pull/6786#issuecomment-3351113609. So I guess I can close this and we'll have to go with https://github.com/os-autoinst/openQA/pull/6788 for now.

PR #6788, opened 2025-09-30
Disable Helm Chart CI checks temporarily

Martchus:
We have to adapt to the change mentioned in the notice on https://github.com/bitnami/charts/blob/main/bitnami/postgresql/README.md before we can run this CI check again.

Related ticket: https://progress.opensuse.org/issues/188982

### DIFF
--- a/.mergify.yml
+++ b/.mergify.yml
@@ -13,7 +13,7 @@ pull_request_rules:
           # "unresolvable" is not reported in general:
           # https://trello.com/c/0N3jHq5M/2257-report-back-to-scm-when-build-results-arent-failed-and-succeeded
           # So we need to require the number of tests explicitly:
-          - "#check-success>=45"
+          - "#check-success>=51"
           - status-success=static-check-containers
           - status-success=webui-docker-compose
           - "#check-failure=0"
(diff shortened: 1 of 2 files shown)


Martchus:
I'll create a follow-up ticket with my findings and the options we have if this is merged.

.mergify.yml
```diff
@@ -13,7 +13,7 @@ pull_request_rules:
           # "unresolvable" is not reported in general:
           # https://trello.com/c/0N3jHq5M/2257-report-back-to-scm-when-build-results-arent-failed-and-succeeded
           # So we need to require the number of tests explicitly:
-          - "#check-success>=45"
+          - "#check-success>=44"
```
Martchus: It actually says 51 successful checks. This let's me think that we haven't updated this in a while and the current value of 45 is actually already too low. So should I put 51 here or do we have a reason for requiring fewer checks than we actually have?

Martchus: Looks like it was even me who put in this number: https://github.com/os-autoinst/openQA/commit/dbe84bc4548b68390a4de9a13a6ffdbf4b5eda85

Not sure how I cam to this conclusion so I'll try my luck with 51 now.

kalikiana: I guess we need to check what we get if all checks run. 51 looks to be correct in this PR 🤔

PR #6789, opened 2025-09-30
Ensure a failing cache service is seen as such by the worker/scheduler

Martchus:
The service `openqa-worker-cacheservice-minion.service` already fails quickly if running into some database errors. With this change the same applies to `openqa-worker-cacheservice.service`. It will now also fail if any kind of database error happens. To make this work in as most cases as possible the event loop is now stopped immediately.

With this the availability check of the worker can now detect a broken cache service when the disk where the cache directory is stored is broken (and hence database errors occurs). This way the worker will no longer try to run jobs that would only end up with a cache service error anyway.

Note that before this change `openqa-worker-cacheservice.service` would stay running but fail to process the unavailability of `openqa-worker-cacheservice-minion.service` and thus might keep telling the worker that there's a functioning Minion worker. Hence we had many jobs running into `Cache service enqueue error 500`, see https://progress.opensuse.org/issues/189300.

Note that the systemd services are configured for automatic restarts but in my tests the services nevertheless fail very quickly with messages like `openqa-worker-cacheservice.service: Start request repeated too quickly.` in cases like this. So regardless of this change there will also almost immediately be at least one failing systemd unit.

### DIFF
--- a/lib/OpenQA/CacheService.pm
+++ b/lib/OpenQA/CacheService.pm
@@ -55,7 +55,7 @@ sub startup {
         'reply.exception' => sub {
             my ($c, $error) = @_;
             $error = $c->$code($error)->stash('exception');
-            return unless $error =~ qr/(database disk image is malformed|no such (table|column))/;
+            return unless $error =~ qr/(database|no such (table|column))/;
 
             my $app = $c->app;
             $app->exit_code(1);    # ensure the return code is non-zero


t/25-cache.t
```diff
@@ -434,7 +434,7 @@ subtest 'cache directory is corrupted' => sub {
     my $t = Test::Mojo->new('OpenQA::CacheService');
     $t->app->log($log);
     $cache_log = '';
-    $t->post_ok('/enqueue')->status_is(500);
+    $t->post_ok('/enqueue');
```
okurz: This seems like removing test assertions, not adding additional assertions for the additional case we want to cover. Can you elaborate?

Martchus: When not stopping gracefully we might not get a reply here. 

I removed stopping gracefully because it wasn't working without preforking (and also might not work with preforking, I haven't tested that yet).

Martchus: I removed the change regarding the stopping as it isn't necessary when using prefork mode which we do in production.

perlpunk (approved):

Martchus:
Merging manually with 50 successful checks. (See https://github.com/os-autoinst/openQA/pull/6791 for fixing Mergify.)

PR #6790, opened 2025-09-30
Introduce array job settings

Martchus:
With this settings where the key ends with `[]` are split on `,`. So e.g. `ISSUES[]=foo,bar,baz` will be stored as `ISSUES[]=foo`, `ISSUES[]=bar` and `ISSUES[]=baz` in the database. So these settings now behave exactly like `WORKER_CLASS`.

This allows searching for jobs with a certain job setting value more efficiently. This can now be done via an API query like `/api/v1/job_settings/jobs?key=ISSUES[]&value=foo`. This query would also succeed with the values `bar` and `baz` for a job created with `ISSUES[]=foo,bar,baz`.

Note that such an array setting will still be returned in its joined form when doing other queries like `/api/v1/jobs/$id`.

Note that the suffix `[]` is still part of the setting key and always needs to be specified. So e.g. `ISSUES` and `ISSUES[]` are two completely different settings.

Related ticket: https://progress.opensuse.org/issues/188652


lib/OpenQA/WebAPI/Controller/API/V1/JobSettings.pm
```diff
@@ -15,31 +15,47 @@ Filters jobs based on a single setting key/value pair.
...
-    my $list_value = $validation->param('list_value');
-    my $jobs = $self->schema->resultset('JobSettings')->jobs_for_setting({key => $key, list_value => $list_value});
+    my $job_settings = $self->schema->resultset('JobSettings');
+    my $jobs = [];
+    if (defined(my $list_value = $validation->param('list_value'))) {
+        $jobs = $job_settings->jobs_for_setting({key => $key, list_value => $list_value});
+    }
+    elsif (defined(my $value = $validation->param('value'))) {
+        $jobs = [map { $_->job_id } $job_settings->search({key => $key, value => $value}, {columns => ['job_id']})];
+    }
```
okurz: This looks less pretty than before and might pose a problem in the "else all" branch. How about


```suggestion
    my $list_value, $value = $validation->param('list_value'), $validation->param('value');
    die "Found neither 'value' nor 'list_value'" unless defined $list_value || defined $value;
    my $jobs = defined $value ? [map { $_->job_id } $job_settings->search({key => $key, value => $value}, {columns => ['job_id']})] : $job_settings->jobs_for_setting({key => $key, list_value => $list_value});
```

Probably even nicer would be to put this into a method of `$job_settings` which can decide what to return based on the type. Something like this

```perl
# …/ResultSet/JobSettings.pm (?)
sub get_jobs_by_type ($key, $value, $list_value) {
    die "Found neither 'value' nor 'list_value'" unless defined $value || defined $list_value;
    return $self->jobs_for_setting({key => $key, list_value => $list_value}) if defined $list_value;
    return [map { $_->job_id } $self->search({key => $key, value => $value}, {columns => ['job_id']})] if defined $value;
}

# lib/OpenQA/WebAPI/Controller/API/V1/JobSettings.pm
sub jobs ($self) {
    …
    my $jobs = $job_settings->get_jobs_by_type($validation->param('key'), $validation->param('value'), $validation->param('list_value'));
}
```

okurz (changes requested):

lib/OpenQA/WebAPI/Controller/API/V1/JobSettings.pm
```diff
@@ -15,31 +15,47 @@ Filters jobs based on a single setting key/value pair.
...
+=item C<value>
+
+Returns all jobs where the value of C<key> is C<value>. This can be a string
+without special characters.
 
 =item C<list_value>
 
-The value to match for the given key. This can be a string without special characters.
+Returns all jobs where the value of C<key> contains a comma-separated list that
+in turn contains C<list_value>. This can be a string without special characters.
```
perlpunk: Why without special characters?

Martchus: This sentence was present before. I only changed the description to account for the new `value` parameter.

perlpunk: Sorry, I meant to comment above.

lib/OpenQA/WebAPI/Controller/API/V1/JobSettings.pm
```diff
@@ -15,31 +15,47 @@ Filters jobs based on a single setting key/value pair.
...
 
-The setting key to filter by. It accepts a string or a glob of a job variable.
-Some variables are not stored and they can not be used.
+The setting key to filter by. Some variables are not stored and they can not be
+used.
+
+=item C<value>
+
+Returns all jobs where the value of C<key> is C<value>. This can be a string
+without special characters.
```
perlpunk: Why without special characters?

perlpunk: I mean, for the list_value we have the limited use case of searching for numbers in a string with comma separated numbers, and we are using `like`, so we don't want people to insert `%` somewhere else, but for a simple `value = '...'` query I don't see why we should limit it

Martchus: Copied from the other setting but it is actually wrong.

okurz (approved):

lib/OpenQA/WebAPI/Controller/API/V1/JobSettings.pm
```diff
@@ -10,37 +10,58 @@ use Mojo::Base 'Mojolicious::Controller', -signatures;
...
     return $self->reply->validation_error({format => 'json'}) if $validation->has_error;
 
     my $key = $validation->param('key');
+    my $value = $validation->param('value');
     my $list_value = $validation->param('list_value');
-    my $jobs = $self->schema->resultset('JobSettings')->jobs_for_setting({key => $key, list_value => $list_value});
-    $self->render(json => {jobs => $jobs});
+    my $limit = $validation->param('limit');
+    return $self->render(json => {error => 'either "value" or "list_value" needs to be specified'}, status => 400)
+      unless defined($value) ^^ defined($list_value);
```
perlpunk: That's a syntax error (should be `^`). But I would also prefer `xor` here because ["Binary "^" returns its operands XORed together bit by bit."](https://perldoc.perl.org/perlop#Bitwise-Or-and-Exclusive-Or) and it's not intended for checking for truthiness, and `xor` is also a bit more readable.
```suggestion
      unless defined($value) xor defined($list_value);
```

Martchus: I was using the `^^` operator as documented on https://perldoc.perl.org/perlop#C-style-Logical-Or,-Xor,-and-Defined-Or.

> Binary "^^" performs a logical XOR operation.

It worked in my local tests. Maybe it has only been introduced in a recent version of Perl.

I avoided `xor` because tidy did some very ugly reformatting when it was used. I guess we'll have to live with that.

perlpunk: yeah, that's new in https://perldoc.perl.org/5.40.0/perldelta#New-%5E%5E-logical-xor-operator

lib/OpenQA/WebAPI/Controller/API/V1/JobSettings.pm
```diff
@@ -10,37 +10,59 @@ use Mojo::Base 'Mojolicious::Controller', -signatures;
...
 =item C<list_value>
 
-The value to match for the given key. This can be a string without special characters.
+Returns all jobs where the value of the setting specified by C<key> contains a
+comma-separated list that in turn contains C<list_value>. This can be a string
+without special characters. When C<list_value> is used, C<key> might contain an
+asterisks for globbing.
+
+Use of C<list_value> is expensive so the search is limited via the server-side
+setting C<{misc_limits}{job_settings_max_recent_jobs}>.
```
d3flex, line 38: `{misc_limits}{job_settings_max_recent_jobs}` doesnt seem to be used any more. And actually I cant find it anywhere else except openqa.ini as ref, Setup.pm as part of initialization and UserGuide. Should that be remove completely or I miss something?

d3flex (dismissed):

lib/OpenQA/WebAPI/Controller/API/V1/JobSettings.pm
perlpunk: I can see it used in jobs_for_settings

lib/OpenQA/WebAPI/Controller/API/V1/JobSettings.pm
```diff
@@ -10,37 +10,59 @@ use Mojo::Base 'Mojolicious::Controller', -signatures;
...
 =back
 
 =cut
 
 sub jobs ($self) {
     my $validation = $self->validation;
-    $validation->required('key')->like(qr/^[\w\*]+$/);
-    $validation->required('list_value')->like(qr/^\w+$/);
+    $validation->required('key')->like(qr/^[\w\*\[\]]+$/);
+    $validation->optional('value')->like(qr/^\w+$/);
```
perlpunk: The value is still restricted to \w here, you probably missed that?

lib/OpenQA/WebAPI/Controller/API/V1/JobSettings.pm
Martchus: It is still being used, I just did another search on this branch to verify.

Martchus:
I tested this locally with the OSD database and e.g. `script/openqa-cli api --host http://localhost:9526 cli api job_settings/jobs key="ISSUES[]" value=foo` is very fast with returning an empty set of jobs (even though *all* jobs/settings in the entire database were considered). But I guess we've already established this and it is just the index on `key`/`value` at work.

If one uses the extended route to query for a setting that returns tons of jobs like `script/openqa-cli api --host http://localhost:9526 job_settings/jobs key=BACKEND value=qemu` it takes a significant time but eventually returns without running into a timeout. With e.g. `limit=50` it is still very fast of course.

PR #6791, opened 2025-10-01
Reduce the number of required checks for Mergify again

Martchus:
On e.g. https://github.com/os-autoinst/openQA/pull/6790 there are just 50 checks but all packages have been built on OBS and were reported back. There were 51 checks on
https://github.com/os-autoinst/openQA/pull/6788 (where I took the 51 in my previous change from) but only because it changed the Mergify configuration and thus triggered the additional check "Configuration changed The new Mergify configuration is valid".

Note that one of the 50 checks is still the summary of Mergify itself but I guess the summary should be counted as well as it is a successful check (even if the merge conditions aren't satisfied) and always present.


Martchus:
And by the way, this PR is expected to have 51 checks itself - but also only because it changes the Mergify config.

PR #6795, opened 2025-10-06
parser: ktap: Fix SKIP case

rbmarliere:
Verification runs:
- https://rmarliere-openqa.qe.prg2.suse.org/tests/1599#step/selftests:cgroup:test_freezer/11 **before**
- https://rmarliere-openqa.qe.prg2.suse.org/tests/1604#step/selftests:cgroup:test_freezer/10 **after**

### DIFF
--- a/t/30-test_parser.t
+++ b/t/30-test_parser.t
@@ -967,7 +967,7 @@ subtest ktap_parse => sub {
     my $parser = OpenQA::Parser::Format::KTAP->new;
     $parser->load($ktap_file);
 
-    is $parser->results->size, 4, 'Expected four test groups';
+    is $parser->results->size, 5, 'Expected five test groups';
 
     my $result0 = $parser->results->get(0);
     is $result0->{name}, 'selftests:cgroup:test_core', 'The first test group has expected name';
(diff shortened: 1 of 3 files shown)


rbmarliere:
Good call. Added a new test case, which fails in the old code:

```
    #   Failed test 'Has 10 subtest details'
    #   at t/30-test_parser.t line 987.
    #          got: '17'
    #     expected: '10'
    # Looks like you failed 1 test of 9.
```

Martchus (approved):

t/30-test_parser.t
```diff
@@ -967,7 +967,7 @@ subtest ktap_parse => sub {
     my $parser = OpenQA::Parser::Format::KTAP->new;
     $parser->load($ktap_file);
 
-    is $parser->results->size, 4, 'Expected four test groups';
+    is $parser->results->size, 5, 'Expected five test groups';
```
perlpunk, line 970: As a general thing for the future, I recommend trying to leave the expected number out of the label, because then you only have to adjust it in one place.
e.g.

```suggestion
    is $parser->results->size, 5, 'Expected number of test groups';
```

PR #6797, opened 2025-10-07
Run `optipng` for real and handle errors if it fails

Martchus:
The `openQA-worker` package already has a hard requirement on `optipng` so it doesn't need to be adjusted. `optipng` also seems already available in the CI environment for full test coverage.

Related ticket: https://progress.opensuse.org/issues/189564

### DIFF
--- a/lib/OpenQA/Worker/Job.pm
+++ b/lib/OpenQA/Worker/Job.pm
@@ -873,14 +873,16 @@ sub _upload_results_step_2_1_upload_images ($self) {
 }
 
 my $OPTIMIZE_ERROR = 'Unable to optimize image:';
+my $OPTIMIZE_SUGGESTION = 'You may disable optimizations via OPTIMIZE_IMAGES=0.';
 
 sub _upload_results_step_2_2_upload_images ($self, $callback, $error) {
     $self->{_images_to_send} = {};
     $self->{_files_to_send} = {};
     return $callback->() unless $error;
     chomp $error;
+    $error =~ s/\s+(at.*line|OpenQA::).*//s;    # avoid too lengthy log message so $OPTIMIZE_SUGGESTION is not elided
     my $is_optimize_error = index($error, $OPTIMIZE_ERROR) == 0;
-    $error = "Unable to upload images: $error" unless $is_optimize_error;
+    $error = $is_optimize_error ? "$error - $OPTIMIZE_SUGGESTION" : "Unable to upload images: $error";
     log_error($self->{_result_upload_error} = $error);
     return $callback->() unless $is_optimize_error;
     $self->{_result_upload_internal_error} = $error;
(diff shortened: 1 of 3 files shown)


t/24-worker-jobs.t
```diff
@@ -1219,12 +1220,16 @@ subtest 'Dynamic schedule' => sub {
...
 
-subtest 'optipng' => sub {
-    is OpenQA::Worker::Job::_optimize_image('foo', {}), undef, 'optipng call is "best-effort"';
-};
+subtest 'image optimization' => sub {
+    my $opt = \&OpenQA::Worker::Job::_optimize_image;
+    is $opt->('foo', {OPTIMIZE_IMAGES => 0}), 0, 'image optimization can be skipped';
+
+    throws_ok { $opt->('foo', {}) }
+    qr/(Failed to execute optipng|optipng exited with non-zero return code)/, 'failing to run optipng is a hard error';
```
perlpunk: Why the alternation here? Does the result depending on where this test is run?

Martchus: Yes, it depends on whether optipng is completely missing or whether it is present but runs into an error. I guess we don't really care in this test so I just made it accept both error messages.

Martchus:
With the 2nd commit the jobs now stopped as soon as the error occurs (usually when isotovideo is already at the 2nd module):

screenshot_20251008_133651

Logs and other result files are still uploaded.

lib/OpenQA/Worker/Job.pm
```diff
@@ -1222,19 +1222,19 @@ sub _log_snippet ($self, $file, $offset_name) {
...
-        no warnings;
-        if ($job_settings->{USE_PNGQUANT}) {
-            system('pngquant', '--force', '--output', $image, $image);
-        }
-        else {
-            system('optipng', '-quiet', '-o2', $image) if ($? == -1);
-        }
-    }
-    return undef;
+    my @params
```
d3flex: Can we rename the `@params` to `@cmd` or `@bin`?

lib/OpenQA/Worker/Job.pm
```diff
@@ -1222,19 +1232,19 @@ sub _log_snippet ($self, $file, $offset_name) {
     return \%ret;
 }
 
-sub _optimize_image ($image, $job_settings) {
+sub _optimize_image ($image, $job_settings, $optipng_bin = 'optipng', $pngquant_bin = 'pngquant') {
+    my $optimize_flag = $job_settings->{OPTIMIZE_IMAGES};
```
okurz, line 1238: This flag is never documented. Can you put that into the error message as in "optipng failed. You can set OPTIMIZE_IMAGES=0 to disable"

okurz (changes requested):

PR #6804, opened 2025-10-15
Use direct link to subdomain for the test assets

d3flex:
Replace the `url_for with helper functions which return the domain name whenever this is applicable based on the setting.

- Link to domain for download assets like iso and `Uploaded logs` (includes also vars.json)
- No change in the redirection, assuming that this is out of scope here, and that in general it stays despite the direct link in the href.

issue: https://progress.opensuse.org/issues/189888

### DIFF
--- a/lib/OpenQA/WebAPI/Plugin/Helpers.pm
+++ b/lib/OpenQA/WebAPI/Plugin/Helpers.pm
@@ -381,6 +381,27 @@ sub register ($self, $app, $config) {
             }
             return $regex_problem && $context ? "$context: $regex_problem" : $regex_problem;
         });
+
+    $app->helper(
+        asset_url => sub ($c, $asset, $testid) {
+            my $url
+              = $c->url_for('test_asset_name', testid => $testid, assettype => $asset->type, assetname => $asset->name);
+            return _domain_url_for($c, $url);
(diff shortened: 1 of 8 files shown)


okurz (changes requested):
https://progress.opensuse.org/issues/189888 says "to simplify our implementation" but you add a net 21 lines and miss coverage. How does that match?

Martchus:
And probably we even still need the redirect to support old links which might still be present/used on places we don't control.

So I don't think this can simplify this aspect for now but the PR as-is does improve efficiency.

perlpunk:
> And probably we even still need the redirect to support old links which might still be present/used on places we don't control.
> 
> So I don't think this can simplify this aspect for now but the PR as-is does improve efficiency.

why do we need to redirect these links?
we just deliver the requested file under the regular domain, with the only difference that we make sure it is downloaded and not rendered. which is code which we already have in place in case there is no file domain configured

Martchus:
> we just deliver the requested file under the regular domain, with the only difference that we make sure it is downloaded and not rendered.

If we make sure it is downloaded then we can remove the redirection.

perlpunk:
I assume this can be closed, See #6810

PR #6805, opened 2025-10-16
Assure git_clone is only run for valid targets

r-richardson:
Previously we attempted git_clone including get_origin_url on non-repos or undefined remotes, causing failures before outage handling kicks in.

This PR makes it so we verify repos/URLs first.

Also i sringified the error object in order to make the regex check more reliable.

Related progress issue: https://progress.opensuse.org/issues/190410

### DIFF
--- a/lib/OpenQA/Task/Git/Clone.pm
+++ b/lib/OpenQA/Task/Git/Clone.pm
@@ -53,16 +53,26 @@ sub _git_clone_all ($job, $clones) {
         die "Don't even think about putting '..' into '$path'." if $path =~ /\.\./;
 
         my $git = OpenQA::Git->new(app => $app, dir => $path);
-        my $origin_url = -d $path ? $git->get_origin_url : undef;
-        my $url = $origin_url // $clones->{$path};
-        my $server_host = _extract_server_host($url);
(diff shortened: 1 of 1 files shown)


okurz (changes requested):
s/i sringified/I stringified/

lib/OpenQA/Task/Git/Clone.pm
```diff
@@ -53,16 +53,26 @@ sub _git_clone_all ($job, $clones) {
...
+                $ctx->info("Skipping '$path': no URL configured and not a git repo.");
+                next;
+            }
+
+            _git_clone($git, $ctx, $path, $url_cfg, $origin_url);
             report_server_available($app, $server_host);
             next;
         }
-        catch ($e) { $error = $e }
+        catch ($e) { $error = "$e" }
```
perlpunk, line 75: In perl, if applied to a regex, a variable is automatically stringified, so that should not be necessary

perlpunk:
And the according ticket for the "Failed to get origin url (/var/lib/openqa/share/tests): fatal: not a git repository" is actually https://progress.opensuse.org/issues/190686

PR #6810, opened 2025-10-20
Directly link to file domain without redirect

perlpunk:
This reverts #6793, #6759 (exceptions from redirection).
Then I amend #6804 to only keep the `log_url` helper, and in the next commit I only use it for files generated by the test, as files generated by openQA itself should be safe.
In the last commit I remove the redirection and replace it by serving the files as attachments if necessary.

Issue: https://progress.opensuse.org/issues/189888

Probably some tests will fail


templates/webapi/test/result_file_list.html.ep
```diff
@@ -17,14 +17,14 @@
         % }
         % else {
             <i class="fa fa-file-o"></i>
-            <a href="<%= url_for('test_file', testid => $testid, filename => $resultfile) %>">
+            <a href="<%= $userfiles ? log_url($testid, $resultfile) : url_for('test_file', testid => $testid, filename => $resultfile) %>">
```
okurz: I suggest to abstract that ternary into a plugin function with clear name as well. Or rather move the logic into `log_url` itself so that this would look like


```suggestion
            <a href="<%= log_url($testid, $resultfile, $is_userfiles) %>">
```

t/ui/07-file.t
```diff
@@ -150,13 +150,6 @@ subtest 'redirection to different domain' => sub {
 
     $t->get_ok('/tests/99961/asset/repo/testrepo/README')->status_is(302);
     $t->header_like(Location => qr|^//openqa-files(:\d+)?/assets/repo/testrepo/README$|);
-
-    # no redirect for images and thumbs
-    $t->get_ok('/tests/99938/images/logpackages-1.png')->status_is(200)->content_type_is('image/png')
-      ->header_is('Content-Length' => '48019');
-    $t->get_ok('/tests/99938/images/thumb/logpackages-1.png')->status_is(200)->content_type_is('image/png')
-      ->header_is('Content-Length' => '6769');
-    $t->get_ok('/image/347/da6/.thumbs/61d0c3faf37d49d33b6fc308f2.png')->status_is(200)->content_type_is('image/png');
```
okurz, line 159: I guess we still have some expectations about paths which are not redirected, isn't it?

lib/OpenQA/WebAPI/Controller/File.pm
```diff
@@ -164,7 +159,11 @@ sub _set_headers ($self, $path) {
                 $headers->header('X-Content-Type-Options', 'nosniff');
                 $is_text = 1;
             }
-            my $allow_insecure = $self->app->config->{global}->{file_security_policy} ne 'download-prompt';
+            my $file_security_policy = $self->app->config->{global}->{file_security_policy};
+            my $allow_insecure = $file_security_policy eq 'insecure-browsing';
+            if ($file_security_policy =~ m/^domain:/ and my $file_domain = $self->app->config->{global}->{file_domain}) {
+                $allow_insecure = 1 if $file_domain eq $self->req->url->to_abs->host_port;
```
d3flex, line 166: maybe you can skip some assignments here. And use directly `$headers->content_disposition("attachment; filename=$filename;")`. 
To me it looks like 
```perl
if ($file_security_policy =~ m/^domain:/ and my $file_domain = $self->app->config->{global}->{file_domain}) {
    if $file_domain eq $self->req->url->to_abs->host_port || $filetype !~ m/(html|svg)/) && $ext ne 'iso' {
        $headers->content_disposition("attachment; filename=$filename;");}}```

t/ui/07-file.t
okurz: Nevermind

lib/OpenQA/WebAPI/Controller/File.pm
perlpunk: I don't see how that helps.
First, the logic seems to be reversed (if the current domain is the file domain, serve as attachment -> no, it should not be served as an attachment in that case).
Second, why do all this and then repeat calling `$headers->content_disposition(...)` below? Note that this needs to be called even if `if (my $filetype = $self->app->types->type($ext))` is false.
Give me a full patch instead so I can better see what you mean. As it is, it doesn't make sense to me.

templates/webapi/test/result_file_list.html.ep
perlpunk: done

lib/OpenQA/WebAPI/Controller/File.pm
d3flex: ok I see. I was not clear. but I dont think it is worth to hold it on. It is good enough as is, so I will resolve and approve. thanks for looking on that anyway.

d3flex (approved):

PR #6812, opened 2025-10-20
Avoid creation of git_clone tasks for jobs with empty DISTRI

Martchus:
These jobs would otherwise just fail with:

```
Failed to get origin url (/var/lib/openqa/share/tests): fatal: not a git repository (or any parent up to mount point /var/lib/openqa)
```

Related ticket: https://progress.opensuse.org/issues/190686


okurz (commented):
+1. Need to fix CI tests

perlpunk (approved):

okurz (approved):

PR #6813, opened 2025-10-21
Emit restart events from _hook_script

d3flex:
issue: https://progress.opensuse.org/issues/190557

### DIFF
--- a/lib/OpenQA/Task/Job/HookScript.pm
+++ b/lib/OpenQA/Task/Job/HookScript.pm
@@ -4,6 +4,7 @@
 package OpenQA::Task::Job::HookScript;
 use Mojo::Base 'Mojolicious::Plugin', -signatures;
 use OpenQA::Task::SignalGuard;
+use OpenQA::Events;
 
 sub register ($self, $app, $config) {
     $app->minion->add_task(hook_script => \&_hook_script);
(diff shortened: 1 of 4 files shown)


lib/OpenQA/Task/Job/HookScript.pm
```diff
@@ -23,6 +24,7 @@ sub _hook_script ($job, $hook, $openqa_job_id, $options) {
 
     if ($retries && $skip_rc && defined $rc && $rc == $skip_rc && $job->retries < $retries) {
         $job->app->log->debug("Job @{[$job->id]}: Retrying @{[$job->retries]} times with linear delay: ${delay}s");
+        $job->app->events->emit('openqa_hook_script_job_restart', [{job_id => $openqa_job_id, minion_id => $job->id}]);
```
perlpunk, line 27: That's the wrong place. There is no openQA job restart happening here. That's the hook script called from the finalize_job_results task. It runs investigation scripts for example.

In the ticket I linked to
https://github.com/os-autoinst/openQA/blob/a802d1dd235bc62d964281a687097c20b0f46cd4/lib/OpenQA/Task/Job/Restart.pm and https://github.com/os-autoinst/openQA/blob/a802d1dd235bc62d964281a687097c20b0f46cd4/lib/OpenQA/Schema/Result/Jobs.pm#L2051

PR #6816, opened 2025-10-22
Emit restart events when a job is enqueued

d3flex:
With this commit once the job goes to its end, it will check if it can restarted and if it fulfills the criteria, it will send the new job to the minion queue via `enqueue_restart`, where now it should populate the event.

issue: https://progress.opensuse.org/issues/190557

### DIFF
--- a/lib/OpenQA/Task/Job/Restart.pm
+++ b/lib/OpenQA/Task/Job/Restart.pm
@@ -18,9 +18,8 @@ sub restart_delay { $ENV{OPENQA_JOB_RESTART_DELAY} // 5 }
 sub restart_openqa_job ($minion_job, $openqa_job) {
     my $cloned_job_or_error = $openqa_job->auto_duplicate;
     my $is_ok = ref $cloned_job_or_error || $cloned_job_or_error =~ qr/(already.*clone|direct parent)/i;
-    if ($is_ok) {
-        my $openqa_job_id = $openqa_job->id;
-        my %event_data = (id => $openqa_job_id, result => $cloned_job_or_error, auto => 1);
+    if (ref $cloned_job_or_error) {
+        my %event_data = (id => $openqa_job->id, result => $cloned_job_or_error->{cluster_cloned}, auto => 1);
         OpenQA::Events->singleton->emit_event('openqa_job_restart', data => \%event_data);
     }
 
@@ -64,12 +63,8 @@ sub _init_amqp_plugin ($app) {
 
 sub _wait_for_event_publish ($app) {
     return undef unless $app->config->{amqp}->{enabled};
-    OpenQA::Events->singleton->once(
-        'amqp_handled',
-        sub {
-            Mojo::IOLoop->singleton->stop;
-        });
-    Mojo::IOLoop->singleton->start;
+    OpenQA::Events->singleton->once(amqp_handled => sub { Mojo::IOLoop->stop });
+    Mojo::IOLoop->start;
 }
 
 1;
--- a/t/10-jobs.t
+++ b/t/10-jobs.t
@@ -21,9 +21,9 @@ use Test::MockModule 'strict';
 use Test::Mojo;
 use Test::Output;
 use Test::Warnings qw(:report_warnings warning);
-use Mojo::File 'path';
+use Mojo::File qw(path tempdir);
 use Mojo::JSON qw(decode_json encode_json);
-use OpenQA::Test::Utils qw(perform_minion_jobs);
+use OpenQA::Test::Utils qw(perform_minion_jobs mock_io_loop);
 use OpenQA::Test::TimeLimit '30';
 
 binmode(STDOUT, ':encoding(UTF-8)');
@@ -933,30 +933,56 @@ subtest 'job setting based retriggering' => sub {
     is $lastest_job->latest_job->id, $lastest_job->id, 'found the latest job from latest job itself';
 };
 
-subtest 'AMQP event emission for minion restarts' => sub {
-    my $events_module = Test::MockModule->new('OpenQA::Events');
-    my @emitted_events;
-    $events_module->redefine(
-        emit_event => sub ($self, $event_type, %args) {
-            push @emitted_events, {type => $event_type, data => $args{data}};
-            $events_module->original('emit_event')->($self, $event_type, %args);
+subtest 'AMQP event emission for job restarts within Minion tasks' => sub {
+    my $plugin_mock = Test::MockModule->new('OpenQA::WebAPI::Plugin::AMQP');
+    my $conf = "[global]\nplugins=AMQP\n[amqp]\npublish_attempts = 2\npublish_retry_delay = 0\n";
+    my $tempdir = tempdir;
+    path($ENV{OPENQA_CONFIG} = $tempdir)->make_path->child('openqa.ini')->spew($conf);
+    my %published;
+    my @event_body;
+    my $io_loop_mock = Test::MockModule->new('Mojo::IOLoop');
+    $io_loop_mock->redefine(start => sub { });
+    $plugin_mock->redefine(
+        publish_amqp => sub ($self, $topic, $data) {
+            $published{$topic} = $data;
+            Mojo::IOLoop->next_tick(sub { OpenQA::Events->singleton->emit('amqp_handled'); });
         });
 
+    my $events_mock = Test::MockModule->new('OpenQA::Events');
+    $events_mock->redefine(
+        emit => sub ($self, $type, $args) {
+            if ($type eq 'openqa_job_restart') {
+                @event_body = ($type, $args);
+            }
+            $events_mock->original('emit')->($self, $type, $args);
+        });
+    # use a new app; otherwise it runs slowly, maybe trying to run none-mocked stuff
+    my $t = Test::Mojo->new('OpenQA::WebAPI');
     my $minion = $t->app->minion;
-    my %_settings = %settings;
-    $_settings{TEST} = 'test_restart';
-    $_settings{RETRY} = '1';
-    my $job = _job_create(\%_settings);
+    is $t->app->config->{amqp}->{enabled}, 1, 'AMQP enabled from config file';
+
+    my %retry_settings = %settings;
+    $retry_settings{TEST} = 'test_amqp_restart';
+    $retry_settings{RETRY} = '1';
+    my $job = $jobs->create_from_settings(\%retry_settings);
+    $job->discard_changes;
     my $job_id = $job->id;
 
-    @emitted_events = ();
+    %published = ();
     $job->done(result => FAILED);
     stdout_like { perform_minion_jobs($minion) } qr/Job \d+ duplicated as \d+/;
-    my @restart_events = grep { $_->{type} eq 'openqa_job_restart' } @emitted_events;
-    is scalar(@restart_events), 1, 'exactly one job restart event emitted';
-    my $event_data = $restart_events[0]->{data};
-    is $event_data->{id}, $job_id, 'event contains original job ID';
-    ok exists($event_data->{result}), 'event contains result';
+    is scalar keys %published, 1, 'exactly one job restart event emitted';
+    my $event = $published{'suse.openqa.job.restart'};
+    is $event->{id}, $job_id, 'event contains original job ID';
+    is $event->{auto}, 1, 'event marked as auto restart';
+    my ($user_id, $connection, $type, $data) = @{$event_body[1]};
+    is $user_id, undef, 'user_id is undef for Minion restart';
+    is $connection, undef, 'connection is undef for Minion restart';
+    is $type, 'openqa_job_restart', 'type matches event type';
+    is_deeply $data, $event, 'published event equals emitted event';
+    ok exists $data->{result}->{$job_id}, 'old job id is in result';
+    $job->discard_changes;
+    is $job->clone_id, $data->{result}->{$job_id}, 'clone_id points to reported id';
 };
 
 subtest '"race" between status updates and stale job detection' => sub {


d3flex:
> code looks good so far. I suggest you do a manual system test with a local openQA instance and ensure that no duplicate events are sent.
> 
> Also in your commit message I don't quite understand "With this commit once the job goes to its end, it will check if it can restarted and if it fulfills the criteria". What "end" and what "criteria"? Also "check if it can restarted" is grammatically incorrect

image

perlpunk:
> With this commit once the job goes to execute its final stage, it will
check if it can be restarted. If yes, it will send the new job to
the minion queue via `enqueue_restart`, where now it should populate
the event.

The only change in this commit is that, if the job is restarted via `enqueue_restart`, it now publishes a restart event.
The description sounds overly complicated ("goes to execute its final stage").
" it will check if it can be restarted" - that's already happening without your commit.

How about: "When a job is enqueued to be restarted via the `RETRY` feature, it now publishes a `job.restart` event."

perlpunk:
When a job is enqueued to be restarted as part of the `$job->done` handling, a `job.restart` event is now published

t/23-amqp.t
```diff
@@ -334,4 +334,16 @@ subtest 'extra TLS query params' => sub {
...
 
+subtest 'enqueue_restart triggers amqp event and log' => sub {
+    my $test_job = $schema->resultset('Jobs')->find(99945);
+    my $minion_id;
+    %published = ();    # just to be sure
+    combined_like { $minion_id = $test_job->enqueue_restart }
+    qr/Enqueued restarting openQA job 99945 via Minion job \d+/, 'Debug log for enqueue_restart is present';
+    ok $minion_id, 'Got a Minion job ID back';
+    # each event contains keys(headers args body)
+    is scalar(%published), 3, 'Exactly one event was published';
```
Martchus: The number and the description are inconsistent.

t/23-amqp.t
```diff
@@ -334,4 +334,16 @@ subtest 'extra TLS query params' => sub {
...
 };
 
+subtest 'enqueue_restart triggers amqp event and log' => sub {
+    my $test_job = $schema->resultset('Jobs')->find(99945);
+    my $minion_id;
+    %published = ();    # just to be sure
+    combined_like { $minion_id = $test_job->enqueue_restart }
+    qr/Enqueued restarting openQA job 99945 via Minion job \d+/, 'Debug log for enqueue_restart is present';
+    ok $minion_id, 'Got a Minion job ID back';
+    # each event contains keys(headers args body)
```
Martchus: I would remove this comment. If this is important it should be checked with code.

lib/OpenQA/Schema/Result/Jobs.pm
```diff
@@ -2050,6 +2050,7 @@ sub enqueue_restart ($self, $options = {}) {
     my $openqa_job_id = $self->id;
     my $minion_job_id = OpenQA::App->singleton->gru->enqueue(restart_job => [$openqa_job_id], $options)->{minion_id};
     log_debug "Enqueued restarting openQA job $openqa_job_id via Minion job $minion_job_id";
+    OpenQA::App->singleton->emit_event(openqa_job_restart => {id => $openqa_job_id});
```
Martchus: I'm not sure whether this is the right place. Here the job hasn't actually been restarted yet because the according Minion job has only been enqueued. I would prefer to emit the event only when the restart has actually successfully happened.

Martchus (changes requested):

t/23-amqp.t
d3flex: I will answer in your both comments here.
The amqp event contains three things, right? It has a headers, a body and the args. This is one amqp and the description is emphasize to that. the comment+ the description intended to make that clear. 
the importance is that there is one event and not what is in this.

Martchus: Maybe just remove the comment then. It doesn't help me understand the test code better. (And it lacks punctuation.)

okurz (approved):

lib/OpenQA/Schema/Result/Jobs.pm
```diff
@@ -1001,6 +1002,7 @@ sub auto_duplicate ($self, $args = {}) {
     $dup->{cluster_cloned} = {map { $_ => $clones->{$_}->{clone} } keys %$clones};
     $dup->{comments_created} = $args->{comments};
     log_debug("Job $job_id duplicated as $clone_id");
+    OpenQA::Events->singleton->emit_event('job_restart', data => {id => $clone_id});
```
perlpunk: `auto_duplicate` is also called when doing a restart via API or web UI, and in that case the code emits an event already, see 
https://github.com/os-autoinst/openQA/blob/master/lib/OpenQA/WebAPI/Controller/API/V1/Job.pm#L821
So we would get two events.

Martchus: Yes, so please take care we don't get two events.

t/20-stale-job-detection.t
```diff
@@ -43,9 +44,24 @@ subtest 'worker with job and not updated in last 120s is considered dead' => sub
     $workers->update_all({t_seen => undef});
     is($jobs->stale_ones->count, 3, 'jobs considered stale if t_seen is not set');
 
+    my @emitted_events;
+    my $mock_events = Test::MockModule->new('OpenQA::Events');
+    $mock_events->redefine(
+        emit_event => sub {
+            my ($self, $type, %args) = @_;
```
Martchus: Use signatures in new code.

lib/OpenQA/Task/Job/Restart.pm
```diff
@@ -17,6 +17,7 @@ sub restart_delay { $ENV{OPENQA_JOB_RESTART_DELAY} // 5 }
 sub restart_openqa_job ($minion_job, $openqa_job) {
     my $cloned_job_or_error = $openqa_job->auto_duplicate;
     my $is_ok = ref $cloned_job_or_error || $cloned_job_or_error =~ qr/(already.*clone|direct parent)/i;
+    OpenQA::Events->singleton->emit_event('job_restart', data => {id => $openqa_job->id}) if $is_ok;
```
Martchus: To make this actually work we would probably need to make sure the AMQP plugin is loaded in the Gru service. It *probably* isn't so emitting this event probably has no effect. Now that I think about it, running the AMQP plugin within the Gru service and the processes it forks to run the particular Minion jobs is probably not so easy and could end up becoming quite a mess.

t/api/18-job-amqp-event.t
```diff
@@ -0,0 +1,55 @@
...
+    my %params = (TEST => 'event_test', DISTRI => 'foo', VERSION => 'bar', ARCH => 'baz');
+    $t->post_ok('/api/v1/jobs', form => \%params)->status_is(200);
+    my $job_id = $t->tx->res->json->{id};
+    is(scalar @emitted_events, 1, 'exactly one event was emitted on create');
+    is($emitted_events[0]->{data}{id}, $job_id, 'job_id is emitted on event');
+    is($emitted_events[0]->{type}, 'openqa_job_create', 'data has correct event type');
+
+    @emitted_events = ();
+    combined_like { $t->post_ok('/api/v1/jobs/99926/restart?force=1')->status_is(200) } qr/Job 99926 duplicated/,
+      'Job restarted successfully';
```
Martchus: This route is not what your ticket is about. Of course it makes also sense to add an explicit test for this in case we don't already have one. However, the relevant route would be the one that ends with `/done` I suppose. And then we'd of course need to run the Minion job enqueued by that route. And then you'll probably run into the problem mentioned in my other review comment.

Martchus (changes requested):
As mentioned in the inline comments, this might be much harder to make actually work than we estimated. Maybe we should reevaluate the best way of implementing this and possibly also re-consider whether it is worth it.

lib/OpenQA/Task/Job/Restart.pm
perlpunk: This event does not contain what the event for a "regular" job restart contains.
Please run the test and look at the event data. It needs to have the old id plus the new id(s).

d3flex: That might be easy to add. Maybe I could redo what the `_restart` in  lib/OpenQA/WebAPI/Controller/API/V1/Job.pm does. 
But I think we need to address first the points from Marius.

d3flex: That's a good one. Lets take a step back. 

> It probably isn't so emitting this event probably has no effect

I am not sure about that but  the `emit_event` in lib/OpenQA/Events.pm implies that this method is intended for non-controller contexts, such as Minion background tasks. right?

> To make this actually work we would probably need to make sure the AMQP plugin is loaded in the Gru service.

Maybe is enough to make the Openqa::events::emit_events clever and enqueue an event from there? could that possibly work?

perlpunk: Also the name of the event is wrong.
I just want to make sure you do know the existing emit_event call, and it looks like you don't, judging from how you call the event here and your questions about the event data.
Did you actually look into the event data and the current code?

https://github.com/os-autoinst/openQA/blob/master/lib/OpenQA/WebAPI/Controller/API/V1/Job.pm#L870
Here we have:
```
$self->emit_event(openqa_job_restart => \%event_data);
```
So it's called `openqa_job_restart` while you name the event just `job_restart`. Why do you give it a different name?

And about the contents of the event, just do this:
```diff
diff --git a/t/api/04-jobs.t b/t/api/04-jobs.t
index 6b68f26ff..305ec4198 100644
--- a/t/api/04-jobs.t
+++ b/t/api/04-jobs.t
@@ -438,6 +438,7 @@ subtest 'restart single job passing settings' => sub {
     $t->json_is('/warnings' => undef, 'no warnings generated');
     $t->json_is('/errors' => undef, 'no errors generated');
     my $event = OpenQA::Test::Case::find_most_recent_event($schema, 'job_restart');
+    always_explain $event;
     is $event->{id}, 99926, 'restart produces event';
     my $clone = $jobs->find(99926)->clone;
     isnt $clone, undef, 'job has been cloned' or return;
```
and run the test.
The output is:
```
t/api/04-jobs.t .. 17/?     # {
    #   'auto' => 0,
    #   'id' => '99926',
    #   'result' => {
    #     '99926' => 99997
    #   }
    # }
```
So you can see that you get a mapping of new id(s) for old id(s).
Have you been looking at this? I would expect this to be one of the first things when starting to work on the ticket. Look at the existing restart event.
We can not always tell you those things. It's part of the research for the ticket.

Martchus: Yes, it is intended to be used from non-controller context but *not* Minion tasks. I think "non-controller context" mainly refers to DBIx result classes or some utility modules.

We probably don't need to make `emit_events` clever. We just need to make sure that in the process that runs the Minion task the AMQP plugin is initialized (and maybe also other similar plugins for the sake of consistency). The problem with that is just that it might get messy due to the involved forking in combination with having to manage a permanent connection to the AMQP server.

Of course we could *actually* make `emit_event` clever and make it invoke some internal route of the normal web UI service to emit the event in case it is running from the Gru service. That *might* be the easiest solution - even though it requires a new internal route to do IPC.

d3flex: > Did you actually look into the event data and the current code?

I did but `auto` and `results` were added from the function in the function of the API. The focus was to find the right place for the emission so I left it out.

> So it's called openqa_job_restart while you name the event just job_restart. Why do you give it a different name?

because looking at the @job_events in lib/OpenQA/WebAPI/Plugin/AMQP.pm I thought that this is how it should be in the firstplace. I forgot to change it.

perlpunk: > > Did you actually look into the event data and the current code?
> 
> I did but `auto` and `results` were added from the function in the function of the API.

I don't understand the sentence.

> The focus was to find the right place for the emission so I left it out.

Then please mark things you leave out. I was writing this because we talked about how to look at the event data this morning, so I assumed this is clear.


> > So it's called openqa_job_restart while you name the event just job_restart. Why do you give it a different name?
> 
> because looking at the @job_events in lib/OpenQA/WebAPI/Plugin/AMQP.pm I thought that this is how it should be in the firstplace. I forgot to change it.

How I searched for existing restart events, maybe that helps you searching for things in the future:
```
% ack emit_event.*restart
lib/OpenQA/WebAPI/Controller/API/V1/Job.pm
870:        $self->emit_event(openqa_job_restart => \%event_data);
```
That result is pretty clear to me.

t/api/18-job-amqp-event.t
perlpunk: Yes, and I use this opportunity to remind about TDD (test driven development).
I took this PR, but only the new test and not the code changes.
The test passed.

It should be clear that this should not be the case. If you write a new test for a new or changed feature, then the test **must** fail until you add the actual code changes. If it passes, then you are not testing the new feature.

For checking this rule it isn't even important whether you _started_ development with a test or not.

So for the future: This is an easy rule to check for yourself before you push something to a PR.

lib/OpenQA/Task/Job/Restart.pm
```diff
@@ -17,6 +18,12 @@ sub restart_delay { $ENV{OPENQA_JOB_RESTART_DELAY} // 5 }
 sub restart_openqa_job ($minion_job, $openqa_job) {
     my $cloned_job_or_error = $openqa_job->auto_duplicate;
     my $is_ok = ref $cloned_job_or_error || $cloned_job_or_error =~ qr/(already.*clone|direct parent)/i;
+    if ($is_ok) {
+        my $openqa_job_id = $openqa_job->id;
+        my %event_data = (id => $openqa_job_id, result => $cloned_job_or_error, auto => 1);
```
perlpunk: Please make sure the event_data has the same kind of data as the other `openqa_job_restart` we send elsewhere.
This is not the case currently.

perlpunk (changes requested):

perlpunk:
> I submit a second commit to register the AMQP inside the minion space.

That's good news!
So now you can work on the open discussions.

lib/OpenQA/Shared/Plugin/Gru.pm
```diff
@@ -98,6 +98,19 @@ sub register ($self, $app, $config) {
...
 
     $self->register_tasks;
+    my $plugins = $app->config->{global}->{plugins};
+    if (defined $plugins && $plugins =~ /\bAMQP\b/) {
+        push @{$app->plugins->namespaces}, 'OpenQA::WebAPI::Plugin'
+          unless grep { $_ eq 'OpenQA::WebAPI::Plugin' } @{$app->plugins->namespaces};
+
+        try {
+            $app->plugin('AMQP');
+            Mojo::IOLoop->singleton->one_tick;
```
perlpunk: Why is this needed?

d3flex: It is described in the commit (if you think it is not clear let me know to update it.)
but to have a proper answer. 
The `register` function https://github.com/os-autoinst/openQA/blob/1801a6dd1ec72c42027785448bef68908316a275/lib/OpenQA/WebAPI/Plugin/AMQP.pm#L31 registers the events inside the `Mojo::IOLoop->singleton->next_tick`. The task tries to emit an event but it isnt actually registered yet. So I tried to force it, in order to have it registered asap. `one_tick` should trigger `next_tick` in the `register` sub.

d3flex:
Today I will take care of the coverage.

lib/OpenQA/Shared/Plugin/Gru.pm
Martchus: This will probably work but is of course a bit fragile as there's no explicit condition for stopping the event loop this way. I would at least add a short note:

```suggestion
            Mojo::IOLoop->singleton->one_tick;   # handle async plugin registration
```

Of course it would be better to register *some* event handler that is fired after the registration has completed in which we will stop the event loop explicitly. Then we would just start the event loop.

Maybe this kind of plugin registration should be its own function, e.g. `_register_amqp_plugin_synchronously`.

d3flex: I added the comment. And it took me a while to find how to write the test for the test coverage. I hope it is fine to keep it as is for now and write a note in the ticket, in case it is something to look at later.

lib/OpenQA/Task/Job/Restart.pm
d3flex: I think this can be closed now

lib/OpenQA/Task/Job/Restart.pm
d3flex: closing as outdated. reopen it if you feel that there is something needs attention.

lib/OpenQA/Task/Job/Restart.pm
perlpunk: I don't see why

lib/OpenQA/Schema/Result/Jobs.pm
```diff
@@ -2050,6 +2050,7 @@ sub enqueue_restart ($self, $options = {}) {
     my $openqa_job_id = $self->id;
     my $minion_job_id = OpenQA::App->singleton->gru->enqueue(restart_job => [$openqa_job_id], $options)->{minion_id};
     log_debug "Enqueued restarting openQA job $openqa_job_id via Minion job $minion_job_id";
+    OpenQA::App->singleton->emit_event(openqa_job_restart => \(id => $openqa_job_id));
```
perlpunk: Why?

t/api/18-job-amqp-event.t
d3flex: if I remember correctly, I added that change mainly because was not covered in the api tests. Apparently thats why it is added in `t/api`. it could be something else as well simultaneously. in a few words, this was not meant to cover the new case. The emission from the minion job should have implemented in `10-jobs.t:AMQP event emission for minion restarts`

lib/OpenQA/Task/Job/Restart.pm
d3flex: I guess this can closed as well?

t/23-amqp.t
d3flex: code has been removed.

t/23-amqp.t
d3flex: outdated and code has been removed

t/api/18-job-amqp-event.t
perlpunk: If this is covering an existing case that was not covered before, it should be in its own commit.

perlpunk:
I tried this out with a local rabbitmq and reported the problems here: https://progress.opensuse.org/issues/190557#note-27
So currently this doesn't work as expected.

lib/OpenQA/Schema/Result/Jobs.pm
d3flex: I think I have removed this. I will check

perlpunk:
Thanks to @Martchus we got the AMQP plugin working in the gru server during our collab session: https://progress.opensuse.org/issues/190557#note-29
I pushed two commits.

t/15-shared-plugin-gru.t
```diff
@@ -3,7 +3,6 @@
 # SPDX-License-Identifier: GPL-2.0-or-later
 
 use Test::Most;
-
```
okurz, line 6: why?

d3flex: obviously left over. I had edited it when i tried to add a test. I think I added signatures. When I remove it I missed that there was a line here previously. fixed

lib/OpenQA/Task/Job/Restart.pm
```diff
@@ -17,6 +18,13 @@ sub restart_delay { $ENV{OPENQA_JOB_RESTART_DELAY} // 5 }
 sub restart_openqa_job ($minion_job, $openqa_job) {
     my $cloned_job_or_error = $openqa_job->auto_duplicate;
     my $is_ok = ref $cloned_job_or_error || $cloned_job_or_error =~ qr/(already.*clone|direct parent)/i;
+    if ($is_ok) {
+        my $openqa_job_id = $openqa_job->id;
+        my $result = ref $cloned_job_or_error ? $cloned_job_or_error->{cluster_cloned} : $cloned_job_or_error;
+        my %event_data = (id => $openqa_job_id, result => $result, auto => 1);
```
perlpunk: We shouldn't emit an event if cloning failed.
```suggestion
    if (ref $cloned_job_or_error) {
        my %event_data = (id => $openqa_job->id, result => $cloned_job_or_error->{cluster_cloned}, auto => 1);

```

t/10-jobs.t
```diff
@@ -931,6 +933,66 @@ subtest 'job setting based retriggering' => sub {
...
+    # new app otherwise it runs slow. Maybe tries no-mocked stuff
+    my $t = Test::Mojo->new('OpenQA::WebAPI');
+    my $minion = $t->app->minion;
+    my $jobs2 = $t->app->schema->resultset('Jobs');
+    is $t->app->config->{amqp}->{enabled}, 1, 'AMQP enabled from config file';
+
+    my %_settings = %settings;
+    $_settings{TEST} = 'test_amqp_restart';
+    $_settings{RETRY} = '1';
+    my $job = $jobs2->create_from_settings(\%_settings);
```
perlpunk: No need for a new variable `$jobs2`.
```suggestion
    is $t->app->config->{amqp}->{enabled}, 1, 'AMQP enabled from config file';

    my %_settings = %settings;
    $_settings{TEST} = 'test_amqp_restart';
    $_settings{RETRY} = '1';
    my $job = $jobs->create_from_settings(\%_settings);
```

lib/OpenQA/Task/Job/Restart.pm
d3flex: But we do not emit if there is any error. Only if `$is_ok`

perlpunk: Sorry, to be more clear: I meant "when cloning actually happened".
```
my $is_ok = ref $cloned_job_or_error || $cloned_job_or_error =~ qr/(already.*clone|direct parent)/i;
```
In the latter case cloning did not happen.

d3flex: Done. I will try to test it also local one more time.

t/10-jobs.t
```diff
@@ -931,6 +933,65 @@ subtest 'job setting based retriggering' => sub {
...
+    $_settings{TEST} = 'test_amqp_restart';
+    $_settings{RETRY} = '1';
+    my $job = $jobs->create_from_settings(\%_settings);
+    $job->discard_changes;
+    my $job_id = $job->id;
+
+    %published = ();
+    $job->done(result => FAILED);
+    stdout_like { perform_minion_jobs($minion) } qr/Job \d+ duplicated as \d+/;
+    my $event = OpenQA::Test::Case::find_most_recent_event($schema, 'job_restart');
```
perlpunk: The `$event` variable is never used anywhere

d3flex: true. I used it to test the output (in always_explain) which I removed later but that line was left over. fixed.

lib/OpenQA/Task/Job/Restart.pm
d3flex: test looked good in machine. 
- parent job restarted and child job run
- one instance of events emitted per each (suse.openqa.job.create > suse.openqa.job.done > suse.openqa.job.restart > suse.openqa.job.done)

t/10-jobs.t
```diff
@@ -931,6 +933,64 @@ subtest 'job setting based retriggering' => sub {
...
+    $_settings{TEST} = 'test_amqp_restart';
+    $_settings{RETRY} = '1';
+    my $job = $jobs->create_from_settings(\%_settings);
+    $job->discard_changes;
+    my $job_id = $job->id;
+
+    %published = ();
+    $job->done(result => FAILED);
+    stdout_like { perform_minion_jobs($minion) } qr/Job \d+ duplicated as \d+/;
+    my $expected_topic = 'suse.openqa.job.restart';
```
perlpunk: IMHO it's hard to read what is tested here, and some tests are redundant.
I suggest the following diff:
```diff
diff --git a/t/10-jobs.t b/t/10-jobs.t
index d8a706c28..9719980d7 100644
--- a/t/10-jobs.t
+++ b/t/10-jobs.t
@@ -974,22 +974,20 @@ subtest 'AMQP event emission for minion restarts' => sub {
     %published = ();
     $job->done(result => FAILED);
     stdout_like { perform_minion_jobs($minion) } qr/Job \d+ duplicated as \d+/;
-    my $event = OpenQA::Test::Case::find_most_recent_event($schema, 'job_restart');
-    my $expected_topic = 'suse.openqa.job.restart';
-    my @restart_events = grep { $_ eq $expected_topic } keys %published;
-    ok exists $published{$expected_topic}, 'restart event published via AMQP';
-    is $published{$expected_topic}{id}, $job_id, 'event contains original job ID';
-    is $published{$expected_topic}{auto}, 1, 'event marked as auto restart';
-    is scalar(@restart_events), 1, 'exactly one job restart event emitted';
-    is $event_body[0], 'openqa_job_restart', 'event type is openqa_job_restart';
+    is scalar keys %published, 1, 'exactly one job restart event emitted';
+    my $event = $published{'suse.openqa.job.restart'};
+    is $event->{id}, $job_id, 'event contains original job ID';
+    is $event->{auto}, 1, 'event marked as auto restart';
+
     my ($user_id, $connection, $type, $data) = @{$event_body[1]};
     is $user_id, undef, 'user_id is undef for Minion restart';
     is $connection, undef, 'connection is undef for Minion restart';
     is $type, 'openqa_job_restart', 'type matches event type';
-    is $data->{id}, $job_id, 'data contains original job ID';
-    is $data->{auto}, 1, 'data marked as auto restart';
-    is ref($data->{result}), 'HASH', 'result should be a plain hash like API';
-    is_deeply $data->{result}, {$job_id => $data->{result}{$job_id}}, 'result shows cloned job info';
+
+    is_deeply $data, $event, 'published event equals emitted event';
+    ok exists $data->{result}->{$job_id}, 'old job id is in result';
+    $job->discard_changes;
+    is $job->clone_id, $data->{result}->{$job_id}, 'clone_id points to reported id';
 };
```

t/10-jobs.t
```diff
@@ -931,6 +933,64 @@ subtest 'job setting based retriggering' => sub {
...
+    is scalar(@restart_events), 1, 'exactly one job restart event emitted';
+    is $event_body[0], 'openqa_job_restart', 'event type is openqa_job_restart';
+    my ($user_id, $connection, $type, $data) = @{$event_body[1]};
+    is $user_id, undef, 'user_id is undef for Minion restart';
+    is $connection, undef, 'connection is undef for Minion restart';
+    is $type, 'openqa_job_restart', 'type matches event type';
+    is $data->{id}, $job_id, 'data contains original job ID';
+    is $data->{auto}, 1, 'data marked as auto restart';
+    is ref($data->{result}), 'HASH', 'result should be a plain hash like API';
+    is_deeply $data->{result}, {$job_id => $data->{result}{$job_id}}, 'result shows cloned job info';
```
perlpunk: Alternatively here as a suggestion:
```suggestion
    is scalar keys %published, 1, 'exactly one job restart event emitted';
    my $event = $published{'suse.openqa.job.restart'};
    is $event->{id}, $job_id, 'event contains original job ID';
    is $event->{auto}, 1, 'event marked as auto restart';

    my ($user_id, $connection, $type, $data) = @{$event_body[1]};
    is $user_id, undef, 'user_id is undef for Minion restart';
    is $connection, undef, 'connection is undef for Minion restart';
    is $type, 'openqa_job_restart', 'type matches event type';

    is_deeply $data, $event, 'published event equals emitted event';
    ok exists $data->{result}->{$job_id}, 'old job id is in result';
    $job->discard_changes;
    is $job->clone_id, $data->{result}->{$job_id}, 'clone_id points to reported id';
```

d3flex: I used the later

perlpunk (commented):
I don't like the third commit message subject much "Make restart_openqa_job emit event same as API" (what should "same as API" tell us?)
Also the message from the first commit wasn't fixed yet, it still contains the "TBD" thing from me.
Otherwise I'd approve.

d3flex:
> > I don't like the third commit message subject much "Make restart_openqa_job emit event same as API" (what should "same as API" tell us?) Also the message from the first commit wasn't fixed yet, it still contains the "TBD" thing from me. Otherwise I'd approve.
> 
> I will look at the commit. the `TBD` was not totally sure what was about. edit the commit or feature to be done. However I have modified it a bit. Do not approve yet. I try to fix the codecov

Commit messages were updated.
I didnt find a solution for the missing coverage (https://app.codecov.io/gh/os-autoinst/openQA/pull/6816?dropdown=coverage&src=pr&el=h1&utm_medium=referral&utm_source=github&utm_content=checks&utm_campaign=pr+comments&utm_term=os-autoinst). I marked it with `# uncoverable statement`. theoretically it can test it as a unit test. but in 10-jobs.t? is it worth it after all?

okurz (approved):

lib/OpenQA/Task/Job/Restart.pm
```diff
@@ -46,4 +55,20 @@ sub _restart_job ($minion_job, @args) {
...
+    Mojo::IOLoop->singleton->one_tick;
+}
+
+sub _wait_for_event_publish ($app) {
+    return undef unless $app->config->{amqp}->{enabled};
+    OpenQA::Events->singleton->once(
+        'amqp_handled',
+        sub {
+            Mojo::IOLoop->singleton->stop;    # uncoverable statement
+        });
```
Martchus: Maybe this way you can avoid the "uncoverable statement" comment and make it much shorter:
```suggestion
    OpenQA::Events->singleton->once(amqp_handled => sub { Mojo::IOLoop->stop });
```

lib/OpenQA/Task/Job/Restart.pm
```diff
@@ -46,4 +55,20 @@ sub _restart_job ($minion_job, @args) {
...
+    Mojo::IOLoop->singleton->start;
```
Martchus: For consistency with my previous suggestion:
```suggestion
    Mojo::IOLoop->start;
```

t/10-jobs.t
```diff
@@ -931,6 +933,61 @@ subtest 'job setting based retriggering' => sub {
     is $lastest_job->latest_job->id, $lastest_job->id, 'found the latest job from latest job itself';
 };
 
+subtest 'AMQP event emission for minion restarts' => sub {
```
Martchus:
```suggestion
subtest 'AMQP event emission for job restarts within Minion tasks' => sub {
```

t/10-jobs.t
```diff
@@ -931,6 +933,61 @@ subtest 'job setting based retriggering' => sub {
...
+    my $conf = "[global]\nplugins=AMQP\n[amqp]\npublish_attempts = 2\npublish_retry_delay = 0\n";
+    my $tempdir = tempdir;
+    path($ENV{OPENQA_CONFIG} = $tempdir)->make_path->child('openqa.ini')->spew($conf);
+    my %published;
+    my @event_body;
+    my $io_loop_mock = Test::MockModule->new('Mojo::IOLoop');
+    $io_loop_mock->redefine(
+        start => sub ($loop) {
+            $loop->one_tick while $loop->is_running;
+        });
```
Martchus: Is this really required? To me it looks like you're just replacing the start function with another (more convoluted) way of starting the loop.

t/10-jobs.t
```diff
@@ -931,6 +933,61 @@ subtest 'job setting based retriggering' => sub {
...
+
+    my $events_mock = Test::MockModule->new('OpenQA::Events');
+    $events_mock->redefine(
+        emit => sub ($self, $type, $args) {
+            if ($type eq 'openqa_job_restart') {
+                @event_body = ($type, $args);
+            }
+            $events_mock->original('emit')->($self, $type, $args);
+        });
+    # new app otherwise it runs slow. Maybe tries no-mocked stuff
```
Martchus: That's not great but if you don't know a better way than let's at least improve the grammar here:
```suggestion
    # use a new app; otherwise it runs slowly, maybe trying to run none-mocked stuff
```

t/10-jobs.t
```diff
@@ -931,6 +933,61 @@ subtest 'job setting based retriggering' => sub {
...
+    my $t = Test::Mojo->new('OpenQA::WebAPI');
+    my $minion = $t->app->minion;
+    is $t->app->config->{amqp}->{enabled}, 1, 'AMQP enabled from config file';
+
+    my %_settings = %settings;
```
Martchus:
```suggestion
    my %retry_settings = %settings;
```

Martchus (approved):

PR #6818, opened 2025-10-23
Log output of `optipng` when it fails

Martchus:
* Remove the `-quiet` parameter
* Use `run_cmd_with_log_return_error` for logging the command output in the error case
* Improve error messages in `run_cmd_with_log_return_error` to include the program name so the error message is also useful on its own (without the prior `Running cmd: …` message)
* Make `run_cmd_with_log_return_error` return the error message so it can be passed as reason when using it to run `optipng`
* See https://progress.opensuse.org/issues/190920

### DIFF
--- a/lib/OpenQA/Utils.pm
+++ b/lib/OpenQA/Utils.pm
@@ -305,6 +305,7 @@ sub run_cmd_with_log_return_error ($cmd, %args) {
     my $stdout_level = $args{stdout} // 'debug';
     my $stderr_level = $args{stderr} // 'debug';
     my $output_file = $args{output_file};
+    my $program = $cmd->[0] // 'cmd';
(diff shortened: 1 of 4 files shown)


Martchus:
It looks like using the IPC run-based `run_cmd_with_log_return_error` cannot be used here - at least not without further modification. The problem is that when `run_cmd_with_log_return_error` is used after `call_driver` (like we do in fullstack tests) the `$?` variable doesn't return the correct value. And also the return value of `IPC::Run::run(…)` is false when it should be true.

Martchus:
Thanks for approving - however, some concrete advice on how to proceed would be more useful. As mentioned before, we cannot use IPC::Run here - at least not without figuring out why it breaks in our UI tests that launch Selenium.

PR #6820, opened 2025-10-24
Retry image optimizations

Martchus:
The optimization can fail with `optipng` returning with a non-zero return code, see https://progress.opensuse.org/issues/190920. We don't know exactly why it fails sometimes. Perhaps it is because the image is actually still written when we invoke `optipng`. This change introduces a simple retry which should help if it is just that. If `optipng` cannot be executed at all (e.g. because it isn't installed) the worker will still fail immediately.

Related ticket: https://progress.opensuse.org/issues/190920

### DIFF
--- a/t/24-worker-jobs.t
+++ b/t/24-worker-jobs.t
@@ -9,6 +9,8 @@ use Test::Warnings ':report_warnings';
 BEGIN {
     $ENV{OPENQA_UPLOAD_DELAY} = 0;
     $ENV{OPENQA_WORKER_ACCEPT_ATTEMPTS} = 2;
+    $ENV{OPENQA_WORKER_OPTIMIZE_ATTEMPTS} = 2;
+    $ENV{OPENQA_WORKER_OPTIMIZE_RETRY_DELAY} = 0;
 }
 
 use FindBin;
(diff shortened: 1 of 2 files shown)


t/24-worker-jobs.t
```diff
@@ -9,6 +9,8 @@ use Test::Warnings ':report_warnings';
 BEGIN {
     $ENV{OPENQA_UPLOAD_DELAY} = 0;
     $ENV{OPENQA_WORKER_ACCEPT_ATTEMPTS} = 2;
+    $ENV{OPENQA_WORKER_OPTIMIZE_ATTEMPTS} = 2;
+    $ENV{OPENQA_WORKER_OPTIMIZE_RETRY_DELAY} = 0;
```
kalikiana, line 13: How would we know that these are effective? The unit test should probably make them fail on purpose.

kalikiana: Considering the commit message saying we don't really know why these fail, I could also see that we can't test it, but then just setting variables would seem misleading to me.

Martchus: The existing tests are already testing the retry. The coverage check would fail if it wasn't tested.

These env variables are also definitely effective. I ran tests without them and then the delay was notable.

kalikiana: use constant OPTIMIZE_ATTEMPTS => $ENV{OPENQA_WORKER_OPTIMIZE_ATTEMPTS} // 5;
    use constant OPTIMIZE_RETRY_DELAY => $ENV{OPENQA_WORKER_OPTIMIZE_RETRY_DELAY} // 1;

These are new variables, no? They were not even being used before.

Martchus: Yes, they are new. However, the case when the command fails is tested and with this change it tests also the retry. I could add additional checks but this would also require additional mocking. I was hoping to keep this simple so maybe if I manage to implement the capturing/logging the conflict isn't too big.

kalikiana (approved):

PR #6824, opened 2025-10-30
Mark non-critical Schema::Result::Workers log message as log_info

okurz:
Related progress issue: https://progress.opensuse.org/issues/191446

### DIFF
--- a/lib/OpenQA/Schema/Result/Workers.pm
+++ b/lib/OpenQA/Schema/Result/Workers.pm
@@ -215,10 +215,10 @@ sub send_command {
     my $client = OpenQA::WebSockets::Client->singleton;
     try { $client->send_msg($self->id, $args{command}, $args{job_id}) }
     catch ($e) {
-        log_error(
+        log_info(
             sprintf(
                 'Failed dispatching message to websocket server over ipc for worker "%s:%n": %s',
-                $self->host, $self->instance, $e
+                $self->host || '?', $self->instance || '?', $e
             ));
         return undef;
     }


aylen384 (approved):

lib/OpenQA/Schema/Result/Workers.pm
```diff
@@ -193,7 +193,7 @@ sub info ($self) {
 
 sub send_command {
     my ($self, %args) = @_;
-    return undef if (!defined $args{command});
+    return undef unless defined $args{command} && defined $self->host && defined $self->instance;
```
Martchus: We should probably prevent the registration of workers with empty `host` and `instance` names in the first place. (And with empty I mean empty. Your check is about whether those values are defined at all.)

Note that in theory `$self->host` and `$self->instance` are already supposed to be never undefined because those columns are not nullable. We should rather find out why `$self->instance` was rendered as empty string here and fix a potential bug elsewhere.

I also doubt that the websocket send attempt was invalid here. We don't need an instance number for this to work. The `Connection refused` error indicates that the web socket server was not responding which has nothing to do with the concrete worker the command is sent to.

Martchus (changes requested):

lib/OpenQA/Schema/Result/Workers.pm
Martchus: This happened when sending a command to abort a job, see https://progress.opensuse.org/issues/191446#note-7. Maybe we can replicate the missing `$self->instance` in unit tests for this use-case.

Note that `$self->instance` was *not* `undef`; otherwise we would have gotten a Perl warning about it but the logs don't contain such a warning. So it was probably an empty string.

okurz: Ok, so I would definitely need to change the message. But isn't it still a good idea to prevent such Perl warning about use of undefined values? We can also have a fatal pre-condition check with

```suggestion
    die "Undefined command/host/instance" unless defined $args{command} && defined $self->host && defined $self->instance;
```

or

```suggestion
    die "Undefined command" unless defined $args{command};
    die "Undefined host" unless defined $self->host;
    die "Undefined instance" unless defined $self->instance;
```

Martchus: I don't think this should be fatal. We still want to run subsequent code despite this function failing. Hence also the try-catch blocks. For simply having a warning about this we can of course rely on the Perl warning we get for free (and not prevent it).

Note that the issue at hand it *not* about undefined values, though. Otherwise we would see a Perl warning about it but I haven't found a warning in the logs.

In general it is useful to specify/check post conditions in some way (if there's a good way). I don't think that checks like you propose that are always run unconditionally at runtime causing hard failures are conventional and a good idea in Perl. And it is likely also not what we want here as mentioned before.

Martchus: By the way, maybe we should just change the error message into a debug or info message if we don't want to improve the actual problem here. (The actual problem is that the different services can run into "connection refused" errors when they are restarted.)

Ideally we also tell the user about it so they know that not all jobs could be aborted after clicking the restart button.

Martchus: I'd also like to note that knowing the host and instance are not pre-conditions of that function except for printing a log message in some error cases. The only thing we need here is the worker ID and the command to send.

okurz: If we only use those variables in a log message then we should ensure a valid string for the log messages, e.g.

```perl
log_error(sprintf($msg, $args{command}, $self->host // '', $self->instance // ''));
```

> By the way, maybe we should just change the error message into a debug or info message if we don't want to improve the actual problem here. (The actual problem is that the different services can run into "connection refused" errors when they are restarted.)
> 
> Ideally we also tell the user about it so they know that not all jobs could be aborted after clicking the restart button.

ok but that probably needs to happen elsewhere and I don't right now how to do that. So should I do only the above improvement or close the PR and nothing at all?

Martchus: I would say you should only do the above improvement - but use `|| '?'` instead of `// ''`. Only that prevents log messages with nothing after the colon. Your version wouldn't make any difference (as there was probably an empty string assigned anyway).

And I would change it from `log_error` to `log_info`.

Martchus: Although it would be good to find out why `$self->instance` was an empty string considering it is supposed to return the value of a non-nullable integer column.

lib/OpenQA/Schema/Result/Workers.pm
```diff
@@ -215,10 +215,10 @@ sub send_command {
     my $client = OpenQA::WebSockets::Client->singleton;
     try { $client->send_msg($self->id, $args{command}, $args{job_id}) }
     catch ($e) {
-        log_error(
+        log_info(
             sprintf(
                 'Failed dispatching message to websocket server over ipc for worker "%s:%n": %s',
-                $self->host, $self->instance, $e
+                $self->host // '?', $self->instance // '?', $e
```
Martchus: If you really want to avoid empty parts before/after the colon (as we discussed when refining the ticket):
```suggestion
                $self->host || '?', $self->instance || '?', $e
```
This would also turn `0` into `?`, though. In my opinion it doesn't really matter as we're talking about a case we shouldn't have to deal with anyway.

lib/OpenQA/Schema/Result/Workers.pm
```diff
@@ -215,10 +215,10 @@ sub send_command {
     my $client = OpenQA::WebSockets::Client->singleton;
     try { $client->send_msg($self->id, $args{command}, $args{job_id}) }
     catch ($e) {
-        log_error(
+        log_info(
```
perlpunk, line 218: Tests are failing. `log_info` needs to be imported in line 11, I guess that's the reason.

PR #6825, opened 2025-10-30
Apply TIMEOUT_SCALE twice when needed

rclobus:
When TIMEOUT_SCALE is set for a worker and the command line also provides the value, apply both.

### DIFF
--- a/lib/OpenQA/Worker/Job.pm
+++ b/lib/OpenQA/Worker/Job.pm
@@ -222,7 +222,7 @@ sub start ($self) {
     # update settings received from web UI with worker-specific stuff
     my $worker = $self->worker;
     my $global_worker_settings = $worker->settings->global_settings;
-    # TIMEOUT_SCALE: apply the factors from both the job settings and the worker settings
+    # apply the time scale factors from both the job settings and the worker settings
     my $timeout_scale = $global_worker_settings->{TIMEOUT_SCALE} // 1;
     $timeout_scale *= $job_settings->{TIMEOUT_SCALE} // 1;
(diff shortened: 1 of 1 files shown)


lib/OpenQA/Worker/Job.pm
```diff
@@ -222,7 +222,13 @@ sub start ($self) {
     # update settings received from web UI with worker-specific stuff
     my $worker = $self->worker;
     my $global_worker_settings = $worker->settings->global_settings;
+    # TIMEOUT_SCALE: apply the factors from both the job settings and the worker settings
```
Martchus:
```suggestion
    # apply the time scale factors from both the job settings and the worker settings
```

rclobus:
> I _guess_ this makes sense. (Not exactly sure what you're trying to achieve here.)

I have a job that takes a long time, and I want to finish it by overriding the timeouts from the openqa-cli command line. However, the worker already has `TIMEOUT_SCALE` defined in the worker.ini file and then the value from the command line gets overwritten. Unfortunately the value in the worker.ini file is not sufficient to make the job pass.

I've run the job in a local, empty openQA instance and there the timeouts are all sufficient. It is our public worker that is under load and I want to run the job first with extended timeouts, so I can extract more suitable values from the autoinst-log.txt file.

rclobus:
> But should TIMEOUT_SCALE be treated in any special way then?

@okurz See my reasoning in https://github.com/os-autoinst/openQA/pull/6825#issuecomment-3475866561. The user who calls 'openqa-cli' cannot know in advance whether the job will be scheduled on a fast or slow worker, so both multiplication factors need to be considered.

PR #6826, opened 2025-10-31
Properly conclude scheduling if there are no jobs

Vogtinator:
I found that openqa-schedule was pretty much always at 100% CPU use on my workstation, which could only be worked around by a restart of the scheduler. I finally had a look and this fixes it.

If a wakeup happens while there are no jobs to be scheduled, it did not emit concluded and the timer for scheduling is left at 0ms, which causes high CPU and DB use by the scheduler. Move the call to emit concluded to the right place to ensure it's called in that case as well.

### DIFF
--- a/lib/OpenQA/Scheduler/Model/Jobs.pm
+++ b/lib/OpenQA/Scheduler/Model/Jobs.pm
@@ -46,7 +46,6 @@ sub _allocate_jobs ($self, $free_workers) {
     my $limit = OpenQA::App->singleton->config->{scheduler}->{max_running_jobs};
     if ($limit >= 0 && $running >= $limit) {
         log_debug("max_running_jobs ($limit) exceeded, scheduling no additional jobs");
-        $self->emit('conclude');
         return ({}, {});
     }
     my $max_allocate = $limit >= 0 ? min(MAX_JOB_ALLOCATION, $limit - $running) : MAX_JOB_ALLOCATION;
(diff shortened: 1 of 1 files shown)


lib/OpenQA/Scheduler/Model/Jobs.pm
```diff
@@ -167,7 +166,10 @@ sub schedule ($self) {
         "Scheduling: Free workers: $free_worker_count/$worker_count; Scheduled jobs: " . scalar(keys %$scheduled_jobs));
 
     my ($allocated_workers, $allocated_jobs) = $self->_allocate_jobs($free_workers);
-    return [] unless keys %$allocated_workers;
```
Vogtinator, line 170: OTOH, might also be possible to just delete this line altogether? That turns the loop below into a noop

lib/OpenQA/Scheduler/Model/Jobs.pm
```diff
@@ -46,7 +46,6 @@ sub _allocate_jobs ($self, $free_workers) {
     my $limit = OpenQA::App->singleton->config->{scheduler}->{max_running_jobs};
     if ($limit >= 0 && $running >= $limit) {
         log_debug("max_running_jobs ($limit) exceeded, scheduling no additional jobs");
-        $self->emit('conclude');
```
Martchus, line 49: Don't we still need it here? This condition has an early return and I suppose the same counts here as for other early returns.

lib/OpenQA/Scheduler/Model/Jobs.pm
Martchus: Yes, you can also just remove this line entirely (and also not add an additional `emit('conclude')` relying on the one at the end of the function). That would probably be the best solution.

Martchus (changes requested):

lib/OpenQA/Scheduler/Model/Jobs.pm
Vogtinator: No, the call at the end of `schedule` is reached in that case.

lib/OpenQA/Scheduler/Model/Jobs.pm
Vogtinator: I just removed all early exits.

lib/OpenQA/Scheduler/Model/Jobs.pm
Martchus: Oh, right. This is a whole different function.

PR #6828, opened 2025-10-31
Adapt os-autoinst-scripts reference after rename

okurz:
Related progress issue: https://progress.opensuse.org/issues/191635

### DIFF
--- a/docs/WritingTests.asciidoc
+++ b/docs/WritingTests.asciidoc
@@ -654,9 +654,9 @@ hash for tests and/or needles (see
 for details) and a concrete worker host (see
 <<WritingTests.asciidoc#_assigning_jobs_to_workers,Assigning jobs to workers>>).
 The script
-https://github.com/os-autoinst/scripts/blob/master/README.md#openqa-investigate---automatic-investigation-jobs-with-failure-analysis-in-openqa[`openqa-investigate`] helps automating retries like this. It will also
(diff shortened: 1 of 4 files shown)


docs/WritingTests.asciidoc
```diff
@@ -2333,7 +2333,7 @@ It is also possible to create a GitHub workflow that will clone and monitor an
 openQA job which is mentioned in the PR description or comment. The scripts
```
d3flex, line 2333: the coment still mentions the repo as `script` here. Not big deal but... could you rename that as well?

kalikiana: image

Actually what *workflow* is this about? I don't see any workflow using openqa-clone-and-monitor-job-from-pr in https://github.com/os-autoinst/os-autoinst-scripts/tree/master/.github/workflows 🤔

kalikiana: Ah but wait there is an action https://github.com/os-autoinst/os-autoinst-scripts/blob/master/actions/clone-job/action.yaml

kalikiana: Maybe we can document how to use that - I'd say that can go in a follow-up PR, though, and unless someone beats me to it I would propose something

kalikiana: https://github.com/os-autoinst/openQA/pull/6834

PR #6840, opened 2025-11-14
 Skip a build if there are no jobs

perlpunk:
My theory: due to a race condition there might not be any job left in a build,
when we start to process it, because all were jobs deleted meanwhile.

Issue: https://progress.opensuse.org/issues/192496


okurz (approved):

d3flex:
> > I was about to approve but I notice a tiny drop in the coverage. https://app.codecov.io/gh/os-autoinst/openQA/pull/6840/indirect-changes it makes me think that it might not that simple as it looks like, Why does it effect this part in the `cancel()`. Could you take a look. Maybe is nothing
> 
> I assume that's a flaky coverage and is unrelated. My change has nothing to do with cancelling. I can do a force push and see if it changes something.

i hear you.. gonna approve

d3flex (approved):

perlpunk:
> I assume that's a flaky coverage and is unrelated. My change has nothing to do with cancelling. I can do a force push and see if it changes something.

yep, the coverage change is gone now

PR #6843, opened 2025-11-17
Make the number of builds per group on the front page configurable

AdamWill:
The number of builds to show per job group on the front page of the web UI is hardcoded to 3.

SUSE's openQA has a ton of job groups so only showing three per group makes sense, but not every instance has so many groups, and the front page can look unnecessarily empty. This allows the instance maintainer to set an appropriate value.

The value can be overridden for a single query with the limit_builds query param, but that's unwieldy and doesn't solve things for casual views of the home page.

### DIFF
--- a/etc/openqa/openqa.ini
+++ b/etc/openqa/openqa.ini
@@ -125,6 +125,9 @@
 ## resilient in case of a high load on the web UI. Lower values reduce this risk but
 ## can cause jobs to incomplete with "timestamp mismatch" error messages.
 #api_hmac_time_tolerance = 300
+## How many builds to show per job group on the web UI front page
+## Can be overridden with the limit_builds query param
+#frontpage_builds = 3
 
 ## Configuration related to actions openQA may perform on a git distri
 #[scm git]
--- a/lib/OpenQA/Setup.pm
+++ b/lib/OpenQA/Setup.pm
@@ -96,6 +96,7 @@ sub read_config ($app) {
             service_port_delta => $ENV{OPENQA_SERVICE_PORT_DELTA} // 2,
             access_control_allow_origin_header => undef,
             api_hmac_time_tolerance => 300,
+            frontpage_builds => 3,
         },
         rate_limits => {
             search => 5,
--- a/lib/OpenQA/WebAPI/Controller/Main.pm
+++ b/lib/OpenQA/WebAPI/Controller/Main.pm
@@ -23,7 +23,7 @@ sub dashboard_build_results ($self) {
     $validation->optional('group');
     return $self->reply->validation_error({format => $self->accepts('html', 'json')}) if $validation->has_error;
 
-    my $limit_builds = $validation->param('limit_builds') // 3;
+    my $limit_builds = $validation->param('limit_builds') // $self->app->config->{global}->{frontpage_builds};
     my $time_limit_days = $validation->param('time_limit_days') // 14;
     my $only_tagged = $validation->param('only_tagged') // 0;
     my $default_expanded = $validation->param('default_expanded') // 0;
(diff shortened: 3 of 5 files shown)


etc/openqa/openqa.ini
```diff
@@ -125,6 +125,9 @@
 ## resilient in case of a high load on the web UI. Lower values reduce this risk but
 ## can cause jobs to incomplete with "timestamp mismatch" error messages.
 #api_hmac_time_tolerance = 300
+## How many builds to show per job group on the web UI front page
+## Can be overridden with the limit_builds query param
+# frontpage_builds = 3
```
okurz: I think we can the page either index page or dashboard. I wouldn't introduce another term but if we already use frontpage elsewhere then it's fine

AdamWill: Well, I think it's usually called 'dashboard', but it seems to be mostly in source code, not user/admin-facing docs. The word appears a few times in docs/UsersGuide.asciidoc , but only one time is it specifically referring to the home page and it doesn't really explain what it means (just says "index dashboard"). As a user/admin, if you say "dashboard" or "index dashboard" I really don't know what you mean. If you say "front page", I know what you mean. Still, I guess the comment explains it.

We can't really change it to 'dashboard_builds', though, because the group overviews are also called 'dashboards' and they do not use this value (and default to 10). So it'd have to be index_dashboard_builds ?

okurz: ok, fine. I guess at least "frontpage" is unique and not ambiguous :)

d3flex: can you remove the whitespace in `# frontpage_builds = 3` 
```suggestion
#frontpage_builds = 3
```
I dont know why all the defaults are without a whitespace and I would expect them with one but for consistency let's just do that here.

lib/OpenQA/Setup.pm
```diff
@@ -96,6 +96,7 @@ sub read_config ($app) {
             service_port_delta => $ENV{OPENQA_SERVICE_PORT_DELTA} // 2,
             access_control_allow_origin_header => undef,
             api_hmac_time_tolerance => 300,
+            frontpage_builds => 3,
```
Martchus, line 99: It would make more sense to add this under `misc_limits` (although this category is now also quite big).

I also think we normally call it the "index" page but I suppose front page is also fine.

etc/openqa/openqa.ini
Martchus:
```suggestion
#frontpage_builds = 3
```

lib/OpenQA/Setup.pm
perlpunk: there were already discussions above about the word "frontpage" and about the config section

etc/openqa/openqa.ini
perlpunk: Already suggested above

lib/OpenQA/Setup.pm
Martchus: As mentioned, I also think "frontpage" is fine. So there's no reason to change that unless @AdamWill now wants to :-)

AdamWill: @okurz said not to put it in `misc_limits`, above.

etc/openqa/openqa.ini
AdamWill: Done, thanks.

lib/OpenQA/Setup.pm
Martchus: Ok, I don't care much about it. I just would have found `misc_limits` more intuitive. (And we do have display limits like `next_jobs_default_limit` there.)

PR #6845, opened 2025-11-18
Give jobs with high MAX_JOB_TIME a priority malus

perlpunk:
The default is to divide by 100 and add the result to the priority. Can be configured.

### DIFF
--- a/lib/OpenQA/Schema/ResultSet/Jobs.pm
+++ b/lib/OpenQA/Schema/ResultSet/Jobs.pm
@@ -116,10 +116,6 @@ sub create_from_settings ($self, $settings, $scheduled_product_id = undef) {
     my %settings = %$settings;
     my %new_job_args;
 
-    my $result_source = $self->result_source;
-    my $schema = $result_source->schema;
-    my $job_settings = $schema->resultset('JobSettings');
-
     my @invalid_keys = grep { $_ =~ /^(PUBLISH_HDD|FORCE_PUBLISH_HDD|STORE_HDD)\S+(\d+)$/ && $settings{$_} =~ /\// }
       keys %settings;
     die 'The ' . join(',', @invalid_keys) . " cannot include / in value\n" if @invalid_keys;
@@ -142,38 +138,7 @@ sub create_from_settings ($self, $settings, $scheduled_product_id = undef) {
         $new_job_args{priority} //= $group->default_priority;
     }
 
-    # handle dependencies
-    # note: The subsequent code only allows adding existing jobs as parents. Hence it is not
-    #       possible to create cyclic dependencies here.
-    my @dependency_definitions = (
-        {
-            setting_name => '_START_AFTER_JOBS',
-            dependency_type => OpenQA::JobDependencies::Constants::CHAINED,
-        },
-        {
-            setting_name => '_START_DIRECTLY_AFTER_JOBS',
-            dependency_type => OpenQA::JobDependencies::Constants::DIRECTLY_CHAINED,
-        },
-        {
-            setting_name => '_PARALLEL_JOBS',
-            dependency_type => OpenQA::JobDependencies::Constants::PARALLEL,
-        },
-    );
-    for my $dependency_definition (@dependency_definitions) {
-        next unless my $ids = delete $settings{$dependency_definition->{setting_name}};
-
-        # support array ref or comma separated values
-        $ids = [split(/\s*,\s*/, $ids)] if (ref($ids) ne 'ARRAY');
-
-        my $dependency_type = $dependency_definition->{dependency_type};
-        for my $id (@$ids) {
-            if ($dependency_type eq OpenQA::JobDependencies::Constants::DIRECTLY_CHAINED) {
-                my $parent_worker_classes = join(',', @{$job_settings->all_values_sorted($id, 'WORKER_CLASS')});
-                _handle_directly_chained_dep($parent_worker_classes, $id, \%settings);
-            }
-            push(@{$new_job_args{parents}}, {parent_job_id => $id, dependency => $dependency_type});
-        }
-    }
+    $self->_handle_dependency_settings(\%settings, \%new_job_args);
 
     for my $key (keys %settings) {
         my $value = $settings{$key};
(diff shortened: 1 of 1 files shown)


etc/openqa/openqa.ini
```diff
@@ -325,6 +325,8 @@ concurrent = 0
 #wait_for_grutask_retries = 6
 ## Maximum size of MCP results in bytes (to prevent performance issues)
 # mcp_max_result_size = 500000
+## If MAX_JOB_TIME is greater than DEFAULT_MAX_JOB_TIME, divide it by this number and add the result to the job priority
+#max_job_time_priority_malus_scale = 100
```
perlpunk: I'm not completely happy with `misc_limits.max_job_time_priority_malus_scale`. Any suggestions?

lib/OpenQA/Schema/ResultSet/Jobs.pm
```diff
@@ -190,6 +191,11 @@ sub create_from_settings ($self, $settings, $scheduled_product_id = undef) {
     # assign scheduled product
     $new_job_args{scheduled_product_id} = $scheduled_product_id;
 
+    if ($settings{MAX_JOB_TIME} and $settings{MAX_JOB_TIME} > DEFAULT_MAX_JOB_TIME) {
```
```suggestion
    if (looks_like_number($settings{MAX_JOB_TIME}) and $settings{MAX_JOB_TIME} > DEFAULT_MAX_JOB_TIME) {
```

perlpunk: done

etc/openqa/openqa.ini
```diff
@@ -325,6 +325,8 @@ concurrent = 0
 #wait_for_grutask_retries = 6
 ## Maximum size of MCP results in bytes (to prevent performance issues)
 # mcp_max_result_size = 500000
+## If MAX_JOB_TIME is greater than DEFAULT_MAX_JOB_TIME, divide it by this number and add the result to the job priority
```
okurz: I suggest for this comment to explain the use case not (just) the how

etc/openqa/openqa.ini
okurz: Just misc_limits.max_job_time_prio_scale

A positive value is effectively a "malus", a negative value is a "bonus". The scale itself is neutral

okurz (changes requested):

etc/openqa/openqa.ini

etc/openqa/openqa.ini

etc/openqa/openqa.ini
```diff
@@ -325,6 +325,8 @@ concurrent = 0
 #wait_for_grutask_retries = 6
 ## Maximum size of MCP results in bytes (to prevent performance issues)
 # mcp_max_result_size = 500000
+## To discourage long running job scenarios: If MAX_JOB_TIME is greater than DEFAULT_MAX_JOB_TIME, divide it by this number and add the result to the job priority
```
okurz: Three points:
1. We can use positive phrasing to exlpain how openQA is actually "nice" to short running jobs :)
2. The phrase "divide it" sounds like a request for the reader but it's an explanation for existing behaviour
3. We can explicitly mention how zero and negative values work


```suggestion
## To encourage job scenarios with shorter runtime: If MAX_JOB_TIME is greater than
## DEFAULT_MAX_JOB_TIME, MAX_JOB_TIME divided by this number is added to the
## job priority. Can be 0 to disable and negative to prioritize long running job
## scenarios.
```

lib/OpenQA/Schema/ResultSet/Jobs.pm
```diff
@@ -190,6 +192,11 @@ sub create_from_settings ($self, $settings, $scheduled_product_id = undef) {
     # assign scheduled product
     $new_job_args{scheduled_product_id} = $scheduled_product_id;
 
+    if (looks_like_number($settings{MAX_JOB_TIME}) and $settings{MAX_JOB_TIME} > DEFAULT_MAX_JOB_TIME) {
+        my $scale = OpenQA::App->singleton->config->{misc_limits}->{max_job_time_prio_scale};
+        $new_job_args{priority} += int($settings{MAX_JOB_TIME} / $scale);
```
```suggestion
        $new_job_args{priority} += int($settings{MAX_JOB_TIME} / $scale) if $scale;
```

to handle 0 and effectively disable the feature.

By the way the function you are adding this code to is already rather big. How about splitting it?

t/api/04-jobs.t
```diff
@@ -1002,6 +1002,18 @@ subtest 'priority correctly assigned when posting job' => sub {
     $t->json_is('/job/group', 'opensuse', 'group assigned (1)');
     $t->json_is('/job/priority', 50, 'global default priority assigned');
 
+    my $max = 7300;
+    $jobs_post_params{MAX_JOB_TIME} = $max;
+    $t->post_ok('/api/v1/jobs', form => \%jobs_post_params)->status_is(200);
+    $t->get_ok('/api/v1/jobs/' . $t->tx->res->json->{id})->status_is(200);
+    $t->json_is('/job/priority', 50 + $max / 100, 'Increased priority due to high MAX_JOB_TIME');
```
okurz:
```suggestion
    $t->json_is('/job/priority', 50 + $max / 100, 'Increased prio value due to high MAX_JOB_TIME');
```

as the effective priority is *de*-creased, not increased. Or just say "prio" as that is how we tried to consistently call the value. If "prio" goes up, "priority" goes down and vice-versa.

okurz (changes requested):

etc/openqa/openqa.ini

lib/OpenQA/Schema/ResultSet/Jobs.pm
perlpunk: > to handle 0 and effectively disable the feature.

done

> By the way the function you are adding this code to is already rather big. How about splitting it?

on it

perlpunk: I split out `_handle_dependency_settings`. The method now fits into my window at least :)
For more I would have to think about what else to split...

t/api/04-jobs.t
perlpunk: done

PR #6847, opened 2025-11-19
Ensure that /var/lib/openqa/cache/tmp is always present

aylen384:


okurz (approved):

lib/OpenQA/CacheService/Model/Cache.pm
```diff
@@ -70,8 +70,8 @@ sub init ($self) {
     unless (-e $db_file) {
         $log->info(qq{Creating cache directory tree for "$location"});
         $location->remove_tree({keep_root => 1});
-        $location->child('tmp')->make_path;
     }
+	$location->child('tmp')->make_path;
```
d3flex: Maybe 
```suggestion
	my $tmp = $location->child('tmp'); 
	$tmp->make_path unless $tmp->exists;
```
Why to call `make_path` if it does exist

d3flex: you might also need to check whether is dir or not.

aylen384: > Why to call make_path if it does exist

Because it will check if the dir is there and create it if needed. No need to explicitly code that.


> you might also need to check whether is dir or not.

If someone actually creates a file there called `tmp`, it will raise an error which is what we want in this case.

Martchus (approved):

lib/OpenQA/CacheService/Model/Cache.pm
d3flex: > Because it will check if the dir is there and create it if needed. No need to explicitly code that.

Does the `make_path` checks the existence? it wont try to create it anyway?

Martchus: @d3flex It is best to let `make_path` handle all of it. Why replicating the logic to check if it is already there and handle the case when it is already there but a file? `make_path` handles all those cases and behaves exactly like we want in any of them.

perlpunk (approved):

lib/OpenQA/CacheService/Model/Cache.pm
perlpunk: perltidy complains.
I just had a closer look and you actually have a tab there.

```suggestion
    $location->child('tmp')->make_path;
```

d3flex: you guys are right. I just check that to make sure.

d3flex (approved):

PR #6848, opened 2025-11-21
Add debug message about priority malus

perlpunk:
Since we don't know the original value (or cannot find out very easily), this will help us to track the feature.

### DIFF
--- a/lib/OpenQA/Schema/ResultSet/Jobs.pm
+++ b/lib/OpenQA/Schema/ResultSet/Jobs.pm
@@ -157,12 +157,18 @@ sub create_from_settings ($self, $settings, $scheduled_product_id = undef) {
     # assign scheduled product
     $new_job_args{scheduled_product_id} = $scheduled_product_id;
 
+    my $debug_msg;
     if (looks_like_number($settings{MAX_JOB_TIME}) and $settings{MAX_JOB_TIME} > DEFAULT_MAX_JOB_TIME) {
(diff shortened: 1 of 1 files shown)


lib/OpenQA/Schema/ResultSet/Jobs.pm
```diff
@@ -157,12 +157,17 @@ sub create_from_settings ($self, $settings, $scheduled_product_id = undef) {
...
-        $new_job_args{priority} += int($settings{MAX_JOB_TIME} / $scale) if $scale;
+        if (my $scale = OpenQA::App->singleton->config->{misc_limits}->{max_job_time_prio_scale}) {
+            my $malus = int($settings{MAX_JOB_TIME} / $scale);
+            $debug_msg = sprintf 'Adding priority malus to newly created job (old: %d, malus: %s)', $new_job_args{priority}, $malus;
+            $new_job_args{priority} += $malus;
+        }
     }
 
     my $job = $self->create(\%new_job_args);
+    log_info(sprintf "(Job %d) $debug_msg", $job->id) if $debug_msg;
```
okurz: This can be thousand of lines. I recommend against "info" and use "debug". That's also what the commit says: "debug"

perlpunk: Yes, that was accidental. fixed

Martchus (approved):

PR #6850, opened 2025-11-21
Fix possibly excessive memory use when computer test result overview

Martchus:
When accessing the test results overview page (especially without parameters, e.g. http://localhost:9526/tests/overview) a very huge set of jobs is considered. This leads to a very excessive memory usage, e.g. with current OSD data the RSS rises over 7 GiB and the request takes very long. There's already a limit of 2000 jobs which is also effective but unfortunately not passed to `complex_query` and only used later on. I tracked the memory usage while running the code with debug printing it it was very clear that `complex_query` is the culprit.

This change uses the limit from the start. This means additional filtering done in Perl code also needs to happen as part of the now limited initial query. So this change replaces these Perl loops/grep with passing the filter parameters directly to the database. This slightly changes the order of things but should lead to the same end result.

---

Still a draft as I'll also have to take care of other places where we use `complex_query` in a similar way. I probably also still need to change some details depending on test results. If I took care of other places I can probably also remove code that is then no longer required, e.g. `latest_jobs`.

---

Related ticket: https://progress.opensuse.org/issues/192448

### DIFF
--- a/lib/OpenQA/Schema/ResultSet/Jobs.pm
+++ b/lib/OpenQA/Schema/ResultSet/Jobs.pm
@@ -23,6 +23,7 @@ use Mojolicious::Validator;
 use Mojolicious::Validator::Validation;
 use Time::HiRes 'time';
 use DateTime;
+use List::Util qw(any);
 use Scalar::Util qw(looks_like_number);
 
 =head2 latest_build
@@ -55,20 +56,16 @@ sub latest_build ($self, %args) {
     $attrs{order_by} = {-desc => 'me.id'};    # More reliable for tests than t_created
     $attrs{columns} = qw(BUILD);
 
+    my $job_settings = $schema->resultset('JobSettings');
     foreach my $key (keys %args) {
+        my $key_uc = uc $key;
         my $value = $args{$key};
-
-        if (grep { $key eq $_ } qw(distri version flavor machine arch build test)) {
-            push(@conds, {'me.' . uc($key) => $value});
+        if (any { $key_uc eq $_ } OpenQA::Schema::Result::Jobs::MAIN_SETTINGS) {
+            push @conds, {'me.' . $key_uc => $value};
         }
         else {
-
-            my $subquery = $schema->resultset('JobSettings')->search(
-                {
-                    key => uc($key),
-                    value => $value
-                });
-            push(@conds, {'me.id' => {-in => $subquery->get_column('job_id')->as_query}});
+            my $subquery = $job_settings->search({key => $key_uc, value => $value});
+            push @conds, {'me.id' => {-in => $subquery->get_column('job_id')->as_query}};
         }
     }
 
@@ -97,16 +94,8 @@ sub latest_jobs ($self, $until = undef) {
     my @latest;
     my %seen;
     foreach my $job (@jobs) {
-        my $test = $job->TEST;
-        my $distri = $job->DISTRI;
-        my $version = $job->VERSION;
-        my $build = $job->BUILD;
-        my $flavor = $job->FLAVOR;
-        my $arch = $job->ARCH;
-        my $machine = $job->MACHINE // '';
-        my $key = "$distri-$version-$build-$test-$flavor-$arch-$machine";
-        next if $seen{$key}++;
-        push(@latest, $job);
+        my $key = join('-', map { $job->$_ // '' } OpenQA::Schema::Result::Jobs::MAIN_SETTINGS);
+        push @latest, $job unless $seen{$key}++;
     }
 
     return @latest;
@@ -379,7 +368,7 @@ sub complex_query_latest_ids ($self, %args) {
     $attrs->{order_by} = \['max(me.id) DESC'];
     $attrs->{select} = ['max(me.id)'];
     $attrs->{as} = ['id'];
-    $attrs->{group_by} = [qw(TEST DISTRI VERSION BUILD FLAVOR ARCH MACHINE)];
+    $attrs->{group_by} = [OpenQA::Schema::Result::Jobs::MAIN_SETTINGS];
 
     # execute the search; use a sub query if filtering is enabled
     my $search = $self->search({-and => $conds}, $attrs);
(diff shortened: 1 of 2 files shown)


Martchus:
Looks like the current code determines the latest build automatically when no `build` is specified. Since that actually seems to be a wanted feature I made my change behave similarly. With this there is *almost* no change in behavior anymore. Some tests still fail which I'll have to work on tomorrow.

I also added the following to the commit message because this in one of the remaining differences and I haven't found a good way to avoid it:

> With this change the "preferred" machine (basically the most frequent machine per architecture) is only determined based on the jobs that are actually displayed. That means a jobs will no longer show as e.g. `kde@uefi` when filtering for `machine=uefi` because with this filter the machine `uefi` becomes the preferred (most frequent) machine (as now all displayed jobs have that machine). I think this change is acceptable. It means that some tests had to be adjusted accordingly.

lib/OpenQA/Schema/ResultSet/Jobs.pm
```diff
@@ -353,16 +353,53 @@ sub _prepare_complex_query_search_args ($self, $args) {
     return (\@conds, \%attrs);
 }
 
-sub complex_query ($self, %args) {
-    # For args where we accept a list of values, allow passing either an
-    # array ref or a comma-separated list
+sub _accept_comman_separated_arg_values ($args) {
```
```suggestion
sub _accept_comma_separated_arg_values ($args) {
```

t/10-tests_overview.t
```diff
@@ -151,10 +151,8 @@ subtest 'time parameter' => sub {
 };
 
 subtest 'limit parameter' => sub {
-    $t->get_ok(
-        '/tests/overview?distri=opensuse&version=Factory&limit=2&limit=2',
-        'no database error when specifying more than one limit'
-    );
+    $t->get_ok('/tests/overview?distri=opensuse&version=Factory&build=0048&limit=2&limit=2',
+        'no database error when specifying more than one limit');
```
okurz, line 155: why is this change necessary? Not specifying the build should still work to resolve the latest build

Martchus: And it still works. The problem is that this is `0048` vs. `0048@0815` (where `@0815` is part of the `BUILD`, *not* some appended `MACHINE`). The sorting between those two doesn't seem stable and apparently changes slightly with my change (even though the logic/query to determine the latest build hasn't changed). In my opinion this test shouldn't depend on that and rather explicitly select a build that has enough jobs to be able to test whether the `limit` parameter is effective.

There are other tests which check this explicitly (and probably less ambiguously). I know this because at some point detecting the latest build was actually broken with my changes and more tests failed. I think these other tests are sufficient and we can change this test which is just about limiting anyway.

okurz (approved):

lib/OpenQA/Schema/ResultSet/Jobs.pm
```diff
@@ -353,16 +353,53 @@ sub _prepare_complex_query_search_args ($self, $args) {
...
+    _accept_comma_separated_arg_values(\%args);
     my ($conds, $attrs) = $self->_prepare_complex_query_search_args(\%args);
-    my $jobs = $self->search({-and => $conds}, $attrs);
-    return $jobs;
+    my $filters = $args{filters};
+    my $has_filters = $filters && @$filters > 0;
+    my $rows = $has_filters ? delete $attrs->{rows} : undef;  # when filtering, limit rows only in outer/filtering query
+    if (my $until = $args{until}) { push @$conds, {'me.t_created' => {'<=' => $until}} }
+
+    # set attributes to return the only the latest job IDs for a certain combination of TEST, DISTRI, VERSION, …
```
d3flex: `to return the only the latest` should be `to return only the latest`, right?

lib/OpenQA/Schema/ResultSet/Jobs.pm
```diff
@@ -353,16 +353,53 @@ sub _prepare_complex_query_search_args ($self, $args) {
...
+    my $filters = $args{filters};
+    my $has_filters = $filters && @$filters > 0;
+    my $rows = $has_filters ? delete $attrs->{rows} : undef;  # when filtering, limit rows only in outer/filtering query
+    if (my $until = $args{until}) { push @$conds, {'me.t_created' => {'<=' => $until}} }
+
+    # set attributes to return only the latest job IDs for a certain combination of TEST, DISTRI, VERSION, …
+    $attrs->{order_by} = \['max(me.id) DESC'];
+    $attrs->{select} = ['max(me.id)'];
+    $attrs->{as} = ['id'];
+    $attrs->{group_by} = [qw(TEST DISTRI VERSION BUILD FLAVOR ARCH MACHINE)];
```
d3flex: this set of variables are commonly used. they are already used in this file. What I am thinking is to create a constant.  it might be a subset of (TEST DISTRI VERSION BUILD FLAVOR ARCH MACHINE), as it depends what you will find in the other occurences. 

However a search shows similar results in other files too, which is true especially for `qw(DISTRI VERSION FLAVOR ARCH)` (this is the smallest set I found). So there is also the idea to make this a constant in the Constants.pm. Maybe apply it only in this case for now to keep it simple and easy.

wdyt?

lib/OpenQA/WebAPI/Controller/Test.pm
```diff
@@ -699,25 +698,8 @@ sub _prepare_job_results ($self, $all_jobs, $limit) {
...
-          and (not $results or $results->{$_->result})
-          and (not $archs or $archs->{$_->ARCH})
-          and (not $machines or $machines->{$_->MACHINE})
-          and (not $failed_modules or $_->result eq OpenQA::Jobs::Constants::FAILED)
-    } @$all_jobs;
-    my $limit_exceeded = @jobs >= $limit;
-    @jobs = @jobs[0 .. ($limit - 1)] if $limit_exceeded;
-    my @jobids = map { $_->id } @jobs;
+    my @jobs = $jobs->all;
+    my $preferred_machines = _calculate_preferred_machines(\@jobs);
```
d3flex, line 702: just a note. The `$jobs` are used here only for calculate the `$preferred_machines` which is used later for this condition
``` if (   $job->MACHINE
            && $preferred_machines->{$job->ARCH}
            && $preferred_machines->{$job->ARCH} ne $job->MACHINE)```

So the `_prepare_job_results` needs this only and not the whole ResultSet. If you call `_calculate_preferred_machines(\@jobs);` before https://github.com/Martchus/openQA/blob/79f52e8c934da96eaa20e95fca56a8891dfc361a/lib/OpenQA/WebAPI/Controller/Test.pm#L848C5-L848C94 and just pass this. I think that way, it would simplify the dependencies between the invocations and the functions would be more clear. 

But lets not change anything now.

Martchus: > The $jobs are used here only for calculate …

The "only" in this sentence makes it wrong. Just search for `@jobs` in the subsequent code of this function; you'll find 4 occurrences.

PR #6852, opened 2025-11-24
Fix ordering of job groups after 2ad929ceca43d

Martchus:


lib/OpenQA/Schema/ResultSet/JobGroupParents.pm
```diff
@@ -10,7 +10,9 @@ use Mojo::Base 'DBIx::Class::ResultSet';
 sub job_groups_and_parents {
     my $self = shift;
 
-    my $params = {order_by => [{-asc => 'me.sort_order'}, {-asc => 'me.name'}], prefetch => 'children'};
+    my @order
+      = ({-asc => 'me.sort_order'}, {-asc => 'children.sort_order'}, {-asc => 'me.name'}, {-asc => 'children.name'});
+    my $params = {order_by => \@order, prefetch => 'children'};
```
okurz: With the not so nice line-wrapping how about


```suggestion
    my @order = map { {-asc=>$_} } qw(me.sort_order children.sort_order me.name children.name);
    my $params = {order_by => \@order, prefetch => 'children'};
```

Martchus: `-asc` is actually the default so we can just use the array without map.

lib/OpenQA/Schema/ResultSet/JobGroupParents.pm
```diff
@@ -10,7 +10,8 @@ use Mojo::Base 'DBIx::Class::ResultSet';
 sub job_groups_and_parents {
     my $self = shift;
 
-    my $params = {order_by => [{-asc => 'me.sort_order'}, {-asc => 'me.name'}], prefetch => 'children'};
+    my @order = (qw(me.sort_order children.sort_order me.name children.name));
```
perlpunk:
```suggestion
    my @order = qw(me.sort_order children.sort_order me.name children.name);
```

d3flex (approved):

PR #6854, opened 2025-11-26
Add task to limit scheduled products similar to audit events

Martchus:
This task deletes all scheduled products where the created jobs have
already been deleted. So the retention of scheduled products is basically
the retention of their jobs and doesn't need to be configured separately.

We still need to prevent deleting scheduled products that didn't create any
jobs immediately (e.g. due to an error). Hence this change still ensures
that scheduled products are kept at least 4 weeks (which is also
configurable but probably doesn't need to be changed in practice).

Related ticket: https://progress.opensuse.org/issues/193120

### DIFF
--- a/lib/OpenQA/Schema/ResultSet/ScheduledProducts.pm
+++ b/lib/OpenQA/Schema/ResultSet/ScheduledProducts.pm
@@ -6,8 +6,6 @@ package OpenQA::Schema::ResultSet::ScheduledProducts;
 use Mojo::Base 'DBIx::Class::ResultSet', -signatures;
 use OpenQA::Schema::Result::ScheduledProducts qw(CANCELLED);
 use OpenQA::App;
-use Time::Piece;
-use Time::ParseDate;
 
 sub create_with_event ($self, $params, $user, $webhook_id = undef) {
     my $scheduled_product = $self->create(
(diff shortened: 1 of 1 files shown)


okurz:
Nice! Just the packaging problem:

```
[  718s] RPM build errors:
[  718s]     Installed (but unpackaged) file(s) found:
[  718s]    /usr/share/openqa/script/openqa-enqueue-scheduled-product-cleanup
```

lib/OpenQA/Schema/ResultSet/ScheduledProducts.pm
```diff
@@ -118,4 +120,21 @@ sub job_statistics ($self, $distri, $version, $flavor, $arch, $build) {
...
+    $time_constraint = localtime($time_constraint)->ymd;
+    my $sth = $self->result_source->schema->storage->dbh->prepare(
+        <<~'END_SQL'
+        DELETE FROM scheduled_products
+            WHERE
+                (t_created < ?)
+                AND (SELECT count(id) FROM jobs WHERE jobs.scheduled_product_id = scheduled_products.id LIMIT 1) = 0
+        END_SQL
+    );
+    $sth->execute($time_constraint);
```
perlpunk:
```suggestion
                (t_created < current_date - interval '?'
                AND (SELECT count(id) FROM jobs WHERE jobs.scheduled_product_id = scheduled_products.id LIMIT 1) = 0
        END_SQL
    );
    $sth->execute("$min_storage_duration days");
```
Then we wouldn't need Time::ParseDate

perlpunk: I see it's already used in the code for AuditEvents as well. But maybe we can keep it in mind for a future improvement

PR #6860, opened 2025-11-28
Avoid partial deletion of a screenshot if Minion job is aborted

Martchus:
See individual commit messages and https://progress.opensuse.org/issues/187953.

**I haven't tested this for real**, hence this is still a draft.

I also considered scanning for "untracked" screenshots (like we have scanning for untracked assets). This would also be useful but considering that the scanning would take hours to complete in production (e.g. over 15 hours on OSD), it is probably not the best approach to solve the specific ACs we defined in https://progress.opensuse.org/issues/187953.

I also considered adding a callback in the existing `SignalGuard`. This is [very easy to do](https://github.com/os-autoinst/openQA/compare/master...Martchus:openQA:signal-guard-callback?expand=1) but also not very helpful. In the context where we can easily assign the callback (the outer function of the Minion task) we can't do much within this callback as we don't really know what's currently being processed. In the context where we do the processing (where this PR adds the `SignalBlocker`) we could so something meaningful by assigning a callback but we don't have the `SignalGuard` object at hand. So we needed to pass it there or add some global variable - both not nice - especially when we have to take care not to leave any copies so the destructor is still called as expected.

I also considered improving the order in which we do things and also the error handling. Maybe we should do an upfront DB query to know whether we can delete the screenshot, do the deletion on disk (not continuing if that fails) and only as the last step delete the database entry. Of the the deletion of the database entry can then still fail. So there is still a race condition to solve here. Hence I haven't looked further into that solution. However, it would be a bit more effective as it would also help with the case when the screenshots cannot be deleted on disk (e.g. due to a permission error) or when the cleanup job is interrupted more forcefully than receiving `SIGTERM` or `SIGINT`.

### DIFF
--- a/lib/OpenQA/SignalBlocker.pm
+++ b/lib/OpenQA/SignalBlocker.pm
@@ -5,7 +5,7 @@ package OpenQA::SignalBlocker;
 use Mojo::Base -base, -signatures;
 use Scalar::Util qw(weaken);
 
-my @SIGNALS = qw(TERM INT);
+my @SIGNALS = qw(TERM INT QUIT);
 
 # delays execution of signal handlers while an instance exists
 sub new ($class, @attributes) {
--- a/lib/OpenQA/ScreenshotDeletion.pm
+++ b/lib/OpenQA/ScreenshotDeletion.pm
@@ -7,6 +7,7 @@ use Mojo::Base -base;
 use File::Basename qw(basename dirname);
 use File::Spec::Functions qw(catfile);
 use OpenQA::Log qw(log_debug);
+use OpenQA::SignalBlocker;
 use OpenQA::Utils qw(imagesdir);
 use Feature::Compat::Try;
 
@@ -30,6 +31,9 @@ sub delete_screenshot {
     # note: This might fail due to foreign key violation because a new screenshot link might
     #       have just been created. In this case the screenshot should not be deleted in the
     #       database or the file system.
+    #       Using OpenQA::SignalBlocker to delay handling cancellation signals to avoid doing
+    #       only some of the deletions if the Minion job is aborted.
+    my $signal_blocker = OpenQA::SignalBlocker->new;
     try { $self->{_deletion_query}->execute($screenshot_id) }
     catch ($e) { return undef }
     # keep track of the deleted size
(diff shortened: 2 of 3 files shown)


lib/OpenQA/SignalBlocker.pm
```diff
@@ -0,0 +1,35 @@
+# Copyright SUSE LLC
+# SPDX-License-Identifier: GPL-2.0-or-later
+
+package OpenQA::SignalBlocker;
+use Mojo::Base -base, -signatures;
```
d3flex, line 5: I checked with other modules because I was not sure but it seems like `-base` can be removed.

lib/OpenQA/SignalBlocker.pm
```diff
@@ -0,0 +1,35 @@
...
+use Scalar::Util qw(weaken);
+
+my @SIGNALS = qw(TERM INT);
+
+# delays execution of signal handlers while an instance exists
+sub new ($class, @attributes) {
+    my $self = $class->SUPER::new(@attributes);
```
d3flex, line 12: Also not sure about the convension and for sure not a big deal but `cls` is one is used in other languages. feel free to close the comment...

lib/OpenQA/SignalBlocker.pm
okurz: `-base` is correct when defining a class in this package which is not based on any other class. If a file does not define a class then one can use `use Mojo::Base -strict, -signatures;`. If inheriting from a class Foo then `use Mojo::Base 'Foo', -signatures;`.

https://docs.mojolicious.org/Mojo/Base for details

lib/OpenQA/SignalBlocker.pm
perlpunk: Using `$class` is common in Perl.

lib/OpenQA/SignalBlocker.pm
```diff
@@ -0,0 +1,35 @@
+# Copyright SUSE LLC
+# SPDX-License-Identifier: GPL-2.0-or-later
+
+package OpenQA::SignalBlocker;
```
perlpunk, line 4: Why a completely new class and not reuse `OpenQA::Task::SignalGuard`?

Martchus: Because it does something different. (Both classes are dealing with signals and both use their destructor for automatic cleanup. However, that's it with the similarities. So I think it is better to keep `SignalGuard` simple and not do two different things.)

lib/OpenQA/SignalBlocker.pm
Martchus: I think in Perl we use `$class` (or at least in this project we do).

kalikiana (approved):

PR #6861, opened 2025-11-28
Also use TIMEOUT_SCALE for priority malus calculation

perlpunk:


okurz (approved):
It's fine though I would avoid the word "malus". Either use positive phrasing "bonus" or neutral "scaling" :)

PR #6865, opened 2025-12-05
feat: Allow to configure key+secret with env variables

okurz:
This is a better variant to
https://github.com/os-autoinst/openQA/pull/6767 which only adds the
support in the worker authentication. If we change the user agent then
we can enable clients as well as workers to use environment variables.

This commit adds support for optional environment variables
OPENQA_API_KEY and OPENQA_API_SECRET which if specified take precedence
over config file entries. Command line arguments take precedence over
environment variables but are now explicitly discouraged in help texts
and documentation to reduce the risk of leaking credentials.

### DIFF
--- a/docs/Client.asciidoc
+++ b/docs/Client.asciidoc
@@ -82,8 +82,12 @@ key = 1234567890ABCDEF
 secret = ABCDEF1234567890
 ----
 
-For ad-hoc use all `openqa-cli` subcommands use the `--apikey` and `--apisecret`
-options. Which will override whatever the config files may contain.
+Alternatively specify key and secret in environment variables `OPENQA_API_KEY`
+and `OPENQA_API_SECRET`.
+
+For ad-hoc use all `openqa-cli` subcommands use the `--apikey` and
+`--apisecret` options which will override whatever the environment or config
+files may contain.
 
 [source,sh]
 ----
--- a/t/api/03-auth.t
+++ b/t/api/03-auth.t
@@ -49,6 +49,16 @@ subtest 'authentication routes for plugins' => sub {
 
 client($t, apikey => undef, apisecret => undef);
 
+subtest 'key+secret from env variables' => sub {
+    my $t_new = Test::Mojo->new('OpenQA::WebAPI');
+    $t_new->delete_ok('/api/v1/assets/1')->status_is(403, 'clean user agent has no credentials');
+    local $ENV{OPENQA_API_KEY} = 'LANCELOTKEY01';
+    local $ENV{OPENQA_API_SECRET} = 'MANYPEOPLEKNOW';
+    client($t_new, apikey => undef, apisecret => undef);
+    $t_new->delete_ok('/api/v1/assets/1')->status_is(403)
+      ->json_is('/error' => 'Administrator level required', 'key read from env (but missing administrator level)');
+};
+
 subtest 'access limiting for non authenticated users' => sub {
     $t->get_ok('/api/v1/jobs')->status_is(200);
     is $t->tx->res->headers->access_control_allow_origin, '*', 'CORS header set for non authenticated routes';
--- a/lib/OpenQA/UserAgent.pm
+++ b/lib/OpenQA/UserAgent.pm
@@ -15,6 +15,8 @@ has [qw(apikey apisecret base_url)];
 sub new {
     my $self = shift->SUPER::new(@_);
     my %args = @_;
+    $args{apikey} //= $ENV{OPENQA_API_KEY};
+    $args{apisecret} //= $ENV{OPENQA_API_SECRET};
     for my $i (qw(apikey apisecret)) {
         $self->$i($args{$i}) if $args{$i};
     }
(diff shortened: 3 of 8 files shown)


t/api/03-auth.t
```diff
@@ -49,6 +49,18 @@ subtest 'authentication routes for plugins' => sub {
...
+subtest 'key+secret from env variables' => sub {
+    my $t_new = Test::Mojo->new('OpenQA::WebAPI');
+    $t_new->delete_ok('/api/v1/assets/1')->status_is(403, 'clean user agent has no credentials');
+    local $ENV{OPENQA_API_KEY} = 'PERCIVALKEY02';
+    local $ENV{OPENQA_API_SECRET} = 'PERCIVALSECRET02';
+    $t_new->delete_ok('/api/v1/assets/1')->status_is(403)
+      ->json_is('/error' => 'no api key', 'key read from env (but missing administrator level)');
+    delete $ENV{OPENQA_API_KEY};
+    delete $ENV{OPENQA_API_SECRET};
+    client($t, apikey => undef, apisecret => undef);
```
d3flex: Could you explain what is the indention with the last lines.  they are not really needed. I run the test and it just fine without anyways.

d3flex (changes requested):
```diff --git a/t/api/03-auth.t b/t/api/03-auth.t
index 61b5fa724..71a49f8b0 100644
--- a/t/api/03-auth.t
+++ b/t/api/03-auth.t
@@ -52,13 +52,11 @@ client($t, apikey => undef, apisecret => undef);
 subtest 'key+secret from env variables' => sub {
     my $t_new = Test::Mojo->new('OpenQA::WebAPI');
     $t_new->delete_ok('/api/v1/assets/1')->status_is(403, 'clean user agent has no credentials');
-    local $ENV{OPENQA_API_KEY} = 'PERCIVALKEY02';
-    local $ENV{OPENQA_API_SECRET} = 'PERCIVALSECRET02';
+
     $t_new->delete_ok('/api/v1/assets/1')->status_is(403)
       ->json_is('/error' => 'no api key', 'key read from env (but missing administrator level)');
-    delete $ENV{OPENQA_API_KEY};
-    delete $ENV{OPENQA_API_SECRET};
-    client($t, apikey => undef, apisecret => undef);
+
+
 };```

and test still pass, so I assume it doesnt really test the variables. sorry if I am mistaken anything here

File: t/api/03-auth.t (hunk 1 above)
Comment (okurz) [reply t/api/03-auth.t]: Those lines are clean up so that the next tests are not impacted by creating a new user agent with the default testing key+secret. We can also remove those lines which would make tests not independent anymore but for now they would still work

**Comment by okurz** [comment]
> ```diff
> index 61b5fa724..71a49f8b0 100644
> --- a/t/api/03-auth.t
> +++ b/t/api/03-auth.t
> @@ -52,13 +52,11 @@ client($t, apikey => undef, apisecret => undef);
>  subtest 'key+secret from env variables' => sub {
>      my $t_new = Test::Mojo->new('OpenQA::WebAPI');
>      $t_new->delete_ok('/api/v1/assets/1')->status_is(403, 'clean user agent has no credentials');
> -    local $ENV{OPENQA_API_KEY} = 'PERCIVALKEY02';
> -    local $ENV{OPENQA_API_SECRET} = 'PERCIVALSECRET02';
> +
>      $t_new->delete_ok('/api/v1/assets/1')->status_is(403)
>        ->json_is('/error' => 'no api key', 'key read from env (but missing administrator level)');
> -    delete $ENV{OPENQA_API_KEY};
> -    delete $ENV{OPENQA_API_SECRET};
> -    client($t, apikey => undef, apisecret => undef);
> +
> +
>  };```
> 
> and test still pass, so I assume it doesnt really test the variables. sorry if I am mistaken anything here
> ```

You are right. I messed up two different ideas. I wanted to use a dedicated new test instance with separate user agent that ensures that we use really different credentials and also ensure cleanup after the subtest. I know changed the test to use the actually created `$t_new` and also use the "Lancelot" key which is no operator and no admin so that the different used key is obvious. This now also improves the test to not need the cleanup you mentioned in https://github.com/os-autoinst/openQA/pull/6865#discussion_r2594700787

File: t/api/03-auth.t (hunk 1 above)
Comment (okurz) [reply t/api/03-auth.t]: Improved with https://github.com/os-autoinst/openQA/pull/6865#issuecomment-3619923485

File: docs/Client.asciidoc [hunk 2]
```diff
@@ -82,8 +82,12 @@ key = 1234567890ABCDEF
 secret = ABCDEF1234567890
 ----
 
-For ad-hoc use all `openqa-cli` subcommands use the `--apikey` and `--apisecret`
-options. Which will override whatever the config files may contain.
+Alternatively specify key and secret in environment variables `OPENQA_API_KEY`
+and `OPENQA_API_SECRET`.
+
+For ad-hoc use all `openqa-cli` subcommands use the `--apikey` and
+`--apisecret` options. Which will override whatever the environment or config
```
Martchus: Since you're changing this anyway:

```suggestion
`--apisecret` options which will override whatever the environment or config
```

okurz: done

PR #6867, opened 2025-12-08
React to stop signal in OpenQA::Task::Needle::Delete

perlpunk:
OpenQA::Task::SignalGuard now supports a callback that can be used in a minion task to have more control over what should happen.

This way we can abort after every single needle and save the already processed needle into the minion job notes.

Maybe we should have called the module
You::are::about::to::be::shot::do::you::have::any::last::Words instead.

Still needs tests

### DIFF
--- a/lib/OpenQA/Task/SignalGuard.pm
+++ b/lib/OpenQA/Task/SignalGuard.pm
@@ -17,6 +17,10 @@ has retry => 1;
 #       action twice.
 has abort => 0;
 
+# Code to execute when the minion server receives a signal to stop the server
+# If the call returns true, we just return and let the task handle it
+has 'callback';
+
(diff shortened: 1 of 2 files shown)


lib/OpenQA/Task/SignalGuard.pm
```diff
@@ -33,6 +37,9 @@ sub new ($class, $job, @attributes) {
 }
 
 sub _handle_signal ($self_weak, $signal) {
+    if (my $cb = $self_weak->callback) {
+        return if $cb->($signal);
```
Martchus, line 41: Not sure whether it'll be useful at some point:
```suggestion
        return if $cb->($self_weak, $signal);
```

This would be consistent with callbacks in e.g. Mojo::UserAgent which always pass the agent object as first argument.

PR #6869, opened 2025-12-09
Add scanner for untracked screenshots

Martchus:
Related ticket: https://progress.opensuse.org/issues/187953

---

I tested this also manually:

```
$ OPENQA_LOGFILE=delete-log.txt script/openqa eval 'say(STDERR app->schema->resultset("Screenshots")->scan_untracked_screenshots)' > to-delete.txt
0
$ cat to-delete.txt 
[INFO] Reading openqa config from: /hdd/openqa-devel/config/openqa.ini /hdd/openqa-devel/config/openqa.ini.d/01-test.ini
/hdd/openqa-devel/openqa/images/0a4/a66/0647f51a4bef105e96ad0fad32.png
/hdd/openqa-devel/openqa/images/f62/868/e3806bbf8f1fc12756a4dc6ab2.png
```

It isn't nice that we get this info message in the list of files but I guess that's another more general problem and we can live with that here. Otherwise, the list of files looks good. (I moved most of the images to a different directory so the list of present images becomes manageable. With this it found only 2 images which were really not in the DB anymore.)

I would probably run this only in dry mode in production, see what files are considered, do a sanity check on some of them and only delete the files afterwards.

I hope holding all those paths in memory at the same time isn't too much as we have 32 GiB of RAM in production and usually only use < 10 GiB.

### DIFF
--- a/lib/OpenQA/Schema/ResultSet/Screenshots.pm
+++ b/lib/OpenQA/Schema/ResultSet/Screenshots.pm
@@ -1,4 +1,4 @@
-# Copyright 2019 SUSE LLC
+# Copyright SUSE LLC
 # SPDX-License-Identifier: GPL-2.0-or-later
 
 package OpenQA::Schema::ResultSet::Screenshots;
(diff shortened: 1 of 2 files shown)


lib/OpenQA/Schema/ResultSet/Screenshots.pm
```diff
@@ -1,12 +1,15 @@
...
 # SPDX-License-Identifier: GPL-2.0-or-later
 
 package OpenQA::Schema::ResultSet::Screenshots;
 
 
 use Mojo::Base 'DBIx::Class::ResultSet', -signatures;
 
-use OpenQA::Log qw(log_trace);
+use Mojo::File qw(path);
+use Mojo::Path;
```
perlpunk: I don't see this used anywhere

PR #6873, opened 2025-12-10
Worker: add codecov coverage on OVS DBus checks

m-dati:
Improve test coverage in Worker.pm on DBus check

ticket: https://progress.opensuse.org/issues/193549

### DIFF
--- a/t/24-worker-overall.t
+++ b/t/24-worker-overall.t
@@ -16,6 +16,7 @@ use Mojo::Util 'scope_guard';
 use Mojolicious;
 use Test::Output qw(combined_like combined_from);
 use Test::MockModule;
+use Test::MockObject;
 use OpenQA::Constants qw(WORKER_COMMAND_QUIT WORKER_SR_API_FAILURE WORKER_SR_DIED WORKER_SR_DONE WORKER_SR_FINISH_OFF);
 use OpenQA::Worker;
 use OpenQA::Worker::Job;
@@ -593,6 +594,16 @@ subtest 'check availability of Open vSwitch related D-Bus service' => sub {
     $worker->settings->global_settings->{WORKER_CLASS} = 'foo,bar';
     is $worker->check_availability, undef, 'worker considered always available if not a tap worker';
 };
+
+subtest 'check is_ovs_dbus_service_running of D-Bus service' => sub {
+    my $mock_net_dbus = Test::MockObject->new();
+    $worker->{_system_dbus} = $mock_net_dbus;
+    $mock_net_dbus->set_true('get_service');
+    is $worker->is_ovs_dbus_service_running, 1, 'OvS D-Bus service passed on passing get_service';
+    $mock_net_dbus->mock(get_service => sub { die "Fake error" });
+    is $worker->is_ovs_dbus_service_running, 0, 'OvS D-Bus service failed on failing get_service';
+};
+
 
 subtest 'handle client status changes' => sub {
     my $fake_client = OpenQA::Worker::WebUIConnection->new('some-host', {apikey => 'foo', apisecret => 'bar'});


t/24-worker-overall.t.tdy
perlpunk, line 1: Please delete this file.
How did you run tidy?
```
make tidy-perl
```

t/24-worker-overall.t
```diff
@@ -22,6 +22,8 @@ use OpenQA::Worker::Job;
 use OpenQA::Worker::WebUIConnection;
 use Socket qw(getaddrinfo);
 
+$ENV{MOCK_ERROR_OVS_DBUS_SERVICE} = undef;
```
perlpunk: If undef is the default, that line is not really necessary

t/24-worker-overall.t
```diff
@@ -90,7 +92,10 @@ package Test::FakeDBusObject {
 package Test::FakeDBus {
     use Mojo::Base -base, -signatures;
     has mock_service_value => 1;
-    sub get_service ($self, $service_name) { $self->mock_service_value }
+    sub get_service ($self, $service_name) {
+        die("OvS DBus Error triggered on: $service_name") if ($ENV{MOCK_ERROR_OVS_DBUS_SERVICE} == 1);
```
perlpunk:
```suggestion
        die "OvS DBus Error triggered on: $service_name" if $ENV{MOCK_ERROR_OVS_DBUS_SERVICE};
```
We can just check for "truth" here, we don't care if it's a certain number.
That will also get rid of the warning:
```
Use of uninitialized value $ENV{"MOCK_ERROR_OVS_DBUS_SERVICE"} in numeric eq (==)
```

t/24-worker-overall.t
m-dati: Ok. I did so to explicitly initialize that new parameter at higer level, being  used in a redefined routine called from inside a test and I was not sure of that parameter's scope.

t/24-worker-overall.t.tdy
m-dati: Sorry :) I forgot to remove that file

t/24-worker-overall.t
m-dati: Yes thanks: it noted it late yesterday evening :), going to update it

t/24-worker-overall.t.tdy
m-dati: done

t/24-worker-overall.t
m-dati: removed

t/24-worker-overall.t
m-dati: done

t/24-worker-overall.t
```diff
@@ -90,7 +90,10 @@ package Test::FakeDBusObject {
 package Test::FakeDBus {
     use Mojo::Base -base, -signatures;
     has mock_service_value => 1;
-    sub get_service ($self, $service_name) { $self->mock_service_value }
+    sub get_service ($self, $service_name) {
+        die "OvS DBus Error triggered on: $service_name" if $ENV{MOCK_ERROR_OVS_DBUS_SERVICE};
```
Martchus: We probably don't want this test to be influenced by the environment. So don't use an environment variable here. A normal global variable will do.

t/24-worker-overall.t.tdy
okurz: That file shouldn't even be left over. See @perlpunk 's comment about `make tidy-perl`

t/24-worker-overall.t
m-dati: ok

m-dati: To clarify: 
to make the coverage "enter" in the [last catch](https://app.codecov.io/gh/os-autoinst/openQA/blob/master/lib%2FOpenQA%2FWorker.pm#L643), I shall cause an error _on purpose_ in the above `try`: 
so, to trigger that error I put there, _like a "bomb"_ 😄, a parametr, global to the `24-worker-overall.t`, to tweak the status on-demand in the test flow, that should be visible in `get_service` and make it fail.

okurz (approved):

PR #6885, opened 2025-12-18
feat(OpenQA::Git): Cleanup git dir in commit() on shutdown

perlpunk:

OpenQA::Git::_run_cmd will throw an exception when
* The process received a signal
* or when a git command was aborted due to a signal

OpenQA::Git::commit will attempt to cleanup after itself if aborted by a signal in the middle.

The `delete_needles` task will abort and return an error message to the user
if it was stopped by a signal, but the task itself will succeed.

Next step:
* in `OpenQA::Task::Needle::Delete` we can now react to that exception and simply retry
* in `OpenQA::Task::Needle::Save` the same plus some cleanup before

### DIFF
--- a/lib/OpenQA/Task/Needle/Delete.pm
+++ b/lib/OpenQA/Task/Needle/Delete.pm
@@ -12,10 +12,10 @@ use Feature::Compat::Try;
 
 sub register {
     my ($self, $app) = @_;
-    $app->minion->add_task(delete_needles => sub { _delete_needles($app, @_) });
+    $app->minion->add_task(delete_needles => sub { _task_delete_needles($app, @_) });
 }
 
-sub _delete_needles ($app, $minion_job, $args) {
+sub _task_delete_needles ($app, $minion_job, $args) {
     # SignalGuard will prevent the delete task to interrupt with no recovery,
     # instead will retry once the gru server returned up and running. The popup
     # on the frontend will wait until the retried job finished.
@@ -42,13 +42,23 @@ sub _delete_needles ($app, $minion_job, $args) {
     }
 
     $signal_guard->retry(0);
+    _delete_needles($app, $user, \%to_remove, \@removed_ids, \@errors);
 
-    for my $dir (sort keys %to_remove) {
-        my $needles = $to_remove{$dir};
+
+    return $minion_job->finish(
+        {
+            removed_ids => \@removed_ids,
+            errors => \@errors
+        });
+}
+
+sub _delete_needles ($app, $user, $to_remove, $removed_ids, $errors) {
+  DIR: for my $dir (sort keys %$to_remove) {
+        my $needles = $to_remove->{$dir};
         # prevent multiple git tasks to run in parallel
         my $guard;
         unless ($guard = $app->minion->guard("git_clone_${dir}_task", 2 * ONE_HOUR)) {
-            push @errors,
+            push @$errors,
               {
                 id => $_,
                 message => "Another git task for $dir is ongoing. Try again later.",
@@ -63,27 +73,22 @@ sub _delete_needles ($app, $minion_job, $args) {
             }
             catch ($e) {
                 my $msg = "$e";
-                if (ref $e and $e->shutting_down) {
+                my $shutting_down = ref $e && $e->shutting_down;
+                if ($shutting_down) {
                     $msg = "Aborted due to server restart, please try again in a bit ($msg)";
                 }
-                push @errors,
+                push @$errors,
                   {
                     id => $needle_id,
                     display_name => $needle->filename,
                     message => $msg,
                   };
-                last if ref $e && $e->shutting_down;
+                last DIR if $shutting_down;
                 next;
             }
-            push @removed_ids, $needle_id;
+            push @$removed_ids, $needle_id;
         }
     }
-
-    return $minion_job->finish(
-        {
-            removed_ids => \@removed_ids,
-            errors => \@errors
-        });
 }
 
 1;
--- a/t/14-grutasks-git.t
+++ b/t/14-grutasks-git.t
@@ -21,6 +21,9 @@ use Mojo::Util qw(dumper scope_guard);
 use Mojo::File qw(path tempdir);
 use Time::Seconds;
 use File::Copy::Recursive qw(dircopy);
+use Test::MockModule;
+use Test::MockObject;
+use OpenQA::Error;
 
 # Avoid using tester's ~/.gitconfig
 delete $ENV{HOME};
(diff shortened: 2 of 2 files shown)


lib/OpenQA/Error/Cmd.pm
d3flex, line 1: I wonder why do we need this file? Why not just use lib/OpenQA/Error.pm?

perlpunk: Because `OpenQA::Error::Cmd` is specifically for when a command execution failed, and later we can check `$e->return_code for example`

okurz (changes requested):
missing patch coverage

perlpunk:
> missing patch coverage

should be fixed now.
Took me a while because I tried `Test::MockObject `first, but that won't work if you want to mock a method that is called `remove`, because that is a `Test::MockObject `method :-/

lib/OpenQA/Error/Cmd.pm
d3flex: Sorry but I do not see the point. the answer doesnt seem to say why it is really needed.
this introduces some sort of unnecessary complexity IMO. If I would change something it would be to rename this to lib/OpenQA/Error.pm - but it is just fine to keep it as is now - . It is a very simple class. But both, at least to me, look to share the same behavior for the same type of error. No need for abstraction.

d3flex (dismissed):

lib/OpenQA/Task/Needle/Delete.pm
```diff
@@ -12,17 +12,15 @@ use Feature::Compat::Try;
...
 
-sub _delete_needles ($app, $minion_job, $args) {
+sub _task_delete_needles ($app, $minion_job, $args) {
     # SignalGuard will prevent the delete task to interrupt with no recovery,
     # instead will retry once the gru server returned up and running. The popup
     # on the frontend will wait until the retried job finished.
     my $signal_guard = OpenQA::Task::SignalGuard->new($minion_job);
-    my $schema = $app->schema;
-    my $needles = $schema->resultset('Needles');
-    my $user = $schema->resultset('Users')->find($args->{user_id});
```
d3flex, line 25: I think there was a reason for this split.
And it should be that the following queries do not repeat a db request. Someone needs to confirm.

lib/OpenQA/Task/Needle/Delete.pm
```diff
@@ -62,23 +71,22 @@ sub _delete_needles ($app, $minion_job, $args) {
...
+                    $msg = "Aborted due to server restart, please try again in a bit ($msg)";
+                }
+                push @$errors,
                   {
                     id => $needle_id,
                     display_name => $needle->filename,
-                    message => ref $e ? $e->msg : $e,
+                    message => $msg,
                   };
+                last DIR if ref $e && $e->shutting_down;
```
d3flex: So here if there is an error and if an interruption like qw(INT TERM) go back to the next dir?!
I still try to make my mind around this but one comment to make it a little more readable is,

```
$interrupt_flag = ref $e and $e->shutting_down;
$interrupt_msg = "Aborted due to server restart, please try again in a bit ";
push @$errors,
{
{
id => $needle_id,
display_name => $needle->filename,
 message => $interrupt_flag ? $interrupt_msg . $e : $e,
};
 };
last DIR if $interrupt_flag;
```

lib/OpenQA/Task/Needle/Delete.pm
perlpunk: I don't understand. What split do you mean, and which db request would be repeated?

lib/OpenQA/Task/Needle/Delete.pm
perlpunk: Before my change it would do `next` after any error.
with my change is still does a `next` for errors, except if it was a signal error, then we know we should abort the task as soon as we can, so I do `last` instead.

lib/OpenQA/Error/Cmd.pm
perlpunk: I don't want to add `status`, `return_code`, `stdout` and `stderr` to an error class and just leave them empty sometimes. There can be many reasons for errors, and the most generic one is the superclass OpenQA::Error.
In the future there can be more kinds of errors, and I would hate to put all their possible attributes in one class.

lib/OpenQA/Task/Needle/Delete.pm
perlpunk: I put it into a variable `$shutting_down` now.

lib/OpenQA/Task/Needle/Delete.pm
d3flex: > and which db request would be repeated?

    my $schema = $app->schema;                                                                          # fetch schema once                                                
    my $needles = $schema->resultset('Needles');                                               # schema already have data
    my $user = $schema->resultset('Users')->find($args->{user_id});               # same
    
 instead:

        my $needles = $app->schema->resultset('Needles');                                 # in line 23 
        my $user = $app->schema->resultset('Users')->find($args->{user_id}); # line 43 -- goes again to fetch schema

lib/OpenQA/Error/Cmd.pm
d3flex: > In the future there can be more kinds of errors, and I would hate to put all their possible attributes in one class.

this is what I try to say.. We do not have "more kinds" at that moment. So it would be better to keep it simple.

> I don't want to add status, return_code, stdout and stderr to an error class and just leave them empty

Whats the problem with the variables been undef in this case? From my POV there is no behavioral difference. We present two type of errors but it is just different data, right?!

lib/OpenQA/Task/Needle/Delete.pm
perlpunk: There is no database request that is made twice.
It is just a method call. A method call in perl is not that horribly slow that we have to avoid it at all costs.

lib/OpenQA/Error/Cmd.pm
perlpunk: > this is what I try to say.. We do not have "more kinds" at that moment.

We already have two right now.

perlpunk: > Whats the problem with the variables been undef in this case? From my POV there is no behavioral difference. We present two type of errors but it is just different data, right?!

One error happens when a command was executed and failed, and one is a generic error that currently mainly has the function to check if the error was because of a signal.

Having objects with different data in the same class is not very clean.

Unless there are other votes for merging the two classes I'm not going to change it.

lib/OpenQA/Task/Needle/Delete.pm
d3flex: @perlpunk I phrase it wrong, calling it bd query. However there is no reason to do this change as it doesnt (as you say) have any significant performance impact, but the old version is a better practice. Additional and according to those arguments, it will keep the history better. it is another change, not directly connected to the scope, without some benefit. Can we leave it as is please?

t/14-grutasks-git.t
```diff
@@ -481,6 +483,22 @@ subtest 'delete_needles' => sub {
...
     };
+
+    subtest 'error handling' => sub {
+        my $user = Test::MockObject->new;
+        my @needles = map { $schema->resultset('Needles')->new({id => $_}) } (23, 42);
+        my $mock_needle = Test::MockModule->new('OpenQA::Schema::Result::Needles');
+        $mock_needle->redefine(
+            remove => sub ($self, $user) {
+                return if $self->id == 23;
+                die OpenQA::Error->new(signal => 'INT', msg => "Not running git command (...) because of signal INT");
```
Martchus, line 495: Maybe we should have an explicit `use OpenQA::Error::Cmd;` in this test file.

t/16-utils-runcmd.t
```diff
@@ -333,6 +333,66 @@ subtest 'signal guard aborts when git is disabled and do_cleanup is "no"' => sub
...
+            {status => 1, return_code => 0, stdout => 'add ok', stderr => '', signal => 0},
+            {status => 0, return_code => 0, stdout => '', stderr => '', signal => 'INT'},
+            {status => 1, return_code => 0, stdout => 'restore ok', stderr => '', signal => 0},
+        );
+        my @executed;
+        $git_mock->redefine(
+            run_cmd_with_log_return_error => sub ($cmd, %args) {
+                push @executed, $cmd;
+                my $res = shift @results;
+                return $res;
```
Martchus:
```suggestion
                return shift @results;
```
Or does this slightly change the meaning of the code?

perlpunk: done

t/14-grutasks-git.t
perlpunk: done (`OpenQA::Error`)

lib/OpenQA/Task/Needle/Delete.pm
perlpunk: Ok, I changed this back.

lib/OpenQA/Error/Cmd.pm
d3flex: I hope it was a funny debate. ))

PR #6888, opened 2026-01-05
Skip caching for KERNEL INITRD variables

d3flex:
Treat KERNEL and INITRD similar to UEFI_PFLASH_VARS. The variables might be assigned with absolute paths. There are not used as regular variables and when they do they are meant to be used in replace of UEFI_PFLASH_VARS.

The change, mainly, prevents the given asset to try be downloaded to the cache directory. Together with `ABSOLUTE_TEST_CONFIG_PATHS=1` they allow to be used in booting a custom kernle for testing purposes.

related_issue: https://progress.opensuse.org/issues/190056

### DIFF
--- a/lib/OpenQA/Worker/Engines/isotovideo.pm
+++ b/lib/OpenQA/Worker/Engines/isotovideo.pm
@@ -33,6 +33,7 @@ use Feature::Compat::Try;
 use constant CGROUP_SLICE => $ENV{OPENQA_CGROUP_SLICE};
 use constant CACHE_SERVICE_POLL_DELAY => $ENV{OPENQA_CACHE_SERVICE_POLL_DELAY} // 5;
 use constant CACHE_SERVICE_TEST_SYNC_ATTEMPTS => $ENV{OPENQA_CACHE_SERVICE_TEST_SYNC_ATTEMPTS} // 3;
+my %ASSETS_TO_SKIP = map { $_ => 1 } qw(UEFI_PFLASH_VARS KERNEL INITRD);
 
 my $isotovideo = '/usr/bin/isotovideo';
 my $workerpid;
@@ -64,7 +65,7 @@ sub detect_asset_keys ($vars) {
         # test (which we should treat as an hdd asset), or it may point
         # to an absolute filesystem location of e.g. a template file from
         # edk2 (which we shouldn't).
-        next if $key eq 'UEFI_PFLASH_VARS' && $value =~ m,^/,;
+        next if exists $ASSETS_TO_SKIP{$key} && $value =~ m,^/,;
         my $type = asset_type_from_setting($key, $value);
 
         # Exclude repo assets for now because the cache service does not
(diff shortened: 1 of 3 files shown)


kalikiana (approved):

lib/OpenQA/Worker/Engines/isotovideo.pm
```diff
@@ -56,6 +56,7 @@ sub _save_vars ($pooldir, $vars) {
 
 sub detect_asset_keys ($vars) {
     my %res;
+    my @skip_assets = qw(UEFI_PFLASH_VARS KERNEL INITRD);
```
Martchus: Move this outside the function and use a hash:

```suggestion
my %ASSETS_TO_SKIP = map { $_ => 1 } qw(UEFI_PFLASH_VARS KERNEL INITRD);
```

lib/OpenQA/Worker/Engines/isotovideo.pm
```diff
@@ -64,7 +65,7 @@ sub detect_asset_keys ($vars) {
         # test (which we should treat as an hdd asset), or it may point
         # to an absolute filesystem location of e.g. a template file from
         # edk2 (which we shouldn't).
-        next if $key eq 'UEFI_PFLASH_VARS' && $value =~ m,^/,;
+        next if (grep { $key eq $_ } @skip_assets) && $value =~ m,^/,;
```
Martchus:
```suggestion
        next if $ASSETS_TO_SKIP{$key} && $value =~ m,^/,;
```

Or at least use `any` instead of  `grep`.

Martchus (changes requested):
`s/There are not used/They are not used/`
`s/KERNEL INITRD variables/KERNEL and INITRD variables/`
`s/related_issue: /Related issue: /`

lib/OpenQA/Worker/Engines/isotovideo.pm
d3flex: I wonder if the comment also should go above the %ASSETS_TO_SKIP? wdyt?

Martchus (approved):

lib/OpenQA/Worker/Engines/isotovideo.pm
```diff
@@ -64,7 +65,7 @@ sub detect_asset_keys ($vars) {
         # test (which we should treat as an hdd asset), or it may point
         # to an absolute filesystem location of e.g. a template file from
         # edk2 (which we shouldn't).
-        next if $key eq 'UEFI_PFLASH_VARS' && $value =~ m,^/,;
+        next if exists $ASSETS_TO_SKIP{$key} && $value =~ m,^/,;
```
perlpunk, line 68: FTR, the value is set to 1, so the exists is unnecessary here:
```suggestion
        next if $ASSETS_TO_SKIP{$key} && $value =~ m,^/,;
```

Martchus: Yes - and it wasn't part of my suggestion :-)

PR #6891, opened 2026-01-08
feat(scheduled products): Allow adding note to result data

Martchus:
The note will be displayed on the web UI like the other result data. It can be used by tools like qem-bot to display additional information, e.g. "Not approving maintenance request X".

Related ticket: https://progress.opensuse.org/issues/193843

### DIFF
--- a/lib/OpenQA/WebAPI/Controller/API/V1/Iso.pm
+++ b/lib/OpenQA/WebAPI/Controller/API/V1/Iso.pm
@@ -56,6 +56,15 @@ sub show_scheduled_product {
     $self->render(json => $scheduled_product->to_hash(@args));
 }
 
+sub _get_iso_params_and_validate ($self) {
+    my $validation = $self->validation;
+    my @param_keys = (qw(distri version flavor arch build));
+    $validation->required($_) for @param_keys;
+    return [map { $validation->param($_) } @param_keys] unless $validation->has_error;
+    $self->reply->validation_error({format => 'json'});
+    return 0;
+}
+
 =over 4
 
 =item job_statistics()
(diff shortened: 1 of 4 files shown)


lib/OpenQA/WebAPI/Controller/API/V1/Iso.pm
```diff
@@ -89,13 +98,17 @@ could also be used to generate a more detailed report.
 =cut
 
 sub job_statistics ($self) {
+    return undef unless my $params = $self->_get_iso_params_and_validate;
+    my $scheduled_products = $self->app->schema->resultset('ScheduledProducts');
+    $self->render(json => $scheduled_products->job_statistics(@$params));
+}
+
+sub update_note ($self) {
```
okurz, line 119: Please add a documentation on top so that our API help page also explains the usage

lib/OpenQA/WebAPI/Controller/API/V1/Iso.pm
```diff
@@ -89,13 +98,30 @@ could also be used to generate a more detailed report.
...
+}
+
+=over 4
+
+=item update_note()
+
+Adds/updates a note on the latest scheduled product matching the specified
+DISTRI, VERSION, FLAVOR, ARCH and BUILD parameters. This note is shown on the
+web UI. The ID of the scheduled product where the note was added/updated is
+returned, e.g. `{updated_product_id: 42}`.
```
okurz, line 113: Can one also retrieve the note over API?

Martchus: Yes, the note will be added to the results of the scheduled product which can be queried via the API. The added test also does this to verify that the note is present.

PR #6892, opened 2026-01-08
Kernel bug report - Add plugin-based reporting templates

schlad:
https://progress.opensuse.org/issues/194474

WiP. This PR introduces a new IssueReporter framework that moves external issue reporting logic out of the branding-specific EP template into a structured WebAPI plugin with supporting helpers.

**The main goal** is to allow multiple templates which could help reporting good quality bugs - it should be possible to easily add more templates.
**Another goal is to address the kernel devs' feedback** which is described in the ticket. see: https://progress.opensuse.org/issues/194474

* OpenQA::WebAPI::Plugin::IssueReporter is meant as a single entry point for external issue reporting.
* external_reporting.html.ep is stripped down to bare minimum and is responsible only for the UI.
* 3 templates are added - GenericBug: which preserves the previous behaviour, ProgressIssue: which also preserves the original behaviour. KernelBug: new template and button.

This is deployed and can be tested here:
check for openSuse/TW: https://rmarliere-openqa.qe.prg2.suse.org/tests/1700
SLE15SP7: https://rmarliere-openqa.qe.prg2.suse.org/tests/1663

* if we agree to have something like this, I will prepare the documentation and the unit tests
* I think adding a specific 'kernel report bug' button is reasonable since a lot of QE engineers do report them and it would make sense to have a specific bug for this system component. 
* I consider working on the UI side to allow sort of a drop down to allow to pick a template.

### DIFF
--- a/lib/OpenQA/WebAPI/Plugin/IssueReporter/OpenSuseIssueReporter.pm
+++ b/lib/OpenQA/WebAPI/Plugin/IssueReporter/OpenSuseIssueReporter.pm
@@ -11,13 +11,11 @@ use OpenQA::WebAPI::Plugin::IssueReporter::OpenSuseProgressIssue;
 sub register ($self, $app, $config) {
     $app->helper(
         report_external_issue => sub ($c) {
-            my @actions;
-
-            push @actions, @{OpenQA::WebAPI::Plugin::IssueReporter::OpenSuseGenericBug::actions($c)};
-            push @actions, @{OpenQA::WebAPI::Plugin::IssueReporter::OpenSuseKernelBug::actions($c)};
-            push @actions, @{OpenQA::WebAPI::Plugin::IssueReporter::OpenSuseProgressIssue::actions($c)};
-
-            return \@actions;
+            return [
+                @{OpenQA::WebAPI::Plugin::IssueReporter::OpenSuseGenericBug::actions($c)},
+                @{OpenQA::WebAPI::Plugin::IssueReporter::OpenSuseKernelBug::actions($c)},
+                @{OpenQA::WebAPI::Plugin::IssueReporter::OpenSuseProgressIssue::actions($c)},
+            ];
         });
 }
 


schlad:
> Nice idea! I think we have the space to accomodate that separate "kernel bug" icon :) It seems the blocker flag isn't set anymore which IIRC was done automatically in before. More functionality might have become lost. I am not sure about the location and name of files. So far any "SUSE specifics" were neatly separated into the "branding" folder which means that only a instance actually configured to be "SUSE" would also load this SUSE code. With the new approach the "SUSE" stuff is generally imported. Maybe somebody has a good idea how to still support the more elaborate functionality while also preserving the instance specific configuration.

"the blocker flag isn't set anymore which IIRC was done automatically in before." - ah OK, I removed it assuming we should not have that by default. I will restore that.

templates/webapi/branding/openSUSE/external_reporting.html.ep
```diff
@@ -1,158 +1,5 @@
...
-    %= stepaction_for('Report product bug' => url_for($product_url_new)->query(%product_details), 'fa-bug', 'report product_bug');
-% }
-<% my %test_issue_params = (
-    'issue[subject]' => "test fails in $module",
-    'issue[description]' => $description,
-    'issue[category_id]' => 152,
-); %>
-% my $test_url_new = 'https://progress.opensuse.org/projects/openqatests/issues/new';
-%= stepaction_for('Report test issue' => url_for($test_url_new)->query(%test_issue_params), 'fa-bolt', 'report test_issue');
+% if (my $actions = eval { report_external_issue() }) {
```
Martchus: An `eval` without error handler can make this hard to debug. I suppose you just want to cover the case here that the helper doesn't exist (because the plugin hasn't been loaded). @kraih Do you know how to check whether a helper exists from a template? I briefly checked the documentation but couldn't find anything.

However, maybe this is also the place where we now want to load the new plugin. After all, when setting the branding from "plain" to "openSUSE" one would expect this to be effective. I guess the plugin could be loaded via `% app->plugin('IssueReporter');`. (Maybe you'll need to use `app->report_external_issue` then.)

lib/OpenQA/WebAPI/Plugin/IssueReporter/OpenSuseIssueReporter.pm
```diff
@@ -0,0 +1,21 @@
...
+sub register ($self, $app, $config) {
+    $app->helper(
+        report_external_issue => sub ($c) {
+            my @actions;
+
+            push @actions, @{OpenQA::WebAPI::Plugin::IssueReporter::GenericBug::actions($c) // []};
+            push @actions, @{OpenQA::WebAPI::Plugin::IssueReporter::KernelBug::actions($c) // []};
+            push @actions, @{OpenQA::WebAPI::Plugin::IssueReporter::ProgressIssue::actions($c) // []};
+
+            return \@actions;
```
Martchus: It seems like `actions` always returns an array ref so `// []` isn't needed.

```suggestion
            return [
                @{OpenQA::WebAPI::Plugin::IssueReporter::GenericBug::actions($c),
                @{OpenQA::WebAPI::Plugin::IssueReporter::KernelBug::actions($c),
                @{OpenQA::WebAPI::Plugin::IssueReporter::ProgressIssue::actions($c),
            ];
```

templates/webapi/branding/openSUSE/external_reporting.html.ep
schlad: I run into the debugging issue with missing handler for the eval, so I resorted to an try-catch and `app->log->error`. (I've checked it and it behaved well giving me proper error message:
```
an 12 19:37:22 rmarliere-openqa.qe.prg2.suse.org openqa-webui-daemon[127424]: [info] Creating process id file "/var/lib/openqa/webui/prefork-0.pid"
Jan 12 19:37:22 rmarliere-openqa.qe.prg2.suse.org openqa-webui-daemon[127462]: [info] Worker 127462 started
Jan 12 19:37:24 rmarliere-openqa.qe.prg2.suse.org openqa-webui-daemon[127459]: WARNING:  database "openqa" has a collation version mismatch
Jan 12 19:37:24 rmarliere-openqa.qe.prg2.suse.org openqa-webui-daemon[127459]: DETAIL:  The database was created using collation version 2.40, but the operating system provides version 2.42.
Jan 12 19:37:24 rmarliere-openqa.qe.prg2.suse.org openqa-webui-daemon[127459]: HINT:  Rebuild all objects in this database that use the default collation and run ALTER DATABASE openqa REFRESH COLLATION VERSION, or build PostgreSQL with the right library version.
Jan 12 19:37:25 rmarliere-openqa.qe.prg2.suse.org openqa-webui-daemon[127460]: WARNING:  database "openqa" has a collation version mismatch
Jan 12 19:37:25 rmarliere-openqa.qe.prg2.suse.org openqa-webui-daemon[127460]: DETAIL:  The database was created using collation version 2.40, but the operating system provides version 2.42.
Jan 12 19:37:25 rmarliere-openqa.qe.prg2.suse.org openqa-webui-daemon[127460]: HINT:  Rebuild all objects in this database that use the default collation and run ALTER DATABASE openqa REFRESH COLLATION VERSION, or build PostgreSQL with the right library version.
Jan 12 19:37:25 rmarliere-openqa.qe.prg2.suse.org openqa-webui-daemon[127460]: [info] branding: external_reporting.html.ep reached
Jan 12 19:37:25 rmarliere-openqa.qe.prg2.suse.org openqa-webui-daemon[127460]: [error] IssueReporter helper failed: Can't locate object method "report_external_issueeee" via package "OpenQA::WebAPI::Controller::Test" at template branding/openSUSE/external_reporting.html.ep line 4.
Jan 12 19:37:25 rmarliere-openqa.qe.prg2.suse.org openqa-webui-daemon[127460]: Context:
Jan 12 19:37:25 rmarliere-openqa.qe.prg2.suse.org openqa-webui-daemon[127460]:   1: % app->plugin('OpenQA::WebAPI::Plugin::IssueReporter::OpenSuseIssueReporter');
Jan 12 19:37:25 rmarliere-openqa.qe.prg2.suse.org openqa-webui-daemon[127460]:   2: % app->log->info('branding: external_reporting.html.ep reached');
Jan 12 19:37:25 rmarliere-openqa.qe.prg2.suse.org openqa-webui-daemon[127460]:   3:
Jan 12 19:37:25 rmarliere-openqa.qe.prg2.suse.org openqa-webui-daemon[127460]:   4: % if (my $actions = eval { $c->report_external_issueeee }) {
Jan 12 19:37:25 rmarliere-openqa.qe.prg2.suse.org openqa-webui-daemon[127460]:   5: %   for my $a (@$actions) {
Jan 12 19:37:25 rmarliere-openqa.qe.prg2.suse.org openqa-webui-daemon[127460]:   6:       %= stepaction_for($a->{label} => $a->{url}, $a->{icon}, "report $a->{id}");
Jan 12 19:37:25 rmarliere-openqa.qe.prg2.suse.org openqa-webui-daemon[127460]:   7: %   }
Jan 12 19:37:25 rmarliere-openqa.qe.prg2.suse.org openqa-webui-daemon[127460]:   8: % } else {
Jan 12 19:37:25 rmarliere-openqa.qe.prg2.suse.org openqa-webui-daemon[127460]:   9: %   app->log->error("IssueReporter helper failed: $@") if $@;
Jan 12 19:37:25 rmarliere-openqa.qe.prg2.suse.org openqa-webui-daemon[127460]: Traceback (most recent call first):
...
```

I also did apply this: "maybe this is also the place where we now want to load the new plugin". 
Current deployment in https://rmarliere-openqa.qe.prg2.suse.org does not load IssueReporter/OpenSuseIssueReporter in the openqa.ini

lib/OpenQA/WebAPI/Plugin/IssueReporter/OpenSuseIssueReporter.pm
schlad: I'll fix it as soon as we agree on general structure and approach (naming of the files etc.)

lib/OpenQA/WebAPI/Plugin/IssueReporter/Context.pm
```diff
@@ -0,0 +1,51 @@
+package OpenQA::WebAPI::Plugin::IssueReporter::Context;
```
okurz, line 4:
```suggestion
# Copyright SUSE LLC
# SPDX-License-Identifier: GPL-2.0-or-later

package OpenQA::WebAPI::Plugin::IssueReporter::Context;
```

same for other files

lib/OpenQA/WebAPI/Plugin/IssueReporter/OpenSuseIssueReporter.pm
```diff
@@ -0,0 +1,21 @@
...
+sub register ($self, $app, $config) {
+    $app->helper(
+        report_external_issue => sub ($c) {
+            my @actions;
+
+            push @actions, @{OpenQA::WebAPI::Plugin::IssueReporter::OpenSuseGenericBug::actions($c) // []};
+            push @actions, @{OpenQA::WebAPI::Plugin::IssueReporter::OpenSuseKernelBug::actions($c) // []};
+            push @actions, @{OpenQA::WebAPI::Plugin::IssueReporter::OpenSuseProgressIssue::actions($c) // []};
+
+            return \@actions;
```
okurz:
```suggestion
my @plugins = qw(
    OpenQA::WebAPI::Plugin::IssueReporter::OpenSuseGenericBug
    OpenQA::WebAPI::Plugin::IssueReporter::OpenSuseKernelBug
    OpenQA::WebAPI::Plugin::IssueReporter::OpenSuseProgressIssue
);
return [ map { @{ $_->actions($c) // [] } } @plugins ];
```

lib/OpenQA/WebAPI/Plugin/IssueReporter/OpenSuseKernelBug.pm
```diff
@@ -0,0 +1,69 @@
...
+1;
+
+__DATA__
+@@ kernel_bug.txt.ep
+
+== EDIT ==
+IMPORTANT: For kernel bugs please provide detailed kernel information (`uname -a`, rpm -qi kernel-default, ...)
+
+See for detailes and context:
+https://confluence.suse.com/spaces/qasle/pages/1836712232/Kernel+Bug+Reporting+Guidelines
```
okurz: that's a SUSE internal resource and can't be referenced externally. This should be replaced to a properly published document.

lib/OpenQA/WebAPI/Plugin/IssueReporter/OpenSuseKernelBug.pm
```diff
@@ -0,0 +1,69 @@
...
+
+1;
+
+__DATA__
+@@ kernel_bug.txt.ep
+
+== EDIT ==
+IMPORTANT: For kernel bugs please provide detailed kernel information (`uname -a`, rpm -qi kernel-default, ...)
+
+See for detailes and context:
```
okurz:
```suggestion
See for details and context:
```

lib/OpenQA/WebAPI/Plugin/IssueReporter/OpenSuseProgressIssue.pm
```diff
@@ -0,0 +1,81 @@
...
+            scenario => $scenario,
+            module => $ctx->{module},
+            step_url => $ctx->{step_url},
+            scenario_description => $scenario_description,
+            first_known_bad => $first_known_bad,
+            last_good => $last_good,
+            latest => "$latest",
+        });
+
+
```
okurz: here and also in many other places there are quite many redundant blank lines. Please trim those.

templates/webapi/branding/openSUSE/external_reporting.html.ep
```diff
@@ -1,158 +1,11 @@
...
+% app->plugin('OpenQA::WebAPI::Plugin::IssueReporter::OpenSuseIssueReporter');
+% try {
+%   my $actions = $c->report_external_issue;
+%   for my $a (@$actions) {
+      %= stepaction_for($a->{label} => $a->{url}, $a->{icon}, "report $a->{id}");
+%   }
+% }
+% catch ($err) {
+%   app->log->error("IssueReporter helper failed: $err");
+% }
```
Martchus, line 4: Maybe we don't need the try-catch as now the plugin is being loaded and hence expected to be there. If the code runs into an unexpected error it will of course produce a 500 response without the try-catch but that's actually how we usually handle these kinds of errors.

```suggestion
% app->plugin('OpenQA::WebAPI::Plugin::IssueReporter::OpenSuseIssueReporter');
% for my $a (@{$c->report_external_issue}) {
      %= stepaction_for($a->{label} => $a->{url}, $a->{icon}, "report $a->{id}");
% }
```

Note that with the try-catch the user won't see any error; only the buttons will be missing (or only be partially present or the HTML will be invalid). That might be more confusing than running into a 500 page which at least clearly indicates something went wrong.

schlad: valid point. Thx

lib/OpenQA/WebAPI/Plugin/IssueReporter/OpenSuseIssueReporter.pm
schlad: This won't work as-is. 
```
Too many arguments for subroutine 'OpenQA::WebAPI::Plugin::IssueReporter::OpenSuseGenericBug::actions' (got 2; expected 1) at /usr/share/openqa/script/../lib/OpenQA/WebAPI/Plugin/IssueReporter/OpenSuseIssueReporter.pm line 18.
Jan 13 11:55:20 rmarliere-openqa.qe.prg2.suse.org openqa-webui-daemon[132482]: Context:
Jan 13 11:55:20 rmarliere-openqa.qe.prg2.suse.org openqa-webui-daemon[132482]:   13:             my @plugins = qw(
Jan 13 11:55:20 rmarliere-openqa.qe.prg2.suse.org openqa-webui-daemon[132482]:   14:               OpenQA::WebAPI::Plugin::IssueReporter::OpenSuseGenericBug
Jan 13 11:55:20 rmarliere-openqa.qe.prg2.suse.org openqa-webui-daemon[132482]:   15:               OpenQA::WebAPI::Plugin::IssueReporter::OpenSuseKernelBug
Jan 13 11:55:20 rmarliere-openqa.qe.prg2.suse.org openqa-webui-daemon[132482]:   16:               OpenQA::WebAPI::Plugin::IssueReporter::OpenSuseProgressIssue
Jan 13 11:55:20 rmarliere-openqa.qe.prg2.suse.org openqa-webui-daemon[132482]:   17:             );
Jan 13 11:55:20 rmarliere-openqa.qe.prg2.suse.org openqa-webui-daemon[132482]:   18:             return [map { @{$_->actions($c) // []} } @plugins];
Jan 13 11:55:20 rmarliere-openqa.qe.prg2.suse.org openqa-webui-daemon[132482]:   19:         });
Jan 13 11:55:20 rmarliere-openqa.qe.prg2.suse.org openqa-webui-daemon[132482]:   20: }
Jan 13 11:55:20 rmarliere-openqa.qe.prg2.suse.org openqa-webui-daemon[132482]:   21:
Jan 13 11:55:20 rmarliere-openqa.qe.prg2.suse.org openqa-webui-daemon[132482]:   22: 1;
Jan 13 11:55:20 rmarliere-openqa.qe.prg2.suse.org openqa-webui-daemon[132482]: Traceback (most recent call first):
```

I can adapt sub actions but perhaps current approach is sufficient.

I will leave it as is, without your suggestion and will wait for your feedback on v3.

Meanwhile I will apply earlier comment from Marius.

perlpunk: That can be fixed by adding a class argument to the actions functions:
```
sub actions ($class, $c) {
```

perlpunk: Alternatively
```
return [ map { @{ $_->can('actions')->($c) // [] } } @plugins ];
```

schlad: > That can be fixed by adding a class argument to the actions functions:
> 
> ```
> sub actions ($class, $c) {
> ```

yeah I applied that approach and tested it - of course worked.

I can adapt this or keep it as pushed in v3 - up to you really - I will simply follow your guidance here as personally I'm out of my depth here.

Martchus: I don't really care but would actually still prefer the version I initially suggested. Using a map makes this just harder to read and feels like over-engineering.

schlad: afaik I applied that suggestion but correct me if I'm wrong. We had quite some points in the discussions, so I might have missed something.

I'll resolve this point now and will request the reviews since - I think - I addressed all points and all checks are green now.

okurz: @Martchus I guess you meant you would prefer https://github.com/os-autoinst/openQA/pull/6892/changes#r2675781108 , right? I am fine with either your version or the one from perlpunk (or mine) but please no more procedural code with multiple "push"

schlad: updated and pushed.

schlad: > @Martchus I guess you meant you would prefer https://github.com/os-autoinst/openQA/pull/6892/changes#r2675781108 , right? I am fine with either your version or the one from perlpunk (or mine) but please no more procedural code with multiple "push"

updated and pushed.

t/ui/30-issue-reporter-plugin.t
```diff
@@ -0,0 +1,73 @@
...
+    $distri = 'SUSE Linux Enterprise';
+    is get_bugzilla_product_name(Test::FakeJob->new(FLAVOR => 'Server-Updates', VERSION => '15-SP3'), 'sle', \$distri),
+      'Server 15 SP3', 'sle flavor suffix stripped and version "-" becomes space';
+    is $distri, 'PUBLIC SUSE Linux Enterprise', 'PUBLIC rewrite still applies';
+
+    $distri = 'SUSE Linux Enterprise';
+    is get_bugzilla_product_name(Test::FakeJob->new(FLAVOR => 'WeirdFlavor', VERSION => '15 SP2'), 'sle', \$distri),
+      'Server 15 SP2', 'unknown flavor falls back to Server';
+    is $distri, 'SUSE Linux Enterprise', 'no PUBLIC rewrite for SP2';
+
```
okurz, line 62: I would prefer if the variable `$distri` is not reset all the time as it's not expected to change and also properly asserted. By the way, why should a "get_…" function change passed variable contents anyway?

Martchus: `$distri` is clearly an out parameter - otherwise it would also be quite unconventional to pass it as ref.

I guess the idea here was to allow these tests to pass independently of each other (considering `is` does not not stop the test execution in Perl; so it isn't really the same as a usual `assert`).

Martchus (approved):

t/ui/30-issue-reporter-plugin.t
okurz: ok, IMHO it would be cleaner to refactor "get_bugzilla_product_name" to not return a value *and* have a side-effect?

schlad: Please note my comment:
```
# taken from the external_reporting.html.ep
# likely this could be improved
```
IMHO we could roll it out as-is knowing there are shortcomings and then we could open another progress ticket to figure better way to handle that.

As I wrote in that file:
```
# below constants are taken directly from the external_reporting.html.ep
# ideally this should be in a config file to make openQA product/brand-agnostic 
```

I'm not sure if `ideally` is the right word we all agree on, but personally I do not like `OpenSuseBugzillaUtils` and I'm fully committed to improve it under your guidance.

Does it make sense? As mentioned, I'm out of my depth here, so I count on your guidance.

Martchus: I'm fine with this as-is. We can do further improvements but having the logic not within templates is already a big improvement on its own. I think the way the return value is used and parameter is passed makes it also quite obvious what's going on here so refactoring this wouldn't gain us much.

schlad:
> A review from @perlpunk is still missing as it was explicitly requested. Otherwise 2 approvals (which this PR has) would be enough.

ack

perlpunk (approved):

PR #6894, opened 2026-01-09
Update PostgresQL minimum required version to 14

Martchus:
So we can use [jsonb Subscripting](https://www.postgresql.org/docs/current/datatype-json.html#JSONB-SUBSCRIPTING). Note that PostgreSQL 13 isn't supported anymore by upstream anyway.

Related ticket: https://progress.opensuse.org/issues/193843

### DIFF
--- a/docs/Installing.asciidoc
+++ b/docs/Installing.asciidoc
@@ -500,8 +500,8 @@ httpsonly = 0
 [[database]]
 === Database
 
-openQA uses PostgreSQL as database. By default, a database with name `openqa`
-and `geekotest` user as owner is used. An automatic setup of a freshly 
+openQA requires PostgreSQL 14 or newer. By default, a database with name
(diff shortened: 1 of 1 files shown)


docs/Installing.asciidoc
```diff
@@ -1420,7 +1420,7 @@ version:
 +
 [source,sh]
 ----
-oldver=10 newver=12
+oldver=10 newver=14
```
okurz: Not sure if many steps are a good idea. Better


```suggestion
oldver=17 newver=18
```

to relate to a more current example

Martchus:
But to see whether it breaks https://build.opensuse.org/package/show/devel:openQA:ci/base we'll have to merge this. So it doesn't make sense to request changes due to this point. Or do you want me to override the Dockerfile in this project temporarily?

d3flex (approved):

PR #6897, opened 2026-01-12
Retry delete_needles job on server restart

perlpunk:

If the delete_needles job is interrupted by a gru server restart, we can note down the already deleted needles and enqueue a retry.

Note that apparently 'expire' is set to one minute when the minion server is shutting down and you do a retry. That can lead to a disappearing job if the timing is unfortunate. So I set expire to one day explicitly.


perlpunk:
Fixed patch coverage

PR #6900, opened 2026-01-14
feat: Consider archive as well in df-based cleanup of job results

Martchus:
* Apply the deletion logic for non-archived jobs also to archived jobs considering the free disk space within the archive directory
* Introduce `archive_min_free_disk_space_percentage` to configure this
* See https://progress.opensuse.org/issues/193999

### DIFF
--- a/docs/UsersGuide.asciidoc
+++ b/docs/UsersGuide.asciidoc
@@ -1513,11 +1513,12 @@ the retention for important jobs it is still subject to cleanup.
 === Space-aware cleanup ===
 The cleanup of logs/results uses time-based retentions and hence is independent
 of actually available space on any assigned storage volumes. To ensure enough
-free space on file system storing results one can use the configuration settings
-`…_min_free_disk_space_percentage` to control automatic cleanup behavior. These
-percentages extend the cleanup of logs/results so that they are deleted until
-the specified percentage of file system space is free. This extended deletion
-happens independently of configured thresholds.
+free space on the file systems storing results one can use the configuration
+settings `…_min_free_disk_space_percentage` to control automatic cleanup
+behavior. These percentages extend the cleanup of logs/results so that they are
+deleted until the specified percentage of file system space is free. This
+extended deletion happens independently of configured time-based retention
+thresholds.
 
 The deletion happens in the following order:
 
@@ -1532,9 +1533,9 @@ It is also possible to abort cleanup tasks early in case there is still
 enough headroom on relevant file systems anyway. This can be enabled using the
 configuration settings `…_cleanup_max_free_percentage`.
 
-WARNING: The space-aware cleanup relies on `df` reporting a valid file system
-space usage. The algorithms also assume that screenshots and test results are
-stored on the same file system. The `…_min_free_disk_space_percentage` settings
+NOTE: The space-aware cleanup relies on `df` reporting a valid file system space
+usage. The algorithms also assume that screenshots and test results are stored
+on the same file system. The `…_min_free_disk_space_percentage` settings
 specifically are still experimental.
 
 [id="asset_cleanup"]
--- a/docs/UsersGuide.asciidoc
+++ b/docs/UsersGuide.asciidoc
@@ -1479,6 +1479,18 @@ Further remarks:
   section of <<GettingStarted.asciidoc#_configuration,the web UI configuration>>.
   Jobs outside of any group use the limits configured in the `[no_group_limits]`
   section of the web UI configuration.
+* The cleanup of logs/results uses time-based retentions and hence is
+  independant of actually available space on any assigned storage volumes. To
+  ensure enough free space on filesystem storing results one can use the
+  configuration settings `…_min_free_disk_space_percentage` to control automatic
+  cleanup behavior. These percentages extend the cleanup of logs/results so that
+  they are deleted until the specified percentage of filesystem space is free.
+  This extended deletion happens independently of configured thresholds. Old
+  jobs are deleted first. Videos are deleted first; further results are only
+  deleted if deleting videos is not sufficient. Un-important jobs are deleted
+  first; important jobs are only deleted if deleting un-important jobs is not
+  sufficient. The algorithm assumes that screenshots are stored on the same
+  partition test results are stored. This feature is still experimental.
 * The <<Installing.asciidoc#auditing,Auditing>> section explains the cleanup of
   the audit log.
 
--- a/docs/UsersGuide.asciidoc
+++ b/docs/UsersGuide.asciidoc
@@ -1510,7 +1510,7 @@ This means a job is "in the archive" if:
 NOTE: Archiving does *not* prevent cleanup. If an archived important job exceeds
 the retention for important jobs it is still subject to cleanup.
 
-=== Space-aware cleanup of job results ===
+=== Space-aware cleanup ===
 The cleanup of logs/results uses time-based retentions and hence is independent
 of actually available space on any assigned storage volumes. To ensure enough
 free space on filesystem storing results one can use the configuration settings
@@ -1528,9 +1528,14 @@ The deletion happens in the following order:
 
 Within each group the oldest jobs are deleted first.
 
-WARNING: This feature is still experimental. It relies on `df` reporting a valid
-file system space usage. The algorithm also assumes that screenshots and test
-results are stored on the same file system.
+It is also possible to abort cleanup tasks early in case there is still
+enough headroom on relevant filesystems anyway. This can be enabled using the
+configuration settings `…_cleanup_max_free_percentage`.
+
+WARNING: The space-aware cleanup relies on `df` reporting a valid file system
+space usage. The algorithms also assume that screenshots and test results are
+stored on the same file system. The `…_min_free_disk_space_percentage` settings
+specifically are still experimental.
 
 [id="asset_cleanup"]
 === Cleanup strategy for assets ===
--- a/docs/UsersGuide.asciidoc
+++ b/docs/UsersGuide.asciidoc
@@ -1479,18 +1479,6 @@ Further remarks:
   section of <<GettingStarted.asciidoc#_configuration,the web UI configuration>>.
   Jobs outside of any group use the limits configured in the `[no_group_limits]`
   section of the web UI configuration.
-* The cleanup of logs/results uses time-based retentions and hence is
-  independant of actually available space on any assigned storage volumes. To
-  ensure enough free space on filesystem storing results one can use the
-  configuration settings `…_min_free_disk_space_percentage` to control automatic
-  cleanup behavior. These percentages extend the cleanup of logs/results so that
-  they are deleted until the specified percentage of filesystem space is free.
-  This extended deletion happens independently of configured thresholds. Old
-  jobs are deleted first. Videos are deleted first; further results are only
-  deleted if deleting videos is not sufficient. Un-important jobs are deleted
-  first; important jobs are only deleted if deleting un-important jobs is not
-  sufficient. The algorithm assumes that screenshots are stored on the same
-  partition test results are stored. This feature is still experimental.
 * The <<Installing.asciidoc#auditing,Auditing>> section explains the cleanup of
   the audit log.
 
@@ -1521,6 +1509,28 @@ This means a job is "in the archive" if:
 
 NOTE: Archiving does *not* prevent cleanup. If an archived important job exceeds
 the retention for important jobs it is still subject to cleanup.
+
+=== Space-aware cleanup of job results ===
+The cleanup of logs/results uses time-based retentions and hence is independent
+of actually available space on any assigned storage volumes. To ensure enough
+free space on filesystem storing results one can use the configuration settings
+`…_min_free_disk_space_percentage` to control automatic cleanup behavior. These
+percentages extend the cleanup of logs/results so that they are deleted until
+the specified percentage of filesystem space is free. This extended deletion
+happens independently of configured thresholds.
+
+The deletion happens in the following order:
+
+1. Videos of unimportant jobs
+2. Results of unimportant jobs
+3. Videos of important jobs
+4. Results of important jobs
+
+Within each group the oldest jobs are deleted first.
+
+WARNING: This feature is still experimental. It relies on `df` reporting a valid
+file system space usage. The algorithm also assumes that screenshots and test
+results are stored on the same file system.
 
 [id="asset_cleanup"]
 === Cleanup strategy for assets ===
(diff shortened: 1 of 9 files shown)


etc/openqa/openqa.ini
```diff
@@ -287,6 +287,9 @@ concurrent = 0
 ## Extends the job result cleanup to ensure the partition results are stored on does not become too full
 ## (still experimental, relies on df)
 #results_min_free_disk_space_percentage = 0
+## Extends the job result cleanup to ensure the partition results are archived on does not become too full
```
okurz: this text is mostly the same as above so it's hard for the admin to realize which parts are actually different.

etc/openqa/openqa.ini
```diff
@@ -287,6 +287,9 @@ concurrent = 0
...
+## (in the same way as results_min_free_disk_space_percentage does for the results dir, same caveats apply)
```
okurz: … and the text in this line in parentheses is relevant so how about turning it around and start with "Similar as results_min_free_disk_space_percentage"

But please extend the documentation for archiving so that we have a better high level description of what feature is possible and how it behaves

okurz (changes requested):

etc/openqa/openqa.ini
Martchus: I now changed it to simply have one explanation for both variables.

Martchus:
> So is this PR ready to review?

Yes - and it is in my opinion also ready to be merged :-)

etc/openqa/openqa.ini
```diff
@@ -284,9 +284,10 @@ concurrent = 0
 ## Specify the number of screenshot ID ranges (with a size as configured by screenshot_cleanup_batch_size) to process in a single Minion
 ## job (reduce to avoid Minion jobs from running very long and possibly being interrupted, increase to reduce the number of Minion jobs)
 #screenshot_cleanup_batches_per_minion_job = 450
-## Extends the job result cleanup to ensure the partition results are stored on does not become too full
+## Extend the job result cleanup to ensure the partitions results are stored and archived do not become too full
```
perlpunk: Now I don't really understand the sentence. The "results of the partitions" (?) are stored and archived? and then "do not become too full" seems to be half of another sentence...

docs/UsersGuide.asciidoc
```diff
@@ -1479,6 +1479,20 @@ Further remarks:
...
+  `results_min_free_disk_space_percentage`
+  and `archive_min_free_disk_space_percentage` exist. These percentages extend
+  the cleanup of logs/results so that they are deleted until the specified
+  percentage of disk space is free on the partitions logs/results are regularly
+  stored and archived respectively. This extended deletion happens independently
+  of configured thresholds. Old jobs are deleted first. Videos are deleted
+  first; further results are only deleted if deleting videos is not sufficient.
+  Un-important jobs are deleted first; important jobs are only deleted if
+  deleting un-important jobs is not sufficient. The algorithm assumes that
+  screenshots are stored on the same partition test results are stored. This
```
okurz:
```suggestion
  screenshots are stored on the same partition test results are stored on. This
```

and please also explain the difference behind `results_min_free_disk_space_percentage`
  and `archive_min_free_disk_space_percentage`

okurz (changes requested):

etc/openqa/openqa.ini
Martchus: I now split the sentence. I hope it makes now more sense.

Martchus: Sorry, the change isn't there yet. I messed up with Git.

docs/UsersGuide.asciidoc
Martchus: I've now changed it to avoid mentioning the explicit variables to keep the explanation more high-level. When going to the settings file one will now find comments explaining the specific variables.

docs/UsersGuide.asciidoc
```diff
@@ -1479,6 +1479,17 @@ Further remarks:
   section of <<GettingStarted.asciidoc#_configuration,the web UI configuration>>.
   Jobs outside of any group use the limits configured in the `[no_group_limits]`
   section of the web UI configuration.
+* Since the cleanup of logs/results uses time-based retentions it is not
+  directly possible to ensure the partition results are stored on is not getting
+  too full. Hence the configuration settings `…_min_free_disk_space_percentage`
```
okurz: I would prefer if you can avoid the double negation and give this positive phrasing. Also "partition results" reads weirdly. Also it's not about a partition but can we say "filesystem"? So how about

```suggestion
Note that the cleanup of logs/results uses time-based retentions and hence is
independant of actually available space on any assigned storage volumes.
To ensure enough free space on filesystem storing results one can use the
configuration settings `…_min_free_disk_space_percentage` to control automatic
cleanup behavior.
```

docs/UsersGuide.asciidoc
```diff
@@ -1479,6 +1479,17 @@ Further remarks:
...
+  exist. These percentages extend the cleanup of logs/results so that they are
+  deleted until the specified percentage of disk space is free. This extended
```
okurz:
```suggestion
  deleted until the specified percentage of filesystem space is free. This extended
```

docs/UsersGuide.asciidoc
```diff
@@ -1479,6 +1479,17 @@ Further remarks:
...
+  deletion happens independently of configured thresholds. Old jobs are deleted
+  first. Videos are deleted first; further results are only deleted if deleting
+  videos is not sufficient. Un-important jobs are deleted first; important jobs
```
okurz: There can only be one "first". How about just stating "In this order…"

etc/openqa/openqa.ini
```diff
@@ -284,9 +284,11 @@ concurrent = 0
 ## Specify the number of screenshot ID ranges (with a size as configured by screenshot_cleanup_batch_size) to process in a single Minion
 ## job (reduce to avoid Minion jobs from running very long and possibly being interrupted, increase to reduce the number of Minion jobs)
 #screenshot_cleanup_batches_per_minion_job = 450
-## Extends the job result cleanup to ensure the partition results are stored on does not become too full
-## (still experimental, relies on df)
+## Extend the job result cleanup via experimental `…_min_free_disk_space_percentage` settings to ensure storage does not become too full (relies on df)
+## Specifies the threshold for the for partition where job results are regularly stored on
```
okurz: Wrap please

etc/openqa/openqa.ini
```diff
@@ -284,9 +284,11 @@ concurrent = 0
...
 #results_min_free_disk_space_percentage = 0
+## Specifies the threshold for the partition where job results are archived
```
okurz:
```suggestion
## Specifies the threshold for the filesystem where job results are archived
```

okurz (changes requested):

etc/openqa/openqa.ini
Martchus: That would be inconsistent with other lines in the same section. Should I wrap all comments then? At what length?

etc/openqa/openqa.ini
Martchus: This would be inconsistent with phrasing in other comments. I guess I'll change all of them then.

docs/UsersGuide.asciidoc
```diff
@@ -1510,6 +1510,33 @@ This means a job is "in the archive" if:
 NOTE: Archiving does *not* prevent cleanup. If an archived important job exceeds
 the retention for important jobs it is still subject to cleanup.
 
+=== Space-aware cleanup ===
+The cleanup of logs/results uses time-based retentions and hence is independant
+of actually available space on any assigned storage volumes. To ensure enough
+free space on filesystem storing results one can use the configuration settings
```
okurz:
```suggestion
free space on the filesystem storing results one can use the configuration settings
```

docs/UsersGuide.asciidoc
```diff
@@ -1510,6 +1510,33 @@ This means a job is "in the archive" if:
...
+`…_min_free_disk_space_percentage` to control automatic cleanup behavior. These
+percentages extend the cleanup of logs/results so that they are deleted until
+the specified percentage of filesystem space is free. This extended deletion
+happens independently of configured thresholds.
```
okurz:
```suggestion
happens independently of configured time-based retention thresholds.
```

as the "free_space_percentage" can also be seen as threshold so we should be explicit to distinguish

docs/UsersGuide.asciidoc
```diff
@@ -1510,6 +1510,33 @@ This means a job is "in the archive" if:
...
+3. Videos of important jobs
+4. Results of important jobs
+
+Within each group the oldest jobs are deleted first.
+
+It is also possible to abort cleanup tasks early in case there is still
+enough headroom on relevant filesystems anyway. This can be enabled using the
+configuration settings `…_cleanup_max_free_percentage`.
+
+WARNING: The space-aware cleanup relies on `df` reporting a valid file system
```
okurz: I don't think that df is that dangerous that we should give a WARNING. How about

```suggestion
NOTE: The space-aware cleanup relies on `df` reporting a valid file system
```

which we use much more often

```
$ git grep -c NOTE: docs/
docs/Contributing.asciidoc:3
docs/Installing.asciidoc:4
docs/Networking.asciidoc:14
docs/UsersGuide.asciidoc:11
docs/WritingTests.asciidoc:13
$ git grep -c WARNING: docs/
docs/Contributing.asciidoc:1
docs/GettingStarted.asciidoc:1
docs/UsersGuide.asciidoc:2
docs/WritingTests.asciidoc:1
```

etc/openqa/openqa.ini
```diff
@@ -276,17 +276,19 @@ concurrent = 0
 
 [misc_limits]
 #untracked_assets_storage_duration = 14
-## Performs the cleanup of results/assets only if the free disk space on the relevant partition is below the specified percentage (and aborts early otherwise)
+## Performs the cleanup of results/assets only if the free disk space on the relevant filesystem is below the specified percentage (and aborts early otherwise)
```
okurz:
```suggestion
## Performs the cleanup of results/assets only if the free space on the relevant filesystem is below the specified percentage (and aborts early otherwise)
```

etc/openqa/openqa.ini
```diff
@@ -276,17 +276,19 @@ concurrent = 0
...
 #asset_cleanup_max_free_percentage = 100
 ## Specify the screenshot ID range to query at once from the database (reduce to avoid big queries, increase to lower query overhead)
 #screenshot_cleanup_batch_size = 200000
 ## Specify the number of screenshot ID ranges (with a size as configured by screenshot_cleanup_batch_size) to process in a single Minion
 ## job (reduce to avoid Minion jobs from running very long and possibly being interrupted, increase to reduce the number of Minion jobs)
 #screenshot_cleanup_batches_per_minion_job = 450
-## Extends the job result cleanup to ensure the partition results are stored on does not become too full
-## (still experimental, relies on df)
+## Extend the job result cleanup via experimental `…_min_free_disk_space_percentage` settings to ensure storage does not become too full (relies on df)
+## Specifies the threshold for the for filesystem where job results are regularly stored on
```
okurz: Wrap please
```suggestion
## Specifies the threshold for the filesystem where job results are regularly stored on
```

etc/openqa/openqa.ini
```diff
@@ -276,17 +276,19 @@ concurrent = 0
...
 #results_min_free_disk_space_percentage = 0
+## Specifies the threshold for the filesystem where job results are archived
+#archive_min_free_disk_space_percentage = 0
```
okurz: From the comments I still don't understand how to pick each value. Also the meaning of the default value "0" should be explained

okurz (changes requested):

docs/UsersGuide.asciidoc
```diff
@@ -1510,6 +1510,33 @@ This means a job is "in the archive" if:
 NOTE: Archiving does *not* prevent cleanup. If an archived important job exceeds
 the retention for important jobs it is still subject to cleanup.
 
+=== Space-aware cleanup ===
+The cleanup of logs/results uses time-based retentions and hence is independant
```
perlpunk:
```suggestion
The cleanup of logs/results uses time-based retentions and hence is independent
```

docs/UsersGuide.asciidoc
```diff
@@ -1479,6 +1479,18 @@ Further remarks:
...
   section of the web UI configuration.
+* The cleanup of logs/results uses time-based retentions and hence is
+  independant of actually available space on any assigned storage volumes. To
+  ensure enough free space on filesystem storing results one can use the
+  configuration settings `…_min_free_disk_space_percentage` to control automatic
+  cleanup behavior. These percentages extend the cleanup of logs/results so that
+  they are deleted until the specified percentage of filesystem space is free.
+  This extended deletion happens independently of configured thresholds. Old
+  jobs are deleted first. Videos are deleted first; further results are only
+  deleted if deleting videos is not sufficient. Un-important jobs are deleted
```
perlpunk: "Un-important" looks a bit weird to me. Why not just "Unimportant"?

docs/UsersGuide.asciidoc
```diff
@@ -1522,6 +1510,28 @@ This means a job is "in the archive" if:
...
+1. Videos of unimportant jobs
+2. Results of unimportant jobs
+3. Videos of important jobs
+4. Results of important jobs
+
+Within each group the oldest jobs are deleted first.
+
+WARNING: This feature is still experimental. It relies on `df` reporting a valid
+file system space usage. The algorithm also assumes that screenshots and test
+results are stored on the same file system.
```
perlpunk: Sometimes "filesystem" is used, sometimes "file system".

docs/UsersGuide.asciidoc
perlpunk: This seems to have gone with the latest commit

docs/UsersGuide.asciidoc
Martchus: Considering https://en.wikipedia.org/wiki/File_system there's no clear preference. I go for "file system" then as that's the title of the article.

etc/openqa/openqa.ini
Martchus: I'll create an additional PR showing how a setting for moving results to the archive could look like. But this is not it; so you really just assign these percentages say how much disk space should be kept free on the different file systems. For for this PR the explanation should be good enough.

I also added an explanation about the special meaning of zero.

okurz (approved):

PR #6901, opened 2026-01-14
feat: optionally configure fake auth key+secret+expiration

okurz:
Related progress issue: https://progress.opensuse.org/issues/194513

### DIFF
--- a/lib/OpenQA/WebAPI/Auth/Fake.pm
+++ b/lib/OpenQA/WebAPI/Auth/Fake.pm
@@ -5,14 +5,17 @@ package OpenQA::WebAPI::Auth::Fake;
 use Mojo::Base -base, -signatures;
 use Time::Seconds;
 
+use constant KEY => $ENV{OPENQA_FAKE_AUTH_KEY} // '1234567890ABCDEF';
+use constant SECRET => $ENV{OPENQA_FAKE_AUTH_SECRET} // '1234567890ABCDEF';
+use constant EXPIRATION => $ENV{OPENQA_FAKE_AUTH_EXPIRATION_S} // ONE_DAY;
+
 sub auth_logout ($self) { return }
 
 sub auth_login ($self) {
     my $headers = $self->req->headers;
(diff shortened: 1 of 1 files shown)


lib/OpenQA/WebAPI/Auth/Fake.pm
```diff
@@ -5,14 +5,17 @@ package OpenQA::WebAPI::Auth::Fake;
 use Mojo::Base -base, -signatures;
 use Time::Seconds;
 
+use constant KEY => $ENV{OPENQA_FAKE_AUTH_KEY} // '1234567890ABCDEF';
+use constant SECRET => $ENV{OPENQA_FAKE_AUTH_SECRET} // '1234567890ABCDEF';
+use constant EXPIRATION => $ENV{OPENQA_FAKE_AUTH_EXPIRATION_S} // ONE_DAY;
```
d3flex, line 10: this cant be a constant. it would be better as non constant vars - this is a quick first impression

perlpunk: > this cant be a constant.

why not?

perlpunk: > this cant be a constant.

why not?

edit: not sure why this comment appears twice...

d3flex: We had a chat in the meeting but I can tell you here too.. IMO once is defined it cant be changed. And I can think one two cases that this is not gonna play nice

perlpunk: > And I can think one two cases that this is not gonna play nice

Then let us know about those use cases, please

perlpunk: > IMO once is defined it cant be changed

It couldn't be changed before either, it was simply hardcoded in the function below. Now you can at least change it on server startup.

PR #6902, opened 2026-01-15
feat: Add dry run to df-based cleanup of job results

Martchus:
Based on https://github.com/os-autoinst/openQA/pull/6900, check out only the last commit.

I think this will be useful to be able to enable this feature in production without having to worry that it deletes too much, see the 2nd paragraph of https://github.com/os-autoinst/openQA/pull/6900#issuecomment-3750541077.

### DIFF
--- a/etc/openqa/openqa.ini
+++ b/etc/openqa/openqa.ini
@@ -277,7 +277,7 @@ concurrent = 0
 [misc_limits]
 #untracked_assets_storage_duration = 14
 ## Performs the cleanup of results/assets only if the free disk space on the
-## relevant filesystem is below the specified percentage (and aborts early
+## relevant file system is below the specified percentage (and aborts early
 ## otherwise)
 #result_cleanup_max_free_percentage = 100
 #asset_cleanup_max_free_percentage = 100
@@ -292,10 +292,10 @@ concurrent = 0
 ## Extend the job result cleanup via experimental
 ## `…_min_free_disk_space_percentage` settings to ensure storage does not become
 ## too full (relies on df)
-## Specifies the threshold for the for filesystem where job results are
+## Specifies the threshold for the for file system where job results are
 ## regularly stored on
 #results_min_free_disk_space_percentage = 0
-## Specifies the threshold for the filesystem where job results are archived
+## Specifies the threshold for the file system where job results are archived
 #archive_min_free_disk_space_percentage = 0
 ## Duration to keep minion jobs in the database, in seconds
 #minion_job_max_age = 604800
--- a/etc/openqa/openqa.ini
+++ b/etc/openqa/openqa.ini
@@ -276,9 +276,8 @@ concurrent = 0
 
 [misc_limits]
 #untracked_assets_storage_duration = 14
-## Performs the cleanup of results/assets only if the free disk space on the
-## relevant file system is below the specified percentage (and aborts early
-## otherwise)
+## Performs the cleanup of results/assets only if the free space on the relevant
+## file system is below the specified percentage (and aborts early otherwise)
 #result_cleanup_max_free_percentage = 100
 #asset_cleanup_max_free_percentage = 100
 ## Specify the screenshot ID range to query at once from the database (reduce to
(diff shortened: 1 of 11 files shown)


okurz (commented):
+1

etc/openqa/openqa.ini
```diff
@@ -284,9 +284,10 @@ concurrent = 0
 ## Specify the number of screenshot ID ranges (with a size as configured by screenshot_cleanup_batch_size) to process in a single Minion
 ## job (reduce to avoid Minion jobs from running very long and possibly being interrupted, increase to reduce the number of Minion jobs)
 #screenshot_cleanup_batches_per_minion_job = 450
-## Extends the job result cleanup to ensure the partition results are stored on does not become too full
+## Extend the job result cleanup to ensure the partitions results are stored and archived do not become too full
```
d3flex: I think 
```
## Extends the job result cleanup to ensure the partitions' results are stored on does not become too full
``` is more grammatically correct. not sure

File: lib/OpenQA/Task/Job/Limit.pm [hunk 2]
```diff
@@ -207,23 +211,26 @@ sub _ensure_results_below_threshold ($job, @) {
...
     # load configured free percentage
-    my $min_free_percentage = $job->app->config->{misc_limits}->{results_min_free_disk_space_percentage};
+    my $limits = $job->app->config->{misc_limits};
+    my $min_free_percentage = $limits->{results_min_free_disk_space_percentage};
+    my $min_free_percentage_ar = $limits->{archive_min_free_disk_space_percentage};
     return $job->finish('No minimum free disk space percentage configured') unless defined $min_free_percentage;
     return $job->fail('Configured minimum free disk space is not a number between 0 and 100')
       unless _is_valid_percentage($min_free_percentage);
+    return $job->fail('Configured archive_min_free_disk_space_percentage is not a number between 0 and 100')
+      if defined $min_free_percentage_ar && !_is_valid_percentage($min_free_percentage_ar);
```
d3flex, line 233: Why not include the condition in the log above?

lib/OpenQA/Task/Job/Limit.pm
d3flex, line 1: seems like the `_ensure_results_below_threshold` has two paths for two limits. However the code is more or less the same. I think it would be more clear if you extract some logic to helper function(s) to eliminate duplication and use the limits as parameters. wdyt?

d3flex (changes requested):
I would like to ask to simply the flow in the `_ensure_results_below_threshold`. I just walk the individual commits, so ignore if it changes in the following ones.

etc/openqa/openqa.ini
Martchus: I have not simplified/split this anyway due to @perlpunk comment.

lib/OpenQA/Task/Job/Limit.pm
Martchus: I would like to postpone further refactoring like this after https://github.com/os-autoinst/openQA/pull/6902. I would generally refrain from over-engineering this, though.

Martchus: Oh, this actually is about https://github.com/os-autoinst/openQA/pull/6902. I am still not sure whether it makes sense trying to do improvements like this right now. I think what I already have is already a step in the right direction and doing more it probably outside the scope of the ticket I'm currently working on.

lib/OpenQA/Task/Job/Limit.pm
Martchus: I now just added support for this additional variable in the same way the existing variable was handled. I guess I can do this small improvement but if I would start doing small improvements like this everywhere I don't think we'd get very far with what the referenced ticket is actually about.

Martchus: I've pushed another commit to improve the error message.

lib/OpenQA/Task/Job/Limit.pm
d3flex: I am good if you file a new issue to track this and not forget it, so we can move on (considering that it is out of scope - which I am not sure - and considering you already look into this yourself, it would be easier for you to implement the changes). So if you do not mind, would you like to do this and resolve the thread?

etc/openqa/openqa.ini
```diff
@@ -298,6 +298,8 @@ concurrent = 0
 ## Specifies the threshold for the file system where job results are archived,
 ## zero keeps archived results as long as configured per retentions
 #archive_min_free_disk_space_percentage = 0
+## Allows running the extended job result cleanup via `…_min_free_disk_space_percentage` settings in dry-run mode
```
okurz: as you already wrapped your comments above now wrap this as well pretty plz

okurz (changes requested):

okurz (approved):

lib/OpenQA/Task/Job/Limit.pm
d3flex: As descussed I will follow up with this, creating a draft when I found time. I will resolve this and give me 5 min to review the latest changes

d3flex (approved):

PR #6904, opened 2026-01-19
feat: Show when "Next & Previous" jobs are limited

Martchus:
Related ticket: https://progress.opensuse.org/issues/174439

---

screenshot_20260119_150428

(I changed the phrasing now a little bit to also include the concrete limit.)

### DIFF
--- a/lib/OpenQA/WebAPI/Controller/Test.pm
+++ b/lib/OpenQA/WebAPI/Controller/Test.pm
@@ -627,9 +627,16 @@ sub job_next_previous_ajax ($self) {
     my ($children_by_job, $parents_by_job) = $self->_fetch_dependencies_by_jobs([map $_->id, @jobs]);
 
     my $comment_data = $self->schema->resultset('Comments')->comment_data_for_jobs(\@jobs, {bugdetails => 1});
-    my @data;
+    my (@data, @info);
+    my %source_count;    # keep track of how many jobs of which source (latest, next, previous) we got for limiting
+    if (@jobs >= 2 && $jobs[0]->id == $jobs[1]->id) {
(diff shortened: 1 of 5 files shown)


lib/OpenQA/WebAPI/Controller/Test.pm
```diff
@@ -653,14 +658,23 @@ sub job_next_previous_ajax ($self) {
...
                 finished => $job->t_finished ? $job->t_finished->datetime() . 'Z' : undef,
                 duration => $job->t_started
                   && $job->t_finished ? $self->format_time_duration($job->t_finished - $job->t_started) : 0,
                 comment_data => $rendered_data,
             });
     }
-    $self->render(json => {data => \@data});
+    $data[0]->{islatest} = 1 if @data;
+    if (($source_count{p} // 0) > $p_limit) {  # the query requests `$p_limit + 1` so we know when the limit was reached
+        push @info, "The number of \"Previous\" jobs exceeds the display limit of $p_limit.";
```
okurz, line 669: I guess the obvious question a user might ask then is how to "un-limit"

Martchus: There are query parameters but also limits on the server-side one cannot exceed.

However, I think one hopefully doesn't need that. This message is mainly to avoid the confusion https://progress.opensuse.org/issues/174439 was about - not to support some workflow where a user would need to go though that many jobs.

okurz: Yes, that's ok. Maybe we can rephrase to make that clear and don't even invite such question ;)

okurz (approved):

PR #6912, opened 2026-01-21
GenericBug: Add [QE] to the subject

schlad:


lib/OpenQA/WebAPI/Plugin/IssueReporter/OpenSuseGenericBug.pm
```diff
@@ -36,7 +36,7 @@ sub actions ($c) {
 
     my $url = Mojo::URL->new($bugzilla_url)->query(
         {
-            short_desc => "[Build $ctx->{build}] openQA test fails in $ctx->{module}",
+            short_desc => "[QE][Build $ctx->{build}] openQA test fails in $ctx->{module}",
```
okurz, line 39: Not everyone clicking the report bug button on openQA is related to "QE", depending on definition. Could you please elaborate *in your commit message* why this should make sense?

okurz (changes requested):
https://github.com/os-autoinst/openQA/pull/6912/changes#r2711256436

lib/OpenQA/WebAPI/Plugin/IssueReporter/OpenSuseGenericBug.pm
schlad: Git commit amended

okurz: Ok, reading it again shows the problem: Not everyone clicking the report bug button on openQA is related to "QE". This does not make sense at this point also what I understood from the original request: To separate what was created by SUSE LSG QE employees vs. others. By just making any contributor of openSUSE on o3 create bugs with "QE" in the subject that will defeat the purpose, right?

schlad: QE tag here is meant not as something which indicates the team assignment but a tag to understand from where the bugs are coming from, so we would help the kernel development team, who is asking for this, with the triaging of the bugs, and effectively with solving the bugs.

I'm personally aware there are other flags one can filter on however if that could help with the product quality I think it is worth trying.

Here is the internal reference point: https://confluence.suse.com/spaces/labs/pages/1943142773/Reliable+process+to+handle+QA+bugs+reported+for+released+products+kernels
Where you actively voted, so I assume you are aware of this.

Also: the button opens the bugzilla with pre-defined strings and what not.
If one does not like the [QE], one can remove it.
o3 and osd is more actively used by suse qe folks than other folks, so I think proposed way is simply more efficient and since this comes with a promise of better kernel bug triage and thus quicker turnaround ... well I think we could give it a try.

wdyt?

okurz: Yes, I am aware about https://confluence.suse.com/spaces/labs/pages/1943142773/Reliable+process+to+handle+QA+bugs+reported+for+released+products+kernels which explicitly mentions that its important to distinguish the group of people reporting an issue, not from where it was created . There is already "found by" set to "openQA" so if you say they mean the same then this is redundant and I see it as not necessary. If you insist then I will accept and we will merge this changing all openSUSE+SUSE bug reporting templates regardless who uses that. I would leave it to you to clarify if it is ok to change the template for all openQA generated bug reports for all openQA instances running the "openSUSE/SUSE" branding, e.g. primarily o3+osd

schlad: Honestly, yes, I insist.
And it isn't like I did the easiest - added QE to the ep file and moved on.
Knowing how problematic that could have been, I've worked on plugin-based templates to respect the needs of various groups and to allow to customize specific template only if that's what required.

So please go ahead with merging.

Rest assured I am not deaf to your points and I added this: https://progress.opensuse.org/issues/194780
I imagine we can come up with the idea to parametrize the subject further, perhaps do sub-branding: Suse and play with the generic template specific for suse only. I will defiantly consider how to make all that more suitable and fitting all use-cases.
That also includes working with kernel development team to make sure they reduce the redundancy in the process, so we use certain filters more effectively. We are really working on it.

okurz (approved):

PR #6915, opened 2026-01-21
feat: add API endpoint for creating API keys

okurz:
Implement POST /api/v1/users/me/api_keys to allow creating API keys via API.

Related progress issue: https://progress.opensuse.org/issues/194513

### DIFF
--- a/lib/OpenQA/WebAPI/Controller/API/V1/User.pm
+++ b/lib/OpenQA/WebAPI/Controller/API/V1/User.pm
@@ -59,8 +59,10 @@ sub create_api_key ($self) {
         json => {error => 'Date must be in format ' . DateTime::Format::Pg->format_datetime(DateTime->now())},
         status => 400
     ) if $validation->has_error;
-    $expiration = DateTime::Format::Pg->parse_datetime($validation->param('expiration'))
-      if $validation->is_valid('expiration');
+    $expiration
+      = $validation->is_valid('expiration')
+      ? DateTime::Format::Pg->parse_datetime($validation->param('expiration'))
+      : DateTime->now->add(years => 1);
     my $apikey = $user->api_keys->create({t_expiration => $expiration});
     $self->render(json => {id => $apikey->id, key => $apikey->key, t_expiration => $apikey->t_expiration});
 }
@@ -83,7 +85,8 @@ sub list_api_keys ($self) {
             t_expiration => $_->t_expiration,
             t_created => $_->t_created,
             t_updated => $_->t_updated,
-        } } $user->api_keys->all;
+        }
+    } $user->api_keys->all;
     $self->render(json => {keys => \@keys});
 }
 
--- a/t/api/15-users.t
+++ b/t/api/15-users.t
@@ -38,7 +38,8 @@ is $app->schema->resultset('Users')->find(99902)->feature_version, 42, 'feature 
 
 my $res = $t->post_ok('/api/v1/users/me/api_keys')->status_is(200, 'create api key')->tx->res->json;
 ok $res->{key}, 'key returned';
-is $res->{t_expiration}, undef, 'default expiration is undef (never)';
+my $expected_year = time2str('%Y', time + ONE_YEAR, 'UTC');
+like $res->{t_expiration}, qr/^$expected_year-/, 'default expiration is set to 1 year from now';
 my $key1 = $res->{key};
 
 my $new_key = $app->schema->resultset('ApiKeys')->find({key => $res->{key}});
@@ -46,11 +47,13 @@ ok $new_key, 'key found in DB';
 is $new_key->user_id, 99902, 'key belongs to correct user';
 
 subtest 'create_api_key with expiration' => sub {
-    my $expiration = time2str('%Y-%m-%d %H:%M:%S', time + ONE_YEAR, 'UTC');
+    my $expiration_time = time + ONE_YEAR;
+    my $expiration = time2str('%Y-%m-%d %H:%M:%S', $expiration_time, 'UTC');
     $res = $t->post_ok('/api/v1/users/me/api_keys' => form => {expiration => $expiration})
       ->status_is(200, 'create api key with expiration')->tx->res->json;
     ok $res->{key}, 'key returned';
-    like $res->{t_expiration}, qr/^[0-9]+/, 'expiration looks like date';
+    my $expected_year = time2str('%Y', $expiration_time, 'UTC');
+    like $res->{t_expiration}, qr/^$expected_year-/, 'expiration matches expected year';
 };
 
 subtest 'create_api_key with invalid expiration' => sub {


lib/OpenQA/WebAPI/Controller/API/V1/User.pm
```diff
@@ -26,17 +28,46 @@ Implements User API.
 
 Deletes an existing user.
 
+=item create_api_key()
+
+Create a new API key.
+
```
Martchus: Shouldn't this be placed before the function?

lib/OpenQA/WebAPI/Controller/API/V1/User.pm
```diff
@@ -26,17 +28,46 @@ Implements User API.
...
+    return $self->render(
+        json => {error => 'Date must be in format ' . DateTime::Format::Pg->format_datetime(DateTime->now())},
+        status => 400
+    ) if $validation->has_error;
+    if ($validation->is_valid('expiration')) {
+        try { $expiration = DateTime::Format::Pg->parse_datetime($validation->param('expiration')) }
+        catch ($e) { return $self->render(json => {error => "$e"}, status => 400) }
+    }
+    my $apikey;
+    try { $apikey = $self->schema->resultset('ApiKeys')->create({user_id => $user->id, t_expiration => $expiration}) }
```
Martchus: *Maybe* the following works, too:
```suggestion
    try { $apikey = $user->api_keys->create({t_expiration => $expiration}) }
```

lib/OpenQA/WebAPI/Controller/API/V1/User.pm
```diff
@@ -26,17 +28,46 @@ Implements User API.
...
+    catch ($e) { return $self->render(json => {error => "$e"}, status => 500) }
```
Martchus: Returning internal server errors makes sense in development mode but we shouldn't do it in production mode.

```suggestion
    catch ($e) { return $self->render(json => {error => 'Internal server error'}, status => 500) }
```

lib/OpenQA/WebAPI/Controller/API/V1/User.pm
```diff
@@ -26,17 +28,46 @@ Implements User API.
...
+    $self->render(
+        json => {
+            key => $apikey->key,
+            secret => $apikey->secret,
+            t_expiration => $apikey->t_expiration ? '' . $apikey->t_expiration : undef,
+            t_created => '' . $apikey->t_created,
+            t_updated => '' . $apikey->t_updated,
```
Martchus: Is the concatenation with an empty string literal really needed?

lib/OpenQA/WebAPI/Controller/API/V1/User.pm
okurz: Yes, moved

lib/OpenQA/WebAPI/Controller/API/V1/User.pm
okurz: seems to work, thx

lib/OpenQA/WebAPI/Controller/API/V1/User.pm
okurz: ok, done

lib/OpenQA/WebAPI/Controller/API/V1/User.pm
okurz: shouldn't be needed, removed.

perlpunk:
https://app.circleci.com/pipelines/github/os-autoinst/openQA/19023/workflows/9a00ebca-e9b3-45f6-be0d-000701338eca/jobs/188230
```

#   Failed test 'POD test for lib/OpenQA/WebAPI/Controller/API/V1/User.pm'
#   at /usr/lib/perl5/vendor_perl/5.26.1/Test/Pod.pm line 187.
# lib/OpenQA/WebAPI/Controller/API/V1/User.pm (25): =over without closing =back
# Looks like you failed 1 test of 238.
[14:06:11] t/02-pod.t ................................................ 
Failed 1/238 subtests 
```

lib/OpenQA/WebAPI/Controller/API/V1/User.pm
```diff
@@ -26,17 +28,50 @@ Implements User API.
...
+        json => {error => 'Date must be in format ' . DateTime::Format::Pg->format_datetime(DateTime->now())},
+        status => 400
+    ) if $validation->has_error;
+    if ($validation->is_valid('expiration')) {
+        try { $expiration = DateTime::Format::Pg->parse_datetime($validation->param('expiration')) }
+        catch ($e) { return $self->render(json => {error => "$e"}, status => 400) }
+    }
+    my $apikey;
+    try { $apikey = $user->api_keys->create({t_expiration => $expiration}) }
+    catch ($e) { return $self->render(json => {error => 'Internal server error'}, status => 500) }
```
d3flex: Should we add a log entry as well to be able to debug it or something?

okurz: yeah, good point. Basically this was over-engineering. I just removed the exception handling here as the create should be ensured to not fail if the expiration is already validated. Also this would introduce uncovered code so I significantly simplified this code.

lib/OpenQA/WebAPI/Controller/API/V1/User.pm
```diff
@@ -30,13 +32,78 @@ Deletes an existing user.
...
+    my $user = $self->current_user;
+    my @keys = map {
+        {
+            key => $_->key,
+            t_expiration => $_->t_expiration,
+            t_created => $_->t_created,
+            t_updated => $_->t_updated,
+        }
+    } $user->api_keys->all;
+    $self->render(json => \@keys);
```
Martchus: It is generally a good idea to return objects as it makes extending the response in the future easier:

```suggestion
    $self->render(json => {keys => \@keys});
```

okurz: done

t/api/15-users.t
```diff
@@ -38,7 +38,7 @@ is $app->schema->resultset('Users')->find(99902)->feature_version, 42, 'feature
 
 my $res = $t->post_ok('/api/v1/users/me/api_keys')->status_is(200, 'create api key')->tx->res->json;
 ok $res->{key}, 'key returned';
-is $res->{t_expiration}, undef, 'default expiration is undef (never)';
+like $res->{t_expiration}, qr/^[0-9]+/, 'default expiration is set';
```
d3flex: the regex is too generic. isnt it better to check that actually is <1y? would you mind to make the check explicit.

okurz: Done

perlpunk (approved):

d3flex (approved):
nice done.

PR #6917, opened 2026-01-21
feat(UI): add delete button for job groups and parent groups

okurz:
Screenshot_20260130_101340_openqa_delete_job_group

### DIFF
--- a/assets/javascripts/admin_groups.js
+++ b/assets/javascripts/admin_groups.js
@@ -410,3 +410,40 @@ function handleQuery(query) {
     .then(success)
     .catch(error);
 }
+function deleteGroup(elem, isParent) {
+  const li = $(elem).closest('li');
+  const idAttr = li.attr('id');
+  const id = parseInt(idAttr.replace(isParent ? 'parent_group_' : 'job_group_', ''));
+  const name = li
+    .find(isParent ? '.parent-group-name' : 'span > a')
+    .text()
+    .trim();
+  const type = isParent ? 'parent group' : 'job group';
+
+  if (!confirm(`Are you sure you want to delete the ${type} "${name}"?`)) {
+    return false;
+  }
+
+  const url = `/api/v1/${isParent ? 'parent_groups' : 'job_groups'}/${id}`;
+
+  fetchWithCSRF(url, {method: 'DELETE'})
+    .then(response => {
+      if (!response.ok) {
+        return response.json().then(json => {
+          throw json.error || response.statusText;
+        });
+      }
+      return response.json();
+    })
+    .then(() => {
+      li.fadeOut('slow', function () {
+        $(this).remove();
+      });
+    })
+    .catch(error => {
+      console.error(error);
+      alert(`Error deleting ${type}: ${error}`);
+    });
+
+  return false;
+}
--- a/lib/OpenQA/WebAPI/Controller/API/V1/JobGroup.pm
+++ b/lib/OpenQA/WebAPI/Controller/API/V1/JobGroup.pm
@@ -349,11 +349,11 @@ If not empty (there are existing jobs), it will return an error.
 sub delete ($self) {
     return unless my $group = $self->find_group;
 
-    if ($group->can('jobs') && scalar($group->jobs) != 0) {
+    if ($group->can('jobs') && $group->jobs->count != 0) {
         return $self->render(json => {error => 'Job group ' . $group->id . ' is not empty'}, status => 409);
     }
 
-    if ($group->can('children') && scalar($group->children) != 0) {
+    if ($group->can('children') && $group->children->count != 0) {
         return $self->render(json => {error => 'Parent job group ' . $group->id . ' is not empty'}, status => 409);
     }
 
--- a/t/ui/23-job_group_delete.t
+++ b/t/ui/23-job_group_delete.t
@@ -1,0 +1,72 @@
+# Copyright SUSE LLC
+# SPDX-License-Identifier: GPL-2.0-or-later
+
+use Test::Most;
+
+use FindBin;
+use lib "$FindBin::Bin/../lib", "$FindBin::Bin/../../external/os-autoinst-common/lib";
+use Test::Mojo;
+use Test::Warnings ':report_warnings';
+use OpenQA::Test::TimeLimit '8';
+use OpenQA::Test::Case;
+
+use OpenQA::Test::Client;
+
+OpenQA::Test::Case->new->init_data(fixtures_glob => '01-jobs.pl 03-users.pl');
+
+my $t = Test::Mojo->new('OpenQA::WebAPI');
+my $schema = $t->app->schema;
+my $job_groups = $schema->resultset('JobGroups');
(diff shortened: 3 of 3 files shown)


t/ui/23-job_group_delete.t
```diff
@@ -0,0 +1,33 @@
...
+
+subtest 'delete buttons not visible for not-logged-in/non-admin' => sub {
+    my $t = Test::Mojo->new('OpenQA::WebAPI');
+    $t->get_ok('/admin/groups')->status_is(200);
+    $t->element_exists_not('a[title^="Delete job group"]', 'Delete job group button not visible');
+    $t->element_exists_not('a[title^="Delete parent group"]', 'Delete parent group button not visible');
+};
+
+subtest 'delete buttons visible for admin' => sub {
+    my $t = Test::Mojo->new('OpenQA::WebAPI');
```
Martchus: The test would be more lightweight if we would avoid instantiating another app.

Maybe it would also be good to avoid a separate test just for that entirely. (Of course that depends on how "full" other tests already are.)

t/ui/23-job_group_delete.t
```diff
@@ -0,0 +1,33 @@
...
+    $t->post_ok('/login', form => {user => 'Demo'})->status_is(302);
+    my $parent = $t->app->schema->resultset('JobGroupParents')->create({name => 'Test Parent Group'});
+    $t->get_ok('/admin/groups')->status_is(200);
+    $t->element_exists('a[title^="Delete job group"]', 'Delete job group button visible');
+    $t->element_exists('a[title^="Delete parent group"]', 'Delete parent group button visible');
+    $t->element_exists('a[title="Delete parent group Test Parent Group"]',
+        'Delete button for specific parent group visible');
```
Martchus, line 35: Considering this can be quite destructive if it goes wrong it would make sense to test whether the deletion actually works and whether only the selected group is deleted.

templates/webapi/admin/group/parent_group_row.html.ep
```diff
@@ -23,6 +23,7 @@
         % }
         % if (is_admin) {
             <a href="#" onclick="return showAddJobGroup(this);" class="fa fa-2 fa-plus" title="Add new job group in <%= $group->name %>"></a>
+            <a href="#" onclick="return deleteGroup(this, true);" class="fa fa-trash" title="Delete parent group <%= $group->name %>"></a>
```
Martchus, line 26: If the order of the buttons was reversed the trash icons would line up more nicely.

assets/javascripts/admin_groups.js
```diff
@@ -410,3 +410,40 @@ function handleQuery(query) {
...
+function deleteGroup(element, isParent) {
+  const li = $(element).closest('li');
+  const idAttr = li.attr('id');
+  const id = parseInt(idAttr.replace(isParent ? 'parent_group_' : 'job_group_', ''));
+  const name = li
+    .find(isParent ? '.parent-group-name' : 'span > a')
+    .text()
+    .trim();
+  const type = isParent ? 'parent group' : 'job group';
+
```
Martchus, line 422:
```suggestion
```

assets/javascripts/admin_groups.js
```diff
@@ -410,3 +410,40 @@ function handleQuery(query) {
...
+  if (!confirm(`Are you sure you want to delete the ${type} "${name}"?`)) {
+    return false;
+  }
+
+  const url = `/api/v1/${isParent ? 'parent_groups' : 'job_groups'}/${id}`;
+
```
Martchus, line 428:
```suggestion
```

lib/OpenQA/WebAPI/Controller/API/V1/JobGroup.pm
```diff
@@ -368,6 +368,10 @@ sub delete ($self) {
         return $self->render(json => {error => 'Job group ' . $group->id . ' is not empty'}, status => 400);
     }
 
+    if ($group->can('children') && scalar($group->children) != 0) {
+        return $self->render(json => {error => 'Parent job group ' . $group->id . ' is not empty'}, status => 400);
```
d3flex: look at https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Status/409. It should be more appropiate and accurate than a general `400`

okurz: done

assets/javascripts/admin_groups.js
okurz: done

assets/javascripts/admin_groups.js
okurz: done

templates/webapi/admin/group/parent_group_row.html.ep
okurz: done, see updated screenshot

t/ui/23-job_group_delete.t
okurz: ok, "destructive test" added :)

t/ui/23-job_group_delete.t
okurz: well, there is no generic test file for "job groups". I have looked into other test files in t/ui/ but haven't found good candidates.

Martchus: Tests in that area are in `13-admin.t`. However, I haven't asked to add this test elsewhere. It is just that the 2nd instantiation of the app (the second occurrence of `my $t = Test::Mojo->new('OpenQA::WebAPI');`) seems a bit wasteful/unnecessary.

t/ui/23-job_group_delete.t
```diff
@@ -0,0 +1,53 @@
...
+    my $t = Test::Mojo->new('OpenQA::WebAPI');
+    $t->get_ok('/admin/groups')->status_is(200);
+    $t->element_exists_not('a[title^="Delete job group"]', 'Delete job group button not visible');
+    $t->element_exists_not('a[title^="Delete parent group"]', 'Delete parent group button not visible');
+};
+
+subtest 'delete buttons visible for admin' => sub {
+    my $t = Test::Mojo->new('OpenQA::WebAPI');
+    $t->post_ok('/login', form => {user => 'Demo'})->status_is(302);
+    my $parent = $t->app->schema->resultset('JobGroupParents')->create({name => 'Test Parent Group'});
```
Martchus: I would store `$t->app->schema->resultset('JobGroupParents')` and other resultsets in a variable once in the beginning of the test instead of having these repeated accesses.

t/ui/23-job_group_delete.t
okurz: oh, that. I thought we need it to change the user. Will look into this

t/ui/23-job_group_delete.t
okurz: done

t/ui/23-job_group_delete.t
okurz: done

PR #6919, opened 2026-01-21
Limit url parameters to maximum possible value

perlpunk:
With the placeholder `<id:bigint>` we can limit the id to 19 digits, but that will still allow numbers that are too big. For that we can use conditions:
https://metacpan.org/dist/Mojolicious/view/lib/Mojolicious/Guides/Routing.pod#Adding-conditions

The disadvantage is, we have to add them to every route definition. That will be a bit nicer with OpenAPI.

Issue: https://progress.opensuse.org/issues/193897

TODO: Add the `->requires(bigint_range => '...')` to the rest of the routes.

### DIFF
--- a/lib/OpenQA/WebAPI.pm
+++ b/lib/OpenQA/WebAPI.pm
@@ -203,11 +203,8 @@ sub startup ($self) {
     # but this route is actually matched (in case apache is not catching this earlier)
     # due to the split md5_dirname having a /
     $r->get('/image/:md5_1/:md5_2/.thumbs/#md5_basename')->to('file#thumb_image');
-
     $r->get('/group_overview/<groupid:bigint>' => [format => ['json', 'html']])->requires(bigint_range => 'groupid')
       ->name('group_overview')->to('main#job_group_overview', format => undef);
-
-
     $r->get('/parent_group_overview/<groupid:bigint>' => [format => ['json', 'html']])->name('parent_group_overview')
       ->to('main#parent_group_overview', format => undef)->requires(bigint_range => 'groupid');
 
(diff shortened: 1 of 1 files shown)


lib/OpenQA/WebAPI.pm
```diff
@@ -16,6 +16,8 @@ use Mojo::IOLoop;
 use Mojo::Util 'trim';
 use Feature::Compat::Try;
 
+use constant MAX_BIGINT => 9223372036854775807;
```
d3flex: Should this go to the Constants.pm? or can it?

okurz: Or why should we even need to define it manually? I am sure such constant already exists. If not then please specify in hex, not dec

okurz (changes requested):
Would be nice to not return 404 as a job "max_int+1" can never be found as it can't be referenced like that at all. How about
422 (Unprocessable Entity)

lib/OpenQA/WebAPI.pm
perlpunk: Changed it to `2 ** (8 * $Config{ivsize} - 1) - 1`.
I also thought about putting it into the Constants module. Should I do that?

perlpunk: Note that in the previous version I used the number from https://www.postgresql.org/docs/8.1/datatype.html#DATATYPE-NUMERIC which is different from `2**64-1`.
Gemini suggested `2 ** (8 * $Config{ivsize} - 1) - 1` which is `2**63-1`.
2**64-1 would be 18446744073709600000

Martchus: I'm wondering whether  `ivsize` is appropriate here. We're running into a limit on database-level here, right? So it shouldn't matter what some Perl-internal config says. Or would this limit prevent us from processing such big numbers in Perl at all so we also need to consider the Perl-internal limit? (I would expect that even a 32-bit build of Perl can handle 64-bit numbers by *some* means.)

I guess `2**63-1` makes more sense than `2**64-1` because we probably care about signed integers - even though we only use positive IDs. That let's me think: we should probably also forbid negative values.

perlpunk: Negative values are already excluded by the regex

perlpunk: I would hope that both perl and postgres would have the same limit if they are running on the same platform.

perlpunk: I put it into the Constants module now

lib/OpenQA/WebAPI.pm
```diff
@@ -391,83 +422,110 @@ sub startup ($self) {
...
+    $api_ra->delete('/assets/<id:bigint>')->name('apiv1_delete_asset')->to('asset#delete')
+      ->requires(bigint_range => 'id');
     $api_ra->delete('/assets/#type/#name')->name('apiv1_delete_asset_name')->to('asset#delete');
 
     # api/v1/test_suites
     $api_public_r->get('test_suites')->name('apiv1_test_suites')->to('table#list', table => 'TestSuites');
     $api_ra->post('test_suites')->to('table#create', table => 'TestSuites');
-    $api_public_r->get('test_suites/:id')->name('apiv1_test_suite')->to('table#list', table => 'TestSuites');
-    $api_ra->put('test_suites/:id')->name('apiv1_put_test_suite')->to('table#update', table => 'TestSuites');
+    $api_public_r->get('test_suites/<id:bigint>')->name('apiv1_test_suite')->to('table#list', table => 'TestSuites');
```
d3flex: why `test_suites` do not include the condition `->requires(bigint_range => 'id');` here??

perlpunk: Because I overlooked that apparently. fixed :)

lib/OpenQA/WebAPI.pm
```diff
@@ -184,18 +193,23 @@ sub startup ($self) {
...
     $r->get('/image/:md5_1/:md5_2/.thumbs/#md5_basename')->to('file#thumb_image');
-    $r->get('/group_overview/<groupid:num>' => [format => ['json', 'html']])->name('group_overview')
-      ->to('main#job_group_overview', format => undef);
-    $r->get('/parent_group_overview/<groupid:num>' => [format => ['json', 'html']])->name('parent_group_overview')
-      ->to('main#parent_group_overview', format => undef);
+
+    $r->get('/group_overview/<groupid:bigint>' => [format => ['json', 'html']])->requires(bigint_range => 'groupid')
+      ->name('group_overview')->to('main#job_group_overview', format => undef);
+
+
```
okurz:
```suggestion
    $r->get('/group_overview/<groupid:bigint>' => [format => ['json', 'html']])->requires(bigint_range => 'groupid')
      ->name('group_overview')->to('main#job_group_overview', format => undef);
```

perlpunk: I don't see the diff

perlpunk: oh, removing newlines. I see

PR #6922, opened 2026-01-22
Update CircleCI image to Leap 16.0

d3flex:
Some changes:
- drop obsolete 'rubygem(sass)>=3.7.4' replacing with perl-CSS-Sass
- Remove some duplicated packages from zypper
- fix paths which are related to the new usr-merge structure
- Apply workaround to https://github.com/openSUSE/obs-build/issues/1107 using `--no-refresh` on zypper.

related-issue: https://progress.opensuse.org/issues/180731

### DIFF
--- a/container/devel:openQA:ci/base/Dockerfile
+++ b/container/devel:openQA:ci/base/Dockerfile
@@ -4,27 +4,30 @@
 # in one go
 # see https://progress.opensuse.org/issues/67855
 #!NoSquash
-FROM opensuse/leap:15.6
+FROM opensuse/leap:16.0
 
 # only dependencies for CircleCI to be able to load/save the package cache
 # and fix permissions for the unprivileged user we use
-RUN zypper -n in tar gzip sudo
+# Apply `no-refresh` to work around 
+# https://github.com/openSUSE/obs-build/issues/1107
+RUN zypper -n --no-refresh in tar gzip sudo
 
 # these are autoinst dependencies
-RUN zypper install -y gcc-c++ cmake ninja pkgconfig\(opencv4\) pkg-config perl\(Module::CPANfile\) pkgconfig\(fftw3\) pkgconfig\(libpng\) pkgconfig\(sndfile\) pkgconfig\(theoraenc\) tesseract-ocr
+RUN zypper --no-refresh install -y gcc-c++ cmake ninja pkgconfig\(opencv4\) pkg-config perl\(Module::CPANfile\) pkgconfig\(fftw3\) pkgconfig\(libpng\) pkgconfig\(sndfile\) pkgconfig\(theoraenc\) tesseract-ocr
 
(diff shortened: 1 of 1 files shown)


container/devel:openQA:ci/base/Dockerfile
```diff
@@ -4,27 +4,29 @@
...
-RUN zypper install -y chromedriver
+RUN zypper --no-refresh install -y chromedriver
 
 ENV LANG=en_US.UTF-8
 
 ENV NORMAL_USER=squamata
 ENV USER=squamata
 
-RUN echo "$NORMAL_USER ALL=(ALL) NOPASSWD: ALL" >> /etc/sudoers
+RUN echo "$NORMAL_USER ALL=(ALL) NOPASSWD: ALL" >> /etc/sudoers.d/$NORMAL_USER
```
baierjan: I would guess that `/etc/sudoers.d/squamata` does not exists by default in a standard leap image and I would guess that the default umask will create a file with `0644` here. Hence the sudo will complain later with something like: `/etc/sudoers.d/squamata: bad permissions, should be mode 0440`. You should add `&& chmod` to fix it.

perlpunk: I added the not-ready label, thx

d3flex:
```
❯ podman run --rm -it testleap16 sudo ls -l /etc/sudoers.d/squamata
-r--r----- 1 root root 33 Jan 26 09:38 /etc/sudoers.d/squamata
```

baierjan: I don't know how testleap16 was build, but I am getting:
```
$  podman run --rm opensuse/leap:16.0 bash -c 'zypper -n --no-refresh in sudo &>/dev/null ; echo "squamata ALL=(ALL) NOPASSWD: ALL" >> /etc/sudoers.d/squamata; ls -la /etc/sudoers.d/'
total 4
drwxr-x--- 1 root root   16 Jan 26 12:37 .
drwxr-xr-x 1 root root 1212 Jan 26 12:37 ..
-rw-r--r-- 1 root root   33 Jan 26 12:37 squamata
```

perlpunk: it looks ok in `registry.opensuse.org/devel/openqa/ci/containers16.0/base:latest`

perlpunk: Ah, now I understand your question @baierjan 
You probably missed that @d3flex added a chmod call already. @d3flex maybe try to be clearer about that in the future

baierjan: Ah, so it was a "validation run" and not a reproducer attempt then. Cool! Yeah, I saw the fix in the other PR.

PR #6923, opened 2026-01-23
test: Improve test coverage under `lib/OpenQA/WebAPI/Controller`

Martchus:
Related ticket: https://progress.opensuse.org/issues/189204

### DIFF
--- a/lib/OpenQA/WebAPI/Controller/Step.pm
+++ b/lib/OpenQA/WebAPI/Controller/Step.pm
@@ -262,14 +262,16 @@ sub edit ($self) {
     $self->render('step/edit');
 }
 
+sub _new_click_point ($click_point) { $click_point ? (click_point => $click_point) : () }
+
 sub _new_match ($area) {
     return {
         xpos => int $area->{x},
         ypos => int $area->{y},
         width => int $area->{w},
         height => int $area->{h},
-        click_point => $area->{click_point},
         type => 'match',
+        _new_click_point($area->{click_point}),
     };
 }
 
@@ -470,7 +472,7 @@ sub _calc_match ($area) {
         height => int $area->{h},
         type => $area->{result},
         similarity => int($area->{similarity} + 0.5),
-        click_point => $area->{click_point},
+        _new_click_point($area->{click_point}),
     };
 }
 


lib/OpenQA/Schema/Result/JobModules.pm
```diff
@@ -93,6 +93,7 @@ sub results ($self, %options) {
         catch ($e) { push @$errors, qq{Unable to read "$file": $e} if $errors; undef }
       }
       or return {};
+    print("details for $file: $json_data\n");
```
okurz: ^^

lib/OpenQA/WebAPI/Controller/Step.pm
```diff
@@ -262,36 +262,30 @@ sub edit ($self) {
     $self->render('step/edit');
 }
 
+sub _new_match ($area) {
+    return {
+        xpos => int $area->{x},
+        ypos => int $area->{y},
+        width => int $area->{w},
+        height => int $area->{h},
+        click_point => $area->{click_point},
```
perlpunk: This will result in `click_point => undef` if `$area->{click_point}` does not exist or is undef, which was prevented by the previous code.

lib/OpenQA/WebAPI/Controller/Step.pm
```diff
@@ -462,24 +462,22 @@ sub save_needle_ajax ($self) {
...
-    }
+sub _calc_match ($area) {
+    return {
+        xpos => int $area->{x},
+        ypos => int $area->{y},
+        width => int $area->{w},
+        height => int $area->{h},
+        type => $area->{result},
+        similarity => int($area->{similarity} + 0.5),
+        click_point => $area->{click_point},
```
perlpunk: Same as my other comment

lib/OpenQA/WebAPI/Controller/Step.pm
Martchus: Yes, but I don't think it'll make a difference.

perlpunk: It will consume more space, so depending on where this is stored it might make a tiny difference. Just saying, not blocking.

Martchus: I guess I can nevertheless remove it so we don't have `"click_point": null` in the JSON.

PR #6924, opened 2026-01-23
draft: Allow archiving jobs in space-aware result cleanup

Martchus:
Related ticket: https://progress.opensuse.org/issues/193999

### DIFF
--- a/docs/UsersGuide.asciidoc
+++ b/docs/UsersGuide.asciidoc
@@ -1534,12 +1534,22 @@ thresholds.
 
 The deletion happens in the following order:
 
-1. Videos of unimportant jobs
-2. Results of unimportant jobs
-3. Videos of important jobs
-4. Results of important jobs
+1. Videos of unimportant jobs are deleted
+2. Results of unimportant jobs are deleted
+3. Videos of important jobs are deleted
+4. Results of important jobs are deleted
 
-Within each group the oldest jobs are deleted first.
+Within each group the oldest jobs are deleted first. The deletion is stopped as
+soon as the specified percentage of space is free again. This is evaluated after
+handling each individual job. The archive is cleaned up first.
+
+When enabling `archive_via_min_free_disk_space_cleanup` jobs will be archived
+as long as the archive is not full. That means cleanup actions will be performed
+in this order:
+
+1. Unimportant jobs are archived if there is enough room in the archive
+2. Important jobs are archived if there is enough room in the archive
+3. Continue with steps 1 to 4 from the previously mentioned list.
 
 It is also possible to abort cleanup tasks early in case there is still
 enough headroom on relevant file systems anyway. This can be enabled using the
--- a/etc/openqa/openqa.ini
+++ b/etc/openqa/openqa.ini
@@ -301,6 +301,10 @@ concurrent = 0
 ## Allows running the extended job result cleanup via
 ## `…_min_free_disk_space_percentage` settings in dry-run mode
 #dry_min_free_disk_space_cleanup = 0
+## Archives results if set to `1` and `results_min_free_disk_space_percentage`
+## is exceeded (instead of deleting videos/results, as long as the archive is
+## not full)
+#archive_via_min_free_disk_space_cleanup = 0
 ## Duration to keep minion jobs in the database, in seconds
 #minion_job_max_age = 604800
 ## Default number of templates to include in job_templates api response (to
--- a/lib/OpenQA/Task/Job/Limit.pm
+++ b/lib/OpenQA/Task/Job/Limit.pm
@@ -165,7 +165,7 @@ sub _account_for_deletion ($margin_bytes, $margin_bytes_main_storage, $deleted_r
 }
 
 sub _delete_results ($dry, $jobs, $max_job_id, $not_important_cond, $important_cond, $margin_bytes,
-    $margin_bytes_main_storage, $archived)
+    $margin_bytes_main_storage, $margin_bytes_archive_storage, $archived)
 {
     # caveat: The subsequent cleanup simply deletes stuff from old jobs first. It does not take the retention periods
     #         configured on job group level into account anymore.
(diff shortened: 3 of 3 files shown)


docs/UsersGuide.asciidoc
```diff
@@ -1534,12 +1534,22 @@ thresholds.
...
+Within each group the oldest jobs are deleted first. The deletion is stopped as
+soon as the specified percentage of space is free again. This is evaluated after
+handling each individual job. The archive is cleaned up first.
+
+When enabling `archive_via_min_free_disk_space_cleanup` jobs will be archived
+as long as the archive is not full. That means cleanup actions will be performed
+in this order:
+
+1. Unimportant jobs are archived if there is enough room in the archive
+2. Important jobs are archived if there is enough room in the archive
```
Martchus, line 1551: I'm not sure whether this order actually makes sense. Preferring unimportant jobs would of course be in-line with how we order the deletions. However, it also means that unimportant jobs are more likely archived (instead of deleted) than important jobs.

The "if there is enough room in the archive" part would in our case mean "if there is enough space left in the archive at all" considering you don't want to enable space-aware cleanup for archiving.

docs/UsersGuide.asciidoc
```diff
@@ -1534,12 +1534,22 @@ thresholds.
...
-
-Within each group the oldest jobs are deleted first.
+1. Videos of unimportant jobs are deleted
+2. Results of unimportant jobs are deleted
+3. Videos of important jobs are deleted
+4. Results of important jobs are deleted
+
+Within each group the oldest jobs are deleted first. The deletion is stopped as
+soon as the specified percentage of space is free again. This is evaluated after
+handling each individual job. The archive is cleaned up first.
```
Martchus, line 1544: > The archive is cleaned up first.

This is just how it currently is. Considering you don't plan enabling cleanup of the archive in production it shouldn't make a difference for us so we probably don't need to think too much about this. I'm just mentioning it here for explicitness.

docs/UsersGuide.asciidoc
```diff
@@ -1534,12 +1534,22 @@ thresholds.
...
+
+When enabling `archive_via_min_free_disk_space_cleanup` jobs will be archived
+as long as the archive is not full. That means cleanup actions will be performed
+in this order:
+
+1. Unimportant jobs are archived if there is enough room in the archive
+2. Important jobs are archived if there is enough room in the archive
+3. Continue with steps 1 to 4 from the previously mentioned list.
```
Martchus, line 1552: That's *maybe* something one doesn't want; maybe one is *just* interested in the space-aware cleanup for additional archiving (but not additional deleting).

So maybe the boolean setting `archive_via_min_free_disk_space_cleanup = 0` should be `archiving_via_min_free_disk_space_cleanup = no` where the value can be `no`, `exclusively` or `additionally`.

docs/UsersGuide.asciidoc
```diff
@@ -1534,12 +1534,22 @@ thresholds.
...
-1. Videos of unimportant jobs
-2. Results of unimportant jobs
-3. Videos of important jobs
-4. Results of important jobs
-
-Within each group the oldest jobs are deleted first.
+1. Videos of unimportant jobs are deleted
+2. Results of unimportant jobs are deleted
+3. Videos of important jobs are deleted
+4. Results of important jobs are deleted
```
okurz, line 1540: Why the addition of "are deleted"? The header still says "The deletion happens in the following order:". sounds redundant

docs/UsersGuide.asciidoc
okurz: As the section https://github.com/os-autoinst/openQA/pull/6924/files#diff-4703823e594413b96467dcfe16688eea4421730bc6275ec634afc777ec8f9602R1498 "Archiving of *important* jobs can be enabled:" so that section would need to be updated as well.

Why is there a step "important jobs are archived" here at all? Isn't that "normal" archiving anyway?


> considering you don't want to enable space-aware cleanup for archiving.

The "archive" sounds like something that openQA should not delete from unless it's explicitly selected by test result owners, e.g. job group admins that set a non-zero expiration. Of course we can leave such choice to instance admins if they want to enable space-aware cleanup but right now how o3 and osd is used I see a more important feature than space-aware-archive-cleanup https://progress.opensuse.org/issues/192190 "Prevent openQA depleting storage with new openQA jobs" so that no data corruption happens because the archive filesystem would run full. The openQA instance would simply not add more results which are stored in "results" first but then due to normal archiving or "archive_via_min_free_disk_space_cleanup" move results to archive.

I see good arguments for both orders. Archiving unimportant jobs first can be good because that means that important jobs can be  kept on a potentially faster storage for longer. But I guess first I need to understand what you would respond to my above question.

etc/openqa/openqa.ini
```diff
@@ -301,6 +301,10 @@ concurrent = 0
 ## Allows running the extended job result cleanup via
 ## `…_min_free_disk_space_percentage` settings in dry-run mode
 #dry_min_free_disk_space_cleanup = 0
+## Archives results if set to `1` and `results_min_free_disk_space_percentage`
+## is exceeded (instead of deleting videos/results, as long as the archive is
+## not full)
+#archive_via_min_free_disk_space_cleanup = 0
```
okurz, line 307:
```suggestion
## Archives results if set to `1` and `results_min_free_disk_space_percentage`
## is exceeded instead of deleting videos/results, as long as the archive is
## not full
#archive_via_min_free_disk_space_cleanup = 0
```

docs/UsersGuide.asciidoc
okurz: ok, that's a good point. Your suggestion sounds good but I would avoid the words "yes" or "no" and reserve those for booleans (if we ever want to go from "0|1" to "no|yes"). So better `none`, `exclusively` or `additionally`

docs/UsersGuide.asciidoc
Martchus: I wanted to be explicit here, especially also because the list is now referred to later on - when it isn't just about deletions anymore.

etc/openqa/openqa.ini
Martchus: > Why is there a step "important jobs are archived" here at all? Isn't that "normal" archiving anyway?

The "normal" archiving works by moving jobs that are only kept longer because they are important based on the configured retentions. **This is how you wanted me to implement it.** So the "normal" archiving does not take a full results file system into account. This is what that step is about. So if results get too full (because time-based retentions are "promising" longer duration then what is possible size-wise) this cleanup takes place - and it applies to both - important and unimportant jobs. So there can be unarchived important jobs that would be archived at this step.

docs/UsersGuide.asciidoc
okurz: Ok, then make the header more generic, e.g. cleanup

etc/openqa/openqa.ini
okurz: > This is how you wanted me to implement it.

If you think I have unreasonable requests then please let me know. If I ask you to implement something that I could not make clear yet please also let me know. Otherwise I see no point in mentioning this.

> So there can be unarchived important jobs that would be archived at this step.

Maybe my initial feature request is flawed then asking for unimportant jobs to be included in archiving. How about we try to make it simpler and only try to archive more important jobs (and never unimportant jobs) if results needs cleanup and if archive has space?

Martchus: I guess we could indeed only consider important jobs here. That way we wouldn't need to find an answer to my question.

PR #6925, opened 2026-01-23
feat: Make test creation discoverable to all users

Martchus:
See individual commit messages for details.

Related ticket: https://progress.opensuse.org/issues/189921

### DIFF
--- a/templates/webapi/test/create.html.ep
+++ b/templates/webapi/test/create.html.ep
@@ -63,13 +63,13 @@
       <%= help_popover('Test repository' => '<p>Specifies additional settings as newline-separated <code>KEY=value</code>-pairs that will be assigned to each job. ' .
         'Some settings also influence the job creation itself, e.g. <code>_OBSOLETE</code>.</p>',
         'https://open.qa/docs/#_spawning_multiple_jobs_based_on_templates_isos_post', 'the documentation', 'left') %>
-      <textarea type="text" class="form-control key-value-pairs" id="create-tests-settings" rows="15"></textarea>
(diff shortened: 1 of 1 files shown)


templates/webapi/layouts/navbar.html.ep
```diff
@@ -10,7 +10,7 @@
          <li class='nav-item' id="all_tests">
             %= link_to 'All Tests' => url_for('tests') => class => 'nav-link', title => 'Lists all tests grouped by state'
          </li>
-          % if ($current_user && $current_user->is_operator) {
+          % if ($current_user) {
```
okurz, line 13: This looks a bit like a trap that we allow logged in users to click a link that sounds like "oh, cool, I can create tests" only to be greeted by a warning. How about to always display but grey out and disable the action with a hover message instead of the flash? Or maybe grey and if one *still* clicks the flash if the former is not feasible

Martchus: I don't think a tooltip on a disabled menu entry is good UX as users might think they cannot interact with the disabled element at all. Graying text out also means it isn't going to be very readable anymore - at least with out color scheme (see e.g. https://getbootstrap.com/docs/5.3/components/navbar/#supported-content).

I'm also not sure why you're considering it a trap. It is not one like could just move on after clicking on "Create new …".

Martchus: But if you nevertheless want the tooltip I can implement it of course. It is probably quite easy.

okurz: ok, fine. Leave it as is

PR #6927, opened 2026-01-23
navbar: add new item in menu to link MCP documentation

m-dati:
Add in the _openQA webUI_, in the rightmost drop-down menu, a new item to launch a link pointing to *MCP Support* paragraph of the open.qa documentation, to help configuring openQA MCP in AI clients.

- Ticket:  https://progress.opensuse.org/issues/189069

### DIFF
--- a/t/ui/05-auth_mcp.t
+++ b/t/ui/05-auth_mcp.t
@@ -1,0 +1,41 @@
+# Copyright SUSE LLC
+# SPDX-License-Identifier: GPL-2.0-or-later
+
+use Test::Most;
+
+use Test::Mojo;
+use Test::Warnings ':report_warnings';
+use FindBin;
+use lib "$FindBin::Bin/../lib", "$FindBin::Bin/../../external/os-autoinst-common/lib";
+use OpenQA::Test::TimeLimit '8';
+use OpenQA::Test::Case;
+
+my $test_case = OpenQA::Test::Case->new;
+$test_case->init_data(fixtures_glob => '03-users.pl');
+
+my $t = Test::Mojo->new('OpenQA::WebAPI');
+
+subtest 'First access no login' => sub {
+    $t->get_ok('/tests')->status_is(200);
+    my $actions = OpenQA::Test::Case::trim_whitespace($t->tx->res->dom->at('#user-action')->all_text);
+    is $actions, 'Login', 'no-one logged in';
+};
+
+subtest 'MCP disabled' => sub {
+    $test_case->login($t, 'percival');
+    $t->app->config->{global}{mcp_enabled} = 'no';
+    my $mcp_help = '';
+    $t->get_ok('/tests')->status_is(200);
+    my $actions = OpenQA::Test::Case::trim_whitespace($t->tx->res->dom->at('#user-action')->all_text);
+    like $actions, qr/API help ${mcp_help}Changelog Logout/, 'no MCP entry shown for operator when MCP not enabled';
+};
+
+subtest 'MCP enabled' => sub {
+    $t->app->config->{global}{mcp_enabled} = 'read_only';
+    my $mcp_help = 'MCP help ';
+    $t->get_ok('/tests')->status_is(200);
+    my $actions = OpenQA::Test::Case::trim_whitespace($t->tx->res->dom->at('#user-action')->all_text);
+    like $actions, qr/API help ${mcp_help}Changelog Logout/, 'MCP entry is shown for operator when MCP is enabled';
+};
+
+done_testing();
--- a/templates/webapi/admin/mcp_help.html.ep
+++ b/templates/webapi/admin/mcp_help.html.ep
@@ -1,0 +1,8 @@
+% layout 'bootstrap';
+% title 'MCP help';
+  <h2><%= title %></h2>
+  %= include 'layouts/info'
+  <p>
+      To configure an MCP client to connect to openQA, check out the
+      <%= link_to 'documentation about MCP support' => url_for('https://open.qa/docs/#_mcp_support') %>.
+  </p>
--- a/templates/webapi/layouts/navbar.html.ep
+++ b/templates/webapi/layouts/navbar.html.ep
@@ -97,6 +97,9 @@
                   % }
                   %= link_to "Appearance" => url_for('appearance') => class => 'dropdown-item'
                   %= link_to "API help" => url_for('api_help') => class => 'dropdown-item'
+                  % if (config->{global}{mcp_enabled} eq 'read_only') {
+                      %= link_to 'MCP help' => url_for('mcp_help') => class => 'dropdown-item'
+                  % }
                   %= link_to 'Changelog' => url_for('changelog') => class => 'dropdown-item'
                   %= link_to 'Logout' => url_for('logout') => 'data-method' => 'delete' => class => 'dropdown-item'
                 </div>
--- a/lib/OpenQA/WebAPI.pm
+++ b/lib/OpenQA/WebAPI.pm
@@ -203,6 +203,7 @@ sub startup ($self) {
     $r->get('/dashboard_build_results' => [format => ['json', 'html']])->name('dashboard_build_results')
       ->to('main#dashboard_build_results', format => undef);
     $r->get('/api_help' => sub ($c) { $c->render('admin/api_help') })->name('api_help');
+    $r->get('/mcp_help' => sub ($c) { $c->render('admin/mcp_help') })->name('mcp_help');
 
     # Default route
     $r->get('/' => sub ($c) { $c->render('main/index') })->name('index');


templates/webapi/layouts/navbar.html.ep
```diff
@@ -97,6 +97,7 @@
                   % }
                   %= link_to "Appearance" => url_for('appearance') => class => 'dropdown-item'
                   %= link_to "API help" => url_for('api_help') => class => 'dropdown-item'
+                  %= link_to "AI-MCP help" => url_for('https://open.qa/docs/#_mcp_support') => class => 'dropdown-item'
```
okurz: As I wrote in https://progress.opensuse.org/issues/189069#note-27

I suggest to use ✦ (U+2726) with a usual coloring

```
.ai-icon {
    color: #7b61ff;
    font-style: normal;
}
```

```suggestion
                  %= link_to "✦ AI-MCP help" => url_for('https://open.qa/docs/#_mcp_support') => class => 'dropdown-item'
```

but *only* show that if the MCP is enabled on the host

m-dati: Do you mean, _only_ when openqa.ini contains `mcp_enabled`?

m-dati: I think that this help should be always present, to inform and drive the eventual user to tweak that value too, when MCP is not present or disabled.

okurz: > I think that this help should be always present, to inform and drive the eventual user to tweak that value too, when MCP is not present or disabled.

What values should a user "tweak" if the instance does not support MCP?

m-dati: If we talk about `openqa.ini`, the linked documentation informs that 
```
MCP ... needs to be enabled manually in openqa.ini:

[global]
mcp_enabled = read-only

Once enabled the experimental MCP endpoint becomes available ...
```
Therefore, if one is not aware yet of this new capability, the presence of that help _could_ spur to activate it.

m-dati: Please look [here](http://10.146.4.26/) a temporary instance of the UI with the last update of the menu.

templates/webapi/layouts/navbar.html.ep
```diff
@@ -97,6 +97,7 @@
                   % }
                   %= link_to "Appearance" => url_for('appearance') => class => 'dropdown-item'
                   %= link_to "API help" => url_for('api_help') => class => 'dropdown-item'
+                  %= link_to "✦ AI-MCP help" => url_for('https://open.qa/docs/#_mcp_support') => class => 'dropdown-item'
```
m-dati: The proposal in https://github.com/os-autoinst/openQA/pull/6927#discussion_r2722133738 and [poo189069](https://progress.opensuse.org/issues/189069#note-27) for a special icon or character near the MCP menu item (the symbol of `gemini`), is not well clear to me, because nowhere else we use icons or dotted lists in openQA and applying such element we create a strong difference in the items visualization. 
Also, it is failing in the CI checks. So, May be I didn't catch well that note's meaning.

m-dati: But did you mean to set it as a check box trigger, making the "only symbol" to appear/disappear, when mcp is enabled/disabled, near the item title ?
Like:
`* AI-MCP...`: enabled
`  AI-MCP...`: disabled 
?

okurz: No, really just as part of static text. But the star misaligns the entry with the other ones. Unless you add padding to all entries in the menu you should put that on the right hand side then.

templates/webapi/layouts/navbar.html.ep
okurz: > Therefore, if one is not aware yet of this new capability, the presence of that help _could_ spur to activate it.

But a user is not the same as an instance admin. The *user* should only be presented with what works on this instance.

m-dati: done

templates/webapi/layouts/navbar.html.ep
m-dati: done

templates/webapi/layouts/navbar.html.ep
perlpunk: How about a greyed out entry like you also suggested for the "Create" menu?
If the goal is to make the MCP feature popular, then it might be useful to have that menu entry in every instance, just greyed out.

templates/webapi/layouts/navbar.html.ep
m-dati: The star * seems still causing errors in CI.... Unless that check routine itself is unable to handle the star  🙃

m-dati: The star * seems still causing errors in CI.... Unless that check routine itself is wrong 🙃

templates/webapi/layouts/navbar.html.ep
m-dati: normal users cannot configure mcp, no key management: they could see the Grey but do nothing,... unless they push for a role change 😃

templates/webapi/layouts/navbar.html.ep
m-dati: Just managed to have CI passed, with the special star too: an unit test update needed 🙂

templates/webapi/layouts/navbar.html.ep
```diff
@@ -97,6 +97,9 @@
                   % }
                   %= link_to "Appearance" => url_for('appearance') => class => 'dropdown-item'
                   %= link_to "API help" => url_for('api_help') => class => 'dropdown-item'
+                  % if (is_operator) {
```
okurz: why only operator?

And, again, the menu entry must only be shown if MCP is enabled on the server. Same as I explained in https://github.com/os-autoinst/openQA/pull/6927#discussion_r2722655414 : Do not point the user to documentation that he can't follow as the user is not the instance admin.

templates/webapi/layouts/navbar.html.ep
perlpunk: > Just managed to have CI passed, with the special star too: an unit test update needed 🙂

Are you also testing this locally before pushing? Might be more efficient than waiting for CI for every fix :)

templates/webapi/layouts/navbar.html.ep
m-dati: The link to the documentation is a simple help and an information to make people aware of the openQA MCP and should *always* be present; it isn't a tool to use only when needed.

Instead, when it comes to _show for mcp-enabled only_, it means that user shall already be aware about the mcp feature, to make the help link appear by updating the `openqa.ini`: the ones that don't know yet this feature and haven't yet that setting, will not be informed of it, in the webUI, making this link less useful.

But, to followup your request and explain the `is_operator` limitation, it means "at least operator" and includes 'admin', because the 'normal' "users" have _no api key management_ then they can't perform any mcp configuration at all, therefore I applied that limitation.

Anyway, given that the menu entry **must** *only be shown if MCP is enabled*, I'm going to apply it.

Martchus: I'm wondering whether we should allow normal users to manage their API keys - even though they probably can't do much with them. Using read-only MCP APIs would perhaps be a use case for this now - although it raises the question why one needs an API key at all to use read-only APIs that probably only access data that's publicly accessible anyway.

templates/webapi/layouts/navbar.html.ep
perlpunk: How about showing a greyed out entry for admin users at least, if MCP is not configured? This way anyone trying out setting up an openQA instance would see that. (Note: I don't need that, but I thought the goal was to advertise this functionaliy)

templates/webapi/admin/mcp_help.html.ep
```diff
@@ -0,0 +1,11 @@
+% layout 'bootstrap';
+% title 'AI✦MCP help';
+<div>
+    <h2><%= title %></h2>
+    
+    %= include 'layouts/info'
+    
+    <p>To configure the openQA MCP Server, please look</p> 
+    <p>in the open.qa documentation the section <%= link_to 'MCP support' => url_for('https://open.qa/docs/#_mcp_support') %>.
+    </p>
```
Martchus: Please don't start another paragraph in the middle of a sentence. Also, don't use a domain in the middle of a sentence like this. "server" should probably also be lower case.


```suggestion
    <p>
        To configure the openQA MCP server, check out the
        <%= link_to 'documentation section about MCP support' => url_for('https://open.qa/docs/#_mcp_support') %>.
    </p>
```

templates/webapi/admin/mcp_help.html.ep
```diff
@@ -0,0 +1,11 @@
+% layout 'bootstrap';
+% title 'AI✦MCP help';
+<div>
+    <h2><%= title %></h2>
+    
+    %= include 'layouts/info'
+
```
Martchus:
```suggestion
    %= include 'layouts/info'
```

templates/webapi/admin/mcp_help.html.ep
```diff
@@ -0,0 +1,11 @@
+% layout 'bootstrap';
+% title 'AI✦MCP help';
+<div>
```
Martchus: Is this `div` actually needed?

m-dati: It is the same, similar, as `templates/webapi/admin/api_help.html.ep`

templates/webapi/admin/mcp_help.html.ep
m-dati: As I replied in https://github.com/os-autoinst/openQA/pull/6927#discussion_r2728323784, then also `api_help.html.ep`, that I did't touch, should be updated accordingly? BTW, I change it in next commit.

templates/webapi/admin/mcp_help.html.ep
Martchus: That doesn't answer my question. **If** the `div` is not needed we shouldn't have it, regardless where this is copied from.

t/ui/05-auth.t
```diff
@@ -71,16 +72,28 @@ is($res, 'Login', 'no-one logged in');
...
     $actions,
-    qr/Logged in as perci Operators Menu.*Manage API keys Appearance API help Changelog Logout/,
+    qr/Logged in as perci Operators Menu.*Manage API keys Appearance API help ${mcp_help}Changelog Logout/,
+    'perci has operator links'
+);
+# mcp enabled
+$t->app->config->{global}{mcp_enabled} = 'read_only';
+$mcp_help = 'AI\x{2726}MCP help ';
+$actions = OpenQA::Test::Case::trim_whitespace(
+    $t->get_ok('/tests')->status_is(200)->tx->res->dom->at('#user-action')->all_text);
```
Martchus: Add some point this method-chaining madness should end:
```suggestion
$t->get_ok('/tests')->status_is(200);
$actions = OpenQA::Test::Case::trim_whitespace($t->tx->res->dom->at('#user-action')->all_text);
```

t/ui/05-auth.t
```diff
@@ -9,6 +9,7 @@ use Test::Mojo;
 use Test::Warnings ':report_warnings';
 use OpenQA::Test::TimeLimit '8';
 use OpenQA::Test::Case;
+use Data::Dump qw(dump);
```
Martchus: This doesn't seem to be used.

perlpunk: @m-dati  and if you just want to look at data during testing, use `always_explain` (it's automatically exported by Test::Most).

m-dati: Sorry I forgot to remove it, used for temporary local testing :)

m-dati: done

t/ui/05-auth.t
m-dati: ~I agree with you, but IMHO this update should be done for **all** the similar occurrences, in a separated **full refactoring** ticket, or at least in another PR,   not only once. Thus it seems out of scope here, compared with the minimal changes needed for this ticket.~

templates/webapi/admin/mcp_help.html.ep
m-dati: ~Infact it is not about the file that the new one is coming from, but the similarity and comparison with all the other `*htmp.ep` present, where an initial \<div\> block is a clean way to envelop, separate and handle a text block. It does't appear syntattically wrong.~

t/ui/05-auth.t
m-dati: Done

templates/webapi/admin/mcp_help.html.ep
m-dati: done

t/ui/05-auth.t
```diff
@@ -71,16 +71,28 @@ is($res, 'Login', 'no-one logged in');
...
+# mcp disabled
+$t->app->config->{global}{mcp_enabled} = 'no';
+my $mcp_help = '';
+$t->get_ok('/tests')->status_is(200);
+my $actions = OpenQA::Test::Case::trim_whitespace($t->tx->res->dom->at('#user-action')->all_text);
+like(
+    $actions,
+    qr/Logged in as perci Operators Menu.*Manage API keys Appearance API help ${mcp_help}Changelog Logout/,
+    'perci has operator links'
+);
```
okurz: use subtests instead of comments, avoid parentheses around test actions, correct the test description string and prevent line breaks with tests only testing the relevant parts

```suggestion
subtest 'mcp disabled' => sub {
    $t->app->config->{global}{mcp_enabled} = 'no';
    my $mcp_help = '';
    $t->get_ok('/tests')->status_is(200);
    my $actions = OpenQA::Test::Case::trim_whitespace($t->tx->res->dom->at('#user-action')->all_text);
    like $actions, qr/API help ${mcp_help}Changelog/, 'mcp help shown when disabled';
};
```

but obviously this does not make sense as we wanted the link to *not* show up when MCP is disabled.

Also this does not fit into t/ui/05-auth.t

templates/webapi/admin/mcp_help.html.ep
```diff
@@ -0,0 +1,9 @@
+% layout 'bootstrap';
+% title 'AI✦MCP help';
+  <h2><%= title %></h2>
+  %= include 'layouts/info'
+  <p>
+      To configure the openQA MCP server, check out the
```
okurz: The openQA MCP server *is* configured. This is about the client

templates/webapi/layouts/navbar.html.ep
```diff
@@ -97,6 +97,9 @@
                   % }
                   %= link_to "Appearance" => url_for('appearance') => class => 'dropdown-item'
                   %= link_to "API help" => url_for('api_help') => class => 'dropdown-item'
+                  % if (config->{global}{mcp_enabled} eq 'read_only') {
+                      %= link_to 'AI✦MCP help' => url_for('mcp_help') => class => 'dropdown-item'
```
okurz: why "AI✦MCP"?

m-dati: Why not 😃 ? For aesthetic reasons: 
you asked to put the char ✦ (an AI icon) at end "`AI MCP help ✦`": but this way it seemed to have not much meaning, for the item.
Setting it inbetween the AI and MCP `AI✦MCP` is much more a sense of connection, like MCP does.
If you prefer the other way around, I will change it.

t/ui/05-auth.t
m-dati: The mcp help unit test should be in this auth file because the check is based on the **role** of the logging **user**.
Unless a brand new test file like `5-mcp.t` is preferred.

This specific user "`percival`" is **operator** role then we expect that his webUI menu shall contain the new item, when the mcp is enabled, against the other user `morgana` being normal user, that shall not show that item.
As counterproof, the unit test shall also show the opposite case, when the mcp is *not configured*, so that `percival` has no mcp help in his menu.

m-dati: Going to apply the subtest suggestion.
But the full file shoud be refactored applying subtests in many other places.

m-dati: A more kind clarification about `...avoid parentheses around test actions,...`: in this module, parentheses are used in every `like/unlike`, therefore I considered that it was the preferred style 🙂. Going to remove parentheses.

templates/webapi/layouts/navbar.html.ep
okurz: > […] although it raises the question why one needs an API key at all to use read-only APIs that probably only access data that's publicly accessible anyway.

At least that covers authentication even though authorization isn't the issue here, i.e. "at least we know who was it" :)

templates/webapi/layouts/navbar.html.ep
Martchus: I also find it nicer in the middle. In any case it isn't really meaningful but in the middle it looks at least nice, like some kind of weird dash.

t/ui/05-auth_mcp.t
```diff
@@ -0,0 +1,44 @@
+# Copyright 2014-2026 SUSE LLC
```
m-dati: As followup of https://github.com/os-autoinst/openQA/pull/6927#discussion_r2731394859, I moved those mcp checks into a new unit test file, hoping to meet the expectations

Martchus:
```suggestion
# Copyright SUSE LLC
```

t/ui/05-auth_mcp.t
```diff
@@ -0,0 +1,44 @@
...
+my ($actions, $mcp_help);
+
+subtest 'first access no login' => sub {
+    my $actions = OpenQA::Test::Case::trim_whitespace(
+        $t->get_ok('/tests')->status_is(200)->tx->res->dom->at('#user-action')->all_text);
+    is $actions, 'Login', 'no-one logged in';
+    $t->get_ok('/api_keys')->status_is(302);
+};
+
+subtest 'webUI operator user login, with mcp disabled' => sub {
```
Martchus:
```suggestion
subtest 'web UI operator user login, with MCP disabled' => sub {
```

t/ui/05-auth_mcp.t
```diff
@@ -0,0 +1,44 @@
...
+    $test_case->login($t, 'percival');
+    $t->app->config->{global}{mcp_enabled} = 'no';
+    $mcp_help = '';
+    $t->get_ok('/tests')->status_is(200);
+    $actions = OpenQA::Test::Case::trim_whitespace($t->tx->res->dom->at('#user-action')->all_text);
+    like $actions, qr/API help ${mcp_help}Changelog Logout/, 'perci has operator links, case MCP disabled';
+};
+
+subtest 'webUI operator user after mcp enabled' => sub {
```
Martchus:
```suggestion
subtest 'web UI operator user after MCP enabled' => sub {
```

t/ui/05-auth_mcp.t
```diff
@@ -0,0 +1,44 @@
...
+my $test_case = OpenQA::Test::Case->new;
+$test_case->init_data(fixtures_glob => '03-users.pl');
+
+my $t = Test::Mojo->new('OpenQA::WebAPI');
+
+my ($actions, $mcp_help);
+
+subtest 'first access no login' => sub {
+    my $actions = OpenQA::Test::Case::trim_whitespace(
+        $t->get_ok('/tests')->status_is(200)->tx->res->dom->at('#user-action')->all_text);
```
Martchus:
```suggestion
    $t->get_ok('/tests')->status_is(200);
    my $actions = OpenQA::Test::Case::trim_whitespace($t->tx->res->dom->at('#user-action')->all_text);
```

Although I'm wondering whether we need this subtest at all. We have many menu entries that are only shown to logged-in users and we probably don't need to test like that for each individual menu entry.

m-dati: In t/ui/05-auth.t L#78, the full operator menu is checked: given that a more item could appear or not depending on the openqa configuration, I thought that a proper test was expected.

templates/webapi/layouts/navbar.html.ep
nicksinger: From an accessibility standpoint it is not so nice to have a random unicode char in the middle of two abbreviations (which is btw. called "BLACK FOUR POINTED STAR" - not "AI icon")

m-dati: It isn't a random char, but indeed, that char is generally used specifically as `Gemini AI` [icon](https://icons8.com/icon/mrNoLXFmvXDX/gemini-ai). It was a proposed in this PR https://github.com/os-autoinst/openQA/pull/6927#discussion_r2722133738 days ago. 
The AI generic is [this](https://icons8.com/icon/POBc2SrrhhnF/ai)

m-dati: About _icons_ in the menu items: I see the issue that actually _no icon_ is used (yet) in the menu items, therefore, any single icon addition would deeply change the menu appearance and that item would have a big difference in terms of separation, making uneven the menu style. 

Therefore, a _single char_ is less invasive and the more fitting the intent of pointing an AI-like item out. 
But its value and position seems that still need to be better agreed.

templates/webapi/layouts/navbar.html.ep
m-dati: @okurz: should we _grey_ the help item when MCP not configured, for operators/admin, instead of hiding it?

templates/webapi/layouts/navbar.html.ep
okurz: I think it is more aggreable if we leave out the character completely. I consider @nicksinger's point about accessibility important. I see this addressable only with using either an icon or character but hidden using aria attributes. If you can do that then cool otherwise just leave out the character completely.

templates/webapi/layouts/navbar.html.ep
okurz: No, the operator or admin of a webUI is not the same role as an instance admin controlling the configuration of the instance. We do not want to have greyed out items for all other potential features available over config. I don't see a reason why MCP should be treated as exception in this regard.

templates/webapi/admin/mcp_help.html.ep
```diff
@@ -0,0 +1,8 @@
+% layout 'bootstrap';
+% title 'AI✦MCP help';
+  <h2><%= title %></h2>
+  %= include 'layouts/info'
+  <p>
+      To configure the openQA MCP client, check out the
```
Martchus: This makes it sound like we would provide a special client as part of openQA.


```suggestion
      To configure an MCP client to connect to openQA, check out the
```

templates/webapi/layouts/navbar.html.ep
Martchus: In theory `<span aria-hidden="true">✦</span>` would do it but then you probably cannot use the `link_to` helper anymore. So maybe just:


```suggestion
                      %= link_to 'MCP help' => url_for('mcp_help') => class => 'dropdown-item'
```

m-dati: Only a kind note about ARIA: it is a special set of attributes to address _screen readers_ to **read on speakers** the status 'icon is hidden', as support in disabilites, but not to _trigger_ the action itself i.e. icon-hiding, performed via Css/Js or React.
Anyway, I just removed the discussed symbol

templates/webapi/admin/mcp_help.html.ep
m-dati: done

t/ui/05-auth_mcp.t
```diff
@@ -0,0 +1,43 @@
...
+    is $actions, 'Login', 'no-one logged in';
+};
+
+subtest 'web UI operator user login, with MCP disabled' => sub {
+    $test_case->login($t, 'percival');
+    $t->app->config->{global}{mcp_enabled} = 'no';
+    $mcp_help = '';
+    $t->get_ok('/tests')->status_is(200);
+    $actions = OpenQA::Test::Case::trim_whitespace($t->tx->res->dom->at('#user-action')->all_text);
+    like $actions, qr/API help ${mcp_help}Changelog Logout/, 'perci has operator links, case MCP disabled';
```
okurz:
```suggestion
    $t->get_ok('/tests')->status_is(200);
    $actions = OpenQA::Test::Case::trim_whitespace($t->tx->res->dom->at('#user-action')->all_text);
    like $actions, qr/API help Changelog Logout/, 'no MCP entry shown for operator when MCP not enabled';
```

t/ui/05-auth_mcp.t
```diff
@@ -0,0 +1,43 @@
...
+
+my ($actions, $mcp_help);
+
+subtest 'First access no login' => sub {
+    $t->get_ok('/tests')->status_is(200);
+    $actions = OpenQA::Test::Case::trim_whitespace($t->tx->res->dom->at('#user-action')->all_text);
+    is $actions, 'Login', 'no-one logged in';
+};
+
+subtest 'web UI operator user login, with MCP disabled' => sub {
```
okurz:
```suggestion
subtest 'MCP disabled' => sub {
```

as the former is not a proper English sentence and weird use of comma

t/ui/05-auth_mcp.t
```diff
@@ -0,0 +1,43 @@
...
+my $t = Test::Mojo->new('OpenQA::WebAPI');
+
+my ($actions, $mcp_help);
+
+subtest 'First access no login' => sub {
+    $t->get_ok('/tests')->status_is(200);
+    $actions = OpenQA::Test::Case::trim_whitespace($t->tx->res->dom->at('#user-action')->all_text);
+    is $actions, 'Login', 'no-one logged in';
+};
+
```
okurz, line 23:
```suggestion
```

I see no need for this subtest here

t/ui/05-auth_mcp.t
```diff
@@ -0,0 +1,43 @@
...
+use Test::Warnings ':report_warnings';
+use OpenQA::Test::TimeLimit '8';
+use OpenQA::Test::Case;
+
+my $test_case = OpenQA::Test::Case->new;
+$test_case->init_data(fixtures_glob => '03-users.pl');
+
+my $t = Test::Mojo->new('OpenQA::WebAPI');
+
+my ($actions, $mcp_help);
```
okurz: You are not really reusing the variables so I suggest to work with only local variables within each scope

```suggestion
```

t/ui/05-auth_mcp.t
```diff
@@ -0,0 +1,43 @@
+# Copyright SUSE LLC
+# SPDX-License-Identifier: GPL-2.0-or-later
+
+use Test::Most;
+
+use FindBin;
+use lib "$FindBin::Bin/../lib", "$FindBin::Bin/../../external/os-autoinst-common/lib";
+use Test::Mojo;
+use Test::Warnings ':report_warnings';
+use OpenQA::Test::TimeLimit '8';
```
okurz:
```suggestion
use Test::Mojo;
use Test::Warnings ':report_warnings';
use FindBin;
use lib "$FindBin::Bin/../lib", "$FindBin::Bin/../../external/os-autoinst-common/lib";
use OpenQA::Test::TimeLimit '8';
```

as we should only use the local lib lookup for modules that actually need it

t/ui/05-auth_mcp.t
```diff
@@ -0,0 +1,43 @@
...
+subtest 'web UI operator user login, with MCP disabled' => sub {
+    $test_case->login($t, 'percival');
+    $t->app->config->{global}{mcp_enabled} = 'no';
+    $mcp_help = '';
+    $t->get_ok('/tests')->status_is(200);
+    $actions = OpenQA::Test::Case::trim_whitespace($t->tx->res->dom->at('#user-action')->all_text);
+    like $actions, qr/API help ${mcp_help}Changelog Logout/, 'perci has operator links, case MCP disabled';
+};
+
+subtest 'web UI operator user after MCP enabled' => sub {
```
okurz:
```suggestion
subtest 'MCP enabled' => sub {
```

as "operator user" does not make much sense

t/ui/05-auth_mcp.t
m-dati: The next tests didn't work, without it. It resulted to be needed for the initialization of the web client.

t/ui/05-auth_mcp.t
m-dati: it was a mention, in the test title, of the user role, different than a normal unpriviledged user.
Anyway, I'll change it.

m-dati: it was a mention, in the test title, of the user role, different than a normal unpriviledged user

t/ui/05-auth_mcp.t
m-dati: ok, for this module; but then let's remember of it and do the same changes for all the other `t/*.t` tests, that have the same unwanted order.

t/ui/05-auth_mcp.t
m-dati: When I prepared this unit test, I already tried to remove it, but the full test didn't work anymore; keeping it in, instead, all works fine.
Therefore it is needed, to prepare the first access without login, for the next tests with a logged-in user.

m-dati: Please, find here the results _without_ and _with_ this subtest:
* Run t/ui/5 mcp, without the first test:
```
t/ui/05-auth_mcp.t 
...
    #   Failed test 'no MCP entry shown for operator when MCP not enabled'
    #   at t/ui/05-auth_mcp.t line 30.
    #                   'Login'
    #     doesn't match '(?^:API help Changelog Logout)'
t/ui/05-auth_mcp.t .. 1/? 
#   Failed test 'MCP disabled'
#   at t/ui/05-auth_mcp.t line 31.

    #   Failed test 'MCP entry is shown for operator when MCP is enabled'
    #   at t/ui/05-auth_mcp.t line 38.
    #                   'Login'
    #     doesn't match '(?^:API help MCP help Changelog Logout)'

#   Failed test 'MCP enabled'
#   at t/ui/05-auth_mcp.t line 39.
# Looks like you failed 2 tests of 3.
```
* Rerun the t/ui/5 test with it:
```
t/ui/05-auth_mcp.t 
t/ui/05-auth_mcp.t .. ok    
All tests successful.
```

t/ui/05-auth_mcp.t
okurz: Yes, I would appreciate if you can cleanup more. Unfortunately in Perl we are far behind with such stylistic rules to be enforced automatically.

t/ui/05-auth_mcp.t
okurz: ok, fine. keep it

t/ui/05-auth_mcp.t
m-dati: done

t/ui/05-auth_mcp.t
m-dati: done

t/ui/05-auth_mcp.t
m-dati: done

t/ui/05-auth_mcp.t
m-dati: done

t/ui/05-auth_mcp.t
m-dati: done

PR #6929, opened 2026-01-24
feat: Add symlink for aeon in openqa-bootstrap script

jwharm:
Related to PR https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/24395

I created a new product for aeon, but openQA by default will try to use `/var/lib/openqa/tests/aeon` but that doesn't exist; it's located under `/var/lib/openqa/tests/opensuse`. A symlink from `/var/lib/openqa/tests/aeon -> opensuse` fixes this. 

(There's already a similar link `/var/lib/openqa/tests/sle -> opensuse` for SLE, for the same reason.)

This PR extends the `openqa-bootstrap` script to create this symlink.

### DIFF
--- a/script/openqa-bootstrap
+++ b/script/openqa-bootstrap
@@ -149,6 +149,9 @@ if [[ -z $skip_suse_tests ]]; then
     if [[ ! -e /var/lib/openqa/tests/sle ]]; then
         ln -s opensuse /var/lib/openqa/tests/sle
     fi
+    if [[ ! -e /var/lib/openqa/tests/aeon ]]; then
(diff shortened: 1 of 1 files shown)


script/openqa-bootstrap
```diff
@@ -162,6 +162,11 @@ if [[ -z $skip_suse_tests ]]; then
     fi
 fi
 
+# link aeon tests to opensuse
```
okurz: Please put those into the if condition for SUSE specific tests above. The bootstrap script primarily targets any user of openQA so independent of openSUSE/SUSE

jwharm: Done

PR #6935, opened 2026-01-26
Add MCP tool annotations for Claude connector compliance

d3flex:
Enhance MCP tool definitions with comprehensive metadata to comply with Claude connector submission requirements and MCP safety principles. All tools are annotated as `read-only` and `non-destructive`. you can visit #security-and-trust-&-safety on
https://modelcontextprotocol.io/specification/2025-03-26 for reference.

related-issue: https://progress.opensuse.org/issues/193555

UPDATE: If you have `perl-MCP-0.70.0-1.1` you can test it out

### DIFF
--- a/cpanfile
+++ b/cpanfile
@@ -50,7 +50,7 @@ requires 'JSON::Validator';
 requires 'JavaScript::Minifier::XS', '>= 0.11';
 requires 'LWP::Protocol::https';
 requires 'LWP::UserAgent';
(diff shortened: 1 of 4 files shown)


d3flex:
> If this means we need a certain version of the MCP module we should probably be explicit about that and add a constraint in the dependency YAML and/or spec file.

@Martchus feel free to check again. I have added he version on the dependencies

PR #6938, opened 2026-01-26
Restart: handle subclassed AMQP plugin

AdamWill:
This code, introduced in #6816, assumed the plugin called AMQP was only ever used as itself. However, Fedora has a plugin called FedoraMessaging which subclasses AMQP, using its `register` sub unmodified and wrapping its `publish_amqp` sub to add some more fields. #6816 broke this setup because it causes openQA to load the AMQP plugin even when the configured plugin is actually FedoraMessaging, and publish messages without the extra fields, which confuses Fedora's messaging system.

This should fix the problem in a generic way (it'll work for any other similar subclassed message publisher plugin) by having `register` stash the plugin's name rather than just a boolean, then having the load code read in that name and load that plugin.

### DIFF
--- a/lib/OpenQA/WebAPI/Plugin/AMQP.pm
+++ b/lib/OpenQA/WebAPI/Plugin/AMQP.pm
@@ -29,7 +29,11 @@ sub register ($self, $app, @args) {
     $self->{app} = $app;
     $self->{config} = $app->config;
     my $config = $self->{config}->{amqp};
-    $config->{enabled} = 1;    # Needed for reloading the plugin later in the forked process
+    # this gives us the plugin's name; we need to know it to reload
+    # the correct plugin later in the forked process, and it may not
+    # be AMQP, it may be a subclass like FedoraMessaging
+    $config->{plugin} = (split /::/, ref $self)[-1];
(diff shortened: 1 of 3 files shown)


lib/OpenQA/WebAPI/Plugin/AMQP.pm
```diff
@@ -29,7 +29,11 @@ sub register ($self, $app, @args) {
     $self->{app} = $app;
     $self->{config} = $app->config;
     my $config = $self->{config}->{amqp};
-    $config->{enabled} = 1;    # Needed for reloading the plugin later in the forked process
+    # this gives us the plugin's name; we need to know it to reload
+    # the correct plugin later in the forked process, and it may not
+    # be AMQP, it may be a subclass like FedoraMessaging
+    $config->{plugin} = (split(/::/, ref($self)))[-1];
```
perlpunk:
```suggestion
    $config->{plugin} = (split /::/, ref $self)[-1];
```

AdamWill: sure, if you like. I never know what's preferred in perl...

perlpunk: less noise :D

PR #6940, opened 2026-01-27
Handle links on test_log for different url in .git/config

d3flex:
The investigation function creates links to commits using the URL from the .git/config file. This URL is expected to have a `.git` extension. However, manual intervention when cloning the repo can result in a URL without the extension, which breaks the constructed commit links.

This commit introduces an enhancement to handle both cases.

related-issue: https://progress.opensuse.org/issues/191914


d3flex:
> Dismissed my requested changes, I created
> 
> * [refactor(investigation): Use TEST_GIT_URL and NEEDLES_GIT_URL #6941](https://github.com/os-autoinst/openQA/pull/6941)
> 
> for the refactoring

I had started working on this request. So I found that https://github.com/os-autoinst/openQA/pull/6941 already merged. I will rebase. nice done @perlpunk

PR #6942, opened 2026-01-27
Update openQA containers to Leap 16.0

d3flex:
Those container are missing from the registry and as we have moved on
and noticed with other upgrades, we can also upgrade those ones.
    
- drop obsolete 'rubygem(sass)>=3.7.4' replacing with perl-CSS-Sass
    
This also required and intends to fix the helm charts in a followup
commit(s)
    

related-issue: https://progress.opensuse.org/issues/180731

### DIFF
--- a/container/webui/Dockerfile
+++ b/container/webui/Dockerfile
@@ -1,6 +1,6 @@
 # SPDX-License-Identifier: MIT
 #!BuildTag: openqa_webui:latest opensuse/openqa-webui:latest opensuse/openqa-webui:%PKG_VERSION% opensuse/openqa-webui:%PKG_VERSION%.%RELEASE%
-FROM opensuse/leap:15.6
+FROM opensuse/leap:16.0
 
 # labelprefix=org.opensuse.openqa-webui
 LABEL org.opencontainers.image.title="openQA webui container"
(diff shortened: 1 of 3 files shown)


d3flex:
> > Some changes:
> > * remove devel::openQA repos and use packages directly from leap:16.0
> 
> that does not seem to be true and also is not what you should do

yea , lets over from previous commit. We've already clarified with @perlpunk, that they should not removed

container/webui/Dockerfile
```diff
@@ -14,13 +14,10 @@ LABEL org.opencontainers.image.created="%BUILDTIME%"
 ENV LC_ALL=C.UTF-8
 
 # hadolint ignore=DL3037
-RUN zypper ar -p 95 -f http://download.opensuse.org/repositories/devel:openQA/15.6 devel_openQA && \
-    zypper ar -p 90 -f http://download.opensuse.org/repositories/devel:openQA:Leap:15.6/15.6 devel_openQA_Leap && \
+RUN zypper ar -p 95 -f http://download.opensuse.org/repositories/devel:openQA/16.0 devel_openQA && \
+    zypper ar -p 90 -f http://download.opensuse.org/repositories/devel:openQA:Leap:16.0/16.0 devel_openQA_Leap && \
     zypper --gpg-auto-import-keys ref && \
-    # openQA might need rubygem:sass to compile assets in this context due to
```
perlpunk, line 20: you replaced rubygem sass with perl-CSS-Sass below. why do you remove the comment instead of replacing just the package name here as well?

Martchus: Out of my head I'd say what this comment says is not true anymore. So it is probably good to remove it.

perlpunk: ok, good to know

PR #6947, opened 2026-01-28
Prepare documentation generation for Leap 16.0

perlpunk:

### DIFF
--- a/tools/generate-documentation
+++ b/tools/generate-documentation
@@ -200,15 +200,25 @@ check_asciidoctor() {
 install_asciidoctor() {
     # install dependencies
     gem install asciidoctor
-    gem install pygments.rb -v 2.4.1 # newer versions need Ruby 2.6
-    [[ ${formats[pdf]} ]] && {
-        gem install public_suffix -v 4.0.7 # newer versions need Ruby 2.6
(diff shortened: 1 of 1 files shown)


d3flex (approved):

tools/generate-documentation
```diff
@@ -200,15 +200,25 @@ check_asciidoctor() {
...
-        gem install public_suffix -v 4.0.7 # newer versions need Ruby 2.6
-        gem install css_parser -v 1.12.0   # newer versions need Ruby 2.7
-        gem install ttfunk -v 1.7.0
-        gem install Ascii85 -v 1.1.1         # newer versions need Ruby 2.7
-        gem install afm -v 0.2.2             # newer versions need Ruby 3.2.0
-        gem install asciidoctor-pdf -v 1.6.2 # newer versions need Ruby 2.7
-    }
+
+    source /etc/os-release
+    if [[ $VERSION == 15.6 ]]; then
```
okurz: That's very openSUSE specific and also specific to one version. Can we at least do

```suggestion
    if [[ $VERSION =~ 15 && $ID_LIKE =~ suse ]]; then
```

perlpunk: I now changed it to read the ruby version

okurz: much better, great

PR #6948, opened 2026-01-29
feat(ui): add "Reset Filters" button to test overview page

okurz:
The button allows users to quickly clear all applied filter selections in the
filter box on the /tests/overview page without completely removing filter parameters that identify a build
Screenshot_20260129_114142_openqa_filter_box

### DIFF
--- a/assets/javascripts/filter_form.js
+++ b/assets/javascripts/filter_form.js
@@ -26,6 +26,15 @@ function setupFilterForm(options) {
       );
     }
   });
+
+  $('#filter-reset-button').on('click', function () {
+    const form = $('#filter-form');
+    form.find('input[type="text"], input[type="number"]').val('');
+    form.find('input[type="checkbox"]').prop('checked', false);
+    form.find('input[hidden]').remove();
+    form.find('select').val([]).trigger('chosen:updated');
+    $('#filter-panel .card-header span').text('no filter present, click to toggle filter form');
+  });
 }
 
 function parseFilterArguments(paramHandler) {
--- a/t/ui/10-tests_overview.t
+++ b/t/ui/10-tests_overview.t
@@ -315,6 +315,16 @@ subtest 'filtering by flavor' => sub {
     }
 };
 
+subtest 'reset filters' => sub {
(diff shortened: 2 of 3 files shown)


perlpunk:
When clicking on the image I get: "This private-user-images.githubusercontent.com page can’t be found"

edit: nevermind, I just reloaded the page and it works now. I guess it was edited

assets/javascripts/filter_form.js
```diff
@@ -26,6 +26,15 @@ function setupFilterForm(options) {
...
     }
   });
+
+  $('#filter-reset-button').on('click', function () {
+    const form = $('#filter-form');
+    form.find('input[type="text"], input[type="number"]').val('');
+    form.find('input[type="checkbox"]').prop('checked', false);
+    form.find('input[hidden]').remove();
+    form.find('select').val([]).trigger('chosen:updated');
+    $('#filter-panel .card-header span').text('no filter present, click to toggle filter form');
```
Martchus, line 36: It would be nice if `applyFilterChanges` would set this text and if we would just all this function from here. (Now this text is duplicated between backend/frontend.) You can leave it like that if the deduplication is too much effort.

assets/javascripts/filter_form.js
```diff
@@ -26,6 +26,15 @@ function setupFilterForm(options) {
       );
     }
   });
+
+  $('#filter-reset-button').on('click', function () {
+    const form = $('#filter-form');
+    form.find('input[type="text"], input[type="number"]').val('');
+    form.find('input[type="checkbox"]').prop('checked', false);
+    form.find('input[hidden]').remove();
```
Martchus, line 34: Removing hidden inputs might break things. What values do these elements have here? Do we need to remove these?

okurz: yes, looks like this does not behave as most people would like the "reset filters" to work. I am not sure we can get this to work. I guess https://github.com/os-autoinst/openQA/pull/6956 is a better alternative that covers the most important use case for me, to clear the filter on the job state+result checkboxes

PR #6949, opened 2026-01-29
Update subrepo external/os-autoinst-common

perlpunk:
This includes deprecating the B::Deparse patch.

I didn't see any warnings in Leap 16.0 anymore.
So this was fixed in some perl version > 5.26.
We should remove the diag message here so we can upgrade without having unhandled output.
After having switched consistently we can delete it.


perlpunk:
See https://github.com/os-autoinst/os-autoinst-common/pull/61

PR #6950, opened 2026-01-29
Improve test coverage of controller code

Martchus:
Related ticket: https://progress.opensuse.org/issues/189204

### DIFF
--- a/codecov.yml
+++ b/codecov.yml
@@ -26,15 +26,7 @@ coverage:
           - lib/OpenQA/WebAPI/Controller/File.pm
           - lib/OpenQA/WebAPI/Controller/Step.pm
           - lib/OpenQA/WebAPI/Controller/Test.pm
-          - lib/OpenQA/WebAPI/Controller/API/V1/Iso.pm
-          - lib/OpenQA/WebAPI/Controller/API/V1/Job.pm
-          - lib/OpenQA/WebAPI/Controller/API/V1/JobGroup.pm
-          - lib/OpenQA/WebAPI/Controller/API/V1/JobTemplate.pm
-          - lib/OpenQA/WebAPI/Controller/API/V1/Locks.pm
-          - lib/OpenQA/WebAPI/Controller/API/V1/Mm.pm
-          - lib/OpenQA/WebAPI/Controller/API/V1/Search.pm
-          - lib/OpenQA/WebAPI/Controller/API/V1/Table.pm
-          - lib/OpenQA/WebAPI/Controller/API/V1/Worker.pm
+          - lib/OpenQA/WebAPI/Controller/API/V1/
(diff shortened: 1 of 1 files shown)


t/13-joblocks.t
```diff
@@ -187,7 +188,9 @@ $t->post_ok($b_prefix)->status_is(400);
 # create barrier without number of expected tasks fails
 $t->post_ok($b_prefix, form => {name => 'barrier1'})->status_is(400);
 # create barrier succeeds with 1 expected task
-$t->post_ok($b_prefix, form => {name => 'barrier1', tasks => 1},)->status_is(200);
+$t->post_ok($b_prefix, form => {name => 'barrier1', tasks => 1})->status_is(200);
+$t->post_ok($b_prefix, form => {name => 'barrier1', tasks => 1});
```
d3flex, line 192: I guess you can remove this line, right? `$t->post_ok($b_prefix, form => {name => 'barrier1', tasks => 1})->status_is(200);` should be cover it

Martchus: fixed

t/13-joblocks.t
```diff
@@ -187,7 +188,8 @@ $t->post_ok($b_prefix)->status_is(400);
 # create barrier without number of expected tasks fails
 $t->post_ok($b_prefix, form => {name => 'barrier1'})->status_is(400);
 # create barrier succeeds with 1 expected task
-$t->post_ok($b_prefix, form => {name => 'barrier1', tasks => 1},)->status_is(200);
+$t->post_ok($b_prefix, form => {name => 'barrier1', tasks => 1})->status_is(200);
+$t->status_is(409, 'return code in case of conflict (barrier)');
```
perlpunk, line 193: Now I wonder how that can be. `status_is(200)` above, and then it shall be 400 without another request?

Martchus: Oh, right. I think the code was actually correct before.

@d3flex The 2nd invocation was *not* already covered because it is the case when the barrier already exists (after the 1st invocation created it).

t/lib/OpenQA/Test/Case.pm
```diff
@@ -1,16 +1,19 @@
...
 
 require OpenQA::Test::Database;
+use Exporter 'import';
 use OpenQA::Test::Testresults;
 use OpenQA::Schema::Result::Users;
 use OpenQA::Schema;
 use Date::Format 'time2str';
 use Mojo::JSON 'decode_json';
 
+our @EXPORT = qw(find_most_recent_event);
```
perlpunk: I would prefer
```suggestion
our @EXPORT_OK = qw(find_most_recent_event);
```
so that it isn't exported by default.

PR #6951, opened 2026-01-29
Refactor t/lib/OpenQA/Test/Case.pm with signatures

d3flex:
Observed on review of ce11d42 that the specific module is missing signatures.

### DIFF
--- a/t/lib/OpenQA/Test/Case.pm
+++ b/t/lib/OpenQA/Test/Case.pm
@@ -2,7 +2,7 @@
 # SPDX-License-Identifier: GPL-2.0-or-later
 
 package OpenQA::Test::Case;
-use Mojo::Base -base;
+use Mojo::Base -base, -signatures;
 
 require OpenQA::Test::Database;
 use Exporter 'import';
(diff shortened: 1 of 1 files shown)


Martchus:
This broke tests: `Too many arguments for subroutine 'OpenQA::Test::Case::login' at t/ui/06-operator_links.t line 28.`

t/lib/OpenQA/Test/Case.pm
```diff
@@ -46,7 +39,7 @@ sub login {
     # Hack the existing session cookie and add a user to pretend we logged in
     $c->req->cookies($cookie);
     $sessions->load($c);
-    OpenQA::Schema->singleton->resultset('Users')->create_user($username);
+    OpenQA::Schema->singleton->resultset('Users')->create_user($username, %attrs);
```
Martchus, line 42:
```suggestion
    OpenQA::Schema->singleton->resultset('Users')->create_user($username, \%attrs);
```

But it would make more sense to already pass a hash-ref to `login`. Or to remove this parameter completely as it is seemingly not required.

PR #6952, opened 2026-01-29
Throttle openQA jobs resources based on test parameters size

m-dati:
Define a priority component for some resources to control, to be added to a job priority setting at test start time, to limit the jobs resources requests.

Store the configuration for all the needed resources in `openqa.ini` file, in `prio_throttling_parameters` parameter, in a string of a format described in the ticket: "param1:scale1[:reference1],param2:scale2,...", to load the control values at openQA server start.

- Ticket [poo#192952](https://progress.opensuse.org/issues/192952)

### DIFF
--- a/lib/OpenQA/Schema/ResultSet/Jobs.pm
+++ b/lib/OpenQA/Schema/ResultSet/Jobs.pm
@@ -146,35 +146,7 @@ sub create_from_settings ($self, $settings, $scheduled_product_id = undef) {
     # assign scheduled product
     $new_job_args{scheduled_product_id} = $scheduled_product_id;
 
-    my $debug_msg;
-    my $max_job_time = looks_like_number $settings{MAX_JOB_TIME} ? $settings{MAX_JOB_TIME} : 0;
-    my $timeout_scale = looks_like_number $settings{TIMEOUT_SCALE} ? $settings{TIMEOUT_SCALE} : 0;
-    $max_job_time *= $timeout_scale if $max_job_time and $timeout_scale > 1;
-    if ($max_job_time and $max_job_time > DEFAULT_MAX_JOB_TIME) {
-        if (my $scale = OpenQA::App->singleton->config->{misc_limits}->{max_job_time_prio_scale}) {
-            my $malus = int($max_job_time / $scale);
-            $debug_msg = sprintf 'Adding priority malus to newly created job (old: %d, malus: %s)',
-              $new_job_args{priority}, $malus;
-            $new_job_args{priority} += $malus;
-        }
-    }
-    # apply resources throttling control
-    if (my $throttling
-        = OpenQA::App->singleton && OpenQA::App->singleton->config->{misc_limits}->{prio_throttling_data})
-    {
-        my $throttling_info;
-        for my $resource (keys %$throttling) {
-            next if !defined $settings{$resource};
-            my $scale = $throttling->{$resource}->{scale};
-            my $reference = $throttling->{$resource}->{reference};
-            my $prio = int(($settings{$resource} - $reference) * $scale);
-            $throttling_info .= "$resource, scale: $scale" . ($reference ? ", reference: $reference;" : ';');
-            $new_job_args{priority} += $prio;
-        }
-        $debug_msg .= sprintf("\nAdjusting job priority by %s based on resource requirement(s): %s",
-            $new_job_args{priority}, $throttling_info)
-          if $throttling_info;
-    }
+    my $debug_msg = _apply_prio_throttling(\%settings, \%new_job_args);
 
     my $job = $self->create(\%new_job_args);
     log_debug(sprintf "(Job %d) $debug_msg", $job->id) if $debug_msg;
@@ -196,6 +168,42 @@ sub create_from_settings ($self, $settings, $scheduled_product_id = undef) {
     $job->calculate_blocked_by;
     return $job;
 }
+
+sub _apply_prio_throttling ($settings, $new_job_args) {
+    my $debug_msg;
+    my $max_job_time = looks_like_number $settings->{MAX_JOB_TIME} ? $settings->{MAX_JOB_TIME} : 0;
+    my $timeout_scale = looks_like_number $settings->{TIMEOUT_SCALE} ? $settings->{TIMEOUT_SCALE} : 0;
+    $max_job_time *= $timeout_scale if $max_job_time and $timeout_scale > 1;
+    if ($max_job_time and $max_job_time > DEFAULT_MAX_JOB_TIME) {
+        if (my $scale = OpenQA::App->singleton->config->{misc_limits}->{max_job_time_prio_scale}) {
+            my $malus = int($max_job_time / $scale);
+            $debug_msg = sprintf 'Adding priority malus to newly created job (old: %d, malus: %s)',
+              $new_job_args->{priority}, $malus;
+            $new_job_args->{priority} += $malus;
+        }
+    }
+
+    if (my $throttling
+        = OpenQA::App->singleton && OpenQA::App->singleton->config->{misc_limits}->{prio_throttling_data})
+    {
+        my $throttling_info;
+        for my $resource (keys %$throttling) {
+            next unless defined $settings->{$resource};
+            my $scale = $throttling->{$resource}->{scale};
+            my $reference = $throttling->{$resource}->{reference};
+            my $prio = int(($settings->{$resource} - $reference) * $scale);
+            $throttling_info .= "$resource, scale: $scale" . ($reference ? ", reference: $reference;" : ';');
+            $new_job_args->{priority} += $prio;
+        }
+        $debug_msg .= sprintf(
+            '. Adjusting job priority by %s based on resource requirement(s): %s',
+            $new_job_args->{priority},
+            $throttling_info
+        ) if $throttling_info;
+    }
+    return $debug_msg;
+}
+
 
 sub _handle_dependency_settings ($self, $settings, $new_job_args) {
     my $job_settings = $self->result_source->schema->resultset('JobSettings');
--- a/t/api/04-jobs.t
+++ b/t/api/04-jobs.t
@@ -1021,71 +1021,73 @@ subtest 'array settings correctly assigned when posting job' => sub {
 };
 
 subtest 'priority correctly assigned when posting job' => sub {
+    my $default_prio = 50;
+
     # post new job and check default priority
     $t->post_ok('/api/v1/jobs', form => \%jobs_post_params)->status_is(200);
     $t->get_ok('/api/v1/jobs/' . $t->tx->res->json->{id})->status_is(200);
     $t->json_is('/job/group', 'opensuse', 'group assigned (1)');
-    $t->json_is('/job/priority', 50, 'global default priority assigned');
+    $t->json_is('/job/priority', $default_prio, 'global default priority assigned');
 
     subtest 'priority malus due to high MAX_JOB_TIME' => sub {
-        my $max = 7300;
-        local $jobs_post_params{MAX_JOB_TIME} = $max;
+        my %new_job_args = (priority => $default_prio);
+        my $max_time = 7300;
+        local $jobs_post_params{MAX_JOB_TIME} = $max_time;
         $t->post_ok('/api/v1/jobs', form => \%jobs_post_params)->status_is(200);
         $t->get_ok('/api/v1/jobs/' . $t->tx->res->json->{id})->status_is(200);
-        $t->json_is('/job/priority', 50 + $max / 100, 'increased prio value');
+        $t->json_is('/job/priority', $default_prio + $max_time / 100, 'increased prio value');
 
         local $jobs_post_params{TIMEOUT_SCALE} = 2;
-        $t->post_ok('/api/v1/jobs', form => \%jobs_post_params)->status_is(200);
-        $t->get_ok('/api/v1/jobs/' . $t->tx->res->json->{id})->status_is(200);
-        $t->json_is('/job/priority', 50 + $max * 2 / 100, 'increased prio value with TIMEOUT_SCALE');
+        OpenQA::Schema::ResultSet::Jobs::_apply_prio_throttling(\%jobs_post_params, \%new_job_args);
+        is $new_job_args{priority}, $default_prio + $max_time * 2 / 100, 'increased prio value with TIMEOUT_SCALE';
         delete $jobs_post_params{TIMEOUT_SCALE};
 
         my $limits = OpenQA::App->singleton->config->{misc_limits};
+        %new_job_args = (priority => $default_prio);
         $limits->{max_job_time_prio_scale} = 10;
-        $t->post_ok('/api/v1/jobs', form => \%jobs_post_params)->status_is(200);
-        $t->get_ok('/api/v1/jobs/' . $t->tx->res->json->{id})->status_is(200);
-        $t->json_is('/job/priority', 50 + $max / 10, 'custom scale value: increased prio value');
+        OpenQA::Schema::ResultSet::Jobs::_apply_prio_throttling(\%jobs_post_params, \%new_job_args);
+        is $new_job_args{priority}, $default_prio + $max_time / 10, 'custom scale value: increased prio value';
 
+        %new_job_args = (priority => $default_prio);
         $limits->{max_job_time_prio_scale} = 0;
-        $t->post_ok('/api/v1/jobs', form => \%jobs_post_params)->status_is(200);
-        $t->get_ok('/api/v1/jobs/' . $t->tx->res->json->{id})->status_is(200);
-        $t->json_is('/job/priority', 50, 'feature disabled: prio value unchanged');
+        OpenQA::Schema::ResultSet::Jobs::_apply_prio_throttling(\%jobs_post_params, \%new_job_args);
+        is $new_job_args{priority}, $default_prio, 'feature disabled: prio value unchanged';
     };
 
     subtest 'priority scaled up due to QEMURAM demand' => sub {
+        my %new_job_args = (priority => $default_prio);
         my $add = 20;
         local $jobs_post_params{QEMURAM} = 4096;
 
         my $config = OpenQA::Setup::read_config($t->app);
         $config->{misc_limits}->{prio_throttling_parameters} = 'XXX :0.2, QEMURAM:0.01:2048';
         $config->{misc_limits}->{prio_throttling_data} = OpenQA::Setup::_load_prio_throttling($t->app, $config);
-        $t->post_ok('/api/v1/jobs', form => \%jobs_post_params)->status_is(200);
-        $t->get_ok('/api/v1/jobs/' . $t->tx->res->json->{id})->status_is(200);
-        $t->json_is('/job/priority', 50 + $add, 'increased prio value');
+        OpenQA::Schema::ResultSet::Jobs::_apply_prio_throttling(\%jobs_post_params, \%new_job_args);
+        is $new_job_args{priority}, $default_prio + $add, 'increased prio value';
     };
 
     subtest 'priority improved due to QEMURAM low demand' => sub {
+        my %new_job_args = (priority => $default_prio);
         my $add = -10;
         local $jobs_post_params{QEMURAM} = 1024;
 
         my $config = OpenQA::Setup::read_config($t->app);
         $config->{misc_limits}->{prio_throttling_parameters} = 'XXX :0.2, QEMURAM:0.01:2048';
         $config->{misc_limits}->{prio_throttling_data} = OpenQA::Setup::_load_prio_throttling($t->app, $config);
-        $t->post_ok('/api/v1/jobs', form => \%jobs_post_params)->status_is(200);
-        $t->get_ok('/api/v1/jobs/' . $t->tx->res->json->{id})->status_is(200);
-        $t->json_is('/job/priority', 50 + $add, 'decreased prio value');
+        OpenQA::Schema::ResultSet::Jobs::_apply_prio_throttling(\%jobs_post_params, \%new_job_args);
+        is $new_job_args{priority}, $default_prio + $add, 'decreased prio value';
     };
 
     subtest 'priority scaled up due to HDDSIZEGB demand' => sub {
+        my %new_job_args = (priority => $default_prio);
         my $add = 2;
         local $jobs_post_params{HDDSIZEGB} = 40;
         my $config = OpenQA::Setup::read_config($t->app);
         $config->{misc_limits}->{prio_throttling_parameters}
           = 'XXX :0.2,  FAKE_HDDSIZEGB:0.01, HDDSIZEGB:0.05,  YYY: 0.1';
         $config->{misc_limits}->{prio_throttling_data} = OpenQA::Setup::_load_prio_throttling($t->app, $config);
-        $t->post_ok('/api/v1/jobs', form => \%jobs_post_params)->status_is(200);
-        $t->get_ok('/api/v1/jobs/' . $t->tx->res->json->{id})->status_is(200);
-        $t->json_is('/job/priority', 50 + $add, 'increased prio value');
+        OpenQA::Schema::ResultSet::Jobs::_apply_prio_throttling(\%jobs_post_params, \%new_job_args);
+        is $new_job_args{priority}, $default_prio + $add, 'increased prio value';
     };
 
     # post new job in job group with customized default priority


t/21-needles.t
```diff
@@ -54,6 +55,23 @@ my $schema = OpenQA::Test::Database->new->create;
...
+    'scm git' => {checkout_needles_sha => 'yes', allow_arbitrary_url_fetch => 'yes'}};
+my $mock_app = Test::MockModule->new('OpenQA::App');
+my $mock_log = Test::MockModule->new('Mojo::Log');
+$mock_log->redefine('error', sub { print STDERR $_[1] . "\n" });
+$mock_log->redefine('info', sub { print STDOUT $_[1] . "\n" });
+$mock_log->redefine('debug', sub { print STDOUT $_[1] . "\n" });
+my $mock_singleton = bless {config => $mock_config}, 'OpenQA::App';
+# Add config and log as methods
+no strict 'refs';
+*{'OpenQA::App::config'} = sub { return $_[0]->{config} };
```
perlpunk: You probably want to use Test::MockObject here

t/21-needles.t
perlpunk, line 1: Did you look at the previous PRs, e.g. #6861 ?
This test doesn't seem to be a good place, I suggest to use the same test as the previous PRs for that, `t/api/04-jobs.t`

lib/OpenQA/Schema/ResultSet/Jobs.pm
```diff
@@ -101,6 +101,14 @@ sub latest_jobs ($self, $until = undef) {
     return @latest;
 }
 
+sub load_throttling_config ($config_string) {
+    # parse to hash "{ PAR1 => [THR1, MAL1], PAR2 => [THR2, MAL2], ...}"
+    # the prio_throttling_parameters configuration string; see openqa.ini
+    return unless ($config_string && $config_string =~ /=\d+:/);
+    my %hash = map { my ($k, $v) = split /=/, $_, 2; uc($k) => [split(/:/, $v)] } split /,/, $config_string;
```
perlpunk: From this I can't really see which format you actually chose. But I guess I will see when you add the first test :)
But I can't see a suggestion where it would be necessary to split on `=`.

t/21-needles.t
m-dati: In reality this test failed as soon I ran it locally, without any change. SO I had to work much to fix cascading issues coming after each change. In the end it passed, but still a strange git error remains.

etc/openqa/openqa.ini
```diff
@@ -369,6 +369,13 @@ concurrent = 0
 ## scheduled products without any jobs (e.g. due to errors) are not immediately
 ## cleaned up.
 #scheduled_product_min_storage_duration = 34
+## Job resources throttling configuration:
+## a string of comma-separated jobs resources, 
+## formatted like: "RES1=THR1:MAL1,RES2=THR2:MAL2,..."
+## where THR# is the maximum threshold of a RESOURCE# value,
+## and MAL# is a malus priority to multiply as many times as THR# increases:
+## PRIO1 = base + int(RESOURCE1/THR1)*MAL1
+#prio_throttling_parameters
```
okurz: I recommend to use the format from
https://progress.opensuse.org/issues/192952#Suggestions
to show a better example how this string will actually look like

lib/OpenQA/Schema/ResultSet/Jobs.pm
m-dati: A.t.m. the usage indication mainly is in openqa.ini, at file end. I will add more tests and helping text in next commits.

t/21-needles.t
```diff
@@ -54,6 +55,23 @@ my $schema = OpenQA::Test::Database->new->create;
...
+
+my $mock_config = {
+    misc_limits => {prio_throttling_parameters => '', max_job_time_prio_scale => 100},
+    global => {worker_timeout => 900},
+    'scm git' => {checkout_needles_sha => 'yes', allow_arbitrary_url_fetch => 'yes'}};
+my $mock_app = Test::MockModule->new('OpenQA::App');
+my $mock_log = Test::MockModule->new('Mojo::Log');
+$mock_log->redefine('error', sub { print STDERR $_[1] . "\n" });
+$mock_log->redefine('info', sub { print STDOUT $_[1] . "\n" });
+$mock_log->redefine('debug', sub { print STDOUT $_[1] . "\n" });
```
Martchus: Use signatures in new code.

t/21-needles.t
```diff
@@ -54,6 +55,23 @@ my $schema = OpenQA::Test::Database->new->create;
...
+my $mock_singleton = bless {config => $mock_config}, 'OpenQA::App';
+# Add config and log as methods
+no strict 'refs';
+*{'OpenQA::App::config'} = sub { return $_[0]->{config} };
+*{'OpenQA::App::log'} = sub { return bless {}, 'Mojo::Log' };
+use strict 'refs';
```
Martchus: Why are you not just using MockModule here as above?

lib/OpenQA/Setup.pm
```diff
@@ -293,6 +294,12 @@ sub read_config ($app) {
     if (my $minion_fail_job_blocklist = $config->{influxdb}->{ignored_failed_minion_jobs}) {
         $config->{influxdb}->{ignored_failed_minion_jobs} = [split(/\s+/, $minion_fail_job_blocklist)];
     }
+    my $throttling = $config->{misc_limits}->{prio_throttling_parameters};
+    if ($throttling && length $throttling) {
+        $config->{misc_limits}->{prio_throttling_parameters} =~ s/\s+//g;
+        die("Wrong formatting for 'prio_throttling_parameters' in openqa.ini")
+          unless ($throttling =~ /^[A-Z_]+=\d+:\d+(?:,[A-Z_]+=\d+:\d+)*$/i);
+    }
```
Martchus: This should probably be moved into its own function.

lib/OpenQA/Schema/ResultSet/Jobs.pm
```diff
@@ -158,6 +166,21 @@ sub create_from_settings ($self, $settings, $scheduled_product_id = undef) {
...
+            next if !defined $settings{$resource};
+            my @limits = $throttling->{$resource};
+            if ($settings{$resource} > $limits[0]) {
+                my $malus = int($settings{$resource} / $limits[0]) * $limits[1];
+                $debug_msg = sprintf 'Adding priority malus to newly created job (old: %d, malus: %s)',
+                  $new_job_args{priority}, $malus;
+                $new_job_args{priority} += $malus;
+            }
+        }
+    }
```
Martchus: This should be moved into its own function.

etc/openqa/openqa.ini
```diff
@@ -369,6 +369,13 @@ concurrent = 0
 ## scheduled products without any jobs (e.g. due to errors) are not immediately
 ## cleaned up.
 #scheduled_product_min_storage_duration = 34
+## Job resources throttling configuration:
+## a string of comma-separated jobs resources, 
+## formatted like: "RES1=THR1:MAL1,RES2=THR2:MAL2,..."
+## where THR# is the maximum threshold of a RESOURCE# value,
+## and MAL# is a malus priority to multiply as many times as THR# increases:
+## PRIO1 = base + int(RESOURCE1/THR1)*MAL1
```
Martchus: Maybe we should just refer to the documentation instead of explaining the syntax here.

etc/openqa/openqa.ini
```diff
@@ -369,6 +369,13 @@ concurrent = 0
 ## scheduled products without any jobs (e.g. due to errors) are not immediately
 ## cleaned up.
 #scheduled_product_min_storage_duration = 34
+## Job resources throttling configuration:
+## a string of comma-separated jobs resources, 
+## formatted like: "RES1=THR1:MAL1,RES2=THR2:MAL2,..."
```
Martchus: This syntax is not very intuitive. Just a suggestion: If it becomes to cumbersome using INI we can always just add a setting like `prio_throttling_parameters_file` that can point to a JSON or YAML document.

perlpunk: I also don't get what `RES` and `THR` stand for. for `MAL` it's clear (the value for the prio adjustment).

perlpunk: The ticket suggests:
```
[throttling]
prio_throttling_parameters = QEMURAM,QEMUCPUS,HDDSIZEGB,RETRY
prio_throttling_scale_<lowercase parameter_name> = value

# or
prio_throttling_parameters = QEMURAM:.003,QEMUCPUS:1,HDDSIZEGB:.5,RETRY:1
```

t/21-needles.t
m-dati: After a clarification, this change was reverted, pls see https://github.com/os-autoinst/openQA/pull/6952#discussion_r2747663013 and last commit

t/21-needles.t
m-dati: After a clarification, this change was reverted and removed

lib/OpenQA/Schema/ResultSet/Jobs.pm
m-dati: It is a settings control block similar (sibling) to the previous one above of Tina for `max_job_time`: therefore, to apply the logic, for code uniformity, I should do either of
- other than mine, move Tina's too in another function, but it's out of scope here,
- let here be unchanged, but do the movement for both in another PR,
- don't do that;
what is the best option ?

etc/openqa/openqa.ini
m-dati: A documentation of this feature is not present yet, to be pointed.
But when the code will be more consolidated, indeed somewhere a description should be added, then, we'll think how to arrange the refrences. For now, this is a draft reference likewise for the other  parameters present in this .ini file.

etc/openqa/openqa.ini
m-dati: All This was changed, after the help-session with Tina, and the ticket's main suggestion applied. See https://github.com/os-autoinst/openQA/pull/6952#discussion_r2747663013 and last commit.

But, for the record, I was proposing another way to achieve the throttling, now no more the case to explain it here.

t/21-needles.t
m-dati: Change reverted, no more present. A simple line movement in an upper position, resolved the initial tests issues. Thanks to Tina.

t/21-needles.t
m-dati: pls see https://github.com/os-autoinst/openQA/pull/6952#discussion_r2747663013

lib/OpenQA/Schema/ResultSet/Jobs.pm
m-dati: Code changed, please look at the new version.

etc/openqa/openqa.ini
m-dati: I will change it in next updates.

perlpunk:
it seems that https://app.circleci.com/pipelines/github/os-autoinst/openQA/19199/workflows/5a5b04f1-4a01-4606-8df0-4208907ae0b1/jobs/190404
```
Can't call method "config" on an undefined value at /home/squamata/project/lib/OpenQA/Schema/ResultSet/Jobs.pm line 170.
[19:01:53] t/ui/10-tests_overview.t ................... 
Dubious, test returned 255 (wstat 65280, 0xff00)
No subtests run 
```
is the same problem as the one we fixed in 21-needles.t.
In this case, just put a new `Test::Mojo` app before `prepare_database`:
```
prepare_database;
```
and add it to the same commit as the 21-needles.t fix

lib/OpenQA/Schema/ResultSet/Jobs.pm
m-dati: I've tried to move the code in a separated function, but  it resulted not much convenient, because the code is used once only and because many hashes should be passed in input, then also losing the symmetry with the similar previous max_job_time block.
I suggest to leave the code here, in its place.

etc/openqa/openqa.ini
m-dati: Moreover, looking at the default `openqa.ini`, each configurable parameter does have a synthetical description of the purpose and usage. So this new one also requires a dedicated description block.

lib/OpenQA/Setup.pm
m-dati: done

etc/openqa/openqa.ini
m-dati: done

lib/OpenQA/Setup.pm
```diff
@@ -61,6 +61,15 @@ sub _load_config ($app, $defaults, $mode_specific_defaults) {
     return $config;
 }
 
+sub validate_prio_throttling_format ($config) {
```
perlpunk: It's a convention in perl to prefix internal functions/methods with an underscore:
```suggestion
sub _validate_prio_throttling_format ($config) {
```

lib/OpenQA/Setup.pm
```diff
@@ -61,6 +61,15 @@ sub _load_config ($app, $defaults, $mode_specific_defaults) {
...
+    my $throttling = $config->{misc_limits}->{prio_throttling_parameters};
+    if ($throttling && length $throttling) {
```
perlpunk: The check for length is unnecessary here IMHO
```suggestion
    if ($throttling) {
```

lib/OpenQA/Setup.pm
```diff
@@ -61,6 +61,15 @@ sub _load_config ($app, $defaults, $mode_specific_defaults) {
...
+        $config->{misc_limits}->{prio_throttling_parameters} =~ s/\s+//g;
+        die("Wrong formatting for 'prio_throttling_parameters' in openqa.ini")
```
perlpunk:
```suggestion
        die "Wrong format for 'prio_throttling_parameters' in openqa.ini"
```
But we usually don't die in other places where the config has invalid values, but we issue a warning with `$app->log->warn`. So I suggest to do that here as well.

lib/OpenQA/Setup.pm
```diff
@@ -61,6 +61,15 @@ sub _load_config ($app, $defaults, $mode_specific_defaults) {
...
+          unless ($throttling =~ /^[A-Z_]+:\d+(?:,[A-Z_]+:\d+)*$/i);
```
perlpunk: That doesn't allow decimal numbers.

lib/OpenQA/Schema/ResultSet/Jobs.pm
```diff
@@ -158,6 +166,18 @@ sub create_from_settings ($self, $settings, $scheduled_product_id = undef) {
             $new_job_args{priority} += $malus;
         }
     }
+    my $throttling = OpenQA::App->singleton->config->{misc_limits}->{prio_throttling_parameters};
+    $throttling = load_throttling_config($throttling);
+    # apply resources throttling control
+    if ($throttling) {
```
```suggestion
    if (my $throttling = load_throttling_config(OpenQA::App->singleton->config->{misc_limits}->{prio_throttling_parameters}) {
```

lib/OpenQA/Schema/ResultSet/Jobs.pm
```diff
@@ -158,6 +166,18 @@ sub create_from_settings ($self, $settings, $scheduled_product_id = undef) {
...
+        for my $resource (keys %$throttling) {
+            next if !defined $settings{$resource};
+            my $malus = int($settings{$resource} * $throttling->{$resource});
```
```suggestion
            my $summand = int($settings{$resource} * $throttling->{$resource});
```

lib/OpenQA/Schema/ResultSet/Jobs.pm
```diff
@@ -158,6 +166,18 @@ sub create_from_settings ($self, $settings, $scheduled_product_id = undef) {
...
+            $debug_msg = sprintf 'Adding priority malus to newly created job due to %s (old: %d, malus: %s)',
```
```suggestion
            $debug_msg = sprintf "Adjusting job priority by %s based on resource requirement '%s' (old: %d)",
```

okurz (changes requested):

lib/OpenQA/Schema/ResultSet/Jobs.pm
perlpunk: I think the two blocks manipulating the priority can be moved to one separate function, but maybe we can do that together in a separate PR.

m-dati:
Going to apply the last discussed throttling logic, that includes a reference value, as described in [ticket note-10](https://progress.opensuse.org/issues/192952#note-10)

lib/OpenQA/Schema/ResultSet/Jobs.pm
m-dati: In the next commit I'll check `if ($throttling ref eq 'HASH') ...`, so to me it seems not much clear a local definition with that check:
```
if (my $throttling = OpenQA::App->singleton->config->{misc_limits}->{prio_throttling_parameters} ref eq 'HASH') {...}
```

lib/OpenQA/Setup.pm
```diff
@@ -61,6 +61,31 @@ sub _load_config ($app, $defaults, $mode_specific_defaults) {
...
+sub _load_prio_throttling($config) {
+    my $throttling = $config->{misc_limits}->{prio_throttling_parameters};
+    my $ret;
+    if ($throttling) {
+        my $n = qr/[+-]?(?:\d+\.?\d*|\.\d+)/;    # number
+        my $u = qr/([A-Z_]+):($n)(?::($n))?/i;    # Key:N:M = $1:$2:$3
+
+        # remove spaces and validate comma-separated keys sequence
+        $throttling =~ s/\s+//g;
+        die("Wrong format in openqa.ini 'prio_throttling_parameters': $throttling")
```
perlpunk: Please see my previous comment regarding `die` and `warn`.
https://github.com/os-autoinst/openQA/pull/6952#pullrequestreview-3744149564

lib/OpenQA/Setup.pm
```diff
@@ -61,6 +61,31 @@ sub _load_config ($app, $defaults, $mode_specific_defaults) {
...
+        # Implicit iteration along the sequence
+        while ($throttling =~ /,?$u/g) {
+            $hash{$1} = {scale => $2, reference => $3 // 0};
+        }
+        $ret = \%hash;
+    }
+    else {
+        $ret = undef;
+    }
+    $config->{misc_limits}->{prio_throttling_parameters} = $ret;
```
perlpunk: You don't need to set something to undef which is already undef:
```suggestion
        $config->{misc_limits}->{prio_throttling_parameters} = \%hash;
    }
```

lib/OpenQA/Schema/ResultSet/Jobs.pm
```diff
@@ -158,6 +158,23 @@ sub create_from_settings ($self, $settings, $scheduled_product_id = undef) {
             $new_job_args{priority} += $malus;
         }
     }
+    my $throttling = OpenQA::App->singleton->config->{misc_limits}->{prio_throttling_parameters};
+    # apply resources throttling control
+    if ($throttling ref eq 'HASH') {
```
perlpunk: This is a syntax error.
And you don't need to check for the type, if the function sets it to a hashref or leaves it undef.
You only need to check if it returns something true (only undef, 0 / `false` and the empty string are 'false" in perl).
```suggestion
    if ($throttling) {
```
If we extend to function to return an array ref at some point as an alternative, we can still add that check.

lib/OpenQA/Schema/ResultSet/Jobs.pm
m-dati: done

lib/OpenQA/Schema/ResultSet/Jobs.pm
m-dati: names changed

lib/OpenQA/Schema/ResultSet/Jobs.pm
m-dati: code updated

lib/OpenQA/Setup.pm
m-dati: changed

lib/OpenQA/Schema/ResultSet/Jobs.pm
m-dati: done

lib/OpenQA/Setup.pm
m-dati: changed

lib/OpenQA/Setup.pm
m-dati: done

lib/OpenQA/Setup.pm
m-dati: done

lib/OpenQA/Setup.pm
m-dati: done

lib/OpenQA/Setup.pm
```diff
@@ -61,6 +61,26 @@ sub _load_config ($app, $defaults, $mode_specific_defaults) {
     return $config;
 }
 
+sub _load_prio_throttling($config) {
+    my $throttling = $config->{misc_limits}->{prio_throttling_parameters};
+    my %hash;
+    if ($throttling) {
```
perlpunk: In this case, you can actually turn around the logic:
```suggestion
    return undef unless $throttling;
```
because in this case there is nothing to be done. One level less of indentation / complexity

perlpunk: or
```
return undef unless my $throttling = $config->{misc_limits}->{prio_throttling_parameters};
```

lib/OpenQA/Setup.pm
```diff
@@ -61,6 +61,26 @@ sub _load_config ($app, $defaults, $mode_specific_defaults) {
...
+        my $n = qr/[+-]?(?:\d+\.?\d*|\.\d+)/;    # number
+        my $u = qr/([A-Z_]+):($n)(?::($n))?/i;    # Key:N:M = $1:$2:$3
+
+        # remove spaces and validate comma-separated keys sequence
```
perlpunk: IMHO the comment is just stating the obvious and can be removed

lib/OpenQA/Setup.pm
```diff
@@ -61,6 +61,26 @@ sub _load_config ($app, $defaults, $mode_specific_defaults) {
...
+        $throttling =~ s/\s+//g;
+        $app->log->warn("Wrong format in openqa.ini 'prio_throttling_parameters': $throttling")
+          unless ($throttling =~ qr/^$u(?:,$u)*$/i);
+
+        # Implicit iteration along the sequence
```
perlpunk: I think this comment can also be removed.

etc/openqa/openqa.ini
```diff
@@ -369,6 +369,14 @@ concurrent = 0
 ## scheduled products without any jobs (e.g. due to errors) are not immediately
 ## cleaned up.
 #scheduled_product_min_storage_duration = 34
+## Throttling
+## A comma-separated list of test parameters eachone followed by a colon and a scaling factor,
```
d3flex:
```suggestion
## A comma-separated list of test parameters, each one followed by a colon and a scaling factor,
```

lib/OpenQA/Setup.pm
```diff
@@ -61,6 +61,26 @@ sub _load_config ($app, $defaults, $mode_specific_defaults) {
...
+        while ($throttling =~ /,?$u/g) {
+            $hash{$1} = {scale => $2, reference => $3 // 0};
+        }
+    }
+    $config->{misc_limits}->{prio_throttling_parameters} = \%hash;
```
perlpunk: I think you applied my suggestion in the wrong way.
Here `$config->{misc_limits}->{prio_throttling_parameters}` is always set, but I suggested to do it in the `if`.
Or better, like my other suggestion:
```
return undef unless my $throttling = $config->{misc_limits}->{prio_throttling_parameters};
```
at the beginning.

That should fix the config.t test

lib/OpenQA/Setup.pm
m-dati: Thanks, I applied this solution on my local code and checking the unit test locally.

lib/OpenQA/Setup.pm
m-dati: yes, thanks, I already did it on my local instance and re-checking the unit tests, before pushing commit

etc/openqa/openqa.ini
Martchus: > A documentation of this feature is not present yet, to be pointed.

Of course. I meant you should put what you're adding here into the documentation and just refer to that in the config file. It is just that 6 lines of comments with explanations is a bit much for the config file.

t/api/04-jobs.t
```diff
@@ -1052,6 +1052,26 @@ subtest 'priority correctly assigned when posting job' => sub {
         $t->json_is('/job/priority', 50, 'feature disabled: prio value unchanged');
     };
 
+    subtest 'priority malus due to high QEMURAM demand' => sub {
+        my $add = 40;
+        local $jobs_post_params{QEMURAM} = 4096;
+        my $limits = OpenQA::App->singleton->config->{misc_limits};
+        $limits->{prio_throttling_parameters} = 'QEMURAM:0.01';
+        $t->post_ok('/api/v1/jobs', form => \%jobs_post_params)->status_is(200);
```
perlpunk: https://app.circleci.com/pipelines/github/os-autoinst/openQA/19241/workflows/7bc258dc-0b54-4a1b-923b-fcfc71fcb4d5/jobs/190887

If you put a
```
always_explain $t->tx->res->json;
```
after that, you will see
```
        # {                                                    
        #   'error' => 'Can\'t use string ("QEMURAM:0.01") as a HASH ref while "strict refs" in use at /path/to/openQA/lib/OpenQA/Schema/ResultSet/Jobs.pm line 164.',
        #   'error_status' => 400                                                                                             
        # }                                                                                                                   
```
The problem is now that we turn the config string into a hash in Setup.
And for the test we set it to a string, but at that point it is already expected to be the processed hash ref.
I think we have other places where we preprocess the config, but I would have to look how we usually do this.
I think the original string should not be overwritten, but we could store the processed hash ref into a different place.

m-dati: I'd add a new element to the config hash, like i.e.: 
`config->...->{throttling_hash}`, dynamically, when config string isn't empty, to pass to priority elaboration in Jobs.

lib/OpenQA/Setup.pm
m-dati: done

etc/openqa/openqa.ini
m-dati: done

lib/OpenQA/Setup.pm
m-dati: done

lib/OpenQA/Setup.pm
m-dati: done

lib/OpenQA/Setup.pm
m-dati: done

t/api/04-jobs.t
m-dati: I added a new element: `$config->{misc_limits}->{prio_throttling_data}`
Does it make sense?

perlpunk: Yeah, let's go with it for now.
But we still need to set it in the test now to make it work.

m-dati: Indeed now the new `t/api/03-auth.t` test is failing in L1062 🙁

m-dati: I managed to fix the t/* and more. Pls, see last commit

etc/openqa/openqa.ini
m-dati: I'll update the docs.: any suggestion what is the better place to add this part?
Anyway I'd suggest to consider this work for a new and dedicated PR: I think that such topic shall be agreed in a broader way, if we go touching the central openQA documentation.

lib/OpenQA/Schema/ResultSet/Jobs.pm
m-dati: Ok, then a new function for both codes will be managed in antoher PR/

m-dati:
CI Just passed fine, including new and updated u.tests.

lib/OpenQA/Setup.pm
```diff
@@ -54,13 +54,29 @@ sub _load_config ($app, $defaults, $mode_specific_defaults) {
...
     return $config;
 }
 
+sub _load_prio_throttling ($app, $config) {
+    my $throttling = $config->{misc_limits}->{prio_throttling_parameters};
+    return undef unless $throttling;
+    my %hash;
+    my $n = qr/[+-]?(?:\d+\.?\d*|\.\d+)/;    # number
+    my $u = qr/([A-Z_][A-Z0-9_]*):($n)(?::($n))?/i;    # Key:N:M = $1:$2:$3
+    my $v = qr/^$u(?:,$u)*$/i;
```
m-dati: To be deleted Line 69, having used the expr. directly below.

etc/openqa/openqa.ini
m-dati: I updated the comments in .ini. About the documentation as discussed today, The feature could still change soon, so it's better we don't touch documentation yet 🙂

etc/openqa/openqa.ini
```diff
@@ -369,6 +369,11 @@ concurrent = 0
 ## scheduled products without any jobs (e.g. due to errors) are not immediately
 ## cleaned up.
 #scheduled_product_min_storage_duration = 34
+## Throttling: priority scaling for job-tests requesting high resources in the settings.
+## List configurable here in the format: PAR1:SCAL1[:REF1][,…,PAR#:SCAL#[:REF#]
+## PAR# parameter name,SCAL# decimal num.,REF# (optional) param.threshold before scaling.
+## The formula is: job-prio = 'base-prio + int((<PAR#.val> - <REF#>) * <SCAL#>) + ...'.
+#prio_throttling_parameters =
```
okurz: What was wrong with the next from

https://progress.opensuse.org/issues/192952#Suggestions ? I can't parse the sentence "priority scaling for job-tests requesting high resources in the settings." There is no verb included.

why did you not want to create a section `[throttling]` and decided to put everything into one setting instead of multiple ones?

The original suggestion incorporating the "reference" would be

```suggestion
[throttling]
## Set a comma-separated list of test parameters that should trigger throttling of scheduled openQA jobs by priority. Example: 'prio_throttling_parameters = QEMURAM,QEMUCPUS,HDDSIZEGB,RETRY'. Empty by default
#prio_throttling_parameters =
## Set 'prio_throttling_scale_<lowercase parameter_name>' to a divisor that the parameter
## value $value minus optional $reference is divided by and added to each job prio as
## '$base_prio + ($value - $reference)*$prio_throttling_scale'
## Example: 'prio_throttling_scale_retry = 3' for a base prio of 50 and a parameter value RETRY=1 leading to a job prio of '50 + 1*3=53'
## Example including reference: `prio_throttling_scale_qemuram = .003:1024` with a
## value of 2048 would lead to a priority of 50 + (2048-1024) * .003 = 53
#prio_throttling_scale_<lowercase parameter_name> =
```

but as we discussed we are ok to go with the alternative. We should just improve the comment above.

etc/openqa/openqa.ini
```diff
@@ -369,6 +369,11 @@ concurrent = 0
 ## scheduled products without any jobs (e.g. due to errors) are not immediately
 ## cleaned up.
 #scheduled_product_min_storage_duration = 34
+## Throttling: priority scaling for job-tests requesting high resources in the settings.
+## List configurable here in the format: PAR1:SCAL1[:REF1][,…,PAR#:SCAL#[:REF#]
+## PAR# parameter name,SCAL# decimal num.,REF# (optional) param.threshold before scaling.
+## The formula is: job-prio = 'base-prio + int((<PAR#.val> - <REF#>) * <SCAL#>) + ...'.
```
Martchus: I would avoid abbreviations where just one more letter would be required. I would also use a bit less punctuation.

```suggestion
## List configurable here in the format: PARAM1:SCALE1[:REF1][,…,PARAM#:SCALE#[:REF#]
## PARAM: parameter name, SCALE: decimal number, REF (optional): param threshold before scaling
## formula: job-prio = base-prio + int((<PARAM#.val> - <REF#>) * <SCALE#>) + …
```

lib/OpenQA/Schema/ResultSet/Jobs.pm
```diff
@@ -158,6 +158,21 @@ sub create_from_settings ($self, $settings, $scheduled_product_id = undef) {
             $new_job_args{priority} += $malus;
         }
     }
+    # apply resources throttling control
+    if (my $throttling = OpenQA::App->singleton->config->{misc_limits}->{prio_throttling_data}) {
+        my $debug_msg_2;
```
```suggestion
        my $throttling_info;
```

lib/OpenQA/Setup.pm
```diff
@@ -54,13 +54,28 @@ sub _load_config ($app, $defaults, $mode_specific_defaults) {
...
-
     # ensure default values are assigned; warn if config files were supplied at all
     _read_config_file($config, undef, $defaults, $mode_defaults);
     $app->log->warn('No configuration files supplied, will fallback to default configuration') unless $config_file;
     return $config;
 }
 
+sub _load_prio_throttling ($app, $config) {
+    my $throttling = $config->{misc_limits}->{prio_throttling_parameters};
+    return undef unless $throttling;
```
Martchus:
```suggestion
    return undef unless my $throttling = $config->{misc_limits}->{prio_throttling_parameters};
```

m-dati: In that proposed compressed form we mix all together, defining the variable and its initialization in a collateral block, out of the main flow, even if the variable scope is in the main block, because in perl it is allowed to do that.

I preferred the splitted form, for a more clear style: keep the definition evident in the main flow for the next use in the code body, making the return-condition more clear and clean.

etc/openqa/openqa.ini
m-dati: Thanks. As discussed, I considered that _alternative_ an option available and being that one easier and faster to implement, other than soon visible and quickly configurable in one shot, I opted for that implementation.
But in a next step we could even add the other logic too, as another way around to do the job.

lib/OpenQA/Setup.pm
Martchus: We use this combined form all over the place in our code base and it became a well-recognized pattern. You don't have to follow my suggestion here but I don't think we want to avoid the combined form.

etc/openqa/openqa.ini
Martchus: I said it should be more compact because we shouldn't have a whole paragraph of text in our config file for a single setting. I still think we shouldn't. However, if you want to merge this nevertheless, go ahead. I hope we'll improve this at some point. I don't understand what's so hard about putting this in the documentation in the first place, though.

perlpunk: We said that we want to wait with putting it into the official documentation because it's still experimental and the format / implementation might change

m-dati: I'll update the .ini comments in the next commit.

lib/OpenQA/Setup.pm
m-dati: If this is a team common style, then I'll change the code.

m-dati: Done

lib/OpenQA/Schema/ResultSet/Jobs.pm
m-dati: done

etc/openqa/openqa.ini
m-dati: done

etc/openqa/openqa.ini
m-dati: Done

okurz: I don't see the changes. There is still text like "job-tests" with no verb in the sentence

lib/OpenQA/Schema/ResultSet/Jobs.pm
```diff
@@ -158,6 +158,21 @@ sub create_from_settings ($self, $settings, $scheduled_product_id = undef) {
...
+            my $scale = $throttling->{$resource}->{scale};
+            my $reference = $throttling->{$resource}->{reference};
+            my $prio = int(($settings{$resource} - $reference) * $scale);
+            $throttling_info .= $resource . ", scale: $scale" . ($reference ? ", reference: $reference;" : ';');
+            $new_job_args{priority} += $prio;
+        }
+        $debug_msg .= sprintf("\nAdjusting job priority by %s based on resource requirement(s):\n %s",
+            $new_job_args{priority}, $throttling_info)
+          if $throttling_info;
+    }
```
okurz: If you need a comment explaining this block of code that means that this code wants to be a function :)

lib/OpenQA/Schema/ResultSet/Jobs.pm
```diff
@@ -158,6 +158,21 @@ sub create_from_settings ($self, $settings, $scheduled_product_id = undef) {
...
+        my $throttling_info;
+        for my $resource (keys %$throttling) {
+            next if !defined $settings{$resource};
+            my $scale = $throttling->{$resource}->{scale};
+            my $reference = $throttling->{$resource}->{reference};
+            my $prio = int(($settings{$resource} - $reference) * $scale);
+            $throttling_info .= $resource . ", scale: $scale" . ($reference ? ", reference: $reference;" : ';');
+            $new_job_args{priority} += $prio;
+        }
+        $debug_msg .= sprintf("\nAdjusting job priority by %s based on resource requirement(s):\n %s",
```
okurz: Can you give us an example of how a log message including that will look like?

lib/OpenQA/Setup.pm
```diff
@@ -54,13 +54,30 @@ sub _load_config ($app, $defaults, $mode_specific_defaults) {
...
+    my $u = qr/([A-Z_][A-Z0-9_]*):($n)(?::($n))?/i;    # Key:N:M = $1:$2:$3
+    $throttling =~ s/\s+//g;
+    unless ($throttling =~ qr/^$u(?:,$u)*$/i) {
+        $app->log->warn("Wrong format in openqa.ini 'prio_throttling_parameters': $throttling");
+        return;
+    }
+    while ($throttling =~ /,?$u/g) {
+        $hash{$1} = {scale => $2, reference => $3 // 0};
+    }
+    return \%hash;
```
okurz: I don't quite understand

```
my $n = qr/[+-]?(?:\d+\.?\d*|\.\d+)/;    # number
```

If it would be a number you could just use `[0-9]\+`. What is this actually?

And overall I suggest to construct a hash with map, not procedurally, e.g.

```suggestion
    my $n = qr/[+-]?(?:\d+\.?\d*|\.\d+)/;
    my $u = qr/([A-Z_][A-Z0-9_]*):($n)(?::($n))?/i;    # Key:N:M = $1:$2:$3
    $throttling =~ s/\s+//g;
    unless ($throttling =~ qr/^$u(?:,$u)*$/i) {
        $app->log->warn("Wrong format in openqa.ini 'prio_throttling_parameters': $throttling");
        return;
    }
    return { map { $_->[0] => { scale => $_->[1], reference => $_->[2] // 0 } } $throttling =~ /,?$u/g };
```

you can of course extract the input to map or the internal command into helper functions.

lib/OpenQA/Setup.pm
```diff
@@ -293,6 +312,9 @@ sub read_config ($app) {
     if (my $minion_fail_job_blocklist = $config->{influxdb}->{ignored_failed_minion_jobs}) {
         $config->{influxdb}->{ignored_failed_minion_jobs} = [split(/\s+/, $minion_fail_job_blocklist)];
     }
+
+    $config->{misc_limits}->{prio_throttling_data} = _load_prio_throttling($app, $config);
+
```
okurz:
```suggestion
    $config->{misc_limits}->{prio_throttling_data} = _load_prio_throttling($app, $config);
```

t/api/04-jobs.t
```diff
@@ -1052,6 +1052,42 @@ subtest 'priority correctly assigned when posting job' => sub {
         $t->json_is('/job/priority', 50, 'feature disabled: prio value unchanged');
     };
 
+    subtest 'priority scaled up due to QEMURAM demand' => sub {
+        my $add = 20;
+        local $jobs_post_params{QEMURAM} = 4096;
+
+        my $config = OpenQA::Setup::read_config($t->app);    # OpenQA::App->singleton->config;
```
okurz: What is `# OpenQA::App->singleton->config;` for?

t/config.t
```diff
@@ -392,4 +395,78 @@ subtest 'Lookup precedence/hiding' => sub {
...
+        [misc_limits]
+        max_job_time_prio_scale = 100
+        prio_throttling_parameters = DEFAULT1 : 0.5: 20, DEFAULT2: 1.0
+    ";
+    $t_dir->child('openqa.ini')->spew($data);
+    $log_mock->mock('warn', sub { $warn_called = 1; $warn_msg = $_[1]; });
+    $app_mock->mock('log', sub { $log_mock });
+
+    OpenQA::App->set_singleton(my $app = Mojolicious->new(log => $quiet_log));
+    my $config = OpenQA::Setup::read_config($app);
```
okurz: We don't need to read+write extra config files but can focus on testing only the algorithms you introduced. I suggest you mock out reading the config file reading to just return the expected string content.

t/config.t
```diff
@@ -392,4 +395,78 @@ subtest 'Lookup precedence/hiding' => sub {
...
+
+    subtest 'prio_throttling_parameters loaded from openqa.ini' => sub {
+        my $prio_params = $config->{misc_limits}->{prio_throttling_parameters};
+        $prio_params =~ s/\s+//g;
+        is $prio_params, "DEFAULT1:0.5:20,DEFAULT2:1.0", 'prio_throttling_parameter correctly loaded from openqa.ini';
```
okurz: What is this test for? We can trust that config files are read and written properly by both trusting the upstream module which does that as well as the former tests which were already executed. I don't see the benefit of this subtest

t/21-needles.t
okurz, line 1: Why this change? Your commit only states what you did but no reason why

lib/OpenQA/Setup.pm
m-dati: Oli, sorry, but this also was discussed yesterday in our collab section, I explained that the expected number here is mainly a `floating point`, not an integer, like you mention here, and the alternative `looks_like_number` would match and let pass other unneeded formats.

etc/openqa/openqa.ini
m-dati: I do changed the text, from the 11 lines, indeed too much, to the actual 4 lines, explaining the meaning as short as possible (waiting for a more stable version when we'll write it into the central documentation).
But, please feel free to show how it'd be your preferable sentence here.

lib/OpenQA/Schema/ResultSet/Jobs.pm
m-dati: This question was long discussed already, here above in https://github.com/os-autoinst/openQA/pull/6952#discussion_r2745515300 and during last collab.session and all we agreed that also the max-job-time funtionality will be moved togheter with this block, into a routine, but in a next PR.

t/api/04-jobs.t
m-dati: typo

t/config.t
m-dati: The config file used here is not an extra, but the real name and type ascii than the one of a production implementation and in my opinion the preparation of the environment more near to the real behavior (instead of mocking), by loading the configuration from a real file, makes more complete and certified those unit tests.
But if you don't like that, I'll change the code.

t/21-needles.t
m-dati: It was discussed days ago with Tina, that helped much to resolve an error that impacted this test after the first code push, simply moving a mock definition in an upper position. I'll update the message.

etc/openqa/openqa.ini
m-dati: I could propose update to:
```
## Throttling: scale priority for job-tests requesting high resources in the settings.
## configure elements in the format: PAR1:SCAL1[:REF1][,…,PAR#:SCAL#[:REF#]
...
```

okurz: your suggestion still has a sentence without a verb and using this weird "job-tests" whatever it should be. Again, why not the text suggested in https://github.com/os-autoinst/openQA/pull/6952/files#r2768987752

t/21-needles.t
perlpunk: How about:
```
Create Mojo::Test app before it is needed

In `create_from_settings` we need the app singleton, but previously the test app was created after that.
```

lib/OpenQA/Schema/ResultSet/Jobs.pm
perlpunk: Yes, it was mentioned here and discussed in dailies, that we plan to do this afterwards as a refactoring exercise.

t/21-needles.t
okurz: And please use https://www.conventionalcommits.org/ . I still don't know if this is a fix, refactoring or a feature

perlpunk: I would say it's a fix

etc/openqa/openqa.ini
m-dati: The text proposed in `Suggestions` is no more applicable as it is: it was intended for the logic splitted in more parameters, with each resource scale in its own parameter and the resource name appended: `prio_throttling_scale_<resource>`.

Here, using the `Alternative`, one only parameter is needed, with a specific lexycon AA:XX:YY, that shall be explained.

"Sentence without a verb": `scale` and `configure` do are verbs 🙂, but well: `scale the...`, `configure the...`.

Anyway, then, what if:
```
## Throttling: set a comma-separated list of test parameters that trigger throttling of scheduled openQA jobs by priority. 
## Example: "PARAM1:SCALE1[:THRESH1][,...,PARAM#:SCALE#[:THRESH#]]"
## Set a scale that the parameter value, minus an optional threshold, is multiplied by and added to each job prio, for each configured resource, as: 
## '$base_prio + $scale# * ($param# - $thresh#)'
...
```

Let's also remember that the optional _threshold reference_  should be also explained, having introduced a scaled prio "bonus" (prio decrease) for values lower than it, against the case without it where unconditional prio increasing is applied, that would need more text here.

lib/OpenQA/Schema/ResultSet/Jobs.pm
m-dati: Assuming throttling "QEMURAM:0.03:1024,HDDSIZEGB:0.01", the log should report:
```
...
Adjusting job priority by <new-prioriy-value> based on resource requirement(s):
 QEMURAM, scale: 0.03, reference: 1024; HDDSIZEGB, scale: 0.01;
```

t/config.t
m-dati: you don't see it, but it is there, because it is the only test where prio_throttling_parameters is directly filled from the ini file, by the call `read_config`, while in the other tests it is tweaked in the local test. Anyway, you don't like that: I'll remove it.

t/config.t
okurz: Yes, please. We should be mindful of test execution time which is a challenge in openQA. So whenever we add tests we should be striving for an efficient approach with sufficient test coverage but also having a long-term sustainable ratio of unit tests, e.g. ~70%, together with around 20% integration tests and  10% system tests, aka. "full stack tests". Every test that writes or reads external tests can not be considered a unit test.

t/api/04-jobs.t
okurz: A typo would be something like "htis", not writing multiple words ;)

lib/OpenQA/Setup.pm
okurz: sure, then just write that and call it "floating point number" in the comment. But better use https://perldoc.perl.org/perlre#/x-and-/xx

By the way, my main suggestion was to use `map`.

lib/OpenQA/Schema/ResultSet/Jobs.pm
okurz: ok. please remove the line breaks. Grepping log lines is way easier that way. Also, where does the semicolon come from?

lib/OpenQA/Schema/ResultSet/Jobs.pm
okurz: > afterwards

after what? But anyway, ok

etc/openqa/openqa.ini
okurz: your suggestion in the code block is good. Yes, of course we should also document the threshold reference.

lib/OpenQA/Schema/ResultSet/Jobs.pm
m-dati: In my opinion, the meaning could be: being the code to move in a routine made of 2 different activities and poo#s, the part of `max-job-time` here is out scope for this PR; then is better to consider a 'fusion' in a proper 'quick' (I hope) new PR, done on purpose 🙂.

lib/OpenQA/Schema/ResultSet/Jobs.pm
m-dati: The line break help to easier read the full text.
This debug line could be longer, if more paramters will be added, skipping more the line on the right side, out sight.
But let's shorten comments and remove breaks.
Semicolon? It is to divide the elements in a sentence, following the language grammar rules to well separate sentences with a subject inside, like each parameter name is.
Is it a problem?

okurz: I mean only the lone semicolon at the end. As separator in the middle it's fine. Regarding the line break you did not respond to what I wrote about grepping

t/api/04-jobs.t
m-dati: My intention already was to remove it in the next commit

t/21-needles.t
m-dati: updated, see last commits.

t/config.t
m-dati: Removed.

t/config.t
m-dati: Changed.

t/api/04-jobs.t
m-dati: removed

lib/OpenQA/Setup.pm
m-dati: done

lib/OpenQA/Setup.pm
m-dati: I applied the map, and tested your command, but even in a local separated snippet, to me it fail, as it is: I had to apply some changes to make it run successfully.

lib/OpenQA/Schema/ResultSet/Jobs.pm
m-dati: As you can easly see in the `$throttling_info` line, the semicolon at end is the sentence terminator of each throttled resource, but needed as separator from the next coming parameter in the loop, so that,the last one of the loop remains 'alone'.
Introducing some long non-functional coding to simply have in the log a different character (like a dot) at end of the sequence, it'd be a pointless over-engineering, only making code heavier.
Lines breaks removed.

etc/openqa/openqa.ini
m-dati: Comments updated, trying to make the text the shorter the possible.

m-dati:
About the non-squashed commits, I did so after a peer-coding section, fixing unit tests impacted by the new code, having been suggested to keep separated the commits.

lib/OpenQA/Setup.pm
perlpunk: Yeah, I can't see how @okurz ' suggestion could work. m/.../g returns a flat list.
Since the validity of the string is already checked above, maybe a split would work here:
```
my %hash;
for my $item (split m/,/, $throttling) {
    my ($resource, $scale, $reference) = split m/:/, $item, 3;
    $hash{$resource} = {scale => $scale, reference => $reference // 0 };
}
```
If @okurz insists on a oneliner, I would ask him to provide a working one :)

lib/OpenQA/Setup.pm
```diff
@@ -54,13 +54,29 @@ sub _load_config ($app, $defaults, $mode_specific_defaults) {
...
+    my $n = qr/[+-]?(?:\d+\.?\d*|\.\d+)/;
+    # Unit format = Param.name: scale: reference(optional)
+    my $u = qr/([A-Z_][A-Z0-9_]*):($n)(?::($n))?/i;
+    $throttling =~ s/\s+//g;
+    unless ($throttling =~ qr/^$u(?:,$u)*$/i) {
+        $app->log->warn("Wrong format in openqa.ini 'prio_throttling_parameters': $throttling");
+        return;
+    }
+    my %hash = map { my ($k, $s, $r) = $_ =~ /$u/g; $k => {scale => $s, reference => $r // 0} }
+      split(',', $throttling);
```
perlpunk: Remember that `split` always splits on a regex, even if you pass it as a simple string.

perlpunk (approved):

lib/OpenQA/Setup.pm
m-dati: Thank you much, I did already a change using `map` in some way and the code seems working fine :D

lib/OpenQA/Schema/ResultSet/Jobs.pm
```diff
@@ -158,6 +158,21 @@ sub create_from_settings ($self, $settings, $scheduled_product_id = undef) {
...
     }
+    # apply resources throttling control
+    if (my $throttling = OpenQA::App->singleton->config->{misc_limits}->{prio_throttling_data}) {
+        my $throttling_info;
+        for my $resource (keys %$throttling) {
+            next if !defined $settings{$resource};
+            my $scale = $throttling->{$resource}->{scale};
+            my $reference = $throttling->{$resource}->{reference};
+            my $prio = int(($settings{$resource} - $reference) * $scale);
+            $throttling_info .= $resource . ", scale: $scale" . ($reference ? ", reference: $reference;" : ';');
```
```suggestion
            $throttling_info .= "$resource, scale: $scale" . ($reference ? ", reference: $reference;" : ';');
```

lib/OpenQA/Setup.pm
```diff
@@ -54,13 +54,29 @@ sub _load_config ($app, $defaults, $mode_specific_defaults) {
...
-
     # ensure default values are assigned; warn if config files were supplied at all
     _read_config_file($config, undef, $defaults, $mode_defaults);
     $app->log->warn('No configuration files supplied, will fallback to default configuration') unless $config_file;
     return $config;
 }
 
+sub _load_prio_throttling ($app, $config) {
+    return
+      unless my $throttling = $config->{misc_limits}->{prio_throttling_parameters};
```
Martchus: Is tidy really requiring the wrapping here?


```suggestion
    return unless my $throttling = $config->{misc_limits}->{prio_throttling_parameters};
```

t/api/04-jobs.t
```diff
@@ -1052,6 +1052,42 @@ subtest 'priority correctly assigned when posting job' => sub {
...
+        $t->post_ok('/api/v1/jobs', form => \%jobs_post_params)->status_is(200);
+        $t->get_ok('/api/v1/jobs/' . $t->tx->res->json->{id})->status_is(200);
+        $t->json_is('/job/priority', 50 + $add, 'increased prio value');
+    };
+
+    subtest 'priority improved due to QEMURAM low demand' => sub {
+        my $add = -10;
+        local $jobs_post_params{QEMURAM} = 1024;
+
+        my $config = OpenQA::Setup::read_config($t->app);    # OpenQA::App->singleton->config;
```
Martchus: Not sure what this comment is suppose to tell one. I'd just remove it.

```suggestion
        my $config = OpenQA::Setup::read_config($t->app);
```

t/config.t
```diff
@@ -392,4 +395,55 @@ subtest 'Lookup precedence/hiding' => sub {
     is_deeply lookup_config_files(@args), \@expected, 'drop-in in overriden dir hides all other config';
 };
 
+subtest 'check throttling configuration validation and application' => sub {
+    my $warn_called = 0;
+    my $log_mock = Test::MockObject->new();
+    my $app_mock = Test::MockObject->new();
+    $log_mock->mock('warn', sub { $warn_called = 1; });
```
Martchus: It would make sense to verify whether the expected warning was logged. (Don't require an exact match but e.g. compare it via `like`.)

t/ui/10-tests_overview.t
```diff
@@ -21,6 +21,7 @@ my $schema = $test_case->init_data(
     schema_name => $schema_name,
     fixtures_glob => '01-jobs.pl 02-workers.pl 04-products.pl 05-job_modules.pl 06-job_dependencies.pl'
 );
+my $t = Test::Mojo->new('OpenQA::WebAPI');
```
Martchus: Why? We shouldn't instantiate a 2nd app unless it is necessary so tests stay as lightweight as possible. If it is now required that the app singleton is set than do just that.

perlpunk: Where is the other test instance created? I couldn't find it, and the test was failing because create_from_settings was trying to access the app.

perlpunk: > If it is now required that the app singleton is set than do just that.

And how would you do that?

Martchus: Looks like other UI tests also just instantiate a 2nd app if needed. However, most at least make at least also proper use of that 2nd app. Some do the instantiation via `my $fake_app = Mojo::Server->new->build_app('OpenQA::WebAPI');` but that's probably not really much better.

By the way, the problem here is that we instantiate the app in a sub process in UI tests. This is ok on its own but of course if we then need the app in the main test process after all we have to create an app there as well. So we have two independent apps which can lead to confusion and is of course not that lightweight.

Maybe it would be best to remove the requirement of having an app in the main process. That would mean finding out where exactly this UI test executes your code from the main test process and avoid that. If it is unavoidable than you could also change your code to avoid the hard dependency on `OpenQA::App->singleton`.

lib/OpenQA/Schema/ResultSet/Jobs.pm
```diff
@@ -158,6 +158,21 @@ sub create_from_settings ($self, $settings, $scheduled_product_id = undef) {
             $new_job_args{priority} += $malus;
         }
     }
+    # apply resources throttling control
+    if (my $throttling = OpenQA::App->singleton->config->{misc_limits}->{prio_throttling_data}) {
```
perlpunk:
```suggestion
    if (my $throttling = OpenQA::App->singleton && OpenQA::App->singleton->config->{misc_limits}->{prio_throttling_data}) {
```

perlpunk: This should fix the failure in 10-tests-overview.t, so you can remove the additional Test::Mojo app

PR #6953, opened 2026-01-29
Decouple dependency of a functionwith duplicated logic on variables

d3flex:
In https://github.com/os-autoinst/openQA/pull/6902/changes, `min_free_percentage` and `min_free_percentage_ar` variable introduced, which contain similar logic. This makes the function more complicated and difficult to follow or debug. The solution here begins with a design which allows:
- to simplify the `_ensure_results_below_threshold`
- handle the variables as one data structure and as such iterate over this structure instead of handling each variable separately
- following this design, reduce if/else
- the need for additional support variables (see any *_ar)
- makes it easier to add new logic

There is a small change at the tests but it should make sense, as it should follow the correct sequence, first archive then results. there was space for more refactoring (wrapping the `_delete_results` in a loop) but it should be simple enough at that point. The `_validate_storage_locations` could also be left out and use the `_get_storage_locations` for the validation, but that would require to pass `$job`. Providing two function, it keeps a clear design with SOLID in mind - doing on thing and without dependencies.

### DIFF
--- a/lib/OpenQA/Task/Job/Limit.pm
+++ b/lib/OpenQA/Task/Job/Limit.pm
@@ -158,6 +158,42 @@ sub _is_valid_percentage ($value) { looks_like_number($value) && $value >= 0 && 
 
 sub _format_percentage_error ($key, $value) { "Configured $key ($value) is not a number between 0 and 100" }
 
+sub _get_storage_struct ($limits) {
+    my @storage_struct;
+    my $results_percentage = $limits->{results_min_free_disk_space_percentage};
+    my $archive_percentage = $limits->{archive_min_free_disk_space_percentage};
+
+    if (defined $results_percentage && $results_percentage ne 'none') {
+        push @storage_struct,
+          {
+            config_key => 'results_min_free_disk_space_percentage',
+            percentage => $results_percentage,
+            dir => resultdir,
+            archived => 0,
+          };
+    }
+
+    if (defined $archive_percentage) {
+        push @storage_struct,
+          {
+            config_key => 'archive_min_free_disk_space_percentage',
+            percentage => $archive_percentage,
+            dir => archivedir,
+            archived => 1,
+          };
+    }
+    return \@storage_struct;
+}
+
+sub _validate_storage_locations (@storage_struct) {
+    for my $tstdir (@storage_struct) {
+        return _format_percentage_error($tstdir->{config_key} => $tstdir->{percentage})
+          unless _is_valid_percentage($tstdir->{percentage});
+    }
+    return undef;
+}
+
+
 sub _account_for_deletion ($margin_bytes, $margin_bytes_main_storage, $deleted_results, $deleted_screenshots = 0) {
     $$margin_bytes += $deleted_results;
     $$margin_bytes_main_storage += $deleted_screenshots;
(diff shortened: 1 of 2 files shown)


t/42-df-based-cleanup.t
```diff
@@ -292,8 +292,8 @@ subtest 'jobs in archive considered via archive_min_free_disk_space_percentage'
     $app->config->{misc_limits}->{archive_min_free_disk_space_percentage} = 31;
 
     my @expected_messages = (
-        'Unable to cleanup enough results from results dir',
```
okurz, line 295: I don't quite understand the commit message and especially seeing this change I recommend to use "conventional commits" same as I and others introduced recently. Is this supposed to be a refactoring or fix or feature?

okurz (changes requested):

t/42-df-based-cleanup.t
d3flex: This comes as a RFC following the discussion on https://github.com/os-autoinst/openQA/pull/6902#discussion_r2720748931. It is a refactoring aiming simplifying and decoupling the function from handling two indentical variables.

okurz: How can it be a refactoring when the order of execution changes?

lib/OpenQA/Task/Job/Limit.pm
```diff
@@ -158,6 +158,42 @@ sub _is_valid_percentage ($value) { looks_like_number($value) && $value >= 0 &&
...
+    if (defined $archive_percentage) {
+        push @storage_struct,
+          {
+            config_key => 'archive_min_free_disk_space_percentage',
+            percentage => $archive_percentage,
+            dir => archivedir,
+            archived => 1,
+          };
+    }
+    return @storage_struct;
```
Martchus:
```suggestion
    return \@storage_struct;
```
And adjust call-sites accordingly.

lib/OpenQA/Task/Job/Limit.pm
```diff
@@ -158,6 +158,42 @@ sub _is_valid_percentage ($value) { looks_like_number($value) && $value >= 0 &&
 
 sub _format_percentage_error ($key, $value) { "Configured $key ($value) is not a number between 0 and 100" }
 
+sub _get_storage_struct ($limits) {
```
Martchus, line 161: A function that pushes other function parameters into an array like this is not how I would envision improving this piece of code. I haven't put much thought into it yet but there's surely a better way to structure this code. Now I find it even more complicated than before. Before there was a slight repetition but it was still simpler than that. The fact that you're adding much more lines than you're removing is also a good indication that we're heading in the wrong direction here.

lib/OpenQA/Task/Job/Limit.pm
```diff
@@ -279,16 +314,33 @@ sub _ensure_results_below_threshold ($job, @) {
...
+
+    if ($results_dir) {
+        my ($ok, @message) = _delete_results(
+            $dry, $jobs, $max_job_id, \@not_important_cond, \@important_cond,
+            \$results_dir->{margin_bytes},
+            \$results_dir->{margin_bytes}, 0
+        );
+        push @messages, @message;
+        $all_ok &&= $ok;
+    }
```
Martchus, line 340: Now the same code section is even longer and frankly also harder to make sense of. For instance `@message, @message_ar` in my version was simpler than tracking the `push @messages, …` invocations in your version.

t/42-df-based-cleanup.t
Martchus: I think this is only changing the order in which the messages occur (but not the order in which the different parts of the cleanup are executed). Still a functional change but at least a very slight one.

Martchus (changes requested):
Adding effectively 52 lines of code just to prevent a little bit of repetition is not worth it. We should probably still avoid the repetition but we can probably do better than that. Note that your PR also *introduces* new repeating patterns.

So I think this PR - as is - changes this code for the worse.

lib/OpenQA/Task/Job/Limit.pm
d3flex: > The fact that you're adding much more lines than you're removing is also a good indication that we're heading in the wrong direction here

It is also wrong to evaluate an approach from the numbers of lines IMO. or justify a good/bad design. I tried to reason in the commit some design decision. Somewhere should mention `more space for refactoring` or something like that. I think my arguments are valid. because is it easier to answer whether the abstraction wrong? or whether the function have clearer responsibility or not? do you think that this PR increase complexity? isnt easier to unit test it? or extended?

>  but there's surely a better way to structure this code

very likely. but it is getting harder and harder to refactor if we do not keep it in some good design which I believe this commit does.

lib/OpenQA/Task/Job/Limit.pm
d3flex: It is debatable. I wont disagree.. lets say it was simpler.. there is still two variables doing mostly the same thing, which makes you write things like 
```
my $margin_bytes = _check_remaining_disk_usage($job, $resultdir, $min_free_percentage);
my $margin_bytes_ar = _check_remaining_disk_usage($job, $archivedir, $min_free_percentage_ar);
```
It is easy to think that this will introduce if/else in the future or special handling. In contrast a data structure makes future changes less possible (if a function has good abstraction) or easier done.

t/42-df-based-cleanup.t
d3flex: > How can it be a refactoring when the order of execution changes?

I wrote it in the commit. It troubles me in the first sight but I didnt spend too much time to dig it. What I made sure is that tracking the proposed changes this made sense. It is exactly as @Martchus mentioned at the end

update: likely the cause is https://github.com/os-autoinst/openQA/pull/6953/changes#diff-2ef6d893f06b310ebe36bf8b60650f5ebfbd1885b30820679957a7696f69d672L245

okurz (changes requested):
Sorry, I also don't see the improvement. This is now introducing more lines of code, more procedural code with for-loops and push which in before was more concise and looking more functional

lib/OpenQA/Task/Job/Limit.pm
Martchus: This is not just about the number of lines. I find your version with an extra function just to compute function parameters harder to make sense of.

I also don't think that there are lots of different "responsibilities" to separate here. I also don't think it is easier to test considering we already have tests that are good enough.

The goal for improvement should be to eliminate repeating patterns but considering the repetition isn't that bad it has to come at a smaller cost than what you're proposing. (And I would also guess that there's a more elegant way.)

d3flex: > I also don't think that there are lots of different "responsibilities" to separate here. I also don't think it is easier to test considering we already have tests that are good enough.

The goal for improvement should be to eliminate also dependencies and abstraction. Having good tests I would argue that it is good enough until you need to refactor and then you realize you have to change a lot of things. Which it wasnt in this case thanks to your previous work. 

> I find your version with an extra function just to compute function parameters harder to make sense of
I am pretty sure that you understand that:
- `_get_storage_struct` has single responsibility
- `_ensure_results_below_threshold` is more flexible and extensible

Yes it could break down more to avoid some duplication but it left out of the scope of this commit.
This commit meant to provide the idea from the https://github.com/os-autoinst/openQA/pull/6902#discussion_r2720748931. Before you have to track a number of variables.  that would make debugging harder. But I dont wish to spend more time for a perfect solution. The problem is in fact already solved. We just can agree on the design. But I just dont see good arguments against these changes to be honest.

d3flex:
not approved in favor of other approach

PR #6956, opened 2026-01-31
feat(ui): add bulk action checkboxes to test overview filters

okurz:
Inspired by github notification selection checkboxes I implemented bulk
action checkboxes allowing to select "All" and "None". Also I added an
additional button to invert the current selection, e.g. to go from
"Passed" to "All but Passed".

## Select all
Screenshot_20260131_131920_filter_box_all

## Select none
Screenshot_20260131_131847_filter_box_none

## Some selected
Screenshot_20260131_131905_filter_box_some

## Invert curent selection
Screenshot_20260131_131937_invert

### DIFF
--- a/assets/javascripts/filter_form.js
+++ b/assets/javascripts/filter_form.js
@@ -26,6 +26,56 @@ function setupFilterForm(options) {
       );
     }
   });
+
+  $('#filter-reset-button').on('click', function () {
+    const form = $('#filter-form');
+    form.find('input[type="text"], input[type="number"]').val('');
+    form.find('input[type="checkbox"]').prop('checked', false).prop('indeterminate', false);
+    form.find('input[hidden]').remove();
+    form.find('select').val([]).trigger('chosen:updated');
+    $('#filter-panel .card-header span').text('no filter present, click to toggle filter form');
+  });
+
+  const updateMasterCheckbox = function (container) {
+    const master = container.find('.filter-bulk-master');
+    if (!master.length) return;
+    const checkboxes = container.find('input[type="checkbox"]:not(.filter-bulk-master)');
+    const checkedCount = checkboxes.filter(':checked').length;
+
+    if (checkedCount === 0) {
+      master.prop('checked', false);
+      master.prop('indeterminate', false);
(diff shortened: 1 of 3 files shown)


d3flex (changes requested):
I am gonna block you only because I found weird this use case:
- select *scheduled*: I see `-` on job state
- select then *setup*: still  `-` on job state
- Now click *Job state   `-`* : all actions are selected **<- the problem**
- click again will select them all

expected:
When nothing is selected, the group action should select all
When at least one is selected the group action `-` should deselect then all

assets/javascripts/filter_form.js
```diff
@@ -26,6 +26,56 @@ function setupFilterForm(options) {
...
+    const container = $(this).closest('.mb-3');
+    container.find('input[type="checkbox"]:not(.filter-bulk-master)').each(function () {
+      $(this).prop('checked', !$(this).prop('checked'));
+    });
+    updateMasterCheckbox(container);
+  });
+
+  $('#filter-results, #filter-states').each(function () {
+    updateMasterCheckbox($(this));
+  });
```
Martchus, line 78: Just and example how code could look like without jQuery:


```suggestion
  Arrray.from(document.querySelectorAll('#filter-results, #filter-states')).forEach(updateMasterCheckbox);
```

okurz: oh, I didn't realize. I was just taking over other codes as template :) I will see what I can do for this function and maybe more code.

Martchus: Well, now you have already jQuery-based code which is not bad per-se. We don't have to rewrite that immediately; I'd just avoid it when writing new code anyway.

Oh, and the example could be even simpler (at least for `forEach`):

```suggestion
  document.querySelectorAll('#filter-results, #filter-states').forEach(updateMasterCheckbox);
```

(And in any case the function `updateMasterCheckbox` needs to deal with getting a normal DOM node and not a jQuery list.)

okurz: Yes, using more `forEach` makes sense. I created
* #6965

PR #6957, opened 2026-02-02
Use Leap 16.0 container in CircleCI

perlpunk:

Needs:
* #6964 
  plus meta change in https://build.opensuse.org/project/show/devel:openQA:ci


tools/ci/run_unit_tests.sh
```diff
@@ -1,5 +1,8 @@
 #!/bin/bash
 set -ex
+
+sudo cpanm TAP::Harness::JUnit Perl::Tidy@20260109
```
d3flex: just a question (other than that all good) - why specific version for Perl::Tidy?

perlpunk: We have always been using a fixed Perl::Tidy version in CI.
And I don't want this to suddenly break if there is a new CPAN release that hasn't made its way to our OBS project yet.

.circleci/config.yml
```diff
@@ -104,11 +104,11 @@ images:
       HELM_TEST: 0
 
   - &base
-    image: registry.opensuse.org/devel/openqa/ci/containers/base:latest
+    image: registry.opensuse.org/devel/openqa/ci/containers16.0/base:latest
```
okurz: this should change back to use a non-versioned path as soon as you have a full verification

perlpunk: of course

perlpunk:
Almost forgot: pushed a commit for the github fullstack workflow

perlpunk:
> Almost forgot: pushed a commit for the github fullstack workflow

That one fails:
https://github.com/perlpunk/openQA/actions/runs/21670305982/job/62476184671

But it fails with the same error in master:
https://github.com/os-autoinst/openQA/actions/runs/21670376976/job/62476422958

and we only run this as dispatch when needed, so nothing we need to fix right now

d3flex:
Should this still be draft? is it ready for review?

PR #6963, opened 2026-02-03
Mark some one line catch statements uncoverable

perlpunk:
Between perl 5.26 / 5.42 and Syntax::Keyword::Try 0.280 / 0.300
the behaviour of coverage changed. An opening
```
catch ($e) {
```
on its own line is now not counted as code anymore, but
```
catch ($e) { something }
```
is counted as code, whereas it was automatically counted as covered
before.

Related issue: https://progress.opensuse.org/issues/180731

### DIFF
--- a/lib/OpenQA/CacheService/Model/Cache.pm
+++ b/lib/OpenQA/CacheService/Model/Cache.pm
@@ -167,7 +167,7 @@ sub track_asset ($self, $asset) {
         $db->query($sql, $asset);
         $tx->commit;
     }
-    catch ($e) { $self->log->error("Tracking asset failed: $e") }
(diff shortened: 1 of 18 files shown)


okurz (approved):

perlpunk:
> would the right solution be [...]

ideally we would try to cover them now, but we need to mark them now to be able to proceed with the Leap 16 ticket.

Martchus (approved):
> ideally we would try to cover them now, but we need to mark them now to be able to proceed with the Leap 16 ticket.

If that's to move forward with the Leap 16 ticket I'm ok with it. (Otherwise it would make more sense to ensure those lines are covered.) Maybe we should create a ticket to re-visit those at some point, though.

PR #6965, opened 2026-02-04
refactor(ui): use more native APIs instead of jquery in filter_form

okurz:
* refactor(ui): replace remaining jQuery usage in bulk action listeners
* refactor(ui): remove jQuery usage from setupFilterForm and parseFilterArguments
* refactor(ui): use native DOM APIs for bulk action logic

### DIFF
--- a/assets/javascripts/filter_form.js
+++ b/assets/javascripts/filter_form.js
@@ -1,114 +1,146 @@
 function setupFilterForm(options) {
   // make filter form expandable
-  $('#filter-panel .card-header').on('click', function () {
-    $('#filter-panel .card-body').toggle(200);
-    if ($('#filter-panel').hasClass('filter-panel-bottom')) {
-      $('html,body').animate({
-        scrollTop: $(document).height()
-      });
-    }
-  });
+  const cardHeader = document.querySelector('#filter-panel .card-header');
+  if (cardHeader) {
+    cardHeader.addEventListener('click', function () {
+      const cardBody = document.querySelector('#filter-panel .card-body');
+      if (cardBody) {
+        $(cardBody).toggle(200);
+        if (document.getElementById('filter-panel').classList.contains('filter-panel-bottom')) {
+          window.scrollTo({top: document.documentElement.scrollHeight, behavior: 'smooth'});
+        }
+      }
+    });
+  }
 
-  $('#filter-panel .help_popover').on('click', function (event) {
-    event.stopPropagation();
+  document.querySelectorAll('#filter-panel .help_popover').forEach(helpPopover => {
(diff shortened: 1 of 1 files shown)


assets/javascripts/filter_form.js
```diff
@@ -1,114 +1,152 @@
...
 
-  $('#filter-panel .help_popover').on('click', function (event) {
-    event.stopPropagation();
-  });
+  const helpPopover = document.querySelector('#filter-panel .help_popover');
+  if (helpPopover) {
+    helpPopover.addEventListener('click', function (event) {
+      event.stopPropagation();
+    });
+  }
```
Martchus: This is only equivalent if we only have a single help popover within that panel.

assets/javascripts/filter_form.js
```diff
@@ -1,114 +1,152 @@
...
+      input.value = val;
+      input.name = key;
+      input.hidden = true;
+      hiddenInputs.push(input);
     }
+  });
+
+  if (form) {
+    hiddenInputs.forEach(input => form.appendChild(input));
   }
```
Martchus, line 140:
```suggestion
  if (form) {
    form.append(...hiddenInputs);
}
```

assets/javascripts/filter_form.js
```diff
@@ -1,114 +1,152 @@
...
-  for (var i = 0; i < hiddenInputs.length; i++) {
-    $('#filter-form').append(hiddenInputs[i]);
-  }
+
   if (filterLabels.length > 0) {
-    $('#filter-panel .card-header')
-      .find('span')
-      .text('current: ' + filterLabels.join(', '));
+    const headerSpan = document.querySelector('#filter-panel .card-header span');
+    if (headerSpan) {
```
Martchus: Checking whether the span exists and doing nothing otherwise is quite close to the previous behavior. However, maybe it makes more sense to avoid checks likes these in most cases where the elements are expected to exist.

assets/javascripts/filter_form.js
okurz: right, this should be forEach as well, I guess? Pushed an update

assets/javascripts/filter_form.js
okurz: done

assets/javascripts/filter_form.js
okurz: done

assets/javascripts/filter_form.js
Martchus: It depends on whether the "if" part is true or false. (But of course a forEach can't hurt.)

d3flex (approved):

PR #6971, opened 2026-02-05
feat(ui): make test overview summary counts clickable for quick filtering

okurz:
The job counts in the "Overall Summary" section of the test overview
page are now buttons that link to the same page with appropriate
'result' or 'state' filters applied. This allows for quick filtering
of jobs by their status.

An "All" button was also added to clear these filters.

https://github.com/user-attachments/assets/b0735c25-a17f-4919-bf64-62232c07aa50

Related progress issue: https://progress.opensuse.org/issues/184264

Before:
* #6978

### DIFF
--- a/assets/stylesheets/overview.scss
+++ b/assets/stylesheets/overview.scss
@@ -21,12 +21,11 @@
     margin-left: 0.5rem;
 }
 
+#summary .badge-btn.btn-outline-light {
+    color: black;
+}
 #summary .badge {
-    font-size: 90%;
-    padding: 4px 6px 4px;
-    border-radius: 0.5rem;
-    color: #fff;
-    background-color: #bbb;
+    top: 0px;
 }
 #summary .time-params {
     font-weight: lighter;
--- a/lib/OpenQA/WebAPI/Controller/Test.pm
+++ b/lib/OpenQA/WebAPI/Controller/Test.pm
@@ -106,6 +106,17 @@ my %BADGE_CHARLENS = (
     '[0-9]' => 0.64
 );
 
+my %SUMMARY_CATEGORY_QUERY = (
+    passed => {result => PASSED, state => undef},
+    softfailed => {result => SOFTFAILED, state => undef},
+    failed => {result => FAILED, state => undef},
+    not_complete => {result => [NOT_COMPLETE_RESULTS], state => undef},
+    scheduled => {result => undef, state => SCHEDULED},
+    running => {result => undef, state => [EXECUTION_STATES]},
+    aborted => {result => [ABORTED_RESULTS], state => undef},
+    none => {result => NONE, state => undef},
+);
+
 sub referer_check ($self) {
     return $self->reply->not_found if (!defined $self->param('testid'));
     my $referer = $self->req->headers->header('Referer') // '';
@@ -904,6 +915,7 @@ sub overview ($self) {
         until => $search_args->{until},
         parallel_children_collapsable_results_sel => $config->{global}->{parallel_children_collapsable_results_sel},
         summary_parts => \@summary_parts,
+        summary_category_query => \%SUMMARY_CATEGORY_QUERY,
         only_distri => $only_distri,
         limit_exceeded => $exceeded_limit,
     );
--- a/t/10-tests_overview.t
+++ b/t/10-tests_overview.t
@@ -28,7 +28,7 @@ $t->get_ok('/tests/overview' => form => {distri => 'opensuse', version => '13.1'
 
 my $summary = get_summary;
 like($summary, qr/Overall Summary of opensuse 13\.1 build 0091/i);
-like($summary, qr/Passed: 3 Scheduled: 2 Running: 2 None: 1$/i);
+like($summary, qr/Passed: 3 Scheduled: 2 Running: 2 None: 1/i);
 
 # Check the headers
 $t->element_exists('#flavor_DVD_arch_i586');
@@ -101,7 +101,7 @@ $t->text_is('#res_DVD_x86_64_doc .failedmodule *' => 'logpackages', 'failed modu
 $t->get_ok('/tests/overview' => form => {distri => 'opensuse', version => '13.1'})->status_is(200);
 $summary = get_summary;
 like($summary, qr/Summary of opensuse 13\.1 build 0091/i, 'summary for 13.1');
-like($summary, qr/Passed: 3 Scheduled: 2 Running: 2 None: 1$/i, 'summary badges for 13.1');
+like($summary, qr/Passed: 3 Scheduled: 2 Running: 2 None: 1/i, 'summary badges for 13.1');
 
 $form = {distri => 'opensuse', version => '13.1', groupid => 1001};
 $t->get_ok('/tests/overview' => form => $form)->status_is(200);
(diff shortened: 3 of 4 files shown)


lib/OpenQA/WebAPI/Controller/Test.pm
```diff
@@ -890,6 +890,17 @@ sub overview ($self) {
...
+    my %summary_cat_query = (
+        passed => {result => PASSED, state => undef},
+        not_complete => {result => [NOT_COMPLETE_RESULTS], state => undef},
+        softfailed => {result => SOFTFAILED, state => undef},
+        failed => {result => FAILED, state => undef},
+        scheduled => {state => SCHEDULED, result => undef},
+        running => {state => [EXECUTION_STATES], result => undef},
+        aborted => {result => [ABORTED_RESULTS], state => undef},
+        none => {result => NONE, state => undef},
+    );
```
Martchus: This looks like static data so it would probably be more efficient to instantiate it once outside the function instead of every time this function is executed.

lib/OpenQA/WebAPI/Controller/Test.pm
```diff
@@ -890,6 +890,17 @@ sub overview ($self) {
         _add_distri_and_version_to_summary(\@summary_parts, $distri, $version, 0);
     }
 
+    my %summary_cat_query = (
+        passed => {result => PASSED, state => undef},
+        not_complete => {result => [NOT_COMPLETE_RESULTS], state => undef},
+        softfailed => {result => SOFTFAILED, state => undef},
+        failed => {result => FAILED, state => undef},
+        scheduled => {state => SCHEDULED, result => undef},
```
Martchus: It would be nice to keep the order consistent.

Martchus (commented):
The video shows that the padding when hovering is very uneven.

okurz:
> The video shows that the padding when hovering is very uneven.

I don't see what you mean. Can you clarify?

lib/OpenQA/WebAPI/Controller/Test.pm
okurz: done

lib/OpenQA/WebAPI/Controller/Test.pm
okurz: done

okurz:
> The left and top padding is notably smaller than the bottom and right padding. This is when hovering over the figures and e.g. the whole "Soft-Failed: 2" text has a background.

Got it.

Is this better with +4px left and +2px top

Screenshot_20260209_111955_clickable_boxes_padding_p4left_p2top

lib/OpenQA/WebAPI/Controller/Test.pm
```diff
@@ -106,6 +106,17 @@ my %BADGE_CHARLENS = (
     '[0-9]' => 0.64
 );
 
+my %SUMMARY_CAT_QUERY = (
```
d3flex: what about `%SUMMARY_GROUP_STATE` or something similar?

lib/OpenQA/WebAPI/Controller/Test.pm
```diff
@@ -106,6 +106,17 @@ my %BADGE_CHARLENS = (
...
+    passed => {result => PASSED, state => undef},
+    softfailed => {result => SOFTFAILED, state => undef},
+    failed => {result => FAILED, state => undef},
+    not_complete => {result => [NOT_COMPLETE_RESULTS], state => undef},
```
d3flex, line 113: Why did you go with `[]` for some of them? And maybe if you could use the same order because is tricky to have some with one order and some with another

assets/stylesheets/overview.scss
```diff
@@ -22,12 +22,8 @@
...
 
 #summary .badge {
-    font-size: 90%;
-    padding: 4px 6px 4px;
-    border-radius: 0.5rem;
-    color: #fff;
-    background-color: #bbb;
-}
+    top: 0px;
+  }
```
d3flex: indentation is off

lib/OpenQA/WebAPI/Controller/Test.pm
perlpunk: > Why did you go with `[]` for some of them?

because `NOT_COMPLETE_RESULTS` is a list

t/10-tests_overview.t
```diff
@@ -176,10 +176,36 @@ $t->element_exists_not('.state_cancelled');
...
+like(get_summary, qr/current time Failed: 1/i);
 $t->element_exists('#res_DVD_x86_64_doc .result_failed');
 $t->element_exists_not('#res_DVD_x86_64_kde .result_passed');
 
+subtest 'clickable summary buttons' => sub {
+    $t->get_ok('/tests/overview' => form => {distri => 'opensuse', version => 'Factory', build => '0048'})
+      ->status_is(200);
+    my $summary = $t->tx->res->dom->at('#summary .card-body');
+    my $failed_link = $summary->at('a[href*="result=failed"]');
+    ok($failed_link, 'Failed count is clickable');
```
perlpunk:
```suggestion
    ok $failed_link, 'Failed count is clickable';
```
same for the rest of the tests

templates/webapi/test/overview.html.ep
```diff
@@ -49,17 +49,18 @@
                 </span>
             </button>
             % }
-            % my @badges = qw(success secondary  warning     danger info      primary light   light);
-            % my @labels = qw(Passed  Incomplete Soft-Failed Failed Scheduled Running Aborted None);
+            % my @badges = qw(success warning  danger secondary info      primary light   light);
+            % my @labels = qw(Passed  Soft-Failed Failed Incomplete Scheduled Running Aborted None);
```
perlpunk: Now the spaces seem to be off.
Previously the columns were matching.

lib/OpenQA/WebAPI/Controller/Test.pm
Martchus: What is a "cat" query?

t/10-tests_overview.t
```diff
@@ -176,10 +176,36 @@ $t->element_exists_not('.state_cancelled');
...
+    like($failed_link->attr('href'), qr/distri=opensuse/, 'preserves distri');
```
Martchus: I'd save `$failed_link->attr('href')` in a variable.

lib/OpenQA/WebAPI/Controller/Test.pm
okurz: short for "category". This is based on templates/webapi/test/overview.html.ep

```
55             % for my $category (qw(passed softfailed failed not_complete scheduled running aborted none)) {
```

I don't know what a "summary group state" is but "state" would be ambiguous as there is a "job state". I now called this `%SUMMARY_CATEGORY_QUERY`

assets/stylesheets/overview.scss
okurz: done

t/10-tests_overview.t
okurz: I also prefer this style however here I followed the style of the existing file. Would be cool to have a perlcritic rule to ensure consistency. Anyway, fixed.

templates/webapi/test/overview.html.ep
okurz: done

lib/OpenQA/WebAPI/Controller/Test.pm
okurz: short for "category". This is based on templates/webapi/test/overview.html.ep

```
55             % for my $category (qw(passed softfailed failed not_complete scheduled running aborted none)) {
```

Changed to `%SUMMARY_CATEGORY_QUERY`

t/10-tests_overview.t
okurz: done

okurz:
> > Image
> > Not sure if I did some different but I do not see the colours
> 
> what are the results you look at? There is incomplete but what are the other two?

ah, cancelled job has no result. Fixed:

Screenshot_20260209_185918_none_color_fixed

lib/OpenQA/WebAPI/Controller/Test.pm
d3flex: @okurz Can we address the second part as well before resolve it?

okurz: > And maybe if you could use the same order because is tricky to have some with one order and some with another

do you mean that "scheduled" and "running" mention state first and result second? I selected to mention the key first that has a value other than undef but I agree that ordering by key is more common. Updated.

PR #6977, opened 2026-02-07
Fix instabilities

okurz:
* test(25-cache-service): fix race conditions
* test(ui/21-admin-needles): properly wait for modal dialog and deletion
* test(ui/13-admin): properly wait for API key deletion
* test(40-openqa-clone-job): properly isolate from system config
* test(15-asset): bump timeout to current runtime

### DIFF
--- a/t/25-cache-service.t
+++ b/t/25-cache-service.t
@@ -44,7 +44,7 @@ use POSIX '_exit';
 use Mojo::IOLoop::ReadWriteProcess qw(queue process);
 use Mojo::IOLoop::ReadWriteProcess::Session 'session';
 use OpenQA::Test::Utils
-  qw(fake_asset_server cache_minion_worker cache_worker_service wait_for_or_bail_out perform_minion_jobs);
+  qw(fake_asset_server cache_minion_worker cache_worker_service wait_for_or_bail_out perform_minion_jobs wait_for);
 use OpenQA::Test::TimeLimit '90';
 use Mojo::Util qw(md5_sum);
 use OpenQA::CacheService;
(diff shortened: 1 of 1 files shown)


Martchus (approved):

t/40-openqa-clone-job.t
```diff
@@ -4,12 +4,16 @@
...
 use Test::Most;
 use Test::Warnings qw(warning :report_warnings);
+use Mojo::File 'tempdir';
 use FindBin;
 use lib "$FindBin::Bin/lib", "$FindBin::Bin/../external/os-autoinst-common/lib";
 use OpenQA::Test::TimeLimit '10';
 use OpenQA::Test::Utils qw(run_cmd test_cmd);
 use Test::MockModule;
-$ENV{OPENQA_CONFIG} = '';
+
```
perlpunk, line 12: Maybe we should just set it to `$FindBin::Bin/../etc/config`?

t/25-cache-service.t
```diff
@@ -434,9 +437,15 @@ subtest 'Multiple minion workers (parallel downloads, almost simulating real sce
     }
     'assets';
 
-    ok($cache_client->asset_exists('localhost', 'sle-12-SP3-x86_64-0368-200_88888@64bit.qcow2'),
-        "Asset $_ downloaded correctly")
-      for @assets;
+    for my $asset (@assets) {
+        ok(
+            wait_for(
+                sub { $cache_client->asset_exists('localhost', 'sle-12-SP3-x86_64-0368-200_88888@64bit.qcow2') },
```
perlpunk, line 443: Why do we even wait for the same asset multiple times? `$asset` is not even used in there.

t/40-openqa-clone-job.t
okurz: That would make tests depend on this default config when we want to be independent

t/25-cache-service.t
okurz: I don't know. I didn't change that part

perlpunk: I know. But I wanted to mention it. It looks wrong.
Should I create a ticket for it? Otherwise we'll forget

t/40-openqa-clone-job.t
perlpunk: But several tests already depend on it, I think.
Having three new lines for having an empty config seems a bit too much.
Maybe we should have a test config then?

t/25-cache-service.t
okurz: Yes, please create a ticket

t/40-openqa-clone-job.t
okurz: ok, I found that we define multiple different config dirs in multiple cases but there is actually a proper test "mock" config which we can use so this is just one line now

```
$ENV{OPENQA_CONFIG} = "$FindBin::Bin/data";
```

PR #6978, opened 2026-02-09
feat: support inverted result filters in /tests/overview

okurz:
* feat: support inverted result filters in /tests/overview
* feat(ui): use clickable test overview summary counts for quick filtering

## Only "failed" selected
Screenshot_20260208_201556_failed

## All but "failed" selected with updated URL
Screenshot_20260217_161617_openQA_not_failed

After:
* #6971

### DIFF
--- a/assets/javascripts/filter_form.js
+++ b/assets/javascripts/filter_form.js
@@ -25,9 +25,22 @@ function setupFilterForm(options) {
 
   const filterForm = document.getElementById('filter-form');
   if (filterForm) {
-    filterForm.addEventListener('submit', function () {
+    filterForm.addEventListener('submit', function (event) {
+      event.preventDefault();
       const currentQuery = window.location.search.substring(1);
       const formData = new FormData(filterForm);
+
+      // optimize results and states to use negative filters if shorter
+      ['result', 'state'].forEach(key => {
+        const checkboxes = Array.from(filterForm.querySelectorAll(`input[name="${key}"]`));
+        const checked = checkboxes.filter(cb => cb.checked);
+        const unchecked = checkboxes.filter(cb => !cb.checked);
+        if (checked.length > 1 && unchecked.length > 0 && unchecked.length < checked.length) {
+          formData.delete(key);
+          unchecked.forEach(cb => formData.append(`${key}__not`, cb.value));
+        }
+      });
+
       const newQuery = new URLSearchParams(formData).toString();
 
       if (newQuery !== currentQuery) {
@@ -40,6 +53,7 @@ function setupFilterForm(options) {
           progress.innerHTML = '<i class="fa fa-cog fa-spin fa-2x fa-fw"></i> <span>Applying filter…</span>';
           cardBody.appendChild(progress);
         }
+        window.location.search = newQuery;
       }
     });
   }
--- a/assets/javascripts/overview.js
+++ b/assets/javascripts/overview.js
@@ -223,6 +223,18 @@ function setupOverview() {
     if (key === 'result' || key === 'state') {
       flags[key][val] = true;
       return formatFilter(val);
+    } else if (key === 'result__not' || key === 'state__not') {
+      const baseKey = key.slice(0, -5);
+      const isFirst = !flags[baseKey].__inverted;
+      flags[baseKey].__inverted = true;
+      document.querySelectorAll(`#filter-${baseKey}s input[name="${baseKey}"]`).forEach(cb => {
+        if (isFirst) {
+          flags[baseKey][cb.value] = cb.value !== val;
+        } else if (cb.value === val) {
+          flags[baseKey][cb.value] = false;
+        }
+      });
+      return 'NOT ' + formatFilter(val);
     } else if (key === 'todo') {
       form.todo.checked = val !== '0';
       return 'TODO';
--- a/lib/OpenQA/WebAPI/Plugin/Helpers.pm
+++ b/lib/OpenQA/WebAPI/Plugin/Helpers.pm
@@ -516,16 +516,21 @@ sub _every_non_empty_param ($c, $param_key) {
 sub _compute_overview_filtering_params ($c) {
     my $states = $c->every_non_empty_param('state');
     my $results = $c->every_non_empty_param('result');
+    my $states_not = $c->every_non_empty_param('state__not');
+    my $results_not = $c->every_non_empty_param('result__not');
     my $archs = $c->every_non_empty_param('arch');
     my $machines = $c->every_non_empty_param('machine');
     my $failed_modules = $c->every_non_empty_param('failed_modules');
-    my %filters;
-    $filters{'me.state'} = {-in => $states} if @$states;
-    $filters{'me.result'} = {-in => $results} if @$results;
-    $filters{'me.result'} = FAILED if @$failed_modules;
-    $filters{ARCH} = {-in => $archs} if @$archs;
-    $filters{MACHINE} = {-in => $machines} if @$machines;
-    return [\%filters];
+    my @conds = (
+        (@$states ? {'me.state' => {-in => $states}} : ()),
+        (@$results ? {'me.result' => {-in => $results}} : ()),
+        (@$states_not ? {'me.state' => {-not_in => $states_not}} : ()),
+        (@$results_not ? {'me.result' => {-not_in => $results_not}} : ()),
+        (@$failed_modules ? {'me.result' => FAILED} : ()),
+        (@$archs ? {ARCH => {-in => $archs}} : ()),
+        (@$machines ? {MACHINE => {-in => $machines}} : ()),
+    );
+    return \@conds;
 }
 
 sub _groups_for_globs ($c) {
--- a/t/10-tests_overview.t
+++ b/t/10-tests_overview.t
@@ -444,6 +444,29 @@ $t->get_ok(
 like(get_summary, qr/current time/i, 'Job was successful, so failed_modules does not show it');
 $t->element_exists_not('#res-99946', 'no module has failed');
 
+subtest 'Inverted filters' => sub {
+    $t->get_ok('/tests/overview?distri=opensuse&version=13.1&build=0091&result__not=passed')->status_is(200);
+    my $summary = get_summary;
+    unlike($summary, qr/Passed: [1-9]/i, 'Passed jobs are excluded');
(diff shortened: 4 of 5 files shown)


assets/javascripts/overview.js
```diff
@@ -223,6 +223,16 @@ function setupOverview() {
     if (key === 'result' || key === 'state') {
       flags[key][val] = true;
       return formatFilter(val);
+    } else if (key === 'result__n' || key === 'state__n') {
+      const baseKey = key.slice(0, -3);
+      const containerId = 'filter-' + baseKey + 's';
+      const container = document.getElementById(containerId);
+      if (container) {
+        Array.from(container.querySelectorAll(`input[name="${baseKey}"]`)).forEach(cb => {
```
Martchus:
```suggestion
        container.querySelectorAll(`input[name="${baseKey}"]`).forEach(cb => {
```

assets/javascripts/overview.js
```diff
@@ -223,6 +223,16 @@ function setupOverview() {
     if (key === 'result' || key === 'state') {
       flags[key][val] = true;
       return formatFilter(val);
+    } else if (key === 'result__n' || key === 'state__n') {
+      const baseKey = key.slice(0, -3);
+      const containerId = 'filter-' + baseKey + 's';
+      const container = document.getElementById(containerId);
```
Martchus:
```suggestion
      const container = document.getElementById(`filter-${baseKey}s`);
```

assets/javascripts/overview.js
```diff
@@ -223,6 +223,16 @@ function setupOverview() {
...
+      if (container) {
+        Array.from(container.querySelectorAll(`input[name="${baseKey}"]`)).forEach(cb => {
+          if (cb.value !== val) flags[baseKey][cb.value] = true;
+        });
+      }
```
Martchus: Or even just:
```suggestion
      container.querySelectorAll(`#filter-${baseKey}s input[name="${baseKey}"]`).forEach(cb => {
        if (cb.value !== val) flags[baseKey][cb.value] = true;
      });
```

assets/javascripts/filter_form.js
```diff
@@ -25,9 +25,22 @@ function setupFilterForm(options) {
...
       const formData = new FormData(filterForm);
+
+      // optimize results and states to use negative filters if shorter
+      ['result', 'state'].forEach(key => {
+        const checkboxes = Array.from(filterForm.querySelectorAll(`input[name="${key}"]`));
+        const checked = checkboxes.filter(cb => cb.checked);
+        if (checked.length > 1 && checked.length === checkboxes.length - 1) {
+          const unchecked = checkboxes.find(cb => !cb.checked);
+          formData.delete(key);
+          formData.set(`${key}__n`, unchecked.value);
```
Martchus: There my be more than one unchecked checkboxes. Why do we only care about the value of the first one?

lib/OpenQA/WebAPI/Plugin/Helpers.pm
```diff
@@ -516,12 +516,16 @@ sub _every_non_empty_param ($c, $param_key) {
...
+    my $states_n = $c->every_non_empty_param('state__n');
+    my $results_n = $c->every_non_empty_param('result__n');
     my $archs = $c->every_non_empty_param('arch');
     my $machines = $c->every_non_empty_param('machine');
     my $failed_modules = $c->every_non_empty_param('failed_modules');
     my %filters;
     $filters{'me.state'} = {-in => $states} if @$states;
     $filters{'me.result'} = {-in => $results} if @$results;
+    $filters{'me.state'} = {-not_in => $states_n} if @$states_n;
+    $filters{'me.result'} = {-not_in => $results_n} if @$results_n;
```
Martchus: So if `state__n`/`result__n` is present we completely ignore the non-negative parameters. That's not very intuitive.

assets/javascripts/overview.js
```diff
@@ -223,6 +223,16 @@ function setupOverview() {
     if (key === 'result' || key === 'state') {
       flags[key][val] = true;
       return formatFilter(val);
+    } else if (key === 'result__n' || key === 'state__n') {
```
Martchus: Why the double underscore?

okurz: because netbox does it this way. Is there a different, more common or more convenient format that you prefer?

Martchus: It is certainly very uncommon to have to consecutive underscores. So I'd just use a single one.

Martchus: Probably `not_result` and `not_state` would be most intuitive.

okurz: @nicksinger wdyt?

nicksinger: the `__<filter>`-syntax is certainly special but I guess was chosen for a good reason by the netbox devs. But they also support numerical comparisons like `__gt` or `__lt` which we don't really need here, do we?

Then again, I like the flow of reading: `result__n=passed` - field, operator, comparison. How about `__not`?

okurz: So regarding the format I did a bit of research. Let's compare:

1. Suffix-based negation with `__` (e.g., field__n, field__not)
  This approach uses a delimiter followed by an operator. django also seems to use this as separator also for multiple operators, e.g. "pub_date__gte". I like the brevity of `__n` but if you insist we can also go further with `__not`. Reasoning: The double underscore is a "meta-separator" that prevents name collisions with fields that might contain a single. Reference: Django QuerySet Field Lookups https://docs.djangoproject.com/en/stable/ref/models/querysets/#field-lookups
2. Prefix operators (e.g., -field, !field)
  GitHub/Google Search Style: Using a minus sign (-) to exclude results (e.g., -label:bug). Pros: Highly intuitive and extremely concise. Cons: In URL query parameters, - is sometimes interpreted as a space or needs careful handling. ! (exclamation mark) is another candidate but requires URL encoding (%21), which reduces readability, at least in logs. Reference: GitHub Search Syntax - Excluding terms https://docs.github.com/en/search-github/searching-on-github/searching-for-repositories#exclude-query-terms
3. Value-based operators (e.g., field=ne:value)
   MongoDB/JSON API Style where the operator is part of the value Reasoning: Keeps the parameter key clean. Very extensible, e.g., field=gt:10, field=ne:value. Reference: JSON API Recommendation for Filtering https://jsonapi.org/recommendations/#filtering
4. Suffix-based negation with single character separator, e.g. `_`
   As alternative to 1. the single separator might look clean in some cases but as soon as we have a key that already uses the underscore as separator which we use in many other cases the meaning becomes ambiguous as well as makes parsing more complicated

Given the above I would favor 1. or 2.

Martchus: Ok, I guess that is enough justification to go with the double underscore. I would still prefer `not` because a single letter is just not very intuitive to understand.

assets/javascripts/overview.js
okurz: done

assets/javascripts/overview.js
okurz: done

assets/javascripts/overview.js
okurz: done

assets/javascripts/filter_form.js
okurz: changed to iterate over all, done

assets/javascripts/overview.js
okurz: done

lib/OpenQA/WebAPI/Plugin/Helpers.pm
okurz: basically done but I like the code less with `push @conds`. will look into this

okurz: done

lib/OpenQA/WebAPI/Plugin/Helpers.pm
```diff
@@ -516,16 +516,21 @@ sub _every_non_empty_param ($c, $param_key) {
 sub _compute_overview_filtering_params ($c) {
     my $states = $c->every_non_empty_param('state');
     my $results = $c->every_non_empty_param('result');
+    my $states_n = $c->every_non_empty_param('state__not');
```
d3flex: why not `$states_not` `$results_not`. IIUC they are equivelant to their long form. Now the URL  can show other times `result__n=timeout_exceeded` and other times `result__not=timeout_exceeded`, no?

okurz: > why not `$states_not` `$results_not`. IIUC they are equivelant to their long form. Now the URL can show other times `result__n=timeout_exceeded` and other times `result__not=timeout_exceeded`, no?

where did you find "result__n" as query parameter? This line just defines a local variable, nothing else. Anyway, renamed.

d3flex: image
http://localhost/tests/overview?arch=&flavor=&machine=&test=&modules=&module_re=&group_glob=&not_group_glob=&comment=&distri=microos&distri=opensuse&version=Tumbleweed&build=20251121&groupid=1**&result__n=timeout_exceeded**

it is just happened to find it like that. in some other cases where like `result__not`.

okurz: @d3flex I don't see that as possible as the code does not support `__n` anywhere anymore. I assume you had an outdated checkout of this PR. Anyway I rebased once more and pushed.

asmorodskyi (approved):
LGTM

lib/OpenQA/WebAPI/Plugin/Helpers.pm
d3flex: you even updated the regex in the t/ui/10-tests_overview.t and is more clear now. thanks

d3flex (approved):

Martchus (commented):
It would be nice to have a test for the JavaScript code. That shouldn't be too hard with Selenium.

Martchus:
> > It would be nice to have a test for the JavaScript code. That shouldn't be too hard with Selenium.
> 
> There is already https://github.com/os-autoinst/openQA/pull/6978/files#diff-e273ca4b4bd6fbd253c514adb40b3b8d4549c16039168b20333491b6ca5f57ddR722 but I guess you mean clicking buttons

Yes, I mean clicking the buttons.

PR #6979, opened 2026-02-09
feat(ui): remove empty query parameters from /tests/overview URL

okurz:
Cleans up the URL after submitting the filter form by removing
parameters with empty values.

Related progress issue: https://progress.opensuse.org/issues/184264

### DIFF
--- a/assets/javascripts/filter_form.js
+++ b/assets/javascripts/filter_form.js
@@ -41,7 +41,13 @@ function setupFilterForm(options) {
         }
       });
 
-      const newQuery = new URLSearchParams(formData).toString();
+      const params = new URLSearchParams(formData);
+      const keysToDelete = [];
+      params.forEach((val, key) => {
+        if (val === '') keysToDelete.push(key);
+      });
+      keysToDelete.forEach(key => params.delete(key));
+      const newQuery = params.toString();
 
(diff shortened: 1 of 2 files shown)


assets/javascripts/filter_form.js
```diff
@@ -41,7 +41,13 @@ function setupFilterForm(options) {
         }
       });
 
-      const newQuery = new URLSearchParams(formData).toString();
+      const params = new URLSearchParams(formData);
+      const keysToDelete = [];
+      params.forEach((val, key) => {
+        if (val === '') keysToDelete.push(key);
+      });
```
perlpunk, line 48: Why an extra `keysToDelete` variable and a second loop instead of doing the `params.delete` in that loop already?

Martchus: Looks to me that @okurz (or the AI) was afraid that the deletion within the for loop causes problems (e.g. in C++ it would as iterators might be invalidated). Since I am not 100 % sure about JavaScript and `forEach` I didn't comment. However, I *suppose* this is indeed needless in JavaScript with `forEach`. (My guess is that the AI is mixing coding patterns from different programming languages.)

okurz: ->
* #7017

PR #6988, opened 2026-02-09
use native JS instead of jQuery everywhere

okurz:
## Reasoning
- **Modern Standards:** Native browser APIs now provide everything jQuery once solved.
- **Performance:** Removing a 30KB+ (minified/gzipped) library and its overhead.
- **Maintenance:** Reducing dependencies and aligning with modern JS best practices.

After:
* #7839
* #7840
* #7841
* #7842
* #7843
* #7844
* #7845

### DIFF
--- a/assets/javascripts/admintable.js
+++ b/assets/javascripts/admintable.js
@@ -489,9 +489,7 @@ function setupAdminTable(isAdmin) {
 
   // make the height of text areas fit its contents
   dataTable.on('draw', function () {
-    table.find('textarea').each(function () {
-      updateTextArea(this);
-    });
+    tableEl.querySelectorAll('textarea').forEach(updateTextArea);
   });
 
   // set/update page-global state (there can only be one admin table at a page anyways)
--- a/assets/javascripts/admintable.js
+++ b/assets/javascripts/admintable.js
@@ -31,7 +31,8 @@ function showAdminTableRow(row) {
   }
 
   // scroll to the new row
-  $('html').animate({scrollTop: $(row.node()).offset().top}, 250);
+  const rowNode = row.node();
+  window.scrollTo({top: rowNode.getBoundingClientRect().top + window.scrollY, behavior: 'smooth'});
 }
 
 function addAdminTableRow() {
@@ -40,7 +41,7 @@ function addAdminTableRow() {
   // add new row
   const newRow = adminTable.row.add(adminTable.emptyRow);
   const newRowIndex = newRow.index();
-  adminTable.rowData[newRowIndex] = jQuery.extend({isEditing: true}, adminTable.emptyRow);
+  adminTable.rowData[newRowIndex] = Object.assign({isEditing: true}, adminTable.emptyRow);
   newRow.invalidate().draw();
 
   showAdminTableRow(newRow);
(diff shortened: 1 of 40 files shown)


assets/javascripts/admintable.js
```diff
@@ -472,25 +468,25 @@ function setupAdminTable(isAdmin) {
...
     });
   });
 
   // make the height of text areas fit its contents
   dataTable.on('draw', function () {
-    table.find('textarea').each(function () {
-      updateTextArea(this);
+    tableEl.querySelectorAll('textarea').forEach(function (textarea) {
+      updateTextArea(textarea);
     });
```
Martchus:
```suggestion
    tableEl.querySelectorAll('textarea').forEach(updateTextArea);
```

assets/javascripts/audit_log.js
```diff
@@ -282,27 +282,48 @@ function rescheduleProductForActionLink(link) {
...
+      return response.text().then(text => {
+        let json;
+        try {
+          json = JSON.parse(text);
+        } catch (e) {
+          // not JSON
+        }
+        return {response, json, text};
+      });
+    })
```
Martchus: We should probably streamline this with other pieces of code we already have that use native the AJAX APIs.

assets/javascripts/filter_form.js
```diff
@@ -5,7 +5,11 @@ function setupFilterForm(options) {
     cardHeader.addEventListener('click', function () {
       const cardBody = document.querySelector('#filter-panel .card-body');
       if (cardBody) {
-        $(cardBody).toggle(200);
+        if (cardBody.style.display === 'none' || cardBody.style.display === '') {
+          cardBody.style.display = 'block';
+        } else {
+          cardBody.style.display = 'none';
+        }
```
Martchus: I'm wondering whether this is actually equivalent. Probably the new code now lacks the animation. That could of course be done via CSS these days but after a brief look at this PR I haven't seen such a change. Maybe we can also use a Bootstrap API? Or is bootstrap handling this even when just setting raw CSS styling?

assets/javascripts/filter_form.js
```diff
@@ -58,7 +62,9 @@ function setupFilterForm(options) {
       filterForm.querySelectorAll('input[hidden]').forEach(input => {
         input.remove();
       });
-      $(filterForm).find('select').val([]).trigger('chosen:updated');
+      filterForm.querySelectorAll('select').forEach(select => {
+        Array.from(select.options).forEach(opt => (opt.selected = false));
+      });
```
Martchus: We should query `select options` directly via `querySelectorAll` instead of using a nested `forEach` and `Array.from`.

Martchus (changes requested):
After a very brief look this doesn't look too bad.

I guess the real work is now fixing all the little details that are wrong with it. So I have only commented on a few places as it would otherwise lead way too far for a code review.

Judging by test results there is also probably quite a lot wrong with it. This also needs extensive manual testing as we don't even track test coverage for JavaScript code (and even then there are probably many details that might not be caught by automatic tests, especially interactions with CSS).

Note that for this being actually worthwhile we needed to drop/replace DataTables (which is not the best library anyway but replacing it will be lots of work).

assets/javascripts/admintable.js
okurz: done

assets/javascripts/audit_log.js
okurz: ok, streamlined into a `handleJSONResponse` function

assets/javascripts/filter_form.js
okurz: I guess bootstrap is doing fine on its own *if* we use `bootstrap.Collapse`. Did that now. Seems to work fine according to my tests

assets/javascripts/filter_form.js
okurz: done

PR #6990, opened 2026-02-09
 fix(worker): reject jobs explicitly when worker is stopping

okurz:
Workers previously ignored new job assignments while finishing a previous job,
leading to stuck "assigned" jobs in the database.

Note that "$worker->is_stopping" is not a guarantee that the worker will
accomplish to stop. It just means that the current job is stopping.

This change ensures workers send a "rejected" message when in the "stopping"
state. Additionally, t/43-scheduling-and-worker-scalability.t is robustified to
handle rescheduled jobs during polling.

Fixes sporadic failures in t/43-scheduling-and-worker-scalability.t.

Related progress issue: https://progress.opensuse.org/issues/196295

### DIFF
--- a/lib/OpenQA/Worker/CommandHandler.pm
+++ b/lib/OpenQA/Worker/CommandHandler.pm
@@ -148,14 +148,12 @@ sub _can_grab_job {
     # refuse new job(s) if the worker is in an error state or stopping
     if ($worker->is_stopping) {
         $reason_to_reject_job = 'currently stopping';
-        # note: Not rejecting the job here; declaring the worker as offline which is done in any case
-        #       should be sufficient.
     }
     elsif (my $current_error = _reevaluate_and_return_current_error($client, $worker)) {
         $reason_to_reject_job = $current_error;
-        $client->reject_jobs($job_ids_to_grab, $reason_to_reject_job);
     }
     if (defined $reason_to_reject_job) {
+        $client->reject_jobs($job_ids_to_grab, $reason_to_reject_job);
         log_debug("Refusing to grab job from $webui_host: $reason_to_reject_job");
         return 0;
     }
--- a/t/24-worker-webui-connection.t
+++ b/t/24-worker-webui-connection.t
@@ -505,6 +505,7 @@ qr/Ignoring WS message from http:\/\/127\.0\.0\.1:1 with type livelog_stop and j
         [
             \%fake_error_status, $rejection->([42], 'some error'),    # status + rejection due to error state
             \%fake_error_status, $rejection->([42], 'some error'),    # status + rejection due to error state
+            $rejection->([42], 'currently stopping'),    # rejection due to stopping state
             $rejection->(['but no settings'], 'the provided job is invalid'),
             $rejection->(['42'], 'job info lacks execution sequence'),
             $rejection->(['42'], 'already busy with a job from foo'),
--- a/t/43-scheduling-and-worker-scalability.t
+++ b/t/43-scheduling-and-worker-scalability.t
@@ -220,21 +220,17 @@ subtest 'assign and run jobs' => sub {
         is_deeply([sort @allocated_worker_ids], [sort @expected_worker_ids], 'all workers allocated');
     }
     for my $try (1 .. $polling_tries_jobs) {
-        last if $jobs->search({state => DONE})->count == $job_count;
-        if ($jobs->search({state => SCHEDULED})->count > max(0, $remaining_jobs)) {
-            # uncoverable statement
-            note('At least one job has been set back to scheduled; aborting to wait until all jobs are done');
-            last;    # uncoverable statement
-        }
-        if ($remaining_jobs > 0) {
-            note("Trying to assign remaining $remaining_jobs jobs");
+        my $done_count = $jobs->search({state => DONE})->count;
+        last if $done_count == $job_count;
+        my $scheduled_count = $jobs->search({state => SCHEDULED})->count;
+        if ($scheduled_count > 0) {
+            note("Trying to assign $scheduled_count scheduled jobs");
             if (my $allocated = OpenQA::Scheduler::Model::Jobs->singleton->schedule) {
                 my $assigned_job_count = scalar @$allocated;
-                $remaining_jobs -= $assigned_job_count;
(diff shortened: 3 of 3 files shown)


lib/OpenQA/Worker/CommandHandler.pm
```diff
@@ -148,8 +148,7 @@ sub _can_grab_job {
     # refuse new job(s) if the worker is in an error state or stopping
     if ($worker->is_stopping) {
         $reason_to_reject_job = 'currently stopping';
-        # note: Not rejecting the job here; declaring the worker as offline which is done in any case
-        #       should be sufficient.
+        $client->reject_jobs($job_ids_to_grab, $reason_to_reject_job);
```
okurz: @Martchus can you comment on this? The comment "not rejecting the job here" makes it really tempting to *do* reject the job as that helps with 43-scheduling-and-worker-stability but breaks other tests

Martchus: The worker will declare itself offline if `$worker->is_stopping` is true so the assignment will be undone anyway. Maybe the test should be adjusted to deal with this.

If it is easier to reject jobs here you should at least add a comment that this is done for the sake of a specific test.

okurz: We shouldn't write out production code *just* for tests. But maybe in production we are never that quick so we wouldn't run into a condition where this makes a problem. So we could state that for "consistency" we reject jobs here or we find a better solution not needing that

Martchus: > But maybe in production we are never that quick so we wouldn't run into a condition where this makes a problem.

I would assume it is the other way around: In the test we are not patient enough for the worker to stop. Or we simulate/mock the stopping of the worker in a way that leads to a race condition.

> So we could state that for "consistency" we reject jobs here or we find a better solution not needing that

I would not be that vague. Consistency with what? These functions distinguishes *different* cases and just saying we should be consistent between those (if that's what you mean) make not much sense.

okurz: > In the test we are not patient enough for the worker to stop

Then we should change that. The test must wait for the worker then but still to not waste time we can mock or configure the worker to not wait in between. I will see what I can do

> I would not be that vague. Consistency with what?

That's why I am asking you because I am not able to phrase it less vague yet :)

Martchus: After having a closer look we established that the note is wrong and you can go though with that change after all. Avoiding the duplication like that would be preferable:

```
    # refuse new job(s) if the worker is in an error state or stopping
    if ($worker->is_stopping) {
        $reason_to_reject_job = 'currently stopping';
    }
    elsif (my $current_error = _reevaluate_and_return_current_error($client, $worker)) {
        $reason_to_reject_job = $current_error;
    }
    if (defined $reason_to_reject_job) {
        $client->reject_jobs($job_ids_to_grab, $reason_to_reject_job);
        log_debug("Refusing to grab job from $webui_host: $reason_to_reject_job");
        return 0;
    }
```

My wrong assumption here was that `$worker->is_stopping` means the worker will stop. However, it just means that the current job is stopping. Otherwise we'll end up with a situation like this that could also happen in production:

```
    # Waiting until all jobs are done, try 2
[debug] [pid:37047] Considering job 1 as regularly done
[debug] [pid:37047] REST-API call: POST "http://localhost:9526/api/v1/jobs/1/status"
[debug] [pid:37048] Considering job 2 as regularly done
[debug] [pid:37048] REST-API call: POST "http://localhost:9526/api/v1/jobs/2/status"
[debug] [pid:37047] Upload concluded (at fake)
[debug] [pid:37047] REST-API call: POST "http://localhost:9526/api/v1/jobs/1/set_done?worker_id=1" <-- worker considered free by web UI and scheduler, in the middle of processing this request
[debug] [pid:37048] Upload concluded (at fake)
[debug] [pid:37048] REST-API call: POST "http://localhost:9526/api/v1/jobs/2/set_done?worker_id=2"
[debug] [pid:37038] _carry_over_candidate(1): _failure_reason=GOOD
    # Trying to assign remaining 3 jobs
[debug] [pid:37030] Scheduling: Free workers: 1/2; Scheduled jobs: 3
[debug] [pid:37030] Need to schedule 1 parallel jobs for job 3 (with priority 50)
[debug] [pid:37030] limit reached, scheduling no additional jobs (max_running_jobs=-1, free workers=1, running=1, allocated=1)
[debug] [pid:37030] Assigned job '3' to worker ID '1'
[debug] [pid:37030] [Job#3] Prepare for being processed by worker 1
[debug] [pid:37037] Started to send message to 1 for job(s) 3
[debug] [pid:37047] Refusing to grab job from http://localhost:9526: currently stopping <-- worker still thinks it is stopping
[debug] [pid:37030] Sent job(s) '3' to worker '1'
[debug] [pid:37030] Scheduler took 0.03996s to perform operations and allocated 1 jobs
[debug] [pid:37030] Allocated: { job => 3, worker => 1 }
    # Assigned 1 more jobs: [
    #   {
    #     "job" => 3,
    #     "worker" => 1
    #   }
    # ]
    # Waiting until all jobs are done, try 3
[debug] [pid:37047] Job 1 from http://localhost:9526 finished - reason: done
[debug] [pid:37038] _carry_over_candidate(2): _failure_reason=GOOD
[debug] [pid:37048] Job 2 from http://localhost:9526 finished - reason: done
    # Trying to assign remaining 2 jobs
```

(It would resolve itself on some subsequent worker status update but that's of course nothing we should wait on.)

okurz: done. I included your reasoning in the commit message as additional note.

t/24-worker-webui-connection.t
```diff
@@ -505,6 +505,7 @@ qr/Ignoring WS message from http:\/\/127\.0\.0\.1:1 with type livelog_stop and j
         [
             \%fake_error_status, $rejection->([42], 'some error'),    # status + rejection due to error state
             \%fake_error_status, $rejection->([42], 'some error'),    # status + rejection due to error state
+            $rejection->([42], 'currently stopping'),    # rejection due to stopping
```
d3flex: picky one if you could: `# rejection due to stopping state`

okurz: done

Martchus (approved):

PR #6992, opened 2026-02-10
Add dependencies to openqa-setup-db to ensure timely startup

gauravpathak:
Related to: https://progress.opensuse.org/issues/196163

### DIFF
--- a/systemd/openqa-setup-db.service
+++ b/systemd/openqa-setup-db.service
@@ -2,7 +2,6 @@
 Description=Setup local PostgreSQL database for openQA
 Requires=postgresql.service
 After=postgresql.service network.target remote-fs.target
-Before=openqa-scheduler.service openqa-livehandler.service
 
 [Service]
 User=postgres


systemd/openqa-setup-db.service
```diff
@@ -1,7 +1,8 @@
 [Unit]
 Description=Setup local PostgreSQL database for openQA
 Requires=postgresql.service
-After=postgresql.service
+After=postgresql.service network.target remote-fs.target
```
okurz, line 4: Why does openqa-setup-db need network or anything remote?

systemd/openqa-setup-db.service
```diff
@@ -1,7 +1,8 @@
...
+Before=openqa-scheduler.service openqa-livehandler.service
```
Martchus: I don't think that this needs the scheduler and livehandler started before. It is actually the scheduler which calls the web socket service.

Note that these orderings between our own services are also not very helpful anyway because our own services have no means to signal systemd when they are actually ready to serve requests.

```suggestion
```

systemd/openqa-setup-db.service
Martchus: I don't think it needs that so this change doesn't make much sense to me.

Martchus (changes requested):
Oh, wait - this is about `openqa-setup-db.service`. I thought it was about `openqa-websockets.service` due to the referenced progress ticket. Then my previous comment still holds - we don't need these ordering dependencies.

In addition I'm now wondering now how this will solve problems with `openqa-websockets.service` on OSD.

PR #6993, opened 2026-02-10
Allow using different test distributions without symlinks

d3flex:
Added DISTRIBUTION_DIR global variable to run tests from different test
distribution. This removes the need to create symlinks for any new DISTRI
which uses the same test repo.
    
The variable supports to be set in 4 different ways. In this order
  - Job settings
  - global settings
  - settings in the worker
  - or the way that it was before aka DISTRI
    
Added resolve_distribution_dir() helper that works in both workers and API to
avoid code duplication and handle the priority chain in one place, and support
the configuration in both webui and workers.
    
If DISTRIBUTION_DIR is not set, it should use DISTRI like before, so existing
setups with symlinks should continue working. However, this could be a
breaking change for needles matching. Needles might not load correctly after
this applied. And there may be other test variables that depend on DISTRI
matching the path which it hasn't fully investigated yet.
   

related-issue: https://progress.opensuse.org/issues/195437

*The new approach is https://github.com/os-autoinst/openQA/pull/6998*

### DIFF
--- a/etc/openqa/openqa.ini
+++ b/etc/openqa/openqa.ini
@@ -1,6 +1,7 @@
 [global]
 ## Web site name for tab titles and bookmarks
 #appname = openQA
+distribution_dir = opensuse
 
 ## type of branding - [ openSUSE, plain, openqa.suse.de ]
 #branding = plain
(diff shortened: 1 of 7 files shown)


etc/openqa/openqa.ini
```diff
@@ -1,6 +1,7 @@
 [global]
 ## Web site name for tab titles and bookmarks
 #appname = openQA
+distribution_dir = opensuse
```
okurz, line 4: We do not set any defaults in config files. This might be a development left-over from you?

PR #6996, opened 2026-02-12
Switch to Leap 16.0 in all leftover places

perlpunk:


okurz (approved):

perlpunk:
> If you don't want to switch over helm charts in this PR that's fine but the ticket must only be resolved if all references to Leap 15.6 are gone, i.e. also helm charts need to be updated.

So according to https://github.com/os-autoinst/os-autoinst/pull/2803 you didn't mean **all** references to 15.6.
I removed the commit for the OBS CI workflow again.

PR #6998, opened 2026-02-13
Allow using different test distributions without symlinks

d3flex:
Added DISTRIBUTION_DIR global variable to run tests from different test distribution. This removes the need to create symlinks for any new DISTRI which uses the same test repo.

The variable supports to be set the CASEDIR if this is undefined.

If DISTRIBUTION_DIR is not set, it should use DISTRI like before, so existing setups with symlinks should continue working.
This commit could be a potential breaking change for needles matching. And there may be other circumstance which they are not very obvious yet, However the use of CASEDIR and the way the change approach the solution should minimize the possibility of regressions.

related-issue: https://progress.opensuse.org/issues/195437

### DIFF
--- a/etc/openqa/openqa.ini
+++ b/etc/openqa/openqa.ini
@@ -1,6 +1,8 @@
 [global]
 ## Web site name for tab titles and bookmarks
 #appname = openQA
+## Set the default CASEDIR if the distribution directory/symlink doesn't already exist
+#distribution_dir = opensuse
 
 ## type of branding - [ openSUSE, plain, openqa.suse.de ]
 #branding = plain
--- a/lib/OpenQA/Setup.pm
+++ b/lib/OpenQA/Setup.pm
@@ -80,6 +80,7 @@ sub read_config ($app) {
     my %defaults = (
         global => {
             appname => 'openQA',
+            distribution_dir => 'opensuse',
             base_url => undef,
             branding => 'openSUSE',
             download_domains => undef,
--- a/lib/OpenQA/Utils.pm
+++ b/lib/OpenQA/Utils.pm
@@ -206,12 +206,15 @@ sub testcasedir ($distri = undef, $version = undef, $rootfortests = undef) {
         $rootfortests ||= $dir if -d $dir;
     }
     $rootfortests ||= $defaultroot;
-    $distri //= '';
-    # TODO actually "distri" is misused here. It should rather be something
-    # like the name of the repository with all tests
-    my $dir = catdir($rootfortests, $distri);
-    $dir .= "-$version" if $version && -e "$dir-$version";
-    return $dir;
+    try {
+        my $dir = catdir($rootfortests, $distri);
+        $dir .= "-$version" if $version && -e "$dir-$version";
+        log_info "testcasedir path exists at $dir" if (-e $dir || -l $dir);
+        return $dir;
+    }
+    catch ($e) {
+        log_error "testcasdir failed using '$distri': $e";
+    }
 }
 
 =head2 git_commit_url
--- a/docs/WritingTests.asciidoc
+++ b/docs/WritingTests.asciidoc
@@ -2296,7 +2296,9 @@ follows:
 * If `CASEDIR` or `NEEDLES_DIR` is customized the customized location is used
   instead of the default repository.
 * If only one of `CASEDIR` or `NEEDLES_DIR` is customized the other variable
-  will still be initialized to point to the default repository.
+  will still be initialized to point to the default repository. In case of 
+  undefined `CASEDIR` the variable will attempt to set the value of 
+  `DISTRIBUTION_DIR` from the openQA global settings.
 * A relative `NEEDLES_DIR` is treated to be relative to the default `CASEDIR`
   (even if `CASEDIR` is customized). To have it treated to be relative to the
   custom `CASEDIR`, prefix the relative path with `%CASEDIR%/`. So specifying
--- a/lib/OpenQA/Schema/Result/Jobs.pm
+++ b/lib/OpenQA/Schema/Result/Jobs.pm
@@ -370,7 +370,18 @@ sub prepare_for_work ($self, $worker = undef, $worker_properties = {}) {
     my $job_token = $worker_properties->{JOBTOKEN} // random_string();
     $worker->set_property(JOBTOKEN => $job_token);
     $job_hashref->{settings}->{JOBTOKEN} = $job_token;
-
+    unless (exists $job_hashref->{settings}->{CASEDIR}) {
+        if (my $dist_dir = OpenQA::App->singleton->config->{global}->{distribution_dir}) {
+            my $distri = $job_hashref->{settings}->{DISTRI};
+            my $version = $job_hashref->{settings}->{VERSION};
+            my $testdir_found = testcasedir($distri, $version);
+            if ($testdir_found && !-e $testdir_found) {
+                log_debug
+                  "CASEDIR is not set and test directory for DISTRI=$distri not found, CASEDIR will use $dist_dir";
+                $job_hashref->{settings}->{CASEDIR} = $dist_dir;
+            }
+        }
+    }
     my $updated_settings = $self->register_assets_from_settings();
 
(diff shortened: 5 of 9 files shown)


etc/openqa/openqa.ini
```diff
@@ -1,6 +1,8 @@
 [global]
 ## Web site name for tab titles and bookmarks
 #appname = openQA
+## Set the default CASEDIR if the distribution directory/symlink doesn't already exist
+#distribution_dir = opensuse
```
Martchus, line 5: Having to add another config setting for this seems wrong/needles. I wouldn't change what `CASEDIR` is set to by default unless that other variable is used.

perlpunk: I stared at the `wrong/needles` for a long time until I understood that it's meant to be `wrong/needless` :D

perlpunk: Also I'm wondering again how you imagine a global variable per instance.
So let's say we set this on o3. Probably you would set it to `opensuse`. How would the obs tests work then? They also don't set CASEDIR and rely on the existing directory. Suddenly their CASEDIR would be `opensuse` instead of `obs`.

d3flex: As far as I can say nothing should change for the tests that set CASEDIR. I didnt consider the `obs` tests and I do not know what would be different between obs and opensuse, but if some tests needs a particular testrepo it would have to be defined explicitly. To give you a concrete answer I would have to test it I guess. But I do not see really the problem.

perlpunk: The way I understood this feature is (if we decide for the CASEDIR approach):
Tests for Aeon (for example)
*  would set `DISTRI=aeon` which they already do
* set `CASEDIR=opensuse` so it would work automatically without having a symlink

That means any new DISTRI would have to set `CASEDIR=opensuse` to make this work, but that seems like a requirement that is easy to fulfill.

Your global `distribution_dir` instead: I assume  you would set it to `opensuse` on o3.
Then any test that sets no CASEDIR (e.g. https://openqa.opensuse.org/group_overview/63 OBS) would suddenly use opensuse as a CASEDIR because of the global `distribution_dir`.
So now any existing test like those would stop working.

perlpunk: > To give you a concrete answer I would have to test it I guess.

Yes please. That's the task of the ticket assignee I would say.

Martchus: @d3flex I understood this feature in the same way as @perlpunk and see the same problem with your approach.

d3flex: I checked o3:
```ybonatakis@ariel:~> ls -la /var/lib/openqa/share/tests/
total 16
drwxr-xr-x  9 geekotest root     286 Feb 16 14:33 .
drwxr-xr-x  5 root      root      95 May 23  2022 ..
lrwxrwxrwx  1 geekotest nogroup    8 Jan 30 12:21 aeon -> opensuse
lrwxrwxrwx  1 geekotest nogroup    8 Jun 15  2022 alp -> opensuse
lrwxrwxrwx  1 geekotest nogroup    8 Apr  5  2022 d-installer -> opensuse
drwxr-xr-x  8 geekotest nogroup 4096 Sep 23  2024 example
lrwxrwxrwx  1 root      root       8 Sep  4  2017 kubic -> opensuse
lrwxrwxrwx  1 root      root       8 Jun 24  2022 leap-micro -> opensuse
lrwxrwxrwx  1 geekotest nogroup    8 Apr 16  2019 microos -> opensuse
drwxr-xr-x 15 geekotest nogroup 4096 Sep  5 07:25 none
drwxr-xr-x  5 geekotest nogroup  121 Aug  5  2024 obs
drwxr-xr-x  8 geekotest nogroup  294 Dec 10 15:54 openqa
lrwxrwxrwx  1 root      root       6 Apr 16  2024 openQA -> openqa
drwxr-xr-x 15 geekotest nogroup 4096 Feb 16 11:18 opensuse
drwxr-xr-x  4 geekotest nogroup   65 Apr 16  2020 repositories
drwxr-xr-x 15 geekotest nogroup 4096 Dec  9 12:00 sle
lrwxrwxrwx  1 root      root       8 Jul 22  2016 windows -> opensuse
```

That mean that nothing breaks on `obs`
gonna run a test in addition. stay tuned

Martchus: Before you follow this approach too far: It would nevertheless be good to know whether it is really not possible to just rely on `CASEDIR` avoiding a 2nd variable that is almost the same and therefore potentially causes confusion.

d3flex: I copied a job like that
`❯ openqa-clone-job --skip-chained-deps _GROUP_ID=0 _GROUP=0 _PRIORITY=0 BUILD=":iob" TEST+="-iob-obs-test-new2" --from https://openqa.opensuse.org/tests/5669094`

I copied the `obs` in the /var/lib and in the logs I saw `[info] testcasedir path exists at /home/iob/openqadev/varlib/openqa/share/tests/obs` on worker and 
```[2026-02-16T18:13:01.617718+01:00] [debug] [pid:3370571] git hash in 'obs/needles': 25da2e536ed4dd6b834a0d6abf1d8569478cdff5
21	[2026-02-16T18:13:01.618056+01:00] [debug] [pid:3370571] git url in 'obs/needles': "UNKNOWN (no .git found)"
``` 
in autoinst-log.txt

I have to do some tweaks and fix some import but other than that, it worked as expected. Except that it didnt reproduce the original failure.

>  It would nevertheless be good to know whether it is really not possible to just rely on CASEDIR avoiding a 2nd variable that is almost the same and therefore potentially causes confusion.

Everything is possible right? it is more what's your trade-off.

Without `distribution_dir`, the trade off is that the jobs will require a CASEDIR if i understand your idea. 

But I think it doesnt need to as I assume that like opensuse, every distro has one common testdir and

With a `distribution_dir` everything is almost the same for:
- jobs with set CASEDIR
- jobs which have dedicated test_repo in /var/lib/openqa/share/tests
- jobs which depends on the "default" repo (using symlink)

For my POV, this release some burden from the user.

lib/OpenQA/Utils.pm
```diff
@@ -206,12 +206,15 @@ sub testcasedir ($distri = undef, $version = undef, $rootfortests = undef) {
         $rootfortests ||= $dir if -d $dir;
     }
     $rootfortests ||= $defaultroot;
-    $distri //= '';
-    # TODO actually "distri" is misused here. It should rather be something
-    # like the name of the repository with all tests
-    my $dir = catdir($rootfortests, $distri);
-    $dir .= "-$version" if $version && -e "$dir-$version";
-    return $dir;
+    try {
```
perlpunk, line 209: what kind of exception do you expect here in this block?

lib/OpenQA/Setup.pm
```diff
@@ -80,6 +80,7 @@ sub read_config ($app) {
     my %defaults = (
         global => {
             appname => 'openQA',
+            distribution_dir => 'opensuse',
```
perlpunk, line 83: so when this is released, every openQA instance, also not used for openSUSE, would have 'opensuse' as a default? is that intended?

lib/OpenQA/Utils.pm
d3flex: I used it because of `catdir` and to print a different log in case something happen, but it might not have real use here. I will remove it

lib/OpenQA/Setup.pm
d3flex: that was the idea, even thought I didnt thought about that thoughtfully. This PR is still a draft and WIP. We can have it `opensuse`, _empty_ ` or undef` as well, as far it can be controlled by configuration, right?

okurz: openQA should definitely not default anything to `opensuse` as openQA is independent of ~openQA~ openSUSE.

perlpunk: > openQA should definitely not default anything to `opensuse` as openQA is independant of openQA.

You probably meant "openQA is independent of openSUSE".

PR #7000, opened 2026-02-16
Refactor helm, make it work with Ingress, and fix worker connectivity issues

d3flex:
- Remove old helm chart structure
- Add restructured openqa parent chart with ingress and nginx
- Move worker subchart under openqa/ and fix connectivity
- Update Helm chart documentation

related-issue: https://progress.opensuse.org/issues/183608 https://progress.opensuse.org/issues/189807


container/helm/openqa/templates/_helpers.tpl
```diff
@@ -1,16 +1,16 @@
...
-{{- define "webui.name" -}}
+{{- define "openqa.name" -}}
 {{- default .Chart.Name .Values.nameOverride | trunc 63 | trimSuffix "-" }}
 {{- end }}
 
 {{/*
 Create a default fully qualified app name.
 We truncate at 63 chars because some Kubernetes name fields are limited to this (by the DNS naming spec).
-If the release name contains a chart name it will be used as a full name.
+If release name contains chart name it will be used as a full name.
```
Martchus: The use of articles was correct and your version is now grammatically incorrect. So please revert.

container/helm/openqa/Chart.yaml
```diff
@@ -1,19 +1,16 @@
...
 dependencies:
-  - name: webui
-    repository: file://../webui/
-    version: ~0.1.0
-    condition: webui.enabled
   - name: postgresql
-    version: 15.5.38
+    version: 18.0.7
     repository: https://charts.bitnami.com/bitnami
     condition: postgresql.enabled
```
Martchus, line 12: I remember we struggled in the past with upgrading this container as it was no longer supported. So have things changed?

container/helm/README.md
```diff
@@ -13,41 +13,90 @@ a parent chart, _openqa_.
...
+helm install openqa openqa/ -f my_values.yaml
 ```
 
 ### Worker
 
 The worker requires some basic configuration, as described in the
-[documentation](http://open.qa/docs/#_run_openqa_workers). You can also set up
-a [cache service](http://open.qa/docs/#asset-caching), which may improve
-performance when handling assets, tests, and needles.
+[documentation](http://open.qa/docs/#_run_openqa_workers).
```
Martchus, line 127: So the cache service is setup automatically? Or is this no longer possible? Or was this never working anyway?

container/helm/openqa/Chart.yaml
d3flex: In simple words the containers. I cant remember where exactly was failing, but we were trying to update the postgres version while we keep using leap15.6 container for webui. There was a conflict, something with the schema, if my notes are correct, the problem could happen with either postgresql version 15 or 16. I think at the moment the versions above that, were not available in the distrubution. Cant say for sure - that might be a memory from the summer -

perlpunk: https://progress.opensuse.org/issues/189807 Adapt Helm Chart after Bitnami changed their PostgreSQL container and reintroduce according CI checks

container/helm/README.md
d3flex: No. or it depends what do you mean? the setup is there but under condition
```
      # Future enhancement option: initContainer for git-pulling tests into /data
      containers:
        {{ if .Values.cacheService -}}
        - name: {{ include "worker.fullname" . }}-cacheservice
 ```

this is how it was even before. And I assume it works. The only reason I removed the documentation is that I didnt experiment with this, so I left out to keep what I really know it works, in case you follow the instruction.

File: container/helm/openqa/Chart.yaml (hunk 2 above)
Comment (perlpunk) [reply container/helm/openqa/Chart.yaml:12]: I think this PR should only be merged if it actually works. Currently we only lint, but we don't test that the helm charts are building. We skip the test. So we have no proof that your change actually works

Comment (d3flex) [reply container/helm/openqa/Chart.yaml:12]: I tested it a couple of times and it was ok. I even was able to run a job(failed for other reasons). So you can just try and tell me

Comment (d3flex) [reply container/helm/openqa/Chart.yaml:12]: > https://progress.opensuse.org/issues/189807 Adapt Helm Chart after Bitnami changed their PostgreSQL container and reintroduce according CI checks

I will just link the PR but I do know if we really want another ticket for the helms. or do you want me to mention the progress on this PR?

Comment (perlpunk) [reply container/helm/openqa/Chart.yaml:12]: > I tested it a couple of times and it was ok.

I believe you, but that's not how we ensure something is working.

> So you can just try and tell me

tell you what?

I wrote before that we SKIP the check-helm-chart / install-chart at the moment. Because of https://progress.opensuse.org/issues/189807 .
So the first thing is to fix the charts and enable that test again.

In this PR the helm charts are only linted, but install-chart is disabled. So how should we know whether this works? Isn't that what we have CI checks for?

Comment (perlpunk) [reply container/helm/openqa/Chart.yaml:12]: So how about you enable the install-chart check again as part of this PR and we'll see if it works?

Comment (perlpunk) [reply container/helm/openqa/Chart.yaml:12]: > I will just link the PR but I do know if we really want another ticket for the helms

I don't understand that sentence.

**Review by perlpunk — CHANGES_REQUESTED** [review]
The install-chart check needs to be enabled in order to see if this works

File: container/helm/openqa/Chart.yaml (hunk 2 above)
Comment (Martchus) [reply container/helm/openqa/Chart.yaml:12]: > was ok.

"was ok" like "did not format my root file system"?

**Comment by d3flex** [comment]
> The install-chart check needs to be enabled in order to see if this works

I enabled it. But the CI is different story for me

File: container/helm/openqa/Chart.yaml (hunk 2 above)
Comment (d3flex) [reply container/helm/openqa/Chart.yaml:12]: > "was ok" as in "did not format my root file system"?

in what way? the containers run in a kubernetes cluster.

File: container/helm/README.md [hunk 4]
```diff
@@ -13,41 +13,90 @@ a parent chart, _openqa_.
 
 ## Installation
 
-To install openQA, update helm dependencies and install the parent chart.
-
+Make sure, first that a Kubernetes cluster is running
+```bash
+minikube start
+minikube addons enable ingress
```
okurz: isn't that relying on nginx ingress which is obsolete? https://www.kubernetes.dev/blog/2025/11/12/ingress-nginx-retirement/

d3flex: indeed. However we still can merge this and follow up

okurz: I am not sure. We don't know when we have the time to follow-up. Apparently #7008 fixes tests for now so I prefer to have smaller changes to fix what we have right now and rebase the bigger restructuring on top. And I checked the code changed and it looks like rebasing on top of #7008 is trivial. As you already applied the according changes but moved files you can just forcefully checkout your version.

container/helm/README.md
```diff
@@ -13,41 +13,115 @@ a parent chart, _openqa_.
...
+
+The chart uses the [Kubernetes Gateway API](https://gateway-api.sigs.k8s.io/)
+for external access. This has to be installed once before the provision of the
+cluster. Install the Gateway API CRDs and
+[Envoy Gateway](https://gateway.envoyproxy.io/) controller:
+
+```bash
+kubectl apply -f https://github.com/kubernetes-sigs/gateway-api/releases/download/v1.2.1/standard-install.yaml --server-side
+helm install eg oci://docker.io/envoyproxy/gateway-helm --version v1.4.1 \
+  -n envoy-gateway-system --create-namespace --skip-crds
```
okurz: Does this have to be version specific? This will outdate soon

container/helm/README.md
```diff
@@ -13,41 +13,115 @@ a parent chart, _openqa_.
...
+```
+
+Add the gateway address to `/etc/hosts`, pointing to the hostname configured
+in `values.yaml` (`baseUrl`):
+
+```
+<GATEWAY_ADDRESS> openqa.host.org
+```
+
+Then access the UI at http://openqa.host.org.
```
okurz: host.org is not a conventional internal-only domain. I recommend to use a domain that is reserved for internal use. https://www.rfc-editor.org/rfc/rfc6762#appendix-G mentions multiple examples. How about .internal.?

d3flex: please  lets merge this and I will make this change. But not in this PR

container/helm/README.md
d3flex: No it is the one I used. So who ever follow this guide will reproduce the same. And here, I would like to remind you that it would be nice to merge and improve later. Do we have to address everything here, when we can have smaller manageable commits for minor things?

okurz: Having smaller manageable commits is good. You can add commits on top, no problem with that. I would prefer if we don't introduce such version specific call that will soon be obsolete.

container/helm/README.md
okurz: Why not just do a simple search and replace?

okurz (changes requested):
see still open and unanswered discussions

container/helm/README.md
```diff
@@ -13,41 +13,115 @@ a parent chart, _openqa_.
...
+
+### Gateway API setup
+
+The chart uses the [Kubernetes Gateway API](https://gateway-api.sigs.k8s.io/)
+for external access. This has to be installed once before the provision of the
+cluster. Install the Gateway API CRDs and
+[Envoy Gateway](https://gateway.envoyproxy.io/) controller:
+
+```bash
+kubectl apply -f https://github.com/kubernetes-sigs/gateway-api/releases/download/v1.2.1/standard-install.yaml --server-side
```
okurz: there is still a version pin to a patch-level version

okurz (approved):

PR #7001, opened 2026-02-16
Improve `enqueue_and_keep_track`

Martchus:
See individual commit messages and https://progress.opensuse.org/issues/178954.

### DIFF
--- a/codecov.yml
+++ b/codecov.yml
@@ -24,8 +24,7 @@ coverage:
           - lib/OpenQA/Worker.pm
           - lib/OpenQA/Utils.pm
           - lib/OpenQA/WebAPI/Controller/
-          - lib/OpenQA/Shared/Controller/Running.pm
-          - lib/OpenQA/Shared/Controller/Session.pm
+          - lib/OpenQA/Shared/
       tests:
         target: 100.0
         threshold: 0


lib/OpenQA/Shared/Plugin/Gru.pm
```diff
@@ -312,33 +309,21 @@ sub enqueue_and_keep_track {
 
     # keep track of the Minion job and continue rendering if it has completed
     return $self->app->minion->result_p($minion_id, {interval => TRACK_INTERVAL})->then(
-        sub {
-            my ($info) = @_;
-
-            unless (ref $info) {
-                return Mojo::Promise->reject({error => "Minion job for $task_description has been removed."});
-            }
+        sub (@results) {
```
perlpunk, line 312: It looks like this could be:
```suggestion
    sub ($info, @) {
```
and the other sub below 
```suggestion
    sub ($result, @) {
```

Martchus: The `@results` array can unfortunately be completely empty (meaning no parameters are passed at all). That is the case for the first sub when the Minion job was removed completely. I guess for the 2nd sub I *could* change it as that sub probably gets at least one argument but I'm not 100 % sure of that (as whatever is passed here might come directly from the Minion framework).

perlpunk: ah ok, it's not very obvious from the code that this can be empty as it explicitly checks the first item...

PR #7002, opened 2026-02-16
feat: Make allowed hosts for SCENARIO_DEFINITIONS_YAML_FILE configurable

Martchus:
So far we only allow downloads from `raw.githubusercontent.com`. With this change we also allow downloads from `github.com`. That makes writing GitHub actions without hardcoding URLs easier because only the latter is available as variable (`GITHUB_SERVER_URL`).

With this change the list of allowed hosts is now also configurable and the fact that not all hosts are allowed by default is documented.

Related ticket: https://progress.opensuse.org/issues/195584

### DIFF
--- a/lib/OpenQA/Task/Iso/Schedule.pm
+++ b/lib/OpenQA/Task/Iso/Schedule.pm
@@ -18,12 +18,13 @@ sub register ($self, $app, @) {
     $app->minion->add_task(schedule_iso => sub { _schedule_iso($app, @_) });
 }
 
-sub _download_scenario_definitions ($minion_job, $scheduled_product, $scheduling_params) {
+sub _download_scenario_definitions ($app, $minion_job, $scheduled_product, $scheduling_params) {
     return 1 unless my $file = $scheduling_params->{SCENARIO_DEFINITIONS_YAML_FILE};
     my $url = Mojo::URL->new($file);
     my $scheme = $url->scheme;
     return 1 unless $scheme eq 'http' || $scheme eq 'https';
-    return 1 unless $url->host eq 'raw.githubusercontent.com';    # restrict this for now to GitHub
+    state $allowed_hosts = {map { ($_ => 1) } split ' ', $app->config->{global}->{scenario_definitions_allowed_hosts}};
+    return 1 unless $allowed_hosts->{$url->host};
     my $ua = Mojo::UserAgent->new(SCENARIO_DEFS_YAML_UA_ARGS);
     my $tx = $ua->get($url);
     if (my $err = $tx->error) {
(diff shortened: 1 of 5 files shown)


Martchus:
I removed the use of `state` to avoid `[  374s] Initialization of state variables in list context currently forbidden at /home/abuild/rpmbuild/BUILD/openQA-5.1771257763.962f0749/lib/OpenQA/Task/Iso/Schedule.pm line 26, near "};"` for SLE15 builds.

lib/OpenQA/Task/Iso/Schedule.pm
```diff
@@ -18,12 +18,13 @@ sub register ($self, $app, @) {
...
 
-sub _download_scenario_definitions ($minion_job, $scheduled_product, $scheduling_params) {
+sub _download_scenario_definitions ($app, $minion_job, $scheduled_product, $scheduling_params) {
     return 1 unless my $file = $scheduling_params->{SCENARIO_DEFINITIONS_YAML_FILE};
     my $url = Mojo::URL->new($file);
     my $scheme = $url->scheme;
     return 1 unless $scheme eq 'http' || $scheme eq 'https';
-    return 1 unless $url->host eq 'raw.githubusercontent.com';    # restrict this for now to GitHub
+    my %allowed_hosts = map { ($_ => 1) } split ' ', $app->config->{global}->{scenario_definitions_allowed_hosts};
+    return 1 unless $allowed_hosts{$url->host};    # restrict this for now to GitHub
```
perlpunk:
```suggestion
    return 1 unless $allowed_hosts{$url->host};    
```

lib/OpenQA/Task/Iso/Schedule.pm
```diff
@@ -18,12 +18,13 @@ sub register ($self, $app, @) {
...
 
-sub _download_scenario_definitions ($minion_job, $scheduled_product, $scheduling_params) {
+sub _download_scenario_definitions ($app, $minion_job, $scheduled_product, $scheduling_params) {
     return 1 unless my $file = $scheduling_params->{SCENARIO_DEFINITIONS_YAML_FILE};
     my $url = Mojo::URL->new($file);
     my $scheme = $url->scheme;
     return 1 unless $scheme eq 'http' || $scheme eq 'https';
-    return 1 unless $url->host eq 'raw.githubusercontent.com';    # restrict this for now to GitHub
+    my %allowed_hosts = map { ($_ => 1) } split ' ', $app->config->{global}->{scenario_definitions_allowed_hosts};
+    return 1 unless $allowed_hosts{$url->host};
```
perlpunk: `state` would would work with a hashref:
```suggestion
    state $allowed_hosts = { map { ($_ => 1) } split ' ', $app->config->{global}->{scenario_definitions_allowed_hosts} };
    return 1 unless $allowed_hosts->{$url->host};
```

PR #7003, opened 2026-02-17
fix(43-scheduling-and-worker-scalability): prevent sporadic issues

okurz:
Change the verification logic to move from the fragile synchronous
expectations to an eventually consistent model by checking the
corresponding final states.

- Shift to state-based verification: Check job states in the database
  (ASSIGNED, RUNNING, or DONE) instead of relying on the transient return
  value of a single $scheduler->schedule call.
- Implement polling retry loops for job assignment to robustly handle
  transient WebSocket connection delays.
- Ensure the test verifies the expected system state reaching
  the desired configuration within a reasonable timeout.

Together with the commits before this ensures a fully stable test now:

count-fail-ratio: Run: 200. Fails: 0. Fail ratio 0±0%. No fails, computed failure probability < 1.50%
mean runtime: 2952±66.71 ms

Related progress issue: https://progress.opensuse.org/issues/196295

Next:
* #7004

### DIFF
--- a/t/43-scheduling-and-worker-scalability.t
+++ b/t/43-scheduling-and-worker-scalability.t
@@ -148,87 +148,61 @@ my $polling_tries_workers = $seconds_to_wait_per_worker / $polling_interval * $w
 my $polling_tries_jobs = $seconds_to_wait_per_job / $polling_interval * $job_count;
 
 subtest 'wait for workers to be idle' => sub {
-    # wait for all workers to register
-    my @worker_search_args = ({'properties.key' => 'WEBSOCKET_API_VERSION'}, {join => 'properties'});
+    # wait for all workers to register, have correct API version and be idle
+    # this ensures they are fully visible to the scheduler
     my $actual_count = 0;
     for my $try (1 .. $polling_tries_workers) {
-        last if ($actual_count = $workers->search(@worker_search_args)->count) == $worker_count;
-        note("Waiting until all workers are registered, try $try");    # uncoverable statement
+        my @idle
+          = grep { $_->status eq 'idle' && ($_->websocket_api_version || 0) == WEBSOCKET_API_VERSION } $workers->all;
+        $actual_count = scalar @idle;
+        last if $actual_count == $worker_count;
+        note("Waiting until all workers are registered and idle, try $try");    # uncoverable statement
         sleep $polling_interval;    # uncoverable statement
     }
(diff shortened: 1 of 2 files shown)


okurz:
> Have you tested all three cases by setting the env variables that test uses? (worker count = job count, worker count > job count, worker count < job count)
> 
> Have you tested with a high number of jobs and workers?

No and no.

okurz:
> Have you tested all three cases by setting the env variables that test uses? (worker count = job count, worker count > job count, worker count < job count)
> 
> Have you tested with a high number of jobs and workers?

I did both now. We should definitely try to cover that automatically. Trying …

t/43-scheduling-and-worker-scalability.t
```diff
@@ -148,87 +148,61 @@ my $polling_tries_workers = $seconds_to_wait_per_worker / $polling_interval * $w
...
-    # wait for all workers to register
-    my @worker_search_args = ({'properties.key' => 'WEBSOCKET_API_VERSION'}, {join => 'properties'});
+    # wait for all workers to register, have correct API version and be idle
+    # this ensures they are fully visible to the scheduler
     my $actual_count = 0;
     for my $try (1 .. $polling_tries_workers) {
-        last if ($actual_count = $workers->search(@worker_search_args)->count) == $worker_count;
-        note("Waiting until all workers are registered, try $try");    # uncoverable statement
+        my @idle
+          = grep { $_->status eq 'idle' && ($_->websocket_api_version || 0) == WEBSOCKET_API_VERSION } $workers->all;
```
perlpunk, line 156: I don't get from the description why you changed from `->count` to `->all` and then only get the number of items in the array after all.

okurz: With "description" do you mean the PR description, the commit description or the comment above in lines 151-152?

So in before we relied on the SQL `COUNT(*)` for each worker that registered the WEBSOCKET_API_VERSION property. The new logic verifies that workers are actually idle additionally.

Martchus: And because `status` is unfortunately not just a column but a computation done in Perl we need to retrieve the rows.

perlpunk: I meant the commit message.
But now I see it, sorry. Let's just say it was late when I reviewed it :)

PR #7006, opened 2026-02-17
feat: Allow specifying `CASEDIR` to avoid symlinking

Martchus:
Tested by cloning an Aeon job locally without having symlinking in place.

When just cloning the job it fails to locate tests as it assumes the directory `aeon` as per `DISTRI`. That was expected.

The idea of this change is to be able to set e.g. `CASEDIR=opensuse` in cases like this instead of creating a symlink. To test this I cloned an Aeaon job locally again, this time with that variable set:

```
openqa-clone-job https://openqa.opensuse.org/tests/5677313 CASEDIR=opensuse
```

With this I ran into the problem that the worker does not do the required symlinking of osado into the pool dir based on `CASEDIR`. It just looks at `DISTRI`. That is therefore what this change adjusts. After cloning the job again with this small change in place the test was able to start.

Note that the job I've cloned already specifies `NEEDLES_DIR` pointing to a Git repository. So combination with a Git repo for needles works.

I've also tested cloning with `NEEDLES_DIR=` (clearing the variable). Then the worker symlinked osado needles which is reasonable and in-line with the documentation.

I also tested cloning with `NEEDLES_DIR=%CASEDIR%/products/aeon` to cover the case if one wanted to use a distinct needle repo for Aeon but still not specify a Git URL. With this the worker created a symlink to `products/aeon` within the `opensuse` `CASEDIR`. So that also works as expected and is also in-line with the documentation.

Related ticket: https://progress.opensuse.org/issues/195437

### DIFF
--- a/lib/OpenQA/Worker/Engines/isotovideo.pm
+++ b/lib/OpenQA/Worker/Engines/isotovideo.pm
@@ -390,12 +390,14 @@ sub _engine_workit_step_2 ($job, $job_settings, $vars, $shared_cache, $callback)
 
     # ensure a CASEDIR and a PRODUCTDIR is assigned and create a symlink if required
     my $absolute_paths = $vars->{ABSOLUTE_TEST_CONFIG_PATHS};
-    my @vars_for_default_dirs = ($vars->{DISTRI}, $vars->{VERSION}, $shared_cache);
+    my $casedir = $vars->{CASEDIR};
+    my $effective_distri = defined $casedir && !looks_like_url_with_scheme($casedir) ? $casedir : $vars->{DISTRI};
+    my @vars_for_default_dirs = ($effective_distri, $vars->{VERSION}, $shared_cache);
     my $default_casedir = testcasedir(@vars_for_default_dirs);
     my $default_productdir = productdir(@vars_for_default_dirs);
     my $target_name = path($default_casedir)->basename;
-    my $has_custom_dir = $vars->{CASEDIR} || $vars->{PRODUCTDIR};
-    my $casedir = $vars->{CASEDIR} //= $absolute_paths ? $default_casedir : $target_name;
+    my $has_custom_dir = $casedir || $vars->{PRODUCTDIR};
+    $casedir = $vars->{CASEDIR} //= $absolute_paths ? $default_casedir : $target_name;
     unless (_is_job_only_relying_on_git($vars)) {
         if ($casedir eq $target_name) {
             $vars->{PRODUCTDIR} //= substr($default_productdir, rindex($default_casedir, $target_name));
--- a/t/24-worker-engine.t
+++ b/t/24-worker-engine.t
@@ -466,7 +466,20 @@ subtest 'symlink testrepo, logging behavior, variable expansion' => sub {
         my $vars_data = get_job_json_data($pool_directory);
         is $vars_data->{NEEDLES_DIR}, 'needles', 'relative NEEDLES_DIR is set to name of symlink';
         is $result->{error}, undef, 'no error occurred (4)';
+    };
 
+    subtest 'good case: custom "default" CASEDIR specified overriding DISTRI' => sub {
+        my %settings = (DISTRI => 'opensuse', CASEDIR => 'fedora', NEEDLES_DIR => 'https://foo', JOBTOKEN => 'token');
+        my ($job, $result) = OpenQA::Worker::Job->new($worker, $client, {id => 12, settings => \%settings});
+        combined_like { $result = _run_engine($job) }
+        qr {Symlinked from "t/data/openqa/share/tests/fedora" to "$pool_directory/fedora"},
+          'symlink for CASEDIR=fedora created, taking precedence over DISTRI=opensuse';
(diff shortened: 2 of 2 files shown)


lib/OpenQA/Worker/Engines/isotovideo.pm
```diff
@@ -390,7 +390,7 @@ sub _engine_workit_step_2 ($job, $job_settings, $vars, $shared_cache, $callback)
 
     # ensure a CASEDIR and a PRODUCTDIR is assigned and create a symlink if required
     my $absolute_paths = $vars->{ABSOLUTE_TEST_CONFIG_PATHS};
-    my @vars_for_default_dirs = ($vars->{DISTRI}, $vars->{VERSION}, $shared_cache);
+    my @vars_for_default_dirs = ($vars->{CASEDIR} // $vars->{DISTRI}, $vars->{VERSION}, $shared_cache);
```
perlpunk: Probably we should check for `^http` in CASEDIR first.

But that approach was what I came up with as well, basically, when I tried this.

Martchus: Maybe. I'll test how I can "break" it to see whether we need more checks. Maybe in case of a Git repo this is not used anyway.

perlpunk: I wonder if we should check at least that it doesn't contain an absolute path.
I sometimes used this when running tests in my local instance, but it doesn't seem nice to allow everybody to specify an arbitrary directory there.

Martchus: I added this kind of check as it is indeed needed.

lib/OpenQA/Worker/Engines/isotovideo.pm
```diff
@@ -390,12 +390,14 @@ sub _engine_workit_step_2 ($job, $job_settings, $vars, $shared_cache, $callback)
...
+    my $casedir = $vars->{CASEDIR};
+    my $effective_distri = defined $casedir && !looks_like_url_with_scheme($casedir) ? $casedir : $vars->{DISTRI};
+    my @vars_for_default_dirs = ($effective_distri, $vars->{VERSION}, $shared_cache);
     my $default_casedir = testcasedir(@vars_for_default_dirs);
     my $default_productdir = productdir(@vars_for_default_dirs);
     my $target_name = path($default_casedir)->basename;
-    my $has_custom_dir = $vars->{CASEDIR} || $vars->{PRODUCTDIR};
-    my $casedir = $vars->{CASEDIR} //= $absolute_paths ? $default_casedir : $target_name;
+    my $has_custom_dir = $casedir || $vars->{PRODUCTDIR};
+    $casedir = $vars->{CASEDIR} //= $absolute_paths ? $default_casedir : $target_name;
```
d3flex, line 400: Can we use a debug log here? or do you think it will be overkill?

Martchus: I'd say it is an overkill. At least I didn't need that when debugging. One can just look at what the worker did (symlinks in the pool directory) and `vars.json`. If a hard error occurs it already ends up on the worker logs.

d3flex: I was thinking for debugging purpose. when fails, it leaves some trace in one or other way. But when I can say from my tests, that there might be not so clear when the test can fail for not obvious reason. Neither the worker or the os-autoinst logs anything about the casedir. All this happens in silence. it could be even at the trace level if you like.

Martchus: > Neither the worker or the os-autoinst logs anything about the casedir. All this happens in silence. it could be even at the trace level if you like.

That's not true. The Git hash the the checkouts is logged by os-autoinst. The worker will also complain if the specified directory does not exist.

t/24-worker-engine.t
```diff
@@ -466,7 +466,20 @@ subtest 'symlink testrepo, logging behavior, variable expansion' => sub {
         my $vars_data = get_job_json_data($pool_directory);
         is $vars_data->{NEEDLES_DIR}, 'needles', 'relative NEEDLES_DIR is set to name of symlink';
         is $result->{error}, undef, 'no error occurred (4)';
+    };
 
+    subtest 'good case: custom "default" CASEDIR specified overriding DISTRI' => sub {
+        my %settings = (DISTRI => 'opensuse', CASEDIR => 'fedora', NEEDLES_DIR => 'https://foo', JOBTOKEN => 'token');
+        my ($job, $result) = OpenQA::Worker::Job->new($worker, $client, {id => 12, settings => \%settings});
+        combined_like { $result = _run_engine($job) }
+        qr {Symlinked from "t/data/openqa/share/tests/fedora" to "$pool_directory/fedora"},
```
perlpunk, line 475:
```suggestion
        qr{Symlinked from "t/data/openqa/share/tests/fedora" to "$pool_directory/fedora"},
```
or was that on purpose?

perlpunk: Oh I see this was also used in the previous subtest. Too bad that neither perltidy nor perlcritic complain about it...

PR #7007, opened 2026-02-17
test(43-scheduling-and-worker-scalability): automate job count scenarios

okurz:
- Derives job counts (nr_workers-2, nr_workers, nr_workers+2) from worker count.
- Wraps core test logic in a loop to cover all scenarios automatically.
- Ensures a fresh database and services for each iteration to prevent state leakage.
- Restores SCALABILITY_TEST_WORKER_COUNT as the primary environment variable.

After:
* #7003

### DIFF
--- a/t/43-scheduling-and-worker-scalability.t
+++ b/t/43-scheduling-and-worker-scalability.t
@@ -15,7 +15,7 @@ use Mojo::File qw(path tempfile);
 use Mojo::Util qw(dumper scope_guard);
 use IPC::Run qw(start);
 use FindBin;
-use lib "$FindBin::Bin/lib", "$FindBin::Bin/../external/os-autoinst-common/lib";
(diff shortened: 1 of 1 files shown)


t/43-scheduling-and-worker-scalability.t
```diff
@@ -31,36 +31,44 @@ use OpenQA::Utils 'testcasedir';
 
 BEGIN {
     # set defaults
-    $ENV{SCALABILITY_TEST_JOB_COUNT} //= 5;
```
Martchus, line 34: It would still be good to be able to set this manually. The idea of this test was to be able to test with a very high job count and/or worker count to see how the scheduler, ws server and worker scale. For this it makes sense if each count can be configured individually.

Martchus (changes requested):

t/43-scheduling-and-worker-scalability.t
okurz: ok, done

Martchus (approved):

d3flex (approved):

PR #7025, opened 2026-02-21
Demo Test2::V0

perlpunk:
This is to show how [Test2::V0](https://metacpan.org/pod/Test2::V0) can be used.
Some differences:
* PRO `is` can replace `is_deeply` and it can be more clever - you can leave out parts of the data structure you're not interested in
* PRO Comes with its own `mock` and `mock_obj` function. Call tracking is possible. It also offers `before`, `after` and `around`. https://metacpan.org/pod/Test2::Tools::Mock
* CONTRA It doesn't have an `explain` function :( (but there is https://metacpan.org/dist/Test2-Tools-Explain )
* CONTRA we need `-no_srand => 1` because otherwise the test database schema name will not be random and complain that it exists already (from a previous run). Maybe there is a better way to work around that
* PRO `ref_ok` for checking a type of a variable
* PRO lots of comparison functions https://metacpan.org/pod/Test2::V1#COMPARE

### DIFF
--- a/t/05-scheduler-dependencies.t
+++ b/t/05-scheduler-dependencies.t
@@ -3,7 +3,12 @@
 # Copyright 2014-2021 SUSE LLC
 # SPDX-License-Identifier: GPL-2.0-or-later
 
-use Test::Most;
+use Test2::V0 -no_srand => 1;
+use Test::More (); #qw(no_plan explain);
+
+#use Test::Most;
+
+sub always_explain { Test::More::explain(@_) }
 
 BEGIN { $ENV{OPENQA_SCHEDULER_STARVATION_PROTECTION_PRIORITY_OFFSET} = 5 }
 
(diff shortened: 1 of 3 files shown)


perlpunk:
`set_series` can be emulated like this though:
```perl
my $obj = mock obj => {} => (
    add => [
        methodname => sub { state @ret = qw(a b c); return shift @ret }
    ]
);
```

t/05-scheduler-dependencies.t
```diff
@@ -634,13 +643,9 @@ subtest 'duplicate parallel parent in tree with all dependency types' => sub {
...
-    my @sorted_got = sort(@{$jobQ_deps->{children}->{Chained}});
-    my @sorted_exp = sort(($jobW->id, $jobU->id, $jobR->id, $jobT->id));
-    is_deeply(\@sorted_got, \@sorted_exp, 'jobQ is chained parent to all jobs except jobTA')
-      or always_explain \@sorted_got;
-    @sorted_got = sort(@{$jobQ_deps->{children}->{'Directly chained'}});
-    @sorted_exp = sort(($jobTA->id));
-    is_deeply(\@sorted_got, \@sorted_exp, 'jobQ is directly chained parent to jobTA') or always_explain \@sorted_got;
+    my $check = bag { item $_ for $jobW->id, $jobU->id, $jobR->id, $jobT->id; end() };
+    is $jobQ_deps->{children}->{Chained}, $check, 'jobQ is chained parent to all jobs except jobTA';
+    is $jobQ_deps->{children}->{'Directly chained'}, [$jobTA->id], 'jobQ is directly chained parent to jobTA';
```
Martchus, line 648: This new test API allows for shorter code but the code is also less intuitive with "magic" functions like `bag`, `item` and `end()` that are not really self-explanatory.

PR #7026, opened 2026-02-21
perf: handle huge job groups gracefully

okurz:
Motivation
Fetching and iterating all job objects for huge job groups causes
significant performance degradation and memory issues. The lack of
database-level aggregation and missing limits per build makes the system
vulnerable to excessive resource consumption. Additionally, some newly
introduced configuration names were confusing.

Design Choices
1. Optimized compute_build_results by using database-level aggregation
   instead of fetching and iterating all job objects.
2. Moved job deduplication (per scenario) to the database level.
3. Optimized comment/review tracking by only checking failed jobs.
4. Added t/61-job_group_aggregation.t to verify aggregation and limit
   enforcement, including specific subtests for parent group oversized
   build coverage.
5. Extracted category mapping into _get_job_result_category helper and
   reused it in both count_job and _count_job_aggregated for consistency.
6. Renamed the new config 'job_group_overview_max_jobs' to
   'build_results_max_jobs_per_build' across the application, tests,
   and default config to clearly distinguish it from the dashboard limit.

Benefits
These optimizations drastically reduce memory usage and processing time
for large job groups. Job figures are safely handled when encountering
oversized builds, preventing unhandled regressions. Config renaming
reduces confusion for administrators and developers.

Related progress issue: https://progress.opensuse.org/issues/196913

### DIFF
--- a/lib/OpenQA/BuildResults.pm
+++ b/lib/OpenQA/BuildResults.pm
@@ -5,15 +5,16 @@ package OpenQA::BuildResults;
 
 use Mojo::Base -strict, -signatures;
 
-use OpenQA::Jobs::Constants;
+use OpenQA::Jobs::Constants
+  qw(DONE CANCELLED PENDING_STATES PASSED SOFTFAILED ABORTED FAILED NOT_COMPLETE NOT_OK_RESULTS meta_result);
 use OpenQA::Constants qw(BUILD_SORT_BY_NAME BUILD_SORT_BY_NEWEST_JOB BUILD_SORT_BY_OLDEST_JOB);
 use OpenQA::Schema::Result::Jobs;
 use OpenQA::Utils;
 use Date::Format;
 use DateTime::Format::Pg;
-use List::Util qw(any);
 use Sort::Versions;
 use Time::Seconds;
+use List::Util qw(any);
 
 use constant DEFAULT_BUILD_RESULTS_LIMIT => 400;
 
@@ -31,6 +32,51 @@ sub init_job_figures ($job_result) {
     $job_result->{total} = 0;
 }
 
+sub _reset_stats_for_oversized_build ($jr, $total_for_build) {
+    init_job_figures($jr);
+    $jr->{total} = $total_for_build;
+    init_job_figures($_) for values %{$jr->{children} // {}};
+}
+
+my %META_MAP = (
+    PASSED() => 'passed',
+    SOFTFAILED() => 'softfailed',
+    ABORTED() => 'skipped',
+    FAILED() => 'failed',
+    NOT_COMPLETE() => 'failed',
+);
+
+sub _get_job_result_category ($state, $result) {
+    return $META_MAP{meta_result($result)} if $state eq DONE;
+    return 'skipped' if $state eq CANCELLED;
+    return 'unfinished';
+}
+
+sub _count_job_aggregated ($stat, $jr, $count) {
+    $jr->{total} += $count;
+    my $category = _get_job_result_category($stat->state, $stat->result);
+    $jr->{$category} += $count;
+}
+
+sub count_job ($job, $jr, $labels) {
+    _count_job_aggregated($job, $jr, 1);
+    my ($state, $result) = ($job->state, $job->result);
+    my $category = _get_job_result_category($state, $result);
+
+    if ($category eq 'failed') {
+        my $comment_data = $labels->{$job->id};
+        if ($comment_data) {
+            $jr->{labeled}++ if $comment_data->{reviewed};
+            $jr->{comments}++ if $comment_data->{comments} || $comment_data->{reviewed};
+        }
+    }
+    elsif ($category eq 'unfinished') {
+        require OpenQA::Log;
+        OpenQA::Log::log_error('Encountered not-implemented state:' . $state . ' result:' . $result)
+          unless any { $state eq $_ } (PENDING_STATES);
+    }
+    return;
+}
 sub add_review_badge ($build_res) {
     $build_res->{all_passed} = $build_res->{passed} + $build_res->{softfailed} >= $build_res->{total} ? 1 : 0;
     $build_res->{reviewed} = $build_res->{labeled} >= $build_res->{failed} ? 1 : 0;
@@ -65,7 +111,9 @@ sub _get_latest_job_ids ($jobs_resultset, $version, $buildnr, $group_ids) {
 sub compute_build_results (
     $group, $limit, $time_limit_days, $tags, $subgroup_filter, $show_tags,
     $max_jobs_limit = undef,
-    $ignored_groups = undef
+    $ignored_groups = undef,
+    $max_jobs_per_build = undef,
+    $app = undef
   )
 {
     # find relevant child groups taking filter into account
@@ -83,6 +131,11 @@ sub compute_build_results (
         }
     }
 
+    if (!$max_jobs_per_build && $app) {
+        $max_jobs_per_build = $app->config->{misc_limits}->{build_results_max_jobs_per_build};
+    }
+    require OpenQA::Setup;
+    $max_jobs_per_build //= OpenQA::Setup::default_config()->{misc_limits}->{build_results_max_jobs_per_build};
     my $total_jobs_seen = 0;
     my $limit_exceeded = 0;
     my @sorted_results;
@@ -97,8 +150,8 @@ sub compute_build_results (
         });
     return \%result if defined($limit) && int($limit) <= 0;
     # build sorting
-    my $buildver_sort_mode = BUILD_SORT_BY_NAME;
-    $buildver_sort_mode = $group->build_version_sort if $group->can('build_version_sort');
+    my $buildver_sort_mode
+      = $group->can('build_version_sort') ? ($group->build_version_sort // BUILD_SORT_BY_NAME) : BUILD_SORT_BY_NAME;
     my $sort_column = $buildver_sort_mode == BUILD_SORT_BY_OLDEST_JOB ? 'oldest_job' : 'newest_job';
 
     # 400 is the max. limit selectable in the group overview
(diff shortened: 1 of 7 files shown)


lib/OpenQA/Setup.pm
```diff
@@ -242,6 +242,7 @@ sub read_config ($app) {
             generic_default_limit => 10000,
             generic_max_limit => 100000,
             tests_overview_max_jobs => 2000,
+            job_group_overview_max_jobs => 5000,
```
Martchus: This should be added to the config file template as well.

lib/OpenQA/BuildResults.pm
```diff
@@ -185,14 +192,24 @@ sub compute_build_results ($group, $limit, $time_limit_days, $tags, $subgroup_fi
...
-        my $jobs = $jobs_resultset->search(
+        my $jobs_search_filter = {
+            VERSION => $version,
+            BUILD => $buildnr,
+            group_id => {in => $group_ids},
+            clone_id => undef,
+        };
+        my $latest_job_ids_rs = $jobs_resultset->search($jobs_search_filter,
+            {select => [{max => 'id'}], as => [qw(id)], group_by => [qw(TEST ARCH FLAVOR MACHINE)]});
+        my @latest_ids = grep { defined } map { $_->get_column('id') } $latest_job_ids_rs->all;
```
Martchus:
```suggestion
        my @latest_ids = map { $_->get_column('id') // () } $latest_job_ids_rs->all;
```

lib/OpenQA/BuildResults.pm
```diff
@@ -201,34 +218,80 @@ sub compute_build_results ($group, $limit, $time_limit_days, $tags, $subgroup_fi
...
+            my $distri = $stat->get_column('DISTRI');
+            $jr{distris}->{$distri} = 1;
+            if ($jr{children}) {
+                my $child = $jr{children}->{$stat->group_id};
+                _count_job_aggregated($stat, $child, $count);
+                $child->{distris}->{$distri} = 1;
+            }
+        }
+        if ($total_for_build > $max_jobs_per_build) {
+            die 'Build '
```
Martchus: Will this be displayed to the user in a nice way? The error handling looks like one would be presented with some JSON object. Or is there some JavaScript involved that handles that JSON response?

lib/OpenQA/BuildResults.pm
```diff
@@ -201,34 +218,80 @@ sub compute_build_results ($group, $limit, $time_limit_days, $tags, $subgroup_fi
...
+                $t_created = DateTime::Format::Pg->parse_datetime($t_created);
+            }
             if ($newest) {
-                $jr{oldest_newest} //= $job->t_created;
+                $jr{oldest_newest} //= $t_created;
             }
             else {
-                $jr{oldest_newest} = $job->t_created;
+                $jr{oldest_newest} = $t_created;
             }
```
Martchus: Could be a one-liner. (Of course you're only changing the rhs here.)

lib/OpenQA/BuildResults.pm
```diff
@@ -201,34 +218,80 @@ sub compute_build_results ($group, $limit, $time_limit_days, $tags, $subgroup_fi
...
+        my $extra_info_rs = $jobs_resultset->search(
+            {id => {-in => \@latest_ids}},
+            {
+                select => [{($newest ? 'max' : 'min') => 't_created'}],
+                as => [qw(t_created)],
+            });
+        while (my $info = $extra_info_rs->next) {
+            my $t_created = $info->get_column('t_created');
+            if ($t_created && !ref $t_created) {
+                require DateTime::Format::Pg;
```
Martchus: This should probably be a use at the beginning.

lib/OpenQA/BuildResults.pm
```diff
@@ -30,41 +33,41 @@ sub init_job_figures ($job_result) {
     $job_result->{total} = 0;
 }
 
+sub _get_job_result_category ($state, $result) {
```
Martchus, line 49: I'm wondering whether we can reuse some of the mapping functions we already have to avoid repetitiveness here. Of course this is already better than how it was before your change.

Martchus (changes requested):
Look generally good. Have you tested this with production data and compared before and after? I think for bigger changes like that we should do that (like when I recently did a similar optimization for the test results overview page). This way we could also confirm whether it helps with the problem from the motivating ticket (https://progress.opensuse.org/issues/196709).

t/61-job_group_aggregation.t
```diff
@@ -0,0 +1,246 @@
...
+    my $group_id_api = $schema->resultset('JobGroups')->create({name => 'api_group'})->id;
+    my $version = 'api_limit_version';
+    my $build = 'api_limit_build';
+    my $num_jobs = 10;
+    my $t_created = time2str '%Y-%m-%d %H:%M:%S', time, 'UTC';
+    my %common = (
+        group_id => $group_id_api,
+        priority => 50,
+        state => 'done',
+        result => 'passed',
```
Martchus: It would make sense to use constants.

t/61-job_group_aggregation.t
```diff
@@ -0,0 +1,246 @@
...
+        FLAVOR => 'flavor',
+        VERSION => $version,
+        BUILD => $build,
+        DISTRI => 'distri',
+        t_created => $t_created,
+        t_updated => $t_created,
+    );
+    my @jobs_data;
+    push @jobs_data, {%common, id => 600000 + $_, TEST => "test_$_"} for (1 .. $num_jobs);
+    $schema->resultset('Jobs')->populate(\@jobs_data);
```
Martchus: Use `map` or avoid the intermediate array by calling `create` in a loop.

t/61-job_group_aggregation.t
```diff
@@ -0,0 +1,246 @@
...
+    );
+    my @jobs_data;
+    push @jobs_data, {%common, id => 500000 + $_, TEST => "test_$_"} for (1 .. $num_jobs);
+    $schema->resultset('Jobs')->populate(\@jobs_data);
+    # Set a very low limit in the app config
+    $t->app->config->{misc_limits}->{job_group_overview_max_jobs} = 5;
+    $t->get_ok("/group_overview/$group_id_ctrl" => form => {distri => 'distri', version => $version, build => $build})
+      ->status_is(400)->content_like(qr/exceeds the limit of 5/);
+};
+subtest 'API Controller limit enforcement' => sub {
```
Martchus: This subtest is almost like the previous ones and they are all very verbose. It would make sense to avoid this kind of duplication.

t/61-job_group_aggregation.t
```diff
@@ -0,0 +1,246 @@
...
+use OpenQA::BuildResults;
+use Date::Format 'time2str';
+use Feature::Compat::Try;
+use Test::Mojo;
+my $test_case = OpenQA::Test::Case->new;
+my $schema = $test_case->init_data(fixtures_glob => '01-jobs.pl 03-users.pl');
+my $t = Test::Mojo->new('OpenQA::WebAPI');
+my $group_id = 1001;
+my $group = $schema->resultset('JobGroups')->find($group_id);
+subtest 'Aggregation with deduplication' => sub {
```
Martchus, line 48: Please use blank lines to separate different sections.

lib/OpenQA/BuildResults.pm
```diff
@@ -30,41 +34,42 @@ sub init_job_figures ($job_result) {
...
+    if ($state eq OpenQA::Jobs::Constants::DONE) {
+        my $meta = OpenQA::Jobs::Constants::meta_result($result);
+        return 'passed' if $meta eq OpenQA::Jobs::Constants::PASSED;
+        return 'softfailed' if $meta eq OpenQA::Jobs::Constants::SOFTFAILED;
+        return 'skipped' if $meta eq OpenQA::Jobs::Constants::ABORTED;
+        return 'failed' if $meta eq OpenQA::Jobs::Constants::FAILED || $meta eq OpenQA::Jobs::Constants::NOT_COMPLETE;
+    }
+    return 'skipped' if $state eq OpenQA::Jobs::Constants::CANCELLED;
```
Martchus: It would look less noisy if the constants were imported.

Martchus (changes requested):
I have just some style nitpicks but enough to say that at least some of them should be improved.

lib/OpenQA/Setup.pm
okurz: done

lib/OpenQA/BuildResults.pm
okurz: done

lib/OpenQA/BuildResults.pm
okurz: I thought this is what works in other cases when we have "die" in the execution path triggered by webUI users. "nice way" is debatable. This is my first attempt but I want to try to just limit and provide a message that it's limited. Now I remember that we have a similar approach on /tests/overview

lib/OpenQA/BuildResults.pm
okurz: done now with

```
$jr{oldest_newest} = $newest ? ($jr{oldest_newest} // $t_created) : $t_created
```

lib/OpenQA/BuildResults.pm
okurz: done

t/61-job_group_aggregation.t
okurz: also for tests? I considered it safer to be explicit here so that tests would break if we muss up constant definitions, isn't it?

lib/OpenQA/BuildResults.pm
```diff
@@ -111,13 +116,17 @@ sub find_child_groups ($group, $subgroup_filter) {
...
+sub compute_build_results ($group, $limit, $time_limit_days, $tags, $subgroup_filter, $show_tags,
+    $max_jobs_per_build = undef)
+{
 
     # find relevant child groups taking filter into account
     my $child_groups = find_child_groups($group, $subgroup_filter);
     my $group_ids = $child_groups->{group_ids};
     my $children = $child_groups->{children};
 
+    $max_jobs_per_build //= DEFAULT_MAX_JOBS_PER_BUILD;
```
d3flex: I guess you can move  the `DEFAULT_MAX_JOBS_PER_BUILD` in the function signature and remove this line altogether

d3flex (changes requested):
it seems to me that this will have impact in the performance. there are multiple `$jobs_resultset->search` which can stress the system with long results. is there anything better to avoid this? or do i miss something?

t/61-job_group_aggregation.t
Martchus: I guess it is a matter of taste. I somehow like to avoid repeating these values also in tests.

lib/OpenQA/BuildResults.pm
okurz: I think not because when the function is called with `undef` value explicitly then the behaviour would be undefined. Also in other places we use this approach like here.

lib/OpenQA/BuildResults.pm
okurz: This is how it would look like for an obviously ridiculous limit of 1:

Screenshot_20260301_001147_openqa_build_job_lmiit_exceeded

but again, as we already have a better way on /tests/overview I would prefer to use something similar to this.

t/61-job_group_aggregation.t
okurz: will do … if I use this code at all

lib/OpenQA/BuildResults.pm
```diff
@@ -201,34 +220,74 @@ sub compute_build_results ($group, $limit, $time_limit_days, $tags, $subgroup_fi
...
+            }
+            if ($cd->{comments} || $cd->{reviewed}) {
+                $jr{comments}++;
+                if ($jr{children} && (my $child = $jr{children}->{$job->group_id})) {
+                    $child->{comments}++;
+                }
             }
         }
+        if ($jr{children}) {
+            for my $child_id (keys %{$jr{children}}) {
```
Martchus:
```suggestion
        if (my $children = $jr{children}) {
            for my $child_id (keys %$children) {
```

lib/OpenQA/BuildResults.pm
okurz: done. More reuse

t/61-job_group_aggregation.t
okurz: ok, done

lib/OpenQA/BuildResults.pm
```diff
@@ -104,14 +115,105 @@ sub find_child_groups ($group, $subgroup_filter) {
...
+sub _get_latest_job_ids ($jobs_resultset, $version, $buildnr, $group_ids) {
+    my $jobs_search_filter = {
+        VERSION => $version,
+        BUILD => $buildnr,
+        group_id => {in => $group_ids},
+        clone_id => undef,
+    };
+    my $latest_job_ids_rs = $jobs_resultset->search($jobs_search_filter,
+        {select => [{max => 'id'}], as => [qw(id)], group_by => [qw(TEST ARCH FLAVOR MACHINE)]});
+    return map { $_->get_column('id') // () } $latest_job_ids_rs->all;
```
Martchus: It would probably be more efficient to return this as array ref or to return the rs object.

lib/OpenQA/BuildResults.pm
```diff
@@ -83,6 +131,11 @@ sub compute_build_results (
         }
     }
 
+    if (!$max_jobs_per_build && $app) {
+        $max_jobs_per_build = $app->config->{misc_limits}->{build_results_max_jobs_per_build};
+    }
+    require OpenQA::Setup;
```
Martchus, line 137: Why can't this be included via a normal `use` at the beginning of the file?

lib/OpenQA/BuildResults.pm
```diff
@@ -31,6 +32,51 @@ sub init_job_figures ($job_result) {
...
+
+    if ($category eq 'failed') {
+        my $comment_data = $labels->{$job->id};
+        if ($comment_data) {
+            $jr->{labeled}++ if $comment_data->{reviewed};
+            $jr->{comments}++ if $comment_data->{comments} || $comment_data->{reviewed};
+        }
+    }
+    elsif ($category eq 'unfinished') {
+        require OpenQA::Log;
```
Martchus, line 74: Why can't this be included via a normal `use` at the beginning of the file?

Martchus (changes requested):

PR #7064, opened 2026-03-03
feat: Use CASEDIR to avoiding manual symlinking for running test

d3flex:
if CASEDIR is set to a non-URL relative path, it's used as the effective distri instead of DISTRI. This avoids the need to manually create symlinks like aeon -> opensuse when cloning/triggering jobs.

find related changes also os-autoinst/openQA#7006

related-issue: https://progress.opensuse.org/issues/196856


okurz (approved):

Martchus (approved):
I've added the "not-ready" label as we should only merge this as soon as SLE tests are actually triggered with `CASEDIR=opensuse`.

Martchus (commented):
It wouldn't hurt to fix grammar and spelling in the commit message.

PR #7068, opened 2026-03-03
test: Move author tests to xt/ and rename target to test-author

okurz:
Motivation:
Following Perl ecosystem best practices, author tests (static analysis,
style checks, POD, etc.) should be separated from functional/dynamic
tests. This allows build environments and package maintainers to focus
exclusively on tests that verify the correctness of the software in its
runtime environment, while avoiding dependencies and runtime overhead
associated with developer-only quality gates.

A similar change was done in os-autoinst in os-autoinst commit 090ad748
in PR https://github.com/os-autoinst/os-autoinst/pull/1596. We already
had a Makefile target called "tidy-compile-style" which sounds like
"author" tests.

Design Choices:
- Moved 00-tidy.t, 01-compile-check-all.t, 01-style.t, 02-pod.t, and
  45-make-update-deps.t from t/ to xt/.
- Renamed the Makefile target 'test-tidy-compile-style' to 'test-author'
  to match industry-standard naming and ensure consistency with the
  'os-autoinst-common' sub-module.
- Updated xt/01-compile-check-all.t to include the new xt/ directory in
  its syntax checks.
- Updated xt/01-style.t to include xt/ in style checks and correctly
  exclude itself from self-referential style rules.
- Simplified Makefile targets (test-t, test-ui, test-api) by removing
  GLOBIGNORE exclusions for moved tests, as they are no longer in t/.
- Updated t/testrules.yml to reflect the new test locations.
- Simplified dist/rpm/openQA.spec as author tests can be ignored
  completely.

User Benefit:
- Developers get quick feedback via the generic 'make test-author'.
- CI and build environments can more easily exclude non-functional
  tests, speeding up build-time verification and reducing dependency
  bloat.
- Improved project structure consistency with other projects in the
  os-autoinst ecosystem.


perlpunk:
> > I think `01-compile-check-all.t,` still belongs under `t/` and not `xt/`.
> 
> and can you elaborate why?

It's a test for functionality and not style.
https://metacpan.org/pod/Dist::Zilla::Plugin::Test::Compile (a plugin for dzil, a tool to maintain a CPAN module) can also generate such a test, and it puts it into the `t/` directory by default. (It has an option to put it into `xt` though.)

Makefile
```diff
@@ -369,9 +369,13 @@ endif
...
-.PHONY: test-tidy-compile-style
-test-tidy-compile-style: ## Run tidy, compile and style tests
-	$(MAKE) test-unit-and-integration TIMEOUT_M=20 PROVE_ARGS="$$HARNESS t/*{tidy,compile,style}*.t" GLOBIGNORE="$(unstables)"
+.PHONY: test-author
+test-author: ## Run author tests (tidy, style, pod, etc.)
+	$(MAKE) test-unit-and-integration TIMEOUT_M=20 PROVE_ARGS="$$HARNESS xt/*.t" GLOBIGNORE="$(unstables)" COVEROPT=""
+
+.PHONY: test-compile
+test-compile: ## Run compile tests
+	$(MAKE) test-unit-and-integration TIMEOUT_M=20 PROVE_ARGS="$$HARNESS t/compile/*.t" GLOBIGNORE="$(unstables)" COVEROPT=""
```
perlpunk: Why without `COVEROPT`?
This way it can theoretically happen that a module that is completely uncovered would go unnoticed, because it wouldn't show up at all in the coverage report, so codecov would not consider it.

For the author tests it's fine.

okurz: I added the COVEROPT to reduce overhead and save execution time. 

> This way it can theoretically happen that a module that is completely uncovered would go unnoticed, because it wouldn't show up at all in the coverage report, so codecov would not consider it.

So you mean a perl file which we don't even load as dependency anywhere else? I don't see why the compile tests needs to take care of that.

perlpunk: I mean if someone adds a script `script/foo` which uses `lib/OpenQA/Foo.pm`, but doesn't test this. This might be covered by the `44-scripts.t`, though...

okurz: I will test the performance differences with and without coverage analysis

perlpunk: See also my other comment about using Test::Compile, which reduces runtime to ~33%.
https://github.com/os-autoinst/openQA/pull/7162#discussion_r2969578811

okurz: like this https://github.com/os-autoinst/os-autoinst/pull/2848 ?

perlpunk: yes

okurz:
All ready for review

@os-autoinst/tools-team

aylen384 (approved):

Martchus (approved):

xt/02-perlcritic.t
```diff
@@ -2,6 +2,7 @@
 # SPDX-License-Identifier: GPL-2.0-or-later
 
 use Test::Most;
+use Test::Warnings qw(:no_end_test :report_warnings);
```
perlpunk, line 5: I think, `:no_end_test` isn't necessary here. The files are not really loaded, just parsed by Perl:Critic.

PR #7127, opened 2026-03-18
RFD: Ensure code executed on the worker is confined

d3flex:
Issue: https://progress.opensuse.org/issues/194717

### DIFF
--- a/docs/194717-ensure-code-executed-on-worker-is-confined.md
+++ b/docs/194717-ensure-code-executed-on-worker-is-confined.md
@@ -1,0 +1,87 @@
+# 194717 Ensure code executed on the worker is confined
+Issue: https://progress.opensuse.org/issues/194717
+
+## Summary
+
+Test modules run as Perl code inside the `isotovideo` process on the **host**, not inside the VM.
+That means a test module can read files, write to the filesystem, or make network requests as the worker OS user — even on a production server with a dedicated account.
+
+This document describes what was investigated, what works, and what still needs to be done.
+
+## Investigation
+
+To demonstrate the attack surface I wrote a test module (`tests/install/explore.pm`) that uses standard testapi calls (`script_run`, `upload_logs`) alongside direct Perl code to do sensitive things from inside the worker process on the host side.
+TThis covers the main ways a malicious test module could exploit the worker: testapi calls that run commands in the VM but also give access to the host process, direct host file reads, and leaking files via `upload_logs` to the web UI.
+Other attack vectors acknowledged but not demonstrated: external code pulled in via container images or git repos during a job, and arbitrary commands passed through job variables or openQA settings.
+
+I ran the same test module against a confined and an unconfined worker to compare results.
+
+### systemd hardening — tested and working
+
+I added standard systemd hardening directives to the worker unit. This solution seems to be the less efortless with great value return, which should be applied first .
+The key confirmed results:
+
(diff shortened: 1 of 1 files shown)


docs/194717-ensure-code-executed-on-worker-is-confined.md
```diff
@@ -0,0 +1,87 @@
...
+## Summary
+
+Test modules run as Perl code inside the `isotovideo` process on the **host**, not inside the VM.
+That means a test module can read files, write to the filesystem, or make network requests as the worker OS user — even on a production server with a dedicated account.
+
+This document describes what was investigated, what works, and what still needs to be done.
+
+## Investigation
+
+To demonstrate the attack surface I wrote a test module (`tests/install/explore.pm`) that uses standard testapi calls (`script_run`, `upload_logs`) alongside direct Perl code to do sensitive things from inside the worker process on the host side.
```
okurz, line 13: > I wrote a test module (`tests/install/explore.pm`)

where can I find this?

okurz (commented):
content looks good although I don't think we should commit this file into the repo at the current state.

docs/194717-ensure-code-executed-on-worker-is-confined.md
d3flex: that not really important. I will add something similar to the openqa-in-opena as suggested in the RFD.

docs/194717-ensure-code-executed-on-worker-is-confined.md
```diff
@@ -0,0 +1,87 @@
...
+This covers the main ways a malicious test module could exploit the worker: testapi calls that run commands in the VM but also give access to the host process, direct host file reads, and leaking files via `upload_logs` to the WebUI.
```
Martchus:
```suggestion
This covers the main ways a malicious test module could exploit the worker: testapi calls that run commands in the VM but also give access to the host process, direct host file reads, and leaking files via `upload_logs` to the web UI.
```

docs/194717-ensure-code-executed-on-worker-is-confined.md
```diff
@@ -0,0 +1,87 @@
...
+Other attack vectors acknowledged but not demonstrated: external code pulled in via container images or git repos during a job, and arbitrary commands passed through job variables or openQA settings.
+
+I ran the same test module against a confined and an unconfined worker to compare results.
+
+### systemd hardening — tested and working
+
+I added standard systemd hardening directives to the worker unit. This solution seems to be the simplier less efortless to great value approach, which should be applied first .
```
Martchus: "less efortless to great value approach" is incomprehensible.

Martchus (commented):
For this to become part of our documentation you should avoid the use of "I". Maybe that's also not strictly necessary right now.

It would be interesting to test the isolation of tests runs from each other. I think it would be easily possible to restrict accessible directories even more so tests would only be able to read needed files from the system and from their *own* pool directory (but not pool directories of other worker slots).

It would also be interesting to test whether it works with the cache service or whether further settings need to be added.

Martchus:
> > For this to become part of our documentation you should avoid the use of "I". Maybe that's also not strictly necessary right now.
> 
> As I stated in [#7127 (review)](https://github.com/os-autoinst/openQA/pull/7127#pullrequestreview-3969160440) I don't see the benefit of submitting this to the repo.

And with no updates for so long it makes sense to close this.

PR #7162, opened 2026-03-20
author tests + common perlcritic

okurz:


Makefile
```diff
@@ -369,9 +369,13 @@ endif
...
-.PHONY: test-tidy-compile-style
-test-tidy-compile-style: ## Run tidy, compile and style tests
-	$(MAKE) test-unit-and-integration TIMEOUT_M=20 PROVE_ARGS="$$HARNESS t/*{tidy,compile,style}*.t" GLOBIGNORE="$(unstables)"
+.PHONY: test-author
+test-author: ## Run author tests (tidy, style, pod, etc.)
+	$(MAKE) test-unit-and-integration TIMEOUT_M=20 PROVE_ARGS="$$HARNESS xt/*.t" GLOBIGNORE="$(unstables)" COVEROPT=""
+
+.PHONY: test-compile
+test-compile: ## Run compile tests
+	$(MAKE) test-unit-and-integration TIMEOUT_M=20 PROVE_ARGS="$$HARNESS t/compile/*.t" GLOBIGNORE="$(unstables)" COVEROPT=""
```
perlpunk, line 415: I think I commented elsewhere in another PR with the same change that it could theoretically happen that a file (script or module) goes completely uncovered if it isn't tested at all, and the compile check doesn't run with coverage.
Just saying, in my role as the chief scepticism officer ;-)

script/openqa-load-templates
```diff
@@ -215,30 +215,32 @@ sub post_entry ($table, $entry) {
...
-            # unless empty, but we *can* replace existing YAML
-            # template strings with 'empty' ones
-            if ($table eq "JobGroups") {
-                # unlike all other tables, is not a hash
-                for my $jg (@$result) {
-                    next unless ($jg->{template});
-                    $res = post_yaml_templates($jg->{name}, "scenarios: {}\nproducts: {}\n");
-                    last unless $res->is_success;
-                }
+        unless ($res->is_success) {
```
perlpunk, line 222: I don't like too much refactoring in a commit titled `fix: resolve perlcritic violations in scripts`. Was perlcritic complaining about the complexity here?

t/24-worker-jobs.t
```diff
@@ -1033,7 +1034,7 @@ subtest 'Job stopped while uploading' => sub {
     # pretend there's an image with thumbnail and a regular file to be uploaded
     my $test_image = 'test.png';
     path($job->_result_file_path(".thumbs/$test_image"))->touch;
-    $job->images_to_send->{'098f6bcd4621d373cade4e832627b4f6'} = $test_image;
+    $job->images_to_send->{'098f6bcd4621d373cade4e832627b4f6'} = $test_image;    ## no critic (OpenQA::HashKeyQuotes)
```
perlpunk, line 1037: The `no critic` should not be needed after https://github.com/os-autoinst/os-autoinst-common/pull/74

t/api/02-iso-download.t
```diff
@@ -4,7 +4,7 @@
 # SPDX-License-Identifier: GPL-2.0-or-later
 
 use Test::Most;
-use Mojo::Base -base, -signatures;
+use experimental 'signatures';
```
perlpunk, line 7: This won't work in cases where Mojo::Base is actually used as a base class.
However it shouldn't be needed anymore after https://github.com/os-autoinst/os-autoinst-common/pull/75

Makefile
perlpunk: I had a look at Test::Strict.
It checks if a file compiles and if it has strict.
As we already have a Perl::Critic rule to check for strict, we can actually drop Test::Strict and replace it with Test::Compile.
Runtime 60s (Test::Compile) vs. 170s (Test::strict).
```
my $test = Test::Compile->new();
$test->all_files_ok(qw(t lib script ));
```
Test::Compile isn't a drop-in replacement; It doesn't check `*.t` files, but maybe that's ok, as we still ensure they compile by running them.

perlpunk: Or we can keep Test::Strict and just set `$Test::Strict::TEST_STRICT` and `$Test::Strict::TEST_WARNINGS` to 0.

perlpunk: > Or we can keep Test::Strict and just set `$Test::Strict::TEST_STRICT` and `$Test::Strict::TEST_WARNINGS` to 0.

Hm, that actually doesn't reduce the runtime.

t/24-worker-jobs.t
okurz: https://github.com/os-autoinst/os-autoinst-common/pull/74 merged. I should rebase and remove the obsolete exclusion markers.

t/api/02-iso-download.t
okurz: https://github.com/os-autoinst/os-autoinst-common/pull/75 merged. I should rebase and revert those now not anymore needed changes.

script/openqa-validate-yaml
```diff
@@ -76,7 +76,7 @@ use Getopt::Long::Descriptive;
 use OpenQA::YAML qw(load_yaml validate_data);
 
 my ($opt, $usage) = describe_options(
-    <<"EOM",
+    <<'EOM',
```
perlpunk, line 79: This way it won't interpolate the `$0` below

perlpunk (changes requested):

PR #7177, opened 2026-03-24
chore(AGENTS.md): avoid by default adoption of agentic conventions

d3flex:
There are many reasons to not enforce use of AGENTS.md by default Here some of them:
- let the user decide whether to use AI coding assistants at all
- user might already have his/her own one
- avoids conflicts
- maintainance (prevent outdated conventions)
- keeps the repo free of tooling-specific


okurz (changes requested):
You can always combine personal preferences by putting them into local and home folder AGENTS.md, no need to disable the project one. If you see certain guidelines not generally applicable please address those specifically, don't move the whole file

Martchus (changes requested):
To me this specific AGENTS.md file looks like a good default.

okurz:
I guess after more than a month we can conclude that we won't merge that.

PR #7206, opened 2026-03-28
feat: add scheduling skip reasons to the webUI

okurz:
Motivation:
When openQA jobs remain in the scheduled state, users have no way to know why
they are not being assigned. Surfacing scheduler skip reasons reduces
confusion and support requests.

Design Choices:
- Used the existing "reason" column in the "jobs" table to store per-job
  scheduling skip reasons.
- Implemented a write-on-change strategy in the scheduler to update the
  reasons only when they change, minimizing database write load.
- Replaced the direct "df" call in the WebUI with a check on the aggregated
  job reasons to improve efficiency.
- Added a popover next to the "Scheduled jobs" heading and tooltips on job
  status icons to display the reasons.

Benefits:
- Provides immediate, per-job feedback to users about scheduling delays.
- Decouples the WebUI from low-level system checks like disk space.
- Leverages existing database schema for zero-migration implementation.

Screenshot_20260328_221952_oqa_scheduling_no_issues

Screenshot_20260328_235656_oqa_scheduling_however_all_tests

Screenshot_20260328_235718_oqa_scheduling_popover

Screenshot_20260328_235735_oqa_details_scheduling_info

Related issue: https://progress.opensuse.org/issues/198647

### DIFF
--- a/assets/javascripts/tests.js
+++ b/assets/javascripts/tests.js
@@ -79,10 +79,14 @@ function renderTestName(data, type, row) {
   if (row.result !== 'none') {
     html += '<i class="status fa-solid fa-circle result_' + row.result + '" title="Done: ' + row.result + '"></i>';
   } else if (row.state === 'scheduled') {
+    let title = 'Scheduled';
+    if (row.reason) {
+      title += ': ' + row.reason;
+    }
     if (typeof row.blocked_by_id === 'number') {
       html += '<i class="status fa-solid fa-circle state_blocked" title="Blocked"></i>';
     } else {
-      html += '<i class="status fa-solid fa-circle state_scheduled" title="Scheduled"></i>';
+      html += '<i class="status fa-solid fa-circle state_scheduled" title="' + htmlEscape(title) + '"></i>';
     }
   } else if (row.state === 'assigned') {
     html += '<i class="status fa-solid fa-circle state_running" title="Assigned"></i>';
@@ -323,9 +327,13 @@ function renderTestLists() {
       dataSrc: function (json) {
         // update heading when JSON is available
         let blockedCount = 0;
-        jQuery.each(json.data, function (index, row) {
-          if (row && typeof row.blocked_by_id === 'number') {
+        const reasons = {};
+        json.data.forEach(row => {
+          if (row && row.blocked_by_id) {
             ++blockedCount;
+          }
+          if (row && row.reason) {
+            reasons[row.reason] = (reasons[row.reason] || 0) + 1;
           }
         });
         let text = json.data.length + ' scheduled jobs';
@@ -334,6 +342,26 @@ function renderTestLists() {
         }
         $('#scheduled_jobs_heading').text(text);
         $('#scheduled_jobs_warning').toggleClass('d-none', !json.job_skipped_by_disk_limits);
+
+        const reasonKeys = Object.keys(reasons).sort();
+        const popoverContent = reasonKeys.length
+          ? `<ul>${reasonKeys.map(r => '<li>' + reasons[r] + ' job(s): ' + htmlEscape(r) + '</li>').join('')}</ul>`
+          : 'No scheduling issues detected';
+
+        const popoverIcon = document.getElementById('scheduled_jobs_help_icon');
+        if (popoverIcon) {
+          popoverIcon.classList.toggle('text-muted', !reasonKeys.length);
+          popoverIcon.setAttribute('data-bs-content', popoverContent);
+
+          // Update active instance if it exists
+          if (window.bootstrap && bootstrap.Popover) {
+            const popover = bootstrap.Popover.getInstance(popoverIcon);
+            if (popover) {
+              popover.setContent({'.popover-body': popoverContent});
+            }
+          }
+        }
+
         return json.data;
       }
     },
--- a/lib/OpenQA/Scheduler/Model/Jobs.pm
+++ b/lib/OpenQA/Scheduler/Model/Jobs.pm
@@ -55,15 +55,19 @@ sub effective_job_limit ($self) {
 sub _allocate_jobs ($self, $free_workers) {
     my ($allocated_workers, $allocated_jobs) = ({}, {});
     my $scheduled_jobs = $self->scheduled_jobs;
+    $scheduled_jobs->{$_}->{current_reason} = undef for keys %$scheduled_jobs;
+
     my $schema = OpenQA::Schema->singleton;
     my $running = $schema->resultset('Jobs')->count({state => [OpenQA::Jobs::Constants::EXECUTION_STATES]});
     my $limit = $self->effective_job_limit;
     if (storage_below_threshold()) {
         log_debug('Skipping job scheduling: free storage space in results directory below threshold');
+        $_->{current_reason} = 'storage space below threshold' for values %$scheduled_jobs;
         return ({}, {});
     }
     if ($limit >= 0 && $running >= $limit) {
         log_debug("job limit ($limit) exceeded, scheduling no additional jobs");
+        $_->{current_reason} = "job limit reached (max_running_jobs=$limit)" for values %$scheduled_jobs;
         return ({}, {});
     }
     my $max_allocate = $limit >= 0 ? min(MAX_JOB_ALLOCATION, $limit - $running) : MAX_JOB_ALLOCATION;
@@ -73,7 +77,10 @@ sub _allocate_jobs ($self, $free_workers) {
     for my $id (keys %$scheduled_jobs) {
         my $jobinfo = $scheduled_jobs->{$id};
         $jobinfo->{matching_workers} = _matching_workers($jobinfo, $free_workers, \%rejected);
-        delete $scheduled_jobs->{$id} unless @{$jobinfo->{matching_workers}};
+        if (!@{$jobinfo->{matching_workers}}) {
+            my @needed = sort @{$jobinfo->{worker_classes}};
+            $jobinfo->{current_reason} = @needed ? "no free workers for class @needed" : 'no workers online';
+        }
     }
     if (keys %rejected) {
         my @rejected = sort { $rejected{$b} <=> $rejected{$a} || $a cmp $b } keys %rejected;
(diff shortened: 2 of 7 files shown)


assets/javascripts/tests.js
```diff
@@ -311,17 +315,42 @@ function renderTestLists() {
...
         $('#scheduled_jobs_warning').toggleClass('d-none', !json.job_skipped_by_disk_limits);
+
+        const reasonKeys = Object.keys(reasons).sort();
+        const popoverContent = reasonKeys.length
+          ? '<ul>' +
+            reasonKeys.map(r => '<li>' + htmlEscape(reasons[r]) + ' job(s): ' + htmlEscape(r) + '</li>').join('') +
+            '</ul>'
+          : 'No scheduling issues detected';
+
+        const popoverIcon = $('#scheduled_jobs_help_icon');
```
Martchus: It would be nice to avoid jQuery in new code.

assets/javascripts/tests.js
```diff
@@ -311,17 +315,42 @@ function renderTestLists() {
...
+        popoverIcon.toggleClass('text-muted', !reasonKeys.length).attr('data-bs-content', popoverContent);
+
+        // Update active instance if it exists
+        if (window.bootstrap && bootstrap.Popover) {
+          const popover = bootstrap.Popover.getInstance(popoverIcon[0]);
+          if (popover) {
+            popover.setContent({
+              '.popover-body': popoverContent
+            });
```
Martchus:
```suggestion
            popover.setContent({'.popover-body': popoverContent});
```

assets/javascripts/tests.js
```diff
@@ -311,17 +315,42 @@ function renderTestLists() {
...
         if (blockedCount > 0) {
           text += ' (' + blockedCount + ' blocked by other jobs)';
         }
         $('#scheduled_jobs_heading').text(text);
         $('#scheduled_jobs_warning').toggleClass('d-none', !json.job_skipped_by_disk_limits);
+
+        const reasonKeys = Object.keys(reasons).sort();
+        const popoverContent = reasonKeys.length
+          ? '<ul>' +
+            reasonKeys.map(r => '<li>' + htmlEscape(reasons[r]) + ' job(s): ' + htmlEscape(r) + '</li>').join('') +
```
Martchus: This could use template literals. The first `htmlEscape` is probably also unnecessary.

lib/OpenQA/Scheduler/Model/Jobs.pm
```diff
@@ -86,10 +95,14 @@ sub _allocate_jobs ($self, $free_workers) {
     my %checked_jobs;
     for my $j (@sorted) {
         my $job_id = $j->{id};
-        next if $checked_jobs{$job_id} || defined $allocated_jobs->{$job_id} || !@{$j->{matching_workers}};
+        if ($checked_jobs{$job_id} || defined $allocated_jobs->{$job_id} || !@{$j->{matching_workers}}) {
+            $checked_jobs{$job_id} = 1;
```
Martchus, line 105: This is so far only set in the `@tobescheduled` loop. I hope setting it here as well has no unwanted side-effects.

lib/OpenQA/Scheduler/Model/Jobs.pm
```diff
@@ -178,7 +198,7 @@ sub _assign_multiple_jobs ($self, $schema, $worker, $worker_id, $jobs, $directly
     my %worker_assignment = (state => ASSIGNED, t_started => undef, assigned_worker_id => $worker_id);
     $schema->txn_do(
         sub {
-            $_->update(\%worker_assignment) for @$jobs;
```
Martchus, line 189: Just add `reason => undef` to `%worker_assignment` instead of making a new hash in every iteration.

And don't we need the same in the code path for assigning a single job? (Or is the change in `reschedule_state` covering that?)

lib/OpenQA/Scheduler/Model/Jobs.pm
```diff
@@ -326,6 +348,25 @@ sub schedule ($self) {
...
+        my $last = $info->{last_reason};
+        if (($curr // '') ne ($last // '')) {
+            push @{$reasons_to_update{$curr // 'NULL'}}, $id;
+            $info->{last_reason} = $curr;
+        }
+    }
+    for my $reason (keys %reasons_to_update) {
+        my $ids = $reasons_to_update{$reason};
+        my $val = $reason eq 'NULL' ? undef : $reason;
+        $schema->resultset('Jobs')->search({id => {-in => $ids}})->update({reason => $val});
```
Martchus: Avoid the resolution of `$schema->resultset('Jobs')` in each loop iteration.

```suggestion
        $jobs->search({id => {-in => $ids}})->update({reason => $val});
```

templates/webapi/test/infopanel.html.ep
```diff
@@ -69,7 +69,7 @@
                         <abbr class="timeago" title="<%= $job->t_created->datetime() %>Z"><%= format_time($job->t_created) %></abbr>
                     % }
                     % if (my $reason = $job->reason) {
-                        <br>Reason: <%= $reason %>
+                        <br><%= $job->state eq 'scheduled' ? 'Scheduling info' : 'Reason' %>: <%= $reason %>
```
Martchus: This should use the constant. Maybe this check is also completely unnecessary, though. (Assigned jobs should not have a reason, right?)

Martchus (changes requested):
Looks generally good. I only have a few nitpicks and questions.

assets/javascripts/tests.js
okurz: done

assets/javascripts/tests.js
okurz: done

assets/javascripts/tests.js
okurz: done

lib/OpenQA/Scheduler/Model/Jobs.pm
okurz: none that I am aware of :)

lib/OpenQA/Scheduler/Model/Jobs.pm
okurz: Regarding the single job assignment: Yes, it is also covered. `_assign_single_job` calls `$job->set_assigned_worker`, which I have also updated to clear the `reason` field. While `reschedule_state` ensures the reason is cleared when a job is reverted to the scheduled state, updating it during assignment ensures that any previous "scheduling info" would not persist once a worker is successfully assigned, I guess

lib/OpenQA/Scheduler/Model/Jobs.pm
okurz: done

templates/webapi/test/infopanel.html.ep
okurz: done

d3flex:
image

I think the number of jobs in the popup is redudant. is there a case where the number will not match?

okurz:
> I think the number of jobs in the popup is redudant. is there a case where the number will not match?

of course. Jobs can be scheduled and waiting for various things, different worker classes, global job limit, the "df-thingy" you implemented, etc.

d3flex:
> > I think the number of jobs in the popup is redudant. is there a case where the number will not match?
> 
> of course. Jobs can be scheduled and waiting for various things, different worker classes, global job limit, the "df-thingy" you implemented, etc.

but the amount of scheduled jobs will be always the same. Such, why to add it in the popup?

assets/javascripts/tests.js
Martchus: Done where? I still see `const popoverIcon = $('#scheduled_jobs_help_icon');`.

kalikiana:
> but the amount of scheduled jobs will be always the same. Such, why to add it in the popup?

The total won't be the same if they are delayed for other reasons.

lib/OpenQA/Scheduler/Model/Jobs.pm
```diff
@@ -136,6 +148,10 @@ sub _allocate_jobs ($self, $free_workers) {
             my $free_worker_count = @$free_workers;
             log_debug('limit reached, scheduling no additional jobs'
                   . " (max_running_jobs=$limit, free workers=$free_worker_count, running=$running, allocated=$busy)");
+            for my $id (keys %$scheduled_jobs) {
+                next if defined $allocated_jobs->{$id} || $checked_jobs{$id};
+                $scheduled_jobs->{$id}->{current_reason} //= 'scheduling limit reached';
```
kalikiana:
```suggestion
                $scheduled_jobs->{$id}->{current_reason} //= 'scheduling limit reached' . " (max_running_jobs=$limit)";
```

Can we include the allowed amount here?

templates/webapi/test/list.html.ep
```diff
@@ -29,7 +29,9 @@
 </div>
 
 <div>
-    <h2 id="scheduled_jobs_heading">Scheduled jobs</h2>
+    <h2 id="scheduled_jobs_heading" style="display: inline-block">Scheduled jobs</h2><%=
+      help_popover 'Why is my test not executed?' => 'No scheduling issues detected',
```
kalikiana:
```suggestion
      help_popover 'Why is my test not getting executed?' => 'No scheduling issues detected',
```

assets/javascripts/tests.js
okurz: Now it's

```
const popoverIcon = document.getElementById('scheduled_jobs_help_icon');
```

templates/webapi/test/list.html.ep
okurz: done

lib/OpenQA/Scheduler/Model/Jobs.pm
okurz: done

okurz:
@Martchus @kalikiana @d3flex all your suggestions have been included. Ready for re-review

Martchus (approved):

d3flex (approved):

PR #7232, opened 2026-04-01
feat: enhance dynamic job limit with fast ramp-up and centralization

okurz:
Motivation:
The initial dynamic global job limit was too slow to scale up when the system
was idle, and its scaling factors were hardcoded and repeated across the
codebase, making maintenance difficult.

Design Choices:
- Implemented 'fast ramp-up' logic doubling the step size when load is low.
- Centralized all default scaling parameters as constants in DynamicLimit.pm.
- Updated OpenQA::Setup to use these constants for the configuration schema
  by referencing DynamicLimit->DEFAULTS.
- Refactored tests to rely on centralized constants and DEFAULTS, reducing
  repetition of literal values like '0.85' and '50'.
- Added current min/max values to "Dynamic job limit" log messages.

Benefits:
- Faster utilization of spare host capacity (reaching ceilings ~4x faster).
- Single source of truth for all dynamic limit defaults.
- Cleaner and more maintainable codebase and tests.
- Better operational visibility into the dynamic scaling behavior.

Related issue: https://progress.opensuse.org/issues/198770

### DIFF
--- a/lib/OpenQA/Scheduler/DynamicLimit.pm
+++ b/lib/OpenQA/Scheduler/DynamicLimit.pm
@@ -10,10 +10,29 @@ use OpenQA::Log qw(log_debug);
 use OpenQA::Utils qw(load_avg);
 use Time::HiRes 'time';
 
-# Fraction of threshold below which all load averages must fall to scale up.
-use constant SCALE_UP_HYSTERESIS => 0.7;
-# Multiplier applied to step on an emergency (critical) cutback.
-use constant EMERGENCY_STEP_MULTIPLIER => 3;
+use constant {
+    SCALE_UP_HYSTERESIS => 0.7,
+    FAST_RAMP_UP_LOAD_FACTOR => 0.3,
+    LOAD_THRESHOLD_FACTOR => 0.85,
+    LOAD_CRITICAL_FACTOR => 1.5,
+    STEP => 10,
+    MIN => 50,
+    INTERVAL => 60,
+    EMERGENCY_STEP_MULTIPLIER => 3,
+};
+
+use constant DEFAULTS => {
+    dynamic_job_limit_load_threshold => 0,
+    dynamic_job_limit_load_threshold_factor => LOAD_THRESHOLD_FACTOR,
+    dynamic_job_limit_load_critical => 0,
+    dynamic_job_limit_load_critical_factor => LOAD_CRITICAL_FACTOR,
+    dynamic_job_limit_step => STEP,
+    dynamic_job_limit_min => MIN,
+    max_running_jobs => -1,
+    dynamic_job_limit_interval => INTERVAL,
+    dynamic_job_limit_scale_up_hysteresis => SCALE_UP_HYSTERESIS,
+    dynamic_job_limit_fast_ramp_up_load_factor => FAST_RAMP_UP_LOAD_FACTOR,
+};
 
 has effective_limit => undef;
 
@@ -25,13 +44,14 @@ sub block_adjustment_for ($self, $seconds = 9999) { $self->{_last_adjusted} = ti
 
 # Returns the number of online CPUs by counting processor entries in /proc/cpuinfo.
 sub _nproc () {
-    scalar grep { /^processor\s*:/ } split /\n/, path('/proc/cpuinfo')->slurp;
+    state $cpus = scalar grep { /^processor\s*:/ } split /\n/, path('/proc/cpuinfo')->slurp;
+    return $cpus;
 }
 
 # Returns a resolved (non-zero) threshold, falling back to nproc * $factor.
 sub _resolve_threshold ($configured, $factor) {
     return $configured if $configured > 0;
-    state $nproc = _nproc();
+    my $nproc = _nproc();
     return $nproc > 0 ? $nproc * $factor : 10 * $factor;
 }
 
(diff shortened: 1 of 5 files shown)


Martchus (approved):

lib/OpenQA/Scheduler/DynamicLimit.pm
```diff
@@ -32,20 +67,32 @@ sub _nproc () {
 sub _resolve_threshold ($configured, $factor) {
     return $configured if $configured > 0;
     state $nproc = _nproc();
-    return $nproc > 0 ? $nproc * $factor : 10 * $factor;
+    my $val = $FIXED_NPROC // $nproc;
+    return $val > 0 ? $val * $factor : 10 * $factor;
```
perlpunk: Why not mock `_nproc()` instead?
`our` variables can be used by anyone, and when we change this or remove the variable, other code breaks.
We rarely use `our` in our code for this kind of thing. Sometimes we use env variables.

And if you introduce a global variable, the `state` variable is kind of useless.

perlpunk (changes requested):

lib/OpenQA/Scheduler/DynamicLimit.pm
```diff
@@ -9,11 +9,46 @@ use Mojo::File 'path';
...
+        dynamic_job_limit_load_critical => 0,
+        dynamic_job_limit_load_critical_factor => LOAD_CRITICAL_FACTOR,
+        dynamic_job_limit_step => STEP,
+        dynamic_job_limit_min => MIN,
+        max_running_jobs => -1,
+        dynamic_job_limit_interval => INTERVAL,
+        dynamic_job_limit_scale_up_hysteresis => SCALE_UP_HYSTERESIS,
+        dynamic_job_limit_fast_ramp_up_load_factor => FAST_RAMP_UP_LOAD_FACTOR,
+    };
+}
```
perlpunk:
```suggestion
use constant DEFAULTS => {
    dynamic_job_limit_load_threshold => 0,
    dynamic_job_limit_load_threshold_factor => LOAD_THRESHOLD_FACTOR,
    dynamic_job_limit_load_critical => 0,
    dynamic_job_limit_load_critical_factor => LOAD_CRITICAL_FACTOR,
    dynamic_job_limit_step => STEP,
    dynamic_job_limit_min => MIN,
    max_running_jobs => -1,
    dynamic_job_limit_interval => INTERVAL,
    dynamic_job_limit_scale_up_hysteresis => SCALE_UP_HYSTERESIS,
    dynamic_job_limit_fast_ramp_up_load_factor => FAST_RAMP_UP_LOAD_FACTOR,
};
```
and then later you can use:
```
    DEFAULTS->{dynamic_job_limit_load_threshold}
```

lib/OpenQA/Scheduler/DynamicLimit.pm
```diff
@@ -9,11 +9,46 @@ use Mojo::File 'path';
...
+    LOAD_THRESHOLD_FACTOR => 0.85,
+    LOAD_CRITICAL_FACTOR => 1.5,
+    STEP => 10,
+    MIN => 50,
+    INTERVAL => 60,
+    EMERGENCY_STEP_MULTIPLIER => 3,
+};
+
+our @EXPORT_OK = qw(
+  SCALE_UP_HYSTERESIS
```
perlpunk: Are these used anywhere except in this module?

perlpunk (commented):
Before making this more complicated, what do you think about [throttling worker upload speed](https://progress.opensuse.org/issues/199232) ?

t/26-dynamic-job-limit.t
```diff
@@ -13,7 +13,7 @@ use OpenQA::Test::Utils qw(simulate_load);
 use Test::MockModule;
 use Test::Output qw(combined_like);
 use OpenQA::Utils qw(load_avg);
-use OpenQA::Scheduler::DynamicLimit;
+use OpenQA::Scheduler::DynamicLimit qw(MIN STEP EMERGENCY_STEP_MULTIPLIER);
```
perlpunk: Hm, is it necessary and a goo design to export constants to only use them in tests? And do we really need them in tests? If for some reason we change the value for STEP accidentally, the tests will still pass, right? Shouldn't we have rather have hardcoded values in the tests?

lib/OpenQA/Scheduler/DynamicLimit.pm
okurz: ok, removed the `our` and moved the `state` to the low-level function `_nproc()` plus using mocking as well as an explicit (coverage) test for `_nproc()`

lib/OpenQA/Scheduler/DynamicLimit.pm
okurz: nice, done

lib/OpenQA/Scheduler/DynamicLimit.pm
okurz: addressed together with https://github.com/os-autoinst/openQA/pull/7232#discussion_r3047901466, not exporting anything anymore

t/26-dynamic-job-limit.t
okurz: ok, hardcoded expectations with the explicit examples. Solved.

okurz:
addressed all open code review suggesitions from @perlpunk

perlpunk (approved):

PR #7234, opened 2026-04-01
fix: only report job limit in UI when globally reached

okurz:
Motivation:
The /tests page often displayed "(limited by server config)" even when far
below the actual limit. This happened because the backend was reporting the
limit based on filtered job counts rather than global counts, and was
unconditionally sending the static limit parameter when dynamic limits were
enabled.

Design Choices:
- Updated list_running_ajax to calculate the global count of running jobs.
- Only include dynamic_job_limit or max_running_jobs in the API response when
  the global count actually reaches or exceeds the respective limit.
- Deduplicated configuration extraction in DynamicLimit.pm by mapping over
  the DEFAULTS hash.
- Refactored t/ui/01-list.t to reduce repetition of the scheduler config path
  by using a shared $scheduler_config variable.
- Improved test description strings in t/ui/01-list.t and
  t/26-dynamic-job-limit.t to be self-explanatory, eliminating the need for
  in-file comments.
- Added comprehensive subtests to verify limit reporting behavior with
  filters and both dynamic/static configurations.

Benefits:
- Eliminates misleading "limited by server config" messages.
- Provides accurate server-wide limit status to users even in filtered views.
- Cleaner and more maintainable test and implementation code.

Related issue: https://progress.opensuse.org/issues/198770

### DIFF
--- a/lib/OpenQA/Scheduler/DynamicLimit.pm
+++ b/lib/OpenQA/Scheduler/DynamicLimit.pm
@@ -60,23 +60,18 @@ sub _resolve_threshold ($configured, $factor) {
 # DynamicLimit to the config key names.
 sub _extract_config ($config) {
     my $defaults = DEFAULTS();
+    my %cfg = map { $_ => $config->{$_} // $defaults->{$_} } keys %$defaults;
     return {
-        threshold => _resolve_threshold(
-            $config->{dynamic_job_limit_load_threshold} // $defaults->{dynamic_job_limit_load_threshold},
-            $config->{dynamic_job_limit_load_threshold_factor} // $defaults->{dynamic_job_limit_load_threshold_factor}
-        ),
-        critical => _resolve_threshold(
-            $config->{dynamic_job_limit_load_critical} // $defaults->{dynamic_job_limit_load_critical},
-            $config->{dynamic_job_limit_load_critical_factor} // $defaults->{dynamic_job_limit_load_critical_factor}
-        ),
-        step => $config->{dynamic_job_limit_step} // $defaults->{dynamic_job_limit_step},
-        min => $config->{dynamic_job_limit_min} // $defaults->{dynamic_job_limit_min},
-        max => $config->{max_running_jobs} // $defaults->{max_running_jobs},
-        interval => $config->{dynamic_job_limit_interval} // $defaults->{dynamic_job_limit_interval},
-        scale_up_hysteresis => $config->{dynamic_job_limit_scale_up_hysteresis}
-          // $defaults->{dynamic_job_limit_scale_up_hysteresis},
-        fast_ramp_up_load_factor => $config->{dynamic_job_limit_fast_ramp_up_load_factor}
-          // $defaults->{dynamic_job_limit_fast_ramp_up_load_factor},
+        threshold =>
+          _resolve_threshold($cfg{dynamic_job_limit_load_threshold}, $cfg{dynamic_job_limit_load_threshold_factor}),
+        critical =>
+          _resolve_threshold($cfg{dynamic_job_limit_load_critical}, $cfg{dynamic_job_limit_load_critical_factor}),
+        step => $cfg{dynamic_job_limit_step},
+        min => $cfg{dynamic_job_limit_min},
+        max => $cfg{max_running_jobs},
+        interval => $cfg{dynamic_job_limit_interval},
+        scale_up_hysteresis => $cfg{dynamic_job_limit_scale_up_hysteresis},
+        fast_ramp_up_load_factor => $cfg{dynamic_job_limit_fast_ramp_up_load_factor},
     };
 }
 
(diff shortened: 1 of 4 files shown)


kalikiana (approved):

okurz:
CI failures introduced by rebasing fixed again.

t/26-dynamic-job-limit.t
```diff
@@ -116,9 +116,9 @@ my @cases = (
 
 test_dynamic_limit(%$_) for @cases;
 
-subtest 'auto-detects threshold from nproc when configured as 0' => sub {
+subtest 'auto-detect threshold (nproc * factor) and fast ramp-up' => sub {
     my $mock_dl = Test::MockModule->new('OpenQA::Scheduler::DynamicLimit');
-    $mock_dl->mock(_nproc => sub { 4 });    # 4 CPUs; threshold = 4*0.85 = 3.4; load 1.0 < 3.4*0.3=1.02 => fast up
+    $mock_dl->mock(_nproc => sub { 4 });
```
Wabri: question(non-blocking): why did you remove the comment? I think it's quite useful to understand what is going on here!

okurz: ok, restored.

okurz:
suggestion by @Wabri incorporated

lib/OpenQA/Scheduler/DynamicLimit.pm
```diff
@@ -60,26 +60,21 @@ sub _resolve_threshold ($configured, $factor) {
 # DynamicLimit to the config key names.
 sub _extract_config ($config) {
     my $defaults = DEFAULTS();
+    my %c = map { $_ => $config->{$_} // $defaults->{$_} } keys %$defaults;
```
d3flex: Due to the controller convention which use `$c`, this might just be better be renamed to `cfg`

d3flex (changes requested):
`my $global_running = $self->schema->resultset('Jobs')->count({state =>` seems very much similar to the `my $running = $self->schema->resultset('Jobs')->complex_query(` 
not sure is you really need a second query

Related issue: https://progress.opensuse.org/issues/198770 is resolved btw. is that the correct issue link?

lib/OpenQA/Scheduler/DynamicLimit.pm
```diff
@@ -60,26 +60,21 @@ sub _resolve_threshold ($configured, $factor) {
...
+        step => $c{dynamic_job_limit_step},
+        min => $c{dynamic_job_limit_min},
+        max => $c{max_running_jobs},
+        interval => $c{dynamic_job_limit_interval},
+        scale_up_hysteresis => $c{dynamic_job_limit_scale_up_hysteresis},
+        fast_ramp_up_load_factor => $c{dynamic_job_limit_fast_ramp_up_load_factor},
     };
 }
 
+
```
Martchus:
```suggestion
```

Martchus (approved):

okurz:
> `my $global_running = $self->schema->resultset('Jobs')->count({state =>` seems very much similar to the `my $running = $self->schema->resultset('Jobs')->complex_query(` not sure is you really need a second query

Ok, optimized the global job count in lib/OpenQA/WebAPI/Controller/Test.pm. Now uses the already-fetched job list size when no filters are applied, avoiding a redundant database query.

> Related issue: https://progress.opensuse.org/issues/198770 is resolved btw. is that the correct issue link?

yes, correct

lib/OpenQA/Scheduler/DynamicLimit.pm
okurz: done

lib/OpenQA/Scheduler/DynamicLimit.pm
okurz: done

lib/OpenQA/WebAPI/Controller/Test.pm
```diff
@@ -343,14 +343,26 @@ sub list_running_ajax ($self) {
...
     my %response = (data => \@running);
-    my $config = OpenQA::App->singleton->config->{scheduler};
+    my $app = OpenQA::App->singleton;
+    my $config = $app->config->{scheduler};
     my $max_running = $config->{max_running_jobs};
+    my $has_filters
+      = $self->get_match_param
+      || $self->param('comment')
+      || $self->param('groupid')
+      || $self->every_key_value_param('job_setting');
```
Wabri: question(blocking): I'm not sure about this, if `every_key_value_param('job_setting')` found no match could this lead into a always true condition?

perlpunk: I think so. `every_key_value_param` always returns a true value (a hashref).
That's probably not wanted here...

okurz: So, let's check the value behind that, right? Changed to

```suggestion
    my $has_filters
      = $self->get_match_param
      || $self->param('comment')
      || $self->param('groupid')
      || %{$self->every_key_value_param('job_setting')};
```

Wabri: It could be only if the `every_key_value_param` don't throw an undef

okurz:
@perlpunk I found the OBS test failure now in this PR as well

d3flex (approved):

Martchus (approved):

PR #7244, opened 2026-04-02
test: Fix flaky full-stack tests by mocking service ports

okurz:
Motivation:
Multiple full-stack tests were failing intermittently when run in parallel
due to "Address already in use" errors. They all defaulted to the same
base port (9526), causing conflicts in concurrent test environments.

Design Choices:
- Restored mock_service_ports in OpenQA::Test::Utils which was previously
  removed as dead code in 8d7b9a590. It mocks OpenQA::Utils::service_port
  via Test::MockModule to allocate independent free ports per service using
  Mojo::IOLoop::Server->generate_port, which asks the OS for a truly free
  port instead of blindly guessing.
- Applied mock_service_ports to t/05-scheduler-full.t, t/25-cache-service.t,
  t/33-developer_mode.t, t/43-scheduling-and-worker-scalability.t, and
  t/full-stack.t, replacing the prior setup_random_base_port approach.
- Each service gets an independently chosen free port (not a fixed-offset
  block), eliminating the risk that one occupied port breaks all services.
- Removed duplicate randomization logic and redundant comments across
  multiple test files.
- Cleaned up BEGIN blocks and imports in full-stack tests.

Benefits:
- Guarantees free ports via OS-level allocation, eliminating race conditions
  that setup_random_base_port could not prevent.
- Eliminates port conflicts during parallel test execution.
- Improves test stability and reliability for CI/CD.
- Centralizes logic for easier future maintenance and consistent
  port selection strategy.
- Cleaner and more maintainable test code.

### DIFF
--- a/t/lib/OpenQA/Test/Utils.pm
+++ b/t/lib/OpenQA/Test/Utils.pm
@@ -49,6 +49,7 @@ BEGIN {
 }
 
 our @EXPORT_OK = qw(
+  assign_free_service_ports
   setup_mojo_app_with_default_worker_timeout create_user_for_workers
   create_webapi create_websocket_server create_scheduler create_live_view_handler
   unresponsive_worker broken_worker rejective_worker setup_share_dir setup_fullstack_temp_dir run_gru_job
@@ -69,6 +70,12 @@ our @EXPORT_OK = qw(
 # reuse it in listen to prevent race condition
 #
 # Potentially this approach can also be used in production code.
+
+sub assign_free_service_ports {
+    for my $service (qw(webui websocket livehandler scheduler cache_service)) {
+        $ENV{"OPENQA_PORT_\U$service"} //= Mojo::IOLoop::Server->generate_port;
+    }
+}
 
 sub setup_mojo_app_with_default_worker_timeout ($class = 'Mojolicious') {
     my $app = $class->new(config => {global => {worker_timeout => DEFAULT_WORKER_TIMEOUT}}, log => undef);
--- a/lib/OpenQA/Utils.pm
+++ b/lib/OpenQA/Utils.pm
@@ -803,6 +803,10 @@ sub set_listen_address ($port) {
 }
 
 sub service_port ($service) {
+    # \U uppercases the interpolated service name for the environment variable lookup
+    if (my $env_port = $ENV{"OPENQA_PORT_\U$service"}) {
+        return $env_port;
+    }
     my $base = $ENV{OPENQA_BASE_PORT} ||= DEFAULT_OPENQA_BASE_PORT;
     my $offsets = {
         webui => 0,
--- a/t/05-scheduler-full.t
+++ b/t/05-scheduler-full.t
@@ -27,20 +27,20 @@ use OpenQA::Scheduler::Client;
 use OpenQA::Scheduler::Model::Jobs;
 use OpenQA::WebAPI;
 use OpenQA::Worker::WebUIConnection;
-use OpenQA::Utils;
+use OpenQA::Utils qw(service_port);
 require OpenQA::Test::Database;
 use OpenQA::Test::Utils qw(
-  setup_mojo_app_with_default_worker_timeout
+  assign_free_service_ports setup_mojo_app_with_default_worker_timeout
   setup_fullstack_temp_dir create_user_for_workers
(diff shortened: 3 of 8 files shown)


t/lib/OpenQA/Test/Utils.pm
```diff
@@ -60,16 +61,9 @@ our (@EXPORT, @EXPORT_OK);
...
-# sequential range starting with OPENQA_BASE_PORT. This can cause problems
-# especially in repeated testing if any of the ports in that range is already
-# occupied. So we inject random, free ports for the services here.
-#
-# Potential point for later improvement: In
-# Mojo::IOLoop::Server::generate_port keep the sock object on the port and
-# reuse it in listen to prevent race condition
-#
-# Potentially this approach can also be used in production code.
+sub setup_random_base_port {
```
perlpunk: Now I'm wondering why `mock_service_ports` was removed: 8d7b9a5905051ca2b30ccbeffe51c7d9676c6991
I guess the according comment block was overlooked.

okurz: huh. That's why this comment block was a bit dangling :D

Martchus (approved):

t/lib/OpenQA/Test/Utils.pm
okurz: replaced all the randomization by using re-introduced `mock_service_ports`

Martchus (approved):

Martchus (approved):

t/lib/OpenQA/Test/Utils.pm
```diff
@@ -70,6 +71,12 @@ our @EXPORT_OK = qw(
 #
 # Potentially this approach can also be used in production code.
 
+sub assign_free_service_ports {
+    for my $service (qw(webui websocket livehandler scheduler cache_service)) {
+        $ENV{"OPENQA_PORT_\U$service"} //= Mojo::IOLoop::Server->generate_port;
```
Martchus, line 76: We already assign `$ENV{OPENQA_BASE_PORT} ||= Mojo::IOLoop::Server->generate_port;` in the regular fullstack test and the developer fullstack test. Maybe it is good enough to extend that to the other fullstack tests.

(That we run into https://progress.opensuse.org/issues/202662 means that this is perhaps indeed not good enough. However, then I'm also wondering whether this approach which is also just using `Mojo::IOLoop::Server->generate_port` in a non-race-free manner is good enough.)

Martchus (changes requested):
> They all defaulted to the same base port (9526), causing conflicts in concurrent test environments.

Remove that from the commit message. At least the regular and devel fullstack test do *not* default to the same base ports. I've just verified that locally, see https://progress.opensuse.org/issues/202662#note-9.

Martchus:
I've also just ran the regular and devel fullstack tests in parallel on my machine and the problem there was that both use the same ports on os-autoinst level. The command server ports were definitely conflicting but perhaps the same counts for the VNC ports.

EDIT: I ran the tests again and couldn't reproduce it anymore. It looks like we pick a random number between 1 and 64 for the worker instance number. That means this conflict happens only around 1.56 % of the time (1/64).

Martchus:
If I get https://github.com/os-autoinst/openQA/pull/7573 working we don't need this PR anymore. https://github.com/os-autoinst/openQA/pull/7573 is preferable as it would resolve the problem without leaving a small window for race conditions.

t/lib/OpenQA/Test/Utils.pm
okurz: simply setting OPENQA_BASE_PORT still relies on hardcoded offsets (e.g., base + 1, base + 2). As we discussed today this still makes zero guarantees about the offset ports.

By using assign_free_service_ports, we force generate_port to independently query the OS and validate a free socket for every individual service (webui, websocket, scheduler, etc.). 

You are correct that generate_port isn't entirely race-free between the port being returned and the service binding to it. Still, the chance for this being a problem is likely still significantly lower than just blindly offsetting from a single verified port. I think there is an approach where we can bind a socket and re-use it or something?

Martchus: Yes, see https://github.com/os-autoinst/openQA/pull/7573 for a race-free alternative.

PR #7252, opened 2026-04-08
fix: Apply force result label correctly if `force_result_regex` is empty

Martchus:
This seems to work as-is. I also temporarily removed the code responsible for this in `sub carry_over_bugrefs` which caused the newly added checks to fail.

Related ticket: https://progress.opensuse.org/issues/199124

### DIFF
--- a/etc/openqa/openqa.ini
+++ b/etc/openqa/openqa.ini
@@ -103,9 +103,9 @@
 ## The maximum amount of times a job will be restarted if the auto_clone_regex matches
 #auto_clone_limit = 20
 
-## A regex pattern that a "force_result" label description in job comments
-## must match to be accepted. If undefined and by default no rules are applied
-## and no description is expected
+## A regex pattern that a "force_result" label description in job comments must
+## match to be applied. If empty (the default), all "force_result" labels are
+## applied.
 #force_result_regex =
 
 ## Job results to collapse by default as parallel children on the test result
--- a/t/17-labels_carry_over.t
+++ b/t/17-labels_carry_over.t
@@ -138,7 +138,7 @@ $schema->txn_rollback;
 $schema->txn_begin;
 
 subtest 'failed->failed flag:carryover comments are carried over' => sub {
-    my $comment_text = 'flag:carryover label:force_result:softfailed:bsc#1257825';
+    my $comment_text = "flag:carryover label:force_result:softfailed:bsc#1257825\nAutomatic carryover";
     $t->post_ok("/api/v1/jobs/$old_job/comments", $auth => form => {text => $comment_text})->status_is(200);
     $t->post_ok("/api/v1/jobs/$job/set_done", $auth => form => {result => FAILED})->status_is(200);
     my @comments_new = @{comments("/tests/$job")};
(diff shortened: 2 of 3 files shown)


t/17-labels_carry_over.t
```diff
@@ -138,10 +138,13 @@ $schema->txn_rollback;
 $schema->txn_begin;
 
 subtest 'failed->failed flag:carryover comments are carried over' => sub {
-    $t->post_ok("/api/v1/jobs/$old_job/comments", $auth => form => {text => 'flag:carryover'})->status_is(200);
-    $t->post_ok("/api/v1/jobs/$job/set_done", $auth => form => {result => 'failed'})->status_is(200);
+    my $comment_text = 'flag:carryover label:force_result:softfailed:bsc#1257825';
```
perlpunk: I can make the test fail with this:

```suggestion
    my $comment_text = 'flag:carryover label:force_result:softfailed:bsc#1257825 Automatic carryover';
```
See my comment https://progress.opensuse.org/issues/199124#note-5 regarding the special (and annoying) empty pattern in perl

Martchus: > See my comment https://progress.opensuse.org/issues/199124#note-5 regarding the special (and annoying) empty pattern in perl

I'll change the check according to your comment.

> I can make the test fail with this:

~~Ok, but the comment being carried over in production was literally just "flag:carryover label:force_result:softfailed:bsc#1257825" (see https://openqa.suse.de/tests/21624088#comment-2457094).
But I guess that's nevertheless something we should improve.~~ I now see that you added "Automatic carryover" to trigger the problem with the empty pattern.

Martchus: I adjusted the test and the code accordingly. Thanks for your help, I guess it would have taken me quite a while to figure out this very non-intuitive behavior of Perl.

perlpunk (approved):

etc/openqa/openqa.ini
```diff
@@ -103,9 +103,9 @@
 ## The maximum amount of times a job will be restarted if the auto_clone_regex matches
 #auto_clone_limit = 20
 
-## A regex pattern that a "force_result" label description in job comments
-## must match to be accepted. If undefined and by default no rules are applied
-## and no description is expected
+## A regex pattern that a "force_result" label description in job comments  must
+## match to be applied. If empty (the default), all "force_result" labes are
```
kalikiana:
```suggestion
## match to be applied. If empty (the default), all "force_result" labels are
```

etc/openqa/openqa.ini
```diff
@@ -103,9 +103,9 @@
 ## The maximum amount of times a job will be restarted if the auto_clone_regex matches
 #auto_clone_limit = 20
 
-## A regex pattern that a "force_result" label description in job comments
-## must match to be accepted. If undefined and by default no rules are applied
-## and no description is expected
+## A regex pattern that a "force_result" label description in job comments  must
```
okurz:
```suggestion
## A regex pattern that a "force_result" label description in job comments must
```

okurz (dismissed):

kalikiana (approved):

okurz (approved):

perlpunk:
We cannot even force merge because of the broken WIP check

perlpunk:
I merged it locally

PR #7255, opened 2026-04-09
fix(apparmor): forbid breaking out of confinement with ssh again

okurz:
This fixes a security regression introduced in 4ec6f2aa8 allowing to
execute ssh globally in the complete worker process. Allowing to ssh out
from the main worker process would deny most of the confinement benefit
of using apparmor.

Related to
https://github.com/os-autoinst/openQA/pull/7249#discussion_r3049550719


profiles/apparmor.d/usr.share.openqa.script.worker
```diff
@@ -253,6 +252,7 @@
     /proc/*/fd/ r,
     /proc/filesystems r,
     /proc/meminfo r,
+    /usr/bin/ssh rix,
```
Wabri, line 255: question: I don't get it, why did you move that line?

perlpunk (approved):

profiles/apparmor.d/usr.share.openqa.script.worker
okurz: I partially revert https://github.com/os-autoinst/openQA/pull/7249/files where one commit removed the line from one process group and moved it to the general worker rules. I guess we are fine to still allow ssh for the specific non-confined support processes even though I don't think we even had or can enforce apparmor in that context. That is why we have https://progress.opensuse.org/projects/openqav3/wiki/#o3-s390-and-other-eg-bare-metal-workers-running-as-podman-containers-on-o3-worker-openqaworker23

kalikiana (approved):

Wabri (approved):

Martchus:
Maybe a stupid question: Don't certain backends *need* to make SSH connections? I suppose they don't do them via the `ssh` command, though. But does it really matter what tool/library is used? As long as we can make SSH connections in general I don't see a big improvement here.

I guess it makes generally sense to trim down the number of executables to a minimum but the title "forbid breaking out of confinement" makes it sound like this is about a specific attack vector being prevented here (SSH connections in general).

Vogtinator (changes requested):
This will break ssh backends again

Vogtinator:
> Maybe a stupid question: Don't certain backends need to make SSH connections? I suppose they don't do them via the ssh command, though.

They actually do: The worker spawns Xvnc, with icewm and xterm-console inside, then runs ssh.

https://github.com/os-autoinst/openQA/commit/4ec6f2aa8788c1a4914ef6755be2818275dec2da was necessary to make this work again.

> But does it really matter what tool/library is used? As long as we can make SSH connections in general I don't see a big improvement here.

Correct. Any code, including test code can just connect to whereever using TCP sockets.

profiles/apparmor.d/usr.share.openqa.script.worker
Vogtinator, line 255: x3270 does not invoke ssh

Vogtinator: > though I don't think we even had or can enforce apparmor in that context. That is why we have https://progress.opensuse.org/projects/openqav3/wiki/#o3-s390-and-other-eg-bare-metal-workers-running-as-podman-containers-on-o3-worker-openqaworker23

Actually, the profile from the worker is enforced inside the containers as well, possibly explicitly configured that way. That's why PR 7249 was necessary in the first place.

kalikiana:
> > But does it really matter what tool/library is used? As long as we can make SSH connections in general I don't see a big improvement here.
> 
> Correct. Any code, including test code can just connect to whereever using TCP sockets.

Let's verify this properly then. https://openqa.opensuse.org/tests/5827767 is the test that was used to verify #7249 - are there other cases we should consider?

Vogtinator:
> Let's verify this properly then. https://openqa.opensuse.org/tests/5827767 is the test that was used to verify https://github.com/os-autoinst/openQA/pull/7249 - are there other cases we should consider?

That should work, also svirt, pvm and generalhw. Also any git repo referenced through e.g. `CASEDIR` may do a clone over ssh.

okurz:
> Maybe a stupid question: Don't certain backends _need_ to make SSH connections?

Of course. See https://github.com/os-autoinst/openQA/pull/7255#discussion_r3059334108

> I suppose they don't do them via the `ssh` command, though. But does it really matter what tool/library is used? As long as we can make SSH connections in general I don't see a big improvement here.

What "improvement" do you mean? I am trying to break a security regression from https://github.com/os-autoinst/openQA/commit/4ec6f2aa8788c1a4914ef6755be2818275dec2da

> I guess it makes generally sense to trim down the number of executables to a minimum but the title "forbid breaking out of confinement" makes it sound like this is about a specific attack vector being prevented here (SSH connections in general).

please read the commit message and tell me where I can phrase it better. This PR is intended to be a security regression fix, not a general improvement.

Martchus (changes requested):
Then I don't get what regression this fixes considering code can make SSH connections anyway (e.g. using libopenssh directly or some custom implementation).

Considering @Vogtinator's comments this will even break existing tests and allows using `ssh` in the wrong place.

okurz:
> Then I don't get what regression this fixes considering code can make SSH connections anyway (e.g. using libopenssh directly or some custom implementation).
> 
> Considering @Vogtinator's comments this will even break existing tests and allows using `ssh` in the wrong place.

You should know even better. We have apparmor enforced in o3 except for https://progress.opensuse.org/projects/openqav3/wiki/#o3-s390-and-other-eg-bare-metal-workers-running-as-podman-containers-on-o3-worker-openqaworker23 . And we don't enforce apparmor within the whole OSD domain. Why is that then?

Vogtinator:
> We have apparmor enforced in o3 except for https://progress.opensuse.org/projects/openqav3/wiki/#o3-s390-and-other-eg-bare-metal-workers-running-as-podman-containers-on-o3-worker-openqaworker23

The apparmor denials caused failures in the containerized workers and adjusting the policy on ow23 (the container host!) made it work again. Not sure which part of that is intentional.

In any case, the tests should work even outside of containers, like on local workstations with apparmor enabled.

okurz:
> > We have apparmor enforced in o3 except for https://progress.opensuse.org/projects/openqav3/wiki/#o3-s390-and-other-eg-bare-metal-workers-running-as-podman-containers-on-o3-worker-openqaworker23
> 
> The apparmor denials caused failures in the containerized workers and adjusting the policy on ow23 (the container host!) made it work again. Not sure which part of that is intentional.
> 
> In any case, the tests should work even outside of containers, like on local workstations with apparmor enabled.

That all sounds nice and dandy but doesn't explain why someone (not me) made the decision that o3 can have enforced apparmor profiles for all standard qemu workers, apparmor disabled but confined in containers in o3 for non-qemu and no apparmor enabled at all within OSD

Vogtinator:
> > > We have apparmor enforced in o3 except for https://progress.opensuse.org/projects/openqav3/wiki/#o3-s390-and-other-eg-bare-metal-workers-running-as-podman-containers-on-o3-worker-openqaworker23
> > 
> > 
> > The apparmor denials caused failures in the containerized workers and adjusting the policy on ow23 (the container host!) made it work again. Not sure which part of that is intentional.
> > In any case, the tests should work even outside of containers, like on local workstations with apparmor enabled.
> 
> That all sounds nice and dandy but doesn't explain why someone (not me) made the decision that o3 can have enforced apparmor profiles for all standard qemu workers, apparmor disabled but confined in containers in o3 for non-qemu

It's enabled on all workers FWICT? For the container workers I suspect the difference in behaviour started with Leap 16. Maybe some product regression there (the container engines no longer support AppArmor?).

> and no apparmor enabled at all within OSD

Would be nice if that could change.

Vogtinator:
> (the container engines no longer support AppArmor?).

FWICT that is indeed the case. On ow23, `podman info` prints:

```
  security:
    apparmorEnabled: false
    capabilities: CAP_CHOWN,CAP_DAC_OVERRIDE,CAP_FOWNER,CAP_FSETID,CAP_KILL,CAP_NET_BIND_SERVICE,CAP_SETFCAP,CAP_SETGID,CAP_SETPCAP,CAP_SETUID,CAP_SYS_CHROOT
    rootless: false
    seccompEnabled: true
    seccompProfilePath: /usr/share/containers/seccomp.json
    selinuxEnabled: false
```

so it does not load an apparmor profile and it apparently ends up inheriting the host confinement. No idea if that's how it should be, but it is rather questionable. It definitely makes AppArmor on Leap 16 less usable.

With that it mind, it may be a valid option to disable AppArmor on ow23 as it's not working as intended, but arguably that makes it less secure.

I'd still like to understand how being able to run ssh (or connect to port 22) allows test or worker code to break out of containment, IMO that should be cleared up in any case.

okurz:
thanks for your assessment. Based on https://github.com/os-autoinst/openQA/pull/7255#issuecomment-4259354774 then I will close this PR and we will accept the current state acceptable giving us even the option to enforce apparmor on more hosts with of course a long-term migration to better supported solutions as alternative

PR #7258, opened 2026-04-10
feat: faster comment_data_for_jobs

Vogtinator:
Most of the time spent in the dashboard calculation was due to passing thousands of IDs as `WHERE ID IN ($1, $2, $3, ...., $33567)`.

If $jobs is a result set, it's much faster to use the source query directly
instead of fetching all ids and passing them individually.

There is a theoretical change in behaviour: If the database changes in
between the jobs and comments queries, it may now miss comments for jobs
that no longer match the query.

This improves performance noticably:

Before:
```
> time curl -s http://fussheizung/dashboard_build_results?group=Investigation >/dev/null

real    0m29,465s
user    0m0,005s
sys     0m0,008s
```

After:

```
> time curl -s http://fussheizung/dashboard_build_results?group=Investigation >/dev/null

real    0m7,773s
user    0m0,008s
sys     0m0,002s
```

### DIFF
--- a/lib/OpenQA/Schema/ResultSet/Comments.pm
+++ b/lib/OpenQA/Schema/ResultSet/Comments.pm
@@ -99,8 +99,9 @@ if you need the Bug objects themselves.
 =cut
 
 sub comment_data_for_jobs ($self, $jobs, $args = {}) {
-    my @job_ids = map { $_->id } ref $jobs eq 'ARRAY' ? @$jobs : $jobs->all;
-    my $comments = $self->search({job_id => {in => \@job_ids}}, {order_by => 'me.id', select => [qw(text job_id)]});
+    my $job_id_clause = ref $jobs eq 'ARRAY' ? [map { $_->id } @$jobs] : $jobs->get_column('id')->as_query;
+    my %args = (order_by => 'me.id', select => [qw(text job_id)]);
+    my $comments = $self->search({job_id => {in => $job_id_clause}}, \%args);
     my $bugs = $self->result_source->schema->resultset('Bugs');
 
     my (%res, %bugdetails);
--- a/lib/OpenQA/BuildResults.pm
+++ b/lib/OpenQA/BuildResults.pm
@@ -206,7 +206,7 @@ sub compute_build_results ($group, $limit, $time_limit_days, $tags, $subgroup_fi
             $seen{$key}++ ? () : $_;
         } @all_jobs;
         next unless @jobs;
-        my $comment_data = $group->result_source->schema->resultset('Comments')->comment_data_for_jobs(\@jobs);
+        my $comment_data = $group->result_source->schema->resultset('Comments')->comment_data_for_jobs($jobs);
         for my $job (@jobs) {
             $jr{distris}->{$job->DISTRI} = 1;
             if ($newest) {


Vogtinator:
CC @okurz

lib/OpenQA/Schema/ResultSet/Comments.pm
```diff
@@ -99,8 +99,17 @@ if you need the Bug objects themselves.
...
+    if (ref $jobs eq 'ARRAY') {
+        my @job_ids = map { $_->id } @$jobs;
+        $comments = $self->search({job_id => {in => \@job_ids}}, {order_by => 'me.id', select => [qw(text job_id)]});
+    }
+    else {
+        $comments = $self->search(
+            {job_id => {in => $jobs->get_column('id')->as_query}},
+            {order_by => 'me.id', select => [qw(text job_id)]});
+    }
+
```
okurz: this looks a bit duplicate. We can extract the value for the "in" parameter, e.g. something like

```
    my $jobs_ref = ref $jobs eq 'ARRAY' ? { map { $_->id } @$jobs } : $jobs->get_column('id')->as_query;
    my $comments = $self->search({job_id => {in => $jobs_ref}}, {order_by => 'me.id', select => [qw(text job_id)]});
```

okurz (changes requested):
Great idea! This goes in the same direction as I was heading with #7026 which became much more complicated and elaborate. If we can get all tests to pass then we should merge this first and I will happily rebase/rework #7026 on top

lib/OpenQA/Schema/ResultSet/Comments.pm
Vogtinator: Yeah, I wasn't sure how to handle the different types involved there with perl.

`{ map { $_->id } @$jobs }` didn't work, but apparently `[ map { $_->id } @$jobs ]` does. Still readable!

Vogtinator:
Tests passed, but the WIP app is down...

Bildschirmfoto_20260410_105709

https://github.com/wip/app/issues/951

lib/OpenQA/Schema/ResultSet/Comments.pm
```diff
@@ -99,8 +99,10 @@ if you need the Bug objects themselves.
 =cut
 
 sub comment_data_for_jobs ($self, $jobs, $args = {}) {
-    my @job_ids = map { $_->id } ref $jobs eq 'ARRAY' ? @$jobs : $jobs->all;
-    my $comments = $self->search({job_id => {in => \@job_ids}}, {order_by => 'me.id', select => [qw(text job_id)]});
+    my $job_id_clause = ref $jobs eq 'ARRAY' ? [map { $_->id } @$jobs] : $jobs->get_column('id')->as_query;
+    my $comments
+      = $self->search({job_id => {in => $job_id_clause}}, {order_by => 'me.id', select => [qw(text job_id)]});
+
```
okurz:
```suggestion
    my %args = (order_by => 'me.id', select => [qw(text job_id)]);
    my $comments = $self->search({job_id => {in => $job_id_clause}}, \%args);
```

also please delete the blank line afterwards

Vogtinator: Done.

okurz:
CI fails with a seemingly unrelated error:
```

        #   Failed test 'all workers have a job or all jobs assigned'
        #   at t/43-scheduling-and-worker-scalability.t line 198.
        # Allocated count: 2, expected at least: 3

    #   Failed test 'assign and run jobs'
    #   at t/43-scheduling-and-worker-scalability.t line 218.

#   Failed test 'Scalability test with 5 worker(s) and 3 job(s)'
#   at t/43-scheduling-and-worker-scalability.t line 240.
# Looks like you failed 1 test of 4.
[12:18:18] t/43-scheduling-and-worker-scalability.t .................. 
Failed 1/4 subtests
```

I will handle this and restarted the CI test.

okurz (approved):

Martchus (approved):

okurz:
Next to the broken "WIP" we are still missing one check, likely from OBS which currently has performance problems. Merging manually.

PR #7260, opened 2026-04-10
feat: Prevent job assignments when running out of space by default

Martchus:
Set a low default value of 5 % for `results_min_free_storage_space_percentage`. We usually set higher thresholds for cleanup and alerting so 5 % makes sense for this measure which should only be the *last* measure to prevent incomplete jobs and running completely out of storage space.

Related ticket: https://progress.opensuse.org/issues/198848

### DIFF
--- a/lib/OpenQA/Setup.pm
+++ b/lib/OpenQA/Setup.pm
@@ -159,6 +159,7 @@ sub default_config () {
             dynamic_job_limit_load_critical => 0,
             dynamic_job_limit_step => 10,
             dynamic_job_limit_interval => 60,
+            results_min_free_storage_space_percentage => 0,
         },
         logging => {
             level => undef,
--- a/lib/OpenQA/Setup.pm
+++ b/lib/OpenQA/Setup.pm
@@ -159,7 +159,7 @@ sub default_config () {
             dynamic_job_limit_load_critical => 0,
             dynamic_job_limit_step => 10,
             dynamic_job_limit_interval => 60,
-            results_min_free_storage_space_percentage => 0,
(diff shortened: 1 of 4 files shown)


kalikiana (approved):

okurz (approved):

lib/OpenQA/Setup.pm
```diff
@@ -321,7 +322,11 @@ sub read_config ($app) {
     my $defaults = default_config();
 
     # in development and test mode we use fake auth and log to stderr
-    my %devel_and_test_defaults = (auth => {method => 'Fake'}, logging => {file => undef, level => 'debug'});
+    my %devel_and_test_defaults = (
+        auth => {method => 'Fake'},
+        logging => {file => undef, level => 'debug'},
+        scheduler => {results_min_free_storage_space_percentage => 0},
```
okurz, line 329: why is this needed?

okurz (changes requested):

lib/OpenQA/Setup.pm
Martchus: CI checks were failing. The same tests passed for me locally. I could make them fail locally as in the CI env by mocking df so I have only very little disk space left. So that's why they were failing in the CI env.

I think we shouldn't make the outcome of our tests depend on how much disk space is left - especially since the jobs scheduled in most tests aren't even creating a big amount of results.

Of course there's still the test of the feature itself which already sets `results_min_free_storage_space_percentage` to some specific value and also mocks df for consist results.

okurz (approved):

PR #7262, opened 2026-04-10
fix: address remaining XSS vulnerabilities using safe DOM APIs

okurz:
Motivation:
CodeQL identified cross-site scripting (XSS) vulnerabilities where
untrusted exception messages and server responses were rendered as HTML.
The previous approach of manually escaping at every call site was
error-prone and hard to maintain.

Design Choices:
1.  Moved the `createElement` helper to `openqa.js` to make it globally
    available for safe UI construction.
2.  Refactored `makeFlashElement` to use `textContent` by default for
    strings, ensuring all text passed to `addFlash` is safely escaped.
3.  Updated `addFlash` to accept both safe strings and DOM elements,
    providing a concise but secure API for callers.
4.  Refactored `job_templates.js`, `tests.js`, `overview.js`,
    `needleeditor.js`, and `create_tests.js` to use the `createElement`
    helper instead of interpolated HTML strings for status messages.
5.  Removed redundant `htmlEscape` calls where standard DOM APIs now
    handle the escaping implicitly.

Benefits:
-   Eliminates XSS vulnerabilities by securing the underlying DOM sinks.
-   Reduces code duplication and boilerplate through centralized helpers.
-   Improves long-term maintainability by adopting a consistent, safe
    pattern for dynamic UI updates across the project.

Related issue: https://github.com/os-autoinst/openQA/security/code-scanning/11
Related issue: https://github.com/os-autoinst/openQA/security/code-scanning/12

### DIFF
--- a/assets/javascripts/create_tests.js
+++ b/assets/javascripts/create_tests.js
@@ -54,7 +54,14 @@ function createTests(form) {
     success: function (response) {
       const id = response.scheduled_product_id;
       const url = `${form.dataset.productlogUrl}?id=${id}`;
-      addFlash('info', `Tests have been scheduled, checkout the <a href="${url}">product log</a> for details.`);
+      addFlash(
+        'info',
+        createElement('span', [
+          'Tests have been scheduled, checkout the ',
+          createElement('a', ['product log'], {href: url}),
+          ' for details.'
+        ])
+      );
     },
     error: function (xhr, ajaxOptions, thrownError) {
       addFlash('danger', 'Unable to create tests: ' + (xhr.responseJSON?.error ?? xhr.responseText ?? thrownError));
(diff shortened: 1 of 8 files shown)


Martchus (approved):
Let's hope none of these flash messages contain HTML elements to be rendered.

kalikiana (approved):

Martchus (changes requested):
Looks like the tests have even uncovered a few cases where the HTML elements are actually supposed to be rendered.

Note that this is probably not fixing any XSS vulnerabilities anyway. The HTML code we're inserting here comes from the same Mojolciious app that also served the HTML page to begin with and the Mojo app is expected to escape any user provided data accordingly. So we should not have to take care of this in the frontend as well. So maybe we should simply leave out changes that cause problems and somehow ignore the false alerts.

okurz:
> […] somehow ignore the false alerts.

I am fine with that. I referenced the according alerts. Maybe you had already look into those as you worked on the security scanner feature in before?

EDIT: Found it. There is a "dismiss alert" button and I added an according explanation in there

PR #7263, opened 2026-04-10
feat: Speed up asset uploads

Martchus:
* Increase default upload junk size from 1 MB to 10 MB
* Adjust `client_body_buffer_size` setting as well adding a little bit of headroom
* Define the constant for the default upload junk size only in one place

Related ticket: https://progress.opensuse.org/issues/199397

---

I'm adding the "not-ready" label because ~~I still want to do some local tests and I also need to check whether the NGINX config needs to be changed manually in production~~. It looks like `/etc/nginx/vhosts.d/openqa-endpoints.inc` is modified on o3 and OSD so we'll have to change those files manually. So ~~I'll change~~ I changed the setting there manually.

### DIFF
--- a/lib/OpenQA/Worker/Job.pm
+++ b/lib/OpenQA/Worker/Job.pm
@@ -12,6 +12,7 @@ use OpenQA::Worker::Engines::isotovideo;
 use OpenQA::Worker::Isotovideo::Client;
 use OpenQA::Log qw(log_error log_warning log_debug log_info redact_settings_in_file);
 use OpenQA::Utils qw(find_video_files usleep_backoff);
+use OpenQA::Client::Upload qw(DEFAULT_CHUNK_SIZE);
 
 use Digest::MD5;
 use Fcntl;
@@ -27,8 +28,6 @@ use File::Map 'map_file';
 use List::Util 'max';
 use Time::HiRes qw(time usleep);
 use Feature::Compat::Try;
-
-use constant DEFAULT_UPLOAD_CHUNK_SIZE => 10_000_000;
 
 # define attributes for public properties
 has 'worker';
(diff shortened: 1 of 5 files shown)


okurz (approved):

perlpunk (approved):
Tests need to be adjusted.
Maybe I would have conducted some more tests in production, but if turns out problematic, it's easy to fix

perlpunk:
And maybe it would be good to split the refactoring of the constant and the adjustment of the value to different commits.

lib/OpenQA/Worker/Job.pm
```diff
@@ -12,6 +12,7 @@ use OpenQA::Worker::Engines::isotovideo;
 use OpenQA::Worker::Isotovideo::Client;
 use OpenQA::Log qw(log_error log_warning log_debug log_info redact_settings_in_file);
 use OpenQA::Utils qw(find_video_files usleep_backoff);
+use OpenQA::Client::Upload qw(DEFAULT_CHUNK_SIZE);
```
perlpunk: I guess I would have moved that to the Constants module. `OpenQA::Client::Upload` inherits from `OpenQA::Client::Handler`, which inherits from `Mojo::EventEmitter`.
I would say it's not a module that should be used for providing constants or functions to other modules.

perlpunk (dismissed):

d3flex (approved):

okurz (approved):

perlpunk (commented):
For the recent PR regarding the VNC port it was requested to split the definition of the constant to a separate commit.
I don't know why @okurz approves here without that split.

okurz:
> For the recent PR regarding the VNC port it was requested to split the definition of the constant to a separate commit. I don't know why @okurz approves here without that split.

Because I trust that besides this change what you "requested" @Martchus will not include more

perlpunk (dismissed):
`s/refactor(upload): Define junk size only once/refactor(upload): Define chunk size only once/`

PR #7264, opened 2026-04-10
feat: enable videos for all jobs again

okurz:
openQA disabled recording videos for jobs based on MAX_JOB_TIME. Since
then multiple factors changed. os-autoinst by default records with very
efficient video encoders, e.g. AV1, while on the other hand capacities
and computing powers have increased. For a 1h openQA job a usual video
has a size of only 2MiB so there should be no problem to simply enable
recording videos by default again.

This is a refresh of #4746, rebased with conflicts resolved and an
updated motivation.


aylen384 (approved):

Martchus (approved):

Martchus:
Btw, AV1 is a video *format* (and not a video *encoder*).

PR #7265, opened 2026-04-10
test: fix sporadic failure in scheduling-and-worker-scalability

okurz:
Motivation:
The test was sporadically failing with fewer jobs allocated than expected.
This was likely caused by the scheduler singleton retaining state from
previous scenarios, which is problematic when scenarios use separate
databases but reuse job IDs.

Design Choices:
- Clear the scheduler singleton's scheduled_jobs hash at the start of
  each scenario.
- Improved diagnostics in case of failure to show both job and worker
  status.
- Replaced `$log_jobs` with `$log_status` to provide more context.

### DIFF
--- a/t/43-scheduling-and-worker-scalability.t
+++ b/t/43-scheduling-and-worker-scalability.t
@@ -91,6 +91,10 @@ for my $job_count (@job_counts) {
     subtest "Scalability test with $worker_count worker(s) and $job_count job(s)" => sub {
         # ensure fresh database for each scenario
         my $schema = OpenQA::Test::Database->new->create;
+
+        # ensure the scheduler starts with a fresh state for each scenario
+        # (the scheduler singleton's scheduled_jobs hash must be cleared since we reuse job IDs across fresh databases)
+        OpenQA::Scheduler::Model::Jobs->singleton->scheduled_jobs({});
         local $ENV{OPENQA_DATABASE_SEARCH_PATH} = $schema->search_path_for_tests;
         my $workers_rs = $schema->resultset('Workers');
         my $jobs_rs = $schema->resultset('Jobs');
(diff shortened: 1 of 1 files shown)


t/43-scheduling-and-worker-scalability.t
```diff
@@ -212,9 +221,18 @@ for my $job_count (@job_counts) {
...
+            for my $try (1 .. $polling_tries_workers) {
+                my @not_idle = grep { $_->status ne 'idle' } $workers_rs->all;
+                last if @not_idle == 0;
+                sleep $polling_interval;    # uncoverable statement
+            }
+
             my $done = is $jobs_rs->search({state => DONE})->count, $job_count, 'all jobs done';
             my $passed = is $jobs_rs->search({result => PASSED})->count, $job_count, 'all jobs passed';
-            $log_jobs->() unless $done && $passed;
+            $log_status->() if !$done || !$passed;
```
d3flex: I dont see a good reaon to not use the `unless` here and I had the impression that it is in general in favor in our codebase

Martchus (approved):

t/43-scheduling-and-worker-scalability.t
perlpunk: I also don't see a reason.

okurz: Reverted the change to the condition.

okurz:
addressed https://github.com/os-autoinst/openQA/pull/7265#discussion_r3081098571

d3flex (approved):
I think you could save the `$allocated_count` during the for loop instead of call the  `$allocated_count = $jobs_rs->search({state => {'!=' => SCHEDULED}})->count;` in L205. But it is fine as is. approved

PR #7266, opened 2026-04-10
feat: allow ignoring job groups on dashboard evaluation

okurz:
Motivation:
Some large job groups (like "Investigation" on o3) cause 502 errors when
evaluating the dashboard. Allowing these groups to be ignored reduces the
evaluation load on the main page.

Design Choices:
- Added `ignored_job_groups` configuration to `openqa.ini` for easy global
  management.
- Added `ignore_on_dashboard` boolean column to `job_groups` and
  `job_group_parents` for fine-grained control via the Admin UI.
- Implemented filtering logic in both `OpenQA::WebAPI::Controller::Main`
  (for top-level groups) and `OpenQA::BuildResults` (for subgroups).
- Updated the Admin UI templates to allow toggling this property.

Benefits:
- Improves dashboard performance and stability.
- Provides administrators with flexibility to hide irrelevant or problematic
  job groups from the main page without deleting them.

Screenshot_20260410_212853_oqa_ignore_on_dashboard


lib/OpenQA/Schema/ResultSet/Assets.pm
```diff
@@ -40,7 +40,7 @@ sub register ($self, $type, $name, $options = {}) {
     my $sth_jobs_assets = $dbh->prepare(<<~'END_SQL');
         INSERT INTO jobs_assets (job_id, asset_id, t_created, t_updated, created_by)
                         VALUES (?,      ?,        now(),     now(),      true)
-            ON CONFLICT ON CONSTRAINT jobs_assets_job_id_asset_id DO UPDATE SET created_by = true
+            ON CONFLICT (job_id, asset_id) DO UPDATE SET created_by = true
```
Vogtinator, line 43: Unrelated change?

aylen384 (approved):
LGTM in general

perlpunk (commented):
I don't see a test for `ignored_job_groups`.
Also adding both (config setting and new column) in one commit makes it a bit confusing too see where each is used.
Both should have the same effect, right? If the group is ignored via a database setting or via config - it should not appear on the index dashboard. Should it appear in the parent job group view though?

okurz:
> I don't see a test for `ignored_job_groups`. Also adding both (config setting and new column) in one commit makes it a bit confusing too see where each is used. Both should have the same effect, right?

Yes

> If the group is ignored via a database setting or via config - it should not appear on the index dashboard.

Correct

> Should it appear in the parent job group view though?

yes

lib/OpenQA/Schema/ResultSet/Assets.pm
Martchus: Good question. I suppose the change won't make a difference but why have you changed that?

lib/OpenQA/WebAPI/Controller/Main.pm
```diff
@@ -36,14 +37,18 @@ sub dashboard_build_results ($self) {
     my $regex_problem = $self->regex_problem($group_params, 'group parameter is invalid');
     return $self->render(json => {error => $regex_problem}, status => 400) if $regex_problem;
 
-    my $groups = $self->stash('job_groups_and_parents');
+    my $groups = $self->stash('job_groups_and_parents')
+      // $self->schema->resultset('JobGroupParents')->job_groups_and_parents;
+    my $ignored_job_groups = $config->{ignored_job_groups} // '';
+    my %ignored_groups = map { $_ => 1 } grep { length } map { trim($_) } split /,/, $ignored_job_groups;
```
Martchus: It would be good to avoid computing `%ignored_groups` for each and every request. The code that does the config parsing could already turn `$config->{ignored_job_groups}` into a hash.

lib/OpenQA/BuildResults.pm
```diff
@@ -73,45 +73,52 @@ sub add_review_badge ($build_res) {
...
     # find relevant child groups taking filter into account
     my $child_groups = find_child_groups($group, $subgroup_filter);
     my $group_ids = $child_groups->{group_ids};
     my $children = $child_groups->{children};
 
+    if ($ignored_groups) {
+        my @filtered_ids;
+        my @filtered_children;
+        for my $i (0 .. $#$group_ids) {
+            my $child = $children->[$i];
```
Martchus: Not nice that this is not a for-each loop. Maybe once could just use `$group_ids->[scalar @filtered_ids]` below - although I'm not sure whether that's more readable.

t/ui/14-dashboard.t
```diff
@@ -14,7 +14,9 @@ use OpenQA::Test::TimeLimit '30';
 use OpenQA::Test::Case;
 use OpenQA::SeleniumTest;
 
-OpenQA::Test::Case->new->init_data(fixtures_glob => '01-jobs.pl 02-workers.pl 03-users.pl');
+my $test_case = OpenQA::Test::Case->new;
+my $schema = $test_case->init_data(fixtures_glob => '01-jobs.pl 02-workers.pl 03-users.pl');
+my $t = Test::Mojo->new('OpenQA::WebAPI');
```
Martchus: This creates yet another Mojo app. It would be better to move `Mojo::Test`-based tests to e.g. `t/22-dashboard.t` and only do Selenium-based testing in `t/ui/…`.

t/ui/14-dashboard.t
```diff
@@ -143,7 +145,36 @@ subtest 'filter form' => sub {
...
+
+    # 1. Ignore via job group property
+    $group2->update({ignore_on_dashboard => 1});
+    $driver->get('/');
+    wait_for_ajax;
+    is scalar @{$driver->find_elements('h2', 'css')}, 1, 'opensuse test ignored via property';
+    is $driver->find_element('h2')->get_text(), 'opensuse', 'only opensuse shown';
+    $group2->update({ignore_on_dashboard => 0});
+
+    # 2. Ignore via parent group property
```
Martchus: Why the numbering? These blocks will obviously executed in sequence. Maybe subtests would make more sense than comments anyway. The same counts for the other test changes.

t/22-dashboard.t
```diff
@@ -722,4 +723,75 @@ subtest 'dashboard_build_results coverage' => sub {
...
+    @group_names = map { $_->{group}->{name} } @{$res->{results}};
+    ok !grep({ $_ eq 'Test Parent Config' } @group_names),
+      'Parent group is hidden when its ignore_on_dashboard flag is set';
+    $parent->update({ignore_on_dashboard => 0});
+
+    # Cleanup
+    $jobs->search({group_id => {in => [$g1->id, $g2->id]}})->delete;
+    $g1->delete;
+    $g2->delete;
+    $parent->delete;
```
Martchus: Just create a txn before and roll it back here.

Martchus (changes requested):

lib/OpenQA/Schema/ResultSet/Assets.pm
okurz: I thought I had written a comment pointing to https://github.com/os-autoinst/openQA/pull/7266/commits/69b6fd4067dc4eca40f6f6590cc01a04e3378eb8 to make that explicit.

t/22-dashboard.t
okurz: done

t/ui/14-dashboard.t
okurz: ok, subtests

t/ui/14-dashboard.t
okurz: done

lib/OpenQA/WebAPI/Controller/Main.pm
okurz: ok. made that a pre-computed attribute.

lib/OpenQA/BuildResults.pm
okurz: not sure what you actually prefer but I refactored this not needing the for loop anymore. Hope you like it.

Martchus (approved):

PR #7267, opened 2026-04-10
feat(Makefile): add convenience targets for all services

okurz:


aylen384 (approved):

Makefile
```diff
@@ -305,6 +305,18 @@ run-webui-test-db: setup-database ## Run a local web UI instance using a test da
...
+run-websockets-test-db: setup-database ## Run a local websockets server instance using a test database
+	$(call RUN_SERVICE_TEST_DB,script/openqa-websockets-daemon)
+
+.PHONY: run-scheduler-test-db
+run-scheduler-test-db: setup-database ## Run a local scheduler instance using a test database
+	$(call RUN_SERVICE_TEST_DB,script/openqa-scheduler-daemon)
+
+.PHONY: run-worker-test-db
+run-worker-test-db: setup-database ## Run a local worker instance using a test database
+	$(call RUN_SERVICE_TEST_DB,script/worker)
```
Martchus: The worker doesn't use the database.

Martchus (changes requested):

Martchus (approved):
Ok, looks like "test-db" also means custom basedir and test mode. That does apply to the worker. Maybe "test-db" should be "test-setup" or something but I guess it is ok.

okurz:
> Ok, looks like "test-db" also means custom basedir and test mode. That does apply to the worker. Maybe "test-db" should be "test-setup" or something but I guess it is ok.

you mean the make target name? how about `run-worker-test-env` and other targets accordingly?

Martchus:
Yes, `…-test-env` would be clearer.

Makefile
okurz: renamed the target to "…-test-env" and the make macro/command to "…TEST_ENV"

okurz:
> Yes, `…-test-env` would be clearer.

done

PR #7279, opened 2026-04-15
feat: ignore "Investigations" job group on dashboard by default

okurz:
Motivation:
Automatic investigation jobs often create many job groups that clutter the
dashboard evaluation and might lead to performance issues or 502 errors.
Hiding them by default improves the initial user experience and dashboard
load time.

Design Choices:
- Updated `ignored_job_groups` default in `lib/OpenQA/Setup.pm` to include
  "Investigations".
- Updated the example in `etc/openqa/openqa.ini` to reflect this default.

Benefits:
- Cleaner dashboard by default.
- Potentially reduced load on the main page by ignoring common but
  irrelevant job groups for general evaluation.

### DIFF
--- a/etc/openqa/openqa.ini
+++ b/etc/openqa/openqa.ini
@@ -126,7 +126,7 @@
 ## can cause jobs to incomplete with "timestamp mismatch" error messages.
 #api_hmac_time_tolerance = 300
 ## space-separated list of job groups to ignore on the dashboard
-#ignored_job_groups =
+#ignored_job_groups = Investigations
 ## How many builds to show per job group on the web UI front page
 ## Can be overridden with the limit_builds query param
 #frontpage_builds = 3
--- a/lib/OpenQA/Setup.pm
+++ b/lib/OpenQA/Setup.pm
@@ -129,7 +129,7 @@ sub default_config () {
             frontpage_builds => 3,
             frontpage_time_limit_days => 14,
             scenario_definitions_allowed_hosts => 'github.com raw.githubusercontent.com',
-            ignored_job_groups => '',
+            ignored_job_groups => 'Investigations',
         },
         rate_limits => {
             search => 5,


d3flex:
I just wonder why not at the https://openqa.opensuse.org/admin/job_templates/132?

image

okurz:
> I just wonder why not at the https://openqa.opensuse.org/admin/job_templates/132?

huh?

Martchus (approved):

etc/openqa/openqa.ini
```diff
@@ -126,7 +126,7 @@
 ## can cause jobs to incomplete with "timestamp mismatch" error messages.
 #api_hmac_time_tolerance = 300
 ## space-separated list of job groups to ignore on the dashboard
-#ignored_job_groups =
+#ignored_job_groups = Investigations
```
perlpunk, line 129: There is no group "Investigations" by default and no mechanism in openQA for such a thing. (Only in the scripts repo).
So as @d3flex suggested, instance admins can set this group to be ignored when they create it.

okurz: So are you suggesting I close the PR as there is no other content or should I change something?

perlpunk: yes, I think it's not necessary

okurz: as @Martchus approved I am not sure if we should vote or try to find consensus

kalikiana: Effectively this is an example and using such a specific example job group out of context might be unclear.

```suggestion
#ignored_job_groups = Example
```

okurz: That's wrong. We write the default value in that commented string so if the default is "Investigations" then this string should say "Investigations"

kalikiana: Well, it's very random without context. But I see no harm, so feel free to go ahead 🙂

kalikiana (approved):

d3flex:
I added "question" flag because @kalikiana approved and to give @Martchus a chance to reply on https://github.com/os-autoinst/openQA/pull/7279#discussion_r3089124210

PR #7280, opened 2026-04-15
feat: Handle deleted/altered system user more gracefully

Martchus:
* Avoid 500 responses
* Log an error message that states the problem and consequences so admins are alerted, e.g. using `openqa-logwarn`
* Do *not* just create a new system user on the fly because the existing system user might just have been altered (as it was recently in production) and we want to avoid ending up with two system user accounts

Related ticket: https://progress.opensuse.org/issues/199664

### DIFF
--- a/lib/OpenQA/Schema/Result/Jobs.pm
+++ b/lib/OpenQA/Schema/Result/Jobs.pm
@@ -2046,7 +2046,8 @@ sub handle_retry ($self) {
     return 0 unless looks_like_number $retry;
     my $ancestors = $self->ancestors;
     return 0 if $ancestors >= $retry;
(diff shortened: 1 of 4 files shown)


okurz (approved):

d3flex (approved):

perlpunk (approved):

Martchus:
~~In Mergify we have `#check-success>=50` but here I see only 37 successful checks but a few others. I'm wondering what's missing because the list of checks looks complete.~~ Looks like many OBS checks are missing.

Martchus:
Now all checks are there except the WIP check so I'm merging manually.

PR #7283, opened 2026-04-16
feat: Return to previous page after login via OAuth2

Martchus:
Related ticket: https://progress.opensuse.org/issues/199574

---

Still a draft because I haven't tested this because I don't know how I can test this.

### DIFF
--- a/lib/OpenQA/Shared/Controller/Session.pm
+++ b/lib/OpenQA/Shared/Controller/Session.pm
@@ -61,6 +61,8 @@ sub _redirect_to_referrer ($self, $ref, $res) {
     return $self->redirect_to($ref);
 }
 
+sub return_page ($self) { $self->param('return_page') || $self->req->headers->referrer }
+
 sub create ($self) {
     my $ref = $self->req->headers->referrer;
     my $config = $self->app->config;
--- a/lib/OpenQA/WebAPI/Auth/OAuth2.pm
+++ b/lib/OpenQA/WebAPI/Auth/OAuth2.pm
@@ -8,6 +8,12 @@ use Carp 'croak';
 use Mojo::Util qw(dumper);
 use OpenQA::Log qw(log_debug);
 
+# The OAuth2 provider handles login requests and responses from the auth provider both via the `/login` route.
+# So both cases are handled in `auth_login` and `auth_response` is not implemented. This is in-line with the
+# helper `get_token_p` from `Mojolicious::Plugin::OAuth2` which is also designed to do these two things at the
+# same time (see https://metacpan.org/pod/Mojolicious::Plugin::OAuth2#oauth2.get_token_p). The distinction
+# between the cases is made by whether `$data` is passed to `update_user`.
+
 sub auth_setup ($server) {
     my $app = $server->app;
     my $config = $app->config->{oauth2};
--- a/lib/OpenQA/WebAPI/Auth/OAuth2.pm
+++ b/lib/OpenQA/WebAPI/Auth/OAuth2.pm
@@ -85,18 +85,26 @@ sub update_user ($controller, $main_config, $provider_config, $data) {
         fullname => $details->{name},
         email => $details->{email});
 
-    $controller->session->{user} = $user->username;
-    $controller->redirect_to('index');
+    my $session = $controller->session;
+    $session->{user} = $user->username;
+    $controller->redirect_to(Mojo::URL->new($session->{return_page} // 'index')->path_query);
 }
+
+sub auth_logout ($controller) { delete $controller->session->{return_page} }
 
 sub auth_login ($controller) {
     croak 'Config was not parsed' unless my $main_config = $controller->app->config->{oauth2};
     croak 'Setup was not called' unless my $provider_config = $main_config->{provider_config};
 
+    my $handle_login_response_from_provider = defined $controller->param('code');
+    my $handle_login_request = !$handle_login_response_from_provider;
     my $base_url = $controller->app->config->{global}->{base_url};
     my $host = $base_url ? Mojo::URL->new($base_url)->host : $controller->req->url->host;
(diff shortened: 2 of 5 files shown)


okurz:
> Related ticket: https://progress.opensuse.org/issues/199574
> 
> Still a draft because I haven't tested this because I don't know how I can test this.

test in OSD as the code change isn't too intrusive

Martchus:
I've just tested this on OSD:

```
martchus@openqa:~> wget 'https://github.com/os-autoinst/openQA/pull/7283.patch'
…
martchus@openqa:~> sudo patch -d /usr/share/openqa -p1 -i "$PWD"/7283.patch 
patching file lib/OpenQA/Shared/Controller/Session.pm
patching file lib/OpenQA/WebAPI/Auth/OAuth2.pm
patching file lib/OpenQA/WebAPI/Auth/OpenID.pm
martchus@openqa:~> sudo systemctl restart openqa-webui.service
```

It seems to work partially. When trying to login and actually had to go though SUSEID I actually was redirected back to the specific group overview page I was on. When logging out again and trying it again I didn't need to go though SUSEID again (1) and I was only redirected to the index page (2). I'm wondering whether (1) is normal and what's missing regarding (2).

EDIT: I reverted my change and restarted again and (1) was still reproducible. ~~So to me it looks like the logout is broken but regardless of my change.~~ Judging by the network analysis tab it actually still goes through SUSEID. That's just *very* fast and doesn't require user interaction because I didn't log out from SUSEID (only from openQA). So I guess that's ok.

Martchus:
I have now come to a version that works consistently on OSD as far as I can tell. It is still deployed in case you want to try it (but of course might be reverted by an automated package update at any time).

The OpenID provider uses two endpoints: One for requesting the login and one for the provider to provide the response. This makes it easy to distinguish the two cases. We need to distinguish the cases because when requesting the login we need to save the return page and when getting the response we need to return to it.

The OAuth2 provider uses just one route. Hence I'm using `defined $controller->param('code')` to distinguish the cases here as that parameter is only present on responses from the provider. That's maybe a bit hacky but the easiest way to achieve what we want. I guess I'll add an explaining comment in the code.

Martchus:
I tried to void checking the `code` parameter relying only on the presence of the `$data` returned by `oauth2.get_token_p`. This didn't work; after `oauth2.get_token_p` it is too late to modify the session. So I reverted this PR to the version that uses the `code` parameter which previously worked on OSD. It isn't nice that we have to check the parameter as this kind of logic belongs in `oauth2.get_token_p` but I guess that is the best/simplest approach for now. I was at least able to also improve/extend unit tests.

I have also reverted all changes on OSD by re-installing packages as changes from my other experiments made it quite messy.

okurz (approved):

lib/OpenQA/Shared/Controller/Session.pm
```diff
@@ -61,6 +61,8 @@ sub _redirect_to_referrer ($self, $ref, $res) {
     return $self->redirect_to($ref);
 }
 
+sub return_page ($self) { $self->param('return_page') || $self->req->headers->referrer }
```
d3flex, line 64: this means tha the `return_page` can given as an input from a malicious user, right? I am not sure how serious vulnerability can this be but is there a way to avoid this? Actually why do you need this when there is `$self->req->headers->referrer`? if there is a case where the $self->req->headers->referrer is not present, it should just fallback to index, no?

d3flex (changes requested):
Can we descuss a little bit the security implications. See the comments but I believe this should be prevented. unless I miss something.

lib/OpenQA/Shared/Controller/Session.pm
Martchus: > this means tha the `return_page` can given as an input from a malicious user, right?

Yes, the user can control `return_page`. So they are in control of the page they will return to after login. That's not a security problem as far as I know. If it was, then the exact same security problem would already exist in the OpenID provider which contains the same code.

> I am not sure how serious vulnerability can this be but is there a way to avoid this? Actually why do you need this when there is `$self->req->headers->referrer`?

I think we need it so the `_return_back` function used in the `ensure_…` functions works. Note that both, `$self->param('return_page')` *and* `$self->req->headers->referrer` are user controlled. (`$self->req->headers` represents the headers present in the user's request.)

> if there is a case where the $self->req->headers->referrer is not present, it should just fallback to index, no?

As both are user provided we had to remove both and always return to the index page.

---

I don't think this is a security problem. The URL parameter or referrer are stored as-is and added as `Location` header as-is in the reply after the login. There's no processing of this user-supplied data on our side. I don't think one can do much bad with it. One could give someone else a link to the openQA login that will do a login and after that return to a malicious site. That's not great but hopefully easy enough to spot and still doesn't allow to hijack the user session. And again, we have this in place for years for the OpenID login.

Martchus: I now did two changes to improve this a little bit:

1. Redirect users only to the `path_query` of the URL. This is what we already do in the OpenID plugin and will prevent redirections to other pages. So that concern from my last comment is addressed.
2. De-duplicate the code between the OpenID and OAuth2 code to make it clear that we do the same thing in both cases.

Martchus:
As I did a few amendments I have now just redeployed the changes on OSD to see whether they still work. It seems like they do.

okurz (approved):

lib/OpenQA/Shared/Controller/Session.pm
d3flex: I tested manual using OAuth and passing various variables and looks good

for ref I used among others those queries
`localhost/login?return_page=https://openqa.suse.de/tests/21900638'
`localhost/login?return_page=../../'
'localhost/login?return_page=%2Ftests%2F5836090`

d3flex (approved):

PR #7284, opened 2026-04-16
fix+improve CSRF token handling when using None authentication provider

okurz:
* refactor: improve CSRF and auth logic for better maintainability
* fix: prevent "Bad CSRF token!" when using "None" authentication provider

Related issue: https://progress.opensuse.org/issues/194786

### DIFF
--- a/lib/OpenQA/Shared/Controller/Auth.pm
+++ b/lib/OpenQA/Shared/Controller/Auth.pm
@@ -56,12 +56,16 @@ sub auth ($self) {
     # Browser with a logged in user
     my ($user, $reason) = (undef, 'Not authorized');
     if ($user = $self->current_user) {
-        ($user, $reason) = (undef, 'Bad CSRF token!') unless $self->req->method eq 'GET' || $self->valid_csrf;
+        if ($self->req->method ne 'GET' && !$self->valid_csrf) {
+            # Invalidate session-based auth if CSRF is missing; fallback to API headers if present
+            $user = undef;
+            $self->render(json => {error => 'Bad CSRF token!'}, status => 403) and return 0
+              unless $self->is_api_request;
+        }
     }
 
     # No session (probably not a browser)
-    else {
-
+    if (!$user) {
         # Personal access token
         if (($self->req->headers->authorization // '') =~ /^Bearer\s+(.+)$/) {
             ($user, $reason) = $self->_token_auth($reason, $1);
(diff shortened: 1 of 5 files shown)


kalikiana (approved):

lib/OpenQA/Shared/Controller/Auth.pm
```diff
@@ -55,13 +55,20 @@ sub auth ($self) {
 
     # Browser with a logged in user
     my ($user, $reason) = (undef, 'Not authorized');
+    my $csrf_failed = 0;
     if ($user = $self->current_user) {
-        ($user, $reason) = (undef, 'Bad CSRF token!') unless $self->req->method eq 'GET' || $self->valid_csrf;
+        if ($self->req->method ne 'GET' && !$self->valid_csrf) {
+            # Invalidate session-based auth if CSRF is missing; fallback to API headers if present
+            $user = undef;
```
Martchus, line 61: How does this fallback actually work? The logic doesn't seem obvious to me. I only see that the auth via API headers (API key and personal access token) is implemented in the else block below - which we will never reach anymore at this point. So either I'm missing something or this doesn't work and the test changes below don't check for the right behavior.

lib/OpenQA/Shared/Controller/Auth.pm
```diff
@@ -55,13 +55,20 @@ sub auth ($self) {
...
+            if (!$self->is_api_request) {
+                ($user, $reason) = (undef, 'Bad CSRF token!');
+                $csrf_failed = 1;
```
Martchus: We only read this variable from code that is not reachable anymore from here - or am I missing something?

Martchus (changes requested):

lib/OpenQA/Shared/Controller/Auth.pm
kalikiana: Hum. I guess the "API Key" branch below should handle that, but it's misplaced.

okurz: I changed the code to use an early return with `render … && return`, so no need for this variable anymore

lib/OpenQA/Shared/Controller/Auth.pm
okurz: sorry, I had refactored this kaput! I dropped the else block entirely and changed it to an early return on hard CSRF failures. This now falls through to check the API headers. I also updated the test to use a valid token so we actually verify the fallback works, not only relying on an invalid API key to yield 403

okurz:
still ready for review. All review suggestions have been incorporated already 3 days ago

perlpunk (approved):

Martchus (approved):

PR #7287, opened 2026-04-17
fix: prevent "Not enough storage" with consistent setting

okurz:
os-autoinst in bb08a48 added a feature to prevent depleting available
storage. To provide consistent test results and prevent sporadic issues
in CI systems we must disable that feature in the openQA full-stack test
as well.

Related issue: https://progress.opensuse.org/issues/199061


perlpunk (approved):

kalikiana (approved):
Good catch!

PR #7288, opened 2026-04-17
perf: Select only needed columns in _previous_scenario_jobs

perlpunk:
This query can return a lot of rows, and in the case of the IssueReporter plugin really only 3 columns are used.

On master an example `get_details_ajax` takes 128ms, with this PR 68ms.

With the "cache regression_links" PR #7289  together it takes 45ms.

### DIFF
--- a/lib/OpenQA/Schema/Result/Jobs.pm
+++ b/lib/OpenQA/Schema/Result/Jobs.pm
@@ -1660,7 +1660,7 @@ sub needle_dir ($self) {
 }
 
 # return the last X complete jobs of the same scenario
-sub _previous_scenario_jobs ($self, $rows = undef) {
+sub _previous_scenario_jobs ($self, $rows = undef, $search_attrs = {}) {
     my $schema = $self->result_source->schema;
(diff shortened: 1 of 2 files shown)


okurz (approved):

perlpunk:
fixed commit message

lib/OpenQA/Schema/Result/Jobs.pm
```diff
@@ -1660,15 +1660,16 @@ sub needle_dir ($self) {
 }
 
 # return the last X complete jobs of the same scenario
-sub _previous_scenario_jobs ($self, $rows = undef) {
+sub _previous_scenario_jobs ($self, $rows = undef, $more_attrs = {}) {
```
d3flex: I would rather rename the `more_attrs`. what more attrs are for? filter particular columns. what about `search_attrs` or `select_cols`?

perlpunk: done

okurz (approved):

d3flex (approved):

PR #7290, opened 2026-04-17
feat: Throttle consistently failing test scenarios

okurz:
Motivation
Consistently failing openQA tests are usually more expensive than
passing ones because they are kept around longer, take longer to exhaust
all timeouts, run post_fail_hooks, and generate additional uploaded
logs. Such jobs can bog down an openQA instance and impact other jobs.
openQA should throttle such test scenarios by automatically reducing
their priority (i.e. increasing the priority value) based on the job
scenario's recent history.

Design Choices
A new helper method `_consecutive_failures` is introduced in
`ResultSet::Jobs` to query the recent history of completed jobs for a
given scenario (derived from `SCENARIO_WITH_MACHINE_KEYS`). If the
number of consecutive failures reaches a configurable threshold, the
priority is adjusted by a configurable step per consecutive failure.
The configuration settings `throttle_failing_job_threshold` and
`throttle_failing_job_prio_step` are added to `openqa.ini` under
`[misc_limits]` to make this behavior easily customizable.

Benefits
This prevents consistently failing test scenarios from repeatedly
overloading workers and exhausting storage space, while still stable
scenarios are preferred and are not badly impacted.

Related issue: https://progress.opensuse.org/issues/186371


lib/OpenQA/Schema/ResultSet/Jobs.pm
```diff
@@ -233,6 +248,21 @@ sub _apply_prio_throttling ($settings, $new_job_args, $group = undef) {
...
+        {'me.state' => 'done'},
+        {'me.result' => {-not_in => [OpenQA::Jobs::Constants::ABORTED_RESULTS]}},
+        map { {"me.$_" => $job_args->{$_}} } OpenQA::Schema::Result::Jobs::SCENARIO_WITH_MACHINE_KEYS
+    ];
+    my %attrs = (
+        order_by => ['me.id DESC'],
+        rows => $history_length
+    );
+    my @recent_jobs = $self->search({-and => $conds}, \%attrs)->all;
+    my ($failures) = grep { OpenQA::Jobs::Constants::is_ok_result($recent_jobs[$_]->result) } 0 .. $#recent_jobs;
```
```suggestion
    my $failures = first { OpenQA::Jobs::Constants::is_ok_result($recent_jobs[$_]->result) } 0 .. $#recent_jobs;
```

lib/OpenQA/Schema/ResultSet/Jobs.pm
```diff
@@ -233,6 +248,21 @@ sub _apply_prio_throttling ($settings, $new_job_args, $group = undef) {
     return $debug_msg;
 }
 
+sub _consecutive_failures ($self, $job_args, $history_length) {
+    my $conds = [
+        {'me.state' => 'done'},
```
Martchus: Use constants.

lib/OpenQA/Schema/ResultSet/Jobs.pm
```diff
@@ -233,6 +248,21 @@ sub _apply_prio_throttling ($settings, $new_job_args, $group = undef) {
...
+        {'me.result' => {-not_in => [OpenQA::Jobs::Constants::ABORTED_RESULTS]}},
+        map { {"me.$_" => $job_args->{$_}} } OpenQA::Schema::Result::Jobs::SCENARIO_WITH_MACHINE_KEYS
+    ];
+    my %attrs = (
+        order_by => ['me.id DESC'],
+        rows => $history_length
+    );
```
Martchus: Could be one line or would maybe even still fit into the next.

etc/openqa/openqa.ini
```diff
@@ -394,6 +394,12 @@ concurrent = 0
 ## To configure a custom set of scaling parameters, add also the below default 
 ## rule, e.g. "prio_throttling_parameters = MAX_JOB_TIME:0.007,FOO:3:2000"
 #prio_throttling_parameters = MAX_JOB_TIME:0.007
+## Number of consecutive failures before throttling kicks in.
```
Martchus:
```suggestion
## Number of consecutive failures before throttling kicks in; set to 0 to disable
## additional throttling
```

Martchus (changes requested):
I'm only requesting changes for a few code suggestions.

However, I'm still wondering whether this is a good idea. It makes applying a fix or investigation of a job after it failed (e.g. using the developer mode) or re-needling more annoying. Maybe `throttle_failing_job_threshold` shouldn't be one by default (but a little bit higher).

lib/OpenQA/Schema/ResultSet/Jobs.pm
okurz: done

lib/OpenQA/Schema/ResultSet/Jobs.pm
okurz: done

lib/OpenQA/Schema/ResultSet/Jobs.pm
okurz: done

etc/openqa/openqa.ini
okurz: done

okurz:
> I'm only requesting changes for a few code suggestions.
> 
> However, I'm still wondering whether this is a good idea. It makes applying a fix or investigation of a job after it failed (e.g. using the developer mode) or re-needling more annoying. Maybe `throttle_failing_job_threshold` shouldn't be one by default (but a little bit higher).

ok, changed to 2. That *is* a little bit higher, right? ;)

etc/openqa/openqa.ini
```diff
@@ -394,6 +394,13 @@ concurrent = 0
 ## To configure a custom set of scaling parameters, add also the below default 
 ## rule, e.g. "prio_throttling_parameters = MAX_JOB_TIME:0.007,FOO:3:2000"
 #prio_throttling_parameters = MAX_JOB_TIME:0.007
+## Number of consecutive failures before throttling kicks in; set to 0 to disable
+## additional throttling
+#throttle_failing_job_threshold = 2
```
kalikiana, line 399: I feel like we should be more forgiving here.

```suggestion
#throttle_failing_job_threshold = 10
```

kalikiana (approved):
Now that we can make it clearly visible in the web UI when it happens, I like it.

okurz:
> Now that we can make it clearly visible in the web UI when it happens, I like it.

Yes, we have the visualization of how priority was changed controlled by the `$throttling_info`, see https://github.com/os-autoinst/openQA/pull/7290/changes#diff-0cacec0ef410b23ecf692cc0795a84af469a3e96ad1b590e08541e3917053116R238

etc/openqa/openqa.ini
okurz: That's way too high. Note that this feature is only introducing throttling by changing the priority meaning that other currently scheduled jobs are more and more preferred because they did not have a chance to run because maybe a (more or less) consistently failing scenario is ignored by test owners. If there are enough free resources then the priority will actually not have any impact as all tests are immediately picked up.

lib/OpenQA/Schema/ResultSet/Jobs.pm
```diff
@@ -233,6 +248,18 @@ sub _apply_prio_throttling ($settings, $new_job_args, $group = undef) {
     return $debug_msg;
 }
 
+sub _consecutive_failures ($self, $job_args, $history_length) {
+    my $conds = [
+        {'me.state' => OpenQA::Jobs::Constants::DONE},
+        {'me.result' => {-not_in => [OpenQA::Jobs::Constants::ABORTED_RESULTS]}},
+        map { {"me.$_" => $job_args->{$_}} } OpenQA::Schema::Result::Jobs::SCENARIO_WITH_MACHINE_KEYS
+    ];
+    my @recent_jobs = $self->search({-and => $conds}, {order_by => ['me.id DESC'], rows => $history_length})->all;
```
perlpunk:
```suggestion
    my @recent_jobs = $self->search({-and => $conds}, {order_by => ['me.id DESC'], rows => $history_length, select => [qw(me.id me.result)]})->all;
```

t/api/04-jobs.t
```diff
@@ -1049,7 +1049,8 @@ subtest 'priority correctly assigned when posting job' => sub {
         $t->app->config->{misc_limits}->{prio_throttling_parameters} = 'MAX_JOB_TIME:1000';
         my $config = OpenQA::Setup::read_config($t->app);
         $config->{misc_limits}->{prio_throttling_data} = OpenQA::Setup::_load_prio_throttling($t->app, $config);
-        OpenQA::Schema::ResultSet::Jobs::_apply_prio_throttling(\%jobs_post_params, \%new_job_args);
+        OpenQA::Schema::ResultSet::Jobs::_apply_prio_throttling($schema->resultset('Jobs'),
```
perlpunk: `$schema->resultset('Jobs')` is repeated may times

lib/OpenQA/Schema/ResultSet/Jobs.pm
okurz: done

t/api/04-jobs.t
okurz: fixed

Martchus (approved):

PR #7291, opened 2026-04-17
perf: optimize compute_build_results with DB-level aggregation

okurz:
Motivation:
Fetching and iterating all job objects for huge job groups causes significant
performance degradation and memory issues. The lack of database-level
aggregation makes the system vulnerable to excessive resource consumption.

Design Choices:
1. Replaced the Perl-level job deduplication and iteration with a database-level
   group_by query (aggregation).
2. Introduced _get_latest_job_ids helper to handle scenario-based job
   deduplication entirely in the database.
3. Optimized comment data fetching by only targeting failed job IDs instead of
   all jobs in a build.
4. Preserved existing parent-group rollup behavior while leveraging the new
   aggregated data structure.

Verified manually on openqa.opensuse.org with the Tumbleweed job group
overview calling

```
runs=3 count-fail-ratio curl \
        https://openqa.opensuse.org/api/v1/job_groups/1/build_results
```

* Before: mean runtime: 1255±52.37 ms
* After: mean runtime: 832±26.79 ms

That is a relative speedup of approximately 1.5x, a 34% reduction in
execution time.

Benefits:
Significantly reduces processing time and potentially memory footprint
for builds with many scenarios. Hydrating fewer job objects results in
faster response times for group overview and dashboard pages.

Related progress issue: https://progress.opensuse.org/issues/196913

### DIFF
--- a/lib/OpenQA/BuildResults.pm
+++ b/lib/OpenQA/BuildResults.pm
@@ -9,8 +9,9 @@ use OpenQA::Jobs::Constants;
 use OpenQA::Constants qw(BUILD_SORT_BY_NAME BUILD_SORT_BY_NEWEST_JOB BUILD_SORT_BY_OLDEST_JOB);
 use OpenQA::Schema::Result::Jobs;
 use OpenQA::Utils;
-use OpenQA::Log qw(log_error);
 use Date::Format;
+use DateTime::Format::Pg;
+use List::Util qw(any);
 use Sort::Versions;
 use Time::Seconds;
 
@@ -30,43 +31,6 @@ sub init_job_figures ($job_result) {
     $job_result->{total} = 0;
 }
 
-sub count_job ($job, $jr, $labels) {
-    $jr->{total}++;
-    if ($job->state eq OpenQA::Jobs::Constants::DONE) {
-        if ($job->result eq OpenQA::Jobs::Constants::PASSED) {
-            $jr->{passed}++;
-            return;
-        }
-        if ($job->result eq OpenQA::Jobs::Constants::SOFTFAILED) {
-            $jr->{softfailed}++;
-            return;
-        }
-        if (grep { $job->result eq $_ } OpenQA::Jobs::Constants::ABORTED_RESULTS) {
-            $jr->{skipped}++;
-            return;
-        }
-        if (grep { $job->result eq $_ } OpenQA::Jobs::Constants::NOT_OK_RESULTS) {
-            my $comment_data = $labels->{$job->id};
-            $jr->{failed}++;
-            if ($comment_data) {
-                $jr->{labeled}++ if $comment_data->{reviewed};
-                $jr->{comments}++ if $comment_data->{comments} || $comment_data->{reviewed};
-            }
-            return;
-        }
-        # note: Incompletes and timeouts are accounted to both categories - failed and skipped.
-    }
-    if ($job->state eq OpenQA::Jobs::Constants::CANCELLED) {
-        $jr->{skipped}++;
-        return;
-    }
-    my $state = $job->state;
-    log_error('Encountered not-implemented state:' . $job->state . ' result:' . $job->result)
-      unless grep { /$state/ } (OpenQA::Jobs::Constants::PENDING_STATES);
-    $jr->{unfinished}++;
-    return;
-}
-
 sub add_review_badge ($build_res) {
     $build_res->{all_passed} = $build_res->{passed} + $build_res->{softfailed} >= $build_res->{total} ? 1 : 0;
     $build_res->{reviewed} = $build_res->{labeled} >= $build_res->{failed} ? 1 : 0;
@@ -74,7 +38,6 @@ sub add_review_badge ($build_res) {
 }
 
 sub find_child_groups ($group, $subgroup_filter) {
-    # handle regular (non-parent) groups
     return {
         group_ids => [$group->id],
         children => [],
@@ -91,6 +54,12 @@ sub find_child_groups ($group, $subgroup_filter) {
         group_ids => [map { $_->id } @children],
         children => \@children,
     };
+}
+
+sub _get_latest_job_ids ($jobs_resultset, $version, $buildnr, $group_ids) {
+    return map { $_->id } $jobs_resultset->search(
+        {VERSION => $version, BUILD => $buildnr, group_id => {in => $group_ids}, clone_id => undef},
+        {select => [{max => 'id'}], as => ['id'], group_by => [qw(TEST ARCH FLAVOR MACHINE)]})->all;
 }
 
 sub compute_build_results (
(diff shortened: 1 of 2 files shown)


Vogtinator:
FYI: One of the bottlenecks of the previous code was the frequent access of `t_created`. Every time that field is accessed, it gets converted from a postgres datetime to a native perl one, which is somewhat expensive. With the aggregation here this is avoided nicely.

lib/OpenQA/BuildResults.pm
```diff
@@ -197,36 +200,81 @@ sub compute_build_results (
...
+            my $count = $stat->get_column('count');
+            $jr{total} += $count;
+            $jr{distris}->{$stat->DISTRI} = 1;
+            my $t_agg = $stat->get_column('t_created_agg');
+            if ($t_agg && !ref $t_agg) {
+                $t_agg = DateTime::Format::Pg->parse_datetime($t_agg);
+            }
             if ($newest) {
-                $jr{oldest_newest} //= $job->t_created;
+                $jr{oldest_newest} = $t_agg if !$jr{oldest_newest} || $t_agg gt $jr{oldest_newest};
```
Martchus:
```suggestion
                $jr{oldest_newest} = $t_agg if !$jr{oldest_newest} || $t_agg > $jr{oldest_newest};
```

lib/OpenQA/BuildResults.pm
```diff
@@ -197,36 +200,81 @@ sub compute_build_results (
...
             }
             else {
-                $jr{oldest_newest} = $job->t_created;
+                $jr{oldest_newest} = $t_agg if !$jr{oldest_newest} || $t_agg lt $jr{oldest_newest};
```
Martchus:
```suggestion
                $jr{oldest_newest} = $t_agg if !$jr{oldest_newest} || $t_agg < $jr{oldest_newest};
```

lib/OpenQA/BuildResults.pm
```diff
@@ -197,36 +200,81 @@ sub compute_build_results (
...
+            }
+
+            my $cat = 'unfinished';
+            if ($stat->state eq OpenQA::Jobs::Constants::DONE) {
+                if ($stat->result eq OpenQA::Jobs::Constants::PASSED) { $cat = 'passed' }
+                elsif ($stat->result eq OpenQA::Jobs::Constants::SOFTFAILED) { $cat = 'softfailed' }
+                elsif (grep { $stat->result eq $_ } OpenQA::Jobs::Constants::ABORTED_RESULTS) {
```
Martchus: This should probably use `any`.

lib/OpenQA/BuildResults.pm
```diff
@@ -197,36 +200,81 @@ sub compute_build_results (
...
+                    $cat = 'skipped';
+                }
+                elsif (grep { $stat->result eq $_ } OpenQA::Jobs::Constants::NOT_OK_RESULTS) {
```
Martchus: This should probably use `any`.

Martchus (changes requested):

lib/OpenQA/BuildResults.pm
okurz: done

lib/OpenQA/BuildResults.pm
okurz: done

lib/OpenQA/BuildResults.pm
okurz: done

lib/OpenQA/BuildResults.pm
okurz: done

okurz:
Incorporated all review suggestions from @Martchus

t/34-build_results-unit.t
```diff
@@ -0,0 +1,96 @@
+#!/usr/bin/env perl
+use Test::Most;
+use FindBin;
+use lib "$FindBin::Bin/../lib", "$FindBin::Bin/lib", "$FindBin::Bin/../../external/os-autoinst-common/lib";
+use OpenQA::Test::Case;
+use OpenQA::BuildResults;
+use OpenQA::Jobs::Constants;
+use OpenQA::Constants qw(BUILD_SORT_BY_NAME BUILD_SORT_BY_NEWEST_JOB);
+use Data::Dumper;
```
d3flex: left over likely

lib/OpenQA/BuildResults.pm
```diff
@@ -30,51 +30,14 @@ sub init_job_figures ($job_result) {
...
-    }
-    my $state = $job->state;
-    log_error('Encountered not-implemented state:' . $job->state . ' result:' . $job->result)
-      unless grep { /$state/ } (OpenQA::Jobs::Constants::PENDING_STATES);
-    $jr->{unfinished}++;
-    return;
-}
-
 sub add_review_badge ($build_res) {
+
```
d3flex: nits: remove all the blank/empty lines

t/34-build_results-unit.t
okurz: yes, sorry. I had initially created a separate test file. That is now gone completely as the test addition was added to the pre-existing file t/22-dashboard.t

lib/OpenQA/BuildResults.pm
okurz: done

lib/OpenQA/BuildResults.pm
```diff
@@ -30,51 +30,14 @@ sub init_job_figures ($job_result) {
...
-            return;
-        }
-        # note: Incompletes and timeouts are accounted to both categories - failed and skipped.
-    }
-    if ($job->state eq OpenQA::Jobs::Constants::CANCELLED) {
-        $jr->{skipped}++;
-        return;
-    }
-    my $state = $job->state;
-    log_error('Encountered not-implemented state:' . $job->state . ' result:' . $job->result)
```
d3flex, line 64: This is missing. Not sure why yet. but as a side effect, it makes the import unused, as i didnt find it anwhere else in the module.

okurz: Not sure what you mean with "This is missing". I removed the function as I am using database-based aggregation instead of doing it on the Perl level. However, I removed the now unused import of log_error.

lib/OpenQA/BuildResults.pm
```diff
@@ -197,36 +160,71 @@ sub compute_build_results (
...
+            }
             if ($newest) {
-                $jr{oldest_newest} //= $job->t_created;
+                $jr{oldest_newest} = $t_agg if !$jr{oldest_newest} || $t_agg > $jr{oldest_newest};
             }
             else {
-                $jr{oldest_newest} = $job->t_created;
+                $jr{oldest_newest} = $t_agg if !$jr{oldest_newest} || $t_agg < $jr{oldest_newest};
+            }
+            my $cat = 'unfinished';
```
kalikiana, line 185: Just a nitpick, but wouldn't `$state` make more sense?


```suggestion
            my $state = 'unfinished';
```

okurz: We kept it as `$cat` because it represents the result category ('passed', 'failed', 'skipped', 'unfinished'), while `$stat->state` represents the actual job state ('DONE', 'CANCELLED'). Renaming it to `$state` would conflate the two concepts.

lib/OpenQA/BuildResults.pm
```diff
@@ -197,36 +160,71 @@ sub compute_build_results (
...
+            my $count = $stat->get_column('count');
+            $jr{total} += $count;
+            $jr{distris}->{$stat->DISTRI} = 1;
+            my $t_agg = $stat->get_column('t_created_agg');
+            if ($t_agg && !ref $t_agg) {
+                $t_agg = DateTime::Format::Pg->parse_datetime($t_agg);
+            }
             if ($newest) {
-                $jr{oldest_newest} //= $job->t_created;
+                $jr{oldest_newest} = $t_agg if !$jr{oldest_newest} || $t_agg > $jr{oldest_newest};
```
perlpunk, line 180: If you order by `t_created_agg`, you could keep the old logic `$jr{oldest_newest} //= $job->t_created;` here. Have you tried that out?

perlpunk: Somehow DBIC doesn't allow `order_by => 't_created_agg'`, I only get it working by using `order_by => 'MAX(t_created)'` (or `MIN`).
But I guess it doesn't make a big difference, the number of rows here is usually not very high.

okurz: so, are you still suggesting I change something or not?

d3flex (approved):

okurz:
@Martchus can you please review again?

Martchus (approved):

PR #7294, opened 2026-04-20
feat: disable dashboard inclusion checkbox if ignored by config

okurz:
Motivation:
Job groups can be ignored from the dashboard via .ini configuration. To
visualize that explicitly in the web UI, the "Ignore on dashboard"
checkbox should be disabled and reflect the effective state if a group
is already ignored via configuration.

Design Choices:
- Updated the group property editor template to check against the
  ignored_groups configuration.
- The disabled state is also inherited from parent groups if they are
  configured to be ignored.
- Added tooltips to explain why the checkbox is disabled.

### DIFF
--- a/templates/webapi/admin/group/group_property_editor.html.ep
+++ b/templates/webapi/admin/group/group_property_editor.html.ep
@@ -180,7 +180,7 @@
                 </div>
                 <div class="mb-3 row">
                     <label for="editor-keep-important-logs-in-days"
-                    class="form-label col-sm-2 control-label" data-bs-toggle="tooltip" title="Number of days to keep logs of important jobs" title="currently not used">Keep important logs for</label>
+                    class="form-label col-sm-2 control-label" data-bs-toggle="tooltip" title="Number of days to keep logs of important jobs">Keep important logs for</label>
                     <div class="col-sm-10">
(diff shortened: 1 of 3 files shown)


templates/webapi/admin/group/group_property_editor.html.ep
```diff
@@ -53,16 +53,26 @@
                     %>
                 </div>
             </div>
+            % my $ignored_groups = app->ignored_groups;
+            % my $ignored_by_config = $ignored_groups->{$group->name} || (!$is_parent && $group->parent && $ignored_groups->{$group->parent->name});
+            % my $ignored_title = 'Exclude from evaluation for the main dashboard page';
+            % if ($ignored_by_config) {
+            %     my $reason = $ignored_groups->{$group->name} ? 'the group name' : 'the parent group name';
+            %     $ignored_title = "Excluded from evaluation for the main dashboard page because $reason is configured in the 'ignored_job_groups' ini file setting";
+            % }
```
Martchus: It would be better to avoid having lots of code in template files (where we don't track coverage and everything is mixed with HTML). So I suggest you move this to the controller or a helper.

okurz: done

okurz:
incorporated sole review suggestion from @Martchus

kalikiana (approved):

d3flex (approved):

PR #7295, opened 2026-04-20
feat(Git): Use nickname as a fallback for fullname

perlpunk:
It can happen that we have a user without a fullname.

Issue: https://progress.opensuse.org/issues/199913


okurz (approved):
Consider to hot-patch OSD

Martchus (approved):

PR #7296, opened 2026-04-20
feat: Make assignment of fullname on login via OAuth2 configurable

Martchus:
Other assignments are already configurable. This might help to populate the fullname field after migrating to SUSEID which doesn't set `name`.

Related ticket: https://progress.opensuse.org/issues/199913

### DIFF
--- a/etc/openqa/openqa.ini
+++ b/etc/openqa/openqa.ini
@@ -200,6 +200,7 @@
 #token_scope = read_user
 #token_label = Bearer
 #id_from = id
+#fullname_from = name
 #nickname_from = username
 #
 # The default value for `id_from` is "id" for backwards compatibility.
--- a/etc/openqa/openqa.ini
+++ b/etc/openqa/openqa.ini
@@ -202,6 +202,7 @@
 #id_from = id
 #fullname_from = name
 #nickname_from = username
+#email_from = email
 #
 # The default value for `id_from` is "id" for backwards compatibility.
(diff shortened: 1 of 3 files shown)


kalikiana (approved):
Makes sense

perlpunk (commented):
What I don't understand yet is: will this now put the full name into the fullname field, or is SUSEID not providing the full name at all?
This PR just says that it's now configurable, but the value actually didn't change.

okurz (approved):
LGTM but I suggest a check on OSD, e.g. hot-patch the PR

Martchus:
> What I don't understand yet is: will this now put the full name into the fullname field

No, it will not put the full name into the fullname field for SUSEID.

> , or is SUSEID not providing the full name at all?

I don't know that. For that I'll have to add some debug logging on the production instance to see what fields we get as I've already suggested in the ticket. **If** we get something that looks like a fullname we can use this change to use it.

> This PR just says that it's now configurable, but the value actually didn't change.

Yes, because it is really just that for now.

Martchus:
> LGTM but I suggest a check on OSD, e.g. hot-patch the PR

I've hot-patch-deployed this on OSD and restarted the web UI. Logout and login still work in Firefox and Chromium. So this didn't cause a regression. As stated in the previous comment, this will allow us only to do further configuration. So there's no change in behavior to be expected.

Martchus:
Merging manually due to the WIP check (otherwise there are 49 successful checks present).

PR #7298, opened 2026-04-21
fix(apparmor): adapt snd2png path to os-autoinst#2881

okurz:
os-autoinst 607dccd in
https://github.com/os-autoinst/os-autoinst/pull/2881
changed the path of snd2png. This commit allows both the old and the new
path.

### DIFF
--- a/profiles/apparmor.d/usr.share.openqa.script.worker
+++ b/profiles/apparmor.d/usr.share.openqa.script.worker
@@ -134,7 +134,7 @@
   /usr/bin/rsync rix,
   /usr/bin/sha1sum rix,
   /usr/bin/slirpvde rix,
-  /usr/bin/snd2png rix,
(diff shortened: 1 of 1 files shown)


Vogtinator (approved):

d3flex (approved):

Martchus (approved):

rclobus:
For os-autoinst/os-autoinst#2881 you would probably also need to adjust the paths of `debugviewer`

okurz:
> For [os-autoinst/os-autoinst#2881](https://github.com/os-autoinst/os-autoinst/pull/2881) you would probably also need to adjust the paths of `debugviewer`

I don't think that "debugviewer" is ever executed from any service. And calling that as user is a different story

Martchus:
> I don't think that "debugviewer" is ever executed from any service.

And if it was it would probably fail because it needs a graphical env.

PR #7299, opened 2026-04-21
feat: automatically add configurable worker classes to jobs

okurz:
Motivation:
Provide a mechanism to automatically assign default worker classes to
jobs based on configuration in openqa.ini. This allows administrators to
ensure that jobs meet certain infrastructure requirements (e.g. resource
limits, specific hardware availability) without requiring explicit
user input for every job.

Design Choices:
- Configuration is read from [worker_class_auto_assignment] section.
- Pattern is a regex matched against the job's existing worker classes.
- If no existing class matches the pattern, the specified class is added.
- Logic is implemented in lib/OpenQA/Config.pm for parsing and
  lib/OpenQA/Schema/ResultSet/Jobs.pm for application during job creation.

Benefits:
- Decouples infrastructure requirements from test definitions.
- Allows instance-wide defaults for worker class assignment.
- Reduces manual configuration errors during job submission.

Related issue: https://progress.opensuse.org/issues/191650

### DIFF
--- a/lib/OpenQA/Config.pm
+++ b/lib/OpenQA/Config.pm
@@ -6,16 +6,45 @@ use Mojo::Base -strict, -signatures;
 
 use Config::IniFiles;
 use Exporter qw(import);
-use OpenQA::Log qw(log_info);
+use OpenQA::Log qw(log_info log_warning);
 use Mojo::File qw(path);
+use Feature::Compat::Try;
 
-our @EXPORT = qw(config_dir_within_app_home lookup_config_files parse_config_files parse_config_files_as_hash);
+our @EXPORT
+  = qw(config_dir_within_app_home lookup_config_files parse_config_files parse_config_files_as_hash parse_worker_class_auto_assignment);
 
 sub _config_dirs ($config_dir_within_home) {
     return [[$ENV{OPENQA_CONFIG} // ()], [$config_dir_within_home // ()], ['/etc/openqa', '/usr/etc/openqa']];
 }
 
 sub config_dir_within_app_home ($app) { $app->child('etc', 'openqa') }
+
+sub _parse_auto_assignment_entry ($entry) {
+    my ($pattern_str, $class) = split /\s*->\s*/, $entry, 2;
+    if (!defined $class) {
+        log_warning("Missing delimiter ' -> ' in worker_class_auto_assignment: $entry");
+        return ();
+    }
+    my $pattern_regex;
+    try { $pattern_regex = qr/$pattern_str/ }
+    catch ($e) {
+        log_warning("Invalid regex pattern in worker_class_auto_assignment: $pattern_str");
+        return ();
+    }
+    return {pattern => $pattern_regex, class => $class};
+}
+
+sub parse_worker_class_auto_assignment ($config) {
+    return $config->{_worker_class_auto_assignment_rules} if $config->{_worker_class_auto_assignment_rules};
+
+    my $section = $config->{worker_class_auto_assignment} or return [];
+    my $entries = $section->{add_worker_class_if_missing} or return [];
+    $entries = [$entries] unless ref $entries eq 'ARRAY';
+
+    my $rules = [map { _parse_auto_assignment_entry($_) } grep { $_ } @$entries];
+    $config->{_worker_class_auto_assignment_rules} = $rules;
+    return $rules;
+}
 
 sub lookup_config_files ($config_dir_within_home, $name, $silent = 0) {
     my $config_name;
@@ -32,7 +61,7 @@ sub lookup_config_files ($config_dir_within_home, $name, $silent = 0) {
             last if $has_main_config;
         }
         if (@config_file_paths) {
-            log_info "Reading $config_name config from: @config_file_paths" unless $silent;
+            log_info("Reading $config_name config from: @config_file_paths") unless $silent;
             last;
         }
     }
--- a/lib/OpenQA/Schema/ResultSet/Jobs.pm
+++ b/lib/OpenQA/Schema/ResultSet/Jobs.pm
@@ -9,6 +9,7 @@ use Date::Format 'time2str';
 use Encode qw(decode_utf8);
 use File::Basename 'basename';
 use IPC::Run;
+use OpenQA::Config;
 use OpenQA::App;
 use OpenQA::Jobs::Constants;
 use OpenQA::Constants qw(DEFAULT_MAX_JOB_TIME);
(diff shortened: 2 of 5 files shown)


lib/OpenQA/Schema/ResultSet/Jobs.pm
```diff
@@ -102,6 +103,39 @@ sub latest_jobs ($self, $until = undef) {
...
+        map {
+            my $key = $_;
+            my @values
+              = $key =~ qr/(^WORKER_CLASS|\[\])$/
+              ? grep { defined $_ && $_ ne '' }
+              split(m/,/, $settings_ref->{$key} // '')
+              : (defined $settings_ref->{$key} ? ($settings_ref->{$key}) : ());
+            map { {t_created => $now, t_updated => $now, key => $key, value => $_} } @values;
+        } keys %$settings_ref
+    ];
```
Martchus: The previous code was much more readable and might also perform better.

lib/OpenQA/Config.pm
```diff
@@ -6,17 +6,47 @@ use Mojo::Base -strict, -signatures;
...
+    catch ($e) {
+        OpenQA::Log::log_warning("Invalid regex pattern in worker_class_auto_assignment: $pattern_str");
+        return ();
+    }
+    return {pattern => $pattern_regex, class => $class};
+}
+
+sub parse_worker_class_auto_assignment ($config) {
+    return $config->{_worker_class_auto_assignment_rules}
+      if ref $config eq 'HASH' && $config->{_worker_class_auto_assignment_rules};
```
Martchus: Do we really need these `ref $config eq 'HASH'` checks in this function?

lib/OpenQA/Config.pm
```diff
@@ -6,17 +6,47 @@ use Mojo::Base -strict, -signatures;
...
+
+    my $section = $config->{worker_class_auto_assignment} // return [];
```
Martchus: This style would be a bit more conventional:
```suggestion
    my $section = $config->{worker_class_auto_assignment} or return [];
```

Martchus (changes requested):

lib/OpenQA/Config.pm
okurz: would be good to have perlcritic rules for that :)

lib/OpenQA/Config.pm
okurz: no, we don't. Removed

lib/OpenQA/Schema/ResultSet/Jobs.pm
okurz: ok. extracted helper functions to have readable single-line map calls.

okurz: Nevermind, I have changed it (back) to a for loop as you mentioned "might also perform better" considering "function call overhead in Perl" :)

lib/OpenQA/Config.pm
okurz: done

okurz:
incorporated all review suggestions from @Martchus

Martchus (approved):

lib/OpenQA/Config.pm
```diff
@@ -6,17 +6,46 @@ use Mojo::Base -strict, -signatures;
 
 use Config::IniFiles;
 use Exporter qw(import);
-use OpenQA::Log qw(log_info);
+use OpenQA::Log;
```
d3flex: isnt better with `use OpenQA::Log qw(log_info log_warning); ` and load only what actual used?
what's the logic for this approach?

kalikiana: Also the call becomes `OpenQA::Log::log_info`. Should we generally change this? Otherwise I'd say for consistently let's keep it the way we usually do 🤔

okurz: ok, done

kalikiana (approved):

d3flex (changes requested):
I see that `_apply_auto_worker_class_assignment` needs to have the $config. the function is also called inside the same block with `_apply_prio_throttling` which also make use  the config. better to extract the config beforehand and pass it to the functions.

okurz:
> I see that `_apply_auto_worker_class_assignment` needs to have the $config. the function is also called inside the same block with `_apply_prio_throttling` which also make use the config. better to extract the config beforehand and pass it to the functions.

I assume your statement is based on the assumption that retrieving the config would incur significant performance cost. I considered this, but extracting config once and passing it to both functions would be inconsistent with the rest of the codebase (see 39+ other instances where functions access OpenQA::App->singleton->config directly). The singleton pattern is designed to handle repeated access efficiently, and breaking encapsulation here would make the code less maintainable without measurable performance benefits.

d3flex:
> > I see that `_apply_auto_worker_class_assignment` needs to have the $config. the function is also called inside the same block with `_apply_prio_throttling` which also make use the config. better to extract the config beforehand and pass it to the functions.
> 
> I assume your statement is based on the assumption that retrieving the config would incur significant performance cost. I considered this, but extracting config once and passing it to both functions would be inconsistent with the rest of the codebase (see 39+ other instances where functions access OpenQA::App->singleton->config directly). The singleton pattern is designed to handle repeated access efficiently, and breaking encapsulation here would make the code less maintainable without measurable performance benefits.

No I wasnt considering performance issues. But digging in the code and seeing one function which calls two others which both need the config, it made me rethink the approach. 

>  breaking encapsulation here would make the code less maintainable without measurable performance benefits.

I do not think this is a think. you do not really gain anything from encapsulate the $config and as a matter of fact I believe keeping it like this is less maintanable. which is the main concern as opposed the performance in this case.

okurz:
> I do not think this is a think. you do not really gain anything from encapsulate the $config and as a matter of fact I believe keeping it like this is less maintanable. which is the main concern as opposed the performance in this case.

Then I suggest you create a separate, explicit PR showing all reviewers what your intention is making this consistent. Please don't block this PR on it when I am just following the existing consistent style

d3flex (approved):

okurz:
46 successful checks, merging manually, see https://github.com/os-autoinst/openQA/pull/7356

lib/OpenQA/Schema/ResultSet/Jobs.pm
```diff
@@ -171,6 +191,7 @@ sub create_from_settings ($self, $settings, $scheduled_product_id = undef) {
     return $job;
 }
 
+
```
perlpunk, line 194: ?

PR #7310, opened 2026-04-22
feat(scheduler): Consider all relevant paths in storage space check

Martchus:
Related ticket: https://progress.opensuse.org/issues/199091

### DIFF
--- a/etc/openqa/openqa.ini
+++ b/etc/openqa/openqa.ini
@@ -475,6 +475,14 @@ concurrent = 0
 ## regularly stored on; scheduling is suspended if the free storage falls under
 ## the specified percentage. Defaults to 5. Set to 0 to disable.
 #results_min_free_storage_space_percentage = 5
+## Minimum percentage of free storage in the file system where job results are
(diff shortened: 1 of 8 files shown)


okurz (commented):
+1

CI checks fail. Note that https://progress.opensuse.org/issues/199091 says "any storage", not those specifically,
> any connected storages, e.g. anything that appears in df with a configurable exclusion list

Martchus:
> CI checks fail. Note that https://progress.opensuse.org/issues/199091 says "any storage", not those specifically,

Yes, but those are the ones that make sense to cover. (The DB storage doesn't make sense. For this we needed to prevent adding new jobs at all. The root filesystem is also mainly independent from whether we run jobs. Even the archive - which I covered as it was mentioned explicitly in the ticket - is actually not really relevant here.)

okurz (approved):

kalikiana (approved):

Martchus:
All checks are present and successful except WIP so merging manually.

PR #7314, opened 2026-04-24
fix(ci): Avoid conflict when building cache with `crypto-policies-scripts-*`

Martchus:
This will hopefully fix errors when building the cache, e.g.:

```
file /usr/share/crypto-policies/BSI/opensslcnf.txt from install of
crypto-policies-20250714.cd6043a-160000.1.1.noarch conflicts with file from
package crypto-policies-20250124.4d262e7-160000.2.1.noarch
```

PR #7315, opened 2026-04-24
fix: try to resolve package conflicts with zypper

okurz:
Lately circleCI failed to install packages in our CI environment with
weird errors which are likely resolvable and avoidable if we use zypper
instead of raw rpm.

### DIFF
--- a/tools/ci/build_cache.sh
+++ b/tools/ci/build_cache.sh
@@ -8,6 +8,7 @@ set -ex
 sudo zypper ar -f -p 90 https://download.opensuse.org/repositories/devel:/openQA:/Leap:/16.0/16.0 openQA
 sudo zypper ar -f -p 95 http://download.opensuse.org/repositories/devel:openQA/16.0 devel
 tools/retry sudo zypper --gpg-auto-import-keys ref
-sudo zypper -n install --download-only $(sed -e 's/\r//' < tools/ci/ci-packages.txt)
-sudo rpm -i -f $(find /var/cache/zypp/packages/ | grep '.rpm$')
+ci_packages=$(sed -e 's/\r//' < tools/ci/ci-packages.txt)
+touch /var/cache/zypp/packages/.keep_packages
+sudo zypper -n in $ci_packages
 echo "build_cache.sh done"


Martchus (approved):

okurz:
maybe not "urgently" needed after all. https://app.circleci.com/pipelines/github/os-autoinst/openQA/20761/workflows/1c98395d-fa5f-4833-a5e1-89d0cad14268/jobs/209386 just passed after another re-push by me which does not have this "fix"

EDIT: No, other PRs like #7299 still fail "weirdly", e.g. in https://app.circleci.com/pipelines/github/os-autoinst/openQA/20761/workflows/1c98395d-fa5f-4833-a5e1-89d0cad14268/jobs/209394

tools/ci/build_cache.sh
```diff
@@ -8,6 +8,7 @@ set -ex
 sudo zypper ar -f -p 90 https://download.opensuse.org/repositories/devel:/openQA:/Leap:/16.0/16.0 openQA
 sudo zypper ar -f -p 95 http://download.opensuse.org/repositories/devel:openQA/16.0 devel
 tools/retry sudo zypper --gpg-auto-import-keys ref
-sudo zypper -n install --download-only $(sed -e 's/\r//' < tools/ci/ci-packages.txt)
-sudo rpm -i -f $(find /var/cache/zypp/packages/ | grep '.rpm$')
```
Martchus, line 12: This line should also be fixed in `build_or_install_from_cache.sh`. That is nevertheless works is strange. Maybe the problem has resolved itself regardless? (I tried fixing the other occurrence in https://github.com/os-autoinst/openQA/pull/7314 but this doesn't work yet.)

okurz: I prefer to do that in a separate PR for more obvious testing results and feedback

kalikiana (approved):

tools/ci/build_cache.sh
Martchus: https://github.com/os-autoinst/openQA/pull/7313 still shows the problem after rebasing. So the problem hasn't resolved itself and this PR seems to help - even though I'm still wondering why there's no problem in `build_or_install_from_cache.sh`.

PR #7316, opened 2026-04-24
fix(ci): Avoid conflicts when installing from cache

Martchus:
https://github.com/Martchus/openQA/commit/c5f81e366edcfa4a1fcaa525eb39c7dfa955912a replaced the separate `zypper … install --download-only … && rpm -i …`
invocations with simply using `zypper install …`. This change does the same for
the `rpm -i` call in `build_or_install_from cache.sh` to fix the same issue,
e.g.:

```
file /usr/share/crypto-policies/BSI/opensslcnf.txt from install of
crypto-policies-20250714.cd6043a-160000.1.1.noarch conflicts with file from
package crypto-policies-20250124.4d262e7-160000.2.1.noarch
```

I am not sure why https://github.com/Martchus/openQA/commit/c5f81e366edcfa4a1fcaa525eb39c7dfa955912a could be merged. Maybe because it didn't change
`ci-packages.txt`. Nightly jobs that change that file like
https://github.com/os-autoinst/openQA/pull/7326 are still failing. So with this
change I am still removing the conflicting `crypto-policies-scripts-*` from
`ci-packages.txt` to have a change there that triggers cache invalidation and
this way the changed script. The nightly job will add the correct version back
if it is still required.


okurz (approved):

kalikiana (approved):

Martchus:
Now with my last commit it doesn't find that package when building the cache. That's very annoying and I don't understand it yet. Maybe it means that the `sudo zypper ms --no-refresh --all` in the previous commit works at least after all. Maybe we also need to cache the repo dir.

PR #7317, opened 2026-04-24
chore(ci): use consistent zypper calls instead of rpm

okurz:
In c5f81e366 we changed a rpm call to zypper to fix package conflict
resolution problems. To prevent potential further problems and for
consistency we should use zypper also in the 2nd occurence.


kalikiana:
I'm guessing this is not ready for review. Please look into the CI.

Martchus:
The CI has been simplified so this can be closed.

PR #7319, opened 2026-04-24
perf: tweak jobs evaluated on job groups based on recent improvement

okurz:
379aa88b4 recently introduced performance optimizations which allow us
to bump the default limit of maximum jobs that are evaluated on job
group pages. This also addresses a concern raised within the openSUSE
community in
https://lists.opensuse.org/archives/list/factory@lists.opensuse.org/thread/ZSHLUFUHMUMGX3AUQTTQB2YOPKVR6WRZ/

Tested and verified on o3 since 2026-04-25

Related issue: https://progress.opensuse.org/issues/196913

### DIFF
--- a/etc/openqa/openqa.ini
+++ b/etc/openqa/openqa.ini
@@ -333,7 +333,7 @@ concurrent = 0
 
 ## Maximum number of jobs to include in job group overview and dashboard pages
 ## (to prevent performance issues)
-#job_groups_overview_max_jobs = 10000
+#job_groups_overview_max_jobs = 25000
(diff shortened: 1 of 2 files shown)


d3flex (approved):
Approved - I just wonder if you tested it local

Martchus:
> I just wonder if you tested it local

I'm also wondering about that.

okurz:
> > I just wonder if you tested it local
> 
> I'm also wondering about that.

Tested and verified on o3 since 2026-04-25 as well as locally

Martchus (approved):

PR #7320, opened 2026-04-24
chore(AGENTS.md): refine coverage requirements

okurz:
Meanwhile the openQA project ensures and requires full statement
coverage in the complete project so we should adjust our AGENTS
guidelines accordingly.

Related issue: https://progress.opensuse.org/issues/165393

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -9,6 +9,11 @@ JavaScript (Bootstrap).
   `eslint/prettier`).
 - `make test-checkstyle`: Must pass before committing.
 - `make test-unit-and-integration TESTS=t/your_test.t`: Run specific tests.
+- `COVERAGE=1 make test-unit-and-integration TESTS=t/your_test.t`: Run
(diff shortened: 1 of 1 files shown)


Martchus (changes requested):
Are the commits that remove comments accidentally on this branch?

okurz:
> Are the commits that remove comments accidentally on this branch?

yes, fixed. thx

kalikiana (approved):

Martchus (approved):
Ok, although that deviates from my workflow which is much simpler: Putting a `die` in the code temporarily and see whether tests fail and add an explicit check for the new behavior.

Good to document how to get the coverage report for a single file. I guess one has to be careful with that when going back-and-forth between states and reset the coverage DB as needed.

okurz:
Missing OBS check due to non-critical https://progress.opensuse.org/issues/200477 . Merging manually.

PR #7323, opened 2026-04-25
feat: include parent job group name in priority check

okurz:
Motivation:
Job group priority adjustment based on name regex failed when the
matching string was only present in the parent job group name. For
example, a "Functional" group under "Development" wouldn't match a
"Development" regex rule.

Design Choices:
Modified OpenQA::Schema::ResultSet::Jobs to prefer the 'full_<property>'
method if it exists on the job group object. Since 'full_name' is already
implemented in OpenQA::Schema::Result::JobGroups, this correctly
includes the parent name (e.g., "Development / Functional"). A 'can'
check ensures compatibility with properties that don't have a 'full_'
variant.

Benefits:
Allows more flexible and intuitive priority rules that can target
entire hierarchies of job groups by matching on parent names.

Related issue: https://progress.opensuse.org/issues/198551

### DIFF
--- a/t/api/04-jobs.t
+++ b/t/api/04-jobs.t
@@ -1129,6 +1129,16 @@ subtest 'priority correctly assigned when posting job' => sub {
         OpenQA::Schema::ResultSet::Jobs::_apply_prio_throttling($jobs, {}, \%new_job_args, $group);
         is $new_job_args{priority}, $default_prio + 15, 'priority increased by multiple matching rules';
 
+        my $parent_group = $schema->resultset('JobGroupParents')->create({name => 'Development'});
+        $group->update({parent_id => $parent_group->id});
+        %new_job_args = (priority => $default_prio);
+        $t->app->config->{misc_limits}->{prio_group_parameters} = 'full_name:Development:12';
+        $t->app->config->{misc_limits}->{prio_group_data}
+          = OpenQA::Setup::_load_prio_group_throttling($t->app, $t->app->config);
+        OpenQA::Schema::ResultSet::Jobs::_apply_prio_throttling($jobs, {}, \%new_job_args, $group);
+        is $new_job_args{priority}, $default_prio + 12, 'priority increased based on full group name';
+        $group->update({parent_id => undef});
+
         # reset for following tests
         $t->app->config->{misc_limits}->{prio_group_parameters} = '';
         $t->app->config->{misc_limits}->{prio_group_data} = undef;
--- a/etc/openqa/openqa.ini
+++ b/etc/openqa/openqa.ini
@@ -407,10 +407,10 @@ concurrent = 0
 ## Throttling based on job group properties (e.g. name, description).
 ## Format: property:regex:priority_increment[,...]
 ## By default, job groups with "Development" in the name add +50 to the priority.
-## For example, "prio_group_parameters = name:Development:50,description:^$:100"
+## For example, "prio_group_parameters = full_name:Development:50,description:^$:100"
 ## would add 50 for development groups and 100 for groups with an empty description,
 ## because ownership and purpose of a job group are unclear if the description is missing.
-#prio_group_parameters = name:Development:50
+#prio_group_parameters = full_name:Development:50
 
 [archiving]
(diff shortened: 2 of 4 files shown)


t/api/04-jobs.t
```diff
@@ -1129,6 +1129,16 @@ subtest 'priority correctly assigned when posting job' => sub {
...
         is $new_job_args{priority}, $default_prio + 15, 'priority increased by multiple matching rules';
 
+        my $parent_group = $schema->resultset('JobGroupParents')->create({name => 'Development'});
+        $group->update({parent_id => $parent_group->id});
+        %new_job_args = (priority => $default_prio);
+        $t->app->config->{misc_limits}->{prio_group_parameters} = 'name:Development:12';
+        $t->app->config->{misc_limits}->{prio_group_data}
+          = OpenQA::Setup::_load_prio_group_throttling($t->app, $t->app->config);
+        OpenQA::Schema::ResultSet::Jobs::_apply_prio_throttling($jobs, {}, \%new_job_args, $group);
+        is $new_job_args{priority}, $default_prio + 12, 'priority increased based on parent group name';
```
d3flex: should say `priority increased based on parent group settings` here, no? or something else anyway but `parent group name`

asmorodskyi (approved):
LGTM

t/api/04-jobs.t
okurz: I don't understand what you mean. This test is explicitly testing that a prio value of "12" is added because the parent group name and not the child group itself has "Development" included.

lib/OpenQA/Schema/ResultSet/Jobs.pm
```diff
@@ -216,7 +216,8 @@ sub _apply_prio_throttling ($self, $settings, $new_job_args, $group = undef) {
     if ($config && $group && (my $group_throttling = $config->{misc_limits}->{prio_group_data})) {
         for my $rule (@$group_throttling) {
             my $prop = $rule->{property};
-            my $val = $group->$prop // '';
+            my $full_prop = "full_$prop";
+            my $val = $group->can($full_prop) ? $group->$full_prop : ($group->$prop // '');
```
Martchus: Why do you use `can` here? Why not just `$group->full_name`? Group is always a regular job group, isn't it? And if it is sometimes a parent group it would be more elegant to add a `full_name` alias there.

So it would probably make most sense to simply use `full_name` where `$rule->{property}` is configured.

Martchus (changes requested):

lib/OpenQA/Schema/ResultSet/Jobs.pm
okurz: > Why do you use `can` here? Why not just `$group->full_name`? Group is always a regular job group, isn't it? And if it is sometimes a parent group it would be more elegant to add a `full_name` alias there.
> 
> So it would probably make most sense to simply use `full_name` where `$rule->{property}` is configured.

The feature works by looking at *any* job group property so I decided to check for any potential "full_$property" here. Alternatively we can of course be explicit and check here only on `full_name`. Is this what you would prefer?

Martchus: Yes, I would be explicit and check only for `full_name`. This is the only property with a `full_…` prefix and I wouldn't pretend there's a pattern when there's none. Maybe it is good enough to simply make this a configuration (and perhaps documentation) change then.

okurz: ok. changed the code to simply rely on the rule that is provided by config and changed the default config to `full_name:Development:50`. This means that the whole commit does not change the actual implementation, only the config default, the description and adding tests.

t/api/04-jobs.t
d3flex: which means that the prio is accountable of the parent group settings. Anyway it is not clear to me but lets say semantics.. I will resolve

Martchus (approved):

okurz:
All good except "WIP" which I don't know why that is still there as I disabled the github app for the org

PR #7328, opened 2026-04-27
fix(AMQP): Avoid keeping unused AMQP connections around

Martchus:
We don't keep a persistent connection with the AMQP server. That is not ideal but until we have improved that we should at least close connections when we don't use them anymore. This will help with log messages about missed heartbeats on the AMQP-server side because it reduces the number of AMQP connections we keep open. However, it does not completely fix this issue as the event loop might still be blocked for too long (so we might not even be able to keep the connections we use alive).

Related ticket: https://progress.opensuse.org/issues/200195


okurz (approved):

kalikiana (approved):

Martchus:
Merging manually as only the WIP check is missing.

Issue #7297, opened 2026-04-20
Request: Use the new 'None' authentication method in the 5-minutes podman


rclobus:
Hi,

Today I ran the podman from the https://open.qa/docs/#single-instance-container documentation.
It looks like the podman container it not using the new 'None' authentication scheme yet.
Could you update the bootstrap script?

okurz:
Yes, good idea. First we should have
* #7284 

merged though

PR #7331, opened 2026-04-27
feat: switch bootstrap to 'None' authentication provider

okurz:
Motivation:
The single-instance container setup was using the 'Fake' authentication method,
which is less ideal for modern development/testing than the newer 'None' method.
'None' provides better defaults and doesn't require manual API key injection or
explicit login.

Design Choices:
- Updated script/openqa-bootstrap to configure method = None.
- Removed manual database injection and random key generation since None handles
this automatically.
- Updated documentation to reflect these changes and promote None over Fake.

Benefits:
Simplified bootstrap process, better alignment with recommended development and
testing practices.

Related issue: https://github.com/os-autoinst/openQA/issues/7297


script/openqa-bootstrap
```diff
@@ -180,16 +180,11 @@ while ! curl -o /dev/null -w "%{http_code}" -sIL http://localhost/ | grep 200; d
     sleep 3
 done
 
-# create api key
-curl http://localhost/login # create demo user (id=2)
-API_KEY=$(hexdump -n 8 -e '2/4 "%08X" 1 "\n"' /dev/random)
-API_SECRET=$(hexdump -n 8 -e '2/4 "%08X" 1 "\n"' /dev/random)
```
Martchus, line 186: This generated a new, random key and configured it. So one ended up with a more secure setup than with with your change. The only missing step to make it secure would be switching the auth method for the web UI.

So I'm not sure whether that's a change for the better.

script/openqa-bootstrap
```diff
@@ -180,16 +180,11 @@ while ! curl -o /dev/null -w "%{http_code}" -sIL http://localhost/ | grep 200; d
...
-echo "INSERT INTO api_keys (key, secret, user_id, t_created, t_updated) VALUES ('${API_KEY}', '${API_SECRET}', 2, NOW(), NOW());" | su postgres -c "psql $dbname"
-
+# setup api key for client
 cat >> /etc/openqa/client.conf << EOF
 [localhost]
-key = ${API_KEY}
-secret = ${API_SECRET}
+key = ${OPENQA_AUTH_NONE_KEY:-DEADBEEFDEADBEEF}
+secret = ${OPENQA_AUTH_NONE_SECRET:-DEADBEEFDEADBEEF}
```
Martchus: It is a bit weird that we need this file at all with the `None` provider. Shouldn't `None` also mean that no API key/secret are required when using `openqa-cli`?

Martchus (changes requested):
The PR description mentions that this is motivated by improving the single instance container but the change itself it changing the bootstrap script. And it does it in a way that makes it less secure. If this is the work of AI, I really suggest you review these changes yourself before creating a PR.

script/openqa-bootstrap
kalikiana: I kind of agree. On the other hand changing the method away from None/Fake is necessary anyway?

okurz:
> The PR description mentions that this is motivated by improving the single instance container but the change itself it changing the bootstrap script. And it does it in a way that makes it less secure. If this is the work of AI, I really suggest you review these changes yourself before creating a PR.

No, I got completely tunnel sighted and have forgot that we use openqa-bootstrap outside the single instance container :D

okurz:
> The PR description mentions that this is motivated by improving the single instance container but the change itself it changing the bootstrap script. And it does it in a way that makes it less secure.

Actually, can you comment what you mean is making this less secure? In before we used the fake authentication means everyone is an admin. How is it less secure now?

Martchus:
> > The PR description mentions that this is motivated by improving the single instance container but the change itself it changing the bootstrap script. And it does it in a way that makes it less secure.
> 
> Actually, can you comment what you mean is making this less secure? In before we used the fake authentication means everyone is an admin. How is it less secure now?

Read https://github.com/os-autoinst/openQA/pull/7331#discussion_r3148243924. Before one was only *one* step away from a secure setup (changing the auth provider). Now one also needs to generate a new key/secret.

okurz:
> > > The PR description mentions that this is motivated by improving the single instance container but the change itself it changing the bootstrap script. And it does it in a way that makes it less secure.
> > 
> > 
> > Actually, can you comment what you mean is making this less secure? In before we used the fake authentication means everyone is an admin. How is it less secure now?
> 
> Read [#7331 (comment)](https://github.com/os-autoinst/openQA/pull/7331#discussion_r3148243924). Before one was only _one_ step away from a secure setup (changing the auth provider). Now one also needs to generate a new key/secret.

But reusing a key that was previously openly available is also a bad idea

lib/OpenQA/Shared/Controller/Auth.pm
```diff
@@ -78,6 +78,11 @@ sub auth ($self) {
         elsif (my $key = $self->req->headers->header('X-API-Key')) {
             ($user, $reason) = $self->_key_auth($reason, $key);
         }
+        # Fallback for None auth
+        elsif ($self->app->config->{auth}->{method} eq 'None') {
+            $user = $self->schema->resultset('Users')->find({username => 'admin'});
```
Martchus: Maybe this should use the `DEFAULT_ADMIN` defined in `None.pm`.

Martchus (commented):
Looks generally good with supplying no key at all.

lib/OpenQA/Shared/Controller/Auth.pm
okurz: so moving `DEFAULT_ADMIN` to Constants.pm?

okurz: done. Separate commit on top

kalikiana:
> > Read [#7331 (comment)](https://github.com/os-autoinst/openQA/pull/7331#discussion_r3148243924). Before one was only _one_ step away from a secure setup (changing the auth provider). Now one also needs to generate a new key/secret.
> 
> But reusing a key that was previously openly available is also a bad idea

This is exactly the discussion we already had [drafting **None** authentication](https://progress.opensuse.org/issues/194786). Fake is not secure. The login route can be used even after a key is present.

The **only** difference here is explicitly using a provider that consistently disabling authentication. Security by obscurity is not helping us imho (my only remark above was about documentation in that regard)

Martchus (approved):

lib/OpenQA/Shared/Controller/Auth.pm
```diff
@@ -32,12 +33,17 @@ sub check ($self) {
             if (!$exp || $exp->epoch > time) {
                 if (my $secret = $api_key->secret) {
                     my $sum = hmac_sha1_sum($self->req->url->to_string . $remote_timestamp, $secret);
-                    $user = $api_key->user;
-                    log_trace(sprintf 'API auth by user: %s, operator: %d', $user->username, $user->is_operator);
+                    $user = $api_key->user if secure_compare($hash, $sum);
```
okurz, line 38: This fixes a (minor) security invulnerability from https://github.com/os-autoinst/openQA/blob/30a97add05c726d058d25bc4dc8b3bfb750f8637/lib/OpenQA/Shared/Controller/Auth.pm#L35

okurz:
only OBS checks failed. Trying a rebase

Martchus:
Looks like the failing OBS checks are related to the changes:

```
[  362s] # Subtest: None authentication provider fallback and admin auto-login
[  362s] [info] [pid:5717] Reading openqa config from: t/data/openqa.ini
[  362s] [info] [pid:5717] setting database search path to tmp_Zb9oi6_dcNX7ryM3 when registering Minion plugin
[  362s]     1..0
[  362s] not ok 15 - No tests run for subtest "None authentication provider fallback and admin auto-login"
[  362s] Dubious, test returned 255 (wstat 65280, 0xff00)
[  362s] Failed 1/15 subtests
```

perlpunk (approved):

perlpunk:
This broke openqa-in-openqa tests: 
https://openqa.opensuse.org/tests/5941690/logfile?filename=worker-journal.log.txt#line-2707
https://progress.opensuse.org/issues/194786

PR #7336, opened 2026-04-28
fix: Fix rendering bugrefs as links in lists and other structures

Martchus:
The regex hard-codes `<p>` but bugrefs might also occur directly inside other elements like `<li>` when inside a list or `<strong>` when marked as important.

This change allows bugrefs to start after most elements so there is still a link rendered in those cases. It does not allow bugrefs to start after links because this would lead to rendering a link inside a link which would be invalid and rendered incorrectly.

Note that using variable length quantifiers in lookbehind expressions is experimental and leads to the following warning:

> Variable length positive lookbehind with capturing is experimental in regex

Using a quantifier that *can* exceed 255 characters even leads to a hard error:

> Lookbehind longer than 255 not implemented in regex

Hence this change uses separate lookbehind expressions to cover HTML element names from 1 to 6 characters. This should be sufficient to cover most uses and as stated before we want to exclude links with `href="…"` anyway.

### DIFF
--- a/lib/OpenQA/Utils.pm
+++ b/lib/OpenQA/Utils.pm
@@ -35,7 +35,7 @@ use List::Util qw(all min);
 
 my $FRAG_REGEX = FRAGMENT_REGEX;
 
-my (%BUGREFS, %BUGURLS, $MARKER_REFS, $MARKER_URLS, $BUGREF_REGEX);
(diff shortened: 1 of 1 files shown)


okurz (approved):

d3flex (approved):
tested local 👍

okurz:
Missing OBS check due to non-critical https://progress.opensuse.org/issues/200477 . Merging manually.

PR #7338, opened 2026-04-28
chore(ci): Simplify CI runtime using devel container image

Martchus:
* Remove the need to update dependencies via cron jobs that currently fail due to file conflicts
* Remove applying tidy automatically - that is something we might want to re-introduce later
* Remove the need to install packages in each pipeline (through CircleCI cache)
* Keep building/tracking a custom version of os-autoinst


container/devel:openQA:ci/base/Dockerfile
```diff
@@ -1,39 +0,0 @@
...
-# these are autoinst dependencies
-RUN zypper install -y gcc-c++ cmake ninja pkgconfig\(opencv4\) pkg-config perl\(Module::CPANfile\) pkgconfig\(fftw3\) pkgconfig\(libpng\) pkgconfig\(sndfile\) pkgconfig\(theoraenc\) tesseract-ocr && zypper clean -a
-
-# openQA dependencies
-RUN zypper install -y perl-CSS-Sass npm python3-base python3-requests python3-weasyprint git-core rsync curl 'postgresql-devel>=14' 'postgresql-server>=14' qemu qemu-tools tar xorg-x11-fonts sudo make && zypper clean -a
-
-# openQA chromedriver for Selenium tests
-RUN zypper install -y chromedriver && zypper clean -a
-
-ENV LANG=en_US.UTF-8
```
okurz, line 22: I guess from here downwards is what we still need to do somewhere to allow CI tests to run. For now I suggest you can just do all those tasks when actually executing CI tests as they don't need any remote operations and are lightweight but maybe we can also move some of those steps to the devel container build instructions themselves

Martchus:
I think there was a reason why our base container didn't use `root` (even though it meant having to use `chown` here and there):

```
make[1]: Entering directory '/root/project'
test -d /dev/shm/tpg && (pg_ctl -D /dev/shm/tpg -s status >&/dev/null || pg_ctl -D /dev/shm/tpg -s start) || ./t/test_postgresql /dev/shm/tpg
initdb: error: cannot be run as root
initdb: hint: Please log in (using, e.g., "su") as the (unprivileged) user that will own the server process.
```

container/devel:openQA:ci/base/Dockerfile
Martchus: Not really, e.g. `LANG` we also already set in the Makefile.

We apparently need another user (besides `root`) but the we can probably also simplify that a little bit.

Martchus:
Now almost everything besides ShellCheck works after adding a few `zypper` calls to install necessary dependencies.

I also created https://github.com/os-autoinst/openQA/pull/7385 to add these dependencies directly into the image.

I'm not sure about ShellCheck. It should already be part of `openQA-devel`.
EDIT: It is also installed judging by the most recent build logs on OBS:
```
[   24s] The following 1124 NEW packages are going to be installed:
[   24s]   Mesa Mesa-KHR-devel Mesa-dri Mesa-libEGL1 Mesa-libGL-devel Mesa-libGL1 Mesa-vulkan-device-select ShellCheck …
```
The most recent container hasn't been published yet, though. So maybe we'll just have to wait.
EDIT: This was due to the use of `which` which is not present in the container env. I've pushed a fix.

---

EDIT: The fullstack test fails as well. I'll look into it after lunch.
EDIT: Fixed by invalidating the fullstack cache because os-autoinst needs a rebuild against the changed container env. This revealed that the fullstack caching broke as well but I could also fix that by simplifying it.

container/devel:openQA:ci/base/Dockerfile
Martchus: Somehow `LANG` nevertheless seems to be required. So I'm adding it after all. We can investigate this later.

Martchus:
I'm not sure how long we'll have to wait for `python3-weasyprint` being available in the devel container. If it installs `openQA-devel` only from `openSUSE:Factory` it would mean we'd have to wait for the next SR and TW snapshot. However, considering the container is built under https://build.opensuse.org/package/show/devel:openQA/openQA-devel-container it might also install `openQA-devel` from `devel:openQA`.

Otherwise this is probably ready to be merged. I've also included a job to log installed RPM packages (replacing the existing `check_cache` job which would not work with the new container anymore).

okurz (approved):

Martchus:
The checkstyle check is now broken again. Not sure what I did wrong:

```
+ sudo --preserve-env -u testuser make test-checkstyle
^@^@shellcheck -x $(file --mime-type script/* t/* container/worker/*.sh tools/* | sed -n 's/^\(.*\):.*text\/x-shellscript.*$/\1/p')
yamllint --strict $((git ls-files "*.yml" "*.yaml" 2>/dev/null || find -name '*.y*ml') | grep -v ^dbicdh)
shfmt -d -i 4 -bn -ci -sr $(file --mime-type script/* t/* container/worker/*.sh tools/* | sed -n 's/^\(.*\):.*text\/x-shellscript.*$/\1/p')
The authenticity of host 'github.com (140.82.113.4)' can't be established.
ED25519 key fingerprint is: SHA256:+DiY3wvvV6TuJJhbpZisF/zLDA0zPMSvHdkr4UvCOqU
This key is not known by any other names.
```

However, it looks very wrong that it tries to make an ssh connection. It is probably `gitlint`:

```
.PHONY: test-gitlint
test-gitlint: ## Run commit message checks using gitlint
	@command -v gitlint >/dev/null 2>&1 || (echo "Command 'gitlint' not found, can not execute commit message checks. Install with 'python3-gitlint' (openSUSE) or 'pip install gitlint-core'" && false)
	@BASES=$$(for i in upstream/master upstream/main origin/master origin/main master main; do git rev-parse --verify $$i 2>/dev/null; done ||:); \
	BASE=$$(git merge-base --independent $$BASES | head -n 1); \
	gitlint --commits "$$BASE..HEAD"
```

Of course now it runs into:

```
An error occurred while executing 'git diff-tree --no-commit-id --numstat -r --root 73d9e4c71c83e0136687b4f9f18a0c609f665278': b"Warning: Permanently added 'github.com' (ED25519) to the list of known hosts.\r\ngit@github.com: Permission denied (publickey).\r\nfatal: Could not read from remote repository.\n\nPlease make sure you have the correct access rights\nand the repository exists.\nfatal: could not fetch bb9c3bc3988488c2eaf1a1cd33248ad88029b10e from promisor remote"
make: *** [Makefile:418: test-gitlint] Error 254
```

Martchus:
Not sure whether we still want this after https://github.com/os-autoinst/openQA/commit/6777c925bea431e3e5e87cc0f46821cf78cfd402 fixes the problem with cron updates (although in a bit of a hacky way). Maybe it is still worthwhile simplifying this so I've just rebased this PR. The `build-docs` check should now pass as `weasyprint` is in the container now (according to my local tests).

okurz:
> Not sure whether we still want this after https://github.com/os-autoinst/openQA/commit/6777c925bea431e3e5e87cc0f46821cf78cfd402 fixes the problem with cron updates (although in a bit of a hacky way). Maybe it is still worthwhile simplifying this

Agreed. Maybe we give it another day seeing the stability of the hacky fix.

okurz (approved):
Still approved. Remove the "not-ready" after a day or so just to see the stability of the last "hack" fix

tools/ci/prepare_dependency_pr.sh
```diff
@@ -5,20 +5,22 @@ depid=$(date -I)
...
+sha_file=tools/ci/autoinst.sha
+current_sha=$(cat "$sha_file")
+new_sha=$(curl -s https://api.github.com/repos/os-autoinst/os-autoinst/commits | grep sha | head -n 1 | grep -o -E '[0-9a-f]{40}')
+[[ $current_sha == "$new_sha" ]] && exit 0
+
+echo -n "$new_sha" > "$sha_file"
+
+sudo zypper -n install git hub curl
+hub --version
+git "$sha_file"
```
d3flex: does it miss `git add`?

d3flex (changes requested):

d3flex (approved):

PR #7340, opened 2026-04-28
feat: Add IP and hostname allow-list for assets

okurz:
Motivation:
Some hosts (e.g. openQA developer instances) need to access OSD assets
without authentication, even when require_for_assets is enabled. This
allow-list should also support hostnames and domains, with a secure
default of "localhost" to maintain ease of use for local installations.

Design Choices:
- Added a new configuration option `assets_allowlist` under the `[auth]`
  section in `openqa.ini` and set its default to "localhost" in
  `OpenQA::Setup`.
- Implemented `auth_assets` in `OpenQA::Shared::Controller::Auth` to
  bypass authentication if the remote IP or its resolved hostname
  matches an entry in the allow-list.
- Added a forward DNS lookup (`_forward_dns_match`) after reverse
  resolution to prevent DNS spoofing, ensuring the resolved IPs contain
  the original request IP.
- Refactored access checks using declarative `List::Util::any`.

Benefits:
Provides flexible, secure configuration for authorized access to assets
via IPs, hostnames, or domains without credentials. This simplifies
integration and mirroring for development purposes while maintaining
strict access control for other users. The test suite maintains zero
duplication and high maintainability.

Related issue: https://progress.opensuse.org/issues/176670

### DIFF
--- a/lib/OpenQA/Shared/Controller/Auth.pm
+++ b/lib/OpenQA/Shared/Controller/Auth.pm
@@ -9,6 +9,9 @@ use OpenQA::Log qw(log_trace);
 use Mojo::Util qw(hmac_sha1_sum secure_compare);
 use Mojo::URL;
 use Time::Seconds;
+use Socket
+  qw(AF_INET AF_INET6 inet_pton getnameinfo getaddrinfo NI_NAMEREQD NI_NUMERICHOST NI_NUMERICSERV SOCK_STREAM pack_sockaddr_in pack_sockaddr_in6);
+use List::Util 'any';
 
 sub check ($self) {
     my $config = $self->app->config;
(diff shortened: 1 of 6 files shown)


Martchus:
Looks generally good - although I'm not sure whether this can be considered fully secure with that reverse-DNS check. @nicksinger What do you think?

Have you tested this change outside of unit tests?

lib/OpenQA/Shared/Controller/Auth.pm
```diff
@@ -94,6 +97,46 @@ sub auth ($self) {
...
 }
 
+sub auth_assets ($self) {
+    my $allowlist = $self->app->config->{auth}->{assets_allowlist} // 'localhost';
+    my $remote_address = $self->remote_address;
+    my @allowed_items = split /\s+/, $allowlist;
+
+    return 1 if any { $_ eq 'localhost' ? $self->is_local_request : $remote_address eq $_ } @allowed_items;
+
+    my @host_allowed = grep { $_ ne 'localhost' && $_ !~ /^[0-9.:]+$/ } @allowed_items;
```
nicksinger, line 107: I think this should cover https://www.rfc-editor.org/rfc/rfc4007.html#section-11 as well

nicksinger (changes requested):
> whether this can be considered fully secure

No, not at all. You cannot even argue it is remotely secure. This should IMHO be implemented behind an enormous `I KNOW THAT THIS IS STUPID AND INSECURE`-switch.

IPs can be spoofed and DNS can be altered. You could lower the impact by using a fully trusted DNSSEC-chain with a fixed DNS resolver but even that would make us still susceptible to spoofed IPs which cannot be prevented on an application level.

I'm against merging this and have to ask why one cannot simple deactivate the authentication if it is not needed after all.

okurz:
> IPs can be spoofed and DNS can be altered. You could lower the impact by using a fully trusted DNSSEC-chain with a fixed DNS resolver but even that would make us still susceptible to spoofed IPs which cannot be prevented on an application level.
> 
> I'm against merging this and have to ask why one cannot simple deactivate the authentication if it is not needed after all.

Well, the initial request was https://progress.opensuse.org/issues/176670 mentioning specifically OSD where we can't "simple deactivate the authentication". So @nicksinger would you agree that the request in https://progress.opensuse.org/issues/176670 is not feasible at all? I can ask the ticket author what authentication based alternatives would be acceptable.

okurz:
rejecting as per https://github.com/os-autoinst/openQA/pull/7340#pullrequestreview-4242797989

PR #7344, opened 2026-04-29
feat(api): filter job statistics by group globs

okurz:
Motivation
Users need to filter out jobs in development job groups from the statistics
returned by the /api/v1/isos/job_stats endpoint, similar to what is currently
available for the /tests and /tests/overview endpoints.

Design Choices
We reused the existing `groups_for_globs` helper in the Iso API controller to
parse the `group_glob` and `not_group_glob` parameters. This helper provides
a list of matching job groups. We extract the group IDs and pass them to the
ScheduledProducts resultset's `job_statistics` method. The SQL query is then
dynamically constructed to filter by the target group IDs in the
`most_recent_jobs` block.

To ensure consistency with other filtered views in openQA, jobs without a
group (group_id IS NULL) are included when using only negative filters
(not_group_glob), but excluded when explicit positive filters (group_glob)
are applied.

Benefits
This enhancement gives users consistent and flexible filtering capabilities
across different endpoints, allowing them to isolate or ignore specific
environments (like "Development") and obtain more relevant job statistics.

This also helps me for qem-bot's handling of increment-approve, see
https://progress.opensuse.org/issues/198728

### DIFF
--- a/lib/OpenQA/Schema/ResultSet/ScheduledProducts.pm
+++ b/lib/OpenQA/Schema/ResultSet/ScheduledProducts.pm
@@ -53,9 +53,16 @@ sub update_note ($self, $distri, $version, $flavor, $arch, $build, $note) {
     return {updated_product_id => $sth->fetchrow_arrayref->[0]};
 }
 
-sub job_statistics ($self, $distri, $version, $flavor, $arch, $build) {
+sub job_statistics ($self, $distri, $version, $flavor, $arch, $build, $group_ids = undef, $include_null_groups = 0) {
+    my $group_filter = '';
+    my @binds;
+    if ($group_ids && @$group_ids) {
+        $group_filter = 'AND (mrj.group_id = ANY(?)' . ($include_null_groups ? ' OR mrj.group_id IS NULL' : '') . ')';
+        @binds = ($group_ids);
+    }
+
     my $sth = $self->result_source->schema->storage->dbh->prepare(
-        <<~'END_SQL'
+        <<~"END_SQL"
         WITH RECURSIVE
         -- get the initial set of jobs in the scheduled product
         initial_job_ids AS (
@@ -114,6 +121,7 @@ sub job_statistics ($self, $distri, $version, $flavor, $arch, $build) {
             JOIN jobs AS mrj ON mrj.id = latest_job_id
             WHERE
                 latest_job_id IS NOT NULL
+                $group_filter
             ORDER BY
                 job_id,
                 level DESC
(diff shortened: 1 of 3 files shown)


lib/OpenQA/Schema/ResultSet/ScheduledProducts.pm
```diff
@@ -53,9 +53,20 @@ sub update_note ($self, $distri, $version, $flavor, $arch, $build, $note) {
...
 }
 
-sub job_statistics ($self, $distri, $version, $flavor, $arch, $build) {
+sub job_statistics ($self, $distri, $version, $flavor, $arch, $build, $group_ids = undef, $include_null_groups = 0) {
+    my $group_filter = '';
+    my @binds = ($distri, $version, $flavor, $arch, $build);
+    if ($group_ids && @$group_ids) {
+        my $placeholders = join ', ', ('?') x @$group_ids;
+        $group_filter
+          = 'AND (mrj.group_id IN ('
```
Martchus: This should use `IN ANY(?)` and pass the IDs for this single static placeholder as array ref. At least I *think* this is possible with PostgreSQL. This has the great advantage of simplifying the code and then PostgreSQL always sees the same SQL code which is also more efficient. (Or two versions of the SQL code in case we still need different versions for with group IDs and for without group IDs.)

lib/OpenQA/Schema/ResultSet/ScheduledProducts.pm
```diff
@@ -132,12 +144,7 @@ sub job_statistics ($self, $distri, $version, $flavor, $arch, $build) {
             latest_job_result
         END_SQL
     );
-    $sth->bind_param(1, $distri);
-    $sth->bind_param(2, $version);
-    $sth->bind_param(3, $flavor);
-    $sth->bind_param(4, $arch);
-    $sth->bind_param(5, $build);
-    $sth->execute;
+    $sth->execute(@binds);
```
Martchus: The binds that are always present could still be in this line, e.g.:

```suggestion
    $sth->execute($distri, $version, $flavor, $arch, $build, @binds);
```

This way we would avoid copying the variables to an array just to copy them again when making the argument list.

(And with my other suggestion `@binds` would probably only contain 0 or 1 elements.)

Martchus (changes requested):

lib/OpenQA/Schema/ResultSet/ScheduledProducts.pm
okurz: done

lib/OpenQA/Schema/ResultSet/ScheduledProducts.pm
okurz: done

d3flex (approved):

Martchus (approved):

PR #7346, opened 2026-04-29
test: use isolated environment for helm chart tests

okurz:
Motivation:
Running tests directly in the source tree causes side effects like
modifying the versioned Chart.lock file and downloading dependencies
into the charts/ directory. While a "git restore" trap can undo some
of this, a truly isolated environment is more robust.

Design Choices:
- Modified the test script to create a temporary directory for each run.
- Copies the helm chart source into the temp directory and executes
  tests there.

Benefits:
Zero side effects on the source tree and better isolation between
test runs.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -446,7 +446,7 @@ test-helm-lint: ## Lint helm chart
 	tools/test_helm_chart lint
 
 .PHONY: test-helm-install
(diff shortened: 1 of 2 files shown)


Martchus:
Looks like this is already included in some other PR now.

okurz:
yes, included in #7347

PR #7349, opened 2026-04-29
style: Enforce perlcritic rule ProhibitImplicitNewlines

perlpunk:
ValuesAndExpressions::ProhibitImplicitNewlines


perlpunk:
The problem with this rule is that it complains about signatures spanning multiple lines:
```
#   ::warning file=t/lib/OpenQA/Test/Utils.pm,line=553,col=14,title=Literal line breaks in a string - severity 3::[ValuesAndExpressions::ProhibitImplicitNewlines] See pages 60,61 of PBP
[23:16:41] xt/02-perlcritic.t ........ 
```
```perl
sub test_cmd (
    $cmd, $args,
    $expected = qr//,
    $test_msg = 'command line is correct',
    $exit_code = 0,
    $exit_code_msg = 'command exits successfully'
  )
{
```
The reasons are:
* This is by default read as a prototype, not a signature
* The class `PPI::Token::Prototype` inherits from `PPI::Token::Quote::Literal` surprisingly, so the perlcritic rule also considers those as quoted strings

There is an issue about it:
* https://github.com/Perl-Critic/Perl-Critic/issues/591
It contains a suggested fix to enable recognizing subroutine signatures:
https://github.com/Perl-Critic/Perl-Critic/issues/591#issuecomment-2564771794

```
PPI_CUSTOM_FEATURE_INCLUDES='Mojo::Base: {signatures: perl}' perlcritic t/lib/OpenQA/Test/Utils.pm
```
With this it passes, but maybe this is not important enough for now.
It would also automatically pass when we switch to `use v5.40` at some point.

So I would mark this as not-ready for now, and we can exclude the policy.

PR #7353, opened 2026-04-29
style: Enforce perlcritic policy BuiltinFunctions::RequireBlockGrep

perlpunk:

https://metacpan.org/pod/Perl::Critic::Policy::BuiltinFunctions::RequireBlockGrep

Policy description:
> The expression forms of grep and map are awkward and hard to read. Use the
> block forms instead.


d3flex (commented):
looks fine but why do we need this? and can you add the ticket to the commit?

okurz (approved):

perlpunk:
> looks fine but why do we need this? and can you add the ticket to the commit?

I didn't know what you meant with "this" exactly - perlcritic or this specific rule. I now added a short sentence about perlcritic as well as the description of the policy.

d3flex (approved):

PR #7354, opened 2026-04-30
style: Enforce perlcritic policy Subroutines::RequireArgUnpacking

perlpunk:

https://metacpan.org/pod/Perl::Critic::Policy::Subroutines::RequireArgUnpacking

Policy description:
> Subroutines that use `@_` directly instead of unpacking the arguments to
> local variables first have two major problems. First, they are very hard to
> read. If you're going to refer to your variables by number instead of by
> name, you may as well be writing assembler code! Second, `@_` contains
> aliases to the original variables! If you modify the contents of a `@_`
> entry, then you are modifying the variable outside of your subroutine.


Martchus (approved):

perlpunk:
ready for review

okurz (approved):

PR #7356, opened 2026-04-30
chore(mergify): update required checks

okurz:
After the removal of the WIP service we must require less checks.


Martchus (approved):

perlpunk:
https://github.com/os-autoinst/openQA/pull/7366 was just merged automatically with 48 checks, so I don't see the need

PR #7361, opened 2026-04-30
refactor: extract config dependency from throttling functions

d3flex:
Remove internal OpenQA::App->singleton calls from _apply_auto_worker_class_assignment and _apply_prio_throttling, accepting $config as a parameter instead. This makes both functions pure with respect to global state, improving testability (callers can pass arbitrary config hashes) and eliminating hidden dependencies of the callees.

This follows discussion and implements the suggestion from https://github.com/os-autoinst/openQA/pull/7299#issuecomment-4350639121

### DIFF
--- a/lib/OpenQA/Schema/ResultSet/Jobs.pm
+++ b/lib/OpenQA/Schema/ResultSet/Jobs.pm
@@ -103,9 +103,8 @@ sub latest_jobs ($self, $until = undef) {
     return @latest;
 }
 
-sub _apply_auto_worker_class_assignment ($settings_ref) {
-    my $app = OpenQA::App->singleton or return undef;
-    my $rules = OpenQA::Config::parse_worker_class_auto_assignment($app->config // {});
+sub _apply_auto_worker_class_assignment ($settings_ref, $config = {}) {
+    my $rules = OpenQA::Config::parse_worker_class_auto_assignment($config);
     return undef unless @$rules;
 
     my @existing = grep { $_ } split qr/,/, $settings_ref->{WORKER_CLASS} // '';
@@ -172,11 +171,12 @@ sub create_from_settings ($self, $settings, $scheduled_product_id = undef) {
     }
 
     $settings{WORKER_CLASS} ||= 'qemu_' . ($new_job_args{ARCH} // 'x86_64');
-    _apply_auto_worker_class_assignment(\%settings);
+    my $config = OpenQA::App->singleton && OpenQA::App->singleton->config;
+    _apply_auto_worker_class_assignment(\%settings, $config);
 
     $new_job_args{scheduled_product_id} = $scheduled_product_id;
 
-    my $debug_msg = $self->_apply_prio_throttling(\%settings, \%new_job_args, $group);
+    my $debug_msg = $self->_apply_prio_throttling(\%settings, \%new_job_args, $group, $config);
     $settings{_PRIORITY_EXPLANATION} = $debug_msg if $debug_msg;
 
     my $job = $self->create(\%new_job_args);
@@ -212,11 +212,10 @@ sub _apply_max_job_time_prio ($factor, $time, $throt_config, $job_args) {
     return $info;
 }
 
-sub _apply_prio_throttling ($self, $settings, $new_job_args, $group = undef) {
+sub _apply_prio_throttling ($self, $settings, $new_job_args, $group = undef, $config = undef) {
     my $debug_msg;
     my $base_prio = $new_job_args->{priority} // 0;
     my @throttling_info;
-    my $config = OpenQA::App->singleton && OpenQA::App->singleton->config;
     if ($config && (my $throttling = $config->{misc_limits}->{prio_throttling_data})) {
         if (
             my $mjt_info = _apply_max_job_time_prio(
--- a/t/api/04-jobs.t
+++ b/t/api/04-jobs.t
@@ -1049,7 +1049,8 @@ subtest 'priority correctly assigned when posting job' => sub {
         $t->app->config->{misc_limits}->{prio_throttling_parameters} = 'MAX_JOB_TIME:1000';
         my $config = OpenQA::Setup::read_config($t->app);
         $config->{misc_limits}->{prio_throttling_data} = OpenQA::Setup::_load_prio_throttling($t->app, $config);
-        OpenQA::Schema::ResultSet::Jobs::_apply_prio_throttling($jobs, \%jobs_post_params, \%new_job_args);
+        OpenQA::Schema::ResultSet::Jobs::_apply_prio_throttling($jobs, \%jobs_post_params, \%new_job_args, undef,
+            $config);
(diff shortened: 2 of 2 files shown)


Martchus (commented):
The functions in `ResultSet/Jobs.pm` generally access the config - before and after your change. So I'm wondering what the point is. It doesn't really matter here at which level we access the config. Even the tests don't seem simpler with this.

d3flex:
> The functions in `ResultSet/Jobs.pm` generally access the config - before and after your change. So I'm wondering what the point is. It doesn't really matter here at which level we access the config. Even the tests don't seem simpler with this.

I can give paradigms but I am pretty sure that you can see it by your own. I want only to turn your focus on this particular part I actually refactor as follow up of  https://github.com/os-autoinst/openQA/pull/7299#issuecomment-4350639121. If the review is about the quality (code and logic), there is nothing to debate. Can you say if this commit:
- improving slightly the module in the direction that make the function a little better? what happens if you need another function which needs $config?
- is it really better to fetch a global config internally in two different private functions within another?
- Does it expose a hidden dependency in `create_from_settings`?
- is there any good reason it should be like that in the first place?

lib/OpenQA/Schema/ResultSet/Jobs.pm
```diff
@@ -261,7 +261,7 @@ sub _apply_prio_throttling ($self, $settings, $new_job_args, $group = undef, $co
     }
     if (@throttling_info) {
         my $info_str = join '; ', @throttling_info;
-        $debug_msg .= sprintf
+        $debug_msg = sprintf
```
okurz: please elaborate on this change in the git commit message.

perlpunk (commented):
Re 24015b8e82003cdf0a10681e4ecebc3a82032da6
> fix: avoid concatenation on undef in _apply_prio_throttling

> `$debug_msg` is likely not initialized when `sprintf` tries to expand to the
> variable.

That's true. It's even a fact and not just likely :)

> The warning is not visible likely due to the lack of strict
> in the module.

Sorry, but that does not make sense in several regards.

First of all, using `.=` on an uninitialized variable does not produce a warning:
```
% perl
use strict;
use warnings;
my $debug_msg;
$debug_msg .= sprintf "%s", "foo";
__END__
%
```
Second, this file has
```
```
at the top, and `Mojo::Base` automatically enables strict and warnings. I would have thought you know that.

Third, a "lack of strict" has nothing to do with warnings not appearing. You meant "a lack of warnings" instead, I assume.

Fourth, "is not visible likely due" sounds like an assumption that you did not actually check yourself. Was this suggested by an LLM?

The change itself is ok, although the use of `.=` allows to easily already append to the variable before that point in the subroutine.

Edit: and fifth, if you assume that we are not using strict in this module, I would have expected that you add it. We should use strict everywhere, right?

lib/OpenQA/Schema/ResultSet/Jobs.pm
perlpunk: my suggestion:
Along with this change, put the declaration of `$debug_msg` directly above this `if` statement.
That way it will be clear that the variable is not set / changed before.

d3flex:
> Re [24015b8](https://github.com/os-autoinst/openQA/commit/24015b8e82003cdf0a10681e4ecebc3a82032da6)
> 
> > fix: avoid concatenation on undef in _apply_prio_throttling
> 
> > `$debug_msg` is likely not initialized when `sprintf` tries to expand to the
> > variable.
> 
> That's true. It's even a fact and not just likely :)
> 
> > The warning is not visible likely due to the lack of strict
> > in the module.
> 
> Sorry, but that does not make sense in several regards.
> 
> First of all, using `.=` on an uninitialized variable does not produce a warning:
> 
> ```
> % perl
> use strict;
> use warnings;
> my $debug_msg;
> $debug_msg .= sprintf "%s", "foo";
> __END__
> %
> ```
> 
> Second, this file has
> 
> ```
> use Mojo::Base 'DBIx::Class::ResultSet', -signatures;
> ```
> 
> at the top, and `Mojo::Base` automatically enables strict and warnings. I would have thought you know that.
> 

thats a reminder. 

> Third, a "lack of strict" has nothing to do with warnings not appearing. You meant "a lack of warnings" instead, I assume.
> 
> Fourth, "is not visible likely due" sounds like an assumption that you did not actually check yourself. Was this suggested by an LLM?
> 
> The change itself is ok, although the use of `.=` allows to easily already append to the variable before that point in the subroutine.
> 
> Edit: and fifth, if you assume that we are not using strict in this module, I would have expected that you add it. We should use strict everywhere, right?

Here is the thing. I do not know. I was expecting a comment like that. I did check shortly a few things. But I think I didnt see any problem. The change might be wrong in some way. it is not AI. I decided to submit it because I was afraid that if not, it will go unoticed. it could be because it is initialized somewhere else. I changed it and seems it was working. The truth is that I didnt give too much attention though as far the tests and logs locally didnt show any problem.

d3flex:
> > Re [24015b8](https://github.com/os-autoinst/openQA/commit/24015b8e82003cdf0a10681e4ecebc3a82032da6)
> > > fix: avoid concatenation on undef in _apply_prio_throttling
> > 
> > 
> > > `$debug_msg` is likely not initialized when `sprintf` tries to expand to the
> > > variable.
> > 
> > 
> > That's true. It's even a fact and not just likely :)
> > > The warning is not visible likely due to the lack of strict
> > > in the module.
> > 
> > 
> > Sorry, but that does not make sense in several regards.
> > First of all, using `.=` on an uninitialized variable does not produce a warning:
> > ```
> > % perl
> > use strict;
> > use warnings;
> > my $debug_msg;
> > $debug_msg .= sprintf "%s", "foo";
> > __END__
> > %
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Second, this file has
> > ```
> > use Mojo::Base 'DBIx::Class::ResultSet', -signatures;
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > at the top, and `Mojo::Base` automatically enables strict and warnings. I would have thought you know that.
> 
> thats a reminder.
> 
> > Third, a "lack of strict" has nothing to do with warnings not appearing. You meant "a lack of warnings" instead, I assume.
> > Fourth, "is not visible likely due" sounds like an assumption that you did not actually check yourself. Was this suggested by an LLM?
> > The change itself is ok, although the use of `.=` allows to easily already append to the variable before that point in the subroutine.
> > Edit: and fifth, if you assume that we are not using strict in this module, I would have expected that you add it. We should use strict everywhere, right?
> 
> Here is the thing. I do not know. I was expecting a comment like that. I did check shortly a few things. But I think I didnt see any problem. The change might be wrong in some way. it is not AI. I decided to submit it because I was afraid that if not, it will go unoticed. it could be because it is initialized somewhere else. I changed it and seems it was working. The truth is that I didnt give too much attention though as far the tests and logs locally didnt show any problem.

removed from this PR

PR #7362, opened 2026-04-30
style: Enforce perlcritic policy ProhibitUnusedCapture

perlpunk:
RegularExpressions::ProhibitUnusedCapture

https://metacpan.org/pod/Perl::Critic::Policy::RegularExpressions::ProhibitUnusedCapture

Policy description:
> Perl regular expressions have multiple types of grouping syntax. The basic
> parentheses (e.g. `m/(foo)/`) captures into the magic variable `$1`.
> Non-capturing groups (e.g. `m/(?:foo)/`) are useful because they have better
> runtime performance and do not copy strings to the magic global capture
> variables.
>
> It's also easier on the maintenance programmer if you consistently use
> capturing vs. non-capturing groups, because that programmer can tell more
> easily which regexps can be refactored without breaking surrounding code
> which may use the captured values.


lib/OpenQA/Parser/Format/KTAP.pm
```diff
@@ -54,9 +54,9 @@ sub _parse_subtest ($self, $result) {
         push @{$steps->{unparsed_lines}}, $line;
         return;
     }
+    my ($status, $index, $subtest_name) = ($+{status}, $+{index}, $+{name});
```
perlpunk, line 65: perlcritic doesn't recognize the `@+{ list of names }`, so it complained that captures were unused.
Also it has to be before the next regex to be recognized (which is generally a good idea, although in this case the logic ensures that the captures will survive, as we will only reach that code if the next regex does not match).

lib/OpenQA/Schema/Result/Comments.pm
```diff
@@ -148,7 +148,7 @@ e.g. 'tag:0123:important:GM' returns a list of '0123', 'important' and 'GM'.
 
 sub tag ($self) {
     $self->text
-      =~ /\btag:(((?<version>[-.@\d\w]+)-)?(?<build>[-.@\d\w]+)|"((?<version>[-.@\d\w]+)-)?(?<build>[-.@\d\w\s\+:]+)"):(?<type>[-@\d\w]+)(:(?<description>[-.@\d\w]+))?\b/;
+      =~ /\btag:(((?<version>[-.@\d\w]+)-)?(?<build>[-.@\d\w]+)|"((?<version>[-.@\d\w]+)-)?(?<build>[-.@\d\w\s\+:]+)"):(?<type>[-@\d\w]+)(:(?<description>[-.@\d\w]+))?\b/; ## no critic (RegularExpressions::ProhibitUnusedCapture)
```
perlpunk: I think this regex should be split to several lines and have the `/x` modifier for readability.
perlcritic complains about unused captures, but it's hard to figure out.

perlpunk: And perltidy complains about the long line, and adds the comment to the next line, which makes perlcritic complain again...

Wabri:
Closed in favor of #7804

PR #7365, opened 2026-04-30
style and refactor: Enforce perlcritic policies regarding eval

perlpunk:
* BuiltinFunctions::ProhibitStringyEval
* ErrorHandling::RequireCheckingReturnValueOfEval
    
https://metacpan.org/pod/Perl::Critic::Policy::ErrorHandling::RequireCheckingReturnValueOfEval
Description:
> A common idiom in perl for dealing with possible errors is to use eval
> followed by a check of` $@`/`$EVAL_ERROR`.
>
> There's a problem with this: the value of `$EVAL_ERROR` can change between
> the end of the eval and the if statement. The issue is object destructors.
    
https://metacpan.org/pod/Perl::Critic::Policy::BuiltinFunctions::ProhibitStringyEval
Description:
> The string form of eval is recompiled every time it is executed, whereas the
> block form is only compiled once. Also, the string form doesn't give
> compile-time warnings.


t/lib/OpenQA/Test/Database.pm
```diff
@@ -58,11 +58,12 @@ sub insert_fixtures ($self, $schema, $fixtures_glob = '*.pl') {
...
     my $cwd = getcwd;
 
-    chdir $self->fixture_path;
+    my $path = $self->fixture_path;
+    chdir $path;
     my %ids;
     foreach my $fixture (glob "$fixtures_glob") {
 
-        my $info = eval path($fixture)->slurp;
+        my $info = do "$path/$fixture";
```
Martchus: This was reading a file. Is `do` doing the same? Even if it does the same that's not very intuitive.

Martchus (changes requested):

t/lib/OpenQA/Test/Database.pm
perlpunk: Yes, `do` is doing basically the same.
https://perldoc.perl.org/functions/do#do-EXPR
> Uses the value of EXPR as a filename and executes the contents of the file as a Perl script:
```
# load the exact specified file (./ and ../ special-cased)
do '/foo/stat.pl';
do './stat.pl';
do '../foo/stat.pl';

# search for the named file within @INC
do 'stat.pl';
do 'foo/stat.pl';
```

perlpunk: But somehow coverage shows that some lines are not executed anymore.
I will look into that.
More opinions on this from others?

Martchus: You could add a comment `  # eval file` at the end of the line.

perlpunk: I also tried `require`, but the same issue with the coverage appeared.
Currently I don't see how those lines could be covered at all, as our fixtures don't contain a hashref as a first element, so why are they reported as covered in current master?
Could be a bug in coverage. See #7432 .
No time to further investigate it.
I removed the change and added a no critic comment.

okurz (approved):

t/lib/OpenQA/Test/Database.pm
Martchus: Reminds me of https://github.com/os-autoinst/openQA/pull/7338 where I also had to add a commit to add coverage for one line of code. There I'm also wondering why this line was considered covered before as I couldn't find any test that would cover it. Probably also a bug in the coverage tracking.

t/api/17-mcp.t
```diff
@@ -106,7 +106,7 @@ subtest 'openqa_get_job_info tool' => sub {
     };
 
     subtest 'Input validation failed' => sub {
-        eval { $client->call_tool('openqa_get_job_info', {job_id => 'abc'}) };
+        eval { $client->call_tool('openqa_get_job_info', {job_id => 'abc'}) };    ## no critic (ErrorHandling::RequireCheckingReturnValueOfEval)
         like $@, qr/Error -32602: Invalid arguments/, 'right error message';
```
Martchus, line 110: We check `$@` here. Shouldn't this be enough?

script/openqa-load-templates
```diff
@@ -87,7 +87,7 @@ try {
     dd $info;
 }
 catch ($e) {
-    $info = eval $datafile;
+    $info = eval $datafile;    ## no critic (BuiltinFunctions::ProhibitStringyEval)
```
Martchus, line 90: We check `$@` here. Shouldn't this be enough?

perlpunk: This is about `ProhibitStringyEval`, not block eval. String eval is something that should rarely be used.
This is blindly evaluating perl code from a possibly untrusted source.
I'm wondering whether we can remove that at some point.

t/api/17-mcp.t
perlpunk: We could also leave out this rule.
In most cases we should actually use `try/catch` or in case of unit tests `throws_ok` instead.

But this policy is about some weird behaviour with `$@`, and I think there was some fix in perl regarding that, but I would have to look it up.

Martchus (approved):

PR #7366, opened 2026-04-30
style: Enforce perlcritic policy Variables::ProtectPrivateVars

perlpunk:

https://metacpan.org/pod/Perl::Critic::Policy::Variables::ProtectPrivateVars

Description:
> By convention Perl authors (like authors in many other languages) indicate
> private methods and variables by inserting a leading underscore before the
> identifier. This policy catches attempts to access private variables from
> outside the package itself.

We currently only violate that in tests, which is ok


okurz (approved):

d3flex (commented):
really wee request but it seems like it should be two commits. All the changes other than .perlcriticrc seems like disabling the policy, no?

perlpunk:
> really wee request but it seems like it should be two commits. All the changes other than .perlcriticrc seems like disabling the policy, no?

I don't think the change requires two commits.

d3flex (approved):

Martchus (commented):
If we're ok with adding further `## no critic (Variables::ProtectPrivateVars)` exceptions in test code this is ok.

PR #7367, opened 2026-04-30
fix(apparmor): allow hooks access to os-autoinst-scripts

kalikiana:
apparmor=DENIED operation=open class="file"
        profile=/usr/share/openqa/script/openqa
        name=/opt/os-autoinst-scripts/ pid=1759680
        comm=openqa-llm-inve requested_mask=r denied_mask=r
        fsuid=geekotest ouid=geekotest

See: https://progress.opensuse.org/issues/200393


okurz:
obsoleted by https://github.com/os-autoinst/openQA/pull/7382

PR #7369, opened 2026-04-30
style: Rewrite some for loops in foreach style

perlpunk:
Could be enforced with the perlcritic policy
ControlStructures::ProhibitCStyleForLoops, but it complains about too many
loops that can't be rewritten with a simple `for my $i (@array)`
or `for my $i (1 .. 23)`

https://metacpan.org/pod/Perl::Critic::Policy::ControlStructures::ProhibitCStyleForLoops

Description:
> The 3-part for loop that Perl inherits from C is butt-ugly, and only really
> necessary if you need irregular counting. The very Perlish .. operator is
> much more elegant and readable.

### DIFF
--- a/t/ui/12-needle-edit.t
+++ b/t/ui/12-needle-edit.t
@@ -555,7 +555,7 @@ subtest 'Showing new needles limited to the 5 most recent ones' => sub {
 
     # add 6 new needles (makes 7 new needles in total since one has already been added)
     my $needle_name_input = $driver->find_element_by_id('needleeditor_name');
-    for (my $i = 0; $i != 7; ++$i) {
+    for my $i (0 .. 6) {
         # enter new needle name
         my $new_needle_name = "$needlename-$i";
         $needle_name_input->clear();
--- a/lib/OpenQA/File.pm
+++ b/lib/OpenQA/File.pm
@@ -86,7 +86,7 @@ sub split ($self, $chunk_size = undef) {
     my $n_chunks = _chunk_size($self->size(), $chunk_size);
     my $files = OpenQA::Files->new();
 
-    for (my $i = 1; $i <= $n_chunks; $i++) {
+    for my $i (1 .. $n_chunks) {
         #$piece->generate_sum; # XXX: Generate sha here and ditch content?
         $files->add($self->_chunk($i, $n_chunks, $chunk_size, $residual, $total_cksum));
     }
--- a/lib/OpenQA/Parser.pm
+++ b/lib/OpenQA/Parser.pm
@@ -136,7 +136,7 @@ sub restore_tree_section ($ref) {
     try {
         walker $ref => sub ($key, $value, $keys, $parent) {
             my $hash = $ref;
-            for (my $i = 0; $i < scalar @$keys - 1; $i++) {
+            for my $i (0 .. $#$keys - 1) {
                 my ($type, $kk) = @{$keys->[$i]};
                 $hash = $hash->{$kk} if $type eq 'HASH';
                 $hash = $hash->[$kk] if $type eq 'ARRAY';
--- a/lib/OpenQA/Scheduler/WorkerSlotPicker.pm
+++ b/lib/OpenQA/Scheduler/WorkerSlotPicker.pm
@@ -42,7 +42,7 @@ sub _reduce_matching_workers ($self) {
     #       of those jobs and prioritize the jobs as usual via _allocate_worker_with_priority.
     my $picked_matching_worker_slots = $self->{_picked_matching_worker_slots};
     my $to_be_scheduled = $self->{_to_be_scheduled};
-    for (my $i = 0; $i != @$to_be_scheduled; ++$i) {
+    for my $i (0 .. $#$to_be_scheduled) {
         $to_be_scheduled->[$i]->{matching_workers} = [$picked_matching_worker_slots->[$i] // ()];
     }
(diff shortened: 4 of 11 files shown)


okurz (approved):

t/ui/12-needle-edit.t
```diff
@@ -555,7 +555,7 @@ subtest 'Showing new needles limited to the 5 most recent ones' => sub {
 
     # add 6 new needles (makes 7 new needles in total since one has already been added)
     my $needle_name_input = $driver->find_element_by_id('needleeditor_name');
-    for (my $i = 0; $i != 7; ++$i) {
+    for my $i (0 .. 6) {
```
d3flex, line 558: should this starts with 1? `(my $i = 0; $i != 7; ++$i)` indicates that the `i` will first increased and then run? it slightly change the way it was

d3flex (changes requested):
Likely there is not an impact. However some loops use `++var` rather than `var++` and the initialization havent changed. likely they do not cause any harm but I think it s worth mention.

t/28-logging.t
```diff
@@ -82,7 +82,7 @@ subtest 'Logging to stdout' => sub {
 
     like $output, qr/$$/, 'Pid is printed in debug mode';
     is @matches / 2, 3, 'Worker log matches';
-    for (my $i = 0; $i < @matches; $i += 2) {
+    for (my $i = 0; $i < @matches; $i += 2) {    ## no critic (ControlStructures::ProhibitCStyleForLoops)
```
d3flex: why this can be refactor in a way that `$i += 2` is in the for?

t/ui/12-needle-edit.t
okurz: The change presented in this PR *is* equivalent to the previous style. Are you referring about the use of the pre-increment `++$i`? That does not make a difference in this style. It would only make a difference in code like `print $i++` vs. `print ++$i`

d3flex: wouldnt be different value, in the first iteration, for each of them? 
inside `for (my $i = 0; $i != 7; ++$i)` I would expect `i=1`
inside  `for my $i (0 .. 6)` I would expect `i=0`
might not make difference though but I didnt dig to match. I saw only that it expands to a needle string or something.

perlpunk: I did check that as well and there is no difference:
```
% perl -wE'
for (my $i = 0; $i != 7; ++$i) {
  say $i
}'
0
1
2
3
4
5
6
```
```
% perl -wE'
for (my $i = 0; $i != 7; $i++) {
  say $i
}'
0
1
2
3
4
5
6
```

t/28-logging.t
perlpunk: Do you have a suggestion? I wouldn't know how to rewrite this.

t/ui/12-needle-edit.t
d3flex: interesting. but good enough. thanks for checking

t/28-logging.t
d3flex: what I had in mind initial seems doesnt work with `for my $i (0 .. #$matches)`
some gemini suggestions tho:

```bash
Option 1: List::Util::pairs (cleanest)                                                                                             
  use List::Util qw(pairs);
                                                                                                                                     
  for my $pair (pairs @matches) {                                                                                                    
      my ($got, $pattern) = @$pair;
      like $got, qr/$pattern/, "OK $got";                                                                                            
  }           
                                                                                                                                     
  Option 2: splice on a copy
  my @m = @matches;                                                                                                                  
  while (my ($got, $pattern) = splice @m, 0, 2) {
      like $got, qr/$pattern/, "OK $got";        
  }
```

Martchus (commented):
You had to use `## no critic (ControlStructures::ProhibitCStyleForLoops)` many times and in those cases the C-style for-loop also makes most sense. So I'm not sure whether this policy is something we should enforce.

perlpunk:
> You had to use `## no critic (ControlStructures::ProhibitCStyleForLoops)` many times and in those cases the C-style for-loop also makes most sense. So I'm not sure whether this policy is something we should enforce.

If we don't use that policy, I could at least rewrite those loops that make sense to rewrite.
Other opinions?

okurz:
> > You had to use `## no critic (ControlStructures::ProhibitCStyleForLoops)` many times and in those cases the C-style for-loop also makes most sense. So I'm not sure whether this policy is something we should enforce.
> 
> If we don't use that policy, I could at least rewrite those loops that make sense to rewrite.

Yes, do that first. Those should be easy to accept. And I personally would be ok to still enforce the rule but everywhere where we see it as the better approach we apply the "no critic" as that means an explicit decision was made

perlpunk:
Ok, I did that now and removed the policy from .perlcriticrc.

t/28-logging.t
perlpunk: Since perl 5.40 you can use
```
% perl
my @a = qw/ a b c d e f /;
for my ($x, $y, $z) (@a) {
    say "$x, $y, $z"
}'
__END__
a, b, c
d, e, f
```
so I'd say we wait for that

Martchus (approved):

okurz (approved):

okurz (approved):

d3flex (approved):

PR #7371, opened 2026-04-30
style: Enforce perlcritic policy Community::EmptyReturn

perlpunk:

https://metacpan.org/pod/Perl::Critic::Policy::Community::EmptyReturn

Description:
> Context-sensitive functions, while one way to write functions that DWIM (Do
> What I Mean), tend to instead lead to unexpected behavior when the function
> is accidentally used in a different context, especially if the function's
> behavior changes significantly based on context. This also can lead to
> vulnerabilities when a function is intended to be used as a scalar, but is
> used in a list, such as a hash constructor or function parameter list. return
> with no arguments will return either undef or an empty list depending on
> context. Instead, return the appropriate value explicitly.

---
Note: there are about 50 violations currently, so I only fixed two of them and wanted to ask first whether we want to enforce this or not, before I go on.


okurz (approved):

d3flex (approved):

Martchus (approved):

perlpunk:
> Note: there are about 50 violations currently, so I only fixed two of them and wanted to ask first whether we want to enforce this or not, before I go on.

Ok, I added a return value in all places. Left the not-ready label, so you can have a look again.

okurz (approved):

perlpunk:
Rebased and removed the not-ready label

PR #7377, opened 2026-05-04
perf(cleanup): replace N+1 per-group queries with single batch query

d3flex:
The status() method in ResultSet::Assets previously executed one SQL query per job group to fetch asset-group associations. With 100+ job groups, this meant 100+ queries inside a single REPEATABLE READ transaction.

Replace this N+1 pattern with a single batch query that fetches all (asset_id, group_id, max_job) tuples at once, stored in a hash and looked up per group. This reduces the query count from N+2 to 2.

The compute_max_job_by_group option is removed since the batch query always includes max_job. Previously this option controlled whether max(j.id) appeared in the per-group SELECT:

- When true (_limit caller): max_job was used for the inconsistency check ($res_max_job > $init_max_job -> die) and stored in $asset_info->{groups}->{$group_id}.
- When false (Admin/Asset.pm, second status() call): the query only returned asset_id to populate group membership without max_job.

The batch query now always returns max_job for all callers. This is harmless for the non _limit callers because the inconsistency check only fires when `fail_on_inconsistent_status` is true, which only the _limit code path sets.

### DIFF
--- a/lib/OpenQA/Schema/ResultSet/Assets.pm
+++ b/lib/OpenQA/Schema/ResultSet/Assets.pm
@@ -141,28 +141,15 @@ sub status ($self, %options) {
             END_SQL
     }
 
-    # define a query to find the latest job for each asset by group
-    my $max_job_by_group_query;
-    if ($options{compute_max_job_by_group}) {
-        $max_job_by_group_query = <<~'END_SQL';
-            select a.id as asset_id, max(j.id) as max_job
-            from jobs_assets ja
-              join jobs j on j.id=ja.job_id
-              join assets a on a.id=ja.asset_id
-              where group_id = ?
-            group by a.id;
-            END_SQL
-    }
-    else {
-        $max_job_by_group_query = <<~'END_SQL';
-            select a.id as asset_id
-            from jobs_assets ja
-              join jobs j on j.id=ja.job_id
-              join assets a on a.id=ja.asset_id
-              where group_id = ?
-            group by a.id;
-            END_SQL
-    }
+    # fetch all asset-group associations in a single batch query
+    my $max_job_by_group_query = <<~'END_SQL';
+        select a.id as asset_id, j.group_id, max(j.id) as max_job
+        from jobs_assets ja
+          join jobs j on j.id=ja.job_id
+          join assets a on a.id=ja.asset_id
+          where j.group_id IS NOT NULL
+        group by a.id, j.group_id;
+        END_SQL
     my $max_job_by_group_prepared_query = $dbh->prepare($max_job_by_group_query);
 
     # define variables; add "zero group" for groupless assets
@@ -217,6 +204,13 @@ sub status ($self, %options) {
                 push @assets, \%asset;
             }
 
+            # query all asset-group associations in one batch
+            $max_job_by_group_prepared_query->execute();
+            my %group_assets;
+            while (my $result = $max_job_by_group_prepared_query->fetchrow_hashref) {
+                $group_assets{$result->{group_id}}{$result->{asset_id}} = $result->{max_job};
+            }
+
             # query list of job groups to show assets by job group
             # note: We collect data required for /admin/assets *and* the limit_assets task.
             my $groups = $schema->resultset('JobGroups');
@@ -249,12 +243,12 @@ sub status ($self, %options) {
                     group => $group->full_name,
                 };
 
-                # add the max job ID for this group
-                $max_job_by_group_prepared_query->execute($group_id);
-                while (my $result = $max_job_by_group_prepared_query->fetchrow_hashref) {
-                    my $asset_info = $asset_info{$result->{asset_id}} or next;
+                # populate asset-group associations from the pre-fetched batch
+                my $assets_for_group = $group_assets{$group_id} // {};
+                for my $asset_id (keys %$assets_for_group) {
+                    my $asset_info = $asset_info{$asset_id} or next;
                     my $init_max_job = $asset_info->{max_job} || 0;
-                    my $res_max_job = $result->{max_job};
+                    my $res_max_job = $assets_for_group->{$asset_id};
                     $asset_info->{groups}->{$group_id} = $res_max_job;
                     $asset_info->{parents}->{$parent_id} = 1 if defined $parent_id;
 
--- a/t/37-limit_assets.t
+++ b/t/37-limit_assets.t
@@ -440,10 +440,7 @@ subtest 'asset registration considers chained and directly chained parent jobs' 
 subtest 'asset status with pending state, max_job and max_job by group' => sub {
     my $asset_status;
     stdout_like {
-        $asset_status = $schema->resultset('Assets')->status(
-            compute_pending_state_and_max_job => 1,
-            compute_max_job_by_group => 1,
-        )
+        $asset_status = $schema->resultset('Assets')->status(compute_pending_state_and_max_job => 1)
     }
     qr/Skipping asset $empty_asset_id because its name is empty/, 'warning about skipped asset';
     my ($assets_with_max_job, $assets_without_max_job) = prepare_asset_status($asset_status);
@@ -466,20 +463,13 @@ subtest 'asset status without pending state, max_job and max_job by group' => su
     for my $asset (@expected_assets_with_max_job) {
         $asset->{pending} = undef;
         $asset->{last_job} = undef;
-        my $groups = $asset->{groups};
-        for my $group_id (keys %$groups) {
-            $groups->{$group_id} = undef;
-        }
     }
     for my $asset_name (keys %expected_assets_without_max_job) {
         my $asset = $expected_assets_without_max_job{$asset_name};
         $asset->{pending} = undef;
     }
 
-    my $asset_status = $schema->resultset('Assets')->status(
-        compute_pending_state_and_max_job => 0,
-        compute_max_job_by_group => 0,
-    );
+    my $asset_status = $schema->resultset('Assets')->status(compute_pending_state_and_max_job => 0);
     my ($assets_with_max_job, $assets_without_max_job) = prepare_asset_status($asset_status);
     $assets_with_max_job->[5]->{id} = undef;    # might vary
     is_deeply $assets_with_max_job, \@expected_assets_with_max_job, 'assets with max job'
--- a/lib/OpenQA/Task/Asset/Limit.pm
+++ b/lib/OpenQA/Task/Asset/Limit.pm
@@ -64,7 +64,6 @@ sub _limit ($app, $job) {
 
         my $asset_status = $schema->resultset('Assets')->status(
             compute_pending_state_and_max_job => 1,
-            compute_max_job_by_group => 1,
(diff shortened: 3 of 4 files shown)


okurz (approved):
Looks nice. I don't know if the performance benefit is real but you reduce the LOC so why not ;)

d3flex:
> Looks nice. I don't know if the performance benefit is real but you reduce the LOC so why not ;)

the LoC shouldnt matter. I tried to measure it but I do not get realistic/complete cleanup runs. However if you run the queries and the logic, i think it is easy to make the maths in your head

lib/OpenQA/Schema/ResultSet/Assets.pm
```diff
@@ -217,6 +204,13 @@ sub status ($self, %options) {
                 push @assets, \%asset;
             }
 
+            # query all asset-group associations in one batch (replaces N per-group queries)
```
Martchus: It isn't clear what "N per-group queries" refers to after it has been replaced, so just remove that:

```suggestion
            # query all asset-group associations in one batch
```

lib/OpenQA/Schema/ResultSet/Assets.pm
```diff
@@ -217,6 +204,13 @@ sub status ($self, %options) {
...
+            $max_job_by_group_prepared_query->execute();
+            my %group_assets;
+            while (my $result = $max_job_by_group_prepared_query->fetchrow_hashref) {
+                $group_assets{$result->{group_id}}{$result->{asset_id}} = $result->{max_job};
+            }
```
Martchus, line 212: I'm wondering whether this Perl-loop can be avoided using `fetchall_hashref` like I did in https://github.com/os-autoinst/openQA/blob/b3bb564f3b396b911486bd29e3904db65535ce66/lib/OpenQA/Schema/ResultSet/ScheduledProducts.pm#L141.

t/37-limit_assets.t
```diff
@@ -440,10 +440,7 @@ subtest 'asset registration considers chained and directly chained parent jobs'
 subtest 'asset status with pending state, max_job and max_job by group' => sub {
     my $asset_status;
     stdout_like {
-        $asset_status = $schema->resultset('Assets')->status(
-            compute_pending_state_and_max_job => 1,
-            compute_max_job_by_group => 1,
-        )
+        $asset_status = $schema->resultset('Assets')->status(compute_pending_state_and_max_job => 1,)
```
Martchus:
```suggestion
        $asset_status = $schema->resultset('Assets')->status(compute_pending_state_and_max_job => 1)
```

t/37-limit_assets.t
```diff
@@ -466,20 +463,13 @@ subtest 'asset status without pending state, max_job and max_job by group' => su
...
     for my $asset_name (keys %expected_assets_without_max_job) {
         my $asset = $expected_assets_without_max_job{$asset_name};
         $asset->{pending} = undef;
     }
 
-    my $asset_status = $schema->resultset('Assets')->status(
-        compute_pending_state_and_max_job => 0,
-        compute_max_job_by_group => 0,
-    );
+    my $asset_status = $schema->resultset('Assets')->status(compute_pending_state_and_max_job => 0,);
```
Martchus:
```suggestion
    my $asset_status = $schema->resultset('Assets')->status(compute_pending_state_and_max_job => 0);
```

Martchus (changes requested):
What do you mean with `N+1` pattern and `N+2` in `This reduces the query count from N+2 to 2`? So far we do one query per job group and your change reduces this to a single query overall. I suggest you change the commit message to use simpler terms as in my previous sentence.

By the way, this is probably *not* where the cleanup spends most of its time. I guess it can't hurt to optimize this nevertheless, though.

lib/OpenQA/Schema/ResultSet/Assets.pm
d3flex: good comment (kinda i like this ones). So I can take a look but this is not something I changed. and it looks tricky in a quick look. So I hope is not a blocker for now

d3flex:
> What do you mean with `N+1` pattern and `N+2` in `This reduces the query count from N+2 to 2`? So far we do one query per job group and your change reduces this to a single query overall. I suggest you change the commit message to use simpler terms as in my previous sentence.
> 

Originally there was something like:

```bash
$groups = query 
while $group:
   query(group_id)
```
that was one query per group.
that replaced now with a single query that returns all asset-group at once

> By the way, this is probably _not_ where the cleanup spends most of its time. I guess it can't hurt to optimize this nevertheless, though.

No. it is not. but it was the simpliest easy improvement.

lib/OpenQA/Schema/ResultSet/Assets.pm
Martchus: You're introducing *new* code that might be written better following an existing example in our code base. However, it is not a blocker. Your code is still an improvement over how is was before.

Martchus:
> that was one query per group.

Yes. So just use that wording in the commit message and leave out "N + 1", "N + 2" and other more complicated expressions. I'd just call it:

> perf(cleanup): Retrieve job/assets for all groups in single query
>
> With 100+ job groups in our production instance this is a significant reduction of queries.
>
> The compute_max_job_by_group option is removed since the batch query always includes max_job. Previously this > option controlled whether max(j.id) appeared in the per-group SELECT:
>
> - When true (_limit caller): max_job was used for the inconsistency check ($res_max_job > $init_max_job -> die) and stored in $asset_info->{groups}->{$group_id}.
> - When false (Admin/Asset.pm, second status() call): the query only returned asset_id to populate group membership without max_job.
> 
> The batch query now always returns max_job for all callers. This is harmless for the non _limit callers because the inconsistency check only fires when `fail_on_inconsistent_status` is true, which only the _limit code path sets.

lib/OpenQA/Schema/ResultSet/Assets.pm
```diff
@@ -141,28 +141,15 @@ sub status ($self, %options) {
...
-            from jobs_assets ja
-              join jobs j on j.id=ja.job_id
-              join assets a on a.id=ja.asset_id
-              where group_id = ?
-            group by a.id;
-            END_SQL
-    }
+    # fetch all asset-group associations in a single batch query
+    my $max_job_by_group_query = <<~'END_SQL';
+        select a.id as asset_id, j.group_id, max(j.id) as max_job
```
perlpunk, line 146: Before the `status` method was called once with `compute_max_job_by_group` and once without, and depending on that `max(j.id)` was calculated.
Now it's calculated in both calls, although only needed in the first call.
Have you checked if that makes a measurable runtime difference?

perlpunk: on an o3 database:
```
# explain analyze select a.id as asset_id, j.group_id, max(j.id) as max_job
from jobs_assets ja
join jobs j on j.id=ja.job_id
join assets a on a.id=ja.asset_id
where j.group_id is not null
group by a.id, j.group_id;
...
Execution Time: 13249.670 ms

# explain analyze select a.id as asset_id                                  
from jobs_assets ja
join jobs j on j.id=ja.job_id
join assets a on a.id=ja.asset_id
where group_id is not null
group by a.id;

Execution Time: 5021.392 ms
```
on an osd database:
```
# with max(j.id)
Execution Time: 54618.296 ms

# without
Execution Time: 20362.996 ms
```
That doesn't look like an improvement.
And it's not related to your actual change you described in the PR.
If I'm missing  something, let me know.
The change is probably ok if you leave out this part and still keep two different queries.

perlpunk (changes requested):

lib/OpenQA/Schema/ResultSet/Assets.pm
perlpunk: ah sorry, I left out the group_id in the old query for comparison.
If I use it in both, the additional `max(j.id)` doesn't seem to make a difference.

perlpunk: Of course, fetching the additional column in the perl code when it is not needed probably still makes a small difference.
I wonder if the calculation originally was left out because of performance or for other reasons.

d3flex: Exactly what you did + one extra query which have been removed now

```bash
openqa=# explain analyze select a.id as asset_id, j.group_id, max(j.id) as max_job
openqa-# from jobs_assets ja
openqa-# join jobs j on j.id=ja.job_id
openqa-# join assets a on a.id=ja.asset_id
openqa-# where j.group_id is not null
openqa-# group by a.id, j.group_id;

                                                                   QUERY PLAN                                                        
           
-------------------------------------------------------------------------------------------------------------------------------------
-----------
 HashAggregate  (cost=208670.78..234098.55 rows=1170775 width=24) (actual time=1736.262..1748.531 rows=54279.00 loops=1)

 Planning:
   Buffers: shared hit=324 read=67
 Planning Time: 5.537 ms
 Execution Time: 1754.846 ms
(28 rows)

openqa=# 
openqa=# explain analyze select a.id as asset_id                                  
openqa-# from jobs_assets ja
openqa-# join jobs j on j.id=ja.job_id
openqa-# join assets a on a.id=ja.asset_id
openqa-# where group_id is not null
openqa-# group by a.id;
                                                                            QUERY PLAN                                               
                             
-------------------------------------------------------------------------------------------------------------------------------------
-----------------------------
 Finalize HashAggregate  (cost=100758.44..101148.73 rows=39029 width=8) (actual time=589.951..603.531 rows=39060.00 loops=1)

 Planning:
   Buffers: shared hit=39
 Planning Time: 0.790 ms
 Execution Time: 605.915 ms
(38 rows)

openqa=# explain analyze select a.id as asset_id, j.group_id, max(j.id) as max_job
openqa-#         from jobs_assets ja
openqa-#           join jobs j on j.id=ja.job_id
openqa-#           join assets a on a.id=ja.asset_id
openqa-#           where j.group_id IS NOT NULL
openqa-#         group by a.id, j.group_id;

                                                                   QUERY PLAN                                                        
           
-------------------------------------------------------------------------------------------------------------------------------------
-----------
 HashAggregate  (cost=208670.78..234098.55 rows=1170775 width=24) (actual time=1618.177..1626.808 rows=54279.00 loops=1)

 Planning:
   Buffers: shared hit=17 read=22
 Planning Time: 2.224 ms
 Execution Time: 1629.648 ms
(28 rows)
```

Yes the aggregate query takes *1754.846 ms* in one call
and yes the other query takes *605.915 ms* but in additional it loops over each group_id and for a single query gives *1629.648 ms* total *2235.563 ms* only for one group

add this is the a big win 1754.846 outperform as you do not need to do the loop anymore

d3flex:
updated the commit

Martchus (approved):
I still see that the commit message mentions "N+1" pattern even though the "+1" part is likely not the case and we could just avoid the formula completely. I'm still approving because at least the wrong "N+2" phrase is removed and at some point this gets too tiresome.

perlpunk (approved):

PR #7378, opened 2026-05-04
Include log in IPA results

alvarocarvajald:
Current implementation of `OpenQA::Parser::Format::IPA` shows very limited information in OpenQA results, both in the `Details` tab and in the `External Results` tab. It causes it to render in both tabs, only 2 lines:

* A first line with the test name
* A second line with the result

However, the parser also supports a log to be included in the external JSON results file which can be uploaded, but not rendered anywhere.

This commit adds to the results' content the log if it is defined, so that is rendered both in `Details` (the first few lines) and in `External Results` tabs (full log).

* Related Ticket: https://jira.suse.com/browse/TEAM-11107
* Run without changes: http://mango.qe.nue2.suse.org/tests/7719
* Run **with** changes: http://mango.qe.nue2.suse.org/tests/7730

Before changes, IPA results will be shown in the `Details` tab like this:

Screenshot from 2026-05-04 11-53-28

Similarly, in the `External Results` tab information is limited to 2 lines:

Screenshot from 2026-05-04 11-54-58

With the changes, results would include the first few lines of the log in the `Details` tab:
Screenshot from 2026-05-04 11-53-53

And the full log is shown in the `External Results` tab:

Screenshot from 2026-05-04 11-54-22

### DIFF
--- a/lib/OpenQA/Parser/Format/IPA.pm
+++ b/lib/OpenQA/Parser/Format/IPA.pm
@@ -60,6 +60,7 @@ sub parse ($self, $json) {
         my $details = {result => $result->{result}};
         my $text_fn = "IPA-$t_name.txt";
         my $content = join "\n", $t_name, $result->{result};
+        $content .= "\nlog:\t$res->{test}->{log}" if (defined $res->{test}->{log});
 
         $details->{text} = $text_fn;
         $details->{title} = $t_name;


okurz (approved):

mloviska (approved):
Me gusta, gracias

grisu48 (approved):

Martchus (approved):

PR #7385, opened 2026-05-05
ci: Ensure all required dependencies are part of the devel container

Martchus:
* Move `pandoc` from style requires to devel requires as it is used for
  generating documentation (and not just for style checks)
* Add `python3-weasyprint` as it is required for generating PDF documentation
  and the script to generate documentation fails with a hard error if it is
  missing
* Add `perl(TAP::Harness::JUnit)` as we use it in the CI environment to generate
  JUnit-compatible test results
    * Add it only to the devel container (and not `openQA-devel`) as it is not
      strictly required for normal development and might not be available on all
      targets we build `openQA-devel` for.
    * It is still small enough to add it to the devel container without causing
      too much bloat. This way we can use the devel container in the CI env, not
      needing a separate container.

### DIFF
--- a/.mergify.yml
+++ b/.mergify.yml
@@ -13,7 +13,7 @@ pull_request_rules:
           # "unresolvable" is not reported in general:
           # https://trello.com/c/0N3jHq5M/2257-report-back-to-scm-when-build-results-arent-failed-and-succeeded
           # So we need to require the number of tests explicitly:
-          - "#check-success>=47"
+          - "#check-success>=45"
           - status-success=static-check-containers
           - status-success=webui-docker-compose
           - "#check-failure=0"
--- a/container/devel/Dockerfile
+++ b/container/devel/Dockerfile
@@ -20,7 +20,7 @@ LABEL org.opencontainers.image.created="%BUILDTIME%"
 # endlabelprefix
 
 # hadolint ignore=DL3037
-RUN zypper in -y os-autoinst-devel openQA-devel vim nodejs-default git mc perl-Devel-REPL perl-Term-ReadKey && \
+RUN zypper in -y os-autoinst-devel openQA-devel vim nodejs-default git mc perl-Devel-REPL perl-Term-ReadKey perl-TAP-Harness-JUnit && \
     zypper clean -a
 
 ENV OPENQA_DIR=/opt/openqa
--- a/cpanfile
+++ b/cpanfile
@@ -121,6 +121,7 @@ on 'develop' => sub {
     requires 'Perl::Critic::Community';
     requires 'Perl::Tidy', '== 20260204.0.0';
     requires 'Pod::Markdown';
+    requires 'TAP::Harness::JUnit';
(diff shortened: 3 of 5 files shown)


okurz (changes requested):
please provide more details in the commit message. I don't understand your choices yet

Martchus:
I provided further details. I'm also wondering whether we should add `hub` at least to the `Dockerfile` as it is installed manually via zypper so far to create documentation PRs. (It is also manually installed to create dependency PRs but we won't need those anymore anyway if https://github.com/os-autoinst/openQA/pull/7338 has been merged.)

Oh, and the sentence

> This way we can use the devel container in the CI env, not needing a separate container.

which I've added of course also refers to https://github.com/os-autoinst/openQA/pull/7338 but I didn't want to add a reference to an unmerged/draft PR into the commit message.

okurz (approved):

d3flex (approved):

okurz:
openQA:openQA-devel-test lists python3-weasyprint as unresolvable in 15.6 . I recommend you exclude those dependencies for 15.6 or disable the build of the "devel" environment for 15.6 or linkpac that package (and dependencies) to https://build.opensuse.org/project/show/devel:openQA:Leap:15.6

Martchus:
I've already added 18 packages to the devel project but now I'm running into a build error with `python-ruff`. So this is probably too much effort and I'll remove those packages again.

I'm not sure how I can simply exclude those dependencies for Leap 15.6 while still using our `dependency.yaml` structure as-is. So I guess I'll just disable building the devel package for Leap 15.6 then.

Martchus:
Looks like disabling the devel package for Leap 15.6 worked, it is already gone in the backports build were it only occurs as removed package in the build logs:

```
[  538s] --- /tmp/tmp.dMFi92Q56J	2026-05-05 14:10:07.708000000 +0000
[  538s] +++ /tmp/tmp.Y5hSzGHAyf	2026-05-05 14:10:07.732000000 +0000
[  538s] @@ -5,7 +5,6 @@
[  538s]  openQA-client
[  538s]  openQA-common
[  538s]  openQA-continuous-update
[  538s] -openQA-devel
[  538s]  openQA-doc
[  538s]  openQA-local-db
[  538s]  openQA-mcp
```

The normal build for Leap 15.6 is unfortunately still blocked.
EDIT: The build has now finished and the openQA-devel package is also removed. So my spec file change works.

I guess I'll need to reduce the number of required CI checks by 2 (as `OBS: openQA:openQA-devel-test - openSUSE_Leap_15.6/x86_64` and `OBS: openQA:openQA-devel-test - SLE_15_SP6_Backports/x86_64` will no longer occur.)

perlpunk:
Now we probably have to decrease the number of required checks by 2, because there are 2 less OBS checks

edit: sorry, ignore me, that was already part of this PR

perlpunk (approved):

okurz (approved):

PR #7386, opened 2026-05-05
parser: ktap: Ignore selftest status comments

rbmarliere:
Kselftest logs can include inner status comments like: 

\# selftests: test [PASS]

These are diagnostics, not group headers, so do not use them to create groups. Keep the group name based on the actual selftest header only, without the collection prefix.

VR:
- **before** https://rmarliere-openqa.qe.prg2.suse.org/tests/1998#step/selftests:test_xdp_redirectxdpdrv[PASS]/1
- **after** https://rmarliere-openqa.qe.prg2.suse.org/tests/2000#step/test_xdp_redirect_sh/1

### DIFF
--- a/lib/OpenQA/Parser/Format/KTAP.pm
+++ b/lib/OpenQA/Parser/Format/KTAP.pm
@@ -9,6 +9,14 @@ use Carp qw(confess);
 use OpenQA::Parser::Result::OpenQA;
 use TAP::Parser;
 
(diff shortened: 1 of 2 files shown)


okurz (changes requested):
As you are repeating the same non-trivial regex three times please extract it into a constant or the check into a function

okurz (approved):

Martchus (approved):

PR #7388, opened 2026-05-06
test: stabilize test_containers_compose for parallel and robust execution

okurz:
Motivation:
Running container compose tests in parallel or repeated runs was unstable due
to port conflicts, database migration race conditions, and non-unique project
names.

Design Choices:
- Use unique, sanitized project names for each docker compose invocation.
- Randomize host port mappings to avoid collisions.
- Remove redundant webui-db-init service from tests to prevent database
  migration race conditions.
- Implement robust health checking for all containers before executing commands.
- Use 'docker compose run' instead of 'exec' for better network alias support.
- Fix worker test data dependency by copying all relevant container source.
- Ensure correct environment passing to sudo commands.
- Make nginx entrypoint project-agnostic.

Benefits:
Reliable and parallel-safe execution of container integration tests.

Verified with:

```
count-fail-ratio make -j test-containers-compose
```

### DIFF
--- a/container/helm/openqa/templates/deployment.yaml
+++ b/container/helm/openqa/templates/deployment.yaml
@@ -130,7 +130,6 @@ spec:
               command: ["curl", "-f", "http://localhost:9526"]
         - name: {{ include "openqa.fullname" . }}-nginx
           image: "nginx:alpine"
-          imagePullPolicy: IfNotPresent
           volumeMounts:
             - name: nginx-config
               mountPath: /etc/nginx/nginx.conf
--- a/tools/test_containers_compose
+++ b/tools/test_containers_compose
@@ -27,37 +27,62 @@ wait_until() {
 }
 
 setup_containers() {
-    "$thisdir"/retry sudo docker compose build
+    local project_name="$1"
+    # Remove fixed port mappings and webui-db-init service to avoid conflicts and race conditions
+    python3 -c "
+import yaml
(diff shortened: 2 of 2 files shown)


Martchus (approved):

perlpunk (approved):

kalikiana:
```
shfmt -d -i 4 -bn -ci -sr $(file --mime-type script/* t/* container/worker/*.sh tools/* | sed -n 's/^\(.*\):.*text\/x-shellscript.*$/\1/p')
--- tools/test_containers_compose.orig
+++ tools/test_containers_compose
@@ -50,7 +50,7 @@
     fi
     # Wait for all services with healthchecks to become healthy
     for id in $(sudo docker compose -p "$project_name" ps -q); do
-        if sudo docker inspect "$id" --format '{{.State.Health}}' 2>/dev/null | grep -q "<nil>"; then
+        if sudo docker inspect "$id" --format '{{.State.Health}}' 2> /dev/null | grep -q "<nil>"; then
             continue
         fi
         wait_until "sudo docker inspect $id --format '{{.State.Health.Status}}' | grep -q healthy" 60
make: *** [Makefile:413: test-shfmt] Error 1
```

```
1: T1 Title exceeds max length (73>72): "test: stabilize test_containers_compose for parallel and robust execution"
Error: Process completed with exit code 1.
```

PR #7395, opened 2026-05-08
feat(search): Add job id to search results response

foursixnine:
Often I need to find out in which job was a module last run, in the current state I need to use the webUI to navigate to the correct job group which is not optimal, having the job id at least exposed via api makes it easier to find what I'm looking for, so I can build scripts on top


perlpunk (approved):
Yeah. We could then also link to the job in the web ui.

Martchus (approved):

foursixnine:
> Yeah. We could then also link to the job in the web ui.

That's the idea :), but wanted to submit it separately https://github.com/os-autoinst/openQA/pull/7397

PR #7405, opened 2026-05-09
style: Enforce perlcritic policy Subroutines::ProhibitExcessComplexity

perlpunk:
This PR sets a high threshold for the complexity so that it won't currently fail, and we can gradually lower the threshold in a followup ticket.

https://metacpan.org/pod/Perl::Critic::Policy::Subroutines::ProhibitExcessComplexity

Policy Description:
> All else being equal, complicated code is more error-prone and more expensive
> to maintain than simpler code. The first step towards managing complexity is
> to establish formal complexity metrics. One such metric is the McCabe score,
> which describes the number of possible paths through a subroutine. This
> Policy approximates the McCabe score by summing the number of conditional
> statements and operators within a subroutine. Research has shown that a
> McCabe score higher than 20 is a sign of high-risk, potentially untestable
> code. See http://en.wikipedia.org/wiki/Cyclomatic_complexity for some
> discussion about the McCabe number and other complexity metrics.


okurz (approved):

.perlcriticrc
```diff
@@ -53,3 +53,7 @@ min_value = 1000000
 [Modules::ProhibitExcessMainComplexity]
 # Reduce me!
 max_mccabe = 83
+
+[Subroutines::ProhibitExcessComplexity]
+# Reduce me!
```
d3flex, line 62: I wonder how we gonna track this? seems easy forgotten thing. despite is there a ticket which could attach here?

d3flex (approved):
All those reduce me - it would be nice to link to some progress ticket for reminders/tracking.

.perlcriticrc
perlpunk: I already planned to create at least one followup ticket, but I don't know yet how to split them up, so I'd like to avoid adding a ticket reference at this point that might be the wrong one later.

PR #7407, opened 2026-05-09
style: Enforce perlcritic policy Subroutines::ProhibitManyArgs

perlpunk:
This PR sets a high threshold for the allowed number so that it won't currently fail, and we can gradually lower the threshold in a followup ticket.

https://metacpan.org/pod/Perl::Critic::Policy::Subroutines::ProhibitManyArgs

Policy description:
> Subroutines that expect large numbers of arguments are hard to use because
> programmers routinely have to look at documentation to remember the order of
> those arguments. Many arguments is often a sign that a subroutine should be
> refactored or that an object should be passed to the routine.


perlpunk:
This still fails with the current version of Perl::Critic in Leap because in version 1.152 the policy does not work correctly with signatures.
I tested 1.156 and there it works.

The current "winner" is `_parse_job_template_machines` in lib/OpenQA/Schema/Result/JobGroups.pm with 10 arguments.
https://github.com/os-autoinst/openQA/blob/master/lib/OpenQA/Schema/Result/JobGroups.pm#L459-L462

perlpunk:
> This still fails with the current version of Perl::Critic in Leap because in version 1.152 the policy does not work correctly with signatures.
> I tested 1.156 and there it works.

I created:
* https://src.opensuse.org/pool/perl-Perl-Critic/pulls/1
* https://src.opensuse.org/pool/perl-Perl-Critic/pulls/2

perlpunk:
New Perl::Critic is in Leap 16.0. Rebased to check if the container also already contains it

edit: cool, tests pass

d3flex (approved):

okurz (approved):

PR #7410, opened 2026-05-09
WIP style: Enforce perlcritic rule BuiltinFunctions::ProhibitComplexMappings

perlpunk:
Do we want this or at least limit the number of statements?

---
https://metacpan.org/pod/Perl::Critic::Policy::BuiltinFunctions::ProhibitComplexMappings

Policy description:
> The map function can be confusing to novices in the best of circumstances.
> Mappings with multiple statements are even worse. They're also a maintainer's
> nightmare because any added complexity decreases readability precipitously.
> Why? Because map is traditionally a one-liner converting one array to
> another. Trying to cram lots of functionality into a one-liner is a bad idea
> in general.

> The best solutions to a complex mapping are: 1) write a subroutine that
> performs the manipulation and call that from map; 2) rewrite the map as a for
> loop.


Wabri:
I'm preparing a draft with all changes needed to enforce this policy

perlpunk:
I don't like that you put all these changes in the very same commit that I created just for adding the policy.

lib/OpenQA/Parser/Result/OpenQA.pm
```diff
@@ -18,7 +18,9 @@ sub new ($class, @args) { $class->SUPER::new(@args)->parsed_details }
 
 # Adds _source => 'parser' to all the details of the result
 sub parsed_details ($self) {
-    return $self->details([map { $_->{_source} = 'parser'; $_ } @{$self->details}]);
+    my $details = $self->details;
+    $_->{_source} = 'parser' for @$details;
+    return $self->details($details);
```
perlpunk, line 23: I think we should *at least* allow 2 statements in a map. Otherwise we restrict ourselves too much IMHO.

perlpunk (commented):
I would request changes, but github doesn't let me. Maybe because I am still the author of the PR.
@Wabri please create a new PR

lib/OpenQA/Parser/Result/OpenQA.pm
Wabri: From 20+ critics to 7, not bad :) Let me update and propose a new pr

Wabri:
New pr opened #7789

PR #7419, opened 2026-05-11
fix(test overview): Bring query for "Aborted" in-line with accounting

Martchus:
We count only jobs in state `done` as "Aborted" so this change brings the filtering query in-line with that.

Note that this was already the case before recent improvements. (Before those improvements, jobs in state `cancelled` were counted as "None" so "Aborted" was always just jobs in state `done`.)

Related ticket: https://progress.opensuse.org/issues/200594


okurz (approved):

Martchus:
E.g. if one is on https://openqa.opensuse.org/tests/overview?groupid=1&version=Tumbleweed&t=2026-05-11+07%3A10%3A05+%2B0000 there are two aborted jobs. If one clicks on the filter there are still 3 jobs. Appending `&state=done` fixes that.

We could of course consolidate "Cancelled" and "Aborted" so there's just one counter and filtering query that counts/shows both. I can do that in another commit if you want.

perlpunk (approved):

PR #7421, opened 2026-05-11
feat: Add `pyproject.toml` as in other repositories

Martchus:
Related ticket: https://progress.opensuse.org/issues/198023

---

There's no CI job yet because there are no files this would apply to anyway.

I could create a separate PR enabling this for `container/openqa_data/scripts/client-conf` which is the only Python script I found in the repository. This file is not used by us and has no tests. So making it fulfill all the rules we set here would be quite some work and we probably don't want to spend that effort. (That's how the changes would look like: https://github.com/os-autoinst/openQA/commit/e0e56112414fc37e34db738c05ff32da8b19ecde - but there are still 33 remaining errors including to use type hints everywhere.)

So I'm not sure whether it makes sense to proceed with this at this point where we simply don't really use Python anyway.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,7 +1,7 @@
 [project]
 name = "openQA"
 dynamic = ["version"]
-description = "openQA web-frontend, scheduler and tools."
+description = "Framework for automated system-level testing (web-frontend, scheduler and tools)"
 readme = "README.md"
 requires-python = ">=3.6"
 license = "GPL-2-or-later"
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,0 +1,149 @@
+[project]
+name = "openQA"
+dynamic = ["version"]
+description = "openQA web-frontend, scheduler and tools."
+readme = "README.md"
+requires-python = ">=3.6"
+license = "GPL-2-or-later"
+authors = [
+    { name = "SUSE LLC" },
+]
+classifiers = [
+    "Programming Language :: Python :: 3",
+    "Operating System :: OS Independent",
+]
+dependencies = [
+]
+
+[project.urls]
+Homepage = "https://open.qa"
+
+[dependency-groups]
+dev = [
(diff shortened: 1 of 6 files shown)


okurz (changes requested):
https://github.com/os-autoinst/openQA/blob/master/script/openqa-label-all *is* Python so I guess you need to adjust rules to catch files like those

Martchus:
We can actually format the Python code snippets in the documentation with `ruff`. So I added an additional commit for that.

perlpunk (approved):

pyproject.toml
```diff
@@ -0,0 +1,149 @@
+[project]
+name = "openQA"
+dynamic = ["version"]
+description = "openQA web-frontend, scheduler and tools."
```
d3flex: this likely should also change. see https://github.com/os-autoinst/openQA/pull/7459 and obs builds for the rpmlint output (`name-repeated-in-summary`)

d3flex (dismissed):

pyproject.toml
Martchus: What has the description in this file to do with the summary in rpm spec files?

Martchus: By the way, I took this from the "About" meta-data of the GitHub repository (not the spec file).

okurz (approved):

pyproject.toml
d3flex: just to align with the description on the spec. we use the same wording but I had to change that due to the rpmlint check

Martchus: What we have in the spec file right now sounds wrong:

> Automated web-frontend testing framework, scheduler and tools

This makes it sound like openQA tests web frontends. However, it *is* a web frontend and tests are system-level tests.

I'll fix the spec file and streamline both descriptions.

okurz (approved):

PR #7423, opened 2026-05-11
style: Enforce perlcritic policy Community::ConditionalImplicitReturn

perlpunk:
https://metacpan.org/pod/Perl::Critic::Policy::Community::ConditionalImplicitReturn

Policy description:
> If the last statement in a subroutine is a conditional block such as `if
> ($foo) { ... }`, and the else condition is not handled, the subroutine will
> return an unexpected value when the condition fails, and it is most likely a
> logic error. Specify a return value after the conditional, or handle the else
> condition.


okurz (approved):

lib/OpenQA/Worker/Job.pm
```diff
@@ -293,6 +293,7 @@ sub _handle_engine_startup ($self, $engine, $max_job_time) {
 sub kill ($self) {
     return undef unless my $engine = $self->engine;
     if (my $child = $engine->{child}) { $child->stop }
+    return undef;
```
d3flex, line 296: just a note but this could have been refactored to avoid the first return.

d3flex (approved):

Martchus (approved):

perlpunk:
rebased and marked that one line as uncoverable

PR #7426, opened 2026-05-11
style: enforce perlcritic policy ProhibitCascadingIfElse

perlpunk:
https://metacpan.org/pod/Perl::Critic::Policy::ControlStructures::ProhibitCascadingIfElse

Policy description:
> Long if-elsif chains are hard to digest, especially if they are longer than a
> single page or screen. If testing for equality, use a hash lookup instead.


okurz (approved):

d3flex (approved):
`# Reduce me!` ))

PR #7429, opened 2026-05-12
test: Add `python3-ruff` for static checks of Python scripts

Martchus:
So far this is only about `openqa-label-all`. Static tests will be added for this script part of https://progress.opensuse.org/issues/198023.

### DIFF
--- a/dependencies.yaml
+++ b/dependencies.yaml
@@ -223,3 +223,4 @@ style_check_requires:
   ShellCheck:
   shfmt:
   python3-gitlint:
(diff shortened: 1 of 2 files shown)


okurz (approved):

perlpunk (approved):

Martchus:
By the way, I cross checked that the dependency will only be added to the `-devel` package which we already don't build anymore for Leap 15.6 so this shouldn't cause any unresolvable dependencies.

PR #7430, opened 2026-05-12
feat(worker): Prevent jobs from being skipped without obvious reason

Martchus:
Check out individual commit messages for details.

Related ticket: https://progress.opensuse.org/issues/198239

### DIFF
--- a/lib/OpenQA/Worker.pm
+++ b/lib/OpenQA/Worker.pm
@@ -519,7 +519,7 @@ sub _accept_or_skip_next_job_in_queue ($self) {
     if (my $e = $self->current_error) {
         if ($self->current_error_is_fatal) {
             log_info "Skipping job $next_job_id from queue because worker is broken ($e)";
-            return $self->_prepare_and_skip_job($next_job);
+            return $self->_prepare_and_skip_job($next_job, WORKER_SR_BROKEN);
         }
         else {
             log_info "Continuing with job $next_job_id as it is already enqueued despite current error ($e)";
--- a/lib/OpenQA/Worker.pm
+++ b/lib/OpenQA/Worker.pm
@@ -666,12 +666,12 @@ sub check_availability ($self) {
     if (my $cache_service_client = $self->{_cache_service_client}) {
         my $error = $cache_service_client->info->availability_error;
         my $host = $cache_service_client->host // '?';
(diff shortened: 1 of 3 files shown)


okurz (approved):

t/24-worker-overall.t
```diff
@@ -887,8 +889,14 @@ subtest 'handle job status changes' => sub {
             };
             $reset_job_and_queue->();
             $worker->current_error('fake fatal error')->current_error_is_fatal(1);
+            #$worker->_accept_or_skip_next_job_in_queue();
```
d3flex: left over?

lib/OpenQA/Worker.pm
```diff
@@ -666,23 +666,23 @@ sub check_availability ($self) {
...
         my $error = $cache_service_client->info->availability_error;
         my $host = $cache_service_client->host // '?';
-        return (0, "Worker cache not available via $host: $error") if $error;
+        return (1, "Worker cache not available via $host: $error") if $error;
     }
 
     # check whether qemu is still running
     if (my $qemu_pid = $self->is_qemu_running) {
-        return (0, "A QEMU instance using the current pool directory is still running (PID: $qemu_pid)");
+        return (1, "A QEMU instance using the current pool directory is still running (PID: $qemu_pid)");
```
d3flex, line 674: Does this make sense? I guess there are consequences as this looks quite fatal. What happens to the other running job?

d3flex (changes requested):
blocked because I worry about the `return (1, "A QEMU instance using the current pool directory is still running (PID: $qemu_pid)");`. I do not think it is a good idea to let it go from here. I guess any other solution will require further changes. What about to leave this as is and come up with something in another PR?

lib/OpenQA/Worker.pm
Martchus: Nothing bad will happen to the other running job if we ignore this. The other job has probably already finished (probably not successfully) or it is stuck (and will incomplete or time out anyway). So starting the next job despite this error will not make the mess bigger than it already is so we might as well try.

perlpunk (approved):

lib/OpenQA/Worker.pm
d3flex: lets try

d3flex (approved):

PR #7431, opened 2026-05-12
ci: Add distinct container image for CI

Martchus:
This container installs packages from `devel:openQA` (instead of only from `openSUSE:Tumbleweed`). This has the advantage that we don't have to wait for package submissions to be accepted to add a new dependency for our CI environment. This container also avoids dependencies like `mc` and `vim` that might be useful for development but are not needed in a CI environment.

Commit 1a8dbebc8 added `perl-TAP-Harness-JUnit` to the devel container. We can remove that package from there again after configuring the new container added by this commit.

### DIFF
--- a/container/ci/Dockerfile
+++ b/container/ci/Dockerfile
@@ -1,0 +1,18 @@
+# SPDX-License-Identifier: MIT
+#!BuildTag: opensuse/openqa_ci:latest opensuse/openqa_ci:%PKG_VERSION% opensuse/openqa_ci:%PKG_VERSION%.%RELEASE%
+
+# Match chromium architectures
+#!ExclusiveArch:  x86_64 aarch64 riscv64
+
+# hadolint ignore=DL3006
+FROM opensuse/tumbleweed
+
+# hadolint ignore=DL3037
+RUN zypper ar -p 95 -f 'http://download.opensuse.org/repositories/devel:openQA/openSUSE_Tumbleweed' devel_openQA && \
+    zypper --gpg-auto-import-keys ref && \
+    zypper in -y os-autoinst-devel openQA-devel nodejs-default perl-TAP-Harness-JUnit && \
+    zypper clean -a
+
+ENV OPENQA_DIR=/opt/openqa
+VOLUME $OPENQA_DIR
+WORKDIR $OPENQA_DIR


perlpunk (approved):

container/ci/Dockerfile
```diff
@@ -0,0 +1,30 @@
+# Specify the license of the container build description (see also the LICENSE file)
+# SPDX-License-Identifier: MIT
+# Define the names/tags of the container
+#!BuildTag: opensuse/openqa_devel:latest opensuse/openqa_devel:%PKG_VERSION% opensuse/openqa_devel:%PKG_VERSION%.%RELEASE%
```
okurz:
```suggestion
# SPDX-License-Identifier: MIT
#!BuildTag: opensuse/openqa_devel:latest opensuse/openqa_devel:%PKG_VERSION% opensuse/openqa_devel:%PKG_VERSION%.%RELEASE%
```

container/ci/Dockerfile
```diff
@@ -0,0 +1,30 @@
...
+
+# Match chromium architectures
+#!ExclusiveArch:  x86_64 aarch64 riscv64
+
+# hadolint ignore=DL3006
+FROM opensuse/tumbleweed
+
+# Define labels according to https://en.opensuse.org/Building_derived_containers
+# labelprefix=org.opensuse.openqa_devel
```
okurz: Is this (also) boilerplate left-overs?

okurz (changes requested):

container/ci/Dockerfile
Martchus: I'm not sure. I've just copied over what we have in the devel container. We can probably remove most of these comments.

container/ci/Dockerfile
```diff
@@ -0,0 +1,18 @@
...
+#!BuildTag: opensuse/openqa_ci:latest opensuse/openqa_ci:%PKG_VERSION% opensuse/openqa_ci:%PKG_VERSION%.%RELEASE%
+
+# Match chromium architectures
+#!ExclusiveArch:  x86_64 aarch64 riscv64
+
+# hadolint ignore=DL3006
+FROM opensuse/tumbleweed
+
+# hadolint ignore=DL3037
+RUN zypper ar -p 95 -f 'http://download.opensuse.org/repositories/devel:openQA/openSUSE_Tumbleweed' devel_openQA && \
```
d3flex, line 11: is there any reason that you dont use `https`?

container/ci/Dockerfile
```diff
@@ -0,0 +1,18 @@
+# SPDX-License-Identifier: MIT
+#!BuildTag: opensuse/openqa_ci:latest opensuse/openqa_ci:%PKG_VERSION% opensuse/openqa_ci:%PKG_VERSION%.%RELEASE%
+
+# Match chromium architectures
+#!ExclusiveArch:  x86_64 aarch64 riscv64
```
d3flex, line 5: I guess you do not need an extra space here

okurz (approved):

container/ci/Dockerfile
Martchus: Good question. This was C&P from [our documentation](https://open.qa/docs/#development-version-repository). If there's no good reason we should probably change the documentation and the `Dockerfile`.

PR #7440, opened 2026-05-12
fix: parse URL-derived settings for cloned jobs

okurz:
Motivation:
Cloned jobs (using `is_clone_job => 1`) skip `_generate_job_setting` to
avoid overriding cloned settings with database defaults. However, this
also bypasses `parse_url_settings`, meaning that if a user provides an
`ISO_URL` override while cloning, the corresponding `ISO` variable is
not automatically deduced. This leads to jobs that download the ISO but
fail to inform the test backend which ISO to use.

Design Choices:
Explicitly call `OpenQA::JobSettings::parse_url_settings` for cloned
jobs in `_create_job`. This ensures that URL-derived settings are
populated from any provided `*_URL` variables without interfering with
the cloning process's goal of avoiding database defaults.

Benefits:
Ensures uniform behavior for `*_URL` variables across all job creation
paths. Users can now override `ISO_URL` during cloning and expect the
`ISO` variable to be correctly set, matching the documented behavior.

Related issue: https://github.com/openSUSE/qem-bot/pull/522

### DIFF
--- a/lib/OpenQA/WebAPI/Controller/API/V1/Job.pm
+++ b/lib/OpenQA/WebAPI/Controller/API/V1/Job.pm
@@ -321,6 +321,9 @@ sub _create_job ($self, $global_params, $job_suffix, $job_specific_params, $mini
         die "$result->{error_message}\n" if defined $result->{error_message};
         $job_settings = $result->{settings_result};
     }
+    else {
+        OpenQA::JobSettings::parse_url_settings($job_settings);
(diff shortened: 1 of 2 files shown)


Amrysliu (approved):
LGTM

jlausuch (approved):

Martchus (approved):

nicksinger (approved):

PR #7446, opened 2026-05-14
fix: align productlog ajax columns with DataTable order

### LINKED ISSUE #6232: Sorting for /admin/productlog not working
Sorting by 'Flavor' did not result in an ordered view.

To reproduce:
* Go to https://openqa.opensuse.org/admin/productlog
* Click on the 'Flavor' column header

See the image at https://pioneers.game-host.org/adminlog_sorting.png

officialasishkumar:
## Motivation

Clicking the column headers on `/admin/productlog` sorts the rows by a
different attribute than the one displayed. For instance, sorting by
`Flavor` produces a list ordered by `Version`. This affects every
sortable column except `ID` and `Time` because the server-side list of
sortable columns does not match the column order rendered by the
DataTable.

## Design Choices

The DataTable defined in `assets/javascripts/audit_log.js` maps each
column index (0..10) to a specific data field. The controller passes an
array to `OpenQA::WebAPI::ServerSideDataTable::render_response` that
must use exactly the same order so the `order[<n>][column]` parameter
from DataTables refers to the intended database column.

The previous implementation reused the `@searchable_columns` list as
the sortable columns and also included `me.settings` and `me.results`
which are not represented as columns in the table. As a result the
indices drifted by two starting at the `Status` column and produced
incorrect or undefined sort orders.

This change replaces the reused list with an explicit array whose
indices line up with the DataTable. A comment links the two sides so
future modifications cannot silently desynchronize them.

## Benefits

Sorting by `Status`, `Distri`, `Version`, `Flavor`, `Arch`, `Build` and
`ISO` now matches what the user clicks. The placeholder entries used
for the `User` and `Actions` columns are inert because both are marked
`orderable: false` in the DataTable, but they keep the array length
correct so any future change that makes those columns sortable can pick
a sensible default rather than reading past the end of the array.

A new `t/45-productlog-sort.t` test populates the table with scrambled
column values and asserts the expected order for every sortable column
to prevent the columns from drifting out of sync again.

Closes https://github.com/os-autoinst/openQA/issues/6232


lib/OpenQA/WebAPI/Controller/Admin/AuditLog.pm
```diff
@@ -34,10 +34,27 @@ sub productlog_ajax ($self) {
         push @filter_conds, {-or => [map { $_ => \%condition } @searchable_columns]};
     }
 
+    # NOTE: The column list must be aligned with the column order of the
+    # DataTable defined in `assets/javascripts/audit_log.js` so that
+    # sorting requests refer to the intended database columns.
+    my @columns = (
+        'me.id',    # 0: ID
+        'me.t_created',    # 1: Time
+        'me.t_created',    # 2: User (orderable: false, placeholder)
```
d3flex: I would say that this is misleading. The second index in the UI points to `User`. Functional doesnt make sense apparently, which smells IMO. Same for the index 10.

d3flex (changes requested):
This seems to work but it would be nice to address the use of the placeholders in `@columns`. 

In additional I wonder if there are alredy fixtures which you could use, in order to simulate this with more realistic datasets

Lastly a related issue is https://progress.opensuse.org/issues/186870. please add it to the commit.

t/45-productlog-sort.t
```diff
@@ -0,0 +1,104 @@
...
+use Test::Most;
+
+use FindBin;
+use lib "$FindBin::Bin/lib", "$FindBin::Bin/../external/os-autoinst-common/lib";
+use Mojo::Base -signatures;
+use Mojo::URL;
+use Test::Mojo;
+use Test::Warnings ':report_warnings';
+use OpenQA::Test::Case;
+use OpenQA::Test::TimeLimit '10';
```
d3flex:
```suggestion
use OpenQA::Test::TimeLimit '5';
```

Likely 10 is too much for that, no?

lib/OpenQA/WebAPI/Controller/Admin/AuditLog.pm
officialasishkumar: Addressed by using explicit `undef` entries for non-orderable User/Actions columns and skipping undefined mappings in server-side ordering.

t/45-productlog-sort.t
officialasishkumar: Changed to `OpenQA::Test::TimeLimit '5'`.

officialasishkumar:
I checked for reusable scheduled-products fixtures and did not find one, so the focused test keeps locally created rows but now uses more realistic product values. The progress issue is included in the review-fix commit body.

Martchus (commented):
Looks generally good. Make sure the style check and other checks pass. (I've just approved the workflow run which still hasn't finished so you probably want to await pending checks until pushing changes again.)

You can have multiple commits on the branch but it would be good to avoid merge commit and *all* commits have to follow https://www.conventionalcommits.org.

d3flex:
@officialasishkumar do you need any help?

PR #7450, opened 2026-05-15
feat: add llm server deployment support

okurz:
* feat: add Nginx proxy template for local LLM load balancing
* feat: add openQA-llm-server sub-package for local LLM support

### DIFF
--- a/container/systemd/openqa-llm-server.container
+++ b/container/systemd/openqa-llm-server.container
@@ -1,0 +1,18 @@
+[Unit]
+Description=Llama Cpp Server for openQA
+After=network-online.target
+
+[Container]
+Image=ghcr.io/ggml-org/llama.cpp:server
+ContainerName=openqa-llm-server
+Volume=/opt/llm/models:/models
+PublishPort=8080:8080
+Exec=-hf unsloth/gemma-4-E4B-it-GGUF:UD-IQ2_M --temp 1.0 --top-p 0.95 --top-k 64 --ctx-size 65536
+User=_openqa-worker
+
+[Service]
+Restart=always
+TimeoutStartSec=900
+
+[Install]
+WantedBy=multi-user.target
--- a/dist/rpm/openQA.spec
+++ b/dist/rpm/openQA.spec
@@ -361,6 +361,16 @@ Supplements:    (%{name}-client and zsh)
 %description client-zsh-completion
 The official zsh completion script for openqa-cli.
 
+%package llm-server
+Summary:        Local LLM Server features for openQA workers
+Requires:       %{name}-worker = %{version}
+Requires:       podman
+
+%description llm-server
+openQA workers can optionally host a local llama.cpp server to provide
+LLM features directly from the worker node. This package provides the
+Podman Quadlet configuration to automatically manage the server.
+
 %prep
 %setup -q
 sed -e 's,/bin/env python,/bin/python,' -i script/openqa-label-all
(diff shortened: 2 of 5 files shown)


container/systemd/openqa-llm-server.container
```diff
@@ -0,0 +1,18 @@
+[Unit]
+Description=Llama Cpp Server for openQA
+After=network-online.target
+
+[Container]
+Image=ghcr.io/ggml-org/llama.cpp:server
+ContainerName=openqa-llm-server
+Volume=/opt/llm/models:/models
+PublishPort=8080:8080
+Exec=-hf unsloth/gemma-4-E4B-it-GGUF:UD-IQ2_M --temp 1.0 --top-p 0.95 --top-k 64 --ctx-size 65536
```
kalikiana, line 10: Might be nice to use a template unit here so the model can be changed depending on what's suitable for where it's deployed?

kalikiana: Alternative read this from /etc so it can be changed that way

d3flex: it would be nice to have in the the commit the reasoning of the selected options. the ctx-size seems small enough but maybe too small?!

d3flex:
Should this have a review from the SUSE IT security team?

okurz:
> Should this have a review from the SUSE IT security team?

The SUSE IT security team has nothing to do with this openQA upstream contribution.

container/systemd/openqa-llm-server.container
okurz: I considered using a template unit (@.container), but I decided to stick with a standard unit for a few reasons:
1. Automatic Deployment: The primary design goal for this sub-package is a "zero-config" install. By using a standard unit, the RPM scriptlets can automatically enable and start the service upon installation (zypper in openQA-llm-server), which isn't as straightforward (or not possible?) with template units.
2. Configurability: Changing the model often requires adjusting other parameters like context size or sampling flags. A template unit only gives us one variable (%I), which is not be enough for a clean configuration.
3. Extensibility: Users who need a different model can still easily customize the deployment using standard systemd drop-ins (e.g., `systemctl edit openqa-llm-server.service`) to override the `Exec=` line without breaking the package's default behavior.

That said, if we see a need for multiple concurrent model instances on the same worker in the future, we can revisit the template approach but maybe we wouldn't use quadlets in this form anyway.

> it would be nice to have in the the commit the reasoning of the selected options. the ctx-size seems small enough but maybe too small?!

This context size is what is currently in use in the o3 infrastructure for the automated LLM investigation. 64k is already considerably bigger than default. Note that increased context means significantly more use of memory and also slower operations so we should not just bump it up further unless it's actually necessary for the specific use cases.

dist/rpm/openQA.spec
```diff
@@ -592,6 +602,15 @@ fi
 %postun local-db
 %service_del_postun %{openqa_localdb_services}
 
+%post llm-server
+%service_add_post openqa-llm-server.service
```
Martchus, line 623: This references `.service` units but the unit is actually a `.container` unit. Are the `.service` units generated automatically? Have you tested this? I'm just asking because form my work on https://progress.opensuse.org/projects/openqav3/wiki/#o3-s390-and-other-eg-bare-metal-workers-running-as-podman-containers-on-o3-worker-openqaworker23 I know that one at least needed a `systemctl daemon-reload` to make it work.

d3flex (approved):

container/systemd/openqa-llm-server.container
kalikiana: Fair. We can always change it.

kalikiana (approved):

okurz:
The service unit is indeed generated automatically from the `.container` file by `systemd-podman-generator`. I have added an explicit `systemctl daemon-reload` to the `%post` scriptlet of the `llm-server` sub-package to ensure the unit is generated and recognized immediately after installation.

Martchus (approved):

PR #7455, opened 2026-05-18
fix: remove zero-length warnings from openqa build

d3flex:
The `sponsorbox.html.ep` template was a zero-byte file. OBS rpmlint
flags zero-length files as errors. Adding minimal content satisfies
rpmlint without changing any visible output.

`ocs/images/created_with_draw.io` was an empty file used as a
convention marker. OBS rpmlint flags zero-length files as errors.
Removing it avoids the warning without any functional impact, since
the file carried no content.

Issue: https://progress.opensuse.org/issues/65837

### DIFF
--- a/docs/images/README.md
+++ b/docs/images/README.md
@@ -1,0 +1,1 @@
+Architecture diagrams in this directory were created with [draw.io](https://www.drawio.com/).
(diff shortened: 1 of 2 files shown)


perlpunk (approved):

docs/images/created_with_draw.io
okurz, line 1: we should still keep this information somewhere as a file like docs/images/openqa_architecture.xml does otherwise not make it clear how this was used. How about adding a README file in the folder instead? Alternatively remove the file dynamically during the build process.

okurz (changes requested):

okurz (approved):

PR #7456, opened 2026-05-18
feat: Distinguish parallel jobs in info when cluster cannot be assigned

Martchus:
Add a special info if a worker for a job has been picked but the pick was discarded because another job in the parallel cluster misses a free worker.

Related ticket: https://progress.opensuse.org/issues/201015

### DIFF
--- a/lib/OpenQA/Scheduler/Model/Jobs.pm
+++ b/lib/OpenQA/Scheduler/Model/Jobs.pm
@@ -136,7 +136,7 @@ sub _allocate_jobs ($self, $free_workers) {
                     my ($picked_worker, $job_info) = @{$taken{$worker}};
                     $self->_allocate_worker_with_priority($prio, $job_info, $j, $allocated_workers, $picked_worker);
                 }
-                $_->{current_reason} = 'incomplete parallel cluster' for @tobescheduled;
+                $_->{current_reason} = 'no matching worker is free' for @tobescheduled;
                 %taken = ();
                 last;
             }
--- a/lib/OpenQA/Scheduler/Model/Jobs.pm
+++ b/lib/OpenQA/Scheduler/Model/Jobs.pm
@@ -136,7 +136,7 @@ sub _allocate_jobs ($self, $free_workers) {
                     my ($picked_worker, $job_info) = @{$taken{$worker}};
                     $self->_allocate_worker_with_priority($prio, $job_info, $j, $allocated_workers, $picked_worker);
                 }
-                $_->{current_reason} = 'no matching worker is free' for @tobescheduled;
+                $_->{current_reason} //= 'no matching worker is free' for @tobescheduled;
                 %taken = ();
                 last;
             }
@@ -174,6 +174,7 @@ sub _allocate_worker_with_priority ($self, $prio, $job_info, $j, $allocated_work
           . ', reducing priority by '
           . STARVATION_PROTECTION_PRIORITY_OFFSET;
         $j->{priority_offset} += STARVATION_PROTECTION_PRIORITY_OFFSET;
+        $j->{current_reason} //= 'no matching worker for all other jobs in the parallel cluster is free';
     }
     else {
         # don't "take" the worker, but make sure it's not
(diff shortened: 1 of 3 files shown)


Martchus:
This now fails with:

```
not ok - Test::Perl::Critic for "t/05-scheduler-dependencies.t"

#   Failed test 'Test::Perl::Critic for "t/05-scheduler-dependencies.t"'
#   at /usr/lib/perl5/vendor_perl/5.42.0/Test/Perl/Critic.pm line 121.
# 
#   ::warning file=t/05-scheduler-dependencies.t,line=1,col=1,title=Main code has high complexity score (84) - severity 3::[Modules::ProhibitExcessMainComplexity] Consider refactoring
```

I'm wondering why that's the case. I haven't added anything to the main code (only to a sub test).

I guess I could split the test. However, it is relatively quick to execute and probably not that easy to split. Btw, is there a way to execute these style checks for a single file and is it possible to ignore the main complexity check? I've added `## no critic (Modules::ProhibitExcessMainComplexity)` at the beginning of the file but it didn't work.

Martchus:
By the way, I also tested this manually (because missing manual testing was probably why the problem was introduced in the first place). It works but I got a misleading `Scheduling info: scheduling limit reached (max_running_jobs=-1)`. So I guess that's the next problem to be fixed. (It should probably say `dynamic limit n reached` where `n` is the actually computed dynamic limit.)

d3flex:
> This now fails with:
> 
> ```
> not ok - Test::Perl::Critic for "t/05-scheduler-dependencies.t"
> 
> #   Failed test 'Test::Perl::Critic for "t/05-scheduler-dependencies.t"'
> #   at /usr/lib/perl5/vendor_perl/5.42.0/Test/Perl/Critic.pm line 121.
> # 
> #   ::warning file=t/05-scheduler-dependencies.t,line=1,col=1,title=Main code has high complexity score (84) - severity 3::[Modules::ProhibitExcessMainComplexity] Consider refactoring
> ```
> 
> I'm wondering why that's the case. I haven't added anything to the main code (only to a sub test).
> 
> I guess I could split the test. However, it is relatively quick to execute and probably not that easy to split. Btw, is there a way to execute these style checks for a single file and is it possible to ignore the main complexity check? I've added `## no critic (Modules::ProhibitExcessMainComplexity)` at the beginning of the file but it didn't work.

quoting from https://metacpan.org/dist/Perl-Critic/view/lib/Perl/Critic/Policy/Modules/ProhibitExcessMainComplexity.pm

> Complexity is subjective, but formal complexity metrics are still incredibly valuable. Every problem has an inherent level of complexity, so it is not necessarily optimal to minimize the McCabe number. So don't get offended if your code triggers this Policy. Just consider if there might be a simpler way to get the job done.

I think the complexity is valuable feedback to remove. but maybe it shouldnt count the t/ dir. you hit 84 when the limit is 83. I would suggest to just increase that threshold for now.

perlpunk:
> Btw, is there a way to execute these style checks for a single file

```
./tools/perlcritic filename
```

okurz (approved):

perlpunk:
> I guess I could split the test.

The file has a lot of subtests. Maybe some can be grouped together into toplevel subtests.

perlpunk:
> I've added `## no critic (Modules::ProhibitExcessMainComplexity)` at the beginning of the file but it didn't work.

Hm, it works if I add it as the very first line, which would make the shebang not a shebang anymore.
OTOH, shebangs for tests are usually not needed.

perlpunk:
actually this works:
```
#!/usr/bin/env perl ## no critic (Modules::ProhibitExcessMainComplexity)
```
:D

perlpunk:
The problem seems to be that subtest subs (or any anonymous subs) are not treated the same as "real" named subroutines.
https://metacpan.org/module/Perl::Critic::Utils::McCabe/source
I would say that's a bug.
It's possible that most users don't run perlcritic on their tests.
However, I agree that `t/05-scheduler-dependencies.t` can be improved. A lot of subtests mostly on one level after each other, in between some variable initializations where it is not clear where the variables will be used.

okurz (approved):

perlpunk (approved):

PR #7459, opened 2026-05-18
fix(packaging): avoid repeating package name in Summary

d3flex:
rpmlint flags name-repeated-in-summary when the package name appears in the Summary field. Reworded to describe the package without repeating "openQA".

Issue: https://progress.opensuse.org/issues/65837

### DIFF
--- a/dist/rpm/openQA.spec
+++ b/dist/rpm/openQA.spec
@@ -106,7 +106,7 @@
 Name:           openQA
 Version:        5
 Release:        0
-Summary:        The openQA web-frontend, scheduler and tools
+Summary:        Automated web-frontend testing framework, scheduler and tools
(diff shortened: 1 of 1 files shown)


okurz (approved):

dist/rpm/openQA.spec
```diff
@@ -106,7 +106,7 @@
 Name:           openQA
 Version:        5
 Release:        0
-Summary:        The openQA web-frontend, scheduler and tools
+Summary:        Automated web-frontend testing framework, scheduler and tools
```
perlpunk, line 109: Just for the record, I assume you could also use

```suggestion
Summary:        The %{name} web-frontend, scheduler and tools
```

perlpunk (approved):

PR #7460, opened 2026-05-19
feat(needle-editor): Validate that there is at least one match area

Martchus:
The needle editor already validates that areas are present at all. With this change it also validates that at least one match area is present which is required by os-autoinst.

### DIFF
--- a/assets/javascripts/needleeditor.js
+++ b/assets/javascripts/needleeditor.js
@@ -468,6 +468,10 @@ function updateToggleClickCoordinatesButton(hasClickCoorinates) {
   }
 }
 
+function hasMatchAreas(areas) {
+  return Array.isArray(areas) && areas.find(area => area.type === 'match');
+}
+
 function saveNeedle(overwrite) {
   const form = document.getElementById('save_needle_form');
   const errors = [];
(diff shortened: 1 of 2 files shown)


Martchus:
I tested it only manually so far but that's probably also enough.

okurz:
The commit title sounds wrong. Not "all areas" need to be match areas

assets/javascripts/needleeditor.js
```diff
@@ -477,8 +481,8 @@ function saveNeedle(overwrite) {
   }
   const areaSelection = window.needles[$('#area_select').val()];
   const takeMatches = $('#take_matches').prop('checked');
-  if ((!takeMatches && !areaSelection.area.length) || (takeMatches && !areaSelection.matches.length)) {
-    errors.push('No areas defined.');
+  if (!hasMatchAreas(takeMatches ? areaSelection.matches : areaSelection.area)) {
+    errors.push('Not at least one match area defined.');
```
kalikiana: This looks good.

Note the title is saying something else. Maybe `Validate that there is at least one match area` would make more sense?

kalikiana (approved):

assets/javascripts/needleeditor.js
perlpunk: Yeah, "Not at least ..." sounds a bit akward. Maybe phrase it like this?
```suggestion
    errors.push('Please define at least one match area.');
```
or
```suggestion
    errors.push('At least one match area must be defined.');
```

perlpunk: And the PR title still says "Validate that all areas ..."

perlpunk (approved):

PR #7462, opened 2026-05-19
feat: Link to log and source code location from details step view

perlpunk:
Issue: https://progress.opensuse.org/issues/197177

This is how the link to the log looks like:
details-log-link-grey

This is how the source links in the log currently look like:
autoinst-log-with-source-links

### DIFF
--- a/assets/javascripts/render.js
+++ b/assets/javascripts/render.js
@@ -184,9 +184,21 @@ function renderModuleRow(module, snippets) {
       box.push(E('span', [content], {class: 'resborder ' + resborder}));
     }
     if (step.text && title !== 'Soft Failed') {
+      const params = new URLSearchParams({
+        filename: 'autoinst-log.txt',
+        filter: `[step:${module.category},${module.name},${step.num}]`
+      });
+      const logHref = `${window.location.pathname}/logfile?${params.toString()}`;
+      const linkText = E('i', [''], {
+        class: 'fa-solid fa-file-lines fa-xl',
+        title: 'View Logfile',
+        style: 'color: #0077ff'
+      });
+      const sourceLink = E('a', [linkText], {href: logHref});
       const stepActions = E('span', [], {class: 'step_actions', style: 'float: right'});
       stepActions.innerHTML = renderTemplate(snippets.bug_actions, {MODULE: module.name, STEP: step.num});
       const textresult = E('pre', [textData]);
+      stepActions.prepend(sourceLink);
       let html = stepActions.outerHTML;
       html += textresult.outerHTML;
       const txt = escape(html);
--- a/assets/javascripts/test_result.js
+++ b/assets/javascripts/test_result.js
@@ -1,6 +1,8 @@
 // jshint multistr: true
 // jshint esversion: 6
 
+const testGitInfoRe = /TEST_GIT_HASH=([a-fA-F0-9]+) TEST_GIT_URL=([^\p{Cc}]+)/u;
+
 const testStatus = {
   state: null,
   result: null,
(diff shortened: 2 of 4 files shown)


assets/javascripts/render.js
```diff
@@ -184,9 +184,11 @@ function renderModuleRow(module, snippets) {
       box.push(E('span', [content], {class: 'resborder ' + resborder}));
     }
     if (step.text && title !== 'Soft Failed') {
+      const sourceLink = E('a', ['log'], {href: window.location.pathname + '/logfile?filename=autoinst-log.txt&filter=[step:' + module.category + ':' + module.name + ':' + step.num + ']'});
```
Martchus: The URL parameters should be encoded correctly. Maybe it makes also sense to use `URLSearchParams`.

Martchus (commented):
The way the HTML document is changed could probably be improved but this is of course just a draft. So how does this look like?

assets/javascripts/render.js
perlpunk: yep, done

perlpunk:
> So how does this look like?

This is what I currently have:
details-log-link

> The way the HTML document is changed could probably be improved but this is of course just a draft.

So far I'm not even sure how to add it to the screenshot boxes as well.
Hints welcome :)

perlpunk:
> So far I'm not even sure how to add it to the screenshot boxes as well.

I found it. Added the link.
Unfortunately for the text boxes this is done in the frontend, and for the screenshots in the template, so two implementations of the same thing, which means we ideally also test both of them.

perlpunk:
I have to work around the fact that we only add the category to a module in details_ajax, if it's the first module in this category (so that a header is displayed).
I don't have access to previous modules in some places, so I'm not sure how to solve that.
I'm thinking ideally every module should have the category, and we add a `heading => 1` in the backend, when it's the first module in a category.

perlpunk:
This is how the source links in the log currently look like:
autoinst-log-with-source-links

assets/javascripts/test_result.js
```diff
@@ -616,6 +618,17 @@ function delay(callback, ms) {
   };
 }
 
+let testGitHash = '';
+let testGitUrl = '';
```
Martchus, line 622: We probably shouldn't use global variables like this.

assets/javascripts/test_result.js
```diff
@@ -669,6 +685,21 @@ function filterEmbeddedLogFiles() {
...
+const stepRe = / \[step:[a-zA-Z0-9-]+,[a-zA-Z0-9-]+,[0-9]+\] /;
+const sourceRe = /(?<= )(([a-zA-Z0-9_/.-]+\.p[my])):(\d+)/g;
+function createSourceLinks(lineContentElement) {
+  const baseUrl = testGitUrl + '/blob/' + testGitHash;
+  const found = lineContentElement.innerHTML.match(stepRe);
+  if (found) {
+    lineContentElement.innerHTML = lineContentElement.innerHTML.replace(
+      sourceRe,
+      (match, filePath, executionMatch, lineNumber) => {
+        return `<a href="${baseUrl}/${filePath}#L${lineNumber}" target="_blank"><i class="fa-solid fa-file-code fa-lg"></i> ${filePath}:${lineNumber}</a>`;
```
Martchus, line 697: It would be better do do direct DOM manipulations instead of rewriting HTML code.

Martchus (changes requested):

perlpunk:
I created two new PRs out of this which can be independent:
* #7567
* #7568

assets/javascripts/test_result.js
perlpunk: addressed in #7568

assets/javascripts/test_result.js
perlpunk: addressed in #7568, but still using `innerHTML.replace`

PR #7465, opened 2026-05-20
fix(jobs): work around DBD::Pg bug in insert_module row count detection

d3flex:
DBD::Pg returns -1 from $sth->rows() for INSERT ... ON CONFLICT DO NOTHING statements when using prepared statements, instead of the correct value of 0. This caused insert_module to incorrectly increment skipped_module_count when duplicate modules were inserted, breaking the test "inserting the same module twice keeps the job module statistics intact" in t/10-jobs.t.

The fix parses $insert_sth->{pg_cmd_status} instead, which correctly reports "INSERT 0 0" when no rows are inserted due to conflict.

https://github.com/bucardo/dbdpg/issues/194
Issue: https://progress.opensuse.org/issues/201279

Co-Authored-By: Opus 4.6

### DIFF
--- a/lib/OpenQA/Schema/Result/Jobs.pm
+++ b/lib/OpenQA/Schema/Result/Jobs.pm
@@ -1116,7 +1116,9 @@ sub insert_module ($self, $tm, $skip_jobs_update = undef) {
     # execute query to insert job module
     # note: We have 'important' in the DB but 'ignore_failure' in the flags for historical reasons (see #1266).
     my $flags = $tm->{flags};
-    $insert_sth->execute(
+    # note: Using execute() return value instead of $sth->rows which is unreliable for ON CONFLICT DO NOTHING
+    # https://github.com/bucardo/dbdpg/issues/194
+    my $rows = $insert_sth->execute(
         $self->id, @required_fields,
         $flags->{milestone} ? 1 : 0,
         $flags->{ignore_failure} ? 0 : 1,
@@ -1124,7 +1126,7 @@ sub insert_module ($self, $tm, $skip_jobs_update = undef) {
         $flags->{always_rollback} ? 1 : 0,
         $flags->{always_run} ? 1 : 0,
     );
-    return 0 unless $insert_sth->rows;
+    return 0 if $rows <= 0;
 
     # update job module statistics for that job (jobs with default result NONE are accounted as skipped)
     $self->update({skipped_module_count => \'skipped_module_count + 1'}) unless $skip_jobs_update;
@@ -1619,9 +1621,12 @@ sub allocate_network ($self, $name) {
     for ($vlan = 1;; ++$vlan) {
         log_debug "at vlan $name:$vlan";
         next if $used{$vlan};
-        try { $sth->execute($job_id, $name, $vlan) }
+        # note: Using execute() return value instead of $sth->rows which is unreliable for ON CONFLICT DO NOTHING
+        # https://github.com/bucardo/dbdpg/issues/194
+        my $rows;
+        try { $rows = $sth->execute($job_id, $name, $vlan) }
         catch ($e) { die "Failed to create new vlan tag '$vlan' for job $job_id: $e\n" }    # uncoverable statement
-        die "Unable to allocate network for job $job_id: network '$name' already exists" unless $sth->rows;
+        die "Unable to allocate network for job $job_id: network '$name' already exists" if $rows <= 0;
         log_debug "Created network for $job_id: $vlan";
         for my $cluster_job_id (keys %{$self->cluster_jobs}) {
             # apply it for the whole cluster so that the vlan only appears if all of the cluster is gone
--- a/t/04-scheduler.t
+++ b/t/04-scheduler.t
@@ -674,10 +674,10 @@ subtest 'allocating network' => sub {
         #       a transaction (as of 3c52abbe3d6364c02f73c6f9fe4afe44f89688f6) makes one think it may be
         #       necassary. If we ever encounter this error in production we should find out what exactly is
         #       causing it and what behavior would make most sense in that situation.
-        my $jobs_mock = Test::MockModule->new('OpenQA::Schema::Result::Jobs');
-        $jobs_mock->redefine(_find_network => undef);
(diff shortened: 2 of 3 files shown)


okurz (changes requested):
As https://github.com/bucardo/dbdpg/issues/194 is *just* your recent created own issue report I wouldn't see this as proof that there is an issue we need to fix. If this is a "fix" then there should be an according test addition.

d3flex:
> As [bucardo/dbdpg#194](https://github.com/bucardo/dbdpg/issues/194) is _just_ your recent created own issue report I wouldn't see this as proof that there is an issue we need to fix. If this is a "fix" then there should be an according test addition.

you can switch version DBD::Pg and reproduce it. No sure what do you have in mind for according test addition. it is a bug and the test caught it

perlpunk:
It's rather a workaround for the bug in DBD::Pg than a fix. Is `workaround: ...` a commit that will be accepted by gitlint?

lib/OpenQA/Schema/Result/Jobs.pm
```diff
@@ -1124,7 +1124,10 @@ sub insert_module ($self, $tm, $skip_jobs_update = undef) {
         $flags->{always_rollback} ? 1 : 0,
         $flags->{always_run} ? 1 : 0,
     );
-    return 0 unless $insert_sth->rows;
+    # note: $sth->rows is unreliable for ON CONFLICT DO NOTHING
+    # https://github.com/bucardo/dbdpg/issues/194
+    my ($affected) = ($insert_sth->{pg_cmd_status} // '') =~ /INSERT \d+ (\d+)/;
```
perlpunk: `$insert_sth->execute` will actually also return the number, in this case it returns `0E0` which is a special "number" in perl which will be treated as a number without a warning if you use it in a numeric context.
That looks more reliable than looking at `pg_cmd_status` with a regex.
So how about:
```
my $rows = $insert_sth->execute(...);
return 0 if $rows <= 0;
```

d3flex:
> It's rather a workaround for the bug in DBD::Pg than a fix. Is `workaround: ...` a commit that will be accepted by gitlint?

indeed a workaround. I try to find the acceptable keys for the commit but I cant

okurz:
> > It's rather a workaround for the bug in DBD::Pg than a fix. Is `workaround: ...` a commit that will be accepted by gitlint?
> 
> indeed a workaround. I try to find the acceptable keys for the commit but I cant

"fix" is fine.

perlpunk:
btw, there is one more place where we use `ON CONFLICT` and `$sth->rows`
https://github.com/os-autoinst/openQA/blob/master/lib/OpenQA/Resource/Locks.pm#L96-L99
That might just not be covered by our tests, but could possibly break in production.

perlpunk:
I wonder why we don't see failures in CircleCI.
We use registry.opensuse.org/devel/openqa/containers/opensuse/openqa_ci:latest, and that also has perl-DBD-Pg 3.20.2 already

d3flex:
> I wonder why we don't see failures in CircleCI. We use registry.opensuse.org/devel/openqa/containers/opensuse/openqa_ci:latest, and that also has perl-DBD-Pg 3.20.2 already

what do you mean I think the test also failing. at least that was the case in https://github.com/os-autoinst/openQA/pull/7361

image

perlpunk:
> what do you mean I think the test also failing. at least that was the case in #7361

Well, I hadn't seen the failure in CircleCI so far.
Failures started yesterday, and only in OBS.
So the ci container was just built with the new version a bit later.

okurz (approved):

t/04-scheduler.t
```diff
@@ -674,10 +674,10 @@ subtest 'allocating network' => sub {
         #       a transaction (as of 3c52abbe3d6364c02f73c6f9fe4afe44f89688f6) makes one think it may be
         #       necassary. If we ever encounter this error in production we should find out what exactly is
         #       causing it and what behavior would make most sense in that situation.
-        my $jobs_mock = Test::MockModule->new('OpenQA::Schema::Result::Jobs');
-        $jobs_mock->redefine(_find_network => undef);
+        $mock_result->redefine(_find_network => undef);
```
perlpunk, line 677: How is this related?

perlpunk (changes requested):
There are two commits with the same title and body

t/04-scheduler.t
d3flex: this should likely be separate change but I ignore it because the PR was to test the builds. it came from the first thing I checked - which it another version change on Test::Module. from https://metacpan.org/dist/Test-MockModule/changes

## What's Changed
* Fix GH #83: Reverts to pre v0.181.0 behavior. new() returns singleton-per-package by default again. pass distinct => 1 for v0.181's fresh-object-per-call behavior.
* New: Test::MockModule->original_for($pkg, $sub) — call the original sub from inside a mock closure without capturing $mock.

likely is harmless and just a correction after all

perlpunk: > this should likely be separate change but I ignore it

Please make it a separate commit.

d3flex:
> There are two commits with the same title and body

yes. I just coppy paste one to the other. they solve the same thing. but the Locks.pm thought to make it separate one as it was not the reason of the failures

perlpunk:
> > There are two commits with the same title and body
> 
> yes. I just coppy paste one to the other. they solve the same thing. but the Locks.pm thought to make it separate one as it was not the reason of the failures

They don't solve the same thing. Both talk about `insert_module`. How can that be?

Squash the two commits and make a separate one for the Test::MockModule change, ok?

perlpunk:
Also the commit still talks about your previous fix and not the updated fix which uses the return value of `execute`.

perlpunk:
Also the first commit id about two separate fixes, but the description just talks about one of them, and only t/10-jobs.t, but not t/04-scheduler.t

perlpunk:
I created #7470 (but with a more useful split of commits and a more correct commit message). You're still the author of the commits @d3flex

Martchus:
We can close this in favor of https://github.com/os-autoinst/openQA/pull/7470.

PR #7469, opened 2026-05-20
ci: Update package builds after Leap 15.6/SP6 have reached EOL

Martchus:
Related ticket: https://progress.opensuse.org/issues/200814

### DIFF
--- a/.obs/workflows.yml
+++ b/.obs/workflows.yml
@@ -16,13 +16,6 @@ pr:
               - target_project: devel:openQA
                 target_repository: openSUSE_Tumbleweed
             architectures: [ x86_64 ]
-          - name: openSUSE_Leap_15.6
-            paths:
-              - target_project: devel:openQA:Leap:15.6
-                target_repository: openSUSE_Leap_15.6
-              - target_project: devel:openQA
-                target_repository: openSUSE_Leap_15.6
-            architectures: [ x86_64 ]
           - name: SLE_15_SP6_Backports
             paths:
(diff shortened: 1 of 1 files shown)


.obs/workflows.yml
```diff
@@ -16,18 +16,9 @@ pr:
               - target_project: devel:openQA
                 target_repository: openSUSE_Tumbleweed
             architectures: [ x86_64 ]
-          - name: openSUSE_Leap_15.6
+          - name: SLE_15_SP7_Backports
```
perlpunk, line 19: Will this work?
At https://build.opensuse.org/package/show/devel:openQA/openQA I can see that 15.7 is unresolvabe:
> nothing provides perl(MCP) >= 0.80.0

Martchus: The problem is that `perl-MCP` is indeed missing under SP7 (while it is present in SP6):

`perl-MCP` is present under https://build.opensuse.org/project/show/openSUSE:Backports:SLE-15-SP6:Update (under "Packages" specifically) but not under https://build.opensuse.org/project/show/openSUSE:Backports:SLE-15-SP7:Update (under none of the tabs). So I guess that's our problem. Now I just have to find out how to fix that. I would have expected that packages from SP6 are carried over to SP7.

Martchus: I added https://build.opensuse.org/project/show/devel:openQA:Leap:15.7 for now which already looks good for x86_64.

perlpunk (approved):

okurz (approved):

PR #7470, opened 2026-05-20
fix: work around DBD::Pg bug regarding INSERT ON CONFLICT row count

perlpunk:
DBD::Pg returns -1 from $sth->rows() for INSERT ... ON CONFLICT DO
NOTHING statements when using prepared statements, instead of the
correct value of 0. This broke t/10-jobs.t and t/04-scheduler.t.

The fix uses the return value of `$sth->execute` instead, which
also returns the row count and doesn't seem to be affected by
the bug.

bucardo/dbdpg#194
Issue: https://progress.opensuse.org/issues/201279

---
Same as #7465 but with corrected commit message and more useful split of commits. Commit author is still @d3flex


Wabri (approved):
Awesome!

okurz (approved):

PR #7472, opened 2026-05-22
refactor: Use Feature::Compat::Try consistently

perlpunk:
Syntax::Keyword::Try provides a try feature that is compatible with the try function builtin in newer perl versions. Feature::Compat::Try is a module that uses Syntax::Keyword::Try in older versions and automatically uses the builtin try in newer versions, and that is what we use elsewhere as well.

I had simply confused the two modules in this file.

I had to mark one `catch` block as uncoverable. This is probably due to
Devel::Cover missing features regarding those Syntax::Keyword modules, which
previously considered this line falsely as covered.


Martchus (approved):

perlpunk:
Marked a `catch` block as uncovered

okurz (approved):

PR #7476, opened 2026-05-24
feat: add run-test-env target to start all local services

okurz:
Motivation:
Starting each openQA service individually is tedious and error-prone.

Design Choices:
Added a 'run-test-env' target that launches all necessary daemons
(webui, websockets, scheduler, worker, gru) in the background and
uses 'wait' with a 'trap' to ensure they all exit when the main
process is interrupted.

Benefits:
Single command to bring up a full local development/test environment.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -316,6 +316,16 @@ define RUN_SERVICE_TEST_ENV
 	$(1)
 endef
 
+.PHONY: run-test-env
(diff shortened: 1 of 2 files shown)


Martchus (approved):
There are better ways to start openQA for development (see my helpers repo) but if you want a target like this I'm not against it. At least that's a generic "solution" (unlike my helpers which only work with Tmux or Konsole).

perlpunk (approved):

PR #7481, opened 2026-05-25
feat: allow operators to fetch unredacted settings + warn non-operators about redacted settings

okurz:
* feat: request unredacted settings in openqa-clone-job and warn
* feat: allow authenticated operators to fetch unredacted job settings

### DIFF
--- a/lib/OpenQA/Script/CloneJob.pm
+++ b/lib/OpenQA/Script/CloneJob.pm
@@ -109,8 +109,15 @@ sub _get_vars ($jobid, $url_handler, $options) {
 sub clone_job_get_job ($jobid, $url_handler, $options) {
     my $url = $url_handler->{remote_url}->clone;
     $url->path("jobs/$jobid");
+    $url->query->merge(unredacted => 1);
     $url->query->merge(check_assets => 1) unless $options->{'ignore-missing-assets'};
     my $job = _get_with_retry($url_handler, $url, $jobid, 'job', $options)->{job};
+
+    if (my @redacted = grep { ($job->{settings}->{$_} // '') eq '[redacted]' } keys %{$job->{settings}}) {
+        warn "Job $jobid has redacted secrets: " . join(', ', sort @redacted) . "\n";
+        warn "These clone as '[redacted]'. Use operator credentials for actual secrets.\n";
+    }
+
     print STDERR Cpanel::JSON::XS->new->pretty->encode($job) if $options->{verbose};
     $job->{vars} = _get_vars($jobid, $url_handler, $options) if $options->{reproduce};
     return $job;
--- a/lib/OpenQA/WebAPI.pm
+++ b/lib/OpenQA/WebAPI.pm
@@ -78,6 +78,8 @@ sub startup ($self) {
       = $api_public->under('/')->to('Auth#auth_admin')->name('api_ensure_admin');
     my $api_auth_any_user
       = $api_public->under('/')->to('Auth#auth')->name('api_ensure_user');
+    my $api_maybe_auth = $api_public->under('/')->to('Auth#auth_maybe');
+
 
     OpenQA::Setup::setup_template_search_path($self);
     OpenQA::Setup::load_plugins($self, $auth);
@@ -289,6 +291,8 @@ sub startup ($self) {
     my $api_ro = $api_auth_operator->any('/')->to(namespace => 'OpenQA::WebAPI::Controller::API::V1');
     my $api_ra = $api_auth_admin->any('/')->to(namespace => 'OpenQA::WebAPI::Controller::API::V1');
     my $api_public_r = $api_public->any('/')->to(namespace => 'OpenQA::WebAPI::Controller::API::V1');
+    my $api_maybe_auth_r = $api_maybe_auth->any('/')->to(namespace => 'OpenQA::WebAPI::Controller::API::V1');
+
 
     $api_public_r->get('/routes')->name('api_v1_list_routes')->to('routes#list');
     push @api_routes, $api_ru, $api_ro, $api_ra, $api_public_r;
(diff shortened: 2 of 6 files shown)


lib/OpenQA/WebAPI.pm
```diff
@@ -323,20 +327,21 @@ sub startup ($self) {
...
-    $api_public_r->get('/job_settings/jobs')->name('apiv1_get_jobs_for_job_settings')->to('job_settings#jobs');
+    $api_maybe_auth_r->get('/job_settings/jobs')->name('apiv1_get_jobs_for_job_settings')->to('job_settings#jobs');
 
     my $job_r = $api_ro->any('/jobs/<jobid:num>');
     push @api_routes, $job_r;
-    $api_public_r->any('/jobs/<jobid:num>')->name('apiv1_job')->to('job#show');
-    $api_public_r->get('/experimental/jobs/<jobid:num>/status')->name('apiv1_get_status')->to('job#get_status');
-    $api_public_r->any('/jobs/<jobid:num>/details')->name('apiv1_job')->to('job#show', details => 1);
+    $api_maybe_auth_r->any('/jobs/<jobid:num>')->name('apiv1_job')->to('job#show');
+    $api_maybe_auth_r->get('/experimental/jobs/<jobid:num>/status')->name('apiv1_get_status')->to('job#get_status');
```
Martchus: Am I missing something here? It looks like the code doesn't touch `get_status` so changing this line seems wrong.

lib/OpenQA/WebAPI.pm
```diff
@@ -323,20 +327,21 @@ sub startup ($self) {
...
-    $api_public_r->get('/jobs/overview')->name('apiv1_jobs_overview')->to('job#overview');
+    $api_maybe_auth_r->get('/jobs')->name('apiv1_jobs')->to('job#list');
+    $api_maybe_auth_r->get('/jobs/overview')->name('apiv1_jobs_overview')->to('job#overview');
     $api_ro->post('/jobs')->name('apiv1_create_job')->to('job#create');
     $api_ro->post('/jobs/cancel')->name('apiv1_cancel_jobs')->to('job#cancel');
     $api_ro->post('/jobs/restart')->name('apiv1_restart_jobs')->to('job#restart');
 
     # api/v1/job_settings/jobs
-    $api_public_r->get('/job_settings/jobs')->name('apiv1_get_jobs_for_job_settings')->to('job_settings#jobs');
+    $api_maybe_auth_r->get('/job_settings/jobs')->name('apiv1_get_jobs_for_job_settings')->to('job_settings#jobs');
```
Martchus: Am I missing something here? It looks like the code doesn't touch `JobSettings::jobs` so changing this line seems wrong.

t/35-script_clone_job.t
```diff
@@ -78,20 +78,47 @@ subtest 'running command' => sub {
...
+    my $clone_mock = Test::MockModule->new('OpenQA::Script::CloneJob');
+    $clone_mock->redefine(
+        _get_with_retry => sub {
+            return {
+                job => {
+                    id => 42,
+                    settings => {
+                        _SECRET_VAR => '[redacted]',
+                        NORMAL_VAR => 'visible'
+                    }}};
```
Martchus: It would  be good to format this in a more compact way.

t/35-script_clone_job.t
```diff
@@ -78,20 +78,47 @@ subtest 'running command' => sub {
...
+        });
+    $clone_mock->redefine(_get_vars => sub { {} });
+
+    my $url_handler = {remote_url => Mojo::URL->new('http://remote')};
+    my $options = {'ignore-missing-assets' => 1};
+
+    my @w;
+    {
+        local $SIG{__WARN__} = sub { push @w, @_ };
```
Martchus: Isn't there some test helper to check for warnings we normally use instead?

Martchus (changes requested):

t/35-script_clone_job.t
okurz: duh, brainmelt ;) Done, now using Test::Warnings as suggested.

t/35-script_clone_job.t
okurz: Done, formatted more compactly.

lib/OpenQA/WebAPI.pm
okurz: Done, reverted to public route.

lib/OpenQA/WebAPI.pm
okurz: Done, reverted to public route

Martchus (approved):

lib/OpenQA/Script/CloneJob.pm
```diff
@@ -109,8 +109,15 @@ sub _get_vars ($jobid, $url_handler, $options) {
 sub clone_job_get_job ($jobid, $url_handler, $options) {
     my $url = $url_handler->{remote_url}->clone;
     $url->path("jobs/$jobid");
+    $url->query->merge(unredacted => 1);
     $url->query->merge(check_assets => 1) unless $options->{'ignore-missing-assets'};
     my $job = _get_with_retry($url_handler, $url, $jobid, 'job', $options)->{job};
+
+    if (my @redacted = grep { ($job->{settings}->{$_} // '') eq '[redacted]' } keys %{$job->{settings}}) {
+        warn "Job $jobid has redacted secrets: " . join(', ', sort @redacted) . "\n";
+        warn "These clone as '[redacted]'. Use operator credentials for actual secrets.\n";
```
kalikiana, line 118: Imho this should be an error. The message already explains this won't work otherwise.

kalikiana (changes requested):

lib/OpenQA/Script/CloneJob.pm
okurz: We can't decide at this time if jobs will fail due to this or not. I would keep it a warning as I have seen very weird test setups so I wouldn't be surprised if there are secrets that are not actually necessary. Or maybe the test code wants to handle it conditionally. We should keep this non-fatal.

kalikiana (approved):

PR #7483, opened 2026-05-27
feat: watch worker auto-restart systemd drop-ins for reloads

gauravpathak:

Updates `openqa-reload-worker-auto-restart@.path` to monitor the `/etc/systemd/system/openqa-worker-auto-restart@.service.d` directory.  
Any changes to systemd drop-in files will now automatically trigger a graceful worker reload without interrupting active jobs.


okurz (changes requested):
I don't think this should be necessary. I thought we have the .path units to ensure that there is already reloading happening

Martchus (approved):

kalikiana (approved):
Also discussed in the Unblock

r-richardson (approved):

okurz (approved):

PR #7486, opened 2026-05-27
fix: Set correct settings for scheduling example test

Martchus:
* Set `NEEDLES_DIR=%%CASEDIR%%/needles` as the worker otherwise assigns
  `$OPENQA_BASEDIR/openqa/share/tests/example/products/example/needles` as
  per documentation for the sake of compatibility but here it leads to
  `The source directory …/tests/example/products/example/needles does not
  exist` as the needles directory is directly under `CASEDIR`
    * We already do this in the example command in the "Installing"
      documentation.

* Set `PRODUCTDIR=.` as the worker otherwise assigns `products/example` for
  the sake of compatibility but here it leads to `unable to load main.pm`
  as `main.pm` is directly under `CASEDIR`
    * This is not explicitly documented but in-line with the behavior for
      assigning `NEEDLES_DIR`. So for now I'm considering this behavior as
      given; we might want to reconsider this, though.
    * We don't do this in the example command in the "Installing" section
      so it runs into the same error when executed as-is. So this commit
      adds the required setting there as well.

* Note that this once worked out of the box as far as I remember; we have
  probably tweaked the behavior since then to restore compatibility with
  other use cases
    * We might want to reconsider this but here I'm focusing on getting the
      example test working again without affecting anything else.

Related ticket: https://progress.opensuse.org/issues/201546

### DIFF
--- a/docs/Installing.md
+++ b/docs/Installing.md
@@ -69,6 +69,7 @@ openqa-cli schedule --monitor async=1 \
     DISTRI=example VERSION=0 FLAVOR=DVD ARCH=x86_64 \
     TEST=simple_boot _GROUP_ID=0 BUILD=test \
     CASEDIR=https://github.com/os-autoinst/os-autoinst-distri-example.git \
+    PRODUCTDIR=. \
     NEEDLES_DIR=%%CASEDIR%%/needles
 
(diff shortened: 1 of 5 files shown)


okurz (approved):

gauravpathak (approved):
LGTM!

PR #7487, opened 2026-05-27
docs: Fix links in documentation generated via `generate-docs`

Martchus:
The simpler script `generate-docs` which really only generates docs does not process the Markdown files as needed to make links work. This change fixes that by using the code from `generate-documentation` in `generate-docs` as well. This also avoids repeating this not so trivial Perl command.

### DIFF
--- a/tools/process-docs
+++ b/tools/process-docs
@@ -1,0 +1,2 @@
+#!/bin/bash -e
+perl -e 'sub p{my $f=shift;open my $h,"<",$f or die $!;while(<$h>){if(/^include::([^\]]+)\[\]/){p($1)}else{s/\[([^\]]+)\]\([\w-]+\.md(#[\w-]+)\)/[$1]($2)/g;s/\[([^\]]+)\]\(([\w-]+)\.md\)/"[$1](#".lc($2).")"/ge;print}}}p("index.md")'
--- a/tools/process-docs
+++ b/tools/process-docs
@@ -10,7 +10,7 @@ sub process_file ($file) {
         if ($line =~ /^include::([^\]]+)\[\]/) { process_file($1) }
         else {
             $line =~ s/\[([^\]]+)\]\([\w-]+\.md(#[\w-]+)\)/[$1]($2)/g;
-            $line =~ s/\[([^\]]+)\]\(([\w-]+)\.md\)/"[$1](#".lc($2).")"/ge;
+            $line =~ s/\[([^\]]+)\]\(([\w-]+)\.md\)/[$1](#\L$2\E)/g;
             print $line;
         }
     }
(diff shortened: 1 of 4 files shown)


okurz (approved):

okurz (approved):

tools/process-docs
perlpunk, line 1: If it's a bash script and the only command is `perl -e`, I suggest making it a perl script instead :)

perlpunk (changes requested):

tools/process-docs
```diff
@@ -0,0 +1,21 @@
...
+use v5.10;
+use experimental 'signatures';
+
+sub process_file ($file) {
+    open my $handle, "<", $file or die $!;
+    while (<$handle>) {
+        if (/^include::([^\]]+)\[\]/) { process_file($1) }
+        else {
+            s/\[([^\]]+)\]\([\w-]+\.md(#[\w-]+)\)/[$1]($2)/g;
+            s/\[([^\]]+)\]\(([\w-]+)\.md\)/"[$1](#".lc($2).")"/ge;
```
perlpunk: Not sure if everyone finds this more readable, but at least wanted to suggest it:
```suggestion
            s/\[([^\]]+)\]\(([\w-]+)\.md\)/[$1](#\L$2\E)/g;
```
But maybe this can be done in a followup.

perlpunk (approved):

perlpunk:
https://app.circleci.com/pipelines/github/os-autoinst/openQA/21415/workflows/163785c1-f8d2-4958-b7b9-3e3ca6ab1833/jobs/217400
```
 ::warning file=tools/process-docs,line=5,col=1,title=Version string used - severity 3::[ValuesAndExpressions::ProhibitVersionStrings] Use a real number instead
```
so that's about the `use v5.10;` line. I disagree with this perlcritic policy.

https://metacpan.org/pod/Perl::Critic::Policy::ValuesAndExpressions::ProhibitVersionStrings
> Whenever you use or require a module, you can specify a minimum version requirement. To ensure compatibility with older Perls, this version number should be expressed as a floating-point number.

So this is meant for older perls.

tools/process-docs
```diff
@@ -0,0 +1,21 @@
+#!/bin/perl
+
+use strict;
+use warnings;
+use v5.10;
+use experimental 'signatures';
+
+sub process_file ($file) {
+    open my $handle, "<", $file or die $!;
+    while (<$handle>) {
```
perlpunk: perlcritic wants this:
```suggestion
    while (my $line = <$handle>) {
```
or
```suggestion
    while (local $_ = <$handle>) {
```

PR #7489, opened 2026-05-29
feat(scheduling): Allow globbing in version specifications

Martchus:
There is already a fallback to search for product definitions with the version `*` in case no product definition with an exact version match was found. This change extends this fallback so no only `*` is checked for.

With this change it is possible to create a product definition with a version like `15.7:PR-*`. When then scheduling a new product with e.g. `VERSION=15.7:PR-0815` this product definition will be considered as fallback.

This will allow us to create jobs with e.g. `VERSION=15.7:PR-0815` instead of just `VERSION=15.7` so these jobs are not wrongly considered to be for consecutive builds.

This change keeps the sub test `fallback to version "*"` essentially as-is. It is only renamed to `a single "*" can be used to match all versions` under `version matching with "*"` and the expected not has been slightly adjusted. The sub test `no templates found for product` also still tests the same behavior and was only adjusted to no longer rely on the now restructured inner-workings of the previous sub test.

I tested this change also manually by creating a medium type (product definition) with version `15.7:PR-*` and by creating a new job group with the following job templates:

```
defaults:
  x86_64:
    machine: 64bit
    priority: 50
products:
  distri-x86_64:
    distri: distri
    flavor: flavor
    version: '15.7:PR-*'
scenarios:
  x86_64:
    distri-x86_64:
      - test-suite:
          settings:
            FOO: bar
```

Then I scheduled tests via:
```
script/openqa-cli api … -X POST isos DISTRI=distri FLAVOR=flavor VERSION=15.7:PR-1234 ARCH=x86_64
```

Without this change I ran into `no products found for distri-flavor-x86_64`. With this change one job was created as expected. Of course that new job had the setting `VERSION=15.7:PR-1234` and not just `VERSION=15.7`. So if we wanted to use this in osado we might need to adjust test code to deal with the suffix `:PR-1234`.

Related ticket: https://progress.opensuse.org/issues/200991

### DIFF
--- a/docs/UsersGuide.md
+++ b/docs/UsersGuide.md
@@ -122,21 +122,23 @@ Some example variables used by openSUSE are:
 - `PROMO` marks the promotional product.
 - `RESCUECD` is set to 1 for rescue CD images.
 
-The version of a medium can contain the `*` character. This enables a fallback
-lookup using globbing in case no medium with an exactly matching version is
-present when scheduling a product. This can be useful in the following cases:
+#### Using Wildcards in Medium Versions
+You can use the `*` character within a medium version to enable a fallback
+lookup using globbing. This is particularly helpful for the following scenarios:
 
-- One can avoid having individual medium types per version by specifying `*` as
-  medium type version. Note that having a medium with a concrete version and one
-  with `*` at the same time is usually not a good idea as you will likely run
-  into the scheduling error `no templates found for product …` (unless you
-  actually have job templates for all these medium types).
-- One can schedule products with e.g. `VERSION=15-SP7:PR-1234` if those jobs
-  are not supposed to be considered for consecutive builds (to avoid e.g. wrong
-  carry over and a cluttered "Next & Previous" table). To make this work one can
-  specify a version like `15-SP7:PR-*` in the medium type. This medium type will
-  then be used for all scheduled products where the `VERSION` starts with
-  `15-SP7:PR-`.
+- *Consolidating generic medium types:* Instead of creating a unique medium
+  type for every single version, you can use `*` as a catch-all version.
+    - Tip: Avoid defining a concrete version (e.g., `16.1`) and a wildcard
+      version (`*`) within one schedule definition unless you have matching job
+      templates for both, as this can trigger a scheduling error
+      (`no templates found for product...`).
+- *Handling isolated tests for individual submissions (e.g. pull-requests):*
+  When testing such submissions, schedule the products with e.g.,
+  `VERSION=16.1:PR-1234`. This will avoid cluttering the "Next & Previous" table
+  and prevent unintended carry-overs. Then specify a partial-match as the medium
+  version. For example by specifying `16.1:PR-*` in the medium type, openQA will
+  automatically match and use this type for any scheduled product where the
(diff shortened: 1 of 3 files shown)


okurz (changes requested):
Very nice. Can you please extend the documentation accordingly?

Martchus:
I've just tested the performance of the query. On o3 e.g. `select * from products where distri = 'opensuse' and '15.7:PR-0815' like replace(replace(replace(version, '%', '\%'), '_', '\_'), '*', '%') and flavor = 'Staging-DVD' and arch = 'x86_64';` returns the expected row with `version=*` in an instant. One OSD a similar query also returns in an instant. So that this is not very efficient is probably not going to be a problem.

I'll extend the documentation.

Martchus:
osado needs a change like https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/25684.

docs/UsersGuide.md
```diff
@@ -122,11 +122,21 @@ Some example variables used by openSUSE are:
...
+  medium type version. Note that having a medium with a concrete version and one
+  with `*` at the same time is usually not a good idea as you will likely run
+  into the scheduling error `no templates found for product …` (unless you
+  actually have job templates for all these medium types).
+- One can schedule products with e.g. `VERSION=15-SP7:PR-1234` if those jobs
+  are not supposed to be considered for consecutive builds (to avoid e.g. wrong
+  carry over and a cluttered "Next & Previous" table). To make this work one can
+  specify a version like `15-SP7:PR-*` in the medium type. This medium type will
+  then be used for all scheduled products where the `VERSION` starts with
+  `15-SP7:PR-`.
```
okurz: I would prefer a more user-facing, positive phrasing

```
### Using Wildcards (*) in Medium Versions

You can use the `*` character within a medium version to enable flexible fallback lookup using globbing. This is particularly helpful for the following scenarios:

- **Consolidating generic medium types:** Instead of creating a unique medium type for every single version, you can use `*` as a catch-all version. 
  *Tip: Avoid defining a concrete version (e.g., `16.1`) and a wildcard version (`*`) within one schedule definition unless you have matching job templates for both, as this can trigger a scheduling error (`no templates found for product...`).*
- **Handling isolated pull-request or branch builds:** When testing temporary variations (e.g., `VERSION=16.1:PR-1234`) specify a partial-match. For example by specifying `16.1:PR-*` in the medium type, openQA will automatically match and use this type for any product where the `VERSION` begins with `16.1:PR-`. ), This will ensure clean historical data for each branch and prevent unintended carry-overs.
```

okurz (changes requested):

docs/UsersGuide.md
Martchus: I see what makes sense from this clearly AI generated suggestion. "clean historical data" is definitely making not much sense, as well as "),".

okurz: > I see what makes sense from this clearly AI generated suggestion. "clean historical data" is definitely making not much sense, as well as "),".

please stop blaming the poor AI for my mistakes ;)

okurz (approved):

aylen384 (approved):

PR #7492, opened 2026-06-02
ci: Auto-import keys to avoid interactive prompt for devel_openQA

Martchus:
Related ticket: https://progress.opensuse.org/issues/201858

### DIFF
--- a/tools/ci/build-docs
+++ b/tools/ci/build-docs
@@ -2,7 +2,7 @@
 if [[ "${CIRCLE_JOB}" != *nightly ]]; then
     bash -xe tools/generate-documentation
 else
-    zypper -n install hub
+    zypper -n --gpg-auto-import-keys install hub
     export PULL_REQUEST_USER=${CIRCLE_PROJECT_USERNAME:-os-autoinst}
     export PUBLISH=1
     bash -xe tools/generate-documentation https://token@github.com/os-autoinst-bot/openQA.git gh-pages-"$(date +%y%m%d%H%M%S)"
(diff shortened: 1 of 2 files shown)


tools/ci/build-docs
```diff
@@ -2,7 +2,7 @@
 if [[ "${CIRCLE_JOB}" != *nightly ]]; then
     bash -xe tools/generate-documentation
 else
-    zypper -n install hub
+    zypper -n --gpg-auto-import-keys install hub
```
perlpunk, line 5: I'm wondering - can't this be done already when adding the repo to zypper?

okurz (approved):

tools/ci/build-docs
okurz: I think we already do: https://github.com/os-autoinst/openQA/blob/master/container/ci/Dockerfile#L12 . Apparently not enough. Our container might be a bit outdated. But regardless there can be the case when the container is created, the keys change and only then we use the container so I think it's correct that we need to import keys on use-time

perlpunk (approved):

tools/ci/build-docs
Martchus: Yes, we "unfortunately" already do which means that doing it when building the container doesn't seem to cover all cases. We already use `zypper -n --gpg-auto-import-keys` in some other places so that's probably also not the first time we run into this problem.

And by the way, the problem was also easily reproducible locally with podman and *not* using `--pull always`.

PR #7499, opened 2026-06-03
ci: Run OBS scm checks only for master branch

perlpunk:
Issue: https://progress.opensuse.org/issues/201978

I tried to test it in my fork, but somehow it complained about the secret, but this is the documented way to do it:
https://openbuildservice.org/help/manuals/obs-user-guide/cha-obs-scm-ci-workflow-integration#sec-obs-obs-scm-ci-workflow-integration-setup-obs-workflows-filters-delimiters

### DIFF
--- a/.obs/workflows.yml
+++ b/.obs/workflows.yml
@@ -33,6 +33,9 @@ pr:
 
   filters:
     event: pull_request
+    branches:
+      only:
+        - master
 
 # Setup:
 # 1. Put this .obs/workflows.yml in the main branch of openQA


Martchus (approved):

okurz (approved):

okurz:
According to the docs there is also "ignore", why didn't you pick that?

okurz (changes requested):
According to the docs there is also "ignore", why didn't you pick that?

Does this still trigger tests in PRs?

perlpunk:
> According to the docs there is also "ignore", why didn't you pick that?

Because we really only want it to run on PRs that have master as a target.
What other use case do you have in mind?

> Does this still trigger tests in PRs?

Yes. I can try harder to test it in my fork, but according to the docs this should be working.

okurz:
> > According to the docs there is also "ignore", why didn't you pick that?
> 
> Because we really only want it to run on PRs that have master as a target. What other use case do you have in mind?

As target, yes. If it's only about that then ok

> > Does this still trigger tests in PRs?
> 
> Yes. I can try harder to test it in my fork, but according to the docs this should be working.

Ok, let's try.

okurz (approved):

PR #7500, opened 2026-06-03
WIP just checking

perlpunk:
https://progress.opensuse.org/issues/201978


perlpunk:
Confirming that this is still running the OBS checks

PR #7506, opened 2026-06-05
feat(scheduler): distinguish busy vs offline workers

okurz:
Motivation:
Previously, unallocated jobs got the reason "no free workers for class
@needed" even if no online workers had that class. This made it hard to
diagnose if workers were busy or offline.

Design Choices:
- Introduced `determine_online_workers` to query the database once for all
  active workers.
- Refactored `determine_free_workers` to filter the active workers in memory.
- Updated `_allocate_jobs` signature and reason assignment to verify if a
  matching worker is online before claiming it is just busy.

Benefits:
- Users now get precise diagnostic feedback on whether workers are busy or
  completely offline.
- Performance remains optimal by avoiding multiple DB queries.

Screenshot:
Screenshot_20260709_201749_oqa_no_worker_details

### DIFF
--- a/lib/OpenQA/Scheduler/Model/Jobs.pm
+++ b/lib/OpenQA/Scheduler/Model/Jobs.pm
@@ -29,16 +29,15 @@ has scheduled_jobs => sub { {} };
 has shuffle_workers => 1;
 has dynamic_limit => sub { OpenQA::Scheduler::DynamicLimit->new };
 
-sub determine_free_workers ($shuffle = 0) {
-    my @free_workers = grep { !$_->dead } OpenQA::Schema->singleton->resultset('Workers')->search(
+sub determine_online_workers ($shuffle = 0) {
+    my @online_workers = grep { !$_->dead } OpenQA::Schema->singleton->resultset('Workers')->search(
         {
-            job_id => undef,
             error => undef,
             'properties.key' => 'WEBSOCKET_API_VERSION',
             'properties.value' => WEBSOCKET_API_VERSION
         },
         {join => 'properties'})->all;
-    return $shuffle ? shuffle(\@free_workers) : \@free_workers;
+    return $shuffle ? shuffle(\@online_workers) : \@online_workers;
 }
 
 sub determine_scheduled_jobs ($self) {
@@ -52,7 +51,8 @@ sub effective_job_limit ($self) {
     return $self->dynamic_limit->current_limit($config);
 }
 
-sub _allocate_jobs ($self, $free_workers) {
+sub _allocate_jobs ($self, $online_workers = undef) {
+    $online_workers //= determine_online_workers($self->shuffle_workers);
     my ($allocated_workers, $allocated_jobs) = ({}, {});
     my $scheduled_jobs = $self->scheduled_jobs;
     $scheduled_jobs->{$_}->{current_reason} = undef for keys %$scheduled_jobs;
@@ -76,10 +76,14 @@ sub _allocate_jobs ($self, $free_workers) {
     my %rejected;
     for my $id (keys %$scheduled_jobs) {
         my $jobinfo = $scheduled_jobs->{$id};
-        $jobinfo->{matching_workers} = _matching_workers($jobinfo, $free_workers, \%rejected);
+        my $has_matching_online = 0;
+        $jobinfo->{matching_workers} = _matching_workers($jobinfo, $online_workers, \%rejected, \$has_matching_online);
         if (!@{$jobinfo->{matching_workers}}) {
             my @needed = sort @{$jobinfo->{worker_classes}};
-            $jobinfo->{current_reason} = @needed ? "no free workers for class @needed" : 'no workers online';
+            $jobinfo->{current_reason}
+              = !@needed ? ($has_matching_online ? 'no free workers' : 'no workers online')
+              : $has_matching_online ? "no free workers for class @needed"
+              : "no workers online for requested class @needed at all";
         }
     }
     if (keys %rejected) {
@@ -99,6 +103,7 @@ sub _allocate_jobs ($self, $free_workers) {
 
     my @sorted = sort { $a->{priority} <=> $b->{priority} || $a->{id} <=> $b->{id} } values %$scheduled_jobs;
     my %checked_jobs;
+    my $free_workers = [grep { !$_->job_id } @$online_workers];
     for my $j (@sorted) {
         my $job_id = $j->{id};
         if ($checked_jobs{$job_id} || defined $allocated_jobs->{$job_id} || !@{$j->{matching_workers}}) {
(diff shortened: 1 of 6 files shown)


Wabri (approved):

lib/OpenQA/Scheduler/Model/Jobs.pm
```diff
@@ -29,16 +29,21 @@ has scheduled_jobs => sub { {} };
...
             'properties.value' => WEBSOCKET_API_VERSION
         },
         {join => 'properties'})->all;
-    return $shuffle ? shuffle(\@free_workers) : \@free_workers;
+    return $shuffle ? shuffle(\@online_workers) : \@online_workers;
+}
+
+sub determine_free_workers ($shuffle = 0) {
+    my $online_workers = determine_online_workers($shuffle);
+    my @free_workers = grep { !$_->job_id } @$online_workers;
```
Martchus: It would be more efficient to still let the DB do this filtering. Or, do the query and shuffling form `determine_online_workers` once, store the result and pass that as argument to `determine_free_workers` instead of redoing this work. The latter is probably the best solution.

lib/OpenQA/Scheduler/Model/Jobs.pm
```diff
@@ -79,7 +85,10 @@ sub _allocate_jobs ($self, $free_workers) {
         $jobinfo->{matching_workers} = _matching_workers($jobinfo, $free_workers, \%rejected);
         if (!@{$jobinfo->{matching_workers}}) {
             my @needed = sort @{$jobinfo->{worker_classes}};
-            $jobinfo->{current_reason} = @needed ? "no free workers for class @needed" : 'no workers online';
+            $jobinfo->{current_reason}
+              = !@needed ? 'no workers online'
```
Martchus: This reason was there before but it seems a bit odd. Shouldn't it be "job has no worker class"?

lib/OpenQA/Scheduler/Model/Jobs.pm
```diff
@@ -79,7 +85,10 @@ sub _allocate_jobs ($self, $free_workers) {
...
+              : @{_matching_workers($jobinfo, $online_workers)} ? "no free workers for class @needed"
```
Martchus: Going though all online workers again is a bit inefficient. We could at least return early after finding one matching online worker.

Martchus (changes requested):
I have some inline comments. However, I guess it would be best if we'd just do one round of `_matching_workers` on `$online_workers` and let `_matching_workers` deal with the `job_id`. So the filtering by `job_id` would be done within `_matching_workers` on the fly. `_matching_workers` could then directly return a boolean if there was a match regardless of `job_id`. This way we don't need two `determine_` functions and we don't need to call `_matching_workers` twice (and thus not go through all those workers twice).

lib/OpenQA/Scheduler/Model/Jobs.pm
okurz: Good catch. haven't thought about it when I just moved it. But I have added a new commit that clarifies this exact scheduling reason. If a job does not define the proper setting now its scheduling reason is explicitly set to "job has no worker class" instead of "no workers online". Tests in t/05-scheduler-full.t have also been updated accordingly.

lib/OpenQA/Scheduler/Model/Jobs.pm
okurz: Absolutely. I've refactored this to avoid the second call to _matching_workers entirely. Now tracks $has_matching_online inside the first pass of _matching_workers using a reference, allowing us to determine if any online workers match on the fly, I guess

lib/OpenQA/Scheduler/Model/Jobs.pm
okurz: Indeed. Originally I didn't know how but then I managed. determine_free_workers has been removed. We now query the DB exactly once in schedule via `determine_online_workers`, and then filter `$online_workers` in memory via `grep` to define `$free_workers`

lib/OpenQA/Scheduler/Model/Jobs.pm
```diff
@@ -218,15 +223,16 @@ sub _assign_jobs ($self, @args) {
...
+    my $free_workers = [grep { !$_->job_id } @$online_workers];
     my $worker_count = $schema->resultset('Workers')->count;
     my $free_worker_count = @$free_workers;
 
     my $scheduled_jobs = $self->determine_scheduled_jobs;
     log_debug("Scheduling: Free workers: $free_worker_count/$worker_count; Scheduled jobs: "
           . (scalar keys %$scheduled_jobs));
 
-    my ($allocated_workers, $allocated_jobs) = $self->_allocate_jobs($free_workers);
+    my ($allocated_workers, $allocated_jobs) = $self->_allocate_jobs($online_workers);
```
Martchus:
```suggestion
    my ($allocated_workers, $allocated_jobs) = $self->_allocate_jobs($online_workers, $free_workers);
```

Then avoid the duplicated computation `my $free_workers = [grep { !$_->job_id } @$online_workers];` in `_allocate_jobs`.

lib/OpenQA/Scheduler/Model/Jobs.pm
```diff
@@ -99,6 +103,7 @@ sub _allocate_jobs ($self, $free_workers) {
 
     my @sorted = sort { $a->{priority} <=> $b->{priority} || $a->{id} <=> $b->{id} } values %$scheduled_jobs;
     my %checked_jobs;
+    my $free_workers = [grep { !$_->job_id } @$online_workers];
```
Martchus: Even with my other suggestion we probably need:

```suggestion
    $free_workers //= [grep { !$_->job_id } @$online_workers];
```

So maybe it makes sense to extract `[grep { !$_->job_id } @$online_workers]` into a function that takes `$online_workers`, e.g. `determine_free_workers`.

Martchus (changes requested):

Martchus (approved):

Martchus (approved):

Martchus:
The coverage diff doesn't make much sense. Maybe it helps to retrigger tests.

perlpunk:
The coverage report is a bit weird:
https://output.circle-artifacts.com/output/job/101c773c-1397-417e-888e-aa56aa3131f7/artifacts/0/cover_db/t-full-stack-t.html
```
124 1 | for (; $elapsed <= $setup_timeout; $elapsed += sleep $setup_poll_interval) {
      |
125	1 | $free_workers = OpenQA::Scheduler::Model::Jobs::determine_free_workers(
      |
126	  |            OpenQA::Scheduler::Model::Jobs::determine_online_workers());
      |
127 0 | if (List::Util::any { $_->check_class($worker_class) } @$free_workers) {    # <-----------------
    1 |
      |
128	1 | pass "at least one free worker with class $worker_class registered";
      |
129	1 | return ($relevant_jobs, $elapsed);
      |
130	  | }
```
So because the List::Util line consists of two statements, in the HTML report we see two numbers, 0 and 1.
The weird thing is, if I try to reproduce that locally with an empty array, the first number is 1 and the second 0.
I would have to check if there is some change in the perl version.

And now the Codecov plugin turns that into a JSOn file with either `null`, 1 or 0 for each line.

I think in such a case the Codecov plugin should not take the 0, but the 1.
Since I'm the maintainer of the plugin, I will have a look :)

perlpunk:
OTOH it's also not really explainable with an empty array, because then the code inside the if block would also be uncovered...

okurz:
> I think in such a case the Codecov plugin should not take the 0, but the 1. Since I'm the maintainer of the plugin, I will have a look :)

Thanks. Anyway I pushed a commit to make the code more "Devel::Cover friendly". Maybe helps

EDIT: … maybe not :(

PR #7511, opened 2026-06-08
fix: Use reload-or-restart for openQA auto restart worker slots

gauravpathak:

Use reload-or-restart for openQA auto restart worker slots after openQA-worker package uninstallation step to avoid Unit cannot be reloaded because it is inactive error


okurz (approved):

Martchus (approved):

gauravpathak:
@Mergifyio queue

PR #7512, opened 2026-06-08
test: Avoid waiting on devel console if job is not running anymore

Martchus:
Related ticket: https://progress.opensuse.org/issues/201972


okurz (approved):

Martchus:
Annoyingly that's not possible because some fullstack tests fail with this. The developer mode tests passes but on 2nd though I'm also not sure whether it will now *reliably* pass. So it is probably better to avoid this optimization.

PR #7516, opened 2026-06-09
fix: move tests/.../livelog and liveterminal below /liveviewhandler

Vogtinator:
The apache and nginx configs have special treatment for /liveviewhandler that makes it more suitable for event sources, like disabled buffering. Moving those endpoints to this location fixes an indeterminate delay due to buffering.

### DIFF
--- a/lib/OpenQA/LiveHandler.pm
+++ b/lib/OpenQA/LiveHandler.pm
@@ -59,6 +59,8 @@ sub startup ($self) {
 
     # register route for live streaming of image
     $test_r->get('/streaming')->name('streaming')->to('running#streaming');
+    $test_r->get('/livelog')->name('streaming')->to('running#livelog');
+    $test_r->get('/liveterminal')->name('streaming')->to('running#liveterminal');
 
     OpenQA::Setup::setup_plain_exception_handler($self);
 }
--- a/container/webui/nginx.conf
+++ b/container/webui/nginx.conf
@@ -37,6 +37,7 @@ http {
             proxy_set_header Upgrade $http_upgrade;
             proxy_set_header Connection $connection_upgrade;
             proxy_set_header Host $host;
+            proxy_buffering off;
         }
 
         location /assets/ {
--- a/etc/nginx/vhosts.d/openqa-endpoints.inc
+++ b/etc/nginx/vhosts.d/openqa-endpoints.inc
@@ -53,6 +53,7 @@ location /liveviewhandler/ {
     proxy_set_header Upgrade $http_upgrade;
     proxy_set_header Connection "upgrade";
     proxy_set_header Host $host;
+    proxy_buffering off;
 }
 
 location / {
--- a/lib/OpenQA/WebAPI.pm
+++ b/lib/OpenQA/WebAPI.pm
@@ -167,8 +167,6 @@ sub startup ($self) {
     $test_r->get('/investigation_ajax')->name('test_investigation')->to('test#investigate');
     $test_r->get('/infopanel_ajax')->name('test_infopanel')->to('test#infopanel');
     $test_r->get('/status')->name('status')->to('running#status');
-    $test_r->get('/livelog')->name('livelog')->to('running#livelog');
-    $test_r->get('/liveterminal')->name('liveterminal')->to('running#liveterminal');
     $test_r->get('/edit')->name('edit_test')->to('running#edit');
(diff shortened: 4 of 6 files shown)


okurz (approved):

lib/OpenQA/LiveHandler.pm
```diff
@@ -59,6 +59,8 @@ sub startup ($self) {
 
     # register route for live streaming of image
     $test_r->get('/streaming')->name('streaming')->to('running#streaming');
+    $test_r->get('/livelog')->name('streaming')->to('running#livelog');
+    $test_r->get('/liveterminal')->name('streaming')->to('running#liveterminal');
```
Martchus, line 63: Don't move anything to `LiveHandler.pm` that doesn't need to be there. The separate live handler service only exists because requests handled by it need access to web socket connections maintained by this service - and to make this work the live handler service cannot use preforking. This doesn't scale so we should at least try to minimize the load in the live handler service.

You can still move the routes under `/liveviewhandler` keeping them in the main web API service.

Martchus (changes requested):

lib/OpenQA/LiveHandler.pm
Vogtinator: > You can still move the routes under /liveviewhandler keeping it in the main web API service.

That would need more complex routing in nginx again to route specific /liveviewhandler endpoints to webapi.

At that point you could just keep the route as is and disable buffering for `/tests/[0-9]+/(livelog|liveterminal)`.
Should I do that?

Martchus: Oh, it looks like the `/liveviewhandler` prefix is defined on NGINX-level and not on Mojolicious-level.

> At that point you could just keep the route as is and disable buffering for /tests/[0-9]+/(livelog|liveterminal).
> Should I do that?

Yes, that's probably better than adding anything to the livehandler service that doesn't need to be there.

Btw, where do we disable buffering for `/liveviewhandler` as you mention? I just see:

```
location /liveviewhandler/ {
    proxy_pass http://livehandler;
    proxy_http_version 1.1;
    proxy_read_timeout 3600;
    proxy_send_timeout 3600;
    proxy_set_header Upgrade $http_upgrade;
    proxy_set_header Connection "upgrade";
    proxy_set_header Host $host;
}
```

That's only enabling web sockets and has increased timeouts (compared to `/`) but no setting like `proxy_buffering off;`. It would probably make sense to actually add that setting there (and not *only* for `/tests/[0-9]+/(livelog|liveterminal)`).

Vogtinator: > Btw, where do we disable buffering for /liveviewhandler as you mention?

In the other commit in this PR.

Martchus: I've just checked on o3 where we have `proxy_buffering off;` in `/etc/nginx/vhosts.d/openqa-endpoints.inc`. So that's probably where you found it. It would make sense to upstream this configuration.

Vogtinator: > I've just checked on o3 where we have proxy_buffering off; in /etc/nginx/vhosts.d/openqa-endpoints.inc. So that's probably where you found it. It would make sense to upstream this configuration.

I added that with @aylen384 yesterday, the result was this PR.

aylen384: @Martchus See my comment on slack: https://suse.slack.com/archives/C02AJ1E568M/p1781020996330179

> Fabian and I finally found the problem with the livelog/liveterminal loading only after a few minutes if at all on o3.
> It was due to nginx buffering.
> We moved it to the liveviewhandler and set its route to unbuffered, which greatly improves responsiveness (even for the live screen itself).
https://github.com/os-autoinst/openQA/pull/7516
https://github.com/os-autoinst/openQA/pull/7517
When the PRs are merged, the current workaround in the nginx config (`location ~ (livelog|liveterminal)$` sections in `frontend.conf` on o3 and in `openqa-endpoints.inc` on both o3 and osd) need to be removed.

Vogtinator: Alternative nginx-config-only PR: https://github.com/os-autoinst/openQA/pull/7521

Compared to the workaround in production from yesterday I made the regex a bit stricter.

Vogtinator:
Closing in favor of https://github.com/os-autoinst/openQA/pull/7521

PR #7517, opened 2026-06-09
fix: avoid initial delay when replying from livestream

Vogtinator:
It waits until the image streaming timer (0.3s by default) has expired before replying, even if there's data available directly. Try it immediately as well.

### DIFF
--- a/lib/OpenQA/Shared/Controller/Running.pm
+++ b/lib/OpenQA/Shared/Controller/Running.pm
@@ -203,34 +203,35 @@ sub streaming ($self) {
     my $last_png = "$basepath/last.png";
     my $backend_run_file = "$basepath/backend.run";
     my $lastfile = '';
-    $timer_id = Mojo::IOLoop->recurring(
-        IMAGE_STREAMING_INTERVAL() => sub (@) {
-            my $newfile = readlink($last_png) || '';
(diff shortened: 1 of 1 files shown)


aylen384 (approved):

okurz (approved):

lib/OpenQA/Shared/Controller/Running.pm
```diff
@@ -203,34 +203,35 @@ sub streaming ($self) {
...
-            catch ($e) {
-                # log the error as server-sent events message and close the connection
-                # note: This should be good enough for debugging on the client-side. The client will re-attempt
-                #       streaming. Avoid logging on the server-side here to avoid flooding the log for this
-                #       non-critical problem.
-                chomp $e;
-                $self->write("data: Unable to read image: $e\n\n", $close_connection);
-            }
-        });
+    my $sendImage = sub (@) {
```
Martchus:
```suggestion
    my $send_image = sub (@) {
```

Martchus:
I re-triggered the fullstack test to see whether the failure is related.

lib/OpenQA/Shared/Controller/Running.pm
Vogtinator: Done

PR #7520, opened 2026-06-10
fix: prevent openqa-worker-auto-restart@* restart on uninstall

gauravpathak:

Previously, the uninstallation process would trigger a reload or restart of the openqa-worker-auto-restart@* services causing error openqa-worker-auto-restart@<n>.service is not active, cannot reload

This update ensures to not explicity reload-or-restart after the package is removed.


dist/rpm/openQA.spec
```diff
@@ -600,7 +600,6 @@ fi
 #        "$1 -ge 1" checks for a package upgrade
 if [ -x /usr/bin/systemctl ] && [ $1 -ge 1 ]; then
     /usr/bin/systemctl daemon-reload || :
-    /usr/bin/systemctl reload-or-restart 'openqa-worker-auto-restart@*.service' || :
```
Martchus, line 603: Won't this also prevent the reload on updates?

Martchus (changes requested):

dist/rpm/openQA.spec
gauravpathak: Maybe it will prevent the reload on updates. I created this PR just to test the behavior to see if it helps with the issue but it didn't. So, I'll close this PR.

gauravpathak: Or maybe I'll keep it for now to perform some more tests locally and close it later.

PR #7521, opened 2026-06-10
fix(livelog): Prevent timeouts due to nginx buffering

Vogtinator:
/liveviewhandler/... as well as /tests/.../livelog and .../liveterminal are endpoints for streaming events. Those need to be sent immediately without waiting for a full buffer or connection close.

Alternative to PR #7516

### DIFF
--- a/etc/nginx/vhosts.d/openqa-endpoints.inc
+++ b/etc/nginx/vhosts.d/openqa-endpoints.inc
@@ -53,6 +53,7 @@ location /liveviewhandler/ {
     proxy_set_header Upgrade $http_upgrade;
     proxy_set_header Connection "upgrade";
     proxy_set_header Host $host;
+    proxy_buffering off;
 }
 
 location / {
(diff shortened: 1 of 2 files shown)


etc/nginx/vhosts.d/openqa-endpoints.inc
```diff
@@ -66,3 +67,18 @@ location / {
     proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
     proxy_set_header X-Forwarded-Proto $scheme;
 }
+
+location ~ /tests/[0-9]+/(livelog|liveterminal)$ {
+    # Duplicate settings for location /, nginx can't merge locations :-(
```
Martchus, line 72: I guess the duplication is ok but you could move these settings into a separate file and include it via `include …`.

Martchus (approved):

etc/nginx/vhosts.d/openqa-endpoints.inc
Vogtinator: I could, but IMO that makes it even harder to follow what's going on...

On o3 it's already includes two layers deep.

okurz (approved):

PR #7523, opened 2026-06-10
ci: Log if TAP output check is disabled

Martchus:
A draft for debugging https://progress.opensuse.org/issues/200934


Martchus:
Ah, that's the problem:

> Skipping TAP output check as tests are running in verbose mode (t/16-utils-vcs-provider.t) ...

There's a `-v` in one of the test names and those get passed as CLI args. So the regex within `prove_wrapper` needs anchoring.

PR #7524, opened 2026-06-10
feat: Throttle jobs not using Git-URL as `CASEDIR`

Martchus:
* Throttle jobs that don't use a supported Git-URL as `CASEDIR` by default
* Allow throttling jobs based on any setting and regex pattern; this
  complements the existing throttling mechanism (which applies priority
  scaling for numerical settings)

Related ticket: https://progress.opensuse.org/issues/201276

### DIFF
--- a/docs/Installing.md
+++ b/docs/Installing.md
@@ -700,6 +700,20 @@ Its scale only applies to jobs having that value exceeding
 configuration, administrators must explicitly include the `MAX_JOB_TIME` rule,
 to retain runtime-based throttling.
 
+It is also possible to define throttling patterns:
+
+    prio_throttling_patterns = parameter_n:adjustment_n:=~regex_n[,…]
+    prio_throttling_patterns = parameter_n:adjustment_n:!~regex_n[,…]
+
+where:
+
+- `parameter_n`: the name of the parameter
+- `adjustment_n`: a positive or negative number that is added to the job
+  priority
+- `=~` or `!~`: the former applies `adjustment_n` if `regex_n` matches, the
+  latter applies `adjustment_n` if `regex_n` does not match
+- `regex_n`: a regex to match the value of `parameter_n` against
+
 The final priority of a job is the sum of a base priority, defined in the job
 templates or scenario definitions, plus all adjustments calculated for the
 configured parameters that match the job settings.
--- a/lib/OpenQA/Schema/ResultSet/Jobs.pm
+++ b/lib/OpenQA/Schema/ResultSet/Jobs.pm
@@ -212,6 +212,11 @@ sub _apply_max_job_time_prio ($factor, $time, $throt_config, $job_args) {
     return $info;
 }
 
+sub _change_prio_returning_sign ($new_job_args, $adjustment) {
+    $new_job_args->{priority} += $adjustment;
+    return $adjustment >= 0 ? '+' : '';
+}
+
 sub _apply_prio_throttling ($self, $settings, $new_job_args, $group = undef) {
     my $debug_msg;
     my $base_prio = $new_job_args->{priority} // 0;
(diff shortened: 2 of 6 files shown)


okurz (commented):
Doc changes are good

docs/Installing.md
```diff
@@ -700,6 +700,20 @@ Its scale only applies to jobs having that value exceeding
...
 
+It is also possible to define throttling patterns:
+
+    prio_throttling_patterns = parameter_n:adjustment_n:=~regex_n[,…]
+    prio_throttling_patterns = parameter_n:adjustment_n:!~regex_n[,…]
+
+where:
+
+- `parameter_n`: the name of the parameter
+- `adjustment_n`: a positive or negative number that is added to the job
```
kalikiana, line 711: Why is the existing scaling not good enough for `CASEDIR`? Do we have other cases where we need that kind of granularity?

I'm wondering because it seems overkill for the specific case the ticket is about.

kalikiana: And please also consider a more descriptive title for the PR. We already have throttling. What are we missing here and why?

kalikiana (changes requested):

okurz (changes requested):
"Throttle jobs that done"? Please state the important effect in the title. Think about users that want to read the changelog and wonder why their tests have a different priority after this change

docs/Installing.md
Martchus: > I'm wondering because it seems overkill for the specific case the ticket is about.

That's why I asked about this two days ago when I only proposed the documentation change. However, I think we should actually think a little bit bigger than just the ticket at hand. I mean, if someone comes up with yet another idea to throttle jobs based on settings it wouldn't be nice if we have to introduce yet another special feature for that. With this change it'll just be a configuration change.

Martchus: > Why is the existing scaling not good enough for `CASEDIR`? Do we have other cases where we need that kind of granularity?

The existing scaling is exactly that - *scaling*. It takes a value, e.g. `MAX_JOB_TIME`, and *scales* the priority based on that. Here we want something different: A simple check. If the contents of e.g. `CASEDIR` match a certain pattern we want to throttle those jobs. There's no scaling involved but instead we need to do pattern matching.

lib/OpenQA/Schema/ResultSet/Jobs.pm
```diff
@@ -218,20 +218,32 @@ sub _apply_prio_throttling ($self, $settings, $new_job_args, $group = undef) {
...
+                my $val = $value;
+                my $regex = $rule->{regex};
+                my $adj = $rule->{adjustment};
+                my $matches = $val =~ $regex;
+                if (($op eq '=~' && $matches) || ($op eq '!~' && !$matches)) {
+                    $new_job_args->{priority} += $adj;
+                    my $sign = $adj >= 0 ? '+' : '';
+                    push @throttling_info, sprintf "%s [%s%s: value %s %s %s]", $resource, $sign, $adj, $val, $op,
+                      $rule->{regex_str};
+                }
```
okurz, line 249: This seems to duplicate what further below is already done for job group properties. Can this be de-duplicated?

okurz (changes requested):

lib/OpenQA/Schema/ResultSet/Jobs.pm
Martchus: Both pieces of code do regex matching in a loop and push something to `@throttle_info`. But that's where the similarity ends. So I don't think these loops can be consolidated in a good way.

okurz (approved):

lib/OpenQA/Schema/ResultSet/Jobs.pm
kalikiana: Well, the code in the loop also looks quite similar to me:

    $new_job_args->{priority} += $rule->{increment};
    my $sign = $rule->{increment} >= 0 ? '+' : '';
    push @throttling_info, "$sign$rule->{increment} because job group $prop matches $rule->{regex}";

Plus it would be nice having the same phrasing here.

Martchus: I extracted the first two lines you mentioned in a function. Of course that means more lines of code in total but maybe still worth it.

It makes no sense to use the same phrasing because the code I'm introducing here has nothing to do with group properties. The code I'm introducing is closer to the scaling we already have and follows the style/format of this existing code. (It also does not match it exactly because it is slightly different due to being a slightly different feature.)

kalikiana (approved):

Martchus:
There's nothing missing. I created https://github.com/os-autoinst/openQA/pull/7547 to adjust the number of checks.

PR #7527, opened 2026-06-11
test: try to reproduce poo202209 problem

okurz:


Martchus:
The 16.0 build failed with:

```
[82754s] Bailout called.  Further testing stopped:  test 't/deploy.t' exceeds runtime limit of '200' seconds
```

The other builds also failed according to the GitHub status but on OBS they are actually building. Maybe they are retried by OBS itself after a certain build time has been exceeded. Maybe it is best to close this PR instead of wasting further build resources on OBS because these builds might go on forever and don't lead to any further insights.

PR #7535, opened 2026-06-12
docs: Mention how to enable UEFI and recently added variables

Martchus:
Related ticket: https://progress.opensuse.org/issues/200156


okurz (approved):

okurz:
Should be good enough

PR #7539, opened 2026-06-15
build: Let package builds for Tumbleweed and Leap>16 fail on any warning

Martchus:
* Depend on `rpmlint-strict` which makes many rpmlint checks fatal
* Note that this does not necessarily cover *all* warnings and errors

Related ticket: https://progress.opensuse.org/issues/201477

### DIFF
--- a/dist/rpm/openQA.spec
+++ b/dist/rpm/openQA.spec
@@ -155,6 +155,9 @@ Requires(pre):  group(nogroup)
 BuildRequires:  sysuser-tools
 %sysusers_requires
 %endif
+%if 0%{?is_opensuse} && 0%{?suse_version} >= 1600
+BuildRequires:  rpmlint-strict
+%endif
 
 %description
 openQA is a testing framework that allows you to test GUI applications on one


Martchus:
This is only one attempt at trying to let the build fail on any kind of warning or error.

The following does not work:

* Adding `setOption('BadnessThreshold', 0)` to rpmlintrc as this seems to be completely ignored by rpmlint 2 (where this config format is only still supported for compatibility).
* Adding `setBadness('.*', 100000)` to rpmlintrc. The regex is probably only supported for `addFilter`.
* Adding `BadnessThreshold = -1` or `BadnessThreshold = 0` to a new `rpmlint.toml` config file. The negative/zero value is probably internally discarded.
* Using the `--strict` flag of `rpmlint`. This actually does exactly what's being asked for in the ticket ACs but I haven't found a way to pass custom flags to `rpmlint` when building on OBS. OBS actually checks for a custom `rpmlint` binary under a custom location (see https://github.com/openSUSE/obs-build/blob/d588de479d7dd76ddf4af16e0094e288c9f7edcf/build-recipe-spec#L345). However, I'm not sure whether we can place an `rpmlint` wrapper there (e.g. by creating a custom `rpmlint-strict` package and adding it to `BuildRequires`) as it might conflict with how OBS itself adds its custom script there.

What kind of works:

* Adding `BadnessThreshold = 1`  to a new `rpmlint.toml` config file. It works for our openQA package in its current state but if there are only warnings of badness zero it probably won't work.
* The approach of this PR. It only covers checks that are on the list of the `rpmlint-strict` package, though. I should probably reword the commit message subject accordingly. With this our current openQA package still passes (but it only has very few warnings and errors that are probably really not a big deal).

It is not nice that `rpmlintrc` is currently only managed in the OBS repo but still referenced in the spec-file managed in Git. That's also why I cannot create a PR for the `BadnessThreshold = 1` version without adding the file on OBS first.

Note that what the ticket ACs ask for is also a bit questionable anyway. If we make it fail on any warning or error we'll probably sometimes run into the case that a new rpmlint version "breaks" our CI checks - which seems needlessly disruptive.

dist/rpm/openQA.spec
```diff
@@ -155,6 +155,9 @@ Requires(pre):  group(nogroup)
 BuildRequires:  sysuser-tools
 %sysusers_requires
 %endif
+%if 0%{?is_opensuse} && 0%{?suse_version} >= 1600
+Recommends:     rpmlint-strict
```
okurz: And this does have an effect on build time? I would have guessed recommends only affects runtime recommendations

Martchus: Right, that was wrong. In my local tests (with just the main `openQA` package) it nevertheless didn't make any difference but I'll change the PR.

okurz (approved):

PR #7540, opened 2026-06-16
feat: optionally fail ISO API if no jobs created

okurz:
Motivation:
Allow automation bots to strictly detect and fail when no jobs are scheduled for
a product, while keeping the failed record for investigation.

Design Choices:
Introduce an optional _FAIL_IF_NO_JOBS parameter that fails with 400 when no
jobs are created, while preserving the scheduled product record with error
details for later troubleshooting.

Benefits:
Automation bots can keep track of failures directly via the audit logs of
scheduled products.

Related issue: https://progress.opensuse.org/issues/194047

### DIFF
--- a/lib/OpenQA/Schema/Result/ScheduledProducts.pm
+++ b/lib/OpenQA/Schema/Result/ScheduledProducts.pm
@@ -26,6 +26,7 @@ use constant {
     ADDED => 'added',    # no jobs have been created yet
     SCHEDULING => 'scheduling',    # jobs are being created
     CANCELLING => 'cancelling',    # jobs are being cancelled (so far only possible as reaction to webhook event)
+    FATAL_ERROR => 'fatal_error',
 };
 
 __PACKAGE__->table('scheduled_products');
@@ -109,7 +110,7 @@ __PACKAGE__->inflate_column(
         deflate => sub { encode_json(shift) },
     });
 
-our @EXPORT_OK = qw(ADDED SCHEDULING SCHEDULED CANCELLING CANCELLED);
+our @EXPORT_OK = qw(ADDED SCHEDULING SCHEDULED CANCELLING CANCELLED FATAL_ERROR);
 our %EXPORT_TAGS = (STATE => [@EXPORT_OK]);
 
 sub sqlt_deploy_hook ($self, $sqlt_table, @) {
@@ -215,13 +216,14 @@ sub schedule_iso ($self, $args, $guard) {
 }
 
 sub set_done ($self, $result) {
+    my $status = ref $result eq 'HASH' ? (delete $result->{status} // SCHEDULED) : SCHEDULED;
     # set the status to be either …
     if ($self->_update_status_if(CANCELLED, status => CANCELLING)) {
         $self->update({results => $result});
         $self->cancel;    # … CANCELLED if meanwhile CANCELLING and invoke cancel again (as it backed off)
     }
     else {
-        $self->update({status => SCHEDULED, results => $result});    # … SCHEDULED if remained SCHEDULING
+        $self->update({status => $status, results => $result});    # … SCHEDULED if remained SCHEDULING
         $self->report_status_to_github;
     }
 }
@@ -333,6 +335,7 @@ sub _schedule_iso ($self, $args, $guard) {
 
     # read arguments for deprioritization and obsoleten
     my $deprioritize = delete $args->{_DEPRIORITIZEBUILD} // 0;
+    my $fail_if_no_jobs = delete $args->{_FAIL_IF_NO_JOBS} // 0;
     my $deprioritize_limit = delete $args->{_DEPRIORITIZE_LIMIT};
     my $obsolete = delete $args->{_OBSOLETE} // 0;
     my $onlysame = delete $args->{_ONLY_OBSOLETE_SAME_BUILD} // 0;
@@ -360,8 +363,10 @@ sub _schedule_iso ($self, $args, $guard) {
     else {
         $result = $self->_generate_jobs($args, \@notes, $skip_chained_deps, $include_children);
     }
-    return {error => $result->{error_message}, error_code => $result->{error_code} // HTTP_BAD_REQUEST}
-      if defined $result->{error_message};
+    if (defined $result->{error_message}) {
+        return $self->_fail_scheduling($result->{error_message}) if $fail_if_no_jobs;
+        return {error => $result->{error_message}, error_code => $result->{error_code} // HTTP_BAD_REQUEST};
+    }
     my $jobs = $result->{settings_result};
     # take some attributes from the first job to guess what old jobs to cancel
     # note: We should have distri object that decides which attributes are relevant here.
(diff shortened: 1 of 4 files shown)


lib/OpenQA/Schema/Result/ScheduledProducts.pm
```diff
@@ -109,7 +110,7 @@ __PACKAGE__->inflate_column(
         deflate => sub { encode_json(shift) },
     });
 
-our @EXPORT_OK = qw(ADDED SCHEDULING SCHEDULED CANCELLING CANCELLED);
+our @EXPORT_OK = qw(ADDED SCHEDULING SCHEDULED CANCELLING CANCELLED NO_JOBS_SCHEDULED_ERROR);
```
asmorodskyi: what "EXPORT_OK" array is holding ? "_OK" in name of variable contradicts with "_ERROR" in value for me at least

lib/OpenQA/Schema/Result/ScheduledProducts.pm
```diff
@@ -208,7 +209,12 @@ sub schedule_iso ($self, $args, $guard) {
         try { $self->_schedule_iso($args, $guard) }
         catch ($e) { {error => $e} }    # uncoverable statement
     };
-    $self->set_done($result);
+    if ($result->{delete_product}) {
```
asmorodskyi: what we are deleting ?

lib/OpenQA/Schema/Result/ScheduledProducts.pm
```diff
@@ -432,6 +444,14 @@ sub _schedule_iso ($self, $args, $guard) {
     return \%results;
 }
 
+sub _fail_scheduling ($self, $error, $error_code = HTTP_BAD_REQUEST) {
+    return {
+        error => $error,
+        error_code => $error_code,
+        delete_product => 1,
```
asmorodskyi: I don't think we want to delete scheduling product . it would be very useful for investigation why schedule haven't produced any jobs

lib/OpenQA/Schema/Result/ScheduledProducts.pm
asmorodskyi: nevermind , I figure out what

lib/OpenQA/Schema/Result/ScheduledProducts.pm
```diff
@@ -360,8 +367,10 @@ sub _schedule_iso ($self, $args, $guard) {
     else {
         $result = $self->_generate_jobs($args, \@notes, $skip_chained_deps, $include_children);
     }
-    return {error => $result->{error_message}, error_code => $result->{error_code} // HTTP_BAD_REQUEST}
-      if defined $result->{error_message};
+    if (defined $result->{error_message}) {
```
asmorodskyi, line 366: it took me a while to understand this piece . 

```
if (defined $result->{error_message}) {
        my $err_res = {error => $result->{error_message}, error_code => $result->{error_code} // HTTP_BAD_REQUEST};
        $err_res->{delete_product} = 1 if $fail_if_no_jobs;
        return $err_res;
   }
```

such version in my opinion would be easier to guess what is going on here for random reader of this code

asmorodskyi: but yeah above I am questioning overall need to delete product so this whole chunk might be just reverted ...

asmorodskyi (changes requested):
see comments inline

lib/OpenQA/Schema/Result/ScheduledProducts.pm
perlpunk: https://perldoc.perl.org/Exporter#Selecting-What-to-Export

lib/OpenQA/Schema/Result/ScheduledProducts.pm
okurz: but all the information is already present on the side of the message sender

lib/OpenQA/Schema/Result/ScheduledProducts.pm
asmorodskyi: ok , sorry my ignorance

lib/OpenQA/Schema/Result/ScheduledProducts.pm
asmorodskyi: if I would need to investigate such case I would really like to see proves on openQA side . this would allow me to exclude room for speculations about openQA API bug

okurz: ok, I removed that part. That means that the relevant change includes only the "fail if no jobs"

lib/OpenQA/Schema/Result/ScheduledProducts.pm
okurz: done

asmorodskyi: > done

what exactly is done ? now after whole point of this chunk is gone ( we don't need to add `delete => 1` to the hash when `$fail_if_no_jobs`  is True )  it should be simply reverted because there is no difference of what we returning in both branches . just in one case it happens from extra function call and in another directly

lib/OpenQA/Schema/Result/ScheduledProducts.pm
```diff
@@ -26,6 +26,7 @@ use constant {
     ADDED => 'added',    # no jobs have been created yet
     SCHEDULING => 'scheduling',    # jobs are being created
     CANCELLING => 'cancelling',    # jobs are being cancelled (so far only possible as reaction to webhook event)
+    NO_JOBS_SCHEDULED_ERROR => 'No jobs scheduled.',
```
Martchus: The constant value should not be a human-readable text that might be changed in the future. The concrete error message should probably just be part of the result data.

Maybe it also makes more sense to just have `ERROR`. Otherwise, if we add further error states in the future one needed to check for all error states to determine whether the scheduled product ran into an error. And maybe `FATAL_ERROR` also makes more sense because I think we can have non-fatal errors here and the fact that we have an error we want to treat as fatal is probably what's important here.

Martchus (changes requested):
Remove that from benefits:

> Automation bots can use strict error checking on the ISO post route

This only works if not using `async=1` but using that flag is highly recommended.

lib/OpenQA/Schema/Result/ScheduledProducts.pm
okurz: got it. Done

okurz:
> Remove that from benefits:
> 
> > Automation bots can use strict error checking on the ISO post route
> 
> This only works if not using `async=1` but using that flag is highly recommended.

done.

lib/OpenQA/WebAPI/Controller/API/V1/Iso.pm
```diff
@@ -158,6 +159,9 @@ Schedule jobs for assets matching the required settings DISTRI, VERSION, FLAVOR
 passed to the method as arguments. Returns a JSON block containing the number of jobs
 created, their job ids and the information for jobs that could not be scheduled.
 
+Supports an optional C<_FAIL_IF_NO_JOBS> parameter. When set to C<1>, the request fails with
+a C<400 Bad Request> if no jobs are scheduled.
```
Martchus: Not with `async=1`. It would be more important to mention that the scheduled product status will be set to `fatal_error`.

Martchus (changes requested):

Martchus (approved):

lib/OpenQA/Schema/Result/ScheduledProducts.pm
okurz: Not equivalent: `_fail_scheduling` adds `status => FATAL_ERROR`, which `set_done` uses to mark the product `fatal_error`; the fallback hash omits it and keeps the legacy `scheduled` status. Reverting would drop that and break t/api/02-iso.t + t/39-scheduled_products-table.t

lib/OpenQA/WebAPI/Controller/API/V1/Iso.pm
okurz: Addressed in the current revision — the POD now notes the `async=1` caveat and that the scheduled product status is set to `fatal_error`.

okurz:
Addressed all comments. Waiting for re-review from @asmorodskyi

asmorodskyi (approved):

PR #7554, opened 2026-06-19
tests: Make parallel execution of the fullstack test more reliable

Martchus:
Before there was a change of around 1.5625 % for choosing a conflicting worker instance number when running the developer and regular fullstack tests in parallel. This change does not fix the problem completely but now
the change is at least only 0.09765625 %.


Martchus:
Oh, I think this will lead to choosing completely invalid ports. I'll have to pick a lower number. Back to draft.

okurz (approved):

okurz:
Knowingly reducing the probability of the failure occuring sounds like you are already on the right track and that you should be and to fix it properly :) How about trying and on failure retrying?

perlpunk (approved):

PR #7556, opened 2026-06-19
fix(obs_rsync): use scope guards to avoid stalls

r-richardson:
When an openQA worker is restarted or dies, manual project locks in `obs_rsync_run` leak because the `unlock` call at the end of the script is never reached. This causes subsequent jobs for the same project to get stuck in a retry loop continually cycling between `active` and `inactive` states.

* Use a scope-based guard (`$helper->guard`) in `ObsRsync/Task.pm` to have Perl's garbage collector automatically release the database lock when the job (scope) exits for any reason.
* The `$job->note(project_lock => 1)` assignment is retained because the `Gru` queue controller explicitly relies on this metadata note to differentiate between jobs that are actively syncing versus those waiting in the queue.

Related Progress Issue: https://progress.opensuse.org/issues/202725

### DIFF
--- a/lib/OpenQA/WebAPI/Plugin/ObsRsync/Task.pm
+++ b/lib/OpenQA/WebAPI/Plugin/ObsRsync/Task.pm
@@ -24,7 +24,6 @@ sub _retry_or_finish ($job, $helper, $project = undef, $retry_interval = undef, 
     return $job->retry({delay => $retry_interval})
       if !$retry_max_count || $job->retries < $retry_max_count;
 
-    $helper->unlock($project) if $project;
     return $job->finish(
         {code => 2, message => "Exceeded retry count $retry_max_count. Consider job will be re-triggered later"});
 }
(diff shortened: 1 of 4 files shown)


lib/OpenQA/WebAPI/Plugin/ObsRsync/Task.pm
```diff
@@ -46,18 +46,18 @@ sub run ($job, $args) {
     my $helper = $app->obs_rsync;
     my $home = $helper->home;
 
-    if ($job->info && !$job->info->{notes}{project_lock}) {
-        return _retry_or_finish($job, $helper) unless $helper->lock($project);
-        $job->note(project_lock => 1);
-    }
+    return _retry_or_finish($job, $helper)
+      unless my $project_guard = $helper->guard($project);
+    $job->note(project_lock => 1) if $job->can('note');
```
Martchus: What kind of job might be passed here that does not have the `note` method? If the `if` is only there for tests to pass I suggest you rather extend the mocking in test code.

lib/OpenQA/WebAPI/Plugin/ObsRsync/Task.pm
```diff
@@ -46,18 +46,18 @@ sub run ($job, $args) {
     my $helper = $app->obs_rsync;
     my $home = $helper->home;
 
-    if ($job->info && !$job->info->{notes}{project_lock}) {
```
Martchus, line 49: You're removing the `$job->info && !$job->info->{notes}{project_lock}` check. I am not against that but I can't say whether that will break things or not.

lib/OpenQA/WebAPI/Plugin/ObsRsync/Task.pm
```diff
@@ -46,18 +46,18 @@ sub run ($job, $args) {
...
-        return _retry_or_finish($job, $helper) unless $helper->lock($project);
```
Martchus, line 50: I briefly checked where we call `lock` and after this removal we are *probably* not calling `lock` anymore anywhere. So it would probably make sense to remove it from `lib/OpenQA/WebAPI/Plugin/ObsRsync.pm`, as well as `unlock`.

Martchus (changes requested):

Martchus (approved):
The easiest way to find out whether this breaks anything is probably merging it :-)

okurz (approved):

okurz:
@r-richardson or
@mergifyio rebase

okurz (approved):

Martchus:
According to codecode line `sub info (@) { {notes => {project_lock => 1}} }` in `t/ui/27-plugin_obs_rsync_gru.t` is no longer covered. You can probably just remove that line.

PR #7557, opened 2026-06-19
refactor: Calculate module category headings in frontend

perlpunk:
This way it's easier to access each modules category.

Issue: https://progress.opensuse.org/issues/201840

### DIFF
--- a/assets/javascripts/render.js
+++ b/assets/javascripts/render.js
@@ -240,6 +240,7 @@ function batchProcess(items, processor, options = {}) {
   const {batchSize = DEFAULT_BATCH_SIZE, shouldContinue = () => true} = options;
   let currentIndex = 0;
 
+  let category = '';
   return new Promise((resolve, reject) => {
     function processNextBatch() {
       if (!shouldContinue()) {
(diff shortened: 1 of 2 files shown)


okurz (approved):

assets/javascripts/render.js
```diff
@@ -293,13 +294,14 @@ async function renderModuleTable(container, response, shouldContinue = () => tru
 
   return batchProcess(
     response.modules,
-    module => {
-      if (module.category) {
+    (module, i, catgory) => {
+      if (module.category != category) {
```
Martchus:
```suggestion
      if (module.category !== category) {
```

assets/javascripts/render.js
```diff
@@ -240,6 +240,7 @@ function batchProcess(items, processor, options = {}) {
   const {batchSize = DEFAULT_BATCH_SIZE, shouldContinue = () => true} = options;
   let currentIndex = 0;
 
+  category = '';
```
Martchus: This shouldn't be a global variable, right?

```suggestion
  let category = '';
```

Martchus (changes requested):

assets/javascripts/render.js
perlpunk: done

assets/javascripts/render.js
perlpunk: done

Martchus (approved):

assets/javascripts/render.js
perlpunk: hm, that made several tests fail... I have to debug this

perlpunk: found the problem. I mistyped category elsewhere...

PR #7563, opened 2026-06-23
fix: avoid foreign key error in screenshot_links

okurz:
Motivation
When a worker updates screenshots for a job that was recently deleted or
canceled, inserting into screenshot_links fails with a foreign key violation.

Design Choices
We add a pre-check to verify the job exists before proceeding. We also wrap
the populate call in a try/catch block from Feature::Compat::Try to handle
any race condition.

Benefits
This prevents unexpected database errors and logwarn alerts for normal user or
cleanup actions.

Related issue: https://progress.opensuse.org/issues/202605

### DIFF
--- a/lib/OpenQA/Schema/ResultSet/Screenshots.pm
+++ b/lib/OpenQA/Schema/ResultSet/Screenshots.pm
@@ -7,8 +7,9 @@ package OpenQA::Schema::ResultSet::Screenshots;
 use Mojo::Base 'DBIx::Class::ResultSet', -signatures;
 
 use Mojo::File qw(path);
-use OpenQA::Log qw(log_trace log_error);
+use OpenQA::Log qw(log_trace log_debug log_error);
 use OpenQA::Utils qw(imagesdir);
+use Feature::Compat::Try;
 
 sub create_screenshot ($self, $img) {
     my $dbh = $self->result_source->schema->storage->dbh;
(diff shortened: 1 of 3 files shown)


lib/OpenQA/Schema/ResultSet/Screenshots.pm
```diff
@@ -22,14 +23,21 @@ sub create_screenshot ($self, $img) {
...
     }
     my @data = map { [$_, $job_id] } values %ids;
-    $self->result_source->schema->resultset('ScreenshotLinks')->populate([[qw(screenshot_id job_id)], @data]);
+    try {
+        $self->result_source->schema->resultset('ScreenshotLinks')->populate([[qw(screenshot_id job_id)], @data]);
+    }
+    catch ($err) {
+        die $err unless $err =~ /screenshot_links_fk_job_id/;
+        log_debug "Job $job_id has been deleted, skipping screenshot links";
+    }
```
Martchus, line 50: The screenshot might be deleted as well. Maybe we should do the creation of the screenshot itself and the creation of the links in one transaction Or are we already doing that?

Maybe it is also more elegant to use raw SQL here.

lib/OpenQA/Schema/ResultSet/Screenshots.pm
```diff
@@ -22,14 +23,21 @@ sub create_screenshot ($self, $img) {
 
 # insert the symlinks into the DB
 sub populate_images_to_job ($self, $imgs, $job_id) {
+    return unless $self->result_source->schema->resultset('Jobs')->find($job_id);
```
Martchus: This check is not strictly needed as we need to handle this kind of error later on anyway again (although one can consider that check some kind of optimization for a probably very unlikely case).

lib/OpenQA/Schema/ResultSet/Screenshots.pm
```diff
@@ -22,14 +23,32 @@ sub create_screenshot ($self, $img) {
...
+                sub {
+                    my $res = $self->create_screenshot($img)->fetchrow_arrayref;
+                    my $screenshot_id = $res ? $res->[0] : $self->find({filename => $img})->id;
+                    $screenshot_links_rs->create({screenshot_id => $screenshot_id, job_id => $job_id});
+                });
+        }
+        catch ($err) {
+            if ($err =~ /screenshot_links_fk_job_id/) {
+                log_debug "Job $job_id has been deleted, skipping screenshot links";
+                return;
```
Martchus: This `return` is probably not needed and makes these branches inconsistent.

okurz: done

kalikiana (approved):

Martchus (approved):

perlpunk:
Picking my commit from #7634 with only 1 approval into a PR with already two approvals is a funny way to sneak in my commit :)

okurz:
> Picking my commit from #7634 with only 1 approval into a PR with already two approvals is a funny way to sneak in my commit :)

That's why it's possible to dismiss review automatically on updates. I honestly hadn't considered this caveat. I actually added your commit assuming the corresponding PR would be approved eventually and my PR could then automatically get merged

PR #7568, opened 2026-06-26
feat: Turn code locations in autoinst-log into links

perlpunk:
Issue: https://progress.opensuse.org/issues/197177

This will turn the file info in lines like

    [step:install,boot,5] tests/install/boot.pm:5 called utils::wait_for_desktop
      -> lib/utils.pm:48 called testapi::assert_screen

into links to the source on GitHub.

It will only do that if the filename is autoinst-log.txt, and if the page
has the special query parameter `sl=1`, as doing it by default might not
be wanted for long logfiles.

autoinst-log-with-source-links

### DIFF
--- a/assets/javascripts/test_result.js
+++ b/assets/javascripts/test_result.js
@@ -1,6 +1,8 @@
 // jshint multistr: true
 // jshint esversion: 6
 
+const testGitInfoRe = /TEST_GIT_HASH=([a-fA-F0-9]+) TEST_GIT_URL=([^\p{Cc}]+)/u;
+
 const testStatus = {
   state: null,
   result: null,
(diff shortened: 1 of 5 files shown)


okurz (approved):

assets/javascripts/test_result.js
```diff
@@ -669,6 +681,24 @@ function filterEmbeddedLogFiles() {
   loadEmbeddedLogFiles(filterLogLines.bind(null, searchBox, false));
 }
 
+const stepRe = / \[step:[a-zA-Z0-9-]+,[a-zA-Z0-9-]+,[0-9]+\] /;
+const sourceRe = /(?<= )([a-zA-Z0-9_/.-]+\.p[my]):(\d+)/g;
+function createSourceLinks(logFileElement, lineContentElement) {
+  const E = createElement;
+  const found = lineContentElement.innerHTML.match(stepRe);
+  if (found) {
+    lineContentElement.innerHTML = lineContentElement.innerHTML.replace(sourceRe, (match, filePath, lineNumber) => {
```
Martchus, line 690: Maybe it is possible to use `lineContentElement.querySelectorAll(…)` with https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Selectors/Attribute_selectors and replace those elements via the DOM `.replaceWith()` function instead of doing replacements in HTML code directly.

Martchus (approved):
I guess we can still improve the DOM handling later. (My suggestion is not very concrete and I see now that using Attribute selectors would probably not help.)

PR #7569, opened 2026-06-26
fix: allow concurrent archiving and result cleanup + avoid blocking result cleanup on stale locks

okurz:
* fix: allow concurrent archiving and result cleanup
* fix: avoid blocking result cleanup on stale locks

### DIFF
--- a/lib/OpenQA/Task/Job/ArchiveResults.pm
+++ b/lib/OpenQA/Task/Job/ArchiveResults.pm
@@ -16,11 +16,9 @@ sub _archive_results ($minion_job, @args) {
     my $app = $minion_job->app;
     return $minion_job->fail('No job ID specified.') unless defined $openqa_job_id;
 
-    # avoid archiving during result cleanup, avoid running too many cleanup/archiving jobs in parallel
+    # avoid running too many cleanup/archiving jobs in parallel
     return $minion_job->retry({delay => ONE_MINUTE})
       unless my $process_job_results_guard = $app->minion->guard('process_job_results_task', ONE_DAY, {limit => 5});
-    return $minion_job->retry({delay => ONE_MINUTE})
-      if $app->minion->is_locked('limit_results_and_logs_task');
 
     # avoid running any kind of result post processing task for a particular openQA job in parallel
     return $minion_job->retry({delay => 30})
--- a/lib/OpenQA/Task/Job/Limit.pm
+++ b/lib/OpenQA/Task/Job/Limit.pm
@@ -28,14 +28,10 @@ sub register ($self, $app, @args) {
 sub _limit ($job, $args = undef) {
     my $ensure_task_retry_on_termination_signal_guard = OpenQA::Task::SignalGuard->new($job);
 
-    # prevent multiple limit_results_and_logs tasks and limit_screenshots_task/archive_job_results to run in parallel
+    # prevent multiple limit_results_and_logs tasks to run in parallel
     my $app = $job->app;
-    return $job->retry({delay => ONE_MINUTE})
-      unless my $process_job_results_guard = $app->minion->guard('process_job_results_task', ONE_DAY);
     return $job->finish('Previous limit_results_and_logs job is still active')
       unless my $limit_results_and_logs_guard = $app->minion->guard('limit_results_and_logs_task', ONE_DAY);
-    return $job->finish('Previous limit_screenshots_task job is still active')
-      unless my $limit_screenshots_guard = $app->minion->guard('limit_screenshots_task', ONE_DAY);
 
     return undef unless my $limit_guard = acquire_limit_lock_or_retry($job);
 


lib/OpenQA/Task/Job/ArchiveResults.pm
```diff
@@ -16,11 +16,9 @@ sub _archive_results ($minion_job, @args) {
     my $app = $minion_job->app;
     return $minion_job->fail('No job ID specified.') unless defined $openqa_job_id;
 
-    # avoid archiving during result cleanup, avoid running too many cleanup/archiving jobs in parallel
+    # avoid running too many cleanup/archiving jobs in parallel
     return $minion_job->retry({delay => ONE_MINUTE})
       unless my $process_job_results_guard = $app->minion->guard('process_job_results_task', ONE_DAY, {limit => 5});
-    return $minion_job->retry({delay => ONE_MINUTE})
-      if $app->minion->is_locked('limit_results_and_logs_task');
```
perlpunk, line 23: That contradicts b9ff4a43c5240e4c4207c8059ba7dd7e3852318e

okurz: yes. But https://github.com/os-autoinst/openQA/commit/b9ff4a43c5240e4c4207c8059ba7dd7e3852318e merely says
> Archiving and the result cleanup can interfere with each other and are therefore prevented to run in parallel

Do you want me to reference this commit explicitly and explain better in commit messages or do you want to see actual code changes?

lib/OpenQA/Task/Job/Limit.pm
```diff
@@ -28,14 +28,10 @@ sub register ($self, $app, @args) {
...
 
-    # prevent multiple limit_results_and_logs tasks and limit_screenshots_task/archive_job_results to run in parallel
+    # prevent multiple limit_results_and_logs tasks to run in parallel
     my $app = $job->app;
-    return $job->retry({delay => ONE_MINUTE})
-      unless my $process_job_results_guard = $app->minion->guard('process_job_results_task', ONE_DAY);
     return $job->finish('Previous limit_results_and_logs job is still active')
       unless my $limit_results_and_logs_guard = $app->minion->guard('limit_results_and_logs_task', ONE_DAY);
-    return $job->finish('Previous limit_screenshots_task job is still active')
-      unless my $limit_screenshots_guard = $app->minion->guard('limit_screenshots_task', ONE_DAY);
```
Martchus, line 38: When I remember correctly, the `limit_screenshots_task` task is basically and extension of `limit_results_and_logs_task`. It makes most sense to keep this sequence - although probably no bad things will happen without.

lib/OpenQA/Task/Job/ArchiveResults.pm
Martchus: I would also like at least an explanation why

> Archiving and the result cleanup can interfere with each other and are therefore prevented to run in parallel¹.

is wrong.

Martchus (changes requested):
Please state the *concrete* high-level goal in one of the commit messages. The phrase "This prevents necessary cleanups and fills up the disk." reads rather vague.

Note that the locking was also to avoid putting the system under too much pressure which could impact the responsiveness of the web UI.

---

> can lead to deadlocks or stale locks

> De-coupling them avoids deadlocks

Locks are supposed to be acquired so that no deadlocks happen. Can you be more specific about where a deadlock can happen? Don't just say existing code is bad but give an explanation.

By the way, if there are problems with stale locks it would make most sense to improve that immediate problem instead of just removing *some* of the locks.

perlpunk:
> Locks are supposed to be acquired so that no deadlocks happen. Can you be more specific about where a deadlock can happen? Don't just say existing code is bad but give an explanation.
> 
> By the way, if there are problems with stale locks it would make most sense to improve that immediate problem instead of just removing _some_ of the locks.

Yes, I would also rather like to investigate that.
Acoording to @okurz he used the minion web UI functionality to send a TERM signal to the cleanup job instead of using the stop / remove functionality. Still, sending a TERM signal should actually lead to a retry, but instead the job was failed and locks were still there.

okurz:
> > Locks are supposed to be acquired so that no deadlocks happen. Can you be more specific about where a deadlock can happen? Don't just say existing code is bad but give an explanation.
> > By the way, if there are problems with stale locks it would make most sense to improve that immediate problem instead of just removing _some_ of the locks.
> 
> Yes, I would also rather like to investigate that. Acoording to @okurz he used the minion web UI functionality to send a TERM signal to the cleanup job instead of using the stop / remove functionality. 

I first used the "stop" functionality (not STOP signal) which had no visible effect within about 3m of waiting time. Then I used the minion web UI to send TERM.

Martchus:
Maybe this kind of stopping leads to stale locks. I don't see how only removing *some* of the locks fixes the problem, though. We also can't just remove *all* locks (which would fix the problem) so I suppose we have to solve this problem at the source.

okurz:
Closing due to reviewer feedback. My idea was that I would leave a decision to the filesystem and storage drives how to enqueue requests but if we actually have conceptual conflicts in how openQA treats archiving and deleting results then we should keep the locks

PR #7572, opened 2026-06-29
test: Enable faster re-connects in full scheduler test consistently

Martchus:
This rules out one possibility for the full scheduler being sometimes very slow. Unfortunately, this doesn't completely solve the issue in my local test environment.

I created this change while working on
https://progress.opensuse.org/issues/202803 but that's a different issue.


okurz (approved):
full-stack.t is different failure

Martchus:
https://progress.opensuse.org/issues/203298

aylen384 (approved):
I hope this doesn't create cpu spikes if the webui is offline

PR #7574, opened 2026-06-29
feat: Abort results cleanup early to prevent load surges

okurz:
Motivation:
As identified in #197123 we had 35k+ cleanup+archiving minion jobs due
to a delayed and longer running cleanup causing load surges.

Design Choices:
We evaluate df iteratively during the JobGroups loop using a new shared
utility method `is_disk_usage_below_percentage`. This allows the
task to abort results cleanup if enough space has already been freed.

Benefits:
This keeps results longer for users and prevents load surges on openQA
web UI hosts by avoiding enqueuing thousands of minion jobs needlessly.

Related issue: https://progress.opensuse.org/issues/197483

### DIFF
--- a/lib/OpenQA/Task/Job/Limit.pm
+++ b/lib/OpenQA/Task/Job/Limit.pm
@@ -8,7 +8,8 @@ use OpenQA::Jobs::Constants;
 use OpenQA::Log 'log_debug';
 use OpenQA::ScreenshotDeletion;
 use OpenQA::Utils qw(:DEFAULT resultdir archivedir check_df);
-use OpenQA::Task::Utils qw(acquire_limit_lock_or_retry finish_job_if_disk_usage_below_percentage);
+use OpenQA::Task::Utils
+  qw(acquire_limit_lock_or_retry finish_job_if_disk_usage_below_percentage is_disk_usage_below_percentage);
 use OpenQA::Task::SignalGuard;
 use Scalar::Util 'looks_like_number';
 use List::Util 'min';
@@ -55,6 +56,16 @@ sub _limit ($job, $args = undef) {
     my $gru = $app->gru;
     my %options = (priority => -20, ttl => 2 * ONE_DAY);
     while (my $group = $groups->next) {
+        if (
+            my $msg = is_disk_usage_below_percentage(
+                job => $job,
+                setting => 'result_cleanup_max_free_percentage',
+                dir => resultdir
+            ))
+        {
+            $job->note(early_abort_results => $msg);
+            last;
+        }
         my @preserved_important_jobs;
         $group->limit_results_and_logs(\@preserved_important_jobs);
 
--- a/lib/OpenQA/Task/Job/Limit.pm
+++ b/lib/OpenQA/Task/Job/Limit.pm
@@ -5,7 +5,7 @@ package OpenQA::Task::Job::Limit;
 use Mojo::Base 'Mojolicious::Plugin', -signatures;
 
 use OpenQA::Jobs::Constants;
-use OpenQA::Log 'log_debug';
+use OpenQA::Log qw(log_debug log_info);
 use OpenQA::ScreenshotDeletion;
 use OpenQA::Utils qw(:DEFAULT resultdir archivedir check_df);
 use OpenQA::Task::Utils
@@ -63,6 +63,7 @@ sub _limit ($job, $args = undef) {
                 dir => resultdir
             ))
         {
+            log_info "Early abort during job groups loop: $msg";
             $job->note(early_abort_results => $msg);
             last;
         }
--- a/t/42-df-based-cleanup.t
+++ b/t/42-df-based-cleanup.t
@@ -62,14 +62,19 @@ $mock_df->();
 subtest 'abort early if there is enough free disk space' => sub {
 
     $app->config->{misc_limits}->{result_cleanup_max_free_percentage} = 10;
-    my $job = run_gru_job($app, limit_results_and_logs => []);
+    my $job;
+    combined_like { $job = run_gru_job($app, limit_results_and_logs => []) }
+    qr/Skipping, free disk space on '.*' exceeds configured percentage 10 % \(free percentage: 11 %\)/,
+      'result cleanup early abort logged';
     is $job->{state}, 'finished', 'result cleanup still considered successful';
     like $job->{result},
       qr|Skipping.*/openqa/testresults.*exceeds configured percentage 10 % \(free percentage: 11 %\)|,
       'result cleanup aborted early';
 
     $app->config->{misc_limits}->{asset_cleanup_max_free_percentage} = 9;
-    $job = run_gru_job($app, limit_assets => []);
+    combined_like { $job = run_gru_job($app, limit_assets => []) }
+    qr/Skipping, free disk space on '.*' exceeds configured percentage 9 % \(free percentage: 11 %\)/,
+      'asset cleanup early abort logged';
     is $job->{state}, 'finished', 'asset cleanup still considered successful';
     like $job->{result},
       qr|Skipping.*/openqa/share/factory.*exceeds configured percentage 9 % \(free percentage: 11 %\)|,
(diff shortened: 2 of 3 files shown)


lib/OpenQA/Task/Job/Limit.pm
```diff
@@ -55,6 +56,16 @@ sub _limit ($job, $args = undef) {
     my $gru = $app->gru;
     my %options = (priority => -20, ttl => 2 * ONE_DAY);
     while (my $group = $groups->next) {
```
okurz, line 58: I have opted to only lookup the storage usage within this loop. We could theoretically push the evaluation deeper which would mix how we look up cleanup candidates from the database together with calling `df` which would not be so nice though.

aylen384 (approved):

t/42-df-based-cleanup.t
```diff
@@ -88,6 +88,34 @@ subtest 'abort early if there is enough free disk space' => sub {
...
+    $df_mock->redefine(
+        df => sub ($dir, @) {
+            $df_call_count++;
+            if ($df_call_count == 1) {
+                return {bavail => 5, blocks => 100};    # below 10%
+            }
+            else {
+                return {bavail => 20, blocks => 100};    # above 10%
+            }
+        });
```
Martchus: We shouldn't have a branch for each invocation (even if there are just two branches right now).

```suggestion
    my $df_call_count = 0;
    my @bavail = (5);
    $df_mock->redefine(df => sub ($dir, @) { {bavail => $bavail[$df_call_count++] // 20, blocks => 100} });
```

Or even:
```suggestion
    my $df_call_count = 0;
    my @bavail = (5, 20);
    $df_mock->redefine(df => sub ($dir, @) { {bavail => $bavail[$df_call_count++] // fail '…', blocks => 100} });
```

Maybe it makes also sense to be exactly below/above the 10 % threshold to avoid one-off errors.

And btw, if I wrote needlessly if-else-heavy code like this you often objected. Apparently you don't care if it is your code? :-)

t/42-df-based-cleanup.t
```diff
@@ -88,6 +88,34 @@ subtest 'abort early if there is enough free disk space' => sub {
     $mock_df->();
     ok !finish_job_if_disk_usage_below_percentage(@check_args), 'cleanup done if not enough free disk space available';
     is $job->{state}, undef, 'job not finished when proceeding with cleanup';
+
+    # Now test abort early DURING the loop
+    # We will simulate `df` returning low space on the first call, and high space on the second call.
```
Martchus: No need to capitalize "during". This should probably be a sub test.

Martchus (changes requested):
Looks generally good. I only have a few style nitpicks. Oh, and don't ever ask me again why some job was not cleaned up or archived after we've merged this.

t/42-df-based-cleanup.t
```diff
@@ -88,6 +88,34 @@ subtest 'abort early if there is enough free disk space' => sub {
...
+    # We need to make sure there are JobGroups to iterate over
+    # There is already a `JobGroup` created below, but maybe we can just run the minion job and check the note.
+    # To check the note, we must retrieve the minion job from the db.
+    my $group_bar = $app->schema->resultset('JobGroups')->create({name => 'bar'});
+    my $loop_job = run_gru_job($app, limit_results_and_logs => []);
+    is $loop_job->{state}, 'finished', 'result cleanup still considered successful';
+    like $loop_job->{notes}->{early_abort_results},
+      qr|Skipping.*/openqa/testresults.*exceeds configured percentage 10 % \(free percentage: 20 %\)|,
+      'result cleanup aborted early inside the loop';
+    $group_bar->delete;
```
kalikiana, line 114: We should probably also test if the next run will clean the previously skipped results?

kalikiana:
> Oh, and don't ever ask me again why some job was not cleaned up or archived after we've merged this.

Not sure what's with the cynicism here, but we have log messages we can check in that case. If that's not enough and we actually want this to be discoverable, we need openQA job details to show cleanup status of jobs.

t/42-df-based-cleanup.t
Martchus: With this design the runs are still independent. It is not like aborting early means the next run will continue where the previous run aborted. So unless the next run has to clean up more it also won't.

By the way, this design also means that the first job groups in the iteration are always the ones that are cleanup up and only the last job groups in the iteration benefit from having their results stored longer. That is not really fair. I suppose this could be easily solved by making sure that we iterate over the job groups in a truly random order.

Martchus:
> > Oh, and don't ever ask me again why some job was not cleaned up or archived after we've merged this.
> 
> Not sure what's with the cynicism here, but we have log messages we can check in that case. If that's not enough and we actually want this to be discoverable, we need openQA job details to show cleanup status of jobs.

It is because with this approach we more or less randomly (but probably consistently) exempt certain job groups from cleanup and archiving. So I can already tell that someone will ask me: Why is this old job not archived? Then I could of course check the logs but I'm getting tired of answering questions like that and with this change questions like that might come up more frequently.

perlpunk:
I'm wondering if this saves us anything. If we abort earlier, then used disk space won't be as low as it would be otherwise, so in the next run we would have to delete more. Isn't it just shifting the work a bit into the future?

okurz:
> I'm wondering if this saves us anything. If we abort earlier, then used disk space won't be as low as it would be otherwise, so in the next run we would have to delete more. Isn't it just shifting the work a bit into the future?

Not necessarily. Any next run might also just abort early if there is enough free space. And a lot of things can change until "the next run" or whenever we go below the triggering threshold again. maybe the server gains more free space because other data was removed or more storage was added or other non-important jobs can be removed rather than removing important jobs and possibly many more things.

t/42-df-based-cleanup.t
okurz: I partially addressed this in #7590 and I plan on top to ensure "fair" cleanup with a stateful tracking in #7591

t/42-df-based-cleanup.t
okurz: done

t/42-df-based-cleanup.t
okurz: you are right. I didn't initially understand how I could condense the code.

Done

okurz:
> Looks generally good. I only have a few style nitpicks. Oh, and don't ever ask me again why some job was not cleaned up or archived after we've merged this.

I added a change on top https://github.com/os-autoinst/openQA/pull/7574/changes/6c706b686c46e84b5b60cdce32c021170faf4bc0 to introduce some informative log messages. Maybe then everyone has a better chance to understand themselves what's going on.

t/42-df-based-cleanup.t
```diff
@@ -90,6 +95,30 @@ subtest 'abort early if there is enough free disk space' => sub {
...
+subtest 'abort early during the loop' => sub {
+    # We will simulate `df` returning low space on the first call, and high space on the second call.
+    my $df_call_count = 0;
+    my @bavail = (5);
+    $df_mock->redefine(df => sub ($dir, @) { {bavail => $bavail[$df_call_count++] // 20, blocks => 100} });
+    $app->config->{misc_limits}->{result_cleanup_max_free_percentage} = 10;
+
+    # We need to make sure there are JobGroups to iterate over
+    # There is already a `JobGroup` created below, but maybe we can just run the minion job and check the note.
+    # To check the note, we must retrieve the minion job from the db.
```
Martchus:
```suggestion
```

None of these comments make sense:

* The first two lines are not really clear so it would be better to just let the code speak for itself.
* If we added comments like the last line everywhere, our code would be very cluttered.

Martchus (changes requested):

t/42-df-based-cleanup.t
okurz: ok, removed those lines

okurz:
addressed the last open review suggestion https://github.com/os-autoinst/openQA/pull/7574/changes/fbb1f527aaca798e44fb35f4a0b7f3a9e9ce2a1c#pullrequestreview-4643527948

Martchus (approved):

PR #7582, opened 2026-07-01
ci: Exclude dependabot commits from gitlint

Martchus:
These commits are auto-generated and otherwise often run into "Line exceeds max length".

### DIFF
--- a/external/os-autoinst-common/.gitlint
+++ b/external/os-autoinst-common/.gitlint
@@ -12,3 +12,8 @@ regex=^https?:\/\/\S+$
 # https://progress.opensuse.org/issues/198254
 regex=^git subrepo pull
 ignore=all
+
+[ignore-by-author-name]
+# Match commits by author name (e.g. ignore dependabot commits)
+regex=dependabot
+ignore=all


Martchus:
Btw, this is directly from the [official docs](https://jorisroovers.com/gitlint/latest/ignoring_commits/) so I guess it is common practice to exclude these commits.

okurz (changes requested):
In qem-bot+qem-dashboard so far we have
https://github.com/openSUSE/qem-bot/blob/master/.gitlint#L10
I would prefer to stay consistent. If you see your change as superior I don't mind but then please include it in https://github.com/os-autoinst/os-autoinst-common/ (you are changing data in "external/") and apply the same in qem-bot+qem-dashboard or ask me to do it

external/os-autoinst-common/.gitlint
```diff
@@ -12,3 +12,8 @@ regex=^https?:\/\/\S+$
 # https://progress.opensuse.org/issues/198254
 regex=^git subrepo pull
 ignore=all
+
```
perlpunk, line 15: Hm, the checklist plugin should have warned about directly modifying files under `external/`

Martchus:
Ok, I'll go though the common repo.

EDIT: https://github.com/os-autoinst/os-autoinst-common/pull/96

external/os-autoinst-common/.gitlint
okurz: I wondered about the same. https://github.com/os-autoinst/openQA/actions/runs/28516968009/job/84531011863#step:3:14 says
> No paths were modified that match checklist paths

perlpunk: ah: https://github.com/wyozi/contextual-qa-checklist-action#include-hidden-files

perlpunk: * #7583

PR #7586, opened 2026-07-01
feat: throw 404 for nonexistent groups in overview

okurz:
Motivation
When a user specifies a non-existent group ID/name in the test overview
endpoint, openQA silently ignores the parameter and returns all test
results. This causes massive performance overhead and returns confusing jobs.

Design Choices
Check if explicit group parameters are specified but resolve to an empty
list of groups, and respond with a 404 error using the appropriate format.
Update callers to abort gracefully when the search parameters are undefined.

Benefits
This prevents unnecessary database load and provides clear and expected
error responses to users and API clients for non-existent job groups.

### DIFF
--- a/lib/OpenQA/WebAPI/Plugin/Helpers.pm
+++ b/lib/OpenQA/WebAPI/Plugin/Helpers.pm
@@ -408,6 +408,13 @@ sub _compose_job_overview_search_args ($c) {
         my @group_name_search = map { {name => $_} } @{$v->every_param('group')};
         my @search_terms = (@group_id_search, @group_name_search);
         @groups = $schema->resultset('JobGroups')->search(\@search_terms)->all;
+        if (!@groups) {
+            $c->stash(error_message => 'Group does not exist');
+            $c->respond_to(
+                json => {json => {error => 'Group does not exist'}, status => 400},
+                any => sub { $c->render(template => 'main/specific_not_found', status => 400) });
+            return undef;
+        }
     }
 
     else {
--- a/lib/OpenQA/WebAPI/Controller/API/V1/Job.pm
+++ b/lib/OpenQA/WebAPI/Controller/API/V1/Job.pm
@@ -228,6 +228,7 @@ So this works in the same way as the test results overview in the GUI.
 
 sub overview ($self) {
     my ($search_args, $groups) = $self->compose_job_overview_search_args;
+    return undef unless $search_args;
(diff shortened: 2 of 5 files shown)


pdostal (approved):
Awesome, thank you Oliver!

pdostal:
It has been reported [here](https://suse.slack.com/archives/C02CANHLANP/p1782226021262609).

lib/OpenQA/WebAPI/Plugin/Helpers.pm
```diff
@@ -408,6 +408,12 @@ sub _compose_job_overview_search_args ($c) {
         my @group_name_search = map { {name => $_} } @{$v->every_param('group')};
         my @search_terms = (@group_id_search, @group_name_search);
         @groups = $schema->resultset('JobGroups')->search(\@search_terms)->all;
+        if (!@groups) {
+            $c->respond_to(
+                json => {json => {error => 'Group does not exist'}, status => 404},
```
grisu48: I suggest to use status code [400 Bad Request](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Status/400) instead. 404 is not wrong, as the requested resource doesn't exist, but a 404 error makes the user rather believe that the used API endpoint does not exist. A 400 error is more generic, but gives the user a hint that they used the right API endpoint and the request was ill-formated.

This is just a suggestion. Both status codes (400 and 404) are valid, I just think that a 400 status code leads to less confusion, as it tells the end user "this is the right path, just your request is invalid due to some circumstances" (which are then described in the error message).

perlpunk (approved):

Martchus (approved):

lib/OpenQA/WebAPI/Plugin/Helpers.pm
okurz: ok, done

okurz:
I incorporated the suggestion from https://github.com/os-autoinst/openQA/pull/7586#discussion_r3511380362 so now returning error code 400 instead of 404

lib/OpenQA/WebAPI/Plugin/Helpers.pm
perlpunk: Hm, I'd chose 404, that's what we also return for `/api/v1/job_groups/$id` if group `$id` does not exist

PR #7589, opened 2026-07-02
feat: Retry git_clone minion job with exponential backoff

perlpunk:
Issue: https://progress.opensuse.org/issues/202881

### DIFF
--- a/t/14-grutasks-git.t
+++ b/t/14-grutasks-git.t
@@ -181,13 +181,23 @@ subtest 'git clone' => sub {
     };
 
     subtest 'git clone retried on failure' => sub {
-        $ENV{OPENQA_GIT_CLONE_RETRIES} = 1;
+        $ENV{OPENQA_GIT_CLONE_RETRIES} = 3;
         my $openqa_clone = Test::MockModule->new('OpenQA::Task::Git::Clone');
         $openqa_clone->redefine(_git_clone => sub (@) { die "fake error\n" });
+
         $res = run_gru_job(@gru_args);
         is $res->{retries}, 1, 'job retries incremented';
         is $res->{state}, 'inactive', 'job set back to inactive';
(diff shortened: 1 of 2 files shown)


t/14-grutasks-git.t
```diff
@@ -181,13 +181,23 @@ subtest 'git clone' => sub {
...
     subtest 'git clone retried on failure' => sub {
-        $ENV{OPENQA_GIT_CLONE_RETRIES} = 1;
+        $ENV{OPENQA_GIT_CLONE_RETRIES} = 3;
         my $openqa_clone = Test::MockModule->new('OpenQA::Task::Git::Clone');
         $openqa_clone->redefine(_git_clone => sub (@) { die "fake error\n" });
+
         $res = run_gru_job(@gru_args);
         is $res->{retries}, 1, 'job retries incremented';
         is $res->{state}, 'inactive', 'job set back to inactive';
+        is +($res->{delayed} - $res->{retried}), 2, 'delay as expected';
```
kalikiana, line 191:
```suggestion
        is +($res->{delayed} - $res->{retried}), 2, 'delayed as expected';
```

kalikiana (approved):

t/14-grutasks-git.t
Wabri: thought: It should be a request change if you suggest a change

Wabri (changes requested):
See Liv suggestion

t/14-grutasks-git.t
perlpunk: It's short for "The delay is as expected". And if that should be changed, then the next test label should also be changed, otherwise it's inconsistent.

okurz (approved):

Wabri (approved):

t/14-grutasks-git.t
Wabri: Ok, for me!

PR #7590, opened 2026-07-03
PART 2: feat: Prioritize cleanup of busiest job groups - After #7574

okurz:
Motivation:
Iterating over job groups in default DB order creates an unfair bias, always
cleaning up the same groups first while newer groups escape early aborts.

Design Choices:
Query and sort JobGroups by 'COUNT(jobs.id) DESC' to dynamically process the
most active groups first, which likely consume the most disk space.

Benefits:
Ensures a fair distribution of cleanup operations and maximizes disk space
reclaimed early to prevent load surges during high-concurrency periods.

Related issue: https://progress.opensuse.org/issues/197483

After:
* #7574

### DIFF
--- a/lib/OpenQA/Task/Job/Limit.pm
+++ b/lib/OpenQA/Task/Job/Limit.pm
@@ -8,7 +8,8 @@ use OpenQA::Jobs::Constants;
 use OpenQA::Log 'log_debug';
 use OpenQA::ScreenshotDeletion;
 use OpenQA::Utils qw(:DEFAULT resultdir archivedir check_df);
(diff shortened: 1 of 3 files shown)


Martchus:
This is as biased as the ordering we had before - the bias is just a different one. So either strike wording like "creates an unfair bias … Ensures a fair distribution" from the commit message or make it actually unbiased by selecting job groups randomly.

Martchus (changes requested):
Directly go for https://github.com/os-autoinst/openQA/pull/7591 which makes more sense.

okurz:
Closed in favor of #7591

PR #7597, opened 2026-07-06
fix: ensure podman cache dir is created by code+RPM

okurz:
* fix(packaging): create podman cache dir via RPM
* fix(worker): ensure podman cache directory exists


kalikiana (approved):
Good catch!

Martchus (approved):

PR #7602, opened 2026-07-07
feat: support global instance-wide test settings

okurz:
Motivation:
Allow defining default, instance-wide test settings in openqa.ini to easily
manage global defaults across all jobs.

Design Choices:
Added a test_settings section in Setup.pm and extracted those keys with
lowest precedence when generating job settings.

Benefits:
Provides a clean, centralized way to set baseline test parameters that can
be overridden by products, machines, test suites, or API POST requests.

Related issue: https://progress.opensuse.org/issues/203688

### DIFF
--- a/docs/UsersGuide.md
+++ b/docs/UsersGuide.md
@@ -263,15 +263,10 @@ a test suite and a product that appear in the same job template. The precedence
 order for variables is as follows (from lowest to highest):
 
 - Instance-wide test settings (defined in the `[test_settings]` section of `openqa.ini`)
-
 - Product
-
 - Machine
-
 - Test suite
-
 - Job template
-
 - API POST query parameters
 
 That is, variable values set as part of the API request that triggers the jobs will
--- a/docs/UsersGuide.md
+++ b/docs/UsersGuide.md
@@ -262,6 +262,8 @@ used for a single job - for instance, you may have a variable defined in both
 a test suite and a product that appear in the same job template. The precedence
 order for variables is as follows (from lowest to highest):
 
+- Instance-wide test settings (defined in the `[test_settings]` section of `openqa.ini`)
+
 - Product
 
 - Machine
--- a/lib/OpenQA/JobSettings.pm
+++ b/lib/OpenQA/JobSettings.pm
@@ -8,6 +8,7 @@ use Mojo::Base -strict, -signatures;
 use File::Basename;
 use Mojo::URL;
 use Mojo::Util 'url_unescape';
+use OpenQA::App;
 use OpenQA::Log 'log_debug';
 use OpenQA::Utils qw(asset_type_from_setting get_url_short);
 use Feature::Compat::Try;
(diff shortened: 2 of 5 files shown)


docs/UsersGuide.md
```diff
@@ -262,6 +262,8 @@ used for a single job - for instance, you may have a variable defined in both
 a test suite and a product that appear in the same job template. The precedence
 order for variables is as follows (from lowest to highest):
 
+- Instance-wide test settings (defined in the `[test_settings]` section of `openqa.ini`)
+
```
Martchus: The blank lines between these points should probably be removed.

lib/OpenQA/JobSettings.pm
```diff
@@ -17,6 +18,16 @@ my $PLACEHOLDER_RE = qr/(%+)(\w+)(%+)/;
 sub generate_settings ($params) {
     my $settings = $params->{settings};
     my @worker_class;
+
+    if (my $app = OpenQA::App->singleton) {
+        if (my $global_test_settings = $app->config->{test_settings}) {
+            push @worker_class,
+              map { $global_test_settings->{$_} } grep { uc eq 'WORKER_CLASS' } keys %$global_test_settings;
+            map { my $uc_key = uc; $settings->{$uc_key} = $global_test_settings->{$_} }
+              grep { uc ne 'WORKER_CLASS' } keys %$global_test_settings;
```
Martchus: Don't use `map` when just discarding the result.

Martchus: It is not nice that the `grep { uc eq 'WORKER_CLASS' } keys %$global_test_settings` part is redundant and doing the iteration twice is probably also inefficient. I would prefer *one* loop with `next if $uc_key eq 'WORER_CLASS';` that handles populating the array and the hash in one go.

Martchus (changes requested):

docs/UsersGuide.md
okurz: done. separate commit as this touches more than just the line I added

lib/OpenQA/JobSettings.pm
okurz: done

okurz:
adressed review suggestions from Martchus.

Martchus (approved):

lib/OpenQA/JobSettings.pm
```diff
@@ -17,6 +18,21 @@ my $PLACEHOLDER_RE = qr/(%+)(\w+)(%+)/;
 sub generate_settings ($params) {
     my $settings = $params->{settings};
     my @worker_class;
+
+    if (my $app = OpenQA::App->singleton) {
+        if (my $global_test_settings = $app->config->{test_settings}) {
+            for my $k (keys %$global_test_settings) {
```
perlpunk, line 24: Whenever a list of hash keys ends up in a list somewhere, I'd suggest sorting it for easier reproducing and comparing:
```suggestion
            for my $k (sort keys %$global_test_settings) {
```

perlpunk: Nevermind, `@worker_class` is sorted later in the code anyway :)

asmorodskyi (approved):

PR #7603, opened 2026-07-07
fix: OpenSuseKernelBug: Never route kernel bug reports to PUBLIC product

schlad:
Kernel regressions found by openQA are frequently against unreleased kernels, so filing them straight into the customer-visible PUBLIC SLE Bugzilla product (as done for generic bug reports on SLE 15 SP3+) can disclose pre-release kernel issues. Add an allow_public flag to get_bugzilla_product_name/_sle_product and have the kernel bug reporter opt out, keeping generic bug reports unaffected.

new behaviour can be checked here: 
https://rmarliere-openqa.qe.prg2.suse.org/tests/1686#step/selftests:cgroup:test_cpu/4
Screenshot from 2026-07-07 17-42-32

vs:
https://openqa.suse.de/tests/23212669#step/test_cpu/8
Screenshot from 2026-07-07 17-41-58

### DIFF
--- a/lib/OpenQA/WebAPI/Plugin/IssueReporter/OpenSuseBugzillaUtils.pm
+++ b/lib/OpenQA/WebAPI/Plugin/IssueReporter/OpenSuseBugzillaUtils.pm
@@ -60,8 +60,8 @@ sub get_bugzilla_distri_name ($raw_distri) {
 
 # taken from the external_reporting.html.ep
 # likely this could be improved
-sub get_bugzilla_product_name ($job, $raw_distri, $distri_ref) {
-    return _sle_product($job, $distri_ref) if $raw_distri eq 'sle';
+sub get_bugzilla_product_name ($job, $raw_distri, $distri_ref, $allow_public = 1) {
+    return _sle_product($job, $distri_ref, $allow_public) if $raw_distri eq 'sle';
     return _sle_micro_product($job) if $raw_distri eq 'sle-micro';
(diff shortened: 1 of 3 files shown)


Martchus (approved):

okurz (changes requested):
I don't understand why the kernel needs to be treated different here. The default should always be public. Why need to hide something when it's about the kernel? This hardly contains any customer data. As needed anyone can change visibility. That is the same for manual kernel reports and the same for other components which can also include pre-release code.

schlad:
> I don't understand why the kernel needs to be treated different here. The default should always be public. Why need to hide something when it's about the kernel? This hardly contains any customer data. As needed anyone can change visibility. That is the same for manual kernel reports and the same for other components which can also include pre-release code.

fair point. Request came from @mdoucha and I in fact agree with it.
It less about `hiding` and more about noise from my point of view. Also we deal at times with `embargoed` kernel bugs, so its safer to have the default `non-public` and switch to public than the other way around.

@mdoucha what do you think?

hrommel:
I think as long as the bug reports for the embargoed issues are locked on a need-to-know base and as long as the test artefacts (kernel rpms) and not published we should be safe with sending new bug report to the public part of BSC?

hrommel:
I have asked the SUSE Solutions Security team for advise. They told me that the only issue would be mentioning of fallout of embargoed bugs here. They don't see a need to change it currently.

schlad:
all clear thanks. Closing.

PR #7604, opened 2026-07-07
feat(worker): Allow configuring lock to throttle job execution

Martchus:
This is still a draft to show how a worker-side approach could look like. The comment in `workers.ini` contains more details. The locking via Minion worked in a few manual tests I did and is probably simpler on our side compared to adding a lock file. There's only one FIXME to resolve and of course I need to do more manual testing and add unit tests.

The main concern I have is the `LOCK_EXPIRATION` setting. I put 60 seconds there but considering asset downloads (that happen before QEMU even starts and the load increases) might take significantly longer we might need to put a higher value there. That has the disadvantage that other worker slots will be blocked for longer with re-evaluating the load threshold (so we possibly limit ourselves to just a few workers needlessly long).

It would be good if the worker could change the expiration of the lock it holds after the setup (without temporarily releasing it). This way we could initially set a high expiration (definitely covering any asset downloads that might take long) and lower it when we started isotovideo. Unfortunately, I haven't found a way to do this with Minion locks.

We could also only acquire the lock once the setup is done. Unfortunately, if we then cannot acquire the job we needed to set the job to `incomplete` or `setup_failure`. (We could set it back to `scheduled` and pretend nothing happened but that's nothing we currently already do at this point and I'm not sure whether we should open this can of worms.)

Note that a centralized scheduler-based approach has even less chances of taking the setup time into account. There would probably also be some cool-down setting (similar to `LOCK_EXPIRATION` here, but as part of the web UI settings) and we would have the same problem there as well.

The advantage a scheduler-based approach has is being able to distribute jobs across as many worker hosts as possible. That's of course only useful if we have sufficiently many worker hosts (e.g. of a certain architecture). So maybe we could combine this worker-side locking/throttling with a change in the scheduler to max out distribution of jobs across worker hosts (but not having any limit/throttling in the scheduler).

Related ticket: https://progress.opensuse.org/issues/203478

### DIFF
--- a/etc/openqa/workers.ini
+++ b/etc/openqa/workers.ini
@@ -85,6 +85,27 @@
 # Set to 0 to disable.
 #CRITICAL_LOAD_AVG_THRESHOLD = 40
 
+# Let the worker acquire a lock with the specified name before accepting a job.
+# The lock will be released after the job is done or after the specified
+# expiration. The expiration is effective when isotovideo has been started. The
+# worker will reject the job if it cannot acquire the lock. The number of local
+# worker instances that can acquire the lock can be limited.
+# By default no lock is acquired. The purpose of configuring a lock is
+# throttling the startup of jobs on a particular worker host so the setting
+# CRITICAL_LOAD_AVG_THRESHOLD is more effective. (Otherwise many worker
+# instances might still all start newly scheduled jobs at the same time. With
+# the example config below, only 5 worker instances would be able to start newly
+# scheduled jobs at the same time. More worker instances would only start
+# further jobs after 30 seconds - unless the load exceeds the configured
+# threshold then.)
+# Workers that could not acquire a lock will temporarily show as unavailable
+# with an according note. This prevents the scheduler from immediately assigning
+# a job on these slots again (increasing the chances it will spread future
+# job assignments across different/other worker hosts).
+#LOCK_NAME = qemu-job
+#LOCK_EXPIRATION = 30
+#LOCK_LIMIT = 5
+
 # If IPMI_HOSTNAME, IPMI_USER and IPMI_PASSWORD is set the worker will periodically
 # ensure that the IPMI controlled SUT is powered off when no job is running.
 # The default period is 300 seconds.
--- a/lib/OpenQA/CacheService/DB.pm
+++ b/lib/OpenQA/CacheService/DB.pm
@@ -1,7 +1,7 @@
 # Copyright SUSE LLC
 # SPDX-License-Identifier: GPL-2.0-or-later
 
-package OpenQA::CacheService::Db;
+package OpenQA::CacheService::DB;
 use Mojo::Base -strict, -signatures;
 
 use Mojo::SQLite;
--- a/lib/OpenQA/Worker.pm
+++ b/lib/OpenQA/Worker.pm
@@ -47,7 +47,7 @@ use OpenQA::Constants
   qw(WEBSOCKET_API_VERSION WORKER_COMMAND_QUIT WORKER_SR_BROKEN WORKER_SR_DONE WORKER_SR_DIED WORKER_SR_FINISH_OFF);
 use OpenQA::Client;
 use OpenQA::CacheService::Client;
-use OpenQA::CacheService::Db;
+use OpenQA::CacheService::DB;
 use OpenQA::Log qw(log_error log_warning log_info log_debug add_log_channel remove_log_channel);
 use OpenQA::Utils qw(prjdir load_avg);
 use OpenQA::Worker::WebUIConnection;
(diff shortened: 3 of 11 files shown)


etc/openqa/workers.ini
```diff
@@ -85,6 +85,22 @@
...
+# By default no lock is acquired. The purpose of configuring a lock is
+# throttling the startup of jobs on a particular worker host so the setting
+# CRITICAL_LOAD_AVG_THRESHOLD is more effective. (Otherwise many worker
+# instances might still all start newly scheduled jobs at the same time. With
+# the example config below, only 5 worker instances would be able to start newly
+# scheduled jobs at the same time. More worker instances would only start
+# further jobs after one minute - unless the load exceeds the configured
+# threshold then.)
+#LOCK_NAME = qemu-job
+#LOCK_EXPIRATION = 60
```
okurz: The ticket suggests
> Introduce sleep jitter and load-proportionate delays to idle slot polling, giving the load average time to rise and trigger the threshold. 

how about simply sleeping `factor * current_load_value`? either then this value `LOCK_EXPIRATION` can be computed this way or the algorithm replaced

Martchus: I'm not sure whether using `current_load_value` makes sense as the problem is the time window where the load hasn't increased yet.

Martchus: I think it is easily possible to change the expiration of an existing lock.

So I would initially put a high expiration when creating a lock (as we do for locks on Minion jobs where the expiration is mainly to prevent locks from staying around for forever in case something goes wrong). If the job is started for real I would lower the expiration to `LOCK_EXPIRATION`. With that this can still be a low value, e.g. 60 seconds or even lower (considering QEMU will probably start producing high load very quickly).

Martchus:
I've tested this locally with a few workers and quickly restarting a few jobs. It worked as expected. There's only one flaw I've noticed:

```
[warn] [pid:50489] Limited by configured lock 'qemu-job' - checking again for web UI 'http://localhost:9526' in 251.53 s
```

The 251.53 seconds are way too long. It makes sense to consider the worker unavailable for a *short* time to give the scheduler a chance to try slots on another host. However, the worker should not only reevaluate its availability after minutes because maybe the load isn't too bad after all. Reducing this waiting time shouldn't be too hard to tweak, though.

EDIT: I've just pushed a change that improves this. With that change everything seems to work quite nicely in my local tests:

1. Start 4 initially idling worker instances with a limit of 2 and a lock expiration of 30 seconds.
2. Restart 4 jobs.
3. Only 2 jobs are started. So the lock limit works. The other jobs are tried to be assigned but set back to scheduled. At this point 2 workers are shown as working and 2 workers are shown as unavailable.
4. The remaining 2 jobs are started after 30 seconds. Now all 4 workers are shown as working. So the lock expiration works and the re-check is done quickly enough to not cause any further delays.
    * If my system load would have been too high from starting the first two jobs, the 2nd two jobs would not have been started and the other workers would be unavailable with the usual note. I haven't tested that for real but we have unit tests for re-evaluating the availability.

I'm not sure whether we need to tweak the scheduler for now. There's a window when the workers will show as unavailable due to the lock so the scheduler is already forced to be more clever than just assigning the jobs to the throttled slots again.

etc/openqa/workers.ini
Martchus: The latest change contains adjusting the expiration after isotovideo has been started. I also lowered the default to 30 s because without having to worry about job setup delaying the load we probably don't need to go higher. This worked in my local tests, e.g. a 3rd worker was able to pick up another job after being initially held back only the configured 30 s (despite the lock being initially created with an expiration of an hour).

lib/OpenQA/Worker.pm
```diff
@@ -103,10 +113,40 @@ sub new ($class, $cli_options) {
...
+# reduces the expiration of the lock with the longest expiration from the initial ONE_HOUR to the configured LOCK_EXPIRATION
+# note: This function is called by the engine when the setup is over and the actual execution starts and thus an increased load
+#       is expected.
+sub update_job_guard_expiration ($self) {
+    return 0 unless ref $self->{_guard_or_error};    # skip if we don't hold a lock
+    my $settings = $self->settings->global_settings;
+    return undef unless defined(my $lock_name = $settings->{LOCK_NAME});
+    my $expiration = int($settings->{LOCK_EXPIRATION} // 30);
+    my $query = q(update minion_locks set expires = datetime('now', ?) where name = ? order by expires desc limit 1);
+    $self->{_minion_sqlite}->db->query($query, "+$expiration seconds", $lock_name);
```
Martchus, line 145: This will just update the lock with the highest expiration and not necessarily the lock we hold. However, I don't think this will cause any problems. (There's no API to get the ID an acquired lock so I can't just use `where id = ?`.)

okurz (approved):
Seems to work fine and looks fine. I also tested this locally with just a single instance. Please also see my related PR inspired by your change:
* #7609

lib/OpenQA/CacheService/DB.pm
perlpunk, line 1: I would call this file lib/OpenQA/CacheService/DB.pm

kalikiana (approved):

etc/openqa/workers.ini
```diff
@@ -85,6 +85,27 @@
...
+# scheduled jobs at the same time. More worker instances would only start
+# further jobs after 30 seconds - unless the load exceeds the configured
+# threshold then.)
+# Workers that could not acquire a lock will temporarily show as unavailable
+# with an according note. This prevents the scheduler from immediately assigning
+# a job on these slots again (increasing the chances it will spread future
+# job assignments across different/other worker hosts).
+#LOCK_NAME = qemu-job
+#LOCK_EXPIRATION = 30
+#LOCK_LIMIT = 5
```
okurz, line 107: please use the default values here, i.e. empty lock_name and such

PR #7607, opened 2026-07-08
feat: allow UEFI_PFLASH_CODE as asset

Vogtinator:
Just like UEFI_PFLASH_VARS, allow to download it as an asset.

This allows to supply a consistent firmware image independent of the worker the test runs on.

Related ticket: https://progress.opensuse.org/issues/203625

Depends on: https://github.com/os-autoinst/os-autoinst/pull/3011

Verification run: https://openqa.opensuse.org/tests/6081282 (monkeypatched on experimental ow28 worker. It attempts to download it at least)

### DIFF
--- a/lib/OpenQA/Utils.pm
+++ b/lib/OpenQA/Utils.pm
@@ -379,12 +379,12 @@ sub run_cmd_with_log_return_error ($cmd, %args) {
 }
 
 # passing $value is optional but makes the result more accurate
-# (it affects only UEFI_PFLASH_VARS currently).
+# (it affects only UEFI_PFLASH_(CODE|VARS) currently).
 sub asset_type_from_setting ($setting, $value = undef) {
     return 'iso' if $setting eq 'ISO' || $setting =~ /^ISO_\d+$/;
     return 'hdd' if $setting =~ /^HDD_\d+$/;
-    # non-absolute-path value of UEFI_PFLASH_VARS treated as HDD asset
-    return 'hdd' if $setting eq 'UEFI_PFLASH_VARS' && ($value // '') !~ m,^/,;
+    # non-absolute-path value of UEFI_PFLASH_(CODE|VARS) treated as HDD asset
+    return 'hdd' if $setting =~ /^UEFI_PFLASH_(?:CODE|VARS)$/ && ($value // '') !~ m{^/};
     return 'repo' if $setting =~ /^REPO_\d+$/;
     return 'other' if $setting =~ /^ASSET_\d+$/ || $setting eq 'KERNEL' || $setting eq 'INITRD';
     # empty string if this doesn't look like an asset type
--- a/docs/UsersGuide.md
+++ b/docs/UsersGuide.md
@@ -1468,6 +1468,7 @@ The following job settings are specifying that an asset is required by a job:
 - `ISO` (type `iso`)
 - `ISO_n` (type `iso`)
 - `HDD_n` (type `hdd`)
+- `UEFI_PFLASH_CODE` (type `hdd`) (in some cases, see below)
 - `UEFI_PFLASH_VARS` (type `hdd`) (in some cases, see below)
 - `REPO_n` (type `repo`)
 - `ASSET_n` (type `other`)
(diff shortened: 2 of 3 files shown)


okurz (commented):
I wonder why was only UEFI_PFLASH_VARS handled as a special case

Vogtinator:
> I wonder why was only UEFI_PFLASH_VARS handled as a special case

Probably just to implement `PUBLISH_PFLASH_VARS`, to propagate EFI variables to child jobs alongside the HDD.

The difference to _CODE is that there, publishing it does not make sense (it's effectively read-only), but downloading does.

lib/OpenQA/Utils.pm
```diff
@@ -379,12 +379,12 @@ sub run_cmd_with_log_return_error ($cmd, %args) {
...
 # passing $value is optional but makes the result more accurate
-# (it affects only UEFI_PFLASH_VARS currently).
+# (it affects only UEFI_PFLASH_(CODE|VARS) currently).
 sub asset_type_from_setting ($setting, $value = undef) {
     return 'iso' if $setting eq 'ISO' || $setting =~ /^ISO_\d+$/;
     return 'hdd' if $setting =~ /^HDD_\d+$/;
-    # non-absolute-path value of UEFI_PFLASH_VARS treated as HDD asset
-    return 'hdd' if $setting eq 'UEFI_PFLASH_VARS' && ($value // '') !~ m,^/,;
+    # non-absolute-path value of UEFI_PFLASH_(CODE|VARS) treated as HDD asset
+    return 'hdd' if $setting =~ '^UEFI_PFLASH_(CODE|VARS)$' && ($value // '') !~ m,^/,;
```
perlpunk: Use the match operator instead of a string if you are using `=~`.
And the default delimiter for the match operator should be `m{}` if it can't be `/`.
```suggestion
    return 'hdd' if $setting =~ /^UEFI_PFLASH_(CODE|VARS)$/ && ($value // '') !~ m{^/};
```
I'm wondering why perlcritic didn't complain though.

perlpunk: > I'm wondering why perlcritic didn't complain though.

Ah, we have excluded this policy, there are many many violations.
https://metacpan.org/pod/Perl::Critic::Policy::RegularExpressions::ProhibitUnusualDelimiters
Maybe we should actually use it.

Vogtinator: Done!

Vogtinator:
Not sure how to fully test this without a dev setup with separate webui and worker hosts.

lib/OpenQA/Utils.pm
```diff
@@ -379,12 +379,12 @@ sub run_cmd_with_log_return_error ($cmd, %args) {
...
 # passing $value is optional but makes the result more accurate
-# (it affects only UEFI_PFLASH_VARS currently).
+# (it affects only UEFI_PFLASH_(CODE|VARS) currently).
 sub asset_type_from_setting ($setting, $value = undef) {
     return 'iso' if $setting eq 'ISO' || $setting =~ /^ISO_\d+$/;
     return 'hdd' if $setting =~ /^HDD_\d+$/;
-    # non-absolute-path value of UEFI_PFLASH_VARS treated as HDD asset
-    return 'hdd' if $setting eq 'UEFI_PFLASH_VARS' && ($value // '') !~ m,^/,;
+    # non-absolute-path value of UEFI_PFLASH_(CODE|VARS) treated as HDD asset
+    return 'hdd' if $setting =~ /^UEFI_PFLASH_(CODE|VARS)$/ && ($value // '') !~ m{^/};
```
perlpunk: Oh, and use `?:` to avoid capturing.

```suggestion
    return 'hdd' if $setting =~ /^UEFI_PFLASH_(?:CODE|VARS)$/ && ($value // '') !~ m{^/};
```

There is also a perlcritic policy for that, but there was one place hard to fix, so it's still an open pull request.

perlpunk:
btw, to avoid the fullstack-unstable test failure, rebase

lib/OpenQA/Utils.pm
Vogtinator: Done as well, still "works" (https://openqa.opensuse.org/tests/6081332)

perlpunk (approved):

okurz (approved):

Vogtinator:
FTR, works! https://openqa.opensuse.org/tests/6083176#downloads

PR #7608, opened 2026-07-08
ci: Add cli tool to report uncovered lines

perlpunk:
Issue: https://progress.opensuse.org/issues/203757

This is executed after the codecov upload and will fail the codecov step if there are any uncovered lines.

Example output:
```
Checking for uncovered lines 
t/01-test-utilities.t:64     | pass "something 2";
t/01-test-utilities.t:65     | pass "something 3";
t/01-test-utilities.t:66     | pass "something 4";
t/01-test-utilities.t:67     | pass "something 5";
t/01-test-utilities.t:68     | pass "something 6";
t/01-test-utilities.t:69     | pass "something 7";
t/01-test-utilities.t:70     | pass "something 8";
t/01-test-utilities.t:71     | pass "something 9";
t/01-test-utilities.t:72     | pass "something 10";
t/01-test-utilities.t:73     | pass "something 11";
Reached limit of 10 printed lines. Stopping.
10 uncovered lines

Exited with code exit status 1
```
Demo: https://app.circleci.com/pipelines/github/perlpunk/openQA/518/workflows/a11e59ee-9392-43e9-b75f-434abaa2ce98/jobs/4524?invite=true&expanded=true#step-114-0_49

### DIFF
--- a/tools/check-line-coverage
+++ b/tools/check-line-coverage
@@ -1,0 +1,30 @@
+#!/bin/bash
+
+input_file=$1
+limit=${2:-500}
+
+declare -i counter=0
+
+echo "Checking for uncovered lines" >&2
+
+while read -r file line; do
+    [[ $line -eq 0 ]] && continue
+    if [[ -f "$file" ]]; then
+        content=$(sed -n "${line}p" "$file")
+        printf "%-60s %s\n" "$file:$line" "| $content"
+        counter+=1
+        if [[ $counter -ge $limit ]]; then
+            echo "Reached limit of $limit printed lines. Stopping." >&2
+            break
+        fi
+    else
+        echo "Warning: File '$file' not found." >&2
+    fi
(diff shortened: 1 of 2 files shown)


tools/check-line-coverage
okurz, line 1: Please compare to https://github.com/os-autoinst/os-autoinst/blob/master/tools/check_coverage . I would prefer using the same approach for both repos, at best with shared code

perlpunk: Sure, we can put the script from this PR into the common repo. But maybe try this out first here?

The check_coverage:
* Only checks Devel::Cover coverage, which is statement coverage, and we don't have 100% yet there
* Does not report the uncovered lines

okurz: Yes, that's fine. We can continue with the script in this repo first and then consider to move to common

okurz (approved):

tools/check-line-coverage
perlpunk: And besides that, tools/check_coverage is not used anywhere... :)

Martchus (approved):

PR #7609, opened 2026-07-08
feat(worker): Implement load-proportionate delay

okurz:
Motivation:
During wave concurrent test triggers, multiple idle workers read the same low
load due to OS load average sampling lag, launching heavy QEMU instances at the
exact same millisecond and causing severe boot avalanches.

Design Choices:
Instead of a hard critical limit, we introduce parameters to calculate a
load-proportionate sleep delay before accepting a job. Sleeping staggers job
acceptance, letting the OS sampling lag catch up so subsequent workers can
correctly reject jobs when overloaded. Low loads bypass sleep to preserve
fast performance on local single-instance development hosts.

Benefits:
This safely prevents thundering herd load storms on high-density hosts while
maintaining immediate responsiveness for local test developers.

Related issue: https://progress.opensuse.org/issues/203478

This is related to #7604 which inspired me to a complementary idea.

### DIFF
--- a/docs/Installing.md
+++ b/docs/Installing.md
@@ -1065,23 +1065,14 @@ to access files of the host) use the following setting:
 HOST = http://openqa.example.com
 ```
 
-The load of the system can affect test execution and cause failures due to delays of command execution and [thrashing](https://en.wikipedia.org/wiki/Thrashing_(computer_science)). If the average over a period of 15 minutes exceeds the specified value the worker will not accept new jobs.
-
-
-
-
-
-workers.ini
-
-
-
-``` ini
-[global]
-# Set to 0 to disable.
-#CRITICAL_LOAD_AVG_THRESHOLD = 40
-```
-
-
+The load of the system can affect test execution and cause failures due to delays
+of command execution and [thrashing](<https://en.wikipedia.org/wiki/Thrashing_(computer_science)>).
+To prevent "load storms" (synchronized boot avalanches), workers check the
+system load average before picking up jobs. If the load is very high, they
+temporarily reject jobs. If the load is moderately high, they introduce a
+load-proportionate sleep delay to stagger job acceptance. The specific load
+thresholds, factor, and jitter are configurable in `workers.ini` (e.g.,
+`CRITICAL_LOAD_AVG_THRESHOLD`, `WORKER_LOAD_LOW_LIMIT`).
 
 Once you got workers running they should show up in the admin section of openQA in
 the workers section as 'idle'. When you get so far, you have your own instance
--- a/lib/OpenQA/Worker.pm
+++ b/lib/OpenQA/Worker.pm
@@ -885,6 +885,28 @@ sub _check_system_utilization (
     return undef;
 }
 
+sub _calculate_load_delay (
+    $self,
+    $threshold = $self->settings->global_settings->{CRITICAL_LOAD_AVG_THRESHOLD},
+    $load = load_avg())
+{
+    return 0 if !$threshold || @$load < 3;
+    my $global_settings = $self->settings->global_settings;
+    my $low_limit = $global_settings->{WORKER_LOAD_LOW_LIMIT};
+    if ($load->[0] >= $low_limit && $load->[0] < $threshold) {
+        my $factor = $global_settings->{WORKER_LOAD_FACTOR};
+        my $jitter = $global_settings->{WORKER_LOAD_JITTER};
+        my $delay = $factor * $load->[0] + ($jitter > 0 ? rand $jitter : 0);
+        if ($delay > 0) {
+            my $format
+              = 'Load is %.2f (between %d and %d). Delaying job acceptance for %.2f seconds to prevent load storms.';
+            log_debug sprintf $format, $load->[0], $low_limit, $threshold, $delay;
+            return $delay;
+        }
+    }
+    return 0;
+}
+
 sub _setup_pool_directory ($self) {
     # skip if we have already locked the pool directory
     return undef if defined $self->{_pool_directory_lock_fd};
--- a/lib/OpenQA/Worker/Job.pm
+++ b/lib/OpenQA/Worker/Job.pm
@@ -139,6 +139,19 @@ sub accept ($self) {
     my $info = $self->info;
     die 'attempt to accept job without ID and job info' unless $id && defined $info && ref $info eq 'HASH';
     die 'attempt to accept job which is not newly initialized' unless $self->is_supposed_to_start;
+
+    my $worker = $self->worker;
+    my $delay = $worker->_calculate_load_delay;
+    if ($delay > 0 && !$self->{_delay_applied}) {
+        $self->{_delay_applied} = 1;
+        $self->_set_status(accepting => {}) if $self->status ne 'accepting';
+        Mojo::IOLoop->timer(
+            $delay => sub {
+                $self->accept if $self->is_supposed_to_start;
+            });
+        return 1;
+    }
+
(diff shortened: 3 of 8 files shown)


docs/Installing.md
```diff
@@ -1065,23 +1065,7 @@ to access files of the host) use the following setting:
...
-
-
-``` ini
-[global]
-# Set to 0 to disable.
-#CRITICAL_LOAD_AVG_THRESHOLD = 40
-```
-
-
+The load of the system can affect test execution and cause failures due to delays of command execution and [thrashing](https://en.wikipedia.org/wiki/Thrashing_(computer_science)). To prevent "load storms" (synchronized boot avalanches), workers check the system load average before picking up jobs. If the load is very high, they temporarily reject jobs. If the load is moderately high, they introduce a load-proportionate sleep delay to stagger job acceptance. The specific load thresholds, factor, and jitter are configurable in `workers.ini` (e.g., `CRITICAL_LOAD_AVG_THRESHOLD`, `WORKER_LOAD_LOW_LIMIT`).
```
Martchus: This should be wrapped.

lib/OpenQA/Worker.pm
```diff
@@ -811,6 +812,20 @@ sub _check_system_utilization (
...
+    my $low_limit = $global_settings->{WORKER_LOAD_LOW_LIMIT};
+    if ($load->[0] >= $low_limit && $load->[0] < $threshold) {
+        my $factor = $global_settings->{WORKER_LOAD_FACTOR};
+        my $jitter = $global_settings->{WORKER_LOAD_JITTER};
+        my $delay = $factor * $load->[0] + rand $jitter;
+        if ($delay > 0) {
+            my $format
+              = 'Load is %.2f (between %d and %d). Sleeping for %.2f seconds before re-checking to prevent load storms.';
+            log_debug sprintf $format, $load->[0], $low_limit, $threshold, $delay;
+            sleep $delay;
```
Martchus: Don't use sleep here. Worker code should always return the control flow to the event loop as soon as possible to maintain the ws connection. This means in practice that the worker will be unresponsive for many seconds, maybe even a minute. In that time the job will just be considered assigned (as it is not rejected) leaving the scheduler no window to re-assign it to a different worker and leaving users wondering why the job is stuck in the state assigned (which normally only happens when there are directly chained dependencies).

lib/OpenQA/Worker/Settings.pm
```diff
@@ -21,6 +21,9 @@ has 'webui_host_specific_settings';
 
 use constant QEMU_PORT_OFFSET => 20_002;
 use constant DEFAULT_CRITICAL_LOAD_AVG_THRESHOLD => 40;
+use constant DEFAULT_WORKER_LOAD_LOW_LIMIT => 4;
+use constant DEFAULT_WORKER_LOAD_FACTOR => 1;
+use constant DEFAULT_WORKER_LOAD_JITTER => 5;
```
Martchus: I wouldn't enable this by default, e.g. on a local development setup it would be rather annoying having to randomly wait a few seconds before a test can be started just because the average system load exceeds the low limit.

lib/OpenQA/Worker/Settings.pm
```diff
@@ -21,6 +21,9 @@ has 'webui_host_specific_settings';
 
 use constant QEMU_PORT_OFFSET => 20_002;
 use constant DEFAULT_CRITICAL_LOAD_AVG_THRESHOLD => 40;
+use constant DEFAULT_WORKER_LOAD_LOW_LIMIT => 4;
+use constant DEFAULT_WORKER_LOAD_FACTOR => 1;
```
Martchus: Ok, let's say the load is 10 (so above the low limit). Then with this default of 1 we'd wait 10 seconds to re-check the load. That's probably not enough waiting time until the threshold of 40 is reached (considering test setup is probably just causing a little bit I/O but might take minutes).

lib/OpenQA/Worker.pm
```diff
@@ -811,6 +812,20 @@ sub _check_system_utilization (
...
 {
     return undef if !$threshold || @$load < 3;
+    my $global_settings = $self->settings->global_settings;
+    my $low_limit = $global_settings->{WORKER_LOAD_LOW_LIMIT};
+    if ($load->[0] >= $low_limit && $load->[0] < $threshold) {
+        my $factor = $global_settings->{WORKER_LOAD_FACTOR};
+        my $jitter = $global_settings->{WORKER_LOAD_JITTER};
+        my $delay = $factor * $load->[0] + rand $jitter;
+        if ($delay > 0) {
+            my $format
```
Martchus, line 901: Doing the throttling here seems conceptually wrong because this function is supposed to check the system utilization and *not* supposed to block the general operation of the worker.

Martchus (changes requested):

docs/Installing.md
okurz: done.

lib/OpenQA/Worker.pm
okurz: ok, changed to a Mojo::IOLoop->timer invocation

lib/OpenQA/Worker/Settings.pm
okurz: true. set both factor+jitter to default 0

lib/OpenQA/Worker/Settings.pm
okurz: true. so we should default the load to 0 but suggest to use bigger numbers when wanting to use this feature. The comment in .ini now says

> Delay formula: WORKER_LOAD_FACTOR * load + rand(WORKER_LOAD_JITTER)
> The feature is disabled by default. If enabling, a WORKER_LOAD_FACTOR of 10 to 15
> is recommended to allow the OS load average sufficient time to reflect the actual load.

lib/OpenQA/Worker.pm
okurz: done. only returning the calculated value here.

kalikiana (approved):

lib/OpenQA/Worker/Job.pm
```diff
@@ -139,6 +139,19 @@ sub accept ($self) {
     my $info = $self->info;
     die 'attempt to accept job without ID and job info' unless $id && defined $info && ref $info eq 'HASH';
     die 'attempt to accept job which is not newly initialized' unless $self->is_supposed_to_start;
+
+    my $worker = $self->worker;
+    my $delay = $worker->_calculate_load_delay;
+    if ($delay > 0 && !$self->{_delay_applied}) {
+        $self->{_delay_applied} = 1;
+        $self->_set_status(accepting => {}) if $self->status ne 'accepting';
```
Martchus, line 147: This line should be moved before the `if` instead of being duplicated.

lib/OpenQA/Worker/Job.pm
```diff
@@ -139,6 +139,19 @@ sub accept ($self) {
...
+        Mojo::IOLoop->timer(
+            $delay => sub {
```
Martchus, line 149: This keeps the job in suspense for `$delay` seconds. It'll look to the web UI (and users of the web UI) like the worker is slow to accept the job for *whatever* reason. That's not ideal. My approach explicitly made the worker reject the job to avoid this.

I'm also wondering whether delaying the acceptance makes sense here. We're already past `_can_grab_job` which re-evaluates e.g. the load. So if we delay the acceptance here we'll just wait a few seconds and then just accept the job regardless of whether something has changed in those seconds. That seems not *that* useful, or am I wrong about the re-checking? Or do you want to wait here purely for the external effect on other worker slots?

Martchus (changes requested):

PR #7615, opened 2026-07-10
fix(apparmor): allow webui access to fixed assets

marmarek:
Allow reading files in hdd/fixed and iso/fixed, not only listing them.
Without this, a worker that uses cache service (instead of shared NFS
mount) would fail to fetch those, getting HTTP 500.


Martchus (approved):

okurz:
> Allow reading files in hdd/fixed and iso/fixed, not only listing them. Without this, a worker that uses cache service (instead of shared NFS mount) would fail to fetch those, getting HTTP 500.

Strange. can anyone explain why we don't have a problem with that for 10 years using the cache service and apparmor?

marmarek:
Are you using that, _and fixed assets_? I used fixed assets before, but only in `repo` or `other` dirs (where apparmor rule is more relaxed), and only recently added some to `iso` dir. This worked fine for workers using NFS share, and finally failed on those with cache service..

okurz (changes requested):
> Are you using that, _and fixed assets_? 

yes, that is used already. anyway, we can approve

marmarek:
miss-click? or is here something to change?

Martchus:
I was also wondering why we haven't encountered this, though. We use fixed assets of all kinds (including iso and hdd) on o3 where we have also AppArmor enabled and rely on the cache service.

Martchus:
Maybe we don't see this because it isn't actually the openQA process that serves these assets in our setups but NGINX. I recommend you also configure Apache, NGINX or whatever you're using to serve assets because it is more efficient. There are config examples under https://github.com/os-autoinst/openQA/tree/master/etc for Apache and NGINX.

marmarek:
> Maybe we don't see this because it isn't actually the openQA process that serves these assets in our setups but NGINX

That's very likely.

PR #7616, opened 2026-07-10
test: reuse reserved socket in create_webapi

okurz:
Motivation:
When tests passed the plain reserved port to create_webapi, Mojo bound a fresh
listener that only succeeded on IPv6 loopback and deadlocked IPv4 clients,
hanging the test. This surfaced locally as a lengthy timeout in t/14-grutasks.t
under HEAVY=1 with a real PostgreSQL, while CI's networking happened to
mask it.

### DIFF
--- a/t/lib/OpenQA/Test/Utils.pm
+++ b/t/lib/OpenQA/Test/Utils.pm
@@ -269,10 +269,7 @@ sub stop_service {
 }
 
 sub create_webapi ($port = undef, $no_cover = undef) {
-    # Callers commonly pass the plain reserved port via `reserve_ports(...)->sockport`;
-    # restore its socket fd so the daemon reuses the parent's bound socket.
-    $port = OpenQA::Utils::raw_port_for_reserved($port) if defined $port && $port =~ /^\d+$/;
-    $port //= raw_service_port 'webui';
+    $port = raw_service_port('webui', $port);
     note("Starting WebUI service. Port: $port");
 
     my $h = _setup_sigchld_handler 'openqa-webapi', start sub {
@@ -299,7 +296,7 @@ sub create_webapi ($port = undef, $no_cover = undef) {
 }
 
 sub create_websocket_server ($port, $bogus, $with_embedded_scheduler = undef, $no_cover = undef) {
-    OpenQA::WebSockets::Client->singleton->port(to_plain_service_port($port //= raw_service_port('websocket')));
+    OpenQA::WebSockets::Client->singleton->port(to_plain_service_port($port = raw_service_port('websocket', $port)));
     note "Starting WebSocket service (port: $port, bogus: $bogus)";
     my $h = _setup_sigchld_handler 'openqa-websocket', start sub {
         _setup_sub_process 'openqa-websocket';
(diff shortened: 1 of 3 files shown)


t/lib/OpenQA/Test/Utils.pm
```diff
@@ -269,6 +269,9 @@ sub stop_service {
 }
 
 sub create_webapi ($port = undef, $no_cover = undef) {
+    # Callers commonly pass the plain reserved port via `reserve_ports(...)->sockport`;
+    # restore its socket fd so the daemon reuses the parent's bound socket.
+    $port = OpenQA::Utils::raw_port_for_reserved($port) if defined $port && $port =~ /^\d+$/;
```
Martchus: I'm wondering whether it makes sense to do the same in the other `create_…` functions. Maybe it would make sense to have a function that combines this with `raw_service_port` so we still have to call only one function to handle the port.

Martchus (approved):
Looks good. I have one suggestion we *might* want to look into but I'm also approving your minimal change.

Martchus (changes requested):
Although the commit message is quite wrong:

> Removes cross-file format coupling and the fragile regex

Where is it removing any regexes? And I'm also not sure about what "cross-file format coupling" is even referring to here exactly.

> while upstream CI's networking happened

You want to merge this as upstream commit so we shouldn't refer to "upstream" like that.

Martchus:
Maybe it makes sense to keep commit messages shorter in general. You don't need to describe the code change itself in such a detailed way. Reviewers have to read the code anyway and this way we just have to read a lengthy description plus the code. (And I think sometimes I'm already just skipping these lengthy descriptions.)

t/lib/OpenQA/Test/Utils.pm
okurz: ok, done. Added an optional `port` parameter to `raw_service_port`. Added in second commit

okurz:
> Although the commit message is quite wrong:
> 
> > Removes cross-file format coupling and the fragile regex
> 
> Where is it removing any regexes?

ok, that part was outdated, sorry.

> And I'm also not sure about what "cross-file format coupling" is even referring to here exactly.

The 'coupling` refers to the initial version where I initially tried parsing the `&fd=` string manually in the test file before realizing it should just be encapsulated in a helper function. Anyway, I removed that as you asked for shorter commit messages.

> > while upstream CI's networking happened
> 
> You want to merge this as upstream commit so we shouldn't refer to "upstream" like that.

How do we call "os-autoinst/openQA" then, origin? I have removed the word "upstream" so now it says "CI's". I guess nowadays commonly people refer to CI in the joint target repository as CI despite anyone being able to execute *all* tests locally :)

> Maybe it makes sense to keep commit messages shorter in general. You don't need to describe the code change itself in such a detailed way. Reviewers have to read the code anyway and this way we just have to read a lengthy description plus the code. (And I think sometimes I'm already just skipping these lengthy descriptions.)

ok, stripped the "Design Choices" now. I hope you still consider Motivation helpful. "Benefits" can be seen as redundant for fixes of course.

Martchus (approved):

kalikiana (approved):

PR #7620, opened 2026-07-13
docs: Remove example from `LOCK_NAME` worker setting

Martchus:
Locking is disabled by default and the example lock name was maybe a bit too specific.

Related ticket: https://progress.opensuse.org/issues/203478

---

Note that we already allow workers to write to the cache directory and it seems to work on openqaworker23. So I haven't included any AppArmor changes here.


Wabri (approved):

Martchus:
Looks like we'll have to update tidy. I'll create a PR.

kalikiana (approved):

PR #7622, opened 2026-07-13
feat(overview-page): Allow restarting jobs with advanced options

Martchus:
Related ticket: https://progress.opensuse.org/issues/203967


okurz:
screenshot plz

Martchus:
Bildschirmfoto_20260714_103614

Like on the test details page it only shows up if the test has dependencies.

okurz (approved):

pevik:
While this is already a nice improvement, I was also thinking about adding a new button on next to `Force restart`. And this button would be a single option `Restart parent skip restarting "OK" (passed/softfailed) children`. Because that's the option we mostly want when assets has been cleared.

force-restart-button

pevik (approved):

Martchus:
I can also add the options to "Force restart" - ~~although I'd simply replicate the menu we already have instead of introducing another code branch with reduced options.~~ Oh, wait you mean "Restart **parent** …". That would be another option we don't have so far. So I guess it would be a different implementation after all. I now also see that this is done entirely in JavaScript so that also means it is probably the easiest to implement it separately anyway.

pevik:
>  Oh, wait you mean "Restart parent …"

Yes.  Do I need to create a separate poo for it or can you just update that the ticket would be for 2 implementations?

And I'd be for merging this as it's an improvement.

Martchus:
> Yes. Do I need to create a separate poo for it or can you just update that the ticket would be for 2 implementations?

I can work on it as part of the first poo. It shouldn't be a big deal.

EDIT: PR: https://github.com/os-autoinst/openQA/pull/7625

kalikiana (approved):

pevik:
@Martchus BTW what is the default when one selects nothing? I hope restarting just that particular job, right? It's not that much obvious from that UI.

Martchus:
> I hope restarting just that particular job, right?

The "Force" button is restarting the job with default options (or the options you've selected before when restarting from the advanced restart menu). So it does what the regular "Restart" button does (or what the selected advanced restarting option does) - just forcefully. That means it will restart dependent jobs based on certain rules (that are explained in the documentation and probably also in some help popover).

PR #7625, opened 2026-07-15
feat(job-restart): Allow restarting parent job skipping OK children

Martchus:
When a job restart is prevented by a directly chained parent dependency, provide a dropdown option next to "Force restart" to restart the parent job directly while skipping "OK" (passed/softfailed) children. This avoids re-running successfully completed children while ensuring the dependency chain remains correct.

Related ticket: https://progress.opensuse.org/issues/203967

### DIFF
--- a/assets/javascripts/openqa.js
+++ b/assets/javascripts/openqa.js
@@ -193,13 +193,57 @@ function showJobRestartResults(responseJSON, newJobUrl, retryFunction, targetEle
   }
   const container = document.createElement('div');
   if (hasResponse && responseJSON.enforceable && retryFunction) {
+    let parentId;
+    if (Array.isArray(errors)) {
+      for (const error of errors) {
+        const match = error.match(/Direct parent (\d+) needs to be cloned as well/);
+        if (match) {
+          parentId = match[1];
+          break;
+        }
+      }
+    }
+
     const button = document.createElement('button');
     button.onclick = retryFunction;
     button.className = 'btn btn-danger force-restart';
-    button.style.float = 'right';
     button.appendChild(document.createTextNode('Force restart'));
-    container.appendChild(button);
+    const btnGroup = document.createElement('div');
+    btnGroup.className = 'btn-group';
+    btnGroup.style.float = 'right';
+    btnGroup.appendChild(button);
+
+    if (parentId) {
+      const dropdownToggle = document.createElement('button');
+      dropdownToggle.type = 'button';
+      dropdownToggle.className = 'btn btn-danger dropdown-toggle dropdown-toggle-split';
+      dropdownToggle.setAttribute('data-bs-toggle', 'dropdown');
+      dropdownToggle.setAttribute('aria-expanded', 'false');
+
+      const srOnly = document.createElement('span');
+      srOnly.className = 'visually-hidden sr-only';
+      srOnly.appendChild(document.createTextNode('Toggle Dropdown'));
+      dropdownToggle.appendChild(srOnly);
+      btnGroup.appendChild(dropdownToggle);
+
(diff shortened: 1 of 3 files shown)


okurz (approved):

pevik:
@Martchus Thanks for working on this. Can I see it on your testing instance? Or at least a screenshot?

Martchus:
I don't like this style check:

```
[10:20:06] xt/00-tidy.t .............. ok   129685 ms ( 0.17 usr  0.01 sys + 128.88 cusr  0.77 csys = 129.83 CPU)

#   Failed test 'Consistent Test::More call format (no parentheses)'
#   at xt/01-style.t line 33.
#          got: 't/ui/26-jobs_restart.t:        # cleanup after leaving subtest (not using txn here as the forked process needs to see changes)
# '
#     expected: ''
# Looks like you failed 1 test of 15.
[10:20:07] xt/01-style.t .............
```

It doesn't tell the line number of the problematic line. I was looking back and forth in the file to spot the mistake - only to realize that the **comment** is the problem. Which is of course the next problem with this check …

I'm wondering whether we should have certain style checks if their implementation is clearly lacking quality. Being able to treat comments as such is a basic requirement for a style check and apparently not implemented here which let's me question the overall quality. (So using regexes like we do in `01-style.t` is not a good approach.)

EDIT: I added a fix for this particular style check but I still find this approach questionable.

Martchus:
Bildschirmfoto_20260715_142856

assets/javascripts/openqa.js
```diff
@@ -193,13 +193,57 @@ function showJobRestartResults(responseJSON, newJobUrl, retryFunction, targetEle
...
-    container.appendChild(button);
+    const btnGroup = document.createElement('div');
+    btnGroup.className = 'btn-group';
+    btnGroup.style.float = 'right';
+    btnGroup.appendChild(button);
+
+    if (parentId) {
+      const dropdownToggle = document.createElement('button');
+      dropdownToggle.type = 'button';
+      dropdownToggle.className = 'btn btn-danger dropdown-toggle dropdown-toggle-split';
```
kalikiana, line 219: Styling seems off here. Tho I guess that's a general issue and not just here. The popup isn't using the button's "danger" variant.

Martchus: The `button` and `dropwonToggle` are styled in-line with each other and `btn-danger` was used like that before. Not sure what you're asking for with regards to the "popup" (do you mean the dropdown menu?).

kalikiana: Yes. It's visible in your screenshot. The button is red, the "Restart parent job..] bit is white.

kalikiana (approved):
Wouldn't block on the style aspect, if it's non-trivial. Looks good otherwise.

assets/javascripts/openqa.js
Martchus: Oh, I'm not sure whether it would good to color the whole context menu red - especially because chosing this option isn't actually "dangerous". This should conceptionally be its own button that is not dangerous. However, "Restart parent job skipping …" is very long so the button would be awfully big.

pevik: I'd personally prefer 2 buttons to save extra click (even with short text + full text in the tooltip. But even this implemented is a great usability improvement, thank you.

> Styling seems off here. Tho I guess that's a general issue and not just here. The popup isn't using the button's "danger" variant.

Once I see it live I'll have look, maybe it can be fixed with CSS modification (change CSS selector).

Martchus: Ok, I guess one could make it a 2nd button with a shorter text and a bootstrap tooltip with the full explanation.

PR #7626, opened 2026-07-15
feat: Improve scalability of dependency tree computation

Martchus:
* Optimize finding the latest job using raw SQL query
    * Reduce lookup time from almost 7 minutes to 3 seconds on my system with a clone chain of 742 jobs with parallel dependencies
* Avoid warning `Deep recursion on subroutine "OpenQA::WebAPI::Controller::Test::_add_job"` if there are long clone chains
* Evaluate `is_child_of` only for the latest job (and not each and every job in the chain); that changes the behavior a little bit requiring the tests change
    * The number of displayed jobs did not notably increase in the big dependency trees I checked.

Related ticket: https://progress.opensuse.org/issues/204120

### DIFF
--- a/lib/OpenQA/Schema/Result/Jobs.pm
+++ b/lib/OpenQA/Schema/Result/Jobs.pm
@@ -2046,8 +2046,20 @@ sub incomplete_ancestors ($self, $limit = -1) {
 }
 
 sub latest_job ($self) {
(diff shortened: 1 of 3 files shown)


okurz (approved):

Wabri (approved):
Nice

PR #7627, opened 2026-07-15
feat: Limit dependency tree

Martchus:
Based on https://github.com/os-autoinst/openQA/pull/7626 (so check out the 2nd commit to review this change individually).

---

* Limit the lookup breadth and depth to avoid timeouts as we do in most
  other routes.
* These limits will probably not be actually enforced in production as even
  our biggest dependency trees are smaller; this is mainly to follow our
  practice of avoiding unbounded computations and deep recursions (which
  Perl would warn about).
    * This even applies if jobs are restarted over 700 times (as it
      sometimes happens on o3) because the whole clone chain is not
      displayed anyway and also not factored into these limits (where the
      depth refers to dependency depth; not cloning depth).

Related ticket: https://progress.opensuse.org/issues/204120

### DIFF
--- a/lib/OpenQA/WebAPI/Controller/Test.pm
+++ b/lib/OpenQA/WebAPI/Controller/Test.pm
@@ -22,6 +22,8 @@ use List::Util qw(any first min);
 use HTTP::Status qw(:constants);
 
 use constant DEPENDENCY_DEBUG_INFO => $ENV{OPENQA_DEPENDENCY_DEBUG_INFO};
+use constant DEPENDENCY_BREADTH_LIMIT => $ENV{OPENQA_DEPENDENCY_BREADTH_LIMIT} // 200;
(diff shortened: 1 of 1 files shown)


lib/OpenQA/WebAPI/Controller/Test.pm
```diff
@@ -1164,7 +1164,7 @@ sub _add_dependency_to_node ($node, $parent, $dependency_type) {
     }
 }
 
-sub _add_job ($dependency_data, $job, $as_child_of, $preferred_depth) {
+sub _add_job ($dependency_data, $job, $as_child_of, $preferred_depth, $breadth_limit = 200, $depth_limit = 25) {
```
Wabri: nitpick: I don't like magic number, can you please put somewhere the meanings of those 200 and 25 values?

okurz: In this syntax I consider it clear that those are default values for the well named variables but a named constant at the top of the file would be more conventional

okurz (approved):

perlpunk (approved):

PR #7629, opened 2026-07-16
feat(cli): improve connection error reporting

okurz:
Motivation:
When network connection failures occur, the response code defaults to 0.
This prints as "hit error 0", which is confusing and lacks clarity.

Design Choices:
Map code 0 to "connection error" while retaining other HTTP status codes.
Mock Mojo transaction errors to verify the fallback formatting in tests.

Benefits:
Provides much clearer error messages when network connectivity issues occur.

### DIFF
--- a/lib/OpenQA/Command.pm
+++ b/lib/OpenQA/Command.pm
@@ -132,11 +132,12 @@ sub retry_tx ($self, $client, $tx, $retries = undef, $delay = undef) {
         my $waited = time - $start;
         my $error = $new_tx->error;
         my $error_msg = $error ? " ($error->{message})" : '';
+        my $error_info = $res_code == 0 ? "connection error$error_msg" : "error $res_code$error_msg";
         my $url = $new_tx->req->url;
         my $port = $url->port // ($url->scheme && $url->scheme eq 'https' ? 443 : 80);
         my $target = ($url->host || 'localhost') . ":$port";
         print STDERR encode('UTF-8',
-"Request to $target failed, hit error $res_code$error_msg, retrying up to $retries more times after waiting … (delay: $delay; waited ${waited}s)\n"
+"Request to $target failed, hit $error_info, retrying up to $retries more times after waiting … (delay: $delay; waited ${waited}s)\n"
         );
         sleep $delay;
         $delay *= $factor;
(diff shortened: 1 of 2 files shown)


lib/OpenQA/Command.pm
```diff
@@ -132,11 +132,12 @@ sub retry_tx ($self, $client, $tx, $retries = undef, $delay = undef) {
         my $waited = time - $start;
         my $error = $new_tx->error;
         my $error_msg = $error ? " ($error->{message})" : '';
+        my $error_info = $res_code == 0 ? "connection error$error_msg" : "error $res_code$error_msg";
```
Wabri, line 135: question: No space between the message and the error? 
```suggestion
        my $error_info = $res_code == 0 ? "connection error $error_msg" : "error $res_code$error_msg";
```

okurz: This is on purpose to prevent spaces when there is no error. Also this is unchanged from before

Wabri: suggestion: I think something like this is not readable enough and a little too much

```suggestion
my $error_info = $res_code == 0
    ? ($error_msg eq '' ? "connection error" : $error_msg)
    : ($error_msg eq '' ? "error $res_code$error_msg" : "error $res_code");
```

okurz: we shouldn't compare a string multiple times against an empty string this way

t/05-scheduler-serialize-directly-chained-dependencies.t
```diff
@@ -142,7 +142,7 @@ $cluster_info{$_}->{state} = SCHEDULED for (1, 9);
 # provide a sort function to control the order between multiple children of the same parent
 my %sort_criteria = (12 => 'anfang', 7 => 'danach', 6 => 'mitte', 8 => 'nach mitte', 1 => 'zuletzt');
 my $sort_function = sub ($items) {
-    return [sort { ($sort_criteria{$a} // $a) cmp ($sort_criteria{$b} // $b) } @$items];
+    return [sort { ($sort_criteria{$a} // $a) cmp($sort_criteria{$b} // $b) } @$items];
```
perlpunk: Unrelated change, likely due to outdated perltidy on your side.

okurz: that's one of the few problems of using Slowroll vs. Tumbleweed :)

Fixed

perlpunk (approved):

Wabri (approved):

Martchus (approved):

perlpunk:
rebase to avoid the CI failure

PR #7632, opened 2026-07-17
Revert #7620+#7604 due to https://progress.opensuse.org/issues/204252

okurz:
* Revert "refactor: Move database setup of cache service to separate module"
* Revert "feat(worker): Add local locking on worker-instance scope"
* Revert "feat(worker): Allow configuring lock to throttle job execution"
* Revert "feat(worker): Re-evaluate unavailability due to job lock quickly"
* Revert "refactor: Rename `OpenQA::CacheService::DB`"
* Revert "test(worker): Cover rejecting job when acquiring job lock failed"
* Revert "feat: Change worker job lock limit to be in line with config example"

I prepared this draft in case we need it.

https://progress.opensuse.org/issues/204252


perlpunk:
ui/02-csrf.t test failure:
https://progress.opensuse.org/issues/204255

perlpunk:
Likely not necessary and fixed with #7633

okurz:
> Likely not necessary and fixed with #7633

I am confident as well that we shouldn't need this revert

PR #7635, opened 2026-07-17
feat: extend API timestamp mismatch error message with diagnostic info

asmorodskyi:
To simplify debugging clock synchronization issues
 between local hosts (e.g. workers/CLI) and the
 web UI host, include key caller and time context
 in the error response.
The mismatch error now reports:
- Client IP
- Server epoch time
- Client epoch time
- Configured tolerance

### DIFF
--- a/lib/OpenQA/Shared/Controller/Auth.pm
+++ b/lib/OpenQA/Shared/Controller/Auth.pm
@@ -29,7 +29,7 @@ sub check ($self) {
     my $schema = OpenQA::Schema->singleton;
     my $api_key = $schema->resultset('ApiKeys')->find({key => $key});
     if ($api_key) {
-        if ($self->_is_timestamp_valid(time, $remote_timestamp)) {
+        if (($self->_is_timestamp_valid(time, $remote_timestamp))[0]) {
             my $exp = $api_key->t_expiration;
             # It has no expiration date or it's in the future
             if (!$exp || $exp->epoch > time) {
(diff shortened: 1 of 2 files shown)


okurz (changes requested):
* s/extetnd/extend/
* some example output from actual operation would be good

lib/OpenQA/Shared/Controller/Auth.pm
```diff
@@ -206,7 +206,10 @@ sub _key_auth ($self, $reason, $key) {
           = sprintf 'Rejecting authentication for user "%s" with ip "%s", valid key "%s", secret "%s"',
           $username, $self->tx->remote_address, $api_key->key, $api_key->secret;
         if (!$self->_is_timestamp_valid($our_timestamp, $remote_timestamp)) {
-            $reason = 'timestamp mismatch - check whether clocks on the local host and the web UI host are in sync';
+            my $tolerance = defined $self->config->{api_hmac_time_tolerance} ? $self->config->{api_hmac_time_tolerance} . 's' : 'unknown';
```
Martchus: Don't repeat `$self->config->{api_hmac_time_tolerance}`. Maybe it would make more sense to compute the reason in `_is_timestamp_valid` which presumably also already deals with that configuration.

Martchus (changes requested):

lib/OpenQA/Shared/Controller/Auth.pm
asmorodskyi: done . WDYT now ?

t/05-scheduler-serialize-directly-chained-dependencies.t
```diff
@@ -142,7 +142,7 @@ $cluster_info{$_}->{state} = SCHEDULED for (1, 9);
 # provide a sort function to control the order between multiple children of the same parent
 my %sort_criteria = (12 => 'anfang', 7 => 'danach', 6 => 'mitte', 8 => 'nach mitte', 1 => 'zuletzt');
 my $sort_function = sub ($items) {
-    return [sort { ($sort_criteria{$a} // $a) cmp ($sort_criteria{$b} // $b) } @$items];
+    return [sort { ($sort_criteria{$a} // $a) cmp($sort_criteria{$b} // $b) } @$items];
```
perlpunk: Unrelated change, probably because of an outdated perltidy

lib/OpenQA/Shared/Controller/Auth.pm
```diff
@@ -137,11 +137,18 @@ sub _is_timestamp_valid ($self, $our_timestamp, $remote_timestamp) {
...
-    return 1 if (abs($our_timestamp - $remote_timestamp) <= $tolerance);
+    return (1, undef) if (abs($our_timestamp - $remote_timestamp) <= $tolerance);
     $log->debug(
 qq{Timestamp mismatch over ${tolerance}s; our_timestamp: $our_timestamp, X-API-Microtime (from worker): $remote_timestamp}
     );
-    return 0;
+    my $tolerance_str
+      = defined $self->config->{api_hmac_time_tolerance}
+      ? $self->config->{api_hmac_time_tolerance} . 's'
+      : 'unknown';
```
Martchus:
```suggestion
    my $tolerance_str = defined $tolerance ? "${tolerance}s" : 'unknown';
```

Martchus (changes requested):

perlpunk (approved):

lib/OpenQA/Shared/Controller/Auth.pm
asmorodskyi: done

okurz (approved):

Martchus (approved):

